PR #1840, opened 2018-11-24
Document that alacritty requires winpty-agent on windows

zacps:
I'm not sure this is the best place to put this information because it's only supposed to be for *from-source* builds, but this affects binary releases as well. I don't know if there's a better place to put it though.

Fixes https://github.com/jwilm/alacritty/issues/1818

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -166,11 +166,8 @@ nix-shell -A alacritty '<nixpkgs>'
 
 #### Windows
 
-On windows you will need to have the `{architecture}-pc-windows-msvc` toolchain installed as well as [Clang 3.9 or greater](http://releases.llvm.org/download.html).
-
-Alacritty requires winpty on windows to emulate UNIX's PTY API. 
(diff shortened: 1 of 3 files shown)


chrisduerr:
Hmm, I'm not sure if the INSTALL.md is the best place to document this. I'd imagine most windows users will just download the exe from the releases section directly, so it might be better suited for the README.md. Maybe add a short comment to the `Others` section?

rennex:
I was also stumped by the lack of winpty-agent for a while, because without it, an alacritty window appeared and instantly closed without explanation. Only when I ran it from a console window, I saw the error messages.

One idea might be to just distribute a zip file on the releases page, containing both the required exe files.

chrisduerr:
@rennex I think that's a good idea. @zacps would you mind adding that to this PR too? I believe we'd just have to add the `zip` package as a dependency and then we could use it in our `before_deploy.sh`.

PR #1856, opened 2018-12-01
Use 'login' as the default shell on macOS

neersighted:
If the `shell:` key is not specified in alacritty.yml, default to `login -fp $USER` on macOS, following the behavior of Terminal.app and iTerm2.

This isn't the most elegant fix, but it works well for now in lieu of a more robust, platform-independent solution.


neersighted:
After conferring with @chrisduerr on IRC, we've decided to use bash a wrapper to launch a login shell, as `login(1)` discards the PWD, and we duplicate everything it does. `std::process::Command` is incapable of manually setting argv[0]; so I will open a PR against rust to resolve this later, as bash is guaranteed to be in all past macOS releases, but not necessarily future releases.

neersighted:
I just noticed the missing CHANGELOG entry as you commented. That should be done now. Let me know if you have any more concerns.

neersighted:
Good catch, @dm1try. I forgot to update how `$SHELL` is set. It should be good now.

benmp:
Does this mean I can remove

```
shell:
  program: login
  args: ["-fp", "myuser.name"]
```

from my alacritty.config on mac?

neersighted:
Yes, it does. If you've set alacritty to start a login shell in any way `/bin/myshell --login`, `login -fp $USERNAME`, etc, you no longer need to do that, as long as you used `chsh` to change your login shell.

PR #1857, opened 2018-12-02
Make windows config location more sensible

zacps:
Contrary to my initial idea this does *not* remove platform specific code, mostly because after dealing with fallbacks and supporting the old windows location it's not worth it.

This changes the configuration file location from (in fallback order):

* `%HOME%\alacritty.yml`
* `%USERPROFILE%\alacritty.yml`
* `GetUserProfileDirectory()\alacritty.yml`

To:

* `%APPDATA%\Roaming\alacritty\alacritty.yml`

If a configuration file is found in the old location a warning is emitted pointing to the new location.

Fixes https://github.com/jwilm/alacritty/issues/1660

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Moved the Windows configuration location from %USERPROFILE%\alacritty.yml (usually) to %APPDATA%\Roaming\alacritty\alacritty.yml
+- Windows configuration location has been moved from %USERPROFILE\alacritty.yml
+    to %APPDATA%\Roaming\alacritty\alacritty.yml
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New configuration field `visual_bell.color` allows changing the visual bell color
 - Crashes on Windows are now also reported with a popup in addition to stderr
 
+### Changed
+
+- Moved the Windows configuration location from %USERPROFILE%\alacritty.yml (usually) to %APPDATA%\Roaming\alacritty\alacritty.yml
+
 ### Fixed
 
 - Fix color issue in ncurses programs by updating terminfo pairs from 0x10000 to 0x7FFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1523,14 +1523,22 @@ impl Config {
 
     #[cfg(windows)]
     pub fn installed_config<'a>() -> Option<Cow<'a, Path>> {
-        if let Some(mut path) = ::std::env::home_dir() {
-            path.push("alacritty");
-            path.set_extension("yml");
-            if path.exists() {
-                return Some(path.into());
-            }
+        #[allow(deprecated)]
+        let old = ::std::env::home_dir()
+            .map(|path| path.join("alacritty.yml"));
+        let new = dirs::config_dir()
+            .map(|path| path.join("alacritty\\alacritty.yml"));
+        // TODO: Remove old configuration location warning (Deprecated 03/12/2018)
+        if let Some(old) = old.as_ref().filter(|old| old.exists()) {
+            warn!(
+                "Found old configuration at: '{}'. The file should be moved to the new location: '{}'.", 
+                old.to_string_lossy(), 
+                new.as_ref().map(|new| new.to_string_lossy()).unwrap_or_else(|| Cow::from("(UNKNOWN)"))
+            );
         }
-        None
+
+        new.filter(|new| new.exists()).map(Cow::from)
+            .or_else(|| old.filter(|old| old.exists()).map(Cow::from))
     }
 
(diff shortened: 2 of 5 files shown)


src/config.rs
```diff
@@ -1513,14 +1513,22 @@ impl Config {
...
-            path.set_extension("yml");
-            if path.exists() {
-                return Some(path.into());
-            }
+        #[allow(deprecated)]
+        let old = ::std::env::home_dir()
+            .map(|path| path.join("alacritty.yml"));
+        let new = dirs::config_dir()
+            .map(|path| path.join("alacritty\\alacritty.yml"));
+        // TODO: Remove old configuration location warning (Deprecated 03/12/2018)
```
chrisduerr: I think this comment should be moved from within the method to before the method definition. That should give it a little better visibility so it doesn't get lost.

src/config.rs
```diff
@@ -1513,14 +1513,22 @@ impl Config {
...
+        if let Some(old) = old.as_ref().filter(|old| old.exists()) {
+            warn!(
+                "Found old configuration at: '{}'. The file should be moved to the new location: '{}'.",
```
chrisduerr: The file is not necessarily an `old` configuration. It's just the old configuration location. Stating that clearly takes a few too many words imho, but just removing the `old` here should prevent confusion I think. Just something like `Found configuration file at: '{}'`.

Also in the change log it's noted that the old path "usually" was %USERPROFILE%. Would it be possible that the old and new path is the same, thus creating an warning the user can't resolve, telling him to move it from /x/y/z to /x/y/z?

src/config.rs
```diff
@@ -1513,14 +1513,22 @@ impl Config {
...
+        let old = ::std::env::home_dir()
+            .map(|path| path.join("alacritty.yml"));
+        let new = dirs::config_dir()
+            .map(|path| path.join("alacritty\\alacritty.yml"));
+        // TODO: Remove old configuration location warning (Deprecated 03/12/2018)
+        if let Some(old) = old.as_ref().filter(|old| old.exists()) {
+            warn!(
+                "Found old configuration at: '{}'. The file should be moved to the new location: '{}'.", 
+                old.to_string_lossy(), 
+                new.as_ref().map(|new| new.to_string_lossy()).unwrap_or_else(|| Cow::from("(UNKNOWN)"))
```
chrisduerr: I think an `expect(...)` here makes more sense.

src/config.rs
```diff
@@ -1534,9 +1542,12 @@ impl Config {
 
     #[cfg(windows)]
     pub fn write_defaults() -> io::Result<Cow<'static, Path>> {
-        let path = ::std::env::home_dir()
-            .ok_or_else(|| io::Error::new(io::ErrorKind::NotFound, "could not find profile directory"))
-            .and_then(|mut p| {p.push("alacritty"); p.set_extension("yml"); Ok(p)})?;
+        let mut path = dirs::config_dir()
+            .ok_or_else(|| io::Error::new(io::ErrorKind::NotFound, "could not find profile directory"))?;
```
chrisduerr: Line length should not exceed 100 characters.

src/config.rs
```diff
@@ -1534,9 +1542,12 @@ impl Config {
...
+        if !path.exists() {
+            fs::create_dir(&path)?
+        }
```
chrisduerr: The `create_dir_all` method might make this a little simpler. It doesn't create a directory if it already exists.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Extra padding is not evenly spread around the terminal by default anymore
+- Moved the Windows configuration location from %USERPROFILE%\alacritty.yml (usually) to %APPDATA%\Roaming\alacritty\alacritty.yml
```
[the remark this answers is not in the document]
chrisduerr: If there are multiple different locations the old configuration file could have been add, just saying `Moved the Windows configuration file location to ...` would also be fine.

src/config.rs
zacps: `%APPDATA%` is a child of `%USERPROFILE%`, so they should never point to the same location.

src/config.rs
zacps: Do you mean expect on `dirs::config_dir` and `std::env::home_dir`? Those are the failable part of this.

zacps:
Using `dirs` instead of `xdg` is *possible*, but ensuring they have identical behaviour is difficult. 

They follow the same usual case if `XDG_CONFIG_HOME` is set. However, the way it's written now it also searches `XDG_CONFIG_DIRS` which `dirs` has no equivalent for.

I think it's probably not worth the effort.

src/config.rs
chrisduerr: I was specifically talking about the `unwrap_or_else(|| Cow::from("(UNKNOWN)"))`, since that could actually lead to unexpected behavior. If there's no good way to recover, Alacritty should crash with a helpful error message rather than trying to fill in the gaps with garbage.

zacps: It's the equivalent of failing if `~` doesn't exist on linux, so I think it's probably reasonable to panic.

PR #1858, opened 2018-12-03
Fix-#1833:Detach Child process to avoid zombie processes

### LINKED ISSUE #1833: Disown child processes to avoid zombie processes
When you click an URL which triggers the opening of the browser, **the browser's process will stay as a Zombie process when closed**.

You can have multiple Z processes as long as you don't close alacritty.

As discussed on IRC 5 minutes ago, it may be interesting to **launch the browser as a daemon process and disown it** to avoid this issue.

vsag96:
Attempts to fix #1833. While I am new to learning rust(Very new). I intend to use alacritty for a long time my attempt to give something back. So if I have done something wrong or something you would like to see added please let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed a bad type conversion which could cause underflow on a window resize
 - Alacritty now spawns a login shell on macOS, as with Terminal.app and iTerm2
+- Fixed zombie processes sticking around after launching URLs
 
 ## Version 0.2.3
 
--- a/src/input.rs
+++ b/src/input.rs
@@ -238,8 +238,8 @@ impl Action {
                 trace!("running command: {} {:?}", program, args);
 
                 match start_daemon(program, args) {
-                    Ok(child) => {
-                        debug!("spawned new proc with pid: {}", child.id());
+                    Ok(_) => {
+                        debug!("spawned new proc");
                     },
                     Err(err) => {
                         warn!("couldn't run command: {}", err);
--- a/src/input.rs
+++ b/src/input.rs
@@ -20,10 +20,7 @@
 //! determine what to do when a non-modifier key is pressed.
 use std::borrow::Cow;
 use std::mem;
-use std::process::Command;
 use std::time::Instant;
-#[cfg(not(windows))]
-use std::os::unix::process::CommandExt;
 
 use copypasta::{Clipboard, Load, Buffer as ClipboardBuffer};
 use glutin::{ElementState, MouseButton, TouchPhase, MouseScrollDelta, ModifiersState, KeyboardInput};
@@ -35,6 +32,7 @@ use index::{Line, Column, Side, Point};
 use term::SizeInfo;
 use term::mode::TermMode;
 use util::fmt::Red;
+use util::start_daemon;
 
 pub const FONT_SIZE_STEP: f32 = 0.5;
 
@@ -239,24 +237,7 @@ impl Action {
             Action::Command(ref program, ref args) => {
                 trace!("running command: {} {:?}", program, args);
 
-                #[cfg(not(windows))]
-                let spawned = Command::new(program)
-                    .args(args)
-                    .before_exec(|| {
-                        // Detach forked process from Alacritty. This will cause
-                        // init or whatever to clean up child processes for us.
-                        unsafe { ::libc::daemon(1, 0); }
-                        Ok(())
-                    })
-                    .spawn();
-
-                #[cfg(windows)]
-                let spawned = Command::new(program)
-                    .args(args)
-                    .spawn();
-
-                match spawned
-                {
+                match start_daemon(program, args) {
                     Ok(child) => {
                         debug!("spawned new proc with pid: {}", child.id());
                     },
@@ -557,30 +538,10 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         let mut args = launcher.args().to_vec();
         args.push(text);
 
-        #[cfg(not(windows))]
-        let _spawned = Command::new(launcher.program())
-                    .args(&args)
-                    .before_exec(|| {
-                        // Detach forked process from Alacritty. This will cause
-                        // init or whatever to clean up child processes for us.
-                        unsafe { ::libc::daemon(1, 0); }
-                        Ok(())
-                    })
-                    .spawn();
-        #[cfg(windows)]
-        let _spawned = Command::new(launcher.program())
-                    .args(&args)
-                    .spawn();
-
-        match _spawned
-                {
-                    Ok(child) => {
-                        debug!("spawned new proc with pid: {}", child.id());
-                    },
-                    Err(err) => {
-                        warn!("couldn't run command: {}", err);
-                    },
-                }
+        match start_daemon(launcher.program(), &args) {
+            Ok(_) => debug!("Launched: {} {:?}", launcher.program(), args),
+            Err(_) => warn!("Unable to launch: {} {:?}", launcher.program(), args),
+        }
 
         Some(())
     }
--- a/src/input.rs
+++ b/src/input.rs
@@ -557,10 +557,30 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         let mut args = launcher.args().to_vec();
         args.push(text);
 
-        match Command::new(launcher.program()).args(&args).spawn() {
-            Ok(_) => debug!("Launched: {} {:?}", launcher.program(), args),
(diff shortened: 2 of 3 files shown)


vsag96:
So I forgot to add the compiler flags (cfg) which caused windows build to fail. So add changes and forced pushed them and this fixes it.

src/input.rs
```diff
@@ -557,10 +557,30 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         let mut args = launcher.args().to_vec();
         args.push(text);
 
-        match Command::new(launcher.program()).args(&args).spawn() {
-            Ok(_) => debug!("Launched: {} {:?}", launcher.program(), args),
-            Err(_) => warn!("Unable to launch: {} {:?}", launcher.program(), args),
-        }
+        #[cfg(not(windows))]
+        let _spawned = Command::new(launcher.program())
```
chrisduerr: The variable shouldn't be prefixed with `_`. This is usually used to prevent warnings about unused variables and thus could lead to issues when refactoring code in the future.

src/input.rs
```diff
@@ -557,10 +557,30 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+
+        match _spawned
+                {
+                    Ok(child) => {
+                        debug!("spawned new proc with pid: {}", child.id());
+                    },
+                    Err(err) => {
+                        warn!("couldn't run command: {}", err);
+                    },
+                }
```
chrisduerr: The code formatting is a bit messed up here. It should look like this:
```
match _spawned {
    ...
}
```

src/input.rs
```diff
@@ -557,10 +557,30 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+                    .args(&args)
+                    .spawn();
+
+        match _spawned
+                {
+                    Ok(child) => {
+                        debug!("spawned new proc with pid: {}", child.id());
+                    },
+                    Err(err) => {
+                        warn!("couldn't run command: {}", err);
```
chrisduerr: Log messages should stay the same.

vsag96:
Thanks for the review, I have two questions

1.  Should the function be moved to utils.rs in the same PR? I think both changes should be done in different PR's, if you feel otherwise please let me know. 
 
2. Should the change log entry be in the unreleased part of the file?

chrisduerr:
> Should the function be moved to utils.rs in the same PR? I think both changes should be done in different PR's, if you feel otherwise please let me know.

I think it would be best to do it in this PR, rather than separating these changes and potentially introducing additional code changes in between them which might complicate things. It should just be a very minimal change either way, so there shouldn't be a problem.

> Should the change log entry be in the unreleased part of the file?

Exactly.

vsag96:
Hi,
I am getting the error below, When I use the rustc --explain I got the gist of the error, so when I tried adding the type inference to the function like `start_daemon::<io::Result<Child>>(program,args)`. I get an error saying I am passing only 1 argument. Can you help me out here, I am sorry.

`error[E0282]: type annotations needed                                                                                                                                                                              
   --> src/input.rs:240:30                                                                                                                                                                                         
    |                                                                                                                                                                                                              
240 |                 let spawned= start_daemon(program,args);                                                                                                                                                     
    |                              ^^^^^^^^^^^^ cannot infer type for `S`                                                                                                                                          
       `

chrisduerr:
@vsag96 The type parameters `<S,I>` for static dyspatch are not necessary.

vsag96:
Oh you're right, forgot to remove them, are you okay with the current changes should I rebase and push? or do you have any suggestions?

src/input.rs
```diff
@@ -238,25 +236,9 @@ impl Action {
...
-                    .spawn();
-
-                #[cfg(windows)]
-                let spawned = Command::new(program)
-                    .args(args)
-                    .spawn();
-
-                match spawned
-                {
+                //.collect::<Vec<_>>();
```
chrisduerr: Unneeded comments should be removed.

src/input.rs
```diff
@@ -238,25 +236,9 @@ impl Action {
...
+                let spawned= start_daemon(program,args);
+                match spawned{
```
chrisduerr: This can be rewritten to just the following:

```rust
match start_daemon(program, args) {
```

Please make sure that spaces are set appropriately to ensure readability of the code.

src/input.rs
```diff
@@ -556,10 +538,14 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         let launcher = self.mouse_config.url.launcher.as_ref()?;
         let mut args = launcher.args().to_vec();
         args.push(text);
-
-        match Command::new(launcher.program()).args(&args).spawn() {
-            Ok(_) => debug!("Launched: {} {:?}", launcher.program(), args),
-            Err(_) => warn!("Unable to launch: {} {:?}", launcher.program(), args),
+        let spawned=start_daemon(launcher.program(),&args);
+        match spawned {
```
chrisduerr: This can also be rewritten to `match start_daemon(launcher.program(), &args) {`.

src/input.rs
```diff
@@ -556,10 +538,14 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+            Ok(child) => {
+                debug!("Launched: {} {:?}", launcher.program(), args);
+            },
+            Err(err) => {
+                warn!("Unable to launch: {} {:?}", launcher.program(), args);
+            },
```
chrisduerr: The variables `child` and `err` are not used here and should not be defined.

These lines should stay unchanged since it's not necessary to alter anything here.

src/util.rs
```diff
@@ -92,3 +96,24 @@ mod tests {
...
+        .spawn();
+
+    #[cfg(windows)]
+    let spawned = Command::new(program)
+    .args(args)
+    .spawn();
+
+    return spawned;
+
+}
```
chrisduerr: Since 100% of the method body is split between Windows and macOS/Linux/BSD, two separate methods should be defined and annotated.

It should look something like this instead:
```rust
#[cfg(not(windows))]
pub fn start_daemon(...) -> ... {
    ...
}

#[cfg(windows)]
pub fn start_daemon(...) -> ... {
    ...
}
```

Instead of using `return spawned;`, which is unidiomatic Rust code, you should also just remove the semicolon after `.spawn();` and delete the `return ...;` statement completely.

Please also make sure there's a newline at the end of the file. This is currently missing.

vsag96:
@chrisduerr This is ready for merging I think.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed a bad type conversion which could cause underflow on a window resize
 - Alacritty now spawns a login shell on macOS, as with Terminal.app and iTerm2
+- Fixed the problem of zombie processes by detaching the child process.
```
chrisduerr: No punctuation should be at the end of change log comments.

I'd also like a mention to the feature in question here (url-launching). Without that it's not really clear under which circumstances these zombies were created.

src/input.rs
chrisduerr: These should be changed back to single-line, the way they have been before this PR.

src/util.rs
```diff
@@ -92,3 +96,27 @@ mod tests {
...
+
+#[cfg(windows)]
+pub fn start_daemon(program: &str,args: &Vec<String>) -> io::Result<std::process::Child>{
+    Command::new(program)
+        .args(args)
+        .spawn()
+
+}
+
```
chrisduerr: The three additional whitespaces inside the `start_daemon` methods should be removed.

Additionally a whitespace should be added before `args: &Vec<String>` and after `<std::process::Child>{`.

vsag96:
So two things

1) Is this okay for a changelog message or is it too hairy to be in  changelog, English is not my first language.

When ever a url is launched from alacritty the browser process will stay as a zombie process to avoid this launch the child process as a daemon and disown it so the init system will take of the launched browser process.

2) Thanks for all the comments, even at work we don't have such rigorous code review, not even for the core and important features of a service or product. The ironic part is we are paid to do the job. It really warms my heart that you strive to keep the code readable and you read others code and let them know all the good practices and subtle things. You are awesome man!

chrisduerr:
> You are awesome man!

Thanks, but this only works out because there are so many other people happily sending PRs and working on bugs and features!

> Is this okay for a changelog message

I don't think your original message was that far off. It was short and concise, which is the most important thing, it was just missing more specifics about the issue itself.

I'd probably just change it slightly like this:
```
Fixed zombie processes sticking around after launching URLs
```

With the change log messages it doesn't matter how it's done, since that's what our commit messages are for. This is just for the enduser so he knows which issues have been resolved. So it's not necessary to go into much detail on how the problem was resolved.

vsag96:
@chrisduerr done.

Issue #1860, opened 2018-12-03
Some font characters look cut-off (incomplete rendering)


talal:
OS: macOS 10.14.1
Alacritty: 0.2.3 ([config-file](https://github.com/talal/dotfiles/blob/master/alacritty/alacritty.yml))
tmux: 2.8 ([config-file](https://github.com/talal/dotfiles/blob/master/tmux/))
font: Pragmata Pro Mono

Some font characters are not fully rendered, they are cut at the top. It is not that big of an issue but with some characters the difference is noticeable. I have particularly noticed it with `u` and `v`.

tmux is not the issue and neither is neovim since this issue also occurs outside neovim/tmux. e.g. `cat some-file`. The same font looks okay in iTerm.

Please refer to the screenshot below and look at the `u`, `v`, `x`, and `y` characters, their top part is incomplete and looks cut-off on the Alacritty side:
![font-cut](https://user-images.githubusercontent.com/3526562/49393976-07bcb300-f733-11e8-8b9c-291010dc160a.png)

Other fonts don't have this problem, I tried Monaco and it looked okay.

Also: 
- `use_thin_strokes` doesn't work, but I already saw another issue regarding that.
- `start_maximized` also doesn't work, it always starts with the `window.dimensions`. This feature is not that important to me.

Your help on this issue would be appreciated. To be honest, these issues don't bother me much, the speed difference in comparison with iTerm is worth it 😄

dm1try:
@talal I was unable to reproduce the issue like on your screenshot above (`osx 10.14.1/alacritty master/Pragmata Pro Mono Regular/size`)
<img width="251" alt="screen shot 2018-12-04 at 02 38 42" src="https://user-images.githubusercontent.com/486807/49408540-bed62000-f76d-11e8-8945-df1e1132f33a.png">
 (disabling all font settings except antialising)
<img width="251" alt="screen shot 2018-12-04 at 02 29 40" src="https://user-images.githubusercontent.com/486807/49408286-b6c9b080-f76c-11e8-8c98-41d156573e7d.png">

> start_maximized also doesn't work, it always starts with the window.dimensions. This feature is not that important to me.

this one is fixed on master 0d47cd25b90deecc2dbf40c8361d2310049fdde7

>  the speed difference in comparison with iTerm

did you try to enable metal renderer in iTerm?

chrisduerr:
[...]

PR #1866, opened 2018-12-06
[macos] [doc] add instructions to enable font smoothing

talal:
Doc changes as discussed in #1860


talal:
Thank you for the extensive review :)

I agree that a long text can be off-putting for users and many will not read it. All your suggestions were good and the PR has been updated to reflect them.

PR #1869, opened 2018-12-06
Rust 2018, NLL fixes

jwilm:
This updates Alacritty to Rust 2018 while fixing old workarounds for lack of NLLs. Almost all `// borrowck` commens have been resolved except for two instances of overlapping mutable borrows not quite covered by the initial NLL support.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ build = "build.rs"
 description = "GPU-accelerated terminal emulator"
 readme = "README.md"
 homepage = "https://github.com/jwilm/alacritty"
+edition = "2018"
 
 [workspace]
 members = [
--- a/build.rs
+++ b/build.rs
@@ -12,15 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 #[cfg(windows)]
-extern crate embed_resource;
+use embed_resource;
 #[cfg(windows)]
-extern crate tempfile;
+use tempfile;
 #[cfg(windows)]
-extern crate reqwest;
+use reqwest;
 #[cfg(windows)]
-extern crate zip;
-
-extern crate gl_generator;
+use zip;
 
 use gl_generator::{Api, Fallbacks, GlobalGenerator, Profile, Registry};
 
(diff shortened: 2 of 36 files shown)


chrisduerr:
IIRC the Rust 2018 edition shouldn't require `extern crate` anymore, right? That should probably be removed too.

I think there was another rustfix command for idiomatic Rust 2018 code vs correct Rust 2018 code, though I'd have to look that up myself.

jwilm:
WRT extern crate, the cargo —fix command didn’t change it. I’m indifferent on whether it’s included or not, but I guess if it’s considered idiomatic not to have it, we should remove those too.

chrisduerr:
@jwilm Looking at `cargo fix --help`, there seems to be `cargo fix --edition` and `cargo fix --edition-idioms`. I'd try running that and just checking out what rustfix does, I think that should resolve the extern crate thing.

jwilm:
Ah, I didn't run the `-idioms` variation

chrisduerr:
Looking at the output of the edition idioms, it looks like there's **a lot** of work to be done. Instead of just simply running it and everything's good to do.

I'd propose just working on these separately. This would give people the chance to fix one idiom issue at a time, instead of having to do all at the same time.

chrisduerr:
I've managed to apply the fixes running `cargo fix --edition-idioms --broken-code` which is supposed to break a lot of stuff but it worked out great for Alacritty.

I've had to manually touch up some changes, but after doing a full review of the new changes, everything's looking good.

Some `#[macro_use] extern crate` and `#[cfg(not(Linux))] extern crate` imports are still present. I'd remove those myself, but I don't want to rely on travis to provide me feedback about the success of my changes, so I'll leave that to Windows/macOS users.

neersighted:
I just wanted to note that `macro_use` should no longer be necessary as macros can be imported using `use`.

chrisduerr:
@neersighted I'm aware of that, however my attempt to remove these lead to unnecessarily much work, since especially serde macros are used everywhere.

neersighted:
Fair enough. That can always be cleaned up later.

chrisduerr:
I've decided to remove some more `extern crate` imports which import macros. Though `log` and `serde` are still untouched in the `src/lib.rs`.

I think that should be plenty for this PR, so once CI is finished this can be merged. Further changes can be submitted in separate PRs, which will make things much easier since rebasing this PR would be a major pain.

PR #1871, opened 2018-12-06
use tool lints for clippy allow/deny lint attributes

matthiaskrgr:


winpty/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg_attr(feature = "cargo-clippy", deny(clippy, if_not_else, enum_glob_use, wrong_pub_self_convention))]
+#![cfg_attr(feature = "cargo-clippy", deny(clippy::all, clippy::if_not_else, clippy::enum_glob_use, clippy::wrong_pub_self_convention))]
```
matthiaskrgr: @zacps  hmm, not sure why I missed this..
anyway I think this can just be `!#[deny(clippy::all, clippy::if_not_else, clippy::enum_glob_use, clippy::wrong_pub_self_convention)]` (without the `cfg_attr....` stuff)

zacps: Ah whoops, do you want to change that?

PR #1877, opened 2018-12-07
Add -x flag to tic command

### LINKED ISSUE #1876: "Unknown capability" error when installing terminfo
OS: macOS 10.14.1
Alacritty: built from `master`

I get multiple "unknown capability" errors when installing Alacritty's terminfo by following the instructions from the `INSTALL.md` file.

An example of the error:
```
"alacritty.info", line 27, col 43, terminal 'alacritty+common': unknown capability 'AX'
```

talal:
Fixes #1876 

Adding `-x` flag fixes the issue. With this flag, unknown terminfo capabilities are treated as user-defined.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -226,7 +226,7 @@ To choose a default terminal app, use Debian's `update-alternatives`.
 
 ```sh
 update-alternatives --config x-terminal-emulator
-``` 
+```
 
 ## Manual Page
 
@@ -303,5 +303,5 @@ instead.
 To install alacritty's terminfo entry globally:
 
 ```sh
-sudo tic -e alacritty,alacritty-direct alacritty.info
+sudo tic -xe alacritty,alacritty-direct alacritty.info
 ```


talal:
Does using Alacritty's terminfo definition offer any benefit over `xterm-256color`? 

I am asking because I share my dotfiles between two machines, and on the other one I use iTerm. This creates a problem as in order to get truecolor support inside tmux, I would have to add the following in the `.tmux.conf`:

```
set-option -ga terminal-overrides ",alacritty:Tc"
```

which breaks the truecolor support in iTerm.

chrisduerr:
There are certain features xterm supports, which Alacritty doesn't and the other way around.

The Alacritty terminfo is based on xterm-256color and Alacritty should probably work fine with it. However there is a significant difference between those files.

chrisduerr:
Feel free to test if you can work around this issue @talal, however I think just sticking to what we have is probably best.

Feel free to close this PR and the related issue yourself if you can't get it to work.

PR #1878, opened 2018-12-07
Add a contributing.md file

### LINKED ISSUE #1870: Need for contributing file
I have searched the repo but could not find it.This is a popular application it has over 13k stars, it's time that we have a contributing.md file. Off the top of my head it can talk about formatting the code and adding a changelog entry. It saves new people time and that's the first thing some people search for in a repository. Thoughts?

vsag96:
It contains a few rules for making a PR. @chrisduerr you can add the formatting or style changes that you want to see the folks do before opening a PR. I think the point 6 can be done with a fewer commands. Any thing you would like to be removed or added?

Attempts to fix #1870

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,68 @@
+## Contributing to Alacritty
+
+Looks like you are interested in contributing to alacritty, thank you for your time.The following are just guidelines except for the code formatting which should be adhered to. So if you know something that can be done in fewer commands, use that or
+even better submit a PR.
+
+This document will introduce some things that will get you started with 
+- Setting up local development environment.
+- Formatting Code before PR's 
+
+## Setting up local development environment.
+
+1. Clone a copy of Alacritty from source 
+
+    ``` git clone --origin upstream https://github.com/jwilm/alacritty.git ```
+
+   The above command does two things for you, it clones the repository into your local directory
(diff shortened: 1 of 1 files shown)


zacps:
First of all instructions for compiling from source are in `INSTALL.md`, so that should be pointed to instead of replicating the instructions. Another note about the IRC channel `#alacritty` could also be worthwhile.

I'm not sure having git workflow notes in `CONTRIBUTING.md` is a good idea. It can be pretty subjective and there is already a lot of information about for people who are really new. I never use `git pull` for example.

I think including something like @chrisduerr's checklist is a good idea.

From the Windows side it'd be good have a reminder that changes to `alacritty.yml` should usually be reflected in `alacritty_windows.yml` and `alacritty_macos.yml`.

As for formatting it's something that should be done; but there are a lot of open pull requests at the moment, so getting the number down a bit before formatting will help reduce unnecessary conflicts.

chrisduerr:
I've decided to rework the `CONTRIBUTING.md` file completely based on what I feel like are issues that people often run into, or which might be relevant to new contributors.

I agree with @nixpulvis that it is better not to document any specific workflow, since that is better left to other channels and highly subjective.

I'm sure there's plenty of typos and other things left to be fixed, so feedback would be much appreciated. I feel like this should set us up pretty well though. If I've missed anything critical, please let me know.

PR #1911, opened 2018-12-18
Send alt key with actual key in one flush

skliew:
The delay between the alt key and the actual received key might cause
certain key sequences to be missed, ex. when tmux has its escape-time
set to 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix color issue in ncurses programs by updating terminfo pairs from 0x10000 to 0x7FFF
 - Fix panic after quitting Alacritty on macOS
 - Tabs are no longer replaced by spaces when copying them to the clipboard
+- Alt modifier is no longer sent separately from the modified key
 
 ## Version 0.2.4
 
--- a/src/input.rs
+++ b/src/input.rs
@@ -687,14 +687,14 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
             self.ctx.clear_selection();
 
             let utf8_len = c.len_utf8();
-            if *self.ctx.received_count() == 0 && self.ctx.last_modifiers().alt && utf8_len == 1 {
-                self.ctx.write_to_pty(b"\x1b".to_vec());
-            }
-
             let mut bytes = Vec::with_capacity(utf8_len);
             unsafe {
                 bytes.set_len(utf8_len);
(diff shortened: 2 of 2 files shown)


src/input.rs
```diff
@@ -687,17 +687,21 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
             let utf8_len = c.len_utf8();
-            if *self.ctx.received_count() == 0 && self.ctx.last_modifiers().alt && utf8_len == 1 {
-                self.ctx.write_to_pty(b"\x1b".to_vec());
-            }
+            let mut all_bytes =
+                if *self.ctx.received_count() == 0 && self.ctx.last_modifiers().alt && utf8_len == 1 {
+                    b"\x1b".to_vec()
+                } else {
+                    Vec::new()
+                };
```
chrisduerr: I feel like it's not an optimal solution here to basically always create an empty vector when the alt key hasn't been pressed. Not necessarily only for performance (though that's also a factor), but mainly because of style issues.

Rather than creating a potentially empty vector and then pushing the other elements onto it, I think it would be a better idea to modify the `bytes` vector after the character has been written to it, before it's written to the pty. Like that the standard case that no alt key is pressed can be ignored most of the time.

skliew: @chrisduerr Thank you for the feedback. I have made some changes following your advice. Apparently it looks much simpler and straight forward now. Just realized that the unsafe `set_len` was only for the `encode_utf8` to work and we could easily `insert` another byte into the vector later on.

This is how I test if it's working:
- Run tmux with this in .tmux.conf `bind -n M-j select-pane -t .-`
- Run `split-window -v` so that we'd have two panes in a window
- Run `showkey -a` in both panes
- Spam `M-j`

Without the fix you might intermittently see this:
```
^[       27 0033 0x1b
j       106 0152 0x6a
```

Which means the Alt and j keys are captured separately by tmux and thus do not cause a pane switch.

Thanks and regards.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix color issue in ncurses programs by updating terminfo pairs from 0x10000 to 0x7FFF
 - Fix panic after quitting Alacritty on macOS
 - Tabs are no longer replaced by spaces when copying them to the clipboard
+- Send alt key along with actual key in a single flush
```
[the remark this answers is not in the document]
skliew: @chrisduerr I have changed the change log to use the message you suggested. Thank you for the feedback.

PR #1921, opened 2018-12-24
Fix line metrics

chrisduerr:
The height of the line drawing for underline and strikethrough has been
incorrectly substracted from the Y position instead of adding it.
Leading to an incorrect Y position for underline and strikeout by
exactly the width of the line.

By simply adding half of the height instead of substracting it, the
position should be corrected and the distanace between the underline and
the baseline should be increased.

### DIFF
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -432,13 +432,9 @@ impl Font {
 
         // Strikeout and underline metrics
         // CoreText doesn't provide strikeout so we provide our own
-        let underline_position =
-            (self.ct_font.underline_position() - descent)
-            .round() as f32;
-        let underline_thickness = self.ct_font.underline_thickness()
-            .round()
-            .max(1.) as f32;
-        let strikeout_position = (line_height as f32 / 2. - descent as f32).round();
+        let underline_position = (self.ct_font.underline_position() - descent) as f32;
+        let underline_thickness = self.ct_font.underline_thickness().max(1.) as f32;
+        let strikeout_position = line_height as f32 / 2. - descent as f32;
         let strikeout_thickness = underline_thickness;
 
         Metrics {
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -430,17 +430,17 @@ impl Font {
         let leading = self.ct_font.leading() as f64;
         let line_height = (ascent + descent + leading + 0.5).floor();
 
-        // Strikeout and underline metrics
-        // CoreText doesn't provide strikeout so we provide our own
+        // Strikeout and underline metrics.
+        // CoreText doesn't provide strikeout so we provide our own.
         let underline_position = (self.ct_font.underline_position() - descent) as f32;
-        let underline_thickness = self.ct_font.underline_thickness().max(1.) as f32;
-        let strikeout_position = line_height as f32 / 2. - descent as f32;
+        let underline_thickness = self.ct_font.underline_thickness() as f32;
+        let strikeout_position = (line_height / 2. - descent) as f32;
         let strikeout_thickness = underline_thickness;
 
         Metrics {
             average_advance,
             line_height,
-            descent: -(self.ct_font.descent() as f32),
+            descent: -(descent as f32),
             underline_position,
             underline_thickness,
             strikeout_position,
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -97,15 +97,13 @@ impl ::Rasterize for FreeTypeRasterizer {
 
         // Get underline position and thickness in device pixels
         let x_scale = full.size_metrics.x_scale as f32 / 65536.0;
-        let mut underline_position =
-            f32::from(face.ft_face.underline_position()) * x_scale / 64.;
-        let mut underline_thickness =
-            (f32::from(face.ft_face.underline_thickness()) * x_scale / 64.).max(1.);
+        let mut underline_position = f32::from(face.ft_face.underline_position()) * x_scale / 64.;
+        let mut underline_thickness = f32::from(face.ft_face.underline_thickness()) * x_scale / 64.;
 
         // Fallback for bitmap fonts which do not provide underline metrics
         if underline_position == 0. {
             underline_thickness = (descent / 5.).round();
-            underline_position = descent / 2. + underline_thickness / 2.;
+            underline_position = descent / 2.;
         }
 
         // Get strikeout position and thickness in device pixels
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -97,22 +97,24 @@ impl ::Rasterize for FreeTypeRasterizer {
 
         // Get underline position and thickness in device pixels
         let x_scale = full.size_metrics.x_scale as f32 / 65536.0;
-        let underline_position =
-            (f32::from(face.ft_face.underline_position()) * x_scale / 64.).round();
-        let underline_thickness =
-            (f32::from(face.ft_face.underline_thickness()) * x_scale / 64.)
-            .round()
-            .max(1.);
+        let mut underline_position =
+            f32::from(face.ft_face.underline_position()) * x_scale / 64.;
+        let mut underline_thickness =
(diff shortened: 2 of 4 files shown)


cole-h:
On my Linux machine using Iosevka, the strikeout looks somewhat strange. It's not a big deal, but I thought I would bring it up and see what could be done (if anything).

In the screenshot below, the `# comment` is describing the previous output.

chrisduerr:
@cole-h Thanks for the heads-up. Looking at the metrics again, I can see no errors in the current behavior. The initial version of this PR should actually be a regression.

Since the metrics look right to me, I decided to just round them instead of changing the metrics itself. This produces a significant improvement for my font, so hopefully that's good to go.

chrisduerr:
@FichteFoll We get these metrics directly from the font, so they should look fine. I'm not a huge fan of the idea to add an additional offset. At that point we might as well just draw it wherever we want and ignore what the font has to say. If the font is actually at fault, the font should be fixed. Of course at this point this quite possibly still has some minor bugs.

chrisduerr:
@FichteFoll Would you mind providing a screenshot of the latest update of this PR? Does this look similar to termite/urxvt/kitty?

FichteFoll:
@chrisduerr did you perhaps forget to push?

chrisduerr:
@FichteFoll Thanks for the heads-up. I must have forgotten to push my changes. Should work now though.

FichteFoll:
@chrisduerr looks exactly the same as in my earlier screeshot here.

chrisduerr:
Hmm, so this definitely improves things, but still isn't completely matching URxvt/VTE/Kitty.

I'll look into this a bit more to see if there's any minor rounding/off by one errors I've potentially missed with the metrics calculation. Otherwise just merging this and following up in the future should also be a good option.

DarkDefender:
I'm having trouble with my bitmap fonts (as FichteFoll). Even with the patch here, the underline is glued onto the bottom of the text. In the other terminals I've tried, the rendering is kinda like in urxvt with some space between the text and the underline.

Urxvt (tewi font):
urxvt

Alacritty (with patch):
ala_patch

chrisduerr:
@FichteFoll It would help a lot if you could tell me the metrics which are reported with your font. You can do so by applying the following patch. Of course if others have the same problem, that would help too.

Just trying to figure out where an extra pixel or two might could come from.

```diff
diff --git a/font/src/ft/mod.rs b/font/src/ft/mod.rs
index 03e8bb3..8db1544 100644
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -97,12 +97,16 @@ impl ::Rasterize for FreeTypeRasterizer {
 
         // Get underline position and thickness in device pixels
         let x_scale = full.size_metrics.x_scale as f32 / 65536.0;
-        let underline_position =
-            (f32::from(face.ft_face.underline_position()) * x_scale / 64.).round();
-        let underline_thickness =
-            (f32::from(face.ft_face.underline_thickness()) * x_scale / 64.)
-            .round()
-            .max(1.);
+        let mut underline_position =
+            (f32::from(face.ft_face.underline_position()) * x_scale / 64.);
+        let mut underline_thickness =
+            (f32::from(face.ft_face.underline_thickness()) * x_scale / 64.).max(1.);
+
+        println!("UL POS: {} * {} / 64 = {}", face.ft_face.underline_position(), x_scale, underline_position);
+        println!("UL HEI: {} * {} / 64 = {}", face.ft_face.underline_thickness(), x_scale, underline_thickness);
+
+        underline_position = underline_position.round();
+        underline_thickness = underline_thickness.round();
 
         // Get strikeout position and thickness in device pixels
         let (strikeout_position, strikeout_thickness) =
diff --git a/src/renderer/lines.rs b/src/renderer/lines.rs
index 781f4d4..85d96eb 100644
--- a/src/renderer/lines.rs
+++ b/src/renderer/lines.rs
@@ -132,6 +132,10 @@ fn create_rect(
         y = max_y;
     }
 
+    if flag == Flags::UNDERLINE {
+        println!("UNDERLINE:\n    CELL BOTTOM: {}\n    BASELINE: {}\n    UL POS: {}\n    UL HEIGHT: {}\n    MAX Y: {}\n    Y: {}", cell_bottom, baseline, position, height, max_y, y);
+    }
+
     let rect = Rect::new(
         start_x + size.padding_x,
         y.round() + size.padding_y,

```

chrisduerr:
The thing I notice immediately here is that your output has a `UL POS` which has decimal points. IIRC with my testing that was never the case and these are currently all truncated before they are reported. So maybe these few additional decimal numbers might affect the pixel position by a single pixel.

The other metrics all look perfectly fine to me. But iirc @cole-h, the current metrics should work great for you, right? Removing the rounding would only move the `Y` position to about `56.7`, which would still be the same with rounding applied.

Might be interesting if the `UL POS` affects @FichteFoll differently. That might be the extra pixel which is missing.

FichteFoll:
Looks pretty weird, I guess.

```
~/code/alacritty Δ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/alacritty`
UL POS: 0 * 1 / 64 = 0
UL HEI: 0 * 1 / 64 = 1
UL POS: 0 * 1 / 64 = 0
UL HEI: 0 * 1 / 64 = 1
UL POS: 0 * 1 / 64 = 0
UL HEI: 0 * 1 / 64 = 1
UL POS: 0 * 1 / 64 = 0
UL HEI: 0 * 1 / 64 = 1
```

chrisduerr:
@FichteFoll That looks great! Makes a lot of sense that this wouldn't produce good results, your font isn't providing any reasonable metrics for underline positioning!

I wonder if this is related to bitmap fonts and if there is any bitmap font which properly provides underline metrics. Supporting the rusttype underline metrics when the underline position is `0` would be a possible fix for this.

DarkDefender:
I get the same thing. 

```
UL POS: 0 * 1 / 64 = 0
UL HEI: 0 * 1 / 64 = 1
UNDERLINE:
    CELL BOTTOM: 24
    BASELINE: 21
    UL POS: 0
    UL HEIGHT: 1
    MAX Y: 23
    Y: 20.5
```
As you can see, there is no underline information in the bdf format: https://en.wikipedia.org/wiki/Glyph_Bitmap_Distribution_Format
So I guess that underline pos is not defined for bitmap fonts. I think that bitmap fonts actually provides a character for underline instead. But that won't really work as alacritty is now drawing the lines instead.

chrisduerr:
I think it should be fine to skip the character and render with our own metrics. Hopefully that should resolve the remaining issues with bitmap fonts.

It also seems like a good idea to remove the rounding from the metrics though and only apply it once the rect is created, just to make sure.

chrisduerr:
@FichteFoll @DarkDefender I've added some custom metrics for bitmap fonts which should help with drawing the underline. Can you confirm that the behavior is now improved compared to the latest master? Or are there still any issues?

cole-h:
Even though your latest commit is to help with bitmap fonts, it makes strikethrough look relatively normal on Iosevka:

image

DarkDefender:
That seems to produce acceptable results for me @chrisduerr. It doesn't exactly match urxvt and so on. but it doesn't have to really. The underline position is fine for me now.

FichteFoll:
Much better now.



```
UNDERLINE:
    CELL BOTTOM: 48
    BASELINE: 44
    UL POS: -2.5
    UL HEIGHT: -1
    MAX Y: 49
    Y: 47
```

chrisduerr:
@FichteFoll @DarkDefender I have further improved the line metrics by properly centering the underline in the space that is available. I think that should look even better, let me know if that's not the case please.

In @FichteFoll's case, this should raise the underline's `Y` position to `46`, hopefully making it interfere less with symbols like `y`.

DarkDefender:
@chrisduerr Now it matches exactly urxvt and the other terms I've tried!

FichteFoll:
Seems unchanged.

chrisduerr:
Ah yeah, makes sense that it's unchanged because the 0.5 offset that is added puts it back at the same position with rounding.

I think this should be good to go now. It seems to produce reasonable results with bitmap fonts too.

PR #1922, opened 2018-12-24
Set min osx deployment target to 10.11

### LINKED ISSUE #1853: Doesn't work for Mac OSX Sierra
image

I'd expect it to at-least work for last two versions of a major OS

dm1try:
to get maximum coverage for pre-built binaries

closes #1853

in addition, `MACOSX_DEPLOYMENT_TARGET` can be overridden

BTW
I was able to reproduce the referenced issue by providing not existing version(`MACOSX_DEPLOYMENT_TARGET=10.14.3 make app`):
screen shot 2018-12-25 at 00 40 49

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,6 @@ APP_DIR = $(RELEASE_DIR)/osx
 APP_BINARY = $(RELEASE_DIR)/$(TARGET)
 APP_BINARY_DIR  = $(APP_DIR)/$(APP_NAME)/Contents/MacOS
 
-export MACOSX_DEPLOYMENT_TARGET = $(shell defaults read loginwindow SystemVersionStampAsString)
-
 DMG_NAME = Alacritty.dmg
 DMG_DIR = $(RELEASE_DIR)/osx
 
@@ -24,9 +22,6 @@ help: ## Prints help for targets with comments
 
 binary: | $(TARGET) ## Build release binary with cargo
 $(TARGET):
-ifneq ( "${MACOSX_DEPLOYMENT_TARGET}" , "")
-	@echo MACOSX_DEPLOYMENT_TARGET=$${MACOSX_DEPLOYMENT_TARGET}
-endif
 	cargo build --release
 
 app: | $(APP_NAME) ## Clone Alacritty.app template and mount binary
--- a/ci/before_deploy.sh
+++ b/ci/before_deploy.sh
@@ -22,6 +22,7 @@ mkdir "./target/deploy"
 name="Alacritty-${TRAVIS_TAG}"
 
 if [ "$TRAVIS_OS_NAME" == "osx" ]; then
+    export MACOSX_DEPLOYMENT_TARGET=10.11
     make dmg
     mv "./target/release/osx/Alacritty.dmg" "./target/deploy/${name}.dmg"
 elif [ "$TRAVIS_OS_NAME" == "linux" ] && [ "$ARCH" != "i386" ]; then


chrisduerr:
I wonder if this is the best place to set this. I'd imagine if someone would compile it on their own machine, we'd very much want to set the deployment target to be exactly what is installed on the system.

Maybe it would be a better idea to just set this when building on CI before deploying the binaries? That way the pre-built binaries would still work, but on a local machine you might still benefit from using system-specific versions.

dm1try:
@chrisduerr I think you are right.
though
> I'd imagine if someone would compile it on their own machine, we'd very much want to set the deployment target to be exactly what is installed on the system.

should we set it at all? I mean if we do not set any target version then the host version should be used as default one anyway. so I little bit confused by the comment in https://github.com/jwilm/alacritty/commit/24533d2c5d86857dbb2f75f171430da7073ce65c
> On macOS 10.12.6, running `make app` creates a bundle that refuses to
start (it complains application is only compatible with more recent
versions of macOS). The binary works great when running it directly.

as there are no any target specific restrictions in the build script.

chrisduerr:
> should we set it at all? I mean if we do not set any target version then the host version should be used as default one anyway. so I'm a little bit confused by the comment in 24533d2

I can not tell you *why* it didn't work, but apparently this does help building on some systems. Now I don't know if this is because the user had an improperly configured system with the variable set already, or if it's actually necessary, but I think the safest thing would be to just keep it set.

This raises a problem though. If we want to run `make` with a different value of this variable during CI, we'd need to make it conditional. So we'd only set the variable if it's not already present in the system. If we make the assumption that the only reason why it didn't work was because the environment variable was already set, then this will not help.

Maybe @Carreau is still around and is able to provide some feedback?

dm1try:
> Now I don't know if this is because the user had an improperly configured system with the variable set already

while investigating this I found that system caches the result of minimum required version of a binary and does not invalidate the result on replacing a content of application directory. so I have an additional possible scenario:
1. user with old(`10.12`) osx version downloads the binary which was build on CI for higher version of osx; this version does not work for him (*this is expected*)
2. he decides to follow instructions for manual build and run `make app` and `cp -r target/release/osx/Alacritty.app /Applications/`; this does not invalidate the cache mentioned above and he still has the same problem

this scenario is reproducible
screen shot 2018-12-25 at 15 46 19

removing directory before  copying certainly helps with cache invalidation(not sure about other triggers) and fixes the problem.

if this is @Carreau's case, I'll vote for setting explicit `MACOSX_DEPLOYMENT_TARGET` only on CI.

it would also help if someone with osx `10.12.6` will make a fresh install **without** this line
https://github.com/jwilm/alacritty/blob/f0180430dfea33d806ab530e789d2edb1488af6b/Makefile#L11
and provide the results here.
in the case of failure, please run this command to check minimum os version of the compiled binary
 ```
otool -l target/release/osx/Alacritty.app/Contents/MacOS/alacritty | grep -E "minos|sdk"
       sdk 10.14
     minos 10.14.2
```

**Comment by chrisduerr** [comment]
@dm1try In that case it seems like just setting it to 10.11 for everyone might just be the easiest solution. If that works on every macOS platform, maybe that's just the best choice. I doubt there will be much performance difference, though that's something that can be tested.

**Comment by chrisduerr** [comment]
@dm1try Do you think this PR is good to go the way it is right now? It works on both 10.11 and later versions of macOS, right?

Or do you think there would be any benefit in following a different approach?

**Comment by Carreau** [comment]
I might not be able to fully test before a couple of days (travelling +
promised to be offline, but already breaking promise) but will in a couple
of days. I was aware of the caching of binary compatibility (which made
testing supper annoying), but I'm pretty sure that option was necessary for
it to work on my machine.

I think people on "old" osx version are a minority, so I believe it would
be ok to not default to using the option for local build. If the effect of
building for "old" version have minimal impact, I would be happy for CI
built versions to have maximal compatibility.

**Comment by dm1try** [comment]
@chrisduerr  I want to update the PR 
> for setting explicit MACOSX_DEPLOYMENT_TARGET only on CI.

but let's wait for some feedback from @Carreau or other users who use the old version of osx.

as I mentioned above it's strange that the default behavior on the old OS version is different: the minimum required version is not equal the version of host machine while compiling a binary(I posted results for my environment above, see `otool` check). it might be a compiler/environment related issue(new Xcode on old system/or specific SDK settings). But even so, this should affect not only alacritty build but any other program build also, so passing `MACOSX_DEPLOYMENT_TARGET` by the user explicitly might be the preferred way to do it(this can be mentioned in wiki). 

> Or do you think there would be any benefit in following a different approach?

simplicity: less code to support, relying on the default behavior :)

**Comment by Carreau** [comment]
So as far as I can tell, rustc [default to 10.7](https://github.com/rust-lang/rust/blob/f381a962550436f74dd6e9021e4df2fdefb96cfa/src/librustc_target/spec/apple_base.rs#L17-L23) if MACOSX_DEPLOYMENT_TARGET is not set (at least for the linker). This is (possibly) unrelated to xcode ? But I'm out of my depth here. 

This is a strange as it is incoherent with my test (slightly modified from above). 

```
otool -l target/release/osx/Alacritty.app/Contents/MacOS/alacritty | grep -E "version|minos|sdk" | head -n2
```

gives respectively,


MACOS_DEPLOYMENT_TARGET unset. 
```
version 10.12.7 # (I have 10.12.6)
      sdk 10.12
```
`.app` does not works


MACOS_DEPLOYMENT_TARGET=10.7
```
  version 10.7
      sdk 10.12
```
`.app` works

MACOS_DEPLOYMENT_TARGET=10.11
```
  version 10.11
      sdk 10.12
```
`.app` works


Xcode is Version 9.2 (9C40b), I don't believe I can use more recent. I don't usually develop with XCode or have specific setting – I'm not even sure how to change the default deployment target if you don't have a xcode project. 

I'm happy to try other things, but I'm unsure where to go from there.

**Comment by chrisduerr** [comment]
Looks to me like the best way forward is to keep the makefile the way it is right now and just override it to `10.11` during the release deployment process using the environment variable.

Thanks for getting back to this and providing some feedback regarding the original issue @Carreau.

**Comment by Carreau** [comment]
That seem reasonable. If there is any other things I can help test on 10.12 let me know. 

Also 10.12.7 is not a version of MacOS that exists, so I'm not sure what might be setting this.

Thanks to you for all the work you are doing.

**Comment by dm1try** [comment]
I've noticed that in the current build process `MACOS_DEPLOYMENT_TARGET` is ignored at all; the cached binary is reused on packing an application, for example see [this build](https://travis-ci.org/jwilm/alacritty/jobs/465755130)
```
MACOSX_DEPLOYMENT_TARGET=10.13.3
cargo build --release
    Finished release [optimized + debuginfo] target(s) in 0.47s
Created 'Alacritty.app' in 'target/release/osx'
Packing disk image...
created: /Users/travis/build/jwilm/alacritty/target/release/osx/Alacritty.dmg
Packed 'Alacritty.dmg' in 'target/release/osx'
```
and the result
```
Load command 9
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.13
      sdk 10.13
```
instead of
```
Load command 9
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.13.3
      sdk 10.13
```

`MACOSX_DEPLOYMENT_TARGET` should be set before the first `cargo --release` run to get the expected result

**Comment by chrisduerr** [comment]
> anyway, I disagree with the decision

Why? It seems like compiling for the platform the user is on is the best choice in most cases. Caching should be irrelevant outside of CI builds.

Caching on CI is also not a big deal. We'd just have to set this once and clear the cache and everything should be good to go.

I've never been able to test or reproduce this issue since I don't have a macOS machine. So I'm very curious why you've decided to close this issue @dm1try since it seems like there should be a way forward which satisfies everyone. You said that just removing the variable completely is the best way forward you can recommend, but it seems like that would break both on CI and on setups like Carreau's?

**Comment by dm1try** [comment]
> Why? It seems like compiling for the platform the user is on is the best choice in most cases.

As I already said, this is how it works by default without providing `MACOSX_DEPLOYMENT_TARGET`. Particularly, the minimum required version is set to the version of SDK which is used on a compilation. The good example of default behavior is the current release binary which has the minimum version is equal `10.13`(see above, `MACOSX_DEPLOYMENT_TARGET` is ignored at all because it is set after the binary was compiled)

> Caching on CI is also not a big deal. We'd just have to set this once and clear the cache and everything should be good to go.

just for clarity, the cache discussed above is about caching the compiled binary between compilations
```
cargo test --release
export MACOSX_DEPLOYMENT_TARGET=10.13.3 
make dmg #  MACOSX_DEPLOYMENT_TARGET is ignored as the binary was already compiled in the first step, so the minimum version is equal the version of used SDK
```

> I've never been able to test or reproduce this issue since I don't have a macOS machine. So I'm very curious why you've decided to close this issue

I commented above about why I closed this:
> I've noticed that in the current build process MACOS_DEPLOYMENT_TARGET is ignored at all; the cached binary is reused on packing an application

so [the provided PR change](https://github.com/jwilm/alacritty/pull/1922/commits/d0703b44048744500337d84eb3a8bf41c4fdbf63) does not solve the original problem(#1853). `MACOS_DEPLOYMENT_TARGET` should be set in `.travis.yml` or `ci/script.sh` to make an impact on a compiled binary.

> since it seems like there should be a way forward which satisfies everyone.

please clarify that you mean by the satisfaction. I'm interested in the proper solving of the problem.
lets look at this information provided by @Carreau 
> version 10.12.7 # (I have 10.12.6)
>
> Xcode is Version 9.2 (9C40b)
> sdk 10.12
1. the minimum version in the test binary looks wrong.
2. the default sdk for this version of Xcode is `10.13.2`

so adding the line with `export MACOSX_DEPLOYMENT_TARGET = $(shell defaults read loginwindow SystemVersionStampAsString)` to `Makefile` looks like a dirty workaround for fixing a system setup for this particular case, but the consequences are:
- changing a default behavior and don't use `MACOSX_DEPLOYMENT_TARGET` for [its intended purpose](https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/cross_development/Configuring/configuring.html)
- the discussion in this issue that partially unrelated to the original problem(set a minimum required version for the pre-built binaries)

 > but it seems like that would break both on CI and on setups like Carreau's?

1. see above, CI setup already ignores `MACOSX_DEPLOYMENT_TARGET` and has the default values:
```
    # the information for a downloaded release binary
      cmd LC_VERSION_MIN_MACOSX
  version 10.13
      sdk 10.13
```
so it should not break anything

2. if the problem in Carreau's setup is common for users on `10.12.6` then this information should be added to the wiki. And again, it looks like a system-wide problem so the user should solve it on the system setup level but not in the makefile of every compiled binary.

chrisduerr:
The `MACOSX_DEPLOYMENT_TARGET` variable is used to specify the minimum version of macOS which should be supported by the application. So as far as I can tell, always setting it to `10.11` should resolve the problem for most users.

You're talking about caching of the binary, however this is not an issue on CI and neither is it a problem with local builds. The assumption for local builds is always that they're made from a clean repo clone, we don't need to worry about previous build caches unless they're going to create issues in the future. On CI we have full controll over the cache, cleaning that once is not an issue.

> The good example of default behavior is the current release binary which has the minimum version is equal 10.13 (see above, MACOSX_DEPLOYMENT_TARGET is ignored at all because it is set after the binary was compiled).

The current release binary has a version of `10.13` because our CI server (the default travis setup) runs macOS `10.13` by default. It's not an issue with caching, this is the expected behavior.

dm1try:
> You're talking about caching of the binary, however this is not an issue on CI and neither is it a problem with local builds.

Re-read above.

> The current release binary has a version of 10.13 because our CI server (the default travis setup) runs macOS 10.13 by default. It's not an issue with caching, this is the expected behavior.

That's it. This is the default behavior without any dirty workarounds placed in Makefile.

chrisduerr:
> That's it! This is the default behavior without any dirty workarounds that needed to be placed in Makefile.

It doesn't work though, since we don't want to produce binaries for `10.13`.

Which gives us the option to either always specify `10.11`, or specify `10.11` just for CI. As far as I know we don't need any newer features, so it should be irrelevant which version is specified exactly. Setting it to `10.11` in the makefile, will solve both @Carreau issue and the problem with binaries not being available on CI. Setting it just on CI deployment will only fix one of those problems.

I'll create a separate PR to set this up, but it seems like that's the best way forward since it makes everyone happy and means we don't need to deal with potential system configuration issues.

dm1try:
> it doesn't work though, since we don't want to produce binaries for 10.13.

this was discussed in the context of https://github.com/jwilm/alacritty/commit/24533d2c5d86857dbb2f75f171430da7073ce65c, as the example of how it should work by default and it was indirectly related to the original #1853.

PR #1931, opened 2018-12-27
Clearify create_rect Comment

nixpulvis:
Both me and @chrisduerr seem to agree this comment was a little confusing. Hopefully this is better.

### DIFF
--- a/src/renderer/lines.rs
+++ b/src/renderer/lines.rs
@@ -105,7 +105,8 @@ impl<'a> Lines<'a> {
     }
 }
 
-/// Create a rectangle between two cells.
+/// Create a rectangle that starts on the left of `start` and ends on the right
+/// of `end`, based on the given flag and size metrics.
 fn create_rect(
     start: &RenderableCell,
     end: &RenderableCell,


src/renderer/lines.rs
```diff
@@ -105,7 +105,7 @@ impl<'a> Lines<'a> {
     }
 }
 
-/// Create a rectangle between two cells.
+/// Create a rectangle spanning two cells.
```
chrisduerr: I don't think `spanning` two cells is the correct word here either. If the start cell is at index `0` and the end cell is at index `5`, the underline would go from `0` to the end of `5`. So it's not just from `0` to `1` or two other neighboring cells.

Maybe something like this is more understandable:

```rust
/// Create a vertical rectangle from the start of one, to the end of another cell.
```

I'm open to put even more information into this comment though. So there's no need to limit it to a short one-liner.

nixpulvis: updated

src/renderer/lines.rs
```diff
@@ -105,7 +105,8 @@ impl<'a> Lines<'a> {
     }
 }
 
-/// Create a rectangle between two cells.
+/// Create a rectangle that starts on the left of `start` and ends on the right
+/// of `end` based on the given flags and size metrics.
```
chrisduerr: That sounds pretty clear to me! Just some minor nitpicks, then let's ship it.

`flag` should be singular since it's never all cell flags that are passed, but just the underline OR the strikeout flag. So even if a cell has both strikeout and underline, these are handled separately since they require to distinct rectangles.

```suggestion
 /// of `end`, based on the given flag and size metrics.
```

PR #1959, opened 2018-12-31
Change default Windows shell to PowerShell

### LINKED ISSUE #1928: Starts umaintained 'cmd' shell on Windows
Alacritty 0.24 on Windows 10.0.17763.0 will start `cmd` by default. The default shell (see Win+X) on Windows is powershell and cmd is no longer maintained by Microsoft. 

Alacritty should start powershell by default, which is actively maintained by Microsoft and supports modern features like key bindings, readline, arrays, objects, and functions, and separates content from presentation avoiding the need for regex scraping. 

Ideally, Alacritty should do what Terminus does and start `pwsh` (Powershell 6) if installed, then fall back to `powershell`

lzybkr:
Fix #1928

### DIFF
--- a/alacritty_windows.yml
+++ b/alacritty_windows.yml
@@ -276,12 +276,12 @@ live_config_reload: true
 
 # Shell
 #
-# You can set `shell.program` to the path of your favorite shell, e.g. `powershell`.
+# You can set `shell.program` to the path of your favorite shell, e.g. `cmd`.
 # Entries in `shell.args` are passed unmodified as arguments to the shell.
 #shell:
-#  program: cmd
+#  program: powershell
 #  args:
-#    - /K env.cmd
+#    -
 
 # Windows 10 ConPTY backend
 #
--- a/alacritty_windows.yml
+++ b/alacritty_windows.yml
@@ -276,12 +276,12 @@ live_config_reload: true
 
 # Shell
 #
-# You can set `shell.program` to the path of your favorite shell, e.g. `/bin/fish`.
+# You can set `shell.program` to the path of your favorite shell, e.g. `powershell`.
 # Entries in `shell.args` are passed unmodified as arguments to the shell.
-shell:
-  program: powershell
-  #args:
-  #  - --login
+#shell:
+#  program: cmd
+#  args:
+#    - /K env.cmd
 
 # Windows 10 ConPTY backend
 #
--- a/alacritty_windows.yml
+++ b/alacritty_windows.yml
@@ -280,8 +280,7 @@ live_config_reload: true
 # Entries in `shell.args` are passed unmodified as arguments to the shell.
 #shell:
 #  program: powershell
-#  args:
-#    -
+#  args: []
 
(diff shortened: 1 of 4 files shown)


zacps:
Please add a changelog entry.

lzybkr:
I changed the fallback from `cmd` to `powershell`, but didn't comment out the shell in the configuration file because COMSPEC is always set and no reasonable person would set it to a shell that doesn't work like `cmd`.

How do people feel about ignoring COMSPEC? If that's fine, then we can comment out the shell entry in the configuration file.

chrisduerr:
@lzybkr What is `COMSPEC` set to by default? Also `cmd` should work just fine, right? It just is less functional than powershell.

lzybkr:
I do think `SHELL` and `COMSPEC` are different enough that we should ignore `COMSPEC`.

I believe the most important use of `COMSPEC` is in the C `system` api - see the [Windows](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/system-wsystem?view=vs-2017) and [Linux](http://man7.org/linux/man-pages/man3/system.3.html) documentation. Notably - Linux always uses `/bin/sh`, not the value of any environment variable like Windows.

Windows does not have the notion of a logon shell.

zacps:
@lzybkr Is correct, COMSPEC should really never be changed without a very good reason.

I think we should just ignore COMSPEC entirely, because it's going to be cmd 99% of the time.

iKlsR:
Is/will there be a way to use `(ba)sh` on windows akin to what cygwin or git-bash does?

chrisduerr:
@iKlsR It's already possible. I've been running Alacritty with bash on Windows 7 for a while.

iKlsR:
@chrisduerr Ok, we can clean the conversation after. *How?* I'm on windows 10 so bash goes to WSL. Setting program to git-bash, sh or any other variant causes the app to hang, I don't like powershell and cmd is next to useless,

alacritty_windows.yml
```diff
@@ -276,12 +276,12 @@ live_config_reload: true
...
+# You can set `shell.program` to the path of your favorite shell, e.g. `cmd`.
 # Entries in `shell.args` are passed unmodified as arguments to the shell.
-shell:
-  program: cmd
-  #args:
-  #  - --login
+#shell:
+#  program: powershell
+#  args:
+#    -
```
chrisduerr: Uncommenting this `shell` block should work properly without any issues. However the `args:` section could cause issues here.

I'd recommend just changing `args` to an empty array like this: `args: []`.

lzybkr: It might be a tiny bit harder for folks to actually specify args if they don't know yaml, but I made the change.

nixpulvis: Or supply an example use case. Just a thought.

lzybkr: I had supplied an example use case in a previous commit because I also thought that was a reasonable idea, but specifying the default is reasonable as well.

PR #1968, opened 2019-01-03
Fix double-triggering of mouse bindings

chrisduerr:
The 2d9afb9b395ea0cc71432613e0df104aefcf24c1 commit lead to mouse
actions being triggered on both press and release of mouse buttons.

This reverts the mouse binding behavior back to the previous state where
they are only triggered when the button is pressed, not when it's
released.


matthiaskrgr:
Looks good to me, thanks! (tested on x11 linux)

Issue #607, opened 2017-06-11
Support multiple windows


net:
I use a window manager instead of tmux. With other terminal emulators when I want a new terminal instance, I simply open a new window. As far as I can tell, the only way to accomplish this with Alacritty is to open a new instance of the application. This is inferior for two reasons:

1. Load times when opening a new instance of Alacritty are subpar. For a program that prides itself on being the fastest, opening new terminals should be no exception. (Loading a new instance of Alacritty lasts about 1 second for me vs instant when opening a new window in iTerm.)

2. On macOS opening multiple instances of Alacritty fills up the Dock and Application Switcher.

From what I can tell from other issues, using window managers as a replacement for tmux is within Alacritty's scope. Even when using tmux, it can be desirable to open a second tmux instance for an unrelated session without detaching from the current session. I think multiple windows should be supported natively by Alacritty.

maximbaz:
Related: #241

robertgzr:
I [add this](https://gist.github.com/robertgzr/34e1a90b885037e2ccef870740f4763d) to get a 'new window' command.

It's not very nice but it works when I need two tmux sessions in seperate windows...

blark:
It would be nice to get "new window" support. Nesting tmux is terrible, and I use it remotely via SSH a lot. Maybe someone can help me out here, is there a good way to do this?

tbodt:
My opinion on this is alacritty should become a sort of terminal emulator UI component, so other programs could use alacritty as a renderer and provide tabs, split panes, multiple windows, GUI configuration, and all that cool stuff.

mkeeler:
The main reason I would like multi-window is I ran into some scenarios where I wanted to monitor a whole lot of panes simultaneously and wanted a window on one monitor and one on another. Ideally after starting the Alacritty.app there would be some hotkey to spawn a new window.

brycefisher:
Quoting from #674:

> While I can create some alias to launch a new alacritty window easily, we still cannot:
> 
>    * use spotlight to bring a running version of alacritty to the front - which I guess should be the default behaviour.
>    * open a new alacritty window without dropping to a shell first - maybe the macOS app should have a "new window" menu item, seems like it might be a common usage.
> 
> These two things are surmountable, but would greatly improve the UX on macOS (I have no idea the difficulty of implementation however)

I feel the same way!

My current workaround for having no way to launch a second alacritty window from SpotLight is to launch terminal.app for times I would use a secondary alacritty window.

[...]

Issue #241, opened 2017-01-08
daemon mode


untoreh:
a daemon mode would be nice, for reference urxvtd uses 20mb + 100kb ram per window

aktau:
It'd need to be pretty resistant to crashing though. I stopped using urxvt daemon mode because it crashed (only a few times, not many) and took all of its windows with it. Is current alacritty memory usage that high?

untoreh:
I have never experienced a crash on my end with urxvtd, anyway alacritty seems to use roughly 30mb per window? libvte terminals with dbus use <1mb per window

mathstuf:
I've seen urxvtd crash a few times (as in maybe 5 in the past 7 years), but `tmux` buffers me against those anyways.

I think the idea, at least with urxvtd, is more about speed of opening new windows rather than memory usage.

beta-phenylethylamine:
It's mostly exactly what @mathstuf said. I tend to open and close terminal emulators a lot and have a lot open at the same time at the same time, so client/daemon terminal emulators are a no brainer for me usually. It and scrollback not having arrived in stable yet are the two big reasons I haven't switched to Alacritty even though it runs better and has much cleaner font rendering than xiate (my TE of choice right now) does

mpe:
I tried switching to alacritty this week and found the memory usage is a show-stopper for me. Admittedly I do tend to run a lot of terminals.

As a small experiment running 10 alacritty's uses ~135MB each, or ~1.3GB in total. That doesn't change noticeably between just after startup vs with 10,000 lines of scrollback each.
 
Running 10 gnome-terminal windows uses ~50MB in total at startup.
After 10,000 lines of scrollback on each it's up to ~100MB in total.

It's true that daemon mode increases your risk if there's a crash, but honestly I can't remember the last time gnome-terminal crashed on me.

That all sounds a bit negative but is meant to be constructive! What's working so far is awesome :tada:

maximbaz:
While I'm curious myself about the daemon mode, I cannot reproduce your results on memory consumption.

If I interpret the output of `htop` correctly, the memory usage is the physical memory used by process (`RES`) minus the memory that is shared by processes (`SHR`). 

> Could someone confirm if this is correct way to measure real memory usage?

Here's my `htop`:

![image](https://user-images.githubusercontent.com/1177900/38363345-9ca41062-38d4-11e8-99e4-cac836d69603.png)

[...]

PR #1971, opened 2019-01-04
Add action to create new instance of alacritty within the current working directory

Camotubi:
Other terminals allow you to spawn a new instance of the terminal with the same working directory. This pull request adds this feature.

Tested on: 
- Linux Xorg
- Linux Wayland ( Sway )

Notes:
- It will probably crash on windows and mac.
- I changed the shell PID variable on tty::unix::PID to public. Don't know if that is allowed. 
- I didn't write any tests 😅. 

This fixes https://github.com/jwilm/alacritty/issues/808.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Crashes on Windows are now also reported with a popup in addition to stderr
 - Windows: New configuration field `enable_experimental_conpty_backend` which enables support
     for the Pseudoconsole API (ConPTY) added in Windows 10 October 2018 (1809) update
+- New key action `SpawnNewInstance` for launching another instance of Alacritty
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Crashes on Windows are now also reported with a popup in addition to stderr
 - Windows: New configuration field `enable_experimental_conpty_backend` which enables support
     for the Pseudoconsole API (ConPTY) added in Windows 10 October 2018 (1809) update
-- New key action `SpawnNewInstance` for launching another instance of Alacritty
+- New mouse and key action `SpawnNewInstance` for launching another instance of Alacritty
 
 ### Changed
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -4,7 +4,6 @@ use std::fs::File;
 use std::io::Write;
 use std::sync::mpsc;
 use std::time::{Instant};
-use std::process;
 use std::process::Command;
 use std::env;
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -4,7 +4,6 @@ use std::fs::{self,File};
 use std::io::Write;
 use std::sync::mpsc;
 use std::time::{Instant};
-use std::process::Command;
 use std::env;
 
 use crate::tty;
@@ -25,7 +24,7 @@ use crate::selection::Selection;
 use crate::sync::FairMutex;
 use crate::term::{Term, SizeInfo, TermMode, Search};
 use crate::term::cell::Cell;
-use crate::util::limit;
+use crate::util::{limit, start_daemon};
 use crate::util::fmt::Red;
 use crate::window::Window;
 use glutin::dpi::PhysicalSize;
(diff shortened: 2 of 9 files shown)


src/event.rs
```diff
@@ -177,6 +181,39 @@ impl<'a, N: Notify + 'a> input::ActionContext for ActionContext<'a, N> {
     fn clear_log(&mut self) {
         self.terminal.clear_log();
     }
+
+    fn new_instance_same_dir(&mut self) {
+        #[cfg(target_os = "linux")]
+        unsafe {
+            let current_working_dir = Command::new("pwdx")
```
quininer: I think you don't need `pwdx`. try [cwd](https://github.com/danburkert/procinfo-rs/blob/master/src/pid/cwd.rs#L12).

Camotubi: It works. :fire:

Camotubi:
I have no idea, why is not compiling on linux. On my machine is compiling ._.

nixpulvis:
It's working for me locally too! :tada:

I would be curious what performance optimizations are avalible here, without getting into the full fledged shared backend. How much of this could be done in a `fork` like fashion where it shares some of the memory etc. Hopefully this makes sense.

chrisduerr:
@nixpulvis You might be talking about #241? URxvt supports that with `urxvtd`. It requires #607 so that a new `alacritty` command would just tell the Alacritty deamon to spawn an additional window.

Unless you're talking about something different here, but I don't think so?

nixpulvis:
#241 was what I meant by "full fledged shared backed", I'm just wondering if there's ways to easily shared things like say the glyph cache with a child process. 

https://linux.die.net/man/3/shm_open

I don't have any experience with this kind of stuff in Rust and most of my searches (on mobile atm) are giving me IPC stuff for bi-directional communication. I'm more interested in stuff that is basically read only.

Maybe this is all not worth it, as we probably will have a different and better solution with #241.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Crashes on Windows are now also reported with a popup in addition to stderr
 - Windows: New configuration field `enable_experimental_conpty_backend` which enables support
     for the Pseudoconsole API (ConPTY) added in Windows 10 October 2018 (1809) update
+- New key action `SpawnNewInstance` for launching another instance of Alacritty
```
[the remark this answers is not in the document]
chrisduerr: I've initially called this just `action`, but attached the `key` to make it more obvious what it's talking about. But it's absolutely correct that this also applies to mouse bindings. I feel like it might be confusing to call this `binding action` since not everyone knows what that would mean, but `New key and mouse action ...` should do the job.

chrisduerr:
Oh an important note I wanted to make is that this will not work with subshells spawned by the shell which is launched by Alacritty. Alacritty will only respect the direct child shell. But I don't think there's much Alacritty can do about that (might be possible, but seems unreasonable to include).

Camotubi:
Before using the PID static variable I was using pgrep to find the PID of the shell with the PID of alacritty. It could be implemented that way. But I think is fine how it is right now.

zacps:
Windows doesn't have /proc, there should be a cross platform way to do this.

chrisduerr:
@zacps Are you aware of how this works on Windows? Because I'm not and I'm not sure if it's worth holding this back for potentially months when there's nobody who's interested in implementing it. The current behavior should work great on Windows, it just doesn't keep the working directory.

chrisduerr:
@zacps Yeah, that's the same results I've found when looking into implementing this on Windows. I think every platform would benefit from this feature even without the working directory, so I think just keeping it as a tiny bonus on Unix for now while tracking feature parity for Windows should be okay.

dm1try:
it's linux-only feature. on osx a process will be crashed with panic if anyone tries to use this new binding.

at least CHANGELOG should mention that the feature is linux-only.

nixpulvis:
I should have been more skeptical with my review maybe, is there a reason we didn't go with [current_dir](https://doc.rust-lang.org/std/env/fn.current_dir.html) and instead wrote our own? The `unsafe` alone should be a red flag.

nixpulvis:
This blog post might help the situation on macOS, https://astojanov.wordpress.com/2011/11/16/mac-os-x-resolve-absolute-path-using-process-pid/.

Though I'm starting to think it would be easier to have the PTY just tell us the output of [`current_dir`](https://doc.rust-lang.org/stable/std/env/fn.current_dir.html) directly.

dm1try:
> This blog post might help the situation on macOS, https://astojanov.wordpress.com/2011/11/16/mac-os-x-resolve-absolute-path-using-process-pid/. See this SO post for more.

Thanks, I already use something similar, see [this example](https://github.com/jwilm/alacritty/issues/808#issuecomment-451645481).

The point in the comment above was only about the delivering a new feature to users, nothing more.

chrisduerr:
> it's linux-only feature. on osx a process will be crashed with panic if anyone tries to use this new binding.

I was not aware of this. Is anyone aware of how this works on BSD?

A separate issue should probably be opened, since this is not the right place to discuss this.

> Though I'm starting to think it would be easier to have the PTY just tell us the output of current_dir directly.

The subprocess in a shell, so it can't just run any Rust code. Running this function anywhere inside the Alacritty process, will just return the directory Alacritty was launched from.

I've created a new issue for discussing this here: https://github.com/jwilm/alacritty/issues/1988

Issue #1981, opened 2019-01-07
Best way to get tabs on Windows?


mikemaccana:
Similar to https://github.com/jwilm/alacritty/issues/366 - what's the best way to get tabs on Windows? On nix it's tmux (even then a lot of people prefer using graphical tabs rather than text ones), but on Windows tmux isn't an option. I'm currently using Terminus, which has tabs out of the box. What's the best way to open a new tab on Alacritty/Windows?

chrisduerr:
There is no intention for Alacritty to implement its own tabs, however with the Windows 'Sets' feature which is currently in available for testing in Windows 10, it should be able to have multiple tabs of Alacritty in the same window.

mikemaccana:
Sets got cancelled, and removed from current Windows builds.

mikemaccana:
https://www.theregister.co.uk/2018/06/28/sets_scrapped_from_windows_10/

mikemaccana:
Maybe Groupy might be a good alternative...I;m going to investigate...

chrisduerr:
Oh, I was not aware that sets got canceled. There might be other alternatives to provide the same feature, but even without sets it's not planned to manually implement this in Alacritty just for Windows.

mikemaccana:
I still think adding tabs would be useful for all OSs, but I do understand your view that this is a job better suited to the window manager or an app running inside the terminal. 

Looks like Groupy does this:

![image](https://user-images.githubusercontent.com/172594/50767350-5a2e4980-1274-11e9-90aa-17dad8ed5919.png)

liufuyang:
Thank you very much, I was wondering the same. Perhaps the doc can suggest a good windows manager for Windows. Alacrity seems working out of the box in Mac with "tabs" but not in Windows.

chrnz008:
its seems to be groupy is the only alternative. there is no windowmanager or multiplexer for windows that does this. are there any other alternatives

PR #1983, opened 2019-01-07
Add tabs on Windows to FAQ per #1981

mikemaccana:


chrisduerr:
I have personally never used Groupy and I do not think it's a good idea to advertise third party applications in this fashion. It's up to the user to decide which application he wants to use to make up for the lacking Windows support for tabs and I don't think there's any need for guidance here.

mikemaccana:
It's not advertising - there's 0 other apps that can do this (I've researched the topic fairly extensively).

mikemaccana:
@chrisduerr Would you accept the PR if either:

 - A We explicitly mention Groupy is not endorsed by thee Alacritty developers
 - B We didn't mention Groupy by name but said 'third party applications that add tabs to Windows' 

?

mikemaccana:
Looks like you can't update a PR once it's been closed, so this became #1984

chrisduerr:
> Looks like you can't update a PR once it's been closed, so this became #1984 

I am able to re-open closed PRs. So please don't keep re-opening closed PRs which have been closed for a good reason.

I do not see why Alacritty would explicitly document the lack of tabs *again* in the faq, after it is already documented that some advanced features like tabs and splits are not supported.

There is nothing that separates tabs from other advanced features and it makes little sense to document every unsupported feature separately in the FAQ section. As the name might indicate, the FAQ section is about frequently asked questions which are unanswered in the rest of the readme. This is not the case for tabs.

PR #1984, opened 2019-01-07
Add tabs on Windows to FAQ per #1981

mikemaccana:
Don't mention specific apps.


mikemaccana:
Damn @chrisduerr I sent this specifically to respond to your feedback - what did I do wrong?

chrisduerr:
@mikemaccana See comment in the other PR you've opened. It's already documented that Alacritty does not support tabs and has no plan to do so in the future. As an alternative window managers and terminal multiplexers can be used. There is no need to document this again.

Issue #1988, opened 2019-01-08
Crash on macOS with SpawnNewInstance action


chrisduerr:
This has been introduced by https://github.com/jwilm/alacritty/pull/1971.

It has already been accepted in #1979 that implementing this feature on Windows takes significantly more effort than on Linux, so it's not fully supported initially.

For macOS this feature has been enabled, but it currently leads to a crash whenever it is used. It should be investigated if this can be easily fixed to achieve feature parity with Linux, otherwise it should be disabled on macOS, similar to the current Windows implementation.

nixpulvis:
Reposting my comment from the PR threads here:

This blog post might help the situation on macOS, https://astojanov.wordpress.com/2011/11/16/mac-os-x-resolve-absolute-path-using-process-pid/. See [this](https://stackoverflow.com/questions/8327139/working-directory-of-running-process-on-mac-os) SO post for more.

Though I'm starting to think it would be easier to have the PTY just tell us the output of [`current_dir`](https://doc.rust-lang.org/stable/std/env/fn.current_dir.html) directly.

Camotubi:
There is a wrapper for the libproc 
https://github.com/andrewdavidmackenzie/libproc-rs
There is a function called pidpath that may solve the issue.

@nixpulvis current_dir wont work. That will return the path where alacritty was spawned. We need the shell cwd.

chrisduerr:
I've also just looked into `libproc-rs` and it seems to be that [`pidpath`](https://docs.rs/libproc/0.3.2/libproc/libproc/proc_pid/fn.pidpath.html) should indeed make this quite simple.

Considering that, it's probably worth just fixing this issue so it will work properly on macOS rather than disabling it temporarily.

Note that this would not solve the problem on BSD though, which is likely to have similar problems.

chrisduerr:
I've talked to some BSD users and apparently this will likely crash on BSD too.

However the `libproc-rs` library is macOS specific. It seems like the `procstat` utility is able to get working directory information about processes owned by the current user, so there likely is a way to solve this on BSD too.

Camotubi:
I have read [here](https://stackoverflow.com/questions/27145686/how-to-get-the-current-working-directory-of-a-process-in-freebsd-in-c) that in BSD you could use lsof to get the cwd.
With this I could resolve the shell cwd :
```bash
lsof -p $$ | sed '2q;d'| awk '{print $NF}'
# Note: $$ is the PID of the shell.
```

[...]

Issue #1979, opened 2019-01-06
Spawn new Alacritty instance in working directory on macOS using proc_pidinfo


chrisduerr:
The `SpawnNewInstance` action allows spawning the new Alacritty instance inside the current working directory of the child shell, however this is currently not supported on Windows since `/proc/*/cwd` is not available on non-unix systems.

For feature parity between platforms, this should also be supported on Windows.

The action was originally implemented in https://github.com/jwilm/alacritty/pull/1971.

nixpulvis:
Possible hacking solution, or inspiration: https://github.com/Eugeny/terminus/blob/3aaf490f5790564d851c29a76c62ef5c5450a2d7/terminus-terminal/src/services/sessions.service.ts#L242

davidhewitt:
I've been looking into this a little tonight for Windows. It seems really hard to do it reliably - most robust suggestions I can find involve using low level windows APIs: `CreateRemoteThread` to inject a DLL into the child cmd.exe process and from that DLL execute `GetCurrentDirectory` and send the result back to Alacritty. Seems awfully complicated. There's probably a better solution than that somewhere.

cfiet:
@davidhewitt, @chrisduerr It seems that `CreateRemoteThread` trick is a lot more complex than just running `GetCurrentDirectory`, since the handler that starts the thread has to be in target process address space. For `cmd.exe`/`powershell.exe` that would be limited to WinAPI functions. So the actual code solution would be:
- Have a DLL on Windows that can be loaded into `cmd.exe` process that, when [initialising](https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-entry-point-function) runs `GetCurrentDirectory` and sends a message back to Alacritty process
- Allacritty, after `cmd.exe` has started, runs `CreateRemoteThread` with a handler that would run [`LoadLibrary`](https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibrarya) with that DLL path

Overall the entire process sounds very iffy and error prone and can potentially trigger some AV scanners.

[...]

PR #1990, opened 2019-01-08
[WIP] Fix crash on macOS and BSD for spawn_new_instance action

Camotubi:
Currently spawn new instance action is crashing on macOS and BSD systems. This pull request tries to fix the crash and implement a way to get the current working directory of the child shell for this 2 systems. See #1988

This was implemented blindly. I'm not sure if it works since I have no macOs or BSD machine to test it.  ```libproc-rs ``` was used to get the current working directory for macOS.
For BSD, it parses the output from ```procstat -f $PID ```command. 

I would like macOS and BSD to test this changes.

This fixes https://github.com/jwilm/alacritty/issues/1988.


src/event.rs
```diff
@@ -201,6 +222,24 @@ impl<'a, N: Notify + 'a> input::ActionContext for ActionContext<'a, N> {
...
+    let cwd = std::str::from_utf8(&cwd)
+        .expect("Failed to parse current working directory.")
+        .split("\n");
+    let cwd = cwd.filter(|line| line.find("cwd").is_some())
+        .next();
+    match cwd {
+        Some(c) => Ok(c.to_string()),
+        None => Err("Failed to get BSD shell current working_directory.".to_string())
+    }
+}
```
chrisduerr: We shouldn't spawn a subprocess to a CLI application and attempt to parse its output. Instead the solution should be to call whatever the application itself is calling to acquire the required information.

Calling C code from Rust should be pretty straight-forward, so if there's a library on BSD which allows to retrieve this information, the Rust bindings to that C library can be generated automatically.

Camotubi: What do you mean by log entries? if you are referring to all the previous commits that are appearing here.  I don't know why they are here. I'm sorry but this activity of contributing to projects is new to me. If you could tell me how avoid this from happening I will be grateful.

okuuva:
This seems to prevent the crash on macOS, but isn't quite complete fix. When calling SpawnNewInstance action, the window flashes (refreshes?) but new one doesn't spawn.

Tested on macOS Mojave 10.14.2.

chrisduerr:
The working directory passed to the newly spawned instance might be empty (`""`), this would lead to the newly spawned Alacritty crashing instantly.

It should help to put a few debug outputs in the code to troubleshoot this, but that would be my first guess.

nixpulvis:
Would it make sense to just fallback on the default working directory when we can't use the provided one? I don't like crashing.

chrisduerr:
@nixpulvis That's what this PR should be doing. Unless you're talking about something else.

src/event.rs
```diff
@@ -185,14 +188,23 @@ impl<'a, N: Notify + 'a> input::ActionContext for ActionContext<'a, N> {
...
     fn spawn_new_instance(&mut self) {
         let alacritty = env::args().next().unwrap();
-        #[cfg(unix)]
+        #[cfg(not(any(target_os = "macos", target_os = "linux")))]
+        let args: [&str; 0] = [];
+
+        #[cfg(target_os = "macos")]
+        let args = [
+            "--working-directory".into(),
+            proc_pid::pidpath(unsafe { tty::PID })
```
dm1try: `pidpath` is about returning the absolute path to the process(not the current working directory)

dm1try: according to the information in [the answer on SO](https://stackoverflow.com/a/33852283) mentioned above, `proc_pidinfo` should be used to get the current working directory of a process.

Camotubi: Alright, I will disable the feature on macOS for now. If someone wants to implement the call to proc_pidinfo, please do it. I still don't know how to call C call from Rust and I have no way to test it here.

dm1try: I've looked at [libproc-rs](https://github.com/andrewdavidmackenzie/libproc-rs/blob/9a72e779374caa1e5911b3dbf46d0696d60c24ce/src/libproc/proc_pid.rs) and did not find a proper "mapping" for used structure(`vpi.pvi_cdir.vip_path`).

dm1try: I'm not a "rustman" but does it make sense to add a dependency for a single C call? Is it too hard to interact with C interface from Rust?

dm1try: > I still don't know how to call C call from Rust and I have no way to test it here.

oops, did not see your answer :)

Camotubi: Yep, I'm a newbie on rust .-.
Anyway, could you test the last commit? It should spawn a new instance without crashing.

nixpulvis: @dm1try I think it's fine to add a dependency for this one call. My heuristic for dependencies is, will it be more likely to be maintained and fixed locally here, or upstream in a logical dependency. If there's a crate aiming to cover these kinds of C calls then I think it makes sense to depend on that.

nixpulvis: @dm1try @Camotubi If you're interested in learning a bit about Rust's [FFI](), I recomend you read [this section of the Rust Book on Unsafe Rust](https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html#using-extern-functions-to-call-external-code).

dm1try: ok I implemented a [POC](https://github.com/jwilm/alacritty/commit/4c5b2b9d2d669194fc8a4da86eac54c04135a69f) for retrieving the current working directory of a process on osx. it's not a "rustish" but should be easy to understand.

let me know if you wanna to support it. I was tried to extend `libproc`, but there are many moving parts for me in Rust, sorry)

dm1try: beside retrieving CWD, there is another problem with `start_daemon`
```
Unable to start new Alacritty process: target/debug/alacritty ["--working-directory", "/Users/dmitrydedov"]
```

do we need `procpath` here for Alacritty proc? 😄

Camotubi: I don't think so. Could you try calling ```trim()``` on the cwd. I do remeber having issues with a breakline on the str. Other than that, I don't know why it is failing.

dm1try: forgot to mention, that the problem is reproducible without the changes related to CWD:
```rust
        let args = [
            "--working-directory".into(),
            "/Users/dmitrydedov".to_string(),
        ];
```

it looks like we need an absolute path to alacritty binary.

dm1try: btw, it's not the issue for a usual user because after installing binaries into the system an absolute path is used on the start everywhere, so the change above works as expected.

though an alacritty developer might use `cargo run` or run the binary directly and it will fail

dm1try: screen shot 2019-01-09 at 03 31 06

nixpulvis:
Here is Terminus's implementation for reference. Not sure how well it works.

https://github.com/Eugeny/terminus/blob/3aaf490f5790564d851c29a76c62ef5c5450a2d7/terminus-terminal/src/services/sessions.service.ts#L242

dm1try:
> Not sure how well it works.

It works well, I use the same implementation for a couple of weeks in the AppleScript mentioned in the referenced issue. 
basically `exec(`lsof -p ${this.truePID} -Fn`))[0].toString().split('\n')`
eq `lsof -a -d cwd -p " & shellPID & " -F n | cut -c 2- | tail -n1`

nixpulvis:
@dm1try, cool! Though it's outside the scope of this PR, I'd be curious how well the windows workaround works.

src/event.rs
chrisduerr: @dm1try Might be interesting to see why exactly this is failing. Because running `target/debug/alacritty` from the directory you're running `cargo run` in should work perfectly, right?

Also I wasn't aware that the temporary file directory where the log is written to is so huge on macOS.

chrisduerr:
I'm fine with getting all platforms fixed up if there's a good implementation. Though the primary reason against that is that most people only run one or two platforms, so it's usually easier to separate things.

nixpulvis:
@chrisduerr, depending how far this PR goes we can open another tracking ticket for Windows. Who knows maybe one day I'll learn how windows works, seeing as I have a gaming PC lying around.

chrisduerr:
@nixpulvis Well an issue for Windows already exists. That could be updated if there are any crazy breakthroughs.

Though I'm not sure if the terminus source code is the best choice for a reference implementation, judging by their macOS implementation which forks it out to a separate process.

nixpulvis:
Yeaaaaa that's what I'm saying. But it does look like they have a trick, so it might be possible. 

Sorry I didn't know about #1979.

src/event.rs
dm1try: > Because running target/debug/alacritty from the directory you're running cargo run

yep, it works because I'm in the CWD of the shell.
`rust::Command` inherits CWD of the Alacritty process and it's a home directory, so it makes sense that this fails.

dm1try: precise error
> Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/libcore/result.rs:

setting `.current_dir` in `start_daemon` "helps")

does it behave different on linux?

Camotubi: On linux it works regardless of how you open alacritty (cargo run or directly calling the executable).

dm1try: it's interesting because it's not clear how the path to the binary is resolved.
assuming you do not have alacritty source dir in the `PATH`, so the resolution should be based on the CWD of running alacritty process(if I'm not missing something here)

in my case, after running `cargo run` CWD of the process is home dir.
```
lsof -a -d cwd -p $(pgrep -n alacritty) -F -n | cut -c 2- | tail -n1
/Users/dmitrydedov
```
could you check on linux if the current working directory of running alacritty process is different from home dir?

paulrouget:
If we go with that current PR, please keep `unix` (procfs exists on other unixes), fall back to an empty vec instead of using `expect`. Note: you will likely need to use a `Vec<>` during the mapping/matching otherwise you will have 2 different types (`[_;0]` vs. `[str;2]`).

src/event.rs
dm1try: FYI, the answer is not needed, alacritty changes CWD on osx on the start, see this code
https://github.com/jwilm/alacritty/blob/cf0cece47afde2b1392ea19fc6e7e672c3072861/src/main.rs#L74-L76

so the behavior above is expected

PR #1991, opened 2019-01-08
Fix crash when starting Alacritty on full drives

### LINKED ISSUE #1936: Alacritty can't start with a full disk
Alacritty does not seem to be able to start if it is not able to write to the disk. This is not ideal in case you want to open a terminal to deal with that very situation and are unable to.

This is on Linux, with xorg.

chrisduerr:
Since the Alacritty configuration file is written to the filesystem at
startup, this could create issues when the system does not have any free
space left.

To circumvent this problem, the default configuration is now returned
even when the configuration file could not be created. Instead of
crashing Alacritty, an error is now emitted.

Fixes #1936.


jwilm:
This seems like a great change. Nice work!

PR #2015, opened 2019-01-18
Fix off-by-one error in erase_chars

abulimov:
I had an issue with last characters on the line remaining on the screen even after line they belonged to moved up or down. Easy repro: open any text file with long lines (alacritty.yml for example around line 100) in an editor of your choice, and then move around so the editor scrolls up or down. Last char of the line will not be erased.

Here is the example:
alacritty_bug

It looked like off-by-one kind of error from the start, so it was just a matter of seeing what kind of events are happening when bug is present.

According to rust docs, slicing with [..N] doesn't include N-th element, so we don't have to subtract 1 from num_cols().

On a side note, for some reason the issue doesn't manifest itself with winpty on Windows, only with conpty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Line's last character will no longer escape erasing
+- Resolved off-by-one issue with erasing characters in the last column
 
 ## Version 0.2.7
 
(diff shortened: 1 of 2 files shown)


chrisduerr:
Judging just by the code itself, this *looks* correct to me. However I'm not able to reproduce on Linux at all.

@zacps could you maybe confirm that this issue indeed exists on Windows and is resolved by this fix?

I'm going to run this branch myself for a bit to make sure there is no odd behavior on Linux either, but it should be good to go, thanks for the PR!

zacps:
Can confirm the original issue and that this patch fixes it, thanks!

Just add a changelog entry and it should be good to go!

abulimov:
Thanks for a quick review! I've rebased and added changelog line.

chrisduerr:
Waiting for https://github.com/jwilm/alacritty/pull/2021 to be merged to make a separate release, to prevent potential regressions.

chrisduerr:
Thanks a lot for the PR!

PR #2022, opened 2019-01-20
Add support for double-click bracket-pair selection

### LINKED ISSUE #2014: Smart selection, select everything between brackets
In Terminal.app, you can double-click a starting bracket (for example, `{`), or the closing one, to select it and everything up to and including the bracket pair:

terminal

I believe this behaviour is different (but not exclusive) to the already existing "[smart selection](https://github.com/jwilm/alacritty/issues/30)" functionality. 

I use this functionality a lot when dealing with large (log) outputs that include JSON for example, to quickly see where an object starts and ends. It serves as a quick visual aid to limit the scope in which I have to search for some key.

I know this use case can also be tackled by using `less` or `grep` or some other way, but it's not always obvious when I run a command that I need to find a needle in a haystack, so in Terminal.app I could use this handy functionality to quickly find what I'm looking for, instead of having to execute the command again with some extra commands piped onto it, to achieve the same result.

JeanMertz:
closes #2014

This functionality is a direct copy of the same implementation in Apple's Terminal.app. See the original issue for an example of how it works in that terminal.

Selection logic now behaves as follows (all using the left mouse button):

| click state  | button state | character under cursor | selection type |
|--------------|--------------|------------------------|----------------|
| single click | pressed      | `*`                    | simple         |
| double click | pressed      | `*`                    | semantic       |
| double click | released     | `()[]{}<>`             | bracket-pair   |
| triple click | pressed      | `*`                    | line           |

Example:

alacritty

Since this is my first contribution to Alacritty, and I'm still fairly new at Rust, I'd welcome any and all feedback on how this implementation could be further optimised.

I'm still looking into adding more tests (the current selection tests all assume selection based on mouse drag behaviour), and perhaps a benchmark to prevent performance regressions when working with large buffers.

I did test this locally, including with a _somewhat_ large JSON blob, and saw no performance regression in the selection actions, and no unexpected cross-over situations between the different selection types.


src/term/mod.rs
```diff
@@ -96,6 +98,48 @@ impl Search for Term {
         point
     }
 
+    fn bracket_pair_search(&self, start: Point<usize>, direction: Side, start_char: char, needle: char) -> Option<Point<usize>> {
```
JeanMertz: I'm not entirely happy with the verbosity of this method, so if anyone knows of a better implementation (potentially using a more functional style), let me know!

JeanMertz:
Just a heads-up: this is still on my radar, I just haven't gotten the time to improve the PR yet. I hope I can find some time this weekend.

JeanMertz:
@chrisduerr I finally got around to making the requested changes. Let me know what you think.

src/selection.rs
```diff
@@ -125,6 +135,33 @@ impl Selection {
...
+            None => point,
+        };
+
+        Some(Selection::BracketPair {
+            region: Range {
+                start: Anchor::new(point.into(), side.opposite()),
+                end: Anchor::new(end_point.into(), side),
+            }
+        })
+    }
```
chrisduerr: There's a whitespace missing after the function here.

src/term/mod.rs
```diff
@@ -99,6 +107,48 @@ impl Search for Term {
...
+            if c == end_char {
+                if skip_pairs == 0 {
+                    return true;
+                }
+
+                skip_pairs -= 1;
+            }
+
+            false
+        };
```
chrisduerr: I think this would look a lot better with a match statement.

src/term/mod.rs
```diff
@@ -99,6 +107,48 @@ impl Search for Term {
...
+
+        match direction {
+            Direction::Right => {
+                while let Some(cell) = iter.next() {
+                    if match_bracket(cell.c) {
+                        return Some(iter.cur);
+                    }
+                }
```
chrisduerr: Instead of using `while` loops on iterators, `for` loops should be preferred in general. However we can go even one step further here.

The following should do the job:
`if let Some(cell) = iter.find(|cell| match_bracket(cell.c)) { return Some(cell) }`

src/term/mod.rs
```diff
@@ -41,6 +41,12 @@ use crate::tty;
 pub mod cell;
 pub mod color;
 
+/// Determines the direction the search is going.
+pub enum Direction {
+    Left,
+    Right,
+}
```
chrisduerr: I know I've said myself that we shouldn't use `Side` for the search direction, however I think using the struct still makes sense since it's basically the same either way.

I'm not sure if we should really use the same for both, but I'd propose renaming `Side` to `Direction` and using it to indicate cell direction too.

I'm open for feedback here though, since I'm not 100% sure which approach I like most here.

src/input.rs
```diff
@@ -729,6 +738,18 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+            let elapsed = self.ctx.mouse().last_click_timestamp.elapsed();
+
+            match self.ctx.mouse().click_state {
+                ClickState::DoubleClick
+                    if !button_changed && elapsed < self.mouse_config.double_click.threshold =>
+                {
+                    self.on_mouse_double_click(button, state, point);
+                }
+                _ => {}
+            }
```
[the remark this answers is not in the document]
JeanMertz: The problem I see is that you want a couple different steps to be taken on a mouse action:

1. set the click state according to the current click state and the current mouse button state (e.g. `ClickState::Click` + `ElementState::Pressed` == `ClickState::DoubleClick`)
2. then, you want to perform any action related to the new `ClickState`, plus potentially the current `ElementState` (such as executing any of the possible selection types)

`on_mouse_press` is (mostly) now performing the first step, and all the selections from the second step, since all of those are triggered on `ElementState::Pressed`.

This new selection type doesn't fit that model, since it should act on `Released`, not pressed.

I don't yet see a very clean way to fit that into the current code, without pulling apart the `on_mouse_press` and basically split it into a function that _only_ does (1), and then separate functions that trigger based on the current `ClickState` and `ElementState`, after that, to resolve (2).

I'm not quite sure what you'd prefer. I worked off of this comment you gave before:

> I think it might make sense to move the `on_mouse_double_click` and `on_mouse_triple_click` into the `mouse_input` function and then just passing the pressed/released state as arguments. That should clean stuff up and resolve the issue that you mentioned with expecting the double click method to be called on release too.

But I didn't want to start refactoring all of it just yet.

There are several ways to represent the end-goal, I have no preference for either of them, but if you have an idea of what you'd like to see, let me know and I'll implement that.

src/term/mod.rs
JeanMertz: I could just get rid of the `Direction` enum completely, and either do something like `bracket_pair_search_{right,left}` (there's precedence for this, since there's also `semantic_search_{right,left}`).

I just thought the extra duplication didn't make much sense, and providing the direction as a function argument makes most sense.

I have no feedback on renaming `Side`. I'm fine with that, although I thought it had a bit of a different semantic meaning, as in `Side::Left` represents "the left edge of something", not so much "towards the left of something".

src/input.rs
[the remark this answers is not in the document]
chrisduerr: I'd have to play around with the code myself to see which exact limitations apply, but that would likely take quite a while. So going off the comment I gave earlier, I think that would still make sense.

Using the current functions as single press functions seems fine to me. Then we can just extract the `on_mouse_double_click` and `on_mouse_triple_click` functions into the the `mouse_input` function. So we split single/double/triple click first and then by press/release, instead of the other way around.

However I'm open for alternative solutions. Basically the only thing that doesn't seem like a good solution is the way it's in right now. I don't think there's any way around refactoring. We need our code to stay structured to be able to find stuff and the way it is right now, I don't think that's the case.

If you have a better idea than my proposal, I'm open to that too. I just don't think there's any way around refactoring things at least a little bit. Since as you've mentioned yourself the functionality just doesn't fit into the current structure very well. Tacking it on as a separate code path just doesn't feel like the right approach.

src/term/mod.rs
chrisduerr: > I thought it had a bit of a different semantic meaning, as in Side::Left represents "the left edge of something", not so much "towards the left of something".

That is absolutely the case and why I've complained initially.

> I could just get rid of the Direction enum completely, and do something like bracket_pair_search_{right,left} (there's precedence for this, since there's also semantic_search_{right,left}).

I'm not sure if this is the best choice, however I think it might work quite well so it could be a simple solution that easily fixes all our options.

The common code can be split into a separate method and then be called from the left/right methods.

The callsite looks trivial too. Right now we're just returning `let (direction, side, end_char)` from a closure and then using that to search left/right. But the direction and end_char are only required for the `bracket_pair_search` call, so we could just call the left/right methods in the `find_map` closure and then return `let (end_point, side)` and we'd be good to go.

JeanMertz:
Thanks for the feedback @chrisduerr. I'll give the proposed refactoring a go, but it'll probably have to wait until next weekend.

By the way, as a sort-of test, I recorded me working on this PR in an "[Open Source Contributions](https://www.youtube.com/watch?v=1e53UNn8Y_k)" video. I plan on doing more of those in the future, so perhaps I'll tackle the refactor of this PR in an upcoming video as well.

JeanMertz:
It took me a couple of weeks, but I found some time to do another pass on this pull request, based on the provided feedback (I also [recorded me working](https://youtu.be/rt1jV4skwaA) on it again, in case anyone is interested in these kinds of videos, like I am: [1](https://youtu.be/rt1jV4skwaA), [2](https://youtu.be/22jAipwVIfQ), [3](https://youtu.be/TNyxqhGcpWw)).

I think it's in a mergeable state now, but there are definitely some stylistic changes that could still be made. Also, I think there are quite some gaps in the test coverage when it comes to the (existing) mouse input logic (I've seen the tests pass quite some times, when I could manually validate things not working as expected). I tested things manually, and I _believe_ things still work as they did before, with the added bracket-pair selection feature, but I can't be 100% sure.

One way forward is to merge this, if you are happy with the result, and maybe I can find some time (and gain more knowledge on testing in Rust) somewhere in the next couple of weeks to increase the overall test coverage of this part of Alacritty, or to leave the PR open for the duration of that time, until test coverage is increased.

src/input.rs
```diff
@@ -757,37 +663,163 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         }
     }
 
+    fn update_click_state(&mut self, button: MouseButton) {
+        let elapsed = self.ctx.mouse().last_click_timestamp.elapsed();
+        let button_changed = self.ctx.mouse().last_button != button;
+        let click_state = self.ctx.mouse().click_state;
+
+        self.ctx.mouse_mut().last_click_state = click_state;
```
JeanMertz: I just realised some of my code changes got lost in a rebase. I'm not using `last_click_state` anywhere, which was added to mimic the exact workings of the previous `on_mouse_press` `match` behaviour.

You should probably hold off your review until I manage to put the missing code back in place.

chrisduerr: Thanks for letting me know. Please ping me when you think that everything's ready to go and I'll get on it asap.

JeanMertz:
@chrisduerr I rebased and did some cleanup. Let me know what you think.

JeanMertz:
Thanks @chrisduerr. Your implementation definitely looks much better. Using the existing semantic selection logic definitely trims down on the amount of changes needed.

I still think the existing mouse input handling could use a refactor to make it a bit more clear and easier to work with, but given the fact that you were able to make this work without any of that refactor work might prove me wrong (and even if it doesn't, it's clearly not needed for this PR, so 👍).

chrisduerr:
> I still think the existing mouse input handling could use a refactor to make it a bit more clear and easier to work with, but given the fact that you were able to make this work without any of that refactor work might prove me wrong (and even if it doesn't, it's clearly not needed for this PR, so +1).

I definitely don't doubt that some refactoring could make things better, though it's a tough task to juggle priorities between click count and press/release state.

But as you've rightfully mentioned, if we do still want to refactor the logic here, we should do that separately in a different PR.

PR #2024, opened 2019-01-20
Bump version to 0.2.7

chrisduerr:
This fixes a major regression on Windows which would cause a crash on startup.


chrisduerr:
@zacps Can you verify if this works on Alacritty with the default config? I'd like to make sure that we won't have to push another version bump in a few hours.

Also like someone to test on macOS, just to make sure. Maybe @dm1try or @sodiumjoe could help out?

PR #2037, opened 2019-01-25
Fix URL detection i.c.w. tmux pane borders

svanharmelen:
Properly detect a URL that is right next to a tmux pane border (see screenshot)

image


svanharmelen:
Thanks for the feedback @chrisduerr! And I agree that would indeed be a better, more correct, solution.

 I'll have a go at it and will update the PR accordingly.

svanharmelen:
@chrisduerr it turned out that the correct approach (following RFC 3986) was already used, but the implementation contained a bug. The used `String.find` method returns the byte index but was used as the character index.

I added some logic to convert the byte index to the correct character index, and then noticed the same kind issue with using the `String.split_off` method.

As this is my first Rust PR I guess there are probably more (and maybe better) ways to solve it, but this seemed to make the most sense to me. Yet I'm of course open to any feedback 😉

src/url.rs
```diff
@@ -45,13 +45,21 @@ impl UrlParser {
...
+                    index = char_index;
+                    break;
+                }
+            }
             for i in (0..index - 1).rev() {
                 match self.state.chars().nth(i).unwrap() {
                     'a'...'z' | 'A'...'Z' => (),
                     _ => {
-                        self.state = self.state.split_off(i + 1);
+                        self.state = self.state.chars().skip(i+1).collect::<String>();
```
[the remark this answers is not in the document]
svanharmelen: That is a very nice and clean solution indeed. Manually tested it and as the tests already indicated, it works as expected. Thanks!

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Resolved off-by-one issue with erasing characters in the last column
 - Excessive polling every 100ms with `live_config_reload` enabled
+- Removal of characters before the URL scheme now take UTF8 byte sizes into account
```
chrisduerr: This is a very technical description of what has been fixed (which is great for devs), but doesn't really help the user understand what has been fixed.

I'd propose something more like this:
`- Unicode characters at the beginning of URLs are now ignored`

Of course this is just a suggestion, feel free to improve on it if you have any ideas.

svanharmelen: @chrisduerr, just updated the changelog entry

PR #2047, opened 2019-01-29
Add wayland dark themed client decorations option

trimental:
Adds the option to set the wayland client side decorations theme to a dark one. The colors in the theme are changeable but I chose colors that in my opinion look good on the ConceptFrame. Once a new winit and SCTK version has been released then this PR can be considered merge-able. 

I don't really know that much about alacritty's configuration process so please tell me if I missed something. 

image

Fixes https://github.com/jwilm/alacritty/issues/2092.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -45,9 +45,9 @@ window:
   # Values for `decorations`:
   #     - full: Borders and title bar
   #     - none: Neither borders nor title bar
-  #
-  # Values for `decorations` (macOS only):
+  # Values for `decorations` (wayland and macOS only):
   #     - transparent: Title bar, transparent background and title bar buttons
+  # Values for `decorations` (macOS only):
   #     - buttonless: Title bar, transparent background, but no title bar buttons
   decorations: full
 
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -21,7 +21,9 @@ use glutin::event_loop::EventLoop;
 #[cfg(target_os = "macos")]
 use glutin::platform::macos::{RequestUserAttentionType, WindowBuilderExtMacOS, WindowExtMacOS};
 #[cfg(not(any(target_os = "macos", windows)))]
-use glutin::platform::unix::{EventLoopWindowTargetExtUnix, WindowBuilderExtUnix, WindowExtUnix};
+use glutin::platform::unix::{
+    EventLoopWindowTargetExtUnix, WaylandTheme, WindowBuilderExtUnix, WindowExtUnix,
+};
 #[cfg(not(target_os = "macos"))]
 use glutin::window::Icon;
 use glutin::window::{CursorIcon, Fullscreen, Window as GlutinWindow, WindowBuilder, WindowId};
@@ -150,6 +152,9 @@ impl Window {
         let windowed_context =
             create_gl_window(window_builder.clone(), &event_loop, false, logical)
                 .or_else(|_| create_gl_window(window_builder, &event_loop, true, logical))?;
+
+        // Set Wayland client-side decorations theme
+        Window::set_wayland_theme(event_loop, windowed_context.window(), config);
 
         // Text cursor
         windowed_context.window().set_cursor_icon(CursorIcon::Text);
@@ -391,6 +396,33 @@ impl Window {
     fn window(&self) -> &GlutinWindow {
         self.windowed_context.window()
     }
+
+    #[cfg(not(any(windows, target_os = "macos")))]
+    fn set_wayland_theme(event_loop: &EventLoop<Event>, window: &GlutinWindow, config: &Config) {
+        if config.window.decorations == Decorations::Transparent && event_loop.is_wayland() {
+            let bg = config.colors.primary.background;
+            let fg = config.colors.primary.foreground;
+            let red = config.colors.normal().red;
+            let green = config.colors.normal().green;
+            let blue = config.colors.normal().blue;
+
+            window.set_wayland_theme(WaylandTheme {
+                primary_active: [0xFF, fg.r, fg.g, fg.b],
+                primary_inactive: [0xFF, fg.r, fg.g, fg.b],
+                secondary_active: [0xFF, bg.r, bg.g, bg.b],
+                secondary_inactive: [0x00, 0x00, 0x00, 0x00],
+                close_button_hovered: [0xFF, red.r, red.g, red.b],
+                close_button: [0x00, 0x00, 0x00, 0x00],
+                maximize_button_hovered: [0xFF, green.r, green.g, green.b],
+                maximize_button: [0x00, 0x00, 0x00, 0x00],
+                minimize_button_hovered: [0xFF, blue.r, blue.g, blue.b],
+                minimize_button: [0x00, 0x00, 0x00, 0x00],
+            });
+        }
+    }
+
+    #[cfg(any(windows, target_os = "macos"))]
+    fn set_wayland_theme(_: &EventLoop<Event>, _: &GlutinWindow, _: &Config) {}
 }
 
 #[cfg(not(any(target_os = "macos", windows)))]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,7 +39,7 @@ dependencies = [
  "rustc_tools_util 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.101 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.41 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_yaml 0.8.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_yaml 0.8.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.42 (registry+https://github.com/rust-lang/crates.io-index)",
  "unicode-width 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -72,7 +72,7 @@ dependencies = [
  "rfind_url 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.101 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.41 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_yaml 0.8.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_yaml 0.8.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "signal-hook 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "static_assertions 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "terminfo 0.6.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -675,7 +675,7 @@ name = "foreign-types-macros"
 version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "proc-macro2 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "proc-macro2 1.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "quote 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "syn 1.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
@@ -1477,7 +1477,7 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.4"
+version = "1.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "unicode-xid 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -1506,7 +1506,7 @@ name = "quote"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "proc-macro2 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "proc-macro2 1.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -1869,7 +1869,7 @@ name = "serde_derive"
 version = "1.0.101"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "proc-macro2 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "proc-macro2 1.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "quote 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "syn 1.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
@@ -1886,7 +1886,7 @@ dependencies = [
 
 [[package]]
 name = "serde_yaml"
-version = "0.8.9"
+version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "dtoa 0.4.4 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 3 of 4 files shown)


chrisduerr:
Thanks for proposing this PR. There are a few questions that I'd like to address with this PR.

First of all, if the user wants a custom decoration theme that fits the system theme, shouldn't he use server-side decorations?

Wouldn't it be easier to just support the `buttonless` / `transparent` window decoration options that already exist for macOS? It seems like we don't really need customizable client-side decorations when we can just use our background color as title bar color.

If we do accept this as a separate thing even though it only exists on Linux and it only exists on a single display server on Linux and is very similar to the decoration option, we likely wouldn't want to use fixed colors. It's so specialized already that I can't see people being a huge fan of customized CSDs when at the end they don't fit in with the terminal dark theme and the system theme anyways. So we'd either have to use the existing colors or allow full color customization for the decorations.

What makes Alacritty different than other applications here is that the background and foreground colors are almost always customized I'd say. So "normal" applications have it quite easy to just define CSDs that fit their theme, but if we just decide on 'arbitrary' colors for CSDs it's very likely they won't fit.

trimental:
To answer the first question, some wayland compositors such as mutter (used by gnome) have stated to never support server side decorations, and wayland clients such never assume as much.

Secondly, I don't really know much about the alacritty configuration system but if you want a third option called `background-colored`, or something, as a decorations option instead and then use the background color then I can do that instead.

Thirdly, the CSD colors are arbitrary and can run a risk of not fitting in with the terminal colors; however, saying that, https://github.com/Smithay/client-toolkit/commit/af198d1c3f50f81747d6261de5861bb789a0b379 is also using 'arbitrary' colors so just using standard dark theme colors doesn't seem like much of a compromise at the moment, I dare say people using it just don't want to burn their eyes. That problem could be averted by matching the background color, or allowing full color config control, like you stated however.

I do definitely see where you are coming from though, I run alacritty with server side decorations so I don't have to worry about this personally however people have requested this https://github.com/tomaka/winit/issues/765 so I thought it might be worth a PR.

chrisduerr:
It's unfortunate that Wayland creates problems like this, but if the idea is that every application should just randomly throw some colors together, then I think the thing that makes most sense for Alacritty would be to use the background colors of Alacritty as the background colors for the CSDs.

We should be able to use the `transparent` decoration option for this, because that's basically exactly what it would be:
https://github.com/jwilm/alacritty/blob/master/alacritty.yml#L43

This should make it trivial to have the background color and the foreground color of the window actions, however it seems like we'd also need to specify things like hover/active colors for the buttons. This is a bit more troublesome, since there's absolutely no good way to decide what should be used here, but we might be able to abuse some other colors for this too (like red/blue maybe?).

If it turns out that using existing colors for these actions doesn't do the job, then there'd still be the option to brighten/darken existing colors. So we could make the background XX% lighter or something like that (which would of course have to be reversed for light colors, so lotta work again).

The primary thing I'd be trying to prevent here is that Alacritty just decides on some colors. Because I know I'd **really** hate that myself and that would just annoy me a lot if the colors don't coincidentally line up with what I want (though I'd never use a WM that forces CSDs, but that's a different story).

If it turns out that there's no way to find good colors for the buttons, it would be possible to add colors to the existing color configuration. However this would create new entries in the config file which are completely useless for probably a big majority of users.

trimental:
@chrisduerr does this latest commit do what you were describing?

src/window.rs
```diff
@@ -156,6 +183,43 @@ impl Window {
...
+                    primary_inactive: vscaled_bg,
+                    secondary_active: [0xFF, fg.r, fg.g, fg.b],
+                    secondary_inactive: [0x00, 0x00, 0x00, 0x00],
+                    close_button_hovered: [0xFF, red.r, red.g, red.b],
+                    close_button: [0x00, 0x00, 0x00, 0x00],
+                    maximize_button_hovered: [0xFF, green.r, green.g, green.b],
+                    maximize_button: [0x00, 0x00, 0x00, 0x00],
+                    minimize_button_hovered: [0xFF, blue.r, blue.g, blue.b],
+                    minimize_button: [0x00, 0x00, 0x00, 0x00],
+                });
```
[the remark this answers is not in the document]
trimental:
```rust
let mut scale = 382.5 / (f32::from(bg.r) + f32::from(bg.g) + f32::from(bg.b));
```
^ This scales the window focused color by increasing the brightness if the background color is dark and darkening if its light. I made the focused color equal to a lightened/darkened version of the bg color for distinction however I'll change it to that then.

These are the color options
* primary_active = window headerbar color when the window is focused
* primary_inactive = window headerbar color when the window is unfocused
* secondary_active = color of button symbols
* secondary_inactive = ignored in the window theme that winit uses

The buttons are pretty self explanatory

I don't really know what to do about the other points like the buttons. Perhaps they can be followed up once you have the chance to test this on sway.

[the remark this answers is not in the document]
chrisduerr: Yeah, I don't think it makes much sense to discuss this in depth until I've tested it. I'll try to setup sway this weekend and play around with it and then follow up with another review.

Thanks for your explanations.

JMurph2015:
Yeah, I don't know where I land on the discussion here.  On one hand, if the automatic colors from Alacritty look good in context to the Alacritty theme, that's pretty good.  On the other hand, exposing essentially full color control in the config file would enable users to match their system theme much more faithfully.  

Honestly, I'd settle for either.  Just anything to end this bright grey title bar on a Monokai themed Alacritty in an Adapta (dark Material UI) themed GNOME.  It's such an eyesore, it makes it hard to use Alacritty, performance be damned.

Also, w.r.t. user demographics.  Considering Wayland is well on its way to replacing X11, I would guess the number of target users that this impacts is larger than you give it credit for.  Sure, Alacritty is probably used on macOS, but macOS users (generalizing) are not exactly the type to go hunt down a compile-from-source, Rust language, alternative terminal emulator.  It seems pretty firmly in the "Linux geek" territory, which is also where Wayland is.  This having it stand out like a sore thumb there is pretty rough.  Just my 2¢.

chrisduerr:
> On the other hand, exposing essentially full color control in the config file would enable users to match their system theme much more faithfully

This is exactly what SSDs are for. There is no point in using CSDs when you're trying to match your system theme anyways. This is exactly the issue with Wayland's approach, so CSDs should be things that fit with the application, rather than the theme of the system. If you want decorations that fit the system, you should use SSDs and a compositor which supports them.

Note that this would also add things to the configuration file which would be completely useless on all other platforms. So I really see no point in giving in too much here for something that Wayland has essentially brought upon itself. If Wayland wants everyone to use CSDs, the consequence of things not matching up with your system theme are inevitable. Using the other terminal colors should fix this at least a little bit, but there's not a lot to be done here.

> Considering Wayland is well on its way to replacing X11

Wayland still has a long, long way to go. It would be extremely surprising to me if Alacritty's Wayland userbase would be bigger than either X11 or macOS users, based on the activity I've seen on the issue tracker.

JMurph2015:
Problem is, Wayland *without SSD* is the default on GNOME (and by extension, basically all of Ubuntu's recent versions) which is the at the very least a large majority (maybe closer to overwhelming majority, but there aren't good stats for this) of Linux desktop users right there?

I honestly have no idea about OS breakdown for Alacritty, but it very well could be the case that there are disproportionately more problems getting things to work on Mac and certainly Windows due to their unique quirks.  Or there are more Mac hackers than I thought.

chrisduerr:
> which is the at the very least a large majority (maybe closer to overwhelming majority, but there aren't good stats for this) of Linux desktop users right there?

I've got no definite metrics for this, but it seems like Ubuntu/GNOME users aren't as dominant with Alacritty as they are with Linux in general. Which is probably at least partially due to Alacritty not being in the repos and Alacritty generally attracting 'powerusers' more, which generally have a more diverse spread across Linux distros.

Regardless of how big the userbase actually is though, this is a problem that GNOME has put upon its users. The only way for me to interpret this decision is that they do not wish for consistent decorations across Applications, so it's not really up to Alacritty to work around this.

Of course this doesn't mean we can't still do the best job we can at providing great CSDs. However the focus with CSDs should (at least in my opinion) always be on fitting with the application, rather than the system. After all that's exactly what CSDs are for.

JMurph2015:
I'm cool with that, focusing on blending with Alacritty's theme is much better than blending with neither, and it's a valid point about CSD's being better suited for blending with the client (though really the killer use for them is to compress the controls area like you see with Chrome and Firefox on Windows, regardless of theme)

src/window.rs
```diff
@@ -156,6 +158,36 @@ impl Window {
...
 
+        #[cfg(not(any(windows, target_os = "macos")))]
+        {
+            if window_config.decorations() == Decorations::Transparent && event_loop.is_wayland() {
+                let bg = colors.primary.background;
+                let mut scale = 382.5 / (f32::from(bg.r) + f32::from(bg.g) + f32::from(bg.b));
+                scale /= 3.;
+                let scaled_bg = [0xFF, (f32::from(bg.r) * scale) as u8, (f32::from(bg.g) * scale) as u8, (f32::from(bg.b) * scale) as u8];
+                scale *= 1.5;
+                let vscaled_bg = [0xFF, (f32::from(bg.r) * scale) as u8, (f32::from(bg.g) * scale) as u8, (f32::from(bg.b) * scale) as u8];
```
chrisduerr: Lines should be shorter than 100 characters. Generally following rustfmt is a good idea, even though it's not possible to run it on the whole codebase yet.

src/config/mod.rs
```diff
@@ -287,10 +287,12 @@ impl Default for Alpha {
...
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[derive(Debug, Copy, Clone, Deserialize, PartialEq, Eq)]
 pub enum Decorations {
     Full,
+    #[cfg(not(widows))]
     Transparent,
+    #[cfg(target_os = "macos")]
     Buttonless,
     None,
 }
```
chrisduerr: We should make these either case insensitive or change them to lowercase, otherwise we'll break existing configuration files.

src/window.rs
```diff
@@ -156,6 +158,36 @@ impl Window {
...
+                    primary_active: scaled_bg,
+                    primary_inactive: vscaled_bg,
+                    secondary_active: [0xFF, fg.r, fg.g, fg.b],
+                    secondary_inactive: [0x00, 0x00, 0x00, 0x00],
+                    close_button_hovered: [0xFF, red.r, red.g, red.b],
+                    close_button: [0x00, 0x00, 0x00, 0x00],
+                    maximize_button_hovered: [0xFF, green.r, green.g, green.b],
+                    maximize_button: [0x00, 0x00, 0x00, 0x00],
+                    minimize_button_hovered: [0xFF, blue.r, blue.g, blue.b],
+                    minimize_button: [0x00, 0x00, 0x00, 0x00],
```
chrisduerr: These are all mixed up I think. Testing on weston, the first three fields are RGB, not the last three.

src/window.rs
```diff
@@ -156,6 +158,36 @@ impl Window {
             .or_else(|_| create_gl_window(window_builder, &event_loop, true))?;
         window.show();
 
+        #[cfg(not(any(windows, target_os = "macos")))]
+        {
+            if window_config.decorations() == Decorations::Transparent && event_loop.is_wayland() {
+                let bg = colors.primary.background;
+                let mut scale = 382.5 / (f32::from(bg.r) + f32::from(bg.g) + f32::from(bg.b));
+                scale /= 3.;
```
chrisduerr: This will be infinity when the background is completely black, casting it to `u8` then sets it to 0, leading to the background not getting scaled.

src/window.rs
chrisduerr: Actually, looks like they are BGR.

chrisduerr:
I've tested this on Wayland now and while I think that this works pretty well, there are some major issues besides what I've already mentioned in the reviews.

The colors (red, green, blue after fixing them) look *great* for my configuration file. But the window icons are pretty much unreadable on my colors, I'm not sure if I've mentioned this before but we shouldn't assume that light text on the colored backgrounds will provide great contrast, since that's not the case with a lot of themes. I think it would be much better to use the r/g/b colors as the foreground of the symbols on hover, but I don't think that's possible? Another solution would be to make the foreground of the icons the background color on hover, but I don't think that works either?

Since colors don't seem to work, I'd probably choose to use the inactive color as the color of all icons. That way icons aren't colored anymore, but at least it will always be readable and consistent.

Another problem is that the scaling doesn't work well with light themes (like 0xffffff background) and with a solid black background it just doesn't work at all. It's also counter-intuitive that a focused window has a darker border than an unfocused one.

Now these are a lot of complaints and not a lot of solutions, but hopefully we can figure something out. If you have any more questions, please let me know. But at its current state, this looks like it still needs a bit of refinement work.

src/window.rs
trimental: Ah welcome to another ‘feature’ of wayland.

Pixels written to the buffer must be of native endian (little endian on Linux). Technically two versions of the colours should be supplied and used according to the system endianness.

JMurph2015: There should be code out there to switch the endianness to system endian, or it would be trivial to do for some custom type.

trimental: Yeah there is. SCTK used to hardcode for performance however because this would only be called once I think you’re right that a trivial utility function would be best.

JMurph2015:
Maybe the MVP here is to just pick some defaults and delegate to config file overrides?  Maybe make it a separate feature to do auto-determination?

chrisduerr:
> Maybe the MVP here is to just pick some defaults and delegate to config file overrides? Maybe make it a separate feature to do auto-determination?

If we want to support CSDs, we should support CSDs that will always fit the Alacritty theme properly. If the goal is to pick some arbitrary default, then SSDs should be used instead.

JMurph2015:
> > Maybe the MVP here is to just pick some defaults and delegate to config file overrides? Maybe make it a separate feature to do auto-determination?
> 
> If we want to support CSDs, we should support CSDs that will always fit the Alacritty theme properly. If the goal is to pick some arbitrary default, then SSDs should be used instead.

It just occurs to me that it is a rather challenging proposition to create an attractive matching set of window decorations for any given Alacritty color configuration.  And for that matter, what constitutes "attractive" is almost certainly going to depend on the user.  What you are looking for here doesn't really exist for arbitrary color pallettes like what Alacritty allows.
Thus I am suggesting that a reasonable attempt (that works with say 70% of configs) should be mainlined in the short term, with override support for the remaining 30%.  Then as more effort is spent in this area, perhaps fewer and fewer users will find the overrides useful.

JMurph2015:
Just an FYI, I've been working on some upstream changes that would enable recoloring the icon stroke color, if that helps.  See Smithay/client-toolkit#68 for more details.  `winit` still uses a version that is a few versions old now, so it would have to be updated to use a more recent version of SCTK, but that is a decent bit of work.

chrisduerr:
@trimental I've decided to go for the simple route and just make the window decorations the inverse of the window content. So the foreground color is used as the titlebar and the background color is used as the button colors. Then on hover it uses the red/green/blue colors of the colorscheme.

Testing it with weston this seemed to work out pretty well, so I think we should be good to go with this. Unfortunately I think this requires #2438 to land first, since unfortunately there hasn't been a glutin release which includes this feature since january.

XVilka:
There are some conflicts that should be solved first.

alacritty.yml
```diff
@@ -45,9 +45,9 @@ window:
   # Values for `decorations`:
   #     - full: Borders and title bar
   #     - none: Neither borders nor title bar
-  #
-  # Values for `decorations` (macOS only):
+  # Values for `decorations` (wayland and macOS only):
```
kchibisov: `wayland` -> `Wayland`

alacritty/src/window.rs
```diff
@@ -151,6 +153,9 @@ impl Window {
             create_gl_window(window_builder.clone(), &event_loop, false, logical)
                 .or_else(|_| create_gl_window(window_builder, &event_loop, true, logical))?;
 
+        // Set Wayland client-side decorations theme
+        Window::set_wayland_theme(event_loop, windowed_context.window(), config);
```
kchibisov: I think we should follow our X11 approach here and move call to `#[cfg(not(any(..))` and do event_loop.is_wayland() then call `set_wayland_theme`.

alacritty/src/window.rs
```diff
@@ -391,6 +396,33 @@ impl Window {
...
+                maximize_button_hovered: [0xFF, green.r, green.g, green.b],
+                maximize_button: [0x00, 0x00, 0x00, 0x00],
+                minimize_button_hovered: [0xFF, blue.r, blue.g, blue.b],
+                minimize_button: [0x00, 0x00, 0x00, 0x00],
+            });
+        }
+    }
+
+    #[cfg(any(windows, target_os = "macos"))]
+    fn set_wayland_theme(_: &EventLoop<Event>, _: &GlutinWindow, _: &Config) {}
```
kchibisov: We don't need this method on macOS/Windows at all if we follow X11 approach.

alacritty/src/window.rs
chrisduerr: We do.

chrisduerr:
Blocked by https://github.com/rust-windowing/winit/pull/1209.

kchibisov:
Blocked again by https://github.com/rust-windowing/winit/pull/1444

kchibisov:
Closing in favor of  #3455.

PR #2054, opened 2019-01-30
Fix reloading with empty config

chrisduerr:
When loading an empty configuration file, Alacritty only prints an info
message and then proceeds to load the default config. However when
reloading the configuration file it would throw a hard error.

This has been fixed and a hard error is now only thrown when an error is
returned during reload which isn't the empty file error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Resolved off-by-one issue with erasing characters in the last column
 - Excessive polling every 100ms with `live_config_reload` enabled
 - Unicode characters at the beginning of URLs are now properly ignored
+- Remove error message when reloading an empty config
 
 ## Version 0.2.7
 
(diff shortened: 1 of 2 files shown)


nixpulvis:
I've been able to verify this does indeed work, though I struggled a bit with trying to replicate by running `rm` then `touch` on the config file. It would seem that events stop coming in when a file is deleted and recreated, or something is strange here. However, when I tested by simply deleting everything from my existing config it worked as intended now, vs on master where it put's a nasty red message on my screen :tada:

I'm also noticing that sometimes the message appears after breaking the config and sometimes it doesn't. This is not any different than master, and if it is indeed an issue, I don't see it being related to this PR. I'm still not really sure when/why this is happening, but I'm happy to dig more into it. For now though, I'm going to sleep :wink: 

### Aside

Also I find it just a little strange that an empty file works, but a file containing only whitespace doesn't. I understand _why_ this is happening, but it's just a little weird to run into this error when it's a file that's effectively empty:

```
[ERROR] Unable to load default config: Problem with config: EOF while parsing a value
```

chrisduerr:
The problem with the file containing only whitespace is that it's actually serde giving out there. So it's not a specialized error but a straight-up serde crash.

nixpulvis:
Yea, but if my memory serves (it was like > 2 years ago now) I've dealt with this kind of issue with Serde before, and it can be avoided. I'll have to play with the deserialization to figure out exactly how. But that's probably best served as another PR.

PR #2059, opened 2019-02-01
 Fix regression in the URL launcher config 

### LINKED ISSUE #2058: How to disable URL launcher?
How to completely disable URL launcher?

Every time I am trying to highlight an URL and then browser shows up,

I would like to disable it.

```yaml
url:
    # URL launcher
    #
    # This program is executed when clicking on a text which is recognized as a URL.
    # The URL is always added to the command as the last parameter.
    #
    # Default:
    #   - (macOS) open
    #   - (Linux) xdg-open
    #   - (Windows) explorer
    #launcher: xdg-open
```

chrisduerr:
Due to the merging of configuration files on all platforms, it has been
made impossible to completely disable URL launching without still
executing some kind of program like `true`.

Setting the launcher to `None` in the config, will now disable it
completely.

This fixes #2058.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -291,6 +291,8 @@ mouse:
     # This program is executed when clicking on a text which is recognized as a URL.
     # The URL is always added to the command as the last parameter.
     #
+    # When set to `None`, URL launching will be disabled completely.
+    #
     # Default:
     #   - (macOS) open
     #   - (Linux) xdg-open
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -116,12 +116,38 @@ pub struct Mouse {
 #[derive(Clone, Debug, Deserialize, PartialEq, Eq)]
 pub struct Url {
     // Program for opening links
-    #[serde(deserialize_with = "failure_default")]
+    #[serde(deserialize_with = "deserialize_launcher")]
     pub launcher: Option<CommandWrapper>,
 
     // Modifier used to open links
     #[serde(deserialize_with = "deserialize_modifiers")]
     pub modifiers: ModifiersState,
+}
+
+fn deserialize_launcher<'a, D>(deserializer: D) -> ::std::result::Result<Option<CommandWrapper>, D::Error>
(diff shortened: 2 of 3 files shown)


alacritty.yml
```diff
@@ -291,6 +291,8 @@ mouse:
     # This program is executed when clicking on a text which is recognized as a URL.
     # The URL is always added to the command as the last parameter.
     #
+    # When set to `None`, URL launching will be disabled completely.
+    #
```
nixpulvis: Would it be possible to set the value `None` on the whole `url` key? That might be even better.

src/config/mod.rs
```diff
@@ -116,14 +116,40 @@ pub struct Mouse {
...
 }
 
+fn deserialize_launcher<'a, D>(deserializer: D) -> ::std::result::Result<Option<CommandWrapper>, D::Error>
+    where D: de::Deserializer<'a>
+{
+    // Deserialize to generic value
+    let val = match serde_yaml::Value::deserialize(deserializer) {
+        Ok(val) => val,
+        Err(err) => {
+            error!("Problem with config: {}; using default value", err);
```
nixpulvis: Is "using default value" the best message? What about "disabling URL luancher". I know this is a departure from convention, but it might be nice to start pointing users towards the failing configs in the message.

alacritty.yml
chrisduerr: It would be possible, however that might cause issues in the future I'd imagine.

Like this, it disables one specific feature, setting the whole `url` key to `None` would potentially disable multiple features.

Imagine we one day introduce a `url.copy_to_clipboard` option, which just copies the URL to the clipboard when clicking on it. Then it wouldn't be possible to disable just one of these.

So I think the current way is a good way to go about it.

src/config/mod.rs
chrisduerr: Great point, this is just the generic error but with most other specialized deserializers we say exactly which value is used.

PR #2063, opened 2019-02-02
Remove unused coordinate from rect shader

mstoeckl:
To test, verify that strikeout, underline, and visual bell behave as before.

### DIFF
--- a/res/rect.f.glsl
+++ b/res/rect.f.glsl
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 #version 330 core
-in vec4 color;
+
+uniform vec4 color;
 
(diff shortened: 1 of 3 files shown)


res/rect.f.glsl
```diff
@@ -12,11 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 #version 330 core
-in vec4 color;
+
+uniform vec4 col;
```
nixpulvis: Super not-picky, but I'd prefer `color` unless it's like this for a reason I don't understand. Alacritty does a lot with columns (though to be fair, not in this context) and I generally like avoiding potential reading hazards.

mstoeckl: So do I.

PR #2065, opened 2019-02-02
Scale IME position by hidpi_factor.

### LINKED ISSUE #2056: Input method select list position is not correct
MacOS Mojave 10.14.2
Alacritty v0.2.7

awused:
Fixes #2056

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remove error message when reloading an empty config
 - Allow disabling URL launching by setting the value of `mouse.url.launcher` to `None`
 - Corrected the `window.decorations` config documentation for macOS
+- Scale IME position by DPI
 
 ## Version 0.2.7
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remove error message when reloading an empty config
 - Allow disabling URL launching by setting the value of `mouse.url.launcher` to `None`
 - Corrected the `window.decorations` config documentation for macOS
-- Scale IME position by DPI
+- Fix IME position on HiDPI displays
 
 ## Version 0.2.7
 
--- a/src/display.rs
+++ b/src/display.rs
@@ -488,17 +488,17 @@ impl Display {
 
     /// Adjust the IME editor position according to the new location of the cursor
     pub fn update_ime_position(&mut self, terminal: &Term) {
-        use crate::index::{Column, Line, Point};
-        use crate::term::SizeInfo;
-        let Point{line: Line(row), col: Column(col)} = terminal.cursor().point;
-        let SizeInfo{cell_width: cw,
-                    cell_height: ch,
-                    padding_x: px,
-                    padding_y: py, ..} = *terminal.size_info();
+        let point = terminal.cursor().point;
+        let SizeInfo{
+            cell_width: cw,
+            cell_height: ch,
+            padding_x: px,
+            padding_y: py, ..
+        } = *terminal.size_info();
         let dpr = self.window().hidpi_factor() as f32;
 
-        let nspot_y = (py + (row + 1) as f32 * ch / dpr) as i32;
-        let nspot_x = (px + col as f32 * cw / dpr) as i32;
+        let nspot_x = ((px + point.col.0 as f32 * cw) / dpr) as i32;
+        let nspot_y = ((py + (point.line.0 + 1) as f32 * ch) / dpr) as i32;
         self.window().set_ime_spot(LogicalPosition::from((nspot_x, nspot_y)));
     }
 }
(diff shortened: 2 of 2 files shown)


src/display.rs
```diff
@@ -495,8 +495,10 @@ impl Display {
                     cell_height: ch,
                     padding_x: px,
                     padding_y: py, ..} = *terminal.size_info();
-        let nspot_y = (py + (row + 1) as f32 * ch) as i32;
-        let nspot_x = (px + col as f32 * cw) as i32;
+        let dpr = self.window().hidpi_factor() as f32;
+
+        let nspot_y = (py + (row + 1) as f32 * ch / dpr) as i32;
+        let nspot_x = (px + col as f32 * cw / dpr) as i32;
```
chrisduerr: You're not including the padding in the DPI calculation here, is that intentional? We scale our padding by DPI, so I think it might be relevant to include it? I don't have any IME thingy setup myself, but it should be simple to test this on a monitor without 1.0 DPI by setting the padding to a large value, that should make things obvious.

If this is indeed the case, I'd propose going for a different approach though. To make sure things are done properly I'd construct a `PhysicalPosition` using `nspot_y` and `nspot_x` completely unscaled, then applying the scaling when calling `psize.to_logical(dpr)`. That should make sure the calculation is done properly.

awused:
Made requested changes.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remove error message when reloading an empty config
 - Allow disabling URL launching by setting the value of `mouse.url.launcher` to `None`
 - Corrected the `window.decorations` config documentation for macOS
+- Scale IME position by DPI
```
chrisduerr: This will likely not tell the user at all what actually got fixed, I'd recommend putting it in simpler terms like this:

```
- Fix IME position on hidpi displays
```

src/display.rs
```diff
@@ -488,15 +488,17 @@ impl Display {
...
-                    padding_x: px,
-                    padding_y: py, ..} = *terminal.size_info();
-        let nspot_y = (py + (row + 1) as f32 * ch) as i32;
-        let nspot_x = (px + col as f32 * cw) as i32;
+        let point = terminal.cursor().point;
+        let SizeInfo{
+            cell_width: cw,
+            cell_height: ch,
+            padding_x: px,
+            padding_y: py, ..
```
chrisduerr: The `..` should be on the next line.

src/display.rs
```diff
@@ -488,15 +488,17 @@ impl Display {
...
+        } = *terminal.size_info();
+        let dpr = self.window().hidpi_factor() as f32;
+
+        let nspot_x = ((px + point.col.0 as f32 * cw) / dpr) as i32;
+        let nspot_y = ((py + (point.line.0 + 1) as f32 * ch) / dpr) as i32;
         self.window().set_ime_spot(LogicalPosition::from((nspot_x, nspot_y)));
```
chrisduerr: As I've mentioned, we should be making use of the build-in winit functionality and use `PhysicalPosition` and then convert it using the `.to_logical` method.

CHANGELOG.md
nixpulvis: HiDPI even.

src/display.rs
awused: You only said to do that "If this is indeed the case," and not being familiar with your codebase I opted to not do anything unnecessary. It's done now.

src/display.rs
awused: Is there a style guide? You don't seem to run rustfmt, and I'm not familiar with rust conventions.

CHANGELOG.md
awused: Done. Though this would also affect low DPI displays like projectors, but the code only ever mentions high DPI so this is consistent.

src/display.rs
nixpulvis: I haven't seen a style guide for this repo, but generally all Rust examples I've seen, and more generally across languages it's rare to put all items separated by a comma on separate lines except the last one, even if it's an `..` like thing.

The way it is now, it's too easy to miss.

nixpulvis: FWIW, there are many Rust style guide in the community, the most official (and incomplete) probably being https://doc.rust-lang.org/1.0.0/style/, however it doesn't mention this case.

chrisduerr: Generally I try to follow `rustfmt` and I plan to run it on the complete repo some time in the future. However since there are constantly new PRs coming in, this is extremely difficult to do.

awused: As long as you're asking people to do things manually there are going to be differences and omissions. In this case I would normally have put .. on its own line but since I had to format it manually I missed it, which is why I asked if there were a style guide so I could at least have a set of rules to keep in mind. 

@nixpulvis Do you mean now as in how the code is now, or now as in how it was when this comment thread started?

chrisduerr: > As long as you're asking people to do things manually there are going to be differences and omissions

Which is why I've pointed it out in the review. It does put an additional strain on the review process and it's easier to miss things, however just running `rustfmt` on everything isn't as simple as it might sound at first.

Though the plan to run `rustfmt` came up originall when there were like 40/50+ open PRs. So we're getting closer and closer to an acceptable number of pending changes to put through something like this.

awused: I only skimmed the documentation since you asked me to make this three line change myself, so I did miss that. But since you're planning to run rustfmt anyway I'm not sure it was worth the time to restyle existing code in such a small CL.

PR #2066, opened 2019-02-03
Simplify text shader

mstoeckl:
This should be orthogonal to #2063. I do not observe any performance improvement or regression, but assume that reducing the total amount of work done cannot hurt.

This PR:
* Replaces the `projection` 4x4 matrix uniform with a single vec4 ; the former contained mostly zeros.
* Specializes the text vertex shader to act on instances of rectangles. [I am unsure about the utility of this change. Keep or drop?]
* Moves most of the y-flipping logic out of the vertex shader, getting rid of a uniform in the process.
* Disables updates to the depth buffer. I haven't figured out how to make OpenGL not allocate it (or the stencil buffer) in the first place. [I guessed at the correct place to put the `gl::DepthMask` call; is there a better spot?]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,7 +35,6 @@ dependencies = [
  "arraydeque 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "base64 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "bitflags 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
- "cgmath 0.17.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "clap 2.32.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "copypasta 0.0.1",
  "dirs 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


chrisduerr:
@mstoeckl Just out of curiosity, how did you attempt to measure the performance difference? Alacritty isn't usually bottlenecked by rendering on modern systems, so it doesn't make sense to run benchmarks like vtebench.

I'd probably recommend running Alacritty with the render timer enabled to check if there's additional free time in between rendering passes. Testing with super small font sizes to increase grid size could also be helpful.

mstoeckl:
Indeed, the code is CPU limited, my guess in the latency thread was wrong. I have enabled the render timer as per suggestion, although it isn't particularly useful; all the rendering is "done" in less than 1 msec. Further profiling (with `perf`) confirms that the terminal logic remains the most expensive task.

My ad-hoc testing methods:
* Running a vtebench file under `intel_gpu_time` -- no statistically significant difference is observed; for the ascii random write dataset, CPU% is 130-131%, GPU% is 7.8-7.9%, both pre and post-PR.
* Using e.g. `apitrace trace alacritty -e cat testfile`, and then replaying the test file with `apitrace replay alacritty.N.trace`. I haven't yet figured out how to get rid of vsync on the replays, but this does let one push the CPU usage down to e.g. 10.2%, with GPU at 5.2%. This makes it easier to isolate e.g. the cost of uploading new glyph texture data for unicode-heavy tests, small as it is.

(I'm submitting these changes not because of the (unmeasurable) performance impact, but because I had already written them, and they do simplify the shader logic and remove a dependency.)

mstoeckl:
I have a more detailed measurement, produced using `apitrace trace alacritty -e hexdump /dev/urandom` for 10-20 seconds, followed by `glretrace --pgpu alacritty.N.trace` to rerun the program output with time queries to record draw call times on the GPU side, and a script to parse the results and compute averages. The average times for the background pass are 454 μs +/- 9 μs , and for the foreground pass are 218 μs +/- 7 μs. (Uncertainties are standard deviations.) The discrepancy in means is 0.4 μs for the background pass, and 2 μs for the foreground pass, in both cases statistically insignificant, and likely dominated by variation due to unshared system environment (i.e, /dev/urandom state.) All but one cell of the background pass is `discard`ed.

chrisduerr:
Thanks for going into so much detail with testing this!

src/display.rs
```diff
@@ -150,6 +151,11 @@ impl Display {
         let mut viewport_size = window.inner_size_pixels()
             .expect("glutin returns window size").to_physical(dpr);
 
+        // Disable depth mask, as the renderer never uses depth tests
+        unsafe {
+            gl::DepthMask(gl::FALSE);
+        }
```
chrisduerr: I'm not sure if this is the right place for this. Is there any specific reason why this isn't done at the top of the `QuadRenderer::new`?

PR #2077, opened 2019-02-05
Set window title on Wayland

### LINKED ISSUE #1582: wmclass is null in Wayland
### Which operating system does the issue occur on?

Fedora 28, 64-bit Linux with Wayland

### If on linux, are you using X11 or Wayland?

Wayland

### Description of the problem

Alacritty shows up as an unknown icon and application, e.g.:

unknown-app

Using Gnome's Looking Glass (alt+f2 and run 'lg'), we can see that the `wmclass` of alacritty is null:

wmclass-null-alacritty

I _believe_ this causes running instances of alacritty to not be associated to the `alacritty.desktop` file.

Not sure which part of alacritty's upstream dependencies sets this `wmclass` value. If this belongs in a more general Wayland ticket, feel free to close it/move it.

### LINKED ISSUE #1875: Title is set to "Unbekannt" ("Unknown") and there's no icon if started as Wayland client

Archlinux with Gnome Shell 3.30. System is up to date.

> If on linux, are you using X11 or Wayland?

Wayland.

Picture to visualize everything:
Screenshot
The picture shows part of the application menu of the Gnome Shell. The first icon is the icon for alacritty if started as Wayland client, the text next to it is its title ("Unbekannt", German for "Unknown"). The icon below shows the correct behaviour (terminal icon as icon, title is "Alacritty") after starting alacritty as XWayland client.

It would be nice if alacritty shows the correct behaviour regardless of X11 or Wayland.

chrisduerr:
Fixes #1582.
Fixes #1875.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -38,18 +38,6 @@ static WINDOW_ICON: &'static [u8] = include_bytes!("../assets/windows/alacritty.
 /// In X11, this the default value for the `WM_NAME` property.
 pub const DEFAULT_TITLE: &str = "Alacritty";
 
-/// Default text for general window class, X11 specific.
-///
-/// In X11, this is the default value for the `WM_CLASS` property. The
-/// second value of `WM_CLASS` is **never** changed to anything but
-/// the default value.
-///
-/// ```ignore
-/// $ xprop | grep WM_CLASS
-/// WM_CLASS(STRING) = "Alacritty", "Alacritty"
-/// ```
-pub const DEFAULT_CLASS: &str = "Alacritty";
-
 /// Window errors
 #[derive(Debug)]
 pub enum Error {
@@ -150,8 +138,7 @@ impl Window {
 
         let title = options.title.as_ref().map_or(DEFAULT_TITLE, |t| t);
         let class = options.class.as_ref().map_or(DEFAULT_TITLE, |c| c);
-        let window_builder = Window::get_platform_window(title, window_config);
-        let window_builder = Window::platform_builder_ext(window_builder, &class);
+        let window_builder = Window::get_platform_window(title, class, window_config);
         let window = create_gl_window(window_builder.clone(), &event_loop, false)
             .or_else(|_| create_gl_window(window_builder, &event_loop, true))?;
         window.show();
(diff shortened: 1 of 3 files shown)


src/window.rs
```diff
@@ -272,7 +272,11 @@ impl Window {
     )]
     fn platform_builder_ext(window_builder: WindowBuilder, wm_class: &str) -> WindowBuilder {
         use glutin::os::unix::WindowBuilderExt;
-        window_builder.with_class(wm_class.to_owned(), "Alacritty".to_owned())
+        window_builder
+            // X11
+            .with_class(wm_class.to_owned(), "Alacritty".to_owned())
+            // Wayland
+            .with_app_id("Alacritty".to_owned())
```
nixpulvis: Might as well make a let binding for the string at this point. Makes it a bit more clear what this is being used as anyway.

```rs
let title = "Alacritty".to_owned();
```

chrisduerr:
I've decided to refactor the platform-specific window builder a bit to remove pointless duplication. It should work fine on all platforms, however it still needs a sanity check and some testing.

src/window.rs
```diff
@@ -262,35 +249,14 @@ impl Window {
...
-    #[cfg(
-        any(
-            target_os = "linux",
-            target_os = "freebsd",
-            target_os = "dragonfly",
-            target_os = "openbsd"
-        )
-    )]
-    fn platform_builder_ext(window_builder: WindowBuilder, wm_class: &str) -> WindowBuilder {
+    #[cfg(not(any(target_os = "macos", windows)))]
```
nixpulvis: Shouldn't `windows` be in quotes as `"windows"` here?

Oh... it's not a `target_os` but a whole `cfg` of it's own. This seems weird, but probably has a reason? Not an issue with this PR.

src/window.rs
```diff
@@ -38,18 +38,6 @@ static WINDOW_ICON: &'static [u8] = include_bytes!("../assets/windows/alacritty.
...
-///
-/// In X11, this is the default value for the `WM_CLASS` property. The
-/// second value of `WM_CLASS` is **never** changed to anything but
-/// the default value.
-///
-/// ```ignore
-/// $ xprop | grep WM_CLASS
-/// WM_CLASS(STRING) = "Alacritty", "Alacritty"
-/// ```
-pub const DEFAULT_CLASS: &str = "Alacritty";
```
nixpulvis: Do we really want to remove all support for separate classes and titles under X11? I'm personally pretty indifferent, but I don't really know all the cases we might want to use window classes?

Given that Wayland doesn't seem to have window classes (this is correct right?), I can understand this removal.

nixpulvis:
I can confirm no regressions on X11 inside Linux, nice refactoring.

src/window.rs
chrisduerr: I'm not familiar with Rust's internals, but I believe this is related to operating system internals.

FreeBSD, OpenBSD, macOS, Linux are all at least  unix-ish, but Windows is a completely separate architecture. That's probably why `#[cfg(windows)]` and `#[cfg(unix)]` exist, but I don't think `#[cfg(macos)]` does.

src/window.rs
chrisduerr: We aren't. The functionality hasn't been removed. The only thing that has been removed is a separation between two constants which should always be the same anyways.

The `DEFAULT_CLASS` constant was actually only used in `src/cli.rs`, even the default class in the `window.rs` was using `DEFAULT_TITLE`. If `_TITLE` is the best name or not could be discussed here, maybe `DEFAULT_NAME` would be better. But I don't think there's any point in storing the same thing twice.

src/window.rs
nixpulvis: Interesting, thanks for some insight.

src/window.rs
nixpulvis: Sure, I agree with that. I think I agree that `DEFAULT_NAME` is a better constant name here, if you want to update that.

PR #2089, opened 2019-02-08
 Change shift+pgup/pgdown to scroll history

### LINKED ISSUE #1989: Add a command to scroll the scrollback buffer so we can bind it to keys
As discussed on IRC:

It would be nice to have a way to bind a 'scroll up' action to keys, so we can simulate the behaviour on other terminals. For example, on termite I can use shift+pgup to see the scrollback, and pgup/down do whatever my shell does (scroll through history). inside vim shift+pgup/down scrolls by page, but I suspect I have something on my to do that, and on tmux it doesn nothing.  

Thanks!

chrisduerr:
The default shift+pgup/pgdown buttons were sending the escape sequences
specified by the official standard, however most terminal emulators like
XTerm, URxvt and VTE make an exception for this special case and instead
scroll the native history buffer.

Both XTerm and URxvt do never send the escapes for Shift+PgUp/PgDown,
however VTE does send them in the alternate screen.

Since Alacritty already supports keybindings based on terminal mode and
the binding to scroll the history is useless when in the alternate
screen buffer, Alacritty is now following VTEs behavior here, allowing
applications in the alt screen (like vim) to handle this escape.

Fixes #1989.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Error/Warning bar doesn't overwrite the terminal anymore
 - Full error/warning messages are displayed
 - Config error messages are automatically removed when the config is fixed
+- Scroll history on Shift+PgUp/PgDown when scrollback history is available
 
 ### Fixed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -444,122 +444,124 @@ alt_send_esc: true
 #   - AppKeypad
 key_bindings:
   # (Windows/Linux only)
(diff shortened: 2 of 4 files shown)


ngortheone:
@chrisduerr  @jwilm 
Is this fix in master? I am on 0.2.9 and I still get 2~2~ when I try to scroll up?

chrisduerr:
I'd recommend trying with the default config using `alacritty --config-file /dev/null` to see if it works for you or not.

If that resolves it there's likely an issue with your config (potentially due to duplicate entries).

ngortheone:
@chrisduerr 
I found the problem. There were old keybindings left in config file that were duplicating same bindings with different values. I removed my config and used a vanilla one and now it is fixed.

Thanks!

Issue #2087, opened 2019-02-07
Inaccurate claim


jcayzac:
Hi,

> Alacritty is the fastest terminal emulator in existence. Using the GPU for rendering enables optimizations that simply aren't possible in other emulators.

This is a bit misleading. [iTerm2 for instance has had GPU rendering since version 3.2.0](https://gitlab.com/gnachman/iterm2/wikis/Metal-Renderer).

nixpulvis:
Maybe we can simply add "in most other emulators."

I don't have strong feelings about this, and I'm not a marketing person really.

chrisduerr:
I think that's the least inaccurate claim made in the README.

`other emulators` here implies other terminal emulators that do not use GPU rendering, otherwise the statement about the GPU rendering would be pretty pointless anyways.

Discussions about reworking the readme a bit have been made before, however I see no rush to jump to action here. It's likely things will get reworded in the future though, taking out the `alpha` is another thing worth considering (at least switch it to beta, it's by far not as unpolished anymore as it was when it was first put there).

PR #2091, opened 2019-02-08
Small Updates to the README

nixpulvis:
Here I've taken the liberty of changing things about as dramatically as I could imagine being reasonable. Please feel free to push back on anything I've changed, or suggest a different set of changes.

- Call Alacritty "beta" tier software (I think it's time).
- Add "most" to the sentence about GPU terminals (others might be catching on)
- Generalize section about performance measurements, mostly removing the claim we're always faster in every way. Latency is an issue we care about, and will be improved
- Small formatting updates

Partially inspired by #2087, among others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,10 @@ Alacritty currently supports macOS, Linux, BSD, and Windows.
 ## About
 
 Alacritty is a terminal emulator with a strong focus on simplicity and
-performance. Bloated features are avoided, and the included features are
-carefully considered. With such a strong focus on performance, you can expect
-Alacritty to be blazing fast, always. Alacritty should also "just work" by
-making sane choices in defaults, but allowing [configuration](#configuration)
-of many aspects of the terminal.
+performance. With such a strong focus on performance, included features are
+carefully considered and you can always expect Alacritty to be blazingly fast.
+Alacritty should also "just work" by making sane choices in defaults, but
+allowing [configuration](#configuration) of many aspects of the terminal.
 
 The software is considered to be at a **beta** level of readiness -- there are
 a few missing features and bugs to be fixed, but it is already used by many as
@@ -150,9 +149,9 @@ feature request, feel free to ask about it. Please just keep in mind Alacritty
 is focused on simplicity and performance, and not all features are in line with
 that goal.
 
-Before posting an issue try to give the existing issues a read over, there's a
-chance someone else has already reported it, and you can subscribe to that
-issue to keep up on the latest developments.
+Before opening a new issue, please check if it has already been reported.
+There's a chance someone else has already reported it, and you can subscribe to
+that issue to keep up on the latest developments.
 
 ## FAQ
 
@@ -165,10 +164,10 @@ case where this isn't true, please report a bug.
 **_Why isn't feature X implemented?_**
 
 Alacritty has many great features, but not every feature from every other
-terminal. This could be for a number of reasons, but sometimes it's just
-because we aren't interested in the added complexity. This means you won't find
-things like tabs or splits (which are best left to a window manager or
-[terminal multiplexer][tmux]) nor niceties like a GUI config editor.
+terminal. This could be for a number of reasons, but sometimes it's just not a
+good fit for Alacritty. This means you won't find things like tabs or splits
+(which are best left to a window manager or [terminal multiplexer][tmux]) nor
+niceties like a GUI config editor.
 
 **_macOS + tmux + vim is slow! I thought this was supposed to be fast!_**
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Alacritty
 
 Alacritty is the fastest terminal emulator in existence. Using the GPU for
 rendering enables optimizations that simply aren't possible without it.
-Alacritty currently supports macOS, Linux, and Windows.
+Alacritty currently supports macOS, Linux, BSD, and Windows.
 
 <p align="center">
   <img width="600"
--- a/README.md
+++ b/README.md
@@ -16,11 +16,12 @@ Alacritty currently supports macOS, Linux, BSD, and Windows.
 
 ## About
 
-Alacritty is focused on simplicity and performance. The performance goal means
-it should be faster than any other terminal emulator available. The simplicity
-goal means that it doesn't have features such as tabs or splits (which can be
-better provided by a window manager or [terminal multiplexer][tmux]) nor
-niceties like a GUI config editor.
+Alacritty is a terminal emulator with a strong focus on simplicity and
+performance. Bloated features are avoided, and the features we do include are
+carefully considered. This means you won't find things like tabs or splits
+(which are best left to a window manager or [terminal multiplexer][tmux]) nor
+niceties like a GUI config editor. We also strive to keep alacritty faster than
+any other terminal available. Performance matters!
 
 The software is considered to be at a **beta** level of readiness -- there are
 a few missing features and bugs to be fixed, but it is already used by many as
--- a/README.md
+++ b/README.md
@@ -19,8 +19,9 @@ Alacritty currently supports macOS, Linux, BSD, and Windows.
 Alacritty is a terminal emulator with a strong focus on simplicity and
 performance. With such a strong focus on performance, included features are
 carefully considered and you can always expect Alacritty to be blazingly fast.
-Alacritty should also "just work" by making sane choices in defaults, but
-allowing [configuration](#configuration) of many aspects of the terminal.
+By making sane choices for defaults, Alacritty requires no additional setup.
+However, it does allow [configuration](#configuration) of many aspects of the
+terminal.
 
 The software is considered to be at a **beta** level of readiness -- there are
 a few missing features and bugs to be fixed, but it is already used by many as
--- a/README.md
+++ b/README.md
@@ -154,8 +154,9 @@ and widgets.
 
 **_Is it really the fastest terminal emulator?_**
 
-In the terminals we've benchmarked against, Alacritty is either faster, *way*
-faster, or at least neutral.
+In the terminals we've [benchmarked](http://github.com/jwilm/vtebench),
+Alacritty is either faster or **way** faster than the others. If you've found a
+case where this isn't true, please report a bug.
 
 **_macOS + tmux + vim is slow! I thought this was supposed to be fast!_**
 
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ have found a better experience using XWayland which can be achieved launching
 Alacritty with the `WAYLAND_DISPLAY` environment variable cleared:
 
 ```sh
-env WAYLAND_DISPLAY=alacritty
+env WAYLAND_DISPLAY="" alacritty
 ```
 
 If you're interested in seeing our Wayland support improve, please head over to
--- a/README.md
+++ b/README.md
@@ -5,11 +5,13 @@ Alacritty
 [![Appveyor build Status](https://ci.appveyor.com/api/projects/status/github/jwilm/alacritty?svg=true)](https://ci.appveyor.com/project/jwilm/alacritty)
 
 Alacritty is the fastest terminal emulator in existence. Using the GPU for
-rendering enables optimizations that simply aren't possible in other emulators.
-Alacritty currently supports macOS, Linux, and Windows.
+rendering enables optimizations that simply aren't possible in most other
+emulators. Alacritty currently supports macOS, Linux, and Windows.
 
 <p align="center">
-  <img width="600" alt="Alacritty running vim inside tmux" src="https://cloud.githubusercontent.com/assets/4285147/21585004/2ebd0288-d06c-11e6-95d3-4a2889dbbd6f.png">
+  <img width="600"
+       alt="Alacritty running vim inside tmux"
+       src="https://cloud.githubusercontent.com/assets/4285147/21585004/2ebd0288-d06c-11e6-95d3-4a2889dbbd6f.png">
 </p>
 
 ## About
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -182,7 +182,7 @@ have found a better experience using XWayland which can be achieved launching
 Alacritty with the `WAYLAND_DISPLAY` environment variable cleared:
 
 ```sh
-env WAYLAND_DISPLAY= alacritty
+env WAYLAND_DISPLAY=alacritty
```
cole-h: Above section says "which can be achieved launching Alacritty with the `WAYLAND_DISPLAY` environment variable cleared", which makes me believe this change (_setting_ the `WAYLAND_DISPLAY` env var as opposed to _clearing_ it) was just a brainfart. If this change is intentional, maybe the above should be reworded.

Also in that same section, maybe "which can be achieved launching Alacritty" -> "which can be achieved **by** launching Alacritty"

README.md
```diff
@@ -20,9 +22,9 @@ goal means that it doesn't have features such as tabs or splits (which can be
 better provided by a window manager or [terminal multiplexer][tmux]) nor
 niceties like a GUI config editor.
 
-The software is considered to be at an **alpha** level of readiness--there are
-missing features and bugs to be fixed, but it is already used by many as a daily
-driver.
+The software is considered to be at a **beta** level of readiness -- there are
+a few missing features and bugs to be fixed, but it is already used by many as
+a daily driver.
```
chrisduerr: While bumping Alacritty's version, I noticed that it's also titled as 'alpha' in the snap file. Should probably be bumped there too.

Might be worth searching through the other files too.

README.md
nixpulvis: @cole-h much thanks for catching this. I've updated the PR to hopefully be a bit more clear, since I honestly forgot `env A= ...` is clearing `A` (not technically unsetting). An empty string should make this more obvious hopefully.

nixpulvis:
I've at least made another pass at this.

Let me know if the updated "About" section is good, or if there's anything better to add/change. Once we're happy with it, I can port it over to `alacritty.man` too.

I've mostly copied your FAQ section, but tweaked the sentence a little to be shorter and flow (maybe) a tiny bit better.

I've also addressed the other review comments.

README.md
```diff
@@ -5,24 +5,27 @@ Alacritty
...
-it should be faster than any other terminal emulator available. The simplicity
-goal means that it doesn't have features such as tabs or splits (which can be
-better provided by a window manager or [terminal multiplexer][tmux]) nor
-niceties like a GUI config editor.
+Alacritty is a terminal emulator with a strong focus on simplicity and
+performance. Bloated features are avoided, and the features we do include are
+carefully considered. This means you won't find things like tabs or splits
+(which are best left to a window manager or [terminal multiplexer][tmux]) nor
+niceties like a GUI config editor. We also strive to keep alacritty faster than
+any other terminal available. Performance matters!
```
chrisduerr: I don't like the way this reads since it kinda feels like it's mostly `Alacritty is not [...]`, rather than `Alacritty is [...]`. While it's important to state what is not the focus of Alacritty, I feel like we should make it clear what the advantages of Alacritty are first.

I'm also not sure if I'd mention specific features like tabs/splits/GUI config editor in the about section, Alacritty is not *about* not having those things. Not having those things is just a result of what Alacritty is about.

I'm a big fan of the first sentence, that already gives a pretty good idea of what Alacritty is. But I think we can tweak the rest a bit more to showcase what Alacritty is for. Then maybe we could put the tabs/splits/config editor in the FAQ section?

README.md
```diff
@@ -146,16 +148,15 @@ On Windows the config file is located at:
 If you run into a problem with Alacritty, please file an issue. If you've got a
 feature request, feel free to ask about it. Keep in mind that Alacritty is very
 much not looking to be a feature-rich terminal emulator with all sorts of bells
-and widgets. It's primarily a cross-platform, blazing fast `tmux` renderer that
-Just Works.
+and widgets.
```
chrisduerr: Even though this is already incorrect in the current version, I believe the correct phrasing here would be `... with all sorts of bells and whistles.`.

README.md
nixpulvis: I agree about moving the bit about tabs and splits into the FAQ, I was frankly just being conservative with my edits here. I'm tempted to let this about section shrink a little as a result. See updated commit incoming for another draft.

README.md
nixpulvis: heh, I was having a little too much fun eh? Widgets are the bane of many over-complex UIs. Regardless noted, I don't like it anyway.

nixpulvis: OK, nevermind I think I've fixed it.

README.md
nixpulvis: Actually decided to try adding a sentence about defaults and configuration. Let me know.

README.md
nixpulvis: Is there any sane way to DRY up all this? I'm fine making manual changes to the `man` entry since it's formatting is a bit of a special case, but I feel like package descriptions like this should be copy pastes (essentially) from the README. Worst case, it's not too hard to just do that, but I've never dealt with a snapcraft file before.

README.md
```diff
@@ -17,11 +17,11 @@ Alacritty currently supports macOS, Linux, BSD, and Windows.
...
-performance. Bloated features are avoided, and the features we do include are
-carefully considered. This means you won't find things like tabs or splits
-(which are best left to a window manager or [terminal multiplexer][tmux]) nor
-niceties like a GUI config editor. We also strive to keep alacritty faster than
-any other terminal available. Performance matters!
+performance. Bloated features are avoided, and the included features are
+carefully considered. With such a strong focus on performance, you can expect
+Alacritty to be blazing fast, always. Alacritty should also "just work" by
+making sane choices in defaults, but allowing [configuration](#configuration)
+of many aspects of the terminal.
```
chrisduerr: This might just be me, but the words `Bloated features are avoided` sound a bit judgemental torwards other terminals and in general the word `Bloated` feels very opinionated.

```
With such a strong focus on performance, included features are carefully considered and you can expect Alacritty to always be blazingly fast.
```

I think the content of the last sentence is great and a welcome addition, however I'm not sure if I'd put something like `"just work"` in a manpage (in the readme it would be fine I think, but reusability would be nice).

Maybe something like this?
```
By choosing sane defaults, Alacritty should also be pleasant to use out of the box, while still allowing [configuration](#configuration) of many aspects of the terminal.
```

I'm not sure if this is an improvement, so let me know what you think please.

README.md
```diff
@@ -146,9 +146,13 @@ On Windows the config file is located at:
...
-feature request, feel free to ask about it. Keep in mind that Alacritty is very
-much not looking to be a feature-rich terminal emulator with all sorts of bells
-and widgets.
+feature request, feel free to ask about it. Please just keep in mind Alacritty
+is focused on simplicity and performance, and not all features are in line with
+that goal.
+
+Before posting an issue try to give the existing issues a read over, there's a
+chance someone else has already reported it, and you can subscribe to that
+issue to keep up on the latest developments.
```
chrisduerr: `try to give the existing issues a read over` sounds a bit awkward I think.

I've got another proposal, though I'm not sure if it's not too 'forceful'. I'd rather have a duplicate than an issue that hasn't been reported.

```
Before opening a new issue, please make sure if it has already been reported. You can subscribe to existing issues to stay up to date with their progress.
```

README.md
```diff
@@ -158,6 +162,14 @@ In the terminals we've [benchmarked](http://github.com/jwilm/vtebench),
 Alacritty is either faster or **way** faster than the others. If you've found a
 case where this isn't true, please report a bug.
 
+**_Why isn't feature X implemented?_**
+
+Alacritty has many great features, but not every feature from every other
+terminal. This could be for a number of reasons, but sometimes it's just
+because we aren't interested in the added complexity. This means you won't find
+things like tabs or splits (which are best left to a window manager or
+[terminal multiplexer][tmux]) nor niceties like a GUI config editor.
```
chrisduerr: `we aren't interested in the added complexity` feels a bit dismissive I think. Putting it in a better light might be a good idea.

Something like this is a bit marketingy and kinda downplays the issue, but I think it might be more welcoming:

```
but sometimes it's just because the feature isn't a good fit for Alacritty.
```

README.md
chrisduerr: I'm completely unfamiliar with snapcraft and don't even know if this file is actually used by the snap package, since no Alacritty maintainer also maintains the snapcraft file.

It might be interesting to get a hold on whoever maintains that and put that effort on them? I feel like this is something that should be done by the packager.

If we do keep it in the repo, I don't think there's any good way around it though. Though it might be possible to automatically generate the manpage from the CLI --help and readme.

README.md
cole-h: `please make sure if it has already been reported` sounds a little awkward as well.

I think this flows better, IMHO:
`Before opening a new issue, check if it has already been reported. <...>`

chrisduerr: Throw another `please` in front of the `check`  and it sounds awesome to me!

README.md
cole-h: Slight nit on the flow of your first proposition (moved `always` before `expect`):
```
With such a strong focus on performance, included features are carefully considered and you can always expect Alacritty to be blazingly fast.
```

I like your second proposition regarding the configuration wording,  though.

chrisduerr: I like that!

README.md
nixpulvis: Yep, that was my plan.

README.md
nixpulvis: Yep, that basically fixes the problem I had reading it. I'll update these all in a bit.

nixpulvis:
OK, I've pushed what I expect to be the last revision of the README. I've given up on trying to make the "About" section be copy-paste-able to the man page, since I feel like they are different enough. For example I'd like to put a bit about platform support into the man page, but I don't feel like it makes sense to have it twice in the README at the very top (it's already in the short paragraph above the image).

I've taken a first swing at the man page as well.

chrisduerr:
> I'd like to put a bit about platform support into the man page

I don't think that's necessary. The manpage should give info on usage mainly, so I'd never expect information about cross-platform support to be in there. Especially since manpages don't even exist on all platforms.

> I've given up on trying to make the "About" section be copy-paste-able to the man page

I think it should be just fine to copy the README over to the manpage. I don't think we need to explicitly state the part of it being beta in the manpage. Just copying the first paragraph should be fine.

Also note that the about section of the manpage is not the correct place to tell people where bugs should be reported. We already have a dedicate `BUGS` section, which is the common way of handling this.

README.md
chrisduerr: Any feedback on my proposal to change the "just works" section @nixpulvis? I think it's a bit colloquial.

nixpulvis:
OK, fixed the man page, and tweaked the colloquialism. Now I think we just need to figure out what to do about the snapcraft file.

cole-h:
Just my opinion, but couldn't we continue doing what's already being done? Copy the intro and about sections to the `snapcraft.yml` almost word-for-word and match that up with your current changes? Just a quick bit of research, but the `description` key is mandatory and should contain a `more in-depth look at what your snap does and who may find it most useful` (from: https://docs.snapcraft.io/snapcraft-yaml-reference/4276). IMHO, the fixed-up intro and about sections do a good job of that. 

We could also ping Chris MacNaughton -- who is listed on [Alacritty's snapcraft page](https://snapcraft.io/alacritty) -- for his thoughts.

(I'm unsure of the etiquette on mentioning users, so I'll leave that up to one of you, if you decide it is necessary)

chrisduerr:
@macnaughton are you actually using the snapcraft.yml in the Alacritty repos for https://snapcraft.io/alacritty? Considering how no other build files for distributions are in Alacritty's repos, it might be nice to mantain than separately?

macnaughton:
@chrisduerr Sorry but I'm not the Chris Macnaughton you're looking for!

chrisduerr:
@macnaughton Oh, sorry. But I think I found the right one!

@ChrisMacNaughton looks pretty good since he seems to be a Canonical employee. There can't be *that* many Chris McNaughtons on this world!

ChrisMacNaughton:
@chrisduerr it is being used to maintain the snap - the desire is to actually get alacritty building (into the edge channel) on each commit, which is much harder to do in an out-of-tree mechanism. Given how much progress has been made on that goal, it may be worth re-evaluating that as a goal.

From a practical point of view, there are actually other distribution files in-tree (see `./debian`) but it's not necessary to maintain the snapcraft files in-tree; I suggest that maintenance in-tree would depend on deciding if the project would like to have automatic snap builds on commit, or if it's better left to a highly asynchronous release process dependent on my schedule?

chrisduerr:
@ChrisMacNaughton Thanks for the swift reply.

As far as I'm concerned, that's more than enough reason to keep things the way they are. At least for now.

> the desire is to actually get alacritty building (into the edge channel) on each commit

I'd hope this would just be an alternative to the latest release though. Since Alacritty's master is not intended to be the recommended version to use for endusers and that would likely cause additional work for Alacritty and provide users with a worse experience.

ChrisMacNaughton:
> 
> > the desire is to actually get alacritty building (into the edge channel) on each commit
> 
> I'd hope this would just be an alternative to the latest release though. Since Alacritty's master is not intended to be the recommended version to use for endusers and that would likely cause additional work for Alacritty and provide users with a worse experience.

The idea would be that a user could `snap install --edge alacritty` and be running the latest, but that `snap install alacritty` would give them the latest release, and that the release process to stable would probably be manual, and could step through beta -> candidate -> stable, rather than just going from tip.

PR #2099, opened 2019-02-11
Add Control+Back, Control+Delete bindings to default config

jplatte:
Bound to delete-to-word-begin, delete-to-word-end as for example in VS Code.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -481,9 +481,11 @@ key_bindings:
   - { key: PageDown,                chars: "\x1b[6~"                         }
   - { key: Tab,      mods: Shift,   chars: "\x1b[Z"                          }
   - { key: Back,                    chars: "\x7f"                            }
+  - { key: Back,     mods: Control, chars: "\x17"                            }
   - { key: Back,     mods: Alt,     chars: "\x1b\x7f"                        }
   - { key: Insert,                  chars: "\x1b[2~"                         }
   - { key: Delete,                  chars: "\x1b[3~"                         }
+  - { key: Delete,   mods: Control, chars: "\x1bD"                           }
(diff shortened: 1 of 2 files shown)


alacritty.yml
```diff
@@ -481,9 +481,11 @@ key_bindings:
   - { key: PageDown,                chars: "\x1b[6~"                         }
   - { key: Tab,      mods: Shift,   chars: "\x1b[Z"                          }
   - { key: Back,                    chars: "\x7f"                            }
+  - { key: Back,     mods: Control, chars: "\x17"                            }
```
chrisduerr: The `Back` control sequence should be `^H`.

alacritty.yml
```diff
@@ -481,9 +481,11 @@ key_bindings:
...
   - { key: Back,     mods: Alt,     chars: "\x1b\x7f"                        }
   - { key: Insert,                  chars: "\x1b[2~"                         }
   - { key: Delete,                  chars: "\x1b[3~"                         }
+  - { key: Delete,   mods: Control, chars: "\x1bD"                           }
```
chrisduerr: The delete escape sequence should be `\x1b[3;5~`. Additionally an action for Shift+Delete should be added (`\x1b[3;2~`).

jplatte:
I thought some terminals actually implemented this behaviour apart form VS Code's internal terminal, but I just checked again and you are right. I will just have to keep this in my personal config then.

chrisduerr:
@jplatte This is usually done in the .bashrc/.bash_profile (or whatever configuration your shell might be using). It's possible that VS Code's internal terminal comes with a non-default config.

jplatte:
@chrisduerr: VS Code's internal terminal sends the same escape codes that I added in this PR. (checked with `showkey - a`)

PR #2101, opened 2019-02-11
 Fix font spacing error 

### LINKED ISSUE #2098: Display issue with 2 monitors on MacOS after update to 0.2.8
MacOS 10.14.3.

After the update from 0.2.7 to 0.2.8 Alacritty has a problem with different monitors and different display resolutions. In my case I'm using the default monitor of a MBP (1680x1050) + an external 24" Monitor (1200x1920 correct value).

Using just one monitor is working perfectly fine. 

Starting Alacritty on the MBP looks good: 
screenshot 2019-02-11 at 11 40 16

Moving Alacritty to the external monitor: 
screenshot 2019-02-11 at 11 40 36

And moving it back to the MBP one:
screenshot 2019-02-11 at 11 40 48

chrisduerr:
Fixes #2098.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


nixpulvis:
I can confirm this fixes the issue.

There may still be a little room for improvement in the way we scale font sizes and cell sizes, but that's not critical for this fix at all, and should be considered more closely at a later time.

Unless anyone else has feelings I think it would make sense to cut another release ASAP since I suspect this will affect a large number of users.

PR #2110, opened 2019-02-13
Update fish completion installation

chrisduerr:


nixpulvis:
Fish says NO to backwards compatibility. LOL. Anyway? Thanks for doing this, and removing the duplicate install wiki page.

Issue #779, opened 2017-09-08
EHN: inherited configuration files.


musically-ut:
I use alacritty on multiple platforms such that the dotfiles between them is kept in sync (using softlinks and Dropbox, though that is not important).

However, I would like to have slightly tweaked configuration on each system (e.g. larger font sizes on one system, `use_thin_strokes` on one system, light/dark color themes, etc.).

Right now, the config is read only from one single file:

https://github.com/jwilm/alacritty/blob/8990aa4568ea8a7481f827512b7487be4d39332a/src/main.rs#L65-L75

Would there be interest in converting `config_path` to an array and such that all the configs files are read in the said order:

    /// 1. $XDG_CONFIG_HOME/alacritty/alacritty.yml
    /// 2. $XDG_CONFIG_HOME/alacritty.yml
    /// 3. $HOME/.config/alacritty/alacritty.yml
    /// 4. $HOME/.alacritty.yml

and the configurations read from each is _merged_ with preferences being overwritten by files read later.

This would allow me to keep, say, `$HOME/.config/alacritty/alacritty.yml` synced between the systems while keeping local configuration in `$HOME/.alacritty.yml`.

Alternatively, we can add a file to the list, i.e. `$HOME/.alacritty.local.yml`, which will be exclusively used to override the preferences read from other files.

lukaslueg:
Related to #169

mathstuf:
I'd prefer there be another filename that is used rather than using the same one in different locations to be the trigger. One of the benefits of XDG paths is the decluttering of `$HOME`.

2m:
One more advantage of merging config files is the easier management of config as alacritty evolves. This would allow having only the custom settings in a config file, and all of the default ones would be merged in from the default config.

nixpulvis:
Reducing the filesize of `.config/alacritty/alacritty.yml` is a nice win indeed.

Sidenote: it drives me crazy that it's named `alacritty.yml` and not `config` or `config.yml`, even though it hardly matters.

mathstuf:
Alternatively, a directory could be read instead of specific paths. For example, anything matching `$XDG_CONFIG_HOME/alacritty/conf.d/*.yml` is read in ASCII-betical order and later files override previous files.

2m:
I thought about merging multiple yml files as well, but then there would still be no way to override mouse and keyboard bindings without modifying the original configuration file. Here is an example mouse binding:
```yml
mouse_bindings:
[...]

PR #2113, opened 2019-02-14
Add inherited configuration file support.

norcalli:
Adds two methods for including another yaml file inside of the config
file that is referenced by alacritty: `include PATH` and `include!  PATH`.

`include!` will throw an error if the path is not found while `include`
will only throw a warning.

It also substitutes environment variables which are referenced in the
file path to allow for a greater degree of customization via scripting.

Addresses #779

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -1904,6 +1904,67 @@ impl Config {
             return Ok(Config::default());
         }
 
+        contents = {
+            let mut new_contents = String::new();
+
+            let env_vars = env::vars().collect();
+
+            struct IncludeFile {
+                file: PathBuf,
+                required: bool,
+            }
+
+            impl IncludeFile {
+                fn parse_line(line: &str) -> Option<Self> {
+                    if line.starts_with("include ") {
+                        Some(IncludeFile {
+                            file: PathBuf::from(&line["include ".len()..]),
+                            required: false,
+                        })
+                    } else if line.starts_with("include! ") {
+                        Some(IncludeFile {
+                            file: PathBuf::from(&line["include! ".len()..]),
+                            required: true,
+                        })
+                    } else {
+                        None
+                    }
+                }
+
+                fn substitute(line: &str, vars: &HashMap<String, String>) -> String {
+                    let mut line = String::from(line);
+                    for (key, value) in vars.iter() {
+                        line = line.replace(&format!("${}", key), value);
+                    }
+                    line
+                }
+            }
+
+            contents.lines()
+                .for_each(|line| {
+                    if let Some(include) = IncludeFile::parse_line(&IncludeFile::substitute(line, &env_vars)) {
+                        if include.file.exists() {
+                            if let Ok(mut file) = File::open(&include.file) {
+                                if let Err(err) = file.read_to_string(&mut new_contents) {
+                                    warn!("Failed to include config file {:?}: {}", &include.file, err);
+                                }
+                            }
+                        } else if include.required {
+                            error!("Include file does not exist {:?}!", &include.file);
+                        } else {
+                            warn!("Include file does not exist {:?}!", &include.file);
+                        }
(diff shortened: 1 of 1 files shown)


norcalli:
I made some decisions in this which should be discussed, but considering the small scope of the PR, it's fairly easy to change.

Open questions in my mind are:
- syntax of the include statement
  - currently, I've gone with `include` and `include!`
- whether to validate the files being include
  - currently I leave that to the final parsing
- whether to have a separate "required" mode
  - currently, `include!` notes a stricter mode in case a path is not found
- the operation mode of required vs non-required mode
  - currently, non-required mode will log a `warn!` vs an `error!`
  - an alternative that I may be favorable towards is to be silent for non-required mode, and to have required mode log error.

I believe that substituting environment variables is vital to making this useful for a situation like `hostname` based configuration or via scripting.

norcalli:
Just for fun, here is how I'm currently using it:

```yml
# Configuration for Alacritty, the GPU enhanced terminal emulator

include! $HOME/.config/alacritty/base.yml
include! $HOME/.config/alacritty/hack-font.yml
# include! $HOME/.config/alacritty/source-code-pro-font.yml

# env.BASE16_THEME: default-dark
# include! $HOME/.config/alacritty/colors/base16-default-dark-256.yml

include $HOME/.config/alacritty/colors/base16-$BASE16_THEME-256.yml
```

chrisduerr:
Thanks for the proposal.

I don't think the approach taken here is a great fit for Alacritty. There are several issues which are directly connected to this solution which I'm not comfortable with putting on users.

First of all this extension is proprietary and completely undocumented. While we could provide documentation for Alacritty, people just expecting normal yaml files would be completely now to this without the ability to use it.

It also doesn't follow any yaml syntax, which makes this approach actually break perfectly valid configuration files since it is possible to have `include ` at the start of a line without actually wanting to include something (multiline strings). I'm sure you can understand that this is not acceptable. We can't just break yaml with proprietary extensions nobody knows about. That would likely cause a lot of problems.

Lastly there's the issue with live config reload. This is something that should just work for all settings that are documented to work with it, without weird issues. But if we watch just the one file we load the configuration from, we won't actually know when any included file has been reloaded.

As far as I can tell, at this point this doesn't actually provide any benefit over what is already possible with shell scripts, so I think those standardized and documented tools should be preferred. Adding two config files together should be pretty simple using `alacritty --config-file <( cat file1 file2 )`. This approach wouldn't allow live config reload, though it matches this PR for that. But at the same time it has none of the other issues.

While I appreciate the attempt at solving this, I don't think this is something that can just be quickly rushed in a PR. A sufficient approach should be thoroughly discussed and involve multiple parties which can bring up their different usecases. So I'd like to continue this discussion in #779, if you have any comments on how to resolve some of the issues with your approach, feel free to bring them up there.

norcalli:
You are absolutely correct. I was up for 24 hours (so not the best place for thinking) when I started the PR and it turned out to be way more complicated than I expected in terms of features I'd have to support. I've abandoned the idea for doing it now, and I'll just use a preprocessor. I think if I'll contribute anything, it would be turning alacritty into a library so that features like this that I may want I can maintain in a private fork.

nixpulvis:
@norcalli I think there is valid room for improvement in our configuration file, it just need to be carefully thought out and possibly run through a few iterations of trial and error. I wouldn't expect the first idea to be perfect, and we really wouldn't want to get this wrong (changing configs once may already be too much).

I can't speak for everyone, but I personally would like to move away from YAML altogether. But I can't honestly recommend that until I have a better option to offer. That means more than just proposing TOML (for example) since we'd want to have a clear idea of how the config would look under the new format, and @chrisduerr has rightfully raised concerns about how things like keybindings will look in TOML.

I think it would be a bit of a shame to maintain a fork just over configuration issues, but I don't know yet how to solve the concerns of everyone.

PR #1367, opened 2018-06-11
Static display notification on SIGCHLD

semaj:
Ensures that the window processor wakes up upon SIGCHLD.
Otherwise, after `exit` or `logout`, the user had to trigger
a window event before alacritty closed.

Fixes #1313.
Fixes #1276.
Fixes #915.

This may not be the best solution. It's certainly not pretty. Also, my editor is forcing rustfmt on the codebase - let me know if that's annoying. 

Relevant parts: `main.rs` :  110 - 114, `tty.rs` : 43 - 64

PR #1589, opened 2018-09-26
[wip] Sodiumjoe/607 multiple windows

sodiumjoe:
This is working off of `mkeeler:glutin-0.17-upgrade`, so just the last two commits are mine. You can check out just the relevant diff [here](https://github.com/mkeeler/alacritty/compare/glutin-0.17-upgrade...sodiumjoe:sodiumjoe/607-multiple-windows?expand=1).

I've basically been just trying to factor things (in a cargo-culty way) so that it's possible to create multiple windows without changing any logic. This compiles and runs, and will even create two windows (if hardcoded in).

But I seem to be missing something, because even with just one window, there are some issues.
1. sometimes the terminal doesn't update
2. certain keys are triggered twice (like backspace)

I haven't tried to get the second window working in earnest yet, but I'll describe what happens in case it helps diagnose the single window case. One window seems to work just like the single-window case, but the second window at first renders with just the blank background color, and on any interaction becomes transparent. It seems to still take window events, like resize events and keyboard input, but it doesn't display anything. With both windows attached to a tmux session, I can see keyboard input from the second window take effect in the first window, but not immediately, only after interacting with the second window.

I'd like to get the single window case working properly before trying to get multiple windows working and I'm kinda stuck, so I'm hoping someone else can take a look and see if I missed something obvious or point me at some better debugging methods. I've tried with `--print-events`, and the events there appear to be totally normal.

---
This change is [Reviewable](https://reviewable.io/reviews/jwilm/alacritty/1589)

PR #2114, opened 2019-02-15
Fix lingering window after child process exit on unix

### LINKED ISSUE #915: Quitting Vim after "alacritty -e vim" requires additional keystroke to close the terminal
If I run 

``alacritty -e vim``

And then quit right away with

``:q``

There remains a dangling Alacritty window with no text and just a cursor in the top left corner.
It persists until I move the mouse or press a key, which is a little bit annoying.
I've tested the same command with ``urxvt``, ``terminator``, ``st`` and behaviour of all seems to be the same in that they all close the terminal once ``vim`` quits.

For the record, I am writing this from Arch Linux with an i3 window manager. My server is ``X11``.

Other than that, thank you for an amazing terminal... I've been struggling to find one that takes advantage of my 144 Hz monitor, especially since I am a former FPS player and can immediately tell whether something responds to my keystrokes in 16ms or 7ms!

### LINKED ISSUE #1276: alacritty processes continue running after closing
On a fully up to date Arch Linux set up on X11 with NVIDIA proprietary driver's (396.18) and the composite extension disabled, transparency and vsync disabled (via .yml), alacritty processes continue to run and don't stop even after closing the terminal windows. My CPU cores will rise to 100% and there will be many alacritty processes running, even if you shut down the X.Org display server. I have to manually kill each alacritty process with htop for my CPU to stop maxing out at 100%.

### LINKED ISSUE #1313: Alacritty doesn't close until the first event after exiting shell
Which operating system does the issue occur on? Linux, Ubuntu 18.04

If on linux, are you using X11 or Wayland? Running Wayland with XWayland (`env WAYLAND_DISPLAY= alacritty`)

When I leave the shell with ^d or "exit", alacritty will stay alive until I do the next thing (move my mouse, click, type a character, or defocus). I'd expect it to die immediately after exiting the shell.

Screen recording of the issue: https://webmshare.com/DdYGG

glxinfo:

```console
$ glxinfo | grep OpenGL                           
OpenGL vendor string: Intel Open Source Technology Center
OpenGL renderer string: Mesa DRI Intel(R) HD Graphics 620 (Kaby Lake GT2) 
OpenGL core profile version string: 4.5 (Core Profile) Mesa 18.0.0-rc5
OpenGL core profile shading language version string: 4.50
OpenGL core profile context flags: (none)
OpenGL core profile profile mask: core profile
OpenGL core profile extensions:
OpenGL version string: 3.0 Mesa 18.0.0-rc5
OpenGL shading language version string: 1.30
OpenGL context flags: (none)
OpenGL extensions:
OpenGL ES profile version string: OpenGL ES 3.2 Mesa 18.0.0-rc5
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.20
OpenGL ES profile extensions:
```

rkanati:
This is a crack at handing `SIGCHLD` (and conceivably others) safely and correctly, taking a different approach to #1367 by using the well-known "self-pipe trick". This should allay concerns about non- _async-signal-safe_ functions in signal handlers.

The signal handler writes received `SIG*` numbers into a pipe, which wakes up the existing I/O event loop. The loop then delegates handling back to the `Pty`, which reads the pipe, and reacts appropriately.

I'm not 100% certain on the finer points of this design, though, so I'd appreciate some feedback.

- Fixes #915
- Fixes #1276
- Fixes #1313

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -176,7 +176,8 @@ impl<'a, N: Notify + 'a> input::ActionContext for ActionContext<'a, N> {
             let proc_prefix = "";
             #[cfg(target_os = "freebsd")]
             let proc_prefix = "/compat/linux";
-            if let Ok(path) = fs::read_link(format!("{}/proc/{}/cwd", proc_prefix, unsafe { tty::PID })) {
+            let link_path = format!("{}/proc/{}/cwd", proc_prefix, tty::child_pid());
+            if let Ok(path) = fs::read_link(link_path) {
                 vec!["--working-directory".into(), path]
             } else {
                 Vec::new()
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -34,7 +34,7 @@ pub enum Msg {
 ///
 /// Handles all the pty I/O and runs the pty parser which updates terminal
 /// state.
-pub struct EventLoop<T: tty::EventedReadWrite> {
+pub struct EventLoop<T: tty::EventedPty> {
     poll: mio::Poll,
     pty: T,
     rx: Receiver<Msg>,
@@ -162,7 +162,7 @@ impl Writing {
 
 impl<T> EventLoop<T>
     where
-        T: tty::EventedReadWrite + Send + 'static,
+        T: tty::EventedPty + Send + 'static,
 {
     /// Create a new event loop
     pub fn new(
@@ -338,18 +338,19 @@ impl<T> EventLoop<T>
             let mut state = state.unwrap_or_else(Default::default);
             let mut buf = [0u8; 0x1000];
 
+            let mut tokens = (0..).map(Into::into);
+
             let poll_opts = PollOpt::edge() | PollOpt::oneshot();
 
-            const CHANNEL_TOKEN: mio::Token = mio::Token(0);
-            let mut tokens = (1..).map(Into::into);
-
+            let channel_token = tokens.next().unwrap();
             self.poll
-                .register(&self.rx, CHANNEL_TOKEN, Ready::readable(), poll_opts)
+                .register(&self.rx, channel_token, Ready::readable(), poll_opts)
                 .unwrap();
 
             // Register TTY through EventedRW interface
             self.pty
-                .register(&self.poll, &mut tokens, Ready::readable(), poll_opts).unwrap();
+                .register(&self.poll, &mut tokens, Ready::readable(), poll_opts)
+                .unwrap();
 
             let mut events = Events::with_capacity(1024);
 
@@ -369,41 +370,50 @@ impl<T> EventLoop<T>
 
                 for event in events.iter() {
                     match event.token() {
-                        CHANNEL_TOKEN => if !self.channel_event(CHANNEL_TOKEN, &mut state) {
-                            break 'event_loop;
+                        token if token == channel_token => {
+                            if !self.channel_event(channel_token, &mut state) {
+                                break 'event_loop;
+                            }
                         },
+
+                        #[cfg(unix)]
+                        token if token == self.pty.child_event_token() => {
+                            if let Some(tty::ChildEvent::Exited) = self.pty.next_child_event() {
+                                self.terminal.lock().exit();
+                                self.display.notify();
+                                break 'event_loop;
+                            }
+                        },
+
                         token if token == self.pty.read_token() || token == self.pty.write_token() => {
-                            #[cfg(unix)]
-                                {
-                                    if UnixReady::from(event.readiness()).is_hup() {
-                                        break 'event_loop;
-                                    }
+                            #[cfg(unix)] {
+                                if UnixReady::from(event.readiness()).is_hup() {
+                                    // don't try to do I/O on a dead PTY
+                                    continue;
                                 }
+                            }
+
                             if event.readiness().is_readable() {
-                                if let Err(err) = self.pty_read(&mut state, &mut buf, pipe.as_mut())
-                                    {
-                                        error!(
-                                            "[{}:{}] Event loop exiting due to error: {}",
-                                            file!(),
-                                            line!(),
-                                            err
-                                        );
-                                        break 'event_loop;
+                                if let Err(e) = self.pty_read(&mut state, &mut buf, pipe.as_mut()) {
+                                    #[cfg(target_os = "linux")] {
+                                        // On Linux, a `read` on the master side of a PTY can fail
+                                        // with `EIO` if the client side hangs up.  In that case,
+                                        // just loop back round for the inevitable `Exited` event.
+                                        // This sucks, but checking the process is either racy or
+                                        // blocking.
+                                        if e.kind() == ErrorKind::Other {
+                                            continue;
+                                        }
                                     }
 
-                                if crate::tty::process_should_exit() {
+                                    error!("Error reading from PTY in event loop: {}", e);
                                     break 'event_loop;
                                 }
                             }
 
                             if event.readiness().is_writable() {
-                                if let Err(err) = self.pty_write(&mut state) {
-                                    error!(
-                                        "[{}:{}] Event loop exiting due to error: {}",
-                                        file!(),
-                                        line!(),
-                                        err
-                                    );
+                                if let Err(e) = self.pty_write(&mut state) {
+                                    error!("Error writing to PTY in event loop: {}", e);
                                     break 'event_loop;
                                 }
                             }
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -160,9 +160,6 @@ impl Writing {
     }
 }
 
-/// `mio::Token` for the event loop channel
-const CHANNEL: mio::Token = mio::Token(0);
-
 impl<T> EventLoop<T>
     where
         T: tty::EventedReadWrite + Send + 'static,
@@ -217,13 +214,13 @@ impl<T> EventLoop<T>
 
     // Returns a `bool` indicating whether or not the event loop should continue running
     #[inline]
-    fn channel_event(&mut self, state: &mut State) -> bool {
+    fn channel_event(&mut self, token: mio::Token, state: &mut State) -> bool {
         if self.drain_recv_channel(state).is_shutdown() {
             return false;
         }
 
         self.poll
-            .reregister(&self.rx, CHANNEL, Ready::readable(), PollOpt::edge() | PollOpt::oneshot())
+            .reregister(&self.rx, token, Ready::readable(), PollOpt::edge() | PollOpt::oneshot())
             .unwrap();
 
         true
@@ -343,15 +340,16 @@ impl<T> EventLoop<T>
 
             let poll_opts = PollOpt::edge() | PollOpt::oneshot();
 
-            let tokens = [1, 2];
+            const CHANNEL_TOKEN: mio::Token = mio::Token(0);
+            let mut tokens = (1..).map(Into::into);
 
             self.poll
-                .register(&self.rx, CHANNEL, Ready::readable(), poll_opts)
+                .register(&self.rx, CHANNEL_TOKEN, Ready::readable(), poll_opts)
                 .unwrap();
 
             // Register TTY through EventedRW interface
             self.pty
-                .register(&self.poll, &mut tokens.iter(), Ready::readable(), poll_opts).unwrap();
+                .register(&self.poll, &mut tokens, Ready::readable(), poll_opts).unwrap();
 
             let mut events = Events::with_capacity(1024);
 
@@ -371,7 +369,7 @@ impl<T> EventLoop<T>
 
                 for event in events.iter() {
                     match event.token() {
-                        CHANNEL => if !self.channel_event(&mut state) {
+                        CHANNEL_TOKEN => if !self.channel_event(CHANNEL_TOKEN, &mut state) {
                             break 'event_loop;
                         },
                         token if token == self.pty.read_token() || token == self.pty.write_token() => {
--- a/src/term/mod.rs
+++ b/src/term/mod.rs
@@ -34,6 +34,8 @@ use crate::url::UrlParser;
 use crate::message_bar::MessageBuffer;
 use crate::term::color::Rgb;
 use crate::term::cell::{LineLength, Cell};
+
+#[cfg(windows)]
 use crate::tty;
 
 pub mod cell;
--- a/src/term/mod.rs
+++ b/src/term/mod.rs
@@ -1336,7 +1336,7 @@ impl Term {
 
     #[inline]
     pub fn should_exit(&self) -> bool {
-        tty::process_should_exit() || self.should_exit
+        self.should_exit
     }
 }
 
--- a/src/tty/mod.rs
+++ b/src/tty/mod.rs
@@ -40,7 +40,7 @@ pub trait EventedReadWrite {
     fn register(
         &mut self,
         _: &mio::Poll,
-        _: &mut dyn Iterator<Item = &usize>,
+        _: &mut dyn Iterator<Item = mio::Token>,
         _: mio::Ready,
         _: mio::PollOpt,
     ) -> io::Result<()>;
--- a/src/tty/mod.rs
+++ b/src/tty/mod.rs
@@ -53,6 +53,29 @@ pub trait EventedReadWrite {
     fn write_token(&self) -> mio::Token;
 }
 
+/// Events concerning TTY child processes
+#[derive(PartialEq)]
+pub enum ChildEvent {
+    /// Indicates the child has exited
+    Exited
+}
+
+/// A pseudoterminal (or PTY)
+///
+/// This is a refinement of EventedReadWrite that also provides a channel through which we can be
+/// notified if the PTY child process does something we care about (other than writing to the TTY).
+/// In particular, this allows for race-free child exit notification on UNIX (cf. `SIGCHLD`).
+pub trait EventedPty : EventedReadWrite {
+    #[cfg(unix)]
+    fn child_event_token(&self) -> mio::Token;
+
+    /// Tries to retrieve an event
+    ///
+    /// Returns `Some(event)` on success, or `None` if there are no events to retrieve.
+    #[cfg(unix)]
+    fn next_child_event(&mut self) -> Option<ChildEvent>;
+}
+
 // Setup environment variables
 pub fn setup_env(config: &Config) {
     // Default to 'alacritty' terminfo if it is available, otherwise
--- a/src/tty/unix.rs
+++ b/src/tty/unix.rs
@@ -327,11 +327,11 @@ impl EventedReadWrite for Pty {
     fn register(
         &mut self,
         poll: &mio::Poll,
-        token: &mut dyn Iterator<Item = &usize>,
+        token: &mut dyn Iterator<Item = mio::Token>,
         interest: mio::Ready,
         poll_opts: mio::PollOpt,
     ) -> io::Result<()> {
-        self.token = (*token.next().unwrap()).into();
+        self.token = token.next().unwrap();
         poll.register(
             &EventedFd(&self.fd.as_raw_fd()),
             self.token,
--- a/src/tty/unix.rs
+++ b/src/tty/unix.rs
@@ -23,17 +23,15 @@ use crate::cli::Options;
 use mio;
 
 use libc::{self, c_int, pid_t, winsize, SIGCHLD, TIOCSCTTY, WNOHANG};
+use nix::pty::openpty;
 
-use std::os::unix::io::{FromRawFd};
+use std::os::unix::{process::CommandExt, io::{FromRawFd, AsRawFd, RawFd}};
 use std::fs::File;
-use std::os::unix::process::CommandExt;
 use std::process::{Command, Stdio};
 use std::ffi::CStr;
 use std::ptr;
 use mio::unix::EventedFd;
 use std::io;
-use std::os::unix::io::AsRawFd;
-
 
 /// Process ID of child process
 ///
@@ -71,50 +69,15 @@ fn errno() -> c_int {
 }
 
 /// Get raw fds for master/slave ends of a new pty
-#[cfg(target_os = "linux")]
-fn openpty(rows: u8, cols: u8) -> (c_int, c_int) {
-    let mut master: c_int = 0;
-    let mut slave: c_int = 0;
+fn make_pty(size: winsize) -> (RawFd, RawFd) {
+    let mut win_size = size;
+    win_size.ws_xpixel = 0;
+    win_size.ws_ypixel = 0;
 
-    let win = winsize {
-        ws_row: libc::c_ushort::from(rows),
-        ws_col: libc::c_ushort::from(cols),
-        ws_xpixel: 0,
-        ws_ypixel: 0,
-    };
+    let ends = openpty(Some(&win_size), None)
+        .expect("openpty failed");
 
-    let res = unsafe {
-        libc::openpty(&mut master, &mut slave, ptr::null_mut(), ptr::null(), &win)
-    };
-
-    if res < 0 {
-        die!("openpty failed");
-    }
-
-    (master, slave)
-}
-
-#[cfg(any(target_os = "macos",target_os = "freebsd",target_os = "openbsd"))]
-fn openpty(rows: u8, cols: u8) -> (c_int, c_int) {
-    let mut master: c_int = 0;
-    let mut slave: c_int = 0;
-
-    let mut win = winsize {
-        ws_row: libc::c_ushort::from(rows),
-        ws_col: libc::c_ushort::from(cols),
-        ws_xpixel: 0,
-        ws_ypixel: 0,
-    };
-
-    let res = unsafe {
-        libc::openpty(&mut master, &mut slave, ptr::null_mut(), ptr::null_mut(), &mut win)
-    };
-
-    if res < 0 {
-        die!("openpty failed");
-    }
-
-    (master, slave)
+    (ends.master, ends.slave)
 }
 
 /// Really only needed on BSD, but should be fine elsewhere
@@ -214,11 +177,11 @@ pub fn new<T: ToWinsize>(
     size: &T,
     window_id: Option<usize>,
 ) -> Pty {
-    let win = size.to_winsize();
+    let win_size = size.to_winsize();
     let mut buf = [0; 1024];
     let pw = get_pw_entry(&mut buf);
 
-    let (master, slave) = openpty(win.ws_row as _, win.ws_col as _);
+    let (master, slave) = make_pty(win_size);
 
     let default_shell = if cfg!(target_os = "macos") {
         let shell_name = pw.shell.rsplit('/').next().unwrap();
--- a/src/tty/unix.rs
+++ b/src/tty/unix.rs
@@ -24,7 +24,7 @@ use mio;
 
 use libc::{self, c_int, pid_t, winsize, SIGCHLD, TIOCSCTTY, WNOHANG};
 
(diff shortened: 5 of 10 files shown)


src/event.rs
```diff
@@ -47,6 +47,7 @@ pub struct ActionContext<'a, N> {
     pub suppress_chars: &'a mut bool,
     pub last_modifiers: &'a mut ModifiersState,
     pub window_changes: &'a mut WindowChanges,
+    pub child_info: &'a tty::ChildInfo
```
rkanati: `ActionContext` needs a way to get at the child PID for `spawn_new_instance`, and this seemed better than a `mut static`.

src/event_loop.rs
```diff
@@ -343,7 +343,7 @@ impl<T> EventLoop<T>
 
             let poll_opts = PollOpt::edge() | PollOpt::oneshot();
 
-            let tokens = [1, 2];
+            let tokens = [1, 2, 3];
```
rkanati: Obviously we know that one token is shared for both read + write on unix, so we could use `2` for the signal pipe... but better safe than sorry, right?

src/event_loop.rs
```diff
@@ -374,13 +374,26 @@ impl<T> EventLoop<T>
                         CHANNEL => if !self.channel_event(&mut state) {
                             break 'event_loop;
                         },
+
+                        #[cfg(unix)]
+                        token if token == self.pty.signal_token() => {
+                            let should_exit = self.pty.process_signals();
+                            if should_exit {
```
rkanati: Returning a `bool` here sucks, but I'm not sure what would be better.

src/event_loop.rs
```diff
@@ -374,13 +374,26 @@ impl<T> EventLoop<T>
...
                         token if token == self.pty.read_token() || token == self.pty.write_token() => {
-                            #[cfg(unix)]
-                                {
-                                    if UnixReady::from(event.readiness()).is_hup() {
-                                        break 'event_loop;
-                                    }
+                            #[cfg(unix)] {
+                                if UnixReady::from(event.readiness()).is_hup() {
+                                    // don't try to do I/O on a dead Pty
+                                    continue;
```
rkanati: Not doing this caused the `pty_read` / `_write` to `EIO`, and I'm still not totally sure I've fixed that.

src/tty/mod.rs
```diff
@@ -30,10 +30,18 @@ mod windows;
 #[cfg(windows)]
 pub use self::windows::*;
 
+/// This struct contains information about the child process.
+pub struct ChildInfo {
+    #[cfg(unix)]
+    pub pid: libc::pid_t,
+}
```
rkanati: I feel like this could be useful on any platform, but since `spawn_new_instance` currently only works on 'nix anyway, I thought I'd leave it alone for now.

src/tty/mod.rs
```diff
@@ -30,10 +30,18 @@ mod windows;
...
+
 /// This trait defines the behaviour needed to read and/or write to a stream.
 /// It defines an abstraction over mio's interface in order to allow either one
 /// read/write object or a seperate read and write object.
 pub trait EventedReadWrite {
+    // TODO: should this maybe just be called `Pseudoterminal` or `Pty`?
+    // The current attempt at generality is probably unnecessary.
```
rkanati: If we dropped the pretense and had `trait Pseudoterminal` or similar, it'd feel less odd adding a `should_exit` method, which would avoid the gross `bool` on `process_signals`.

src/tty/unix.rs
```diff
@@ -15,59 +15,79 @@
...
-static mut SHOULD_EXIT: bool = false;
-
-extern "C" fn sigchld(_a: c_int) {
-    let mut status: c_int = 0;
-    unsafe {
-        let p = libc::waitpid(PID, &mut status, WNOHANG);
-        if p < 0 {
-            die!("Waiting for pid {} failed: {}\n", PID, errno());
+/// We write any signals into the pipe, and the event loop wakes up and reads them.
+static mut SIG_PIPE_WRITE: Option<File> = None;
```
rkanati: This is an `Optional` because it gets initialized when spawning the child - adding a little sync to this would probably work out nicely when we eventually get around to multiple windows (like #1589).

src/tty/unix.rs
```diff
@@ -15,59 +15,79 @@
...
+
+unsafe extern "C" fn on_signal(sig: c_int) {
+    let old_errno = errno();
+
+    let byte = [sig as u8];
+    let file: &mut File = SIG_PIPE_WRITE.as_mut().unwrap();
+    match file.write_all(&byte) {
+        Ok(_) => {
+            // TODO: a flush is almost certainly unnecesary
+            file.flush().unwrap();
```
rkanati: ...or _is it_? :thinking:

src/tty/unix.rs
```diff
@@ -15,59 +15,79 @@
...
-/// Get the current value of errno
-fn errno() -> c_int {
-    ::errno::errno().0
+// lifted from nix::unistd (MIT)
+fn pipe() -> (File, File) {
+    unsafe {
+        let mut fds: [c_int; 2] = mem::uninitialized();
+        if let 0 = libc::pipe(fds.as_mut_ptr()) {
+            set_nonblocking(fds[0]);
+            set_nonblocking(fds[1]);
```
rkanati: Non-blocking wouldn't seem to have any benefit, but _everyone_ seems to do it with self-pipes. Let me know if it's unnecessary.

src/tty/unix.rs
```diff
@@ -185,9 +205,11 @@ fn get_pw_entry(buf: &mut [i8; 1024]) -> Passwd<'_> {
 }
 
 pub struct Pty {
+    child: Child,
     pub fd: File,
-    pub raw_fd: RawFd,
```
rkanati: I took the liberty of cleaning this up - why keep this when you can `fd.as_raw_fd()`?

src/tty/unix.rs
```diff
@@ -206,6 +228,10 @@ impl Pty {
             die!("ioctl TIOCSWINSZ failed: {}", errno());
         }
     }
+
+    pub fn child_info(&self) -> ChildInfo {
+        ChildInfo { pid: self.child.id() as _ }
+    }
```
rkanati: If we turn `trait EventedReadWrite` into `trait Pty` or similar, this should either go there, or maybe come back in a tuple from `Pty::new`.

src/tty/unix.rs
```diff
@@ -294,23 +320,28 @@ pub fn new<T: ToWinsize>(
...
-                PID = child.id() as _;
+            // Handle SIGCHLD
+            let sig_pipe_read = {
+                let (read_end, write_end) = pipe();
+                unsafe { SIG_PIPE_WRITE = Some(write_end); }
+                read_end
+            };
+
+            // TODO: replace with sigaction?
+            unsafe { libc::signal(SIGCHLD, on_signal as _) };
```
rkanati: Please advise.

src/tty/unix.rs
```diff
@@ -335,26 +366,47 @@ impl EventedReadWrite for Pty {
...
+        )?;
+
+        self.sig_token = (*token.next().unwrap()).into();
+        poll.register(
+            &EventedFd(&self.sig_pipe_read.as_raw_fd()),
+            self.sig_token,
+            mio::Ready::readable(),
+            mio::PollOpt::level()
         )
     }
```
rkanati: Sticking the `register`/`reregister`/`deregister` together like this seems really gross, but I'm hard-pushed to think up a better approach.

src/tty/unix.rs
```diff
@@ -376,6 +428,33 @@ impl EventedReadWrite for Pty {
...
+            let mut sig = [0u8];
+            self.sig_pipe_read.read_exact(&mut sig).unwrap();
+            sig[0]
+        };
+
+        match sig as i32 {
+            libc::SIGCHLD => {
+                if let Some(_) = self.child.try_wait().expect("error collecting pty child") {
+                    return true;
+                }
```
rkanati: We have to be careful here to flat-out _ignore_ `SIGCHLD` from other sources, or we risk breaking anything that uses `spawn_daemon` or similar.

rkanati:
Once this has been reviewed and tweaked, I'll squash down all those messy commits.

chrisduerr:
Thanks a ton for the PR. Fixing this would be a huge step forward and lovely to include in the next release.

Unfortunately I'm not very familiar with signal safety, so I'll have to set aside some time to give this a proper review and test and I'm currently hung up on polishing the #2120 PR a bit.

I just wanted to let you know that this PR has not been missed and I'll definitely take a look this weekend (hopefully I can even get to it a bit earlier).

rkanati:
Nice! No worries.

src/event_loop.rs
chrisduerr: I'd either rename the method to something that clearly return a bool, or maybe return an enum with the type of signal that was processed? Or just an action that should be taken maybe.

src/tty/mod.rs
chrisduerr: I've also had issues with the name when this was initially introduced, but iirc there was a good reason for it and the naming was taken from a related API.

(See https://docs.rs/mio/0.6.16/mio/event/trait.Evented.html)

src/tty/unix.rs
```diff
@@ -15,59 +15,79 @@
...
 
-pub fn process_should_exit() -> bool {
-    unsafe { SHOULD_EXIT }
+    set_errno(old_errno);
 }
 
-/// Get the current value of errno
-fn errno() -> c_int {
-    ::errno::errno().0
+// lifted from nix::unistd (MIT)
```
chrisduerr: Why would we copy a function from a library, rather than using the library?

src/tty/unix.rs
chrisduerr: I feel like we should either handle a `None` value here or choose a different way to store this, otherwise I'm not sure which benefit we get from having an Option if we always crash anyways when it's `None`.

src/event_loop.rs
rkanati: I'll try an `enum`, see how it feels.

src/tty/unix.rs
rkanati: It might be my weak rust-fu - I used `Option` just because it has to be initialized dynamically.

src/tty/unix.rs
rkanati: Because iirc we don't depend on `nix` (which is honestly kind of astounding), and I didn't want to incur something so chunky for one bugfix.

rkanati:
I'll mess with some of these bits sometime soon. Ta!

src/tty/unix.rs
chrisduerr: We have an indirect dependency on nix anyways iirc. So we're pulling in one version of it anyways. Also I think in general it's the much better approach compared to copy/pasting code over, which might cause all kinds of trouble.

src/event_loop.rs
rkanati: So, it turned out the `EIO` was actually a racy bug that's been present for who-knows-how-long, and it's a *miracle* it hasn't burnt us before - after thinking I'd screwed up somehow for a couple days, I finally reproduced it on 0.2.9. See the comments in the new code below.

src/tty/unix.rs
rkanati: Turns out all the alternatives are worse, currently. The only other legit way to do this on `stable` is `lazy_static!`, and that's even more overkill that `Option`.

src/tty/mod.rs
```diff
@@ -59,6 +59,29 @@ pub trait EventedReadWrite {
...
+    #[cfg(unix)]
+    fn child_event_token(&self) -> mio::Token;
+
+    /// Tries to retrieve an event
+    ///
+    /// Returns `Some(event)` on success, or `None` if there are no events to retrieve.
+    #[cfg(unix)]
+    fn get_child_event(&mut self) -> Option<ChildEvent>;
+}
+
```
rkanati: I think this is better than bloating `EventedReadWrite`, and it's free to be used by the win32 `tty` if they want to.

rkanati:
I'm doing some finishing touches, and giving the travis instance a workout (I don't have win or mac boxes to build/test on), but hopefully this will be ready soon.

rkanati:
Provided there are no _really_ hidden screw-ups here, I'm pretty happy with the state of this. I can make fixups or additions as requested.

src/tty/unix.rs
rkanati: Using `nix` now.

rkanati:
Okay, hold up - I've been reading about `signal` vs `sigaction` and I need to switch it over. Again, sheer luck that this hasn't messed us up before.

src/tty/mod.rs
```diff
@@ -53,6 +59,29 @@ pub trait EventedReadWrite {
...
+/// In particular, this allows for race-free child exit notification on UNIX (cf. `SIGCHLD`).
+pub trait EventedPty : EventedReadWrite {
+    #[cfg(unix)]
+    fn child_event_token(&self) -> mio::Token;
+
+    /// Tries to retrieve an event
+    ///
+    /// Returns `Some(event)` on success, or `None` if there are no events to retrieve.
+    #[cfg(unix)]
+    fn get_child_event(&mut self) -> Option<ChildEvent>;
```
chrisduerr: `get_child_event` sounds like it just retrieves a field variable (like a getter), however if I understand things correctly, this is more like pulling the next event out of the PTY, right? It should probably be called something like `next_child_event` then, to represent that calling it twice in quick succession usually never returns the same value?

src/tty/unix.rs
```diff
@@ -15,106 +15,76 @@
...
+use crate::tty::{EventedReadWrite, EventedPty, ChildEvent, ChildInfo};
 use crate::term::SizeInfo;
 use crate::display::OnResize;
 use crate::config::{Config, Shell};
 use crate::cli::Options;
 use mio;
 
-use libc::{self, c_int, pid_t, winsize, SIGCHLD, TIOCSCTTY, WNOHANG};
+use libc::{self, c_int, winsize, TIOCSCTTY};
+use nix::{ unistd::pipe, pty::openpty, sys::signal::* };
```
chrisduerr: Why the glob import here? Elements should probably be imported individually and the extra whitespace should be removed.

src/tty/unix.rs
```diff
@@ -15,106 +15,76 @@
...
-    let mut status: c_int = 0;
-    unsafe {
-        let p = libc::waitpid(PID, &mut status, WNOHANG);
-        if p < 0 {
-            die!("Waiting for pid {} failed: {}\n", PID, errno());
-        }
-
-        if PID == p {
-            SHOULD_EXIT = true;
+/// We write and signals into the pipe, and the event loop wakes up and reads them.
```
chrisduerr: Either there's something missing here or the `and` should be removed.

src/tty/unix.rs
```diff
@@ -15,106 +15,76 @@
...
+static mut SIG_PIPE_WRITE: Option<File> = None;
+
+extern "C" fn on_signal(sig: c_int) {
+    let old_errno = errno();
+
+    let byte = [sig as u8];
+    let file = unsafe { SIG_PIPE_WRITE.as_mut().unwrap() };
```
chrisduerr: While this is perfectly safe, I'm not sure if we want to take any chances here. It's an `Option` already anyways and we just unwrap that, but we could just make it something thread-safe and then return in case there's anything wrong.

An `Option<Arc<Mutex<File>>>` or something similar could just be checked without any potential for error at all, that would probably be a nice change.

src/tty/unix.rs
```diff
@@ -277,12 +253,14 @@ pub fn new<T: ToWinsize>(
...
-            libc::signal(libc::SIGTERM, libc::SIG_DFL);
-            libc::signal(libc::SIGALRM, libc::SIG_DFL);
+            let dfl = SigAction::new(SigHandler::SigDfl, SaFlags::empty(), SigSet::empty());
+            sigaction(Signal::SIGCHLD, &dfl).unwrap();
+            sigaction(Signal::SIGHUP,  &dfl).unwrap();
+            sigaction(Signal::SIGINT,  &dfl).unwrap();
+            sigaction(Signal::SIGQUIT, &dfl).unwrap();
+            sigaction(Signal::SIGTERM, &dfl).unwrap();
+            sigaction(Signal::SIGALRM, &dfl).unwrap();
+            sigaction(Signal::SIGPIPE, &dfl).unwrap();
```
chrisduerr: Just for documentation, the `sigaction` call will fail when either the signal is invalid or a signal that can't be caught, or the `SaFlags` contain `SA_SIGINFO`. Since the flags are empty and the signals are constant, it is safe to unwrap here and guaranteed to work (assuming it was tested at least once and none of the signals were illegal).

src/tty/unix.rs
```diff
@@ -378,6 +400,40 @@ impl EventedReadWrite for Pty {
...
+    }
+
+    #[inline]
+    fn child_event_token(&self) -> mio::Token {
+        self.sig_token
+    }
+}
+
+pub fn process_should_exit() -> bool {
+    false // TODO: remove this stub if/when windows no longer needs it
```
chrisduerr: By this I'd assume you mean that we should implement the `get_child_event` method on the winpty and return a new child event when the PTY should quit on Windows?

If this is the case and you don't want to include it in this PR (which is perfectly fine by me, since that's kinda separate anyways), I'd propose opening an issue and removing this TODO, so it's tracked in github. Alternatively the TODO could also document what actually needs to be done and maybe put a link to the issue.

rkanati:
I skimmed these quickly, but I should have time to go through properly this
evening. Thanks!

src/event_loop.rs
```diff
@@ -371,41 +370,51 @@ impl<T> EventLoop<T>
...
+                        CHANNEL_TOKEN => if !self.channel_event(CHANNEL_TOKEN, &mut state) {
                             break 'event_loop;
                         },
+
+                        #[cfg(unix)]
+                        token if token == self.pty.child_event_token() => {
+                            while let Some(child_event) = self.pty.get_child_event() {
+                                if child_event == tty::ChildEvent::Exited {
+                                    let mut term = self.terminal.lock();
+                                    term.exit();
```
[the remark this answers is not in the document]
rkanati: Done.

src/tty/mod.rs
rkanati: Agreed - done.

src/tty/unix.rs
rkanati: I'm not confident putting `Arc` or `Mutex` in there without first being certain they don't do anything that's not async-signal-safe. `Arc`, I assume, can `drop` and thus `free`? Not okay for signal handlers.

Honestly, even using `File` makes me slightly uncomfortable. I'm thinking about using a `RawFd` in an atomic and a `libc::write` instead.

chrisduerr: `Arc` or `Mutex` don't inherently do anything async unsafe. Quite the opposite they're used to *synchronize* access to a variable.

Now if the variable itself can't be accessed from multiple threads, that's a different thing. But then it shouldn't be publicly accessible at any time anyways. `Arc` or `Mutex` don't make this any worse, they just don't make it any better either.

Using a `RawFd` isn't necessarily any better either. And a `RawFd` can be wrapped in synchronization structs too.

The goal here is to make it impossible to do anything unsafe with it, so adding synchronization across threads is not going to make things worse.

src/event_loop.rs
```diff
@@ -371,41 +370,51 @@ impl<T> EventLoop<T>
...
                     match event.token() {
-                        CHANNEL => if !self.channel_event(&mut state) {
+                        CHANNEL_TOKEN => if !self.channel_event(CHANNEL_TOKEN, &mut state) {
                             break 'event_loop;
                         },
+
+                        #[cfg(unix)]
+                        token if token == self.pty.child_event_token() => {
+                            while let Some(child_event) = self.pty.get_child_event() {
+                                if child_event == tty::ChildEvent::Exited {
```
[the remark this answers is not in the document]
rkanati: That works if `Exited` is the only possible event, and currently it is, but I thought it couldn't hurt to be suggestive here.

src/tty/unix.rs
rkanati: Apologies if I'm misreading you, but I think you may have misunderstood the issue here. Signal handlers aren't merely concurrent like threads, they also operate in a really strange, restrictive context, which makes various seemingly-innocuous operations suddenly really unsafe.

So far, `SIG_PIPE_WRITE` has been okay because it's set once _before_ the handler is installed, and then remains constant... but technically even then we're not 100% correct, because

1. We don't enforce any ordering between setting `SIG_PIPE_WRITE` and installing the handler, and
2. The access in `on_signal` isn't volatile.

There may be better solution to this, however; let me get back to you.

src/event_loop.rs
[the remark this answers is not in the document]
chrisduerr: We shouldn't style code for ways things might potentially be in the future.

If there are more events in the future, this can be refactored. Until then a cleaner solution should be preferred.

src/tty/unix.rs
chrisduerr: > Signal handlers aren't merely concurrent like threads, they also operate in a really strange, restrictive context, which makes various seemingly-innocuous operations suddenly really unsafe.

I'm aware of this, however I wanted to point out that the current solution is also vulnerable to those issues. If we want to make things Rust-safe, using Arc and Mutex is an option in the global scope, but obviously an approach should be taken that is not only thread-safe, but also signal-safe.

Of course sometimes there's no way around adding an `unsafe` somewhere, however that should be carefully considered for potential abuse.

src/event.rs
rkanati: Okay, I replaced it with an `AtomicIsize` and a `pub fn`, and stripped out the extra plumbing.

Originally I did this thinking it'd be easier to deal with for multiple windows (if we ever do that), but that's probably premature.

rkanati:
Implemented most of your suggestions, @chrisduerr, but I've got one final big trick up my sleeve that might sidestep a lot of it - stay tuned.

rkanati:
...So here's v2.

It's amazing what you can find on crates.io.

chrisduerr:
> It's amazing what you can find on crates.io.

I didn't even think about using a crate for this, but it makes a lot of sense that this is something commonly used and should have a well maintained crate.

src/tty/mod.rs
```diff
@@ -53,6 +53,29 @@ pub trait EventedReadWrite {
...
+/// In particular, this allows for race-free child exit notification on UNIX (cf. `SIGCHLD`).
+pub trait EventedPty : EventedReadWrite {
+    #[cfg(unix)]
+    fn child_event_token(&self) -> mio::Token;
+
+    /// Tries to retrieve an event
+    ///
+    /// Returns `Some(event)` on success, or `None` if there are no events to retrieve.
+    #[cfg(unix)]
+    fn next_child_event(&mut self) -> Option<ChildEvent>;
```
chrisduerr: We already have a `next_child_event` method, so that would likely raise the question why `EventedPty` doesn't just have a `next` function which returns an `Option<ChildEvent>` using the Iterator trait. And I think that might actually be the correct thing to do here, since a `next_*` method which acts pretty much exactly like an iterator but isn't implemented as an iterator is a bit odd idiomatically.

Just changing this to `pub trait EventedPty: EventedReadWrite + Iterator<Item = ChildEvent>` should take care of the `next` method automatically. The obvious drawback here would be that this is more difficult to do cleanly with the platform-specific annotations, but I don't think that's a huge deal.

rkanati:
I think that's fair that `next_child_event` isn't idiomatic, but an
`Iterator for EventedPty` setup seems a bit... clever.

How about we meet halfway?

```rust
fn child_events(&mut self) -> SomeKindaIterator<ChildEvent>;
```

chrisduerr:
> an `Iterator for EventedPty` setup seems a bit... clever

Hmm it definitely might be a bit clever and usually simple and stupid solutions are the way to go, however I'm not sure if it's too bad in this case.

Having a `child_events` function that returns an iterator seems a bit pointless, since we don't really gain anything here but just make it harder to access the child events since we now need two methods to access events.

Looking at the actual implementation of it on unix, it actually is very close to an iterator though, since we are actually calling the `next()` method on the `pending()` call there. So I feel like an iterator would be the idomatic and clean way to represent what's actually going on, even though we might not need anything beyond the usual `next()` method.

rkanati:
I'm mostly just saying that the call site will look like
```rust
for child_event in pty {
    ...
```
which reads strangely to me... but I'm not gonna die on this hill or anything. I'll go ahead and change it.

chrisduerr:
@rkanati You can just call `next` manually, in which case the callsite should look exactly the same way it does now. Or am I missing something here?

rkanati:
@chrisduerr It just feels less explicit to me. The `pty.child_events()` seems clearer.

chrisduerr:
> It just feels less explicit to me.

That's a great point, since this is a trait the name of the implementing struct doesn't necessarily make it clear what this iterates over.

Since we'll never need any advanced iterator functions, I think keeping it the way it is right now is probably a good idea. Straight-forward and simple too. Thanks for the input!

rkanati:
I'll have a look at the changelog, then.

Also, do you want these commits left as-is (they're fairly neat, I think), or all squashed into one?

chrisduerr:
> Also, do you want these commits left as-is (they're fairly neat, I think), or all squashed into one?

Please do not squash commits. That just makes things tougher to review.

I'll squash them myself before merging, thanks!

rkanati:
There we go.

src/term/mod.rs
```diff
@@ -1336,7 +1336,7 @@ impl Term {
 
     #[inline]
     pub fn should_exit(&self) -> bool {
-        tty::process_should_exit() || self.should_exit
+        self.should_exit
```
chrisduerr: The `use crate::tty;` import should be removed.

src/tty/unix.rs
```diff
@@ -15,54 +15,33 @@
...
-        if p < 0 {
-            die!("Waiting for pid {} failed: {}\n", PID, errno());
-        }
-
-        if PID == p {
-            SHOULD_EXIT = true;
-        }
-    }
-}
+static PID: AtomicIsize = ATOMIC_ISIZE_INIT;
```
chrisduerr: This has been deprecated and should be replaced with `AtomicIsize::new(0)`, it's also questionable if an isize should be used here at all, since I don't think PIDs can ever be negative.

rkanati: Huh, weird. I only chose `Isize` since `pid_t` is `i32`, but `Usize` works too.

src/term/mod.rs
rkanati: It's actually needed for `tty::isconpty` on windows - I'll pop a `#[cfg(windows)]` on there.

src/tty/unix.rs
chrisduerr: I believe that group IDs can be negative, but I don't think PIDs can be. So using a usize here should be fine.

src/term/mod.rs
chrisduerr: Thanks, it's mainly about the compile-time warning. Including it conditionally makes sense here.

chrisduerr:
I think this should be good to go now. I've tested everything and it's looking good. No code issues are jumping out either and the approach seems sane.

Thanks a ton for your work @rkanati, I'm glad this issue is finally resolved.

jwilm:
Thanks for tackling this very long standing issue @rkanati and thanks @chrisduerr for helping to land it! Exciting to finally have this fixed.

rkanati:
Happy to help. I've been on a roll with Rust recently, and this seemed like a good excuse to cut my teeth. Thanks for the learning experience!

PR #2115, opened 2019-02-16
rustfmt src/{event,input,selection}.rs

JeanMertz:
I was picking up work on https://github.com/jwilm/alacritty/pull/2022, and had to disable my editor's rustfmt.

I noticed that the [contributing guidelines](https://github.com/jwilm/alacritty/blob/master/CONTRIBUTING.md#style) mention:

> Alacritty currently does not have any automatically enforced style guidelines. As a result of that, it is not possible to run `rustfmt` on existing files. New code should however follow the default ruleset of `rustfmt` and for newly created files it is possible to run the `rustfmt` tool directly.

This runs `rustfmt +stable` (without any custom configurations) on these files:

* src/event.rs
* src/input.rs
* src/selection.rs

I checked the existing 14 open PRs and there are only a couple of them that touch these files, so I figured I'd try to get this formatting PR through.

* https://github.com/jwilm/alacritty/pull/1398 – marked as "This will need a complete rework since it's not clean at all"
* https://github.com/jwilm/alacritty/pull/1512 - will have most breakage, but it should be relatively easy to solve the merge conflicts (I can do it, if requested)
* https://github.com/jwilm/alacritty/pull/1589 - marked as "I would like to pick this back up at some point, but not sure when that will be at this point"
* https://github.com/jwilm/alacritty/pull/2114 - only 6 lines are affected

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -1,33 +1,33 @@
 //! Process window events
+use std::borrow::Cow;
+use std::env;
 #[cfg(unix)]
 use std::fs;
-use std::borrow::Cow;
 use std::fs::File;
 use std::io::Write;
 use std::sync::mpsc;
-use std::time::{Instant};
-use std::env;
+use std::time::Instant;
 
-use serde_json as json;
-use parking_lot::MutexGuard;
-use glutin::{self, ModifiersState, Event, ElementState, MouseButton};
-use copypasta::{Clipboard, Load, Store, Buffer as ClipboardBuffer};
+use copypasta::{Buffer as ClipboardBuffer, Clipboard, Load, Store};
 use glutin::dpi::PhysicalSize;
+use glutin::{self, ElementState, Event, ModifiersState, MouseButton};
+use parking_lot::MutexGuard;
+use serde_json as json;
 
-#[cfg(unix)]
-use crate::tty;
-use crate::grid::Scroll;
-use crate::config::{self, Config};
 use crate::cli::Options;
+use crate::config::{self, Config};
 use crate::display::OnResize;
-use crate::index::{Line, Column, Side, Point};
-use crate::input::{self, MouseBinding, KeyBinding};
+use crate::grid::Scroll;
+use crate::index::{Column, Line, Point, Side};
+use crate::input::{self, KeyBinding, MouseBinding};
 use crate::selection::Selection;
 use crate::sync::FairMutex;
-use crate::term::{Term, SizeInfo};
 use crate::term::cell::Cell;
-use crate::util::{limit, start_daemon};
+use crate::term::{SizeInfo, Term};
+#[cfg(unix)]
+use crate::tty;
 use crate::util::fmt::Red;
+use crate::util::{limit, start_daemon};
 use crate::window::Window;
 
 /// Byte sequences are sent to a `Notify` in response to some events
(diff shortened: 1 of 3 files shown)


JeanMertz:
Sidenote: it might make sense to close PRs that are marked as "won't be updated in the forseeable future". That way, it's easier to see what parts of the code can already be formatted in follow-up PRs.

A closed PR can always be reopened later, so there's no downside here, as far as I can tell.

rkanati:
Re: #2114, I can rebase if/when necessary, so no worries.

chrisduerr:
Thanks for the PR, but I don't think this is something that PRs are welcome on. I might make an exception for this one, since it's the first one in line, however there are a few issues.

First of all, I'd like to run rustfmt on everything or nothing. Having half the codebase follow rustfmt with the other half not following that ruleset achieves very little and will require us to run rustfmt on these same files in the future, just causing unnecessary repeated repository changes.

Following this last point, it's not easily possible to tell what the state of forks is just by looking at PRs. I have a couple of local changes on my machine for example which I haven't sent a PR for yet. This isn't a big deal, but if someone has to rebase large chunks repeatedly because we don't do everything in one go, it can become very annoying and difficult to merge.

Thirdly there's the problem with code reviews. If a collaborator who is trusted to know what he's doing is running rustfmt, there's no need for review. But it's hard to make any assumptions about other people sending rustfmt PRs. So I'd have to review the changes. Of course reviewing each line changed would be ridiculous, I'd probably just run rustfmt on these files myself and check for a diff, but at that point I could have just sent the PR myself.

Lastly I'd like to point out that to make sure that files don't have to be changed repeatedly, I want to run rustfmt checks on CI. To prevent having to run rustfmt multiple times, this should be done in the same PR which introduces these changes, to immediately solve many of the problems I've just talked about.

Since you didn't know about anything of this, I'll try to find the time to review and test this locally because in general I think this is a welcome change. It might take a while though because I'd like to follow up with a PR to run rustfmt on the rest of the files shortly after that to prevent some of the issues I've talked about. So you might have to rebase and rustfmt again once it's time.

Also please refrain from sending additional rustfmt PRs for now. Contributions are always welcome, but some things just need to be handled in a specific way to make sure they cause as little friction as possible.

JeanMertz:
> Thanks for the PR, but I don't think this is something that PRs are welcome on.

No worries. I've closed the PR, since I only spent a couple of minutes on it, and if it takes you longer to review, then it's just not worth it.

Some quick notes:

> it's not easily possible to tell what the state of forks is just by looking at PRs. I have a couple of local changes on my machine for example which I haven't sent a PR for yet

That's a very good point. I did consider this, but I went by [your comment](https://github.com/jwilm/alacritty/pull/2065#discussion_r254060549) noting "the plan to run `rustfmt` came up originall when there were like 40/50+ open PRs. So we're getting closer and closer to an acceptable number of pending changes". Which suggested to me that you too were mainly looking at open PRs. However, that completely misses the fact that you (and other core contributors) do the bulk of the changes, so obviously your local changes that haven't been upstreamed yet are the biggest to be impacted, other than the open PRs.

> Thirdly there's the problem with code reviews. If a collaborator who is trusted to know what he's doing is running rustfmt, there's no need for review.

Thank you for pointing this out. This is something which I hadn't considered, but it makes total sense. For me, I opened the three files, saved them, and let `rustfmt` do its thing, so _I_ know what's in this PR. for you as a reviewer, that context is completely lost, and without a certain trust level, you're going to have to review every line.

I should've thought of this, but I didn't. Thank you for reminding me.

> I want to run rustfmt checks on CI. To prevent having to run rustfmt multiple times

Also a good point. Though if doing a big-bang PR is difficult, doing it step by step is still better than not doing it at all, I think. Also this is where something like (Ruby's) "[.rubocop_todo.yml](https://github.com/rubocop-hq/rubocop/blob/master/.rubocop_todo.yml)" feature could benefit Rustfmt.

chrisduerr:
Thank you a lot for being so understanding. It's easy to get frustrated when proposing changes and they're just rejected. I'll take this as another incentive to run rustfmt as quickly as possible though, so hopefully this problem will be resolved soon!

PR #2123, opened 2019-02-18
Windows: Remove undefined assertion on return value of ClosePseudoConsole

davidhewitt:
Windows: ClosePseudoConsole has no return value, but `impl Drop for ConPty` was mistakenly trying to assert the return value was "OK".

When I wrote the ConPty implementation I'd been working from this document: https://github.com/MicrosoftDocs/Console-Docs/blob/master/docs/closepseudoconsole.md, and as you will see there the documentation states the function has **no return value**.

If you check the history of that file you'll see the docs used to state that it had an HRESULT return type. I wrote the assertion based on that erroneous doc.

I've removed the usage of the return value.


davidhewitt:
Thanks, changelog entry added!

PR #1708, opened 2018-10-27
WIP: Reenable dynamic title support on windows

zacps:
It turns out there were no issues with dynamic title support on windows,
I simply had an issue with my configuration which made me think it was
broken.

Fixes https://github.com/jwilm/alacritty/issues/1695

PR #2127, opened 2019-02-22
Add dynamic title support for Windows

### LINKED ISSUE #1695: Title does not get updated in Windows
Window title is always "Alacritty" on Windows OS

davidhewitt:
As discussed previously in #1708 I have successfully restored support for dynamic title for Windows, but only for the Conpty backend.

From what I can tell, winpty support would require us to fork the `winpty-agent.exe` binary. This would be incompatible with the current approach of downloading prebuilt versions of it. Perhaps worth closing #1708 PR and opening a new issue for winpty support to remember it for the future?

Fixes #1695.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - MSI installer for Windows is now available
 - New default key bindings Alt+Home, Alt+End, Alt+PageUp and Alt+PageDown
+- Dynamic title support on Windows
 
 ### Fixed
 
--- a/src/term/mod.rs
+++ b/src/term/mod.rs
@@ -799,7 +799,7 @@ pub struct Term {
     message_buffer: MessageBuffer,
 
     /// Hint that Alacritty should be closed
-    should_exit: bool
+    should_exit: bool,
 }
 
 /// Terminal size info
@@ -1357,19 +1357,14 @@ impl ansi::Handler for Term {
     #[inline]
     fn set_title(&mut self, title: &str) {
         if self.dynamic_title {
-            let mut title = title.to_owned();
+            self.next_title = Some(title.to_owned());
 
             #[cfg(windows)]
             {
-                // Winpty - the title is missing Alacritty
-                // at the front, must prepend it to get equivalent
-                // behaviour between Conpty and Winpty
                 if !tty::is_conpty() {
-                    title = String::from("Alacritty") + &title;
+                    self.next_title = Some(format!("Alacritty{}", title));
                 }
             }
-
-            self.next_title = Some(title);
         }
     }
 
--- a/src/term/mod.rs
+++ b/src/term/mod.rs
@@ -786,6 +786,7 @@ pub struct Term {
     /// Default style for resetting the cursor
     default_cursor_style: CursorStyle,
 
+    /// Whether to permit updating the terminal title
     dynamic_title: bool,
 
     /// Number of spaces in one tab
@@ -798,7 +799,7 @@ pub struct Term {
     message_buffer: MessageBuffer,
 
     /// Hint that Alacritty should be closed
-    should_exit: bool,
+    should_exit: bool
 }
 
 /// Terminal size info
@@ -1356,7 +1357,19 @@ impl ansi::Handler for Term {
     #[inline]
     fn set_title(&mut self, title: &str) {
         if self.dynamic_title {
-            self.next_title = Some(title.to_owned());
+            let mut title = title.to_owned();
+
+            #[cfg(windows)]
+            {
+                // Winpty - the title is missing Alacritty
+                // at the front, must prepend it to get equivalent
+                // behaviour between Conpty and Winpty
+                if !tty::is_conpty() {
+                    title = String::from("Alacritty") + &title;
+                }
+            }
+
+            self.next_title = Some(title);
         }
     }
 
--- a/src/term/mod.rs
+++ b/src/term/mod.rs
@@ -1361,8 +1361,14 @@ impl ansi::Handler for Term {
 
             #[cfg(windows)]
             {
-                if !tty::is_conpty() {
-                    self.next_title = Some(format!("Alacritty{}", title));
+                // cmd.exe in winpty: winpty incorrectly sets the title to ' ' instead of
+                // 'Alacritty' - thus we have to substitute this back to get equivalent
+                // behaviour as conpty.
+                //
+                // The starts_with check is necessary because other shells e.g. bash set a
+                // different title and don't need Alacritty prepended.
+                if !tty::is_conpty() && title.starts_with(' ') {
+                    self.next_title = Some(format!("Alacritty {}", title.trim()));
                 }
             }
         }
--- a/src/tty/windows/mod.rs
+++ b/src/tty/windows/mod.rs
@@ -34,6 +34,7 @@ mod winpty;
 
 /// Handle to the winpty agent or conpty process. Required so we know when it closes.
 static mut HANDLE: *mut c_void = 0usize as *mut c_void;
+static mut IS_CONPTY: bool = false;
 
 pub fn process_should_exit() -> bool {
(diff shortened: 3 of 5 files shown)


src/tty/windows/mod.rs
```diff
@@ -76,6 +76,20 @@ impl<'a> Pty<'a> {
...
+            PtyHandle::Conpty(_) => false
+        }
+    }
+
+    pub fn is_conpty(&self) -> bool {
+        match self.handle {
+            PtyHandle::Winpty(_) => false,
+            PtyHandle::Conpty(_) => true
+        }
+    }
```
chrisduerr: It seems a bit pointless to me to have both `is_conpty` and `is_winpty` when one is just the inverse of the other.

I think just having `is_conpty` would be good, since that's currently the 'new feature' we want to check for. In the future if this is the commonly used thing, it might make more sense to switch this around to `is_winpty` to query for backwards-compatability support, but I wouldn't bother with that for now.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Support dynamic title on Windows (Conpty backend only)
```
chrisduerr: Super tiny bikeshed, but I feel like `Dynamic title support on Windows` would sound a bit better when talking about an 'added feature'. Though I don't *really* care either way, just as a sidenote.

src/tty/windows/mod.rs
davidhewitt: I think that's reasonable. Will modify.

src/term/mod.rs
```diff
@@ -1350,6 +1363,11 @@ impl ansi::Handler for Term {
     /// Set the window title
     #[inline]
     fn set_title(&mut self, title: &str) {
+        #[cfg(windows)]
+        {
+            if !self.dynamic_title_supported { return; }
```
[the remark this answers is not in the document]
davidhewitt: (Added a comment in the source)

CHANGELOG.md
davidhewitt: No problem, reworded. I kept the note in brackets saying conpty-only. Wasn't sure if you wanted that removed too.

davidhewitt:
Thanks, have pushed changes as per your feedback 👍

src/term/mod.rs
```diff
@@ -1350,6 +1363,21 @@ impl ansi::Handler for Term {
...
+            // Windows - if using the winpty backend then Alacritty will
+            // recieve instructions to set the title to either blank, or
+            // e.g. the running process "- python".
+            //
+            // With the conpty backend the title is set correctly, so
+            // these same titles would be set from the conpty backend
+            // as e.g. "Alacritty" or "Alacritty - python"
+            //
+            // For this reason we disable setting the title when using
+            // the winpty backend.
```
chrisduerr: I don't see how completely disabling dynamic title support would be better than allowing WinPTY to set blank titles or something like `- python`. If people wish to have Alacritty as title, just disabling dynamic title support from the config would still be an option. Imho that is a better solution than adding a bunch of ugly workarounds just to disable a functionality which doesn't necessarily harm the user (and definitely doesn't cause any crashes).

Besides that, it sounds like the difference between WinPTY and ConPTY is just that ConPTY adds an `Alacritty` at the beginning. If that's all there is to it, why don't we just do the same with the WinPTY? That wouldn't require patching anything upstream, it seems like a pretty simple fix that should 'just work'?

davidhewitt: Fair point. I agree it's quite an ugly workaround.

> Besides that, it sounds like the difference between WinPTY and ConPTY is just that ConPTY adds an `Alacritty` at the beginning. If that's all there is to it, why don't we just do the same with the WinPTY?

I wondered about that - the titles are coming in from the term escape codes. Instead of a `dynamic_title_supported` field in `Term` we could have a `title_prefix` field. Which defaults empty, but in the `winpty` case we can set it to `"Alacritty"`. I'll do that.

chrisduerr: I don't like the idea of having a WinPTY specific hack like that in the config. If it makes sense to prepend `Alacritty` to all titles set by WinPTY, we should do exactly that, instead of shoving the responsibility on the user. Either it makes sense to do this (and we should do it), or it doesn't.

If users want to modify the title, then they can always set it explicitly, but we shouldn't be exposing workarounds to the user like that, when they've got no clue what the hell's going on.

davidhewitt: >  If it makes sense to prepend `Alacritty` to all titles set by WinPTY, we should do exactly that, instead of shoving the responsibility on the user.

Fully agree that the solution should be transparent to the user. I meant that this behaviour of appending `Alacritty` to all titles would need to be implemented in the `Term::set_title` code and based at runtime on either a `bool` or `String` configuring whether to do it at runtime.

As that also is quite ugly - for the moment I've just removed the Winpty workaround entirely.

davidhewitt:
Sure thing, I've added an `IS_CONPTY` global and brought parity between the winpty and conpty support using it.

davidhewitt:
Given the number of revisions, I've squashed everything into a single commit.

chrisduerr:
@davidhewitt Were you able to confirm that this actually does something? I've tested it myself and it doesn't seem like the escape is ever sent. Both in WinPTY and ConPTY.

davidhewitt:
Yep, definitely works (gif below). I've tested on current insiders slow ring (which I'm currently running) and also Windows 10 1809 update (current released windows 10).

This is with conpty but I also tested with winpty.



Thanks for fixing the style issues btw.

src/term/mod.rs
```diff
@@ -1352,6 +1353,18 @@ impl ansi::Handler for Term {
...
             self.next_title = Some(title.to_owned());
+
+            #[cfg(windows)]
+            {
+                // cmd.exe in winpty sends titles with blank space at the beginning
+                // because the title is not able to be passed to cmd.exe.
+                //
+                // The startswith check is necessary because other shells e.g. bash send a
+                // different title and don't need Alacritty appended.
+                if !tty::is_conpty() && self.next_title.as_ref().unwrap().starts_with(" ") {
```
chrisduerr: How does WinPTY reset the terminal title? If it just sends `""` and expects the terminal to reset the title to `Alacritty`, this obviously wouldn't work.

davidhewitt: Winpty always sends at least `' '` (a single space), as it incorrectly tells cmd.exe that that's the title (instead of `'Alacritty'`). That said, I've realised that what actually needs to be done is thus to replace the first space with `'Alacritty'` - otherwise we end up with an extra space compared to the conpty version. Have pushed a correction for that.

src/term/mod.rs
```diff
@@ -1352,6 +1353,22 @@ impl ansi::Handler for Term {
...
+                // 'Alacritty' - thus we have to substitute this back to get equivalent
+                // behaviour as conpty.
+                //
+                // The startswith check is necessary because other shells e.g. bash send a
+                // different title and don't need Alacritty prepended.
+                if !tty::is_conpty() && title.starts_with(' ') {
+                    self.next_title = Some(match title.get(1..) {
+                        None => "Alacritty".into(),
+                        Some(rest) => format!("Alacritty{}", rest)
+                    });
```
chrisduerr: Is there any difference between `"Alacritty"` and `"Alacritty "` as a Window title? It seems like if `" "` is really used to reset it instead of `""`, then just using `format!("Alacritty{}", title)` would be fine.

It's also possible to just slap a `trim()` on the title at any point, which would remove leading and trailing whitespace, so even if it makes a difference I think that would be an easier option than having to match the first character.

davidhewitt: Thanks - I think avoiding the double space is desirable (it's noticable, just barely) - using `trim()` is a much neater solution though so I've switched to that.

davidhewitt:
Rebased 😄

davidhewitt:
Also closes #1695 I think

Issue #1900, opened 2018-12-14
Add an installer for Windows


zacps:
Because Windows has no dedicated package manager, unlike Linux/OSX, the main method used for installing Windows programs is through the `.msi` format.

I started investigating using `cargo wix` for generating an installer, this could potentially be used with [self_update](https://crates.io/crates/self_update) for automatic updates.

Initial work at [zacps/alacritty#cargo-wix](https://github.com/zacps/alacritty/tree/cargo-wix).

chrisduerr:
> Because Windows has no dedicated package manager, unlike Linux/OSX

I've recently merged a PR with install instructions for scoop, I believe there's also another package manager with a name closed to chocolate. So while a linux-like package manager doesn't exist, a macOS-like third party addon does.

I still think it would be a good idea to have a way to install on Windows without these package managers though, since I don't think they're commonly used. The user would still have to update manually, but maybe the `watch releases` option on github is sufficient to resolve that problem for now.

zacps:
Yeah, there are a couple of them, but they're very much a minority.

What do you think about using that crate for updates? I can see why you'd want to keep packaging/updates separate from the main binary, but it does seem like the simplest solution to me.

chrisduerr:
As long as the code is clean and I am never bothered by it on Linux, I'd be fine with it. Though I'm not a huge fan, but there's little that can be done since it's just due to Windows' awful package management.

I'd be very concerned with the UX of updating though. We aren't gonna add any UI options, since Alacritty doesn't do that kinda stuff, so the only option seems to be to automatically update at startup. That might not only affect startup times significantly, it could also lead to a forced update situation which might prevent users from running older builds, when the latest stable breaks stuff. Both of these seem like deal-breakers to me.

If there's a decent way to do it, sure. I just don't know how I would go about doing it. There's a good reason why this is done by external applications on most operating systems. Patching it into Alacritty itself is just not going to work out extremely well.

Just showing an update notice with the text display we use for warnings/errors might be an option. But at that point there's no reason why we'd need to depend on a crate to update things. Combining that with an external script for updates is of course an option, but I think we're already distributing enough files on Windows, that's just going to make things way too complicated for the user.

If you have any ideas how this could work out well, I'm all ears. But it doesn't seem like we can just pull in the crate and we're good to go.

zacps:
[...]

PR #2129, opened 2019-02-25
Add initial installer

eberkund:
Work in progress for #1900 

- [x] Copy alaritty.exe and winpty-agent.exe to Program Files
- [x] Copy the default config file to `%APPDATA%`
- [x] Add the Program Files folder to `$PATH`
- [x] Add right click context menu


wix/alacitty.wxs
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="windows-1252"?>
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi"
+    xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+    <Product Name="Alacritty" Id="*" UpgradeCode="87c21c74-dbd5-4584-89d5-46d9cd0c40a7" Language="1033" Codepage="1252" Version="0.2.0">
+        <Package InstallerVersion="200" Compressed="yes" InstallScope="perMachine"/>
+        <MajorUpgrade AllowSameVersionUpgrades="yes" DowngradeErrorMessage="A newer version of [ProductName] is already installed."/>
```
chrisduerr: Have you tested that `[ProductName]` gets replaced properly, or might this be from a template which is supposed to be replaced with the actual application name?

ci/script.sh
```diff
@@ -28,6 +28,7 @@ if [ "$TRAVIS_OS_NAME" == "windows" ]; then
         cp "./target/debug/winpty-agent.exe" "./target/debug/deps"
     fi
     cargo test -p winpty || error=true
+    ./wix/build.sh
```
chrisduerr: Builds should never be done for normal CI builds, but only before deployment.

wix/license.rtf
```diff
@@ -0,0 +1 @@
+{\rtf1}
```
chrisduerr: I'm not familiar with wix, but Alacritty's license should probably be put here.

eberkund: It displays during installation and asks the user to accept the agreement. Should I copy and paste the license from [here](https://github.com/jwilm/alacritty/blob/master/LICENSE-APACHE)?

ci/script.sh
chrisduerr: Since you've forked Alacritty you can test deployment by creating releases for your local branch. That actually tests properly if the releases will be uploaded and everything.

eberkund: Oh okay, thanks I'll try that.

wix/license.rtf
chrisduerr: Yes, please.

wix/alacitty.wxs
eberkund: That gets replaced automatically from the Product element

eberkund:
Should be good now if anyone wants to test.

chrisduerr:
@eberkund Were you able to build releases and upload them to github releases? Would be the quickest way to confirm that this actually works.

eberkund:
I have to setup the builds under my own Travis account to do that right?

chrisduerr:
Yeah, you have to enable travis to build it, then change the secure token and repository (see https://github.com/jwilm/alacritty/blob/master/.travis.yml#L48-L56) and create a new tag.

Creating a new secure token requires you to add another developer token in github with the `public repo` permission and then use the travis CLI application to encrypt that token (using `travis encrypt -r eberkund/alacritty <GITHUB_PUBLIC_REPO_TOKEN>`).

That's usually how I do it and it worked fairly well for testing. If you're running into troubles with this, let me know and I'll set it up myself.

eberkund:
I can't get it to work, I created a new API token and encrypted it but it looks like it is getting an unauthenticated error for some reason

https://github.com/eberkund/alacritty/blob/master/.travis.yml

chrisduerr:
I'll set it up for testing myself. Thanks for trying though!

eberkund:
Thanks

ci/before_deploy.sh
```diff
@@ -54,9 +54,17 @@ elif [ "$TRAVIS_OS_NAME" == "linux" ] && [ "$ARCH" == "i386" ]; then
     # Make sure all files can be uploaded without permission errors
     sudo chown -R $USER:$USER "./target"
 elif [ "$TRAVIS_OS_NAME" == "windows" ]; then
-    choco install 7zip
+    choco install 7zip nuget.commandline
+    nuget install WiX
+
+    # Create zip archive
     7z a -tzip "./target/deploy/${name}-windows.zip" "./target/release/alacritty.exe" \
         "./target/release/winpty-agent.exe"
```
chrisduerr: I feel like it might be nice to rename this from `Alacritty-vX.X.X-windows.zip` to `Alacritty-vX.X.X-windows-portable.zip` (or `...-portable-windows.zip`). It seems like that would make the difference to the wix installer pretty clear and make it easy to figure out what the difference is.

ci/before_deploy.sh
```diff
@@ -54,9 +54,17 @@ elif [ "$TRAVIS_OS_NAME" == "linux" ] && [ "$ARCH" == "i386" ]; then
...
+
+    # Create msi installer
+    "./WiX.3.11.1/tools/candle.exe" -nologo -arch "x64" -ext WixUIExtension -ext WixUtilExtension -out "wix/alacritty.wixobj" "wix/alacritty.wxs"
```
chrisduerr: This will break and I'm not a fan of that at all. We shouldn't rely on a specific version of WiX.

The great thing is that we're running bash here though, so we are able to make use of the wonders of unix by changing this to `./WiX.*/tools/candle.exe`.

If it's possible to make sure that the version is not part of the wix folder name, that would of course be even better and less error-prone, but I'm not sure if it is. But maybe there's a nuget flag for this?

wix/alacritty.wxs
```diff
@@ -0,0 +1,84 @@
+<?xml version="1.0" encoding="windows-1252"?>
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" 
+    xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+    <Product Name="Alacritty" Id="*" UpgradeCode="87c21c74-dbd5-4584-89d5-46d9cd0c40a7" Language="1033" Codepage="1252" Version="0.2.9" Manufacturer="Alacritty">
```
chrisduerr: I'm not familiar with WiX, but just to make sure I have read up on it correctly:

The `UpgradeCode` is used as an ID to replace existing installs when a new version is installed, correct? So we will be able to keep this `UpgradeCode` the same until the end of time? Or does this need to be changed every time a new version is released?

wix/alacritty.wxs
```diff
@@ -0,0 +1,84 @@
...
+                <File Id="WinPtyAgentExeFile" Source="..\target\release\winpty-agent.exe" Name="winpty-agent.exe" KeyPath="yes"/>
+            </Component>
+        </DirectoryRef>
+
+        <DirectoryRef Id="AlacrittyAppData">
+            <!-- Copy default config -->
+            <Component Id="AlacrittyConfig" Guid="*" Permanent="yes">
+                <File Id="AlacrittyConfigFile" Source="..\alacritty.yml" Name="alacritty.yml" KeyPath="yes"/>
+            </Component>
+        </DirectoryRef>
```
chrisduerr: We don't actually want to copy over the configuration file. It's generated automatically at the right location once Alacritty is started, there's no point to duplicate this here.

wix/alacritty.wxs
```diff
@@ -0,0 +1,84 @@
...
+
+        <DirectoryRef Id="TARGETDIR">
+            <!-- Add context menu -->
+            <Component Id="ContextMenu" Guid="449f9121-f7b9-41fe-82da-52349ea8ff91">
+                <RegistryKey Root="HKCU" Key="Software\Classes\Directory\Background\shell\Alacritty Here\command">
+                    <RegistryValue Type="string" Value="[AlacrittyProgramFiles]alacritty.exe --working-directory &quot;%V&quot;" KeyPath="yes"/>
+                </RegistryKey>
+                <RegistryKey Root="HKCU" Key="Software\Classes\Directory\Background\shell\Alacritty Here">
```
chrisduerr: I've tested this with powershell and it uses `Open PowerShell here`, we should follow this convention and change this to `Open Alacritty here`.

wix/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+"WiX.3.11.1/tools/candle.exe" -nologo -arch "x64" -ext WixUIExtension -ext WixUtilExtension -out "wix/alacritty.wixobj" "wix/alacritty.wxs"
+
+"WiX.3.11.1/tools/light.exe" -nologo -ext WixUIExtension -ext WixUtilExtension -out "wix/alacritty.msi" -sice:ICE61 -sice:ICE91 "wix/alacritty.wixobj"
```
chrisduerr: We don't want to depend on a specific wix version here either.

wix/alacritty.wxs
eberkund: True, okay I will remove this part.

wix/alacritty.wxs
eberkund: The `UpgradeCode` should remain the same until the end of time, the `Id` should change with each update but it is set to `Id="*"` so that will be handled automatically.

ci/before_deploy.sh
eberkund: The more commonly used MSI distributed WiX package set an environment variable with the path of the WiX installation however unfortunately chocolatey has a known bug preventing installing this package so I had to use the Nuget distribution as a workaround. Using a wildcard is a good suggestion though.

ci/before_deploy.sh
eberkund: I think the .msi vs .zip extension should make this clear and that would be consistent with the way the rest of the releases are named. Also, "portable" usually means that the program is self-contained and writes config files into a relative directory.

wix/alacritty.wxs
eberkund: Okay, sounds good.

ci/before_deploy.sh
chrisduerr: That's a great point, though the way the configuration file is generated is likely to change soon (it's only going to be generated on demand). So I feel like this would still help with telling the user what the difference is without having to download it.

I don't think there's any precedence with the current releases, since 'portable' installs aren't really something commonly used outside of Windows in my experience.

eberkund: Fixed.

eberkund:
> Based on what I understand, it should be possible to add checkboxes to the wix installer, right? I think both the context menu and the shortcut should be optional, to make sure we don't push things on users that they do not want.

I agree this would be good to have. This isn't something I've done before and is something I would need to look into to see how this can be done. I was playing around with the feature tree dialog set: http://wixtoolset.org/documentation/manual/v3/wixui/dialog_reference/wixui_featuretree.html but it adds a lot of other unnecessary stuff.

Would it be possible to merge this as is and open a new improvement ticket for making certain features optional? I think as it is now it is just one more distribution option available, those who don't want it can continue using the old zip archive.

chrisduerr:
> Would it be possible to merge this as is and open a new improvement ticket for making certain features optional? I think as it is now it is just one more distribution option available, those who don't want it can continue using the old zip archive.

I think that's a reasonable suggestion, yes. Though I might not get to it until this weekend. So feel free to let me know if you've found out more until then.

PR #2134, opened 2019-02-26
Fix terminfo support for extended capabilities

chrisduerr:
To make sure all extended capabilities can be queried correctly, it is
necessary to compile the terminfo file using the `-x` flag.

Since Alacritty specified support for the XTerm mouse mode (XM/xm),
using the `-x` flag was not possible until now without breaking programs
like `htop`. By removing this flag, these issues should be resolved.

This fixes https://github.com/jwilm/alacritty/issues/2131.


JeanMertz:
> By removing this flag, these issues should be resolved.

One question: what's the downside of removing these flags? Is there something obvious that was supported before that will no longer work now?

chrisduerr:
The `XM` capability basically says we support XTerm mouse mode, which we do not apparently. So if we add it, `htop` is likely trying to use this capability, which then breaks. So by removing it everything should behave the same way it has been until now without the `-x` flag.

PR #2139, opened 2019-02-28
Add Alt+Home, Alt+End, Alt+PageUp and Alt+PageDown to default key bindings

stfnm:
These key bindings worked for me in xterm and urxvt but seemed to be missing in alacritty's defaults.

They're used for example in [WeeChat's default key bindings](https://weechat.org/files/doc/stable/weechat_user.en.html#key_bindings_buffers_windows) to scroll to the top of the buffer, the bottom, a few lines up or a few lines down.

Cheers!


stfnm:
Done! I added the changelog entry and rebased my branch.

PR #2140, opened 2019-02-28
Make start_daemon behaviour on Windows consistent with Unix

### LINKED ISSUE #2128: Windows: Cannot spawn new "cmd" instance using "command" keybinding
On windows if I add the following keybind:

      - { key: T,        mods: Control|Shift, command: {program: "cmd", args: []} }

Then when I hit Ctrl+Shift+T, the following happens:

- If the `Alacritty` instance running was launched from the desktop or taskbar, the `cmd` window immediately crashes.
- If the `Alacritty` instance was launched from another terminal, then that terminal is taken over by the new `cmd` process (wat!)

The fault all seems to lie in the `spawn_daemon` function. I think to fix we would need to use windows APIs directly rather than `std::process::Command`, which is a bit sad.

I will write up a PR, but I wanted to survey opinion first on what "correct" behaviour of this keybind should be. I propose the following:

Keybind | Result
------------ | -------------
`{program: "cmd", args: []}` | New command window opens
`{program: "cmd", args: ["echo", "hello"]}`| New command window opens, prints hello, immediately closes.
`{program: "foo", args: []}`| Program "foo" opens in new window

In particular, in Linux daemons as I understand are background processes. So should there be any situations where I can spawn a process with a hidden console window?

I discovered this while digging around with solving #1979. The `cmd.exe` pseudoconsole immediately crashes with the conpty backend.

davidhewitt:
Closes #2128.

As agreed in that thread, I have modified the `start_daemon` function on Windows to make behaviour as close to Unix as possible.

At the same time I have also fixed a crash where new Alacritty windows opened with `SpawnNewInstance` keybind would close instantly with the ConPTY backend. This was due to invalid stdio handles being passed to the conpty child process.

If you think this all looks reasonable I'll create a ChangeLog entry and add it to the PR.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -109,16 +109,10 @@ pub fn start_daemon<I, S>(program: &str, args: I) -> io::Result<()>
         I: IntoIterator<Item = S>,
         S: AsRef<OsStr>,
 {
-    // As per issue 2128 the decision was made to make this call as
-    // similar to the Unix behaviour as possible.
-    //
     // Setting all the I/O handles to null and setting the
     // CREATE_NEW_PROCESS_GROUP and CREATE_NO_WINDOW has the effect
     // that console applications will run without opening a new
     // console window.
-    //
-    // If visible output is desirable the DETACHED_PROCESS flag could
-    // be used instead of CREATE_NO_WINDOW.
     Command::new(program)
         .args(args)
         .stdin(Stdio::null())
--- a/src/util.rs
+++ b/src/util.rs
@@ -11,11 +11,20 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
+use std::{cmp, io};
+use std::ffi::OsStr;
+use std::process::Command;
+
 #[cfg(not(windows))]
 use std::os::unix::process::CommandExt;
-use std::process::Command;
-use std::ffi::OsStr;
-use std::{cmp, io};
+
+#[cfg(windows)]
+use std::os::windows::process::CommandExt;
+#[cfg(windows)]
+use std::process::Stdio;
+#[cfg(windows)]
+use winapi::um::winbase::{CREATE_NEW_PROCESS_GROUP, CREATE_NO_WINDOW};
 
 /// Threading utilities
 pub mod thread {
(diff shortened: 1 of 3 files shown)


src/util.rs
```diff
@@ -100,7 +109,24 @@ pub fn start_daemon<I, S>(program: &str, args: I) -> io::Result<()>
...
+    // that console applications will run without opening a new
+    // console window.
+    //
+    // If visible output is desirable the DETACHED_PROCESS flag could
+    // be used instead of CREATE_NO_WINDOW.
+    Command::new(program)
+        .args(args)
+        .stdin(Stdio::null())
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
```
chrisduerr: Shouldn't this be the default anyways? What happens when all this `Stdio::null()` mappings are removed?

davidhewitt: According to https://doc.rust-lang.org/std/process/struct.Command.html#method.new, for `.spawn()` the default is to inherit handles rather than null. This is also true for linux - but the `::libc::daemon(1, 0)` call will set all of the stdio handles to null in that case.

chrisduerr: You are absolutely correct, I just tested it again to confirm it and I must have remembered that incorrectly. Stdout of a child will get redirected to the stdout of the parent by default.

davidhewitt:
Thanks, ChangeLog entry added. I tried to keep it short - let me know if it's unclear and I'll expand it.

src/util.rs
```diff
@@ -100,7 +109,24 @@ pub fn start_daemon<I, S>(program: &str, args: I) -> io::Result<()>
         I: IntoIterator<Item = S>,
         S: AsRef<OsStr>,
 {
-    Command::new(program).args(args).spawn().map(|_| ())
+    // As per issue 2128 the decision was made to make this call as
+    // similar to the Unix behaviour as possible.
+    //
```
chrisduerr: I think putting this part explicitly into a comment doesn't provide much benefit. If anyone is curious, `git blame` will suffice here.

src/util.rs
```diff
@@ -100,7 +109,24 @@ pub fn start_daemon<I, S>(program: &str, args: I) -> io::Result<()>
...
+    // Setting all the I/O handles to null and setting the
+    // CREATE_NEW_PROCESS_GROUP and CREATE_NO_WINDOW has the effect
+    // that console applications will run without opening a new
+    // console window.
+    //
+    // If visible output is desirable the DETACHED_PROCESS flag could
+    // be used instead of CREATE_NO_WINDOW.
```
chrisduerr: While this might be a beneficial comment if this is planned to change in the future, I would recommend against putting it here since the plan is to keep things the way they are.

If it is desired to change this in the future, looking into the Windows documentation or this PR should be the first target anyways. This way we can keep the code comments to the essentials and don't have to waste time with info that is likely irrelevant.

src/util.rs
```diff
@@ -100,7 +109,24 @@ pub fn start_daemon<I, S>(program: &str, args: I) -> io::Result<()>
...
         S: AsRef<OsStr>,
 {
-    Command::new(program).args(args).spawn().map(|_| ())
+    // As per issue 2128 the decision was made to make this call as
+    // similar to the Unix behaviour as possible.
+    //
+    // Setting all the I/O handles to null and setting the
+    // CREATE_NEW_PROCESS_GROUP and CREATE_NO_WINDOW has the effect
+    // that console applications will run without opening a new
+    // console window.
```
chrisduerr: Just to be clear, this part of the docs is great and I'm all for leaving it in!

davidhewitt:
Thanks for the feedback on the comments - have removed the suggested pieces.

PR #2151, opened 2019-03-03
Fix mouse events ignored by padding

### LINKED ISSUE #2109: Allow highlighting drag to start from the border
This is sort of a papercut type issue, often when I go to highlight some text in the window especially when it's maximized and against the left edge of my screen I can't start the drag without moving a couple of pixels to the right which appears to be the width of the border of the window. It usually means I end up having to attempt to drag more than once to get it right or start from the other end. I think this would improve the ergonomics of alacritty.

Linux

Wayland, though it appears alacritty runs in XWayland as it shows up in xwininfo.

ladotonia:
SizeInfo ctx with padding was being used to determine if the last mouse location was valid, however the padding doesn't need to be applied to mouse locations. Mouse x,y is updated by CursorMoved events which only send events if the mouse is inside of the alacritty window bounds.

Instead, use usize::max_value() as a marker for invalid mouse positions and set to that value on CursorLeft events to ensure that mouse_coords() returns None when the mouse position is invalid. This should allow any mouse location inside the alacritty window to return a proper row/col coord.

This fixes #2109

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Cursor jumping around when leaving alt screen while not in the alt screen
 - Text lingering around when resetting while scrolled up in the history
 - Terminfo support for extended capabilities
+- Mouse presses are no long ignored in padding except for event propagation
 
 ## Version 0.2.9
 
@@ -249,7 +250,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Empty lines in selections are now properly copied to the clipboard
 - Selection start point lagging behind initial cursor position
 - Rendering of selections which start above the visible area and end below it
-- Mouse press no longer ignored in padding width
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -249,6 +249,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Empty lines in selections are now properly copied to the clipboard
 - Selection start point lagging behind initial cursor position
 - Rendering of selections which start above the visible area and end below it
+- Mouse press no longer ignored in padding width
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Cursor jumping around when leaving alt screen while not in the alt screen
 - Text lingering around when resetting while scrolled up in the history
 - Terminfo support for extended capabilities
-- Mouse presses are no long ignored in padding except for event propagation
+- Allow mouse presses and beginning of mouse selection in padding
 
 ## Version 0.2.9
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -124,7 +124,11 @@ impl<'a, N: Notify + 'a> input::ActionContext for ActionContext<'a, N> {
     }
 
     fn mouse_coords(&self) -> Option<Point> {
-        self.terminal.pixels_to_coords(self.mouse.x as usize, self.mouse.y as usize)
+        if self.mouse.x == usize::max_value() {
+            None
+        } else {
+            Some( self.size_info.pixels_to_coords(self.mouse.x as usize, self.mouse.y as usize) )
+        }
     }
 
     #[inline]
@@ -238,8 +242,8 @@ pub struct Mouse {
 impl Default for Mouse {
     fn default() -> Mouse {
         Mouse {
-            x: 0,
-            y: 0,
+            x: usize::max_value(),
+            y: usize::max_value(),
             last_click_timestamp: Instant::now(),
             left_button_state: ElementState::Released,
             middle_button_state: ElementState::Released,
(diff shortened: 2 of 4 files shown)


chrisduerr:
I'm not sure if you've read my review yet, but I think it would help significantly with simplifying this code.

If you have any questions about it, please let me know.

ladotonia:
@chrisduerr Thanks for the feedback. I did review it, and am open to closing this PR and submitting a a simpler option with the changes that you suggested.

I reviewed the other places where those methods are used and feel that this approach simplifies a few things. It also explicitly and only stores mouse positions within the window bounds without worrying about padding used elsewhere.

One thing I'm not familiar with is a pty interface, should events be propagated to that interface with a drag that extends the current selection? It seems like maybe that should receive the moving cursor location on the first column even when drags are made outside the bounds. Again, I'm not quick clear on how the pty is used, just a thought.

chrisduerr:
> I did review it, and am open to closing this PR and submitting a a simpler option with the changes that you suggested.

I see no reason to close this PR rather than just improving it.

> I reviewed the other places where those methods are used and feel that this approach simplifies a few things.

Are you talking about the PR you've proposed? Since it seems significantly more complex than my alternative. Though I've just tested a quick draft and haven't checked any edge cases.

> One thing I'm not familiar with is a pty interface, should events be propagated to that interface with a drag that extends the current selection? It seems like maybe that should receive the moving cursor location on the first column even when drags are made outside the bounds.

Updates to the selection are not propagated to the PTY. However based on mouse modes clicks and motion is reported. This should however not be the case for events over the padding, since the PTY does not care about padding at all.

I'd propose something like this maybe (based on master):
```diff
diff --git a/src/input.rs b/src/input.rs
index 7779dca..32e4dcb 100644
--- a/src/input.rs
+++ b/src/input.rs
@@ -411,7 +411,7 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         } else if self.ctx.terminal().mode().intersects(motion_mode)
             // Only report motion when changing cells
             && (prev_line != self.ctx.mouse().line || prev_col != self.ctx.mouse().column)
-            && size_info.contains_point(x, y)
+            && size_info.contains_point(x, y, false)
         {
             if self.ctx.mouse().left_button_state == ElementState::Pressed {
                 self.mouse_report(32, ElementState::Pressed, modifiers);
diff --git a/src/term/mod.rs b/src/term/mod.rs
index 966ac18..f6ac33f 100644
--- a/src/term/mod.rs
+++ b/src/term/mod.rs
@@ -838,11 +838,15 @@ impl SizeInfo {
         Column(((self.width - 2. * self.padding_x) / self.cell_width) as usize)
     }
 
-    pub fn contains_point(&self, x: usize, y:usize) -> bool {
-        x < (self.width - self.padding_x) as usize
-            && x >= self.padding_x as usize
-            && y < (self.height - self.padding_y) as usize
-            && y >= self.padding_y as usize
+    pub fn contains_point(&self, x: usize, y: usize, include_padding: bool) -> bool {
+        if include_padding {
+            x < self.width as usize && y < self.height as usize
+        } else {
+            x < (self.width - self.padding_x) as usize
+                && x >= self.padding_x as usize
+                && y < (self.height - self.padding_y) as usize
+                && y >= self.padding_y as usize
+        }
     }
 
     pub fn pixels_to_coords(&self, x: usize, y: usize) -> Point {
@@ -1106,7 +1110,7 @@ impl Term {
     ///
     /// Returns None if the coordinates are outside the screen
     pub fn pixels_to_coords(&self, x: usize, y: usize) -> Option<Point> {
-        if self.size_info.contains_point(x, y) {
+        if self.size_info.contains_point(x, y, true) {
             Some(self.size_info.pixels_to_coords(x, y))
         } else {
             None
```

ladotonia:
The reason I suggested removing the PR would be to remove the unecessary history. Not sure if the commits are squashed on merge.

I agree that your proposal is a much simpler change and it has the intended results, including filtering drag events in the padding from being sent to the pty interface. My goal was to provide an alternative that only ever stores valid in-window positions in the mouse object, to ensure consistency. In this sense, the current approach has no obvious negative effect now, but the mouse x,y may or may not store a location inside the window, based on whether or not the user dragged the cursor outside the extents.

I appreciate the feedback as it is very helpful to me as a Rust Newbie.

I'll gladly update / re-open a PR with your proposed changes at your request, or revert contains_point to filter out the padding from the pty event propagation.

chrisduerr:
> The reason I suggested removing the PR would be to remove the unecessary history. Not sure if the commits are squashed on merge.

Yeah no worries about that. All PRs are usually squashed before merging.

> My goal was to provide an alternative that only ever stores valid in-window positions in the mouse object, to ensure consistency.

This is a great point and I think it's something that is important to do. Since we're using the `CursorMoved` event only this is ensured by winit though.

> I'll gladly update / re-open a PR with your proposed changes at your request, or revert contains_point to filter out the padding from the pty event propagation.

Updating this PR would be much appreciated, thanks for your work and quick responses. :)

CHANGELOG.md
```diff
@@ -248,6 +248,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Empty lines in selections are now properly copied to the clipboard
 - Selection start point lagging behind initial cursor position
 - Rendering of selections which start above the visible area and end below it
+- Mouse press no longer ignored in padding width
```
chrisduerr: I think we could be a little more precise here. Since the mouse press is still ignored in regards to the PTY (as you've previously pointed out).

So I'd recommend explicitly stating that it is now possible to start selection from within the padding.

ladotonia: Made the comment more explicit. The change includes allowing following links in the first or last column by clicking in the padding (or other mouse interactions excluding pty propagation).

wkral:
Thanks very much for fixing this.

PR #2152, opened 2019-03-03
Add URL Highlighting

### LINKED ISSUE #1618: Highlight URLs
This is an extension of the work done in #1617.

Since #1617 implements everything necessary to open URLs when clicked on them, the basic requirements of #113 are now resolved. However it does not provide any indication of which URLs are recognized.

Ideally a URL would be underlined when it is recognized and the cursor shape should change to a pointer to indicate that an interaction is possible. However there are several different ways other terminal emulators do this, so as long as there's a clear indication that an URL has been recognized, it should suffice.

chrisduerr:
This PR adds URL highlighting to give better feedback to the user about
which text strings are recognized as an URL.

This includes changing the cursor to indicate that it is possible to
click URLs and underlining the URLs to show which cells are part of the
URL.

This fixes #1618.

Currently this is only an initial draft, there's still some things that need to be done:
 - [X] Change cursor when hovering over URLs
 - [x] Underline URLs
 - [x] Fix style/performance issues

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -407,7 +407,7 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         }
 
         // Underline URLs and change cursor on hover
-        self.update_url_highlight(point);
+        self.update_url_highlight(point, modifiers);
 
         if self.ctx.mouse().left_button_state == ElementState::Pressed
             && (modifiers.shift || !self.ctx.terminal().mode().intersects(report_mode))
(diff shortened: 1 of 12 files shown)


src/input.rs
```diff
@@ -425,6 +434,84 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+    /// Underline URLs and change the mouse cursor when URL hover state changes.
+    fn update_url_highlight(&mut self, point: Point, modifiers: ModifiersState) {
+        let mouse_mode =
+            TermMode::MOUSE_MOTION | TermMode::MOUSE_DRAG | TermMode::MOUSE_REPORT_CLICK;
+
+        // Only show URLs as launchable when all required modifiers are pressed
+        if self.mouse_config.url.modifiers.relaxed_eq(modifiers)
+            && (!self.ctx.terminal().mode().contains(TermMode::ALT_SCREEN) || modifiers.shift)
+        {
+            if let Some(Url { text, origin }) = self.ctx.terminal().url_search(point.into()) {
```
chrisduerr: This nesting and early `return` is pretty ugly but done for performance reasons. It would be nice to get the same effect with a little bit cleaner code.

chrisduerr:
Everything should be good to go now assuming that CI finishes.

src/input.rs
```diff
@@ -435,54 +435,63 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
 
-        if let Some(Url { text, origin }) = self.ctx.terminal().url_search(point.into()) {
-            let mouse_cursor = if self.ctx.terminal().mode().intersects(mouse_mode) {
-                MouseCursor::Default
-            } else {
-                MouseCursor::Text
-            };
+        // Only show URLs as launchable when all required modifiers are pressed
+        if self.mouse_config.url.modifiers.relaxed_eq(modifiers)
+            && (!self.ctx.terminal().mode().contains(TermMode::ALT_SCREEN) || modifiers.shift)
```
victorhsieh: Is this internal that in alt screen mode, shift has to be the modifier?  This seems to force me to use shift in tmux :)

BTW, thanks for the work!  I've been wanting this for a long time.

chrisduerr: I'm not sure what got over me here. This shouldn't check for `ALT_SCREEN`, but instead check for the mouse mode to be enabled. Thanks for pointing it out, current implementation makes no sense at all.

I'm going to include that into https://github.com/jwilm/alacritty/pull/2195.

Issue #34, opened 2017-01-02
Proper full screen support


jwilm:
Currently there's no full screen support in Alacritty. Doing to with `glutin` requires recreation of the window and thus recreation of the OpenGL context.

jwilm:
At least on macOS, there's the automagical full screen button that does the trick. It does occupy another workspace, however, and it doesn't allow stacking the full screen application on top of other apps.

davesque:
I can't get the fullscreen button to work on OS X 10.11.6.

jwilm:
@jimeh That's precisely what this issue is for :)

jimeh:
@jwilm Good good :)...  I just wanted to point out that the native full screen implementation on OSX sucks... lol

jedahan:
@jimeh @jwilm you can maybe take inspiration from https://github.com/mpv-player/mpv/pull/4676/files

caiocutrim:
Well, it was implemented? It seems that is not working in ubuntu when I hit f11...

jedahan:
@caiocutrim since the issue is still open I assume no one has taken a crack at it yet.

TaylorDennisLee:
For what it is worth, on Linux, you can install `wmctrl` which allows you to control windows from the terminal. In my .zshrc, I have an alias: `alias wq='wmctrl -r 'Alacritty' -b toggle,fullscreen'` which allows you to toggle fullscreen with the command `wq`. If you use tmux, you can bind this command to F11 by putting the following into your `.tmux.conf`:

`bind -n F11 run-shell 'wmctrl -r 'Alacritty' -b toggle,fullscreen'`

Notice that this works because 'Alacritty' is the title of the window. So, if your tmux config changes the title, you have to be aware of this and adjust either the new title or the key-binding/alias to match.

Also, I have found trying to invoke an alias with a tmux key-binding does not work, so just type in the entire command it references instead.

`wmctrl` is pretty legit. Type in `wmctrl  -l` to get a listing of all your windows to see what to target with it. You may be able to hit Alacritty in other ways besides with the window title. I don't know what the OS X equivalent it, but if we are preserving the 'purity' of Alacritty by farming out scrollback to tmux, we might consider trying to use wmctrl for as much functionality as we can.

What are your thoughts on binding shell commands to the keys in the Alacritty config? Can you do that with other terminal emulators?

[...]

PR #2157, opened 2019-03-06
Feature/mac os fullscreen

drew7721:
Judging by the latest comments on #34 seems I'm not the only one irritated by the fact that Alacritty does not start in _fullscreen_ mode. 

This changes the `start_maximized` config behaviour to start the app in full-screen on the current primary monitor (macos : where the mouse is located and spotlight bar is displayed).

I guess an additional configuration can be added for this. For now, I rather keep this simple.

I've tested this on latest macOs Mojave version and it works as expected.

Can't test how this change behaves on Windows but I will update this PR once tested.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -136,9 +136,14 @@ impl Window {
 
         let title = options.title.as_ref().map_or(DEFAULT_NAME, |t| t);
         let class = options.class.as_ref().map_or(DEFAULT_NAME, |c| c);
-        let window_builder = Window::get_platform_window(title, class, window_config, &event_loop);
+        let window_builder = Window::get_platform_window(title, class, window_config);
         let window = create_gl_window(window_builder.clone(), &event_loop, false)
             .or_else(|_| create_gl_window(window_builder, &event_loop, true))?;
+
+        if window_config.start_maximized() {
+            window.set_fullscreen(Some(event_loop.get_primary_monitor()));
+        }
+
         window.show();
 
         // Text cursor
@@ -251,8 +256,7 @@ impl Window {
     pub fn get_platform_window(
         title: &str,
         class: &str,
-        window_config: &WindowConfig,
-        event_loop: &EventsLoop
+        window_config: &WindowConfig
     ) -> WindowBuilder {
         use glutin::os::unix::WindowBuilderExt;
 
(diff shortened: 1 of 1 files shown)


drew7721:
hey @chrisduerr, I completely agree with you, it should not be introduced as a regression. But do you think it would be a good feature to have? 

Initially, I did this just to implement the feature for myself and give a solution to those that were looking for one, other than creating an AppleScript. But, I agree it would be better to create a new config for it on the same level as the *start_maximized*. 

I'll try to update the PR as soon as I have some time if you rule that it's worth having this option. Let me know.

Also, know that this does start on the _active_ screen even if the method is named _primary_ screen (at least it does so on mac os). Not sure about how it behaves on Windows, yet. I'll have to test and see if it behaves the same.

chrisduerr:
> But do you think it would be a good feature to have?

Yes. This was introduced fully aware of the difference between maximized and fullscreen and people have explicitly requested both.

It does not necessarily have to be a a separate config option either, since we can make use of enums. Something like `start_state: Maximized|Fullscreen|Windowed` should work perfectly fine, though we'd have to make sure `start_maximized: true` is mapped automatically for backwards compatability.

> Also, know that this does start on the active screen even if the method is named primary screen (at least it does so on mac os). Not sure about how it behaves on Windows, yet. I'll have to test and see if it behaves the same.

While `set_fullscreen` does not work on my Linux machine, it does work on Windows and there it definitely does not start the window on the active monitor.

src/window.rs
```diff
@@ -139,6 +139,11 @@ impl Window {
         let window_builder = Window::get_platform_window(title, class, window_config);
         let window = create_gl_window(window_builder.clone(), &event_loop, false)
             .or_else(|_| create_gl_window(window_builder, &event_loop, true))?;
+
+        if window_config.start_maximized() {
+            window.set_fullscreen(Some(event_loop.get_primary_monitor()));
```
chrisduerr: Another issue is that `set_fullscreen` does not correctly spawn the Window in fullscreen mode, but rather spawns it in windowed mode and then fullscreens it. The `with_fullscreen` method should be used instead.

chrisduerr:
Thanks for your work, but it looks like https://github.com/jwilm/alacritty/pull/2308 addresses the issues mentioned in this PR while also introducing a few more features.

So I'm closing this in favor of the #2308 PR, which should hopefully be merged pretty soon.

Issue #2069, opened 2019-02-03
Add option to override colour of selected text


awused:
Right now selected text always inverts the colours, but this isn't always desirable, especially when the background and foreground colours are similar. It'd be nice to be able to override that so that text can always be made legible on selection.

A similar setting already exists for cursor colours, this would function equivalently for selected text.

chrisduerr:
Seems reasonable to have something in line with what the already have for cursor colors, since both behave very similar anyways.

MarioRicalde:
@chrisduerr isn't this closed by #2159 ?

chrisduerr:
It is, thanks!

PR #2159, opened 2019-03-07
Added the ability to change the selection color.

kchibisov:
Now selection color and selection text color can be specified in config.

Implements [#2069](https://github.com/jwilm/alacritty/issues/2069)

### DIFF
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -543,10 +543,6 @@ pub enum NamedColor {
     CursorText,
     /// Color for the cursor itself
     Cursor,
-    /// Color for the selected text
-    SelectionText,
-    /// Color for the selection itself
-    Selection,
     /// Dim black
     DimBlack,
     /// Dim red
@@ -567,6 +563,10 @@ pub enum NamedColor {
     BrightForeground,
     /// Dim foreground
     DimForeground,
+    /// Color for the selected text
+    SelectionText,
+    /// Color for the selection itself
+    Selection,
 }
 
 impl NamedColor {
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -566,7 +566,7 @@ pub enum NamedColor {
     /// Color for the selected text
     SelectionText,
     /// Color for the selection itself
-    Selection,
+    SelectionBackground,
 }
 
 impl NamedColor {
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -563,10 +563,6 @@ pub enum NamedColor {
     BrightForeground,
     /// Dim foreground
     DimForeground,
-    /// Color for the selected text
-    SelectionText,
-    /// Color for the selection itself
-    SelectionBackground,
 }
 
 impl NamedColor {
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -543,6 +543,10 @@ pub enum NamedColor {
     CursorText,
     /// Color for the cursor itself
     Cursor,
+    /// Color for the selected text
+    SelectionText,
+    /// Color for the selection itself
+    Selection,
     /// Dim black
     DimBlack,
     /// Dim red
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -1285,6 +1285,8 @@ pub struct Colors {
     pub primary: PrimaryColors,
     #[serde(deserialize_with = "failure_default")]
     pub cursor: CursorColors,
+    #[serde(deserialize_with = "failure_default")]
+    pub selection: SelectionColors,
     #[serde(deserialize_with = "deserialize_normal_colors")]
     pub normal: AnsiColors,
     #[serde(deserialize_with = "deserialize_bright_colors")]
@@ -1300,6 +1302,7 @@ impl Default for Colors {
         Colors {
             primary: Default::default(),
             cursor: Default::default(),
+            selection: Default::default(),
             normal: default_normal_colors(),
             bright: default_bright_colors(),
             dim: Default::default(),
@@ -1419,6 +1422,15 @@ pub struct CursorColors {
     pub text: Option<Rgb>,
     #[serde(deserialize_with = "deserialize_optional_color")]
     pub cursor: Option<Rgb>,
+}
+
+#[serde(default)]
+#[derive(Debug, Copy, Clone, Default, Deserialize, PartialEq, Eq)]
+pub struct SelectionColors {
+    #[serde(deserialize_with = "deserialize_optional_color")]
+    pub text: Option<Rgb>,
+    #[serde(deserialize_with = "deserialize_optional_color")]
+    pub selection: Option<Rgb>,
 }
 
 #[serde(default)]
@@ -1854,6 +1866,18 @@ impl Config {
     #[inline]
     pub fn cursor_cursor_color(&self) -> Option<Color> {
         self.colors.cursor.cursor.map(|_| Color::Named(NamedColor::Cursor))
+    }
+
+    /// Selected text color.
+    #[inline]
+    pub fn selection_text_color(&self) -> Option<Color> {
+        self.colors.selection.text.map(|_| Color::Named(NamedColor::SelectionText))
+    }
+
+    /// Selected text background color.
+    #[inline]
+    pub fn selection_selection_color(&self) -> Option<Color> {
+        self.colors.selection.selection.map(|_| Color::Named(NamedColor::Selection))
     }
 
     /// Enable experimental conpty backend (Windows only)
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -1876,7 +1876,7 @@ impl Config {
 
     /// Selected text background color.
     #[inline]
-    pub fn selection_selection_color(&self) -> Option<Color> {
+    pub fn selection_background_color(&self) -> Option<Color> {
         self.colors.selection.selection.map(|_| Color::Named(NamedColor::Selection))
     }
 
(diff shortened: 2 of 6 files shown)


kchibisov:
Hi @chrisduerr . Sorry if i give you any inconvenience, but could you provide some status on this PR ( any notes, or mark as delayed for review). Thanks and sorry again.

chrisduerr:
I will review it once I find some time.

src/config/mod.rs
```diff
@@ -1847,6 +1859,18 @@ impl Config {
...
 
+    /// Selected text color.
+    #[inline]
+    pub fn selection_text_color(&self) -> Option<Color> {
+        self.colors.selection.text.map(|_| Color::Named(NamedColor::SelectionText))
+    }
+
+    /// Selected text background color.
+    #[inline]
+    pub fn selection_selection_color(&self) -> Option<Color> {
```
chrisduerr: This name is pretty bad, I think `selection_background_color` would be the right choice here.

src/term/mod.rs
```diff
@@ -453,13 +453,30 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
...
+                        cell.bg
+                    }),
+                self.config.selection_text_color()
+                    .unwrap_or_else(|| cell.fg))
+            } else {
+                (cell.bg, cell.fg)
+            };
 
-            let bg_alpha = if selected ^ cell.inverse() {
+            // Apply inversion (if required) and lookup RGB values.
```
chrisduerr: This comment is a bit confusing, since the immediately following block just looks up the RGB values.

The `Apply inversions if required` comment should be added to the block which actually does the inversions (the `let bg_alpha` one).

src/term/mod.rs
```diff
@@ -453,13 +453,30 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
                 (cell, selected)
             };
 
-            // Apply inversion and lookup RGB values
-            let mut fg_rgb = self.compute_fg_rgb(cell.fg, &cell);
-            let mut bg_rgb = self.compute_bg_rgb(cell.bg);
+            let mut invert_selection = false;
```
chrisduerr: It's probably cleaner to change this to the following:

```rust
let selection_background_color = self.config.selection_background_color();
let mut invert_selection = selected && selection_background_color.is_none();
```

src/term/mod.rs
```diff
@@ -453,25 +453,24 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
                 (cell, selected)
             };
 
-            let mut invert_selection = false;
             // Set custom selection color and selection text color, if they were provided.
```
chrisduerr: This comment should be moved to the `let (bg_color, fg_color) = if selected { ...` comment.

src/term/mod.rs
```diff
@@ -453,25 +453,24 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
...
+            let selection_background_color = self.config.selection_background_color();
```
chrisduerr: Moving this under the selection color comment and over the `let invert_selection` statement would probably make things look a little cleaner after all the other suggestions are applied.

src/config/mod.rs
```diff
@@ -1417,6 +1420,15 @@ pub struct CursorColors {
     pub cursor: Option<Rgb>,
 }
 
+#[serde(default)]
+#[derive(Debug, Copy, Clone, Default, Deserialize, PartialEq, Eq)]
+pub struct SelectionColors {
+    #[serde(deserialize_with = "deserialize_optional_color")]
+    pub text: Option<Rgb>,
+    #[serde(deserialize_with = "deserialize_optional_color")]
+    pub selection: Option<Rgb>,
```
chrisduerr: This should probably be changed to `background` too.

src/ansi.rs
```diff
@@ -563,6 +563,10 @@ pub enum NamedColor {
     BrightForeground,
     /// Dim foreground
     DimForeground,
+    /// Color for the selected text
+    SelectionText,
+    /// Color for the selection itself
+    Selection,
```
chrisduerr: To be in line with all other variables, this should be `SelectionBackground`, which is a little more explicit too.

src/term/mod.rs
```diff
@@ -453,13 +453,28 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
...
-            let mut fg_rgb = self.compute_fg_rgb(cell.fg, &cell);
-            let mut bg_rgb = self.compute_bg_rgb(cell.bg);
+            // If selection color wasn't set by user we should invert cell colors.
+            let selection_background_color = self.config.selection_background_color();
+            let invert_selection = selected && selection_background_color.is_none();
+
+            // Set custom selection color and selection text color, if they were provided.
+            let (bg_color, fg_color) = if selected {
+                (selection_background_color.unwrap_or_else(|| cell.bg ),
+                    self.config.selection_text_color().unwrap_or_else(|| cell.fg))
```
chrisduerr: The formatting here is a bit messy, I propose doing it like this:

```rust
(
    selection_background_color.unwrap_or_else(|| cell.bg),
    self.config.selection_text_color().unwrap_or_else(|| cell.fg),
)
```

This should probably feel a bit more uniform and I think it's probably more like what rustfmt would do by itself.

kchibisov:
The current behavior ( in PR ) is like: apply selection text color to foreground, use default background and then swap them. (e.g. on picture)

Screenshot-Mon_11_Mar_2019_05

I did this on purpose, since i thought that the next variant is a little bit strange, since there wouldn't be any visual selection on the empty background.

Screenshot-Mon_11_Mar_2019_33

Although the behavior may be the same as with the cursor.

Screenshot-Mon_11_Mar_2019_34

So, which one should be implemented (1,2 or 3)?  But now for me the right behavior is the second one, since I realized that the 1st one can be implemented just by setting text + background for selection.

chrisduerr:
> So, which one should be implemented (1,2 or 3)?

None of the above. I'm not exactly sure if I understand what you're saying here, but if just the text color is set, just the text color should be changed. The rest should behave exactly the same way it does with selection by default - inverting the foreground to the background color.

kchibisov:
Sorry if I confused you, you’re suggesting the third one ( the block cursor will behave the way you describe, if its background color is missing ). Ok, I’ll do it.

src/term/mod.rs
```diff
@@ -455,14 +455,20 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
...
 
             // Set custom selection color and selection text color, if they were provided.
             let (bg_color, fg_color) = if selected {
-                (
-                    selection_background_color.unwrap_or_else(|| cell.bg ),
-                    self.config.selection_text_color().unwrap_or_else(|| cell.fg),
-                )
+                // If selection text color was set by the user we shouldn't invert it.
+                if invert_selection && selection_text_color.is_some() {
+                    (selection_text_color.unwrap(), cell.fg)
```
chrisduerr: We should use the `let` pattern here to skip the unwrap.

kchibisov: Are you talking about ```if let```? In this case it was ignored because rust does not allow writing
```rust
if let Some(x) = x && true
```

chrisduerr: Nested `if`s would still be better than unwraps which might lead to bugs down the road.

If nesting is not desired, there's also a workaround that can be applied:

```
if let ((true, Some(value))) = (invert_selection, selection_text_color.is_some())
```

kchibisov: What do you think about this one?
```rust
let (bg_color, fg_color) = if selected {
    match (selection_background_color, self.config.selection_text_color()) {
        (None, None) => (cell.bg, cell.fg),
        (None, Some(fg_color)) => (fg_color, cell.fg),
        (Some(bg_color), None) => (bg_color, cell.fg),
        (Some(bg_color), Some(fg_color)) => (bg_color, fg_color),
    }
} else {
    (cell.bg, cell.fg)
};
```

chrisduerr: There's a lot of duplication in this one. 4/5 times the `fg_color` is just `cell.fg` for example, so that doesn't look like a great choice.

I think your current solution might actually make things harder on you than it has to be, I've played around with it myself for a bit and the following code worked great:

```rust
// Apply inversion and lookup RGB values
let mut fg_rgb = self.compute_fg_rgb(cell.fg, &cell);
let mut bg_rgb = self.compute_bg_rgb(cell.bg);

let selection_background = self.config.colors().selection.background;
let bg_alpha = if let (true, Some(col)) = (selected, selection_background) {
    bg_rgb = col;
    1.0
} else if selected ^ cell.inverse() {
    mem::swap(&mut fg_rgb, &mut bg_rgb);
    self.compute_bg_alpha(cell.fg)
} else {
    self.compute_bg_alpha(cell.bg)
};

if let (true, Some(col)) = (selected, self.config.colors().selection.text) {
    fg_rgb = col;
}
```

This would also make it possible to just remove all code from `color.rs` and remove the methods from the `config/mod.rs`. I think it would significantly clean things up here.

kchibisov: I like your idea, since it's much easier to maintain. In my PR I just wanted to be consistent with the existing code for cursor colors specification ( e.g. method `selection_selection_color` in the first version of this PR  was named after [`cursor_cursor_color`](https://github.com/jwilm/alacritty/blob/e240da9ab3b819a8845ced1ab72d0a4239eac789/src/config/mod.rs#L1855)). But if you consider such a difference between implementations in almost identical things, then I would prefer your option.

chrisduerr: I don't think the named colors have a slightly different background and are mainly focused on the traditional 256 (/257/88) terminal colorspace.

Going for the simple solution here should be fine.

kchibisov: Ok. So, waiting for tests and your review.

PR #2162, opened 2019-03-08
Use /compat/linux/proc in SpawnNewInstance on FreeBSD

t6:
This fixes opening a new instance in the shell's current working
directory.  The code currently assumes that there is a Linux
compatible procfs mounted on /proc which is not the case on FreeBSD.
However linprocfs(5) is often mounted on /compat/linux/proc so we
can use that here for the time being.

A proper solution would look up the kern.proc.cwd.<pid> sysctl
instead, but that is not currently easily doable due to missing
KERN_PROC_CWD and struct kinfo_file definitions in the libc crate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Terminfo support for extended capabilities
 - Allow mouse presses and beginning of mouse selection in padding
 - Windows: Conpty backend could close immediately on startup in certain situations
+- FreeBSD: SpawnNewInstance will now open new instances in the shell's current working directory as long as linprocfs(5) is mounted on `/compat/linux/proc`
 
 ## Version 0.2.9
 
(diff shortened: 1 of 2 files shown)


chrisduerr:
How commonly is this filesystem actually mounted? I'd assume that this has to be explicitly done by the user?

This seems like a pretty cheap workaround that could make it work for some people at least, but if it's never mounted but just exists in theory there's obviously not much benefit to it.

t6:
Yes, it has to be mounted explicitly by users.  Anyone who uses linux(4) to run some Linux applications on FreeBSD already has to do it.  Some ported applications are also patched to look in /compat/linux/proc.  It's probably mounted commonly enough.

t6:
> Is this a thing on other BSD systems too? In that case it would be possible to make it work under OpenBSD/...?

I don't know. I'm unfamiliar with the other BSDs.

> This still needs a change log entry before it can be merged.

Added.

t6:
Thank you.

PR #2163, opened 2019-03-08
Configurable window position

### LINKED ISSUE #41: Add config for window starting position
Gnome terminal and others support specifying a starting geometry for the screen, usually in terms of chars. So "200x50" is 200 vertical chars and 50 horizontal ones. That plus a pixel offset from top-left of screen lets the user decide how big should the alacritty window be and where should it be placed. With this in config, the window is always where I want it.

Something like this:

```yaml
window:
  vertical_chars: 50
  horizontal_chars: 100
  starting_x_offset: 100
  starting_y_offset: 100
```

You could put the `fullscreen` property (from #38) on the same `window` block too.

cole-h:
Adds the ability to specify window geometry through either a config option or
command line flag. Geometry is set early to avoid "teleporting" the window from
its initial position (typically 0x0) to its configured position.

(Somewhat) resolves #41 -- Ideally, at some point this should become POSIX compliant à la `COLUMNSxROWS+X+Y`. For now, a combination of `dimensions` and `geometry` is a somewhat competent workaround.

-----

This is my first "substantial" contribution to this project and I'm new to the code base, so I would appreciate any feedback on style and the like. I tested this on my local machine and nothing seemed out of the ordinary. I mostly copied the code present for setting the dimensions of the window and modified it as necessary.

The one thing I'm not very happy with is the `set_position` function in `src/window.rs` -- it looks messy in my eyes. If there is a better or more majestic way to do what I did, I would appreciate the tip.

Should I add the window geometry to the `info!` call here: https://github.com/jwilm/alacritty/blob/master/src/main.rs#L136?

-----

Current problems:

- [x] FIXED: Terminal now spawns at absolute 0x0, as opposed to 0x0 of the current screen.
- [x] FIXED: Terminal does not respect DPI settings set via `WINIT_HIDPI_FACTOR` env var


cole-h:
Thoughts on making geometry configurable via the flag only, rather than by flag _and_ config file? Now that I think about it, it doesn't make sense (to me) to be able to set the position of the window via the config file. Removing that part also makes it easier to decide whether or not to respect the specified geometry: if the flag is present, use it. Otherwise, respect current behavior of letting the window manager take care of it.

Unless there's any dissent on removing the config file stuff, I think I'll go ahead and do just that in the next push.

alacritty.yml
```diff
@@ -21,6 +21,14 @@ window:
     columns: 0
     lines: 0
 
+  # Window geometry (changes require restart)
+  #
+  # Specified in number of pixels.
+  # If both are `0`, this setting is ignored.
+  geometry:
+    x: 0
+    y: 0
```
chrisduerr: This should *not* be ignored when both are `0`, since `0x0` is not the default window position.

I think the best option here would be to just have `geometry` commented out by default. While that's usually a bad choice because there's no way to explicitly disable it, I think it might be better than something like `geometry: Default`.

src/config/mod.rs
```diff
@@ -1988,6 +2003,40 @@ impl Dimensions {
...
+    #[inline]
+    pub fn x(self) -> i32 {
+        self.x
+    }
+
+    /// Get y-pos
+    #[inline]
+    pub fn y(self) -> i32 {
+        self.y
+    }
```
chrisduerr: I think it just making these x and y fields public is a better choice than having getters here.

alacritty.yml
cole-h: Is everything in `config.rs` required to have a partner in `alacritty.yml`? If not, what's your opinion on not letting this be configurable by the config file at all? Frankly, it makes everything a whole lot easier to implement, and I can't think of a time when you'd want to set it from there.

chrisduerr: > Is everything in config.rs required to have a partner in alacritty.yml?

Yes.

> If not, what's your opinion on not letting this be configurable by the config file at all?

Not an option at all.

> it makes everything a whole lot easier to implement

I don't think that's the right thing to focus on. Even with a configuration option this is a fairly straight-forward feature. Skimping out on it and putting the load of working around it on the user is not the right move here.

cole-h:
Changed names to more accurately reflect that we're setting position, not geometry (for now). Commented out `position` in config by default, make `position` field in `WindowConfig` an option (is there a better way you would prefer I do this?). Let window manager handle positions if no flag or config specifies it. Marked `set_position` uses with `TODO` so we don't forget to use `with_position` if/when it becomes available.

Brain was working a little slow this morning, hopefully I've addressed most/all of the issues.

alacritty.yml
```diff
@@ -21,6 +21,14 @@ window:
     columns: 0
     lines: 0
 
+  # Window position (changes require restart)
+  #
+  # Specified in number of pixels.
+  # Comment out to let the window manager handle placement.
```
chrisduerr: To fall in line with the other documentation comments about commenting-out values, I'd reword this to the following:

```
If the position is not set, the window manager will handle the placement.
```

src/config/mod.rs
```diff
@@ -414,6 +414,10 @@ pub struct WindowConfig {
     #[serde(default, deserialize_with = "failure_default")]
     dimensions: Dimensions,
 
+    /// Initial position
+    #[serde(default, deserialize_with = "failure_default")]
+    position: Option<Position>,
+
```
chrisduerr: Having the `dimensions` in the global `Config` is actually deprecated, I'm not sure why it hasn't been moved to the bottom and marked as such, but it will print a warning when it's used.

So the position should not be added here. I'd also appreciate it if you could move the `dimensions` to the bottom where the other items are and mark it with a deprecation comment, so this issue will not occur again.

src/display.rs
```diff
@@ -150,6 +150,13 @@ impl Display {
         // Create renderer
         let mut renderer = QuadRenderer::new(viewport_size)?;
 
+        // TODO: replace `set_position` with `with_position` once available
+        // Set window position early so it doesn't "teleport"
+        let position = options.position().or_else(|| config.position());
+        if let Some(position) = position {
+            window.set_position(position.x, position.y);
```
chrisduerr: Is there any reason why the position isn't set right after window initialization?

src/window.rs
```diff
@@ -182,6 +182,11 @@ impl Window {
         self.window.set_inner_size(size);
     }
 
+    // TODO: use `with_position` once available
+    pub fn set_position(&mut self, x: i32, y: i32) {
+        self.window.set_position(LogicalPosition::from((x, y)));
```
chrisduerr: The x/y coordinates in the config and CLI are in physical units, to get the expected behavior for the user we need to convert them to logical units using the DPR. So you should be creating a `PhysicalPosition` with `x` and `y` here and then using `dpr` to convert it to a `LogicalPosition` (there's a method on `PhysicalPosition` to do that I believe).

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - MSI installer for Windows is now available
 - New default key bindings Alt+Home, Alt+End, Alt+PageUp and Alt+PageDown
 - Dynamic title support on Windows
+- Ability to specify starting position with the `--position` flag
```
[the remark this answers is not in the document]
cole-h: Should it be in the same entry, or should I make a new one below it?

src/config/mod.rs
cole-h: Should I add a changelog entry to document this? Or is the already-present `print_deprecation_warnings` enough?

CHANGELOG.md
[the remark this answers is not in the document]
chrisduerr: I'd create a separate one to keep them short and concise.

src/config/mod.rs
chrisduerr: This is just a code change that does not affect the user. The warning is already present. So there's no need for a change log entry.

src/display.rs
cole-h: Moved right under `let mut window`. Hopefully that's what you meant.

src/window.rs
```diff
@@ -182,6 +182,15 @@ impl Window {
         self.window.set_inner_size(size);
     }
 
+    // TODO: use `with_position` once available
+    pub fn set_position(&mut self, x: i32, y: i32) {
+        let dpr = self.window.get_hidpi_factor();
+        let physical = PhysicalPosition::from((x, y));
+        let logical = physical.to_logical(dpr);
```
chrisduerr: This is a tiny nitpick, but I think it's a bit too verbose (too many variables that aren't reused, but just temporary).

Something like this should work a bit nicer:
```rust
let logical = PysicalPosition::from((x, y)).to_logical(self.window.get_hidpi_factor());
self.window.set_position(logical);
```

I believe this would still be under the 100 column limit, so it shouldn't be a problem. Alternatively just declaring a separate variable for `dpr` would be enough and we could at least drop the `physical` variable.

src/config/mod.rs
```diff
@@ -1988,6 +1997,28 @@ impl Dimensions {
...
+#[derive(Default, Debug, Copy, Clone, Deserialize, PartialEq, Eq)]
+pub struct Position {
+    /// Window x-position
+    #[serde(deserialize_with = "failure_default")]
+    pub x: i32,
+
+    /// Window y-position
+    #[serde(deserialize_with = "failure_default")]
+    pub y: i32,
+}
```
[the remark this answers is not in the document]
cole-h: Using `Delta` does seem like a good idea.

>It might make sense to rename `Delta`

Do you mean renaming `Delta` to `Position`, or just renaming `Delta` in general?

Negative positions do indeed spawn the window outside of the screen (at least on Linux), so I'll leave that as is.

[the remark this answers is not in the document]
cole-h: Also, negative positions cannot be passed to the `--position` flag, but setting `AllowLeadingHyphen` fixes that. I'm assuming we want users to be able to pass negative values, so should `ArgSettings` be `use`d at the top or should I just `.set(clap::ArgSettings::AllowLeadingHyphen)`?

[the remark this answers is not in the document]
chrisduerr: > Do you mean renaming Delta to Position, or just renaming Delta in general?

I'm talking about renaming `Delta` in general. I'd say `Vector` since that's commonly used to store dimensional positions, but with `Vec` that namespace is kinda already occupied.

Since `padding` also uses a `Delta` even though `Delta` might not be the greatest name, I'm almost inclined to just keep the name though. If you have a good idea, I'm all ears, otherwise I'd just use Delta for now.

> Also, negative positions cannot be passed to the --position flag, but setting AllowLeadingHyphen fixes that. I'm assuming we want users to be able to pass negative values, so should ArgSettings be used at the top or should I just .set(clap::ArgSettings::AllowLeadingHyphen)?

Usually importing at the top of the file is a good idea, just makes things a lot cleaner in the actual code.

Does this create any issues with single-character flags though? We wouldn't want to introduce a regression and there has to be a reason why it doesn't work by default.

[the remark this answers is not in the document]
cole-h: I'll leave `Delta` alone for now, then.

The only issue it has from my brief testing is, if you don't specify x and y coordinates (i.e. running `cargo run -- --position -e bash`), it eats the argument after it. Without the setting, it properly errors out notifying the user that `--position` requires 2 values.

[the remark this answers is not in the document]
chrisduerr: > The only issue it has from my brief testing is, if you don't specify x and y coordinates (i.e. running cargo run -- --position -e bash), it eats the argument after it. Without the setting, it properly errors out notifying the user that --position requires 2 values.

I think we can just ignore this issue as user error. I don't think there is any well defined 'best option' to go for here.

> If we do go forward with allowing hyphen values despite that issue, I'd use .allow_hyphen_values(true), which does the same thing without needing use clap::ArgSettings

Sounds good.

src/display.rs
```diff
@@ -140,6 +140,13 @@ impl Display {
         // Create the window where Alacritty will be displayed
         let mut window = Window::new(&options, config.window())?;
 
+        // TODO: replace `set_position` with `with_position` once available
+        // Set window position early so it doesn't "teleport"
```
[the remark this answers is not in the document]
cole-h: If/once this gets merged, I'll submit the tracking issue.

cole-h:
Thanks for bearing with me (and fixing up those last bits of mess)!

PR #2165, opened 2019-03-08
Adds keybinding for 'de' layout keyboards #2010

### LINKED ISSUE #2010: increas font keybinding not working
arch linux

X11 

using `alacritty-git` from aur

executing `setxkbmap de`
decreasing and resettings works, but increasing does not.

amgoodfellow:
This closes #2010  by implementing the suggestion made by @chrisduerr to include the IncreaseFontSize keybinding for the 'Add' key by default.

I tested this on MacOS and Arch Linux, and it seems to work. All of the cargo tests pass as well.

Also, I realize there's not a lot going on in this PR, but it's my first 'real' one so any suggestions or critiques about the code or the message would be most welcome! Thanks :)

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -466,11 +466,13 @@ key_bindings:
   #- { key: Insert,   mods: Shift,         action: PasteSelection   }
   #- { key: Key0,     mods: Control,       action: ResetFontSize    }
   #- { key: Equals,   mods: Control,       action: IncreaseFontSize }
+  #- { key: Add,      mods: Control,       action: IncreaseFontSize }
   #- { key: Subtract, mods: Control,       action: DecreaseFontSize }
 
   # (macOS only)
   #- { key: Key0,   mods: Command, action: ResetFontSize    }
   #- { key: Equals, mods: Command, action: IncreaseFontSize }
+  #- { key: Add,    mods: Control, action: IncreaseFontSize }
   #- { key: Minus,  mods: Command, action: DecreaseFontSize }
   #- { key: K,      mods: Command, action: ClearHistory     }
   #- { key: K,      mods: Command, chars: "\x0c"            }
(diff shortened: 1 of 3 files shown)


alacritty.yml
```diff
@@ -449,11 +449,13 @@ key_bindings:
   #- { key: Insert,   mods: Shift,         action: PasteSelection   }
   #- { key: Key0,     mods: Control,       action: ResetFontSize    }
   #- { key: Equals,   mods: Control,       action: IncreaseFontSize }
+  #- { key: Add,      mods: Control,       action: IncreaseFontSize }
   #- { key: Subtract, mods: Control,       action: DecreaseFontSize }
 
   # (macOS only)
   #- { key: Key0,   mods: Command, action: ResetFontSize    }
   #- { key: Equals, mods: Command, action: IncreaseFontSize }
+  #- { key: Add,    mods: Control, action: IncreaseFontSize }
```
chrisduerr: On macOS we need to use `Command` here.

amgoodfellow: Ope, sorry about that. I just pushed a new commit that should fix this - I also made the same mistake in `bindings.rs`

dnkl:
For what it's worth; this is a nice default for the `se` layout too.

dnkl:
Btw, since this is slightly related; for me (`se` layout) `Subtract` (for font _decrease_) doesn't work, but `Minus` does.

chrisduerr:
I'd be fine with adding a `Minus` binding too, it's already the case for macOS.

Otherwise just rebasing this should be enough and it can be merged.

amgoodfellow:
Hi guys - sorry about the force push - if I was smart I would've rebased and _then_ pushed the `Minus` keybinding stuff. But I wasn't thinking about it at the time.

chrisduerr:
> Hi guys - sorry about the force push

No worries, there's no way around that when rebasing.

I think everything's looking good now, thanks for the PR.

PR #2168, opened 2019-03-10
config and term: introduce explicit selection color.

matttproud:
This commit introduces a new feature for Alacritty wherein the user may
request explicit text and background colors for selection ranges.  If
no color is specified, the behavior falls back on the Alacritty factory
original of using the inverted fore- and background colors for the
current context.

This had been something I was pining for in the last two years of using
Alacritty.  Looked into how hard it would be to implement and seemed
easy enough.  Please let me know what your feedback is.

Screenshot_20190310_125724

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -177,6 +177,14 @@ colors:
   #  text: '0x000000'
   #  cursor: '0xffffff'
 
+  # Selection colors
+  #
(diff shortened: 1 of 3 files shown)


chrisduerr:
Duplicate of https://github.com/jwilm/alacritty/pull/2159.

PR #2175, opened 2019-03-14
Update glutin to version 0.20

chrisduerr:
Fixes https://github.com/jwilm/alacritty/issues/458.
Fixes https://github.com/jwilm/alacritty/issues/1681.


atouchet:
I think cargo.lock needs to be re-updated. The image crate no longer matches what's in Cargo.toml.

chrisduerr:
@atouchet The lockfile has been updated. We're now actually depending on both image files, one likely being the dependency for a different crate.

Issue #923, opened 2017-11-28
Complete implementation of scrollback


shreve:
Scrollback (#657) is nearly complete, but does still require some heavy lifting to get it merged into master.

Bountysource won't let me put a bounty on a pull request, so I've made this issue. 

[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=52061265)](https://www.bountysource.com/issues/52061265-complete-implementation-of-scrollback?utm_source=52061265&utm_medium=shield&utm_campaign=ISSUE_BADGE)

buster:
👍 Added 15$. I really hope someone get's this merged, i don't need or want tmux in my local terminal.

lancelet:
+$15 from me. I've been using alacritty for a while and would like to have scrollback too.

ght:
As a workaround you could use a terminal multiplexer like [tmux](/tmux/tmux), which supports scrollback (you might need to `set-option -g mouse on`) and has several other advantages.

buster:
I use that workaround and it's a major hazzle. I see the point of tmux and screen on remote and plain text systems but not at all in my desktop environment. I'm fine when other people like to use tmux all day but please don't try to sell it to me as "advantage". For you it may be, for me it is not. 

I'm aware that alacritty was supposed to be used with tmux from the beginning. I'd like to use it without tmux and put some money behind it. That's how painful using tmux is for me. I'm paying money whereas i could just use one of the gazillion other shells.

Still, i hope  someone will find the time to do that :)

unphased:
I can't live without tmux! @buster, could you share some of your reasons why you prefer not to use tmux? I'm just really curious!

lord-re:
For me tmux is excellent throught ssh, but i locally i prefer opening multiple terminals and arranging easily with my windows manager rather than learning how to arrange them in tmux. 
I also don't use tabs in my web browser for the exact same reason : i prefer managing them through the wm rather than learning how to move them in the browser… 
Idem for (neo)vi(m), i don't use multiple buffers but multiple terminals with multiple text editor…

unphased:
vim -> windows/panes in one instance lets your search highlights work in all of the open files, copy & paste across them, etc. Also reduce process ram usage overhead.

[...]

PR #2186, opened 2019-03-16
Add ScrollLineUp and ScrollLineDown actions for scrolling line by line

enjmiah:
It was mentioned in #923 that there is currently no way to scroll line by line using the keyboard.  This PR adds two actions, `ScrollLineUp` and `ScrollLineDown`, which can be optionally bound to keyboard keys by the user.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -438,6 +438,8 @@ alt_send_esc: true
 #   - ResetFontSize
 #   - ScrollPageUp
 #   - ScrollPageDown
+#   - ScrollLineUp
+#   - ScrollLineDown
 #   - ScrollToTop
 #   - ScrollToBottom
 #   - ClearHistory
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added the ability to change the selection color
 - Text will reflow instead of truncating when resizing Alacritty
 - Underline text and change cursor when hovering over URLs with required modifiers pressed
+- Added ScrollLineUp and ScrollLineDown actions for scrolling line by line
 
 ### Changed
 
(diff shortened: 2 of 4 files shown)


alacritty.yml
```diff
@@ -518,12 +520,14 @@ key_bindings:
   - { key: Right,    mods: Alt,     chars: "\x1b[1;3C"                       }
   - { key: Right,                   chars: "\x1b[C",        mode: ~AppCursor }
   - { key: Right,                   chars: "\x1bOC",        mode: AppCursor  }
-  - { key: Up,       mods: Shift,   chars: "\x1b[1;2A"                       }
+  - { key: Up,       mods: Shift,   action: ScrollLineUp,   mode: ~Alt       }
```
chrisduerr: We can *not* bind these by default, since they are escape sequences which can be read by terminal applications.

enjmiah: Ah, okay.  Fixed.  (It's <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Up / Down</kbd> now.)

chrisduerr: This is not standard functionality. While some people might be interested in scrolling line-by-line (which is fine), it should not be bound by default.

enjmiah: Sorry for the delayed response.

The following terminal emulators bind keys to line-by-line scrolling by default:

+ xfce4-terminal binds Ctrl + Shift + Up/Down
+ LXTerminal binds Ctrl + Shift + Up/Down
+ GNOME Terminal binds Ctrl + Shift + Up/Down
+ kitty binds Ctrl + Shift + Up/Down
+ tilda binds Ctrl + Shift + Up/Down
+ Terminator binds Ctrl + Shift + Alt + Up/Down
+ Konsole binds Shift + Up/Down

The only terminal emulator I tested that didn't have a keybinding for line-by-line scrolling is XTerm.  I don't think having a keybinding for line-by-line scrolling by default is uncommon.  I think users will appreciate if something like this works out of the box.  That being said, if you still want to remove the keybinding, then I'll remove it.

chrisduerr: URxvt also does not set any binding by default.

I think this topic is too opinionated to find a good default binding which would still be easy to trigger without breaking workflows of some people.

Generally it shouldn't be a huge issue to set these up yourself if you're interested in them.

enjmiah: Fair enough; fixed.  :)

PR #2192, opened 2019-03-19
Fix URLs getting incorrectly extended to next line

### LINKED ISSUE #1906: alacritty-0.2.4:  URLs can be incorrectly extended onto the next line
Which operating system does the issue occur on?    Probably any
If on linux, are you using X11 or Wayland?   X11, but I don't think that's relevant

There is a minor bug in the URL grabber.  I'm running irssi in an alacritty, and this is an excerpt of my session:

13:17  * Grimes examines the latest issue of Stars and Stripes
13:22 < ToCoTox> https://www.geologyin.com/2018/12/largest-known-diamond-in-north-america.html
13:22  * ShowTitle <Title> Largest Known Diamond in North America Found

It just so happens that the URL in the middle line ends **EXACTLY** in the last column of the terminal.  When I click the link to open it, alacritty does not realize that the URL ends there, and wraps around and grabs the timestamp from the next line and appends it to the URL, which then does not resolve because largest-known-diamond-in-north-america.html13:22 does not exist.

I'm not certain how alacritty can best detect this condition.  Clearly I can work around it **in this use case** by making my URL handler strip anything that looks like an Irssi timestamp off the end of the URL, but that's not going to work for everyone.

chrisduerr:
If a URL ends right at the end of the terminal, it would sometimes
incorrectly include the characters from the following line when
launching the URL.

Similar to the semantic search function, the URL parsing iterator will
now stop if it encounters a cell at the end of the line which does not
contain the `WRAPLINE` flag.

This fixes #1906.

### DIFF
--- a/src/term/mod.rs
+++ b/src/term/mod.rs
@@ -85,7 +85,7 @@ impl Search for Term {
         point.line = min(point.line, self.grid.len() - 1);
 
         let mut iter = self.grid.iter_from(point);
-        let last_col = self.grid.num_cols() - Column(1);
+        let last_col = self.grid.num_cols() - 1;
 
         while let Some(cell) = iter.next() {
             if self.semantic_escape_chars.contains(cell.c) {
@@ -103,6 +103,8 @@ impl Search for Term {
     }
 
     fn url_search(&self, mut point: Point<usize>) -> Option<Url> {
+        let last_col = self.grid.num_cols() - 1;
+
         // Switch first line from top to bottom
         point.line = self.grid.num_lines().0 - point.line - 1;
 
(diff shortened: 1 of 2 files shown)


src/term/mod.rs
```diff
@@ -103,26 +103,33 @@ impl Search for Term {
...
+            {
                 break;
             }
         }
-        for cell in iterf {
-            if url_parser.advance_right(cell.c) {
+
+        while let Some(cell) = iterf.next() {
+            if url_parser.advance_right(cell.c)
+                || (iterf.cur.col == last_col && !cell.flags.contains(cell::Flags::WRAPLINE))
```
cole-h: Just a small nit:

Above, you have the `WRAPLINE` check before `advance_left`, and here you have it the other way around. If there's no reason for that I think it would look nicer to keep the same order (so either `WRAPLINE` before `advance_*` in both, or vice versa).

chrisduerr: This is intentional. The short-circuiting here is used to return *before* the url parser is advanced when moving to the left.

PR #2195, opened 2019-03-19
Fix URL highlight crash

### LINKED ISSUE #2194: Crash when resizing with URL highlight
The URL highlight stores the state of the last URL highlight with the position of the URL start position. However when resizing, it's possible that the indices of this point change which will cause a crash if the old positions are not within the grid anymore.

The best approach to fix this seems to be to clear the URL highlight state when the window is resized.

chrisduerr:
The URL highlight stores the state of the last URL highlight with the
position of the URL start position. However when resizing, it's possible
that the indices of this point change which will cause a crash if the
old positions are not within the grid anymore.

This has been resolved by resetting the URL highlight state whenever the
terminal is resized.

This fixes #2194.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -10,7 +10,7 @@ use std::env;
 
 use serde_json as json;
 use parking_lot::MutexGuard;
-use glutin::{self, ModifiersState, Event, ElementState, MouseButton, MouseCursor};
+use glutin::{self, ModifiersState, Event, ElementState, MouseButton};
 use copypasta::{Clipboard, Load, Store, Buffer as ClipboardBuffer};
 use glutin::dpi::PhysicalSize;
 
@@ -222,13 +222,6 @@ pub enum ClickState {
     TripleClick,
 }
 
-/// Temporary save state for restoring mouse cursor and underline after unhovering a URL.
-pub struct UrlHoverSaveState {
-    pub mouse_cursor: MouseCursor,
-    pub underlined: Vec<bool>,
-    pub start: Point<usize>,
-}
-
 /// State of the mouse
 pub struct Mouse {
     pub x: usize,
@@ -245,7 +238,6 @@ pub struct Mouse {
     pub lines_scrolled: f32,
     pub block_url_launcher: bool,
     pub last_button: MouseButton,
-    pub url_hover_save: Option<UrlHoverSaveState>,
 }
 
 impl Default for Mouse {
@@ -265,7 +257,6 @@ impl Default for Mouse {
             lines_scrolled: 0.0,
             block_url_launcher: false,
             last_button: MouseButton::Other(0),
-            url_hover_save: None,
         }
     }
 }
--- a/src/input.rs
+++ b/src/input.rs
@@ -463,14 +463,12 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
             let start = Point::new(line, Column(col));
 
             // Update URLs only on change, so they don't all get marked as underlined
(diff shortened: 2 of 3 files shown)


chrisduerr:
I think there are still some code style improvements which can be made here, but it seems like it's already working as intended.

nixpulvis:
I couldn't reproduce any crash on master, but I could get the underline to be stuck. This bad behavior is indeed fixed by this branch FWIW.

chrisduerr:
@victorhsieh I've just added a new commit, is this behavior changed now?

Commit message for explanation:
```
The original PR incorrectly required the shift modifier to be required
when the user was in the alternate screen buffer. However the correct
behavior is to require it when the mouse mode is enabled.

This has been resolved and URLs are now highlighted in the alt screen
even if no shift modifier is pressed.
```

victorhsieh:
> Are you certain you're not just moving the mouse while clicking?
> 
> I can't reproduce this issue, for me it works as intended right now.

Not sure what's wrong.  Setting the modifier to "Control" then click works for me, but not Alt ...  So it's probably not related to your fix.

chrisduerr:
@victorhsieh Could you go into a little more detail of what's actually wrong with steps to reproduce?

victorhsieh:
Sure.  Here is my configuration.
 - Linux
 - Sync to master, with the 3 patches here applied.
 - My alacritty.xml is same to the example but with minor size adjustments.
 - Tried both with tmux (TERM=screen-256color) and without (TERM=xterm-256color)

When mouse.url.modifiers is None/Shift/Control, clicking on the underlined url does open.  But strangely Alt doesn't.

chrisduerr:
@victorhsieh So if you clone this branch, run `cargo run -- --config-file /dev/null` and then execute `echo -e "\e[?1049h"` and after that print `https://example.com`, the URL is underlined but when clicking on it, it doesn't launch?

If this is the case (please try these exact steps to make sure reproduction is simple), is it maybe launched in a browser instance that is on a different workspace/monitor?

When running `alacritty -vv`, does it print any debug output when clicking on the URL?

victorhsieh:
With other mods, I see  "Launched xdg-open with args", but not with Alt.

chrisduerr:
I can not reproduce. Please make sure you don't have alt setup for a different keybinding or it's blocked by your system.

chrisduerr:
@victorhsieh Would you be willing to add some debug output to check where this issue might be coming from?

The following diff should give some insights:
```diff
diff --git a/src/input.rs b/src/input.rs
index afcf862..04fc7a6 100644
--- a/src/input.rs
+++ b/src/input.rs
@@ -435,6 +435,8 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
 
     /// Underline URLs and change the mouse cursor when URL hover state changes.
     fn update_url_highlight(&mut self, point: Point, modifiers: ModifiersState) {
+        println!("RECEIVED MODS: {:?}", modifiers);
+        println!("REQUIRED MODS: {:?}", self.mouse_config.url.modifiers);
         let mouse_mode =
             TermMode::MOUSE_MOTION | TermMode::MOUSE_DRAG | TermMode::MOUSE_REPORT_CLICK;
```

It should be possible to just apply that to the latest commit in this PR. If you're willing to help and have any questions, please let me know.

chrisduerr:
I've fixed one remaining crash and everything should be good to go now.

If the alt-click issue is still reproducible, I recommend opening a separate issue for it, though I still cannot reproduce it.

PR #2201, opened 2019-03-21
Maximize window after mapping

### LINKED ISSUE #1902: start_maximized doesn't work on KDE
ThinkPad x1 carbon 5gen(1920×1080)
Latest KDE neon

X11

Width not be maximized.

CrackedP0t:
Because Alacritty's window starts out unmapped with `with_visibility(false)`, adding `with_maximized()` makes Winit try to maximize the window before it's mapped. Compliant window managers ignore the request, so when `show()` is called, it's not maximized. This pull request explicitly maximizes the window after it's mapped, and fixes #1902.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixes increase/decrease font-size keybindings on international keyboards
 - On Wayland, the `--title` flag will set the Window title now
 - Parsing issues with URLs starting in the first or ending in the last column
+- Fix not starting maximized on X11
 
 ## Version 0.2.9
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixes increase/decrease font-size keybindings on international keyboards
 - On Wayland, the `--title` flag will set the Window title now
 - Parsing issues with URLs starting in the first or ending in the last column
-- Fix not starting maximized on X11
+- Fix `start_maximized` option on X11
 
 ## Version 0.2.9
 
--- a/src/window.rs
+++ b/src/window.rs
@@ -138,6 +138,7 @@ impl Window {
         let window = create_gl_window(window_builder.clone(), &event_loop, false)
             .or_else(|_| create_gl_window(window_builder, &event_loop, true))?;
         window.show();
+        window.set_maximized(window_config.start_maximized());
 
         // Text cursor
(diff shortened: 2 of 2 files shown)


CrackedP0t:
Alright, now on all platforms it calls `with_maximized()` on the builder, so it will start maximized on all platforms except X11, and it's possible that some WMs will let it start maximized before mapping. Then, on X11 only, it will maximize it after `show()`.

src/window.rs
```diff
@@ -139,6 +140,11 @@ impl Window {
             .or_else(|_| create_gl_window(window_builder, &event_loop, true))?;
         window.show();
 
+        // Maximize window after mapping in X11
+        if cfg!(not(any(target_os = "macos", windows))) && event_loop.is_x11() {
```
chrisduerr: We should use `#[cfg(...)]` here instead of `cfg!`. It's a little uglier since it requires an extra scope, but that makes sure it's only checked at compile-time.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixes increase/decrease font-size keybindings on international keyboards
 - On Wayland, the `--title` flag will set the Window title now
 - Parsing issues with URLs starting in the first or ending in the last column
+- Fix not starting maximized on X11
```
chrisduerr: I think we can improve this a little bit. Right now people might think this means Alacritty will always start maximized, which is not the case.

So we should say that this fixed not starting maximized on X11 when the config setting is enabled.

src/window.rs
CrackedP0t: Oh, OK, thanks - yeah, on MacOS and Windows it was failing because of that.

CHANGELOG.md
CrackedP0t: Alright, changed that

src/window.rs
```diff
@@ -141,8 +141,11 @@ impl Window {
         window.show();
 
         // Maximize window after mapping in X11
-        if cfg!(not(any(target_os = "macos", windows))) && event_loop.is_x11() {
-            window.set_maximized(window_config.start_maximized());
+        #[cfg(not(any(target_os = "macos", windows)))]
+        {
+            if event_loop.is_x11() {
+                window.set_maximized(window_config.start_maximized());
```
chrisduerr: Another issue with doing it like this is that we could be potentially unsetting the WMs maximized state if a user sets the window to maximized through the WM instead of Alacritty.

So when using `set_maximized` after start, we should only do so when maximized is `true`, we shouldn't unmaximize the window when it is `false`.

So `if event_loop.is_x11() && window_config.start_maximized() { window.set_maximized(true); }` is probably what we want.

PR #2207, opened 2019-03-23
Clipboard rework

chrisduerr:
This switches our own `copypasta` crate with the more standardized
`clipboard` library, which allows us to get rid of the `xclip`
dependency on X11.

Additionally, this lays the foundation for native Wayland clipboard
support once the clipboard crate is updated (or a fork is created).

Fixes https://github.com/jwilm/alacritty/issues/5.


Cargo.toml
```diff
@@ -99,7 +98,9 @@ maintainer = "Joe Wilm <joe@jwilm.com>"
 license-file = ["LICENSE-APACHE", "3"]
 extended-description = """\
 Alacritty is the fastest terminal emulator in existence. Using the GPU for \
-rendering enables optimizations that simply aren't possible without it.  """
+rendering enables optimizations that simply aren't possible in other emulators. \
+Alacritty currently supports FreeBSD, Linux, macOS, and OpenBSD. Windows \
+support is planned before the 1.0 release.  """
```
atouchet: This should be changed back.

src/clipboard.rs
```diff
@@ -0,0 +1,52 @@
...
+use clipboard::{ClipboardProvider, ClipboardContext};
+
+#[derive(Debug)]
+pub enum Clipboard {
+    Primary,
+    Secondary,
+}
+
+impl Clipboard {
+    pub fn store(&self, text: impl Into<String>) {
```
nixpulvis: I think I'd prefer to return a `Result` here and log warning messages and such where this is used.

chrisduerr: We never need to handle the error case but only want to log it. This effectively just reduces code bloat by having the same code at all callsites.

Since this is for internal use only, not a public API, I think this is the better approach here.

nixpulvis: Makes sense then

c-edw:
>Additionally, this lays the foundation for native Wayland clipboard
support once the clipboard crate is updated (or a fork is created).

Is this still relevant? I'm testing this branch on Wayland and everything seems to be working normally.

chrisduerr:
This does not change anything about Wayland clipboard handling yet. Which means there is no native support.

gustafla:
Seems to work on sway on Gentoo but also experienced terminals freezing (might be unrelated).

Flakebi:
This works better than no clipboard support but it sometimes freezes alacritty. I was able to get reproducible freezes when marking some text and using middle-click to paste it into *the same window*.

I know that in x11, the problem in this case is that we wait for the paste data but the same process has to deliver them from the same thread. This leads to the deadlock.
My guess is that this is the same for wayland.

chrisduerr:
I've created an issue upstream: https://github.com/Smithay/smithay-clipboard/issues/3

trimental:
smithay-clipboard 0.1.1 has been released and should fix a few problems. Still no Weston support but hopefully someone can test that KDE and Gnome work as well as Sway does. Weston is a bit of a problem child though I think.

chrisduerr:
I'd personally be fine with not supporting Weston as long as KDE, Gnome and Sway work. I've been able to confirm myself that Sway does, however I can't easily test the rest of them.

So if anyone is able to test, I'd much appreciate it. I've bumped this PR to include the latest smithay-clipboard version, so it should be possible to test even with the existing merge conflicts.

gustafla:
As a sway user, at least with a few minutes of testing I encountered no freezes of panics and copying and pasting stuff around with firefox and wl-clipboard worked flawlessly

dvergeylen:
As a Gnome user, I encountered no freezes of panics from copying and pasting stuff via <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>v</kbd> or middle mouse button. :+1: 

I had to install `libxcb-shape0-dev libxcb-xfixes0-dev` as dependencies to make it compile (Debian 10).

```bash
sudo apt install libxcb-shape0-dev libxcb-xfixes0-dev
```

trimental:
I've tested this PR with smithay-clipboard v0.2 on KDE (Wayland), Gnome, Weston and Sway and all platforms seem to work well without freezing or panics and multiple seats also now work properly :+1:

t184256:
I used 1cae9f3 on sway for a couple of days and it worked fine.

rumpelsepp:
Firefox on wayland seems to use `\r\n` line endings for the wl clipboard. Pasting into the terminal often leads me to annoying surprises... vte based terminal seems to convert the line endings on paste. What do you think about converting line endings on paste to the host platform?

chrisduerr:
I think that Firefox should fix that. It seems extremely odd to me that we would need to handle any kind of workaround for that. If `\r\n` is pasted to the clipboard, the clipboard should **never** alter the clipboard because it thinks that's incorrect. It's a tool for storing data not editing it.

Now if it's our fault that the line endings are incorrect, that would be a different story. But if an application pastes incorrect line endings for an application into the clipboard, that needs to be fixed in the application.

I'd assume that clipboards in Firefox are done through GTK though, so it surprises me that it would have the incorrect line endings.

rumpelsepp:
> I'd assume that clipboards in Firefox are done through GTK though, so it surprises me that it would have the incorrect line endings.

I am struggling with this for quite some time now on `sway` and it seems that indeed firefox does something wrong:

Copy the alacritty config file:

```
$ wl-paste | hexdump -C | grep 0a | tail
00005990  20 20 7d 0d 0a 20 20 2d  20 7b 20 6b 65 79 3a 20  |  }..  - { key: |
000059e0  20 20 7d 0d 0a 20 20 2d  20 7b 20 6b 65 79 3a 20  |  }..  - { key: |
00005a30  20 20 7d 0d 0a 20 20 2d  20 7b 20 6b 65 79 3a 20  |  }..  - { key: |
00005a80  20 20 7d 0d 0a 20 20 2d  20 7b 20 6b 65 79 3a 20  |  }..  - { key: |
00005ad0  20 20 7d 0d 0a 20 20 2d  20 7b 20 6b 65 79 3a 20  |  }..  - { key: |
00005b20  20 20 7d 0d 0a 20 20 2d  20 7b 20 6b 65 79 3a 20  |  }..  - { key: |
00005b70  20 20 7d 0d 0a 20 20 2d  20 7b 20 6b 65 79 3a 20  |  }..  - { key: |
00005bc0  20 20 7d 0d 0a 20 20 2d  20 7b 20 6b 65 79 3a 20  |  }..  - { key: |
00005c10  20 20 7d 0d 0a 20 20 2d  20 7b 20 6b 65 79 3a 20  |  }..  - { key: |
00005c60  20 20 7d 0d 0a 0a
```
vs

```
$ curl -Ls https://raw.githubusercontent.com/jwilm/alacritty/master/alacritty.yml | hexdump -C | grep 0a | tail
00005730  20 20 20 20 20 20 20 20  20 20 20 20 20 7d 0a 20  |             }. |
00005780  20 20 20 20 20 20 20 20  20 20 20 20 7d 0a 20 20  |            }.  |
000057d0  20 20 20 20 20 20 20 20  20 20 20 7d 0a 20 20 2d  |           }.  -|
00005820  20 20 20 20 20 20 20 20  20 20 7d 0a 20 20 2d 20  |          }.  - |
00005870  20 20 20 20 20 20 20 20  20 7d 0a 20 20 2d 20 7b  |         }.  - {|
000058c0  20 20 20 20 20 20 20 20  7d 0a 20 20 2d 20 7b 20  |        }.  - { |
00005910  20 20 20 20 20 20 20 7d  0a 20 20 2d 20 7b 20 6b  |       }.  - { k|
00005960  20 20 20 20 20 20 7d 0a  20 20 2d 20 7b 20 6b 65  |      }.  - { ke|
000059b0  20 20 20 20 20 7d 0a 20  20 2d 20 7b 20 6b 65 79  |     }.  - { key|
00005a00  20 20 20 20 7d 0a                                 |    }.|
```

I only notice this behaviour on non vte terminals, since the vte library seems to clean the `0x0d` bytes up.

chrisduerr:
Yeah I would strongly recommend complaining on the mozilla bugtracker, that is the right place for this. Maybe there's already an existing issue too.

VTE might be stripping them either way, I'm not sure if this was done explicitly because of Firefox, but I think I'd rather not if somehow possible.

rumpelsepp:
yeah. sorry for being somewhat offtopic.

doronbehar:
Awesome :beers: Can't wait to get xclip out of my system.

7415963987456321:
This seems to work fine on wayland(sway) for me, except I've seem to have run into a bug and a performance issue. 

The bug is that if a large amount of text is copied from one terminal to another the terminal that is being pasted into "freezes", i.e it still takes input fine, I can still continue pasting or write into it but I cannot escape the text with CTRL+C or clear the terminal by pressing enter. 

As for the performance, it is quite slow, both when pasting directly into the terminal or into vim. And in the case of directly pasting into the terminal it just gets slower for each successive paste.

Both of these accounts were tested with about a pages worth of dmesg output.

Issue #1673, opened 2018-10-17
[Windows] Text a bit blurry


Geobert:
Just tested the Windows' build and I see great potential!

However, I found the text is a bit blurry and I didn't find the Windows' way to modify the aliasing setting (no fontconfig on Windows).

zacps:
Rusttype (the font backed on Windows) doesn't currently support hinting. 

Given plans to switch to font kit I'll probably not make any changes until that lands.

PR #2224, opened 2019-03-28
Add DirectWrite font rasterizer

benpye:
This adds a DirectWrite font rasterizer for Windows. This enables subpixel rendering and hinting on Windows, resolving #1673 . I have stuck this behind a feature flag, directwrite, for now, I'm not sure what would be desirable going forward. I was not confident enough to outright replace rusttype.

I have tested this at 100% DPI, 150% DPI and 200% DPI and it appears to behave correctly. There are no test regressions though I haven't checked to see if there are any tests that would catch a renderer bug! I have run rustfmt with the rustfmt config in #2212 on the presumption that PR will be merged in the future, I wanted to minimise any future required changes.

I did try to run vtebench but it seems that it doesn't work on Windows, is there something I'm missing? For example, am I meant to run vtebench in WSL to check Windows performance or is Windows perf something that hasn't been looked at yet?

Whilst not a regression when compared to the rusttype, this rasterizer currently does not attempt to fallback if the font isn't found. The DirectWrite APIs for fallbacks appear pretty complex (https://docs.microsoft.com/en-gb/windows/desktop/api/dwrite_2/nf-dwrite_2-idwritefontfallback-mapcharacters) and are not exposed in dwrote. Perhaps it would be sufficient to hardcode some fallback familes as Windows ships with a well known set of fonts, it seems to me that Consolas should be a safe fallback but perhaps we should look for Lucida Console too.

Apologies if I've made any silly mistakes, this is the most Rust I've ever written, I think I like it though!

Fixes https://github.com/jwilm/alacritty/issues/2316.


chrisduerr:
I haven't been able to test this at all and there are some style issues which I will comment on later, but in general I really like this change.

On Linux and macOS we're already using the native font toolkits and I think this should definitely be the default on Windows too. Unless there's any feature we're losing because of this (which I'm not aware of at this moment) I'm very in favor of completely replacing rusttype with this.

> Perhaps it would be sufficient to hardcode some fallback familes as Windows ships with a well known set of fonts, it seems to me that Consolas should be a safe fallback but perhaps we should look for Lucida Console too.

As far as I'm aware (as someone not using Windows), Consolas is the default monospace font on Windows, so using that as a fallback if the font can't be found at all makes sense to me.

font/src/directwrite/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+            / f32::from(vmetrics.designUnitsPerEm);
+
+        let underline_position = f32::from(vmetrics.underlinePosition) * scale;
+        let underline_thickness = f32::from(vmetrics.underlineThickness) * scale;
+
+        let strikeout_position = f32::from(vmetrics.strikethroughPosition) * scale;
+        let strikeout_thickness = f32::from(vmetrics.strikethroughThickness) * scale;
+
+        let line_height = f64::from(
+            i32::from(vmetrics.ascent) + i32::from(vmetrics.descent) + i32::from(vmetrics.lineGap),
```
chrisduerr: There are a lot of conversions here, could we instead use one of the types of the metrics to convert to so we do not need to convert everything?

font/src/directwrite/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//! Rasterization powered by DirectWrite
+extern crate dwrote;
+use self::dwrote::{
+    FontCollection, FontStretch, FontStyle, FontWeight, GlyphOffset, GlyphRunAnalysis,
+};
```
chrisduerr: Using `extern crate` isn't required (and discouraged) in Rust 2018, so it should be removed.

Additionally, the `self::` should not be necessary here.

font/src/directwrite/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+        let system_fc = FontCollection::system();
+
+        let weight = match desc.style {
+            Style::Description { weight, .. } => match weight {
+                Weight::Normal => FontWeight::Regular,
+                Weight::Bold => FontWeight::Bold,
+            },
+
+            _ => FontWeight::Regular,
+        };
```
chrisduerr: Rust's pattern matching actually has quite a neat trick to simplify this even more, I think this should work:

```rust
let style = if let Style::Description { weight: Weight::Bold, .. } = desc.style {
    FontWeight::Bold
} else {
    FontWeight::Regular
};
```

font/src/directwrite/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+
+        let style = match desc.style {
+            Style::Description { slant, .. } => match slant {
+                Slant::Normal => FontStyle::Normal,
+                Slant::Oblique => FontStyle::Oblique,
+                Slant::Italic => FontStyle::Italic,
+            },
+
```
chrisduerr: This empty line should be removed, but I think the pattern matching trick wouldn't fit in here really well so the match can stay.

chrisduerr:
I'd be interested in what @zacps or maybe @davidhewitt think about making this the default on Windows and completely deleting rusttype.

While in the future we might want to use fontkit, that will likely still take a while. I'd be in favor of removing rusttype, since it has troubles with some fonts and doesn't support some font features. It's also the only non-native fontstack Alacritty is using.

Though I'm not completely aware of what advantages rusttype brings us (or directwrite's disadvantages) except for rusttype being a pure Rust solution (which definitely isn't enough of a reason I think).

font/src/directwrite/mod.rs
benpye: The font crate isn't set to use Rust 2018 yet - do we want to change that as part of this PR or should it be split out?

benpye:
I've resolved the style comments but I'm not sure about removing the `extern crate dwrote`.

benpye:
I think that last commit should address #2234 . I have tested their example and the following config works:

```
normal:
  family: Roboto Mono for Powerline
  style: Medium
```

davidhewitt:
> I'd be interested in what @zacps or maybe @davidhewitt think about making this the default on Windows and completely deleting rusttype.

I'm all for simplification, and using the native font kit seems like a good idea. I just built and ran this PR successfully, seems to be working fine to me.

I'm not a font expert by any means, but I had a couple of thoughts / worries:

- I know DirectWrite can be hardware-accelerated, but I think it has to be used in tandem with Direct2D to achieve that. So I guess we're using CPU rendering here. Not sure if rusttype is using the GPU?
- At least for me, it looks like text anti-aliasing is not working with DirectWrite. Is this something we can enable? (The glyphs rendered by rusttype just appear slightly smoother. It's kinda hard to tell with modern screens with high PPI anyway!)

font/src/directwrite/mod.rs
chrisduerr: I wouldn't make any major changes to the font crate until we've made up our mind about fontkit, which I'd expect to happen after the 0.3 release.

chrisduerr:
> I think that last commit should address #2234

That font you've used there is still shorter than 31 character though. So you'd need to test something longer.

> So I guess we're using CPU rendering here. Not sure if rusttype is using the GPU?

I believe so does rusttype? Though it's possible to cache glyphs on the GPU with rusttype. A performance benchmark would certainly make sense here though. @benpye you said that vtebench doesn't seem to work, what exactly was your issue here? You might have to specify `--term xterm`.

> At least for me, it looks like text anti-aliasing is not working with DirectWrite. Is this something we can enable?

That's definitely worth investigating. I'd imagine by default it should behave the same way as the native font system though? And I feel like that should enable it?

benpye:
> > I think that last commit should address #2234
> 
> That font you've used there is still shorter than 31 character though. So you'd need to test something longer.

That is true - I will try and find a 31 character+ named font though I believe it should work

> > So I guess we're using CPU rendering here. Not sure if rusttype is using the GPU?
> 
> I believe so does rusttype? Though it's possible to cache glyphs on the GPU with rusttype. A performance benchmark would certainly make sense here though. @benpye you said that vtebench doesn't seem to work, what exactly was your issue here? You might have to specify `--term xterm`.

I believe this is CPU rendering, yeah. vtebench didn't work on Windows due to a terminfo issue, but I will try passing in `--term xterm`.
 
> > At least for me, it looks like text anti-aliasing is not working with DirectWrite. Is this something we can enable?
> 
> That's definitely worth investigating. I'd imagine by default it should behave the same way as the native font system though? And I feel like that should enable it?

It should work (and certianly does on my machine). I would check that you have built with the `directwrite` feature - if you have then I am a little surprised. I believe it should be using your system's ClearType settings (hence `DWRITE_RENDERING_MODE_NATURAL`). If you still get no ClearType rendering in Alacritty with this patch set and the same font as cmd.exe then that would definitely indicate an issue.

davidhewitt:
> I believe this is CPU rendering, yeah. vtebench didn't work on Windows due to a terminfo issue, but I will try passing in --term xterm.

If I remember right I had to use Ubuntu (on WSL) to generate the vtebench files, because of that terminfo issue. I've got a couple of busy weeks ahead but if I find a spare moment I can try and figure that out again and upload some benchmarks.

> It should work (and certianly does on my machine). I would check that you have built with the directwrite feature - if you have then I am a little surprised. I believe it should be using your system's ClearType settings (hence DWRITE_RENDERING_MODE_NATURAL). If you still get no ClearType rendering in Alacritty with this patch set and the same font as cmd.exe then that would definitely indicate an issue.

I was definitely using the `directwrite` feature - if I can find a spare moment I'll try playing around with the rendering mode and also make some enlarged screenshots so you can see what it looks like on my statement.

chrisduerr:
> If I remember right I had to use Ubuntu (on WSL) to generate the vtebench files, because of that terminfo issue. I've got a couple of busy weeks ahead but if I find a spare moment I can try and figure that out again and upload some benchmarks.

That's definitely possible. I'd share the benchmark files I usually use but unfortunately these are huge usually. But it does kinda depend on the Unix terminfo structure so that might cause issues on Windows.

davidhewitt:
I had a look at benchmarks today. I couldn't get any observable difference between the two branches.

Also, I managed to work out what's going on with the anti-aliasing: rusttype is using greyscale anti-aliasing, whereas DirectWrite is using rgb anti-aliasing. (Screenshots below.)

This seems fine to me! So I think we're good to replace rusttype with DirectWrite entirely, if that's what you guys want to do! 😄 

rusttype:
rusttype

DirectWrite:
directwrite

font/src/directwrite/mod.rs
benpye: I think this should now be fixed - let me know if this isn't what you meant.

font/src/directwrite/mod.rs
benpye: Okay, I'll leave it at 2015 then meaning we still need this.

font/src/directwrite/mod.rs
benpye: Sorted!

zacps:
Moving to DirectWrite makes sense and I hoped to do that via font-kit However, I've not been able to work on it for a while so I'm in favour of merging this now (this is not a review though).

font/src/directwrite/mod.rs
chrisduerr: That was actually not exactly what I was proposing.

Ascent and descent are both u16 while the line gap is i16. So to reduce conversions as much as possible, it should be possible to add all of them together by just converting the ascent and descent instead of having to convert everything.

benpye:
Shall I go ahead and update this PR to replace rusttype with directwrite at this point now that 0.3.0 is released?

chrisduerr:
@benpye Please do. I think that would be the right path forward.

musm:
I was just going to open an issue on how bad the fonts look on windows because they're way too blurry. 👍  on this change.

bmisiak:
@benpye, font fallback appears to be properly exposed in Servo's `dwrote`. It is significantly newer, see the changes:
https://github.com/vvuk/dwrote-rs/compare/master...servo:master#diff-5f00dd21f727b764279d6a48d3616

font/Cargo.toml
```diff
@@ -24,3 +27,4 @@ core-foundation-sys = "0.6"
 [target.'cfg(windows)'.dependencies]
 font-loader = "0.6.0"
 rusttype = "0.4.1"
+dwrote = { version = "0.8.0", optional = true }
```
bmisiak:
```suggestion
dwrote = { git = "https://github.com/servo/dwrote-rs.git", optional = true }
```
Two weeks ago, the `dwrote` fork for servo added support for font fallback.

chrisduerr: Please note that the `servo-dwrote` fork would probably be a better idea than a git dependencies.

bmisiak: @chrisduerr Unfortunately, the `servo-dwrote` crate, while it links to the right github repository, was last updated 2 years ago, so it's wayyy behind.

chrisduerr: Then the way to go would probably be to ask them to push a new version.

benpye:
@chrisduerr Sorry for this slow progress on this - I'll find time either today or tomorrow to update this PR.

Do you have a preferred path for the font fallback? I haven't looked at what it would take to use the interface exposed by the newer dwrote crate.

chrisduerr:
If servo's dwrote crate exposes font fallback, I don't see why we wouldn't tap into that. That seems like the logical path forwards.

font/Cargo.toml
atouchet: `servo-dwrote` was a temporary fork that is no longer used. `dwrote` is the up to date crate.

atouchet:
dwrote 0.9 with font fallback is out now.

chrisduerr:
@benpye The primary thing left for this seems to be the removal of rusttype and replacing it with the new direct write API. I think merging this early in the 0.3.2 cycle would be a good idea so it can be tested thoroughly.

Depending on your availability to work on this, I'd propose removing the font fallback and moving that to a follow-up PR to prevent this from going stale. Let me know if you think that would be a better approach so we con reduce the amount of rebase work required.

benpye:
@chrisduerr That sounds good to me. I've just pushed changes to resolve some code conflicts and replace rusttype with directwrite. I haven't squashed yet, though I think GitHub lets you do that when you merge now?

Let me know if there's anything remaining!

chrisduerr:
> I haven't squashed yet, though I think GitHub lets you do that when you merge now?

This has worked in github for years, yes. So I'd really appreciate not squashing commits since that makes reviews very difficult.

PR #2250, opened 2019-04-03
Add notes about icon installation on Linux

kchibisov:
After completing all the steps presented in INSTALL.md, Alacritty will correctly report its icon (e.g. screenshot from rofi).
out


kchibisov:
Yes, adding information about modified desktop file in changelog is a good idea.
Maybe we should also rename ```alacritty-term.svg```  to ```Alacritty.svg```? This should help avoid some mistakes when doing packaging or installation. (I don't know how to properly rename ```alacritty.svg``` in this case though).

chrisduerr:
I think if a maintainer would want to pick one of the different icons, that's perfectly fine too. The default is documented in our installation guide and that should get people going.

Though I could understand people thinking that `alacritty.svg` would be the 'default' icon so I'd be open to change that one to `alacritty-simple.svg` or something like that so no judgement is made in naming which one should be the default.

kchibisov:
Sorry for asking here, shouldn’t we change desktop entries in snapcraft.yml, since desktop file was moved from ./ to extra/linux/ ?

chrisduerr:
That's a good point, I've fixed it in https://github.com/jwilm/alacritty/pull/2255.

PR #2251, opened 2019-04-03
Bump dependencies

chrisduerr:


atouchet:
I don't know how much work they'd take to update but font-loader and rusttype in the font crate could use updates as well.

chrisduerr:
Thanks for the input, looks like deps.rs doesn't actually show Windows dependencies.

I've bumped all those too, hopefully that should be good to go now.

PR #2252, opened 2019-04-03
Upgrade to winapi 0.3.7

### LINKED ISSUE #1920: Use HPCON definition from winapi
In `src/tty/conpty.rs`, near the top we have the definition `pub type HPCON = *mut c_void;`

Once https://github.com/retep998/winapi-rs/pull/699 is released then we can update to a newer winapi and `use winapi::um::wincontypes::HPCON;`

(I added this definition there so that we could merge the support for `ConPty` apis without having to depend on a patched `winapi`.)

davidhewitt:
Bump to winapi 0.3.7.

Closes #1920

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,8 +62,8 @@ x11-dl = "2"
 winpty = { path = "./winpty" }
 mio-named-pipes = "0.1"
 miow = "0.3"
-winapi = { version = "0.3.5", features = ["impl-default", "winuser", "synchapi", "roerrorapi", "winerror", "wincon"]}
 dunce = "1.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -62,8 +62,8 @@ x11-dl = "2"
 winpty = { path = "./winpty" }
 mio-named-pipes = "0.1"
 miow = "0.3"
-winapi = { version = "0.3.5", features = ["impl-default", "winuser", "synchapi", "roerrorapi", "winerror", "wincon"]}
 dunce = "1.0"
+winapi = { version = "0.3.7", features = ["impl-default", "winuser", "synchapi", "roerrorapi", "winerror", "wincon"]}
```
atouchet: Do you need wincontypes here?

davidhewitt: It's not needed as it's a sub-feature of the wincon feature. But I think we might aswell be explicit and add it.

PR #2255, opened 2019-04-04
Fix snap desktop

chrisduerr:


kchibisov:
Probably [this one](https://github.com/jwilm/alacritty/blob/91f889849e13e720f3803a7b0405ec4a9cf6005b/extra/linux/snap/snapcraft.yaml#L32) too, if I understand [reference](https://docs.snapcraft.io/snapcraft-yaml-reference/4276) correctly.

chrisduerr:
@kchibisov The reference says that this path is relative to the prime directory. Which is the directory of the snap after it's packaged. The instance changed in this PR copies the desktop file to the root of the prime directory so it should be fine.

PR #2261, opened 2019-04-06
Improve errors for missing shell file on Windows 

### LINKED ISSUE #2016: screen is just black
Windows 10 home edition, Alacritty-v0.2.5.exe.  I've also tried with 0.2.4 and get the same results.

I suppose this is something environmental - I mean you'd have noticed if this thing just didn't work on Windows! - but I've no idea what.

When I run Alacritty, I just get a black screen, and a cursor.  Typing gives the impression that I might be getting black text on a black background, the cursor looks as though it's moving the right amount.

chrisduerr:
Currently with WinPTY there is no error at all if the shell does not
exist and on ConPTY the error is cryptic 'assertion failed: success'.

This changes the ConPTY assertion to a human-readable error which
clearly states that the file could not be found.

Currently the WinPTY backend has not been fixed yet.

This fixes #2016.

### DIFF
--- a/src/tty/windows/conpty.rs
+++ b/src/tty/windows/conpty.rs
@@ -15,11 +15,11 @@
 use super::{Pty, HANDLE};
 
 use std::i16;
+use std::io::Error;
 use std::mem;
 use std::os::windows::io::IntoRawHandle;
 use std::ptr;
 use std::sync::Arc;
-use std::io::Error;
 
 use dunce::canonicalize;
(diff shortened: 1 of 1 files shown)


chrisduerr:
To get these improvements out to the ConPTY users a bit faster, I think the better approach here is to merge this change and open a tracking issue for the WinPTY side.

I've created an issue for the WinPTY here: https://github.com/jwilm/alacritty/issues/2344

PR #2262, opened 2019-04-06
Consistent treatment of numeric types

dimbleby:
Don't mix floats and integers.

As mentioned in https://github.com/jwilm/alacritty/issues/2016#issuecomment-480379478.

### DIFF
--- a/res/text.v.glsl
+++ b/res/text.v.glsl
@@ -45,8 +45,8 @@ void main()
 
     // Compute vertex corner position
     vec2 position;
-    position.x = (gl_VertexID == 0 || gl_VertexID == 1) ? 1. : 0;
-    position.y = (gl_VertexID == 0 || gl_VertexID == 3) ? 0. : 1;
+    position.x = (gl_VertexID == 0 || gl_VertexID == 1) ? 1. : 0.;
+    position.y = (gl_VertexID == 0 || gl_VertexID == 3) ? 0. : 1.;
 
     // Position of cell from top-left
     vec2 cellPosition = cellDim * gridCoords;


chrisduerr:
Does this actually resolve the issue for you?

dimbleby:
I've no idea - setting up a build environment looks like being a fair amount of work so, I'm slightly embarrassed to admit, I've submitted this without even compiling it.  I mean it can't be wrong, right...?

chrisduerr:
> I've no idea - setting up a build environment looks like being a fair amount of work so, I'm slightly embarrassed to admit, I've submitted this without even compiling it. I mean it can't be wrong, right...?

Without knowing if this PR solves any actual problems, it unfortunately becomes completely useless.

If this indicates that the user's GLSL support might have some incompatibilities, this might actually be useful for debugging.

Don't get me wrong, I completely agree that fundamentally these should both be floats as this PR proposes. But the goal of code changes should always be to fix a problem and unless someone can verify that this is actually the case there's little point in putting time, effort and work to fix problems that do not exist.

dimbleby:
Well, take it or leave it...

As things stand I definitely can't use Alacritty; and there looks to be a high chance that this change will at least get me past the problem that currently blocks that.  I don't see the downside in taking this, myself - but hey, it's not my project.

If you're able to compile an executable that includes this change, I'd be willing to try it out and report back.

chrisduerr:
[alacritty.zip](https://github.com/jwilm/alacritty/files/3050892/alacritty.zip)

dimbleby:
Well: good news and bad news - much as expected, probably.  Alacritty does now start up, but we're back as we were when I opened #2016: the screen is black. 

I notice that if I select the bit of blackness where there ought to be text I do get a fairly clear hint that there is stuff there, it's just not being rendered as expected.  It looks like this: <https://imgur.com/a/7oEwdmv>.

So: this fix is working as expected - but no progress on the original problem.

PR #2278, opened 2019-04-09
Set maximum Rust version to 1.31.0

### LINKED ISSUE #2277: Maximum Rust Version
Especially for distros that usually have an outdated version of stable Rust on their system, it's almost impossible to package Alacritty since it always targets the latest Rust master. This also causes issues for users trying to build it themselves on systems like Debian.

As a solution Alacritty should follow the path that many other applications have gone already and limit itself to a specific Rust version and enforce this in the CI process.

chrisduerr:
By setting the minimum Rust version and enforcing it with CI, Alacritty
should hopefully make it possible for maintainers to package the
application even on distributions which are not rolling release.

The 1.31.0 target has been chosen here because it's the first version of
the Rust 2018 release. Bumping this version in the future should be
considered to be a breaking change and should only be done with caution
and in consideration of the supported Rust versions of the major
distributions available.

This fixes #2277.


mmstick:
You could add a `rust-toolchain` file with `1.31.0` in it, to allow rustup to default to this version, and to make the minimum version more visible.

chrisduerr:
After some consideration I've decided *not* to add a `rust-toolchain` file because it makes working with Rust a bit more difficult since it's enforced too strictly imo (I might want to run nightly rustfmt and clippy, even though my code builds on 1.31.0).

However I did add it to the CONTRIBUTING.md file that version 1.31.0 is the minimum supported versions and all patches need to work with it, so hopefully that will help with awareness a bit.

mmstick:
May want to be wary that sometimes `cargo +nightly fmt` will format code in a backwards-incompatible way.

chrisduerr:
The CI process should make sure that's not the case. However the nightly formatter is far more powerfull than the stable one which is (imo) missing a lot of features that are necessary for a sane formatting standard.

Thanks for mentioning that though.

mmstick:
With this merged, I'll create debian packages for Pop!_OS later today.

PR #2293, opened 2019-04-10
Fix macos subprocess execution

### LINKED ISSUE #2259: Keybinding open -n -a /Application/Alacritty.app not working
We've had a hack for a while on macOS to open multiple windows as a workaround to #607. I dusted of my MBP this morning and found that it's not working. Running the command from a shell in Alacritty works just fine, but the keybinding version causes Alacritty to exit immediately. Seems that this may be a problem with how we create child processes.

If this is a dupe, I blame GitHub search

chrisduerr:
This fixes the execution of subprocesses on macOS which could sometimes
prevent actions like `SpawnNewProcess` or custom commands from launching
their processes correctly.

This fixes #2259.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -93,10 +93,16 @@ where
         .stdout(Stdio::null())
         .stderr(Stdio::null())
         .before_exec(|| unsafe {
-            if ::libc::fork() != 0 {
-                ::libc::_exit(0);
+            match ::libc::fork() {
+                -1 => return Err(io::Error::last_os_error()),
+                0 => (),
+                _ => ::libc::_exit(0),
             }
-            ::libc::setsid();
+
+            if ::libc::setsid() == -1 {
+                return Err(io::Error::last_os_error());
+            }
+
             Ok(())
         })
         .spawn()?
--- a/src/util.rs
+++ b/src/util.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use std::ffi::OsStr;
-use std::process::Command;
+use std::process::{Command, Stdio};
 use std::{cmp, io};
 
 #[cfg(not(windows))]
@@ -21,8 +21,6 @@ use std::os::unix::process::CommandExt;
 
 #[cfg(windows)]
 use std::os::windows::process::CommandExt;
-#[cfg(windows)]
-use std::process::Stdio;
 #[cfg(windows)]
 use winapi::um::winbase::{CREATE_NEW_PROCESS_GROUP, CREATE_NO_WINDOW};
 
@@ -91,9 +89,14 @@ where
 {
     Command::new(program)
         .args(args)
+        .stdin(Stdio::null())
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
         .before_exec(|| unsafe {
-            #[allow(deprecated)]
-            libc::daemon(1, 0);
+            if ::libc::fork() != 0 {
+                ::libc::_exit(0);
+            }
+            ::libc::setsid();
             Ok(())
         })
         .spawn()?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Reset scrolling region when the RIS escape sequence is received
+- Subprocess spawning on macos
 
 ## Version 0.3.0
 


src/util.rs
```diff
@@ -91,9 +89,14 @@ where
...
         .args(args)
+        .stdin(Stdio::null())
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
         .before_exec(|| unsafe {
-            #[allow(deprecated)]
-            libc::daemon(1, 0);
+            if ::libc::fork() != 0 {
+                ::libc::_exit(0);
+            }
```
dnkl: I can barely read rust and perhaps I'm missing something, but you're not checking that `fork()` succeeds? Meaning, if it fails (granted, very unlikely), you'll terminate the main Alacritty process.

chrisduerr: This executes in the context of the new process, so it should never lead to Alacritty quitting.

As far as I can tell, if `fork()` doesn't succeed it would just prevent the subprocess from spawning.

I'm not sure *exactly* how `std::process::Command` works, but I'm assuming it forks already. So what we do here is just a second fork to get the double-fork child process zombie fix behavior.

If you don't trust me, it should be possible to test this by just putting `::libc::_exit(0);` in the `before_exec` closure. That should just lead to no subprocess spawning, not to Alacritty quitting.

dnkl:
Right, of course it has already forked once. 

Would it make sense to exit with a non-zero value when fork() fails? Or is that ignored by Command().

I'm thinking it would be nice to let the user know it failed...

On April 10, 2019 8:17:46 PM GMT+02:00, Christian Duerr <notifications@github.com> wrote:
>chrisduerr commented on this pull request.
>
>
>
>>          .before_exec(|| unsafe {
>-            #[allow(deprecated)]
>-            libc::daemon(1, 0);
>+            if ::libc::fork() != 0 {
>+                ::libc::_exit(0);
>+            }
>
>This executes in the context of the new process, so it should never
>lead to Alacritty quitting.
>
>As far as I can tell, if `fork()` doesn't succeed it would just prevent
>the subprocess from spawning.
>
>I'm not sure *exactly* how `std::process::Command` works, but I'm
>assuming it forks already. So what we do here is just a second fork to
>get the double-fork child process zombie fix behavior.
>
>If you don't trust me, it should be possible to test this by just
>putting `::libc::_exit(0);` in the `before_exec` closure. That should
>just lead to no subprocess spawning, not to Alacritty quitting.
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/jwilm/alacritty/pull/2293#discussion_r274093881

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

chrisduerr:
Unfortunately I'm not sure if there's any way to talk to the user after the process has been spawned. Since we explicitly redirect all stdin/stdout/stderr to /dev/null.

Exiting with a status of -1 would be possible, but I doubt it would actually help anyone. Since nobody's going to know either way.

dnkl:
It should be possible to propagate that back up to the main process.

Surely `std::process::Command`  is (or can be made to) `wait()`:ing for the (first) child process? If so, the exit code should be available, somehow, allowing you to log, or write to stderr.

dnkl:
Don't know how difficult it is, but seeing that fork() rarely fails, maybe it's not worth it.

What _would_ be nice though, is to detect and report exec() failures (maybe the user misspelled a command etc). But that is a bit difficult even in plain C. Doable, but not straight forward.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

chrisduerr:
I've decided to report the last os error when setsid or fork fails. That should hopefully be good now. Thanks for the idea!

dnkl:
Nice!

Are you sure reporting setsid() works though? I would expect it to be as difficult as reporting exec() failure, since it is executing in the context of the double-forked child.


Unless I'm mistaken, std:: process:: Command will only see the result of it's direct child (i.e the explicit _exit()-calls). What happens in the grand-child shouldn't be visible in the main process.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

chrisduerr:
> Are you sure reporting setsid() works though? I would expect it to be as difficult as reporting exec() failure, since it is executing in the context of the double-forked child.

That is correct and I do not expect it to report back to Alacritty. However it might prevent errors that are caused by just going on after the setsid call fails. This way it should explicitly bail when that is the case.

dnkl:
Ah! All good then! (y)

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #2296, opened 2019-04-10
Add Pop!_OS to the list of supported Linux OSs

mmstick:


chrisduerr:
I'm not super familiar with Ubuntu and PPAs, but since this is a PPA it should be possible to use this on other Ubuntu distros that are *not* Pop!_OS by adding the PPA, right?

I feel like it might make sense to make these more general Ubuntu installation instructions and add a note that on Pop!_OS it is not required to add the PPA.

mmstick:
Adding the Pop!_OS PPA on an Ubuntu system will conflict with Ubuntu's packaging, as we override some of Ubuntu's packages. Unexpected things may happen upon adding it to an Ubuntu system. In the future, we may separate open source application packaging from the Pop!_OS PPA, as we do for our proprietary repository. Currently, the Pop!_OS PPA is a dumping ground for our non-proprietary packaging until we have the infrastructure in place to move things.

chrisduerr:
Thanks for the cooperation on this!

mmstick:
If anyone wants to create a PPA specifically for alacritty, or add it to one of their own, they can build debian packages for alacritty from [our fork](https://github.com/pop-os/alacritty), which just adds the `debian` directory to the repository. You can run `dch` to modify the changelog, `dpkg-buildpackage -S -d` to create a source tarball, and then `dput ppa:username/repo ../path_to_source.changes` to upload it.

PR #2298, opened 2019-04-10
Add support for running commands on BEL

### LINKED ISSUE #1072: Alacritty does not report BELs to X11
I realise that this is perhaps a *desired* feature for many people, who just use Alacritty's built-in visual bell and are happy about it, but a visual bell built into the terminal is, well, only noticed when you're looking at the terminal.

For quite a while now, I've been using [xvisbell](https://github.com/rianhunter/xvisbell) to flash the *entire* screen (not just my terminal) when an X11 bell is received, and further supplementing that using xkbevd to play an audio signal if a bell is received and the system is idle or the screen is locked (i.e., I still am signalled even when not paying attention to the system and won't see a visual bell).

Alacritty unfortunately breaks that workflow. Would a PR that modifies the `bell()` method on Handler to report them to X11 (in addition to the built-in visual bell functionality) be accepted?

### LINKED ISSUE #1528: Add support for audio bell
macOS High Sierra v10.13.6

I don't get any audio notification for the terminal bell. The config file has options for `visual_bell`, but that's of no help when I'm looking at other fullscreen windows.

For the time being, I've worked around the problem using tmux:
```tmux
# Some hacks for OSX Alacritty audio bell
set -g monitor-bell on
set -g bell-action other
set-hook -g alert-bell 'run-shell "afplay /System/Library/Sounds/Funk.aiff"'
```

Does Alacritty not provide an audio bell? If not, is the feature planned or out of scope?

robertgzr:
Adds a new configuration option under `visual_bell` called
`bell_command` that reuses the command formatting of the keybindings.

This allows for things like playing a notification sound.
See https://github.com/jwilm/alacritty/issues/1528#issuecomment-425691029

Closes #1528

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -597,7 +597,7 @@ pub struct Cursor {
     charsets: Charsets,
 }
 
-pub struct VisualBell {
+pub struct Bell {
     /// Visual bell animation
     animation: VisualBellAnimation,
 
@@ -608,7 +608,7 @@ pub struct VisualBell {
     start_time: Option<Instant>,
 
     /// Command to run when bell is rang
-    bell_command: Option<CommandWrapper>,
+    command: Option<CommandWrapper>,
 }
 
 fn cubic_bezier(p0: f64, p1: f64, p2: f64, p3: f64, x: f64) -> f64 {
@@ -618,19 +618,20 @@ fn cubic_bezier(p0: f64, p1: f64, p2: f64, p3: f64, x: f64) -> f64 {
         + x.powi(3) * p3
 }
 
-impl VisualBell {
-    pub fn new(config: &Config) -> VisualBell {
-        let visual_bell_config = &config.visual_bell;
-        VisualBell {
-            animation: visual_bell_config.animation,
-            duration: visual_bell_config.duration(),
-            bell_command: visual_bell_config.bell_command(),
+impl Bell {
+    pub fn new(config: &Config) -> Bell {
+        let bell_config = &config.bell;
+        Bell {
+            animation: bell_config.animation,
+            duration: bell_config.duration(),
+            command: bell_config.command.clone(),
             start_time: None,
         }
     }
 
     /// Ring the visual bell, and return its intensity.
     pub fn ring(&mut self) -> f64 {
+        self.run_command();
         let now = Instant::now();
         self.start_time = Some(now);
         self.intensity_at_instant(now)
@@ -715,8 +716,8 @@ impl VisualBell {
         }
     }
 
-    pub fn run_bell_command(&self) {
-        if let Some(ref cmd) = self.bell_command {
+    pub fn run_command(&self) {
+        if let Some(ref cmd) = self.command {
             match start_daemon(cmd.program(), cmd.args()) {
                 Ok(_) => debug!("Launched {} with args {:?} on bell", cmd.program(), cmd.args()),
                 Err(_) => {
@@ -727,10 +728,10 @@ impl VisualBell {
     }
 
     pub fn update_config(&mut self, config: &Config) {
-        let visual_bell_config = &config.visual_bell;
-        self.animation = visual_bell_config.animation;
-        self.duration = visual_bell_config.duration();
-        self.bell_command = visual_bell_config.bell_command();
+        let bell_config = &config.bell;
+        self.animation = bell_config.animation;
+        self.duration = bell_config.duration();
+        self.command = bell_config.command.clone();
     }
 }
 
@@ -784,7 +785,7 @@ pub struct Term {
 
     pub dirty: bool,
 
-    pub visual_bell: VisualBell,
+    pub bell: Bell,
     pub next_is_urgent: Option<bool>,
 
     /// Saved cursor from main grid
@@ -938,7 +939,7 @@ impl Term {
             next_title: None,
             next_mouse_cursor: None,
             dirty: false,
-            visual_bell: VisualBell::new(config),
+            bell: Bell::new(config),
             next_is_urgent: None,
             input_needs_wrap: false,
             grid,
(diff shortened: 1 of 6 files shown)


robertgzr:
Opening this before I document the changes and add them to config file in case there are any comments ;)

src/term/mod.rs
```diff
@@ -632,6 +633,9 @@ pub struct VisualBell {
 
     /// The last time the visual bell rang, if at all
     start_time: Option<Instant>,
+
+    /// Command to run when bell is rang
+    bell_command: Option<CommandWrapper>,
```
nixpulvis: I'd call this just `command` we know it's a bell command.

src/term/mod.rs
```diff
@@ -737,10 +742,20 @@ impl VisualBell {
         }
     }
 
+    pub fn run_bell_command(&self) {
```
nixpulvis: Another small naming note, would simply `run`, `spawn` or `start` be better here? `..._bell_command` is again somewhat redundant.

src/config/mod.rs
```diff
@@ -249,6 +253,12 @@ impl VisualBellConfig {
     pub fn color(&self) -> Rgb {
         self.color
     }
+
+    /// Visual bell command
+    #[inline]
+    pub fn bell_command(&self) -> Option<CommandWrapper> {
+        self.bell_command.clone()
+    }
```
nixpulvis: We don's have getters for the other launcher (it's simply a `pub` field). I don't love the idea of having a lone `pub` field on the struct, but at the same time, I really would like to avoid `clone`. Thoughts?

chrisduerr: Using `.as_ref()` would probably work here to return an `Option<&CommandWrapper>`. No clones necessary.

HeyBanditoz:
Do you still intend to work on this?

src/config/mod.rs
robertgzr: I'm not quite sure how to do this? surely you don't want me to litter all the structures with lifetime annotations just because of this?

alacritty_terminal/src/term/mod.rs
```diff
@@ -614,18 +618,20 @@ fn cubic_bezier(p0: f64, p1: f64, p2: f64, p3: f64, x: f64) -> f64 {
...
+            animation: bell_config.animation,
+            duration: bell_config.duration(),
+            command: bell_config.command.clone(),
             start_time: None,
         }
     }
 
     /// Ring the visual bell, and return its intensity.
     pub fn ring(&mut self) -> f64 {
+        self.run_command();
```
robertgzr: I moved the running of the command to the start of the `ring()` method, instead of calling it separate from that...

robertgzr:
re-ping @chrisduerr @nixpulvis in case this slipped through your queues ;)

alacritty_terminal/src/term/mod.rs
chrisduerr: A newline after `self.run_command()` would be nice to make it clear it's logically separated from the rest of the bell logic.

alacritty_terminal/src/term/mod.rs
```diff
@@ -614,18 +618,20 @@ fn cubic_bezier(p0: f64, p1: f64, p2: f64, p3: f64, x: f64) -> f64 {
...
+        let bell_config = &config.bell;
+        Bell {
+            animation: bell_config.animation,
+            duration: bell_config.duration(),
+            command: bell_config.command.clone(),
             start_time: None,
         }
     }
 
     /// Ring the visual bell, and return its intensity.
```
chrisduerr: This should be changed from visual bell to just bell, since it also executes the bell command.

alacritty_terminal/src/term/mod.rs
```diff
@@ -706,14 +712,26 @@ impl VisualBell {
                 // Since we want the `intensity` of the VisualBell to decay over
                 // `time`, we subtract the `inverse_intensity` from 1.0.
                 1.0 - inverse_intensity
-            },
+            }
+        }
+    }
+
+    pub fn run_command(&self) {
```
chrisduerr: This function name does not correctly represent what is done inside the function. A generic command execution should not print `bell` errors.

I'd recommend inlining this into the `run` method.

alacritty_terminal/src/term/mod.rs
```diff
@@ -706,14 +712,26 @@ impl VisualBell {
...
+                }
+            }
         }
     }
 
     pub fn update_config(&mut self, config: &Config) {
-        let visual_bell_config = &config.visual_bell;
-        self.animation = visual_bell_config.animation;
-        self.duration = visual_bell_config.duration();
+        let bell_config = &config.bell;
```
chrisduerr: Defining this variable here has no benefits, since `config.bell` and `bell_config` are exactly the same length.

src/config/mod.rs
nixpulvis: @robertgzr like this:

```rust
struct A(Option<u32>);

impl A {
    fn get(&self) -> Option<&u32> {
        self.0.as_ref()
    }
}

fn main() {
    println!("{:?}", A(Some(10)).get());
    println!("{:?}", A(None).get());
}
```

alacritty_terminal/src/term/mod.rs
nixpulvis: I personally don't mind a `Bell::run_command` function like this, though I don't care much either way since `Bell::ring` is so short.

If this were to be left as it's own function it could be hidden as an internal detail by removing the `pub` declaration.

XVilka:
There are many conflicts to be fixed, please rebase.

kchibisov:
While I like the idea of running commands on BEL, I don't really like the proposed integration into `alacritty`, however I've created a proposal on how an integration of this and similar features should be performed in alacritty in some kind of maintainable way. see #3182

chrisduerr:
> While I like the idea of running commands on BEL, I don't really like the proposed integration into alacritty, however I've created a proposal on how an integration of this and similar features should be performed in alacritty in some kind of maintainable way. see #3182

Due to this improved design for handling things like this and the amount of time this PR has been stale without any interest in it, I'm going to close this PR.

If there is interest in picking this back up, I'd recommend starting out with the new design suggested in #3182. If there are any uncertainties, please leave them in that issue.

PR #2305, opened 2019-04-12
Call TIOCSWINSZ only on grid change

### LINKED ISSUE #2177: Resizing calls TIOCSWINSZ for every pixel
Linux, Wayland.

When resizing something, alacritty calls TIOCSWINSZ even when padding was just added/removed. This causes processes to update their screens a lot, and this is slow in some cases, especially over ssh (try it with `watch date +%s%N`). TIOCSWINSZ does not use the pixel arguments, according to documentation.
It might be better to only call it when the grid size changes.

ekarbe:
Instead of calling TIOCSWINSZ for every pixel change it will now be called only on changes to the grid size. This should reduce screen refreshes.

This is my first pull-request for this project and first time really working with Rust. I'm open to improvements and hope that my code will be at least a bit useful.

This fixes #2177

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -388,8 +388,8 @@ impl Display {
             let height = psize.height as f32;
             let cell_width = self.size_info.cell_width;
             let cell_height = self.size_info.cell_height;
-            let cols = self.size_info.cols().0;
-            let lines = self.size_info.lines().0;
+            let previous_cols = self.size_info.cols().0;
+            let previous_lines = self.size_info.lines().0;
 
             self.size_info.width = width;
             self.size_info.height = height;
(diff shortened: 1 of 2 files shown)


src/tty/unix.rs
```diff
@@ -58,6 +61,11 @@ fn make_pty(size: winsize) -> (RawFd, RawFd) {
     win_size.ws_xpixel = 0;
     win_size.ws_ypixel = 0;
 
+    unsafe {
+        OLD_ROW = win_size.ws_row;
+        OLD_COL = win_size.ws_col;
+    };
```
chrisduerr: In Rust, things are usually marked as `unsafe` for a reason and adding unsafe code when there is an alternative is strictly discouraged.

In a multi-threaded context, this could easily cause problem which is why variables should be stored on structs instead of the global scope.

ekarbe: Thanks for your feedback.

I've adapted everything to your proposals.

src/display.rs
```diff
@@ -388,6 +388,8 @@ impl Display {
             let height = psize.height as f32;
             let cell_width = self.size_info.cell_width;
             let cell_height = self.size_info.cell_height;
+            let cols = self.size_info.cols().0;
+            let lines = self.size_info.lines().0;
```
chrisduerr: I think naming this something like `old_*` or `previous_*` instead of just cols/lines would make it a little more obvious what's going on here.

PR #2308, opened 2019-04-12
feat: add proper fullscreen support

### LINKED ISSUE #34: Proper full screen support

### LINKED ISSUE #2012: Implement Alt + Enter to fullscreen on Windows
\- Feature request -

Alt + Enter is a standard, expected shortcut on Windows systems to fullscreen the active application.
It is supported by the default conhost terminal too - it should be enabled by default in alacritty as well.

I often use it when I don't want to be distracted by other stuff that's open on my monitor,
it would be fantastic if you could add it - thanks a lot!

acheronfail:
This adds fullscreen support to Alacritty.
It also adds a "simple" fullscreen support (the highly requested feature for macOS where the fullscreened window doesn't occupy its own space).

Added platform keybindings:

* macOS: <kbd>ctrl</kbd> + <kbd>cmd</kbd> + <kbd>F</kbd> (native fullscreen)
* windows: <kbd>Alt</kbd> + <kbd>Enter</kbd>

Added actions (can be used to create your own keybindings):

* `ToggleFullscreen`:
    * this toggles the native fullscreen on each platform
* `ToggleSimpleFullscreen`:
    * _on macOS_, this toggles the nice "simple" fullscreen (no space required)
    * _non macOS_, this is just an alias for `ToggleFullscreen`. The rationale for this choice is so that the one `alacritty.yml` config file is still compatible cross-platform.

I've manually tested this on the following:

* macOS (mojave)
* Windows 10
* Linux (X11)

---

Fixes #34.
Fixes #2012.
This is different to #2157 (that one is for _starting_ in fullscreen mode).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,12 +21,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added ScrollLineUp and ScrollLineDown actions for scrolling line by line
-- Added ToggleFullscreen action (macos also has a ToggleSimpleFullscreen action
-    which allows switching to fullscreen without occupying another space)
+- Added ToggleFullscreen action
+- On macOS, there's a ToggleSimpleFullscreen action which allows switching to 
+    fullscreen without occupying another space
+- A new window option `startup_mode` which controls how the window is created
 
 ### Changed
 
 - Alacritty now has a fixed minimum supported Rust version of 1.31.0
+- The `start_maximized` window option is now `startup_mode: Maximized`
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added ToggleFullscreen action
+- On macOS, there's a ToggleSimpleFullscreen action which allows switching to 
+    fullscreen without occupying another space
+- A new window option `startup_mode` which controls how the window is created
+
 ### Changed
 
 - On Windows, Alacritty will now use the native DirectWrite font API
+- The `start_maximized` window option is now `startup_mode: Maximized`
 
 ## Version 0.3.2
 
@@ -21,15 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added ScrollLineUp and ScrollLineDown actions for scrolling line by line
-- Added ToggleFullscreen action
-- On macOS, there's a ToggleSimpleFullscreen action which allows switching to 
-    fullscreen without occupying another space
-- A new window option `startup_mode` which controls how the window is created
 
 ### Changed
 
 - Alacritty now has a fixed minimum supported Rust version of 1.31.0
-- The `start_maximized` window option is now `startup_mode: Maximized`
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added ScrollLineUp and ScrollLineDown actions for scrolling line by line
+- Added ToggleFullscreen action (macos also has a ToggleSimpleFullscreen action
+    which allows switching to fullscreen without occupying another space)
 
 ### Changed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -184,11 +184,11 @@ colors:
   #cursor:
   #  text: '0x000000'
   #  cursor: '0xffffff'
-  
+
   # Selection colors
   #
-  # Colors which should be used to draw the selection area. If selection 
-  # background is unset, selection color will be the inverse of the cell colors. 
+  # Colors which should be used to draw the selection area. If selection
+  # background is unset, selection color will be the inverse of the cell colors.
   # If only text is unset the cell text color will remain the same.
   #selection:
   #  text: '0xeaeaea'
@@ -449,6 +449,8 @@ alt_send_esc: true
 #   - Quit
 #   - ClearLogNotice
 #   - SpawnNewInstance
+#   - ToggleFullscreen
+#   - ToggleSimpleFullscreen # macOS only
 #   - None
 #
 # Values for `command`:
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -51,8 +51,16 @@ window:
   #     - buttonless: Title bar, transparent background, but no title bar buttons
   decorations: full
 
-  # When true, alacritty starts maximized.
-  start_maximized: false
+  # Startup Mode (changes require restart)
+  #
+  # Values for `startup_mode`:
+  #   - Windowed
+  #   - Maximized
+  #   - Fullscreen
+  #
+  # Values for `startup_mode` (macOS only):
+  #   - SimpleFullscreen
+  startup_mode: Windowed
 
 scrolling:
   # Maximum number of lines in the scrollback buffer.
@@ -450,8 +458,10 @@ alt_send_esc: true
 #   - ClearLogNotice
 #   - SpawnNewInstance
 #   - ToggleFullscreen
-#   - ToggleSimpleFullscreen # macOS only
 #   - None
+#
+# Values for `action` (macOS only):
+#   - ToggleSimpleFullscreen: Enters fullscreen without occupying another space
 #
 # Values for `command`:
 #   The `command` field must be a map containing a `program` string and
--- a/src/config/bindings.rs
+++ b/src/config/bindings.rs
@@ -179,7 +179,17 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
 
 #[cfg(not(any(target_os = "macos", test)))]
 pub fn platform_key_bindings() -> Vec<KeyBinding> {
-    bindings!(
+    let mut bindings = vec![];
+
+    #[cfg(all(target_os = "macos", not(test)))]
+    {
+        bindings.extend(bindings!(
+            KeyBinding;
+            Key::Return, [alt: true]; Action::ToggleFullscreen;
+        ));
+    }
+
+    bindings.extend(bindings!(
         KeyBinding;
         Key::V, [ctrl: true, shift: true]; Action::Paste;
         Key::C, [ctrl: true, shift: true]; Action::Copy;
@@ -189,7 +199,9 @@ pub fn platform_key_bindings() -> Vec<KeyBinding> {
         Key::Add, [ctrl: true]; Action::IncreaseFontSize;
         Key::Subtract, [ctrl: true]; Action::DecreaseFontSize;
         Key::Minus, [ctrl: true]; Action::DecreaseFontSize;
-    )
+    ));
+
+    bindings
 }
 
 #[cfg(all(target_os = "macos", not(test)))]
@@ -200,6 +212,7 @@ pub fn platform_key_bindings() -> Vec<KeyBinding> {
         Key::Equals, [logo: true]; Action::IncreaseFontSize;
         Key::Add, [logo: true]; Action::IncreaseFontSize;
         Key::Minus, [logo: true]; Action::DecreaseFontSize;
+        Key::F, [ctrl: true, logo: true]; Action::ToggleFullscreen;
         Key::K, [logo: true]; Action::ClearHistory;
         Key::K, [logo: true]; Action::Esc("\x0c".into());
         Key::V, [logo: true]; Action::Paste;
--- a/src/config/bindings.rs
+++ b/src/config/bindings.rs
@@ -178,18 +178,8 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
 }
 
 #[cfg(not(any(target_os = "macos", test)))]
-pub fn platform_key_bindings() -> Vec<KeyBinding> {
-    let mut bindings = vec![];
-
-    #[cfg(all(target_os = "macos", not(test)))]
-    {
-        bindings.extend(bindings!(
-            KeyBinding;
-            Key::Return, [alt: true]; Action::ToggleFullscreen;
-        ));
-    }
-
-    bindings.extend(bindings!(
+fn common_keybindings() -> Vec<KeyBinding> {
+    bindings!(
         KeyBinding;
         Key::V, [ctrl: true, shift: true]; Action::Paste;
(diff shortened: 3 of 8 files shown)


src/config/bindings.rs
```diff
@@ -177,9 +177,11 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
     bindings
 }
 
+#[allow(clippy::let_and_return)]
```
acheronfail: I had to add this override (and the `unused_mut` one below) because the `bindings` vec is only mutated on Windows. Open to suggestions on better ways to do this.

src/event.rs
```diff
@@ -281,6 +312,9 @@ pub struct Processor<N> {
     window_changes: WindowChanges,
     save_to_clipboard: bool,
     alt_send_esc: bool,
+    is_fullscreen: bool,
+    #[cfg(target_os = "macos")]
+    is_simple_fullscreen: bool,
```
acheronfail: I've opened another PR for `winit` which allows you to query the `Window` itself to see if it's fullscreen or not: https://github.com/tomaka/winit/pull/838.

If adding these flags to `Processor` isn't preferred, we can wait for that PR to be merged, and then wait for `glutin` to be updated with the next version of `winit` and then use the `window.get_fullscreen()` methods.

Although to be honest, I think this way is fine as it is.

src/window.rs
```diff
@@ -369,7 +369,7 @@ impl Window {
         self.window().hide();
     }
 
-    fn window(&self) -> &glutin::Window {
+    pub fn window(&self) -> &glutin::Window {
```
acheronfail: In order to toggle fullscreen, we needed access to the `glutin::Window` so I had to make this method `pub`.

I'm not sure if that's okay? We can always make it `pub(crate)` instead if that's preferred.

src/event.rs
```diff
@@ -546,8 +583,27 @@ impl<N: Notify> Processor<N> {
             window.hide();
         }
 
-        if self.window_changes.hide {
-            window.hide();
```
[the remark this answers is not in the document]
acheronfail: Ah, I meant to ask about this. Look at the lines above - is this meant to be there? Right now the two blocks are completely identical...

acheronfail:
> Please note that this does not fix #34 unless a config option is also implemented (unfortunately #2157 seems stale).

I think I might have misunderstood: right now there is a new action `ToggleFullscreen`. This can be used in `alacritty.yml` like:

```yml
- { key: F, mods: Command, action: ToggleFullscreen }
```

Unless there's something extra to add as well?

src/config/mod.rs
```diff
@@ -896,6 +896,9 @@ impl<'a> de::Deserialize<'a> for ActionWrapper {
                     "Quit" => Action::Quit,
                     "ClearLogNotice" => Action::ClearLogNotice,
                     "SpawnNewInstance" => Action::SpawnNewInstance,
+                    "ToggleFullscreen" => Action::ToggleFullscreen,
+                    #[cfg(target_os = "macos")]
+                    "ToggleSimpleFullscreen" => Action::ToggleSimpleFullscreen,
```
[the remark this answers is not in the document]
acheronfail: In terms of Apple's guidelines, they recommend only the "native" fullscreen which occupies its own space, and is quite obtrusive.

I implemented both here and made the both available because in #34 @jwilm suggests that [he wants the "simple" fullscreen mode](https://github.com/jwilm/alacritty/issues/34#issuecomment-271131103):

> At least on macOS, there's the automagical full screen button that does the trick. It does occupy another workspace, however, and it doesn't allow stacking the full screen application on top of other apps.
>
> That's precisely what this issue is for :)

And I've left in the "native" fullscreen mode so that Alacritty can follow Apple's guidelines and also have the standard fullscreen mode.

src/event.rs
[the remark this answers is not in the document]
chrisduerr: Oh, the github diff didn't include enough context here unfortunately.

That looks like a mistake. Thanks for removing it, I'm fine with handling it in this PR.

chrisduerr:
I'm talking about a config option to start the window as fullscreen window.

src/config/mod.rs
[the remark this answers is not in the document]
chrisduerr: Is there any plan to deprecate the old fullscreen mode? If Apple doesn't want to support it anymore, it seems like that needs to addressed with them and not in Alacritty.

Also is there any requirement to add keybindings for both, or is there already a default keybinding in macos that toggles the new fullscreen mode anyways?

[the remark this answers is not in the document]
acheronfail: Sorry, I was wrong to assume you were familiar with the fullscreen situation on macOS, let me explain:

Basically, macOS has three distinct fullscreen modes:

* Standard fullscreen (happens when you click the green fullscreen button)
    * Also called "native" fullscreen
    * This moves the window to its own space
    * There's no control over the animation or transition - all handled by macOS itself
    * This is the most "recent" of the fullscreen modes and was added to OSX around version 10.7
* [`enterFullScreenMode`](https://developer.apple.com/documentation/appkit/nsview/1483780-enterfullscreenmode?language=objc)
    * This takes over the monitor, so _no spaces or other applications may be used at the same time_ (think of how fullscreen mode works in kiosk applications)
    * Also completely controlled by macOS itself (ie: unconfigurable)
* "Simple Fullscreen"
    * The window just fills the screen completely (hiding the dock and menu bar)
    * This used to be how all fullscreen windows worked pre-macOS Lion
    * This is a popular feature, so much so that many apps have implemented it manually since Apple removed the functionality.
        * Some apps are: VLC, iTerm2, [The `electron` framework](https://electronjs.org/docs/api/browser-window#winsetsimplefullscreenflag-macos)
        * It's also worth noting that web browsers (Chrome, Firefox, Opera, etc) also implement this type of fullscreen mode when viewing videos and other such media
    * This means you can still use other applications, spaces, and all other functionality, while having your application in fullscreen mode.
    * This fullscreen mode is most similar to the fullscreen modes of Windows and Linux (using Gnome, etc).

So, there's no "deprecated" fullscreen mode, and macOS defaults to setting <kbd>ctrl</kbd> + <kbd>cmd</kbd> + <kbd>F</kbd> as the Standard Fullscreen option.

acheronfail:
> I'm talking about a config option to start the window as fullscreen window.

Ah, I see. I'll do some testing and try to include it in this PR as well. :)

acheronfail:
Regarding starting the window in fullscreen mode with a `start_fullscreen` option, it looks like `winit` currently has a limitation where it can only provide a list of monitors or the primary monitor.

At this stage I don't think there's any way to query which monitor is the "active" or "current" one.

This means that if we add the `start_fullscreen` option, then Alacritty will start on the "primary" monitor, not the "active" monitor, which may be confusing.

chrisduerr:
I'm fully willing to accept flickering at the start in favor of landing this. If we can't detect the display it's going to launch on, we can't and have to accept that and move on. Flickering into fullscreen shouldn't be too jarring and better than not supporting this at all.

So I'm in favor of ignoring the complexity of using `with_fullscreen` and just using `set_fullscreen` after the window has been mapped.

The thing I still do not want to concede is that this should be grouped together with `start_maximized` into a single configuration setting though. With proper deprecation warnings and fallback to the old value if it is set.

acheronfail:
> The thing I still do not want to concede is that this should be grouped together with start_maximized into a single configuration setting though. With proper deprecation warnings and fallback to the old value if it is set.

To clarify, are you saying that you don't want a new config option `start_fullscreen`, but that you want `start_maximized` to be grouped together? So that a value of `start_maximized` makes Alacritty start in fullscreen mode?

If so, I personally think that separate configuration values for maximized and fullscreen make more sense. 🤷‍♂️

chrisduerr:
No, there should be a separate option (something like `startup_mode` maybe) which takes the options `Fullscreen` and `Maximized` and the `start_maximized` option should be deprecated. If the `start_maximized` option is still specified, `startup_mode: Maximized` should be used regardless of what `startup_mode` is actually set to.

acheronfail:
Okay, I've deprecated `start_maximized` in favour of `startup_mode` which has three options:

* `Normal` just open the window normally
* `Maximized` start the window maximized (this replaces `start_maximized`)
* `Fullscreen` start the window fullscreen

src/window.rs
```diff
@@ -175,6 +175,13 @@ impl Window {
             window.set_position(logical);
         }
 
+        // TODO: once `winit` supports getting the active monitor before window creation
+        // use with `window.with_fullscreen()` to create the window on the right monitor.
```
chrisduerr: I'm not sure if this is actually planned due to the technical limitations of some platforms.

So I'd recommend just removing this comment and accepting this limitation as part of the design.

src/event.rs
```diff
@@ -546,8 +583,27 @@ impl<N: Notify> Processor<N> {
...
+
+            if self.window_changes.toggle_fullscreen && !self.is_simple_fullscreen {
+                self.set_fullscreen(window, !self.is_fullscreen);
+                self.is_fullscreen = !self.is_fullscreen;
+            }
+
+            if self.window_changes.toggle_simple_fullscreen && !self.is_fullscreen {
+                window.window().set_simple_fullscreen(!self.is_simple_fullscreen);
+                self.is_simple_fullscreen = !self.is_simple_fullscreen;
+            }
```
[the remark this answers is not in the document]
acheronfail: After testing some other apps, there are two common implementations:

1. the current implementation, which is:
    * `set_simple_fullscreen` only works if it's not currently `fullscreen`
    * `set_fullscreen` only works if it's not currently `simple_fullscreen`
2. choose either `simple_fullscreen` or `fullscreen` and only support one until the entire application is restarted (and the configuration option changed)

IMHO it makes the most sense the way it is, and it's the way that iTerm2 does it too. I think the way it is now is the way most macOS users would expect it to work.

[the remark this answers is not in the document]
chrisduerr: > IMHO it makes the most sense the way it is, and it's the way that iTerm2 does it too

Yeah that sounds like a good lead to follow, thanks!

alacritty.yml
```diff
@@ -51,8 +51,16 @@ window:
...
+  # Startup Mode (changes require restart)
+  #
+  # Values for `startup_mode`:
+  #   - Windowed
+  #   - Maximized
+  #   - Fullscreen
+  #
+  # Values for `startup_mode` (macOS only):
+  #   - SimpleFullscreen
+  startup_mode: Normal
```
chrisduerr: The default value needs to be changed to `Windowed` now.

CHANGELOG.md
```diff
@@ -9,12 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added ScrollLineUp and ScrollLineDown actions for scrolling line by line
-- Added ToggleFullscreen action (macos also has a ToggleSimpleFullscreen action
-    which allows switching to fullscreen without occupying another space)
+- Added ToggleFullscreen action
+- Added ToggleSimpleFullscreen action for macos which allows switching to fullscreen 
+    without occupying another space
```
chrisduerr: I'd recommend this to move the `macOS` mention to the front, since that way non-macos users don't have to read through half of it before they realize that it doesn't affect them.

I think some other entries in our change log have used `On macOS, ...` for this.

alacritty.yml
```diff
@@ -450,9 +458,11 @@ alt_send_esc: true
 #   - ClearLogNotice
 #   - SpawnNewInstance
 #   - ToggleFullscreen
-#   - ToggleSimpleFullscreen # macOS only
 #   - None
 #
+# Values for `action` (macOS only):
+#   - ToggleSimpleFullscreen: fullscreens without occupying another space
```
chrisduerr: I think using `fullscreens` as a verb doesn't sound great, `Enters fullscreen` sounds better to me I think.

src/config/bindings.rs
```diff
@@ -199,9 +189,18 @@ pub fn platform_key_bindings() -> Vec<KeyBinding> {
         Key::Add, [ctrl: true]; Action::IncreaseFontSize;
         Key::Subtract, [ctrl: true]; Action::DecreaseFontSize;
         Key::Minus, [ctrl: true]; Action::DecreaseFontSize;
-    ));
+    );
 
-    bindings
+    if cfg!(all(target_os = "windows", not(test))) {
```
chrisduerr: We shouldn't use `cfg!`, since this is a runtime check instead of a compile-time check.

I'd also moving this into the `platform_key_bindings (using a dedicated one for Windows), these were created exactly for the issue of test requiring no platform-specific bindings.

To prevent us from having to duplicate the windows and linux bindings since they're largely similar, we could put those in a separate method that is only offered on linux/bsd/windows and called from the more specific platform_bindings.

src/event.rs
```diff
@@ -546,8 +583,27 @@ impl<N: Notify> Processor<N> {
...
-        if self.window_changes.hide {
-            window.hide();
+        #[cfg(target_os = "macos")]
+        {
+            use glutin::os::macos::WindowExt;
+
+            if self.window_changes.toggle_fullscreen && !self.is_simple_fullscreen {
+                self.set_fullscreen(window, !self.is_fullscreen);
+                self.is_fullscreen = !self.is_fullscreen;
+            }
```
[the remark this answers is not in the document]
acheronfail: Actually, since for `macos` we need to first check `self.is_simple_fullscreen` I think the code is simplest if we just leave it as is.

Open to suggestions though :)

alacritty.yml
```diff
@@ -51,8 +51,8 @@ window:
   #     - buttonless: Title bar, transparent background, but no title bar buttons
   decorations: full
 
-  # When true, alacritty starts maximized.
-  start_maximized: false
+  # Valid options: `Normal`, `Maximized` or `Fullscreen`.
+  startup_mode: Normal
```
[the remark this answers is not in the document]
acheronfail: I like `Windowed` - I've changed it all to that now. 😄

src/event.rs
```diff
@@ -546,8 +568,25 @@ impl<N: Notify> Processor<N> {
...
+        }
+
+        #[cfg(not(target_os = "macos"))]
+        {
+            if self.window_changes.toggle_fullscreen {
+                window.set_fullscreen(!self.is_fullscreen);
+                self.is_fullscreen = !self.is_fullscreen;
+            }
         }
```
chrisduerr: I think I've mentioned this in a previous comment, but we should aim to merge the macos and non-macos targets together here, after all the body of this conditional is completely identical.

The easiest solution would probably be to just make `is_simple_fullscreen` a field that is available on every platform and just make it always false on linux/bsd/windows.

acheronfail: :+1:

PR #2311, opened 2019-04-13
Remove live shader reload feature

### LINKED ISSUE #1523: shaders not live reloading multiple times (with --features "live-shader-reload"). Config live reload works fine
`rx.recv` at https://github.com/jwilm/alacritty/blob/master/src/renderer/mod.rs#L595 seems to hang after a second update of the shader files

On freenode, UndeadLeech suggested I do a few things.
Here's what didn't work, if I applied the instructions correctly:
 - increasing `fs.inotify.max_user_instances` and `fs.inotify.max_user_watches` through `sysctl`
 - using nano to write to the shaders
 - increasing the delay to 3s on https://github.com/jwilm/alacritty/blob/master/src/renderer/mod.rs#L588

Here's what worked:
 - changing the code to watch based on file path and not inode
 - watching the res dir instead of the individual files

The confusing thing is that the watcher for the config live reload should be the same as the one for the shader live reload, so neither should work, or both should work, if it was a watcher issue

Ubuntu 16.04, i3, ext4, no DE

X11

chrisduerr:
This removes the live-shader-reload feature to make the renderer
significantly more concise and maintainable.

As far as I'm aware, this feature was never really used by anyone since
there isn't much of an advantage to recompiling Alacritty yourself just
to make live shader reload work.

This feature has also been broken for a long time, since the projection
is updated after the reload without considering the padding. Editing
with standard editors like vim also deletes the file and replaces it,
making it impossible to use them for live shader reload.

Closes #1523.

This is mostly a proposal at this point, since there is no *need* to remove
live shader reload at this point. However I currently do not see much
advantage in keeping this code around.

If anyone is actually using Alacritty with live shader reload enabled,
please let me know. I believe @towc is the only person to ever create
a live shader reload related issue. Do you still care about this?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,9 +75,6 @@ dirs = "1.0.2"
 
 [features]
 default = []
-# Enabling this feature makes shaders automatically reload when changed
-live-shader-reload = []
-nightly = []
 bench = []
 
 [build-dependencies]
--- a/src/display.rs
+++ b/src/display.rs
@@ -156,7 +156,7 @@ impl Display {
             window.inner_size_pixels().expect("glutin returns window size").to_physical(dpr);
 
         // Create renderer
-        let mut renderer = QuadRenderer::new(viewport_size)?;
+        let mut renderer = QuadRenderer::new()?;
 
         let (glyph_cache, cell_width, cell_height) =
             Self::new_glyph_cache(dpr, &mut renderer, config)?;
@@ -189,6 +189,8 @@ impl Display {
         }
 
         window.set_inner_size(viewport_size.to_logical(dpr));
+
+        // Update the shader projection
         renderer.resize(viewport_size, padding_x as f32, padding_y as f32);
 
         info!("Cell Size: ({} x {})", cell_width, cell_height);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,7 +14,6 @@
 //
 //! Alacritty - The GPU Enhanced Terminal
 #![deny(clippy::all, clippy::if_not_else, clippy::enum_glob_use, clippy::wrong_pub_self_convention)]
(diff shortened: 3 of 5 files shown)


towc:
I think it added a certain edge to this project that others just don't have. I still use it, but I can come up with alternatives :)

Thanks for asking for my opinion, it's really nice of you

nixpulvis:
I personally support the reduction in complexity this allows. I'm not sure how often anyone is tweaking shaders, but I've gotten pretty ok with the compile times for Alacritty anyway myself.

jansol:
I have live config reloading enabled and it's particularly nice when tuning background opacity. But the issue is a bit unclear on whether it's just about shaders (not sure bg opacity changes those) or the entire config reload feature.

chrisduerr:
Live config reload is not affected by this and I have absolutely no plan to ever remove that (quite the opposite, I'd rather make more things work with it).

This is only about shader reload.

CrackedP0t:
Hey - your description of this PR makes it sound like it removes live config reloading, entirely, not just shaders - you might want to edit that

chrisduerr:
Thanks a lot for pointing that out, looks like I've got shader and config reload mixed up a couple of times there.

chrisduerr:
I've thought about this quite a bit and came to the conclusion that it's just not worth it.

Even removing it completely as proposed in this PR would just remove ~100 lines and there are some things that are very useful that wouldn't be possible in this current form.

The main issue being that this would only allow changing it through compilation. I don't think live reloading is super important but I think that changing the shaders at runtime vs compile time definitely has some benefits.

I still don't think this is used a lot, but I don't want to get rid of it just to get rid of it. If there are some more convincing arguments in the future it would likely make sense to revisit this, but right now I feel like we'd mostly lose something, without directly gaining much.

PR #1023, opened 2018-01-13
[very WIP] Make Alacritty into a library

tbodt:
My work on #450. As far as I can tell very little progress has been made on that issue, so I'm posting this just to get a bit of code out there, that maybe others can build on. Feedback would be appreciated.

PR #2312, opened 2019-04-14
Split alacritty main into a separate crate

tbodt:
Spiritual successor of #1023. The eventual goal is to have alacritty-core not depend on glutin or winit.

@chrisduerr Would you be willing to merge this in the current state? I'd prefer to avoid what happened with the last attempt, where the final goal was so far off that it was never merged and eventually it got impossible to rebase.


Cargo.toml
```diff
@@ -1,26 +1,22 @@
 [package]
-name = "alacritty"
+name = "alacritty-core"
 version = "0.3.0"
 authors = ["Joe Wilm <joe@jwilm.com>"]
 license = "Apache-2.0"
 build = "build.rs"
-description = "GPU-accelerated terminal emulator"
+description = "GPU-accelerated terminal emulator - library for embedders"
```
chrisduerr: This is a tiny nitpick, but if the plan is to separate the frontend from the backend completely, there's no more GPU acceleration in alacritty-core and advertising it as such would likely be confusing.

I'd recommend just switching it to something like `Library for writing terminal emulators` (suggestions welcome!).

Cargo.toml
```diff
@@ -1,26 +1,22 @@
 [package]
-name = "alacritty"
+name = "alacritty-core"
```
chrisduerr: I think calling this `core` is a decent choice, though it's not the most general name (people might not realize that it can be used outside of Alacritty).

Generally I also think that while I think that `-` to separate words looks better, using `_` just makes more sense since it will need to be used as such in the source code anyways.

Cargo.toml
```diff
@@ -1,26 +1,22 @@
...
 build = "build.rs"
-description = "GPU-accelerated terminal emulator"
+description = "GPU-accelerated terminal emulator - library for embedders"
 readme = "README.md"
 homepage = "https://github.com/jwilm/alacritty"
 edition = "2018"
 
 [workspace]
 members = [
+    "app",
```
chrisduerr: I think `app` isn't great either. I feel like just calling this directory `alacritty` would be good. That should be a nice uniform name for final application and would likely make it easiest for contributors to find the main source files for the frontend.

app/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+version = "0.3.0"
+authors = ["Joe Wilm <joe@jwilm.com>"]
+license = "Apache-2.0"
+description = "GPU-accelerated terminal emulator"
+readme = "../README.md"
+homepage = "https://github.com/jwilm/alacritty"
+edition = "2018"
+
+[dependencies]
+alacritty-core = { path = ".." }
```
chrisduerr: Currently this makes the root of the project the core alacritty library. However I feel like organization becomes kinda cluttered at that point. I think moving core to `alacritty-core` directory so the root doesn't have any source directory would make things a bit more uniform.

Similar to what [glutin](https://github.com/tomaka/glutin) and many others do already.

Cargo.toml
nixpulvis: I'm not giving this as much thought right now as it deserves, but I'll just say that in general, the `-core` notion works pretty well when there's high coupling, and generally a "porcelain" and "plumbing" style to the code. Here I think we have something else...

Even `alacritty-backend` might make more sense. Thoughts?

Cargo.toml
tbodt: In my view of things, the core would in fact have the GPU acceleration code. Embedders would activate an OpenGL context and then call the core library to render a frame.

Cargo.toml
tbodt: That sounds better to me as well.

Cargo.toml
nixpulvis: lol, fwiw the more I think about "backend" the more I don't like it...

tbodt: I'm starting to think alacritty-core or alacritty-backend is too generic for all the things it would be doing. Maybe we should split it into more crates:

- alacritty-term
- alacritty-render
- alacritty-ptys

chrisduerr: `alacritty-backend` also seems reasonable, but I think I prefer core since it sounds cooler and it's shorter.

However, I like @tbodt's proposal. I think `alacritty-term` would be a better name than `alacritty-core` because it doesn't sound like it's necessarily Alacritty-internal but it's still short and describes well what it does.

My biggest gripe with that would be that people that don't see the other directories like `alacritty-render` might think that `alacritty-term` is the complete terminal emulator.

Regarding the other names I'm not a huge fan of `alacritty-render` (I can already see me typing renderer and having to correct that) and `alacritty-ptys` should definitely be singular, but since this isn't necessary right now, it's not too important so I'd like to figure out the primary two names first (UI and Term).

I'd also like to mention that using just `alacritty` for frontend is probably a bad idea, since most of the code is likely going to be in the terminal part. So I'm quite happy with the `alacritty-term` name the more I think about it, but the frontend might still need some consideration.

Maybe something like `alacritty-ui`?

nixpulvis: I'm currently in favor of `alacritty-terminal` and `alacritty-ui`, after much bikeshedding with @chrisduerr in IRC. I also belive personally that `Term` would be better of as `Terminal`, but I don't have super strong feelings.

Cargo.toml
chrisduerr: > In my view of things, the core would in fact have the GPU acceleration code

That seems like an odd choice. I feel like one of the reasons for separating things is so we can have frontends that do not use Alacritty's OpenGL rendering code. This would allow using Alacritty for various things other than writing your own terminal emulator too (like writing a multiplexer).

chrisduerr:
I've created an alternative proposal here:
https://github.com/tbodt/alacritty/pull/1

I've tried to work my proposals into a kind of personal ideal, however that doesn't mean it's perfect.

A big issue that becomes clear with separation like this is that it definitely makes development harder, which is why I haven't tried to remove the binary executable from the root completely and moved it into an `alacritty` subdirectory, but I think there's no way around a little bit of hassle.

chrisduerr:
I've actually gone over things again and it seems like creating a separate module for the entry point isn't too bad and actually works pretty well. So I've updated the PR so no code is in the root of the project anymore and everything has been moved to the `alacritty` module.

Cargo.toml
tbodt: I'm interested in using alacritty as the renderer for my terminal-like iOS app, so I would want to include the GPU code. Other use cases include terminal apps that, say, have a native settings UI. You're right about there being use cases with no graphics, though. I'll go back to the suggestion of an `alacritty_ui` crate for the graphics stuff, so you can use the grid-related code separately.

chrisduerr: So maybe an `alacritty_ui` crate, an `alacritty_terminal` crate and the `alacritty` crate which is just superglue to hook them both up to each other.

If we're talking priority I'd definitely rank the terminal usecase over the UI one though, since Alacritty's UI is far less complex than what the terminal does. The UI is also very poorly suited for anything other than a terminal emulator due to the limitations that have been imposed to improve performance.

So my main concern is that it wouldn't be possible to have good abstractions for both without any performance hit (which is an absolute requirement), however if both can be pulled out I'd be fine with that too.

Note that my branch already separates an `alacritty` and `alacritty_terminal` subcrate, so I think that would be a decent starting point and first incremental step for this.

tbodt: Agreed, I was going to suggest keeping UI in the alacritty crate for the time being, since that's harder to separate. The terminal part is already fairly cleanly separated from things that depend on UI, I think.

What exactly should the scope be for alacritty_terminal, though? I'll propose something in the main comment thread, since this is getting pretty long for a code comment.

chrisduerr: > I'll propose something in the main comment thread, since this is getting pretty long for a code conversation.

That's what I was about to recommend too. Keeping things in the main comment thread will likely make it a lot easier to follow the discussion, thanks.

chrisduerr:
> What exactly should the scope be for alacritty_terminal, though?

Alacritty terminal should be completely decoupled from anything UI related. The primary goal for this would be to be able to write any terminal multiplexer/emulator in it, regardless of curses/opengl/vulkan or any other UI that should be used for it.

Concretely this means that there should be absolutely no dependency on glutin, winit, openGL or copypasta.

I'm not entirely sure about separation of the font crate, since we rely a lot on metrics and stuff like that, but this would probably need a bit of investigation. In theory I don't think the font crate should be necessary either.

tbodt:
I went through the list of files in alacritty and tried to decide which ones should go in which crate.

Files for `alacritty_terminal`:

- grid/
- renderer/: alacritty
- term/
- tty/
- ansi.rs
- event_loop.rs
- index.rs
- selection.rs

Files for `alacritty`:

- cli.rs
- config.rs
- display.rs
- event.rs
- input.rs
- locale.rs
- logging.rs
- meter.rs
- panic.rs
- window.rs

How's that sound?

I wasn't quire sure about sync.rs or util.rs.

chrisduerr:
I think in general that seems like a reasonable approach for an initial separation. Though there are some things where I would probably switch things up a bit (selections are more a UI thing, same for url.rs which is missing in your list).

Further separations that might be possible would be to have the `renderer` and similar stuff in `alacritty_ui` and the `tty` stuff in an `alacritty_tty` subcrate.

To start things out I'd keep everything in the `alacritty_terminal` crate for now (similar to what I've done in my branch) and then try to pull out one thing after the other from the stuff mentioned in your list. Once it's possible to use `alacritty_term` without any of Alacritty's GUI code, we can start separating stuff from `alacritty` into `alacritty_ui` and just keep going like that until everything is separated (ofc some of these things can be done in parallel too).

chrisduerr:
@tbodt At this point I'd be curious about feedback on my branch as a first step. I think it's a little closer to the plan than this PR currently. Though I haven't put a lot of time into it either so there's basically no separation yet.

But in general I think the sooner we can start the first step, the easier it will be to iterate on it without keeping this PR open forever.

Generally I'd like to release 0.3.1 pretty soon-ish so I think it would be nice to have the first step(s) of this in the 0.3.2 release (without anything usable at that point of course).

tbodt:
The project structure from your PR looks good to me. I'll do some kind of rebase with that structure tomorrow.

tbodt:
Apparently I was wrong about "tomorrow" (I blame game of thrones). Just pushed a new branch to this PR based on your suggestion for source layout. Turns out 5 days is plenty of time to make an automatic rebase near impossible, so let's try and get this merged now.

alacritty_terminal/Cargo.toml
```diff
@@ -0,0 +1,95 @@
+[package]
+name = "alacritty_terminal"
+version = "0.3.2"
+authors = ["Joe Wilm <joe@jwilm.com>"]
+license = "Apache-2.0"
+build = "build.rs"
+description = "GPU-accelerated terminal emulator - library for embedders"
```
chrisduerr: I think I've mentioned this previously but you've probably missed it when including my changes, but this library is completely unrelated to GPU acceleration.

In my PR I've used the following description:
```
description = "Library for writing terminal emulators"
```

I have added a bunch of little things to the `Cargo.toml` files like keywords and categories that should help with separating the different subcrates, even though it's not necessary I think these might be helpful to just add here.

alacritty_terminal/Cargo.toml
```diff
@@ -0,0 +1,95 @@
...
+image = "0.21.0"
+
+[target.'cfg(target_os = "macos")'.dependencies]
+objc = "0.2.2"
+
+[features]
+default = []
+# Enabling this feature makes shaders automatically reload when changed
+live-shader-reload = []
+nightly = []
```
chrisduerr: I've removed this from my branch because it's not useful anymore, but to keep things completely uncontroversial and not add any actual source changes I think it's a good idea to handle this separately.

So even if you apply the changes I've made to the config, I think this can stay the way you have it.

alacritty_terminal/Cargo.toml
```diff
@@ -0,0 +1,95 @@
...
+assets = [
+    ["target/release/alacritty", "usr/bin/", "755"],
+    ["extra/linux/alacritty.desktop", "usr/share/applications/", "644"],
+    ["extra/logo/alacritty-term.svg", "usr/share/pixmaps/Alacritty.svg", "644"],
+    ["extra/completions/alacritty.bash", "usr/share/bash-completion/completions/alacritty", "644"],
+    ["extra/completions/alacritty.fish", "usr/share/fish/completions/alacritty.fish", "644"],
+    ["extra/completions/_alacritty", "usr/share/zsh/vendor-completions/_alacritty", "644"],
+    ["extra/alacritty.info", "usr/share/terminfo/a/alacritty", "644"],
+]
+maintainer-scripts = "extra/linux/debian"
```
chrisduerr: My Cargo.toml has this moved in the `alacritty` subcrate. I've actually made sure that debian still builds with my fork by test-building it once so it's probably a good idea to follow its lead in this regard.

Since we can only create debs from binaries, this isn't useful here.

chrisduerr:
I'd like to get this merged asap as the next PR, so we won't have to rebase it again. Let me know if you don't have the time to work on my suggestions within the next few days and I'll apply them myself to get things going a bit quicker.

tbodt:
Tomorrow morning I'll either do it or let you know that you'll have to do it.

chrisduerr:
Thanks! No worries either way, I'd just like to know.

tbodt:
cargo test for me does seem to test everything. cargo run runs the alacritty crate.

alacritty_terminal/Cargo.toml
tbodt: Done

alacritty_terminal/Cargo.toml
tbodt: Done

tbodt:
The window.rs thing was an overzealous git add. I'll fix that and also fix CI.

chrisduerr:
Oh yeah, it looks like the tests do work, maybe they just didn't before I made the main Cargo.toml completely virtual (so when there was still a `src` in the root directory).

However, I noticed that the ref tests are not run. The `tests` directory needs to be moved to `alacritty_terminal`.

chrisduerr:
Also it looks like CI fails because of the copypasta subcrate. This was expected and all the tests can just be deleted.

https://github.com/jwilm/alacritty/pull/2207 will remove the copypasta crate entirely anyways.

alacritty/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+assets = [
+    ["target/release/alacritty", "usr/bin/", "755"],
+    ["extra/linux/alacritty.desktop", "usr/share/applications/", "644"],
+    ["extra/logo/alacritty-term.svg", "usr/share/pixmaps/Alacritty.svg", "644"],
+    ["extra/completions/alacritty.bash", "usr/share/bash-completion/completions/alacritty", "644"],
+    ["extra/completions/alacritty.fish", "usr/share/fish/completions/alacritty.fish", "644"],
+    ["extra/completions/_alacritty", "usr/share/zsh/vendor-completions/_alacritty", "644"],
+    ["extra/alacritty.info", "usr/share/terminfo/a/alacritty", "644"],
+]
+maintainer-scripts = "extra/linux/debian"
```
chrisduerr: These paths all need to be changed. See https://github.com/chrisduerr/alacritty/blob/alacritty-split/alacritty/Cargo.toml#L41-L50 and the live shader reload feature needs to get passed through.

I really recommend taking a look at the Cargo.tomls from my branch, these have been tested to work with all features, tests and tools.

tbodt:
Went through your Cargo.toml files and everything should be good now. Hopefully windows should also be working now, but we'll have to wait for the CI to finish on that.

tbodt:
Now it can't find winpty-agent.exe. I don't have any way to test windows so I won't be able to fix that.

chrisduerr:
Looks like the build file from winpty-agent was just completely broken and would have never worked, so since all subcrates are tested now this showed up.

I've tried to fix the build on windows, hopefully CI will not complain.

PR #2329, opened 2019-04-18
Scroll visible region into history on `CSI 2 J`.

### LINKED ISSUE #2199: Ctrl-L should at least optionally not remove contents from the scrollback buffer
Which operating system does the issue occur on? Ubuntu 18.04

I will often clear my terminal screen just so I can start issuing commands with a fully empty page. I like however to maintain the full screen history just the same. gnome-terminal reacts to Ctrl-L by actually adding empty lines in the scrollback. alacritty will remove the lines cleared from the screen from the scrollback. Apparently both are valid behaviors:

https://github.com/kovidgoyal/kitty/issues/268#issuecomment-355764257

Ideally I'd like a third behavior where just the position in the buffer changes and Ctrl-L is just like a scroll down past the current point so the rest of the terminal below the current position is still empty.

nixpulvis:
Upon pressing <kbd>ctrl-L</kbd> in many terminal applications (including most shells) the ANSI escape command `CSI 2 J` is triggered to clear the entire screen. There are various implementations of this in the wild, however we've decided the best course of action should be to retain all output by scrolling the visible region up each time. This ensures you never lose anything on the screen with `CSI 2 J`. Then to clear the history another escape, `CSI 3 J` can be used, as `clear` does.

### TODO
- [x] `clear` no longer clearing all history, last page preserved (might need to be config?)
- [x] `clear` causes a full page of scrollback history each time (most other terminals also do strange things)
- [x] Scrolling regions (`printf "\e[1;5r" `)
- [x] Broken ref tests

fixes #2199


alacritty_terminal/src/term/mod.rs
```diff
@@ -1909,17 +1909,6 @@ impl ansi::Handler for Term {
...
-            ansi::ClearMode::Below => {
-                for cell in &mut self.grid[self.cursor.point.line][self.cursor.point.col..] {
-                    cell.reset(&template);
-                }
-                if self.cursor.point.line < self.grid.num_lines() - 1 {
-                    self.grid
-                        .region_mut((self.cursor.point.line + 1)..)
-                        .each(|cell| cell.reset(&template));
-                }
-            },
```
nixpulvis: Just reordering above, above below ;)

alacritty_terminal/src/term/mod.rs
```diff
@@ -1934,7 +1923,26 @@ impl ansi::Handler for Term {
...
+            ansi::ClearMode::Below => {
+                for cell in &mut self.grid[self.cursor.point.line][self.cursor.point.col..] {
+                    cell.reset(&template);
+                }
+                if self.cursor.point.line < self.grid.num_lines() - 1 {
+                    self.grid
+                        .region_mut((self.cursor.point.line + 1)..)
+                        .each(|cell| cell.reset(&template));
+                }
+            },
```
nixpulvis: Just reordering above, above below ;)

nixpulvis:
### Reference

##### Escapes

- `H`: Goto 0,0
- `0J`: clear below
- `1J`: clear above
- `2J`: clear all
- `3J`: clear saved

##### Application Support

```
`clear`:    3J H 2J
`clear -x`:    H 2J
^L:            H 2J
```

### Today

##### ^L (and `clear -x`)

We currently "clear" by deleting the visible region on 2J so all current work is not moved into the history. You can see this by opening a new (master) alacritty, `date`, then ^L. The date will be lost.

##### `clear`

We first delete history when we receive 3J, then we delete the visible region when we receive 2J. Just like the comment above, since we don't save the visible region on 2J into history, we don't see anything in history. Everything is cleared, and this matches my expectations.

### Tomorrow

We want to allow users to save all output with ^L, meaning that it should no longer delete the text, but shift it into the history buffer. Now we're stuck with a few options, none seem perfect or feasible.

1. Clear *saves* the current view in fact, but does clear what *was* in history and leaves you with a blank screen.

2. We break the granularity of `ClearMode::Saved` and have it also clear the visible region. This would make clear a reliable means to remove all text, like today, while still allowing for ^L to save all history.

There may be a way to fix this somehow with a custom clear entry in out terminfo?

I've also found a wealth of information on this issue here: https://unix.stackexchange.com/questions/375743/why-clear-do-not-clear-whole-screen/375753 most notably https://bugzilla.gnome.org/show_bug.cgi?id=506438.

chrisduerr:
@nixpulvis Thanks a bunch for linking the different discussions, the Konsole bug actually solves everything we need to know.

> Though I'm really curious why I haven't read any suggestions to change clear to be H 2J 3J since I think that avoids this whole mess.

This seems to be exactly what tmux creator thought too, so a patch has been sent to ncurses which has been merged since august of 2018. So now the order of events is `H 2J 3J` fixing all problems with `clear` not doing what one might expect.

This leaves the question of how to handle the ^L behavior and I think we've got the option to do better than anyone else here. I'm not sure why this isn't simple to implement for GNOME Terminal, but for Alacritty it should be trivial to determine how many lines of the visible area are occupied. So we have the choice to always scroll up just a minimal amount, regardless of current cursor position. Hopefully that should solve the issue with big blocks of whitespace in the history.

Generally I was a bit concerned this might break certain applications, but it doesn't seem like it judging by that conversation. And generally it should be fine as long as everything visible is removed, so we should be good to go moving forward like this.

nixpulvis:
For anyone on Arch Linux (or able to use `makepkg`) here is a `PKGBUILD` with the latest `ncurses` release. Using this we can test the proper behavior of `clear`.

nixpulvis:
@chrisduerr OK, I've made some progress here, and I know I can improve the logic to avoid checking every cell like it is currently. It is mostly working as is though.

I am seeing a very strange behaviour I'm having trouble figuring out. ^L seems to be working both with nothing in the scrollback and when there are lines in the scrollback. And it's correctly only saving the lines with nonempty cells. However, if I use `yes` to flood the scrollback both `clear` and ^L leave `y` in line's first column coming in from the bottom of the grid.

chrisduerr:
@nixpulvis This sounds like an issue of the raw buffer wrapping around, which might indicate an issue with the `zero` in `storage.rs`, or you're rotating old data into the screen. Since the raw buffer might be bigger than the actual buffer, because resizes do not always truncate the buffer, this can lead to weird results during rotation with incorrect zero positions. You can call `truncate` in storage to clear the extra bits and see if that messes with things. Otherwise it's probably just a `zero` issue I'd assume, maybe you're rotating too far?

alacritty_terminal/src/grid/storage.rs
```diff
@@ -221,8 +237,9 @@ impl<T> Storage<T> {
         }
     }
 
+    // TODO: The a & b lines here seem to start from 0 at the top and grow downwards.
     pub fn swap_lines(&mut self, a: Line, b: Line) {
-        let offset = self.inner.len() + self.zero + *self.visible_lines;
+        let offset = self.zero + *self.visible_lines;
```
nixpulvis: @chrisduerr this seems like something that should be safe to remove since `x + y % x == y % x`.

alacritty_terminal/src/grid/storage.rs
```diff
@@ -312,6 +336,75 @@ mod test {
...
+    #[test]
+    fn rotate() {
+        let mut storage = Storage::with_capacity(Line(3),
+                                                 Row::new(Column(0), &' '));
+        storage.rotate(2);
+        assert_eq!(storage.zero, 2);
+        storage.shrink_lines(2);
+        // TODO: Shouldn't `shrink_lines` update the value of `zero`?
+        assert_eq!(storage.len, 1);
+        assert_eq!(storage.zero, 0);
```
nixpulvis: @chrisduerr I'm not 100% sure what I expect here. Thoughts?

(This is probably unrelated to me issues with the PR atm I fear however)

chrisduerr: Why would zero move? It just reduces the number of elements. All that requires to do is reduce the `len` and we should be good, the rest is then automatically treated as excess buffer.

Zero is reset though when more than a certain number of lines have been removed, to prevent excessive memory consumption.

nixpulvis: That relies on every use of `zero` correctly handling the case where it's outside the inner length. But this does seem to be the case, so ignore this comment.

nixpulvis:
Just a little update on this, since I've been rather slow in fixing all the remaining issues. It's finally at a point where at least the primary feature works.

In this recording you can see that at first I press <kbd>ctrl-L</kbd> and it properly saves the history without any empty lines, and then `clear` fully deletes the history. **Note:** This is running on my system with the updated `ncurses` installed.

clear

chrisduerr:
I've looked into the failing reftests and I think I've figured out why things are going wrong here.

### Issue 1

The first issue is very simple, this fixes all tests other than the `vttest_origin_mode_*` tests. The problem you have is that when `clear`ing, you're expected to reset all cells to the background color (which might be a color other than the normal terminal background).

However, the `scroll_up` method calls `reset` on each row, which only resets lines which are occupied. This fails when you want to clear the background to a specific color (in vim for example), because the cells might not contain any relevant info (they're empty), but you still need to redraw them.

This can be fixed by just setting the `occ` of all rows to the number of columns before clearing the viewport. Not sure if that's the best possible solution, however it should hopefully get the point across what the underlying issue is.

### Issue 2

This one's a bit tougher. It took me a while to figure out what exactly is going on and I'm still not entirely certain. This affects only the `vttest_origin_mode_*` benchmarks.

There are two ways in which this can be easily fixed. Either we reset all cells of the last line after calling `clear_viewport` (last line being `Term::lines().0 - 1`), or we just subtract 1 from `bottom_nonempty_line` in the `clear_viewport` method.

I haven't really looked into this, but to me this indicates that the `bottom_nonempty_line` calculation isn't entirely correct. Note that in both cases the `bottom_nonempty_line` is `1` and it's fixed when it is set to `0`. So maybe there's some special case when `bottom_nonempty_line` is `1` and there's a good reason why it needs to be `0`. Alternatively maybe it's just not tracking things correctly. I haven't really looked a ton into this but hopefully this should help.

If you have any questions about my findings, feel free to ask me @nixpulvis. But I thought I'd give things a look to help out a bit. :)

alacritty_terminal/src/grid/mod.rs
```diff
@@ -508,11 +558,33 @@ impl<T: GridCell + Copy + Clone> Grid<T> {
         }
     }
 
+    pub fn clear_viewport(&mut self, template: &T) {
+        // TODO: More efficiently using occ.
+        let mut iter = self.iter_from(Point { line: 0, col: Column(0) });
```
chrisduerr: Found the cause for the second issue!

```suggestion
        let mut iter = self.iter_from(Point { line: 0, col: self.num_cols() });
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -1934,7 +1923,25 @@ impl ansi::Handler for Term {
...
+            ansi::ClearMode::All => {
+                // Scroll the viewport into the history.
+                self.grid.clear_viewport(&template);
+                // Reset the now empty viewport.
+                self.grid.region_mut(..).each(|c| c.reset(&template));
```
chrisduerr: Instead of resetting all lines here, we can just do the following above `clear_viewport`:

```rust
for i in 0..self.lines().0 {
    self.grid[i].occ = self.cols().0;
}
```

alacritty_terminal/src/grid/mod.rs
```diff
@@ -479,10 +529,10 @@ impl<T: GridCell + Copy + Clone> Grid<T> {
...
             self.url_highlight = None;
 
-            // // This next loop swaps "fixed" lines outside of a scroll region
-            // // back into place after the rotation. The work is done in buffer-
-            // // space rather than terminal-space to avoid redundant
-            // // transformations.
+            // This next loop swaps "fixed" lines outside of a scroll region
+            // back into place after the rotation. The work is done in buffer-
+            // space rather than terminal-space to avoid redundant
+            // transformations.
```
nixpulvis: Just removing some redundant comment characters.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -72,6 +72,37 @@ pub trait GridCell {
...
+/// |                                            |
+/// +--------------------------------------------+ <--- display_offset
+/// |                                            |
+/// |              BELOW SCROLLING               |
+/// |                   REGION                   |
+/// |                                            |
+/// +--------------------------------------------+ <--- raw.zero
+///                                              |
+///                                              v
+///                                             cols
```
chrisduerr: I think if we want to add this documentation comment, we need to add a few more words to explain exactly what's going on and remove the `???` parts.

Additionally we should make sure it formats properly, which will likely require putting this in a codeblock.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -83,6 +114,7 @@ pub struct Grid<T> {
 
     /// Number of lines.
     ///
+    /// TODO: this invariant is incorrect.
```
chrisduerr: How is it incorrect? We might have talked about this previously, but this should probably be captured in an issue rather than a code comment that doesn't really explain the issue.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -508,11 +540,36 @@ impl<T: GridCell + Copy + Clone> Grid<T> {
...
+        let mut iter = self.iter_from(Point { line: 0, col: self.num_cols() });
+        let mut bottom_nonempty_line = 0;
+        while let Some(cell) = iter.next_back() {
+            if !cell.is_empty() {
+                bottom_nonempty_line = iter.cur.line;
+                break;
+            }
+
+            // In case the whole terminal is empty somehow.
+            // TODO: Test.
```
chrisduerr: Should be tested.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -508,11 +540,36 @@ impl<T: GridCell + Copy + Clone> Grid<T> {
         }
     }
 
+    pub fn clear_viewport(&mut self, template: &T) {
+        for i in 0..self.num_lines().0 {
+            self[i].occ = self.num_cols().0;
+        }
```
chrisduerr: I'm slightly concerned about this, since it will leave the `occ` at way too high values after each clear.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -543,6 +600,9 @@ impl<T> Grid<T> {
     }
 
     pub fn clear_history(&mut self) {
+        // Explicitly purge all lines from history
+        let shrinkage = self.raw.len() - self.lines.0;
+        self.raw.shrink_lines(shrinkage);
```
chrisduerr: This fixes a separate bug, right? Need to remember that when writing a change log entry.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -629,6 +689,23 @@ impl<'a, T> Iterator for GridIterator<'a, T> {
     }
 }
 
+impl<'a, T> DoubleEndedIterator for GridIterator<'a, T> {
```
chrisduerr: Why do we need the `DoubleEndedIterator` here, shouldn't the `BidirectionalIterator` work already? If this is a problem with iterating over all cells vs iterating over only empty cells, I think it would be a big mistake to have a `DoubleEndedIterator` which iterates over all, but a `BidirectionalIterator` which iterates only over non-empty cells.

alacritty_terminal/src/grid/storage.rs
```diff
@@ -22,10 +9,35 @@ use crate::index::Line;
...
+/// [`Deref`](std::ops::Deref) implementation is provided. Anything from `Vec`
+/// that should be exposed must be done so manually.
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct Storage<T> {
     inner: Vec<Row<T>>,
+
+    /// Starting point for the storage of rows
+    ///
+    /// This value represents the starting line offset within the ring buffer. Doing things this
+    /// way allows for scrolling stored lines in a performant way.
```
chrisduerr: > Doing things this way allows for scrolling stored lines in a performant way.

We don't need to reiterate these things on the documentation of a field. This kind of documentation should be on the struct or module (I think it's already sufficiently covered in the struct docs).

alacritty_terminal/src/grid/storage.rs
```diff
@@ -22,10 +9,35 @@ use crate::index::Line;
...
+    ///
+    /// The value of this offset may be larger than the `len` itself, and will wrap around to the
+    /// start to form the ring buffer.
     zero: usize,
+
+    /// An **index** separating the visible and scrollback regions
```
chrisduerr: Why is `index` bolded here? Seems like there's no need for emphasis.

alacritty_terminal/src/grid/storage.rs
```diff
@@ -22,10 +9,35 @@ use crate::index::Line;
...
+    ///
+    /// TODO: Why is this needed in storage, and not just in grid.
```
chrisduerr: I think I've explained this previously, but this is needed for growing/shrinking the available scrollback lines.

alacritty_terminal/src/grid/storage.rs
```diff
@@ -325,6 +337,75 @@ mod test {
...
+
+    #[test]
+    #[should_panic]
+    fn rotate_overflow() {
+        let mut storage = Storage::with_capacity(Line(3),
+                                                 Row::new(Column(0), &' '));
+        storage.rotate(4);
+    }
+
+
```
chrisduerr: Formatting in these tests seems messed up? And there's an extra newline here.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1950,7 +1939,22 @@ impl ansi::Handler for Term {
...
+                        .region_mut((self.cursor.point.line + 1)..)
+                        .each(|cell| cell.reset(&template));
+                }
+            },
+            ansi::ClearMode::All => {
+                self.grid.clear_viewport(&template);
+            },
+            ansi::ClearMode::Saved => {
+                self.grid.clear_history();
+            }
```
chrisduerr: These calls should be formatted to single line. Unfortunately rustfmt can't do this automatically so it needs to be done manually. Rustfmt can only break them when they are too long.

alacritty_terminal/src/grid/mod.rs
nixpulvis: Calling `prev` does indeed work. I forget why I felt the need to write this separate iterator. Since I'm checking for `!cell.empty()` inside the `while let Some(cell) = iter.prev()` I don't think it was a functional reason like that.

alacritty_terminal/src/grid/storage.rs
nixpulvis: Yes you have, sorry. I think I simply forgot to remove this `TODO`.

alacritty_terminal/src/grid/mod.rs
nixpulvis: Would it make sense to reset them back to 0 at the end of `clear_viewport`? As I'm understanding things right now, setting `occ = self.num_cols()` is telling the grid to consider each row as fully occupied for the purposes of clearing the viewport. And after a clear it should be completely empty.

I've tested this, and it seems to work.

chrisduerr: Actually I'm not sure if that will do anything. I have to investigate again what exactly is going on, but in theory if we set the occ to num_cols before scrolling, the stuff with incorrect occ should be what has been scrolled offscreen.

So I'm a bit confused about what exactly is happening. First thing we do is reset the occ, then we rotate, meaning that these lines are gone and then we reset the fixed lines (which I think should be what has been 'rotated in' from the bottom). So I don't quite get why the reftest failed exactly and I'd have to look more into how the occ affects things and changes over time to give a definite answer.

I'd recommend checking what the `occ` is before you reset them to 0, I'd expect them to be 0 already.

nixpulvis:
Interesting, I think I've found a clippy/rust bug. Clippy is telling me `error: statement with no effect`, yet commenting it out fails the test.

chrisduerr:
> Interesting, I think I've found a clippy/rust bug. Clippy is telling me error: statement with no effect, yet commenting it out fails the test.

Hah, you're really giving clippy a hard time here though. Usually these statements shouldn't do anything unless you expect your code to fail, which makes little sense of course.

But to fix this in the tests, I think you can just do `let _ = &storage[1];` instead, to explicitly tell it that you don't care about the return type. That's also a decent idea in general just to siginal things better to future readers of these tests (though they're so simple it doesn't really matter).

alacritty_terminal/src/grid/mod.rs
nixpulvis: @chrisduerr I don't think I was trying to fix a bug with this so much as grouping it so I could reuse it more easily.

nixpulvis:
I've rebased this on master, and I'm interested in wrapping this up soon. As far as I'm concerned the only remaining tasks are:

1. Figure out what's going on with `occ` exactly
2. Refine the added documentation

chrisduerr:
I think to figure out the `occ` issue, it helps to show a bit how clearing the terminal works.

When running `printf "\e[41m"` in bash, that should set the template cell to red for example. Now you'll notice that whenever you write something and delete the last character, it will paint the entire line red for example. This is because bash will send a Clear::Right at that point.

If we look at what happens when clearing, we'll notice that now all the cells are bright red in the visible region. That's the reason why the reftest fails when you set the new lines to zero, because their state depends on `template.is_empty()`. Only if the template is empty, then all the new added lines will have an occ of `0`.

Now if we're looking at the behavior of this PR, we'll notice that our clearing routine is actually incorrect, since we end up with a completely clear screen, however we should reset to the template cell which should leave us with a completely red screen.

A very interesting thing when doing `printf "\e[41m"` is also when you use `ls -lah` until there's rotation going on. You'll notice that the end of the line is empty (default bg) until you hit new lines that weren't filled before. At that point the lines start being completely filled until the end. This is the correct behavior, but something that might come as a surprise. So maybe that helps a bit with understanding too.

Initially we changed the `occ` to fix the failing ref tests, however this did not correctly represent what we actually want to do. I think the following change to the row's `reset` function shows more clearly what we actually would like to do:

```rust
if self.occ < self.inner.len() && self.inner.last() == Some(template) {
    // Previously this is all we did, just reset the cells that are empty
    for item in &mut self.inner[..self.occ] {
        *item = *template;
    }
} else {
    // If 'template' isn't even empty, we need to reset everything though on clear
    for item in &mut self.inner[..] {
        *item = *template;
    }
}
```

I'm not sure if it's going to be possible to change the `reset` method to this though, both because of performance and because it's used in a lot of different places too. But I think this should appropriately represent why the initial version of the PR did not work.

Another fix, which should be more viable and less risky is the following change to the `scroll_up` function:

```
for i in 0..*positions {
    if template.is_empty() {
        self.raw[i + fixed_lines].reset(&template);
    } else {
        self.raw[i + fixed_lines] = Row::new(self.num_cols(), &template);
    }
}
```

Hopefully this should help with understanding the issue and how to best fix it. I can take another look to determine what the best possible way to resolve this would be and compare the different versions, benchmark them and see if there are any bugs with any of them. However I'd first like to know if you understand the problem @nixpulvis and if you're interested to explore a few different options yourself.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -501,11 +527,41 @@ impl<T: GridCell + Copy + Clone> Grid<T> {
...
+        let positions = self.lines - bottom_nonempty_line;
+        let region = Line(0)..self.num_lines();
+
+        // Clear the viewport.
+        self.scroll_up(&region, positions, template);
+
+        // We've now cleared each row, so they are empty.
+        for i in 0..self.num_lines().0 {
+            self[i].occ = 0;
+        }
```
chrisduerr: All this occ code can be deleted now.

nixpulvis:
This PR on the left, and master on the right. Clearly something is still amiss here. :crying_cat_face:

chrisduerr:
@nixpulvis See my comment on IRC:

> Wrote the tests in https://github.com/jwilm/alacritty/pull/2964 and ran them against the latest version of the PR and immediately found a bug.

> Since we rotate only the minimum amount, all other cells are not reset (say we only rotate 3/30 lines, 27 will be unchanged). So we need to reset the rest ourselves (without rotating) even if we do not rotate a full screen.

> And that needs to be done always, regardless of template.is_empty or not. So just calling `reset(template)` on every row again that hasn't rotated in.

I didn't think about this either, but in retrospect it makes a lot of sense.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -73,6 +73,34 @@ pub trait GridCell {
...
+/// |                                            |
+/// |                                            |
+/// |                                            |
+/// |                                            |
+/// |               SCROLLUP REGION              |
+/// |                                            |
+/// |                                            |
+/// |                                            |
+/// |                                            |
+/// +--------------------------------------------+ <--- self.lines
```
chrisduerr: This is also a bit misleading, since `self.lines` is not the distance away from `zero`, but rather the distance from `display_offset`. It's of course difficult to show that in a graphic like this, but maybe something like the following could work out:

```
┌────────────────────────┐ <── max_scroll_limit + display_offset
│                        │
│      UNINITIALIZED     │
│                        │
├────────────────────────┤ <── scroll_limit + display_offset | raw.len
│                        │
│          ABOVE         │
│                        │
├────────────────────────┤ v lines
│                        │ │
│     VISIBLE REGION     │ │
│                        │ │
├────────────────────────┤ ^ <── display_offset
│                        │
│          BELOW         │
│                        │
└────────────────────────┘ <── zero
                         ^
                         |
                        cols
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -668,6 +668,25 @@ impl VisualBell {
...
+/// |              SCROLLING REGION              |
+/// |                                            |
+/// +--------------------------------------------+ <--- scroll_region.end
+/// |                                            |
+/// |                                            |
+/// |          (See Grid's Documentation)        |
+/// |                                            |
+/// |                                            |
+/// +--------------------------------------------+
+/// ```
```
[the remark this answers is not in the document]
nixpulvis: Yea, I agree that this doesn't make the most sense yet. It evolved a bit as I was understanding things myself. I think I agree we can make this just some text like you mention.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -73,6 +73,32 @@ pub trait GridCell {
...
+/// │                         │
+/// ├─────────────────────────┤  <-- raw.len()
+/// │                         │
+/// │      RESIZE BUFFER      │
+/// │                         │
+/// ├─────────────────────────┤  <-- scroll_limit + lines
+/// │                         │
+/// │     SCROLLUP REGION     │
+/// │                         │
+/// ├─────────────────────────┤v lines
```
nixpulvis:
```suggestion
/// ├─────────────────────────┤v <-- lines
```

chrisduerr: I've explicitly omitted the arrows there, since `lines` is not at that point, but it's the range of visible lines. You might want to compile the docs to see it better.

nixpulvis:
This LGTM too. It's worth noting that at this time your version of `clear` may still be outdated. Refer to https://github.com/jwilm/alacritty/pull/2329#issuecomment-510533778 for an easy install `PKGFILE` from source.

aneutron:
Thanks a lot lads !

PR #2331, opened 2019-04-19
Call on_resize() if font changes

nicolashahn:
Fixes https://github.com/jwilm/alacritty/issues/2327

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -415,7 +415,7 @@ impl Display {
                 pty_size.height -= pty_size.cell_height * message.text(&size).len() as f32;
             }
 
-            if previous_cols != size.cols() || previous_lines != size.lines()  {
+            if font_changed || previous_cols != size.cols() || previous_lines != size.lines() {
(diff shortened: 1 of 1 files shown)


src/display.rs
```diff
@@ -415,7 +415,7 @@ impl Display {
                 pty_size.height -= pty_size.cell_height * message.text(&size).len() as f32;
             }
 
-            if previous_cols != size.cols() || previous_lines != size.lines()  {
+            if font_changed || previous_cols != size.cols() || previous_lines != size.lines() {
```
chrisduerr: While this solves the problem, I think there would be a solution that makes the problem a little more obvious.

We only need to check if cols or lines have changed, that's perfectly fine since the shell doesn't care about font size, however we previously update the size info in the `if font_changed || self.last_message != terminal.message_buffer_mut().message() { ... }` block.

So as a solution, I'd propose just moving the definition of `previous_cols` and `previos_lines` at the very beginning of the method, that should fix this bug too.

PR #2336, opened 2019-04-19
Update glutin

atouchet:
This was updated in Cargo.lock in #2328 but not in Cargo.toml.


chrisduerr:
Since this is a non-breaking change there is no need to bump the version in the Cargo.toml and the automatic update using the Cargo.lock file is entirely sufficient.

PR #2339, opened 2019-04-20
Fix cursor colors

chrisduerr:
This fixes a recent regression in
cfc20d4f34dca535654cc32df18e785296af4cc5 which broke cursor colors when
specified in the `colors.cursor` field in the config.

It also removes a lot of unneeded code from the font crate related to
the cursor rendering.


kchibisov:
Cursor's colors are working fine for me in this PR.

Issue #1554, opened 2018-09-18
Select with Shift + Click


elbaro:
example:

select a few characters, scrollback with pageUp or mouse wheel, and shift+click to extend the selection.

chrisduerr:
Thanks for the suggestion.

There might be some issues with this though since `shift` is already used to disable mouse mode for selection.

One scenario where the behavior might be unexpected, would be this:
 * Enable mouse mode
 * Make selection with shift+drag
 * cancel selection with left click+shift
 * keep holding shift to start a new selection

In this scenario it wouldn't be certain if a new selection should be started, from the point where the selection was cancelled, to the new start, or if the first click was just to cancel the selection.
I think there should be a way to resolve this, in this case it's probably best to just assume the first click was just to cancel the selection, but this becomes more complicated when there is no selection currently visible (but still present in the history).

Do you have any examples for terminal emulators which already do this? It might be nice to capture their behavior to make sure if Alacritty should receive this feature, that the behavior would be consistent and battle-tested.

elbaro:
gnome-terminal and OSX default Terminal app do this.
This is necessary when I have a huge output (>1000 lines) to select and copy.

kvelicka:
`xterm` does the same by left clicking somewhere and right clicking somewhere else (everything in between gets selected).

I think that the general idea of "mark start/end and select in between" is the interesting suggestion, the keys/shortcuts to achieve that are just a minor detail.

kchibisov:
We decided to go for right mouse button for selection expansion, since `Shift` is a bit overloaded with other implicit things already, and they were conflicting with each other.

adsick:
> We decided to go for right mouse button for selection expansion, since `Shift` is a bit overloaded with other implicit things already, and they were conflicting with each other.

Does `Shift + Mouse Click` do something else (you mentioned "conflicting")? I was super surprised to find out about `Click + Right Click`, I don't know any other software that works like this. Right click usually opens a context menu (Copy, Paste, etc.)

PR #2361, opened 2019-04-24
Modify Text Selection

Ben496:
Allow the user to modify a selection. This can be triggered with either a key combination or modifier + mouse button. The end point of the selection will be moved to the mouse location.

The selection will still be cleared when clicking the mouse without any modifiers activated.

Fixes: #1554


chrisduerr:
As far as I know all terminal emulators that support it just map it to shift. Is there any reason why we would make this configurable over just making it default behavior when shift is held?

Ben496:
Xterm allows the user to modify a text selection by using right click. I think, since most applications follow the same behavior for shift+left click, that either setting the default to be shift+left click or removing the configuration entirely and only using shift+left click would work.

chrisduerr:
Interested in @nixpulvis opinion on this. I think I'd personally be for using just shift without configuration option. Configuration is great, however only when it's being used. Since all applications I'm aware of use shift for this already, I don't think a config option is necessary.

nixpulvis:
I would naturally expect <kbd>shift</kbd> + left click to work. So I think at a minimum we should support that out of the box.

Based on my exceptions of how this should work from applications like Firefox:

- <kbd>shift</kbd> should be the only modifier for this
- The selection should be able to grow **and shrink**
- Clicking, not just dragging should trigger this

I'm assuming we don't really want to follow XTerm on this one since, as far as I have seen, it's the only program using right click. It's also a weird behavior, since it remembers the last place you clicked and always starts from there, regardless of if there's a selection or not.

I don't feel like we **need** to add an action, personally, but maybe I'm missing a use case?

nixpulvis:
Interestingly Firefox also remembers my last click and starts the selection from there when holding <kbd>shift</kbd> and clicking. So maybe this is a more generally accepted "feature" than I expected. In fact I kind of like it, mark and select.

Of course, Firefox doesn't do this on right click.

Ben496:
Alright, I will remove the configuration option and bind it to <kbd>Shift</kbd>, if this becomes an issue an action can always be added later. The only use case that I can think of that would conflict with using <kbd>Shift</kbd> would be if someone already bound one of the other actions to that in their config.

Right now only the endpoint of the selection is moved when modifying the selection. Do we want to support moving both ends of the selection, or just move the endpoint of the selection? Firefox and most other gui applications seem follow the current behavior of just moving the endpoint of the selection with <kbd>Shift+Left Click</kbd>. On the other hand Termite, URXVT, and XTerm are some applications that allow both ends of the selection to be moved.

nixpulvis:
@Ben496 I'd personally prefer to follow a "mark and select/modify" functionality, so I'd expect the selection endpoint to be all that's updated (like Firefox). I'm somewhat indifferent if a <kbd>shift</kbd> + left click should start a selection from the last mouse click to the current mouse position or not however. We should make a decision about that explicitly.

Ben496:
I do not have a strong preference as to whether <kbd>Shift</kbd> + left click without a visible selection should be allowed. I will personally always initiate a short drag so that I can see that I have the proper start point.

chrisduerr:
I think we need to move the end point and keep the start point static for a consistent deterministic behavior. I'm not sure how URxvt/XTerm handle it but if it's possible to modify both start and end I'd assume that it will move whatever is closer, so that might be super confusing especially if you're trying to click somewhere in the middle of the current selection to shrink it.

Ben496:
Commit af403b6 fixes <kbd>Shift+LMB</kbd> not working when Alacritty is initially launched.

This also changes the behavior when using <kbd>Shift</kbd> to temporarily disable mouse mode. The new behavior is to require the selection to be explicitly cleared before a new one can be created, but I think that this could still be problematic because it could cause user confusion. Some potential options for dealing with this would be:
1. Disable modifying the selection when shift is being used to disable mouse mode.
2. Require the selection to be cleared before LMB can be passed through to the application
3. A key other than shift could be used to trigger modifying the selection.

I have noticed that gnome-terminal appears to use option 1 and only allow the selection to be modified when shift is not being used to disable mouse mode. I haven't checked other terminal emulators yet, but this is the solution that I am in favor of.

chrisduerr:
> A key other than shift could be used to trigger modifying the selection

I don't think we have that option. Shift is the only one that users would easily find through exploration by getting used to it from other terminals and I believe ctrl is already reserved for block selection. So we should try to make things work with shift.

> Require the selection to be cleared before LMB can be passed through to the application

I don't think I hate this idea. But if we think about things like selections possibly being outside of the visible region, I think it could lead to a ton of confusion.

> Disable modifying the selection when shift is being used to disable mouse mode

The inconsistency of this is pretty bad I think, so I'm not sure how much I like this option. It seems to be the best solution for a good chunk of users, but others that spend a lot of time with mouse mode enabled in programs like IRC (which is probably where this would be the most useful), will gain nothing which would be unfortunate.

Unless  you have any other proposals, I don't think it's possible to solve these issues though, so I feel inclined to just go with the option gnome-terminal goes for (they've probably also thought about this a lot and came to the same conclusion).

Is there anything that still needs to be addressed in this PR @Ben496? It seems to me like it should be identical to gnome-terminal's behavior in the current version so we could go ahead with merging it after a bit more testing and a full code review.

nixpulvis:
I'll give this another more thorough testing and read over later, but just glancing quickly at it again, I notice that the selection mode is left untouched. I'm not sure this is what we want. If I selected some text with a double click (for example), thus triggering semantic selection, I _think_ I'd expect a modification later to be a simple selection, allowing me to select just a part of a semantic block.

Hopefully that made sense.

Ben496:
@nixpulvis I think I understand what you are asking. For example, be able to switch from line selection (triple click) to simple selection without clearing the selection. For me, if I start with line selection, I would expect to be able to extend as a line selection, instead of a simple selection. Looking at gnome-terminal as a reference again, it doesn't touch the selection method when modifying with shift. Do you have an example of a terminal emulator that changes the selection method?

@chrisduerr In other terminals that use mouse mode when running an application such as htop the selection is always cleared when the terminal redraws under the selection while Alacritty's existing behavior is to keep the selection in the same spot on the screen (even when scrolling), even when scrolling through the process list. I haven't looked at irc clients specifically yet to see if they behave any different.

As for stuff that still needs to be addressed, I have noticed some unexpected behavior with my latest commit when scrolling with the mouse wheel in neovim. Currently, if I make a selection and then scroll down so that the selection is off the screen and then <kbd>Shift+LMB</kbd> the selection appears to extend, but in reality anything above the top of the terminal is lost. On the other hand when scrolling up the selection stays in the same location. This behavior also exists in the existing release of Alacritty too (if you just make a selection and then scroll). Hopefully that makes some sense.

nixpulvis:
@Ben496 very interesting that the other terminals implementing this seem to generally keep the same mode after the initial selection while modifying with a shift click. This isn't consistent with the expectations I have from programs like Firefox or Chrome (for example).

I think the **best** implementation would allow the user to double and triple shift click to convert the selection to a new kind of selection. I'm not sure it's worth the complexity, since it may be necessary to complicate the  selection to allow a `Simple` start, but a `Semantic` end.

Barring that solution, I would personally rather have all shift click selection modifications to be simple personally. You've already selected the text you want, now it's refinement. I don't personally see following other terminals here to be required, since I would think most user expectations would be set elsewhere.

chrisduerr:
I've decided to take a look at some other terminal emulators and I think we should aim to align ourselves with them.

XTerm and Kitty are two terminal emulators that do not seem to support this. I'd recommend to just ignore them.

Termite and URxvt both support it. Both also allow modifying both directions of the selection, so I think it would probably be a good idea to follow that. This means that when clicking to the left of an existing selection, the start should be extended, when clicking to the right the end should be extended and when clicking inside, the closer side should be moved. I'd recommend testing yourself though since I might have missed something.

Both termite and URxvt do not allow extension when mouse mode is enabled, we should keep what we're doing right now and follow that.

The big difference between Termite and URxvt is the behavior when extending after semantic/line selection. URxvt seems to shift-expand normally when it's a normal selection, however both line and semantic selection are expanded with semantic selection. Termite expands every selection style with the style of the current selection. I think Termite's behavior is more consistent here and should be followed.

So in general, I'd recommend copying Termite (or any other VTE terminal) 100%. I think this is the best option forward and the only thing that makes sense (I wouldn't want to be the odd one out here).

chrisduerr:
Since @nixpulvis and I have both commented at the same time, I've shortly discussed this on IRC and come to the conclusion that while in every other area we should follow Termite, when extending special selection modes it makes sense to act similarly to initial selection.

So if the user holds shift and clicks normally, it should be extended as a normal selection. If a user double-clicks it should be extended as semantic selection and a triple-click should be line selection. I think that makes a lot of sense and is something where we can actually provide something *better* than VTE.

The more input on this, the merrier though, so I'd love to hear what you think yourself @Ben496. Hopefully this hasn't just caused a ton of confusion, so if you have any questions please let me know before heading off in the wrong direction.

Ben496:
I like the idea of the user being able to choose which selection method to use when extending the selection with shift. This gives good base behavior of extending via simple selection (like Firefox), but also allows the other methods. My understanding is to allow the user to modify the selection to both finer and coarser selection types, ex: start the selection with semantic selection and then can extended as any of the three selection types depending on if the user single, double, or triple clicks.

chrisduerr:
> start the selection with semantic selection and then can extended as any of the three selection types depending on if the user single, double, or triple clicks.

I just wanted to make it clear that this is exactly what I mean. Instead of the current behavior, no matter which way a selection was started, it should always be possible to extend it as a normal selection, a semantic selection or a line selection.

Ben496:
I have implemented heterogeneous selection modes. This allows the user to create a selection in one mode and update it with another mode. This allows for the ability to create a selection that, for example, has a `simple` start and a `semantic` end. The user chooses which selection mode to update by single, double, or triple <kbd>LMB</kbd> when updating the selection with <kbd>Shift</kbd>.

nixpulvis:
Just taking a quick look now, and it seems to be working really great :tada: I'm really looking forward to this.

Just one questions about functionality... I'm not sure I like deciding which part of the selection to modify by a midpoint in the selection, I'm much rather always update the second selection anchor. This is (again) how Firefox text selection works.

Ben496:
Yeah, looking back it looks like I should always update the endpoint unless the start point is between the clicked location and the end point of the selection instead of using a midpoint method to align with the way that VTE handles this.

nixpulvis:
I think I've found a bug:

screen

Ben496:
@nixpulvis Just looking for a confirmation before working on a solution.

It looks to me like the expected behavior for the first selection you make would be to have the start point of the selected cells stay at the left side of `nixpulvis` when you move the mouse to the left. This would end up with `xr-x 3 ` highlighted, but not `nixpulvis` when expanding to 7 characters left of the n in nixpulvis. In other words the end the span that is not being updated should not move no matter how the selection is updated.

As for your last selection, where you start with simple selection and expand as a semantic selection. I do not think this is a bug because updating a selection should not affect the stationary point on the span.

Ben496:
I have also noticed a bug when reversing a multi-line selection by clicking on the same line as the start of the selection.

nixpulvis:
I think the first issue can be resolved by treating a semantic selsction's start as the leftmost point that it matches. It's weird that a selection "remembers" where you double clicked the initial semantic selection. It should be irrelevant by the time you come to edit the selection.

As for the next point, I'm thinking in terms of invariants about the selected text. A semantic selection should never start in the middle of a semantic match, since before this change there was no way to achieve this, and it's a nice property. So I would argue that a simple selection that is promoted to a semantic selection should update it's start to cover the whole word.

This would mean shift double click dragging (semantic selection modification) the selection from either side of the start would leave the word semantic match completely selected throughout.

Ben496:
>So I would argue that a simple selection that is promoted to a semantic selection should update it's start to cover the whole word

@nixpulvis I would argue the opposite. If I start a selection with a simple selection I want to select part of a word. Then if I choose to extend the selection as a semantic selection, I would not want my initial point to change. I would instead want to keep the initial selection point and only change the end. This would end up with a selection with the start as simple and the end as semantic. If I converted the entire selection to semantic then both ends would be semantic and I would loose the fine selection that I started with. By extension I would expect the same thing with simple to line and semantic to line.

I am interested in also hearing what @chrisduerr thinks about this.

chrisduerr:
I personally wouldn't expect the other side of the selection to change at all. So extending by semantic at the left side, shouldn't change line/semantic/simple at the right if possible.

alacritty_terminal/src/selection.rs
```diff
@@ -18,52 +18,53 @@
...
 use crate::term::{Search, Term};
 
-/// Describes a region of a 2-dimensional area
+/// Describes a type of a 2-dimensional area
 ///
-/// Used to track a text selection. There are three supported modes, each with its own constructor:
+/// Used to track a text selection type. There are three supported modes, each with its own constructor:
 /// [`simple`], [`semantic`], and [`lines`]. The [`simple`] mode precisely tracks which cells are
 /// selected without any expansion. [`semantic`] mode expands the initial selection to the nearest
 /// semantic escape char in either direction. [`lines`] will always select entire lines.
```
chrisduerr: > There are three supported modes, each with its own constructor

This comment part doesn't fit the enum. There should also not be any duplication between the different documentations, since that would mean both have to be edited on a change and people are likely to miss that.

alacritty_terminal/src/selection.rs
```diff
@@ -81,200 +82,92 @@ pub trait Dimensions {
...
-            start.line = min(start.line, region.start.line);
-            end.line = max(end.line, region.end.line);
-        } else {
-            // Start is above end
-            start.line = min(start.line, region.end.line);
-            end.line = max(end.line, region.start.line);
-        }
-
+                (start_orig, end_orig, start_side, end_side, self.start_type.clone(), self.end_type.clone())
+            };
```
chrisduerr: Why separate start side/point here? Seems like it just creates additional complexity.

And if we have selection type tied to start/end tied anyways, we might as well move that into `Anchor` (and probably rename it).

That way we go from 6 variable definitions down to 2.

alacritty_terminal/src/selection.rs
```diff
@@ -81,200 +82,92 @@ pub trait Dimensions {
...
-        };
-
-        if alt_screen {
-            Selection::alt_screen_clamp(&mut start, &mut end, lines, cols)?;
-        }
-
-        let (mut start, mut end) = if start == end {
-            if let Some(end) = grid.bracket_search(start.into()) {
-                (start.into(), end)
+    pub fn to_span(&self, term: &Term, alt_screen: bool) -> Option<Span> {
```
chrisduerr: This function is massive now, it should be split into smaller more compact parts.

alacritty_terminal/src/selection.rs
```diff
@@ -283,18 +176,55 @@ impl Selection {
...
+            span
+        })
+    }
+
+    pub fn is_empty(&self) -> bool {
+        if self.start_type == SelectionType::Simple && self.end_type == SelectionType::Simple {
+            self.region.start == self.region.end
+        } else {
+            false
+        }
```
chrisduerr: This body can be changed to `self.start_type == SelectionType::Simple && self.end_type == SelectionType::Simple && self.region.start == self.region.end`.

Ben496:
@chrisduerr @nixpulvis We should reach a consensus on how to shrink the selection since each of you are thinking a different method. Do we want to always shrink the end, or use a midpoint method to shrink the closest side? Right now the I have implemented the way VTE (termite) seems to handle this, so the end (the last updated side) is always the side that shrinks.

chrisduerr:
If the last updated side is always changed, that would make it impossible to edit one of the sides, right? Seems like that just brings in restrictions.

Ben496:
It would be possible to edit both sides still. The way you can achieve this in termite and my current code (commit: 67186d6) is to click outside the selection on the side that you started on, then you can drag to shrink the selection.

I personally would also prefer using a midpoint to determine which side to shrink, but that deviates away from the behavior of the VTE terminals that I've looked at.

chrisduerr:
I think URxvt does closest side. I'd honestly go for that. Nobody would ever be able to figure out that you need to grow it first to be able to shrink it, that's just way too complicated.

@nixpulvis Any opposing thoughts?

alacritty_terminal/src/event.rs
```diff
@@ -80,11 +80,27 @@ impl<'a, N: Notify + 'a> input::ActionContext for ActionContext<'a, N> {
...
     }
 
+    fn selection_bounds(&self) -> Option<(&Point<isize>, &Point<isize>)> {
+        if let Some(ref selection) = self.terminal.selection() {
+            let start = &selection.region.start.point;
+            let end = &selection.region.end.point;
+            Some((start, end))
+        } else {
+            None
+        }
```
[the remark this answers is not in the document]
Ben496: I also think it makes sense for this be neither a function nor an event, as this code is only used once.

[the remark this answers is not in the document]
chrisduerr: If it's only used once, that probably makes sense, yeah.

nixpulvis:
I liker how Firefox for example does this, it's easily predictable to just be updating the endpoint of the selection. It should be very easy to determine which side it will shrink from, it should always be shrinking towards your initial selection start marker.

chrisduerr:
I'd be fine with the way Firefox does it. It is by far the clearest when it comes to how the selection shrinks/grows, though it is more limited than what URxvt provides.

chrisduerr:
@Ben496 Me and @nixpulvis have discussed this in a lot of detail on IRC and came to the conclusion, that a fixed starting point is likely our best bet. So for the simple selection, it should function identical to the way it works in Firefox (as a comparison).

Semantic and line selection are a bit more complicated. However they shouldn't be too tough to handle. Expansion should always be in the opposite direction of the starting point from the position of the click.

So even if you semantic expand inside the selection, if that would semantically expand to a point outside of the current selection it will grow. Semantic shrinking is possible only if the semantic selection expands to a point that is smaller than the current selection.

With line selection the behavior is identical. So triple-clicking the last line (no matter if inside or outside the selection) will always grow it to the end of the line. Clicking inside a multi-line selection with triple-click will shrink it to the end of the line.

The goal of this approach is to be simple to implement, easy to understand for the user in all scenarios and to be consistent. If you have any questions or you'd like to share your own feedback, please let me know @Ben496. This definitely wasn't an obvious decision for nixpulvis and me, so if you have potentially other insights that would be much appreciated. However if you just want to get this done (which I'd understand very much), feel free to just go ahead with the suggested implementation.

chrisduerr:
Just in case https://github.com/jwilm/alacritty/pull/2552 lands before this, I think it would make sense there to just expand the endpoint, but keep the selection as a block selection. So the end moves like a semantic, line selection, but the beginning stays the same and thus even with a triple-click not the whole line would be selected if the block selection start doesn't include the first cell.

Ben496:
While I think that it would be nice to be able to update both ends of the selection, I like that it does keep things simple to only update the end of the selection. It is definitely more limited than URxvt, or Termite, but I personally will not miss that functionality either.

chrisduerr:
@Ben496 Once this PR is in a shape where all recommendations have been implemented, please ping me so I can review it. That way I don't need to watch for changes and you can take your time. Just to make sure you're not waiting for a review without me realizing.

Ben496:
@chrisduerr I am ready for you to do another review of my code.

Ben496:
@chrisduerr Thanks for the input!

> So when expanding semantically on the middle of `README`, it jump to the end of the word and then immediately jump back to only `READM`.

When I went to reproduce this I lost only the first letter of `README`, instead of the last letter, so I ended up with `EADME` selected.

As for the block selection. The current implementation of block selection in master completely ignores semantic selection so I had just kept it the same because it seemed like I was adding another feature to this pull request. I should've just asked you about that right away instead of guessing. I think I see now what you meant in your previous comment about block selection. Given this example:

```
The quick brown
fox jumped
over the lazy dog
```

If you start the semantic block selection in the word `quick` you should be able to make a semantic block selection that is contained within the `|` characters when ending the selection in the word `lazy`:

```
The |quick bro|wn
fox |jumped   |
over| the lazy| dog
```

And then of course have block selection support all the different combinations that normal selection supports.

chrisduerr:
> When I went to reproduce this I lost only the first letter of README, instead of the last letter, so I ended up with EADME selected.

That's probably because you expanded to the left while I expanded to the right (I started the selection left to right).

> As for the block selection. The current implementation of block selection in master completely ignores semantic selection so I had just kept it the same because it seemed like I was adding another feature to this pull request.

I wouldn't expect semantic/line selection to work for the initial block selection. However, I thought this would be a good idea for the expansion only, so the expansions are all consistent and the user doesn't get results he's surprised about.

I'm not sure how initial block selection should behave with semantic/line, I think it doesn't quite fit it which is why I've left it out. Though for expansion I do think it makes sense.

Ben496:
> That's probably because you expanded to the left while I expanded to the right (I started the selection left to right).

That's what I was doing differently.

> I'm not sure how initial block selection should behave with semantic/line

I think the way this would work is if the user double or triple clicks while pressing <kbd>ctrl</kbd> then a block selection should be created with the ends marked as semantic or line respectively. Then the user can expand as the same type that the block selection was started just by dragging the mouse after creating the selection. This might be useful if trying to select specific columns in a table, as a potential example.

> I think it doesn't quite fit it which is why I've left it out. Though for expansion I do think it makes sense.

I think that this fits if we want to allow expanding as semantic or line. While having a block selection with both ends being line selection wouldn't really be different than normal line selection, it would be more consistent with normal selection, and would also allow changing the ending to say semantic or simple. I think it also makes more sense than currently because right now double clicking with <kbd>ctrl</kbd> pressed will create a normal selection instead of a block selection.

I do not think that it makes sense to allow the user to convert between a normal selection and a block selection.

Ben496:
@chrisduerr I fixed the incorrect semantic & block expansions as well as implemented expanding block selection as semantic & line.

I'm not sure if we wanted to allow the user to create a block selection as semantic or line so I didn't touch that functionality at all.

nixpulvis:
@Ben496 I was testing this out, and the functionality is seeming really good. However the responsiveness of the selection updates has gotten laggy. I'm worried about the performance of this ATM.

Can you reproduce this by moving the mouse around quickly while selecting?

chrisduerr:
@nixpulvis Did you get that result in release mode too? I've had the same impression when running that in debug mode, but thought that might just have been due to lacking optimizations. If that's always present that would be troublesome.

Ben496:
@nixpulvis @chrisduerr I tested a release build and noticed that there does appear to be a performance issue, it appears to be more noticeable when moving the mouse diagonally.

chrisduerr:
@Ben496 You think you could give this a test within `perf` or similar performance tracking tools? With a bit of luck, the relevant piece of code might be pretty obvious. From a quick glance at the code I wasn't able to see anything stand out a lot, though I didn't have the time to look that closely unfortunately. We don't seem to be doing more on redraw which is good, however Alacritty can generate **a lot** of mouse motion events.

Ben496:
Yeah, I'll run a performance profile, once I get a chance, and see what that turns up. I did, however, run a git bisect to try and find the first bad commit, which ended up being on one of the commits I merged in from master (84aca672964e29b5b4503b7da7bc34fc395f08ab). I haven't looked past that yet other than to note that performance appears to be back to normal after reverting that commit. The performance profile should give more useful information though.

chrisduerr:
@Ben496 Huh, yeah that could be a performance issue, but for that reason I've asked them to add a reference to the current cursor in the window so it shouldn't actually call any winit code unless the cursor actually changes.

Can you reproduce this on the latest master too? If so, you shouldn't have to worry about it obviously. I can look into it then in a few days and we can just create a separate issue.

kchibisov:
@Ben496 Would you mind testing patch from #2666, it should improve some things?

Ben496:
@chrisduerr I cannot reproduce the issue in master, only in this branch.

@kchibisov Merging #2666 into this branch appears to fix the performance issue for me.

Ben496:
Err, on second thought, if I apply the change in #2666 manually it still occurs.

Ben496:
Ok, I brought this branch up to date with master and that fixed the performance issues that I was seeing with moving the mouse quickly while dragging a selection. @nixpulvis does it work better for you now, or are you still seeing lag while dragging the mouse?

kchibisov:
@Ben496 I didn't notice that your branch was a bit behind. I think I've fixed performance issue in a99547cc6dc1d752b29de785f969591c1ed74782 , since `mouse_moved` was just incorrect.

alacritty_terminal/src/input.rs
```diff
@@ -575,6 +578,21 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         }
     }
 
+    fn is_expand_selection(
```
nixpulvis: Tiny nitpick, feel free to ignore.

Would this make more sense as `should_expand_selection`?

alacritty_terminal/src/selection.rs
```diff
@@ -40,35 +40,39 @@ use crate::term::{Search, Term};
...
-        region: Range<Point<isize>>,
-    },
 }
 
 /// A Point and side within that point.
 #[derive(Debug, Clone, PartialEq)]
 pub struct Anchor {
-    point: Point<isize>,
+    pub point: Point<isize>,
+    pub ty: SelectionType,
```
nixpulvis: I'm sorry if I'm missing something, but why do the anchors need to store the `SelectionType`? As I understand it, a selection is defined by a starting point and an ending point. Editing a selection only depends on the starting point, and depending on the mouse clicks will update the end point according to one of the three selections types.

Storing the `SelectionType` could be useful if we were trying to preserve the type while modifying the selection, but this isn't the case. I don't think this is a functionality issue, but more of a data structures concern/question.

Seems related to https://github.com/jwilm/alacritty/pull/2361/files#r294029075

Thoughts?

Ben496: > why do the anchors need to store the `SelectionType`?

I currently have `SelectionType` stored in the anchors because `to_span` needs to know how to expand the visual selection. I moved it into the `Anchor` because it explains how the `Anchor` should be expanded and that makes more sense stored in the `Anchor` than in `Selection`.

> Editing a selection only depends on the starting point, and depending on the mouse clicks will update the end point according to one of the three selections types.

There is one situation where editing (updating) a selection will have an effect on both ends of the selection. This is where the user expands the selection in the same mouse click that created a semantic or line expansion. In this case if the user changes the direction that they drag the selection then the semantic/line selection start point needs to expand in the opposite direction than what was initially done.

On another note, perhaps a better name for the enum would be `ExpansionType` instead of `SelectionType`.

alacritty_terminal/src/input.rs
Ben496: I do like `should_expand_selection` better.

alacritty_terminal/src/input.rs
```diff
@@ -575,6 +578,21 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+        button: MouseButton,
+        modifiers: ModifiersState,
+        report_mode_intersection: bool,
+    ) -> bool {
+        // We do not want to expand the selection when using mouse mode because
+        // shift conflicts with disabling the mouse when using mouse mode.
+        modifiers.shift
+            && button == MouseButton::Left
+            && !self.ctx.selection_is_empty()
+            && !report_mode_intersection
```
chrisduerr: Why pass this in as parameter instead of just calculating it inside?

alacritty_terminal/src/input.rs
```diff
@@ -587,40 +605,62 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+        let report_mode_intersection = self.ctx.terminal().mode().intersects(report_modes);
+        let side = self.ctx.mouse().cell_side;
+
         self.ctx.mouse_mut().click_state = match self.ctx.mouse().click_state {
             ClickState::Click
                 if !button_changed && elapsed < self.mouse_config.double_click.threshold =>
             {
                 self.ctx.mouse_mut().block_url_launcher = true;
-                self.on_mouse_double_click(button, point);
+                if self.is_expand_selection(button, modifiers, report_mode_intersection) {
```
chrisduerr: We start new selections in `on_mouse_double_click`, why should we expand it here instead of inside `on_mouse_double_click`? This is the same for `on_mouse_triple_click`.

alacritty_terminal/src/input.rs
```diff
@@ -587,40 +605,62 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+                    // update current selection
+                    if let Some(point) = point {
+                        self.ctx.update_selection_as(point, side, SelectionType::Simple);
+                    }
+                } else {
+                    self.ctx.clear_selection();
+
+                    // Start new empty selection
+                    let side = self.ctx.mouse().cell_side;
+                    if let Some(point) = point {
```
chrisduerr: We unpack the point from the option in both of these branches. Seems like it's better extracted to outside of them so the code isn't unnecessarily duplicated.

alacritty_terminal/src/selection.rs
```diff
@@ -40,35 +40,39 @@ use crate::term::{Search, Term};
...
 /// [`lines`]: enum.Selection.html#method.lines
 #[derive(Debug, Clone, PartialEq)]
+pub enum SelectionType {
+    Simple,
+    Semantic,
+    Lines,
+}
+
+/// Describes a region of a 2-dimensional area
+///
```
chrisduerr: Trailing whitespace.

alacritty_terminal/src/selection.rs
```diff
@@ -81,89 +85,175 @@ pub trait Dimensions {
...
     pub fn update(&mut self, location: Point<usize>, side: Side) {
-        // Always update the `end`; can normalize later during span generation.
         match *self {
-            Selection::Simple { ref mut region } | Selection::Block { ref mut region } => {
-                region.end = Anchor::new(location.into(), side);
-            },
-            Selection::Semantic { ref mut region } | Selection::Lines { ref mut region } => {
-                region.end = location.into();
-            },
+            Selection::Normal {ref mut region } | Selection::Block {ref mut region } => {
```
chrisduerr: Formatting.

alacritty_terminal/src/selection.rs
```diff
@@ -81,89 +85,175 @@ pub trait Dimensions {
...
+        match *self {
+            Selection::Normal {ref mut region } | Selection::Block { ref mut region } => {
+                region.end.ty = selection_type;
+            }
+        }
+        self.update(location, side);
+    }
+
+    // Expand selection following the similar rules as creating a span
+    pub fn expand_selection(term: &mut Term) {
```
chrisduerr: Formatting of this function is very messed up. Additionally the control flow is incredibly confusing, so I don't really feel like it makes much sense to review it in its current state. This should be simplified to make it easier to maintain and understand. Extraction in sub functions might be one option.

alacritty_terminal/src/selection.rs
```diff
@@ -174,26 +264,30 @@ impl Selection {
...
-            Selection::Lines { .. } => Selection::span_lines(term, start, end),
+            Selection::Normal { .. } => {
+                // Simple Selection
+                let (start, end): (Anchor, Anchor) = self.span_simple(cols, start, end)?;
+
+                // Semantic expansion
+                let (start, end) = Selection::span_semantic(term, start, end);
+
+                // Line expansion
+                let (start, end) = Selection::span_line(cols, start, end);
```
chrisduerr: It's a bit unidiomatic and bad for performance to attempt every type of expansion regardless of what previous expansions have resulted in. A side will never require simple expansion, semantic expansion **and** line expansion, since it can only be one.

Ben496: I'll refactor this code to check the expansion type and then call the required expansion function.

alacritty_terminal/src/selection.rs
Ben496: Yeah, I agree that this function is messy. I'll do some cleanup on it.

XVilka:
There are conflicts, so should be rebased on top of the master. Or, probably better, on top of https://github.com/jwilm/alacritty/pull/2438

nixpulvis:
Based on some discussion in #3144, I'm thinking it might make a lot of sense to add an action for this. It seems like we may not want to have right click be a default action for modifying the selection, but some terminals use this binding, and it would be nice to support that configuration.

chrisduerr:
Thanks for all your pathfinding work on this. Thanks to your existing code and some major refactorings to the selection code, this was trivial to implement on top of https://github.com/alacritty/alacritty/pull/3315, so I've sent a PR to do that: https://github.com/alacritty/alacritty/pull/3322.

PR #2367, opened 2019-04-25
Fix cursor disappearing

### LINKED ISSUE #2353: Redraw failure
Since the last two releases I've been experiencing these redrawing errors with neovim/tmux on macOS. On some lines the screen simply does not redraw. Reverting back to 0.3.0, the problem disappears again.

Gif of the problem:
screen

OS: macOS Mojave 10.14.4
Alacritty version: 0.3.1/0.3.2

Config changes (all working in 0.3.0):
- 0 padding
- no decorations
- fonts
- colors scheme
- two keybindings

[config file](https://gist.github.com/SV3A/7fd92b5a50e13bc34117d2fed35539ed)

Furthermore, the following commands don't report anything unordinary during the event:
```
alacritty -vvv
alacritty --print-events
```

### LINKED ISSUE #2355: Insert cursor disappearing in Neovim
I've have noticed an issue with the cursor in **INSERT** mode in `neovim`. After having `neovim` open in `alacritty` for a while the cursor disappears when in insert mode. I'm having a hard time replicating the issue.

Peek 2019-04-23 10-52

Worth noting is that restarting `neovim` does not fix the issue. I have to restart `alacritty` for it to disappear. I did not notice this issue with version `0.3.0`.

**EDIT:** I also now noted that the cursor does not respect my set color. I have: 

```
  cursor:
    text: '0x343434'
    cursor: '0xf8f8f2'
```
But it looks like this:
Peek 2019-04-23 10-58

I'm running `alacritty` built from https://github.com/jwilm/alacritty/commit/4fbae5e397c514c245432a6415a4c99c52b7978f on **Arch Linux** with Xorg.

chrisduerr:
The cfc20d4f34dca535654cc32df18e785296af4cc5 commit introduced a
regression which would cause the cursor to disappear after the glyph
cache has been filled.

Since the cursor was not cached on the glyph cache, the cursor would
quickly fill up the OpenGL texture with lots of cursor textures and then
things would break after the atlas was filled completely.

This adds a separate cursor cache which is keyed by the cursor style
that will persist the texture without flooding the atlas.

This fixes #2355.
This fixes #2353.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - PTY size not getting updated when message bar is shown
+- Text Cursor disappearing
 
 ## Version 0.3.2
 
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -343,7 +343,7 @@ pub trait Handler {
 }
 
 /// Describes shape of cursor
-#[derive(Debug, Eq, PartialEq, Copy, Clone, Deserialize)]
+#[derive(Debug, Eq, PartialEq, Copy, Clone, Hash, Deserialize)]
 pub enum CursorStyle {
     /// Cursor is a block like `▒`
     Block,
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -26,6 +26,7 @@ use font::{self, FontDesc, FontKey, GlyphKey, Rasterize, RasterizedGlyph, Raster
 use glutin::dpi::PhysicalSize;
 use notify::{watcher, DebouncedEvent, RecursiveMode, Watcher};
 
+use crate::ansi::CursorStyle;
 use crate::config::{self, Config, Delta};
 use crate::gl;
 use crate::gl::types::*;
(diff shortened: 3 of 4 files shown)


kchibisov:
Seems like I can't reproduce it anymore, but I wasn't able to enter this [logic](https://github.com/jwilm/alacritty/blob/e964af8a5ea390d96667e80873060d33079134fd/src/renderer/mod.rs#L1056) after 15 minutes of intensive scrolling the file, so I can't check it in situation where it broke.

chrisduerr:
This fix should always keep the cursor in the first atlas, meaning you should never reach that logic to move it to the second atlas.

So that sounds like the fix worked as intended, thanks for the feedback.

jokeyrhyme:
@chrisduerr @jwilm do you know if there are plans to tag a release that includes this fix? The latest (0.3.2) is from before this fix was merged
Cheers, love your work <3

cole-h:
Chris is waiting for #2438 to be merged before he releases 0.3.3 I believe. Progress has been slow but steady. From my testing, it is working fine right now except for a minor cursor issue. I expect it will be out soon™, hopefully!

chrisduerr:
> Chris is waiting for #2438 to be merged before he releases 0.3.3 I believe. Progress has been slow but steady. From my testing, it is working fine right now except for a minor cursor issue. I expect it will be out soon™, hopefully!

That has changed, I don't really have any hope that this will be released within the next few weeks, so unless I should get some new insights, a new RC should be released this weekend (or earlier).

You can take a look at the 0.3.3 milestone to see what I'd still like to tackle before that. It's mostly minor things.

PR #2368, opened 2019-04-25
Perform icon loading on unix non macos platforms

kchibisov:
On linux some wms relies on `_NET_WM_ICON` when reporting application/window icon ( e.g. awesomewm with its tasklist and titlebars), so to set this we should call `with_window_icon` as we did it on windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - On macOS, there's a ToggleSimpleFullscreen action which allows switching to
     fullscreen without occupying another space
 - A new window option `startup_mode` which controls how the window is created
+- `_NET_WM_ICON` property is set on X11 now, allowing for WMs to show icons in titlebars
 
 ### Changed
 
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -52,6 +52,8 @@ winapi = { version = "0.3.7", features = ["impl-default", "winuser", "synchapi",
 dirs = "1.0"
 widestring = "0.4"
 mio-anonymous-pipes = "0.1"
+
+[target.'cfg(not(target_os = "macos"))'.dependencies]
 image = "0.21.0"
 
 [target.'cfg(target_os = "macos")'.dependencies]
--- a/alacritty_terminal/src/window.rs
+++ b/alacritty_terminal/src/window.rs
@@ -18,19 +18,20 @@ use crate::gl;
 use glutin::dpi::{LogicalPosition, LogicalSize, PhysicalPosition, PhysicalSize};
 #[cfg(not(any(target_os = "macos", windows)))]
 use glutin::os::unix::EventsLoopExt;
-#[cfg(windows)]
+#[cfg(not(target_os = "macos"))]
 use glutin::Icon;
 use glutin::{
     self, ContextBuilder, ControlFlow, Event, EventsLoop, MouseCursor, PossiblyCurrent,
     WindowBuilder,
 };
-#[cfg(windows)]
+#[cfg(not(target_os = "macos"))]
 use image::ImageFormat;
 
 use crate::cli::Options;
 use crate::config::{Decorations, StartupMode, WindowConfig};
 
-#[cfg(windows)]
+// It's required to be in this directory due to the `windows.rc` file
+#[cfg(not(target_os = "macos"))]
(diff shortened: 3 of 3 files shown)


src/window.rs
```diff
@@ -33,6 +33,9 @@ use crate::config::{Decorations, StartupMode, WindowConfig};
 #[cfg(windows)]
 static WINDOW_ICON: &'static [u8] = include_bytes!("../extra/windows/alacritty.ico");
 
+#[cfg(not(any(target_os = "macos", target_os = "windows")))]
+static WINDOW_ICON: &'static [u8] = include_bytes!("../extra/linux/alacritty.ico");
```
chrisduerr: Why can't we use the Windows icon here? I don't see any reason why this should be duplicated.

src/window.rs
```diff
@@ -283,6 +286,8 @@ impl Window {
     ) -> WindowBuilder {
         use glutin::os::unix::WindowBuilderExt;
 
+        let icon= Icon::from_bytes_with_format(WINDOW_ICON, ImageFormat::ICO).unwrap();
```
chrisduerr: We can just pass this to the `get_platform_window` method, then we don't have to duplicate this code either.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - On macOS, there's a ToggleSimpleFullscreen action which allows switching to
     fullscreen without occupying another space
 - A new window option `startup_mode` which controls how the window is created
+- Alacritty now loads its icon on unix platforms, except macOS
```
chrisduerr: This statement is a bit misleading, since this change should be completely unrelated to Wayland for example.

I'd be a bit more specific here and say that the `_NET_WM_ICON` property is set on X11 now, allowing for WMs to show icons in titlebars.

src/window.rs
kchibisov: For me it was strange to point to `windows` resources from linux feature and since I can't move icon to it's own folder ( due to windows rc files requirements) I decided to go this way. If using windows folder is ok, I'll change it.

kchibisov:
I'll test it on wayland in a moment.

src/window.rs
chrisduerr: It should be possible to use relative paths for the window.rs file, right?

I also think that we shouldn't point to windows files for the linux target, but we can just use the `extra/logo` folder and put the ico there.

kchibisov: [No](https://docs.microsoft.com/en-us/windows/desktop/menurc/icon-resource), that's why i did it this way.

chrisduerr: Microsoft strikes again I guess. In that case I'd go for just using the `.ico` from the windows directory. Maybe adding a comment that says it's required to be in that directory because of the .rc would help explain this in the future though.

kchibisov:
It doesn't crash, but in the same time it doesn't do anything ( tested in sway ). So, yes, I should change changelog to X11 only

src/window.rs
```diff
@@ -18,19 +18,20 @@ use crate::gl;
...
-#[cfg(windows)]
+#[cfg(not(target_os = "macos"))]
 use image::ImageFormat;
 
 use crate::cli::Options;
 use crate::config::{Decorations, StartupMode, WindowConfig};
 
-#[cfg(windows)]
+// It's required to be in this direcroty due to `windows.rc` file
+#[cfg(not((target_os = "macos"))]
```
chrisduerr: There's too many parenthesis here.

kchibisov: Thx, seems like I'm a little bit blind today.

src/window.rs
```diff
@@ -18,19 +18,20 @@ use crate::gl;
...
 };
-#[cfg(windows)]
+#[cfg(not(target_os = "macos"))]
 use image::ImageFormat;
 
 use crate::cli::Options;
 use crate::config::{Decorations, StartupMode, WindowConfig};
 
-#[cfg(windows)]
+// It's required to be in this direcroty due to `windows.rc` file
```
chrisduerr:
```suggestion
// It's required to be in this directory due to the `windows.rc` file
```

I spotted another minor error.

src/window.rs
nixpulvis: Would it make any sense to store a single file, and copy them into the appropriate locations in a `build.rs`, script? Too heavy?

chrisduerr: Yeah, at that point we're just adding way too much complexity for anyone to intuitively understand what's going on I think.

This way we have only one point where a workaround is in place and it allows us to document it at that same place too. So I think this is the best solution we have available right now.

PR #2371, opened 2019-04-26
Fix double-width characters only highlighting halfway

CrackedP0t:
Currently, `RenderableCellsIter` skips over the `WIDE_CHAR_SPACER`, as it's empty, and doesn't underline it, resulting in only the left half of the character being underlined. This PR makes it so that `RenderableCellsIter` always underlines every cell in `url_highlight`, even if it's empty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- URLs with double-width characters not correctly highlighting
 - PTY size not getting updated when message bar is shown
 - Text Cursor disappearing
 - Incorrect positioning of zero-width characters over double-width characters
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+
+
+
 # Changelog
 All notable changes to this project will be documented in this file.
 
@@ -21,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- URLs with double-width characters not correctly highlighting
+- Double-width characters in URLS only being highlit on the left half
 - PTY size not getting updated when message bar is shown
 - Text Cursor disappearing
 - Incorrect positioning of zero-width characters over double-width characters
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,3 @@
-
-
-
 # Changelog
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- URLs with double-width characters not correctly highlighting
```
chrisduerr: The word `correctly` is often a bit difficult in a change log since people might not know what `correct` means.

Being more concrete should help here, so this might be better:

```suggestion
- URLs ending in double-width characters not highlighting the last cell
```

src/term/mod.rs
```diff
@@ -418,15 +418,12 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
...
-                    continue;
-                }
-
                 // Underline URL highlights
                 if self.url_highlight.as_ref().map(|range| range.contains_(index)).unwrap_or(false)
                 {
                     cell.inner.flags.insert(Flags::UNDERLINE);
+                } else if cell.is_empty() && !selected {
+                    continue;
                 }
```
chrisduerr: I feel like putting the code in the `else` block here is more confusing than helpful. While it does improve performance (unless LLVM is smart enough to optimize it itself, which I wouldn't be surprised about), I feel like it harms readability. And the performance difference will never show up in any benchmarks, because the url highlighting is not in any hot path.

So I'd propose something more like this:

```rust
if !cell.is_empty() || selected {
    return Some(RenderableCell::new(self.config, self.colors, cell, selected));
}
```

Since `cell.is_empty()` checks if the underline flag is set, this should not affect functionality, but would hopefully make it extremely clear that this conditional is not related to url highlighting at all, but just skips all empty cells.

CHANGELOG.md
CrackedP0t: OK, that's a good point, although all double-width characters in URLs are affected, not just ones at the end

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+
+
+
```
chrisduerr: Not sure where this came from, but it was probably unintentional?

CHANGELOG.md
```diff
@@ -21,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Double-width characters in URLS only being highlit on the left half
```
chrisduerr: I completely forgot that this is not only the case at the end of URLs, but everywhere. Thanks for catching that and going the correct route.

This sounds good to me now, though maybe styling URLS as URLs is a little more in line with the other occurrences in the change log. But this is such a minor thing that I don't really care *that* much either way.

src/term/mod.rs
CrackedP0t: Cool, done!

CHANGELOG.md
CrackedP0t: Yeah, no idea

CHANGELOG.md
CrackedP0t: Yeah, 's' is plural, not part of the acronym :smile:

PR #2372, opened 2019-04-26
Add config option to always highlight URLs on mouseover

kchibisov:
In my daily use of Alacritty I've started noticing that I always forget that URLs are clickable, since they are not highlighted unless you press modifier key. So this PR makes them to be highlighted by default, like in other terminals ( e.g. terminator  ). As alternative, config option could be added to control such behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - On macOS, there's a ToggleSimpleFullscreen action which allows switching to
     fullscreen without occupying another space
 - A new window option `startup_mode` which controls how the window is created
+- Added config option to always highlight URLs on mouseover
 
 ### Changed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -326,6 +326,10 @@ mouse:
     #  program: xdg-open
     #  args: []
 
+    # If this is `true`, the URLs will be highlighted on mouseover regardless
+    # of pressed modifiers
+    always_highlight_on_mouseover: false
+
     # URL modifiers
     #
     # These are the modifiers that need to be held down for opening URLs when clicking
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -122,6 +122,10 @@ pub struct Url {
     #[serde(deserialize_with = "deserialize_launcher")]
     pub launcher: Option<CommandWrapper>,
 
+    // When the links will be highlighted
+    #[serde(deserialize_with = "failure_default")]
+    pub always_highlight_on_mouseover: bool,
+
     // Modifier used to open links
     #[serde(deserialize_with = "deserialize_modifiers")]
     pub modifiers: ModifiersState,
@@ -167,6 +171,7 @@ impl Default for Url {
             launcher: Some(CommandWrapper::Just(String::from("open"))),
             #[cfg(windows)]
             launcher: Some(CommandWrapper::Just(String::from("explorer"))),
+            always_highlight_on_mouseover: Default::default(),
             modifiers: Default::default(),
         }
     }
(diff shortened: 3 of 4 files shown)


chrisduerr:
This has been added intentionally and it is absolutely not an option to remove it completely.

I'm willing to consider a configuration option, but I feel like there's little benefit. Under which circumstances do you usually 'forget' that Alacritty supports URLs? Do you use multiple different terminal emulators?

Also would using no modifiers be a possible solution?

kchibisov:
No, I use only alacritty and I've been using it for a long time. When this option was added (without highlighting first, as far as I remember) I've started accidentally click on URLs while performing copying them (No modifiers is not a solution) , so I've set a modifier for url launcher, and since then I've never used it ( until recently I've decided to sync my config with the one from master in which there are no modifiers for url launching). So, this whole situation made me think that interactive elements should provide some feedback ( I've also asked some co-workers about url highlighting in their terminals and most of them was highlighting them by default and use `ctrl or shift + left mouse click` for opening). I'm fine with option for it in a config, also I think this option will provide some familiar functionality for people migrating to alacritty.

chrisduerr:
> I think this option will provide some familiar functionality for people migrating to alacritty

Just to be clear, I do not think it's a good idea to make this the default. It does not correctly show the user when links can be clicked and when not, especially when mouse mode is enabled this can easily lead to confusion. Terminal emulators that just bind it to shift click without giving the user an option to configure it at all do not have this problem, so they shouldn't be used as reference here.

Terminal emulators that just always highlight URLs even when they aren't hovered are a separate thing too, so we can't follow their lead either.

So this makes me ask myself the following question: If we make this an opt-in configuration option, how likely is it that a user is going to find out about it and change it before getting used to Alacritty supporting URL launching? Because based on your usage feedback it seems like this option would get obsolete as soon as the user has gotten 'used to it'.

In general I'm pro config-options if there's a good use for them. But with this I'm not certain how much real world usage it would actually get. So if anyone else is coincidentally following this PR and interested in this, please let me know. It's always easier to add a config option than to remove it, so I'd like to carefully consider it.

I'm also curious about what @nixpulvis thinks about this. It's possible that I'm just a bit too cautious here.

kchibisov:
I've updated PR to reflect changes resulting from our discussion. So it's should be easier to understand what's actually is going on here.

chrisduerr:
I've talked to @nixpulvis on IRC and I don't think this is a good fit for Alacritty right now. URL highlighting is a relatively fresh feature and it's natural that people are still getting used to it, but this band-aid feels more harmful than beneficial in the long run.

It would also make it significantly harder to distinguish when links are actually clickable and it's not unlikely that users enabling this would actually get themselves confused more often than getting anything out of it once they've used it for a while. It would likely be required to change only the underline state without changing the cursor to still represent when a URL is clickable, which would introduce additional complexity to the config option.

If there should be a significant desire for this in the future, I'm willing to reconsider this. However with the low interest right now and the general feedback that prolonged usage might fix the underlying issue anyways, I don't think I'm ready to add a configuration flag for this.

Thanks for proposing this feature with a PR and explaining your reasoning behind it.

nixpulvis:
FWIW, I played around with the following diff to this PR. This prevents the mouse from being styled as a hand when the click wont actually open the URL... However, the lack of mouse style updating when the modifier alone is pressed makes this PR as it stands unsatisfactory.

```diff
-            self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Hand);
+
+            if self.mouse_config.url.modifiers.relaxed_eq(modifiers)
+                && (!self.ctx.terminal().mode().intersects(mouse_mode) || modifiers.shift)
+                && self.mouse_config.url.launcher.is_some()
+            {
+                self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Hand);
+            }
```

I agree with @chrisduerr overall though, and even given a fix for this, I'm not sold on the need for this configuration. Hopefully if clicking URLs  is valuable, you'll eventually get used to the modifier. Otherwise, as mentioned, we could revisit this.

kchibisov:
@nixpulvis , thx for your feedback. It seems like there wasn't any desire for this feature from someone else besides me ( at least in this PR ). But I'm ok with @chrisduerr decision, the most important thing that we can get from this PR right now is that such option were ever discussed, so we could reference this in case someone wants something similar.

PR #2387, opened 2019-04-29
Update INSTALL.md for fresh openSUSE Tumbleweed install

cbcoutinho:
A fresh openSUSE Tumbleweed installation was missing a build dependency


cbcoutinho:
@chrisduerr Indeed this was on a wayland-enabled system running gnome.

Not sure what the differences would be for a fresh X11 system

PR #2394, opened 2019-04-30
Update UPDATE.md with Fedora X11 dependency

krlvi:
This is similar to https://github.com/jwilm/alacritty/pull/2387
On a Fedora system running X11 I had to install the `libxcb-devel` dependency in order to compile.


krlvi:
Hi @chrisduerr I added the Debian dependency. Unfortunately I don't have any other systems at hand to test with

PR #2397, opened 2019-05-01
Add git commit sha to output of --version

nixpulvis:
This changes the output of the `--version` (`-V`) flag. Before it was printing just the version of alacritty, now it will include the current git revision built. For example, this branch currently prints: `alacritty 0.3.2 (531976a)`.

This also fixes a small issue caused by dbd8538762ef8968a493e1bf996e8693479ca783, which caused `alacritty -V` to print `alacritty_terminal 0.3.2`.

- [x] output is using `crate_name!()` now yielding `alacritty_terminal` not `alacritty`


alacritty_terminal/src/cli.rs
```diff
@@ -60,8 +60,11 @@ impl Options {
     pub fn load() -> Options {
         let mut options = Options::default();
 
+        let git_sha = env!("GIT_SHA");
+        let version_string = format!("{} ({})", crate_version!(), git_sha);
```
nixpulvis: I think the solution here might be to move this into the `alacritty` crate itself. It seems like a better home for this.

chrisduerr: Feel free to go ahead with this. It's definitely not going to stay in `alacritty_terminal` in the long run. So moving it to `alacritty` sounds good to me.

matthiaskrgr:
If want, you could also use this crate (it's already used in rls and clippy): 
https://crates.io/crates/rustc_tools_util

nixpulvis:
@matthiaskrgr cool, I've updated this to use that crate. Thanks.


I'm pretty happy with this overall atm. One crazy idea I had though, would be to check for local changes to the repo and print a `alacritty 0.3.2 (531976a*)`  (notice the `*`) when detected. Worth added complexity? I'm happy to keep this simple, but just thought it might help detect weird patches and other discrepancies.

alacritty/build.rs
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+fn main() {
+    println!("cargo:rustc-env=GIT_HASH={}",
+             rustc_tools_util::get_commit_hash().unwrap_or_default())
```
chrisduerr: I'm not sure if we want to `unwrap_or_default` this. I think an `expect("getting commit hash")` might be better because it would be easier to catch errors?

I'm slightly on the edge, but I think it's better than printing no commit hash and not noticing.

alacritty_terminal/src/config/mod.rs
```diff
@@ -22,11 +22,12 @@ use serde::{self, de, Deserialize};
 use serde_yaml;
 
 use crate::ansi::CursorStyle;
-use crate::cli::Options;
 use crate::index::{Column, Line};
 use crate::input::{Action, Binding, KeyBinding, MouseBinding};
 use crate::term::color::Rgb;
 
+pub use self::options::Options;
```
chrisduerr: I'm personally not a huge fan of `pub use` I think that directory structures mostly exist for a good reason and imports should be done at the top of the file so the lengths doesn't play a huge role either.

So I think removing this and using `use crate::config::options::Options` give a better overview over what's actually going on and where to find the struct itself.

nixpulvis: I personally feel like `...::config::Options` is better than `...::config::options::Options`, I'm only really using the file to avoid bloating config/mod.rs. So in this case I don't see the reason for the nesting, but I agree with you in some cases.

Either way, I don't care much, and if you'd be happier with it as `config::options::Options` then I'm fine with that.

chrisduerr: I'd be much happier without re-exporting. That usually just creates a mess an inconsistencies in my experience. I believe we currently have a bunch of instances both using `crate::term::Cell` and `crate::term::cell::Cell` and I would like for that not to happen with this.

nixpulvis: Ok, so I guess my question is then, should I put it in config directly or leave it and change all the `use `statements. I think I'm leaning towards just moving it into `config`. Maybe this leads nicely to some unification we might be able to do there?

chrisduerr: I'd leave it in a separate module right now.

There's nothing wrong with `crate::config::options::Options`.

PR #2416, opened 2019-05-05
Fix regression causing the cursor to only show over the left half of wide characters

CrackedP0t:
Commit 494348abe80f591dfdd68fd4987bafc59fcb32c1 only cached by cursor style, which meant that the cached cursor, which was probably single-width, would be used to draw over wide characters.
This commit caches by height, width, line_width, and style, so any time the cursor requirements change, a new cursor will be cached.

### DIFF
--- a/alacritty_terminal/src/cursor.rs
+++ b/alacritty_terminal/src/cursor.rs
@@ -23,8 +23,8 @@ use crate::ansi::CursorStyle;
 /// Width/Height of the cursor relative to the font width
 pub const CURSOR_WIDTH_PERCENTAGE: i32 = 15;
 
+/// A key for caching cursor glyphs
 #[derive(Debug, Eq, PartialEq, Copy, Clone, Hash, Deserialize)]
-/// An enum that represents all data about the state of the cursor
 pub struct CursorKey {
     pub style: CursorStyle,
     pub is_wide: bool,
--- a/alacritty_terminal/src/cursor.rs
+++ b/alacritty_terminal/src/cursor.rs
@@ -23,29 +23,53 @@ use crate::ansi::CursorStyle;
 /// Width/Height of the cursor relative to the font width
 pub const CURSOR_WIDTH_PERCENTAGE: i32 = 15;
 
-pub fn get_cursor_glyph(
-    cursor: CursorStyle,
-    metrics: Metrics,
-    offset_x: i8,
-    offset_y: i8,
-    is_wide: bool,
-) -> RasterizedGlyph {
-    // Calculate the cell metrics
-    let height = metrics.line_height as i32 + i32::from(offset_y);
-    let mut width = metrics.average_advance as i32 + i32::from(offset_x);
-    let line_width = cmp::max(width * CURSOR_WIDTH_PERCENTAGE / 100, 1);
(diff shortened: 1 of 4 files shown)


CrackedP0t:
We could just cache by wideness and style, but I figured caching by everything would be more future-proof.

chrisduerr:
> We could just cache by wideness and style, but I figured caching by everything would be more future-proof.

There's no need to cache by anything. If cell dimensions change the cache is emptied anyways.

CrackedP0t:
Oh, OK - that makes sense. We still need to cache by `is_wide`, though, or else cached single-width cursors will be used for double-width characters

chrisduerr:
Yeah, that sounds fine to me. Caching width + style seems like a reasonable key for the hashmap.

alacritty_terminal/src/renderer/mod.rs
```diff
@@ -96,7 +96,7 @@ impl ::std::fmt::Display for Error {
         match *self {
             Error::ShaderCreation(ref err) => {
                 write!(f, "There was an error initializing the shaders: {}", err)
-            },
+            }
```
chrisduerr: There are a few formatting regressions that were introduced in this PR, these should be fixed.

CrackedP0t: Oh, whoops, looks like rustfmt didn't see the config file

alacritty_terminal/src/cursor.rs
```diff
@@ -23,6 +23,13 @@ use crate::ansi::CursorStyle;
 /// Width/Height of the cursor relative to the font width
 pub const CURSOR_WIDTH_PERCENTAGE: i32 = 15;
 
+#[derive(Debug, Eq, PartialEq, Copy, Clone, Hash, Deserialize)]
+/// An enum that represents all data about the state of the cursor
```
chrisduerr: The comment should be above the annotation here.

alacritty_terminal/src/renderer/mod.rs
```diff
@@ -994,12 +994,21 @@ impl<'a> RenderApi<'a> {
...
     pub fn render_cell(&mut self, cell: RenderableCell, glyph_cache: &mut GlyphCache) {
         let chars = match cell.inner {
-            RenderableCellContent::Cursor((cursor_style, ref raw)) => {
+            RenderableCellContent::Cursor(cursor_key) => {
                 // Raw cell pixel buffers like cursors don't need to go through font lookup
-                let glyph = glyph_cache
-                    .cursor_cache
-                    .entry(cursor_style)
-                    .or_insert_with(|| self.load_glyph(raw));
+                let metrics = &glyph_cache.metrics;
```
chrisduerr: Instead of borrowing and dereferencing it right after, we probably should just use `glyph_cache.metrics` when calling `get_cursor_glyph`.

If this is done because of the borrow checker, we should at least remove the borrow so we don't borrow and then dereference immediately after.

CrackedP0t: The borrow checker won't let us just write `glyph_cache.metrics`, but the borrow was unnecessary

CrackedP0t:
Alright, thanks for rebasing!

PR #2424, opened 2019-05-07
Fix message bar not updating PTY size on hide

### LINKED ISSUE #2418: Message bar messing up PTY
The message bar which is displayed for warnings and errors can apparently mess up applications like Vim. This is likely because the PTY is not resized appropriately.

I was not yet able to reproduce this myself, it has only been reported to me on IRC.

kchibisov:
Fixes #2418

### DIFF
--- a/alacritty_terminal/src/display.rs
+++ b/alacritty_terminal/src/display.rs
@@ -367,7 +367,10 @@ impl Display {
             }
         }
 
-        if font_changed || self.last_message != terminal.message_buffer_mut().message() {
+        // Message bar update detected
+        let message_bar_changed = self.last_message != terminal.message_buffer_mut().message();
+
+        if font_changed || message_bar_changed {
             if new_size == None {
                 // Force a resize to refresh things
                 new_size = Some(PhysicalSize::new(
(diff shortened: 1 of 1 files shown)


chrisduerr:
Oh, now I understand. We never store the 65 lines as a size but we just pass it to the PTY. So when we grow it we don't have the 65 as last size but the 68 raw size (not PTY size).

So yeah, your solution looks like a good approach then.

alacritty_terminal/src/display.rs
```diff
@@ -368,7 +368,10 @@ impl Display {
             }
         }
 
-        if font_changed || self.last_message != terminal.message_buffer_mut().message() {
+        // Message bar message update detected
```
chrisduerr: I think `Message bar update detected` would be a little easier to read here.

```suggestion
        // Message bar update detected
```

alacritty_terminal/src/display.rs
```diff
@@ -368,7 +368,10 @@ impl Display {
...
+        let message_changed = self.last_message != terminal.message_buffer_mut().message();
```
chrisduerr: I think being a little bit more verbose here would be beneficial. So `message_bar_changed` would imho be the better variable name. It's easy to get confused when talking about messages, so when possible we should not skip the full name here.

The line should also stay under 100 columns, so it's just a little bit more verbose without any formatting disadvantages.

kchibisov:
@chrisduerr , sorry for pinging, but what’s status of this PR?

PR #2426, opened 2019-05-07
Don't attempt shader compilation if not supported.

markuspeloquin:
I was trying to get Alacritty to work in WSL. Hint: it doesn't. But I was getting a useless error that could be better:

```
Created log file at "/tmp/Alacritty-6058.log"
[2019-05-06 14:34] [ERROR] Alacritty encountered an unrecoverable error:

        There was an error initializing the shaders: Failed compiling shader at /home/markus/dist/alacritty/res/text.v.glsl: 

```

It turns out that shader compilation isn't supported, so it never executed and there is no log for `glGetShaderInfoLog()` to return. I guess that's because VcXsrv is starting with `Wgl="True"`, indicating that all OpenGL stuff should be indirectly executed by Windows WGL.

I'm not sure what the solution to that is. Precompiling GLSL files? Wait for WSL 2? I really wanted to use Alacritty :(

Also, I've never written a line of Rust before so this might not be the best.

### DIFF
--- a/alacritty_terminal/src/renderer/mod.rs
+++ b/alacritty_terminal/src/renderer/mod.rs
@@ -1307,6 +1307,16 @@ fn create_shader(
 
     let len: [GLint; 1] = [source.len() as GLint];
 
+    let shader_compiler = unsafe {
+        let mut val: GLboolean = gl::FALSE;
+        gl::GetBooleanv(gl::SHADER_COMPILER, &mut val);
+        val == gl::TRUE
+    };
+    if !shader_compiler {
+        return Err(ShaderCreationError::Compile(PathBuf::from(path),
+                "no shader compiler support".to_string()));
+    }
+
     let shader = unsafe {
         let shader = gl::CreateShader(kind);
         gl::ShaderSource(shader, 1, &(source.as_ptr() as *const _), len.as_ptr());


chrisduerr:
What exactly does this PR attempt to solve? Judging by your comments, it still doesn't work?

I don't really see how the error is useless, it seems like it clearly pointed in the right direction?

markuspeloquin:
What it attempts to solve is to help a user debug their problems without reading source code. Perhaps the message could be better as 'GL implementation doesn't support shader compilation'.

I don't agree that it was clear. It said that compilation failed, but no compilation occurred. I tried searching for the text `"Failed compiling shader"` online and found only irrelevant information.

markuspeloquin:
(As far as getting it to work in WSL, it appears that xming/vcxsrv only support OpenGL 1.4, but Cygwin/X supports 4.6, so it should work. I'm having trouble forwarding OpenGL. It *seems* like for remote connections from WSL, the call to `glXCreateContext()` is failing but `glxinfo`/`glxgears` don't check for the error.)

chrisduerr:
Missing support to compile shaders is just a subset of errors of the failed shader compilation and either way the solution is likely the same, that you'll need to install proper OpenGL drivers or that Alacritty does not support your device.

So I'm going to close this issue since it adds additional checks at startup, which are irrelevant since they don't help with solving the issue anyways.

PR #2429, opened 2019-05-08
Add Ubuntu PPA

mmstick:
Packages are pulled from the Pop!_OS PPA so that Ubuntu and Ubuntu-derivatives may use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ else the instructions to build Alacritty from source can be found [here](INSTALL
 
 ### Pop!\_OS / Ubuntu
 
-> Ubuntu requires a third party repository, with packages copied from the Pop!\_OS PPA.
+> If you're not running Pop!_OS, you'll have to add a third party repository first:
 >
 > ```sh
 > add-apt-repository ppa:mmstick76/alacritty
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,9 +40,13 @@ Precompiled binaries are available from the [GitHub releases page](https://githu
 Some operating systems already provide binaries for Alacritty, for everyone
 else the instructions to build Alacritty from source can be found [here](INSTALL.md).
 
-### Pop!\_OS
+### Pop!\_OS / Ubuntu
 
-Provided in the [Pop!\_OS repositories](https://launchpad.net/~system76/+archive/ubuntu/pop) for 18.04 (bionic) and higher.
+> Ubuntu requires a third party repository, with packages copied from the Pop!\_OS PPA.
```
chrisduerr: I'm not sure if I trust users to understand that this isn't required for pop with this sentence. Being a little more clear would be beneficial I think.

Maybe this makes it more clear:

```
> If you're not running Pop!_OS, you'll have to add a third party repository first:
```

mmstick: Sure. I'll add that.

PR #2443, opened 2019-05-15
Compatible without git

Hentioe:
Follow the discussion plan for [PR#2441](https://github.com/jwilm/alacritty/pull/2441). The ultimate goal is to continue building without `git` installed or without getting a commit hash, and correctly output `--version`.
The build script for openSUSE or other Linux distribution continues to be compatible without any source code patches.

### DIFF
--- a/alacritty/build.rs
+++ b/alacritty/build.rs
@@ -13,10 +13,6 @@
 // limitations under the License.
 
 fn main() {
-    let mut hash = rustc_tools_util::get_commit_hash().unwrap_or("".to_owned());
-    if hash.is_empty() {
-        hash = std::env::var("GIT_HASH").expect("couldn't get commit hash");
-    }
-
+    let hash = rustc_tools_util::get_commit_hash().unwrap_or("".to_owned());
     println!("cargo:rustc-env=GIT_HASH={}", hash);
 }
--- a/alacritty/build.rs
+++ b/alacritty/build.rs
@@ -13,6 +13,6 @@
 // limitations under the License.
 
 fn main() {
-    let hash = rustc_tools_util::get_commit_hash().unwrap_or("".to_owned());
+    let hash = rustc_tools_util::get_commit_hash().unwrap_or_default();
     println!("cargo:rustc-env=GIT_HASH={}", hash);
 }
--- a/alacritty/build.rs
+++ b/alacritty/build.rs
@@ -13,6 +13,10 @@
 // limitations under the License.
 
 fn main() {
-    let hash = rustc_tools_util::get_commit_hash().expect("couldn't get commit hash");
+    let mut hash = rustc_tools_util::get_commit_hash().unwrap_or("".to_owned());
+    if hash.is_empty() {
+        hash = std::env::var("GIT_HASH").expect("couldn't get commit hash");
+    }
+
     println!("cargo:rustc-env=GIT_HASH={}", hash);
 }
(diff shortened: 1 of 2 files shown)


alacritty/build.rs
```diff
@@ -13,6 +13,6 @@
 // limitations under the License.
 
 fn main() {
-    let hash = rustc_tools_util::get_commit_hash().expect("couldn't get commit hash");
+    let hash = rustc_tools_util::get_commit_hash().unwrap_or("".to_owned());
     println!("cargo:rustc-env=GIT_HASH={}", hash);
```
nixpulvis: I think I liked the idea of checking if the environment variable is set already before overwriting it.

Here's an idea for how this might look:

```rust
use std::env;

fn main() {
    match env::var("GIT_HASH") {
        Ok(_) => {},
        _ => {
            let hash = rustc_tools_util::get_commit_hash()
                                        .unwrap_or("".to_owned());
            println!("cargo:rustc-env=GIT_HASH={}", hash);
        },
    }
}
```

alacritty/src/cli.rs
```diff
@@ -62,8 +62,12 @@ impl Options {
...
     pub fn new() -> Self {
         let mut options = Options::default();
-
-        let version_string = format!("{} ({})", crate_version!(), env!("GIT_HASH"));
+        let commit_hash = env!("GIT_HASH");
+        let version_string = if commit_hash.is_empty() {
+            format!("{}", crate_version!())
+        } else {
+            format!("{} ({})", crate_version!(), commit_hash)
+        };
```
nixpulvis: I'd consider moving this into a `version` function for `Options`.

alacritty/build.rs
chrisduerr: > I think I liked the idea of checking if the environment variable is set already before overwriting it.

I don't think I do. What do we gain by that? Best case scenario we save ourselves some compile-time code, which is basically free relative to the complete compile time of Alacritty.

However, worst-case the hash might be outdated or contain completely garbage info (like the hash from a different project, set by the user manually). That could lead to some really annoying results when trying to troubleshoot bugs.

Since this is not a standardized environment variable set by git, I'd highly recommend against just assuming that when it's set, it's gonna be set properly.

alacritty/src/cli.rs
chrisduerr: I also think that the following might be a bit cleaner:

```rust
let mut version = crate_version!().to_owned();
if !commit_hash.is_empty() {
    version = format!("{} ({})", version, commit_hash);
}
```

alacritty/build.rs
```diff
@@ -13,6 +13,6 @@
 // limitations under the License.
 
 fn main() {
-    let hash = rustc_tools_util::get_commit_hash().expect("couldn't get commit hash");
+    let hash = rustc_tools_util::get_commit_hash().unwrap_or("".to_owned());
```
chrisduerr: The correct method to use here is `unwrap_or_else(|| _)`. We should also switch `"".to_owned()` into something a little more clear like `String::new`:

```suggestion
    let hash = rustc_tools_util::get_commit_hash().unwrap_or_else(String::new);
```

alacritty/build.rs
nixpulvis: Fair enough, I was thinking from the last PR there was some desire to be able to set this manually when it's not inside a git repository. Though now I'm thinking I agree with you, if it's not in a git directory, the solution to getting it to build with git info would be to clone the repo.

alacritty/build.rs
quininer: just `unwrap_or_default()`

chrisduerr: Absolutely correct, thanks!

Issue #2409, opened 2019-05-03
alacritty 0.3.2:  URL parser drops trailing /


UnixRonin:
Which operating system does the issue occur on?  Linux

If on linux, are you using X11 or Wayland?  xorg-server 1.20.4

If a URL ends with a trailing /, alacritty's URL parser in 0.3.2 is (sometimes?  always?  **LOOKS** like always, in all of my test cases so far) dropping the trailing slash, which **in most cases** causes the link to fail.

I'm fairly sure this is a regression from 0.2.x.

chrisduerr:
> which in most cases causes the link to fail

That should not be the case. The trailing slash is dropped intentionally, but it should not cause the link to fail.

Trailing slashes should usually be optional, could you share an example on when this will break a link?

UnixRonin:
https://www.leaf.tv/articles/difference-between-virginia-black-forest-ham/

Without the trailing /, this link does not resolve.

UnixRonin:
(I should probably have said "in most of my test cases".)

chrisduerr:
In my experience this is more than extremely unusual. Since usually pages end in `.html` and the `.html` is truncated, leaving just the name.

With a trailing slash at the end, that would make a rewrite from `site.html` to `site/` which doesn't make much sense to me.

But trying to keep context whenever possible makes sense, the reason trailing slashes are stripped is primarily to be able to strip other content.

So the text `Testing (my website is alacritty.io/testing)/` should currently resolve properly to the `alacritty.io/testing` URL, but if we do not allow to truncate `/` at the end it will resolve to `alacritty.io/testing)/`.

So I think there is some benefit in truncating slashes from the end, but the question is probably how common this scenario is. I think removing the truncation just for `/` should be okay, but either way will likely introduce some issues. That's just the nature of the problem with having to parse URLs inside plain text.

UnixRonin:
Hmm.  Yeah, not a simple problem, and I've never yet seen a text parser that someone couldn't construct a crafted example to break.

I suspect the problem mainly arises when URLs point to dynamically generated content.

Normally, http://foo......bar/ is parsed as http://foo......bar/index.html or whatever the designated default index is.  I've been doing further testing on a wider variety of URLs and found that most of the time it does work, but I found a significant number of cases where it didn't.

I thought that () were illegal characters in URLs which should cause your above example to be correctly parsed anyway, but looking it up, I see that () are not only allowed in URLs but considered "safe".

[...]

PR #2449, opened 2019-05-16
Allow URLs to end with trailing slash

shreve:
Addresses #2409


CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Selections not automatically expanding across double-width characters
 - On macOS, automatic graphics switching has been enabled again
 - Text getting recognized as URLs without slashes separating the scheme
+- URL parser drops trailing slashes from valid URLs
```
cole-h: s/drops/dropping/

"Fixed: URL parser drops trailing slashes from valid URLs" vs "Fixed: URL parser dropping trailing slashes from valid URLs" -- flows better imo, but just a nit.

PR #2465, opened 2019-05-21
Move the project description to the center

Gymmasssorla:


chrisduerr:
Thanks for proposing this, however I've tested this myself previously and centered text is significantly more difficult to read than normal left-aligned text.

So I think it's best to keep things the way they are right now.

Issue #540, opened 2017-05-03
Sometimes alacritty doesn't fully render output


maximbaz:
Quite often alacritty doesn't finish rendering the output, it is reproducible even after executing the same command over and over again. When it's stuck, if I hit Enter, this will cause it to refresh window, but until I do anything inside the window, it will remain like that.

It's best described with a gif:

![alacritty](https://cloud.githubusercontent.com/assets/1177900/25645462/9866ced4-2fae-11e7-987f-1389363ae17d.gif)

It's not the same as #459, as here window is always focused.

System info: Arch, i3, https://github.com/jwilm/alacritty/commit/f298e0455ab549aac1835ba59467ce06111b9bc9 installed via AUR.

jwilm:
@maximbaz  thanks for the report!

This is a bummer; I thought we had squashed that issue last year before release. Did this start recently for you?

maximbaz:
Well, to be honest I was waiting for glyph offset to be merged, so I only started using alacritty yesterday.

I wanted to do a git bisect, but I can't build almost any old commits, either rust language changes or something else.

I managed to build e0980df from 2 months ago, and there the issue exists.

I should also mention that I removed my custom config and the issue still persists, so it is not a config problem.

iv597:
Affects me as well. As seen in the attached picture, the bottom row of neovim is about 25% cut off for me.

Installed from AUR:
```
local/alacritty-git 0.1.0.492.g6659810-1
    A cross-platform, GPU enhanced terminal emulator
```

Window manager is herbstluft if it makes any difference.

![1493838893](https://cloud.githubusercontent.com/assets/799439/25677255/da5b38fa-300a-11e7-899d-ec7bdaf94005.png)

jwilm:
Would one of you mind trying to build from git, directly? I haven't been able to repro this, and you are both building from AUR.

maximbaz:
I did that already, when I mentioned that I reproduced the issue on a 2 month old commit https://github.com/jwilm/alacritty/commit/e0980df48ef50ddb48021d9375ef005eac268b7e.

At that time I also built the `master` from git, and both times I could reproduce it.

And by "built" I mean I cloned the repo and ran `cargo build --release`. This is what AUR is doing anyway, it is cloning repo and running this one command.

jwilm:
@maximbaz ah, thanks for clarifying.

Since I'm unable to repro at this point, I'll try staring at the code for a bit for anything obviously wrong.

[...]

Issue #2217, opened 2019-03-26
Lagging while using nano, vim, etc.


svaderia:
Which operating system does the issue occur on?
`Ubuntu 16.04`
If on linux, are you using X11 or Wayland?
`X11`

When I use nano it isn't loaded immediately, after I press a key it gets rendered. even while typing in nano, A key press will load the previous key and not the current key. Even when exiting, the save file option is shown only after I press some key. This issue remains with any editor I use. Even when using `git log --graph --pretty='\''%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset'\'' --abbrev-commit` command, it gets rendered only after I press some key. Following is the video to showcase the issue. 

![simplescreenrecorder-2019-03-26_14 22 30](https://user-images.githubusercontent.com/26005890/54984228-6ce32f80-4fd4-11e9-98c0-f0c079789787.gif)

If you need any other info related to system config or want me to try out some experiments, I can help.

EDIT 1 : Using `alacritty 0.2.7`

svaderia:
Another thing I noticed is, everything works perfectly when I use a new instance of alacritty. But when an instance stays inactive for a while, it starts lagging. Sometimes writing in the terminal starts lagging.

cole-h:
Can you reproduce this in another terminal emulator? What if you try a different shell (`bash`, `zsh`, `fish`)? And finally, can you see if this is reproducible on the latest master? I don't have this issue and that's what I'm currently running. Even checking out the 0.2.7 tag doesn't give me the same behavior.

chrisduerr:
Could you try running Alacritty with `alacritty --print-events` to show if Alacritty is updated with a delay or if the key events are actually coming in *after* the key is pressed (with a delay).

svaderia:
@cole-h 

> What if you try a different shell (`bash`, `zsh`, `fish`)?

The issue remains irrespective of what I use.


> Can you reproduce this in another terminal emulator?

I'm unable to reproduce.


> can you see if this is reproducible on the latest master?

I will try it and let you know.

svaderia:
@chrisduerr, Following is the output of the `print-events`. 

```
[2019-03-26 23:57] [INFO] Welcome to Alacritty
[2019-03-26 23:57] [INFO] Configuration loaded from "/home/shyamal/.config/alacritty/alacritty.yml"
[...]

PR #2467, opened 2019-05-22
Bugfix to rerender the output #540

andys8:
This PR would change the optimization level. I think this is most likely _not_ an option, is it?
Are there other options to fix the behavior?

See comment of @maximbaz for more info <https://github.com/jwilm/alacritty/issues/540#issuecomment-304479048>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ members = [
 lto = true
 debug = 1
 incremental = false
+opt-level = 1
 
 [patch.crates-io]
 servo-freetype-sys = { path = "servo-freetype-proxy" }


chrisduerr:
> I think this is most likely not an option, is it?

It is not. This is likely a racing condition or something similar. Just making Alacritty slower until the racing condition doesn't pop up anymore is not an option.

I personally can't reproduce the issue at all though, so it's tough for me to look into it.

andys8:
Thanks for the reply. I guess it's very tricky, if the issue is not reproducible. I'm wondering if it's related to i3 or i3 users simply aren't using their mouses often, which trigger's a refresh 😁

chrisduerr:
@andys8 I'm using i3 myself, so I doubt it.

Also just to make sure because your phrasing sounds a bit ambiguous, the #540 issue is about incorrect text getting rendered and actual elements missing.

If you run into issues with Alacritty not refreshing/redrawing when it should, you might want to take a look at #2217.

andys8:
You're right! And I'm happy to read, that it might be solved :) https://github.com/jwilm/alacritty/issues/2217#issuecomment-489367208

PR #2470, opened 2019-05-22
Re-invert cursor when in selection

CrackedP0t:
Currently, when the cursor is inside the selection, it isn't visible, because both it and the selection are the inverse of the regular terminal. This PR inverts the cursor and its cell again, making sure that the cursor is still visible.


nixpulvis:
This seems to be working as intended. Thanks for implementing this!

chrisduerr:
I didn't have any time to look at this yet, but did you check out alternative cursor shapes when testing this @nixpulvis? Currently this bug only exists with the block cursor, so I'd hope that this fix doesn't break the behavior with all the other cursors.

nixpulvis:
@chrisduerr yep, I've tested all three of: `Beam`, `Underline`, and `Block` (the focus of this PR) work as I'd expect. This branch also fixs an issue of always using block inside a selection, so now each cursor mode is fully respected, **and** visible in selections.

I don't exactly know what you mean by "Currently this bug only exists with the block cursor" however, since (unless my machine is different in some way) all selections use a re-re-inverted (aka normal) block in place of the cursor, which should only be the case for a `Block`.

PR #2478, opened 2019-05-25
Add Gentoo install dependencies

chrisduerr:
While Gentoo was listed in the table of contents of the INSTALL.md, the
instructions to install the build dependencies were missing.

This adds the emerge command necessary to install all dependencies of
Alacritty (other than Rust) so building from source without the overlay
is possible.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -156,7 +156,7 @@ Alacritty. Here's a `eopkg` command that should install all of them. If
 something is still found to be missing, please open an issue.
 
 ```sh
-sudo eopkg install fontconfig-devel
+eopkg install fontconfig-devel
 ```
 
 #### NixOS/Nixpkgs
@@ -175,7 +175,7 @@ command should install all of them. If something is still found to be missing,
 please open an issue.
 
 ```sh
-emerge media-libs/fontconfig media-libs/freetype:2 sys-devel/make dev-util/cmake dev-util/pkgconfig
+emerge --onlydeps x11-terms/alacritty
 ```
 
 #### Windows
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -168,6 +168,16 @@ dependencies on [NixOS](https://nixos.org).
 nix-shell -A alacritty '<nixpkgs>'
 ```
 
+#### Gentoo
+
+On Gentoo, you need a few extra libraries to build Alacritty. The following
+command should install all of them. If something is still found to be missing,
+please open an issue.
+
+```sh
+emerge media-libs/fontconfig media-libs/freetype:2 sys-devel/make dev-util/cmake dev-util/pkgconfig
+```
+
 #### Windows
 
 On windows you will need to have the `{architecture}-pc-windows-msvc` toolchain


kchibisov:
Should we mention runtime dependencies?
```
 sys-libs/zlib
 x11-libs/libXcursor
 x11-libs/libXi
 x11-libs/libXrandr
 x11-misc/xclip
 virtual/opengl
```

chrisduerr:
I don't think we should. XClip isn't necessary anymore and the X/Wayland dependencies aren't added for the other platforms and rather obvious. So I think this should be good.

However I have not been able to test this myself, since I don't have a gentoo system. So these are basically just guesses. The main goal was to get the absolute basics on what's necessary to just build Alacritty, similar to what we already have for the other platforms. I expect most people to use the overlay.

I don't think cmake is necessary, since I don't have it on my system. But I've left it for now since we do depend on the cmake crate.

kchibisov:
I'm a using `Gentoo` on my main system and I'm building `Alacritty` from source everytime (from git, not emerge). You need only `media-libs/fontconfig` and  `media-libs/freetype:2` to build `Alacritty` (cmake, make and pkgconfig are "essentials" in `Gentoo` since you can't live without them). But it wouldn't run unless you have `DRI` and all runtime dependencies.

chrisduerr:
> make and pkgconfig are "essentials" in Gentoo since you can't live without them

Yeah that's kinda what I expected, though adding them should do no harm, right? That way it's the same as all the other platforms and we're on the safe side.

I'm fine with not listing runtime dependencies unless they're unusual. Something like `X` or `Wayland` being required should be fine.

kchibisov:
Yeah ( you'll just recompile them in bad case). There's a tricky part about `dri` (USE FLAG), since you can ran `X`/`Wayland` without it, but `Alacritty` can behave very strange ( it's my experience on setting up `Gentoo`).

chrisduerr:
Can you specify that use flag in the emerge command? Like `USE=dri emerge ...`? Or what would you recommend. I've installed Gentoo only once for a few weeks, so you're probably more familiar with what the best approach for this would be.

kchibisov:
It's a global USE FLAG, so it should go in `/etc/portage/make.conf`. But it really depends on your system setup ( in my case, I wasn't able to launch `Alacritty` in fullscreen in awesomewm due to 'DRI' 2 related crash). Probably `opengl` flag is also needed, but  I've never tested without it.

chrisduerr:
@kchibisov Do you think it makes sense to keep it simple the way it is right now, or would it make more sense to expand and include the runtime dependencies?

I'd imagine a full list with runtime dependencies might bloat that section quite a bit.

kchibisov:
It depends, if you want to `emerge <mentioned dependencies>` ,`git clone`, `cargo install` you'll be surprised if it'll crash on startup, but you can add reference to official ebuild file which covers all runtime dependencies, so it'll be easier to troubleshoot. Right now I think we should keep it the way it's in this PR ( probably add reference to ebuild file) and If there are a lot of problems, such as "I can not run` Alacritty` on `Gentoo`", then it'll make sense. You can remove make ( you'll always have it, even without package) and pkgconfig ( same as make) in case you want shorter install section.

chrisduerr:
Sounds like a reasonable suggestion, thanks!

lovesegfault:
The command as listed is problematic. It selects all the packages into the world set, while most of those probably shouldn't be there, such as CMake. It'd be more reasonable to add the oneshot (`-1` or `--oneshot`) flag to portage (i.e. `emerge`) so that the packages are emerged, but not selected.

This, however, would cause them to be removed during a depclean.

lovesegfault:
I think the best reference would be to link to the Alacritty ebuild and tell the user to go see the deps listed there. Those will always be up-to-date and they're available locally to any Gentoo user as well (`/usr/portage/x11-terms/alacritty/*.ebuild`). Seems easier to be than us maintaining what fundamentally is a duplicate of the `DEPEND`, `RDEPEND`, and `BDEPEND` portions of the ebuild here.

chrisduerr:
That doesn't seem completely unreasonable either. Probably a good idea though that does cause problems when you're offline (but then it's likely you won't have much luck with cargo or emerge anyways).

lovesegfault:
@chrisduerr I think the best instructions might be to tell them to do `sudo emerge --onlydeps x11-terms/alacritty` which, well, emerges only the ebuild deps but not the package itself. If you add `-p` or `--pretend` then it effectively only lists them.

chrisduerr:
@lovesegfault Thanks for your advice, that sounds like a reasonable solution and is very similar to what is done on nix anyways.

PR #2482, opened 2019-05-26
Skip UTF-8 BOM when reading config file

kazarmy:
I'm currently stuck with Notepad when editing the config file and it insists on adding a BOM when saving it. Alacritty errors when loading the config file due to it and so this patch adds code to skip the BOM.

Note that my Rust knowledge is basically based on Googling for how-to-do-stuff and interpreting the results, so any improvements to the code are welcome.

### DIFF
--- a/alacritty/src/config.rs
+++ b/alacritty/src/config.rs
@@ -176,8 +176,17 @@ pub fn reload_from(path: &PathBuf) -> Result<Config> {
 }
 
 fn read_config(path: &PathBuf) -> Result<Config> {
-    let mut contents = String::new();
-    File::open(path)?.read_to_string(&mut contents)?;
+    let mut bytes_vec = Vec::new();
+    let mut f = File::open(path)?;
+    f.read_to_end(&mut bytes_vec)?;
+    let mut bytes: &[u8] = &bytes_vec;
+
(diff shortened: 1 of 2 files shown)


alacritty/src/config.rs
```diff
@@ -176,8 +176,17 @@ pub fn reload_from(path: &PathBuf) -> Result<Config> {
...
+    let mut f = File::open(path)?;
+    f.read_to_end(&mut bytes_vec)?;
+    let mut bytes: &[u8] = &bytes_vec;
+
+    // Skip UTF-8 BOM
+    if bytes.starts_with(&[0xef, 0xbb, 0xbf]) {
+        bytes = &bytes[3..];
+    }
+
+    let contents = String::from_utf8_lossy(&bytes);
```
chrisduerr: We should be able to keep the current code, then add this to truncate the start of the string:
```rust
// Remove UTF-8 BOM
if buf.chars().nth(0) == Some('\u{FEFF}') {
    buf = buf.split_off(3);
}
```

kazarmy: Ok 52b8d065051e0ba1bc4ff383135be6c32c2a9821

PR #2495, opened 2019-06-02
Add foreground/background request escape codes

rbong:
When a "?" is given rather than a color in xterm for foreground/background escape codes, the current color is returned. This can be and is used by programs such as vim to preserve readability.

I have based the output format directly on xterm.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1875,11 +1875,7 @@ impl ansi::Handler for Term {
     fn dynamic_color_sequence<W: io::Write>(&mut self, writer: &mut W, code: u8, index: usize) {
         trace!("Writing escape sequence for dynamic color code {}: color[{}]", code, index);
         let color = self.colors[index];
-        let _ = write!(
-            writer,
-            "\x1b]{};rgb:{:x}/{:x}/{:x}\x07",
-            code, color.r, color.g, color.b
-        );
+        let _ = write!(writer, "\x1b]{};rgb:{:x}/{:x}/{:x}\x07", code, color.r, color.g, color.b);
     }
 
     /// Reset the indexed color to original value
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1870,6 +1870,18 @@ impl ansi::Handler for Term {
         self.color_modified[index] = true;
     }
 
+    /// Write a foreground/background color escape sequence with the current color
+    #[inline]
+    fn dynamic_color_sequence<W: io::Write>(&mut self, writer: &mut W, code: u8, index: usize) {
+        trace!("Writing escape sequence for dynamic color code {}: color[{}]", code, index);
+        let color = self.colors[index];
+        let _ = write!(
+            writer,
+            "\x1b]{};rgb:{:x}/{:x}/{:x}\x07",
+            code, color.r, color.g, color.b
+        );
+    }
+
     /// Reset the indexed color to original value
     #[inline]
     fn reset_color(&mut self, index: usize) {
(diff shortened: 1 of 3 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -1870,6 +1870,20 @@ impl ansi::Handler for Term {
         self.color_modified[index] = true;
     }
 
+    /// Get the foreground/background color
+    #[inline]
+    fn request_special_color<W: io::Write>(&mut self, writer: &mut W, code: u8, index: usize) {
+        let color = self.colors[index];
+        let _ = write!(
+            writer,
+            "\x1b]{};rgb:{:x}{:x}/{:x}{:x}/{:x}{:x}",
```
chrisduerr: The docs say the following about this escape:
> If a "?" is given rather than a name or RGB specification,
          xterm **replies with a control sequence of the same form** which
          can be used to set the corresponding dynamic color.  Because
          **more than one pair of color number and specification can be
          given in one control sequence**, xterm can make more than one
          reply.

The most important part here is that the control sequence we return needs to be identical to the way the color is set, this is currently not the case. The biggest issue is the missing BEL character at the end, so a `\x07` needs to be added here.

The next problem is that Alacritty doesn't handle RGB in the specified format. Instead of accepting `rgb:rrrr/gggg/bbbb` it only accepts `rgb:rr/gg/bb` from what I can tell. Since I haven't written this code I'm not 100% certain about its correctness, but we should respond with what we can actually understand.

The last problem is the part about multiple codes. If multiple parameters are specified (like `\x1b]10;?;?\x07`), then we should print the escapes for setting both the foreground color (10) and the color following it (11 = background). So the `10` just defines the first color which will be set/get and every parameter after that is used to set/get the following colors.

It seems like VTE also doesn't handle this case of setting multiple colors in one escape, however it does seem like it would be better if we supported it. It seems trivial to me to add support, however if we do, it should be added for both setting and getting. If you want to handle that in this PR that would be fine, otherwise I'd open an issue separately and ignore that part for this PR.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1870,6 +1870,20 @@ impl ansi::Handler for Term {
         self.color_modified[index] = true;
     }
 
+    /// Get the foreground/background color
+    #[inline]
+    fn request_special_color<W: io::Write>(&mut self, writer: &mut W, code: u8, index: usize) {
```
chrisduerr: I'm not a huge fan of the method name here, I think `request_` indicates that the terminal is requesting something, however it's responding to a request.

Since they're often referred to as `dynamic` colors, I also think that `dynamic` is a better way to name this than `special`.

I think something like `dynamic_color_request` would be a better way to name this, however I'm open for suggestions here.

chrisduerr: In (most) of the other methods we add some trace output at the start of the handler functions. I think it would be worthwhile to do the same here.

rbong: Added in a trace [here](https://github.com/rbong/alacritty/blob/fc3f86eded9af7b45b18c7a8354dda8e3ff93daf/alacritty_terminal/src/term/mod.rs#L1876)

rbong: Good point, I think `dynamic_color_sequence` is more indicative of what the function actually does as oposed to why it is being called. I've changed it to that, let me know if I need to change it again.

alacritty_terminal/src/term/mod.rs
rbong: I've added in the bell and I've switched to rr/gg/bb. I have too many questions about how multiple request codes actually work. I think I'd like to keep this PR to single requests. I've created #2497 for this.

rbong: Hold off on merging this, the new format causes some problems in Vim.

rbong: Never mind, it's handling it fine. My was overwriting the background color and I thought Vim was not picking up the color correctly. Things should be able to handle this code fine if Vim can.

chrisduerr: I don't think this escape is used 'in the wild' anyways, so it should be fine to stick to what we have now.

rbong:
I've added the feature to the changelog

PR #2504, opened 2019-06-04
Follow full xparsecolor spec in escape sequences

rbong:
Escape sequences in xterm are parsed according to xparsecolor. xparsecolor supports 1, 2, 3, and 4 digit hex colors. Previously, only 2 digits were supported.

This also fixes a bug where "fX" was parsed as "0xf" instead of "0xf0". xterm handles invalid hex values as 0, except for '/'.

The response to a request for fg/bg must be a valid escape sequence. Most terminals output a 4-digit hex response on request for fg/bg color. This changes the response to 4-digit, improving compatibility.

- [xterm control sequence document referencing xparsecolor](https://www.xfree86.org/current/ctlseqs.html)
- [xparsecolor spec describing the valid colors](https://linux.die.net/man/3/xparsecolor)
- [context for why we were using 2-digit color hex responses](https://github.com/jwilm/alacritty/pull/2495#discussion_r289637940)

### DIFF
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -24,59 +24,56 @@ use vte;
 
 use crate::term::color::Rgb;
 
-fn is_rgb_separator(ch: Option<char>) -> bool {
-    ch == None || ch == Some('/') || ch == Some('\x07')
+// Parse colors in XParseColor format
+fn xparse_color(color: &[u8]) -> Option<Rgb> {
+    if !color.is_empty() && color[0] == b'#' {
+        let len = color.len().saturating_sub(1);
+        parse_legacy_color(&color[1..len])
+    } else if color.len() >= 4 && &color[..4] == b"rgb:" {
+        let len = color.len().saturating_sub(1);
+        parse_rgb_color(&color[4..len])
+    } else {
+        None
+    }
 }
 
-// Parse color arguments
-// Expect the color argument in the form "rgb:x/x/x" or "#xxx", where x is 1-4 digit hex numbers
+// Parse colors in `rgb:r(rrr)/g(ggg)/b(bbb)` format
 fn parse_rgb_color(color: &[u8]) -> Option<Rgb> {
-    let len = color.len();
-    // Legacy colors are in the form "#xxx"
-    // The use of this format is no longer encouraged by xparsecolor
-    let mut is_legacy = false;
+    let colors = str::from_utf8(color).ok()?.split('/').collect::<Vec<_>>();
 
-    if len >= 5 && color[0] == b'#' {
-        is_legacy = true;
-    } else if len < 10 || &color[..4] != b"rgb:" {
+    if colors.len() != 3 {
         return None;
     }
 
-    let mut iter = color.iter().map(|v| *v as char).skip(if is_legacy { 1 } else { 4 });
-    let mut colors = [0; 3];
+    // Scale values instead of filling with `0`s
+    let scale = |input: &str| {
+        let max = u32::pow(16, input.len() as u32) - 1;
+        let value = u32::from_str_radix(input, 16).ok()?;
+        Some((255 * value / max) as u8)
+    };
 
-    for color in &mut colors {
-        // parse the first digit
-        let first_digit = iter.next().and_then(|v| v.to_digit(16))? as u8;
-        *color = first_digit << 4;
+    Some(Rgb {
+        r: scale(colors[0])?,
+        g: scale(colors[1])?,
+        b: scale(colors[2])?,
+    })
+}
 
-        // parse the remaining 3 digits
-        for i in 0..4 {
-            // legacy color format does not have separators, break based on length
-            if is_legacy && (i == 3 || len < 8 + i * 3) {
-                break;
-            }
+// Parse colors in `#r(rrr)g(ggg)b(bbb)` format
+fn parse_legacy_color(color: &[u8]) -> Option<Rgb> {
+    let item_len = color.len() / 3;
 
-            let next = iter.next();
+    // Truncate/Fill to two byte precision
+    let color_from_slice = |slice: &[u8]| {
+        let col = usize::from_str_radix(str::from_utf8(slice).ok()?, 16).ok()? << 4;
+        Some((col >> (4 * slice.len().saturating_sub(1))) as u8)
+    };
 
-            if !is_legacy && is_rgb_separator(next) {
-                if i == 0 {
-                    // "rgb:" format colors treat single-digit colors as repeated
-                    *color += first_digit;
-                }
-                break;
-            } else if i == 3 {
-                // expect to have encountered a separator
-                return None;
-            }
-
-            if i == 0 {
-                *color += next.and_then(|v| v.to_digit(16))? as u8;
-            }
-        }
-    }
-
-    Some(Rgb { r: colors[0], g: colors[1], b: colors[2] })
+    Some(Rgb {
+        r: color_from_slice(&color[0..item_len])?,
+        g: color_from_slice(&color[item_len..item_len * 2])?,
+        b: color_from_slice(&color[item_len * 2..])?,
+    })
 }
 
 fn parse_number(input: &[u8]) -> Option<u8> {
@@ -777,7 +774,7 @@ where
                 if params.len() > 1 && params.len() % 2 != 0 {
                     for chunk in params[1..].chunks(2) {
                         let index = parse_number(chunk[0]);
-                        let color = parse_rgb_color(chunk[1]);
+                        let color = xparse_color(chunk[1]);
                         if let (Some(i), Some(c)) = (index, color) {
                             self.handler.set_color(i as usize, c);
                             return;
@@ -802,7 +799,7 @@ where
                                 break;
                             }
 
-                            if let Some(color) = parse_rgb_color(param) {
+                            if let Some(color) = xparse_color(param) {
                                 self.handler.set_color(index, color);
                             } else if param == b"?" {
                                 self.handler.dynamic_color_sequence(writer, dynamic_code, index);
@@ -1199,7 +1196,7 @@ fn attrs_from_sgr_parameters(parameters: &[i64]) -> Vec<Option<Attr>> {
             37 => Some(Attr::Foreground(Color::Named(NamedColor::White))),
             38 => {
                 let mut start = 0;
-                if let Some(color) = parse_color(&parameters[i..], &mut start) {
+                if let Some(color) = parse_sgr_color(&parameters[i..], &mut start) {
                     i += start;
                     Some(Attr::Foreground(color))
                 } else {
@@ -1217,7 +1214,7 @@ fn attrs_from_sgr_parameters(parameters: &[i64]) -> Vec<Option<Attr>> {
             47 => Some(Attr::Background(Color::Named(NamedColor::White))),
             48 => {
                 let mut start = 0;
-                if let Some(color) = parse_color(&parameters[i..], &mut start) {
+                if let Some(color) = parse_sgr_color(&parameters[i..], &mut start) {
                     i += start;
                     Some(Attr::Background(color))
                 } else {
@@ -1252,7 +1249,7 @@ fn attrs_from_sgr_parameters(parameters: &[i64]) -> Vec<Option<Attr>> {
 }
 
 /// Parse a color specifier from list of attributes
-fn parse_color(attrs: &[i64], i: &mut usize) -> Option<Color> {
+fn parse_sgr_color(attrs: &[i64], i: &mut usize) -> Option<Color> {
     if attrs.len() < 2 {
         return None;
     }
@@ -1452,7 +1449,7 @@ pub mod C1 {
 #[cfg(test)]
 mod tests {
     use super::{
-        parse_number, parse_rgb_color, Attr, CharsetIndex, Color, Handler, Processor,
+        parse_number, xparse_color, Attr, CharsetIndex, Color, Handler, Processor,
         StandardCharset, TermInfo,
     };
     use crate::index::{Column, Line};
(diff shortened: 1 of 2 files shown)


alacritty_terminal/src/ansi.rs
```diff
@@ -28,28 +28,64 @@ use crate::term::color::Rgb;
...
-                digit <<= 4;
-                digit += value as u8;
+            let mut value: u8 = 0;
+            if let Some(digit) = convert_digit!(next) {
+                value = digit as u8;
+                value <<= 4;
             }
-            digit
+
+            if peek!() != Some('/') {
```
rbong: I'm doing a last once-over and I realized that this structure will eat the bell character. Not eating the bell character was the entire reason for making the iterator peekable. Fixing this now.

rbong:
Besides the issue above I will add this to the changelog.

rbong:
Everything looks good from my end now.

rbong:
Looks like rustfmt failed on CI and it failed a check for cognitive complexity. I will run rustfmt. Any tips for bringing down cognitive complexity is appreciated, since I'm not pretty new to Rust, but I will figure it out if not. I'll run any local checks before pushing next time.

chrisduerr:
I've only taken a super small peek at this, but this looks like extreme amounts of macros. You'll want to keep macros to a minimum since they add a lot of complexity and compile time, instead you should prefer abstracting over complexity with methods when possible.

rbong:
Good to know, thanks

rbong:
Hopefully that should fix rustfmt and clippy problems.

As far as the issues with macros goes, I removed some unnecessary macros. I was able to cut out one macro that was originally in the function by using `.map` instead. I'm about as far as I can get reducing the complexity of this function while somewhat preserving the original style, if I need to take a step further and refactor parsing colors entirely now that it's a bit more complex, let me know.

alacritty_terminal/src/ansi.rs
```diff
@@ -25,60 +25,88 @@ use vte;
...
-        };
+// Parse a color argument in the form "#xxxxxx"
+// The use of this format is no longer encouraged by xparsecolor
+fn parse_legacy_rgb_color(color: &[u8]) -> Option<Rgb> {
+    if color.len() < 7 || color[0] != b'#' {
+        return None;
     }
 
+    let mut iter = color.iter().map(|v| (*v as char));
+    iter.next();
```
chrisduerr: Instead of creating an iterator and then calling `iter.next()`, it should be possible to just add `.skip(1)` after the `map`.

alacritty_terminal/src/ansi.rs
```diff
@@ -25,60 +25,88 @@ use vte;
...
+
+    let r = parse_hex!();
+    let g = parse_hex!();
+    let b = parse_hex!();
 
-            let next = next!().and_then(|v| v.to_digit(16));
-            if let Some(value) = next {
-                digit <<= 4;
-                digit += value as u8;
+    Some(Rgb { r, g, b })
```
chrisduerr: I know that you've probably just extended what already existed in this method, but calling a macro that takes things out of the current scope and even mutates it is *really* bad style. So I'd like to take the chance to improve the style and do it right here.

This is what I'd propose for this method to look like, including the previous change I've mentioned:

```rust
fn parse_legacy_rgb_color(color: &[u8]) -> Option<Rgb> {
    if color.len() < 7 || color[0] != b'#' {
        return None;
    }

    let mut iter = color.iter().map(|v| (*v as char)).skip(1).skip(1);

    let mut colors = [0; 3];
    for i in 0..3 {
        colors[i] = iter.next().and_then(|v| v.to_digit(16))? as u8;

        colors[i] <<= 4;

        colors[i] += iter.next().and_then(|v| v.to_digit(16))? as u8;
    }

    Some(Rgb { r: colors[0], g: colors[1], b: colors[2] })
}
```

alacritty_terminal/src/ansi.rs
```diff
@@ -25,60 +25,88 @@ use vte;
...
+                    hex += digit as u8;
+                }
+                if (0..3).find(|_| is_hex_separator(iter.next())) == None {
+                    return None;
+                }
             }
-            digit
+
+            hex
         }};
```
chrisduerr: Since this is pretty much a direct copy of the legacy rgb code, we should extract it to a separate function.

The `/` separation should have no impact on the legacy color parsing anyways, so we can just have one function that handles both.

alacritty_terminal/src/ansi.rs
```diff
@@ -25,60 +25,88 @@ use vte;
...
                 return None;
             }
-            if next!() != Some('b') {
+            if iter.next() != Some('b') {
                 return None;
             }
-            if next!() != Some(':') {
+            if iter.next() != Some(':') {
                 return None;
             }
```
chrisduerr: I think this would be a worthwhile improvement:

```rust
let chars: Vec<u8> = iter.take(3).map(|c| *c as u8).collect();
if std::str::from_utf8(&chars) != Ok("gb:") {
    return None;
}
```

rbong:
I have refactored the function based on the example given above - thank you, that was a much better starting point. All of the issues should be resolved. There are no macros, legacy colors are parsed in the same function, and a simple loop is used.

alacritty_terminal/src/ansi.rs
rbong: I went for `len < 9 || &color[..4] != b"rgb:"` instead, which is similar to the legacy color approach - check the length and then check the beginning of the string before initializing the iterator.

rbong:
I am working on the CI issues.

rbong:
Fixed clippy and rustfmt problems, it's ready to be reviewed again whenever.

alacritty_terminal/src/ansi.rs
```diff
@@ -25,62 +25,49 @@ use vte;
...
+}
+
+// Expect that color argument looks like "rgb:x[xxx]/x[xxx]/x[xxx]" or "#xxxxxx"
+fn parse_rgb_color(color: &[u8]) -> Option<Rgb> {
+    let len = color.len();
+    // Legacy colors are in the form "#xxxxxx"
+    // The use of this format is no longer encouraged by xparsecolor
+    let mut is_legacy = false;
+
+    if len >= 2 && color[0] == b'#' {
```
chrisduerr: This should probably be `>= 4`, since `#RGB` is at least 4 characters.

alacritty_terminal/src/ansi.rs
```diff
@@ -25,62 +25,49 @@ use vte;
...
-        _ => None,
+        *color += next.and_then(|v| v.to_digit(16))? as u8;
+
+        if is_legacy {
+            continue;
+        }
+
+        if (0..3).find(|_| is_rgb_separator(iter.next())) == None {
+            return None;
+        }
```
chrisduerr: We should not just skip it if colors have additional precision. XParseColors has very clear instructions on how additional depth should be handled.

alacritty_terminal/src/ansi.rs
```diff
@@ -25,62 +25,49 @@ use vte;
...
-            iter.next().map(|v| *v as char)
-        };
+fn is_rgb_separator(ch: Option<char>) -> bool {
+    ch == None || ch == Some('/') || ch == Some('\x07')
+}
+
+// Expect that color argument looks like "rgb:x[xxx]/x[xxx]/x[xxx]" or "#xxxxxx"
+fn parse_rgb_color(color: &[u8]) -> Option<Rgb> {
+    let len = color.len();
+    // Legacy colors are in the form "#xxxxxx"
```
chrisduerr: This statement is not true. Legacy colors can be in the format `#RGB` up to `#RRRRGGGGBBBB`. This should be handled properly, if we're already trying to improve the XParseColor compatability.

alacritty_terminal/src/ansi.rs
```diff
@@ -25,62 +25,49 @@ use vte;
 use crate::term::color::Rgb;
 
 // Parse color arguments
```
chrisduerr: This comment is outdated/out of place.

alacritty_terminal/src/ansi.rs
rbong: [Updated](https://github.com/rbong/alacritty/blob/cdbcd8357fc2955c011785ed02bd93b386509561/alacritty_terminal/src/ansi.rs#L39-L43). I used 5 instead and started using the bell character instead, and modified the length check for non-legacy colors as well. Tests have been updated to use the bell character.

alacritty_terminal/src/ansi.rs
rbong: I'm sorry, but I can't find any references of what to do with the additional bit depth. If you could point me in the right direction I would apprecitae it.

alacritty_terminal/src/ansi.rs
rbong: Full disclosure, my main goal with this PR is really to support 4-digit color queries, which has already been added to the main branch separate from this issue. Improving XParseColor compatibility was just a prerequisite. However I am happy to help and I have added support for more color formats [here](https://github.com/rbong/alacritty/blob/cdbcd8357fc2955c011785ed02bd93b386509561/alacritty_terminal/src/ansi.rs#L53-L72) with tests [here](https://github.com/rbong/alacritty/blob/cdbcd8357fc2955c011785ed02bd93b386509561/alacritty_terminal/src/ansi.rs#L1547-L1550)

alacritty_terminal/src/ansi.rs
rbong: [Moved](https://github.com/rbong/alacritty/blob/cdbcd8357fc2955c011785ed02bd93b386509561/alacritty_terminal/src/ansi.rs#L31-L32)

chrisduerr:
I thought I'd answer in a top-level comment for visibility and to make things a bit easier to read.

Regarding the way additional depth should be calculated, you can find the full information I know about here:
https://linux.die.net/man/3/xparsecolor

To go into a bit more detail, the legacy #RRGGBB colors are supposed to define the most significant bits of the color. So #RGB would be the same as #R000G000B000.

The rgb: colors are handled a bit differently. If not enough depth is specified, they're supposed to be scaled. So `f[fff]` would be equal to `1.0`/`255` and `[0/000]` is equal to `0`. So a value of `42` would be equal to `25.88`/`66`, if we'd scale that back up it would be `428F` instead of `4200` or `4242`. At least that's how I understand it based on this documentation.

> Full disclosure, my main goal with this PR is really to support 4-digit color query responses, which has already been added to the main branch for an unrelated issue. Improving XParseColor compatibility was just a prerequisite. However I am happy to help and I have added support for more color formats here with tests here

If you're not interested in adding full XParseColor support, we can decide on a subset of it too. Though I feel like it would be nice to get this out of the way. In theory with XLookupColors even colors like `red` would be valid, however I'd understand if the line would be drawn here.

If we decide to go all the way, it would probably be a nice idea to separate that into a different workspace module (/crate/library) so we could offer an XParseColor crate for the broader Rust community, however that might be a bigger project than anyone's signed up for with this PR. :)

rbong:
>The rgb: colors are handled a bit differently. If not enough depth is specified, they're supposed to be scaled. So f[fff] would be equal to 1.0/255 and [0/000] is equal to 0. So a value of 42 would be equal to 25.88/66

I missed that, thank you. It should now be fixed and tested. I thought you meant that we were supposed to do something with the third and fourth bit.

>If you're not interested in adding full XParseColor support, we can decide on a subset of it too.

I think at this point, if I missed a format, I'd be more comfortable with creating an issue than fixing it here. Though obviously, if some color format is parsed incorrectly here, I will fix it.

>If we decide to go all the way, it would probably be a nice idea to separate that into a different workspace module

It's a good idea, but I wouldn't be comfortable with maintaining a rust library of any size right now given my skill level with the language.

alacritty_terminal/src/ansi.rs
```diff
@@ -24,63 +24,59 @@ use vte;
...
-    macro_rules! next {
-        () => {
-            iter.next().map(|v| *v as char)
-        };
+    let len = color.len();
+    // Legacy colors are in the form "#xxx"
+    // The use of this format is no longer encouraged by xparsecolor
+    let mut is_legacy = false;
+
+    if len >= 5 && color[0] == b'#' {
```
chrisduerr: We shouldn't account for the bell character here, instead it should be stripped before passing the string to `parse_rgb_color`. Otherwise it might cause problems when we want to reuse this in other places, since `4` is a perfectly valid length.

alacritty_terminal/src/ansi.rs
```diff
@@ -1544,13 +1540,30 @@ mod tests {
     }
 
     #[test]
-    fn parse_valid_rgb_color() {
-        assert_eq!(parse_rgb_color(b"rgb:11/aa/ff"), Some(Rgb { r: 0x11, g: 0xaa, b: 0xff }));
+    fn parse_valid_rgb_colors() {
+        assert_eq!(parse_rgb_color(b"rgb:11/aa/ff\x07"), Some(Rgb { r: 0x11, g: 0xaa, b: 0xff }));
+        assert_eq!(parse_rgb_color(b"rgb:f/ed1/cb23\x07"), Some(Rgb { r: 0xff, g: 0xed, b: 0xcb }));
+        assert_eq!(parse_rgb_color(b"rgb:f/e/d\x07"), Some(Rgb { r: 0xff, g: 0xee, b: 0xdd }));
+    }
```
chrisduerr: Would be nice to cover 16-bit colors in a test too.

alacritty_terminal/src/ansi.rs
```diff
@@ -24,63 +24,59 @@ use vte;
...
 
-            let next = next!().and_then(|v| v.to_digit(16));
-            if let Some(value) = next {
-                digit <<= 4;
-                digit += value as u8;
-            }
-            digit
-        }};
-    }
+    for color in &mut colors {
```
chrisduerr: I think we can simplify this whole block a bit.

Instead of having a lot of branching statements what should be done depending on `i` and the different modes we're currently in, I think it might be useful to just pull as many characters as each color is long and then parse them together in bulk.

So we'd first collect all elements into a string, str, [u8; X], [char; X] or similar. This can be done for legacy mode by dividing the length by three or in the new mode by taking from the iterator while `!is_separator` is true.

Now that we have that, instead of going for branching we can just use `std` and some math to calculate the correct value.

I've got a playground link that should help with how this can be done, this should be simple to adapt depending on the different input types one might choose as most reasonable (I think for us, `str` is probably best):
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=fff750f405a6b7e1578ce80916163dd5

rbong:
Awesome, thank you!

PR #2506, opened 2019-06-04
RPM packaging

pschyska:
I added a specfile template that enables building alacritty with [rpkg](https://docs.pagure.org/rpkg-util/), for example in [COPR](https://copr.fedorainfracloud.org). I would be happy to maintain an inofficial COPR repo that we could mention in README.md for Fedora and EL7/Centos users.

I hope I got all the building and installation steps right, but it seems to work fine in my testing.

Is libEGL a build or a runtime dependency? I was able to build without the package mesa-libEGL-devel . I could add mesa-libEGL (no -dev) as Requires: if it's a runtime dependency.

I had to base this on #2443, as the rpmbuild chroot is not a git repository, and the build failed otherwise.
Unfortunately, I was unable to place the files into extra/linux/rpm or similar, because rpkg expects the macros in the project root.

Lastly, I made some assumptions when generating the RPM package version: I take the output of git describe, replace dashes with dots and drop the leading "v" from the tag name. I didn't want to repeat the crate version in the spec template so it doesn't need to be bumped for every release.
(N.B. `git_version` is a built-in rpkg macro, but I override it because it expects a specific tagging scheme in the repo).

### DIFF
--- a/extra/linux/redhat/alacritty.spec
+++ b/extra/linux/redhat/alacritty.spec
@@ -7,7 +7,7 @@ URL:           https://github.com/jwilm/alacritty
 VCS:           https://github.com/jwilm/alacritty.git
 Source:        alacritty-%{version}.tar
 
-BuildRequires: rust >= 1.31.0
+BuildRequires: rust >= 1.32.0
 BuildRequires: cargo
 BuildRequires: cmake
 BuildRequires: freetype-devel
--- a/extra/linux/redhat/alacritty.spec
+++ b/extra/linux/redhat/alacritty.spec
@@ -1,0 +1,56 @@
+Name:          alacritty
+Version:       0.3.2
+Release:       1%{?dist}
+Summary:       A cross-platform, GPU enhanced terminal emulator
+License:       ASL 2.0
+URL:           https://github.com/jwilm/alacritty
+VCS:           https://github.com/jwilm/alacritty.git
+Source:        alacritty-%{version}.tar
+
+BuildRequires: rust >= 1.31.0
+BuildRequires: cargo
+BuildRequires: cmake
+BuildRequires: freetype-devel
+BuildRequires: fontconfig-devel
+BuildRequires: libxcb-devel
+BuildRequires: desktop-file-utils
+BuildRequires: python36
+
+%description
+Alacritty is a terminal emulator with a strong focus on simplicity and
+performance. With such a strong focus on performance, included features are
+carefully considered and you can always expect Alacritty to be blazingly fast.
+By making sane choices for defaults, Alacritty requires no additional setup.
+However, it does allow configuration of many aspects of the terminal.
+
+%prep
+%setup -q -n alacritty-%{version}
+
+%build
+cargo build --release
+
+%install
+install -p -D -m755 target/release/alacritty         %{buildroot}%{_bindir}/alacritty
+install -p -D -m644 extra/linux/alacritty.desktop    %{buildroot}%{_datadir}/applications/alacritty.desktop
+install -p -D -m644 extra/logo/alacritty-term.svg    %{buildroot}%{_datadir}/pixmaps/Alacritty.svg
+install -p -D -m644 alacritty.yml                    %{buildroot}%{_datadir}/alacritty/alacritty.yml
+tic     -xe alacritty,alacritty-direct \
+                    extra/alacritty.info       -o    %{buildroot}%{_datadir}/terminfo
+install -p -D -m644 extra/completions/alacritty.bash %{buildroot}%{_datadir}/bash-completion/completions/alacritty
+install -p -D -m644 extra/completions/_alacritty     %{buildroot}%{_datadir}/zsh/site-functions/_alacritty
+install -p -D -m644 extra/completions/alacritty.fish %{buildroot}%{_datadir}/fish/vendor_completions.d/alacritty.fish
+install -p -D -m644 extra/alacritty.man              %{buildroot}%{_mandir}/man1/alacritty.1
+
+%check
+desktop-file-validate %{buildroot}%{_datadir}/applications/alacritty.desktop
+
+%files
+%{_bindir}/alacritty
+%{_datadir}/applications/alacritty.desktop
+%{_datadir}/pixmaps/Alacritty.svg
+%{_datadir}/alacritty/alacritty.yml
+%{_datadir}/terminfo/a/alacritty*
+%{_datadir}/bash-completion/completions/alacritty
+%{_datadir}/zsh/site-functions/_alacritty
+%{_datadir}/fish/vendor_completions.d/alacritty.fish
+%{_mandir}/man1/alacritty.1*
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -1,0 +1,14 @@
+version := $(shell rpmspec -q --srpm --qf "%{version}\n" extra/linux/redhat/alacritty.spec)
+commands = git
+
+srpm: $(commands)
+	$(eval top := $(shell mktemp -d))
+	mkdir -p "$(top)/SOURCES"
+	git archive HEAD --output "$(top)/SOURCES/alacritty-$(version).tar" --prefix "alacritty-$(version)/"
+	rpmbuild -bs "$(spec)" --define "_topdir $(top)" --define "_srcrpmdir $(outdir)"
+	rm -rf "$(top)"
+
+$(commands):
+	command -v $@ &> /dev/null || dnf -y install $@
+
+.PHONY: srpm $(commands)


hargoniX:
If this spec file should ever get accepted into the repository it'd be my pleasure to try and get it into the official Fedora repositories (I'm already a package maintainer there so getting it in there would be a matter of sticking to the guidelines and opening an issue on the Fedora Bugzilla instead of the longer process for first time package maintainers where you have to find a sponsor etc. (which speaking from experience can take weeks or even months)) So hit me up once you're done with it maybe :wink:

pschyska:
Awesome, thanks! I will.

hargoniX:
Oh and just for your information, if we were to actually add this to Fedora we wouldn't be allowed to use cargo build do download and build all the dependencies but would instead have to specify each dependency as a BuildRequires attribute as fedora like many other Linux distribution s likes to keep control of the source of their packages instead of relying on some external registry

chrisduerr:
This sounds like a great proposal and I appreciate the help of both of you.

For an external build system it makes some sense to have the .spec.rpkg in our repository, however I'd expect that Fedora has the configurations for all packages stored in their own git repository?

Is that the case @hargoniX? If so, is there any point in adding this to the Alacritty repository too, or should this just be added to the Fedora package repo instead?

hargoniX:
Yeah Fedora has a completely self contained build system so adding the spec file to this repository would only make sense if we were to go for a COPR repository as all build files would be stored on Fedora infrastructure anyways.

pschyska:
> For the record I have tried packaging rust binary packages before and due to fedoras policy of only having one version of a lib at a time it can actually be pretty hard as for example some sub dependencies might rely on several different versions of popular libraries like lazy_static etc. But I'd definitely give this one a try and suggest the following patches if required

But isn't that only for dynamically-linked libraries? Alacritty only links these libraries when compiled via this spec:
```
$ ldd /usr/bin/alacritty
	linux-vdso.so.1 (0x00007ffd4edc3000)
	libfontconfig.so.1 => /lib64/libfontconfig.so.1 (0x00007f7bb8059000)
	libfreetype.so.6 => /lib64/libfreetype.so.6 (0x00007f7bb7f9c000)
	libxcb.so.1 => /lib64/libxcb.so.1 (0x00007f7bb7f71000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f7bb7f6b000)
	libutil.so.1 => /lib64/libutil.so.1 (0x00007f7bb7f66000)
	librt.so.1 => /lib64/librt.so.1 (0x00007f7bb7f5c000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f7bb7f39000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f7bb7f1f000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f7bb7d59000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f7bb7c13000)
	libexpat.so.1 => /lib64/libexpat.so.1 (0x00007f7bb7be5000)
	libuuid.so.1 => /lib64/libuuid.so.1 (0x00007f7bb7bdb000)
	libbz2.so.1 => /lib64/libbz2.so.1 (0x00007f7bb7bc5000)
	libpng16.so.16 => /lib64/libpng16.so.16 (0x00007f7bb7b8f000)
	libz.so.1 => /lib64/libz.so.1 (0x00007f7bb7b75000)
	libXau.so.6 => /lib64/libXau.so.6 (0x00007f7bb7b6f000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f7bb87b3000)
```
These are all packaged.

pschyska:
> For an external build system it makes some sense to have the .spec.rpkg in our repository, however I'd expect that Fedora has the configurations for all packages stored in their own git repository?
> 
> Is that the case @hargoniX? If so, is there any point in adding this to the Alacritty repository too, or should this just be added to the Fedora package repo instead?

I think there is value in having this now for two reasons:

1. We can have a unofficial COPR linked in README.md for Fedora users to use as soon as this is merged (or as soon as the next tag, i.e. 0.3.3). Submitting to Fedora proper sounds like a "1.0" thing to do and it might take a while until the project considers that. This would of course be possible when the spec is managed externally, but it's less friction that way.
2. As the rpmspec has to evolve with the software in terms of library dependencies, and things to install (contributions like shell completions, icons etc.) and possible metadata (Summary, Description and things like that) it should live in the repo IMHO. That way contributors might make the effort to update it when they add something related.

pschyska:
I just rebased the branch on f2b1bfd

pschyska:
After rebasing, the RPM (still) builds: https://copr.fedorainfracloud.org/coprs/pschyska/alacritty/build/928060/

hargoniX:
@pschyska to your earlier comment about the libraries, I am talking about the rust libraries here, C is perfectly fine and always works but if we for example consider the parking_lot = "0.7" requirement in alacritty_terminal, I can almost guarantee you that very likely some of the sub dependencies of this project will also use parking_lot however if the version is different that is incompatible with the Fedora idea of libraries as only one version may be available at a time (even if they are source packages only). So if that should be the case, building this with a Fedora packaging specification compliant spec file is impossible without bumping the versions etc.

chrisduerr:
> For the record I have tried packaging rust binary packages before and due to fedoras policy of only having one version of a lib at a time it can actually be pretty hard as for example some sub dependencies might rely on several different versions of popular libraries like lazy_static etc. But I'd definitely give this one a try and suggest the following patches if required.

This will likely be near impossible with any reasonably sized Rust application, including Alacritty. There have been some people helping out greatly with this, however completely removing all duplicates for every release seems unlikely to ever be possible. Rust's version management just isn't really built to force this kind of restriction.

With that in mind, it probably makes sense to move forward with this PR then. We'll get this building as an unofficial package first and hopefully Fedora can figure something out to reliably package Rust applications.

alacritty.spec.rpkg
```diff
@@ -0,0 +1,56 @@
+Name:          {{{ git_name }}}
+Version:       {{{ git_version }}}
+Release:       1%{?dist}
+Summary:       A cross-platform, GPU enhanced terminal emulator
+License:       ASL 2.0
```
chrisduerr: I'm not sure what the ASL 2.0 license is, but I don't think it's the Apache License which Alacritty is licensed under.

cole-h: "ASL" appears to stand for "Apache Software License" -- see https://wiki.debian.org/DFSGLicenses#The_Apache_Software_License_.28ASL.29 (one of the few references to that nomenclature).

chrisduerr: Huh, interesting. I usually only see it shortened to Apache-2.0 and that seems to be what most websites refer to it as.

Though if Debian says that's good, this is fine by me!

cole-h: SPDX agrees with you (and "most websites"), so it may be better to use that anyways. See the "Short Identifier" at https://spdx.org/licenses/Apache-2.0.html

chrisduerr: Yeah I'd say if @pschyska has picked this for a specific reason, it's fine. But if this isn't a Fedora requirement the Apache-2.0 name is probably better.

pschyska: I think they prefer ASL, see https://fedoraproject.org/wiki/Licensing:Main. That files is linked from COPR because you are only allowed to build software with licenses mentioned in that wiki there.

pschyska:
As I wrote I was unable to move the files into extra/linux/, at least not rpkg.macros as the tool expects it in the project root...
I can try something else tomorrow.

I can't build a 0.3.2 manually without modification, as it requires f2b1bfdf37aea62e79b1791127eaf1507edfecd9 to build. The builds will work in COPR after the next tag.
If you just want to test if the package works, you can also check https://copr.fedorainfracloud.org/coprs/pschyska/alacritty/. The latest version there is built from this branch. (`sudo dnf copr enable pschyska/alacritty && sudo dnf install alacritty`)

The version number encodes the HEAD via `git-describe`: 
`<most recent tag>.<commits between tag and HEAD>.<committish>-<package release>.<series>`
e.g. `0.3.2.44.g489552d-1.fc30` (package release is hard-coded in the spec and should change when just the spec changes but no other code).
For 0.3.3 the package name will be `0.3.3-1.fc30`, etc.

(The COPR repo is not public yet, use at your own risk :-). We could enable automatic builds on every push to master to build "pre-release" packages, but I think these should live in another COPR then, as otherwise people on stable versions would auto-upgrade to non-tagged builds. I'd suggest `pschyska/alacritty-git` or similar. I will remove the pre-release packages from `pschyska/alacritty` as soon as I can build a stable 0.3.3.

chrisduerr:
> I can't build a 0.3.2 manually without modification, as it requires f2b1bfd to build. The builds will work in COPR after the next tag.

I'll likely release a release candidate for the next version soon, that should hopefully allow doing a test build, right?

> (The COPR repo is not public yet, use at your own risk :-). We could enable automatic builds on every push to master to build "pre-release" packages, but I think these should live in another COPR then, as otherwise people on stable versions would auto-upgrade to non-tagged builds. I'd suggest pschyska/alacritty-git or similar. I will remove the pre-release packages from pschyska/alacritty as soon as I can build a stable 0.3.3.

I'm generally in favor of mainly supporting the last release. If a git release is desired, that can always be added later separately.

> As I wrote I was unable to move the files into extra/linux/, at least not rpkg.macros as the tool expects it in the project root...

My bad, seems like I missed that in the original post. That's definitely not optimal, but if there's no way around it that might be fine. Though you said that the only thing required to be in the root is the macros file, right? So could we move this to extra by manually specifying the version? I'm fine with bumping the version myself since I have to do that for several files anyways.

hargoniX:
So I've just tried looking at packaging this possible night mare for Fedora according to the standards and it's basically like @chrisduerr predicted....I might ask on the Fedora mailing list for guidance on this but it doesn't seem like any bigger rust project is gonna be packagable according to the standards soon so good luck with your copr repo @pschyska

pschyska:
> I'll likely release a release candidate for the next version soon, that should hopefully allow doing a test build, right?

Correct.
> So could we move this to extra by manually specifying the version? I'm fine with bumping the version myself since I have to do that for several files anyways.

I'm questioning the value of `rpkg` right now. It is essentially just a bunch of macros that try to make it easier for git-based projects, but we can't use their version and tagging scheme, as they expect tag names to be name-major-minor (they have a similar way of conveying the HEAD state like git-describe). They generate Changelogs automatically, but they pull them from the commit messages of tags that are made via their scheme. They pre-process the spec.template with bash, although rpm has a sophisticated macro system itself. Honestly, it's probably easier to just commit a plain .spec file and maintain that.

So COPR has 3 build modes: rpkg, tito, and make srpm. Tito is another tool like rpkg, which might be more flexible, but it also seems to expect a release process focused around building rpms, i.e. making tags with the tool itself. The last option just expects a `srpm` target in the Makefile, and we can do anything we want in there. As we already have a Makefile it's probably better to commit just the spec file and add the srpm target. Then we can do anything we want without tool constraints :-).

I also saw that Fedora has a set of macros called [rust-packaging](https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/), maybe there is something we can use to use cargo metadata instead of hard coding things into the specfile. I'll invest an hour or two to check that out.

alacritty.spec.rpkg
nixpulvis: Just double checked that a quick google for "ASL 2.0" didn't give me some revision of the **A**merican **S**ign **L**anguage. :rofl:

chrisduerr:
> Honestly, it's probably easier to just commit a plain .spec file and maintain that.

Would you be willing to contribute one? I'm fine with maintaining it if it just requires updating some version number, however I don't have any Fedora system to write or test it, so help in that area would be much appreciated.

I personally don't really care either way, since I'm not really familiar with Fedora build things, but going closer to the target by removing useless abstractions is usually a good thing.

pschyska:
> Would you be willing to contribute one?

Yes, of course! I'm working on it. Taking longer than expected :smirk:

pschyska:
@chrisduerr Can you tell me if `mesa-libEGL` should be a dependency? Or `mesa-libEGL-dev`?
I have `mesa-libEGL` on my system and I can't remove it, so I can't really check if alacritty would work without it. There was a note in the installation instruction that nvidia-users might want it. My machine has an intel chip but I still have it on a default install. I'm thinking it's probably correct to add it as `Requires:`?

I'm pretty sure it's not a `BuildRequires`, and we don't need `mesa-libEGL-dev` to build, as I never had that. Unless alacritty would build differently with `mesa-libEGL-dev` installed, but I didn't find a reference in the project that would indicate that.

chrisduerr:
@pschyska EGL is required when running Wayland. On X11 it's just a fallback and GLX is used by defaualt.

Not sure how Fedora differentiates between Wayland/X11, but that probably means it should always be required.

pschyska:
But that means it's not a direct dependency, but a dependency on running Wayland at all if I understand correctly? Cursory search on the internet seems to imply that the proprietary nvidia drivers install their own conflicting version of this library. If that is true, and we depend on it, nvidia user wouldn't be able to install alacritty.

mesa-libEGL is required by libglvnd which is required by gnome-session, etc.
I expect it to be available wherever any wayland compositor is installed, and we should *not* depend on it.

pschyska:
I went deep in the rabbit hole of trying to make the defined version more flexible, and derive it from the source tree (i.e. parsing `cargo metadata`), but failed. I can only control the building of the SRPM to some extend in COPR, and the version number generation logic failed during the build of the actual RPM. I have some other ideas but for now I think it's fine to leave the version hardcoded.
It needs to be bumped when a new Version is tagged and can't contain `-`. [Fedora has a notion of pre-release packages](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_prerelease_versions) which we probably should obey.
For tagging `0.3.3-rc1` I recommend the following:
```
Version:       0.3.2
Release:       0.1%{?dist}
```
`0.3.3-rc2` would be
```
Version:       0.3.2
Release:       0.1%{?dist}
```

I'm personally disappointed because I planned to set up a private COPR that builds every revision of master for testing, and that's not easily possible now, but I might figure it out eventually and open a new PR for that :-).

As for the file paths, there is a `.copr` now in the project root although the spec file could move. If we use the third build options, we would have a `.tito` in the project root, so there doesn't seem a way around it...

chrisduerr:
> But that means it's not a direct dependency, but a dependency on running Wayland at all if I understand correctly? Cursory search on the internet seems to imply that the proprietary nvidia drivers install their own conflicting version of this library. If that is true, and we depend on it, nvidia user wouldn't be able to install alacritty.

Alacritty just requires any EGL library, it doesn't have to be mesa's. I don't know how this works in Fedora, but there's usually standards about what dependencies need to be included and which can be assumed to exist (like the linux kernel for example).

I know that with Archlinux, there exists a concept of a 'provider' for certain things. So both the proprietary nvidia driver and mesa provide 'opengl-dri', so a package would only have to depend on 'opengl-dri' and the user could then decide which package he wants to install that provides this.

Generally it's not a direct dependency, in the way that not every Alacritty instance requires it. However every Alacritty instance requires some form of GPU drivers for graphics to work and on linux either a working X11 or Wayland server.

You can take a look at the Archlinux package dependencies as a reference, though I can't say anything about Fedora standards:
https://www.archlinux.org/packages/community/x86_64/alacritty/

> As for the file paths, there is a .copr now in the project root although the spec file could move. If we use the third build options, we would have a .tito in the project root, so there doesn't seem a way around it...

I think that should be fine. There's also a .travis.yml and similar stuff, for things that require to work based on the directory root.

However, I am curious how COPR polls this. Does it require manual interaction to build or will it just build whenever the file is changed? Usually only tagged versions are actual releases, so even though it shouldn't matter, one might not want to directly build on every change to master.

pschyska:
> Generally it's not a direct dependency, in the way that not every Alacritty instance requires it. However every Alacritty instance requires some form of GPU drivers for graphics to work and on linux either a working X11 or Wayland server.

I don't know for sure how to handle that. I would rather leave the dependency out than to run with it and possible break the package for nvidia users (as in: making it not installable). I would expect everyone trying to run alacritty to have a working graphical environment though 😏
I'll try figure it out and amend the dependencies in a follow-up PR.

> You can take a look at the Archlinux package dependencies as a reference, though I can't say anything about Fedora standards:
> https://www.archlinux.org/packages/community/x86_64/alacritty/

I don't see any dependency on any GL libs, and/or wayland libs there. Just some supplemental X input libs.
Am I blind? Unless libEGL is pulled in by one of these...

> However, I am curious how COPR polls this. Does it require manual interaction to build or will it just build whenever the file is changed? Usually only tagged versions are actual releases, so even though it shouldn't matter, one might not want to directly build on every change to master.

They support webhooks, but you'd need custom scripts to unpack the json to figure out if the webhook is about a newly-pushed HEAD in refs/tags, for example. By default they just trigger a build everytime the webhook is invoced, and you can only specify a single commitish to build like `master`. AFAICS they don't support something like `build refs/tags/*` ☹️.

As soon as 0.3.3(-rc) is tagged, I would push the button and manually insert `refs/tags/v0.3.3-rc1` into the form and that version would be built. For now. And then remove the private flag from the COPR and send a PR on README.md that mentions how to install it from there. Hey, it's a start, right?

chrisduerr:
> I don't see any dependency on any GL libs, and/or wayland libs there. Just some supplemental X input libs.
Am I blind? Unless libEGL is pulled in by one of these...

That's the point I was trying to make. Archlinux does not add these as dependencies.

> They support webhooks, but you'd need custom scripts to unpack the json to figure out if the webhook is about a newly-pushed HEAD in refs/tags, for example. By default they just trigger a build everytime the webhook is invoked, and you can only specify a single commitish to build like master. AFAICS they don't support something like build refs/tags/* frowning_face. It would also be possible to ping COPR from travis, or something similar.

It should be fine for now, though that's something we might want to keep in mind. Triggering from travis sounds like a decent idea.

pschyska:
> That's the point I was trying to make. Archlinux does not add these as dependencies.

Alright sorry, I misunderstood that.

Let me know if you think there is something else to do before this can be merged. Thanks!

chrisduerr:
@pschyska I've got another question with the pre-release stuff. Will people have to opt-into that or would 3.3/0.1 automatically get installed for someone on 3.2/1.0?

pschyska:
I doubt it, but I will test tomorrow.

pschyska:
@chrisduerr I have researched Fedora versioning guidelines, and how rpm (or rather dnf) will behave.
For reference: [Versioning guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/), [Examples](https://fedoraproject.org/wiki/Package_Versioning_Examples).
The traditional way to release a RC would be to "abuse" the Release: tag to signify that this is a pre-release, e.g.:

| Alacritty   | Version | Release   | RPM             |
|-------------|---------|-----------|-----------------|
| 0.3.3-beta1 | 0.3.3   | 0.1.beta1 | 0.3.3-0.1.beta1 |
| 0.3.3-rc1   | 0.3.3   | 0.2.rc1   | 0.3.3-0.2.rc1   |
| 0.3.3-rc2   | 0.3.3   | 0.3.rc2   | 0.3.3-0.3.rc2   |
| 0.3.3       | 0.3.3   | 1         | 0.3.3-1         |

However, I don't like that, as the 0.N in Release has to be manually bumped every time, and for that you need to know what the last version used, and there is no relation of the integer to the upstream version.
There is a new proposal that allows the use of the `~` in the Version: which makes it always sort lower than anything else with the same prefix. (See this [discussion](https://pagure.io/packaging-committee/issue/398)).
In that scheme, the `Release:` would be used for versions where no code except the specfile itself changed (like it was originally intended, AFAICS). So it would normally stay at `Release: 1` for all versions. E.g:

| Alacritty   | Version     | Release | RPM           |
|-------------|-------------|---------|---------------|
| 0.3.3~beta1 | 0.3.3~beta1 | 1       | 0.3.3~beta1-1 |
| 0.3.3-rc1   | 0.3.3-rc1   | 1       | 0.3.3-rc1-1   |
| 0.3.3-rc2   | 0.3.3-rc2   | 1       | 0.3.3-rc2-1   |
| 0.3.3       | 0.3.3       | 1       | 0.3.3-1       |

This is much more sane IMHO and I think we should use that scheme. There is a pathological case where the "prerelease name" is not alphabetically sorting. Imagine there is now a release called "pre" which should come *after* "rc2". In that case its Version must be changed from the upstream name to make it sort again, i.e. by introducing a `1.` prefix:

| Alacritty   | Version     | Release | RPM           |
|-------------|-------------|---------|---------------|
| 0.3.3~beta1 | 0.3.3~beta1 | 1       | 0.3.3~beta1-1 |
| 0.3.3-rc1   | 0.3.3-rc1   | 1       | 0.3.3-rc1-1   |
| 0.3.3-rc2   | 0.3.3-rc2   | 1       | 0.3.3-rc2-1   |
| 0.3.3-pre   | 0.3.3-1.pre   | 1       | 0.3.3-1.pre-1   |
| 0.3.3       | 0.3.3       | 1       | 0.3.3-1       |

So it's getting ugly again (package version diverging from upstream), but at least this situation goes away when 0.3.3 is released and is only necessary for this one version. I also don't expect this to ever happen in practice :smirk: 

The bad news is that dnf would the upgrade from `0.3.2` to `0.3.3-rc1` regardless of the way we use. So that also means I *must* have separate release and prerelease COPRs.

tldr; I'm updating my recommendation on how to tag the upcoming 0.3.3-rc1:
```
Version:       0.3.3~rc1
Release:       1%{?dist}
```
@hargoniX Can you chime in if that makes sense, as you are probably more experienced with the guidelines than I am?

chrisduerr:
> The bad news is that dnf would the upgrade from 0.3.2 to 0.3.3-rc1 regardless of the way we use. So that also means I must have separate release and prerelease COPRs.

In that case it seems like the best solution is to just not bump the fedora copr build for RCs at all and instead just update it when versions are released.

pschyska:
> In that case it seems like the best solution is to just not bump the fedora copr build for RCs at all and instead just update it when versions are released.

But then I get no chance to test if packages will build correctly for RC's, which I'd like to do. If a packaging issue comes up (i.e. new dependency), I would only be able to correct it for the next release tag...
Please just bump the version in specfile like I described when you tag RC1, I will deal with it. We don't have to mention the COPR repo containing only RC's anywhere, but it would make my life much easier.
pschyska/alacritty will always only contain releases in any case.
Cheers

pschyska:
But, I would argue will still should offer pschyska/alacritty-pre as a COPR so we don't lose Fedora users that don't want to manually build as potential prerelease testers.

chrisduerr:
Okay, so for version 0.3.3 I will tag a pre-release since none exists yet, then for future releases we can just skip the pre-release and go straight to the versioned releases.

I'm not concerned with pre-releases at this point since people can just build from source, that shouldn't be much of a problem for people interested in testing especially since Rust projects are so straight-forward to build.

If I understand everything correctly, COPR will be watching only the master branch and build whenever I change the version in the spec file, right?

pschyska:
> If I understand everything correctly, COPR will be watching only the master branch and build whenever I change the version in the spec file, right?

No, I will trigger the builds manually for now. You can ping me when you make a new git tag, but I will regularly check anyways.

> Okay, so for version 0.3.3 I will tag a pre-release since none exists yet, then for future releases we can just skip the pre-release and go straight to the versioned releases.

Alright, let's say whenever you tag a RC, you bump the specfile accordingly. If you are not tagging anything in git anyways, there is no need to build RC rpms, obviously. I just saw that there used to be RC's, and you mentioned it, so I thought you might want to use prereleases more.

I'm fine either way, I just want to avoid that there is a 0.3.3-rc1 tagged but the specfile still says 0.3.2, and then someone wants a 0.3.3~rc1 rpm package to test something and I can't build it without hand-patching the specfile.

chrisduerr:
> I'm fine either way, I just want to avoid that there is a 0.3.3-rc1 tagged but the specfile still says 0.3.2, and then someone wants a 0.3.3~rc1 rpm package to test something and I can't build it without hand-patching the specfile.

That's exactly what's going to happen though. If a RC is going to overwrite the existing release, I'm not going to tag RCs for the Fedora build. RCs should ***not*** get published to package repositories. If you want to test a RC, you can always build from source.

It completely ruins the point of RCs when everyone gets upgraded to them, since that just makes them equivalent to normal releases and everything breaks for everyone. It's meant as a best-effort to catch some bugs before releasing to the world.

extra/linux/redhat/alacritty.spec
```diff
@@ -0,0 +1,56 @@
+Name:          alacritty
+Version:       0.3.2
+Release:       1%{?dist}
+Summary:       A cross-platform, GPU enhanced terminal emulator
+License:       ASL 2.0
+URL:           https://github.com/jwilm/alacritty
+VCS:           https://github.com/jwilm/alacritty.git
+Source:        alacritty-%{version}.tar
+
+BuildRequires: rust >= 1.31.0
```
chrisduerr: This needs to be changed to 1.32.0, since I've just updated the minimum supported rust version.

pschyska:
> That's exactly what's going to happen though. If a RC is going to overwrite the existing release, I'm not going to tag RCs for the Fedora build. RCs should _**not**_ get published to package repositories. If you want to test a RC, you can always build from source.
> 
> It completely ruins the point of RCs when everyone gets upgraded to them, since that just makes them equivalent to normal releases and everything breaks for everyone. It's meant as a best-effort to catch some bugs before releasing to the world.

But that's not going to happen, as it's in a different repository. Every COPR has their own rpm repo, so if you never enabled, let's say pschyska/alacritty-pre, in addition to pschyska/alacritty you would not get pre-releases.

(If you are like me, you would have enabled it and get the RC's as soon as they come out.
I don't mind occassional breakage as it's easy to recover, and I like when I'm able to report meaningful bugs if I can.)
But again, enabling alacritty-pre would be a deliberate choice, with a warning attached.

chrisduerr:
> The bad news is that dnf would the upgrade from 0.3.2 to 0.3.3-rc1 regardless of the way we use. So that also means I must have separate release and prerelease COPRs.

Based on this I thought that would mean this wouldn't be a reasonable approach, but if you're willing to maintain them separately and it's an opt-in for the pre-releases, that's fine by me.

I just don't want people that install the regular one to suddenly run RCs. But it sounds like we should be good to go.

extra/linux/redhat/alacritty.spec
pschyska: Done, currently build new packages for testing

pschyska:
Unsurprisingly, the package still built with the bumped rust dependency :smile: 
Was there anything else?

chrisduerr:
@pschyska I think everything should be tagged properly now for 0.3.3-rc1, let me know if there are any issues.

pschyska:
@chrisduerr For some reason, my e-mail reply didn't get posted. I went ahead and create 2 public repos (stable and testing). You can check them out here:
[alacritty](https://copr.fedorainfracloud.org/coprs/pschyska/alacritty/)
[alacritty-testing](https://copr.fedorainfracloud.org/coprs/pschyska/alacritty-testing/)

I worded alacritty-testing differently, and removed the detailed installation instructions to dissuade users that might not want to test pre-releases from using that one, I hope that works for you :smirk:.

The prerelease build is currently running. When you tag 0.3.3 stable, I will send you a pull request on README.md.

chrisduerr:
@pschyska Version 0.3.3 is released now, hopefully everything should work out.

pschyska:
@chrisduerr Thanks, the build was successful: https://copr.fedorainfracloud.org/coprs/pschyska/alacritty/build/936407/
I'll do a quick check and then send a PR for README.md.

PR #2507, opened 2019-06-04
Dynamic color code escapes for foreground/background

kev-zheng:
Saw https://github.com/jwilm/alacritty/issues/2497 and decided to take a crack at it

Recursively causes os dispatch to parse the next
parameters, as if they were entered with their
own escape codes

Not too familiar with rust but seems like this 
would be tail recursive

before:
Screen Shot 2019-06-04 at 4 56 34 PM

after:
Screen Shot 2019-06-04 at 4 56 39 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -389,6 +389,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Pressing enter on the numpad should now insert a newline
 - The mouse bindings now support keyboard modifiers (shift/ctrl/alt/super)
 - Add support for the bright foreground color
+- Support for setting foreground, background colors in one escape sequence
 
 ### Changed
 
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -801,7 +801,7 @@ where
                             // 10 is the first dynamic color, also the foreground
                             let offset = dynamic_code as usize - 10;
                             let index = NamedColor::Foreground as usize + offset;
-                            
+
                             // End of setting dynamic colors
                             if index > NamedColor::Cursor as usize {
                                 unhandled(params);
@@ -811,11 +811,7 @@ where
                             if let Some(color) = parse_rgb_color(param) {
                                 self.handler.set_color(index, color);
                             } else if param == b"?" {
-                                self.handler.dynamic_color_sequence(
-                                    writer,
-                                    dynamic_code,
-                                    index,
-                                );
+                                self.handler.dynamic_color_sequence(writer, dynamic_code, index);
                             } else {
                                 unhandled(params);
                             }
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -537,10 +537,10 @@ pub enum NamedColor {
     Foreground = 256,
     /// The background color
     Background,
-    /// Color for the text under the cursor
-    CursorText,
     /// Color for the cursor itself
     Cursor,
+    /// Color for the text under the cursor
+    CursorText,
     /// Dim black
     DimBlack,
     /// Dim red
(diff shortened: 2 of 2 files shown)


kev-zheng:
Got it, thanks for the helpful explanation. Checked out some behaviors on xterm; this revision should have similar behavior.

Given a starting escape code, you can get (via ?) or set (via color); this uses the ordering in NamedColor. I added cursor support since that case was similar; had to switch the order of Cursor and CursorText.

since only fg, bg, and cursor were available and NamedColors started to go into the dim colors, there is an explicit check to end set/get at the Cursor value.

Examples:

querying fg, bg
```
bash-3.2$ echo -e '\033]10;?;?\007'

bash-3.2$ 10;rgb:c5/c8/c610;rgb:31/35/38
```

set fg, get bg, verify fg is different
```
bash-3.2$ echo -e '\033]10;#4eeeee;?\007'

bash-3.2$ 10;rgb:31/35/38

bash-3.2$ echo -e '\033]10;?\007'

bash-3.2$ 10;rgb:4e/ee/ee
```

on xterm, malformed colors were just skipped, similar behavior:
skip fg, get bg
```
bash-3.2$ echo -e '\033]10;blah;?\007'

bash-3.2$ 10;rgb:31/35/38
```

setting cursor...
```
bash-3.2$ echo -e '\033]12;?\007'
bash-3.2$ 12;rgb:0/0/0
bash-3.2$ echo -e '\033]12;#4eeeee\007'
bash-3.2$ echo -e '\033]12;?\007'
bash-3.2$ 12;rgb:4e/ee/ee
```

all three...
```
bash-3.2$ echo -e '\033]10;?;?;?\007'
bash-3.2$ 10;rgb:c5/c8/c610;rgb:31/35/3810;rgb:4e/ee/ee
bash-3.2$ echo -e '\033]10;#ffffff;#000000;#000000\007'
```

kev-zheng:
Those commands work as it is right now on my end (bash in OSX) - what are you using where it doesn't work?

```
# Start with white foreground, black background
printf '\e]10;rgb:ff/ff/ff;rgb:00/00/00\007'

# Should print escape for setting first color and set the second
printf '\e]10;?;rgb:ff/00/00\007'   # Sets background to red
^[]10;rgb:ff/ff/ff^G       # Returns escape code for foreground

# Reset
printf '\e]10;rgb:ff/ff/ff;rgb:00/00/00\007'

# Should print escape for setting second color and set the first
printf '\e]10;rgb:ff/00/00;?\007'         # Sets foreground to red
^[]10;rgb:0/0/0^G      # Escape code for background

# Reset (also set third color)
printf '\e]10;rgb:ff/ff/ff;rgb:00/00/00;rgb:ee/ee/ee\007'

# Should print escape for setting first and third color and set the second
printf '\e]10;?;rgb:ff/00/00;?\007'
^[]10;rgb:ff/ff/ff^G^[]10;rgb:ee/ee/ee^G # Escapes for first and third color
```

used this script to capture output with character representations of escape/bell codes
```
#!/bin/sh
oldstty=$(stty -g)

# What to query?
query=$1

stty raw -echo min 0 time 0
printf "$1"

sleep 0.00000001

read -r answer
echo $answer | cat -v
stty $oldstty
```

chrisduerr:
Take a look at the output of your commands:

```
# Should print escape for setting second color and set the first
printf '\e]10;rgb:ff/00/00;?\007'         # Sets foreground to red
^[]10;rgb:0/0/0^G      # Escape code for background
```

This should start with the escape '\e]11', not '\e]10'. Your example would just return the escape for setting the first color again.

Same here, the escapes should start with 10 and 12:

```

# Should print escape for setting first and third color and set the second
printf '\e]10;?;rgb:ff/00/00;?\007'
^[]10;rgb:ff/ff/ff^G^[]10;rgb:ee/ee/ee^G # Escapes for first and third color
```

kev-zheng:
Ah, thanks. This should work now
```
printf '\e]10;rgb:ff/00/00;?\007'
^[]11;rgb:0/0/0^G                          # Second color

printf '\e]10;rgb:ff/00/00;?\007'
^[]10;rgb:ff/ff/ff^G^[]12;rgb:0/0/0^G  # First, third color
```

Issue #2457, opened 2019-05-17
Key binding `mode`s need docs


OJFord:
The readme/wiki/sample config does not currently explain what `AppCursor`, `AppKeypad`, `Alt`, or the `~` modify mean.

Additionally, the wiki points to the sample config for an explanation, but uses line numbers on master, which have clearly changed since it was written. (If it's going to use line numbers, to highlight the relevant section, then it should point at a specific commit, not master.)

The sample config does not include `Alt`/`~Alt` in available values, though it does use them in a couple of examples.

chrisduerr:
Thanks for the issue. Just to be clear these are terminal  modes like the alt screen mode (`Alt`) which can be used to restrict certain keys to only the normal or alternate screen buffer. `~` matches whenever a mode is **not** active.

Documenting this properly certainly makes sense and of course all valid values should be presented. Though this is likely useless for users and primarily used to make the default bindings work properly.

OJFord:
Thanks, I think `~` is fairly self-evident (a comment would be friendly though) - I was just struggling to see which I needed and eventually found it as Alt in an issue. (The nane makes perfect sense, once you know about 'alt screen'!)

I'd be happy to write something that I think would've helped, but I confess I still don't know what the other two are for!

chrisduerr:
> I confess I still don't know what the other two are for!

There's an escape sequence to switch the Cursor Keys and Keypad Keys between normal and application mode. In these modes the escape sequences sent are different. That's what they are for, since we need to send other escape sequences to comply with the official standard.

nixpulvis:
I've at least started by fixing the Wiki link to be a perma link to the correct place for now.

I'd be interested in making this more clear in the default config. I'll have a PR ready soon, possibly with some questions.

PR #2509, opened 2019-06-05
Expand/Refactor Config Key Bindings Docs

nixpulvis:
This is in response to #2457, and should fully cover the `key_bindings` configuration.

- [x] Document `mode` values

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -280,58 +280,6 @@ visual_bell:
 # The value `0.0` is completely transparent and `1.0` is opaque.
 background_opacity: 1.0
 
-# Mouse bindings
-#
-# Available fields:
-#   - mouse
-#   - action
-#   - mods (optional)
-#
-# Values for `mouse`:
-#   - Middle
-#   - Left
-#   - Right
-#   - Numeric identifier such as `5`
-#
-# All available `mods` and `action` values are documented in the key binding
-# section.
-mouse_bindings:
-  - { mouse: Middle, action: PasteSelection }
-
-mouse:
-  # Click settings
-  #
-  # The `double_click` and `triple_click` settings control the time
-  # alacritty should wait for accepting multiple clicks as one double
-  # or triple click.
-  double_click: { threshold: 300 }
-  triple_click: { threshold: 300 }
-
-  # If this is `true`, the cursor is temporarily hidden when typing.
-  hide_when_typing: false
-
-  url:
-    # URL launcher
-    #
-    # This program is executed when clicking on a text which is recognized as a URL.
-    # The URL is always added to the command as the last parameter.
-    #
-    # When set to `None`, URL launching will be disabled completely.
-    #
-    # Default:
-    #   - (macOS) open
-    #   - (Linux) xdg-open
-    #   - (Windows) explorer
-    #launcher:
-    #  program: xdg-open
-    #  args: []
-
-    # URL modifiers
-    #
-    # These are the modifiers that need to be held down for opening URLs when clicking
-    # on them. The available modifiers are documented in the key binding section.
-    modifiers: None
-
 selection:
   semantic_escape_chars: ",│`|:\"' ()[]{}<>"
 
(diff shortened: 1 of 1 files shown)


cole-h:
A suggestion, since we're already dealing with key bindings docs: could we also move the `mouse_bindings` section below the `key_bindings` section (or `key_bindings` above `mouse_bindings`)?

I say this because the mouse bindings docs/comments state "All available `mods` and `action` values are documented in the key binding section." However, the key binding section is at the bottom of the file. To me, it feels very cumbersome scrolling 200+ lines to read possible values.

chrisduerr:
I'm not against this, though right now the mouse binding docs are intentionally placed next to the mouse docs.

So if we decide to do that, I'd recommend reordering all of them, so we'd have `mouse` section first, then `mouse_bindings`, which references the `key_bindings` section, which then just follows until the end of the file.

alacritty.yml
```diff
@@ -486,26 +474,47 @@ debug:
...
+#    - Option
+#    - Super
+#    - Shift
+#    - Alt
+#
+#    Multiple `mods` can be combined using `|` like this:
+#       `mods: Control|Shift`.
+#    Whitespace and capitalization are relevant and must match the example.
+#
+# - `mode`: ...
```
nixpulvis: I'm struggling a bit to fill in these last `...`s. I can see the complete list (embeded below), but only `ALT_SCREEN`, `APP_KEYPAD`, and `APP_CURSOR` are exposed through the config.

https://github.com/jwilm/alacritty/blob/55d346c7d44215c51187cafcac79e3504209f942/alacritty_terminal/src/term/mod.rs#L489-L508

However for the config values we only need to document `AppCursor`, `AppKeypad`, and `Alt`.

chrisduerr: > However for the config values we only need to document AppCursor, AppKeypad, and Alt.

That's correct, only the values relevant for keybindings are available, so only those need to be documented. That shouldn't be a problem, right?

nixpulvis: Ah, if I'm understanding this correctly, we just reexport the mode through `AppCursor`, since `App` is appropriate for a Digital Equipment Corporation private control sequence :trollface:. And there are differences in the escape sequences between modes. I'd never dug into this before. How strange!

It would be really nice if we could abstract this away from the user config altogether. But that's another story.


```
ansi::Mode::CursorKeys => self.mode.insert(TermMode::APP_CURSOR),
```

> DECCKM – Cursor Keys Mode (DEC Private)
>
> This is a private parameter applicable to set mode (SM) and reset mode (RM) control sequences. This mode is only effective when the terminal is in keypad application mode (see DECKPAM) and the ANSI/VT52 mode (DECANM) is set (see DECANM). Under these conditions, if the cursor key mode is reset, the four cursor function keys will send ANSI cursor control commands. If cursor key mode is set, the four cursor function keys will send application functions.

chrisduerr: If we abstract this away from user config, the issue is that the user doesn't have full control about the special bindings anymore.

Sure we could set all our default bindings to never get overwritten at all, but I don't think that would actually benefit the user at all.

And the app keypad/app cursor modes are actually quite common.

alacritty.yml
```diff
@@ -486,26 +474,50 @@ debug:
...
+#
+#    - Command
+#    - Control
+#    - Option
+#    - Super
+#    - Shift
+#    - Alt
+#
+#    Multiple `mods` can be combined using `|` like this: `mods:
+#    Control|Shift`. Whitespace and capitalization are relevant and must match
```
chrisduerr: Linebreak here for the `mods: Control|Shift` code blocks seems a bit odd and I don't think it goes beyond maximum line length? Would probably best to not split it apart if not necessary.

nixpulvis: I literally had this correct before, and lazily formatted it with `gq`.

mjiricka:
Hi, I got here from #2457 when I was wondering what are `AppCursor`, `AppKeypad` and `Alt` modes. I still don't understand the first two, probably because I am not familiar with unix terminal specs etc. But I find confusing there is "Alt mode" and "Alt mod", each meaning something completely different. I would suggest to notice that "Alt mod" means "Alternate screen buffer" in example config file.

PR #2511, opened 2019-06-05
Fix index out of bounds during selection

### LINKED ISSUE #2486: Index out of bounds when selecting during `yes`
When creating a selection while running the `yes` command, Alacritty will crash with an index out of bounds exception.

chrisduerr:
This reworks the selection logic to prevent any possible index out of
bounds exceptions by clamping the start and end points before doing
anything else with them when converting selections to spans.

This also fixes a bug where semantic selections would not automatically
expand across double-width characters.

Fixes #2486.

### DIFF
--- a/alacritty_terminal/src/index.rs
+++ b/alacritty_terminal/src/index.rs
@@ -71,12 +71,6 @@ impl From<Point> for Point<usize> {
     }
 }
 
-impl From<Point<usize>> for Point {
-    fn from(point: Point<usize>) -> Self {
-        Point::new(Line(point.line), point.col)
-    }
-}
-
 impl From<&RenderableCell> for Point<Line> {
     fn from(cell: &RenderableCell) -> Self {
         Point::new(cell.line, cell.column)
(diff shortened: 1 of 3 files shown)


cole-h:
Confirmed fixed for me, using both `yes` and `while true; do echo no; done`. Just as I was starting to look into it :sweat_smile: 

Good work!

chrisduerr:
@cole-h Hah, thanks for testing. If you found alternative solutions that might be better, feel free to let me know.

Though I think this is pretty good already, since I've chosen to refactor a fair bit and clean stuff up. Hopefully that makes everything a little more readable now too.

cole-h:
Your solution looks much more elegant than what I had going (which wasn't much; only recently found time to try and work on it). The refactor is a nice bonus.

I also like that sneaky typo-fix at https://github.com/jwilm/alacritty/pull/2511/files#diff-7e09fea17462febd06d3b5bdd2b4512aR444 :wink:

PR #2438, opened 2019-05-13
Update to winit/glutin EventLoop 2.0

chrisduerr:
This takes the latest glutin master to port Alacritty to the EventLoop
2.0 rework.

This changes a big part of the event loop handling by pushing the event
loop in a separate thread from the renderer and running both in
parallel.

TODO:
 - [x] Resize issues (https://github.com/jwilm/alacritty/pull/2438#issuecomment-505518232)
 - [x] `run_return` support for macOS (https://github.com/rust-windowing/winit/issues/1005)
 - [x] Resizing extremely slow
 - [x] Font size change not updating pty size
 - [x] Message bar does not update PTY size properly
 - [x] Incorrect window dimensions when moving Alacritty between different DPI monitors
 - [x] Font size not updating when moving Alacritty between different DPI monitors
 - [x] Clean up code / resolve remaining TODOs
 - [x] Incorrect startup DPI on X11 (https://github.com/rust-windowing/winit/issues/998)
    -> ~~:zap: Still broken when spawning a floating window in i3~~
    -> Workaround implemented
 - [x] Make sure before merging all the linked issues have change log entries
 - [x] Clear errors on message bar reload

Fixes #2796.
Fixes #2694.
Fixes #2643.
Fixes #2625.
Fixes #2618.
Fixes #2601.
Fixes #2564.
Fixes #2456.
Fixes #2438.
Fixes #2334.
Fixes #2254.
Fixes #2217.
Fixes #1789.
Fixes #1750.
Fixes #1125.

PR #2517, opened 2019-06-07
fix compiler warnings with latest nightly

matthiaskrgr:


matthiaskrgr:
Mmmh,  `pre_exec()` does not exist on rust 1.31...

chrisduerr:
I think the #2438 PR should fix all warnings in a Rust 1.31.0 compatible way.

matthiaskrgr:
Was failing to build with 1.32 as well, I removed the before_exec/pre_exec changes from the pr.

Issue #1071, opened 2018-01-31
Default shell on macOS should use `login`


JamesDunne:
Which operating system does the issue occur on?
MacOS 10.12.3

Default process for a shell on MacOS should use `login -fp $USERNAME` and not a direct invocation of `/bin/bash` because the user's environment gets set up differently with `login`. iTerm2.app and Console.app both use `login` and share a consistent environment for my user account but alacritty's environment differs with a buggy $PATH among other oddities.

JamesDunne:
For anyone else running into this issue, add the following to your `alacritty.yml` in your config location (mine is `~/.config/alacritty/alacritty.yml`):
```
shell:
  program: login
  args: ["-fp", "USERNAMEHERE"]
```

fallwith:
Thank you for posting this. Alacritty was not lighting up my `~/.zsh/.zprofile` file, but following your example I was able to get things working with:

```shell
shell:
  program: /usr/local/bin/zsh
  args:
    - --login
```

acheronfail:
I think a good example of how this can be fixed in Alacritty can be seen in how it was fixed in kitty: https://github.com/kovidgoyal/kitty/issues/247

PR #2541, opened 2019-06-14
Launch a login shell by default on Linux

gentooboontoo:
Alacritty did not launch a login shell by default on Linux contrary to
what is claimed by the comments in the configuration file. As a result,
`bash` does not read `/etc/profile` file leading to an inconsistent
environment compared to those with other terminals.

This commit simply extends what has been done by PR #1856 referenced by
issue #1071.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -363,7 +363,8 @@ live_config_reload: true
 # Entries in `shell.args` are passed unmodified as arguments to the shell.
 #
 # Default:
-#   - (Linux/macOS) /bin/bash --login
+#   - (macOS) /bin/bash --login
+#   - (Linux) user login shell
 #   - (Windows) powershell
 #shell:
 #  program: /bin/bash
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
-
-### Fixed
-
-- Launch a login shell by default on Linux too
-
 ## 0.3.3-rc1
 
 ### Packaging
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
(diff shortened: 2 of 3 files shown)


NickeZ:
This does not seem to be the defualt behavior on `gnome-terminal`. The following should print 0 if it is a login shell.

```
$ shopt -q login_shell; echo $?
1
```

gentooboontoo:
You are right, it is different between my system and other pristine systems (I wrongly assumed it was the same).

I was misled because I always use `tmux` inside my terminals and `tmux` launch shells as login shells for new windows. Without `tmux`, my other installed terminals (`gnome-terminal`, `terminator`) also launch login shells but I should have configured them long time ago and I forgot that (so it does not match anymore their default configuration).

My questions now are:

* Should MacOS and linux `alacritty` terminals' behavior be consistent?
* If not, why is it different on MacOS? (out of curiosity, I have no Mac)
* If yes, I could reword the commit description removing the part about other terminals' behavior.
* If this PR is not the right thing to do, I could close it and make another one fixing the comment about default login shells in [`alacritty.conf` (line 366)](https://github.com/jwilm/alacritty/blob/v0.3.3-rc1/alacritty.yml#L366).

chrisduerr:
> Should MacOS and linux alacritty terminals' behavior be consistent?

No, the reason why `--login` is explicitly required on macOS is because otherwise the environment will be messed up and the locale is not set correctly. Linux does not have this issue.

> If this PR is not the right thing to do, I could close it and make another one fixing the comment about default login shells in alacritty.conf

I don't think it's necessary to open a new PR, but that seems reasonable. The current docs are probably incorrect because there has been some refactorings in the past, Alacritty will not use `bash` by default on Linux at all for example.

gentooboontoo:
> Alacritty will not use bash by default on Linux at all for example

I missed that point.

I have reverted my change and fixed the docs about default shell. I am unsure about how format the description of the linux default shell since it is not a "static" default value.

When and if OK, I will squash the commits and rename the title of the PR.

alacritty.yml
```diff
@@ -363,7 +363,8 @@ live_config_reload: true
 # Entries in `shell.args` are passed unmodified as arguments to the shell.
 #
 # Default:
-#   - (Linux/macOS) /bin/bash --login
+#   - (macOS) /bin/bash
```
chrisduerr: The --login argument was correct for macOS, it should be kept.

alacritty.yml
```diff
@@ -363,7 +363,8 @@ live_config_reload: true
...
+#   - (Linux) shell program defined in password file for current user
```
chrisduerr: I think this can be reduced to just `user login shell`, that should make things clear since it uses common terminology.

chrisduerr:
> When and if OK, I will squash the commits and rename the title of the PR.

No need to, I squash and fix the commit message of every PR when merging anyways.

gentooboontoo:
Requested changes applied.

Issue #2529, opened 2019-06-12
Build failure for 0.3.3-rc1 on OpenBSD 6.5


pwrdwnsys:
OpenBSD 6.5 (Release) / amd64. Rust 1.33.0 from OpenBSD packages. Checked out the  v0.3.3-rc1 tag.

```
error[E0425]: cannot find function `lutimes` in module `libc`
   --> /home/pwrdwnsys/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.12.1/src/sys/stat.rs:216:15
    |
216 |         libc::lutimes(cstr.as_ptr(), &times[0])
    |               ^^^^^^^
help: a function with a similar name exists
    |
216 |         libc::futimes(cstr.as_ptr(), &times[0])
    |               ^^^^^^^
help: possible candidate is found in another module, you can import it into scope
    |
1   | use sys::stat::lutimes;
    |
```
After finding nix-related problems in issue #2244, I tried those suggestions by bumping `copypasta/Cargo.toml`'s wayland-client to the latest (0.23) and also did a `cargo update`, but build still failed with the same error.

chrisduerr:
I've sent an upstream PR to update the dependencies of smithay-clipboard, so this is currently blocked on https://github.com/Smithay/smithay-clipboard/pull/5.

chrisduerr:
@pwrdwnsys Can you verify if the #2530 PR builds on BSD?

pwrdwnsys:
@chrisduerr Thank you for the rapid response. I checked out #2530 locally, the build is now failing at:
```
error[E0599]: no variant named `EPROTO` found for type `nix::errno::Errno` in the current scope
  --> /home/pwrdwnsys/.cargo/registry/src/github.com-1ecc6299db9ec823/wayland-client-0.23.3/src/native_lib/display.rs:72:40
   |
72 |         if ret == ::nix::errno::Errno::EPROTO as i32 {
   |                   ---------------------^^^^^^
   |                   |
   |                   variant not found in `nix::errno::Errno`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: Could not compile `wayland-client`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

It looks like the problem is referenced at https://github.com/rust-windowing/winit/issues/825

chrisduerr:
[...]

PR #2545, opened 2019-06-15
Add note about OpenBSD default user datasize limit

pwrdwnsys:
Ref #2529 LLVM out of memory on OpenBSD default login.conf settings


pwrdwnsys:
Thanks @chrisduerr, I have made changes.

PR #2551, opened 2019-06-16
Mention the Copr repositories in README.md

pschyska:
Add pointer to Copr in README.md for Fedora builds.


pschyska:
> Following the example of the other documentation, the `sudo` command should be removed.

Done.

PR #2552, opened 2019-06-16
Add block selection

### LINKED ISSUE #526: Rectangle selection
Most terminals support rectangular region selection. In Terminal.app this is done by holding alt, and on most X11 terminals (mostly VTE) this is done by holding ctrl.

chrisduerr:
This implements a block selection mode which can be triggered by holding
Control before starting a selection.

If text is copied using this block selection, newlines will be
automatically added to the end of the lines.

This fixes #526.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Block selection mode when Control is held while starting a selection
+
 ### Fixed
 
 - GUI programs launched by Alacritty starting in the background on X11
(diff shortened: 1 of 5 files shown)


chrisduerr:
This has been discussed before and the shortcut will not be configurable. There are just too many different modifiers that play into mouse interaction to allow for that reliably without causing trouble.

Just as an example, shift is already reserved for extension of selections in normal mode and disabling mouse mode when it is enabled. So that wouldn't be an option.

So the only choice would be between alt and control. And alt is already passed through to some applications which can't use anything else since shift and control is already occupied.

kchibisov:
Assume next
```
-------------
|  2  |  1  |
|-----|-----|
|  3  |  4  |
-------------
```
Is it intended that block selection started in area 4(3) can be dynamically changed only to area 2(1), and not 1(2),3(4)? 

selection

chrisduerr:
@kchibisov Should be fixed now, I've also fixed a problem where the block selection would go to the end of the terminal when leaving the mouse during a selection. And there should be no more off-by-one errors.

PR #2563, opened 2019-06-19
Bump appdata to v0.3.3

1ace:
The timestamp is that of when the `v0.3.3` tag was created.

While at it, I also changed the date format to follow ISO 8601 :+1:


1ace:
> Also a big fan of ISO 8601, if that works it's fine by me.

That's actually the date format that [should be used] (https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases)

That said, the `<releases>` list is optional, and looking at the various appdata installed on my machine, none of them have it. They only have the stuff that's "always true", nothing that needs regular updating like the "latest release".

Maybe alacritty's `<releases>` list could simply be removed?

chrisduerr:
Reading through that, I think we should keep the `<releases>` section and instead of editing the current release, it would actually be better to add a new release for every release.

Adding a URL linking to the github release notes would also be a good idea I think.

Issue #1017, opened 2018-01-11
Scrollback Buffer Search


nixpulvis:
Once we have scrollback (#124), it's pretty reasonable to want to be able to search that. The challenge here is building something that works well, while still feeling simple, clean and lightweight. This is I believe the first "mode" proposed to alacritty. As in I hit some key binding and suddenly the rules for what my keys do has changed. I can't think of any other modes right now, but maybe that helps inspire the design.

Anyway sorry for getting ahead of things here, but this is a feature I'm dying to have. For now re-running the command and passing it to `rg` will have to suffice.

alexozer:
A line must be drawn at some point though to delegate to tmux. I'm wondering if implementing scrollback search in a mode would be a slippery slope to implementing something like termite's entire selection mode, and whether that's in scope simplicity and performance-wise here.

maximbaz:
Search is one of the reasons I will keep using tmux even after scrollback is merged, assuming #1000 is implemented too. 

I was considering to report this exact feature request myself a few days ago, but then I thought of what search capabilities I use today in tmux, and I realized that reimplementing them natively in alacritty would be too expensive, so I don't think vanilla alacritty will ever suffice my needs. 

I support the feature request, perhaps for many people that all they need, but here's for the reference the list of search capabilities that I identified as currently used by me with tmux:

- Incremental search (not type-then-search)
- Prepared search queries (e.g. `❯` is my prompt character, and I have a tmux shortcut that starts a search for the character `❯`, which allows me to jump to previous prompts in one key press)
- Visual selection (i.e. once you find something in the scrollback history (e.g. start of a log entry), use keyboard shortcuts to expand the selection (e.g. to select the entire log entry)
- Operations on the visual selection (via `tmux-yank` plugin):
    - Copy the selection to clipboard
    - Paste the selection to the prompt directly without copying it to clipboard
- Fuzzy-search + smart text blocks extraction (via `extrakto` plugin):
    - Open URL in browser
    - Open file path in the text editor
    - Copy to clipboard
    - and many more.

And after I saw this list, I realized that this is too much for alacritty, so I'll always be using tmux.

chrisduerr:
I'm not saying something like a vim-mode might solve all of those. But it kinda would.

maximbaz:
[...]

Issue #1615, opened 2018-10-02
Piping visible text to external programs


Deewiant:
(This was previously mentioned in https://github.com/jwilm/alacritty/issues/113#issuecomment-298948392 by a now-deleted user but seems to have been lost in the noise, so I thought I'd write it up properly.)

There is a patch for st called [externalpipe](https://st.suckless.org/patches/externalpipe/). It allows you to bind shortcut keys for piping the text from the terminal's screen to arbitrary external programs. I would very much like to have this functionality available in Alacritty.

The example there uses it as a workaround for URL recognition (#113) by piping to an external URL recognizer and calling `dmenu` to select among the options. I find this quite convenient, using the `urifind` tool from http://metacpan.org/pod/URI::Find. One could also use other tools to implement different kinds of pattern matching, thereby solving the use cases mentioned by @casey in https://github.com/jwilm/alacritty/issues/113#issuecomment-298079977 and https://github.com/jwilm/alacritty/issues/113#issuecomment-298081114. The main difference being that this would be a keyboard-driven instead of a mouse-driven interface.

In the end I would expect a configuration interface like the following in `key_bindings` (approximating what @casey's examples might look like using `dmenu` to pick among the matches visible on-screen):

```yaml
key_bindings:
  - { key: F, mods: Control, externalpipe: { program: "sh", args: ["-c", "rg -o 'https://[^\\s]+' | dmenu | xargs firefox"] } }
  - { key: I, mods: Control, externalpipe: { program: "sh", args: ["-c", "rg -o '[-_0-9a-zA-Z/]+:[0-9]+' | dmenu | parallel -d: 'some-ide {1} --line {2}'"] } }
  - { key: G, mods: Control, externalpipe: { program: "sh", args: ["-c", "rg -o '#[0-9]+' | dmenu | xargs -i firefox https://github.com/jwilm/alacritty/issues/{}"] } }
  - { key: G, mods: Control|Shift, externalpipe: { program: "sh", args: ["-c", "rg -o '[a-f0-9]{40}' | dmenu | xargs -i firefox https://github.com/jwilm/alacritty/commit/{}"] } }
```

This seems like a simpler way of approaching URL recognition and similar issues, since with this approach Alacritty doesn't have to concern itself with performing the pattern matching, only providing a straightforward data dump. It may be a relatively quick way of having a solution that is good enough for a lot of people.

chrisduerr:
[...]

PR #2568, opened 2019-06-20
RFC: Dump scrollback buffer to new instance of alacritty running a pager

jynnantonix:
This is a potential compromise for fixing #1017.  Currently being unable to search the scrollback buffer severely limits its usefulness but at the same it's questionable whether this should be implemented directly in alacritty itself.

This PR adds an action to spawn a new instance of alacritty with a user specified pager and then dump the contents of the scrollback buffer into the stdin of the new instance.  This neatly sidesteps the issue of having to implement search functionality in alacritty itself while giving users the freedom to use whichever pager best suits their needs.  The default is `less` but there's nothing stopping someone from using an editor like `vim` as their pager instead.

The contents of the scrollback buffer are first converted into a string (by iterating over every occupied cell) before being written out, which means we end up with a copy of the buffer in memory.  However since this is something that is only done in response to explicit user action, it shouldn't end up getting out of hand.

This is probably not fit for merging in its current state but I wanted to ask if this is something that would be accepted upstream before spending time cleaning it up.

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -28,6 +28,7 @@ pub struct Options {
     pub live_config_reload: Option<bool>,
     pub print_events: bool,
     pub ref_test: bool,
+    pub inherit_stdin: bool,
     pub dimensions: Option<Dimensions>,
     pub position: Option<Delta<i32>>,
     pub title: Option<String>,
@@ -45,6 +46,7 @@ impl Default for Options {
             live_config_reload: None,
             print_events: false,
             ref_test: false,
+            inherit_stdin: false,
             dimensions: None,
             position: None,
             title: None,
@@ -84,6 +86,11 @@ impl Options {
                     .long("no-live-config-reload")
                     .help("Disable automatic config reloading")
                     .conflicts_with("live-config-reload"),
+            )
+            .arg(
+                Arg::with_name("inherit-stdin")
+                    .long("inherit-stdin")
+                    .help("Forward stdin to the command to be run.  Only useful with `-e`."),
             )
             .arg(
                 Arg::with_name("print-events")
@@ -172,6 +179,10 @@ impl Options {
             options.print_events = true;
         }
 
+        if matches.is_present("inherit-stdin") {
+            options.inherit_stdin = true;
+        }
+
         if matches.is_present("live-config-reload") {
             options.live_config_reload = Some(true);
         } else if matches.is_present("no-live-config-reload") {
@@ -246,6 +257,7 @@ impl Options {
             self.working_dir.or_else(|| config.working_directory().to_owned()),
         );
         config.shell = self.command.or(config.shell);
+        config.inherit_stdin = self.inherit_stdin;
 
         config.window.dimensions = self.dimensions.unwrap_or(config.window.dimensions);
         config.window.position = self.position.or(config.window.position);
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -95,6 +95,10 @@ pub struct Config {
     #[serde(default, deserialize_with = "failure_default")]
     pub shell: Option<Shell<'static>>,
 
+    /// Path to a pager to use for searching the scrollback buffer
+    #[serde(default, deserialize_with = "failure_default")]
+    pub pager: Option<Shell<'static>>,
+
     /// Path where config was loaded from
     #[serde(default, deserialize_with = "failure_default")]
     pub config_path: Option<PathBuf>,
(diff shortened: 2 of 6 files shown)


chrisduerr:
I think that implementing this into Alacritty would be a far better solution than spawning a second instance.

Do you think this has a merit beyond just searching or was that the only goal? For search I'd like to dismiss this completely, but I feel like there might potentially be some alternative usecases.

This is similar to what https://github.com/jwilm/alacritty/issues/1615 proposes, so it might have some benefit beyond search.

chrisduerr:
As an additional note, I'd like to mention that I'd also like to have URL opening support directly in Alacritty. So I'm not sure if #1615 would be solved by that already.

Generally regex searching the complete buffer and selecting all matches would already offer URL keyboard selection to some degree, but I think we'd want to have separate dedicated support for it.

jynnantonix:
Having it natively in alacritty would be ideal but the last comment on #1017 was 11 months ago so I hope you can understand my skepticism about if/when it gets implemented.

I personally only need it for searching but I could see this being a more general mechanism to run an arbitrary program with the contents of the scrollback buffer (rather than just another instance of alacritty) without adding too much complexity.

We could add something like the current command action where different keybindings run different programs but they all get the buffer as their stdin.

chrisduerr:
> Having it natively in alacritty would be ideal but the last comment on #1017 was 11 months ago so I hope you can understand my skepticism about if/when it gets implemented.

Well it's just a matter of putting in the time and effort to implement it. There's nothing preventing it from getting implemented.

Choosing subpar solutions just because people are not interested in putting in the effort to do it properly just seems very unproductive for writing good software.

I'll definitely get around to it, but I can only do one thing at a time, so things just take time.

> We could add something like the current command action where different keybindings run different programs but they all get the buffer as their stdin.

I'd definitely like to make it possible to pipe it into an arbitrary command if possible. So that would definitely need to be required.

But before putting any more work into this, I'd like to see an actual real-world usecase for this beyond URL launching and scrollback buffer search. Because both of these will 100% land in Alacritty at some point.

jynnantonix:
Personally I don't think there's anything wrong with landing a stop-gap solution now and then getting rid of it once the "proper" solution is ready. 

Not having this feature was making me pretty unproductive at writing any kind of software (let alone good software) so I will continue to maintain these patches locally as the benefit to effort ratio is pretty high for me personally.  I look forward to the day that I can drop them.

chrisduerr:
> Personally I don't think there's anything wrong with landing a stop-gap solution now and then getting rid of it once the "proper" solution is ready.

I'm not a fan of landing subpar code when things could be done properly with a little more effort. I'm happy to guide anyone interested in implementing these features, however hacking in a bunch of tacked-on workarounds because nobody is interested in doing it properly is not a valid solution to me. Especially because it's much harder to maintain or remove something once it has been added than it is to do it properly.

chrisduerr:
I'm going to close this issue for now, to make it clear that there is no plan to work on this feature at the moment.

With the latest changes to URL searching, there have been major strides torwards being able to do some of the things what this PR was intended for, so hopefully we can make progress on that end.

nicklan:
I find this is a generally useful function to have. It happens to address a couple of current pain points that currently exist, but I would use this even if alacritty supported searching.

Some examples of what I would use this for:
- A quick way to get a log that was dumped to stdout that I want to paste into a ticket/comment/etc
- When I need more rich searching (For complex search I imagine many users prefer their editor to whatever their terminal implements)
- If I'm creating documentation it's nice to run a bunch of commands, then pop that into an editor so I can clean it up and save it

That being said, I think a simpler `Action` to just "dump scrollback buffer to file" (where file has a default path but can be configured) would be ideal for my use cases (and generally useful).  Possibly an extra option to also trigger a command where the file dumped can be supplied as an argument.

Would there be any interest in revisiting/modifying this feature given the above? I'd be happy to take a stab and making the needed changes if the above flow might get merged.

jynnantonix:
I have a branch of the 3.3 release with this change applied on top: https://github.com/jynnantonix/alacritty

I plan on keeping it up to date as new releases of alacritty come out.  Like yourself, I probably will keep using this feature even if alacritty gains the ability to do search.  In addition to the things you mentioned, other features that I've found really useful:

* `less` has a command to only show matching lines (`&`)
* Being able to dump part of the scrollback buffer into a file; sort of like a retroactive `tee`
* Just being able to keep the scrollback buffer open in one window as I keep working in the other one

If you'd like, please give it a try.  I'm more than happy to fix any issues you run into.

chrisduerr:
Just fyi, I'm open to continue investigating this in #1615. So arguments for or against it are best made in there.

nicklan:
Thanks @jynnantonix working great for me at the moment.

I'd still prefer it as being able to dump to a file since I usually want to open it in emacs, but a gross combination of arguments as my pager has things working for now :)

Issue #2570, opened 2019-06-21
Weird carat behavior when scrolling


lovesegfault:
Using the latest Alacritty, on Gentoo, and under Sway I see this interesting behavior when scrolling. I think the best way to describe it is to show it:

[![Demo](https://i.imgur.com/Z9PlPfP.gif)](https://i.imgur.com/tLckMYA.mp4)

P.S. Click on image for full video.

chrisduerr:
Can you reproduce this in X or XWayland? Also does setting a different cursor (using `printf '\e[4q'` for example) fix this?

kchibisov:
@chrisduerr If the issue is about cursor, I can reproduce it on X11 with zsh just by opening new `alacritty` instance typing `ls -laF` bunch of times, then `Ctrl + l`, `ls -laF` and scroll up. It only happens when your prompt isn't at the bottom of the screen though (in the middle e.g).

chrisduerr:
Huh, I don't think I was able to reproduce this in bash on macOS at work, but on X11/Arch/zsh I can reproduce it when following your steps.

I'm surprised I haven't noticed this, so my guess would be that this might be a regression. Testing the previous versions and bisecting the broken commit if those work might be a good idea.

kchibisov:
@chrisduerr I'll bisect in a moment.

chrisduerr:
That would be much appreciated, thanks for the help!

kchibisov:
@chrisduerr This one cfc20d4f34dca535654cc32df18e785296af4cc5

chrisduerr:
My best bet would be the following line which hasn't been carried over in the cursor rework:
https://github.com/jwilm/alacritty/commit/cfc20d4f34dca535654cc32df18e785296af4cc5#diff-7b3174257e73247b46a5a32b142d45beL440

I'm not sure *why* it is required, but replacing `self.cursor.line` with `self.inner.line` might be worth investigating. If that's not it, I'd have to look into this a bit more.

kchibisov:
@chrisduerr I've applied the following diff on top of the cfc20d4 and it seems like it resolves the issue.
```
diff --git a/src/term/mod.rs b/src/term/mod.rs
index 9e879ad..2e1236a 100644
--- a/src/term/mod.rs
+++ b/src/term/mod.rs
@@ -378,7 +378,7 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
                     let cell = Indexed {
[...]

PR #2573, opened 2019-06-21
Fix inconsitent cursor position when scrolling

kchibisov:
This commit fixes regression introduced in cfc20d4f34dca535654cc32df18e785296af4cc5.
`self.cursor.line` forced the cursor to hold a fixed location while scrolling
until its "original" location (usually the shell prompt) went off the screen.
So cursor position should be keep updated, which can be achieved by using
`self.inner.line()`.

cc @chrisduerr @lovesegfault


kchibisov:
Fixes #2570

lovesegfault:
Y'all rock

PR #2580, opened 2019-06-22
Perform clear + `swap_buffers` before `window.show`

kchibisov:
This should fill window with background color while it is offscreen instead of
showing uninitilized window (just borders) and then performing `clear`. So, the new behavior
should prevent glitches during startup (e.g. content of a windows below,
garbage from drivers, so on).

--
However, I was able to get glitches ( previous app content ) with the new behavior, but it happens 1 time out of 15-20 openings, instead of every time. So if someone knows how should I prevent this 1 time, I'll appreciate it.

P.s. I'm not very familiar with `opengl`, so if you see something completely wrong, let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Text Cursor position when scrolling
 - Performance issues while resizing Alacritty
 - First unfullscreen action ignored on window launched in fullscreen mode
+- The window is now filled with the background color before displaying
 
 ## 0.3.3
 
--- a/alacritty_terminal/src/display.rs
+++ b/alacritty_terminal/src/display.rs
@@ -224,6 +224,33 @@ impl Display {
             api.clear(background_color);
         });
 
+        // We should call `clear` when window is offscreen, so when `window.show()` happens it
+        // would be with background color instead of uninitialized surface.
+        window.swap_buffers()?;
+
+        window.show();
+
+        // Set window position
+        //
+        // TODO: replace `set_position` with `with_position` once available
+        // Upstream issue: https://github.com/tomaka/winit/issues/806
+        if let Some(position) = config.window.position {
+            let physical = PhysicalPosition::from((position.x, position.y));
+            let logical = physical.to_logical(window.hidpi_factor());
+            window.set_position(logical);
+        }
+
+        use StartupMode::*;
+        #[allow(clippy::single_match)]
+        match config.window.startup_mode() {
+            Fullscreen => window.set_fullscreen(true),
+            #[cfg(target_os = "macos")]
+            SimpleFullscreen => window.set_simple_fullscreen(true),
+            #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+            Maximized if window.is_x11() => window.set_maximized(true),
+            _ => (),
+        }
+
         Ok(Display {
             window,
             renderer,
--- a/alacritty_terminal/src/display.rs
+++ b/alacritty_terminal/src/display.rs
@@ -240,14 +240,13 @@ impl Display {
             window.set_position(logical);
         }
 
-        use StartupMode::*;
         #[allow(clippy::single_match)]
         match config.window.startup_mode() {
-            Fullscreen => window.set_fullscreen(true),
+            StartupMode::Fullscreen => window.set_fullscreen(true),
             #[cfg(target_os = "macos")]
-            SimpleFullscreen => window.set_simple_fullscreen(true),
-            #[cfg(not(any(target_os = "macos", target_os = "windows")))]
-            Maximized if window.is_x11() => window.set_maximized(true),
+            StartupMode::SimpleFullscreen => window.set_simple_fullscreen(true),
+            #[cfg(not(any(target_os = "macos", windows)))]
+            StartupMode::Maximized if window.is_x11() => window.set_maximized(true),
             _ => (),
         }
 
--- a/alacritty_terminal/src/window.rs
+++ b/alacritty_terminal/src/window.rs
@@ -17,7 +17,7 @@ use std::ffi::c_void;
 use std::fmt::Display;
 
 use crate::gl;
-use glutin::dpi::{LogicalPosition, LogicalSize, PhysicalPosition, PhysicalSize};
+use glutin::dpi::{LogicalPosition, LogicalSize, PhysicalSize};
 #[cfg(target_os = "macos")]
 use glutin::os::macos::WindowExt;
 #[cfg(not(any(target_os = "macos", windows)))]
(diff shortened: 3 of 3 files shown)


chrisduerr:
Which GPU/driver are you using? I wasn't aware this is an issue.

kchibisov:
@chrisduerr I think it's about de + wm ( I was experienced artifacts from a driver during startup long time ago with `GTX 670` and `nvidia proprietary driver`, but now I'm using `RX 570` and `amdgpu` driver, so this `PR` wasn't tested against driver artifacts).

I'm using `kde` (`kwix_x11`) with `tiling` script. So, when a new window is spawning, it immediately moving to a new location. `alacritty` (master) spawns a window, just a bunch of borders + transparent background, and after some time it actually gets filled with colors and stuff like that. So if I have `firefox` on the left and then open `alacritty`, it will be opened on top of `firefox`, but tiling script will send it to the right to form a new tile ( it happens instantly, you can't see this happens ). However the thing you can see is a `firefox` page in `alacritty` for a bit. This thing happens only with `alacritty` in my environment.

Also, assume you have `HDD` and you're doing a cold start (no other apps, only `alacritty`). `alacritty` will draw borders + transparent background, and only after 4-5 secs ( very slow `HDD` though :- ( ), it'll draw background. `Firefox` in this scenario doing everything in a background and appears when ready to use (just e.g.).

kchibisov:
@chrisduerr These [recordings](https://www.dropbox.com/sh/d0ximfa19ez87ul/AAB9WQJpcDSJ97zBqSTmh_9ua?dl=0) could give you an understanding of what I'm trying to achieve here (bit rate is low due to my cpu limitations).

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Block selection mode when Control is held while starting a selection
 
+### Changed
+
+- The window is now filled with the `background color` before displaying
```
chrisduerr: I think I'd move that in the fixed section, since it solves an existing problem.

alacritty_terminal/src/window.rs
```diff
@@ -408,6 +386,16 @@ impl Window {
...
 
+    pub fn set_maximized(&self, maximized: bool) {
+        #[cfg(not(any(target_os = "macos", windows)))]
+        {
+            let window = self.window();
+            if self.event_loop.is_x11() {
+                window.set_maximized(maximized);
+            }
+        }
+    }
```
chrisduerr: The way this method is split apart now is a bit dangerous. The only reason why things are maximized after starting on X11 is because doing it before showing doesn't work, not because `set_maximized` only works on X11. This changes the window API in a way that makes it seem like `set_maximized` only works on X11, which is not the case.

A better way to do this would be to always call `window.set_maximized(bool)` in the Window's method and handle the logic at the callsite in display.rs.

kchibisov:
> However, I was able to get glitches ( previous app content ) with the new behavior, but it happens 1 time out of 15-20 openings

@chrisduerr Do you have any ideas about this one? I've tried to throttle before opening, but it doesn't help, seems like pipeline related issue.

chrisduerr:
I'm familiar with the issue in general, however I can not reproduce it at all on my system, even without this patch. So I'm not sure why it would happen at all for you.

You're running X11, right? Do you use a compositor? I'd assume you're just running standard KDE with its compositor enabled?

kchibisov:
@chrisduerr Yes, standard kde on x11, with its default compositor `kwin`, but a custom script for windows placement.

So in default kde without any scripts + compositing everything is fine.
Script + placing by kwin - seems fine ( not shure if it is)
Script + placing in the end instead of creating new master window - the issue that I'm fixing.

I've tried different scripts and all of them have problems with `alacritty` without my patch.

But I think I was able to reproduce it with `GTX 670` + `nvidia proprietary driver` + `i3` long time ago. In `nvidia` setup alacritty was the only window to have strange artifacts on startup (with and without compositor). With `nouveau` everything was fine.

In theory I can check `nvidia` setup, but I'm a little bit lazy here. Since It's not that simple (just long) to replace `GPU` in my machine.

kchibisov:
@chrisduerr Just to make it clear. I'm already satisfied with behavior I've achieved here.

P.s.
If you want me to retest nvidia setup let me know.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - GUI programs launched by Alacritty starting in the background on X11
 - Text Cursor position when scrolling
+- The window is now filled with the `background color` before displaying
```
chrisduerr: This suggests that `background color` is a specific config variable, which it is not. So the graves around it should just be removed.

alacritty_terminal/src/display.rs
```diff
@@ -224,6 +224,44 @@ impl Display {
             api.clear(background_color);
         });
 
+        // We should call `clear` when window is offscreen, so when `window.show()` happens it
+        // would be already with background color instead of bunch of borders + uninitialized
```
chrisduerr: `bunch of borders` is implementation specific to the WM which is used, if your WM has no borders around your windows, Alacritty will not just render the borders..

So just calling it an `unitialized surface` should be enough, without any mention of borders.

alacritty_terminal/src/display.rs
```diff
@@ -224,6 +224,44 @@ impl Display {
...
+            }
+        }
+
+
+        #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+        {
+            if window.x11_window() && StartupMode::Maximized == startup_mode {
+                window.set_maximized(true);
+            }
+        }
```
chrisduerr: Looking at the three `startup_mode` statements here, I feel like it would be great to just put this into a `match` statement. That would clean up the code a lot.

This isn't related to this PR at all, but I usually think that if there's a chance to refactor things a bit, it's probably good to take it. Do you want to implement that or should I send a separate PR?

I'm thinking about something like this:

```rust
match config.window.startup_mode() {
    Fullscreen => window.set_fullscreen(true),
    #[cfg(target_os = "macos")]
    SimpleFullscreen => window.set_simple_fullscreen(true),
    #[...]
    Maximized if window.x11_window() => window.set_maximized(true),
    _ => (),
}
```

I also think the `use glutin::os::macos::WindowExt` shouldn't be required here, so that can probably be removed while you're at it.

alacritty_terminal/src/window.rs
```diff
@@ -387,6 +365,11 @@ impl Window {
         }
     }
 
+    #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+    pub fn x11_window(&self) -> bool {
+        self.event_loop.is_x11()
+    }
```
chrisduerr: I think the method name could probably be improved, since `window.x11_window()` is a bit of unnecessary duplication. I'd probably just go with `window.is_x11` to stay consistent with the glutin API's name.

Alternatively since Display::new has access to the event_loop too, before the window is created, we could store the x11 state in a variable before creating the window and query that later, but I think it's nicer to expose this method on the window. I'm also not sure how this is handled in the evlp2 branch, but it's possible that things need to be changed there anyways, so I wouldn't worry *too* much.

chrisduerr:
> @chrisduerr Just to make it clear. I'm already satisfied with behavior I've achieved here.

I think this is our best bet. We set the window content before making it visible, as far as I'm concerned there's not a ton more we can do about it.

alacritty_terminal/src/display.rs
kchibisov: Yes, sure.

Issue #2414, opened 2019-05-04
disable text reflow on alt screen mode


yorickvP:
(writing it down here after some discussion on irc)
When I run any application that uses alt screen mode (emacs, htop, weechat), it probably does it's own text reflow. Resizing such applications currently reflows their rendered UI's, which is never useful. The application then quickly redraws.

Other terminals should be checked to see what they do in this case.

PR #2581, opened 2019-06-23
Disable text reflow on `ALT_SCREEN` mode

kchibisov:
Apps that are using `ALT_SCREEN` mode are probably doing their own
reflow, so `our` reflow can lead to strange "artifacts" during resize.

> Other terminals should be checked to see what they do in this case.

I've testing `kitty` for this, and it seems like its doing text reflow on `ALT_SCREEN` mode. 

The behavior without reflow feels better in apps with `ALT_SCREEN` mode though.

This `PR` uses the approach that was before b1032bc , when we need resize without reflow.

Implements #2414

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Block selection mode when Control is held while starting a selection
 
+### Changed
+
+- Text reflow is now disabled in `ALT_SCREEN` mode
+
 ### Fixed
 
 - GUI programs launched by Alacritty starting in the background on X11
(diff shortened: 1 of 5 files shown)


chrisduerr:
I've actually got a PR working on this myself, fixing the performance bug with text reflow too.

I'm not sure if generally disabling reflow is the best choice here, there's specific escape sequences to disable reflow though, which should likely be respected for automatic text reflow too.

Investigating if it is enough to listen for those escapes is something I still wanted to do before submitting the PR though.

You can see my current work in progress here: https://github.com/jwilm/alacritty/pull/2582

kchibisov:
@chrisduerr You can close this one, if you want.

PR #2591, opened 2019-06-25
Fix dependencies for arch in INSTALL.md

benjaminbauer:
current master does not build on arch w/o libxcb
https://www.archlinux.org/packages/extra/x86_64/libxcb/


benjaminbauer:
Thanks for all the great work you put into the project!

PR #2598, opened 2019-06-27
Better font antialiasing using DirectWrite

ark0f:
`DWRITE_RENDERING_MODE_NATURAL`:

`DWRITE_RENDERING_MODE_NATURAL_SYMMETRIC`:

### DIFF
--- a/font/src/directwrite/mod.rs
+++ b/font/src/directwrite/mod.rs
@@ -147,7 +147,7 @@ impl crate::Rasterize for DirectWriteRasterizer {
             &glyph_run,
             self.device_pixel_ratio * (96.0 / 72.0),
             None,
-            dwrote::DWRITE_RENDERING_MODE_NATURAL,
+            dwrote::DWRITE_RENDERING_MODE_NATURAL_SYMMETRIC,
             dwrote::DWRITE_MEASURING_MODE_NATURAL,
             0.0,
             0.0,


chrisduerr:
Looking at it like this, it's tough to argue, however smaller sizes might be a bit different. Usually more aliasing at smaller sizes can easily lead to increased blur.

Do you know why this has to be set manually by the application? I would think that fonts should look the same across all applications, so it doesn't make sense to me that this would be set on a per-window basis.

nixpulvis:
@chrisduerr looking at the [docs](https://docs.microsoft.com/en-us/windows/win32/api/dwrite_3/ne-dwrite_3-dwrite_rendering_mode1) it seems like this is a large render optimized setting:

> Similar to natural mode except that antialiasing is performed in both the horizontal and vertical directions. This is typically used at larger sizes to make curves and diagonal lines look smoother. The antialiasing may be either ClearType or grayscale depending on the text antialiasing mode.

It sounds like the only real difference is that this will do two antialiasing passes on both the x and y axis, while currently it only does the x axis. So it seems like a sane performance setting, I'm not sure how exactly we should decide tradeoffs like this in general, but given how this seems like it's mostly intended for large font sizes, it seems out of place as the default here.

Am I missing something? Should we have antialiasing "quality" settings like many OpenGL applications (games)?

chrisduerr:
> but given how this seems like it's mostly intended for large font sizes, it seems out of place as the default here.

This is my concern too, with fonts you usually risk having blurry fonts if your anti aliasing is too agressive.

> Am I missing something? Should we have antialiasing "quality" settings like many OpenGL applications (games)?

Games are way different than fonts though, anti-aliasing settings usually make a big difference in performance, but the quality is usually always better the more performance you're giving away. With fonts that's not the same, more anti aliasing isn't necessarily better.

nixpulvis:
@chrisduerr yea, I'm not trying to say we're the same as a game... but just that we _could_ make a config setting. Though I'd really rather not expose that if possible. A terminal's primary job is reading a writing text, we should do that well by default.

chrisduerr:
> we could make a config setting

I think things like font anti aliasing are better configured on a system-wide basis. It doesn't make a lot of sense to me to override this for a specific app unless there's a very good reason.

TheDecryptor:
I made a couple of builds to see the difference, and honestly at the default font size it's barely noticeable to me.

DWRITE_RENDERING_MODE_NATURAL
trim_nat

DWRITE_RENDERING_MODE_NATURAL_SYMMETRIC
trim_symm

Here's the difference between the two
dwrite_diff

🤷‍♂ 

DirectWrite offers an enticing `DWRITE_RENDERING_MODE_DEFAULT` option, which unfortunately doesn't actually work (It's for the higher level layout API). But if you've got a specific font face (Which is the case here) you can ask for what to use with `get_recommended_rendering_mode_default_params`, which in quick testing results in natural for plain text, and symmetric for bold text.

```rust
let rendering_mode = font.get_recommended_rendering_mode_default_params(
    glyph.size.as_f32_pts(),
    self.device_pixel_ratio * (96.0 / 72.0),
    dwrote::DWRITE_MEASURING_MODE_NATURAL
);

let glyph_analysis = GlyphRunAnalysis::create(
    &glyph_run,
    self.device_pixel_ratio * (96.0 / 72.0),
    None,
    rendering_mode,
    dwrote::DWRITE_MEASURING_MODE_NATURAL,
    0.0,
    0.0,
)
.or_else(|_| Err(Error::MissingGlyph(glyph.c)))?;
```

chrisduerr:
@TheDecryptor Since you've implemented the alternative solution, do you think it makes sense to push for it? If so, would you be willing to send a PR with your changes applied? That way not all the work is on @ark0f and we might be able to get this merged a little quicker.

If you don't think there's any benefit to it (I'm not super concerned either way, since I'm not very familiar with Windows), please let me know. Then it might be better to just clearly state that we're not interested in this.

TheDecryptor:
Sure, I'll make a PR.

chrisduerr:
Closing this in favor of https://github.com/jwilm/alacritty/pull/2598.

Thanks for doing the initial pathfinding work to get this all figured out @ark0f.

PR #2605, opened 2019-06-30
Suggest installing fish completions per-user

faho:
This uses $fish_complete_path[1], which should be in the user's home directory, so `sudo` is no longer needed.


faho:
Done.

PR #2606, opened 2019-06-30
Gnome tweaks

bkoropoff:
This makes two small changes for better support under gnome:

- Allow setting the `_GTK_THEME_VARIANT` hint.  This permits selecting dark title bars to go with a dark terminal theme
- Allow setting both the instance and class portions of the `WM_CLASS` hint (with CLI support).  This gives complete control over which desktop entry gnome-shell associates with the window.  Without this, gnome-shell would always associated the created window with alacritty.desktop, even if it was launched from a custom desktop entry (e.g. to create a dedicated vim window).  The commit message gives an example.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -72,6 +72,12 @@ window:
     # General application class
     general: Alacritty
 
+  # GTK theme variant (Linux only)
+  #
+  # Override the variant of the GTK theme. Commonly supported values are `dark` and `light`.
+  # Set this to `None` to use the default theme variant.
+  gtk_theme_variant: None
+
 scrolling:
   # Maximum number of lines in the scrollback buffer.
   # Specifying '0' will disable scrolling.
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -66,7 +66,11 @@ window:
   #title: Alacritty
 
   # Window class (Linux only):
-  #class: Alacritty
+  class:
+    # Application instance name
+    instance: Alacritty
+    # General application class
+    general: Alacritty
 
 scrolling:
   # Maximum number of lines in the scrollback buffer.
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -250,7 +250,14 @@ impl Options {
         config.window.dimensions = self.dimensions.unwrap_or(config.window.dimensions);
         config.window.position = self.position.or(config.window.position);
         config.window.title = self.title.or(config.window.title);
-        config.window.class = self.class.or(config.window.class);
+
+        if let Some(class) = self.class {
+            let parts : Vec<_> = class.split(',').collect();
+            config.window.class.instance = parts[0].into();
+            if let Some(&general) = parts.get(1) {
+                config.window.class.general = general.into();
+            }
+        }
 
         config.set_dynamic_title(config.dynamic_title() && config.window.title.is_none());
 
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -136,8 +136,8 @@ pub struct Config {
     alt_send_esc: DefaultTrueBool,
 
     /// Shell startup directory
-    #[serde(default, deserialize_with = "failure_default")]
-    working_directory: WorkingDirectory,
+    #[serde(default, deserialize_with = "option_explicit_none")]
+    working_directory: Option<PathBuf>,
 
     /// Debug options
     #[serde(default, deserialize_with = "failure_default")]
@@ -214,37 +214,12 @@ impl Config {
 
     #[inline]
     pub fn working_directory(&self) -> &Option<PathBuf> {
-        &self.working_directory.0
+        &self.working_directory
     }
 
     #[inline]
     pub fn set_working_directory(&mut self, working_directory: Option<PathBuf>) {
-        self.working_directory.0 = working_directory;
-    }
-}
-
-#[derive(Default, Debug, PartialEq, Eq)]
-struct WorkingDirectory(Option<PathBuf>);
-
-impl<'de> Deserialize<'de> for WorkingDirectory {
-    fn deserialize<D>(deserializer: D) -> Result<WorkingDirectory, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        let value = serde_yaml::Value::deserialize(deserializer)?;
-
-        // Accept `None` to use the default path
-        if value.as_str().filter(|v| v.to_lowercase() == "none").is_some() {
-            return Ok(WorkingDirectory(None));
-        }
-
-        Ok(match PathBuf::deserialize(value) {
-            Ok(path) => WorkingDirectory(Some(path)),
-            Err(err) => {
-                error!("Problem with config: {}; using None", err);
-                WorkingDirectory(None)
-            },
-        })
+        self.working_directory = working_directory;
     }
 }
 
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -94,7 +94,7 @@ pub struct Config {
     pub mouse: Mouse,
 
     /// Path to a shell program to run on startup
-    #[serde(default, deserialize_with = "failure_default")]
+    #[serde(default, deserialize_with = "option_from_string_or_deserialize")]
     pub shell: Option<Shell<'static>>,
 
     /// Path where config was loaded from
@@ -331,6 +331,12 @@ impl<'a> Shell<'a> {
         S: Into<Cow<'a, str>>,
     {
         Shell { program: program.into(), args }
+    }
+}
+
+impl From<String> for Shell<'static> {
+    fn from(value: String) -> Self {
+        Shell::new(value)
     }
 }
 
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -94,7 +94,7 @@ pub struct Config {
     pub mouse: Mouse,
 
     /// Path to a shell program to run on startup
-    #[serde(default, deserialize_with = "option_from_string_or_deserialize")]
+    #[serde(default, deserialize_with = "from_string_or_deserialize")]
     pub shell: Option<Shell<'static>>,
 
     /// Path where config was loaded from
@@ -334,9 +334,9 @@ impl<'a> Shell<'a> {
     }
 }
 
-impl From<String> for Shell<'static> {
-    fn from(value: String) -> Self {
-        Shell::new(value)
+impl FromString for Option<Shell<'_>> {
+    fn from(input: String) -> Self {
+        Some(Shell::new(input))
     }
 }
 
@@ -426,7 +426,7 @@ pub fn option_explicit_none<'de, T, D>(deserializer: D) -> Result<Option<T>, D::
 }
 
 pub fn from_string_or_deserialize<'de, T, D>(deserializer: D) -> Result<T, D::Error>
-    where D: Deserializer<'de>, T: Deserialize<'de> + From<String> + Default
+    where D: Deserializer<'de>, T: Deserialize<'de> + FromString + Default
 {
     Ok(match Value::deserialize(deserializer)? {
         Value::String(value) => T::from(value),
@@ -434,17 +434,7 @@ pub fn from_string_or_deserialize<'de, T, D>(deserializer: D) -> Result<T, D::Er
     })
 }
 
-pub fn option_from_string_or_deserialize<'de, T, D>(deserializer: D) -> Result<Option<T>, D::Error>
-    where D: Deserializer<'de>, T: Deserialize<'de> + From<String>
-{
-    Ok(match Value::deserialize(deserializer)? {
-        Value::String(value) => Some(T::from(value)),
-        value => match T::deserialize(value) {
-            Ok(value) => Some(value),
-            Err(err) => {
-                error!("Problem with config: {}; using None", err);
-                None
-            }
-        }
-    })
+// Used over From<String>, to allow implementation for foreign types
+pub trait FromString {
+    fn from(input: String) -> Self;
 }
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -15,8 +15,10 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::path::PathBuf;
+use std::fmt::Display;
 
 use serde::{Deserialize, Deserializer};
+use serde_yaml::Value;
 
 mod bindings;
 mod colors;
@@ -418,17 +420,50 @@ impl Default for DefaultTrueBool {
     }
 }
 
+fn fallback_default<T, E>(err: E) -> T
+    where T: Default, E: Display
+{
+    error!("Problem with config: {}; using default value", err);
+    T::default()
+}
+
 pub fn failure_default<'a, D, T>(deserializer: D) -> Result<T, D::Error>
 where
     D: Deserializer<'a>,
     T: Deserialize<'a> + Default,
 {
-    let value = serde_yaml::Value::deserialize(deserializer)?;
-    match T::deserialize(value) {
-        Ok(value) => Ok(value),
-        Err(err) => {
-            error!("Problem with config: {}; using default value", err);
-            Ok(T::default())
-        },
-    }
+    Ok(T::deserialize(Value::deserialize(deserializer)?).unwrap_or_else(fallback_default))
+}
+
+pub fn option_explicit_none<'de, T, D>(deserializer: D) -> Result<Option<T>, D::Error>
+    where D: Deserializer<'de>, T: Deserialize<'de> + Default
+{
+    Ok(match Value::deserialize(deserializer)? {
+        Value::String(ref value) if value.to_lowercase() == "none" => None,
+        value => Some(T::deserialize(value).unwrap_or_else(fallback_default))
+    })
+}
+
+pub fn from_string_or_deserialize<'de, T, D>(deserializer: D) -> Result<T, D::Error>
+    where D: Deserializer<'de>, T: Deserialize<'de> + From<String> + Default
+{
+    Ok(match Value::deserialize(deserializer)? {
+        Value::String(value) => T::from(value),
+        value => T::deserialize(value).unwrap_or_else(fallback_default)
+    })
+}
+
+pub fn option_from_string_or_deserialize<'de, T, D>(deserializer: D) -> Result<Option<T>, D::Error>
+    where D: Deserializer<'de>, T: Deserialize<'de> + From<String>
+{
+    Ok(match Value::deserialize(deserializer)? {
+        Value::String(value) => Some(T::from(value)),
+        value => match T::deserialize(value) {
+            Ok(value) => Some(value),
+            Err(err) => {
+                error!("Problem with config: {}; using None", err);
+                None
+            }
+        }
+    })
 }
--- a/alacritty_terminal/src/config/window.rs
+++ b/alacritty_terminal/src/config/window.rs
@@ -1,5 +1,6 @@
-use crate::config::{failure_default, Delta};
+use crate::config::{failure_default, from_string_or_deserialize, Delta};
 use crate::index::{Column, Line};
+use crate::window::DEFAULT_NAME;
 
 #[serde(default)]
 #[derive(Deserialize, Debug, Clone, Default, PartialEq, Eq)]
@@ -33,8 +34,12 @@ pub struct WindowConfig {
     pub title: Option<String>,
 
     /// Window class
+    #[serde(deserialize_with = "from_string_or_deserialize")]
+    pub class: Class,
+
+    /// Window class instance
     #[serde(deserialize_with = "failure_default")]
-    pub class: Option<String>,
+    pub instance: Option<String>,
 
     /// TODO: DEPRECATED
     #[serde(deserialize_with = "failure_default")]
@@ -115,5 +120,25 @@ impl Dimensions {
     #[inline]
     pub fn columns_u32(&self) -> u32 {
         self.columns.0 as u32
+    }
+}
+
+/// Window class hint
+#[serde(default)]
+#[derive(Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct Class {
+    pub instance: String,
+    pub general: String
+}
+
+impl Default for Class {
+    fn default() -> Self {
+        Class { instance: DEFAULT_NAME.into(), general: DEFAULT_NAME.into() }
+    }
+}
+
+impl From<String> for Class {
+    fn from(value: String) -> Self {
+        Class { instance: value, general: DEFAULT_NAME.into() }
     }
 }
--- a/alacritty_terminal/src/config/window.rs
+++ b/alacritty_terminal/src/config/window.rs
@@ -1,4 +1,4 @@
-use crate::config::{failure_default, option_explicit_none, from_string_or_deserialize, Delta};
+use crate::config::{failure_default, option_explicit_none, from_string_or_deserialize, Delta, FromString};
 use crate::index::{Column, Line};
 use crate::window::DEFAULT_NAME;
 
@@ -137,7 +137,7 @@ impl Default for Class {
     }
 }
 
-impl From<String> for Class {
+impl FromString for Class {
     fn from(value: String) -> Self {
         Class { instance: value, general: DEFAULT_NAME.into() }
     }
--- a/alacritty_terminal/src/config/window.rs
+++ b/alacritty_terminal/src/config/window.rs
@@ -1,4 +1,4 @@
-use crate::config::{failure_default, from_string_or_deserialize, Delta};
+use crate::config::{failure_default, option_explicit_none, from_string_or_deserialize, Delta};
 use crate::index::{Column, Line};
 use crate::window::DEFAULT_NAME;
 
@@ -37,9 +37,9 @@ pub struct WindowConfig {
     #[serde(deserialize_with = "from_string_or_deserialize")]
     pub class: Class,
 
-    /// Window class instance
-    #[serde(deserialize_with = "failure_default")]
-    pub instance: Option<String>,
+    /// GTK theme variant
+    #[serde(deserialize_with = "option_explicit_none")]
+    pub gtk_theme_variant: Option<String>,
 
     /// TODO: DEPRECATED
     #[serde(deserialize_with = "failure_default")]
--- a/alacritty_terminal/src/window.rs
+++ b/alacritty_terminal/src/window.rs
@@ -151,9 +151,8 @@ impl Window {
         dimensions: Option<LogicalSize>,
     ) -> Result<Window> {
         let title = config.window.title.as_ref().map_or(DEFAULT_NAME, |t| t);
-        let class = config.window.class.as_ref().map_or(DEFAULT_NAME, |c| c);
 
-        let window_builder = Window::get_platform_window(title, class, &config.window);
+        let window_builder = Window::get_platform_window(title, &config.window);
         let windowed_context =
             create_gl_window(window_builder.clone(), &event_loop, false, dimensions)
                 .or_else(|_| create_gl_window(window_builder, &event_loop, true, dimensions))?;
@@ -255,7 +254,6 @@ impl Window {
     #[cfg(not(any(target_os = "macos", windows)))]
     pub fn get_platform_window(
         title: &str,
-        class: &str,
         window_config: &WindowConfig,
     ) -> WindowBuilder {
         use glutin::os::unix::WindowBuilderExt;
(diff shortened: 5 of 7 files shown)


alacritty_terminal/src/config/window.rs
```diff
@@ -36,9 +36,17 @@ pub struct WindowConfig {
...
+    #[serde(deserialize_with = "failure_default")]
+    pub instance: Option<String>,
+
     /// TODO: DEPRECATED
     #[serde(deserialize_with = "failure_default")]
     pub start_maximized: Option<bool>,
+    ///
+    /// Window class instance
+    #[serde(deserialize_with = "failure_default")]
+    pub gtk_theme_variant: Option<String>,
```
chrisduerr: New entries should be added *above* the deprecated values, to keep them all collected together at the bottom. Also please make sure the whitespace between elements is consistent.

alacritty_terminal/src/window.rs
```diff
@@ -267,23 +269,28 @@ impl Window {
...
             .with_window_icon(icon.ok())
             // X11
-            .with_class(class.into(), DEFAULT_NAME.into())
+            .with_class(instance.into(), class.into())
             // Wayland
-            .with_app_id(class.into())
+            .with_app_id(class.into());
+        if let Some(ref val) = window_config.gtk_theme_variant {
+            builder = builder.with_gtk_theme_variant(val.clone())
+        }
```
chrisduerr: Some whitespace around this would improve readability.

alacritty_terminal/src/window.rs
```diff
@@ -267,23 +269,28 @@ impl Window {
...
+        return builder
```
chrisduerr: This should just be `builder` without the `return`.

bkoropoff:
@chrisduerr Would that mean removing the commented out `gtk_theme_variant` line entirely (to represent the default of `None`)?

chrisduerr:
> @chrisduerr Would that mean removing the commented out gtk_theme_variant line entirely (to represent the default of None)?

No, all available configuration values need to be documented in the configuration file. Possible options would be either an empty string (which I'd assume is the default for the property itself), or an explicit `None`, which you can already find in other places of the config. Then the option should be uncommented by default if possible (which is why None is usually preferrable, though I don't think there's a general-purpose deserializer we have for that yet).

bkoropoff:
@chrisduerr 

OK, the `class` YAML setting can now be a sequence of `[instance, class]`, or a string (which preserves the old behavior).  The command line argument can now be of the form `instance,class`, with the old form working as before.

`gtk_theme_variant` now uses a custom deserializer which respects `None`

bkoropoff:
Also fixed unintentional Wayland app ID behavior change and formatting stuff

bkoropoff:
@chrisduerr Are any changes still needed, here?

alacritty/src/cli.rs
```diff
@@ -250,7 +250,14 @@ impl Options {
...
         config.window.title = self.title.or(config.window.title);
-        config.window.class = self.class.or(config.window.class);
+
+        if let Some(class) = self.class {
+            let parts : Vec<_> = class.split(',').collect();
+            config.window.class.instance = Some(parts[0].into());
+            if parts.len() >= 2 {
+                config.window.class.class = Some(parts[1].into());
+            }
+        }
```
chrisduerr: If we update the CLI, we also need to update the manpage and completions.

alacritty.yml
```diff
@@ -66,7 +66,10 @@ window:
   #title: Alacritty
 
   # Window class (Linux only):
-  #class: Alacritty
+  #class: [Alacritty, Alacritty]
+
+  # GTK theme variants (Linux only):
+  #gtk_theme_variant: None
```
chrisduerr: These options should both be uncommented by default if possible. That way users can just directly change them without having to uncomment them first.

I also think it might be worthwhile to describe `gtk_theme_variant` a bit more, since I'm not familiar with it myself. Is this just Dark/Light? If so, then this should be an enum, not a string.

alacritty_terminal/src/config/window.rs
```diff
@@ -34,13 +36,35 @@ pub struct WindowConfig {
...
+    /// GTK theme variant
+    #[serde(deserialize_with = "deserialize_gtk_theme_variant")]
+    pub gtk_theme_variant: Option<String>,
 
     /// TODO: DEPRECATED
     #[serde(deserialize_with = "failure_default")]
     pub start_maximized: Option<bool>,
 }
 
+fn deserialize_gtk_theme_variant<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
```
chrisduerr: You've copied the working directory deserializer here, right? I think it should be possible to make this a generic `deserialize_option` with the Result being `Result<Option<T>, D::Error>`. If `Deserialize` is a trait bound for `T`, then we can just use it for every `Option` which we want to accept explicit nones for.

alacritty_terminal/src/config/window.rs
```diff
@@ -117,3 +141,45 @@ impl Dimensions {
         self.columns.0 as u32
     }
 }
+
+/// Window class
+#[derive(Clone, Default, Debug, Eq, PartialEq)]
+pub struct Class {
+    pub instance: Option<String>,
+    pub class: Option<String>
```
chrisduerr: I think neither instance or class should be optional here. We should probably just fall back to `Alacritty` for both.

alacritty_terminal/src/config/window.rs
```diff
@@ -117,3 +141,45 @@ impl Dimensions {
...
+
+impl<'de> Deserialize<'de> for Class {
+    fn deserialize<D>(deserializer: D) -> Result<Class, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let value = serde_yaml::Value::deserialize(deserializer)?;
+
+        match value {
+            Value::Null => Ok(Default::default()),
```
chrisduerr: I don't think there's a need to support an explicit `Null` here, is there? If this value is not present, it shouldn't be null but never call the deserializer, so this would probably just trigger for `value: Null`, right? So I think we can get rid of that.

alacritty_terminal/src/config/window.rs
```diff
@@ -117,3 +141,45 @@ impl Dimensions {
...
+            Value::String(val) => Ok(Class { instance: Some(val), class: None }),
+            Value::Sequence(mut vals) => {
```
chrisduerr: I feel like we might be making it harder on ourselves with parsing here by using an array. Instead, maybe we can consider doing something similar to what `shell` allows? So both of the following would be allowed:

```yml
class: Alacritty
```

```yml
class:
  instance: Alacritty
  class: Alacritty
```

I think it would be nice to find a better name for class->class, since duplication here will inevitably lead to confusion, but I'm not sure if there's a standardized name for the part of the class which should always stay the same across instances.

>  A string that names the general class of applications to which the client that owns this window belongs. Resources that are specified by class apply to all applications that have the same class name. Class names are specified by the application writer. Examples of commonly used class names include: "Emacs", "XTerm", "XClock", "XLoad", and so on. 

Based on this documentation, maybe `general` is an option?

alacritty.yml
bkoropoff: Only dark/light tend to be used, but GTK apparently allows arbitrary strings:

https://developer.gnome.org/gtk3/stable/GtkCssProvider.html#gtk-css-provider-get-named

alacritty_terminal/src/config/window.rs
bkoropoff: YAML has an explicit null (`null`).

alacritty_terminal/src/config/window.rs
bkoropoff: The official X11 structure is:

```c
typedef struct {
    char *res_name;
    char *res_class; 
} XClassHint; 
```

So I could change `instance` to `name`, but the class/class nomenclature is standard.

alacritty_terminal/src/config/window.rs
chrisduerr: Yes, I was just trying to point out that we should not accept this explicit null. It should be treated like any other invalid value (like boolean, object, or whatever).

alacritty_terminal/src/config/window.rs
chrisduerr: I'm aware that by default this is just the 'window class', however if we want to have a reasonable configuration file and be able to communicate reasonably about our own code, we need to find good words to identify the struct and the struct's fields.

I think `instance` is great already, this is often what it's referred to in the documentation too (we don't need to stick to the exact struct). So the only remaining question is struct name and general class.

I think naming the struct `XClassHint would be more confusing than helpful, we should name this closely similar to the config field and I think the most reasonable for the user there would be the `class` field identifier. So we need to find a good name for the 'opposite' of instance. And that's what I'd propose `general` for. The general class is used for all instances and that's the most reasonable naming I could come up with looking at the documentation of X.

bkoropoff:
@chrisduerr I reworked the deserialization and fixed some of the other issues.  The only unresolved issue is final naming on the config keys, which I'm not particularly invested in.  Let me know what you prefer.

alacritty.yml
```diff
@@ -63,10 +63,20 @@ window:
...
+  class:
+    # Application instance name
+    name: Alacritty
+    # General application class
+    class: Alacritty
+
+  # Override GTK theme variant (Linux only).  "None" does not override default.
+  # Most themes support the following variants:
+  #   - light
+  #   - dark
```
chrisduerr: This should be formatted to match the other configuration options.

Something like this should be better:

```yml
# GTK theme variant (Linux only)
# 
# Override the variant of the GTK theme. Commonly supported values are `dark` and `light`.
# Set this to `None` to use the default theme variant.
```

alacritty/src/cli.rs
chrisduerr: It seems like the completions currently do not support --class at all, do you want to include that in this PR or should I open a separate issue?

alacritty_terminal/src/config/mod.rs
```diff
@@ -432,3 +413,97 @@ where
...
+    {
+        Deserialize::deserialize(MapAccessDeserializer::new(map))
+    }
+}
+
+pub fn struct_or_str<'de, T, D>(deserializer: D) -> Result<T, D::Error>
+where D: Deserializer<'de>, T: Deserialize<'de> + FromStr<Err = Void>
+{
+    deserializer.deserialize_any(OrStrVisitor(PhantomData))
+}
```
chrisduerr: This deserializer seems unnecessarily complex. Is there a reason why you did not follow my recommendation to just deserialize using a generic type parameter?

alacritty_terminal/src/config/window.rs
```diff
@@ -117,3 +123,25 @@ impl Dimensions {
         self.columns.0 as u32
     }
 }
+
+/// Window class hint
+#[serde(default)]
+#[derive(Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct ClassHint {
```
chrisduerr: If you do not care, then I'd name this struct `Class` and the fields `instance` and `general`.

extra/alacritty.man
```diff
@@ -39,8 +39,8 @@ Increases the level of verbosity (the max level is \fB\-vvv\fR)
 Prints version information
 .SH "OPTIONS"
 .TP
-\fB\-\-class\fR <class>
-Defines the window class on Linux [default: Alacritty]
+\fB\-\-class\fR [ <name> | <name>,<class> ]
+Defines the window class hint on Linux [default: Alacritty]
```
chrisduerr: This default should be adjusted to show that both instance and general class are set to Alacritty.

alacritty_terminal/src/config/mod.rs
bkoropoff: I'm not super familiar with serde, but using a custom visitor seems to be the right way to 
intermediate the deserialization process.  This particular use is actually one of the documentation examples: https://serde.rs/string-or-struct.html

`serde_yaml::Value` doesn't implement `IntoDeserializer`, so in order to feed it back into deserialization after handling the special case I'd need to re-serialize it.  This avoids the intermediate data structure.  The same applies to the `or_none` helper.

chrisduerr: You can just deserialize it to Value and then use T::deserialize(value) to deserialize it from the value to the T, as long as T implements deserialize. The method shouldn't look much different than the working directory deserializer. You can look at failure_default as an example.

bkoropoff: It looks like that would work, but this method is more efficient since it doesn't create the intermediate object tree.

chrisduerr: It's also unmaintainable. The performance for the config file deserialization is not an issue. Maintaining the code for it is a lot more difficult though.

bkoropoff:
@chrisduerr OK, I simplified the deserialization code by using intermediate `serde_yaml::Value`s.  Interestingly, the existing `failure_default` didn't actually need it, so I removed it there.

alacritty/src/cli.rs
bkoropoff: I'd rather it be separate.  There's nothing in particular to complete for the option argument, anyway (it's completely arbitrary).

bkoropoff:
@chrisduerr fixed

alacritty.yml
```diff
@@ -66,7 +66,17 @@ window:
   #title: Alacritty
 
   # Window class (Linux only):
-  #class: Alacritty
+  class:
+    # Application instance name
+    name: Alacritty
+    # General application class
+    class: Alacritty
```
chrisduerr: This field should be named `general`.

alacritty/src/cli.rs
chrisduerr: Created https://github.com/jwilm/alacritty/issues/2612 to track this.

alacritty_terminal/src/config/mod.rs
```diff
@@ -134,8 +139,8 @@ pub struct Config {
     alt_send_esc: DefaultTrueBool,
 
     /// Shell startup directory
-    #[serde(default, deserialize_with = "failure_default")]
-    working_directory: WorkingDirectory,
+    #[serde(default, deserialize_with = "or_none")]
```
chrisduerr: I think we can do better than naming this `or_none`, which doesn't actually say a lot about what it does.

Something like `explicit_option` or `allow_none_option` might be more descriptive here. That should make it clear that we're only deserializing options here, not just any value or none.

alacritty_terminal/src/config/mod.rs
```diff
@@ -418,17 +398,35 @@ impl Default for DefaultTrueBool {
...
+pub fn or_none<'de, T, D>(deserializer: D) -> Result<Option<T>, D::Error>
+    where D: Deserializer<'de>, T: Deserialize<'de> + Default
+{
+    Ok(match Value::deserialize(deserializer)? {
+        Value::String(ref value) if value.to_lowercase() == "none" => None,
+        value => Some(T::deserialize(value).unwrap_or_else(defaulted))
+    })
+}
+
+pub fn struct_or_str<'de, T, D>(deserializer: D) -> Result<T, D::Error>
```
chrisduerr: If we want to make this generic, we should use it for the shell too. Otherwise it should be renamed and moved closer to where it's actually used.

alacritty_terminal/src/config/mod.rs
```diff
@@ -418,17 +398,35 @@ impl Default for DefaultTrueBool {
...
+where D: Deserializer<'de>, T: Deserialize<'de> + FromStr<Err = Void> + Default
```
chrisduerr: There's no need to use `Void` here, especially since it's an external dependency for something extremely trivial. Instead we should just use `Err = ()`, which is more idiomatic Rust anyways.

alacritty_terminal/src/window.rs
```diff
@@ -267,23 +269,29 @@ impl Window {
...
-            .with_class(class.into(), DEFAULT_NAME.into())
+            .with_class(instance.into(), class.into())
             // Wayland
-            .with_app_id(class.into())
+            .with_app_id(instance.into());
+
+        if let Some(ref val) = window_config.gtk_theme_variant {
+            builder = builder.with_gtk_theme_variant(val.clone())
+        }
+        builder
```
chrisduerr: I'd personally prefer an empty line before `builder` too, but that's me being extremely pedantic. I do think that it would look nicer though.

extra/alacritty.man
```diff
@@ -39,8 +39,8 @@ Increases the level of verbosity (the max level is \fB\-vvv\fR)
 Prints version information
 .SH "OPTIONS"
 .TP
-\fB\-\-class\fR <class>
-Defines the window class on Linux [default: Alacritty]
+\fB\-\-class\fR [ <name> | <name>,<class> ]
```
chrisduerr: This should be named with `[ <instance> | <instance>,<general> ]` to follow the rest of the naming conventions.

alacritty_terminal/src/config/mod.rs
bkoropoff: `Void` isn't unit, but an uninhabited type.  This means an error is impossible, which is why it's safe to call `unwrap`.  The `Void` type is just `enum Void{}`, but the equality constraint presumably requires equality of types.  The `void` crate ensures everyone uses the same type.

alacritty_terminal/src/window.rs
bkoropoff: I forgot about that one.

alacritty_terminal/src/config/mod.rs
chrisduerr: Requiring an external dependency for something so simple is still pretty fragile. In theory this should be done with the never type, though that's not available yet.

However, in this example we could just require `From<&str>` or `From<String>`. This can never throw an error and the unwrap is actually safe since there is none (unlike Void which is vulnerable to refactorings).

alacritty_terminal/src/config/mod.rs
bkoropoff: It won't work for the shell because of the `Option`, which will require a different function (without the `Default` bound, and defaulting to `None` on type error)

chrisduerr: It might be possible to get rid of the option for shell and use default instead, but if we don't want to use it we should change the name at least.

bkoropoff: I added a second function to handle the option case.  They'll probably be generically useful for future settings where you want to have an abbreviated string form or a more detailed full form.

bkoropoff:
@chrisduerr I think that's everything.  Probably.

alacritty_terminal/src/config/mod.rs
```diff
@@ -418,17 +401,50 @@ impl Default for DefaultTrueBool {
...
-        Ok(value) => Ok(value),
-        Err(err) => {
-            error!("Problem with config: {}; using default value", err);
-            Ok(T::default())
-        },
-    }
+    Ok(T::deserialize(Value::deserialize(deserializer)?).unwrap_or_else(defaulted))
+}
+
+pub fn none_str_as_none<'de, T, D>(deserializer: D) -> Result<Option<T>, D::Error>
```
chrisduerr: I think the naming of this deserializer is still a bit unfortunate. The name would suggest to me that it just deserializes the `none` string to `None` and that's it. It doesn't really encode that it parses anything else generically too and None is just an option.

Maybe something like `deserialize_option_explicit_none` might be better?

alacritty_terminal/src/config/mod.rs
```diff
@@ -418,17 +401,50 @@ impl Default for DefaultTrueBool {
...
+    where D: Deserializer<'de>, T: Deserialize<'de> + Default
+{
+    Ok(match Value::deserialize(deserializer)? {
+        Value::String(ref value) if value.to_lowercase() == "none" => None,
+        value => Some(T::deserialize(value).unwrap_or_else(defaulted))
+    })
+}
+
+pub fn abbrev_from_str<'de, T, D>(deserializer: D) -> Result<T, D::Error>
```
chrisduerr: I think the naming of this is also a bit troublesome. I wouldn't know what `abbrev_from_str` does at all for example. Maybe `deserialize_string_or_generic` might be better? The `or_generic` part is kinda ugly, but I can't think of anything better myself right now.

alacritty_terminal/src/config/mod.rs
```diff
@@ -418,17 +401,50 @@ impl Default for DefaultTrueBool {
     }
 }
 
+fn defaulted<T, E>(err: E) -> T
```
chrisduerr: I think `defaulted` is a bit weird for a Rust function name. I'd expect `default` to be the name, not `defaulted`. Of course it does more than just picking the default, but I'm not sure if there's a good way to encode that in its name. Maybe `fallback_default` could encode it in a way that makes it clear that it's not just normal default, but an error case?

alacritty_terminal/src/config/mod.rs
```diff
@@ -418,17 +401,50 @@ impl Default for DefaultTrueBool {
...
+        Value::String(value) => Some(T::from(value)),
+        value => match T::deserialize(value) {
+            Ok(value) => Some(value),
+            Err(err) => {
+                error!("Problem with config: {}; using None", err);
+                None
+            }
+        }
+    })
 }
```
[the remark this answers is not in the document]
bkoropoff: Rust's trait coherence rules do not allow it since both `From` and `Option` are not defined in this crate.

alacritty_terminal/src/config/mod.rs
bkoropoff: It follows the same format as `failure_default`: `<condition>_<result>`.

alacritty_terminal/src/config/mod.rs
bkoropoff: I could go back to `struct_or_str`, which is more likely to be recognized since it's the name used in the official serde documentation (although that version enforces the requirement that it be a struct if it's not a string).

alacritty_terminal/src/config/mod.rs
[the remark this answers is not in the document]
chrisduerr: Yeah that's true, but we could still just write a thin wrapper around Option<Shell> and that would allow us to implement it.

I still think that just implementing Default for shell would be the easiest solution here, but allowing all fields to be optional might be an alternative to write a completely separate type wrapping around Option<Shell>.

alacritty_terminal/src/config/mod.rs
chrisduerr: Failure default wasn't really written with that in mind though. It's just a good way to describe all things going on with this deserializer. In case of failure, go with the default.

None str as none does not do that. It's easy to confuse the name and think it might do other things than it really does. The double-use of 'none' also is confusing to read and it's hard to even understand exactly what is meant without reading it multiple times. Using `str`, which is also a Rust type is additionally confusing, since that does not actually exist for serde/yaml.

alacritty_terminal/src/config/mod.rs
chrisduerr: Generally `str` is not a good abbreviation here, since it's easily confused with the Rust type. Which is probably why the serde documentation example also uses `string_or_struct`. I believe the yaml type is also called `String`.

Generally for one-off deserializers, I've also added the `deserialize_` at the beginning to make clear what they are for. `failure_default` is an exception here simply because it's used everywhere and previously there was an additional function which has been relplaced with `#[serde(default)]` which is why a shorter name was beneficial.

Having `struct` in the name is also misleading. This is in no way limited to structs, so the function name would actually be completely incorrect. We either deserialize from a serde string, or from whatever other type the field has which implements deserialize. This is why I proposed `or_generic` instead of `or_struct`.

Generally another problem that could be mentioned is that `string` does not explicitly mention that its about the origin and not the target type (neither in my example nor yours). So `deserialize_string_or_generic` could be easily mistaken for a deserializer which deserializes either a field with the value string, or a generic value.

I'll happily admit that I'm not really happy with any solution proposed so far, neither by you nor me. I'm not sure if there is a perfect solution either, so the better choice might be to just settle for something reasonable.

Considering most of the points I've made, maybe `from_string_or_deserialize` could be an option? It's a quite literal way to translate what exactly is happening, so it might be a relatively uncontroversial solution that still represents things accurately. It doesn't really say that from string is also deserialization, but I guess you can't win them all.

bkoropoff: `from_string_or_deserialize` sounds pretty good

alacritty_terminal/src/config/mod.rs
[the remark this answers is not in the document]
bkoropoff: I'm inclined to leave it like this since both functions are generically useful.  Future settings may not have a meaningful `Default` but can still benefit from an shorthand string form.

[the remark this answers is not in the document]
chrisduerr: There's another option, which is creating a custom `FromString` trait. This allows us to implement it for any type, including wrappers from std like `Arc` or `Option`.

Since this works very well, I've decided to go for this and have made the required changes myself. I think that's a good solution for the problem and it removes the need for two functions which are extremely similar.

jazzy57:
Wondering if this was ever released to the pkg on Arch LInux? Trying to set this in my conf but having no effect. On version 0.4.0 of alacritty.

chrisduerr:
> Wondering if this was ever released to the pkg on Arch LInux? Trying to set this in my conf but having no effect. On version 0.4.0 of alacritty.

It is part of the 0.4.0 release, please check the alacritty.yml file in the 0.4.0 release on github to see the latest version of the configuration file. Your config will not be rewritten for you, we do not touch the file once it has been created.

jazzy57:
I'm aware, the new config doesn't seem to be working.  Did a diff from the file on master and I don't seem to be missing anything.  Here is my `window` section.

```
window:
  # Window dimensions (changes require restart)
  #
  # Specified in number of columns/lines, not pixels.
  # If both are `0`, this setting is ignored.
  dimensions:
    columns: 0
    lines: 0

  # Window position (changes require restart)
  #
  # Specified in number of pixels.
  # If the position is not set, the window manager will handle the placement.
  #position:
  #  x: 0
  #  y: 0

  # Window padding (changes require restart)
  #
  # Blank space added around the window in pixels. This padding is scaled
  # by DPI and the specified value is always added at both opposing sides.
  padding:
    x: 0
    y: 0

  # Spread additional padding evenly around the terminal content.
  dynamic_padding: false

  # Window decorations
  #
  # Values for `decorations`:
  #
  # Values for `decorations` (macOS only):
  #     - transparent: Title bar, transparent background and title bar buttons
  #     - buttonless: Title bar, transparent background, but no title bar buttons
  decorations: full

  # Startup Mode (changes require restart)
  #
  # Values for `startup_mode`:
  #   - Windowed
  #   - Maximized
  #   - Fullscreen
  #
  # Values for `startup_mode` (macOS only):
  #   - SimpleFullscreen
  startup_mode: Windowed

  # Window title
  #title: Alacritty

  #Window class (Linux only):
  #class: Alacritty
  class:
    instance: Alacritty
    general: Alacritty
  gtk_theme_variant: dark
```

jazzy57:
Running Gnome Shell 3.34.1 and have dark mode turned on via gnome-tweak-tools.

chrisduerr:
@jazzy57 Are you running Wayland?

jazzy57:
Yep

chrisduerr:
I don't think such a thing exists on Wayland.

jazzy57:
Newer versions of Gnome use Wayland and not X by default. Other apps seem to have the dark title bar.

chrisduerr:
Of course apps running GTK will have your GTK theme. Alacritty is not a GTK application and you give up the ability to have uniform window decorations when you choose to install Gnome Wayland.

There's nothing any application can do about this. Gnome makes this decision for you and by choosing gnome, you agree with that decision. If you do not like that behavior, please do not choose gnome.

jazzy57:
Sounds good didn't realize this was the case, will give away another try soon.

teohhanhui:
> Of course apps running GTK will have your GTK theme. Alacritty is not a GTK application and you give up the ability to have uniform window decorations when you choose to install Gnome Wayland.
> 
> There's nothing any application can do about this. Gnome makes this decision for you and by choosing gnome, you agree with that decision. If you do not like that behavior, please do not choose gnome.

That's a completely anti-choice and user hostile position. Gnome is a major desktop environment, and Wayland is the now and the future. It's the app developer's responsibility to abide to platform guidelines. Forcing the user to choose between the two is a false dichotomy.

chrisduerr:
If you dislike that, you're free to write your own Alacritty frontend in GTK.

kchibisov:
>on. Gnome is a major desktop environment, and Wayland is the now and the future.

KDE and most wlroots based compositors have server side decorations. Server side decorations is a standard thing on Wayland.The position of GNOME is not to support SSD in any shape. don't ping us, ping them, but they won't change their decision either. But yeah, just write your frontend with gtk if you like it so much.

>  It's the app developer's responsibility to abide to platform guidelines.

What guidelines? There are a lot of ton specs, gui toolkits, windowing libs, protocols, etc. There's no standard, nor guidelines. Wayland says that you should handle both CSD and SSD, we handle both of them. Does it says that it should be GTK themed? No, it doesn't. 

>Forcing the user to choose between the two is a false dichotomy.

GNOME doesn't provide you a choice to use CSD or SSD. Everyone else is trying to do so.

PR #2611, opened 2019-07-04
Use Menlo as fallback font on macOS

### LINKED ISSUE #2585: Crashes on startup MacOS Mojave

MacOS Mojave 10.14.5

What version of Alacritty?

Alacritty-v0.3.3 downloaded from the releases tab. I tested it with some previous versions and they had it as well. 

Behavior:

Alacritty crashes on startup.

Context:
I believe the issue is related to finding the font family because this happens when I set the normal font in my config to something that does not exist (in the example below the correct font name should be "Fira Code"):
```
# Font configuration (changes require restart)
font:
  # Normal (roman) font face
  normal:
    # Font family
    #
    # Default:
    #   - (macOS) Menlo
    #   - (Linux) monospace
    #   - (Windows) Consolas
    family: "FiraCode"
```
Alacritty behaves normally when provided with an existing font family (e.g. "Fira Code"). 

I attached my config file for more information.
[config.txt](https://github.com/jwilm/alacritty/files/3318843/config.txt)

nangtrongvuon:
This commit makes alacritty use Menlo as a fallback font on macOS if the config specified font family isn't found.

closes #2585

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The window is now filled with the background color before displaying
 - Cells sometimes not getting cleared correctly
 - X11 clipboard hanging when mime type is set
+- On macOS, Alacritty will now fallback to Menlo if a font specified in the config cannot be loaded
 
 ## 0.3.3
 
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -314,10 +314,11 @@ pub fn descriptors_for_family(family: &str) -> Vec<Descriptor> {
     let mut out = Vec::new();
 
     trace!("Family: {}", family);
-    let ct_collection = match create_for_family(family) {
-        Some(c) => c,
-        None => return out,
-    };
+    let ct_collection = create_for_family(family).unwrap_or_else(|| {
+        // Fallback to Menlo if we can't find the config specified font family.
+        warn!("Unable to load specified font {}, falling back to Menlo", &family);
+        create_for_family("Menlo").expect("Menlo exists")
+    });
 
     // CFArray of CTFontDescriptorRef (i think)
     let descriptors = ct_collection.get_descriptors();


font/src/darwin/mod.rs
```diff
@@ -316,7 +316,14 @@ pub fn descriptors_for_family(family: &str) -> Vec<Descriptor> {
...
-        None => return out,
+        None => {
+            // Fallback to Menlo if we can't find the config specified font family.
+            if let Some(menlo_fallback) = create_for_family("Menlo") {
+                menlo_fallback
+            } else {
+                return out;
+            }
+        }
     };
```
chrisduerr: I think it might be a bit nicer if we use `unwrap_or_else` here, since we don't want to change the `Some` case anyways.

Then, since this should never happen, we should probably also print a warning when the font could not be loaded using the `warn!` macro. This will lead to a warning in the Alacritty UI too, so hopefully this would make users change their config.

nangtrongvuon: Do you mean the fallback `Some` case? Or do you mean `ct_collection`'s `Some` case?

chrisduerr: I mean that we should do `create_from_family(family).unwrap_or_else(|| menlo.expect(...))`.

One thing I think I forgot to mention in the review is that we probably want to crash hard if menlo load isn't successful, since that's a broken system we don't need to support. Using `expect` for that should allow us to crash with a reasonable message.

nangtrongvuon:
@chrisduerr I updated the PR with your feedback, does it look alright now?

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The window is now filled with the background color before displaying
 - Cells sometimes not getting cleared correctly
 - X11 clipboard hanging when mime type is set
+- On macOS, Alacritty will now fallback to Menlo if a font specified in the config cannot be loaded.
```
chrisduerr: There should be no punctuation at the end of change log entries.

nangtrongvuon: Done.

nangtrongvuon:
Thank you for your patience, one final check?

nangtrongvuon:
Thanks for your work on alacritty!

PR #2614, opened 2019-07-06
Change mouse cursor when hovering over the message bar

CrackedP0t:
This PR sets the mouse cursor to an arrow when over the message bar, and to a pointer when over the close button. This indicates that the message bar isn't part of the regular terminal, and that the close button can be clicked on and closed.

Feedback welcome!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow setting general window class on X11 using CLI or config (`window.class.general`)
 - Config option `window.gtk_theme_variant` to set GTK theme variant
 - Completions for `--class` and `-t` (short title)
-- Change the mouse cursor when mousing over the message bar and its close button
+- Change the mouse cursor when hovering over the message bar and its close button
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow setting general window class on X11 using CLI or config (`window.class.general`)
 - Config option `window.gtk_theme_variant` to set GTK theme variant
 - Completions for `--class` and `-t` (short title)
+- Change the mouse cursor when mousing over the message bar and its close button
 
 ### Fixed
 
--- a/alacritty_terminal/src/input.rs
+++ b/alacritty_terminal/src/input.rs
@@ -946,7 +946,6 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         let size = self.ctx.size_info();
         point.col + message_bar::CLOSE_BUTTON_TEXT.len() >= size.cols()
             && point.line == size.lines() - message.text(&size).len()
-
     }
 
     /// Handle clicks on the message bar.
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Block selection mode when Control is held while starting a selection
+- Change the mouse cursor when mousing over the message bar and its close button
```
```suggestion
- Change the mouse cursor when hovering over the message bar and its close button
```

alacritty_terminal/src/input.rs
```diff
@@ -414,9 +414,17 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
             prev_line != self.ctx.mouse().line || prev_col != self.ctx.mouse().column;
         let mouse_moved = cell_changed || prev_side != cell_side;
 
+        if !mouse_moved {
```
chrisduerr: I think we can just inline the `mouse_moved` code here.

alacritty_terminal/src/input.rs
```diff
@@ -910,6 +918,15 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         has_binding
     }
 
+    /// Set the cursor depending on where the mouse is on the message bar
+    fn message_cursor(&mut self, point: Point, message: Message) {
```
chrisduerr: The name `message_cursor` sounds like a Rust-style getter. Just renaming it to `update_message_cursor` would probably be fine here. Using `set_message_cursor` wouldn't be ideal since it's not actually a setter, so it doesn't take a mouse cursor as parameter.

alacritty_terminal/src/input.rs
```diff
@@ -414,9 +414,17 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+        }
+
         // Only report motions when cell changed and mouse is not over the message bar
-        if self.message_at_point(Some(point)).is_some() || !mouse_moved {
+        if let Some(message) = self.message_at_point(Some(point)) {
+            self.message_cursor(point, message);
+
             return;
+        } else {
+            self.ctx.terminal_mut().reset_mouse_cursor();
```
chrisduerr: I think this might be a bit problematic. Setting the cursor regardless of its current state whenever the cell is changed (or cell side is changed) makes things a bit better, but it could still call winit's mouse cursor function potentially hundreds of times.

It might be worthwhile to keep track of the current state and only call the winit function when it changes. This would be significantly easier on the #2438 PR though since that allows direct access to the window from the input.rs file.

alacritty_terminal/src/input.rs
CrackedP0t: Sounds good!

alacritty_terminal/src/input.rs
CrackedP0t: Yeah, that makes sense

alacritty_terminal/src/input.rs
CrackedP0t: Alright, I added a field to `Term` to track the mouse cursor and only set `next_mouse_cursor` if it changes. Does that seem like a good solution?

chrisduerr: Generally the approach seems good, but I feel like the terminal is the wrong place for this.

The `Window`'s `set_mouse_cursor` is always called when a mouse cursor is set, so that's probably the better place. It just makes more sense for the window to track the mouse cursor state instead of the terminal.

CrackedP0t: Yeah, that definitely makes sense.

Issue #2617, opened 2019-07-06
How can i change the cursor color in insert


taigacute:
the color of beam same as block. how can i change it?

kchibisov:
@taigacute can you elaborate? Do you want different colors for the different cursors?

taigacute:
@kchibisov   use neovim .  i think the different color between normal mode and insert mode is good.
now . i set the cursor color yellow . when use neovim .  i want cursor color is green in insert mode
normal
![1](https://user-images.githubusercontent.com/41671631/60754465-e1862b00-a013-11e9-811a-2447ff453af6.png)
insert
![2](https://user-images.githubusercontent.com/41671631/60754467-e6e37580-a013-11e9-8ff9-5ce13196a572.png)

kchibisov:
@taigacute Neovim can handle this by default `:help guicursor`. However it seems like the cursor color changing with escape sequence is broken atm.

@chrisduerr This command changes the cursor on `alacritty v0.3.0` `printf '\033]12;#333333\007'` , but doesn't work at all on the latest master. So apps like `neovim` can't change cursor color anymore. I guess this was introduced by cursor rework.

taigacute:
@kchibisov  thanks for your reply. got it.

chrisduerr:
Bisect might be useful on this, I suspect that you're right @kchibisov and this is about the cursor change, however the parsing for this escape has also changed recently. But I think that is less likely to be the problem.

PR #2619, opened 2019-07-06
Fix cursor color setting with escape sequence

kchibisov:
The cursor rework introduced a regression where cursor color was always picked
from a config file, rather then using `ansi::NamedColor::Cursor` and
`ansi::NamedColor::CursorText` for this purposes.

Fixes:  #2617

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -447,8 +447,8 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
 
                     renderable_cell.inner = RenderableCellContent::Cursor(cursor_key);
 
-                    if let Some(color) = self.config.colors.cursor.cursor {
-                        renderable_cell.fg = color;
+                    if let Some(color) = self.config.cursor_cursor_color() {
+                        renderable_cell.fg = RenderableCell::compute_bg_rgb(self.colors, color);
                     }
 
                     return Some(renderable_cell);
@@ -459,7 +459,7 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
                     if self.cursor_style == CursorStyle::Block {
                         std::mem::swap(&mut cell.bg, &mut cell.fg);
 
-                        if let Some(color) = self.config.colors.cursor.text {
+                        if let Some(color) = self.config.cursor_text_color() {
                             cell.fg = color;
                         }
                     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - X11 clipboard hanging when mime type is set
 - On macOS, Alacritty will now fallback to Menlo if a font specified in the config cannot be loaded
 - Debug ref tests are now written to disk regardless of shutdown method
+- Cursor color setting with escape sequence
 
 ## 0.3.3
 
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -539,8 +539,6 @@ pub enum NamedColor {
     Background,
     /// Color for the cursor itself
     Cursor,
-    /// Color for the text under the cursor
-    CursorText,
     /// Dim black
     DimBlack,
     /// Dim red
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -32,7 +32,7 @@ mod test;
 mod visual_bell;
 mod window;
 
-use crate::ansi::CursorStyle;
+use crate::ansi::{Color, CursorStyle, NamedColor};
 use crate::input::{Binding, KeyBinding, MouseBinding};
 
 pub use crate::config::bindings::Key;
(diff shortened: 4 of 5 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -447,10 +447,14 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
...
                     renderable_cell.inner = RenderableCellContent::Cursor(cursor_key);
 
-                    if let Some(color) = self.config.colors.cursor.cursor {
-                        renderable_cell.fg = color;
+                    if self.config.colors.cursor.cursor.is_some() {
+                        // `ansi::NamedColor::Cursor` is in sync with `cursor.cursor` color from a
+                        // config, but we can change cursor color with escape sequence during
+                        // runtime.
+                        renderable_cell.fg = *self.colors.index(ansi::NamedColor::Cursor);
                     }
```
chrisduerr: Did the previous implementation behave the same way and only apply the cursor colors when it is fixed in the config?

It would be possible to always set it to a fixed color once it's set to a specific color once, even if it was just inversed before, but I'd mimic the previous behavior here. Not sure what other terminals do either, but I'd think usually they'd probably try to respect the change cursor color escape.

alacritty_terminal/src/term/mod.rs
```diff
@@ -447,10 +447,14 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
...
 
+
```
chrisduerr: Looks like a duplicate whitespace slipped in here.

alacritty_terminal/src/term/mod.rs
```diff
@@ -459,8 +463,10 @@ impl<'a> Iterator for RenderableCellsIter<'a> {
                     if self.cursor_style == CursorStyle::Block {
                         std::mem::swap(&mut cell.bg, &mut cell.fg);
 
-                        if let Some(color) = self.config.colors.cursor.text {
-                            cell.fg = color;
+                        if self.config.colors.cursor.text.is_some() {
+                            // `ansi::NamedColor::CursorText` is in sync with `cursor.text` color
+                            // from a config.
+                            cell.fg = *self.colors.index(ansi::NamedColor::CursorText);
                         }
```
chrisduerr: Is there any way to change this? Or are we just storing it in our colors without any real benefit?

kchibisov: I didn't find anyway to change it, and the `alacritty_terminal/src/ansi.rs` doesn't have any handler for it atm. Just searching the web result in nothing and `nvim` can change only background color, and not foreground one. So, It seems to me like there is no way to change it. However we can introduce some `proprietary` sequence to change it.

alacritty_terminal/src/term/mod.rs
kchibisov: > Did the previous implementation behave the same way and only apply the cursor colors when it is fixed in the config?

Yes.

alacritty_terminal/src/term/mod.rs
chrisduerr: I'd be surprised if there was a way to change it and I'm not familiar with it. I'm strongly against introducing a proprietary escape, but we might be able to remove it from colors if we don't need it there.

It's a bit unergonomic to have to load things from two different places. We should just be able to `let Some` it and be good with it.

kchibisov: > It's a bit unergonomic to have to load things from two different places. We should just be able to let Some it and be good with it.

You mean returning old cursor colors supporting methods ( You've removed them during rework)?

chrisduerr: I mean it's unergonomic to check the config colors and then load it from a completely different place, even though they're always the exact same. For someone just looking at that part of the code that makes absolutely no sense at all.

I think I was actually able to change it from the old way to simplify things a bit more. Generally it seems like the `NamedColor::CursorText` just isn't necessary anymore. If this is the only place where we're using it, I think it might be cleaner just to remove it.

kchibisov: > I mean it's unergonomic to check the config colors and then load it from a completely different place, even though they're always the exact same.

I've understood this in the first place, I mean, you've removed the following method on purpose.
```rust
impl Config {
...
/// Cursor background color                                             
#[inline]                                                              
pub fn cursor_cursor_color(&self) -> Option<Color> {                   
    self.colors.cursor.cursor.map(|_| Color::Named(NamedColor::Cursor))
}
...                                                                      
}
```
So, I want to bring it back. So cursor color setting will change to:
```rust
if let Some(cursor) = self.config.cursor_cursor_color() {
    renderable_cell.fg = RenderableCell::compute_bg_rgb(self.colors, cursor);
}
```

I think removing `NamedColor::CursorText` is a good idea. So, should I add the following method to the `struct Config` to mimic method suggested above, or access it directly, like it's now?

```rust
#[inline]
pub fn cursor_text_color(&self) -> Option<Rgb> {                   
    self.colors.cursor.text
}
```

chrisduerr: Sure, looks good to me. One concern might be that one of these is returning a Color and the other Rgb, but I personally wouldn't worry about that too much.

alacritty_terminal/src/term/mod.rs
chrisduerr: Sounds good to me then.

Issue #2612, opened 2019-07-04
Add --class to shell completions


chrisduerr:
The shell completions currently do not support the `--class` flag. This should be added as an arbitrary text variable.

PR #2621, opened 2019-07-07
Update shell completions

kchibisov:
Add `--class` and `-t` options to shell completions.

Fixes: #2612


kchibisov:
@chrisduerr The changes for shell completions were never documented (even adding them), so this is why I've skipped it entirely.

kchibisov:
@chrisduerr Anyway, if change log entry is required here, should it go to `Fixed` or `Added` section?

chrisduerr:
@kchibisov Ah, that's a reasonable explanation, however I think that's a mistake. Generally shell completions and terminfo seem related enough to source code to document changes and fixes, especially to inform people that they might want to inform them.

So I'd add it to the `Added` section. Chances are the fixed section is going to be the biggest most of the time anyways, so if anything's right on the edge it probably makes sense to slim down the fixes. So I'd just say this added completions for --class and -t (short title).

PR #2623, opened 2019-07-07
Fix saving of ref tests

chrisduerr:
Since ref tests were only stored whenever winit requested the window
close, they would not get stored properly when the terminal was closed
through Alacritty using `exit`, Ctrl+D or similar.

This moves the ref test code to the and of the main entry point, which
will always be executed regardless of how the terminal was shutdown.


chrisduerr:
Created from incorrect branch, see https://github.com/jwilm/alacritty/pull/2624.

PR #2624, opened 2019-07-07
Fix saving of ref tests

chrisduerr:
Since ref tests were only stored whenever winit requested the window
close, they would not get stored properly when the terminal was closed
through Alacritty using `exit`, Ctrl+D or similar.

This moves the ref test code to the and of the main entry point, which
will always be executed regardless of how the terminal was shutdown.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Cells sometimes not getting cleared correctly
 - X11 clipboard hanging when mime type is set
 - On macOS, Alacritty will now fallback to Menlo if a font specified in the config cannot be loaded
+- Debug ref tests are now written to disk regardless of shutdown method
 
 ## 0.3.3
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,6 +32,7 @@ dependencies = [
  "env_logger 0.6.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustc_tools_util 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 1.0.39 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_yaml 0.8.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.42 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/alacritty/Cargo.toml
+++ b/alacritty/Cargo.toml
@@ -16,6 +16,7 @@ time = "0.1.40"
 env_logger = "0.6.0"
 crossbeam-channel = "0.3.8"
 serde_yaml = "0.8"
+serde_json = "1"
 
 [build-dependencies]
 rustc_tools_util = "0.2.0"
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -252,7 +252,7 @@ impl Options {
         config.window.title = self.title.or(config.window.title);
 
         if let Some(class) = self.class {
-            let parts : Vec<_> = class.split(',').collect();
+            let parts: Vec<_> = class.split(',').collect();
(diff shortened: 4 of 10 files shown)


kchibisov:
It sometime panics on close with `Ctrl+D` when launching with `alacritty --ref-test`. Probably some race, since in `release` mode it happens more seldom.
```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: EventsLoopClosed', src/libcore/result.rs:999:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:59
             at src/libstd/panicking.rs:197
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:211
   4: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:474
   5: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:381
   6: rust_begin_unwind
             at src/libstd/panicking.rs:308
   7: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
   8: core::result::unwrap_failed
             at /rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c/src/libcore/macros.rs:18
   9: core::result::Result<T,E>::unwrap
             at /rustc/a53f9df32fbb0b5f4382caaad8f1a46f36ea887c/src/libcore/result.rs:800
  10: winit::platform::platform::wayland::keyboard::init_keyboard::{{closure}}
             at /home/kchibisov/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.19.1/src/platform/linux/wayland/keyboard.rs:111
  11: smithay_client_toolkit::keyboard::implement_kbd::{{closure}}::{{closure}}::{{closure}}
             at /home/kchibisov/.cargo/registry/src/github.com-1ecc6299db9ec823/smithay-client-toolkit-0.4.6/src/keyboard/mod.rs:720

```

chrisduerr:
@kchibisov Can you reproduce this with master at all? It seems like a winit/glutin problem.

It would be possible to store these after the PTY event loop is down, however it shouldn't make a difference. You can still try it though:

```diff
diff --git a/alacritty/src/main.rs b/alacritty/src/main.rs
index b5d0629..17112a4 100644
--- a/alacritty/src/main.rs
+++ b/alacritty/src/main.rs
@@ -255,13 +255,13 @@ fn run(config: Config, message_buffer: MessageBuffer) -> Result<(), Box<dyn Erro
         }
     }
 
+    loop_tx.send(Msg::Shutdown).expect("Error sending shutdown to event loop");
+
     // Write ref tests to disk
     if config.debug.ref_test {
         write_ref_test_results(&terminal.lock());
     }
 
-    loop_tx.send(Msg::Shutdown).expect("Error sending shutdown to event loop");
-
     // FIXME patch notify library to have a shutdown method
     // config_reloader.join().ok();
```

The error kinda seems to me like there might have been some window events after the window has been closed? The mention of `keyboard.rs` makes me feel like this could be about window events after the event loop closed.

Note that I cannot reproduce this on X11, but it does seem like an issue in winit/glutin which should be present even before this change.

kchibisov:
@chrisduerr I've tested on master (debug mode), and there is no such issue. However, it wasn't perform `correct` saving on `Ctrl+D`.

chrisduerr:
@kchibisov Have you tried closing through window decorations (or close shortcut)? Pressing a bunch of keys while doing it is likely a good idea to try and repro it.

kchibisov:
@chrisduerr So, your diff doesn't work. 

This PR:
- If I close with `shortcut` ( I'm on sway, so no decorations with buttons),- fine.
- If I close with `shortcut` and then press key - fine
- If I close with `Ctrl-D` and not pressing any keys - fine
- If I close with `Ctrl-D` and then press something - panic

Master:
both 4 cases are fine, so the 4'th one can be broken on master too, since I'm not too fast. Seems like not our bug.

chrisduerr:
Yeah, seems like it should be unrelated to this PR. Possibly fixed in the Event Loop 2 rework though.

kchibisov:
@chrisduerr So, I'll open an issue, right after merge.

PR #2630, opened 2019-07-09
Fix row occ not set during new and reset

### LINKED ISSUE #2613: Regression with scrolling performance
While doing some tests with profile guided optimization, I've noticed that normal scrolling (the type that is done by `yes`), is extremely slow on the latest master.

I've tested this with the version I'm currently running (f002171) and there's no such performance issue there.

So somewhere in between 707f4293667d4d8564ab0afc13128463c2ee7fd5 and f002171 there must have been a change significantly impacting the performance of scrolling content inside of Alacritty.

My best bet would be https://github.com/jwilm/alacritty/commit/0815774cbfb9d1421b8077d5d9d641faf5f818c1, however I haven't bisected it yet, so I'll update this issue when I've got some additional information.

chrisduerr:
Fixes #2613.

### DIFF
--- a/alacritty_terminal/src/grid/row.rs
+++ b/alacritty_terminal/src/grid/row.rs
@@ -45,8 +45,12 @@ impl<T: PartialEq> PartialEq for Row<T> {
 }
 
 impl<T: Copy> Row<T> {
-    pub fn new(columns: Column, template: &T) -> Row<T> {
-        Row { inner: vec![*template; *columns], occ: 0 }
+    pub fn new(columns: Column, template: &T) -> Row<T>
+    where
+        T: GridCell,
+    {
+        let occ = if template.is_empty() { 0 } else { columns.0 };
+        Row { inner: vec![*template; columns.0], occ }
     }
 
     pub fn grow(&mut self, cols: Column, template: &T) {
@@ -70,7 +74,7 @@ impl<T: Copy> Row<T> {
         let index = new_row.iter().rposition(|c| !c.is_empty()).map(|i| i + 1).unwrap_or(0);
         new_row.truncate(index);
 
-        self.occ = min(self.occ, *cols);
+        self.occ = min(self.occ, cols.0);
 
         if new_row.is_empty() {
             None
@@ -80,11 +84,17 @@ impl<T: Copy> Row<T> {
     }
 
     /// Resets contents to the contents of `other`
-    pub fn reset(&mut self, other: &T) {
-        for item in &mut self.inner[..] {
-            *item = *other;
+    pub fn reset(&mut self, template: &T)
+    where
+        T: GridCell,
+    {
+        for item in &mut self.inner[..self.occ] {
+            *item = *template;
         }
-        self.occ = 0;
+
+        if template.is_empty() {
+            self.occ = 0;
+        }
     }
 }
 
@@ -116,13 +126,14 @@ impl<T> Row<T> {
     where
         T: GridCell,
     {
+        self.occ += vec.len();
         self.inner.append(vec);
-        self.occ = self.inner.iter().rposition(|c| !c.is_empty()).map(|i| i + 1).unwrap_or(0);
     }
 
     #[inline]
     pub fn append_front(&mut self, mut vec: Vec<T>) {
         self.occ += vec.len();
+
         vec.append(&mut self.inner);
         self.inner = vec;
     }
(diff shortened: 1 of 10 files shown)


alacritty_terminal/src/grid/row.rs
```diff
@@ -80,11 +84,17 @@ impl<T: Copy> Row<T> {
     }
 
     /// Resets contents to the contents of `other`
-    pub fn reset(&mut self, other: &T) {
-        for item in &mut self.inner[..] {
-            *item = *other;
+    pub fn reset(&mut self, template: &T)
```
nixpulvis: Nice, `other` was confusing.

chrisduerr: Yeah, we've used `template` everywhere else, so being consistent here makes a lot more sense.

alacritty_terminal/src/grid/row.rs
```diff
@@ -80,11 +84,17 @@ impl<T: Copy> Row<T> {
...
+    where
+        T: GridCell,
+    {
+        for item in &mut self.inner[..self.occ] {
+            *item = *template;
+        }
+
+        if template.is_empty() {
+            self.occ = 0;
```
nixpulvis: Sorry, I think I'm clearly not understanding the template quite right. But I'd expect the `self.occ` to be set to the template's length. Or something like that here.

alacritty_terminal/src/grid/row.rs
```diff
@@ -116,13 +126,14 @@ impl<T> Row<T> {
     where
         T: GridCell,
     {
+        self.occ += vec.len();
         self.inner.append(vec);
-        self.occ = self.inner.iter().rposition(|c| !c.is_empty()).map(|i| i + 1).unwrap_or(0);
```
nixpulvis: :racing_car: :checkered_flag:

alacritty_terminal/src/grid/row.rs
chrisduerr: The template is a single cell. So we're resetting every occupied cell with the template cell.

Now if the cell we're using to replace it is empty, we know that none of the cells are occupied anymore and we can set it to 0.

However if the cell is not empty, we know that all cells in a row might have changed, however they are all still occupied.

Originally I iterated over each cell and checked if it's empty, setting the `occ` to the last element which is not empty. But since all elements are the same (they are the template cell), this is an optimized way of doing the same.

alacritty_terminal/src/grid/storage.rs
```diff
@@ -310,8 +310,21 @@ impl<T> IndexMut<Line> for Storage<T> {
 mod test {
     use crate::grid::row::Row;
     use crate::grid::storage::Storage;
+    use crate::grid::GridCell;
     use crate::index::{Column, Line};
 
+    impl GridCell for char {
+        fn is_empty(&self) -> bool {
+            *self == ' ' || *self == '\t'
```
nixpulvis: Are these really the only two "empty" characters?

alacritty_terminal/src/grid/row.rs
nixpulvis: Ahhhhh, I understand. Setting all cells to an empty template collapses the whole thing to 0, otherwise it stays the same.

A template is always essentially length 1 or 0.

nixpulvis: That makes sense, thanks for clearing that up for me.

alacritty_terminal/src/grid/storage.rs
chrisduerr: Yes. Generally this isn't about empty text characters, but more about what Alacritty treats as empty characters. By default all characters are filled with `' '` if they're not filled with anything, so that's basically the most important thing that needs to be checked.

The tabs are basically just an addition to that. Since we set those ourselves with the tabstop logic, we can treat them as empty too, since often with ncurses they are just used to move the cursor around more efficiently.

Everything else are basically characters which come in from the PTY. These are explicitly set so we don't really want to treat them as empty. The method is more about finding cells which are unchanged, rather than cells which don't need to be rendered.

nixpulvis: Would a different name actually make more sense here then. Like `is_spacer` since a "spacer" sounds more like something we've added ourselves.

alacritty_terminal/src/grid/row.rs
kchibisov: Maybe add a comment is a good idea, since this is not something obvious?

alacritty_terminal/src/grid/storage.rs
chrisduerr: No, since we also want to check other things like foreground color. This is basically just to check if a cell can be ignored or assumed to be default, so `empty` is good here I think.

PR #2642, opened 2019-07-11
Override default bindings with subset mode match

### LINKED ISSUE #2641: Double binding - ScrollPageUp
When setting a binding like:
```
- { key: PageUp,   mods: Shift,   action: ScrollPageUp                 }
```
The ScrollPageUp gets triggered twice, and I miss lines in the buffer.

After a suggestion from irc, I removed the keybinding, and it works as expected. There might be an issue that register binding twice.

The issues appears on: Linux X11 using git master from 2019/07/05

chrisduerr:
Fixes #2641.

### DIFF
--- a/alacritty_terminal/src/input.rs
+++ b/alacritty_terminal/src/input.rs
@@ -152,17 +152,17 @@ impl<T: Eq> Binding<T> {
         // the most likely item to fail so prioritizing it here allows more
         // checks to be short circuited.
         self.trigger == *input
-            && self.mode_matches(mode)
-            && self.not_mode_matches(mode)
+            && mode.contains(self.mode)
+            && !self.notmode.intersects(mode)
             && self.mods_match(mods, relaxed)
     }
 
     #[inline]
     pub fn triggers_match(&self, binding: &Binding<T>) -> bool {
         self.trigger == binding.trigger
-            && self.mode == binding.mode
-            && self.notmode == binding.notmode
             && self.mods == binding.mods
+            && self.mode.contains(binding.mode)
+            && self.notmode.contains(binding.notmode)
     }
 }
 
@@ -171,16 +171,6 @@ impl<T> Binding<T> {
     #[inline]
     fn execute<A: ActionContext>(&self, ctx: &mut A, mouse_mode: bool) {
         self.action.execute(ctx, mouse_mode)
-    }
-
-    #[inline]
-    fn mode_matches(&self, mode: TermMode) -> bool {
-        self.mode.is_empty() || mode.intersects(self.mode)
-    }
-
-    #[inline]
-    fn not_mode_matches(&self, mode: TermMode) -> bool {
-        self.notmode.is_empty() || !mode.intersects(self.notmode)
     }
 
     /// Check that two mods descriptions for equivalence
@@ -995,6 +985,72 @@ mod tests {
     use super::{Action, Binding, Processor};
 
     const KEY: VirtualKeyCode = VirtualKeyCode::Key0;
+
+    type MockBinding = Binding<usize>;
+
+    impl Default for MockBinding {
+        fn default() -> Self {
+            Self {
+                mods: Default::default(),
+                action: Default::default(),
+                mode: Default::default(),
+                notmode: Default::default(),
+                trigger: Default::default(),
+            }
+        }
+    }
+
+    #[test]
+    fn binding_matches_itself() {
+        let binding = MockBinding::default();
+        let identical_binding = MockBinding::default();
+
+        assert!(binding.triggers_match(&identical_binding));
+        assert!(identical_binding.triggers_match(&binding));
+    }
+
+    #[test]
+    fn binding_matches_different_action() {
+        let binding = MockBinding::default();
+        let mut different_action = MockBinding::default();
+        different_action.action = Action::ClearHistory;
+
+        assert!(binding.triggers_match(&different_action));
+        assert!(different_action.triggers_match(&binding));
+    }
+
+    #[test]
+    fn subset_mode_binding_matches_superset() {
+        let mut superset_mode = MockBinding::default();
+        superset_mode.mode = TermMode::ALT_SCREEN | TermMode::INSERT | TermMode::ORIGIN;
+        let mut subset_mode = MockBinding::default();
+        subset_mode.mode = TermMode::ALT_SCREEN | TermMode::ORIGIN;
+
+        assert!(superset_mode.triggers_match(&subset_mode));
+        assert!(!subset_mode.triggers_match(&superset_mode));
+    }
+
+    #[test]
+    fn subset_notmode_binding_matches_superset() {
+        let mut superset_notmode = MockBinding::default();
+        superset_notmode.notmode = TermMode::ALT_SCREEN | TermMode::INSERT | TermMode::ORIGIN;
+        let mut subset_notmode = MockBinding::default();
+        subset_notmode.notmode = TermMode::ALT_SCREEN | TermMode::ORIGIN;
+
+        assert!(superset_notmode.triggers_match(&subset_notmode));
+        assert!(!subset_notmode.triggers_match(&superset_notmode));
+    }
+
+    #[test]
+    fn mods_binding_requires_strict_match() {
+        let mut superset_mods = MockBinding::default();
+        superset_mods.mods = ModifiersState { alt: true, logo: true, ctrl: true, shift: true };
+        let mut subset_mods = MockBinding::default();
+        subset_mods.mods = ModifiersState { alt: true, logo: false, ctrl: false, shift: false };
+
+        assert!(!superset_mods.triggers_match(&subset_mods));
+        assert!(!subset_mods.triggers_match(&superset_mods));
+    }
 
     #[derive(PartialEq)]
     enum MultiClick {
--- a/alacritty_terminal/src/input.rs
+++ b/alacritty_terminal/src/input.rs
@@ -153,7 +153,7 @@ impl<T: Eq> Binding<T> {
         // checks to be short circuited.
         self.trigger == *input
             && mode.contains(self.mode)
-            && !self.notmode.intersects(mode)
+            && !mode.intersects(self.notmode)
             && self.mods_match(mods, relaxed)
     }
 
@@ -161,8 +161,8 @@ impl<T: Eq> Binding<T> {
     pub fn triggers_match(&self, binding: &Binding<T>) -> bool {
         self.trigger == binding.trigger
             && self.mods == binding.mods
-            && self.mode.contains(binding.mode)
-            && self.notmode.contains(binding.notmode)
+            && (self.mode.contains(binding.mode) || binding.mode.contains(self.mode))
+            && (self.notmode.contains(binding.notmode) || binding.notmode.contains(self.notmode))
     }
 }
 
@@ -993,8 +993,8 @@ mod tests {
             Self {
                 mods: Default::default(),
                 action: Default::default(),
-                mode: Default::default(),
-                notmode: Default::default(),
+                mode: TermMode::empty(),
+                notmode: TermMode::empty(),
                 trigger: Default::default(),
             }
         }
@@ -1027,7 +1027,7 @@ mod tests {
         subset_mode.mode = TermMode::ALT_SCREEN | TermMode::ORIGIN;
 
         assert!(superset_mode.triggers_match(&subset_mode));
-        assert!(!subset_mode.triggers_match(&superset_mode));
+        assert!(subset_mode.triggers_match(&superset_mode));
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


baloo:
I can confirm it fixes #2641 for me

chrisduerr:
Since we've talked about this on IRC, I'd be interested what you think about this PR before merging it @nixpulvis. Let me know if you have any concerns.

nixpulvis:
@chrisduerr yep, this is on my radar. I'll make a point to test this weekend.

chrisduerr:
@nixpulvis No worries, didn't want to rush you into it. Just trying to get things into order so I don't lose track of things. :smile:

alacritty_terminal/src/input.rs
```diff
@@ -996,6 +986,72 @@ mod tests {
...
+    fn mods_binding_requires_strict_match() {
+        let mut superset_mods = MockBinding::default();
+        superset_mods.mods = ModifiersState { alt: true, logo: true, ctrl: true, shift: true };
+        let mut subset_mods = MockBinding::default();
+        subset_mods.mods = ModifiersState { alt: true, logo: false, ctrl: false, shift: false };
+
+        assert!(!superset_mods.triggers_match(&subset_mods));
+        assert!(!subset_mods.triggers_match(&superset_mods));
+    }
+
```
nixpulvis: Might be nice to add a test for `is_triggered_by` to cover the change in functionality.

chrisduerr:
@nixpulvis Regarding the opposite case, I've asked you about this on IRC iirc but I don't think you were around anymore. Do you think it would be better to override the default, even if it is a subset of the existing config option? I think we can go for that, I don't think there should be any issues with that.

kchibisov:
Just want to add my 2cents here, I think all implicit (default) binding with the same `key` + `mods` as a user provided ones must be overwritten. From a "user" point of view I found the following very confusing.

> This binding and the default will still be executed, resulting in a double paste when not in the alt buffer.

I'd like to add the following example to a @nixpulvis 's one:

"I want to write explicit binding for paste in `Alt` and in `~Alt` mode. To write everything correctly I should create the following bindings:

```
- { key: V,        mods: Control|Shift, action: None                       }
- { key: V,        mods: Control|Shift, action: Paste, mode: ~Alt          }
- { key: V,        mods: Control|Shift, action: Paste, mode: Alt           }
```

For me, the first binding is something strange. I understand it when I need to “unbind” something, but when I add similar bindings, it is intuitively not clear to me.

chrisduerr:
I don't think I'd go for unbinding all bindings. If there's a default binding for `Alt` and a user binds something just for `~Alt` mode, I would definitely not expect the other binding to disappear.

But we can do two-way intersection. So when any mode is present in both the default and the custom binding, the binding will be overwritten.

kchibisov:
> But we can do two-way intersection. So when any mode is present in both the default and the custom binding, the binding will be overwritten.

That's a good idea, it should also make unbinding process easier, since you can't simply unbind for specific mode atm. e.g I'm expecting the following to unbind `Paste` only in `Alt` mode.
```
- { key: V,        mods: Control|Shift, action: None, mode: Alt            }
```

chrisduerr:
@kchibisov That would not change. I don't think I want to support unbinding existing bindings in only one mode. A binding either exists or it doesn't, they don't mutate they only get filtered.

kchibisov:
>  I don't think I want to support unbinding existing bindings in only one mode.

@chrisduerr Ok, just to make it clear, you don't want to support "unbinding" and **changing**? So, you're expecting that every binding's mode should have the same `action` as a binding without any mods, am I understand you correctly here?

chrisduerr:
@kchibisov I'm just saying that all we should be doing is either disabling a default binding, or enabling or keeping it.

Your proposal would mean that we basically throw away the default binding, create a new one specific to the modes not overridden and add that again. I think that's too much complexity and we should just keep it simple.

I've updated the PR with what I think should be a good and straight-forward solution. Let me know what you think @kchibisov / @nixpulvis.

It is possible to specify more than one mode, which is something that needs to be considered. So there's lots of different combination of modes and inverted modes possible, but I think my current version should be a reasonable default. It will generally disable the default binding whenever the new binding's modes are either a superset or subset of all default modes.

If this doesn't provide a desirable outcome, I think the best alternative is to just always ignore modes when overriding. That is likely the simplest for the user. That way it would just always override the defaults regardless of mode and the user could adjust accordingly.

kchibisov:
@chrisduerr Maybe I didn't understand you correctly here, but you've just implemented the following.

> I don't think I'd go for unbinding all bindings. If there's a default binding for Alt and a user binds something just for ~Alt mode, I would definitely not expect the other binding to disappear.

So, this binding
```
- { key: V,        mods: Control|Shift, action: Paste, mode: Alt   }
```
unbinds the one for `~Alt`. That's what I was writing here:

>  I think all implicit (default) bindings with the same key + mods as a user provided ones must be overwritten

Is it miscommunication/understanding here or a bug in a current implementation?

chrisduerr:
Ah okay, I think it's a misunderstanding. Binding for `Alt` does not unbind the one for `~Alt`, there is none for `~Alt`, there's only one for all modes, which is unbound. So while in practice it unbinds the one for `~Alt`, in reality it just removes the binding completely, it doesn't modify the binding to remove it *just* for that one mode. So in `AppCursor` mode or something like that it's also removed now, not only for `~Alt`.

So we only make the choice if we want to completely remove a default binding or completely keep it. What I thought you meant was that we'd take a binding that applies to all modes and change it to a binding that only applies to what the new binding doesn't match. So the default binding doesn't stick around for everything that isn't `Alt`, it's just completely removed.

Does that clear things up?

kchibisov:
>  it's just completely removed.

@chrisduerr Yes, that clear things up, this is the thing I've initially suggested here, and huge thanks for detailed explanation, sorry for confusing you here.

> Your proposal would mean that we basically throw away the default binding, create a new one specific to the modes not overridden and add that again. I think that's too much complexity and we should just keep it simple.

My "proposal" was a production from the following

> So when any mode is present in both the default and the custom binding, the binding will be overwritten

I've initially thought of a default binding like a `key + mods + mode`, so forget about it ( one more misunderstanding).

nixpulvis:
This seems to be working well. Though I wonder if we should make this behavior more documented in the template config file.

https://github.com/jwilm/alacritty/blob/1c05b3bb0cfe29ddaa349906a7359f4b65e78f19/alacritty.yml#L526-L528

Maybe this could be changed to something along the lines of:

> Below is the list of default bindings for Alacritty. Manually configuring a binding that overlaps with any of these will remove the default. This is only true for default bindings, since it is possible to define two actions for the same binding that execute sequentially. If you'd simply like to remove a binding, set it's action to `None`.

A bit wordy, but something like that.

chrisduerr:
I honestly feel like that's just too verbose and confusing at that point. I feel like it might be smarter to just give a hint that bindings will get overwritten and let the user figure out exactly how it works.

In the past, the number of people reading this documentation has always been about zero, so I don't really feel like making it even more verbose will lead to much benefit. I'd rather keep it short, only to its important bits and simple to understand so that users are more inclined to actually give it a quick read.

nixpulvis:
@chrisduerr first off, I'll say I'm fine with whatever we decide here for this PR. The rest of this comment is a bike shed.

I do think that this speaks toward my idea for having configurations documented in the `config` module's rustdoc. It's a place to allow more verbose explanations, without cluttering the "example" config.

I'm actually quite frustrated with the `alacritty.yml` file ATM. It's a maintenance issue, since it requires updates to keep in sync with the `impl Default` and other config code changes. You've expressed liking this over other options on a few occasions, but I don't see how we can easily do but `impl Default` and a "default" (example) YAML config file.

chrisduerr:
> I do think that this speaks toward my idea for having configurations documented in the config module's rustdoc. It's a place to allow more verbose explanations, without cluttering the "example" config.

What would be the point of such a documentation? That seems to me like it would just be additional maintenance overhead, lead to diverging documentation (potentially having inaccurate and outdated info in one of them) and get completely ignored by everyone. Rustdoc is not an enduser documentation for Rust binaries.

I don't think that the `alacritty.yml` is much of a maintenance issue, I think something like rustdoc would be far, far worse. The default values can also never be incorrect, since they're checked by unit tests to be accurate. This provides great documentation and allows users to go for a full deep dive into all configuration options if they're interested, starting with something that already works great. This makes it easy to configure Alacritty without having to open external resources like manpages or websites.

nixpulvis:
I think given that this is a config file you're right about Rustdoc being inappropriate. I'm mostly trying (too hard?) to avoid needing to edit two places for every config change. I wonder if a `DefaultFromConfig` trait could be implemented at compile time. Or if it's worth it at this point.

Realizing there's a test for this will let me sleep well again however :D

chrisduerr:
I'd say it's very similar to traditional code documentation in this case. If the code changes, of course the documentation has to be adjusted too, that's just the nature of documentation. If the documentation just says exactly what has been defined in code, there's probably little reason to have the documentation since I can just look at the source at that point (except for styling maybe).

If a code example in rustdoc is outdated, it will need to be adjusted for code changes too. That's hardly a problem, that's just the nature of documentation really. However, it's almost possible to document code examples in languages other than Rust, simply because they aren't checked at all and it's so simple to get them out of sync. That's why I've written a test to check it and why it's so great that documentation examples are actually run, it makes sure that the interface between code and documentation is accurate and I think as long as that's the case, we should be safe.

I'm open to further discussion on this/go into a bit more detail, however I think that should be largely unrelated to this PR. There's probably better ways to word the documentation, however I think at this point it strikes a nice balance between simplicity and information, so I don't think this PR is the best place to do start large bikeshedding about it.

kchibisov:
I don't like idea the idea of using `rustdoc`, I'd rather move all config documentation to manpages ( `man 5 alacritty`/ `man alacritty.yml`) and leave `alacritty.yml` just with markup comments and default stuff.

I don't see any issue with "maintaining" `alacritty.yml`, but I see an issue with shipping documentation updates to end users (not to a new users). I rarely(never) update documentation in my `alacritty.yml` file and I feel like a lot of people also ignores doc updating and just consulting `alacritty.yml` file on a `github` or just raises an issue.

chrisduerr:
> I don't see any issue with "maintaining" alacritty.yml, but I see an issue with shipping documentation updates to end users (not to a new users). I rarely(never) update documentation in my alacritty.yml file and I feel like a lot of people also ignores doc updating and just consulting alacritty.yml file on a github or just raises an issue.

I feel like this is indeed one of the actual serious problems. Especially because this means that configs that are completely unchanged might often break.

This is the main reason why I've been advocating to not create the default configuration file anymore if it's not present. I think that would solve most of these issues with documentation/fields changing.

Then I'd go the route of i3 to have a config file in /etc/alacritty/alacritty.yml (or something like that) which is just our full config file that document everything. So a user can just adapt that whenever they're interested in configuring it.

An alternative would be to go more for a sshd_config like approach and have basically no information about the fields in the alacritty.yml and have all the fields documented in manpages.

kchibisov:
@chrisduerr 
@nixpulvis 
What do you think about creating survey about `alacritty.yml`? This way we can gather more opinions on how to ship docs/should we generate `alacritty.yml` or force users to grab one from `github.com`/`etc/alacritty/alacritty.yml` and so on.

chrisduerr:
I wouldn't be against it, just to get a few more opinions, however I don't think it'll provide any significant insight. If it's about making the actual decision, I think asking on github is a bit pointless, since likely a big chunk of users wouldn't be represented. Having only the powerusers respond that are actively tracking github would likely just result in a one-sided result that would probably favor minimal defaults with maximum user configuration.

nixpulvis:
@kchibisov & @chrisduerr first off, thank you for basically taking my frustration and finding the right solution.

> Then I'd go the route of i3 to have a config file in /etc/alacritty/alacritty.yml (or something like that) which is just our full config file that document everything. So a user can just adapt that whenever they're interested in configuring it.

I think this is a great idea.

> I'd rather move all config documentation to manpages ( man 5 alacritty/ man alacritty.yml) 

If we can keep a manpage updated easily, I'd be very happy to have the config included.

kchibisov:
>  If it's about making the actual decision. I think asking on github is a bit pointless, since likely a big chunk of users wouldn't be represented.  Having only the powerusers respond that are actively tracking github would likely just result in a one-sided result that would probably favor minimal defaults with maximum user configuration.

@chrisduerr Hm, thats a good point, I'm not comfortable on creating surveys outside of "github" and I  don't think that I'm the one who should do this (I mean, creating surveys outside of `alacritty`s `github` bug tracker, which cannot be easily moderated by **maintainers**). I think that it's been suggested mostly everything that we can do with our documentation. However, it's a little bit chaotic, so creating `RFC` to structure everything and put in "one place" would be much better idea than survey. I'd like to write this one and I'll try to post it in the next few days.

P.s. Thanks @nixpulvis for starting this discussion.

chrisduerr:
Would be totally open to an RFC on what people (especially the macOS/Windows) crowd would think about not spawning the default config by default (and other alternatives).

PR #2644, opened 2019-07-13
Query DirectWrite for the rendering mode to use

TheDecryptor:
As an alternative to #2598, query DirectWrite for the font rendering mode to use instead of hard-coding one.

With the settings on my system this doesn't make much of a difference visually, but asking for the OS to decide the mode seems better simply because there could be different system configurations that result in different modes, or it could even be varied on a per font basis by Microsoft.

### DIFF
--- a/font/src/directwrite/mod.rs
+++ b/font/src/directwrite/mod.rs
@@ -143,11 +143,17 @@ impl crate::Rasterize for DirectWriteRasterizer {
             bidiLevel: 0,
         };
 
+        let rendering_mode = font.get_recommended_rendering_mode_default_params(
+            glyph.size.as_f32_pts(),
+            self.device_pixel_ratio * (96.0 / 72.0),
+            dwrote::DWRITE_MEASURING_MODE_NATURAL
+        );
+
         let glyph_analysis = GlyphRunAnalysis::create(
             &glyph_run,
             self.device_pixel_ratio * (96.0 / 72.0),
             None,
-            dwrote::DWRITE_RENDERING_MODE_NATURAL,
+            rendering_mode,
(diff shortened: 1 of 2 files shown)


font/src/directwrite/mod.rs
```diff
@@ -143,11 +143,17 @@ impl crate::Rasterize for DirectWriteRasterizer {
             bidiLevel: 0,
         };
 
+        let rendering_mode = font.get_recommended_rendering_mode_default_params(
+            glyph.size.as_f32_pts(),
+            self.device_pixel_ratio * (96.0 / 72.0),
+            dwrote::DWRITE_MEASURING_MODE_NATURAL
```
chrisduerr: Just out of curiosity, do you know anything about the measuring mode by chance? Is this maybe something we can also ask Windows about rather than setting it ourselves?

TheDecryptor: The measuring mode is left entirely up to the application, there's no real default because it's down to how you're trying to use DirectWrite.

It's basically a switch between how DirectWrite (Natural) and GDI works, the intent was for applications replacing their existing text layout to be able to keep their existing UI metrics (via GDI compatible layout), while also allowing resolution independent layout for things like eBook/PDF readers, browser engines, etc.

But that should only have an effect if you're doing "flowing" text layout, Alacritty is rendering single glyph buffers to a fixed grid which should completely negate any differences.

TheDecryptor:
Ok, added a line to the change-log, sound alright?

font/src/directwrite/mod.rs
chrisduerr: Thanks a lot for the detail, good to know!

PR #2648, opened 2019-07-15
URL parser: add angle brackets as delimiters

### LINKED ISSUE #2629: Links surrounded by angle brackets aren't correctly parsed
I've noticed that Alacritty does not correctly recognize URLs surrounded by angle brackets. With "angle brackets" here I specifically mean these two Unicode codepoints:

- [U+27E8](https://www.fileformat.info/info/unicode/char/27e8/index.htm) (`⟨`)
- [U+27E9](https://www.fileformat.info/info/unicode/char/27e9/index.htm) (`⟩`)

These angle brackets are rendered in manual pages when the `.UR`+`.UE` combination of groff macros is used ([see here](http://man7.org/linux/man-pages/man7/groff_man.7.html)). Since this is the recommended way to format URLs and e-mails in man pages, it is a quite common scenario (see for example `man 1 proxy`, `man 1 groff`, `man 1 script`, `man 1 renice`).

In particular, Alacritty identifies the closing angle bracket as part of the URL, while it is not, so in something like this:

    ⟨https://example.com/x/y⟩

the identified URL is:

    https://example.com/x/y⟩

while it should be:

    https://example.com/x/y

mebeim:
After having #2629 sitting there for a while I decided to have a go at it. The change is pretty straightforward. There isn't much to say in addition to what I already explained there. Let me know if this is ok. 

Fixes #2629.

### DIFF
--- a/alacritty_terminal/src/url.rs
+++ b/alacritty_terminal/src/url.rs
@@ -93,7 +93,7 @@ impl UrlParser {
                 ')' if open_parens_count > 0 => open_parens_count -= 1,
                 '[' => open_bracks_count += 1,
                 ']' if open_bracks_count > 0 => open_bracks_count -= 1,
-                ')' | ']' => {
+                ')' | ']' | '⟩' => {
                     self.state.truncate(i);
                     break;
                 },
@@ -246,9 +246,11 @@ mod tests {
         url_test("https://example.org)", "https://example.org");
         url_test("https://example.org(", "https://example.org");
         url_test("(https://one.org/)(https://two.org/)", "https://one.org/");
+        url_test("⟨https://example.org/test⟩", "https://example.org/test");
 
         url_test("https://[2001:db8:a0b:12f0::1]:80", "https://[2001:db8:a0b:12f0::1]:80");
         url_test("([(https://example.org/test(ing))])", "https://example.org/test(ing)");
+        url_test("⟨([(https://example.org/test(ing))])⟩", "https://example.org/test(ing)");
         url_test("https://example.org/]()", "https://example.org/");
         url_test("[https://example.org]", "https://example.org");
 


alacritty_terminal/src/url.rs
```diff
@@ -93,7 +93,9 @@ impl UrlParser {
                 ')' if open_parens_count > 0 => open_parens_count -= 1,
                 '[' => open_bracks_count += 1,
                 ']' if open_bracks_count > 0 => open_bracks_count -= 1,
-                ')' | ']' => {
+                '⟨' => open_bracks_count += 1,
+                '⟩' if open_bracks_count > 0 => open_bracks_count -= 1,
```
nixpulvis: These angle brackets should not be treated as the same as `[` in terms of nesting level.

nixpulvis:
I found [this SO](https://stackoverflow.com/questions/13535172/list-of-all-unicodes-open-close-brackets) post about a list of all Unicode matching characters. Would this be something we want to expand support to more of?

What is the "correct" logic here?

chrisduerr:
> I found this SO post about a list of all Unicode matching characters. Would this be something we want to expand support to more of?

I don't think we want to include anything that isn't absolutely necessary. Every single one of these pairs not only increases the complexity of the code, but also significantly impacts the runtime performance of URL checking.

> What is the "correct" logic here?

The 'correct' way according to standard would be to include all brackets regardless of matching state, since they're all valid URL characters. However this is simple not reasonable in a terminal context, so we need to break that and go with what makes sense rather than going for what's 'correct'.

Generally I intentionally haven't implemented this yet because I'm currently in the process of rewriting the way URLs are detected in Alacritty, making the changes in this code victim to refactoring in the future, so this code would just get removed (hopefully very soon). The plan is to replace the URL searching and bracket logic with [rfind_url](https://crates.io/crates/rfind_url). This is significantly faster than what we have right now and does not require a starting position.

I'd be okay with solving the remaining issues in this PR and getting it merged before my refactoring, however I have planned to get this done within the next two weeks, so I'm not sure how much sense it makes to put any effort into this. However the library I've written currently also doesn't check for this bracket character, though it would be significantly easier to add it there (a quick look at the code should show how simple it is).

nixpulvis:
@chrisduerr cool. Sounds like we do want to support this bracket character, and we can discuss other matching characters as they come up on a case by case basis.

chrisduerr:
Yeah, I'd just wait until someone complains, the same way it happened in this case. If the reasoning is good enough (manpages seem pretty important for terminals), then we can go for it. I think that's the best choice to keep things simple.

alacritty_terminal/src/url.rs
mebeim: Got it.

chrisduerr:
Unfortunately, this PR has been sent at a bit of an unfortunate time while I was working on https://github.com/jwilm/alacritty/pull/2681 myself. This completely reworks the URL searching mechanism and in its latest edition supports angle brackets too, however this does mean that there would be little benefit to merge this PR first (besides some remaining issues with it).

I appreciate the contribution and want to apologize that things didn't quite work out this time, hopefully that is not too discouraging. Thanks for bringing awareness to the problem by creating an issue and looking for a solution, even if the code itself wasn't able to make it into Alacritty.

mebeim:
@chrisduerr no worries. I'm using my modified version. I opened this PR just to "fix" this issue for everyone, didn't know you were already working on a major update to the URL parsing. That's definitely a better solution. Keep up the good work, I love this terminal emulator!

PR #2661, opened 2019-07-20
Always show Text mouse cursor when using Shift to select a block of text

### LINKED ISSUE #2550: Cursor in apps with mouse mode doesn't set to `MouseCursor::Text` on shift press.
Linux X11
Alacritty version: `alacritty 0.3.3-rc2 (58e2365)`

In apps like `htop` (with mouse mode) when pressing `Shift` the cursor is left arrow (default one on my system) instead of text. So, text selection is a bit hard. Cursor should probably be `Default` without `Shift` and `Text` with `Shift`.

jjsullivan5196:
Resolves #2550

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - URL parser incorrectly handling Markdown URLs and angled brackets
 - Intermediate bytes of CSI sequences not checked
 - Wayland clipboard integration
+- Use text mouse cursor when mouse mode is temporarily disabled with shift
 
 ## 0.3.3
 
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -52,11 +52,7 @@ fn parse_rgb_color(color: &[u8]) -> Option<Rgb> {
         Some((255 * value / max) as u8)
     };
 
-    Some(Rgb {
-        r: scale(colors[0])?,
-        g: scale(colors[1])?,
-        b: scale(colors[2])?,
-    })
+    Some(Rgb { r: scale(colors[0])?, g: scale(colors[1])?, b: scale(colors[2])? })
 }
 
 // Parse colors in `#r(rrr)g(ggg)b(bbb)` format
@@ -1449,8 +1445,8 @@ pub mod C1 {
 #[cfg(test)]
 mod tests {
     use super::{
-        parse_number, xparse_color, Attr, CharsetIndex, Color, Handler, Processor,
-        StandardCharset, TermInfo,
+        parse_number, xparse_color, Attr, CharsetIndex, Color, Handler, Processor, StandardCharset,
+        TermInfo,
     };
     use crate::index::{Column, Line};
     use crate::term::color::Rgb;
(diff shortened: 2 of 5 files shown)


jjsullivan5196:
Yeah, it would change the cursor, but the URL clicks still work. I'll just put the two together and prefer the Hand cursor for URLs.

jjsullivan5196:
I changed the behavior in `mouse_moved` so that if any URL is highlighted, the `Hand` cursor is shown instead of `Text`.

I did notice a possible regression from release v0.3.3 (on master 76dd84b) while testing this. The cursor appears to flip between `Hand` and the default cursor for the context when moving the mouse across an URL, in comparison to release v0.3.3 where the cursor is always `Hand` when hovering across the URL. 

This behavior is still present in this patch, but I don't believe it contributes to the problem. I'll see if I can find the commit that caused it and file a bug report.

kchibisov:
@jjsullivan5196 I guess this was introduced by this one 84aca672964e29b5b4503b7da7bc34fc395f08ab

jjsullivan5196:
@kchibisov just checked the parent of that commit and it has the same behavior as release, 84aca67 seems to have broken URL highlights outright and the next commit (a99547c) shows the behavior I mentioned.

kchibisov:
@jjsullivan5196 a99547c fixed a different bug introduced by 84aca67 :- ).

> Would you mind opening an issue? I can open in case you don't want, just let me know.

?

jjsullivan5196:
Sorry @kchibisov was in the middle of something, I can create the issue :+1:

jjsullivan5196:
Put in some notes about what was going on and those commits. Gonna be away for a bit but if noone gets to this I can see if I can add a fix.

jjsullivan5196:
Took at look at #2666 and it fixes the cursor shape flicker, I'd recommend pulling that first and then I can merge that into my pull. Everything should be working correctly by then.

kchibisov:
@jjsullivan5196 #2666 was merge and we've refactored a lot of related code to your PR. Implementing the feature you're working on here should be much easier now.

jjsullivan5196:
@kchibisov Thanks! I'll merge and rework this a bit

jjsullivan5196:
So I basically took my original commit and re-wrote it in the context of the `MousePosition` enum, so holding `Shift` will only force the `Text` cursor when the cursor is in the main part of the screen and not intersecting URLs/other special areas.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Config option `window.gtk_theme_variant` to set GTK theme variant
 - Completions for `--class` and `-t` (short title)
 - Change the mouse cursor when hovering over the message bar and its close button
+- Text mouse cursor is shown when shift-selecting text
```
chrisduerr: This comment is a bit imprecise. It should state that the text mouse cursor is shown when holding shift while in mouse mode.

jjsullivan5196:
Don't have time to work on this right now, anyone's welcome to take the issue

chrisduerr:
No worries, I'm sorry this took longer than expected. I think it should be fairly straight-forward from here though. I'll take a look as soon as some time frees up.

PR #2664, opened 2019-07-20
Ignore unsupported CSI sequences

### LINKED ISSUE #2171: Interpretation of "private" CSI diverges from other terminals.
Please don't threat CSI sequences with "private use" (<, = , >, ?) chars the same as sequences without with the same final character.

These are code space extensions and often have quite different functions. Interpreting them the same as the sequence without will create glitches and confuse applications. The same applies to the intermediates (` !\"#$%&'()*+,-./`).

Properly ignoring unknown sequences is essential for the ongoing terminal ecosystem evolution as new sequences are allocated.

The current code in ansi.rs: `csi_dispatch` seems to get both these sequence parts in `intermediates` but mostly ignores them, except for select cases.

The following is based on information of the sequences interpreted by xterm, vte/gnome-terminal, konsole and rxvt-unicode. Also see my ongoing [documentation project](https://terminalguide.netlify.com/seq/) for further details.

I think for the currently implemented sequences the only ones that should trigger with an non empty `intermediates` are:
* `J` and `K` for `intermediates` exactly `?`. Strictly this is for protected cells, but it's commonly implemented on terminals that don't actually implement protected cells (if protected cells are not supported it behaves like the non `?` version)
* And of course `h` and `l` for `intermediates` exactly `?`. Similar to what alacritty currently does.
* possibly `n` for `intermediates` exactly `?`. The expected behavior here is different from the not `?` versions and alacritty should either ignore them or properly implement this different behavior. alacritty currently only supports 5 and 6 so a proper implementation can just add a additional `?` in the output for the dec private variants.
* `q` should set cursor style only when `intermediates` is exactly ` ` (space). Without it's a different sequence understood by xterm (and DEC compatible terminals; DECLL).

Currently i think the following claimed sequences are misinterpreted:
```
ESC [ Ⓝ ␣ @  	Scroll Left
ESC [ Ⓝ ␣ A  	Scroll Right
ESC [ > Ⓝ T  	Unset Title Mode
ESC [ > Ⓝ c  	Secondary Device Attributes
ESC [ = Ⓝ c  	Tertiary Device Attributes 
ESC [ Ⓝ ␣ d  	Unset Tab Stop
ESC [ > Ⓝ ; Ⓝ m  	xterm keyboard mode configuration
ESC [ ? Ⓝ n  	private Device Status Report (different expected response)
ESC [ Ⓝ q  	Load LEDs (DECLL)
ESC [ Ⓝ * q  	DEC secure reset
ESC [ Ⓝ " q  	Select Character Protection Attribute (DECSCA)
ESC [ ? Ⓝ s  	Save Mode
ESC [ Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ $ r  	Change attributes in rectangular area 
ESC [ Ⓝ ␣ u  	Set Margin Bell Volume
```
But in the future there are a lot more that have different meanings with non empty `intermediates`
```
ESC [ ? W  	Set Tab At Every 8 Columns
ESC [ Ⓝ $ p  	Request Mode
ESC [ Ⓝ + p     DEC secure reset
ESC [ ? Ⓝ $ p  	Request Private Mode
ESC [ > Ⓝ p  	Change Mouse Pointer Auto Hide
ESC [ ! p  	Soft Reset
ESC [ ? Ⓝ r  	Restore Mode
ESC [ ? Ⓝ t  	Toggle Mode
ESC [ > Ⓝ t  	Set Title Mode
ESC [ Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ $ t  	Reverse Attributes in Rectangular Area
ESC [ Ⓝ ␣ t  	Set Warning Bell Volume
ESC [ Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ $ v  	Copy rectangular area
ESC [ Ⓝ $ w  	Request Presentation State Report
ESC [ 1 $ w  	Request Cursor Information Report
ESC [ 2 $ w  	Request Tab Stop Report
ESC [ Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ ' w  	DEC Locator
ESC [ Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ $ x  	Fill Rectangular Area
ESC [ Ⓝ * x  	Select Attribute Change Extent
ESC [ Ⓝ # y  	Setup checksum parameters
ESC [ Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ * y  	Request Checksum of Rectangular Area
ESC [ Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ $ z  	Erase Rectangular Area
ESC [ Ⓝ ; Ⓝ ' z  	DEC Locator
ESC [ Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ $ {  	Selective erase rectangular area
ESC [ Ⓝ # {  	SGR Push
ESC [ Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ ' {  	DEC Locator
ESC [ Ⓝ $ |  	DECSCPP
ESC [ Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ # |  	Report SGR Rectangle
ESC [ Ⓝ * |  	DECSNLS
ESC [ Ⓝ ' |  	DEC Locator
ESC [ # }  	SGR Pop State
ESC [ Ⓝ ' }  	Insert Column
ESC [ Ⓝ ' ~  	Delete Column
```

akinomyoga:
I came from https://github.com/akinomyoga/ble.sh/issues/28, https://github.com/jwilm/alacritty/issues/2652, https://github.com/jwilm/alacritty/issues/2171 .

For details of the problem, please refer to these issues. If it does not support some control functions, it should ignore them rather than triggering random control functions which have similar coded representations. The problem is not restricted in private sequences, but even some ANSI sequences (such as SL, SR, PPA, etc.) triggers wrong control functions in the current Alacritty where parameter/extra-intermediate bytes of CSI sequences are not checked. In this patch, the sequences with extra bytes are handled separately from the sequences without extra bytes.

Note: the support for DA2 (Secondary Device Attribute) is not included in this patch.

Fixes #2171.


akinomyoga:
Travis CI produces the following error.

```
error: the function has a cognitive complexity of 32
    --> alacritty_terminal/src/ansi.rs:893:5
     |
893  | /     fn csi_dispatch(&mut self, args: &[i64], intermediates: &[u8], _ignore: bool, action: char) {
894  | |         let handler = &mut self.handler;
895  | |         let writer = &mut self.writer;
896  | |
...    |
1165 | |         }
1166 | |     }
     | |_____^
     |
```

Honestly I'm new to Rust, so I don't know what is the cleanest way (with minimal changes) to reduce the "cognitive complexity". Also I'm not sure what defines the "cognitive complexity". For example, does creating local functions increase or decrease the cognitive complexity? I'm sorry but let me make several experiments here...

kchibisov:
>  I'm sorry but let me make several experiments here...

@akinomyoga It's fine, however you'll load ci this way, which is quite slow. You can install `clippy` on your local machine with `rustup component add clippy` and run it with `cargo clippy`, so it'll be faster to make your experiments (just an advice).

> For example, does creating local functions increase or decrease the cognitive complexity?

It should decrease

P.s. You should add entry to `Fixed` section in `CHANGELOG.md`. (one more advice before @chrisduerr  's review)

akinomyoga:
@kchibisov Thank you very much for your useful advice for `clippy`! It helped me a lot. I managed to suppress the cognitive complexity, but I'm not sure whether I'm doing things in the right way. Anyway I would like to wait for review.

akinomyoga:
Thank you very much for your comments and advices!

First of all, as I have already mentioned above, I'm new to Rust, so I have to admit that I need some help. Initially when I opened the PR, I didn't wanted to change existing styles/approaches with little knowledge on Rust, so I tried to minimize the changes. After I realized that `clippy` produces errors, I had to update codes largely with the little knowledge, so I'm actually not satisfied with the current PR quality.

> First of all, the extraction of `process_sgr` is fine, however there's absolutely no reason to nest it inside the csi_dispatch function, that just makes things difficult to und\
erstand.

Thank you for your comment. Initially I wanted place the function in `impl` as something like private members in other languages but realized that's not allowed. What is the common strategy for such a case? How about putting the function at the top level? Or could I add another function to `trait Handler`?

> However I'm not sure if there's any immediate incentive to move this, was this done because of a clippy warning?

Yes, initially I didn't wanted change largely the existing codes, so without the warning I wouldn't have moved the codes. Anyway, after applying your advices, `clippy` still produces the same error on cognitive complexity if I moved back the codes into `csi_dispatch`. So currently I still have a separate function `process_sgr`. Are there other ways to suppress the cognitive complexity?

(If this was my project and could freely change the existing codes, I would prepare a function for each ANSI/private control function, and make `csi_dispatch` really dispatch procedures to such functions. But that's just my taste with little knowledge on Rust...)

> but since we're only going to use it twice (see following paragraph), I think this is best inlined. We can then just set the private/nonprivate based on the intermediates.

Thank you, I have inlined this function along with the changes mentioned in the following paragraph.

> However I think there's a good solution for it. Instead of matching against action twice, we can just match against action and intermediates.

Thank you very much for the nice suggestion. I have updated the codes.

> Lastly, I think the early return should be moved to the top of the function

I have applied the change.

akinomyoga:
> I think the main thing is the J and K escape, though checking all of them probably makes sense. I'll give this a more detailed test once the style issues are resolved.

I hesitate to support DECSED (<kbd>CSI ? J</kbd>) and DECSEL (<kbd>CSI ? K</kbd>) as ED (<kbd>CSI J</kbd>) and EL (<kbd>CSI K</kbd>). DECSED/DECSEL is aimed to be used in combination with DECSCA (<kbd>CSI " q</kbd>). So it appears to me that it is natural that DECSED/DECSEL/DECSCA (and maybe also DECSERA) are supported at the same time. It is true that DECSED/DECSEL behaves equivalently to ED/ED if DECSCA isn't used at all as described in #2171. But if some terminal application doesn't use DECSCA, there is no reason to use DECSED/DECSEL instead of standardized ED/EL. If some terminal application uses DECSED/DECSEL, it probably uses DECSCA before that. In such a case I'm not sure which is better to erase everything including protected areas or erase nothing. Other related information:

- A terminal can include parameter value 6 in DA1 response (as second or later parameters) to indicate that it supports DECSED/DECSEL/DECSCA to terminal applications. So terminal application can switch its behavior using DA1.
- Not so many terminals support DECSED/DECSEL/DECSCA, and as far as I know there is no other terminal that supports DECSED/DECSEL without support of DECSCA.

chrisduerr:
> First of all, as I have already mentioned above, I'm new to Rust, so I have to admit that I need some help.

No worries about that, always happy to help out where it's needed.

> Thank you for your comment. Initially I wanted place the function in impl as something like private members in other languages but realized that's not allowed. What is the common strategy for such a case? How about putting the function at the top level? Or could I add another function to trait Handler?

I was kinda expecting this issue to pop up since it might be a bit confusing for new people, but I forgot about it in the original review. I think a little example should explain how this works in Rust:

```rust
struct MyStruct {
    field: String,
}

// This is the impl for the struct, we define methods on the struct here
impl MyStruct {
    fn output(&self) { MyStruct::foobar(self.field); }

    // We can also define methods on the struct which don't take 'self', commonly 'new' for example
    fn foobar(input: String) { print(input); }
}

// This is a trait implementation, all methods on the trait need to be implemented here and only methods of the trait are allowed in here
impl Handler for MyStruct {
    // we can reference methods directly on the struct here!
    osc_dispatch(&self) { self.output(); }
}

// This is a 'static' method, no `self`, but we can call it from everywhere
fn print(test: String) {
    println!("{}", test);
}
```

I think that should compile, though I haven't tried. But it's hopefully sufficient for getting across the options you have. So in your case you need to move the implementation from the trait impl to the struct impl.

> (If this was my project and could freely change the existing codes, I would prepare a function for each ANSI/private control function, and make csi_dispatch really dispatch procedures to such functions. But that's just my taste with little knowledge on Rust...)

If you look a bit deeper at the code, this is actually what we do. With each of these different escapes, we just call `handler.<escape_handler>`. The SGR codes are just transformed to their values in the `ansi.rs` file, since it has the control over colors and attribute codes, so moving that to the terminal wouldn't be much cleaner.

However, I do think we can clean this up a lot, in a way which reduces the cognitive complexity **and** which is more idiomatic Rust code.

Basically, what we're doing and what most of the code is about is converting a number to an `Attr`. So instead of doing that where we handle the SGR codes, we can make it so `Attr` itself takes an integer as input and returns `Self`.

That would look something like this:

```rust
impl Attr {
    fn from_i64(from: i64) -> Option<Attr> {
        match i64 {
        // ...
        }
    }
}
```

I think that would be the best solution for reducing the amount of code in the sgr parser. That way, all we have to do would be to just call `Attr::from_i64(args[i])` and we'd be good.

> I hesitate to support DECSED (CSI ? J) and DECSEL (CSI ? K) as ED (CSI J) and EL (CSI K). DECSED/DECSEL is aimed to be used in combination with DECSCA (CSI " q). So it appears to me that it is natural that DECSED/DECSEL/DECSCA (and maybe also DECSERA) are supported at the same time. It is true that DECSED/DECSEL behaves equivalently to ED/ED if DECSCA isn't used at all as described in #2171. But if some terminal application doesn't use DECSCA, there is no reason to use DECSED/DECSEL instead of standardized ED/EL. If some terminal application uses DECSED/DECSEL, it probably uses DECSCA before that. In such a case I'm not sure which is better to erase everything including protected areas or erase nothing. Other related information:

You make some good points here, however in this case I'd say usually following the example of other terminals is the best choice. I haven't tested this myself, but based on the comment in #2171 it sounds like most terminals which have similar support as Alacritty just clear the screen, so I feel like we probably should follow that example. I haven't actually tested that myself though, so I can't say for sure.

akinomyoga:
### SGR

> I think a little example should explain how this works in Rust:

Thank you very much for the illustration! I can learn many things from the example.

> Basically, what we're doing and what most of the code is about is converting a number to an Attr. So instead of doing that where we handle the SGR codes, we can make it so Attr itself takes an integer as input and returns Self.
>
> That would look something like this:

Thank you. I had to slightly change the function parameters as follows because the SGR parameter 38 and 48 takes additional parameters. If you have alternative ideas, could you please let me know?

```rust
impl Attr {
    fn from_sgr_parameters(parameters: &[i64], i: &mut usize) -> Option<Attr> {
        ...
    }
}
```

By the way, maybe an existing function `parse_color` is also better to be moved to `impl Attr`?

### DECSED/DECSEL

> I haven't tested this myself, but based on the comment in #2171 it sounds like most terminals which have similar support as Alacritty just clear the screen, so I feel like we probably should follow that example. I haven't actually tested that myself though, so I can't say for sure.

Actually after I read #2171, I tried several terminals, but I have not yet found terminals which supports DECSED/DECSEL without support of DECSCA (except for Alacritty). You can test with the following command:

```bash
$ printf 'A\e[1"qB\e[2"qC\e[?2K\n'
```

When both DECSEL and DECSCA are supported, it will print "<code>&nbsp;B&nbsp;</code>". When only DECSEL is supported, it will print "<code>&nbsp;&nbsp;&nbsp;</code>". When none of DECSEL and DECSCA are supported, it will print "`ABC`". The following is the list of terminals I tested with:

```
Alacritty (0.3.3)  "   " (the current release version)
GNU screen (4.6.2) "ABC"
tmux (2.9a)        "ABC"
urxvt256c (9.22)   "ABC"
lxterminal (0.3.2) "ABC"
konsole (19.4.2)   "ABC"
Poderosa (4.4.3)   "ABC"
xterm (#334)       " B "
mintty (3.0.2)     " B "
RLogin (2.24.4)    " B "
mlterm (3.8.8)     " B "
Tera Term (4.103)  " B "
```

Maybe we can ask @textshell (the author of #2171) for help. According to @textshell's site, `vte` and `konsole` support DECSED/DECSEL without support of DECSCA [1-3], but as far as I tested, `lxterminal` (a libvte-based terminal) and `konsole` don't support DECSED/DECSEL. Maybe I miss something.

- [1] [Selective Erase Display [DISPATCH] (DECSED) — Terminal Guide](https://terminalguide.netlify.com/seq/csi_cj__p/)
- [2] [Selective Erase Line [DISPATCH] (DECSEL) — Terminal Guide](https://terminalguide.netlify.com/seq/csi_ck__p/)
- [3] [Select Character Protection Attribute (DECSCA) — Terminal Guide](https://terminalguide.netlify.com/seq/csi_sq_t_quote/)

chrisduerr:
> Thank you. I had to slightly change the function parameters as follows because the SGR parameter 38 and 48 takes additional parameters. If you have alternative ideas, could you please let me know?

I didn't see that `i` is incremented in the match like that. It might be a better solution then to move the loop into `Attr` and return a `Vec<Attr>`. Then we can just do the following: 

```rust
for attr in Attr::from_sgr_parameters(parameters) {
    handler.terminal_attribute(attr);
}
```

I think that should work a bit better. However there's the slight problem that this doesn't just create a single `Attr` anymore, making the code a bit unidiomatic (a `from_` method usually should return `Self`). This is why I'd recommend moving it just above the `parse_color` function and maybe renaming it to `attrs_from_sgr_parameters()` or something like that.

> By the way, maybe an existing function parse_color is also better to be moved to impl Attr?

I think `parse_color` is good where it's at right now.

> Maybe we can ask @textshell (the author of #2171) for help. According to @textshell's site, vte and konsole support DECSED/DECSEL without support of DECSCA [1-3], but as far as I tested, lxterminal (a libvte-based terminal) and konsole don't support DECSED/DECSEL. Maybe I miss something.

The only VTE terminal that works properly on my system (Termite) behaves the same way as Alacritty, so I'm a bit surprised lxterminal would behave differently here. But if urxvt doesn't do anything and neither does Konsole, that's sufficient precedent for me to do nothing (which I think makes slightly more sense). So we should be good in that regard.

chrisduerr:
Also it looks like your latest change has broken Alacritty, it's probably good to check that the tests match up to make sure there's no regression.

Additionally, it seems like the perfect opportunity to introduce a test for the DECSED/DECSEL behavior if possible.

akinomyoga:
### SGR

> Then we can just do the following:
>
> ```rust
> for attr in Attr::from_sgr_parameters(parameters) {
>     handler.terminal_attribute(attr);
> }
> ```


When there is unknown SGR parameters, I wanted to output error messages as done in the original codes, so I implemented a modified version:

```rust
for attr in attrs_from_sgr_parameters(args) {
    match attr {
        Some(attr) => handler.terminal_attribute(attr),
        None => unhandled!(),
    }
}

(snip)

fn attrs_from_sgr_parameters(parameters: &[i64]) -> Vec<Option<Attr>> {
```

[ -- Note: I initially tried the following approach, but it broke a ref test `fish_cc`. The reason is because, in the original code, known parameters before the unknown parameter are processed, but in the following approach, the entire parameters wouldn't be processed when there are some unknown parameters.

```rust
match attrs_from_sgr_parameters(args) {
    Some(attrs) => {
        for attr in attrs {
            handler.terminal_attribute(attr);
        }
    },
    None => unhandled!(),
}

(snip)

fn attrs_from_sgr_parameters(parameters: &[i64]) -> Option<Vec<Attr>> {
```

-- End of Note ]

> I think parse_color is good where it's at right now.

OK

### DECSED/DECSEL

> so I'm a bit surprised lxterminal would behave differently here.

It turned out that LXTerminal uses an old version of VTE which is compatible with GTK+ 2.0 [1]. In my system, LXTerminal uses VTE v0.28.2 which is released on 2011-08-29 [2]. While, DECSED/DECSEL seems be introduced in VTE on 2018-03-28 and available from v0.53.0 [3].

- [1] [#790184 - lxterminal: depends on vte which is deprecated - Debian Bug report logs](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=790184)
- [2] [Release 0.28.2 · GNOME/vte](https://github.com/GNOME/vte/releases/tag/0.28.2)
- [3] [parser: Import new parser · GNOME/vte@78b37c0](https://github.com/GNOME/vte/commit/78b37c0ef)

> Also it looks like your latest change has broken Alacritty, it's probably good to check that the tests match up to make sure there's no regression.

I'm sorry. I somehow thought `cargo clippy` includes tests.

> Additionally, it seems like the perfect opportunity to introduce a test for the DECSED/DECSEL behavior if possible.

Can I think we add a tentative test for unsupported DECSED/DECSEL/DECSCL which will be removed (or updated) when DECSED/DECSEL/DECSCL are supported if any? I tried to create a ref test using the following command, but the resulting `grid.json` becomes 161 MiB...

```bash
$ cat reftest_vttest_erase.sh
#!/bin/sh

printf '\x1b[H'
printf 'A\x1b[1"qB\x1b[2"qC\x1b[?2J\n'
printf 'A\x1b[1"qB\x1b[2"qC\x1b[?2K\n'
$ target/release/alacritty --ref-test -e ./reftest_vttest_erase.sh
```

alacritty_terminal/tests/ref.rs
```diff
@@ -48,6 +48,7 @@ ref_tests! {
     vttest_origin_mode_2
     vttest_scroll
     vttest_tab_clear_set
+    vttest_selective_erasure
```
chrisduerr: The `vttest` are named this way because they were created with the `vttest` application. Looking at your test it doesn't look like that's the case. So you should probably rename it to prevent confusion, simply dropping the `vttest_` should do the job.

alacritty_terminal/src/ansi.rs
```diff
@@ -905,6 +904,10 @@ where
             }};
         }
 
+        if has_ignored_intermediates || intermediates.len() > 1 {
+            unhandled!();
+        }
```
chrisduerr: This should either be above all in-function macros or before the macros, there's little reason to have it in between.

I'd propose taking the `let handler`, `let writer` and `if has_ignored_intermediates` and moving it after the `arg_or_default` macro.

alacritty_terminal/src/ansi.rs
```diff
@@ -1174,6 +1105,97 @@ where
     }
 }
 
+fn attrs_from_sgr_parameters(parameters: &[i64]) -> Vec<Option<Attr>> {
```
chrisduerr: I think this looks good now, thanks!

akinomyoga:
> The vttest are named this way because they were created with the vttest application. 

Thank you for the information! I renamed it to `selective_erasure`.

I noticed that there is a test for DECSCA/DECSED/DECSEL in `vttest` which is found in the following place:

- `11. Test non-VT100 (e.g., VT220, XTERM) terminals`
  - `1. Test of VT220 features`
    - `2. Test screen-display functions`
      - `4. Test Protected-Areas (DECSCA)`

But Alacritty fails to reproduce the expected result as it doesn't support selective erasure. Maybe a ref test for this failure (i.e., how it fails) could also be added this time. What do you think of this?

> I'd propose taking the let handler, let writer and if has_ignored_intermediates and moving it after the arg_or_default macro.

I moved them.

chrisduerr:
> But Alacritty fails to reproduce the expected result as it doesn't support selective erasure. Maybe a ref test for this failure (i.e., how it fails) could also be added this time. What do you think of this?

I don't think it's a good idea to use a reftest of vttest to make sure something fails. We don't want to force the test case to fail, we only don't implement it because there seems to be little need for the escape. If it would just accidentally work, that would be fine by me.

So I'd rather stay away from vttest for this. Instead just having a separate little test case should be perfectly fine.

alacritty_terminal/src/ansi.rs
```diff
@@ -970,129 +973,68 @@ where
...
+            ('L', None) => handler.insert_blank_lines(Line(arg_or_default!(idx: 0, default: 1) as usize)),
+            ('l', intermediate) => {
+                let is_private_mode = match intermediate {
+                    Some(b'?') => true,
+                    None => false,
+                    _ => unhandled!(),
+                };
                 for arg in args {
-                    let mode = Mode::from_primitive(private, *arg);
+                    let mode = Mode::from_primitive(is_private_mode, *arg);
```
nixpulvis: TIL: https://github.com/rust-lang/rust/issues/62358

Ignore since this is a nightly only opt-in feature for now.

chrisduerr: Very interesting, didn't know about this one. Thanks for the info!

akinomyoga:
Thank you very much!

textshell:
Thanks for working on this. And sorry for being so late.

This seem to solve #2171. I wonder if the matching could be done on the whole intermediates list (limited in vte to 2 elements anyway iiuc), because at least one sequences alacritty might want to support does use two. E.g. `ESC [ ? Ⓝ $ p`. Of course thinking about that in this PR is strictly optional, it can be done without more work later too. I think allowing 2 would cover all sequences currently in use in xterm-like terminals.

textshell:
As for the question of implementing the selective erase sequences. I don't see much point in implementing them without their actual meaning just because they are the same without actually being able to set protection. And i don't think they play much of an role in the terminal ecosystem as far as i can tell. So if alacritty is ok in trying to prove that they are not actually needed that's nice too.

chrisduerr:
> This seem to solve #2171. I wonder if the matching could be done on the whole intermediates list (limited in vte to 2 elements anyway iiuc), because at least one sequences alacritty might want to support does use two. E.g. ESC [ ? Ⓝ $ p. Of course thinking about that in this PR is strictly optional, it can be done without more work later too. I think allowing 2 would cover all sequences currently in use in xterm-like terminals.

Since we do not have support for this yet, I don't think we should do that in this PR. But in theory it would be pretty simple, instead of doing `get(0)` we'd just have to take until there's nothing left and collect that to an array we can match instead of matching a single intermediate.

chrisduerr:
I've rebased this PR on top of master, sounds like everything's good now. Just waiting for CI to confirm the rebase didn't break anything. Thanks for your contribution.

Issue #2536, opened 2019-06-13
Feature Request: Hints for Regex


shibumi:
Hi,
I stumbled upon tmux-fingers. A really nice tmux plugin, that has a hint flag feature for several types of regexes. I thought about this a lot and I think it would be really cool if such features would be inside of alacritty: https://github.com/Morantron/tmux-fingers

What I would like to have is:
I type [ctrl]+[shift]+[space] [x] and i get hints for:

File paths
git SHAs
numbers ( 4+ digits )
hex numbers
IP addresses
kubernetes resources
UUIDs
URLs

then I can type a flag number in the terminal and i will copy the matched string

chrisduerr:
Absolutely love this and it has actually been planned and discussed for URLs at least for a way.

Somewhat related is https://github.com/jwilm/alacritty/issues/262 I guess.

In general I think Alacritty would do extremely well with a way to open URLs with the keyboard and select stuff with the keyboard, as an i3 user that would be incredibly helpful to me. Having the URL detection flexible enough to handle other stuff (like just regexes) too would be a nice extra benefit.

shibumi:
Smal addition to the Feature-Request: Would be nice to support Wayland (just switched to sway.. seems like it's the year of the wayland desktop)

d-vb:
Are there any updates on this one? This would be really cool.

rodolf0:
I built something related to this in https://github.com/jwilm/alacritty/pull/3126 it doesn't have keyboard navigation but it does recognize bits of text by matching regexes and allowing to click on them.

pwnorbitals:
Have we seen any progress on this issue recently ?

shibumi:
@chrisduerr  Thanks for your work :heart:

PR #2666, opened 2019-07-21
Fix cursor flickering on url hover

### LINKED ISSUE #2665: Mouse cursor flickers between default and Hand shape when hovering URL
## System
Archlinux kernel 4.19
rust 1.32.0
i3 on x11
alacritty master 76dd84b

## Behavior
When hovering across a highlighted URL, the cursor will flip back and forth between the default cursor for the context and the `Hand` shapes.

## Expectation
Mouse cursor shape should always be `Hand` when on top of an URL.

## Reproduce
Try `echo "https://google.com"` or similar command/input to get an URL
Move mouse cursor across the output, cursor should change shape as described

## Extra
This issue was discovered while trying to make an enhancement to text selection, and is introduced in 84aca67. The behavior described is first seen in a99547c.

More context in this pull #2661

kchibisov:
This commit fixes the regression introduced in
84aca672964e29b5b4503b7da7bc34fc395f08ab.

Fixes #2665.

`reset_mouse_cursor()` was called most of the time twice during mouse move actions and in case of hovering over already highlighted url, we were resetting `hand` cursor and then immediately setting it back,  this behavior was causing cursor "flickering".

### DIFF
--- a/alacritty_terminal/src/input.rs
+++ b/alacritty_terminal/src/input.rs
@@ -383,7 +383,30 @@ impl From<&'static str> for Action {
     }
 }
 
+enum MousePosition {
+    Url(Url),
+    MessageBar,
+    MessageBarButton,
+    Terminal,
+}
+
 impl<'a, A: ActionContext + 'a> Processor<'a, A> {
+    fn mouse_position(&mut self, point: Point) -> MousePosition {
+        if let Some(message) = self.message_at_point(Some(point)) {
+            if self.message_close_at_point(point, message) {
+                MousePosition::MessageBarButton
+            } else {
+                MousePosition::MessageBar
+            }
+        // Check for url should be after check for message bar, since we're not looking into
+        // message bar content.
+        } else if let Some(url) = self.ctx.terminal().url_search(point.into()) {
+            MousePosition::Url(url)
+        } else {
+            MousePosition::Terminal
+        }
+    }
+
     #[inline]
     pub fn mouse_moved(&mut self, x: usize, y: usize, modifiers: ModifiersState) {
         self.ctx.mouse_mut().x = x;
@@ -408,21 +431,38 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
             return;
         }
 
-        // Only report motions when cell changed and mouse is not over the message bar
-        if let Some(message) = self.message_at_point(Some(point)) {
-            self.update_message_cursor(point, message);
-
-            return;
-        } else {
-            self.ctx.terminal_mut().reset_mouse_cursor();
-        }
-
         // Don't launch URLs if mouse has moved
         self.ctx.mouse_mut().block_url_launcher = true;
 
-        // Underline URLs and change cursor on hover
-        if cell_changed {
-            self.update_url_highlight(point, modifiers);
+        match self.mouse_position(point) {
+            MousePosition::Url(url) => {
+                let mouse_mode =
+                    TermMode::MOUSE_MOTION | TermMode::MOUSE_DRAG | TermMode::MOUSE_REPORT_CLICK;
+
+                if self.mouse_config.url.mods().relaxed_eq(modifiers)
+                    && (!self.ctx.terminal().mode().intersects(mouse_mode) || modifiers.shift)
+                    && self.mouse_config.url.launcher.is_some()
+                {
+                    let url_bounds = self.url_bounds_at_point(url, point);
+                    self.ctx.terminal_mut().set_url_highlight(url_bounds);
+                    self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Hand);
+                    self.ctx.terminal_mut().dirty = true;
+                }
+            },
+            MousePosition::MessageBar => {
+                self.ctx.terminal_mut().reset_url_highlight();
+                self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Default);
+                return;
+            },
+            MousePosition::MessageBarButton => {
+                self.ctx.terminal_mut().reset_url_highlight();
+                self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Hand);
+                return;
+            },
+            MousePosition::Terminal => {
+                self.ctx.terminal_mut().reset_url_highlight();
+                self.ctx.terminal_mut().reset_mouse_cursor();
+            },
         }
 
         if self.ctx.mouse().left_button_state == ElementState::Pressed
@@ -445,50 +485,45 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         }
     }
 
-    /// Underline URLs and change the mouse cursor when URL hover state changes.
-    fn update_url_highlight(&mut self, point: Point, modifiers: ModifiersState) {
-        let mouse_mode =
-            TermMode::MOUSE_MOTION | TermMode::MOUSE_DRAG | TermMode::MOUSE_REPORT_CLICK;
+    fn url_bounds_at_point(&self, url: Url, point: Point) -> RangeInclusive<Linear> {
+        let Url { origin, text } = url;
+        let cols = self.ctx.size_info().cols().0;
 
-        // Only show URLs as launchable when all required modifiers are pressed
-        let url = if self.mouse_config.url.mods().relaxed_eq(modifiers)
-            && (!self.ctx.terminal().mode().intersects(mouse_mode) || modifiers.shift)
-            && self.mouse_config.url.launcher.is_some()
-        {
-            self.ctx.terminal().url_search(point.into())
+        // Calculate the URL's start position
+        let lines_before = (origin + cols - point.col.0 - 1) / cols;
+        let (start_col, start_line) = if lines_before > point.line.0 {
+            (0, 0)
         } else {
-            None
+            let start_col = (cols + point.col.0 - origin % cols) % cols;
+            let start_line = point.line.0 - lines_before;
+            (start_col, start_line)
         };
 
-        if let Some(Url { origin, text }) = url {
-            let cols = self.ctx.size_info().cols().0;
+        let start = Point::new(start_line, Column(start_col));
 
-            // Calculate the URL's start position
-            let lines_before = (origin + cols - point.col.0 - 1) / cols;
-            let (start_col, start_line) = if lines_before > point.line.0 {
-                (0, 0)
-            } else {
-                let start_col = (cols + point.col.0 - origin % cols) % cols;
-                let start_line = point.line.0 - lines_before;
-                (start_col, start_line)
-            };
-            let start = Point::new(start_line, Column(start_col));
+        // Calculate the URL's highlight end position
+        let len = text.width();
+        let url_end_col_denormilized = point.col.0 + len - origin;
 
-            // Calculate the URL's end position
-            let len = text.width();
-            let end_col = (point.col.0 + len - origin) % cols - 1;
-            let end_line = point.line.0 + (point.col.0 + len - origin) / cols;
-            let end = Point::new(end_line, Column(end_col));
-
-            let start = Linear::from_point(Column(cols), start);
-            let end = Linear::from_point(Column(cols), end);
-
-            self.ctx.terminal_mut().set_url_highlight(RangeInclusive::new(start, end));
-            self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Hand);
-            self.ctx.terminal_mut().dirty = true;
+        // This means that url ends at the last cell of the line
+        let end_col = if url_end_col_denormilized % cols == 0 {
+            cols - 1
         } else {
-            self.ctx.terminal_mut().reset_url_highlight();
-        }
+            url_end_col_denormilized % cols - 1
+        };
+
+        let end_line = if end_col == cols - 1 {
+            point.line.0 + (url_end_col_denormilized) / cols - 1
+        } else {
+            point.line.0 + (url_end_col_denormilized) / cols
+        };
+
+        let end = Point::new(end_line, Column(end_col));
+
+        let start = Linear::from_point(Column(cols), start);
+        let end = Linear::from_point(Column(cols), end);
+
+        RangeInclusive::new(start, end)
     }
 
     fn get_mouse_side(&self) -> Side {
@@ -908,15 +943,6 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         has_binding
     }
 
-    /// Set the cursor depending on where the mouse is on the message bar
-    fn update_message_cursor(&mut self, point: Point, message: Message) {
-        if self.message_close_at_point(point, message) {
-            self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Hand);
-        } else {
-            self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Default);
-        }
-    }
-
     /// Return the message bar's message if there is some at the specified point
     fn message_at_point(&mut self, point: Option<Point>) -> Option<Message> {
         if let (Some(point), Some(message)) =
@@ -945,11 +971,7 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
             ElementState::Pressed => {
                 if self.message_close_at_point(point, message) {
                     self.ctx.terminal_mut().message_buffer_mut().pop();
-                    if let Some(message) = self.message_at_point(Some(point)) {
-                        self.update_message_cursor(point, message);
-                    } else {
-                        self.ctx.terminal_mut().reset_mouse_cursor();
-                    }
+                    self.ctx.terminal_mut().reset_mouse_cursor();
                 }
 
                 self.ctx.clear_selection();
--- a/alacritty_terminal/src/event.rs
+++ b/alacritty_terminal/src/event.rs
@@ -410,6 +410,7 @@ impl<N: Notify> Processor<N> {
                             processor.ctx.terminal.next_is_urgent = Some(false);
                         } else {
                             processor.ctx.terminal.reset_url_highlight();
+                            processor.ctx.terminal.reset_mouse_cursor();
                             processor.ctx.terminal.dirty = true;
                             *hide_mouse = false;
                         }
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -896,6 +896,7 @@ impl Term {
     pub fn scroll_display(&mut self, scroll: Scroll) {
         self.grid.scroll_display(scroll);
         self.reset_url_highlight();
+        self.reset_mouse_cursor();
         self.dirty = true;
     }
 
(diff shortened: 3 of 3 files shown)


jjsullivan5196:
Tested it, this pull has the original behavior of v0.3.3, cursor is always `Hand` when hovering an URL

kchibisov:
> However, I had to search quite a bit to find out why it still works without this. Which is because of the reset_url_highlight call, which also resets the mouse cursor.

Yes, same problem for me here, just forgot to add this to PR message to make things easier for you. That's probably the case, why you(author of this issue) didn't notice(added) the explicit reset here during review.

>  seems like we could probably just reset the mouse cursor at the beginning of mouse moved

Unfortunately, I think this shouldn't work, we need to conditionally reset mouse right after `update_url_hightligh`, otherwise the behavior we'll be identical to the issue, I'm fixing here.

> or just set it in one place on
That's probably a good idea. I'm a bit conservative in fixing the regression issues. However it's already proposed to make `update_url_hightlight` to return a `bool` in #2661 (since it's a private function I think everything will be fine here). So we can probably move resetting logic from `update_url_hightlight` and place it in `move_moved` (add something like if `should_reset`).

>  I tested this and everything still seems to be working great, I can't find any regressions with the message bar or underlines.

Yeah, in addition I've tested under `tmux`/`htop` with message bars an other stuff (everything was fine), otherwise I wouldn't submit this bug fix.

chrisduerr:
@kchibisov I can't go into too much detail for now, but generally I don't like the idea of having `update_url_highlight` return a boolean. It doesn't really make sense to me, since it's supposed to do something, not check something. If we really would want to get that info, we should probably create a method just for checking it. But at the same time I question why we can't just put it in there. If it needs to be done whenever the URL highlight needs to be updated, then it should probably be in that method.

Regarding my idea, I thought we could reset the cursor at the beginning and then just have it change over the course of the mouse motion event. So it would only get unset when there's nothing else overriding it, similar to the window changes that we also 'stage' and then apply after the event loop run is done. However I don't know what exactly is causing the flickering, I haven't looked into the order of operations too much here, so if the issue is caused by the assumption that the cursor should be reset by default, then that obviously wouldn't work.

Giving the code a really quick glance, maybe it could just work to move the cursor reset to `if cell_changed` before the url highlight is updated. Then we should get the behavior that after a cell change the cursor is reset by default, but over a URL it should still work correctly. However this is just a quick guess for a change that would introduce a minimal set of changes. Then we could remove it from `reset_url_highlight`.

If you've got any better ideas, feel free to either run them by me or just try them out and I can see what I think. Though if you need more detailed feedback, that'll likely take me till thursday.

kchibisov:
> However I don't know what exactly is causing the flickering
```
`https://github.com`
~~~~~~~~~~~^
```
Assume you're hovering url with mouse on `t` key as on a "picture". Your cursor icon will be changed to `Hand`. Then you've decided to move one cell left. Here what ig going to happen:
1) `mouse_moved` will be called.
2) You've changed cell, so `cell_changed` will be set to true. So `self.ctx.terminal_mut().reset_mouse_cursor()` in `mouse_moved` will be called, which change your cursor from `Hand` to `Text`.
3) But your cell was changed so we call `self.update_url_highlight(point, modifiers)` and since we are **hovering** over url, we'll call `self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Hand)` in `update_url_hightlight`. Our cursor will be changed from `Text` to `Hand`.

As you can see, you're changing cursor from `Hand` to `Text` and then immediately to `Hand`, which results in cursor flickering when hovering over url. 

In case you've moved from `t` one cell down, so you're leaving url.

1) `mouse_moved` will be called.
2) You've changed cell, so `cell_changed` will be set to true. So `self.ctx.terminal_mut().reset_mouse_cursor()` in `mouse_moved` will be called, which change your cursor from `Hand` to `Text`.
3)  But your cell was changed so we call `self.update_url_highlight(point, modifiers)` and since we're **leaving** url, we'll call `self.ctx.terminal_mut().reset_url_highlight()` in `update_url_hightlight`. Our cursor will be changed from `Text` to `Text`.

Here you're changing your cursor from `Hand` to `Text` and then once again to `Text`, which is redundant.

I guess, that should make it clear without reading code.

> If you've got any better ideas, feel free to either run them by me or just try them out and I can see what I think.

So, my idea ( just what "I" think, regarding being conservative) here is to make
`update_url_highlight(&mut self, point: Point, modifiers: ModifiersState)` to return `Option<RangeInclusive::new(start, end)>` and rename it to `find_url_at_point`/`url_bounds_at_point` or something similar and move the following logic (maybe not all?) to `mouse_moved` since it'll make things so much clear in my opinion.
```rust
  self.ctx.terminal_mut().set_url_highlight(RangeInclusive::new(start, end));
  self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Hand);
  self.ctx.terminal_mut().dirty = true;
} else {
  self.ctx.terminal_mut().reset_url_highlight();
}
```

Also maybe moving this block
```rust
let mouse_mode =
    TermMode::MOUSE_MOTION | TermMode::MOUSE_DRAG | TermMode::MOUSE_REPORT_CLICK;

// Only show URLs as launchable when all required modifiers are pressed
let url = if self.mouse_config.url.mods().relaxed_eq(modifiers)
    && (!self.ctx.terminal().mode().intersects(mouse_mode) || modifiers.shift)
    && self.mouse_config.url.launcher.is_some()
 {
    self.ctx.terminal().url_search(point.into())
} else {
    None
 };
```
 somewhere else (separate function or `mouse_moved`) will also make sense here, since will end up with a "generic_function" to get url bounds which can be reused, when we decide to implement this #2536 , I guess.

@chrisduerr As I understand you're planning to integrate https://github.com/chrisduerr/rfind_url into `alacritty`. From this perspective, what do you think on splitting `update_url_highlight` to smaller functions, since now it can be used only in `mouse_moved`?

jjsullivan5196:
@kchibisov I think the idea you have there makes a lot more sense. @chrisduerr I agree what I tried to do with `bool` is a code smell, I just didn't want to start hacking a bunch of stuff without review.

@kchibisov would you mind if I gave your suggestion a try in my PR? I have enough of an idea of how this is working from your description and what I've seen in `input.rs`.

kchibisov:
> @kchibisov would you mind if I gave your suggestion a try in my PR? I have enough of an idea of how this is working from your description and what I've seen in input.rs.

If you can't wait or want to practice some rust you can post here your implementation on what I've proposed, since your original PR is doing other things and fixing + implementing a feature in one PR is a bad idea.

I'd like to wait for @chrisduerr  opinion on my proposal though. I didn't implemented it right after I've posted a comment, because there is no need to hurry on this thing (read @chrisduerr comment) and because it was 2:30 am and I want to chill/sleep/"think about the solution I've proposed here" a bit before work.

chrisduerr:
I'll definitely be able to play around with a bit of code myself tomorrow, so I'd like to hold out till then. But if anyone else wants to give it a shot, I can give it a quick read through when I can find some time. Though since we're just talking about one or two days here, there might not be any need to rush things.

Originally I forgot that we're calling through to winit directly though, which of course doesn't relate to our event loop and redraw frequency at all. So if we switch multiple times during a frame it will actually flicker instead of just repainting on the frame boundary. So maybe locking things down can help a bit, though I think restructuring stuff is likely also very promising.

jjsullivan5196:
Figured it would be a good idea to wait, was just a bit eager :-)

I'll just wait until another pull re-factors this bit of the code, I'll work #2661 around that. Maybe I'll post a snippet or something like @kchibisov suggested.

chrisduerr:
Okay @kchibisov, after taking a look at some of the code, your solution seems reasonable to me, however I don't think it solves the problem at hand.

Splitting apart the `update_url_highlight` function makes sense, generally the method behaves in a bit of an ugly way, being seemingly disconnected from everything else, while still significantly mutating the state of the terminal.

However, even after splitting the method, I can still see it being difficult to convey in which way the mouse cursor is going to be mutated. I don't think it makes sense to enforce at which place the call to reset it should be made, that itself seems like a bit of a code smell.

Generally I think the problem itself is a bit different. The problem isn't necessarily how resetting the URL is called, but rather that the URL cursor reset is also responsible for resetting the cursor when over the message bar, which took me completely by surprise. This is because we always reset the URL highlight state, unless we hover over a URL.

I'm unsure about what the best possible solution for this is though. One option would be to 'stage' these changes, so instead of applying the cursor change immediately, we just do it at the end of the mouse motion function, at the end of the event processing or when redrawing.

Another possible solution would be to `match` a mouse position state. So we'd create a method that returns if the mouse cursor is over a URL, over the message bar, or anywhere else. In that case it would likely be pretty simple to separate the different cases and reset the cursor explicitly whenever required.

Of course the way it is right now also works and I'm unsure if there's going to be a significant improvement with any of the proposed changes so far. So I'm almost inclined to keep the PR the way it is right now, or just make a very minimal but potentially suboptimal change, rather than reworking a chunk of things and creating more of a mess than a solution.

@kchibisov what do you think about the issue that the mouse cursor reset is not clearly communicated to affect the message bar too, not only the URL state? Do you think your proposal improves that, or do you also think that this still comes with some troubles? Also do you have any simple fixes you can think of that would clear this situation up without reworking everything?

kchibisov:
> what do you think about the issue that the mouse cursor reset is not clearly communicated to affect the message bar too, not only the URL state?

Probably all cursor manipulations (setting icon/resetting) should be in one "block"(match statement or function). So If we took something I've proposed and add some ideas from "here"

> nother possible solution would be to match a mouse position state. So we'd create a method that returns if the mouse cursor is over a URL, over the message bar, or anywhere else. In that case it would likely be pretty simple to separate the different cases and reset the cursor explicitly whenever required.

and perform cursor icon fallback url -> message_bar_button -> message_bar -> terminal, that would be something clear (we move a mouse to location, we've found what's there, and set to appropriate cursor directly in mouse_moved ( or some support function).

> Do you think your proposal improves that, or do you also think that this still comes with some troubles?

My proposal clarifies things about urls hightlight, trying to reset cursor in a place where it make sense and expected, and probably make some help with integrating `rfind_url`, since I have a feeling, that you'll write some similar functions at some point (just to highlight multiple urls on key press). However we can improve overall situating with handling mouse interactive elements in alacritty.

> So I'm almost inclined to keep the PR the way it is right now, or just make a very minimal but potentially suboptimal change, rather than reworking a chunk of things and creating more of a mess than a solution.

I'm also fine with proposed solution, however adding one more handler to `mouse_moved` will introduce a lot of hassle and code smell. It already introduced it, when message bar handling was added, because this call to reset mouse in url highlight is not something obvious and we've ended up with 2 calls. So implementing some reliable mechanism for such handling would make sense in a long term.  

> Also do you have any simple fixes you can think of that would clear this situation up without reworking everything?

Move things a bit? like
```rust
if url { set to url } 
else if message_bar { set_to_message_bar } 
else { default }
```
My proposed solutions in this comment is something similar to this, but with additional refactoring to a function, that doesn't make sense.

chrisduerr:
> I'm also fine with proposed solution, however adding one more handler to mouse_moved will introduce a lot of hassle and code smell. It already introduced it, when message bar handling was added, because this call to reset mouse in url highlight is not something obvious and we've ended up with 2 calls. So implementing some reliable mechanism for such handling would make sense in a long term.

I think this is irrelevant to the discussion, but I don't quite understand what you mean by `handler` in this context. What do you mean by `reliable mechanism for such handling`?

> My proposal clarifies things about urls hightlight, trying to reset cursor in a place where it make sense and expected, and probably make some help with integrating rfind_url, since I have a feeling, that you'll write some similar functions at some point (just to highlight multiple urls on key press). However we can improve overall situating with handling mouse interactive elements in alacritty.

I'd like to completely ignore future changes for now. Usually that just creates additional complexity and often never pays off in the long run. So I'd like to solve the problems we have right now, instead of the problems we might face in the future.

> Move things a bit? like
>```rust
> if url { set to url } 
> else if message_bar { set_to_message_bar } 
> else { default }
> ```

I think that would very closely match up with my `match` proposal, so maybe that makes sense. I personally think that a method that just outputs above which element the cursor is when moved could be very beneficial. Using an enum that could potentially be pretty clean:

```rust
match where_are_thou() {
    Url => ...,
    MessageBar => ...,
    _ => ...,
}
```

If you think that makes sense and you're willing to give it a shot, I'd be very curious how that would turn out. Otherwise I'd just go for what we currently have and keep changes to future refactorings. At the end of the day, this is about fixing a bug.

kchibisov:
> I don't quite understand what you mean by handler in this context. What do you mean by reliable mechanism for such handling?

I mean, if we decided to add FancyMessageBar or git_hash or whatever e.g.. match or if else solutions should handle such things fine and clear.

> If you think that makes sense and you're willing to give it a shot, I'd be very curious how that would turn out. 

This makes sense, because of #2661 waiting at the corner, and it doesn't seem like something overly complex.

Probably we need `if else` + `match` + `flag for return`?
`if else` for probing locations, since we want to expose that some element dominates another.
`match` for quick select
`flag for return` if we want to some logic

So should be like this, I guess.

```rust
let are_we_term_gui_element_yet = if url {} else if message_bar {} else {};
let leaving_this_mouse_thing = match are_we_term_gui_element_yet {
    Url => ...,
    MessageBar => ...,
    _ => ...,
};
/* Some handling I don't remember, but it will be here */
if leaving_this_mouse_thing {
 return;
}
```

chrisduerr:
That seems overly complicated to be honest. I see no reason why we'd have `leaving_this_mouse_thing` separated instead of just returning inside the match for example.

The whole `if/else if/else` is probably mainly confusing since it's pseudocode, but just to be clear I wouldn't first store the status in a variable and then use those to determine the state, I'd just create a method which goes through all checks and returns the mouse position state. There's no need to check for URLs when the mouse is over the message bar after all.

alacritty_terminal/src/input.rs
```diff
@@ -443,50 +484,45 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
-            self.ctx.terminal_mut().set_url_highlight(RangeInclusive::new(start, end));
-            self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Hand);
-            self.ctx.terminal_mut().dirty = true;
+        let end_line = if end_col == cols - 1 {
+            point.line.0 + (url_end_col_denormilized) / cols - 1
         } else {
-            self.ctx.terminal_mut().reset_url_highlight();
-        }
+            point.line.0 + (url_end_col_denormilized) / cols
+        };
```
kchibisov: There was a bug, when url ends at the last column there was an unsigned overflow (0 - 1).

alacritty_terminal/src/input.rs
```diff
@@ -445,50 +484,45 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
-                let start_col = (cols + point.col.0 - origin % cols) % cols;
-                let start_line = point.line.0 - lines_before;
-                (start_col, start_line)
-            };
-            let start = Point::new(start_line, Column(start_col));
+        let start = Point::new(start_line, Column(start_col));
 
-            // Calculate the URL's end position
-            let len = text.width();
-            let end_col = (point.col.0 + len - origin) % cols - 1;
```
kchibisov: There was a bug, when url ends at the last column there was an unsigned overflow (0 - 1).

kchibisov:
Naming maybe is a little bit odd, but I like the result of this small refactoring.

@chrisduerr What do you think about removing call to `reset_mouse_cursor` in `reset_url_highlight`? So each function will end up doing their job as their names advertise.

alacritty_terminal/src/input.rs
```diff
@@ -408,21 +431,37 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+                self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Default);
+                return;
+            },
+            MousePosition::MessageBarButton => {
+                self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Hand);
+                return;
+            },
+            MousePosition::Terminal => {
+                // `reset_url_highlight` calls `reset_mouse_cursor`, so no need to perform explicit
+                // call to the second one.
```
chrisduerr: Instead of adding this as a comment, couldn't we just remove `reset_mouse_cursor` from `reset_url_highlight` and make this a little more explicit?

alacritty_terminal/src/input.rs
```diff
@@ -408,21 +431,37 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+                self.ctx.terminal_mut().reset_url_highlight();
```
chrisduerr: I haven't tested this, but I'm a bit concerned about calling `reset_url_highlight` only when the mouse is over the terminal. What if the URL is in the last line and there is a message bar open? Wouldn't that cause the URL to stay highlighted until the mouse is moved to the terminal again?

I'd consider an `if/else` block here too where the `else` always resets url highlight, however I think that might get very deeply nested and convoluted. So I think I'd just go for putting `reset_url_highlight` in every block and making sure it doesn't reset the mouse cursor itself.

alacritty_terminal/src/input.rs
```diff
@@ -445,50 +484,45 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         }
     }
 
-    /// Underline URLs and change the mouse cursor when URL hover state changes.
-    fn update_url_highlight(&mut self, point: Point, modifiers: ModifiersState) {
-        let mouse_mode =
-            TermMode::MOUSE_MOTION | TermMode::MOUSE_DRAG | TermMode::MOUSE_REPORT_CLICK;
+    fn url_bounds_at_point(&self, url: Url, point: Point) -> RangeInclusive<Linear> {
```
chrisduerr: This method is a bit messy I think and `url_bounds_at_point` might be a little hard to understand, however I think with the rfind_url changes I've made it should be possible to simplify this down to something more straight-forward like `url_bounds(url) -> Range<Linear>` which would make it pretty clear.

So I think this is good to go for now and I'd just worry about potential improvements to this with the other PR which I'd like to land after this one.

alacritty_terminal/src/input.rs
kchibisov: Yeah, I've already asked you about this one ;- ).

alacritty_terminal/src/input.rs
kchibisov: Good one, I'll test this scenario and update.

> reset_url_highlight in every block and making sure it doesn't reset the mouse cursor itself.

After removing `reset_mouse_cursor` from `reset_url_highlight` everything should be very simple to handle.

alacritty_terminal/src/input.rs
```diff
@@ -408,21 +431,38 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+                return;
+            },
+            MousePosition::MessageBarButton => {
+                self.ctx.terminal_mut().reset_url_highlight();
+                self.ctx.terminal_mut().set_mouse_cursor(MouseCursor::Hand);
+                return;
+            },
+            MousePosition::Terminal => {
+                self.ctx.terminal_mut().reset_url_highlight();
+                self.ctx.terminal_mut().reset_mouse_cursor()
```
chrisduerr: `;` missing at the end of the line here. Just as a minor style thing.

Issue #50, opened 2017-01-04
Support for ligatures


fuine:
I would love to see the support for ligatures (an example of their use might be found in https://github.com/tonsky/FiraCode). Please note that i have almost no idea how do they work under the hood, and so i don't know how hard is it to implement support for them, I'm just throwing an idea here.

Anyhow i think this is a great project, congratulation! :)

jwilm:
I would really like to support this! It falls into the categories of things Alacritty cares about which include quality font rendering and performance.

> Anyhow i think this is a great project, congratulation! :)

Thanks :)

aperezdc:
Support for ligatures would be indeed super nice. One terminal emulator that handles them quite well in my experience is [Pangoterm](http://www.leonerd.org.uk/code/pangoterm/), though I have not looked at the code to see whether it does something explicitly or it just delegates the task to [Pango](http://www.pango.org/) — anyway, leaving the link here as reference, in case it's useful to gather inspiration from its code.

For reference, the following screenshot shows an interesting behaviour of Pangoterm: ligatures are *not* handled in the lines/cells where used input is done, but only in the ones where output is shows. I suppose that it's easier to handle because it's not needed to backtrack and redraw the line where the cursor is to have ligatures applied during input:

![screenshot from 2017-01-23 12-04-35](https://cloud.githubusercontent.com/assets/723451/22199898/8c310b12-e165-11e6-9c0e-79bba1523ab6.png)

(Pangoterm on top, Alacritty on bottom. Of course, if the same line used for input receives output from a program e.g. Vim, then ligatures “appear”.)

michelboaventura:
I use [qterminal](https://github.com/lxde/qterminal), which has full support for ligatures and seems to have a more recent implementation of a terminal. I think it's a nice example of how to handle ligature (and unicode in general).

3goliad:
Qterminal and Konsole share the same implementation, AFAIK, and correctly render ligatures in input lines, excepting when under the cursor.
For further reference:
![cropped ligatures in konsole](https://cloud.githubusercontent.com/assets/20362502/22620945/a7e15214-eb0f-11e6-8f01-de7c4256f9b4.png)

f-koehler:
@3goliad I use konsole and always considered this behaviour to be intended. It would be nice to have a similar implementation in alacritty.

mrkgnao:
I'd love this too!

zovt:
Anyone currently working on this?

[...]

PR #2677, opened 2019-07-26
Support for Font Ligatures using Harfbuzz on Linux

thunderseethe:
Initial support for font ligatures using harfbuzz and free type on linux.
This builds on the work done by @anirudhb in [[WIP] Basic text shaping and rendering using Harfbuzz](https://github.com/jwilm/alacritty/pull/2181)
Step in the direction towards closing #50 .

This is relatively early in it's implementation but I wanted to get some eyes on it as changes past this point will start to effect core pieces of infrastructure (rendering etc.).

Currently code will render ligatures on linux with a monospace font (I've only tested with Fira Code).

image

## Major Changes 
---
* TextRun replaces RenderableCell as main method of rendering text
* Harfbuzz used to shape runs of text (allowing ligatures)
* GlyphKey uses u32 (glyph index) instead of char

Support for font ligatures is enabled by changing the base rendering unit from a RenderableCell to a TextRun (name flexible). A TextRun bundles a series of RenderableCells that share the same render properties (line, fg, bg, etc.) into one unit. Each TextRun gets shaped by harfbuzz returning the glyph indices for each character.  From here each glyph is rasterized as normal through the glyph cache.

Changes are guarded behind the `hb-ft` feature in `alacritty_terminal`. 
Along with adding new code under this feature, some code is removed when this feature is turned on. Mainly code dealing with rendering RenderableCell's (and GlyphKey's c member).
The idea behind this is to ensure that there aren't any unknown uses of RenderableCell in the codebase. Once the code is free of RenderableCell when hb-ft is on, more intrusive changes can be made to rendering that break the conventions set by RenderableCell.

## Caveats
---
* Only works on linux.
* Incurs performance hit of shape() call for each text run for each draw call (haven't done any testing yet to see how steep this is).
* Ignores position information from harfbuzz, this means it probably only works for monospace ligatures (Fira Code) since each ligature is still rendered as if it's in it's own cell.
* Each Face object carries an instance of hb_font with it, in addition to `freetype::Face`, this doesn't seem hugely problematic but is should definitely be noted.
* Post shaping work is mainly done in Glyph Indices, this makes it hard to employ the same font recovery technique currently used by `load_face_with_glyph` as we don't have a char available to guess at the font. This functionality can be recovered by keeping the char stored with it's indice however this means `GlyphKey` would no longer fit in a `u64`.

## Questions
---
* Should TextRuns be cached similar to how GlyphKeys are currently?
* Can this approach be adapted to Mac OSX and Windows, and if it can how different does the code need to be?
* Currently code grabs a `Vec<RenderableCell>` from `grid` and wraps that as a `TextRunIter` to produce `TextRun`s. Would it make more sense for `grid` to directly produce a `TextRunIter` and avoid the unecessary middle allocation (hopefully)?
* Should alacritty only support monospace ligatures and ignore "proper" ligatures?
* Can `TextRun` completely replace `RenderabelCell`?
* Does it make sense to use glyph index in `GlyphKey` (instead of char)?

## Remaining Work
---
* Rendering and Atlas construction should hopefully be able to take more advantage of TextRuns then they are currently (none).
* Font ligatures should be a configurable option, this is a matter of enabling/disabling the harfbuzz feature based on a bool.
* Harfbuzz supports building a font from a freetype Face directly, however none of the rust bindings expose that functionality currently.
* Glyph Index is represented by a raw u32, this should be replaced with a newtype wrapper such as `GlyphIndex(u32)` similar to Line and Column.
* Benchmarking performance with new TextRun implementation

I'm relatively new to the open source community so my apologies if something is out of place or I've missed a process I should've completed.

I initially ran my code through `cargo fmt` however it changed a ton of the codebase that I hadn't touched so I undid those changes. Is there a common `rustfmt.toml` config the repo uses?


kchibisov:
Cool, some bugs I've found just without looking at code.
- Unicode font rendering seems broken most of the time.
- Underline/Strikethrough rendering is also broken.

So here is a test case for you that must work: `printf "\e[4;9m日本人\n"`.

The output should be(master):
good

Your PRs output:
broken

I can copy this text though.

I've also ran some benchmarks with `vtebench` + `perf` and everything was fine( less then 1%  deviation from the master on my system), except `unicode-random-write`, which affected by bugs I've mentioned (rendering is wrong, but time is fine), so to gather proper results, rendering should be consistent. We probably need more benchmarks with unicode involved.

thunderseethe:
Thanks for the fast turn around and succinct test case. I've resolved the issue with strikethrough/underline and they should now be working as intended.

As you noted the issue with not rendering unicode characters was due to a lack of glyphs in the font (at least for me on Fira Code). I was able to properly render unicode glyphs that Fira Code does contain (↑ and →). 

Either way I have added back the `KeyType` enum to `GlyphKey`. It is either a `GlyphIndex(u32)` or it is a `Fallback(char)`. In the Fallback case font loading assumes the glyph is missing from current font and directly tries to load a font for this char. This is a small departure from the current (master) logic which looks up the char in the ft face before trying to load a font.

unicode_fallback

Above is an example of font fallback working. Font is still Fira Code and the unicode characters use the fallback font to render (this also is an example of strikethrouhg/underline behaving ^_^).

Right now fallback case is slower than necessary as it iterates over the text run to find each character (as opposed to iterating over text run once). Also GlyphKey is using derived PartialEq, and Hash impls for the time being.

I plan to remove the unnecessary iteration in a future commit. It is unclear to me the best way to move forward with Hash/PartialEq for GlyphKey operating on the assumption it doesn't fit in a u64. I'd appreciate any guidance on that.

kchibisov:
The performance is good(same deviation I've reported before, but now on all tests) on my system and fallback seems to work fine. Functionality mimics master at the moment, as far as I can see. We probably need more benchmarks to [vtebench](https://github.com/jwilm/vtebench) (like drawing a lot of random ligatures and scrolling).


> It is unclear to me the best way to move forward with Hash/PartialEq for GlyphKey operating on the assumption it doesn't fit in a u64. I'd appreciate any guidance on that.

For now, it's fine. We should probably look at this once again, when the PR will be almost ready and the last performance measurements will be performed.

> Font ligatures should be a configurable option, this is a matter of enabling/disabling the harfbuzz feature based on a bool.

Implementing this option early on, will probably make sense (you should add `ligatures`(or smth like that) option to font configuration with default set to `false`).

> Is there a common `rustfmt.toml` config the repo uses?

Yes, in the root of the project, just run `rustfmt <file>`, It was added a bit later, so some files gets additional reformatting (from my experience on running `rustfmt` on files I'm editing in `alacritty`), which you should probably undo yourself.

> Harfbuzz supports building a font from a freetype Face directly, however none of the rust bindings expose that functionality currently.

If rust binding you're using here don't expose this functionality, you should probably open an issue and post a link to it in this PR, if you want to exploit this behavior.

> Benchmarking performance with new TextRun implementation

We're using [vtebench](https://github.com/jwilm/vtebench) forthis. I'm running it with `perf stat -r 10 vtebench -b 10000000 <test_name>`, however you can use [hyperfine](https://github.com/sharkdp/hyperfine), which is more about benchmarking then perf (I've just got used to run perf). 

P.s.
You've also fixed #2680 in "hb-ft" feature (not with this feature turn off). Don't worry about fixing it outside your feature, just noted.

chrisduerr:
> Font ligatures should be a configurable option, this is a matter of enabling/disabling the harfbuzz feature based on a bool.

I'm a bit concerned about maintaining two completely separate ways to do the same thing if all we want to do is display ligatures or not. That's gonna introduce a lot of complexity and make things very hard to understand, especially if it is done this way on all platforms, leading to basically four different ways to do the same thing.

thunderseethe:
I should clarify, the config option I'm thinking of would be passed down to Harfbuzz and would toggle whether the 'liga' feature is used during shaping. This would still use the new TextRun way of rendering things and would mostly exist in case a user is using a font with ligatures but wishes not to see them.

I'm in agreement that we would ideally only have one way to turn text to glyphs and put them on a screen.

thunderseethe:
I've added a config option named "use_font_ligatures" that toggles Harfbuzz's 'clig' and 'liga' feature. This is a conservative choice and could be expanded up on to toggle 'hlig', 'dlig' and 'slig'. One that should not be included in this toggle is 'rlig' as thing under this category are required ligatures and are necessary for the font.

A caveat is that in it's current manifestation this actually doesn't turn off ligatures in Fira Code (I have not tested other fonts). Fira Code's "ligatures" are all done as glyph substitutions under the 'calt' (Contextual Alternatives) feature. I don't know how widely used this feature is in standard fonts and have concerns toggling it via "use_font_ligatures" could have surprising and unexpected results in fonts other than Fira Code.

I've also done some light benchmarking with the command `perf stat -r 100 vtebench -w $(tput cols) -h $(tput lines) -b 10000000 unicode-random-write`. Results are inline with @kchibisov's results so far:
master: `0.85382 +- 0.00405`
font-ligatures: `0.83850 += 0.00286`

chrisduerr:
> Results are inline with @kchibisov's results so far:
> master: 0.85382 +- 0.00405
> font-ligatures: 0.83850 += 0.00286

Please note that runtimes of under a second are insufficient of properly testing Alacritty in my experience. To get a reliable performance benchmark, it is necessary to run for at least around 5-10 seconds and instead of running it through `vtebench` directly, it is necessary to use a file buffer.

The `unicode-random-write` benchmark is also **extremely** slow in Alacritty. So other benchmarks like the `alt-screen-random-write` should be preferred.

thunderseethe:
> I don't see what other fonts would do with this. We should not do any character substitution at all, except for ligatures. Everything else should be treated as separate glyphs the way it is currently done in Alacritty.

I took a look at some other popular fonts that have monospace ligatures for common programming symbols. I looked at: 
* [Hasklig](https://github.com/i-tu/Hasklig)
* [Hask](https://sourcefoundry.org/hack/)
* [Iosveka](https://github.com/be5invis/Iosevka)
* [Monoid](https://larsenwork.com/monoid/)
* [Victor Mono](https://rubjo.github.io/victor-mono/)

Of these fonts only Victor Mono does ligatures via the open type 'liga' feature. The rest of the fonts use the 'calt' featire to do glyph substitutions the same way Fira Code does.

This is not statistically significant or anything. However, if the idea behind ligature support is to support these kind of monospaced programming fonts (as opposed to like sanskrit or text that flows vertically) then 'calt' will have to be toggled to get the desired behavior.

chrisduerr:
> opposed to like sanskrit or text that flows vertically

There is no desire to properly support vertical or RTL text as far as I'm concerned.

thunderseethe:
Do we want to support non monospace ligatures? That one will involve some hefty changes to rendering since it breaks rendering each cell at a set position based on column and line.

It also doesn't need to happen in this pull request (although I certainly don't mind taking a stab at it). This is already a size-able modification.

I've ran some more benchmarks with a larger data set.  
`perf stat -r 100 vtebench -w $(tput cols) -h $(tput lines) -b 100000000 alt-screen-random-write`:
 * font-ligatures: `3.0292 +- 0.0328 seconds time elapsed  ( +-  1.08% )`
 * master: `3.06846 +- 0.00980 seconds time elapsed  ( +-  0.32% )`

`perf stat -r 100 vtebench -w $(tput cols) -h $(tput lines) -b 100000000 unicode-random-write`:
 * font-ligatures: `8.1640 +- 0.0180 seconds time elapsed  ( +-  0.22% )`
 * master: `8.2763 +- 0.0134 seconds time elapsed  ( +-  0.16% )`

chrisduerr:
> Do we want to support non monospace ligatures? That one will involve some hefty changes to rendering since it breaks rendering each cell at a set position based on column and line.

I don't think so.

The benchmarks look pretty good, though I'm a bit skeptical that there's so little change here to be quite honest. So at the moment I'm a bit suspicious that things might have slowed down in areas that aren't perfectly covered by benchmarks. I'll try to give it a run for its money myself though once it's all looking good.

@thunderseethe Is there anything left to do from your perspective? This has been a bit of a work in progress, so I've been holding out on reviews until you think it's good. So please just let me know when you're waiting for a full review so I don't forget about it.

thunderseethe:
@chrisduerr It should be ready for review now.

I agree with your assessment that the benchmarks look too good to be true. I ran a normal perf on one of the benchmarks and calls out to harfbuzz were a significant amount of cpu time in the result. 

I plan to try some bench marks that use the new code except don't call harfbuzz shape to see if they're magically faster for some reason and that might explain things.

nixpulvis:
Just to assist anyone else trying to test this, I found it difficult to run with the feature enabled out of the box. So I've added a feature to the `alacritty/Cargo.toml` as well:

```diff
+hb-ft = ["alacritty_terminal/hb-ft"]
```

Then, sadly due to rust-lang/cargo#5015, we can't simply run use this feaure at the top level workspace, so I've found `cargo run --manifest-path alacritty/Cargo.toml --features hb-ft` to work. Given that I might expect this feature to be feature gated until it's supported on all platforms (thoughts @chrisduerr, @jwilm) this might be worth updating.

I'll have to play with this a bit more, but upon first glance it's working quite well. Although I generally don't like ligatures in my programming fonts, it's still very cool watching this work. This might have inspired me to start looking for a minimally invasive set of ligatures to have on top of Fira Mono, but I digress.

In response to non-monospace fonts: I'm perfectly happy landing monospace ligatures before anything else, and given how poorly non-monospace rendering looks inside most terminals I don't see a dire need to support this. But I would like to go on record saying that I'd be happier if Alacritty did support non-monosapce fonts, and I don't seem to see an open issue for this. What reason would we have to turn down this feature?

thunderseethe:
@nixpulvis I've been running up against that same issue. My hackneyed solution was simply to keep toggling the default features of alacritty_terminal, far from optimal. Thank you for a more reasonable solution!

font/src/lib.rs
```diff
@@ -135,13 +142,46 @@ impl FontKey {
     }
 }
 
+#[cfg(feature = "hb-ft")]
+pub mod key_type {
```
chrisduerr: Why is this a separate module? Just for the feature restriction?

font/src/lib.rs
```diff
@@ -135,13 +142,46 @@ impl FontKey {
...
+        fn from(val: char) -> Self {
+            Fallback(val)
+        }
+    }
+}
+#[cfg(feature = "hb-ft")]
+use key_type::KeyType;
+
+// For now use derived impls, this can be swapped for more efficient implementations once things are
+// working
```
chrisduerr: Things are mostly working, right? Should this actually be implemented manually?

If this is not performance relevant, I'm fine with leaving it derived and removing the comment to shorten the code.

Otherwise this should be implemented before merging.

font/src/lib.rs
```diff
@@ -267,3 +321,27 @@ pub trait Rasterize {
     /// Update the Rasterizer's DPI factor
     fn update_dpr(&mut self, device_pixel_ratio: f32);
 }
+
+/// Config option specific to the Rasterizer.
+/// Since the Rasterizer lives in the subcrate font we do not want it to depend on the Font config
+/// struct from alacritty, as this would introduce a circular dependency between the crates.Clone
+/// This struct specifies the subset of Font that the Rasterizer cares about, then traits can be
+/// used to convert from Font to this struct when constructing a Rasterizer.
```
chrisduerr: I don't think we need this comment. It should be understood that crates should not depend on each other.

Rather than documenting this, the doc comment for this configuration should describe what it is used for.

> Config option specific to the Rasterizer.

I think the wording here specifically is already a bit odd. It's not a `config option`, it's the configuration for the rasterizer. Crates in the workspace should all be treated largely independently, so please comment this in a way you would for a crate completely unrelated to Alacritty.

font/src/lib.rs
```diff
@@ -267,3 +321,27 @@ pub trait Rasterize {
...
+#[cfg(feature = "hb-ft")]
+pub struct RasterizeConfig {
+    /// Toggle thin strokes on mac osx
+    // Technically this is impossible while under "hb-ft" but is included for compatiblity in the
+    // api
```
chrisduerr: If it is impossible, we should remove `Technically`. If there is a way to somehow do this, we should implement it.

I'd also be curious if there's any upstream tracking issue with freetype for this. Is this not relevant for harfbuzz at all? Then we probably don't even need it, right? In which case we could also just pass `use_font_ligatures` alone without the `RasterizeConfig` struct?

font/src/lib.rs
```diff
@@ -267,3 +321,27 @@ pub trait Rasterize {
...
+    /// Toggle rendering of font ligatures
+    pub use_font_ligatures: bool,
+}
+
+// Only implemented for the FreeType rasterizer so far.
+/// Conceptually this extends the Rasterizer trait with Harfbuzz specific functionality.
+#[cfg(feature = "hb-ft")]
+pub trait HbFtExt {
+    /// Shape the provided text into a set of glyphs.
+    /// TODO: properly report HarfBuzz errors
```
chrisduerr: This is something we should probably do before merging.

We need to be clear about which errors we want to ignore/handle ourselves, which we want to just crash on and which we want to propagate to Alacritty.

thunderseethe:
Thank you for the review. I'll get cooking on fixing up the code based on your comments. Before I dive into that I wanted to ask. I'm having a lot of trouble getting rustfmt to work on my code. I've run it on each file I've touched, a couple of times.

It appears to make changes on some runs and undo/make different changes on another run. Also I'm working on stable so it's quite possibly not even using a lot of the config options specified. What is the normal workflow here? Is there something obvious I'm missing and/or is there a way I can format my code via `cargo fmt` or rls directly instead of manually running `rustfmt`?

I'm all for standard formats and would very much like to conform to whatever style the codebase is using, but so far that has proved difficult.

chrisduerr:
> It appears to make changes on some runs and undo/make different changes on another run. Also I'm working on stable so it's quite possibly not even using a lot of the config options specified. What is the normal workflow here? Is there something obvious I'm missing and/or is there a way I can format my code via cargo fmt or rls directly instead of manually running rustfmt?

You'll need to run nightly to format the code. Then just run `cargo fmt` in the project root. I'm fine with doing this for you too if you don't want to setup nightly, but usually it's something that you might want to do yourself since it often cleans up the code a bit.

thunderseethe:
Cool, I'm down to run `cargo fmt` and happy to do it on nighly. I'll give that a shot. Perhaps it was on an old version of master but when I originally ran `cargo fmt` it touched ~100 files most of which I had not changed and I've been too scared to try since.

chrisduerr:
Running `cargo fmt` in the latest master should do absolutely nothing (and it doesn't do anything for me). So feel free to give that a shot to confirm if everything's setup correctly.

Note that I've just pushed a new commit to fix all current formatting issues. The latest commit didn't have rustfmt run on it apparently, but now everything should work great. Maybe that's why you saw things changing that you didn't touch!

alacritty_terminal/Cargo.toml
```diff
@@ -61,6 +61,8 @@ default = []
 live-shader-reload = []
 nightly = []
 bench = []
+# Use FreeType+FontConfig on all platforms, and enable HarfBuzz text shaping
+hb-ft = ["font/hb-ft"]
```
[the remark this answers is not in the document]
thunderseethe: The runtime option in it's current manifestation is harfbuzz and linux specific. This code is pretty narrow in scope in that it only works for harfbuzz and atm harfbuzz is only tested on linux. The scope can expand to cover other OSs but I would be wary to move this outside a feature flag while support on other OSes isn't in place.

alacritty_terminal/src/cursor.rs
```diff
@@ -56,13 +55,19 @@ pub fn get_cursor_glyph(
     }
 }
 
+// This default is done as a constant to avoid duplicating the feature toggle for each cursor type.
+#[cfg(not(feature = "hb-ft"))]
+const CURSOR: char = ' ';
+#[cfg(feature = "hb-ft")]
+const CURSOR: font::key_type::KeyType = font::key_type::KeyType::GlyphIndex(1u32);
```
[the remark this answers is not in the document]
thunderseethe: That is correct

alacritty_terminal/Cargo.toml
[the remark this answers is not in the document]
kchibisov: You can make it specific to `linux/bsd` rather than hided by "feature", which is perfectly fine as long as the requirements specified by @chrisduerr are met. Our font rendering will be feature less on macOs/Windows, but we shouldn't care about it for a now (e.g. we already has colored emojis on macOs, but not on linux). In fact we can use other libraries to make ligatures work on macOs/Windows.

> The runtime option in it's current manifestation is harfbuzz and linux specific.

FYI, we already has platform specific options in config (e.g.  `use_thin_strokes` for macOs).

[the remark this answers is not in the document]
chrisduerr: Having a feature flag on master is pretty much the same as not having the feature at all. In that case we could just keep the PR open until the reasons for having it as a feature flag are resolved.

However, for implementing something platform specific we can do precisely what @kchibisov said and use linux/bsd specific cfg tags.

However, even in that case we do want to make sure to reduce code duplication instead of just slapping feature/cfg attributes everywhere. The goal should be to use the same code for as much stuff is possible.

thunderseethe:
What's the policy on adding new dependencies? I'd like to rely on [itertools::Either](https://docs.rs/itertools/0.8.0/itertools/enum.Either.html). Can I add that library as a dependency for such a relatively narrow usage? Or would it be preferred to replicate the bit of code I need and avoid a new dependency?

chrisduerr:
I'm not a fan of it to be honest. Especially adding a completely new dependency just for that. In most cases it's a better idea to create a proper enum which is nicely named.

alacritty_terminal/src/renderer/mod.rs
```diff
@@ -1128,7 +1128,7 @@ where
 {
     // If text_run is hidden we don't want to shape text.
     // But we still want to run each cell through add_render_item so colors get handled
-    // appropiately. We construct a dummy key that should be cached a majority of the time so
+    // appropiately. We construct a dummy key that should bee cached a majority of the time so
```
asymmetric: Typo?

TheZoq2:
Nice to see this being worked on! I couldn't wait for ligatures, so I started using this branch a few days ago.

Pretty much everything works fine, except for rendering of a the glyph for `++` in the shell. It works fine in vim, but in pretty much everything else (zsh, bash, cat, echo...) only the second + character is rendered.

I'm using the hasklig font installed from the AUR. My alacritty config can be found in my [dotfiles repo](https://github.com/thezoq2/dotfiles)

TheZoq2:
Never mind, looks like the issue is something different. I changed the `font family` from `Hasklig` to `Hasklig Semibold`, which seems to have broken ligatures.

TheZoq2:
> @TheZoq2 I pulled a copy of your alacritty.yml to test but was unable to reproduce either of the issues.

I made some changes to it since then, so things may have changed. My issues were with `family: Hasklig Semibold` and I believe `style:regular`

> Do you have a screen shot to show what you're seeing? Also I believe if you want to enable Hasklig Semibold it's font: Hasklig style: Semibold as opposed to font: "Hasklig Semibold" style: Regular

Yes, that's the conclusion I came to as well, however semi bold looks very non-bold for me so I use   `style: Medium` instead

Arvedui:
I have been using this branch for some time mostly without problems, however there seems to be an issue with at least Japanese and Chinese characters.
You can see it in the first image already, in the second the line is just highlighted to make it more pronounced.

thunderseethe:font-ligatures:
sel-2019-09-04-10:19:32
sel-2019-09-04-10:19:46

master for comparison:
sel-2019-09-04-10:26:16
sel-2019-09-04-10:27:03


The first 5 are Japanese Hiragana the last two some random Chinese copied from a Wikipedia page.

alacritty_terminal/src/renderer/rects.rs
```diff
@@ -90,33 +91,19 @@ impl RenderLines {
             .collect()
     }
 
-    /// Update the stored lines with the next cell info.
-    pub fn update(&mut self, cell: &RenderableCell) {
+    /// Update the stored lines with the next text_run info.
+    pub fn update(&mut self, text_run: &TextRun) {
```
chrisduerr: Currently it seems like this method doesn't have a ton of logic in it anymore.

I don't think we need a dedicated struct for this anymore, we already have a collection of text runs and those are already separated by flags, so we've got all the info we need.

I think we could just make a method on render rect which takes a text run as parameter and then we'll just create all rects by iterating over all text runs, filtering by flags and then creating render rects for it.

Does that make sense to you? I just think the middle man isn't super necessary anymore, we already have the data structure ready as text runs, so all we need is a single method to convert them.

alacritty_terminal/src/term/text_run.rs
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /// End point covered by this TextRun
+    pub fn end_point(&self) -> Point {
+        Point { line: self.line, col: self.span.1 }
+    }
+
+    /// Returns iterator over range of columns [run.0, run.1]
+    pub fn cols(&self) -> impl Iterator<Item = Column> {
+        let (start, end) = self.span;
```
chrisduerr: We can take destructuring even further here!

```rust
let (Column(start), Column(end)) = self.span;
```

This way there's no need to do `.0` anymore.

I'd also move this right above the call itself, just since this is a very temporary variable definition just for ergonomics.

alacritty_terminal/src/term/text_run.rs
```diff
@@ -0,0 +1,309 @@
...
+        let step = if self.flags.contains(Flags::WIDE_CHAR) {
+            // If our run contains ide chars treat each cell like it's 2 cells wide
```
chrisduerr: I don't think we need this comment. Returning `2` instead of `1` when something's a wide cell should be pretty obvious.

alacritty_terminal/src/term/text_run.rs
```diff
@@ -0,0 +1,309 @@
...
+        Point { line: self.line, col: self.span.0 }
+    }
+
+    /// End point covered by this TextRun
+    pub fn end_point(&self) -> Point {
+        Point { line: self.line, col: self.span.1 }
+    }
+
+    /// Returns iterator over range of columns [run.0, run.1]
+    pub fn cols(&self) -> impl Iterator<Item = Column> {
```
chrisduerr: Doesn't need to be public. However, I don't think we need this method anyways.

If we fuse it together with `fn cells`, the resulting method would actually be pretty clean:

```rust
    pub fn cells(&self) -> impl Iterator<Item = RenderableCell> {
        let step = if self.flags.contains(Flags::WIDE_CHAR) {
            2
        } else {
            1
        };
        let (Column(start), Column(end)) = self.span;
        (start..=end).step_by(step).map(move |x| self.cell_at(Column(x)))
    }
```

alacritty_terminal/src/term/text_run.rs
```diff
@@ -0,0 +1,309 @@
...
+    pub fn start_point(&self) -> Point {
+        Point { line: self.line, col: self.span.0 }
+    }
+
+    /// End point covered by this TextRun
+    pub fn end_point(&self) -> Point {
+        Point { line: self.line, col: self.span.1 }
+    }
+
+    /// Returns iterator over range of columns [run.0, run.1]
```
chrisduerr: I think `range of columns [run.0, run.1]` is a bit unnecessarily complicated. Cleartext description would be better I think. If I understand things correctly this does the following:

```
/// Returns iterator over all columns in text range
```

I think that's more clear? Think that makes it a bit easier to understand than `run.0` and `run.1` which could mean anything (is there a run.2? who knows!).

alacritty_terminal/src/term/text_run.rs
```diff
@@ -0,0 +1,309 @@
...
+    pub fn cols(&self) -> impl Iterator<Item = Column> {
+        let (start, end) = self.span;
+        let step = if self.flags.contains(Flags::WIDE_CHAR) {
+            // If our run contains ide chars treat each cell like it's 2 cells wide
+            2
+        } else {
+            1
+        };
+        // unpacking is neccessary while Step trait is nightly
+        // hopefully this compiles away.
```
chrisduerr: If we need to do it this way, we need to do it this way. So it doesn't really matter if it compiles away or not.

Being able to fix this once the `Step` trait is stable is a very important comment though, it should be highlighted to be easier to search for with a `TODO`.

```
// TODO: Use `Step` trait once stabilized
```

Would we want to implement the step trait for Columns here? Would this even be possible with both full and half width columns? Or does the trait not actually help here?

alacritty_terminal/src/term/text_run.rs
```diff
@@ -0,0 +1,309 @@
...
+                self.buffer_zero_width.push(arr);
+            },
+            RenderableCellContent::Cursor(cursor) => {
+                self.cursor = Some(cursor);
+            },
+        }
+    }
+
+    /// Empty out pending buffer producing owned collections that can be moved into a TextRun
+    fn drain_buffer(&mut self) -> (String, Vec<[char; MAX_ZEROWIDTH_CHARS]>) {
```
chrisduerr: This should be `bufferS`, it doesn't drain only one buffer. Documentation should state accordingly.

alacritty_terminal/src/term/text_run.rs
```diff
@@ -0,0 +1,309 @@
...
+                // Save leftover buffer and empty it
+                Some(TextRun::from_iter_state(start, latest, self.drain_buffer()))
+            } else if let Some(cursor) = self.cursor {
+                let start = self.run_start.take()?;
+                Some(TextRun::from_cursor_rc(start, cursor))
+            } else {
+                None
+            }
+        })
+    }
```
chrisduerr: Okay, so the `TextRunIterator` functions are a bit messy I think. I don't have a clear instruction here on how to improve things since that would require a lot of experimenting on my side too, but to try to get you some information about what needs to be changed I'll list some issues I see with the current approach:

```
buffer_content -> Why do we set cursor here to notify next col to start a new run?
produce_cursor -> Do we really need to create a new run with relatively generic logic? Wouldn't it be easier to just handle this separately as soon as a cursor is encountered?
produce_char_run -> Why `prev_buffer` isn't it the current buffer? We first collect all text, then start a new run, then construct a new run from the last run? That's a bit confusing.
start_run -> Does this return the info of the last run? A bit unclear and not documented in name or comments, there's probably a better solution.
TextRunIterator::new
    -> It seems like it would be much easier to early return rather than buffering `output`
    -> output.or_else is used to return the last text run? I think this is a bit confusing and caused by the start_run being unclear too.
    -> The whole cursor handling in output.or_else could probably be handled by explicitly handling cursors separately
```

Of course these are very general statements without clear instructions, so please let me know when you have any more questions. But I think it should be pretty clear here that control flow is not straight forward and I think we can do better.

alacritty_terminal/src/term/text_run.rs
```diff
@@ -0,0 +1,309 @@
+use super::{
```
chrisduerr: I'm not sure that `term/text_run.rs` is the best place for the text run. Sure, it's related to the terminal in some way, but it's also related to font and rendering.

I think the most uncontroversial thing would be to just put it in `src/`. That way we can keep terminal clean with just the terminal logic and the cells and things clearly related to it (please ignore src/term/color.rs, that's in a completely bad spot too).

anirudhb:
Wow @thunderseethe thank you so much for picking back up on this! I really want to use Alacritty, and this is my only blocker! I haven't had time lately :(

Arvedui:
Selection looks better now, but …

sel-2019-09-10-10:29:28

DanilaFe:
I don't know Alacritty's internals (or rust), but can someone explain whether using Harfbuzz can only be implemented on Linux, or if eventually macOS and Windows will receive this feature, too? 
If the latter is true, what's the hurdle to supporting ligatures on other platforms? 

Thanks for working on this!

chrisduerr:
> I don't know Alacritty's internals (or rust), but can someone explain whether using Harfbuzz can only be implemented on Linux, or if eventually macOS and Windows will receive this feature, too?
If the latter is true, what's the hurdle to supporting ligatures on other platforms?

Harfbuzz only works natively on Linux. So other platforms would have strange rendering that does not conform to their system's font configuration.

thunderseethe:
@Arvedui latest commits should resolve that issue, can you give it another look and confirm it's not just on my machine?

Cargo.lock
```diff
@@ -664,6 +665,15 @@ name = "foreign-types-shared"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
+[[package]]
+name = "freetype"
```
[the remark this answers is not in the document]
chrisduerr: It's a dependency of harfbuzz.

font/src/lib.rs
```diff
@@ -135,39 +140,48 @@ impl FontKey {
...
+/// Captures possible outcomes of shaping, if shaping succeeded it will return a `GlyphIndex`.
+/// If shaping failed or did not occur, `Fallback` will be returned.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub enum KeyType {
+    /// Shaping returned a valid index and we can render that as expected.
+    GlyphIndex(u32),
+    /// Shaping returned a missing glyph or shaping did not occur. If glyph is missing system will
+    /// attempt to load character glyph from a fallback font. If shaping did not occur this will
+    /// first try the configured font then fallback.
+    Fallback(char),
```
chrisduerr: I'd like a clarification here since I'm not entirely certain how things work and it seems like the comments/naming are not sufficiently explaining this at the moment.

If I understand things correctly, on non-harfbuzz systems we do never use the GlyphIndex and on harfbuzz supported systems we only use `Fallback` for font fallback, right?

That's what I remembered and the comments seem to support this, however I don't think the `Fallback` name sufficiently explains this. One might think that this is only used for font fallback, but that's not accurate. So I'd like to steer away from it.

Maybe just having this as `KeyType::Char` would be better? It's simple and accurately describes the type, without going into too much detail what it might be used for. I think that might be a bit better? But alternatives are certainly possible. Maybe `RawChar` if we want to steer away from using just `Char`, but since it's capitalized I don't think it should create an issue.

Very curious what you'd think about this, it just seems that `Fallback` currently isn't ideal.

font/src/lib.rs
```diff
@@ -135,39 +140,48 @@ impl FontKey {
...
-            // This transmute is fine:
-            //
-            // - If GlyphKey ever becomes a different size, this will fail to compile
-            // - Result is being used for hashing and has no fields (it's a u64)
-            ::std::mem::transmute::<GlyphKey, u64>(*self)
-        }
-        .hash(state);
+impl Default for KeyType {
+    // Arbitrary non missing (non 0) glyph
+    // Used as a placeholder when glyph doesn't matter, such as when representing a cursor.
```
chrisduerr: I think with having this `PLACEHOLDER_GLYPH` as constant, the `non 0` comment is more confusing than helpful. It also doesn't seem arbitrary now, since it's a well named constant. So I'd just remove this comment.

Since this is the font library, we also shouldn't conflate it with topics like 'cursors', that's a terminal and not a font thing. As a library it's up to the user what they want to use this for.

So in general I think the useful part of this comment is that it hints at the default glyph key being completely empty, but I don't think that's super crucial information and pretty much the only thing that makes sense.

So I'd remove both of these comments myself, though if you think they bring some good value I'd recommend shortening them down to just a small statement saying that the default glyph key is for an empty glyph.

font/src/lib.rs
```diff
@@ -267,3 +278,11 @@ pub trait Rasterize {
     /// Update the Rasterizer's DPI factor
     fn update_dpr(&mut self, device_pixel_ratio: f32);
 }
+
+// Only implemented for the FreeType rasterizer so far.
```
chrisduerr: I think leaving a comment here will make it easy to just forget about it even after implementation. Missing features are better documented in github issues, where we don't have to go crawl through code to find remaining TODOs (especially since this isn't even tagged as TODO).

So I'd just remove this and we can then create a new issue about ligature support on Windows and macOS.

font/src/lib.rs
```diff
@@ -267,3 +278,11 @@ pub trait Rasterize {
...
+/// Conceptually this extends the Rasterizer trait with Harfbuzz specific functionality.
```
chrisduerr: I think we can make this comment a little bit more concise.

```suggestion
/// Extend the Rasterizer trait with Harfbuzz specific functionality.
```

Though generally, this is currently not an extension to the rasterizer trait, so we either need to reword it or change the `HbFtExt` trait to require the `Rasterizer` trait to be implemented. I don't care either way, since the `Rasterizer` trait will always be implemented anyways and the `HbFtExt` should technically be independent.

But we should still make things clear. If you should decide to remove the extension part, the trait also needs to be renamed though.

font/src/lib.rs
thunderseethe: I think that's a good point, originally it was intended to specify that the variant immediately fell back to system font and didn't check the user font at all. This was also made while everything was still under the hb-ft flag so it was assumed things were harfbuzz. However based on what it looks like font ligatures will look like under windows (and probably mac) I agree it no longer makes sense.

TL;DR I'll change it to Char and linux font fetching code can be commented to explain the fallback behavior.

font/src/lib.rs
thunderseethe: I've reworded the comment to hopefully better capture what it does. 

At a larger scope, I actually think this trait might go away entirely as font ligature support is added to windows and mac osx. Both of their APIs appear (from a quick glance) to rasterize batches of glyphs at a time and expose shaping through that. Freetype is the odd man out in that regard, so I'd imagine the Rasterizer trait would change to rasterize batches of glyphs instead of one glyph at a time and then Harfbuzz would entirely be an implementation detail of the freetype rasterizer.

However until taking a further look at the directwrite and core text apis I don't want to take a stab at modifying rasterizer so for now this is an okay approach.

font/src/lib.rs
thunderseethe: :sweat_smile: It appears I've commented on the wrong reply. My reply to the comment about `// Only implemented for the FreeType rasterizer so far.` also applies to this comment. I'm gonna resolve this comment in the interest of consolidating discussion.

alacritty_terminal/src/renderer/rects.rs
thunderseethe: I missed this one entirely, sorry about that. I wholeheartedly agree, I've removed the RenderLines struct and replaced it with a `from_text_run` method on RenderRect. The lines local in draw is replaced with a Vec of RenderRects.

font/src/lib.rs
chrisduerr: My problem with this comment here is that it doesn't serve much of a purpose and it requires maintenance in the future when the trait is implemented for another platform. Having ta change this to say it's implemented for FreeType and CoreText would be a waste for example. And since this is harfbuzz specific at the moment anyways, there's not really a good way to reuse it on other platforms as far as I'm aware.

So in general this specific comment just doesn't help much imho. If I want to see where the trait is used, I can just look at the references in the code.

font/Cargo.toml
```diff
@@ -22,4 +23,4 @@ core-graphics = "0.17"
 core-foundation-sys = "0.6"
 
 [target.'cfg(windows)'.dependencies]
-dwrote = { version = "0.9.0" }
+dwrote = { version = "0.9.0" }
```
chrisduerr: File is missing newline at the end.

alacritty_terminal/src/display.rs
```diff
@@ -514,20 +517,17 @@ impl Display {
             let mut lines = RenderLines::new();
 
             // Draw grid
-            {
-                let _sampler = self.meter.sampler();
```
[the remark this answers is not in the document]
chrisduerr: The render timer is still not actually drawn.

font/src/directwrite/mod.rs
```diff
@@ -122,15 +124,20 @@ impl crate::Rasterize for DirectWriteRasterizer {
...
+                .get_glyph_indices(&[' ' as u32])
+                .first()
+                .filter(|index| **index != 0)
+                .ok_or_else(|| Error::MissingGlyph(' '))?,
+            KeyType::Char(c) => *font
+                .get_glyph_indices(&[c as u32])
+                .first()
+                .filter(|index| **index != 0)
+                .ok_or_else(|| Error::MissingGlyph(c))?,
+        };
```
chrisduerr: Just had an idea. I think we can do things a lot better here if we never actually receive GlyphIndex.

In this match arm we can just return either `c` or `' '` and mark the `GlyphIndex` branch an usreachable. Then we do all the other stuff once instead of twice, since the code is identical.

This allows us to get rid of `keytype_unwrap_char` too!

alacritty_terminal/src/term/text_run.rs
```diff
@@ -0,0 +1,309 @@
...
+/// Represents a set of renderable cells that all share the same rendering propreties.
+/// The assumption is that if two cells are in the same TextRun they can be sent off together to
+/// be shaped. This allows for ligatures to be rendered but not when something breaks up a ligature
+/// (e.g. selection hightlight) which is desired behavior.
+#[derive(Debug)]
+pub struct TextRun {
+    /// By definition a run is on one line.
+    pub line: Line,
+    /// Span of columns the text run covers.
+    pub span: (Column, Column),
```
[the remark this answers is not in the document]
chrisduerr: I don't see why this wouldn't be possible?

```diff
diff --git a/alacritty_terminal/src/renderer/mod.rs b/alacritty_terminal/src/renderer/mod.rs
index fff5187..c3e5243 100644
--- a/alacritty_terminal/src/renderer/mod.rs
+++ b/alacritty_terminal/src/renderer/mod.rs
@@ -1042,7 +1042,7 @@ impl<'a> RenderApi<'a> {
 
         let text_run = TextRun {
             line,
-            span: (Column(0), Column(string.len() - 1)),
+            span: Column(0)..=Column(string.len() - 1),
             content: TextRunContent::CharRun(string.to_owned(), vec![]),
             fg: Rgb { r: 0, g: 0, b: 0 },
             bg: color.unwrap_or(Rgb { r: 0, g: 0, b: 0 }),
diff --git a/alacritty_terminal/src/text_run.rs b/alacritty_terminal/src/text_run.rs
index 4099647..c38dd6c 100644
--- a/alacritty_terminal/src/text_run.rs
+++ b/alacritty_terminal/src/text_run.rs
@@ -41,7 +41,7 @@ pub struct TextRun {
     /// A run never spans multiple lines.
     pub line: Line,
     /// Span of columns the text run covers.
-    pub span: (Column, Column),
+    pub span: std::ops::RangeInclusive<Column>,
     /// Cursor or sequence of characters.
     pub content: TextRunContent,
     /// Foreground color of text run content.
@@ -63,7 +63,7 @@ impl TextRun {
         let end_column = if is_wide { latest + 1 } else { latest };
         TextRun {
             line: start.line,
-            span: (start.column, end_column),
+            span: start.column..=end_column,
             content: TextRunContent::CharRun(buffer.0, buffer.1),
             fg: start.fg,
             bg: start.bg,
@@ -75,7 +75,7 @@ impl TextRun {
     fn from_cursor_rc(start: RunStart, cursor: CursorKey) -> Self {
         TextRun {
             line: start.line,
-            span: (start.column, start.column),
+            span: start.column..=start.column,
             content: TextRunContent::Cursor(cursor),
             fg: start.fg,
             bg: start.bg,
@@ -99,24 +99,25 @@ impl TextRun {
 
     /// First cell in the TextRun
     pub fn start_cell(&self) -> RenderableCell {
-        self.cell_at(self.span.0)
+        self.cell_at(*self.span.start())
     }
 
     /// First point covered by this TextRun
     pub fn start_point(&self) -> Point {
-        Point { line: self.line, col: self.span.0 }
+        Point { line: self.line, col: *self.span.start() }
     }
 
     /// End point covered by this TextRun
     pub fn end_point(&self) -> Point {
-        Point { line: self.line, col: self.span.1 }
+        Point { line: self.line, col: *self.span.end() }
     }
 
     /// Iterates over each RenderableCell in column range [run.0, run.1]
     pub fn cells<'a>(&'a self) -> impl Iterator<Item = RenderableCell> + 'a {
         let step = if self.flags.contains(Flags::WIDE_CHAR) { 2 } else { 1 };
-        let (Column(start), Column(end)) = self.span;
+
         // TODO: impl Step for Column (once Step is stable) to avoid unwrapping then rewrapping.
+        let (&Column(start), &Column(end)) = (self.span.start(), self.span.end());
         (start..=end).step_by(step).map(move |col| self.cell_at(Column(col)))
     }
 }
```

alacritty_terminal/src/term/text_run.rs
chrisduerr: Sure this is an implementation detail inside of text_run, but this method is private and only called inside of text_run. It's not hidden behind some kind of API that people aren't looking into. This is the place where things are implemented and it's returning two completely separate buffers not one. So I don't see how they are one.

alacritty_terminal/src/term/text_run.rs
chrisduerr: > At a high level I believe a lot of the issue here spawn from the fact that we have to destructively grab enough RenderableCells from our internal iterator.

Maybe we could use `peek`?

I can take a deeper look at it and rewrite things myself, though that would take time I currently do not have (at least for this PR). But I'm pretty confident we can improve things a lot here.

alacritty_terminal/src/renderer/mod.rs
```diff
@@ -1002,74 +1067,138 @@ impl<'a> RenderApi<'a> {
...
+where
+    L: LoadGlyph,
+{
+    // If text_run is hidden we don't want to shape text.
+    // But we still want to run each cell through add_render_item so colors get handled
+    // appropiately. We construct a dummy key that should be cached a majority of the time so
+    // that we have a valid glyph.
+    let key = GlyphKey { id: PLACEHOLDER_GLYPH, font_key, size: glyph_cache.font_size };
+    *glyph_cache.get(key, loader)
+}
```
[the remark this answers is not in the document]
thunderseethe: Does a Default glyph behave correctly for any of the font key/size combos? My concern with this is that since it doesn't account for font size it's spacing may not relative too large or small font sizes. However it's also possible all the spacing stuff is handled by the RenderableCell that gets passed and the glyph layout members aren't used at all for that purpose.

alacritty_terminal/src/text_run.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// Background color of text run content.
+    pub bg: Rgb,
+    /// Background color opacity of the text run
+    pub bg_alpha: f32,
+    /// Attributes of this text run.
+    pub flags: Flags,
+}
+
+impl TextRun {
+    fn from_iter_state(
```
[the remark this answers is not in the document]
thunderseethe: I would expect a new method to be pub and to have a signature akin to `(Line, (Column, Column), TextRunContent, Rgb, Rgb, f32, Flags) -> Self` or some subset. Some parameter set that clearly maps to the fields of text run perhaps with some reasonable defaults.

This method by contrast should not be used publicly and only makes sense in the context of the `TextRunIter`. I'm happy to change the name but I don't think it makes sense as a new method. Whatever the name is it should capture the semantics that this method constructs a text run inside the `TextRunIter` specifically from state the `TextRunIter` maintains, and shouldn't be public.

alacritty_terminal/src/text_run.rs
```diff
@@ -0,0 +1,289 @@
...
+    iter: I,
+    run_start: Option<RunStart>,
+    latest_col: Option<LatestCol>,
+    cursor: Option<CursorKey>,
+    buffer_text: String,
+    buffer_zero_width: Vec<[char; MAX_ZEROWIDTH_CHARS]>,
+}
+
+// This is an explicit function (as opposed to a closure) to make it easy to use as a function
+// pointer.
```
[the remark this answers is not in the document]
thunderseethe: Oh nice I didn't realize rust would coerce a closure to a function pointer now, previously you had to manually coerce it with `(|cell| ....) as fn(&RenderableCell) -> bool` which is verbose and non-obvious compared to using a normal function pointer.

Either way, I've swapped it back to a normal closure now.

alacritty_terminal/src/text_run.rs
```diff
@@ -0,0 +1,289 @@
...
+        is_cell_not_adjacent || is_col_not_adjacent
+    }
+
+    /// Add content of cell to pending TextRun buffer
+    fn buffer_content(&mut self, inner: RenderableCellContent) {
+        // Add to buffer only if the next rc is a Char (not a cursor)
+        match inner {
+            RenderableCellContent::Chars(chars) => {
+                self.buffer_text.push(chars[0]);
+                let mut arr: [char; MAX_ZEROWIDTH_CHARS] = Default::default();
```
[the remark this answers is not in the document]
thunderseethe: What would be the advantage of using that when `[char; N]` impls Default? It would hard code behavior here that we don't need to hard code.

alacritty_terminal/src/renderer/mod.rs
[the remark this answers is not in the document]
chrisduerr: Based on my testing, everything should work perfectly. The only issue I saw was that the default texture ID might not be defined, but since we always start with texture id 0 and number types default to 0, we should always have the texture id 0 valid.

Feel free to test this yourself though.

alacritty_terminal/src/text_run.rs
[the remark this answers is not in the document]
chrisduerr: I'm fine with not naming it `new`, I'm just not a huge fan of the name. It shouldn't really matter where the method is called from, but instead what it does. And I see it as a bit troublesome since there's no iterator anywhere to be seen in the method signature.

alacritty_terminal/src/text_run.rs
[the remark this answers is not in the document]
chrisduerr: It's shorter and more clear. It's a bit pointless to use `Default::default()` when you then have to specify the type anyways.

It's hardcoded either way, the code is completely identical. If the type changes, you need to change the type annotations. If you change it to mine, you have to change the assigned value. Doesn't really give you any advantages.

sQVe:
I just tried this out with [**Iosevka**](https://github.com/be5invis/iosevka), which does not work with `kitty` due to the dynamic nature of the ligatures, and **it works well**. Thank you for all your work @thunderseethe. :tada: :pray:

There is some minor jaggedness to ligatures like `===`. Given the echo of `echo "=, ==, ===, !=, !==, ------>, >="` it will output:

font/src/ft/mod.rs
```diff
@@ -265,13 +289,18 @@ impl FreeTypeRasterizer {
                 Some(FixedSize { pixelsize: pixelsize.next().expect("has 1+ pixelsize") })
             };
 
+            // Construct harfbuzz font
+            let hb_font = Font::new(HbFace::from_file(&path, index as u32)?);
```
Lompik: You could create from ```ft_face``` like so:

```rust
    use harfbuzz_rs::HarfbuzzObject;
    use harfbuzz_rs::hb as harfbuzz_sys;
    let hb_font_raw = unsafe{harfbuzz_sys::hb_ft_font_create_referenced(ft_face)};
    let hb_font = unsafe {harfbuzz_rs::Font::from_raw(hb_font_raw)};
```
(same amount of ```unsafe``` as before).

chrisduerr:
> There is some minor jaggedness

This is actually intentional to make it easier to see the difference between =, == and ===. Which I do think makes some sense if you think about the reason behind it.

kchibisov:
Underlining and strikeout are incorrect on this PR. You can try `printf "\e[4;9mhello\n"` to understand what I'm talking about.

Screenshot_Wed 25 Sep 2019 05:38:37 PM MSK

thunderseethe:
@kchibisov yea that's a good old off by one error, I have a fix for it but  it looks like I haven't pushed it yet.

I'll push the the fix to this PR next I'm at my computer.

alacritty_terminal/src/text_run.rs
```diff
@@ -0,0 +1,289 @@
...
+    line: Line,
+    column: Column,
+    fg: Rgb,
+    bg: Rgb,
+    bg_alpha: f32,
+    flags: Flags,
+}
+impl RunStart {
+    /// Compare cell and check if it belongs to the same run.
+    fn is_adjacent_to_cell(&self, rc: &RenderableCell) -> bool {
```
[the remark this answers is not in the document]
kchibisov: and probably adding `#[inline]` here will be a good idea.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Completions for `--class` and `-t` (short title)
 - Change the mouse cursor when hovering over the message bar and its close button
 - Support combined bold and italic text (with `font.bold_italic` to customize it)
+- Font Ligatures on Linux
```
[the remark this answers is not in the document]
thunderseethe: I grepped through the code base this instance and the comment in renderer were the only instances and they've both been updated

alacritty_terminal/src/renderer/mod.rs
```diff
@@ -962,32 +1039,20 @@ impl<'a> RenderApi<'a> {
...
+            bg_alpha,
+        };
 
-        for cell in cells {
-            self.render_cell(cell, glyph_cache);
-        }
+        self.render_text_run(text_run, glyph_cache);
     }
 
-    #[inline]
```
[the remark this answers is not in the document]
thunderseethe: Accident

font/src/ft/mod.rs
thunderseethe: I agree with the sentiment, it's wasteful to create the HbFont when we already have a Ft Face ripe for the picking.

There's a frustrating issue going on here with two different freetype crates each exporting their own version of FT_Face. So if you try to convert ft_face to the expected *mut FT_FaceRec_ from freetype-0.4.1 it's actually a *mut FT_FaceRec from freetype-rs (transitively freetype-sys).  I took a look at manually converting but you end up having to recursively convert the pointers FT_Face holds and it got too gnarly.

In the long term we should upstream a variant of the harfbuzz_sys bindings that use freetype-rs types. Until then I think we're stuck with the current approach. Unless you know of a workaround I'm unaware of, that would be sweet.

alacritty_terminal/src/text_run.rs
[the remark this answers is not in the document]
thunderseethe: I've swapped it out to construct the Array explicitly but use char's default. Hopefully this is a reasonable compromise.

alacritty_terminal/src/text_run.rs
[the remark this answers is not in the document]
thunderseethe: I've moved it to TextRunIter and renamed it to `build_text_run`. The name should now match the behavior of the function and it's new location should clarify the scope of usage.

alacritty_terminal/src/text_run.rs
```diff
@@ -0,0 +1,289 @@
...
+}
+impl<BaseIter> TextRunIter<std::iter::Filter<BaseIter, fn(&RenderableCell) -> bool>>
+where
+    BaseIter: Iterator<Item = RenderableCell>,
+{
+    pub fn new(iter: BaseIter) -> Self {
+        TextRunIter {
+            // Logic for WIDE_CHAR is handled internally by TextRun
+            // So we no longer need WIDE_CHAR_SPACER at this point.
+            iter: iter.filter(is_not_wide_char_spacer),
```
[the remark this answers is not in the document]
thunderseethe: I think we can. The rationale for putting it here is because TextRunIter is the one that cares about filtering out the WIDE_CHAR_SPACER characters. While there currently aren't any, in theory any other consumer of the renderable_cells() method would want the full stream of RenderableCells unfiltered.

That being said RenderableCellsIter and TextRunIter are very tightly coupled and I think they could more aggressively be combined. For example, we could build TextRuns directly from `Cell` and avoid going through RenderableCellsIter entirely. I haven't taken a look at grid code at all though so I have don't know how involved that would be.

font/src/ft/mod.rs
Lompik: Also, I did not realize ```harfbuzz_rs``` types have a ```PhantomData``` marker so the code above wouldn't have cut it because ```Font::hb_font``` is marked ```'static```. It turns out though, current ```harfbuzz_rs``` leaks `hbface` when calling `hb_rs::Font::new`. So if we didn't want to leak  data, we'd need to the reduce reference count of the inner `hbbuff` and `hbface` of `hb_rs::Font` by implementing `Drop` on `Font`. Or avoid create `hbface`: the code below worked for me and does not leak from HB. But this quite a messy situation.

```rust
use harfbuzz_rs::HarfbuzzObject;
use harfbuzz_rs::hb as harfbuzz_sys;
let hb_font_raw = unsafe{harfbuzz_sys::hb_ft_font_create_referenced(ft_face.raw_mut() as *mut _ as *mut u64 as *mut _)};
let hb_font = unsafe {harfbuzz_rs::Font::from_raw(hb_font_raw)};

let face = Face {
    hb_font: unsafe{std::mem::transmute(hb_font)},
```

alacritty_terminal/src/text_run.rs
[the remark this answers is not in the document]
chrisduerr: As far as I can tell, we're currently using the TextRun on all platforms anyways and it's a fixed part of our pipeline. So I don't see why we wouldn't integrate it properly.

We only need the renderable cells to render text. If we always render text through the text run iter, we should be fine with reworking the pipeline to fit us as well as possible.

font/src/ft/mod.rs
thunderseethe: Alright silly question, how did you get harfbuzz_sys to build with ft linked? So far I've been unsuccessful in doing so and haven't been able to figure out if it's building from source and just not linking it in or if it's my system copy and I need to recompile that.

Also, thank you for pointing out the memory leak and providing a workaround for the library incompatibility! While the pointer conversion is quite messy, it's preferable to introducing a memory leak by far.

anirudhb: Ditto - That's why I did it that way in the original code

Lompik: > Alright silly question, how did you get harfbuzz_sys to build with ft linked? 

with hb-sys 0.3.2: try something like ```HARFBUZZ_SYS_NO_PKG_CONFIG=1 cargo build```. It seems possible to build on windows although I have not tested that. Also ```alpine```, ```arch``` & ```fedora 29``` seems to build hb with freetype already. If your native harbuzz comes with ```hb-ft.h``` (usually in ```/usr/include/harfbuzz```)  then it has been built with freetype support.

chrisduerr:
I've just merged the winit/glutin rework PR which changed a lot of code, so that caused a few conflicts. If you need any help resolving these, please let me know.

sQVe:
What's missing from this getting merged? I've used it for quite some time now and it works really well.

thunderseethe:
AFAIK there are 4 things, barring anything that arises in follow up reviews:

- The current harfbuzz font construction has a memory leak, that's gotta be resolved
- The windows change needs to be made see @chrisduerr's comment for more context
- TextRun needs to be integrated into Grid so we don't expose RenderableCell, and possibly get rid of RenderableCell entirely.
- Benchmarking needs to be done to get a sense of how heavy harfbuzz is, and almost certainly with the results of those benchmarks we'll need something to help alleviate the problem. Most major browsers handle it with a word cache but that creates issues around word wrapping and determining what a word is for our use case.

The PR isn't abandoned yet buy my october has been full of IRL stuff so I haven't had a chance to clean up these residual things yet. I was able to merge the large PR successfully, and despite the failing CI builds have tested locally that things are stil WAI with regard to ligatures and runs (can't speak to the winit things but it still had a window)

lf94:
Keep it up guys, we're all watching :) Excellent work by everyone. Can't wait.

XVilka:
There is interesting article from JaneStreet about [adding underlines or digits numbering](https://blog.janestreet.com/commas-in-big-numbers-everywhere/) using font shaping, which might be a good test case for this work too. The font sources [available at GitHub repository](https://github.com/trishume/numderline).  The online demo site is [here](https://thume.ca/numderline/).

image

image

ronasimi:
> Keep it up guys, we're all watching :) Excellent work by everyone. Can't wait.

So, based on the fact that 4.0 is in RC status is it safe to assume this won't make it in?

kchibisov:
>So, based on the fact that 4.0 is in RC status is it safe to assume this won't make it in?

Yes, it wont be in v0.4.0.

davidhewitt:
> The windows change needs to be made see @chrisduerr's comment for more context

I could potentially be interested in helping get the windows support done. Which comment are you referring to? I had a read of the thread but could not find anything which stood out to me...

thunderseethe:
> I've just merged the winit/glutin rework PR which changed a lot of code, so that caused a few conflicts. If you need any help resolving these, please let me know.

My comment was in reference to @chrisduerr landing a large window rework in master that I needed to merge in this PR. I believe I have that merging done. However since this PR has been sitting stagnant there is still plenty of merging to be done if you'd like to take a look.

The next merge on my stack is to look at the new urls.rs file and add a method to it to work with TextRuns.

davidhewitt:
Oh I see; you meant support for the updated windowing system, not the Windows OS! :smile:

Once this PR lands I would be interested in attempting support for Windows. Perhaps using the allsorts crate (which might also work for macOS).

thunderseethe:
I'm taking a look at the URL stuff wrt to TextRuns and wanted to ask, can a URL span a change in color/font/metadata (essentially span two or more text runs). At the very least it appears they can span multiple lines which means they can be at least two runs in that regard, but wanted to clarify the other semantics.

chrisduerr:
> I'm taking a look at the URL stuff wrt to TextRuns and wanted to ask, can a URL span a change in color/font/metadata (essentially span two or more text runs). At the very least it appears they can span multiple lines which means they can be at least two runs in that regard, but wanted to clarify the other semantics.

A URL is in the following format: https://github.com/jwilm/alacritty/blob/master/alacritty/src/url.rs#L18-L23

As you can see, the `lines` is a `Vec<RenderLine>`. Which means one URL can contain multiple lines and each line can have a different color. That's the main reason why there are multiple lines, because URLs are not interrupted when colors change.

Each of these `RenderLine`s might also span multiple lines in the terminal.

svalaskevicius:
had a quick go at merging master to this branch... tests pass, the terminal shows up and I can work in it, but have not tested much more

angvp:
I guess this branch needs rebase from master as I have lots of merge conflicts and in order to keep a clean history .. instead of merging master into this branch, good to know that way works tho.

roychoo:
Hi, looking forward to this, wouldn’t mind helping but I am a total rust newbie. Can I ask a noob question, why only linux?

kchibisov:
Due to use of hardbuzz library, which is used primary on linux, for macOS we should like use CoreText and for Windows DirectWrite.

Keating950:
Do people anticipate that this will be merged in the near future? If not, I'd be interested in packaging this branch for the [AUR](https://aur.archlinux.org).

chrisduerr:
> Do people anticipate that this will be merged in the near future? If not, I'd be interested in packaging this branch for the AUR.

This branch is **severely** outdated and requires some significant work just to get it back up to date.

angvp:
I am not a rust dev but I'm willing to volunteer to update the branch if the maintainer keeps MIA? ping @thunderseethe ?

rkanati:
@thunderseethe <polite coughing>

alacritty_terminal/src/renderer/mod.rs
```diff
@@ -286,7 +295,76 @@ impl GlyphCache {
...
 
-    pub fn get<'a, L>(&'a mut self, glyph_key: GlyphKey, loader: &mut L) -> &'a Glyph
+    //pub fn font_metrics(&self) -> font::Metrics {
+    //    self.rasterizer
+    //        .metrics(self.font_key, self.font_size)
+    //        .expect("metrics load since font is loaded at glyph cache creation")
+    //}
+
+    // Since shaping is not available on Windows/Mac OSX, text run glyphs are rasterized one by one
+    // as characters
```
blaggacao, line 305: https://github.com/yeslogic/allsorts ?

chrisduerr: We're open to it if the performance works well and it properly supports all necessary features on the platforms it could be used for.

However someone would actually have to do the work to try that out of course.

kchibisov: And please, avoid pinging folks without a reason for it. thx.

blaggacao: The reason is that Weslay, if I recall correctly, had offered working together with anyone who would want to pick this up.

kchibisov: yeah, and if anyone had a code with the mentioned lib and they had problems with it then maybe it might be something where pinging could make sense.

blaggacao:
@thunderseethe I made a patch between your branch and @zenixls2 's fork which is reported to work atop recent versions. Hope it helps: 
[zenixls2_patch.zip](https://github.com/alacritty/alacritty/files/4775485/zenixls2_patch.zip)

Since his version is heavily inspired by this PR, the relevant diffs should be easy to spot. (hopefully)

alacritty_terminal/src/renderer/mod.rs
blaggacao: Or when trying to summon the wizards before the raid, but that arguably doesn't always work :wink:

blaggacao:
>I'm taking a look at the URL stuff wrt to TextRuns and wanted to ask, can a URL span a change in color/font/metadata (essentially span two or more text runs). At the very least it appears they can span multiple lines which means they can be at least two runs in that regard, but wanted to clarify the other semantics.

You (@thunderseethe) might want to have a look at this changeset:

https://github.com/zenixls2/alacritty/commit/126093ab33664247298db91538058d67310695f3

alacritty_terminal/src/renderer/mod.rs
wezm: Hello all, I would be happy to answer any allsorts related queries. There should be no cross-platform issues — there are no OS dependencies except libc. We ship builds of Prince, which uses allsorts on at least Linux, macOS, Windows, and FreeBSD. I will admit the the API is currently quite low level, we hope to eventually add an easier to use higher-level API, but have not had a chance to do that yet.

chrisduerr: Allsorts is certainly looking like a great solution for supporting all platforms at the same time. So the primary question would be how well it performs and how nicely the API integrates with Alacritty, but I think generally a lower level API should be a pretty good fit for Alacritty.

Since you're already around @wezm, I'd imagine you might be aware of how allsorts compares to harfbuzz in performance? Specifically for running relatively simple combination of characters like ligatures on short-ish strings.

wezm: >  I'd imagine you might be aware of how allsorts compares to harfbuzz in performance? Specifically for running relatively simple combination of characters like ligatures on short-ish strings.

We have not done a lot of comparison with Harfbuzz at this point as the primary focus was accurate shaping, and matching or exceeding the performance of our previous shaping engine. There is an open issue to explore this further https://github.com/yeslogic/allsorts/issues/2

Whilst quite fast, it's probably safe to assume that Allsorts is not as fast as Harfbuzz, as Harfbuzz has been optimised for performance over many years. That is not to say we ignored performance though. We have profiled and optimised it in order to meet, and then exceed the performance of shaping engine it replaced in Prince.

felippemr:
👋 , what are the things we need to do in order to merge this pr?

chrisduerr:
@felippemr Check the previous comments please.

Mic92:
For nix there is also a [NUR](https://github.com/nix-community/NUR) package: `nix-shell -p nur.repos.xeals.alacritty-ligatures`

exorcist365 (approved):
This is fucking  awesome

Mic92:
Tested on both x11 and wayland.

alacritty_terminal/src/renderer/mod.rs
```diff
@@ -1046,39 +1041,21 @@ impl<'a> RenderApi<'a> {
...
-                    chars[0] = c;
-                    chars
-                }),
-                bg: color.unwrap_or(Rgb { r: 0, g: 0, b: 0 }),
-                fg: Rgb { r: 0, g: 0, b: 0 },
-                flags: Flags::empty(),
-                bg_alpha,
-            })
-            .collect::<Vec<_>>();
->>>>>>> 856cddc8739c32d8bbfff72dd3692f49359142a9
```
AsafFisher: WTF

Cargo.toml
```diff
@@ -13,4 +13,4 @@ debug = 1
 incremental = false
 
 [patch.crates-io]
-servo-freetype-sys = { path = "servo-freetype-proxy" }
+servo-freetype-sys = { path = "servo-freetype-proxy" }
```
AsafFisher: NL

AsafFisher:
Some parts of this code are poorly written. Please review it again!

Mic92:
Has anybody a newer rebase than https://github.com/zenixls2/alacritty/commit/126093ab33664247298db91538058d67310695f3 ?

carlosala:
Hi! This seems to be quite near to be possible to merge. I am not rust dev, but I would use ligatures quite a lot.
If any dev is interested, it seems to work but this changes have to be upstreamed to actual repo status and revise the @chrisduerr changes.
Thanks!

PR #2681, opened 2019-07-28
Switch to rfind_url for URL detection

### LINKED ISSUE #2627: Links without a space after it
If there is text after a link with non white space delimiter between them hovering over this text will show the underline under the link and clicking it will open the link but the underline is only under the link not the text.  Additionally if there is a second link in this text it will not detected.

Example:
A markdown link `[http://example.com](http://example.com)`

I'm on macos using 0.3.2

### LINKED ISSUE #2629: Links surrounded by angle brackets aren't correctly parsed

chrisduerr:
This switches to rfind_url for detecting URLs inside the grid. Instead
of expanding at the cursor position, the complete terminal is searched
from the bottom until the visible region is left with no active URL.

Fixes #2629.
Fixes #2627.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -75,11 +75,11 @@ impl Search for Term {
                 break;
             }
 
-            if iter.cur.col == last_col && !cell.flags.contains(cell::Flags::WRAPLINE) {
+            if iter.point().col == last_col && !cell.flags.contains(cell::Flags::WRAPLINE) {
                 break; // cut off if on new line or hit escape char
             }
 
-            point = iter.cur;
+            point = iter.point();
         }
 
         point
@@ -97,7 +97,7 @@ impl Search for Term {
                 break;
             }
 
-            point = iter.cur;
+            point = iter.point();
 
             if point.col == last_col && !cell.flags.contains(cell::Flags::WRAPLINE) {
                 break; // cut off if on new line or hit escape char
@@ -139,7 +139,7 @@ impl Search for Term {
 
             // Check if the bracket matches
             if c == end_char && skip_pairs == 0 {
-                return Some(iter.cur);
+                return Some(iter.point());
             } else if c == start_char {
                 skip_pairs += 1;
             } else if c == end_char {
@@ -1333,7 +1333,11 @@ impl Term {
         let mut extra_url_len = 0;
         let mut urls = Vec::new();
 
-        while let Some(cell) = iter.prev() {
+        let mut c = Some(iter.cell());
+        while let Some(cell) = c {
+            let point = iter.point();
+            c = iter.prev();
+
             // Skip double-width cell but extend URL length
             if cell.flags.contains(cell::Flags::WIDE_CHAR_SPACER) {
                 extra_url_len += 1;
@@ -1341,20 +1345,20 @@ impl Term {
             }
 
             // Interrupt URLs on line break
-            if iter.cur.col == last_col && !cell.flags.contains(cell::Flags::WRAPLINE) {
+            if point.col == last_col && !cell.flags.contains(cell::Flags::WRAPLINE) {
                 extra_url_len = 0;
                 parser.reset();
             }
 
             match parser.advance(cell.c) {
                 ParserState::Url(length) => {
-                    urls.push(Url::new(iter.cur, length + extra_url_len, num_cols))
+                    urls.push(Url::new(point, length + extra_url_len, num_cols))
                 },
                 ParserState::NoUrl => {
                     extra_url_len = 0;
 
                     // Stop searching for URLs once the viewport has been left without active URL
-                    if iter.cur.line > last_line.0 + display_offset {
+                    if point.line > last_line.0 + display_offset {
                         break;
                     }
                 },
@@ -1369,14 +1373,18 @@ impl Term {
         let mut url_text = String::new();
 
         let mut iter = self.grid.iter_from(url.start);
-        loop {
-            let cell = iter.next().unwrap();
 
-            url_text.push(cell.c);
+        let mut c = Some(iter.cell());
+        while let Some(cell) = c {
+            if !cell.flags.contains(cell::Flags::WIDE_CHAR_SPACER) {
+                url_text.push(cell.c);
+            }
 
-            if iter.cur == url.end {
+            if iter.point() == url.end {
                 break;
             }
+
+            c = iter.next();
         }
 
         url_text
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -20,17 +20,17 @@ use std::{io, mem, ptr};
 
 use font::{self, Size};
 use glutin::MouseCursor;
+use rfind_url::{Parser, ParserState};
 use unicode_width::UnicodeWidthChar;
 
 use crate::ansi::{
-    self, Attr, CharsetIndex, Color, CursorStyle, Handler, NamedColor, StandardCharset,
+    self, Attr, CharsetIndex, Color, CursorStyle, Handler, NamedColor, StandardCharset, TermInfo,
 };
 use crate::clipboard::{Clipboard, ClipboardType};
 use crate::config::{Config, VisualBellAnimation};
 use crate::cursor::CursorKey;
 use crate::grid::{
     BidirectionalIterator, DisplayIter, Grid, GridCell, IndexRegion, Indexed, Scroll,
-    ViewportPosition,
 };
 use crate::index::{self, Column, Contains, IndexRange, Line, Linear, Point};
 use crate::input::FONT_SIZE_STEP;
@@ -38,7 +38,7 @@ use crate::message_bar::MessageBuffer;
 use crate::selection::{self, Selection, SelectionRange, Span};
 use crate::term::cell::{Cell, Flags, LineLength};
 use crate::term::color::Rgb;
-use crate::url::{Url, UrlParser};
+use crate::url::Url;
 
 #[cfg(windows)]
 use crate::tty;
@@ -58,8 +58,6 @@ pub trait Search {
     fn semantic_search_left(&self, _: Point<usize>) -> Point<usize>;
     /// Find the nearest semantic boundary _to the point_ of provided point.
     fn semantic_search_right(&self, _: Point<usize>) -> Point<usize>;
-    /// Find the nearest URL boundary in both directions.
-    fn url_search(&self, _: Point<usize>) -> Option<Url>;
     /// Find the nearest matching bracket.
     fn bracket_search(&self, _: Point<usize>) -> Option<Point<usize>>;
 }
@@ -107,40 +105,6 @@ impl Search for Term {
         }
 
         point
-    }
-
-    fn url_search(&self, mut point: Point<usize>) -> Option<Url> {
-        let last_col = self.grid.num_cols() - 1;
-
-        // Switch first line from top to bottom
-        point.line = self.grid.num_lines().0 - point.line - 1;
-
-        // Remove viewport scroll offset
-        point.line += self.grid.display_offset();
-
-        // Create forwards and backwards iterators
-        let mut iterf = self.grid.iter_from(point);
-        point.col += 1;
-        let mut iterb = self.grid.iter_from(point);
-
-        // Find URLs
-        let mut url_parser = UrlParser::new();
-        while let Some(cell) = iterb.prev() {
-            if (iterb.cur.col == last_col && !cell.flags.contains(cell::Flags::WRAPLINE))
-                || url_parser.advance_left(cell)
-            {
-                break;
-            }
-        }
-
-        while let Some(cell) = iterf.next() {
-            if url_parser.advance_right(cell)
-                || (iterf.cur.col == last_col && !cell.flags.contains(cell::Flags::WRAPLINE))
-            {
-                break;
-            }
-        }
-        url_parser.url()
     }
 
     fn bracket_search(&self, point: Point<usize>) -> Option<Point<usize>> {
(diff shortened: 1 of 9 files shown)


alacritty_terminal/src/url.rs
```diff
@@ -1,318 +1,40 @@
...
-            for scheme in &URL_SCHEMES {
-                if self.state.starts_with(scheme) {
-                    return Some(Url { origin: self.origin - 1, text: self.state });
-                }
-            }
-        }
-
-        None
-    }
+    pub fn linear_bounds(&self, terminal: &Term) -> RangeInclusive<Linear> {
```
chrisduerr: @kchibisov This is what I meant in your previous PR by refactoring the `url_bounds_at_point` method. Making this significantly simpler seems to be one advantage of the rfind_url change, besides the obvious advantages when searching for more than just one URL.

I think this is pretty uncontroversial, but thought it might be interesting to see the changes, though if you have any further ideas for improvements, do let me know please.

kchibisov:
Fixes: #2627

alacritty_terminal/src/url.rs
kchibisov: Looks pretty simple. I'll take a look at other changes, but tomorrow (later today).

chrisduerr:
@kchibisov Good point, forgot about that. Is also covered by tests in rfind_url, so should be good to close that with this PR.

kchibisov:
@chrisduerr What do you think about adding the brackets mentioned in  #2648 to `rfind_url` before merging this work? So, no additional version bumpings in `Cargo.toml` + `Cargo.lock` will occur on a master.

chrisduerr:
> @chrisduerr What do you think about adding the brackets mentioned in #2648 to rfind_url before merging this work? So, no additional version bumpings in Cargo.toml + Cargo.lock will occur on a master.

I've wondered how to best do this, since there's two ways really.

Either we make the character illegal in URLs, which would likely be the simplest option and be pretty efficient.

Alternatively we could just add them to the bracket characters in rfind_url, which would also be trivial, but I'm not sure if it's worth it. I'm not sure if we ever want to allow them otherwise.

My plan was definitely to fix that with this PR too, but I just haven't made up my mind yet about which of these approaches I prefer.

kchibisov:
@chrisduerr from a logical point of view they a surround characters, from a practical point of view making them illegal will make sense, since these chars are super rare, and I personally didn't know about them until `man groff`. So the choice here is `correctness` vs `performance`, I'd personally go for correctness and logical consistency with open/close brackets.

chrisduerr:
> So the choice here is correctness vs performance, I'd personally go for correctness and logical consistency with open/close brackets.

Yeah I was kinda thinking the same thing. Though the performance difference is quite significant when adding additional brackets (roughly 20% in my benchmark), so I was kinda not looking forward to that. So it just doesn't really seem worth the difference, since the penalty will obviously always be there regardless of the presence of that character.

kchibisov:
@chrisduerr You should probably compare adding this char to `is_illegal` vs one bracket to `SURROUND_CHARACTERS`. It seems like overhead of surround_chars is something like `O(len(SURROUND_CHARACTERS) * n)` (n - number of chars, that you'll parse), so `+1` to `4` is a lot, but only just because its `4` (I've not done any serious math around here, just at first glance). However you can probably improve this with `HashMap`, that uses on stack buffer (no dynamic allocations, until they needed), but you should probably wait for const generics here.

chrisduerr:
I've tested it with `is_illegal` and adding it there slightly improves performance (probably some weird optimization thing). The performance difference of the surrounding characters always seems to be around 20%, regardless of the amount (varying slightly). However this is +20% of the previous value, so it is exponential, not linear growth.

Also regarding the `HashMap`, are you talking about the stored `surround_states`? Because the `SURROUND_CHARACTERS` do not use a static buffer. If you are talking about the `surround_states`, a dynamic buffer is necessary since it tracks more than one surround character at a time, since something like `(https://example.org/test((ing)heh))` needs to be properly handled.

kchibisov:
> I've tested it with is_illegal and adding it there slightly improves performance (probably some weird optimization thing).

`is_illigal` should be a jump table, which is pretty fast and adding one more `char`(since its just an `IntegerTy`) should be fine, probably you've decreased "gaps"(forgot the official name of this) in it, so that might be faster. It looks something like this in assembly.

```
.LJTI0_0:
  .long .LBB0_1-.LJTI0_0
  .long .LBB0_5-.LJTI0_0
  .long .LBB0_2-.LJTI0_0
  .long .LBB0_3-.LJTI0_0
  .long .LBB0_5-.LJTI0_0
  .long .LBB0_4-.LJTI0_0
```

> Because the SURROUND_CHARACTERS do not use a static buffer.

I was about this [one](https://github.com/chrisduerr/rfind_url/blob/39beb4f59aa2477558216e8c1c5a6166795558f5/src/lib.rs#L55). It seems like it doesn't use dynamic memory for it's storage (It's an array in the end of the day). However, I'm not very familiar on how rust handles such things (from `llvm-ir`, seems like just an static array somewhere, but I'm not quite sure), but I can be wrong here, don't have a dev version of rustc to check this. 

> If you are talking about the surround_states, a dynamic buffer is necessary since it tracks more than one surround character at a time, since something like (https://example.org/test((ing)heh)) needs to be properly handled.

I've forgotten to mention this one, you can probably eliminate all dynamic buffer overhead with `SmallVector` or with preallocating buffer in a constructor (e.g. `Vec::with_capacity(16)`), since the number of surrounding_states should be small.

Just a small micro optimization hint, you should probably use `self.surround_states.clear()` [here](https://github.com/chrisduerr/rfind_url/blob/39beb4f59aa2477558216e8c1c5a6166795558f5/src/lib.rs#L214), to avoid allocations.

kchibisov:
I think, I've found a bug here. If you start a `zsh`(without config). write this line `git clone https://github.com/jwilm/alacritty` and try to click on url's `y` char this will result in nothing (no opening and highlighting), clicking on `t` works (`y` will be highlighted). Seems like off-by-one somewhere. (on master everything works fine)

alacritty_terminal/src/term/mod.rs
```diff
@@ -1368,9 +1319,71 @@ impl Term {
...
+        }
+
+        urls
+    }
+
+    pub fn url_to_string(&self, url: &Url) -> String {
+        let mut url_text = String::new();
+
+        let mut iter = self.grid.iter_from(url.start);
+        loop {
```
kchibisov: I think this function can be much simpler and expressed entirely in iterators. Oh, and you're skipping first char here you can add `println!("URL: {}", url_text)`, so opening `https://github.com/chrisduerr/rfind_url` results in `URL: ttps://github.com/chrisduerr/rfind_url`

alacritty_terminal/src/term/mod.rs
```diff
@@ -1368,9 +1319,71 @@ impl Term {
     pub fn clipboard(&mut self) -> &mut Clipboard {
         &mut self.clipboard
     }
+
+    pub fn urls(&self) -> Vec<Url> {
+        let display_offset = self.grid.display_offset();
+        let num_cols = self.grid.num_cols().0;
+        let last_col = Column(num_cols - 1);
```
kchibisov: You don't need `-1` here, you have a big chance of skipping last `WRAPLINE` cell (in fish shell you're skipping it).

chrisduerr:
> is_illigal should be a jump table, which is pretty fast and adding one more char(since its just an IntegerTy) should be fine, probably you've decreased "gaps"(forgot the official name of this) in it, so that might be faster.

The `match` statement in `is_illegal` does indeed compile to a jump table and I've considered that the performance difference might be related to memory alignment, however I'd have to actually take a look at the output to be sure.

> I was about this one. It seems like it doesn't use dynamic memory for it's storage (It's an array in the end of the day). However, I'm not very familiar on how rust handles such things (from llvm-ir, seems like just an static array somewhere, but I'm not quite sure), but I can be wrong here, don't have a dev version of rustc to check this.

Arrays in Rust do not use any dynamic memory, so they can just be easily copied around. However, when testing I noticed that it is possible to improve performance by unrolling the loop looking for surround characters. Trying to either use a function here or a macro might give a good amount of extra performance.

> I've forgotten to mention this one, you can probably eliminate all dynamic buffer overhead with SmallVector or with preallocating buffer in a constructor (e.g. Vec::with_capacity(16)), since the number of surrounding_states should be small.

I've used SmallVec before but the performance benefits I've gotten out of it have been very mediocre. Though it's obviously possible that just didn't fit the usecase for it, so looking into it again would likely be worth it. I don't think `Vec::with_capacity` would improve anything though, since we're unlikely to frequently grow this vector and having even a capacity of `1` would likely just be overhead for (probably) the majority of URLs.

> Just a small micro optimization hint, you should probably use self.surround_states.clear() here, to avoid allocations.

I wasn't able to benchmark any performance differences (at least nothing outside of margin of error) between `Vec::new` and `Vec::clear` and neither of them should allocate. However calling `Vec::clear` is likely the better choice here.

alacritty_terminal/src/term/mod.rs
kchibisov: Maybe, I'm wrong here, but there's something broken with multiline urls, but there's definitely something odd with all of this story. For e.g. if I create a long line in vim with url, it wouldn't be considered as multiline one (I don't have any any vim decorations on the left), same here for the fish shell, if I write a url and move it to the right with space chars(to make it appear on multiple lines), it wouldn't be considered as multiline one, until I resize a terminal.

kchibisov:
> I wasn't able to benchmark any performance differences (at least nothing outside of margin of error) between Vec::new and Vec::clear and neither of them should allocate.

It's about reusing buffer, that you've already allocated during parsing, with `Vec::new()` you'll drop it, but with `Vec::clear` you'll reuse it. So in case you'll face more surrounding chars after `reset` you'll probably have space for them.

alacritty_terminal/src/term/mod.rs
chrisduerr: > opening https://github.com/chrisduerr/rfind_url results in URL: ttps://github.com/chrisduerr/rfind_url

This explains so much. I've actually extended the URL in another place because I thought there was an off by one issue, but it seems like this is the problem.

Doesn't help that Firefox apparently doesn't complain at all when opening `ttps://example.org`.

kchibisov: Yeah, it auto adds h to url.

chrisduerr: Regarding simplification though, I don't think iterators would make this much cleaner. Especially since we need a reference to the iterator to make sure we can return early.

alacritty_terminal/src/term/mod.rs
chrisduerr: We kinda want a `-1` for correctness here, similar to the issue in `url_to_string` though the iterators do not include the cell they're right above. Since cells are 0-based the last column has the index num_cols - 1.

It might just work to start 'outside' the actual grid, however that seems kinda troublesome and could actually lead to index out of bounds if things ever change in the future.

kchibisov: Yeah, I’m just wondering what’s going wrong with long wrapped url in neovim/fish. Unfortunately, I can’t debug this one atm.

alacritty_terminal/src/term/mod.rs
kchibisov: > Regarding simplification though, I don't think iterators would make this much cleaner. Especially since we need a reference to the iterator to make sure we can return early.

After looking at this function once again, making it correct probably will be enough here.

alacritty_terminal/src/term/mod.rs
chrisduerr: I'll try to find a clean solution for fixing the iterators and I'll try to repro once that's fixed.

chrisduerr: I've resolved the mentioned issues in the latest version. I don't think there's anything we can do about long wrapped URLs in neovim, since these are actually broken by neovim with a hard linebreak at the end.

chrisduerr:
@kchibisov There's been a lot to unpack in the comments here and some very great suggestions. I think everything should work properly now and all potential performance improvements for `rfind_url` should either be fixed or have issues to track them available.

If there's anything left that you think hasn't been addressed, please let me know.

kchibisov:
> I think, I've found a bug here. If you start a zsh(without config). write this line git clone https://github.com/jwilm/alacritty and try to click on url's y char this will result in nothing (no opening and highlighting), clicking on t works (y will be highlighted). Seems like off-by-one somewhere. (on master everything works fine)

This issue is steel here.

kchibisov:
Also left most url character is not clickable.

kchibisov:
Also I've found some odd behavior in fish shell (everything is fine on master under the same conditions). If my text ends on `-1` from line end and I press `space` it results in doubling the character and other odd behaviors after pressing space.

Screenshot_Tue 30 Jul 2019 06:01:55 PM MSK

Screenshot_Tue 30 Jul 2019 06:02:07 PM MSK

alacritty_terminal/src/url.rs
```diff
@@ -1,318 +1,40 @@
...
-        if cell.flags.contains(Flags::WIDE_CHAR_SPACER) {
-            return false;
-        }
-
-        self.advance(cell.c, self.state.len())
+    pub fn contains(&self, point: impl Into<Point<usize>>) -> bool {
+        let point = point.into();
+        point.line <= self.start.line
+            && point.line >= self.end.line
+            && (point.line != self.start.line || point.col > self.start.col)
```
kchibisov: You need inclusive check `point.col >= self.start.col` and `point.col <= self.end.col`

chrisduerr: Makes sense yeah, thanks!

chrisduerr:
> I think, I've found a bug here. If you start a zsh(without config). write this line git clone https://github.com/jwilm/alacritty and try to click on url's y char this will result in nothing (no opening and highlighting), clicking on t works (y will be highlighted). Seems like off-by-one somewhere. (on master everything works fine)

I can not reproduce this.

> Also I've found some odd behavior in fish shell (everything is fine on master under the same conditions). If my text ends on -1 from line end and I press space it results in doubling the character and other odd behaviors after pressing space more times.

Is this only present with fish shell and not with bash or zsh? I can't reproduce this with either. Also I'd assume that URLs are irrelevant in this case?

kchibisov:
Only fish, yes, irrelevant.

kchibisov:
> I can not reproduce this

You’ve just fixed it with check that I’ve suggested

chrisduerr:
So the only issue left is with the fish shell in the last column? I'll have to install fish in a sec to give it a shot, very surprised it would only happen on fish though tbh.

kchibisov:
> Also I've found some odd behavior in fish shell (everything is fine on master under the same conditions). If my text ends on -1 from line end and I press space it results in doubling the character and other odd behaviors after pressing space more times.

> Is this only present with fish shell and not with bash or zsh? I can't reproduce this with either. Also I'd assume that URLs are irrelevant in this case?

Alarm, I was able to repro this on master (seems like fish shell problem). This is just super odd, you should move your text to the end only by one space at a time. I've probably pressed it for a while on master and failed to repro this. Sorry

chrisduerr:
> Alarm, I was able to repro this on master (seems like fish shell problem). This is just super odd, you should move your text to the end only by one space at a time. I've probably pressed it for a while on master and failed to repro this. Sorry

Now you've kinda gotten me curious though. Would you be willing to test another terminal emulator real quick?

kchibisov:
> Would you be willing to test another terminal emulator real quick?

Sure

kchibisov:
I wasn't able to repro this in other terminal emulators. The most important part of this, that I can't repro this when `alacritty` is built in release mode (only debug). Just for you, how this look like after additional spaces
Screenshot_Tue 30 Jul 2019 07:24:33 PM MSK

chrisduerr:
> I wasn't able to repro this in other terminal emulators. The most important part of this, that I can't repro this when alacritty is built in release mode (only debug). Just for you, how this look like after additional spaces

This really sounds like there might be a real issue somewhere, would you mind opening a separate issue to maybe further troubleshoot? Sounds like this PR is unrelated then.

alacritty_terminal/src/url.rs
```diff
@@ -1,318 +1,40 @@
...
-
-// See https://tools.ietf.org/html/rfc3987#page-13
-const URL_SEPARATOR_CHARS: [char; 10] = ['<', '>', '"', ' ', '{', '}', '|', '\\', '^', '`'];
-const URL_DENY_END_CHARS: [char; 7] = ['.', ',', ';', ':', '?', '!', '('];
-const URL_SCHEMES: [&str; 8] =
-    ["http://", "https://", "mailto:", "news:", "file://", "git://", "ssh://", "ftp://"];
-
-/// URL text and origin of the original click position.
-#[derive(Debug, PartialEq)]
 pub struct Url {
```
kchibisov: Would you mind adding `#derive(Debug)` (at least) here?

chrisduerr: Why? I usually like to leave out derives when they aren't necessary, since they always have a decent impact on compile time. And currently this isn't exported publicly either, which is usually a big argument for adding all the possible derives.

If it's about debugging issues, it shouldn't be a problem to add that while debugging. But it's never printed anywhere, so a debug output isn't really necessary I think.

kchibisov: On my system system the build times with `#[derive(Debug)]` and without it are identical ( in debug mode, which I'm usually running to test usability PRs ). I was just adding this macro anyway, while testing the PR.

If you're about release builds, it's about the way rust is doing `lto` and how it calls linker ( and the choice of the linker ) with this target triple `x86_64-pc-linux-gnu`.

It's up to you though.

chrisduerr: > On my system system the build times with #[derive(Debug)] and without it are identical ( in debug mode, which I'm usually running to test usability PRs ). I was just adding this macro anyway, while testing the PR.

Oh yeah, being cautious with adding a lot of derives will only really make a significant impact when you have a ton or complex derives (or both).

Honestly if there was a `derive(Debug)` statement and you would have asked for its removal, I would have likely responded in a similar fashion. In general I just don't think worrying too much about derives makes a lot of sense unless the type is involved in a public API.

Really the primary reason why I'm still considering this is that with the library split, we'll likely have the `Url` as part of `alacritty_terminal`'s API, in which case it would be a good idea to provide as many derives as possible.

Issue #2680, opened 2019-07-28
Underline starting in last line not rendered when scrolled up


chrisduerr:
If an underline is started in the last line while scrolled up in the terminal, the underline is not rendered correctly.

This applies both for underlines printed using the escape sequence and for underlines created for URL highlighting.

Quick investigation shows this might be caused by the `update_lines` method or some other step in the rendering process. It seems like the cells have the underline flag set correctly, or at least some flags have an underline flag set, I haven't checked if they're the correct cells yet.

kchibisov:
Seems like `strikethrough` is also affected. However selecting any cell (even empty) on line with underline + strikethrough shows both underline and strikethrough correctly.

chrisduerr:
Since strikeout and underline are handled pretty much exactly the same way, this does make sense.

Good catch on the selection though. I'd expect this to be relatively easy to resolve once the cause is found, however I was caught up with finishing up my PR for moving Alacritty to rfind_url.

kchibisov:
@chrisduerr I guess, I've found a starting point and it seems like something is super weird here. The code, which is seems to be responsible for it is [here](https://github.com/jwilm/alacritty/blob/54dca06bfbac50da0adb33c928e376b5c7986c4a/alacritty_terminal/src/renderer/rects.rs#L78). This "if" condition is **always** false and the cell.column, cell.line values are super odd. Also I've never seen `cell.line == size_info.lines()`.

```diff
diff --git a/alacritty_terminal/src/renderer/rects.rs b/alacritty_terminal/src/renderer/rects.rs
index b4f8a01..95f4028 100644
--- a/alacritty_terminal/src/renderer/rects.rs
+++ b/alacritty_terminal/src/renderer/rects.rs
@@ -17,6 +17,7 @@ use crate::index::Point;
 use crate::term::cell::Flags;
 use crate::term::color::Rgb;
 use crate::term::{RenderableCell, SizeInfo};
+use std::process;

 #[derive(Debug, Copy, Clone)]
 pub struct Rect<T> {
@@ -66,6 +67,10 @@ impl<'a> Rects<'a> {
     /// Update the stored lines with the next cell info.
     pub fn update_lines(&mut self, size_info: &SizeInfo, cell: &RenderableCell) {
         for line in self.active_lines.iter_mut() {
+            if size_info.lines() == cell.line {
+                process::exit(127);
+            }
+
             match line.range {
                 // Check for end if line is present
[...]

PR #2682, opened 2019-07-28
Fix last line update criterion in update_lines.

kchibisov:
`SizeInfo` lines/cols start from 1, however `RenderableCell` lines/cols start from 0, so explicit inc/dec is required. Also checking for `cell.column == size_info.cols() + 1` seems incorrect here, since it will be called, when our terminal last line will be entirely occupied by text, so underline/strikethrough wouldn't be updated most of the time.

Fixes: #2680

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - On Linux, respect fontconfig's `embeddedbitmap` configuration option
 - Selecting trailing tab with semantic expansion
 - URL parser incorrectly handling Markdown URLs and angled brackets
+- Last line underline/strikeout update
 
 ## 0.3.3
 
--- a/alacritty_terminal/src/display.rs
+++ b/alacritty_terminal/src/display.rs
@@ -519,20 +519,12 @@ impl Display {
 
                 self.renderer.with_api(config, &size_info, |mut api| {
                     // Iterate over all non-empty cells in the grid
-                    if let Some(last_cell) = grid_cells.last() {
-                        let (last_col, last_line) = (last_cell.column, last_cell.line);
-                        for cell in grid_cells {
-                            // Update underline/strikeout
-                            if cell.line == last_line && cell.column == last_col {
-                                // Mark last cell
-                                rects.update_lines(&cell, true);
-                            } else {
-                                rects.update_lines(&cell, false);
-                            }
+                    for cell in grid_cells {
+                        // Update underline/strikeout
+                        rects.update_lines(&cell);
 
-                            // Draw the cell
-                            api.render_cell(cell, glyph_cache);
-                        }
+                        // Draw the cell
+                        api.render_cell(cell, glyph_cache);
                     }
                 });
             }
--- a/alacritty_terminal/src/display.rs
+++ b/alacritty_terminal/src/display.rs
@@ -519,12 +519,20 @@ impl Display {
 
                 self.renderer.with_api(config, &size_info, |mut api| {
                     // Iterate over all non-empty cells in the grid
-                    for cell in grid_cells {
-                        // Update underline/strikeout
-                        rects.update_lines(&size_info, &cell);
+                    if let Some(last_cell) = grid_cells.last() {
+                        let (last_col, last_line) = (last_cell.column, last_cell.line);
+                        for cell in grid_cells {
+                            // Update underline/strikeout
+                            if cell.line == last_line && cell.column == last_col {
+                                // Mark last cell
+                                rects.update_lines(&cell, true);
+                            } else {
+                                rects.update_lines(&cell, false);
+                            }
 
-                        // Draw the cell
-                        api.render_cell(cell, glyph_cache);
+                            // Draw the cell
+                            api.render_cell(cell, glyph_cache);
+                        }
                     }
                 });
             }
--- a/alacritty_terminal/src/renderer/rects.rs
+++ b/alacritty_terminal/src/renderer/rects.rs
@@ -64,7 +64,7 @@ impl<'a> Rects<'a> {
     }
 
     /// Update the stored lines with the next cell info.
-    pub fn update_lines(&mut self, size_info: &SizeInfo, cell: &RenderableCell) {
+    pub fn update_lines(&mut self, cell: &RenderableCell, last_cell: bool) {
         for line in self.active_lines.iter_mut() {
             match line.range {
                 // Check for end if line is present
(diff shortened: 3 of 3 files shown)


chrisduerr:
> Also checking for cell.column == size_info.cols() + 1 seems incorrect here, since it will be called, when our terminal last line will be entirely occupied by text, so underline/strikethrough wouldn't be updated most of the time.

That's a very good point, I forgot that we filter out empty cells in the `RenderableCellsIter::next` method. However, I think logically this is still necessary. With multiple underlines over multiple cells in the last line, we wouldn't want to call this multiple times. We only want to call this once, after all other cells have been processed instead of pushing a ton of tiny lines when in the last line.

I think it might be possible to check at the end of the `update_lines` function if there still is an active line and then push that to the vector, regardless of the position of the line. Then we can just remove [this](https://github.com/jwilm/alacritty/blob/f9e414a9cdf42adc84487c757f5c59237bcdf9f3/alacritty_terminal/src/renderer/rects.rs#L78-L85) completely.

kchibisov:
> I think it might be possible to check at the end of the update_lines function if there still is an active line and then push that to the vector, regardless of the position of the line. Then we can just remove this completely.

I probably don't understand, how to implement this one only by touching `update_lines`, since there is no state sharing, if I'll check for active lines in the end of the `update_lines`, I'll do this for all cells eventually, so the same amount of update will be performed, as in the current implementation. Returning something from `update_lines` seems a little bit odd here.

However I think, I've found one more problem here. The behavior of `create_rect` depends on which cell `create_rect` was called.

If you run with following check `if size_info.lines() == cell.line + 1 && size_info.cols() == cell.column + 1`, the resulted highlight will be fine.
Screenshot_Sun 28 Jul 2019 06:17:03 PM MSK

But with `if size_info.lines() == cell.line + 1 || size_info.cols() == cell.column + 1` (or without `size_info.cols` part) the result is different, look at the end of the line.

Screenshot_Sun 28 Jul 2019 06:16:08 PM MSK

I'll check what's wrong with this one later. Seems like some computations in `create_rect` round in a wrong way.

chrisduerr:
> I probably don't understand, how to implement this one only by touching update_lines, since there is no state sharing, if I'll check for active lines in the end of the update_lines, I'll do this for all cells eventually, so the same amount of update will be performed, as in the current implementation. Returning something from update_lines seems a little bit odd here.

Sorry, I just saw the `for` loop and immediately assumed that it would handle all cells, not just one cell, which obviously is incorrect and which is why I've done it the way I have originally. So it's correct that my proposal would not work.

> But with if size_info.lines() == cell.line + 1 || size_info.cols() == cell.column + 1 (or without size_info.cols part) the result is different, look at the end of the line.

This is an incorrect logic which doesn't really make sense since you basically add a lot of single cell underlines (or zero-cell in the last one I think). So I wouldn't worry about it not working, since it's super incorrect anyways.

I think I've got a better solution though. I'm not sure if this would work 100%, but previously I think I've added the last line in the `rects` method. I'm unsure why I changed this and the reason for that change might prevent us from doing it, but if it works, we could just create the last line in there.

kchibisov:
> I think I've got a better solution though. I'm not sure if this would work 100%, but previously I think I've added the last line in the rects method. I'm unsure why I changed this and the reason for that change might prevent us from doing it, but if it works, we could just create the last line in there.

I've tried it, but it feels like more like a hack, then a real solution, but maybe I'm implementing it wrong.

What do you think about sending entire grid to this function, so we can organize processing in a more cleaner way(implement your first idea)? We don't use `grid_cells` anywhere else other than for `update_lines`, so having access to all grid can be beneficial here.

chrisduerr:
> I've tried it, but it feels like more like a hack, then a real solution, but maybe I'm implementing it wrong.

I don't think it's a super uncommon pattern to finish off a loop like this with a statement that collects whatever is 'left over'. However having this done in a completely separate method is certainly unpractical and likely a bit ugly.

> What do you think about sending entire grid to this function, so we can organize processing in a more cleaner way(implement your first idea)? We don't use grid_cells anywhere else other than for update_lines, so having access to all grid can be beneficial here.

I'm not a fan of iterating over the complete grid twice if pushing the last cell in `rects` would do the job.

Another very, very simple solution would be simply to tell `update_lines` when things are over. Like passing a boolean `last_cell` or something like that. That wouldn't be super clean or idiomatic, but it would certainly be pragmatic.

Of course there might be a better way to do this in a larger scale, however I can't think of a concrete example and I think I'd rather stay away from that unless there's a good proposal.

Thinking about it a bit, the only 'bigger' rework I can think of which might make sense is to always push a finished rectangle for every new cell with a line and then update the last element in the vector when the line just keeps going. This way the last element would be automatically added to the vector, since it's added from the beginning. Maybe that would be cleaner and works out?

kchibisov:
@chrisduerr I've added some info in issue tracker for #2680 . I have a strongly feeling, that the thing that is addressed in this PR is only a part of the problem, but if you don't think so, just let me know.

chrisduerr:
> @chrisduerr I've added some info in issue tracker for #2680 . I have a strongly feeling, that the thing that is addressed in this PR is only a part of the problem, but if you don't think so, just let me know.

Looking at that comment, to me it sounds like it just confirms what has already been said. If it's on a prompt line, of course it'll also cause problems since then that's the last line. As we've previously established this is about an outstanding underline/strikeout not getting comitted.

This is also why it only affects the last one. If there is any cell after the underline it will get comitted because all underlines get comitted as soon as they end. With the last one we just can't tell, since there is no character sent which 'ends' the underline after it has been completed.

kchibisov:
The approach is a little bit messy, but at least it works and rendering of underline/strikeout is consistent.

chrisduerr:
@kchibisov I think this approach is very hackish, have you considered my last proposal?

> Thinking about it a bit, the only 'bigger' rework I can think of which might make sense is to always push a finished rectangle for every new cell with a line and then update the last element in the vector when the line just keeps going. This way the last element would be automatically added to the vector, since it's added from the beginning. Maybe that would be cleaner and works out?

kchibisov:
@chrisduerr I'll try to implement it. I've implemented this one first, since it was the simplest one and it just works, however I tend to dislike it.

chrisduerr:
Yeah, while I think it should work great in general, it's more of a worst case scenario. It's good that we have a working solution ready, but I'd like to try for something more optimal if possible.

kchibisov:
This should implement what you've suggested here, it feels simpler, but the last element update is something I'm worry about here. We have to look `2` elements from the end, because we have to update proper rects (underline or strikeout).

I decided to leave structs/other functions in this file untouched, because I want to hear your ideas on how better to finish this rework (since It's your proposal).

Anyway, this approach seems better at doing things.

chrisduerr:
@kchibisov I think I actually want to play around with tweaking this myself for a bit just to compare a few different approaches. I'll get back to you on this within the next one or two days hopefully.

chrisduerr:
I've decided to create a pull request to your branch [here](https://github.com/kchibisov/alacritty/pull/1), to make it easy to find and comment on the code.

Please let me know what you think. I feel like this approach might be a little better, though it is a significant change.

chrisduerr:
See follow-up PR here: https://github.com/jwilm/alacritty/pull/2701

PR #2697, opened 2019-08-03
Move modifier check before URL search

chrisduerr:
This makes sure that the URL search is only initiated when all required
modifiers are held down. This should improve performance with long URLs.

This was previously the case, however 72dfa47 introduced a regression
which broke this.

### DIFF
--- a/alacritty_terminal/src/input.rs
+++ b/alacritty_terminal/src/input.rs
@@ -389,23 +389,32 @@ enum MousePosition {
 }
 
 impl<'a, A: ActionContext + 'a> Processor<'a, A> {
-    fn mouse_position(&mut self, point: Point) -> MousePosition {
+    fn mouse_position(&mut self, point: Point, modifiers: ModifiersState) -> MousePosition {
+        let mouse_mode =
+            TermMode::MOUSE_MOTION | TermMode::MOUSE_DRAG | TermMode::MOUSE_REPORT_CLICK;
+
         let buffer_point = self.ctx.terminal().visible_to_buffer(point);
 
         // Check message bar before URL to ignore URLs in the message bar
         if let Some(message) = self.message_at_point(Some(point)) {
             if self.message_close_at_point(point, message) {
-                MousePosition::MessageBarButton
+                return MousePosition::MessageBarButton;
             } else {
-                MousePosition::MessageBar
+                return MousePosition::MessageBar;
             }
-        } else if let Some(url) =
-            self.ctx.terminal().urls().drain(..).find(|url| url.contains(buffer_point))
-        {
-            MousePosition::Url(url)
-        } else {
-            MousePosition::Terminal
         }
+
+        // Check for URL at point with required modifiers held
+        if self.mouse_config.url.mods().relaxed_eq(modifiers)
+            && (!self.ctx.terminal().mode().intersects(mouse_mode) || modifiers.shift)
+            && self.mouse_config.url.launcher.is_some()
+        {
+            if let Some(url) = self.ctx.terminal().urls().drain(..).find(|url| url.contains(buffer_point)) {
+                return MousePosition::Url(url);
+            }
+        }
+
+        MousePosition::Terminal
     }
 
     #[inline]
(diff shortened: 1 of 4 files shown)


kchibisov:
@chrisduerr Just a few notes on why this "regression" happened. Generally, this is not a regression, since users without `modifiers` for `url` will do the search anyway, even with your patch, so my changes made all users equal in this direction (this is how equality works these days). So here, you're introducing advantages for users with modifiers and the problem for users without them are unresolved (this is a problem, since we're advertising `None` modifiers in default config).

I just want to clarify things why I've introduced this regression.

chrisduerr:
> So here, you're introducing advantages for users with modifiers and the problem for users without them are unresolved (this is a problem, since we're advertising None modifiers in default config).

Yes, that is correct, however this also affects active mouse mode regardless of modifiers.

In the original code this was intentionally done this way, since it makes sense to stagger the operations in a way where the most expensive one is done only when all other conditions are met. I think generally that's a very good idea.

I'd still say this is a 'regression' because of it. Even if a performance improvement only affects a limited amount of users (with modifiers), or a limited number of usecases (mouse mode enabled), it's still a performance optimization. Removing that impacts the performance for these people and usecases negatively, so if it is not done intentionally (by which I mean I wasn't aware of it when I merged it), it's a regression.

kchibisov:
One more clarification, I'm not against your changes here (I also like the idea of checking before searching), but with available `None` modifier these performance improvements are not improvements in general.

alacritty_terminal/src/input.rs
```diff
@@ -389,23 +389,32 @@ enum MousePosition {
...
+                return MousePosition::MessageBar;
             }
-        } else if let Some(url) =
-            self.ctx.terminal().urls().drain(..).find(|url| url.contains(buffer_point))
+        }
+
+        // Check for URL at point with required modifiers held
+        if self.mouse_config.url.mods().relaxed_eq(modifiers)
+            && (!self.ctx.terminal().mode().intersects(mouse_mode) || modifiers.shift)
+            && self.mouse_config.url.launcher.is_some()
```
kchibisov: I think url modifiers check can be moved to a separate function or back to `mouse_moved`, and we can just pass a `bool` (smth like search_for_urls) to mouse_position. This should make sense, since deciding to do smth is a `bool` and `ModifiersState`.

kchibisov: The move to a separate function is motivating by this [problem](https://github.com/jwilm/alacritty/pull/2413#issuecomment-495264005)

chrisduerr: Hmm, I think I'd like to keep it here for now. It makes sense where it is and it's currently not needed anywhere else. Changing it also doesn't directly resolve the problem.

So I think this change is probably best done once a PR to resolve this is proposed.

kchibisov: I just don't like that `mouse_position` requires now `ModifierState` as argument, so if we decided to reuse `mouse_position`, we'll deal with this restriction, which isn't generally a problem now, but can be annoying in some cases (using in `on_message_bar_click` in case we decide to add more buttons there or making `mouse_moved` public). So for now I have a strongly feeling that this code will be removed from this function anyway (and probably in a near feature), but it's up to you ofc.

kchibisov:
What do you think about advertising `Control` or `Shift` instead of `None` in default configuration file?

alacritty_terminal/src/input.rs
chrisduerr: I don't think it's ever a good idea to write code for something that might or might not happen in the future. Even if there's a 100% chance that things are going to change in the future, the job of this PR is to write the best code possible *right now*. And I think that extracting that right now would just introduce additional complexity, rather than improving anything in its current situation.

If it's guaranteed to be changed in the future, that's fine. Let's do it in the future then.

chrisduerr:
> What do you think about advertising Control or Shift instead of None in default configuration file?

I can only go based on what I'd expect a default to be. And I personally run it without any modifiers. The same thing goes for most other terminal emulators I'm familiar with.

I don't think the performance difference with normal URLs is sufficient to change this, even though  Alacritty is focused on performance in this case the performance is only impacted with very big URLs. So I think our best choice is to just limit the URL length (I'm personally for 65535, but I don't have any very old systems) if we're interested in resolving the DoS issue.

kchibisov:
> So I think our best choice is to just limit the URL length (I'm personally for 65535, but I don't have any very old systems) if we're interested in resolving the DoS issue.

In case you want to test things on something slow I can help with this one, just @ me at some point, when you'll think that these checks are required to get something done.

Thanks for answering my questions though.

Issue #2699, opened 2019-08-04
Hovering mouse over message bar results in crash


kchibisov:
When you hover mouse over message bar `alacritty` will crash, this issue was introduced by 9dddf649a15d103295f4ce97b8ae4c178c9623e0 and presented on master.

PR #2700, opened 2019-08-04
Fix mouse hover over message bar

kchibisov:
`visible_to_buffer` can't be used on a point from message bar, such use
of function will result in crash.

This commit fixes the regression introduced in 9dddf64.

Fixes: #2699

### DIFF
--- a/alacritty_terminal/src/input.rs
+++ b/alacritty_terminal/src/input.rs
@@ -393,7 +393,6 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         let mouse_mode =
             TermMode::MOUSE_MOTION | TermMode::MOUSE_DRAG | TermMode::MOUSE_REPORT_CLICK;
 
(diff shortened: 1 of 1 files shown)


chrisduerr:
> visible_to_buffer can't be used on a point from message bar, such use
of function will result in crash.

It seems like we've got an `assert!` in place which will cause the crash, is that correct? Are there other instances where there's potential for this to crash Alacritty? I'm wondering if returning a `Result` might be better here, though performance should be considered. But it seems like a pretty simple mistake to make.

kchibisov:
> It seems like we've got an assert! in place which will cause the crash, is that correct?

Yes, you're correct here.

> I'm wondering if returning a Result might be better here, though performance should be considered. But it seems like a pretty simple mistake to make.

I don't think that returning result matters here, since we have an assert, which absolutely make sense. The only problem I see, that we don't have UI tests to catch such problems during reviews.

PR #2705, opened 2019-08-05
Fix multiple lines with the same color on the same line

kchibisov:
This commit fixes regression introduced in ddee14a.

P.s.
I really with feel bad for not spotting this one in a first place.

### DIFF
--- a/alacritty_terminal/src/display.rs
+++ b/alacritty_terminal/src/display.rs
@@ -27,7 +27,7 @@ use crate::config::{Config, StartupMode};
 use crate::index::Line;
 use crate::message_bar::Message;
 use crate::meter::Meter;
-use crate::renderer::rects::{Rect, Lines};
+use crate::renderer::rects::{RenderRect, Lines};
 use crate::renderer::{self, GlyphCache, QuadRenderer};
 use crate::sync::FairMutex;
 use crate::term::color::Rgb;
@@ -537,7 +537,7 @@ impl Display {
                 // Create a new rectangle for the background
                 let start_line = size_info.lines().0 - text.len();
                 let y = size_info.padding_y + size_info.cell_height * start_line as f32;
-                rects.push(Rect::new(
+                rects.push(RenderRect::new(
                     0.,
                     y,
                     size_info.width,
--- a/alacritty_terminal/src/display.rs
+++ b/alacritty_terminal/src/display.rs
@@ -27,7 +27,7 @@ use crate::config::{Config, StartupMode};
 use crate::index::Line;
 use crate::message_bar::Message;
 use crate::meter::Meter;
-use crate::renderer::rects::{Rect, Rects};
+use crate::renderer::rects::{Rect, Lines};
 use crate::renderer::{self, GlyphCache, QuadRenderer};
 use crate::sync::FairMutex;
 use crate::term::color::Rgb;
@@ -511,7 +511,7 @@ impl Display {
 
         {
             let glyph_cache = &mut self.glyph_cache;
-            let mut rects = Rects::new();
+            let mut lines = Lines::new();
 
             // Draw grid
             {
@@ -521,7 +521,7 @@ impl Display {
                     // Iterate over all non-empty cells in the grid
                     for cell in grid_cells {
                         // Update underline/strikeout
-                        rects.update_lines(&cell, &size_info, &metrics);
+                        lines.update(&cell);
 
                         // Draw the cell
                         api.render_cell(cell, glyph_cache);
@@ -529,14 +529,21 @@ impl Display {
                 });
             }
 
+            let mut rects = lines.into_rects(&metrics, &size_info);
+
             if let Some(message) = message_buffer {
                 let text = message.text(&size_info);
 
                 // Create a new rectangle for the background
                 let start_line = size_info.lines().0 - text.len();
                 let y = size_info.padding_y + size_info.cell_height * start_line as f32;
-                let rect = Rect::new(0., y, size_info.width, size_info.height - y);
-                rects.push(rect, message.color());
+                rects.push(Rect::new(
+                    0.,
+                    y,
+                    size_info.width,
+                    size_info.height - y,
+                    message.color(),
+                ));
 
                 // Draw rectangles including the new background
                 self.renderer.draw_rects(config, &size_info, visual_bell_intensity, rects);
--- a/alacritty_terminal/src/display.rs
+++ b/alacritty_terminal/src/display.rs
@@ -27,7 +27,7 @@ use crate::config::{Config, StartupMode};
 use crate::index::Line;
 use crate::message_bar::Message;
 use crate::meter::Meter;
-use crate::renderer::rects::{RenderRect, Lines};
+use crate::renderer::rects::{RenderRect, RenderLines};
 use crate::renderer::{self, GlyphCache, QuadRenderer};
 use crate::sync::FairMutex;
 use crate::term::color::Rgb;
@@ -511,7 +511,7 @@ impl Display {
 
         {
             let glyph_cache = &mut self.glyph_cache;
-            let mut lines = Lines::new();
+            let mut lines = RenderLines::new();
 
             // Draw grid
             {
--- a/alacritty_terminal/src/renderer/rects.rs
+++ b/alacritty_terminal/src/renderer/rects.rs
@@ -37,6 +37,7 @@ impl<T> Rect<T> {
 struct Line {
     rect: Rect<f32>,
     start: Point,
+    end: Point,
     color: Rgb,
 }
 
@@ -65,11 +66,12 @@ impl Line {
 
         let rect = Rect::new(cell_x + size.padding_x, y + size.padding_y, size.cell_width, height);
 
-        Self { start: cell.into(), color: cell.fg, rect }
+        Self { start: cell.into(), end: cell.into(), color: cell.fg, rect }
     }
 
     fn update_end(&mut self, end: Point, size: &SizeInfo) {
         self.rect.width = (end.col + 1 - self.start.col).0 as f32 * size.cell_width;
+        self.end.col = end.col;
     }
 }
 
@@ -103,10 +105,9 @@ impl Rects {
 
             // Check if there's an active line
             if let Some(line) = self.inner.get_mut(flag).and_then(|lines| lines.last_mut()) {
-                let line_end = line.start.col.0 + (line.rect.width / size.cell_width) as usize;
                 if cell.line == line.start.line
                     && cell.fg == line.color
-                    && line_end == cell.column.0
+                    && cell.column == line.end.col + 1
                 {
                     // Update the length of the line
                     line.update_end(cell.into(), size);
@@ -128,7 +129,7 @@ impl Rects {
 
     // Add a rectangle
     pub fn push(&mut self, rect: Rect<f32>, color: Rgb) {
-        let line = Line { start: Point::default(), color, rect };
+        let line = Line { start: Point::default(), end: Point::default(), color, rect };
 
         // Flag `HIDDEN` for hashmap index is arbitrary
         match self.inner.get_mut(&Flags::HIDDEN) {
--- a/alacritty_terminal/src/renderer/rects.rs
+++ b/alacritty_terminal/src/renderer/rects.rs
@@ -26,25 +26,20 @@ pub struct Rect<T> {
     pub y: T,
     pub width: T,
     pub height: T,
+    pub color: Rgb
 }
 
 impl<T> Rect<T> {
-    pub fn new(x: T, y: T, width: T, height: T) -> Self {
-        Rect { x, y, width, height }
+    pub fn new(x: T, y: T, width: T, height: T, color: Rgb) -> Self {
+        Rect { x, y, width, height, color}
     }
 }
 
-struct Line {
-    rect: Rect<f32>,
-    start: Point,
-    end: Point,
-    color: Rgb,
-}
-
-impl Line {
-    /// Create a line that starts on the left of `cell` and is one cell wide
-    fn from_cell(cell: &RenderableCell, flag: Flags, metrics: &Metrics, size: &SizeInfo) -> Line {
-        let cell_x = cell.column.0 as f32 * size.cell_width;
+impl Rect<f32> {
+    fn from_line(line: &Line, flag: Flags, metrics: &Metrics, size: &SizeInfo, color: Rgb) -> Rect<f32> {
+        let start_x = line.start.col.0 as f32 * size.cell_width;
+        let end_x = (line.end.col.0 + 1) as f32 * size.cell_width;
+        let width = end_x - start_x;
 
         let (position, mut height) = match flag {
             Flags::UNDERLINE => (metrics.underline_position, metrics.underline_thickness),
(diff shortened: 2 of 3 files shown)


chrisduerr:
Can you provide an example for a case where this is broken? I've actually removed this check intentionally since it was originally just to compensate for whitespace which isn't rendered, but that shouldn't be necessary with the new approach.

kchibisov:
Sure,

current master:
Screenshot_Mon 05 Aug 2019 06:49:42 PM MSK

my PR:
Screenshot_Mon 05 Aug 2019 06:50:11 PM MSK

chrisduerr:
That's odd. With the current version, we should never update the end of a URL unless the cell is actually underlined, since we have to explicitly expand for every single cell.

I'll try to look into this in a few minutes, however that seems a bit odd to me.

kchibisov:
@chrisduerr You have two underlined cells with the same color on the same line, but there is an arbitrary text between them, thats the problem.

alacritty_terminal/src/renderer/rects.rs
```diff
@@ -103,7 +103,11 @@ impl Rects {
 
             // Check if there's an active line
             if let Some(line) = self.inner.get_mut(flag).and_then(|lines| lines.last_mut()) {
-                if cell.line == line.start.line && cell.fg == line.color {
+                let line_end = line.start.col.0 + (line.rect.width / size.cell_width) as usize;
```
nixpulvis: Would it be better to store this `end` on the `Line`? We already store the `start` in addition to the underlying `Rect`, so maybe this would be a natural addition.

I will say however, I'm not overly in love with storing a `Rect` with it's own positions independently of a `start` and `end` point, but I can see why it might be needed for performance reasons. It might be a good idea to add a note on the `start` (and potential `end`) fields explaining the constraint with the `rect.x` & `rect.width` values.

chrisduerr: > I will say however, I'm not overly in love with storing a Rect with it's own positions independently of a start and end point, but I can see why it might be needed for performance reasons.

My thought was similar, but if we store both start and end, we can just transform it to rects in the `rects()` method. There's no performance benefit to creating the rects once versus creating them every time `rects()` is called, since the method is only called once per draw call.

Generally I'm not a super huge fan of the current code, since it seems vulnerable to slight mathematical errors. A cast to `usize` could easily reduce the line length by a first cell just because it's been off by like `.00000001`.

I haven't had a ton of time yet today to explore alternatives, but storing the `end` and transforming in the `rects()` call was the best thing I could come up with at this point.

kchibisov: We're storing rectangles other than underline/strikeout in our `HashMap<Flags, Vec<Lines>>`, so creating such rectangles will become less  flexible (e.g. `push` method on `Rects`).

However I have enough time today to implement all of these stuff, so I'll send 2 commits with just adding `end: Point` and with refactoring you've suggested here.

nixpulvis: Yea, the `push` method is kinda a sore point for this code at the moment. For the purposes of this fix we don't need to go crazy refactoring things, but at some point we may want to split up the line functions from the more general rect functions in a way that addresses our concerns @chrisduerr.

chrisduerr: > Yea, the push method is kinda a sore point for this code at the moment. But for the purposes of this fix we don't need to go crazy refactoring things. At some point we may want to split up the line functions from the more general rect functions in a way that addresses our concerns @chrisduerr.

Well we've already refactored things a lot, so I'm not too concerned personally.

However I do question if there would be much benefit to separating them more.

However, we never need to call `update_lines` after underline and strikeout are finished, so I'd be perfectly fine with moving `update_lines` on `Line` maybe, then having the `rects` method take `rects(self)` and return an owned rects array where we can just push new rects in.

I think this should be a minor change which could also separate concerns a bit better? Curious about your opinions on this @kchibisov / @nixpulvis.

kchibisov:
This is just how it'll look like with simple storing things in `end: Point`. Major refactoring are not implemented yet.

alacritty_terminal/src/renderer/rects.rs
kchibisov: > I think this should be a minor change which could also separate concerns a bit better? Curious about your opinions on this @kchibisov / @nixpulvis.

I'll try to split things more, I just don't want to touch render with these refactorings.

chrisduerr: > I'll try to split things more, I just don't want to touch renderer with these refactorings.

Completely agree with that. At the end it should spit out a `Vec<Rects>` and we should be good. Though some minor changes in `display.rs` should be fine.

I think making the underline/strikeout solely handled by the `Line` struct which is then converted to a `Vec<Rect>` is the way to go. With Rust's ownership semantics that should provide a very nice interface when `rects` moves `self`.

kchibisov:
Okay, I've tried to split things here.

alacritty_terminal/src/display.rs
```diff
@@ -521,14 +521,16 @@ impl Display {
                     // Iterate over all non-empty cells in the grid
                     for cell in grid_cells {
                         // Update underline/strikeout
-                        rects.update_lines(&cell, &size_info, &metrics);
+                        lines.update_lines(&cell);
```
chrisduerr: `lines.update` might be better for naming.

alacritty_terminal/src/display.rs
```diff
@@ -521,14 +521,16 @@ impl Display {
...
 
                         // Draw the cell
                         api.render_cell(cell, glyph_cache);
                     }
                 });
             }
 
+            let mut rects = Rects::from(lines.into_rects(&metrics, &size_info));
```
chrisduerr: I think `rects` should just be `Vec<Rect>` and `Line::into_rects` should return `Vec<Rect>`. I think that should work.

I just saw that `Rects` obviously has colors too, but I don't think we use `Rect` anywhere else, right? So we could probably just add the color to `Rect` instead?

alacritty_terminal/src/renderer/rects.rs
```diff
@@ -35,69 +35,66 @@ impl<T> Rect<T> {
...
+/// Lines for underline and strikeout.
+#[derive(Default)]
+pub struct Lines {
+    inner: HashMap<Flags, Vec<Line>>,
+}
 
-    fn update_end(&mut self, end: Point, size: &SizeInfo) {
-        self.rect.width = (end.col + 1 - self.start.col).0 as f32 * size.cell_width;
-        self.end.col = end.col;
+fn create_rect(
```
chrisduerr: This should probably be moved into `Line`, since it's exclusive to that. It even takes in a `&Line`, so that could just be `&self`.

alacritty_terminal/src/renderer/rects.rs
```diff
@@ -35,69 +35,66 @@ impl<T> Rect<T> {
...
-            .collect()
+    pub fn into_rects(self, metrics: &Metrics, size: &SizeInfo) -> Vec<(Rect<f32>, Rgb)> {
+        let mut rects = Vec::with_capacity(self.inner.capacity());
+        for (flag, lines) in self.inner {
+            for line in lines {
+                rects.push(create_rect(&line, flag, &metrics, &size, line.color));
+            }
+        }
+        rects
     }
```
chrisduerr: Can this not be done with an iterator? That would save us from having to first define a vector and then fill it bit by bit.

alacritty_terminal/src/renderer/rects.rs
```diff
@@ -110,33 +107,52 @@ impl Rects {
...
+}
+
+/// Generall purpose rects.
+#[derive(Default)]
+pub struct Rects {
+    inner: Vec<(Rect<f32>, Rgb)>,
+}
+
+impl Rects {
+    pub fn new() -> Self {
```
chrisduerr: Neither `new` or `default` are used anymore, so I think we can remove the derive and the new impl.

alacritty_terminal/src/renderer/rects.rs
```diff
@@ -110,33 +107,52 @@ impl Rects {
...
+        Self::default()
+    }
+
+    /// Convert the stored rects to rectangles for the renderer.
+    pub fn rects(&self) -> &Vec<(Rect<f32>, Rgb)> {
```
chrisduerr: Instead of using a getter, I'd usually recommend just making the field `pub`. However this shouldn't be relevant anymore if we decide to add color to `Rect`.

alacritty_terminal/src/renderer/rects.rs
```diff
@@ -110,33 +107,52 @@ impl Rects {
...
+        &self.inner
+    }
+
+    /// Add rectangles
+    pub fn append(&mut self, other: &mut Vec<(Rect<f32>, Rgb)>) {
```
chrisduerr: I don't think this is ever called?

kchibisov: I've forgotten to delete this one.

alacritty_terminal/src/renderer/rects.rs
kchibisov: Ok, just wanted to save the thing to easy create Rects.

alacritty_terminal/src/renderer/rects.rs
kchibisov: I've tried to create a working solution here to expose structure. I'll try to create an iterator here, just It takes me more time, than such for loops.

alacritty_terminal/src/renderer/rects.rs
chrisduerr: Makes sense, but I don't think we need this in our current code and adding it should be simple enough once it actually is required.

alacritty_terminal/src/renderer/rects.rs
kchibisov: Kind of, just placed it here for now, since wasn't sure.

alacritty_terminal/src/display.rs
kchibisov: So, we're changing renderer? I'm fine with it though.

alacritty_terminal/src/renderer/rects.rs
chrisduerr: I think `self.inner.drain(..).map(|line| create_rect(...))..collect()` should work.

alacritty_terminal/src/display.rs
kchibisov: Sure, just mechanically renamed this.

alacritty_terminal/src/renderer/rects.rs
chrisduerr: Biggest issue here is probably the naming. I think `impl From<Line> for Rect` might do the job?

kchibisov: >  I think impl From<Line> for Rect might do the job?

That seems better.

alacritty_terminal/src/display.rs
kchibisov: Oh, and in this case we don't actually need `Rects` any more.

kchibisov: I've checked the use of `Rect` struct and we can't add color to it, without reworking all the stuff.

alacritty_terminal/src/renderer/rects.rs
kchibisov: Unfortunately that shouldn't work, since there is a restriction of only 1 parameter in from. However we can add custom method `from_line` to `Rect`.

alacritty_terminal/src/renderer/rects.rs
nixpulvis: I don't think I agree with this. Defining a non `mut` getter with a private field allows the library developer to be sure that that field is never mutated from the outside, whereas a `pub` field can be mutated if it's binding is `mut`. A somewhat small thing, but I find it useful.

alacritty_terminal/src/renderer/rects.rs
kchibisov: After some time I give up on this one, since things were more complicated with iterators here, since you have 2 loops and HashMap + Vec.

alacritty_terminal/src/display.rs
chrisduerr: > I've checked the use of Rect struct and we can't add color to it, without reworking all the stuff.

What do you mean by `all the stuff`, that sounds quite dramatic. As far as I can tell, wed just have to merge the function parameters in the renderer everywhere where it takes `rect` and `color` and have it both in one, right?

I don't think there's any instance of `Rect` without `color`, is there? So that kinda sounds like it would almost be an improvement anyways.

My main concern is that `Rect` sounds like a very generic name and we couldn't use it anymore if we ever wanted to have just a normal rectangle without any colors not used in the renderer, but I don't think that's currently the case anywhere?

alacritty_terminal/src/renderer/rects.rs
chrisduerr: I think `from_line` makes sense on `Rect`, yeah.

alacritty_terminal/src/renderer/rects.rs
chrisduerr: You can create a vec from a hashmap using flatten, just take a look at the previous code.

`self.inner.drain(..).map(|(_, lines)| lines).flatten()` should create an iterator over `Line`, then from there my previous comment with the `.map(|line| create_rect(...)).collect()` should be applicable.

alacritty_terminal/src/display.rs
kchibisov: > I don't think there's any instance of Rect without color, is there?

I've accidentally found another `Rect` struct, which isn't related to `Rect` in renderer. So, I'll change stuff in a moment.

alacritty_terminal/src/renderer/rects.rs
chrisduerr: Is there any good reason why we would want to prevent that here though? That's precisely why we do have `mut` to mark things as mutable or immutable.

Currently this just adds five lines of code which would be completely identical to just accessing it directly.

kchibisov: FYI, this code will be deleted in a moment.

alacritty_terminal/src/display.rs
chrisduerr: > I've accidentally found another Rect struct, which isn't related to Rect in renderer. So, I'll change stuff in a moment.

There is? I don't see any other `Rect` struct in the latest master.

kchibisov: I was grepping entire workspace, so there is one in `font/src/darwin/mod.rs`, which is unrelated here.

alacritty_terminal/src/renderer/rects.rs
kchibisov: Yeah, but in my case I need something like Vec<(Flags, Line)> at some point, since I need Flag, which isn't the case in your previous code.

kchibisov:
The all issues we're addressed here, beside rewriting `into_rects` in iterators, which I still think will be more complex, than it's right now.

kchibisov:
As one more thing, that can be addressed here is the naming of `Rect<T>` struct, which can be changed to `RenderableRect`/`RendererRect` and specialized to use f32 in it or the same name, but with `Rect<f32>` inside, which I think adds more complexity here.

alacritty_terminal/src/renderer/rects.rs
chrisduerr: `self.inner.drain(..).flatten()` will give you an iterator over `(Flags, Line)`, that should be even easier.

kchibisov: `self.inner.drain().flatten()` wouldn't compile since you'll need to implement all that stuff `the trait `std::iter::Iterator` is not implemented for `(term::cell::Flags, std::vec::Vec<renderer::rects::Line>)``

kchibisov: I've implemented this one in iterators though

alacritty_terminal/src/renderer/rects.rs
```diff
@@ -26,24 +26,20 @@ pub struct Rect<T> {
...
-    start: Point,
-    color: Rgb,
-}
-
-impl Line {
-    /// Create a line that starts on the left of `cell` and is one cell wide
-    fn from_cell(cell: &RenderableCell, flag: Flags, metrics: &Metrics, size: &SizeInfo) -> Line {
-        let cell_x = cell.column.0 as f32 * size.cell_width;
+impl Rect<f32> {
+    fn from_line(line: &Line, flag: Flags, metrics: &Metrics, size: &SizeInfo, color: Rgb) -> Rect<f32> {
```
chrisduerr: I think moving this to `Line` and calling it as `into_rect` is better since we can then use `&self`.

kchibisov:
> Why do we need a Rect<f32> inside?

I've just provided an alternative, in case we want to hold `Rect<T>` for some reason. I don't like the idea of keeping Rect<f32>.

Anyway, everything should be fine now.

chrisduerr:
I think the last "issue" I have now is `Line` and `Lines`. I feel like these are named exceptionally poorly (which I guess is my fault), especially since we already have a `Line` struct in `index.rs`.

Do you have any ideas on how to improve this? Name suggestions would be very welcome here, especially now that `Lines` is exposed publicly from the module.

kchibisov:
@chrisduerr `RenderLine`/`RenderLines`?

chrisduerr:
> @chrisduerr RenderLine/RenderLines?

I think that's better already. Do you have any ideas of how to solve that we have just `line` and `lines` structs? It seems a bit odd to have an extra struct just for collecting lines but I don't see a way around this because the functions implemented for it are necessary. Just adding an `s` to the name feels kinda odd to me.

kchibisov:
I don't find this odd, since `Line` is private (and likely will never go public) and just a helper for `Lines` to simplify things a bit. However we can make `Line` public, remove `Lines` entirely and just pass a `HashMap<Flags, Vec<Line>>` in and out to a simple functions in `rects.rs`, I don't like this approach much though.

chrisduerr:
> I don't find this odd, since Line is private (and likely will never go public)

Yeah, this is a very good point.

> However we can make Line public, remove Lines entirely and just pass a HashMap<Flags, Vec<Line>> in and out to a simple functions in rects.rs, I don't like this approach much though.

I don't like it much either, I think having the functions defined on the `Lines` struct helps a lot with a nice and structured API.

So it's just the `RenderLine(s)` change then.

kchibisov:
> So it's just the RenderLine(s) change then.

done

Issue #2651, opened 2019-07-16
Blurred alacritty.ico in Large Icon Task Switcher on Linux


xuzhao9:
Hello! I am an Alacritty fan on KDE and using the Large Icon task switcher. Everything works great except that Alacritty icon is blurred icon under this scheme. See the screenshot here: https://ibb.co/z4S7wwV
After some exploration, I found replacing the extra/windows/alacritty.ico fixes this problem. I can provide the new ico file at: 
https://user.fm/files/v2-86512d5494230eebac9323053bfe8dba/alacritty.ico

I can also submit a pull request, if you want.

Which operating system does the issue occur on?
openSUSE Tumbleweed 20190708

If on linux, are you using X11 or Wayland?
X11

kchibisov:
@xuzhao9 How does `KDE` pickup icons for task switche? Does it uses `NET_WM_ICON` or it just query them from `/usr/share/pixmaps` ? I guess the first one, since you've mentioned `extra/windows/alacritty.ico`, which is used for `NET_WM_ICON`, but I remember that it uses `/usr/share/pixmaps` over `NET_WM_ICON` in some places, so just wondering what you've changed here.

xuzhao9:
@kchibisov Yes, it uses the NET_WM_ICON. /usr/share/pixmaps/Alacritty.svg is the vector one therefore should not be blurred, however it is not being used.

chrisduerr:
Since the .ico is primarily for the standard Windows icon, it's probably a good idea to check how it looks on Windows. As far as I remember, there were some good reasons for building the ico the way it has been built, however it's possible that just increasing the resolution would still work fine on Windows.

kchibisov:
@xuzhao9 Yes, seems like that, unfortunately we can't use `svg` for `NET_WM_ICON`, since our dependency doesn't support loading them (`KDE`/`Qt` had some problems with `Alacritty.svg` though).

kchibisov:
@chrisduerr As far as I've [understood](https://docs.microsoft.com/en-us/windows/win32/uxguide/vis-icons) we can use icon with `256x256` for windows (The maximum size for `ico`). It will downscale it to appropriate size.

@xuzhao9 How did you build this icon though? Upscale?

chrisduerr:
[...]

PR #2707, opened 2019-08-07
Update the README logo to be crisper

nixpulvis:
Following the work in #2651, I've lifted a better looking icon for use in our README. I've also added the other `.ico` file sizes provided by @xuzhao9 until we have a way to generate these ourselves.


kchibisov:
@nixpulvis What display did you use to test our icon in `README.md`?

Here are two icons try to guess which one is from the current readme.
Screenshot_Wed 07 Aug 2019 09:57:46 PM MSK

Screenshot_Wed 07 Aug 2019 09:58:11 PM MSK

chrisduerr:
Isn't the point of SVGs kinda that they scale properly to appropriate sizes? It seems like using an image in a scenario where an SVG is supported properly is not a great choice.

I am however not 100% certain how svgs behave on the `img` tag as source. There might be a better way to embed it.

xuzhao9:
I generated the .ico file from svg with ImageMagick. It loses some details comparing with the browser version. I will try to generate a better version and resubmit to the original issue post.

kchibisov:
> Isn't the point of SVGs kinda that they scale properly to appropriate sizes? It seems like using an image in a scenario where an SVG is supported properly is not a great choice.

I'm absolutely agree on this, we should use svg instead of ico for README. However our README.md icon is fine only when it's big, so on mobile phone it can be blurry.

For e.g. this is how it looks like on my phone. You can see sharp corner/borders but the icon is blurry because of the number of scanlines.

nixpulvis:
@kchibisov interesting I hadn't noticed the edges... This is unfortunate, because I find the scan lines to look better in this new version (the issue I was trying to fix) and not the rest of the image is a mess :frowning:

@chrisduerr yes ideally vectors scale but something is wrong with the scanlines.

kchibisov:
> @kchibisov interesting I hadn't noticed the edges... 

The problem is not the edges, look at the light around blue rhombus, it's not smooth. Same for the red-orange `A`.

@nixpulvis FYI on my 4k display scanlines from our README are super clear, but it's 4k ;- ).

chrisduerr:
> This is unfortunate, because I find the scan lines to look better in this new version

I actually do not agree with this. In the SVG version the gradients look much more uniform and gradual, the image looks a lot more spotty and noisy.

Really the only problem I could see is on mobile as @kchibisov has said. I'm not sure if it's possible to do media queries in github markdown though and I think if I had to choose, I'd accept a bit of a worse quality on mobile since I'd expect most visitors to come from desktop.

kchibisov:
> and I think if I had to choose, I'd accept a bit of a worse quality on mobile since I'd expect most visitors to come from desktop.

Same here though, since github README is mostly for developers/advanced users to give a short meaningful information and not for advertising an app (I mean we can advertise, but that's not the main purpose of it in my opinion).

nixpulvis:
@chrisduerr I was most offended on mobile, that's probably what inspired this. I guess I agree with you now that I hear it out loud about the scanlines. That said, I think it's poor design to have the "A" of Alacritty look blurry by default. Scanlines are a cool effect, but it's hard to pull off well it seems.

That said, I'll either close this PR, or find a way to make it look the same on desktop, and better on mobile. Maybe @xuzhao9 can think of a clever way to sharpen up the scanlines without making it all look worse like I've done here.

nixpulvis:
@chrisduerr @kchibisov, FWIW not that I'm a prefect example of a user, but I use GitHub on mobile pretty heavily.

kchibisov:
However the icon that we can update is `alacritty/extra/windows/alacritty.ico`, which is low res (relative to the format `ico`). So, for this purposes I recommend to use just one big (256x256) `alacritty/logo/alacritty-term.svg`

kchibisov:
There is a compromise solution to use `alacritty/extra/logo/alacritty-term.svg` for logo in README, but these will make scanlines version quite useless for a while. One more alternative is to use `png` (512x512 or something similar) for README which should be better than ico.

chrisduerr:
> There is a compromise solution to use alacritty/extra/logo/alacritty-term.svg for logo in README, but these will make scanlines version quite useless for a while.

Yeah, this is kinda my concern. I like seeing the scanline version, it gives the icon a bit more of a terminal emulator touch I think and it would be a shame to basically scrap it completely.

I've actually gotten out my phone to test it myself and I think it looks pretty nice on mobile too, dare I say better than desktop even? Might be interesting to see what OS you run @nixpulvis, I'd imagine iOS and Android might have some differences in SVG rendering, since there's only one browser on iOS (I use IceCat on Android).

kchibisov:
> I've actually gotten out my phone to test it myself and I think it looks pretty nice on mobile too, dare I say better than desktop even?

Thats interesting, I've just tested on android phone (with absolutely random browser) and the logo was as sharp and clean as on my 4k screen. So it seems like a browser problem, since I'm using ios on my phone.

nixpulvis:
> since there's only one browser on iOS

Oh how wrong you are. I'm one of *very* very small cult that uses Firefox on iOS. Even though there's no way to change the default browser, and even though no developers try to make things work, and even though the gmail app will gladly open attachments in chrome or safari, still I use Firefox Mobile. Why, well... you gotta walk the talk... and I'm a hopeless masochist. \</rant>

kchibisov:
> Oh how wrong you are.

He is not wrong. All ios browsers are using WebKit under the hood, which means they are all the same. Firefox on ios (which I'm also using) is built with WebKit and not gecko.

chrisduerr:
So that seems to be it then? Webkit just doesn't render the SVGs nicely? I mean it doesn't change a ton since we still need to consider if we want to just ignore iOS users or not, but at least it shows that the issue is limited to just a small number of configurations.

I think playing around with a better way to embed the SVGs might be worth it, but I don't think it would be worth it to make it look worse on everything but iOS just so it can look okay on iOS, when it's the fault of webkit and not our SVG.

kchibisov:
@chrisduerr It seems like a problem in WebKit svg rendering, there are a bunch of bug reports on their end about blurry svg and so on. (You can search their bug tracker for such stuff, I just don't want to link a lot here)

nixpulvis:
@kchibisov good point, since we'd expect Firefox and Safari to look the same as it relates to this issue. I was just using this as a good time to push back on the Safari monopoly. Browser != browser engine, etc.

Anyway, I'm going to close this issue since it's a know issue in WebKit, and any discussion of what to do about other uses of our icon can happen there.

This has been a fun thread :stuck_out_tongue_winking_eye:

Issue #2574, opened 2019-06-22
Wayland clipboard intermittently broken


pschyska:
Which operating system does the issue occur on?
Fedora 30, on both alacritty 0.3.3 and alacritty master. I'm running in gnome-session-wayland-session 3.32.0.
If on linux, are you using X11 or Wayland?
Wayland. This issue doesn't occur in X11 at all.

I noticed that sometimes, neither the Selection nor Clipboard works at all. I can't fix it without restarting alacritty. I ran a test just running it from my shell from another alacritty instance many times, and it's 50-50 each trial if the clipboard is either completely broken, and never works, or always works.

I've added some debug statements:

```diff
diff --git a/alacritty_terminal/src/clipboard.rs b/alacritty_terminal/src/clipboard.rs
index 0e40cd4..85e3378 100644
--- a/alacritty_terminal/src/clipboard.rs
+++ b/alacritty_terminal/src/clipboard.rs
@@ -83,18 +83,21 @@ pub enum ClipboardType {

 impl Clipboard {
     pub fn store(&mut self, ty: ClipboardType, text: impl Into<String>) {
+        debug!("Clipboard#store ty={:?}", ty);
         let clipboard = match (ty, &mut self.selection) {
             (ClipboardType::Selection, Some(provider)) => provider,
             (ClipboardType::Selection, None) => return,
             _ => &mut self.clipboard,
         };
-
-        clipboard.set_contents(text.into()).unwrap_or_else(|err| {
+        let text = text.into();
+        debug!("Clipboard#store text={}", text);
+        clipboard.set_contents(text).unwrap_or_else(|err| {
             warn!("Unable to store text in clipboard: {}", err);
         });
     }

     pub fn load(&mut self, ty: ClipboardType) -> String {
+        debug!("Clipboard#load ty={:?}", ty);
         let clipboard = match (ty, &mut self.selection) {
             (ClipboardType::Selection, Some(provider)) => provider,
             _ => &mut self.clipboard,
@@ -105,7 +108,10 @@ impl Clipboard {
                 debug!("Unable to load text from clipboard: {}", err);
                 String::new()
             },
-            Ok(text) => text,
+            Ok(text) => {
+                debug!("Clipboard#load text={}", text);
+                text
+            },
         }
     }
 }
``` 

The output is as follows:

```
[2019-06-22 11:08] [DEBUG] Clipboard#store ty=Selection
[2019-06-22 11:08] [DEBUG] Clipboard#store text=alacritty
[...]

PR #2715, opened 2019-08-11
Fix wayland clipboard integration

kchibisov:
What was fixed here:
- [x] Implement `gtk_primary_selection_device_manager` ( `smithay-clipborad`)
- [X] Fix race? or whatever with normal clipboard on GNOME ( `smithay-clipboard` or `alacritty` )
- [ ] Clipboard dies over time on sway/probably GNOME (`smithay-clipboard`)

Initial attempt to fix clipboard on wayland.

Fixes: #2574

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Selecting trailing tab with semantic expansion
 - URL parser incorrectly handling Markdown URLs and angled brackets
 - Intermediate bytes of CSI sequences not checked
+- Wayland clipboard integration
 
 ## 0.3.3
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -360,7 +360,7 @@ dependencies = [
  "objc-foundation 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "objc_id 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "smithay-client-toolkit 0.6.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "smithay-clipboard 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "smithay-clipboard 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "wayland-client 0.23.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "x11-clipboard 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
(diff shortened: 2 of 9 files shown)


kchibisov:
I've added implementation of `gtk-primary-selection`, for now it's in my fork, but both clipboards are working fine for me on GNOME + wayland.

The last issue I have here, is that clipboard die, when tty switch occurs(not always dies though), however I don't think this remaining issue should block merging this PR, since we've already has some problems with tty switching.

trimental:
smithay-clipboard v0.3.4 has been released with your gtk fallback implementation, good work

kchibisov:
@trimental I've noticed, thx for your help.

>  The last issue I have here, is that clipboard dies, when tty switch occurs(not always dies though), however I don't think this remaining issue should block merging this PR, since we've already has some problems with tty switching.

@trimental , What do you think about this issue though? Asking here for now, probably move this question to Matrix.

nixpulvis:
@kchibisov FWIW, we are also pretty active in the #alacritty Freenode channel too.

kchibisov:
@nixpulvis I'm aware of it, but the last issue is likely with Smithay's stuff, rather then alacritty's, so asking there would make much more sense in my opinion (I'm about smithay chatroom on matrix).

P.s.
@nixpulvis, I'm too young for IRC stuff though, but I'll investigate in solving this problem ;- ). I've already tried that in the past, but I've wiped `weechat` during storage upgrades on my PC.

trimental:
@kchibisov mm I’m unsure about the tty switching, there are people on the smithay matrix channel with more general wayland knowledge then me that may have an idea, although I’ll hopefully be able to investigate over the next few days.

About the IRC, you can use matrix IRC bridges to connect without needing an IRC client, I think you need to connect to 	`#freenode_#alacritty:matrix.org`

kchibisov:
@chrisduerr I think this PR is ready for a review.

There is one more issue with wayland clipboard, but it wasn't the subject in #2574 and I guess only I noticed that issue. I'm switching to X11 sessions to test alacritty's cursor stuff, so that was the first time I've experienced some problems with clipboard, but it doesn't die over time as I've initially stated, since I had 10+ terminals running for more than 24 hours+ in a row and everything was fine with clipboards, unless I've decided to switch tty.

pschyska:
@kchibisov I've also finally got around testing your changes. Seems to work perfectly!

I ran alacritty about 20 times and couldn't make the clipboard misbehave. Additionally, primary selection worked -- in true Heisenbug fashion, the primary selection never worked for me as well anymore on the stable version :smile:, which makes more sense due to the interface change discussed earlier.
Either I tested incorrectly, or something in Gnome changed recently (I switched to I3 because of the clipboard issues).

Thanks a lot!

kchibisov:
@chrisduerr Looks fine. Just push "Fix clippy issues" and "my Fix" in a separate commits.

chrisduerr:
I've merged this into master manually.

PR #2717, opened 2019-08-12
Triple-select include wraplines

ike:
Hey there! First PR on Alacritty.

This enables triple-select to select all lines that have wrapped in either direction from the clicked point.

Screen Shot 2019-08-12 at 1 20 29 PM

I am unable to get one of the tests to work -- the `line_selection_works` test fails with the following error:

```
thread 'term::tests::line_selection_works' panicked at 'assertion failed: requested < self.len', alacritty_terminal/src/grid/storage.rs:195:9
```

Also, the contribution guide recommends running `cargo fmt`, but the formatter changes a couple dozen files. I did not commit these changes.


nixpulvis:
Found a small, what apears to be some kind of off-by-one error when testing this out.

alacritty-2717-1

kchibisov:
> Also, the contribution guide recommends running cargo fmt, but the formatter changes a couple dozen files. I did not commit these changes.

FYI you have to run `cargo fmt` using rust nightly toolchain for now.

chrisduerr:
> FYI you have to run cargo fmt using rust nightly toolchain for now.

Just to elaborate on this, it's not because rustfmt isn't available on the nightly toolchain only, but because some of its best features are restricted to nightly unfortunately.

So I don't expect this to change anytime soon, but it's intentionally not a hard CI failure since I'm always willing to do that myself if it causes too much hassle.

ike:
@nixpulvis I think I got the off-by-one error sorted in the latest commit.

nixpulvis:
@ike still seems to be a bit off, though definitely better than before. Let me see if I can explain what I'd like to see ideally.

In this first example, I have triple clicked to select a line that wraps across two lines. This PR properly selects both lines. :tada: However then upon dragging down to the line below the wrapped line it deselects the initially selected lines, and incorrectly selects the `Column(0)` cell alone.



Similarly it seems that it waits to select a new line until the mouse reaches the next line after, whereas I'd expect the line to be selected as soon as the mouse makes it about halfway into the line.



Hopefully this makes sense.

ike:
@nixpulvis yeah, that makes total sense. Are there any examples of that half-selected line? Not sure how I would approach it, but I'll experiment a bit tonight and see what I can find.

nixpulvis:
> Are there any examples of that half-selected line?

@ike I'm not sure I understand what you're asking, sorry.

ike:
Basically -- how to select the line I am over right now, halfway, instead of waiting until the next line is the one I'm over.

nixpulvis:
@ike Oh, you can look at how it works on master, or how the column-wise logic works. It should be roughly the same.

kchibisov:
> I am unable to get one of the tests to work -- the line_selection_works test fails

@ike Just launch a terminal and triple click on the top line or bottom one(without additional scrolling), it'll fail on the same assertion.

chrisduerr:
Thanks @kchibisov and @nixpulvis for testing this. Sounds like the main issues have been identified already. I'll give some comments on the code once the remaining issues are resolved.

Please let me know if you have any outstanding questions.

ike:
Alright -- I think I have resolved the bugs that @nixpulvis identified, and I got tests passing. Ready for another review!

🖖  ike

alacritty_terminal/src/term/mod.rs
```diff
@@ -2252,6 +2270,10 @@ mod tests {
         }
     }
 
+    /// Test selection across lines
+    ///
+    ///
+    /// 1.  [a  ][a" ][a  ][a  ][a" ]
```
ike: this might be an incorrect picture of the line that's being tested.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2275,7 +2297,7 @@ mod tests {
         mem::swap(&mut term.grid, &mut grid);
 
         *term.selection_mut() = Some(Selection::lines(Point { line: 0, col: Column(3) }));
-        assert_eq!(term.selection_to_string(), Some(String::from("\"aa\"a\n")));
+        assert_eq!(term.selection_to_string(), Some(String::from("\n\"aa\"a\n")));
```
chrisduerr: This test still applies after this PR, this indicates an error with this PR.

alacritty_terminal/src/grid/storage.rs
```diff
@@ -192,7 +192,7 @@ impl<T> Storage<T> {
     #[inline]
     /// Compute actual index in underlying storage given the requested index.
     fn compute_index(&self, requested: usize) -> usize {
-        debug_assert!(requested < self.len);
+        debug_assert!(requested <= self.len);
```
chrisduerr: This would allow index out of bounds. If this is triggered, the code triggering it should be fixed.

alacritty_terminal/src/term/mod.rs
chrisduerr: Not quite sure what this "picture" is supposed to mean.

alacritty_terminal/src/term/mod.rs
```diff
@@ -58,6 +58,8 @@ pub trait Search {
     fn semantic_search_left(&self, _: Point<usize>) -> Point<usize>;
     /// Find the nearest semantic boundary _to the point_ of provided point.
     fn semantic_search_right(&self, _: Point<usize>) -> Point<usize>;
+    /// Find the extent of wrap lines to either side of provided point.
+    fn search_wrapline(&self, _: Point<usize>) -> Span;
```
chrisduerr: To match the semantic search, I think it's better to use `line_search_left/right` here. That would also remove the `Span` as return type which is a bit problematic here since it encodes more information than just start/end points.

alacritty_terminal/src/selection.rs
```diff
@@ -525,8 +544,8 @@ mod test {
         selection.rotate(-3);
 
         assert_eq!(selection.to_span(&term(5, 10)).unwrap(), Span {
-            start: Point::new(0, Column(4)),
-            end: Point::new(2, Column(0)),
+            start: Point::new(0, Column(5)),
+            end: Point::new(3, Column(5)),
```
chrisduerr: This seems to indicate an error in this PR.

alacritty_terminal/src/selection.rs
```diff
@@ -272,12 +272,31 @@ impl Selection {
...
+            start.line += 1;
+        } else if start_span.start.line > end_span.end.line {
+            start = start_span.start.into();
+            end = end_span.end.into();
+            end.line -= 1;
+        } else {
+            start = start_span.end.into();
+            end = end_span.start.into();
+            end.line += 1;
+        }
```
chrisduerr: All of this code shouldn't be necessary if the search left/right methods would just return the point after expanding it all the way to the left/right. The same way the semantic search functions.

ike:
Alright, I split the search method into two left/right searches and modified the selection method accordingly. Selection is working as expected. Does this look like the right direction?

Now there are _two_ tests failing. What's the best way to debug these tests?

chrisduerr:
> Now there are two tests failing. What's the best way to debug these tests?

I mean the tests should be pretty self descriptive in what they're trying to do. Usually looking at the code they're calling and the code you've changed is a good way to start.

ike:
Okay, I found another test that I think I had previously modified -- so we're down to just one that's failing. When I comment out the following lines, the test passes. Gonna keep digging.
 
https://github.com/jwilm/alacritty/blob/2bb97fe655c87244ace6b0050474db823613ebbf/alacritty_terminal/src/selection.rs#L277-L284

chrisduerr:
@ike It's probably a good idea to check that no wrapping is going on in the test. As long as that's the case, it shouldn't require any changes to the test itself.

chrisduerr:
Thanks for your work on this, fixing the test was trivial, so I've created a PR for that here: https://github.com/alacritty/alacritty/pull/3207.

ike:
Awesome, sorry I went AWOL -- got super busy with regular work. Still using Alacritty daily and loving it.

chrisduerr:
> Awesome, sorry I went AWOL -- got super busy with regular work. Still using Alacritty daily and loving it.

No worries, thanks for getting this started and doing most of the work!

PR #2719, opened 2019-08-13
Reset the Mouse Cursor While Selecting

nixpulvis:
This change makes the mouse cursor reset to it's default state while a selection is in progress. A click to clear the selection doesn't trigger a URL action, and so the mouse hand style isn't applied until the selection is cleared.

Before:
before

After:
after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Wayland clipboard integration
 - Use text mouse cursor when mouse mode is temporarily disabled with shift
 - Wayland primary selection clipboard not storing text when selection is stopped outside of the window
+- Block URL highlight while a selection is active
 
 ## 0.3.3
 
--- a/alacritty_terminal/src/input.rs
+++ b/alacritty_terminal/src/input.rs
@@ -449,6 +449,8 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         if self.mouse_config.url.mods().relaxed_eq(mods)
             && (!self.ctx.terminal().mode().intersects(mouse_mode) || mods.shift)
             && self.mouse_config.url.launcher.is_some()
+            && self.ctx.selection_is_empty()
+            && self.ctx.mouse().left_button_state != ElementState::Pressed
         {
             let buffer_point = self.ctx.terminal().visible_to_buffer(point);
             if let Some(url) =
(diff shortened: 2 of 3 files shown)


alacritty_terminal/src/input.rs
```diff
@@ -455,7 +455,11 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
             if let Some(url) =
                 self.ctx.terminal().urls().drain(..).find(|url| url.contains(buffer_point))
             {
-                return MouseState::Url(url);
+                if self.ctx.selection_is_empty() &&
+                   self.ctx.mouse().left_button_state != ElementState::Pressed
```
chrisduerr: Based on my testing, this shouldn't be necessary. I've asked about this before, but could you go into more detail why this is needed?

alacritty_terminal/src/input.rs
```diff
@@ -455,7 +455,11 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
             if let Some(url) =
                 self.ctx.terminal().urls().drain(..).find(|url| url.contains(buffer_point))
             {
-                return MouseState::Url(url);
+                if self.ctx.selection_is_empty() &&
```
chrisduerr: I think we can move this check to the other `if` block, the one before the `if let` binding. We don't need access to the `url` so it should be easier to check it early in one block.

alacritty_terminal/src/input.rs
```diff
@@ -692,6 +696,17 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+            let mouse_state = self.mouse_state(point);
+            self.update_mouse_cursor(mouse_state);
+            match mouse_state {
+                MouseState::Url(url) => {
+                    let url_bounds = url.linear_bounds(self.ctx.terminal());
+                    self.ctx.terminal_mut().set_url_highlight(url_bounds);
+                },
+                _ => {
+                    self.ctx.terminal_mut().reset_url_highlight();
+                }
```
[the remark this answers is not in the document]
chrisduerr: Why would we need this? This code is just to make sure that when the selection is cleared by clicking, that URLs are shown again, right? We're never concerned about removing URLs when the selection is cleared (since it shouldn't exist in the first place).

So this can just be transformed into an `if let`.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Intermediate bytes of CSI sequences not checked
 - Wayland clipboard integration
 - Use text mouse cursor when mouse mode is temporarily disabled with shift
+- Remove URL styles when selecting text
```
chrisduerr: I wouldn't know what `URL styles` are and I'd assume I'm not alone here. I think we can describe better what we're doing here.

Not much better, but maybe this would work:

> - Block URL highlight while a selection is active

Open for other suggestions though. I just don't think `Remove URL styles` says a lot.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Intermediate bytes of CSI sequences not checked
 - Wayland clipboard integration
 - Use text mouse cursor when mouse mode is temporarily disabled with shift
+- Block URL highlight while a selection is active.
```
chrisduerr: No punctuation at the end here.

alacritty_terminal/src/input.rs
```diff
@@ -692,6 +694,12 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
             self.mouse_report(code, ElementState::Released, modifiers);
             return;
         } else if let (Some(point), true) = (point, button == MouseButton::Left) {
+            let mouse_state = self.mouse_state(point);
+            self.update_mouse_cursor(mouse_state);
+            if let MouseState::Url(url) = mouse_state {
+                let url_bounds = url.linear_bounds(self.ctx.terminal());
+                self.ctx.terminal_mut().set_url_highlight(url_bounds);
+            }
```
chrisduerr: Why is this required here? The mouse highlight is accurate for me after clicking on URLs even without this.

chrisduerr: Ah, got it!

nixpulvis:
Fixed the rebase, and squashed down to a single commit.

alacritty_terminal/src/input.rs
chrisduerr: This makes me think though, couldn't we move the `self.launch_url()` call into this then and remove the check for modifiers at the start of the `launch_url` method?

nixpulvis:
Note to self, from @chrisduerr: https://paste.rs/YUx.diff

nixpulvis:
I've uncovered a bug while testing this, however this is actually a bug on master, so I don't think it should block merging this.

block-selection-url-bug-2019-09-07_00 03 52

To reproduce on master, create a 0 width N height block selection, and then attempt to click a URL. I believe a proper fix for this is to disallow 0 width block selections, similarly to how other selections disappear.

chrisduerr:
Very good catch @nixpulvis. I'll see if I can find a simple patch, but I do agree that it should be unrelated to this PR.

alacritty_terminal/src/input.rs
```diff
@@ -691,34 +693,34 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
         }
 
-        let point = self.ctx.terminal().visible_to_buffer(point);
-        let url = self.ctx.terminal().urls().drain(..).find(|url| url.contains(point))?;
-        let text = self.ctx.terminal().url_to_string(&url);
-
-        let launcher = self.mouse_config.url.launcher.as_ref()?;
-        let mut args = launcher.args().to_vec();
-        args.push(text);
+        let text = self.ctx.terminal().url_to_string(url);
```
chrisduerr: This is useless if we don't have url launcher set. So instead we can do it later inside the `if let`.

PR #2723, opened 2019-08-16
Update euclid to 0.20

Eijebong:


chrisduerr:
It seems like the `parking_lot` and `url` deps of `alacritty_terminal` are also outdated. Would you be willing to bump these too? I can't imagine that they'd require significant changes.

Eijebong:
There you go :p

chrisduerr:
Looks like the euclid bump broke macOS, it seems like `Rect` has an additional parameter now.

I don't quite understand what the `U` parameter is supposed to be used for, but I think just adding `()` should work. Can you fix this even though you probably don't have a macOS system for testing? Feel free to abuse CI.

lespea:
> All types are generic over the scalar type of their component (f32, i32, etc.), and tagged with **a generic Unit parameter which is useful to prevent mixing values from different spaces**. For example it should not be legal to translate a screen-space position by a world-space vector and this can be expressed using the generic Unit parameter.

> This unit system is not mandatory and all * structures have an alias with the default unit: UnknownUnit. for example Point2D<T> is equivalent to Point2D<T, UnknownUnit>. Client code typically creates a set of aliases for each type and doesn't need to deal with the specifics of typed units further. For example:

Issue #2745, opened 2019-08-24
Italic doesn’t work on Windows


chris-morgan:
Started in #894, but I now believe that this is a different issue.

On Windows 10 version 1903 (build 18362.295), alacritty master, `CSI [ 3 m` (italic) and `CSI [ 23 m` (cancel italic) just don’t work. You echo it like this:

```sh
echo -e 'Normal, \x1b[1mbold\x1b[22m, \x1b[3mitalic\x1b[23m, \x1b[1;3mbold italic\x1b[22;23m'
```

… but it disappears before `<alacritty_terminal::ansi::Performer as vte::Perform>::csi_dispatch`.

Reproducible with both WinPTY and ConPTY backends.

I don’t know how WinPTY and ConPTY are structured, but I suspect that they’re meddling, and filtering out things that *they* don’t know about. When I log from `csi_dispatch` when `action == 'm'`, I observe substantial differences between WinPTY and ConPTY—after filtering some things out, WinPTY mostly passes what you asked for, while ConPTY seems to split combined commands up and rewrite some things like bold to add bright text, and turn unbold (22) into reset () if possible—sure looks like there’s something vaguely intelligent in there, which is super annoying for Alacritty, which would much prefer to implement everything itself.

Can anything be done about this? Some ideas:

1. Disable or bypass the bad WinPTY/ConPTY intelligence;
2. Ask Microsoft to allow same;
3. Ask Microsoft to at least pass unknown CSI through;
4. Ask Microsoft to implement `3` and `23` at least enough to pass them through, even if they don’t implement it visually yet. (This is probably a bigger job than I’d like, since the unbold → reset transformation suggests they’re keeping track of active attributes.)

Or I might be completely wrong, and something else is to blame.

I suspect *something* to be possible, since ConEmu [includes this line in its README](https://github.com/Maximus5/ConEmu#description):

> - using normal/bold/italic fonts for different parts of console simultaneously;

But I haven’t actually tried it.

*(Now I have; it looks like it actually achieves that by sacrificing a background colour, reinterpreting specific background colours as bold or italic. Yuck! Maybe there is no current solution, then.)*

chrisduerr:
If Windows PTYs do not send the escape sequence at all, I don't think there's anything we can do.

It might be interesting to see if the new Windows Terminal can handle it though and if it does, how it's doing it.

I'm strongly against introducing some weird hacks that break support for other things to support this, in this case this bug should be propagated up to Microsoft to resolve this properly.

[...]

PR #2746, opened 2019-08-24
Support bold italic font

chris-morgan:
It’s perfectly reasonable to combine bold and italic.

I also normalised the ordering of the four variants, which was inconsistent in some places (including also the doc comment not matching the code in one place), to this: regular, bold, italic, bold italic.

This may functionally be a slight regression for fonts that don’t have a bold italic variant, because I suppose it’ll pick some other variant, I know not which, rather than the bold font. If we care about that enough, we can fiddle it in config::font to fall back to bold rather than normal, and in renderer to fall back to bold rather than I know not what. I didn’t do that due to the (very) mild complexity involved, because of the rareness I expect of manually setting a bold font that differs materially from the normal font, and because I naively expect the renderer fallback to do a sane thing out of the box.

----

Untested because I’m on Windows and italics aren’t working there (#2745). It builds and runs, but I can’t definitely say further than that.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -145,6 +145,17 @@ font:
     # The `style` can be specified to pick a specific face.
     #style: Italic
 
+  # Bold italic font face
+  #bold_italic:
+    # Font family
+    #
+    # If the bold italic family is not specified, it will fall back to the
+    # value specified for the normal font.
+    #family: monospace
+
+    # The `style` can be specified to pick a specific face.
+    #style: Bold Italic
+
   # Point size
   size: 11.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Config option `window.gtk_theme_variant` to set GTK theme variant
 - Completions for `--class` and `-t` (short title)
 - Change the mouse cursor when hovering over the message bar and its close button
+- Support combined bold and italic text (with `font.bold_italic` to customize it)
 
 ### Changed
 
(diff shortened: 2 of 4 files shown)


alacritty.yml
```diff
@@ -145,6 +145,17 @@ font:
...
+  # Bold italic font face
+  #bold_italic:
+    # Font family
+    #
+    # If the bold italic family is not specified, it will fall back to the
+    # value specified for the normal font.
+    #family: monospace
+
+    # The `style` can be specified to pick a specific face.
+    #style: Bold Italic
```
chrisduerr: Just as a note, I thought that `Bold Italic` might not be the correct thing here and instead maybe `BoldItalic` or something entirely different would be required to make things work. However, after testing it with `Fira Sans` (since `Fira Mono` does not have this variant), this does work properly and `BoldItalic` does not.

So looking good like this.

chris-morgan: Yeah, just as the normal font face is conventionally “Regular”, this one is conventionally “Bold Italic”.

chris-morgan:
This PR is properly formatted. `cargo fmt --all` makes two changes, from code on master (from 629ea247). I am loath to insert the fmt changes into this PR (despite the fact they’d be in a separate commit).

Good call on the changelog; I forgot to check if there was one. I used en-us for the spelling “customize”, matching what appears to be the convention.

chris-morgan:
Yeah, I’ve grown used to GitLab at work which makes *principled* force pushing quite tolerable, because it’s willing to show you the difference between the commits. I’ve certainly guided others away from *unprincipled* force-pushing or rebasing myself! But I do like my artisanally-crafted history. I shall avoid doing that here. ☺

chrisduerr:
> But I do like my artisanally-crafted history.

I do too! Which is why I squash all commits before merging and try to give them sensible commit messages. So no worries. :)

chris-morgan:
I am enjoying reading this repository’s git logs. It’s nice coming across others that care about these things too! (I am known at work for writing long and expressive commit messages. I think I have 17–18 of the longest 20 commit messages. 😀)

chrisduerr:
Thanks for the PR!

PR #2749, opened 2019-08-24
use std::mem::MaybeUninit::uninit().assume_init() instead of deprecated std::mem::uninitialized().

matthiaskrgr:


alacritty_terminal/src/tty/unix.rs
```diff
@@ -91,7 +90,7 @@ struct Passwd<'a> {
 /// If `buf` is changed while `Passwd` is alive, bad thing will almost certainly happen.
 fn get_pw_entry(buf: &mut [i8; 1024]) -> Passwd<'_> {
     // Create zeroed passwd struct
-    let mut entry: libc::passwd = unsafe { ::std::mem::uninitialized() };
+    let mut entry: libc::passwd = unsafe { ::std::mem::MaybeUninit::uninit().assume_init() };
```
rhysd: This is not correct usage of `MaybeUninit`. It causes undefined behavior. Somehow you need to initialize `libc::passwd` before calling `.assume_init()`.

From [official document](https://doc.rust-lang.org/beta/std/mem/union.MaybeUninit.html):

> Similarly, entirely uninitialized memory may have any content, while a bool must always be true or false. Hence, creating an uninitialized bool is undefined behavior:

```rust
use std::mem::{self, MaybeUninit};

let b: bool = unsafe { mem::uninitialized() }; // undefined behavior!
// The equivalent code with `MaybeUninit<bool>`:
let b: bool = unsafe { MaybeUninit::uninit().assume_init() }; // undefined behavior!
```

I recently created a similar PR to `term` crate. FWIW, please refer it how to handle initializing uninitialized memory with `MaybeUninit`:

https://github.com/Stebalien/term/pull/101

chrisduerr:
Small update, the glutin rework is on 1.36.0 minimum Rust version now, since that has been adopted by most distros now and it was causing some build failures.

I want to merge that in the near future for a broader testing audience, so I think we can get on this again after that.

chrisduerr:
I've decided to add the 1.36.0 update to the latest master separate from the glutin PR. So it should be possible to rebase this on the latest master and fix outstanding issues now @matthiaskrgr.

PR #2755, opened 2019-08-26
Fix url highlight not showing with required modifiers

chrisduerr:


kchibisov:
> Oh, add an entry to the CHANGELOG.md 😉

There is no need in `CHANGELOG.md` update, because this bug wasn't released yet.

chrisduerr:
> I find it interesting that having mouse: url: modifiers: Shift updates the pointer to a hand without moving the mouse (not underlining the URL however), while mouse: url: modifiers: Control requires the mouse to move before updating the cursor style.

This is a very good point, currently we only track the shift modifier. Generally I think the original PR should have been held back until after the evlp2 rework, but there'll likely be a second revision then.

So once https://github.com/rust-windowing/winit/issues/1124 is merged, we can probably just get rid of `Modifiers` and that will resolve all these issues.

PR #2762, opened 2019-08-28
Continuously track visible URLs

CrackedP0t:
This PR scans for URLs every frame, and it will only go 4096 characters away from the current viewport. It starts scanning 4096 characters below the viewport, and it stops scanning either once the last URL that extends above the viewport ends, or once the last URL reaches 4096 characters.

I made sure to scan before events are processed but after they're gathered, so that the events always operate on the same lock as the URL scan.

Thanks, feedback welcome!

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -47,6 +47,9 @@ pub mod color;
 /// Used to match equal brackets, when performing a bracket-pair selection.
 const BRACKET_PAIRS: [(char, char); 4] = [('(', ')'), ('[', ']'), ('{', '}'), ('<', '>')];
 
+/// The highest number of characters Alacritty will search for a URL
+const MAX_URL_LENGTH: usize = 2048;
+
 /// A type that can expand a given point to a region
 ///
 /// Usually this is implemented for some 2-D array type since
@@ -684,6 +687,9 @@ impl VisualBell {
 }
 
 pub struct Term<T> {
+    /// Terminal focus
+    pub is_focused: bool,
+
     /// The grid
     grid: Grid<Cell>,
 
@@ -755,11 +761,11 @@ pub struct Term<T> {
     /// Clipboard access coupled to the active window
     clipboard: Clipboard,
 
+    /// All the URL locations in the current view
+    urls: Vec<Url>,
+
     /// Proxy for sending events to the event loop
     event_proxy: T,
-
-    /// Terminal focus
-    pub is_focused: bool,
 }
 
 /// Terminal size info
@@ -887,6 +893,7 @@ impl<T> Term<T> {
             clipboard,
             event_proxy,
             is_focused: true,
+            urls: Vec::new(),
         }
     }
 
@@ -1233,18 +1240,23 @@ impl<T> Term<T> {
         &mut self.clipboard
     }
 
-    pub fn urls(&self) -> Vec<Url> {
+    pub fn update_urls(&mut self) {
+        self.urls.clear();
+
         let display_offset = self.grid.display_offset();
         let num_cols = self.grid.num_cols().0;
         let last_col = Column(num_cols - 1);
         let last_line = self.grid.num_lines() - 1;
 
-        let grid_end_point = Point::new(0, last_col);
-        let mut iter = self.grid.iter_from(grid_end_point);
+        let start_line = display_offset.saturating_sub(MAX_URL_LENGTH / num_cols);
+        let start_col = if display_offset == 0 { last_col } else { Column(MAX_URL_LENGTH % num_cols) };
+        let start_search_point = Point::new(start_line, start_col);
+        let mut iter = self.grid.iter_from(start_search_point);
 
         let mut parser = Parser::new();
         let mut extra_url_len = 0;
-        let mut urls = Vec::new();
+
+        let mut current_url_len = 0;
 
         let mut c = Some(iter.cell());
         while let Some(cell) = c {
@@ -1263,24 +1275,39 @@ impl<T> Term<T> {
                 parser.reset();
             }
 
+            let above_viewport = point.line > last_line.0 + display_offset;
+
+            current_url_len += 1;
+
             // Advance parser
             match parser.advance(cell.c) {
                 ParserState::Url(length) => {
-                    urls.push(Url::new(point, length + extra_url_len, num_cols))
+                    current_url_len = 0;
+
+                    self.urls.push(Url::new(point, length + extra_url_len, num_cols));
                 },
                 ParserState::NoUrl => {
                     extra_url_len = 0;
+                    current_url_len = 0;
 
                     // Stop searching for URLs once the viewport has been left without active URL
-                    if point.line > last_line.0 + display_offset {
+                    if above_viewport {
                         break;
                     }
                 },
                 _ => (),
             }
-        }
 
-        urls
+            // Don't cut off URLs in viewport, because we'll iterate over them anyway
+            if above_viewport && current_url_len > MAX_URL_LENGTH {
+                break;
+            }
+        }
+    }
+
+    #[inline]
+    pub fn urls(&self) -> &[Url] {
+        &self.urls
     }
 
     pub fn url_to_string(&self, url: Url) -> String {
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -358,7 +358,7 @@ impl Display {
     /// This call may block if vsync is enabled
     pub fn draw<T>(
         &mut self,
-        terminal: MutexGuard<'_, Term<T>>,
+        mut terminal: MutexGuard<'_, Term<T>>,
         message_buffer: &MessageBuffer,
         config: &Config,
     ) {
@@ -372,6 +372,11 @@ impl Display {
         // Update IME position
         #[cfg(not(windows))]
         self.window.update_ime_position(&terminal, &self.size_info);
+
+        // Update visible URLs
+        if config.ui_config.mouse.url.launcher.is_some() {
+            terminal.update_urls();
+        }
 
         // Drop terminal as early as possible to free lock
         drop(terminal);
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -252,9 +252,9 @@ impl<'a, T: EventListener, A: ActionContext<T> + 'a> Processor<'a, T, A> {
         {
             let buffer_point = self.ctx.terminal().visible_to_buffer(point);
             if let Some(url) =
-                self.ctx.terminal().urls().drain(..).find(|url| url.contains(buffer_point))
+                self.ctx.terminal().urls().iter().find(|url| url.contains(buffer_point))
             {
-                return MouseState::Url(url);
+                return MouseState::Url(*url);
             }
         }
 


alacritty_terminal/src/selection.rs
```diff
@@ -566,7 +566,7 @@ mod test {
         assert_eq!(selection.to_span(&term(5, 10)).unwrap(), Span {
             start: Point::new(0, Column(4)),
             end: Point::new(2, Column(4)),
-            is_block: true,
+            is_block: true
```
chrisduerr: All changes to this file should be reverted.

alacritty_terminal/src/input.rs
```diff
@@ -707,15 +707,15 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
+        let url = self.ctx.terminal().urls().iter().find(|url| url.contains(point))?;
         let text = self.ctx.terminal().url_to_string(&url);
 
         let launcher = self.mouse_config.url.launcher.as_ref()?;
         let mut args = launcher.args().to_vec();
         args.push(text);
 
         match start_daemon(launcher.program(), &args) {
-            Ok(_) => debug!("Launched {} with args {:?}", launcher.program(), args),
+            Ok(_) => println!("Launched {} with args {:?}", launcher.program(), args),
```
chrisduerr: I'd assume this `println` is for debugging? It should be removed before merging.

alacritty/src/main.rs
```diff
@@ -213,8 +213,16 @@ fn run(config: Config, message_buffer: MessageBuffer) -> Result<(), Box<dyn Erro
...
+        processor.wait_for_events(display.window());
+
+        let mut terminal_lock = terminal.lock();
+
+        // Make sure that events are processed with current URLs
+        terminal_lock.update_urls();
+
         // Process input and window events
-        let mut terminal_lock = processor.process_events(&terminal, display.window());
+        processor.process_events(&mut terminal_lock, display.window());
```
chrisduerr: There have been a lot of changes to `main.rs` and `event.rs`, which would likely introduce a lot of troubles with the #2438 PR.

Looking at these changes, it seems like it might be possible to just move `update_urls()` inside of the `process_events` method and that would just be a single line change instead of the differences introduced here.

This was just assessed based on a quick read over the code without any testing however, so it's possible that this is incorrect. If that's the case, maybe it's better to base this PR on top of the glutin rework instead?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Continuously track all URLs in view
```
chrisduerr: If there's no change for the user, we shouldn't add a change log entry.

I don't think this fixes any issues with URLs updating below the cursor yet, right? So we probably shouldn't add any change log entry if nothing changed for the user.

alacritty_terminal/src/selection.rs
CrackedP0t: Whoops, looks like my `rustfmt` did that by mistake

alacritty_terminal/src/input.rs
CrackedP0t: Oh, I missed that, thanks

alacritty/src/main.rs
CrackedP0t: OK, yeah, putting it in process_events makes sense if it makes the glutin rework easier - I made this change mostly just for ergonomics.

CHANGELOG.md
CrackedP0t: No, this doesn't fix those issues, although that would be a lot easier to fix with this change. However, fixing the URL update issues will probably require a lot of changes to `input.rs` and `update.rs` - I'm totally happy to work on that, but should I wait until the glutin rework lands first?

chrisduerr: Probably a good idea, yeah. I think I'm going to try to land the glutin rework for initial testing this weekend though.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1356,7 +1372,12 @@ impl Term {
             }
         }
 
-        urls
+        self.urls = urls;
```
chrisduerr: Instead of creating a new `Vec` and assigning it here, we can just call `self.urls.clear()` and then push directly to the `Vec` on `Term`.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1309,14 +1316,19 @@ impl Term {
...
         let last_col = Column(num_cols - 1);
         let last_line = self.grid.num_lines() - 1;
 
-        let grid_end_point = Point::new(0, last_col);
-        let mut iter = self.grid.iter_from(grid_end_point);
+        let max_above = Linear::new(Column(num_cols), Column(0), last_line + Line(display_offset))
+            + Linear(MAX_URL_LENGTH);
+
+        let start_line = display_offset.saturating_sub(MAX_URL_LENGTH / num_cols);
+        let start_col = if start_line == 0 { last_col } else { Column(MAX_URL_LENGTH % num_cols) };
```
chrisduerr: Is it correct that this might go slightly beyond `MAX_URL_LENGTH` when on the last line, since we always go to the last column here instead of making sure that the URL wouldn't have ended before it?

It seems like always returning `last_col` here is to make the codepath simpler where `saturating_sub` would usually make the line go negative, is that correct?

alacritty_terminal/src/term/mod.rs
CrackedP0t: Oh, that's a good idea, thanks!

alacritty_terminal/src/term/mod.rs
CrackedP0t: I can't see any situations when this will go beyond `MAX_URL_LENGTH`; am I missing some edge case?

`start_line` will never be negative, because it's a `usize` and `saturating_sub` clamps it to zero. The reason behind the `if` statement there is to ensure that when there are fewer characters past the viewport than `MAX_URL_LENGTH` that all of them are checked.

chrisduerr: This case is not only triggered when the length goes beyond the bottom-most line, but it's also triggered when it stops right in it. Then the columns will still get artificially extended to the last column, even though it would not be necessary.

This is because like this there's no way to differentiate between a `0` starting line which was truncated and one which was not.

At least that's my understanding. However that might not necessarily be a problem since it shouldn't make a massive difference to the URL length, however we should be aware that this basically makes it possible in theory to have infinite length URLs.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1356,20 +1351,32 @@ impl Term {
...
 
                     // Stop searching for URLs once the viewport has been left without active URL
-                    if point.line > last_line.0 + display_offset {
+                    if above_viewport {
                         break;
                     }
                 },
-                _ => (),
+                ParserState::MaybeUrl => {
+                    current_url_len += 1;
```
chrisduerr: Just a note since I've caught this change, this does not accurately track the URL length. When characters like brackets or quotes are included in the URL, the URL will be truncated when there is no matching opening quote/bracket.

CrackedP0t: It doesn't need to be super accurate; this is just used to make sure the URL isn't too long above the viewport

CrackedP0t:
Commit 8d3ea7c stops scanning for URLs once the maximum URL length has been hit immediately above the viewport, which increases performance in that benchmark. I also set the max URL length to 2048, although I'm happy to change it if some other limit seems better

alacritty_terminal/src/term/mod.rs
CrackedP0t: Alright, thanks for explaining it - does 
63aa56d solve this?

chrisduerr:
Thanks for the updates, I've run some benchmarks again and it seems like we're consistently at about 1% of a performance hit here on my system with the alt-screen-random-write benchmark.

Running things through `perf` it seems like some more enhancements could be made in rfind-url, though the hot codepaths are already pretty lean. I'll try to see if we can maybe work around this a little bit. Potentially there are also some simple places where we can skip the URL updating or we could do more of the work without the `Term` locked, so we wouldn't block up the parsing thread.

CrackedP0t:
Alright, makes sense - would you mind recommending a way to profile and identify good places to optimize? Thanks!

Also, I think we could only recheck for URLs when the terminal's dirty - I can't see any case where we'd want to check when it wouldn't be dirty.

chrisduerr:
> Also, I think we could only recheck for URLs when the terminal's dirty - I can't see any case where we'd want to check when it wouldn't be dirty.

We shouldn't redraw at all when it's not dirty, right? Forgot how this PR works again, but we shouldn't even reach the call in that case?

> would you mind recommending a way to profile and identify good places to optimize?

I've been testing things with `perf` and so far it looks like the rfind_url crate itself might need more optimizations. However since it's pretty optimized already, we might not be able to resolve the issues by improving the performance of the URL parsing alone, but instead have to parse URLs less somehow.

chrisduerr:
Okay, since we have a maximum URL length now, there's something I'd like to give a try.

Currently the `rfind_url` library parses the grid backwards, this was mainly with the goal to just start at the bottom of the grid and keep going without limiting URL length.

However, now that we have a limit on URL length, we have a fixed point where we can start when parsing forwards. So we know we'll never need to start parsing more than a few thousand characters above the viewport. With viewports containing easily ~20K+ (on my 1440p monitor) cells, that difference shouldn't matter to performance.

Since we have a limited number of protocols, parsing forwards should be much easier than parsing backwards, since we can discard sequences more easily before having to check semantics or track state. I'm not sure if this will make a massive difference, but I think it's worth a shot. Currently we're lacking only like 1 or 2 % of performance before this would be a no-brainer to merge, so small differences could matter a lot.

I'll give this a quick shot by writing a simple streaming URL parser in the next few days, I think I should be able to get something good going within a few days. If the comparison to rfind_url shows that we can gain maybe ~10%+ performance just by switching direction, I think it would be worth to try this.

I'll report back, sorry for the long silence on this. I really want to get this going before the next release though.

CrackedP0t:
Cool, seems worth a shot!

chrisduerr:
Okay, update on the URL situation. I've written a forwards-parsing URL library and I think it's much nicer than the backwards parsing one. I'm currently working on integration, though performance is pretty much identical. I'm not quite sure yet which of these libraries are best suited, but both should be fine.

I've also tested that VTE does not highlight URLs outside the visible region. I think that makes sense and I don't see a necessity for us to do it. So we can do the same here.

The reason why this is important is that it might allow us to use the renderable_cells vector to compute URLs only for display time without storing it on the term. This allows us to make sure we keep the term locked as little as possible, freeing up time for the parser to do its thing. If we do it this way, I would expect zero performance impact, since the parser is our hottest code (display is relatively cold).

I'll be playing around with this over the weekend.

chrisduerr:
I've decided to create a separate PR for my rework instead of pushing to this, so we have both available in theory in case anyone is curious.

I also have changed the code so much, that I'm not sure if any of your changes are still around, so you'd probably have a hard time resolving the remaining issues anyways. However if you find problems that aren't tracked yet and want to address them, feel free to create a PR against my repo's branch.

Thanks for your work, hopefully we can push this through quickly now. All performance issues have been resolved in my new PR, so we should be good now.

PR #2763, opened 2019-08-28
Fix legacy xparsecolor regression

### LINKED ISSUE #2759: Color issue introduced in commit 629e24
>(linux 5.1.5)

> Wayland (sway)

While testing the latest master I ran into an issue where the entire terminal will turn green.
Doing a git bisect points towards commit 629ea247cdccc33c6df0037f357ba9be48c7178d
I looked over the commit but could not find any obvious issues with the code, but then again I barely know rust. 

Let me know I you need any more info in order to debug.

chrisduerr:
The legacy xparsecolor implementation assumed that the \007 ending would
be passed to the parser, however it never is. This caused colors in the
format #rrggbb to be interpreted as #rrggb, leading to incorrect colors
showing up in Alacritty.

Fixes #2759.

### DIFF
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -27,11 +27,9 @@ use crate::term::color::Rgb;
 // Parse colors in XParseColor format
 fn xparse_color(color: &[u8]) -> Option<Rgb> {
     if !color.is_empty() && color[0] == b'#' {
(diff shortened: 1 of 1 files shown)


7415963987456321:
Currently at work. Will test this once I get home.

7415963987456321:
Just tested. Works as intended.

PR #2774, opened 2019-09-03
Fix Wayland primary clipboard storing when stopping selection outside…

kchibisov:
… of the window

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Intermediate bytes of CSI sequences not checked
 - Wayland clipboard integration
 - Use text mouse cursor when mouse mode is temporarily disabled with shift
+- Wayland primary selection clipboard not storing text when selection is stopped outside of the window
 
 ## 0.3.3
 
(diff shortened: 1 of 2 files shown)


chrisduerr:
I don't quite understand what you mean here? Does this not store the active selection when the mouse is released outside the window? That seems like something that isn't desirable.

kchibisov:
>  Does this not store the active selection when the mouse is released outside the window

Yes, and this is not intended behavior, since your pointer is steel bounded to `alacritty` and other apps(firefox/gtk apps) behave like `alacritty` with this patch applied. It also makes our primary selection on Wayland consistent with it on `X11`.

chrisduerr:
Oh, so you're saying this fix makes it so the selection is now stored when stopping it outside of the window, not removing this feature?

I got a bit confused by the wording of the PR then, sorry.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Intermediate bytes of CSI sequences not checked
 - Wayland clipboard integration
 - Use text mouse cursor when mouse mode is temporarily disabled with shift
+- Wayland primary clipboard storing when stopping selection outside of the window
```
chrisduerr: I think instead of saying we fixed that the clipboard stores the selection when stopping outside the window, I think it makes more sense to phrase this a bit differently:

```
- Wayland primary selection clipboard not storing text when selection is stopped outside of the window
```

Issue #2688, opened 2019-07-30
Add escape bindings for multiple combined modifiers


nifker:
Which operating system does the issue occur on?
Linux
If on linux, are you using X11 or Wayland?
AFAIK both

It seems like [Ctrl/Alt]+Shift+Arrowkey doesnt get catched by terminal applications when using alacritty.

See this issue for reference: https://github.com/zyedidia/micro/issues/1335

kchibisov:
Would you mind trying bindings suggested [here](https://github.com/jwilm/alacritty/issues/360#issuecomment-516360460)?

chrisduerr:
Based on some quick testing, it looks like Alacritty does not send the correct escapes when multiple modifiers are held down. We have bindings for shift, alt **or** control, but not combinations of those.

In theory, this is all the combinations we should support:
```
                    Code     Modifiers
                  ---------+---------------------------
                     2     | Shift
                     3     | Alt
                     4     | Shift + Alt
                     5     | Control
                     6     | Shift + Control
                     7     | Alt + Control
                     8     | Shift + Alt + Control
                     9     | Meta
                     10    | Meta + Shift
                     11    | Meta + Alt
                     12    | Meta + Alt + Shift
                     13    | Meta + Ctrl
                     14    | Meta + Ctrl + Shift
                     15    | Meta + Ctrl + Alt
                     16    | Meta + Ctrl + Alt + Shift
                  ---------+---------------------------
```

This is not only the case for the arrow keys, but also for everything else which uses that mechanism like PageUp/PageDown and function keys.

Implementing this is trivial, it's just a matter of adding additional key bindings, however I question whether we do want to have all of that in your configuration file. I feel like the key bindings are at a point, where we would benefit from not actually having all of them listed in the default config.

nifker:
@kchibisov seems to be working for alt Shift+Alt+Arrowkey - just missing the sequences for Shift+Ctrl+Arrowkey

kchibisov:
@nifker try this one
```yml
- { key: Up,       mods: Control|Shift,  chars: "\x1b[1;6A"                }
- { key: Down,     mods: Control|Shift,  chars: "\x1b[1;6B"                }
- { key: Left,     mods: Control|Shift,  chars: "\x1b[1;6D"                }
- { key: Right,    mods: Control|Shift,  chars: "\x1b[1;6C"                }
```

nifker:
That is working. Thanks. Shall I close this one now?

[...]

PR #2785, opened 2019-09-08
Rework default bindings

kchibisov:
This commit removes all bindings which are sending escapes from
the default configuration file, adds bindings for F13-F24, adds bindings
for ScrollToTop/ScrollToBottom actions, removes bindings for Super + F1-F12,
fixes bindings for Alt + F1-F12.

Fixes: #2688


alacritty.yml
```diff
@@ -567,105 +567,3 @@ key_bindings:
...
-  - { key: Home,                    chars: "\x1b[H",        mode: ~AppCursor }
-  - { key: End,      mods: Alt,     chars: "\x1b[1;3F"                       }
-  - { key: End,                     chars: "\x1bOF",        mode: AppCursor  }
-  - { key: End,                     chars: "\x1b[F",        mode: ~AppCursor }
-  - { key: PageUp,   mods: Shift,   action: ScrollPageUp,   mode: ~Alt       }
-  - { key: PageUp,   mods: Shift,   chars: "\x1b[5;2~",     mode: Alt        }
-  - { key: PageUp,   mods: Control, chars: "\x1b[5;5~"                       }
-  - { key: PageUp,   mods: Alt,     chars: "\x1b[5;3~"                       }
-  - { key: PageUp,                  chars: "\x1b[5~"                         }
-  - { key: PageDown, mods: Shift,   action: ScrollPageDown, mode: ~Alt       }
```
kchibisov: I wasn't able to keep `PageDown`/`PageUp` bindings in the config since if we leave them here, the predefined ones with modifiers will be erased which is not a desired behavior.

kchibisov: resolved

CHANGELOG.md
```diff
@@ -43,6 +45,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Wayland clipboard integration
 - Use text mouse cursor when mouse mode is temporarily disabled with shift
 - Wayland primary selection clipboard not storing text when selection is stopped outside of the window
+- Bindings for Alt + F1-F12
+- Missing bindings for combined modifiers
```
chrisduerr: These have been added, not fixed.

alacritty.yml
```diff
@@ -563,109 +563,15 @@ key_bindings:
...
+  - { key: L,         mods: Control, action: ClearLogNotice                   }
+  - { key: L,         mods: Control, chars: "\x0c"                            }
+  - { key: PageUp,    mods: Shift,   action: ScrollPageUp,   mode: ~Alt       }
+  - { key: PageUp,    mods: Shift,   chars: "\x1b[5;2~",     mode: Alt        }
+  - { key: PageDown,  mods: Shift,   action: ScrollPageDown, mode: ~Alt       }
+  - { key: PageDown,  mods: Shift,   chars: "\x1b[6;2~",     mode: Alt        }
+  - { key: Home,      mods: Shift,   action: ScrollToTop,    mode: ~Alt       }
+  - { key: Home,      mods: Shift,   chars: "\x1b[1;2H",     mode: Alt        }
+  - { key: End,       mods: Shift,   action: ScrollToBottom, mode: ~Alt       }
+  - { key: End,       mods: Shift,   chars: "\x1b[1;2F",     mode: Alt        }
```
chrisduerr: I don't think we should specify any of the `chars` default bindings here. Having only some of them would likely be even more confusing.

alacritty_terminal/src/config/bindings.rs
```diff
@@ -127,57 +113,107 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
...
+        Key::F15; Action::Esc("\x1b[1;2R".into());
+        Key::F16; Action::Esc("\x1b[1;2S".into());
+        Key::F17; Action::Esc("\x1b[15;2~".into());
+        Key::F18; Action::Esc("\x1b[17;2~".into());
+        Key::F19; Action::Esc("\x1b[18;2~".into());
+        Key::F20; Action::Esc("\x1b[19;2~".into());
+        Key::F21; Action::Esc("\x1b[20;2~".into());
+        Key::F22; Action::Esc("\x1b[21;2~".into());
+        Key::F23; Action::Esc("\x1b[23;2~".into());
+        Key::F24; Action::Esc("\x1b[24;2~".into());
```
chrisduerr: Where did you get these from? They don't seem to match XTerm's specification.

alacritty_terminal/src/config/bindings.rs
```diff
@@ -127,57 +113,107 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
...
+        ModifiersState { shift: true, ..ModifiersState::default() },
+        ModifiersState { alt: true, ..ModifiersState::default() },
+        ModifiersState { shift: true, alt: true, ..ModifiersState::default() },
+        ModifiersState { ctrl: true, ..ModifiersState::default() },
+        ModifiersState { shift: true, ctrl: true, ..ModifiersState::default() },
+        ModifiersState { alt: true, ctrl: true, ..ModifiersState::default() },
+        ModifiersState { shift: true, alt: true, ctrl: true, ..ModifiersState::default() },
+    ];
+
+    let mut modifers_num = 2;
```
chrisduerr: There's a typo here, should be modifiers, not modifers.

alacritty_terminal/src/config/bindings.rs
```diff
@@ -127,57 +113,107 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
...
+    for mods in modifiers.iter() {
```
chrisduerr: Can't we just use `modifiers.iter().enumerate()` to get the number? We just need to add 2.

alacritty_terminal/src/config/bindings.rs
```diff
@@ -127,57 +113,107 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
...
+        let mods_num = modifers_num.to_string();
+        bindings.extend(bindings!(
+            KeyBinding;
+            Key::Up,    [shift: mods.shift, alt: mods.alt, ctrl: mods.ctrl];
+            Action::Esc("\x1b[1;".to_owned()  + &mods_num + "A");
```
chrisduerr:
```suggestion
            Action::Esc(format!("\x1b[1;{}{}", mods_num, "A");
```

I think this should be shorter and it doesn't require the `modifiers_num.to_string()`.

alacritty.yml
chrisduerr: I don't think this has been addressed?

kchibisov:
blocked by #2794

alacritty_terminal/src/config/bindings.rs
```diff
@@ -71,113 +71,158 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
     let mut bindings = bindings!(
         KeyBinding;
         Key::Paste; Action::Paste;
-        Key::Copy; Action::Copy;
+        Key::Copy;  Action::Copy;
```
nixpulvis: Worlds smallest nit-pick. But I think you've added an extra space here :stuck_out_tongue_winking_eye:

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Completions for `--class` and `-t` (short title)
 - Change the mouse cursor when hovering over the message bar and its close button
 - Support combined bold and italic text (with `font.bold_italic` to customize it)
+- Extra bindings for F13-F20
+- Bindings with combined modifiers
```
nixpulvis: Could this be more clear? I'm not sure which bindings this is referring too.

kchibisov: I don't want this to be too bloated.

chrisduerr: I think it could be clarified a tiny bit without much extra wording.

Something like this maybe:

```
 - Terminal escape bindings with combined modifiers
```

That way people know just slightly more about it, but it's still short and clear.

alacritty_terminal/src/config/bindings.rs
chrisduerr: I don't think there's any reason for this whitespace change?

alacritty_terminal/src/config/bindings.rs
```diff
@@ -71,113 +71,158 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
...
-        Key::PageDown, [alt: true]; Action::Esc("\x1b[6;3~".into());
+        Key::Home,     [shift: true], ~TermMode::ALT_SCREEN; Action::ScrollToTop;
+        Key::Home,     [shift: true], +TermMode::ALT_SCREEN; Action::Esc("\x1b[1;2H".into());
+        Key::End,      [shift: true], ~TermMode::ALT_SCREEN; Action::ScrollToBottom;
+        Key::End,      [shift: true], +TermMode::ALT_SCREEN; Action::Esc("\x1b[1;2F".into());
+        Key::Home, +TermMode::APP_CURSOR; Action::Esc("\x1bOH".into());
+        Key::Home, ~TermMode::APP_CURSOR; Action::Esc("\x1b[H".into());
+        Key::End,  +TermMode::APP_CURSOR; Action::Esc("\x1bOF".into());
+        Key::End,  ~TermMode::APP_CURSOR; Action::Esc("\x1b[F".into());
+        Key::PageUp;   Action::Esc("\x1b[5~".into());
```
chrisduerr: Previously these bindings were nicely sorted by `key`. I think that's a good idea to keep confusion to a minimum.

Also I don't think I've stated this before, but just wanted to make sure that I think that adding these additional `Home` and `End` bindings makes a lot of sense (for jumping to start/end of scrollback).

alacritty_terminal/src/config/bindings.rs
```diff
@@ -71,113 +71,158 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
...
+    //    10    | Meta + Shift
+    //    11    | Meta + Alt
+    //    12    | Meta + Alt + Shift
+    //    13    | Meta + Ctrl
+    //    14    | Meta + Ctrl + Shift
+    //    15    | Meta + Ctrl + Alt
+    //    16    | Meta + Ctrl + Alt + Shift
+    // ---------+---------------------------
+    //
+    // from: https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-PC-Style-Function-Keys
```
chrisduerr: Great idea to add a table and source here.

alacritty_terminal/src/config/bindings.rs
```diff
@@ -71,113 +71,158 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
...
+            Action::Esc(format!("\x1b[31;{}~", modifiers_code));
+            Key::F18,   [shift: mods.shift, alt: mods.alt, ctrl: mods.ctrl];
+            Action::Esc(format!("\x1b[32;{}~", modifiers_code));
+            Key::F19,   [shift: mods.shift, alt: mods.alt, ctrl: mods.ctrl];
+            Action::Esc(format!("\x1b[33;{}~", modifiers_code));
+            Key::F20,   [shift: mods.shift, alt: mods.alt, ctrl: mods.ctrl];
+            Action::Esc(format!("\x1b[34;{}~", modifiers_code));
+        ));
+
+        if modifiers_code != 2 {
```
chrisduerr: I think a comment explaining why we do things differently here would go a long way. Otherwise it would probably strike people as odd.

alacritty.yml
chrisduerr: I still kinda wish we could do something about this. Maybe the new PR helps here too? I think we should be able to work around this.

alacritty_terminal/src/config/bindings.rs
nixpulvis: I had the same comment, but it's to align with the above action. This is consistent with the rest of this `bindings!` macro invocation:

https://github.com/jwilm/alacritty/blob/748138240508f9a032cfde9f0e397f7fc3be5eb8/alacritty_terminal/src/config/bindings.rs#L89-L90

chrisduerr: Oh right. Somehow I checked the action above it and came to the conclusion they do not line up. But very clearly they do.

I think I even noticed this in the last review and didn't comment because I saw that they line up. So please just disregard.

alacritty_terminal/src/config/bindings.rs
```diff
@@ -71,113 +71,158 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
...
+        ModifiersState { alt: true, ..ModifiersState::default() },
+        ModifiersState { shift: true, alt: true, ..ModifiersState::default() },
+        ModifiersState { ctrl: true, ..ModifiersState::default() },
+        ModifiersState { shift: true, ctrl: true, ..ModifiersState::default() },
+        ModifiersState { alt: true, ctrl: true, ..ModifiersState::default() },
+        ModifiersState { shift: true, alt: true, ctrl: true, ..ModifiersState::default() },
+    ];
+
+    for (modifiers_code, mods) in modifiers.iter().enumerate() {
+        let modifiers_code = modifiers_code + 2;
```
nixpulvis: This might be clearer if you don't reuse `modifiers_code`, since technically, until you shadow the binding introduced from the `for` it's inaccurate. Instead I might write:

```rust
let modifiers_code = index + 2;
```

Very small point tho.

alacritty_terminal/src/config/bindings.rs
nixpulvis: Could we give names to each of these magic numbers from the table with a bitflags struct? I'd much prefer to be comparing like `if modifiers_code != ModifierCode::Shift`. It might clean up the iteration too a little bit, IDK.

alacritty_terminal/src/config/bindings.rs
nixpulvis: I too like these new bindings :tada:

kchibisov: These bindings were never sorted by key. They were placed in groups by key, however I can't group them like it was before, because our tests are position dependent.

Just a reference on how they look now and before.

Screenshot_Wed 18 Sep 2019 02:51:45 PM MSK

alacritty_terminal/src/config/bindings.rs
kchibisov: It would be nice, if `2 | 3 == 4` and other bitwise math would map on the codes, but that's not the case, so the array + table in a comment should do the job right. I also think that's too much hassle for the code that will likely remain untouched anymore.

kchibisov: I've added a comment on why we're skipping things here btw

alacritty_terminal/src/config/bindings.rs
```diff
@@ -71,113 +71,160 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
...
+    //    7     | Alt + Control
+    //    8     | Shift + Alt + Control
+    //    9     | Meta
+    //    10    | Meta + Shift
+    //    11    | Meta + Alt
+    //    12    | Meta + Alt + Shift
+    //    13    | Meta + Ctrl
+    //    14    | Meta + Ctrl + Shift
+    //    15    | Meta + Ctrl + Alt
+    //    16    | Meta + Ctrl + Alt + Shift
```
nixpulvis: Should we include the "Meta" codes if we're not using them? Would we ever want to use these?

kchibisov: make sense

alacritty_terminal/src/config/bindings.rs
nixpulvis: OK, yea I tried for a moment to make something nicer and it's indeed probably not worth it. However I think the comment should address the main difference I see, which is that in the bindings defined above, there's explicit actions, and conditions for the `mode`. I think ideally we'd just overwrite them, but since I don't think that's easy to do ATM, so this is fine.

alacritty_terminal/src/config/bindings.rs
nixpulvis: @kchibisov I'd say we should feel free to change the tests to match the format we want here. Especially if this is going to be the order they appear in our reference `alacritty.yml` file. As a user, I'd really like to see all keys together so I can think about what I want that key to do in the same place.

This is a pretty small issue tho.

Issue #1544, opened 2018-09-17
Investigate macOS tabs


jwilm:
Per @pcwalton's [comment], this may actually be somewhat easy to support.

[comment]: https://news.ycombinator.com/item?id=18007940

paulrouget:
I have some experience in this area. We ran into the same issues with Servo.

On a general note, for tabs, menu, popups (etc) support or any proper integration with MacOS, we tried:

1. create an xcode + obj-c/swift project that would load Servo as a library (no winit/glutin)
2. create a xib the we would in Rust (see https://github.com/paulrouget/servoshell, no winit/glutin)
3. use the [native handle exposed by winit](https://docs.rs/winit/0.11.3/winit/os/index.html) and dynamically build the menu and all the missing MacOS integration bits

*2* was my favorite approach as it didn't require writing Obj-c/Swift code, but we could still design the UI with XCode. But we lose winit events which was great for the multiplatform abstraction. It's probably possible now to tell winit to re-use an existing window (the xib loader creates its own window iirc).

*3* works. It's not optimal as everything needs to be done manually.

I think the best long term approach is to add these features to winit.

As for tabs, specifically, both for Mac and Windows (window support is still in beta, see Redstone 5 release), it's just matter of supporting multiple windows and flagging them a "tab-able", and the OS will do the rest.

Hope that helps.

jwilm:
Thanks for sharing your experience here @paulrouget! Even better that there's a big example program to learn from. :smile:

I hope you don't mind; I have a few questions.

1. When you say "adding these features to winit", are you calling out xib loader/initialization for a winit window?
2. In servoshell, did you reimplement event handling on top of the handle returned by the xib loader?
3. Reflecting on this a bit further, it sounds like servo isn't using winit any longer. Is that true? What major features are you missing from winit today beyond xib support?

cc @francesca64 who might be interested in this discussion

paulrouget:
> 1. When you say "adding these features to winit", are you calling out xib loader/initialization for a winit window?

No. I never used xib + winit.

I meant adding methods to the winit builder to support tabs and menus.

> 2. In servoshell, did you reimplement event handling on top of the handle returned by the xib loader?

[...]

PR #2786, opened 2019-09-08
Add Xembed support.

mkosem:
This is a basic implementation of XEmbed.  Please go easy on me.  I'm a Java developer, and haven't seen rust code until about 3 days ago.  Nevertheless, my interest in XEmbed for this application is to allow it to be used in conjunction with Suckless tabbed - which seems to work fine with not too much effort.

I normally run a basic terminal in tabbed in dwm, but here's what it looks like in Xfce:
image

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -134,7 +134,7 @@ impl Options {
                 Arg::with_name("embed")
                     .long("embed")
                     .takes_value(true)
-                    .help("Defines the X11 window ID (as a decimal integer) to embed Alacritty within."),
+                    .help("Defines the X11 window ID (as a decimal integer) to embed Alacritty within"),
             )
             .arg(
                 Arg::with_name("q")
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -71,106 +71,104 @@ impl Options {
 
         let mut options = Options::default();
 
-        let matches = App::new(crate_name!())
-            .version(version.as_str())
-            .author(crate_authors!("\n"))
-            .about(crate_description!())
-            .arg(Arg::with_name("ref-test").long("ref-test").help("Generates ref test"))
-            .arg(
-                Arg::with_name("live-config-reload")
-                    .long("live-config-reload")
-                    .help("Enable automatic config reloading"),
-            )
-            .arg(
-                Arg::with_name("no-live-config-reload")
-                    .long("no-live-config-reload")
-                    .help("Disable automatic config reloading")
-                    .conflicts_with("live-config-reload"),
-            )
-            .arg(
-                Arg::with_name("print-events")
-                    .long("print-events")
-                    .help("Print all events to stdout"),
-            )
-            .arg(
-                Arg::with_name("persistent-logging")
-                    .long("persistent-logging")
-                    .help("Keep the log file after quitting Alacritty"),
-            )
-            .arg(
-                Arg::with_name("dimensions")
-                    .long("dimensions")
-                    .short("d")
-                    .value_names(&["columns", "lines"])
-                    .help(
-                        "Defines the window dimensions. Falls back to size specified by window \
-                         manager if set to 0x0 [default: 0x0]",
-                    ),
-            )
-            .arg(
-                Arg::with_name("position")
-                    .long("position")
-                    .allow_hyphen_values(true)
-                    .value_names(&["x-pos", "y-pos"])
-                    .help(
-                        "Defines the window position. Falls back to position specified by window \
-                         manager if unset [default: unset]",
-                    ),
-            )
-            .arg(
-                Arg::with_name("title")
-                    .long("title")
-                    .short("t")
-                    .takes_value(true)
-                    .help(&format!("Defines the window title [default: {}]", DEFAULT_NAME)),
-            )
-            .arg(
-                Arg::with_name("class")
-                    .long("class")
-                    .takes_value(true)
-                    .help(&format!("Defines window class on Linux [default: {}]", DEFAULT_NAME)),
-            )
-            .arg(
-                Arg::with_name("embed")
-                    .long("embed")
-                    .takes_value(true)
-                    .help("Defines the X11 window ID (as a decimal integer) to embed Alacritty within"),
-            )
-            .arg(
-                Arg::with_name("q")
-                    .short("q")
-                    .multiple(true)
-                    .conflicts_with("v")
-                    .help("Reduces the level of verbosity (the min level is -qq)"),
-            )
-            .arg(
-                Arg::with_name("v")
-                    .short("v")
-                    .multiple(true)
-                    .conflicts_with("q")
-                    .help("Increases the level of verbosity (the max level is -vvv)"),
-            )
-            .arg(
-                Arg::with_name("working-directory")
-                    .long("working-directory")
-                    .takes_value(true)
-                    .help("Start the shell in the specified working directory"),
-            )
-            .arg(Arg::with_name("config-file").long("config-file").takes_value(true).help(
-                "Specify alternative configuration file [default: \
-                 $XDG_CONFIG_HOME/alacritty/alacritty.yml]",
-            ))
-            .arg(
-                Arg::with_name("command")
-                    .long("command")
-                    .short("e")
-                    .multiple(true)
-                    .takes_value(true)
-                    .min_values(1)
-                    .allow_hyphen_values(true)
-                    .help("Command and args to execute (must be last argument)"),
-            )
-            .get_matches();
+        let matches =
+            App::new(crate_name!())
+                .version(version.as_str())
+                .author(crate_authors!("\n"))
+                .about(crate_description!())
+                .arg(Arg::with_name("ref-test").long("ref-test").help("Generates ref test"))
+                .arg(
+                    Arg::with_name("live-config-reload")
+                        .long("live-config-reload")
+                        .help("Enable automatic config reloading"),
+                )
+                .arg(
+                    Arg::with_name("no-live-config-reload")
+                        .long("no-live-config-reload")
+                        .help("Disable automatic config reloading")
+                        .conflicts_with("live-config-reload"),
+                )
+                .arg(
+                    Arg::with_name("print-events")
+                        .long("print-events")
+                        .help("Print all events to stdout"),
+                )
+                .arg(
+                    Arg::with_name("persistent-logging")
+                        .long("persistent-logging")
+                        .help("Keep the log file after quitting Alacritty"),
+                )
+                .arg(
+                    Arg::with_name("dimensions")
+                        .long("dimensions")
+                        .short("d")
+                        .value_names(&["columns", "lines"])
+                        .help(
+                            "Defines the window dimensions. Falls back to size specified by \
+                             window manager if set to 0x0 [default: 0x0]",
+                        ),
+                )
+                .arg(
+                    Arg::with_name("position")
+                        .long("position")
+                        .allow_hyphen_values(true)
+                        .value_names(&["x-pos", "y-pos"])
+                        .help(
+                            "Defines the window position. Falls back to position specified by \
+                             window manager if unset [default: unset]",
+                        ),
+                )
+                .arg(
+                    Arg::with_name("title")
+                        .long("title")
+                        .short("t")
+                        .takes_value(true)
+                        .help(&format!("Defines the window title [default: {}]", DEFAULT_NAME)),
+                )
+                .arg(
+                    Arg::with_name("class").long("class").takes_value(true).help(&format!(
+                        "Defines window class on Linux [default: {}]",
+                        DEFAULT_NAME
+                    )),
+                )
+                .arg(Arg::with_name("embed").long("embed").takes_value(true).help(
+                    "Defines the X11 window ID (as a decimal integer) to embed Alacritty within",
+                ))
+                .arg(
+                    Arg::with_name("q")
+                        .short("q")
+                        .multiple(true)
+                        .conflicts_with("v")
+                        .help("Reduces the level of verbosity (the min level is -qq)"),
+                )
+                .arg(
+                    Arg::with_name("v")
+                        .short("v")
+                        .multiple(true)
+                        .conflicts_with("q")
+                        .help("Increases the level of verbosity (the max level is -vvv)"),
+                )
+                .arg(
+                    Arg::with_name("working-directory")
+                        .long("working-directory")
+                        .takes_value(true)
+                        .help("Start the shell in the specified working directory"),
+                )
+                .arg(Arg::with_name("config-file").long("config-file").takes_value(true).help(
+                    "Specify alternative configuration file [default: \
+                     $XDG_CONFIG_HOME/alacritty/alacritty.yml]",
+                ))
+                .arg(
+                    Arg::with_name("command")
+                        .long("command")
+                        .short("e")
+                        .multiple(true)
+                        .takes_value(true)
+                        .min_values(1)
+                        .allow_hyphen_values(true)
+                        .help("Command and args to execute (must be last argument)"),
+                )
+                .get_matches();
 
         if matches.is_present("ref-test") {
             options.ref_test = true;
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -134,7 +134,7 @@ impl Options {
                 Arg::with_name("embed")
                     .long("embed")
                     .takes_value(true)
-                    .help("Defines an X window ID (as a decimal integer) which will serve as the parent when embedding Alacritty within other applications."),
+                    .help("Defines the X11 window ID (as a decimal integer) to embed Alacritty within."),
             )
             .arg(
                 Arg::with_name("q")
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -134,7 +134,7 @@ impl Options {
                 Arg::with_name("embed")
                     .long("embed")
                     .takes_value(true)
-                    .help("Sets the XEmbed parent window ID (as a long value)"),
+                    .help("Defines an X window ID (as a decimal integer) which will serve as the parent when embedding Alacritty within other applications."),
             )
             .arg(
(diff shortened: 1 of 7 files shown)


chrisduerr:
Just as a quick update, looks like we will be merging this into Alacritty for now.

If you don't mind, I'd like to get #2336 ready first and then get to this PR afterwards. So I'll wait with the review until then to make sure things don't need to be needlessly changed.

kchibisov:
Just to clarify things, this PR is blocked by #2438 and not by #2336, as were originally stated.

P.s.
It seems like `github` autocomplete failure.

alacritty_terminal/src/config/window.rs
```diff
@@ -39,6 +39,10 @@ pub struct WindowConfig {
     #[serde(deserialize_with = "from_string_or_deserialize")]
     pub class: Class,
 
+    /// XEmbed parent
+    #[serde(deserialize_with = "failure_default")]
+    pub embed: Option<u64>,
```
chrisduerr: I think we should only offer this on Linux/BSD.

extra/alacritty.man
```diff
@@ -57,6 +57,9 @@ Defines the window position. Falls back to position specified by window manager
 \fB\-t\fR, \fB\-\-title\fR <title>
 Defines the window title [default: Alacritty]
 .TP
+\fB\-\-embed\fR <parent>
+Sets the XEmbed parent window
```
chrisduerr: The type we expect (decimal) should be documented, this is also the case for the `--help` output.

kchibisov:
You've added `fish`/`zsh` completions, however you're missing `bash` one.

extra/alacritty.man
kchibisov: In addition to @chrisduerr comment, the man entry should be consistent with `--help` output, but as I can see they differ a bit.

alacritty_terminal/src/window.rs
```diff
@@ -167,6 +169,14 @@ impl Window {
         // Set OpenGL symbol loader. This call MUST be after window.make_current on windows.
         gl::load_with(|symbol| windowed_context.get_proc_address(symbol) as *const _);
 
+        // On X11, embed the window inside another if the parent ID has been set
+        #[cfg(not(any(target_os = "macos", windows)))]
+        {
+            if let Some(parent_window_id) = config.window.embed {
```
kchibisov: It's not exposed, that it's on X11. You should add  `if event_loop.is_x11()` or something, since there's Wayland.

alacritty_terminal/src/config/window.rs
mkosem: I'm new to all this.  Can you elaborate on what you're expecting based on this coment?

extra/alacritty.man
mkosem: This should (hopefully) be set in my last commit.

alacritty_terminal/src/config/window.rs
kchibisov: You should add `#[cfg(not(any(target_os = "macos", windows)))]` above `#[serde(...)]`. It's expecting that this option will only work on Linux/BSD systems, take a look on `StartupMode`.`SimpleFullscreen` below, as an example.

chrisduerr: I'm actually not so sure anymore if this is a good idea. Since this is an `Option` anyways, I don't think it would make any difference at all.

The same thing goes for the command line arguments, while it would be nice to only show the commands available on that platform, I don't want to ship multiple manpages etc. So I think the best bet is to just skip that for now. I don't think it's such a big deal.

So let's focus on the error handling first, that's far more important.

extra/alacritty.man
```diff
@@ -57,6 +57,9 @@ Defines the window position. Falls back to position specified by window manager
 \fB\-t\fR, \fB\-\-title\fR <title>
 Defines the window title [default: Alacritty]
 .TP
+\fB\-\-embed\fR <parent>
+Sets the XEmbed parent window ID (as a long value)
```
chrisduerr: I don't think using `long` here is doing us any benefits, since it's more likely to confuse or help.

Just calling it a decimal integer is likely to be more user friendly. We could optionally also add a little more info to what this does, I think XTerm does a good job here: https://linux.die.net/man/1/xterm (search for 'embed')

mkosem: Ok.  Hopefully the current rendition is better.

extra/alacritty.man
mkosem: Ok, so I looked into making things more resilient if the window ID is bogus or the window doesn't exist.  The X FAQ suggests that checking for the existence of the target isn't really recommended:

Subject: 121)  How do I check whether a window ID is valid?
My program has the ID of a window on a remote display. I want to check whether
the window exists before doing anything with it.

	Because X is asynchronous, there isn't a guarantee that the window would
still exist between the time that you got the ID and the time you sent an event
to the window or otherwise manipulated it. What you should do is send the event
without checking, but install an error handler to catch any BadWindow errors, 
which would indicate that the window no longer exists. This scheme will work 
except on the [rare] occasion that the original window has been destroyed and 
its ID reallocated to another window.

Based on that, it seems like gracefully tolerating the issue is the most likely route to success.  It seems like std::panic::catch_unwind is the way this would be done in rust, but I don't see any examples of that in this codebase.  Before I try implementing it, I'd like to ask - Would using that meet your expectations?

alacritty_terminal/src/window.rs
mkosem: K.  Added.

extra/completions/alacritty.fish
```diff
@@ -29,6 +29,9 @@ complete -c alacritty \
 complete -c alacritty \
   -l "class" \
   -d "Defines the window class"
+complete -c alacritty \
+  -l "embed" \
+  -d "Defines the XEmbed parent"
```
kchibisov: You should keep completions help persistent across zsh/fish/help etc.

kchibisov:
Agree, there's no reason to add this one to config. I tend to think that this option was added to a config file unintentionally, since the `alacritty.yml` change is missing, am I correct here @mkosem?.

alacritty/src/cli.rs
```diff
@@ -128,6 +130,12 @@ impl Options {
                     .takes_value(true)
                     .help(&format!("Defines window class on Linux [default: {}]", DEFAULT_NAME)),
             )
+            .arg(
+                Arg::with_name("embed")
+                    .long("embed")
+                    .takes_value(true)
+                    .help("Defines an X window ID (as a decimal integer) which will serve as the parent when embedding Alacritty within other applications."),
```
nixpulvis: This is rather long still, and doesn't make it super clear that `--embed` is the setting determining that Alacritty **will** be embedded. Maybe something simple like this would be better?

> Defines the X window ID of the parent to embed Alacritty within.

Food for thought.

nixpulvis:
Just tried this out with `tabbed`, and it seems to be working very well :tada: I personally use a WM capable of creating tabs, however this effectively solves the problem for anyone on X11 using another kind of WM. I'd be curious if it's nearly this simple to add native macOS tab support (see #1544), since people would really like that.

FWIW, here's what it looks like:

alacritty-2786

alacritty/src/cli.rs
mkosem: All: I made one more attempt at satisfying everyone's requests here.  It seems like some are in favor of less info, and others in favor of more, I tried to encompass the necessary details everyone has requested in as few words as reasonable.  Hopefully this helps.

@kchibisov Yeah, that was not intentional.  I'm not familiar with this code base and used the most similar other argument I could find as an example of where to add this stuff.  Could you do me a favor and clue me in to what, specifically, is causing this to be bound to the config file?  Is it the presence in src/config/window.rs coupled with some ?  If so, what would be the preferred mechanism for propagating this argument from the command line while bypassing that code path?  The current mechanism seems convenient/clean.

chrisduerr: Current version seems good to me, though please adjust punctuation accordingly, we want to be consistent here.

I'm very much in favor of short, but it's absolutely necessary to point out the datatype since the default window ID is always printed in the format `0x01238047` as hexadecimal number. And we do not support this.

extra/alacritty.man
chrisduerr: > The X FAQ suggests that checking for the existence of the target isn't really recommended

That doesn't mean we can't handle the errors. There's nothing garceful about the current approach, we just crash with a bogus error. `catch_unwind` should basically never used, since this is Rust we don't catch errors, we handle them.

This can simply be resolved by handling the `BadWindow` response that X will give when attempting to embed with an invalid window and printing an appropriate error message when that is the case.

alacritty/src/cli.rs
mkosem: Ok.  Should be consistent now.

extra/alacritty.man
mkosem: K.  I'll try to figure out what that involves tonight.  Thanks for the guidance.

chrisduerr: No worries, I think this just involves one of the usual C error handling ways where you ask X11 for the last error and it'll tell you about it. I haven't looked too much into it but I think the X11 bindings have something like this.

alacritty/src/cli.rs
kchibisov: @mkosem

You should just put `#[serde(skip_deserializing)]` instead of what you're having now iirc in config/window.rs. You're already ignoring this option when you're setting it from cli though, but removing this `+1` config parsing should be a good idea.

extra/alacritty.man
mkosem: K.  I pushed a change to handle re-parenting failures.  I'm not sure what the deal with the result from XReparentWindow is, but it isn't meaningful (always 1, success or failure) and the error handler doesn't get fired when it fails.  I couldn't find any examples, even in C, of code making use of the result.  Nevertheless, attempting to get attributes from the window after reparenting does get the feedback needed to assert the success of the reparenting.

alacritty/src/cli.rs
mkosem: Thanks! :)  Done.

alacritty_terminal/src/window.rs
```diff
@@ -409,6 +421,49 @@ impl Window {
...
+        (xlib.XSetErrorHandler)(std::mem::transmute(old_handler));
+    }
+}
+
+unsafe extern "C" fn ctx_error_handler(
+        _dpy: *mut x11_dl::xlib::Display,
+        _ev: *mut x11_dl::xlib::XErrorEvent
+        ) -> i32 {
+    eprintln!("Could not embed into specified window.");
+    ::std::process::exit(1);
```
kchibisov: It's quite old syntax here, please, convert it to `process::exit(1)` + import at top.

alacritty_terminal/src/window.rs
```diff
@@ -409,6 +421,49 @@ impl Window {
...
+            [0, 1].as_ptr(),
+            2,
+        );
+        
+        // set a handler to deal with unsuccessful reparenting
+        let old_handler = (xlib.XSetErrorHandler)(Some(ctx_error_handler));
+        // reparent the window with a specific handler in-place
+        (xlib.XReparentWindow)(xlib_display as _, xlib_window as _, parent_window_id, 0, 0);
+        // XReparentWindow doesn't give helpful results, so poke the window here
+        let mut attribs: x11_dl::xlib::XWindowAttributes = std::mem::uninitialized();
```
kchibisov: We should replace `std::mem::uninitialized` with `MaybeUninit::uninit()`, since the first one is deprecated.

mkosem: Thanks.  Done.

alacritty_terminal/src/window.rs
mkosem: Done.

alacritty_terminal/src/window.rs
```diff
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 use std::convert::From;
-#[cfg(not(any(target_os = "macos", target_os = "windows")))]
+use std::process;
```
kchibisov: this import is useless on macOS/Windows so we need the same `#cfg` as for `c_void`. This one should be the last thing from my side.

alacritty_terminal/src/window.rs
```diff
@@ -409,6 +422,50 @@ impl Window {
...
+            2,
+        );
+        
+        // set a handler to deal with unsuccessful reparenting
+        let old_handler = (xlib.XSetErrorHandler)(Some(ctx_error_handler));
+        // reparent the window with a specific handler in-place
+        (xlib.XReparentWindow)(xlib_display as _, xlib_window as _, parent_window_id, 0, 0);
+        // XReparentWindow doesn't give helpful results, so poke the window here
+        let mut attribs = std::mem::MaybeUninit::<x11_dl::xlib::XWindowAttributes>::uninit();
+        (xlib.XGetWindowAttributes)(xlib_display as _, xlib_window as _, attribs.as_mut_ptr());
```
chrisduerr: I don't quite understand why we're reading the window attributes here?

mkosem: It's sort of explained in the comment.  The result from XReparentWindow isn't a helpful way to detect failures there.  xterm does this the same way.

See the same in their detection here:
https://github.com/joejulian/xterm/blob/defc6dd5684a12dc8e56cb6973ef973e7a32caa3/misc.c#L5744

alacritty_terminal/src/window.rs
mkosem: Should be sorted.  Thanks again.

alacritty_terminal/src/window.rs
chrisduerr: Hmm, that's very interesting. But has lead me to look at how URxvt does it and that might actually be even better.

While URxvt prints more information that indicates they've got an error hadler setup, they basically just check the return type of `XGetWindowAttributes`. If that works (which it does for URxvt apparently), we can just skip the whole error handler nonsense and check for the return type to determine what's going on.

https://github.com/exg/rxvt-unicode/blob/4b8969cc0eab47944f5964c917fda2b4a9c431d7/src/init.C#L1353-L1365

What both seem to do is also updating the viewport dimensions. But since we actually check the viewport size *after* the window was created, I think we should be able to get away with not doing this.

alacritty_terminal/src/window.rs
```diff
@@ -409,6 +423,51 @@ impl Window {
...
+            xlib_window as _,
+            atom,
+            atom,
+            32,
+            PropModeReplace,
+            [0, 1].as_ptr(),
+            2,
+        );
+        
+        // set a handler before verifying target window existence
```
kchibisov: Oh, just noticed. We're starting our `//` comments with uppercase later. So should be `Set` etc.

chrisduerr: Oh yeah, noticed that too but forget to mention it. Thanks for pointing it out.

mkosem: Sorted :)

alacritty_terminal/src/window.rs
mkosem: I like it, and it works well with less code.  Thanks.

mkosem:
Thanks!  All set from my end. Thanks for being patient with me :)!

alacritty_terminal/src/window.rs
```diff
@@ -409,6 +425,39 @@ impl Window {
...
 }
 
+#[cfg(not(any(target_os = "macos", windows)))]
+fn x_embed_window(window: &GlutinWindow, parent_id: u64) {
+    let (xlib_display, xlib_window) = match (window.get_xlib_display(), window.get_xlib_window()) {
+        (Some(display), Some(window)) => (display, window),
+        _ => return,
+    };
+
+    let xlib = Xlib::open().expect("get xlib");
```
nixpulvis: If there's a way to get the `xconn.xlib` that winit already has? That way, ideally, we don't need to check for this error again here.

alacritty_terminal/src/window.rs
```diff
@@ -409,6 +425,39 @@ impl Window {
...
+            32,
+            PropModeReplace,
+            [0, 1].as_ptr(),
+            2,
+        );
+
+        // Check for the existence of the target before attempting reparenting
+        let mut attrs = MaybeUninit::uninit();
+        if (xlib.XGetWindowAttributes)(xlib_display as _, parent_id, attrs.as_mut_ptr()) == 0 {
+            eprintln!("Could not embed into specified window.");
```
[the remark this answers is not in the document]
chrisduerr: > this is a sign that this functionality might really want to live upstream

We've already established that this is not an option. So I don't think there's any point in trying to pursue this.

> I generally don't like killing processes deep in the call stack

Fundamentally this isn't a bad notion, however handling this error genericallly means we either have to specifically catch this and print the error itself, or we'll have to live with what is more a crash than error handling.

I think there's value in making sure an error is as clean as possible when we're talking about something that is purely there for user feedback and likely to be encountered.

alacritty_terminal/src/window.rs
chrisduerr: I think we could use [get_xcb_connection](https://docs.rs/winit/0.19.3/winit/os/unix/trait.WindowExt.html#tymethod.get_xcb_connection) and [get_xlib_display](https://docs.rs/winit/0.19.3/winit/os/unix/trait.WindowExt.html#tymethod.get_xlib_display) to construct an [XCB Connection](http://rtbo.github.io/rust-xcb/xcb/base/struct.Connection.html) and then use [reparent_window](http://rtbo.github.io/rust-xcb/xcb/xproto/fn.reparent_window.html) to embed the [winit window](https://docs.rs/winit/0.19.3/winit/os/unix/trait.WindowExt.html#tymethod.get_xlib_window).

mkosem:
Yeah, this is along the lines of something I tried over the weekend. The XSync stuff didn't seem to help, for me when I was trying that, but I may well have been doing something incorrectly and plenty of other apps are doing similar stuff so it looks like a sound approach.

mkosem:
Sorry, didn't mean to close that.

PR #2787, opened 2019-09-08
Fix #1165 - Implement the "hold" feature

### LINKED ISSUE #1165: Add support for --hold flag
The `--hold` flag is supported by many other terminal emulators and causes the terminal window to stay open after the command returns.

This is useful when using gdb to debug a console application: one can run `favorite_terminal --hold -e tty` to get the tty name and leave it open. 
Then in gdb,  call `tty <tty-name>` before running the application. It will be started on the terminal held open. 
This doesn't work if the command being run is not finished, as the terminal already has a "session leader".

valignatev:
Hi there! Thanks a lot for your work on Alacritty, this project been helpful for me in countless ways, so I want to contribute something back. This PR implements `--hold` flag together with the `hold` configuration option which keeps Alacritty open after its child process exits.

Maybe I should clarify why I decided to put the `hold` flag in the `Term` struct instead of passing it directly to the `EventLoop` where the handling of child process exit happens. I've tried this approach first, but it turned out to be extremely hard to write a test for (at least for me, I'm a total noob in Rust). I've also tried ref-tests, but for some reason, even with just opening and closing Alacritty I get 60 megabytes of `grid.json`, and a generated ref test is always failing. I'm not sure why it happens as I'm not very familiar with how `Grid` works yet.

So I've taken the easy-to-test approach, I hope it's good enough :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bindings for ScrollToTop and ScrollToBottom actions
 - `ReceiveChar` key binding action to insert the key's text character
 - Live reload font size from config
-- New CLI flag `--hold` for keeping Alacritty opened after its child process exits.
+- New CLI flag `--hold` for keeping Alacritty opened after its child process exits
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bindings for ScrollToTop and ScrollToBottom actions
 - `ReceiveChar` key binding action to insert the key's text character
 - Live reload font size from config
+- New CLI flag `--hold` for keeping Alacritty opened after its child process exits.
 
 ### Changed
 
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -36,6 +36,7 @@ pub struct Options {
     pub embed: Option<String>,
     pub log_level: LevelFilter,
     pub command: Option<Shell<'static>>,
+    pub hold: bool,
     pub working_dir: Option<PathBuf>,
     pub config: Option<PathBuf>,
     pub persistent_logging: bool,
@@ -54,6 +55,7 @@ impl Default for Options {
             embed: None,
             log_level: LevelFilter::Warn,
             command: None,
+            hold: false,
             working_dir: None,
             config: None,
             persistent_logging: false,
@@ -169,6 +171,7 @@ impl Options {
                         .allow_hyphen_values(true)
                         .help("Command and args to execute (must be last argument)"),
                 )
+                .arg(Arg::with_name("hold").long("hold").help("Remain open after child process exits"))
                 .get_matches();
 
         if matches.is_present("ref-test") {
@@ -239,6 +242,10 @@ impl Options {
             options.command = Some(Shell::new_with_args(command, args));
         }
 
+        if matches.is_present("hold") {
+            options.hold = true;
+        }
+
         options
     }
 
@@ -277,6 +284,7 @@ impl Options {
         if config.debug.print_events {
             config.debug.log_level = max(config.debug.log_level, LevelFilter::Info);
         }
+        config.hold = self.hold;
 
         config
     }
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -171,7 +171,11 @@ impl Options {
                         .allow_hyphen_values(true)
                         .help("Command and args to execute (must be last argument)"),
                 )
-                .arg(Arg::with_name("hold").long("hold").help("Remain open after child process exits"))
+                .arg(
+                    Arg::with_name("hold")
+                        .long("hold")
+                        .help("Remain open after child process exits")
+                )
                 .get_matches();
 
         if matches.is_present("ref-test") {
@@ -262,6 +266,8 @@ impl Options {
         );
         config.shell = self.command.or(config.shell);
 
+        config.hold = self.hold;
+
         config.window.dimensions = self.dimensions.unwrap_or(config.window.dimensions);
         config.window.position = self.position.or(config.window.position);
         config.window.title = self.title.or(config.window.title);
@@ -284,7 +290,6 @@ impl Options {
         if config.debug.print_events {
             config.debug.log_level = max(config.debug.log_level, LevelFilter::Info);
         }
-        config.hold = self.hold;
 
         config
     }
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -74,109 +74,106 @@ impl Options {
 
         let mut options = Options::default();
 
-        let matches =
-            App::new(crate_name!())
-                .version(version.as_str())
-                .author(crate_authors!("\n"))
-                .about(crate_description!())
-                .arg(Arg::with_name("ref-test").long("ref-test").help("Generates ref test"))
-                .arg(
-                    Arg::with_name("live-config-reload")
-                        .long("live-config-reload")
-                        .help("Enable automatic config reloading"),
-                )
-                .arg(
-                    Arg::with_name("no-live-config-reload")
-                        .long("no-live-config-reload")
-                        .help("Disable automatic config reloading")
-                        .conflicts_with("live-config-reload"),
-                )
-                .arg(
-                    Arg::with_name("print-events")
-                        .long("print-events")
-                        .help("Print all events to stdout"),
-                )
-                .arg(
-                    Arg::with_name("persistent-logging")
-                        .long("persistent-logging")
-                        .help("Keep the log file after quitting Alacritty"),
-                )
-                .arg(
-                    Arg::with_name("dimensions")
-                        .long("dimensions")
-                        .short("d")
-                        .value_names(&["columns", "lines"])
-                        .help(
-                            "Defines the window dimensions. Falls back to size specified by \
-                             window manager if set to 0x0 [default: 0x0]",
-                        ),
-                )
-                .arg(
-                    Arg::with_name("position")
-                        .long("position")
-                        .allow_hyphen_values(true)
-                        .value_names(&["x-pos", "y-pos"])
-                        .help(
-                            "Defines the window position. Falls back to position specified by \
-                             window manager if unset [default: unset]",
-                        ),
-                )
-                .arg(
-                    Arg::with_name("title")
-                        .long("title")
-                        .short("t")
-                        .takes_value(true)
-                        .help(&format!("Defines the window title [default: {}]", DEFAULT_NAME)),
-                )
-                .arg(
-                    Arg::with_name("class").long("class").takes_value(true).help(&format!(
-                        "Defines window class on Linux [default: {}]",
-                        DEFAULT_NAME
-                    )),
-                )
-                .arg(Arg::with_name("embed").long("embed").takes_value(true).help(
+        let matches = App::new(crate_name!())
+            .version(version.as_str())
+            .author(crate_authors!("\n"))
+            .about(crate_description!())
+            .arg(Arg::with_name("ref-test").long("ref-test").help("Generates ref test"))
+            .arg(
+                Arg::with_name("live-config-reload")
+                    .long("live-config-reload")
+                    .help("Enable automatic config reloading"),
+            )
+            .arg(
+                Arg::with_name("no-live-config-reload")
+                    .long("no-live-config-reload")
+                    .help("Disable automatic config reloading")
+                    .conflicts_with("live-config-reload"),
+            )
+            .arg(
+                Arg::with_name("print-events")
+                    .long("print-events")
+                    .help("Print all events to stdout"),
+            )
+            .arg(
+                Arg::with_name("persistent-logging")
+                    .long("persistent-logging")
+                    .help("Keep the log file after quitting Alacritty"),
+            )
+            .arg(
+                Arg::with_name("dimensions")
+                    .long("dimensions")
+                    .short("d")
+                    .value_names(&["columns", "lines"])
+                    .help(
+                        "Defines the window dimensions. Falls back to size specified by window \
+                         manager if set to 0x0 [default: 0x0]",
+                    ),
+            )
+            .arg(
+                Arg::with_name("position")
+                    .long("position")
+                    .allow_hyphen_values(true)
+                    .value_names(&["x-pos", "y-pos"])
+                    .help(
+                        "Defines the window position. Falls back to position specified by window \
+                         manager if unset [default: unset]",
+                    ),
+            )
+            .arg(
+                Arg::with_name("title")
+                    .long("title")
+                    .short("t")
+                    .takes_value(true)
+                    .help(&format!("Defines the window title [default: {}]", DEFAULT_NAME)),
+            )
+            .arg(
+                Arg::with_name("class")
+                    .long("class")
+                    .takes_value(true)
+                    .help(&format!("Defines window class on Linux [default: {}]", DEFAULT_NAME)),
+            )
+            .arg(
+                Arg::with_name("embed").long("embed").takes_value(true).help(
                     "Defines the X11 window ID (as a decimal integer) to embed Alacritty within",
-                ))
-                .arg(
-                    Arg::with_name("q")
-                        .short("q")
-                        .multiple(true)
-                        .conflicts_with("v")
-                        .help("Reduces the level of verbosity (the min level is -qq)"),
-                )
-                .arg(
-                    Arg::with_name("v")
-                        .short("v")
-                        .multiple(true)
-                        .conflicts_with("q")
-                        .help("Increases the level of verbosity (the max level is -vvv)"),
-                )
-                .arg(
-                    Arg::with_name("working-directory")
-                        .long("working-directory")
-                        .takes_value(true)
-                        .help("Start the shell in the specified working directory"),
-                )
-                .arg(Arg::with_name("config-file").long("config-file").takes_value(true).help(
-                    "Specify alternative configuration file [default: \
-                     $XDG_CONFIG_HOME/alacritty/alacritty.yml]",
-                ))
-                .arg(
-                    Arg::with_name("command")
-                        .long("command")
-                        .short("e")
-                        .multiple(true)
-                        .takes_value(true)
-                        .min_values(1)
-                        .allow_hyphen_values(true)
-                        .help("Command and args to execute (must be last argument)"),
-                )
-                .arg(
-                    Arg::with_name("hold")
-                        .long("hold")
-                        .help("Remain open after child process exits")
-                )
-                .get_matches();
+                ),
+            )
+            .arg(
+                Arg::with_name("q")
+                    .short("q")
+                    .multiple(true)
+                    .conflicts_with("v")
+                    .help("Reduces the level of verbosity (the min level is -qq)"),
+            )
+            .arg(
+                Arg::with_name("v")
+                    .short("v")
+                    .multiple(true)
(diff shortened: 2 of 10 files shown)


chrisduerr:
Thanks for the PR.

Before going into detail with this PR, I'm curious why we need dedicated support for this. It seems like it should be possible to do this just by specifying a shell parameter/config which doesn't exit?

valignatev:
> It seems like it should be possible to do this just by specifying a shell parameter/config which doesn't exit

I thought this as well until I've found out that shells don't do this in a convenient way. They usually offer `-c` flag as a way to evaluate specific command instead of reading stdin, but they present it as an alternative of the interactive mode and usually you can't keep the shell going after the command finishes unless you try to work around it somehow. Examples for bash: https://stackoverflow.com/a/3531426/3606603, https://unix.stackexchange.com/q/123103/245683

Similar hacks might work for zsh, and I couldn't find similar hacks for fish. Maybe that's the reason why this `hold` feature exists in every other terminal emulators (Termite, kitty, xterm, gnome-terminal, konsole, Xfce4-terminal and Terminal are the ones I know for sure to have this).

P.S. With that being said, I don't want to sound like "every other terminal has this so we should as well", I understand that it is not a good argument as is, and that the power of Alacritty is about _not having_ features. Also, maybe I'm missing something really obvious and simple way of doing it without a separate terminal flag.

chrisduerr:
First of all, I'd like to point out that the issues you've linked and the hold flag are fundamentally two completely different things. The hold flags just keeps the window open, while the stackexchange and stackoverflow links request to be able to input additional commands, which is not possible with the hold flag.

If you want to enter additional commands, you have to do something like this:
```
alacritty -e sh -c "ls && exec sh"
```

However, if the goal is to keep the terminal emulator open, there's a very simple way to do this with the exact same pattern:
```
alacritty -e sh -c "ls && sleep infinity"
```

As far as I can tell, the only drawback of this way of doing things is that it's a bit longer and that it keeps the shell around. I am not aware of any case where this would be relevant, so I don't think that's significant justification.

I'm generally not against adding this option (though it should not be a config option), but I'd obviously like a good reason to use it over the solution we already have. There's no reason to support two different ways to do things when one of them doesn't have significant advantages.

I've personally never used the hold flag in any terminal emulator, so I might not have a complete overview over the different usecases. So I'd appreciate all feedback on this. Otherwise it seems tough to justify.

valignatev:
> First of all, I'd like to point out that the issues you've linked and the hold flag are fundamentally two completely different things

I understand and agree, it's just the only relevant thing I found about "not closing the terminal after program exits" without using a terminal feature

> though it should not be a config option
Do you mean it shouldn't be in the `alacritty.yml` file or it shouldn't stay in the `Config` struct at all?

TIL about `sleep infinity` trick, never seen it before, cool. It looks less convenient than `--hold` flag, but definitely not the end of the world. I'm not sure though if it is possible to do this as simple in Windows though.

My main case for this is to sort off a web browser's "open in the new window". Previously I was using Termite, and I would do something like this from the project root:
```
$ termite -e ./start_server --hold &
$ termite -e ./start_webpack --hold &
```
This way if the server dies I will have a chance to read its output. Also, I was often using `--hold` to open some noisy command in the new terminal window quickly so it doesn't ruin my main working window with its noise.
Without hold it will be like:
```
$ termite -e sh -c "./start_server && sleep infinity" &
$ termite -e sh -c "./start_webpack && sleep infinity" &
```
Well, one might argue that you have to type it only once for every often used command and doesn't justify a whole CLI option, I don't know, such are my cases :)

chrisduerr:
Okay, I think we want to support `--hold`, since it's fairly straight-forward and commonly used. It allows re-running with a different command a bit easier and doesn't require any shell.

However, I don't think there's much of a reason to add a configuration option, so I'd skip that.

Since the code touched by this is significantly changed in the glutin rework (#2438), which is expected to be merged in the reasonably near future, I think it makes sense to wait until after that. So I'll get back to this and add a proper review once that's done.

valignatev:
Horay! :tada: 
> However, I don't think there's much of a reason to add a configuration option, so I'd skip that.
I agree with that, the only reason I've added it is that I was under impression that the `Config` struct should fully match config file, now I see that the struct has some additional fields.

So I'll patiently wait for glutin rework to be merged then.

Once again, thanks a lot for your time and patience :heart:

chrisduerr:
The glutin rework PR has been merged now. If you need any help with rebasing, please let me know.

valignatev:
Great! I'll take a look and ask for help if I need any

valignatev:
Ok, I've rebased the branch onto the new master. I removed the option from the config file. And I think that the tests are more legit with the new event_proxy, I was able to get rid of `should_exit` which I've added before solely for testing purposes.

alacritty_terminal/src/config/mod.rs
```diff
@@ -133,6 +133,10 @@ pub struct Config<T> {
     #[serde(flatten)]
     pub ui_config: T,
 
+    /// Remain open after child process exits
+    #[serde(default, deserialize_with = "failure_default")]
```
kchibisov: FYI, you can still set this option in `alacritty.yml`. To fix this issue you should replace current `#[serde(...)` with `#[serde(skip)`.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bindings for ScrollToTop and ScrollToBottom actions
 - `ReceiveChar` key binding action to insert the key's text character
 - Live reload font size from config
+- New CLI flag `--hold` for keeping Alacritty opened after its child process exits.
```
kchibisov: We don't have dots in the end, please remove it.

alacritty_terminal/src/event_loop.rs
```diff
@@ -327,7 +327,7 @@ where
                         #[cfg(unix)]
                         token if token == self.pty.child_event_token() => {
                             if let Some(tty::ChildEvent::Exited) = self.pty.next_child_event() {
-                                self.terminal.lock().exit();
+                                self.terminal.lock().handle_child_process_exit();
```
chrisduerr: Why not store `hold` on `self` here and check in here rather than checking in the terminal? Seems a bit complicated to store things on the terminal, if it's only needed here.

alacritty/src/cli.rs
```diff
@@ -169,6 +171,7 @@ impl Options {
                         .allow_hyphen_values(true)
                         .help("Command and args to execute (must be last argument)"),
                 )
+                .arg(Arg::with_name("hold").long("hold").help("Remain open after child process exits"))
```
kchibisov: This formatting seems odd. You should format your code with `cargo +nightly fmt` (p.s. nightly branch is required for alacritty atm).

alacritty_terminal/src/config/mod.rs
chrisduerr: This is a good idea.

alacritty/src/cli.rs
```diff
@@ -277,6 +284,7 @@ impl Options {
         if config.debug.print_events {
             config.debug.log_level = max(config.debug.log_level, LevelFilter::Info);
         }
+        config.hold = self.hold;
```
kchibisov: I think moving this under config.shell will be better for structure, so

```rust
config.shell = ...

config.hold = self.hold

config.window = ...
```
or without empty line between shell and hold, idk.

chrisduerr: I don't think this is related to shell, so I see little reason to move it. However I do think we should add an empty line above it, since it's also not related to `print_events`.

kchibisov: I've suggested it because there is a giant `debug` section in the end, so I treat this `debug` block like a finishing one. So moving somewhere `hold` about it end emphasize with new line will make sense in my opinion. Feel free to ignore this comment.

chrisduerr: If the `debug` block is at the end or not shouldn't matter that much as long as it's not interrupted and cleanly separated from everything else I think. Before it would be fine too though.

nixpulvis: I might put it under line 263, under the command bit, just like it is for the CLI matching: https://github.com/jwilm/alacritty/pull/2787/files/8c501bcdcb2794eeb42928294dd1cdbf900ed691#diff-d72ceef0f1f8da7f6ac545eb5dbd0f63R236-R247

alacritty_terminal/src/term/mod.rs
```diff
@@ -2301,6 +2326,52 @@ mod tests {
...
+        config.hold = false;
+        let mut term = Term::new(
+            &config,
+            &size,
+            Clipboard::new_nop(),
+            EventSpy { latest_event: RefCell::new(None) },
+        );
+        term.handle_child_process_exit();
+        assert_eq!(term.event_proxy.latest_event.into_inner(), Some(Event::Exit));
+    }
```
nixpulvis: Nice tests :tada:

We should really have a faster way to build a test `Terminal`...

alacritty/src/cli.rs
valignatev: @kchibisov Hmm, interesting. When I do `cargo + nightly fmt` in the project root, it reformats a `let matches` section completely, but it doesn't touch my `hold` match addition. I'm not sure if I'm doing it right.

```patch
diff --git a/alacritty/src/cli.rs b/alacritty/src/cli.rs
index d1bd919..c697850 100644
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -74,105 +74,106 @@ impl Options {
 
-        let matches =
-            App::new(crate_name!())
-                .version(version.as_str())
-                .author(crate_authors!("\n"))
-                .about(crate_description!())
-                .arg(Arg::with_name("ref-test").long("ref-test").help("Generates ref test"))
-                .arg(
-                    Arg::with_name("live-config-reload")
-                        .long("live-config-reload")
-                        .help("Enable automatic config reloading"),
-                )
-                .arg(
-                    Arg::with_name("no-live-config-reload")
-                        .long("no-live-config-reload")
-                        .help("Disable automatic config reloading")
-                        .conflicts_with("live-config-reload"),
-                )
-                .arg(
-                    Arg::with_name("print-events")
-                        .long("print-events")
-                        .help("Print all events to stdout"),
-                )
-                .arg(
-                    Arg::with_name("persistent-logging")
-                        .long("persistent-logging")
-                        .help("Keep the log file after quitting Alacritty"),
-                )
-                .arg(
-                    Arg::with_name("dimensions")
-                        .long("dimensions")
-                        .short("d")
-                        .value_names(&["columns", "lines"])
-                        .help(
-                            "Defines the window dimensions. Falls back to size specified by \
-                             window manager if set to 0x0 [default: 0x0]",
-                        ),
-                )
-                .arg(
-                    Arg::with_name("position")
-                        .long("position")
-                        .allow_hyphen_values(true)
-                        .value_names(&["x-pos", "y-pos"])
-                        .help(
-                            "Defines the window position. Falls back to position specified by \
-                             window manager if unset [default: unset]",
-                        ),
-                )
-                .arg(
-                    Arg::with_name("title")
-                        .long("title")
-                        .short("t")
-                        .takes_value(true)
-                        .help(&format!("Defines the window title [default: {}]", DEFAULT_NAME)),
-                )
-                .arg(
-                    Arg::with_name("class").long("class").takes_value(true).help(&format!(
-                        "Defines window class on Linux [default: {}]",
-                        DEFAULT_NAME
-                    )),
-                )
-                .arg(Arg::with_name("embed").long("embed").takes_value(true).help(
+        let matches = App::new(crate_name!())
+            .version(version.as_str())
+            .author(crate_authors!("\n"))
+            .about(crate_description!())
+            .arg(Arg::with_name("ref-test").long("ref-test").help("Generates ref test"))
+            .arg(
+                Arg::with_name("live-config-reload")
+                    .long("live-config-reload")
+                    .help("Enable automatic config reloading"),
+            )
+            .arg(
+                Arg::with_name("no-live-config-reload")
+                    .long("no-live-config-reload")
+                    .help("Disable automatic config reloading")
+                    .conflicts_with("live-config-reload"),
+            )
+            .arg(
+                Arg::with_name("print-events")
+                    .long("print-events")
+                    .help("Print all events to stdout"),
+            )
+            .arg(
+                Arg::with_name("persistent-logging")
+                    .long("persistent-logging")
+                    .help("Keep the log file after quitting Alacritty"),
+            )
+            .arg(
+                Arg::with_name("dimensions")
+                    .long("dimensions")
+                    .short("d")
+                    .value_names(&["columns", "lines"])
+                    .help(
+                        "Defines the window dimensions. Falls back to size specified by window \
+                         manager if set to 0x0 [default: 0x0]",
+                    ),
+            )
+            .arg(
+                Arg::with_name("position")
+                    .long("position")
+                    .allow_hyphen_values(true)
+                    .value_names(&["x-pos", "y-pos"])
+                    .help(
+                        "Defines the window position. Falls back to position specified by window \
+                         manager if unset [default: unset]",
+                    ),
+            )
+            .arg(
+                Arg::with_name("title")
+                    .long("title")
+                    .short("t")
+                    .takes_value(true)
+                    .help(&format!("Defines the window title [default: {}]", DEFAULT_NAME)),
+            )
+            .arg(
+                Arg::with_name("class")
+                    .long("class")
+                    .takes_value(true)
+                    .help(&format!("Defines window class on Linux [default: {}]", DEFAULT_NAME)),
+            )
+            .arg(
+                Arg::with_name("embed").long("embed").takes_value(true).help(
                     "Defines the X11 window ID (as a decimal integer) to embed Alacritty within",
-                ))
-                .arg(
-                    Arg::with_name("q")
-                        .short("q")
-                        .multiple(true)
-                        .conflicts_with("v")
-                        .help("Reduces the level of verbosity (the min level is -qq)"),
-                )
-                .arg(
-                    Arg::with_name("v")
-                        .short("v")
-                        .multiple(true)
-                        .conflicts_with("q")
-                        .help("Increases the level of verbosity (the max level is -vvv)"),
-                )
-                .arg(
-                    Arg::with_name("working-directory")
-                        .long("working-directory")
-                        .takes_value(true)
-                        .help("Start the shell in the specified working directory"),
-                )
-                .arg(Arg::with_name("config-file").long("config-file").takes_value(true).help(
-                    "Specify alternative configuration file [default: \
-                     $XDG_CONFIG_HOME/alacritty/alacritty.yml]",
-                ))
-                .arg(
-                    Arg::with_name("command")
-                        .long("command")
-                        .short("e")
-                        .multiple(true)
-                        .takes_value(true)
-                        .min_values(1)
-                        .allow_hyphen_values(true)
-                        .help("Command and args to execute (must be last argument)"),
-                )
-                .arg(Arg::with_name("hold").long("hold").help("Remain open after child process exits"))
-                .get_matches();
+                ),
+            )
+            .arg(
+                Arg::with_name("q")
+                    .short("q")
+                    .multiple(true)
+                    .conflicts_with("v")
+                    .help("Reduces the level of verbosity (the min level is -qq)"),
+            )
+            .arg(
+                Arg::with_name("v")
+                    .short("v")
+                    .multiple(true)
+                    .conflicts_with("q")
+                    .help("Increases the level of verbosity (the max level is -vvv)"),
+            )
+            .arg(
+                Arg::with_name("working-directory")
+                    .long("working-directory")
+                    .takes_value(true)
+                    .help("Start the shell in the specified working directory"),
+            )
+            .arg(Arg::with_name("config-file").long("config-file").takes_value(true).help(
+                "Specify alternative configuration file [default: \
+                 $XDG_CONFIG_HOME/alacritty/alacritty.yml]",
+            ))
+            .arg(
+                Arg::with_name("command")
+                    .long("command")
+                    .short("e")
+                    .multiple(true)
+                    .takes_value(true)
+                    .min_values(1)
+                    .allow_hyphen_values(true)
+                    .help("Command and args to execute (must be last argument)"),
+            )
+            .arg(Arg::with_name("hold").long("hold").help("Remain open after child process exits"))
+            .get_matches();
 
         if matches.is_present("ref-test") {
             options.ref_test = true;
diff --git a/alacritty_terminal/src/term/mod.rs b/alacritty_terminal/src/term/mod.rs
index d2014ce..f073174 100644
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -2340,12 +2340,9 @@ mod tests {
         };
         let mut config = MockConfig::default();
         config.hold = true;
-        let mut term = Term::new(
-            &config,
-            &size,
-            Clipboard::new_nop(),
-            EventSpy { latest_event: RefCell::new(None) },
-        );
+        let mut term = Term::new(&config, &size, Clipboard::new_nop(), EventSpy {
+            latest_event: RefCell::new(None),
+        });
         term.handle_child_process_exit();
         assert_eq!(term.event_proxy.latest_event.into_inner(), None);
     }
@@ -2363,12 +2360,9 @@ mod tests {
         };
         let mut config = MockConfig::default();
         config.hold = false;
-        let mut term = Term::new(
-            &config,
-            &size,
-            Clipboard::new_nop(),
-            EventSpy { latest_event: RefCell::new(None) },
-        );
+        let mut term = Term::new(&config, &size, Clipboard::new_nop(), EventSpy {
+            latest_event: RefCell::new(None),
+        });
         term.handle_child_process_exit();
         assert_eq!(term.event_proxy.latest_event.into_inner(), Some(Event::Exit));
     }
```

valignatev: `config-file` arg formatting looks kinda similar.

alacritty_terminal/src/event_loop.rs
valignatev: I did it because I wanted to test it as well, and testing it if it sits on the EventLoop beyond my current skills :) Maybe, since the whole feature boils down to pretty much one `if` statement, tests don't matter that much, but I'm not sure. What do you think?

alacritty/src/cli.rs
valignatev: Anyways, I reformatted it in the same ways as most of the args formatted, I hope it's good enough now.

valignatev:
Travis isn't in the mood today, but I've fixed formatting issues and added the man entry.

alacritty_terminal/src/event_loop.rs
chrisduerr: Testing the `exit` method makes sense, but testing a single conditional doesn't hold much value. I'd much prefer a cleaner code style over that personally.

valignatev: Ok, I've moved `hold` to the EventLoop, not the condition is under the `ChildEvent::Exited` match arm and there's no point in having `handle_child_process_exit` on the `Term`, so I've deleted it as well. Much less new code this way :)

extra/completions/_alacritty
```diff
@@ -11,6 +11,7 @@ _arguments \
   "--print-events[print all events to stdout]" \
   '(-v)'{-q,-qq}"[reduce the level of verbosity (min is -qq)]" \
   "--ref-test[generate ref test]" \
+  "--hold[Remain open after child process exits]" \
```
kchibisov: `Remain` -> `remain`, for consistency with the rest of zsh complitions.

valignatev: Good catch, fixed

alacritty_terminal/src/event_loop.rs
```diff
@@ -43,6 +43,7 @@ pub struct EventLoop<T: tty::EventedPty, U: EventListener> {
     tx: Sender<Msg>,
     terminal: Arc<FairMutex<Term<U>>>,
     event_proxy: U,
+    hold: bool,
     ref_test: bool,
```
chrisduerr: We're passing two config values here, usually at that point it's shorter to just pass `&config` in and then take out the specific fields in the struct construction.

valignatev: You're right, I should've thought about that. Just did.

valignatev:
:tada: 
@chrisduerr @nixpulvis @kchibisov  Thanks a lot for your patience, guidance and super fast responses!

chrisduerr:
Thanks for working on this!

PR #2793, opened 2019-09-10
Fix empty block selection detection

### LINKED ISSUE #2789: 0 Width Vertical Block Selections
I've uncovered a bug while testing #2719 (as noted there). Basically we shouldn't allow empty vertical block selections of width 0. This can cause a situation where it looks like there is no selection, but in fact there is one. Since a selection prevents actions like clicking a URL, this can lead to strange user interactions.

block-selection-url-bug-2019-09-07_00 03 52

To reproduce on master, create a 0 width N height block selection, and then attempt to click a URL. I believe a proper fix for this is to disallow 0 width block selections, similarly to how other selections disappear.

chrisduerr:
Fixes #2789.

### DIFF
--- a/alacritty_terminal/src/selection.rs
+++ b/alacritty_terminal/src/selection.rs
@@ -140,7 +140,8 @@ impl Selection {
                         (&region.start, &region.end)
                     };
 
-                // Empty when single cell with identical sides or two cell with right+left sides
+                // Simple selection is empty when the points are identical
+                // or two adjacent cells have the sides right -> left
                 start == end
                     || (start.side == Side::Left
                         && end.side == Side::Right
@@ -148,7 +149,9 @@ impl Selection {
                         && start.point.col == end.point.col + 1)
             },
             Selection::Block { region: Range { ref start, ref end } } => {
-                // Empty when at most line different or two cells with right+left sides
+                // Block selection is empty when the points' columns and sides are identical
+                // or two cells with adjacent columns have the sides right -> left,
+                // regardless of their lines
                 (start.point.col == end.point.col && start.side == end.side)
                     || (start.point.col == end.point.col + 1
                         && start.side == Side::Left
--- a/alacritty_terminal/src/selection.rs
+++ b/alacritty_terminal/src/selection.rs
@@ -132,7 +132,7 @@ impl Selection {
 
     pub fn is_empty(&self) -> bool {
         match *self {
-            Selection::Simple { ref region } | Selection::Block { ref region } => {
+            Selection::Simple { ref region } => {
                 let (start, end) =
                     if Selection::points_need_swap(region.start.point, region.end.point) {
                         (&region.end, &region.start)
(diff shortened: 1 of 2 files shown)


nixpulvis:
I've written two tests for `is_empty`. One for simple and one for block selections:

```rust
#[test]
fn simple_is_empty() {
    let mut selection = Selection::simple(Point::new(0, Column(0)), Side::Right);
    assert!(selection.is_empty());
    selection.update(Point::new(0, Column(1)), Side::Left);
    assert!(selection.is_empty());
    selection.update(Point::new(1, Column(0)), Side::Right);
    assert!(!selection.is_empty());
}

#[test]
fn block_is_empty() {
    let mut selection = Selection::block(Point::new(0, Column(0)), Side::Right);
    assert!(selection.is_empty());
    selection.update(Point::new(0, Column(1)), Side::Left);
    assert!(selection.is_empty());
    selection.update(Point::new(0, Column(1)), Side::Right);
    assert!(!selection.is_empty());
    selection.update(Point::new(1, Column(0)), Side::Right);
    assert!(selection.is_empty());
    selection.update(Point::new(1, Column(1)), Side::Left);
    assert!(selection.is_empty());
    selection.update(Point::new(1, Column(1)), Side::Right);
    assert!(!selection.is_empty());
}
```

chrisduerr:
Ah, so looks like the problem is pretty straight-forward.

The current code just checks if it's either the exact same cell, or if it's opposing cells with the correct sides.

My fix just removed the check for this secondary check to be on the same line, but there's another scenario where block selections can be empty. They can be the same side, but just different lines.

I won't be able to get to this until friday, but it should be a trivial fix. Thanks a ton for these tests @nixpulvis.

chrisduerr:
@nixpulvis Think I fixed it. Let me know if you can find any other issues.

alacritty_terminal/src/selection.rs
```diff
@@ -144,9 +144,19 @@ impl Selection {
                 start == end
                     || (start.side == Side::Left
                         && end.side == Side::Right
-                        && start.point.line == end.point.line
+                        && (start.point.line == end.point.line)
                         && start.point.col == end.point.col + 1)
             },
+            Selection::Block { region: Range { ref start, ref end } } => {
+                // Empty when at most line different or two cells with right+left sides
```
nixpulvis: I'm not really sure what this comment means. What is "line" referring to here exactly? Also, I think we could be more clear about the right+left logic. Would you mind rewording this a little bit?

chrisduerr: I think this is the best way to word this shortly without getting into too much detail. If it takes longer to read the comment than the code, there's not much point either.

And I'm not sure why you're asking about what `line` refers to here? Every point has a column and a line, which other interpretation of `line` would make sense here?

nixpulvis: Yea, every point has a line, but you don't say which point's line you're talking about. In fact I don't see a reference to `line` anywhere below this comment, so it's somewhat unclear what this means to me.

> If it takes longer to read the comment than the code, there's not much point either.

I don't agree with this, and I believe it's the wrong way to think about comments. You should read comments and code together to get a complete understanding. If the code is clear enough on it's own, then there's no reason for a comment, but on the other hand some code requires vastly more comments than it's own syntax.

chrisduerr: > Yea, every point has a line, but you don't say which point's line you're talking about. In fact I don't see a reference to line anywhere below this comment, so it's somewhat unclear what this means to me.

Because it doesn't matter which line. Because it's a block selection and line number is completely irrelevant for detecting if it's empty. That's exactly what this comment says.

> If the code is clear enough on it's own, then there's no reason for a comment

Fyi, I only added a comment because the other codepath already had one. I don't think there's much that needs to be explained here, so I'd be fine with just removing it.

nixpulvis: I'd just remove this comment then. But if you want to keep it it that's fine too. It just confused me a bit.

chrisduerr: No worries, any idea for improvements? Maybe there's something better.

Issue #565, opened 2017-05-13
support for background images


OliverLeitner:
hi

id love to see background image support...

something like "rxvt*pixmap: /path/to/image.jpg"

is this in the plannings?

anyways, thanks for this great approach to a new generation of terminals=)

jwilm:
Probably not going to support this, but we do have plans for making the window translucent.

Radagan:
I wish you would reconsider this decision. It adds a lot to a terminal window and distinguishes it from the background in a way transparency does not.

Catman155:
I would also like this feature

ColdSauce:
I would really love this feature, as well.

KhaimovMR:
Same. Please reconsider.

i30817:
It'd be rad on a login shell ;)

chrisduerr:
Adding a background image might be something that could be reconsidered, however playing videos in the background will definitely not be. That's just too much work for a feature which is not going to be used by anyone and hardly makes much sense for a terminal emulator.

Of course some crazy ideas are fun to play around and experiment with, however I don't see something like that ever landing in Alacritty. :)

i30817:
Depressingly conservative, as is the unix tradition.

nixpulvis:
I wouldn't say "depressingly conservative". If Alacritty was closed source maybe... but there's nothing stopping a motivated member from adding this feature to their own fork. Don't be depressed :D

As for the goals of this community and mainline project; we are more focused on performance and minimalism, while still being a feature complete, working terminal emulator.

If the community truly wants this, I'd expect to see prototypes and forks pop up long before they are merged into master here.

Just my 2 cents.

pencil:
@chrisduerr Would a PR for this feature be accepted?

chrisduerr:
Honestly, I think at this point there's no way to implement this in a nice way where I'd be inclined to accept it.

It's possible that when support for something like libsixel is implemented, that it would be so trivial to implement this that it wouldn't be a lot of work to maintain, which would make it fine for me.

But generally I don't want to introduce any additional codepaths and rendering modes to support this and Alacritty currently has no way to render anything other than glyphs and rectangles.

Cyberlane:
I am experiencing ghosting when I have the `background_opacity` set, which is why I started to look for a way to set an image on the terminal itself.

sinistersnare:
One of the reasons for the closing of #2798 was that there wasnt enough interest in the feature.

[...]

PR #2798, opened 2019-09-14
Add background image

tamamu:
Add background image with a new configuration option `background_image`.
If you would like to try on it, add a new row in your `alacritty.yml` to set the path to the image, like as:
```yaml
background_image: /home/tamamu/Pictures/background.jpg
```
PNG is also ok.

There is a certain demand of this feature.
See #565

Screenshot from 2019-09-15 01-24-06


chrisduerr:
Thanks for the pull request, however as mentioned in #565, I do not think this aligns with Alacritty's goals. While it's always nice to have options and customization, I do not think it's worth the maintenance burden and overhead.

nixpulvis:
Thanks for your interest in Alacritty, and for taking the time to write this feature.

Some of us have talked a bit about this (before, and now again), and we're leaning towards not adding this feature to mainline Alacritty. I know that there some users who would like to see this added, but it's also not "make it or break it" kind of thing in our eyes. Alacritty is a pretty minimal terminal emulator, and we don't always add things just because we can.

Our concerns are specifically:

1. Feature bloat (what next, video backgrounds?)
2. Extra maintenance burden
3. Potential availability via the transparency feature

I know not everyone will agree with this decision, but we believe it's in the best interest of the project as a whole. I hope you understand.

I do encourage you to maintain this branch and use it as you need if you so desire.

aswinpajayan:
Is there a feature branch with this modification

aswinpajayan:
This feature would have been really nice

aswinpajayan:
can you give build instructions

aswinpajayan:
i was not able to get the feature when build from the sources

tamamu:
@aswinpajayan 

The branch is no longer maintained but you should be able to build it as below.

```
git clone -b background-image https://github.com/tamamu/alacritty; cd alacritty
cargo build --release
```

then edit your alacritty.yml to set the path to the background image.

Thanks!

casserlyprogramming:
>  did not show a significant amount of interest for this feature.

The 29 comments on the original issue, a follow up issue and this PR isn't enough interest from the community?

chrisduerr:
29 comments isn't a lot, no. And you conveniently left out the part where I mentioned that the lack of interest is just an additional hurdle on top of not fitting in with Alacritty's goals.

PR #2802, opened 2019-09-15
Fix bindings incorrectly getting replaced

### LINKED ISSUE #2794: Key binding mode replacement consistency
This is a follow-up to https://github.com/jwilm/alacritty/pull/2642.

Currently the idea is that whenever a user defines a binding with a bigger scope `mode`, that the default binding is overwritten. The idea behind this is that the user's binding will always work for all scopes he has defined, while default bindings are left untouched as much as possible.

This already works well when staying either only in positive, or in negative modes, but as soon as modes are mixed like `Alt` and `~Alt` things get messed up. Overwriting a default binding by only overwriting the `~Alt` one for example, will currently also disable the `Alt` binding. This is inconsistent with what would be expected by the user, since he has only requested a change in Alacritty's behavior in this specific mode.

I think to make sure that we do not create a regression in another area by fixing this, it's probably useful to lay out a table with all possible relevant combinations and creating dedicated tests for them to assure that the specification and implementation match up properly this time.

chrisduerr:
Fixes #2794.

### DIFF
--- a/alacritty_terminal/src/input.rs
+++ b/alacritty_terminal/src/input.rs
@@ -198,10 +198,21 @@ impl<T: Eq> Binding<T> {
 
     #[inline]
     pub fn triggers_match(&self, binding: &Binding<T>) -> bool {
-        self.trigger == binding.trigger
-            && self.mods == binding.mods
-            && (self.mode.contains(binding.mode) || binding.mode.contains(self.mode))
-            && (self.notmode.contains(binding.notmode) || binding.notmode.contains(self.notmode))
+        // Check the binding's key and modifiers
+        if self.trigger != binding.trigger || self.mods != binding.mods {
+            return false;
+        }
+
+        // Completely empty modes match all modes
+        if (self.mode.is_empty() && self.notmode.is_empty())
+            || (binding.mode.is_empty() && binding.notmode.is_empty())
+        {
+            return true;
+        }
+
+        // Check for intersection (equality is required since empty does not intersect itself)
+        (self.mode == binding.mode || self.mode.intersects(binding.mode))
+            && (self.notmode == binding.notmode || self.notmode.intersects(binding.notmode))
     }
 }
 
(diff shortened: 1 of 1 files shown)


alacritty_terminal/src/input.rs
```diff
@@ -198,10 +198,21 @@ impl<T: Eq> Binding<T> {
...
     #[inline]
     pub fn triggers_match(&self, binding: &Binding<T>) -> bool {
-        self.trigger == binding.trigger
-            && self.mods == binding.mods
-            && (self.mode.contains(binding.mode) || binding.mode.contains(self.mode))
-            && (self.notmode.contains(binding.notmode) || binding.notmode.contains(self.notmode))
+        // Check the binding's key and modifiers
+        if self.trigger != binding.trigger || self.mods != binding.mods {
+            return false;
+        }
```
nixpulvis: Just a small thing, but I think you could apply good ol' De Morgan's law to make this a bit more readable.

```
if self.trigger == binding.trigger && self.mods == binding.mods
```

This is also more similar to what we had written before.

chrisduerr: Can't early return if we do it that way though. That's the only reason why I've done it like this, otherwise you're 100% correct.

I thought it would be more readable to split this up into separate conditionals with early return statements than having just one giant boolean statement and trying to add comments to that.

alacritty_terminal/src/input.rs
```diff
@@ -198,10 +198,21 @@ impl<T: Eq> Binding<T> {
...
+        // Completely empty modes match all modes
+        if (self.mode.is_empty() && self.notmode.is_empty())
+            || (binding.mode.is_empty() && binding.notmode.is_empty())
+        {
+            return true;
+        }
+
+        // Check for intersection (equality is required since empty does not intersect itself)
+        (self.mode == binding.mode || self.mode.intersects(binding.mode))
+            && (self.notmode == binding.notmode || self.notmode.intersects(binding.notmode))
```
nixpulvis: Depending on how you feel, we might be able to make this a bit clearer by breaking this into a new helper function: `modes_match`, which would deal with both of these last two chained conditionals. Just might be a little nicer since the first conditional is just about the keyboard input, while the second two are about the terminal's mode.

chrisduerr: Problem is that this is just something we want to use for this specific function. In the other areas where we match the different modes we do things differently, which is why I haven't done this yet.

It's definitely a bit confusing with how contains and intersects work, but I feel like this might be better than a bigger separation.

alacritty_terminal/src/input.rs
nixpulvis: I mean, it still passes all the tests and is logically equivalent (see [De Morgan's Laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws).

I guess I forgot to add that you need to early return `true` and not `false`, but it's still possible.

chrisduerr: Generally this is even linted by clippy, but this is not accurate with an early return. So changing it would cause test failure.

The only alternative would be the following:
```rust
!(self.trigger == binding.trigger && self.mods == binding.mods)
```

And I don't think the introduction of brackets here is beneficial (it makes the whole thing a bit harder to read, since you need to read it inside-out).

nixpulvis: Yea, sorry for the noise... Today's a mess for me.

PR #2808, opened 2019-09-19
Initialize only visible characters

chrisduerr:
This fixes an off-by-two error in the renderer which initializes
characters 32 until 128 (inclusive) for each font whenever it is loaded.
The ascii visible range however just goes from 32 until 126 (inclusive).


nixpulvis:
ASCII is the best! :tada:

PR #2809, opened 2019-09-20
Remove outdated TODO/FIXME comments

chrisduerr:


alacritty_terminal/src/term/cell.rs
```diff
@@ -26,6 +26,7 @@ bitflags! {
         const INVERSE           = 0b00_0000_0001;
         const BOLD              = 0b00_0000_0010;
         const ITALIC            = 0b00_0000_0100;
+        const BOLD_ITALIC       = 0b00_0000_0110;
```
nixpulvis: :thinking: nice.

alacritty_terminal/src/renderer/mod.rs
```diff
@@ -1026,19 +1027,15 @@ impl<'a> RenderApi<'a> {
...
-        let font_key = match (
-            cell.flags.contains(cell::Flags::BOLD),
-            cell.flags.contains(cell::Flags::ITALIC),
-        ) {
-            (false, false) => glyph_cache.font_key,
-            (true, false) => glyph_cache.bold_key,
-            (false, true) => glyph_cache.italic_key,
-            (true, true) => glyph_cache.bold_italic_key,
+        let font_key = match cell.flags | (Flags::BOLD & Flags::ITALIC) {
+            Flags::BOLD_ITALIC => glyph_cache.bold_italic_key,
```
dnkl: Shouldn't the match expression be `cell.flags & (Flags::BOLD | Flags::ITALIC)`?

dnkl: Or simply `cell.flags & Flags::BOLD_ITALIC`

chrisduerr: Completely correct, looks like I've twisted things around. I haven't used the `BOLD_ITALIC` flag since I added that later, but now that's also a great suggestion.

PR #2810, opened 2019-09-20
Remove bindings with action `None`

zsugabubus:
Currently, there is no option to disable default key bindings while keep receiving pressed keys. It's because specifying action `None` is actually a no-op and does not result in an unset binding. This pull request aims to overcome this limitation.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -496,8 +496,8 @@ mouse_bindings:
 #   - ToggleFullscreen
 #   - SpawnNewInstance
 #   - ClearLogNotice
-#   - None
 #   - ReceiveChar
+#   - None
 #
 #   (macOS only):
 #   - ToggleSimpleFullscreen: Enters fullscreen without occupying another space
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -497,6 +497,7 @@ mouse_bindings:
 #   - SpawnNewInstance
 #   - ClearLogNotice
 #   - None
+#   - ReceiveChar
 #
 #   (macOS only):
 #   - ToggleSimpleFullscreen: Enters fullscreen without occupying another space
@@ -536,7 +537,8 @@ mouse_bindings:
 #
 # Bindings are always filled by default, but will be replaced when a new
 # binding with the same triggers is defined. To unset a default binding, it can
-# be mapped to the `None` action.
+# be mapped to the `ReceiveChar` action. Alternatively, you can use `None` for
+# a no-op if you do not wish to receive input characters for that binding.
 key_bindings:
   # (Windows/Linux only)
   #- { key: V,        mods: Control|Shift, action: Paste            }
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -269,6 +269,7 @@ where
 {
     let mut bindings: Vec<Binding<T>> = failure_default(deserializer)?;
 
+    // Remove matching default bindings
     for binding in bindings.iter() {
         default.retain(|b| !b.triggers_match(binding));
     }
--- a/alacritty_terminal/src/input.rs
+++ b/alacritty_terminal/src/input.rs
@@ -219,7 +219,7 @@ impl<T: Eq> Binding<T> {
 impl<T> Binding<T> {
     /// Execute the action associate with this binding
     #[inline]
-    fn execute<A: ActionContext>(&self, ctx: &mut A, mouse_mode: bool) {
+    fn execute<A: ActionContext>(&self, ctx: &mut A, mouse_mode: bool) -> bool {
         self.action.execute(ctx, mouse_mode)
     }
 
@@ -236,60 +236,60 @@ impl<T> Binding<T> {
 
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
 pub enum Action {
-    /// Write an escape sequence
+    /// Write an escape sequence.
     #[serde(skip)]
     Esc(String),
 
-    /// Paste contents of system clipboard
+    /// Paste contents of system clipboard.
     Paste,
 
-    // Store current selection into clipboard
+    /// Store current selection into clipboard.
     Copy,
 
-    /// Paste contents of selection buffer
+    /// Paste contents of selection buffer.
     PasteSelection,
 
-    /// Increase font size
+    /// Increase font size.
     IncreaseFontSize,
 
-    /// Decrease font size
+    /// Decrease font size.
     DecreaseFontSize,
 
-    /// Reset font size to the config value
+    /// Reset font size to the config value.
     ResetFontSize,
 
-    /// Scroll exactly one page up
+    /// Scroll exactly one page up.
     ScrollPageUp,
 
-    /// Scroll exactly one page down
+    /// Scroll exactly one page down.
     ScrollPageDown,
 
-    /// Scroll one line up
+    /// Scroll one line up.
     ScrollLineUp,
 
-    /// Scroll one line down
+    /// Scroll one line down.
     ScrollLineDown,
 
-    /// Scroll all the way to the top
+    /// Scroll all the way to the top.
     ScrollToTop,
 
-    /// Scroll all the way to the bottom
+    /// Scroll all the way to the bottom.
     ScrollToBottom,
 
-    /// Clear the display buffer(s) to remove history
+    /// Clear the display buffer(s) to remove history.
     ClearHistory,
 
-    /// Run given command
+    /// Run given command.
     #[serde(skip)]
     Command(String, Vec<String>),
 
-    /// Hides the Alacritty window
+    /// Hide the Alacritty window.
     Hide,
 
-    /// Quits Alacritty.
+    /// Quit Alacritty.
     Quit,
 
-    /// Clears warning and error notices.
+    /// Clear warning and error notices.
     ClearLogNotice,
 
     /// Spawn a new instance of Alacritty.
@@ -304,6 +304,9 @@ pub enum Action {
 
     /// No action.
     None,
+
+    /// Allow receiving char input.
+    ReceiveChar,
 }
 
 impl Default for Action {
@@ -314,7 +317,7 @@ impl Default for Action {
 
 impl Action {
     #[inline]
-    fn execute<A: ActionContext>(&self, ctx: &mut A, mouse_mode: bool) {
+    fn execute<A: ActionContext>(&self, ctx: &mut A, mouse_mode: bool) -> bool {
         match *self {
             Action::Esc(ref s) => {
                 ctx.scroll(Scroll::Bottom);
@@ -392,7 +395,10 @@ impl Action {
                 ctx.spawn_new_instance();
             },
             Action::None => (),
+            Action::ReceiveChar => return false
         }
+        // Suppress input by default
+        true
     }
 
     fn paste<A: ActionContext>(&self, ctx: &mut A, contents: &str) {
@@ -716,7 +722,7 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         self.copy_selection();
     }
 
-    // Spawn URL launcher when clicking on URLs
+    /// Spawn URL launcher when clicking on URLs.
     fn launch_url(&self, url: Url) {
         if self.ctx.mouse().block_url_launcher {
             return;
@@ -846,7 +852,7 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         self.ctx.mouse_mut().last_button = button;
     }
 
-    /// Process key input
+    /// Process key input.
     pub fn process_key(&mut self, input: KeyboardInput) {
         self.ctx.modifiers().update(input);
 
@@ -864,17 +870,13 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         match input.state {
             ElementState::Pressed => {
                 *self.ctx.received_count() = 0;
-                *self.ctx.suppress_chars() = false;
-
-                if self.process_key_bindings(input) {
-                    *self.ctx.suppress_chars() = true;
-                }
+                *self.ctx.suppress_chars() = self.process_key_bindings(input);
             },
             ElementState::Released => *self.ctx.suppress_chars() = false,
         }
     }
 
-    /// Process a received character
+    /// Process a received character.
     pub fn received_char(&mut self, c: char) {
         if *self.ctx.suppress_chars() {
             return;
@@ -903,15 +905,14 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         *self.ctx.received_count() += 1;
     }
 
-    /// Attempts to find a binding and execute its action
+    /// Attempt to find a binding and execute its action.
     ///
     /// The provided mode, mods, and key must match what is allowed by a binding
     /// for its action to be executed.
     ///
-    /// Returns true if an action is executed.
+    /// Return true if chars should be suppressed.
     fn process_key_bindings(&mut self, input: KeyboardInput) -> bool {
-        let mut has_binding = false;
-        for binding in self.key_bindings {
+        self.key_bindings.iter().filter_map(|binding| {
             let is_triggered = match binding.trigger {
                 Key::Scancode(_) => binding.is_triggered_by(
                     *self.ctx.terminal().mode(),
@@ -935,21 +936,27 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
             };
 
             if is_triggered {
-                // binding was triggered; run the action
-                binding.execute(&mut self.ctx, false);
-                has_binding = true;
+                // Binding was triggered; run the action
+                Some(binding.execute(&mut self.ctx, false))
+            } else {
+                None
             }
-        }
-
-        has_binding
+        }).fold(None, |should_suppress_char, suppress_char| {
+            // Suppress char by default as soon as a binding has been triggered
+            let should_suppress_char = should_suppress_char.unwrap_or(true);
+            Some(should_suppress_char && suppress_char)
+        }).unwrap_or_else(|| {
+            // Do not suppress char if no bindings were triggered
+            false
+        })
     }
 
-    /// Attempts to find a binding and execute its action
+    /// Attempt to find a binding and execute its action.
     ///
     /// The provided mode, mods, and key must match what is allowed by a binding
     /// for its action to be executed.
     ///
-    /// Returns true if an action is executed.
+    /// Return true if an action is executed.
     fn process_mouse_bindings(&mut self, mods: ModifiersState, button: MouseButton) -> bool {
         let mut has_binding = false;
         for binding in self.mouse_bindings {
--- a/alacritty_terminal/src/input.rs
+++ b/alacritty_terminal/src/input.rs
@@ -930,7 +930,10 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
                 // Don't suppress when there has been a `ReceiveChar` action
                 Some(suppress_chars.unwrap_or(true) && binding.action != Action::ReceiveChar)
             })
-            .unwrap_or(false);
+            .unwrap_or_else(|| {
+                // Do not suppress char if no bindings were triggered
+                false
+            });
     }
 
     /// Attempt to find a binding and execute its action.
--- a/alacritty_terminal/src/input.rs
+++ b/alacritty_terminal/src/input.rs
@@ -219,7 +219,7 @@ impl<T: Eq> Binding<T> {
 impl<T> Binding<T> {
     /// Execute the action associate with this binding
     #[inline]
-    fn execute<A: ActionContext>(&self, ctx: &mut A, mouse_mode: bool) -> bool {
+    fn execute<A: ActionContext>(&self, ctx: &mut A, mouse_mode: bool) {
         self.action.execute(ctx, mouse_mode)
     }
 
@@ -302,11 +302,11 @@ pub enum Action {
     #[cfg(target_os = "macos")]
     ToggleSimpleFullscreen,
 
-    /// No action.
-    None,
-
     /// Allow receiving char input.
     ReceiveChar,
+
+    /// No action.
+    None,
 }
 
 impl Default for Action {
@@ -317,7 +317,7 @@ impl Default for Action {
 
 impl Action {
     #[inline]
-    fn execute<A: ActionContext>(&self, ctx: &mut A, mouse_mode: bool) -> bool {
+    fn execute<A: ActionContext>(&self, ctx: &mut A, mouse_mode: bool) {
         match *self {
             Action::Esc(ref s) => {
                 ctx.scroll(Scroll::Bottom);
@@ -394,11 +394,8 @@ impl Action {
             Action::SpawnNewInstance => {
                 ctx.spawn_new_instance();
             },
-            Action::None => (),
-            Action::ReceiveChar => return false
+            Action::ReceiveChar | Action::None => (),
         }
-        // Suppress input by default
-        true
     }
 
     fn paste<A: ActionContext>(&self, ctx: &mut A, contents: &str) {
@@ -870,7 +867,7 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
         match input.state {
             ElementState::Pressed => {
                 *self.ctx.received_count() = 0;
-                *self.ctx.suppress_chars() = self.process_key_bindings(input);
+                self.process_key_bindings(input);
             },
             ElementState::Released => *self.ctx.suppress_chars() = false,
         }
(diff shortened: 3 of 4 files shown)


zsugabubus:
> The goal of None action is to provide nop, isn't it?

From "alacritty.yml":
> To unset a default binding, it can be mapped to the `None` action.

In my interpretation, if I unset such a binding I don't set a no-op, right? For me also doesn't obvious why anybody would like to set a no-op keybinding. If you don't want receive those keys just don't press them.

> If you still want to receive pressed keys, you can do a proper rebind with chars. 

I did it, but it would be cleaner if I could use `None`(?) as action. Or maybe all hard-coded "platform-specific" bindings could be removed and necessary things could be shipped inside alacritty.yml. This way one could avoid such "surprise bindings".

chrisduerr:
@zsugabubus Which binding did you want to unbind?

And the problem with this approach is that there's no good way to map a binding to a noop anymore. And the `NONE` action would be much more surprising to act as a reset, since the name itself already says that it binds a key to an explicit None action.

zsugabubus:
@chrisduerr I wanted to get rid of every `{Increase,Decrease,Reset}FontSize` because I don't use them and they are bound to Control+something that I use too frequently. My specific issue was with `{ key: Key0, mods: Control, action: None }` because it's `Control-]` on my keyboard (I'm absolutely not sure and I don't know how, but as I remember it just went wrong lately, because as I can tell it worked before. Maybe there was an update?).

Although I managed to solved it with "chars" it would be simpler to just unmap it.

chrisduerr:
Hmm maybe a better solution would be to introduce a separate binding action with a more appropriate name? I have no clue how I would name that, but my biggest issue with this is that the action `None` actually does something.

kchibisov:
I think adding action `Unset` (`Unbind`) and keeping `None` for compatibility would make sense. What do you think?

zsugabubus:
And what's your opinion about removing platform-specific keybindings completely and move them to user's config file?  `Unset` would have only one use case, to unset such a platform-specific keybinding.

kchibisov:
I don't like the idea of removing these options.

First of all it's better for cross platform use scenario, so I can have `alacrtitty.yml` only with color scheme and `zero` bindings section in it and use alacritty on Linux/macOS/Windows system without adding options to config file, which tend to break things.

Secondly, These bindings are kind of default in all terminal apps on macOS and Copy/Paste follows the system Copy/Paste, so forcing users to explicitly uncomment things to make native expected bindings work is a bad practice in my opinion.

So, `Unset` is just better here, even if it's purpose is just to handle these few bindings. We can probably add more in a future though.

chrisduerr:
> First of all it's better for cross platform use

Yeah, this is one of the primary reasons why this is done in code right now. We should have a configuration file that works appropriately on all platforms without any conflicts. Moving all things into the config file only, will not allow for this.

alacritty_terminal/src/input.rs
```diff
@@ -283,27 +283,33 @@ pub enum Action {
...
+    /// Toggle fullscreen
     ToggleFullscreen,
 
-    /// Toggle simple fullscreen on macos.
+    /// Toggle simple fullscreen on macos
     #[cfg(target_os = "macos")]
     ToggleSimpleFullscreen,
 
-    /// No action.
+    /// No action
```
chrisduerr: I'm not against unifying the documentation, that's actually much appreciated. But the canonical format for doc comments in Rust is to have the punctuation both in the description and the headline. So that should be kept.

alacritty_terminal/src/input.rs
```diff
@@ -392,6 +398,7 @@ impl Action {
                 ctx.spawn_new_instance();
             },
             Action::None => (),
+            Action::Unset => unreachable!(),
```
chrisduerr: I'm not sure if I `unreachable` is the correct choice here, but I think it does make some sense and should help us detect errors early. Though if we don't detect them, it's possible that the effects are more severe.

In any case, if we do choose to use `unreachable!` here, I'd recommend passing at least a short message to it to make it absolutely clear that it's intentional. That will also make debugging trivial even without any backtraces.

zsugabubus:
My last thoughts @kchibisov.

> I can have alacrtitty.yml only with color scheme and zero bindings section in it and use alacritty on Linux/macOS/Windows system without adding options to config file

You can enjoy the cross-platform nature of the config file up until you don't start adding your own custom "platform-specific" bindings. So it's a config file that you'd better not to modify.

I would also note that most of (if not all) non-qwert{y,z}/non-english users unable to use some of the default keybindings because they have to press Shift key for symbols but Alacritty expects only Control. So these users have to add their own non-cross-platform bindings, making whole config file non-cross-platform.

One other convenient advantage of moving these bindings to the config file would be that users could easily look them up if they forget them.

If it would be up to me then I would add `platform` key, or extend configuration file to allow including files.

> forcing users to explicitly uncomment things to make native expected bindings work is a bad practice in my opinion.

Package maintainers could ship Alacritty with platform-specific keybindings uncommented by default. "alacritty.yml" only populated once at the first run and it never modified afterwards by updates.

> which tend to break things

Why? Removing or worst adding a hard-coded binding won't break things? If a new binding appears, it potentially breaks every single (terminal) application of user. If there are no hard-coded bindings only permanently removed actions cause breaking changes (that breaking change anyway).

alacritty_terminal/src/input.rs
zsugabubus: Eh, sorry. There were no dots for the most of the lines at the start, so I thought that's the standard.

chrisduerr:
@zsugabubus fyi I use Alacritty on multiple platforms with a non-us keyboard myself and I do not have any custom keybindings on any of my systems. So the system currently in place works extremely well in my experience and makes it super easy to have a cross-platform config file.

Adding a `platform` key would just create a giant mess. At that point you're likely better off maintaining your dotfiles with separate branches, which is likely necessary for different systems anyways.

> Package maintainers could ship Alacritty with platform-specific keybindings uncommented by default. "alacritty.yml" only populated once at the first run and it never modified afterwards by updates.

We cannot rely on this at all and package maintainers cannot ship files to the home directory.

> If a new binding appears, it potentially breaks every single (terminal) application of user.

No. If a new binding appears, it fixes all Alacritty instances without requiring manual intervention unless it's explicitly overwritten by the user. The default bindings are mainly used for basic terminal functionality, we don't ship anything crazy that would break applications.

alacritty_terminal/src/input.rs
zsugabubus: `Unset` cannot be added by users, because they are removed (by definition). If that code is reached it can "only happen" if it comes from a hard-coded binding that's absolutely useless.

chrisduerr: I'm aware of that. It still adds potential for a crash when that is absolutely not necessary. If we ever hit this, we could handle this like the `None` action without any issues.

zsugabubus:
> multiple platforms with a non-us keyboard

What layout do you use? I'm unable to zoom in on [Hungarian layout](https://upload.wikimedia.org/wikipedia/commons/a/a8/KB_Hungary.svg) with defaults.

> it fixes all Alacritty instances without requiring manual intervention

To clarify myself, I meant `zsh/bash/vim/mutt...` under terminal applications.

> We cannot rely on this at all and package maintainers cannot ship files to the home directory.

I thought about a `/etc` fallback file.

> we don't ship anything crazy that would break applications.

It don't have to be a crazy thing. Every key you make a hard-coded binding for it could be also used by an application. See my scenario, for example.

kchibisov:
Absolutely agree with @chrisduerr , I was using macOS in the past(Russian keyboard layout) and was running only with my own crossplatform bindings. The thing that you're describing is likely a bug which is already tracked somewhere and probably will be addressed in a future in our dependency iirc.

alacritty_terminal/src/input.rs
zsugabubus: So then an error message for `uncreachable!` is enough? Or choose "`()`"?

chrisduerr:
> Every key you make a hard-coded binding for it could be also used by an application.

That's not accurate. Most bindings we have is explicitly because we want to let applications know about the pressed keys. We have extremely few other bindings that do not use command/super as modifiers, which is not passed to terminals anyways.

zsugabubus:
I meant "platform-specific" bindings as always, just omitted because I thought it's obvious from history.

alacritty_terminal/src/input.rs
chrisduerr: I'd choose an error message for `unreachable!` and rely on people reporting the crash if it should ever happen. It's an easy fix if it does, so it's not a bad idea to know about it.

alacritty_terminal/src/config/mod.rs
```diff
@@ -269,9 +269,11 @@ where
 {
     let mut bindings: Vec<Binding<T>> = failure_default(deserializer)?;
 
-    for binding in bindings.iter() {
+    bindings.retain(|binding| {
+        // Remove matching default bindings.
```
kchibisov: No dot at the end here

alacritty_terminal/src/input.rs
```diff
@@ -304,6 +304,12 @@ pub enum Action {
 
     /// No action.
     None,
+
+    /// Mark binding as unset.
+    ///
+    /// It is just a meta action that will be never received. Bindings with this action will be
+    /// removed shortly after deserialization.
+    Unset,
```
nixpulvis: The more I think about this the more I dislike adding a new "meta"-action. For a few reasons:

1. A new `Unset` meta-action cannot (and should not) ever be executed, making it weird and special
2. The current `None` action's utility remains unknown to me
3. We're trying to communicate something which isn't communicated in the distinction between `None` and `Unset` (at least to me)

Given the historic documentation's wording around `None`, and the lack of a proper use case for a NOOP, I think I'd simply argue for `None` to be re-purposed.

However, this doesn't completely address my first point, namely that we're providing an action which should never be called. I'd like to avoid needing two data structures for this (`ConfigAction` and `Action`), so how might we be able to do this? Allowing a config entry like `{ key: X mods: Shift }` without an action altogether might be an option. Or allow passing an explicit "intrinsic" action, like how "default" here means the baseline (intrinsic) action of `receive_char`, plus any default actions. Something like `{ key: X, mods: Shift, action: ReceiveChar }`. This is awkward, because it's unpredictable. User's aren't going to guess to call a `ReceiveChar` action, although, neither will they guess to call an `Unset` action.

Long story short... can we avoid this meta-action. I'd be a bit happier if we could.

chrisduerr: > I think I'd simply argue for None to be re-purposed

I think repurposing an existing thing and making it do something completely different is about the worst approach here.

We can't just completely break things without any warnings. There's a good chance there are people out there relying on this and it would be extremely annoying to just make something do something entirely different from one day to another.

Renaming this binding to something different at least gives us the option to deprecate one in the future without breaking user configurations today. I see no reason why we would so actively aim to break things when there's a clear and simple way to resolve this without any breakage.

> I view action names in the config as the name of the behavior I get when I activate the binding from input to Alacritty, not as actions being preformed on the configuration itself.

I agree with this and have never been a huge fan of the different choices for actions proposed here, though I do not have a better proposal myself.

> Long story short... can we avoid this meta-action? I'd be a bit happier if we could.

Fundamentally I agree with this. The question is just how to do it.

As @zsugabubus has already correctly realized, the main problem here might not be how to correctly unbind things, but rather that we cannot have things in the default configuration as the only source of truth because we need to support multiple platforms.

The proposal to have a global `platform` key is also quite troublesome, because it would cause quite a mess, however I think it could be reasonable to have a `platform` key for keybindings (maybe @zsugabubus even meant that in his original comment?). That way we could have all bindings that are questionable and might need to be overwritten (the action bindings) in the default config.

I don't think there ever is any good reason to overwrite the `char` bindings that are used to sent escape sequences. But if someone has a different experience here I'd be very curious. And even then mapping to `chars: ""` should still be possible to overwrite the defaults.

zsugabubus: To be honest, I also don't like this meta action thing for the same reasons as you @nixpulvis.

If we want a nicer solution, I think, we can close out every possibilities when the user has to specify one-by-one which bindings she wants to unset. Alternatives:
* Introduce config option to disable builtin mappings -> adds unnecessary complexity + not cross-platform friendly.
* Automagically disable builtin mappings based on some kind of heuristics -> inconvenient when someone just wants to extend her bindings and want to keep others + it's too magical + still not cross-platform friendly.
* Make user responsible to maintain her cross-platform config, by placing every platform-specific bindings into user's config (see my previous comments why I consider it useful). I'm sure whoever would use Alacritty on multiple platforms has a complete solution how to maintain different configs across different platforms, since probably not Alacritty would be the only one that needs it. However it's likely more than inconvenient because only these few keybindings are platform-dependent.
* Alacritty handles cross-platform bindings, for example by a `platform` key  (per binding). It's like the previous, except users don't have to mess with different config files. Seems the simplest solution.

I'd vote for the last, but I don't have other just Linux systems, so I would happy with the last two.

Other ideas?

kchibisov: I'm against forcing users to add **common** platform bindings to make things work as expected. e.g. I want to run alacritty without default configuration file and everything should **work**, if it'll force me to create config and place some bindings there I'll be pretty angry, since you'll broke my contributor workflow. ( I run `alacritty --config-file /dev/null` a lot to test things).

zsugabubus: That's where system-wide config file comes in, shipped with Alacritty.

chrisduerr: Oh yeah that's a great point I completely forgot about myself again. I personally run a completely empty configuration file and I wouldn't want to need part of the default bindings to be required in the configuration file to have them active. That's just unnecessary complexity we're putting in the user config.

We also can't make it that completely empty bindings are different than bindings that are slightly filled, since that makes it extremely hard for the user to understand what's going on.

So I don't immediately see a solution for this which doesn't include having a binding which can override defaults. Though I'd love to hear any suggestions that take this into account.

zsugabubus: 1) You wouldn't have to create config file since if user-local doesn't exists Alacritty falls back to the system-wide.
2) If you would like to test font resizing than you could use a dummy config file with 5 lines. If you don't like it for some reason, then Alacritty could fall back to the system-wide `key_bindings` in such cases.

kchibisov: if user config doesn't exist alacritty generates one for it (this behavior will be removed in a near future). 

If you force `alacritty` to use empty config it'll read it and fallback to default options and not to system wide config. alacritty doesn't have a notion of system wide configuration atm.

kchibisov: In my opinion `Unset` is the least evil we have here. Keeping some file to make alacritty work is not an option (system-wide/user one, doesn't matter).

zsugabubus: > if user config doesn't exist alacritty generates one for it (this behavior will be removed in a near future)
> doesn't have a notion of system wide configuration atm

Cannot change these now?

kchibisov: Oh, and just in case we introduce some key to distinguish between platforms it won't solve the problem. It'll make alacritty.yml more portable if user decide to add more keybindings for each platform, but you'll need `Unset` action, since generating/relaying on external file is not an option.

zsugabubus: I don't get what you mean. Why would we need `Unset`?

> since generating/

Alacritty would have to do generate nothing. System-wide config file written once by the "installer", and user-local config only written by user.

> relaying on external file is not an option

We are talking about the config file, don't? Why we can't rely on config files as the source of configuration settings? Or what do you mean?

zsugabubus: ###### Why I see messages only later?

> Keeping some file to make alacritty work is not an option

1) Don't have to. User-local > System-wide > Hard-coded. 👍 
2) Anyway, if you delete Alacritty binary you don't have to be surprised if Alacritty doesn't start up next time. Why keep it? Alacritty comes with sensible defaults that are shipped with the config file.

kchibisov: > Alacritty would have to do generate nothing. System-wide config file written once by the "installer", and user-local config only written by user.

We can't create a system-wide config file, that is something distro package maintainers should do. We shouldn't rely on them here and I personally don't want to see `/etc/alacritty/alacritty.yml` file at some point, since it doesn't make sense to me.

> I don't get what you mean. Why would we need Unset?

I personally don't need it, since I'm fine with the current `None` action. However if we want to achieve the initial behavior proposed in your PR we need `Unset` thing ( or something similar ). 

>  Anyway, if you delete Alacritty binary you don't have to be surprised if Alacritty doesn't start up next time. Why keep it? Alacritty comes with sensible defaults that are shipped with the config file.

Why it shouldn't startup, I'm excepting application to work. That's why we have message bar to notify users of broken config, instead of crashing. If alacritty crashes on startup it's a bug, that should be fixed.

> Don't have to. User-local > System-wide > Hard-coded. +1

We already have a relationship User-local > Hard-coded, why do we need system-wide then? The problem is that None do nothing instead of sending chars.

chrisduerr: > System-wide config file written once by the "installer"

This is not an option. We cannot rely on system-wide configuration files. We need to be able to run on a system without any Alacritty configuration file whatsoever and it should still provide all the default key bindings. There's no way around this.

chrisduerr:
So I've discussed this a bit with @kchibisov and @nixpulvis on IRC and we came to the conclusion that the only real solution would be to have a new separate action to support this usecase.

However, the `Unset` action is not really ideal, since it acts a bit counter-intuitively. Instead the consensus was that a better choice would be to have a `ReceiveChar` action, which would have the effect that the character will be forwarded (kinda similar to `Unset`).

However, when having an already existing binding action like `ScrollToBottom` and a separate binding is added with the action `ReceiveChar`, it should result in both scrolling to the bottom of the terminal and printing the character itself.

I hope this makes this a bit more functional and clear for the users, if I've forgotten anything please let me know @nixpulvis. Let me know if you're interested in implementing this @zsugabubus, there have been a lot of changes unfortunately, so I'm glad to help out with implementation if this has gone beyond what you're willing to invest into this PR.

nixpulvis:
Just to put things in YAML explicitly, here's how I'm imagining a few use cases:

```yaml
key_bindings:
# Overwrite ctrl-l default action(s) with an explict `ReceiveChar` action.
#
# You can think of `ReceiveChar` as the action for all sets of key+mod+mode not
# set as bindings either in the configuration, or defaults.
#
# This effectively disables default key bindings while keeping pressed keys received
# by the terminal.
- { key: L, mods: Control, action: ReceiveChar }

# Do nothing at all when 'x' is pressed.
- { key: X, action: None }
# Re-enable the terminal's receiving characters 'x'.
- { key: X, action: ReceiveChar }
```

zsugabubus:
Made modifications on [my local branch](https://github.com/zsugabubus/alacritty/tree/introduce-receivechar-action). Do not know if you like it or how you imagined it should be implemented, but if you're satisfied with it I can push it to here or create a new pull request.

alacritty_terminal/src/input.rs
zsugabubus: Okay. Made modifications on another branch. See my comment below.

chrisduerr:
@zsugabubus I'd say the code does exactly what I would have expected. So I'd recommend pushing it to this PR so I can give the code itself a proper review.

One thing I've noticed is that multiple `ReceiveChar` bindings do not actually receive the raw character multiple times, though I think we can accept that trade-off. Other than that I haven't found any issues.

nixpulvis:
Yea I'm unsure what the best functionality for multiple `ReceiveChar` actions should be.

I think ideally it would be a warning to the user. I can't imagine ever wanting to do this personally... am I forgetting an obvious use case?

Leaving things as is would also be fine, I think, since I don't expect users to have additional trouble with this. They may just open an issue "can't use multiple `ReceiveChar`s".

Basically we just need to decide if `ReceiveChar` *should* be allowed to be called twice on one binding. But I don't think we **need** to decide now either.

chrisduerr:
> They may just open an issue "can't use multiple ReceiveChars".

I'd honestly just wait for this and then evaluate if we want to solve this. I do not expect anyone will run into this and even if they do, chances are they still get the behavior they'd want. Someone who intentionally wants to receive the char twice is more likely to use `chars` I think.

It's also not a terrible way to fail, the binding just can't be used twice, it's super friendly about that since it's clear that it's not being sent twice and there's nothing it overwrites or anything. So even if people run into this I'd expect them to just do things differently and move on, since it's such a simple fix with `chars`.

alacritty_terminal/src/input.rs
```diff
@@ -935,21 +936,27 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
-
-        has_binding
+        }).fold(None, |should_suppress_char, suppress_char| {
+            // Suppress char by default as soon as a binding has been triggered
+            let should_suppress_char = should_suppress_char.unwrap_or(true);
+            Some(should_suppress_char && suppress_char)
+        }).unwrap_or_else(|| {
+            // Do not suppress char if no bindings were triggered
+            false
+        })
```
nixpulvis: Generally I like this approach, but I'm a little sad to see `process_key_bindings` and `process_mouse_bindings` diverging. This might be an argument for continuing to have `execute` set `ctx.suppress_chars()` itself.

Thoughts?

chrisduerr: Can't have execute call that itself, otherwise binding order matters.

LaGvidilo:
please, just can run on PPC ?

chrisduerr:
> please, just can run on PPC ?

I'm sorry, but I don't understand what you mean @LaGvidilo?

alacritty_terminal/src/input.rs
```diff
@@ -392,7 +395,10 @@ impl Action {
                 ctx.spawn_new_instance();
             },
             Action::None => (),
+            Action::ReceiveChar => return false
```
kchibisov: We should also swap these (None/ReceiveChar) match arms. To be consistent with already suggested changes on swapping things.

chrisduerr: With my suggested changes, these actions would just be merged together anyways.

kchibisov: I see, I've suggested it, since `return false` was presented in your diff. Anyway it should be `Action::ReceiveChar | Action::None => ...`  (not None | ReceiveChar) , right?

LaGvidilo:
i have a powermac G5 (PPC not intel) and I want Alacritty for my Mac OS X
Leopard.
It's possible or dependencies not compatible ?

Le jeu. 26 sept. 2019 à 18:41, Christian Duerr <notifications@github.com> a
écrit :

kchibisov:
@LaGvidilo you should better open an issue, since you're writing on a PR that doesn't related to your problem and being offtopic.

Answering your question, I'm not sure if we're x86_64 specific the main problem is OpenGL 3.3 support on your system, no one stops you from trying compiling alacritty from source and seeing the result of it.

/mark me offtopic or delete

alacritty_terminal/src/input.rs
chrisduerr: Yeah, I'd go for that. Though I don't particularly care a lot about the order when it's in the same statement. That's a bit too nitpicky even for me. :)

alacritty_terminal/src/input.rs
```diff
@@ -314,7 +317,7 @@ impl Default for Action {
 
 impl Action {
     #[inline]
-    fn execute<A: ActionContext>(&self, ctx: &mut A, mouse_mode: bool) {
+    fn execute<A: ActionContext>(&self, ctx: &mut A, mouse_mode: bool) -> bool {
```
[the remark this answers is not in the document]
zsugabubus: I feel testing for `Action::ReceiveChar` in a different code section (different from core execution logic) more arbitrary. `ReceiveChar` is at the same level as every other action but it's raised from `execute` to a higher level `process_key_bindings`. My simple `bool` meant whether to allow or disallow suppressing keys (not completely true). To keep my comment short, just asking that using an enum maybe wouldn't be a nicer solution?

alacritty.yml
```diff
@@ -497,6 +497,7 @@ mouse_bindings:
 #   - SpawnNewInstance
 #   - ClearLogNotice
 #   - None
+#   - ReceiveChar
```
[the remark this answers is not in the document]
zsugabubus: Of course. (I didn't dare to modify the order.)

alacritty_terminal/src/input.rs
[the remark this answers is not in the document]
chrisduerr: > I feel testing for Action::ReceiveChar in a different code section (different from core execution logic) more arbitrary.

Execution logic is solely about executing actions. If we would print the char manually, that's where it would go. However we shouldn't instruct it to set something that should be done later. There's nothing executed when we're setting a flag.

> My simple bool meant whether to allow or disallow suppressing keys

Actions are not bound to keys. They're arbitrary actions that should be executed. It makes no sense to return if we want to suppress chars or not when executing mouse bindings. So this logic should be handled where it makes sense, at the place where we handle key bindings specifically.

> just asking that using an enum as execute() result maybe wouldn't be a nicer solution?

If execute would ever have any kind of return type, it would likely just indicate success or not. Anything else doesn't make sense.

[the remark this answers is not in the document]
zsugabubus: `ReceiveChar` once executes some code in `execute()`, than _independently_ from this it suppresses keys as another distinct side effect. These two are relating to each other therefore should be one unit. What came into my mind was, what if some day `ClearLogNotice` only wants to suppress keys only when really there is some log notice to clear, otherwise it let keys through. (Just now see that it's two distinct key actions.)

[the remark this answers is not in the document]
chrisduerr: > ReceiveChar once executes some code in execute(), than independently from this it suppresses keys as another distinct side effect

No it doesn't. `ReceiveChar` doesn't execute any code in the `execute` method

> What came into my mind was, what if some day ClearLogNotice only wants to suppress keys only when really there is some log notice to clear, otherwise it let keys through

First of all, that would be awful for the user when actions suddenly change behavior like that. Second of all, we already leave chars through with `ClearLogNotice`. Thirdly, if changes are required in the future, they should be made in the future, not now. And finally, even if `ClearLogNotice` would want to mingle with suppressing characters we could still do that in the key press handling where it isn't called from other functions.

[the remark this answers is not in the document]
nixpulvis: I'm pretty torn on this. There **is** a difference between mouse and keyboard actions in the sense that a keyboard action typically suppresses the underlying default action. This isn't relevant to the mouse. However, I feel like this boils down to an argument for functional style, pass value as bool "flag" in the return type, vs setting some state.

I'm generally a fan of functional styles, but I agree with @chrisduerr that `execute` seems like the wrong place. An `Action` shouldn't care about the surrounding processing in general, and I'd opt to handle `RecieveChar` in a special case within `process_key_bindings`. I haven't played with this in code, so hopefully it makes sense.

alacritty_terminal/src/input.rs
```diff
@@ -911,54 +908,36 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
-        }).unwrap_or_else(|| {
-            // Do not suppress char if no bindings were triggered
-            false
-        })
+                binding.execute(&mut self.ctx, false);
+
+                // Don't suppress when there has been a `ReceiveChar` action
+                Some(suppress_chars.unwrap_or(true) && binding.action != Action::ReceiveChar)
+            })
+            .unwrap_or(false);
```
zsugabubus: In my previous commit I added a comment for this to explain this `false`. If you don't feel that it's necessary, just ignore.

chrisduerr: Don't think there's anything wrong with a comment on this, definitely makes some sense, feel free to add it back!

alacritty_terminal/src/input.rs
```diff
@@ -930,7 +930,10 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
                 // Don't suppress when there has been a `ReceiveChar` action
                 Some(suppress_chars.unwrap_or(true) && binding.action != Action::ReceiveChar)
             })
-            .unwrap_or(false);
+            .unwrap_or_else(|| {
+                // Do not suppress char if no bindings were triggered
+                false
+            });
```
chrisduerr: We shouldn't use `unwrap_or_else` here. You can just put the comment above the `unwrap_or`.

The `unwrap_or_else` method should only be used with non-constant values that are calculated at runtime.

zsugabubus: Won't it be optimized out?

kchibisov: Relying on optimizer, when things can be easily changed to a faster code is a bad idea.

zsugabubus: Seemed so trivial. I just used it to match style with above lines, but should be corrected new as you requested.

chrisduerr: Just as a sidenote, you're both right here.

For constant values, both will result in exactly the same code in every case as far as I remember. The difference is just that cases which are not constant benefit from `_or_else` in the case where the value is `Some`.

However, if we have the option to express something properly, we should do so. This is simply a style issue and using a function meant for non-constant values to create a closure which just returns a constant simply isn't a good code style.

alacritty_terminal/src/input.rs
```diff
@@ -903,55 +902,42 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
-                            input.modifiers,
-                            &key,
-                            false,
-                        )
-                    } else {
-                        false
-                    }
-                },
-            };
+    /// Return true if chars should be suppressed.
```
chrisduerr: It doesn't return anything anymore.

alacritty_terminal/src/input.rs
```diff
@@ -903,55 +902,42 @@ impl<'a, A: ActionContext + 'a> Processor<'a, A> {
...
                 binding.execute(&mut self.ctx, false);
-                has_binding = true;
-            }
-        }
 
-        has_binding
+                // Don't suppress when there has been a `ReceiveChar` action
+                Some(suppress_chars.unwrap_or(true) && binding.action != Action::ReceiveChar)
+            })
+            .unwrap_or(false); // Don't suppress char if no bindings were triggered
```
chrisduerr: Putting this comment above the `unwrap_or` on the same indentation level as the `.` should hopefully make things a little easier to read.

I'm pretty sure I've tested that and `rustfmt` handles that very well (rustfmt really works well).

Issue #2818, opened 2019-09-23
Remove automatic config generation


kchibisov:
Alacritty can work without any configuration file being presented, so generating one for user in `$HOME/.config/alacritty/alacritty.yml`(1) by default is a bad practice, since users don't expect alacritty to generate configuration file for them (just look at the number of issues with default config, most of the time it's being obtain from github.com etc). Moreover some users are running Alacritty without any additional configuration, so they don't need it anyway. In addition, our config file is a big YAML file and lots of it sections are optional, so advertising default config is bad for migration.

I'm proposing to remove automatic config generation entirely. This approach has a drawback of forcing users to obtain *right* config from external resources (e.g. github.com), however we can add command line option to generate default configuration file and put it in specified path (user have to put generated file to a proper location themself).

(1) This is an example for macOS/Linux systems, the same thing applies to Windows.

chrisduerr:
> we can add command line option to generate default configuration file and put it in specified path

I'm not 100% sure if we should do that tbh, other than that I think I agree.

macOS/Windows users who do not have a package manager will likely have to visit github anyways, so the default config is right there.

On Linux, I think shipping the default config with the package manager seems reasonable, similar to what something like sshd_config does.

The primary question would be if such a config should have all fields commented out by default or not. The advantage of commenting options out is that it makes it very obvious which configuration options have been changed by the user, so migrating to a new config is a little easier.

kchibisov:
> I'm not 100% sure if we should do that tbh, other than that I think I agree.

The goal is to provide default config for your version, you have always a change to grab a wrong one from github.

> On Linux, I think shipping the default config with the package manager seems reasonable, similar to what something like sshd_config does.

iirc Arch Linux is shipping our default config in `/usr/share/doc/alacritty/example/alacritty.yml` and I think it's the right place for it. I'm against of adding config to `/etc/*` and reading anything from it.

I think our default config should have all fields commented out, but we should change our tests then to handle such thing properly (we're relying on things being uncommented atm).

chrisduerr:
> The goal is to provide default config for your version, you have always a change to grab a wrong one from github.

In the github releases where people are likely to download their versions, the configuration file is always archived for each specific release. This was specifically setup for this purpose.

> iirc Arch Linux is shipping our default config in /usr/share/doc/alacritty/example/alacritty.yml and I think it's the right place for it. I'm against of adding config to /etc/* and reading anything from it.

[...]

PR #2820, opened 2019-09-24
Remove automatic config generation

kchibisov:
Alacritty can work without any configuration file being presented,
so generating one for user in $HOME/.config/alacritty/alacritty.yml by default
is considered a bad practice.

Fixes: #2818


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Packaging
 
 - Minimum Rust version has been bumped to 1.36.0
+- Automatic config generation was removed, please ship alacritty.yml in /usr/share/docs
```
kchibisov: I'm not sure about this comment though

kchibisov:
remind to self: recheck every doc about auto config generation once again.

alacritty/src/cli.rs
```diff
@@ -280,44 +280,3 @@ impl Options {
...
-    #[test]
-    fn dynamic_title_overridden_by_config() {
-        let mut config: Config =
-            ::serde_yaml::from_str(DEFAULT_ALACRITTY_CONFIG).expect("deserialize config");
-
-        config.window.title = Some("foo".to_owned());
-        let config = Options::default().into_config(config);
-
-        assert!(!config.dynamic_title());
-    }
```
nixpulvis: Why are you removing these tests instead of updating them?

alacritty_terminal/src/config/test.rs
```diff
@@ -1,22 +0,0 @@
...
-}
-
-#[test]
-fn default_match_empty() {
-    let default = Config::default();
-
-    let empty = serde_yaml::from_str("key: val\n").unwrap();
-
-    assert_eq!(default, empty);
-}
```
nixpulvis: I agree that testing the default config like this isn't a good idea anymore... However, I'd really like it if we could test it somehow. Ideally we'd test that our `Default` matches `alacritty.yml` when everything is uncommented. IDK how best to do this however.

Thoughts?

chrisduerr: Not possible to do this with the options commented out. And it's really not a good idea to try and uncomment things while parsing. That's just a bigger mess than it will benefit us.

alacritty/src/cli.rs
kchibisov: Oops, Will update after evlp2 merge

alacritty_terminal/src/config/test.rs
nixpulvis: I feel pretty uncomfortable about pointing people towards a template config file that, at this point, is completely separated from the `Config` (and it's `Default`) implementation. We just need to manually ensure that `alacritty.yml` is updated as the `Config` changes?

This is a tricky issue though. I think ideally (no idea how to make this happen), we'd document the config options on the `Config` struct (and sub-structs) directly, and have a template (with comments) derived from that. This may be possible with a fancy compiler extension, but otherwise I don't see how to do this. So I guess I'm fine with this for now, and we can continue to be vigilant about config changes.

CHANGELOG.md
nixpulvis: I'm also not sure about this.

- Arch Linux's `community/alacritty` is currently putting things in `/usr/share/doc/alacritty/example/alacritty.yml`.
- I'd be curious to hear @ChrisMacNaughton's thoughs, as the maintainer of our snapcraft
- macOS is another story since we're mainly distributed as a "cask" which is just a `.app` moved into the Applications folder. We don't (unless I'm mistaken) get the opportunity to install anything. Alacritty itself could check for the template and install it when opened, but I don't like this approach at all.

Our own `INSTALL.md` may want to be updated as well. It currently refers to installing things like the icons, and man pages. This seems like a natural addition there.

alacritty_terminal/src/config/test.rs
kchibisov: We've been doing this already. We just had sanity checks for *some* fields, not for all of them. So, I don't any major problem here.

CHANGELOG.md
kchibisov: The key point is not to install config. It's about obtaining the *right* default one, since packagers can ship whatever version they want, so they should also ship a config for this specific version somewhere now. We're just acknowledge them that they can't rely on automatic config generation for this purpose anymore. So adding instructions to INSTALL.md is kind of strange, since we're already have one in README, however I'm not opposed to it and would like to hear more opinions on this.

nixpulvis: @kchibisov how do you propose macOS packagers to solve this problem? I'm not a fan of the only method to find the default config requiring the internet (e.g. checking GitHub).

kchibisov: homebrew can ship docs. I've proposed adding a flag for that purposes, but @chrisduerr wasn't happy about it iirc. We can create alacritty-config(5) man page for this purposes, but it's a not cross platform solution and a maintenance burden.

> I'm not a fan of the only method to find the default config requiring the internet (e.g. checking GitHub).

You need to download Alacritty anyway, it's not shipped as a default terminal on macOS/Windows atm.

kchibisov: I think the message in CHANGELOG is misleading atm, but I've already stated it. Maybe `Automatic config generation was removed, please ship alacritty.yml for easy documentation obtain` will make more sense.

nixpulvis: > You need to download Alacritty anyway, it's not shipped as a default terminal on macOS/Windows atm.

Thing is, one might download and install once, but then configure later. Ideally we make the offline experience as good as possible.

kchibisov: flag for alacritty to generate default config manually

kchibisov: So `alacritty --generate-config`

alacritty_terminal/src/config/test.rs
nixpulvis: Yea, I guess this is a small-ish loss. I'd still be interested in how this could be better though. Not blocking moving forward with this ATM though.

CHANGELOG.md
nixpulvis: I've looked into `brew cask` a bit, and unless I'm missing something it's not really designed to support this use case. But... we could hack the `postflight` hook to install the configuration:

```rb
# ... rest of cask description

postflight do
  path_to_config = ...  # From DMG or separate download?
  prefix = `/usr/local/bin/brew --prefix`.chomp
  doc = "#{prefix}/share/doc/alacritty"
  FileUtils.mkdir_p(doc)
  # TODO: Should really be in the Cellar and symlinked to here.
  FileUtils.cp(path_to_config, "#{doc}/alacritty.yml"
end
```
In a normal brew formula (not a cask) this could be as simple as `doc.install 'alacritty.yml'` but this doesn't seem to work here.

chrisduerr: I don't think it's reasonable to ship our default config file on Windows or macOS to be honest. No matter where we put it, people are much, much more likely to find it on github than somewhere on their system.

kchibisov:
Just to clarify status.

Blocked by #2438

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Packaging
 
 - Minimum Rust version has been bumped to 1.36.0
+- Automatic config generation was removed, please ship alacritty.yml for easy documentation obtain
```
```suggestion
- Config is not generated anymore, please consider distributing the alacritty.yml as documentation
```

README.md
```diff
@@ -154,19 +154,14 @@ properly.
 
 Although it's possible the default configuration would work on your system,
 you'll probably end up wanting to customize it anyhow. There is a default
-`alacritty.yml` at the Git repository root. Alacritty looks for the
-configuration file at the following paths:
+`alacritty.yml` at the Git repository root and on the GitHub's releases page.
+Alacritty looks for the configuration file at the following paths:
```
chrisduerr: I think I'd rework the entire section a bit and skip out the yml in the repo.

Something like this maybe:

```
You can find the default configuration file with documentation for all available fields on the [GitHub releases page](https://github.com/jwilm/alacritty/releases) for each release.

Alacritty looks for the configuration file at the following paths:
```

alacritty/src/config/test.rs
```diff
@@ -1,24 +0,0 @@
...
-}
-
-#[test]
-fn default_match_empty() {
-    let default = Config::default();
-
-    let empty = serde_yaml::from_str("key: val\n").unwrap();
-
-    assert_eq!(default, empty);
-}
```
chrisduerr: We don't need a separate `test.rs` file, but the `EOF` error is something that occurs when trying to load the default configuration file, right?

So it would probably make sense to add just one simple test that would inform us when our EOF error message checking has failed. Just a simple deserialization of the documentation file and checking that it matches the default.

alacritty/src/main.rs
```diff
@@ -85,18 +85,10 @@ fn main() {
...
-        .or_else(|| config::write_defaults().ok())
-        .map(|path| path.to_path_buf());
-    let config = if let Some(path) = config_path {
-        config::load_from(path)
-    } else {
-        error!("Unable to write the default config");
-        Config::default()
-    };
+    let config_path =
+        options.config_path().or_else(config::installed_config).map(|path| path.to_path_buf());
```
chrisduerr: Do we need to convert to `PathBuf` here? I think opening files should work with `Path` too, but I'm not entirely certain. No need to convert unless we need it of course.

alacritty/src/main.rs
```diff
@@ -85,18 +85,10 @@ fn main() {
...
+    let config =
+        if let Some(path) = config_path { config::load_from(path) } else { Config::default() };
```
chrisduerr: I think `config_path.map(|path| config::load_from(path)).unwrap_or_else(Config::default)` would be a bit more idiomatic if that works out without errors. Should also save us the annoying line break here.

alacritty/src/main.rs
chrisduerr: @kchibisov This is not resolved.

alacritty/src/config/mod.rs
```diff
@@ -232,3 +201,17 @@ fn print_deprecation_warnings(config: &Config) {
...
 }
+
+#[cfg(test)]
+mod test {
+    use super::Config;
+    use std::path::PathBuf;
+
+    #[test]
+    fn config_read_eof() {
+        let config_path: PathBuf = ["..", "alacritty.yml"].iter().collect();
```
chrisduerr:
```suggestion
        let config_path = PathBuf::from("../alacritty.yml");
```

This should do the job, right?

kchibisov: I don't think that it is going to work on Windows.

DivineGod: For what it's worth, the two are equivalent as long as you're dealing with strings, so will work on windows as well due to Path and PathBuf solving the platform issues.

ref: https://doc.rust-lang.org/std/path/struct.PathBuf.html#examples

kchibisov: This reference doesn't say anything about their equivalence at all. It turned out that `/` is also a separator on Windows, however this platform uses `\` as a **main** separator. `Path` and `PathBuf` doesn't solve this for me actually, it's done by Windows (or at least modern windows). However there is some edge cases, like root prefixes, `C:` and so on If I'm understanding the code in path.rs correctly.

So, the following conversion on Windows is lossless. 
```rust
let p: = std::path::PathBuf("hello/world/zoo");
let p = p.to_str();
``` 

P.s. default windows terminal perfectly understand navigation with `/` and file explorer perfectly understand navigation with them.

Issue #2822, opened 2019-09-24
Crash on `ESC [ Pn; Pn r` with Invalid Range


nixpulvis:
Currently if you run `echo -e '\e[15;1r'` in debug mode, Alacritty will panic. This is because a range should start with a smaller number than it ends with. XTerm (and I) think that a sequence like this should simply be ignored.

It's also worth noting that XTerm also ignores a range where `A=B` so `echo -e '\e[15;15r'` would also be ignored.

PR #2823, opened 2019-09-24
Fix overflow on wrong scroll region setting

kchibisov:
Fixes: #2822

### DIFF
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -14,7 +14,6 @@
 //
 //! ANSI Terminal Stream Parsing
 use std::io;
-use std::ops::Range;
 use std::str;
 
 use crate::index::{Column, Contains, Line};
@@ -296,7 +295,7 @@ pub trait Handler {
     fn unset_mode(&mut self, _: Mode) {}
 
     /// DECSTBM - Set the terminal scrolling region
-    fn set_scrolling_region(&mut self, _: Range<Line>) {}
+    fn set_scrolling_region(&mut self, _top: usize, _bottom: usize) {}
 
     /// DECKPAM - Set keypad to applications mode (ESCape instead of digits)
     fn set_keypad_application_mode(&mut self) {}
@@ -1079,22 +1078,10 @@ where
                 handler.set_cursor_style(style);
             },
             ('r', None) => {
-                // Bottom should be included in the range, but range end is not
-                // usually included. One option would be to use an inclusive
-                // range, but instead we just let the open range end be 1
-                // higher.
                 let top = arg_or_default!(idx: 0, default: 1) as usize;
                 let bottom = arg_or_default!(idx: 1, default: handler.lines().0 as _) as usize;
 
-                if top >= bottom {
-                    debug!("Invalid scroll region escape: ({};{})", top, bottom);
-                    return;
-                }
-
-                let top = Line(top - 1);
-                let bottom = Line(bottom);
-
-                handler.set_scrolling_region(top..bottom);
+                handler.set_scrolling_region(top, bottom);
             },
             ('s', None) => handler.save_cursor_position(),
             ('u', None) => handler.restore_cursor_position(),
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -1080,12 +1080,18 @@ where
             },
             ('r', None) => {
                 let arg0 = arg_or_default!(idx: 0, default: 1) as usize;
+                let arg1 = arg_or_default!(idx: 1, default: handler.lines().0 as _) as usize;
+
+                if arg0 >= arg1 {
+                    debug!("Invalid scroll region setting: ({};{})", arg0, arg1);
+                    return;
+                }
+
                 let top = Line(arg0 - 1);
                 // Bottom should be included in the range, but range end is not
-                // usually included.  One option would be to use an inclusive
+                // usually included. One option would be to use an inclusive
                 // range, but instead we just let the open range end be 1
                 // higher.
-                let arg1 = arg_or_default!(idx: 1, default: handler.lines().0 as _) as usize;
                 let bottom = Line(arg1);
 
                 handler.set_scrolling_region(top..bottom);
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -1079,20 +1079,20 @@ where
                 handler.set_cursor_style(style);
             },
             ('r', None) => {
-                let arg0 = arg_or_default!(idx: 0, default: 1) as usize;
-                let arg1 = arg_or_default!(idx: 1, default: handler.lines().0 as _) as usize;
-
-                if arg0 >= arg1 {
-                    debug!("Invalid scroll region setting: ({};{})", arg0, arg1);
-                    return;
-                }
-
-                let top = Line(arg0 - 1);
                 // Bottom should be included in the range, but range end is not
                 // usually included. One option would be to use an inclusive
                 // range, but instead we just let the open range end be 1
                 // higher.
-                let bottom = Line(arg1);
+                let top = arg_or_default!(idx: 0, default: 1) as usize;
+                let bottom = arg_or_default!(idx: 1, default: handler.lines().0 as _) as usize;
+
+                if top >= bottom {
+                    debug!("Invalid scroll region escape: ({};{})", top, bottom);
+                    return;
+                }
+
+                let top = Line(top - 1);
+                let bottom = Line(bottom);
 
                 handler.set_scrolling_region(top..bottom);
             },
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -2119,7 +2119,7 @@ impl ansi::Handler for Term {
     #[inline]
     fn set_scrolling_region(&mut self, top: usize, bottom: usize) {
         if top >= bottom {
-            debug!("Invalid scroll region: ({};{})", top, bottom);
+            debug!("Invalid scrolling region: ({};{})", top, bottom);
             return;
         }
 
@@ -2130,7 +2130,7 @@ impl ansi::Handler for Term {
         let start = Line(top - 1);
         let end = Line(bottom);
 
-        trace!("Setting scrolling region: {}x{}", start, end);
+        trace!("Setting scrolling region: ({};{})", start, end);
 
         self.scroll_region.start = min(start, self.grid.num_lines());
         self.scroll_region.end = min(end, self.grid.num_lines());
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1260,8 +1260,7 @@ impl Term {
     fn deccolm(&mut self) {
         // Setting 132 column font makes no sense, but run the other side effects
         // Clear scrolling region
-        let scroll_region = Line(0)..self.grid.num_lines();
-        self.set_scrolling_region(scroll_region);
+        self.set_scrolling_region(1, self.grid.num_lines().0);
 
         // Clear grid
         let template = self.cursor.template;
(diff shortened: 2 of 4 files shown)


alacritty_terminal/src/ansi.rs
```diff
@@ -1080,12 +1080,18 @@ where
...
+                }
+
                 let top = Line(arg0 - 1);
                 // Bottom should be included in the range, but range end is not
-                // usually included.  One option would be to use an inclusive
+                // usually included. One option would be to use an inclusive
                 // range, but instead we just let the open range end be 1
                 // higher.
-                let arg1 = arg_or_default!(idx: 1, default: handler.lines().0 as _) as usize;
                 let bottom = Line(arg1);
```
chrisduerr: I think there's little reason to separate top/bottom like this. It looks a bit silly I think. The comment talks about both anyways, so it should be fine to do comment -> let top -> let bottom. Then everything's in a nice order and looks good.

alacritty_terminal/src/ansi.rs
```diff
@@ -1080,12 +1080,18 @@ where
             },
             ('r', None) => {
                 let arg0 = arg_or_default!(idx: 0, default: 1) as usize;
+                let arg1 = arg_or_default!(idx: 1, default: handler.lines().0 as _) as usize;
+
+                if arg0 >= arg1 {
```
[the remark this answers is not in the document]
kchibisov: I'm just constantly follow old code in that hot fixes for whatever reason.

alacritty_terminal/src/ansi.rs
```diff
@@ -1079,14 +1079,20 @@ where
                 handler.set_cursor_style(style);
             },
             ('r', None) => {
-                let arg0 = arg_or_default!(idx: 0, default: 1) as usize;
-                let top = Line(arg0 - 1);
                 // Bottom should be included in the range, but range end is not
-                // usually included.  One option would be to use an inclusive
+                // usually included. One option would be to use an inclusive
                 // range, but instead we just let the open range end be 1
                 // higher.
```
chrisduerr: This comment should stay above the other `top/bottom` block, since that's where the `-1` stuff is happening.

alacritty_terminal/src/ansi.rs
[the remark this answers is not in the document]
nixpulvis: I think rather than shadowing (rebinding) `top` and `bottom` we could just use call `set_scrolling_region(Line(top - 1)..Line(bottom))` here honestly.

This is a comment of arguable merit, please feel free to ignore.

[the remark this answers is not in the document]
chrisduerr: I think I prefer shadowing, especially because of the comment explaining that the subtraction here is intentional.

kchibisov:
Oh, *cognitive complexity* thing strikes back

kchibisov:
I'm not sure that we can split this function though.

kchibisov:
I'll silence it for now, but in case we decide to go with a different approach we can just revert the last commit on this PR.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2118,10 +2117,23 @@ impl ansi::Handler for Term {
     }
 
     #[inline]
-    fn set_scrolling_region(&mut self, region: Range<Line>) {
-        trace!("Setting scrolling region: {:?}", region);
-        self.scroll_region.start = min(region.start, self.grid.num_lines());
-        self.scroll_region.end = min(region.end, self.grid.num_lines());
+    fn set_scrolling_region(&mut self, top: usize, bottom: usize) {
+        if top >= bottom {
+            debug!("Invalid scroll region escape: ({};{})", top, bottom);
```
kchibisov: I’m not sure that escape is accurate here. It was fine ansi.rs, since we’re working with escapes. Probably dropping escapes will make sense.

chrisduerr: Yeah you're right, feel free to fix this yourself.

kchibisov:
Ok, the approach you've suggested ofc fixes clippy error, the problem is, that the idea of restring escape is missing. Now it feels like we're restricting our `set_scrolling_region`, which is fine, but it's not "semantically" the same code as before.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2118,10 +2117,23 @@ impl ansi::Handler for Term {
...
+        }
+
+        // Bottom should be included in the range, but range end is not
+        // usually included. One option would be to use an inclusive
+        // range, but instead we just let the open range end be 1
+        // higher.
+        let start = Line(top - 1);
+        let end = Line(bottom);
+
+        trace!("Setting scrolling region: {}x{}", start, end);
```
nixpulvis: I wouldn't use a `x` to denote a range. To be `x` means width x height, or age x gender or generally any two separate dimensions. While `..` or `...` (depending who you ask) is better for a range. However since we're in Rust I'd stick with `..` meaning half-open. If you're trying to avoid that then maybe `x-y` would even be better.

Sorry I love being pedantic.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2118,10 +2117,23 @@ impl ansi::Handler for Term {
...
+            debug!("Invalid scroll region: ({};{})", top, bottom);
+            return;
+        }
+
+        // Bottom should be included in the range, but range end is not
+        // usually included. One option would be to use an inclusive
+        // range, but instead we just let the open range end be 1
+        // higher.
+        let start = Line(top - 1);
+        let end = Line(bottom);
```
nixpulvis: I'm going to keep making some noise about these pervasive `Line` constructs. :boom: 

We're moving away from a `Range<Line>`, which I can understand since ranges are less clear in this context than "top" and "bottom", but we're replacing it with two `usize` values. This seems most natural to me, but defeats the point of the `Line`. I think this is further evidence that the `Line` in general is doing more harm than good in it's current form, and would be better off eliminated.

That said, we use `Line` in many places as indices, and this isn't necessarily bad, however, we seem to be using `Line` to mean both index, and count, depending on context, and sometimes just a plain old `usize` (re, this PR)... This is further confused when interfacing with specs (we don't control) that use 1 indexing.

https://github.com/jwilm/alacritty/blob/c1f089970fd3d4b9137d07647f8cd028b2b5b3a9/alacritty_terminal/src/grid/mod.rs#L142-L150

https://github.com/jwilm/alacritty/blob/c1f089970fd3d4b9137d07647f8cd028b2b5b3a9/alacritty_terminal/src/grid/mod.rs#L286

Either we make a type for each kind of thing, i.e. `Line`, `Lines`, `OneIndexedLine`, or we should bail on the concept altogether IMO.

I've made this argument before, but I'm making it again. Sorry.

alacritty_terminal/src/ansi.rs
```diff
@@ -1079,16 +1078,10 @@ where
...
-                // range, but instead we just let the open range end be 1
-                // higher.
-                let arg1 = arg_or_default!(idx: 1, default: handler.lines().0 as _) as usize;
-                let bottom = Line(arg1);
-
-                handler.set_scrolling_region(top..bottom);
+                let top = arg_or_default!(idx: 0, default: 1) as usize;
+                let bottom = arg_or_default!(idx: 1, default: handler.lines().0 as _) as usize;
+
+                handler.set_scrolling_region(top, bottom);
```
nixpulvis: Nice, this is much more consistent with the rest of this function. :tada:

alacritty_terminal/src/term/mod.rs
nixpulvis: To go further off the deep end on this topic, let's talk about indices and counts in general here. Please forgive my rambling:

```rust
// This is an *empty* array `[]`.
let empty = [];
let n = 0;
let i = None;

// This is an array with a 0th element `["A"]`.
let single = ["A"];
let n = 1;
let i_a = Some(0);
let i_wtf = Some(1);

// You can see where this is going...
let many = ["A", "B", "C"];
let n = 3;
let i_c = Some(2);
let i_wtf = Some(3);

// So we can see two important facts about indices with respect to counts:
//
// 1. When the count is 0, there isn't an index to be had at all
// 2. When the count is > 0, the index is bound to be < the count
//
// OK.

// So we need to decide if we want a type for cases when we already have an
// element in mind, so no optional is needed, or if we want to store an index
// generally along side some collection, making an optional needed.

// For Alacritty's purposes, we generally already have elements in mind, and
// simply want to address them. Maybe I'm just not giving enough credit to the
// case when the grid is empty though.
//
// For the next bit tho, let's pretend like we've always got at least one
// element...
struct Count(usize);
struct Index(usize);

assert_eq!(many.count(), Count(3));
assert_eq!(many.index_of("A"), Index(0));

// Indices shouldn't be added to each other, but instead with a count.
assert_eq!(Index(0) + Index(10), ???);
assert_eq!(Index(0) + Count(10), Index(10));

// This doesn't help with the fundamental issue motivating Alacritty's index
// types in the first place however, which is mainly to prevent confusing
// columns and lines.
//
// Here we call a poorly defined function, since it's *VERY* easy to call in
// the wrong order.
let column = 0;
let line = 1;
poorly_designed_function(column, line);

// If Rust had keyword arguments we could have designed a function:
properly_designed_function(column: column, line: line);
// And struct style elision might even allow us to write
properly_designed_function(column, line);
// As in the poorly designed function before... (this syntax is contentious)

// However in the real world (for now), we can write
let column = Column(0);
let line = Line(1);
our_designed_function(column, line);

// This comes with other benefits. Like causing a type error when I fail to
// copy and paste this simple arithmetic. This is a somewhat contrived example,
// but does honestly prevent some errors.
let ca = Column(0);
let cb = Column(1);
let la = Line(0);
let lb = Line(3);
let cr = ca + cb;
let lr = la + cb;

// But, here's where the problem begins! Now we're tempted to do this
let current_position = Index(9);
let current_lines = current_position.0 + 1;
// Which gets more confusing if we mistakenly use an index as a count!
let lines = Index(10);
let position = lines.0 - 1;
// And we forget to write `.0` and end up with indices where counts should be!

// But this is just a matter of `+ 1` vs `- 1` you say! Yes, and the other
// issue was just a matter of `Column` vs `Line`... (Un)fortunately, these
// things make a difference.

// Of course none of this talks about translating coordinate systems that grow
// from the top, or bottom.

// Is this the top or the bottom?
Index(0)

// Luckily, this is why god gave us names.
let bottom = Index(0);

// We could encode this into the type as well.
let index = BottomIndexedIndex(0);
// But I can't think of a pretty way to do this...
let index = bottom::Index(0);
// Better, but still...

// So we need (Line, Count + Index) and (Column, Count + Index).
struct Count<T>(usize, PhantomData<T>);
struct Index<T>(usize, PhantomData<T>);
struct Line;
struct Column;

// Now between the value types (Count, and Line), the marker types (Line, and
// Column), and our binding names (top, and bottom), we've completely specified
// the three critical aspects of our grid's elements. Although, in some cases
// the binding name will imply the type, this can't be relied on in general.
// This is however a step up in expressive power than we have today.
let top = Index::<Line>::new(23);
let replacement = Index::<Column>::new(13);
let visible = Count::<Line>::new(1337);

// However this method requires a `new` function, and suddenly I find myself
// wondering about the overhead of these supposedly helpful types.

// Finally I arrive at
struct Line(usize);   // An index.
struct Lines(usize);  // A count.

let top = Line(23);
let history = Lines(1337);
assert_eq!(top + history, Line(1360));

// Simple, elegant. DHH would be proud.

// I'm sure there's more to this still, but I'm going to sleep now.
```

I'm worried the most *interesting* point in all this remains unaddressed, which is that an `Index` is generally an `Option<usize>`, which isn't something I think we care to talk about much.

chrisduerr:
> Ok, the approach you've suggested ofc fixes clippy error, the problem is, that the idea of restring escape is missing. Now it feels like we're restricting our set_scrolling_region, which is fine, but it's not "semantically" the same code as before.

I don't see the issue here? The escape itself is handled in the VTE crate. If something would be properly suppressed, it would be done there. Handling it in the ansi.rs handler or the term/mod.rs handler doesn't make much difference I think.

The functionality is still exactly the same and we've successfully removed the complexity from the broad method to the more specific handler, so that makes a lot of sense to me. If some other part of our code would want to create a scrolling region with bottom above top, we should also block it. So as far as I'm concerned this is actually a far better place to handle things.

Let me know what you think though, but I'm open to merge.

PR #2827, opened 2019-09-26
Fix selection not inverting terminal background

chrisduerr:
Fixes a regression introduced in
9a0555bbba30c264f617ec9260ca00e0eab70870 where the terminal background
would not get inverted when selected.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -289,6 +289,8 @@ impl RenderableCell {
                 // Invert cell fg and bg colors
                 mem::swap(&mut fg_rgb, &mut bg_rgb);
             }
+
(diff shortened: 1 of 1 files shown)


chrisduerr:
Could you review this @kchibisov? Just to make sure this wasn't left out intentionally on the old PR.

Also pinging @xPMo since he found the issue.

kchibisov:
@chrisduerr my pr was 1 minute earlier btw. I've somehow forgot this logic.

chrisduerr:
@kchibisov Curious why you recompute instead of fixing the bg_alpha to 1.0.

I don't think it's possible to set the foreground color to NamedColor::Background?

xPMo:
I applied the patch and built it, everything looks good!

PR #2831, opened 2019-09-28
Concatenate parameters of title OSC

pcapriotti:
A semicolon in a title OSC should be interpreted literally, not as a parameter
separator, but the OSC parser is very simple and does not know about arities of
commands.

Therefore, this patch takes all the parameters returned by the OSC parser and
reconstructs the original string by interspersing semicolons. Now an OSC like
`\e]2;hello;world` will set the title to 'hello;world' and not 'hello' like
before.

### DIFF
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -750,10 +750,13 @@ where
             // Set window title
             b"0" | b"2" => {
                 if params.len() >= 2 {
-                    if let Ok(utf8_title) = str::from_utf8(params[1]) {
-                        self.handler.set_title(utf8_title);
-                        return;
-                    }
+                    let title = params[1..]
+                        .iter()
+                        .flat_map(|x| str::from_utf8(x))
+                        .collect::<Vec<&str>>()
+                        .join(";");
+                    self.handler.set_title(&title);
+                    return;
                 }
                 unhandled(params);
             },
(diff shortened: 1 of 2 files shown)


alacritty_terminal/src/ansi.rs
```diff
@@ -750,10 +750,17 @@ where
...
+                    }
+                    for item in &params[2..] {
+                        if let Ok(title_segment) = str::from_utf8(item) {
+                            buf.push(';');
+                            buf.push_str(title_segment);
+                        }
                     }
+                    self.handler.set_title(&buf);
                 }
                 unhandled(params);
```
chrisduerr: I think we can simplify this code significantly:

```rust
                if params.len() >= 2 {
                    let title = params[1..]
                        .iter()
                        .flat_map(str::from_utf8)
                        .collect::<Vec<&str>>()
                        .join(";");
                    self.handler.set_title(&title);
                    return;
                }
                unhandled(params);
```

pcapriotti: Yes, that looks nicer. However, I've had to eta-expand the argument to `flat_map` because of some error with lifetimes I didn't totally understand.

chrisduerr: Oh yeah, sorry. I meant to post it with a dedicated closure since the types of the functions don't match up entirely, looks like I forgot to edit that, your version is exactly what I was going for!

Issue #2752, opened 2019-08-26
Add line numbers to config error messages


ghost:
Which operating system does the issue occur on?
Linux
If on linux, are you using X11 or Wayland?
X11

I just updated alacritty after a few months and it's complaining about config problems.
`[2019-08-25 17:41] [ERROR] Problem with config: invalid type: unit value, expected a map; using default value`
The error log in /tmp really isn't very helpful and I have no idea where the error actually is. Please add line numbers so I can get _some_ idea of what is wrong easily.

chrisduerr:
The problem is that `serde_yaml::Error` will report location on error, however only when directly deserializing from the file.

What we do is deserialize it to `serde_yaml::Value` first. If that fails, we get correct line error messages.

However, when deserializing the `serde_yaml::Value` to the specific type, the line information will never be present in the resulting Error, since it is lost after it has been successfully deserialized to the `serde_yaml::Value`.

So as far as I'm concerned, there's nothing we can do with the current API of `serde_yaml`. We could require `Clone` or `Copy` on all fields deserialized with `failure_default` and then deserialize everything twice, first to check if there's an error and if there is one we clone and deserialize it to get the line number. However that's extremely hacky and I don't think it's a good approach.

martskins:
I managed to get the error line in everything but the types that are deserialized from both string or map (`window.class` and `shell`) by changing the `failure_default` implementation to not use a `serde_yaml::Value`, unless I'm missing something that should make things better for most of the cases.

Here's what it looks like for non string-or-map fields:

![explained-error](https://user-images.githubusercontent.com/4250565/65929125-2b926900-e3d7-11e9-948d-4c18e9b39c44.png)

As for the fields that deserialize from `string` or `map`, the error displays exactly the same:

![image](https://user-images.githubusercontent.com/4250565/65929182-64cad900-e3d7-11e9-8b11-dfb634c910fe.png)

If this means an improvement I can submit a (really small) PR for this :smile:

nixpulvis:
@martskins yes this would be great. Even small QoL things like this are valuable.

kchibisov:
@chrisduerr is toml any better for us here? I think it always will provide a line along the error?

chrisduerr:
![tmp](https://github.com/alacritty/alacritty/assets/8886672/8f253756-4ffc-489d-a5d1-5dc600a25ee7)

I think we can consider this fixed. It will depend on the type of error, but I think at least syntax errors should be linted with line.

PR #2833, opened 2019-10-01
Add error detail, when possible, on the event of an invalid config file

martskins:
This PR aims to help with error descriptions (#2752) by removing the intermediate deserialization to `serde_yaml::Value` in the `failure_default` deserializer, which is used in the vast majority of fields, resulting in:

explained-error

Instead of

image

This change has no effect in how errors are displayed for fields that use `from_string_or_deserialize` or `option_explicit_none`, as they would need to either require `Clone` on `D` or to be able to deserialize from a borrowed field, which I believe is not possible in serde_yaml as it stands.

### DIFF
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -429,7 +429,7 @@ where
     D: Deserializer<'a>,
     T: Deserialize<'a> + Default,
 {
-    Ok(T::deserialize(Value::deserialize(deserializer)?).unwrap_or_else(fallback_default))
+    Ok(T::deserialize(deserializer).unwrap_or_else(fallback_default))
 }
 
 pub fn option_explicit_none<'de, T, D>(deserializer: D) -> Result<Option<T>, D::Error>


chrisduerr:
This is intentionally done this way, since serde does not support recovering from all errors and this will cause problems (and has done in the past). As a result it's possible to crash Alacritty without falling back to the default configuration at all, which should never happen.

If deserialization to a value fails, we need to propagate an error. There's no way around this.

And we don't just want to fail when we can instead provide a reasonable default, because falling back to the default config (by raising the error when deserializing as `T`), is usually a lot worse then just using our defaults.

nixpulvis:
@martskins I wanted to encourage you to think more about this issue, and hopefully find a nice solution. I wasn't sure if things in serde had changed or if there was a nice new solution out there.

However after recalling the issues we've had with this in the past (thanks @chrisduerr for the refresher above), it may be harder than anyone's willing to work out ATM. For example, this branch as it stands will `panic!` on the following configuration file:

```yaml
# Duplicate top level seems to work...
#
#dynamic_title: true
#dynamic_title: true

colors:
  primary:
    # Duplicate nested will panic!
    background: "0x000000"
    background: "0x000000"
```

As I'm understanding this limitation with serde, it's simply that it wont let users recover from a failed parse, and in our case fall back on the default. I believe this to be a **real** limitation on serde, but I'd love to be proven wrong (I might have to play with it again myself too).

As far as I can tell the next step here would be to explain the need for this functionality in serde-rs/serde#1041, since I suppose this is an "advanced feature" and maybe with enough support from the community, there could be some movement on.

Otherwise, while we're using serde and serde-yaml I don't know how feasible this is. :frowning_face: Sorry for baiting you into this can of worms.

martskins:
Hey, no worries! I knew there had to be something I was missing there. That deserializing to Value makes perfect sense now, so thank you for the explanation!

I'll give this another twist to see if I can find a solution for this, but yeah, that `panic!` there is making things hard now :smile:

Issue #2737, opened 2019-08-21
Live reload font config


rgtk:
Changes to color scheme in config are reflected in all windows without need for restarts. Would it be possible to have the same for font size change?

I jump between displays with different DPI throughout the day. In one 11 size looks good, on the other it's too big. I would love to amend alacritty.yml when computer detects screen change to appropriate font size.

I know it is possible to change font dynamically with `C-+`/`C--` but I with many windows in use, it's very tiresome.

chrisduerr:
This has been on my roadmap for a while actually, I've just never gotten around to create the issue itself.

There's no reason why we would keep it to just the font size either. We should also be able to live reload the font family and style.

chrisduerr:
> This would presumably override the current font size set by the user of a running terminal (set using the *FontSize actions). This seems fine to me, but if people are trying to use the live config feature to update visuals on the fly, I wouldn't want my font sizes resetting.

We should be able to just compare if the current font size is equal to the original font size and reset it only if that is the case.

> Generally live updating configurations that can be changed at runtime seems hard. How do we handle this with any other configs today?

I don't think any other option where this is possible comes to mind.

waynr:
Has anyone begun working on this? It's kind of a personal itch that I'd like to scratch and I am looking for a good [hacktoberfest](https://hacktoberfest.digitalocean.com/details) challenge.

chrisduerr:
@waynr As far as I know, nobody is working on this right now and it would be a very welcome addition.

If you have any questions about how to approach this, please let me know.

These are some of the relevant places for this issue:
 - [This is the config update handler closest to where our font is cached](https://github.com/jwilm/alacritty/blob/master/alacritty_terminal/src/display.rs#L129-L131)
 - [Our font cache is created here](https://github.com/jwilm/alacritty/blob/master/alacritty_terminal/src/display.rs#L308-L309)

As far as I can tell, the following steps should be taken:
 - Check if font has changed
 - Evict existing glyph cache
 - Fill new glyph cache
 - Update our cell metrics

This is of course just necessary if we want to go all the way and enable live reloading for the font. If we just want to live update the default font size, things get much easier.

The font size update code can just be handled in the [terminal's config update handler](https://github.com/jwilm/alacritty/blob/master/alacritty_terminal/src/term/mod.rs#L926-L941) and we'd just need to check if the current font size is equal to the old default font size and if that's the case we can change it to the new font size.

[...]

PR #2834, opened 2019-10-02
Live reload font size from alacritty config

waynr:
Addresses first part of \#2737. Second part, which will be font family and styles to also be live-reloaded, is coming in a follow-up PR.

### DIFF
--- a/alacritty_terminal/src/selection.rs
+++ b/alacritty_terminal/src/selection.rs
@@ -438,7 +438,7 @@ mod test {
         assert_eq!(selection.to_span(&term(1, 1)).unwrap(), Span {
             start: location,
             end: location,
-            is_block: false,
+            is_block: false
         });
     }
 
@@ -456,7 +456,7 @@ mod test {
         assert_eq!(selection.to_span(&term(1, 1)).unwrap(), Span {
             start: location,
             end: location,
-            is_block: false,
+            is_block: false
         });
     }
 
@@ -579,7 +579,7 @@ mod test {
         assert_eq!(selection.to_span(&term(5, 10)).unwrap(), Span {
             start: Point::new(0, Column(4)),
             end: Point::new(2, Column(4)),
-            is_block: true,
+            is_block: true
         });
     }
 
--- a/alacritty_terminal/src/selection.rs
+++ b/alacritty_terminal/src/selection.rs
@@ -438,7 +438,7 @@ mod test {
         assert_eq!(selection.to_span(&term(1, 1)).unwrap(), Span {
             start: location,
             end: location,
-            is_block: false
+            is_block: false,
         });
     }
 
@@ -456,7 +456,7 @@ mod test {
         assert_eq!(selection.to_span(&term(1, 1)).unwrap(), Span {
             start: location,
             end: location,
-            is_block: false
+            is_block: false,
         });
     }
 
@@ -579,7 +579,7 @@ mod test {
         assert_eq!(selection.to_span(&term(5, 10)).unwrap(), Span {
             start: Point::new(0, Column(4)),
             end: Point::new(2, Column(4)),
-            is_block: true
+            is_block: true,
         });
     }
 
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -939,8 +939,7 @@ impl Term {
         self.auto_scroll = config.scrolling.auto_scroll;
         self.grid.update_history(config.scrolling.history() as usize, &self.cursor.template);
 
-        if self.original_font_size == self.font_size
-        {
+        if self.original_font_size == self.font_size {
             self.original_font_size = config.font.size;
             self.font_size = config.font.size;
         }
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -939,10 +939,10 @@ impl Term {
         self.auto_scroll = config.scrolling.auto_scroll;
         self.grid.update_history(config.scrolling.history() as usize, &self.cursor.template);
 
-        self.original_font_size = config.font.size;
         if self.original_font_size == self.font_size {
             self.font_size = config.font.size;
         }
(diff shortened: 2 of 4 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -933,6 +933,11 @@ impl Term {
                 self.colors[i] = self.original_colors[i];
             }
         }
+        if self.original_font_size != config.font.size && self.original_font_size == self.font_size
```
chrisduerr: It's probably not worth it to check for `self.original_font_size != config.font.size`, since all that saves us are two assignments which should basically be free. The font itself is only ever changed if it actually changed something anyways.

I'd also like to put this at the end of the method with one newline separating it. Just to separate the different blocks a bit from each other and not have one massive function without a single newline.

alacritty_terminal/src/term/mod.rs
```diff
@@ -933,16 +933,17 @@ impl Term {
...
-            self.font_size = config.font.size;
-        }
         self.visual_bell.update_config(config);
         self.default_cursor_style = config.cursor.style;
         self.dynamic_title = config.dynamic_title();
         self.auto_scroll = config.scrolling.auto_scroll;
         self.grid.update_history(config.scrolling.history() as usize, &self.cursor.template);
+
+        if self.original_font_size == self.font_size
+        {
```
chrisduerr: Alacritty's (and Rust's) codestyle is to open parenthesis on the same line. Running `cargo fmt` with `rustfmt` installed should fix this.

waynr: Okay; my editor was running `rustfmt` automatically before I started working on this but it was also making a bunch of other trivial changes (and preferring the style shown here). I switched from `stable` to `nightly` and it seems to do what you want.

alacritty_terminal/src/selection.rs
```diff
@@ -579,7 +579,7 @@ mod test {
         assert_eq!(selection.to_span(&term(5, 10)).unwrap(), Span {
             start: Point::new(0, Column(4)),
             end: Point::new(2, Column(4)),
-            is_block: true,
+            is_block: true
```
chrisduerr: Rustfmt needs to be run with nightly to include the nightly formatting options. That's why it has removed the trailing semicolons here for you. Please undo these changes either manually or by running rustfmt with a nightly compiler.

waynr: Hmm, I was running with nightly:

```
zsh/4 9472  (git)-[live-reload-font-size]-% rustup update nightly-x86_64-unknown-linux-gnu
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'

  nightly-x86_64-unknown-linux-gnu unchanged - rustc 1.40.0-nightly (702b45e40 2019-10-01)

info: checking for self-updates
zsh/4 9473  (git)-[live-reload-font-size]-% rustup which cargo
/home/wayne/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/cargo
zsh/4 9474  (git)-[live-reload-font-size]-% /home/wayne/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/cargo fmt
zsh/4 9475  (git)-[live-reload-font-size]-% git diff
zsh/4 9476  (git)-[live-reload-font-size]-%
```

> Please undo these changes either manually or by running rustfmt with a nightly compiler.

I'd be happy to remove them but I want to make sure we have a common understanding that I was using nightly. Also, it seems like the next person to run `cargo fmt` will produce the same changes anyway.

waynr: Do you think it would be useful to add a travis check that fails if `rustfmt --check` returns non-zero?

kchibisov: We have this check on CI, it's optional though, because we're running it on nightly I guess.

waynr: Ah I didn't think to look at what was already running in travis. Looks like the `rustfmt` check is currently passing: https://travis-ci.org/jwilm/alacritty/jobs/592595414

chrisduerr: Make sure to install rustfmt through `rustup` to make sure it's pulling the correct version.

alacritty_terminal/src/term/mod.rs
```diff
@@ -938,6 +938,11 @@ impl Term {
         self.dynamic_title = config.dynamic_title();
         self.auto_scroll = config.scrolling.auto_scroll;
         self.grid.update_history(config.scrolling.history() as usize, &self.cursor.template);
+
+        if self.original_font_size == self.font_size {
+            self.original_font_size = config.font.size;
```
chrisduerr: I found one issue with the functionality of this PR. I think the original font size should be changed regardless of the current font size. So whenever resetting to the original, it would reset to the one set in the config.

waynr: Oh, good catch I'll address that shortly.

alacritty_terminal/src/selection.rs
waynr: That is how I have it installed. One possibility here is that I had an older version of `nightly` installed before this discussion and I updated it specifically to try to fix this; it may be that the older `nightly` was more opinionated about commas whereas the new nightly will allow either style. Whatever the case, I manually reverted the changes in a newer commit.

alacritty_terminal/src/term/mod.rs
```diff
@@ -938,6 +938,11 @@ impl Term {
         self.dynamic_title = config.dynamic_title();
         self.auto_scroll = config.scrolling.auto_scroll;
         self.grid.update_history(config.scrolling.history() as usize, &self.cursor.template);
+
+        self.original_font_size = config.font.size;
+        if self.original_font_size == self.font_size {
```
chrisduerr: We need to check first and set it later, not set it first and then check. Otherwise we'll compare to the changed value.

chrisduerr:
Thanks for the PR, everything working great now. LGTM.

PR #2836, opened 2019-10-02
Live reload font family and style from alacritty config

waynr:
Follow-up to \#2834, addresses second part of \#2737.

This seems to work for me as a naive first pass, but I kind of just blindly copied what looked to be the relevent parts of `Display.new` where the initial `GlyphCache` is populated.

One notable difference is that I left out the logic around estimated DPR since at this point we already have a known `Display.size_info.dpr` calculated when `Display::new` is run.

### DIFF
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -14,7 +14,6 @@
 
 //! The display subsystem including window management, font rasterization, and
 //! GPU drawing.
-use std::cmp::max;
 use std::f64;
 use std::fmt;
 use std::time::Instant;
@@ -24,9 +23,9 @@ use glutin::event_loop::EventLoop;
 use log::{debug, info};
 use parking_lot::MutexGuard;
 
-use font::{self, Rasterize, Size};
+use font::{self, Rasterize};
 
-use alacritty_terminal::config::StartupMode;
+use alacritty_terminal::config::{Font, StartupMode};
 use alacritty_terminal::event::{Event, OnResize};
 use alacritty_terminal::index::Line;
 use alacritty_terminal::message_bar::MessageBuffer;
@@ -37,11 +36,8 @@ use alacritty_terminal::term::color::Rgb;
 use alacritty_terminal::term::{RenderableCell, SizeInfo, Term};
 
 use crate::config::Config;
-use crate::event::{FontResize, Resize};
+use crate::event::DisplayUpdate;
 use crate::window::{self, Window};
-
-/// Font size change interval
-pub const FONT_SIZE_STEP: f32 = 0.5;
 
 #[derive(Debug)]
 pub enum Error {
@@ -116,7 +112,6 @@ impl From<glutin::ContextError> for Error {
 /// The display wraps a window, font rasterizer, and GPU renderer
 pub struct Display {
     pub size_info: SizeInfo,
-    pub font_size: Size,
     pub window: Window,
 
     renderer: QuadRenderer,
@@ -228,14 +223,7 @@ impl Display {
             _ => (),
         }
 
-        Ok(Display {
-            window,
-            renderer,
-            glyph_cache,
-            meter: Meter::new(),
-            size_info,
-            font_size: config.font.size,
-        })
+        Ok(Display { window, renderer, glyph_cache, meter: Meter::new(), size_info })
     }
 
     fn new_glyph_cache(
@@ -270,11 +258,9 @@ impl Display {
     }
 
     /// Update font size and cell dimensions
-    fn update_glyph_cache(&mut self, config: &Config, size: Size) {
+    fn update_glyph_cache(&mut self, config: &Config, font: Font) {
         let size_info = &mut self.size_info;
         let cache = &mut self.glyph_cache;
-
-        let font = config.font.clone().with_size(size);
 
         self.renderer.with_loader(|mut api| {
             let _ = cache.update_font_size(font, size_info.dpr, &mut api);
@@ -286,27 +272,22 @@ impl Display {
         size_info.cell_height = cell_height;
     }
 
-    /// Process resize events
-    pub fn handle_resize<T>(
+    /// Process update events
+    pub fn handle_update<T>(
         &mut self,
         terminal: &mut Term<T>,
         pty_resize_handle: &mut dyn OnResize,
         message_buffer: &MessageBuffer,
         config: &Config,
-        resize_pending: Resize,
+        update_pending: DisplayUpdate,
     ) {
         // Update font size and cell dimensions
-        if let Some(resize) = resize_pending.font_size {
-            self.font_size = match resize {
-                FontResize::Delta(delta) => max(self.font_size + delta, FONT_SIZE_STEP.into()),
-                FontResize::Reset => config.font.size,
-            };
-
-            self.update_glyph_cache(config, self.font_size);
+        if let Some(font) = update_pending.font {
+            self.update_glyph_cache(config, font);
         }
 
         // Update the window dimensions
-        if let Some(size) = resize_pending.dimensions {
+        if let Some(size) = update_pending.dimensions {
             self.size_info.width = size.width as f32;
             self.size_info.height = size.height as f32;
         }
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -225,8 +225,8 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
 
     fn change_font_size(&mut self, delta: f32) {
         *self.font_size = max(*self.font_size + delta, Size::new(FONT_SIZE_STEP));
-        self.display_update_pending.font =
-            Some(self.config.font.clone().with_size(*self.font_size));
+        let font = self.config.font.clone().with_size(*self.font_size);
+        self.display_update_pending.font = Some(font);
         self.terminal.dirty = true;
     }
 
@@ -473,15 +473,13 @@ impl<N: Notify> Processor<N> {
                         processor.ctx.terminal.update_config(&config);
 
                         if processor.ctx.config.font != config.font {
-                            let mut font = config.font.clone();
-
                             // Do not update font size if it has been changed at runtime
-                            if *processor.ctx.font_size != processor.ctx.config.font.size {
-                                font.size = *processor.ctx.font_size;
+                            if *processor.ctx.font_size == processor.ctx.config.font.size {
+                                *processor.ctx.font_size = config.font.size;
                             }
 
-                            processor.ctx.display_update_pending.font = Some(font.clone());
-                            *processor.ctx.font_size = font.size;
+                            let font = config.font.clone().with_size(*processor.ctx.font_size);
+                            processor.ctx.display_update_pending.font = Some(font);
                         }
 
                         *processor.ctx.config = config;
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -66,7 +66,7 @@ pub struct ActionContext<'a, N, T> {
     pub message_buffer: &'a mut MessageBuffer,
     pub display_update_pending: &'a mut DisplayUpdate,
     pub config: &'a mut Config,
-    font_size: Size,
+    font_size: &'a mut Size,
 }
 
 impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionContext<'a, N, T> {
@@ -224,15 +224,14 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     }
 
     fn change_font_size(&mut self, delta: f32) {
-        self.font_size = max(self.font_size + delta, Size::new(FONT_SIZE_STEP));
-        let mut font = self.config.font.clone();
-        font.size = self.font_size;
-        self.display_update_pending.font = Some(font);
+        *self.font_size = max(*self.font_size + delta, Size::new(FONT_SIZE_STEP));
+        self.display_update_pending.font =
+            Some(self.config.font.clone().with_size(*self.font_size));
         self.terminal.dirty = true;
     }
 
     fn reset_font_size(&mut self) {
-        self.font_size = self.config.font.size;
+        *self.font_size = self.config.font.size;
         self.display_update_pending.font = Some(self.config.font.clone());
         self.terminal.dirty = true;
     }
@@ -394,7 +393,7 @@ impl<N: Notify> Processor<N> {
                 message_buffer: &mut self.message_buffer,
                 display_update_pending: &mut display_update_pending,
                 window: &mut self.display.window,
-                font_size: self.font_size,
+                font_size: &mut self.font_size,
                 config: &mut self.config,
             };
             let mut processor = input::Processor::new(context);
@@ -402,8 +401,6 @@ impl<N: Notify> Processor<N> {
             for event in event_queue.drain(..) {
                 Processor::handle_event(event, &mut processor);
             }
-
-            self.font_size = processor.ctx.font_size;
 
             // TODO: Workaround for incorrect startup DPI on X11
             // https://github.com/rust-windowing/winit/issues/998
@@ -479,12 +476,12 @@ impl<N: Notify> Processor<N> {
                             let mut font = config.font.clone();
 
                             // Do not update font size if it has been changed at runtime
-                            if processor.ctx.font_size != processor.ctx.config.font.size {
-                                font.size = processor.ctx.font_size;
+                            if *processor.ctx.font_size != processor.ctx.config.font.size {
+                                font.size = *processor.ctx.font_size;
                             }
 
                             processor.ctx.display_update_pending.font = Some(font.clone());
-                            processor.ctx.font_size = font.size;
+                            *processor.ctx.font_size = font.size;
                         }
 
                         *processor.ctx.config = config;
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -66,7 +66,7 @@ pub struct ActionContext<'a, N, T> {
     pub message_buffer: &'a mut MessageBuffer,
     pub display_update_pending: &'a mut DisplayUpdate,
     pub config: &'a mut Config,
-    font: &'a mut Font,
+    font_size: Size,
 }
 
 impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionContext<'a, N, T> {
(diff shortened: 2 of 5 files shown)


chrisduerr:
The glutin rework PR has just been merged, making some major changes to the display.rs file. If you need any help rebasing on the latest master, please let me know.

The configuration file update is now mainly handled in `alacritty/src/event.rs` in the `handle_event` function and `ConfigReload` handler.

waynr:
@chrisduerr I finally got around to updating this, looking forward to your feedback!

alacritty/src/display.rs
```diff
@@ -286,6 +287,21 @@ impl Display {
         size_info.cell_height = cell_height;
     }
 
+    /// Evict and replace glyph cache
+    pub fn handle_glyph_cache_refresh(&mut self, config: &Config) {
```
chrisduerr: Having a `handle_glyph_cache_refresh` and an `update_glyph_cache` function is a bit confusing, since the names do not clearly separate these two functions. We should make this more clear, renaming the old function is perfectly fine too.

alacritty/src/display.rs
```diff
@@ -286,6 +287,21 @@ impl Display {
...
+        let (glyph_cache, ..) =
```
chrisduerr: I'm not sure if we've discussed this before, but we should handle the cell size changes too. We should not rely on the font size changing at the same time, instead we should either change the font size or evict the complete cache and create a new one. We shouldn't do both.

alacritty/src/event.rs
```diff
@@ -414,6 +428,11 @@ impl<N: Notify> Processor<N> {
                 }
             }
 
+            // Process font change events
+            if !glyph_cache_refresh_pending.is_empty() {
+                self.display.handle_glyph_cache_refresh(&self.config);
+            }
```
chrisduerr: Okay, bear with me here but I think we can do better than having another method that is weirdly strung through our event loop here. The resize should be sufficient.

If you take a look at the font resize code, you'll see that it already reloads the font completely. It also refills the cache with the new glyphs. If I'm reading this correctly, that should mean that our code for changing size and changing font should be identical.

The only thing we should need to change is to pass a `Font` to resize instead of a `Size`. So this would allow us to skip this whole `GlyphCacheRefresh` and solely use the `font_size` to change both font size and font family.

I'm a bit conflicted about naming it `font_size` if it does more than just resizing and I think even naming it `Resize` when it does more than resizing is a bit suboptimal. But our best approach should be to rename the font resize from `font_size` to just `font` and adding a comment that this is a font resize or family change.

If you have any questions about this, please let me know, but hopefully this should eliminate the need for separately handling font size and family changes.

alacritty/src/event.rs
```diff
@@ -461,7 +480,12 @@ impl<N: Notify> Processor<N> {
...
 
-                        if *processor.ctx.font_size == processor.config.font.size {
+                        if processor.ctx.font.size == processor.config.font.size {
+                            processor.ctx.resize_pending.font_size = Some(FontResize::Reset);
+                        }
+
+                        if processor.config.font != config.font {
+                            processor.ctx.glyph_cache_refresh_pending.font = Some(());
                             processor.ctx.resize_pending.font_size = Some(FontResize::Reset);
                         }
```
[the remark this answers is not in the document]
waynr: Ah that's a good point, I hadn't considered that this would accidentally refresh the glyph cache even when only the font size changes.

alacritty/src/display.rs
waynr: Yeah I was going to carry over a lot of the cell size change work that I had included in the glyph cache refresh code prior to the rebase, but I thought it made more sense to trigger a `Resize` event for the sake of avoiding duplication of `handle_resize` code. I was kind of suspecting that you might ask me to rework that, but I also wasn't sure how you would feel about having the glyph cache refresh entirely separate from the `Resize` event either. Looking at your next comment, it seems like you might prefer them to be handled together anyway so I'm glad I didn't try to handle cell resizes here :)

alacritty/src/event.rs
waynr: I think the approach you describe sounds totally reasonable. I did miss that `update_font_cache` is essentially reloading the font anyway. Are you okay with me renaming some things to make it all semantically correct? ie `Resize`, `resize_pending`, etc

chrisduerr: What would you rename `Resize`, `resize_pending` and `handle_resize` to? I'm fine with finding better names, however I'm not sure if that's so simple.

waynr: The common theme I see is that the thing being updated is the configuration of the `Display`, so I was thinking:

* `Resize` -> `ReconfigureDisplay`
* `resize_pending` -> `display_reconfigure_pending`
* `handle_resize` -> `handle_reconfigure`

I also considered something along the lines of `UpdateDisplay`, but that seems to me like it could be misunderstood to refer to a per-loop-iteration update of the display's contents.

`UpdateDisplayConfig` might be a good fit but feels it would get a bit long-winded   since we would end up with `update_display_config_pending`.

alacritty/src/event.rs
```diff
@@ -456,13 +457,13 @@ impl<N: Notify> Processor<N> {
...
                     processor.ctx.message_buffer.remove_target(LOG_TARGET_CONFIG);
-                    processor.ctx.resize_pending.message_buffer = Some(());
+                    processor.ctx.display_reconfig_pending.message_buffer = Some(());
 
                     if let Ok(config) = config::reload_from(&path) {
                         processor.ctx.terminal.update_config(&config);
 
-                        if *processor.ctx.font_size == processor.config.font.size {
-                            processor.ctx.resize_pending.font_size = Some(FontResize::Reset);
+                        if *processor.ctx.font != config.font {
```
waynr: I do still need to think about how to properly address https://github.com/jwilm/alacritty/issues/2737#issuecomment-524929073 here, will work on that tonight or tomorrow morning.

alacritty/src/event.rs
chrisduerr: We're not really changing any display configuration though. The `Display` is not an actual display, it's a struct responsible for displaying content.

Also since we're talking about an enum, we should name things appropriately. `UpdateDisplay` might be a trait name, but an enum should be named `DisplayUpdate`. The enum isn't actually doing anything, it's storing information to do something, so we can't name it like it does something itself.

We also don't only update the display or renderer in this method. The resize is also responsible for updating the terminal and PTY sizes, so naming it something specifically related to the renderer or display is very misleading. It's not stateless outside of `display.rs`.

waynr: Ah shoot, I forgot to submit this comment last week...

> Also since we're talking about an enum, we should name things appropriately. UpdateDisplay might be a trait name, but an enum should be named DisplayUpdate

Good point, I renamed it `DisplayUpdate` in my latest set of updates.

> We also don't only update the display or renderer in this method. The resize is also responsible for updating the terminal and PTY sizes, so naming it something specifically related to the renderer or display is very misleading.

The only place I saw the `Resize` (now `DisplayUpdate`) enum being used was to pass information to the `Display` struct's `handle_resize` method. Could you point out to me where else it is being used to pass update information along?

chrisduerr: `Resize` isn't used in other places, but `handle_resize` does other things too.

alacritty/src/display.rs
```diff
@@ -286,27 +287,27 @@ impl Display {
...
     ) {
         // Update font size and cell dimensions
-        if let Some(resize) = resize_pending.font_size {
-            self.font_size = match resize {
-                FontResize::Delta(delta) => max(self.font_size + delta, FONT_SIZE_STEP.into()),
-                FontResize::Reset => config.font.size,
+        if let Some(update) = update_pending.font {
+            self.font.size = match update {
+                FontUpdate::Delta(delta) => max(self.font.size + delta, FONT_SIZE_STEP.into()),
+                FontUpdate::Reset => config.font.size,
```
chrisduerr: We should pass the font explicitly instead of reading from a seemingly unrelated struct. Doing it like this makes it really hard to reason about where the font update is coming from and where it is sent to the display.

alacritty/src/display.rs
```diff
@@ -234,7 +235,7 @@ impl Display {
             glyph_cache,
             meter: Meter::new(),
             size_info,
-            font_size: config.font.size,
+            font: config.font.clone(),
```
chrisduerr: This clone is problematic, since it means the `font` exist twice. Both in the event processor and the display. This makes it very difficult to keep track of state.

We should check if it is even needed at all or if we can just keep all the state on the event processor.

alacritty/src/event.rs
```diff
@@ -456,13 +455,22 @@ impl<N: Notify> Processor<N> {
...
-                            processor.ctx.resize_pending.font_size = Some(FontResize::Reset);
+                        // use copy of new config with font size overridden to match the current
+                        // font size in order to determine if any font value other than size has
+                        // changed
+                        let mut font_copy = config.font.clone();
+                        font_copy.size = processor.ctx.font.size;
+
+                        if *processor.ctx.font != font_copy {
+                            processor.ctx.display_update_pending.font = Some(FontUpdate::Reset);
                         }
```
chrisduerr: What if the font size has changed, but the current font size is not the default font size? We don't want to reset to the config's font then, because that has the incorrect size.

alacritty/src/event.rs
```diff
@@ -456,13 +455,22 @@ impl<N: Notify> Processor<N> {
...
+                        }
                         processor.ctx.terminal.update_config(&config);
 
-                        if *processor.ctx.font_size == processor.config.font.size {
-                            processor.ctx.resize_pending.font_size = Some(FontResize::Reset);
+                        // use copy of new config with font size overridden to match the current
+                        // font size in order to determine if any font value other than size has
+                        // changed
+                        let mut font_copy = config.font.clone();
+                        font_copy.size = processor.ctx.font.size;
```
chrisduerr: The necessity of the comment should already indicate that this isn't really a great solution to this problem, it might be easier to resolve this once the other issues are addressed, but I thought I'd point it out now already so you can start working if you have an idea.

alacritty/src/event.rs
```diff
@@ -40,21 +39,21 @@ use crate::input::{self, ActionContext as _, Modifiers};
 use crate::window::Window;
 
 #[derive(Copy, Clone, Debug, PartialEq)]
-pub enum FontResize {
+pub enum FontUpdate {
     Delta(f32),
     Reset,
 }
```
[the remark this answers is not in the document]
waynr: > I'd assume that Reset is used both for resetting size and for actually changing the font?

That's right; in either case we are resetting the font (size, family, style) on the `Display` and its child structs to the value in the config, whether the event is triggered by:

* a config reload that may include a size, family, or style changes or
* a size reset action (ie whatever is bound to ctrl-0 by default).

> Maybe it makes sense to just remove this option and change the update field type to Option<Font>?

I think this would complicate things a bit since in many `Delta` cases it looks like there is no explicit `Font` available, just an incremental change that needs to be passed to the `handle_update` function (the `f32` parameter on `Delta`).

I'll probably rename `Delta` -> `SizeDelta` in my next set of updates.

alacritty/src/display.rs
waynr: Are you saying that we should do something like `FontUpdate::ResetFont(font) => font.size`?

chrisduerr: We'd probably need more than just font size if we don't want to store it on the Display, but yeah.

alacritty/src/event.rs
waynr: There are two alternatives I can think of:

* write a one-off function that compares every field (except for size) of the processor font with the config font to determine if anything other than the size has changed; the problem with this approach in my opinion is that it leaves us open to missing future additions to the `Font` struct
* write a procedural macro to make this function for us, parameterizing it to exclude the `size` field (i'm not certain this is possible, just vaguely suspect it might be); the problem with this approach in my opinion is that it seems unnecessarily complicated.

I went with the current approach because it seems both simple and resistant to future changes to the `Font` struct. I am open to other alternatives though of course.

alacritty/src/event.rs
waynr: I guess I was assuming that if something other than the font size changed we would want the font reset regardless of whether the font size has changed dynamically since starting. But I can see how some might find that behavior surprising.

I think if we pass a `Font` as a parameter to the `Reset` enum in this case we could pass in a copy with the current size overridden by the value on `processor.ctx.font` to ensure that the `Display`'s font doesn't change when its family or style changes.

Based on your reaction to the `config.font.clone()` below, I suspect you might not be comfortable with that in which case I'm not sure how we can change any other part of the font when the size has been dynamically changed.

chrisduerr: Why would this be a problem? Cloning the font to send it as an event shouldn't be a massive problem.

alacritty/src/event.rs
waynr: I think you can forget my last comment here, after https://github.com/jwilm/alacritty/pull/2836#discussion_r338853910 I do think that parameterizing `Reset` can resolve this (though not without a clone of the font still)

alacritty/src/event.rs
waynr: I guess in my ignorance I'm just not quite at the point where I can adequately interpret what was wrong that prompted https://github.com/jwilm/alacritty/pull/2836/files#r338834447; but that is kind of moot now as I am about to push an update that should address all the feedback so far.

alacritty/src/event.rs
```diff
@@ -227,17 +229,17 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
 
     fn change_font_size(&mut self, delta: f32) {
-        self.resize_pending.font_size = Some(FontResize::Delta(delta));
+        self.display_update_pending.font = Some(FontUpdate::SizeDelta(delta));
         self.terminal.dirty = true;
     }
 
     fn reset_font_size(&mut self) {
-        self.resize_pending.font_size = Some(FontResize::Reset);
+        self.display_update_pending.font = Some(FontUpdate::ResetToConfig);
```
waynr: @chrisduerr  just FYI, this is where I found it necessary to distinguish between `ResetTo(Font)` and `ResetToConfig`; at this point in the `ActionContext` we don't have access to the `Font` to which we want to reset. We could get more specific here in the enum value's name, ie `ResetToProcessorConfigFont`, what do you think?

chrisduerr: > at this point in the ActionContext we don't have access to the Font to which we want to reset

In theory it would be possible to modify the font though. We can just store `&'a mut Font` in the context, modify it in here, then stage the latest version of the font in the `DisplayUpdate`.

waynr: I didn't see a good way to use `&'a mut Font` since the font we are interested comes from a `&'a mut Config` when initializing the `ActionContext` just before draining the event queue; I also didn't see a good way to work around that `Config` being mutable since it needs to be replaced whenever the config file is reloaded. Because of this I instead [cloned the `Font` onto the `ActionContext`](https://github.com/jwilm/alacritty/pull/2836/commits/2a7d15ee590efee8024607168228dfda8ffd0b75) to avoid the `ResetToConfig` variant.

I think this does introduce an unlikely race condition where the font size reset command is issued around the exact same time the config file is reloaded such that the two events land on the queue at the same time in which case the font size reset event may end up resetting to the old version of the font size (assuming my understanding of the way event processing works is adequate).

alacritty/src/event.rs
```diff
@@ -392,7 +394,7 @@ impl<N: Notify> Processor<N> {
                 message_buffer: &mut self.message_buffer,
                 display_update_pending: &mut display_update_pending,
                 window: &mut self.display.window,
-                font: &mut self.font,
+                font_size: self.font_size,
```
chrisduerr: You're copying the font size here and later assigning it to `self.font_size` again. Couldn't we just pass `&mut self.font_size`?

alacritty/src/event.rs
```diff
@@ -413,7 +417,9 @@ impl<N: Notify> Processor<N> {
 
                     let size = self.display.window.inner_size().to_physical(dpr);
 
-                    display_update_pending.font = Some(self.font.clone());
+                    if display_update_pending.font.is_none() {
+                        display_update_pending.font = Some(self.config.font.clone());
+                    }
```
chrisduerr: Why do we need to check for `None` here?

alacritty/src/event.rs
```diff
@@ -224,14 +224,16 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     }
 
     fn change_font_size(&mut self, delta: f32) {
-        self.font.size = max(self.font.size + delta, Size::new(FONT_SIZE_STEP));
-        self.display_update_pending.font = Some(self.font.clone());
+        self.font_size = max(self.font_size + delta, Size::new(FONT_SIZE_STEP));
+        let mut font = self.config.font.clone();
+        font.size = self.font_size;
+        self.display_update_pending.font = Some(font);
```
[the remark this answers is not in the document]
waynr: Ah, I didn't know about `with_size`, thanks!

alacritty/src/event.rs
waynr: I already removed it by the time you commented; my thinking was that `display_update_pending.font` might already be `Some` from  earlier in the event handling loop but then I had a closer look at the logic for this block and realized that won't ever happen.

alacritty/src/event.rs
waynr: Good point, I've updated this appropriately. Thanks for the quick review!

alacritty/src/event.rs
```diff
@@ -227,23 +224,30 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     }
 
     fn change_font_size(&mut self, delta: f32) {
-        self.resize_pending.font_size = Some(FontResize::Delta(delta));
+        *self.font_size = max(*self.font_size + delta, Size::new(FONT_SIZE_STEP));
+        self.display_update_pending.font =
+            Some(self.config.font.clone().with_size(*self.font_size));
```
chrisduerr: I'm not sure if I've mentioned this in the previous comment, but since this line is longer than 100 columns, I think it makes sense to extract the stuff inside the `Some` into a `let font = self.config...;`.

If we have two lines anyways, it's usually a bit more readable to have separate statements.

alacritty/src/event.rs
```diff
@@ -460,23 +467,31 @@ impl<N: Notify> Processor<N> {
...
+                        if processor.ctx.config.font != config.font {
+                            let mut font = config.font.clone();
+
+                            // Do not update font size if it has been changed at runtime
+                            if *processor.ctx.font_size != processor.ctx.config.font.size {
+                                font.size = *processor.ctx.font_size;
+                            }
+
+                            processor.ctx.display_update_pending.font = Some(font.clone());
+                            *processor.ctx.font_size = font.size;
```
chrisduerr: I think we can structure this code a bit better:

```rust
if *processor.ctx.font_size == processor.ctx.config.font.size {
    *processor.ctx.font_size = config.font.size;
}

let font = config.font.clone().with_size(*processor.ctx.font_size);
processor.ctx.display_update_pending.font = Some(font);
```

PR #2843, opened 2019-10-03
Simplify logo SVG for better compatibility

jansol:
This removes the fancy masked bit that makes up the white part of the logo in favor of a simple radial gradient. This makes all the SVGs compatible with the SVG Tiny subset that QtSVG supports. Sadly webkit will still butcher some of the blurs, so I also rendered a PNG for the README. The results are somewhat different from the original though, so not sure if this is an acceptable workaround.

Opening this PR since there has been little discussion on the matter beyond acknowledging the issue, and to increase the possibility of having this fixed in 0.3.4, if consensus is to go with these simplified versions :wink: .


jansol:
For comparison: https://imgur.com/a/BcskA0k
top: simplified
bottom: original

jansol:
Changes done, CI failures seem unrelated.

kchibisov:
fwiw you can rebase and push, so CI should pass after it.

README.md
```diff
@@ -140,7 +140,11 @@ scoop install alacritty
 ### Other
 
 Prebuilt binaries for Linux, macOS, and Windows can be downloaded from the
-[GitHub releases page](https://github.com/jwilm/alacritty/releases).
+[GitHub releases page](https://github.com/jwilm/alacritty/releases). Note
+that some desktop environments have trouble rendering SVG icons. For these
```
kchibisov: I guess it should be: `For those`.

chrisduerr: I think both would be fine here, but it might sound a bit strange because of the sentence structure regardless.

This might flow a bit nicer:

```
If your desktop environment has trouble rendering the default SVG icons, you can find a PNG and simplified SVG in the `extra/logo/compat` directory.
```

nixpulvis:
Just for what it's worth, this simplified icon is what I was trying to do in #2707. I'm not a fan of the blurring effect in general, but that might just be me.

So I'm clearly in support of any new uses of the "compat" logo's use.

jansol:
It is done

PR #2845, opened 2019-10-04
validate if working directory configuration exists

tylersouthwick:
Currently, this gives a system failure with no clear indication of what the issue is (a general file not found error)

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -18,7 +18,7 @@ use std::path::PathBuf;
 use clap::{crate_authors, crate_description, crate_name, crate_version, App, Arg};
 use log::{self, LevelFilter};
 
-use alacritty_terminal::config::{Delta, Dimensions, Shell, DEFAULT_NAME};
+use alacritty_terminal::config::{ConfigValidationError, Delta, Dimensions, Shell, DEFAULT_NAME};
 use alacritty_terminal::index::{Column, Line};
 
 use crate::config::Config;
@@ -253,7 +253,10 @@ impl Options {
         self.config.clone()
     }
 
-    pub fn into_config(self, mut config: Config) -> Config {
+    pub fn into_config(
+        self,
+        mut config: Config,
+    ) -> ::std::result::Result<Config, ConfigValidationError> {
         config.set_live_config_reload(
             self.live_config_reload.unwrap_or_else(|| config.live_config_reload()),
         );
@@ -289,7 +292,10 @@ impl Options {
             config.debug.log_level = max(config.debug.log_level, LevelFilter::Info);
         }
 
-        config
+        match config.validate() {
+            None => Ok(config),
+            Some(err) => Err(err),
+        }
     }
 }
 
@@ -303,7 +309,7 @@ mod test {
         let config = Config::default();
         let old_dynamic_title = config.dynamic_title();
 
-        let config = Options::default().into_config(config);
+        let config = Options::default().into_config(config).unwrap();
 
         assert_eq!(old_dynamic_title, config.dynamic_title());
     }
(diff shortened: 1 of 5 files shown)


chrisduerr:
@tylersouthwick If you need another review, feel free to request it in github, otherwise it's difficult for me to keep track. But it looks like you've still got some CI failures to figure out.

Just to make sure you're not waiting on me without me knowing about it. :)

alacritty/src/cli.rs
```diff
@@ -253,7 +253,10 @@ impl Options {
         self.config.clone()
     }
 
-    pub fn into_config(self, mut config: Config) -> Config {
+    pub fn into_config(
+        self,
+        mut config: Config,
+    ) -> ::std::result::Result<Config, ConfigValidationError> {
```
chrisduerr: Do we really want to error here? We have the option to just print an error to stderr and then start in the default directory anyways. That seems like a better idea to me.

alacritty/src/cli.rs
```diff
@@ -289,7 +292,10 @@ impl Options {
             config.debug.log_level = max(config.debug.log_level, LevelFilter::Info);
         }
 
-        config
+        match config.validate() {
```
chrisduerr: A `validate` method doesn't make a lot of sense to me. The config file is valid, the path just isn't accessible, but that doesn't mean the config is invalid. This is also the only case we want to check, so we shouldn't over-complicate things.

alacritty/src/config/mod.rs
```diff
@@ -176,7 +190,11 @@ fn parse_config(contents: &str) -> Result<Config> {
         },
         Ok(config) => {
             print_deprecation_warnings(&config);
-            Ok(config)
+
+            match config.validate() {
+                None => Ok(config),
+                Some(err) => Err(Error::InvalidConfiguration(err)),
+            }
```
chrisduerr: I don't see why we need to check things both for the CLI and for the Config loading. The CLI values just replace the config, so we should just verify after both have been combined.

alacritty/src/main.rs
```diff
@@ -88,7 +89,10 @@ fn main() {
     // Load configuration file
     let config_path = options.config_path().or_else(config::installed_config);
     let config = config_path.map(config::load_from).unwrap_or_else(Config::default);
-    let config = options.into_config(config);
+    let config = options.into_config(config).unwrap_or_else(|err| {
+        error!(target: LOG_TARGET_CONFIG, "Unable to apply command line parameters {}", err);
+        Config::default()
+    });
```
chrisduerr: Resetting to the default config, only because the path isn't valid seems unnecessarily hostile. Why would we throw away all that good info, just because our working directory isn't good?

alacritty_terminal/src/config/mod.rs
```diff
@@ -143,7 +143,32 @@ pub struct Config<T> {
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        Display::fmt(&self.message, f)
+    }
+}
+
 impl<T> Config<T> {
+    pub fn validate(&self) -> Option<ConfigValidationError> {
+        self.working_directory.as_ref().filter(|x| !x.exists()).map(|_| ConfigValidationError {
+            field: String::from("working_directory"),
+            message: String::from("working directory does not exist"),
```
nixpulvis: FWIW, this is our current handling for the config file, maybe this could be similar:

https://github.com/jwilm/alacritty/blob/master/alacritty/src/config/mod.rs#L150

chrisduerr:
Since this issue has been stale for a while, I've implemented my suggestions in https://github.com/alacritty/alacritty/pull/3190.

PR #2846, opened 2019-10-04
Resolve race between DPI change and resize events.

### LINKED ISSUE #2334: When displays are (un)plugged under Windows, parts of the terminal might become invisible
Windows 10 1809

Went from from dock to laptop screen to (crappy, slow) StarLink USB2.0 external graphics card, and afterwards I noticed that the top my vim and bash were hidden out of sight behind and below the window titlebar and buttons.

Basically the OpenGL buffer was displayed offset, so the top was not visible inside the window (or anywhere for that matter).

### LINKED ISSUE #2456: Windows 10 - text rendering artifacts after adjusting DPI scaling in Windows settings
On Windows 10, if you adjust DPI scaling down in the OS, and then lower the font size in Alacritty, text will be rendered with artifacts:
image

This is built on the latest commit.

If you keep hitting ctrl+= or ctrl--, there are certain sizes that don't have artifacts, but others that do. Eventually, after changing the size enough times, the artifacts disappear altogether, but return after a short while.

davidhewitt:
On windows, when the DPI changes, alacritty updates the terminal dimensions as part of the `handle_resize` callback. However, the `PhysicalSize` with which alacritty does the calculation can occasionally be the wrong DPI, depending on the order in which the DPI change and resize events are sent.

I suggest working around this race by asking the window for its current DPI and size when we detect a DPI change.

The race problem seems to occur on [event.rs line 369](https://github.com/davidhewitt/alacritty/blob/d1b08c02d8a670fa163ce6de399d8621289d0735/alacritty_terminal/src/event.rs#L369) when we convert the `LogicalSize` to a `PhysicalSize`. Sometimes the DPR used is correct, but for me the DPR is wrong when I switch back and forth from my laptop display to an external monitor. The result is that the pty has the wrong dimensions (either has loads of empty space or overflows the window, depending on whether the DPR went up or down).

I experimented with delaying the conversion from `LogicalSize` to `PhysicalSize` into the `handle_resize` callback, which also fixed the issue for me, but I couldn't guarantee that it's race free. If you prefer that solution rather than asking the `Window` directly for its size I can change this PR to do that.

Fixes #2456.
Fixes #2334.

### DIFF
--- a/alacritty_terminal/src/display.rs
+++ b/alacritty_terminal/src/display.rs
@@ -397,13 +397,9 @@ impl Display {
         let message_bar_changed = self.last_message != terminal.message_buffer_mut().message();
 
         if font_changed || message_bar_changed {
-            if new_size == None {
-                // Force a resize to refresh things
-                new_size = Some(PhysicalSize::new(
-                    f64::from(self.size_info.width) / self.size_info.dpr * dpr,
-                    f64::from(self.size_info.height) / self.size_info.dpr * dpr,
-                ));
-            }
+            // Force a resize to refresh things. Ask the window directly for its size to avoid
+            // potential races between resize & DPI change events.
+            new_size = self.window.inner_size_pixels().map(|ls| ls.to_physical(dpr));
 
             self.font_size = terminal.font_size;
             self.last_message = terminal.message_buffer_mut().message();


chrisduerr:
@davidhewitt Can this be reproduced in the #2438 PR?

davidhewitt:
Ah, indeed PR #2438 does also resolve this (at least for me)! 

In which case I'm going to close this PR and I guess #2456 and #2334 can be marked as fixed by that.

PR #2855, opened 2019-10-06
Add support for alternate scroll escape

### LINKED ISSUE #2727: Support alternateScroll escape
Currently Alacritty allows sending arrow keys in the alternate screen buffer through the `scrolling.faux_multiplier` configuration option. However, there is a terminal escape sequence dedicated to set and unset this option.

According to the XTerm documentation `CSI ? 1007 h` should enable sending arrow keys in the alternate screen mode and `CSI ? 1007 l` should disable it.

XTerm by default has this option set to be disabled, so the first `CSI ? 1007 h` is required to enable it. With Termite it is enabled by default and `CSI ? 1007 l` has to be used to disable it by default. URxvt does not seem to support this.

Alacritty should respect this option and enable/disable what we call faux scrolling using it. Additionally it might make sense to get rid of our configuration option, however that would remove the ability to have separate scroll modifiers for normal and faux scrolling.

The default of Alacritty should be to enable it, since this follows the example of what Termite does and is consistent with what Alacritty has been doing by itself so far.

zummenix:
Closes #2727

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Live reload font size from config
 - New CLI flag `--hold` for keeping Alacritty opened after its child process exits
 - Escape sequence to save and restore window title from stack
-- Support for alternate scroll escape
+- Support for alternate scroll escape. `\e[?1007h`/`\e[?1007l` escape sequences enable/disable
+alternate scroll (faux scrolling) respectively. Enabled by default.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Live reload font size from config
 - New CLI flag `--hold` for keeping Alacritty opened after its child process exits
 - Escape sequence to save and restore window title from stack
+- Support for alternate scroll escape
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bindings for Super/Command + F1-F12
 - Automatic config generation
+- Deprecated `scrolling.faux_multiplier`, use `scrolling.multiplier` instead
 
 ## 0.3.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,8 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Live reload font size from config
 - New CLI flag `--hold` for keeping Alacritty opened after its child process exits
 - Escape sequence to save and restore window title from stack
-- Support for alternate scroll escape. `\e[?1007h`/`\e[?1007l` escape sequences enable/disable
-alternate scroll (faux scrolling) respectively. Enabled by default.
+- Alternate scroll escape sequence (`CSI ? 1007 h` / `CSI ? 1007 l`)
 
 ### Changed
 
@@ -78,9 +77,8 @@ alternate scroll (faux scrolling) respectively. Enabled by default.
 
 - Bindings for Super/Command + F1-F12
 - Automatic config generation
-- Deprecated `scrolling.faux_multiplier`, use `scrolling.multiplier` for both usual and faux
-scrolling. You can enable/disable faux scrolling using `\e[?1007h`/`\e[?1007l` escape sequences
-respectively.
+- Deprecated `scrolling.faux_multiplier`, the alternate scroll escape can now be used to disable it
+    and `scrolling.multiplier` controls the number of scrolled lines
 
 ## 0.3.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,7 +77,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bindings for Super/Command + F1-F12
 - Automatic config generation
-- Deprecated `scrolling.faux_multiplier`, use `scrolling.multiplier` instead
+- Deprecated `scrolling.faux_multiplier`, use `scrolling.multiplier` for both usual and faux
+scrolling. You can enable/disable faux scrolling using `\e[?1007h`/`\e[?1007l` escape sequences
+respectively.
 
 ## 0.3.3
 
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -206,8 +206,9 @@ fn print_deprecation_warnings(config: &Config) {
     if config.scrolling.faux_multiplier().is_some() {
         warn!(
             target: LOG_TARGET_CONFIG,
-            "Config scrolling.faux_multiplier is deprecated; please use scrolling.multiplier \
-             instead"
+            "Config scrolling.faux_multiplier is deprecated; please use scrolling.multiplier for \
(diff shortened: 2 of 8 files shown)


zummenix:
I think this is ready for review. A couple of questions:
- Could you point out to examples of tests that applicable for this feature? I think ref tests won't work for this or I just don't understand.
- Is naming good?
- The removal of `scrolling.faux_multiplier` setting is still undecided. Do we need it? Personally, I'd remove it for simplicity.

alacritty_terminal/src/term/mod.rs
```diff
@@ -901,6 +902,11 @@ impl<T> Term<T> {
             }
         }
         self.visual_bell.update_config(config);
+        if config.scrolling.faux_multiplier().unwrap_or(config.scrolling.multiplier()) > 0 {
+            self.mode.insert(TermMode::ALTERNATE_SCROLL)
+        } else {
+            self.mode.remove(TermMode::ALTERNATE_SCROLL)
+        }
```
zummenix: Actually, I'm not sure this is needed. `TermMode::ALTERNATE_SCROLL` is enabled by default and even if multiplier is 0, scrolling won't work.

zummenix:
@chrisduerr 
I've rebased my commits. The CI is green. Please take a look.

CHANGELOG.md
```diff
@@ -74,6 +75,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bindings for Super/Command + F1-F12
 - Automatic config generation
+- Deprecated `scrolling.faux_multiplier`, use `scrolling.multiplier` instead
```
chrisduerr: We should probably point out that the new escape can be used to enable/disable faux scrolling completely. We might need to break this over two lines with all the info necessary, but that's not the end of the world.

alacritty/src/config/mod.rs
```diff
@@ -202,6 +202,14 @@ fn print_deprecation_warnings(config: &Config) {
             "Config persistent_logging is deprecated; please use debug.persistent_logging instead"
         );
     }
+
+    if config.scrolling.faux_multiplier().is_some() {
+        warn!(
+            target: LOG_TARGET_CONFIG,
+            "Config scrolling.faux_multiplier is deprecated; please use scrolling.multiplier \
+             instead"
```
chrisduerr: Same as deprecation notice, we need to tell the user both how to enable/disable faux scrolling and how to set the multiplier.

alacritty/src/input.rs
```diff
@@ -551,11 +551,9 @@ impl<'a, T: EventListener, A: ActionContext<T> + 'a> Processor<'a, T, A> {
     fn scroll_terminal(&mut self, modifiers: ModifiersState, new_scroll_px: i32) {
         let mouse_modes =
             TermMode::MOUSE_REPORT_CLICK | TermMode::MOUSE_DRAG | TermMode::MOUSE_MOTION;
+        let alt_screen_modes = TermMode::ALT_SCREEN | TermMode::ALTERNATE_SCROLL;
```
chrisduerr: I'd name this alt_scroll_modes. Since it's not just indicating alt screen.

alacritty/src/input.rs
```diff
@@ -565,11 +563,14 @@ impl<'a, T: EventListener, A: ActionContext<T> + 'a> Processor<'a, T, A> {
             for _ in 0..lines {
                 self.mouse_report(code, ElementState::Pressed, modifiers);
             }
-        } else if self.ctx.terminal().mode().contains(TermMode::ALT_SCREEN)
-            && faux_multiplier > 0
-            && !modifiers.shift
-        {
-            self.ctx.mouse_mut().scroll_px += new_scroll_px * faux_multiplier as i32;
+        } else if self.ctx.terminal().mode().contains(alt_screen_modes) && !modifiers.shift {
```
chrisduerr: ~~We still need to support disabling faux scrolling with the `faux_multiplier` option for the time being.~~

After looking at the rest of the code, this is not necessary since we set the escape sequence at startup according to the `faux_multiplier`. So this comment is purely informational for other readers.

alacritty_terminal/src/config/scrolling.rs
```diff
@@ -12,7 +12,8 @@ pub struct Scrolling {
     #[serde(deserialize_with = "failure_default")]
     multiplier: ScrollingMultiplier,
     #[serde(deserialize_with = "failure_default")]
-    faux_multiplier: ScrollingMultiplier,
+    // TODO: DEPRECATED
+    faux_multiplier: Option<ScrollingMultiplier>,
```
chrisduerr: Usually we move deprecated options to the bottom of the structs to make it easier to find them. You should be able to see this in other places where items have been deprecated too.

Also a newline between the deprecated item and the last non-deprecated will make it stand out even more.

alacritty_terminal/src/config/scrolling.rs
```diff
@@ -26,8 +27,9 @@ impl Scrolling {
         self.multiplier.0
     }
 
-    pub fn faux_multiplier(self) -> u8 {
-        self.faux_multiplier.0
+    // TODO: DEPRECATED
```
chrisduerr: Adding this comment just on the field should be enough. All other instances are automatically brought to our attention by Rust's typesystem once the field is removed :fireworks:.

alacritty_terminal/src/term/mod.rs
```diff
@@ -901,6 +902,11 @@ impl<T> Term<T> {
             }
         }
         self.visual_bell.update_config(config);
+        if config.scrolling.faux_multiplier().unwrap_or_else(|| config.scrolling.multiplier()) > 0 {
```
chrisduerr: We don't want to fallback to `multiplier` here. We just need to check if the faux multiplier is `0`.

We also don't want to remove it otherwise, the default should be to enable it, I'm thinking more like this:

```rust
if let Some(0) == config.scrolling.faux_multiplier() {
    self.mode.remove(TermMode::ALTERNATE_SCROLL);
}
```

The `else` is irrelevant since we set `ALTERNATE_SCROLL` by default.

CHANGELOG.md
```diff
@@ -74,6 +76,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bindings for Super/Command + F1-F12
 - Automatic config generation
+- Deprecated `scrolling.faux_multiplier`, use `scrolling.multiplier` for both usual and faux
+scrolling. You can enable/disable faux scrolling using `\e[?1007h`/`\e[?1007l` escape sequences
+respectively.
```
```
- Deprecated `scrolling.faux_multiplier`, the alternate scroll escape can now be used to disable it and `scrolling.multiplier` controls the number of scrolled lines
```

This should be a little shorter and reduce repetition with the other entry. If a change log entry spans multiple lines, it should also be indented by two additional columns like this:

```
- First line
    Second line
```

alacritty/src/config/mod.rs
```diff
@@ -202,6 +202,15 @@ fn print_deprecation_warnings(config: &Config) {
             "Config persistent_logging is deprecated; please use debug.persistent_logging instead"
         );
     }
+
+    if config.scrolling.faux_multiplier().is_some() {
+        warn!(
+            target: LOG_TARGET_CONFIG,
+            "Config scrolling.faux_multiplier is deprecated; please use scrolling.multiplier for \
+             both usual and faux scrolling. You can enable/disable faux scrolling using \
+             `\\e[?1007h`/`\\e[?1007l` escape sequences respectively"
```
chrisduerr: This should be adjusted similarly to the comment about the change log entry. The `Removed` part is the only relevant thing here, we don't need to clarify what the alternate scroll escape sequence is.

PR #2857, opened 2019-10-08
Support saving window title in a stack

### LINKED ISSUE #2840: Add support for saving window title on stack
Arch Linux

Observed on both

I noticed that when I am using either Vim or NeoVim in Alacritty, the titlestring remains active even after the editor has been closed. I also tested this for other terminal emulators but they all behaved as expected. This occurs both under X and Wayland, I have tested swaywm and i3-gaps. See also the attached video example. The dynamic_title setting is set to true. How could I modify my setup so that Alacritty's default title is restored after exiting Vim?

alacritty-title

DSpeckhals:
This commit adds the concept of a "title stack" to the terminal. Some programs (e.g. vim) send control sequences `CSI 22 ; 0` (push title) and `CSI 23 ; 0` (pop title). Clues on how to handles these were taken from a couple of other terminal emulators ([Gnome Terminal](https://github.com/GNOME/vte/commit/23c44f1f18b7985d9fba3205500d8873ecde51aa), [kitty](https://github.com/kovidgoyal/kitty/commit/3067103b188c55d594918d1525d397cb523f8fea)).

The title stack is mainly just a history of previous titles. Applications can push the current title onto the stack, and pop it back off (setting the window title in the process).

To implement this, the Alacritty `Term` must have two new pieces of state: `title` (the mirror of what is set in the Glutin window) and `title_stack` (the history of titles if pushed onto). When control sequence 22 is received, then a title is pushed, and 23 pops and sets the Glutin window title.

This fixes #2840

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ReceiveChar` key binding action to insert the key's text character
 - Live reload font size from config
 - New CLI flag `--hold` for keeping Alacritty opened after its child process exits
-- Save and restore window title from stack
+- Escape sequence to save and restore window title from stack
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bindings for ScrollToTop and ScrollToBottom actions
 - `ReceiveChar` key binding action to insert the key's text character
 - Live reload font size from config
+- Save and restore window title from stack
 
 ### Changed
 
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -269,7 +269,7 @@ impl Options {
         config.window.embed = self.embed.and_then(|embed| embed.parse().ok());
 
         if let Some(title) = self.title {
-            config.window.set_title(title);
+            config.window.title = title.clone();
         }
 
         if let Some(class) = self.class {
@@ -280,7 +280,7 @@ impl Options {
             }
         }
 
-        config.set_dynamic_title(config.dynamic_title() && config.window.title() == DEFAULT_NAME);
+        config.set_dynamic_title(config.dynamic_title() && config.window.title == DEFAULT_NAME);
 
         config.debug.print_events = self.print_events || config.debug.print_events;
         config.debug.log_level = max(config.debug.log_level, self.log_level);
@@ -301,7 +301,8 @@ mod test {
 
     #[test]
     fn dynamic_title_ignoring_options_by_default() {
-        let config = Config::default();
+        let mut config = Config::default();
+        config.window.title = "Alacritty".to_string();
         let old_dynamic_title = config.dynamic_title();
 
         let config = Options::default().into_config(config);
@@ -324,7 +325,7 @@ mod test {
     fn dynamic_title_overridden_by_config() {
         let mut config = Config::default();
 
-        config.window.set_title("foo".to_owned());
+        config.window.title = "foo".to_owned();
         let config = Options::default().into_config(config);
 
         assert!(!config.dynamic_title());
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -256,7 +256,7 @@ impl Options {
 
         config.window.dimensions = self.dimensions.unwrap_or(config.window.dimensions);
         config.window.position = self.position.or(config.window.position);
-        config.window.title = self.title.or(config.window.title);
+        config.window.title = self.title.unwrap_or(config.window.title);
         config.window.embed = self.embed.and_then(|embed| embed.parse().ok());
 
         if let Some(class) = self.class {
@@ -267,7 +267,7 @@ impl Options {
             }
         }
 
-        config.set_dynamic_title(config.dynamic_title() && config.window.title.is_none());
+        config.set_dynamic_title(config.dynamic_title());
 
         config.debug.print_events = self.print_events || config.debug.print_events;
         config.debug.log_level = max(config.debug.log_level, self.log_level);
@@ -311,7 +311,7 @@ mod test {
     fn dynamic_title_overridden_by_config() {
         let mut config = Config::default();
 
-        config.window.title = Some("foo".to_owned());
+        config.window.title = "foo".to_owned();
         let config = Options::default().into_config(config);
 
         assert!(!config.dynamic_title());
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -256,8 +256,11 @@ impl Options {
 
         config.window.dimensions = self.dimensions.unwrap_or(config.window.dimensions);
         config.window.position = self.position.or(config.window.position);
-        config.window.title = self.title.unwrap_or(config.window.title);
         config.window.embed = self.embed.and_then(|embed| embed.parse().ok());
+
+        if let Some(title) = self.title {
+            config.window.set_title(title);
+        }
 
         if let Some(class) = self.class {
             let parts: Vec<_> = class.split(',').collect();
@@ -267,7 +270,7 @@ impl Options {
             }
         }
 
-        config.set_dynamic_title(config.dynamic_title());
+        config.set_dynamic_title(config.dynamic_title() && config.window.title() == DEFAULT_NAME);
 
         config.debug.print_events = self.print_events || config.debug.print_events;
         config.debug.log_level = max(config.debug.log_level, self.log_level);
@@ -311,7 +314,7 @@ mod test {
     fn dynamic_title_overridden_by_config() {
         let mut config = Config::default();
 
-        config.window.title = "foo".to_owned();
+        config.window.set_title("foo".to_owned());
         let config = Options::default().into_config(config);
 
         assert!(!config.dynamic_title());
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -328,6 +328,12 @@ pub trait Handler {
 
     /// Run the dectest routine
     fn dectest(&mut self) {}
+
+    /// Push a title onto the stack
+    fn push_title(&mut self) {}
+
+    /// Pop the last title from the stack
+    fn pop_title(&mut self) {}
 }
 
 /// Describes shape of cursor
@@ -991,6 +997,13 @@ where
                 handler.clear_line(mode);
             },
             ('S', None) => handler.scroll_up(Line(arg_or_default!(idx: 0, default: 1) as usize)),
+            ('t', None) => match arg_or_default!(idx: 0, default: 1) as usize {
+                22 => handler.push_title(),
+                23 => handler.pop_title(),
+                _ => {
+                    unhandled!();
+                },
+            },
             ('T', None) => handler.scroll_down(Line(arg_or_default!(idx: 0, default: 1) as usize)),
             ('L', None) => {
                 handler.insert_blank_lines(Line(arg_or_default!(idx: 0, default: 1) as usize))
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -418,7 +418,13 @@ impl Mode {
     /// Create mode from a primitive
     ///
     /// TODO lots of unhandled values..
-    pub fn from_primitive(private: bool, num: i64) -> Option<Mode> {
+    pub fn from_primitive(intermediate: Option<&u8>, num: i64) -> Option<Mode> {
+        let private = match intermediate {
+            Some(b'?') => true,
+            None => false,
+            _ => return None,
+        };
+
         if private {
             Some(match num {
                 1 => Mode::CursorKeys,
@@ -883,7 +889,6 @@ where
         }
     }
 
-    #[allow(clippy::cognitive_complexity)]
     #[inline]
     fn csi_dispatch(
         &mut self,
@@ -1008,17 +1013,8 @@ where
                 handler.insert_blank_lines(Line(arg_or_default!(idx: 0, default: 1) as usize))
             },
             ('l', intermediate) => {
-                let is_private_mode = match intermediate {
-                    Some(b'?') => true,
-                    None => false,
-                    _ => {
-                        unhandled!();
-                        return;
-                    },
-                };
                 for arg in args {
-                    let mode = Mode::from_primitive(is_private_mode, *arg);
-                    match mode {
+                    match Mode::from_primitive(intermediate, *arg) {
                         Some(mode) => handler.unset_mode(mode),
                         None => {
                             unhandled!();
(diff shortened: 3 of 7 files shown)


alacritty_terminal/src/ansi.rs
```diff
@@ -991,6 +997,13 @@ where
                 handler.clear_line(mode);
             },
             ('S', None) => handler.scroll_up(Line(arg_or_default!(idx: 0, default: 1) as usize)),
+            ('t', None) => match arg_or_default!(idx: 0, default: 1) as usize {
+                22 => handler.push_title(),
+                23 => handler.pop_title(),
+                _ => {
+                    unhandled!();
+                },
```
chrisduerr: We don't need a dedicated block for this, it can just be written on a single line.

alacritty_terminal/src/term/mod.rs
```diff
@@ -47,6 +47,9 @@ pub mod color;
 /// Used to match equal brackets, when performing a bracket-pair selection.
 const BRACKET_PAIRS: [(char, char); 4] = [('(', ')'), ('[', ']'), ('{', '}'), ('<', '>')];
 
+/// Max size of the window title stack
+const TITLE_STACK_MAX_DEPTH: usize = 8;
```
chrisduerr: Why do we need a max depth?

alacritty_terminal/src/term/mod.rs
```diff
@@ -887,6 +896,8 @@ impl<T> Term<T> {
             clipboard,
             event_proxy,
             is_focused: true,
+            title: config.window.title.as_ref().map_or(DEFAULT_NAME.to_string(), |t| t.to_string()),
```
chrisduerr: We don't need to use `map_or` here, instead this is a bit more clear as to what's actually happening:

```suggestion
            title: config.window.title.clone().unwrap_or_else(|| DEFAULT_NAME.to_owned()),
```

However, I'd assume that we use this in window.rs too and that's where you got this code from, in which case we should really rather centralize it.

I'd recommend either adding a `title` method to our config, or to simply use serde to always deserialize to a `String` instead of `Option<String>` since it seems we do always need it anyways. So we can just use serde's defaults for this.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2089,6 +2102,24 @@ impl<T: EventListener> ansi::Handler for Term<T> {
...
+
+    #[inline]
+    fn push_title(&mut self) {
+        let title = self.title.to_owned();
+        trace!("Pushing '{}' onto title stack", title);
+        if self.title_stack.len() >= TITLE_STACK_MAX_DEPTH {
+            self.title_stack.remove(0);
+        }
+        self.title_stack.push(title);
+    }
```
chrisduerr: This function would benefit from a bit of reordering and empty lines to improve clarity.

First of all we can just put the `trace` right at the top with a newline separating it from everything else and printing `self.title`, we don't need an owned string for that.

Then we don't need to actually define the title as variable at all, it's only used in one place and we can inline it instead.

Instead of using `to_owned` we should also use `clone` instead. This makes it clear that we are cloning an existing string rather than converting a `&str` to a `String`.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2089,6 +2102,24 @@ impl<T: EventListener> ansi::Handler for Term<T> {
...
+
+    #[inline]
+    fn pop_title(&mut self) {
+        if let Some(popped) = self.title_stack.pop() {
+            trace!("Popping '{}' from title stack", popped);
```
chrisduerr: I think we should trace whenever this was called, regardless of there actually is a title or not. We can add a separate `trace` to `set_title` maybe if we care about title updates instead of just escape sequences.

alacritty_terminal/src/ansi.rs
DSpeckhals: Got it. One problem is that this function `csi_dispatch` has a cognitive complexity of 26 vs. the recommended 24. Should we ignore the warning with the function attribute `#[allow(clippy::cognitive_complexity)]` or refactor it?

alacritty_terminal/src/term/mod.rs
DSpeckhals: I saw it like this with Gnome Terminal (8) and Kitty (10). This conversation in the Gnome Terminal bug report explains it: https://bugzilla.gnome.org/show_bug.cgi?id=699819#c8

alacritty_terminal/src/ansi.rs
chrisduerr: I don't see how this would change anything about that? I'm just suggesting to change the formatting.

DSpeckhals: Oh yeah. I'm not suggesting that. With or without the change I get a Clippy warning.

alacritty_terminal/src/term/mod.rs
chrisduerr: > should clear the stack on reset

That's a good point. We should do this too.

I don't think the arguments made in the gnome bugtracker make a lot of sense. They seem to assume that a bad app (talking about faulty for implementation) would only cause troubles the moment the stack is filled, but the process of the application pushing a ton of titles on the stack already messes it up completely anyways, so there's really no point I think.

alacritty_terminal/src/term/mod.rs
DSpeckhals: One thing about changing `title` from `Option<String>` to just `String` is that the override for `dynamic_title` depends on checking for the value's existence. Here's a couple of options I can think of to resolve this:

- Check if the title is equal to the default value "Alacritty" instead. This seems a little bit like a magic string if you ask me.
- Keep it as an `Option<String>` and duplicate the logic (`unwrap_or_else` with some sort of utility function might work).
- Don't override `dynamic_title` based on the `title` config field being defined. This would be a breaking change, of course.

Do you have any suggestions?

alacritty_terminal/src/term/mod.rs
DSpeckhals: I see what you mean. I'm happy to remove that logic if you find it unnecessary. I'll also clear the stack on reset. It would be nice if we could restore the original title as well, but that would require the terminal having the original saved in it's state somewhere.

alacritty_terminal/src/term/mod.rs
DSpeckhals: Agreed. I'll change that.

alacritty_terminal/src/term/mod.rs
nixpulvis: :+1: for centralizing the default. We will now have duplication between the window value `title` and the term value `title`, ideally we should only have one. Either the `Term` should populate the title by default with the value of the `Window`'s title, or the window should be wired up to use the `Term` default IMO. This isn't necessarily going to look super clean since the `Display` (which owns the `Window`) is pretty separate from the `Term`.

I'm wondering if it doesn't make sense to reuse the `Event::Title` to send the default from the `Term`. Just a thought...

alacritty_terminal/src/term/mod.rs
```diff
@@ -760,6 +763,12 @@ pub struct Term<T> {
 
     /// Terminal focus
     pub is_focused: bool,
+
+    /// Title of the window
+    title: String,
+
+    /// The stack of titles; the most recent is last
+    title_stack: Vec<String>,
```
nixpulvis: I almost hate to say it (I generally don't like long names), but would `saved_title_stack` or `pushed_title_stack` make things a bit better? Otherwise it may be a little bit less clear why we have a `title` and a `title_stack`. Otherwise I'd just expand the comment a tiny bit to say something about how popping a title from this stack effects the `title`.

alacritty_terminal/src/term/mod.rs
nixpulvis: This makes sense to me, in fact one of the first things I tested was this:

```fish
while true; echo -en "\033[22;0t"; end
```

:tada:

DSpeckhals:
I just pushed a commit that addresses some of the review comments. Some unresolved issues:

- [ ] Duplication of window name in the Terminal and Window.
- [ ] Should there be a maximum title stack size?
- [ ] How should the lack of a `None` value for `title` be handled, e.g. overriding `dynamic_title`?

nixpulvis:
> Should there be a maximum title stack size?

I don't see why we wouldn't want a maximum size. I'm generally of the opinion that we should perform well even when users are abusing us to the best of our abilities. In this case limiting the stack size to avoid a runaway title push seems completely reasonable. Though I wouldn't expect this to come up much, so maybe it's not a big deal either way.

We could certainly support a large maximum however, if we choose to go this route.

alacritty_terminal/src/term/mod.rs
chrisduerr: > We will now have duplication between the window value title and the term value title

This is incorrect, we do not store the title on the window.

> One thing about changing title from Option<String> to just String is that the override for dynamic_title depends on checking for the value's existence.

Seems like the simplest option is to just add a `title` method to the config then.

alacritty_terminal/src/ansi.rs
chrisduerr: We should try to resolve cognitive complexity errors rather than ignoring them. I can take a look if you're having trouble figuring something out. Though this has been an issue with this function multiple times before.

alacritty_terminal/src/term/mod.rs
chrisduerr: > It would be nice if we could restore the original title as well, but that would require the terminal having the original saved in it's state somewhere.

I think on reset we can just reset to `Alacritty` honestly.

chrisduerr:
> I don't see why we wouldn't want a maximum size. I'm generally of the opinion that we should perform well even when users are abusing us to the best of our abilities. In this case limiting the stack size to avoid a runaway title push seems completely reasonable. Though I wouldn't expect this to come up much, so maybe it's not a big deal either way.

I don't see why we would bother. There already are enough things you can totally mess up if you're malicious. Protecting against a malicious user is completely pointless. The only thing we might want to protect against are accidents and since a vector scales pretty freely all this should do is nuke the user's memory, which he is free to fix after he fucked up.

I'd definitely check what happens when running `while true; echo -en "\033[22;0t"; end` without limitation on stack size and judge depending on failure case. But generally I don't see what concrete benefit we would gain in practice.

alacritty_terminal/src/ansi.rs
DSpeckhals: I'll start looking and see what I can do. Any advice is appreciated, of course!

alacritty_terminal/src/term/mod.rs
nixpulvis: @chrisduerr, oh your totally right. I think I was looking at the `WindowConfig`, which of course makes sense to have a title option.

Sorry for the confusion.

alacritty_terminal/src/ansi.rs
DSpeckhals: It's been interesting trying to understand cognitive complexity...I've never come across this Clippy lint until now.

One thing I found would be possible, is to combine the `'l'` and `'t'` arms. It lowers the cognitive complexity, but perhaps it's a little contrived. What do you think?

```rust
(action @ 'l', intermediate) | (action @ 'h', intermediate) => {
    let is_private_mode = match intermediate {
        Some(b'?') => true,
        None => false,
        _ => {
            unhandled!();
            return;
        },
    };
    for arg in args {
        let mode = Mode::from_primitive(is_private_mode, *arg);
        match (action, mode) {
            ('l', Some(mode)) => handler.unset_mode(mode),
            ('h', Some(mode)) => handler.set_mode(mode),
            _ => {
                unhandled!();
                return;
            },
        }
    }
},
```

alacritty_terminal/src/term/mod.rs
DSpeckhals: > Seems like the simplest option is to just add a title method to the config then.

Usually we'd make the `title` field private in this case, and always access it with the `title()` method which still wouldn't indicate if it was defined in the config or not.

DSpeckhals:
> I'd definitely check what happens when running while true; echo -en "\033[22;0t"; end without limitation on stack size and judge depending on failure case.

With 8Gb of RAM on Ubuntu 19.04, it took about 5 minutes until I started swapping to disk. I didn't wait for swap to fill up. Closing Alacritty cleared the RAM up. Nothing locked up. There was some lag, but I think that was because of the `while true`.

chrisduerr:
> With 8Gb of RAM on Ubuntu 19.04, it took about 5 minutes until I started swapping to disk. I didn't wait for swap to fill up. Closing Alacritty cleared the RAM up. Nothing locked up. There was some lag, but I think that was because of the while true.

Thanks for testing, that sounds about right.

After thinking about it for a bit, this sanity check is probably fine to keep around just to make sure nobody will ever create an Alacritty bug because an application messes up stuff. But I'd increase the title limit significantly to something like `4096` (or at least `1024` I think). That way no application should ever run into it and the limit is virtually non-existent while still having a safety fallback in place.

alacritty_terminal/src/term/mod.rs
chrisduerr: I honestly don't feel like `saved_` or `pushed_` improve things much, but I do see the issue itself.

I think with proper documentation on the field, this should be fine though. But I'd steer away from language like `the most recent is last`, since it might also mean that the current title is the last element. So we definitely need to be clear here.

alacritty_terminal/src/term/mod.rs
chrisduerr: > Usually we'd make the title field private in this case, and always access it with the title() method which still wouldn't indicate if it was defined in the config or not.

That sounds like what I had in mind, yeah.

alacritty_terminal/src/ansi.rs
chrisduerr: I don't get any cognitive complexity warning in your latest revision, is this still relevant?

alacritty_terminal/src/term/mod.rs
DSpeckhals: I updated the comments to be a bit more explanatory and clear.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ReceiveChar` key binding action to insert the key's text character
 - Live reload font size from config
 - New CLI flag `--hold` for keeping Alacritty opened after its child process exits
+- Save and restore window title from stack
```
chrisduerr: I think we should clarify that this is an escape.

```suggestion
- Escape sequence to save and restore window title from stack
```

alacritty/src/cli.rs
```diff
@@ -277,7 +280,7 @@ impl Options {
             }
         }
 
-        config.set_dynamic_title(config.dynamic_title() && config.window.title.is_none());
+        config.set_dynamic_title(config.dynamic_title() && config.window.title() == DEFAULT_NAME);
```
chrisduerr: Didn't we say the only reason why we need title to be an `Option` is because we set dynamic title depending on it?

If we change things here to match against `DEFAULT_NAME`, we don't need it as an option anymore and as a result we don't need to have setters and getters, right? Or did I forget about a different reason we did things with optionals?

Seems like it might result in a cleaner solution too, we don't need to support setting the title to `Alacritty` to disable dynamic title anyways, so it should be fine.

CHANGELOG.md
DSpeckhals: Done.

alacritty/src/cli.rs
DSpeckhals: Understood. I pushed a commit that changes it from `Option<String>` to just `String`.

PR #2868, opened 2019-10-10
Disable depth and stencil buffer allocation

Jasu:
Disable allocation of depth and stencil buffers. This reduces active GPU memory consumption by almost a third, at least on Linux.

If I have a large number of terminals open and there is memory pressure, switching between terminals takes a while, presumably because some required GEM objects are swapped to disk. This change makes the number and size of required GEM objects smaller, hopefully alleviating the issue.

**I have only tested this on Linux (X11).** If passing 0 as the depth/stencil buffer size has different semantics on other platforms with glutin, things might break.

With this branch: Note that the amount of active objects is 27MB. (the inactive object count is higher, since I just ran vtebench)

```
archaea# cat i915_gem_objects|grep alacritty
alacritty: 32 objects, 79409152 bytes (27426816 active, 51798016 inactive, 0 global, 29360128 shared, 184320 unbound, 0 closed)
alacritty: 26 objects, 67796992 bytes (0 active, 67694592 inactive, 0 global, 29360128 shared, 102400 unbound, 0 closed)
```

With master, 47MB of active objects are needed.

```
archaea# cat i915_gem_objects|grep alacritty
alacritty: 31 objects, 79147008 bytes (0 active, 78962688 inactive, 0 global, 29360128 shared, 184320 unbound, 0 closed)
alacritty: 26 objects, 67796992 bytes (47620096 active, 20074496 inactive, 0 global, 29360128 shared, 102400 unbound, 0 closed)
```

This had no effect on vtebench results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Crash when resuming after suspension
 - Crash when trying to start on X11 with a Wayland compositor running
 - Crash with a virtual display connected on X11
+- GPU memory usage has been decreased by disabling allocation of depth and stencil buffers
 
 ### Removed
 
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -117,6 +117,8 @@ fn create_gl_window(
     let windowed_context = ContextBuilder::new()
         .with_srgb(srgb)
         .with_vsync(true)
+        .with_depth_buffer(0)
+        .with_stencil_buffer(0)
         .with_hardware_acceleration(None)
         .build_windowed(window, event_loop)?;
 


chrisduerr:
I don't think we should need stencil or depth buffers, so this seems like a great idea.

I'd like to make absolutely sure this doesn't behave strangely on macOS/Windows though, so I'll test that tomorrow.

kchibisov:
I don't see any difference in rendering on Wayland.

nixpulvis:
Awesome, thanks for looking into this. Any chances to save memory and/or cycles is a great thing. :tada: 

Just an initial test of macOS looks fine. I've verified that it still works as it did before in both windowed and fullscreen modes. I've also tested transparency hoping that it might have an effect on the artifacts being left behind, and sadly it doesn't fix this, though I really wouldn't have expected it to, and otherwise works as I'd expect.

Jasu:
@chrisduerr I force-pushed a line to the changelog.

nixpulvis:
@Jasu, so I've looked a bit more into this, and while it doesn't seem to break any of the platforms I tested, I'm unsure it's making a meaningful change. We already disable the depth mask in the renderer, so AFAICT this is just setting an effectively unused configuration value to `0` instead of whatever else it would be.

https://github.com/jwilm/alacritty/blob/2f83f2dd2e131be5ba769655480da0b5fe8c0dc3/alacritty_terminal/src/renderer/mod.rs#L556-L557

If there's something I'm missing, please let me know. Otherwise I'd be curious if your memory benchmarks are really reproducible. If you can reproduce, would you mind sharing the methods used to test this on Linux?

chrisduerr:
As far as I know, `glDepthMask` sets if writing to the depth buffer is enabled or not. I think by default it's not even enabled.

However, what glutin's depth setting does is setup the buffer itself. So setting it to `0` doesn't disable writing to the buffer, but instead makes sure the buffer is zero-sized.

Jasu:
I checked with renderdoc and it reported that the front buffer had depth and stencil buffers with master but not with the changes. 

Also, I ran ftrace on alacritty tracing GEM object allocations. Only master had these additional lines:

```
           <...>-563762 [002] 55438.131380: i915_gem_object_create: obj=0xffff9b8b9e7cf000, size=0xa00000
           <...>-563762 [002] 55438.131395: i915_gem_object_create: obj=0xffff9b8b9e7cc900, size=0x180000
           <...>-563762 [002] 55438.131414: i915_gem_object_create: obj=0xffff9b89ebbcc900, size=0x280000
```

Also, running `INTEL_DEBUG=fbo,dri,buf ~/ThirdPartySources/alacritty/target/release/alacritty -e non-existent-command` emits the following lines on master only, which corresponds to the trace about GEM allocs:

```
intel_alloc_private_renderbuffer_storage: GL_DEPTH_COMPONENT: MESA_FORMAT_Z24_UNORM_X8_UINT (1800x1350)
bo_create: buf 9 (isl-miptree) 10039296b
bo_create: buf 10 (aux-miptree) 1351680b
intel_alloc_private_renderbuffer_storage: GL_STENCIL_INDEX: MESA_FORMAT_S_UINT8 (1800x1350)
bo_create: buf 11 (isl-miptree) 2613248b
```

kchibisov:
I don't see any difference on Windows either.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - On Windows, query DirectWrite for recommended anti-aliasing settings
+- GPU memory usage has been decreased by disabling allocation of depth and stencil buffers
```
chrisduerr: I think I'd move this to the `fixed` section, usually `changed` is something that we intentionally did one way and now do another way. If we would have done it differently from the beginning if we would have known, it's a fix.

Jasu:
@chrisduerr I moved the comment to the bottom of *fixed* and force-pushed the branch.

Clyybber:
It seems like this has been reverted, was it causing issues?

kchibisov:
@Clyybber yes, for more info https://github.com/jwilm/alacritty/issues/2881 .

PR #2871, opened 2019-10-12
Remove copypasta dependency from alacritty_terminal

kchibisov:
Remove the copypasta dependency used by alacritty_terminal,
which pulls some gui libraries like xcb.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -43,6 +43,7 @@ use alacritty_terminal::tty;
 use alacritty_terminal::util::{limit, start_daemon};
 
 use crate::cli::Options;
+use crate::clipboard::Clipboard;
 use crate::config;
 use crate::config::Config;
 use crate::display::Display;
@@ -67,6 +68,7 @@ impl DisplayUpdate {
 pub struct ActionContext<'a, N, T> {
     pub notifier: &'a mut N,
     pub terminal: &'a mut Term<T>,
+    pub clipboard: &'a mut Clipboard,
     pub size_info: &'a mut SizeInfo,
     pub mouse: &'a mut Mouse,
     pub received_count: &'a mut usize,
@@ -110,7 +112,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     fn copy_selection(&mut self, ty: ClipboardType) {
         if let Some(selected) = self.terminal.selection_to_string() {
             if !selected.is_empty() {
-                self.terminal.clipboard().store(ty, selected);
+                self.clipboard.store(ty, selected);
             }
         }
     }
@@ -280,6 +282,10 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         self.urls
     }
 
+    fn clipboard(&mut self) -> &mut Clipboard {
+        self.clipboard
+    }
+
     /// Spawn URL launcher when clicking on URLs.
     fn launch_url(&self, url: Url) {
         if self.mouse.block_url_launcher {
@@ -357,6 +363,7 @@ pub struct Processor<N> {
     mouse: Mouse,
     received_count: usize,
     suppress_chars: bool,
+    clipboard: Clipboard,
     modifiers: ModifiersState,
     config: Config,
     message_buffer: MessageBuffer,
@@ -375,6 +382,11 @@ impl<N: Notify + OnResize> Processor<N> {
         config: Config,
         display: Display,
     ) -> Processor<N> {
+        #[cfg(not(any(target_os = "macos", windows)))]
+        let clipboard = Clipboard::new(display.window.wayland_display());
+        #[cfg(any(target_os = "macos", windows))]
+        let clipboard = Clipboard::new();
+
         Processor {
             notifier,
             mouse: Default::default(),
@@ -386,6 +398,7 @@ impl<N: Notify + OnResize> Processor<N> {
             message_buffer,
             display,
             event_queue: Vec::new(),
+            clipboard,
         }
     }
 
@@ -471,6 +484,7 @@ impl<N: Notify + OnResize> Processor<N> {
                 terminal: &mut terminal,
                 notifier: &mut self.notifier,
                 mouse: &mut self.mouse,
+                clipboard: &mut self.clipboard,
                 size_info: &mut self.display.size_info,
                 received_count: &mut self.received_count,
                 suppress_chars: &mut self.suppress_chars,
@@ -567,6 +581,13 @@ impl<N: Notify + OnResize> Processor<N> {
                     processor.ctx.message_buffer.push(message);
                     processor.ctx.display_update_pending.message_buffer = true;
                     processor.ctx.terminal.dirty = true;
+                },
+                Event::ClipboardStore(clipboard_type, content) => {
+                    processor.ctx.clipboard.store(clipboard_type, content);
+                },
+                Event::ClipboardLoad(clipboard_type, transform) => {
+                    let text = transform(processor.ctx.clipboard.load(clipboard_type).as_str());
+                    processor.ctx.write_to_pty(text.into_bytes());
                 },
                 Event::MouseCursorDirty => processor.reset_mouse_cursor(),
                 Event::Exit => (),
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -45,6 +45,7 @@ use alacritty_terminal::term::{SizeInfo, Term};
 use alacritty_terminal::util::start_daemon;
 use alacritty_terminal::vi_mode::ViMotion;
 
+use crate::clipboard::Clipboard;
 use crate::config::{Action, Binding, Config, Key, ViAction};
 use crate::event::{ClickState, Mouse};
 use crate::url::{Url, Urls};
@@ -90,6 +91,7 @@ pub trait ActionContext<T: EventListener> {
     fn message(&self) -> Option<&Message>;
     fn config(&self) -> &Config;
     fn event_loop(&self) -> &EventLoopWindowTarget<Event>;
+    fn clipboard(&mut self) -> &mut Clipboard;
     fn urls(&self) -> &Urls;
     fn launch_url(&self, url: Url);
     fn mouse_mode(&self) -> bool;
@@ -140,11 +142,11 @@ impl<T: EventListener> Execute<T> for Action {
             #[cfg(not(any(target_os = "macos", windows)))]
             Action::CopySelection => ctx.copy_selection(ClipboardType::Selection),
             Action::Paste => {
-                let text = ctx.terminal_mut().clipboard().load(ClipboardType::Clipboard);
+                let text = ctx.clipboard().load(ClipboardType::Clipboard);
                 paste(ctx, &text);
             },
             Action::PasteSelection => {
-                let text = ctx.terminal_mut().clipboard().load(ClipboardType::Selection);
+                let text = ctx.clipboard().load(ClipboardType::Selection);
                 paste(ctx, &text);
             },
             Action::Command(ref program) => {
@@ -894,7 +896,7 @@ mod tests {
     };
     use glutin::event_loop::EventLoopWindowTarget;
 
-    use alacritty_terminal::clipboard::{Clipboard, ClipboardType};
+    use alacritty_terminal::clipboard::ClipboardType;
     use alacritty_terminal::event::{Event as TerminalEvent, EventListener};
     use alacritty_terminal::grid::Scroll;
     use alacritty_terminal::index::{Point, Side};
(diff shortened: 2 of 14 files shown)


alacritty/src/event.rs
```diff
@@ -438,6 +439,22 @@ impl<N: Notify> Processor<N> {
         self.write_ref_test_results(&terminal.lock());
     }
 
+    fn paste<T: EventListener, A: crate::input::ActionContext<T>>(ctx: &mut A, contents: &str) {
```
kchibisov: I have no idea where to place this function, we can make it free or keep here. What do you think?

alacritty_terminal/src/clipboard.rs
```diff
@@ -24,74 +25,139 @@ use copypasta::wayland_clipboard;
...
+            ClipboardStore {
+                clipboard: Box::new(ClipboardContext::new().unwrap()),
+                selection: Some(Box::new(
+                    X11ClipboardContext::<X11SelectionClipboard>::new().unwrap(),
+                )),
+            }
+        };
 
+        let (tx, rx) = mpsc::channel();
         Self {
```
kchibisov: I've followed the implementation of our config watcher thread. In case we need some kind of graceful shutdown, let me know.

alacritty/src/event.rs
```diff
@@ -450,6 +467,9 @@ impl<N: Notify> Processor<N> {
         match event {
             GlutinEvent::UserEvent(event) => match event {
                 Event::Title(title) => processor.ctx.window.set_title(&title),
+                Event::Paste(content) => {
+                    Self::paste(&mut processor.ctx, &content);
+                },
```
chrisduerr: This can be changed to a single line.

alacritty/src/event.rs
chrisduerr: We could put this on the `Term` and pass a `<W: io::Write>`, similar to what we do with ansi handlers?

That would also allow potential consumers of `Term` to make use of Alacritty's support for bracketed pasting.

alacritty/src/input.rs
```diff
@@ -150,14 +150,12 @@ impl<T: EventListener> Execute<T> for Action {
                 ctx.copy_selection(ClipboardType::Clipboard);
             },
             Action::Paste => {
-                let text = ctx.terminal_mut().clipboard().load(ClipboardType::Clipboard);
-                paste(ctx, &text);
+                ctx.terminal_mut().clipboard().load(ClipboardType::Clipboard);
             },
```
chrisduerr: I think inlining this should still be under the line limit.

alacritty_terminal/src/clipboard.rs
```diff
@@ -24,74 +25,139 @@ use copypasta::wayland_clipboard;
 use copypasta::x11_clipboard::{Primary as X11SelectionClipboard, X11ClipboardContext};
 use copypasta::{ClipboardContext, ClipboardProvider};
 
+use crate::event::{Event, EventListener};
+use crate::util;
+
 pub struct Clipboard {
+    _thread: std::thread::JoinHandle<()>,
```
chrisduerr: We shouldn't name our variables with underscore.

alacritty_terminal/src/clipboard.rs
```diff
@@ -24,74 +25,139 @@ use copypasta::wayland_clipboard;
...
+        let clipboard = ClipboardStore {
+            clipboard: Box::new(ClipboardContext::new().unwrap()),
+            selection: None,
+        };
+
+        let (tx, rx) = mpsc::channel();
+        Self {
+            _thread: util::thread::spawn_named("clipboard", move || {
+                clipboard_thread(clipboard, rx, event_proxy);
+            }),
```
chrisduerr: Why do we permanently spawn a thread for the clipboard instead of spawning one on-demand? That's not really great for system resources to just have a useless thread hanging around for no reason most of the time.

alacritty_terminal/src/clipboard.rs
```diff
@@ -24,74 +25,139 @@ use copypasta::wayland_clipboard;
...
-
 impl Clipboard {
     pub fn store(&mut self, ty: ClipboardType, text: impl Into<String>) {
-        let clipboard = match (ty, &mut self.selection) {
-            (ClipboardType::Selection, Some(provider)) => provider,
-            (ClipboardType::Selection, None) => return,
-            _ => &mut self.clipboard,
-        };
+        // Our clipboard shouldn't crash Alacritty
+        let _ = self.tx.send(ClipboardRequest { action: ClipboardAction::Store(text.into()), ty });
```
chrisduerr: Storing should work without the requirement for a thread.

alacritty/src/event.rs
kchibisov: The problem is, that we are getting terminal from ActionContext and io::Write should be implemented for it too, however we're ending up with &mut for Writer(rust requirements) and &ctx for terminal, so borrow checker just doesn't allow this code(even its 100% safe).

Generally this type of questions can be solved with RefCell etc, however this is not the case, unfortunately.

So the solution could be: create WriteTerm trait (e.g.) to reflect immutability, or to pass `write_to_pty` directly to term and use it for paste there.

***
Or just leave things as it is.

chrisduerr: > Or just leave things as it is.

I think that it makes sense where it is. Let's just go with this.

Just please don't put it in between the event loop and the event handler, I want to keep those right below each other.

nixpulvis:
How long could the time between the user trying to paste and the actual paste be? Can we ensure that no user input is inserted before the paste comes in? This seems difficult and somewhat expensive.

Otherwise it irritates me a bit to think that a paste could non-deterministically show up in my grid. How do other Wayland compatible terminals handle this?

kchibisov:
Paste should be instant,  if it’s not then something is bad with compositor, etc. however we shouldn’t look the terminal during clipboard data reqwest.

chrisduerr:
Okay, so I've looked through the PR and I think there's some refactoring we can do to improve readibility and make the code structure a bit easier. I've tested this out myself and I think it looks pretty good:

<details>

```diff
diff --git a/alacritty_terminal/src/clipboard.rs b/alacritty_terminal/src/clipboard.rs
index 78d00d8..b3cab8d 100644
--- a/alacritty_terminal/src/clipboard.rs
+++ b/alacritty_terminal/src/clipboard.rs
@@ -15,8 +15,9 @@
 #[cfg(not(any(target_os = "macos", target_os = "windows")))]
 use std::ffi::c_void;
 use std::sync::{Arc, Mutex};
+use std::error::Error;
 
-use log::{debug, warn};
+use log::debug;
 
 use copypasta::nop_clipboard::NopClipboardContext;
 #[cfg(not(any(target_os = "macos", target_os = "windows")))]
@@ -97,39 +98,48 @@ impl Clipboard {
 
 impl Clipboard {
     pub fn store(&mut self, ty: ClipboardType, text: impl Into<String>) {
-        if let Ok(mut clipboard) = self.clipboard.try_lock() {
+        let text = text.into();
+
+        self.with_clipboard_thread(move |clipboard| {
             let clipboard = match (ty, &mut clipboard.selection) {
                 (ClipboardType::Selection, Some(provider)) => provider,
-                (ClipboardType::Selection, None) => return,
+                (ClipboardType::Selection, None) => return Ok(()),
                 _ => &mut clipboard.clipboard,
             };
 
-            clipboard.set_contents(text.into()).unwrap_or_else(|err| {
-                warn!("Unable to store text in clipboard: {}", err);
-            });
-        } else {
-            debug!("Clipboard is busy, skipping store.")
-        }
+            clipboard.set_contents(text)
+        });
     }
 
     pub fn request_load(&mut self, ty: ClipboardType) {
-        let clipboard = self.clipboard.clone();
         let event_proxy = self.event_proxy.clone();
+
+        self.with_clipboard_thread(move |clipboard| {
+            let clipboard = match (ty, &mut clipboard.selection) {
+                (ClipboardType::Selection, Some(provider)) => provider,
+                _ => &mut clipboard.clipboard,
+            };
+
+            let text = clipboard.get_contents()?;
+            event_proxy.lock().unwrap().send_event(Event::Paste(text));
+
+            Ok(())
+        });
+    }
+
+    // TODO: Issue with locking, too many clipboard.lock()s cause input to output much longer than
+    // holding
+    fn with_clipboard_thread<F>(&mut self, f: F)
+    where
+        F: FnOnce(&mut ClipboardStore) -> Result<(), Box<dyn Error>> + Send + 'static
+    {
+        let clipboard = self.clipboard.clone();
+
         util::thread::spawn_named("clipboard", move || {
-            if let Ok(mut clipboard) = clipboard.try_lock() {
-                let clipboard = match (ty, &mut clipboard.selection) {
-                    (ClipboardType::Selection, Some(provider)) => provider,
-                    _ => &mut clipboard.clipboard,
-                };
-
-                let text = clipboard.get_contents().unwrap_or_else(|err| {
-                    debug!("Unable to load text from clipboard: {}", err);
-                    String::new()
-                });
-
-                event_proxy.lock().unwrap().send_event(Event::Paste(text));
-            } else {
-                debug!("Clipboard is busy, skipping load.")
+            if let Ok(mut clipboard) = clipboard.lock() {
+                if let Err(err) = f(&mut clipboard) {
+                    debug!("Clipboard error: {}", err);
+                }
             }
         });
     }
```

</details>

I've opted to put both loading and storing into threads, since it doesn't complicate the code any more but provides additional functionality that would likely be desired.

However as mentioned in this diff, there are some problems both with the existing approach and with this diff. The existing PR makes the choice to just drop events when it can't keep up, which isn't really great since we kinda have threads so we don't need to discard stuff when we can't keep up. However my approach never drops anything but when holding paste, it will output long after it has been released already.

Other terminal emulators perform much, much better here (I've tested only Termite, since URxvt doesn't paste out of the box, and it doesn't drop or is so slow that it can't keep up). So we should probably improve things here to handle this better. If I had to take a guess I'd assume other terminal emulators will go for the choice to paste much later than after releasing the key, but are actually fast enough to keep up with holding paste for text. Though more investigation for text that doesn't return immediately with VTE would likely be helpful.

@kchibisov Hopefully this is a review you can work with. If you need more insights I can take a deeper look at how different terminal emulators do this. However we should try to do better here.

dnkl:
> Paste should be instant, if it’s not then something is bad with compositor, etc.

Just wanted to drop this in case you weren't aware of it...

On Wayland, the compositor is just the facilitator. The receiving program (i..e where the user is pasting) sets up a pipe and hands the write-end to the compositor, which passes it on to the paste source (where the user copied text).

It's then up to that other program to actually write to the pipe, and eventually close it. You can't receive data faster than it writes it.

You might also want to consider malicious programs that writes really really slow on purpose, or never closes the pipe and so on...

chrisduerr:
@dnkl That is true and the same "issue" exists on X11.

However I believe the described scenario was more about a general good-case scenario.

Of course Alacritty should not explode when the application takes seconds to respond, however it should not always take seconds when other applications can clearly do it "instantly".

kchibisov:
Just to give more clarity here.

>Other terminal emulators perform much, much better here (I've tested only Termite, since URxvt doesn't paste out of the box, and it doesn't drop or is so slow that it can't keep up). So we should probably improve things here to handle this better. If I had to take a guess I'd assume other terminal emulators will go for the choice to paste much later than after releasing the key, but are actually fast enough to keep up with holding paste for text. Though more investigation for text that doesn't return immediately with VTE would likely be helpful.

This is mostly our clipboard dependencies problem, like our X11 clipboard backend can't keep up with whatever we have here(it has ~50ms delay, just by processing event queue, etc). However this shouldn't stop this PR to progress, since we can't do anything with it in alacritty, the only thing here is to throttle or not to throttle, I feel like if your clipboard is pretty fast throttling shouldn't affect it tbh, however it'll prevent from blowing up your terminal, I can test it on Wayland, since it should keep up with the high repeat rate. If it can work well with throttling than I don't see a point in a non throttling approach.

chrisduerr:
> This is mostly our clipboard dependencies problem, like our X11 clipboard backend can't keep up with whatever we have here(it has ~50ms delay, just by processing event queue, etc). However this shouldn't stop this PR to progress

I'm not so sure about this. I don't think it's a good idea to move forward with this PR, if the clipboard backends aren't in a usable state yet. I suspect that the performance of the different backends might have an impact on the final decision and design.

MacOS works flawlessly on the latest master for example, so maybe there wouldn't even be a need to over-complicate things there and run on a different thread, but rather just a reasonable timeout and blocking operations might be sufficient. However if we can't get reasonable performance out of our backends, that'll likely mean this approach would be completely impossible.

kchibisov:
Well, the only backend left is X11, since Wayland/macOS are both fine. and X11 clipboard requires some rework.

kchibisov:
It turned out that I already had this branch with name `move-clipboard` and I pushed one more branch with the exact same name, I guess this PR will be for a different purpose.

PR #2872, opened 2019-10-12
Use proper path logging formatting

kchibisov:
It was discovered that we were logging path with `\\` instead of `\` as
separators on Windows due to use of Debug formatting instead of Display
for paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,7 +69,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Crash when trying to start on X11 with a Wayland compositor running
 - Crash with a virtual display connected on X11
 - GPU memory usage has been decreased by disabling allocation of depth and stencil buffers
-- Fix path logging on Windows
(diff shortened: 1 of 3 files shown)


kchibisov:
For this one? I think debug logging is something hided from users.

chrisduerr:
Then why do we need to fix it?

kchibisov:
For me to find my default config when testing Windows.

kchibisov:
But sure, I'll add one.

CHANGELOG.md
```diff
@@ -69,6 +69,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Crash when trying to start on X11 with a Wayland compositor running
 - Crash with a virtual display connected on X11
 - GPU memory usage has been decreased by disabling allocation of depth and stencil buffers
+- Fix path logging on Windows
```
chrisduerr: I'd be more precise here.

```suggestion
- Use `\` instead of `\\` as path separators on Windows for logging config file location
```

PR #2875, opened 2019-10-13
Set stencil buffer to 1

kchibisov:
It was discovered that on some GPUs (Rx 570 + amdgpu driver) this leads to artifacts during startup on X11.

The bug is presented with or without compositor running.

video
https://www.dropbox.com/s/5xcni3cs6p263lc/revert.mkv?dl=0

This reverts commit 3475e449870b382cda4ea6d48f980577cd8c929e.

cc @Jasu

P.s.
I've initially tested only my Wayland system, which is absolutely fine with this changes, however X11 don't like it at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Crash when resuming after suspension
 - Crash when trying to start on X11 with a Wayland compositor running
 - Crash with a virtual display connected on X11
-- GPU memory usage has been decreased by disabling allocation of depth and stencil buffers
+- GPU memory usage has been decreased by disabling allocation of depth
 
 ### Removed
 
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -118,7 +118,7 @@ fn create_gl_window(
         .with_srgb(srgb)
         .with_vsync(true)
         .with_depth_buffer(0)
-        .with_stencil_buffer(0)
+        .with_stencil_buffer(1)
         .with_hardware_acceleration(None)
         .build_windowed(window, event_loop)?;
 


chrisduerr:
> however X11 don't like it at all

I'm running X11 and it doesn't cause any problems.

Which compositor are you running? Also can you fix this by just setting them to `1`, or any value smaller than the defaults? Maybe there's a middle ground.

kchibisov:
@chrisduerr compton + awesome

kchibisov:
Hm, setting both to `1` also fixes. Let me try to tweak driver settings, I think it's a driver + GPU problem, since radeonsi doesn't zero something.

kchibisov:
Tweaking driver also works, however the content flickers on opening, however no artifacts during it.

chrisduerr:
Yeah I just tested again on i3+compton+amdgpu and it works just fine. But if setting both to `1` fixes it, that might be sufficient. I'm not sure if we should expect driver tweaks from users.

Could you try just setting the depth buffer to `1` with the stencil buffer at `0`? Or the other way around, though I'd expect this to be due to the depth buffer.

kchibisov:
stencil - 1 depth - 0 - fine, vise-verse - not fine

chrisduerr:
I guess I understand the issue even less than expected, but if stencil 1 and depth 0 works, that's probably what we should go for since it's the minimal memory allocation without any issues.

I'd like to hold back and see if @Jasu has any further comments though.

kchibisov:
Very strange X11 thing, though I've already seen similar glitches from this GPU under i3 earlier (like 6 month ago), however this one seems like our + GPU fault, since I can control these artifacts by changing stencil buffer.

Jasu:
I think that setting stencil buffer to 1 effectively reverts the whole thing (at least on my machine), so it's better to just revert the whole change until I or someone can figure it out.

When you set stencil buffer to 1, it tells GLX to find a matching visual that has at least 1 bit of stencil buffer support.

However, it seems that all visuals (at least on i915) either have 8 bit stencil + 24 bit depth or 0 bits of both:

```
glxinfo -t
...
94 GLX Visuals
Vis   Vis   Visual Trans  buff lev render DB ste  r   g   b   a      s  aux dep ste  accum buffer   MS   MS
 ID  Depth   Type  parent size el   type     reo sz  sz  sz  sz flt rgb buf th  ncl  r   g   b   a  num bufs caveats
--------------------------------------------------------------------------------------------------------------------
0x 20 24 TrueColor    0     32  0  rgba   1   0   8   8   8   8  .   .   0   24  8   0   0   0   0   0   0   None
0x 21 24 DirectColor  0     32  0  rgba   1   0   8   8   8   8  .   .   0   24  8   0   0   0   0   0   0   None
0x142 24 TrueColor    0     32  0  rgba   1   0   8   8   8   8  .   .   0    0  0   0   0   0   0   0   0   None
0x143 24 TrueColor    0     32  0  rgba   0   0   8   8   8   8  .   .   0    0  0   0   0   0   0   0   0   None
0x144 24 TrueColor    0     32  0  rgba   0   0   8   8   8   8  .   .   0   24  8   0   0   0   0   0   0   None
0x145 24 TrueColor    0     24  0  rgba   1   0   8   8   8   0  .   .   0    0  0   0   0   0   0   0   0   None
0x146 24 TrueColor    0     24  0  rgba   0   0   8   8   8   0  .   .   0    0  0   0   0   0   0   0   0   None
0x147 24 TrueColor    0     24  0  rgba   1   0   8   8   8   0  .   .   0   24  8   0   0   0   0   0   0   None
0x148 24 TrueColor    0     24  0  rgba   0   0   8   8   8   0  .   .   0   24  8   0   0   0   0   0   0   None
0x149 24 TrueColor    0     32  0  rgba   1   0   8   8   8   8  .   s   0    0  0   0   0   0   0   0   0   None
0x14a 24 TrueColor    0     32  0  rgba   0   0   8   8   8   8  .   s   0    0  0   0   0   0   0   0   0   None
0x14b 24 TrueColor    0     32  0  rgba   1   0   8   8   8   8  .   s   0   24  8   0   0   0   0   0   0   None
0x14c 24 TrueColor    0     32  0  rgba   0   0   8   8   8   8  .   s   0   24  8   0   0   0   0   0   0   None
... list goes on, with similar depth / stencil values.
```

Jasu:
@kchibisov Which driver setting helped with the issue? And what's your Compton version and configuration?

My desktop has RX580, so I might be able to reproduce this.

kchibisov:
@Jasu `Zero all vram allocations` and `Force uninitialized variables to default zero`. However it still flickers on opening with both of them, just **no** artifacts.

kchibisov:
compton v7.4  with `--experimental-backends`
```
backend              = "glx";
vsync                = true;
glx-no-stencil       = true;
glx-no-rebind-pixmap = true;
unredir-if-possible = true;

no-fading-openclose = true;
# Shadow
shadow = true;
shadow-radius = 10;
shadow-offset-x = -5;
shadow-offset-y = -0;
shadow-opacity = 0.7

wintypes:
{
  dock = { shadow = false; focus = false; redir-ignore = true; };
  dnd = { shadow = false; };
};

mark-wmwin-focused = true;
detect-client-opacity = true;
detect-client-leader = true;
detect-transient = true;
```

The artifacts *are* presented even without compton.

kchibisov:
@Jasu I can just revert only stencil change and keep `with_depth_buffer(0)`, since the last one doesn't do anything bad for me.

Jasu:
Thanks!

I noticed that Alacritty initialization does this:

```rust
        // Clear screen
        let background_color = config.colors.primary.background;
        renderer.with_api(&config, &size_info, |api| {
            api.clear(background_color);
        });

        // We should call `clear` when window is offscreen, so when `window.show()` happens it
        // would be with background color instead of uninitialized surface.
        window.swap_buffers();
```

There is no call to `glFinish` between the clear and buffer swap. `glutin` does not do that internally, and `with_api` wrapper does not do that either. So basically the first clear likely reaches the server after the buffer swap. This is only conjecture, since I don't have a chance to test anything until later.

kchibisov:
@jasu one more additional note, that these artifacts are presented during runtime too, they appear when I'm switching workspace in awesome.

Speaking about clear call, you're probably right that there's a timing issue. I'll try to take a look at this myself, however I don't think that this should change much.

kchibisov:
fwiw, this patch doesn't do anything, however I'm not super familiar with OpenGL, so leaving it up to you for now.
```
diff --git a/alacritty_terminal/src/renderer/mod.rs b/alacritty_terminal/src/renderer/mod.rs
index 51c0d0f..a2f1c94 100644
--- a/alacritty_terminal/src/renderer/mod.rs
+++ b/alacritty_terminal/src/renderer/mod.rs
@@ -927,6 +927,7 @@ impl<'a, C> RenderApi<'a, C> {
                 alpha,
             );
             gl::Clear(gl::COLOR_BUFFER_BIT);
+            gl::Finish();
         }
     }
```

Jasu:
If I would have been right, that should have fixed the initial glitch. So back to the drawing board, I guess.

Jasu:
I misunderstood `glXSwapBuffers`. It executs an implicit `glFinish`, so that cannot be the issue here.

chrisduerr:
> it seems that all visuals (at least on i915) either have 8 bit stencil + 24 bit depth or 0 bits of both

This was kinda my suspicion, since stencil and depth buffers are shared and 8 + 24 are the default, so a minimum size is not unlikely for various reasons (might just align to 32 either way or something like that).

@Jasu did you run into any concrete issues, or were you just concerned about keeping the memory usage minimal? If I recall your benchmarks correctly, the memory in use was significant enough to where it might impact other GPU heavy applications like video games.

@kchibisov You said that you get the issue when swapping workspaces in awesome and when launching. You've also mentioned you get the same issue with compton+i3 when launching. Do you also get this issue when swapping workspaces on i3?

kchibisov:
@chrisduerr i3 bug is different and its i3 fault, since they don't zero something (it also related to border in i3). I can't repro it anymore, I guess, since someone added a workaround in mesa, driver, whatever.

However our bug is presented in i3 and I still can control it with stencil buffer.

chrisduerr:
Gotcha. I think the following order of operation might be best then:
 - Revert original PR (preferably with `git revert` for uniform message)
 - Open issue to track the potential improvement and further research into this topic

As far as I'm concerned, reverting this isn't super high priority since the bug isn't that bad, but unless someone has a clear idea of what we could try to fix this, it's probably going to take a while and I'd rather not keep a PR open for that (tracking in an issue is usually easier).

Jasu:
@chrisduerr I think the amount of memory used is small enough as not to affect other applications *that* much. What I was actually hoping to achieve was that when GPU memory is swapped and the system performance is swap-bound, there is less pages to swap back for Alacritty.

@kchibisov I managed to replicate the issue on my desktop - not as dramatically as your video shows though, the garbage on the screen only persist for some milliseconds. I suspect that it's 
not as dramatic on my machine because I run i3 and it resizes the window right away, so the effect goes away fairly quickly.

Jasu:
@chrisduerr I replicated this, I'll let you know in a few hours if I can figure this out and a revert won't be needed.

chrisduerr:
> not as dramatically as your video shows though

The lower video framerate might also extend the visibility of the garbage frames by a bit, so that could be another factor.

> I replicated this, I'll let you know in a few hours if I can figure this out and a revert won't be needed.

Thanks for looking into this, but no rush. I'm fine with keeping things open as long as someone is actively working on it. Just don't want this sitting around for weeks, a couple days is no problem.

kchibisov:
@chrisduerr We don't need a full revert here, just a stencil buffer one. The bug is bad, since it's for all amdgpu users with Rx 500+ cards on X11, I guess. However it's fine to keep it around before 0.3.4 (it should be fixed before 0.3.4 other with partial revert or other update).

@Jasu I've also noticed that with i3 its less noticeable, however I also use a 4k screen, so it also matters here, I guess(larger grid).

chrisduerr:
> We don't need a full revert here, just a stencil buffer one

As @Jasu has pointed out, it makes no difference if we completely revert it or just revert the stencil buffer. I've said the bug isn't that bad since it's "just" some graphical glitches at startup, nothing that permanently affects the terminal session.

Jasu:
@kchibisov @chrisduerr I'm quite busy this week, so if I can't figure this out now, a revert will definitely be in place. 

I think that we should do a full revert, since only setting the stencil buffer doesn't seem to be usually possible (my Radeon doesn't support stencil without depth either). So a partial revert could trigger some corner cases if some GL implementation says "OK, whatever you say boss" and then does something completely unexpected.

kchibisov:
>  "just" some graphical glitches at startup

And runtime (workspace switch in i3/awesome) *

kchibisov:
> I'm quite busy this week, so if I can't figure this out now, a revert will definitely be in place.

@Jasu no worries here, we can always revert right before `v0.3.4` release. I'm not directly affected, since I'm using Wayland most of the time.

Jasu:
Sorry, I'll have to give up for today (and probably for a while.)

I'm guessing that this is because the framebuffer visual is different from the root visual or the original window's visual, so Mesa or X cannot reuse or initialize the previous buffer when resizing. I get the same behavior when setting e.g. `with_srgb(true)` instead of `with_depth_buffer` and `with_stencil_buffer`.

chrisduerr:
Created tracking issue https://github.com/jwilm/alacritty/issues/2881.

PR to revert https://github.com/jwilm/alacritty/pull/2882.

PR #2883, opened 2019-10-16
Bump winit version

kchibisov:
Fixes cursor icon managing with hide_when_typing: true on Wayland


kchibisov:
yes, cursor changing on Wayland was not even implemented before evlp2, this one just fixes a wrong new behavior.

PR #2888, opened 2019-10-17
Fix cases where config's "window" is not defined

### LINKED ISSUE #2887: gnome thumbnails in the app switcher are broken

Fedora 30 using Gnome

> If on linux, are you using X11 or Wayland?

X11

Since commit 401c2aab964b60e8c3b072e5098ba8e366d04944 thumbnails in the Gnome app switcher are broken.

It's easier to describe the problem with screenshots.

Here's what I get when using Alt-Tab, then moving over Alacritty and using the arrow to move down (or using the "Switch applications" shortcut)  with commit 3475e449870b382cda4ea6d48f980577cd8c929e:

alacritty_ok

and here's what I get with the same process as described above, since commit 401c2aab964b60e8c3b072e5098ba8e366d04944:

alacritty_bad

the thumbnails are missing for some reason.

Additional information:

* the configuration file doesn't change
* the desktop file doesn't change

Let me know if you have any question.

DSpeckhals:
In the config, if `window` is undefined, the derived `Default` for the String `title` is used, which is an empty String. This was unintended, and causes issues in gnome-shell (e.g. in the alt-tab dialog) when the window title is an empty string.

This commit adds a manually implemented default for the `WindowConfig`, it's the same as the derived `Default`, except for the `title`, which will now always be "Alacritty" as originally intended.

Fixes #2887

### DIFF
--- a/alacritty_terminal/src/config/window.rs
+++ b/alacritty_terminal/src/config/window.rs
@@ -9,7 +9,7 @@ use crate::index::{Column, Line};
 pub const DEFAULT_NAME: &str = "Alacritty";
 
 #[serde(default)]
-#[derive(Deserialize, Debug, Clone, Default, PartialEq, Eq)]
+#[derive(Deserialize, Debug, Clone, PartialEq, Eq)]
 pub struct WindowConfig {
     /// Initial dimensions
     #[serde(deserialize_with = "failure_default")]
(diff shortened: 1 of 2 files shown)


alacritty_terminal/src/config/window.rs
```diff
@@ -69,6 +69,24 @@ impl WindowConfig {
...
+impl Default for WindowConfig {
+    fn default() -> WindowConfig {
+        WindowConfig {
+            dimensions: Default::default(),
+            position: Default::default(),
+            padding: Default::default(),
+            decorations: Default::default(),
+            dynamic_padding: Default::default(),
+            startup_mode: Default::default(),
+            title: DEFAULT_NAME.into(),
```
chrisduerr: We should probably call the `default_title` function here for consistency.

DSpeckhals: Yeah, I was going back and forth on that one. I'll change it.

alacritty_terminal/src/config/window.rs
```diff
@@ -69,6 +69,24 @@ impl WindowConfig {
...
+impl Default for WindowConfig {
+    fn default() -> WindowConfig {
+        WindowConfig {
+            dimensions: Default::default(),
+            position: Default::default(),
+            padding: Default::default(),
+            decorations: Default::default(),
+            dynamic_padding: Default::default(),
+            startup_mode: Default::default(),
+            title: default_title(),
```
chrisduerr: Forgot to mention that in the last review, but would you mind moving this to the bottom? I think that would make it stand out a little more with all these `Default::default()`s. Top would work too I guess.

DSpeckhals: Done.

PR #2890, opened 2019-10-18
Default to ConPTY instead of WinPTY

chris-morgan:
I’ve been using the ConPTY backend since I started using Alacritty, and I don’t think I have *ever* experienced a problem unique to the ConPTY backend. Whenever I disable it to test something on the WinPTY backend, I cringe at how buggy ConPTY’s rendering is. After a year of ConPTY’s existence, I see no compelling reason to stick with the old and inferior. I reckon it’s time to default to the new and improved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - On Windows, query DirectWrite for recommended anti-aliasing settings
+- On Windows, the ConPTY backend will now be used by default if available; the `enable_experimental_conpty_backend` config option has been replaced with `use_winpty_backend`.
 
 ### Fixed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -337,16 +337,16 @@
 # directory of the parent process will be used.
 #working_directory: None
 
-# Windows 10 ConPTY backend (Windows only)
+# Windows 10 WinPTY backend (Windows only)
 #
-# This will enable better color support and may resolve other issues,
-# however this API and its implementation is still young and so is
-# disabled by default, as stability may not be as good as the winpty
-# backend.
+# Alacritty defaults to using the newer ConPTY backend if it is
+# available, since it provides better color support, resolves various
+# other minor rendering bugs and is quite a bit faster. If it is not
+# available, the the WinPTY backend will be used instead.
 #
-# Alacritty will fall back to the WinPTY automatically if the ConPTY
-# backend cannot be initialized.
-#enable_experimental_conpty_backend: false
+# Setting this option to true makes Alacritty use the WinPTY backend,
+# even if the ConPTY backend is available.
+#use_winpty_backend: false
 
 # Send ESC (\x1b) before characters when alt is pressed.
 #alt_send_esc: true
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -108,11 +108,10 @@ pub struct Config<T> {
     #[serde(default, deserialize_with = "failure_default")]
     pub cursor: Cursor,
 
-    /// Enable experimental conpty backend instead of using winpty.
-    /// Will only take effect on Windows 10 Oct 2018 and later.
+    /// Disable conpty backend, and use winpty even if conpty is available.
     #[cfg(windows)]
     #[serde(default, deserialize_with = "failure_default")]
-    pub enable_experimental_conpty_backend: bool,
+    pub use_winpty_backend: bool,
 
     /// Send escape sequences using the alt key.
     #[serde(default, deserialize_with = "failure_default")]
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - On Windows, query DirectWrite for recommended anti-aliasing settings
+- On Windows, the ConPTY backend will now be used by default if available; the `enable_experimental_conpty_backend` config option has been replaced with `use_winpty_backend`.
```
chrisduerr: I think it might be easier to split this off a bit, that way we don't need multiple sentences and run into huge lines.

We also don't use any punctuation (other than commas) in change log entries.

I'd recommend to have one entry in `Changed` which points out that `ConPTY` is now enabled by default and a separate entry that points out that `enable_experimental_conpty_backend` has been replaced with `use_winpty_backend`.

alacritty.yml
```diff
@@ -337,16 +337,16 @@
...
-# backend.
+# Alacritty defaults to using the newer ConPTY backend if it is
+# available, since it provides better color support, resolves various
+# other minor rendering bugs and is quite a bit faster. If it is not
+# available, the the WinPTY backend will be used instead.
 #
-# Alacritty will fall back to the WinPTY automatically if the ConPTY
-# backend cannot be initialized.
-#enable_experimental_conpty_backend: false
+# Setting this option to true makes Alacritty use the WinPTY backend,
```
chrisduerr:
```suggestion
# Setting this option to `true` makes Alacritty use the WinPTY backend,
```

alacritty.yml
```diff
@@ -337,16 +337,16 @@
 # directory of the parent process will be used.
 #working_directory: None
 
-# Windows 10 ConPTY backend (Windows only)
+# Windows 10 WinPTY backend (Windows only)
```
chrisduerr: WinPTY is not about `Windows 10` anymore, technically it's also available on previous Windows versions and enabling this option there might make sense for users (fallback works, but they'd eliminate that codepath).

alacritty.yml
```diff
@@ -337,16 +337,16 @@
...
+# available, since it provides better color support, resolves various
+# other minor rendering bugs and is quite a bit faster. If it is not
+# available, the the WinPTY backend will be used instead.
 #
-# Alacritty will fall back to the WinPTY automatically if the ConPTY
-# backend cannot be initialized.
-#enable_experimental_conpty_backend: false
+# Setting this option to true makes Alacritty use the WinPTY backend,
+# even if the ConPTY backend is available.
+#use_winpty_backend: false
```
chrisduerr: I don't think we should have `use_` in this option at all, it kinda doesn't help the user understand things any better and could be put on nearly every option. So `winpty_backend` is just shorter and quicker to understand.

However I do wonder if we want to mark it explicitly as a legacy support option. Like `legacy_winpty_backend` or something like that, but I don't think it should be relevant. However I think we might be able to sneak a `legacy` in the description just to make this sound super old and dated, to push as many people away from using this option as possible.

I think the following change would be a good choice:
```
# Setting this option to `true` makes Alacritty use the legacy WinPTY backend,
# even if the ConPTY backend is available.
```

alacritty_terminal/src/config/mod.rs
```diff
@@ -108,11 +108,10 @@ pub struct Config<T> {
     #[serde(default, deserialize_with = "failure_default")]
     pub cursor: Cursor,
 
-    /// Enable experimental conpty backend instead of using winpty.
-    /// Will only take effect on Windows 10 Oct 2018 and later.
+    /// Disable conpty backend, and use winpty even if conpty is available.
```
chrisduerr:
```suggestion
    /// Disable conpty backend and use winpty even if conpty is available.
```

davidhewitt:
Thanks for the PR, I've been wondering about proposing similar myself!

> I cringe at how buggy ConPTY’s rendering is

I trust you meant "how buggy *WinPTY's* rendering is" 😄 

The other thing I was tempted to propose: should we feature gate WinPTY support (as part of this PR)? In my experience the `winpty-sys` crate takes quite a while to build, so not having it as part of the dependency tree when building & developing would be nice.

chrisduerr:
> The other thing I was tempted to propose: should we feature gate WinPTY support (as part of this PR)? In my experience the winpty-sys crate takes quite a while to build, so not having it as part of the dependency tree when building & developing would be nice.

I would *like* to do the same thing. In fact if it was up to me, I'd just rip it all out. However I expect that a good chunk of users probably are still on Systems that do not support ConPTY.

This can include systems that are still running Windows 7, but we also have to consider that in corporate networks, it's not unlikely that managed systems are severely behind the latest Windows releases.

Windows 7 has extended support until January 14th, 2020 and Windows 8.1 until January 10th, 2023. I think the Windows 7 EOL might be a good time to switch things behind a feature flag and hopefully by 2023, we won't need it at all anymore (potentially even earlier I'd hope).

I think it's definitely a good idea to stage this incrementally, depending on the feedback of previous stages we can then determine how to continue. If nobody complains about this change, we can assume for example that nobody *chooses* to switch to WinPTY if ConPTY is supported.

davidhewitt:
ping @chris-morgan - I'd love to help get this merged, if you don't mind me finishing it off?

chrisduerr:
Closing in favor of https://github.com/jwilm/alacritty/pull/3113. Thanks for your work and getting this change started!

PR #2896, opened 2019-10-19
Track visible URLs continuously

### LINKED ISSUE #2479: hovering over very long lines slows down alacritty

Linux

Wayland

Reproduce:
`head -c81920 /dev/urandom | base64 | tr -d '\n'`

It also seems to happen with /dev/zero

### LINKED ISSUE #2898: Refactor mouse position handling
Currently there are some problems with handling of mouse position when it comes to padding and the message bar. This is mainly caused by a lack of clear distinction between moving inside the window, outside the window and inside the padding.

One of the issues that is present in the latest master is that when moving from the message bar's close button to the padding, the mouse cursor will not be reset.

However this also creates problems with https://github.com/jwilm/alacritty/pull/2896, since the column/line is always set even in the padding, URLs are highlighted when moving from URL directly to the padding.

Leaving the viewport currently also isn't tracked correctly. While the mouse is outside the window, the stored mouse positions still present it as within the window boundaries.

chrisduerr:
TODO:
 - [x] Resolve issue with leaving/entering Alacritty window straight into URL
 - [x] Update mouse cursor when URL highlight changes without mouse motion
 - [x] Fix URL color
 - [x] Add documentation to urlocator source + readme
 - [x] Alacritty source code documentation
 - [x] Fix remaining style issues

Fixes #2898.
Fixes #2479.


kchibisov:
new TODO:
- [ ] When pasting from multi-line selection lines are in inverted order
- [ ] When modifier is set to `Control` underline doesn't show up (`Hand` cursor is presented, Shift works fine).
- [ ] Url's underline color mismatches url color (always fg color).
- [ ] Underline is broken with multi/multiline Url [1]
- [ ] multi url parsing problem [2]

[1]
Guide to repro
open `bash`
enter this `git clone https://github.com/chrisduerr/alacrittyhttps://github.com`
press `Enter`
Highlight url

Screenshot_Mon 28 Oct 2019 07:38:18 AM MSK
[2]
I should be able to find 2 urls here `https://githubhttps://google.com` instead of one, right? At least on master it works like that.

alacritty_terminal/src/renderer/rects.rs
```diff
@@ -36,16 +36,46 @@ impl RenderRect {
...
+            self.start.line += 1;
+        }
+
+        rects.push(Self::create_rect(metrics, size, flag, self.start, self.end, self.color));
+
+        rects
+    }
+
+    fn create_rect(
+        metrics: &Metrics,
```
kchibisov: Why does RenderLine has a static method, called `create_rect`, which return `RenderRect`? I think it should be moved to RenderRect, right?

chrisduerr:
> Url's underline color mismatches url color (always config.fg color).

I've actually done this intentionally, though I'm not sure if it's the optimal solution. I've attempted to color the URLs intentionally by having a URL consist of a `Vec<Point -> Point>`, however that made things a bit more complicated (no more Copy) and did not work properly.

The issue is that while the scheme is matched, we do not count the cells as part of the URL yet and as such I can't push a new URL if the color changed.

My train of thought was that this isn't really necessary though, since URL highlights are different than underlines. So I just went for what was easier, I'm not sure what would be most desirable here?

> multi url string parsing problem [2]

This is intentional. The previous behavior was mostly a peculiarity of parsing URLs in reverse.

kchibisov:
> I've actually done this intentionally, though I'm not sure if it's the optimal solution. I've attempted to color the URLs intentionally by having a URL consist of a Vec<Point -> Point>, however that made things a bit more complicated (no more Copy) and did not work properly.

Oh, I see, I do like the current master behavior though, since for me this default highlighting seems a bit off. How does other terminals behave here?

chrisduerr:
> How does other terminals behave here?

Both VTE and URxvt color the underline the same as the cell fg. I can take another look if I see a good solution for resolving the scheme problem.

alacritty/src/input.rs
```diff
@@ -616,20 +618,17 @@ impl<'a, T: EventListener, A: ActionContext<T> + 'a> Processor<'a, T, A> {
...
-        let point = self.ctx.mouse_coords();
-
         // Skip normal mouse events if the message bar has been clicked
-        if let Some(message) = self.message_at_point(point) {
-            // Message should never be `Some` if point is `None`
-            debug_assert!(point.is_some());
-            self.on_message_bar_click(state, point.unwrap(), message, modifiers);
+        if self.message_close_at_cursor() && state == ElementState::Pressed {
+            self.ctx.clear_selection();
+            self.ctx.pop_message();
```
kchibisov: You should reset mouse cursor right after closing the button, otherwise you'll still have message modifier.

alacritty_terminal/src/term/mod.rs
```diff
@@ -810,15 +790,14 @@ impl SizeInfo {
...
-                && x >= self.padding_x as usize
-                && y < (self.height - self.padding_y) as usize
-                && y >= self.padding_y as usize
-        }
+    /// Check if coordinates are inside the terminal grid.
+    ///
+    /// The padding is not counted as part of the grid.
+    pub fn contains_point(&self, x: usize, y: usize) -> bool {
+        x < (self.width - self.padding_x) as usize
+            && x >= self.padding_x as usize
```
kchibisov: What the idea behind half opened ranges here? I think we should go with `(,)` instead of `[,)`, right?

alacritty/src/display.rs
```diff
@@ -118,6 +122,10 @@ pub struct Display {
     pub size_info: SizeInfo,
     pub font_size: Size,
     pub window: Window,
+    pub urls: Urls,
+
+    /// Currently highlighted URL.
+    pub url_highlighted: Option<Url>,
```
kchibisov: `url_highlighted` -> `highlighted_url`/`active_url`? It's either a comment a bit misleading to me or a name.

alacritty_terminal/src/renderer/rects.rs
```diff
@@ -36,16 +36,42 @@ impl RenderRect {
...
+    pub start: Point,
+    pub end: Point,
+    pub color: Rgb,
 }
 
 impl RenderLine {
-    fn into_rect(self, flag: Flags, metrics: &Metrics, size: &SizeInfo) -> RenderRect {
-        let start_x = self.start.col.0 as f32 * size.cell_width;
-        let end_x = (self.end.col.0 + 1) as f32 * size.cell_width;
+    pub fn as_rects(&self, flag: Flags, metrics: &Metrics, size: &SizeInfo) -> Vec<RenderRect> {
```
kchibisov: This is not `as_rects`, `as` prefix always assumes cheap conversion (e.g. `O(1)`). However this conversion has a linear complexity, so, please, change its name to avoid confusion.

alacritty_terminal/src/renderer/rects.rs
kchibisov: bump

alacritty_terminal/src/renderer/rects.rs
```diff
@@ -81,11 +107,11 @@ impl RenderLines {
         Self::default()
     }
 
-    pub fn into_rects(self, metrics: &Metrics, size: &SizeInfo) -> Vec<RenderRect> {
+    pub fn as_rects(&self, metrics: &Metrics, size: &SizeInfo) -> Vec<RenderRect> {
```
kchibisov: Same here. I don't like when `as_` is a linear thing.

kchibisov:
In addition, the current approach of handling modifiers doesn't play well, when pressing `modifier` with mouse being outside of the `alacrtitty` window, and then moving it into it and highlighting a url (you'll have the same `Hand` and no underline state).

alacritty_terminal/src/term/mod.rs
```diff
@@ -1959,24 +1818,22 @@ impl<T: EventListener> ansi::Handler for Term<T> {
...
+            Attr::Reverse => self.cursor.template.flags.insert(Flags::INVERSE),
+            Attr::CancelReverse => self.cursor.template.flags.remove(Flags::INVERSE),
+            Attr::Bold => self.cursor.template.flags.insert(Flags::BOLD),
+            Attr::CancelBold => self.cursor.template.flags.remove(Flags::BOLD),
+            Attr::Dim => self.cursor.template.flags.insert(Flags::DIM),
+            Attr::CancelBoldDim => self.cursor.template.flags.remove(Flags::BOLD | Flags::DIM),
+            Attr::Italic => self.cursor.template.flags.insert(Flags::ITALIC),
+            Attr::CancelItalic => self.cursor.template.flags.remove(Flags::ITALIC),
+            Attr::Underscore => self.cursor.template.flags.insert(Flags::UNDERLINE),
+            Attr::CancelUnderline => self.cursor.template.flags.remove(Flags::UNDERLINE),
```
kchibisov: Would you mind to change it from `Underscore` to `Underline`, since we have a little inconsistency right now, and you've already touched this lines?

alacritty/src/input.rs
```diff
@@ -226,39 +237,41 @@ impl From<MouseState> for CursorIcon {
...
-            if self.message_close_at_point(point, message) {
-                return MouseState::MessageBarButton;
-            } else {
-                return MouseState::MessageBar;
-            }
+        if point.is_some() && self.message_close_at_cursor() {
+            return MouseState::MessageBarButton;
+        } else if self.message_at_cursor() {
+            return MouseState::MessageBar;
         }
```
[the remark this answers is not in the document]
chrisduerr: > point.is_some() && self.message_at_cursor()

I haven't. The point doesn't need to be some in this case, which is why we're not doing any early return. I'll take another look at it though, I do think that this isn't perfect yet.

alacritty/src/input.rs
```diff
@@ -492,12 +493,8 @@ impl<'a, T: EventListener, A: ActionContext<T> + 'a> Processor<'a, T, A> {
             };
             self.mouse_report(code, ElementState::Released, modifiers);
             return;
-        } else if let (Some(point), true) = (point, button == MouseButton::Left) {
-            let mouse_state = self.mouse_state(point, modifiers);
-            self.update_mouse_cursor(mouse_state);
-            if let MouseState::Url(url) = mouse_state {
-                let url_bounds = url.linear_bounds(self.ctx.terminal());
-                self.ctx.terminal_mut().set_url_highlight(url_bounds);
+        } else if button == MouseButton::Left {
```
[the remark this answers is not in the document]
chrisduerr: I usually try not to use `if let` in this way to work around missing `&&` support, but if this is less than 100 columns it might be fine to change. It's just generally a bit hacky I think.

I've thought about relying on something like `matches!` for this, though I'm generally not a fan of pulling in depedencies for this kinda stuff. But it allows much nicer code imho:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0e057bc038b7a8f1d557a80003bbf190

alacritty/src/input.rs
chrisduerr: I think we did this before already but I couldn't figure out why. This makes sense, though I don't think I was able to make this happen when testing.

Are you speaking from experience or were you actually able to encounter this bug?

alacritty_terminal/src/term/mod.rs
chrisduerr: This is not code I've written myself, but it's correct the way it is from what I can tell.

Just to 100% confirm it, I made this little 'sketch' which should show it pretty clearly:
```
X X 2 3 4 5 6 7 X X

width: 10

padding: 2

x < 8 -> x < (10 - 2) -> x < (width - padding)

x >= 2 -> x >= padding
```

I think that's a pretty simple way to show what's going on here, should hopefully make it clear.

alacritty/src/display.rs
chrisduerr: This was previously a boolean field, where `url_highlighted` makes a bit more sense. I think `highlighted_url` is better now, thanks.

alacritty_terminal/src/renderer/rects.rs
chrisduerr: What would you propose? I just used `as` since it doesn't take `self`, but instead `&self`.

I don't think `into` is ever used without taking `self`. Though I generally agree with you.

alacritty_terminal/src/term/mod.rs
chrisduerr: I'm not sure why we've chosen `Underscore` here, I'll take a look at changing this, thanks.

It's possible that this naming comes from a spec or other docs though, but I don't think I've written this myself.

alacritty_terminal/src/renderer/rects.rs
chrisduerr: Oh, this actually exists for a pretty good reason.

Previously all this code was inside the `as_rects` method, but the code is now executed multiple times with different parameters. That's why it can't take `self` either, that would be misleading as it takes modified parameters that are stored on `self` in a different fashion.

So the only option would be moving it to `RenderRect`. However I have a hard time coming up with a name for that, which is usually an indication that it's not the perfect place for it either. The parameters are just all `RenderLine` exclusive and it acts purely as a way to deduplicate code.

alacritty/src/input.rs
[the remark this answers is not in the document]
chrisduerr: Turns out this check wasn't required at all anymore, stuff makes a lot more sense now!

alacritty_terminal/src/renderer/rects.rs
chrisduerr: I've decided to go for just `rects`, since it doesn't really convert anything. The return signature isn't `-> RenderRect`, so it's not a direct conversion between types anyways.

chrisduerr:
@kchibisov I've adjusted the glutin workaround to take an intersection of the modifiers reported by glutin and our own tracking. Since our goal is to determine when a modifier is pressed, this should have the highest accuracy. Instead of false negatives, we should now only get false positives, which should be a lot better for our usecase.

alacritty/src/input.rs
kchibisov: I'm speaking from experience here. Just open message bar, press `x` and don't move your mouse. It should stay as `Hand`.

alacritty_terminal/src/term/mod.rs
kchibisov: Ah, make sense.

alacritty_terminal/src/renderer/rects.rs
kchibisov: Maybe just `rects` or `get_rects`?

alacritty/src/input.rs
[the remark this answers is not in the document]
kchibisov: it shouldn't fit 100 columns, so its fine to ignore.

alacritty/src/input.rs
```diff
@@ -87,42 +90,59 @@ pub trait ActionContext<T: EventListener> {
 
 #[derive(Debug, Default, Copy, Clone)]
 pub struct Modifiers {
-    mods: ModifiersState,
+    glutin_state: ModifiersState,
```
kchibisov: I'm not a fan of `glutin_state`, maybe `glutin_mods_state`?

alacritty_terminal/src/renderer/rects.rs
kchibisov: I do understand the reason behind it, since we've already walk though this discussion in a past when we were refactoring/fixing underlines, and I do remember that there were similar options. So we can leave it as it is now, and promise to never make it public in its current state.

alacritty_terminal/src/renderer/rects.rs
```diff
@@ -100,10 +126,7 @@ impl RenderLines {
 
             // Check if there's an active line
             if let Some(line) = self.inner.get_mut(flag).and_then(|lines| lines.last_mut()) {
-                if cell.line == line.start.line
-                    && cell.fg == line.color
-                    && cell.column == line.end.col + 1
-                {
+                if cell.fg == line.color && cell.column == line.end.col + 1 {
```
kchibisov: Would you mind elaborating on removing `cell.line == line.start.line` check, just for me to faster understand things.

kchibisov:
There is a problem when I'm entering message bar area right away.
Guide to repro:
	- break your config, so you'll have a message bar on a next instance spawn.
	- Your mouse should be outside the future window.
	- Move mouse to a new window from a message_bar part.

So, your cursor will be `Text` and not `Default`. It'll be fixed on the "terminal area" enter and going back to message bar, however it should be changed to `Default` right away, without touching terminal thing.

alacritty/src/input.rs
chrisduerr: I think just `glutin_mods` should be superior to both here. It's shorter and still conveys the same message.

alacritty_terminal/src/renderer/rects.rs
chrisduerr: Oh yeah, if we would want to make this public it should be moved onto `RenderRect`.

alacritty_terminal/src/renderer/rects.rs
chrisduerr: I thought that it's not really great that a `RenderLine` which is a `Point -> Point` only supports points on the same line. So the change I've made to the `rects()` function on `RenderLine` is actually responsible itself for converting a `RenderLine` which spans multiple lines into multiple rects.

This nicely extracts the logic of handling multiple lines from this place and the URL handling and places it right at the place closest to conversion.

alacritty/src/input.rs
kchibisov: I'm fine with anything that emphasize `glutin` & `mods` in the same time. `state` was super generic and confusing.

kchibisov:
Just randomly discovered that our underlining is a bit broken here.
Screenshot_Sat 02 Nov 2019 04:28:14 AM MSK

I've typed `/set` to highlight stuff in vim.

kchibisov:
Sequence to repro

`echo -e "\"\e[4;9mh\e[0m\n\e[4;9mh\e[0m\n\""`

you should type it more then once.

alacritty/src/input.rs
chrisduerr: Completely agree, thanks.

nixpulvis:
@chrisduerr sorry I never really had the chance to take a look at this in depth before you merged it. I was hoping that this fix the issue of the mouse cursor and underline updating without the mouse moving, but it still seems to be a bit off:

url-issue

PR #2897, opened 2019-10-20
Fix XEmbed compilation error on 32-bit platforms

cyclopsian:
Compilation is broken on 32-bit X11 platforms since c1f089970fd3d4b9137d07647f8cd028b2b5b3a9:

```
error[E0308]: mismatched types
   --> alacritty/src/window.rs:420:69
    |
420 |         (xlib.XReparentWindow)(xlib_display as _, xlib_window as _, parent_id, 0, 0);
    |                                                                     ^^^^^^^^^ expected u32, found u64
help: you can convert an `u64` to `u32` and panic if the converted value wouldn't fit
    |
420 |         (xlib.XReparentWindow)(xlib_display as _, xlib_window as _, parent_id.try_into().unwrap(), 0, 0);
```

The XID of the XEmbed parent needs to be an unsigned long, as seen here: [xlib.rs#L623](https://github.com/erlepereira/x11-rs/blob/7d81595d630f03ced06747f3e214c59eda438df8/src/xlib.rs#L623).

### DIFF
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -15,6 +15,7 @@ use std::convert::From;
 #[cfg(not(any(target_os = "macos", windows)))]
 use std::ffi::c_void;
 use std::fmt;
+use std::os::raw::c_ulong;
 
 use glutin::dpi::{LogicalPosition, LogicalSize, PhysicalPosition, PhysicalSize};
 use glutin::event_loop::EventLoop;
@@ -392,7 +393,7 @@ impl Window {
 }
 
 #[cfg(not(any(target_os = "macos", windows)))]
-fn x_embed_window(window: &GlutinWindow, parent_id: u64) {
+fn x_embed_window(window: &GlutinWindow, parent_id: c_ulong) {
     let (xlib_display, xlib_window) = match (window.xlib_display(), window.xlib_window()) {
         (Some(display), Some(window)) => (display, window),
         _ => return,
--- a/alacritty_terminal/src/config/window.rs
+++ b/alacritty_terminal/src/config/window.rs
@@ -1,3 +1,5 @@
+use std::os::raw::c_ulong;
+
 use serde::Deserialize;
 
 use crate::config::{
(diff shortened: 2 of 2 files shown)


alacritty/src/window.rs
```diff
@@ -392,7 +393,7 @@ impl Window {
 }
 
 #[cfg(not(any(target_os = "macos", windows)))]
-fn x_embed_window(window: &GlutinWindow, parent_id: u64) {
+fn x_embed_window(window: &GlutinWindow, parent_id: c_ulong) {
```
chrisduerr: Wouldn't a `usize` work just as well without having to rely on non-rust datatypes?

cyclopsian: Unfortunately no, a usize the same size as a pointer type, and longs are not guaranteed to be the same size as pointers. It needs to match the XLib C ABI exactly.

nixpulvis: Make a PR ;)

cyclopsian: It should probably happen eventually, but it's a much bigger change. In the meantime this will fix any compilation issues.

chrisduerr: > Unfortunately no, a usize the same size as a pointer type, and longs are not guaranteed to be the same size as pointers.

c_ulong is defined as the same size as poniters in Rust.

cyclopsian: Yes, practically speaking you're correct. I don't know of any other LLVM platforms besides windows where the size of pointer and long are different. So that would likely work, but isn't strictly correct as far as linking to the C ABI goes. I preferred to be closer to that since the state of the x11 crate right now takes no pains to hide that it's a thin wrapper around the XLib C interface.

alacritty_terminal/src/config/window.rs
```diff
@@ -1,5 +1,7 @@
 use serde::Deserialize;
 
+use std::os::raw::c_ulong;
```
chrisduerr: Std imports should be at the top of all imports.

cyclopsian: Good call, fixing it

PR #2904, opened 2019-10-22
Avoid unused import warning on macOS/Windows

giginet:
Getting rid of an unused import for Windows/macOS.

My rustc version is 1.38.0.

Screen Shot 2019-10-22 at 17 31 23

### DIFF
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -15,7 +15,6 @@ use std::convert::From;
 #[cfg(not(any(target_os = "macos", windows)))]
 use std::ffi::c_void;
 use std::fmt;
-use std::os::raw::c_ulong;
 
 use glutin::dpi::{LogicalPosition, LogicalSize, PhysicalPosition, PhysicalSize};
 use glutin::event_loop::EventLoop;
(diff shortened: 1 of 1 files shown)


kchibisov:
Good catch, however this import is required on Linux/BSD, so please, just add a cfg guard to it (you can find the right one a few lines above).

h-michael:
The minimum supported version of Alacritty is Rust 1.36.0 now.
https://github.com/jwilm/alacritty/blob/8a263414572d8ed4d771461de2fdbd9f7e549a75/CONTRIBUTING.md#patches--pull-requests

alacritty/src/window.rs
```diff
@@ -15,6 +15,7 @@ use std::convert::From;
 #[cfg(not(any(target_os = "macos", windows)))]
 use std::ffi::c_void;
 use std::fmt;
+#[cfg(target_os = "linux")]
```
kchibisov: Please, use #[cfg(not(any(target_os = "macos", windows)))] instead, since Linux **and** BSD

giginet:
```suggestion
#[cfg(not(any(target_os = "macos", windows)))]
```

PR #2915, opened 2019-10-24
Draw dirty terminal even when event queue is empty

### LINKED ISSUE #2914: Visual bell animation stalls
The event loop stalls before completion of visual bell animation, even when the dirty bit is still set.

kennylevinsen:
If the event queue was empty, we would stop the event loop even if the terminal was dirty. This lead to the terminal not being drawn when running animations unless something else was causing events to queue, such as user input.

This change ensures that the terminal is always drawn on EventsCleared when dirty.

Fixes #2914.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -426,10 +426,10 @@ impl<N: Notify> Processor<N> {
             }
 
             if terminal.dirty {
-                // Clear dirty flag
-                let animating = !terminal.visual_bell.completed();
-                terminal.dirty = animating;
-                if animating {
+                terminal.dirty = false;
+
+                // Request immediate re-draw if visual bell animation is not finished yet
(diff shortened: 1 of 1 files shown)


kennylevinsen:
Yeah, I also assumed the move of that lock would be the thing that I got comments for. I usually prefer to not make too many assumptions without data, though.

Solutions:
1. Pump wakeup events. This was my initial approach, but I bumped my head into polymorphism issues as we only have a `Term<T>` in display.rs where I need it, not the `Term<EventListener>` needed to use the event_proxy. Didn't want to rewrite too much.
2. Separate dirty bit, likely on `Display`, which for now only tracks animation.

Comments? Might pick #2 for simplicity.

kennylevinsen:
Used UndeadLeech's suggestion.

PR #2917, opened 2019-10-25
Fix for WinPTY freeze on termination

cfiet:
This is a proof-of-concept for working around [winit:1241](https://github.com/rust-windowing/winit/issues/1241) to fix [alacritty:2889](https://github.com/jwilm/alacritty/issues/2889).

The workaround creates a notification channel between main even loop and `pty thread` that notifies the thread that the main loop is about to exit. This notifies the thread to terminate which stop it from  spamming `Wakeup` messages on Windows message channels.

### DIFF
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -46,7 +46,7 @@ pub struct EventLoop<T: tty::EventedPty, U: EventListener> {
     terminal: Arc<FairMutex<Term<U>>>,
     event_proxy: U,
     hold: bool,
-    ref_test: bool
+    ref_test: bool,
 }
 
 /// Helper type which tracks how much of a buffer has been written.
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -211,7 +211,7 @@ where
 
         loop {
             match self.pty.reader().read(&mut buf[..]) {
-                Ok(0) => return Ok(()),
+                Ok(0) => break,
                 Ok(got) => {
                     // Record bytes read; used to limit time spent in pty_read.
                     processed += got;
@@ -250,8 +250,10 @@ where
             }
         }
 
-        // Queue terminal redraw
-        self.event_proxy.send_event(Event::Wakeup);
+        if processed > 0 {
+            // Queue terminal redraw
+            self.event_proxy.send_event(Event::Wakeup);
+        }
 
         Ok(())
     }
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -211,7 +211,7 @@ where
 
         loop {
             match self.pty.reader().read(&mut buf[..]) {
-                Ok(0) => break,
+                Ok(0) => return Ok(()),
                 Ok(got) => {
                     // Record bytes read; used to limit time spent in pty_read.
                     processed += got;
@@ -286,10 +286,6 @@ where
         }
 
         Ok(())
-    }
-
-    pub fn terminate_signal(&self) -> PtyLoopTerminateSignal {
-        PtyLoopTerminateSignal { channel: self.channel() }
     }
 
     pub fn spawn(mut self) -> thread::JoinHandle<(Self, State)> {
@@ -399,18 +395,5 @@ where
 
             (self, state)
         })
-    }
-}
-
-pub struct PtyLoopTerminateSignal {
-    channel: mio_extras::channel::Sender<Msg>
-}
-
-impl PtyLoopTerminateSignal {
-    pub fn send(&self) {
-        // TODO: Channel might be closed here already
-        //       so this probably should only warn and
-        //       not panic
-        self.channel.send(Msg::Shutdown).unwrap()
     }
 }
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -19,7 +19,6 @@ use crate::sync::FairMutex;
 use crate::term::Term;
 use crate::tty;
 use crate::util::thread;
-use crate::terminate::TerminateSignal;
 
 /// Max bytes to read from the PTY
 const MAX_READ: usize = 0x10_000;
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -327,7 +327,6 @@ where
                             }
                         },
 
-                        #[cfg(unix)]
                         token if token == self.pty.child_event_token() => {
                             if let Some(tty::ChildEvent::Exited) = self.pty.next_child_event() {
                                 if !self.hold {
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -19,6 +19,7 @@ use crate::sync::FairMutex;
 use crate::term::Term;
 use crate::tty;
 use crate::util::thread;
+use crate::terminate::TerminateSignal;
 
 /// Max bytes to read from the PTY
 const MAX_READ: usize = 0x10_000;
@@ -45,7 +46,7 @@ pub struct EventLoop<T: tty::EventedPty, U: EventListener> {
     terminal: Arc<FairMutex<Term<U>>>,
     event_proxy: U,
     hold: bool,
-    ref_test: bool,
+    ref_test: bool
 }
 
 /// Helper type which tracks how much of a buffer has been written.
@@ -288,6 +289,10 @@ where
         Ok(())
     }
 
+    pub fn terminate_signal(&self) -> PtyLoopTerminateSignal {
+        PtyLoopTerminateSignal { channel: self.channel() }
+    }
+
     pub fn spawn(mut self) -> thread::JoinHandle<(Self, State)> {
         thread::spawn_named("pty reader", move || {
             let mut state = State::default();
@@ -395,5 +400,18 @@ where
 
             (self, state)
         })
+    }
+}
+
+pub struct PtyLoopTerminateSignal {
+    channel: mio_extras::channel::Sender<Msg>
+}
+
+impl PtyLoopTerminateSignal {
+    pub fn send(&self) {
+        // TODO: Channel might be closed here already
+        //       so this probably should only warn and
+        //       not panic
+        self.channel.send(Msg::Shutdown).unwrap()
     }
 }
--- a/alacritty_terminal/src/tty/windows/child.rs
+++ b/alacritty_terminal/src/tty/windows/child.rs
@@ -14,28 +14,31 @@ use std::io::Error;
 use std::os::raw::c_void;
 use std::sync::atomic::{AtomicPtr, Ordering};
 
+use log::warn;
+
 use mio_extras::channel::{channel, Receiver};
 
-use winapi::shared::ntdef::{HANDLE, PVOID, BOOLEAN};
+use winapi::shared::ntdef::{BOOLEAN, HANDLE, PVOID};
 use winapi::um::winbase::{RegisterWaitForSingleObject, UnregisterWait, INFINITE};
 use winapi::um::winnt::{WT_EXECUTEINWAITTHREAD, WT_EXECUTEONLYONCE};
 
 use crate::tty::ChildEvent;
 
-
 /// WinAPI callback for `HandleWaitSignal`, unpacks Rust closure reference and calls it
-extern "system" fn child_exit_callback<F>(ctx: PVOID, timed_out: BOOLEAN) 
-    where F: FnOnce() + Send
+extern "system" fn child_exit_callback<F>(ctx: PVOID, timed_out: BOOLEAN)
+where
+    F: FnOnce() + Send,
 {
-    if timed_out != 0 { return }
+    if timed_out != 0 {
+        return;
+    }
 
     let callback: Box<F> = unsafe { Box::from_raw(ctx as *mut F) };
     callback();
 }
 
-
 /// Represents closure attached to Win32 handle wait signal
-/// 
+///
 /// This allows to fire a Rust callback when subprocess exits
 pub(crate) struct HandleWaitSignal {
     wait_handle: AtomicPtr<c_void>,
@@ -48,8 +51,9 @@ impl HandleWaitSignal {
     /// blocking calls. See [`WT_EXECUTEINWAITTHREAD` flag docs] for details
     ///
     /// [`WT_EXECUTEINWAITTHREAD` flag docs]: https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-registerwaitforsingleobject
-    fn new<F>(child_handle: HANDLE, on_exit: F) -> Result<HandleWaitSignal, Error> 
-        where F: FnOnce() + Send
+    fn new<F>(child_handle: HANDLE, on_exit: F) -> Result<HandleWaitSignal, Error>
+    where
+        F: FnOnce() + Send,
     {
         let mut wait_handle: HANDLE = 0 as HANDLE;
         let wait_notify = Box::new(on_exit);
@@ -61,8 +65,8 @@ impl HandleWaitSignal {
                 Some(child_exit_callback::<F>),
                 Box::into_raw(wait_notify) as PVOID,
                 INFINITE,
-                WT_EXECUTEINWAITTHREAD | WT_EXECUTEONLYONCE
-            ) 
+                WT_EXECUTEINWAITTHREAD | WT_EXECUTEONLYONCE,
+            )
         };
 
         if 0 == success {
@@ -82,27 +86,21 @@ impl Drop for HandleWaitSignal {
     }
 }
 
-
-pub(crate) struct SubprocessState {
+pub(crate) struct ChildProcessState {
     _on_exit: HandleWaitSignal,
-    events: Receiver<ChildEvent>
+    events: Receiver<ChildEvent>,
 }
 
-impl SubprocessState {
-    pub fn new(subprocess_handle: HANDLE) -> Result<SubprocessState, Error> {
+impl ChildProcessState {
+    pub fn new(subprocess_handle: HANDLE) -> Result<ChildProcessState, Error> {
         let (sender, receiver) = channel();
         let on_exit = HandleWaitSignal::new(subprocess_handle, move || {
-            let result = sender.send(ChildEvent::Exited);
-            match result {
-                Ok(_) => print!("************* Yay!"),
-                _ => print!("************ Nay!")
+            if let Err(e) = sender.send(ChildEvent::Exited) {
+                warn!("An error occurred while attempting to notify about child process termination: {}", e);
             }
         })?;
 
-        Ok(SubprocessState { 
-            _on_exit: on_exit,
-            events: receiver
-        })
+        Ok(ChildProcessState { _on_exit: on_exit, events: receiver })
     }
 
     pub fn events(&self) -> &Receiver<ChildEvent> {
@@ -110,21 +108,17 @@ impl SubprocessState {
     }
 }
 
-
 mod test {
-    use std::io::{Error};
+    use std::io::Error;
     use std::ptr;
     use std::sync::{Arc, Condvar, Mutex};
     use std::time::Duration;
 
     use widestring::U16CString;
 
-    use winapi::shared::ntdef::{LPWSTR};
+    use winapi::shared::ntdef::LPWSTR;
     use winapi::um::processthreadsapi::{
-        CreateProcessW,
-        PROCESS_INFORMATION,
-        STARTUPINFOW,
-        TerminateProcess
+        CreateProcessW, TerminateProcess, PROCESS_INFORMATION, STARTUPINFOW,
     };
 
     use super::*;
@@ -145,7 +139,7 @@ mod test {
                 ptr::null_mut(),
                 ptr::null_mut(),
                 &mut si as *mut STARTUPINFOW,
-                &mut pi as *mut PROCESS_INFORMATION
+                &mut pi as *mut PROCESS_INFORMATION,
             ) {
                 Err(Error::last_os_error())
             } else {
@@ -154,7 +148,6 @@ mod test {
         }
     }
 
-
     // TODO: This is more of an integration test since it spawns new 'cmd.exe'
     //       process and heavily uses Win32 API. Should this be ignored
     //       by default? Worst-case, this will out after `WAIT_TIMEOUT`.
@@ -162,10 +155,7 @@ mod test {
     pub fn on_handle_wait_completed_signalls() {
         const WAIT_TIMEOUT: Duration = Duration::from_millis(200);
 
-        let signals = Arc::new((
-            Mutex::new(false),
-            Condvar::new()
-        ));
+        let signals = Arc::new((Mutex::new(false), Condvar::new()));
         let subprocess_handle = make_cmd_process().unwrap();
 
         // Setup callback to signal Condvar when process exits
@@ -175,7 +165,8 @@ mod test {
 
             *closed = true;
             cnd.notify_all();
-        }).unwrap();
+        })
+        .unwrap();
 
         // Kill the subprocess
         let kill_succeeded = unsafe { TerminateProcess(subprocess_handle, 0) };
--- a/alacritty_terminal/src/tty/windows/child.rs
+++ b/alacritty_terminal/src/tty/windows/child.rs
@@ -115,9 +115,9 @@ mod test {
             loop {
                 match child_exit_watcher.event_rx().try_recv() {
                     Ok(ChildEvent::Exited) => return, // Success
+                    Err(_) => unreachable!("No event {:?} was received", ChildEvent::Exited),
                 }
             }
         }
-        unreachable!("No event {:?} was received", ChildEvent::Exited);
     }
 }
--- a/alacritty_terminal/src/tty/windows/child.rs
+++ b/alacritty_terminal/src/tty/windows/child.rs
@@ -110,10 +110,20 @@ mod test {
         poll.poll(&mut events, Some(WAIT_TIMEOUT)).unwrap();
         for event in events.iter() {
             if event.token() == child_events_token {
-                // Verify the right event was emitted.
-                assert_eq!(child_exit_watcher.event_rx().try_recv().unwrap(), ChildEvent::Exited);
-                return;
+                // Verify that at least one `ChildEvent::Exited` was received
+                loop {
+                    let result = child_exit_watcher.event_rx().try_recv();
+                    match result {
+                        Ok(ChildEvent::Exited) => {
+                            return; // Success
+                        },
+                        Err(_) => {
+                            break;
+                        },
+                    }
+                }
             }
         }
+        panic!("No event {:?} was received", ChildEvent::Exited);
     }
 }
--- a/alacritty_terminal/src/tty/windows/child.rs
+++ b/alacritty_terminal/src/tty/windows/child.rs
@@ -28,8 +28,8 @@ extern "system" fn child_exit_callback(ctx: PVOID, timed_out: BOOLEAN) {
         return;
     }
 
-    let sender: Box<_> = unsafe { Box::from_raw(ctx as *mut Sender<ChildEvent>) };
-    let _ = sender.send(ChildEvent::Exited);
+    let event_tx: Box<_> = unsafe { Box::from_raw(ctx as *mut Sender<ChildEvent>) };
+    let _ = event_tx.send(ChildEvent::Exited);
 }
 
 pub struct ChildExitWatcher {
@@ -39,10 +39,10 @@ pub struct ChildExitWatcher {
 
 impl ChildExitWatcher {
     pub fn new(child_handle: HANDLE) -> Result<ChildExitWatcher, Error> {
-        let (sender, receiver) = channel::<ChildEvent>();
+        let (event_tx, event_rx) = channel::<ChildEvent>();
 
         let mut wait_handle: HANDLE = 0 as HANDLE;
-        let sender_ref = Box::new(sender);
+        let sender_ref = Box::new(event_tx);
 
         let success = unsafe {
             RegisterWaitForSingleObject(
@@ -58,7 +58,7 @@ impl ChildExitWatcher {
         if success == 0 {
             Err(Error::last_os_error())
         } else {
-            Ok(ChildExitWatcher { wait_handle: AtomicPtr::from(wait_handle), event_rx: receiver })
+            Ok(ChildExitWatcher { wait_handle: AtomicPtr::from(wait_handle), event_rx })
         }
     }
 
@@ -89,12 +89,10 @@ mod test {
     pub fn event_is_emitted_when_child_exits() {
         const WAIT_TIMEOUT: Duration = Duration::from_millis(200);
 
-        let (sender, receiver) = channel::<()>();
-        let mut command = Command::new("cmd.exe");
-        let mut child = command.spawn().unwrap();
+        let mut child = Command::new("cmd.exe").spawn().unwrap();
         let child_exit_watcher = ChildExitWatcher::new(child.as_raw_handle()).unwrap();
 
-        let mut poll = Poll::new().unwrap();
+        let poll = Poll::new().unwrap();
         let mut events = Events::with_capacity(8);
         let child_events_token = Token::from(0usize);
 
--- a/alacritty_terminal/src/tty/windows/child.rs
+++ b/alacritty_terminal/src/tty/windows/child.rs
@@ -86,13 +86,13 @@ impl Drop for HandleWaitSignal {
     }
 }
 
-pub struct ChildProcessState {
+pub struct ChildProcessWatcher {
     _on_exit: HandleWaitSignal,
-    events: Receiver<ChildEvent>,
+    events_rx: Receiver<ChildEvent>,
 }
 
-impl ChildProcessState {
-    pub fn new(subprocess_handle: HANDLE) -> Result<ChildProcessState, Error> {
+impl ChildProcessWatcher {
+    pub fn new(subprocess_handle: HANDLE) -> Result<ChildProcessWatcher, Error> {
         let (sender, receiver) = channel();
         let on_exit = HandleWaitSignal::new(subprocess_handle, move || {
             if let Err(e) = sender.send(ChildEvent::Exited) {
@@ -104,11 +104,11 @@ impl ChildProcessState {
             }
         })?;
 
-        Ok(ChildProcessState { _on_exit: on_exit, events: receiver })
+        Ok(ChildProcessWatcher { _on_exit: on_exit, events_rx: receiver })
     }
 
-    pub fn events(&self) -> &Receiver<ChildEvent> {
-        &self.events
+    pub fn events_rx(&self) -> &Receiver<ChildEvent> {
+        &self.events_rx
     }
 }
 
--- a/alacritty_terminal/src/tty/windows/child.rs
+++ b/alacritty_terminal/src/tty/windows/child.rs
@@ -10,8 +10,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::ffi::c_void;
 use std::io::Error;
-use std::os::raw::c_void;
 use std::sync::atomic::{AtomicPtr, Ordering};
 
 use mio_extras::channel::{channel, Receiver};
@@ -22,7 +22,7 @@ use winapi::um::winnt::{WT_EXECUTEINWAITTHREAD, WT_EXECUTEONLYONCE};
 
 use crate::tty::ChildEvent;
 
-/// WinAPI callback for `HandleWaitSignal`, unpacks Rust closure reference and calls it
+/// WinAPI callback for `HandleWaitSignal`, unpacks Rust closure reference and calls it.
 extern "system" fn child_exit_callback<F>(ctx: PVOID, timed_out: BOOLEAN)
 where
     F: FnOnce() + Send,
@@ -35,21 +35,21 @@ where
     callback();
 }
 
-/// Represents closure attached to Win32 handle wait signal.
+/// Represents closure attached to Win32 wait signal.
 ///
 /// This allows firing a Rust callback when the subprocess exits.
-pub(crate) struct HandleWaitSignal {
+pub(crate) struct WaitSignalHandler {
     wait_handle: AtomicPtr<c_void>,
 }
 
-impl HandleWaitSignal {
+impl WaitSignalHandler {
     /// Registers an asynchronous closure to call when process under `child_handle` exits.
     ///
     /// The `on_exit` is called on Win32 threadpool thread so it should avoid
     /// blocking calls. See [`WT_EXECUTEINWAITTHREAD` flag docs] for details.
     ///
     /// [`WT_EXECUTEINWAITTHREAD` flag docs]: https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-registerwaitforsingleobject
-    fn new<F>(child_handle: HANDLE, on_exit: F) -> Result<HandleWaitSignal, Error>
+    fn new<F>(child_handle: HANDLE, on_exit: F) -> Result<WaitSignalHandler, Error>
     where
         F: FnOnce() + Send,
     {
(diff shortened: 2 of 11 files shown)


chrisduerr:
I'm still not sure how Alacritty is doing anything bad here, the `Wakup` events we send are not increasing after termination, right? This is just our usual terminal updates.

If this is the case, the terminal should exit eventually, since no more data is printed to it. If the Wakeup is getting triggered for no good reason, we should probably try to fix this in the pty reader.

cfiet:
@chirsduerr: Another workaround, for `WinPTY` would be to somehow track if `winpty-agent` is still alive and quit if it has terminated. 

As I understand, the shell process is, when using `WinPTY` spawned and maintained by `winpty-agent` which is a 3rd party C library. The only thing that Alacritty knows of are 2 named pipes for shell `cin` and `cout`. Inside of it's `pty thread` alacrity polls only handles those, one of which is constantly polling `cin` to see if anything could be read from it.

As long as nothing bad happens to shell or agent, sending a `WM_QUIT` message to the windows (ex. by clicking X in the top right corner) works fine, no problems there.

However, named pipes persist even after `winpty-agent` and underlying shell have terminated. Alacritty does not really track that and therefore the `pty thread` still polls the pipes even when there's nothing to write to the. This wouldn't be an issue, but after every read attempt it sends a user-defined windows message to wake-up main Glutin/winit UI thread in case it is paused, I assume so it redraws in case new input was received from cout pipe. This triggers a bug (or at least a weird assumption in winit), where Windows message handling code exits only after all messages have been consumed from the message queue and only after that winit checks, if application has requested a termination. Because `pty thread` does not know yet about the termination (it only know about it after winit UI thread actually exits), it keeps sending messages to window message queue and prevents the main UI thread from exiting. I assume because there's nothing to read/write from.

Let me know if there's a Gitter, Slack or any other channel you're in, I'm happy to arrange some time to show you what triggers this bug step by step, if you're interested. It's definitely not an easy one to explain :)

chrisduerr:
> Let me know if there's a Gitter, Slack or any other channel you're in, I'm happy to arrange some time to show you what triggers this bug step by step, if you're interested.

There's an #alacritty channel on IRC, which has pretty good activity and I'm (almost) always around.

I think I understand the issue a bit better now though, I'm not super familiar with the WinPTY code and the way it works on Windows is extremely messy.

Would something like adding `Ok(0) if processed == 0 => return` work [here](https://github.com/jwilm/alacritty/blob/77127fbb41748eae264b1a7578a3bd7d0f94d5c5/alacritty_terminal/src/event_loop.rs#L213) to fix the issue?

An alternative solution might be just wrapping the `Wakeup` in a `if processed > 0` conditional. 

If these two do fix the issue, it depends a bit on performance which is the better fix, but let me know if my assumption is correct and this resolves it.

cfiet:
I think that would solve the freezing. I think that's event better fix as long as nothing related to rendering relays on those wake-up events happening (blinking cursor maybe?).

I've actually just realised there is a piece that tracks status of WinPTY agent that enters Glutin event handler [here](https://github.com/jwilm/alacritty/blob/master/alacritty/src/event.rs#L353) with [tty:process_should_exit()] returning `true`.

But that's only UI loop and it does not terminate it immediately. All it does is tell winint that it would like to exit next time it's possible. But there is still `pty thread` active and [sending messages to windows message queue here](https://github.com/jwilm/alacritty/blob/master/alacritty_terminal/src/event_loop.rs#L254) when all of this happens, so winit never gets around to check that intent.

So any way of stopping / throttling those messages will work.

cfiet:
This fixes the freeze, but brings back [the original issue from 0.3.3](https://github.com/jwilm/alacritty/issues/2889#issuecomment-545171037), where you have to interact with the window for it to disappear.

This is because the UI thread is put to sleep and does not wake up until a Windows event arrives. Probably it would be useful to re-think how tracking WinPTY/ConPTY is done on Windows (spearate thread?) so it can also wake up UI thread when shell exits / crashes.

chrisduerr:
> I think that would solve the freezing. I think that's event better fix as long as nothing related to rendering relays on those wake-up events happening (blinking cursor maybe?).

We do not have a blinking cursor. These Wakeups are only required when data has been processed and the terminal was mutated in the process. If `processed` is `0`, it is never required.

> This is because the UI thread is put to sleep and does not wake up until a Windows event arrives. Probably it would be useful to re-think how tracking WinPTY/ConPTY is done on Windows (spearate thread?) so it can also wake up UI thread when shell exits / crashes. Or just make it a part of PTY thread but I'm not sure how expensive is tty:process_should_exit()

On Linux a `Wakeup` is sent to the event handler when the process should exit, this should behave the same way on Windows.

cfiet:
> On Linux a Wakeup is sent to the event handler when the process should exit, this should behave the same way on Windows.

So it seems that it is just happening in completly different place than on Windows. On Unix-based systems this is done in PTY thread via some [MIO magic](https://github.com/jwilm/alacritty/blob/master/alacritty_terminal/src/event_loop.rs#L331) that I'm still need to fully grasp. Should be possible to move Windows child process monitoring there as well, I'll try to find some time to dive into `mio::windows` to see if we can use it to do the same thing.

cfiet:
@chrisduerr: That should be the solution. It now uses async WinAPI call to catch child process exit and produce relevant mio event. The event is captured in `pty thread` and terminates it. It is now aligned with how it's done on Unix-based systems until OS-specific code is hit. Have a look and let me know what changes should I make to get it merged. I still might post a couple of small changes here and there, want to double-check all the `unsafe` interactions with WinAPI

cfiet:
I'm done with the changes I wanted to make.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    let callback: Box<F> = unsafe { Box::from_raw(ctx as *mut F) };
+    callback();
+}
+
+/// Represents closure attached to Win32 handle wait signal
+///
+/// This allows to fire a Rust callback when subprocess exits
+pub(crate) struct HandleWaitSignal {
```
chrisduerr: Why is this `pub(crate)`?

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,180 @@
...
+    wait_handle: AtomicPtr<c_void>,
+}
+
+impl HandleWaitSignal {
+    /// Registers an asynchronous closure to call when `child_handle` wait signals.
+    ///
+    /// The `on_exit` is called on Win32 threadpool thread so it should avoid
+    /// blocking calls. See [`WT_EXECUTEINWAITTHREAD` flag docs] for details
```
chrisduerr:
```suggestion
    /// blocking calls. See [`WT_EXECUTEINWAITTHREAD` flag docs] for details.
```

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,180 @@
...
+
+/// Represents closure attached to Win32 handle wait signal
+///
+/// This allows to fire a Rust callback when subprocess exits
+pub(crate) struct HandleWaitSignal {
+    wait_handle: AtomicPtr<c_void>,
+}
+
+impl HandleWaitSignal {
+    /// Registers an asynchronous closure to call when `child_handle` wait signals.
```
chrisduerr: > when `child_handle` wait signals.

What do you mean by this?

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,180 @@
...
+impl Drop for HandleWaitSignal {
+    fn drop(&mut self) {
+        unsafe {
+            // Cancel WinAPI wait for child process exiting
+            UnregisterWait(self.wait_handle.load(Ordering::Relaxed));
+        }
+    }
+}
+
+pub(crate) struct ChildProcessState {
```
chrisduerr: Why is this `pub(crate)` is this used only in tests (if so that's perfect then)?

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,180 @@
...
+    _on_exit: HandleWaitSignal,
```
chrisduerr: Is my assumption correct that this is prefaced with `_` because it's only required to keep around until the destructor runs?

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,180 @@
...
+impl ChildProcessState {
+    pub fn new(subprocess_handle: HANDLE) -> Result<ChildProcessState, Error> {
+        let (sender, receiver) = channel();
+        let on_exit = HandleWaitSignal::new(subprocess_handle, move || {
+            if let Err(e) = sender.send(ChildEvent::Exited) {
+                warn!(
+                    "An error occurred while attempting to notify about child process z
+                    termination: {}",
+                    e
+                );
```
chrisduerr: When would this happen? I think the only case a `send` can fail is when the receiver is dead. Do we need to handle that?

What's the argument against using `let _ = sender.send(ChildEvent::Exited)`? Do we require this to be sent successfully? I'd assume if the other end isn't receiving anymore, it has already exited anyways.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,180 @@
...
+            }
+        })?;
+
+        Ok(ChildProcessState { _on_exit: on_exit, events: receiver })
```
chrisduerr: Good choice not to name the temporary variable `_on_exit` here, that should prevent us from accidentally not assigning it if we choose to change the constructor and forget to do something with this statement.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    pub fn events(&self) -> &Receiver<ChildEvent> {
```
chrisduerr: Why is this named `events`? Makes it sound like it gives back a `Vec<Event>` or something like that.

If there's no deeper meaning behind that, we might want to rename it to something that says more clearly that this is a receiving channel. I personally found `child_state.events()` a bit hard to understand.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,180 @@
...
+                Err(Error::last_os_error())
+            } else {
+                Ok(pi.hProcess)
+            }
+        }
+    }
+
+    // TODO: This is more of an integration test since it spawns new 'cmd.exe'
+    //       process and heavily uses Win32 API. Should this be ignored
+    //       by default? Worst-case, this will timeout after `WAIT_TIMEOUT`.
```
chrisduerr: I'd assume just the shell part of `cmd.exe` is used? Or does this actually pop open a new window?

And is the runtime of this test significant? Is the `WAIT_TIMEOUT` just a maximum and it usually runs in under 200ms? If so, why do we choose 200ms as timeout here?

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+}
+
+pub(crate) struct ChildProcessState {
+    _on_exit: HandleWaitSignal,
+    events: Receiver<ChildEvent>,
+}
+
+impl ChildProcessState {
```
chrisduerr: Why was the choice made to name this `ChildProcessState`? It does not really store the state of the process, right?

I'm not sure what a good name for this would be, but it seems more like a child exit listener to me?

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,180 @@
...
+        unsafe {
+            // Cancel WinAPI wait for child process exiting
+            UnregisterWait(self.wait_handle.load(Ordering::Relaxed));
+        }
+    }
+}
+
+pub(crate) struct ChildProcessState {
+    _on_exit: HandleWaitSignal,
+    events: Receiver<ChildEvent>,
```
chrisduerr: Is there ever anything sent through this other than an `Exit` signal? If not, maybe it makes sense to alter this channel's type and name to make it more clear what it is used for?

alacritty_terminal/src/event_loop.rs
```diff
@@ -250,8 +250,10 @@ where
             }
         }
 
-        // Queue terminal redraw
-        self.event_proxy.send_event(Event::Wakeup);
+        if processed > 0 {
+            // Queue terminal redraw
+            self.event_proxy.send_event(Event::Wakeup);
+        }
```
chrisduerr: Is this necessary even after the event processing changes? I think we've said this only affects ConPTY, right?

alacritty_terminal/src/tty/windows/child.rs
cfiet: I agree it is phrased in a confusing manner. The struct is just a thin wrapper between WinAPI and Rust closures. The wording is paraphrase of docs of [RegisterWaitForSingleObject] where it specifically says that callback is called when:

> The specified object is in the signaled state.

The function works with any of [synchronisation objects], although I only use it for synchronising on Process exiting.

I guess if you don't care about the that level of details, this could be changed to
```suggestion
    /// Registers an asynchronous closure to call when process under `child_handle` exits.
```

If you want this to be more precise, I guess the variable should just be called `handle` and comment could be more along the line of
```suggestion
    /// Registers an asynchronous closure to call when WinAPI Synchronization Object under `handle` sends a signal.
```

Let me know which one seems more fitting.

[RegisterWaitForSingleObject]: https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-registerwaitforsingleobject?redirectedfrom=MSDN 
[synchronisation objects]: https://docs.microsoft.com/en-us/windows/win32/sync/synchronization-objects

alacritty_terminal/src/tty/windows/child.rs
cfiet: Yes, when this is dropped the callback waiting for process exit will get unregistered so we need to tie it's lifetime to this struct.

alacritty_terminal/src/tty/windows/child.rs
cfiet: This test could run any windows process as long as it can be spawned and does not immediately exit. I use `cmd.exe` because it is available on the path easily on any Windows machine. If something goes wrong with the test run when process gets spawned but not killed, it will be running in the background until the test runner exits.

The timeout was just picked in an arbitrary fashion. I know that by default `cargo test` times tests out after a minute, but there is no point waiting for that long. The successful path through the test takes just a few milliseconds so I just scaled it by ~2 orders of magnitudes, which is still way faster failure than 1 minute.

alacritty_terminal/src/tty/windows/child.rs
cfiet: Yeah, it used to store the state in `Arc<AtomicBool>` but then I had to redo it slightly. See the next comment for details.

alacritty_terminal/src/tty/windows/child.rs
cfiet: All of this is based on Unix implementation that converts Unix signals from child process to `ChildEvent` enum which currently has only one case. If at one point someone needs to add another `ChildEvent` case this at least marks the place in Windows implementation which is expected to receive it.

alacritty_terminal/src/event_loop.rs
cfiet: This was mostly WinPTY problem where it seems that `mio` always signals that a Windows named pipe should be read after child process exits, even if there is no actual data in the pipe, which causes UI loop to be flooded with those `Wakeup` events.

This seems to be the best way to fix it since `processed` holds the number of bytes that were actually read from the pipe inside [this read loop](https://github.com/jwilm/alacritty/blob/89ee0127c747a151f2763693af89d06314d8d74f/alacritty_terminal/src/event_loop.rs#L212).

alacritty_terminal/src/tty/windows/child.rs
cfiet: It's used in `alacritty_terminal::tty::windows`, `alacritty_terminal::tty::windows::winpty`, `alacritty_terminal::tty::windows::conpty` but doesn't really need to be visible outside of the crate.

alacritty_terminal/src/tty/windows/child.rs
cfiet: This will probably only fail if the Alacritty process is already exiting and due to some unlucky context switching the receiving end gets cleaned up before callback is deregistered or the deregistering fails. It's probably not worth panicking the thread over (esp. that this is executed in WinAPI thread pool thread), but I thought it might be worth logging. Could be just ignored if you think that's a better solution.

alacritty_terminal/src/tty/windows/child.rs
cfiet: It's prepared to be used with [`EventedPty`](https://github.com/jwilm/alacritty/blob/master/alacritty_terminal/src/tty/mod.rs#L68) in case [`ChildEvent`](https://github.com/jwilm/alacritty/blob/master/alacritty_terminal/src/tty/mod.rs#L58) enum gets more cases.

alacritty_terminal/src/tty/windows/child.rs
chrisduerr: I like the first one. Describe what we do, not what we might be doing.

alacritty_terminal/src/tty/windows/child.rs
chrisduerr: If it is used outside of just tests, we'd usually just make it `pub`. That's the case for all other functions and I think it would be good to keep it that way for consistency.

alacritty_terminal/src/tty/windows/child.rs
chrisduerr: > but I thought it might be worth logging

Why? For issues potentially introduced in the future? As far as I can tell this should never be something we care about in the present implementation, which is why it's probably cleaner to just swallow it.

alacritty_terminal/src/tty/windows/child.rs
chrisduerr: It still doesn't return a list of events, but a receiver. That's a significant difference that should usually be exposed at the callee site. Something like `event_rx` would probably make this more readable.

alacritty_terminal/src/tty/windows/child.rs
chrisduerr: Makes sense, thanks for the details. I think if this is just spawning a subprocess that is completely silent (just the way Alacritty silently spawns 'cmd.exe' without anything else opening), it should be fine.

It's true that it is more like an integration test, but I don't think there's much wrong with having a quick running integration test at this place. The problem with integration is usually just runtime and if what you're saying is correct, this doesn't really impact runtime of the tests.

alacritty_terminal/src/tty/windows/child.rs
chrisduerr: The Unix implementation only defines the `EventedPty` interface though, right? Which is not implemented for `ChildProcessState`, but for `Pty`.

So the `ChildProcessState` is just a struct like any other that the `Pty` uses to gather the necessary informations, it's not necessarily bound by `EventedPty` and is mostly independent.

You're talking about other `ChildEvent`s, though I'm not sure what other child events we would be caring about. However since the `ChildEvent` enum itself is already structured this way with only a single field and a lot of generically named methods just for handling this single case, I don't think it's a bad idea to follow it in this PR. Though I feel like that might not have been the most efficient choice originally.

So I think it's fine to keep things mostly the way they are right now, just to make sure that our different platforms share some similarities.

alacritty_terminal/src/tty/windows/child.rs
cfiet: Sure, sounds reasonable

alacritty_terminal/src/tty/windows/child.rs
cfiet: So with my limited understanding of the codebase, this could be a lot simpler for both Unix and Windows systems with `Arc<AtomicBool>` that loop probes but you still need to send an event to `mio` to wake up the IO loop so it actually checks the value of that boolean which works fine for Unix because it has out-of-the-box support for Unix signals so IO loop would know that some signal was received and would just check it. However, there is no similar support for Windows, so best bet is to push it through a channel which would wake up `mio` loop.

If that's the case, I guess you might just push `ChildEvent` from callback directly to the IO loop instead of jumping through hoops to wake up the loop to check boolean flag.

Not sure if that's the best way of doing this for both Unix and Windows but I'm not familiar enough with the codebase yet to judge that. Happy to take any suggestions.

cfiet:
I'll try to find some time later today to wrap it up and fix the remaining things mentioned in the review.

alacritty_terminal/src/tty/windows/child.rs
cfiet: This is used in tests

cfiet:
This should be all the changes you've requested, let me know if there's anything else.

alacritty_terminal/src/tty/windows/child.rs
nixpulvis: I'm perfectly fine mixing in integration tests here, especially since it's a `windows` module. However we could also start collecting integration style tests in a `tests` directory as recommended by the [Rust Lang Book](https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests), though again, I see no need for this here.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,180 @@
...
+                Ok(pi.hProcess)
+            }
+        }
+    }
+
+    // TODO: This is more of an integration test since it spawns new 'cmd.exe'
+    //       process and heavily uses Win32 API. Should this be ignored
+    //       by default? Worst-case, this will timeout after `WAIT_TIMEOUT`.
+    #[test]
+    pub fn on_handle_wait_completed_signalls() {
```
nixpulvis:
```suggestion
    pub fn on_handle_wait_completed_signals() {
```

alacritty_terminal/src/tty/windows/child.rs
chrisduerr: > However we could also start collecting integration style tests in a tests directory as recommended by the Rust Lang Book, though again, I see no need for this here.

We already do that. However I'm not sure if that's necessary here.

alacritty_terminal/src/tty/windows/child.rs
nixpulvis: It's unclear which comment you're referring to, but I agree, this naming is a bit misleading.

alacritty_terminal/src/tty/windows/child.rs
cfiet: Renamed it to `on_handle_wait_runs_callback_when_process_exits` in 74b7b9a

alacritty_terminal/src/tty/windows/child.rs
cfiet: This got renamed to `ChildProcessWatcher` in e678899

alacritty_terminal/src/tty/windows/child.rs
cfiet: Removed in d1a2dfc

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,173 @@
...
+            }
+        }
+    }
+
+    #[test]
+    pub fn on_handle_wait_runs_callback_when_process_exits() {
+        // The test should only take few milliseconds to succeed but will
+        // only fail due to timeout. It can be safely assumed that if callback
+        // has not been called after `WAIT_TIMEOUT` duration, there is some
+        // issue with the code under test.
```
chrisduerr: I'm still a bit skeptical about this comment. Saying it 'should only take a few milliseconds' seems mostly like a guess, which usually isn't super great. We should instead determine why this could potentially timeout and document this specific case, if we can't determine reliably how long a test will take, we'll have test failures left and right which is terrible for our trust in the test suite.

This sounds a lot worse than it is though, since we already know why we need the timeout and what the potential problems are. So we just need to clarify the comment and talk about exactly what specific things can happen and why this timeout resolves them, rather than having a vague statement about what could be going on.

Additionally this comment kinda makes it sound like the timeout would only happen when the subprocess fails and is unrelated to us. However the timeout is exactly what we use to determine if **our** functionality is correct. So a timeout would also happen if we break our exit handling.

So I'm thinking something more like this:

> We use a timeout to determine a maximum timeframe in which the exit event has to be emitted, since failure will not emit anything.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,173 @@
...
+        const WAIT_TIMEOUT: Duration = Duration::from_millis(200);
+
+        let (sender, receiver) = channel::<()>();
+        let subprocess_handle = make_cmd_process().unwrap();
+
+        // Setup callback to signal Condvar when process exits
+        let _hnd_wait = HandleWaitSignal::new(subprocess_handle, move || {
```
chrisduerr: `_hnd_wait` isn't really a great variable name. I'd assume this is likely to change with the rename of `HandleWaitSignal` anyways though, so I don't want to make any concrete recommendations yet.

alacritty_terminal/src/tty/windows/child.rs
cfiet: Should I just remove it completely?

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,173 @@
...
+                Default::default(),
+                ptr::null_mut(),
+                ptr::null_mut(),
+                &mut si as *mut STARTUPINFOW,
+                &mut pi as *mut PROCESS_INFORMATION,
+            ) {
+                Err(Error::last_os_error())
+            } else {
+                Ok(pi.hProcess)
+            }
```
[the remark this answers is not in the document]
cfiet: I actually removed all of it and replaced it with `std::process::Command`.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,173 @@
...
+            Ok(HandleWaitSignal { wait_handle: AtomicPtr::from(wait_handle) })
+        }
+    }
+}
+
+impl Drop for HandleWaitSignal {
+    fn drop(&mut self) {
+        unsafe {
+            // Cancel WinAPI wait for child process exiting
+            UnregisterWait(self.wait_handle.load(Ordering::Relaxed));
```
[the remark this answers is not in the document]
cfiet: It has to be `Send` since it's being a part of `Term` struct moved between threads by the main event loop.

Otherwise, you get:
```
error[E0277]: `*mut std::ffi::c_void` cannot be sent between threads safely
   --> alacritty\src\main.rs:181:22
    |
181 |     let event_loop = EventLoop::new(Arc::clone(&terminal), event_proxy.clone(), pty, &config);
    |                      ^^^^^^^^^^^^^^ `*mut std::ffi::c_void` cannot be sent between threads safely
    |
    = help: within `alacritty_terminal::tty::windows::Pty<'_>`, the trait `std::marker::Send` is not implemented for `*mut std::ffi::c_void`   
    = note: required because it appears within the type `alacritty_terminal::tty::windows::child::HandleWaitSignal`
    = note: required because it appears within the type `alacritty_terminal::tty::windows::child::ChildProcessWatcher`
    = note: required because it appears within the type `alacritty_terminal::tty::windows::Pty<'_>`
    = note: required by `alacritty_terminal::event_loop::EventLoop::<T, U>::new`
```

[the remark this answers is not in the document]
cfiet: It has to be `Send` since it becomes part of `Term` that is shared between threads by the main event loop.

Otherwise you get:
```
error[E0277]: `*mut std::ffi::c_void` cannot be sent between threads safely
   --> alacritty\src\main.rs:181:22
    |
181 |     let event_loop = EventLoop::new(Arc::clone(&terminal), event_proxy.clone(), pty, &config);
    |                      ^^^^^^^^^^^^^^ `*mut std::ffi::c_void` cannot be sent between threads safely
    |
    = help: within `alacritty_terminal::tty::windows::Pty<'_>`, the trait `std::marker::Send` is not implemented for `*mut std::ffi::c_void`   
    = note: required because it appears within the type `alacritty_terminal::tty::windows::child::HandleWaitSignal`
    = note: required because it appears within the type `alacritty_terminal::tty::windows::child::ChildProcessWatcher`
    = note: required because it appears within the type `alacritty_terminal::tty::windows::Pty<'_>`
    = note: required by `alacritty_terminal::event_loop::EventLoop::<T, U>::new`
```

alacritty_terminal/src/tty/windows/child.rs
chrisduerr: I think it's useful to point out that the timeout is used to determine success of the test, kinda like the example I've provided.

cfiet: I replaced it with [a comment on the actual line](https://github.com/cfiet/alacritty/blob/fix/winpty-freeze-on-termination/alacritty_terminal/src/tty/windows/child.rs#L132) that does the waiting, I feel it's a bit more clear there.

chrisduerr: Yeah, was considering to suggest that myself. Great thinking.

cfiet:
One other suggestion I have, since you still have some second though about names of `ChildProcessWatcher` and `WaitSignalHandler`, they could probably just be merged it all into a single struct, something like `ChildExitWatcher`. It is unlikely that there will be any other use case for `WaitSignalHandler` unrelated to this functionality, so maybe there is no point in having it separated and done in such a generic fashion.

It was done this way because it was a lot easier for me when I was still trying to decide between mio channel and atomic boolean for storing the fact that child has exited, but it doesn't have to stay that way.

chrisduerr:
> One other suggestion I have, since you still have some second though about names of ChildProcessWatcher and WaitSignalHandler, they could probably just be merged it all into a single struct, something like ChildExitWatcher. It is unlikely that there will be any other use case for WaitSignalHandler unrelated to this functionality, so maybe there is no point in having it separated and done in such a generic fashion.

I'd definitely be curious on how this would work out and if this would simplify the implementation. If we'd be able to streamline things a little bit that would of course be very nice, though I'm not sure if I want to request this change and then tell you to revert it again.

If you're willing to test it out, feel free to do so, otherwise I can give it a shot too. Otherwise just rename the one struct and that should be sufficient.

cfiet:
Just merged the structs, I think the code is a lot more straightforward this way. Testing is a bit more complicated, since it requires minimal `mio` setup, but overall I think it's better this way.

One final thing I'm considering now - since `child` module now has just a single struct, might be worth just moving it to parent module.

Let me know what do you think.

alacritty_terminal/src/tty/windows/child.rs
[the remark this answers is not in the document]
chrisduerr: Makes sense, thanks for the clarification.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+pub struct ChildExitWatcher {
+    wait_handle: AtomicPtr<c_void>,
+    event_rx: Receiver<ChildEvent>,
+}
+
+impl ChildExitWatcher {
+    pub fn new(child_handle: HANDLE) -> Result<ChildExitWatcher, Error> {
+        let (sender, receiver) = channel::<ChildEvent>();
```
chrisduerr: Renaming this will allow us to simplify the assignment when creating the `ChildExitWatcher`:

```suggestion
        let (event_tx, event_rx) = channel::<ChildEvent>();
```

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    use super::*;
+
+    #[test]
+    pub fn event_is_emitted_when_child_exits() {
+        const WAIT_TIMEOUT: Duration = Duration::from_millis(200);
+
+        let (sender, receiver) = channel::<()>();
+        let mut command = Command::new("cmd.exe");
+        let mut child = command.spawn().unwrap();
```
chrisduerr: Can't we just do the following?

```rust
let mut child = Command::new("cmd.exe").spawn().unwrap();
```

Or was there a reason for the assignment to a temporary variable? I thought it might be related to borrow checking, but I think this should just work.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    use mio::{Events, Poll, PollOpt, Ready, Token};
+
+    use super::*;
+
+    #[test]
+    pub fn event_is_emitted_when_child_exits() {
+        const WAIT_TIMEOUT: Duration = Duration::from_millis(200);
+
+        let (sender, receiver) = channel::<()>();
```
chrisduerr: Is this ever used by anything?

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,121 @@
...
+        // Poll for the event or fail with timeout if nothing has been sent
+        poll.poll(&mut events, Some(WAIT_TIMEOUT)).unwrap();
+        for event in events.iter() {
+            if event.token() == child_events_token {
+                // Verify the right event was emitted.
+                assert_eq!(child_exit_watcher.event_rx().try_recv().unwrap(), ChildEvent::Exited);
+                return;
+            }
+        }
+    }
```
chrisduerr: This doesn't correctly assert success. If there is no token that matches `child_events_token`, it will never assert anything. What other events could this receive, do we even need this?

In any case, we'd probably want to do something more akin to iterator logic here, like `events.iter().filter().next()` or something, that would make the code cleaner. Then we can just unwrap on the `next` and we'll handle the case that there's nothing too.

alacritty_terminal/src/tty/windows/child.rs
cfiet: Fixed

alacritty_terminal/src/tty/windows/child.rs
cfiet: Fixed

alacritty_terminal/src/tty/windows/child.rs
cfiet: Fixed

alacritty_terminal/src/tty/windows/child.rs
cfiet: Not entirely sure it is that simple, events just tells us that `event_rx` is ready to be read. Have a look how I reworked it and let me know if the solution acceptable.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        child.kill().unwrap();
+
+        // Poll for the event or fail with timeout if nothing has been sent
+        poll.poll(&mut events, Some(WAIT_TIMEOUT)).unwrap();
+        for event in events.iter() {
+            if event.token() == child_events_token {
+                // Verify that at least one `ChildEvent::Exited` was received
+                loop {
+                    let result = child_exit_watcher.event_rx().try_recv();
```
chrisduerr: This is going to produce an extremely hot loop just going at it over and over again, which is not ideal.

We should just `event_rx().recv()` on this.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    #[test]
+    pub fn event_is_emitted_when_child_exits() {
+        const WAIT_TIMEOUT: Duration = Duration::from_millis(200);
+
+        let mut child = Command::new("cmd.exe").spawn().unwrap();
+        let child_exit_watcher = ChildExitWatcher::new(child.as_raw_handle()).unwrap();
+
+        let poll = Poll::new().unwrap();
+        let mut events = Events::with_capacity(8);
```
chrisduerr: I'd put this definition right above `poll.poll`, making it easier to know that the parameter is a freshly initialized events buffer.

Also can you explain to me why this requires a capacity of `8` instead of `1`?

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,128 @@
...
+            PollOpt::oneshot(),
+        )
+        .unwrap();
+
+        child.kill().unwrap();
+
+        // Poll for the event or fail with timeout if nothing has been sent
+        poll.poll(&mut events, Some(WAIT_TIMEOUT)).unwrap();
+        for event in events.iter() {
+            if event.token() == child_events_token {
```
chrisduerr: Making sure that this is the case is good, but is this ever not going to be the case? This is the only token we've registered.

If we want to check that things are in a reasonable state, we should just `assert!` them.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -0,0 +1,128 @@
...
+            Ready::readable(),
+            PollOpt::oneshot(),
+        )
+        .unwrap();
+
+        child.kill().unwrap();
+
+        // Poll for the event or fail with timeout if nothing has been sent
+        poll.poll(&mut events, Some(WAIT_TIMEOUT)).unwrap();
+        for event in events.iter() {
```
chrisduerr: Which other events do we expect here besides `ChildEvent::Exited`? Can't we just check if the first event we receive is `Exited` and panic otherwise?

alacritty_terminal/src/tty/windows/child.rs
chrisduerr: > events just tells us that event_rx is ready to be read

The `poll.poll` function will block until at least one readiness event is received. So `event_rx` should always be ready to be read.

alacritty_terminal/src/tty/windows/child.rs
cfiet: Don't think that would be a hot loop. What it will do is take all the events queued on the received and discard all of them until it either gets `Ok(ChildEvent::Exited)` ending in success or `Err(TryRecvError::Empty)` breaking the loop and failing.

alacritty_terminal/src/tty/windows/child.rs
cfiet: To reflect that this test doesn't care if there are other events emitted as long as one of them is eventually `ChildEvent::Exited`. This is unnecessary now, since there is no other values for `ChildEvent` enum, but ultimately, the functionality is still working, if there were multiple events emitted during the test.

alacritty_terminal/src/tty/windows/child.rs
cfiet: I feel it's too strict of an assumption if at any point in the future there would be other values for `ChildEvent`.

alacritty_terminal/src/tty/windows/child.rs
cfiet: Fixed

alacritty_terminal/src/tty/windows/child.rs
cfiet: Fixed

alacritty_terminal/src/tty/windows/child.rs
chrisduerr: Yeah, but at that point why are we using `try_recv` instead of iterating?

alacritty_terminal/src/tty/windows/child.rs
chrisduerr: Then we'll adjust the tests.

alacritty_terminal/src/tty/windows/child.rs
cfiet: Not sure what should I iterate on? Only way too verify that the right event type was send from the watcher thread is to receive them on the receiver side.

chrisduerr: With the iterator I was talking about the `iter` method which is provided on a `Receiver`, which allows iterating until `try_recv` would block.

However I think that just taking next and asserting that is the better choice and much simpler to understand, assuming we don't send any other events before that yet.

alacritty_terminal/src/tty/windows/child.rs
chrisduerr: Assuming that it works, I feel like the code should look more like this:

```rust
poll.poll(&mut events, Some(WAIT_TIMEOUT)).unwrap();

assert_eq!(events.next().token(), child_events_token);
assert_eq!(child_exit_watcher.event_rx().try_recv(), Ok(ChildEvent::Exited));
```

If there's any issues with this, please let me know. But if my assumptions about our current state of the code are correct, this should work.

cfiet: Fixed

PR #2926, opened 2019-10-28
Fix message bar being drawn over terminal content after the next resize

### LINKED ISSUE #2924: Message bar is overlapping terminal content
Introduced in #2438. 

so instead of
```
-----------------------
|                     |
|       terminal      |
|                     |
----------------------- <- terminal end
----------------------- <- message bar start
|     message bar     |
----------------------- <- message bar end
```
we now have
```
-----------------------
|                     |
|       terminal      |
|                     |
----------------------- <- message bar start
|     message bar     |
----------------------- <- message bar and terminal end
```

kchibisov:
Fixes #2924

### DIFF
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -332,9 +332,8 @@ impl Display {
         let mut pty_size = self.size_info;
 
         // Subtract message bar lines from pty size
-        if resize_pending.message_buffer.is_some() {
-            let lines =
-                message_buffer.message().map(|m| m.text(&self.size_info).len()).unwrap_or(0);
+        if let Some(message) = message_buffer.message() {
+            let lines = message.text(&self.size_info).len();
             pty_size.height -= pty_size.cell_height * lines as f32;
         }
 


alacritty/src/event.rs
```diff
@@ -488,8 +488,15 @@ impl<N: Notify> Processor<N> {
                     CloseRequested => processor.ctx.terminal.exit(),
                     Resized(lsize) => {
                         let psize = lsize.to_physical(processor.ctx.size_info.dpr);
-                        processor.ctx.resize_pending.dimensions = Some(psize);
-                        processor.ctx.terminal.dirty = true;
+                        // Drop resize events, which resize to the current size
+                        if (psize.height as f32 - processor.ctx.size_info.height).abs()
```
kchibisov: This change is an optimization for resizes, that are identical to a current window size. If you don't like it, I'm fine with ignoring it.

chrisduerr: When is this ever hit? Why would a resize event be sent in that case?

kchibisov: Change focus on sway. So if you have two windows next to each other and you change a focus, you'll get a resize.

chrisduerr: That sounds like a winit bug to me?

kchibisov: Oh, it does actually. Doesn't happen on `X11`. Should I revert this `resize` check?

chrisduerr: Yeah and report upstream probably. We shouldn't be checking if resize events make sense, that's something winit should be doing.

kchibisov: https://github.com/rust-windowing/winit/issues/1247

Issue #2925, opened 2019-10-28
reset selection when removing chars from before selection?


matthiaskrgr:
alacritty @ 9ff2838844df45b6a1f20621ced870e9e1855ec8

I can type something like `Hello, this is a test.`, then I select the word "this", move the cursor to the start of "this" and remove the substring "Hello, " which will cause the selection to become " a t" as the text shifts to the left but the selection stays in place. This looks odd imo.
It even looks a bit stranger when I double-click-select and the selection will automatically select the 1-2 words that are under the position that I clicked at initially, as the text scrolls by during deletion.
The actual clipboard content still represents the initially clicked test at least.

Which operating system does the issue occur on?
linux

If on linux, are you using X11 or Wayland?
linux 5.3.5-1-MANJARO, x11

PR #2927, opened 2019-10-28
Fixes selection not being reset on escape send

kchibisov:
This is probably a bit controversial change, however I don't see much difference between `Action::Esc` and arbitrary user input for selection reset.

What do you think @chrisduerr, @nixpulvis?

Fixes: #2925


alacritty/src/input.rs
```diff
@@ -710,6 +710,10 @@ impl<'a, T: EventListener, A: ActionContext<T> + 'a> Processor<'a, T, A> {
 
                 // Don't suppress when there has been a `ReceiveChar` action
                 *suppress_chars.get_or_insert(true) &= binding.action != Action::ReceiveChar;
+
+                if let Action::Esc(_) = binding.action {
+                    self.ctx.clear_selection();
+                }
```
chrisduerr: Can't we move this into the `execute` function?

kchibisov: We don't want all the bindings to reset selection. For example, SrollPageUp/ScrollPageDown shouldn't reset selection imho (or I wouldn't be able to select text on a laptop), so thats an idea of making it for `Esc` only.

chrisduerr: > Oh, are you about just moving clear_selection to execute's Action::Esc arm?

Yes.

CHANGELOG.md
```diff
@@ -75,6 +75,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Use `\` instead of `\\` as path separators on Windows for logging config file location
 - Underline/strikeout drawn above visual bell
 - Terminal going transparent during visual bell
+- Fix selection not being cleared when sending chars
```
```suggestion
- Selection not being cleared when sending chars through a binding
```

nixpulvis:
I like the idea behind this. Nice UX improvement IMO.

PR #2935, opened 2019-10-30
Match font family & style for fontconfig fallback

keidax:
In the previous behavior, when looking for a missing glyph, fontconfig will just return some font containing that glyph based on its own ordering preference. It doesn't try to match the user's configured font; it might not even be monospace. This is probably (at least partially) the cause of https://github.com/jwilm/alacritty/issues/1560.

This is my first time contributing in Rust, and I'm not very familiar with the structure of this project yet. So I took the simplest approach of reading back the values already loaded in FreeType. If there's an easy way to get the values directly from the user config, I'm happy to do that instead.

### DIFF
--- a/font/src/ft/fc/mod.rs
+++ b/font/src/ft/fc/mod.rs
@@ -47,6 +47,11 @@ pub use self::pattern::{Pattern, PatternRef};
 ///
 /// The returned pattern is the result of Pattern::render_prepare.
 pub fn font_match(config: &ConfigRef, pattern: &mut PatternRef) -> Option<Pattern> {
+    if log_enabled!(log::Level::Debug) {
+        debug!("Matching font with pattern:");
+        pattern.print();
+    }
+
     pattern.config_substitute(config, MatchKind::Pattern);
     pattern.default_substitute();
 
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -304,7 +304,8 @@ impl FreeTypeRasterizer {
         if use_initial_face {
             Ok(glyph_key.font_key)
         } else {
-            let key = self.load_face_with_glyph(c).unwrap_or(glyph_key.font_key);
+            let key =
+                self.load_face_with_glyph(c, glyph_key.font_key).unwrap_or(glyph_key.font_key);
             Ok(key)
         }
     }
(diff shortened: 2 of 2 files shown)


font/src/ft/fc/mod.rs
```diff
@@ -47,6 +47,11 @@ pub use self::pattern::{Pattern, PatternRef};
 ///
 /// The returned pattern is the result of Pattern::render_prepare.
 pub fn font_match(config: &ConfigRef, pattern: &mut PatternRef) -> Option<Pattern> {
+    if log_enabled!(log::Level::Debug) {
+        debug!("Matching font with pattern:");
+        pattern.print();
+    }
```
keidax: I found this helpful for debugging, not sure if it would be appropriate to keep

chrisduerr:
> In the previous behavior, when looking for a missing glyph, fontconfig will just return some font containing that glyph based on its own ordering preference. It doesn't try to match the user's configured font

I don't quite understand the issue here exactly. If the glyph is missing, that means it's not part of the configured font, so we should just use fontconfig to lookup the first alternative font that does support it.

Why would we include the currently configured font when we're looking for missing glyphs?

The linked fallback issue also is about not falling back to other fonts, instead of not falling back to the configured font. So I don't think this resolves this problem.

Can you share an example of what this fixes for you?

font/src/ft/fc/mod.rs
chrisduerr: I think we should remove this.

keidax:
Yep, sorry, still planning to follow up! I just haven't had a chance to put together a good demonstration.

chrisduerr:
> Yep, sorry, still planning to follow up! I just haven't had a chance to put together a good demonstration.

No worries, just making sure nothing is missed. Let me know if you need any help.

intrntbrn:
Just confirming that this commit solved my font fallback issues (on arch). Good job.

chrisduerr:
@intrntbrn Ah, that's very useful information.

Could you go into a little more detail on how the font fallback was incorrect previously and how it behaves now in the 'corrected' form? Are you aware of which 'incorrect' font was loaded and specifically which glyph was incorrect?

intrntbrn:
I renamed a font to `Termint14n` to make sure they are no rules from my fontconfig applied. Then i created a fallback definition for the new font in `~/.fonts/fonts.conf`: 
```	
	<alias>
        <family>Termint14n</family>
        <prefer>
            <family>RobotoMono Nerd Font Mono</family>
        </prefer>
    </alias>
```
The fallback was set correctly on termite and kitty but not on alacritty. After compiling with this patch my fontconfig gets respected now without further configuration.

kchibisov:
This is a pure luck that it worked on this PR tbh. This PR doesn't solve any issue I personally have with fallback and it doesn't do anything about it. However the current master works **correctly** according to fontconfig, but not how the users are testing/expected. You should query with `fc-match -s ":charset=b104"`(This value for charset could be wrong), and so on to see how master is trying to match things. However the thing most users are likely want is  #3163, it's already was proved to solve issues for some folks on IRC, unlike this PR and solves my issues with fallbacks and some performance related things. It also implements all the functionality this PR has, but in a that make sense. So on #3163 your fallback list in alacritty is identical to `fc-match -s "FAMILY_FROM_CONFIG:style=STYLE_FROM_CONFIG:pixelsize=SIZE_FROM_CONFIG_CONVERTED_TO_PIXELSIZE"`.

chrisduerr:
Closing this in favor of #3163. If there's anything in this PR that is not resolved by #3163, please let me know and we can try to integrate that into the new PR.

PR #2936, opened 2019-10-30
Windows: fix unused imports

davidhewitt:
Resolves a couple of unused imports warnings when building on windows.

The only other warnings I get are some deprecation warnings originating from the `clap::crate_authors!` macro. Looks like they'll be resolved in the next clap version: https://github.com/clap-rs/clap/pull/1558


chrisduerr:
> The only other warnings I get are some deprecation warnings originating from the clap::crate_authors! macro

Those have been haunting all platforms for a while unfortunately. I'm glad the PR was finally merged though, that's good to hear.

PR #2942, opened 2019-10-31
Update Windows config path in README.md

thepiper:
config file specifically saves into ...\AppData\Roaming\alacritty\ in windows


chrisduerr:
There's a difference between the `AppData` folder and the `%APPDATA%` directory. The `%APPDATA%` environment variable points to `AppData/Roaming` by default as far as I know.

So everything's working exactly as expected.

PR #2946, opened 2019-11-02
Term::reset_url_highlight should not always dirty Term

kennylevinsen:
Term::reset_url_highlight is called on cursor move, so dirtying
unconditionally leads to excessive redraw.


kchibisov:
@kennylevinsen that make sense, however we have a massive URL rework, which is going to land pretty soon (before v0.3.4), so I recommend you to base you work on it, however your issue is probably fixed there.

https://github.com/jwilm/alacritty/pull/2896

chrisduerr:
See https://github.com/jwilm/alacritty/pull/2920.

PR #2948, opened 2019-11-03
Fix F26Dot6-conversion (rounding issue)

cynecx:
Fixes https://github.com/jwilm/alacritty/issues/2780.

### DIFF
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -86,7 +86,7 @@ pub struct FreeTypeRasterizer {
 
 #[inline]
 fn to_freetype_26_6(f: f32) -> isize {
-    ((1i32 << 6) as f32 * f) as isize
+    ((1i32 << 6) as f32 * f).round() as isize
 }
 
 #[inline]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fontconfig's `autohint` and `hinting` options being ignored
 - Ingoring of default FreeType properties
 - Alacritty crashing at startup when the configured font does not exist
+- Font size rounding error
 
 ### Removed
 


cynecx:
Rendered (`Hack, 9.5pt`):

Untitled

font/src/ft/mod.rs
```diff
@@ -68,7 +68,7 @@ pub struct FreeTypeRasterizer {
 
 #[inline]
 fn to_freetype_26_6(f: f32) -> isize {
-    ((1i32 << 6) as f32 * f) as isize
+    ((1i32 << 6) as f32 * f).ceil() as isize
```
chrisduerr: Why `ceil()` here? It seems like that would cause the same issue as floor, depending on input value.

cynecx: Well, the bug sounded like an off by one issue, so I dig around and came across this little conversion and thought that ceil semantics would actually be more correct.

This actually seemed to work so I googled further around to see if other projects were actually doing the same, and it turns out that at least cairo and kitty were also rounding up:

https://chromium.googlesource.com/chromiumos/third_party/cairo/+/master/src/cairo-ft-font.c#784 (cairo)
https://github.com/kovidgoyal/kitty/blob/master/kitty/freetype.c#L190 (kitty)

Besides that, I am not really an expert in this area, so I can't say for sure if this "should be the correct behavior".

chrisduerr: The cairo code you've linked isn't rounding up. It's rounding.

Adding 0.5 and then converting from float to int is a way to round for positive numbers.

floor(1.3 + 0.5) = 1
floor(ceil(1.3)) = 2

cynecx: Oh yes, smh, it's rounding. Though, it seems to "also" work with rounding (Looks like I was lucky.) 

Size: 9.5pt
px size: 9.5 * 96.0 / 72.0 ~= 12.666667

master: to_freetype_26_6(12.666667) = 810
rounding would produce 811, and that one difference really makes a diffenrence.

Well, yeah, I also not entirely sure if we are just shifting the issue around, but by actually rounding the value (to_freetype_26_6), we could at least get the same behavior as cairo and kitty (-ish since it's ceiling and not rounding).

chrisduerr: > we could at least get the same behavior as cairo and kitty (-ish since it's ceiling and not rounding).

As you've mentioned, Kitty is indeed ceiling and not rounding. So I'm a bit doubtful that this change would fix things in general, since both sizes should be valid.

Certain combinations of DPI and font size would still be able to produce any possible value here (since DPI can be any fraction). So I'd say this is almost guaranteed not to fix the issue in all cases.

The only change that could be reliable here would be to fix the font size to a multiple of 64 by rounding the font size before passing it to `to_freetype_26_6`. However I don't think anyone else is doing that, so I'm not sure if that's an issue.

kchibisov:
Would you mind sharing a screenshot on how things are getting underlined before your patch and after, with the next small test suit?

`echo -e "\e[4mhelp ggg\e[0m"`

cynecx:
@kchibisov Well there is no test-suite, I've just taken screenshot and zoomed/cropped them together. Also about the underline thing, there is no difference when I compared both rendered text (the underline is positioned the same as before).

kchibisov:
@cynecx I gave you an escape that underlines a text, probably `suit` was a bad word here, and I should use `case` instead. Executing the line I gave you looks like this (see underline)

Screenshot_Mon 04 Nov 2019 01:36:57 AM MSK

However I was confused by your picture, since it's not really a screenshot (it's a screenshot + your captures).

I don't see much difference at all on my system though, however I'm using hidpi screen, so fonts are always looking fine for me.

Anyway, if we decide to go with some approach, I'd rather choose a more common one (cairo), since `ceil` doesn't make much sense to me honestly.

font/src/ft/mod.rs
cynecx: > The only change that could be reliable here would be to fix the font size to a multiple of 64 by rounding the font size before passing it to to_freetype_26_6. However I don't think anyone else is doing that, so I'm not sure if that's an issue.

I am not sure I quite follow. We wouldn't be able to specify fractional points (pixel?) then, since the fractional bits would always be zero, if the size would be a multiple of 64.

In my case, I actually just want to be able to use `12.5` as size forwarded to `set_char_size` and skip the "pt to px" calculation entirely because right now specifying `9.375` as size in alacritty would not be possible because `font::Size`'s granularity (factor 2) is too coarse.

Perhaps, another idea to "fix" this would be to completely skip the conversion and be able to specify the "px size". Or another hacky solution would be to increase the factor in `font::Size` to 8, so sizes like `9.375` can be represented which will turn to the desired `12.5` (9.375 * 96.0 / 72.0 = 12.5 -> * 64 = 800).

cynecx:
test

Left: This PR; Right: Master (You can notice the subtle difference, that the height is shorter by a pixel.)

> I don't see much difference at all on my system though, however I'm using hidpi screen, so fonts are always looking fine for me.

Yes, that's probably it, I don't see any issues on my macbook pro (with retina) either.

> Anyway, if we decide to go with some approach, I'd rather choose a more common one (cairo), since ceil doesn't make much sense to me honestly.

This PR should do what cairo is basically doing, however as @chrisduerr was saying, maybe we are just shifting this issue to somewhere else. I am not really an expert in this kind of area, so I can't really say.

cynecx:
Just as a side note: Further research has also shown that Chromium & Co. which are using skia for 2d rendering, are also using alacritty's master semantics:

https://github.com/google/skia/blob/master/src/ports/SkFontHost_FreeType.cpp#L957

```c
src/core/SkFDot6.h
64:#define SkScalarToFDot6(x)  (SkFDot6)((x) * 64)
```

However, I am able to specify the font-size in multiple formats (pt, px, ...), so setting the font-size to "12.5px" got me the same result as the rounding approach with "9.5pt".

Reproduced with Chromium 78 (skia rendering enabled):

```html
<!DOCTYPE html>
<html>
<head>
	<meta charset="utf-8">
	<title>Test Code</title>
	<style>
		body {
			background-color: #404552;
		}
		code {
		  font-family: "Hack";
		  font-size: 12.5px;
		  font-weight: bold;
		  color: #F58E8E;
		}
</style>
</head>
<body>
	<code>
		bldi
	</code>
</body>
</html> 
```

font/src/ft/mod.rs
cynecx: Well, another quick & hacky would be to increase `font::Size`'s factor to 8 and change the font-size to `9.375`. That also works.

chrisduerr:
One of the issues might be that we're scaling by DPI, so we're not just taking the size and forwarding it directly. So it's possible that passing a font size like `13.333333334` is what is creating issues and the fonts are not designed to handle that nicely. However I don't really know enough about creating fonts to judge this.

But generally if we round the size before passing it to the 26.6 format conversion function, it should disable subpixel shenanigans. Could you try if that fixes it for you too @cynecx?

cynecx:
@chrisduerr Changed:

```rust
to_freetype_26_6(size)
```
to
```rust
to_freetype_26_6(size.round())
```

And yes that fixes this issue too.

cynecx:
@chrisduerr How would you like to proceed?

chrisduerr:
> How would you like to proceed?

That's a very good question. I have zero confidence that any of the approaches taken so far will improve anything in general, so I don't think there's any point in merging things in the current state. Changing things around to a different wrong isn't going to help anyone.

I think we should try to find some definite source of truth here as to how this should be handled. I'd be surprised if rounding the result of the 26.6 conversion would be the answer, but to confirm that rounding our input font is reasonable we'd have to verify other applications are doing the same and currently it does not look like that.

So I think the best way forward would be to check what other terminal emulators are doing and if their font sizes are ever uneven the way ours are.

Issue #2330, opened 2019-04-18
Underline and Strikethrough Bleed Into Current Line on Backspace


nixpulvis:
When I run `printf "\e[4;9mTEST      asd\n"` once to get our next prompt stuck with terminal attributes of underline and strikethough, then type at least 2 characters, finally using the arrow keys (or <kbd>ctrl-a</kbd>) and delete any but the last character of the text you entered, and notice that the styles bleed onto the right of the terminal.

A gif's worth a million words (even if it's of horrid quality):
![QlrVhaj](https://user-images.githubusercontent.com/715947/56397609-99284f80-6212-11e9-8f39-f02079336406.gif)

```
Linux masva 5.0.7-arch1-1-ARCH #1 SMP PREEMPT Mon Apr 8 10:37:08 UTC 2019 x86_64 GNU/Linux
```

Ratione:
I also have this problem. I am also experiencing a (related?) rendering problem, where the Underline of the font is sometimes a few pixels and  cuts through the chars. 

Running Arch with the Awesome-Fonts. This issue was not present yesterday, and I don't believe I changed any config files.

PR #2952, opened 2019-11-04
Fix underline/strikeout bleed into current line on backspace

kchibisov:
Fixes: #2330


kchibisov:
I wonder whether we should perform similar things in other functions like `erase_chars`/`insert_blank`, etc?

chrisduerr:
> I wonder whether we should perform similar things in other functions like erase_chars/insert_blank, etc?

I think the flags and foreground should probably be removed almost everywhere where we use the template cell outside of the `input` function.

However, that has a big risk for regressions, so for every place where we change this (including this PR), we should have a testcase with escapes that tests both that removing flags/fg behaves exactly the same way it does in other terminals and that the background color is actually required (instead of just using `Cell::default()`.

With this case we already have an escape that proves that without this change things are buggy, but we should still confirm that we actually do need the background cell and that the foreground color needs to be reset.

kchibisov:
Just for the idea behind this PR. This function is doing pretty much the same as clear_line with right/left mode. It seems like it was forgotten to change it, when underline/strikeout were introduced. However adding more tests is probably a good idea. (we're clearing things in the end).

As for the other cases I'll take a look on how could I repro, what `xterm`/`vte` do here and just follow them.

chrisduerr:
> It seems like it was forgotten to change it, when underline/strikeout were introduced.

Just as an explanation, I've fixed all function that removed the `flags` before making use of the template cell, so that's why I haven't changed this one.

kchibisov:
> Are you willing to verify the correctness of the other areas where we haven't adjusted the usage of the template cell yet?

Yeah, I was a bit busy during the last week (and probably will for a few more days) in terms of writing code (I can always reply, but not sure about writing something). However I'm not sure, whether I should prioritize it above clipboard PR (I should write a crate for X11 clipboard and rework PR iirc).

chrisduerr:
I think this is more important than the clipboard PR, which likely isn't going to have an effect on most usecases.

However if you can't get to it, I can take a look too within the next week, so feel free to do whatever you're more excited about.

kchibisov:
@chrisduerr Are you fine with this ref test? Just to make clear on what we're expecting from them. I've used this escape `printf "\e[31;1;7;4;9mTEST asd\n"` in user configless bash.

CHANGELOG.md
```diff
@@ -80,6 +80,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Terminal going transparent during visual bell
 - Selection not being cleared when sending chars through a binding
 - Mouse protocols/encodings not being mutually exclusive within themselves
+- Underline/Strikeout bleeds into current line on backspace
```
kchibisov: Should I also update it to reflect all template fixes?

alacritty_terminal/tests/ref.rs
```diff
@@ -54,6 +54,12 @@ ref_tests! {
     zerowidth
     selective_erasure
     colored_reset
+    delete_chars_reset
+    alt_reset
+    deccolm_reset
+    decaln_reset
```
kchibisov: On `decaln` I've followed `vte` behavior, since `xterm` one was a bit strange here. Other then that I've been testing `vte` and `xterm`, sometimes `urxvt`.

CHANGELOG.md
chrisduerr: I feel like we should phrase it more generally probably.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1112,7 +1112,7 @@ impl<T> Term<T> {
 
     pub fn swap_alt(&mut self) {
         if self.alt {
-            let template = Cell { bg: self.cursor.template.bg, ..Cell::default() };
+            let template = self.cursor.template;
             self.grid.region_mut(..).each(|c| c.reset(&template));
```
chrisduerr: You inlined it every else, might as well do the same here, right? Or does this exceed the 100 column limit?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1160,7 +1160,7 @@ impl<T> Term<T> {
         self.set_scrolling_region(1, self.grid.num_lines().0);
 
         // Clear grid
-        let template = Cell { bg: self.cursor.template.bg, ..Cell::default() };
+        let template = self.cursor.template;
```
chrisduerr: You inlined it every else, might as well do the same here, right? Or does this exceed the 100 column limit?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1315,10 +1315,9 @@ impl<T: EventListener> ansi::Handler for Term<T> {
     }
 
     #[inline]
-    fn dectest(&mut self) {
-        trace!("Dectesting");
-        let mut template = self.cursor.template;
-        template.c = 'E';
+    fn decaln(&mut self) {
+        trace!("Decalnning");
+        let template = Cell { c: 'E', ..self.cursor.template };
```
chrisduerr: Trying out both XTerm and Termite, none of them seem to set the background color based on the template cell. We might just want to use `..Cell::default()` here?

CHANGELOG.md
```diff
@@ -80,6 +80,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Terminal going transparent during visual bell
 - Selection not being cleared when sending chars through a binding
 - Mouse protocols/encodings not being mutually exclusive within themselves
+- Cells reset with template cursor
```
chrisduerr: Template cursor is an internal implementation detail. We should document the behavior from a user perspective.

Maybe something more like this:

```suggestion
- Cell reset not clearing underline, strikeout and foreground color
```

alacritty_terminal/tests/ref.rs
chrisduerr: Could you go into more detail what difference you encountered and why VTE makes more sense?

I ran `decaln` myself using `echo -e "\e#8"` and wasn't able to see any difference.

alacritty_terminal/src/term/mod.rs
kchibisov: Borrow checker.

alacritty_terminal/src/term/mod.rs
kchibisov: Borrow checker.

alacritty_terminal/tests/ref.rs
kchibisov: Try the following in xterm. It should remove underline/strikeout effect.
```
printf "\e[31;1;4;9mTEST asd\n"
echo -e "\033#8"
```

alacritty_terminal/src/term/mod.rs
kchibisov: Oh, I've forgotten to check background thing in some escapes. Let me recheck all the thing.

kchibisov: Seems like `decaln` was the only one affected.

kchibisov:
In the last test I've fixed 3 our previous teste. With `row_reset` I can say that our reflow now working fine for me after I cat this test in full screen mode and then go back to windowed mode, however I'm not sure about vim tests in which we've just changed the color of whitescape chars from Rgb to Foreground, visually I don't see any difference, but who knows.

chrisduerr:
Have you actually checked why the tests were failing specifically, or did you just re-record them under the assumption they were incorrect?

We should be extremely careful about re-recording ref tests.

kchibisov:
They were failing because of a bunch of cells with WHITESPACE char with Rgb color before changed to WHITESPACE char and named foreground color, so since we're resetting fg color now it made sense.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1308,10 +1308,9 @@ impl<T: EventListener> ansi::Handler for Term<T> {
...
     #[inline]
-    fn dectest(&mut self) {
-        trace!("Dectesting");
-        let mut template = self.cursor.template;
-        template.c = 'E';
+    fn decaln(&mut self) {
+        trace!("Decalnning");
+        let template = Cell { c: 'E', ..Cell::default() };
 
         self.grid.region_mut(..).each(|c| c.reset(&template));
```
chrisduerr: Extremely minor complaint, but the placing of the newline here is a bit confusing to me. `template` and the reset code belong together, not the `trace` and `template` statement. So I think in general the newline should be placed above `let template` not below it.

Not really a big deal, but if you plan on changing anything else, this would be a nice minor improvement I think.

Issue #1934, opened 2018-12-28
Add UTF-8 mouse report support


DarkDefender:
> Which operating system does the issue occur on?
Linux

> If on linux, are you using X11 or Wayland?
X11

In weechat, mouse gestures (click and drag) and clicks do not register after a certain horizontal point in alacritty. This works in urxvt and xterm.

![ala_mouse](https://user-images.githubusercontent.com/312503/50513544-c224a900-0a98-11e9-9f81-8c436d2dabaf.png)

If I resize the terminal window to be smaller (so it is smaller than the red line area) mouse clicks and motions now work again at the far right again.

So I'm guessing something funky is happening with mouse actions on large cell count positions.

This is a bit of a problem for me as the nick list is located on the right for me and I can't scroll the nick list or click on names to message them unless I resize the window as it is now.

chrisduerr:
This is related to https://github.com/jwilm/alacritty/issues/698.

If I had to take a guess, weechat is probably using the normal mouse reports in Alacritty, but using sgr mouse reports in other terminal emulators. Normal mouse reports do not support anything past column 223 iirc, so that would lead to issues.

Does changing the `TERM` to something like `xterm-256color` resolve this issue? Then maybe there's some code in weechat which sets the mouse mode based on the terminfo string, rather than terminal capabilities.

DarkDefender:
I forgot to mention that I have tried with both `TERM` set to alacirtty and xterm. Both have the same problem.

chrisduerr:
Interesting, that means that both should be using the same mouse report mode. It would probably be worth investigating which mouse mode exactly is set by weechat. There's potentially some more issues with the SGR mouse mode? If the normal mouse mode is requested, then this is the intended behavior afaik.

chrisduerr:
I can't reproduce this issue. Is this still present for you @DarkDefender?

DarkDefender:
I can still reproduce this with 0.3.3. Has anything changed in the regard with master?

DarkDefender:
@chrisduerr also how many cells is the width of your terminal? You might need to use a very small font to trigger this

chrisduerr:
I can repro on my 1440p display. However haven't looked into the cause of this at all yet. I wouldn't expect this to be too difficult though.

kchibisov:
This issue is caused by a limitation of a normal mouse mode lines/cols to be less then `223`, so clicks after a certain point are just filtered out. See the following line.

[...]

PR #2955, opened 2019-11-04
UTF-8 mouse mode escape sequence (`CSI ? 1005 h` / `CSI ? 1005 l`)

kchibisov:
Fixes: #1934

### DIFF
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -366,21 +366,64 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
         }
     }
 
-    pub fn normal_mouse_report(&mut self, button: u8) {
+    pub fn utf8_mouse_report(&mut self, button: u8) {
         let (line, column) = (self.ctx.mouse().line, self.ctx.mouse().column);
 
-        if line < Line(223) && column < Column(223) {
-            let msg = vec![
+        if line < Line(2015) && column < Column(2015) {
+            let mouse_pos_encode = |pos: usize| -> (u8, u8) {
+                let pos = 32 + 1 + pos;
+                let first = 0xC0 + (pos >> 6);
+                let second =  0x80 + (pos & 0x3F);
+                (first as u8, second as u8)
+            };
+
+            let mut msg = vec![
                 b'\x1b',
                 b'[',
                 b'M',
                 32 + button,
-                32 + 1 + column.0 as u8,
-                32 + 1 + line.0 as u8,
             ];
+
+            if column < Column(95) {
+                msg.push(32 + 1 + column.0 as u8);
+            } else {
+                let encoded_col = mouse_pos_encode(column.0);
+                msg.push(encoded_col.0);
+                msg.push(encoded_col.1);
+            }
+
+            if line < Line(95) {
+                msg.push(32 + 1 + line.0 as u8);
+            } else {
+                let encoded_line = mouse_pos_encode(line.0);
+                msg.push(encoded_line.0);
+                msg.push(encoded_line.1);
+            }
 
             self.ctx.write_to_pty(msg);
         }
+    }
+
+    pub fn normal_mouse_report(&mut self, button: u8) {
+        if self.ctx.terminal().mode().contains(TermMode::UTF8_MOUSE) {
+            self.utf8_mouse_report(button);
+        } else {
+            let (line, column) = (self.ctx.mouse().line, self.ctx.mouse().column);
+
+            if line < Line(223) && column < Column(223) {
+                let msg = vec![
+                    b'\x1b',
+                    b'[',
+                    b'M',
+                    32 + button,
+                    32 + 1 + column.0 as u8,
+                    32 + 1 + line.0 as u8,
+                ];
+
+                self.ctx.write_to_pty(msg);
+            }
+        }
+
     }
 
     pub fn sgr_mouse_report(&mut self, button: u8, state: ElementState) {
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -381,7 +381,7 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
         let mouse_pos_encode = |pos: usize| -> Vec<u8> {
             let pos = 32 + 1 + pos;
             let first = 0xC0 + pos / 64;
-            let second = 0x80 + (pos & 0x3F);
+            let second = 0x80 + (pos & 63);
             vec![first as u8, second as u8]
         };
 
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -366,64 +366,38 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
         }
     }
 
-    pub fn utf8_mouse_report(&mut self, button: u8) {
-        let (line, column) = (self.ctx.mouse().line, self.ctx.mouse().column);
-
-        if line < Line(2015) && column < Column(2015) {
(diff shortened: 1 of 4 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -436,23 +436,24 @@ pub mod mode {
...
+            const MOUSE_MOTION        = 0b0000_0000_0100_0000;
+            const LINE_WRAP           = 0b0000_0000_1000_0000;
+            const LINE_FEED_NEW_LINE  = 0b0000_0001_0000_0000;
+            const ORIGIN              = 0b0000_0010_0000_0000;
+            const INSERT              = 0b0000_0100_0000_0000;
+            const FOCUS_IN_OUT        = 0b0000_1000_0000_0000;
+            const ALT_SCREEN          = 0b0001_0000_0000_0000;
+            const MOUSE_DRAG          = 0b0010_0000_0000_0000;
+            const MOUSE_MODE          = 0b0010_0000_0100_1000;
+            const UTF8_MOUSE          = 0b0100_0000_0000_0000;
```
chrisduerr: `MOUSE_MODE` should include `UTF8_MOUSE`.

alacritty/src/input.rs
```diff
@@ -366,23 +366,66 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
+                msg.push(encoded_line.0);
+                msg.push(encoded_line.1);
+            }
+
             self.ctx.write_to_pty(msg);
         }
     }
 
+    pub fn normal_mouse_report(&mut self, button: u8) {
+        if self.ctx.terminal().mode().contains(TermMode::UTF8_MOUSE) {
```
chrisduerr: I think we shouldn't have the utf8 mouse handling in `normal_mouse_report`, that is extremely confusing. Instead it should be moved to the `mouse_report` function just like all the other mouse reports.

That way `normal_mouse_report` shouldn't change at all, which makes a lot of sense too.

alacritty/src/input.rs
```diff
@@ -366,23 +366,66 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
         }
     }
 
-    pub fn normal_mouse_report(&mut self, button: u8) {
+    pub fn utf8_mouse_report(&mut self, button: u8) {
         let (line, column) = (self.ctx.mouse().line, self.ctx.mouse().column);
 
-        if line < Line(223) && column < Column(223) {
-            let msg = vec![
+        if line < Line(2015) && column < Column(2015) {
```
chrisduerr: Should early return here to decrease level of indentation.

alacritty/src/input.rs
```diff
@@ -366,23 +366,66 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
+                msg.push(encoded_col.1);
+            }
+
+            if line < Line(95) {
+                msg.push(32 + 1 + line.0 as u8);
+            } else {
+                let encoded_line = mouse_pos_encode(line.0);
+                msg.push(encoded_line.0);
+                msg.push(encoded_line.1);
+            }
```
chrisduerr: Disregarding everything I've said about moving the call to the `utf8_mouse_report` into the `mouse_report` function, couldn't we just merge the normal and utf8 code together like this?

```rust
// I'm not actually sure if '>=' is correct here, that's what we're doing right now
// following the xterm docs though, they say the last position is 223/2015, making it sound inclusive
if (utf8 && line >= 2015) || (!utf8 && line >= 223) { return }

if utf8 && line < Line(95) {
    // ...
} else if utf8 {
    // ...
}
```

alacritty/src/input.rs
```diff
@@ -366,23 +366,66 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
     }
 
-    pub fn normal_mouse_report(&mut self, button: u8) {
+    pub fn utf8_mouse_report(&mut self, button: u8) {
         let (line, column) = (self.ctx.mouse().line, self.ctx.mouse().column);
 
-        if line < Line(223) && column < Column(223) {
-            let msg = vec![
+        if line < Line(2015) && column < Column(2015) {
+            let mouse_pos_encode = |pos: usize| -> (u8, u8) {
```
chrisduerr: If we return a slice here, we can directly append all elements to `msg` instead of having to use this tuple.

alacritty/src/input.rs
```diff
@@ -366,23 +366,66 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
+                let pos = 32 + 1 + pos;
+                let first = 0xC0 + (pos >> 6);
```
chrisduerr: Should improve readability a bit.

```suggestion
                let first = 0xC0 + pos / 64;
```

alacritty_terminal/src/term/mod.rs
kchibisov: It shouldn't, it's the same thing as `MOUSE_SGR`, which is not a part of MOUSE_MODE, UTF8_MOUSE is an optional encoding here. Am I missing something?

alacritty/src/input.rs
kchibisov: I don't know about this one, I prefer bit operations when manipulating with bits.

alacritty/src/input.rs
kchibisov: In code they're sending `zero` for 2015 and 233. With doesn't make sense to me, but maybe this is how it's supposed to work.

alacritty_terminal/src/term/mod.rs
chrisduerr: Oh, I thought that SGR was included too, completely missed that. Good point.

alacritty/src/input.rs
chrisduerr: I'm not sure how that is related to this comment?

alacritty/src/input.rs
chrisduerr: We're just dividing by 64 though, we're not `manipulating with bits`.

For the bitwise and it makes sense, but for this it's much, much more difficult to clearly understand the code just by reading it.

alacritty/src/input.rs
```diff
@@ -368,19 +368,36 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
-                b'\x1b',
-                b'[',
-                b'M',
-                32 + button,
-                32 + 1 + column.0 as u8,
-                32 + 1 + line.0 as u8,
-            ];
-
-            self.ctx.write_to_pty(msg);
+        let max_point = if utf8 { 2015 } else { 223 };
```
chrisduerr: I like it, good idea.

alacritty/src/input.rs
```diff
@@ -368,19 +368,36 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
+        if line >= Line(max_point) || column >= Column(max_point) {
+            return;
+        }
+
+        let mut msg = vec![b'\x1b', b'[', b'M', 32 + button];
+
+        let mouse_pos_encode = |pos: usize| -> Vec<u8> {
+            let pos = 32 + 1 + pos;
+            let first = 0xC0 + pos / 64;
+            let second = 0x80 + (pos & 0x3F);
```
chrisduerr: I think if we replace `0x3F` with `63` here, things suddenly start making much more sense. Like that it should be possible to see the correlation between the different statements and what they're supposed to do.

I don't think there's much value in changing `0xC0` and `0x80`, those values seem mostly arbitrary to me.

PR #2960, opened 2019-11-05
set env variable, LC_CTYPE=UTF-8, fixes #2939

### LINKED ISSUE #2939: zsh autocompletion is malformed
- `macOS 10.14.3`
- `alacritty 0.3.3`
- shell: oh-my-zsh: `zsh 5.3 (x86_64-apple-darwin18.0`

# Expected

when I type `cd` and then press `tab` I expect the shell suggestions to be displayed clearly

Screenshot 2019-10-31 at 17 00 51

# How it is

before pressing tab:

Screenshot 2019-10-31 at 17 08 00

after:

Screenshot 2019-10-31 at 17 08 07

or: 

Screenshot 2019-10-31 at 16 59 55

debris:


kchibisov:
Thanks for the PR, however this is not a right way to solve this problem. We should get a proper locale from a system, otherwise it's a work around specific problem on *your* system.

If you'd like to work on this bug, I'd recommend you to look at our macOS specific file for setting locale https://github.com/jwilm/alacritty/blob/master/alacritty_terminal/src/locale.rs and understand why we're not getting a proper locale from a system, compared to `Terminal.app`. Looking at other open source terminals, which support macOS, could help here.

chrisduerr:
Closing since this isn't likely going to lead anywhere. If there is an actual fix, we can always open a new PR. But I suspect it might be related to system configuration.

PR #2964, opened 2019-11-09
Add ref test for verifying colored clear behavior

chrisduerr:
This covers the behavior of clearing the screen and a row with colored
cells.

This covers a bug discovered in #2329 which was not detected in any
existing ref tests.


nixpulvis:
Nice. I was going to bug you to send me that test. Now it appears I don't have to ;)

PR #2966, opened 2019-11-10
Add escape for reading clipboard

chrisduerr:


kchibisov:
Are we trying to follow what xterm does or we'd have our own approach on handling this?

Just reading through xterm spec https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands . It seems it's a bit more tricky then just check first char of `1` parameter.

> 
> The first, Pc, may contain zero or more characters from the set c , p , q , s , 0 , 1 , 2 , 3 , 4 , 5 , 6 , and 7.

chrisduerr:
> It seems it's a bit more tricky then just check first char of 1 parameter.

That's correct, however I don't think we need to support the full spec. Our clipboard library doesn't even support all these different buffers, so there's no way we could support them. So we just support the part of the spec that is most likely to be used.

So we do `p/s` => selection clipboard and `c` => clipboard. The rest `q/0-7` we can't support with our clipboard library, so we just throw it away.

I think that's a reasonable minimal implementation to follow and I don't expect that we really need to support any more than this. Considering how most terminal emulators (like VTE and URxvt) don't support this at all and even XTerm requires configuration to support setting the clipboard data.

kchibisov:
I'm not worry about supporting full spec, I worry about your `p`,`s`,`c` parameters being missed, due to things like `0pc`, since the spec isn't clear here. Am I correct that it shouldn't happen?

chrisduerr:
> I'm not worry about supporting full spec, I worry about your p,s,c parameters being missed, due to things like 0pc, since the spec isn't clear here. Am I correct that it shouldn't happen?

Honestly I don't 100% know myself how the other buffers work, but `0c` as a parameter in the second position does not work with xterm. My assumption is that the buffers are set using `q0` - `q7` as second parameter and `p` `c` `s` are only valid on their own.

So I think we should be good the way we're doing it here. We're not supporting `q` and the multiple buffers, but that's just a completely missing feature, but everything else involving the clipboards we support ourselves seems to work exactly as it does in XTerm.

If you want to test it yourself and find anything else, please let me know. But I think our behavior should be correct for the things we do support.

kchibisov:
> Honestly I don't 100% know myself how the other buffers work, but 0c as a parameter in the second position does not work with xterm.

That's what I'd like to hear. I was wondering whether I can write to Selection and Clipboard, by sending just one escape, since xterm spec says, you can pass more then one char, but doesn't emphasize that `p`,`c`,`s` should be used on their own.

So, looks good to me.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1718,8 +1718,33 @@ impl<T: EventListener> ansi::Handler for Term<T> {
...
+    fn write_clipboard<W: io::Write>(&mut self, clipboard: u8, writer: &mut W) {
+        let clipboard_type = match clipboard {
+            b'c' => ClipboardType::Clipboard,
+            b'p' | b's' => ClipboardType::Selection,
+            _ => return,
+        };
+
+        let text = self.clipboard.load(clipboard_type);
+        let base64 = base64::encode(&text);
+        let escape = format!("\x1b]52;{};{}\x07", clipboard as char, base64);
```
kchibisov: I wonder how we should map this behavior to upcoming clipboard rework(using request_load). Do you think that providing optional callback would make sense for such thing?

chrisduerr: This is actually a very good question. Currently we just have access to a `writer` which is direct mutable access to the PTY's fd.

However we already have things in place for writing to PTY on callback, since our `write_to_pty` uses channels. So I think we'd just have to pass the `event_proxy` and then have it write directly to PTY itself, the same way we load from clipboard when pressing the key combinations.

So I don't think this should be any problem! Especially since our `Term` already has an `event_proxy` available.

kchibisov: The problem is not to write to a pty, the problem is to write a correct string. So I'm talking about optional transform for a loaded text `|String| -> String` lambda, since our async clipboard loader can write_to_pty, however it can't transform loaded text. For bracketed paste we're checking terminal mode, which is fine, however for this thing, we don't have any mode (and I don't want to add one for it).

chrisduerr: Oh, right. Since we need to do more than just load the text.

I think we might want to rethink our loading strategy completely tbh. Automatically writing the output to the event loop seems like a bad idea anyways, we should have something more like this:

```rust
clipboard.load(|data| {
    event_proxy.send(Paste(data));
});
```

Having the clipboard data do more than just load the clipboard seems like a strange design decesion to me anyways, even though I haven't thought about it before. Having a callback that is executed once the data is present makes a lot more sense.

kchibisov: I also like this strategy, but I'm worry about carrying event_proxy around, anyway that's enough info to work with for now. Thx.

PR #2967, opened 2019-11-10
Fix modifier inconsistencies

chrisduerr:
Fixes https://github.com/jwilm/alacritty/issues/2906.


ranisalt:
Why does it change every single class and function name when they could just be fixed?

nixpulvis:
@ranisalt if you read closely, `ModifiersState` is the name from `glutin` while we used to use our own wrapper named, simply `Modifiers`.

ranisalt:
Ah, now I see it, I will try and see if it fixes for me

PR #2971, opened 2019-11-12
Fix mouse protocols/encodings sets not being mutually exclusive within itself

kchibisov:


alacritty_terminal/src/term/mod.rs
```diff
@@ -1878,22 +1878,36 @@ impl<T: EventListener> ansi::Handler for Term<T> {
             },
             ansi::Mode::ShowCursor => self.mode.insert(TermMode::SHOW_CURSOR),
             ansi::Mode::CursorKeys => self.mode.insert(TermMode::APP_CURSOR),
+            // Mouse protocols are mutually exlusive
             ansi::Mode::ReportMouseClicks => {
                 self.mode.insert(TermMode::MOUSE_REPORT_CLICK);
+                self.mode.remove(TermMode::MOUSE_DRAG);
+                self.mode.remove(TermMode::MOUSE_MOTION);
```
chrisduerr: Doing something funky like `self.mode = (self.mode ^ MOUSE_MODE) | MOUSE_REPORT_CLICK` might be an option, though I find it a bit unreadable.

But I think consolidating the `remove`s still makes sense.

I also don't think we should order by name, but rather by logical order. We should first remove everything, then insert the new ones (or the other way around). I think doing remove -> insert -> remove is a bit confusing and it's easy to miss what's done with what exactly.

We also already have a `MOUSE_MODE` which matches all other modes. So in theory we can just do `self.mode.remove(TermMode::MOUSE_MODE); self.mode.insert(new_mode);`.

kchibisov: I think removing `MOUSE_MODE` and then setting a correct one it the most readable solution.

CHANGELOG.md
```diff
@@ -79,6 +79,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Underline/strikeout drawn above visual bell
 - Terminal going transparent during visual bell
 - Selection not being cleared when sending chars through a binding
+- Mouse protocol/encoding handling not following specification
```
chrisduerr: I had trouble finding the specification, I doubt others would have an easy time.

We should just describe what exactly we fixed so people don't have to go hunting for this mysterious 'specification'.

Something like this:
```suggestion
- Mouse protocols/encodings not being mutually exclusive within themselves
```

PR #2973, opened 2019-11-12
fix #1465 hide on macos

### LINKED ISSUE #1465: Proper support for (un)hinding on macOS
After testing the latest changes in dbcb5885adb1f0b23c29838ef8dd837212322409 (#1281) I noticed, that I can't regain focus, or rather partially, of the Alacritty window after using the `CMD + H` keystroke.

~The first part shows the normal behavior and the second part, after the key stroke was pressed. You can also see, that the window action buttons also don't work anymore (keystrokes are still working though).~

nitsky:
Fix #1465 by hiding the application using NSApplication.sharedApplication().hide(nil) on macOS rather than just hiding the window. I ran into this issue because tmux mouse mode no longer worked for me after pressing command-H and re-activating alacritty.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,6 +35,8 @@ dependencies = [
  "libc 0.2.65 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "notify 4.0.14 (registry+https://github.com/rust-lang/crates.io-index)",
+ "objc 0.2.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "objc_id 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "parking_lot 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustc_tools_util 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.102 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


kchibisov:
Thank you for the PR, but I have a strongly feeling, that this fix should belong to [winit](https://github.com/rust-windowing/winit) and not to `alacritty`. Could you try sending your patch to `winit`, I guess they should be interested in this fix too. I also recommend to repro it with their `window` example (`cargo run --example window` in `winit` repo, just tune it to show a bug to them).

nitsky:
@kchibisov I think that winit's implementation is actually correct, because it calls -[NSWindow orderOut:] in response to Window::set_visible. I just think that set_visible is the inappropriate action to take in response to the command-H shortcut on macOS, because "hide" on macOS means "hide application" not "hide window". Alacritty is the application, winit is the window library, so the "hide application" action belongs in the application. Does that logic make sense?

chrisduerr:
> Alacritty is the application, winit is the window library, so the "hide application" action belongs in the application. Does that logic make sense?

It does not. Winit is responsible for handling all our windowing, so we should upstream a `hide` function if our current handling isn't working properly.

Thanks for the PR, but I agree with @kchibisov here that we should attempt to fix this upstream first. Are you willing to send a PR for that yourself?

nitsky:
@chrisduerr I will be happy to submit a PR to winit. If they accept a `hide` function, will you accept a change to call that instead of `set_visible` when the target_os is macOS?

chrisduerr:
> If they accept a hide function, will you accept a change to call that instead of set_visible when the target_os is macOS?

Of course, though we would probably want to call `hide` on all platforms and let winit handle the platform destinctions.

PR #2985, opened 2019-11-15
Fix `CSI Ps M` deleting lines above cursor

### LINKED ISSUE #2984: Don't delete lines above the cursor on CSI Ps M
`Alacritty` should clamp `Ps` to not go above cursor, so if you're on the bottom line, typing `echo -e '\e[1000;M'` shouldn't delete all the thing above and should be clamped to `1`.

P.s. All the other terminals clamp it(vte based, xterm, kitty).

chrisduerr:
Fixes #2984.


alacritty_terminal/src/term/mod.rs
```diff
@@ -1564,9 +1564,12 @@ impl<T: EventListener> ansi::Handler for Term<T> {
...
     #[inline]
     fn delete_lines(&mut self, lines: Line) {
+        let origin = self.cursor.point.line;
+        let lines = min(self.lines() - origin, lines);
+
         trace!("Deleting {} lines", lines);
-        if self.scroll_region.contains(&self.cursor.point.line) {
-            let origin = self.cursor.point.line;
+
+        if self.scroll_region.contains(&self.cursor.point.line) && lines.0 > 0 {
```
kchibisov: Could we swap checks?

kchibisov:
I guess we need a regression test for this.

alacritty_terminal/src/term/mod.rs
chrisduerr: Probably.

chrisduerr:
> I guess we need a regression test for this.

I'd say it makes sense to test both the clear behavior at the bottom and when lines actually get deleted. Should be possible to cover both with the same test.

PR #2986, opened 2019-11-15
Fix DA report

kchibisov:


kchibisov:
Do we need more tests to cover the changes?

alacritty_terminal/src/ansi.rs
```diff
@@ -1119,29 +1121,25 @@ where
             }};
         }
 
-        match byte {
-            b'B' => configure_charset!(StandardCharset::Ascii),
-            b'D' => self.handler.linefeed(),
-            b'E' => {
+        match (byte, intermediates.get(0)) {
+            (b'B', num) => configure_charset!(StandardCharset::Ascii, num),
```
chrisduerr: We shouldn't name the intermediate `num`, since we're usually dealing with characters and that just gets confusing.

Just naming it `intermediate` should be good, even if that means we'll have to linebreak.

alacritty_terminal/src/ansi.rs
```diff
@@ -1119,29 +1121,25 @@ where
...
+            (b'Z', None) => self.handler.identify_terminal(self.writer),
+            (b'c', None) => self.handler.reset_state(),
+            (b'0', num) => configure_charset!(StandardCharset::SpecialCharacterAndLineDrawing, num),
+            (b'7', None) => self.handler.save_cursor_position(),
+            (b'8', Some(b'#')) => self.handler.dectest(),
+            (b'8', None) => self.handler.restore_cursor_position(),
+            (b'=', None) => self.handler.set_keypad_application_mode(),
+            (b'>', None) => self.handler.unset_keypad_application_mode(),
+            (b'\\', None) => (), /* String terminator, do nothing (parser handles as string
+                                   * terminator) */
```
chrisduerr: While this comment style is supported in Rust, I don't think there's any idiomatic way where it's ever used like this.

We should just place a `//` comment above `(b'\\', None)`.

alacritty_terminal/tests/ref.rs
```diff
@@ -55,6 +55,7 @@ ref_tests! {
     selective_erasure
     colored_reset
     delete_lines
+    terminal_identity_dropped
```
chrisduerr: We should use imperative here. So `drop_terminal_identity`.

But I don't think a reftest is the best choice here. I think unit tests would be superior.

alacritty_terminal/src/ansi.rs
```diff
@@ -1442,39 +1444,56 @@ mod tests {
     };
     use crate::index::{Column, Line};
     use crate::term::color::Rgb;
-    use std::io;
+    use std::io::{self, sink};
 
-    /// The /dev/null of `io::Write`
-    struct Void;
+    struct PseudoTermHandler {
```
chrisduerr: I think the word `Term` is a bit misleading here, since it doesn't handle terminals, but it handles escapes. So maybe something more like `PseudoEscapeHandler` would fit better. Though we don't need to be overly specific either, so we can also just do `PseudoHandler`.

I'm also wondering if we should use a name more commonly understood as a test replacement method. So maybe `MockHandler` or something like that would express the intention better? Especially because a `pseudoterminal` is something that does exist (PTY) and I don't think it's accurate here.

alacritty_terminal/src/ansi.rs
```diff
@@ -1483,15 +1502,74 @@ mod tests {
...
-            parser.advance(&mut handler, *byte, &mut Void);
+            parser.advance(&mut handler, *byte, &mut sink());
         }
 
         assert_eq!(handler.attr, Some(Attr::Bold));
     }
 
+    #[test]
+    fn parse_terminal_identity_csi() {
+        let bytes: &[u8] = &[0x1b, 0x5b, 0x31, 0x63];
```
chrisduerr: While I know what `0x1b` is and I think that's the best way to express that sequence in our codebase (since x1b is used everywhere), I have absolutely no clue what the other byte sequences are.

I'd be in favor of changing every readable byte to the byte char representation like this: `b'Z'`. That way I know immediately what we're testing, simply by looking at this input.

alacritty_terminal/src/ansi.rs
```diff
@@ -1442,39 +1444,56 @@ mod tests {
...
-    impl Handler for AttrHandler {
-        fn terminal_attribute(&mut self, attr: Attr) {
-            self.attr = Some(attr);
+        fn identify_terminal<W: io::Write>(&mut self, _: &mut W) {
+            self.identity_reported = true;
+        }
+
+        fn reset_state(&mut self) {
+            *self = Self::default();
         }
```
chrisduerr: Do you think it makes sense to define this as a method, when all it does is reset itself to the default anyways? I'm a bit torn on that.

If we replace `handler.reset_state()` with `handler = Handler::default()` we wouldn't really lose anything with regards to conciseness and it's 100% clear what exactly is happening, since we've lost a layer of abstraction. On the other hand `reset_state` is very self descriptive and might be easier to understand than reading `handler = Handler::default()` since the name describes the exact intention of the function.

alacritty_terminal/src/ansi.rs
```diff
@@ -1132,14 +1132,16 @@ where
...
+            },
             (b'7', None) => self.handler.save_cursor_position(),
             (b'8', Some(b'#')) => self.handler.dectest(),
             (b'8', None) => self.handler.restore_cursor_position(),
             (b'=', None) => self.handler.set_keypad_application_mode(),
             (b'>', None) => self.handler.unset_keypad_application_mode(),
-            (b'\\', None) => (), /* String terminator, do nothing (parser handles as string
-                                   * terminator) */
+            (b'\\', None) => (), // String terminator, do nothing (parser handles as string
+            // terminator)
```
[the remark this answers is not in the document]
kchibisov: This comment was auto updated by rustfmt, so I thought it should be fine. Will update it manually though.

alacritty_terminal/src/ansi.rs
kchibisov: That make sense, but I was thinking about it like parse input, reset state, parse more input, reset state. And this method is a part of a Handler trait, so adding it's implementation maked sense to me. So I'm voting for my approach with using trait's declared method.

alacritty_terminal/src/ansi.rs
```diff
@@ -1446,140 +1446,184 @@ mod tests {
...
+            0x1b, b'[', b'1', b'm', 0x1b, b'[', b'0', b'm', b' ', b' ', b' ', b' ', b' ', b' ',
+            b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ',
+            b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ',
+            b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ',
+            b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ',
+            b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ', b' ',
+            b' ', b' ', b' ', b'\r', b' ', b'\r', b'\r', 0x1b, b'[', b'0', b'm', 0x1b, b'[', b'2',
+            b'7', b'm', 0x1b, b'[', b'2', b'4', b'm', 0x1b, b'[', b'J', b'j', b'w', b'i', b'l',
+            b'm', b'@', b'j', b'w', b'i', b'l', b'm', b'-', b'd', b'e', b's', b'k', b' ', 0x1b,
+            b'[', b'0', b'1', b';', b'3', b'2', b'm', 0xe2, 0x9e, 0x9c, b' ', 0x1b, b'[', b'0',
```
kchibisov: Some bytes were left because they made no sense in char from (I was converting with python's chr func).

chrisduerr: Yeah, that's why I mentioned printable characters specifically. So that's perfectly fine.

CHANGELOG.md
```diff
@@ -81,6 +81,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Selection not being cleared when sending chars through a binding
 - Mouse protocols/encodings not being mutually exclusive within themselves
 - Escape `CSI Ps M` deleting lines above cursor when at the bottom of the viewport
+- Escape `CSI Ps c` honored with a wrong `Ps`
+- Handling of `ESC` with intermediates
```
chrisduerr: `ESC`s with intermediates have been handled before. We just haven't been filtering `ESC` escapes with invalid intermediates before.

```rust
- Ignore `ESC` escapes with invalid intermediates
```

alacritty_terminal/src/ansi.rs
chrisduerr: > And this method is a part of a Handler trait

Oh, that's right, I totally missed that. Let's go with this then.

PR #3009, opened 2019-11-23
Move renderer from alacritty_terminal to alacritty

kchibisov:


alacritty_terminal/src/cursor.rs
```diff
@@ -12,97 +12,13 @@
...
 
-/// Width/Height of the cursor relative to the font width
-pub const CURSOR_WIDTH_PERCENTAGE: i32 = 15;
-
 /// A key for caching cursor glyphs
 #[derive(Debug, Eq, PartialEq, Copy, Clone, Hash, Deserialize)]
 pub struct CursorKey {
     pub style: CursorStyle,
     pub is_wide: bool,
 }
```
chrisduerr: Having a file just for this one struct seems a bit excessive. In the `alacritty_terminal` crate this is only used in `term/mod.rs`, so it would probably make sense to move it there.

Issue #1864, opened 2018-12-05
Always render emojis with default color


jsatk:
<img width="738" alt="screen shot 2018-12-05 at 3 16 44 pm" src="https://user-images.githubusercontent.com/1005550/49550560-e11c8a80-f8a0-11e8-88e9-87fcbf6ba215.png">

I'm on macOS. Here's my ridiculous tmux status bar. I name all my windows emojis because it makes me happy. I tried out Alacritty. But was disheartened when I realized it inverts the color on any window tab with activity. And it's most noticeable if you use colorful emoji as your window titles.

chrisduerr:
Your terminal emulator might actually be telling Alacritty to invert the colors. Judging by the black/white mix, I'd assume that's exactly how that is achieved.

Now if emojis should be affected by that is a different question, but that's likely what's going on. So if you would set the manual color codes instead of using the inversion escape, that should resolve the problem.

jsatk:
That seems plausible, but this doesn't happen in Terminal or iTerm 2.

chrisduerr:
It's hard for me to test this since I do not have any colored emojis installed, but it should be pretty simple to try and reproduce that behavior.

Try running this escape sequence in bash:
`echo -e "\e[7mINVERTED: \U1F600\e[0mNORMAL: \U1F600"`

This should print one inverted emoji and one normal one. In theory both should look the same in iTerm/Terminal.app but look different in Alacritty.

Note that I don't think this works in the version of bash which is installed on macOS by default, so you might need to replace the unicode escapes with emojis yourself or update bash.

chrisduerr:
It should be possible to work around this by changing the formatting of the tmux statusline.

A more proper solution would be to prevent emojis from getting inverted though.

jsatk:
I'm not sure what you mean by "changing the formatting"?  It's the name of the window.  How do I tell tmux to render them not inverse?

jsatk:
@chrisduerr update: I noticed that this affects everything. Not just tmux. You probably already inferred that but I just wanted to make sure it was documented that emoji always invert their colors when highlighted.

chrisduerr:
Changed the title to include the following issue: https://github.com/jwilm/alacritty/issues/2160

felipesere:
@chrisduerr I'd like to give this issue a try!
Could you point me in the right direction? I wonder how to test that that emojis themselves are unaffected by inversion...

chrisduerr:
[...]

Issue #791, opened 2017-09-22
Glyphs sometimes cut off (need to draw background in single pass)


casey:
I'm getting a visual glitch which causes the right half of certain characters to be missing. Restarting alacritty fixes it.

It happens when working with han characters, which are double width. I haven't been able to figure out the exact situation when it occurs. It seems to occur most often after displaying a lot of characters, so perhaps it's due to a cache getting filled up.

In this image, the first han character on the screen is okay, and the last three are truncated.

![half-characters](https://user-images.githubusercontent.com/1945/30730024-a89c2482-9f18-11e7-95e1-e6aec2e34d14.png)

jwilm:
I believe this is due to us not drawing the entire background in a single pass. When the first glyph cache gets filled up, Alacritty starts batching draw calls by texture. This is supported by what you've described

> occur most often after displaying a lot of characters

and

> Restarting alacritty fixes it

I'm updating the title to reflect the required action here.

PR #3011, opened 2019-11-23
Colored emojis on Linux/BSD with freetype.

kchibisov:
This PR implements colored emojis support on Linux/BSD with freetype.

The Linux/BSD part of it is working fine, however the implementation requires some discussion, since integration is not that simple in the current structure and I don't want to tight the other platforms on how Linux/BSD operates here.

picture

~P.s. I'll fix macOS/Windows builds when we agree on integration.~

### DIFF
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -370,6 +370,7 @@ impl GlyphCache {
             // FIXME
             // Assume that all emojis are 2 width for now.
             let h = metrics.line_height as i32;
+
             if rasterized.colored && rasterized.width > h && rasterized.height > h {
                 rasterized = Self::downsample_bitmap(rasterized, h, h);
             }
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -336,7 +336,7 @@ impl GlyphCache {
             n_j += 1;
         }
         // FIXME: better top setting.
-        bitmap_glyph.top = bitmap_glyph.top / ratio as i32;
+        bitmap_glyph.top /= ratio as i32;
         bitmap_glyph.width = width as i32;
         bitmap_glyph.height = height as i32;
         bitmap_glyph.buf = scaled_buffer;
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -22,7 +22,9 @@ use std::sync::mpsc;
 use std::time::Duration;
 
 use fnv::FnvHasher;
-use font::{self, FontDesc, FontKey, GlyphKey, Rasterize, RasterizedGlyph, Rasterizer};
+use font::{
+    self, BitmapBuffer, FontDesc, FontKey, GlyphKey, Rasterize, RasterizedGlyph, Rasterizer,
+};
 use log::{error, info};
 use notify::{watcher, DebouncedEvent, RecursiveMode, Watcher};
 
@@ -1581,22 +1583,42 @@ impl Atlas {
         let offset_x = self.row_extent;
         let height = glyph.height as i32;
         let width = glyph.width as i32;
+        let colored;
 
         unsafe {
             gl::BindTexture(gl::TEXTURE_2D, self.id);
 
             // Load data into OpenGL
-            gl::TexSubImage2D(
-                gl::TEXTURE_2D,
-                0,
-                offset_x,
-                offset_y,
-                width,
-                height,
-                gl::RGB,
-                gl::UNSIGNED_BYTE,
-                glyph.buf.as_ptr() as *const _,
-            );
+            match &glyph.buf {
+                BitmapBuffer::RGB(buf) => {
+                    gl::TexSubImage2D(
+                        gl::TEXTURE_2D,
+                        0,
+                        offset_x,
+                        offset_y,
+                        width,
+                        height,
+                        gl::RGB,
+                        gl::UNSIGNED_BYTE,
+                        buf.as_ptr() as *const _,
+                    );
+                    colored = false;
+                },
+                BitmapBuffer::RGBA(buf) => {
+                    gl::TexSubImage2D(
+                        gl::TEXTURE_2D,
+                        0,
+                        offset_x,
+                        offset_y,
+                        width,
+                        height,
+                        gl::RGBA,
+                        gl::UNSIGNED_BYTE,
+                        buf.as_ptr() as *const _,
+                    );
+                    colored = true;
+                },
+            }
 
             gl::BindTexture(gl::TEXTURE_2D, 0);
             *active_tex = 0;
@@ -1616,7 +1638,7 @@ impl Atlas {
 
         Glyph {
             tex_id: self.id,
-            colored: glyph.colored,
+            colored,
             top: glyph.top as f32,
             width: width as f32,
             height: height as f32,
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -336,7 +336,8 @@ impl GlyphCache {
             n_j += 1;
         }
         // FIXME: better top setting.
-        bitmap_glyph.top /= ratio as i32;
+        let b_top = bitmap_glyph.top as f32;
+        bitmap_glyph.top = ((b_top / factor + b_top / ratio as f32) / 2.0)  as i32;
         bitmap_glyph.width = width as i32;
         bitmap_glyph.height = height as i32;
         bitmap_glyph.buf = scaled_buffer;
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -139,6 +139,7 @@ pub struct RectShaderProgram {
 #[derive(Copy, Debug, Clone)]
 pub struct Glyph {
     tex_id: GLuint,
+    colored: bool,
     top: f32,
     left: f32,
     width: f32,
@@ -462,9 +463,15 @@ impl Batch {
         Batch { tex: 0, instances: Vec::with_capacity(BATCH_MAX) }
     }
 
-    pub fn add_item(&mut self, cell: RenderableCell, glyph: &Glyph) {
+    pub fn add_item(&mut self, mut cell: RenderableCell, glyph: &Glyph) {
         if self.is_empty() {
             self.tex = glyph.tex_id;
+        }
+
+        if glyph.colored {
+            cell.fg.r = 255;
+            cell.fg.g = 255;
+            cell.fg.b = 255;
         }
 
         self.instances.push(InstanceData {
@@ -1091,6 +1098,7 @@ fn load_glyph(
         },
         Err(AtlasInsertError::GlyphTooLarge) => Glyph {
             tex_id: atlas[*current_atlas].id,
+            colored: false,
             top: 0.0,
             left: 0.0,
             width: 0.0,
@@ -1608,6 +1616,7 @@ impl Atlas {
 
         Glyph {
             tex_id: self.id,
+            colored: glyph.colored,
             top: glyph.top as f32,
             width: width as f32,
             height: height as f32,
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -472,13 +472,7 @@ impl Batch {
 
         if glyph.colored {
             // XXX Temporary workaround to prevent emojis being rendered with a wrong colors on, at
-            // least, dark backgrounds. To resolve this issue properly we should have a separate
-            // rendering pass for colored glyphs. For more info see #1864.
-            //
-            // The 255 is not a color, it'll be a part of forming an alpha mask later on. So a
-            // proper color of emojis could be achieved by setting fg to #ffffff and bg
-            // to #000000, however we can't change bg here, because it actually "changes
-            // bg" of a cell.
+            // least, dark backgrounds. For more info see #1864.
             cell.fg.r = 255;
             cell.fg.g = 255;
             cell.fg.b = 255;
@@ -1597,9 +1591,15 @@ impl Atlas {
             gl::BindTexture(gl::TEXTURE_2D, self.id);
 
             // Load data into OpenGL
-            let (format, buf, _) = match &glyph.buf {
-                BitmapBuffer::RGB(buf) => (gl::RGB, buf, colored = false),
-                BitmapBuffer::RGBA(buf) => (gl::RGBA, buf, colored = true),
+            let (format, buf) = match &glyph.buf {
+                BitmapBuffer::RGB(buf) => {
+                    colored = false;
+                    (gl::RGB, buf)
+                },
+                BitmapBuffer::RGBA(buf) => {
+                    colored = true;
+                    (gl::RGBA, buf)
+                },
             };
 
             gl::TexSubImage2D(
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -471,6 +471,13 @@ impl Batch {
         }
 
         if glyph.colored {
+            // XXX Temporary workaround to prevent emojis being rendered with a wrong colors on
+            // dark backgrounds. To resolve this issue properly we should have a separate rendering
+            // pass for colored glyphs. For more info #1864.
+            //
+            // The 255 is not a color, it'll be an alpha mask later on. So a proper color of
+            // emojis could be achieved by setting fg to #ffffff and bg to #000000, however we
+            // can't change bg here, because it actually "changes bg" of a cell.
             cell.fg.r = 255;
             cell.fg.g = 255;
             cell.fg.b = 255;
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -272,78 +272,6 @@ impl GlyphCache {
         Ok((regular, bold, italic, bold_italic))
     }
 
-    // TODO: - take a look on what options we have besides this simple algo.
-    //
-    // area-averaging algorithm to downsample a colored bitmap.
-    fn downsample_bitmap(
-        mut bitmap_glyph: RasterizedGlyph,
-        width: i32,
-        height: i32,
-    ) -> RasterizedGlyph {
-        let width = width as usize;
-        let height = height as usize;
-        let b_width = bitmap_glyph.width as usize;
-        let b_height = bitmap_glyph.height as usize;
-        let b_buf = &bitmap_glyph.buf;
-        let factor = (b_width as f32 / width as f32).max(b_height as f32 / height as f32);
-        let ratio = (factor + 1.) as usize;
-        let mut scaled_buffer = Vec::with_capacity(width * height * 3);
-
-        // Index into new image.
-        let mut n_j = 0;
-        let mut s_j = 0;
-
-        while n_j < height {
-            let mut n_i = 0;
-            let mut s_i = 0;
-            while n_i < width {
-                let mut r: u32 = 0;
-                let mut g: u32 = 0;
-                let mut b: u32 = 0;
-                let mut pixels_picked: u32 = 0;
-
-                let cap_y = std::cmp::min(s_j + ratio, b_height);
-                let cap_x = std::cmp::min(s_i + ratio, b_width);
-
-                let mut s_j = s_j;
-                while s_j < cap_y {
-                    let cur_pixel_index = s_j * b_width * 3;
-                    let mut s_i = s_i;
-                    while s_i < cap_x {
-                        r += b_buf[cur_pixel_index + s_i * 3] as u32;
-                        g += b_buf[cur_pixel_index + s_i * 3 + 1] as u32;
-                        b += b_buf[cur_pixel_index + s_i * 3 + 2] as u32;
-                        s_i += 1;
-                        pixels_picked += 1;
-                    }
-                    s_j += 1;
-                }
-
-                if pixels_picked == 0 {
-                    scaled_buffer.push(0);
-                    scaled_buffer.push(0);
-                    scaled_buffer.push(0);
-                } else {
-                    scaled_buffer.push((r / pixels_picked) as u8);
-                    scaled_buffer.push((g / pixels_picked) as u8);
-                    scaled_buffer.push((b / pixels_picked) as u8);
-                }
-
-                s_i += ratio;
-                n_i += 1;
-            }
-            s_j += ratio;
-            n_j += 1;
-        }
-        // FIXME: better top setting.
-        let b_top = bitmap_glyph.top as f32;
-        bitmap_glyph.top = ((b_top / factor + b_top / ratio as f32) / 2.0)  as i32;
-        bitmap_glyph.width = width as i32;
-        bitmap_glyph.height = height as i32;
-        bitmap_glyph.buf = scaled_buffer;
-        bitmap_glyph
-    }
-
     fn make_desc(
         desc: &config::FontDescription,
         slant: font::Slant,
@@ -367,14 +295,6 @@ impl GlyphCache {
         self.cache.entry(glyph_key).or_insert_with(|| {
             let mut rasterized =
                 rasterizer.get_glyph(glyph_key).unwrap_or_else(|_| Default::default());
-
-            // FIXME
-            // Assume that all emojis are 2 width for now.
-            let h = metrics.line_height as i32;
-
-            if rasterized.colored && rasterized.width > h && rasterized.height > h {
-                rasterized = Self::downsample_bitmap(rasterized, h, h);
-            }
 
             rasterized.left += i32::from(glyph_offset.x);
             rasterized.top += i32::from(glyph_offset.y);
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -471,13 +471,14 @@ impl Batch {
         }
 
         if glyph.colored {
-            // XXX Temporary workaround to prevent emojis being rendered with a wrong colors on
-            // dark backgrounds. To resolve this issue properly we should have a separate rendering
-            // pass for colored glyphs. For more info #1864.
+            // XXX Temporary workaround to prevent emojis being rendered with a wrong colors on, at
+            // least, dark backgrounds. To resolve this issue properly we should have a separate
+            // rendering pass for colored glyphs. For more info see #1864.
             //
-            // The 255 is not a color, it'll be an alpha mask later on. So a proper color of
-            // emojis could be achieved by setting fg to #ffffff and bg to #000000, however we
-            // can't change bg here, because it actually "changes bg" of a cell.
+            // The 255 is not a color, it'll be a part of forming an alpha mask later on. So a
+            // proper color of emojis could be achieved by setting fg to #ffffff and bg
+            // to #000000, however we can't change bg here, because it actually "changes
+            // bg" of a cell.
             cell.fg.r = 255;
             cell.fg.g = 255;
             cell.fg.b = 255;
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -272,6 +272,77 @@ impl GlyphCache {
         Ok((regular, bold, italic, bold_italic))
     }
 
+    // TODO: - take a look on what options we have besides this simple algo.
+    //
+    // area-averaging algorithm to downsample a colored bitmap.
+    fn downsample_bitmap(
+        mut bitmap_glyph: RasterizedGlyph,
+        width: i32,
+        height: i32,
+    ) -> RasterizedGlyph {
+        let width = width as usize;
+        let height = height as usize;
+        let b_width = bitmap_glyph.width as usize;
+        let b_height = bitmap_glyph.height as usize;
+        let b_buf = &bitmap_glyph.buf;
+        let factor = (b_width as f32 / width as f32).max(b_height as f32 / height as f32);
+        let ratio = (factor + 1.) as usize;
+        let mut scaled_buffer = Vec::with_capacity(width * height * 3);
+
+        // Index into new image.
+        let mut n_j = 0;
+        let mut s_j = 0;
+
+        while n_j < height {
+            let mut n_i = 0;
+            let mut s_i = 0;
+            while n_i < width {
+                let mut r: u32 = 0;
+                let mut g: u32 = 0;
+                let mut b: u32 = 0;
+                let mut pixels_picked: u32 = 0;
+
+                let cap_y = std::cmp::min(s_j + ratio, b_height);
+                let cap_x = std::cmp::min(s_i + ratio, b_width);
+
+                let mut s_j = s_j;
+                while s_j < cap_y {
+                    let cur_pixel_index = s_j * b_width * 3;
+                    let mut s_i = s_i;
+                    while s_i < cap_x {
+                        r += b_buf[cur_pixel_index + s_i * 3] as u32;
+                        g += b_buf[cur_pixel_index + s_i * 3 + 1] as u32;
+                        b += b_buf[cur_pixel_index + s_i * 3 + 2] as u32;
+                        s_i += 1;
+                        pixels_picked += 1;
+                    }
+                    s_j += 1;
+                }
+
+                if pixels_picked == 0 {
+                    scaled_buffer.push(0);
+                    scaled_buffer.push(0);
+                    scaled_buffer.push(0);
+                } else {
+                    scaled_buffer.push((r / pixels_picked) as u8);
+                    scaled_buffer.push((g / pixels_picked) as u8);
+                    scaled_buffer.push((b / pixels_picked) as u8);
+                }
+
+                s_i += ratio;
+                n_i += 1;
+            }
+            s_j += ratio;
+            n_j += 1;
+        }
+        // FIXME: better top setting.
+        bitmap_glyph.top = bitmap_glyph.top / ratio as i32;
+        bitmap_glyph.width = width as i32;
+        bitmap_glyph.height = height as i32;
+        bitmap_glyph.buf = scaled_buffer;
+        bitmap_glyph
+    }
+
     fn make_desc(
         desc: &config::FontDescription,
         slant: font::Slant,
@@ -295,6 +366,13 @@ impl GlyphCache {
         self.cache.entry(glyph_key).or_insert_with(|| {
             let mut rasterized =
                 rasterizer.get_glyph(glyph_key).unwrap_or_else(|_| Default::default());
+
+            // FIXME
+            // Assume that all emojis are 2 width for now.
+            let h = metrics.line_height as i32;
+            if rasterized.colored && rasterized.width > h && rasterized.height > h {
+                rasterized = Self::downsample_bitmap(rasterized, h, h);
+            }
 
             rasterized.left += i32::from(glyph_offset.x);
             rasterized.top += i32::from(glyph_offset.y);
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -1589,36 +1589,22 @@ impl Atlas {
             gl::BindTexture(gl::TEXTURE_2D, self.id);
 
             // Load data into OpenGL
-            match &glyph.buf {
-                BitmapBuffer::RGB(buf) => {
-                    gl::TexSubImage2D(
-                        gl::TEXTURE_2D,
-                        0,
-                        offset_x,
-                        offset_y,
-                        width,
-                        height,
-                        gl::RGB,
-                        gl::UNSIGNED_BYTE,
-                        buf.as_ptr() as *const _,
-                    );
-                    colored = false;
-                },
-                BitmapBuffer::RGBA(buf) => {
-                    gl::TexSubImage2D(
-                        gl::TEXTURE_2D,
-                        0,
-                        offset_x,
-                        offset_y,
-                        width,
-                        height,
-                        gl::RGBA,
-                        gl::UNSIGNED_BYTE,
-                        buf.as_ptr() as *const _,
-                    );
-                    colored = true;
-                },
-            }
+            let (format, buf, _) = match &glyph.buf {
+                BitmapBuffer::RGB(buf) => (gl::RGB, buf, colored = false),
+                BitmapBuffer::RGBA(buf) => (gl::RGBA, buf, colored = true),
+            };
+
+            gl::TexSubImage2D(
+                gl::TEXTURE_2D,
+                0,
+                offset_x,
+                offset_y,
+                width,
+                height,
+                format,
+                gl::UNSIGNED_BYTE,
+                buf.as_ptr() as *const _,
+            );
 
             gl::BindTexture(gl::TEXTURE_2D, 0);
             *active_tex = 0;
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -45,7 +45,6 @@ use euclid::{Point2D, Rect, Size2D};
 use super::{BitmapBuffer, FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph};
 
 pub mod byte_order;
-use self::byte_order;
 use self::byte_order::kCGBitmapByteOrder32Host;
 
 use super::Size;
@@ -525,8 +524,8 @@ impl Font {
 
         let rasterized_pixels = cg_context.data().to_vec();
 
-        let buf = if font.is_colored() {
-            BitmapBuffer::RGBA(byte_order::extract_rgba(&rasterized_pixel))
+        let buf = if self.is_colored() {
+            BitmapBuffer::RGBA(byte_order::extract_rgba(&rasterized_pixels))
         } else {
             BitmapBuffer::RGB(byte_order::extract_rgb(&rasterized_pixels))
         };
@@ -573,6 +572,9 @@ impl Font {
 
 #[cfg(test)]
 mod tests {
+
+    use super::BitmapBuffer;
+
     #[test]
     fn get_family_names() {
         let names = super::get_family_names();
@@ -594,11 +596,16 @@ mod tests {
             for c in &['a', 'b', 'c', 'd'] {
                 let glyph = font.get_glyph(*c, 72., false).unwrap();
 
+                let buf = match &glyph.buf {
+                    BitmapBuffer::RGB(buf) => buf,
+                    BitmapBuffer::RGBA(buf) => buf,
+                };
+
                 // Debug the glyph.. sigh
                 for row in 0..glyph.height {
                     for col in 0..glyph.width {
                         let index = ((glyph.width * 3 * row) + (col * 3)) as usize;
-                        let value = glyph.buf[index];
+                        let value = buf[index];
                         let c = match value {
                             0..=50 => ' ',
                             51..=100 => '.',
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -572,7 +572,6 @@ impl Font {
 
 #[cfg(test)]
 mod tests {
-
     use super::BitmapBuffer;
 
     #[test]
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -33,6 +33,7 @@ use core_text::font::{
 };
 use core_text::font_collection::create_for_family;
 use core_text::font_collection::get_family_names as ct_get_family_names;
+use core_text::font_descriptor::kCTFontColorGlyphsTrait;
 use core_text::font_descriptor::kCTFontDefaultOrientation;
 use core_text::font_descriptor::kCTFontHorizontalOrientation;
 use core_text::font_descriptor::kCTFontVerticalOrientation;
@@ -245,7 +246,7 @@ impl Rasterizer {
         font: &Font,
     ) -> Option<Result<RasterizedGlyph, Error>> {
         let scaled_size = self.device_pixel_ratio * glyph.size.as_f32_pts();
-        font.get_glyph(glyph.c, f64::from(scaled_size), self.use_thin_strokes)
+        font.get_glyph(glyph.c, f64::from(scaled_size), font.is_colored(), self.use_thin_strokes)
             .map(|r| Some(Ok(r)))
             .unwrap_or_else(|e| match e {
                 Error::MissingGlyph(_) => None,
@@ -431,6 +432,10 @@ impl Font {
         self.ct_font.symbolic_traits().is_italic()
     }
 
+    pub fn is_colored(&self) -> bool {
+        (self.ct_font.symbolic_traits() & kCTFontColorGlyphsTrait) != 0
+    }
+
     fn glyph_advance(&self, character: char) -> f64 {
         let index = self.glyph_index(character).unwrap();
 
@@ -450,6 +455,7 @@ impl Font {
         &self,
         character: char,
         _size: f64,
+        colored: bool,
         use_thin_strokes: bool,
     ) -> Result<RasterizedGlyph, Error> {
         let glyph_index =
@@ -471,6 +477,7 @@ impl Font {
                 height: 0,
                 top: 0,
                 left: 0,
+                colored: false,
                 buf: Vec::new(),
             });
         }
@@ -528,6 +535,7 @@ impl Font {
             top: (bounds.size.height + bounds.origin.y).ceil() as i32,
             width: rasterized_width as i32,
             height: rasterized_height as i32,
+            colored,
             buf,
         })
     }
@@ -583,7 +591,7 @@ mod tests {
         for font in fonts {
             // Get a glyph
             for c in &['a', 'b', 'c', 'd'] {
-                let glyph = font.get_glyph(*c, 72., false).unwrap();
+                let glyph = font.get_glyph(*c, 72., false, false).unwrap();
 
                 // Debug the glyph.. sigh
                 for row in 0..glyph.height {
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -246,7 +246,7 @@ impl Rasterizer {
         font: &Font,
     ) -> Option<Result<RasterizedGlyph, Error>> {
         let scaled_size = self.device_pixel_ratio * glyph.size.as_f32_pts();
-        font.get_glyph(glyph.c, f64::from(scaled_size), font.is_colored(), self.use_thin_strokes)
+        font.get_glyph(glyph.c, f64::from(scaled_size), self.use_thin_strokes)
             .map(|r| Some(Ok(r)))
             .unwrap_or_else(|e| match e {
                 Error::MissingGlyph(_) => None,
@@ -455,7 +455,6 @@ impl Font {
         &self,
         character: char,
         _size: f64,
-        colored: bool,
         use_thin_strokes: bool,
     ) -> Result<RasterizedGlyph, Error> {
         let glyph_index =
@@ -535,7 +534,7 @@ impl Font {
             top: (bounds.size.height + bounds.origin.y).ceil() as i32,
             width: rasterized_width as i32,
             height: rasterized_height as i32,
-            colored,
+            colored: self.is_colored(),
             buf,
         })
     }
@@ -591,7 +590,7 @@ mod tests {
         for font in fonts {
             // Get a glyph
             for c in &['a', 'b', 'c', 'd'] {
-                let glyph = font.get_glyph(*c, 72., false, false).unwrap();
+                let glyph = font.get_glyph(*c, 72., false).unwrap();
 
                 // Debug the glyph.. sigh
                 for row in 0..glyph.height {
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -42,10 +42,10 @@ use core_text::font_descriptor::{CTFontDescriptor, CTFontOrientation};
 
 use euclid::{Point2D, Rect, Size2D};
 
-use super::{FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph};
+use super::{BitmapBuffer, FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph};
 
 pub mod byte_order;
-use self::byte_order::extract_rgb;
+use self::byte_order;
 use self::byte_order::kCGBitmapByteOrder32Host;
 
 use super::Size;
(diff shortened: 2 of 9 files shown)


font/src/ft/mod.rs
```diff
@@ -37,6 +38,7 @@ struct Face {
     render_mode: freetype::RenderMode,
     lcd_filter: c_uint,
     non_scalable: Option<FixedSize>,
+    has_color: bool,
```
kchibisov: We should acknowledge the renderer that it should check `buf` part and scale colored bitmaps to fit into cells. I guess it should help on other platforms and fix an issue with rendering emojis with the right color.

font/src/ft/mod.rs
```diff
@@ -265,13 +267,35 @@ impl FreeTypeRasterizer {
...
+            let has_color = ft_face.has_color();
+            if has_color {
+                unsafe {
+                    let ft_face = ft_face.raw_mut();
+                    let best = 0;
+                    // XXX we could check sizes info, but we don't know to what size we're going to
+                    // downscale.
+                    //
+                    // let available_sizes =
+                    // std::slice::from_raw_parts::<freetype_sys::FT_Bitmap_Size>(
```
kchibisov: We could try and check bitmap sizes and pick the most close one to our current cell dimensions, however pass a size information here requires a lot of api changing.

font/src/ft/mod.rs
```diff
@@ -320,14 +344,27 @@ impl FreeTypeRasterizer {
...
+        }
+
+        let flags = if face.has_color {
+            freetype::face::LoadFlag::COLOR | freetype::face::LoadFlag::DEFAULT
+        } else {
+            face.load_flags
+        };
+
+        if let Err(res) = face.ft_face.load_glyph(index as u32, flags) {
+            println!("ERROR {:?}", res);
```
kchibisov: Will remove this println later on.

font/src/ft/mod.rs
```diff
@@ -484,6 +522,20 @@ impl FreeTypeRasterizer {
...
                 Ok((bitmap.rows(), bitmap.width(), packed))
             },
+            PixelMode::Bgra => {
+                let buf_size = (bitmap.rows() * bitmap.width() * 4) as usize;
+                let mut i = 0;
+                while i < buf_size {
+                    // Convert BGRA to RGB
+                    //
+                    // XXX our rendring works in rgb and doens't care about urers alpha.
+                    packed.push(buf[i + 2]);
```
kchibisov: I guess we should perform this conversion right here, the question is whether we want to keep alpha data, it's useless for alacritty, but we can drop it there and not here.

alacritty_terminal/src/renderer/mod.rs
```diff
@@ -297,6 +368,13 @@ impl GlyphCache {
             let mut rasterized =
                 rasterizer.get_glyph(glyph_key).unwrap_or_else(|_| Default::default());
 
+            // FIXME
+            // Assume that all emojis are 2 width for now.
+            let h = metrics.line_height as i32;
```
kchibisov: We can pass the WIDECHAR flag and just multiply it properly, I guess.

alacritty_terminal/src/renderer/mod.rs
```diff
@@ -273,6 +273,77 @@ impl GlyphCache {
         Ok((regular, bold, italic, bold_italic))
     }
 
+    // TODO: - take a look on what options we have besides this simple algo.
+    //
+    // area-averaging algorithm to downsample a colored bitmap.
+    fn downsample_bitmap(
```
kchibisov: Yes, we should do it ourselves. Enjoy.

alacritty_terminal/src/renderer/mod.rs
kchibisov: I mean don't use height for both things, this was just for showing what we need and where.

font/src/ft/mod.rs
kchibisov: I mean drop it in renderer.rs.

font/src/ft/mod.rs
kchibisov: It turned out that this has_color field could be pretty useful, so I guess we should keep it.

dnkl:
Not sure exactly how your fontconfig+freetype integration works, and I have to admit I haven't read through this PR that thoroughly. But, fontconfig will expose a `pixelsizefixupfactor` attribute that one can use to scale the bitmap.

This fixup factor depends on the point/pixel-size you requested:
```
% fc-match -v "Noto Color Emoji:size=12"|grep pixelsizefixupfactor
        pixelsizefixupfactor: 0,114679(f)(w)
% fc-match -v "Noto Color Emoji:size=24"|grep pixelsizefixupfactor
        pixelsizefixupfactor: 0,229358(f)(w)
```

kchibisov:
@dnkl The main thing is that you should scale things yourself anyway. I mean fontcofnig can calculate a scaling factor for you, but It wont scale the bitmap, right? In this case I don't know how it'll differ from what I'm doing here. Also the user can have disabled this scaling thing and it'll blow alacritty if we will rely on this approach.

dnkl:
@kchibisov 
> The main thing is that you should scale things yourself anyway. I mean fontcofnig can calculate a scaling factor for you, but It wont scale the bitmap, right?

Yes, you need to scale it yourself. fontconfig only tells you by _how much_.

>  In this case I don't know how it'll differ from what I'm doing here.

The difference is fontconfig uses the font and the requested size to calculate the scaling factor (possibly also weighing in other attributes of the font), while you base the scaling factor on the cell size.

Yours will of course ensure the glyph isn't rendered outside its cell(s). But using the fontconfig provided scaling factor may be more in line with how you handle regular glyphs loaded from fallback fonts (or do you manually re-scale those too?)

But, also consider this; while the typical emoji font has glyphs with height == width, there's nothing that prevents you from creating a color bitmap font where this isn't true.

Furthermore, even if emojis visually are roughly two cells, not all are according to `wcwidth()`. Meaning from a typical clients perspective, you'll still be rendering (some) glyphs outside the cell boundaries.

All that said:
> Also the user can have disabled this scaling thing and it'll blow alacritty if we rely on this approach.

Yes. I was a bit surprised by this. But it isn't really that strange; there isn't really anything special with color bitmap fonts except that they are, well, in color. But they are in many ways "just" bitmap fonts.

It might make sense to check how other layout/rendering engines handle this (cairo, for example).

> Now I'm downscaling the bitmap to make it fit into our renderer cell, so there it no way it can overlap the other content.

Only partially true, I think. See comment about `wcwidth()` above.

> P.s. If you have a reference on how can I force freetype/fontconfig do this scaling for me it'll be appreciated, but I can't find anything related to this.

Not that I'm aware of. With many 2D rendering libs all you need to do is set a scale transform and they'll do it for you...

> P.P.s. My implementation is freetype only and don't rely on fontconfig besides font loading.

That's how it supposed to work :)

I think this comes down to a design decision where you either let the font system decide, or you override it and manually fit the glyphs.

I think what I'll do in my projects is use the fontconfig provided scaling factor if available, and otherwise estimate it using the primary font's line height.

font/src/ft/mod.rs
```diff
@@ -320,14 +344,25 @@ impl FreeTypeRasterizer {
...
 
-        face.ft_face.load_glyph(index as u32, face.load_flags)?;
+        let flags = if face.has_color {
+            freetype::face::LoadFlag::COLOR | freetype::face::LoadFlag::DEFAULT
+        } else {
+            face.load_flags
+        };
+
+        face.ft_face.load_glyph(index as u32, flags)?;
+
```
dnkl: I'm fairly sure you don't need to special case the LCD filter and load flags. AFAIK they only apply to outline fonts. Case in point: regular bitmap fonts already work just fine in Alacritty.

kchibisov:
As you can see this PR is a WIP, however I've mentioned that the implementation affects a lot of code and I want to agree on how we should handle its integration in a way that doesn't change our rendering to be Linux/BSD centric. We've already gathered info about adding `colored` field to `RasterizedGlyph`, the result was that it makes sense and will solve coloring problem. However the downsampling part is still untouched and 1/2 width chars handling was mentioned but not yet implemented, because it's too much of changes to other platforms structure. Anyway @dnkl 

> The difference is fontconfig uses the font and the requested size to calculate the scaling factor (possibly also weighing in other attributes of the font), while you base the scaling factor on the cell size.

Yes, and you wont handle 1 cell emojis well. gnome-terminal can't do this, I've tried in alacritty and also wasn't able to do this. If you now how to get a proper scaling for 1 cell thing from `fontconfig`, please share it with me. I don't want to send information to our rasterizer about how much cells a glyph occupies, since this information is not even needed on other platforms and will be more like a hack then a solution, but it's just me.

> But, also consider this; while the typical emoji font has glyphs with height == width, there's nothing that prevents you from creating a color bitmap font where this isn't true.

This is not true, I know. Noto don't have width == height. We can pass a scaling factor from Rasterizer to renderer, but we don't need it on other platforms, and we can't scale in font crate because of we don't have information about how much cells we need for chars.

> Furthermore, even if emojis visually are roughly two cells, not all are according to wcwidth(). Meaning from a typical clients perspective, you'll still be rendering (some) glyphs outside the cell boundaries.

Yes this is your future with naive use of fontconfig scale factor, this doesn't apply to my approach, so 1/2 width will be displayed without overlapping.

> Only partially true, I think. See comment about wcwidth() above

This is only for the current code, but I've already mentioned in code and in a review that we should do it and scale into a box `wcwidth * cell_width x height`.

>>    P.s. If you have a reference on how can I force freetype/fontconfig do this scaling for me it'll be appreciated, but I can't find anything related to this.

>Not that I'm aware of. With many 2D rendering libs all you need to do is set a scale transform and they'll do it for you...

And I guess they're not using have a thing like wcwidth (e.g. firefox doesn't worry about it).

>I think this comes down to a design decision where you either let the font system decide, or you override it and manually fit the glyphs.

I don't want to see glyphs overlapping each other, If I see it I'll be mad.

Anyway we can implement `fontconfig` based scaling and perform `downsampling` in our font crate if we pass an information to font crate about `wcwidth`. 

To sum up we have 3 approaches.

1 -> pass information about wcwidth to font crate (our Rasterizer) and scale recompute fontconfig scale.
2 -> Scale into `wcwidth * cell_width x height` box in renderer.
3 -> pass fontconfig scaling to a renderer.

Which one to pick I'm leaving to @chrisduerr and @nixpulvis . I'm voting for the 2nd one.

font/src/ft/mod.rs
kchibisov: If you want to render your colored emojis in monochrome variant then go ahead and remove load color. I'll investigate a bit more on what we can do here. I don't know how LCD filter affects, but I've noticed that some implementations disable it for colored glyphs.

dnkl: I know `FT_LOAD_COLOR` is required for color bitmap fonts. That's not what I meant.

kchibisov: Do you mean about that I don't need to manually alter them here? If this is what you wanted to say then you're right.

dnkl:
> This is not true, I know. Noto don't have width == height. We can pass a scaling factor from Rasterizer to renderer, but we don't need it on other platforms, and we can't scale in font crate because of we don't have information about how much cells we need for chars.
> ...
> This is only for the current code, but I've already mentioned in code and in a review that we should do it and scale into a box wcwidth * cell_width x height.

You'll get some very funny looking emojis if you do that. What I was trying to say is that while an emoji glyph may be 2 cells **visually**, `wcwidth()` may return 1. If you scale that according to `wcwidth()`, you'll either get a squeezed up glyph, or a very small glyph, depending on how you scale. But yes, it wont overlap.

This is probably wcwidth implementations lagging behind the unicode standard, or something like that. Those emoji glyphs probably _should_ be two columns.

> And I guess they're not using have a thing like wcwidth (e.g. firefox doesn't worry about it).

Non terminal applications typically don't need `wcwidth()`; they advance by _actual_ glyph width instead.

font/src/ft/mod.rs
dnkl: From my own experience, there's no need to differentiate load flags **at all**. From what I have seen, you can _always_ include `FT_LOAD__COLOR`, even for non-color bitmap fonts. And for color bitmap fonts it appears safe to include the "normal" load flags too.

I'm afraid the documentation of `FT_LOAD_COLOR` doesn't really back this up. But neither does it say anything against it.

But this is just from my own experience. I know others only add `FT_LOAD_COLOR` for color bitmap fonts.

I'm for simplifying code and avoiding special casing where it isn't needed. This appears to be one such case, but I don't have any hard evidence.

chrisduerr:
> If you scale that according to wcwidth(), you'll either get a squeezed up glyph, or a very small glyph, depending on how you scale. But yes, it wont overlap.

Alternatively, it should overlap. If the glyph is designed at double-width but the `wcwidth` is only a single cell, we should render it as overlapping. It's not our job to fix fonts.

> This is probably wcwidth implementations lagging behind the unicode standard, or something like that. Those emoji glyphs probably should be two columns.

I don't think that's the case, since Alacritty doesn't use `wcwidth`, but a Rust crate called `unicode-width`. I'm pretty sure this is intended, as all other emojis were updated to double-width just a short while ago and those work as intended.

> Non terminal applications typically don't need wcwidth(); they advance by actual glyph width instead.

Even Firefox will need it, for stuff like `<code>` blocks. However, they should use `wcswidth`, not `wcwidth`.

dnkl:
> I don't think that's the case, since Alacritty doesn't use wcwidth, but a Rust crate called unicode-width. I'm pretty sure this is intended, as all other emojis were updated to double-width just a short while ago and those work as intended.

That may very well be the case. I've been using libc's wcwidth, which I'm guessing more easily lags behind. I suspect these are issues that will fix themselves over time.

kchibisov:
Ok, then lets overlap things.

dnkl:
If you _do_ decide to use the fontconfig provided scale factor in the final design, you'll want to have a fallback when the user has disabled bitmap font scaling (i.e. removed `/etc/fonts/conf.d/10-scale-bitmap-fonts.conf`).

One can use the requested (point- or pixel) size divided by the actual size. This is how fontconfig itself does (by default, that is - the user can configure things differently).

One interesting point is that fontconfig considers regular bitmap fonts non-scalable, while _color_ bitmap fonts are scalable (the fontconfig attribute `scalable` is `true`). But it still omits `pixelsizefixupfactor` for color bitmap fonts (when disabled by the user)...

Finally, my own reasoning for using the `pixelsizefixupfactor` was to **a)** avoid assumptions - treat emoji fonts like any other font as much as possible (in particular, don't assume a color bitmap font is an emoji font), and **b)** don't override the user's settings.

You may disagree, and that may lead you to a different design. The perhaps biggest argument _against_ using fontconfig for scaling is that it _will_ result in glyphs overlapping cells for certain font combinations. But you have that problem with **all other** fallback glyphs too.

chrisduerr:
> avoid assumptions - treat emoji fonts like any other font as much as possible (in particular, don't assume a color bitmap font is an emoji font)

I think this is the biggest reason honestly. We don't want to do any magic ourselves, the behavior is usually well defined and where definition might be lacking, it's often de-facto standardized by existing implementations.

> The perhaps biggest argument against using fontconfig for scaling is that it will result in glyphs overlapping cells for certain font combinations

Now this is an argument I think many people can originally easily get behind. Fonts not overlapping sounds great, right? But I think we would be setting a bad precedent if we did this.

Generally we should look at most emoji fonts the same way we look at proportional fonts. These are fonts, designed to disregard our monospace restrictions and just use the glyph's advance to define cell width. So should we scale text glyphs of proportional fonts until they fit into our cells? That would look terrible and I don't think anyone would be happy with these tiny fonts that do not fit in.

So fundamentally, if a font provides a glyph for a symbol that is only a single cell wide, but it provides that glyph in a factor where it cannot be rendered in a single cell, then we're dealing with an emoji font that is not compatible with monospace fonts. And because of this, we should not try to hack around this, but rather stick to what we're doing right now with all other proportional fonts and explicitly not support it.

Ultimately, all monospace compatible emoji fonts should properly scale their emojis at the bitmap scaling factor, just the way all other single cell bitmap glyphs do.

> If you do decide to use the fontconfig provided scale factor in the final design, you'll want to have a fallback when the user has disabled bitmap font scaling

Which factor we use for this is another really good question. You've mentioned scaling requested size and actual size, the original version of this PR scaled the glyph to fit inside the width of the cell. Another option would be to go based off of height of the glyph and the cell of course, which is similar to the width but not exactly the same.

I'm not sure myself what the best idea here is, but most seem like valid strategies. It would be interesting to see what common other approaches do in this case and how other terminals might behave with bitmap scaling disabled.

dnkl:
> Which factor we use for this is another really good question. You've mentioned scaling requested size and actual size, the original version of this PR scaled the glyph to fit inside the width of the cell. Another option would be to go based off of height of the glyph and the cell of course, which is similar to the width but not exactly the same.

I actually tried all three.

Scaling glyph height to cell height looked a bit out of place; too tall, and at least Noto Color Emoji overlapped horizontally due to the glyphs being slightly wider than they're tall.

Scaling with wcwidth will result in slightly smaller glyphs than when scaled with `pixelsizefixupfactor`, but still looks good (as long as wcwidth returns the correct column count - which it appears your rust implementation does).

If you decide to use requested size divided by actual size, keep in mind that it is the font's _nominal_ height both sizes refer to, and not glyph image height. These are the `size` and `pixelsize` values returned by fontconfig. The _requested_ size is the size from the input pattern, and the _actual_ size is the size from the final pattern.

kchibisov:
@dnkl Could you take a look on the latest approach I've pushed?

font/src/ft/mod.rs
```diff
@@ -367,7 +367,13 @@ impl FreeTypeRasterizer {
         };
 
         if face.has_color {
-            Ok(downsample_bitmap(bitmap, face.pixelsize_fixup_factor))
+            let fixup_factor = if face.pixelsize_fixup_factor == 0. {
+                // Fallback
+                self.pixel_size.ceil() as f64 / bitmap.width as f64
+            } else {
```
dnkl: This is the final `pixelsize` value divided by the glyph image width, right? That doesn't look quite right. The pixelsize refers to a font's _nominal height_. This is a) not the same as font's line height (the nominal height is smaller) and b) why the width?

I _think_ that you either want to use the primary font's line height (the cell height, that is) divided by the glyph image height. Or, use the nominal point size configured by the user (the `size`  entry in `alacritty.yml`) divided by the `size` attribute you got from fontconfig for this particular font.

Example:
```
 fc-match -v "Noto Color Emoji:size=12"|grep size
Pattern has 43 elts (size 48)
        size: 104,64(f)(w)
        pixelsize: 109(f)(s)
        pixelsizefixupfactor: 0,114679(f)(w)
```
Where `12 / 104.64 == 0,1146`.

dnkl: In case it helps: https://codeberg.org/dnkl/foot/src/branch/master/font.c#L160

kchibisov: >This is the final pixelsize value divided by the glyph image width, right? That doesn't look quite right. The pixelsize refers to a font's nominal height. This is a) not the same as font's line height (the nominal height is smaller) and b) why the width?

@dnkl I'm using the pixelsize thing from a user's config, so I guess I should change bitmap width to pixelsize for a font.

The use of width is because of fixupfactor seems like is for width and not height, so you should calculate height fixup yourself. I haven't found any information about on what I should apply the fixup, however applying it to height doesn't scale things well according to my testing.

dnkl: > I'm using the pixelsize thing from a user's config, so I guess I should change bitmap width to pixelsize for a font.

Isn't the user config in points, not pixels? Anyway, if you want to use the glyph size (and not the emoji font's nominal height), then you should scale using the cell height. But as mentioned, this will lead to too-large-looking emojis. Actually, thinking about it, I guess you could use the nominal height too, but I'm guessing the resulting glyphs will look a bit small?

> The use of width is because of fixupfactor seems like is for width and not height, so you should calculate height fixup yourself. I haven't found any information about on what I should apply the fixup, however applying it to height doesn't scale things well according to my testing.

No, fontconfig uses fonts' **nominal height** when calculating the fixup factor (see example above), and that factor should be applied both to the glyph's height and width (otherwise you'll get disproportional glyphs). In general, fontconfig doesn't care about widths, since most applications render proportional fonts where each glyph has a different width.

kchibisov: @dnkl Thanks for clarification, now the fallback returns the exact same value as a fixup.

font/src/ft/mod.rs
```diff
@@ -529,6 +589,86 @@ impl FreeTypeRasterizer {
...
+    let b_buf = &bitmap_glyph.buf;
+    let scaling_factor =
+        (bitmap_width as f32 / width as f32).max(bitmap_height as f32 / height as f32);
+    let advance_step = scaling_factor.ceil() as usize;
+    let mut scaled_buffer = Vec::with_capacity(width * height * 3);
+
+    let mut new_line_index = 0;
+    let mut source_line_index = 0;
+
+    while new_line_index < height {
```
```suggestion
    for i in 0..height {
```

font/src/ft/mod.rs
```diff
@@ -529,6 +589,86 @@ impl FreeTypeRasterizer {
...
+        let mut new_column_index = 0;
+        let mut source_column_index = 0;
+
+        while new_column_index < width {
```
```suggestion
        for i in 0..width {
```

font/src/ft/mod.rs
```diff
@@ -529,6 +589,86 @@ impl FreeTypeRasterizer {
...
+            let mut r: u32 = 0;
+            let mut g: u32 = 0;
+            let mut b: u32 = 0;
+            let mut pixels_picked: u32 = 0;
+
+            let source_end_line = std::cmp::min(source_line_index + advance_step, bitmap_height);
+            let source_end_column = std::cmp::min(source_column_index + advance_step, bitmap_width);
```
chrisduerr: Usually it's fine to import even `std::cmp::min` at the top, that makes these super simple functions a little easier. `std::cmp` would be fine too, so we'd use `cmp::min`, but since it's used so often in some modules the `min` function is basically a reserved name anyways, so it shouldn't be confusing to anyone.

font/src/ft/mod.rs
```diff
@@ -529,6 +589,86 @@ impl FreeTypeRasterizer {
...
+
+            let mut source_line_index = source_line_index;
+            while source_line_index < source_end_line {
```
```suggestion
            for i in source_line_index..source_end_line {
```

font/src/ft/mod.rs
```diff
@@ -529,6 +589,86 @@ impl FreeTypeRasterizer {
...
+                let cur_pixel_index = source_line_index * bitmap_width * 3;
+
+                let mut source_column_index = source_column_index;
+                while source_column_index < source_end_column {
```
```suggestion
                for i in source_column_index..source_end_column {
```

We definitely do not want to shadow `source_column_index` and `source_line_index`. Especially in the way it is right now, the `source_column_index += 1` could easily explode right in our face.

font/src/ft/mod.rs
```diff
@@ -529,6 +589,86 @@ impl FreeTypeRasterizer {
...
+                    r += b_buf[cur_pixel_index + source_column_index * 3] as u32;
+                    g += b_buf[cur_pixel_index + source_column_index * 3 + 1] as u32;
+                    b += b_buf[cur_pixel_index + source_column_index * 3 + 2] as u32;
+                    source_column_index += 1;
+                    pixels_picked += 1;
+                }
+                source_line_index += 1;
+            }
+
+            if pixels_picked == 0 {
```
chrisduerr: Does this ever actually happen? That would mean that the bitmap is `0x0` in dimensions, right?

Might be useful to just assert that this is bigger than 0?

font/src/ft/mod.rs
```diff
@@ -529,6 +589,86 @@ impl FreeTypeRasterizer {
...
+            }
+
+            source_column_index += advance_step;
+            new_column_index += 1;
+        }
+        source_line_index += advance_step;
+        new_line_index += 1;
+    }
+
+    // This top computation performs better with a scaling algo we use.
```
```suggestion
    // This top computation performs better with the scaling algorithm we use.
```

chrisduerr: I'd also like to know *how* it performs better. I'd assume this is not about performance?

font/src/ft/mod.rs
```diff
@@ -529,6 +589,86 @@ impl FreeTypeRasterizer {
...
+    bitmap_glyph.top = ((bitmap_glyph.top as f32 * fixup_factor as f32) as i32
+        + bitmap_glyph.top / advance_step as i32)
+        / 2;
```
chrisduerr: Where exactly does this formula come from? Is this just using metrics based on what's reasonable, or is there any source recommending this strategy?

font/src/ft/mod.rs
```diff
@@ -265,13 +272,24 @@ impl FreeTypeRasterizer {
...
             };
 
+            let pixelsize_fixup_factor = pattern.pixelsizefixupfactor().next().unwrap_or(0.);
+
+            let has_color = ft_face.has_color();
+            if has_color {
+                unsafe {
+                    freetype_sys::FT_Select_Size(ft_face.raw_mut(), 0);
+                }
+            }
```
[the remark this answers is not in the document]
kchibisov: thats select the sizes for bitmap.

font/src/ft/mod.rs
```diff
@@ -320,34 +338,54 @@ impl FreeTypeRasterizer {
                 glyph_key.size.as_f32_pts() * self.device_pixel_ratio * 96. / 72.
             });
 
-        face.ft_face.set_char_size(to_freetype_26_6(size), 0, 0, 0)?;
+        if !face.has_color {
+            face.ft_face.set_char_size(to_freetype_26_6(size), 0, 0, 0)?;
+        }
```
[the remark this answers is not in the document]
kchibisov: Error.

font/src/ft/mod.rs
```diff
@@ -503,6 +560,9 @@ impl FreeTypeRasterizer {
                         // load it again.
                         Some(&key) => {
                             debug!("Hit for font {:?}; no need to load", path);
+                            // Update fixup factor
+                            self.faces.get_mut(&key).unwrap().pixelsize_fixup_factor =
+                                pattern.pixelsizefixupfactor().next().unwrap_or(0.0);
```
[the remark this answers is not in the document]
kchibisov: You can change font size and it'll break things.

font/src/ft/mod.rs
```diff
@@ -529,6 +589,86 @@ impl FreeTypeRasterizer {
...
+
+    let width = (bitmap_width * fixup_factor) as usize;
+    let height = (bitmap_height * fixup_factor) as usize;
+
+    let bitmap_width = bitmap_width as usize;
+    let bitmap_height = bitmap_height as usize;
+
+    let b_buf = &bitmap_glyph.buf;
+    let scaling_factor =
+        (bitmap_width as f32 / width as f32).max(bitmap_height as f32 / height as f32);
```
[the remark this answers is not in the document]
kchibisov: That's a batch for scaling. nothing to do with a fixup.

font/src/ft/mod.rs
kchibisov: How it could explode?

font/src/ft/mod.rs
kchibisov: It does happen mostly on every downsampling iirc. The thing is that you should scaling to some rect, but you can finish your scaling quite fast, and you should fill the bitmap with garbage.

font/src/ft/mod.rs
kchibisov: No, it's about a glyph position. If you just multiply it by fixup_factor it'll be too tight to the top imho.

font/src/ft/mod.rs
kchibisov: This formula came from hours of testing and comparing with different font sizes etc. You can try to compare it to just fixup * top.

font/src/ft/mod.rs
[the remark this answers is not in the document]
chrisduerr: But how do we change the font size without reloading the font?

I don't think there's currently any way, unless I'm missing things.

font/src/ft/mod.rs
chrisduerr: Shadowing will cause problems when you refactor code, say you want to change to my `for` loop, so you delete the original `let mut variable` statement. The compiler will not warn  you about removing the `variable += 1`, since the previous variable was shadowed, thus causing a bug that could have easily been prevented.

So in this case variable shadowing is a really bad idea, because by moving around or deleting lines the code easily breaks in hard to understand ways.

If you do things like shadow variables of a different type like `let path = path.to_string()`, then you don't run into these problems, due to help from the typesystem.

font/src/ft/mod.rs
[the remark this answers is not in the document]
kchibisov: You're right about reloading, but I guess you're thinking only about alacritty.yml fonts, and this will be fine for them, but we also have a font fallbacks, so to handle them properly you either need to load an entire font during fallback and keep it around (I like this idea though) or do the same thing as I do here. The only way to get `fixup_factor`  is from `font_match` resulted pattern. So you have 3 places in `ft/mod.rs` 2 of them are about default config and 1 for fallback, so you should perform the update in a fallback part.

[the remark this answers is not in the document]
chrisduerr: Okay, here's my understanding of the problem, let me know if that is correct:

 - We load our default font into `self.faces`
 - Due to a missing emoji glyph, we load a fallback font into `self.faces`
 - We resize our default font, effectively discarding it from `self.faces`
    -> To be completely correct, I don't think we ever actually discard the old face? I think we just insert the new one and the old one is just never accessed anymore
 - We fallback to the same emoji font again, but it has not been 'overwritten' and the same font matches both our old size pattern as the new one, thus getting loaded from `self.faces` with incorrect fixup factor

So my initial thought would almost be that we might want to clear `self.faces` whenever a new default font is loaded? However that does not seem ideal for performance, but it will happen much less than font fallback in general.

I'd be curious what you think, though generally I feel like this should be our safest choice from a performance standpoint, since the operation is basically a noop with how cold it is.

font/src/ft/mod.rs
```diff
@@ -529,6 +581,84 @@ impl FreeTypeRasterizer {
...
+    // Don't try to upscale
+    if fixup_factor >= 1.0 {
+        return bitmap_glyph;
+    }
+
+    let bitmap_buffer = if let BitmapBuffer::RGBA(bitmap_buffer) = bitmap_glyph.buf {
+        bitmap_buffer
+    } else {
+        return bitmap_glyph;
+    };
```
chrisduerr: I think the following code is pretty easy to understand and has a little less branching you need to think about:

```rust
    // Only scale colored buffers which are bigger than required
    match (bitmap_glyph.buf, fixup_factor.cmp(&1.0)) {
        (BitmapBuffer::RGBA(buffer), Ordering::Less) => buffer,
        _ => return bitmap_glyph,
    }
```

font/src/ft/mod.rs
```diff
@@ -529,6 +581,84 @@ impl FreeTypeRasterizer {
...
+
+    let bitmap_width = bitmap_glyph.width as usize;
+    let bitmap_height = bitmap_glyph.height as usize;
+
+    let width = (bitmap_width as f64 * fixup_factor) as usize;
+    let height = (bitmap_height as f64 * fixup_factor) as usize;
+
+    let advance_step = (1.0 / fixup_factor).ceil() as usize;
```
chrisduerr: I'm not sure if I love `advance_step` as a variable name. It's not really bad, but I feel like it might not be completely obvious.

Ideally this variable name would encompass that this is the height and width of the source grid. Do you have any ideas?

font/src/ft/mod.rs
```diff
@@ -529,6 +581,84 @@ impl FreeTypeRasterizer {
...
+    let mut scaled_buffer = Vec::<u8>::with_capacity(width * height * 4);
```
chrisduerr: The method is called `downsample_bitmap`, so a good name for the output variable would be `downsampled_buffer` I think, instead of `scaled_`.

font/src/ft/mod.rs
```diff
@@ -529,6 +581,84 @@ impl FreeTypeRasterizer {
...
+        bitmap_buffer
+    } else {
+        return bitmap_glyph;
+    };
+
+    let bitmap_width = bitmap_glyph.width as usize;
+    let bitmap_height = bitmap_glyph.height as usize;
+
+    let width = (bitmap_width as f64 * fixup_factor) as usize;
+    let height = (bitmap_height as f64 * fixup_factor) as usize;
```
chrisduerr: Calling this `target_width` and `target_height` might make things a lot more clear. Same with `bitmap_width` and `bitmap_height`. Especially since both input and output are bitmaps.

font/src/ft/mod.rs
```diff
@@ -529,6 +581,84 @@ impl FreeTypeRasterizer {
...
+    let mut source_line_index = 0;
+
+    for _ in 0..height {
+        let mut source_column_index = 0;
+
+        for _ in 0..width {
+            let mut r: u32 = 0;
+            let mut g: u32 = 0;
+            let mut b: u32 = 0;
+            let mut a: u32 = 0;
```
chrisduerr: Usually this isn't ideal, but in this specific case I think the following notation makes sense:

```rust
let (mut r, mut g, mut b, mut a) = (0, 0, 0, 0);
```

font/src/ft/mod.rs
```diff
@@ -529,6 +581,84 @@ impl FreeTypeRasterizer {
...
+    let width = (bitmap_width as f64 * fixup_factor) as usize;
+    let height = (bitmap_height as f64 * fixup_factor) as usize;
+
+    let advance_step = (1.0 / fixup_factor).ceil() as usize;
+    let mut scaled_buffer = Vec::<u8>::with_capacity(width * height * 4);
+
+    let mut source_line_index = 0;
+
+    for _ in 0..height {
+        let mut source_column_index = 0;
```
chrisduerr: I think I've mentioned this in the previous review, but `source_start_line` and `source_start_column` would make a lot more sense here in combination with the `source_end_line` and `source_end_column` variables.

font/src/ft/mod.rs
```diff
@@ -529,6 +581,84 @@ impl FreeTypeRasterizer {
...
+            let mut a: u32 = 0;
+            let mut pixels_picked: u32 = 0;
+
+            let source_end_line = min(source_line_index + advance_step, bitmap_height);
+            let source_end_column = min(source_column_index + advance_step, bitmap_width);
+
+            for source_j in source_line_index..source_end_line {
+                let cur_pixel_index = source_j * bitmap_width * 4;
+
+                for source_i in source_column_index..source_end_column {
```
chrisduerr: The variable names `source_j` and `source_i` are useless. Naming them `source_line` and `source_column` will be much, much, much more useful for the reader.

font/src/ft/mod.rs
```diff
@@ -529,6 +581,84 @@ impl FreeTypeRasterizer {
...
+                    r += bitmap_buffer[cur_pixel_index + source_i * 4] as u32;
+                    g += bitmap_buffer[cur_pixel_index + source_i * 4 + 1] as u32;
+                    b += bitmap_buffer[cur_pixel_index + source_i * 4 + 2] as u32;
+                    a += bitmap_buffer[cur_pixel_index + source_i * 4 + 3] as u32;
```
chrisduerr: I think I've mentioned this in the previous PR too, but I think logically it makes more sense to do `(source_pixel_index + source_column) * 4 + 3`. So the `*4` can be removed from the pixel index variable.

font/src/ft/mod.rs
```diff
@@ -529,6 +581,84 @@ impl FreeTypeRasterizer {
...
+            let mut g: u32 = 0;
+            let mut b: u32 = 0;
+            let mut a: u32 = 0;
+            let mut pixels_picked: u32 = 0;
+
+            let source_end_line = min(source_line_index + advance_step, bitmap_height);
+            let source_end_column = min(source_column_index + advance_step, bitmap_width);
+
+            for source_j in source_line_index..source_end_line {
+                let cur_pixel_index = source_j * bitmap_width * 4;
```
chrisduerr: I think `source_pixel_index` says much more about this variable than `cur_pixel_index`.

font/src/ft/mod.rs
chrisduerr: I'm not fond of the idea of just testing out random combinations of font metrics until something seems to work out okay, this usually means that the result is over-specialized for the fonts installed when testing, rather than being a good general purpose solution.

We should try to see either what other implementations do, or if there's any reference on how these metrics should be scaled. Otherwise we need to put some reason behind the computation, that is more than just 'it seems to look okay this way'.

font/src/lib.rs
```diff
@@ -220,20 +220,25 @@ pub struct RasterizedGlyph {
...
 
-impl Default for RasterizedGlyph {
-    fn default() -> RasterizedGlyph {
-        RasterizedGlyph { c: ' ', width: 0, height: 0, top: 0, left: 0, buf: Vec::new() }
-    }
+#[derive(Clone, Debug)]
+pub enum BitmapBuffer {
+    RGB(Vec<u8>),
+    RGBA(Vec<u8>),
 }
```
chrisduerr: I'm not sure if I've asked you this on IRC, but have you considered using `Vec<PixelFormat>` instead of `PixelFormat<Vec<u8>>`? That way each individual pixel will have the information attached to it.

alacritty/src/renderer/mod.rs
```diff
@@ -1573,22 +1583,42 @@ impl Atlas {
...
+                        offset_y,
+                        width,
+                        height,
+                        gl::RGBA,
+                        gl::UNSIGNED_BYTE,
+                        buf.as_ptr() as *const _,
+                    );
+                    colored = true;
+                },
+            }
```
chrisduerr: Github didn't let me add any more comments to my existing PR because you pushed a commit I think, so I've added this separately:

I think we could remove duplication a lot by changing to something more like this:

```rust
let (format, buf) = match glyph.buf {
    RGB(buf) => (gl::RGB, buf),
    RGBA(buf) => (gl::RGBA, buf),
};
gl::TexSubImage2D(format, buffer);
```

font/src/ft/mod.rs
kchibisov: It's not a start it's an index. Since you can't change start, but you can change index. So, I'm not sure about it.

chrisduerr: Of course it's a start. It's the start of the source line range.

font/src/lib.rs
kchibisov: It wasn't discussed on IRC, but I've chosen this format for performance reasons and simplicity, since you need to pass these things to properly pass them to OpenGL renderer as a image pixels. So using Vec<PixelFormat> isn't playing well here.

chrisduerr: Oh right, the buffer is passed directly to OpenGL, specifying just the format using `gl::RGB` or `gl::RGBA`. I saw that at the end of my review, makes a lot of sense then.

font/src/ft/mod.rs
dnkl: FWIW, I simply scale it and `ceil` it:
```c
ceil(font->face->glyph->bitmap_top * font->pixel_size_fixup)
```
This scaled top value is then used as offset from the primary font's baseline.

It looks reasonable to me, with the (granted, few) font combinations I've tried: grim-snapshot-cca14db4-469d-47d3-952b-c4f09a52a599

(_Fantasque Sans Mono_ + _JoyPixels_)

font/src/ft/mod.rs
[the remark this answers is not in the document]
kchibisov: You're correct here. However I don't like the idea of dropping things and reloading them once again in the current state.

The main problem with this approach is that we should provide a new API to clear Rasterizer "cache" and it doesn't make much sense with the current font crate, since you'll "pseudo fallback" anyway most of the time. However if we do our fallbacks without using fontconfig api all the time (Like load fonts once and don't call it unless we can't find anything in already loaded fonts) that'll make much more sense and improve overall performance, since now e.g. unicode-random-write is a fontcofnig (`fc-match`) benchmark, which is in the same time funny and not.

Anyway this rework is out of scope of "Add colored emoji PR" in my opinion.

lovesegfault:
The issue with emoji selection exists before this PR, here's my test on 0.4.0

image
image

kchibisov:
@lovesegfault Yeah, we're aware of it. it's #791.

chrisduerr:
I'm actually a bit surprised about this, because I can't reproduce it myself.

nixpulvis:
Zoom in.

lovesegfault:
@chrisduerr Sometimes it works correctly for me too
image
image

alacritty/src/renderer/mod.rs
```diff
@@ -462,11 +465,25 @@ impl Batch {
...
 
+        if glyph.colored {
+            // XXX Temporary workaround to prevent emojis being rendered with a wrong colors on, at
+            // least, dark backgrounds. To resolve this issue properly we should have a separate
+            // rendering pass for colored glyphs. For more info see #1864.
+            //
+            // The 255 is not a color, it'll be a part of forming an alpha mask later on. So a
+            // proper color of emojis could be achieved by setting fg to #ffffff and bg
+            // to #000000, however we can't change bg here, because it actually "changes
+            // bg" of a cell.
```
chrisduerr: There's a lot of complexity to this issue, so I don't feel like going into detail here brings us much benefit and might confuse people rather than help as documentation.

I think just referring to #1864 should be sufficient, then we can document things there and people can read up on the complete extensive documentation, rather than making assumptions based on incomplete data.

alacritty/src/renderer/mod.rs
```diff
@@ -1573,21 +1591,27 @@ impl Atlas {
...
+        let colored;
 
         unsafe {
             gl::BindTexture(gl::TEXTURE_2D, self.id);
 
             // Load data into OpenGL
+            let (format, buf, _) = match &glyph.buf {
+                BitmapBuffer::RGB(buf) => (gl::RGB, buf, colored = false),
+                BitmapBuffer::RGBA(buf) => (gl::RGBA, buf, colored = true),
+            };
```
chrisduerr: This is an interesting idea, however it took me like three reads to understand what's actually going on here.

Could we just simplify this to not use tuples to assign things please?

Having small codeblocks like this should be completely sufficient:
```rust
{
    color = false;
    (gl::RGB, buf)
}
```

font/src/darwin/mod.rs
```diff
@@ -564,6 +572,9 @@ impl Font {
 
 #[cfg(test)]
 mod tests {
+
```
chrisduerr:
```suggestion
```

font/src/ft/mod.rs
```diff
@@ -182,14 +190,16 @@ impl FreeTypeRasterizer {
         // Adjust for DPI
         let size = Size::new(size.as_f32_pts() * self.device_pixel_ratio * 96. / 72.);
```
```suggestion
```

font/src/ft/mod.rs
```diff
@@ -265,13 +269,25 @@ impl FreeTypeRasterizer {
                 Some(FixedSize { pixelsize: pixelsize.next().expect("has 1+ pixelsize") })
             };
 
+            let pixelsize_fixup_factor = pattern.pixelsizefixupfactor().next().unwrap_or(0.);
```
chrisduerr: When is this going to be `None`, is this the case when the user has disabled bitmap scaling?

Do we really want to unwrap to `0` here? I feel like it should be an `Option` if it's potentially missing.

We later check for `pixelsize_fixup_factor == 0.` too, checking with `match` would be nicer.

font/src/ft/mod.rs
```diff
@@ -529,6 +582,93 @@ impl FreeTypeRasterizer {
...
+        source_line_start += line_step;
+    }
+
+    bitmap_glyph.top = (bitmap_glyph.top as f32 * fixup_factor as f32) as i32;
+    bitmap_glyph.left = (bitmap_glyph.left as f64 * fixup_factor) as i32;
+    bitmap_glyph.width = target_width as i32;
+    bitmap_glyph.height = target_height as i32;
+    bitmap_glyph.buf = BitmapBuffer::RGBA(downsampled_buffer);
+    bitmap_glyph
+}
```
chrisduerr: I've played around with this myself and I think the following implementation is identical but a bit cleaner:

```diff
diff --git a/font/src/ft/mod.rs b/font/src/ft/mod.rs
index b39607c..830895f 100644
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -596,43 +596,28 @@ fn downsample_bitmap(mut bitmap_glyph: RasterizedGlyph, fixup_factor: f64) -> Ra
     let target_height = (bitmap_height as f64 * fixup_factor) as usize;
 
     let downsampling_step = 1.0 / fixup_factor;
-    let downsampling_step_fraction = downsampling_step.fract();
-    let downsampling_step = downsampling_step.floor() as usize;
-    let mut downsampled_buffer = Vec::<u8>::with_capacity(target_width * target_height * 4);
-
-    let mut source_line_start = 0;
-    let mut line_step_fraction = 0.0f64;
-
-    for _ in 0..target_height {
-        let line_step = if line_step_fraction > 0.99 {
-            line_step_fraction = line_step_fraction.fract();
-            downsampling_step + 1
-        } else {
-            downsampling_step
-        };
 
-        let mut source_column_start = 0;
-        let mut column_step_fraction = 0.0f64;
+    let mut downsampled_buffer = Vec::<u8>::with_capacity(target_width * target_height * 4);
 
-        let source_end_line = min(source_line_start + line_step, bitmap_height);
+    for line_index in 0..target_height {
+        let line_index = line_index as f64;
+        let source_line_start = (line_index * downsampling_step).ceil() as usize;
+        let source_line_end =
+            min(((line_index + 1.) * downsampling_step).floor() as usize, bitmap_height);
 
-        for _ in 0..target_width {
-            let column_step = if column_step_fraction > 0.99 {
-                column_step_fraction = column_step_fraction.fract();
-                downsampling_step + 1
-            } else {
-                downsampling_step
-            };
+        for column_index in 0..target_width {
+            let column_index = column_index as f64;
+            let source_column_start = (column_index * downsampling_step).ceil() as usize;
+            let source_column_end =
+                min(((column_index + 1.) * downsampling_step).floor() as usize, bitmap_width);
 
             let (mut r, mut g, mut b, mut a) = (0u32, 0u32, 0u32, 0u32);
             let mut pixels_picked: u32 = 0;
 
-            let source_end_column = min(source_column_start + column_step, bitmap_width);
-
-            for source_line in source_line_start..source_end_line {
+            for source_line in source_line_start..source_line_end {
                 let source_pixel_index = source_line * bitmap_width;
 
-                for source_column in source_column_start..source_end_column {
+                for source_column in source_column_start..source_column_end {
                     let offset = (source_pixel_index + source_column) * 4;
                     r += bitmap_buffer[offset] as u32;
                     g += bitmap_buffer[offset + 1] as u32;
@@ -642,26 +627,14 @@ fn downsample_bitmap(mut bitmap_glyph: RasterizedGlyph, fixup_factor: f64) -> Ra
                 }
             }
 
-            if pixels_picked == 0 {
-                downsampled_buffer.push(0);
-                downsampled_buffer.push(0);
-                downsampled_buffer.push(0);
-                downsampled_buffer.push(0);
-            } else {
-                downsampled_buffer.push((r / pixels_picked) as u8);
-                downsampled_buffer.push((g / pixels_picked) as u8);
-                downsampled_buffer.push((b / pixels_picked) as u8);
-                downsampled_buffer.push((a / pixels_picked) as u8);
-            }
-
-            column_step_fraction += downsampling_step_fraction;
-            source_column_start += column_step;
+            downsampled_buffer.push((r / pixels_picked) as u8);
+            downsampled_buffer.push((g / pixels_picked) as u8);
+            downsampled_buffer.push((b / pixels_picked) as u8);
+            downsampled_buffer.push((a / pixels_picked) as u8);
         }
-        line_step_fraction += downsampling_step_fraction;
-        source_line_start += line_step;
     }
 
-    bitmap_glyph.top = (bitmap_glyph.top as f32 * fixup_factor as f32) as i32;
+    bitmap_glyph.top = (bitmap_glyph.top as f64 * fixup_factor) as i32;
     bitmap_glyph.left = (bitmap_glyph.left as f64 * fixup_factor) as i32;
     bitmap_glyph.width = target_width as i32;
     bitmap_glyph.height = target_height as i32;
```

font/src/ft/mod.rs
```diff
@@ -529,6 +581,66 @@ impl FreeTypeRasterizer {
...
+
+    for line_index in 0..target_height {
+        let line_index = line_index as f64;
+        let source_line_start = (line_index * downsampling_step).floor() as usize;
+        let source_line_end =
+            min(((line_index + 1.) * downsampling_step).floor() as usize, bitmap_height);
+
+        for column_index in 0..target_width {
+            let column_index = column_index as f64;
+            let source_column_start = (column_index * downsampling_step).floor() as usize;
```
kchibisov: It should be `floor` and not `ceil`. Your approach has the bug with missing pixels/etc, just zoom in with your diff and you should crash.

font/src/ft/mod.rs
```diff
@@ -529,6 +581,66 @@ impl FreeTypeRasterizer {
...
+    let target_width = (bitmap_width as f64 * fixup_factor) as usize;
+    let target_height = (bitmap_height as f64 * fixup_factor) as usize;
+
+    let downsampling_step = 1.0 / fixup_factor;
+
+    let mut downsampled_buffer = Vec::<u8>::with_capacity(target_width * target_height * 4);
+
+    for line_index in 0..target_height {
+        let line_index = line_index as f64;
+        let source_line_start = (line_index * downsampling_step).floor() as usize;
```
kchibisov: same here

font/src/ft/mod.rs
```diff
@@ -529,6 +581,64 @@ impl FreeTypeRasterizer {
...
+                    b += bitmap_buffer[offset + 2] as u32;
+                    a += bitmap_buffer[offset + 3] as u32;
+                    pixels_picked += 1;
+                }
+            }
+
+            downsampled_buffer.push((r / pixels_picked) as u8);
+            downsampled_buffer.push((g / pixels_picked) as u8);
+            downsampled_buffer.push((b / pixels_picked) as u8);
+            downsampled_buffer.push((a / pixels_picked) as u8);
```
chrisduerr: Another comment idea:

```rust
// Add a single pixel for the downscaled rectangle of the source bitmap into the output buffer.
```

font/src/ft/mod.rs
```diff
@@ -529,6 +581,64 @@ impl FreeTypeRasterizer {
...
+        }
+    }
+
+    bitmap_glyph.top = (bitmap_glyph.top as f64 * fixup_factor) as i32;
+    bitmap_glyph.left = (bitmap_glyph.left as f64 * fixup_factor) as i32;
+    bitmap_glyph.width = target_width as i32;
+    bitmap_glyph.height = target_height as i32;
+    bitmap_glyph.buf = BitmapBuffer::RGBA(downsampled_buffer);
+    bitmap_glyph
```
chrisduerr: I've looked at this for every review I've done so far and I never really enjoyed how big of a 'blob' of code this is. However I never really wanted to say anything because I feel like this does all fit together nicely and we of course don't want to just add useless newlines.

However I'd suggest the following change, which I think improves things a bit:

```rust
bitmap_glyph.buf = BitmapBuffer::RGBA(downsampled_buffer);

// Downscale the metrics
bitmap_glyph.top = (bitmap_glyph.top as f64 * fixup_factor) as i32;
bitmap_glyph.left = (bitmap_glyph.left as f64 * fixup_factor) as i32;
bitmap_glyph.width = target_width as i32;
bitmap_glyph.height = target_height as i32;

bitmap_glyph
```

I think putting the assignment to the buffer at the front is nice because that's mostly what we've been working on until now. So it's a bit more tied to the loop we've been going through.

Having the metrics in a separate block allows us to add a nice comment, making it easier to read (or making it possible not to read it at all!).

Having the return on a separate line is always nice I think because it just separates it a bit and makes it clear that we actually really want to return something and haven't just forgotten a semicolon.

cole-h:
For posterity, a few issues I've observed so far (not at all the fault of this PR; good work @kchibisov!):

- If your emoji still render as black-and-white, you likely have an issue with your fontconfig like I did. On Arch, I solved this by following this guide: https://www.reddit.com/r/archlinux/comments/9q8dlj/how_to_better_enable_color_emojis/
  - (if that ever gets deleted, I've mirrored it here: https://gist.github.com/cole-h/8aab0ed9d65efe38496e8e27b96b6a3d)
- The current version of `fish` shell from the Arch repos (3.0.2-3) has issues with displaying emoji, leading to repeated characters and other issues when emoji are present. If this is you, you have a few choices:
  - Install the [`fish-git`](https://aur.archlinux.org/packages/fish-git/) AUR package (or [install `fish` from source](https://github.com/fish-shell/fish-shell))
  - Wait until the next release in which this problem is fixed
  - Switch to a different shell, such as `zsh` or `bash`

hkunzhe:
> For posterity, a few issues I've observed so far (not at all the fault of this PR; good work @kchibisov!):
> 
> * If your emoji still render as black-and-white, you likely have an issue with your fontconfig like I did. On Arch, I solved this by following this guide: https://www.reddit.com/r/archlinux/comments/9q8dlj/how_to_better_enable_color_emojis/
>   
>   * (if that ever gets deleted, I've mirrored it here: https://gist.github.com/cole-h/8aab0ed9d65efe38496e8e27b96b6a3d)
> * The current version of `fish` shell from the Arch repos (3.0.2-3) has issues with displaying emoji, leading to repeated characters and other issues when emoji are present. If this is you, you have a few choices:
>   
>   * Install the [`fish-git`](https://aur.archlinux.org/packages/fish-git/) AUR package (or [install `fish` from source](https://github.com/fish-shell/fish-shell))
>   * Wait until the next release in which this problem is fixed
>   * Switch to a different shell, such as `zsh` or `bash`

I have try the guide you provided, but the color emoji can't render correctly.
image
image
The former is inTilix, and the latter is in Alacritty.
I'm using Manjaro, and the Alacritty version is the newnest.

kchibisov:
Could you show output of `alacritty --version`?

Findarato:
I wanted to add in that I have the same problems with color emoji

image

alacritty 0.4.1

Top is alacritty, bottom is kitty.  Both have the same font specified **Hack**

hkunzhe:
> Could you show output of `alacritty --version`?
image
The version is 0.4.0.

chrisduerr:
This issue is not about suddenly drawing uncolored emojis in color. Uncolored emojis are completely separate glyphs from colored ones.

This issue resolved the issue of colored emojis not being rendered at all.

If you see uncolored emojis and other terminals choose different (colored) glyphs, then you're probably more interested in #3163.

hkunzhe:
@chrisduerr Thanks for your reply. So I have to wait for the next release:)

PR #3043, opened 2019-11-27
Move icon embed build step out of winpty, etc

sterlingjensen:
- In preparation for feature gating winpty (#2890), move the embed step out of the winpty build process.
- Add default-members to workspace, this permits non-targeted platforms to completely ignore unused packages.
- Remove default features from image dependency, further reducing build time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,12 @@ members = [
     "font",
     "winpty"
 ]
+default-members = [
+    "alacritty",
+    "alacritty_terminal",
+    "copypasta",
+    "font"
+]
 
 [profile.release]
 lto = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,12 +6,6 @@ members = [
     "font",
     "winpty"
 ]
-default-members = [
-    "alacritty",
-    "alacritty_terminal",
-    "copypasta",
-    "font"
-]
 
 [profile.release]
 lto = true
--- a/alacritty/build.rs
+++ b/alacritty/build.rs
@@ -18,6 +18,9 @@ use std::env;
 use std::fs::File;
 use std::path::Path;
 
+#[cfg(windows)]
+use embed_resource;
+
 fn main() {
     let hash = rustc_tools_util::get_commit_hash().unwrap_or_default();
     println!("cargo:rustc-env=GIT_HASH={}", hash);
@@ -28,4 +31,7 @@ fn main() {
     Registry::new(Api::Gl, (4, 5), Profile::Core, Fallbacks::All, ["GL_ARB_blend_func_extended"])
         .write_bindings(GlobalGenerator, &mut file)
         .unwrap();
+
+    #[cfg(windows)]
+    embed_resource::compile("../extra/windows/windows.rc");
 }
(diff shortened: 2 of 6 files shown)


chrisduerr:
> In preparation for feature gating winpty

There's no plan to feature-gate WinPTY anytime soon. I don't see why we would take steps not to achieve that in potentially years?

sterlingjensen:
I've got a branch that does just that, keeping it a default feature. This allows a `cargo build --no-default-features`, which requires substantially less time and fewer dependencies, while having no effect on those who enjoy the default build time.

chrisduerr:
Fair enough, making it possible to opt out of winpty seems like a reasonable addition, I've got no issues with that.

Cargo.toml
```diff
@@ -6,6 +6,12 @@ members = [
     "font",
     "winpty"
 ]
+default-members = [
+    "alacritty",
+    "alacritty_terminal",
+    "copypasta",
+    "font"
+]
```
chrisduerr: Now this is an interesting change. Based on these changes I'd assume that the fundamental idea is that we do not want to build anything unless we actually need it. Even if something is not part of the workspace, the dependency will be pulled in, so theoretically we'd only need `alacritty` as the only `default-member`.

However there are some things that do not 'just work'. While `cargo build` resolves all dependencies through their respective `Cargo.toml` files and `cargo clippy` and `cargo fmt` also always run on the complete workspace, regardless of `default-member` configuration (though I'm not sure if it will stay this way), the `cargo test` command will always only run for the crates specified as `default-members`.

The `default-members` behavior can be overwritten though, by running `cargo test --all`. However the `members` array can also be overwritten by running `cargo test -p <crate>`.

Since all depedencies on the `winpty` crate are feature-gated behind `cfg(windows)` annotations though, the build of `cargo build -p winpty` takes only 0.25s. So as far as I can tell this does not improve build times at all.

So is there any point in excluding winpty here?

sterlingjensen: At present it only makes the non-windows build plan more sane, and changes nothing about the default windows build plan. After the winpty package is put behind a default feature, a windows no-default-features build plan will enjoy similar sanity.

It is might not save much in build time, but it does save a lot of time for those who are trying to clean up the dependencies. It is a nice to have, not a need to have - so I'll drop it if you objection to its inclusion.

alacritty/build.rs
```diff
@@ -28,4 +31,7 @@ fn main() {
     Registry::new(Api::Gl, (4, 5), Profile::Core, Fallbacks::All, ["GL_ARB_blend_func_extended"])
         .write_bindings(GlobalGenerator, &mut file)
         .unwrap();
+
+    #[cfg(windows)]
+    embed_resource::compile("../extra/windows/windows.rc");
```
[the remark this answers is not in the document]
sterlingjensen: I was just happy to find another build.rs to stick it in, after noticing how plain my newly build alacritty.exe appeared.

Cargo.toml
chrisduerr: > and changes nothing about the default windows build plan

It will affect the default windows build plan by not running winpty tests anymore.

> It is might not save much in build time, but it does save a lot of time for those who are trying to clean up the dependencies

It does not save any depedencies, all depedencies on all platforms are always part of the lockfile. The depedencies are never installed on non-windows platforms, regardless of inclusion in the defaults or not.

> After the winpty package is put behind a default feature, a windows no-default-features build plan will enjoy similar sanity.

This could still be done once the change is actually made, so I don't see any advantage in doing it now. Though if you want, we could add the WinPTY flag in this feature too, then there might be a bit more substance and reason for the changes made.

Issue #3046, opened 2019-11-28
Shift-Delete in vim's command mode isn't working as expected


nixpulvis:
When I enter command mode `:` in vim, and type something, then start deleting with shift pressed, it simply moves the cursor instead of deleting the characters. I've tested this with both `alacritty` and `xterm-256color` as my `$TERM`.

I'm curious why it works in insert mode, or generally everywhere else, just not in command mode. Other terminals don't seem to have this behavior.

chrisduerr:
I can't reproduce this in neovim, so I'd assume it's vim specific.

Do you get any different output when pressing shift+del in Alacritty vs a different terminal, when pressing ctrl+v to print the raw key escape?

kchibisov:
FWIW I can repro in neovim.

PR #3049, opened 2019-11-28
Add terminal escape bindings with combined modifiers for Delete and Insert

kchibisov:
These bindings was forgotten, when the first massive addition took place.

Fixes: #3046


alacritty/src/config/bindings.rs
```diff
@@ -344,53 +344,57 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
         let modifiers_code = index + 2;
         bindings.extend(bindings!(
             KeyBinding;
-            Key::Up,    [shift: mods.shift, alt: mods.alt, ctrl: mods.ctrl];
+            Key::Insert, [shift: mods.shift, alt: mods.alt, ctrl: mods.ctrl];
```
chrisduerr: On non-macos platforms we do already have Shift+Insert bound to `PasteSelection`. So I'm not sure what we want to do in this case?

I personally do not care about this `Shift+Insert` binding, but in the current version of the PR this would mean that both bindings are active as far as I can tell, so that could obviously cause issues.

alacritty/src/config/bindings.rs
```diff
@@ -454,6 +458,7 @@ pub fn platform_key_bindings() -> Vec<KeyBinding> {
         Key::Equals, [logo: true]; Action::IncreaseFontSize;
         Key::Add, [logo: true]; Action::IncreaseFontSize;
         Key::Minus, [logo: true]; Action::DecreaseFontSize;
+        Key::Insert, [shift: true]; Action::Esc("\x1b[2;2~");
```
chrisduerr:
```suggestion
        Key::Insert, [shift: true]; Action::Esc("\x1b[2;2~".into());
```

PR #3051, opened 2019-11-28
Add documentation for the release process

chrisduerr:


nixpulvis:
Just going to leave a note about what we've discussed so far:

1. Update "Release Process" number 4 to mention cherry-picking
2. Mention stripping old `-dev` when creating new `-dev` entries in `CHANGELOG.md`
3. Give up on including release tags within master's history

chrisduerr:
@nixpulvis Want to give it another read?

CONTRIBUTING.md
```diff
@@ -11,56 +11,117 @@ Table of Contents:
...
+
+The exact steps for an exemplary `1.2.3` might look like this:
+ 1. Initially, the version on the latest master is `1.2.3-dev`
+ 2. A new `v1.2.3` branch is created for the release
+ 3. On master, the version is bumped to `1.2.4-dev`
+        and the `-dev` is stripped from previous change log entries
+ 4. In the branch, the version is bumped to `1.2.3-rc1`
+ 5. The new commit in the branch is tagged as `1.2.3-rc1`
+ 6. A release is created for the `1.2.3-rc1` tag
+ 7. The changelog since the last release (stable or RC) is added
```
nixpulvis: I'm honestly not 100% sure what this means. I got to this step and had already updated the changelog's latest entry to `0.4.1-rc1` with all the latest changes from `0.4.1-dev`. This step seems like it might be a specific reminder to update the changelog on subsequent runs through the loop. If so, we should try and make this a bit more clear somehow... Maybe it could be as simple as "verify that the changelog include all cherry-picked fixes"? But then why not make this explicit in the next step where we talk about cherry-picking?

A cherry pick of a commit with a changelog entry should merge into the RC section no problem unless there's a case I'm missing. In my mock release this worked without any manual intervention.

CONTRIBUTING.md
```diff
@@ -11,56 +11,117 @@ Table of Contents:
...
+ 8. Bug fixes are cherry-picked from master into the branch and steps 4-7
+        are repeated until no major issues are found in the release candidates
+ 9. In the branch, the version is bumped to `1.2.3`
```
nixpulvis:
```suggestion
 9. Once finalized, the release branch version is bumped to `1.2.3`
```

CONTRIBUTING.md
```diff
@@ -11,56 +11,117 @@ Table of Contents:
...
+ 10. The new commit in the branch is tagged as `1.2.3`
```
nixpulvis:
```suggestion
 10. The new commit in the branch is tagged as `v1.2.3`
```

CONTRIBUTING.md
```diff
@@ -11,56 +11,117 @@ Table of Contents:
...
+ 11. A release is created for the `1.2.3` tag
```
nixpulvis:
```suggestion
 11. A release is created for the `v1.2.3` tag
```

CONTRIBUTING.md
```diff
@@ -11,56 +11,117 @@ Table of Contents:
...
+ 12. The changelog since the last stable release (**not** RC) is added
```
nixpulvis: Shouldn't this step be to consolidate the RC changelog entries into a final `1.2.3` entry? Or have I missed something here?

CONTRIBUTING.md
```diff
@@ -11,56 +11,117 @@ Table of Contents:
...
+To achieve these goals, a new branch is created for every new release. Both the release candidates
+and the final version are only comitted and tagged in this branch. The master branch only tracks
+development versions, allowing us to keep the branches completely separate without merging releases
+back into master.
+
+The exact steps for an exemplary `1.2.3` might look like this:
+ 1. Initially, the version on the latest master is `1.2.3-dev`
+ 2. A new `v1.2.3` branch is created for the release
+ 3. On master, the version is bumped to `1.2.4-dev`
+        and the `-dev` is stripped from previous change log entries
```
nixpulvis: For maintainers following this in order it would be nice to avoid switching branches many times. This step can be moved below the current step number 7.

CONTRIBUTING.md
```diff
@@ -11,56 +11,117 @@ Table of Contents:
...
+ 4. In the branch, the version is bumped to `1.2.3-rc1`
```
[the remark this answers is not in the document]
chrisduerr: > Should we enumerate the places the version string appears here, or just hint at searching for the old string? Is this obvious?

I think having a list would be kinda troublesome, because if a list exists, people expect that list to be correct. I can totally see us forgetting to add a new file to the list and then we forget to bump it.

The best way to check is probably to just run `git show vX.Y.Z --stat` and that will list all files bumped in the last release, that's probably the best way to go about it and that's what I do. Doing a `grep` in the repo if you're uncertain should leave you with way more certainty than a centralized list.

> Also, maybe we should indicate that the versions should be 1.2.3-rcX since this step is referred to in a loop.

The thing with this is that we never print a full example and I think examples is exactly what you want to orient yourself with. I would expect someone who makes releases to realize that the rc1 should be bumped to rc2, given our previous RC periods, however if someone has no exact reference of what such a tag might look like, it might be more confusing (start with 0 or 1 for example).

I've thought about spelling it out twice, with the second part being repeated and the rc1 being layed out on its own, but I think that would just be too much. I think we can expect at least a little bit of thinking.

CONTRIBUTING.md
```diff
@@ -11,56 +11,117 @@ Table of Contents:
...
+ 5. The new commit in the branch is tagged as `1.2.3-rc1`
```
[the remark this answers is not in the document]
chrisduerr: Great point.

CONTRIBUTING.md
chrisduerr: This is about the release's github description. I'll clarify.

CONTRIBUTING.md
chrisduerr: I don't see what needs to be 'finalized' here? It's the next step, it shouldn't depend on any 'state'. As long as all previous steps are successful, this can be executed.

CONTRIBUTING.md
chrisduerr: I think it's easier to have the changelog just in one big section since the last release. This is about the github release description again.

CONTRIBUTING.md
chrisduerr: > This step can be moved below the current step number 7.

No, I specifically don't want to do this. I think it makes sense to separate the two branches as quickly as possible, having a long period of time where the master branch is in an incorrect state does not seem like something we should strive for.

Of course once a contributor is more experienced with the process, they might be able to delay it until the branch is waiting for travis to build for example, which gives a nice timeframe to update master. But generally I think that should be something the dev decides on their own, once they're confident.

> Also, shouldn't we add a new section to the changelog here? We won't know exactly what version number this should be at this time, so maybe this should just be master or something indicating development?

The version is bumped to `1.2.4-dev`, that includes the changelog.

chrisduerr:
> OK, I tried walking through a release, and while I'm pretty sure I understand what's intended to happen, some of this could be rearranged or reworded to make a bit more sense. Sorry if my comments aren't clear, and for the quantity.

Thanks a lot for the feedback, especially the confusion about updating the changelog in the GitHub releases has exposed a very good point.

I've addressed all points that I thought were definite improvements and left comments on anything else, trying to explain my thoughts behind it. If you feel like some comments don't sufficiently justify the current documentation, or you have any further feedback, please let me know.

PR #3054, opened 2019-11-29
Fix deadlock issue in Conpty drop implementation

### LINKED ISSUE #3042: Window close "X" button freezes alacritty
Windows 10 (insiders build 19033)

On my system, attempting to close Alacritty with the X button in the title bar causes it to crash. The whole application freezes. The CPU usage is 0% so I assume some kind of deadlock.

It's been a problem for a few months; not sure when it got introduced. I'd wondered about looking into it myself, but don't have the time right now. I might take a look in the New Year if it's still an issue then.

I'm running powershell as my shell. ConPTY backend.

(I'd been waiting for the glutin event loop update to land in case that fixed it. Then saw some similar work for WinPTY which I wondered if would fix it - doesn't seem to have helped.)

See example in this GIF (goes to not responding right at the end):

[Image from Gyazo](https://gyazo.com/facca0b47b773434d52e3c70c547b091)

davidhewitt:
Fixes #3042 

I also refactored some of the differences between winpty / conpty at the same time - resulting in more uniform interfaces and less code overall. Also got rid of the lifetime on the `Pty` type 😄

### DIFF
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -42,7 +42,7 @@ use crate::config::{Config, Shell};
 use crate::event::OnResize;
 use crate::term::SizeInfo;
 use crate::tty::windows::child::ChildExitWatcher;
-use crate::tty::windows::Pty;
+use crate::tty::windows::{Pty, PtyImpl};
 
 /// Dynamically-loaded Pseudoconsole API from kernel32.dll
 ///
@@ -85,8 +85,8 @@ impl ConptyApi {
 /// RAII Pseudoconsole
 pub struct Conpty {
     inner: Arc<ConptyInner>,
-    pub conout: EventedAnonRead,
-    pub conin: EventedAnonWrite,
+    conout: EventedAnonRead,
+    conin: EventedAnonWrite,
 }
 
 /// ResizeHandle can be freely cloned and moved between threads
@@ -104,17 +104,35 @@ struct ConptyInner {
     closed: AtomicBool,
 }
 
-impl Conpty {
-    pub fn resize_handle(&self) -> ConptyResizeHandle {
+impl PtyImpl for Conpty {
+    type ResizeHandle = ConptyResizeHandle;
+    type Conout = EventedAnonRead;
+    type Conin = EventedAnonWrite;
+
+    fn resize_handle(&self) -> Self::ResizeHandle {
         ConptyResizeHandle { inner: self.inner.clone() }
+    }
+
+    fn conout(&self) -> &Self::Conout {
+        &self.conout
+    }
+
+    fn conout_mut(&mut self) -> &mut Self::Conout {
+        &mut self.conout
+    }
+
+    fn conin(&self) -> &Self::Conin {
+        &self.conin
+    }
+
+    fn conin_mut(&mut self) -> &mut Self::Conin {
+        &mut self.conin
     }
 }
 
 impl Drop for Conpty {
     fn drop(&mut self) {
-        // Mark as closed
-        // so that the resize handles don't attempt to resize a
-        // dead pseudoconsole.
+        // Mark as closed so that the resize handles don't attempt to resize a dead pseudoconsole.
         self.inner.closed.store(false, Ordering::SeqCst);
 
         // The i/o pipes must be dropped at the same time as the
@@ -128,7 +146,11 @@ impl Drop for Conpty {
 
 unsafe impl Send for Conpty {}
 
-pub fn new<C>(config: &Config<C>, size: &SizeInfo, _window_id: Option<usize>) -> Option<Pty> {
+pub fn new<C>(
+    config: &Config<C>,
+    size: &SizeInfo,
+    _window_id: Option<usize>,
+) -> Option<Pty<Conpty>> {
     if !config.enable_experimental_conpty_backend {
         return None;
     }
@@ -273,13 +295,7 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, _window_id: Option<usize>) ->
         conin: EventedAnonWrite::new(conin),
     };
 
-    Some(Pty {
-        inner: super::PtyInner::Conpty(conpty),
-        read_token: 0.into(),
-        write_token: 0.into(),
-        child_event_token: 0.into(),
-        child_watcher,
-    })
+    Some(Pty::new(conpty, child_watcher))
 }
 
 // Panic with the last os error as message
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -17,7 +17,7 @@ use std::io::Error;
 use std::mem;
 use std::os::windows::io::IntoRawHandle;
 use std::ptr;
-use std::sync::Arc;
+use std::sync::{Arc, atomic::{AtomicBool, Ordering}};
 
 use dunce::canonicalize;
 use mio_anonymous_pipes::{EventedAnonRead, EventedAnonWrite};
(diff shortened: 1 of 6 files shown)


alacritty_terminal/src/tty/windows/winpty.rs
```diff
@@ -31,42 +31,40 @@ use crate::term::SizeInfo;
...
+        let (cols, lines) = (size.cols().0, size.lines().0);
+        if cols > 0 && cols <= u16::MAX as usize && lines > 0 && lines <= u16::MAX as usize {
+            let winpty: &mut Winpty = unsafe {
+                // This transmute is actually thread-safe since Winpty uses a mutex internally.
+                std::mem::transmute(&self.winpty as *const _ as *mut Winpty)
+            };
+
+            winpty
+                .set_size(cols as u16, lines as u16)
+                .unwrap_or_else(|_| info!("Unable to set winpty size, did it die?"));
```
chrisduerr: Why do we need two WinPTY `OnResize` handlers?

davidhewitt: We don't; this is the one we do need; the old one `impl OnResize for Winpty` should be deleted. I'll tidy up.

davidhewitt:
> First one is about the change to the conout and conin fields. It seems
like previously in the mod.rs we didn't have to worry much about ConPTY or
WinPTY and we could just have indepedent conout and conin fields. However
you've changed this and now we have to match over ConPTY and WinPTY
everywhere. What was the motivation behind this?

The motivation is also the answer to your second question! Moving the
`conout` and `conin` fields around like this resolves the deadlock.

To be specific, I found after a bit of playing around that if the io pipes
were dropped at the same time as the Conpty then the API call didn't
deadlock. So I moved them into the `Conpty` struct so the interface
expressed this constraint. I'll add comments in the code.

chrisduerr:
> To be specific, I found after a bit of playing around that if the io pipes
were dropped at the same time as the Conpty then the API call didn't
deadlock. So I moved them into the `Conpty` struct so the interface
expressed this constraint. I'll add comments in the code.

Can't we just fix this is the drop implementation?

davidhewitt:
> Can't we just fix this is the drop implementation?

TLDR; not quite, the problem was that the `Conpty` drop implemention causing the deadlock needs the i/o pipes to be dropped at the same time. The i/o pipes were not part of the struct, so this requirement of the `Conpty` interface was not being enforced. The correct solution in my eyes was to move the i/o pipes into the `Conpty` struct so that they necessarily had to be dropped at the same time.

----

The problem with the original code was that the `ConptyHandle`, which both the `Pty` and resize handles used, was an `Arc<Conpty>`. This meant that the `Conpty` drop implementation (i.e. the deadlocking `ClosePseudoConsole` call) would only run when second of these `Arc<Conpty>` pointers was dropped. 

On `0.3.3` and before we appear to have been lucky in that the drop order was resize handle first, then `Pty`. The meant the i/o pipes were dropped at the same time as the `Conpty`, and thus no deadlock. On master, this order has switched, and now the `Pty` is dropped (and with it the i/o pipes) long before the resize handle drops. When the resize handle finally drops, the `Arc<Conpty>` also drops, meaning the `ClosePseudoConsole` API call is fired, and we get the deadlock.

I could have written a one-line fix in `main.rs` by explicitly dropping the resize handle before the `Pty`. But this felt like a hackjob given that the code at fault is the original `Conpty` drop implementation I wrote.

So, to ensure that the i/o pipes drop at the same time as `ClosePseudoConsole` is called, I moved the i/o pipes from the `Pty` struct to the `Conpty` struct.

I then decided to do the same for `Winpty` so that the code paths looked as similar as possible. The new code has `match` in a bunch of places, yes, but the old code also had a lot of uses of `match`. The difference is that the old code had two `EventedWritablePipe` and `EventedReadablePipe` enums, inside which all the matching was occuring to select the `Winpty` and `Conpty` code paths.

In the new code, I didn't need those two types and threw them away. The `match` uses are now inside the `Pty` impl to explicitly choose between `Winpty` and `Conpty` code paths where needed. I think if we ever decide to remove the `Winpty` implementation, it will be easier to do so with the code now structured in the way it is in this PR. (The change would simply require using `Conpty` directly instead of the new `PtyInner` enum in the `Pty` type, and deleting all the `Winpty` match arms.)

The new refactored code is also about 50 lines shorter than the old code 😃

chrisduerr:
> I think if we ever decide to remove the Winpty implementation, it will be easier to do so with the code now structured in the way it is in this PR. 

I'm not sure how true this is, because we now need to modify a lot of things in the `windows/mod.rs` when we want to remove WinPTY, because it is used everywhere. Ideally the PTY itself would be generic and we wouldn't have to care which one it is and instead just use dynamic dispatch to let rust handle that for us at runtime. Then we could just delete the WinPTY module with basically no changes to `windows/mod.rs` and everything should basically still work. At least that's the architecture I'd have chosen.

I'll take another look at the PR though and get back to you with concrete feedback. I'll have to re-read it with your comments in mind.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -112,7 +112,16 @@ impl Conpty {
 
 impl Drop for Conpty {
     fn drop(&mut self) {
+        // Mark as closed
+        // so that the resize handles don't attempt to resize a
+        // dead pseudoconsole.
```
cole-h: This comment is not properly broken.

davidhewitt:
> Ideally the PTY itself would be generic and we wouldn't have to care which one it is and instead just use dynamic dispatch to let rust handle that for us at runtime.

I'd be more than happy to modify this PR to use dynamic dispatch. It uses static dispatch because when I was writing the original ConPTY implementation that's what I was instructed to do (I think zacps wanted, perhaps?).

Give me a day or two to find a moment to change this PR to dynamic dispatch :smile:

alacritty_terminal/src/tty/windows/conpty.rs
davidhewitt: Thanks

davidhewitt:
I've now pushed a commit adding dynamic dispatch to the code paths choosing between WinPTY and `conpty`.

It works by adding a file `dynamic.rs` which adds an extension trait `IntoDynamicPty` which can turn a statically-dispatched `Pty<Winpty>` or a `Pty<Conpty>` into a type-erased `Pty<DynPty>`.

This design should make it very easy in the future to strip out the WinPTY stuff and end up just using a statically-dispatched `Pty<Conpty>`.

davidhewitt:
No worries, that makes total sense. I'll find some time to rewrite this within the next few days.

nixpulvis:
Looks like a few refactors are up in another PR (#3054) now too.

sterlingjensen:
Oops, well that is what I get for sitting on a patch too long. I certainly don't want to further complicate what is going on here, would it be better for me to pull it or simply fix it after this one merges?

nixpulvis:
@sterlingjensen given that this PR needs to be reworked anyway I wouldn't worry about your PR. It's a non-functional change so I expect we can review and merge it rather quickly (I skimmed it tonight, and it seemed pretty straightforward).

I'm mostly bringing it up here so that @davidhewitt doesn't get blindsided by a merge conflict.

davidhewitt:
I've opened a cleaned-up version of the original static dispatch design as #3074. I've left this one open as a different branch so that we can revisit this design if we decide we'd rather have dynamic dispatch after all. I can see ways to simplify/remove `dynamic.rs` if we want to just move over entirely to dynamic dispatch. In the current design of this PR I had wanted to keep the option for static dispatch open.

davidhewitt:
Closing this one as I think we're finally settling on a design which will be merged. (#3084)

PR #3060, opened 2019-12-01
Fix colored row reset performance

### LINKED ISSUE #2989: Performance problems with `yes` and colored background
While `yes` performs exceptionally well with the default background color, it appears that scrolling inside Alacritty with a dirty background color causes a big performance hit that slows it down significantly.

In my tests, the simple benchmark of printing a couple of bytes of `yes`, resulted in 6s without background set and ~25s with a dirty background color.

While some performance impact of having to draw significantly more cells might be expected, the disparity between the two benchmarks seems quite extreme and definitely warrants further research.

chrisduerr:
This fixes a bug where a row would always get reset completely if its
background does not equal the default terminal background. This leads to
big performance bottlenecks when running commands like `echo "\e[41m" &&
yes`.

Instead of resetting the entire row whenever the template cell is not
empty, the template cell is now compared to the last cell in the row.
The last cell will always be equal to the previous template cell when
`row.occ < row.inner.len()` and if `occ` is equal to the row's length,
the entire row is always reset anyways.

Fixes #2989.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Clipboard escape `OSC 52` not working with empty clipboard parameter
 - Direct escape input on Windows using alt
 - Incorrect window size on X11 when waking up from suspend
-- Incorrect width of Unicode 11/12 emojis
+- Width of Unicode 11/12 emojis
 - Minimize on windows causing layout issues
+- Performance bottleneck when clearing colored rows
 
 ## 0.4.0
 
--- a/alacritty_terminal/src/grid/mod.rs
+++ b/alacritty_terminal/src/grid/mod.rs
@@ -70,6 +70,12 @@ pub trait GridCell {
     fn is_empty(&self) -> bool;
     fn is_wrap(&self) -> bool;
     fn set_wrap(&mut self, wrap: bool);
+
+    /// Fast equality approximation.
+    ///
+    /// This is a faster alternative to [`PartialEq`],
+    /// but might report inequal cells as equal.
+    fn fast_eq(&self, other: Self) -> bool;
 }
 
 /// Represents the terminal display contents
@@ -140,7 +146,7 @@ pub enum Scroll {
     Bottom,
 }
 
-impl<T: GridCell + Copy + Clone> Grid<T> {
+impl<T: GridCell + PartialEq + Copy> Grid<T> {
     pub fn new(lines: index::Line, cols: index::Column, scrollback: usize, template: T) -> Grid<T> {
         let raw = Storage::with_capacity(lines, Row::new(cols, &template));
         Grid {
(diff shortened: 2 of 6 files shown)


cole-h:
For comparison purposes, using our standard performance test (alt-screen-random-write) on master:
<details>

```
 Performance counter stats for 'cat alt' (5 runs):

          1,372.82 msec task-clock:u              #    0.212 CPUs utilized            ( +-  0.16% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                61      page-faults:u             #    0.044 K/sec                    ( +-  0.90% )
        12,858,245      cycles:u                  #    0.009 GHz                      ( +-  4.83% )
           676,406      instructions:u            #    0.05  insn per cycle           ( +-  0.00% )
           159,363      branches:u                #    0.116 M/sec                    ( +-  0.00% )
            33,378      branch-misses:u           #   20.94% of all branches          ( +-  1.08% )

            6.4754 +- 0.0404 seconds time elapsed  ( +-  0.62% )
```
</details>

Using this branch:
<details>

```
 Performance counter stats for 'cat alt' (5 runs):

          1,347.28 msec task-clock:u              #    0.246 CPUs utilized            ( +-  0.67% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                62      page-faults:u             #    0.046 K/sec                    ( +-  0.32% )
        10,602,936      cycles:u                  #    0.008 GHz                      ( +-  5.52% )
           708,655      instructions:u            #    0.07  insn per cycle           ( +-  0.00% )
           168,324      branches:u                #    0.125 M/sec                    ( +-  0.00% )
            29,461      branch-misses:u           #   17.50% of all branches          ( +-  4.01% )

            5.4674 +- 0.0402 seconds time elapsed  ( +-  0.73% )
```
</details>

Just for fun, asrw-color:
<details>

```
 Performance counter stats for 'cat alt-color' (5 runs):

          1,374.46 msec task-clock:u              #    0.260 CPUs utilized            ( +-  0.17% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                62      page-faults:u             #    0.045 K/sec                    ( +-  0.51% )
         9,408,972      cycles:u                  #    0.007 GHz                      ( +-  1.42% )
           708,653      instructions:u            #    0.08  insn per cycle           ( +-  0.00% )
           168,319      branches:u                #    0.122 M/sec                    ( +-  0.00% )
            26,595      branch-misses:u           #   15.80% of all branches          ( +-  1.76% )

            5.2771 +- 0.0114 seconds time elapsed  ( +-  0.22% )
```
</details>

chrisduerr:
This is not consistent with my results on the latest master @cole-h, so I'm a bit skeptical about these numbers. Asrw performance has no reason te get improved by this.

chrisduerr:
While there is a minor regression with running just `yes` with a normal background color (.85% in my tests), this should significantly speed up every other case.

So I think we should be good to go. After a lot of testing, I wasn't able to get any better performance than what is currently in this PR.

PR #3069, opened 2019-12-05
Remove unnecessary newtype lifetimes

sterlingjensen:
With the exception of Err, all implement Drop on their owned heap allocations and contain no references - making lifetime declaration unnecessary.
In the case of Err, the unused raw pointer was removed - making Drop unnecessary.
Memory is correctly freed according to sanitizer instrumentation, manually verified in debug session.

### DIFF
--- a/winpty/src/windows.rs
+++ b/winpty/src/windows.rs
@@ -48,16 +48,18 @@ pub struct Err {
     message: String,
 }
 
-// Check to see whether winpty gave us an error
+// Check to see whether winpty gave us an error, and perform the necessary memory freeing
 fn check_err(e: *mut winpty_error_t) -> Option<Err> {
-    let (code, message);
     unsafe {
+        let code = winpty_error_code(e);
         let raw = winpty_error_msg(e);
-        code = winpty_error_code(e);
-        message = String::from_utf16_lossy(std::slice::from_raw_parts(raw, wcslen(raw)));
+        let message = String::from_utf16_lossy(std::slice::from_raw_parts(raw, wcslen(raw)));
         winpty_error_free(e);
-    };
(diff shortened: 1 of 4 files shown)


winpty/src/windows.rs
```diff
@@ -40,61 +43,44 @@ bitflags!(
...
+        code = winpty_error_code(e);
+        message = String::from_utf16_lossy(std::slice::from_raw_parts(raw, wcslen(raw)));
+        winpty_error_free(e);
     };
-    if err.code == 0 {
-        None
-    } else {
-        Some(err)
-    }
+    Some(Err { code, message }).filter(|e| e.code != 0)
```
chrisduerr: I think this is too 'smart'. The if/else might be a bit longer, but I think it's easier to read.

I also don't think it's nice to define a variable and then assign it in an unsafe scope, instead of returning the variables from that scope, because it makes it hard in longer function bodies to understand where exactly variables are initialized.

So `let (code, message) = unsafe { ... }` would be more clear and should be the way we do things by default.

This is probably how I would write this function:
```rust
unsafe {
    let code = match winpty_error_code(e) {
        0 => return None,
        code => code,
    };
    let raw_msg = winpty_error_msg(e);
    let message = String::from_utf16_lossy(std::slice::from_raw_parts(raw, wcslen(raw_msg)));
    Some(Err { code, message })
}
```

sterlingjensen: As a fan of the suckless style, I agree. Unfortunately the ordering is largely dictated by the need to call winpty_error_free(e) before returning, but after copying the referenced string. Obviously code clarity trumps keeping a match statement out of the unsafe block.

Issue #1905, opened 2018-12-15
Use Overlapped I/O for Windows 10 ConPty APIs


davidhewitt:
This is currently not possible as Windows 10 1809's ConPTY APIs only support synchronous I/O. Microsoft have expressed intent to support overlapped I/O in the future at https://github.com/Microsoft/console/issues/262. If this lands then we can simplify a lot of the code in `src/tty/windows` to just use `mio_named_pipes::NamedPipe` and drop the dependency on `mio_anonymous_pipes`.

From the discussion on #1762 I've opened this ticket to track the future intention.

davidhewitt:
https://github.com/microsoft/terminal/issues/262 was just closed, so a future Windows release will support overlapped I/O. I see the `polling` port has already removed the `mio-anonymous-pipes` dependency, so the future improvement here will just be about the ability to remove the in-tree threaded reader / writers.

chrisduerr:
> a future Windows release will support overlapped I/O

Noteworthy that this could be decades out. They're **very** slow to update their ConPTY.

n00bmind:
I'm also following the evolution of this issue with interest.. Is there any way of knowing when a particular Windows Update will incorporate the ConPTY version that includes this feature? Anything else we can track?

PR #3074, opened 2019-12-06
Fix deadlock with closing on Windows using Conpty (v2)

davidhewitt:
The ClosePseudoConsole call was deadlocking if it was called after the i/o pipes had already been closed in some other thread. To avoid this issue the i/o pipes have been moved into the Conpty struct in order to ensure they are dropped (closed) at the same time as the pseudoconsole.

This commit also makes corresponding changes to the Winpty backend to keep the code uniform.

Compared to #3054, this PR:
- Also moves the i/o pipes into the `Conpty` / `WinptyAgent` structs to ensure that `Conpty`'s drop closes the i/o pipes and calls .
- Instead to static dispatch using `match`
- Instead does not remove the `EventedReadablePipe` and `EventedWritablePipe` types
- Instead attempts to inline the necessary `match` calls to choose between the `Winpty` / `Conpty` backends

Also many thanks to @sterlingjensen for #3069 which makes this changeset smaller and neater. No lifetimes 😄 !

### DIFF
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -18,6 +18,7 @@ use std::mem;
 use std::os::windows::io::IntoRawHandle;
 use std::ptr;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use dunce::canonicalize;
 use mio_anonymous_pipes::{EventedAnonRead, EventedAnonWrite};
@@ -26,7 +27,7 @@ use widestring::U16CString;
 use winapi::shared::basetsd::{PSIZE_T, SIZE_T};
 use winapi::shared::minwindef::{BYTE, DWORD};
 use winapi::shared::ntdef::{HANDLE, HRESULT, LPWSTR};
-use winapi::shared::winerror::S_OK;
+use winapi::shared::winerror::{E_HANDLE, S_OK};
 use winapi::um::libloaderapi::{GetModuleHandleA, GetProcAddress};
 use winapi::um::processthreadsapi::{
     CreateProcessW, InitializeProcThreadAttributeList, UpdateProcThreadAttribute,
@@ -83,14 +84,71 @@ impl ConptyApi {
 pub struct Conpty {
     pub handle: HPCON,
     api: ConptyApi,
+    closed: AtomicBool
 }
 
 /// Handle can be cloned freely and moved between threads.
 pub type ConptyHandle = Arc<Conpty>;
 
+impl Conpty {
+    pub fn new(size: &SizeInfo) -> Option<(Self, EventedAnonRead, EventedAnonWrite)> {
+        let api = ConptyApi::new()?;
+
+        let mut handle = 0 as HPCON;
+
+        // Passing 0 as the size parameter allows the "system default" buffer
+        // size to be used. There may be small performance and memory advantages
+        // to be gained by tuning this in the future, but it's likely a reasonable
+        // start point.
+        let (conout, conout_pty_handle) = miow::pipe::anonymous(0).unwrap();
+        let (conin_pty_handle, conin) = miow::pipe::anonymous(0).unwrap();
+
+        let coord =
+            coord_from_sizeinfo(size).expect("Overflow when creating initial size on pseudoconsole");
+
+        // Create the Pseudo Console, using the pipes
+        let result = unsafe {
+            (api.CreatePseudoConsole)(
+                coord,
+                conin_pty_handle.into_raw_handle(),
+                conout_pty_handle.into_raw_handle(),
+                0,
+                &mut handle as *mut HPCON,
+            )
+        };
+
+        assert_eq!(result, S_OK);
+
+        let this = Self {
+            handle,
+            api,
+            closed: AtomicBool::new(false)
+        };
+        let conout = EventedAnonRead::new(conout);
+        let conin = EventedAnonWrite::new(conin);
+
+        Some((this, conout, conin))
+    }
+
+    /// Close the pseudoconsole.
+    ///
+    /// When this function is called, the conout pipe must still be alive, as the pseudoconsole
+    /// will write some final content out on that pipe. This function will block until the final
+    /// output is drained from conout. (So that conout must be drained by a different thread.)
+    ///
+    /// (See the docs for `ClosePseudoConsole` at
+    /// https://docs.microsoft.com/en-us/windows/console/closepseudoconsole)
+    pub fn close(&self) {
+        // Cannot call .close() multiple times on the same pseudoconsole, else segfault
+        debug_assert_eq!(false, self.closed.swap(true, Ordering::SeqCst));
+        unsafe { (self.api.ClosePseudoConsole)(self.handle) }
+    }
+}
+
 impl Drop for Conpty {
     fn drop(&mut self) {
-        unsafe { (self.api.ClosePseudoConsole)(self.handle) }
+        // Must call .close() on a conpty or the pseudoconsole process leaks
+        debug_assert!(self.closed.load(Ordering::SeqCst));
     }
 }
 
@@ -103,32 +161,7 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, _window_id: Option<usize>) ->
         return None;
     }
 
-    let api = ConptyApi::new()?;
-
-    let mut pty_handle = 0 as HPCON;
-
-    // Passing 0 as the size parameter allows the "system default" buffer
-    // size to be used. There may be small performance and memory advantages
-    // to be gained by tuning this in the future, but it's likely a reasonable
-    // start point.
-    let (conout, conout_pty_handle) = miow::pipe::anonymous(0).unwrap();
-    let (conin_pty_handle, conin) = miow::pipe::anonymous(0).unwrap();
-
-    let coord =
-        coord_from_sizeinfo(size).expect("Overflow when creating initial size on pseudoconsole");
-
-    // Create the Pseudo Console, using the pipes
-    let result = unsafe {
-        (api.CreatePseudoConsole)(
-            coord,
-            conin_pty_handle.into_raw_handle(),
-            conout_pty_handle.into_raw_handle(),
-            0,
-            &mut pty_handle as *mut HPCON,
-        )
-    };
-
-    assert_eq!(result, S_OK);
+    let (conpty, conout, conin) = Conpty::new(size)?;
 
     let mut success;
 
@@ -191,8 +224,8 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, _window_id: Option<usize>) ->
             startup_info_ex.lpAttributeList,
             0,
             22 | 0x0002_0000, // PROC_THREAD_ATTRIBUTE_PSEUDOCONSOLE
-            pty_handle,
-            mem::size_of::<HPCON>(),
+            conpty.handle,
+            mem::size_of_val(&conpty.handle),
             ptr::null_mut(),
             ptr::null_mut(),
         ) > 0;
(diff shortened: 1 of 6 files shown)


sterlingjensen:
@davidhewitt Glad I could help, I was kicking myself after seeing your pull request and recognizing a bunch of very familiar code in the diff.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -17,7 +17,10 @@ use std::io::Error;
 use std::mem;
 use std::os::windows::io::IntoRawHandle;
 use std::ptr;
-use std::sync::Arc;
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
```
chrisduerr: We don't use nested imports like this in Alacritty, they're usually too complex and this would just be two normal short lines with the more traditional layout.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -236,33 +268,23 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, _window_id: Option<usize>) ->
...
     panic!("Unable to spawn shell: {}", Error::last_os_error());
 }
 
-impl OnResize for ConptyHandle {
+impl OnResize for ConptyResizeHandle {
     fn on_resize(&mut self, sizeinfo: &SizeInfo) {
-        if let Some(coord) = coord_from_sizeinfo(sizeinfo) {
-            let result = unsafe { (self.api.ResizePseudoConsole)(self.handle, coord) };
-            assert_eq!(result, S_OK);
+        if !self.0.closed.load(Ordering::SeqCst) {
```
chrisduerr: Not checking if this is closed causes the deadlock, right? Instead of failing with an error?

It also might be easier to early return here as a gate clause. Especially if this method should ever get bigger for whatever reason that makes a lot of sense. And cutting down on excessive indentation is always pretty nice too.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -79,63 +82,92 @@ impl ConptyApi {
...
 
 impl Drop for Conpty {
     fn drop(&mut self) {
-        unsafe { (self.api.ClosePseudoConsole)(self.handle) }
+        // Mark as closed so resize handles won't attempt to resize the closed pseudoconsole
+        self.inner.closed.store(true, Ordering::SeqCst);
+
+        // NB the i/o pipes must be closed (dropped) at the same time as the ClosePseudoConsole
+        // call. Else this call can deadlock!
+        unsafe { (self.inner.api.ClosePseudoConsole)(self.inner.handle) }
```
chrisduerr: We can't close two things at the same time, we either have to close them before or after. I'm not quite sure what the exact parameters of the restraints are just from this comment and the existing code.

Based on the code, it seems like the I/O pipes are automatically dropped once `Conpty`'s destructor is done? So *after* the `ClosePseudoConsole`?

I also don't think `NB` has ever been used as a comment type in the Alacritty repo and it seems generally kinda uncommon. For annotations about things to watch out for I've been using `XXX` so far I think, would be good to be consistent here.

alacritty_terminal/src/tty/windows/mod.rs
```diff
@@ -38,20 +38,18 @@ pub fn is_conpty() -> bool {
...
+    Winpty(winpty::WinptyResizeHandle),
+    Conpty(conpty::ConptyResizeHandle),
 }
 
 pub struct Pty {
     handle: PtyHandle,
-    // TODO: It's on the roadmap for the Conpty API to support Overlapped I/O.
-    // See https://github.com/Microsoft/console/issues/262
-    // When support for that lands then it should be possible to use
-    // NamedPipe for the conout and conin handles
```
chrisduerr: Where did this comment go? It seems like that would improve the ConPTY API so tracking it makes sense.

Unless we have an issue on GitHub at least, I'm all for opening an issue instead of tracking it in code.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -79,63 +82,92 @@ impl ConptyApi {
...
+
+pub struct ConptyInner {
+    handle: HPCON,
+    api: ConptyApi,
+    closed: AtomicBool,
+}
 
 impl Drop for Conpty {
     fn drop(&mut self) {
-        unsafe { (self.api.ClosePseudoConsole)(self.handle) }
```
chrisduerr: Would it not be possible to implement `Drop` for `PTY` and upholding the drop order in there? I still absolutely cannot stand that we move the `conin` and `conout` on the WinPTY/ConPTY for no good reason at all. All we're creating here is duplication of code that will come back to haunt us in the future.

alacritty_terminal/src/tty/windows/mod.rs
```diff
@@ -60,7 +58,52 @@ pub struct Pty {
...
+            PtyHandle::Conpty(c) => &c.conin,
+        }
+    }
+
+    #[inline]
+    fn conin_mut(&mut self) -> &mut EventedWritablePipe {
+        match &mut self.handle {
+            PtyHandle::Winpty(w) => &mut w.conin,
+            PtyHandle::Conpty(c) => &mut c.conin,
+        }
```
chrisduerr: All of this would be gone if we keep the `conin` and `conout` on the `PTY`. This code is completely useless and just costs us maintenance burden.

alacritty_terminal/src/tty/windows/mod.rs
```diff
@@ -266,15 +308,15 @@ impl EventedReadWrite for Pty {
...
+        poll.deregister(self.conout())?;
+        poll.deregister(self.conin())?;
         poll.deregister(self.child_watcher.event_rx())?;
         Ok(())
     }
 
     #[inline]
     fn reader(&mut self) -> &mut Self::Reader {
-        &mut self.conout
+        self.conout_mut()
```
chrisduerr: Why do we have a `conout_mut` function here, when it is never used anywhere else? Why wouldn't we inline it? Is it solely because we have a `conout` method that is used in other places, making the API more consistent internally? Same goes for conin.

alacritty_terminal/src/tty/windows/winpty.rs
```diff
@@ -137,24 +116,26 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, _window_id: Option<usize>) ->
...
+    Pty::new(super::PtyHandle::Winpty(agent), child_watcher)
 }
 
-impl OnResize for Winpty {
+impl OnResize for WinptyResizeHandle {
     fn on_resize(&mut self, sizeinfo: &SizeInfo) {
         let (cols, lines) = (sizeinfo.cols().0, sizeinfo.lines().0);
         if cols > 0 && cols <= u16::MAX as usize && lines > 0 && lines <= u16::MAX as usize {
-            self.set_size(cols as u16, lines as u16)
+            // This unsafe cast from const to non-const is actually thread-safe because
```
chrisduerr:
```suggestion
            // This unsafe cast from const to non-const is thread-safe because
```

Just a super, super minor thing, but I think the 'actually' is a bit strange here. We're just stating facts here, we're not trying to prove anyone wrong.

alacritty_terminal/src/tty/windows/conpty.rs
davidhewitt: > Not checking if this is closed causes the deadlock, right? Instead of failing with an error?

I'm not sure what calling `ResizePseudoConsole` on a closed pseudoconsole might do. It might just fail with an error, or it might cause some kind of undefined behaviour. The [docs for ResizePseudoConsole`](https://docs.microsoft.com/en-us/windows/console/resizepseudoconsole) just state the parameter should be "an active pseudoconsole".

I could experiment with removing this check and just ignore any error from the `ResizePseudoConsole` call. But I think the way I've currently set it up is more defensive against the untrusted C++ code :)

alacritty_terminal/src/tty/windows/mod.rs
davidhewitt: There's two nearly identical copies of this comment; I removed this and left the other one: https://github.com/jwilm/alacritty/blob/5f4f253cacbe29c1b473004cfeeb1b2caf25f913/alacritty_terminal/src/tty/windows/mod.rs#L121

We also have an issue: #1905

alacritty_terminal/src/tty/windows/conpty.rs
chrisduerr: > I could experiment with removing this check and just ignore any error from the ResizePseudoConsole call.

I'd definitely like to check. If it just reports an error that the file has been closed, that might make things easier for us.

alacritty_terminal/src/tty/windows/conpty.rs
davidhewitt: > Would it not be possible to implement Drop for PTY and upholding the drop order in there? I still absolutely cannot stand that we move the conin and conout on the WinPTY/ConPTY for no good reason at all.

I think I'm failing to explain the deadlock very well. The [docs for `ClosePseudoConsole`](https://docs.microsoft.com/en-us/windows/console/closepseudoconsole) describe what I believe is the issue:

> A final painted frame may arrive on hOutput from the pseudoconsole when this API is called. It is expected that the caller will drain this information from the communication channel buffer and either present it or discard it. Failure to drain the buffer may cause the Close call to wait indefinitely until it is drained or the communication channels are broken another way.

So we need to ensure that the `output` pipe is still open when `ClosePseudoConsole` is called, and also we need to then continuously read from it until it's empty. This will have to be done in another thread, because the main thread will be blocked by `ClosePseudoConsole`. The `EventedAnonRead` type actually uses a reader thread and an internal buffer, so this is solved for us as long as `conout` is still alive when `ClosePseudoConsole` is called.

Going back to this point:

> Would it not be possible to implement Drop for PTY and upholding the drop order in there?

Yes, we could definitely do that. It would lead to less code. However:

- From my observations, we ***need*** to call `ClosePseudoConsole` or the pseudoconsole is leaked after alacritty closes. If it leaks, the conhost pseudoconsole process (and child shell/process tree) becomes a system process which never closes until the computer shuts down.
- My worry is that if `ClosePseudoConsole` call isn't in `Conpty`'s drop implementation, then we are dependent on other code explicitly closing the pseudoconsole to avoid these leaks. (i.e. I'm doing my best to stick to RAII for the `Conpty` opening & closing pseudoconsoles.)

The frustration is this potential deadlock in `ClosePseudoConsole`. To ensure that `conout` is still alive when `ClosePseudoConsole` is called, and to stick to RAII and call `ClosePseudoConsole` as part of `Drop for Conpty`, it was necessary to give `Conpty` ownership of the `conout` pipe.

I don't love that it leads to more code either, but I think that making the `Conpty` type robust is worth the price.

I'm ok if you prefer shorter code and instead accept that misuse of the `Conpty` type might cause deadlocks in future, I'll rework this PR to just `impl Drop for Pty` and be done with it 😄

chrisduerr: > My worry is that if ClosePseudoConsole call isn't in Conpty's drop implementation, then we are dependent on other code explicitly closing the pseudoconsole to avoid these leaks. (i.e. I'm doing my best to stick to RAII for the Conpty opening & closing pseudoconsoles.)

If we implement the drop on ConPTY and call `ClosePseudoConsole` there, we should be safe. The only thing we need to make sure of then is that we do not free the other code before it. But we do not risk leaking anything.

After that is ensured, the only thing we need to check is that the `ConPTY` handle is dropped last in the `Pty` drop implementation. We could ensure that simply by explicitly dropping the handle first and letting Rust automatically close conout/conin (we could also explicitly close those afterwards).

As far as I can tell, this would make everyone drop the things it needs to have 'knowledge' about, while at the same time we still only have a single point where we would need to ensure ordering. I don't see right now how there's difference between ensuring this in the drop for Pty vs ConPTY, please let me know if there is any.

The additional benefit of this strategy would be, that if there is a ConPTY2 some time in the future, we're already on the safe side, because we make sure our `conout`/`conin` are always dropped last, so we will drain the buffers.

> I'm ok if you prefer shorter code and instead accept that misuse of the Conpty type might cause leaks/deadlocks in future.

I don't see how the misuse of the `ConPTY` type could cause leaks here, assuming we close the PTY in its `Drop` implementation. It would be possible to cause a deadlock if we rewrite the `PTY` module and forget about it, but the same deadlock would appear if we rewrite the `ConPTY` module in your PR and forget about it. So as far as I can tell the issue is still the same, just on a different level.

I guess I just don't see what we gain by rewriting things, other than making the code more complicated? There's no doubt the Windows code is already extremely messy and difficult to maintain, so I'd like to stay away from making it even harder.

alacritty_terminal/src/tty/windows/conpty.rs
davidhewitt: If the pseudoconsole is closed, it appears `ResizePseudoConsole` returns `E_HANDLE`. So we can work with that.

davidhewitt:
I'm going to rewrite this PR one final time.

Issue #3072, opened 2019-12-06
Consider adding /Applications symlink into alacritty dmg


mikroskeem:
That would make installing Alacritty even quicker, as then you'd be able to simply drag Alacritty app in a single window to get it up and running.

Why? Everybody else seems to do that, and throwing in `ln -s /Applications Applications` into dmg generation shouldn't be an hard task :smile:

OS is MacOS, if someone could not figure out from the suggestion text yet.

chrisduerr:
I don't have a macOS machine around to test this at the moment, but since we're talking about symlinks here, wouldn't that mean the source file is required to stay around?

I feel like that would just be annoying and something that always has to be removed anyways, usually you just want to copy it over properly, instead of keeping it around in some random folder.

mikroskeem:
Not sure if you understood me correctly, but I mean something like this:

![image](https://user-images.githubusercontent.com/3490861/70358385-e4ff3600-1881-11ea-997f-db2236d26f14.png)

chrisduerr:
Okay, so you're not necessarily talking about actually symlinking, but instead just want the .dmg to automatically install to the Applications folder?

I'd assume that would require some kind of installer to be packaged within the .dmg, at least that was the case for the few .dmgs I've installed manually (everything is installed through brew mostly).

It definitely seems like something that can be evaluated, we have an installer on Windows too after all. But it would probably depend on complexity if we actually want that. Since it's macOS specific it might also take a while, but I think we can definitely give it a shot.

mikroskeem:
Not quite an automatic installer. But simply an symlink **inside** the dmg to avoid opening another Finder window.

I can try to PR this at some point.

mikroskeem:
That's what I meant:

![image](https://user-images.githubusercontent.com/3490861/70359072-9f436d00-1883-11ea-86ba-78693c41bf7f.png)

Now I can simply drag and drop the Alacritty.app into /Applications folder

chrisduerr:
Ah gotcha, that would significantly make things easier, yes. Should be a trivial change then and seems like it would make sense.

chrisduerr:
Fixed in #3077.

PR #3077, opened 2019-12-06
Create /Applications symlink inside dmg for easier installation

mikroskeem:
Solves #3072

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Terminal escape bindings with combined modifiers for Delete and Insert
+- /Applications symlink into OS X DMG for easier installation
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


mikroskeem:
In a few seconds...

mikroskeem:
There you go.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Packaging
 
 - Added compatibility logo variants for environments which can't render the default SVG
+- Added /Applications symlink into OS X DMG for easier installation
```
chrisduerr: Packaging is usually about changes that affect packagers, I've thought about adding it there too, but I think the `Added` section makes more sense.

The though is right, I just don't think it's the best place in this particular PR, since it has no effect on packagers at all (brew doesn't use the dmg).

mikroskeem: Ah, alright. I thought It's about generic packaging and `Added` is more about functional changes.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Terminal escape bindings with combined modifiers for Delete and Insert
+- Added /Applications symlink into OS X DMG for easier installation
```
```suggestion
- Symlink to /Applications in macOS DMG for easier installation
```

We usually try not to prepend `Added` whenever possible, otherwise the entire `### Added` section would look like `- Added ...` which is just unnecessary noise.

PR #3078, opened 2019-12-07
feature - Draw chart metrics in alacritty

sebosp:
First, thanks so much for the terminal, it is amazing, I use it everyday as my main terminal and I live in a terminal (I'm an SRE).

### Disclaimer
I used this as a learning playground for Rust/etc. 
I tried to make daily small commits, might be a bit annoying to watch the history.
I'm not a real dev and I'm not sure I should be even creating this PR, it is inconsistent with the ideas of making a blazing fast terminal, I just found it useful for my day-to-day live on a terminal.
I think this could become an external "module" or "addon" but I'm not sure how to do that.

### Rationale
I wanted to show graphically in my terminal bottom the status of some computer counters (load/memory/disk free/alerts), I use OSX and Linux and I didn't want to create measure techniques for either separately (i.e. proc_info crate only worked on Linux), so I thought of finding something that deals with the different OSs and gives me the metrics and went for Prometheus (I use it a lot for kuberenetes clusters I babysit).

### How it works:
- A background thread is started for Tokio, and the Charts definition and state are managed there.
- The tokio thread uses Intervals to load the data from remote resources (http or https endpoiunts) and calculates the OpenGL vertices.
- The terminal main thread asks the background thread for the opengl representation and then it is rendered.

I tried to include some documentation as well in alacritty_charts/README.md

Please let me know if there's any interest on merging this or if this should become an "addon/module", otherwise I will continue with my fork. Again, thanks for all the great work!


chrisduerr:
First of all, thanks for showing off what you've been working on. It's always cool not only to see people hack on Alacritty, but doing stuff related to it too.

To address the elephant in the room, Alacritty is not the right place for something like this.

Alacritty is a terminal emulator and because of that, its job is to emulate terminals, not to draw complex graphs about system metrics. Even things like image rendering inside of terminals are treated with caution, since often it's not actually a good idea to use them (terminals fundamentally work differently).

I think this would be a better project for maybe a system bar, if you constantly want to give an overview over certain stats of other systems, or maybe a CLI application rendering inside of Alacritty. But Alacritty is definitely not the right place, but thanks for sharing!

PR #3079, opened 2019-12-07
Transition font crate to 2018 edition

sterlingjensen:
In addition to implementing the path changes, dependencies are now more strictly specified, and a light linting was performed.


font/src/darwin.rs
```diff
@@ -40,14 +40,49 @@ use core_text::font_descriptor::SymbolicTraitAccessors;
...
+    #[cfg(target_endian = "big")]
+    pub fn extract_rgb(bytes: Vec<u8>) -> Vec<u8> {
+        bytes
+            .into_iter()
+            .enumerate()
+            .filter(|&(index, _)| ((index) % 4) != 0)
+            .map(|(_, val)| val)
+            .collect::<Vec<_>>()
+    }
+}
```
chrisduerr: What's the intention behind moving this into the default module? Usually modules in the same file are not desired except for test modules.

font/src/ft/fc.rs
```diff
@@ -11,42 +11,34 @@
...
+use std::{fmt, ptr};
+
+use fontconfig::fontconfig::{
+    FcResultNoMatch, {FcFontList, FcFontMatch, FcFontSort},
+    {FcMatchFont, FcMatchPattern, FcMatchScan}, {FcSetApplication, FcSetSystem},
+    {FC_SLANT_ITALIC, FC_SLANT_OBLIQUE, FC_SLANT_ROMAN},
+    {FC_WEIGHT_BLACK, FC_WEIGHT_BOLD, FC_WEIGHT_EXTRABLACK, FC_WEIGHT_EXTRABOLD},
+    {FC_WEIGHT_BOOK, FC_WEIGHT_MEDIUM, FC_WEIGHT_REGULAR, FC_WEIGHT_SEMIBOLD},
+    {FC_WEIGHT_EXTRALIGHT, FC_WEIGHT_LIGHT, FC_WEIGHT_THIN},
+};
```
chrisduerr: Can't we remove all brackets in here?

font/src/ft/fc.rs
```diff
@@ -11,42 +11,34 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-//
-use std::fmt;
-use std::ptr;
-
+use std::{fmt, ptr};
+
+use fontconfig::fontconfig::{
```
chrisduerr: Having a `fontconfig::fontconfig` import is very confusing, which is why the original version had `use fontconfig::fontconfig as ffi` as an easier to understand import. I think that makes a lot of sense and makes it a ton easier to understand what exactly we're importing here. So I'd be in favor of changing that back.

font/src/ft/fc/pattern.rs
```diff
@@ -11,23 +11,23 @@
...
-use std::str;
-
-use foreign_types::{ForeignType, ForeignTypeRef};
+use std::{
+    ffi::{CStr, CString},
+    fmt, mem,
+    path::PathBuf,
+    ptr::{self, NonNull},
+    str,
+};
```
chrisduerr: We do not use nested imports in Alacritty like this. They're too difficult to read without much benefits at all.

font/src/ft/fc.rs
```diff
@@ -11,42 +11,34 @@
...
-use self::ffi::{FC_WEIGHT_BOOK, FC_WEIGHT_MEDIUM, FC_WEIGHT_REGULAR, FC_WEIGHT_SEMIBOLD};
-use self::ffi::{FC_WEIGHT_EXTRALIGHT, FC_WEIGHT_LIGHT, FC_WEIGHT_THIN};
-
-pub mod config;
-pub use self::config::{Config, ConfigRef};
+pub use char_set::{CharSet, CharSetRef};
+pub use config::{Config, ConfigRef};
+pub use font_set::{FontSet, FontSetRef};
+pub use object_set::{ObjectSet, ObjectSetRef};
+pub use pattern::{Pattern, PatternRef};
```
chrisduerr: This is done incorrectly in the latest master too, but generally we do not want to use `pub use` excessively. Imports like `self::Pattern` make it extremely difficult to understand where things are coming from. Meanwhile `pattern::Pattern` is without a doubt easier to understand.

font/src/ft/fc/pattern.rs
```diff
@@ -528,7 +528,7 @@ impl PatternRef {
         unsafe { self.add_integer(b"rgba\0", rgba.to_isize()) }
     }
 
-    pub fn render_prepare(&self, config: &ConfigRef, request: &PatternRef) -> Pattern {
+    pub fn render_prepare(&self, config: &ConfigRef, request: &PatternRef) -> self::Pattern {
```
chrisduerr: There's three instances of `self::Pattern` in this file, however the module itself is `pattern.rs`. Shouldn't the `Pattern` be defined in this file, thus not needing any prefix at all?

font/src/darwin.rs
sterlingjensen: My intention was to simplify the pre-2018 mod structuring (folder, ambiguous mod.rs file, meaningfully named submodules), while leaving the module path structure unchanged. I can only guess at the original intent, likely a combination of crate constraints and unrealized future plans that needed the scoping. As I can't test for macos before pushing, I'm reluctant to mess around with it more than necessary.

font/src/ft/fc.rs
sterlingjensen: I'd be happy to, I just thought it safer to preserve the grouping - as it seemed to me that someone wanted them grouped logically instead of alphabetically.

font/src/ft/fc.rs
sterlingjensen: I was actually tempted to rebind it to the name of the crate it came from. I'll add ffi back in and bind off of that.

font/src/ft/fc/pattern.rs
sterlingjensen: IntelliJ must be back to its old tricks after the update, I'll correct that.

font/src/ft/fc.rs
sterlingjensen: I've got no problem with that, just trying to preserve the existing paths. Do you want me to refactor the visibility in this merge request? Because I was already getting the feeling that I had touched too many files.

font/src/ft/fc/pattern.rs
sterlingjensen: I thought so as well, and tried to figure out what the problem was - before I gave up. Rustc is clever enough that the prefix is unneeded for compilation. but I have found RLS and IntelliJ to be much less clever about macro expansion (my guess at the root cause, the foreign_type macro). This was the case on both Windows and Debian, stable and nightly. So should I go with the syntactically correct original, or the practical existing submission?

font/src/darwin.rs
chrisduerr: > ambiguous mod.rs file

What do you mean by this? How is a `mod.rs` file ever ambiguous?

Modules should never be in the same file without a good reason for it. That's why it originally was put into a different module, since it's unrelated code that has a nice separation from the primary module.

font/src/ft/fc.rs
chrisduerr: The current grouping does nothing, so there's no point in keeping it.

font/src/ft/fc.rs
chrisduerr: Feel free to do either, I'd understand not wanting to touch any existing code if the goal is just to transition to a new edition.

However when it comes to style improvements, I'm always willing to make exceptions since these should just improve codestyle in general.

font/src/ft/fc/pattern.rs
chrisduerr: We do not need to cater to RLS and IntelliJ, our code is compiled by rustc. So if the IDEs and editors don't understand it, that's their problem not ours.

font/src/darwin.rs
sterlingjensen: `mod.rs` files require a path in order to be distinguished from one another, this was sufficiently annoying to enough people that a new naming scheme was offered.

I'm pretty surprised about the mod aversion, it seems like the perfect application in this case - grouping a small amount of like code in the one file it is used, a grep of this crate's dependencies shows hundreds of non-test mod blocks. I'll revert the darwin directory.

font/src/ft/fc/pattern.rs
sterlingjensen: Ok, I'll revert the catering.

font/src/darwin.rs
chrisduerr: The only reason why you would ever go without a `mod.rs` file is when you have only a single module. We have more than one module here, so there's absolutely no point in trying to cram everything into one massive file.

The `mod.rs` file still has its purpose for structures with more than a single module in one directory.

font/src/ft/fc.rs
```diff
@@ -11,42 +11,35 @@
...
+mod config;
+mod font_set;
+mod object_set;
+mod pattern;
 
 /// Find the font closest matching the provided pattern.
 ///
 /// The returned pattern is the result of Pattern::render_prepare.
-pub fn font_match(config: &ConfigRef, pattern: &mut PatternRef) -> Option<Pattern> {
+pub fn font_match(config: &ConfigRef, pattern: &mut PatternRef) -> Option<self::Pattern> {
```
chrisduerr:
```suggestion
pub fn font_match(config: &ConfigRef, pattern: &mut PatternRef) -> Option<Pattern> {
```

font/src/ft/fc/pattern.rs
```diff
@@ -18,16 +18,15 @@ use std::path::PathBuf;
 use std::ptr::{self, NonNull};
 use std::str;
 
-use foreign_types::{ForeignType, ForeignTypeRef};
+use fontconfig::fontconfig::{
+    FcBool, FcChar8, FcConfigSubstitute, FcDefaultSubstitute, FcFontRenderPrepare, FcPattern,
+    FcPatternAddCharSet, FcPatternAddDouble, FcPatternAddInteger, FcPatternAddString,
+    FcPatternCreate, FcPatternDestroy, FcPatternGetBool, FcPatternGetDouble, FcPatternGetInteger,
+    FcPatternGetString, FcPatternPrint, FcResultMatch,
+};
```
chrisduerr: Here we use `fontconfig::fontconfig` again, instead of `super::ffi`, which is inconsistent.

Generally I think that these `fontconfig::fontconfig` imports are an insane mess. We have cross-module depedencies for the sole reason of using a crate import rename which seems to originate from the old edition, but just doesn't work well with the new 2018 edition.

I think the reasonable thing to do would be to import `fontconfig::fontconfig` everywhere. There's just no way around that and it's just imports so it shouldn't matter too much. If we ever import `fontconfig::fontconfig::self` anywhere, then we can rename that to `fontconfig::fontconfig::self as fontconfig_ffi`.

However we should never use `super::fontconfig::fontconfig` or `self::fontconfig::fontconfig` anywhere, since that just unreasonably complicates the cross-module depedencies for no reason.

Sorry for going back and forth on this, but if we want to have a consistent code style, we need to make sure it's intuitive and `super::ffi::` definitely is not that.

font/src/ft.rs
```diff
@@ -21,6 +21,7 @@ use std::path::PathBuf;
 use freetype::tt_os2::TrueTypeOS2Table;
```
chrisduerr: This file needs to go back to `font/src/ft/mod.rs`.

font/src/directwrite.rs
```diff
@@ -13,10 +13,7 @@
 // limitations under the License.
```
chrisduerr: This file should go back to `font/src/directwrite/mod.rs` for platform consistency.

font/src/ft/fc.rs
```diff
@@ -11,42 +11,35 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
chrisduerr: This file needs to go back to `font/src/ft/fc/mod.rs`.

sterlingjensen:
With all the suggested changes, a single 2018 edition improvement remains... a solitary crate scope absolute path. It would be inaccurate to describe this as a "transition" when platform consistency means 2015 conventions.

chrisduerr:
> It would be inaccurate to describe this as a "transition" when platform consistency means 2015 conventions.

Only because it is supported to have both `foo.rs` and the `foo` directory, does not mean that the `foo/mod.rs` structure is deprecated.

Putting things into the same path is a far cleaner solution, regardless of 2018 edition or not and having both the directory and the module on the top level does not scale.

sterlingjensen:
I'd say that two birds could be killed with one stone, simplified paths and cleaned up `pub use`s, but your justification for `byte_order.rs` similarly justifies all those interlinked `fc` sub-modules.

font/src/darwin.rs
nixpulvis: FWIW, I don't think `mod.rs` is ambiguous, however it is somewhat _redundant_, which is why the Rust team has added the new resolution in 2018. If I had to choose myself, I'd be in favor of the old style. The main justification is that it's consistent with `lib.rs`.

For small personal projects I can see using the new style, but I think I agree we shouldn't make this change.

nixpulvis:
@sterlingjensen I'd be happy to see this finished up and merged. Would you be interested in continuing it with the changes @chrisduerr has suggested? Or should I try and complete it?

sterlingjensen:
I don't think I should be touching this part of the code base, as it appears I'm in the minority on some very fundamental issues :upside_down_face: Please feel free to salvage whatever you can.

PR #3080, opened 2019-12-07
Improve error handling for Winpty

### LINKED ISSUE #2344: Better errors for WinPTY
In the https://github.com/jwilm/alacritty/pull/2261 PR the ConPTY errors have been improved to make it obvious why a shell could not be started. However, the WinPTY messages are unchanged and there currently is no output at all when spawning the WinPTY failed.

It should be made clear to the user that the shell could not be found when starting Alacritty with an invalid shell.

Based on the little research I've done, it seems like there's no place in Alacritty to catch this error and it's swallowed further down in the WinPTY dependency chain. Maybe @zacps knows something about this, but I'd assume that we might have to fix something in the `winpty-sys` crate.

davidhewitt:
Closes #2344

I've managed to make it such that failures to spawn the shell now result in a proper error message being displayed. Turns out the `winpty` crate had a TODO item to plumb through the appropriate error in the `Winpty::spawn` method :smile:

Crashes on the Winpty backend now also display error descriptions such as `SpawnCreateProcessFailed` instead of the Winpty internal error codes (e.g. `2`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Width of Unicode 11/12 emojis
 - Minimize on windows causing layout issues
 - Performance bottleneck when clearing colored rows
-- Lack of error message explaining why shell could not start on Windows (when using Winpty backend)
+- Vague startup crash messages on Windows with WinPTY backend
 
 ## 0.4.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Width of Unicode 11/12 emojis
 - Minimize on windows causing layout issues
 - Performance bottleneck when clearing colored rows
+- Lack of error message explaining why shell could not start on Windows (when using Winpty backend)
 
 ## 0.4.0
 
--- a/winpty/src/windows.rs
+++ b/winpty/src/windows.rs
@@ -20,7 +20,7 @@ pub enum ErrorCode {
     AgentDied,
     AgentTimeout,
     AgentCreationFailed,
-    UnknownError(u32)
+    UnknownError(u32),
 }
 
 pub enum MouseMode {
--- a/winpty/src/windows.rs
+++ b/winpty/src/windows.rs
@@ -94,6 +94,7 @@ impl Config {
         let mut err = null_mut() as *mut winpty_error_t;
         let config = unsafe { winpty_config_new(flags.bits(), &mut err) };
         check_err(err)?;
+
         Ok(Self(config))
     }
 
@@ -152,6 +153,7 @@ impl Winpty {
         let mut err = null_mut() as *mut winpty_error_t;
         let winpty = unsafe { winpty_open(cfg.0, &mut err) };
         check_err(err)?;
+
         Ok(Self(winpty))
     }
 
@@ -223,6 +225,7 @@ impl Winpty {
         }
 
         check_err(err)?;
+
         Ok(process_list)
     }
 
@@ -304,6 +307,7 @@ impl SpawnConfig {
         };
 
         check_err(err)?;
+
         Ok(Self(spawn_config))
     }
 }
--- a/winpty/src/windows.rs
+++ b/winpty/src/windows.rs
@@ -9,7 +9,8 @@ use winpty_sys::*;
 
 use widestring::WideCString;
 
-pub enum ErrorCodes {
+#[derive(Clone, Copy, Debug)]
+pub enum ErrorCode {
     Success,
     OutOfMemory,
     SpawnCreateProcessFailed,
@@ -19,6 +20,7 @@ pub enum ErrorCodes {
     AgentDied,
     AgentTimeout,
     AgentCreationFailed,
+    UnknownError(u32)
 }
 
 pub enum MouseMode {
@@ -44,7 +46,7 @@ bitflags!(
 
 #[derive(Debug)]
 pub struct Err {
-    code: u32,
+    code: ErrorCode,
     message: String,
 }
 
@@ -55,16 +57,30 @@ fn check_err(e: *mut winpty_error_t) -> Option<Err> {
         let raw = winpty_error_msg(e);
         let message = String::from_utf16_lossy(std::slice::from_raw_parts(raw, wcslen(raw)));
         winpty_error_free(e);
+
+        let code = match code {
+            WINPTY_ERROR_SUCCESS => ErrorCode::Success,
+            WINPTY_ERROR_OUT_OF_MEMORY => ErrorCode::OutOfMemory,
+            WINPTY_ERROR_SPAWN_CREATE_PROCESS_FAILED => ErrorCode::SpawnCreateProcessFailed,
+            WINPTY_ERROR_LOST_CONNECTION => ErrorCode::LostConnection,
+            WINPTY_ERROR_AGENT_EXE_MISSING => ErrorCode::AgentExeMissing,
+            WINPTY_ERROR_UNSPECIFIED => ErrorCode::Unspecified,
+            WINPTY_ERROR_AGENT_DIED => ErrorCode::AgentDied,
+            WINPTY_ERROR_AGENT_TIMEOUT => ErrorCode::AgentTimeout,
+            WINPTY_ERROR_AGENT_CREATION_FAILED => ErrorCode::AgentCreationFailed,
+            code => ErrorCode::UnknownError(code)
+        };
+
         match code {
-            0 => None,
-            _ => Some(Err { code, message }),
+            ErrorCode::Success => None,
+            code => Some(Err { code, message }),
         }
     }
 }
 
 impl Display for Err {
     fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
-        write!(f, "Code: {}, Message: {}", self.code, self.message)
+        write!(f, "Code: {:?}, Message: {}", self.code, self.message)
     }
 }
 
@@ -95,12 +111,12 @@ impl Config {
     /// Set the mouse mode
     pub fn set_mouse_mode(&mut self, mode: &MouseMode) {
         let m = match mode {
-            MouseMode::None => 0,
-            MouseMode::Auto => 1,
-            MouseMode::Force => 2,
+            MouseMode::None => WINPTY_MOUSE_MODE_NONE,
+            MouseMode::Auto => WINPTY_MOUSE_MODE_AUTO,
+            MouseMode::Force => WINPTY_MOUSE_MODE_FORCE,
         };
         unsafe {
-            winpty_config_set_mouse_mode(self.0, m);
+            winpty_config_set_mouse_mode(self.0, m as i32);
         }
     }
 
@@ -126,6 +142,21 @@ impl Drop for Config {
 #[derive(Debug)]
 /// A struct representing the winpty agent process
 pub struct Winpty(*mut winpty_t);
+
+pub struct ChildHandles {
+    process: HANDLE,
+    thread: HANDLE
+}
+
+impl ChildHandles {
+    pub fn process(&self) -> HANDLE {
+        self.process
+    }
+
+    pub fn thread(&self) -> HANDLE {
+        self.thread
+    }
+}
 
 impl Winpty {
     /// Starts the agent. This process will connect to the agent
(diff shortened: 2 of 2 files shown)


sterlingjensen:
It only took two days for @chrisduerr to be proven right about those map methods :blush:

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Direct escape input on Windows using alt
 - Incorrect window size on X11 when waking up from suspend
 - Incorrect width of Unicode 11/12 emojis
+- Lack of error message explaining why shell could not start on Windows (when using Winpty backend)
```
chrisduerr: This line is a bit long, maybe we can make things a bit more precise and get down to the interesting parts quicker:

```suggestion
- Vague startup crash messages on Windows with WinPTY backend
```

winpty/src/windows.rs
```diff
@@ -19,6 +20,7 @@ pub enum ErrorCodes {
     AgentDied,
     AgentTimeout,
     AgentCreationFailed,
+    UnknownError(u32)
```
chrisduerr:
```suggestion
    UnknownError(u32),
```

Did you run `rustfmt` on this? I thought trailing commas are added automagically.

winpty/src/windows.rs
```diff
@@ -9,7 +9,8 @@ use winpty_sys::*;
 
 use widestring::WideCString;
 
-pub enum ErrorCodes {
+#[derive(Clone, Copy, Debug)]
```
chrisduerr: I usually try to do `Copy, Clone` instead of `Clone, Copy`. Because the `Clone` constraint is just necessary for silly reasons, the interesting bit is `Copy`. So that should be what the reader sees first, the clone part is implied anyways.

winpty/src/windows.rs
```diff
@@ -55,16 +57,30 @@ fn check_err(e: *mut winpty_error_t) -> Option<Err> {
...
+            WINPTY_ERROR_AGENT_TIMEOUT => ErrorCode::AgentTimeout,
+            WINPTY_ERROR_AGENT_CREATION_FAILED => ErrorCode::AgentCreationFailed,
+            code => ErrorCode::UnknownError(code)
+        };
+
         match code {
-            0 => None,
-            _ => Some(Err { code, message }),
+            ErrorCode::Success => None,
+            code => Some(Err { code, message }),
```
chrisduerr: I love matching against `ErrorCode::Success` here, instead of `0`. Great change!

winpty/src/windows.rs
```diff
@@ -55,16 +57,30 @@ fn check_err(e: *mut winpty_error_t) -> Option<Err> {
...
+            WINPTY_ERROR_OUT_OF_MEMORY => ErrorCode::OutOfMemory,
+            WINPTY_ERROR_SPAWN_CREATE_PROCESS_FAILED => ErrorCode::SpawnCreateProcessFailed,
+            WINPTY_ERROR_LOST_CONNECTION => ErrorCode::LostConnection,
+            WINPTY_ERROR_AGENT_EXE_MISSING => ErrorCode::AgentExeMissing,
+            WINPTY_ERROR_UNSPECIFIED => ErrorCode::Unspecified,
+            WINPTY_ERROR_AGENT_DIED => ErrorCode::AgentDied,
+            WINPTY_ERROR_AGENT_TIMEOUT => ErrorCode::AgentTimeout,
+            WINPTY_ERROR_AGENT_CREATION_FAILED => ErrorCode::AgentCreationFailed,
+            code => ErrorCode::UnknownError(code)
+        };
```
chrisduerr: Maybe we could early return on `ErrorCode::Success` here? That way we wouldn't need to match `code` twice.

Alternatively, we could just return from this match statement directly, but that would lead to a ton of duplication with `Some(Err { code: ..., message })` which I do not think would look very nice.

winpty/src/windows.rs
```diff
@@ -95,12 +111,12 @@ impl Config {
     /// Set the mouse mode
     pub fn set_mouse_mode(&mut self, mode: &MouseMode) {
         let m = match mode {
-            MouseMode::None => 0,
-            MouseMode::Auto => 1,
-            MouseMode::Force => 2,
+            MouseMode::None => WINPTY_MOUSE_MODE_NONE,
+            MouseMode::Auto => WINPTY_MOUSE_MODE_AUTO,
+            MouseMode::Force => WINPTY_MOUSE_MODE_FORCE,
```
chrisduerr: Always nice to get rid of some magic numbers, thank you!

winpty/src/windows.rs
```diff
@@ -127,6 +143,21 @@ impl Drop for Config {
...
+
+impl ChildHandles {
+    pub fn process(&self) -> HANDLE {
+        self.process
+    }
+
+    pub fn thread(&self) -> HANDLE {
+        self.thread
+    }
+}
```
chrisduerr: I don't think there's much reason to have methods for accessing `process` and `thread` here, right? We can just make the fields `pub` and that shouldn't be a problem.

winpty/src/windows.rs
```diff
@@ -213,26 +244,37 @@ impl Winpty {
...
+            match err.code {
+                ErrorCode::SpawnCreateProcessFailed => Result::Err(Err {
+                    code: err.code,
+                    message: format!("{} (error code {})", err.message, create_process_error)
+                }),
+                _ => Result::Err(err),
+            }
+        } else {
+            Ok(handles)
+        }
```
chrisduerr: There are a few ways to rewrite this code to get rid of the nested matching, but I'd like to know for now why we don't always print the full error? Is there any disadvantage to that? I don't see why we would catch just spawnprocess errors with nice messages?

winpty/src/windows.rs
```diff
@@ -262,33 +304,26 @@ impl SpawnConfig {
...
         let mut err = null_mut() as *mut winpty_error_t;
-        let (appname, cmdline, cwd, end) = (
-            appname.map_or(null(), |s| WideCString::from_str(s).unwrap().into_raw()),
-            cmdline.map_or(null(), |s| WideCString::from_str(s).unwrap().into_raw()),
-            cwd.map_or(null(), |s| WideCString::from_str(s).unwrap().into_raw()),
-            end.map_or(null(), |s| WideCString::from_str(s).unwrap().into_raw()),
-        );
+
+        let to_wstring = |s| WideCString::from_str(s).unwrap();
+        let as_ptr = |opt: &Option<WideCString>| opt.as_ref().map_or(null(), |ws| ws.as_ptr());
```
chrisduerr: These two can be moved closer to the code where they're actually used.

`to_wstring` can be moved right above `let appname =` while the `as_ptr` can be moved right above `spawn_config`.

The `as_ptr` method could also be renamed a bit better. Maybe something more like `wstring_ptr`.

That way we don't accidentally write cobol code where everything's defined at the top and then never used until the bottom of the program is reached.

winpty/src/windows.rs
davidhewitt: :D indeed, I saw the constants existed and felt I had to use them!

winpty/src/windows.rs
davidhewitt: Ok 👍

winpty/src/windows.rs
davidhewitt: If spawning the shell fails, WinPTY sends us back the result of Window's `GetLastError` call. It's called `create_process_error` in this code.

I felt it would be desirable to store this information in the `Err` struct. The simplest way to do that was to add this information to the format string.

In summary, this code is simply forwarding all errors except for `SpawnCreateProcessFailed`, where it's rewriting the `message` contents to add additional information.

Happy to rewrite this if you've got a preferred way the additional information should be stored.

winpty/src/windows.rs
davidhewitt: 👍 - though this code will now die to eliminate the double-match 😄

winpty/src/windows.rs
davidhewitt: I'm going to change the `WINPTY_ERROR_SUCCESS` branch to `return None`, and then the second match can just be rewritten `Some(Err { code, message })`.

I'll also delete the `ErrorCode::Success` variant, as it will now be unused.

davidhewitt:
Thanks for the code review. I've fixed up most of what you pointed out (and run `rustfmt`, my bad). 

Just the `SpawnCreateProcessFailed` error match to clean up, once I know what you'd prefer.

winpty/src/windows.rs
chrisduerr: My only question is why we wouldn't always add this additional information?

davidhewitt: > My only question is why we wouldn't always add this additional information?

The `create_process_error` value is only meaningful in the `SpawnCreateProcessFailed` case.

Perhaps a better way is to change `ErrorCode::SpawnCreateProcessFailed` to contain this value?

chrisduerr: So you're saying that the `check_err` function always adds a message, but that message is never useful unless the error is `SpawnCreateProcessFailed`?

In that case we certainly should add the message only to that specific error, because it would prevent people from using this incorrectly, however that sounds a bit strange to me.

Why would `WINPTY_ERROR_SPAWN_CREATE_PROCESS_FAILED` have an error message, but none of the other errors does?

davidhewitt: > So you're saying that the check_err function always adds a message, but that message is never useful unless the error is SpawnCreateProcessFailed?

Not quite; All the winpty error codes have a message. (as far as I know)

However, for `SpawnCreateProcessFailed`, winpty also sends the OS error code back from the agent, completely separately to the message. It's this separate error code which is called `create_process_error` above, and only relevant to this error.

I'll modify things to add the code into `SpawnCreateProcessFailed` variant.

chrisduerr: Ah, I got you now. It's not really possible to add this to `SpawnCreateProcessFailed`, right? Since this is created in the `check_err` method and we wouldn't want to pass an optional field to that method.

chrisduerr: I probably would have just done the following in that case:

```rust
let result = cherk_err(err);
if let Err(Err { code: ErrorCode::SpawnCreateProcessFailed, message }) = &mut result {
    message = format!("{} (error code {})", message, create_process_error);
}
result
```

Alternatively `map_err` could work out I think, though then we would have a useless `err => err,` in one of the match arms.

I'd also change the return type of `check_err` to `Result<(), Err>`, because every place where we use it we map it through `Result::Err`, which seems mostly pointless.

davidhewitt: > Ah, I got you now. It's not really possible to add this to SpawnCreateProcessFailed, right? Since this is created in the check_err method and we wouldn't want to pass an optional field to that method.
> 
>Or how would you have achieved this?

I agree that adding an optional field to `check_err` isn't great. I had a couple of ideas brewing:

- Add `check_error_with_os_error`, which takes as an extra argument the `DWORD` value (of `create_process_error`). This would take on the existing `check_err` implementation as well as drop that `DWORD` into the `SpawnCreateProcessFailed` variant. `check_err` would then change to call into that and pass a `DWORD` value of `0`. Still quite messy, but at least all the rest of the functions don't have to write something like `check_err(err, None)`.
- Alternatively, `SpawnCreateProcessFailed` could contain an `Option<DWORD>`. `check_err` would put `None` in there. In the code above we could then put in the value of `create_process_error`.

But I think both of those create a fair amount of mess. So for the moment I'm going to just change the code in the way you suggest.

davidhewitt: Changing the return type of `check_err` worked nicely. I was able to change the following patterns:

    check_err(err).map_or(Ok(process_list), Result::Err)

into

    check_err(err)?;
    Ok(process_list)

which reads much nicer imo.

sterlingjensen: `
check_err(err)
            .map_or_else(||Ok(handles), |mut e|{
                if e.code == ErrorCode::SpawnCreateProcessFailed {
                    e.message += &format!(" (error code {})", create_process_error);
                }
                Err(e)
            })
`
Too cryptic? I think the add operation on the original message makes the purpose pretty clear, without raising questions about nesting error types...

chrisduerr: You can also write the following, as a replacement for the `map_or`:
```rust
check_err(err).map(|_| process_list)
```

> Too cryptic? I think the add operation on the original message makes the purpose pretty clear, without raising questions about nesting error types...

I don't think this suggestion has any benefit over the `if let` I've proposed, since you basically just add the `Err(e)` as additional requirement. It's pretty similar to just `map` in this regard.

The `+= &format` isn't really very idiomatic Rust. I've originally had that in my proposal too, but using `format!("{} (error code {})")` is just the far more common strategy in Rust at least as far as I know. If format is used at all, it's usually used to format everything.

There are a few other alternatives, but I think `format!`ing everything is just the best choice at that point.

winpty/src/windows.rs
```diff
@@ -260,36 +282,29 @@ impl SpawnConfig {
...
-                WideCString::from_raw(cwd as *mut u16);
-            }
-            if !end.is_null() {
-                WideCString::from_raw(end as *mut u16);
-            }
-        }
-
-        check_err(err).map_or(Ok(Self(spawn_config)), Result::Err)
+        check_err(err)?;
+        Ok(Self(spawn_config))
```
chrisduerr: Could we add an extra newline between the three instances where `check_err` now does an early return and we explicitly return `Ok(...)` ourselves?

Since the two aren't related anymore, but two completely separate operations, I'd like to separate that semantically.

Otherwise it would make more sense to use `check_err(err).map(|_| ...)`, but I think the idea of not mapping random data into an error is good.

davidhewitt:
Added those newlines :)

winpty/src/windows.rs
davidhewitt: > There are a few other alternatives, but I think format!ing everything is just the best choice at that point.

I agree. I also wondered about `+=`, but I think that'll lead to two allocations: one for `format!` and then another for `+=`. `format!`ing would lead to just one. A micro-optimisation, I know 😄

cup:
https://github.com/jwilm/alacritty/issues/2016

PR #3081, opened 2019-12-09
Fix minimize on Windows

### LINKED ISSUE #3047: Rendering issue with minimize windows
alacritty

Once I have minimized the window and restore, the contents alignment is broken

OS: Windows 10
Font: Hack Nerd Font
Font Size: 9.0

And also resizing is still not working properly

davidhewitt:
This is a potential solution for #3047.

Both WinPTY and ConPTY backends have issues when minimizing because a winit `Resize` event fires which sets the size to zero (and for various reasons both backends don't like this).

Alternative solutions:
- Change winit to not emit a `Resize` event when minimizing. (Perhaps emit some other event.)
- Add an API to winit to check if the window is minimized, and use that instead of checking `lsize.width` and `lsize.height`.

(Closes #3047)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Direct escape input on Windows using alt
 - Incorrect window size on X11 when waking up from suspend
 - Incorrect width of Unicode 11/12 emojis
+- Minimize on windows causing layout issues
 
 ## 0.4.0
 
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -486,6 +486,17 @@ impl<N: Notify> Processor<N> {
                 match event {
                     CloseRequested => processor.ctx.terminal.exit(),
                     Resized(lsize) => {
+                        #[cfg(windows)]
+                        {
+                            // Minimizing the window sends a Resize event with zero width and
+                            // height. But there's no need to ever actually resize to this.
+                            // Both WinPTY & ConPTY have issues when resizing down to zero size
+                            // and back.
+                            if lsize.width == 0.0 && lsize.height == 0.0 {
+                                return;
+                            }
+                        }
+
                         let psize = lsize.to_physical(processor.ctx.size_info.dpr);
                         processor.ctx.display_update_pending.dimensions = Some(psize);
                         processor.ctx.terminal.dirty = true;


chrisduerr:
> Change winit to not emit a Resize event when minimizing. (Perhaps emit some other event.)

This is not a winit thing, but a Windows thing.

Generally I don't love this, since it doesn't really fix any issue at all. It just ignores Windows events to try and skip out potential states that could cause problems. But generally resizing will lead to the exact same problems as minimizing, so the user doesn't gain much.

I'm inclined to accept this just to improve the situation, but I feel like that's just going to raise expectations to fix stuff we have no control over. This ultimately is not our problem, but should be fixed upstream instead.

davidhewitt:
> I'm inclined to accept this just to improve the situation, but I feel like that's just going to raise expectations to fix stuff we have no control over. This ultimately is not our problem, but should be fixed upstream instead.

By upstream are you referring to `winit` or ConPTY?

FWIW Windows Terminal avoids this minimising issue by not calling it's own `OnResize` when minimising: https://github.com/microsoft/terminal/blob/5dfc021d8ed24de749504713a2ad8a4bd0903a79/src/cascadia/WindowsTerminal/BaseWindow.h#L100

chrisduerr:
> By upstream are you referring to winit or ConPTY?

This is not a problem with `winit`. They're doing everything right. It's a Windows problem.

davidhewitt:
> Resizing to 1x1 should not be any problem at all, it works just fine on all other platforms.

Agreed. Unfortunately we now have to choose to wait for an upstream fix or introduce a workaround.

Given that this patch is relatively small, and the broken behaviour quite annoying and easy to encounter, I'd vote we (reluctantly) work around this ourselves. 

Btw it looks like v0.3.3 worked because of these bounds checks: https://github.com/jwilm/alacritty/blob/cc3e4d3d5cb2f658d6faef02e4216b9342451efc/alacritty_terminal/src/term/mod.rs#L1174

So we used to accidentally work around this issue 😬

chrisduerr:
> Agreed. Unfortunately we now have to choose to wait for an upstream fix or introduce a workaround.

> Given that this patch is relatively small, and the broken behaviour quite annoying and easy to encounter, I'd vote we (reluctantly) work around this ourselves.

Yeah, that accurately describes my feelings about it as well. LGTM.

nixpulvis:
Thank you for leaving a nice note explaining the situation here. This really helps test later to see if the workaround is still needed. ❤️

PR #3084, opened 2019-12-10
Fix deadlock with closing on Windows using Conpty

### LINKED ISSUE #3042: Window close "X" button freezes alacritty

davidhewitt:
This follows on one more attempt from #3054 and #3074.

This time we're implementing the correct drop order in `Drop for Pty`. To ensure the pseudoconsole is properly closed, I've added a `debug_assert!` to `Drop for Conpty`.

Fingers crossed that this implementation now looks passable :smile:

Fixes #3042

### DIFF
--- a/alacritty/src/main.rs
+++ b/alacritty/src/main.rs
@@ -232,7 +232,7 @@ fn run(window_event_loop: GlutinEventLoop<Event>, config: Config) -> Result<(), 
     // The fix is to ensure that processor is dropped first. That way, when io_thread (i.e. pty)
     // is dropped, it can ensure Conpty is dropped before the conout pipe in the pty drop order.
     //
-    // TODO: refactor the Windows codebase and remove this horrendous workaround!
+    // FIXME: Change PTY API to enforce the correct drop order with the typesystem.
     drop(processor);
 
     // Shutdown PTY parser event loop
--- a/alacritty/src/main.rs
+++ b/alacritty/src/main.rs
@@ -222,6 +222,19 @@ fn run(window_event_loop: GlutinEventLoop<Event>, config: Config) -> Result<(), 
     // Start event loop and block until shutdown
     processor.run(terminal, window_event_loop);
 
+    // This explicit drop is needed for Windows, ConPTY backend. Otherwise a deadlock can occur.
+    // The cause:
+    //   - Drop for Conpty will deadlock if the conout pipe has already been dropped.
+    //   - The conout pipe is dropped when the io_thread is joined below (io_thread owns pty).
+    //   - Conpty is dropped when the last of processor and io_thread are dropped, because
+    //     both of them own an Arc<Conpty>.
+    //
+    // The fix is to ensure that processor is dropped first. That way, when io_thread (i.e. pty)
+    // is dropped, it can ensure Conpty is dropped before the conout pipe in the pty drop order.
+    //
+    // TODO: refactor the Windows codebase and remove this horrendous workaround!
+    drop(processor);
+
     // Shutdown PTY parser event loop
     loop_tx.send(Msg::Shutdown).expect("Error sending shutdown to pty event loop");
     io_thread.join().expect("join io thread");
--- a/alacritty/src/main.rs
+++ b/alacritty/src/main.rs
@@ -226,8 +226,8 @@ fn run(window_event_loop: GlutinEventLoop<Event>, config: Config) -> Result<(), 
     // The cause:
     //   - Drop for Conpty will deadlock if the conout pipe has already been dropped.
     //   - The conout pipe is dropped when the io_thread is joined below (io_thread owns pty).
-    //   - Conpty is dropped when the last of processor and io_thread are dropped, because
-    //     both of them own an Arc<Conpty>.
+    //   - Conpty is dropped when the last of processor and io_thread are dropped, because both of
+    //     them own an Arc<Conpty>.
     //
     // The fix is to ensure that processor is dropped first. That way, when io_thread (i.e. pty)
     // is dropped, it can ensure Conpty is dropped before the conout pipe in the pty drop order.
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -90,7 +90,7 @@ pub type ConptyHandle = Arc<Conpty>;
 
 impl Drop for Conpty {
     fn drop(&mut self) {
-        // Caution! This will block until the conout pipe is drained. May cause deadlocks if the
+        // XXX: This will block until the conout pipe is drained. Will cause a deadlock if the
         // conout pipe has already been dropped by this point.
         //
         // See PR #3084 and https://docs.microsoft.com/en-us/windows/console/closepseudoconsole
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -90,6 +90,10 @@ pub type ConptyHandle = Arc<Conpty>;
 
 impl Drop for Conpty {
     fn drop(&mut self) {
+        // Caution! This will block until the conout pipe is drained. May cause deadlocks if the
+        // conout pipe has already been dropped by this point.
+        //
+        // See PR #3084 and https://docs.microsoft.com/en-us/windows/console/closepseudoconsole
         unsafe { (self.api.ClosePseudoConsole)(self.handle) }
     }
 }
--- a/alacritty_terminal/src/tty/windows/mod.rs
+++ b/alacritty_terminal/src/tty/windows/mod.rs
@@ -45,8 +45,8 @@ pub enum PtyHandle {
 }
 
 pub struct Pty {
-    // It is important for drop order that this handle is defined before conout. Drop for Conpty
-    // will deadlock if the conout pipe has already been dropped.
+    // XXX: Handle is required to be the first field, to ensure correct drop order. Dropping
+    // `conout` before `handle` will cause a deadlock.
     handle: PtyHandle,
     // TODO: It's on the roadmap for the Conpty API to support Overlapped I/O.
     // See https://github.com/Microsoft/console/issues/262
--- a/alacritty_terminal/src/tty/windows/mod.rs
+++ b/alacritty_terminal/src/tty/windows/mod.rs
@@ -45,6 +45,8 @@ pub enum PtyHandle {
 }
 
 pub struct Pty {
+    // It is important for drop order that this handle is defined before conout. Drop for Conpty
+    // will deadlock if the conout pipe has already been dropped.
     handle: PtyHandle,
     // TODO: It's on the roadmap for the Conpty API to support Overlapped I/O.
     // See https://github.com/Microsoft/console/issues/262
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Minimize on windows causing layout issues
 - Performance bottleneck when clearing colored rows
 - Vague startup crash messages on Windows with WinPTY backend
+- Deadlock on Windows when closing Alacritty using the title bar "X" button (ConPTY backend)
 
 ## 0.4.0
 


alacritty_terminal/src/tty/windows/mod.rs
```diff
@@ -306,3 +323,28 @@ impl EventedPty for Pty {
...
+                    let mut buf = [0u8; 1024];
+                    while !done.load(Ordering::SeqCst) {
+                        conout.read(&mut buf).unwrap();
+                    }
+                });
+
+                c.close();
+                done.store(true, Ordering::SeqCst);
+            })
+            .unwrap();
```
chrisduerr: Wait, why do we need crossbeam and a separate thread now? I thought you said that reordering the drop operations is sufficient since the pipes handle things internally.

We didn't need this previously, why do we need this now?

davidhewitt: We *can* live without spawning this thread. This is because, yes, the `EventedAnonRead` pipe is draining the pseudoconsole pipe into its own internal buffer.

But that's not guaranteed to always work:
- If we ever change the pipe implementation (e.g. implement #1905), then this might break.
- It's possible there will be so much to drain that the `EventedAnonRead` pipe won't be able to fit it all in it's internal buffer.

Using a scoped thread and explicitly draining the whole thing is intended to make the code robust against those situations. (I used crossbeam so that the scoped thread can reference `conout`.)

If you prefer I can remove this thread and leave a comment in the code documenting `EventedAnonRead` currently handles the draining issue.

davidhewitt: (This possible need for extra defensiveness only struck me this morning when preparing this PR)

chrisduerr: > It's possible there will be so much to drain that the EventedAnonRead pipe won't be able to fit it all in its internal buffer.

How? I thought that there's always only one event sent after closing it.

Pulling in crossbeam just for this seems absolutely crazy to me. Especially because I do not think this kind of complexity is a good fit for a destructor. I'd assume that this might actually make the code less robust, because it so, so much harder to reason about things.

davidhewitt: > How? I thought that there's always only one event sent after closing it.

All [the docs for `ClosePseudoConsole`](https://docs.microsoft.com/en-us/windows/console/closepseudoconsole) say is "a final painted frame". There's no guarantee how much data that is.

> Pulling in crossbeam just for this seems absolutely crazy to me. Especially because I do not think this kind of complexity is a good fit for a destructor. I'd assume that this might actually make the code less robust, because it so, so much harder to reason about things.

Ok. Let's just scrap this bit and go back to just fixing the drop order.

davidhewitt:
I've undone the `crossbeam` bit. Let me know if you want any more changes. Hopefully we're getting there.

---

(Just a quick note: I care about Alacritty and want to help out. Your review of PRs is important for ensuring all work fits the overall direction of Alacritty, and I think you're a very good and thorough reviewer.

The review on this chain of PRs has had a few strong rejections which felt a bit harsh given I'm volunteering my own time to fix bugs. Reading back I see phrases like "absolutely crazy", "this code is completely useless", "for no good reason at all", etc. I understood the original attempts were not a good fit. I'll gladly take feedback and direction to rethink what isn't right. I learn from taking on feedback. And there's little gained if a PR creates a mess for the next person.

It would be nice if the feedback could be a bit less... crushing... next time. I suspect this was a two-way thing; I was probably trying your patience somewhat by repeatedly offering up designs which were more complicated than you wanted them to be.

All that said, I did get the impression from these PRs that the opinion is the windows code is messy and hard to maintain. If you have notes on what needs to be cleaned up / simplified, I'd be willing to find some time to help do so.)

chrisduerr:
> The review on this chain of PRs has had a few strong rejections which felt a bit harsh given I'm volunteering my own time to fix bugs.

I'm sorry if it feels that way, but I don't think there's any way around that when there is so much stuff that needs to be redone in a PR. The fact of the matter is just that I declined a lot of your suggestions, it takes a lot not to get discouraged by that regardless of how those rejections are phrased.

> Reading back I see phrases like "absolutely crazy", "this code is completely useless", "for no good reason at all", etc.

I've tried to keep most of the more openly negative comments about the existing code rather than the changes you've made. I think it's important to state when code is difficult to work on because the existing code is problematic, otherwise it's often even more discouraging if the author gets the feeling that things should be easier than they are.

However in certain cases it's important for me personally to make stronger statements, just to make it clear which points can be discussed and which points are useless arguing about. If code is not acceptable the way it is, there's no point in sugarcoating it and just drawing the process out unnecessarily.

> All that said, I did get the impression from these PRs that the opinion is the windows code is messy and hard to maintain. If you have notes on what needs to be cleaned up / simplified, I'd be willing to find some time to help do so.)

It is. It's probably some of the worst code in the repository in my opinion and the entire Windows ecosystem is a complete mess. I personally do not enjoy working on that side much and I doubt it would be much fun to try and refactor all of it. But if you're interested to put in the time anyways, it would obviously be very welcome.

alacritty_terminal/src/tty/windows/mod.rs
```diff
@@ -306,3 +323,17 @@ impl EventedPty for Pty {
...
+            // The conpty close call will block until the conout pipe is drained.
+            //
+            // EventedAnonRead has an internal buffer which is continuously draining the conout
+            // pipe into, so as long as this `.close()` call is run before self.conout is dropped
+            // it won't lead to a deadlock. (It will block just for a moment while EventedAnonRead
+            // does its internal workings.)
+            c.close();
+        }
+    }
+}
```
chrisduerr: Okay, so having to call `close` here with runtime-enforced crash in the `Drop` for the `ConPTY` is a bit problematic. We kinda use Rust to make sure things are enforced at compile-time as well as possible and this pushes a lot of the checks to runtime which certainly isn't ideal or intuitive.

I'd assume the problem arose because it's not possible to manually specify the drop order in the `Drop` impl for `Pty`, right? So you weren't able to do what you're currently doing in `close` in the `Drop` impl for `ConPTY`?

Rust actually has a mechanism to change that, which would allow us to have a well defined drop order while still relying on the standard `Drop` trait instead of custom `close` functions. This is documented in https://github.com/rust-lang/rfcs/blob/master/text/1857-stabilize-drop-order.md.

The short version is that variables that are dropped in the order they are defined, like this:

```rust
struct X {
    a: String, // Dropped First
    b: String, // Dropped Second
    c: String, // Dropped Third
}
```

So I think we could save ourselves a lot of runtime logic by making use of this well defined behavior. We probably wouldn't need any of the atomic operations, or the strange `close` method if done like this.

To prevent people from destroying things, we could just add a comment to the `PTY`'s handler field that explains this.

davidhewitt: I'm happy to rely on drop order!

To do that, I'll have to change `PtyHandle` to store an actual `Conpty` rather than `ConptyHandle` (which is `Arc<Conpty>`), so that dropping the `Pty` actually drops the `Conpty` rather than just decreasing the reference count.

This will lead to a little bit of fallout:
- Because doing the above will mean we won't have `Conpty` in an `Arc`, I'll have to add a `ConptyResizeHandle` type which implements `OnResize`. Dropping a `ConptyResizeHandle` won't close the pseudoconsole.
- Change the `Pty::resize_handle` function - because it'll need to return `WinptyHandle` for WinPTY, and `ConptyResizeHandle` for ConPTY. Maybe the best option is for it to return `Box<impl OnResize>`?

It's a mess indeed! I can do some follow-up PRs to solve this tangle of WinPTY / ConPTY code.

Sound good to you?

davidhewitt:
Thanks for taking that on board. I'd like to repeat that you do an awesome job with all the work you do on Alacritty. You don't get thanked enough for all of that!

> However in certain cases it's important for me personally to make stronger statements, just to make it clear which points can be discussed and which points are useless arguing about.

I completely agree with this. Let's not beat around the bush by sugarcoating things. What I ask is that instead of being told

> Pulling in crossbeam just for this seems absolutely crazy to me

which reads to me that my solution (which I thought carefully about) is absolutely crazy (*ouch!*), I'd prefer to be directed with something more like

> Let's please not add a new dependency just for this one fix. Can we do this without crossbeam?

I do appreciate that it can be really hard when writing online to know how the reader will interpret things!

--- 

> It is. It's probably some of the worst code in the repository in my opinion and the entire Windows ecosystem is a complete mess. I personally do not enjoy working on that side much and I doubt it would be much fun to try and refactor all of it. But if you're interested to put in the time anyways, it would obviously be very welcome.

Awesome, let's make it nicer :). I've opened #3085 to coordinate that.

alacritty_terminal/src/tty/windows/mod.rs
chrisduerr: > To do that, I'll have to change PtyHandle to store an actual Conpty rather than ConptyHandle (which is Arc<Conpty>), so that dropping the Pty actually drops the Conpty rather than just decreasing the reference count.

That's a good point, I haven't thought about this. It also *again* makes things a lot more complicated of course.

We won't be able to have `resize_handle` just be `self.handle.clone()` anymore, since the handle itself cannot be shared without any issues so we'd probably have to create a resize handle trait or something like that which the PTY backends would have to implement.

We'd then also have to make sure that when resizing, the ConPTY hasn't been closed yet, adding yet more friction and caveats to the ConPTY API.

From my understanding, what we would want is more similar to just packing `Pty` in an `Arc` instead of the `ConPTY` handle and making sure we keep the entire `Pty` around until it can be destroyed. However we cannot do that since there are a lot of other fields on the `Pty` which we do not want to reference count.

Resizing is also the only thing the handle is ever used for. So a design which does not implement drop for the PtyHandle, but then stores the handle on the `Pty` just to ensure the `ConPTY` is dropped at the right time, is probably a bit strange and confusing.

I'm wondering if the 'resize from anywhere' structure of the resize handle is a good fit at all for ensuring these drop order invariants. In all of the current approaches we had some kind of `closed` atomic, or other various bits that ensured that the resize handle doesn't explode after the PTY was closed.

We already have a channel which we use in `event.rs` to write info to the pty, I wonder if we could use that same channel to send resize messages through. Then the `event_loop.rs` would be the sole owner of the `pty` and we could just read new messages from the channel and directly call `pty.resize(xxx)`. That way we do not need an `Arc` anymore and all our problems would be gone.

However I honestly do not want to keep suggesting changes with this PR just to see how they do not work out once implemented. I think the best way forward is to go for the simplest and smallest fix possible to resolve this issue. If we want to actually fix things, we need to refactor code and that should come after fixing this bug.

That probably means just manually calling `drop(pty)` before the program exits, adding a comment above that why we need it and then we automatically get the correct behavior since our drop order is correct. The resize_handle staying around after that doesn't matter at all, since it will not receive any more events because everything is shut down already anyways, so all it does is delay the close call on the ConPTY which doesn't hurt anything. Just to make sure, I'd still reorder the fields in `ConPTY` just to make sure that if the `resize_handle` is dropped first, everything still works fine. And then we should be golden.

chrisduerr:
> Thanks for taking that on board. I'd like to repeat that you do an awesome job with all the work you do on Alacritty. You don't get thanked enough for all of that!

Thanks, it's always nice to hear positive feedback, even though I do all this solely for personal enjoyment.

>> Pulling in crossbeam just for this seems absolutely crazy to me

> which reads to me that my solution (which I thought carefully about) is absolutely crazy (ouch!), I'd prefer to be directed with something more like

With this comment it certainly could have been phrased a lot better, though I do absolutely think it's crazy how massive depedencies are just thrown into Rust projects without a second thought.

I've certainly done this myself plenty of times and it's partially the fault of `cargo` just being too good and easy to use. But I certainly think that especially in the Rust ecosystem people (including myself) need to be more aware of what they pull in as dependencies and act a little more conservative.

alacritty_terminal/src/tty/windows/mod.rs
davidhewitt: > However I honestly do not want to keep suggesting changes with this PR just to see how they do not work out once implemented. I think the best way forward is to go for the simplest and smallest fix possible to resolve this issue. If we want to actually fix things, we need to refactor code and that should come after fixing this bug.

I totally feel you on this.

Ok, I'm going to do the following: 

- As you suggest, I add an appropriate `drop` call in `main.rs`, with a comment saying why we need it to fix this bug. I already know where this needs to go.
- I undo all the rest of the changes in this PR.
- I'll make a note of the things we came across here in #3085 so we can integrate that into the cleanup plan.

> We already have a channel which we use in event.rs to write info to the pty, I wonder if we could use that same channel to send resize messages through. Then the event_loop.rs would be the sole owner of the pty and we could just read new messages from the channel and directly call pty.resize(xxx). That way we do not need an Arc anymore and all our problems would be gone.

This is an interesting thought. Would I be right in understanding this would be a change for all platforms, not just Windows? (Even if it helps solve our Windows issues.)

davidhewitt:
Heh, indeed `cargo` is a great and dangerous friend for us all 😄

alacritty_terminal/src/tty/windows/mod.rs
chrisduerr: > This is an interesting thought. Would I be right in understanding this would be a change for all platforms, not just Windows? (Even if it helps solve our Windows issues.)

Yes. On Linux the resize handle is just an integer, so it doesn't really matter what we do here.

We'd just change it on all platforms, since the code change that is required for Windows should automatically just work on Linux too.

Assuming that the order of the struct is in correct drop order, that should fix the problem without ever doing any `ConPTY` specific changes (other than the drop order thing). We'd basically just reduce complexity by getting rid of pointless multithreading support.

davidhewitt:
Awesome, I've burned it all and replaced it with a carefully-placed `drop` and a few comments. 🔥🔥🔥🔥 😄

alacritty/src/main.rs
```diff
@@ -222,6 +222,19 @@ fn run(window_event_loop: GlutinEventLoop<Event>, config: Config) -> Result<(),
...
+    // The cause:
+    //   - Drop for Conpty will deadlock if the conout pipe has already been dropped.
+    //   - The conout pipe is dropped when the io_thread is joined below (io_thread owns pty).
+    //   - Conpty is dropped when the last of processor and io_thread are dropped, because
+    //     both of them own an Arc<Conpty>.
+    //
+    // The fix is to ensure that processor is dropped first. That way, when io_thread (i.e. pty)
+    // is dropped, it can ensure Conpty is dropped before the conout pipe in the pty drop order.
+    //
+    // TODO: refactor the Windows codebase and remove this horrendous workaround!
```
chrisduerr: I'd like to be a bit more precise in this comment.

```suggestion
    // FIXME: Change PTY API to enforce the correct drop order with the typesystem.
```

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -90,6 +90,10 @@ pub type ConptyHandle = Arc<Conpty>;
 
 impl Drop for Conpty {
     fn drop(&mut self) {
+        // Caution! This will block until the conout pipe is drained. May cause deadlocks if the
+        // conout pipe has already been dropped by this point.
+        //
+        // See PR #3084 and https://docs.microsoft.com/en-us/windows/console/closepseudoconsole
```
chrisduerr: Many text editors like vim already recognize things like `XXX` and highlight it. Using `Caution!` here will likely make this look more like a normal comment than something important, so I'd recommend replacing `Caution!` with `XXX:`.

The phrasing `May cause deadlocks ...` is also kinda incorrect. It would be more accurate to say `Will cause deadlock ...`.

alacritty_terminal/src/tty/windows/mod.rs
```diff
@@ -45,6 +45,8 @@ pub enum PtyHandle {
 }
 
 pub struct Pty {
+    // It is important for drop order that this handle is defined before conout. Drop for Conpty
+    // will deadlock if the conout pipe has already been dropped.
```
chrisduerr: I'd suggest the following:

```
// XXX: Handle is required to be the first field, to ensure correct drop order. Dropping `conout` before `handle` will cause a deadlock.
```

This should give us the nice benefit to highlight this message in most editors.

davidhewitt:
Cheers, I've corrected the comments as advised. 😄

davidhewitt:
Thanks for fixing my formatting :)

Issue #3085, opened 2019-12-10
Windows: Refactoring & cleanups


davidhewitt:
I have the impression the windows code needs refactoring and work to bring it to parity with other platforms. Let's use this ticket to collect together what could be done. (I'd then be happy to steadily work through the list we create.)

Please add comments to suggest what needs working and I'll update this list below.

To spec out / create tickets for:
- [ ] Font ligatures (see linux WIP PR #2677)
- [ ] Reduce / remove unsafe code
- [ ] Reduce duplication, add cross-pty API
- [x] ~~Remove Arcs / resize handles because of the pain they cause. Probably needs #3086 to be done first.~~
- [ ] Improve the quality of the error messages (get as much information from the OS as possible)

Ticketed items:
- Emoji support (#3082)
- ~~Conpty by default (#2890)~~

chrisduerr:
I think #3082 is largely unrelated to the main cause of headache which would be the Windows PTY code.

Since #3082 is solely about the directwrite implementation, things are a bit easier. I think this was implemented separately and is relatively small, so I don't feel like there are any issues with that code. Looking at it just to make sure, it seems very nice to me compared to some of the other code.

The ConPTY/WinPTY code is what I'd primarily address. Unfortunately due to issues with Windows there is no way we can just kill off WinPTY yet, but changing to ConPTY by default might be a first step. After that things like error handling, unsafety, duplication and cross-pty API unification are some of the things I'd like to see improve.

chrisduerr:
Also I'd generally be wary of making this into just a 'fix windows' issue. While tracking issues might be nice in some cases, they have the tendency to never ever get closed.

So splitting stuff into issues that can actually get tackled and closed is desirable.

davidhewitt:
> Also I'd generally be wary of making this into just a 'fix windows' issue. While tracking issues might be nice in some cases, they have the tendency to never ever get closed.
>
> So splitting stuff into issues that can actually get tackled and closed is desirable.

👍. What I'd like to do is use this issue to discuss what we don't like about the current situation, and then let's break those notes up into well-defined separate tickets and close this one.

davidhewitt:
> After that things like error handling

Can you elaborate a little more what needs fixing about with the existing error handling? 

The alert box that pops up with a crash explicitly talks about panicking and then prints a debug representation of the error (I think?). This could be made to have friendlier error messages. Maybe there's other things you're thinking too?

chrisduerr:
[...]

PR #3094, opened 2019-12-13
Replace dynamically-loaded kernel32 functions with winapi

sterlingjensen:
The libloaderapi implementation pre-dates availability in consoleapi (0.3.7).
The winapi feature flags have fallen out of sync with what is in use, as no relative paths to the crate are used in this project, so they were removed.

### DIFF
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -35,7 +35,10 @@ winpty = { path = "../winpty" }
 mio-named-pipes = "0.1"
 miow = "0.3"
 dunce = "1.0"
-winapi = { version = "0.3.7", features = ["impl-default", "basetsd", "libloaderapi", "minwindef", "ntdef", "processthreadsapi", "winbase", "wincon", "wincontypes", "winerror", "winnt", "winuser"] }
+winapi = { version = "0.3.7", features = [
+    "impl-default", "basetsd", "libloaderapi", "minwindef", "ntdef", "processthreadsapi", "winbase",
+    "wincon", "wincontypes", "winerror", "winnt", "winuser",
+]}
 widestring = "0.4"
 mio-anonymous-pipes = "0.1"
 
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -35,9 +35,12 @@ winpty = { path = "../winpty" }
 mio-named-pipes = "0.1"
 miow = "0.3"
 dunce = "1.0"
-winapi = "0.3.7"
 widestring = "0.4"
 mio-anonymous-pipes = "0.1"
+[target.'cfg(windows)'.dependencies.winapi]
+version = "0.3.7"
+features = ["impl-default", "basetsd", "libloaderapi", "minwindef", "ntdef", "processthreadsapi",
+            "winbase", "wincon", "wincontypes", "winerror", "winnt", "winuser"]
 
 [target.'cfg(target_os = "macos")'.dependencies]
 objc = "0.2.2"
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -35,12 +35,9 @@ winpty = { path = "../winpty" }
 mio-named-pipes = "0.1"
 miow = "0.3"
 dunce = "1.0"
+winapi = { version = "0.3.7", features = ["impl-default", "basetsd", "libloaderapi", "minwindef", "ntdef", "processthreadsapi", "winbase", "wincon", "wincontypes", "winerror", "winnt", "winuser"] }
 widestring = "0.4"
 mio-anonymous-pipes = "0.1"
-[target.'cfg(windows)'.dependencies.winapi]
-version = "0.3.7"
-features = ["impl-default", "basetsd", "libloaderapi", "minwindef", "ntdef", "processthreadsapi",
-            "winbase", "wincon", "wincontypes", "winerror", "winnt", "winuser"]
 
 [target.'cfg(target_os = "macos")'.dependencies]
 objc = "0.2.2"
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -35,7 +35,7 @@ winpty = { path = "../winpty" }
 mio-named-pipes = "0.1"
 miow = "0.3"
 dunce = "1.0"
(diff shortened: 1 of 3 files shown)


chrisduerr:
> The winapi feature flags have fallen out of sync with what is in use, as no relative paths to the crate are used in this project, so they were removed.

I'm not sure what you mean by this? What do feature flags have to do with relative paths?

If they're not necessary that's fine, though it might be worthwhile to check if we might want to do no-default-features and then add only what we need ourselves.

davidhewitt:
As much as I love this, I'm not sure we can do it. Won't it break support for Windows versions which predate the existence of the ConPTY api?

For example, on all Windows 7 versions `kernel32` doesn't have these functions. So with this PR I expect Alacritty would crash during load-time dynamic linking.

(Ofc I could just be totally misunderstanding how things work. Correct me if so.)

Alternatively; go ahead and do it. It would make having ConPTY support a hard requirement for Alacritty on Windows. We could then take out the Winpty implementation completely and simplify the lot 😈 🙈

sterlingjensen:
I've only seen the feature flags used to control what is made available in a `mod.rs` import (`winapi::um`), with sub-modules being controlled in the `Cargo.toml` (...`features = ["d3d11"]`...). This means that all usage would rely on paths relative to that import (`um::d3d11::D3D11_RECT`). I don't see any default features in `winapi`.
I hadn't considered graceful failures for Windows 7... that would certainly make this a non-starter. I'll spin up a vm and check it out.

davidhewitt:
> I hadn't considered graceful failures for Windows 7... that would certainly make this a non-starter. I'll spin up a vm and check it out.

To be fair, Windows 7's end of life is next month, but there's still Windows 8 and old versions of Windows 10 kicking around :)

chrisduerr:
@davidhewitt Thanks for pointing that out, very, very good point. I don't think we should be killing off WinPTY just yet.

davidhewitt:
A quick thought on this - if we decide not to kill off the dynamic loading right now, maybe let's update the comment so that people don't try this again until we're ready for it?

sterlingjensen:
Well that was a frustrating experience, the approaching EOL really shows when it comes to tooling. Long story short: now I know how this seemingly low hanging fruit got overlooked until I dove for it: this change breaks things. I can think of a bunch of ways to fix this, but the only correct way I can imagine involves some high level design choices inside `tty::windows`. I think that discussion is better had at #3085.
Should I close this, change the title to include '[WIP]', or leave it as is for now?

sterlingjensen:
@davidhewitt That would have saved me a day, so it gets my vote :smirk:

chrisduerr:
>  Should I close this, change the title to include '[WIP]', or leave it as is for now?

I'm in favor of closing this, yeah. Having a ton of PRs open which aren't going to see any work done makes things a lot harder for me.

davidhewitt:
> @davidhewitt That would have saved me a day, so it gets my vote 😏

Ah damn, sorry about that (the original dynamic load was written by me, so I feel responsible for that) ... :/

davidhewitt:
>  I can think of a bunch of ways to fix this, but the only correct way I can imagine involves some high level design choices inside tty::windows. I think that discussion is better had at #3085.

Please do drop your thoughts there. Very keen to hear any ideas / feedback for tidying this up :)

sterlingjensen:
I just pushed a patch that leaves the removal of the feature flags, but reverts the changes to dynamic loading and adds a todo comment to ward off others. @chrisduerr let me know if that works as a nice way to close through a merge, otherwise I'll close it.

@davidhewitt Don't feel bad, I almost messaged you when I first noticed the opportunity to drop the dynamic load, and saw your involvement in both it and the winapi implementation that would replace it :) But I resisted the urge, because it felt like volunteering you for more work. Anyway, it wasn't a wasted day: I've now got an up to date golden master for win7, some leads on some potential gl-rs bugs, and a new appreciation for not using Windows 7 as a daily driver.

davidhewitt:
> I just pushed a patch that leaves the removal of the feature flags

Just curious about the winapi feature flags, how does it still build without them? I guess we're probably getting lucky that these features are already pulled in by upstream crates?

sterlingjensen:
It doesn't take much, with all the interdependency - winapi directly pulls in most of it, and its rest likely from its other dependencies. I've resynced the feature flags.

davidhewitt:
:+1: thanks, that seems like a good idea to me

alacritty_terminal/Cargo.toml
```diff
@@ -35,9 +35,12 @@ winpty = { path = "../winpty" }
 mio-named-pipes = "0.1"
 miow = "0.3"
 dunce = "1.0"
-winapi = { version = "0.3.7", features = ["impl-default", "winuser", "synchapi", "roerrorapi", "winerror", "wincon", "wincontypes"]}
 widestring = "0.4"
 mio-anonymous-pipes = "0.1"
+[target.'cfg(windows)'.dependencies.winapi]
+version = "0.3.7"
+features = ["impl-default", "basetsd", "libloaderapi", "minwindef", "ntdef", "processthreadsapi",
+            "winbase", "wincon", "wincontypes", "winerror", "winnt", "winuser"]
```
chrisduerr: Why did you add new feature flags that previously weren't required at all?

I also think that if we need to break it over multiple lines anyways, we might as well stick with the old version of defining `winapi` as a single element instead of a nested field.

davidhewitt:
> Why did you add new feature flags that previously weren't required at all?

I guess that these are all the `winapi` submodules that appear in paths inside `alacritty_terminal`. Previously some of these features must have been pulled in by dependencies (or automatically as dependencies of other `winapi` features).

I'm not sure what the correct pattern is here either. It feels most correct to list them all out here as @sterlingjensen has done. But for the sake of brevity, and to minimise maintenance burden keeping the list in sync, I tend to go for the smallest list of features which are needed to get a successful compile. As we saw here that's actually  no feature flags :smile:

chrisduerr:
I'm all in favor of adding all depedencies we do have ourselves. I just want to know.

alacritty_terminal/Cargo.toml
sterlingjensen: They were in use before, but not specified in the manifest - out of sync.
One single line toml table, here comes the line wrap :metal:

chrisduerr:
> unless they are valid within a value

Well, they are valid within arrays, so there will be no problem. As you can see, there are no newlines outside of the array. That's why the closing `]` and `}` are on the same line.

sterlingjensen:
Oh I'm certainly in favor of doing the thing that doesn't cause a line wrap, I'm just glad that you did it - and not me :grinning:

PR #3096, opened 2019-12-13
Send resize messages through the event loop

### LINKED ISSUE #3086: Send resize messages through event loop
From #3084:

> We already have a channel which we use in event.rs to write info to the pty, I wonder if we could use that same channel to send resize messages through. Then the event_loop.rs would be the sole owner of the pty and we could just read new messages from the channel and directly call pty.resize(xxx). That way we do not need an Arc anymore and all our problems would be gone.

Feel free to assign this issue to me if you want. It feels like a cleanup which is mostly intended to clean up Windows, even if it's cross-platform (#3085)

davidhewitt:
Closes #3086 

This turned out to be a great idea! I was able to leverage the existing `OnResize` trait and use it to plumb through the messages in a couple places.

I was also able to completely delete the `tty::winpty::Agent` type and just re-export `pub use winpty::Winpty as Agent`.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -308,20 +308,18 @@ pub struct Processor<N> {
     suppress_chars: bool,
     modifiers: ModifiersState,
     config: Config,
-    pty_resize_handle: Box<dyn OnResize>,
     message_buffer: MessageBuffer,
     display: Display,
     font_size: Size,
 }
 
-impl<N: Notify> Processor<N> {
+impl<N: Notify + OnResize> Processor<N> {
     /// Create a new event processor
     ///
     /// Takes a writer which is expected to be hooked up to the write end of a
     /// pty.
     pub fn new(
         notifier: N,
-        pty_resize_handle: Box<dyn OnResize>,
         message_buffer: MessageBuffer,
         config: Config,
         display: Display,
@@ -334,7 +332,6 @@ impl<N: Notify> Processor<N> {
             modifiers: Default::default(),
             font_size: config.font.size,
             config,
-            pty_resize_handle,
             message_buffer,
             display,
         }
@@ -405,7 +402,7 @@ impl<N: Notify> Processor<N> {
             if !display_update_pending.is_empty() {
                 self.display.handle_update(
                     &mut terminal,
-                    self.pty_resize_handle.as_mut(),
+                    &mut self.notifier,
                     &self.message_buffer,
                     &self.config,
                     display_update_pending,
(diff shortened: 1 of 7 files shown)


davidhewitt:
(force pushed to fix `clippy` and `fmt` errors)

alacritty_terminal/src/event_loop.rs
```diff
@@ -82,6 +85,14 @@ impl event::Notify for Notifier {
     }
 }
 
+impl event::OnResize for Notifier {
+    fn on_resize(&mut self, size: &SizeInfo) {
+        if self.0.send(Msg::Resize(*size)).is_err() {
+            panic!("expected send event loop msg");
+        }
```
chrisduerr: I'd assume you've mirrored the other code in this file with this, right?

Generally a great idea, but I do not understand the purpose of doing this. Couldn't we just `expect("error")`? I'd assume the original author must have forgotten about it or something got confused in a refactoring.

If you're down, feel free to change all instances of this oddness in this file.

alacritty_terminal/src/tty/windows/mod.rs
```diff
@@ -38,16 +38,15 @@ pub fn is_conpty() -> bool {
     IS_CONPTY.load(Ordering::Relaxed)
 }
 
-#[derive(Clone)]
-pub enum PtyHandle {
-    Winpty(winpty::WinptyHandle),
-    Conpty(conpty::ConptyHandle),
+enum PtyImpl {
```
chrisduerr: Not super happy with this name. `Impl` feels like it's a struct implementing a trait, but `Pty` is a trait and `PtyImpl` is an enum. Additionally things like `Impl` and `Trait` shouldn't be appended in Rust.

I'm getting Java flashbacks from my dayjob here, so I think we should probably rename this to me more rusty

TODO: Suggest better name.

alacritty_terminal/src/event_loop.rs
davidhewitt: Will do 👍

alacritty/src/event.rs
```diff
@@ -308,20 +308,18 @@ pub struct Processor<N> {
...
 }
 
-impl<N: Notify> Processor<N> {
+impl<N: Notify + OnResize> Processor<N> {
     /// Create a new event processor
     ///
     /// Takes a writer which is expected to be hooked up to the write end of a
     /// pty.
     pub fn new(
         notifier: N,
```
davidhewitt: Bikeshed: now that `notifier` implements both `Notify` and `OnResize`, should we come up with a better name?

PR #3102, opened 2019-12-14
Fix cut off full width glyphs in last column

### LINKED ISSUE #2385: Wrap line if double-width character is in last column
Alacritty already correctly wraps lines when the max number of text in a row is reached, however with double-width characters things change a little bit.

If a double-width character is printed in the last cell, a line break needs to be inserted before it. This is because there's only one cell left but the cell is two cells wide, leading to everything going beyond the first cell getting cut off.

When testing, special care should be taken not to break selection and text reflow. Both these features make the assumption that the last cell will be filled when a row is 'full'.

chrisduerr:
This resolves the issue with full width glyphs getting rendered in the
last column. Since they need at least two glyphs, it is not possible to
properly render them in the last column.

Instead of rendering half of the glyph in the last column, with the
other half cut off, an additional spacer is now inserted before the wide
glyph. This means that the specific glyph in question is then three
cells wide.

Fixes #2385.

### DIFF
--- a/alacritty_terminal/src/grid/mod.rs
+++ b/alacritty_terminal/src/grid/mod.rs
@@ -279,94 +279,97 @@ impl<T: GridCell + PartialEq + Copy> Grid<T> {
 
     // Grow number of columns in each row, reflowing if necessary
     fn grow_cols(&mut self, reflow: bool, cols: Column, cursor_pos: &mut Point, template: &T) {
+        // Check if a row needs to be wrapped
+        let should_reflow = |row: &Row<T>| -> bool {
+            let len = Column(row.len());
+            reflow && len < cols && row[len - 1].flags().contains(Flags::WRAPLINE)
+        };
+
         let mut new_empty_lines = 0;
-        let mut new_raw: Vec<Row<T>> = Vec::with_capacity(self.raw.len());
+        let mut reversed: Vec<Row<T>> = Vec::with_capacity(self.raw.len());
         for (i, mut row) in self.raw.drain().enumerate().rev() {
-            if let Some(last_row) = new_raw.last_mut() {
-                // Grow the current line if there's wrapped content available
-                if reflow
-                    && last_row.len() < cols.0
-                    && last_row
-                        .last()
-                        .map(|cell| cell.flags().contains(Flags::WRAPLINE))
-                        .unwrap_or(false)
-                {
-                    // Remove wrap flag before appending additional cells
-                    if let Some(cell) = last_row.last_mut() {
-                        cell.flags_mut().remove(Flags::WRAPLINE);
+            let last_row = match reversed.last_mut() {
+                Some(last_row) if should_reflow(last_row) => last_row,
+                _ => {
+                    reversed.push(row);
+                    continue;
+                },
+            };
+
+            // Remove wrap flag before appending additional cells
+            if let Some(cell) = last_row.last_mut() {
+                cell.flags_mut().remove(Flags::WRAPLINE);
+            }
+
+            // Remove leading spacers when reflowing wide char to the previous line
+            let last_len = last_row.len();
+            if last_len >= 2
+                && !last_row[Column(last_len - 2)].flags().contains(Flags::WIDE_CHAR)
+                && last_row[Column(last_len - 1)].flags().contains(Flags::WIDE_CHAR_SPACER)
+            {
+                last_row.shrink(Column(last_len - 1));
+            }
+
+            // Append as many cells from the next line as possible
+            let len = min(row.len(), cols.0 - last_row.len());
+
+            // Insert leading spacer when there's not enough room for reflowing wide char
+            let mut cells = if row[Column(len - 1)].flags().contains(Flags::WIDE_CHAR) {
+                let mut cells = row.front_split_off(len - 1);
+
+                let mut spacer = *template;
+                spacer.flags_mut().insert(Flags::WIDE_CHAR_SPACER);
+                cells.push(spacer);
+
+                cells
+            } else {
+                row.front_split_off(len)
+            };
+
+            last_row.append(&mut cells);
+
+            if row.is_empty() {
+                let raw_len = i + 1 + reversed.len();
+                if raw_len < self.lines.0 || self.scroll_limit == 0 {
+                    // Add new line and move lines up if we can't pull from history
+                    cursor_pos.line = Line(cursor_pos.line.saturating_sub(1));
+                    new_empty_lines += 1;
+                } else {
+                    // Make sure viewport doesn't move if line is outside of the visible
+                    // area
+                    if i < self.display_offset {
+                        self.display_offset = self.display_offset.saturating_sub(1);
                     }
 
-                    // Remove leading spacers when reflowing wide char to the previous line
-                    let last_len = last_row.len();
-                    if last_len >= 2
-                        && !last_row[Column(last_len - 2)].flags().contains(Flags::WIDE_CHAR)
-                        && last_row[Column(last_len - 1)].flags().contains(Flags::WIDE_CHAR_SPACER)
-                    {
-                        last_row.shrink(Column(last_len - 1));
-                    }
-
-                    // Append as many cells from the next line as possible
-                    let len = min(row.len(), cols.0 - last_row.len());
-
-                    // Insert leading spacer when there's not enough room for reflowing wide char
-                    let mut cells = if row[Column(len - 1)].flags().contains(Flags::WIDE_CHAR) {
-                        let mut cells = row.front_split_off(len - 1);
-
-                        let mut spacer = *template;
-                        spacer.flags_mut().insert(Flags::WIDE_CHAR_SPACER);
-                        cells.push(spacer);
-
-                        cells
-                    } else {
-                        row.front_split_off(len)
-                    };
-
-                    last_row.append(&mut cells);
-
-                    if row.is_empty() {
-                        let raw_len = i + 1 + new_raw.len();
-                        if raw_len < self.lines.0 || self.scroll_limit == 0 {
-                            // Add new line and move lines up if we can't pull from history
-                            cursor_pos.line = Line(cursor_pos.line.saturating_sub(1));
-                            new_empty_lines += 1;
-                        } else {
-                            // Make sure viewport doesn't move if line is outside of the visible
-                            // area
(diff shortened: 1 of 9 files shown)


chrisduerr:
Seems like the performance of this PR is actually better than before, after some minor tweaks.

Waiting for @kchibisov before merge, since he wanted to take a look..

alacritty_terminal/src/term/mod.rs
```diff
@@ -1317,20 +1316,20 @@ impl<T: EventListener> Handler for Term<T> {
         }
 
         if width == 1 {
-            self.write_cursor_cell(c);
+            self.write_at_cursor(c);
         } else if width == 2 {
-            // Insert placeholder and linebreak if only one cell left for wide glyph
+            // Insert extra placeholder before wide char if glyph doesn't fit this row anymore
```
cole-h: nit: `glyph doesn't fit this row anymore` -> `glyph doesn't fit *in* this row anymore`

alacritty_terminal/src/index.rs
```diff
@@ -44,24 +44,32 @@ impl<L> Point<L> {
 
     #[inline]
     #[must_use = "this returns the result of the operation, without modifying the original"]
-    pub fn sub(mut self, num_cols: usize, length: usize) -> Point<L>
+    pub fn sub(mut self, num_cols: usize, length: usize, absolute_indexing: bool) -> Point<L>
```
kchibisov: Could we express this `bool` in function names. I find it a bit odd that function called `sub`/`add` takes a bool now. For e.g. in rust there's a `sub` and a `wrapping_sub`.

kchibisov:
There's a strange bug in bash. I've created a url with these symbols, and then start moving it to the new line. The space, that you can see wasn't added by me. and in fact if I move back to an original line everything will be the same. You can also see that a cursor is broken.

picture

picture

kchibisov:
Moving double width chars(test this with URL) is inconsistent a bit. Sometimes you move entire chars and adding a space, sometimes they moved only half way, resulting a broken highlighting most of the time. To test this thing place a url 
`https://こんにち` in the end of line in bash and start moving it through the end by char and highlight it. You should see that chars sometimes are not fully underlined or not moving to the next line, etc, etc. I guess selection across multiple lines not always working for double width chars is kind of related to that.

chrisduerr:
What do you mean by `moving it` and how does this behave before this patch?

I can't reproduce it, though I'm not sure I understand what you mean.

kchibisov:
The performance of this branch is questionable btw.
```
Master:
nightly asrw 6.00107 +- 0.00357 seconds time elapsed  ( +-  0.06% )
stable  asrw 6.36960 +- 0.00118 seconds time elapsed  ( +-  0.02% )

PR:
nightly asrw 6.35584 +- 0.00284 seconds time elapsed  ( +-  0.04% )
stable asrw 5.84472 +- 0.00150 seconds time elapsed  ( +-  0.03% )
```

I mean it feels like there's a "flip" in rustc, since there's something similar between this benchmark results. I'm not sure what we should do about it, probably nothing, but ideally we can figure out what is causing this results and tune to be the same as our master. And I'm sure that I was benchmarking the right terminals.

kchibisov:
@chrisduerr I hope these gifs can help you with my comment.

PR: 
bad

master:
good

chrisduerr:
> The performance of this branch is questionable btw.

Which nightly are you running?

> @chrisduerr I hope these gifs can help you with my comment.

Thanks, I wasn't aware this is about URL highlighting.

kchibisov:
@chrisduerr `rustc 1.41.0-nightly (c8ea4ace9 2019-12-14)`

chrisduerr:
@kchibisov I still can't reproduce the bash issue with URLs btw. It is also noteworthy that both ZSH and bash seem to do their own custom shenanigans when double-width chars are in the last column.

kchibisov:
The 2 bugs I've mentioned were the bugs related to my old (bash-4.4.23), the bash 5.0.11 is fine (at least on my 2 examples ;- )).

alacritty_terminal/src/selection.rs
```diff
@@ -214,16 +214,49 @@ impl Selection {
         span.map(|mut span| {
```
kchibisov: What other apps do with block selection when on one line you have 2 full width glyphs?

Anyway, I feel like you have a bug with block selection and 2 width chars. All selection on this gif was block one, I hope this gif gives enough explanation, if it's not feel free to ask.

block

alacritty_terminal/src/term/mod.rs
```diff
@@ -1233,82 +1273,66 @@ impl<T: EventListener> ansi::Handler for Term<T> {
...
-
-                let src = line[col..].as_ptr();
-                let dst = line[(col + width)..].as_mut_ptr();
-                unsafe {
-                    // memmove
-                    ptr::copy(src, dst, (num_cols - col - width).0);
-                }
+        if width == 1 {
+            self.write_at_cursor(c);
+        } else if width == 2 {
```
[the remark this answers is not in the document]
chrisduerr: I did this because it is faster this way. I can retest on stable though.

alacritty_terminal/src/selection.rs
chrisduerr: We only ever check the beginning or end of a block selection, that's why the bottom left and top right will not snap to the full width glyphs. I'm not sure if it's worth fixing this, or if we maybe should just disable this completely for block selection.

I'll check what VTE does maybe that can give some insights..

kchibisov: I'm voting for consistency here. So we shouldn't do anything where we're in a block selection or do the same thing as vte(expand selectively on full width chars).

chrisduerr: This is unrelated to this PR, I'll create a follow-up issue about this.

kchibisov: It turned out that this thing is completely unrelated to this PR, so it should be moved to our issue tracker instead.

nixpulvis: I think we've reach agreement that we should do the same thing as VTE here (expand full width in block selections), however that is outside the scope of this PR.

chrisduerr:
@kchibisov reflow should now behave correctly, feel free to give that another shot if you have some time.

I think the functionality should be in a good place now, though I have to test everything again just to make sure. I still need to clean up the grow_cols and shrink_cols methods since they're pretty massive now, but that shouldn't affect functionality.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -362,20 +390,47 @@ impl<T: GridCell + PartialEq + Copy> Grid<T> {
...
+                    if len == 1 {
+                        row[cols - 1].flags_mut().insert(Flags::WRAPLINE);
+                        wrapped = None;
+                    } else {
+                        wrappedx[len - 2].flags_mut().insert(Flags::WRAPLINE);
+                        wrappedx.truncate(len - 1);
+                    }
+                }
+            }
+
```
kchibisov: I feel like this part should be hidden behind `if reflow {}` or partially hidden. Anyway I feel like it's a bit incorrect that you're leaving `WIDE_CHAR_SPACER` when reflow is disabled on deleting of `WIDE_CHAR`.

So enter alt screen with `echo -e  "\e[?1049h"`, do `echo  日本ジャパン八州八洲瑞穂の国` and start shrinking. You can apply the following diff and notice, that when you delete a WIDE_CHAR, you're leaving a spacer for it (just click with your mouse on a place where the char was).

```diff
diff --git a/alacritty/src/input.rs b/alacritty/src/input.rs
index 2a98dc1..d8f8235 100644
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -390,6 +390,8 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
                 self.ctx.mouse_mut().block_url_launcher = !self.ctx.selection_is_empty();

                 self.ctx.clear_selection();
+                let grid = &self.ctx.terminal().grid();
+                println!("GRID {:?}", grid[point.line][point.col]);

                 // Start new empty selection
                 let side = self.ctx.mouse().cell_side;
```

chrisduerr: > I feel like this part should be hidden behind if reflow {} or partially hidden. Anyway I feel like it's a bit incorrect that you're leaving WIDE_CHAR_SPACER when reflow is disabled on deleting of WIDE_CHAR.

Yeah, there's a few more changes necessary to functionality that I've already performed locally. I've reworked a good chunk of `shrink_cols` including this specific point.

chrisduerr:
I wasn't really able to clean up the `shrink_cols` and `grow_cols` methods much, since there's a lot to them and moving things into other functions quickly leads to massive lists of parameters with issues about multiple mutable borrowers.

Likely the only way to clean this up significantly would be to rewrite it, but I don't want to make such big changes in this PR and I'm not sure if that would improve anything. So I think it should be fine for now hopefully.

I've made some fixes to the behavior of shrinking column which should now hopefully give us fully accurate shrinking and growing behavior. I've also copied VTE's behavior of not performing this kind of reflow in the alt screen, which I think makes sense.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -279,125 +280,180 @@ impl<T: GridCell + PartialEq + Copy> Grid<T> {
...
-    fn grow_cols(
-        &mut self,
-        reflow: bool,
-        cols: index::Column,
-        cursor_pos: &mut Point,
-        template: &T,
-    ) {
+    // Grow number of columns in each row, reflowing if necessary
+    fn grow_cols(&mut self, reflow: bool, cols: Column, cursor_pos: &mut Point, template: &T) {
+        // Check if a row needs to be wrapped
```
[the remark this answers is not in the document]
chrisduerr: It does the checking itself I'd say and returns a `bool` whether the checking was successful and the row needs a wrap or if it doesn't. So I feel like this is accurate, though it is true that it's a function that is later executed, not immediately.

I'll see if I can put that into the doc without blowing it up.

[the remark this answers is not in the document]
kchibisov: Just looking at this comment again I feel like I had a hard times yesterday, since it's just a function to be honest, and it follows other function comments, so things should be fine the way they are right now.

PR #3113, opened 2019-12-20
Default to ConPTY instead of WinPTY (PR try 2)

davidhewitt:
I've cleaned up #2890 and improved the Windows notes in `README.md` slightly.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -339,9 +339,8 @@
 
 # WinPTY backend (Windows only)
 #
-# Alacritty defaults to using the newer ConPTY backend if it is
-# available, since it provides better color support, resolves various
-# other minor rendering bugs and is quite a bit faster. If it is not
+# Alacritty defaults to using the newer ConPTY backend if it is available,
+# since it resolves a lot of bugs and is quite a bit faster. If it is not
 # available, the the WinPTY backend will be used instead.
 #
 # Setting this option to `true` makes Alacritty use the legacy WinPTY backend,
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -337,16 +337,16 @@
 # directory of the parent process will be used.
 #working_directory: None
 
-# Windows 10 WinPTY backend (Windows only)
+# WinPTY backend (Windows only)
 #
 # Alacritty defaults to using the newer ConPTY backend if it is
 # available, since it provides better color support, resolves various
 # other minor rendering bugs and is quite a bit faster. If it is not
 # available, the the WinPTY backend will be used instead.
 #
-# Setting this option to true makes Alacritty use the WinPTY backend,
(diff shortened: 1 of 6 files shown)


alacritty.yml
```diff
@@ -337,16 +337,16 @@
...
+# WinPTY backend (Windows only)
 #
-# This will enable better color support and may resolve other issues,
-# however this API and its implementation is still young and so is
-# disabled by default, as stability may not be as good as the winpty
-# backend.
+# Alacritty defaults to using the newer ConPTY backend if it is
+# available, since it provides better color support, resolves various
+# other minor rendering bugs and is quite a bit faster. If it is not
+# available, the the WinPTY backend will be used instead.
```
chrisduerr: I think this might be a bit too much information. Especially things like `resolves various other minor rendering bugs` seems so unspecific, that listing it explicitly will likely just increase the effort it takes to understand this section.

If people really want to know the differences, I'm sure there are better sources, so maybe we can just keep it short like this?

> Alacritty defaults to using the newer ConPTY backend if it is available, since it resolves a lot of bugs and is quite a bit faster. If it is not available, the WinPTY backend will be used instead.

alacritty_terminal/src/config/mod.rs
```diff
@@ -108,11 +108,10 @@ pub struct Config<T> {
     #[serde(default, deserialize_with = "failure_default")]
     pub cursor: Cursor,
 
-    /// Enable experimental conpty backend instead of using winpty.
-    /// Will only take effect on Windows 10 Oct 2018 and later.
+    /// Disable conpty backend and use winpty even if conpty is available.
```
chrisduerr: You've doubled down on being super precise here, which I don't think is necessary. If ConPTY is disabled, it's implied that WinPTY isn't and the other way around.

The following shortens this a bit while conveying the same information:
```suggestion
    /// Use WinPTY backend even if ConPTY is available.
```

alacritty_terminal/src/tty/windows/mod.rs
```diff
@@ -61,11 +61,11 @@ pub struct Pty {
 
 pub fn new<C>(config: &Config<C>, size: &SizeInfo, window_id: Option<usize>) -> Pty {
     if let Some(pty) = conpty::new(config, size, window_id) {
-        info!("Using Conpty agent");
+        info!("Using ConPTY backend");
         IS_CONPTY.store(true, Ordering::Relaxed);
         pty
     } else {
-        info!("Using Winpty agent");
+        info!("Using WinPTY backend");
```
chrisduerr: Big fan of keeping the nomenclature consistent here. Thanks.

chrisduerr:
It seems like Windows builds are having troubles currently, so the failing CI status is fine for now. I'll try to get that figured out asap.

PR #3115, opened 2019-12-20
Fix grid.len() and grid.history_size() misreporting

kchibisov:
The grid reset should fail, but It seems like we should rerecord the test.

@chrisduerr


kchibisov:
>It looks like a simple and pratical solution to the problem, but this seems to have the problem that we might exceed our actual history buffer size with additional allocations, right? Say our history right now is 7 lines initialized and we want to grow to 8 and our maximum is also 8, if the min step is 10, we'll jump straight to 17.

Yeah, I was looking to add something like storage cap, but I decided that it doesn't make much sense for some internal use and a bad abstraction. So I went this way for simplicity.

>Additionally, requiring two separate functions for initialize_exact and initialize isn't great either, since it's likely going to be confusing and easy to write code that doesn't perform super well.

It depends, most of the time you should use `initialize` and when you want more efficient memory usage you should use a different function.

>Maybe we could just have a single method and pass two parameters, like initialize(len, buffer, template)

I don't really like this idea of passing buf, mostly because your arguments should depend on each other, like `len` shouldn't be larger then `buffer.len()` and it's easy to fail. So I've decided to expose the API in names rather then arbitrary function arguments. I've initially was using a bit different approach with always dynamically grow things.

Maybe we should add a different function like `initialize_all` instead of `initialize_exact`, idk.

I'm fine with adding more arguments to `initialize`, however I think we need something better then `buffer`. Idk, maybe it's time to add an arbitrary bool to `initialize`, so it'll be `initialize(len, smart, template)`, but I feel like exposing such thing in function names is much better, probably the name was wrong...

Anyway it just depends on what we're going to do with this storage.rs thing in the end. If we want something to hold our goals and use only in alacritty I think more parameters and tight control should be preferred here, since it's a performance/memory related part and using multiple parameters to a function like `initialize(len, internal_buf_resize_step, template)` should be preferred I guess.

Let me know what you think @chrisduerr ?

chrisduerr:
> I don't really like this idea of passing buf, mostly because your arguments should depend on each other, like len shouldn't be larger then buffer.len() and it's easy to fail. So I've decided to expose the API in names rather then arbitrary function arguments. I've initially was using a bit different approach with always dynamically grow things.

Yeah, that's a fair point.

> Maybe we should add a different function like initialize_all instead of initialize_exact, idk.

I don't think this makes sense, since `storage` has no knowledge of what 'all' means.

> I'm fine with adding more arguments to initialize, however I think we need something better then buffer. Idk, maybe it's time to add an arbitrary bool to initialize, so it'll be initialize(len, smart, template), but I feel like exposing such thing in function names is much better, probably the name was wrong...

What about making the buffer an offset from the first parameter? So we have one parameter that says how much the `len` should be grown and another parameter that says how much additional buffer space *after* len should be initialized for performance sake.

That still doesn't solve the problem that it wouldn't be a great abstraction though, because we still would have everything in `grid/mod.rs` instead of handling these performance optimizations in `storage.rs`.

I'd like to give this another thought myself though, I'll play around a bit more with some ideas tomorrow, so I'll get back to you on this one. If you have any better ideas, please let me know.

kchibisov:
>What about making the buffer an offset from the first parameter? So we have one parameter that says how much the len should be grown and another parameter that says how much additional buffer space after len should be initialized for performance sake.

It's an interesting idea tbh and you can't misuse the API in this way on purpose. However, since rust doesn't have default function args/function overloads this API is not very sound. Anyway I don't have better ideas besides using multiple functions, since it's better to expose different things with different names imho.

chrisduerr:
> However, since rust doesn't have default function args/function overloads this API is not very sound.

Just to clarify, by `not very sound` you mean that the API is difficult to use, not that it has `unsound` behavior, right? Since `unsound` has a bit of a special connotation with Rust.

kchibisov:
>Just to clarify, by not very sound you mean that the API is difficult to use

I talking "about difficult to use, because of some rust limitations".

alacritty_terminal/src/grid/storage.rs
```diff
@@ -173,25 +176,41 @@ impl<T> Storage<T> {
...
     pub fn initialize(&mut self, num_rows: usize, template_row: Row<T>)
     where
         T: Clone,
     {
-        let mut new = vec![template_row; num_rows];
+        if self.len + num_rows > self.inner.len() {
+            let len = self.len;
+            self.initialize_exact(max(num_rows, MIN_INIT_SIZE), template_row);
+            // Restore proper len
+            self.len = len + num_rows;
```
nixpulvis: Why is this necessary? Shouldn't `initialize_exact` properly update `self.len`?

alacritty_terminal/src/grid/storage.rs
```diff
@@ -173,25 +176,41 @@ impl<T> Storage<T> {
...
+    /// reallocations by using minimun step for growing internal storage. If you want to avoid this
+    /// behavior use `initialize_exact`.
     pub fn initialize(&mut self, num_rows: usize, template_row: Row<T>)
     where
         T: Clone,
     {
-        let mut new = vec![template_row; num_rows];
+        if self.len + num_rows > self.inner.len() {
+            let len = self.len;
+            self.initialize_exact(max(num_rows, MIN_INIT_SIZE), template_row);
```
nixpulvis: Using `max` here with `MIN_INIT_SIZE` doesn't make sense to me. This seems like it will always update by `MIN_INIT_SIZE` instead of the requested `num_rows`.

_Related to next comment_

kchibisov: No, it's not. num_rows could be larger then `MIN_INIT_SIZE`.

alacritty_terminal/src/grid/storage.rs
kchibisov: It does the goal is not to have 2 functions with the same body, so I'm just reusing `initialize_exact`. We're not wasting anything, we just preallocate when do a reallocation in storage.rs instead of doing this kind of things in grid/mod.rs.

alacritty_terminal/src/grid/storage.rs
```diff
@@ -173,25 +176,41 @@ impl<T> Storage<T> {
...
+            // Restore proper len
+            self.len = len + num_rows;
+        } else {
+            self.len += num_rows;
+        }
+    }
 
+    /// Dynamically grow the storage buffer at runtime.
+    #[inline]
+    pub fn initialize_exact(&mut self, num_rows: usize, template_row: Row<T>)
```
nixpulvis: Correct me if I'm wrong, but it seems like this function is more like `always_initialize`, since it doesn't check `self.len` vs `self.inner.len` before appending. Whereas, `initialize` is more like `maybe_initialize` since it may or may not actually allocate anything. I see where `_exact` is coming from, but I don't think it makes the most sense.

alacritty_terminal/src/grid/storage.rs
nixpulvis: I'm not following... If we allocate 1,000 new rows (`max(num_rows, MIN_INIT_SIZE)`), but `num_rows` is something less that 1,000, then we'll have left `self.len` smaller than the actual allocated range. Right?

alacritty_terminal/src/grid/storage.rs
kchibisov: Not really, these functions are like `reserve` and `reserve_exact` from std::vec::Vec. The goal of `initialize_exact` is not to be smart, like imagine just initializing 100 elements, knowing that you wont grow any larger then that, so initializing for MIN_INIT_SIZE is pointless.

alacritty_terminal/src/grid/storage.rs
kchibisov: >I'm not following... If we allocate 1,000 new rows (max(num_rows, MIN_INIT_SIZE)), but num_rows is something less that 1,000, then we'll have left self.len smaller than the actual allocated range. Right?

Yeah, that's the whole point. `self.len` is like a `len` on std::vec::Vec, but self.inner.len() is like a `capacity` on std::vec::Vec, so the goal is to grow capacity not by 1 at a time, but by some step `MIN_INIT_SIZE`, however you should grow your len by 1 here. This wasn't the case before, so your grid.len()(from grid/mod.rs) was a self.inner.len() (from storage.rs), which is wrong.

nixpulvis:
How necessary is `grid.initialize_all` being special anyway. It seems this should all be handled the same, as the `initialize` function. This function should appropriately update `self.len` and things should work. What am I missing?

alacritty_terminal/src/grid/storage.rs
nixpulvis: Ah I think I had `self.inner.len` and `self.len` backwards in my head. That makes this make more sense.

kchibisov:
>How necessary is grid.initialize_all being special anyway. It seems this should all be handled the same, as the initialize function. This function should appropriately update self.len and things should work. What am I missing?

You don't want to initialize additional things in `initialize_all`, so thats why we have 2 methods on a storage.rs now, like std::vec::Vec with reserve, etc. We just want to have a good abstraction for our ring buffer.

nixpulvis:
OK, if we need to expose `initialize_exact` for ref tests so be it. I don't see a problem with this implementation.

alacritty_terminal/src/grid/storage.rs
```diff
@@ -813,7 +832,7 @@ mod test {
         };
 
         // Initialize additional lines
-        storage.initialize(3, Row::new(Column(1), &'-'));
+        storage.initialize_exact(3, Row::new(Column(1), &'-'));
```
nixpulvis: This is the `initialize` test. Please rename this test, and add a new test for the behavior of the new `initialize` function.

nixpulvis: Something like:

```rust
    #[test]
    fn initialize() {
        // Setup storage area
        let mut storage = Storage {
            inner: vec![],
            zero: 0,
            visible_lines: Line(0),
            len: 0,
        };

        storage.initialize(3, Row::new(Column(0), &' '));

        assert_eq!(3, storage.len);
        assert_eq!(1_000, storage.inner.len());

        storage.initialize(3, Row::new(Column(0), &' '));

        assert_eq!(6, storage.len);
        assert_eq!(1_000, storage.inner.len());
    }
```

kchibisov: Well, this test make sense.

kchibisov:
I'll rerecord ref-test once we decided not to touch anything else.

chrisduerr:
> How necessary is grid.initialize_all being special anyway. It seems this should all be handled the same, as the initialize function. This function should appropriately update self.len and things should work. What am I missing?

Fundamentally that idea isn't completely off. Our `initialize_all` method is almost never called anyways and we do already overcommit with `initialize`, since we do not make sure that we don't initialize more than visible + history. So in theory it should be fine to just use `initialize` for both cases.

kchibisov:
>Fundamentally that idea isn't completely off. Our initialize_all method is almost never called anyways and we do already overcommit with initialize, since we do not make sure that we don't initialize more than visible + history. So in theory it should be fine to just use initialize for both cases.

That's true, but Storage will be lacking a way to precisely control its growth in this case, so I feel like we should add it for completeness anything, the usage of initialize_exact from initialize_all is just a "picking the right available method to solve specific task more efficiently" imho.

chrisduerr:
> That's true, but Storage will be lacking a way to precisely control its growth in this case

Yeah, but growth is never really an exact science, is it? Since we don't know the actual size? So you never know if it will actually initialize new things, because it's abstracted in the `storage.rs` file. Even with `initialize_exact`, right?

It's not about making sure that the **exact** number of lines is initialized, that should always be the case (as in actually occupied). It's about the initialized, but not occupied rows. So if you do `initialize` and then `initialize_exact`, you don't actually know the exact growth in initialization, but only the exact growth in occupation and even `initialize` grows the occupation exactly, it's only the initialization it does not.

So basically, the only difference between `initialize` and `initialize_exact` is that `initialize_exact` has an upper bound on *newly* added buffer lines, but all that should affect is memory usage. So if we decide `initialize` memory usage doesn't matter, why should we provide `initialize_exact`?

kchibisov:
>Yeah, but growth is never really an exact science, is it? Since we don't know the actual size? So you never know if it will actually initialize new things, because it's abstracted in the storage.rs file. Even with initialize_exact, right?

Well, I agree that it's an exact science, but you should have notion of control over implicit allocations. I think that's the case for rustc `reserve` methods. If you read their docs, their are not that precise, but they give you the notion of being more precise with reserve_exact.

>It's not about making sure that the exact number of lines is initialized, that should always be the case (as in actually occupied). It's about the initialized, but not occupied rows. So if you do initialize and then initialize_exact, you don't actually know the exact growth in initialization, but only the exact growth in occupation and even initialize grows the occupation exactly, it's only the initialization it does not.

That's true, the grows in occupation should be identical between the two functions, the only thing that differs here is grows in initialization.

>So basically, the only difference between initialize and initialize_exact is that initialize_exact has an upper bound on newly added buffer lines, but all that should affect is memory usage. So if we decide initialize memory usage doesn't matter, why should we provide initialize_exact?

The thing is that memory usage matters, but the function feels pointless, since you can't solve a problem with overallocationg on reaching cap, (e.g. having 101k lines allocated instead of 100k). So we should either change it somehow to be able to solve that case or just remove it and leave with a bit more memory consumption.

chrisduerr:
> The thing is that memory usage matters, but the function feels pointless, since you can't solve a problem with overallocationg on reaching cap, (e.g. having 101k lines allocated instead of 100k). So we should either change it somehow to be able to solve that case or just remove it and leave with a bit more memory consumption.

That's exactly my point, if `initialize` has an inaccuracy of `N` elements anyways, the `initialize_all` method isn't really necessary where we use it, since the general inaccuracy is still identical. Chances are even after calling `initialize_all`, we'll end up with `X + N` elements total in the buffer at some point. So then why would we even bother having an `*_exact` method, when it doesn't actually benefit us in any way?

The only place where `initialize_exact` would be used is ref-tests. So as long as the reftests do not fail, the memory usage doesn't matter the slightest bit.

kchibisov:
>The only place where initialize_exact would be used is ref-tests. So as long as the reftests do not fail, the memory usage doesn't matter the slightest bit.

Oh, some of them do fail without exact thing iirc.

alacritty_terminal/src/grid/storage.rs
```diff
@@ -832,8 +837,15 @@ mod test {
...
+        assert_eq!(
+            storage.inner[..shrinking_expected.zero],
+            shrinking_expected.inner[..shrinking_expected.zero]
+        );
+        assert_eq!(
+            storage.inner[storage.zero..],
+            shrinking_expected.inner[shrinking_expected.zero..]
+        );
+        assert_eq!(storage.zero, MIN_INIT_SIZE + 2);
         assert_eq!(storage.len, shrinking_expected.len);
```
kchibisov: I'm not sure about zero thing in this test now tbh.

kchibisov:
Should be fine now?

alacritty_terminal/src/grid/mod.rs
```diff
@@ -585,7 +580,7 @@ impl<T> Grid<T> {
 
     pub fn clear_history(&mut self) {
         // Explicitly purge all lines from history
-        let shrinkage = self.raw.len() - self.lines.0;
+        let shrinkage = self.history_size();
         self.raw.shrink_lines(shrinkage);
```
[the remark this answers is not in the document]
chrisduerr: I think this is still relevant?

alacritty_terminal/src/grid/storage.rs
```diff
@@ -173,27 +176,29 @@ impl<T> Storage<T> {
...
         self.inner.rotate_left(self.zero);
         self.inner.truncate(self.len);
 
         self.zero = 0;
     }
 
-    /// Dynamically grow the storage buffer at runtime
-    pub fn initialize(&mut self, num_rows: usize, template_row: Row<T>)
+    /// Dynamically grow the storage buffer at runtime.
+    pub fn initialize(&mut self, additional: usize, template_row: Row<T>)
```
chrisduerr:
```suggestion
    pub fn initialize(&mut self, additional_rows: usize, template_row: Row<T>)
```

I'd assume that the name `additional` was picked because of the parameters of the `reserve` method on `Vec`, but I think it might be a bit hard to understand what `additional` means in this context. I'm not 100% sure about this, but I feel like clarifying `additional_rows` might make things a little more clear?

alacritty_terminal/src/grid/storage.rs
```diff
@@ -173,27 +176,29 @@ impl<T> Storage<T> {
...
     where
         T: Clone,
     {
-        let mut new = vec![template_row; num_rows];
-
-        let mut split = self.inner.split_off(self.zero);
-        self.inner.append(&mut new);
-        self.inner.append(&mut split);
+        if self.len + additional > self.inner.len() {
```
chrisduerr: Adding a comment above this `if` is probably a good idea. Something like this:

```suggestion
        // Skip initializing new rows if enough are already buffered
        if self.len + additional > self.inner.len() {
```

alacritty_terminal/src/grid/storage.rs
chrisduerr: First of all I think there has been a copy-paste error here? It shouldn't be called `shrinking_expected`, but something different like `initialize_expected`, or probably better just `expected` or `expected_storage`.

I also think that we should probably modify the expected `Storage`, rather than our assertions. Having additional extra lines is part of our logic and thus part of what we should verify in our tests.

Lastly `MIN_INIT_SIZE + 2` as zero is a bit of a magic value, it should probably be `old_zero + MIN_INIT_SIZE` or something like that. Where `old_zero` is the previously defined initial `zero` offset before the `initialize` call.

alacritty_terminal/tests/ref/grid_reset/config.json
```diff
@@ -1 +1 @@
-{"history_size":100}
+{"history_size":100}
```
chrisduerr: There's no change to this file, so we shouldn't update it?

alacritty_terminal/tests/ref/grid_reset/size.json
```diff
@@ -1 +1 @@
-{"width":2532.0,"height":1380.0,"cell_width":9.0,"cell_height":19.0,"padding_x":3.0,"padding_y":3.0}
\ No newline at end of file
+{"width":1262.0,"height":1532.0,"cell_width":18.0,"cell_height":35.0,"padding_x":0.0,"padding_y":0.0,"dpr":2.0}
```
chrisduerr: Can you tell me why this reftest was re-recorded? Seems a bit strange to me.

alacritty_terminal/tests/ref/grid_reset/config.json
kchibisov: Well, alacritty produces a bit different format now, so it was just changed by `mv *.json alacritty_terminal/tests/ref/grid_reset/`. Maybe keeping new "format" make more sense. I can revert it if you insist though.

alacritty_terminal/tests/ref/grid_reset/size.json
kchibisov: It finds out that inner storage differs now... So the failure is a bit strange.
```
failures:

---- grid_reset stdout ----
thread 'grid_reset' panicked at 'Ref test failed; grid doesn't match', alacritty_terminal/tests/ref.rs:141:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.


failures:
    grid_reset

test result: FAILED. 33 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

```

However with some debug output I see.
```
term_grid: len: 172, num_cols: 280
grid: len: 0, num_cols: 280
```
I personally don't quite understand this `grid: len: 0 thing`, but the 172 kind of make sense.

alacritty_terminal/tests/ref/grid_reset/config.json
chrisduerr: I'd revert, just to keep the changes minimal. This change isn't necessary, so I'd rather not commit it.

alacritty_terminal/tests/ref/grid_reset/size.json
chrisduerr: We should find out what exactly is going on here, before doing anything else. Just re-recording reftests without understanding why they fail is not a good idea at all.

I'd assume this might be related to the original reason for initializing everything.

kchibisov: So basically the "problem" is that we're actually can call `truncate` now, since on the current master `self.inner.len()` and `self.len()` are identical, and a stupid impl when you allocate lines without some internal  step (like on this PR right) buffer you will also follow the current master behavior.

However on the current PR you can call `truncate` in `shrink_lines` quite easily now. The implementation of `truncate` seems legit and I don't get why leaving old content on a grid fixes it. There's nothing bad with it tbh, since you're likely reuse it afterwards, and it shouldn't change anything, but maybe I'm wrong...

During the test we're calling `initialize`, `rotate`, `shrink_lines`, and a `truncate`, so the problem could be in a one of those functions.

kchibisov:
I've rerecorded test on master with the following patch. On master `grid.len` is a bit broken, so we should truncate all garbage before `initialize_all`, otherwise will be matching garbage from self.inner in a rows from `initiallize_all`.

```diff
diff --git a/alacritty/src/event.rs b/alacritty/src/event.rs
index d03c439..b99c93a 100644
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -640,8 +640,9 @@ impl<N: Notify + OnResize> Processor<N> {

         // dump grid state
         let mut grid = terminal.grid().clone();
+        grid.truncate(true);
         grid.initialize_all(&Cell::default());
-        grid.truncate();
+        grid.truncate(false);

         let serialized_grid = json::to_string(&grid).expect("serialize grid");

diff --git a/alacritty_terminal/src/grid/mod.rs b/alacritty_terminal/src/grid/mod.rs
index 87de1d7..fdbc664 100644
--- a/alacritty_terminal/src/grid/mod.rs
+++ b/alacritty_terminal/src/grid/mod.rs
@@ -616,7 +616,10 @@ impl<T> Grid<T> {
     }

     /// This is used only for truncating before saving ref-tests
-    pub fn truncate(&mut self) {
+    pub fn truncate(&mut self, trun: bool) {
+        if trun {
+            self.raw.len = self.scroll_limit + self.lines.0;
+        }
         self.raw.truncate();
     }

diff --git a/alacritty_terminal/src/grid/storage.rs b/alacritty_terminal/src/grid/storage.rs
index 5d9339d..ab71dee 100644
--- a/alacritty_terminal/src/grid/storage.rs
+++ b/alacritty_terminal/src/grid/storage.rs
@@ -46,7 +46,7 @@ pub struct Storage<T> {
     /// having to truncate the raw `inner` buffer.
     /// As long as `len` is bigger than `inner`, it is also possible to grow the scrollback buffer
     /// without any additional insertions.
-    len: usize,
+    pub len: usize,
 }

 impl<T: PartialEq> PartialEq for Storage<T> {

```

alacritty_terminal/src/grid/storage.rs
```diff
@@ -831,8 +836,15 @@ mod test {
...
-        assert_eq!(storage.inner, shrinking_expected.inner);
-        assert_eq!(storage.zero, shrinking_expected.zero);
+        assert_eq!(
+            storage.inner[..shrinking_expected.zero],
+            shrinking_expected.inner[..shrinking_expected.zero]
+        );
+        assert_eq!(
+            storage.inner[storage.zero..],
+            shrinking_expected.inner[shrinking_expected.zero..]
+        );
```
nixpulvis: This only tests part of `inner`, for example in this test `storage.inner[..shrinking_expected.zero].len() == 5`, and there are more than 5 elements to check for. I'm not sure if there's a super elegant way to test from `zero` to `zero+len` without some way to wrap around.

kchibisov: I've reverted a fix for this test at some point...

kchibisov: Should be fine now @nixpulvis

alacritty_terminal/src/grid/storage.rs
```diff
@@ -817,35 +817,35 @@ mod test {
...
-        assert_eq!(
-            storage.inner[storage.zero..],
-            shrinking_expected.inner[shrinking_expected.zero..]
-        );
-        assert_eq!(storage.zero, MIN_INIT_SIZE + 2);
-        assert_eq!(storage.len, shrinking_expected.len);
+
+        assert_eq!(storage.inner, expected_storage.inner);
+        assert_eq!(storage.zero, expected_storage.zero);
+        assert_eq!(storage.len, expected_storage.len);
```
nixpulvis: Much better! :tada:

nixpulvis:
This seems like a current strange thing: https://github.com/jwilm/alacritty/blob/5651c3f7114dff611e616865ad02e682779979d9/alacritty_terminal/src/grid/mod.rs#L240-L250.

We set `self.scroll_limit` unconditionally, but then only `self.raw.initialize(...)` based on `history_size < self.scroll_limit`. Ideally this would be cleaned up with what @chrisduerr is talking about I believe.

kchibisov:
I'll try to update PR soonish then.

chrisduerr:
> We set self.scroll_limit unconditionally, but then only self.raw.initialize(...) based on history_size < self.scroll_limit. Ideally this would be cleaned up with what @chrisduerr is talking about I believe.

I actually ran into the problem with `scroll_limit` when trying to refactor this little piece of code.

I believe this conditional is only for catching the case where `scroll_limit == max_scroll_limit`, so we don't initialize with `new == 0`.

Ideally, we'd just early return if `grid.history_size() >= grid.max_scroll_limit`.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -346,8 +341,9 @@ impl<T: GridCell + PartialEq + Copy> Grid<T> {
 
                     // Remove one line from scrollback, since we just moved it to the
                     // viewport
-                    self.scroll_limit = self.scroll_limit.saturating_sub(1);
-                    self.display_offset = min(self.display_offset, self.scroll_limit);
+                    // self.decrease_scroll_limit(1);
+                    history_size -= 1;
+                    self.display_offset = min(self.display_offset, history_size);
```
kchibisov: I feel like this is not the right way to do it.

kchibisov:
Reflow seems broken a bit as far as I can see. It seems like I'm removing some lines from the bottom in addition.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -125,7 +123,8 @@ pub struct Grid<T> {
     /// updates this offset accordingly.
     display_offset: usize,
 
-    /// An limit on how far back it's possible to scroll
+    /// Legacy value, however we can't remove it, since it's in all ref-tests
+    #[serde(skip)]
     scroll_limit: usize,
```
chrisduerr: We shouldn't keep around a field that has no purpose.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -62,7 +60,7 @@ impl<T: PartialEq> ::std::cmp::PartialEq for Grid<T> {
             && self.cols.eq(&other.cols)
             && self.lines.eq(&other.lines)
             && self.display_offset.eq(&other.display_offset)
-            && self.scroll_limit.eq(&other.scroll_limit)
+            && self.history_size().eq(&other.history_size())
```
chrisduerr: Do we need this comparison? It should be automatically assured when the raw buffer is compared, right? So I'm not sure if this can ever be `false`.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -188,25 +187,27 @@ impl<T: GridCell + PartialEq + Copy> Grid<T> {
     pub fn update_history(&mut self, history_size: usize, template: &T) {
         self.raw.update_history(history_size, Row::new(self.cols, &template));
         self.max_scroll_limit = history_size;
-        self.scroll_limit = min(self.scroll_limit, history_size);
-        self.display_offset = min(self.display_offset, self.scroll_limit);
+        if history_size < self.history_size() {
+            self.decrease_scroll_limit(self.history_size() - history_size);
+        }
```
chrisduerr: Does it make sense to call `decrease_scroll_limit` here, or should all changes to the raw buffer be made in `update_history`?

alacritty_terminal/src/grid/mod.rs
```diff
@@ -238,21 +239,14 @@ impl<T: GridCell + PartialEq + Copy> Grid<T> {
...
-            self.raw.initialize(new, Row::new(self.cols, template));
-        }
+        let new = min(count, self.max_scroll_limit - self.history_size());
+        self.raw.initialize(new, Row::new(self.cols, template));
     }
 
     fn decrease_scroll_limit(&mut self, count: usize) {
-        self.scroll_limit = self.scroll_limit.saturating_sub(count);
+        let history_size = self.history_size();
+        let count = history_size - history_size.saturating_sub(count);
```
chrisduerr: I'd assume the `history_size - history_size.saturating_sub(count)` math is here just to prevent the case where `count > history_size`? Does this actually ever happen?

Otherwise just `count` itself should be sufficient.

alacritty_terminal/src/grid/mod.rs
chrisduerr: What happens if you just `decrease_scroll_limit`? Does that cause issues since `self.raw` is drained?

alacritty_terminal/src/grid/mod.rs
```diff
@@ -458,7 +455,7 @@ impl<T: GridCell + PartialEq + Copy> Grid<T> {
                     row = Row::from_vec(wrapped, occ);
 
                     // Increase scrollback history
-                    self.scroll_limit = min(self.scroll_limit + 1, self.max_scroll_limit);
+                    history_size = min(history_size + 1, self.max_scroll_limit);
```
chrisduerr: It seems like you're just modifying a variable that is never used? Since the `self.history_size()` returns a copy and not a mutable reference.

So this is buggy.

alacritty_terminal/src/grid/mod.rs
kchibisov: `scroll_down` unit test was failing without this logic.

chrisduerr: Might be a good opportunity to validate that the test makes sense. But we can do that when all the other issues have been resolved.

Probably makes sense to resolve the resizing issues first.

alacritty_terminal/src/grid/mod.rs
kchibisov: I know that's why I asked that I don't really know hot to update this thing in a more better way, since you're speculating on a storage, and self.history() is also speculating on a history, so you have 2 mut refs that can update the storage/invalidate interator(not in rust). Also you're updating the entire storage in the end anyway, so I thought that this scroll_limit thing was just useful only in a context of this `for loop`, since I don't see how to update it after the storage, and I don't see any sign, that this introduce a bug.

alacritty_terminal/src/grid/mod.rs
kchibisov: I was following the old code, the old code was updating scroll_limit to a new value, so it was decreasing it, so this call makes sense from the point of the old code.

alacritty_terminal/src/grid/mod.rs
kchibisov: It could be false, if your len on a storage defer, since we're matching only by inner.len() etc.

alacritty_terminal/src/grid/mod.rs
chrisduerr: I'd have to test things myself, but maybe it works to increase/decrease the scroll limit after all lines are processed.

Alternatively it might need some slight tweaks to the logic, but it shouldn't be that big of a problem since we're creating a new buffer anyways. So we have good control over the raw storage here.

alacritty_terminal/src/grid/mod.rs
chrisduerr: Well if it is **required** that the raw buffer is modified when the history is shrunk, it might make sense to enforce that in the `storage.rs` file with the `update_history` method.

My only question would be if it actually is possible to do things there, or if it lacks some information that only the grid can provide to `shrink_lines`.

But having to call two separate methods on `grid.raw` to perform one operation is usually a bad thing and will cause bugs in the usage of the API.

alacritty_terminal/src/grid/mod.rs
kchibisov: Yes, you have 2 mut refs, so you're iterating over inner storage, and in the same time decrease_scroll_limit can completely change the internal structure of the internal storage, so the only way to to accumulate storage update in a loop and then update storage. However I don't know whether it's correct way to do it or not right now.

alacritty_terminal/src/grid/mod.rs
chrisduerr: Shouldn't we check the `len` in storage then?

chrisduerr: Shouldn't we check the `len` in `storage.rs` then?

kchibisov:
>After some more looking into the column resizing, I've noticed that the history size is not tracked correctly.

>This is best tested by setting the history size to 0 in the config and shrinking the terminal. This will lead to grid.history_size() of the main grid to report a value bigger than 0, which shouldn't be possible.

>Since the history size is tracked in the storage.rs by number of elements in the grid with the len property and we just create the new storage from a Vec, that means that if we push too many rows, we'll increase the history size, since the new len will be equal to Vec::len.

>We should probably make sure that grid.history_size() will never grow beyond grid.max_scroll_limit just to keep our sanity in future refactorings. While we could just trim the len accordingly (limit it to max_scroll_limit, automatically dropping everything after that), it might make sense to look into dropping lines from the Vec or not even adding them, since that might actually improve performance.

Could you elaborate a bit more on how should I find out it? I wasn't able to make grid.history be bigger than the max_scroll_limit no matter how I try. I was using the following diff and resizing thing randomly.

```diff
diff --git a/alacritty_terminal/src/grid/mod.rs b/alacritty_terminal/src/grid/mod.rs
index 9f32d7a..5ebead5 100644
--- a/alacritty_terminal/src/grid/mod.rs
+++ b/alacritty_terminal/src/grid/mod.rs
@@ -214,6 +214,9 @@ impl<T: GridCell + PartialEq + Copy> Grid<T> {
         cursor_pos: &mut Point,
         template: &T,
     ) {
+        if self.history_size() != 0 {
+            panic!("HOW SO? history_size {}", self.history_size());
+        }
         // Check that there's actually work to do and return early if not
         if lines == self.lines && cols == self.cols {
             return;
```

alacritty_terminal/src/grid/mod.rs
kchibisov: I think we can.

alacritty_terminal/src/grid/mod.rs
kchibisov: Agree, so we should update all ref-tests to remove this value.

alacritty_terminal/src/grid/mod.rs
kchibisov: storage.rs doesn't know what history is, it's just a ring buffer, so let's just keep logic that doesn't make sense to it in grid.

alacritty_terminal/src/grid/mod.rs
kchibisov: Well the test makes sense tbh. You shouldn't try to scroll if you reached the end(0).

chrisduerr: I'm just wondering if there's a more transparent way to do this, since this math seems a bit confusing.

I'd propose using the following instead:

```rust
let count = max(count, self.history_size());
```

I think that's a lot more clear about what we're actually doing, right?

alacritty_terminal/src/grid/mod.rs
chrisduerr: Of course `storage.rs` knows what history is, that's why we have a field called `visible_lines` on `Storage`. We always know that `len - visible_lines` is the history.

We're already calling `shrink_lines(old_history - new_history)` inside `update_history` in `storage.rs`. So if it turns out that `history_size < self.history_size()` after that, there's a bug somewhere in the `update_history` function.

This code should not be necessary and removing it shouldn't do anything.

alacritty_terminal/src/grid/mod.rs
chrisduerr: Yeah, we should update all of them just to get rid of it.

alacritty_terminal/src/grid/mod.rs
chrisduerr: We should do this, querying `self.history_size()` in this function does not make sense.

kchibisov:
During running `scrolling` benchmark it was discovered that commit c90f23273b0a19ab633882188a121c582f489a91 introduced a massive performance regression compared to master and to a0d059b769efd80ce13c3949e4d939be29c541a3 . The a0d059b769efd80ce13c3949e4d939be29c541a3 is identical to master performance wise.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -645,19 +640,25 @@ impl<T> Grid<T> {
...
-        let history_size = self.raw.len().saturating_sub(*self.lines);
-        self.raw.initialize(self.max_scroll_limit - history_size, template, self.cols);
+        // Remove all cached lines to clear them of any content
+        self.truncate();
+
+        // Initialize everything with empty new lines
+        self.raw.initialize(self.max_scroll_limit - self.history_size(), template, self.cols);
     }
 
     /// This is used only for truncating before saving ref-tests
```
cole-h: I don't believe this comment holds true anymore, now that we're using it in `initialize_all` above.

chrisduerr: This is a new comment I've added in this PR. I'm not sure what you mean?

cole-h: Brainfart. Move along, nothing to see here.

alacritty_terminal/src/grid/storage.rs
```diff
@@ -51,81 +52,40 @@ pub struct Storage<T> {
...
-        //     7  8  9  | 0  1  2  3  | 4  5  6
-        //     C3 C3 C3 | C1 C1 C1 C1 | C2 C2 C2
-        let len = self.inner.len();
-        bigger.inner[bigger_zero..]
-            == smaller.inner[smaller_zero..smaller_zero + (len - bigger_zero)]
-            && bigger.inner[..bigger_zero - smaller_zero]
-                == smaller.inner[smaller_zero + (len - bigger_zero)..]
-            && bigger.inner[bigger_zero - smaller_zero..bigger_zero]
-                == smaller.inner[..smaller_zero]
+        self.inner == other.inner && self.len == other.len
```
kchibisov: I don't really like this implementation tbh, since it doesn't make sense to me, that something should be truncated before `eq`. However since `eq` is only used by our tests, maybe it's not a bad idea to keep things simple here, but in this case we should state it.

chrisduerr: It's just unnecessary complexity if we never use it without truncation anyways.

We're also not really implementing a full equality check properly, but when it makes sense to compare them, this is the most reasonable way.

kchibisov: I mean we should state that this thing is only for tests, right?

alacritty_terminal/src/grid/mod.rs
```diff
@@ -83,11 +83,11 @@ pub trait GridCell {
 /// │                         │
 /// │      UNINITIALIZED      │
 /// │                         │
-/// ├─────────────────────────┤  <-- raw.len()
+/// ├─────────────────────────┤  <-- raw.inner.len()
 /// │                         │
 /// │      RESIZE BUFFER      │
 /// │                         │
-/// ├─────────────────────────┤  <-- scroll_limit + lines
+/// ├─────────────────────────┤  <-- self.history_size()
```
kchibisov: This is not a self.history_size(), it's a self.len().

chrisduerr: You're absolutely correct.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -232,21 +221,12 @@ impl<T: GridCell + PartialEq + Copy> Grid<T> {
...
-        let history_size = self.raw.len().saturating_sub(*self.lines);
-        if history_size < self.scroll_limit {
-            let new = min(
-                max(self.scroll_limit - history_size, MIN_INIT_SIZE),
-                self.max_scroll_limit - history_size,
-            );
-            self.raw.initialize(new, Row::new(self.cols, template));
-        }
+        let new = min(count, self.max_scroll_limit - self.history_size());
+        self.raw.initialize(new, template, self.cols);
```
kchibisov: Actually, I think like it'll be a good idea to add a check for zero and don't call this function to initialize zero rows, what do you think? I feel like this was the original issue about perf I've faced, since I forgot to add a proper check for zero here, and once you added a way to move row and do this logic on initialize you basically avoided it.

PR #3117, opened 2019-12-22
Clean up string handling in Windows PTY implementations.

davidhewitt:
Removes `widestring` and `dunce` dependencies, reduces some code duplication and cleans up a few minor pieces regarding `Path` / `PathBuf`.

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -254,12 +254,12 @@ impl Options {
     }
 
     pub fn into_config(self, mut config: Config) -> Config {
-        config.set_live_config_reload(
-            self.live_config_reload.unwrap_or_else(|| config.live_config_reload()),
-        );
-        config.set_working_directory(
-            self.working_dir.or_else(|| config.working_directory().to_owned()),
-        );
+        if let Some(lcr) = self.live_config_reload {
+            config.set_live_config_reload(lcr);
+        }
+        if let Some(wd) = self.working_dir {
+            config.set_working_directory(Some(wd));
+        }
         config.shell = self.command.or(config.shell);
 
         config.hold = self.hold;
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -210,7 +210,7 @@ impl<T> Config<T> {
 
     #[inline]
     pub fn working_directory(&self) -> Option<&Path> {
-        // TODO: On Rust 1.40 use self.working_directory.deref()
+        // TODO: On Rust 1.40 use self.working_directory.as_deref()
         self.working_directory.as_ref().map(|path_buf| path_buf.as_ref())
     }
 
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -210,8 +210,7 @@ impl<T> Config<T> {
 
     #[inline]
     pub fn working_directory(&self) -> Option<&Path> {
-        // TODO: On Rust 1.40 use self.working_directory.as_deref()
-        self.working_directory.as_ref().map(|path_buf| path_buf.as_ref())
+        self.working_directory.as_ref().map(|path_buf| path_buf.as_path())
     }
 
     #[inline]
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -15,7 +15,7 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::fmt::Display;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use log::error;
 use serde::{Deserialize, Deserializer};
@@ -209,8 +209,9 @@ impl<T> Config<T> {
     }
 
     #[inline]
-    pub fn working_directory(&self) -> &Option<PathBuf> {
-        &self.working_directory
+    pub fn working_directory(&self) -> Option<&Path> {
+        // TODO: On Rust 1.40 use self.working_directory.deref()
+        self.working_directory.as_ref().map(|path_buf| path_buf.as_ref())
     }
 
     #[inline]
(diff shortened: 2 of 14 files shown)


alacritty/src/cli.rs
```diff
@@ -254,12 +254,12 @@ impl Options {
...
-        );
-        config.set_working_directory(
-            self.working_dir.or_else(|| config.working_directory().to_owned()),
-        );
+        if let Some(lcr) = self.live_config_reload {
+            config.set_live_config_reload(lcr);
+        }
+        if let Some(wd) = self.working_dir {
+            config.set_working_directory(Some(wd));
+        }
```
davidhewitt: After I changed `config.working_directory()` to return `Option<&Path>` instead of `&Option<PathBuf>`, this needed tweaking too. I figured this new implementation is better as it's no longer copying `config.working_directory()` if `self.working_dir` is None.

alacritty_terminal/src/config/mod.rs
```diff
@@ -209,8 +209,9 @@ impl<T> Config<T> {
     }
 
     #[inline]
-    pub fn working_directory(&self) -> &Option<PathBuf> {
-        &self.working_directory
+    pub fn working_directory(&self) -> Option<&Path> {
+        // TODO: On Rust 1.40 use self.working_directory.deref()
+        self.working_directory.as_ref().map(|path_buf| path_buf.as_ref())
```
davidhewitt: `Option<&Path>` is one pointer indirection less than `&Option<PathBuf>`, so I think it's better style to return it in this case.

Regarding the TODO, I guess we are some way off setting a minimum Rust version of 1.40 so we can use `Option::deref()`? I also considered using the `rustversion` crate instead of a TODO.

alacritty_terminal/src/util.rs
```diff
@@ -94,6 +94,15 @@ where
         .map(|_| ())
 }
 
+// Converts the string slice into a Windows-standard representation for "W"-
+// suffixed function variants, which accept UTF-16 encoded string values.
+#[cfg(windows)]
+pub fn win32_string<S: AsRef<OsStr> + ?Sized>(value: &S) -> Vec<u16> {
```
davidhewitt: I'm generally not a fan of `util` modules as they easily grow to contain all sorts of random functions. Is there a better home for this? (Moved from `panic.rs` so that `tty` can use it too.)

sterlingjensen: This is something I've been trying to figure out as well, hoping to corral things prior to a `winpty` feature gate in order to cut down on the number of `#[cfg(...)]`'s that will demand. I'd suggest something like `sys/$OS` mod + `impl`, but that might require a big bang development cycle to even sell.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -205,19 +203,11 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, _window_id: Option<usize>) ->
...
     }
 
-    // Get process commandline
-    let default_shell = &Shell::new("powershell");
-    let shell = config.shell.as_ref().unwrap_or(default_shell);
-    let mut cmdline = shell.args.clone();
-    cmdline.insert(0, shell.program.to_string());
-
-    // Warning, here be borrow hell
-    let cwd = config.working_directory().as_ref().map(|dir| canonicalize(dir).unwrap());
```
davidhewitt: As far as I can tell `dunce::canonicalize` exists for applications that do not handle the UNC paths that `std::fs::canonicalize()` creates.

The docs for [`CreateProcessW`](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) specifically mention the working directory can be a UNC path. Which is why we can safely remove `dunce`.

alacritty_terminal/src/config/mod.rs
chrisduerr: Why not just use `self.warking_directory.map(|path_buf| path_buf.as_path())`?

I think that's more descriptive than both the existing and the proposed 1.40 version and it should work just as well?

alacritty_terminal/src/util.rs
chrisduerr: I think even `tty/windows/mod.rs` would be a better place than `util.rs`. That's the only real place where we have a big difference between Windows and other OSs anyways.

Though as usual, I think the imports should be moved to the top of the file.

winpty/src/windows.rs
```diff
@@ -283,25 +286,28 @@ impl SpawnConfig {
...
         cmdline: Option<&str>,
-        cwd: Option<&str>,
-        end: Option<&str>,
+        cwd: Option<&Path>,
+        env: Option<&str>,
     ) -> Result<Self, Error> {
         let mut err = null_mut() as *mut winpty_error_t;
 
-        let to_wstring = |s| WideCString::from_str(s).unwrap();
+        fn to_wstring(s: &(impl AsRef<OsStr> + ?Sized)) -> Vec<u16> {
```
chrisduerr: I think a type parameter would likely be much more readable than `impl Trait` here. Would also save us the parenthesis.

alacritty_terminal/src/config/mod.rs
davidhewitt: Can do. Though it would have to be 
 
   self.working_directory.as_ref().map(|path_buf| path_buf.as_path())

Because `Option::map` consumes `self`.

chrisduerr: Sounds good to me.

davidhewitt:
@chrisduerr anything else you want from me to get this merged?

chrisduerr:
> anything else you want from me to get this merged?

It's approved, so I don't think so. I'm not sure what happened here, I must have left because CI was still running and then never came back to it.

Thanks for the bump, LGTM.

PR #3121, opened 2019-12-22
Add minimize action

### LINKED ISSUE #2534: ⌘+m does not minimize window
macOS Mojave

Not a big deal but for some reason hitting <kbd>⌘</kbd>+<kbd>m</kbd> does not minimize Alacritty's window. It just sends the character `m` to stdin.

Maybe this is fixable by adding a custom key binding?

kchibisov:
Fixes #2534

PR is blocked by winit bump, however it's reviewable, since after winit bump we should change only 1 name(we should call `set_minimized(minimized)` instead of `
set_maximized(minimized)` in `alacritty/src/window.rs`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - /Applications symlink into OS X DMG for easier installation
 - Colored emojis on Linux/BSD
 - Value `randr` for `WINIT_HIDPI_FACTOR`, to ignore `Xft.dpi` and scale based on screen dimensions
-- `Minimize` action
+- `Minimize` key binding action, bound to `cmd + m` on macOS
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - /Applications symlink into OS X DMG for easier installation
 - Colored emojis on Linux/BSD
 - Value `randr` for `WINIT_HIDPI_FACTOR`, to ignore `Xft.dpi` and scale based on screen dimensions
+- `Minimize` action
 
 ### Changed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -531,7 +531,7 @@
 # be mapped to the `ReceiveChar` action. Alternatively, you can use `None` for
 # a no-op if you do not wish to receive input characters for that binding.
 #key_bindings:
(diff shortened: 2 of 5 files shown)


alacritty.yml
```diff
@@ -531,7 +532,7 @@
 # be mapped to the `ReceiveChar` action. Alternatively, you can use `None` for
 # a no-op if you do not wish to receive input characters for that binding.
 #key_bindings:
-  # (Windows/Linux only)
+  # (Windows and Linux only)
```
kchibisov: I feel like it'll be a good idea to separate windows only bindings and provide a common section for Linux and Windows bindings.

alacritty.yml
```diff
@@ -532,7 +532,7 @@
 # be mapped to the `ReceiveChar` action. Alternatively, you can use `None` for
 # a no-op if you do not wish to receive input characters for that binding.
 #key_bindings:
-  # (Windows and Linux only)
+  # (Windows, Linux and BSD only)
```
cole-h: nit: Oxford comma, please

`(Windows, Linux and BSD only)` -> `(Windows, Linux**,** and BSD only)`

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Terminal escape bindings with combined modifiers for Delete and Insert
 - /Applications symlink into OS X DMG for easier installation
 - Colored emojis on Linux/BSD
+- `Minimize` action
```
chrisduerr: We should probably go into a bit more detail here, I doubt anyone would understand what this means by just reading the change log entry.

```suggestion
- `Minimize` key binding action, bound to `cmd + m` on macOS
```

chrisduerr:
Once winit is bumped, we also need to add change log entries for the other things that have been fixed and close the appropriate issues.

kchibisov:
@chrisduerr I'd like to bump winit separately, since it closes a lot of unrelated issues, etc.

CHANGELOG.md
kchibisov: I was just using similar description to the one from `v0.3.3` added section about `ToggleFullscreen` action.

I'll update once Windows situation becomes more clear.

chrisduerr:
> I'd like to bump winit separately, since it closes a lot of unrelated issues, etc.

Yeah, good idea.

chrisduerr:
This only works when decorations are enabled. See https://github.com/rust-windowing/winit/issues/1361.

I've created a follow-up issue here: https://github.com/jwilm/alacritty/issues/3150

PR #3122, opened 2019-12-23
Fix screen reset not clearing cell flags

chrisduerr:
TODO:
 - [x] Add reftest


kchibisov:
Could you add a ref test?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1785,6 +1785,10 @@ impl<T: EventListener> ansi::Handler for Term<T> {
                 if self.mode.contains(TermMode::ALT_SCREEN) {
                     self.grid.region_mut(..).each(|c| c.reset(&template));
                 } else {
+                    let template = Cell {
+                        c: template.c,
```
kchibisov: Do we need a background color?

chrisduerr: Oh, I completely forgot that. Of course we do. Good catch, thanks.

chrisduerr:
> Could you add a ref test?

Good idea, will add one before merging.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1785,6 +1785,11 @@ impl<T: EventListener> ansi::Handler for Term<T> {
...
```
kchibisov: Oh, could you elaborate a bit on using character from template here? I do feel like we don't need it, by this I mean that default should be fine.

chrisduerr: It was a bit late yesterday and I copied the code from cell's reset method in here, which sets both char and bg. Then since we don't need the char I removed the background.

So this has been a bit of a mixup, we obviously don't need the char, but only the background. Will fix it when adding the reftest.

kchibisov:
LGTM, could you just run `rustfmt` before merging?

PR #3126, opened 2019-12-26
Add Text Objects

rodolf0:
This adds 'Text Objects' which lift plain text into interactive items that
can trigger actions on click. Similar to URLs but with customization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Colored emojis on Linux/BSD
 - Value `randr` for `WINIT_HIDPI_FACTOR`, to ignore `Xft.dpi` and scale based on screen dimensions
 - `Minimize` key binding action, bound to `cmd + m` on macOS
+- Text-Objects for customizable interactive items to click and launch actions.
 
 ### Changed
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,6 +39,7 @@ dependencies = [
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "notify 4.0.14 (registry+https://github.com/rust-lang/crates.io-index)",
  "parking_lot 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustc_tools_util 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.104 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.44 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -408,6 +408,25 @@
     # on them. The available modifiers are documented in the key binding section.
     #modifiers: None
 
+#text_objects:
+  # These user defined 'text-objects' lift plain text into interactive triggers similar to URLs.
+  #
+  # Each text-object entry has a unique name, a `search` regex to find objects in display,
+  # an `action` to be executed when clicked, and optionally `priority` in case of multiple matches.
+  #
+  # `action` is a list where the 1st element is the program to execute and the rest are arguments.
+  # The last element in the `action` list will be used as a template filled in with captures from
+  # the `search` regex to customize the action. $0 replaces the full object, $1 the first capture.
+
+  #GNOME issues:
+  # search: "((?i)gnome|community|world)/([-+./a-z]+)#([[:digit:]]+)"
+  # action: ["xdg-open", "https://gitlab.gnome.org/$1/$2/issues/$3"]
+  # priority: 2
+
+  #Freedesktop issues:
+  # search: "((?i)xorg|wayland|PulseAudio)/([-+./[:alnum:]]+)#([[:digit:]]+)"
+  # action: ["xdg-open", "https://gitlab.freedesktop.org/$1/$2/issues/$3"]
+
 # Mouse bindings
 #
 # Mouse bindings are specified as a list of objects, much like the key
--- a/alacritty/Cargo.toml
+++ b/alacritty/Cargo.toml
@@ -25,6 +25,7 @@ unicode-width = "0.1"
 parking_lot = "0.9"
 font = { path = "../font" }
 urlocator = "0.1.0"
+regex = "1"
 
 [build-dependencies]
 gl_generator = "0.14.0"
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -16,6 +16,7 @@ use alacritty_terminal::config::{Config as TermConfig, LOG_TARGET_CONFIG};
 mod bindings;
 pub mod monitor;
 mod mouse;
+pub mod text_objects;
 mod ui_config;
 
 pub use crate::config::bindings::{Action, Binding, Key, RelaxedEq};
--- a/alacritty/src/config/text_objects.rs
+++ b/alacritty/src/config/text_objects.rs
@@ -1,0 +1,45 @@
+use log::error;
+use regex::bytes::Regex;
+use serde::{Deserialize, Deserializer};
+
+use alacritty_terminal::config::{failure_default, LOG_TARGET_CONFIG};
+
+#[serde(default)]
+#[derive(Clone, Debug, Deserialize)]
+pub struct TextObject {
+    #[serde(deserialize_with = "deserialize_regex")]
+    pub search: Option<Regex>,
+    #[serde(deserialize_with = "failure_default")]
+    pub action: Vec<String>,
+    #[serde(deserialize_with = "failure_default")]
+    pub priority: usize,
+}
+
+impl Default for TextObject {
+    fn default() -> Self {
+        TextObject { search: None, action: Vec::new(), priority: 0 }
+    }
+}
+
+impl PartialEq for TextObject {
+    // Simple equality for Regex so we add them to containers
+    fn eq(&self, other: &Self) -> bool {
+        self.action == other.action
+            && self.priority == other.priority
+            && self.search.as_ref().map(|v| v.as_str()) == other.search.as_ref().map(|v| v.as_str())
+    }
+}
+
+fn deserialize_regex<'a, D>(deserializer: D) -> Result<Option<Regex>, D::Error>
+where
+    D: Deserializer<'a>,
+{
+    let raw = serde_yaml::Value::deserialize(deserializer)?;
+    if let Some(val) = raw.as_str() {
+        if let Ok(re) = Regex::new(val) {
+            return Ok(Some(re));
+        }
+    }
+    error!(target: LOG_TARGET_CONFIG, "Problem with text-object: {:?}; disabled", raw);
+    Ok(None)
+}
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -1,14 +1,20 @@
 use serde::{Deserialize, Deserializer};
+use std::collections::HashMap;
 
 use alacritty_terminal::config::failure_default;
 
 use crate::config::bindings::{self, Binding, KeyBinding, MouseBinding};
 use crate::config::mouse::Mouse;
+use crate::config::text_objects::TextObject;
 
 #[derive(Debug, PartialEq, Deserialize)]
 pub struct UIConfig {
     #[serde(default, deserialize_with = "failure_default")]
     pub mouse: Mouse,
+
+    /// Custom text-object link handlers
+    #[serde(default, deserialize_with = "failure_default")]
+    pub text_objects: HashMap<String, TextObject>,
 
     /// Keybindings
     #[serde(default = "default_key_bindings", deserialize_with = "deserialize_key_bindings")]
@@ -23,6 +29,7 @@ impl Default for UIConfig {
     fn default() -> Self {
         UIConfig {
             mouse: Mouse::default(),
+            text_objects: HashMap::new(),
             key_bindings: default_key_bindings(),
             mouse_bindings: default_mouse_bindings(),
         }
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -366,6 +366,7 @@ impl Display {
         config: &Config,
         mouse: &Mouse,
         mods: ModifiersState,
+        draw_text_objects: bool,
     ) {
         let grid_cells: Vec<RenderableCell> = terminal.renderable_cells(config).collect();
         let visual_bell_intensity = terminal.visual_bell.intensity();
@@ -382,7 +383,7 @@ impl Display {
         self.window.update_ime_position(&terminal, &self.size_info);
 
         // Only get export display content if text-objects are configured
-        let (display_text, display_cells) = if config.ui_config.text_objects.is_empty() {
+        let (display_text, display_cells) = if draw_text_objects && config.ui_config.text_objects.is_empty() {
             (Vec::new(), Vec::new())
         } else {
             terminal.display_index()
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -42,6 +42,7 @@ use crate::config::Config;
 use crate::event::{DisplayUpdate, Mouse};
 use crate::renderer::rects::{RenderLines, RenderRect};
 use crate::renderer::{self, GlyphCache, QuadRenderer};
+use crate::text_objects::{DisplayTextObjects, DisplayTxtObj};
 use crate::url::{Url, Urls};
 use crate::window::{self, Window};
 
@@ -110,10 +111,14 @@ impl From<glutin::ContextError> for Error {
 pub struct Display {
     pub size_info: SizeInfo,
     pub window: Window,
-    pub urls: Urls,
 
     /// Currently highlighted URL.
+    pub urls: Urls,
     pub highlighted_url: Option<Url>,
+
+    /// Detected objects in visible display
+    pub text_objects: DisplayTextObjects,
+    pub highlighted_txob: Option<DisplayTxtObj>,
 
     renderer: QuadRenderer,
     glyph_cache: GlyphCache,
@@ -240,6 +245,8 @@ impl Display {
             size_info,
             urls: Urls::new(),
             highlighted_url: None,
+            text_objects: DisplayTextObjects::new(),
+            highlighted_txob: None,
         })
     }
 
@@ -374,6 +381,13 @@ impl Display {
         #[cfg(not(windows))]
         self.window.update_ime_position(&terminal, &self.size_info);
 
+        // Only get export display content if text-objects are configured
+        let (display_text, display_cells) = if config.ui_config.text_objects.is_empty() {
+            (Vec::new(), Vec::new())
+        } else {
+            terminal.display_index()
+        };
+
         // Drop terminal as early as possible to free lock
         drop(terminal);
 
@@ -404,6 +418,27 @@ impl Display {
         }
 
         let mut rects = lines.rects(&metrics, &size_info);
+
+        // Find text-objects in freshly drawn display
+        self.text_objects.find_objects(
+            &config.ui_config.text_objects,
+            &display_text,
+            &display_cells,
+        );
+
+        // Highlight text-objects if any under the mouse cursor
+        if let Some(txob) = self.text_objects.highlighted(&mouse, mods, mouse_mode, selection) {
+            rects.append(&mut txob.rects(&metrics, &size_info));
+            self.highlighted_txob = Some(txob);
+            self.window.set_mouse_cursor(CursorIcon::Hand);
+        } else if self.highlighted_txob.is_some() {
+            self.highlighted_txob = None;
+            if mouse_mode {
+                self.window.set_mouse_cursor(CursorIcon::Default);
+            } else {
+                self.window.set_mouse_cursor(CursorIcon::Text);
+            }
+        }
 
         // Update visible URLs
         self.urls = urls;
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -245,7 +245,7 @@ impl Display {
             size_info,
             urls: Urls::new(),
             highlighted_url: None,
-            text_objects: DisplayTextObjects::new(),
+            text_objects: DisplayTextObjects::new(&config.ui_config.text_objects),
             highlighted_txob: None,
         })
     }
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -7,7 +7,7 @@ use std::fs;
 use std::fs::File;
 use std::io::Write;
 use std::sync::Arc;
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 use glutin::dpi::PhysicalSize;
 use glutin::event::{ElementState, Event as GlutinEvent, ModifiersState, MouseButton};
@@ -342,7 +342,9 @@ impl<N: Notify + OnResize> Processor<N> {
     where
         T: EventListener,
     {
+        let txob_refresh_freq = Duration::new(0, 50_000_000);
         let mut event_queue = Vec::new();
+        let mut next_txob_draw = Instant::now();
 
         event_loop.run_return(|event, _event_loop, control_flow| {
             if self.config.debug.print_events {
@@ -358,7 +360,6 @@ impl<N: Notify + OnResize> Processor<N> {
                 // Process events
                 GlutinEvent::RedrawEventsCleared => {
                     *control_flow = ControlFlow::Wait;
-
                     if event_queue.is_empty() {
                         return;
                     }
@@ -416,6 +417,16 @@ impl<N: Notify + OnResize> Processor<N> {
 
             if terminal.dirty {
                 terminal.dirty = false;
+                let draw_text_objects = Instant::now() >= next_txob_draw;
+                if draw_text_objects {
+                    next_txob_draw = Instant::now() + txob_refresh_freq;
+                } else {
+                    // Don't let it wayt forever if we want to trigger draw
+                    if *control_flow == ControlFlow::Wait {
+                        *control_flow = ControlFlow::WaitUntil(next_txob_draw);
+                    }
+                    event_queue.push(GlutinEvent::UserEvent(Event::Wakeup));
+                }
 
                 // Request immediate re-draw if visual bell animation is not finished yet
                 if !terminal.visual_bell.completed() {
@@ -429,6 +440,7 @@ impl<N: Notify + OnResize> Processor<N> {
                     &self.config,
                     &self.mouse,
                     self.modifiers,
+                    draw_text_objects,
                 );
             }
         });
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -391,8 +391,13 @@ impl<N: Notify + OnResize> Processor<N> {
                 font_size: &mut self.font_size,
                 config: &mut self.config,
             };
-            let mut processor =
-                input::Processor::new(context, &self.display.urls, &self.display.highlighted_url);
+            let mut processor = input::Processor::new(
+                context,
+                &self.display.urls,
+                &self.display.text_objects,
+                &self.display.highlighted_url,
+                &self.display.highlighted_txob,
+            );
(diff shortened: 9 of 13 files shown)


rodolf0:
Seems this issue is related: https://github.com/jwilm/alacritty/issues/2536

chrisduerr:
It seems to me like much of this code is similar to what we originally did with our url search, where everything is collected into a buffer and then that is searched. As far as I remember, this was not ideal for performance though, which of course is at a higher priority for Alacritty.

I searched for a streaming regex parser back then, but after talking to burntsushi, that's not something that is easily built at performance comparable to the regex library, which is why I've written my own streaming parser specifically for URLs, which had the additional benefit of some heuristics that improve URL detection (like only including matching parenthesis).

However, I've gotten a new CPU very recently and I'm having some problems getting consistent benchmarking results out of it at the moment, so I was not yet able to fully verify if the performance issue we've encountered with this approach before does still exist. Could you maybe give it a shot @kchibisov if you have some time?

rodolf0:
> It seems to me like much of this code is similar to what we originally did with our url search, where everything is collected into a buffer and then that is searched. As far as I remember, this was not ideal for performance though, which of course is at a higher priority for Alacritty.
> 
> I searched for a streaming regex parser back then, but after talking to burntsushi, that's not something that is easily built at performance comparable to the regex library, which is why I've written my own streaming parser specifically for URLs, which had the additional benefit of some heuristics that improve URL detection (like only including matching parenthesis).

Trying to follow the pattern of `RenderableCellsIter` I also found that streaming support for regex doesn't exist and seems to be quite difficult and a bit of an edge case to implement. Considering we'd have to go over the whole displayable text anyway to find the text-objects currently present on each `draw` call, it felt that the tradeoff is just the memory needed to hold a copy of the display text (probably minimal footprint compared to whole rss). The CPU cost would be there either way if using an iterator or buffering the whole thing into a Vec?

Also note that if there's no text-objects configured you shouldn't pay much cost since I don't try to get a copy of the display.


> However, I've gotten a new CPU very recently and I'm having some problems getting consistent benchmarking results out of it at the moment, so I was not yet able to fully verify if the performance issue we've encountered with this approach before does still exist. Could you maybe give it a shot @kchibisov if you have some time?

I also tried enabling the `debug` functionality but couldn't spot any latency changes while rendering with 10 different text-objects configs.

kchibisov:
>I also tried enabling the debug functionality but couldn't spot any latency changes while rendering with 10 different text-objects configs.

This is not how you benchmark things ;- ). You should use vtebench for such things.

>However, I've gotten a new CPU very recently and I'm having some problems getting consistent benchmarking results out of it at the moment, so I was not yet able to fully verify if the performance issue we've encountered with this approach before does still exist. Could you maybe give it a shot @kchibisov if you have some time?

:- ) . I'll take a look once I go back to my desktop PC, so it'll take some time (2-3 days, idk).

rodolf0:
> > I also tried enabling the debug functionality but couldn't spot any latency changes while rendering with 10 different text-objects configs.
> 
> This is not how you benchmark things ;- ). You should use vtebench for such things.

That's fair, would something like this work?

Generate 100mb output using vtebench with same terminal display size

- text-object branch: https://gist.github.com/rodolf0/43b91b4a8fcd5c8820323a630b005298
- master branch: https://gist.github.com/rodolf0/5f9ec56e5cfcecc2eb726ac6fae811de

kchibisov:
It'll work, but you should also disable some cpu features and kernel ones to make it run on fixed frequency, otherwise it doesn't make sense. Your benchmarking results are not that stable as you can see. It also makes sense to show avarage when reporting things, since manually computing 10 numbers is not a good idea for review.

chrisduerr:
> Considering we'd have to go over the whole displayable text anyway to find the text-objects currently present on each draw call, it felt that the tradeoff is just the memory needed to hold a copy of the display text (probably minimal footprint compared to whole rss).

That's not actually true, since we iterate over the entire grid anyways for rendering. So when we do it again, it's done twice. As you can see with the current URL implementation, it uses the exact same loop that is also used for rendering each cell.

Especially if we have underlines, strikeout, URLs, regexes and possibly other things in the future, it obviously makes sense to integrate as much of those as possible into a single iteration of the grid cells.

> Also note that if there's no text-objects configured you shouldn't pay much cost since I don't try to get a copy of the display.

Alacritty is always fast, there shouldn't be any config options that have a big hidden cost associated to it. Saying Alacritty is a fast terminal emulator, but only if you choose some specific options, makes all of the effort a bit worthless. So unfortunately it's not an option to add things behind configuration options that cause significant performance differences.

> That's fair, would something like this work?

This is actually a great question. I've created https://github.com/jwilm/alacritty/issues/3136 to better document how exactly Alacritty should be benchmarked. I'll try to get on that later today or tomorrow, then you can give things a shot yourself if you've got some time.

rodolf0:
I followed the guidelines in https://www.llvm.org/docs/Benchmarking.html disabling address space randomization, setting performance governor, disabling turbo, except I didn't do the `cset` stuff. I was running just alacritty on my desktop (ie: no browser or other programs).

I ran a 100mb cat on both builds with 10 runs using `perf` for measurement.

```
##### Text-Objects branch
$ perf stat -r 10 cat /tmp/100mb.vte

 Performance counter stats for 'cat /tmp/100mb.vte' (10 runs):

            698.54 msec task-clock:u              #    0.120 CPUs utilized            ( +-  0.06% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.082 K/sec
         3,346,987      cycles:u                  #    0.005 GHz                      ( +-  0.57% )
           367,442      instructions:u            #    0.11  insn per cycle           ( +-  0.00% )
            85,964      branches:u                #    0.123 M/sec                    ( +-  0.00% )
            10,814      branch-misses:u           #   12.58% of all branches          ( +-  0.29% )

           5.82239 +- 0.00718 seconds time elapsed  ( +-  0.12% )
```

```
##### Master branch
$ perf stat -r 10 cat /tmp/100mb.vte

 Performance counter stats for 'cat /tmp/100mb.vte' (10 runs):

            694.38 msec task-clock:u              #    0.157 CPUs utilized            ( +-  0.05% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.082 K/sec
         2,772,748      cycles:u                  #    0.004 GHz                      ( +-  0.73% )
           367,748      instructions:u            #    0.13  insn per cycle           ( +-  0.00% )
            86,028      branches:u                #    0.124 M/sec                    ( +-  0.00% )
             9,759      branch-misses:u           #   11.34% of all branches          ( +-  0.64% )

            4.4246 +- 0.0158 seconds time elapsed  ( +-  0.36% )
```

kchibisov:
@rodolf0 Yeah these results make more sense now. We're not accepting any performance loss in general in alacritty, so to make such things upstream the difference should be in a margin of error. But on your PR it's ~30% which is terrible, unfortunately.

rodolf0:
Most of the extra work seems to come from the amount of text-objects we're searching for, not the copying of the display.

I'll see if using RegexSet removes some of that overhead.

If I run the experiments with different amount of text-objects configured I see the incremental costs, eg: for 1, 3, 11 compared to master branch. For low amount of text objects it doesn't seem like the cost changes that much. Maybe additional measurement from someone else is valuable to validate.

```
######## Text-Object branch - #text-objects = 1
$ perf stat -r 10 cat /tmp/100mb.vte

 Performance counter stats for 'cat /tmp/100mb.vte' (10 runs):

            694.77 msec task-clock:u              #    0.160 CPUs utilized            ( +-  0.06% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.082 K/sec
         2,872,289      cycles:u                  #    0.004 GHz                      ( +-  1.35% )
           366,874      instructions:u            #    0.13  insn per cycle           ( +-  0.00% )
            85,793      branches:u                #    0.123 M/sec                    ( +-  0.00% )
             9,448      branch-misses:u           #   11.01% of all branches          ( +-  0.41% )

            4.3530 +- 0.0366 seconds time elapsed  ( +-  0.84% )
```

```
######## Text-Object branch - #text-objects = 3
$ perf stat -r 10 cat /tmp/100mb.vte

 Performance counter stats for 'cat /tmp/100mb.vte' (10 runs):

            695.83 msec task-clock:u              #    0.161 CPUs utilized            ( +-  0.06% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.082 K/sec
         2,831,868      cycles:u                  #    0.004 GHz                      ( +-  0.81% )
           366,873      instructions:u            #    0.13  insn per cycle           ( +-  0.00% )
            85,792      branches:u                #    0.123 M/sec                    ( +-  0.00% )
             9,856      branch-misses:u           #   11.49% of all branches          ( +-  0.54% )

            4.3186 +- 0.0196 seconds time elapsed  ( +-  0.45% )
```

```
######## Text-Object branch - #text-objects = 11
$ perf stat -r 10 cat /tmp/100mb.vte

 Performance counter stats for 'cat /tmp/100mb.vte' (10 runs):

            699.88 msec task-clock:u              #    0.144 CPUs utilized            ( +-  0.07% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.081 K/sec
         2,972,845      cycles:u                  #    0.004 GHz                      ( +-  0.75% )
           366,874      instructions:u            #    0.12  insn per cycle           ( +-  0.00% )
            85,793      branches:u                #    0.123 M/sec                    ( +-  0.00% )
            10,285      branch-misses:u           #   11.99% of all branches          ( +-  0.48% )

            4.8544 +- 0.0370 seconds time elapsed  ( +-  0.76% )
```

```
########## Master Branch - no text-objects
$ perf stat -r 10 cat /tmp/100mb.vte

 Performance counter stats for 'cat /tmp/100mb.vte' (10 runs):

            693.66 msec task-clock:u              #    0.161 CPUs utilized            ( +-  0.06% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.082 K/sec
         2,758,622      cycles:u                  #    0.004 GHz                      ( +-  0.87% )
           367,753      instructions:u            #    0.13  insn per cycle           ( +-  0.00% )
            86,028      branches:u                #    0.124 M/sec                    ( +-  0.00% )
             8,940      branch-misses:u           #   10.39% of all branches          ( +-  0.37% )

            4.3189 +- 0.0202 seconds time elapsed  ( +-  0.47% )
```

kchibisov:
@rodolf0 Could you benchmark `scrolling` from vtebench?

rodolf0:
> @rodolf0 Could you benchmark `scrolling` from vtebench?

Had to try these runs with smaller unput, 100k `vtebench -w $(tput cols) -h $(tput lines) -b 102400 scrolling> /tmp/100k-scrolling.vte`

```
############## Master branch
$ perf stat -r 10 cat /tmp/100k-scrolling.vte
 Performance counter stats for 'cat /tmp/100k-scrolling.vte' (10 runs):

            370.82 msec task-clock:u              #    0.284 CPUs utilized            ( +-  0.15% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.154 K/sec
           386,722      cycles:u                  #    0.001 GHz                      ( +-  1.05% )
           275,260      instructions:u            #    0.71  insn per cycle           ( +-  0.00% )
            63,750      branches:u                #    0.172 M/sec                    ( +-  0.00% )
             4,455      branch-misses:u           #    6.99% of all branches          ( +-  0.61% )

            1.3079 +- 0.0266 seconds time elapsed  ( +-  2.03% )
```

```
########### Text-Objects branch, #text-objects = 11
$ perf stat -r 10 cat /tmp/100k-scrolling.vte
 Performance counter stats for 'cat /tmp/100k-scrolling.vte' (10 runs):

            374.11 msec task-clock:u              #    0.233 CPUs utilized            ( +-  0.15% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.152 K/sec
           397,095      cycles:u                  #    0.001 GHz                      ( +-  1.98% )
           275,251      instructions:u            #    0.69  insn per cycle           ( +-  0.00% )
            63,750      branches:u                #    0.170 M/sec                    ( +-  0.00% )
             4,465      branch-misses:u           #    7.00% of all branches          ( +-  0.67% )

           1.60757 +- 0.00286 seconds time elapsed  ( +-  0.18% )
```

```
######## Text-Objects, with single text-object configured
 Performance counter stats for 'cat /tmp/100k-scrolling.vte' (10 runs):

            374.39 msec task-clock:u              #    0.232 CPUs utilized            ( +-  0.09% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.152 K/sec
           383,474      cycles:u                  #    0.001 GHz                      ( +-  0.43% )
           275,251      instructions:u            #    0.72  insn per cycle           ( +-  0.00% )
            63,750      branches:u                #    0.170 M/sec                    ( +-  0.00% )
             4,452      branch-misses:u           #    6.98% of all branches          ( +-  0.47% )

           1.61095 +- 0.00386 seconds time elapsed  ( +-  0.24% )
```

kchibisov:
@rodolf0 Could you set your history to 100k lines in your config? the smaller numbers for this benchmark is fine. As you can see now you're stressing the right things, I guess. Since in alt-screen you grid isn't that big.

chrisduerr:
Thanks for all the benchmarks @rodolf0, that procedure is pretty much exactly what we're usually doing, though I might have to look into `cset` for my new CPU.

As expected, the differences are quite significant and even with a single textobject configured it seems like there's a difference big enough to block this PR, assuming that the benchmarks are consistent (running each more than once and mixing the runs with the different revisions is probably a good idea to verify consistency, it's what I've usually done in the past).

I'm surprised the slow part is the regex search though, since regex is extremely well optimized and it seems like collecting a big string would be a more difficult operation. But maybe that's a sign that there are some optimizations that can help with this. Have you checked for the origin of these differences by collecting performance statistics with `perf record`? If so, that's usually a great way to figure these bottlenecks out.

It looks like you've already got an idea for improvements though, so please report back on how that's going and let us know if you need any help.

rodolf0:
Results of `perf record -ag -- cat /tmp/100mb.vte

```
############ Text-Objects

Samples: 18K of event 'cycles', Event count (approx.): 8418431228
  Children      Self  Command          Shared Object                    Symbol
+   21.65%    20.26%  pty reader       alacritty                        [.] <alacritty_terminal::term::Term<T> as alacritty_terminal::ansi::Handler>::input                                                    ◆
+   14.00%     0.00%  cat              [kernel.kallsyms]                [k] entry_SYSCALL_64_after_hwframe                                                                                                     ▒
+   14.00%     0.02%  cat              [kernel.kallsyms]                [k] do_syscall_64                                                                                                                      ▒
+   13.53%    11.75%  pty reader       alacritty                        [.] vte::Parser::perform_action                                                                                                        ▒
+   13.50%     0.00%  cat              libc-2.30.so                     [.] __GI___libc_write                                                                                                                  ▒
+   13.41%     0.00%  cat              [kernel.kallsyms]                [k] ksys_write                                                                                                                         ▒
+   13.41%     0.00%  cat              [kernel.kallsyms]                [k] vfs_write                                                                                                                          ▒
+   13.40%     0.04%  cat              [kernel.kallsyms]                [k] tty_write                                                                                                                          ▒
+   13.23%    13.13%  pty reader       alacritty                        [.] std::sys_common::backtrace::__rust_begin_short_backtrace                                                                           ▒
+   12.90%    11.00%  cat              [kernel.kallsyms]                [k] n_tty_write                                                                                                                        ▒
+    5.22%     5.09%  alacritty        alacritty                        [.] regex::backtrack::Bounded<I>::backtrack                                                                                            ▒
+    3.51%     3.41%  alacritty        alacritty                        [.] alacritty::display::Display::draw                                                                                                  ▒
+    3.31%     0.00%  gnome-shell      [unknown]                        [.] 0000000000000000                                                                                                                   ▒
+    2.89%     0.04%  pty reader       libpthread-2.30.so               [.] __libc_read                                                                                                                        ▒
+    2.84%     0.00%  pty reader       [unknown]                        [.] 0x0000555556844140                                                                                                                 ▒
+    2.63%     0.00%  pty reader       [kernel.kallsyms]                [k] entry_SYSCALL_64_after_hwframe                                                                                                     ▒
+    2.63%     0.31%  pty reader       [kernel.kallsyms]                [k] do_syscall_64                                                                                                                      ▒
+    2.21%     0.00%  alacritty        [kernel.kallsyms]                [k] entry_SYSCALL_64_after_hwframe                                                                                                     ▒
+    2.21%     0.07%  alacritty        [kernel.kallsyms]                [k] do_syscall_64                                                                                                                      ▒
+    1.99%     0.00%  alacritty        libc-2.30.so                     [.] __GI___ioctl                                                                                                                       ▒
+    1.94%     1.89%  alacritty        alacritty                        [.] alacritty_terminal::term::RenderableCell::new                                                                                      ▒
+    1.89%     0.00%  alacritty        [kernel.kallsyms]                [k] __x64_sys_ioctl                                                                                                                    ▒
+    1.89%     0.00%  alacritty        [kernel.kallsyms]                [k] ksys_ioctl                                                                                                                         ▒
+    1.87%     0.00%  alacritty        [kernel.kallsyms]                [k] do_vfs_ioctl                                                                                                                       ▒
+    1.86%     0.00%  alacritty        [kernel.kallsyms]                [k] cnl_dump_hw_state                                                                                                                  ▒
+    1.85%     0.00%  alacritty        [kernel.kallsyms]                [k] ieee80211_change_mac                                                                                                               ▒
+    1.84%     0.01%  pty reader       [kernel.kallsyms]                [k] ksys_read                                                                                                                          ▒
+    1.75%     0.01%  pty reader       [kernel.kallsyms]                [k] vfs_read                                                                                                                           ▒
+    1.72%     0.01%  gnome-shell      [kernel.kallsyms]                [k] entry_SYSCALL_64_after_hwframe                
```


```
########### Master

Samples: 16K of event 'cycles', Event count (approx.): 7754805202
  Children      Self  Command          Shared Object                    Symbol
+   27.66%    26.10%  pty reader       alacritty                        [.] <alacritty_terminal::term::Term<T> as alacritty_terminal::ansi::Handler>::input                                                        ◆
+   15.05%     0.00%  cat              [kernel.kallsyms]                [k] entry_SYSCALL_64_after_hwframe                                                                                                         ▒
+   15.05%     0.01%  cat              [kernel.kallsyms]                [k] do_syscall_64                                                                                                                          ▒
+   14.53%     0.00%  cat              libc-2.30.so                     [.] __GI___libc_write                                                                                                                      ▒
+   14.51%    14.24%  pty reader       alacritty                        [.] std::sys_common::backtrace::__rust_begin_short_backtrace                                                                               ▒
+   14.40%     0.00%  cat              [kernel.kallsyms]                [k] ksys_write                                                                                                                             ▒
+   14.40%     0.01%  cat              [kernel.kallsyms]                [k] vfs_write                                                                                                                              ▒
+   14.38%     0.04%  cat              [kernel.kallsyms]                [k] tty_write                                                                                                                              ▒
+   13.99%    11.98%  cat              [kernel.kallsyms]                [k] n_tty_write                                                                                                                            ▒
+   13.91%    12.13%  pty reader       alacritty                        [.] vte::Parser::perform_action                                                                                                            ▒
+    3.17%     0.00%  gnome-shell      [unknown]                        [.] 0000000000000000                                                                                                                       ▒
+    2.92%     0.04%  pty reader       libpthread-2.30.so               [.] __libc_read                                                                                                                            ▒
+    2.72%     0.01%  pty reader       [kernel.kallsyms]                [k] entry_SYSCALL_64_after_hwframe                                                                                                         ▒
+    2.71%     0.29%  pty reader       [kernel.kallsyms]                [k] do_syscall_64                                                                                                                          ▒
+    2.45%     2.34%  alacritty        alacritty                        [.] alacritty::display::Display::draw                                                                                                      ▒
+    1.88%     0.01%  gnome-shell      [kernel.kallsyms]                [k] entry_SYSCALL_64_after_hwframe                                                                                                         ▒
+    1.87%     0.14%  gnome-shell      [kernel.kallsyms]                [k] do_syscall_64                                                                                                                          ▒
+    1.86%     0.01%  pty reader       [kernel.kallsyms]                [k] ksys_read                                                                                                                              ▒
+    1.77%     1.72%  alacritty        alacritty                        [.] alacritty_terminal::term::RenderableCell::new                                                                                          ▒
+    1.75%     0.01%  pty reader       [kernel.kallsyms]                [k] vfs_read                                                                                                                               ▒
+    1.75%     0.01%  alacritty        [kernel.kallsyms]                [k] entry_SYSCALL_64_after_hwframe    
```

rodolf0:
After using a RegexSet for a single scan on the display for all text-objects and only doing further scans if RegexSet indicates there's a match, seems costs are better: https://gist.github.com/rodolf0/3a71b539a087309e9aa6153c944d67ee

For these checks I followed the same setup as before, disabling turbo, setting performance governor, selecting the online CPUs, etc.


An additional round of perf:
```
############# Text-Objects Branch
 Performance counter stats for 'cat /tmp/100k-scrolling.vte' (10 runs):

            370.26 msec task-clock:u              #    0.284 CPUs utilized            ( +-  0.13% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.154 K/sec
           388,928      cycles:u                  #    0.001 GHz                      ( +-  1.30% )
           274,079      instructions:u            #    0.70  insn per cycle           ( +-  0.00% )
            63,393      branches:u                #    0.171 M/sec                    ( +-  0.00% )
             4,388      branch-misses:u           #    6.92% of all branches          ( +-  0.37% )

            1.3035 +- 0.0178 seconds time elapsed  ( +-  1.36% )

00:38:03 ~/tmp/alacritty 0 $ perf stat -r 10 cat /tmp/100mb.vte

 Performance counter stats for 'cat /tmp/100mb.vte' (10 runs):

            691.93 msec task-clock:u              #    0.160 CPUs utilized            ( +-  0.09% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.082 K/sec
         2,685,062      cycles:u                  #    0.004 GHz                      ( +-  0.70% )
           366,882      instructions:u            #    0.14  insn per cycle           ( +-  0.00% )
            85,796      branches:u                #    0.124 M/sec                    ( +-  0.00% )
             8,971      branch-misses:u           #   10.46% of all branches          ( +-  0.74% )

            4.3258 +- 0.0201 seconds time elapsed  ( +-  0.46% )
```


```
############ Master Branch
 Performance counter stats for 'cat /tmp/100k-scrolling.vte' (10 runs):

            369.37 msec task-clock:u              #    0.293 CPUs utilized            ( +-  0.10% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.154 K/sec
           390,728      cycles:u                  #    0.001 GHz                      ( +-  1.48% )
           274,954      instructions:u            #    0.70  insn per cycle           ( +-  0.00% )
            63,626      branches:u                #    0.172 M/sec                    ( +-  0.00% )
             4,420      branch-misses:u           #    6.95% of all branches          ( +-  0.39% )

           1.26209 +- 0.00769 seconds time elapsed  ( +-  0.61% )

00:38:24 ~ 0 $ perf stat -r 10 cat /tmp/100mb.vte

 Performance counter stats for 'cat /tmp/100mb.vte' (10 runs):

            691.64 msec task-clock:u              #    0.171 CPUs utilized            ( +-  0.06% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
                57      page-faults:u             #    0.082 K/sec
         2,400,034      cycles:u                  #    0.003 GHz                      ( +-  1.35% )
           367,753      instructions:u            #    0.15  insn per cycle           ( +-  0.00% )
            86,028      branches:u                #    0.124 M/sec                    ( +-  0.00% )
             8,515      branch-misses:u           #    9.90% of all branches          ( +-  0.71% )

            4.0506 +- 0.0427 seconds time elapsed  ( +-  1.05% )
```

chrisduerr:
That's still like a 3-7% performance hit, which is a lot for a feature like this.

nixpulvis:
When text objects are used we still separately scan for URLs. Would using the more general regex engine for URLs when already needed for text objects save us anything?

If so, we could choose to use the faster (?) urlocator iff we don't have any text objects configured.

kchibisov:
@nixpulvis this is not about choose to be faster in some cases, you should be fast in general. URL searching doesn't affect performance at all (well it was proven that its performance is identical to master before merge IIRC). So if user is using text objects it should expect things to run as fast as without them.

chrisduerr:
> When text objects are used we still separately scan for URLs. Would using the more general regex engine for URLs when already needed for text objects save us anything?

It shouldn't make any difference in performance, since I've made sure that the impact on performance with URLs is basically nil.

rodolf0:
I don't think this is the correct implementation (https://github.com/jwilm/alacritty/pull/3126/commits/c6d9d5e6e773754d67615f4e53503a293e57c913) but would something along these lines work?

Intention: scan display for text-objects + render those when display is dirty but not more frequently than 50ms or some other interval.

Since text-objects are meant exclusively for human interaction, it'd not make any sense to try rendering them for other purposes.

chrisduerr:
Skipping rendering of certain things to try and pretend to be faster is what a slow terminal would do. Alacritty doesn't want to be a slow terminal though, we're supposed to be fast. So we should be able to do everything we need to do at 30-60 FPS consistently.

This is exactly why the url implementation exists in the way it does right now and why I've mentioned previously that it will be tough to implement this without a streaming regex engine.

rodolf0:
Seems that currently this feature can't be added without taking some perf hit. Since there's no regression budget for such feature I'll just close this one. Thanks for time invested in reviewing this!

cole-h:
Thank you for spending your time on this. It's unfortunate that it couldn't be merged, but maybe this will inspire somebody (or even you!) to create an amazing streaming regex engine, which might enable this in the future :wink:

shibumi:
@chrisduerr  @rodolf0  just an idea: Why do you not make this optional? What I have in my mind is: I press a certain key combination and only then the regex search will start over the current "buffer" or output alacritty is displaying... this way there would be no performance hit.
The feature would be complete optional and only on-demand. Performance doesn't matter in this case as long the regex results will occur in less than 1s.

chrisduerr:
Nothing in Alacritty should ever be slow. Optional or not. If there's a way to opt out of being fast, Alacritty is de-facto slow.

rodolf0:
> @chrisduerr  @rodolf0  just an idea: Why do you not make this optional? What I have in my mind is: I press a certain key combination and only then the regex search will start over the current "buffer" or output alacritty is displaying... this way there would be no performance hit.
> The feature would be complete optional and only on-demand. Performance doesn't matter in this case as long the regex results will occur in less than 1s.

I like this idea. Not paying the cost until a modifier key (for example ctrl) is held down. Only then do the regex approach while holding. The moment you let go of the modifier, you don't pay the cost again.

At the end of the day I guess the feature doesn't have enough demand to justify it :-( I've implemented something similar in gnome terminal of you're interested (https://gitlab.gnome.org/GNOME/gnome-terminal/-/issues/43), but getting it merged upstream has had it's difficulties there too.

chrisduerr:
There currently is regex search implemented in Alacritty directly, however that will **never** block anything. It's just completely asynchronously and immediately gets pushed in the background as soon as an answer can't be returned immediately. But this requires operating on the dynamic buffer, instead of just cloning everything.

By using these systems, which are already in place, one could certainly implement this without too much trouble. It's mostly a question of configuration really.

shibumi:
> Nothing in Alacritty should ever be slow. Optional or not. If there's a way to opt out of being fast, Alacritty is de-facto slow.

I don't think this would cost so much performance. The difference to the old PR is significant, because it would only work on the current buffer that alacritty is displaying.. "PageUp/PageDown" for scrolling would be disabled. 

I don't know the internals nor am I a rust developer, but I am pretty confident this should be doable with good performance.

Issue #3082, opened 2019-12-09
Windows emoji support


chrisduerr:
This is a follow-up to https://github.com/jwilm/alacritty/issues/153.

After Emojis are now implemented on macOS, Linux and BSD, the only platform lacking Emoji support is Windows.

kchibisov:
This issue should be pretty simple btw. So if anyone would like to try implementing it, I could provide with all the required information, at least, I think that its all you should need.

P.s. I just dont have Windows machine to try implementing it myself.

davidhewitt:
@kchibisov if you want to drop the information you have on this ticket, I could have a go at picking this up sometime.

kchibisov:
The things I have for Windows emoji.

Everything should be changed only in the following [file](https://github.com/jwilm/alacritty/blob/master/font/src/directwrite/mod.rs) (folder). For colored emojis you should send `RGBA` buffer to Alacritty here, so you should use `RGB` buffer for all normal glyphs, and for colored once you should use `RGBA` [here](https://gist.github.com/makotokato/9407295).

This code that I found could be [useful](https://gist.github.com/makotokato/9407295) , but I'm not sure about it. Also the official [docs](https://docs.microsoft.com/en-us/windows/win32/directwrite/color-fonts).

If you found out that you should downsample things yourself we already have an algorithm around in https://github.com/jwilm/alacritty/blob/8a833406614b503dcd647015d3cfd239faf5cb0c/font/src/ft/mod.rs#L588.

You shouldn't change any code outside of the font crate in the end.

Also you should be aware of this bug #1864, so it's fine that your emojis would also suffer from this issue, since it's a bug in our renderer. I mean with `#ffffff` foreground and `#000000` background color the color of your emojis should be correct, if it's not correct with the provided colors then it's likely your PR fault. 

If you have more questions ask me(us) here or on IRC.

davidhewitt:
Now that #3127 is merged (:tada:) we have greyscale emoji rendering!

Next is to add color rendering using the notes above from kchibisov. I had originally hoped to do the whole thing in one PR but chose to land PR #3127 on its own for two reasons:

- That in itself was a complicated enough change to warrant enough review and feedback for one PR
- I couldn't find the suggested APIs for color rendering in drwite (and maybe not even in winapi) - so some definitions might want to be added upstream for the color rendering to be added.

My time is split too many ways right now for me to champion a large-ish story like the above, though I would hope to return to it in the future.

[...]

Issue #957, opened 2017-12-23
RFC: Font Configuration


jwilm:
I've been thinking about many of the current limitations of how Alacritty's font configuration and possible solutions. This is an RFC for a fully capable font configuration section.

Feedback is very much invited! I may have missed out on some important requirement, and I may have missed out on an obviously better solution. Please feel welcome to comment and suggest improvements or add to the list of use cases this should handle.

## Font Configuration, today

Current font configuration is achieved through the following block of YAML. Most of the comments are preserved for clarity.

```yaml
font:
  # The normal (roman) font face to use.
  normal:
    family: monospace
    style: Regular

  # The bold font face
  bold:
    family: monospace
    style: Bold

  # The italic font face
  italic:
    family: monospace
    style: Italic

  # Point size of the font
  size: 11.0

  # Offset is the extra space around each character. offset.y can be thought of
  # as modifying the linespacing, and offset.x as modifying the letter spacing.
  offset:
    x: 0.0
    y: 0.0

  # Glyph offset determines the locations of the glyphs within their cells with
  # the default being at the bottom. Increase the x offset to move the glyph to
  # the right, increase the y offset to move the glyph upward.
  glyph_offset:
    x: 0.0
    y: 0.0

  # OS X only: use thin stroke font rendering. Thin strokes are suitable
  # for retina displays, but for non-retina you probably want this set to
  # false.
  use_thin_strokes: true
```

### Limitations in the current system

- Size is set globally. This is a problem because it may be desirable to have different sizes for latin text vs CJK or others.
- Glyph offsets are set globally. Setting this per face would be beneficial for things like powerline which often has alignment problems.
- There's no way to specify fallbacks. It may be desirable for users to provide their own fallback list.
- Several properties are not configurable because we don't yet offer them. These are things like subpixel AA, lcd filter, and antialiasing. In theory, these can all be configured per-font, so it might make sense to include them when thinking about a solution for size.

### Other issues

The `font.offset` block is confusingly named. Its name sounds like it's doing what the `glyph_offset` config does, but in reality it is changing cell dimensions.

A better naming system for `font.offset` might be `font.spacing` and give it properties of `line` and `letter` to match convention.

### Considerations for next system

In addition to solving the limitations outlined above, the next system should interact well with system level font settings such as `fontconfig` on Linux platforms.

### Common Font Options

[...]

PR #3127, opened 2019-12-27
Add windows font fallbacks

### LINKED ISSUE #3215: Support Windows font fallback
OS: Windows
Version: 0.4.1 (4d9253b)
Windows: ConPTY

When using the recently released JetBrains Mono font (https://www.jetbrains.com/lp/mono/) on alacritty, seeing a strange issue where some of the UTF-8 characters are not displayed at all with alacritty.   Other terminals (mintty for example) are able to display this correctly.  First noticed when tmux did not display the borders of the panes.

Tested it with this demo test file: https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-demo.txt
Below is a cat of the file in alacritty - notice how only some of the characters are getting printed, and none of the box drawing characters are printed.
alacritty

Same in mintty for comparison:
mintty

Curious to know if it's just me or if others see this as well.
Thank you.

davidhewitt:
This is the first step for #3082

At the moment Alacritty on windows does not use any font fallbacks, so glyphs for emoji cannot be found. I suspect there are many other cases where lack of font fallbacks causes glyphs to be missing.

By adding font fallbacks, we now can rasterize (downsampled) greyscale emoji. They don't look amazing, but hey, something is better than nothing!

The remaining step to add full emoji support is to implement color font rasterizing, which will be a much bigger change and might need upstream support in the `dwrote` crate.

Perhaps others may have ideas how to improve the render quality of these greyscale emojis too.

If this PR looks like the right direction to be heading, there's a couple of things for me to finish off:
- [x] Remove hard-coding of "en-us" locale
- [x] Try to find fallback font in loaded fonts rather than always creating fresh

Possible bugs from testing:
- [x] ~~Figure out how to size emoji correctly - downsample?~~
- [x] ~~Investigate emoji paste issues?~~ (Upstream bug https://github.com/microsoft/terminal/issues/1503)
- [x] ~~Investigate occasional missing glyph~~ (Sounds a lot like upstream bug https://github.com/microsoft/terminal/issues/386)

Before:

image

After:

image

Fixes #3215.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Vi mode for copying text and opening links
 - `CopySelection` action which copies into selection buffer on Linux/BSD
 - Option `cursor.thickness` to set terminal cursor thickness
+- Font fallback on Windows
 
 ### Changed
 
--- a/font/Cargo.toml
+++ b/font/Cargo.toml
@@ -23,4 +23,5 @@ core-graphics = "0.17"
 core-foundation-sys = "0.6"
 
 [target.'cfg(windows)'.dependencies]
-dwrote = { version = "0.9.0" }
+dwrote = { version = "0.11" }
+winapi = { version = "0.3", features = ["impl-default"] }
--- a/font/src/directwrite/mod.rs
+++ b/font/src/directwrite/mod.rs
@@ -45,8 +45,8 @@ pub struct DirectWriteRasterizer {
     fonts: HashMap<FontKey, Font>,
     keys: HashMap<FontDesc, FontKey>,
     device_pixel_ratio: f32,
-    collection: FontCollection,
-    fallback: Option<FontFallback>,
+    available_fonts: FontCollection,
+    fallback_sequence: Option<FontFallback>,
 }
 
 impl DirectWriteRasterizer {
@@ -122,7 +122,7 @@ impl DirectWriteRasterizer {
     }
 
     fn get_fallback_font(&self, loaded_font: &Font, c: char) -> Option<dwrote::Font> {
-        let fallback = self.fallback.as_ref()?;
+        let fallback = self.fallback_sequence.as_ref()?;
 
         let mut buf = [0u16; 2];
         c.encode_utf16(&mut buf);
@@ -142,7 +142,7 @@ impl DirectWriteRasterizer {
             &text_analysis_source,
             0,
             length,
-            &self.collection,
+            &self.available_fonts,
             Some(&loaded_font.family_name),
             loaded_font.weight,
             loaded_font.style,
@@ -161,8 +161,8 @@ impl crate::Rasterize for DirectWriteRasterizer {
             fonts: HashMap::new(),
             keys: HashMap::new(),
             device_pixel_ratio,
-            collection: FontCollection::system(),
-            fallback: FontFallback::get_system_fallback(),
+            available_fonts: FontCollection::system(),
+            fallback_sequence: FontFallback::get_system_fallback(),
         })
     }
 
@@ -211,7 +211,7 @@ impl crate::Rasterize for DirectWriteRasterizer {
         }
 
         let family = self
-            .collection
+            .available_fonts
             .get_font_family_by_name(&desc.name)
             .ok_or_else(|| Error::MissingFont(desc.clone()))?;
 
--- a/font/src/directwrite/mod.rs
+++ b/font/src/directwrite/mod.rs
@@ -32,11 +32,7 @@ use super::{
     BitmapBuffer, FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph, Size, Slant, Style, Weight,
 };
 
-/// This struct contains the properties from dwrote::Font needed in the rest of this module.
-///
-/// We use this struct rather than dwrote::Font for performance reasons: the getter methods on
-/// dwrote::Font need to do things like String conversion from C++ widestrings to Rust's
-/// UTF-8 strings.
+/// Cached DirectWrite font.
 pub struct Font {
     face: FontFace,
     family_name: String,
@@ -139,7 +135,7 @@ impl DirectWriteRasterizer {
         let text_analysis_source_data = TextAnalysisSourceData { locale: &locale, length };
         let text_analysis_source = TextAnalysisSource::from_text(
             Box::new(text_analysis_source_data),
-            Cow::Borrowed(&utf16_codepoints),
+            Cow::Borrowed(utf16_codepoints),
         );
 
         let fallback_result = fallback.map_characters(
@@ -210,7 +206,7 @@ impl crate::Rasterize for DirectWriteRasterizer {
     }
 
     fn load_font(&mut self, desc: &FontDesc, _size: Size) -> Result<FontKey, Error> {
-        // Fast path if face already loaded
+        // Fast path if face is already loaded
         if let Some(key) = self.keys.get(desc) {
             return Ok(*key);
         }
@@ -257,12 +253,11 @@ impl crate::Rasterize for DirectWriteRasterizer {
         let loaded_font = self.get_loaded_font(glyph.font_key)?;
 
         match self.rasterize_glyph(&loaded_font.face, glyph.size, glyph.c) {
-            Ok(glyph) => Ok(glyph),
             Err(err @ Error::MissingGlyph(_)) => {
                 let fallback_font = self.get_fallback_font(&loaded_font, glyph.c).ok_or(err)?;
                 self.rasterize_glyph(&fallback_font.create_font_face(), glyph.size, glyph.c)
-            }
-            Err(err) => Err(err),
+            },
+            result => result,
         }
     }
 
@@ -338,12 +333,11 @@ fn get_current_locale() -> String {
     let mut buf = vec![0u16; LOCALE_NAME_MAX_LENGTH];
     let len = unsafe { GetUserDefaultLocaleName(buf.as_mut_ptr(), buf.len() as i32) as usize };
 
-    // len includes null byte, we don't need that in Rust strings
+    // `len` includes null byte, which we don't need in Rust
     OsString::from_wide(&buf[..len - 1]).into_string().expect("Locale not valid unicode")
 }
 
-/// Information needed by dwrote's TextAnalysisSource to find a fallback font
-/// for a run of text. See `get_fallback_font`.
+/// Font fallback information for dwrote's TextAnalysisSource.
 struct TextAnalysisSourceData<'a> {
     locale: &'a str,
     length: u32,
--- a/font/src/directwrite/mod.rs
+++ b/font/src/directwrite/mod.rs
@@ -13,30 +13,168 @@
 // limitations under the License.
 //
 //! Rasterization powered by DirectWrite
-use dwrote::{FontCollection, FontStretch, FontStyle, FontWeight, GlyphOffset, GlyphRunAnalysis};
+use std::borrow::Cow;
+use std::collections::HashMap;
+use std::ffi::OsString;
+use std::fmt::{self, Display, Formatter};
+use std::os::windows::ffi::OsStringExt;
+
+use dwrote::{
+    FontCollection, FontFace, FontFallback, FontStretch, FontStyle, FontWeight, GlyphOffset,
+    GlyphRunAnalysis, TextAnalysisSource, TextAnalysisSourceMethods, DWRITE_GLYPH_RUN,
+};
+
+use winapi::shared::ntdef::{HRESULT, LOCALE_NAME_MAX_LENGTH};
+use winapi::um::dwrite;
+use winapi::um::winnls::GetUserDefaultLocaleName;
 
 use super::{
     BitmapBuffer, FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph, Size, Slant, Style, Weight,
 };
 
+/// This struct contains the properties from dwrote::Font needed in the rest of this module.
+///
+/// We use this struct rather than dwrote::Font for performance reasons: the getter methods on
+/// dwrote::Font need to do things like String conversion from C++ widestrings to Rust's
+/// UTF-8 strings.
+pub struct Font {
+    face: FontFace,
+    family_name: String,
+    weight: FontWeight,
+    style: FontStyle,
+    stretch: FontStretch,
+}
+
 pub struct DirectWriteRasterizer {
-    fonts: Vec<dwrote::FontFace>,
+    fonts: HashMap<FontKey, Font>,
+    keys: HashMap<FontDesc, FontKey>,
     device_pixel_ratio: f32,
+    collection: FontCollection,
+    fallback: Option<FontFallback>,
+}
+
+impl DirectWriteRasterizer {
+    fn rasterize_glyph(
+        &self,
+        face: &FontFace,
+        size: Size,
+        c: char,
+    ) -> Result<RasterizedGlyph, Error> {
+        let glyph_index = self.get_glyph_index(face, c)?;
+
+        let em_size = em_size(size);
+
+        let glyph_run = DWRITE_GLYPH_RUN {
+            fontFace: unsafe { face.as_ptr() },
+            fontEmSize: em_size,
+            glyphCount: 1,
+            glyphIndices: &glyph_index,
+            glyphAdvances: &0.0,
+            glyphOffsets: &GlyphOffset::default(),
+            isSideways: 0,
+            bidiLevel: 0,
+        };
+
+        let rendering_mode = face.get_recommended_rendering_mode_default_params(
+            em_size,
+            self.device_pixel_ratio,
+            dwrote::DWRITE_MEASURING_MODE_NATURAL,
+        );
+
+        let glyph_analysis = GlyphRunAnalysis::create(
+            &glyph_run,
+            self.device_pixel_ratio,
+            None,
+            rendering_mode,
+            dwrote::DWRITE_MEASURING_MODE_NATURAL,
+            0.0,
+            0.0,
+        )
+        .map_err(Error::DirectWriteError)?;
+
+        let bounds = glyph_analysis
+            .get_alpha_texture_bounds(dwrote::DWRITE_TEXTURE_CLEARTYPE_3x1)
+            .map_err(Error::DirectWriteError)?;
+
+        let buf = glyph_analysis
+            .create_alpha_texture(dwrote::DWRITE_TEXTURE_CLEARTYPE_3x1, bounds)
+            .map_err(Error::DirectWriteError)?;
+
+        Ok(RasterizedGlyph {
+            c,
+            width: (bounds.right - bounds.left) as i32,
+            height: (bounds.bottom - bounds.top) as i32,
+            top: -bounds.top,
+            left: bounds.left,
+            buf: BitmapBuffer::RGB(buf),
+        })
+    }
+
+    fn get_loaded_font(&self, font_key: FontKey) -> Result<&Font, Error> {
+        self.fonts.get(&font_key).ok_or(Error::FontNotLoaded)
+    }
+
+    fn get_glyph_index(&self, face: &FontFace, c: char) -> Result<u16, Error> {
+        let idx = *face
+            .get_glyph_indices(&[c as u32])
+            .first()
+            // DirectWrite returns 0 if the glyph does not exist in the font
+            .filter(|glyph_index| **glyph_index != 0)
+            .ok_or_else(|| Error::MissingGlyph(c))?;
+
+        Ok(idx)
+    }
+
+    fn get_fallback_font(&self, loaded_font: &Font, c: char) -> Option<dwrote::Font> {
+        let fallback = self.fallback.as_ref()?;
+
+        let mut buf = [0u16; 2];
+        c.encode_utf16(&mut buf);
+
+        let length = c.len_utf16() as u32;
+        let utf16_codepoints = &buf[..length as usize];
+
+        let locale = get_current_locale();
+
+        let text_analysis_source_data = TextAnalysisSourceData { locale: &locale, length };
+        let text_analysis_source = TextAnalysisSource::from_text(
+            Box::new(text_analysis_source_data),
+            Cow::Borrowed(&utf16_codepoints),
+        );
+
+        let fallback_result = fallback.map_characters(
+            &text_analysis_source,
+            0,
+            length,
+            &self.collection,
+            Some(&loaded_font.family_name),
+            loaded_font.weight,
+            loaded_font.style,
+            loaded_font.stretch,
+        );
+
+        fallback_result.mapped_font
+    }
 }
 
 impl crate::Rasterize for DirectWriteRasterizer {
     type Err = Error;
 
     fn new(device_pixel_ratio: f32, _: bool) -> Result<DirectWriteRasterizer, Error> {
-        Ok(DirectWriteRasterizer { fonts: Vec::new(), device_pixel_ratio })
+        Ok(DirectWriteRasterizer {
+            fonts: HashMap::new(),
+            keys: HashMap::new(),
+            device_pixel_ratio,
+            collection: FontCollection::system(),
+            fallback: FontFallback::get_system_fallback(),
+        })
     }
 
     fn metrics(&self, key: FontKey, size: Size) -> Result<Metrics, Error> {
-        let font = self.fonts.get(key.token as usize).ok_or(Error::FontNotLoaded)?;
+        let face = &self.get_loaded_font(key)?.face;
+        let vmetrics = face.metrics().metrics0();
 
-        let vmetrics = font.metrics();
-        let scale = (size.as_f32_pts() * self.device_pixel_ratio * (96.0 / 72.0))
-            / f32::from(vmetrics.designUnitsPerEm);
+        let scale = em_size(size) * self.device_pixel_ratio / f32::from(vmetrics.designUnitsPerEm);
 
         let underline_position = f32::from(vmetrics.underlinePosition) * scale;
         let underline_thickness = f32::from(vmetrics.underlineThickness) * scale;
@@ -51,9 +189,12 @@ impl crate::Rasterize for DirectWriteRasterizer {
         let line_height = f64::from(ascent - descent + line_gap);
 
         // We assume that all monospace characters have the same width
-        // Because of this we take '!', the first drawable character, for measurements
-        let glyph_metrics = font.get_design_glyph_metrics(&[33], false);
-        let hmetrics = glyph_metrics.first().ok_or(Error::MissingGlyph('!'))?;
+        // Because of this we take '!', the first drawable character, for horizontal metrics
+        let c = '!';
+        let glyph_index = self.get_glyph_index(face, c)?;
+
+        let glyph_metrics = face.get_design_glyph_metrics(&[glyph_index], false);
+        let hmetrics = glyph_metrics.first().ok_or_else(|| Error::MissingGlyph(c))?;
 
         let average_advance = f64::from(hmetrics.advanceWidth) * f64::from(scale);
 
@@ -69,27 +210,22 @@ impl crate::Rasterize for DirectWriteRasterizer {
     }
 
     fn load_font(&mut self, desc: &FontDesc, _size: Size) -> Result<FontKey, Error> {
-        let system_fc = FontCollection::system();
+        // Fast path if face already loaded
+        if let Some(key) = self.keys.get(desc) {
+            return Ok(*key);
+        }
 
-        let family = system_fc
+        let family = self
+            .collection
             .get_font_family_by_name(&desc.name)
             .ok_or_else(|| Error::MissingFont(desc.clone()))?;
 
         let font = match desc.style {
             Style::Description { weight, slant } => {
-                let weight =
-                    if weight == Weight::Bold { FontWeight::Bold } else { FontWeight::Regular };
-
-                let style = match slant {
-                    Slant::Normal => FontStyle::Normal,
-                    Slant::Oblique => FontStyle::Oblique,
-                    Slant::Italic => FontStyle::Italic,
-                };
-
                 // This searches for the "best" font - should mean we don't have to worry about
                 // fallbacks if our exact desired weight/style isn't available
-                Ok(family.get_first_matching_font(weight, FontStretch::Normal, style))
-            },
+                Ok(family.get_first_matching_font(weight.into(), FontStretch::Normal, slant.into()))
+            }
             Style::Specific(ref style) => {
                 let mut idx = 0;
                 let count = family.get_font_count();
@@ -107,73 +243,27 @@ impl crate::Rasterize for DirectWriteRasterizer {
 
                     idx += 1;
                 }
-            },
+            }
         }?;
 
-        let face = font.create_font_face();
-        self.fonts.push(face);
+        let key = FontKey::next();
+        self.keys.insert(desc.clone(), key);
+        self.fonts.insert(key, font.into());
 
-        Ok(FontKey { token: (self.fonts.len() - 1) as u32 })
+        Ok(key)
     }
 
     fn get_glyph(&mut self, glyph: GlyphKey) -> Result<RasterizedGlyph, Error> {
-        let font = self.fonts.get(glyph.font_key.token as usize).ok_or(Error::FontNotLoaded)?;
+        let loaded_font = self.get_loaded_font(glyph.font_key)?;
 
-        let offset = GlyphOffset { advanceOffset: 0.0, ascenderOffset: 0.0 };
-
-        let glyph_index = *font
-            .get_glyph_indices(&[glyph.c as u32])
-            .first()
-            .ok_or_else(|| Error::MissingGlyph(glyph.c))?;
-        if glyph_index == 0 {
-            // The DirectWrite documentation states that we should get 0 returned if the glyph
-            // does not exist in the font
-            return Err(Error::MissingGlyph(glyph.c));
+        match self.rasterize_glyph(&loaded_font.face, glyph.size, glyph.c) {
+            Ok(glyph) => Ok(glyph),
+            Err(err @ Error::MissingGlyph(_)) => {
+                let fallback_font = self.get_fallback_font(&loaded_font, glyph.c).ok_or(err)?;
+                self.rasterize_glyph(&fallback_font.create_font_face(), glyph.size, glyph.c)
+            }
+            Err(err) => Err(err),
         }
-
-        let glyph_run = dwrote::DWRITE_GLYPH_RUN {
-            fontFace: unsafe { font.as_ptr() },
-            fontEmSize: glyph.size.as_f32_pts(),
-            glyphCount: 1,
-            glyphIndices: &(glyph_index),
-            glyphAdvances: &(0.0),
-            glyphOffsets: &(offset),
-            isSideways: 0,
-            bidiLevel: 0,
-        };
-
-        let rendering_mode = font.get_recommended_rendering_mode_default_params(
-            glyph.size.as_f32_pts(),
-            self.device_pixel_ratio * (96.0 / 72.0),
-            dwrote::DWRITE_MEASURING_MODE_NATURAL,
-        );
-
-        let glyph_analysis = GlyphRunAnalysis::create(
-            &glyph_run,
-            self.device_pixel_ratio * (96.0 / 72.0),
-            None,
-            rendering_mode,
-            dwrote::DWRITE_MEASURING_MODE_NATURAL,
-            0.0,
-            0.0,
-        )
-        .or_else(|_| Err(Error::MissingGlyph(glyph.c)))?;
-
-        let bounds = glyph_analysis
-            .get_alpha_texture_bounds(dwrote::DWRITE_TEXTURE_CLEARTYPE_3x1)
-            .or_else(|_| Err(Error::MissingGlyph(glyph.c)))?;
-        let buf = glyph_analysis
-            .create_alpha_texture(dwrote::DWRITE_TEXTURE_CLEARTYPE_3x1, bounds)
-            .or_else(|_| Err(Error::MissingGlyph(glyph.c)))?;
-
-        Ok(RasterizedGlyph {
-            c: glyph.c,
-            width: (bounds.right - bounds.left) as i32,
-            height: (bounds.bottom - bounds.top) as i32,
-            top: -bounds.top,
-            left: bounds.left,
-            buf: BitmapBuffer::RGB(buf),
-        })
     }
 
     fn update_dpr(&mut self, device_pixel_ratio: f32) {
@@ -186,29 +276,85 @@ pub enum Error {
     MissingFont(FontDesc),
     MissingGlyph(char),
     FontNotLoaded,
+    DirectWriteError(HRESULT),
 }
 
-impl ::std::error::Error for Error {
-    fn description(&self) -> &str {
-        match *self {
-            Error::MissingFont(ref _desc) => "Couldn't find the requested font",
-            Error::MissingGlyph(ref _c) => "Couldn't find the requested glyph",
-            Error::FontNotLoaded => "Tried to operate on font that hasn't been loaded",
-        }
-    }
-}
+impl std::error::Error for Error {}
 
-impl ::std::fmt::Display for Error {
-    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
-        match *self {
-            Error::MissingGlyph(ref c) => write!(f, "Glyph not found for char {:?}", c),
-            Error::MissingFont(ref desc) => write!(
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            Error::MissingGlyph(c) => write!(f, "Glyph not found for char {:?}", c),
+            Error::MissingFont(desc) => write!(
                 f,
                 "Couldn't find a font with {}\n\tPlease check the font config in your \
                  alacritty.yml.",
                 desc
             ),
             Error::FontNotLoaded => f.write_str("Tried to use a font that hasn't been loaded"),
+            Error::DirectWriteError(hresult) => {
+                write!(f, "A DirectWrite rendering error occurred: {:#X}", hresult)
+            }
         }
+    }
+}
+
+fn em_size(size: Size) -> f32 {
+    size.as_f32_pts() * (96.0 / 72.0)
+}
+
+impl From<dwrote::Font> for Font {
+    fn from(font: dwrote::Font) -> Font {
+        Font {
+            face: font.create_font_face(),
+            family_name: font.family_name(),
+            weight: font.weight(),
+            style: font.style(),
+            stretch: font.stretch(),
+        }
+    }
+}
+
+impl From<Weight> for FontWeight {
+    fn from(weight: Weight) -> FontWeight {
+        match weight {
+            Weight::Bold => FontWeight::Bold,
+            Weight::Normal => FontWeight::Regular,
+        }
+    }
+}
+
+impl From<Slant> for FontStyle {
+    fn from(slant: Slant) -> FontStyle {
+        match slant {
+            Slant::Oblique => FontStyle::Oblique,
+            Slant::Italic => FontStyle::Italic,
+            Slant::Normal => FontStyle::Normal,
+        }
+    }
+}
+
+fn get_current_locale() -> String {
+    let mut buf = vec![0u16; LOCALE_NAME_MAX_LENGTH];
+    let len = unsafe { GetUserDefaultLocaleName(buf.as_mut_ptr(), buf.len() as i32) as usize };
+
+    // len includes null byte, we don't need that in Rust strings
+    OsString::from_wide(&buf[..len - 1]).into_string().expect("Locale not valid unicode")
+}
+
+/// Information needed by dwrote's TextAnalysisSource to find a fallback font
+/// for a run of text. See `get_fallback_font`.
+struct TextAnalysisSourceData<'a> {
+    locale: &'a str,
+    length: u32,
+}
+
+impl TextAnalysisSourceMethods for TextAnalysisSourceData<'_> {
+    fn get_locale_name(&self, _text_position: u32) -> (Cow<str>, u32) {
+        (Cow::Borrowed(&self.locale), self.length)
+    }
+
+    fn get_paragraph_reading_direction(&self) -> dwrite::DWRITE_READING_DIRECTION {
+        dwrite::DWRITE_READING_DIRECTION_LEFT_TO_RIGHT
     }
 }
--- a/font/src/directwrite/mod.rs
+++ b/font/src/directwrite/mod.rs
@@ -33,7 +33,7 @@ use super::{
 };
 
 /// Cached DirectWrite font.
-pub struct Font {
+struct Font {
     face: FontFace,
     family_name: String,
     weight: FontWeight,
@@ -184,8 +184,7 @@ impl crate::Rasterize for DirectWriteRasterizer {
 
         let line_height = f64::from(ascent - descent + line_gap);
 
-        // We assume that all monospace characters have the same width
-        // Because of this we take '!', the first drawable character, for horizontal metrics
+        // Since all monospace characters have the same width, we use `!` for horizontal metrics
         let c = '!';
         let glyph_index = self.get_glyph_index(face, c)?;
 
@@ -345,7 +344,7 @@ struct TextAnalysisSourceData<'a> {
 
 impl TextAnalysisSourceMethods for TextAnalysisSourceData<'_> {
     fn get_locale_name(&self, _text_position: u32) -> (Cow<str>, u32) {
-        (Cow::Borrowed(&self.locale), self.length)
+        (Cow::Borrowed(self.locale), self.length)
     }
 
     fn get_paragraph_reading_direction(&self) -> dwrite::DWRITE_READING_DIRECTION {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -465,7 +465,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "dwrote"
-version = "0.9.0"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -473,6 +473,7 @@ dependencies = [
  "serde 1.0.104 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive 1.0.104 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "wio 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -548,13 +549,14 @@ dependencies = [
  "core-foundation-sys 0.6.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-graphics 0.17.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "core-text 13.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "dwrote 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "dwrote 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "euclid 0.20.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "foreign-types 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "freetype-rs 0.24.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.67 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "servo-fontconfig 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -2124,6 +2126,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "wio"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "ws2_32-sys"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2239,7 +2249,7 @@ dependencies = [
 "checksum dlib 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)" = "77e51249a9d823a4cb79e3eca6dcd756153e8ed0157b6c04775d04bf1b13b76a"
 "checksum downcast-rs 1.1.1 (registry+https://github.com/rust-lang/crates.io-index)" = "52ba6eb47c2131e784a38b726eb54c1e1484904f013e576a25354d0124161af6"
 "checksum dtoa 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)" = "4358a9e11b9a09cf52383b451b49a169e8d797b68aa02301ff586d70d9661ea3"
-"checksum dwrote 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)" = "0bd1369e02db5e9b842a9b67bce8a2fcc043beafb2ae8a799dd482d46ea1ff0d"
+"checksum dwrote 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)" = "439a1c2ba5611ad3ed731280541d36d2e9c4ac5e7fb818a27b604bdc5a6aa65b"
 "checksum embed-resource 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)" = "bbaba4684ab0af1cbb3ef0b1f540ddc4b57b31940c920ea594efe09ab86e2a6c"
 "checksum env_logger 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)" = "44533bbbb3bb3c1fa17d9f2e4e38bbbaf8396ba82193c4cb1b6445d711445d36"
 "checksum euclid 0.20.7 (registry+https://github.com/rust-lang/crates.io-index)" = "3f852d320142e1cceb15dccef32ed72a9970b83109d8a4e24b1ab04d579f485d"
@@ -2431,6 +2441,7 @@ dependencies = [
 "checksum winpty 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)" = "92c5a39bb2408a307dd5ab774039ec3f2c68d052970ae4dacc346101abe202b2"
 "checksum winpty-sys 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)" = "ed8a179a59760dc51d30b5e6eaf1bd6da88461f72f2616e962ddebef7e413210"
 "checksum winreg 0.6.2 (registry+https://github.com/rust-lang/crates.io-index)" = "b2986deb581c4fe11b621998a5e53361efe6b48a151178d0cd9eeffa4dc6acc9"
+"checksum wio 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)" = "5d129932f4644ac2396cb456385cbf9e63b5b30c6e8dc4820bdca4eb082037a5"
 "checksum ws2_32-sys 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)" = "d59cefebd0c892fa2dd6de581e937301d8552cb44489cdff035c6187cb63fa5e"
 "checksum x11-clipboard 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)" = "e5e937afd03b64b7be4f959cc044e09260a47241b71e56933f37db097bf7859d"
 "checksum x11-dl 2.18.5 (registry+https://github.com/rust-lang/crates.io-index)" = "2bf981e3a5b3301209754218f962052d4d9ee97e478f4d26d4a6eced34c1fef8"


kchibisov:
@davidhewitt could you show how glyphs looks like without downsampling on windows?

davidhewitt:
If I don't downsample, then the glyphs are sharp. But their width is much larger than for my current font (Cascadia Code), so we end up with alignment issues:

image

I've highlighted one smiley to help show the size issue. Alacritty is drawing a highlight rect around what it thinks is two glyphs in width (because the emoji are double-width glyphs), but it doesn't quite capture the smiley face properly. Notice the right edge is cut off. (I think Alacritty _is_ actually rendering the rest of the glyph, but it's black on black background so we can't see it.)

davidhewitt:
In Windows Terminal the alignment looks correct, though I don't see any downsampling code. However, its rendering is slightly different to Alacritty's. I think it renders directly to the window surface rather than rasterizing glyphs to bitmaps first as Alacritty does.

image

kchibisov:
>Alacritty is drawing a highlight rect around what it thinks is two glyphs in width (because the emoji are double-width glyphs), but it doesn't quite capture the smiley face properly.

Emojis are not two width glyphs in the context of terminal app, their just glyphs with width defined unicode-width-crate(wcwidth in other terminals, etc), so they can be with width equaled to 1.

>I've highlighted one smiley to help show the size issue.

Ehm, it's a bit strange to be honest, so rather then using downsampling you should understand why they're so much to the right. On Linux/BSD downsampling exists to solve a different issue, since emojis on the these platforms are not scaled at all (you're getting for example 138x128 pixels bitmap for every font size and should scale it yourself).

I don't like the windows terminal approach tbh, however I feel like it's just their rendering.

kchibisov:
>The remaining step to add full emoji support is to implement color font rasterizing, which will be a much bigger change and might need upstream support in the dwrote crate.

Oh, could you go into more details why we can't just implement color font support with what we have right now? I'm not forcing you to add it here, but just wondering.

davidhewitt:
> rather then using downsampling you should understand why they're so much to the right.

I query the metrics for each glyph from the font face returned from the fallback API. IIRC, for the highlighted smiley the advance is expected to be about 25 pixels, but Alacritty is only giving the glyph 22 pixels of space (i.e. two monospace columns). All layout is done from the left, which is why the glyph appears to be off to the right.

> Oh, could you go into more details why we can't just implement color font support with what we have right now?

Starting from this line is where Windows Terminal adds color emojis:

https://github.com/microsoft/terminal/blob/c274b38dccc2abdd1ea2fd5a27a3bd2e391e0bcb/src/renderer/dx/CustomTextRenderer.cpp#L304

There's a number of pieces needed for us to reproduce all that:

- We need access to the DirectWrite factory pointer, but that's currently not exposed by `dwrote`.
- We also then need to cast that pointer to the COM interface `IDWriteFactory4`, which isn't even exposed in `winapi`. So we'd need to define that and a few related interfaces ourselves, and then upstream them.
- Once we've got all that, we're rasterizing to bitmaps instead of the window, so we need to pull in WIC to create bitmap render targets for Direct2d. So even the render code is more complicated than what Windows Terminal is doing.

All of that stuff is also going to add a whole load of `unsafe` calls to various C++ APIs, so I thought we should pause half-way with greyscale rendering. (In the interim I wanted to look at what I might upstream to `dwrote` or `winapi`.)

kchibisov:
I'll have an access to a Windows machine for a next few days, so I'll try to play with this fallbacks and "slightly big" emojis thing.

kchibisov:
You should sort out locale issue, otherwise it's nearly impossible for me to test it. I've changed my locale to much my system locale (hardcode thing), however it still don't work as expected, since I'm getting missing glyphs symbols, followed by emoji in some cases in other just missing glyph symbol.

I also had to install wsl, since emojis wasn't working for me in cmd.exe and powershell (Windows was just removing them from paste!!!!). At least I can say that CJK glyphs are seems to work fine now. (A small win?)

@davidhewitt Do I need something special to make things work like on your screenshot?

davidhewitt:
I've added a commit which uses the correct locale.

> Do I need something special to make things work like on your screenshot?

Not as far as I'm aware. Are you using ConPTY + Windows 10? WinPTY + emojis doesn't work for me in testing.

kchibisov:
Yeah, ConPty + Windows 10. Windows just decides to strip emojis sometime from input and paste random things in alacritty.

davidhewitt:
> Yeah, ConPty + Windows 10. Windows just decides to strip emojis sometime from input and paste random things in alacritty.

Weird, that would be worth investigating as a separate issue perhaps. Might be a clipboard bug our side, or ConPTY bug to report upstream. Paste _seems_ to be working for me. For now I've just manually created a text file in VSCode which has a bunch of emoji in it, and just displaying that with `cat`.

> since I'm getting missing glyphs symbols, followed by emoji in some cases in other just missing glyph symbol.

I _occasionally_ get a missing glyph symbol which disappears on a resize. To me that suggests an upstream bug in ConPTY. (I'll try to debug that.)

kchibisov:
>Weird, that would be worth investigating as a separate issue perhaps. Might be a clipboard bug our side, or ConPTY bug to report upstream. Paste seems to be working for me. For now I've just manually created a text file in VSCode which has a bunch of emoji in it, and just displaying that with cat.

I'll try to create something similar to repro.

>I occasionally get a missing glyph symbol which disappears on a resize. To me that suggests an upstream bug in ConPTY. (I'll try to debug that.)

Yeah, it seems like it. Knowing how alacritty clipboard works it seems like a ConPty issue. The key thing here is that if you copy a missing glyph symbol and paste to e.g. Firefox, it'll still be a missing glyph symbol, but not the underlying text (there should be some char causing missing glyph thing).

davidhewitt:
>  I don't see you passing through the font size to dwrote for font fallback, maybe there's something missing?

This is a really interesting point - perhaps I should not be calling `create_font_face` and should instead be doing something slightly different to get a font face of the right size. I'll try to find some time to investigate some point this week...

davidhewitt:
Some further investigation and I think I've cracked it.

I'm doing the right thing with the font size: the font size is not relevant to the DirectWrite font face. The font size is only relevant when actually rasterizing the glyph - as part of the `DWRITE_GLYPH_RUN` structure.

I've also found how Windows Terminal solves this issue. Instead of downsampling, it dynamically rescales the font size so that the glyph fits within the cell width. The algorithm is here:

https://github.com/microsoft/terminal/blob/6f7ad99d514c55acc02a8052e208b3a923671c9b/src/renderer/dx/CustomTextLayout.cpp#L425

I'm going to adapt this PR to do the same.

kchibisov:
FWIW I don't like the idea of using unicode width crate to make things fit into cell. Just let things overlap, it's not a problem tbh, this 1 width emojis are just bad and you'll be consistent with out other backends. So if you want to follow windows terminal here, just don't use unicode-width to solve problem...

chrisduerr:
> FWIW I don't like the idea of using unicode width crate to make things fit into cell. Just let things overlap, it's not a problem tbh, this 1 width emojis are just bad and you'll be consistent with out other backends. So if you want to follow windows terminal here, just don't use unicode-width to solve problem...

Yeah, this is true. Since I've just recently updated our unicode-width crate to the latest version of the standard, I can confidently say that non-combined glyphs all have the correct width in Alacritty. So any width mismatch between OS and Alacritty's width function should be an indication of a bug in the OS' library.

davidhewitt:
I've tidied up the remaining pieces on this branch. To remove the `unicode-width` dependency, I'm now going for a much simpler downsizing rule:

- If the glyph from the fallback font is wider than two grid cells, shrink the font size for that glyph so the glyph fits neatly in two grid cells.
- Don't scale anything else.

Hopefully you agree this is a reasonable compromise between keeping the rendering tidy and avoiding the `unicode-width` dependency.

Attached is a screenshot of the greyscale rendering now:

image

The remaining problems @kchibisov was encountering (paste & missing glyphs) appear to be upstream ConPTY issues; I may look into them but that'll be upstream bugs.

In summary, this PR is feature-complete as far as I'm concerned. I plan to follow up with a separate one in the future with color rendering for emojis once I figure out the best places to put all the pieces.

kchibisov:
>I've tidied up the remaining pieces on this branch. To remove the unicode-width dependency, I'm now going for a much simpler downsizing rule:
>
>    If the glyph from the fallback font is wider than two grid cells, shrink the font size for that glyph so the glyph fits neatly in two grid cells.
    Don't scale anything else.
>
>Hopefully you agree this is a reasonable compromise between keeping the rendering tidy and avoiding the unicode-width dependency.

Yeah, the picture and the idea LGTM.

>The remaining problems @kchibisov was encountering (paste & missing glyphs) appear to be upstream ConPTY issues; I may look into them but that'll be upstream bugs.

Could you check that our Windows clipboard is not at fault here? Like making it load and the instantly store to it and paste back to some app? If you need help on how to do it let me know.

davidhewitt:
> Could you check that our Windows clipboard is not at fault here? Like making it load and the instantly store to it and paste back to some app?

So I did some quick testing and found the following:
- Copying from Alacritty is fine. I can paste the copied emojis into VSCode no problem.
- Powershell's built-in paste works correctly. This is the paste behavior which I saw was working.
- Alacritty's paste action (Control + Shift + V) does not work correctly. This is the broken paste behaviour you're seeing.

So there _is_ possibly an issue with our clipboard code. I wonder if it's a wider issue beyond just emojis (i.e. might be a problem with a number of unicode glyphs).

font/src/directwrite/mod.rs
```diff
@@ -13,33 +13,207 @@
...
     device_pixel_ratio: f32,
+    font_collection: dwrote::FontCollection,
 }
 
-impl crate::Rasterize for DirectWriteRasterizer {
-    type Err = Error;
+impl DirectWriteRasterizer {
+    fn rasterize_glyph_with_fallback(&mut self, glyph: GlyphKey) -> Result<RasterizedGlyph, Error> {
+        self.rasterize_glyph(glyph).or_else(|error| {
+            if let Error::MissingGlyph(_) = &error {
```
chrisduerr: It's bad style to have two levels of indentation for the entire method. These first two conditions should likely be combined for an early return.

font/src/directwrite/mod.rs
```diff
@@ -13,33 +13,207 @@
...
+                // Try to get a fallback font
+                let font = &self.fonts.get(&glyph.font_key).unwrap().font;
```
chrisduerr: Is this `unwrap` safe? It doesn't look like something we should be doing.

font/src/directwrite/mod.rs
```diff
@@ -13,33 +13,207 @@
...
+                    &dwrote::TextAnalysisSource::from_text_and_number_subst(
+                        Box::new(TextAnalysisSource {
+                            locale: locale.clone(),
+                            length: num_codepoints,
+                        }),
+                        utf16_codepoints,
+                        dwrote::NumberSubstitution::new(
+                            dwrite::DWRITE_NUMBER_SUBSTITUTION_METHOD_NONE,
+                            &locale,
+                            false,
```
chrisduerr: Why clone locale earlier and then pass a reference here, instead of just putting this outside of the `map_characters` function and then not requiring to clone later anymore?

font/src/directwrite/mod.rs
```diff
@@ -13,33 +13,207 @@
...
+    fn rasterize_glyph_with_fallback(&mut self, glyph: GlyphKey) -> Result<RasterizedGlyph, Error> {
+        self.rasterize_glyph(glyph).or_else(|error| {
+            if let Error::MissingGlyph(_) = &error {
+                // Try to get a fallback font
+                let font = &self.fonts.get(&glyph.font_key).unwrap().font;
+                let locale = get_current_locale();
+                let utf16_codepoints = encode_utf16(glyph.c);
+                let num_codepoints = utf16_codepoints.len() as u32;
+
+                let fallback = dwrote::FontFallback::get_system_fallback().unwrap();
```
chrisduerr: Another `unwrap` where I cannot tell if it's safe or not.

Generally we should prefer `except` over `unwrap` too, even if it's mostly safe. Just so we can find the error with stripped info without any issue.

font/src/directwrite/mod.rs
```diff
@@ -13,33 +13,207 @@
...
+                        ),
+                    ),
+                    0,
+                    num_codepoints,
+                    &self.font_collection,
+                    Some(&font.family_name()),
+                    font.weight(),
+                    font.style(),
+                    font.stretch(),
+                );
```
chrisduerr: There's so much identation here that it becomes difficult to understand what exactly is going on in this massive statement.

Could we maybe extract some statements, like the `from_text_and_muber_subst` call? Possibly also the `NumberSubstitution` and `TextAnalysisSource`.

font/src/directwrite/mod.rs
```diff
@@ -13,33 +13,207 @@
...
+
+                if let Some(mapped_font) = fallback_font.mapped_font {
```
chrisduerr: What if `mapped_font` is `None`? When would this be the case? Currently we explode right now, so we might as well make that explicit by returning the error directly from this point, rather than making it look like we're performing error correction?

font/src/directwrite/mod.rs
```diff
@@ -13,33 +13,207 @@
...
+                            weight: font.weight().into(),
+                        },
+                    );
+
+                    let fallback_font_key = self.get_matching_face(&desc)?;
+
+                    let mut fallback_glyph = glyph;
+                    fallback_glyph.font_key = fallback_font_key;
+
+                    return self.rasterize_fallback_glyph(fallback_glyph, glyph);
```
chrisduerr: Hopefully explicit return stataments shouldn't be required anymore when the indentation level is fixed. This kinda makes the control flow messy and hard to understand.

font/src/directwrite/mod.rs
```diff
@@ -13,33 +13,207 @@
...
+    fonts: HashMap<FontKey, LoadedFontFace>,
+    keys: HashMap<FontDesc, FontKey>,
     device_pixel_ratio: f32,
+    font_collection: dwrote::FontCollection,
 }
 
-impl crate::Rasterize for DirectWriteRasterizer {
-    type Err = Error;
+impl DirectWriteRasterizer {
+    fn rasterize_glyph_with_fallback(&mut self, glyph: GlyphKey) -> Result<RasterizedGlyph, Error> {
```
chrisduerr: I'm not sure about the naming here, since we want to always use this method for rasterizing the glyph basically.

So I feel like `rasterize_glyph` would be the correct name and the `rasterize_glyph` method should be renamed instead (since it's basically just a part of the glyph rasterization process).

font/src/directwrite/mod.rs
```diff
@@ -13,33 +13,207 @@
...
+        &mut self,
+        glyph: GlyphKey,
+        original_glyph: GlyphKey,
+    ) -> Result<RasterizedGlyph, Error> {
+        // The fallback font is unlikely to have the same advance size as the monospace font.
+        // Therefore we may need to dynamically decrease the font size so the glyph fits in the
+        // cell width.
+        //
+        // We'll only do this for any glyph > 2 cells wide, downsizing it to be just 2 cells wide.
+        // This should help keep things tidy without resizing many glyphs needlessly.
```
chrisduerr: Isn't a halfwidth font just as likely to exceed the width as a fullwidth one?

It seems a bit strange that the widths would differentiate so strongly, since our fonts sizes are specified in points.

font/src/directwrite/mod.rs
```diff
@@ -13,33 +13,207 @@
...
+        if actual_advance > max_advance {
+            let scale_factor = (max_advance / actual_advance) as f32;
+
+            let mut scaled_glyph = glyph;
+            scaled_glyph.size = Size::new(glyph.size.as_f32_pts() * scale_factor);
+
+            self.rasterize_glyph(scaled_glyph)
+        } else {
+            self.rasterize_glyph(glyph)
+        }
```
chrisduerr: Reusing `rasterize_glyph` is nice, but calling it twice here still seems a bit strange to me.

Can't we just mutate `glyph` in the `if` block and then always call `self.rasterize_glyph(glyph)` afterwards?

font/src/directwrite/mod.rs
```diff
@@ -13,33 +13,207 @@
...
+        // We'll only do this for any glyph > 2 cells wide, downsizing it to be just 2 cells wide.
+        // This should help keep things tidy without resizing many glyphs needlessly.
+
+        let font_metrics = self.metrics(original_glyph.font_key, original_glyph.size)?;
+        let glyph_metrics = self.metrics_from_glyph(glyph)?;
+
+        let max_advance = 2.0 * font_metrics.average_advance;
+        let actual_advance = glyph_metrics.average_advance;
+
+        if actual_advance > max_advance {
```
chrisduerr: As mentioned in another comment, I don't think this is a reliable way to determine if we need to scale the font.

font/src/directwrite/mod.rs
```diff
@@ -13,33 +13,207 @@
...
+            let mut scaled_glyph = glyph;
+            scaled_glyph.size = Size::new(glyph.size.as_f32_pts() * scale_factor);
+
+            self.rasterize_glyph(scaled_glyph)
+        } else {
+            self.rasterize_glyph(glyph)
+        }
+    }
+
+    fn metrics_from_glyph(&self, glyph: GlyphKey) -> Result<Metrics, Error> {
```
chrisduerr:
```suggestion
    fn glyph_metrics(&self, glyph: GlyphKey) -> Result<Metrics, Error> {
```

font/src/directwrite/mod.rs
```diff
@@ -215,3 +356,51 @@ impl ::std::fmt::Display for Error {
...
+}
+
+impl From<FontWeight> for Weight {
+    fn from(other: FontWeight) -> Weight {
+        match other {
+            FontWeight::Bold => Weight::Bold,
+            _ => Weight::Normal,
+        }
+    }
+}
```
chrisduerr: Do we need both here, or is this just for completeness? Generally I think it's nicer to have one 'target' format that is used everywhere and we convert to that as soon as possible.

If we need to pass one of our `Weight`s into the `dwrote` crate and we get a `FontWeight` out of `dwrote`, which we need to convert back to our weight, this is obviously the way to go though.

Same applies for `Slant`/`FontStyle` too of course.

davidhewitt: I've gone for just `Weight` -> `dwrote::FontWeight` etc.

font/src/directwrite/mod.rs
davidhewitt: > Isn't a halfwidth font just as likely to exceed the width as a fullwidth one?

Yes, a halfwidth glyph might overflow one cell just as a fullwidth glyph might overflow two cells.

Now I've removed the `unicode-width` dependency I don't see a way to reliably handle the halfwidth case. (I cannot find any DirectWrite API to ask if a codepoint is half- or full-width.)

font/src/directwrite/mod.rs
davidhewitt: > As mentioned in another comment, I don't think this is a reliable way to determine if we need to scale the font.

I agree that it's not perfect, though haven't managed to come up with better. What would you like to see instead?

davidhewitt:
It seems like the paste issue is likely upstream with ConPTY, as Windows Terminal has the same problem: https://github.com/microsoft/terminal/issues/1503

font/src/directwrite/locale.rs
```diff
@@ -0,0 +1,13 @@
+use std::ffi::OsString;
+use std::os::windows::ffi::OsStringExt;
+
+use super::winapi::shared::ntdef::LOCALE_NAME_MAX_LENGTH;
+use super::winapi::um::winnls::GetUserDefaultLocaleName;
+
+pub fn get_current_locale() -> String {
```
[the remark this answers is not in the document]
davidhewitt: Yes, I was anticipating it to be more complicated and then it turned out short. I'll move it back into `mod.rs`.

font/src/directwrite/mod.rs
```diff
@@ -14,32 +14,223 @@
...
+use std::borrow::Cow;
+use std::collections::HashMap;
+
 use self::dwrote::{
-    FontCollection, FontStretch, FontStyle, FontWeight, GlyphOffset, GlyphRunAnalysis,
+    Font, FontCollection, FontFace, FontFallback, FontStretch, FontStyle, FontWeight, GlyphOffset,
+    GlyphRunAnalysis,
 };
 
+use self::winapi::um::dwrite;
```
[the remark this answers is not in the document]
davidhewitt: Unfortunately these `extern crate` commands aren't done in `lib.rs`, which scopes them to only this module but also means `self` is needed.

I'm actually just going to add a PR to bump the font crate to 2018 edition.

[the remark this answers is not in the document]
chrisduerr: > I'm actually just going to add a PR to bump the font crate to 2018 edition.

Yeah, I think that's a good idea. I honestly don't even know how 2015 works exactly anymore myself (though I swear that worked before!).

font/src/directwrite/mod.rs
```diff
@@ -14,32 +14,223 @@
...
+                        &font_entry.font_face,
+                        glyph.c,
+                        glyph.size,
+                    )
+                });
+
+            if fallback_result.is_ok() {
+                return fallback_result;
+            }
+        }
```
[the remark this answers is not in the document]
davidhewitt: Thanks, that's much nicer!

font/src/directwrite/mod.rs
```diff
@@ -14,32 +14,223 @@
...
+
+        let locale = get_current_locale();
+        let utf16_codepoints = encode_utf16(c);
+        let length = utf16_codepoints.len() as u32;
+
+        let number_subst = dwrote::NumberSubstitution::new(
+            dwrite::DWRITE_NUMBER_SUBSTITUTION_METHOD_NONE,
+            &locale,
+            false,
+        );
```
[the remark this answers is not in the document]
davidhewitt: I don't think we actually need it for what we're doing here; I'm going to submit a quick PR upstream to `dwrote` which allows us to get rid of this here.

font/src/directwrite/mod.rs
```diff
@@ -14,32 +14,223 @@
...
+            Some(&font.family_name()),
+            font.weight(),
+            font.style(),
+            font.stretch(),
+        );
+
+        Ok(fallback_font.mapped_font)
+    }
+
+    fn rasterize_font_char(
```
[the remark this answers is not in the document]
davidhewitt: Yes; I'm struggling to come up with a better name though. Will keep thinking.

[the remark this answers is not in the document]
chrisduerr: I can see if I can come up with some ideas once all the other problems are resolved. Since it's just a style issue getting all the other stuff figured out first probably makes more sense.

font/src/directwrite/mod.rs
davidhewitt: > Is there some documentation on this being required, or recommended? Or did you mainly look at the source code of the windows terminal without a proper name?

There's no documentation I've found on it. I think it's just a sad fact that Windows' emoji font has completely different widths to my monospace font. Despite Windows' having selected this font supposedly as a suitable fallback!

This algorithm is what I ended up with after a few varied attempts:

- My initial implementation had no scaling, and I saw glyphs with slight offset issues like in [this comment](https://github.com/alacritty/alacritty/pull/3127#issuecomment-569404916)
- To fix the offset problem I then attempted downscaling [like in the OP](https://github.com/alacritty/alacritty/pull/3127#issue-357228997). TBH these images don't look great. At this point I was already using glyph advance to calculate the factor to downscale by.
- I then decided to peek at what the Windows Terminal was doing. I felt their strategy of scaling the font size could work here too, which is why I then decided to implement the same.

It's not great, but I think it looks better than either the out-of-bounds fonts or the downsampled ones from the OP.

chrisduerr: I personally find it questionable to implement such a thing on Windows, while opting to overlap on other platforms. We should be consistent across platforms.

And as we found out in this PR, the process of downscaling isn't simple either. Scaling the output glyph can easily lead to poor looking fonts so it's not really an option. But scaling the font size can not guarantee that the result actually works properly, especially with fonts that are not strictly monospace (which is likely the case for emojis).

kchibisov: >Neither macOS nor Linux/BSD require scaling fonts after fallback. It's strange to me that this would be required on Windows, without providing a recommended way of dealing with this.

Speaking of Linux here, it does require your app to scale things(fixup_factor) and it can also provide an odd scale factor, since user can just set one in a config. So technically, we're scaling on Linux to a user defined point, so it's just smarter than whatever Windows is providing.

>I personally find it questionable to implement such a thing on Windows, while opting to overlap on other platforms. We should be consistent across platforms.

We will be consistent with such change on all platforms, since our overlapping is related to `wcwidth`, however this PR doesn't speculate it and just kind of using more smarter `downsampling` than for example Linux/BSD impl.

>And as we found out in this PR, the process of downscaling isn't simple either. Scaling the output glyph can easily lead to poor looking fonts so it's not really an option. But scaling the font size can not guarantee that the result actually works properly, especially with fonts that are not strictly monospace (which is likely the case for emojis).

They are full-width fonts, we can perform such scaling only on colored fonts tbh, since we can query such information IIRC.

Anyway, @davidhewitt, could you show me how this string looks like `❄m❄` on your PR?

font/src/directwrite/mod.rs
```diff
@@ -14,32 +14,223 @@
...
     BitmapBuffer, FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph, Size, Slant, Style, Weight,
 };
 
+mod locale;
+use self::locale::get_current_locale;
+
+pub struct TextAnalysisSource {
+    locale: String,
+    length: u32,
+}
```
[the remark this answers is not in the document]
davidhewitt: Will add comment and remove the superfluous `pub`

font/src/directwrite/mod.rs
```diff
@@ -14,32 +14,223 @@
...
+    fn get_locale_name(&self, _text_position: u32) -> (Cow<str>, u32) {
+        (Cow::Borrowed(&self.locale), self.length)
+    }
+
+    fn get_paragraph_reading_direction(&self) -> dwrite::DWRITE_READING_DIRECTION {
+        dwrite::DWRITE_READING_DIRECTION_LEFT_TO_RIGHT
+    }
+}
+
+pub struct FontEntry {
```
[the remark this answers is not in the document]
davidhewitt: I'll go for `LoadedFont`.

font/src/directwrite/mod.rs
davidhewitt: > Anyway, @davidhewitt, could you show me how this string looks like `❄m❄` on your PR?

Sure, it looks like this when pasted in (I typed a few spaces after to get the cursor out the way):

image

kchibisov: One more note, some odd fonts on Linux like `Inconsolata` are playing really bad with all the other fonts on Linux, so technically even normal chars from other fonts are starts to overlap each other, not speaking of emojis. So technically @chrisduerr is right here, I guess.

davidhewitt:
I fixed a number of the style issues and rebased on master.

This PR won't build at the moment. I opened two PRs against `dwrote` which we will need released before this is mergeable. The code as written in this PR currently assumes both of them:
- https://github.com/servo/dwrote-rs/pull/37
- https://github.com/servo/dwrote-rs/pull/38

font/src/directwrite/mod.rs
davidhewitt: Ok. It looks like the consensus is to just not do the scaling, so I'll remove it now.

kchibisov: >Sure, it looks like this when pasted in (I typed a few spaces after to get the cursor out the way):

The picture is absolutely correct though, we and the most of other terminals, have the same thing on Linux. So now it's just a matter of using the same. However we can implement the same algo for font change on Linux, it should be pretty trivial.

kchibisov: @davidhewitt wait a bit.

davidhewitt: Ok. I'm going to head to bed for the night; will pick this up at a later date when a decision has been made. My personal preference is to have the scaling in so the emojis line up nicely, but I will remove it if that's the consensus from everyone else.

We have some time to mull this over as we need the `dwrote` PRs merged before this is mergable now anyway.

chrisduerr:
> I fixed a number of the style issues and rebased on master.
>
> This PR won't build at the moment. I opened two PRs against dwrote which we will need released before this is mergeable.

Thanks for keeping us posted.

font/src/directwrite/mod.rs
kchibisov: After some discussion on IRC, and further investigation with our Linux/BSD backend, we decided that we shouldn't scale things for now and hold it back until there will be a lot of complains in the future. The algorithm since a bit of fragile tbh, since some font metrics are not that much of reliable, knowing that we can fallback to a glyph from non monospace font and there's always tons of use cases.

Also we do have a font rfc #957 and there's an idea on providing some set of options to a font, so with it you can just change a font size to "bad" fonts in your fallback to better match a primary font, since it's more reliable way to do things than guessing based off some font metrics.

tl;dr You issue should be covered by #957 and we should revert any scaling to match our other backends for now.

davidhewitt:
Font scaling removed. Hopefully the `dwrote` PRs don't get stuck for too long. The one removing `NumberSubstitution` would especially be desirable to merge so we don't have pointless code kicking around here. The other one is just a convenience method that we could include as a free function here if needed.

davidhewitt:
The `dwrote` PRs have now been merged, but not yet released. I've pushed a commit which switches to use the `git` version of `dwrote` (pinned to a specific revision), and rebases on master.

Barring a little bikeshedding on function names (the unresolved comment above), and whether you want to wait for a full `dwrote` release, this is now ready to go.

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,157 @@
...
     device_pixel_ratio: f32,
+    font_collection: FontCollection,
+    font_fallback: Option<FontFallback>,
 }
 
-impl crate::Rasterize for DirectWriteRasterizer {
-    type Err = Error;
+impl DirectWriteRasterizer {
+    fn rasterize_glyph(&mut self, glyph: GlyphKey) -> Result<RasterizedGlyph, Error> {
+        let loaded_font = &self.fonts.get(&glyph.font_key).ok_or(Error::FontNotLoaded)?;
```
kchibisov: This function looks like "pointer to impl" thing, what do you thing about moving this logic into `get_glyph` method and merge `rasterize_glyph` and `rasterize_font_char` in one function, so it'll solve a problem with naming and also remove one layer of abstraction.

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,157 @@
...
+                self.rasterize_font_char(&fallback_font.create_font_face(), glyph.c, glyph.size)
+            },
+            Err(err) => Err(err),
+        }
     }
 
-    fn metrics(&self, key: FontKey, size: Size) -> Result<Metrics, Error> {
-        let font = self.fonts.get(key.token as usize).ok_or(Error::FontNotLoaded)?;
+    fn get_fallback_font(&self, font: &Font, c: char) -> Result<Font, Error> {
+        let fallback = match self.font_fallback.as_ref() {
```
chrisduerr:
```suggestion
        let fallback = match &self.font_fallback {
```

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,157 @@
...
+            Some(fallback) => fallback,
+            None => return Err(Error::MissingGlyph(c)),
+        };
+
+        let utf16_codepoints = encode_utf16(c);
+        let length = utf16_codepoints.len() as u32;
+
+        let locale = get_current_locale();
+        let text_analysis_source = dwrote::TextAnalysisSource::from_text(
```
chrisduerr:
```suggestion
        let text_analysis_source = TextAnalysisSource::from_text(
```

Should probably `use dwrote::TextAnalysisSource` at the top.

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,157 @@
...
+        &self,
+        font_face: &FontFace,
+        c: char,
+        size: Size,
+    ) -> Result<RasterizedGlyph, Error> {
+        let offset = GlyphOffset { advanceOffset: 0.0, ascenderOffset: 0.0 };
+
+        let glyph_index = *font_face
+            .get_glyph_indices(&[c as u32])
+            .first()
```
chrisduerr:
```suggestion
            .first()
            // DirectWrite returns 0 if the glyph does not exist in the font
            .filter(|glyph_index| glyph_index == 0)
```

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,157 @@
...
+
+        let em_size = em_size(size);
+
+        let glyph_run = dwrote::DWRITE_GLYPH_RUN {
+            fontFace: unsafe { font_face.as_ptr() },
+            fontEmSize: em_size,
+            glyphCount: 1,
+            glyphIndices: &(glyph_index),
+            glyphAdvances: &(0.0),
+            glyphOffsets: &(offset),
```
chrisduerr: What are these parenthesis for?

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,157 @@
...
+        let glyph_analysis = GlyphRunAnalysis::create(
+            &glyph_run,
+            self.device_pixel_ratio,
+            None,
+            rendering_mode,
+            dwrote::DWRITE_MEASURING_MODE_NATURAL,
+            0.0,
+            0.0,
+        )
+        .or_else(|_| Err(Error::MissingGlyph(c)))?;
```
chrisduerr: Is this error really correct? We shouldn't be returning a missing glyph error if that's not the actual problem.

Same goes for all the other errors in this function following this one.

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,157 @@
...
+            left: bounds.left,
+            buf: BitmapBuffer::RGB(buf),
+        })
+    }
+
+    fn font_metrics(&self, font_face: &FontFace, size: Size) -> Result<Metrics, Error> {
+        // We assume that all monospace characters have the same width
+        // Because of this we take '!', the first drawable character, for measurements
+        self.glyph_metrics(font_face, '!', size)
+    }
```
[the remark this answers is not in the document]
chrisduerr: In previous versions this was different, but I don't see multiple users of these metric functions anymore.

It seems like font_metrics, glyph_metrics and metrics can all just be merged into one function?

font/src/directwrite/mod.rs
```diff
@@ -107,73 +232,42 @@ impl crate::Rasterize for DirectWriteRasterizer {
...
 
-        let bounds = glyph_analysis
-            .get_alpha_texture_bounds(dwrote::DWRITE_TEXTURE_CLEARTYPE_3x1)
-            .or_else(|_| Err(Error::MissingGlyph(glyph.c)))?;
-        let buf = glyph_analysis
-            .create_alpha_texture(dwrote::DWRITE_TEXTURE_CLEARTYPE_3x1, bounds)
-            .or_else(|_| Err(Error::MissingGlyph(glyph.c)))?;
+    fn load_font(&mut self, desc: &FontDesc, _size: Size) -> Result<FontKey, Error> {
+        self.get_matching_face(desc)
```
chrisduerr: It seems like this `get_matching_face` function is also only ever called from here? Then we can probably inline it.

font/src/directwrite/mod.rs
```diff
@@ -212,3 +306,57 @@ impl ::std::fmt::Display for Error {
...
+}
+
+fn encode_utf16(c: char) -> Vec<u16> {
+    let mut buf = vec![0u16; c.len_utf16()];
+    c.encode_utf16(&mut buf);
+    buf
+}
+
+impl From<Weight> for FontWeight {
+    fn from(other: Weight) -> FontWeight {
```
chrisduerr:
```suggestion
    fn from(weight: Weight) -> FontWeight {
```

font/src/directwrite/mod.rs
```diff
@@ -212,3 +306,57 @@ impl ::std::fmt::Display for Error {
...
+        match other {
+            Weight::Bold => FontWeight::Bold,
+            Weight::Normal => FontWeight::Regular,
+        }
+    }
+}
+
+impl From<Slant> for FontStyle {
+    fn from(other: Slant) -> FontStyle {
```
chrisduerr:
```suggestion
    fn from(slant: Slant) -> FontStyle {
```

font/src/directwrite/mod.rs
```diff
@@ -212,3 +306,57 @@ impl ::std::fmt::Display for Error {
...
+}
+
+/// Information needed by dwrote's TextAnalysisSource to find a fallback font
+/// for a run of text. See `get_fallback_font`.
+struct TextAnalysisSourceData {
+    locale: String,
+    length: u32,
+}
+
+impl dwrote::TextAnalysisSourceMethods for TextAnalysisSourceData {
```
chrisduerr:
```suggestion
impl TextAnalysisSourceMethods for TextAnalysisSourceData {
```

Probably should add `use dwrote::TextAnalysisSourceMethods`.

font/src/directwrite/mod.rs
davidhewitt: Hangover from the previous author's implementation. I'll remove them

font/src/directwrite/mod.rs
davidhewitt: I've gone for adding a `DirectWriteError` variant which stores the `HRESULT` error code returned from these functions. We could potentially give the variant a more specific name like `GlyphRasterizationFailed`?

davidhewitt:
Thanks for all the feedback. I've made a number of tweaks to address those points, and also rebased on master.

font/Cargo.toml
```diff
@@ -23,4 +23,5 @@ core-graphics = "0.17"
 core-foundation-sys = "0.6"
 
 [target.'cfg(windows)'.dependencies]
-dwrote = { version = "0.9.0" }
+dwrote = { git = "https://github.com/servo/dwrote-rs", rev = "c2486e8" }
```
kchibisov: Just to note, I don't really mind in using git dependency on Windows for the time being.

chrisduerr: I do. We should usually stay away from git dependencies because we have no control over those retrospectively. With cargo we know immutability of versions is guaranteed.

It shouldn't be too hard to cooperate a release though.

chrisduerr:
Clippy and rustfmt also seem to have some remaining complaints. Once those are done I'll give it another review.

font/Cargo.toml
kchibisov: I know, I'm just saying that it's fine for me to see them on Windows tbh, since it's likely not that popular platform and support/cooperation could be hard. I don't want to see this mess like we had with winit/glutin on Linux tbh/macOS tbh.

chrisduerr: We've probably got over a thousand Windows users, so it shouldn't be taken lightly.

davidhewitt:
> 
> 
> Mostly looks good from my side, could you provide us with information on how this PR performs on `vtebench unicode-random-write` and also some information on how long does it take to get a fallback during running it?
> 
> ```rust
> let t1 = std::time::Instant::now();
> let fallback_font = self.get_fallback_font(&loaded_font.font, glyph.c).ok_or(err)?;
> let t2 = std::time::Instant::now();
> println!("Time for fallback {:?}", t2 - t1).
> ```

A random sample of the timings for this:

    Time for fallback 62.9µs
    Time for fallback 24.2µs
    Time for fallback 45.1µs
    Time for fallback 40µs
    Time for fallback 92.9µs
    Time for fallback 80.4µs
    Time for fallback 31.4µs
    Time for fallback 120.5µs
    Time for fallback 52.2µs
    Time for fallback 67µs
    Time for fallback 33.8µs
    Time for fallback 40.5µs
    Time for fallback 125.5µs

Also, for `vtebench` with 10MB input, it takes ~17 seconds to render on this branch, versus about 17 seconds on `master` too. For comparison, Windows terminal churns through the same file over about 5 minutes. I suspect `Alacritty` might be skipping frames as it buffers?

With this branch, the vtebench output has a lot more characters rendered as compared to master (e.g. CJK characters). However, the framerate does start noticeably slower than master (but not terrible), and then improves as the fallbacks for missing glyphs in the primary font are cached.

davidhewitt:
Fixed clippy errors.

kchibisov:
@davidhewitt you should output vtebench's output to a file and then `cat` it. 
so `vtebench unicode-random-write > out.txt` and then `cat out.txt`. Alacritty doesn't skip frames on purpose, it can start skipping them if your fallback is really slow, but we don't have any logic to do that, since fast terminal shouldn't skip frames, etc.

>Also, for vtebench with 10MB input, it takes ~17 seconds to render on this branch, versus about 17 seconds on master too. For comparison, Windows terminal churns through the same file over about 5 minutes. I suspect Alacritty might be skipping frames as it buffers?

how it could be so long. on Linux 500MB of unicode-random-write takes me 4.4s. I feel like this is just throttled by ConPTY/WinPTY. 17s for 10MB is odd.

I'm not sure whether times are good or not, compared to what we have now on Linux/BSD they are terrible, like 5-6x slower on average, but Windows also doesn't help here. I encourage you to take a look whether times for fallback could be any better. Ideally you should have max time of 30-40usec.

davidhewitt:
> @davidhewitt you should output vtebench's output to a file and then cat it.
so vtebench unicode-random-write > out.txt and then cat out.txt

This is exactly what I did.

> I'm not sure whether times are good or not, compared to what we have now on Linux/BSD they are terrible, like 5-6x slower on average, but Windows also doesn't help here. I encourage you to take a look whether times for fallback could be any better. Ideally you should have max time of 30-40usec.

Yes, windows performance on Alacritty isn't amazing compared to Linux. I've previously played with this and come to the conclusion that ConPTY is a bottleneck.

WRT making the max time of fallback faster, I can do a few tweaks to cache things like the locale and the font properties to avoid a few Windows API calls. It shaves off a little bit of time.

It would be possible to save a number allocations in the font fallback paths, but I'll need to do some further PRs to dwrote. Using those as well as the caching described, the average seems to be ~30usec, but the max time still occasionally spikes to 80-100usec.

kchibisov:
>Yes, windows performance on Alacritty isn't amazing compared to Linux. I've previously played with this and come to the conclusion that ConPTY is a bottleneck.

Well, this is due to Windows, since our internal structure is purely generic and the same for all backends, and we're using the same OpenGL APIs on each platform. So you're basically fighting ConPTY and WinPTY here.

>but the max time still occasionally spikes to 80-100usec.

Oh, the average of 30usec could be quite good actually for Windows. Spikes for 80-100usecs could be due to initial loading of font or some other stuff, like missing glyph, I guess. Well I don't have much to recommend, however caching some computations unlikely dramatically improve something. I'll recommend to check how much time are you spending on some Windows API calls.

You can also take a look on what we've done recently for Linux/BSD for fallbacks here, but it'll unlikely help you. https://github.com/alacritty/alacritty/blob/6b327b6f8f0f308ff8f46cdf551ce0d0f3eda60b/font/src/ft/mod.rs#L617

kchibisov:
Just to make it clear, the initial slowness is expected, since our render caching things and probing fallback for 30usec(you have 2ms in total for fallback and other stuff + rendering) hurts, however it should cache all the thing during one run of `unicode-random-write` or something similar to this, I guess.

Anyway, I'm just trying to say, that fallback performance is very important and we should make sure that it is compatible to our other backends(at least Linux/BSD one).

davidhewitt:
👍 - I'll go ahead and put together the `dwrote` PR in an evening soon, and push some performance tweaks here.

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,159 @@
...
+    font_collection: FontCollection,
+    font_fallback: Option<FontFallback>,
+}
+
+impl DirectWriteRasterizer {
+    fn rasterize_glyph(
+        &self,
+        font_face: &FontFace,
+        c: char,
+        size: Size,
```
chrisduerr: Super minor thing, but I'd swap these two parameters just because that would be a bit more logical and intuitive.

So `rasterize_glyph(font_face, size, c);`.

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,159 @@
...
+        let offset = GlyphOffset { advanceOffset: 0.0, ascenderOffset: 0.0 };
+        let em_size = em_size(size);
+
+        let glyph_run = DWRITE_GLYPH_RUN {
+            fontFace: unsafe { font_face.as_ptr() },
+            fontEmSize: em_size,
+            glyphCount: 1,
+            glyphIndices: &glyph_index,
+            glyphAdvances: &0.0,
+            glyphOffsets: &offset,
```
chrisduerr:
```suggestion
            glyphOffsets: &GlyphOffset::default(),
```

Think we can simplify things a bit by inlining this here. According to winapi docs, `Default` should be implemented.

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,159 @@
...
+
+    fn get_loaded_font(&self, font_key: FontKey) -> Result<&LoadedFont, Error> {
+        self.fonts.get(&font_key).ok_or(Error::FontNotLoaded)
+    }
+
+    fn get_glyph_index(&self, font_face: &FontFace, c: char) -> Result<u16, Error> {
+        font_face
+            .get_glyph_indices(&[c as u32])
+            .first()
+            .cloned()
```
chrisduerr: Why do we have to `cloned` here? Would it otherwise return a `&u16`? Might be a bit more conventional to just put a `*` before `font_face`? If you've had troubles with `filter`, a `**glyph_index != 0` double deref is sometimes necessary.

font/src/directwrite/mod.rs
```diff
@@ -107,73 +231,29 @@ impl crate::Rasterize for DirectWriteRasterizer {
...
-            // does not exist in the font
-            return Err(Error::MissingGlyph(glyph.c));
+        let loaded_font = self.get_loaded_font(glyph.font_key)?;
+
+        match self.rasterize_glyph(&loaded_font.font_face, glyph.c, glyph.size) {
+            Ok(glyph) => Ok(glyph),
+            Err(err @ Error::MissingGlyph(_)) => {
+                let fallback_font =
+                    self.get_fallback_font(&loaded_font.font, glyph.c).ok_or(err)?;
+                self.rasterize_glyph(&fallback_font.create_font_face(), glyph.c, glyph.size)
```
chrisduerr: This is also super minor, but it seems unfortunate that we *just barely* linebreak here.

I'd be fine with renaming this to `fallback` or just `font`. Though `fallback_font` is definitely the best variable name, so I don't mind that too much either.

font/src/directwrite/mod.rs
```diff
@@ -186,29 +266,79 @@ pub enum Error {
...
-}
+impl std::error::Error for Error {}
 
-impl ::std::fmt::Display for Error {
-    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
-        match *self {
-            Error::MissingGlyph(ref c) => write!(f, "Glyph not found for char {:?}", c),
-            Error::MissingFont(ref desc) => write!(
+impl std::fmt::Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> ::std::fmt::Result {
```
chrisduerr: We can move all the `fmt` stuff to imports here. There are no conflicts with anything as far as I can tell.

kchibisov:
@davidhewitt Could you ensure that your PR preserves style during fallback? It was mentioned there #3271 , I think it would be nice to match style of original glyph.

davidhewitt:
Just a ping to say I haven't forgotten about this; have been very busy!

chrisduerr:
@davidhewitt No worries, just let us know if you need any help.

davidhewitt:
> @davidhewitt Could you ensure that your PR preserves style during fallback? It was mentioned there #3271 , I think it would be nice to match style of original glyph.

I just checked this quickly and confirmed both red colour and bold are preserved. So I believe this is working.

chrisduerr:
> I just checked this quickly and confirmed both red colour and bold are preserved. So I believe this is working.

Fyi you can't do anything about color, but italics and bold are relevant. So the question is if a glyph that is not part of the main font, like '汉' can be stylized bold and italic.

davidhewitt:
Yep, is all good: 

image

chrisduerr:
Awesome, thanks for testing.

davidhewitt:
Rebased on master, included some simple perf optimizations (mostly around doing a bit less text copying of things like the locale) and addressed all style tweaks.

To reduce text copying I also had to adjust dwrote slightly with another PR there. https://github.com/servo/dwrote-rs/pull/41

This PR won't build until that dwrote PR is merged.

chrisduerr:
@davidhewitt Is this ready to go as far as you're concerned once that PR is merged?

If so, please request a review from me once that's done, then I can take another peek at the code.

davidhewitt:
Yes. (And if you want, just merge this pointing at my dwrote branch for now.)

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,170 @@
...
+        })
     }
 
     fn metrics(&self, key: FontKey, size: Size) -> Result<Metrics, Error> {
-        let font = self.fonts.get(key.token as usize).ok_or(Error::FontNotLoaded)?;
+        let font_face = &self.get_loaded_font(key)?.font_face;
+
+        // We assume that all monospace characters have the same width
+        // Because of this we take '!', the first drawable character, for measurements
+        let c = '!';
```
chrisduerr: Could we move this closer to where it's actually used?

font/src/directwrite/mod.rs
```diff
@@ -107,73 +242,32 @@ impl crate::Rasterize for DirectWriteRasterizer {
...
-        let face = font.create_font_face();
-        self.fonts.push(face);
+        let font_face = font.create_font_face();
+        let family_name = font.family_name();
+        let weight = font.weight();
+        let style = font.style();
+        let stretch = font.stretch();
+        let key = FontKey::next();
+        self.keys.insert(desc.clone(), key);
+        self.fonts.insert(key, LoadedFont { font_face, family_name, weight, style, stretch });
```
chrisduerr: All these variables are defined only to be used in the `LoadedFont` initialization. So they shouldn't be defined as separate variables but we should just inline these in the struct definition.

However since all we seem to be doing is querying fields on one thing to create another, this seems like a prime candidate for an `Into` implementation?

I'd assume there's a reason why we can't just use dwrote's `Font` ourselves, right? Otherwise that would be the obvious solution.

tinnick:
Hi, a random rust noob passing by. Is this issue still in progress? Is there anyway I can help here? I'm Japanese and not being able to render any Japanese characters unless I set the primary character to Japanese is quite annoying. Alacritty is awesome btw :)

kchibisov:
>Is this issue still in progress?

Yes it is, as you can see.

>Is there anyway I can help here?

You can test this PR.

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,170 @@
...
+use winapi::shared::ntdef::{HRESULT, LOCALE_NAME_MAX_LENGTH};
+use winapi::um::dwrite;
+use winapi::um::winnls::GetUserDefaultLocaleName;
 
 use super::{
     BitmapBuffer, FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph, Size, Slant, Style, Weight,
 };
 
+pub struct LoadedFont {
+    font_face: FontFace,
```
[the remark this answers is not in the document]
davidhewitt: Nice, will do that.

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,170 @@
...
+    family_name: String,
```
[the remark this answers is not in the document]
davidhewitt: I re-used `family_name` verbatim from `dwrote`, but I can change this to family if preferred.

font/src/directwrite/mod.rs
davidhewitt: I am caching all of these things because otherwise it creates a lot of additional directwrite API calls to retrieve this information every time a fallback font lookup is needed (all this information is input to the `map_characters` API call).

This was done in the name of performance.

In addition I think `font.create_font_face()` is likely to be relatively heavyweight so would prefer avoid doing that on every glyph rasterization, which we'd have to do if we just store `dwrote::Font`.

I really like the `Into` idea though.

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,170 @@
...
+}
+
 pub struct DirectWriteRasterizer {
-    fonts: Vec<dwrote::FontFace>,
+    fonts: HashMap<FontKey, LoadedFont>,
+    keys: HashMap<FontDesc, FontKey>,
     device_pixel_ratio: f32,
+    font_collection: FontCollection,
+    font_fallback: Option<FontFallback>,
+    locale: String,
```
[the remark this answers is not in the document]
davidhewitt: Yes, I am aware that was a drawback of caching it like this. I did this to remove the need to call this every time we search for a font fallback.

Let me see what happens to fallback timings if I revert just this bit...

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,170 @@
...
+        fallback_result.mapped_font
+    }
 }
 
 impl crate::Rasterize for DirectWriteRasterizer {
     type Err = Error;
 
     fn new(device_pixel_ratio: f32, _: bool) -> Result<DirectWriteRasterizer, Error> {
-        Ok(DirectWriteRasterizer { fonts: Vec::new(), device_pixel_ratio })
+        let locale = get_current_locale();
```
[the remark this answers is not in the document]
davidhewitt: 👍 I tend to lean towards field init shorthand for everything but that's just my habit and can undo that here!

font/src/directwrite/mod.rs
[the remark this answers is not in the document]
chrisduerr: No, that's a reasonable argument. Let's just keep things consistent with dwrote then.

font/src/directwrite/mod.rs
chrisduerr: That's fair enough, I'm not aware of how many of these methods are basically just a memory access versus an actual computation. `Into` would definitely be a good solution either way.

davidhewitt: A lot of the `dwrote` methods involving `String` have at least one additional conversion going on to convert back and forth from UTF-16 C++ encoding to Rust `String`. Based on experiments I've done, caching all these saves us a number of microseconds, which counts for a lot when the comment above was asking me to try to make perf ~ 20-30 us per fallback lookup!

Will go down the `Into` road.

chrisduerr: Great job actually looking into this, I'll definitely gladly take those free performance improvements.

font/src/directwrite/mod.rs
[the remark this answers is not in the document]
davidhewitt: Ok, I've reverted just this bit. Btw it looks like `dwrote` also caches this thing:

https://github.com/servo/dwrote-rs/blob/c2486e8db9ef1a6fb51b126ead34d392a5060d82/src/helpers.rs#L14-L21

But that's not necessarily a reason for us to be naughty too. Caching this one thing barely affects performance.

davidhewitt:
> Is this issue still in progress?

It's very nearly complete - we're just waiting for https://github.com/servo/dwrote-rs/pull/41 to be merged and released, which allows some performance optimisations I'm doing here.

davidhewitt:
The dwrote PR got merged so I have rebased this PR and updated the `dwrote` git commit referenced.

Fingers crossed CI is green. It's then your call if we want to merge now or request `dwrote` releases a new version to use instead of a git dependency.

chrisduerr:
> It's then your call if we want to merge now or request dwrote releases a new version to use instead of a git dependency.

I'd like to stick to actual releases. Could you maybe talk to the dwrote people and at least say that we have interest in a release? It doesn't matter when, but I think it's good to share that others are waiting.

davidhewitt:
👍 see https://github.com/servo/dwrote-rs/issues/44

davidhewitt:
`dwrote 0.11` is now released, so I've updated this PR and rebased on master.

CHANGELOG.md
```diff
@@ -33,6 +33,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Live config reload for `window.title`
 
+### Added
+
+- Font fallback on Windows
```
kchibisov: Rebase error, should be in `0.5.0-dev`.

damnskippy:
Just out of curiosity, any chance this can or will be included in the upcoming release? 
Thanks.

chrisduerr:
> Just out of curiosity, any chance this can or will be included in the upcoming release?

The 0.4.2 release will no receive any further updates other than bugfixes for regressions.

davidhewitt:
> Rebase error, should be in 0.5.0-dev.

Good catch, fixed. Also squashed all commits.

chrisduerr:
@davidhewitt I'll try to get to this as soon as my search effort is done. I have not forgotten about this, just to let you know.

codecat:
Seems like there's still some issues with glyph alignment/scaling, at least on the latest commit from this pull request. (Unsure if I should be reporting this though, might already be known. Just figured I'd give some hopefully helpful feedback!)

image

chrisduerr:
> Seems like there's still some issues with glyph alignment/scaling, at least on the latest commit from this pull request.

I don't think there's any plan to fix this if I remember the discussion in this PR correctly. We don't scale fonts on any other platform so it doesn't make sense to start that on Windows.

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,168 @@
...
 use super::{
     BitmapBuffer, FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph, Size, Slant, Style, Weight,
 };
 
+/// This struct contains the properties from dwrote::Font needed in the rest of this module.
+///
+/// We use this struct rather than dwrote::Font for performance reasons: the getter methods on
+/// dwrote::Font need to do things like String conversion from C++ widestrings to Rust's
+/// UTF-8 strings.
+pub struct Font {
```
chrisduerr, line 40: Why is this public?

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,168 @@
...
 
 use super::{
     BitmapBuffer, FontDesc, FontKey, GlyphKey, Metrics, RasterizedGlyph, Size, Slant, Style, Weight,
 };
 
+/// This struct contains the properties from dwrote::Font needed in the rest of this module.
+///
+/// We use this struct rather than dwrote::Font for performance reasons: the getter methods on
+/// dwrote::Font need to do things like String conversion from C++ widestrings to Rust's
+/// UTF-8 strings.
```
chrisduerr, line 39: This documentation isn't really great from a user perspective. It's extremely verbose and contains a lot of information that isn't really necessary or useful.

```suggestion
/// Cached DirectWrite font.
```

I think this might even be sufficient already. You know that it tries to prevent overhead with loading because it's a cache and it says what it is. No need to know anything about C++ widestring conversions.

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,168 @@
...
+    style: FontStyle,
+    stretch: FontStretch,
+}
+
 pub struct DirectWriteRasterizer {
-    fonts: Vec<dwrote::FontFace>,
+    fonts: HashMap<FontKey, Font>,
+    keys: HashMap<FontDesc, FontKey>,
     device_pixel_ratio: f32,
+    collection: FontCollection,
```
chrisduerr, line 52: This is the font fallback list, correct? I think `collection` might not be ideal naming since a collection might be all kinds of things. On Linux this is called `fallback_lists` for example (since every style has a different fallback list).

Or is this a collection of all fonts present on the system?

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,168 @@
...
+    fallback: Option<FontFallback>,
```
chrisduerr, line 53: The naming for this is a bit awkward too. Maybe there's a name that would better describe it than just 'fallback'?

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,168 @@
...
+    fn get_loaded_font(&self, font_key: FontKey) -> Result<&Font, Error> {
+        self.fonts.get(&font_key).ok_or(Error::FontNotLoaded)
+    }
+
+    fn get_glyph_index(&self, face: &FontFace, c: char) -> Result<u16, Error> {
+        let idx = *face
+            .get_glyph_indices(&[c as u32])
+            .first()
+            // DirectWrite returns 0 if the glyph does not exist in the font
+            .filter(|glyph_index| **glyph_index != 0)
```
chrisduerr, line 122: Is there any way for this to return `0` as the first element of the indices, but then continue with a usable non-zero value later?

font/src/directwrite/mod.rs
```diff
@@ -13,30 +13,168 @@
...
+
+        let length = c.len_utf16() as u32;
+        let utf16_codepoints = &buf[..length as usize];
+
+        let locale = get_current_locale();
+
+        let text_analysis_source_data = TextAnalysisSourceData { locale: &locale, length };
+        let text_analysis_source = TextAnalysisSource::from_text(
+            Box::new(text_analysis_source_data),
+            Cow::Borrowed(&utf16_codepoints),
```
chrisduerr, line 142:
```suggestion
            Cow::Borrowed(utf16_codepoints),
```

I don't think this reference is necessary?

font/src/directwrite/mod.rs
```diff
@@ -51,9 +189,12 @@ impl crate::Rasterize for DirectWriteRasterizer {
         let line_height = f64::from(ascent - descent + line_gap);
 
         // We assume that all monospace characters have the same width
-        // Because of this we take '!', the first drawable character, for measurements
-        let glyph_metrics = font.get_design_glyph_metrics(&[33], false);
-        let hmetrics = glyph_metrics.first().ok_or(Error::MissingGlyph('!'))?;
+        // Because of this we take '!', the first drawable character, for horizontal metrics
```
chrisduerr, line 192:
```
        // Since all monospace characters have the same width, we use `!` for horizontal metrics
```

I think this works a bit better, even though it has only slightly been changed in this PR. That way we don't have two sentences that aren't properly delimited with punctuation but still separate sentences.

font/src/directwrite/mod.rs
```diff
@@ -69,27 +210,22 @@ impl crate::Rasterize for DirectWriteRasterizer {
     }
 
     fn load_font(&mut self, desc: &FontDesc, _size: Size) -> Result<FontKey, Error> {
-        let system_fc = FontCollection::system();
+        // Fast path if face already loaded
```
chrisduerr, line 213:
```suggestion
        // Fast path if face is already loaded
```

font/src/directwrite/mod.rs
```diff
@@ -107,73 +243,27 @@ impl crate::Rasterize for DirectWriteRasterizer {
...
-            return Err(Error::MissingGlyph(glyph.c));
+        let loaded_font = self.get_loaded_font(glyph.font_key)?;
+
+        match self.rasterize_glyph(&loaded_font.face, glyph.size, glyph.c) {
+            Ok(glyph) => Ok(glyph),
+            Err(err @ Error::MissingGlyph(_)) => {
+                let fallback_font = self.get_fallback_font(&loaded_font, glyph.c).ok_or(err)?;
+                self.rasterize_glyph(&fallback_font.create_font_face(), glyph.size, glyph.c)
+            }
+            Err(err) => Err(err),
```
chrisduerr, line 265:
```suggestion
            Err(err @ Error::MissingGlyph(_)) => {
                let fallback_font = self.get_fallback_font(&loaded_font, glyph.c).ok_or(err)?;
                self.rasterize_glyph(&fallback_font.create_font_face(), glyph.size, glyph.c)
            },
            result => result,
```

I'm not sure if there isn't an even better way to do this, but I feel like this makes things at least a little nicer.

font/src/directwrite/mod.rs
```diff
@@ -186,29 +276,85 @@ pub enum Error {
...
+            Slant::Normal => FontStyle::Normal,
         }
     }
 }
+
+fn get_current_locale() -> String {
+    let mut buf = vec![0u16; LOCALE_NAME_MAX_LENGTH];
+    let len = unsafe { GetUserDefaultLocaleName(buf.as_mut_ptr(), buf.len() as i32) as usize };
+
+    // len includes null byte, we don't need that in Rust strings
```
chrisduerr, line 341:
```suggestion
    // `len` includes null byte, which we don't need in Rust
```

font/src/directwrite/mod.rs
```diff
@@ -186,29 +276,85 @@ pub enum Error {
...
+    OsString::from_wide(&buf[..len - 1]).into_string().expect("Locale not valid unicode")
+}
+
+/// Information needed by dwrote's TextAnalysisSource to find a fallback font
+/// for a run of text. See `get_fallback_font`.
```
chrisduerr, line 346: Usually the first line of documentation on an item should be a short description of what it is. I feel like if that doesn't even fit on a single line, it's probably too verbose.

```suggestion
/// Font fallback information for dwrote's TextAnalysisSource.
```

Do you think this correctly represents the purpose of this struct, or is there some information missing?

font/src/directwrite/mod.rs
davidhewitt: Doesn't need to be; will change.

font/src/directwrite/mod.rs
davidhewitt: This is the collection of fonts the use. Usually the system collection - and that's what's loaded in `::new()`.

(The dwrote `FontCollection` is a wrapper type around the [DirectWrite interface](https://docs.microsoft.com/en-gb/windows/win32/api/dwrite/nn-dwrite-idwritefontcollection?redirectedfrom=MSDN))

This used to be called `font_collection` but we took the prefix `font_` off most names in the file.

`dwrite_collection` ?

font/src/directwrite/mod.rs
davidhewitt: Similar to above, the name `fallback` leads directly from the directwrite `FontFallback` concept.

`fallback_sequences` ? (Inspired by the description of [FontFallback at Microsoft](https://docs.microsoft.com/en-gb/windows/win32/api/dwrite_2/nn-dwrite_2-idwritefontfallback))

font/src/directwrite/mod.rs
davidhewitt: As per the comment, which comes from the [remarks on GetGlyphIndices](https://docs.microsoft.com/en-gb/windows/win32/api/dwrite/nf-dwrite-idwritefontface-getglyphindices#remarks) `0` is the value returned when the glyph is not in the font, so that's definitely the error condition.

If you're concerned about the length - `get_glyph_indices()` will always return a `Vec` of length equal to the input slice - so this vec will always have only one element in it.

Perhaps you prefer this alternative implementation:

```
        match face.get_glyph_indices(&[c as u32]).as_slice() {
            [0] => Err(Error::MissingGlyph(c)),
            [i] => Ok(*i),

            // get_glyph_indices will always return a Vec with length equal to the input slice
            _ => unreachable!()
        }
```

davidhewitt:
Thanks, have applied suggestions and rebased. Couple of bits of input needed from you to decide how to finish things off.

font/src/directwrite/mod.rs
chrisduerr: Usually when I think collection, I'm directed more torwards `std` than directwrite.

I think `available_fonts` would be a good choice, it seems to correctly represent what is stored.

font/src/directwrite/mod.rs
chrisduerr: Seems like it would be `fallback_sequence`, without the `s`?

The word `sequence` seems almost a bit hard to grasp for me, but I think it's a decent choice that should work just fine.

font/src/directwrite/mod.rs
chrisduerr: I'm fine with the implementation, I was just concerned if it might be necessary to check for later elements in the vec.

davidhewitt:
That's done 👍 

As there's a lot of commits on this branch, probably worth squashing as part of the merge.

madetara:
I just want to thank you all for making it!
Can't wait for it to be released

Thank you!

s-taylor:
I'd also love for this to be released, i spent hours puzzling why my powerline icons were not rendering thinking it was something wrong with my fonts. I'll see if i can try build it manually from source but that's likely painful on windows.

jsinkers:
I can confirm building from source is not too difficult on Windows and
worth the effort. It seems like fixing font fallback would be a minor
change but I found this to be a significant improvement.

s-taylor:
For anyone struggling with this, I was able to resolve it with these steps

* Build Alacritty from source (the steps to this were not clear, but I figured it out, [see notes here](https://gist.github.com/s-taylor/5aa53297cda0a7dd7c3539bcbf952a06))
* Install fira code via choco `choco install firacode`
* Set fira code as my font in the alacritty.yml and set `use_thin_strokes: true`
* Zoom in slightly

I feel like the other font installation methods I was using were not working correctly. Suspect there is more to installing fonts in windows than just double clicking on the font files and choosing install. But I did also try other installation methods such as [this](https://medium.com/@slmeng/how-to-install-powerline-fonts-in-windows-b2eedecace58) which did not work either. Installing fira code through choco seems to have done the magic.

PR #3130, opened 2019-12-29
Fix startup time on Wayland in some cases

kchibisov:


chrisduerr:
Are you aware of why swapping buffers causes issues here? And is there any problem with 'garbage' data filling the space Alacritty occupies before startup?

alacritty/src/display.rs
```diff
@@ -208,7 +210,9 @@ impl Display {
 
         // We should call `clear` when window is offscreen, so when `window.show()` happens it
         // would be with background color instead of uninitialized surface.
-        window.swap_buffers();
+        if event_loop.is_x11() {
```
chrisduerr: Checking for `is_x11()` will probably cause a build failure on non-linux/bsd systems. We need to put this behind a #[cfg(not(any(target_os = "macos", windows)))]`.

kchibisov: It's a draft

chrisduerr: Just pointing it out since I noticed it anyways. :)

kchibisov:
>Are you aware of why swapping buffers causes issues here? And is there any problem with 'garbage' data filling the space Alacritty occupies before startup?

I've just found that removing this swap_buffers actually resolve the startup issue at least on my system, so I should go through Wayland logs, but the issue is in waiting on swap_buffers for a long. However we're not even waiting for 200ms seconds here, but in `fn draw`, the most interesting thing that removing this call actually affects the result and resolves issue on my system. So I want to gather some feedback whether it actually do something for other folks.

Oh, and the window isn't presented until you actually draw something into it, so the call is useless on Wayland tbh.

chrisduerr:
> I've just found that removing this swap_buffers actually resolve the startup issue at least on my system, so I should go through Wayland logs, but the issue is in waiting on swap_buffers for a long. 

Does winit do anything special when this is called? Like query monitor information or something like that?

kchibisov:
During swap_buffers? Nope it just calls to EGL as I can see, maybe it's doing something, but it's not the first call of swap_buffers when it actually blocks for a while. It's like 2-3d one.

tadeokondrak:
To my understanding, eglSwapBuffers() on wayland blocks the thread until the compositor requests a redraw, if vsync is on:
https://github.com/jwilm/alacritty/blob/474032742b3bf325e655eb6cddb19632cb989c1a/alacritty/src/window.rs#L124
I *think* the right way to do it on wayland is to disable vsync, then render on the frame callback yourself manually, which winit exposes as RedrawRequested.

alacritty/src/display.rs
```diff
@@ -207,8 +209,19 @@ impl Display {
         });
 
         // We should call `clear` when window is offscreen, so when `window.show()` happens it
-        // would be with background color instead of uninitialized surface.
-        window.swap_buffers();
+        // would be with background color instead of uninitialized surface. On Wayland we can
+        // safely ignore this call, since the window isn't visible until you actually draw
+        // something into it.
```
chrisduerr: Let's move the `On Wayland [...]` part to the `if event_loop.is_x11()`, since that reduces the information density of this comment a bit and it's not relevant to the other codepaths anyways.

alacritty/src/display.rs
```diff
@@ -207,8 +209,19 @@ impl Display {
...
+        #[cfg(not(any(target_os = "macos", windows)))]
+        {
+            if event_loop.is_x11() {
+                window.swap_buffers()
+            }
+        }
+        #[cfg(any(target_os = "macos", windows))]
+        {
+            window.swap_buffers()
+        }
```
chrisduerr: Honestly I don't think we need this on macOS/Windows at all, right? I can test it on macOS to make sure, but originally this was just introduced for X11.

kchibisov:
>To my understanding, eglSwapBuffers() on wayland blocks the thread until the compositor requests a redraw, if vsync is on:

First of all, the issue is strange, we were doing the same things before, but now if you call swap_buffers 3 times in a row it'll block for ~200ms without any reason for it. So it's just a happened that things are working with lower number of swap_buffer calls. So to solve an issue we can pass `sway -Dtxn-timeout=20` (the default is 200ms). Alacritty doesn't start like 200ms, so I don't understand how to we even reach it now. So this patch is more a workaround then a real solution, but lets just go with it for now, since I can't repro the issue on a normal use anyway.

>I think the right way to do it on wayland is to disable vsync, then render on the frame callback yourself manually, which winit exposes as RedrawRequested.

Yes, but we can't do it in alacritty now, since we're not a Wayland only terminal and it'll add a maintenance burden.

alacritty/src/display.rs
kchibisov: Ehm, they also have a concept of window before drawing, so you should just rely, that your compositor doing something that make sense. I feel like we should just leave it as it is.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Deadlock on Windows when closing Alacritty using the title bar "X" button (ConPTY backend)
 - Crash on `clear` when scrolled up in history
 - Entire screen getting underlined/stroke out when running `clear`
+- Extremely slow startup on some Wayland compositors
```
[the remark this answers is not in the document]
kchibisov: It also happened on wayfire, so basically it's all wlroots compositors, I think.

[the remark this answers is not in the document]
chrisduerr: Gotcha, that's a really good point. Should we maybe say `[...] on wlroots Wayland compositors`?

But I feel like some people might not know what that means, so at that point it's probably better to just say `some Wayland compositors` I think.

alacritty/src/display.rs
chrisduerr: Not all platforms require that you manually clear the window yourself though. Some platforms might make it invisible, which works fine. On macOS, it works great for example and I don't think there has ever been an issue with that.

So why should we potentially delay the startup significantly (remember swap_buffers might block), when it has no benefit?

chrisduerr:
Great addition for the 0.4.1 release, thanks!

PR #3135, opened 2019-12-30
Remove C1 escape support

### LINKED ISSUE #2981: Terminal print `6c6c6c` into the prompt after executing `cat`
# System
|                  |                               |
|------------------|-------------------------------|
| Operating System | Linux 4.19.81-1-MANJARO     |
| Rust Version     | 1.38.0   |
| Display Server   | X11 |
| Window Manager   | i3 |
| Compositor       | compton |

# Summary
Alacritty v0.3.3 print `6c6c6c6c6c` characters into the prompt after executing `cat` or similar program.
This happen everytime on a few files.
Number of `6c` printed depends on the amount of file's content printed (more content = more `6c`).
Seems like alacritty issue, because:
- both ZSH and BASH print this characters in alacritty
- terminator and gnome-terminal works fine (no `6c6c6c` characters)

# Behavior
( `|` stands for cursor position)
```
user-pc$ cat file
Lorem ipsum
...
sit amet.
user-pc$ 6c6c6c6c6c|
```

# Expectation
( `|` stands for cursor position)
```
user-pc$ cat file
Lorem ipsum
...
sit amet.
user-pc$ |
```

# Extra
An example of a file having this issue:
[sample.zip](https://github.com/jwilm/alacritty/files/3848102/sample.zip)

chrisduerr:
Fixes #2981.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
-- Support for non-utf8 C1 escapes 
+- Support for 8-bit C1 escape sequences
 
 ## 0.4.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Crash on `clear` when scrolled up in history
 - Entire screen getting underlined/stroke out when running `clear`
 
+### Removed
+
+- Support for non-utf8 C1 escapes 
+
 ## 0.4.0
 
 ### Packaging
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -719,9 +719,6 @@ where
             C0::SUB => self.handler.substitute(),
             C0::SI => self.handler.set_active_charset(CharsetIndex::G0),
             C0::SO => self.handler.set_active_charset(CharsetIndex::G1),
-            C1::NEL => self.handler.newline(),
-            C1::HTS => self.handler.set_horizontal_tabstop(),
-            C1::DECID => self.handler.identify_terminal(self.writer),
             _ => debug!("[unhandled] execute byte={:02x}", byte),
         }
     }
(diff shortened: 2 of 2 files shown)


chrisduerr:
I've looked at the `vte` crate and removing C1 support from there doesn't really seem like it would improve anything, since it just removes three lines in `table.rs` basically which isn't very expensive to maintain for people that are interested.

So I've chosen to just remove it from Alacritty, since no other terminal emulator supports this by default, so it's basically just dead code that leads to surprises with users (as seen in #2981).

CHANGELOG.md
```diff
@@ -36,6 +36,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Crash on `clear` when scrolled up in history
 - Entire screen getting underlined/stroke out when running `clear`
 
+### Removed
+
+- Support for non-utf8 C1 escapes
```
cole-h: `non-utf8` -> `non-UTF-8`?

Our other changelog entries mentioning UTF-8 do so in the proper case ;-)

If that looks as strange to you as it does to me, maybe change the entire message to `- Support for 8-bit C1 escapes`? (I think 8-bit is the proper name after briefly skimming the formerly-mentioned issue)

kchibisov: technically speaking utf8 and UTF-8 are both fine and in use, so the current changelog message isn't controversial in my opinion and reflects the naming for similar things like mime types (e.g. `text/plain;charset=utf-8`).

cole-h: I'm fine either way, but I figured consistency is key.

chrisduerr: Yeah I chose the lowercase naming because `non-UTF-8` looks strange.

In theory `8-bit` is accurate too, but we've never had 8-bit UTF-8 escapes, so I wanted to choose the word that's more specific here.

nixpulvis:
The only control I might be worried about application support for is the `NEL` C1 encoding. The others seem even more esoteric. I personally haven't seen any applications render improperly because of this, but I don't know what applications are all out there.

You can see the difference easily with `printf '\x85'`.

nixpulvis:
The changelog also makes it sound like proper unicode encodings of the C1 controls should still work, but `printf '\xc2\x85'` doesn't seem to work as I'd expect in this PR. I'm expecting a `NEL` newline.

chrisduerr:
> The only control I might be worried about application support for is the NEL C1 encoding.

Why, because of the wikipedia statement about it? Neither XTerm, VTE, nor URxvt support it by default.

> The changelog also makes it sound like proper unicode encodings of the C1 controls should still work, but printf '\xc2\x85' doesn't seem to work as I'd expect in this PR.

Unicode encodings for the C1 controls never worked, so there's no change wrt that. That's why I specifically mentioned it.

Though maybe being less precise is more clear here, just saying `Support for C1 escape sequences` has been removed might tell everyone everything they need to know, because it doesn't matter that a previous version hasn't supported something either.

nixpulvis:
> Though maybe being less precise is more clear here, just saying Support for C1 escape sequences has been removed might tell everyone everything they need to know, because it doesn't matter that a previous version hasn't supported something either.

Yea I think this would clear up any potential confusion.

chrisduerr:
> Yea I think this would clear up any potential confusion.

I've changed it, hopefully it should be more clear about what the state is **after** this patch, which really is the only thing that matters.

PR #3137, opened 2019-12-31
Replace deprecated Error methods

sterlingjensen:
- Display::fmt instead of Error::description
- Error::source instead of Error::cause
- Adjacent linting opportunities performed.

### DIFF
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -1,5 +1,6 @@
 use std::env;
 use std::fmt::{self, Display, Formatter};
+use std::fs;
 use std::io;
 use std::path::PathBuf;
 
@@ -21,7 +22,6 @@ pub use crate::config::bindings::{Action, Binding, Key, RelaxedEq};
 #[cfg(test)]
 pub use crate::config::mouse::{ClickHandler, Mouse};
 use crate::config::ui_config::UIConfig;
-use std::fs::read_to_string;
 
 pub type Config = TermConfig<UIConfig>;
 
@@ -147,7 +147,7 @@ pub fn reload_from(path: &PathBuf) -> Result<Config> {
 }
 
 fn read_config(path: &PathBuf) -> Result<Config> {
-    let mut contents = read_to_string(path)?;
+    let mut contents = fs::read_to_string(path)?;
 
     // Remove UTF-8 BOM
     if contents.chars().nth(0) == Some('\u{FEFF}') {
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -24,7 +24,7 @@ use crate::config::ui_config::UIConfig;
 pub type Config = TermConfig<UIConfig>;
 
 /// Result from config loading
-pub type Result<T> = ::std::result::Result<T, Error>;
+pub type Result<T> = std::result::Result<T, Error>;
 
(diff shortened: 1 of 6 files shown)


alacritty/src/config/mod.rs
```diff
@@ -43,29 +43,20 @@ pub enum Error {
...
-        }
-    }
-
-    fn description(&self) -> &str {
-        match *self {
-            Error::NotFound => "Couldn't locate config file",
-            Error::ReadingEnvHome(ref err) => err.description(),
-            Error::Io(ref err) => err.description(),
-            Error::Yaml(ref err) => err.description(),
+            Error::ReadingEnvHome(e) => e.source(),
```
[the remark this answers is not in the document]
kchibisov: However I feel like it's just a style and I'd like to wait for @chrisduerr on this thing here. In our code base we do use `e` for such things, but way, way less then `err` as I can see.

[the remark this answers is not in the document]
sterlingjensen: I'd agree with you if it was outside of a match - where I'd drop a lot more than just two characters if I could. I always leave edits from maintainers enabled, and I'd hope that would be exercised in a case where a non-functional style change holds up a pull. Or I can change it, either or - just trying to get rid of the build noise.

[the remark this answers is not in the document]
chrisduerr: If it's just a single variable and line length is not a problem, I think the longer the better.

With a short statement like this, even `error` would work perfectly fine, so that would resolve the issue of having to find the optimal abbreviation.

In general, `err` is the shortest 'acceptable' abbreviation usually though. Because `e` is commonly used for `event` too (`ev` would be better here!), so it's better not to have any reasons for confusion. Especially something like `e.source()` would make it impossible for me to tell if we're talking about an event or an error here, without looking at surrounding code.

sterlingjensen:
I did what I could with the paths, though not much can be done in the touched files.

alacritty/src/config/mod.rs
```diff
@@ -154,7 +147,7 @@ pub fn reload_from(path: &PathBuf) -> Result<Config> {
 }
 
 fn read_config(path: &PathBuf) -> Result<Config> {
-    let mut contents = std::fs::read_to_string(path)?;
+    let mut contents = read_to_string(path)?;
```
chrisduerr: Functions that aren't in this file usually should have like one level of path to it so we know where things are coming from.

In this case it would be `fs::read_to_string`. That's usually the best middle-ground for functions and at least in Alacritty it should be the most common thing.

alacritty/src/window.rs
```diff
@@ -307,7 +299,7 @@ impl Window {
     }
 
     #[cfg(windows)]
-    pub fn set_urgent(&self, _is_urgent: bool) {}
+    pub const fn set_urgent(&self, _is_urgent: bool) {}
```
chrisduerr: What's the purpose of this change?

sterlingjensen: Any function that I can declare `const`, I do. But I just dumped the IR and can see that rustc isn't picking up the opportunity to inline the void return. I'll revert it.

chrisduerr: It should be optimized out entirely regardless.

Making this `const` is just confusing and unnecessary extra text.

PR #3141, opened 2020-01-02
Update winit to 0.20.0-alpha5

gin-ahirsch:
I decided to do `cargo update` today before building and ran into some compile-errors due to a winit update. This PR fixes those.

I get a new warning `use of deprecated item 'winit::event::KeyboardInput::modifiers': Deprecated in favor of DeviceEvent::ModifiersChanged`, though rust-windowing/winit#1151 is still open (for iOS and WASM) so I think it's good to keep using that field for now.


chrisduerr:
See https://github.com/jwilm/alacritty/pull/3133.

Issue #889, opened 2017-11-06
Graphical artifacts with window transparency


mkeeler:
Which operating system does the issue occur on? MacOS (10.12.6 - Intel Iris Pro graphcis)

So when I have window transparency turned on (background_opacity: 0.85) I sometimes see graphical artifacts where text is rendered in the background of what was previously on my terminal. An image is attached. This occurs both within tmux and outside of tmux after detaching.

I cant seem to figure out what makes it occur or why its displaying a particular text but its pretty reproducible for me.

<img width="368" alt="artifact" src="https://user-images.githubusercontent.com/1641976/32444417-8bd8bc1c-c2d0-11e7-98eb-d8b00868ae13.png">

mafredri:
Same issue here, I recorded a screencast demoing the issue:

![alacritty-artifacts2](https://user-images.githubusercontent.com/147409/39093849-cc556e0c-462e-11e8-95f2-4ce1f1edcfda.gif)

(As you might notice, it's easily repeatable when hiding Alacritty via Cmd+H)

epwalsh:
Same issue here on a MacBook Air running OS X 10.13.5.

stephenprater:
Same issue on Macbook Pro 10.12.16

![image](https://user-images.githubusercontent.com/149870/43061704-132eef78-8e0b-11e8-965b-c9dbd3123e76.png)

You can't actually see the background effect in this screen shot - _only_ the ghosting effect.  I had just closed a split where the lighter part is.

kirbpowell:
Seeing the same issue here with `background_opacity: 0.5`, running on Macbook Pro 10.14.2 with Radeon Pro 560X graphics.

tomaskallup:
The problem is window shadow, I had the same problem with iTerm2 borderless patch and removing shadow fixed the issue. Would it be possible to add an option to remove window shadow?

remi-gelinas:
Same issue here with `background_opacity: 0.3` on a 2018 Macbook Pro with Iris Pro graphics.

amagura:
I'm having this issue with `background_opacity: 0.6` on a 2018 MacBook Pro (15).

vaibhav-y:
Re: #2490, these are the steps to reproduce on macOS Mojave 10.14.5
1. Enable window trasparency
2. Resize an alacritty window, note the position of any existing text on the window at the maximum size scaled to by dragging the corner
[...]

PR #3142, opened 2020-01-02
Fix macOS OpenGL context resize ordering

### LINKED ISSUE #2226: Padding does not take effect until I start typing

macOS 10.14.4

I have Alacritty's window padding settings set to the default of:

```
  padding:
    x: 10
    y: 10
```

When I open a new Alacritty window, there is 0 padding, but when I begin to type the padding setting goes into effect and is set to `10px`.

I've tried removing my config file and uninstalling and reinstalling Alacritty, but it still does this even with a fresh install.

chrisduerr:
Fixes #2226.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Crash on `clear` when scrolled up in history
 - Entire screen getting underlined/stroke out when running `clear`
 - Slow startup on some Wayland compositors
+- Padding not consistently visible on macOS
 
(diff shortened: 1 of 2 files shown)


kchibisov:
Could you also test #889 . IIRC it was also related to resize, maybe it do something about it.

kchibisov:
Seems like it's working fine on X11/Wayland on my system.

chrisduerr:
> Could you also test #889 . IIRC it was also related to resize, maybe it'll do something about it.

That's a great idea, I'll get on that later.

chrisduerr:
There's no change to #889 as far as I can tell.

PR #3155, opened 2020-01-05
Fix incorrect config path in --help

### LINKED ISSUE #3154: Alacritty reporting wrong config file location on Windows
On Windows, the configuration file path reported by `alacritty --help` is wrong. I'd expect to see this:

```
OPTIONS:
        --class <class>                            Defines window class on Linux [default: Alacritty]
    -e, --command <command>...                     Command and args to execute (must be last argument)
        --config-file <config-file>
            Specify alternative configuration file [default: %APPDATA%\alacritty\alacritty.yml]
```

But instead I see this:

```
OPTIONS:
        --class <class>                            Defines window class on Linux [default: Alacritty]
    -e, --command <command>...                     Command and args to execute (must be last argument)
        --config-file <config-file>
            Specify alternative configuration file [default: $XDG_CONFIG_HOME/alacritty/alacritty.yml]
```

### System

OS: Windows
Version: `alacritty 0.4.0 (b115b90)`
Windows: WinPTY

chrisduerr:
Fixes #3154.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## 0.4.2-dev
 
+### Fixed
+
+- Incorrect default config path in `--help` on Windows and macOS
+
 ## 0.4.1
 
 ### Packaging
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -28,7 +28,7 @@ const CONFIG_PATH: &str = "$XDG_CONFIG_HOME/alacritty/alacritty.yml";
 #[cfg(windows)]
 const CONFIG_PATH: &str = "%APPDATA%\\alacritty\\alacritty.yml";
 #[cfg(target_os = "macos")]
-const CONFIG_PATH: &str = "~/.config/alacritty/alacritty.yml";
+const CONFIG_PATH: &str = "$HOME/.config/alacritty/alacritty.yml";
 
 /// Options specified on the command line
 pub struct Options {
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -23,6 +23,13 @@ use alacritty_terminal::index::{Column, Line};
 
 use crate::config::Config;
 
+#[cfg(not(any(target_os = "macos", windows)))]
+const CONFIG_PATH: &str = "$XDG_CONFIG_HOME/alacritty/alacritty.yml";
+#[cfg(windows)]
+const CONFIG_PATH: &str = "%APPDATA%\\alacritty\\alacritty.yml";
+#[cfg(target_os = "macos")]
+const CONFIG_PATH: &str = "~/.config/alacritty/alacritty.yml";
+
 /// Options specified on the command line
 pub struct Options {
     pub live_config_reload: Option<bool>,
@@ -158,8 +165,7 @@ impl Options {
                     .help("Start the shell in the specified working directory"),
             )
             .arg(Arg::with_name("config-file").long("config-file").takes_value(true).help(
-                "Specify alternative configuration file [default: \
-                 $XDG_CONFIG_HOME/alacritty/alacritty.yml]",
+                &format!("Specify alternative configuration file [default: {}]", CONFIG_PATH)
             ))
             .arg(
                 Arg::with_name("command")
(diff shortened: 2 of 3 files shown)


alacritty/src/cli.rs
```diff
@@ -23,6 +23,13 @@ use alacritty_terminal::index::{Column, Line};
 
 use crate::config::Config;
 
+#[cfg(not(any(target_os = "macos", windows)))]
+const CONFIG_PATH: &str = "$XDG_CONFIG_HOME/alacritty/alacritty.yml";
+#[cfg(windows)]
+const CONFIG_PATH: &str = "%APPDATA%\\alacritty\\alacritty.yml";
+#[cfg(target_os = "macos")]
+const CONFIG_PATH: &str = "~/.config/alacritty/alacritty.yml";
```
nixpulvis: https://github.com/jwilm/alacritty/issues/3154#issuecomment-570956274

kchibisov: can we use $HOME here for some kind of consistency?

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## 0.4.2-dev
 
+### Fixed
+
+- Incorrect default config path in `--help` on Windows and macOS
```
kchibisov: Can we get this patch into v0.4.1?

chrisduerr: Honestly I think I'd like to put this into 0.4.2, since this isn't really a bug that is very time sensitive and it has been around for a long, long time anyways. So extending the scope of the 0.4.1 release for this after the RC is already out seem strange.

If there were other bugs that required another RC anyways, I wouldn't mind as much, but I think the RC is actually looking pretty good as-is.

chrisduerr: Thoughts @nixpulvis?

chrisduerr:
> I really feel like the definition of CONFIG_PATH should at least be near [...] if not unified in some way.

I don't think it's a good idea to move a constant to a different file, where it is never used.

Unification is also not possible because these are the primary default locations, not the absolute locations. Additionally, environment variables mean nothing to Alacritty when just present in a String.

So I think it's best to keep it simple and just move the constant to the top of the file where it is used. If we ever used it anywhere else, that would be a different story, but we don't, since this is not the actual config location, but just the documentation string for the first config location searched.

CHANGELOG.md
nixpulvis: I don't have a strong opinion either way. I think it's fine to wait, I'm generally a fan of keeping releases contained.

kchibisov:
btw we do have the same problem for man page for macOS, what do you think about being a bit more verbose an just put every path into man page? It's a man page in the end of the day...

chrisduerr:
> btw we do have the same problem for man page for macOS, what do you think about being a bit more verbose an just put every path into man page? It's a man page in the end of the day...

I've mentioned the manpage in the issue, but I haven't actually thought about the idea of just putting everything in it. I think that's a great idea, let me see if I can figure out how manpages work to extend that section a bit.

chrisduerr:
I've extended the manpage to include what is in our readme already. I think this is a great bit of additional information:

tmp

extra/alacritty.man
```diff
@@ -49,7 +49,15 @@ Defines the window class hint on Linux [default: Alacritty,Alacritty ]
...
-Specify alternative configuration file [default: $XDG_CONFIG_HOME/alacritty/alacritty.yml]
+Specify alternative configuration file
+
+Alacritty looks for the configuration file at the following paths:
+    1. $XDG_CONFIG_HOME/alacritty/alacritty.yml
+    2. $XDG_CONFIG_HOME/alacritty.yml
+    3. $HOME/.config/alacritty/alacritty.yml
+    4. $HOME/.alacritty.yml
+
+On Windows, the configuration file is located at %APPDATA%\\alacritty\\alacritty.yml.
```
kchibisov: Are man pages a thing on Windows?

chrisduerr: I mean you can read them if you have a program to look at them. So if you're running bash inside of Alacritty, there's a good chance that `man` is available too.

But since Alacritty still runs on Windows, even when the shell offers posix things, the config file will still be at the Windows location.

kchibisov: Let's keep it then.

PR #3162, opened 2020-01-07
Fix keybinding deserialization

### LINKED ISSUE #3160: Key bindings broken with the latest master (183ceee3adf4ac5a889db3ed904e7adbf9ccb4f1)
After updating to the latest master this morning my key bindings stopped working and I get the following message to the Alacritty logs:

```
[2020-01-07 08:17] [ERROR] Problem with config: unknown variant `I`, expected `Scancode` or `Keycode`; using default value
```

So it seems to me that for some reason Alacritty doesn't recognize `I` as keycode. No idea why, it's a letter amongst the others and have worked fine up until now. Maybe a bug in the config parser or some invisible character in my config that breaks things?

Like said I'm running latest master on macOS Mojave 10.14.6 with 1.14.0 rustc installed using rustup.rs.

```
08:39:20 in alacritty on  master [$] via 𝗥 v1.40.0 •100%
[I] > ~/gits/alacritty/target/release/alacritty --version
alacritty 0.4.2-dev (183ceee)
```

Here's my configuration:
https://l.termbin.com/o0bx

chrisduerr:
Fixes #3160.

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -165,7 +165,7 @@ impl Options {
                     .help("Start the shell in the specified working directory"),
             )
             .arg(Arg::with_name("config-file").long("config-file").takes_value(true).help(
-                &format!("Specify alternative configuration file [default: {}]", CONFIG_PATH)
+                &format!("Specify alternative configuration file [default: {}]", CONFIG_PATH),
             ))
             .arg(
                 Arg::with_name("command")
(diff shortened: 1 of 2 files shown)


kchibisov:
Seems to fix the issue for me ;- )

kchibisov:
Normal bindings seems to work.
Scancode bindings seems to work.
Error for wrong scancode bindings seems to work.
Error for misspelled modifiers/mode, etc seems to work.
Mouse bindings seems to work.

So I don't see any difference from old correct behavior???

chrisduerr:
Thanks for testing, looks like this should be good to go then.

PR #3169, opened 2020-01-08
Bump version to 0.4.1-rc2

chrisduerr:


nixpulvis:
@chrisduerr if we had a script to bump RC versions, would they even need PRs?

chrisduerr:
The PRs are mostly nice as sanity checks, since I'd never trust any update to just work.

I know that I didn't miss anything, since I've bumped everything based on the previous version of the RC. It has the exact same number of lines and files changed, so the only question is if the versions were bumped correctly and that is easily verified.

Currently it's too easy to justify writing a script for it I'd say. And making sure the script says working, regardless of possible other changes made to the input file, would likely be more work than just bumping everything manually.

Though I did think about writing a script for this myself before, so the idea definitely isn't far fetched.

PR #3178, opened 2020-01-10
Improved Numpad Enter support

zoxves:
The default binding for NumpadEnter doesn't work in some cases like `vifm` or `fzf`.
With this change Numpad Enter is perfectly fine everywhere (as far as I tested).

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -319,7 +319,7 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
         F18; Action::Esc("\x1b[32~".into());
         F19; Action::Esc("\x1b[33~".into());
         F20; Action::Esc("\x1b[34~".into());
-        NumpadEnter; Action::Esc("\n".into());
+        NumpadEnter; Action::Esc("\x0d".into());
     );
 
     //   Code     Modifiers


chrisduerr:
@nixpulvis I don't own a numpad. Could you verify that other terminals send the `\x0d` escape when the enter key is pressed on numpad?

nixpulvis:
Upon a quick test, I doesn't seem like my keyboard's numpad is setup the same way. When I press enter (on the numpad) it gives scancode 96 (hex 0x60), which isn't the same as our `NumpadEnter` it seems. 

Looking a bit at how this is implemented in winit, it seems like this could be effected by the platform, @zoxves could you help me out by providing some information about your setup, like: OS, Locale, and Keyboard layout information. Specifically, what keycode is reported when you press your numpad enter key? I think my system needs changes, but it's not 100% clear to me currently if it's a change needed in my configuration, keyboard, or in winit itself.

Once I have an understanding of how to trigger this key, I can better access how this should work in the various terminals, then we can merge this as needed.

cole-h:
`xterm`, `urxvt`, and `kitty` all send `0x0d` (on numpad enter) on my system. Just as extra information.

chrisduerr:
@cole-h Thanks for the feedback, that's exactly what I was looking for. Could you confirm to me how you've tested this? Did you use the command we got in our config file (`showkey --scancodes`)?

alacritty/src/config/bindings.rs
```diff
@@ -319,7 +319,7 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
         F18; Action::Esc("\x1b[32~".into());
         F19; Action::Esc("\x1b[33~".into());
         F20; Action::Esc("\x1b[34~".into());
-        NumpadEnter; Action::Esc("\n".into());
+        NumpadEnter; Action::Esc("\x0d".into());
```
chrisduerr: Could you try just removing this binding entirely? I think it should work without.

chrisduerr:
Oh, a thing I missed in my review: We need a change log entry when changing things. So please make sure to add one!

alacritty/src/config/bindings.rs
zoxves: Yea, I tried it already - doesn't work.

cole-h:
Did you remember to remove the binding from your Alacritty config as well? I didn't at first and also thought this was broken (removing the bind let it function as expected).

zoxves:
Actually, I've removed it in config and in code, but never tested removal both in config and in code - only separately and I don't have opportunity to test it right now.

cole-h:
In my (albeit brief) testing earlier, removal of both was required for expected functionality.

dnkl:
<kbd>KP Enter</kbd> should emit `\eOM` in _keypad application_ mode, with modifiers enabled (so e.g. `\eO2M` for <kbd>shift</kbd>+<kbd>KP Enter</kbd> (with num-lock off).

```sh
% echo -e '\e='
% sudo showkey -a

Press any keys - Ctrl-D will terminate this program

^[OM     27 0033 0x1b   # KP_Enter
         79 0117 0x4f
         77 0115 0x4d
^[O2M    27 0033 0x1b   # Shift+KP_Enter
         79 0117 0x4f
         50 0062 0x32
         77 0115 0x4d
^[O5M    27 0033 0x1b   # Ctrl+KP_Enter
         79 0117 0x4f
         53 0065 0x35
         77 0115 0x4d
#... and so on
```
XTerm does urxvt does too, except it doesn't emit modifiers (i.e. it always emits `\eOM`).

chrisduerr:
> KP Enter should emit \eOM in keypad application mode, with modifiers enabled (so e.g. \eO2M for shift+KP Enter (with num-lock off).

That seems like a different issue, but it should be trivial to add for someone who can test with a numpad. All the required code to add the modifier values exists in `alacritty/src/config/bindings.rs` already. So someone just needs to add it and verify that it works.

kchibisov:
I've tested removing a bindings from a config and from alacritty source and it was sending `0x0d` for me, However I don't see anything bad in clearly stating what we should send for Numpad enter in a bindings. However from my testing we're missing some numpad bindings(e.g. arrows), but winit doesn't know them(at least on Wayland).

LamaRaz:
Is there a way to fix this problem without patching?

In alacritty numpad_enter for some applications like fzf acts as arrow_down while for others like mc works as enter should

very confusing and highly annoying

this problem doesn't exist in all other terminals

notchairmk:
It seems like the issue is potentially on the winit side and the mapping of bindings between winit and alacritty. In winit, NumpadEnter only has bindings for macOS and Wayland. Testing it out on X11, I was getting `virtual_keycode: Some(Return)`, which was being handled "correctly". So just removing the binding works on macOS, because on adding a new line a carriage return `ReturnChar` event is included:

https://github.com/rust-windowing/winit/blob/26775fa0b621c225753cefb2a3babead3be33f46/src/platform_impl/macos/view.rs#L542-L548

But I don't believe it would work for Wayland.

tl;dr it seems like explicitly setting to escaped hex carriage return is the best path to go down. I can open up a new PR if this seems accurate, since this PR may be stale.

chrisduerr:
Closing since stalled with unresponsive author. If someone who owns a numpad is willing to send a PR refining the escapes sent with it, I recommend opening a new PR.

Issue #3179, opened 2020-01-10
copypasta: Fix build on ios


dvc94ch:
Having the clipboard be a noop would be good enough for my purposes. The build currently detects it as a unix and tries to link against x11 libraries.

chrisduerr:
I'm not sure where it fails to build and I don't have any iOS devices to test this on, but if it currently fails on iOS I'd certainly welcome a patch that just fixes this by using the noop clipboard.

dvc94ch:
Same issue here, that's why I opened an issue and not a pr. I noticed the failure through the ci build, still procrastinating borrowing a mac/iphone.

PR #3180, opened 2020-01-10
Fix copypasta compilation on ios

kchibisov:
Fixes: #3179


kchibisov:
I've followed `android` here and took `cfg` for ios from https://github.com/rust-windowing/winit/blob/633d0deeaeee06b1e5b90e26d8c53dd39e4d755d/src/platform_impl/mod.rs#L29 .

kchibisov:
moving thing to a proper repo.

PR #3184, opened 2020-01-11
Fix font size reset when moving between screens

### LINKED ISSUE #3183: Alacritty reverts the font change after moving across the monitors
I'm not sure that it's a desired behavior, however I feel like if a user changed font size from `X` to `Y` on a monitor `1` and then moved a window from a monitor `1` to a monitor `2` we should speculate with the updated font size and not with the default one. So the font size on a monitor `2` should be `Y * scale_factor` and not `X * scale_factor`.

In addition to this the next update after moving alacritty from `1` to `2` will be based on `Y` and not `X`, so it seems like a bug to me.

chrisduerr:
Fixes #3183.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -463,11 +463,8 @@ impl<N: Notify + OnResize> Processor<N> {
                     let display_update_pending = &mut processor.ctx.display_update_pending;
 
                     // Push current font to update its DPR
-                    if display_update_pending.font.is_none() {
-                        display_update_pending.font = Some(
-                            processor.ctx.config.font.clone().with_size(*processor.ctx.font_size),
-                        );
-                    }
+                    display_update_pending.font =
+                        Some(processor.ctx.config.font.clone().with_size(*processor.ctx.font_size));
 
                     // Resize to event's dimensions, since no resize event is emitted on Wayland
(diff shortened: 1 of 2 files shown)


alacritty/src/event.rs
```diff
@@ -460,10 +460,13 @@ impl<N: Notify + OnResize> Processor<N> {
         match event {
             GlutinEvent::UserEvent(event) => match event {
                 Event::DPRChanged(scale_factor, (width, height)) => {
-                    let display_update_pending = &mut processor.ctx.display_update_pending;
+                    let display_update_pending = &mut ctx.display_update_pending;
```
kchibisov: processor.ctx

chrisduerr: Oops, tested out to extract that and looks like I've missed this when undoing. Thanks.

alacritty/src/event.rs
```diff
@@ -460,10 +460,13 @@ impl<N: Notify + OnResize> Processor<N> {
...
 
                     // Push current font to update its DPR
-                    display_update_pending.font = Some(processor.ctx.config.font.clone());
+                    if display_update_pending.font.is_none() {
+                        display_update_pending.font =
```
kchibisov: display_update_pending.font = Some(processor.ctx.config.font.clone().with_size(*processor.ctx.font_size));

PR #3207, opened 2020-01-14
Expand line selection across wrapped lines

chrisduerr:
This is a follow-up to https://github.com/alacritty/alacritty/pull/2717.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -115,7 +115,7 @@ impl<T> Search for Term<T> {
 
     fn line_search_left(&self, mut point: Point<usize>) -> Point<usize> {
         while point.line + 1 < self.grid.len()
-            && self.grid[point.line + 1][self.grid.num_cols() - 1].flags.contains(cell::Flags::WRAPLINE)
+            && self.grid[point.line + 1][self.grid.num_cols() - 1].flags.contains(Flags::WRAPLINE)
         {
             point.line += 1;
         }
@@ -126,7 +126,7 @@ impl<T> Search for Term<T> {
     }
 
     fn line_search_right(&self, mut point: Point<usize>) -> Point<usize> {
-        while self.grid[point.line][self.grid.num_cols() - 1].flags.contains(cell::Flags::WRAPLINE) {
+        while self.grid[point.line][self.grid.num_cols() - 1].flags.contains(Flags::WRAPLINE) {
             point.line -= 1;
         }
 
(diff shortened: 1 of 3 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -56,6 +56,10 @@ pub trait Search {
     fn semantic_search_left(&self, _: Point<usize>) -> Point<usize>;
     /// Find the nearest semantic boundary _to the point_ of provided point.
     fn semantic_search_right(&self, _: Point<usize>) -> Point<usize>;
+    /// Find the beginning of a potentially wrapped line.
+    fn line_search_left(&self, _: Point<usize>) -> Point<usize>;
```
chrisduerr: @nixpulvis Any feedback on the function documentation? I feel like it's not 100% clear in what exactly it does.

ike: `Find the beginning of a line, following line wraps`?

maybe the function name could be clearer, like `wrap_line_search_left`.

chrisduerr: I've intentionally removed the `wrap` from the function name, since `wrap_line_search_left` sounds too much like the function actually does something wrapping related. But it actually just follows across linewraps. It's also more consistent with the other search functions for selection, since we're talking about a line selection and not a wrapline selection.

I like the comment suggestion you've made though.

cole-h: `/// Find the beginning of a line or a line that has been wrapped.`, or something similar?

Otherwise, IMHO, the current is fine.

chrisduerr: That sounds like it's finding a line that has been wrapped. But we're finding the beginning of the line, across potential wrappage. So it's not entirely accurate.

cole-h: I was idling so long on that comment I didn't see @ike's above. I like that wording, or something similar.

alacritty_terminal/src/term/mod.rs
```diff
@@ -109,6 +113,28 @@ impl<T> Search for Term<T> {
         point
     }
 
+    fn line_search_left(&self, mut point: Point<usize>) -> Point<usize> {
+        while point.line + 1 < self.grid.len()
+            && self.grid[point.line + 1][self.grid.num_cols() - 1].flags.contains(cell::Flags::WRAPLINE)
```
kchibisov: Line is toooo long, I guess you should run `rustfmt`.

kchibisov: Oh, and, please, could you refactor this code into a for loop.

PR #3210, opened 2020-01-15
Remove fs::canonicalize from Windows pty startup

### LINKED ISSUE #3198: Powershell starts up with extra stuff prepended to the current directory name
After updating to the latest version, when Alacritty starts, either from the `Open Alacritty Here` context menu, or when running it from a previous Alacritty instance, the path has `Microsoft.PowerShell.Core\FileSystem::\\?\` prepended to. For example, if I open Alacritty in `C:\home` my prompt says `PS Microsoft.PowerShell.Core\FileSystem::\\?\C:\home>`. Other than the long prompt, everything seems to work properly. If I run `cd C:\home` then the prompt appears how I would expect it to on startup: `PS C:\home>`. Oddly enough, if I start a powershell 6 terminal, then run `alacritty.exe` inside there, the resulting instance of Alacritty does not have this long prompt. It does print `Cannot load PSReadline module.  Console is running without PSReadline.` above the prompt only in that case though.

### System

OS: Windows 10
Version: alacritty 0.4.1 (4d9253b)
Windows: ConPTY

davidhewitt:
I was looking further at #3198

`std::fs::canonicalize()` is the call responsible for creating the UNC paths. But I think we don't actually need to canonicalize paths for either of the the ConPTY or WinPTY backends. For example, I tested both with relative paths and they work fine.

So I suggest that rather than making some dance with `canonicalize()` followed by working out how to remove the UNC prefix ourselves, we just faithfully pass the user's input `working_directory` on to the PTY. In that case if there's a problem, that's entirely at the hands of the user's config or command line options.

Related but not necessary for this PR: I also thought it would be wise to check that the `working_directory` in the user's config file is an absolute path. (Relative paths in the config don't seem to be meaningful to me.)

I haven't yet updated the changelog; will write an appropriate message if you agree that this is the right solution for #3198 

Closes #3198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Font size resetting when Alacritty is moved between screens
 - Limited payload length in clipboard escape (used for Tmux copy/paste)
 - Alacritty not ignoring keyboard events for changing WM focus on X11
+- Regression which added a UNC path prefix to the working directory on Windows
 
 ## 0.4.1
 
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -203,10 +203,7 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, _window_id: Option<usize>) ->
     }
 
     let cmdline = win32_string(&cmdline(&config));
-    let cwd = config
-        .working_directory
-        .as_ref()
-        .map(|pb| win32_string(&pb.as_path().canonicalize().unwrap()));
+    let cwd = config.working_directory.as_ref().map(win32_string);
 
     let mut proc_info: PROCESS_INFORMATION = Default::default();
     unsafe {
--- a/alacritty_terminal/src/tty/windows/winpty.rs
+++ b/alacritty_terminal/src/tty/windows/winpty.rs
@@ -42,14 +42,13 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, _window_id: Option<usize>) ->
     let (conin, conout) = (agent.conin_name(), agent.conout_name());
 
     let cmdline = cmdline(&config);
-    let cwd = config.working_directory.as_ref().map(|pb| pb.as_path().canonicalize().unwrap());
 
     // Spawn process
     let spawnconfig = SpawnConfig::new(
         SpawnFlags::AUTO_SHUTDOWN | SpawnFlags::EXIT_AFTER_SHUTDOWN,
         None, // appname
         Some(&cmdline),
-        cwd.as_ref().map(|p| p.as_ref()),
+        config.working_directory.as_ref().map(|p| p.as_path()),
         None, // Env
     )
     .unwrap();


davidhewitt:
(Rebased onto master)

chrisduerr:
> std::fs::canonicalize() is the call responsible for creating the UNC paths. But I think we don't actually need to canonicalize paths for either of the the ConPTY or WinPTY backends.

I've actually stumbled across Windows canonicalizing paths in a recent change I've made to the type of the `working_directory` field on the config. I was a bit surprised myself, since we do not canonicalize the path on unix. So assuming that it works for you, I'd say it makes a lot of sense.

> Related but not necessary for this PR: I also thought it would be wise to check that the working_directory in the user's config file is an absolute path. (Relative paths in the config don't seem to be meaningful to me.)

Why would we make assumptions about what should and shouldn't be done, when the failure case isn't critical and it just removes functionality? As far as I can tell it, there's no benefit in printing an error about requiring absolute paths, since we already print an error when the path does not exist.

> I haven't yet updated the changelog; will write an appropriate message if you agree that this is the right solution for #3198

Please do. Sounds like a good approach to me.

alacritty/src/config/mod.rs
```diff
@@ -168,12 +168,24 @@ fn parse_config(contents: &str) -> Result<Config> {
...
+fn check_working_directory_absolute(config: &Config) {
+    if let Some(dir) = &config.working_directory {
+        if !dir.is_absolute() {
+            warn!(
+                target: LOG_TARGET_CONFIG,
+                "Config working_directory is not absolute: {:?}", dir
+            );
+        }
+    }
+}
```
chrisduerr: This is not related to this PR and thus should not be included here.

Additionally, I think it's not desirable, so I'm not sure if opening a separate PR would prove fruitful.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -206,7 +206,7 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, _window_id: Option<usize>) ->
     let cwd = config
         .working_directory
         .as_ref()
-        .map(|pb| win32_string(&pb.as_path().canonicalize().unwrap()));
+        .map(win32_string);
```
chrisduerr: Do we still need the `.as_ref()`?

alacritty/src/config/mod.rs
davidhewitt: 👍 I'll bin it.

alacritty_terminal/src/tty/windows/conpty.rs
davidhewitt: Sadly yes - `Option::map` takes `self`, so we need to use `as_ref()` first to get a new `Option<&PathBuf>` to consume.

alacritty/src/config/mod.rs
chrisduerr: Just out of curiosity, are you aware of the recent change about printing errors on startup with invalid working directories?

I can see the motivation behind this on top of our latest release, since that has no indication of incorrect working directories and will just crash hard. However with the latest master I feel like we already provide sufficient information.

davidhewitt:
> since we already print an error when the path does not exist.

Oh nice; I was working off a slightly older master originally which didn't have that check! 👍

alacritty_terminal/src/tty/windows/conpty.rs
chrisduerr: Oh, right, we access something on our config and can't just consume the original value. Nevermind then, thought for a second that we had full ownership over the path.

alacritty/src/config/mod.rs
chrisduerr: > Oh nice; I was working off a slightly older master originally which didn't have that check! +1

Nevermind, you answered me already. That makes a lot more sense then!

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Font size resetting when Alacritty is moved between screens
 - Limited payload length in clipboard escape (used for Tmux copy/paste)
 - Alacritty not ignoring keyboard events for changing WM focus on X11
+- Alacritty no longer canonicalizes the working directory on Windows
```
davidhewitt: Alternative message: `Alacritty no longer converts the working directory to a UNC canonical path on Windows`

davidhewitt:
Other than a minor bikeshed on which Changelog entry you prefer, this is ready to go.

Closes #3198

CHANGELOG.md
chrisduerr: I'm not sure if the user have to care about the canonicalization at all? We removed it exactly because it ***shouldn't*** affect the user.

We no longer convert the path to a weird UNC path, but I'm not sure if most people know what that is (as a Linux dev, I didn't). Though there's nothing wrong with using the technically correct terms as long as everyone can still understand the change.

Maybe something more like this would work?

```
- Remove UNC path prefix from Windows working directory
```

It's not optimal either, since the usage of `UNC` might still be confusing, but it should be a bit better. I also feel like it might be worth mentioning this was a regression, but I'm not certain how to best state that without just bloating up the message unnecessarily.

If you've got any better suggestions, please let me know.

davidhewitt: How about

"- Regression which added UNC path prefixes to working directories on Windows"

chrisduerr: Sounds like a winner to me! I'd just make it `working directory` instead of `working directories`. Since I think that more accurately captures that we're talking about the main working directory Alacritty is launched on and not that this is always prefixed to the working directory (it's fixed once you `cd`, afaik).

davidhewitt:
Changelog entry tweaked, and also squashed.

PR #3211, opened 2020-01-15
Update font crate to 2018 edition

davidhewitt:
I've updated the font crate to 2018 edition, which will ease some of the path pain working on #3127.

All I've done is:
- Set the edition in `Cargo.toml`
- Remove all `extern crate` declarations
- Remove all `self::` where unneeded
- Add macro imports where needed


chrisduerr:
Looks like there are some macOS build errors @davidhewitt. Would you mind looking into that? I think it should be simple to fix just using the CI build output. But if you need any help, let me know. I can fix things on actual macOS hardware tomorrow.

davidhewitt:
Naturally this failed on the one platform I couldn't easily test 😢. I should have pushed now, fingers crossed this time the CI is green...

davidhewitt:
CI all green 🎉

PR #3213, opened 2020-01-16
Modded alacritty.ico to be highres non blurry

michaelfresco:
I took the  `/extra/logo/compat/alacritty-term.svg` (the one without scanlines) and converted it to a 256x256 ico.

This is the result:

before_after_icon_change

(before / after)

---------------------

I used [this](https://gist.github.com/michaelfresco/1f7501b190f1a0bee7d598ceee9cb066) bash script to generate the icon file. (requires both inkscape and magick to be installed)


chrisduerr:
This needs to be tested on Windows, where this icon is also used.

davidhewitt:
Some observations:

- The new .ico diplays fine on Windows
- We are loading the 256x256 image from the bytes embedded in `alacritty_terminal/src/window.rs`
- Windows is downscaling this image to fit the window title bar. Its algorithm seems to be fairly bad at this.
- Because the new image is sharper, Window's downscaled version of the new icon appears slightly more pixelated in the title bar on Windows

I just had a play and I have a better way for us to set the window icon on Windows. It would be best done as an an upstream winit patch though.

I suggest you merge this PR, and I'll adjust the Windows icon loading separately in the near future.

michaelfresco:
I added a log entry and merged with the master branch.

PR #3217, opened 2020-01-16
Replace `.nth(0)` with `.next()`

cole-h:
Clippy says that `.next()` is more readable than `.nth(0)`, and I am
inclined to agree.

https://rust-lang.github.io/rust-clippy/master/index.html#iter_nth_zero

### DIFF
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -338,8 +338,7 @@ impl Descriptor {
         let fallbacks = if load_fallbacks {
             descriptors_for_family("Menlo")
                 .into_iter()
-                .filter(|d| d.font_name == "Menlo-Regular")
-                .nth(0)
+                .find(|d| d.font_name == "Menlo-Regular")
                 .map(|descriptor| {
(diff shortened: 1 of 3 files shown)


chrisduerr:
One is fixed in https://github.com/alacritty/alacritty/pull/3216, though I'm fine with doing that in here and then having everything bundled. Since #3216 is likely to be merged after this anyways, since the review will take longer.

The other one is in `font/src/darwin/mod.rs`. I'd assume that wasn't found since `cargo clippy` was run on a linux system. So good catch @kchibisov.

cole-h:
Thanks for the catch @kchibisov. I'll need to rebase after 3216 is merged, but otherwise this should be good to go.

font/src/darwin/mod.rs
```diff
@@ -338,8 +338,7 @@ impl Descriptor {
         let fallbacks = if load_fallbacks {
             descriptors_for_family("Menlo")
                 .into_iter()
-                .filter(|d| d.font_name == "Menlo-Regular")
-                .next()
```
kchibisov: Oh, it was a next over an option, right?

Issue #2779, opened 2019-09-05
Change default draw_bold_text_with_bright_colors behavior


chrisduerr:
At the moment, Alacritty draws bold text with the bright colors by default. This can be extremely annoying when the goal is to just get some bold text with non-bright colors and it's not necessary to use this if you actually do want to use bright colors. To make each escape do one thing and do that well, the option `draw_bold_text_with_bright_colors` is currently available to change this behavior.

Generally, I think it makes more sense to have this option disabled by default, however a choice was made against that for compatibility with other terminal emulators and keeping up with user expectations.

Kitty has never had any way to get this mixed bold escape and it has come to my knowledge in #2776 that gnome has recently disconnected the bold and bright escapes by default too. With other terminal emulators changing this and offering the chance to make right what has been done wrong in the past, I think it would be a good choice to follow suit here.

However, we're talking about a default option here, so what really matters is how most people feel about this, so I'd like to take this opportunity to get some comments on this possible change. Maybe @kchibisov and @nixpulvis can chime in, however I'm really curious about all kinds of opinions (both for and against this).

egmontkob:
In the 2nd & 3rd paragraphs you seem to have mixed up the "enabled" vs "disabled" states. I guess you had in mind _enabling "draw_blah = false"_, didn't you?

nixpulvis:
My initial reaction is that this is a good idea. If I forget all the history, I'm having trouble coming up with a rational for why bold and bright would be tied together at all.

I'll be curious to see what examples @egmontkob digs up for negative feedback, but it's got to all be historical support. I'd hope applications would be able to adapt, generally. Though, we'd just be changing the default value of `draw_bold_text_with_bright_colors` so there'd be an easy solution for our users.

FWIW, I googled "ANSI Bold" and the first link I got was this one: http://www.lihaoyi.com/post/BuildyourownCommandLinewithANSIescapecodes.html, which makes it seem like "Bold" and "Bright" are the same thing, so it's pretty clear this is a confused topic in the wild.

I've had (most) of my bright colors set to the same color as the non-bright version, which I'm thinking was probably a result of me not knowing about this bright/bold option. Now I'm wondering if I can come up with a nice set of actual bright colors for my theme (One Dark). I'll change my setting and see how things go, just as another point of reference.

kchibisov:
[...]

PR #3224, opened 2020-01-19
Default `draw_bold_text_with_bright_colors` to `false`

kchibisov:
The option has some "historical" background, for example it could break
ncurses applications like alsamixer, however the default to `true`
seems strange nowadays, because it makes impossible to draw bold text in
a "normal" color. Some other terminals (e.g. vte, kitty) also started default to
`false` here.

We're not removing the option, but just flipping a bit, and knowing that users
tend to copy paste our config in a past, will likely continue to have it
defaulted to `true` in their configuration files. The only problem is the users
who don't have this option in a config file now or new users, however the change
in color is noticeable and it also stated very clear in our configuration file,
so it shouldn't be a problem to flip things back.

Implements #2779

I think keeping this around isn't a good idea, so let's just push it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Renamed `WINIT_HIDPI_FACTOR` environment variable to `WINIT_X11_SCALE_FACTOR`
 - Print an error instead of crashing, when startup working directory is invalid
 - Line selection will now expand across wrapped lines
+- The default value for `draw_bold_text_with_bright_colors` is now `false`
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


chrisduerr:
> We're not removing the option, but just flipping a bit, and knowing that users
tend to copy paste our config in a past, will likely continue to have it
defaulted to true in their configuration files.

Note that this is only true for people who have done this is the past, since recent releases have everything commented out in the config.

I've personally kinda lost track that I opened this issue, but based on this request for comments it seems like it was unanimously in favor. All comments seem to agree with the idea and even the "reactions" on the original post are exclusively positive.

While this isn't a definitive indication that the majority of users will prefer this choice, it should be sufficient evidence to try out the change in our master. That way we can collect information by confronting users with it, that maybe aren't following our issue tracker.

So I'd vote in favor, though I'd like to make this unanimous and see if @nixpulvis has any reasons why we would maybe wait a bit longer before testing this out.

Issue #1873, opened 2018-12-06
Consider removing auto_scroll


chrisduerr:
The auto_scroll option adds the ability to jump to the bottom of the terminal whenever new content is written to it. This functionality is useful, however there are some issues with the implementation.

I've tested removing this code and the `trace!` in the input method and it turns out this improves Alacritty's performance by 3.39% in the alt-screen-random-write benchmark on my machine. This is with the option disabled (so default).

Now this is an extremely minor performance difference (<5%), however since Alacritty's priority is performance, I think it's worth discussing if this feature is relevant enough to introduce any kind of regression for users who have it disabled.

I'd be curious to know if there is anyone actively using this setting in their Alacritty configuration file or if it's just mostly unnecessary. This is not much of a call to action, since Alacritty's fine the way it is right now. However I think it's worth keeping track of potential performance hazards.

chrisduerr:
Note that there's another feature which seems pretty simple, but has a significant impact on performance.

The escape sequence `echo -e "test\e[b"` requires the previous character to be stored in between calls to the `input` method in `src/term/mod.rs`. Because of that, every time a character is written to the terminal emulator, it's stored in a temporary buffer. Removing this increases performance by 5-10%.

I feel like it might be possible to support this command by looking at the cursor and reading the character of the previous column on demand, however I can see some issues with that implementation that might cause problems.

jakeydoodle123:
**Only about auto-scroll:** 
Are you talking about removing the option to have auto scrolling to remove auto scrolling all together? 

If you are talking about removing it all together I think that isn't a great idea. Though the objective is high performance, there are a lot of tasks where you pretty much need autoscrolling (compiling, updating, etc.) and I think that is an important trade-off especially for such a negligible boost.

nixpulvis:
I agree with the sentiment that we shouldn't be removing features altogether for performance if we can have them disabled and not costing anything. That said if we need to move things to compile time features to speed things up while we figure out better fixes, I'm all for that.

chrisduerr:
This is not about removing the automatic scrolling when a command is running. This option is disabled by default, so unless you've manually enabled it, you likely never noticed it.

What it does is that when you're not at the bottom of the screen, but new output is printed, Alacritty will automatically scroll to the bottom. If the viewport is already scrolled all the way down, this has no effect (other than the performance impact).

> if we can have them disabled and not costing anything

That's the whole idea, checking at runtime is too expensive since the method is called so much.

> move things to compile time features

[...]

PR #3226, opened 2020-01-19
Remove scrolling.auto_scroll feature

kchibisov:
Fixes: #1873

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
-- Config option `auto_scroll` and its corresponding feature
+- Config option `auto_scroll`, which is now always disabled
 
 ## 0.4.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Blurred icons in KDE task switcher (alacritty.ico is now high-res)
 - Consecutive builds failing on macOS due to preexisting `/Application` symlink
 
+### Removed
+
+- `auto_scroll` config option as well as behavior
+
 ## 0.4.1
 
 ### Packaging
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
-- `auto_scroll` config option as well as behavior
+- Config option `auto_scroll` and its corresponding feature
 
 ## 0.4.1
 
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -208,8 +208,8 @@ fn print_deprecation_warnings(config: &Config) {
     if config.scrolling.auto_scroll.is_some() {
         warn!(
             target: LOG_TARGET_CONFIG,
-            "Config scrolling.auto_scroll was removed as well as the corresponding feature, there's \
-             no need to keep it anymore"
+            "Config scrolling.auto_scroll was removed as well as the corresponding feature, \
+             there's no need to keep it anymore"
         );
     }
 }
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -208,7 +208,8 @@ fn print_deprecation_warnings(config: &Config) {
     if config.scrolling.auto_scroll.is_some() {
         warn!(
             target: LOG_TARGET_CONFIG,
-            "Config scrolling.auto_scroll has been removed and is now always disabled, it can be safely removed from the config"
+            "Config scrolling.auto_scroll has been removed and is now always disabled, it can be \
+             safely removed from the config"
         );
     }
 }
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -208,7 +208,7 @@ fn print_deprecation_warnings(config: &Config) {
     if config.scrolling.auto_scroll.is_some() {
         warn!(
             target: LOG_TARGET_CONFIG,
-            "Config scrolling.auto_scroll has been removed, and is now always disabled"
+            "Config scrolling.auto_scroll has been removed and is now always disabled, it can be safely removed from the config"
         );
     }
 }
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -208,7 +208,7 @@ fn print_deprecation_warnings(config: &Config) {
     if config.scrolling.auto_scroll.is_some() {
         warn!(
             target: LOG_TARGET_CONFIG,
-            "Config scrolling.auto_scroll was removed as well as a corresponding feature, there's \
+            "Config scrolling.auto_scroll was removed as well as the corresponding feature, there's \
              no need to keep it anymore"
         );
     }
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -208,8 +208,7 @@ fn print_deprecation_warnings(config: &Config) {
     if config.scrolling.auto_scroll.is_some() {
         warn!(
             target: LOG_TARGET_CONFIG,
-            "Config scrolling.auto_scroll was removed as well as the corresponding feature, \
-             there's no need to keep it anymore"
+            "Config scrolling.auto_scroll has been removed, and now is always disabled"
         );
     }
 }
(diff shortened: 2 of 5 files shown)


alacritty/src/config/mod.rs
```diff
@@ -204,6 +204,14 @@ fn print_deprecation_warnings(config: &Config) {
              lines"
         );
     }
+
+    if config.scrolling.auto_scroll.is_some() {
+        warn!(
+            target: LOG_TARGET_CONFIG,
+            "Config scrolling.auto_scroll was removed as well as a corresponding feature, there's \
```
chrisduerr:
```suggestion
            "Config scrolling.auto_scroll was removed as well as the corresponding feature, there's \
```

CHANGELOG.md
```diff
@@ -31,6 +31,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Blurred icons in KDE task switcher (alacritty.ico is now high-res)
 - Consecutive builds failing on macOS due to preexisting `/Application` symlink
 
+### Removed
+
+- `auto_scroll` config option as well as behavior
```
chrisduerr: I'm not sure why, but "behavior" sounds a bit strange here. I don't have a great suggestion for improvement, but maybe using what you've already used in the removal warning can make this a little nicer to read?

```suggestion
- Config option `auto_scroll` and its corresponding feature
```

kchibisov: since your suggestion mirrors the thing in warning it makes sense.

alacritty/src/config/mod.rs
```diff
@@ -204,6 +204,14 @@ fn print_deprecation_warnings(config: &Config) {
              lines"
         );
     }
+
+    if config.scrolling.auto_scroll.is_some() {
+        warn!(
+            target: LOG_TARGET_CONFIG,
+            "Config scrolling.auto_scroll was removed as well as the corresponding feature, \
+             there's no need to keep it anymore"
```
cole-h: `"Config scrolling.autoscroll and its corresponding feature have been removed"`?

Additionally, do we really need to say "corresponding feature"? Just sounds weird to me. If we have to keep it, maybe change it to something like "related functionality"? (would also need to be changed in the changelog)

alacritty_terminal/src/config/scrolling.rs
```diff
@@ -11,8 +11,10 @@ pub struct Scrolling {
     history: ScrollingHistory,
     #[serde(deserialize_with = "failure_default")]
     multiplier: ScrollingMultiplier,
+
+    // TODO: REMOVED
     #[serde(deserialize_with = "failure_default")]
-    pub auto_scroll: bool,
+    pub auto_scroll: Option<bool>,
```
nixpulvis: Just a little work, but could we make these `Deprecated<bool>` and clean up some of the code a bit?

Might save us some TODO comments, and might make the warning code simpler somehow.

alacritty/src/config/mod.rs
nixpulvis: Personally, I'd go for short and sweet. I think people would understand a warning for just "Config scrolling.auto_scroll has been removed".

alacritty_terminal/src/config/scrolling.rs
cole-h: (in this case it should be something like `Removed<bool>` since it no longer influences functionality)

alacritty/src/config/mod.rs
chrisduerr: If the config option has been removed, it's not immediately clear if it is now default `true` or default `false` though.

But I had the same thought as you, I'm not a huge fan of the naming, but I don't have any better proposal.

alacritty_terminal/src/config/scrolling.rs
chrisduerr: How would it save us from the todo comments and make the warning code simpler?

alacritty/src/config/mod.rs
nixpulvis: "Config scrolling.auto_scroll has been removed, and is now always enabled"?

alacritty_terminal/src/config/scrolling.rs
chrisduerr: Just to expand a bit on this, if we wanted to hide this behind an abstraction we would have to put a bit more effort into it and write a macro for an annotation for example.

Something like `#[deprecated("Deprecation Message")]`.

But I don't think there's any simple way to implement this, so the current implementation should be less complex.

alacritty/src/config/mod.rs
cole-h: I like that wording. Though I would still prefer if the changelog said something more along the lines of `related functionality` than `corresponding feature`. My contrived reasoning is that it could be confused with an actual `#[cfg(feature = "autoscroll")]`.

alacritty_terminal/src/config/scrolling.rs
nixpulvis: Well just wrapping things in `Option` is a bit loose. This way we explicitly be parsing Removed or Deprecated things. We could handle them generically potentially.

But at the very least it's a wrapper for something like an Option that tells us we can check for `is_removed`.

Ideally the `print_deprecation_warnings` could be streamlined a bit too. But this is secondary to just having a reliable way to detect outdated configs.

In a perfect world, this would be a flag to serde deserialization that emits the warning without even needing to allocate the field I'd think, but that could be a later enhancement on the type.

chrisduerr: > Nothing tells me (other than a comment) that the field was removed

Well it wasn't removed entirely yet. It's just not required anymore. I'm also not sure how a type would be different from a comment. Unless specific behavior is associated to it, a named type is mostly the same.

> But at the very least it's a wrapper for something like an Option that tells us we can check for is_removed.

Not all deprecated options have been removed.

> In a perfect world, this would be a flag to serde deserialization that emits the warning without even needing to allocate the field I'd think, but that could be a later enhancement on the type.

First of all, I'm not very fond of rewriting serde, so that's not a realistic option. An annotation would be possible, however as previously mentioned that's more work than it is currently worth imho.

The much bigger problem I see is that there currently is no version attached to the deprecation warnings. So I have no way to tell when something deprecated should be completely removed. If I would write a macro to do all this for me, that's something I would include.

So yeah, enhancements here would be possible. But I wouldn't do that with a custom type, since it does not solve any of the fundamental problems imho. If there's a desire to fix these, a macro annotation is the way to go.

alacritty/src/config/mod.rs
chrisduerr: We don't have such a feature and our changelogs are not geared torwards Rust users. So I don't think this is very relevant.

Though I do think that `functionality` is probably the better term in general.

alacritty_terminal/src/config/scrolling.rs
kchibisov: Let's just keep this thing simple. There could be a state when an option could be completely removed and a new type aka `Removed<T>` will become useless. So using an Option<T> to solve this task makes sense, since its a stdlib feature that we're already heavily using among the code for similar things.

alacritty/src/config/mod.rs
```diff
@@ -208,8 +208,7 @@ fn print_deprecation_warnings(config: &Config) {
     if config.scrolling.auto_scroll.is_some() {
         warn!(
             target: LOG_TARGET_CONFIG,
-            "Config scrolling.auto_scroll was removed as well as the corresponding feature, \
-             there's no need to keep it anymore"
+            "Config scrolling.auto_scroll has been removed, and now is always disabled"
```
cole-h: nit: `now is` -> `is now`

CHANGELOG.md
```diff
@@ -31,6 +31,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Blurred icons in KDE task switcher (alacritty.ico is now high-res)
 - Consecutive builds failing on macOS due to preexisting `/Application` symlink
 
+### Removed
+
+- Config option `auto_scroll` and its corresponding feature
```
```suggestion
- Config option `auto_scroll`, which is now always disabled
```

I think this would improve things, right?

alacritty/src/config/mod.rs
```diff
@@ -204,6 +204,13 @@ fn print_deprecation_warnings(config: &Config) {
              lines"
         );
     }
+
+    if config.scrolling.auto_scroll.is_some() {
+        warn!(
+            target: LOG_TARGET_CONFIG,
+            "Config scrolling.auto_scroll has been removed, and is now always disabled"
```
chrisduerr: I still really liked the instruction that people could safely remove the config option. I think it's worthwhile to put that in, because even if a user has absolutely no clue, that provides a clear instruction for finding a solution.

```suggestion
            "Config scrolling.auto_scroll has been removed and is now always disabled, it can be safely removed from the config"
```

kchibisov:
I’ll run rustfmt once I later...

PR #3237, opened 2020-01-21
Fix underline position for bitmap fonts

### LINKED ISSUE #3235: Off by one pixel in underline
**Alacritty**:
2020-01-21-120809_255x41_scrot

**URxvt**:
2020-01-21-120806_255x47_scrot

`alacritty.yml`:
```
font:
    normal:
        family: Profont
    size: 6
```

Font: [profontn.otb.gz](https://github.com/alacritty/alacritty/files/4090753/profontn.otb.gz)

### System

OS: Linux, X11, i3wm
Version: `0.4.1`

### Logs

```
% alacritty -vv
Created log file at "/tmp/Alacritty-571020.log"
[2020-01-21 12:11] [INFO] Welcome to Alacritty
[2020-01-21 12:11] [INFO] Configuration loaded from "/home/ux/.config/alacritty/alacritty.yml"
[2020-01-21 12:11] [DEBUG] Estimated DPR: 1
[2020-01-21 12:11] [DEBUG] Estimated Cell Size: 5 x 10
[2020-01-21 12:11] [DEBUG] Estimated Dimensions: None
[2020-01-21 12:11] [INFO] Device pixel ratio: 1
[2020-01-21 12:11] [INFO] Initializing glyph cache...
[2020-01-21 12:11] [INFO] ... finished initializing glyph cache in 0.009332156s
[2020-01-21 12:11] [INFO] Cell Size: 5 x 10
[2020-01-21 12:11] [INFO] Padding: 0 x 0
[2020-01-21 12:11] [INFO] Width: 800, Height: 600
[2020-01-21 12:11] [INFO] PTY Dimensions: Line(60) x Column(160)
[2020-01-21 12:11] [INFO] Initialisation complete
[2020-01-21 12:11] [DEBUG] New num_cols is 65 and num_lines is 19
[2020-01-21 12:11] [INFO] Width: 327, Height: 195
[2020-01-21 12:11] [DEBUG] New num_cols is 88 and num_lines is 19
[2020-01-21 12:11] [INFO] Width: 444, Height: 195
[2020-01-21 12:11] [DEBUG] New num_cols is 88 and num_lines is 9
[2020-01-21 12:11] [INFO] Width: 444, Height: 96
[2020-01-21 12:11] [INFO] Goodbye
Deleted log file at "/tmp/Alacritty-571020.log"
```

kchibisov:
Fixes #3235.

### DIFF
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -108,8 +108,8 @@ impl Rasterize for FreeTypeRasterizer {
 
         // Fallback for bitmap fonts which do not provide underline metrics
         if underline_position == 0. {
-            underline_thickness = (descent / 5.).round();
-            underline_position = descent * 2.;
+            underline_thickness = (descent / 5.).round().max(1.0);
+            underline_position = descent / 2. - (underline_thickness / 2.).round();
         }
 
         // Get strikeout position and thickness in device pixels
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -108,7 +108,7 @@ impl Rasterize for FreeTypeRasterizer {
 
         // Fallback for bitmap fonts which do not provide underline metrics
         if underline_position == 0. {
-            underline_thickness = (descent / 5.).round().max(1.0);
+            underline_thickness = (-descent / 5.).round().max(1.);
             underline_position = descent / 2. - (underline_thickness / 2.).round();
         }
 
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -108,8 +108,8 @@ impl Rasterize for FreeTypeRasterizer {
 
         // Fallback for bitmap fonts which do not provide underline metrics
         if underline_position == 0. {
-            underline_thickness = (-descent / 5.).round().max(1.);
-            underline_position = descent / 2. - (underline_thickness / 2.).round();
+            underline_thickness = (descent.abs() / 5.).round();
+            underline_position = descent / 2.;
         }
 
         // Get strikeout position and thickness in device pixels
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -109,7 +109,7 @@ impl Rasterize for FreeTypeRasterizer {
         // Fallback for bitmap fonts which do not provide underline metrics
         if underline_position == 0. {
             underline_thickness = (descent / 5.).round();
-            underline_position = descent / 2.;
+            underline_position = descent * 2.;
         }
 
         // Get strikeout position and thickness in device pixels
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Incorrect selection status of the first cell when selection is off screen
 - Backwards bracket selection
 - Stack overflow when printing shader creation error
+- Underline position for bitmap fonts
 
 ### Removed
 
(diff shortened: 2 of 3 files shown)


font/src/ft/mod.rs
```diff
@@ -109,7 +109,7 @@ impl Rasterize for FreeTypeRasterizer {
         // Fallback for bitmap fonts which do not provide underline metrics
         if underline_position == 0. {
             underline_thickness = (descent / 5.).round();
-            underline_position = descent / 2.;
+            underline_position = descent * 2.;
```
chrisduerr: This doesn't make much sense as a metric. It will probably just always put the underline at the far bottom of the glyph, which likely looks really strange for most fonts.

The reasoning behind `descent / 2` is that the descent is the distance from the bottom of the glyph to the bottom of the cell, so half of that should put it about in the middle of the space below the glyph.

If anything, I'd probably do something more like `underline_position = descent / 2. - (underline_thickness / 2.).ceil()`.

font/src/ft/mod.rs
```diff
@@ -108,8 +108,8 @@ impl Rasterize for FreeTypeRasterizer {
 
         // Fallback for bitmap fonts which do not provide underline metrics
         if underline_position == 0. {
-            underline_thickness = (descent / 5.).round();
-            underline_position = descent / 2.;
+            underline_thickness = (-descent / 5.).round().max(1.);
+            underline_position = descent / 2. - (underline_thickness / 2.).round();
```
chrisduerr: After some more thought, the width of the underline should already be considered outside of our font crate. This is because freetype specifies the position as distance between baseline and **center** of the position. You can find that by searching for `underline_position` in the [freetype docs](https://www.freetype.org/freetype2/docs/tutorial/step2.html).

So as far as I can tell, we probably don't need to make any changes in the `ft/mod.rs` other than changing it from `descent` to `-descent`. We also already do the `max(1.)` in the `alacritty/renderer/rects.rs` file, so if we do a `max(1.)` in the `ft/mod.rs` we should probably check if it can be removed from Alacritty code. If it can't be, then we might not want it in the freetype code either, since a thickness < 1. seems acceptable then.

Which leads me to other things that can be tested out to resolve this. Currently our renderer computes the underline position based on the metrics and also includes the thickness of the underline to position it, since the position is the offset from the center. This is done [here](https://github.com/alacritty/alacritty/blob/906f14b6608480b601d08825461bd1d8eb295cfc/alacritty/src/renderer/rects.rs#L89).

Since the position is calculated based on the center of the underline, we have to divide the thickness and raise the position by half of it. However since we divide by two, this can lead to fractions (like 1.0 / 2.0 = 0.5). In these cases we need to decide if we either want to raise it to the next pixel or lower it. Since we should strive to prevent collisions with the glyphs themselves and we already have an upper limit (baseline - thickness), we should take the approach to lower it when we're not sure, rather than raising it. Currently this is not the case.

This change should thus be worth testing:
```
let mut y = (baseline - position - height / 2.).ceil();
```

font/src/ft/mod.rs
```diff
@@ -108,7 +108,7 @@ impl Rasterize for FreeTypeRasterizer {
 
         // Fallback for bitmap fonts which do not provide underline metrics
         if underline_position == 0. {
-            underline_thickness = (descent / 5.).round();
+            underline_thickness = (descent.abs() / 5.).round();
```
chrisduerr: I don't think `.abs()` is useful here, the descent should always be negative according to freetype docs. Have you found an example where this is not the case?

kchibisov: Their docs aren't clear. Read the "descender" thing.


>The descender is the vertical distance from the horizontal baseline to the lowest ‘character’ coordinate in a font face. Unfortunately, font formats don't define the descender in a uniform way. For some formats, it represents the descent of all capital latin characters (without accents), for others it is the ascent of the lowest accented character, and finally, other formats define it as being equal to bbox.yMin. This field is negative for values below the baseline.


As you can see, it does state "always negative".

PR #3239, opened 2020-01-21
Fix stack overflow when printing shader error

### LINKED ISSUE #3238: Unbounded recursion while printing error message
Note: This issue is not about the actual issue that causes alacritty to not start, but about the crash during printing the error message.

### System

OS: Linux (Alpine Edge), running on a Raspberry Pi 4
Version: alacritty 0.4.1 (a0403c7a16)
Linux/BSD: Wayland, sway
Windows: -- (doesn't apply)

### Logs

Stdout (cropped, the details part of the message is apparently printed recursively):

```
Alacritty encountered an unrecoverable error:

There was an error initializing the shaders: There was an error initializing the shaders: There  was an error initializing the shaders: There was an error initializing the shaders: [...]
```

Stderr (after the stack overflowed):

```
Segmentation fault
```

### Additional Info

https://github.com/alacritty/alacritty/blob/c84cd0fdb0335fa873f397e95a0a728be0fe15f5/alacritty/src/main.rs#L111-L115

The `println!()` in line 113 is the last line I can reach in GDB.

kchibisov:
Fixes #3238.

### DIFF
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -88,7 +88,11 @@ impl std::error::Error for Error {
 
 impl Display for Error {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        write!(f, "There was an error initializing the shaders: {}", self)
+        match self {
+            Error::ShaderCreation(err) => {
+                write!(f, "There was an error initializing the shaders: {}", err)
+            },
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


alacritty/src/renderer/mod.rs
```diff
@@ -88,7 +88,11 @@ impl std::error::Error for Error {
 
 impl Display for Error {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        write!(f, "There was an error initializing the shaders: {}", self)
+        match self {
+            Error::ShaderCreation(err) => {
+                write!(f, "There was an error initializing the shaders: {}", err)
+            },
+        }
```
chrisduerr: My initial thought was that `if let` would be correct here, however if we do add new error variants we will need to add them to this match statement. So a match is the way to go here, to prevent things from going wrong in the future.

Issue #3240, opened 2020-01-21
Background color is always black. Used to work before messing with arandr hiDPI settings.


drewboardman:
### System

OS:  Arch Linux
Version: alacritty 0.4.1
Linux/BSD: i3wm

### Logs

```
drew@muaddib ~ $ alacritty -v       
Created log file at "/tmp/Alacritty-18948.log"
[2020-01-21 16:56] [INFO] Welcome to Alacritty
[2020-01-21 16:56] [INFO] Device pixel ratio: 1.5
[2020-01-21 16:56] [INFO] Initializing glyph cache...
[2020-01-21 16:56] [INFO] ... finished initializing glyph cache in 0.006611714s
[2020-01-21 16:56] [INFO] Cell Size: 13 x 26
[2020-01-21 16:56] [INFO] Padding: 0 x 0
[2020-01-21 16:56] [INFO] Width: 1200, Height: 900
[2020-01-21 16:56] [INFO] PTY Dimensions: Line(34) x Column(92)
[2020-01-21 16:56] [INFO] Initialisation complete
[2020-01-21 16:56] [INFO] Width: 960, Height: 1056
```

Crashes: STDERR, STDOUT
Font/Terminal size:

```
drew@muaddib ~ $ alacritty -vv
Created log file at "/tmp/Alacritty-19190.log"
[2020-01-21 16:56] [INFO] Welcome to Alacritty
[2020-01-21 16:56] [DEBUG] Estimated DPR: 1.5
[2020-01-21 16:56] [DEBUG] Estimated Cell Size: 13 x 26
[2020-01-21 16:56] [DEBUG] Estimated Dimensions: None
[2020-01-21 16:56] [INFO] Device pixel ratio: 1.5
[2020-01-21 16:56] [INFO] Initializing glyph cache...
[2020-01-21 16:56] [INFO] ... finished initializing glyph cache in 0.006373983s
[2020-01-21 16:56] [INFO] Cell Size: 13 x 26
[2020-01-21 16:56] [INFO] Padding: 0 x 0
[2020-01-21 16:56] [INFO] Width: 1200, Height: 900
[2020-01-21 16:56] [INFO] PTY Dimensions: Line(34) x Column(92)
[2020-01-21 16:56] [INFO] Initialisation complete
[2020-01-21 16:56] [DEBUG] New num_cols is 73 and num_lines is 40
[2020-01-21 16:56] [INFO] Width: 960, Height: 1056
```

[...]

PR #3241, opened 2020-01-21
Add debug log for missing config

chrisduerr:
We currently log whenever we fall back to the default config because of
an error in the config itself. We also log when the config was
successfully loaded and where it was loaded from. The only scenario
where no config related message is logged is when there is no
configuration file present.

Logging this case should make it easier to debug issues like #3240,
without requiring any knowledge from maintainers about this edgecase.

### DIFF
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -6,7 +6,7 @@ use std::path::PathBuf;
 
 #[cfg(windows)]
 use dirs;
-use log::{error, warn};
+use log::{error, warn, debug};
 use serde_yaml;
 #[cfg(not(windows))]
 use xdg;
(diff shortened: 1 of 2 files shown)


alacritty/src/config/mod.rs
```diff
@@ -121,6 +121,8 @@ pub fn installed_config() -> Option<PathBuf> {
                     return Some(fallback);
                 }
             }
+
+            debug!(target: LOG_TARGET_CONFIG, "No configuration file found");
```
kchibisov: I feel like it should be on `info` level, since configuration loaded is also on `info` level.

kchibisov:
Also, you should move logging there https://github.com/alacritty/alacritty/blob/767d59155a5d834d22ccb7c7d439b418ffad5404/alacritty/src/main.rs#L132 in `else` branch.

alacritty/src/config/mod.rs
chrisduerr: That's a great point, I just put it on the highest level that doesn't cause a UI error, but info makes more sense.

chrisduerr:
> Also, you should move logging there 

That's a good point, that should make it so it works on Windows too.

PR #3253, opened 2020-01-26
Switch to flamegraph-rs script

chrisduerr:
This cleans up the Alacritty scripts a bit by removing some of them
which are not recommended to be used anymore and switching from the
official FlameGraph tool to the more specialized Rust FlameGraph
implementation.


.gitignore
```diff
@@ -33,3 +33,7 @@ alacritty_*_source.tar.bz2
 *.msi
 *.wixobj
 *.wixpdb
+
+# Perf tools
+perf.data*
+flamegraph.svg
```
kchibisov: I feel like the file is mixed aligned or something like that, could you check that?

scripts/create-flamegraph.sh
```diff
@@ -3,36 +3,29 @@
...
+if [ ! -x "$(command -v cargo-flamegraph)" ]; then
+    echo "cargo-flamegraph not installed; installing ..."
+    cargo install flamegraph
+    installed_flamegraph=1
+fi
 
-# Tell users where the file is.
-echo "Flame graph created at: file://$(pwd)/flame-$TIME.svg"
+# Create flamegraph
+cargo flamegraph --bin=alacritty -- -e true
```
kchibisov: This is not what we want though. we should support passing arbitrary flags to `alacritty` and not just `-e true`, since this `flamegraph` is pointless, unless you want to debug startup, idk.

.gitignore
kchibisov: I hate this github viewer here. This is not an issue.

scripts/create-flamegraph.sh
chrisduerr: I didn't consider this, but it seems like a great idea. Now all parameters passed to the script are forwarded to Alacritty, that should be better.

.gitignore
```diff
@@ -33,3 +33,7 @@ alacritty_*_source.tar.bz2
 *.msi
```
kchibisov: ^ there is a legacy Flamegraph thing, should we delete it.

chrisduerr: You commented the wrong thing in GitHub, but yeah we probably should.

kchibisov: I can't comment on a line that is not part of a diff, so I added `^`  thing.

chrisduerr: Oh, right. Gotta love GitHub.

PR #3260, opened 2020-01-28
Remove `tabspaces` config option

chrisduerr:


chrisduerr:
We've had multiple user shoot themselves in the foot trying to change this, despite the clear warning in the documentation. Hiding it should reduce the risk of abuse, but raises the question if it is still needed at all.

While I've had some discussions with @kchibisov on IRC already, I'd like to know what you think @nixpulvis.

kchibisov:
I have a strong argument on why this thing should be renamed and completely hidden from config. Old value should be immediately removed and warn that it was removed and users should solve a puzzle to find a new name.  We should also don't even name it with a word "tabspaces" and just through it through the sha256 of some random thing, and put somewhere in a wiki through 3 or 5 links and 1 link should be through resource with a broken cert. Also fax for hidden options should be one more option.

https://github.com/search?q=filename%3Aalacritty.yml+tabspaces&type=Code

kchibisov:
In addition, this is an ncurses [file]https://raw.githubusercontent.com/mirror/ncurses/master/misc/terminfo.src) to generate all terminfo entries. You can search `it#4` in it and find out only 1 entry for some odd blackberry thing. So all entries have `it#4` in it.

kchibisov:
Hm, or maybe we can remove it from a config entirely and hide under a CLI flag. We can name it `--terminfo-it`, but on the other hand having a `terminal` section in config file could be nice. Especially if we want to add something like xterm does, with its OSC disabling [thing](https://invisible-island.net/xterm/manpage/xterm.html#VT100-Widget-Resources:allowWindowOps). It'll be nice to disable curtain things, like OSC 7 (If it'll land) or OSC 52.

nixpulvis:
Just two questions:

1. On Linux and macOS users should be able to use `tabs -4` or other means of changing this value, but what about on Windows?
2. If this should should be removed, shouldn't we remove it from the config altogether? Why leave the `Tabspaces` struct for example.

chrisduerr:
> On Linux and macOS users should be able to use tabs -4 or other means of changing this value, but what about on Windows?

Even on Windows terminfo is supported in some contexts, so it's the same as on Linux/BSD/macOS. Assuming you meant setting tabs to `4`, not `-4`.

> If this should should be removed, shouldn't we remove it from the config altogether? Why leave the Tabspaces struct for example.

I think there might be a bit of a misunderstanding, the goal was never to completely remove the option. It's only about removing the option in its current form.

nixpulvis:
@chrisduerr 

> I think there might be a bit of a misunderstanding, the goal was never to completely remove the option. It's only about removing the option in its current form.

Oh, I just assumed we wanted to remove the option, given the title of the PR, and no other information really. If we still want to give users the ability to set this, then what are our concerns? I'll be happy as long as I can at least set this via a tool like `tabs`, although as you've mentioned to me in IRC, this could still break peoples rendering for some applications (tabs are a mess, who knew). If properly fixing this involves a custom terminfo file, I'd be much less happy, though I don't know if it's possible to avoid.

> Assuming you meant setting tabs to 4, not -4.

Actually no haha, I didn't know you could just write explicit sequences of tabstops, `tabs 1,3,8,13` is a new concept to me :smiley:

chrisduerr:
> If properly fixing this involves a custom terminfo file to make our default and the terminfo aligned, I'd be much less happy, though I don't know if it's possible to avoid

It is not. Having a common understanding between terminal and application is necessary and terminfo is the only way of doing so.

kchibisov:
After thinking about this issue for a while I've come to following solution

- Remove `tabspaces` value entirely from config and throw an error on old setting.
- Add new section for terminal specific tweaks, for example we can ban `OSC` in it ( see #3386)
- Add a way to set `it` value and called it like that, however hide it inside some comments, so it won't remind of `tabspaces`.

The motivation behind removal is mostly this [link](https://github.com/search?q=filename%3Aalacritty.yml+tabspaces&type=Code). It won't break anything if you just remove it, since all terminfo's are using `it#8` and one who run their terminfo will smoothly update to a newer `it` setting. The idea behind new section makes sense due to collecting terminal specific tweaks, like banning `OSC 7`/`OSC 52`, and putting `it` into it makes sense, since folks should manually compile terminfo to use `tabspaces: 4`, so they are aware what this setting could do to them and what they should look for.

I don't like removing option entirely, since someone could actually run modified `terminfos`. Also there's a way to set proper `tabspaces` value from terminfo on startup. You should just query database and get `InitTabs` value and just use it. However I'm not sure how it could work on Windows. We're already doing database query to check for our terminfo, so getting `InitTabs` capability should there should also be fine.

nixpulvis:
@kchibisov I'm not sure we can merge this before we have a solid plan for setting the `it` value (which it seems like we're on a good path for). I personally would like to keep my terminal operating with 4 space tabs across platforms.

I don't want to add anything to my shell startup (like executing `tabs`), nor do I want to have to compile and install my own `terminfo`. My understanding is that we should be able to simply override the terminfo values inside our config, which will in essence create a fork of the underlying terminfo used.

Once this is the case, I'll be pretty happy with the removal of the setting, since there's a proper solution supported. I don't want to break peoples configs before they can take that moment to actually fix it.

kchibisov:
>I don't want to add anything to my shell startup (like executing tabs), nor do I want to have to compile and install my own terminfo. My understanding is that we should be able to simply override the terminfo values inside our config, which will in essence create a fork of the underlying terminfo used.

This is not how it works, you can't just override it. You should compile it with `4`, since applications are querying terminfo database, not your terminal emulator, so the only way to go is compile your own thing.

chrisduerr:
You should **never** run with a terminfo that doesn't match your `it` setting in Alacritty and Alacritty should **NEVER** modify the terminfo itself in some wonky startup magic creating random terminfo definitions.

If you do not like the default tab option, which you should basically never see anyways, then you *need* to create your own terminfo entry. There is no way around that and Alacritty shouldn't try to provide such a way.

chrisduerr:
I've created a separate PR that resolves some bugs related to tabstops, which I think make sense to fix before this:

https://github.com/alacritty/alacritty/pull/3398

I've also decided to completely get rid of this option, after researching other terminal emulators and finding out no other terminal exposes this option.

kchibisov:
>I've also decided to completely get rid of this option, after researching other terminal emulators and finding out no other terminal exposes this option.

But do they read your terminfo value? If so we should also do this.

chrisduerr:
> But do they read your terminfo it value? If so we should also do this.

No, not a single one did. I modified their default terminfo values, so replaced the `it` in `xterm-termite` with `it#4` for example and all of them behaved the exact same way as they did with the original terminfo.

kchibisov:
Oh, you already done that.

brian6932:
>Even on Windows terminfo is supported in some contexts

How exactly are Windows users supposed to set this? There's no `terminfo`, if you're referencing MSYS/WSL1, these are translation environments, let alone full on emus like WSL2, not something every user may (want to) use. Why remove something without checking if all supported platforms have an actual solution?

kchibisov:
> Why remove something without checking if all supported platforms have an actual solution?

I'd suggest to understand what it does before commenting.

brian6932:
Sets tab size in spaces, which can cause incompatibility with programs that also set tab size.

kchibisov:
No, this is not what it does. It's a tab size based on terminfo, which is completely different, only 4 terminfos have it set to 4, and these are some arcane stuff.

To set tab stops there's an escape sequence, which we support.

kchibisov:
That's what `tabs` uses `printf "\eH"`. Probable just moves the cursor and executes this escape where needed.

PR #3264, opened 2020-01-29
Fix backspace deleting chars when IME is open

### LINKED ISSUE #1606: macOS: unexpected characters get deleted when using backspace in Pinyin input method
Which operating system does the issue occur on? macOS 10.13

When input Chinese characters, if I use backspace to remove some letters ( as part of the Pinyin structure), it will get the characters which are just before the cursor of alacritty deleted (not what I intend to do). 
I have tested it with the input method in the gifs and macOS's default Pinyin input methods, and with the fresh installed alacritty on other people's maxOS (also 10.13), all reproduced the case.

gifs for different behaviors of alacritty and iTerm2.
alacritty
iterm2

chrisduerr:
Fixes #1606.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Resize of bitmap fonts
 - Crash when using bitmap font with `embeddedbitmap` set to `false`
 - Inconsistent fontconfig fallback
+- Backspace deleting characters while IME is open on macOS
 
 ### Removed
 
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -287,7 +287,6 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
         PageDown; Action::Esc("\x1b[6~".into());
         PageDown, ModifiersState::SHIFT, +TermMode::ALT_SCREEN; Action::Esc("\x1b[6;2~".into());
         Tab,  ModifiersState::SHIFT; Action::Esc("\x1b[Z".into());
-        Back; Action::Esc("\x7f".into());
         Back, ModifiersState::ALT; Action::Esc("\x1b\x7f".into());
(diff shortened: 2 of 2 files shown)


kchibisov:
I'm wondering why Alacritty receiving characters when you're typing in IME window, it's seems to me like winit bug tbh, however I'm not familiar with macOS IME and it's fine to remove binding for a platform if it doesn't make any sense.

chrisduerr:
> I'm wondering why Alacritty receiving characters when you're typing in IME window, it's seems to me like winit bug tbh

If you read the last few comments of the issue, you'll notice that we don't. What we do receive is keyboard input events, but no `ReceivedCharacter` events.

The problem is just that we map the \x7F character to backspace in our keybindings by default, since Linux would otherwise send \x08. But I've tested this on macOS and it already sends \x7F anyways, so that binding is not required here.

kchibisov:
I mean they shouldn't be our window event, if you type backspace in a different window on X11, you'll see `Backspace`, but you wont execute anything. So in theory every binding to IME could trigger binding in alacritty, right? IIRC this is not true for e.g. X11 IME.

chrisduerr:
> I mean they shouldn't be our window event, if you type backspace in a different window on X11, you'll see Backspace, but you wont execute anything.

You won't receive a WindowEvent though, but a DeviceEvent. That concept does not exist in this fashion on macOS (or Wayland actually).

kchibisov:
Oh, I've got why iterm works, it's because it's entering chars in terminal while typing in IME, so backspace on IME is a backspace in terminal and it works fine. However we don't do that and we're just missing chars... So we should likely do what iterm is doing and just enter chars into terminal.

chrisduerr:
We're not receiving any characters though. Winit only forwards key bindings, not characters.

So this just fixes the problem, no need to over-complicate things?

kchibisov:
I know, I just found out an explanation why iterm works fine here, since I do remember that it has binding for backspace.

PR #3270, opened 2020-01-30
Fix handling of OpenType variable fonts

### LINKED ISSUE #3257: Bold font is used for normal font
Hi,

alacritty picks up the wrong font for the normal font: what I have configured as bold font is used as regular font and the `normal` part of the font configuration is ignored.

Cheers.

### System

OS: Linux (Debian Sid(
Version: alacritty 0.4.2-dev (6b327b6)
Linux/BSD: X11, i3

### Logs

### Font part of config

```
# Font configuration
font:
  # Normal (roman) font face
  normal:
    # Font family
    #
    # Default:
    #   - (macOS) Menlo
    #   - (Linux/BSD) monospace
    #   - (Windows) Consolas
    family: Source Code Variable

    # The `style` can be specified to pick a specific face.
    style: Semibold

  # Bold font face
  bold:
    # Font family
    #
    # If the bold family is not specified, it will fall back to the
    # value specified for the normal font.
    family: Source Code Variable

    # The `style` can be specified to pick a specific face.
    style: Bold

  # Italic font face
  italic:
    # Font family
    #
    # If the italic family is not specified, it will fall back to the
    # value specified for the normal font.
    family: Source Code Variable

    # The `style` can be specified to pick a specific face.
    style: Semibold Italic

  # Bold italic font face
  bold_italic:
    # Font family
    #
    # If the bold italic family is not specified, it will fall back to the
    # value specified for the normal font.
    family: Source Code Variable

    # The `style` can be specified to pick a specific face.
    style: Bold Italic

  # Point size
  size: 6.5
```

kchibisov:
Fixes #3257.

### DIFF
--- a/font/src/ft/fc/pattern.rs
+++ b/font/src/ft/fc/pattern.rs
@@ -23,7 +23,7 @@ use libc::{c_char, c_double, c_int};
 
 use super::ffi::FcResultMatch;
 use super::ffi::{FcBool, FcFontRenderPrepare, FcPatternGetBool, FcPatternGetDouble};
-use super::ffi::{FcChar8, FcConfigSubstitute, FcDefaultSubstitute, FcPattern};
+use super::ffi::{FcChar8, FcConfigSubstitute, FcDefaultSubstitute, FcPattern, FcPatternHash};
 use super::ffi::{FcPatternAddCharSet, FcPatternDestroy, FcPatternDuplicate, FcPatternGetCharSet};
 use super::ffi::{FcPatternAddDouble, FcPatternAddString, FcPatternCreate, FcPatternGetString};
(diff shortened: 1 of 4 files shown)


font/src/ft/fc/pattern.rs
```diff
@@ -537,6 +540,10 @@ impl PatternRef {
         }
     }
 
+    pub fn hash(&self) -> PatternHash {
+        unsafe { PatternHash(FcPatternHash(self.as_ptr())) }
+    }
```
chrisduerr: A bit unfortunate that Rust's `Hash` function doesn't just return a u32 or something and we can use that ourselves.

But I suppose this should work just fine.

Issue #2990, opened 2019-11-18
Missing dependency for Clear Linux install instructions


a-lakhani:
Which operating system does the issue occur on?

Clear Linux

If on linux, are you using X11 or Wayland?

X11

Clear Linux users run into Fontconfig errors on startup unless we [sudo swupd bundle-add devpkg-fontconfig](https://github.com/jwilm/alacritty/issues/2675#issuecomment-515274325) and [modify our FONTCONFIG_PATH](https://github.com/clearlinux/distribution/issues/639#issuecomment-495972657) since the default location for `fonts/fonts.conf` is under `/usr/share/defaults/`, not `/etc/`.

The devpkg-fontconfig bundle includes devpkg-freetype and devpkg-expat, so the swupd command in INSTALL.md should read:

`sudo swupd bundle-add devpkg-fontconfig devpkg-libxcb`

as well as include the instruction to create a `/etc/profile` containing `FONTCONFIG_PATH=/usr/share/defaults/fonts'

before building alacritty.

I had the same fontconfig errors mentioned in the referenced issues, and the above fixed my build on Clear Linux as well.

kchibisov:
@a-lakhani does Clear Linux not provide this `FONTCONFIG_PATH` env variable by default. It seems strange that it's in unusual location by default and don't have this EXPORT.

Could you send a PR and fill this gap of our missing dependencies on Clear Linux, thx?

chrisduerr:
Waiting on https://github.com/trimental/andrew/issues/3.

kchibisov:
Since andrew is no longer used closing this.

PR #3275, opened 2020-01-31
Fixed Clear Linux installation instructions

a-lakhani:
Changed the swupd bundle-add command to include missing dependencies, and added instructions to fix issues with finding fonts.conf.

In response to #2990.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Inconsistent fontconfig fallback
 - Backspace deleting characters while IME is open on macOS
 - Handling of OpenType variable fonts
-- Incomplete instructions for compiling Alacritty from source on Clear Linux
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Inconsistent fontconfig fallback
 - Backspace deleting characters while IME is open on macOS
 - Handling of OpenType variable fonts
+- Incomplete instructions for compiling Alacritty from source on Clear Linux
 
 ### Removed
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -190,13 +190,28 @@ installed as well as [Clang 3.9 or greater](http://releases.llvm.org/download.ht
 
 #### Clear Linux
 
-On Clear Linux, you need a few extra libraries to build Alacritty. Here's a
-`swupd` command that should install all of them. If something is still found
-to be missing, please open an issue.
+On Clear Linux, a few extra steps are needed to build Alacritty. 
+
+1. Here's a `swupd` command that should install all missing dependencies. If something is still found to be missing, please open an issue.
 
 ```sh
-swupd bundle-add devpkg-expat devpkg-freetype devpkg-libxcb
+swupd bundle-add devpkg-fontconfig devpkg-libxcb
 ```
+
+2. Clear Linux stores fonts.conf in a non-standard location due to its [stateless](https://github.com/clearlinux/clr-man-pages/blob/master/stateless.7.rst) design. 
+
+* Either set the FONTCONFIG_PATH environment variable as follows:
+
+```
+echo "FONTCONFIG_PATH=/usr/share/defaults/fonts" >> /etc/profile
+```
+
+* or create a symbolic link to the location of fonts.conf:
+
+```
+ln -s /usr/share/defaults/fonts /etc/fonts
+```
+
 
 #### Other
 


INSTALL.md
```diff
@@ -190,13 +190,28 @@ installed as well as [Clang 3.9 or greater](http://releases.llvm.org/download.ht
...
-On Clear Linux, you need a few extra libraries to build Alacritty. Here's a
-`swupd` command that should install all of them. If something is still found
-to be missing, please open an issue.
+On Clear Linux, a few extra steps are needed to build Alacritty. 
+
+1. Here's a `swupd` command that should install all missing dependencies. If something is still found to be missing, please open an issue.
 
 ```sh
-swupd bundle-add devpkg-expat devpkg-freetype devpkg-libxcb
+swupd bundle-add devpkg-fontconfig devpkg-libxcb
```
kchibisov: we actually need freetype, are you sure that it's not required? Also some our packages do require python to build, IIRC, you can take a look on deps from other distros.

INSTALL.md
```diff
@@ -190,13 +190,28 @@ installed as well as [Clang 3.9 or greater](http://releases.llvm.org/download.ht
...
+
+2. Clear Linux stores fonts.conf in a non-standard location due to its [stateless](https://github.com/clearlinux/clr-man-pages/blob/master/stateless.7.rst) design. 
+
+* Either set the FONTCONFIG_PATH environment variable as follows:
+
 ```
+echo "FONTCONFIG_PATH=/usr/share/defaults/fonts" >> /etc/profile
+```
+
+* or create a symbolic link to the location of fonts.conf:
```
kchibisov: I'm not sure about bullet list here tbh. What do you think @chrisduerr ?

INSTALL.md
a-lakhani: devpkg-freetype and devpkg-expat are both included in devpkg-fontconfig.

I don't know what else in devpkg-fontconfig is needed in addition to freetype and expat, just that I needed to install the fontconfig bundle to successfully compile from source.

Python is in the desktop-apps bundle. I don't remember if that is part of the recommended installation configuration, but it could be why I did not need to install python when compiling Alacritty.

kchibisov: I think that this part should be fine the way it is right now then.

INSTALL.md
a-lakhani: Is either method less likely to cause issues for other programs? I agree the list adds ambiguity, I just didn't know which one would be better to recommend.

kchibisov: I'm taking about markdown style issue. using `*` creates a bullet list, so I don't really like how it looks like tbh.

chrisduerr: The entire section looks a bit off, since it's all written as a single big list, but that doesn't really make much sense.

The `Here's a swup command` comment does not fit in with the rest of the documentation and thus should be adjusted. `The following swup command` would be more appropriate.

The entire numbering and sub-lists just create unnecessary indentation that make the text hard to read. We're already listing a bunch of things that all need to be executed and there is no ambiguity about the order of operations, so we don't need to use lists for this.

It also seems like this is something that should be fixed in clear linux, not Alacritty. Even with their 'stateless design', it pretty clearly seems broken? We should be fine just recommending the symlink though, I don't think exporting environment variables is following the goals of the Clear Linux project?

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Inconsistent fontconfig fallback
 - Backspace deleting characters while IME is open on macOS
 - Handling of OpenType variable fonts
+- Incomplete instructions for compiling Alacritty from source on Clear Linux
```
chrisduerr: Since this has no release impact, we do not need a changelog entry.

```suggestion
```

INSTALL.md
a-lakhani: When [an issue](https://github.com/clearlinux/distribution/issues/639) was opened with Clear, one of their devs said 
> Does it search for that file directly, not using libfontconfig? If so, it needs a patch to operate properly on Clear.

I'm not taking a side on that, just sharing in case it is a more general solution.

INSTALL.md
a-lakhani: Should I make another commit reverting the change or will you just not merge this commit?

INSTALL.md
chrisduerr: > Does it search for that file directly, not using libfontconfig

I'm pretty sure the answer to that should be `no`. We call `FcConfigGetCurrent` to get the config, we don't load the file ourselves.

This seems like something that should be fixed, rather than worked around.

kchibisov: I feel like Clear linux passes this thing internally to bundles apps or something like that, since entire /etc is empty on it IIRC. So I adding such things into /etc/profile is a way to go for users and even a recommended way. I could be wrong about "recommended", but I heard something like that in the past.

kchibisov: and Alacritty is just calling to `libfontconfig.so` APIs, it doens't do any parsing of config files itself, etc, etc.

chrisduerr: > So I adding such things into /etc/profile is a way to go for users and even a recommended way. I could be wrong about "recommended", but I heard something like that in the past.

It seems like that should be done for the system then. The comment on the issue makes it sound like the distro is expected to work by default, which it clearly does not then.

So it seems to me like clear linux has got some figuring out to do? Otherwise I'm very curious what their recommended strategy would be?

chrisduerr:
This should be fixed, rather than documented. See https://github.com/trimental/andrew/issues/3.

PR #3277, opened 2020-02-01
Fix non-scalable font resize issue

kchibisov:


font/src/ft/mod.rs
```diff
@@ -38,16 +38,25 @@ struct FixedSize {
...
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Copy, Clone, Hash)]
+struct FontHash(u64);
+
+impl FontHash {
+    fn new(high: PatternHash, low: PatternHash) -> Self {
+        let hash: u64 = low.0 as u64 | ((high.0 as u64) << 32);
+        Self(hash)
```
chrisduerr: We can probably just go for `Self(low.0 as u64 | ((high.0 as u64) << 32))`?

font/src/ft/mod.rs
```diff
@@ -213,16 +222,25 @@ impl FreeTypeRasterizer {
...
+                pattern.set_slant(slant.into_fontconfig_type());
             },
             Style::Specific(ref style) => {
                 // If a name was specified, try and load specifically that font.
-                self.get_specific_face(&desc, &style)
+                pattern.add_style(style);
             },
         }
+
+        self.query_font(pattern, hash, desc)
```
chrisduerr: If I understand this correctly, this is just to deduplicate code, right? Seems like we just get rid of two functions that do not actually need to be separate functions.

font/src/ft/mod.rs
```diff
@@ -213,16 +222,25 @@ impl FreeTypeRasterizer {
         let size = Size::new(size.as_f32_pts() * self.device_pixel_ratio * 96. / 72.);
         self.pixel_size = f64::from(size.as_f32_pts());
 
+        let mut pattern = Pattern::new();
+        pattern.add_family(&desc.name);
+        pattern.add_pixelsize(self.pixel_size);
+
+        let hash = pattern.hash();
```
chrisduerr: Do we need to hash inside of here and then pass it? We pass the pattern anyways, right, so it would just remove one parameter to hash inside of `query_font` without any drawbacks?

font/src/ft/mod.rs
```diff
@@ -238,42 +256,25 @@ impl FreeTypeRasterizer {
...
 
-        let mut font_iter = fonts.into_iter();
+        let mut pattern_iter = patterns.into_iter();
 
-        let base_font = font_iter.next().ok_or_else(|| Error::MissingFont(desc.to_owned()))?;
-        let base_font = pattern.render_prepare(config, base_font);
+        let base_pattern =
+            pattern_iter.next().ok_or_else(|| Error::MissingFont(desc.to_owned()))?;
+        let base_pattern = pattern.render_prepare(config, base_pattern);
+        let base_hash = FontHash::new(hash, base_pattern.hash());
```
chrisduerr: Do we hash the unprepared and prepared pattern together here? That seems like a very strange choice. Is there any benefit to doing that here, or is it only relevant for fallback fonts?

font/src/ft/mod.rs
```diff
@@ -293,19 +294,22 @@ impl FreeTypeRasterizer {
...
-            .map(|font| {
-                let charset = font.get_charset().unwrap_or(&empty_charset);
+        let list: Vec<FallbackFont> = pattern_iter
+            .map(|fallback_pattern| {
+                let charset = fallback_pattern.get_charset().unwrap_or(&empty_charset);
+                let pattern = base_pattern.render_prepare(config, fallback_pattern);
+                let hash = FontHash::new(hash, pattern.hash());
                 let _ = coverage.merge(&charset);
-                FallbackFont::new(font.to_owned())
+                FallbackFont::new(pattern, hash)
```
chrisduerr: We've discussed on IRC that all we'd need to add to the hash would be the font size. What is the reason to hash the two different patterns here instead? Is it to include other settings that might be different and relevant, if so, what would those be?

font/src/ft/mod.rs
```diff
@@ -618,8 +622,9 @@ impl FreeTypeRasterizer {
         }
 
         for fallback_font in &fallback_list.list {
+            let font_hash = fallback_font.hash;
             let font_pattern = &fallback_font.pattern;
```
chrisduerr: Why do we extract these two variables here? Is there any specific reason like a linebreak or something that we prevent by putting things here? Or would this work just as well by inlining these two variables?

font/src/ft/mod.rs
kchibisov: I don't want to use u64 hashes from rust, since XOR of 2 u32 hashes from Fontconfig should be already enough for our needs, I'm a bit afraid of collisions tbh, since we're not doing a "good" thing here. We can just use 2 strings(path + style + pixelsize) and hash by them, however I'm not sure about the performance of it.

font/src/ft/mod.rs
kchibisov: borrow checker.

font/src/ft/mod.rs
kchibisov: it's basically hashing of requested (size + family) and rendered thing, where size could be different to requested one.

font/src/ft/mod.rs
```diff
@@ -389,8 +397,7 @@ impl FreeTypeRasterizer {
         if use_initial_face {
             Ok(glyph_key.font_key)
         } else {
-            let key = self.load_face_with_glyph(glyph_key).unwrap_or(glyph_key.font_key);
-            Ok(key)
+            Ok(self.load_face_with_glyph(glyph_key).unwrap_or(glyph_key.font_key))
         }
     }
```
chrisduerr: Honestly, this entire method is a bloody mess. It has a `has_recursed` parameter, without ever being able to recurse, which is always set to `false` when called.

Then it branches two times, the first time just sets a variable to the status of that branching operation, while the second time checks that to branch on the same thing again.

Would you mind removing the parameter, changing it to `if index != 0 { return Ok(glyph_key.font_key); }` and then just putting the `Ok(self.load_face_with_glyph(glyph_key).unwrap_or(glyph_key.font_key))` at the end of it without any additional conditionals? Can get rid of the pointless `else` too.

font/src/ft/mod.rs
```diff
@@ -213,75 +222,56 @@ impl FreeTypeRasterizer {
...
+        // We should render patterns to get values like `pixelsizefixupfactor`
+        let primary_font = pattern.render_prepare(config, primary_font);
 
-        let mut font_iter = fonts.into_iter();
+        // Compute ID for font identification, we're using hashes from requested pattern
+        // and a rendered one, and combining both with XOR to get a resulted ID. Some fonts
+        // always have the same size after pattern matching match, so if user requests pattern
+        // with size `X`, resulted font could always have `size = const`, which leads to
+        // constant result from e.g. `primary_font.hash()`, so we need to use requested
+        // `pixelsize` in the hash computation in addition to the hash of rendered pattern
```
chrisduerr: I think I've asked you this before, but what happens if we set the font size for a font which loses the size during pattern prepare?

If we set it for one that doesn't lose it, it shouldn't make any difference, right?

What happens if we manually set it again after preparing it for all fonts?

font/src/ft/mod.rs
```diff
@@ -38,13 +38,22 @@ struct FixedSize {
...
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Copy, Clone, Hash)]
+struct FontID(u32);
+
+impl FontID {
+    fn new(lhs: PatternHash, rhs: PatternHash) -> Self {
+        // XOR two hashes to get a font ID
+        Self(((lhs.0 << 1) | (lhs.0 >> 31)) ^ rhs.0)
```
chrisduerr: Could you provide some (~3-5) pattern hashes, just to see what they look like? I'm curious if there's any kind of pattern recognizable, that would indicate that these aren't evenly distributed.

font/src/ft/mod.rs
kchibisov: I've done it on refactoring PR, let me do this here one more time.

font/src/ft/mod.rs
kchibisov: >I think I've asked you this before, but what happens if we set the font size for a font which loses the size during pattern prepare?

size is always with *us*, you can't just lose it. The thing is that some fonts have constant size. So if you request a size 12 for font "Emoji" which has fixed size 108, you'll get size 108 and a fixupfactor to scale thing. For scaled fonts you're getting a size more realistic to what you request and it based on your request of 12.

>If we set it for one that doesn't lose it, it shouldn't make any difference, right?

Again, you're not losing anything, you're just always getting a constant size no matter what you try to request from it.

>What happens if we manually set it again after preparing it for all fonts?

If you manually set size to a already rendered pattern, it'll blow up.

font/src/ft/mod.rs
kchibisov: Do you want pattern hashes or resulted FontIDs? If you're curious on pattern.hash() thing https://gitlab.freedesktop.org/fontconfig/fontconfig/blob/master/src/fcpat.c#L601 , so basically some fast XOR hash. The idea of my computation was from it.

font/src/ft/mod.rs
chrisduerr: A u32 has about 4,294,967,295 different possible values. Assuming an even distribution of font hashes across this, it seems like collisions would be pretty unlikely?

It seems like we'd basically half the precision by trying to combine two different hashes that are potentially in conflict, but that should still leave us with plenty of headroom.

The most problematic thing would be if the input hashes were not evenly distributed. XORing the values should be okay-ish, assuming the two hashes are not identical. If it is possible that hashes could be identical, that would cause big problems since we'd get a lot of `0` hashes. It seems like a lot of people recommend `m * hash(a) + hash(b)` over XOR.

font/src/ft/mod.rs
chrisduerr: > If you manually set size to a already rendered pattern, it'll blow up.

Huh, interesting. Mind telling me how it blows up? Will it just segfault?

font/src/ft/mod.rs
kchibisov: where `m` is a prime number, I'd assume? As you can see those bitshifts are here not as decorations :), so I'm trying to prevent us from such thing.

font/src/ft/mod.rs
chrisduerr: I was only curious about pattern hashes.

font/src/ft/mod.rs
kchibisov: No, it messes up your "computed" fields, etc. So you'll get "EXTRA BIG EMOJIS" for example, since you're modifying the returned `size` for your previous request. Like if you force pixelsize of 12 to a thing with constant size 108, your fixup factor would be based off your forced value, however underlining bitmap metrics are completely different (they match this constant size).

font/src/ft/mod.rs
kchibisov: Sure.

```
name: monospace, size: 9, hash: 92e9
Left hash 92e9, right hash c943209b
Left hash 92e9, right hash 29bc3ae0
Left hash 92e9, right hash 4569a792
Left hash 92e9, right hash 8e1b9f9f
Left hash 92e9, right hash 2a4fac1
Left hash 92e9, right hash a5f3b374
Left hash 92e9, right hash 66eadb05
Left hash 92e9, right hash c785e2d2
Left hash 92e9, right hash d7d1d80f
Left hash 92e9, right hash 626fe89c
Left hash 92e9, right hash 1c7e06b7
Left hash 92e9, right hash 75a316c0
Left hash 92e9, right hash 8498d10e
Left hash 92e9, right hash eba29dda
name: monospace, size: 9, hash: 92e9
Left hash 92e9, right hash c943209b
Left hash 92e9, right hash 29bc3ae0
Left hash 92e9, right hash 4569a792
Left hash 92e9, right hash 8e1b9f9f
Left hash 92e9, right hash 2a4fac1
Left hash 92e9, right hash a5f3b374
Left hash 92e9, right hash 66eadb05
Left hash 92e9, right hash c785e2d2
Left hash 92e9, right hash d7d1d80f
Left hash 92e9, right hash 626fe89c
Left hash 92e9, right hash 1c7e06b7
Left hash 92e9, right hash 75a316c0
Left hash 92e9, right hash 8498d10e
Left hash 92e9, right hash eba29dda
name: monospace, size: 9, hash: 92e9
Left hash 92e9, right hash f648c6c5
Left hash 92e9, right hash 741a66ba
Left hash 92e9, right hash 369ef7f4
Left hash 92e9, right hash 8eba312e
Left hash 92e9, right hash b18844d5
Left hash 92e9, right hash 7d572b94
Left hash 92e9, right hash 2948939d
Left hash 92e9, right hash cae4ad28
Left hash 92e9, right hash f3d4b5c8
Left hash 92e9, right hash eba29dda
Left hash 92e9, right hash 3aa7f96d
Left hash 92e9, right hash ae59d18f
Left hash 92e9, right hash a01d91d6
Left hash 92e9, right hash 87097b8d
Left hash 92e9, right hash ef8cb96a
name: monospace, size: 9, hash: 92e9
Left hash 92e9, right hash 20357cbb
Left hash 92e9, right hash 651a66ba
Left hash 92e9, right hash 8e1b9f9f
Left hash 92e9, right hash 644925a6
Left hash 92e9, right hash f76914d1
Left hash 92e9, right hash 67d1bd45
Left hash 92e9, right hash d394cf45
Left hash 92e9, right hash 4470361e
Left hash 92e9, right hash 3bad9911
Left hash 92e9, right hash e4f2d942
Left hash 92e9, right hash e65e85ad
Left hash 92e9, right hash 7751da1b
Left hash 92e9, right hash 18aa4c9c
Left hash 92e9, right hash 506d304d
name: monospace, size: 9, hash: 92e9
Left hash 92e9, right hash 27c79c84
Left hash 92e9, right hash 8756de0e
Left hash 92e9, right hash 2c9591ba
Left hash 92e9, right hash db3b3546
Left hash 92e9, right hash b18844d5
Left hash 92e9, right hash 9eb05b43
Left hash 92e9, right hash aa391d55
Left hash 92e9, right hash e098f591
Left hash 92e9, right hash b6964309
Left hash 92e9, right hash d5fca1bb
Left hash 92e9, right hash 6451a6c8
Left hash 92e9, right hash daf09927
Left hash 92e9, right hash df2096d9
Left hash 92e9, right hash a4e781b4
Left hash 92e9, right hash 85893641
Left hash 92e9, right hash 506d304d
name: monospace, size: 18.5, hash: 92f2
Left hash 92f2, right hash c94ea09b
Left hash 92f2, right hash 29bc3ae0
Left hash 92f2, right hash 4572a792
Left hash 92f2, right hash 8e009f9f
Left hash 92f2, right hash 2a97ac1
Left hash 92f2, right hash a5e8b374
Left hash 92f2, right hash 66f1db05
Left hash 92f2, right hash c79ee2d2
Left hash 92f2, right hash d7cad80f
Left hash 92f2, right hash 6274e89c
Left hash 92f2, right hash 1c6506b7
Left hash 92f2, right hash 75b816c0
Left hash 92f2, right hash 8483d10e
Left hash 92f2, right hash ebb99dda
name: monospace, size: 18.5, hash: 92f2
Left hash 92f2, right hash f64546c5
Left hash 92f2, right hash 741a66ba
Left hash 92f2, right hash 3685f7f4
Left hash 92f2, right hash 8e8c312e
Left hash 92f2, right hash b1be44d5
Left hash 92f2, right hash 7d5aab94
Left hash 92f2, right hash 2953939d
Left hash 92f2, right hash caffad28
Left hash 92f2, right hash f3cfb5c8
Left hash 92f2, right hash ebb99dda
Left hash 92f2, right hash 3abcf96d
Left hash 92f2, right hash ae42d18f
Left hash 92f2, right hash a00691d6
Left hash 92f2, right hash 873f7b8d
Left hash 92f2, right hash efbab96a
name: monospace, size: 18.5, hash: 92f2
Left hash 92f2, right hash 2038fcbb
Left hash 92f2, right hash 651a66ba
Left hash 92f2, right hash 8e009f9f
Left hash 92f2, right hash 642525a6
Left hash 92f2, right hash f75f14d1
Left hash 92f2, right hash 67cabd45
Left hash 92f2, right hash d3f8cf45
Left hash 92f2, right hash 441c361e
Left hash 92f2, right hash 3bc19911
Left hash 92f2, right hash e49ed942
Left hash 92f2, right hash e63285ad
Left hash 92f2, right hash 773dda1b
Left hash 92f2, right hash 18c64c9c
Left hash 92f2, right hash 5001304d
name: monospace, size: 18.5, hash: 92f2
Left hash 92f2, right hash 27ca1c84
Left hash 92f2, right hash 8756de0e
Left hash 92f2, right hash 2c8e91ba
Left hash 92f2, right hash dbe33546
Left hash 92f2, right hash b1be44d5
Left hash 92f2, right hash 9ebddb43
Left hash 92f2, right hash aa551d55
Left hash 92f2, right hash e083f591
Left hash 92f2, right hash b64e4309
Left hash 92f2, right hash d524a1bb
Left hash 92f2, right hash 6489a6c8
Left hash 92f2, right hash da289927
Left hash 92f2, right hash dff896d9
Left hash 92f2, right hash a43f81b4
Left hash 92f2, right hash 85513641
Left hash 92f2, right hash 5001304d

```

```suggestion
        Self(lhs.0.rotate_left(1) ^ rhs.0)
```

font/src/ft/mod.rs
```diff
@@ -213,75 +222,56 @@ impl FreeTypeRasterizer {
...
-    fn query_font(&mut self, pattern: Pattern, desc: &FontDesc) -> Result<FontKey, Error> {
-        let config = fc::Config::get_current();
-        let fonts = fc::font_sort(&config, &mut pattern.clone())
-            .ok_or_else(|| Error::MissingFont(desc.to_owned()))?;
+        // Compute ID for font identification, we're using hashes from requested pattern
+        // and a rendered one, and combining both with XOR to get a resulted ID. Some fonts
+        // always have the same size after pattern matching match, so if user requests pattern
+        // with size `X`, resulted font could always have `size = const`, which leads to
+        // constant result from e.g. `primary_font.hash()`, so we need to use requested
+        // `pixelsize` in the hash computation in addition to the hash of rendered pattern
```
chrisduerr: The location and length of this comment aren't quite right I'd say.  For the primary font this does not apply, right?

So I'd move it to the fallback fonts and shorten it a bit.

Maybe something like this should be sufficient:

```
// Hash fallback pattern together with request pattern to include requested font size in the hash
```

font/src/ft/mod.rs
```diff
@@ -293,19 +283,24 @@ impl FreeTypeRasterizer {
         };
 
         // Reuse the font_key, since changing it can break library users
-        let font_key = self.face_from_pattern(&base_font, font_key).and_then(|pattern| {
-            pattern.map(Ok).unwrap_or_else(|| Err(Error::MissingFont(desc.to_owned())))
-        })?;
+        let font_key = self.face_from_pattern(&primary_font, primary_font_id, font_key).and_then(
+            |pattern| pattern.map(Ok).unwrap_or_else(|| Err(Error::MissingFont(desc.to_owned()))),
+        )?;
```
chrisduerr: If you want to `?` only once, you should do it like this:

```
        let font_key = self
            .face_from_pattern(&primary_font, primary_font_id, font_key)
            .and_then(|pattern| pattern.ok_or_else(|| Error::MissingFont(desc.to_owned())))?;
```

If we don't care about that, this is another option:

```
        let font_key = self
            .face_from_pattern(&primary_font, primary_font_id, font_key)?
            .ok_or_else(|| Error::MissingFont(desc.to_owned()))?;
```

font/src/ft/mod.rs
```diff
@@ -293,19 +283,24 @@ impl FreeTypeRasterizer {
...
-        // Load fallback list
-        let list: Vec<FallbackFont> = font_iter
-            .map(|font| {
-                let charset = font.get_charset().unwrap_or(&empty_charset);
+
+        // Build fallback list
+        let list: Vec<FallbackFont> = matched_fonts
+            .map(|fallback_pattern| {
+                let charset = fallback_pattern.get_charset().unwrap_or(&empty_charset);
+                let fallback_font = primary_font.render_prepare(config, fallback_pattern);
```
chrisduerr: I don't think we should name patterns `font`. If we name it pattern, everyone immediately knows what we're talking about.

font/src/ft/mod.rs
kchibisov: It applies to all fonts, not only fallback ones, since you can request bitmap font.

font/src/ft/mod.rs
```diff
@@ -213,75 +222,50 @@ impl FreeTypeRasterizer {
...
-        pattern.add_pixelsize(self.pixel_size);
-
-        self.query_font(pattern, desc)
-    }
-
-    fn query_font(&mut self, pattern: Pattern, desc: &FontDesc) -> Result<FontKey, Error> {
-        let config = fc::Config::get_current();
-        let fonts = fc::font_sort(&config, &mut pattern.clone())
+        // Match pattern and get a font list, use first font as a "primary" one and the rest for a
+        // font fallback
```
chrisduerr: The double line comment here just bothered me slightly, so here's a proposed fix:

```
        // Get font list using pattern; first font is the primary while the rest are fallback
```

That doesn't linebreak anymore and contains the exact same information.

PR #3280, opened 2020-02-02
Start window invisible on Windows

kchibisov:
It should make our Windows window creation follow other backends and start window invisible. However it was started visible on purpose, so I'm not sure whether it breaks something or not.

Could Windows folks test whether it breaks anything for them or not?

### DIFF
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -252,7 +252,7 @@ impl Window {
 
         WindowBuilder::new()
             .with_title(title)
-            .with_visible(true)
+            .with_visible(false)
             .with_decorations(decorations)
             .with_transparent(true)
             .with_maximized(window_config.startup_mode() == StartupMode::Maximized)


cole-h:
I don't see any issues on my Windows VM: `Version 1909 (OS Build 18363.592)`.

chrisduerr:
@cole-h If you change [this line](https://github.com/alacritty/alacritty/blob/5f6a4334eae864f55a7bb075783c5d00f9beab0e/alacritty/src/window.rs#L144) to unwrap, does it crash for you?

kchibisov:
I'm not sure about a CHANGELOG for this thing. Do we need something for this change?

chrisduerr:
> I'm not sure about a CHANGELOG for this thing. Do we need something for this change?

I don't think we should, we just change some internal thing for consistency. So this will probably be fine. I didn't notice any difference myself.

PR #3282, opened 2020-02-02
Fix selection expansion across full-width glyphs

### LINKED ISSUE #3106: Expand full-width glyphs individually instead of relying on selection start/end
Currently when a selection starts/ends with a zero-width glyph, expansion is done based on that. However with things like block selection this does not work.

Instead of checking for expansion at the start/end of the selection, we should just have the selection as a fixed span from start/end of two points and then whenever a full-width glyph is selected, we should automatically apply that selection to the spacer glyphs too.

chrisduerr:
Instead of trying to expand the start and end of a selection across
full-width glyphs, the selection should now only go from its origin to
the end without any kind of expansion.

Instead, the expansion is now done where the cells are actually checked
for their selection status, expanding across the entire full-width glyph
whenever any part of it is selected.

Fixes #3106.

### DIFF
--- a/alacritty_terminal/src/index.rs
+++ b/alacritty_terminal/src/index.rs
@@ -44,15 +44,13 @@ impl<L> Point<L> {
 
     #[inline]
     #[must_use = "this returns the result of the operation, without modifying the original"]
-    pub fn sub(mut self, num_cols: usize, rhs: usize, absolute_indexing: bool) -> Point<L>
+    pub fn sub(mut self, num_cols: usize, rhs: usize) -> Point<L>
     where
         L: Copy + Default + Into<Line> + Add<usize, Output = L> + Sub<usize, Output = L>,
     {
         let line_changes =
             f32::ceil(rhs.saturating_sub(self.col.0) as f32 / num_cols as f32) as usize;
-        if absolute_indexing {
-            self.line = self.line + line_changes;
-        } else if self.line.into() > Line(line_changes) {
+        if self.line.into() > Line(line_changes) {
             self.line = self.line - line_changes;
         } else {
             self.line = Default::default();
@@ -63,18 +61,11 @@ impl<L> Point<L> {
 
     #[inline]
     #[must_use = "this returns the result of the operation, without modifying the original"]
-    pub fn add(mut self, num_cols: usize, rhs: usize, absolute_indexing: bool) -> Point<L>
+    pub fn add(mut self, num_cols: usize, rhs: usize) -> Point<L>
     where
-        L: Copy + Default + Into<Line> + Add<usize, Output = L> + Sub<usize, Output = L>,
+        L: Add<usize, Output = L> + Sub<usize, Output = L>,
     {
-        let line_changes = (rhs + self.col.0) / num_cols;
-        if !absolute_indexing {
-            self.line = self.line + line_changes;
-        } else if self.line.into() > Line(line_changes) {
-            self.line = self.line - line_changes;
-        } else {
-            self.line = Default::default();
-        }
+        self.line = self.line + (rhs + self.col.0) / num_cols;
         self.col = Column((self.col.0 + rhs) % num_cols);
         self
     }
--- a/alacritty_terminal/src/index.rs
+++ b/alacritty_terminal/src/index.rs
@@ -44,33 +44,38 @@ impl<L> Point<L> {
 
     #[inline]
     #[must_use = "this returns the result of the operation, without modifying the original"]
-    pub fn sub(mut self, num_cols: usize, length: usize, absolute_indexing: bool) -> Point<L>
+    pub fn sub(mut self, num_cols: usize, rhs: usize, absolute_indexing: bool) -> Point<L>
     where
-        L: Copy + Add<usize, Output = L> + Sub<usize, Output = L>,
+        L: Copy + Default + Into<Line> + Add<usize, Output = L> + Sub<usize, Output = L>,
     {
-        let line_changes = f32::ceil(length.saturating_sub(self.col.0) as f32 / num_cols as f32);
+        let line_changes =
+            f32::ceil(rhs.saturating_sub(self.col.0) as f32 / num_cols as f32) as usize;
         if absolute_indexing {
-            self.line = self.line + line_changes as usize;
+            self.line = self.line + line_changes;
+        } else if self.line.into() > Line(line_changes) {
+            self.line = self.line - line_changes;
         } else {
-            self.line = self.line - line_changes as usize;
+            self.line = Default::default();
         }
-        self.col = Column((num_cols + self.col.0 - length % num_cols) % num_cols);
+        self.col = Column((num_cols + self.col.0 - rhs % num_cols) % num_cols);
         self
     }
 
     #[inline]
     #[must_use = "this returns the result of the operation, without modifying the original"]
-    pub fn add(mut self, num_cols: usize, length: usize, absolute_indexing: bool) -> Point<L>
+    pub fn add(mut self, num_cols: usize, rhs: usize, absolute_indexing: bool) -> Point<L>
     where
-        L: Copy + Add<usize, Output = L> + Sub<usize, Output = L>,
+        L: Copy + Default + Into<Line> + Add<usize, Output = L> + Sub<usize, Output = L>,
     {
-        let line_changes = (length + self.col.0) / num_cols;
-        if absolute_indexing {
+        let line_changes = (rhs + self.col.0) / num_cols;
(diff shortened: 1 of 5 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -281,6 +281,52 @@ impl<'a, C> RenderableCellsIter<'a, C> {
...
+
+            // Check trailing spacer
+            return selection.contains(next.col, next.line)
+                // Check line-wrapping, leading spacer
+                || (self.grid[&prev].flags.contains(Flags::WIDE_CHAR_SPACER)
+                    && !self.grid[&prevprev].flags.contains(Flags::WIDE_CHAR)
+                    && selection.contains(prev.col, prev.line));
+        }
+
+        // Check if spacer's wide char is selected
```
kchibisov: This comment seems strange, I guess wording is a bit messed up.

alacritty_terminal/src/term/mod.rs
```diff
@@ -281,6 +281,52 @@ impl<'a, C> RenderableCellsIter<'a, C> {
...
+        if cell.flags.contains(Flags::WIDE_CHAR_SPACER) {
+            let prev = point.sub(num_cols, 1, false);
+
+            if self.grid[&prev].flags.contains(Flags::WIDE_CHAR) {
+                // Check previous cell for trailing spacer
+                return self.is_selected(prev);
+            } else {
+                // Check next cell for line-wrapping, leading spacer
+                let next = point.add(num_cols, 1, false);
+                return self.is_selected(next);
```
kchibisov: Could you explain to me why do we need recursion?

alacritty_terminal/src/term/mod.rs
```diff
@@ -281,6 +281,52 @@ impl<'a, C> RenderableCellsIter<'a, C> {
...
+            }
+        }
+
+        false
```
kchibisov: I feel like you can just do the following instead of this strange `return` logic.

This is pseudo code.

```rust

if cell.flags... {
	let prev = point.sub
	let point = if self.grid... {
		&prev
	} else {
		&next
	}
	
	self.is_selected(point)
} else {
	false
}

```

alacritty_terminal/src/term/mod.rs
```diff
@@ -949,9 +987,9 @@ impl<T> Term<T> {
 
         if is_block {
             for line in (end.line + 1..=start.line).rev() {
-                res += &(self.line_to_string(line, start.col..end.col) + "\n");
+                res += &(self.line_to_string(line, start.col..end.col, start.col.0 == 0) + "\n");
```
kchibisov: Could you explain to me the idea behind including wrap here? I feel like we shouldn't do it?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1017,6 +1064,15 @@ impl<T> Term<T> {
             text.push('\n');
         }
 
+        // If wide char is not part of the selection, but leading spacer is, include it
+        if line_end == self.grid.num_cols()
```
kchibisov: It seems like `line_end` is zero based, but `self.grid.num_cols()`?

alacritty_terminal/src/index.rs
```diff
@@ -44,33 +44,38 @@ impl<L> Point<L> {
...
+        let line_changes =
+            f32::ceil(rhs.saturating_sub(self.col.0) as f32 / num_cols as f32) as usize;
         if absolute_indexing {
-            self.line = self.line + line_changes as usize;
+            self.line = self.line + line_changes;
+        } else if self.line.into() > Line(line_changes) {
+            self.line = self.line - line_changes;
         } else {
-            self.line = self.line - line_changes as usize;
+            self.line = Default::default();
```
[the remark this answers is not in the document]
chrisduerr: Yeah, we don't want to underflow. Partially because that panics in debug mode and partially because that doesn't logically make sense. If you move a point down and it goes below the viewport, you don't want it to appear at the top.

alacritty_terminal/src/term/mod.rs
```diff
@@ -281,6 +281,52 @@ impl<'a, C> RenderableCellsIter<'a, C> {
...
+                return self.is_selected(prev);
+            } else {
+                // Check next cell for line-wrapping, leading spacer
+                let next = point.add(num_cols, 1, false);
+                return self.is_selected(next);
+            }
+        }
+
+        false
+    }
```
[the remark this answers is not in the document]
chrisduerr: I think the current approach should be pretty good for perf actually, because in most cases we exit early and never go inside of the WIDE_CHAR and WIDE_CHAR_SPACER blocks.

So we just have two conditionals that are absolutely necessary anyways and everything else is skipped. Except for the `self.grid[&point]` and `num_cols` parts, but maybe the optimizer even helps with that, I don't know. They shouldn't impact performance much though.

But the main goal is making things more readable, I think that's definitely worth it here. The performance impact should not be measurable, I certainly can't.

alacritty_terminal/src/term/mod.rs
chrisduerr: How so?

alacritty_terminal/src/term/mod.rs
chrisduerr: Because I did not want any duplicate code in this already way too complex logic. These codepaths should rarely be hit, so performance isn't really much of an issue.

But basically what we're doing here is that the `WIDE_CHAR` cell itself has full authority on checking if it is selected or not. So the spacers around it just query that cell for the selection information.

Otherwise we'd have to check in the left spacer if either WIDE_CHAR or right spacer is selected, check in the right spacer if WIDE_CHAR or left spacer is selected and then check in the WIDE_CHAR if either left spacer or right spacer are selected.

alacritty_terminal/src/term/mod.rs
chrisduerr: I don't think the `let point = if { ... } else { }` benefits us here, but the rest of this looks good.

alacritty_terminal/src/term/mod.rs
chrisduerr: If you look at where `line_end` is defined, you can see that it is not zero based. We check with `min(line_length, cols.end + 1)`, where `cols.end` is zero-based. So we add one because the final line length specifically isn't.

kchibisov: and now if you look at the check above `line-end == Column(0)`, you'll see some odd logic.

alacritty_terminal/src/term/mod.rs
chrisduerr: If we have a block selection that includes part of a wide char at the right, but which doesn't start at the left side of the terminal (in column 0), we need to expand across the wide char.

chrisduerr: But it should be `start.col.0 != 0`.

kchibisov: not it make more sense.

alacritty_terminal/src/term/mod.rs
```diff
@@ -281,6 +281,49 @@ impl<'a, C> RenderableCellsIter<'a, C> {
...
+            if self.grid[&prev].flags.contains(Flags::WIDE_CHAR) {
+                // Check previous cell for trailing spacer
+                self.is_selected(prev)
+            } else {
+                // Check next cell for line-wrapping, leading spacer
+                self.is_selected(point.add(num_cols, 1))
+            }
+        } else {
+            false
+        }
```
kchibisov: absolute beast.

alacritty_terminal/src/term/mod.rs
chrisduerr: It's not actually that odd. You have to look at it more like a count, the count of included cells.

So even if it's 1-based, because the "first element" is at 1, that doesn't mean that a length of 0 doesn't exist. This will be the case when the entire line is empty for example.

alacritty_terminal/src/term/mod.rs
kchibisov: idk, it was hard to read it tbh. maybe `check if wide spacer's wide char is selected`, but you can ignore this one.

alacritty_terminal/src/term/mod.rs
kchibisov: Could you name it `line_len`, since it's what you're trying to say to me.

chrisduerr: Sounds like a great idea, yeah.

kchibisov:
note: I'll give (hopefully) a final look on this thing today and merge it later this evening in case everything will be fine.

Issue #3191, opened 2020-01-12
Incorrect mouse position


DivineGod:
Using the mouse to select text is off when on Retina screen.

Bug introduced in https://github.com/alacritty/alacritty/commit/c34ec12c309695e4c14d8e50b5f3f54198f70775

### System

OS: macOS
Version: master

Reproduction:

Open Alacritty on retina screen
Try to select text
Selection is off from actual cursor position

DivineGod:
start-up log:

[Alacritty-91939.log](https://github.com/alacritty/alacritty/files/4050322/Alacritty-91939.log)

kchibisov:
opened upstream issue https://github.com/rust-windowing/winit/issues/1389

DivineGod:
Confirmed by adding the following to top-level Cargo.toml

```toml
[patch.crates-io]
winit = { git = 'https://github.com/rust-windowing/winit' }
```

that winit `master` fixes the issue.

ddevault:
Can reproduce on Linux/Wayland as well.

kchibisov:
@ddevault are you using some master version of alacritty? it shouldn't be a problem on v0.4.1.

ddevault:
Aye I'm using master, I switched to the stable release for now.

tinmarino:
PR #3307 solved it here (Ubuntu 20.10 / Tmux 3.1b). The Shift left click was one character left and one above the mouse position.
Now it is perfect!
Thanks devs!

PR #3306, opened 2020-02-07
Update glutin dependency to 0.23.0

doublej472:
An update was pushed to the dependency [`glutin`](https://github.com/rust-windowing/glutin/blob/master/CHANGELOG.md#version-0230-2020-02-06), which fixes https://github.com/alacritty/alacritty/issues/3191.

Tested with Sway on Linux, and it fixes that issue for me.


kchibisov:
Thanks for your PR and sorry for closing it.

Your PR is missing a lot of entries to `CHANGELOG`, also with this update we'd like to update our code a bit, due to addition of curtain things to `winit`. So instead of making you go through our bug  tracker and trying to guess what you should do, I'll send a PR shortly myself doing this, but thanks for telling us that the new `glutin` version was released.

I hope you understand.

PR #3310, opened 2020-02-07
Explicitly state that we're not creating config file

kchibisov:
I think we should explicitly state that, since amount of questions on IRC and our issue tracker is increasing. I feel like it wasn't really obvious tbh.

cc @cole-h to check wording etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,8 +158,8 @@ executable and is available through the
 You can find the default configuration file with documentation for all available
 fields on the [GitHub releases page](https://github.com/alacritty/alacritty/releases) for each release.
 
-Alacritty doesn't create config file for you, therefore you should put it in one
-of the following locations yourself:
+Alacritty doesn't create the config file for you, but it looks for one in the
+following locations:
 
 1. `$XDG_CONFIG_HOME/alacritty/alacritty.yml`
 2. `$XDG_CONFIG_HOME/alacritty.yml`
@@ -168,7 +168,7 @@ of the following locations yourself:
 
 ### Windows
 
-On Windows the config file should be at:
+On Windows, the config file should be located at:
 
 `%APPDATA%\alacritty\alacritty.yml`
 
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ of the following locations yourself:
 
 ### Windows
 
-On Windows the config file should be put in:
+On Windows the config file should be at:
 
 `%APPDATA%\alacritty\alacritty.yml`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -167,7 +168,7 @@ Alacritty looks for the configuration file at the following paths:
 
 ### Windows
 
-On Windows the config file is located at:
+On Windows the config file should be put in:
```
chrisduerr:
```suggestion
On Windows the config file should be at:
```

We're directly linking a file location, not a directory. So we shouldn't use 'in'.

README.md
```diff
@@ -158,7 +158,8 @@ executable and is available through the
 You can find the default configuration file with documentation for all available
 fields on the [GitHub releases page](https://github.com/alacritty/alacritty/releases) for each release.
 
-Alacritty looks for the configuration file at the following paths:
+Alacritty doesn't create config file for you, therefore you should put it in one
+of the following locations yourself:
```
chrisduerr: I don't think the old meaning implicated that the configuration file is created by Alacritty, but it did make it more clear that there's a fallback priority here.

Our manpage also still states this exact thing.

It might be easier to just change the previous paragraph from `You can find the default configuration file...` to `You can copy the default configuration file...`.

kchibisov: I feel like this part `Alacritty doesn't create config file for you` is important tbh. Like you can't just rely on `copy` tbh. Like I could understand the idea behind copy, but you shouldn't judge people by yourself here.


`Alacritty doesn't create config file for you, if you want to tune it for your needs create one in one of the following locations:`

?

chrisduerr: It should be "doesn't create **a** config file for you" or *the* works too I suppose.

 The `one in one` part also doesn't really flow very well and I'm not sure if `tune` is the right word, make `tweak` is a bit better? And I think it should be `to your needs`, rather than `for your needs`.

Maybe we should just keep it a bit simpler like this:

```
Alacritty doesn't create the config file for you, but it looks for it in the following locations:
```

Should still be very clear, makes it obvious that there is a fallback behavior going on and it doesn't include any text besides what is necessary. The biggest problem is that the `but` might not be optimal, but it should be fine?

cole-h: I'm a fan of @chrisduerr's wording above. Maybe change `it looks for it in` -> `it looks for one in`, though.

The `but` seems fine to me. Could also go for the super intellectual `...for you; however, it looks for one...`. Instant bonus points for using a semicolon.

README.md
```diff
@@ -167,7 +168,7 @@ Alacritty looks for the configuration file at the following paths:
 
 ### Windows
 
-On Windows the config file is located at:
+On Windows the config file should be at:
```
cole-h:
```suggestion
On Windows, the config file should be located at:
```

PR #3322, opened 2020-02-11
Add selection expansion

### LINKED ISSUE #1554: Select with Shift + Click

chrisduerr:
This allows for expanding the selection using the shift key. The new
selection type depends on the number of clicks and applies to both sides
of the selection.

Fixes #1554.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for Fontconfig embolden and matrix options
 - Opt-out compilation flag `winpty` to disable WinPTY support
 - Scrolling during selection when mouse is at top/bottom of window
-- Expanding existing selections using the Shift key
+- Expanding existing selections using the right mouse button
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for Fontconfig embolden and matrix options
 - Opt-out compilation flag `winpty` to disable WinPTY support
 - Scrolling during selection when mouse is at top/bottom of window
+- Expanding existing selections using the Shift key
 
 ### Changed
 
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -512,36 +512,22 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
         let mut point = self.ctx.size_info().pixels_to_coords(mouse.x, mouse.y);
         point.line = min(point.line, self.ctx.terminal().grid().num_lines() - 1);
 
-        let side = self.ctx.mouse().cell_side;
-
         self.ctx.mouse_mut().click_state = match self.ctx.mouse().click_state {
             ClickState::Click
                 if elapsed < self.ctx.config().ui_config.mouse.double_click.threshold =>
             {
-                self.ctx.mouse_mut().block_url_launcher = true;
-                self.ctx.start_selection(SelectionType::Semantic, point, side);
+                self.on_left_double_click(point);
                 ClickState::DoubleClick
             }
             ClickState::DoubleClick
                 if elapsed < self.ctx.config().ui_config.mouse.triple_click.threshold =>
             {
-                self.ctx.mouse_mut().block_url_launcher = true;
-                self.ctx.start_selection(SelectionType::Lines, point, side);
+
+                self.on_left_triple_click(point);
                 ClickState::TripleClick
             }
             _ => {
-                // Don't launch URLs if this click cleared the selection.
-                self.ctx.mouse_mut().block_url_launcher = !self.ctx.selection_is_empty();
-
-                self.ctx.clear_selection();
-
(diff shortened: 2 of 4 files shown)


alacritty/src/input.rs
```diff
@@ -492,26 +492,77 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
             };
 
             self.mouse_report(code, ElementState::Pressed);
-        } else if button == MouseButton::Left {
-            self.on_left_click();
         } else {
-            // Do nothing when using buttons other than LMB.
-            self.ctx.mouse_mut().click_state = ClickState::None;
+            // Reset click state if button has changed
```
kchibisov, line 496: The comment is in accurate and missing a dot in the end. If anything it should mention that it updates the last button as well.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for Fontconfig embolden and matrix options
 - Opt-out compilation flag `winpty` to disable WinPTY support
 - Scrolling during selection when mouse is at top/bottom of window
+- Expanding existing selections using the Shift key
```
kchibisov, line 28: It's not `Shift` anymore.

alacritty/src/input.rs
```diff
@@ -492,26 +492,77 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
+                self.ctx.mouse_mut().last_click_button = button;
+                self.ctx.mouse_mut().click_state = ClickState::None;
+            }
+
+            // Calculate time since the last click to handle double/triple clicks.
+            let now = Instant::now();
+            let elapsed = now - self.ctx.mouse().last_click_timestamp;
+            self.ctx.mouse_mut().last_click_timestamp = now;
+
+            // Load mouse point, treating message bar and padding as closest cell.
```
kchibisov, line 507:
```suggestion
            // Load mouse point, treating message bar and padding as the closest cell.
```

alacritty/src/input.rs
```diff
@@ -492,26 +492,77 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
 
-    /// Handle left click selection and vi mode cursor movement.
-    fn on_left_click(&mut self) {
-        // Calculate time since the last click to handle double/triple clicks in normal mode.
-        let now = Instant::now();
-        let elapsed = now - self.ctx.mouse().last_click_timestamp;
-        self.ctx.mouse_mut().last_click_timestamp = now;
+    /// Handle selection expansion on right click.
+    fn on_right_click(&mut self, point: Point, elapsed: Duration) {
+        self.ctx.mouse_mut().click_state = match self.ctx.mouse().click_state {
```
kchibisov, line 523: I feel like we can compute click state earlier and use proper click state later? In general it could perform click state computation before `on_left_click`/`on_right_click` and then just match current click state in those handlers.

alacritty/src/input.rs
```diff
@@ -492,26 +492,77 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
+                ClickState::DoubleClick
+            }
+            ClickState::DoubleClick
+                if elapsed < self.ctx.config().ui_config.mouse.triple_click.threshold =>
+            {
+                self.expand_selection(point, SelectionType::Lines);
+                ClickState::TripleClick
+            }
+            _ => {
+                // Determine selection type.
```
kchibisov, line 537: Not sure that this comment is actually useful.

kchibisov:
Also, should we mention somewhere that rmb expands selection or users should figure that out?

alacritty/src/input.rs
chrisduerr: I don't think the comment is inaccurate, it resets the click state so that the current click is only a single click. Updating the last button is just to be able to reset the click state again in the future, it shouldn't have any functional effect on anything else since we never read it except for this block.

The punctuation is incorrect though!

alacritty/src/input.rs
chrisduerr: That's not a bad idea actually. I'll look into it.

alacritty/src/input.rs
chrisduerr: Not really.

alacritty/src/input.rs
chrisduerr: Of course this requires matching twice, but I think it should be worth it for the sake of cleaner code.

chrisduerr:
> Also, should we mention somewhere in alacritty.yml that rmb expands selection or users should figure that out?

We don't really have things like that documented in the config, which makes sense since they cannot be configured. It might make sense to document the mouse behavior in another place, though I couldn't tell you where or how. But that applies to all implicit mouse bindings even just selection itself.

I think for now it should be fine to let people figure it out themselves, especially because I don't think there's any place that people are actually going to read (who reads documentation anyways?!).

Though I think a "user friendly" version of our escape sequence support list might be useful. Just a list with all the highlevel features that Alacritty supports. That can also help discover things like vi mode maybe. What do you think @kchibisov? I'd open a separate issue for this.

PR #3315, opened 2020-02-09
Add modal keyboard motion mode

chrisduerr:
This implements a basic mode for navigating inside of Alacritty's
history with keyboard bindings. They're bound by default to vi's motion
shortcuts but are fully customizable. Since this relies on key bindings
only single key bindings are currently supported (so no `ge`, or
repetition).

Other than navigating the history and moving the viewport, this mode
should enable making use of all available selection modes to copy
content to the clipboard and launch URLs below the cursor.

This also changes the rendering of the block cursor at the side of
selections, since previously it could be inverted to be completely
invisible. Since that would have caused some troubles with this keyboard
selection mode, the block cursor now is no longer inverted when it is at
the edges of a selection.

Fixes #262.

PR #3329, opened 2020-02-12
Add a new cursor style: Glass.

pinkfluid:
Glass looks like just a Hollow cursor (unfocused window), except the
interior is semi-transparent.

Subjectively this is a fancier looking cursor compared to other variants. 

Disclaimer: I have 0 clue about Rust - - the change seemed easy enough and the compiler didn't really complain :)

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -305,6 +305,7 @@
   #   - ▇ Block
   #   - _ Underline
   #   - | Beam
+  #   - ◻ Glass, inner part is semi transparent
   #style: Block
 
   # If this is `true`, the cursor will be rendered as a hollow box when the
--- a/alacritty/src/cursor.rs
+++ b/alacritty/src/cursor.rs
@@ -41,10 +41,11 @@ pub fn get_cursor_glyph(
     }
 
     match cursor {
-        CursorStyle::HollowBlock => get_box_cursor_glyph(height, width, line_width),
+        CursorStyle::HollowBlock => get_box_cursor_glyph(height, width, line_width, 0),
         CursorStyle::Underline => get_underline_cursor_glyph(width, line_width),
         CursorStyle::Beam => get_beam_cursor_glyph(height, line_width),
         CursorStyle::Block => get_block_cursor_glyph(height, width),
+        CursorStyle::Glass => get_box_cursor_glyph(height, width, line_width, 150),
         CursorStyle::Hidden => RasterizedGlyph::default(),
     }
 }
@@ -82,7 +83,7 @@ pub fn get_beam_cursor_glyph(height: i32, line_width: i32) -> RasterizedGlyph {
 }
 
 // Returns a custom box cursor character
-pub fn get_box_cursor_glyph(height: i32, width: i32, line_width: i32) -> RasterizedGlyph {
+pub fn get_box_cursor_glyph(height: i32, width: i32, line_width: i32, fill: u8) -> RasterizedGlyph {
     // Create a new box outline rectangle
     let mut buf = Vec::with_capacity((width * height * 3) as usize);
     for y in 0..height {
@@ -94,7 +95,7 @@ pub fn get_box_cursor_glyph(height: i32, width: i32, line_width: i32) -> Rasteri
             {
                 buf.append(&mut vec![255u8; 3]);
             } else {
-                buf.append(&mut vec![0u8; 3]);
+                buf.append(&mut vec![fill; 3]);
             }
         }
     }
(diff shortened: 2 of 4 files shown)


kchibisov:
Thanks for the PR.

Unfortunately this in not a right solution for such things. I also don't really like this cursor style, but whatever.

Terminals have `Block`,`Beam`, and `Underline` cursors for the reason, that you can manipulate change them with escape sequences. So adding new style isn't an option in general, since there's no way to toggle your custom cursor style and it could be reset with escape sequence to already mentioned ones.

I feel like the way to add such things into alacritty, could be "Styles" for existing 3 cursor styles. So let's say that user can change how its Block or Beam cursor looks like, and just have custom drawing patterns. My motivations to proposing such thing is that I don't really like our `Beam` cursor, since it's too thin for me. However I can't just propose such things in a way to just change default cursor style to be thicker.

I wonder whether we can add some additional GUI cursor styling without messing up original `Block`, `Beam`, `Underline` things. Such thing will purely live in `alacritty/alacritty` and be applied on a stage of drawing.

Hollow cursor is an exception and it's being drawn in a bit different way, but only to indicate focus.

Anyway, I'd like to discuss an option to add similar things to what you propose here with @chrisduerr and @nixpulvis first, after making any decision myself.

pinkfluid:
Hi, thank you for your quick response.

What you describe makes sense, I didn't really know you can switch cursor styles with escape sequences. I think what you're proposing is probably the right way to approach the problem, unfortunately I cannot be much of help here since I don't really know Rust :)

Anyway, feel free to close the PR.

chrisduerr:
Just out of curiosity, where did you get the idea for this PR? Was this inspired by a different terminal emulator?

I'd say that Alacritty already provides a great choice of cursors with four different variants and I don't really see the advantages of adding another cursor.

What makes this cursor different from say a cursor that renders just the top and bottom of a box, not the lines at the side? So why would we merge this, but not the other one? I'd like to know more about the motivation behind this to evaluate if there's a good reason for adding this other than personal preference of one user, since unfortunately it's not reasonable to add a custom cursor style for each user's preferences.

Currently this looks a bit like feature creep, something that is trivial to implement but doesn't provide a lot of benefits. But I much appreciate the PR, so thanks for the suggestion. And definitely let me know about your motivations please.

nixpulvis:
A way to additionally style the cursor could be very useful in general. For example, this comment https://github.com/alacritty/alacritty/pull/3315#discussion_r377914113 talks about a desire for some way to style a cursor that's being controlled in a new "keyboard motion" mode to make it clear that you're in the mode. Ideally this additional style would apply to any user's choice of cursor style (i.e. Block/Beam/Underline).

pinkfluid:
@chrisduerr 
> Just out of curiosity, where did you get the idea for this PR? Was this inspired by a different terminal emulator?

No, it wasn't really inspired by another terminal emulator. For the longest time I've been using st. And looking at a terminal window for most of my daily life, I get itches to fixes small details like this (probably its just me, but block cursors are boring after a while). For this reason I've implemented a patch for st that gives the cursor a white outline, which is somewhat reminiscent of my mouse coursor:

st on x11

Alacritty for me was the first terminal emulator that was able to displace st in terms of simplicity and features. It's really great. But then some ithces came back, and this was one of these. Unfortunately I couldn't get exactly the same effect as st redraws the crusor every time while to alacritty the cursor seems to be some sort of mask:

alacritty on wayland

I kinda liked this cursor over the default block and the patch looked really small so I figured I would shared it upstream. I didn't know you can actually switch cursor shapes with escape code until yesterday (never really used that feature) and the various cursors in alacritty were meant for that purpose. And yes, now that I have that knowledge this this patch doesn't really feel like the correct solution.

> I feel like the way to add such things into alacritty, could be "Styles" for existing 3 cursor styles. So let's say that user can change how its Block or Beam cursor looks like, and just have custom drawing patterns. 

Yeah something like that would be welcome. I would assume the various cursor styles would be defined with a mask?

chrisduerr:
> block cursors are boring after a while

It seems like it's mostly preference then I suppose, with a sprinkle of grass always being greener on the other side.

I think the current block cursor should functionally work just as well, so I don't see a reason to merge this as-is. Of course there's nothing wrong with maintaining this as a fork, which is what I would do if I would actually desire this feature myself.

Also it shouldn't be a problem to patch Alacritty to behave the same way as your ST patch, it would probably just be a slightly bigger change (though compared to other changes still tiny and quite simple). You'd probably want to draw a box cursor on top of a cell and change the background/foreground of the cell.

a-elhag:
I'm coming over from windows, and their terminals do have an option for a semi-transparent cursor. Seeing the character below the cursor is an improvement in my opinion. But I don't mind patching it on my end, is there anything I need to be aware of before doing so @pinkfluid?

PR #3330, opened 2020-02-12
Fix display_offset computation when on top of scrollback

kchibisov:
Regression was introduced in 4cc6421daa4ff5976ab43c67110a7a80a36541e5,
however it was working before only due to grid.len() bug.


kchibisov:
Forgot to mention, to repro the thing I'm fixing here, you should do the following;
```bash
while [ 1 ]; do
	echo 1
	sleep 0.1
done
```

and try to stay on your top line, you'll notice that you're actually scrolling a bit.

PR #3336, opened 2020-02-13
Cache FreeType faces instead of loading new

kchibisov:
This commit introduces a cache for all loading faces, and replaces `Face` with `FaceLoadingProperties`, since we don't need to carry faces around, only loading options for them. It also removes `FontID` in favor of building `FontKey` right away.

I'd also like to test an idea of returning `FaceLoadingProperties` over `FontKey`, however it'll likely require to use `Rc` boxes or cloning them. So I'd like to test performance of the solution mentioned in this paragraph.

In addition, I think it'll be nice to add an API for rasterizer to remove curtain keys from use, since now they're kind of persistent, but I'm not sure that it's up to this PR.

### DIFF
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -16,10 +16,12 @@
 use std::cmp::{min, Ordering};
 use std::collections::HashMap;
 use std::fmt::{self, Display, Formatter};
+use std::path::PathBuf;
+use std::rc::{Rc, Weak};
 
-use freetype::freetype_sys;
 use freetype::tt_os2::TrueTypeOS2Table;
 use freetype::{self, Library};
+use freetype::{freetype_sys, Face as FTFace};
 use libc::c_uint;
 use log::{debug, trace};
 
@@ -32,28 +34,28 @@ use super::{
     Style, Weight,
 };
 
+#[derive(Debug, PartialEq, Copy, Clone)]
 struct FixedSize {
     pixelsize: f64,
 }
 
 struct FallbackFont {
     pattern: Pattern,
-    id: FontID,
+    key: FontKey,
 }
 
 impl FallbackFont {
-    fn new(pattern: Pattern, id: FontID) -> FallbackFont {
-        Self { pattern, id }
+    fn new(pattern: Pattern, key: FontKey) -> FallbackFont {
+        Self { pattern, key }
     }
 }
 
-#[derive(Debug, PartialEq, Eq, Copy, Clone, Hash)]
-struct FontID(u32);
-
-impl FontID {
-    fn new(lhs: PatternHash, rhs: PatternHash) -> Self {
+impl FontKey {
+    fn from_pattern_hashes(lhs: PatternHash, rhs: PatternHash) -> Self {
         // XOR two hashes to get a font ID
-        Self(lhs.0.rotate_left(1) ^ rhs.0)
+        Self {
+            token: lhs.0.rotate_left(1) ^ rhs.0,
+        }
     }
 }
 
@@ -63,22 +65,20 @@ struct FallbackList {
     coverage: CharSet,
 }
 
-struct Face {
-    ft_face: freetype::Face,
-    key: FontKey,
+struct FaceLoadingProperties {
     load_flags: freetype::face::LoadFlag,
     render_mode: freetype::RenderMode,
     lcd_filter: c_uint,
     non_scalable: Option<FixedSize>,
     has_color: bool,
     pixelsize_fixup_factor: Option<f64>,
+    ft_face: Rc<FTFace>,
 }
 
-impl fmt::Debug for Face {
+impl fmt::Debug for FaceLoadingProperties {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         f.debug_struct("Face")
             .field("ft_face", &self.ft_face)
-            .field("key", &self.key)
             .field("load_flags", &self.load_flags)
             .field("render_mode", &match self.render_mode {
                 freetype::RenderMode::Normal => "Normal",
@@ -95,9 +95,9 @@ impl fmt::Debug for Face {
 
 /// Rasterizes glyphs for a single font face.
 pub struct FreeTypeRasterizer {
-    faces: HashMap<FontKey, Face>,
     library: Library,
-    keys: HashMap<FontID, FontKey>,
+    faces: HashMap<FontKey, FaceLoadingProperties>,
+    ft_faces: HashMap<PathBuf, Weak<FTFace>>,
     fallback_lists: HashMap<FontKey, FallbackList>,
     device_pixel_ratio: f32,
 }
@@ -115,7 +115,7 @@ impl Rasterize for FreeTypeRasterizer {
 
         Ok(FreeTypeRasterizer {
             faces: HashMap::new(),
-            keys: HashMap::new(),
+            ft_faces: HashMap::new(),
             fallback_lists: HashMap::new(),
             library,
             device_pixel_ratio,
@@ -123,8 +123,8 @@ impl Rasterize for FreeTypeRasterizer {
     }
 
     fn metrics(&self, key: FontKey, _size: Size) -> Result<Metrics, Error> {
-        let face = self.faces.get(&key).ok_or(Error::FontNotLoaded)?;
-        let full = self.full_metrics(key)?;
+        let face = &mut self.faces.get(&key).ok_or(Error::FontNotLoaded)?;
+        let full = self.full_metrics(&face)?;
 
         let height = (full.size_metrics.height / 64) as f64;
         let descent = (full.size_metrics.descender / 64) as f32;
@@ -142,7 +142,7 @@ impl Rasterize for FreeTypeRasterizer {
 
         // Get strikeout position and thickness in device pixels
         let (strikeout_position, strikeout_thickness) =
-            match TrueTypeOS2Table::from_face(&mut face.ft_face.clone()) {
+            match TrueTypeOS2Table::from_face(&mut (*face.ft_face).clone()) {
                 Some(os2) => {
                     let strikeout_position = f32::from(os2.y_strikeout_position()) * x_scale / 64.;
                     let strikeout_thickness = f32::from(os2.y_strikeout_size()) * x_scale / 64.;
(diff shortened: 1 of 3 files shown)


font/src/ft/mod.rs
```diff
@@ -63,22 +65,20 @@ struct FallbackList {
...
-    ft_face: freetype::Face,
-    key: FontKey,
+struct FaceLoadingProperties {
     load_flags: freetype::face::LoadFlag,
     render_mode: freetype::RenderMode,
     lcd_filter: c_uint,
     non_scalable: Option<FixedSize>,
     has_color: bool,
     pixelsize_fixup_factor: Option<f64>,
+    ft_face: Rc<FTFace>,
```
chrisduerr: Okay, before going deeper into this code I have the question what the benefit is of having a reference to the face in the hashmap here, rather than just storing the PathBuf to the font?

Is the goal solely to make it so the reference is dropped, as soon as `FaceLoadingProperties` is dropped, destroying the face?

Are there any concerns about just trashing the `ft_faces` hashmap with a bazillion paths to faces that don't even exist anymore?

kchibisov: >Is the goal solely to make it so the reference is dropped, as soon as FaceLoadingProperties is dropped, destroying the face

Yes, that was the idea.

>Is the goal solely to make it so the reference is dropped, as soon as FaceLoadingProperties is dropped, destroying the face

IIRC paths are cheaper to keep around compared to faces, since it’s max 4KB per path, but it is a worst case.

chrisduerr: Have you ever checked how much is needed to stare a face in Rust? 4KB seems like a lot.

kchibisov: Oh, in addition having reference to a `FTFace` simplify the logic, since you need `FTFace` to perform rasterizing, so using indirection with `PathBuf` is slower, since you need to query `HashMap` each time to get a proper face. I have an idea on how to solve your zombie `PathBuf` issue though.

kchibisov: Ok I was a bit wrong that I know a way to remove this zombies automatically. However you repro an issue I'm fixing by doing `cat /dev/urandom` and starting changing your font size, you should notice that master uses way more RAM on average.

font/src/ft/mod.rs
```diff
@@ -32,28 +34,28 @@ use super::{
     Style, Weight,
 };
 
+#[derive(Debug, PartialEq, Copy, Clone)]
 struct FixedSize {
     pixelsize: f64,
 }
```
chrisduerr: Unrelated to this PR directly, but since there's a change to this: Why is this a dedicated struct? Seems like we could just as well use a `f64` directly?

font/src/ft/mod.rs
```diff
@@ -253,31 +253,23 @@ impl FreeTypeRasterizer {
...
+                    self.faces.remove(&fallback_font.key);
                 }
             }
-
-            let _ = self.faces.remove(&font_key);
-            Some(font_key)
-        } else {
-            None
-        };
+        }
```
chrisduerr: The `primary_font_key` is a hash that uniquely identifies a font, since it's just a combinations of all patterns required to load it.

So it seems like we should just be able to return the `FontKey` immediately when we find that it is already loaded? Or what do we reload the fonts for?

font/src/ft/mod.rs
```diff
@@ -302,32 +294,55 @@ impl FreeTypeRasterizer {
...
-
-        let width = match face.ft_face.load_char('0' as usize, face.load_flags) {
-            Ok(_) => face.ft_face.glyph().metrics().horiAdvance / 64,
-            Err(_) => size_metrics.max_advance / 64,
-        } as f64;
+        let width =
+            match face_load_props.ft_face.load_char('0' as usize, face_load_props.load_flags) {
+                Ok(_) => face_load_props.ft_face.glyph().metrics().horiAdvance / 64,
+                Err(_) => size_metrics.max_advance / 64,
+            } as f64;
```
chrisduerr: Might be worth to extract `face_load_props.ft_face` into a variable here, since it would prevent two very ugly linebreaks.

font/src/ft/mod.rs
```diff
@@ -302,32 +294,55 @@ impl FreeTypeRasterizer {
...
+
+            let ft_face = if let Some(face) = self.ft_faces.get(&path) {
+                if let Some(ft_face) = face.upgrade() {
+                    ft_face
+                } else {
+                    self.load_ft_face(path, index)?
+                }
+            } else {
+                self.load_ft_face(path, index)?
+            };
```
chrisduerr: Maybe we can do something like this here:

```rust
let face = match self.ft_faces.get(&path).and_then(|face| face.upgrade()).unwrap_or_else(|| self.load_ft_face(path, index)?;
```

I'm not sure about the exact syntax, but it should work. Alternatively if it looks nicer, we could also `match` inside of `unwrap_or_else`, but I think it should format nicely.

kchibisov: It doesn't look nice, since you can't use `?` in a closure, you also can't use `ok_or_else`, since it applies random thing and you don't have a proper function to do something if none, but not an `unwrap`, like `map_none` could help, but we don't have one, so you should add it youself, which isn't helpful here and you can't do it un Rust anyways IIRC.

chrisduerr: Could still apply the `match` suggestion just fine.

kchibisov: Oh, it was really hard to parse your one liner tbh.

font/src/ft/mod.rs
```diff
@@ -306,32 +285,49 @@ impl FreeTypeRasterizer {
...
-            Ok(_) => face.ft_face.glyph().metrics().horiAdvance / 64,
+        let width = match ft_face.load_char('0' as usize, face_load_props.load_flags) {
+            Ok(_) => ft_face.glyph().metrics().horiAdvance / 64,
             Err(_) => size_metrics.max_advance / 64,
         } as f64;
 
         Ok(FullMetrics { size_metrics, cell_width: width })
     }
 
+    fn load_ft_face(&mut self, path: PathBuf, index: isize) -> Result<Rc<FTFace>, Error> {
```
kchibisov: We can inline this function now, what do you think about it?

chrisduerr: I don't see anything wrong with keeping it separate.

font/src/ft/mod.rs
```diff
@@ -253,31 +246,16 @@ impl FreeTypeRasterizer {
...
 
-            let _ = self.faces.remove(&font_key);
-            Some(font_key)
-        } else {
-            None
-        };
+        // Return font if we already loaded it
+        if self.faces.get(&primary_font_key).is_some() {
+            return Ok(primary_font_key);
+        }
```
chrisduerr: Thinking about this early return, there might be an issue with fallback faces, right?

In `self.faces` both primary and fallback faces are stored. However, a primary face requires the additional computation we do in this method to load the fallback list.

So it seems like what we should do is check if the font is already loaded **and** if a fallback list exists for this font.

However even if the fallback list does not exist, we should be able to still skip the call to `face_from_pattern`. Just not the entire rest of this function.

kchibisov: >Thinking about this early return, there might be an issue with fallback faces, right?

Yeah, exactly, forgot about this case.

font/src/ft/mod.rs
```diff
@@ -253,82 +246,91 @@ impl FreeTypeRasterizer {
...
-                        self.keys.remove(&fallback_font.id);
-                    }
-                }
-            }
+        let primary_font_key = FontKey::from_pattern_hashes(hash, primary_font.hash());
 
-            let _ = self.faces.remove(&font_key);
-            Some(font_key)
+        // Return font if we already loaded it
+        if self.faces.get(&primary_font_key).is_some() {
```
```suggestion
        if self.fallback_lists.get(&primary_font_key).is_some() {
```

font/src/ft/mod.rs
```diff
@@ -253,82 +246,91 @@ impl FreeTypeRasterizer {
...
+            self.face_from_pattern(&primary_font, primary_font_key)
+                .and_then(|pattern| pattern.ok_or_else(|| Error::MissingFont(desc.to_owned())))?
         };
 
-        // Reuse the font_key, since changing it can break library users
-        let font_key = self
-            .face_from_pattern(&primary_font, primary_font_id, font_key)
-            .and_then(|pattern| pattern.ok_or_else(|| Error::MissingFont(desc.to_owned())))?;
+        // Build fallback list if we don't have one for a given font key
+        if self.fallback_lists.get(&primary_font_key).is_none() {
```
chrisduerr: Don't need to check that here with my earlier suggestion.

font/src/ft/mod.rs
```diff
@@ -248,21 +248,16 @@ impl FreeTypeRasterizer {
         // Hash pattern together with request pattern to include requested font size in the hash
         let primary_font_key = FontKey::from_pattern_hashes(hash, primary_font.hash());
 
-        // Return font if we already loaded it
-        if self.faces.get(&primary_font_key).is_some() {
-            return Ok(primary_font_key);
-        }
+        // Load font if we haven't loaded one
+        if self.fallback_lists.get(&primary_font_key).is_none() {
```
chrisduerr: Early return here, please `contains_key` instead of actually accessing the value and then calling `is_some()` or `is_none()`.

kchibisov: I did this thing on purpose, because clippy has a bug when using `contains_key` here and then `insert`, like it asks you to use `entry`, but it doesn't make any sense, since you should compute the thing before insert.

font/src/ft/mod.rs
```diff
@@ -253,123 +246,107 @@ impl FreeTypeRasterizer {
...
+        // Return if we already have the same primary font
+        if self.fallback_lists.contains_key(&primary_font_key) {
+            return Ok(primary_font_key);
+        }
 
-        // Reuse the font_key, since changing it can break library users
-        let font_key = self
-            .face_from_pattern(&primary_font, primary_font_id, font_key)
-            .and_then(|pattern| pattern.ok_or_else(|| Error::MissingFont(desc.to_owned())))?;
+        // Load font if we haven't loaded one
```
```suggestion
        // Load font if we haven't loaded it yet
```

Since `one` is not ideal here, we only ever want to load one when talking about primary fonts.

PR #3337, opened 2020-02-13
Add seconds and nanoseconds to log output

chrisduerr:


nixpulvis:
:+1: for now, though I do think we're due for a refactor of our log output format. Just look how much space is used before the message. I was just at a Rust meetup in Boston actually, and one of the talks was about the [`tracing`](https://docs.rs/tracing/0.1.12/tracing/) library, which seems like an awesome potential improvement for us over the basic `log` crate.

Until then however, if you need second and subseconds then this seems fine. Just needs a rebase then merge #EEET.

PR #3343, opened 2020-02-15
Fix parser stopping at unknown modes

### LINKED ISSUE #3339: ANSI SGRs after an unknown SGR are incorrectly ignored
An escape sequence like this:
ESC[31;10;44mHelloESC[0m
... will render incorrectly with a red foreground and black background, instead of a red foreground and blue background.
It's because of this block in ansi.rs, in vte::Perform:

            ('m', None) => {
                if args.is_empty() {
                    handler.terminal_attribute(Attr::Reset);
                } else {
                    for attr in attrs_from_sgr_parameters(args) {
                        match attr {
                            Some(attr) => handler.terminal_attribute(attr),
                            None => {
                                unhandled!();
/* The "return" below causes all attributes after an unknown one to be ignored! */
                                return;
                            },
                        }
                    }
                }
            },
Because the return causes control to bail out without processing any further attributes.
Removing the return corrects the issue, but I don't know what else it may break.

chrisduerr:
This resolves an issue in the parser where it would stop as soon as the
first unknown value is encountered in private mode/sgr attribute
escapes.

Fixes #3339.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Modifier key combinations like `Ctrl + Q` not generating characters on macOS
 - Handling of URLs with single quotes
 - Parser reset between DCS escapes
+- Parser stopping at unknown DEC private modes/SGR character attributes
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


kchibisov:
I feel like their algo to compute complexity is kind of broken tbh... Anyway, I wasn't able to simplify the logic.

chrisduerr:
> I feel like their algo to compute complexity is kind of broken tbh... Anyway, I wasn't able to simplify the logic.

Yeah, I've been wondering about the best strategy to move forward with this and I'm pretty sure I'll at least temporarily disable this lint globally. I definitely want to take this discussion upstream though.

PR #3362, opened 2020-02-20
Don't round pixelsize before quering

kchibisov:


kchibisov:
If you look on this line https://github.com/alacritty/alacritty/blob/8abca441827788d48de49d3552735cdd0999379e/font/src/ft/mod.rs#L222  , you can notice that we're kind of limiting pixelsize to have fraction only 0.5, so we're using it for metrics, etc later on. However it wasn't like that during loading a glyph, so glyph metrics could go wrong and mismatch later on.

kchibisov:
I've pushed inverted solution to an original problem to use metrics identical for querying.

kchibisov:
It was discussed on IRC and there wasn't a clear point on what we should do. This PR just sinks the size we query with the size we use. This change could only affect bitmap fonts and emoji fonts(they also bitmap fonts most of the time), however it'll just sync their size/fixup factors with the one we're currently using for rendering.

I'm in favor of using fractional pixel sizes like we do right now in `get_glyph`(don't confuse it with point sizes in config). Also most apps allow arbitrary fractional sizes, so I don't see a point about speculating with one more rounding after computation of already rounded values. The idea behind current pixel size code is a mystery to me.

kchibisov:
It affected v0.4.1 but only for bitmap fonts and in a very edge case, so I’m not sure about even mentioning it.

chrisduerr:
> It affected v0.4.1 but only for bitmap fonts and in a very edge case, so I’m not sure about even mentioning it.

If it affected the last version, we should mention it. If it doesn't fix anything, there's no point in changing it, right? This is a change that should affect the user, not just a style fix as far as I can tell.

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Parser reset between DCS escapes
 - Parser stopping at unknown DEC private modes/SGR character attributes
 - Block selection appending duplicate newlines when last column is selected
+- Font size quering not matching renderer font size
```
cole-h:
```suggestion
- Font size querying not matching renderer font size
```

:)

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Parser reset between DCS escapes
 - Parser stopping at unknown DEC private modes/SGR character attributes
 - Block selection appending duplicate newlines when last column is selected
+- Font size querying not matching renderer font size
```
chrisduerr: This documents what caused the bug, not what bug we're trying to fix. We should state the effect this has on the user and what they can expect to work properly now.

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Parser reset between DCS escapes
 - Parser stopping at unknown DEC private modes/SGR character attributes
 - Block selection appending duplicate newlines when last column is selected
+- Bitmap fonts being a bit smaller than they should in some cases
```
```suggestion
- Bitmap fonts being a bit smaller than they should be in some cases
```

PR #3366, opened 2020-02-21
Add two missing cursor styles in the documentation

r00ster91:


kchibisov:
Oh, it’s a bug that you can set them, we shouldn’t allow such things, since you can’t make them persistent and it’s used only for internal things.

r00ster91:
Oh, they should probably not be displayed in the bottom then when you specify an incorrect cursor style: 
image

kchibisov:
Yeah, I’ve opened #3367 to track this.

PR #3369, opened 2020-02-22
Fix rotation of ongoing selection

### LINKED ISSUE #3232: Selection doesn't rotate if its ongoing
Start selecting something in e.g. weechat and wait for a message, selection wont rotate, so the entire selection becomes pointless.

note: weechat is using scrolling regions.

chrisduerr:
This makes sure that the beginning of the selection is not rotated when
the mouse button is still pressed.

Fixes #3232.


kchibisov:
Blocked by #3315

chrisduerr:
This doesn't seem to fix things for me. I don't think there's much value in this PR anymore.

Issue #2487, opened 2019-05-28
tmux doesn't recognize alacritty


dinhtungdu:
First, thank you for this superb terminal :)

Which operating system does the issue occur on?
I'm using macOS Mojave 10.14.5

After I install alacritty terminfo as mentioned [here](https://github.com/jwilm/alacritty/blob/master/INSTALL.md#terminfo), I can't open tmux anymore. I killed all tmux session, restarted the machine also, but it didn't work.

This message shows up when I run tmux inside alacritty:
```
open terminal failed: missing or unsuitable terminal: alacritty
```
It seems that I installed the terminfo correctly because I can run imfocmp inside alacritty and get the following result. (click on Details to expand).
<details>

```
~ ❯❯❯ infocmp

# Reconstructed via infocmp from file: /usr/share/terminfo/61/alacritty
alacritty|alacritty terminal emulator,
	am, bce, ccc, hs, km, mc5i, mir, msgr, npc, xenl,
	colors#256, cols#80, it#8, lines#24, pairs#32767,
	acsc=``aaffggiijjkkllmmnnooppqqrrssttuuvvwwxxyyzz{{||}}~~,
	bel=^G, blink=\E[5m, bold=\E[1m, cbt=\E[Z, civis=\E[?25l,
	clear=\E[H\E[2J, cnorm=\E[?12l\E[?25h, cr=^M,
	csr=\E[%i%p1%d;%p2%dr, cub=\E[%p1%dD, cub1=^H,
	cud=\E[%p1%dB, cud1=^J, cuf=\E[%p1%dC, cuf1=\E[C,
	cup=\E[%i%p1%d;%p2%dH, cuu=\E[%p1%dA, cuu1=\E[A,
	cvvis=\E[?12;25h, dch=\E[%p1%dP, dch1=\E[P, dim=\E[2m,
	dl=\E[%p1%dM, dl1=\E[M, dsl=\E]2;\007, ech=\E[%p1%dX,
	ed=\E[J, el=\E[K, el1=\E[1K, flash=\E[?5h$<100/>\E[?5l,
	fsl=^G, home=\E[H, hpa=\E[%i%p1%dG, ht=^I, hts=\EH,
	ich=\E[%p1%d@, il=\E[%p1%dL, il1=\E[L, ind=^J,
	indn=\E[%p1%dS,
[...]

PR #3372, opened 2020-02-23
Explicitly state default TERM value in config

kchibisov:
Well, since the first value is `alacritty` and not `xterm-256color` we should likely explicitly state it in alacritty.yml placeholder value.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -10,7 +10,7 @@
   # each instance of Alacritty. If it is not present, alacritty will
   # check the local terminfo database and use `alacritty` if it is
   # available, otherwise `xterm-256color` is used.
-  #TERM: xterm-256color
+  #TERM: alacritty
 
 #window:
   # Window dimensions (changes require restart)


alacritty.yml
```diff
@@ -10,7 +10,7 @@
   # each instance of Alacritty. If it is not present, alacritty will
   # check the local terminfo database and use `alacritty` if it is
   # available, otherwise `xterm-256color` is used.
-  #TERM: xterm-256color
+  #TERM: alacritty
```
nixpulvis: I think this might be a good place to discuss https://github.com/alacritty/alacritty/issues/2487#issuecomment-587455831. Since people are often confused by this variable, and some applications do behave differently with it (I still use `TERM=xterm-256color`, shhhh).

Though I agree, we should be consistent here and have this set to the most specific thing alacritty can default to.

chrisduerr: > I think this might be a good place to discuss #2487 (comment). Since people are often confused by this variable, and some applications do behave differently with it

I'm not sure what there is to discuss. The only reason why people are confused is because they don't understand how terminal emulators work. If they desire to understand what `TERM` does, then there's plenty of documentation online for them to research.

nixpulvis: I was thinking that adding a line like "if you don't have the`alacritty` terminfo installed, `xterm-color256` may be used with limited issues."

In my eyes it would also be nice to gove an example where we need our terminfo instead of the generic xterm-color256, but we don't need to go into too much detail.

Why make our users hunt around online to understand our config file?

nixpulvis:
> If it is set to alacritty, the default behavior does not match what happens when the terminfo is installed, however it will still work mostly fine.

@chrisduerr, I'm a bit confused by this, since it seems to me that with the alacritty terminfo files provided by ncurses 6.2-1, I get the same behavior with and without setting the `TERM` value manually in our config file.

The issue is only if you don't have the terminfo installed, and you uncomment this configuration, which I think we all agree is fine to have a hard error. They can then decide if they need to install our terminfo, or simply set the value to `xterm-color256` and hope for the best.

chrisduerr:
> @chrisduerr, I'm a bit confused by this, since it seems to me that with the alacritty terminfo files provided by ncurses 6.2-1, I get the same behavior with and without setting the TERM value manually in our config file. This is expected, since we default to alacritty before xterm-color256.

The latest ncurses is not what you will find on most servers.

> The issue is only if you don't have the terminfo installed, and you uncomment this configuration, which I think we all agree is fine to have a hard error.

It's not a hard Alacritty error though. It's the applications inside of Alacritty that will cause problems then. That might manifest immediately, or days after this setting was changed.

alacritty.yml
nixpulvis: Your clearly missing my point. But I'll just have to look into this myself and maybe come up with a concrete suggestion. Maybe finally fixing my guicolors for vim in the process... last time I went down this rabbit hole I didn't make it too far, maybe I'll fare better this time.

PR #3378, opened 2020-02-25
fix #3377 changing scroll px info to f32

### LINKED ISSUE #3377: Trackpad Scroll events are ignored when scrolling really slow (less than one px per event)
As title.
This is caused by casting the `lpos.y` to an i32.

toger5:
`fn scroll_terminal(&mut self, new_scroll_px: f32) ` used a `i32` as the parameter. when scrolling slowly the inputs were ignored because 
```rust
MouseScrollDelta::PixelDelta(lpos)
print!(lpos.y) //this prints values less than 1 when scrolling really slow....
print!(lpos.y as i32) // this prints 0 when scrolling really slow
```


alacritty/src/event.rs
```diff
@@ -272,7 +272,7 @@ pub struct Mouse {
     pub right_button_state: ElementState,
     pub last_click_timestamp: Instant,
     pub click_state: ClickState,
-    pub scroll_px: i32,
+    pub scroll_px: f32,
```
chrisduerr: Might as well go for `f64` here, since that's the highest precision we receive.

alacritty/src/input.rs
```diff
@@ -469,31 +469,30 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
         match delta {
             MouseScrollDelta::LineDelta(_columns, lines) => {
                 let new_scroll_px = lines * self.ctx.size_info().cell_height;
-                self.scroll_terminal(new_scroll_px as i32);
+                self.scroll_terminal(new_scroll_px);
             },
             MouseScrollDelta::PixelDelta(lpos) => {
```
chrisduerr: Currently this returns an `lpos`, which means that it should be scaled by scale factor in theory.

I've sent a PR upstream to address this, so it should be fine for now:
https://github.com/rust-windowing/winit/pull/1483

alacritty/src/event.rs
```diff
@@ -292,7 +292,7 @@ impl Default for Mouse {
             middle_button_state: ElementState::Released,
             right_button_state: ElementState::Released,
             click_state: ClickState::None,
-            scroll_px: 0,
+            scroll_px: 0.0,
```
kchibisov:
```suggestion
            scroll_px: 0.,
```

alacritty/src/input.rs
```diff
@@ -469,31 +469,30 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
                 match phase {
                     TouchPhase::Started => {
                         // Reset offset to zero
-                        self.ctx.mouse_mut().scroll_px = 0;
+                        self.ctx.mouse_mut().scroll_px = 0.0;
```
kchibisov:
```suggestion
                        self.ctx.mouse_mut().scroll_px = 0.;
```

alacritty/src/input.rs
```diff
@@ -469,31 +469,30 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
-        let height = self.ctx.size_info().cell_height as i32;
+    fn scroll_terminal(&mut self, new_scroll_px: f32) {
+        let height = self.ctx.size_info().cell_height as f32;
 
         if self.ctx.terminal().mode().intersects(TermMode::MOUSE_MODE) {
             self.ctx.mouse_mut().scroll_px += new_scroll_px;
 
-            let code = if new_scroll_px > 0 { 64 } else { 65 };
-            let lines = (self.ctx.mouse().scroll_px / height).abs();
+            let code = if new_scroll_px > 0.0 { 64 } else { 65 };
```
kchibisov:
```suggestion
            let code = if new_scroll_px > 0. { 64 } else { 65 };
```

alacritty/src/input.rs
```diff
@@ -514,8 +513,8 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
             );
             self.ctx.mouse_mut().scroll_px += new_scroll_px * multiplier;
 
-            let cmd = if new_scroll_px > 0 { b'A' } else { b'B' };
-            let lines = (self.ctx.mouse().scroll_px / height).abs();
+            let cmd = if new_scroll_px > 0.0 { b'A' } else { b'B' };
```
kchibisov:
```suggestion
            let cmd = if new_scroll_px > 0. { b'A' } else { b'B' };
```

alacritty/src/event.rs
toger5: Also was thinking about that. Wasnt sure though. (updated now)

alacritty/src/input.rs
toger5: thanks for the nightly information. its all properly fomratted now.

CHANGELOG.md
```diff
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `draw_bold_text_with_bright_colors` is now `false`
 - Mirror OSC query terminators instead of always using BEL
 - Increased Beam, Underline, and Hollow Block cursors' line widths
-
+- Pixel scrolling now uses f64. Slow trackpad scrolling is not anymore ignored.
```
kchibisov: It should be in a `Fixed` section. Also, you should avoid implementation details in a `CHANGELOG` entries, since we're not a library. Something like `Ignoring of slow touchpad scrolling` could be better.

alacritty/src/input.rs
nixpulvis: Once this is merged we shouldn't need any more `as f64` casts hopefully. :tada:

nixpulvis:
Wow, this is somewhat amazing to me. I only never noticed this on my laptop because it seems to be randomly throwing in > 1 valued scroll events regardless of how little I scroll. I suppose it's hard for me to still rule out my hand shaking reliably or something, but I don't think that's it :stuck_out_tongue_winking_eye:

I've given this whole situation some thought, and replied to rust-windowing/winit#22, however this by no means blocks this PR.

toger5:
@kchibisov Thank you sooo much! I have very little time atm. So it would have taken a couple more days until i would have checked it again.

PR #3379, opened 2020-02-25
Add kinetic scrolling

toger5:
Kinetic scrolling when using the trackpad as the input device.

I am not at all certain about the implementation (new to both: the codebase of alacritty and winit...)
Feedback is very welcome.


toger5:
I tried running `cargo fmt` Maybe i need to switch to the nightly rust channel?

toger5:
should be much better now. I still think i need a nightly channel, because I get a bunch of warnings:
```
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `format_code_in_doc_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `comment_width = 100`, unstable features are only available in nightly channel.
Warning: can't set `normalize_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `format_strings = true`, unstable features are only available in nightly channel.
Warning: can't set `reorder_impl_items = true`, unstable features are only available in nightly channel.
Warning: can't set `overflow_delimited_expr = true`, unstable features are only available in nightly channel.
Warning: can't set `match_block_trailing_comma = true`, unstable features are only available in nightly channel.
Warning: can't set `condense_wildcard_suffixes = true`, unstable features are only available in nightly channel.
```

alacritty/src/input.rs
```diff
@@ -469,31 +469,60 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
                 match phase {
                     TouchPhase::Started => {
                         // Reset offset to zero
-                        self.ctx.mouse_mut().scroll_px = 0;
+                        self.ctx.mouse_mut().scroll_px = 0.0;
                     },
                     TouchPhase::Moved => {
-                        self.scroll_terminal(lpos.y as i32);
+                        // adapt for setting trackpad speed using 60.0 as fps factor...
+                        let multiplier = 60.0
```
chrisduerr: This cannot be assumed.

nixpulvis:
Correct me if I'm wrong, but isn't this something that should be implemented at the device driver level? I'd expect to have this either enable for all applications or not at all.

toger5:
@chrisduerr The advantage is that you can "fling" the terminal to go up by a lot of lines. Like you are used to in the browser or document viewer or whatever...
When i fix the compilation issues you should try it out. it just feels nice.
@nixpulvis That is a design decision by the wayland protocol. It makes a lot of sense to let the applictaion implement kinetic scrolling. When this is handled by the compositor the the app has no way knowing if the user is scrolling or if its getting fake events from the compositor. For showing an indicater that the bottom of a scroll view has been reached you would want to know if the user is actually trying to scroll further down.

chrisduerr:
I see this as problematic, especially because it will break other platforms that implement acceleration of scrolling on driver level. Those would just multiply which would cause big troubles.

I don't see any good reason why this should be implemented in Alacritty directly.

nixpulvis:
@toger5 Ah exactly what I want, every application implementing a slightly different feeling scroll acceleration. I'll never know what to do with my fingers again, and my whole life will crash and burn.

All drama aside, this seems like something we **really** want to standardize. Sure, there should be an API for enabling/disabling this feature, but applications should not be the one's dealing with the implementation.

toger5:
I agree it has some disadvantages. The idea is, that the toolkits implement kintetic scrolling so the actual aplication developer does not have to deal with it. So it would make much more sense to add it to winit. I just really wanted that feature now so i added it. And created a pr to let you guys know.

chrisduerr:
> So it would make much more sense to add it to winit. I just really wanted that feature now so i added it.

Then you should try to implement this for winit and see if it's something they desire. Of course it's perfectly fine to run a patched Alacritty yourself, but I'm not big on adding hacks to get things in faster, rather than doing it right.

toger5:
@chrisduerr sounds reasonable.  Sorry for the noise than.

PR #3382, opened 2020-02-26
Remove unnecessary macos focus check

chrisduerr:
Since winit now filters out clicks on unfocused windows on macOS
anyways, we don't need to implement this logic ourselves anymore.


chrisduerr:
I want to re-test this tomorrow at work, just to make sure the latest winit change hasn't removed this again, since last time I tested this diff was a while ago.

Issue #3381, opened 2020-02-26
Interpret integers for color configurations


FichteFoll:
Currently, colors must be provided as strings as verify their formatting, which means instead of writing `background: 0x0a0a0a` you have to write `background: '0x0a0a0a'`, since YAML automatically parses hexadecimal numbers starting with `0x`. However, I believe this to be unnecessary as interpreting an integer number `0x0a0a0a` as an RGB24 color is absolutely trivial.

Is there any particular reason for why this hasn't been done or was it just deemed unnecessary? Based on the error log, there is specific error checking code for this.
Maybe a limitation of the yaml parser library and its integration?

chrisduerr:
> Is there any particular reason for why this hasn't been done or was it just deemed unnecessary? 

Why would we support this? It would just make things more complicated and require additional code, just so we can handle something that isn't really necessary?

We already support two different formats, that should be enough.

kchibisov:
We're leaving for a long time with '0xffffff' and '#ffffff' ( I use the later one for example), so there's unlikely a chance old users start using new syntax, I'm not sure that a lot of users are using '#ffffff' form. Adding one more similar option doesn't make any sense to me.

nixpulvis:
See #3383

nixpulvis:
I'm torn on this issue.

It's yet another case of simplicity of configuration vs simplicity of code, and a trade-off must be made. I would personally accept a patch that implemented this cleanly, but I'm not sure others would agree.

If I had things my way, the format `0x0a0a0a` would be numeric (not requiring quotes), and the format `#0a0a0a` would be a string.

kchibisov:
I mean it doesn't make sense to have one more format, you can't use `0x0a0a0a` format instead of `'0x0a0a0a'` format since mostly all config files are using the later one, so you'll break tons of configs.

nixpulvis:
@kchibisov we can deprecate things, breaking configs shouldn't completely block all improvements to the config. Users shouldn't expect the config file to be stable until we say it is.

kchibisov:
I don't see any issue in having surrounding `'` tbh, since we still continue to ban `0x0` things. And string makes more sense for color than arbitrary num in a config. I'd argue to just keep '#rrggbb' format though.

nixpulvis:
> I don't see any issue in having surrounding ' tbh

`less` is `more` :)

chrisduerr:
> less is more :)

Exactly, which is why we shouldn't try to support even one more format.

[...]

PR #3383, opened 2020-02-27
Add notes about '#rrggbb' notation

kchibisov:


alacritty.yml
```diff
@@ -180,6 +180,10 @@
 #draw_bold_text_with_bright_colors: false
 
 # Colors (Tomorrow Night Bright)
+#
+# Color's value could be specified either with '0xrrggbb' or with '#rrggbb'
+# notation.
```
cole-h:
```suggestion
# Colors can be specified using either '0xrrggbb' or '#rrggbb` notation.
```

kchibisov: Not sure about dropping notation, but I was picking between `Colors` and `Color's value` originally, due to use of `colors` outside of `colors` section, so decided to avoid confusion, however the comment is above colors section, so I feel like it should be fine to use your suggestion here.

chrisduerr:
See my thoughts on this: https://github.com/alacritty/alacritty/issues/3381#issuecomment-592158107.

nixpulvis:
@chrisduerr 100% agree.

kchibisov:
closing in favor of #3392

PR #3384, opened 2020-02-27
Fix message bar not hiding after fixing color in config

kchibisov:

### DIFF
--- a/alacritty_terminal/src/term/color.rs
+++ b/alacritty_terminal/src/term/color.rs
@@ -103,7 +103,7 @@ impl FromStr for Rgb {
     type Err = ();
 
     fn from_str(s: &str) -> std::result::Result<Rgb, ()> {
-        let chars =  if s.starts_with("0x") && s.len() == 8 {
+        let chars = if s.starts_with("0x") && s.len() == 8 {
             &s[2..]
         } else if s.starts_with('#') && s.len() == 7 {
             &s[1..]
--- a/alacritty_terminal/src/term/color.rs
+++ b/alacritty_terminal/src/term/color.rs
@@ -7,7 +7,7 @@ use serde::de::Visitor;
 use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::ansi;
-use crate::config::Colors;
+use crate::config::{Colors, LOG_TARGET_CONFIG};
 
 pub const COUNT: usize = 269;
 
@@ -85,7 +85,10 @@ impl<'de> Deserialize<'de> for Rgb {
         match value.deserialize_str(RgbVisitor) {
             Ok(rgb) => Ok(rgb),
             Err(err) => {
-                error!("Problem with config: {}; using color #000000", err);
+                error!(
+                    target: LOG_TARGET_CONFIG,
+                    "Problem with config: {}; using color #000000", err
+                );
                 Ok(Rgb::default())
             },
         }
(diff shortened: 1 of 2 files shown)


alacritty_terminal/src/term/color.rs
```diff
@@ -85,7 +85,10 @@ impl<'de> Deserialize<'de> for Rgb {
         match value.deserialize_str(RgbVisitor) {
             Ok(rgb) => Ok(rgb),
             Err(err) => {
-                error!("Problem with config: {}; using color #000000", err);
+                error!(
+                    target: LOG_TARGET_CONFIG,
+                    "Problem with config: {}; using color #000000", err
```
nixpulvis: Either change the format of the color in the `err`, or use `0x000000` here.

kchibisov: I wonder why we should change it? It's print that you have 2 options to set color and will say that it's using `#000000`.

kchibisov: Oh, it depends on  #3383

nixpulvis: I could leave this comment on either PR, but this error message is starting to get rather long... I'd personally rather standardize on one, and then just leave the other in the config doc comments.

kchibisov: Hm, let me try something.

kchibisov: Hm, after trying to use the same color prefix format as the user specified it seems like it's pretty ugly and hacky, however it's possible to do something like that. For now I've decided to error that we have both formats.

>I'd personally rather standardize on one, and then just leave the other in the config doc comments.

Using only one format could be confusing and it could also force user to stick with the one format over another in the end.

nixpulvis: > Using only one format could be confusing and it could also force user to stick with the one format over another in the end.
 
I don't see any problem with this. Personally I'd be happy with everything being `#a0a0a0`, but I don't actually care nearly at all on this issue.

nixpulvis: The complexity of matching the users own format is definitely not worth it, there are much larger config issues deserving of complexity than this.

chrisduerr: See https://github.com/alacritty/alacritty/issues/3381#issuecomment-592158107.

That would solve all of this by just dropping all `0xRRGGBB` documentation.

kchibisov: Synced.

alacritty_terminal/src/term/color.rs
```diff
@@ -127,7 +134,12 @@ impl FromStr for Rgb {
 
         component!(r, g, b);
 
-        Ok(rgb)
+        if chars.as_str().is_empty() {
+            Ok(rgb)
+        } else {
+            // We still have something to parse, so throw an error
+            Err(())
+        }
```
chrisduerr: I wonder if it wouldn't be much easier to check the length before starting the parsing.

That way we can just skip a lot of things.

I was thinking something like this, wrt method structure:

```rust
let chars = if input.starts_with("0x") && input.len() == 8 {
} else if input.starts_with("#") && input.len() == 7 {
} else {
     return Err(());
};
```

Then we can potentially even clean up the macro a bit, or remove it completely. We can also change it from a chars iterator to a slice reference, by returning something like `&input[1..]`, because that way we can directly use `slice[1]` to index it, instead of having to use `chars.next()`.

Just some ideas though, I think it would be possible to simplify and shorten things up a bit.

PR #3390, opened 2020-02-27
Fix crash on Weston 8.0 during startup

kchibisov:


chrisduerr:
What exactly is the reason behind this crash? I'm just wondering if we could maybe have a more general change log entry about which kind of crash this fixes.

kchibisov:
Well, the issue is binding `wl_seat` version 7, however SCTK isn't capable of doing it, so we should cap version at `6`, smithay-clipboard was just binding the thing right away without version check, so it was failing.

chrisduerr:
Why is this limited to Weston?

kchibisov:
I tried to make entry that actually make sense for users, since I'm not aware of any other compositors implementing v7, besides Weston.

CHANGELOG.md
```diff
@@ -55,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Block selection appending duplicate newlines when last column is selected
 - Bitmap fonts being a bit smaller than they should be in some cases
 - Config reload creating alternate screen history instead of updating scrollback
+- Crash on Weston 8.0 during startup
```
chrisduerr: > I tried to make entry that actually make sense for users, since I'm not aware of any other compositors implementing v7, besides Weston.

Yeah, but that's *right now*. If our next version is released and in the repos, this might be different? Or if someone, for some reason, runs an outdated version and suddenly nothing is working anymore.

I don't think it's a good idea that our change log entries depend on when the changelog was written.

```suggestion
- Crash on compositors supporting version 7+ of Wayland's input protocol
```

PR #3392, opened 2020-02-27
Remove 0xRRGGBB color notation

chrisduerr:
Closes #3383.


chrisduerr:
I'd like to get a unanimous decision on this, but as far as I can tell there is never a reason why the `0xRRGGBB` notation would be preferred over `#RRGGBB`.

nixpulvis:
Double check the `colors` in `alacritty.info`.

PR #3396, opened 2020-02-28
Move debug to bottom of config

chrisduerr:
Since this information is the least relevant to the user, it should be
the last thing in the configuration file.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -338,30 +338,6 @@
 # Send ESC (\x1b) before characters when alt is pressed.
 #alt_send_esc: true
 
-#debug:
-  # Display the time it takes to redraw each frame.
-  #render_timer: false
-
-  # Keep the log file after quitting Alacritty.
-  #persistent_logging: false
-
-  # Log level
-  #
-  # Values for `log_level`:
-  #   - None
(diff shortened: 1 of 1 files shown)


alacritty.yml
```diff
@@ -570,3 +546,27 @@
...
+  #   - Info
+  #   - Debug
+  #   - Trace
+  #log_level: Warn
+
+  # Print all received window events.
+  #print_events: false
+
+  # Record all characters and escape sequences as test data.
+  #ref_test: false
```
kchibisov: If you're decided to touch debug, could we remove `ref_test` setting?

chrisduerr: Why?

kchibisov: It doesn't make any sense to have this setting in a config, it will just create random permanent files for you, so making it CLI only makes sense.

chrisduerr: Can we do this separately? Thinking about it, I really don't want to do anything other than just moving this section around in this PR.

kchibisov: Yeah, we can.

P.s.

Don't be so eager, I could be sleeping for a bit.

PR #3398, opened 2020-02-28
Fix tabs across linewrap

chrisduerr:
This resolves an issue with tabs not breaking across line boundaries,
instead the characters would just all get written to the last column and
thus be lost.

It also tweaks the behavior of what happens when the terminal resizes
with the default tabspaces changed, using something like the `tabs`
program. Previously all tabstops would be reset to the default on
resize, which is what URxvt does. Now the tabspaces are kept and the new
columns are filled with the default tabstops, which emulates Termite.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -44,7 +44,7 @@ pub mod color;
 /// Used to match equal brackets, when performing a bracket-pair selection.
 const BRACKET_PAIRS: [(char, char); 4] = [('(', ')'), ('[', ']'), ('{', '}'), ('<', '>')];
 
-/// Max size of the window title stack
+/// Max size of the window title stack.
 const TITLE_STACK_MAX_DEPTH: usize = 4096;
 
 /// A type that can expand a given point to a region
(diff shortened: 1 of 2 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -1542,9 +1538,14 @@ impl<T: EventListener> Handler for Term<T> {
...
     }
 
+    /// Insert tab at cursor position.
     #[inline]
     fn put_tab(&mut self, mut count: i64) {
-        trace!("Putting tab: {}", count);
+        // A tab after the last column is the same as a linebreak
+        if self.input_needs_wrap {
+            self.wrapline();
+            return;
```
kchibisov: I wasn't able to reach this code, could you share an example on how to do this?

chrisduerr: `echo "\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt\tt"`

chrisduerr: This is about tabs breaking across linewraps, so of course you need to make sure there are tabs breaking across lines.

kchibisov: Oh, I see. I was doing similar thing but it wasn't working for me when I tried it first time for some reason.

PR #3399, opened 2020-02-29
Update terminfo dependency

kchibisov:


kchibisov:
>Is there any concrete issue we're fixing by bumping it?

I don't think so, I've just updated it.

PR #3415, opened 2020-03-03
Remove `ref_test` option from configuration file

chrisduerr:
This removes the `debug.ref_test` option from the configuration file,
after this change was originally requested from kchibisov in
https://github.com/alacritty/alacritty/pull/3396.

While this option is valueable for the CLI, it provides no value in the
configuration file.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -558,6 +558,3 @@
 
   # Print all received window events.
   #print_events: false
-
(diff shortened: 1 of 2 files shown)


chrisduerr:
> You should also remove it from code, just removing mentioning of it doesn't make much value.

Huh, I must not have added all files. Should be good now.

PR #3416, opened 2020-03-04
Fix caching variable font faces

kchibisov:
Fixes regression introduced in 84f57ac8367e3e24e0347515e6681c0478fc3d29.

### DIFF
--- a/font/src/ft/fc/mod.rs
+++ b/font/src/ft/fc/mod.rs
@@ -41,7 +41,7 @@ pub mod char_set;
 pub use char_set::{CharSet, CharSetRef};
 
 pub mod pattern;
-pub use pattern::{Pattern, PatternHash, PatternRef};
+pub use pattern::{FTFaceLocation, Pattern, PatternHash, PatternRef};
 
 /// Find the font closest matching the provided pattern.
 ///
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -318,7 +318,7 @@ impl FreeTypeRasterizer {
                 return Ok(Some(font_key));
             }
 
-            trace!("Got font path={:?}", ft_face_location.path);
+            trace!("Got font path={:?}, index={:?}", ft_face_location.path, ft_face_location.index);
 
             let ft_face = match self.ft_faces.get(&ft_face_location) {
                 Some(ft_face) => Rc::clone(ft_face),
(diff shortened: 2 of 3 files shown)


font/src/ft/mod.rs
```diff
@@ -206,6 +206,18 @@ struct FullMetrics {
     cell_width: f64,
 }
 
+#[derive(Hash, Eq, PartialEq, Debug)]
+struct FTFaceLocation {
+    path: PathBuf,
+    index: isize,
```
chrisduerr: What exactly is this mysterious `index` and why is it necessary to keep that around as part of the font path?

Is this related to OTF fonts and it's basically a selection for which face inside a path should be loaded?

Also couldn't this entire `FTFaceLocation` struct be moved to `pattern.rs`, so we would only have a `.location()` method instead of having to load both the path and the index, introducing the need to combine them whenever calling at and adding potential for misuse?

kchibisov: >What exactly is this mysterious index and why is it necessary to keep that around as part of the font path?

It's an index of a face in a font file. Most of the time it's just 0, but for variable fonts it could have multiple of them. So index X will be for normal face, but index Y for bold face.

>Is this related to OTF fonts and it's basically a selection for which face inside a path should be loaded?

Yes.

>Also couldn't this entire FTFaceLocation struct be moved to pattern.rs, so we would only have a .location() method instead of having to load both the path and the index, introducing the need to combine them whenever calling at and adding potential for misuse?

Probably, it could simplify logic a bit, since we need both of them.

chrisduerr: Sure, let's move it then to make it a little easier to use.

font/src/ft/fc/mod.rs
```diff
@@ -41,7 +41,7 @@ pub mod char_set;
 pub use char_set::{CharSet, CharSetRef};
 
 pub mod pattern;
-pub use pattern::{Pattern, PatternHash, PatternRef};
+pub use pattern::{FTFaceLocation, Pattern, PatternHash, PatternRef};
```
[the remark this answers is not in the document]
kchibisov: I need this thing here as export.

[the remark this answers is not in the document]
chrisduerr: Didn't see it was a `pub use`, you're right of course.

PR #3417, opened 2020-03-04
Remove gui dependencies from alacritty_terminal

### LINKED ISSUE #3393: Decouple gui-related bits out of 'alacritty_terminal'
Even after splitting Alacritty into 2 parts (`alacritty` and `alacritty_terminal`) we still have some bits that are make sense only for gui apps, like dependency to `font` and `copypasta`. Font dependency is purely config related and `copypasta`  could be replaced by creating a clipboard trait in `alacritty_terminal` and implementing it in `alacritty`. Ideally we should move our config related bits out of `alacritty_terminal` into `alacritty`, however I'm not sure how much of it we should move out.

There's also one part with `RenderableCell`'s which should likely be moved somehow out of `term/mod.rs`, but I don't have a good idea on how to do that.

kchibisov:
This commit removes 'font' dependency from alacritty_terminal,
so it'll simplify the usage of alacritty_terminal as a library,
since you won't link to system's libraries anymore. It also
moves many alacritty related config options from it.

Fixes #3393.


alacritty/src/config/window.rs
```diff
@@ -126,15 +128,15 @@ impl Default for Decorations {
...
 #[serde(default)]
 #[derive(Default, Debug, Copy, Clone, Deserialize, PartialEq, Eq)]
 pub struct Dimensions {
-    /// Window width in character columns
+    /// Window width in character column.
     #[serde(deserialize_with = "failure_default")]
     columns: Column,
 
-    /// Window Height in character lines
+    /// Window Height in character line.
```
cole-h: Pretty sure these two (`column` and `line`) should still be plural. Also maybe you could lowercase `Height` to bring it in line with the above comment?
```suggestion
    /// Window height in character lines.
```

alacritty/src/config/window.rs
```diff
@@ -126,15 +128,15 @@ impl Default for Decorations {
 
 /// Window Dimensions
 ///
-/// Newtype to avoid passing values incorrectly
+/// Newtype to avoid passing values incorrectly.
 #[serde(default)]
 #[derive(Default, Debug, Copy, Clone, Deserialize, PartialEq, Eq)]
 pub struct Dimensions {
-    /// Window width in character columns
+    /// Window width in character column.
```
cole-h:
```suggestion
    /// Window width in character columns.
```

alacritty/src/renderer/mod.rs
```diff
@@ -22,23 +23,29 @@ use std::sync::mpsc;
 use std::time::Duration;
 
 use fnv::FnvHasher;
+
 use font::{
     self, BitmapBuffer, FontDesc, FontKey, GlyphKey, Rasterize, RasterizedGlyph, Rasterizer,
 };
+
 use log::{error, info};
+
```
cole-h: Not really a fan of these newlines. Dunno why, they just irk me (not a real reason to change anything, I know).

Going off some of the other imports I looked at, it might look better if the `log` and `fnv` imports were grouped together and `font` separated by a newline, instead of all 3 separated by a newline.

alacritty_terminal/src/term/mod.rs
```diff
@@ -816,14 +816,17 @@ pub struct Term<T> {
...
-    clipboard: Clipboard,
+    clipboard: Box<dyn Clipboard>,
 
     /// Proxy for sending events to the event loop
     event_proxy: T,
 
     /// Current title of the window
     title: String,
 
+    /// Title used during creation.
```
cole-h: Maybe I'm nitpicking the wrong things here, but this is slightly too vague for me. Maybe something more like:
```suggestion
    /// Default title used upon window creation.
```

---

It may be the late hour getting to me, but what is the benefit of this/why is it necessary?

alacritty/src/config/window.rs
kchibisov: My vim macro failed here.

alacritty/src/config/window.rs
```diff
@@ -2,58 +2,60 @@ use std::os::raw::c_ulong;
...
 
-    /// Startup mode
+    /// Startup mode.
     #[serde(deserialize_with = "failure_default")]
     startup_mode: StartupMode,
 
-    /// Window title
+    /// Window title.
     #[serde(default = "default_title")]
     pub title: String,
```
kchibisov: I'm not sure where to put this thing. It makes sense as for a window title, but on the other hand terminal also has its own title and in some cases they could be different (imagine tmux like program with multiple panes). Each pane is a terminal and has it's own title, however the resulted title is a title of the currently active pane.

alacritty/src/config/monitor.rs
```diff
@@ -10,7 +10,7 @@ use alacritty_terminal::util;
 use crate::event::EventProxy;
 
 pub struct Monitor {
-    _thread: ::std::thread::JoinHandle<()>,
+    _thread: std::thread::JoinHandle<()>,
```
chrisduerr: If these are touched, we might as well properly import `std::thread` at the top of the file.

alacritty/src/config/monitor.rs
```diff
@@ -26,7 +26,7 @@ impl Monitor {
                 // The Duration argument is a debouncing period.
                 let mut watcher =
                     watcher(tx, Duration::from_millis(10)).expect("Unable to spawn file watcher");
-                let config_path = ::std::fs::canonicalize(path).expect("canonicalize config path");
+                let config_path = std::fs::canonicalize(path).expect("canonicalize config path");
```
chrisduerr: If these are touched, we might as well properly import `std::fs` at the top of the file.

alacritty/src/config/mouse.rs
```diff
@@ -41,7 +41,7 @@ impl Url {
 
 fn deserialize_launcher<'a, D>(
     deserializer: D,
-) -> ::std::result::Result<Option<CommandWrapper>, D::Error>
+) -> std::result::Result<Option<CommandWrapper>, D::Error>
```
chrisduerr: We can just use `Result`.

alacritty/src/config/mouse.rs
```diff
@@ -100,7 +100,7 @@ fn default_threshold_ms() -> Duration {
     Duration::from_millis(300)
 }
 
-fn deserialize_duration_ms<'a, D>(deserializer: D) -> ::std::result::Result<Duration, D::Error>
+fn deserialize_duration_ms<'a, D>(deserializer: D) -> std::result::Result<Duration, D::Error>
```
chrisduerr: We can just use `Result`.

alacritty/src/renderer/mod.rs
```diff
@@ -21,24 +22,28 @@ use std::ptr;
...
 use std::time::Duration;
 
-use fnv::FnvHasher;
 use font::{
     self, BitmapBuffer, FontDesc, FontKey, GlyphKey, Rasterize, RasterizedGlyph, Rasterizer,
 };
+
+use fnv::FnvHasher;
 use log::{error, info};
 use notify::{watcher, DebouncedEvent, RecursiveMode, Watcher};
```
chrisduerr: `use font` should be below foreign crates and before `use crate`.

The `alacritty_terminal` should also be above `crate`, but below foreign crates or our own crates.

alacritty_terminal/src/config/mod.rs
```diff
@@ -22,124 +22,82 @@ use serde::{Deserialize, Deserializer};
...
-#[derive(Debug, PartialEq, Default, Deserialize)]
+/// Top-level config type.
+#[derive(Debug, Default, PartialEq, Deserialize)]
 pub struct Config<T> {
-    /// Pixel padding
-    #[serde(default, deserialize_with = "failure_default")]
-    pub padding: Option<Delta<u8>>,
-
-    /// TERM env variable
+    /// TERM env variable.
```
chrisduerr: This comment is kinda inaccurate.

alacritty_terminal/src/config/mod.rs
```diff
@@ -280,51 +204,8 @@ impl FromString for Option<Shell<'_>> {
...
-    where
-        D: Deserializer<'a>,
-    {
-        Ok(Alpha::new(f32::deserialize(deserializer)?))
-    }
-}
-
 #[derive(Deserialize, Copy, Clone, Debug, PartialEq, Eq)]
-struct DefaultTrueBool(bool);
+pub struct DefaultTrueBool(pub bool);
```
chrisduerr: Why is this made public? This shouldn't be used for anything other than deserialization.

alacritty_terminal/src/term/mod.rs
```diff
@@ -816,14 +816,17 @@ pub struct Term<T> {
     dynamic_title: bool,
 
     /// Clipboard access coupled to the active window
-    clipboard: Clipboard,
+    clipboard: Box<dyn Clipboard>,
```
chrisduerr: Do we have to use dynamic dispatch here? I'd rather stick with static dispatch whenever possible.

With the `Term<T, C = NopClipboard>` syntax it might be possible to prevent having to edit a bunch of unrelated code.

alacritty_terminal/src/term/mod.rs
```diff
@@ -909,7 +912,8 @@ impl<T> Term<T> {
     pub fn new<C>(
         config: &Config<C>,
         size: &SizeInfo,
-        clipboard: Clipboard,
+        title: Option<String>,
```
chrisduerr: Do we really want to make this optional, just to fallback to a fixed value? What's the intended benefit here?

alacritty_terminal/src/clipboard.rs
```diff
@@ -12,53 +12,9 @@
...
-        Self { clipboard: Box::new(NopClipboardContext::new().unwrap()), selection: None }
-    }
-}
-
-impl Default for Clipboard {
-    fn default() -> Self {
-        Self { clipboard: Box::new(ClipboardContext::new().unwrap()), selection: None }
-    }
+pub trait Clipboard: Send {
+    fn store(&mut self, ty: ClipboardType, text: String);
```
chrisduerr: Why did you change it from `Into<String>` to just `String`? If anything, it might even be more appropriate to take just `&str`.

alacritty_terminal/src/term/mod.rs
kchibisov: I still wonder what we should do with title. If you look at CI failure you'll notice that we're using it in ConPTY. I decided to use `Option` to simplify tests, however maybe we should put it on a config.

alacritty_terminal/src/term/mod.rs
kchibisov: I don't see any issue with dynamic dispatch for clipboard, but we can use static one here, sure.

alacritty_terminal/src/config/mod.rs
kchibisov: I'll just copy this to alacritty then, I was trying to avoid coping of this thing, but it's likely make sense to keep this private.

alacritty_terminal/src/clipboard.rs
kchibisov: It was some issue with `trait` that forces me to do so, I feel like `&str` makes more sense though.

alacritty_terminal/src/term/mod.rs
kchibisov: I've updated to use static dispatch, but I don't really like how it looks and how much it requires to change/carry around. Dynamic dispatch for clipboard made much more sense, since it's not something critical for performance and was integrated naturally.

alacritty_terminal/src/term/mod.rs
kchibisov: I don't see how it answers my question.

The problem is that I don't want to plumb title through the PTY interface only because of Windows. So I was thinking on storing it on a config, which could simplify it a bit.

chrisduerr: I mean if we need the title, I don't see anything wrong with passing it through to the PTY.

I'd have to actually see what the title does on ConPTY to really judge if we actually need it though.

If https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/ns-processthreadsapi-startupinfoa is accurate, this is what it is used for:

> For console processes, this is the title displayed in the title bar if a new console window is created. If NULL, the name of the executable file is used as the window title instead. This parameter must be NULL for GUI or console processes that do not create a new console window.

So it's questionable if this is necessary at all for Alacritty, since we spawn our own Window with its own title. So maybe we don't actually have any problem?

kchibisov: I'd go with it for now.

cc @davidhewitt

kchibisov:
I've updated most bits I've memorized. Should I also update `dynamic_title` logic.

chrisduerr:
> Should I also update dynamic_title logic.

I'm not entirely certain what exactly you mean, but I'd say you should, yes. The logic itself should still behave the way it does on the latest master though, since that's what we've decided on.

davidhewitt:
> So it's questionable if this is necessary at all for Alacritty, since we spawn our own Window with its own title. So maybe we don't actually have any problem?

IIRC, with ConPTY, `cmd.exe` processes will make use of this startup info, but I don't think powershell does. (As discussed on the other thread about dynamic title on Windows.)

chrisduerr:
> IIRC, with ConPTY, cmd.exe processes will make use of this startup info, but I don't think powershell does. (As discussed on the other thread about dynamic title on Windows.)

@davidhewitt We've already figured out in another issue that it does have an effect on the title, since it would link to the full Alacritty executable.

But if you aren't aware of any issues, I feel like it might be a nice option to remove this, so that the Alacritty PTY does not need to know anything about the default title anymore.

alacritty/src/config/ui_config.rs
```diff
@@ -48,6 +44,11 @@ pub struct UIConfig {
     /// Debug options.
     #[serde(default, deserialize_with = "failure_default")]
     pub debug: Debug,
+
+    // TODO - DEPRECATED
+    /// Use dynamic title.
```
cole-h:
```suggestion
    // TODO: DEPRECATED
```

Follows what our deprecation comments look like everywhere else. Also, fairly certain we remove doc comments when we deprecate something.

alacritty/src/input.rs
```diff
@@ -35,18 +35,18 @@ use glutin::window::CursorIcon;
 
 use alacritty_terminal::ansi::{ClearMode, Handler};
 use alacritty_terminal::clipboard::ClipboardType;
-use alacritty_terminal::event::{Event, EventListener};
+use alacritty_terminal::event::EventListener as TerminalEventListener;
```
chrisduerr, line 38: I don't think this is necessary?

kchibisov:
Reworked the entire thing.

alacritty/src/cli.rs
```diff
@@ -263,29 +265,34 @@ impl Options {
...
-        config.debug.log_level = max(config.debug.log_level, self.log_level);
-        config.debug.ref_test = self.ref_test || config.debug.ref_test;
-        config.debug.persistent_logging =
-            self.persistent_logging || config.debug.persistent_logging;
-
-        if config.debug.print_events {
-            config.debug.log_level = max(config.debug.log_level, LevelFilter::Info);
+        let ui_config = &mut config.ui_config;
+        ui_config.set_dynamic_title(ui_config.dynamic_title() && self.title.is_none());
+        let window_config = &mut ui_config.window;
```
chrisduerr, line 276: I'm not sure if extracting the `ui_config` into a variable is worth it? Seems to be infrequently used and short enough to just be accessed directly.

alacritty/src/cli.rs
```diff
@@ -263,29 +265,34 @@ impl Options {
...
+        window_config.dimensions = self.dimensions.unwrap_or(window_config.dimensions);
+        window_config.title = self.title.unwrap_or_else(|| window_config.title.clone());
+        window_config.position = self.position.or(window_config.position);
+        window_config.embed = self.embed.and_then(|embed| embed.parse().ok());
+
+        window_config.class.instance =
+            self.class_instance.unwrap_or_else(|| window_config.class.instance.clone());
+        window_config.class.general =
+            self.class_general.unwrap_or_else(|| window_config.class.general.clone());
```
chrisduerr, line 285: We might be able to skip these clones if we access window_config directly?

alacritty/src/display.rs
```diff
@@ -245,8 +246,9 @@ impl Display {
         renderer: &mut QuadRenderer,
         config: &Config,
     ) -> Result<(GlyphCache, f32, f32), Error> {
-        let font = config.font.clone();
-        let rasterizer = font::Rasterizer::new(dpr as f32, config.font.use_thin_strokes())?;
+        let font = config.ui_config.font.clone();
+        let rasterizer =
+            font::Rasterizer::new(dpr as f32, config.ui_config.font.use_thin_strokes())?;
```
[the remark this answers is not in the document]
chrisduerr: Why did you resolve this? Seems like removing the `font::` should be useful here?

alacritty/src/display.rs
```diff
@@ -304,8 +306,10 @@ impl Display {
         let cell_height = self.size_info.cell_height;
 
         // Recalculate padding
-        let mut padding_x = f32::from(config.window.padding.x) * self.size_info.dpr as f32;
-        let mut padding_y = f32::from(config.window.padding.y) * self.size_info.dpr as f32;
+        let mut padding_x =
+            f32::from(config.ui_config.window.padding.x) * self.size_info.dpr as f32;
+        let mut padding_y =
+            f32::from(config.ui_config.window.padding.y) * self.size_info.dpr as f32;
```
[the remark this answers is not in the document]
chrisduerr: I still think this might be a good idea.

alacritty/src/main.rs
```diff
@@ -59,9 +60,11 @@ use crate::config::monitor::Monitor;
 use crate::config::Config;
 use crate::display::Display;
 use crate::event::{Event, EventProxy, Processor};
+use crate::message_bar::MessageBuffer;
 
 fn main() {
-    panic::attach_handler();
+    #[cfg(windows)]
+    crate::panic::attach_handler();
```
chrisduerr, line 67: Do we want to import `crate::panic`? Accessing things with the full path directly using `crate::` seems very rare in our codebase.

alacritty_terminal/src/config/mod.rs
```diff
@@ -320,6 +218,10 @@ impl Percentage {
             value
         })
     }
+
+    pub fn raw_value(self) -> f32 {
```
chrisduerr, line 222:
```suggestion
    pub fn inner(self) -> f32 {
```

Alternatively: `as_f32`.

alacritty_terminal/src/event.rs
```diff
@@ -1,16 +1,11 @@
...
 use crate::term::SizeInfo;
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum Event {
-    DPRChanged(f64, (u32, u32)),
-    ConfigReload(PathBuf),
     MouseCursorDirty,
-    Message(Message),
-    Title(String),
+    Title(Option<String>),
```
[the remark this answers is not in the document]
chrisduerr: Still curious about this. A `ResetTitle` event would probably be easier for others to understand I think.

alacritty/src/cli.rs
kchibisov: It breaks formatting and makes it ugly.

PR #3421, opened 2020-03-05
Fix selection with invisible start and end

chrisduerr:


alacritty_terminal/src/term/mod.rs
```diff
@@ -222,9 +222,8 @@ impl<'a, C> RenderableCellsIter<'a, C> {
     ) -> RenderableCellsIter<'b, C> {
         let grid = &term.grid;
         let num_cols = grid.num_cols();
-        let num_lines = grid.num_lines();
 
-        let cursor_offset = grid.line_to_offset(term.cursor.point.line);
+        let cursor_offset = grid.num_lines().0 - term.cursor.point.line.0 - 1;
```
chrisduerr: I've inlined this method, since it was not used anywhere else anymore and even this one instance will disappear once the vimode PR is merged. So I don't think there's any need anymore to have it as separate method now.

We could keep it for now and just remove it in vimode, but I'd probably forget until this is merged and it should be fine the way it is.

kchibisov: Makes sense.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -145,24 +145,20 @@ impl<T: GridCell + PartialEq + Copy> Grid<T> {
...
-        if point.line < self.display_offset || point.line >= self.display_offset + self.lines.0 {
-            return None;
+    /// Clamp a buffer point to the visible region.
+    pub fn clamp_buffer_to_visible(&self, point: Point<usize>) -> Point {
+        if point.line < self.display_offset {
+            Point::new(self.lines - 1, self.cols - 1)
+        } else if point.line >= self.display_offset + self.lines.0 {
+            Point::new(Line(0), Column(0))
+        } else {
+            Point::new(self.lines + self.display_offset - point.line - 1, point.col)
```
kchibisov: you're basically doing here `visible_to_buffer`, so maybe we should just call it here?

chrisduerr: Probably because it's wrong. Need to fix that.

Issue #3404, opened 2020-03-01
Rework default bright colors


chrisduerr:
The default bright colors are often indistinguishable from the normal colors. This impacts the functionality of the default configuration file, which should achieve to provide sensible defaults that are usable out of the box for most people.

Since colors are highly subjective, I'd propose that the best approach to take is to imitate the current behavior of the dim colors, by taking the normal colors and changing their brightness by a fixed amount.

By default we use a factor of `0.66` for the dim colors, so a good place to start would probably be to use `1.34`. If that doesn't provide sufficient contrast, increasing that factor would be an option. Using completely different colors however wouldn't be ideal, since that is likely too subjective to get any consensus.

Aelnor:
Hello,

I'd like to take this one. 
If something goes wrong and there's no news from me after a week, feel free to reassign.

PR #3423, opened 2020-03-05
Change default colorscheme to Tomorrow Night

Aelnor:
This is implementation of #3404 

Attached example of the colors with the default config. The first line is black and bright black, which are almost invisible.
Screenshot from 2020-03-05 23-09-37

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Mouse bindings for additional buttons need to be specified as a number not a string
 - Don't hide cursor on modifier press with `mouse.hide_when_typing` enabled
 - `Shift + Backspace` now sends `^?` instead of `^H`
+- Default color scheme is now `Tomorrow Night` with the bright colors of `Tomorrow Night Bright`
 
 ### Fixed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -169,20 +169,20 @@
 # If `true`, bold text is drawn using the bright color variants.
 #draw_bold_text_with_bright_colors: false
 
-# Colors (Tomorrow Night Bright)
+# Colors (Tomorrow Night)
 #colors:
   # Default colors
   #primary:
-  #  background: '#000000'
-  #  foreground: '#eaeaea'
+  #  background: '#1d1f21'
+  #  foreground: '#c5c8c6'
 
     # Bright and dim foreground colors
     #
     # The dimmed foreground color is calculated automatically if it is not present.
     # If the bright foreground color is not set, or `draw_bold_text_with_bright_colors`
     # is `false`, the normal foreground color will be used.
-    #dim_foreground: '#9a9a9a'
-    #bright_foreground: '#ffffff'
+    #dim_foreground: '#828482'
+    #bright_foreground: '#eaeaea'
 
   # Cursor colors
   #
@@ -211,39 +211,39 @@
 
   # Normal colors
   #normal:
-  #  black:   '#000000'
-  #  red:     '#d54e53'
-  #  green:   '#b9ca4a'
-  #  yellow:  '#e6c547'
-  #  blue:    '#7aa6da'
-  #  magenta: '#c397d8'
-  #  cyan:    '#70c0ba'
-  #  white:   '#eaeaea'
+  #  black:   '#1d1f21'
+  #  red:     '#cc6666'
+  #  green:   '#b5bd68'
+  #  yellow:  '#f0c674'
+  #  blue:    '#81a2be'
+  #  magenta: '#b294bb'
+  #  cyan:    '#8abeb7'
+  #  white:   '#c5c8c6'
 
   # Bright colors
   #bright:
   #  black:   '#666666'
-  #  red:     '#ff3334'
-  #  green:   '#9ec400'
+  #  red:     '#d54e53'
+  #  green:   '#b9ca4a'
   #  yellow:  '#e7c547'
   #  blue:    '#7aa6da'
-  #  magenta: '#b77ee0'
-  #  cyan:    '#54ced6'
-  #  white:   '#ffffff'
+  #  magenta: '#c397d8'
+  #  cyan:    '#70c0b1'
+  #  white:   '#eaeaea'
 
   # Dim colors
   #
   # If the dim colors are not set, they will be calculated automatically based
   # on the `normal` colors.
   #dim:
-  #  black:   '#000000'
-  #  red:     '#8c3336'
-  #  green:   '#7a8530'
-  #  yellow:  '#97822e'
-  #  blue:    '#506d8f'
-  #  magenta: '#80638e'
-  #  cyan:    '#497e7a'
-  #  white:   '#9a9a9a'
+  #  black:   '#131415'
+  #  red:     '#864343'
+  #  green:   '#777c44'
+  #  yellow:  '#9e824c'
+  #  blue:    '#556a7d'
+  #  magenta: '#75617b'
+  #  cyan:    '#5b7d78'
+  #  white:   '#828482'
 
   # Indexed Colors
   #
--- a/alacritty_terminal/src/config/colors.rs
+++ b/alacritty_terminal/src/config/colors.rs
@@ -117,11 +117,11 @@ impl Default for PrimaryColors {
 }
 
 fn default_background() -> Rgb {
-    Rgb { r: 0, g: 0, b: 0 }
+    Rgb { r: 0x1d, g: 0x1f, b: 0x21 }
 }
 
 fn default_foreground() -> Rgb {
-    Rgb { r: 0xea, g: 0xea, b: 0xea }
+    Rgb { r: 0xc5, g: 0xc8, b: 0xc6 }
 }
 
 /// The 8-colors sections of config.
@@ -151,14 +151,14 @@ struct NormalColors(AnsiColors);
 impl Default for NormalColors {
     fn default() -> Self {
         NormalColors(AnsiColors {
-            black: Rgb { r: 0x00, g: 0x00, b: 0x00 },
-            red: Rgb { r: 0xd5, g: 0x4e, b: 0x53 },
-            green: Rgb { r: 0xb9, g: 0xca, b: 0x4a },
-            yellow: Rgb { r: 0xe6, g: 0xc5, b: 0x47 },
-            blue: Rgb { r: 0x7a, g: 0xa6, b: 0xda },
-            magenta: Rgb { r: 0xc3, g: 0x97, b: 0xd8 },
-            cyan: Rgb { r: 0x70, g: 0xc0, b: 0xba },
-            white: Rgb { r: 0xea, g: 0xea, b: 0xea },
+            black: Rgb { r: 0x1d, g: 0x1f, b: 0x21 },
+            red: Rgb { r: 0xcc, g: 0x66, b: 0x66 },
+            green: Rgb { r: 0xb5, g: 0xbd, b: 0x68 },
+            yellow: Rgb { r: 0xf0, g: 0xc6, b: 0x74 },
+            blue: Rgb { r: 0x81, g: 0xa2, b: 0xbe },
+            magenta: Rgb { r: 0xb2, g: 0x94, b: 0xbb },
+            cyan: Rgb { r: 0x8a, g: 0xbe, b: 0xb7 },
+            white: Rgb { r: 0xc5, g: 0xc8, b: 0xc6 },
         })
     }
 }
@@ -170,13 +170,13 @@ impl Default for BrightColors {
     fn default() -> Self {
         BrightColors(AnsiColors {
             black: Rgb { r: 0x66, g: 0x66, b: 0x66 },
-            red: Rgb { r: 0xff, g: 0x33, b: 0x34 },
-            green: Rgb { r: 0x9e, g: 0xc4, b: 0x00 },
+            red: Rgb { r: 0xd5, g: 0x4e, b: 0x53 },
+            green: Rgb { r: 0xb9, g: 0xca, b: 0x4a },
             yellow: Rgb { r: 0xe7, g: 0xc5, b: 0x47 },
             blue: Rgb { r: 0x7a, g: 0xa6, b: 0xda },
-            magenta: Rgb { r: 0xb7, g: 0x7e, b: 0xe0 },
-            cyan: Rgb { r: 0x54, g: 0xce, b: 0xd6 },
-            white: Rgb { r: 0xff, g: 0xff, b: 0xff },
+            magenta: Rgb { r: 0xc3, g: 0x97, b: 0xd8 },
+            cyan: Rgb { r: 0x70, g: 0xc0, b: 0xb1 },
+            white: Rgb { r: 0xea, g: 0xea, b: 0xea },
         })
     }
 }


alacritty_terminal/src/term/color.rs
```diff
@@ -204,7 +221,7 @@ impl List {
                 self[ansi::NamedColor::DimMagenta] = colors.normal().magenta * 0.66;
                 self[ansi::NamedColor::DimCyan] = colors.normal().cyan * 0.66;
                 self[ansi::NamedColor::DimWhite] = colors.normal().white * 0.66;
-            },
```
nixpulvis: Might as well define a `coeff` for dim colors here while you're at it.

nixpulvis:
Yea, I think something could _possibly_ be done to fix this by changing our default green to be a bit more green. Even the current green and yellow are a bit too close if you ask me.

Though the real solution to luminescence scaling would be a bit more complicated than just multiplying each RGB component by a fixed coefficient. I believe we're interested in the [relative luminance](https://en.wikipedia.org/wiki/Relative_luminance), since human color perception is a bit weird.

Aelnor:
> 
> 
> Yea, I think something could _possibly_ be done to fix this by changing our default green to be a bit more green. Even the current green and yellow are a bit too close if you ask me.
> 
> Though the real solution to luminescence scaling would be a bit more complicated than just multiplying each RGB component by a fixed coefficient. I believe we're interested in the [relative luminance](https://en.wikipedia.org/wiki/Relative_luminance), since human color perception is a bit weird. Perhaps we could use https://crates.io/crates/palette?

Thanks for the input, I'll play with these and let you know what the results are.

chrisduerr:
> Though the real solution to luminescence scaling would be a bit more complicated than just multiplying each RGB component by a fixed coefficient. I believe we're interested in the relative luminance, since human color perception is a bit weird. Perhaps we could use https://crates.io/crates/palette?

I don't think there's any issue with our 'simple' solution, since we have absolutely no need to accurately modify the luminance of the color. We just need to make it a bit brighter/dim, to a degree where the user can differentiate it. Pulling in a depedency for that which does who knows what when our 'simple' solution seems to work fine doesn't sound like a great idea to me.

nixpulvis:
@chrisduerr our simple solution is not working, as you yourselves noted with the bright green and bright yellow colors. Scaling colors is fundamentally a bit complex and the perfect use for a dependency (I wouldn't want Alacritty to maintain luminance code itself). I am generally on the same page about unnecessary dependencies, but unless we want to hard code the bright and dim colors ourselves (using a similar method as the linked crate), then I don't see a better solution.

If the issue is about having a runtime, I suppose this could be worked into a build time functionality, but I wouldn't think it's worth it.

chrisduerr:
> @chrisduerr our simple solution is not working, as you yourself noted with the bright green and bright yellow colors.

This has nothing to do with our luminance implementation though. Using a 'proper' luminance implementation still makes it difficult to distinguish.

I don't see how an external luminance would improve this and adding it as a dependency is without a doubt more trouble than it's worth. I at least am not looking forward to maintaining an entire color library, just because the author of it went MIA.

nixpulvis:
@chrisduerr is the Rust community can't maintain a color library then I'd be a bit concerned, and surprised. What maintenance are you imagining a color library needing?

That said, we could always rip it out if there are issues which need maintenance which we can't perform. This sounds a bit overly conservative to me honestly.

chrisduerr:
> That said, we could always rip it out if there are issues which need maintenance which we can't perform. This sounds a bit overly conservative to me honestly.

I'd rather not add an entire library just to brighten/dim colors a bit rather than adding it and then ripping it out again.

> @chrisduerr is the Rust community can't maintain a color library then I'd be a bit concerned, and surprised. What maintenance are you imagining a color library needing?

It's not maintained by `the Rust community`. It's basically maintained by a single person. It also has 25 depedencies itself which we need to make sure are always up to date. It also has roughly 15k LoC (Alacritty itself has only 22k). So that's far from 'overly conservative' if you ask me.

nixpulvis:
Alright, then perhaps we should just hardcore the colors.

chrisduerr:
> Alright, then perhaps we should just hardcore the colors.

Maybe we should. Dim colors are basically never used, but I think hand-picking the bright colors makes sense. Which is why this was originally the suggestion in the issue.

Though I've tested this code with my own colorscheme (base16) and that does a lot better with the brightness.

Aelnor:
> > Alright, then perhaps we should just hardcore the colors.
> 
> Maybe we should. Dim colors are basically never used, but I think hand-picking the bright colors makes sense. Which is why this was originally the suggestion in the issue.
> 
> Though I've tested this code with my own colorscheme (base16) and that does a lot better with the brightness.

So, what would be the final call? Just change default brights in the config, roll back the deriving feature?

kchibisov:
I feel like we can change the default colorscheme to be something more readable in general. It shouldn't look very nice, but just be the one with +- normal colors. It was originally stated that colorscheme change isn't really an option, but our default one seem strange tbh. I'm not sure if changing yellow and green could help it much.

nixpulvis:
I don't have any problems with our color scheme. Are there any tangible complaints?

The issue is just defining good bright colors that look "right". Currently we have a green that looks like a yellow in this PR. After the discussion I think hard coding them is the best bet. Which means I would also hard code the dim colors too, while we're at it.

chrisduerr:
> It was originally stated that colorscheme change isn't really an option, but our default one seem strange tbh.

It's not an option unless everyone is immediately happy with the new suggestion and accepts it as an actual improvement. This includes mainly @nixpulvis, @kchibisov and myself. But even if just we have to agree on something, I don't know if there's anything that could easily gain approval without much of a discussion. The only colorscheme where I wouldn't immediately complain would be the base16 colors, but even then the foreground and background colors are more difficult. So I don't think that discussion will be beneficial.

> After the discussion I think hard coding them is the best bet. Which means I would also hard code the dim colors too, while we're at it.

Why? I don't see any issues with the dim colors and the algorithm we currently use works very well even for my colorscheme which doesn't have bright colors. If I had to choose between arbitrary dim colors, or dim colors that are close to my own colorscheme, I'd choose the latter every time.

nixpulvis:
@chrisduerr I wasn't proposing changing the sim colors, just hard coding them to be consistent with the new bright colors. If we want to keep them the same that's fine too.

Either way, I mostly liked the way the colors turned out with the scaling in this PR, except the green. So maybe we can just manually tweak that one and save the colors?

chrisduerr:
> @chrisduerr I wasn't proposing changing the dim colors, just hard coding them to be consistent with the new bright colors. If we want to keep them the same that's fine too.

I mean currently the dim colors are consistent with the normal colors, which seems better than being consistent with the bright ones?

nixpulvis:
Sorry if I'm not being clear. I mean leaving the color of the dims unchanged, but just making them hard coded like the proposed bright colors, for consistency. Again, this is a small issue.

The larger issue is finding a bright green that works.

chrisduerr:
> Sorry if I'm not being clear. I mean leaving the color of the dims unchanged, but just making them hard coded like the proposed bright colors, for consistency. Again, this is a small point.

It's not a small point though. My configuration relies on them not being hardcoded. I don't want Alacritty's shitty default colors to mess with my dim colors. I just want a dimmer version of my normal colors. That works great already and I do **not** want to regress this.

nixpulvis:
Ohhh I see. Sorry, we'll then now I really don't see a way to make bright colors work well without proper scaling... users will have the same issue with brights.

Maybe we could just implement the scaling from the Wiki to good effect without the massive library I linked?

chrisduerr:
> Ohhh I see, sorry. Well then now I really don't see a way to make bright colors work well without proper scaling... users will have the same issue with brights.

Sounds like one of those brights should be dim?

> Maybe we could just implement the scaling from the Wiki to good effect without the massive library I linked?

I still don't see an issue with our current scaling for dim colors. It works flawlessly for the colorschemes I've tested. It might not be 100% accurate to what the human eye might perceive, but it's not like that matters at all.

kchibisov:
>I don't have any problems with our color scheme. Are there any tangible complaints?

I've heard them a few times on IRC and on our [tracker](https://github.com/alacritty/alacritty/commit/d4c1d51e36626b1682b96e746bb32632dadcac2c#r34730200). At least yellow and green should be changed, but I find it way to dark tbh. I don't have any suggestion, since colors are hard and knowing how shitty some monitors are, it's even harder to pick dark background color , like the one I like `#1c1c1c` looks absolutely strange on my old 1080p monitor.

nixpulvis:
OK, I've taken some time to play around with this. I don't think my earlier suggestion about color scaling really applies after all, since the main issue is with our green being much to yellow (something that any kind of scaling wont fix). For what it's worth, **the way we scale colors today is fine.**

-----

Here's a proposal for a new green color:

`#7ada71`

And how the bright, normal, and dim colors would all look together:

DeepinScreenshot_select-area_20200307145739

Versus how they look today:

DeepinScreenshot_select-area_20200307145801

-----

This also highlights a potential issue with bright black, we should consider.

Aelnor:
If you ask me, the first green (`#7ada71`) looks better than the second (`#99ad6a`).

nixpulvis:
@Aelnor we've been arguing about this in IRC a bit. Sadly we're not really in agreement yet about the best course of action, however @chrisduerr strongly opposes changing just a single color of the theme, as I've suggested above. Sorry for the thrashing around on this.

The best suggestion I've heard so far is to use HSL (not from a crate) to scale things, and hope that it fixes the existing colors. Otherwise we may need to find a new, popular, color scheme (also using HSL). Base 16 schemes seem to be acceptable.

nixpulvis:
Testing hard coding the normal and bright to [Tomorrow Night](https://github.com/chriskempson/tomorrow-theme) normal and bright respectively (just always using #000000 for the black) gives pretty good results:

DeepinScreenshot_select-area_20200307170908

nixpulvis:
Testing out HSL with CSS yields promising results at 1.25 scaled "lightness":

DeepinScreenshot_select-area_20200307173335

Aelnor:
@nixpulvis please correct me if I'm wrong.
You suggest to convert RGB to HSL, increase L and then convert it back to RGB?

nixpulvis:
@Aelnor that would be my personal favorite implementation (barring any unforeseen issues/complaints). However, I'd also personally be OK with not deriving the colors in code, and just updating `BrightColors::default` to have HSL scaled values (though @chrisduerr seems not to like this as much).

Aelnor:
I've played with this approach a bit and I'm not really sure I'm happy with the results, but still I'd like to share. 
The current version has 1.25 increased lightness.
Screenshot from 2020-03-11 20-11-22

I'm updated this PR, but the code is dirty and is not ready to be merged, I'm only updating it in case someone wants to play with it. I'm going to clean everything up if approach is promising.

Please let me know what you think.

nixpulvis:
@chrisduerr agreed, hopefully we can do something about this with the S value. I've tried hacking it a bit by changing the line in the current `enlighten` function to the following:

```rust
let result = Hsl { h: self.h, s: (self.s * rhs * 2.).max(0.0).min(1.0), l: (self.l * rhs).max(0.0).min(1.0) };
```
which shows promise, although something is clearly off about the black, since it's starting to turn green:

DeepinScreenshot_select-area_20200312001342

Aelnor:
@nixpulvis cleaned up the code, please take a look.
If we are ok with the current numbers, I'll update default config file.

nixpulvis:
@Aelnor I've been going through testing the color schemes we list in the Wiki, and I've noticed a few things.

1. For some of them, unfortunately the bright colors just aren't brighter:
**Blood Moon**
DeepinScreenshot_select-area_20200315181218

2. Worse, for some schemes the colors get thrown pretty far off:
**Darktooth**
DeepinScreenshot_select-area_20200315181554

I **really** wish I had remembered about this project before suggesting HSL, but there's a better color space for doing these kinds of things. It's called HCL, since "Luminance" is in fact what we care about. To see what I mean, perhaps https://jackdoerner.net/artifacts/perception can give some confidence before following the implementation of a different color primitive. I recommend taking some of the hex color values from the problematic color schemes I mentioned above and playing with them in this tool. The code for this example can be found [here](https://bitbucket.org/jackdoerner/perception/src/master/system/js/PerceptualColor.js) otherwise, there's more information about HCL elsewhere too.

If you do decide to continue working on this direction, then maybe I can help with some of the Rust as well... I'm just not sure we can accept this implementation yet as is. Better to take our time and get it right in this regard.

Conversions which cannot fail should be implemented with [`From`](https://doc.rust-lang.org/std/convert/trait.From.html), and conversions which can fail should be done with [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html). I expect you'll have to handle the case where there isn't a mapping like this. Each of the linked trait docs give good examples of how to do this.

Please don't hesitate to ask questions, I'll try my best to help out.

chrisduerr:
@nixpulvis Have you tried converting a few colorschemes using HCL? I'd personally be very interested in seeing a few screenshots of actual Alacritty text with maybe a few different HCL scaling variations. I personally find your direct screenshot comparisons much better to visualize the actual differences than trying it in some separate tool directly.

So maybe the best idea before trying to go anywhere near Rust would be to at least take a couple of themes that look like decent examples and scale them with a few variations that could potentially lead to good results?

nixpulvis:
OK, I've reworked the above two problematic color schemes using HCL to pretty good effect. I think yellow (and make L values) are still a little tricky, but here's the resulting colors:

1. **Blood Moon**:
DeepinScreenshot_select-area_20200318115252
<details>
<summary>YAML</summary>

```yaml
  normal:
    black:   '#10100E'
    red:     '#C40233'
    green:   '#009F6B'
    yellow:  '#FFD700'
    blue:    '#0087BD'
    magenta: '#9A4EAE'
    cyan:    '#20B2AA'
    white:   '#C6C6C4'
  # NOTE: Multiple L by 1.25, max value is 90 out of 100.
  bright:
    black:   '#121211'
    red:     '#e4354a'
    green:   '#46c790'
    yellow:  '#fee083'  # NOTE: Not a direct match.
    blue:    '#48a9e2'
    magenta: '#ba6bcd'
    cyan:    '#5fe0d7'
    white:   '#e2e2e0'
  # NOTE: Multiple L by 0.66.
  dim:
    black:   '#0b0b08'
    red:     '#85011f' # NOTE
    green:   '#016744' # NOTE
    yellow:  '#a28700' # NOTE
    blue:    '#00587d' # NOTE
    magenta: '#6f2584'
    cyan:    '#04736d' # NOTE
    white:   '#7d7d7c'
```
</details>

2. **Darktooth**:
DeepinScreenshot_select-area_20200318120854
<details>
<summary>YAML</summary>

```yaml
  normal:
    black:   '#282828'
    red:     '#9d0006'
    green:   '#79740e'
    yellow:  '#b57614'
    blue:    '#076678'
    magenta: '#8f3f71'
    cyan:    '#00a7af'
    white:   '#fdf4c1'
  # NOTE: Multiple L by 1.25, max value is 90 out of 100.
  bright:
    black:   '#303030'
    red:     '#b62518'
    green:   '#999231'
    yellow:  '#de993b'
    blue:    '#307e91'
    magenta: '#a95789'
    cyan:    '#51d2d9'
    white:   '#fffffe'
  # NOTE: Multiple L by 0.66.
  dim:
    black:   '#1c1c1c'
    red:     '#6b0600' # NOTE
    green:   '#4f4c00' # NOTE
    yellow:  '#784b01' # NOTE
    blue:    '#014451' # NOTE
    magenta: '#6b1c50'
    cyan:    '#006b71' # NOTE
    white:   '#a19a6b'
```
</details>

I'd say this is pretty good evidence for this color space working out well. However, as `NOTE`ed in the YAML, not all colors map directly into the space, and approximates are needed to be found as well, further complicating the algorithm. Of course, the tool I've linked also performs this task, but it's just something to keep in mind. There's no doubt in my mind this belongs in a separate crate at this point, which assuming @chrisduerr is OK with it, could live under the alacritty org. I'm quite happy to assume responsibility of this repository, with your help @Aelnor (if you're interested).

`hcl` on crates.io is currently being squatted on by @swmon (as are a **bunch** of other popular names), so we could probably get this if we desire.

Thoughts? This is definitely more complex than I think we originally planned, but could work out well, both for us, and the larger Rust community.

Aelnor:
@nixpulvis I love your results.
I'm interested in developing the separate crate especially if this can help Alacritty to manage colors.

chrisduerr:
@nixpulvis Could you post the default config colorscheme too?

I'm honestly not convinced by this approach myself. I think it has the same issues as the HSL approach and does not provide colors which I could reasonably recommend anyone to use.

Due to the loss in saturation, I would not be able to tell which of these colors was the bright and the normal variations if you just put them next to each other. Taking the red and yellow of the blood moon theme for example, I think this is pretty clear.

The changes it makes to the white of the darktooth theme also seem like it's going into a direction that does not fit in with the rest of the colorscheme.

So since I could not recommend to anyone to use the automatically derived bright colors using this HCL method, I do not think it makes sense to merge this approach into Alacritty. If we cannot provide something that works reliably throughout most of the colorspace and schemes available for Alacritty, we shouldn't provide anything.

nixpulvis:
@chrisduerr these examples, if you plug into the linked tool, are truly at the limits of the color space. There's simply nothing better I can imagine doing with them. I mean, perhaps we could train an ML algorithm to each user's eyes individually on everyone's first launch. But I'm obviously joking about that...

This seems pretty good to me, though perhaps the values `0.66` and `1.25` aren't quite right. At least this color space (and approximations) itself seems close to correct. I think many people will end up using thier own bright colors if they have issues with their specific scheme.

I'll try plugging in the default scheme too see how that looks. And I'll try a few more just to get a sense of how many schemes really lie on the edge of what's usable...

chrisduerr:
> @chrisduerr these examples, if you plug into the linked tool, are truly at the limits of the color space. There's simply nothing better I can imagine doing with them.

You say that, but then a color gets dimmer, rather than brighter. If you take that red for example I can guarantee you that `#FF0000` would look brighter than the existing red. So it's certainly possible.

> At least this color space (and approximations) itself seems close to correct.

I disagree. It seems like something that theoretically seems nice to put your mind at ease, but does not work at all in practice.

> I think many people will end up using thier own bright colors if they have issues with their specific scheme.

So why bother?

Aelnor:
Well,
I've started this whole thing by misreading the original proposal. And it seems we're unable to find a solution that everyone is happy about. 
Thus, I'd suggest to fall back to the original idea and only fix hardcoded default bright colors. I'm not immediately happy with different approaches to dimmed and bright colors (dimmed are being calculated while brights are not) and I really like results demonstrated by HCL approach, but it seems this approach does not work on every popular themes.
Thus, let's keep it conservative. Unless you, @chrisduerr, @nixpulvis object, I'll revert everything and will try to find decent default bright colors.

nixpulvis:
@Aelnor not a bad idea at all. Deriving theme appropriate color schemes like this is insanely tricky to get right, and most users will probably be happier copying both `dim` and `bright` colors for themselves anyway. So I agree, we should give up trying to scale `bright` colors.

My next question is then, what should we do if a user only configures the `normal` colors? I would think the best course of action, without the ability to scale them ourselves, would be to simply use the same `normal` colors as their `bright` colors. They may be a bit confused why `dim` colors are derived but `bright` colors aren't, so perhaps a comment in the config is in order. Thoughts?

---

There's still a small question in my head if it's worth using the HSL or HCL color spaces for scaling the `dim` colors, since I believe they also looked quite a bit better. I'm guessing we don't think it's worth it, but I thought I'd just leave that question here.

chrisduerr:
> My next question is then, what should we do if a user only configures the normal colors? I would think the best course of action, without the ability to scale them ourselves, would be to simply use the same normal colors as their bright colors.

We should use the default unless specified otherwise, not fallback to the normal colors.

> They may be a bit confused why dim colors are derived but bright colors aren't, so perhaps a comment in the config is in order. Thoughts?

I think such a comment exists already?

> There's still a small question in my head if it's worth using the HSL or HCL color spaces for scaling the dim colors, since I believe they also looked quite a bit better. I'm guessing we don't think it's worth it, but I thought I'd just leave that question here.

I personally think that the dimming of colors always worked quite well, so I'm not sure if there's much benefit in introducing more complexity here?

nixpulvis:
> We should use the default unless specified otherwise, not fallback to the normal colors.

I guess we have to do this, since bright colors should always be distinguishable from normal colors by default.

> I think such a comment exists already?

I mean, adding to the `bright` config entry something like "unlike dim...",

> I personally think that the dimming of colors always worked quite well, so I'm not sure if there's much benefit in introducing more complexity here?

Yea, that fine. Simpler is better, and I agree the current dim colors aren't awful or anything.

chrisduerr:
> I mean, adding to the bright config entry something like "unlike dim...".

I'm not so sure if we need to document both? Usually I'd say that we only have to document the 'odd one out' and the other one should behave 'normal'.

Aelnor:
Screenshot from 2020-03-26 23-07-13

Please let me know what you think.
I'm usually not really good with the color taste. :)

nixpulvis:
@Aelnor I feel like the colors are kinda washed out. I think this is partially a problem caused (as out investigation into HCL discovered) by the underlying theme.

Using the script:

```sh
#!/bin/sh
echo -e "\e[90mfoo\e[91mbar\e[92mbaz\e[93mbat\e[94mbax\e[95mbal\e[96mbag\e[97mban"
echo -e "\e[30mfoo\e[31mbar\e[32mbaz\e[33mbat\e[34mbax\e[35mbal\e[36mbag\e[37mban"
echo -e "\e[2m\e[30mfoo\e[31mbar\e[32mbaz\e[33mbat\e[34mbax\e[35mbal\e[36mbag\e[37mban"
```

I've looked at a large sample of the themes in our Wiki, and TBH most of them wouldn't pass the test for the default these bright colors IMO. The most notable exceptions were the [Kitty](https://github.com/alacritty/alacritty/wiki/Color-schemes#kitty) and [Terminal.app](https://github.com/alacritty/alacritty/wiki/Color-schemes#terminalapp) themes, which I feel both could be used as a replacement for our defaults. However, this is a larger discussion, I'd be curious what @chrisduerr has to say about. I'm personally not attached to Tomorrow Night Bright as the default at all, though I do like how it's **very** dark, something both linked themes shared in common.

Aelnor:
@nixpulvis I'm not sure I'm following you.
Are we looking for default bright colors that suit every theme out there or for just our default theme?

Aelnor:
@nixpulvis @chrisduerr 
Folks? I'm unsure what my next steps should be. Are we looking for bright colors that suit every theme?

chrisduerr:
> Are we looking for bright colors that suit every theme?

We definitely are not. If people change their theme and the new colors do not work with the bright colors, they can just change the bright colors.

What are the current values of your PR based on?

Aelnor:
Current PR is based on colors converted to HSL and then have their S and L increased

Aelnor:
Folks, I need a guidance on what to do next.
Are we okay with these colors? If not, what are the issues you see? @nixpulvis mentioned these colors do not look well with a lot of themes. This comment confuses me and I'm unsure if I understand it.
If we are okay with the colors, I'll clean up the PR, rebase and update default config file as well.

chrisduerr:
I personally still do not like the HCL/HSL approach at all. I don't think it provides good results and I don't think the colors picked here are enjoyable defaults for Alacritty's users.

My opinion remains largely unchanged that picking some reasonable defaults like Tommorow Night for normals and Tomorrow Night Bright for brights is going to give a better result than trying to use math to derive these colors.

But a large driving force behind this PR has been @nixpulvis, so I wouldn't want to suggest any changes without hearing what his thoughts are.

nixpulvis:
Sorry for the long delay, life and stuff...

I agree the HCL/HSL stuff didn't work out as nicely as I'd hoped. And while the colors proposed (but not implemented) in this PR are a clear improvement, I would hesitate to call them perfect. Regardless, I think hard coding the colors is a fine choice for now. The colors @Aelnor has chosen are clearly better than the Tomorrow Night Bright colors, as you can see by comparing these two images:

Brights = Tomorrow Night Bright Brights:
DeepinScreenshot_select-area_20200511192400

Brights = Custom (as documented in this PR):
DeepinScreenshot_select-area_20200511192446

Perhaps at a later date we'll revisit the whole default color scheme and make things even better. This would almost certainly involve making the colors more distinct, since that way we'd get more range for dims/brights. I would unfortunately expect this color scheme to be generally disliked in terms of aesthetics.

As for the small issue of hard-coded brights, but scaled dims, I think this is potentially confusing to users. I understand why dims are a better fit for scaling, so perhaps all that's needed is a line in the default config file stating that dim colors are scaled automatically when omitted?

chrisduerr:
> Ugly but highly functional defaults should be preferred to pretty but unusable ones, since color preference is highly subjective anyway.

I strongly disagree with this sentiment. Because ugly and unusable go hand in hand. It's completely useless if the colorscheme has the perfect mathematical contrast possible if I cannot look at it without my eyes falling out.

> As for the small issue of hard-coded brights, but scaled dims, I think this is potentially confusing to users. I understand why dims are a better fit for scaling, so perhaps all that's needed is a line in the default config file stating that dim colors are scaled automatically when omitted?

This is already the case.

> The colors @Aelnor has chosen are clearly better than the Tomorrow Night Bright colors

I'm not sure if I understand what you're documenting here. Are the images the bright/normal/dim colors (in that order)? And is it using our current defaults for normal and dim? If so, then it looks to me like the tomorow night bright colors are more distinctive, not less? Unless you mixed them up.

And if we'd use tomorrow night bright as bright colors, the normal colors would change too, right? So you'd have to adjust those to tomorrow night? Or are the normal and dim colors in both already using tomorrow night? Both seem to be using the same normal colors, so clearly one of them is probably wrong?

nixpulvis:
I hereby rescind myself from this issue. I'm getting nowhere attempting to explain myself, and I'm only wasting my own time, and yours attempting to help.

If and when I am able, I'll be back on future issues.

Aelnor:
Thanks for the answers!

Did I get it right -- I should change default normal colors including background/foreground to Tomorrow Night as well?

chrisduerr:
> Did I get it right -- I should change default normal colors including background/foreground to Tomorrow Night as well?

Correct. Basically just use the entire tomorrow night theme. `Tomorrow Night` for normal colors, `Tomorrow Night Bright` for brights. Sorry for the massive delay, hopefully this should just get it over with.

Aelnor:
Here's what I have now
Screenshot from 2020-05-16 21-53-27

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Mouse bindings for additional buttons need to be specified as a number not a string
 - Don't hide cursor on modifier press with `mouse.hide_when_typing` enabled
 - `Shift + Backspace` now sends `^?` instead of `^H`
+- Default color scheme is now `Tomorrow Night`
```
chrisduerr, line 32:
```suggestion
- Default color scheme is now `Tomorrow Night` with the bright colors of `Tomorrow Night Bright`
```

alacritty.yml
```diff
@@ -169,19 +169,19 @@
...
+  #  background: '#1d1f21'
+  #  foreground: '#c5c8c6'
 
     # Bright and dim foreground colors
     #
     # The dimmed foreground color is calculated automatically if it is not present.
     # If the bright foreground color is not set, or `draw_bold_text_with_bright_colors`
     # is `false`, the normal foreground color will be used.
-    #dim_foreground: '#9a9a9a'
+    #dim_foreground: '#828482'
```
chrisduerr, line 184: Where is this from?

alacritty.yml
```diff
@@ -169,19 +169,19 @@
...
     #bright_foreground: '#ffffff'
```
chrisduerr, line 185: I think this should be `#eaeaea`, right?

alacritty.yml
```diff
@@ -211,39 +211,39 @@
...
+  #  green:   '#b5bd68'
+  #  yellow:  '#f0c674'
+  #  blue:    '#81a2be'
+  #  magenta: '#b294bb'
+  #  cyan:    '#8abeb7'
+  #  white:   '#c5c8c6'
 
   # Bright colors
   #bright:
   #  black:   '#666666'
```
chrisduerr, line 225: What is this based on? I'd assume it's not using the Tomorrow Night Bright black since that's darker than the normal's black, but I'd be curious why you've kept it at #666666. Just because you didn't know what better value to change it on, or is this actually a good choice based on some criteria?

alacritty.yml
```diff
@@ -211,39 +211,39 @@
...
-  #  cyan:    '#497e7a'
-  #  white:   '#9a9a9a'
+  #  black:   '#232425'
+  #  red:     '#864343'
+  #  green:   '#777c44'
+  #  yellow:  '#9e824c'
+  #  blue:    '#556a7d'
+  #  magenta: '#75617b'
+  #  cyan:    '#5b7d78'
+  #  white:   '#828482'
```
chrisduerr, line 246: I'd assume these are just the automatically computed colors?

alacritty.yml
Aelnor: default foreground * 0.66

alacritty.yml
Aelnor: You're right, fixed.

alacritty.yml
Aelnor: I wasn't sure what is color is better here and I think it looks decent.

alacritty.yml
Aelnor: Yes. These are normals * 0.66

chrisduerr: That doesn't seem to quite check out though. #1d1f21 * 0.66 should be less than those values. However #232425 is brighter in every single bit? So looks like something went wrong? Should be more around #131818 based on some rough estimation, though we should obviously pick the exact colors Alacritty computes by logging them and just copying them to the config file.

chrisduerr: That doesn't seem to quite check out though. #1d1f21 * 0.66 should be less than those values. However #232425 is brighter in every single bit? So looks like something went wrong? Should be more around #131818 based on some rough estimation, though we should obviously pick the exact colors Alacritty computes by logging them and just copying them to the config file.

alacritty.yml
chrisduerr: It's less than all the foreground colors and more than all the other blacks, so should be good!

alacritty.yml
chrisduerr: Checks out, seems good.

alacritty.yml
Aelnor: yep, thanks for spotting this out. One digit mistake copied three times. :(

Aelnor: Double checked other dims, they seem to be correct

Aelnor:
By the way, in the wiki Tomorrow NIght theme is not correct in my opinion (https://github.com/alacritty/alacritty/wiki/Color-schemes). It has too dark white (almost blackish). Could I fix it as well?

Arnavion:
To future readers, if you're here to figure out how to get the original colors back:

```yaml
# Revert https://github.com/alacritty/alacritty/commit/9a0eac0a14e631485c9e33e7499d84549ba510fc
colors:
  primary:
    background: '#000000'
    foreground: '#eaeaea'
    dim_foreground: '#9a9a9a'
    bright_foreground: '#ffffff'
  normal:
    black:   '#000000'
    red:     '#d54e53'
    green:   '#b9ca4a'
    yellow:  '#e6c547'
    blue:    '#7aa6da'
    magenta: '#c397d8'
    cyan:    '#70c0ba'
    white:   '#eaeaea'
```

PR #3425, opened 2020-03-06
Capitalized the .desktop Filename

nixpulvis:
As spesified by [Freedesktop's File Naming](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html), and as suggested in https://github.com/alacritty/alacritty/issues/1875#issuecomment-592994397, I've capitalized the name of our `.desktop` file, and updated the package files accordingly.

> ... the name of the application, which is conventionally written with words run together and initial capital letters (CamelCase). For example, if the owner of example.org writes "Foo Viewer", they might choose the name org.example.FooViewer, resulting in a file named org.example.FooViewer.desktop.

It seems like the file used to be named like this, and was changed at some point. However, I couldn't find the PR/commit that did it easily. If this was changed away from convention for a reason, I'd be curious what it was.


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `draw_bold_text_with_bright_colors` is now `false`
 - Mirror OSC query terminators instead of always using BEL
 - Increased Beam, Underline, and Hollow Block cursors' line widths
+- Capitalized the default Alacritty.desktop file
```
kchibisov: It should be in packaging section.

nixpulvis: Ah, nice. Forgot that section existed.

kchibisov:
name affects `class/app_id`, so if you want to use `com.alacritty.alacritty.desktop`, you class should also be `com.alacritty.alacritty`. I'm in favor of just using `Alacritty.desktop`, since most apps don't use reversed dns notation, just look at `/usr/share/applications` and it allows us to keep class as `Alacritty`.

nixpulvis:
FWIW, wouldn't we only use `com.alacritty.*` if we actually owned alacritty.com?

kchibisov:
I mean we can, but you should also change class to `com.alacritty.*` which is a breaking change and we can't do that.

nixpulvis:
We **can**, but we shouldn't... Saying we own a domain we don't own seems very wrong to me.

kchibisov:
yeah, and you also break tons of configurations without a good reason, since no-one is using reverse dns notation for desktop files at least on my system(well zathura is doing it, but everything else is similar to what we do), but we already have a lot of users matching `Alacritty` class.

chrisduerr:
> I'm in favor of just using Alacritty.desktop, since most apps don't use reversed dns notation, just look at /usr/share/applications and it allows us to keep class as Alacritty.

Yeah and I also see only .desktop files that are not capitalized. Not a single one.

> We can, but we shouldn't... Saying we own a domain we don't own seems very wrong to me.

I'm not certain if this is still the case, but jwilm owned the alacritty.io domain in the past. We might still have access to it.

nixpulvis:
@myfreeweb we've been discussing this a bit, and looking at files in `/usr/share/applications` it seems like many applications/packagers choose to go with dashized lowercase names. Reading things like [this Gnome page](https://developer.gnome.org/integration-guide/stable/desktop-files.html.en) also gives advice inconsistent with the freedesktop spec I linked.

Could you help us understand what use changing this has for you? Do you have issues with other application's desktop entries, like `firefox.desktop` for example?

myfreeweb:
> Could you help us understand what use changing this has for you?

I'm writing an app launcher and trying to find launched windows' icons by doing `new DesktopAppInfo(id ~ ".desktop")` where `id` is the `app_id` from xdg-shell :) I'll need to do a smarter lookup eventually..

> many applications/packagers choose to go with dashized lowercase names

That's fine! I don't really care about the domain thing, I don't care about that part of the freedesktop naming spec, I only care about [the xdg-shell spec](https://github.com/wayland-project/wayland-protocols/blob/b0a25f26d3e8b5ad24773881a9a895dd9cc75466/stable/xdg-shell/xdg-shell.xml#L642-L646):

> The compositor shell will try to group application surfaces together
> by their app ID. As a best practice, it is suggested to select app
> ID's that match the basename of the application's .desktop file.
> For example, "org.freedesktop.FooViewer" where the .desktop file is
> "org.freedesktop.FooViewer.desktop".

Instead of capitalizing the desktop file, you can lowercase the `app_id` (and window class on X11). They just should be the same.

> I also see only .desktop files that are not capitalized. Not a single one.

Odd, I see quite a lot of the reverse-domain style:

<details>
<summary>
% ls /usr/local/share/applications | rg '^(com|io|org)'
</summary>

```
com.gexperts.Tilix.desktop
com.github.alecaddd.taxi.desktop
com.github.babluboy.bookworm.desktop
com.github.bharatkalluri.gifup.desktop
com.github.danrabbit.lookbook.desktop
com.github.phase1geo.minder.desktop
com.vinszent.GnomeTwitch.desktop
io.elementary.granite.demo.desktop
io.github.celluloid_player.Celluloid.desktop
org.daa.NeovimGtk.desktop
org.gnome.Epiphany.desktop
org.gnome.Evince-previewer.desktop
org.gnome.Evince.desktop
org.gnome.FileRoller.desktop
org.gnome.Fractal.desktop
org.gnome.Geary.desktop
org.gnome.gitg.desktop
org.gnome.Lollypop.desktop
org.gnome.meld.desktop
org.gnome.Nautilus.desktop
org.gnome.Podcasts.desktop
org.gnome.Terminal.desktop
org.gnome.Usage.desktop
org.gtk.Demo4.desktop
org.gtk.IconBrowser4.desktop
org.gtk.WidgetFactory4.desktop
org.inkscape.Inkscape.desktop
org.kde.kcachegrind.desktop
org.kde.krita.desktop
org.musicbrainz.Picard.desktop
org.opensc.notify.desktop
org.pitivi.Pitivi.desktop
org.radare.Cutter.desktop
org.remmina.Remmina.desktop
```

</details>

nixpulvis:
@myfreeweb thanks for the extra info.

So we must make a choice (hopefully once and for all):

1. Capitalize the `.desktop` file, as this PR is currently doing. **Note** that this is at least partially compatible with the freedesktop spec.
2. Lowercase the default class, which is for some reason the more popular format (at least of applications I've seen), as currently set here: https://github.com/alacritty/alacritty/blob/cc2fc0b1c36bbec7eb56d931e435e70a6e6e606c/alacritty_terminal/src/config/window.rs#L10-L11

I'm perfectly happy with either, but it should be noted here that changing the default class will break a number of peoples external configs. The fix should be easy, but it's a concern.

kchibisov:
I'm in favor of capitalization, since it's simple and not breaking.

chrisduerr:
I'm slightly in favor of lowercase. I feel like having things standardized is nice and the majority of things/files on Linux at least seem to be lowercase.

So I'll leave it up to @nixpulvis to decide.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Packaging
 
 - Minimum Rust version has been bumped to 1.37.0
+- Capitalized the default Alacritty.desktop file
```
```suggestion
- Capitalized the Alacritty.desktop file
```

nixpulvis:
We should merge this then. It'll be least likely to break user's configurations, and we're following a standard of some kind at least. Plus, regardless of standards, I think we're mostly in agreement that Alacritty is better than alacritty, just stylistically.

PR #3427, opened 2020-03-06
Fix live config reload for window title

chrisduerr:
This enables live config reload for the window title. This includes
updating the title after it has been pushed and popped from the title
stack.

The dynamic title option also isn't disabled automatically anymore when
the title is set in the config. If the title is set from CLI, the
behavior is unchanged and dynamic title changes are still disabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Mirror OSC query terminators instead of always using BEL
 - Increased Beam, Underline, and Hollow Block cursors' line widths
 - Dynamic title is not disabled anymore when `window.title` is set in config
+- Live config reload for `window.title`
 
 ### Fixed
 
@@ -74,7 +75,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Decorations visible when in fullscreen on Wayland
 - Window size not persisted correctly after fullscreening on macOS
 - Crash on startup with some locales on X11
-- Live config reload for `window.title`
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `draw_bold_text_with_bright_colors` is now `false`
 - Mirror OSC query terminators instead of always using BEL
 - Increased Beam, Underline, and Hollow Block cursors' line widths
+- Dynamic title is not disabled anymore when `window.title` is set in config
 
 ### Fixed
 
@@ -73,6 +74,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Decorations visible when in fullscreen on Wayland
 - Window size not persisted correctly after fullscreening on macOS
 - Crash on startup with some locales on X11
+- Live config reload for `window.title`
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Capitalized the default Alacritty.desktop file
 - Capitalized the Alacritty.desktop file
 
+### Added
+
+- Live config reload for `window.title`
+
 ### Changed
 
 - Pressing additional modifiers for mouse bindings will no longer trigger them
@@ -23,7 +27,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Mirror OSC query terminators instead of always using BEL
 - Increased Beam, Underline, and Hollow Block cursors' line widths
 - Dynamic title is not disabled anymore when `window.title` is set in config
-- Live config reload for `window.title`
 
 ### Fixed
 
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -964,8 +964,7 @@ impl<T> Term<T> {
         self.dynamic_title = config.dynamic_title();
 
         if self.dynamic_title {
-            let title = self.title.as_ref().unwrap_or(&self.default_title).clone();
-            self.set_title(title);
+            self.set_title(self.title.clone());
         } else {
             self.event_proxy.send_event(Event::Title(self.default_title.clone()));
         }
@@ -2090,14 +2089,13 @@ impl<T: EventListener> Handler for Term<T> {
     }
 
     #[inline]
-    fn set_title(&mut self, title: impl Into<String>) {
-        let title = title.into();
+    fn set_title(&mut self, title: Option<String>) {
+        trace!("Setting title to '{:?}'", title);
 
-        trace!("Setting title to '{}'", title);
-
-        self.title = Some(title.clone());
+        self.title = title.clone();
 
         if self.dynamic_title {
+            let title = title.unwrap_or_else(|| self.default_title.clone());
             self.event_proxy.send_event(Event::Title(title));
         }
     }
@@ -2123,9 +2121,7 @@ impl<T: EventListener> Handler for Term<T> {
 
         if let Some(popped) = self.title_stack.pop() {
             trace!("Title '{:?}' popped from stack", popped);
-
-            self.set_title(popped.as_ref().unwrap_or(&self.default_title).clone());
-            self.title = popped;
+            self.set_title(popped);
         }
     }
 }
@@ -2376,12 +2372,12 @@ mod tests {
         assert_eq!(term.title, None);
 
         // Title can be set
-        term.set_title("Test");
+        term.set_title(Some("Test".into()));
         assert_eq!(term.title, Some("Test".into()));
 
         // Title can be pushed onto stack
         term.push_title();
-        term.set_title("Next");
+        term.set_title(Some("Next".into()));
         assert_eq!(term.title, Some("Next".into()));
         assert_eq!(term.title_stack.get(0).unwrap(), &Some("Test".into()));
 
@@ -2405,8 +2401,13 @@ mod tests {
         // Title stack pops back to default
         term.title = None;
         term.push_title();
-        term.set_title("Test");
+        term.set_title(Some("Test".into()));
         term.pop_title();
+        assert_eq!(term.title, None);
+
+        // Title can be reset to default
+        term.title = Some("Test".into());
+        term.set_title(None);
         assert_eq!(term.title, None);
     }
 }
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -26,7 +26,7 @@ use crate::ansi::{
     self, Attr, CharsetIndex, Color, CursorStyle, Handler, NamedColor, StandardCharset, TermInfo,
 };
 use crate::clipboard::{Clipboard, ClipboardType};
-use crate::config::{Config, VisualBellAnimation, DEFAULT_NAME};
+use crate::config::{Config, VisualBellAnimation};
 use crate::event::{Event, EventListener};
 use crate::grid::{
     BidirectionalIterator, DisplayIter, Grid, GridCell, IndexRegion, Indexed, Scroll,
@@ -797,21 +797,24 @@ pub struct Term<T> {
     /// Default style for resetting the cursor
     default_cursor_style: CursorStyle,
 
-    /// Whether to permit updating the terminal title
-    dynamic_title: bool,
-
     /// Clipboard access coupled to the active window
     clipboard: Clipboard,
 
     /// Proxy for sending events to the event loop
     event_proxy: T,
 
-    /// Current title of the window
-    title: String,
+    /// Current title of the window.
+    title: Option<String>,
+
+    /// Default title for resetting it.
+    default_title: String,
+
+    /// Whether to permit updating the terminal title.
+    dynamic_title: bool,
 
     /// Stack of saved window titles. When a title is popped from this stack, the `title` for the
     /// term is set, and the Glutin window's title attribute is changed through the event listener.
-    title_stack: Vec<String>,
+    title_stack: Vec<Option<String>>,
 }
 
 /// Terminal size info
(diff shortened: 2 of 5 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -2128,14 +2107,48 @@ impl<T: EventListener> Handler for Term<T> {
...
+            let title = if !tty::is_conpty() && title.starts_with(' ') {
+                format!("Alacritty {}", title.trim())
+            } else {
+                title.to_owned()
+            };
+
+            self.title = Some(title.clone());
+            self.event_proxy.send_event(Event::Title(title));
+        }
+    }
```
chrisduerr: I'm questioning if this separation of Windows/Unix is really necessary here and what exactly the benefits of `set_title`'s Windows implementation are.

So potentially we can just run `title.trim()` on all platforms and unify the code? If we change that, I think it makes sense to look into that now while already touching the code anyways.

kchibisov: well we need some Windows testing then. I'm always in favor of having less platform specific logic.

chrisduerr: @davidhewitt Can you help with this maybe, since you wrote the original comment?

The comment says that this is about WinPTY setting the title to just ` `, but what we do is *always* prepending `Alacritty `, which seems wrong if the comment is correct? It should be more like `if title == ' ' && winpty { "Alacritty" }`, right?

Also I'm not quite sure under which circumstances the title is set to ` `, does ConPTY reset it to what was initially passed to it at startup? What if the title isn't set in the ConPTY code, does it also set it to ` ` then?

davidhewitt: Yes - the 'title' as far as cmd.exe is concerned is a parameter passed at startup:

https://github.com/alacritty/alacritty/blob/bb0a214a157adc81d1e598602bba7739ecec8672/alacritty_terminal/src/tty/windows/conpty.rs#L143

I'm not sure exactly what cmd.exe's logic for the title is, but it seems to be showing _either_ this startup title plus the cwd, or whatever the hosted application has set the title to.

WinPTY doesn't let us pass this title at startup, instead hard-coding it to ` `. So that's what this code is trying to work around.

In all honesty I think this code can be binned; Alacritty will then behave consistently on all platforms and even though WinPTY + dynamic_title will be slightly "broken", it's a cosmetic issue that will affect a tiny fraction of users.

chrisduerr: @davidhewitt So unless the title is reset on WinPTY, it will always be reasonable and we *shouldn't* change anything about it? So our current logic would be incorrect when the title is set to something that coincidentally starts with some whitespace?

If we don't set the title on ConPTY, should every thing still work out fine just using `Alacritty.exe` as title instead of `Alacritty`?

davidhewitt:
> So unless the title is reset on WinPTY, it will always be reasonable and we shouldn't change anything about it? So our current logic would be incorrect when the title is set to something that coincidentally starts with some whitespace?

Yes. It's become increasingly clear that this attempt to workaround WinPTY's inability to take a title at startup was a mistake.

> If we don't set the title on ConPTY, should every thing still work out fine just using Alacritty.exe as title instead of Alacritty?

I don't completely follow what you're asking here but I think the answer is yes.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `draw_bold_text_with_bright_colors` is now `false`
 - Mirror OSC query terminators instead of always using BEL
 - Increased Beam, Underline, and Hollow Block cursors' line widths
+- Dynamic title is not disabled anymore when `window.title` is set in config
```
nixpulvis: I know this might be out of scope for this PR, but I find it very strange that the setting `window.title` and `dynamic_title` aren't both in the `window` group. Since we're changing the semantics now, maybe it's also a good time to move this configuration?

Otherwise I can open a separate PR, if we agree that this should be done.

chrisduerr: I think @kchibisov might want to play around with this for the configuration changes required for the GUI separation?

nixpulvis: That's fine, but we agree these two configurations should be together correct?

chrisduerr: It depends a bit on the future of our GUI/terminal separation.

Because dynamic_title is an inherent terminal emulation feature, while the current title itself does not need to be known at all by the terminal emulator if we decide to never support the title query escape.

kchibisov: I've already proposed to group `dynamic_title` and `title` somehow(let's say `enum`), I'm not afraid of `yaml` enums here.

chrisduerr: I am, I still don't think enum is a great choice here.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2109,14 +2089,26 @@ impl<T: EventListener> Handler for Term<T> {
...
+
     #[inline]
     fn push_title(&mut self) {
-        trace!("Pushing '{}' onto title stack", self.title);
+        trace!("Pushing '{:?}' onto title stack", self.title);
```
kchibisov: Could we just avoid this `None` thing and push default title? I don't really like `Option<String>` for title stack and for title. We can just set `self.title = self.default_title` and remove all the `Option<String>` thing.

chrisduerr: You're asking to make it impossible to change the title after it has been pushed and retrieved from the stack then?

We can't 'just' change it unless we want to stop supporting this usecase.

CHANGELOG.md
```diff
@@ -70,6 +71,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Decorations visible when in fullscreen on Wayland
 - Window size not persisted correctly after fullscreening on macOS
 - Crash on startup with some locales on X11
+- Live config reload for `window.title`
```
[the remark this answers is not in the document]
chrisduerr: It was never explicitly stated that it required a restart to apply and I thought it was working already before checking. That's why I put it in the `fixed` section. But I suppose our fixed section is already way too big already anyways.

alacritty_terminal/src/term/mod.rs
kchibisov: I've likely commented in a wrong place, treat this comment as a general PR thing. I feel like we don't need things to be `Option`, that's what I'm trying to say.

chrisduerr: That's what I'm saying though. If you want the functionality to behave the way it does right now, we do need it to be an `Option`.

So the question is just if we want to have it the way it is right now and accept the ugly code, or if we want to have worse functionality for the user but cleaner code.

Don't get me wrong, I don't love this option either, I'm fine with `title: Option<String>`, but I think the `title_stack: Vec<Option<String>>` is pretty bad. I was kinda hoping for a comment related to this, but I'm not sure myself what I would prefer. But having a nice reloading functionality that works the way the user would expect definitely seems nice.

CHANGELOG.md
nixpulvis: @kchibisov what is a YAML enum? I've never heard of this, and can't find it in the [YAML spec](https://yaml.org/spec/1.2/spec.html).

chrisduerr: @nixpulvis There is none. That's the problem.

A Rust enum in yaml looks like this:

```yml
enum Test {
    One,
    Two,
}

value: {
    Test: One,
}
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -954,7 +959,16 @@ impl<T> Term<T> {
             self.mode.remove(TermMode::ALTERNATE_SCROLL);
         }
         self.default_cursor_style = config.cursor.style;
+
+        // Update title
         self.dynamic_title = config.dynamic_title();
+        if !self.dynamic_title
+            || (self.title.is_none() && self.default_title != config.window.title)
```
kchibisov: The right expression `self.title.is_none()` is always true for `dynamic_title: false`, during toggle of dynamic title, title should be reset to default. In this way you don't need `Option<String>` at all.

chrisduerr: We don't do `!dynamic_title &&`, but `!dynamic_title ||`. That's a big difference.

Your argument for not needing `Option<String>` at all is incorrect and its introduction was completely unrelated to `dynamic_title`.

alacritty_terminal/src/term/mod.rs
nixpulvis: The data types **must** match what we want the expected behavior to be, never the other way around. That being said, I'm not convinced that we can't have the best of both worlds still...

In the world without both `title` and `title_stack`, I feel like we could have a single `title_stack: Vec<String>` which when empty means "use default" much like how `None` means use default today. I see no value in pushing `None` values to a title stack, other than a minuscule amount of memory saved.

alacritty_terminal/src/term/mod.rs
nixpulvis: Let's keep this discussion together under https://github.com/alacritty/alacritty/pull/3427#discussion_r391945401. But I agree.

kchibisov: didn't noticed `||`.

I don't understand what are you trying to achieve.

I'm expecting reload to work in the following way.

if we do a toggle from `dynamic_title: false` to `dynamic_title: true`, set `title` to `default title`, send new `default title`.
if we do a toggle from `dynamic_title: true` to `dynamic_title: false`, set `title` to `default title`, clear title stack, send new title.
if you reload title, just set `default title` and `title` and send `default title`.

nixpulvis: @kchibisov who does setting `dynamic_title` from `false` to `true` need to update the title? Shouldn't it already be the default, since there's no way to change it? Otherwise I think we're on the same page.

Basically, the behavior of this PR today makes sense to me. My contention is with the implementation.

alacritty_terminal/src/term/mod.rs
chrisduerr: Can't do that, we're still not allowed to pop back to the default title.

alacritty_terminal/src/term/mod.rs
chrisduerr: >  if you reload title, just set default title and title and send default title.

That's exactly what I'm doing here though. If the current title is not changed (see `None` check) and the title has been changed (see `default != config.title`), then we set the title.

That's exactly what's happening here.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1916,8 +1896,8 @@ impl<T: EventListener> Handler for Term<T> {
         self.grid.reset(&Cell::default());
         self.alt_grid.reset(&Cell::default());
         self.scroll_region = Line(0)..self.grid.num_lines();
-        self.title = DEFAULT_NAME.to_string();
         self.title_stack.clear();
```
kchibisov: Should we drop underlying buffer? I feel like we should release as much memory as we can on `reset`.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2129,8 +2122,10 @@ impl<T: EventListener> Handler for Term<T> {
         trace!("Attempting to pop title from stack...");
 
         if let Some(popped) = self.title_stack.pop() {
-            trace!("Title '{}' popped from stack", popped);
-            self.set_title(&popped);
+            trace!("Title '{:?}' popped from stack", popped);
+
+            self.set_title(popped.as_ref().unwrap_or(&self.default_title).clone());
+            self.title = popped;
```
kchibisov: I don't really like that we change to `Default title` and then back to `None`, maybe we should use `Option<String>` in `set_title`.

alacritty_terminal/src/tty/windows/winpty.rs
```diff
@@ -48,7 +48,7 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, _window_id: Option<usize>) ->
         SpawnFlags::AUTO_SHUTDOWN | SpawnFlags::EXIT_AFTER_SHUTDOWN,
         None, // appname
         Some(&cmdline),
-        config.working_directory.as_ref().map(|p| p.as_path()),
+        config.working_directory.as_deref(),
```
kchibisov: You should revert this due to CI failures.

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `draw_bold_text_with_bright_colors` is now `false`
 - Mirror OSC query terminators instead of always using BEL
 - Increased Beam, Underline, and Hollow Block cursors' line widths
+- Dynamic title is not disabled anymore when `window.title` is set in config
+- Live config reload for `window.title`
```
kchibisov: It shouldn't be in `Changed` I guess, it was in `Added` before IIRC, why did you move it?

alacritty_terminal/src/term/mod.rs
```diff
@@ -954,8 +959,17 @@ impl<T> Term<T> {
             self.mode.remove(TermMode::ALTERNATE_SCROLL);
         }
         self.default_cursor_style = config.cursor.style;
+
+        self.default_title = config.window.title.clone();
         self.dynamic_title = config.dynamic_title();
 
+        if self.dynamic_title {
+            let title = self.title.as_ref().unwrap_or(&self.default_title).clone();
+            self.set_title(title);
```
kchibisov: You should just send it and set to `None`, I guess. Also, you should don't only on actual change, not when `self.dynamic_title` is `true`, otherwise you'll change the current title.

alacritty_terminal/src/term/mod.rs
chrisduerr: Yeah, I've had that temporarily. It's probably a good idea. Though it's unfortunate since this is the only place where we need it and it would make things easier not to have it everywhere elese.

CHANGELOG.md
chrisduerr: Oh, it was in `Fixed` and I think I moved it to the wrong section, thanks.

alacritty_terminal/src/term/mod.rs
chrisduerr: Yeah I know, that's exactly what I want to do.

PR #3432, opened 2020-03-08
[WIP] Simplify URL matching

chrisduerr:
This is a big refactor of the URL matching procedure. Instead of
checking for all URLs whenever the terminal redraws, the URL parsing is
now only run on the line below the mouse cursor.

This means that in the best case scenario no URL parsing runs at all
when the mouse is off the screen and the usual case will lead to one
line being checked. To improve the worst case of a massive wrapped line,
the maximum number of lines searched is set to 100.

Since the parsing does not depend on the visible cells anymore, it is
now possible to click on URLs that are only partially visible.

A lot of complexity could also be removed by deleting all mouse cursor
checks from `input.rs`. Instead the current mouse cursor state is now
always checked when redrawing. Since both the terminal and input
handling can cause the cursor to change at any time, this shouldn't
perform any worse but makes it possible to centralize the logic in a
much more sensible manner.


chrisduerr:
There are still some style issues that need to be solved, but the functionality should all be completely done already.

The main question would be if this is a simplification we want to accept, or if there's any actual need for the complexity in our current URL parser.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2183,116 +2164,110 @@ impl IndexMut<Column> for TabStops {
...
-    struct Mock;
-    impl EventListener for Mock {
-        fn send_event(&self, _event: Event) {}
-    }
-
-    #[test]
-    fn semantic_selection_works() {
+    /// Construct a terminal from its content as string.
+    ///
+    /// A `\n` will break line and `\r\n` will break line without wrapping.
```
cole-h:
```suggestion
    /// A `\n` will break [a/the] line and a `\r\n` will break [a/the] line without wrapping.
```

Alternatively:
```suggestion
    /// `\n` will break [a/the] line and `\r\n` will break [a/the] line without wrapping.
```

alacritty/src/event.rs
```diff
@@ -434,6 +437,16 @@ impl<N: Notify + OnResize> Processor<N> {
...
 
+            // Check for URL below cursor
+            let highlighted_url = self.highlighted_url(&terminal);
+
+            // Check for URL highlight change
+            let prev_highlighted = mem::replace(&mut url_highlighted, highlighted_url.is_some());
+            terminal.dirty |= prev_highlighted != url_highlighted;
+
+            // Update mouse cursor shape
+            self.display.window.set_mouse_cursor(self.mouse_cursor(&terminal, url_highlighted));
```
kchibisov: Does it actually improves performance, since you're performing URL search/highlighting inside the terminal lock stage?

chrisduerr: *Improving* performance is not the goal of this PR, but I haven't sufficiently tested it yet. There shouldn't be a measurable performance difference.

kchibisov: Then could you elaborate, what’s the goal of this PR?

chrisduerr: It significantly simplifies the URL matching and removes logic that just isn't necessary for the URLs we query.

It manages to improve our URL matching results, since it allows highlighting URLs regardless of what is visible and at the same time it removes like ~250 lines of code.

As far as I'm concerned, this can do the same thing (and more) our current URL matcher can do, while using simpler code to do it. However of course this approach might be less relevant if we would want to always underline all URLs for example.

alacritty_terminal/src/term/search.rs
```diff
@@ -0,0 +1,390 @@
...
+        None
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use crate::index::{Column, Line};
+    use crate::term::test::mock_term;
```
hhirtz, line 295: `grep -r mock_term alacritty_terminal` yields no results (other than this file).  Also there's a typo on `tests`.

kchibisov: This PR is a WIP and was accidentally half reverted.

chrisduerr:
I don't think there's much of value left here. If things should be simplified it should probably be redone from scratch.

PR #3436, opened 2020-03-09
Add option to pick Linux/BSD backends

### LINKED ISSUE #3340: enable building without X11 for Wayland-only systems
Hi,

I wanted to build Alacritty on a Gentoo box that doesn't have any X11 libs installed. This failes because some X libs are linked to.

I tried to brute-force my way around this by removing all X-related dependencies and code. Unfortunately I'm stuck at
https://github.com/alacritty/alacritty/blob/988a846c52f381bdd2d211d9dbb4b9bf0f9105e5/alacritty/src/main.rs#L169
because there's no waylandish version of this implemented.

Any help in getting this awesome terminal running on Wayland-only is very much appreciated :wink: 
Thanks!

kchibisov:
Fixes #3340.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Packaging
 
 - Minimum Rust version has been bumped to 1.37.0
+- Added ability to pick Linux/BSD backends(X11/Wayland) during build, both are enabled by default
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Packaging
 
 - Minimum Rust version has been bumped to 1.37.0
-- Added ability to pick Linux/BSD backends(X11/Wayland) during build, both are enabled by default
+- Added Rust features `x11` and `wayland` to pick backends, with both enabled by default
 
 ### Changed
 
--- a/alacritty/build.rs
+++ b/alacritty/build.rs
@@ -23,7 +23,7 @@ use embed_resource;
 
 fn main() {
     if cfg!(not(any(feature = "x11", feature = "wayland", target_os = "macos", windows))) {
-        panic!("At least one of the \"x11\"/\"wayland\" features must be enabled");
+        panic!("at least one of the \"x11\"/\"wayland\" features must be enabled");
     }
 
     let hash = rustc_tools_util::get_commit_hash().unwrap_or_default();
--- a/alacritty/build.rs
+++ b/alacritty/build.rs
@@ -23,7 +23,7 @@ use embed_resource;
 
 fn main() {
     if cfg!(not(any(feature = "x11", feature = "wayland", target_os = "macos", windows))) {
-        panic!("Please select at least one Linux/BSD backend!");
+        panic!("Please select at least one Linux/BSD backend using the corresponding features");
     }
 
     let hash = rustc_tools_util::get_commit_hash().unwrap_or_default();
--- a/alacritty/build.rs
+++ b/alacritty/build.rs
@@ -25,6 +25,7 @@ fn main() {
     if cfg!(not(any(feature = "x11", feature = "wayland", target_os = "macos", windows))) {
         panic!("Please select at least one Linux/BSD backend!");
     }
+
     let hash = rustc_tools_util::get_commit_hash().unwrap_or_default();
     println!("cargo:rustc-env=GIT_HASH={}", hash);
 
--- a/alacritty/build.rs
+++ b/alacritty/build.rs
@@ -23,7 +23,7 @@ use embed_resource;
 
 fn main() {
     if cfg!(not(any(feature = "x11", feature = "wayland", target_os = "macos", windows))) {
-        panic!("Please select at least one Linux/BSD backend using the corresponding features");
+        panic!("At least one of the \"x11\"/\"wayland\" features must be enabled");
     }
 
     let hash = rustc_tools_util::get_commit_hash().unwrap_or_default();
--- a/alacritty/build.rs
+++ b/alacritty/build.rs
@@ -22,6 +22,9 @@ use std::path::Path;
 use embed_resource;
 
 fn main() {
+    if cfg!(not(any(feature = "x11", feature = "wayland", target_os = "macos", windows))) {
+        panic!("Please select at least one Linux/BSD backend!");
+    }
     let hash = rustc_tools_util::get_commit_hash().unwrap_or_default();
     println!("cargo:rustc-env=GIT_HASH={}", hash);
 
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -24,7 +24,7 @@ unicode-width = "0.1"
 base64 = "0.11.0"
 terminfo = "0.7.1"
 url = "2"
-copypasta = "0.6.1"
+copypasta = { version = "0.6.3", features = [], default-features = false }
 
 [target.'cfg(unix)'.dependencies]
 nix = "0.16.1"
@@ -44,9 +44,12 @@ mio-anonymous-pipes = "0.1"
 objc = "0.2.2"
 
 [features]
-default = []
+default = ["x11", "wayland"]
+x11 = ["copypasta/x11"]
+wayland = ["copypasta/wayland"]
 nightly = []
 bench = []
+
 
 [dev-dependencies]
 serde_json = "1.0.0"
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -24,7 +24,7 @@ unicode-width = "0.1"
 base64 = "0.11.0"
 terminfo = "0.7.1"
 url = "2"
-copypasta = { version = "0.6.3", features = [], default-features = false }
+copypasta = { version = "0.6.3", default-features = false }
 
 [target.'cfg(unix)'.dependencies]
 nix = "0.16.1"
@@ -49,7 +49,6 @@ x11 = ["copypasta/x11"]
 wayland = ["copypasta/wayland"]
 nightly = []
 bench = []
-
 
 [dev-dependencies]
 serde_json = "1.0.0"
--- a/alacritty_terminal/src/clipboard.rs
+++ b/alacritty_terminal/src/clipboard.rs
@@ -57,7 +57,7 @@ impl Clipboard {
             selection: Some(Box::new(X11ClipboardContext::<X11SelectionClipboard>::new().unwrap())),
         };
 
-        #[cfg(not(any(feature = "x11")))]
+        #[cfg(not(feature = "x11"))]
         return Self::new_nop();
     }
 
--- a/alacritty_terminal/src/clipboard.rs
+++ b/alacritty_terminal/src/clipboard.rs
@@ -18,11 +18,9 @@ use std::ffi::c_void;
 use log::{debug, warn};
 
 use copypasta::nop_clipboard::NopClipboardContext;
-#[cfg(not(any(target_os = "macos", windows)))]
-#[cfg(feature = "wayland")]
+#[cfg(all(not(any(target_os = "macos", windows)), feature = "wayland"))]
 use copypasta::wayland_clipboard;
-#[cfg(not(any(target_os = "macos", windows)))]
-#[cfg(feature = "x11")]
+#[cfg(all(not(any(target_os = "macos", windows)), feature = "x11"))]
 use copypasta::x11_clipboard::{Primary as X11SelectionClipboard, X11ClipboardContext};
 #[cfg(any(feature = "x11", target_os = "macos", windows))]
 use copypasta::ClipboardContext;
--- a/alacritty_terminal/src/clipboard.rs
+++ b/alacritty_terminal/src/clipboard.rs
@@ -18,10 +18,10 @@ use std::ffi::c_void;
 use log::{debug, warn};
 
 use copypasta::nop_clipboard::NopClipboardContext;
-#[cfg(not(any(target_os = "macos", target_os = "windows")))]
+#[cfg(not(any(target_os = "macos", windows)))]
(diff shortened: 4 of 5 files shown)


kchibisov:
To check the build you should do the following, assuming that you're in the root of the repo:

```
cd alacritty
cargo build --no-default-features --features wayland (will create debug build in target/debug/).
cargo run --no-default-features --features wayland (run the thing)
```

alacritty_terminal/Cargo.toml
```diff
@@ -24,7 +24,7 @@ unicode-width = "0.1"
 base64 = "0.11.0"
 terminfo = "0.7.1"
 url = "2"
-copypasta = "0.6.1"
+copypasta = { version = "0.6.3", features = [], default-features = false }
```
chrisduerr: Is this `features = []` necessary?

alacritty_terminal/Cargo.toml
```diff
@@ -44,9 +44,12 @@ mio-anonymous-pipes = "0.1"
...
 
 [features]
-default = []
+default = ["x11", "wayland"]
+x11 = ["copypasta/x11"]
+wayland = ["copypasta/wayland"]
 nightly = []
 bench = []
 
+
```
chrisduerr:
```suggestion
```

alacritty_terminal/src/clipboard.rs
```diff
@@ -19,10 +19,14 @@ use log::{debug, warn};
 
 use copypasta::nop_clipboard::NopClipboardContext;
 #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+#[cfg(feature = "wayland")]
 use copypasta::wayland_clipboard;
 #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+#[cfg(feature = "x11")]
```
chrisduerr: Can we combine these two statements?

alacritty_terminal/src/clipboard.rs
```diff
@@ -19,10 +19,14 @@ use log::{debug, warn};
 
 use copypasta::nop_clipboard::NopClipboardContext;
 #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+#[cfg(feature = "wayland")]
```
chrisduerr: Can we combine these two statements?

alacritty_terminal/src/clipboard.rs
```diff
@@ -36,17 +40,27 @@ impl Clipboard {
...
-        Self {
+        #[cfg(feature = "x11")]
+        return Self {
             clipboard: Box::new(ClipboardContext::new().unwrap()),
             selection: Some(Box::new(X11ClipboardContext::<X11SelectionClipboard>::new().unwrap())),
-        }
+        };
+
+        #[cfg(not(any(feature = "x11")))]
+        return Self::new_nop();
```
chrisduerr: Shouldn't we fail compilation when neither of the backends are selected?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Packaging
 
 - Minimum Rust version has been bumped to 1.37.0
+- Added ability to pick Linux/BSD backends(X11/Wayland) during build, both are enabled by default
```
```suggestion
- Added Rust features `x11` and `wayland` to pick backends, which are both enabled by default
```

cole-h: `pick backends` sounds kinda weird to me -- maybe `choose backends`?

Also `which are both enabled by default` -> `with both enabled by default`, IMHO.

kchibisov: I've decided to avoid `choose` since I read such thing like 'one over another'.

alacritty_terminal/src/clipboard.rs
kchibisov: Not really, you can't have multiple cfg's.

alacritty/build.rs
```diff
@@ -22,6 +22,9 @@ use std::path::Path;
 use embed_resource;
 
 fn main() {
+    if cfg!(not(any(feature = "x11", feature = "wayland", target_os = "macos", windows))) {
+        panic!("Please select at least one Linux/BSD backend!");
+    }
```
chrisduerr:
```suggestion
    }
    
```

alacritty_terminal/src/clipboard.rs
```diff
@@ -18,35 +18,49 @@ use std::ffi::c_void;
 use log::{debug, warn};
 
 use copypasta::nop_clipboard::NopClipboardContext;
-#[cfg(not(any(target_os = "macos", target_os = "windows")))]
+#[cfg(not(any(target_os = "macos", windows)))]
+#[cfg(feature = "wayland")]
```
chrisduerr:
```suggestion
#[cfg(all(not(any(target_os = "macos", windows)), feature = "wayland")]
```

```suggestion
#[cfg(all(not(target_os = "macos"), not(windows)), feature = "wayland")]
```

Wouldn't one of these work?

alacritty/build.rs
```diff
@@ -22,6 +22,10 @@ use std::path::Path;
 use embed_resource;
 
 fn main() {
+    if cfg!(not(any(feature = "x11", feature = "wayland", target_os = "macos", windows))) {
+        panic!("Please select at least one Linux/BSD backend!");
```
nixpulvis: Minor nitpick, but I think I'd word this to include the word "feature" to make it a bit more clear to the user. Perhaps just adding the word feature to the end of the sentence.

Also, an exclamation seems a bit dramatic, don't you think? Most of out messages don't have punctuation, we should leave it off here too.

alacritty_terminal/src/clipboard.rs
```diff
@@ -18,35 +18,47 @@ use std::ffi::c_void;
...
-#[cfg(not(any(target_os = "macos", target_os = "windows")))]
+#[cfg(all(not(any(target_os = "macos", windows)), feature = "wayland"))]
 use copypasta::wayland_clipboard;
-#[cfg(not(any(target_os = "macos", target_os = "windows")))]
+#[cfg(all(not(any(target_os = "macos", windows)), feature = "x11"))]
 use copypasta::x11_clipboard::{Primary as X11SelectionClipboard, X11ClipboardContext};
-use copypasta::{ClipboardContext, ClipboardProvider};
+#[cfg(any(feature = "x11", target_os = "macos", windows))]
+use copypasta::ClipboardContext;
+use copypasta::ClipboardProvider;
```
nixpulvis: I find it hard sometimes to read `#[cfg(...)]`'d `use` statements like this. Would you mind grouping all the conditional imports together below the non-conditional ones? Is this a bad idea for some reason?

alacritty_terminal/src/clipboard.rs
```diff
@@ -18,35 +18,47 @@ use std::ffi::c_void;
...
 
-        Self {
+        #[cfg(feature = "x11")]
+        return Self {
             clipboard: Box::new(ClipboardContext::new().unwrap()),
             selection: Some(Box::new(X11ClipboardContext::<X11SelectionClipboard>::new().unwrap())),
-        }
+        };
+
+        #[cfg(not(any(feature = "x11")))]
```
nixpulvis: Why is this `not(any(...))` of a single feature? Seems like it should be `#[cfg(not(any(feature = "x11", feature = "wayland")))]`

alacritty_terminal/src/clipboard.rs
```diff
@@ -18,35 +18,47 @@ use std::ffi::c_void;
...
     }
 
-    #[cfg(not(any(target_os = "macos", target_os = "windows")))]
-    pub fn new(display: Option<*mut c_void>) -> Self {
-        if let Some(display) = display {
-            let (selection, clipboard) =
-                unsafe { wayland_clipboard::create_clipboards_from_external(display) };
-            return Self { clipboard: Box::new(clipboard), selection: Some(Box::new(selection)) };
+    #[cfg(not(any(target_os = "macos", windows)))]
+    pub fn new(_display: Option<*mut c_void>) -> Self {
```
nixpulvis: `display` should not be given an underscore, as the binding is in fact used. I find this confusing. I personally don't have an issue with shadowing in Rust however, so that's how I'd personally fix this, like it was before.

alacritty_terminal/src/clipboard.rs
chrisduerr: Convention, we currently don't do this anywhere so the current policy is to group the imports just like any other import.

I find imports mixed in with `cfg` statements significantly more difficult to read too, but it only affects the imports that are actually annotated I'd say.

Especially in this block, the two unannotated statements `ClipboardProvider` and `NopClipboardContext` are pretty simple to understand.

So I'd probably vote against changing it, but I'd be open to other suggestions that could work around this issue? But I don't think there are any.

alacritty_terminal/src/clipboard.rs
chrisduerr: The reason why an underscore is used, is because when you compile without the `Wayland` feature, the `display` variable will actually never be read.

I think the underscore is a perfectly fine solution in this case. It's just super nasty code because there are way too many feature flags involved.

alacritty_terminal/src/clipboard.rs
chrisduerr: Because that's the only reason why it could be reached, if `x11` is enabled, the `return Self { ... }` will always be hit and thus trigger a compiler warning about an unreachable statement.

So adding this fallback just for Wayland when X11 is disabled and no Display is present (so XWayland), should be fine.

cole-h: I don't want to put words in @nixpulvis's mouth, but I think the question was why it was left as `any` when it's only checking for a single thing (`feature = "x11`). I tried this change locally and it appears to compile fine.

```suggestion
        #[cfg(not(feature = "x11"))]
```

alacritty/build.rs
```diff
@@ -22,6 +22,10 @@ use std::path::Path;
 use embed_resource;
 
 fn main() {
+    if cfg!(not(any(feature = "x11", feature = "wayland", target_os = "macos", windows))) {
+        panic!("Please select at least one Linux/BSD backend using the corresponding features");
```
chrisduerr:
```suggestion
        panic!("at least one of the "x11"/"wayland" features must be enabled");
```

I think we can be a bit shorter here and don't need to include Linux/BSD, since this error isn't emitted on Wayland/macOS anyways. I also think that we should definitely print the features we're talking about.

kchibisov: Should we capitalize `At` at least?

cole-h: Please :)

chrisduerr: We shouldn't. Rust's panic messages are all lowercase without punctuation by default afaik, so we should follow that.

alacritty/build.rs
```diff
@@ -22,6 +22,10 @@ use std::path::Path;
 use embed_resource;
 
 fn main() {
+    if cfg!(not(any(feature = "x11", feature = "wayland", target_os = "macos", windows))) {
+        panic!("At least one of the \"x11\"/\"wayland\" features must be enabled");
```
chrisduerr:
```suggestion
        panic!("at least one of the \"x11\"/\"wayland\" features must be enabled");
```

Needs to be lowercase to follow Rust's formatting standards, see [here](https://doc.rust-lang.org/src/alloc/string.rs.html#1214) for an example.

alacritty/build.rs
nixpulvis: @chrisduerr yep. https://github.com/rust-lang/api-guidelines/issues/79#issuecomment-316275971

alacritty/build.rs
nixpulvis: Or even, https://github.com/rust-lang/api-guidelines/issues/79#issuecomment-316275971

alacritty/build.rs
chrisduerr: @nixpulvis I knew it's the case from using Rust for a bit, but I didn't know about the api-guidelines repository. That's an amazing resource to help with potential naming problems, thanks a ton for sharing.

alacritty_terminal/src/clipboard.rs
nixpulvis: Maybe better suited for a separate PR, but I could see cleaning these up. Not a big deal though, so I'll ignore it for now.

alacritty/build.rs
chrisduerr: Maybe we should use \`x11\` and \`wayland\` too? I'm not sure if that's standardized?

I was kinda holding out on that because I thought the panic message itself was surrounded with \`, but turns out I misremembered and it's actually '.

kchibisov: I'd prefer to surround with `"` tbh, since ` is too markdownish.

alacritty/build.rs
nixpulvis: Sorry I moved the comment, below :P

alacritty/build.rs
nixpulvis: @kchibisov the quotes are from the `panic!` implementation, we don't have any control over that.

kchibisov: @nixpulvis you don't get the issue. I'm talking about the quotes around `x11` and `wayland`, not about the thing panic inserts there, I'm fine with `'` from it.

chrisduerr: What's wrong with markdownish? Rust documentation uses markdown too.

Since we're talking about strings I think both would be fine (`Debug` for String uses `"` too I believe), but I'm not sure if that's the recommended strategy and uniformity is good here.

I have searched the Rust source code and there's no clear indication for me. There are some panics with `"`, a lot with `\`` but not all with strings obviously and some with `'` used for chars for example. So I think `"` is fine?

chrisduerr: Though we wouldn't need the \\", which might be a good argument already to switch to \`.

But I'll leave this up to you @kchibisov, I'd accept both. But I think I'd personally prefer \` I think.

kchibisov: The problem I see with \` is that they are not that noticeable than `"` when you're trying to find things among all this output from panic, so I'm in favor of using more visible thing `"`. It's not only about generic panic, this panic is an artificial compilation error (read `#error` from C world).

PR #3438, opened 2020-03-10
Remove right click deselection

### LINKED ISSUE #3144: RFE: right click shouldn't clear selected text
given conversations in other issues (#1539, #839) it seems right click isn't supported, which is fine.
however right click does cause text selections to be deselected. which is annoying as old habits die hard.

### System

OS: Linux
Version: 0.4.0
Linux/BSD: Wayland, sway, gnome

stefandevai:
Hi everyone. I'm new to Rust programming and I decided to try to take this easy issue. So now clicking only with the right button won't deselect the text.

Apparently rustfmt removed some redundant commas from input.rs.

Fixes #3144.


kchibisov:
>Apparently rustfmt removed some redundant commas from input.rs.

Please use nightly version of rustfmt. So you should run `cargo +nightly fmt` to format the code.

stefandevai:
> > Apparently rustfmt removed some redundant commas from input.rs.
> 
> Please use nightly version of rustfmt. So you should run `cargo +nightly fmt` to format the code.

Done.

alacritty/src/input.rs
```diff
@@ -377,6 +377,13 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
         point.line = min(point.line, self.ctx.terminal().grid().num_lines() - 1);
 
         self.ctx.mouse_mut().click_state = match self.ctx.mouse().click_state {
+            // Do nothing if clicked with the right mouse button
+            ClickState::Click
+                if mouse.right_button_state == ElementState::Pressed
+                    && mouse.left_button_state == ElementState::Released =>
+            {
+                ClickState::Click
+            },
```
chrisduerr: If we just ignore all right clicks, it won't be possible to right-click in applications anymore.

However that is still necessary for Alacritty to function properly. So please make sure the `self.mouse_report` code can still be reached.

stefandevai: That's right, I didn't notice the mouse report.

Now it's fixed, it does the necessary checks and then reports the mouse.

alacritty/src/input.rs
```diff
@@ -377,6 +377,20 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
+            {
+                if !self.ctx.modifiers().shift()
+                    && self.ctx.terminal().mode().intersects(TermMode::MOUSE_MODE)
+                {
+                    // 2 is the right button code
+                    self.mouse_report(2, ElementState::Pressed);
+                    return;
+                }
+                ClickState::Click
+            }
```
chrisduerr: Could you move this together with the left click logic, just excluding the statements we don't want to execute with a conditional? We don't want to duplicate code.

stefandevai: Alright, I think it looks better now :)

kchibisov:
Should we clear the selection if you're in mouse mode? I think we should.

alacritty/src/input.rs
```diff
@@ -394,17 +394,22 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
                 ClickState::TripleClick
             }
             _ => {
-                // Don't launch URLs if this click cleared the selection
-                self.ctx.mouse_mut().block_url_launcher = !self.ctx.selection_is_empty();
+                // Only clear selection if left or middle button are pressed
+                if mouse.left_button_state == ElementState::Pressed
+                    || mouse.middle_button_state == ElementState::Pressed
```
chrisduerr: Why would we remove it for RMB, but still allow selecting with MMB? To me it seems like we'd either use LMB only, or all of them?

kchibisov:
>Do you mean with left click or with middle/right? @kchibisov

Oh, I forgot to mention that I'm doing right click. Like I've selected something, then performed action with right click in mouse mode capable app, but my selection stays on this PR. I'm expecting it to deselect tbh. Note, If I press right click with shift in mouse mode app, it shouldn't select.

stefandevai:
> Should we clear the selection if you're in mouse mode and not pressing `Shift`? I think we should.

Okay, but can you just explain why we should maintain the selection if pressing shift? It doesn't make a lot of sense to me.

Also, can you give me an example of an app that uses RMB in the way you mention so I can try it here?

stefandevai:
In any case, I've updated the PR taking in consideration mouse mode and pressing shift.

chrisduerr:
> Oh, I forgot to mention that I'm doing right click. Like I've selected something, then performed action with right click in mouse mode capable app, but my selection stays on this PR. I'm expecting it to deselect tbh. Note, If I press right click with shift in mouse mode app, it shouldn't select.

I disagree with this statement. Right clicking while mouse mode is enabled shouldn't inherently clear the selection. Your argument is that the content below the selection might update, however that's completely unrelated to the right click. The content could update without pressing anything too. So we should use the content's change as cause for clearing the selection (which isn't super easily possible), not the right click.

So I think it would be best if right click never cleared any selection, no matter what mode you're in.

In fact, I'd question if left click while in mouse mode should clear the selection at all. I'd say it very clearly shouldn't, which is what many other terminal emulators do. Selection should only cleared if the mouse button pressed is LMB and if either shift is pressed or mouse mode is disabled.

stefandevai:
> Selection should only cleared if the mouse button pressed is LMB and if either shift is pressed or mouse mode is disabled.

So even if mouse button is not LMB but Shift is pressed we should clear the selection?
For instance, the whole clear condition would be `(button == LMB || (button != LMB && (shift_pressed || !mouse_mode)))`

chrisduerr:
> So even if mouse button is not LMB but Shift is pressed we should clear the selection?

No, more like this: `button == LMB && (shift || !mouse_mode)`.

So we only update the selection in cases where the user is able to make a selection. If the user isn't able to start a selection, we do not clear it either. Does that make sense?

stefandevai:
> So we only update the selection in cases where the user is able to make a selection. If the user isn't able to start a selection, we do not clear it either. Does that make sense?

Yes, sounds right to me.

alacritty/src/input.rs
```diff
@@ -394,17 +394,24 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
                 ClickState::TripleClick
             }
             _ => {
-                // Don't launch URLs if this click cleared the selection
-                self.ctx.mouse_mut().block_url_launcher = !self.ctx.selection_is_empty();
+                // Only clear selection if LMB is pressed and
+                // if either SHIFT is pressed or mouse mode is disabled
```
chrisduerr: This comment is basically just as verbose as the code itself, so it doesn't really improve the situation.

It's also not accurate since this is also about starting a selection, not only about clearing it.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Block cursor is no longer inverted at the start/end of a selection
+- Right clicking no longer clears a selection
```
chrisduerr: Neither does middle mouse. We should probably phrase this to say that `LMB` is now the only button which clears the selection:

```suggestion
- Only left click without mouse mode clears the selection
```

cole-h: TBH, I like how you put it in your comment more than in your suggested change.  Slightly reworded:

```suggestion
- Clicking with the left mouse button is now the only way to clear selections
```

chrisduerr: That does not mention mouse mode though. So it is less accurate than my suggestion with a longer changelog entry.

It is also not true that clicking with left mouse is the only way to clear the selection.

I'm open to other suggestions though, I just tried to keep it short. My comment does not accurately state that this is about removing right/middle/other mouse button clicks but instead focuses more on mouse mode.

cole-h: Fair. Your suggestion is fine, then. Wording is a little awkward (IMHO),  but nothing's perfect, and I'd hate to block or otherwise delay this on changelog bikeshedding :P

chrisduerr: I mean if you have a better suggestions, I'd love to hear it. As long as it hasn't been changed yet there's still time.

stefandevai:
> It behaves how I would expect now, just some minor changelog things left and then it should be good to go.

Great! :)

Issue #2653, opened 2019-07-18
Clarify MacOS Brew Instructions for extra files


jessebett:
The installation [instructions for MacOS in the README](https://github.com/jwilm/alacritty#macos) use homebrew. Following the command it gives the advice to set up additional things.
>Once the cask is installed, it is recommended to setup the manual page, shell completions, and terminfo definitions.

However, those instructions provide scripts which reference local folders like `extra`, e.g. 
```
sudo mkdir -p /usr/local/share/man/man1
gzip -c extra/alacritty.man | sudo tee /usr/local/share/man/man1/alacritty.1.gz > /dev/null
```

Following the macOS install procedure, it is not clear where this folders are.

chrisduerr:
As far as I know, they aren't anywhere. This is a brew issue not an Alacritty one. You have to clone the repository and copy it yourself as far as I know, though I'm not super into the odd macOS package management system.

jessebett:
Can the README be updated to something which instructs you to clone the repo :

>Once the cask is installed, it is recommended to clone this repo to setup the manual page, shell completions, and terminfo definitions.

chrisduerr:
I'd think that through the links to the INSTALL.md, most of that should be explained in the document already, however since it links to specific sections I can see how that might confuse some users who aren't super familiar with the process. So I'd be fine with pointing out explicitly that these instructions require a copy of Alacritty's repo to be completed (though not necessarily require a full clone, but people that care about that should know that).

hsxfjames:
I see the files like `alacritty.1.gz` and `alacritty.info` can be download from release page of this repo. I think https://github.com/Homebrew/homebrew-cask/blob/master/Casks/alacritty.rb may have to automatically copy and link these files, so users don't need to do this by themself.

hsxfjames:
I see the files like `alacritty.1.gz` and `alacritty.info` can be downloaded from release page of this repo. I think https://github.com/Homebrew/homebrew-cask/blob/master/Casks/alacritty.rb might be able to automatically copy and link these files, so users don't need to do this by themself. A cask could execute system command in postflight section.

chrisduerr:
> I think https://github.com/Homebrew/homebrew-cask/blob/master/Casks/alacritty.rb might be able to automatically copy and link these files, so users don't need to do this by themself, since a cask could execute system command in postflight section.

The brew package is maintained externally, so proposals about a change to it need to be addressed to its maintainers.

0az:
[...]

PR #3441, opened 2020-03-10
Add manpage, terminfo, and completions to macOS App bundle (#2653)

0az:
- Add manpage, terminfo, and completions to macOS App bundle (#2653)
- Update README with new macOS extras/completions location (#2653)
- Update changelog

I didn't edit INSTALL.md as that's for manual installation which requires the repo anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Live config reload for `window.title`
+- Added manpage, terminfo, and completions to macOS application bundle.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Minimum Rust version has been bumped to 1.41.0
 - Prebuilt Linux binaries have been removed
+- Added manpage, terminfo, and completions to macOS application bundle
 
 ### Removed
 
@@ -81,8 +82,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Live config reload for `window.title`
-- Added manpage, terminfo, and completions to macOS application bundle.
-- Added manpage, terminfo, and completions to macOS application bundle
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Live config reload for `window.title`
 - Added manpage, terminfo, and completions to macOS application bundle.
+- Added manpage, terminfo, and completions to macOS application bundle
 
 ### Changed
 
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ $(APP_NAME): $(TARGET)
 	@mkdir -p $(APP_EXTRAS_DIR)
 	@mkdir -p $(APP_COMPLETIONS_DIR)
 	@gzip -c $(MANPAGE) > $(APP_EXTRAS_DIR)/alacritty.1.gz
-	@tic -xe alacritty,alacritty-direct $(TERMINFO) -o $(APP_EXTRAS_DIR)
+	@tic -xe alacritty,alacritty-direct -o $(APP_EXTRAS_DIR) $(TERMINFO)
 	@cp -fRp $(APP_TEMPLATE) $(APP_DIR)
 	@cp -fp $(APP_BINARY) $(APP_BINARY_DIR)
 	@cp -fp $(COMPLETIONS) $(APP_COMPLETIONS_DIR)
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,11 @@ TARGET = alacritty
 ASSETS_DIR = extra
 RELEASE_DIR = target/release
 COMPLETIONS_DIR = $(ASSETS_DIR)/completions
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -3,10 +3,17 @@ TARGET = alacritty
 APP_NAME = Alacritty.app
 ASSETS_DIR = extra
 RELEASE_DIR = target/release
+COMPLETIONS_DIR = $(ASSETS_DIR)/completions
+
+EXTRAS = $(ASSETS_DIR)/alacritty.info $(ASSETS_DIR)/alacritty.man
+COMPLETIONS = $(COMPLETIONS_DIR)/_alacritty $(COMPLETIONS_DIR)/alacritty.bash $(COMPLETIONS_DIR)/alacritty.fish
+
```
chrisduerr: If we want to use whitespace to separate things logically, we should have some reason to it. I currently don't really see why we would add a newline before and after extra/completions.

We can probably move `APP_NAME` just before `APP_TEMPLATE` and then combine `ASSETS_DIR`, `RELEASE_DIR`, `COMPLETIONS_DIR`, `EXTRAS` and `COMPLETIONS` into one block.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Packaging
 
 - Minimum Rust version has been bumped to 1.37.0
+- Added manpage, terminfo, and completions to macOS application bundle.
```
```suggestion
- Added manpage, terminfo, and completions to macOS application bundle
```

Makefile
```diff
@@ -3,10 +3,17 @@ TARGET = alacritty
 APP_NAME = Alacritty.app
 ASSETS_DIR = extra
 RELEASE_DIR = target/release
+COMPLETIONS_DIR = $(ASSETS_DIR)/completions
+
+EXTRAS = $(ASSETS_DIR)/alacritty.info $(ASSETS_DIR)/alacritty.man
```
chrisduerr: Should we maybe precompile these, so they only have to be moved to the right location without requiring any tools to install them?

0az: Yeah, that's probably a good idea.

0az: I'm not very familiar with tic/terminfo. As for manpages, I think all that needs to be done is rename `alacritty.man` to `alacritty.1`.

<details>
<summary>Investigating terminfo</summary>

 `tic -x -o /tmp -e alacritty,alacritty-direct alacritty.info` creates `/tmp/61/alacritty` and `/tmp/61/alacritty-direct`. Based on Kitty's [build-terminfo](https://github.com/kovidgoyal/kitty/blob/10435c23c275b20f36bb208728830f15405a2077/build-terminfo#L14-L31), this is what needs to get copied over to `~/.terminfo` or similar.

However, I have no idea what the `61` signifies, nor the `78` in the path of Kitty's, or the `x` that results from using Kitty's `ssh` kitten.

I suppose this is why Kitty just overwrites `$TERMINFO` and points it to `kitty.app/Contents/Frameworks/kitty/terminfo` – it eliminates all of this hassle. I suppose that's a possible future feature.
</summary>

chrisduerr: > I think all that needs to be done is rename alacritty.man to alacritty.1.

This is incorrect. See our `INSTALL.md` where all this is explained. The `tic` command we need to run can also be found there, though for the output directory we should probably use `target` instead of `/tmp`.

> However, I have no idea what the 61 signifies, nor the 78 in the path of Kitty's, or the x that results from using Kitty's ssh kitten.

You don't really need to know, since that's not relevant. But usually the directory is named after the terminfo file. So the `alacritty` terminfo can be found in `/usr/share/terminfo/a/alacritty`, but `xterm-256color` you would find in `/usr/share/terminfo/x/xterm-256color`.

> I suppose this is why Kitty just overwrites $TERMINFO and points it to kitty.app/Contents/Frameworks/kitty/terminfo – it eliminates all of this hassle.

There really shouldn't be any hassle, this is all quite standard behavior for terminal emulators. Hacking around that will only lead to confusing users that actually know what they're doing, which I see as the primary userbase of Alacritty.

nixpulvis:
Just out of curiosity, would we be interested in creating an installer for macOS, so these files could be placed in the correct location automatically? I'm unsure how hard this really is, so maybe it's a bad idea. But I do know that `brew cask` supports it.

chrisduerr:
> Just out of curiosity, would we be interested in creating an installer for macOS, so these files could be placed in the correct location automatically?

I really don't want to. Having to deal with that kind of stuff is just always a giant mess and never pleasant to work with. At least the Windows installer should be a pretty good example for that.

Makefile
0az: Courtesy @Celti:

> ```
> + modify the support for filesystems which do not support mixed-case to
>   generate 2-character (hexadecimal) codes for the lower-level of the
>   filesystem terminfo database (request by Michail Vidiassov).
> ```
> https://invisible-island.net/ncurses/NEWS.html#index-t20071117

Well, this explains a lot.

I think this is also set at compile time, so running tic during the install process makes a lot more sense now.

DrGo:
Hello,
will the manpage be included in the next release for macos?
Thanks,

0az:
A few things have fallen off of my radar due to current events. I'll try to finish this up sometime in June or July.

DrGo:
very understandable, many thanks!

chrisduerr:
CI results to verify this is working are available here:
https://travis-ci.org/github/chrisduerr/alacritty/builds/704830897

nixpulvis:
Should probably at least look at this before merging: https://github.com/alacritty/alacritty/issues/2653#issuecomment-646403259

chrisduerr:
> Should probably at least look at this before merging: #2653 (comment)

That should be resolved with this PR. It's not alacritty.1, but alacritty.1.gz should be the correct name anyways.

PR #3455, opened 2020-03-14
Use config colors to theme Wayland decorations

### LINKED ISSUE #2092: Take advantage of winit's new Wayland theming capability
In tomaka/winit#775, support for theming the title bar was added for Wayland for better UI integration.  Can Alacritty expose support for this through the config file (even if in some limited way)?  That would be awesome!

kchibisov:
Fixes #2092.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Mirror OSC query terminators instead of always using BEL
 - Increased Beam, Underline, and Hollow Block cursors' line widths
 - Dynamic title is not disabled anymore when `window.title` is set in config
+- Wayland decorations are now based on config colorscheme
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Block cursor is no longer inverted at the start/end of a selection
 - Preserve selection on non-LMB or mouse mode clicks
+- Wayland client side decorations are now based on config colorscheme
 
 ### Fixed
 - Tabstops not being reset with `reset`
@@ -46,7 +47,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Mirror OSC query terminators instead of always using BEL
 - Increased Beam, Underline, and Hollow Block cursors' line widths
 - Dynamic title is not disabled anymore when `window.title` is set in config
-- Wayland client side decorations are now based on config colorscheme
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Mirror OSC query terminators instead of always using BEL
 - Increased Beam, Underline, and Hollow Block cursors' line widths
 - Dynamic title is not disabled anymore when `window.title` is set in config
-- Wayland decorations are now based on config colorscheme
+- Wayland client side decorations are now based on config colorscheme
 
 ### Fixed
 
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -142,6 +142,13 @@ impl Display {
         // Spawn window
         let mut window = Window::new(event_loop, &config, size)?;
 
+        #[cfg(not(any(target_os = "macos", windows)))]
+        {
+            if event_loop.is_wayland() {
+                window.set_wayland_theme(&config);
+            }
+        }
+
         let dpr = window.scale_factor();
         info!("Device pixel ratio: {}", dpr);
 
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -145,7 +145,7 @@ impl Display {
         #[cfg(not(any(target_os = "macos", windows)))]
         {
             if event_loop.is_wayland() {
-                window.set_wayland_theme(&config);
+                window.set_wayland_theme(&config.colors);
             }
         }
 
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -142,13 +142,6 @@ impl Display {
         // Spawn window
         let mut window = Window::new(event_loop, &config, size)?;
 
-        #[cfg(not(any(target_os = "macos", windows)))]
-        {
-            if event_loop.is_wayland() {
-                window.set_wayland_theme(&config.colors);
-            }
-        }
-
         let dpr = window.scale_factor();
         info!("Device pixel ratio: {}", dpr);
 
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -14,6 +14,8 @@ use glutin::dpi::PhysicalSize;
 use glutin::event::{ElementState, Event as GlutinEvent, ModifiersState, MouseButton, WindowEvent};
 use glutin::event_loop::{ControlFlow, EventLoop, EventLoopProxy, EventLoopWindowTarget};
 use glutin::platform::desktop::EventLoopExtDesktop;
+#[cfg(not(any(target_os = "macos", windows)))]
+use glutin::platform::unix::EventLoopWindowTargetExtUnix;
 use log::{debug, info, warn};
 use serde_json as json;
 
@@ -540,6 +542,13 @@ impl<N: Notify + OnResize> Processor<N> {
 
                             let font = config.font.clone().with_size(*processor.ctx.font_size);
                             processor.ctx.display_update_pending.font = Some(font);
+                        }
+
+                        #[cfg(not(any(target_os = "macos", windows)))]
+                        {
+                            if processor.ctx.event_loop.is_wayland() {
+                                processor.ctx.window.set_wayland_theme(&config);
+                            }
                         }
 
                         *processor.ctx.config = config;
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -547,7 +547,7 @@ impl<N: Notify + OnResize> Processor<N> {
                         #[cfg(not(any(target_os = "macos", windows)))]
                         {
                             if processor.ctx.event_loop.is_wayland() {
-                                processor.ctx.window.set_wayland_theme(&config);
+                                processor.ctx.window.set_wayland_theme(&config.colors);
                             }
                         }
 
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -663,33 +663,36 @@ impl<N: Notify + OnResize> Processor<N> {
         processor.ctx.message_buffer.remove_target(LOG_TARGET_CONFIG);
         processor.ctx.display_update_pending.message_buffer = Some(());
 
-        if let Ok(config) = config::reload_from(&path) {
-            let options = Options::new();
-            let config = options.into_config(config);
+        let config = match config::reload_from(&path) {
+            Ok(config) => config,
+            Err(_) => return,
+        };
 
-            processor.ctx.terminal.update_config(&config);
+        let options = Options::new();
+        let config = options.into_config(config);
 
-            if processor.ctx.config.font != config.font {
-                // Do not update font size if it has been changed at runtime
-                if *processor.ctx.font_size == processor.ctx.config.font.size {
-                    *processor.ctx.font_size = config.font.size;
-                }
+        processor.ctx.terminal.update_config(&config);
 
-                let font = config.font.clone().with_size(*processor.ctx.font_size);
-                processor.ctx.display_update_pending.font = Some(font);
+        if processor.ctx.config.font != config.font {
+            // Do not update font size if it has been changed at runtime
+            if *processor.ctx.font_size == processor.ctx.config.font.size {
+                *processor.ctx.font_size = config.font.size;
             }
 
-            #[cfg(not(any(target_os = "macos", windows)))]
-            {
-                if processor.ctx.event_loop.is_wayland() {
-                    processor.ctx.window.set_wayland_theme(&config.colors);
-                }
-            }
-
-            *processor.ctx.config = config;
-
-            processor.ctx.terminal.dirty = true;
+            let font = config.font.clone().with_size(*processor.ctx.font_size);
+            processor.ctx.display_update_pending.font = Some(font);
         }
+
+        #[cfg(not(any(target_os = "macos", windows)))]
+        {
+            if processor.ctx.event_loop.is_wayland() {
+                processor.ctx.window.set_wayland_theme(&config.colors);
+            }
+        }
+
+        *processor.ctx.config = config;
+
+        processor.ctx.terminal.dirty = true;
     }
 
     // Write the ref test results to the disk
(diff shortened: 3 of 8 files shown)


alacritty/src/window.rs
```diff
@@ -361,6 +366,27 @@ impl Window {
...
+        let theme = AlacrittyWaylandTheme {
+            foreground,
+            background,
+            dim_foreground,
+            hovered_close_icon,
+            hovered_minimize_icon,
+            hovered_maximize_icon,
+        };
+        self.window().set_wayland_theme(theme);
+    }
```
nixpulvis: @kchibisov I thought you said this wouldn't need to be handled inside of Alacritty. It seems unfortunate to need to make all these choices for our users. What if my applications should all use blue to maximize?

kchibisov: You don't have control over it anyways, since it's a client side decorations. Before we had random white Qt'ish color scheme, now we have something that matches content. You can't just draw GTK decorations, etc. Providing direct config option isn't an option, so we should derive it our of users colorscheme.

nixpulvis: I feel like I may be misremembering a conversation... but I was under the impression we were against CSDs, in favor of SSDs. I can maybe understand wanting an escape hatch from poorly configured SSDs, but I really wish it wasn't inside out codebase.

Here's what I'm remembering: https://github.com/alacritty/alacritty/pull/2047#issuecomment-461925986

kchibisov: We support both CSD and SSD this option is for CSD theming, since we control them. We do support CSD and we're not against them. So on sway you'll use SSD on GNOME CSD.

nixpulvis: Sorry if this is obvious, but where do we "support both CSD and SSD" ATM?

nixpulvis: FWIW, my main issue here is with the plan for configuration of things like this. I don't really want to block enhancements that make Alacritty look better for users, since people will be happier to use Alacritty if it looks good.

kchibisov: >Sorry if this is obvious, but where do we "support both CSD and SSD" at the moment?

SCTK handles them for us.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `draw_bold_text_with_bright_colors` is now `false`
 - Mirror OSC query terminators instead of always using BEL
 - Increased Beam, Underline, and Hollow Block cursors' line widths
+- Wayland decorations are now based on config colorscheme
```
```suggestion
- Wayland client side decorations are now based on config colorscheme
```

alacritty/src/window.rs
```diff
@@ -361,6 +366,27 @@ impl Window {
         self.window().wayland_display()
     }
 
+    #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+    pub fn set_wayland_theme(&mut self, config: &Config) {
```
chrisduerr: Do we need to pass `config` here, when all we access is `config.colors`?

alacritty/src/window.rs
```diff
@@ -361,6 +366,27 @@ impl Window {
...
+        let hovered_close_icon = config.colors.normal().red;
+        let hovered_maximize_icon = config.colors.normal().green;
+        let hovered_minimize_icon = config.colors.normal().yellow;
+        let foreground = config.colors.primary.foreground;
+        let background = config.colors.primary.background;
+        let dim_foreground =
+            config.colors.primary.dim_foreground.unwrap_or_else(|| foreground * 0.66);
```
chrisduerr: Not a fan of this, since we duplicate the dimming code here.

It might be better to pass `&config.colors` and `term.colors` so we can do it properly?

alacritty/src/window.rs
```diff
@@ -386,6 +412,95 @@ impl Window {
...
+                self.hovered_close_icon.g,
+                self.hovered_close_icon.b,
+            ],
+            WaylandThemeButtonState::Idle => {
+                [0xff, self.foreground.r, self.foreground.g, self.foreground.b]
+            },
+            WaylandThemeButtonState::Disabled => {
+                [0xff, self.dim_foreground.r, self.dim_foreground.g, self.dim_foreground.b]
+            },
+        }
```
chrisduerr: I wonder if we could maybe write a little helper for that which would shorten things down?

Like `self.button_icon_color(status, self.hovered_close_icon)`.

Could then use that for all three buttons with just a single line in each of the trait implementations.

alacritty/src/window.rs
nixpulvis: OK, point taken. I'd like to propose moving these out into their own `wayland.rs` file somewhere however.

chrisduerr: I'd just keep it simple with a `wayland_theme.rs` file for now and if we ever need more platform specific modules, we can separate it more.

alacritty/src/display.rs
```diff
@@ -142,6 +142,13 @@ impl Display {
         // Spawn window
         let mut window = Window::new(event_loop, &config, size)?;
 
+        #[cfg(not(any(target_os = "macos", windows)))]
+        {
+            if event_loop.is_wayland() {
+                window.set_wayland_theme(&config);
+            }
+        }
```
nixpulvis: Is there a reason these aren't marked with a `#[cfg(feature = "wayland")]`?

kchibisov: well, there's no reason to opt out specific features code, since it doesn't affect compilation.

alacritty/src/window.rs
kchibisov: You don't have access to terminal colors and I don't really like adding an option to get them, maybe we should keep this simple for now?

chrisduerr: I'm fine with keeping it simple, I'm not fine with keeping in bug prone.

If we wanted to keep it simple, we'd have to remove the dimming.

Why do we not have access to the terminal colors, because this is executed right at startup?

kchibisov: I mean they are `private` right now for us. We do have access for them, but it's about adding `pub` or methods to get them. This is not about error prone or not, let's say I've just dimmed color by some value if it wasn't available. I could remove use of dim and just use `0.64 * foreground`.

I feel like we can use dark black though.

chrisduerr: > This is not about error prone or not, let's say I've just dimmed color by some value if it wasn't available. I could remove use of dim and just use 0.64 * foreground.

This absolutely is about introducing bugs. Because if I change the dim colors in our color code, but forget to update the wayland theme, that's a bug.

> I feel like we can use dark black though.

No. That doesn't work for light themes.

kchibisov: Oh yeah, we can use just `white`.

kchibisov:
There was some interest in playing with colors from @chrisduerr , so leaving this `dim` issue up to him. The only option I see is to either add some simple algorithm to get nice dim or to use config dim and add a method to `Term` to get it.

chrisduerr:
@kchibisov Just fyi, I have no plan to play with this in the near future.

I'm open to suggestions from others and I do not particularly care much about the way the theme actually looks as long as reasonable colors from the user's colorscheme is picked. I just cannot approve an approach that uses bug-prone copy/pasting rather than proper clean code reuse.

kchibisov:
Updated to reuse dim factor from terminal.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1260,6 +1260,11 @@ impl<T> Term<T> {
         self.colors[NamedColor::Background]
     }
 
+    #[inline]
+    pub fn dim_foreground_color(&self) -> Rgb {
+        self.colors[NamedColor::DimForeground]
+    }
```
chrisduerr: This is never used?

alacritty/src/window.rs
```diff
@@ -361,6 +368,26 @@ impl Window {
...
     }
 
+    #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+    pub fn set_wayland_theme(&mut self, colors: &TerminalColors) {
+        let hovered_close_icon = colors.normal().red;
+        let hovered_maximize_icon = colors.normal().green;
+        let hovered_minimize_icon = colors.normal().yellow;
+        let foreground = colors.primary.foreground;
+        let background = colors.primary.background;
+        let dim_foreground = colors.primary.dim_foreground.unwrap_or(foreground * DIM_FACTOR);
```
chrisduerr: Didn't you say that dim foreground isn't a good choice, since it doesn't work with light themes?

alacritty/src/window.rs
```diff
@@ -33,14 +33,21 @@ use image::ImageFormat;
 #[cfg(not(any(target_os = "macos", windows)))]
 use x11_dl::xlib::{Display as XDisplay, PropModeReplace, XErrorEvent, Xlib};
 
-use alacritty_terminal::config::{Decorations, StartupMode, WindowConfig};
+use alacritty_terminal::config::{
+    Colors as TerminalColors, Decorations, StartupMode, WindowConfig,
```
chrisduerr: Why did you rename this import, especially to `TerminalColors`? There is no conflict, right? And if anything, aren't these the config colors?

alacritty/src/window.rs
```diff
@@ -361,6 +368,26 @@ impl Window {
...
+        let dim_foreground = colors.primary.dim_foreground.unwrap_or(foreground * DIM_FACTOR);
+
+        let theme = AlacrittyWaylandTheme {
+            foreground,
+            background,
+            dim_foreground,
+            hovered_close_icon,
+            hovered_minimize_icon,
+            hovered_maximize_icon,
+        };
```
chrisduerr: There's a lot of crap in this function body just taking things from `colors` and putting it inside of `AlacrittyWaylandTheme`. Could we just make a `new` function for the theme which takes a color, so that `window.rs` is less impacted?

alacritty_terminal/src/term/mod.rs
kchibisov: I thought I've removed it. Thx.

alacritty/src/window.rs
kchibisov: Well, users can define their own dim foreground and has a control over the scheme, so I don't see any problem here.

alacritty/src/window.rs
```diff
@@ -33,6 +33,8 @@ use image::ImageFormat;
 #[cfg(not(any(target_os = "macos", windows)))]
 use x11_dl::xlib::{Display as XDisplay, PropModeReplace, XErrorEvent, Xlib};
 
+#[cfg(not(any(target_os = "macos", windows)))]
+use alacritty_terminal::config::Colors;
 use alacritty_terminal::config::{Colors, Decorations, StartupMode, WindowConfig};
```
cole-h: Oops.

```suggestion
use alacritty_terminal::config::{Decorations, StartupMode, WindowConfig};
```

alacritty/src/window.rs
```diff
@@ -361,6 +366,13 @@ impl Window {
         self.window().wayland_display()
     }
 
+    #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+    pub fn set_wayland_theme(&mut self, colors: &Colors) {
+        let theme = AlacrittyWaylandTheme::new(colors);
+
+        self.window().set_wayland_theme(theme);
```
chrisduerr:
```suggestion
        self.window().set_wayland_theme(AlacrittyWaylandTheme::new(colors));
```

alacritty/src/window.rs
```diff
@@ -41,6 +43,9 @@ use alacritty_terminal::term::{SizeInfo, Term};
 use crate::config::Config;
 use crate::gl;
 
+#[cfg(not(any(target_os = "macos", windows)))]
+use crate::wayland_theme::AlacrittyWaylandTheme;
```
chrisduerr:
```suggestion
#[cfg(not(any(target_os = "macos", windows)))]
use crate::wayland_theme::AlacrittyWaylandTheme;
```

alacritty/src/event.rs
```diff
@@ -542,6 +544,13 @@ impl<N: Notify + OnResize> Processor<N> {
                             processor.ctx.display_update_pending.font = Some(font);
                         }
 
+                        #[cfg(not(any(target_os = "macos", windows)))]
+                        {
+                            if processor.ctx.event_loop.is_wayland() {
+                                processor.ctx.window.set_wayland_theme(&config.colors);
+                            }
+                        }
```
chrisduerr: This is getting kinda big, maybe we should consider moving the `ConfigReload` handling to a separate method?

alacritty/src/display.rs
```diff
@@ -142,6 +142,13 @@ impl Display {
         // Spawn window
         let mut window = Window::new(event_loop, &config, size)?;
 
+        #[cfg(not(any(target_os = "macos", windows)))]
+        {
+            if event_loop.is_wayland() {
+                window.set_wayland_theme(&config.colors);
+            }
+        }
```
chrisduerr: Can't we put this in `window.rs`?

alacritty/src/event.rs
```diff
@@ -675,6 +654,44 @@ impl<N: Notify + OnResize> Processor<N> {
...
+    pub fn reload_config<T>(
+        path: &PathBuf,
+        processor: &mut input::Processor<T, ActionContext<N, T>>,
+    ) where
+        T: EventListener,
+    {
+        processor.ctx.message_buffer.remove_target(LOG_TARGET_CONFIG);
+        processor.ctx.display_update_pending.message_buffer = Some(());
+
+        if let Ok(config) = config::reload_from(&path) {
```
chrisduerr, line 666:
```suggestion
        if let Ok(config) = config::reload_from(&path) {
        let config = match config::reload(&path) {
            Ok(config) => config,
            Err(_) => return, // Maybe log to debug?
        }
```

alacritty/src/wayland_theme.rs
```diff
@@ -0,0 +1,83 @@
+use glutin::platform::unix::{ButtonState, Theme as WaylandTheme};
+
+use alacritty_terminal::config::Colors;
+use alacritty_terminal::term::color::Rgb;
+use alacritty_terminal::term::color::DIM_FACTOR;
```
chrisduerr, line 5:
```suggestion
use alacritty_terminal::term::color::{Rgb, DIM_FACTOR};
```

alacritty/src/window.rs
```diff
@@ -158,6 +162,10 @@ impl Window {
                 if let Some(parent_window_id) = config.window.embed {
                     x_embed_window(windowed_context.window(), parent_window_id);
                 }
+            } else {
+                let theme = AlacrittyWaylandTheme::new(&config.colors);
+
+                windowed_context.window().set_wayland_theme(theme);
```
chrisduerr, line 168:
```suggestion
                let theme = AlacrittyWaylandTheme::new(&config.colors);
                windowed_context.window().set_wayland_theme(theme);
```

This is mostly just a suggestion based on my preference. Doesn't really matter either way.

alacritty/src/event.rs
kchibisov: This is already logged upper.

kchibisov: Also, clippy is against your change(asks for `if let`), especially you take into account that we log in `config::reload` already, since it's used from multiple paths.

kchibisov:
>Before merging, would you mind sharing a picture (or more) of how these are supposed to look? Would probably be good for documentation anyways.

bitmap

Current CSD decorations looks like:
bitmap(1)

alacritty/src/event.rs
chrisduerr: As discussed in IRC, this should still be done.

kchibisov: was done.

PR #3461, opened 2020-03-14
Add Default Keybinding for SpawnNewInstance on macOS

nixpulvis:
By default <kbd>Command+N</kbd> should open a new terminal. In addition to macOS, I'm wondering if we should add this binding on Windows too. I think on other systems it's best to avoid a default binding for this.

I'll also note here, that there seems to be a bug when running `SpawnNewInstance` on a version of alacritty outside the `PATH`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Live config reload for `window.title`
+- Default Command+N keybinding for SpawnNewInstance on macOS
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## 0.5.0-dev
 
+### Added
+
+- Default Command+N keybinding for SpawnNewInstance on macOS
+
 ## 0.4.2-dev
 
 ### Packaging
@@ -17,7 +21,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Live config reload for `window.title`
-- Default Command+N keybinding for SpawnNewInstance on macOS
 
 ### Changed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -527,6 +527,7 @@
   #- { key: M,      mods: Command,         action: Minimize         }
   #- { key: Q,      mods: Command,         action: Quit             }
   #- { key: W,      mods: Command,         action: Quit             }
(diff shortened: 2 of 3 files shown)


kchibisov:
I don't like this idea, since the expected behavior for `Command + N` is to open a new window, however we're creating a new instance of the app and spamming system with it, so I'm against of this change. Ideally we need #607 and then bind it to `Command + N`, as you also may notice, we don't have this binding by default for the mentioned reason of not being what it should be in general, I guess.

chrisduerr:
What's the difference between Cmd+N before and after #607? Seems like it's the same, behavior but just the window handling is different?

kchibisov:
Yes, that's the thing, and you won't spam application dock.

chrisduerr:
> Yes, that's the thing, and you won't spam application dock.

Yeah, but that's just what happens right now when new Alacritty applications are launched. So that's just what's supposed to happen and the bug is separate. So I'm not against adding this on macOS, assuming that other applications use the same.

@nixpulvis Are you aware of any application on Windows that supports Super+N?

kchibisov:
>Yeah, but that's just what happens right now when new Alacritty applications are launched. So that's just what's supposed to happen and the bug is separate. So I'm not against adding this on macOS, assuming that other applications use the same.

You should spawn a new window and not a new application though, I'm also not completely against it, but just saying that this is not expected behavior on macOS to launch a new instance and spam dock.

chrisduerr:
> You should spawn a new window and not a new application though, I'm also not completely against it, but just saying that this is not expected behavior on macOS to launch a new instance and spam dock.

Yeah, but fundamentally that is completely unrelated to this PR. That's just a separate issue.

nixpulvis:
Well it may not be completly unrelatedly if we decide to keep `SpawnNewInstance` once we have `OpenNewWindow`, but that discussion can happen later. @kchibisov is right, the default for <kbd>Command+N</kbd> should open a new window not spam the process tree.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Live config reload for `window.title`
+- Default Command+N keybinding for SpawnNewInstance on macOS
```
chrisduerr: Needs to be moved to 0.5.0

nixpulvis:
Waiting on CI, then merging. (not with the stupid green button LOL).

nixpulvis:
> Are you aware of any application on Windows that supports Super+N?

@chrisduerr I tested the default Command Prompt application, and it didn't respond to <kbd>Ctrl+N</kbd> so I'm happy leaving it off. I'm sure there are applications that do this, or maybe just open tabs, but we should cross that bridge later.

PR #3472, opened 2020-03-16
don't recreate std::fs::read_to_string()

matthiaskrgr:


alacritty/src/renderer/mod.rs
```diff
@@ -1321,7 +1320,7 @@ fn create_shader(
     let source = if let Some(src) = source {
         src
     } else {
-        from_disk = read_file(path)?;
+        from_disk = std::fs::read_to_string(path)?;
```
chrisduerr: Should import `std::fs` and then just use `fs::read_to_string` here.

matthiaskrgr: Done

PR #3479, opened 2020-03-18
Fix Binding `mode`/`notmode` Matching Logic

nixpulvis:
TODO:

- [ ] Fill in more details to commit
- [ ] Add test(s)
- [ ] Changelog entry

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -1063,6 +1063,28 @@ mod tests {
     }
 
     #[test]
+    fn binding_matches_when_only_notmode_matches() {
+        let mut b1 = MockBinding::default();
+        b1.notmode = TermMode::ALT_SCREEN;
+        b1.mode = TermMode::VI;
+        let mut b2 = MockBinding::default();
+        b2.notmode = TermMode::ALT_SCREEN;
+
+        assert!(b1.triggers_match(&b2));
+    }
+
+    #[test]
+    fn binding_matches_when_only_mode_matches() {
+        let mut b1 = MockBinding::default();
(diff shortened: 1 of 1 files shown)


chrisduerr:
Besides the problems we've discussed on IRC like your algorithm not being symmetric and it not considering combinations of mode and notmode, I still think that even if we go for a simpler solution we should first provide a proper implementation.

If the proper implementation doesn't satisfy us, we can still go for the simpler approach. But I don't really want to just give up because I'm lazy.

I think the following code might work:

```
let selfmode = if self.mode.is_empty() {
    TermMode::ANY
} else {
    self.mode
};

let bindingmode = if binding.mode.is_empty() {
    TermMode::ANY
} else {
    self.mode
};

// You cannot make something intersect by taking away modes with `notmode`, unless modes intersect already
if !selfmode.intersects(bindingmode) {
    return false;
}

// If the binding requires a mode where the other binding is never bound, they are not in conflict
if selfmode.intersects(binding.notmode) || bindingmode.intersects(selfmode.notmode) {
    return false;
}

return true
```

It's mostly pseudocode and I haven't run many examples against it yet, but I think logically it should make sense? I'd be interested in what you think in general about this approach before implementing it in code and running a bunch of example tests against it.

Intuitively this *seems* to be the right thing to do, but it's not the first time I've thought that. There's a lot of things not intuitive about that. But I think the key difference here is that we cross-check `mode` of `self` with `notmode` of the `binding`. Just comparing `notmode` to `notmode` cannot possibly give us any good results, since it's just taking things away from `mode` and isn't functional on its own.

kchibisov:
Closing in favor of #3847

PR #3483, opened 2020-03-18
Add CopyPrimary action on Linux/BSD

kchibisov:
This action copies into primary selection buffer.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -624,7 +624,7 @@
   #- { key: K,      mods: Command,            action: ClearHistory     }
   #- { key: V,      mods: Command,            action: Paste            }
   #- { key: C,      mods: Command,            action: Copy             }
-  #- { key: C,      mods: Command, mode: Vi   action: ClearSelection   }
+  #- { key: C,      mods: Command, mode: Vi,  action: ClearSelection   }
   #- { key: H,      mods: Command,            action: Hide             }
   #- { key: M,      mods: Command,            action: Minimize         }
   #- { key: Q,      mods: Command,            action: Quit             }
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -505,7 +505,7 @@
 #   - ToggleSimpleFullscreen: Enters fullscreen without occupying another space
 #
 #   (Linux/BSD only):
-#   - CopyPrimary: Copies into primary selection buffer
+#   - CopySelection: Copies into selection buffer
 #
 # - `command`: Fork and execute a specified command plus arguments
 #
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -550,7 +550,6 @@
 #key_bindings:
   #- { key: Paste,                                action: Paste          }
   #- { key: Copy,                                 action: Copy           }
-  #- { key: Copy,                     mode: Vi,   action: ClearSelection }
   #- { key: L,         mods: Control,             action: ClearLogNotice }
   #- { key: L,         mods: Control, mode: ~Vi,  chars: "\x0c"          }
   #- { key: PageUp,    mods: Shift,   mode: ~Alt, action: ScrollPageUp,  }
@@ -574,6 +573,7 @@
   #- { key: D,      mods: Control,       mode: Vi, action: ScrollHalfPageDown      }
   #- { key: Y,                           mode: Vi, action: Copy                    }
   #- { key: Y,                           mode: Vi, action: ClearSelection          }
+  #- { key: Copy,                        mode: Vi, action: ClearSelection          }
   #- { key: V,                           mode: Vi, action: ToggleNormalSelection   }
   #- { key: V,      mods: Shift,         mode: Vi, action: ToggleLineSelection     }
   #- { key: V,      mods: Control,       mode: Vi, action: ToggleBlockSelection    }
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -550,6 +550,7 @@
 #key_bindings:
   #- { key: Paste,                                action: Paste          }
   #- { key: Copy,                                 action: Copy           }
+  #- { key: Copy,                     mode: Vi,   action: ClearSelection }
   #- { key: L,         mods: Control,             action: ClearLogNotice }
   #- { key: L,         mods: Control, mode: ~Vi,  chars: "\x0c"          }
   #- { key: PageUp,    mods: Shift,   mode: ~Alt, action: ScrollPageUp,  }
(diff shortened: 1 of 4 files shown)


alacritty/src/input.rs
```diff
@@ -140,6 +140,15 @@ impl<T: EventListener> Execute<T> for Action {
...
                 }
             },
+            #[cfg(not(any(target_os = "macos", windows)))]
+            Action::CopyPrimary => {
+                ctx.copy_selection(ClipboardType::Selection);
+
+                // Clear selection in vi mode for better user feedback
+                if ctx.terminal().mode().contains(TermMode::VI) {
+                    ctx.clear_selection();
+                }
```
nixpulvis: Small thing, but can we remove the duplication between this block and the `Copy` block above?

kchibisov: I don't think so, on other platforms you'll just create separate function to do this, since binding is purely for Linux/BSD. You also can't combine `match` arms due to this, so I'd like to keep things as it is right now.

kchibisov:
>This also raises a question for me, which I'm sure you've thought about, and is probably the motivation behind this PR in some way. But shouldn't we be copying to the primary buffer automatically while in VI mode? Or is it purely a default mouse interaction? I can see arguments both ways. With this addition at least I could add an action for it. But now if I want unified clipboards (like we already support for mouse selections), I need to create two bindings, one for Copy and one for CopyPrimary. Basically, a user trying to configure their clipboard is going to stumble around with the way things are currently.

Automatic copy wasn't implemented intentionally. First of all, it's expected to auto copy on mouse selection, however on keyboard selection there's no such rule, so I'd be very surprised if it will change my primary buffer. Some users just want to use `Primary` buffer instead of default one, so that was the motivation for this.

>Basically, a user trying to configure their clipboard is going to stumble around with the way things are currently.

I don't think so, the main goal is to use different buffer, also having 2 bindings is easier than some implicit heuristic logic.

>Would it make any sense to create two new actions, totally three actions for copying?

    Copy, redefined to copy to both clipboard and primary
    CopyClipboard, as today's Copy
    CopyPrimary, as defined in this PR

Thoughts?

If you want to copy to both clipboards at the same time, you don't need new binding actually, however I feel like we can disable clear selection in code and add a binding for it(just to user config, since we have `ClearSelection`). What do you think @chrisduerr ?

chrisduerr:
> I feel like we can disable clear selection in code and add a binding for it

I'm not sure what you mean?

> Would it make any sense to create two new actions, totally three actions for copying?

No. Fighting against platform defaults doesn't sound like a great idea. If you dislike the behavior of copying to the selection clipboard so much, disabling it in applications is probably not the right way to go. You just can't enforce that in every application. So I don't see any need for it in Alacritty.

nixpulvis:
My main concern is that the keybinding configuration and the selection configurations (with and without proposed addition) would need to configured separately. I suppose this is a small thorn, but would be nice to avoid if we can.

kchibisov:
>I suppose this is a small thorn, but would be nice to avoid if we can.

users may prefer to not clear their selection after copy, so I don't see any need to force their combination.

kchibisov:
I've renamed proposed action from `CopyPrimary` to `CopySelection`, which should follow existing binding `PasteSelection`.

alacritty.yml
```diff
@@ -547,6 +550,7 @@
 #key_bindings:
   #- { key: Paste,                                action: Paste          }
   #- { key: Copy,                                 action: Copy           }
+  #- { key: Copy,                     mode: Vi,   action: ClearSelection }
```
chrisduerr: I'd like to keep the vi mode bindings all together.

alacritty.yml
```diff
@@ -597,14 +602,15 @@
...
-  #- { key: C,        mods: Control|Shift, action: Copy             }
-  #- { key: Insert,   mods: Shift,         action: PasteSelection   }
-  #- { key: Key0,     mods: Control,       action: ResetFontSize    }
-  #- { key: Equals,   mods: Control,       action: IncreaseFontSize }
-  #- { key: Add,      mods: Control,       action: IncreaseFontSize }
-  #- { key: Subtract, mods: Control,       action: DecreaseFontSize }
-  #- { key: Minus,    mods: Control,       action: DecreaseFontSize }
+  #- { key: V,        mods: Control|Shift,           action: Paste            }
+  #- { key: C,        mods: Control|Shift,           action: Copy             }
+  #- { key: C,        mods: Control|Shift, mode: Vi, action: ClearSelection   }
```
chrisduerr: I'd like to keep the vi mode bindings all together.

alacritty.yml
```diff
@@ -618,6 +624,7 @@
   #- { key: K,      mods: Command,            action: ClearHistory     }
   #- { key: V,      mods: Command,            action: Paste            }
   #- { key: C,      mods: Command,            action: Copy             }
+  #- { key: C,      mods: Command, mode: Vi,  action: ClearSelection   }
```
chrisduerr: I'd like to keep the vi mode bindings all together.

alacritty/src/input.rs
```diff
@@ -134,11 +134,10 @@ impl<T: EventListener> Execute<T> for Action {
             },
             Action::Copy => {
                 ctx.copy_selection(ClipboardType::Clipboard);
```
chrisduerr: Can we inline this to be just a single line?

alacritty/src/config/bindings.rs
```diff
@@ -111,6 +111,10 @@ pub enum Action {
     /// Store current selection into clipboard.
     Copy,
 
+    #[cfg(not(any(target_os = "macos", windows)))]
+    /// Store current selection into selection buffer.
+    CopySelection,
+
     /// Paste contents of selection buffer.
     PasteSelection,
```
nixpulvis: Why is `CopySelection` not on macos or windows, while `PasteSelection` is?

chrisduerr: Historical reasons. `PasteSelection` has no reason to exist on macOS or Windows.

alacritty/src/config/bindings.rs
```diff
@@ -297,6 +301,7 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
     let mut bindings = bindings!(
         KeyBinding;
         Copy;  Action::Copy;
+        Copy,  +TermMode::VI; Action::Copy;
```
chrisduerr: Why are there two bindings for Copy here? A binding for all modes is always also active for vimode?

Did you mean to use `ClearSelection` as action here?

PR #3498, opened 2020-03-21
Clear Selection on ANSI Clear Line

nixpulvis:
This should fix selections staying around in applications like weechat, when the screen is changed with mouse actions, for example.

- [ ] Update `clear_screen`
  - [ ] Only clear selections that overlap with cleared lines
  - [ ] Don't clear selection when saved lines are cleared
  - [ ] Decide if we should rotate or clear the selection on `ClearMode::All` (VTE, for example rotates, however I think I'd prefer to clear)
- [ ] Update `clear_line`

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1930,11 +1930,10 @@ impl<T: EventListener> Handler for Term<T> {
         trace!("Clearing screen: {:?}", mode);
         let template = self.cursor.template;
 
-        // Remove active selections
-        self.grid.selection = None;
-
         match mode {
             ansi::ClearMode::Above => {
+                self.grid.selection = None;
+
                 // If clearing more than one line
                 if self.cursor.point.line > Line(1) {
                     // Fully clear all lines before the current line
@@ -1949,6 +1948,8 @@ impl<T: EventListener> Handler for Term<T> {
                 }
             },
             ansi::ClearMode::Below => {
+                self.grid.selection = None;
+
                 for cell in &mut self.grid[self.cursor.point.line][self.cursor.point.col..] {
                     cell.reset(&template);
                 }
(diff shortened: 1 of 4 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -1852,6 +1852,9 @@ impl<T: EventListener> Handler for Term<T> {
     fn clear_line(&mut self, mode: ansi::LineClearMode) {
         trace!("Clearing line: {:?}", mode);
 
+        // Remove active selections
+        self.grid.selection = None;
+
```
chrisduerr: Just to summarize my comment on IRC, I'd like to see more granularity here.

Clearing everything just because a single line somewhere was cleared seems unfortunate to me.

In our other clear function we could probably be a little more granular too, but that's usually only hit when the entire screen is cleared.

kchibisov: I agree, I don't like this "fix" either. For example it'll break tmux. We also have a lot of ways to clear content underneath the selection. In theory we can only clear selection in mouse mode when left clicking, since mouse mode app could have its own selection(tmux) and user can have both selections with the same color, so it could be very confusing.

chrisduerr:
> After thinking more about this, I feel like we should just reset selection on left mouse click in mouse mode too.

No, that's the completely wrong way to go about this.

Clicking with your mouse while in mouse mode has **NO** effect on selection. Those two events are **COMPLETELY** separated.

kchibisov:
It seems like at least xterm doesn't reset selection on left click in mouse mode, it's trying to do the thing @nixpulvis proposed. So you can in theory try to intersect selection box and 'rects' from clear/erase, etc. I don't really like it since it's too much complexity all over the place.

chrisduerr:
> So you can in theory try to intersect selection box and 'rects' from clear/erase, etc. I don't really like it since it's too much complexity all over the place.

Why? It makes sense to just clear it when it's dirty, right? We should have the abstractions in place to make this basically a single line of code.

kchibisov:
Closing in favor #3854

PR #3501, opened 2020-03-23
Add FreeBSD to CI

### LINKED ISSUE #2248: Test BSD on CI

kchibisov:
Fixes #2248.

### DIFF
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -17,14 +17,5 @@ tasks:
       cd alacritty
       $HOME/.cargo/bin/cargo +1.39.0 test
   - stable: |
-      $HOME/.cargo/bin/rustup toolchain install --profile minimal stable
       cd alacritty
       $HOME/.cargo/bin/cargo +stable test
-  - stable-clippy: |
-      $HOME/.cargo/bin/rustup component add --toolchain stable clippy || true
-      cd alacritty || true
-      $HOME/.cargo/bin/cargo +stable clippy || true
-  - nightly: |
-      $HOME/.cargo/bin/rustup toolchain install --profile minimal nightly || true
-      cd alacritty || true
-      $HOME/.cargo/bin/cargo +nightly test || true
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -11,7 +11,7 @@ sources:
   - https://github.com/alacritty/alacritty
 tasks:
   - rustup: |
-      curl https://sh.rustup.rs -sSf | sh -s -- -y
+      curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain stable --profile minimal
   - 1-39-0: |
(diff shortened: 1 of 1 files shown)


ddevault:
FYI, you can test out new CI configs by just dropping them into builds.sr.ht/submit - no need to push commits.

kchibisov:
The problem wasn't a config, the problem was that CI picked up wrong source code somehow.
The code mentioned in this crash doesn't even exists https://builds.sr.ht/~undeadleech/job/173487

ddevault:
Hm, take the trailing / off of the source URL maybe. It should be rewriting the source URL to check out the specific commit which was pushed.

ddevault:
Nice, that looks right for the OpenBSD config. You should do the same for the FreeBSD config, though.

kchibisov:
>Nice, that looks right for the OpenBSD config. You should do the same for the FreeBSD config, though.

I wonder why they matter, I've noticed similar behavior on sourcehut tbh. The problem is that some browsers are automatically insert them, so I've copy pasted both URLs with trailing '/' automatically.

ddevault:
We just use the basename of the repo URL to identify which one to rewrite. Remember that the URL your browser renders is not the same as the git clone URL, even if they're similar. If you check the "clone or download" widget on the repo's home page, you get this canonical clone URL: `https://github.com/alacritty/alacritty.git`

I agree that it's less than ideal, but that's the rationale, fwiw. It might be improved upon at some point.

.builds/freebsd.yml
```diff
@@ -0,0 +1,30 @@
...
+      cd alacritty
+      $HOME/.cargo/bin/cargo +1.39.0 test
+  - stable: |
+      $HOME/.cargo/bin/rustup toolchain install --profile minimal stable
+      cd alacritty
+      $HOME/.cargo/bin/cargo +stable test
+  - stable-clippy: |
+      $HOME/.cargo/bin/rustup component add --toolchain stable clippy || true
+      cd alacritty || true
+      $HOME/.cargo/bin/cargo +stable clippy || true
```
chrisduerr: Is the `|| true` to make this stage 'optional'?

I feel like at that point this stage is entirely useless, because we get no feedback at all about its success or failure.

kchibisov: Yeah, it doesn't support 'Allowed failures' feature. We can remove 'nightly' and 'clippy' for now, I guess.

PR #3507, opened 2020-03-24
Add regex scrollback buffer search

### LINKED ISSUE #1017: Scrollback Buffer Search

### LINKED ISSUE #3672: Allow changing inversed cursor to a fixed colored cursor
Right now there's no way to set fixed colored cursor with escape sequence from inverted cursor, so it should be supported.

chrisduerr:
This adds initial support for regex scrollback buffer search. While
searching should already be fully functional, the match is currently
only logged to stdout and not displayed in Alacritty.

However the searching should already behave as intended.

Fixes #3672.
Fixes #1017.

There are a lot of edgecases with this PR that need to be tested, I went through all of them but I doubt I've tested all permutations. So here's a list with things that would be useful to test:
 - Search from vi mode starting in last line
 - Search from vi mode above last line
 - Search without vi mode
 - Forward search
 - Backwards search
 - Goto match next start/end, previous start/end
 - Resize during search
 - Message bar
 - Start search while scrolled up
 - Search without history
 - Search with history and normal cursor at bottom
 - Search with history and normal cursor not at bottom (ctrl+l)
 - Immediate search while typing with match < 1000 lines away
 - Delayed search while typing with match > 1000 lines away

There are a few features I've intentionally left out that we should likely add in the future. I'll create issues for them as soon as this PR is merged:
 - Moving the cursor inside the search bar
 - Anchored searches using ^ and $

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Opt-out compilation flag `winpty` to disable WinPTY support
 - Scrolling during selection when mouse is at top/bottom of window
 - Expanding existing selections using the right mouse button
+- Terminal search
 
 ### Changed
 
@@ -58,6 +59,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Linewrap tracking when switching between primary and alternate screen buffer
 - Preservation of the alternate screen's saved cursor when swapping to primary screen and back
 - Reflow of cursor during resize
+- Cursor color escape ignored when its color is set to inverted in the config
 
 ## 0.4.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Default Command+N keybinding for SpawnNewInstance on macOS
-- Vi mode for copying text and opening links
+- Vi mode for regex search, copying text, and opening links
 - `CopySelection` action which copies into selection buffer on Linux/BSD
 - Option `cursor.thickness` to set terminal cursor thickness
 - Font fallback on Windows
@@ -27,7 +27,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Opt-out compilation flag `winpty` to disable WinPTY support
 - Scrolling during selection when mouse is at top/bottom of window
 - Expanding existing selections using the right mouse button
-- Vi mode terminal regex search
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Opt-out compilation flag `winpty` to disable WinPTY support
 - Scrolling during selection when mouse is at top/bottom of window
 - Expanding existing selections using the right mouse button
-- Terminal search
+- Vi mode terminal regex search
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dragging files into terminal now adds a space after each path
 - Default binding replacement conditions
 - Adjusted selection clearing granularity to more accurately match content
+- To use the cell's text color for selection with a modified background, the `color.selection.text`
+    variable must now be set to `CellForeground` instead of omitting it
 
 ### Fixed
 
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -453,19 +453,22 @@ impl Display {
         let size_info = self.size_info;
 
         let selection = !terminal.selection.as_ref().map(Selection::is_empty).unwrap_or(true);
-        let mouse_mode =
-            terminal.mode().intersects(TermMode::MOUSE_MODE) && !terminal.mode().contains(TermMode::VI);
+        let mouse_mode = terminal.mode().intersects(TermMode::MOUSE_MODE)
+            && !terminal.mode().contains(TermMode::VI);
 
-        let vi_mode_cursor =
-            if terminal.mode().contains(TermMode::VI) { Some(terminal.vi_mode_cursor) } else { None };
+        let vi_mode_cursor = if terminal.mode().contains(TermMode::VI) {
+            Some(terminal.vi_mode_cursor)
+        } else {
+            None
+        };
 
         // Update IME position.
         #[cfg(not(windows))]
         {
             let point = match &search_regex {
                 Some(regex) => {
-                    let column = min(regex.len() + SEARCH_LABEL.len(), terminal.num_cols().0 - 1);
-                    Point::new(terminal.num_lines() - 1, Column(column))
+                    let column = min(regex.len() + SEARCH_LABEL.len(), terminal.cols().0 - 1);
+                    Point::new(terminal.screen_lines() - 1, Column(column))
                 },
                 None => terminal.grid().cursor.point,
             };
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -454,10 +454,10 @@ impl Display {
 
         let selection = !terminal.selection.as_ref().map(Selection::is_empty).unwrap_or(true);
         let mouse_mode =
-            terminal.mode.intersects(TermMode::MOUSE_MODE) && !terminal.mode.contains(TermMode::VI);
+            terminal.mode().intersects(TermMode::MOUSE_MODE) && !terminal.mode().contains(TermMode::VI);
 
         let vi_mode_cursor =
-            if terminal.mode.contains(TermMode::VI) { Some(terminal.vi_mode_cursor) } else { None };
+            if terminal.mode().contains(TermMode::VI) { Some(terminal.vi_mode_cursor) } else { None };
 
         // Update IME position.
         #[cfg(not(windows))]
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -568,7 +568,7 @@ impl Display {
         config: &Config,
         size_info: &SizeInfo,
         message_bar_lines: usize,
-        search_regex: Option<String>
+        search_regex: Option<String>,
     ) {
         let search_regex = match search_regex {
             Some(search_regex) => search_regex,
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -443,7 +443,7 @@ impl Display {
         config: &Config,
         mouse: &Mouse,
         mods: ModifiersState,
-        search_regex: Option<String>,
+        search_regex: Option<&String>,
     ) {
         let grid_cells: Vec<RenderableCell> = terminal.renderable_cells(config).collect();
         let visual_bell_intensity = terminal.visual_bell.intensity();
@@ -465,7 +465,7 @@ impl Display {
         // Update IME position.
         #[cfg(not(windows))]
         {
-            let point = match &search_regex {
+            let point = match search_regex {
                 Some(regex) => {
                     let column = min(regex.len() + SEARCH_LABEL.len(), terminal.cols().0 - 1);
                     Point::new(terminal.screen_lines() - 1, Column(column))
@@ -609,7 +609,7 @@ impl Display {
         config: &Config,
         size_info: &SizeInfo,
         message_bar_lines: usize,
-        search_regex: Option<String>,
+        search_regex: Option<&String>,
     ) {
         let search_regex = match search_regex {
             Some(search_regex) => search_regex,
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -446,6 +446,7 @@ impl Display {
         search_regex: Option<&String>,
     ) {
         let grid_cells: Vec<RenderableCell> = terminal.renderable_cells(config).collect();
+        let search_regex = search_regex.map(|regex| Self::format_search(&regex));
         let visual_bell_intensity = terminal.visual_bell.intensity();
         let background_color = terminal.background_color();
         let metrics = self.glyph_cache.font_metrics();
@@ -465,9 +466,9 @@ impl Display {
         // Update IME position.
         #[cfg(not(windows))]
         {
-            let point = match search_regex {
+            let point = match &search_regex {
                 Some(regex) => {
-                    let column = min(regex.len() + SEARCH_LABEL.len(), terminal.cols().0 - 1);
+                    let column = min(regex.len() + SEARCH_LABEL.len() - 1, terminal.cols().0 - 1);
                     Point::new(terminal.screen_lines() - 1, Column(column))
                 },
                 None => terminal.grid().cursor.point,
@@ -603,23 +604,8 @@ impl Display {
         }
     }
 
-    /// Draw current search regex.
-    fn draw_search(
-        &mut self,
-        config: &Config,
-        size_info: &SizeInfo,
-        message_bar_lines: usize,
-        search_regex: Option<&String>,
-    ) {
-        let search_regex = match search_regex {
-            Some(search_regex) => search_regex,
-            None => return,
-        };
-        let glyph_cache = &mut self.glyph_cache;
-
-        let label_len = SEARCH_LABEL.len();
-        let num_cols = size_info.cols().0;
-
+    /// Format search regex to account for the cursor and fullwidth characters.
+    fn format_search(search_regex: &str) -> String {
         // Add spacers for wide chars.
         let mut text = String::with_capacity(search_regex.len());
         for c in search_regex.chars() {
@@ -632,10 +618,30 @@ impl Display {
         // Add cursor to show whitespace.
         text.push('_');
 
+        text
+    }
+
+    /// Draw current search regex.
+    fn draw_search(
+        &mut self,
+        config: &Config,
+        size_info: &SizeInfo,
+        message_bar_lines: usize,
+        search_regex: Option<String>,
+    ) {
+        let search_regex = match search_regex {
+            Some(search_regex) => search_regex,
+            None => return,
+        };
+        let glyph_cache = &mut self.glyph_cache;
+
+        let label_len = SEARCH_LABEL.len();
+        let num_cols = size_info.cols().0;
+
         // Truncate beginning of text when it exceeds viewport width.
-        let text_len = text.len();
+        let text_len = search_regex.len();
         let truncate_len = min((text_len + label_len).saturating_sub(num_cols), text_len);
-        let text = &text[truncate_len..];
+        let text = &search_regex[truncate_len..];
 
         // Assure text length is at least num_cols.
         let padding_len = num_cols.saturating_sub(label_len);
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -34,7 +34,7 @@ use alacritty_terminal::selection::Selection;
 use alacritty_terminal::term::{RenderableCell, SizeInfo, Term, TermMode};
 
 use crate::config::Config;
-use crate::event::{DisplayUpdate, Mouse};
+use crate::event::Mouse;
 use crate::renderer::rects::{RenderLines, RenderRect};
 use crate::renderer::{self, GlyphCache, QuadRenderer};
 use crate::url::{Url, Urls};
@@ -100,6 +100,44 @@ impl From<renderer::Error> for Error {
 impl From<glutin::ContextError> for Error {
     fn from(val: glutin::ContextError) -> Self {
         Error::ContextError(val)
+    }
+}
+
+#[derive(Default, Clone, Debug, PartialEq)]
+pub struct DisplayUpdate {
+    pub dirty: bool,
+
+    dimensions: Option<PhysicalSize<u32>>,
+    font: Option<Font>,
+    cursor_dirty: bool,
+}
+
+impl DisplayUpdate {
+    pub fn dimensions(&self) -> Option<PhysicalSize<u32>> {
+        self.dimensions
+    }
+
+    pub fn font(&self) -> Option<&Font> {
+        self.font.as_ref()
+    }
+
+    pub fn cursor_dirty(&self) -> bool {
+        self.cursor_dirty
+    }
+
+    pub fn set_dimensions(&mut self, dimensions: PhysicalSize<u32>) {
+        self.dimensions = Some(dimensions);
+        self.dirty = true;
+    }
+
+    pub fn set_font(&mut self, font: Font) {
+        self.font = Some(font);
+        self.dirty = true;
+    }
+
+    pub fn set_cursor_dirty(&mut self) {
+        self.cursor_dirty = true;
+        self.dirty = true;
     }
 }
 
@@ -304,7 +342,7 @@ impl Display {
     }
 
     /// Update font size and cell dimensions.
-    fn update_glyph_cache(&mut self, config: &Config, font: Font) {
+    fn update_glyph_cache(&mut self, config: &Config, font: &Font) {
         let size_info = &mut self.size_info;
         let cache = &mut self.glyph_cache;
 
@@ -339,9 +377,9 @@ impl Display {
         T: EventListener,
     {
         // Update font size and cell dimensions.
-        if let Some(font) = update_pending.font {
+        if let Some(font) = update_pending.font() {
             self.update_glyph_cache(config, font);
-        } else if update_pending.cursor {
+        } else if update_pending.cursor_dirty() {
             self.clear_glyph_cache();
         }
 
@@ -353,7 +391,7 @@ impl Display {
         let mut padding_y = f32::from(config.window.padding.y) * self.size_info.dpr as f32;
 
         // Update the window dimensions.
-        if let Some(size) = update_pending.dimensions {
+        if let Some(size) = update_pending.dimensions() {
             // Ensure we have at least one column and row.
             self.size_info.width = (size.width as f32).max(cell_width + 2. * padding_x);
             self.size_info.height = (size.height as f32).max(cell_height + 2. * padding_y);
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -1,6 +1,7 @@
 //! The display subsystem including window management, font rasterization, and
 //! GPU drawing.
 
+use std::cmp::min;
 use std::f64;
 use std::fmt::{self, Formatter};
 #[cfg(not(any(target_os = "macos", windows)))]
@@ -15,6 +16,7 @@ use glutin::platform::unix::EventLoopWindowTargetExtUnix;
 use glutin::window::CursorIcon;
 use log::{debug, info};
 use parking_lot::MutexGuard;
+use unicode_width::UnicodeWidthChar;
 #[cfg(not(any(target_os = "macos", windows)))]
 use wayland_client::{Display as WaylandDisplay, EventQueue};
 
@@ -23,12 +25,12 @@ use font::set_font_smoothing;
 use font::{self, Rasterize};
 
 use alacritty_terminal::config::{Font, StartupMode};
-use alacritty_terminal::event::OnResize;
-use alacritty_terminal::index::Line;
+use alacritty_terminal::event::{EventListener, OnResize};
+use alacritty_terminal::grid::Dimensions;
+use alacritty_terminal::index::{Column, Line, Point};
 use alacritty_terminal::message_bar::MessageBuffer;
 use alacritty_terminal::meter::Meter;
 use alacritty_terminal::selection::Selection;
-use alacritty_terminal::term::color::Rgb;
 use alacritty_terminal::term::{RenderableCell, SizeInfo, Term, TermMode};
 
 use crate::config::Config;
@@ -37,6 +39,8 @@ use crate::renderer::rects::{RenderLines, RenderRect};
 use crate::renderer::{self, GlyphCache, QuadRenderer};
 use crate::url::{Url, Urls};
 use crate::window::{self, Window};
+
+const SEARCH_LABEL: &str = "Search: ";
 
 #[derive(Debug)]
 pub enum Error {
@@ -328,9 +332,12 @@ impl Display {
         terminal: &mut Term<T>,
         pty_resize_handle: &mut dyn OnResize,
         message_buffer: &MessageBuffer,
+        search_active: bool,
         config: &Config,
         update_pending: DisplayUpdate,
-    ) {
+    ) where
+        T: EventListener,
+    {
         // Update font size and cell dimensions.
         if let Some(font) = update_pending.font {
             self.update_glyph_cache(config, font);
@@ -369,6 +376,11 @@ impl Display {
             pty_size.height -= pty_size.cell_height * lines as f32;
         }
 
+        // Add an extra line for the current search regex.
+        if search_active {
+            pty_size.height -= pty_size.cell_height;
+        }
+
         // Resize PTY.
         pty_resize_handle.on_resize(&pty_size);
 
@@ -393,6 +405,7 @@ impl Display {
         config: &Config,
         mouse: &Mouse,
         mods: ModifiersState,
+        search_regex: Option<String>,
     ) {
         let grid_cells: Vec<RenderableCell> = terminal.renderable_cells(config).collect();
         let visual_bell_intensity = terminal.visual_bell.intensity();
@@ -402,18 +415,25 @@ impl Display {
         let size_info = self.size_info;
 
         let selection = !terminal.selection.as_ref().map(Selection::is_empty).unwrap_or(true);
-        let mouse_mode = terminal.mode().intersects(TermMode::MOUSE_MODE)
-            && !terminal.mode().contains(TermMode::VI);
+        let mouse_mode =
+            terminal.mode.intersects(TermMode::MOUSE_MODE) && !terminal.mode.contains(TermMode::VI);
 
-        let vi_mode_cursor = if terminal.mode().contains(TermMode::VI) {
-            Some(terminal.vi_mode_cursor)
-        } else {
-            None
-        };
+        let vi_mode_cursor =
+            if terminal.mode.contains(TermMode::VI) { Some(terminal.vi_mode_cursor) } else { None };
 
         // Update IME position.
         #[cfg(not(windows))]
-        self.window.update_ime_position(&terminal, &self.size_info);
+        {
+            let point = match &search_regex {
+                Some(regex) => {
+                    let len = min(regex.len() + SEARCH_LABEL.len(), terminal.num_cols().0);
+                    Point::new(terminal.num_lines() - 1, Column(len - 1))
+                },
+                None => terminal.grid().cursor.point,
+            };
+
+            self.window.update_ime_position(point, &self.size_info);
+        }
 
         // Drop terminal as early as possible to free lock.
         drop(terminal);
@@ -484,11 +504,13 @@ impl Display {
             rects.push(visual_bell_rect);
         }
 
+        let mut message_bar_lines = 0;
         if let Some(message) = message_buffer.message() {
             let text = message.text(&size_info);
+            message_bar_lines = text.len();
 
             // Create a new rectangle for the background.
-            let start_line = size_info.lines().0 - text.len();
+            let start_line = size_info.lines().0 - message_bar_lines;
             let y = size_info.cell_height.mul_add(start_line as f32, size_info.padding_y);
             let message_bar_rect =
                 RenderRect::new(0., y, size_info.width, size_info.height - y, message.color(), 1.);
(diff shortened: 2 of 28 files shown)


goferito:
Is there a way to search so far, even if it's not regex based? Or is every search feature in this PR?

chrisduerr:
> Is there a way to search so far

No.

kchibisov:
It seems like confirming search when in history (display_offset is something non zero) it moves one line down, which shouldn't happen imho, however such thing is when display_offset isn't 0.

kchibisov:
Some random oob when running `yes` on `100k` scrollback and typing `yess` in search prompt.

```
~ ➜ RUST_BACKTRACE=1 alacritty
thread 'main' panicked at 'index out of bounds: the len is 100065 but the index is 18446744073709379424', /home/kchibisov/src/rust/alacritty-workspace/repo/alacritty_terminal/src/term/mod.rs:1174:21
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1069
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1504
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:218
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:511
  11: rust_begin_unwind
             at src/libstd/panicking.rs:419
  12: core::panicking::panic_fmt
             at src/libcore/panicking.rs:111
  13: core::panicking::panic_bounds_check
             at src/libcore/panicking.rs:69
  14: <usize as core::slice::SliceIndex<[T]>>::index
             at ./.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/slice/mod.rs:2872
  15: core::slice::<impl core::ops::index::Index<I> for [T]>::index
             at ./.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/slice/mod.rs:2732
  16: <alloc::vec::Vec<T> as core::ops::index::Index<I>>::index
             at ./.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/liballoc/vec.rs:1945
  17: <alacritty_terminal::grid::storage::Storage<T> as core::ops::index::Index<usize>>::index
             at ./src/rust/alacritty-workspace/repo/alacritty_terminal/src/grid/storage.rs:271
  18: <alacritty_terminal::grid::Grid<T> as core::ops::index::Index<usize>>::index
             at ./src/rust/alacritty-workspace/repo/alacritty_terminal/src/grid/mod.rs:521
  19: alacritty_terminal::term::Term<T>::expand_wide
             at ./src/rust/alacritty-workspace/repo/alacritty_terminal/src/term/mod.rs:1174
  20: alacritty_terminal::term::search::<impl alacritty_terminal::term::Term<T>>::search_next
             at ./src/rust/alacritty-workspace/repo/alacritty_terminal/src/term/search.rs:73
  21: alacritty::event::ActionContext<N,T>::goto_match
             at alacritty/src/event.rs:519
  22: <alacritty::event::ActionContext<N,T> as alacritty::input::ActionContext<T>>::pop_search
             at alacritty/src/event.rs:450
  23: alacritty::input::Processor<T,A>::key_input
             at alacritty/src/input.rs:766
  24: alacritty::event::Processor<N>::handle_event
             at alacritty/src/event.rs:864
  25: alacritty::event::Processor<N>::run::{{closure}}
             at alacritty/src/event.rs:754
  26: winit::platform_impl::platform::sticky_exit_callback
             at ./.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.22.2/src/platform_impl/linux/mod.rs:698
  27: winit::platform_impl::platform::wayland::event_loop::EventLoop<T>::run_return
             at ./.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.22.2/src/platform_impl/linux/wayland/event_loop.rs:537
  28: winit::platform_impl::platform::EventLoop<T>::run_return
             at ./.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.22.2/src/platform_impl/linux/mod.rs:634
  29: <winit::event_loop::EventLoop<T> as winit::platform::desktop::EventLoopExtDesktop>::run_return
             at ./.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.22.2/src/platform/desktop.rs:51
  30: alacritty::event::Processor<N>::run
             at alacritty/src/event.rs:681
  31: alacritty::run
             at alacritty/src/main.rs:190
  32: alacritty::main
             at alacritty/src/main.rs:102
  33: std::rt::lang_start::{{closure}}
             at ./.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/rt.rs:67
  34: std::rt::lang_start_internal::{{closure}}::{{closure}}
             at src/libstd/rt.rs:52
```

chrisduerr:
> Some random oob when running yes on 100k scrollback and typing yess in search prompt.

Fixed, thanks!

nicklan:
This works great in my testing so far. Thanks so much, this will be a huge usability win.

A few bits of feedback:
1. It's not clear at all that hitting Enter puts you into Vi mode and that only from there can you navigate between search results
2. On that note, for those of us who won't use Vi mode, would it make sense to have rather a "Search Mode" that just enables the Next/Previous match? This means that there doesn't need to be an Enter/Leave ViMode keybinding for people who don't want it
3. Alternately or Additionally: why force next/previous match to only be available in ViMode? Could it not just be a keybinding that does nothing if you're not in the middle of a search?
4. There's no indication that the search has looped. Might be nice to echo something where the search bar or similar

chrisduerr:
> It's not clear at all that hitting Enter puts you into Vi mode and that only from there can you navigate between search results

Determining the best way to show that vi mode is active is still a bit of an outstanding question.

> On that note, for those of us who won't use Vi mode, would it make sense to have rather a "Search Mode" that just enables the Next/Previous match? This means that there doesn't need to be an Enter/Leave ViMode keybinding for people who don't want it

We've decided against supporting search completely without vi mode, especially since the basic bindings to navigate through search results are very simple.

> Alternately or Additionally: why force next/previous match to only be available in ViMode? Could it not just be a keybinding that does nothing if you're not in the middle of a search?

There is no concept of 'current match' without the vi cursor. It's purely a way to move the cursor. Moving the vi cursor outside of vi mode doesn't do anything, so there's no reason to allow it.

> There's no indication that the search has looped. Might be nice to echo something where the search bar or similar

Do you mean during search, or while jumping between matches? Usually showing any kind of additional information is quite annoying since it shrinks the terminal size.

nicklan:
> Determining the best way to show that vi mode is active is still a bit of an outstanding question.

Okay, that makes sense. 
 
> We've decided against supporting search completely without vi mode, especially since the basic bindings to navigate through search results are very simple.
> There is no concept of 'current match' without the vi cursor. It's purely a way to move the cursor. Moving the vi cursor outside of vi mode doesn't do anything, so there's no reason to allow it.

That's one way to do it yeah, but it's not inherent to search. Many/most terminals do not enter a special ViMode for search. Usually it's just that you can search and have another binding for next/prev that is a no-op if a search isn't active. Not to say that this is terrible, but it will be confusing to people migrating to alacritty.
 
> > There's no indication that the search has looped. Might be nice to echo something where the search bar or similar
> 
> Do you mean during search, or while jumping between matches? Usually showing any kind of additional information is quite annoying since it shrinks the terminal size.

I mean when I hit (say) `N` and I'm on the last match, it jumps to the first match. This is the right thing to do, but it's helpful to know you've looped (vim for instance echos "search hit BOTTOM, continuing at TOP" in the display line at the bottom). I don't think it would need to appear except for when looping in one direction or another occurred, so hopefully screen real-estate wouldn't be too much of an issue for that.

chrisduerr:
> That's one way to do it yeah, but it's not inherent to search. Many/most terminals do not enter a special ViMode for search. Usually it's just that you can search and have another binding for next/prev that is a no-op if a search isn't active. Not to say that this is terrible, but it will be confusing to people migrating to alacritty.

That's a good point and it's the reason why I have considered implementing a separate search mode one could navigate without ever having to touch vi mode.

However this would require adding the concept of an "active match"  and then add special colors to highlight that special match. We'd also probably have to figure out a good way to copy the current active match, likely adding another binding just for that.

If Alacritty was a terminal emulator that shipped by default on any distro, that would probably be the obvious choice. But since it's a little more power user focused I think it's worth it to at least ship only the current version in the initial release and see how people like it. Especially because it forces people to get familiar with vi mode, automatically introducing that feature too.

However if the user is just trapped in vi mode without knowing what's up, chances are that'll just cause pointless bug reports, so figuring out how to best indicate it is still an important topic.

> I mean when I hit (say) N and I'm on the last match, it jumps to the first match. This is the right thing to do, but it's helpful to know you've looped (vim for instance echos "search hit BOTTOM, continuing at TOP" in the display line at the bottom). I don't think it would need to appear except for when looping in one direction or another occurred, so hopefully screen real-estate wouldn't be too much of an issue for that.

See, that's pretty much a scenario where it would be impossible for us to show a message like the search bar. Since if we shrink the terminal, everything inside the terminal is also shrunk creating a really messy situation of things just jumping around randomly. Not a very pleasant experience for something that just pops up shortly.

nicklan:
> If Alacritty was a terminal emulator that shipped by default on any distro, that would probably be the obvious choice. But since it's a little more power user focused I think it's worth it to at least ship only the current version in the initial release and see how people like it. Especially because it forces people to get familiar with vi mode, automatically introducing that feature too.

Fully agree that it's worth merging/releasing in the current form. Maybe I'm the only one bothered by this :)
 
> > I mean when I hit (say) N and I'm on the last match, it jumps to the first match. This is the right thing to do, but it's helpful to know you've looped (vim for instance echos "search hit BOTTOM, continuing at TOP" in the display line at the bottom). I don't think it would need to appear except for when looping in one direction or another occurred, so hopefully screen real-estate wouldn't be too much of an issue for that.
> 
> See, that's pretty much a scenario where it would be impossible for us to show a message like the search bar. Since if we shrink the terminal, everything inside the terminal is also shrunk creating a really messy situation of things just jumping around randomly. Not a very pleasant experience for something that just pops up shortly.

Could the message not just overwrite whatever is at the bottom for like 1 second? (perhaps the search scrolling would have to be clever enough to move the match to the second-to-last line) Without knowing the internals of how this works I realize that "overwrite the line for a second" may be really tricky to do correctly.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Opt-out compilation flag `winpty` to disable WinPTY support
 - Scrolling during selection when mouse is at top/bottom of window
 - Expanding existing selections using the right mouse button
+- Terminal search
```
kchibisov, line 29: In general, I feel like being a bit more verbose here could be better.

alacritty/src/display.rs
```diff
@@ -546,6 +562,68 @@ impl Display {
...
+        self.renderer.with_api(&config, &size_info, |mut api| {
+            api.render_string(glyph_cache, line, &text, fg, Some(bg));
+        });
+    }
+
+    /// Draw render timer.
+    fn draw_render_timer(&mut self, config: &Config, size_info: &SizeInfo) {
+        if !config.render_timer() {
+            return;
+        }
```
kchibisov, line 615: I really wonder where such check should be, in the caller or inside of function. I feel like checking in caller is better, since the function is called `draw_render_timer`, so it implies that it draws it. And you check in caller whether you should draw it or not, however it looks kind of fine, since you try to be more declarative.

alacritty/src/event.rs
```diff
@@ -74,17 +81,49 @@ impl From<TerminalEvent> for Event {
...
+impl Default for SearchState {
+    fn default() -> Self {
+        Self {
+            direction: Direction::Right,
+            display_offset_delta: 0,
+            vi_cursor_point: Point::default(),
+            regex: None,
+        }
+    }
+}
```
kchibisov, line 114: I kind of feel like those things shouldn't be in `event.rs`, as well as some other search related stuff, the file is already kind of huge.

alacritty/src/event.rs
```diff
@@ -74,17 +81,49 @@ impl From<TerminalEvent> for Event {
...
+
 #[derive(Default, Clone, Debug, PartialEq)]
 pub struct DisplayUpdate {
     pub dimensions: Option<PhysicalSize<u32>>,
-    pub message_buffer: bool,
     pub font: Option<Font>,
     pub cursor: bool,
+    pub dirty: bool,
```
kchibisov, line 121: Could you rename it to something else, since it's really confusing with `terminal.dirty` being used around.

alacritty_terminal/src/term/cell.rs
```diff
@@ -12,18 +12,19 @@ pub const MAX_ZEROWIDTH_CHARS: usize = 5;
...
+        const BOLD_ITALIC               = 0b000_0000_0110;
+        const UNDERLINE                 = 0b000_0000_1000;
+        const WRAPLINE                  = 0b000_0001_0000;
+        const WIDE_CHAR                 = 0b000_0010_0000;
+        const WIDE_CHAR_SPACER          = 0b000_0100_0000;
+        const DIM                       = 0b000_1000_0000;
+        const DIM_BOLD                  = 0b000_1000_0010;
+        const HIDDEN                    = 0b001_0000_0000;
+        const STRIKEOUT                 = 0b010_0000_0000;
+        const LEADING_WIDE_CHAR_SPACER  = 0b100_0000_0000;
```
kchibisov, line 27: I'd kind of like having all WIDE_CHAR things grouped here tbqh.

alacritty_terminal/src/term/color.rs
```diff
@@ -128,6 +120,66 @@ impl FromStr for Rgb {
     }
 }
 
+/// RGB color optionally referencing the cell's foreground or background.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum CellRgb {
```
kchibisov, line 125: I'm not sure that it should be called `CellRgb`, why not `CellColor`?

alacritty_terminal/src/term/search.rs
```diff
@@ -0,0 +1,794 @@
...
+
+use regex_automata::{dense, DenseDFA, Error as RegexError, DFA};
+
+use crate::grid::{BidirectionalIterator, Dimensions, GridIterator};
+use crate::index::{Boundary, Column, Direction, Point, Side};
+use crate::term::cell::{Cell, Flags};
+use crate::term::Term;
+
+/// Used to match equal brackets, when performing a bracket-pair selection.
+const BRACKET_PAIRS: [(char, char); 4] = [('(', ')'), ('[', ']'), ('{', '}'), ('<', '>')];
```
kchibisov, line 13: I really wonder whether we can add `"` and `'` and maybe some other semantic escape chars here as well? Like I don't why not. Probably offtopic here.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1050,17 +972,16 @@ impl<T> Term<T> {
...
-    }
+        // Invalidate selection and tabs only when necessary.
+        if old_cols != num_cols {
+            self.selection = None;
 
-    #[inline]
-    pub fn mode(&self) -> &TermMode {
-        &self.mode
+            // Recreate tabs list.
+            self.tabs.resize(self.num_cols());
```
kchibisov, line 983: I guess here you can use clear PR log to perform invalidation properly? If the goal is to keep selection after exiting search, you may want to compare new and old grid len, I guess it should work when columns are the same?

alacritty/src/display.rs
chrisduerr: My main goal was just to slim out the `draw` function, which is why I decided to move that here. So it's just one line instead of three.

Though I do agree with your points, I've considered the same thing when I wrote it.

CHANGELOG.md
chrisduerr: Any concrete suggestions?

alacritty/src/event.rs
chrisduerr: Where else would you put it? It's event state.

alacritty/src/event.rs
chrisduerr: How so? I've explicitly named it *because* we already have `terminal.dirty`. The `display_update.dirty` is supposed to act exactly the same way, just for display updates. So as far as I can see it, it's just a common way to track generic demand for updates.

alacritty_terminal/src/term/cell.rs
chrisduerr: Grouped in what way? I've moved them together originally but that requires re-recording all our tests.

kchibisov: Oh, I see. then I guess we should keep it that way.

alacritty_terminal/src/term/color.rs
chrisduerr: The reason is that it's an `Rgb` color that optionally references the `Cell`. It's not actually a cell color. Quite the opposite, this can **never** be the cell color, since it references the cell itself which would be cyclical.

I see your issues with calling something `Rgb` that isn't only `Rgb` though, do you have a better suggestion?

alacritty_terminal/src/term/search.rs
chrisduerr: Nesting with quotes is a bit more complicated (read as impossible), but yeah this has nothing to do with this PR.

CHANGELOG.md
kchibisov: Well, it's a search for Vi mode, but it's a Terminal search on the other hand. And I'm not sure that `Scrollback buffer search` is a better name here, since it's not a scrollback buffer search, but generic search, since alt-screen doesn't have a scrollback.

alacritty_terminal/src/term/mod.rs
chrisduerr: I don't see what you mean by this, how would we use the clear PR to perform invalidation "properly"?

There is no invalidation going on when resizing, the content just moves. We cannot use invalidation on column change because the columns reflow and we do not know where from or where to and we do not need any kind of invalidation for line changes.

alacritty/src/event.rs
kchibisov: I mean it's strange that some of our updates are via `self.cursor() || self.font.is_some()` but the rest is `self.dirty`. So for me, it feels like cursor and font is a subset of dirty, when in reality the goal for dirty was a generic update, which doesn't fit into the rest of groups?

chrisduerr: Are you suggesting to always update dirty together with the other options?

alacritty_terminal/src/term/mod.rs
kchibisov: I mean that in some cases when you're changing the amount of visible lines, you may want to invalidate selectiion? Not the other way around. I'm talking about properly invalidating, when we have the same amount of columns before and after, but a bit different lines_total value, since sometimes you may wan to clear selection here.

alacritty/src/event.rs
kchibisov: You should either compute dirty based on all values from the display update or update dirty as well and check just for dirty when checking for update and deciding whether we should handle it.

alacritty_terminal/src/term/color.rs
kchibisov: It's still technically cell color, but how it's computed and the cell color the user see. I don't have better suggestions, but it's not really and Rgb, even a comment says that it's not only an Rgb.

alacritty_terminal/src/term/mod.rs
chrisduerr: > I mean that in some cases when you're changing the amount of visible lines, you may want to invalidate selectiion? 

Why would you ever want to do that? You're only changing the number of invisible lines, clearing any lines.

There is no range that is being invalidated.

alacritty/src/event.rs
kchibisov: it's not really an event state, it's a search state. it's even called `SearchState`, not `SearchEventState`, and event is something we're getting or sending, not something we persist on `ActionContext`. Though we already have `DisplayUpdate`, etc in that particular. Maybe we should keep it here then.

alacritty_terminal/src/term/color.rs
chrisduerr: It's not. It might be an underline cursor, or an I-Beam. It's absolutely not a cell color, but a color referencing a cell's color.

Also it absolutely always is an `Rgb` color to the consumer of this. Since you access the color using the `color` function which takes the foreground/background and spits out an `Rgb` value. It's just that it's either a pre-defined Rgb value, or a reference to an Rgb value stored somewhere else. It's still always fundamentally Rgb.

alacritty/src/event.rs
chrisduerr: Sure, let's always set dirty then.

alacritty_terminal/src/term/mod.rs
kchibisov: I mean than I don't really get an issue with this logic and how it's faulty? The only situation where I can think you may want to clear selection on resize, if the amount of lines was changed to a lower value, meaning that you must invalidate selection, knowing that you scrolled away a bunch of them, when your selection spans the entire grid. So technically selection became invalid, and so you could likely want to reset it as well.

In general, clear PR won't help much here, I guess. So that's why I've started talking about just invalidation. In general, selection should be `None` if it scrolled offscreen, so not in grid range anymore, otherwise I can expand from scrolled away selection, which is strange.

kchibisov:
I guess I'm getting OOB now when I try to move cursor down one line in alt screen app(neomutt).

[crash.txt](https://github.com/alacritty/alacritty/files/4839168/crash.txt)

alacritty/src/event.rs
chrisduerr: It's an event state, just like `Mouse` is. It's a state solely maintained for the event processor.

kchibisov:
Since search is actually changing some default combinations, like

```yaml
selection:
  background: '#303030'
```
which should be in a new semantics 

```yaml
selection:
  text: CellForeground
  background: '#303030'
```

to (due to explicit reference of `CellBackground`)

```yaml
selection:
  text: CellBackground
  background: '#303030'
```

We should explicitly mention in a changelog that config should be updated, if they want to restore old behavior.

chrisduerr:
@kchibisov I tried to put it all in a single sentence that still makes the changes clear. Let me know what you think.

CHANGELOG.md
chrisduerr: I've gone into a little more detail, hopefully that should make things a little more clear.

alacritty_terminal/src/term/mod.rs
```diff
@@ -205,21 +129,23 @@ pub struct RenderableCellsIter<'a, C> {
...
+        term: &'a Term<T>,
+        config: &'a Config<C>,
         selection: Option<SelectionRange>,
-    ) -> RenderableCellsIter<'b, C> {
+    ) -> RenderableCellsIter<'a, C> {
         let grid = &term.grid;
 
-        let inner = grid.display_iter();
+        let viewport_end = grid.display_offset();
+        let viewport_start = viewport_end + grid.num_lines().0 - 1;
```
kchibisov, line 148: Could you move those checks inside closure? there's no need to compute them without having selection.

alacritty_terminal/src/term/mod.rs
```diff
@@ -249,11 +173,27 @@ impl<'a, C> RenderableCellsIter<'a, C> {
...
+            inner: grid.display_iter(),
             selection: selection_range,
             config,
             colors: &term.colors,
+            search: RenderableSearch::new(term),
+        }
+    }
+
+    /// Check if point is part of a search match.
+    fn is_matched(&mut self, point: Point) -> bool {
```
kchibisov, line 185: The check function shouldn't mutate state under the hood. I'm not expecting function that checks wether x is true, to mutate the state. If anything the name should be updated. The return of `bool` as a side effect is likely 'ok'.

alacritty_terminal/src/term/mod.rs
```diff
@@ -331,39 +269,35 @@ pub struct RenderableCell {
...
-            } else {
-                // Invert cell fg and bg colors.
-                mem::swap(&mut fg_rgb, &mut bg_rgb);
+                fg_rgb = iter.colors[NamedColor::Background];
+                bg_rgb = iter.colors[NamedColor::Foreground];
             }
-
-            bg_alpha = 1.0;
+        } else if iter.is_matched(point) {
+            let selected_fg = iter.config.colors.search.matches.foreground.color(fg_rgb, bg_rgb);
```
kchibisov, line 294: Should it be called `matched_fg`?

alacritty_terminal/src/term/mod.rs
```diff
@@ -331,39 +269,35 @@ pub struct RenderableCell {
...
+            bg_rgb = iter.config.colors.search.matches.background.color(fg_rgb, bg_rgb);
+            fg_rgb = selected_fg;
         }
 
-        // Override selection text with config colors.
-        if let (true, Some(col)) = (selected, config.colors.selection.text) {
-            fg_rgb = col;
-        }
+        // Make background transparent if it matches the terminal background.
+        let bg_alpha = if bg_rgb == iter.colors[NamedColor::Background] { 0. } else { 1. };
```
kchibisov, line 300: This is incorrect, you shouldn't set alpha based on rgb value, you should set
it based on whether the color is actually NamedBackground (read old code here).
otherwise you'll make truecolor background transparent.

Also, please add a comment explaining it, since it's not happing at the
first time already, that this code was changed that way to compare rgb values.

alacritty_terminal/src/term/mod.rs
```diff
@@ -376,6 +310,12 @@ impl RenderableCell {
         }
     }
 
+    fn is_empty(&self) -> bool {
+        self.bg_alpha == 0.
+            && !self.flags.intersects(Flags::UNDERLINE | Flags::STRIKEOUT)
+            && self.inner == RenderableCellContent::Chars([' '; cell::MAX_ZEROWIDTH_CHARS + 1])
+    }
```
kchibisov, line 317: I'd assume that it was a performance optimization, since this check is a bit simpler to some point?

alacritty_terminal/src/term/mod.rs
```diff
@@ -767,7 +686,7 @@ pub struct Term<T> {
     tabs: TabStops,
 
     /// Mode flags.
-    mode: TermMode,
+    pub mode: TermMode,
```
kchibisov, line 689: In general, I don't like that change, the thing is that mode is something
that set/unset by various side effects, not by the user directly and the user
can only read this field, not write to it various stuff. I do understand that
you wanted to remove a bunch of brackets, but having API that prevents you from
stupid things like unsetting mode yourself on the fly was kind of making sense.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1170,13 +1097,21 @@ impl<T> Term<T> {
...
     pub fn toggle_vi_mode(&mut self) {
         self.mode ^= TermMode::VI;
-        self.selection = None;
 
-        // Reset vi mode cursor position to match primary cursor.
-        if self.mode.contains(TermMode::VI) {
+        let vi_mode = self.mode.contains(TermMode::VI);
+
+        // Do not clear selection when entering search.
+        if self.regex_search.is_none() || !vi_mode {
```
kchibisov, line 1104: Wouldn't it be cleared anyway because you 'type'? I don't quite get it.
Or what is the idea behind this code, I don't feel like I get it.

alacritty_terminal/src/term/search.rs
```diff
@@ -0,0 +1,794 @@
...
+        let mut regex_iter = RegexIter::new(start, end, Direction::Right, &self).peekable();
+
+        // Check if there's any match at all.
+        let first_match = regex_iter.peek()?.clone();
+
+        let regex_match = regex_iter
+            .find(|regex_match| {
+                let match_point = Self::match_side(&regex_match, side);
+
+                // If the matche's point is beyond the origin, we're done.
```
kchibisov, line 114: matched? I don't quite like that `matche's`, not sure that it's a word.

alacritty_terminal/src/term/search.rs
```diff
@@ -0,0 +1,794 @@
...
+        let mut cell = *iter.cell();
+        self.skip_fullwidth(&mut iter, &mut cell, direction);
+        let mut point = iter.point();
+
+        loop {
+            // Convert char to array of bytes.
+            let mut buf = [0; 4];
+            let utf8_len = cell.c.encode_utf8(&mut buf).len();
+
+            // Pass char to DFA is individual bytes.
```
kchibisov, line 227:
```suggestion
            // Pass char to DFA as individual bytes.
```

alacritty_terminal/src/term/search.rs
chrisduerr: It should be `match's`, thanks.

alacritty_terminal/src/term/mod.rs
chrisduerr: Typing into the search bar does not clear the selection, no.

So you can start a selection, search, then your selection will be updated to the new point.

alacritty_terminal/src/term/mod.rs
chrisduerr: It's just that we don't have access to the `selected` information anymore since that's now in the RenderableCells creation method. So I've opted to instead just decide based on the resulting cell rather than the input cell.

It is true though that this should be more efficient since it skips checks that aren't useful.

alacritty_terminal/src/term/mod.rs
chrisduerr: What name would you suggest then?

alacritty_terminal/src/grid/mod.rs
```diff
@@ -427,6 +401,56 @@ impl<T> Grid<T> {
     }
 }
 
+/// Grid dimensions.
+pub trait Dimensions {
+    /// Total number of lines in the buffer, this includes scrollback and visible lines.
+    fn total_lines(&self) -> usize;
+
+    /// Height of the viewport in lines.
+    fn num_lines(&self) -> Line;
```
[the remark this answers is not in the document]
chrisduerr: I've kinda gotten used to `num_lines` but I've always thought it's a horrible name to settle on. Since there's really no way to know which one this represents. However it is extremely short which is important. `viewport_lines` is significantly longer.

Maybe `visible_lines` would be better, it's at least a little shorter. The name `screen_lines` would be even shorter.

I've decided on `screen_lines` or now, let me know what you think. I've also renamed `num_cols` to just `cols`, since we don't have `num_` as a prefix for the other functions either.

alacritty/src/input.rs
```diff
@@ -177,13 +192,44 @@ impl<T: EventListener> Execute<T> for Action {
...
+                let origin = ctx.terminal().visible_to_buffer(ctx.terminal().vi_mode_cursor.point);
+                let direction = ctx.search_direction().opposite();
+
+                let regex_match = ctx.terminal().search_next(origin, direction, Side::Right, None);
+                if let Some(regex_match) = regex_match {
+                    ctx.terminal_mut().vi_goto_point(*regex_match.end());
+                }
+            },
+            Action::Search => ctx.start_search(Direction::Right),
+            Action::SearchReverse => ctx.start_search(Direction::Left),
```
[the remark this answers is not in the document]
chrisduerr: No it's not? You should be able to start the search without entering search mode first. I explicitly tested for that and wrote code to make this work.

Only the options to jump to next/previous should be limited to vi mode.

alacritty/src/event.rs
```diff
@@ -294,19 +318,138 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
+        self.terminal.dirty = true;
+    }
+
+    #[inline]
+    fn confirm_search(&mut self) {
+        // Enter vi mode once search is confirmed.
+        self.terminal.mode.insert(TermMode::VI);
+
+        // Force limitless search if the previous one was interrupted.
+        if self.scheduler.scheduled(TimerId::DelayedSearch) {
```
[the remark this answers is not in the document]
chrisduerr: Unlimited is probably the better word, yes.

alacritty/src/event.rs
```diff
@@ -46,12 +44,18 @@ use crate::cli::Options;
...
-use crate::scheduler::Scheduler;
+use crate::scheduler::{Scheduler, TimerId};
 use crate::url::{Url, Urls};
 use crate::window::Window;
 
+/// Duration after the last user input until an unlimited search is performed.
+pub const TYPING_SEARCH_DELAY: Duration = Duration::from_millis(500);
+
+/// Maximum number of lines blockingly searched while still typing the search regex.
+const MAX_SEARCH_WHILE_TYPING: Option<usize> = Some(1000);
```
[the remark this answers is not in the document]
chrisduerr: Me neither, but I honestly don't know how to put it any better. I wouldn't want two `while`s thrown in and I think saying that this is blocking is an important note.

I've changed it up a little bit, let me know what you think.

alacritty/src/display.rs
```diff
@@ -328,13 +370,16 @@ impl Display {
         terminal: &mut Term<T>,
         pty_resize_handle: &mut dyn OnResize,
         message_buffer: &MessageBuffer,
+        search_active: bool,
```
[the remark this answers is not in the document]
chrisduerr: With multiple windows, you'd need multiple draw calls anyways, right? So we do want to pass it this way?

I don't think four parameters is reason for concern at this point, we've got much worse already. It's something to look out for, but I still think it's okay.

alacritty/src/display.rs
```diff
@@ -402,18 +453,25 @@ impl Display {
...
+        let vi_mode_cursor =
+            if terminal.mode.contains(TermMode::VI) { Some(terminal.vi_mode_cursor) } else { None };
 
         // Update IME position.
         #[cfg(not(windows))]
-        self.window.update_ime_position(&terminal, &self.size_info);
+        {
+            let point = match &search_regex {
+                Some(regex) => {
+                    let column = min(regex.len() + SEARCH_LABEL.len(), terminal.num_cols().0 - 1);
```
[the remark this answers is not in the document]
chrisduerr: It should be good now, including wide chars. I've tested it myself by logging the position, though I haven't actually run any IME. That said, it should work just as well as text on the last column.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -66,7 +66,7 @@ pub struct Cursor {
     /// The location of this cursor.
     pub point: Point,
 
-    /// Template cell when using this cursor.
+    /// Template cell when usingscreen_linessor.
```
kchibisov, line 69: your renaming got borken here.

alacritty/src/input.rs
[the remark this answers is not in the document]
kchibisov: Oh, and the point was to be in Vi mode in the end? Then I guess it's fine.

alacritty_terminal/src/term/mod.rs
kchibisov: I mean that you can't select while search bar is active, and entering search clears selection anyway. If you leave search you won't have a selection anyway. If you can give me an example how to enter search and have selection, it'll be really helpful to understand the intent of that.

chrisduerr: > entering search clears selection anyway.

It shouldn't. Why would it?

alacritty_terminal/src/term/search.rs
```diff
@@ -0,0 +1,794 @@
...
+        let mut regex_iter = RegexIter::new(start, end, Direction::Left, &self).peekable();
+
+        // Check if there's any match at all.
+        let first_match = regex_iter.peek()?.clone();
+
+        let regex_match = regex_iter
+            .find(|regex_match| {
+                let match_point = Self::match_side(&regex_match, side);
+
+                // If the matche's point is beyond the origin, we're done.
```
kchibisov, line 151:
```suggestion
                // If the match's point is beyond the origin, we're done.
```

PR #3510, opened 2020-03-24
Rename alt grid to inactive grid

### LINKED ISSUE #3504: Rename alt_grid in Term struct
The terminal currently keeps track of two grids, the `grid` and the `alt_grid`. This is done because a terminal has two grids at all times, one containing the currently active content and the other containing either the alternate grid if the alternate screen is not active, or the primary grid if it is.

Naming this variable `alt_grid` is very unfortunate, since it seems to indicate that this is always the alternate screen, which is not true. The same can be said about `cursor_save_alt`.

Since this has caused at least two bugs in the past, these variables should be renamed. One option that comes to mind would be `inactive_grid`. Renaming the primary grid would also be an option, however I think it's best to keep that short and concise.

DivineGod:
Small change to term struct.

Renames alt_grid to inactive_grid, alt to swapped (indicating if grids have been swapped more clearly), swap_alt to swap_grids (indicating that the grids get swapped), various other variables referring to alt changed to inactive

fixes #3504

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1196,6 +1196,9 @@ impl<T> Term<T> {
 
         self.alt = !self.alt;
         mem::swap(&mut self.grid, &mut self.inactive_grid);
+        mem::swap(&mut self.cursor, &mut self.inactive_cursor);
+        self.cursor.point.line = min(self.cursor.point.line, self.grid.num_lines() - 1);
+        self.cursor.point.col = min(self.cursor.point.col, self.grid.num_cols() - 1);
     }
 
     /// Scroll screen down
@@ -1811,16 +1814,12 @@ impl<T: EventListener> Handler for Term<T> {
 
     #[inline]
     fn save_cursor_position(&mut self) {
-        trace!("Saving cursor position");
-        mem::swap(&mut self.cursor, &mut self.inactive_cursor);
+        unimplemented!()
     }
 
     #[inline]
     fn restore_cursor_position(&mut self) {
-        trace!("Restoring cursor position");
-        mem::swap(&mut self.cursor, &mut self.inactive_cursor);
-        self.cursor.point.line = min(self.cursor.point.line, self.grid.num_lines() - 1);
-        self.cursor.point.col = min(self.cursor.point.col, self.grid.num_cols() - 1);
+        unimplemented!()
     }
 
     #[inline]
@@ -2046,9 +2045,7 @@ impl<T: EventListener> Handler for Term<T> {
             ansi::Mode::SwapScreenAndSetRestoreCursor => {
                 if !self.alt {
                     self.mode.insert(TermMode::ALT_SCREEN);
-                    // self.save_cursor_position();
                     self.swap_alt();
-                    self.save_cursor_position();
                 }
             },
             ansi::Mode::ShowCursor => self.mode.insert(TermMode::SHOW_CURSOR),
@@ -2099,9 +2096,7 @@ impl<T: EventListener> Handler for Term<T> {
             ansi::Mode::SwapScreenAndSetRestoreCursor => {
                 if self.alt {
                     self.mode.remove(TermMode::ALT_SCREEN);
-                    // self.restore_cursor_position();
                     self.swap_alt();
-                    self.restore_cursor_position();
                 }
             },
             ansi::Mode::ShowCursor => self.mode.remove(TermMode::SHOW_CURSOR),
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1197,8 +1197,6 @@ impl<T> Term<T> {
         self.alt = !self.alt;
         mem::swap(&mut self.grid, &mut self.inactive_grid);
         mem::swap(&mut self.cursor, &mut self.inactive_cursor);
-        self.cursor.point.line = min(self.cursor.point.line, self.grid.num_lines() - 1);
-        self.cursor.point.col = min(self.cursor.point.col, self.grid.num_cols() - 1);
     }
 
     /// Scroll screen down
@@ -1814,12 +1812,16 @@ impl<T: EventListener> Handler for Term<T> {
 
     #[inline]
     fn save_cursor_position(&mut self) {
-        unimplemented!()
+        trace!("Saving cursor position");
+        mem::swap(&mut self.cursor, &mut self.inactive_cursor);
     }
 
     #[inline]
     fn restore_cursor_position(&mut self) {
-        unimplemented!()
+        trace!("Restoring cursor position");
+        mem::swap(&mut self.cursor, &mut self.inactive_cursor);
+        self.cursor.point.line = min(self.cursor.point.line, self.grid.num_lines() - 1);
+        self.cursor.point.col = min(self.cursor.point.col, self.grid.num_cols() - 1);
     }
 
     #[inline]
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -815,7 +815,7 @@ pub struct Term<T> {
     input_needs_wrap: bool,
 
     /// Alternate grid.
-    alt_grid: Grid<Cell>,
+    inactive_grid: Grid<Cell>,
 
     /// Alt is active.
     alt: bool,
@@ -920,7 +920,7 @@ impl<T> Term<T> {
 
         let history_size = config.scrolling.history() as usize;
         let grid = Grid::new(num_lines, num_cols, history_size, Cell::default());
-        let alt = Grid::new(num_lines, num_cols, 0 /* scroll history */, Cell::default());
+        let alt_grid = Grid::new(num_lines, num_cols, 0, Cell::default());
 
         let tabs = TabStops::new(grid.num_cols());
 
@@ -933,7 +933,7 @@ impl<T> Term<T> {
             visual_bell: VisualBell::new(config),
             input_needs_wrap: false,
             grid,
-            alt_grid: alt,
+            inactive_grid: alt_grid,
             alt: false,
             active_charset: Default::default(),
             cursor: Default::default(),
@@ -990,7 +990,7 @@ impl<T> Term<T> {
         }
 
         if self.alt {
-            self.alt_grid.update_history(config.scrolling.history() as usize);
+            self.inactive_grid.update_history(config.scrolling.history() as usize);
         } else {
             self.grid.update_history(config.scrolling.history() as usize);
         }
@@ -1141,7 +1141,7 @@ impl<T> Term<T> {
         }
 
         self.grid.selection = None;
-        self.alt_grid.selection = None;
+        self.inactive_grid.selection = None;
 
         // Should not allow less than 2 cols, causes all sorts of checks to be required.
         if num_cols <= Column(1) {
@@ -1161,7 +1161,7 @@ impl<T> Term<T> {
 
         // Resize grids to new size
         self.grid.resize(!is_alt, num_lines, num_cols, &mut self.cursor.point, &Cell::default());
-        self.alt_grid.resize(is_alt, num_lines, num_cols, alt_cursor_point, &Cell::default());
+        self.inactive_grid.resize(is_alt, num_lines, num_cols, alt_cursor_point, &Cell::default());
 
         // Reset scrolling region to new size
         self.scroll_region = Line(0)..self.grid.num_lines();
@@ -1197,7 +1197,7 @@ impl<T> Term<T> {
         }
 
         self.alt = !self.alt;
-        mem::swap(&mut self.grid, &mut self.alt_grid);
+        mem::swap(&mut self.grid, &mut self.inactive_grid);
     }
 
     /// Scroll screen down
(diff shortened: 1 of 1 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -814,11 +814,11 @@ pub struct Term<T> {
...
 
-    /// Alternate grid.
-    alt_grid: Grid<Cell>,
+    /// Inactive grid.
+    inactive_grid: Grid<Cell>,
 
-    /// Alt is active.
-    alt: bool,
+    /// Inactive grid is active.
+    swapped: bool,
```
chrisduerr: This variable actually tracks if the alt screen is active. But I don't see why we would need it, we can just replace all usages of it with `self.mode.contains(TermMode::ALT_SCREEN)`.

alacritty_terminal/src/term/mod.rs
```diff
@@ -920,7 +920,8 @@ impl<T> Term<T> {
 
         let history_size = config.scrolling.history() as usize;
         let grid = Grid::new(num_lines, num_cols, history_size, Cell::default());
-        let alt = Grid::new(num_lines, num_cols, 0 /* scroll history */, Cell::default());
+        let inactive_grid =
+            Grid::new(num_lines, num_cols, 0 /* scroll history */, Cell::default());
```
chrisduerr:
```suggestion
        let alt_grid = Grid::new(num_lines, num_cols, 0, Cell::default());
```

Never was a big fan of this comment and especially with this linebreak it's probably better to get rid of it.

But I think I'd like to keep this instance as `alt_grid`, since at this point we need for certain that this is actually the alt grid and thus the `0` history size makes a lot of sense.

Just using `inactive_grid: alt_grid` in the `Term` creation should be fine.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1153,15 +1154,50 @@ impl<T> Term<T> {
...
+            let template = Cell { bg: self.cursor_save_inactive.template.bg, ..Cell::default() };
+            self.inactive_grid.scroll_up(&(Line(0)..old_lines), lines, &template);
+        }
+
+        // Move prompt down when growing if scrollback lines are available
+        if num_lines > old_lines && !self.mode.contains(TermMode::ALT_SCREEN) {
+            let growage = min(num_lines - old_lines, Line(self.grid.history_size()));
+            self.cursor.point.line += growage;
+        }
+
```
chrisduerr: Looks like you didn't rebase this correctly. This code does not exist on the latest master.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1153,15 +1154,50 @@ impl<T> Term<T> {
...
         debug!("New num_cols is {} and num_lines is {}", num_cols, num_lines);
 
-        let is_alt = self.mode.contains(TermMode::ALT_SCREEN);
-        let alt_cursor_point =
-            if is_alt { &mut self.cursor_save.point } else { &mut self.cursor_save_alt.point };
+        let is_swapped = self.mode.contains(TermMode::ALT_SCREEN);
+        let inactive_cursor_point = if is_swapped {
```
chrisduerr:
```suggestion
        let is_alt = self.mode.contains(TermMode::ALT_SCREEN);
        let inactive_cursor_point = if is_alt {
```

In this case `is_alt` is actually the correct term.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1190,14 +1228,14 @@ impl<T> Term<T> {
...
+        if self.swapped {
             let template = self.cursor.template;
             self.grid.region_mut(..).each(|c| c.reset(&template));
         }
 
-        self.alt = !self.alt;
-        mem::swap(&mut self.grid, &mut self.alt_grid);
+        self.swapped = !self.swapped;
+        mem::swap(&mut self.grid, &mut self.inactive_grid);
     }
```
chrisduerr: We should move everything related to alt screen swapping into this method, that would make using it a lot easier.

So this should insert/remove the term mode, save/restore cursor position and swap the screens.

If you feel like doing a little bit more work, @kchibisov and myself have also come to the conclusion that there is no need for three cursor states (cursor, cursor_save and cursor_save_alt). Instead we should be able to do the same thing we do with the grid and just `mem::swap` them out, so that we have a `cursor` and `inactive_cursor` and nothing else. Then we'd never have to check for the mode to determine which save cursor we want to use. The naming would also be more consistent.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2525,14 +2564,14 @@ mod tests {
         assert_eq!(term.grid.history_size(), 10);
         assert_eq!(term.cursor.point, Point::new(Line(9), Column(0)));
 
-        // Enter alt screen
+        // Enter inactive screen
```
chrisduerr:
```suggestion
        // Enter alt screen
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -2525,14 +2564,14 @@ mod tests {
...
         term.set_mode(ansi::Mode::SwapScreenAndSetRestoreCursor);
 
         // Increase visible lines
         size.height = 30.;
         term.resize(&size);
 
-        // Leave alt screen
+        // Leave inactive screen
```
chrisduerr:
```suggestion
        // Leave alt screen
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -2587,14 +2626,14 @@ mod tests {
         assert_eq!(term.grid.history_size(), 10);
         assert_eq!(term.cursor.point, Point::new(Line(9), Column(0)));
 
-        // Enter alt screen
+        // Enter inactive screen
```
chrisduerr:
```suggestion
        // Enter alt screen
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -2587,14 +2626,14 @@ mod tests {
...
         term.set_mode(ansi::Mode::SwapScreenAndSetRestoreCursor);
 
         // Increase visible lines
         size.height = 5.;
         term.resize(&size);
 
-        // Leave alt screen
+        // Leave inactive screen
```
chrisduerr:
```suggestion
        // Leave alt screen
```

alacritty_terminal/src/term/mod.rs
DivineGod: I don't mind trying out a bit of a refactor with all of this 👍

alacritty_terminal/src/term/mod.rs
DivineGod: Something did go horribly wrong it seems. I'll clean this up.

DivineGod:
I've pared back the initial commit and rebased it better.

I'll look into the cleanup of swapping code suggested later

DivineGod:
Having a few issues with the refactoring – I've managed to fix the test fails for now, but removing `self.alt` seems to be harder than I thought, as it seems it's not a 1-to-1 mapping of `self.mode.contains(TermMode::ALT_SCREEN)`

I'll investigate further and see what I can do

DivineGod:
I managed to refactor cursor swapping out into the `swap_alt` method.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1156,23 +1152,25 @@ impl<T> Term<T> {
...
         // Resize grids to new size
         self.grid.resize(!is_alt, num_lines, num_cols, &mut self.cursor.point, &Cell::default());
-        self.alt_grid.resize(is_alt, num_lines, num_cols, alt_cursor_point, &Cell::default());
+        self.inactive_grid.resize(
+            is_alt,
+            num_lines,
+            num_cols,
+            &mut self.inactive_cursor.point,
+            &Cell::default(),
+        );
```
chrisduerr: I feel like it might be worth extracting `&mut self.inactive_cursor.point` into a variable here to prevent this linewrap.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1197,7 +1195,10 @@ impl<T> Term<T> {
         }
 
         self.alt = !self.alt;
-        mem::swap(&mut self.grid, &mut self.alt_grid);
+        mem::swap(&mut self.grid, &mut self.inactive_grid);
+        mem::swap(&mut self.cursor, &mut self.inactive_cursor);
+        self.cursor.point.line = min(self.cursor.point.line, self.grid.num_lines() - 1);
+        self.cursor.point.col = min(self.cursor.point.col, self.grid.num_cols() - 1);
```
chrisduerr: What happens when this is removed? I don't think it should be necessary, since we do reflow saved cursors.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1813,20 +1814,12 @@ impl<T: EventListener> Handler for Term<T> {
...
     #[inline]
     fn restore_cursor_position(&mut self) {
-        trace!("Restoring cursor position");
-        let source = if self.alt { &self.cursor_save_alt } else { &self.cursor_save };
-
-        self.cursor = *source;
-        self.cursor.point.line = min(self.cursor.point.line, self.grid.num_lines() - 1);
-        self.cursor.point.col = min(self.cursor.point.col, self.grid.num_cols() - 1);
+        unimplemented!()
     }
```
chrisduerr: I forgot that these are actually called by `ansi.rs`, which makes a lot of sense. We need these functions for things to work properly.

But you can simply call these from `swap_alt`.

chrisduerr:
After looking at the code again, we actually need two different save cursors. However I still think we should swap them so that they always correspond to the correct grid. So `self.cursor_save` is for `self.grid` and `self.cursor_save_inactive` is for `self.alt_grid`.

alacritty_terminal/src/term/mod.rs
DivineGod: Tests still pass if this is removed

alacritty_terminal/src/term/mod.rs
DivineGod: I looked into it. They seem to be called, but nothing seems to call `csi_dispatch` which is where they're called from. I'll re-implement the methods just in case

alacritty_terminal/src/term/mod.rs
```diff
@@ -815,14 +815,17 @@ pub struct Term<T> {
     input_needs_wrap: bool,
 
     /// Alternate grid.
-    alt_grid: Grid<Cell>,
+    inactive_grid: Grid<Cell>,
 
     /// Alt is active.
     alt: bool,
```
envp: Is this related to the the `alt_grid` field? If so, should this also undergo a rename?

alacritty_terminal/src/term/mod.rs
kchibisov: >but nothing seems to call csi_dispatch

This method is called from outside of alacritty.

alacritty_terminal/src/term/mod.rs
nixpulvis: This is actually the crux of the confusion. The idea is that `alt` field indicates if we're in alt mode, and should probably be removed in favor of checking that directly. See https://github.com/alacritty/alacritty/pull/3510#issuecomment-605546448 and https://github.com/alacritty/alacritty/pull/3510#discussion_r397043039 for more about this.

envp: I see, I missed those. Thanks for clarifying!

chrisduerr, line 821: Is there a reason why this hasn't been removed yet?

alacritty_terminal/src/term/mod.rs
```diff
@@ -844,12 +847,6 @@ pub struct Term<T> {
 
     pub visual_bell: VisualBell,
 
-    /// Saved cursor from main grid.
-    cursor_save: Cursor,
-
-    /// Saved cursor from alt grid.
-    cursor_save_alt: Cursor,
```
chrisduerr, line 851: As mentioned previously, these are necessary. I'd recommend looking into XTerm to see if they're persisted across alt grid switches, but we definitely need at least one of these.

alacritty_terminal/src/term/mod.rs
DivineGod: I’ve been busy, I only really have time to look at this on the weekend. I tried removing this in the first iterations of the PR and ran into test failures.

There are some things I think I can do to remove this and make things work

alacritty_terminal/src/term/mod.rs
DivineGod: Can you explain what their function is, looked to me like they were used to store and restore data for `cursor` but doing `mem::swap` now seems to have the same effect to me.

DivineGod: If there’s a test we can do to show the difference then of course that would be welcome.

chrisduerr: If you send the terminal escape `ESC 7`, move the cursor and then send the escape `ESC 8`, the cursor should return to the position where it was when the `ESC 7` escape was sent. To do that we obviously need to store information about the cursor and that's what this is used for.

This can be used to save and restore the cursor position, charset and cell template.

Based on my testing it is not necessary (or intended?) to preserve this across alt grid boundaries, however it's still necessary to store it.

DivineGod: Makes sense, I’ll see if I can make a ref-test and make sure it works consistently

DivineGod:
Due to the situation that's happening right now, I do not have any capacity to keep working on this implementation at the moment

chrisduerr:
No problem, thanks for your work @DivineGod. There's some upcoming PRs that would make this a little bit easier. Your existing work rebased on that should do the job, which I can do myself once those are merged.

PR #3514, opened 2020-03-25
use fs::read_to_string in tests instead of custom implementation.

matthiaskrgr:
Should have run with --all-targets --all-features previously... :)


kchibisov:
Thanks.

chrisduerr:
For future reference, functions shouldn't be directly imported, instead the module should be used to reference the function. The same way we did it for the last PR:

b48e2fc2d813723ac73030b2e8f5a21ad5fd37ed

PR #3536, opened 2020-03-30
Fix various mouse mode + vi mode interactions

kchibisov:
This commit fixes some issues introduced by
1a8cd172e520e493bacc9c6a2ae6f80de086eaa3:

 1. Vi cursor not moving properly on double/triple click
 2. URL not launching via mouse click in vi mode + mouse mode
 3. Ability to select in mouse mode with double/triple click regardless
    of shift modifier

### DIFF
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -455,20 +455,6 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
         }
     }
 
-    fn on_mouse_double_click(&mut self, button: MouseButton, point: Point) {
-        if button == MouseButton::Left {
-            let side = self.ctx.mouse().cell_side;
-            self.ctx.start_selection(SelectionType::Semantic, point, side);
-        }
-    }
-
-    fn on_mouse_triple_click(&mut self, button: MouseButton, point: Point) {
-        if button == MouseButton::Left {
-            let side = self.ctx.mouse().cell_side;
-            self.ctx.start_selection(SelectionType::Lines, point, side);
-        }
-    }
-
     fn on_mouse_press(&mut self, button: MouseButton) {
         // Handle mouse mode
         if !self.ctx.modifiers().shift() && self.ctx.mouse_mode() {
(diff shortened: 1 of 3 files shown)


kchibisov:
You should click and move your cursor during the click, so your triple click is moving location during it, pretty tricky to repro, but it's possible.

picture

alacritty/src/input.rs
```diff
@@ -496,12 +506,16 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
             {
                 self.ctx.mouse_mut().block_url_launcher = true;
                 self.on_mouse_triple_click(button, point);
+                // Move vi mode cursor to mouse position
+                if self.ctx.terminal().mode().contains(TermMode::VI) {
+                    // Update vi mode cursor position on triple click
+                    self.ctx.terminal_mut().vi_mode_cursor.point = point;
+                }
```
chrisduerr, line 513: If we do this every time the user clicks, could we do this outside of this match to reduce duplication?

alacritty/src/input.rs
```diff
@@ -481,13 +481,23 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
         let mut point = self.ctx.size_info().pixels_to_coords(mouse.x, mouse.y);
         point.line = min(point.line, self.ctx.terminal().grid().num_lines() - 1);
 
+        // Reset click state in mouse mode
+        if !self.ctx.modifiers().shift() && self.ctx.mouse_mode() {
+            self.ctx.mouse_mut().click_state = ClickState::None;
+        }
```
chrisduerr, line 487: I don't quite understand this part. Why would we even get to this point, just to reset it?

It seems like what we actually want is to always execute the same code when mouse mode is active, which is the `self.mouse_report(code, ElementState::Pressed);` one?

It seems like restructuring the method could lead to significant improvements in readability here.

kchibisov: You can change state during triple/double click, so instead of having it everywhere for mouse mode, I've decided to put it here, doing it in `self.mouse_report(code, ElementState::Pressed);` doesn't work in the mentioned scenario, since your first click could be in normal mode, which won't trigger that reset, but the second/third one in mouse mode.

alacritty/src/input.rs
```diff
@@ -455,89 +455,78 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
+            let code = match button {
+                MouseButton::Left => 0,
+                MouseButton::Middle => 1,
+                MouseButton::Right => 2,
+                // Can't properly report more than three buttons.
+                MouseButton::Other(_) => return,
+            };
+
+            self.mouse_report(code, ElementState::Pressed);
+            return;
```
chrisduerr, line 472: Probably the biggest thing standing out in the current state of this PR is these two early returns.

While the `button != Left` early return is fairly small and straight-forward, this is some pretty big chunk of code for an early return. That's usually a bit of a code smell and increases the complexity of understanding the code a bit.

The method is also already 73 lines long, which is relatively big I'd say and is quickly approaching the limits of what I'd say is acceptable.

That's why I would suggest moving everything after these early returns to a separate function for handling just LMB clicks when mouse mode is not enabled.

That would also allow getting rid of early returns by just doing `if { ... } else if button != Left { ... } else { lmb_stuff() }`.

Issue #972, opened 2017-12-27
Background blur


chriskrycho:
Which operating system does the issue occur on? **macOS 10.13.2**

If on linux, are you using X11 or Wayland? **N/A**

Transparent background works nicely; however, for many contexts this is essentially unusable; you also need a background *blur* setting (as Terminal.app and iTerm.app have on macOS).

maximbaz:
Interesting request 🙂 
I was curious if this request applies to Linux too - apparently not, because compton already allows this:

-------

![image](https://user-images.githubusercontent.com/1177900/34395549-66837650-eb61-11e7-8909-c0c77bd4f918.png)

--------------

If you are interested to achieve similar results, play with compton's `--blur-background` and `--blur-kern` CLI arguments.

unphased:
fwiw, this doesn't invalidate the issue in any way (I would love to have a blur option, in fact), but with iTerm2 I've found that I always end up setting no blur and leaving it to a low transparency setting so I can at least get some transparent effect at least: The blurring always incurs a substantial cpu load increase. If this can be done in a way that effectively leverages the GPU, it would be lightweight enough to not appreciably impact battery life. Otherwise, it's less attractive at least for me.

mikelui:
Anyone looking at implementing this can also look at KDE's new blur implementation which is meant to be lightweight:
https://www.reddit.com/r/kde/comments/7szqqk/i_implemented_a_new_blur_effect_in_kde_it/ 

The technique used is described in: https://community.arm.com/cfs-file/__key/communityserver-blogs-components-weblogfiles/00-00-00-26-50/siggraph2015_2D00_mmg_2D00_marius_2D00_notes.pdf

diversario:
I'd like to support this (kinda vain, I admit it) request. It's a (very) nice to have.

sandangel:
may I ask is there any update on this?

chrisduerr:
If there is any progress on an issue, the issue will be updated. Asking about updates when there is none is unlikely to speed things up.

I also do not expect this to get fixed anytime soon, since it's very low priority for me. Unless someone else sends a simple PR which adds this, it's unlikely to be resolved in the near future.

creikey:
Going to take a look at this... need it for my mac os i3 rice

wucke13:
[...]

PR #3537, opened 2020-03-30
Add support for background blur

Soft:
Add new `background_blur` boolean configuration option for enabling background
blur effect.

This setting currently only has any effect when Alacritty is used with KWin and
X11 as the implementation relies on KWin's compositor for blurring.

This feature was requested in #972 

Considering the limited OS/WM support, I realize that you might not want to introduce this feature as-is. However, I still decide to open this PR as a sort of proof-of-concept implementation that works well enough for me. Maybe someone else will also find it useful.

Screenshot_20200330_192224

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Default Command+N keybinding for SpawnNewInstance on macOS
 - Vi mode for copying text and opening links
 - `CopySelection` action which copies into selection buffer on Linux/BSD
+- Config option `background_blur` to enable background blur effect
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


kchibisov:
It seems like a KDE specific feature tbh, I doubt it'll work on GNOME, compton, etc. Also on X11 in kwin you can set this property in wm settings, so I don't think that it should be a part of alacritty here. This feature is likely for macOS/Windows/Wayland users tbh.

Soft:
Yes, the current implementation is likely mostly going to benefit KDE users as it depends on KWin. As far as I know there is no standard way in X11 for applications to request parts of their background to be blurred so any implementation that relies on the compositor is going to be compositor specific.

With Compton I think you might be able to achieve this by manually editing config file. For GNOME, I don't know if they have any official support for this sort of thing. I am not sure if this effect is currently configurable through KDE's settings, although of course you can enable it for a particular window using something like `xprop`.

kchibisov:
>I am not sure if this effect is currently configurable through KDE's settings, although of course you can enable it for a particular window using something like xprop.

It should be configurable, since I've done it in the past, but I have no idea where it was, I feel like there's a lot of articles around on how to do so.

PR #3546, opened 2020-04-02
Convert Shell.program from Cow to String

sterlingjensen:
In the interest of easing future efforts to simplify and decouple the config facilities, I think that this use of Cow can be changed to a String with no discernible performance impact. I've tested it as best I can, but I'd really welcome a second look at the small amount of serde related code I've touched here.

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -41,7 +41,7 @@ pub struct Options {
     pub class: Option<String>,
     pub embed: Option<String>,
     pub log_level: LevelFilter,
-    pub command: Option<Shell<'static>>,
+    pub command: Option<Shell>,
     pub hold: bool,
     pub working_dir: Option<PathBuf>,
     pub config: Option<PathBuf>,
@@ -245,7 +245,7 @@ impl Options {
             // Arg::min_values(1) is set.
             let command = String::from(args.next().unwrap());
             let args = args.map(String::from).collect();
-            options.command = Some(Shell::new_with_args(command, args));
+            options.command = Some(Shell::new_with_args(&command, args));
         }
 
         if matches.is_present("hold") {
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -245,7 +245,7 @@ impl Options {
             // Arg::min_values(1) is set.
             let command = String::from(args.next().unwrap());
             let args = args.map(String::from).collect();
-            options.command = Some(Shell::new_with_args(&command, args));
+            options.command = Some(Shell::new_with_args(command, args));
         }
 
         if matches.is_present("hold") {
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -49,8 +49,8 @@ pub struct Options {
 }
 
 impl Default for Options {
-    fn default() -> Options {
-        Options {
+    fn default() -> Self {
+        Self {
             live_config_reload: None,
             print_events: false,
             ref_test: false,
@@ -78,7 +78,7 @@ impl Options {
             version = format!("{} ({})", version, commit_hash);
         }
 
-        let mut options = Options::default();
+        let mut options = Self::default();
 
         let matches = App::new(crate_name!())
             .version(version.as_str())
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -266,12 +266,12 @@ pub struct Shell {
 }
 
 impl Shell {
-    pub fn new(program: &str) -> Self {
-        Self { program: program.into(), args: Vec::new() }
+    pub const fn new(program: String) -> Self {
+        Self { program, args: Vec::new() }
     }
 
-    pub fn new_with_args(program: &str, args: Vec<String>) -> Self {
-        Self { program: program.into(), args }
+    pub const fn new_with_args(program: String, args: Vec<String>) -> Self {
+        Self { program, args }
     }
 }
 
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -143,7 +143,7 @@ pub struct Config<T> {
 
 impl<T> Config<T> {
     #[inline]
-    pub fn draw_bold_text_with_bright_colors(&self) -> bool {
+    pub const fn draw_bold_text_with_bright_colors(&self) -> bool {
         self.draw_bold_text_with_bright_colors
     }
 
@@ -155,7 +155,7 @@ impl<T> Config<T> {
 
     /// Live config reload
     #[inline]
-    pub fn live_config_reload(&self) -> bool {
+    pub const fn live_config_reload(&self) -> bool {
         self.live_config_reload.0
     }
 
@@ -165,13 +165,13 @@ impl<T> Config<T> {
     }
 
     #[inline]
-    pub fn dynamic_title(&self) -> bool {
+    pub const fn dynamic_title(&self) -> bool {
         self.dynamic_title.0
     }
 
     /// Cursor foreground color.
     #[inline]
-    pub fn cursor_text_color(&self) -> Option<Rgb> {
+    pub const fn cursor_text_color(&self) -> Option<Rgb> {
         self.colors.cursor.text
     }
 
@@ -183,13 +183,13 @@ impl<T> Config<T> {
 
     /// Vi mode cursor foreground color.
     #[inline]
-    pub fn vi_mode_cursor_text_color(&self) -> Option<Rgb> {
+    pub const fn vi_mode_cursor_text_color(&self) -> Option<Rgb> {
         self.colors.vi_mode_cursor.text
     }
 
     /// Vi mode cursor background color.
     #[inline]
-    pub fn vi_mode_cursor_cursor_color(&self) -> Option<Rgb> {
+    pub const fn vi_mode_cursor_cursor_color(&self) -> Option<Rgb> {
         self.colors.vi_mode_cursor.cursor
     }
 
@@ -200,7 +200,7 @@ impl<T> Config<T> {
 
     /// Send escape sequences using the alt key
     #[inline]
-    pub fn alt_send_esc(&self) -> bool {
+    pub const fn alt_send_esc(&self) -> bool {
         self.alt_send_esc.0
     }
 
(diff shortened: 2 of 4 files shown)


alacritty/src/cli.rs
```diff
@@ -245,7 +245,7 @@ impl Options {
             // Arg::min_values(1) is set.
             let command = String::from(args.next().unwrap());
             let args = args.map(String::from).collect();
-            options.command = Some(Shell::new_with_args(command, args));
+            options.command = Some(Shell::new_with_args(&command, args));
```
chrisduerr, line 248: This is kind of bad style. Creating something as String, just to pass it to a function as reference which then converts it to a String again.

If this is the only consumer, a `String` parameter would work just fine. Otherwise `Into<String>` is also a better alternative.

alacritty_terminal/src/config/mod.rs
```diff
@@ -144,7 +143,7 @@ pub struct Config<T> {
 
 impl<T> Config<T> {
     #[inline]
-    pub fn draw_bold_text_with_bright_colors(&self) -> bool {
+    pub const fn draw_bold_text_with_bright_colors(&self) -> bool {
```
chrisduerr, line 146: Why did you change all of these to `const` functions?

alacritty_terminal/src/config/mod.rs
```diff
@@ -253,38 +252,26 @@ pub struct Cursor {
...
-    }
-
-    pub fn new_with_args<S>(program: S, args: Vec<String>) -> Shell<'a>
-    where
-        S: Into<Cow<'a, str>>,
-    {
-        Shell { program: program.into(), args }
+impl Shell {
+    pub fn new(program: &str) -> Self {
+        Self { program: program.into(), args: Vec::new() }
```
chrisduerr, line 270: Taking a `&str` just to call `into()` on it is bad style.

alacritty/src/cli.rs
sterlingjensen: I was hoping to avoid the OS specific files in this pull, but I guess I can't.

alacritty_terminal/src/config/mod.rs
sterlingjensen: The lint commit is full of changes at the direction of the linter. The linter is correct here, those getter functions are const. The untouched getters operate on options, I'm not even sure if there is a feature gate for that.

chrisduerr: Interesting, I wasn't aware of this lint it must be new.

Of course these functions can be marked as `const`, but it's pointless. Could you point me to the exact clippy/Rust version you're running? Since the latest nightly that supports rustfmt and clippy does not warn on this.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -152,9 +152,9 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, window_id: Option<usize>) ->
         let shell_name = pw.shell.rsplit('/').next().unwrap();
         let argv = vec![String::from("-c"), format!("exec -a -{} {}", shell_name, pw.shell)];
 
-        Shell::new_with_args("/bin/bash", argv)
+        Shell::new_with_args(String::from("/bin/bash"), argv)
     } else {
-        Shell::new(pw.shell)
+        Shell::new(String::from(pw.shell))
```
chrisduerr, line 157: Having the parameter take `Into<String>` as suggested earlier would make it much easier to use.

alacritty_terminal/src/config/mod.rs
sterlingjensen: I try to go above and beyond when it comes to annoying the maintainers of software I use. `-W clippy::nursery`

sterlingjensen: I'll add that `const` better expresses intent, and allows for `const` propagation:
https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn

alacritty_terminal/src/tty/unix.rs
sterlingjensen: True, but that is the price of specificity - reducing generality in the interest of simplicity. I thought it was a good trade, given how short the list of useful `Into<String>` producers for this use case is. I suppose the benefit is less evident without the larger intent in view: I'd like to see all the config facilities centralized, because at present they're pretty scattered. I've found practical simplification of hypothetical generality helpful in that endeavor. But I certainly don't want to waste your time, as I did with the 2018 edition font mod episode, so let me know if I'm heading in the wrong direction here and I'll close this pull and begin work on a different way.

alacritty_terminal/src/config/mod.rs
chrisduerr: > that const better expresses intent

It better expresses intent for what?

> allows for const propagation

We don't need that though, you've mentioned in your other comment that you'd like to simplify things and this seems to go in the exact opposite direction.

Adding `const` to all functions just because they could potentially maybe sometime be called from a `const` context doesn't really make much sense when you know that the chance for that is basically nill and they're only for internal use. If we need something to be `const` internally, we can just change it when we need it.

These are functions that read information from a configuration file that is deserialized. There is no usecase to do that in a `const` context. It's just verbose and I'd argue that it goes completely against expressing intent. If you make something `const` you're expressing the intent that it could potentially be interesting to call this from a const context and thus it should not be changed to a 'normal' function unless absolutely necessary. It will be a breaking change to do so and will be completely impossible if consumers already rely on that. I don't think this is what we're trying to do here? It seems like this change is mostly 'it currently does not hurt to make it const', which is completely different.

alacritty_terminal/src/tty/unix.rs
chrisduerr: I don't see how this simplifies things.

With what we currently have, every consumer of this API needs to convert their input to Strings, adding at minimum `.into()` to every single call.

While using `Into<String>` as parameter allows passing both `&str` and `String` without any disadvantages, while doing all the work in the callee, rather than the caller.

If you ask me, removing duplicates from the callsite and moving it into the callee is simplicity, not duplicating everything.

> But I certainly don't want to waste your time, as I did with the 2018 edition font mod episode, so let me know if I'm heading in the wrong direction here and I'll close this pull and begin work on a different way.

Please don't make any rash decisions. Fundamentally I don't see anything wrong with this PR, but I cannot merge something unless it is implemented properly and following all the repositorie's guidelines.

alacritty_terminal/src/config/mod.rs
sterlingjensen: > It better expresses intent for what?

Being a getter, providing no utility (aside from API boundary) beyond direct reference to the struct member. I am presupposing that they are getter functions though, and that they aren't intended to be stuffed with run time data manipulation logic.

> These are functions that read information from a configuration file that is deserialized.

These functions read information from a struct. That is not a distinction without a difference, especially when you start thinking about default values, where they are stored and how they are influenced by build time flags.

alacritty_terminal/src/tty/unix.rs
sterlingjensen: > I don't see how this simplifies things.

An unneeded type (Cow), lifetime (static), trait bounds, and trait implementation was removed. The cost is a more specific parameter being passed from the OS implementation file.

> Please don't make any rash decisions.

I'm not, it isn't as if I take your objections personally. I just know, through prior experience with your reviews, that when it starts getting into matters of opinion - the probability for a merge drops very fast. So the sooner that can be shortcut, the better.

chrisduerr: > An unneeded type (Cow), lifetime (static), trait bounds, and trait implementation was removed.

I was talking about taking a `String` rather than `Into<String>`, not this PR in general.

alacritty_terminal/src/config/mod.rs
chrisduerr: > Being a getter, providing no utility (aside from API boundary) beyond direct reference to the struct member.

Who says we want to guarantee that? It's not a good idea to lock an API down to a restriction like that unless you're absolutely certain that's always going to be the case and you actually gain any benefits from doing so. Currently you're trying to lock things down without any benefits at all.

> These functions read information from a struct. That is not a distinction without a difference

It is pointless to look at code without looking at the bigger picture of where and how it is used. Only because it might sometimes be useful for struct access to be `const`, doesn't mean it should always be `const` whenever possible.

It makes no sense for this struct to be accessed in a `const` environment.

sterlingjensen: > Who says we want to guarantee that? It's not a good idea to lock an API down...

Nobody, as far as I know. I didn't even know that they were there to guarantee anything.

> It is pointless to look at code without looking at the bigger picture of where and how it is used.

I agree, one can't look at the present state of the config code and conclude "yeah, this is a sane place to rely on for default values that can be overridden at run time."

alacritty_terminal/src/tty/unix.rs
sterlingjensen: > ...when your opinions are different than the guidelines of the Alacritty project, there's no chance of them being merged.

I've got no problem with the benevolent dictator model, I'd actually say that it is preferable in most software projects (including this one)... so I don't understand why you'd suggest that there is some kind of codified guideline that has been violated (here or in the prior event that I referenced). I see no such guideline, and I see nothing in my conduct suggesting that I operate under the illusion that a project guideline related to this even exists.

I'm going to assume that your final sentiment on standards was directly related to the code under discussion - and the short circuit I was asking for.

chrisduerr: > I see no such guideline, and I see nothing in my conduct suggesting that I operate under the illusion that a project guideline related to this even exists.

That's why I'm telling you. Of course not every little standard is documented, but I do know how it should be done to be in line with the rest of the Alacritty codebase. Which in this case would be `Into<String>` and not `String`.

alacritty_terminal/src/config/mod.rs
chrisduerr: > this is a sane place to rely on for default values that can be overridden at run time

That's not an accurate description of what the configuration is for. It is entirely a runtime structure, so these aren't just default values that can be overriden at runtime, this is a runtime configuration structure that can change at any time during execution.

chrisduerr:
Closing in favor of https://github.com/alacritty/alacritty/pull/3817.

PR #3547, opened 2020-04-02
Add option 'cursor.thickness' to set cursor thickness

### LINKED ISSUE #3526: New feature: Make various cursor dimensions configurable
I think it'd be great if the beam width could be adjusted from the configuration file. I'm not sure how the cursor is rendered and what scale could be applied, but some way of adjusting it should probably exist. In a discussion on irc @kchibisov said it'd be nice to have some sort of option to tune the style of the underline as well. 

A good solution would likely be a uniform way of tuning the different cursor styles, my proposed format for the configuration file is:
```
cursor:
  style: [Block/Underline/Beam]
  height: [\dpx]
  width: [\dpx]

  Underline:
    height: [\dpx]
    width: [\dpx] (inhibit from default)
  Beam:
    height: [\dpx] (inhibit from default)
    width: [\dpx]
  Box:
    height: [\dpx] (inhibit from default)
    width: [\dpx] (inhibit from default)
    border_thickness: [\dpx]
```

kchibisov:
Fixes #3526.

### DIFF
--- a/alacritty/src/cursor.rs
+++ b/alacritty/src/cursor.rs
@@ -26,13 +26,13 @@ pub fn get_cursor_glyph(
     offset_x: i8,
     offset_y: i8,
     is_wide: bool,
-    cursor_thickness: f32,
+    cursor_thickness: f64,
 ) -> RasterizedGlyph {
     // Calculate the cell metrics
     let height = metrics.line_height as i32 + i32::from(offset_y);
     let mut width = metrics.average_advance as i32 + i32::from(offset_x);
 
-    let line_width = cmp::max((f64::from(cursor_thickness) * f64::from(width)).round() as i32, 1);
+    let line_width = cmp::max((cursor_thickness * f64::from(width)).round() as i32, 1);
 
     // Double the cursor width if it's above a double-width glyph
     if is_wide {
--- a/alacritty/src/cursor.rs
+++ b/alacritty/src/cursor.rs
@@ -20,20 +20,19 @@ use alacritty_terminal::ansi::CursorStyle;
 
 use font::{BitmapBuffer, Metrics, RasterizedGlyph};
 
-/// Width/Height of the cursor relative to the font width
-pub const CURSOR_WIDTH_PERCENTAGE: f64 = 0.15;
-
 pub fn get_cursor_glyph(
     cursor: CursorStyle,
     metrics: Metrics,
     offset_x: i8,
     offset_y: i8,
     is_wide: bool,
+    cursor_thickness: f32,
 ) -> RasterizedGlyph {
     // Calculate the cell metrics
     let height = metrics.line_height as i32 + i32::from(offset_y);
     let mut width = metrics.average_advance as i32 + i32::from(offset_x);
-    let line_width = cmp::max((f64::from(width) * CURSOR_WIDTH_PERCENTAGE).round() as i32, 1);
+
+    let line_width = cmp::max((f64::from(cursor_thickness) * f64::from(width)).round() as i32, 1);
 
     // Double the cursor width if it's above a double-width glyph
     if is_wide {
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -286,7 +286,15 @@ impl Display {
         size_info.cell_height = cell_height;
     }
 
-    /// Process update events
+    /// Clear glyph cache.
+    fn clear_glyph_cache(&mut self) {
+        let cache = &mut self.glyph_cache;
+        self.renderer.with_loader(|mut api| {
+            cache.clear_glyph_cache(&mut api);
+        });
+    }
+
+    /// Process update events.
     pub fn handle_update<T>(
         &mut self,
         terminal: &mut Term<T>,
@@ -300,7 +308,7 @@ impl Display {
             if let Some(font) = update_pending.font {
                 self.update_glyph_cache(config, font);
             } else if update_pending.cursor.is_some() {
-                self.update_glyph_cache(config, config.font.clone());
+                self.clear_glyph_cache();
             }
         }
 
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -296,8 +296,12 @@ impl Display {
         update_pending: DisplayUpdate,
     ) {
         // Update font size and cell dimensions
-        if let Some(font) = update_pending.font {
-            self.update_glyph_cache(config, font);
+        if update_pending.font.is_some() || update_pending.cursor.is_some() {
+            if let Some(font) = update_pending.font {
+                self.update_glyph_cache(config, font);
+            } else if update_pending.cursor.is_some() {
+                self.update_glyph_cache(config, config.font.clone());
+            }
         }
 
         let cell_width = self.size_info.cell_width;
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -304,10 +304,10 @@ impl Display {
         update_pending: DisplayUpdate,
     ) {
         // Update font size and cell dimensions
-        if update_pending.font.is_some() || update_pending.cursor.is_some() {
+        if update_pending.font.is_some() || update_pending.cursor {
             if let Some(font) = update_pending.font {
                 self.update_glyph_cache(config, font);
-            } else if update_pending.cursor.is_some() {
+            } else if update_pending.cursor {
                 self.clear_glyph_cache();
             }
         }
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -304,12 +304,10 @@ impl Display {
         update_pending: DisplayUpdate,
     ) {
         // Update font size and cell dimensions
-        if update_pending.font.is_some() || update_pending.cursor {
-            if let Some(font) = update_pending.font {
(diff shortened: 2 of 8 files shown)


alacritty/src/event.rs
```diff
@@ -532,7 +532,11 @@ impl<N: Notify + OnResize> Processor<N> {
 
                         processor.ctx.terminal.update_config(&config);
 
-                        if processor.ctx.config.font != config.font {
+                        // We're also reloading font on thickness change, since we update
+                        // glyph cache
```
chrisduerr, line 536: The phrasing here is a bit weird, the `also` suggests to me that there was a previous comment already talking about something that was reloaded, but there isn't.

```suggestion
                        // Reload font on thickness change, since we need to update glyph cache
```

Though generally it seems a bit strange to me, since we don't actually want to do half of the things in this method? It also seems a bit strange to me to reload the font just for this, it doesn't really feel right.

alacritty_terminal/src/config/mod.rs
```diff
@@ -249,15 +249,68 @@ pub struct Cursor {
...
+
+impl<'de> Deserialize<'de> for CursorThickness {
+    fn deserialize<D>(deserializer: D) -> ::std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let value = serde_yaml::Value::deserialize(deserializer)?;
+        match u8::deserialize(value) {
+            Ok(thickness) => {
+                if thickness > 30 {
```
chrisduerr, line 281: Where do these limits come from? Neither 10 nor 30 make much sense to me.

alacritty/src/cursor.rs
```diff
@@ -17,23 +17,29 @@
...
 ) -> RasterizedGlyph {
     // Calculate the cell metrics
     let height = metrics.line_height as i32 + i32::from(offset_y);
     let mut width = metrics.average_advance as i32 + i32::from(offset_x);
-    let line_width = cmp::max((f64::from(width) * CURSOR_WIDTH_PERCENTAGE).round() as i32, 1);
+    let cursor_thickness = if cursor == CursorStyle::HollowBlock {
+        CursorThickness::default().0
+    } else {
+        cursor_thickness.0
+    };
```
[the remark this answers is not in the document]
kchibisov: The conversation was purely about `Beam` and `Underline` on the #3526, so I've decided to use default value for `HollowBlock` in the end, however I feel like we can safely apply thickness to it as well, since it makes things kind of consistent.

alacritty/src/event.rs
kchibisov: We should clear loader, which requires to reloading both caches in the same time, doing them separately doesn't work well, due to the fact that you should update atlases.

kchibisov: Updated as well, so resolved.

alacritty/src/renderer/mod.rs
```diff
@@ -302,16 +299,22 @@ impl GlyphCache {
...
         &mut self,
         font: config::Font,
         dpr: f64,
         loader: &mut L,
     ) -> Result<(), font::Error> {
-        // Clear currently cached data in both GL and the registry
-        loader.clear();
-        self.cache = HashMap::default();
-        self.cursor_cache = HashMap::default();
+        self.clear_glyph_cache(loader);
```
chrisduerr, line 317: There's no reason why this needs to be at the top, right?

That's why my previous review just suggested to call `clear_glyph_cache` directly. No point to do the work twice.

alacritty_terminal/src/config/mod.rs
```diff
@@ -242,20 +242,58 @@ impl Default for EscapeChars {
...
     }
+
+    #[inline]
+    pub fn thickness(self) -> f32 {
+        self.thickness.0
+    }
+}
+/// Wrapper around f32 that represents a cursor thickness between 0.0 and 1.0
+#[derive(Clone, Copy, Debug, PartialEq)]
+struct CursorThickness(f32);
```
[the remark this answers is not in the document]
kchibisov: I'm in favor of using `f32`, since it's OpenGL stuff for alpha is using f32 all over the place, and I'm in favor of casting f32 to f64 than downcasting.

alacritty/src/event.rs
```diff
@@ -51,11 +51,15 @@ pub struct DisplayUpdate {
...
+    pub cursor: Option<()>,
 }
 
 impl DisplayUpdate {
     fn is_empty(&self) -> bool {
-        self.dimensions.is_none() && self.font.is_none() && self.message_buffer.is_none()
+        self.dimensions.is_none()
+            && self.font.is_none()
+            && self.message_buffer.is_none()
+            && self.cursor.is_none()
```
chrisduerr, line 62: Quick question, do you know why we use `Option<()>` instead of `bool` for `message_buffer` and `cursor`?

I don't remember having a specific reason to do so and both are basically just single bit invariants.

alacritty/src/renderer/mod.rs
```diff
@@ -344,6 +342,13 @@ impl GlyphCache {
         self.metrics
     }
 
+    fn load_common_glyphs<L: LoadGlyph>(&mut self, loader: &mut L) {
```
chrisduerr, line 345:
```suggestion
    /// Prefetch glyphs that are almost guaranteed to be loaded anyways.
    fn load_common_glyphs<L: LoadGlyph>(&mut self, loader: &mut L) {
```

alacritty_terminal/src/config/mod.rs
```diff
@@ -242,20 +243,47 @@ impl Default for EscapeChars {
...
 impl Cursor {
+    #[inline]
     pub fn unfocused_hollow(self) -> bool {
         self.unfocused_hollow.0
     }
+
+    #[inline]
+    pub fn thickness(self) -> f64 {
+        self.thickness.0 as f64
+    }
```
chrisduerr, line 267: Why do we cast this to `f64` here, rather than making `Percentage` an `f64`?

Precision isn't really very important with these options, but it seems extremely cheap in this context so I feel like bigger might sometimes be better?

alacritty_terminal/src/config/mod.rs
```diff
@@ -242,20 +243,47 @@ impl Default for EscapeChars {
...
 
 #[serde(default)]
-#[derive(Deserialize, Copy, Clone, Debug, Default, PartialEq, Eq)]
+#[derive(Deserialize, Copy, Clone, Debug, PartialEq)]
 pub struct Cursor {
     #[serde(deserialize_with = "failure_default")]
     pub style: CursorStyle,
     #[serde(deserialize_with = "option_explicit_none")]
     pub vi_mode_style: Option<CursorStyle>,
+    #[serde(deserialize_with = "default_thickness")]
```
chrisduerr, line 252: I think you need to add a `default = default_thickness` annotation otherwise it would print an error when the key is missing?

alacritty_terminal/src/config/mod.rs
```diff
@@ -242,20 +243,47 @@ impl Default for EscapeChars {
...
+        Self {
+            style: Default::default(),
+            vi_mode_style: Default::default(),
+            thickness: Percentage::new(DEFAULT_CURSOR_THICKNESS),
+            unfocused_hollow: Default::default(),
+        }
+    }
+}
+
+pub fn default_thickness<'a, D>(deserializer: D) -> Result<Percentage, D::Error>
```
chrisduerr, line 281: The naming here isn't really great, since usually that's more the naming we use for default values, instead of deserializers.

So it should be something like `deserialize_cursor_thickness`.

alacritty/src/display.rs
```diff
@@ -296,8 +304,12 @@ impl Display {
...
         // Update font size and cell dimensions
-        if let Some(font) = update_pending.font {
-            self.update_glyph_cache(config, font);
+        if update_pending.font.is_some() || update_pending.cursor.is_some() {
+            if let Some(font) = update_pending.font {
+                self.update_glyph_cache(config, font);
+            } else if update_pending.cursor.is_some() {
+                self.clear_glyph_cache();
+            }
         }
```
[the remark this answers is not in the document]
kchibisov: You should clear glyph_cache only once, I don't see a reason to do so twice, and for font you need `update_glyph_cache`, but for `cursor` only clear, and changing font should also update cursor anyway, since cursor is based on cell metrics, which are based on font metrics.

alacritty/src/event.rs
kchibisov: I have no idea, but I just followed the old code here for consistency. I feel like we can use `bool`.

alacritty_terminal/src/config/mod.rs
kchibisov: Alpha is f32 and we can't make it f64, so I've chosen to use this cast instead.

alacritty_terminal/src/config/mod.rs
kchibisov: It seems to work just fine?

alacritty/src/display.rs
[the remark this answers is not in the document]
kchibisov: ops, don't get your point initially, let me update it.

Issue #3440, opened 2020-03-10
Document escape sequence support


chrisduerr:
Alacritty supports many escape sequences and many others are not supported. Currently the only way to figure out what exactly is supported is to look at the source code itself.

This both makes it hard to understand why certain things aren't implemented yet and doesn't give users the opportunity to pick out escape sequences they might like to see to go ahead and implement them.

I think it would be valuable to have a full list with all escape sequences in our wiki together with their current status (implemented/todo/denied -> reason).

chrisduerr:
For escapes where we are not certain if we should implement them or not, it would be helpful to have a column in the escape sequence table to document if a maintainer is for or against adding an escape. That should make it pretty simple to rule out escapes quickly without requiring too much discussion about every single one.

If there are some disagreements, discussion can then start based off of that list, but to rule out some basic escapes that are not needed it would likely be a very helpful system.

nixpulvis:
Seems like there's pretty good reason to implement `OSC 4` at least. This will most likely superseed #1771 since that issue is specific to X11, and this escape is more generic. A tool could easily then read Xresources and send the appropriate escapes.

nixpulvis:
I've started the list [here](https://github.com/alacritty/alacritty/wiki/ANSI-References#csi-control-sequence-introducer---esc-). I'm not 100% sure how the voting should really work TBH, but we can add the existing stuff in there, and start with some obvious missing escapes.

chrisduerr:
I've decided to move it to a separate page, since the escape sequence reference is mostly unrelated to Alacritty's support for them. It can now be found under https://github.com/alacritty/alacritty/wiki/Escape-Sequence-Support.

I think for voting it might be good to just have the name of the voter with a red/yellow/green symbol besides it.

The only major change I've made to your proposal is that I've removed the `Justification` field, since different people might have different justifications and it doesn't seem super maintainable to add a thousand different links to various issue comments.

Instead, I've added a `tracking issue` column. If someone is interested in outlining their ideas, a tracking issue should be created for the escape with their reasoning about it. I think that would make things a lot easier since it forces us to keep all the discussion in one place.

chrisduerr:
Also just to make sure this issue can be closed at some point, I'd like to have a very clear definition of when it is completed.

I don't think it's reasonable to expect every single escape to be documented with a decision, so I'd say once all escape sequences are documented that Alacritty currently supports, this issue should be done. I think that's a fair minimum of what Alacritty should document.

[...]

PR #3553, opened 2020-04-03
Adding use_xrdb to config and support for xrdb on colors

### LINKED ISSUE #1771: Load colors from Xresources.
Title is pretty self explanatory. It would be nice to be able to load colors from the Xresources file.

jhonrocha:
Fixes #1771 **Load colors from Xresources.**

## Proposed Changes
  - **alacritty.yml**: Adds a new config option on colors: `use_xrdb: bool`
  - **alacritty_terminal/Cargo.toml**: Adds the external (and very new) crate [xrdb](https://crates.io/crates/xrdb) 
  - **alacritty_terminal/src/config/colors.rs**: Adds new field for deserialize `use_xrdb`
  - **alacritty_terminal/src/term/color.rs**: Adds a new private function
    `xrdb_colors -> bool` that will check for the `use_xrdb` option and try to
    load colors using the xrdb lib. If any of the previous is false or fails, it
    returns false, which leads to previus normal flow on fill_named. Otherwise
    it adds the expected xrdb to NamedColor.

## New dependencies?
Yes, [xrdb](https://crates.io/crates/xrdb).

### Warning:
- I am still starting to learn Rust
- This is my first serious PR ever ;)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -72,6 +72,7 @@ dependencies = [
  "vte 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "winpty 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "xrdb 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -2133,6 +2134,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "x11"
+version = "2.18.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "libc 0.2.67 (registry+https://github.com/rust-lang/crates.io-index)",
+ "pkg-config 0.3.17 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "x11-clipboard"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2169,6 +2179,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 name = "xml-rs"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "xrdb"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "libc 0.2.67 (registry+https://github.com/rust-lang/crates.io-index)",
+ "scopeguard 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "x11 2.18.2 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
 name = "yaml-rust"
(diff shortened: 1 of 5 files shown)


jhonrocha:
I have tested it with this (gruvbox):
```
! hard contrast: *background: #1d2021
*background: #282828
! soft contrast: *background: #32302f
*foreground: #ebdbb2
! Black + DarkGrey
*color0:  #282828
*color8:  #928374
! DarkRed + Red
*color1:  #cc241d
*color9:  #fb4934
! DarkGreen + Green
*color2:  #98971a
*color10: #b8bb26
! DarkYellow + Yellow
*color3:  #d79921
*color11: #fabd2f
! DarkBlue + Blue
*color4:  #458588
*color12: #83a598
! DarkMagenta + Magenta
*color5:  #b16286
*color13: #d3869b
! DarkCyan + Cyan
*color6:  #689d6a
*color14: #8ec07c
! LightGrey + White
*color7:  #a89984
*color15: #ebdbb2
```

And with pywal. Running `xrdb -query all` and checking the specified values seems correct.

jhonrocha:
Hold on. I just deleted my config file and replaced it again. Now I am able to replicate the problem. Let me debug.

jhonrocha:
Nope. It is working! I got an error while uncommenting section #colors and #use_xrdb. Remember to set it to true.

jhonrocha:
Also, if you change .Xresources, remember to merge it:
```
 xrdb -merge ~/.Xresources
```

chrisduerr:
Oh I see, you require all colors to be present, which obviously wouldn't work out well.

Could you go into more detail why you think loading colors from `.Xresources` would be worth it for Alacritty? It seems like with most platforms not supporting this file, it's an extremely niche usecase.

jhonrocha:
Okay, so ... not all colors are necessary. Except of background and foreground, all the others have default values:
```
self[ansi::NamedColor::Black] = if let Some(c) = xrdb_color.get(0) { *c } else { colors.normal().black };
```
The reason for this PR is to have better color integration with the system. This would allow tools like pywal, which set those colors from images (wallpapers), to integrate natively with Alacritty. 

There are many tools and terminals already using those colors: https://github.com/dylanaraps/pywal/wiki/Customization#konsole
There is even a project to build Alacritty config file automatically reading them: https://github.com/egeesin/alacritty-color-export

chrisduerr:
> This would allow tools like pywal, which set those colors from images (wallpapers), to integrate natively with Alacritty.

Pywal is not a good argument for this, since it's partially just a limitation in an external tool. If it can write to .Xresources, there is no reason why it wouldn't be able to write to Alacritty's config.

And there's also escapes to support changing terminal colors. We do not yet support those fully, but those can be used by pywal too (and already are iirc).

So that seems like a strange reason to support .Xresources. At that point we'd basically be arguing to integrate an arbitrary config file from a third party.

nixpulvis:
@jhonrocha I'm personally looking for this functionality, however I'm thinking a separate tool makes a lot more sense. Perhaps you'd be interested in diving deeper into the code by [implementing `OSC 4`](https://github.com/alacritty/alacritty/issues/3440#issuecomment-608902139)?

It's true we really try to avoid platform specific features, especially when more generic solutions could be possible. Having to consider two separate configs to decide what colors will be loaded, on a platform by platform basis, is potentially confusing for users.

jhonrocha:
@nixpulvis Thanks for the advice. I agree it is too specific, only usable for Xorg users, even though they are most of Linux users. I think I got the wrong felling from issue 1771, it seemed to me this was something the project was already willing to do. I am migrating from St and Urxvt and both of them have this option, probably that is why I tough this way.

Anyway, I will colse this PR since the project is moving into escapes direction. And I will keep my fork for my specific usage, since I don't like the idea to add more initial load time by using escape codes and external tools.

MuntashirAkon:
@jhonrocha I think you have enough arguments to create a separate implementation of this awesome application. I too have find this project (thus far) interesting (coming from st, xst, Urxvt which support xrdb), and I think many Linux users will be interested in loading colors from xrdb as most popular Linux terminals do.

jhonrocha:
@MuntashirAkon Thank you! In the next days, I will check and try to compile on a Mac and Windows, to see what is going on. I am still very novice on Rust and I am sure there is plenty to fix on this PR.
In any case, I truly appreciate this project and how the things are going on here, I just missed this specific option. Perhaps there is a way to make a patch from this, something the user can apply to the source code. I will check that too.

PR #3559, opened 2020-04-04
Use frame callbacks instead of vsync on Wayland

### LINKED ISSUE #2851: alacritty doesn't close unless it is visible on sway
Which operating system does the issue occur on? sway on Arch Linux

If on linux, are you using X11 or Wayland? Wayland

How to reproduce: Run a command like "sleep 3 && exit" and then either:
- switch to another workspace
- open another window using a tabbed layout

You'll see alacritty doesn't close itself after 3 seconds, like it should, until it is visible again.

kchibisov:
Fixes #2851.

This PR should work fine on Wayland and solve some serious problems like freeze when window is offscreen or some glitches on resize. 

>~However it doesn't work nicely with winit redraw events, so it should likely be done there. Anyway I'll publish it so some folks could run it.~

>~I'm also not sure how well it plays with input latency, since it could be worse in theory, but repeat rate on my system seems way more responsive with it.~

Fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fallback to `LC_CTYPE=UTF-8` on macOS without valid system locale
 - Resize lag on launch under some X11 wms
 - Increased input latency due to vsync behavior on X11
+- Freeze when application is invisible on Wayland
 
 ## 0.4.2
 
--- a/alacritty/Cargo.toml
+++ b/alacritty/Cargo.toml
@@ -41,7 +41,7 @@ dirs = "2.0.2"
 
 [target.'cfg(not(any(target_os="windows", target_os="macos")))'.dependencies]
 x11-dl = "2"
-wayland-client = { version = "0.25", features = ["dlopen"] }
+wayland-client = { version = "0.26", features = ["dlopen"] }
 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3.7", features = ["impl-default", "wincon"]}
--- a/alacritty/Cargo.toml
+++ b/alacritty/Cargo.toml
@@ -41,6 +41,7 @@ dirs = "2.0.2"
 
 [target.'cfg(not(any(target_os="windows", target_os="macos")))'.dependencies]
 x11-dl = "2"
+wayland-client = { version = "0.25", features = ["dlopen"] }
 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3.7", features = ["impl-default", "wincon"]}
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -51,16 +51,16 @@ use crate::window::{self, Window};
 
 #[derive(Debug)]
 pub enum Error {
-    /// Error with window management
+    /// Error with window management.
     Window(window::Error),
 
-    /// Error dealing with fonts
+    /// Error dealing with fonts.
     Font(font::Error),
 
-    /// Error in renderer
+    /// Error in renderer.
     Render(renderer::Error),
 
-    /// Error during buffer swap
+    /// Error during buffer swap.
     ContextError(glutin::ContextError),
 }
 
@@ -131,11 +131,11 @@ pub struct Display {
 
 impl Display {
     pub fn new(config: &Config, event_loop: &EventLoop<Event>) -> Result<Display, Error> {
-        // Guess DPR based on first monitor
+        // Guess DPR based on first monitor.
         let estimated_dpr =
             event_loop.available_monitors().next().map(|m| m.scale_factor()).unwrap_or(1.);
 
-        // Guess the target window dimensions
+        // Guess the target window dimensions.
         let metrics = GlyphCache::static_metrics(config.font.clone(), estimated_dpr)?;
         let (cell_width, cell_height) = compute_cell_size(config, &metrics);
         let dimensions =
@@ -157,10 +157,10 @@ impl Display {
             }
         }
 
-        // Create the window where Alacritty will be displayed
+        // Create the window where Alacritty will be displayed.
         let size = dimensions.map(|(width, height)| PhysicalSize::new(width, height));
 
-        // Spawn window
+        // Spawn window.
         let mut window = Window::new(
             event_loop,
             &config,
@@ -172,10 +172,10 @@ impl Display {
         let dpr = window.scale_factor();
         info!("Device pixel ratio: {}", dpr);
 
-        // get window properties for initializing the other subsystems
+        // get window properties for initializing the other subsystems.
         let viewport_size = window.inner_size();
 
-        // Create renderer
+        // Create renderer.
         let mut renderer = QuadRenderer::new()?;
 
         let (glyph_cache, cell_width, cell_height) =
@@ -194,7 +194,7 @@ impl Display {
                 window.set_inner_size(PhysicalSize::new(width, height));
             }
         } else if config.window.dynamic_padding {
-            // Make sure additional padding is spread evenly
+            // Make sure additional padding is spread evenly.
             padding_x = dynamic_padding(padding_x, viewport_size.width as f32, cell_width);
             padding_y = dynamic_padding(padding_y, viewport_size.height as f32, cell_height);
         }
@@ -205,7 +205,7 @@ impl Display {
         info!("Cell Size: {} x {}", cell_width, cell_height);
         info!("Padding: {} x {}", padding_x, padding_y);
 
-        // Create new size with at least one column and row
+        // Create new size with at least one column and row.
         let size_info = SizeInfo {
             dpr,
             width: (viewport_size.width as f32).max(cell_width + 2. * padding_x),
@@ -216,7 +216,7 @@ impl Display {
             padding_y,
         };
 
-        // Update OpenGL projection
+        // Update OpenGL projection.
         renderer.resize(&size_info);
 
         // Call `clear` before showing the window, to make sure the surface is initialized.
@@ -242,10 +242,10 @@ impl Display {
 
         window.set_visible(true);
 
-        // Set window position
+        // Set window position.
         //
         // TODO: replace `set_position` with `with_position` once available
-        // Upstream issue: https://github.com/rust-windowing/winit/issues/806
+        // Upstream issue: https://github.com/rust-windowing/winit/issues/806.
         if let Some(position) = config.window.position {
             window.set_outer_position(PhysicalPosition::from((position.x, position.y)));
         }
@@ -283,7 +283,7 @@ impl Display {
         let font = config.font.clone();
         let rasterizer = font::Rasterizer::new(dpr as f32, config.font.use_thin_strokes())?;
 
-        // Initialize glyph cache
+        // Initialize glyph cache.
         let glyph_cache = {
             info!("Initializing glyph cache...");
             let init_start = Instant::now();
@@ -306,7 +306,7 @@ impl Display {
         Ok((glyph_cache, cw, ch))
     }
 
-    /// Update font size and cell dimensions
+    /// Update font size and cell dimensions.
     fn update_glyph_cache(&mut self, config: &Config, font: Font) {
         let size_info = &mut self.size_info;
         let cache = &mut self.glyph_cache;
@@ -315,7 +315,7 @@ impl Display {
             let _ = cache.update_font_size(font, size_info.dpr, &mut api);
         });
 
-        // Update cell size
+        // Update cell size.
         let (cell_width, cell_height) = compute_cell_size(config, &self.glyph_cache.font_metrics());
         size_info.cell_width = cell_width;
         size_info.cell_height = cell_height;
@@ -338,7 +338,7 @@ impl Display {
         config: &Config,
         update_pending: DisplayUpdate,
     ) {
-        // Update font size and cell dimensions
+        // Update font size and cell dimensions.
         if let Some(font) = update_pending.font {
             self.update_glyph_cache(config, font);
         } else if update_pending.cursor {
@@ -348,18 +348,18 @@ impl Display {
         let cell_width = self.size_info.cell_width;
         let cell_height = self.size_info.cell_height;
 
-        // Recalculate padding
+        // Recalculate padding.
         let mut padding_x = f32::from(config.window.padding.x) * self.size_info.dpr as f32;
         let mut padding_y = f32::from(config.window.padding.y) * self.size_info.dpr as f32;
 
-        // Update the window dimensions
+        // Update the window dimensions.
         if let Some(size) = update_pending.dimensions {
-            // Ensure we have at least one column and row
+            // Ensure we have at least one column and row.
             self.size_info.width = (size.width as f32).max(cell_width + 2. * padding_x);
             self.size_info.height = (size.height as f32).max(cell_height + 2. * padding_y);
         }
 
-        // Distribute excess padding equally on all sides
+        // Distribute excess padding equally on all sides.
         if config.window.dynamic_padding {
             padding_x = dynamic_padding(padding_x, self.size_info.width, cell_width);
             padding_y = dynamic_padding(padding_y, self.size_info.height, cell_height);
@@ -370,29 +370,29 @@ impl Display {
 
         let mut pty_size = self.size_info;
 
-        // Subtract message bar lines from pty size
+        // Subtract message bar lines from pty size.
         if let Some(message) = message_buffer.message() {
             let lines = message.text(&self.size_info).len();
             pty_size.height -= pty_size.cell_height * lines as f32;
         }
 
-        // Resize PTY
+        // Resize PTY.
         pty_resize_handle.on_resize(&pty_size);
 
-        // Resize terminal
+        // Resize terminal.
         terminal.resize(&pty_size);
 
-        // Resize renderer
+        // Resize renderer.
         let physical = PhysicalSize::new(self.size_info.width as u32, self.size_info.height as u32);
         self.window.resize(physical);
         self.renderer.resize(&self.size_info);
     }
 
-    /// Draw the screen
+    /// Draw the screen.
     ///
     /// A reference to Term whose state is being drawn must be provided.
     ///
-    /// This call may block if vsync is enabled
+    /// This call may block if vsync is enabled.
     pub fn draw<T>(
         &mut self,
         terminal: MutexGuard<'_, Term<T>>,
@@ -418,11 +418,11 @@ impl Display {
             None
         };
 
-        // Update IME position
+        // Update IME position.
         #[cfg(not(windows))]
         self.window.update_ime_position(&terminal, &self.size_info);
 
-        // Drop terminal as early as possible to free lock
+        // Drop terminal as early as possible to free lock.
         drop(terminal);
 
         self.renderer.with_api(&config, &size_info, |api| {
@@ -432,20 +432,20 @@ impl Display {
         let mut lines = RenderLines::new();
         let mut urls = Urls::new();
 
-        // Draw grid
+        // Draw grid.
         {
             let _sampler = self.meter.sampler();
 
             self.renderer.with_api(&config, &size_info, |mut api| {
-                // Iterate over all non-empty cells in the grid
+                // Iterate over all non-empty cells in the grid.
                 for cell in grid_cells {
-                    // Update URL underlines
+                    // Update URL underlines.
                     urls.update(size_info.cols().0, cell);
 
-                    // Update underline/strikeout
+                    // Update underline/strikeout.
                     lines.update(cell);
 
-                    // Draw the cell
+                    // Draw the cell.
                     api.render_cell(cell, glyph_cache);
                 }
             });
@@ -453,7 +453,7 @@ impl Display {
 
         let mut rects = lines.rects(&metrics, &size_info);
 
-        // Update visible URLs
+        // Update visible URLs.
         self.urls = urls;
         if let Some(url) = self.urls.highlighted(config, mouse, mods, mouse_mode, selection) {
             rects.append(&mut url.rects(&metrics, &size_info));
@@ -471,14 +471,14 @@ impl Display {
             }
         }
 
-        // Highlight URLs at the vi mode cursor position
+        // Highlight URLs at the vi mode cursor position.
         if let Some(vi_mode_cursor) = vi_mode_cursor {
             if let Some(url) = self.urls.find_at(vi_mode_cursor.point) {
                 rects.append(&mut url.rects(&metrics, &size_info));
             }
         }
 
-        // Push visual bell after url/underline/strikeout rects
+        // Push visual bell after url/underline/strikeout rects.
         if visual_bell_intensity != 0. {
             let visual_bell_rect = RenderRect::new(
                 0.,
@@ -494,19 +494,19 @@ impl Display {
         if let Some(message) = message_buffer.message() {
             let text = message.text(&size_info);
 
-            // Create a new rectangle for the background
+            // Create a new rectangle for the background.
             let start_line = size_info.lines().0 - text.len();
             let y = size_info.cell_height.mul_add(start_line as f32, size_info.padding_y);
             let message_bar_rect =
                 RenderRect::new(0., y, size_info.width, size_info.height - y, message.color(), 1.);
 
-            // Push message_bar in the end, so it'll be above all other content
+            // Push message_bar in the end, so it'll be above all other content.
             rects.push(message_bar_rect);
 
-            // Draw rectangles
+            // Draw rectangles.
             self.renderer.draw_rects(&size_info, rects);
 
-            // Relay messages to the user
+            // Relay messages to the user.
             let mut offset = 1;
             for message_text in text.iter().rev() {
                 self.renderer.with_api(&config, &size_info, |mut api| {
@@ -520,11 +520,11 @@ impl Display {
                 offset += 1;
             }
         } else {
-            // Draw rectangles
+            // Draw rectangles.
             self.renderer.draw_rects(&size_info, rects);
         }
 
-        // Draw render timer
+        // Draw render timer.
         if config.render_timer() {
             let timing = format!("{:.3} usec", self.meter.average());
             let color = Rgb { r: 0xd5, g: 0x4e, b: 0x53 };
@@ -534,7 +534,7 @@ impl Display {
         }
 
         // Frame event should be requested before swaping buffers, since it requires surface
-        // `commit`, which is done by swap buffers under the hood
+        // `commit`, which is done by swap buffers under the hood.
         #[cfg(not(any(target_os = "macos", windows)))]
         self.request_frame(&self.window);
 
@@ -564,17 +564,17 @@ impl Display {
 
         let should_draw = self.window.should_draw.clone();
 
-        // Mark that window was drawn
+        // Mark that window was drawn.
         should_draw.store(false, Ordering::Relaxed);
 
-        // Request a new frame
+        // Request a new frame.
         surface.frame().quick_assign(move |_, _, _| {
             should_draw.store(true, Ordering::Relaxed);
         });
     }
 }
 
-/// Calculate padding to spread it evenly around the terminal content
+/// Calculate padding to spread it evenly around the terminal content.
 #[inline]
 fn dynamic_padding(padding: f32, dimension: f32, cell_dimension: f32) -> f32 {
     padding + ((dimension - 2. * padding) % cell_dimension) / 2.
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -526,19 +526,7 @@ impl Display {
         }
 
         #[cfg(not(any(target_os = "macos", windows)))]
-        {
-            // Register wayland frame callback
-            if let Some(surface) = self.window.wayland_surface() {
-                let proxy: Proxy<WlSurface> = unsafe { Proxy::from_c_ptr(surface as _) };
-                let attached = proxy.attach(self.wayland_event_queue.as_ref().unwrap().token());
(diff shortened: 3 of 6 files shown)


kchibisov:
This branch is kind of ready, CI is failing since Wayland client requires `rust-1.40.0`, and I'm in favor of updating to it tbh, since using old wayland-client is just a pain to some point.

michaelweiser:
I can confirm that when used with sway this PR solves for me:
- window titles not updating in tabbed view,
- the processing of backlogged updates upon becoming visible again causing lag, high CPU load and spurious visibility of obvisouly outdated output and,
- likely (to be confirmed by further testing) windows closing randomly upon becoming visible again.

No negative effects so far with master and when backported to 0.4.2. Thanks!

DarkDefender:
I've tried this one out and it seems to work nicely!

One thing it didn't fix though is when you fullscreen alacritty in sway (super+f), the texture buffer seems to become corrupted and it will only become normal again when you exit fullscreen.
I guess this is an other issue, but I'm just asking here to make sure.

kchibisov:
>One thing it didn't fix though is when you fullscreen alacritty in sway (super+f), the texture buffer seems to become corrupted and it will only become normal again when you exit fullscreen.
I guess this is an other issue, but I'm just asking here to make sure.

Could you explain this issue a bit more. It has nothing to do with my PR and my PR can't fix it, I guess, but I don't understand the issue with fullscreen, and I can't repro it. Maybe screenshot could help.

DarkDefender:
Ok, here is a screenshot. As it only happens for me with alacritty (it is fine though it I use the Xorg backend via Xwayland), I think that this might have something to do with how the drawing is handled in the wayland backend. (And that is why I posted here to make sure even though it might not be related)

al_error

kchibisov:
It seems like a GPU/driver bug to me, we clear the entire screen on each redraw.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Packaging
 
 - Minimum Rust version has been bumped to 1.41.0
+- Minimum Rust version has been bumped to 1.40.0
```
lbonn, line 12: This line is redundant now.

kchibisov: Yeah, I'll rebase it properly, once I see review on a horizon. The problem is that this patch will require some more work to be included in upstream.

alacritty/Cargo.toml
```diff
@@ -41,6 +41,7 @@ dirs = "2.0.2"
 
 [target.'cfg(not(any(target_os="windows", target_os="macos")))'.dependencies]
 x11-dl = "2"
+wayland-client = { version = "0.25", features = ["dlopen"] }
```
chrisduerr, line 44: Are there any problems with using multiple different `wayland-client` versions?

Just out of curiosity, do you know why winit isn't updated yet? I'd assume 0.25.0 is the minimum for frame callbacks?

I'd just like to make an effort to not duplicate dependencies unnecessarily.

alacritty/src/event.rs
```diff
@@ -472,6 +498,15 @@ impl<N: Notify + OnResize> Processor<N> {
...
             }
 
+            #[cfg(not(any(target_os = "macos", windows)))]
+            {
+                if event_loop.is_wayland()
+                    && !self.display.window.should_draw.load(Ordering::Relaxed)
+                {
+                    return;
+                }
+            }
```
chrisduerr, line 508: What you seem to do in this PR is basically just ignore redraw opportunities until Wayland says it's ready, but I'm a bit concerned if that's the ideal approach.

At least in theory, that should open us for potential latency issues. If we miss the vsync just by a hair and go into the next event loop, every event we receive leaves us open to miss the vsync window. Which would mean that we'd probably miss the vsync and then latency is increased by one frame whenever this happens.

It's possible that this still improves things in general, but this seems like a bit of a design flaw? Have you considered this and potential solutions for it?

alacritty/src/event.rs
```diff
@@ -411,8 +413,32 @@ impl<N: Notify + OnResize> Processor<N> {
                 GlutinEvent::RedrawEventsCleared => {
                     *control_flow = ControlFlow::Wait;
 
-                    if event_queue.is_empty() {
-                        return;
+                    let event_queue_empty = event_queue.is_empty();
```
chrisduerr, line 416: Seems like there's no need to extract this as a variable? Since it's just conditionally compiled in, it's not executed twice at runtime anyways.

alacritty/src/event.rs
```diff
@@ -411,8 +413,32 @@ impl<N: Notify + OnResize> Processor<N> {
...
+
+                    #[cfg(not(any(target_os = "macos", windows)))]
```
chrisduerr, line 418: It might just be because of the terrible formatting which can be fixed separately, but these are relatively big platform-dependent blocks. Might be worth extracting this into a separate method?

alacritty/src/display.rs
```diff
@@ -206,6 +218,11 @@ impl Display {
             // actually draw something into it.
```
chrisduerr, line 218: The comments on this block should be updated and things should be moved closer to where the `swap_buffers` is actually performed. Since it just seems a bit confusing otherwise.

alacritty/src/event.rs
```diff
@@ -411,8 +413,32 @@ impl<N: Notify + OnResize> Processor<N> {
...
+                                .as_mut()
+                                .unwrap()
+                                .dispatch_pending(&mut (), |_, _, _| {})
+                                .unwrap();
+                            num_dispatched == 0
+                        } else {
+                            true
+                        };
+
+                        if event_queue_empty && wayland_queue_empty {
```
chrisduerr, line 433: Why can't we skip processing if the wayland queue is not empty? We don't need to render on every frame callback, right? So shouldn't matter?

alacritty/src/event.rs
kchibisov: >At least in theory, that should open us for potential latency issues. If we miss the vsync just by a hair and go into the next event loop, every event we receive leaves us open to miss the vsync window. Which would mean that we'd probably miss the vsync and then latency is increased by one frame whenever this happens.

You're relying on when to draw already, since vsync doesn't exists on Wayland at all. Compositors are scheduling frames, in that case they are sending you event to redraw in some before the next vblank, e.g.  max_render_time on sway, weston is using 7ms (Could be changed). If you can't render under the time they are giving to you that's really unfortunate.

Ideally you just put your rendering code into a callback, however that's not something possible sometimes, and some apps are doing the same approach, like Firefox.

The only way you can miss redraw in that case, if you really can't render under the time your compositor is giving to you, but it's not our problem, if it can't schedule frames properly. Compositor can collect statics on how you reply to its redraw requests and tune its scheduling of frame callbacks for each client, so if you start to draw under 13ms or you'll be missing times, it can start sending frame events earlier. So I don't see any issue in doing it this way, and if client is missing redraw, it's compositor's fault.

I've been running this PR for a long time already on slow and a bit faster system, and latency on it is way lower. I've tested weston and sway.

>Have you considered this and potential solutions for it?

Up to compositor.

alacritty/src/event.rs
kchibisov: This is intentional, since without it we were missing redraw in some edge cases. IIRC it could be true for obscure windows and something like that, since compositor could start sending you events to redraw before you're going to switch into that window. This code exists because of a bug of missing redraws in some corner cases, that's all.

alacritty/Cargo.toml
kchibisov: >Are there any problems with using multiple different wayland-client versions?

I don't want to use `wayland-client` winit is using, because I don't understand it at all and structure code with it is just hard even for such tasks. There're no problems, but one versions has good API, but the other one is just hard to use here.

>Just out of curiosity, do you know why winit isn't updated yet? I'd assume 0.25.0 is the minimum for frame callbacks?

Because it's hard to update to new wayland client for winit, however there's some ongoing work https://github.com/rust-windowing/winit/pull/1544 . Also winit is using SCTK, but we're using wayland-client here. Our clipboard will be updates to a new wayland-client soon as well though.

alacritty/src/event.rs
kchibisov: Also, one thing why it's required. You're processing all your events on `RedrawEventsCleared`, so somtimes it happens that you're finished them to early, and compositor hasn't send you frame event, but you won't get any new event, however once compositor will send you frame event, winit will wakeup and send you `RedrawEventsCleared`, because winit reading in a blocking way and has some kind of a `main` queue. if we dispatched something on this queue it means that we got frame event,  but since we have a situation, when you've already processes all events before frame event if you just check pending events you'll skip the redraw. The easiest way to repro it is moving with high repeat rate, since sometimes the last redraw is being missed.

alacritty/src/event.rs
kchibisov: I've updated the mentioned code, however I've kept the variable to reduce the amount of conditional code and made it more linear for reading.

alacritty/src/event.rs
kchibisov: It makes sense to extract it, I agree.

alacritty/src/display.rs
kchibisov: Done.

alacritty/Cargo.toml
kchibisov: >Since clipboard and winit are exchanging data types I think it's required for winit to also be bumped though.

This is not true, they exchange the `C` data type, which is directly related to `libwayland-client` and has nothing to do with `wayland-client`, so it's fine to mix them.

alacritty/src/event.rs
chrisduerr: To summarize our discussion on IRC:

Sway's latency is delayed by an entire vblank interval by default, but can be lowered to any value in the config according to the documentation:

https://github.com/swaywm/sway/blob/58703e4f24ae2738f9459c0c092f41a827be007e/sway/sway.5.scd#L188-L202 

So this means that we can't really miss the vblank, since we have up to an entire frame ahead of it and do not need to render immediately after the callback came in. This is also necessary for us since there might be pending resize events which need to be processed.

Of course if a compositor chose to, they could break this. But that's a compositor's problem not the application's. Sway plays it safe here by basically acting like our X backend by default, but giving the user the choice to optimize it themselves.

So this should give us a great balance of leaving as much time as possible before rendering to reduce input latency and making it in time for the next vblank.

alacritty/src/display.rs
```diff
@@ -118,6 +124,9 @@ pub struct Display {
     renderer: QuadRenderer,
     glyph_cache: GlyphCache,
     meter: Meter,
+
+    #[cfg(not(any(target_os = "macos", windows)))]
+    pub wayland_event_queue: Option<EventQueue>,
```
chrisduerr, line 129: Could we move this up below `highlighted_url`, since it's also `pub`?

alacritty/src/display.rs
```diff
@@ -192,20 +201,28 @@ impl Display {
         // Update OpenGL projection
         renderer.resize(&size_info);
 
-        // Clear screen
+        // We should call `clear` when window is offscreen, so when `window.show()` happens it
+        // would be with background color instead of uninitialized surface.
```
chrisduerr, line 205: I'm not sure what you mean by this? Are you saying this is an outstanding TODO for Wayland? Or are you documenting the existing code?

I'd assume you're documenting the existing code, so I propose some minor changes to make that more clear:

```suggestion
        // Call `clear` before showing the window, to make sure the surface is initialized.
```

alacritty/src/display.rs
```diff
@@ -192,20 +201,28 @@ impl Display {
...
         {
-            // On Wayland we can safely ignore this call, since the window isn't visible until you
-            // actually draw something into it.
+            // Setup and perform platform specific helpers
             if event_loop.is_x11() {
+                // On Wayland we can safely ignore this call, since the window isn't visible until
+                // you actually draw something into it.
                 window.swap_buffers()
+            } else {
+                // Initialize Wayland event queue, to handle various Wayland related routines.
```
chrisduerr, line 222: Do we really handle various Wayland related routines? Seems like we just handle frame callbacks.

If this is made because future additions might also need it, what kind could be integrated here?

Even if we wanted to keep this comment generic, to make sure things don't change in the future rendering this comment outdated without us noticing, I'd remove the word `various` since it's usually kind of useless.

```suggestion
                // Initialize event queue, to handle Wayland callbacks.
```

I've replaced `related routines` with `callbacks` since it's a bit more specific. But there might be better terms for it. If wayland-client names them 'routines' then that would be the better choice.

alacritty/src/display.rs
```diff
@@ -498,8 +517,29 @@ impl Display {
...
+
         self.window.swap_buffers();
     }
+
+    /// Requst a new frame for a window on Wayland.
+    #[inline]
+    #[cfg(not(any(target_os = "macos", windows)))]
+    fn request_frame(&self, window: &Window) {
+        // Register wayland frame callback
+        if let Some(surface) = window.wayland_surface() {
```
chrisduerr, line 531: Could pull out the surface here in a match to reduce indentation. But the method's so short that it might not be worth it to add the extra two lines required to do so.

I'd say it's right on the edge where either thing would work fine, so I'll leave it up to you to decide.

alacritty/src/display.rs
```diff
@@ -498,8 +517,29 @@ impl Display {
...
+            let proxy: Proxy<WlSurface> = unsafe { Proxy::from_c_ptr(surface as _) };
+            let attached = proxy.attach(self.wayland_event_queue.as_ref().unwrap().token());
```
chrisduerr, line 533: The walyand event queue is obviously always present when the wayland surface is, but it doesn't seem ideal to handle things this way.

Do you know if it's possible for the wayland surface to ever change? Would it be an option to always keep the event queue attached for example and store that on Display?

alacritty/src/display.rs
```diff
@@ -498,8 +517,29 @@ impl Display {
...
+            let should_draw = self.window.should_draw.clone();
+
+            // Mark that window was drawn
+            should_draw.store(false, Ordering::Relaxed);
+            attached.frame().quick_assign(move |_, _, _| {
+                should_draw.store(true, Ordering::Relaxed);
+            });
```
chrisduerr, line 540: Might be worth to at least add a newline after the `store(false, ...)`, potentially even a new comment? Like this it kinda looks like the entire block is used to mark the window as rendered, but that doesn't seem to be correct.

alacritty/src/event.rs
```diff
@@ -384,6 +386,23 @@ impl<N: Notify + OnResize> Processor<N> {
         }
     }
 
+    /// Dispatch wayland event queue
```
chrisduerr, line 389:
```suggestion
    /// Dispatch wayland event queue.
```

alacritty/src/event.rs
chrisduerr: This still isn't ideal. I think it would get much, much cleaner if the method was renamed to something more like `event_queue_empty` and was implemented on all platforms checking for `event_queue.is_empty()` and the wayland status.

Then we can just do `if self.event_queue_empty() { return }`.

alacritty/src/event.rs
kchibisov: I'm not sure about this dot, we have a lot of code without dot in a top level comment.

alacritty/src/event.rs
chrisduerr: I'm not sure about all the points raised in your comments, but I think this basically boils down to `terminal.dirty` being `true`, while the event queue is empty.

So basically what we'd want to check is `event_queue_empty && (!redraw_requested || !terminal.dirty)`.

However, the problem with that is to check if the terminal is dirty we require locking it. If we lock it right here we need to drop it immediately and then re-lock it later just to render.

This means that in the worst case for the current solution, we'd lock it once and do all the following things for no reason:
 * create new processor
 * loop over all 0 events and process them
 * verify that there is indeed no display update pending

The worst case for changing the dirtiness would be that the terminal is locked twice with no other disadvantages. However the terminal lock is **very** expensive when there's currently a high PTY usage, however in that case there would be plenty of events lined up.

I feel like it might be worth it to lock the terminal, to save us a bunch of extra work. However I haven't benchmarked it.

alacritty/src/display.rs
kchibisov: It shouldn't change, I could try to store it on a window/display.

alacritty/src/event.rs
chrisduerr: I'm sure about it. The 'headline' should always have a dot. You can see that pretty clearly when looking at Rust's official documentation for modules:

https://doc.rust-lang.org/std/index.html#primitives

Every other place is just doing it wrong. Though I do understand how this can be easily done wrong since headlines aren't usually punctuated with a trailing dot.

alacritty/src/display.rs
chrisduerr: Might be worth a shot? Not sure if it'll clean things up.

alacritty/src/display.rs
```diff
@@ -192,19 +218,18 @@ impl Display {
...
             api.clear(background_color);
         });
 
-        // We should call `clear` when window is offscreen, so when `window.show()` happens it
-        // would be with background color instead of uninitialized surface.
         #[cfg(not(any(target_os = "macos", windows)))]
         {
-            // On Wayland we can safely ignore this call, since the window isn't visible until you
-            // actually draw something into it.
+            // Setup and perform platform specific helpers
```
chrisduerr, line 229:
```suggestion
```

I don't think this is needed. Especially since it's just one X11 thing.

alacritty/src/display.rs
```diff
@@ -192,19 +218,18 @@ impl Display {
...
             if event_loop.is_x11() {
+                // On Wayland we can safely ignore this call, since the window isn't visible until
+                // you actually draw something into it.
```
chrisduerr, line 232: Seems like due to the other changes the `else` branch was removed? In that case I'd move it in front of the `is_x11` again since that's a direct x11/Wayland reference.

alacritty/src/display.rs
```diff
@@ -498,8 +525,31 @@ impl Display {
             });
         }
 
+        #[cfg(not(any(target_os = "macos", windows)))]
+        self.request_frame(&self.window);
+
         self.window.swap_buffers();
```
chrisduerr, line 531: Is there a reason why we do this before `swap_buffers`? Intuitively I would think that we'd ask for the next frame once the current frame has been sent off.

alacritty/src/event.rs
```diff
@@ -381,16 +384,37 @@ impl<N: Notify + OnResize> Processor<N> {
             config,
             message_buffer,
             display,
+            event_queue: Vec::new(),
+        }
+    }
+
+    #[inline]
+    /// Return `true` if `event_queue` is empty, `true` otherwise.
```
chrisduerr, line 392: There's a much more performant way to implement what this comment suggests: `return true`.

```suggestion
    /// Return `true` if `event_queue` is empty, `false` otherwise.
```

That might be more appropriate?

alacritty/src/event.rs
```diff
@@ -381,16 +384,37 @@ impl<N: Notify + OnResize> Processor<N> {
...
+            // On Wayland we also dispatch and check its own event queue
+            let events_dispatched = wayland_event_queue
+                .dispatch_pending(&mut (), |_, _, _| {})
+                .expect("failed to dispatch event queue");
+
+            self.event_queue.is_empty() && events_dispatched == 0
         }
+
+        #[cfg(any(target_os = "macos", windows))]
+        self.event_queue.is_empty()
```
chrisduerr, line 410: I think we should just duplicate the entire method. I usually find it much easier to read than having different blocks inside of it.

That way we don't need the extra empty block just for the config attribute.

alacritty/src/event.rs
```diff
@@ -472,12 +496,21 @@ impl<N: Notify + OnResize> Processor<N> {
...
             }
 
+            #[cfg(not(any(target_os = "macos", windows)))]
+            {
+                if event_loop.is_wayland()
+                    && !self.display.window.should_draw.load(Ordering::Relaxed)
+                {
+                    return;
+                }
+            }
```
chrisduerr, line 506: Maybe add a comment to this? Would probably be generous. Just something short and sweet to explain what we're doing here.

Suggestion:

```
// Skip rendering on Wayland if the next vblank hasn't been reached.
```

Alternatively if 'vblank' is too weird:

```
// Skip rendering on Wayland if the next frame hasn't been requested yet.
```

Maybe that's a little easier to understand, more accessible? I'll leave it up to you.

alacritty/src/window.rs
```diff
@@ -135,13 +146,20 @@ fn create_gl_window(
...
     windowed_context: WindowedContext<PossiblyCurrent>,
     current_mouse_cursor: CursorIcon,
     mouse_visible: bool,
+    /// Wayland setting to control redrawing of the window, according to the frame callbacks.
+    #[cfg(not(any(target_os = "macos", windows)))]
+    pub should_draw: Arc<AtomicBool>,
+
+    /// Attached Wayland surface to create new frame events.
+    #[cfg(not(any(target_os = "macos", windows)))]
+    pub wayland_surface: Option<Attached<WlSurface>>,
```
chrisduerr, line 162: Move `pub` fields to the top? Maybe add one line of whitespace between these 'ugly' `cfg` fields and the normal ones, just to make it feel a little less cluttered.

alacritty/src/window.rs
```diff
@@ -135,13 +146,20 @@ fn create_gl_window(
...
+/// Wraps the underlying windowing library to provide a stable API in Alacritty.
 pub struct Window {
     windowed_context: WindowedContext<PossiblyCurrent>,
     current_mouse_cursor: CursorIcon,
     mouse_visible: bool,
+    /// Wayland setting to control redrawing of the window, according to the frame callbacks.
+    #[cfg(not(any(target_os = "macos", windows)))]
+    pub should_draw: Arc<AtomicBool>,
+
+    /// Attached Wayland surface to create new frame events.
```
chrisduerr, line 160: Is 'create' new frame events the correct term here? Wouldn't `handle` be better? We're not only using it to create them, right? They also come back 'from' that surface.

alacritty/src/window.rs
```diff
@@ -135,13 +146,20 @@ fn create_gl_window(
...
-/// A window which can be used for displaying the terminal
+/// A window which can be used for displaying the terminal.
 ///
-/// Wraps the underlying windowing library to provide a stable API in Alacritty
+/// Wraps the underlying windowing library to provide a stable API in Alacritty.
 pub struct Window {
     windowed_context: WindowedContext<PossiblyCurrent>,
     current_mouse_cursor: CursorIcon,
     mouse_visible: bool,
+    /// Wayland setting to control redrawing of the window, according to the frame callbacks.
```
chrisduerr, line 156:
```suggestion
    /// Flag tracking frame redraw requests from Wayland.
```

Maybe this is a bit better? My main goal is just to shorten this comment down a tiny bit, it seems very long for just a single flag which might make it harder to understand. Though I'm not certain we should change it.

alacritty/src/window.rs
```diff
@@ -165,6 +189,9 @@ impl Window {
         // Set OpenGL symbol loader. This call MUST be after window.make_current on windows.
         gl::load_with(|symbol| windowed_context.get_proc_address(symbol) as *const _);
 
+        #[cfg(not(any(target_os = "macos", windows)))]
+        let mut wayland_surface = None;
+
         // On X11, embed the window inside another if the parent ID has been set
```
chrisduerr, line 195: Move this comment to inside the `is_x11` check maybe?

alacritty/src/window.rs
```diff
@@ -175,10 +202,22 @@ impl Window {
             } else {
                 let theme = AlacrittyWaylandTheme::new(&config.colors);
                 windowed_context.window().set_wayland_theme(theme);
+
+                let surface = windowed_context.window().wayland_surface().unwrap();
+                let proxy: Proxy<WlSurface> = unsafe { Proxy::from_c_ptr(surface as _) };
+                wayland_surface = Some(proxy.attach(wayland_event_queue.as_ref().unwrap().token()));
```
chrisduerr, line 208: Two short comments on each of these blocks might be useful? Since they do two completely different things unrelated to each other.

alacritty/src/window.rs
kchibisov: you're requesting a frame event and run a callback when server send you it back on dispatch_pending on event queue. Not sure about handle because of that.

alacritty/src/display.rs
kchibisov: It's required by protocol to be that way, since frame event will take effect only after commit. Read wl_surface::frame from https://github.com/alacritty/alacritty/blob/6b45780f3afa23e9e4277bf43e6bf381172fa1ea/alacritty.yml#L410 . Commit is done by swap_buffers internally here. If you swap them it won't work.

alacritty/src/event.rs
kchibisov: It's about receiving frame callback from the server, though maybe requested is also fine.

alacritty/src/display.rs
```diff
@@ -508,6 +533,11 @@ impl Display {
             });
         }
 
+        // Frame event should be requested before swaping buffers, since it requires surface
+        // `commit`, which is done by swap buffers under the hood
```
chrisduerr, line 537:
```suggestion
        // Frame event should be requested before swaping buffers, since it requires surface
        // `commit`, which is done by swap buffers under the hood.
```

alacritty/src/event.rs
```diff
@@ -381,16 +384,54 @@ impl<N: Notify + OnResize> Processor<N> {
...
+            let events_dispatched = wayland_event_queue
+                .dispatch_pending(&mut (), |_, _, _| {})
+                .expect("failed to dispatch event queue");
+
+            self.event_queue.is_empty() && events_dispatched == 0
+        }
+
+        #[cfg(any(target_os = "macos", windows))]
+        self.event_queue.is_empty()
+    }
```
chrisduerr, line 428: I mean better safe than sorry, but I think we can assume that Rust won't just suddenly enter unreachable code that has been conditionally compiled out.

alacritty/src/event.rs
```diff
@@ -381,16 +384,54 @@ impl<N: Notify + OnResize> Processor<N> {
...
         }
     }
 
+    /// Return `true` if `event_queue` is empty, `false` otherwise.
+    #[inline]
+    #[cfg(not(any(target_os = "macos", windows)))]
+    fn event_queue_empty(&mut self) -> bool {
+        let wayland_event_queue = match self.display.wayland_event_queue.as_mut() {
+            Some(wayland_event_queue) => wayland_event_queue,
+            None => return self.event_queue.is_empty(),
```
chrisduerr, line 397: This is quite an important little line. Could we add a line of documentation to it that says that on X11 we just check the event queue?

Suggestion:

```
// Since frame callbacks do not exist on X11, just check the event queue.
```

alacritty/src/event.rs
```diff
@@ -381,16 +384,54 @@ impl<N: Notify + OnResize> Processor<N> {
...
+        };
+
+        // On Wayland we also dispatch and check its own event queue
```
chrisduerr, line 400: The `also` sounds a bit strange here, since there's nothing previously that it could refer to. I'd suggest the following:

```suggestion
        // Check for pending frame callbacks.
```

Alternatively:

```suggestion
        // Check for pending frame callbacks on Wayland.
```

alacritty/src/event.rs
```diff
@@ -472,12 +513,22 @@ impl<N: Notify + OnResize> Processor<N> {
                 );
             }
 
+            #[cfg(not(any(target_os = "macos", windows)))]
+            {
+                // Skip rendering on Wayland until we get frame event from compositor
```
chrisduerr, line 518:
```suggestion
                // Skip rendering on Wayland until we get frame event from compositor.
```

alacritty/src/window.rs
```diff
@@ -152,11 +171,17 @@ impl Window {
...
         config: &Config,
         size: Option<PhysicalSize<u32>>,
+        #[cfg(not(any(target_os = "macos", windows)))] wayland_event_queue: Option<&EventQueue>,
     ) -> Result<Window> {
         let window_builder = Window::get_platform_window(&config.window.title, &config.window);
+        // Disable vsync on Wayland
+        #[cfg(not(any(target_os = "macos", windows)))]
+        let vsync = !event_loop.is_wayland();
+        #[cfg(any(target_os = "macos", windows))]
+        let vsync = true;
```
chrisduerr, line 181:
```suggestion

        // Disable vsync on Wayland
        #[cfg(not(any(target_os = "macos", windows)))]
        let vsync = !event_loop.is_wayland();
        #[cfg(any(target_os = "macos", windows))]
        let vsync = true;
        
```

Just to clean up stuff a little bit.

alacritty/src/window.rs
```diff
@@ -165,20 +190,37 @@ impl Window {
...
+
         #[cfg(not(any(target_os = "macos", windows)))]
         {
             if event_loop.is_x11() {
+                // On X11, embed the window inside another if the parent ID has been set
                 if let Some(parent_window_id) = config.window.embed {
                     x_embed_window(windowed_context.window(), parent_window_id);
                 }
             } else {
+                // Apply CSD theme
```
chrisduerr, line 204:
```suggestion
                // Apply client side decorations theme.
```

Just to make it a little bit easier to understand for people who don't know about things.

alacritty/src/window.rs
```diff
@@ -165,20 +190,37 @@ impl Window {
...
                 let theme = AlacrittyWaylandTheme::new(&config.colors);
                 windowed_context.window().set_wayland_theme(theme);
+
+                // Attach surface to Alacritty's internal wayland queue to handle frame callbacks
```
chrisduerr, line 208:
```suggestion
                // Attach surface to Alacritty's internal wayland queue to handle frame callbacks.
```

PR #3561, opened 2020-04-05
Add a keybinding to open OSX Emoji & Symbols picker.

### LINKED ISSUE #1895: Can't bring up input menu
On macOS the keyboard shortcut to bring up the Input Menu to be able to use emoji, math symbols, etc, doesn't work. The keyboard shortcut is Ctrl + Command + Space.

nholstein:
This implements a call to OSX's [`orderfrontcharacterpalette()`](https://developer.apple.com/documentation/appkit/nsapplication/1428455-orderfrontcharacterpalette?language=objc) which is the aptly named function to display the standard on-screen keyboard for choosing emoji and of characters.

Similar functionality exists in an undocumented and [unexported function](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/macos/util/mod.rs#L121-L124
) in the `winit` crate. A cleaner solution is to plumb this command through the crates dependencies and avoid the additional explicit dependencies on Objective-C macros in this crate.

Fixes #1895

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Default Command+N keybinding for SpawnNewInstance on macOS
+- Add keybinding to open "Emoji & Symbols" input dialog on macOS
 - Vi mode for copying text and opening links
 - `CopySelection` action which copies into selection buffer on Linux/BSD
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,6 +19,7 @@ version = "0.5.0-dev"
 dependencies = [
  "alacritty_terminal 0.5.0-dev",
  "clap 2.33.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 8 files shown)


nholstein:
As alluded to in the commit message, I believe the cleaner way to implement this is to plumb the correct function call up through the `winit` crate. I opened this PR as a reference point for #1895. If this functionality is desired I can work on PRs for the underlying crates as well.

chrisduerr:
This definitely should go into winit, but I'm skeptical about this approach in general.

Adding a dedicated keybinding for emoji input seems a bit strange, when the concept of IME already exists. I'm also relatively certain that this does not exist on any other platform.

So I'm wondering why we should add a shortcut for the macOS emoji keyboard explicitly ourselves, when it's a macOS specific thing implemented in a questionable fashion.

chrisduerr:
Closing since there's no intention to merge it in this fashion.

ezzieyguywuf:
does anyone know if there is a PR in winit to resolve this? I've searched around but I'm not really sure what the b est search terms are

PR #3576, opened 2020-04-11
Log critical errors to stderr instead of stdout

kchibisov:


kchibisov:
>Why don't we use the error! macro?

Is there any benefit for it here though? You're crashing anyway, so most of the things error! macro does we don't need.

kchibisov:
>Is there any benefit for it here though? You're crashing anyway, so most of the things error! macro does we don't need.

Actually it forces the log to be saved at some special point, so it's kind of nice to have it.

PR #3578, opened 2020-04-11
Add the feature to build and copy the binary via Docker

mdedonno1337:
This feature allows the user to compile the alacritty binary in release
mode, without having to install all the rust compiler.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM rust
+
+RUN apt-get update && \
+    apt-get install -y cmake pkg-config libfreetype6-dev libfontconfig1-dev libxcb-xfixes0-dev python3
+
+WORKDIR /code
+
+ADD . .
+
+RUN RUSTFLAGS='-C link-arg=-s' cargo build --release
+
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ APP_DIR = $(RELEASE_DIR)/osx
 APP_BINARY = $(RELEASE_DIR)/$(TARGET)
 APP_BINARY_DIR  = $(APP_DIR)/$(APP_NAME)/Contents/MacOS
 
+PWD = $(pwd)
+
 DMG_NAME = Alacritty.dmg
 DMG_DIR = $(RELEASE_DIR)/osx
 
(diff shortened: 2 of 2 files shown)


kchibisov:
>This feature allows the user to compile the alacritty binary in release
mode, without having to install all the rust compiler.

And now they have to install docker to build alacritty, so it doesn't change anything and pulling rust compiler is easier than pulling docker. I'm also completely against this PR, since using docker for such thing seems wrong to me.

mdedonno1337:
On my side, I dont use nor code in Rust, and dont want to clug my system with build tools I dont need.
I use the same logic for everything on my system (I dont have a lot of tools, event the one I use everydays, like Python, R and all the other tools).

I was just thinking it could be interesting to have an alternative way to provide the binary, nothing more. The user dont "have to" install docker, but they can avoir installing rust if they already have docker (like myself).

In any case, if this PR is not interessting/does not provide a useful feature, please close this PR.

Thanks for the reading anyways.

kchibisov:
>I was just thinking it could be interesting to have an alternative way to provide the binary, nothing more. The user dont "have to" install docker, but they can avoid installing rust if they already have docker (like myself).

Also, you'll build for Ubuntu and not for anything else, so the value of this change is low, since what if I use something completely different or build with non default features. I don't see any issue with running `sudo apt install rust`, doing `cargo build --release` and then remove rust. If you don't want to have rust on your system, just package the thing once and use the package.

>On my side, I dont use nor code in Rust, and dont want to clug my system with build tools I dont need.

Same for me, I don't use docker and don't need it, and even if I had docker, I can't use your script, since it doesn't work on my system, because it's not Debian based system.

chrisduerr:
Also even if it is through docker, you're still installing Rust. You gain nothing.

mdedonno1337:
> Also even if it is through docker, you're still installing Rust. You gain nothing.

Well, no, I dont have rust on my runtime machine.
The aim is to produce the binary and copy it on my system, not running alacritty in docker.

Total build size, dependencies and the alacritty binary : 6Mb (and nothing dangling unused on my system)

PS: I dont want/need/ask the PR anymore, I have it on my fork anyways.

Issue #2800, opened 2019-09-15
Certain UTF-8 characters not showing


miguelmota:
Alacritty doesn't show certain UTF-8 characters, for example, like the Russian ruble sign `₽`

iTerm on the left, Alacritty on the right

<img width="1178" alt="Screen Shot 2019-09-14 at 7 23 44 PM" src="https://user-images.githubusercontent.com/168240/64915838-6634aa00-d725-11e9-9234-2339ac974b9f.png">

Using default config settings

macOS version 10.14.6

kchibisov:
What’s your output of `locale` in alacritty?

miguelmota:
Output of `locale` in Alacritty:

```bash
LANG="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_CTYPE="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_ALL=
```

@kchibisov

chrisduerr:
You seem to be using different fonts here. Does switching the Alacritty font to what you're using in iTerm fix this?

miguelmota:
Current font in iTerm

<img width="575" alt="Screen Shot 2019-09-15 at 5 51 15 PM" src="https://user-images.githubusercontent.com/168240/64930026-9477bf80-d7e1-11e9-9816-ff66d38c61b8.png">


Updated alacritty config in `/Users/mota/.config/alacritty/alacritty.yml`

```yaml
# Font configuration (changes require restart)
font:
  # Normal (roman) font face
  normal:
    # Font family
    #
    # Default:
    #   - (macOS) Menlo
    #   - (Linux) monospace
    #   - (Windows) Consolas
    family: DejaVu Sans Mono for Powerline

    # The `style` can be specified to pick a specific face.
    #style: Regular
    style: Book

  # Bold font face
  bold:
    # Font family
    #
    # If the bold family is not specified, it will fall back to the
    # value specified for the normal font.
    family: DejaVu Sans Mono Bold for Powerline

    # The `style` can be specified to pick a specific face.
    style: Bold

  # Italic font face
  italic:
    # Font family
    #
    # If the italic family is not specified, it will fall back to the
    # value specified for the normal font.
    family: DejaVu Sans Mono Oblique for Powerline

    # The `style` can be specified to pick a specific face.
[...]

Issue #2566, opened 2019-06-20
Inconsistent locale when running from command line vs from dock


RichoDemus:
Which operating system does the issue occur on?
Macos, Alacritty version 0.3.3, Swedish keyboard

I installed Alacritty via homebrew

If I start alacritty via the Icon in the dock/ Applications folder then the åäö characters doesn't work, å and ä does nothing and ö prints `(arg: 6)`

however, if I open Alacritty by running `/Applications/Alacritty.app/Contents/MacOS/alacritty` in iTerm2, then those characters works just fine

is env/dotfiles loaded differently or something?

kchibisov:
> is env/dotfiles loaded differently or something?

@RichoDemus You can actually compare them by typing `env`, seems like locale problem. Also, could you share your `shell` section from `alacritty.yml`? On macOS your `shell` should have arg `--login` to work properly. It has this flag by default, but if you've changed it somehow, it can break things.

RichoDemus:
Thanks 2 things:
I can definitely see env differences, it seems like when I launch it from iterm it enherits some stuff from there (I see a bunch of iterm things) but also things like `LC_CTYPE=UTF-8` is missing when it doesn't work and `SHLVL=2` is 1 when it doesn't work and 2 when it works

I did however try to just type `export SHLVL=2` and `export LC_CTYPE=UTF-8` but that didn't fix it

I'm not entirely familiar with how the config works, 
I've copy-pasted the alacritty.yml file from the repo into `~/.config/alacritty/alacritty.yml` is that then the only file that config is read from or are there multiple? (and they kinda override each other)

anyway, the whole shell section is commented out there, so I guess the `--login` thing should apply then

kchibisov:
@RichoDemus `SHLVL` indicate how much shells you are in. So if you open terminal, then `echo $SHLVL` then `bash` and then `echo $SHLVL` the variables would be different ( `n` and `n + 1`). So, setting this one manually is probably a bad idea.

`alacritty` can use only the one config at the time. However, you should ensure, that config in `~/.config/alacritty/alacritty.yml` is actually in use. You can do this by launching `/Applications/Alacritty.app/Contents/MacOS/alacritty --print-events` from `iterm` and searching for line `[INFO] Configuration loaded from` ( one of the first lines).

How did you changed `LC_CTYPE`, just typing export or you've set in your shell configuration? You should probably set `LC_CTYPE` in your `SHELL` configuration file.

chrisduerr:
[...]

PR #3608, opened 2020-04-18
bugfix: LC_CTYPE should default to utf8

casperstorm:
## Description
This is a proposed fix for https://github.com/alacritty/alacritty/issues/2800 and https://github.com/alacritty/alacritty/issues/2566.

## Issue
Alacritty was handling locale in a way which could lead to troubles when not setting any locales in your `bashrc`, `zshrc` or similar when running on Mac OS. Without setting any locales Alacritty would output the following:
```
$ locale
LANG=""
LC_COLLATE="C"
LC_CTYPE="C"
LC_MESSAGES="C"
LC_MONETARY="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_ALL=
```
This however gave problems with some `UTF-8` characters (like the nordic `æ`, `ø` and `å`). Looking at the output from the same command on both `Terminal.app` and `iTerm` shows a difference:

```
$ locale
LANG=""
LC_COLLATE="C"
LC_CTYPE="UTF-8"
LC_MESSAGES="C"
LC_MONETARY="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_ALL=
```
The key here is that they are both falling back to `LC_CTYPE="UTF-8"`. `LC_CTYPE` is defined as:

> LC_CTYPE
This category applies to classification and conversion of characters, and to multibyte and wide characters

[Source](https://www.gnu.org/software/libc/manual/html_node/Locale-Categories.html).

## Solution
I propose we introduce a fallback `LC_CTYPE` value. This value should be `UTF-8`. Right now the code is trying to calculate `LANG` based on `NSLocale` from Core Foundation. While doing this we use a trick where we try to set the given locale, to see if it's a valid locale. We do it this way:

First we save the original `LC_CTYPE`.
```rust
let original = setlocale(LC_CTYPE, null());
let saved_original = if original.is_null() {
    CString::new("").unwrap()
} else {
    CStr::from_ptr(original).to_owned()
};
```

Then we have to verify the locale we found to see if it's valid to propagate to `LANG`:
```rust
let modified = setlocale(LC_CTYPE, locale_ptr);
let result = if modified.is_null() { String::new() } else { locale_id };
setlocale(LC_CTYPE, saved_original.as_ptr());
```

And after this, the `result` will be passed to `LANG` like this: `env::set_var("LANG", &locale_id);`
The issue happens when we reset the `LC_CTYPE` on the last line, in the previous block. We reset it to the default value, hence we don't have a fallback solution.

I am suggesting that we still do the check to see if the locale is valid, but then not resetting back to the original `LC_CTYPE` but back to a `UTF-8`. Remember this is a case which only happens when the user is not providing any locales them self. If setting `LC_CTYPE` to something different in `alacritty.yml` it will still use that value. Ditto if overwriting locale in `bashrc`, `zshrc`, or similar it will use that.

Let me know what you guys think.

### DIFF
--- a/alacritty_terminal/src/locale.rs
+++ b/alacritty_terminal/src/locale.rs
@@ -61,7 +61,7 @@ pub fn set_locale_environment() {
     }
 }
 
-/// Determine system locale based on language and country code.
+// Determine system locale based on language and country code.
 fn system_locale() -> String {
     unsafe {
         let locale_class = Class::get("NSLocale").unwrap();
--- a/alacritty_terminal/src/locale.rs
+++ b/alacritty_terminal/src/locale.rs
@@ -61,8 +61,7 @@ pub fn set_locale_environment() {
     let locale_id = unsafe {
         // try setting `locale_id`
         let modified = setlocale(LC_CTYPE, locale_ptr);
-        let result = if modified.is_null() { String::new() } else { locale_id };
-        result
+        if modified.is_null() { String::new() } else { locale_id }
     };
 
     // Provide a default value for LC_CTYPE if needed
--- a/alacritty_terminal/src/locale.rs
+++ b/alacritty_terminal/src/locale.rs
@@ -61,7 +61,11 @@ pub fn set_locale_environment() {
     let locale_id = unsafe {
         // try setting `locale_id`
         let modified = setlocale(LC_CTYPE, locale_ptr);
-        if modified.is_null() { String::new() } else { locale_id }
+        if modified.is_null() {
+            String::new()
+        } else {
+            locale_id
+        }
     };
 
     // Provide a default value for LC_CTYPE if needed
--- a/alacritty_terminal/src/locale.rs
+++ b/alacritty_terminal/src/locale.rs
@@ -65,24 +65,20 @@ pub fn set_locale_environment() {
         result
     };
 
-    // Provide a default value for LC_CTYPE if needed
+    // provide a default value for LC_CTYPE if needed
     unsafe {
-        let orginal = setlocale(LC_CTYPE, null());
-        let ctype = if orginal.is_null() {
-            CString::new("C").unwrap()
+        // check if we already have LC_CTYPE set as a environment variable
+        if let Ok(ctype) = env::var("LC_CTYPE") {
+            setlocale(LC_CTYPE, CString::new(ctype).unwrap().as_ptr());
         } else {
-            CStr::from_ptr(orginal).to_owned()
+            // if current LC_CTYPE is either null or C, we overwrite it
+            let ctype = setlocale(LC_CTYPE, null());
+            if !ctype.is_null() && CStr::from_ptr(ctype).to_string_lossy() == "C" {
+                let locale = "UTF-8";
+                setlocale(LC_CTYPE, CString::new(locale).unwrap().as_ptr());
+                env::set_var("LC_CTYPE", locale);
+            }
         };
-
-        let ctype_string = ctype.into_string().unwrap();
-        match ctype_string.as_str() {
-            "C" => {
-                let utf8 = CString::new("UTF-8").unwrap();
-                setlocale(LC_CTYPE, utf8.as_ptr());
-                env::set_var("LC_CTYPE", utf8.into_string().unwrap());
-            },
-            _ => (),
-        }
     }
 
     env::set_var("LANG", &locale_id);
--- a/alacritty_terminal/src/locale.rs
+++ b/alacritty_terminal/src/locale.rs
@@ -68,7 +68,7 @@ pub fn set_locale_environment() {
     // Provide a default value for LC_CTYPE if needed
     unsafe {
         // Check if we already have LC_CTYPE set as a environment variable
-        if let Ok(_) = env::var("LC_CTYPE") {
+        if env::var("LC_CTYPE").is_ok() {
             setlocale(LC_CTYPE, CString::new("").unwrap().as_ptr());
         } else {
             let ctype = setlocale(LC_CTYPE, null());
--- a/alacritty_terminal/src/locale.rs
+++ b/alacritty_terminal/src/locale.rs
@@ -12,22 +12,62 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 #![allow(clippy::let_unit_value)]
-#![cfg(target_os = "macos")]
-use libc::{setlocale, LC_CTYPE};
+
 use std::env;
 use std::ffi::{CStr, CString};
 use std::os::raw::c_char;
-use std::ptr::null;
 use std::slice;
 use std::str;
 
+use libc::{setlocale, LC_ALL, LC_CTYPE};
+use log::debug;
 use objc::runtime::{Class, Object};
 
+const FALLBACK_LOCALE: &str = "UTF-8";
+
 pub fn set_locale_environment() {
-    let locale_id = unsafe {
+    let env_locale_c = CString::new("").unwrap();
+    let env_locale_ptr = unsafe { setlocale(LC_ALL, env_locale_c.as_ptr()) };
+    if !env_locale_ptr.is_null() {
+        let env_locale = unsafe { CStr::from_ptr(env_locale_ptr).to_string_lossy() };
+
+        // Assume `C` locale means unchanged, since it is the default anyways
+        if env_locale != "C" {
+            debug!("Using environment locale: {}", env_locale);
+            return;
+        }
+    }
+
+    let system_locale = system_locale();
+
+    // Set locale to system locale
+    let system_locale_c = CString::new(system_locale.clone()).expect("nul byte in system locale");
+    let lc_all = unsafe { setlocale(LC_ALL, system_locale_c.as_ptr()) };
+
+    // Check if system locale was valid or not
+    if lc_all.is_null() {
+        // Use fallback locale
+        debug!("Using fallback locale: {}", FALLBACK_LOCALE);
+
+        let fallback_locale_c = CString::new(FALLBACK_LOCALE).unwrap();
+        unsafe { setlocale(LC_CTYPE, fallback_locale_c.as_ptr()) };
+
+        env::set_var("LC_CTYPE", FALLBACK_LOCALE);
+    } else {
+        // Use system locale
+        debug!("Using system locale: {}", system_locale);
+
+        env::set_var("LC_ALL", system_locale);
+    }
+}
+
+/// Determine system locale based on language and country code.
+fn system_locale() -> String {
+    unsafe {
         let locale_class = Class::get("NSLocale").unwrap();
         let locale: *const Object = msg_send![locale_class, currentLocale];
         let _: () = msg_send![locale_class, release];
+
         // `localeIdentifier` returns extra metadata with the locale (including currency and
         // collator) on newer versions of macOS. This is not a valid locale, so we use
         // `languageCode` and `countryCode`, if they're available (macOS 10.12+):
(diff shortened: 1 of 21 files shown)


casperstorm:
> I'm not sure if just hardcoding the locale to UTF8 is the right choice here.
> 
> IIRC, the default terminal in macOS has a setting to override the locale, do I remember that correctly? Is that enabled by default?

Terminal.app and iTerm both has a GUI setting to change this value, which both defaults to UTF-8. I agree hardcoding it might not seem like the proper choice but in our case I feel like it could work - but again - im open for other suggestions!

alacritty_terminal/src/locale.rs
```diff
@@ -59,18 +58,13 @@ pub fn set_locale_environment() {
...
-        } else {
-            CStr::from_ptr(original).to_owned()
-        };
         // try setting `locale_id`
         let modified = setlocale(LC_CTYPE, locale_ptr);
         let result = if modified.is_null() { String::new() } else { locale_id };
-        // restore original setting
-        setlocale(LC_CTYPE, saved_original.as_ptr());
+        // restore LC_CTYPE to UTF-8
+        let utf8 = CString::new("UTF-8").unwrap();
```
kchibisov, line 65: If we decide to fallback to some arbitrary value, could we use `C.UTF-8` instead of `UTF-8`, because the later one isn't a valid locale if you ssh into some linux(glibc at least) box?

casperstorm: We could. My only reason to choose `UTF-8` was because that is how Terminal.app and iTerm does it.

casperstorm:
Seems like https://github.com/alacritty/alacritty/issues/3612 indeed would have benefited from this as well. How to we move it forward, @chrisduerr?

chrisduerr:
> Terminal.app and iTerm both has a GUI setting to change this value, which both defaults to UTF-8.

This is not true for example. Terminal.app defaults to `en_US.UTF-8`. It might set the text encoding to UTF-8 by default, but it seems like it still keeps around the locale.

> How to we move it forward

I still think this seems like a questionable approach. Blindly overriding an environment variable to something hardcoded in Alacritty seems like it's just bound to cause trouble.

I do understand that the user can always override it themselves, but that should be possible in the `env` section in the config for example. Not just the shellrc. We also shouldn't just blindly omit the locale part of the text encoding.

kchibisov:
>This is not true for example. Terminal.app defaults to en_US.UTF-8. It might set the text encoding to UTF-8 by default, but it seems like it still keeps around the locale.

Idk, for me locale in terminal.app is `UTF-8` for `LC_CTYPE` and `C` for the rest

>I still think this seems like a questionable approach. Blindly overriding an environment variable to something hardcoded in Alacritty seems like it's just bound to cause trouble.

Well, if you check that this value isn't set before launch, and you're getting some strange results (`C` locale), I feel like you're fine to override it to whatever value you want, since it's macOS. iterm2 is just using setting from user config, our mapping to that is setting in `env` section of the config, but we should provide some meaningful value, because it's the common practice to do so.

>We also shouldn't just blindly omit the locale part of the text encoding.

that's why I'm saying that if we got `C` locale for `LC_CTYPE`, we can use `C.UTF-8` instead, since it makes more sense to use such.

chrisduerr:
> Idk, for me locale in terminal.app is UTF-8 for LC_CTYPE and C for the rest

Ah, I suppose it depends on which of the locale variables you're talking about, you're absolutely right. Everything other than `LC_CTYPE` is set to `en_US.UTF-8`, while `LC_CTYPE` is `UTF-8`.

I've never used Terminal.app though, so I don't know how that plays with ssh.

> Well, if you check that this value isn't set before launch, and you're getting some strange results (C locale), I feel like you're fine to override it to whatever value you want

That is true, but as far as I can tell this patch doesn't check if the value is set before changing it?

> that's why I'm saying that if we got C locale for LC_CTYPE, we can use C.UTF-8 instead, since it makes more sense to use such.

That seems reasonable to me, yeah. Definitely better than just `UTF-8` at least.

kchibisov:
>That is true, but as far as I can tell this patch doesn't check if the value is set before changing it?

If it's not it should start doing so.

>I've never used Terminal.app though, so I don't know how that plays with ssh.

that's why I've suggested `C.UTF-8`, since `UTF-8` isn't a valid locale for `glibc` and likely `musl` based systems, so ssh don't like them.

@casperstorm so you should check if `LC_CTYPE` is set before changing it and set locale to `C.UTF-8` if it's not.

casperstorm:
> [...] you should check if `LC_CTYPE` is set before changing it and set locale to `C.UTF-8` if it's not.
If I set `LC_CTYPE` in either `alacritty.yml` or my `bashrc/zshrc` it will overwrite it right now. And setting `LC_CTYPE` to `C.UFT-8` in Mac OS fails. I am not sure thats a locale it supports, @kchibisov.

kchibisov:
>I am not sure thats a locale it supports, @kchibisov?

Oh, ok, seems odd to me that it doesn't allow this locale. and we can't just use `en_US.UTF-8`, etc, since it's not something that user likely wants, so if platform doesn't allow `C.UTF-8` then we're fine with using `UTF-8`, I guess.

casperstorm:
> > I am not sure thats a locale it supports, @kchibisov?
> 
> Oh, ok, seems odd to me that it doesn't allow this locale. and we can't just use `en_US.UTF-8`, etc, since it's not something that user likely wants, so if platform doesn't allow `C.UTF-8` then we're fine with using `UTF-8`, I guess.

I guess thats why Terminal.app and iTerm2.app is doing `UTF-8` as well 🤷‍♂️

> >That is true, but as far as I can tell this patch doesn't check if the value is set before changing it?
>
> If it's not it should start doing so.

It is not overwriting the value from `alacritty.yml`, or your *rc file. It is simply just making sure we don't end up in a situtation where we would have `LC_CTYPE: C`.

chrisduerr:
> It is not overwriting the value from alacritty.yml, or your *rc file. It is simply just making sure we don't end up in a situtation where we would have LC_CTYPE: C.

It isn't overwriting the value from `alacritty.yml`, since the environment variables are coincidentally set after the macOS locale is. Overwriting it in the shell works because it is done after Alacritty is started, so that obviously works too.

But `LC_CTYPE=en_US.UTF8 alacritty` does not work.

Regardless of most things working, we shouldn't override this if it's already set. So it should be checked first if this value is already set to something different.

alacritty_terminal/src/locale.rs
```diff
@@ -61,13 +62,29 @@ pub fn set_locale_environment() {
...
+        let ctype_string = ctype.into_string().unwrap();
+        match ctype_string.as_str() {
+            "C" => {
+                let utf8 = CString::new("UTF-8").unwrap();
+                setlocale(LC_CTYPE, utf8.as_ptr());
+                env::set_var("LC_CTYPE", utf8.into_string().unwrap());
+            },
+            _ => (),
+        }
+    }
```
casperstorm, line 86: @chrisduerr @kchibisov: Let me know what you think.

alacritty_terminal/src/locale.rs
```diff
@@ -59,21 +59,32 @@ pub fn set_locale_environment() {
...
+
+        let ctype_string = ctype.into_string().unwrap();
+        match ctype_string.as_str() {
+            "C" => {
+                let utf8 = CString::new("UTF-8").unwrap();
+                setlocale(LC_CTYPE, utf8.as_ptr());
+                env::set_var("LC_CTYPE", utf8.into_string().unwrap());
+            },
+            _ => (),
+        }
```
chrisduerr, line 85:
```suggestion
        let ctype = setlocale(LC_CTYPE, null());
        if !original.is_null() && CStr::from_ptr(ctype).to_string_lossy() == "C" {
            let locale = "UTF-8";
            setlocale(LC_CTYPE, CString::new(locale).unwrap());
            env::set_var("LC_CTYPE", locale);
        }
```

I think something like this should work a bit better.

However I question how well this will work by just retrieving the locale. I can test this on macOS tomorrow, but does this deal with setting the locale from the CLI correctly?

casperstorm: Hm, @chrisduerr. It seems like passing `LC_CTYPE` like `LC_CTYPE=en_US.UTF8 alacritty` does not pass it correctly. Any idea why?

Also your suggested code has some errors; i fixed those – what about the following?
```rust
let ctype = setlocale(LC_CTYPE, null());
if !ctype.is_null() && CStr::from_ptr(ctype).to_string_lossy() == "C" {
    let locale = "UTF-8";
    setlocale(LC_CTYPE, CString::new(locale).unwrap().as_ptr());
    env::set_var("LC_CTYPE", locale);
}

```

chrisduerr: > Hm, @chrisduerr. It seems like passing LC_CTYPE like LC_CTYPE=en_US.UTF8 alacritty does not pass it correctly. Any idea why?

I think you might be confusing locale with the environment variable. The environment variable is processed with `env::var("LC_CTYPE")`, while `setlocale(LC_CTYPE, null())` just reads the currently used locale.

> Also your suggested code has some errors; i fixed those – what about the following?

Yeah, should be fine. Was mostly about the general idea than the character for character implementation. The functionality is the bigger problem.

casperstorm:
I've updated the code to handle if you pass in an environment variable, good catch @chrisduerr.
Let me know what you think about it.

alacritty_terminal/src/locale.rs
```diff
@@ -65,24 +65,20 @@ pub fn set_locale_environment() {
...
 
-    // Provide a default value for LC_CTYPE if needed
+    // provide a default value for LC_CTYPE if needed
     unsafe {
-        let orginal = setlocale(LC_CTYPE, null());
-        let ctype = if orginal.is_null() {
-            CString::new("C").unwrap()
+        // check if we already have LC_CTYPE set as a environment variable
+        if let Ok(ctype) = env::var("LC_CTYPE") {
+            setlocale(LC_CTYPE, CString::new(ctype).unwrap().as_ptr());
```
chrisduerr, line 72:
```suggestion
            setlocale(LC_CTYPE, CString::new("").unwrap().as_ptr());
```

alacritty_terminal/src/locale.rs
```diff
@@ -65,24 +65,20 @@ pub fn set_locale_environment() {
         result
     };
 
-    // Provide a default value for LC_CTYPE if needed
+    // provide a default value for LC_CTYPE if needed
     unsafe {
-        let orginal = setlocale(LC_CTYPE, null());
-        let ctype = if orginal.is_null() {
-            CString::new("C").unwrap()
+        // check if we already have LC_CTYPE set as a environment variable
```
chrisduerr: Comments should be capitalized.

alacritty_terminal/src/locale.rs
```diff
@@ -65,24 +65,20 @@ pub fn set_locale_environment() {
...
+        if let Ok(ctype) = env::var("LC_CTYPE") {
+            setlocale(LC_CTYPE, CString::new(ctype).unwrap().as_ptr());
         } else {
-            CStr::from_ptr(orginal).to_owned()
+            // if current LC_CTYPE is either null or C, we overwrite it
```
chrisduerr, line 74: The comment isn't actually what we're doing. If the locale is null, we do not set the locale.

Though I don't think that it's possible for a call with a null pointer as parameter to return null, at least according to the documentation. If trying to call it with an invalid locale it will return null, but not set it. So future calls will just return the previous locale.

So I don't think it's necessary to document that specific part.

casperstorm: That is now fixed.

alacritty_terminal/src/locale.rs
casperstorm: > For example we could just always call setlocale with an empty string as parameter and check if the return value is C or not.

Why do you want to do this? I believe what we have right now is quite nice where we have one case to check if we have a environment variable, if we do we set the locale to it:

```rust
setlocale(LC_CTYPE, CString::new(ctype).unwrap().as_ptr());
```
If we don't have a enviorment variable passed along we will check if our `LC_CTYPE` locale is already set, and if it's `C` we will then overwrite it to `UTF-8`

```rust
let ctype = setlocale(LC_CTYPE, null());
if !ctype.is_null() && CStr::from_ptr(ctype).to_string_lossy() == "C" {
    let locale = "UTF-8";
    setlocale(LC_CTYPE, CString::new(locale).unwrap().as_ptr());
    env::set_var("LC_CTYPE", locale);
}
```

I am not sure i understand why you make [this change]( https://github.com/alacritty/alacritty/pull/3608/files#r414091716), could you elaborate on it @chrisduerr? I don't think we are missing any cases at the moment.

alacritty_terminal/src/locale.rs
casperstorm: This is now fixed.

alacritty_terminal/src/locale.rs
chrisduerr: Passing an empty string to `setlocale` tells it to set everything according to command line variables.

casperstorm: What is the benefit of doing that rather than setting it to the passed variable? Care to elaborate on that @chrisduerr?

alacritty_terminal/src/locale.rs
```diff
@@ -59,21 +59,27 @@ pub fn set_locale_environment() {
...
         let result = if modified.is_null() { String::new() } else { locale_id };
-        // restore original setting
-        setlocale(LC_CTYPE, saved_original.as_ptr());
         result
     };
 
+    // Provide a default value for LC_CTYPE if needed
+    unsafe {
+        // Check if we already have LC_CTYPE set as a environment variable
+        if let Ok(_) = env::var("LC_CTYPE") {
```
kchibisov, line 71:
```suggestion
        if env::var("LC_CTYPE").is_ok() {
```

alacritty_terminal/src/locale.rs
```diff
@@ -12,22 +12,62 @@
...
+
+        let fallback_locale_c = CString::new(FALLBACK_LOCALE).unwrap();
+        unsafe { setlocale(LC_CTYPE, fallback_locale_c.as_ptr()) };
+
+        env::set_var("LC_CTYPE", FALLBACK_LOCALE);
+    } else {
+        // Use system locale
+        debug!("Using system locale: {}", system_locale);
+
+        env::set_var("LC_ALL", system_locale);
```
chrisduerr, line 60: We should consider replacing the `LC_ALL` instances here with `LANG`, since that's probably more commonly used and checked against.

If we use `LANG`, then `LC_ALL` will be empty. If we use `LC_ALL`, then `LANG` will be empty. So it's probably better to have `LANG` filled. All other values like `LC_CTYPE` are automatically set when either is present.

casperstorm: I belive this is the right approach.

chrisduerr: Let's go for it then.

casperstorm:
@chrisduerr You are not waiting for me here, right? If so just let me know.

casperstorm:
> LGTM, do you want to give it another shot to verify LANG works @casperstorm?

Apparently theres no binding to `LANG` through https://github.com/rust-lang/libc.
```
Cannot find value `LANG` in crate `libc`
```

chrisduerr:
> Apparently theres no binding to LANG through https://github.com/rust-lang/libc.

Oh yeah, of course not. Because it's not a valid parameter for `setlocale`, only a valid environment variable.

I'll stick to LC_ALL, that should work.

PR #3610, opened 2020-04-19
Add deprecation warning for WINIT_HIDPI_FACTOR

Detegr:
I spent an hour wondering why `WINIT_HIDPI_FACTOR` no longer had any effect with new versions of Alacritty (my bad for not going through CHANGELOG.md). I figured it would be helpful to warn users using the environment variable that it has been renamed to `WINIT_X11_SCALE_FACTOR`.

### DIFF
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -217,6 +217,13 @@ fn print_deprecation_warnings(config: &Config) {
              the config"
         );
     }
+
+    if config.env.contains_key("WINIT_HIDPI_FACTOR") {
+        warn!(
+            target: LOG_TARGET_CONFIG,
+            "WINIT_HIDPI_FACTOR has been renamed to WINIT_X11_SCALE_FACTOR"
+        );
+    }
 }
 
 #[cfg(test)]


Detegr:
I agree that this is too late for 0.4.2, where the change was made, but isn't it better to add it later than leave it out completely? I would rather add the warning now, but I understand your reasoning if you don't want to merge this.

chrisduerr:
> I agree that this is too late for 0.4.2, where the change was made, but isn't it better to add it later than leave it out completely?

The thing is that this is a deprecation warning, for an item that is never publicly advertised in our documentation. It's sole existence is mainly special cases with broken xrandr and other usecases are mostly discouraged I'd say.

Additionally to that, we've already missed the opportunity to warn the majority of users, since it has already been removed. So most people that use this will hopefully already have resolved it.

I completely agree that it probably would have been the correct approach to warn on this with the 0.4.2 release, but now that it's too late already anyways, we can probably afford to just remove it so we won't have to deal with removing a deprecation warning in the future that never actually showed up for anyone.

In general we shouldn't only print deprecation warnings, but also provide a fallback during a transition period. In the future that is definitely the way to go and it's already what we do for our configuration issues. But since this variable comes from one of our depedencies, we've missed it in this case.

Thanks for suggesting this PR, but I think it's best to cut our losses here and move on. Hopefully this still serves as useful documentation for others running into similar problems.

PR #3618, opened 2020-04-20
Fix: clear selection on grid swap

### LINKED ISSUE #3290: Selection in Alt-screen isn't cleared on mode leave
if you run `alt-screen-random-write` from `vtebench`, select something during it, when it finish, you'll go back to a normal mode and won't see any selection. if you start it again you'll see your old selection from it.

I guess we should just always clear selection on mode toggle.

enizor:
Fixes #3290 

It should merge nicely with #3510 but it could wait for it just in case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Tabstops not being reset with `reset`
-- Selection not cleared on mode change
+- Selection not cleared when switching between main and alt grid
 
 ## 0.4.2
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Tabstops not being reset with `reset`
+- Selection not cleared on mode change
```
chrisduerr, line 29: We should be more precise here.

```suggestion
- Selection not cleared when switching between main and alt grid
```

enizor:
> Just for the record, this is already fixed by #3589, but I don't see a reason to wait for that PR to finish before merging this.

I did not see a mention on the original issue so I missed it :/

Do I need to squash the change you requested?

chrisduerr:
> Do I need to squash the change you requested?

No, all changes are squashed before I merge them. Squashing them beforehand just makes reviews more difficult, so it is discouraged.

PR #3630, opened 2020-04-22
Remove prebuilt linux binaries

### LINKED ISSUE #3628: Add man page to deb generated by cargo-deb
Hi, I think it would be useful to add Alacritty's man page to the deb generated with `cargo deb -p alacritty`. The deb already includes the shell completions so adding the man page doesn't seem like it would be difficult, but I have no experience with the Rust build process.

chrisduerr:
Fixes #3628.


alacritty/Cargo.toml
```diff
@@ -56,22 +56,3 @@ wayland = ["alacritty_terminal/wayland"]
...
-priority = "optional"
-assets = [
-    ["../target/release/alacritty", "usr/bin/", "755"],
-    ["../extra/linux/Alacritty.desktop", "usr/share/applications/", "644"],
-    ["../extra/logo/alacritty-term.svg", "usr/share/pixmaps/Alacritty.svg", "644"],
-    ["../extra/completions/alacritty.bash", "usr/share/bash-completion/completions/alacritty", "644"],
-    ["../extra/completions/alacritty.fish", "usr/share/fish/completions/alacritty.fish", "644"],
-    ["../extra/completions/_alacritty", "usr/share/zsh/vendor-completions/_alacritty", "644"],
-]
-maintainer-scripts = "../extra/linux/debian"
```
chrisduerr, line 77: I think removing the prebuilt Linux binaries is pretty straight-forward, since it just supports a fraction of the platforms we support and is more likely to cause troubles than actually be beneficial.

So I think the main remaining question is if we should remove the ability to build with `cargo-deb` completely, or if local deb builds should still be allowed.

kchibisov: I feel like it's fine to keep them around, since we have rpm specs around, etc, especially knowing that this part of `Cargo.toml` is unlikely to change.

chrisduerr: Case in point against this being unlikely to change: #3628.

There's also been a couple changes to `cargo-deb` in the past. I'm generally not super against it, but I don't think 'unlikely to change' is a great argument.

kchibisov: >Case in point against this being unlikely to change

Well, it's a valid issue, since we forget man page, and it's not something new.

chrisduerr: Yeah, of course it's valid. But that says nothing about potential future changes that are just as valid. So I wouldn't say it's particularly unlikely to change.

chrisduerr:
@mmstick I'm curious what your current process for building the Pop ppa is. I'd assume it is built completely from scratch, or are you making use of Alacritty's `cargo-deb` integration?

nixpulvis:
I'm generally in favor of this. But I'd be open to the idea of putting distribution related things into a separate repository if it can be integrated in a way that keeps things easy for some people, but moves packaging issues to that repo, so we avoid clutter in this one.

mmstick:
@chrisduerr I just forked the repository and [added debian packaging to it](https://github.com/pop-os/alacritty/tree/master/debian). Same sbuild-compatible debian rules we use in all our Rust projects.

kchibisov:
Hm, ok, thx.

P.s.

v0.4.2 was released some time ago in case you want to bump it.

kinghajj:
Um, can we at least keep the Debian packaging stuff? I use `cargo deb -p alacritty_terminal` to generate a package for my local computer.

chrisduerr:
The Alacritty upstream aims to solely maintain the Alacritty source code, not the distribution of it. That is much better done by separate parties.

If you're interested in still building a deb from the latest master, you can make use of mmstick's PPA repo which likely has a higher quality .deb than what Alacritty had before.

kinghajj:
I assume [this repo](https://github.com/pop-os/alacritty) is the one that generates the PPA .deb files? It looks like that repo is a bit behind, and doesn't merge cleanly with this repo. So it makes generating a package a bit more of a PITA than before. I guess I'll just maintain my own patch for this repo that retains the Debian packaging info.

chrisduerr:
Looking at the log and commits, it should rebase trivially with our master.

aslatter:
I wasn't able to get a merge or rebase to go through cleanly, but directly checking-out just the *debian* folder from https://github.com/pop-os/alacritty onto alacritty-master worked for me on Ubuntu 20.04.

I'm not-great at Debian-stuff so I've scripted the process here: https://gitlab.com/aslatter/alacritty_build

PR #3635, opened 2020-04-24
Add support for reporting window size in pixels and characters

ayosec:
This patch adds the following sequences:

* `CSI 14 t`, to report the window size in pixels.
* `CSI 18 t`, to report the area size in characters.

These sequences are described in [dtterm(5)](https://www.freebsd.org/cgi/man.cgi?query=dtterm&sektion=5&manpath=FreeBSD+12.1-RELEASE+and+Ports).

### DIFF
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -316,6 +316,12 @@ pub trait Handler {
 
     /// Pop the last title from the stack.
     fn pop_title(&mut self) {}
+
+    /// Report window size in pixels.
+    fn window_size_pixels<W: io::Write>(&mut self, _: &mut W) {}
+
+    /// Report area size in characters.
+    fn area_size_chars<W: io::Write>(&mut self, _: &mut W) {}
 }
 
 /// Describes shape of cursor.
@@ -1080,6 +1086,8 @@ where
             ('s', None) => handler.save_cursor_position(),
             ('T', None) => handler.scroll_down(Line(next_param_or(1) as usize)),
             ('t', None) => match next_param_or(1) as usize {
+                14 => handler.window_size_pixels(writer),
+                18 => handler.area_size_chars(writer),
                 22 => handler.push_title(),
                 23 => handler.pop_title(),
                 _ => unhandled!(),
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -317,11 +317,11 @@ pub trait Handler {
     /// Pop the last title from the stack.
     fn pop_title(&mut self) {}
 
-    /// Report window size in pixels.
-    fn window_size_pixels<W: io::Write>(&mut self, _: &mut W) {}
+    /// Report text area size in pixels.
+    fn text_area_size_pixels<W: io::Write>(&mut self, _: &mut W) {}
 
-    /// Report area size in characters.
-    fn area_size_chars<W: io::Write>(&mut self, _: &mut W) {}
+    /// Report text area size in characters.
+    fn text_area_size_chars<W: io::Write>(&mut self, _: &mut W) {}
 }
 
 /// Describes shape of cursor.
@@ -1086,8 +1086,8 @@ where
             ('s', None) => handler.save_cursor_position(),
             ('T', None) => handler.scroll_down(Line(next_param_or(1) as usize)),
             ('t', None) => match next_param_or(1) as usize {
-                14 => handler.window_size_pixels(writer),
-                18 => handler.area_size_chars(writer),
+                14 => handler.text_area_size_pixels(writer),
+                18 => handler.text_area_size_chars(writer),
                 22 => handler.push_title(),
                 23 => handler.pop_title(),
                 _ => unhandled!(),
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -2774,43 +2774,6 @@ mod tests {
         assert_eq!(version_number("1.2.3-dev"), 1_02_03);
         assert_eq!(version_number("999.99.99"), 9_99_99_99);
     }
-
-    #[test]
-    fn report_size_in_pixels_and_chars() {
-        let size = SizeInfo {
-            width: 21.0,
-            height: 51.0,
-            cell_width: 3.0,
-            cell_height: 3.0,
-            padding_x: 0.0,
-            padding_y: 0.0,
-            dpr: 1.0,
-        };
-
-        let mut term = Term::new(&MockConfig::default(), &size, Mock);
-        let mut processor = ansi::Processor::default();
-
-        macro_rules! run_sequence {
-            ($input:expr, $output:expr) => {
-                let mut writer = Vec::new();
-                for &i in &$input[..] {
-                    processor.advance(&mut term, i, &mut writer);
-                }
-
-                assert_eq!(&$output[..], &writer[..]);
-            };
-        }
-
-        run_sequence!(b"\x1b[14t", b"\x1b[4;51;21t");
-        run_sequence!(b"\x1b[18t", b"\x1b[8;17;7t");
-
-        // Resize
-        let size = SizeInfo { width: 210.0, height: 510.0, ..size };
-        term.resize(&size);
-
-        run_sequence!(b"\x1b[14t", b"\x1b[4;510;210t");
-        run_sequence!(b"\x1b[18t", b"\x1b[8;170;70t");
-    }
 }
 
 #[cfg(all(test, feature = "bench"))]
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -969,6 +969,8 @@ impl<T> Term<T> {
 
     /// Resize terminal to new dimensions.
     pub fn resize(&mut self, size: SizeInfo) {
+        self.size = size;
+
         let old_cols = self.cols();
         let old_lines = self.screen_lines();
         let num_cols = max(size.cols(), Column(MIN_SIZE));
@@ -1002,8 +1004,6 @@ impl<T> Term<T> {
 
         self.grid.resize(!is_alt, num_lines, num_cols);
         self.inactive_grid.resize(is_alt, num_lines, num_cols);
-
-        self.size = size;
 
         // Clamp vi cursor to viewport.
         self.vi_mode_cursor.point.col = min(self.vi_mode_cursor.point.col, num_cols - 1);
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -746,7 +746,7 @@ pub struct Term<T> {
     /// Current forward and backward buffer search regexes.
     regex_search: Option<RegexSearch>,
 
-    /// Terminal size info.
+    /// Information about window dimensions.
     size: SizeInfo,
 }
 
@@ -761,7 +761,7 @@ impl<T> Term<T> {
         self.dirty = true;
     }
 
-    pub fn new<C>(config: &Config<C>, size: &SizeInfo, event_proxy: T) -> Term<T> {
+    pub fn new<C>(config: &Config<C>, size: SizeInfo, event_proxy: T) -> Term<T> {
         let num_cols = size.cols();
         let num_lines = size.lines();
 
@@ -798,7 +798,7 @@ impl<T> Term<T> {
             title_stack: Vec::new(),
             selection: None,
             regex_search: None,
-            size: *size,
+            size,
         }
     }
 
@@ -968,7 +968,7 @@ impl<T> Term<T> {
     }
 
     /// Resize terminal to new dimensions.
-    pub fn resize(&mut self, size: &SizeInfo) {
+    pub fn resize(&mut self, size: SizeInfo) {
         let old_cols = self.cols();
         let old_lines = self.screen_lines();
         let num_cols = max(size.cols(), Column(MIN_SIZE));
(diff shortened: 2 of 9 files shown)


alacritty_terminal/src/ansi.rs
```diff
@@ -334,6 +334,12 @@ pub trait Handler {
 
     /// Pop the last title from the stack
     fn pop_title(&mut self) {}
+
+    /// Report window size in pixels
```
chrisduerr, line 338:
```suggestion
    /// Report window size in pixels.
```

alacritty_terminal/src/ansi.rs
```diff
@@ -334,6 +334,12 @@ pub trait Handler {
...
+    fn window_size_pixels<W: io::Write>(&mut self, _: &mut W) {}
+
+    /// Report area size in characters
```
chrisduerr, line 341:
```suggestion
    /// Report area size in characters.
```

alacritty_terminal/src/ansi.rs
```diff
@@ -1015,6 +1021,8 @@ where
             },
             ('S', None) => handler.scroll_up(Line(arg_or_default!(idx: 0, default: 1) as usize)),
             ('t', None) => match arg_or_default!(idx: 0, default: 1) as usize {
+                14 => handler.window_size_pixels(writer),
```
chrisduerr, line 1024: We probably shouldn't return the text area size when the outer window size is requested.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2230,6 +2242,16 @@ impl<T: EventListener> Handler for Term<T> {
             self.set_title(popped);
         }
     }
+
+    #[inline]
+    fn window_size_pixels<W: io::Write>(&mut self, writer: &mut W) {
+        let _ = write!(writer, "\x1b[4;{};{}t", self.window_height, self.window_width);
+    }
```
chrisduerr, line 2249: This reports the inner window size, not the text area size.

ayosec:
> Have you checked which specific application actually uses these escape sequences?

Yes, [lsix uses it](https://github.com/hackerb9/lsix/blob/ab561bc022057159f70914b63d8932cbfab078e9/lsix#L167-L168) when `CSI ? 2 ; 1 ; 0 S` is missing.

alacritty_terminal/src/term/mod.rs
ayosec: What should be sent? `width + padding_x`?

chrisduerr: You'd have to subtract the padding from the inner size.

chrisduerr:
> Yes, lsix uses it when CSI ? 2 ; 1 ; 0 S is missing.

So this escape would be completely useless if we decide to implement `CSI ? Pi ; Pv ; Pa S`?

ayosec:
> So this escape would be completely useless if we decide to implement `CSI ? Pi ; Pv ; Pa S`?

Yes, but `CSI ? ... S` was rejected in the discussion of https://github.com/alacritty/alacritty/issues/910, since it is Xterm specific.

alacritty_terminal/src/term/mod.rs
ayosec: Is this correct?

```rust
self.window_height = (size.height - size.padding_y * 2) as usize;
self.window_width = (size.width - size.padding_x * 2) as usize;
```

chrisduerr:
> Yes, but CSI ? ... S was rejected in the discussion of #910, since it is Xterm specific.

It would be interesting to see which one is actually more commonly used by terminals supporting sixel. "XTerm specific" doesn't really mean super much, since there are a lot of XTerm extensions that are used by other terminals.

We might also want to look if w3m queries these escapes, it might be useful for that too. We've had issues in the past with w3m ignoring padding, if we implemented the text area position escape it might be possible to fix that. Since this PR already implements text area size, that should be pretty simple. But one would first have to check if `w3m` attempts to check these.

ayosec:
> We might also want to look if w3m queries these escapes

It tries with `TIOCGWINSZ`, and fallback to `CSI 14 t` `CSI 18 t` ([source](https://sources.debian.org/src/w3m/0.5.3-38/terms.c/#L682-L692)).

Do you prefer to keep only `TIOCGWINSZ`?

chrisduerr:
> It tries with TIOCGWINSZ, and fallback to CSI 14 t CSI 18 t (source).

Wouldn't that mean that this doesn't work? Since it will use `TIOCGWINSZ` instead and that does not include the padding?

ayosec:
> > It tries with TIOCGWINSZ, and fallback to CSI 14 t CSI 18 t (source).
> 
> Wouldn't that mean that this doesn't work? Since it will use `TIOCGWINSZ` instead and that does not include the padding?

I guess that the patch to support sixel in w3m assumes that the full window is available to draw images.

If the terminal has padding the image will be rendered smaller than w3m thinks.

chrisduerr:
I'm sorry for the lack of communication, but I think this PR should be good to go after getting it up to date.

I think that its use for w3m or sixel might be questionable, but irrelevant. All major terminals including XTerm, Kitty, URxvt and VTE support CSI 14 and CSI 18, so it really doesn't matter much if it's required for sixel or not, it's just a good idea to provide this information with these commonly supported escape sequences.

If you're interested in finishing this up yourself, please let me know @ayosec. Otherwise I'd take matters into my own hands.

ayosec:
I updated the PR with the changes from the current *master* branch.

Do you want to add the change from https://github.com/alacritty/alacritty/pull/3635#discussion_r414944154 ?

chrisduerr:
> Do you want to add the change from #3635 (comment) ?

Honestly I'm not 100% sure. You'd probably have to test yourself if padding is usually included in the pixel size or not with other terminals. I'd assume it's not, which means that your comment is correct (except for the `as usize` part which doesn't seem to have any use).

ayosec:
> You'd probably have to test yourself if padding is usually included in the pixel size or not with other terminals.

I checked the terminals mentioned in <https://github.com/alacritty/alacritty/pull/3635#issuecomment-670623621>:

* **XTerm**

    It uses the size of the widget. With `CSI 14 ; 2 t` it returns the size of the container window.  I don't see any padding option ([source](https://sources.debian.org/src/xterm/344-1/charproc.c/#L7295)).

* **Kitty**

    It computes the size with `cell size × (columns, rows)`.

    https://github.com/kovidgoyal/kitty/blob/5eefd41059855ccc91d7b836ed678f6c8d70ff67/kitty/screen.c#L1356-L1366

* **URxvt**

    It uses the size of the window ([source](http://cvs.schmorp.de/rxvt-unicode/src/command.C?revision=1.585&view=markup#l3180)).

* **VTE**

    It computes the size with `cell size × (columns, rows)` ([source](https://gitlab.gnome.org/GNOME/vte/-/blob/12599636842931003c01d07e82496018ac56de1e/src/vteseq.cc#L8768-8772)).


I updated the PR to use the approach of VTE and Kitty

alacritty_terminal/src/term/mod.rs
```diff
@@ -745,6 +745,12 @@ pub struct Term<T> {
 
     /// Current forward and backward buffer search regexes.
     regex_search: Option<RegexSearch>,
+
+    /// Width of individual cell.
+    cell_width: f32,
+
+    /// Height of individual cell.
+    cell_height: f32,
```
chrisduerr, line 753: What's the reason for saving the cell width/height here, when what we actually need is the text area size? Wouldn't it be easier to just store the only thing we need?

If we store more information than necessary, why wouldn't we just store `SizeInfo`?

alacritty_terminal/src/term/mod.rs
```diff
@@ -2755,6 +2779,43 @@ mod tests {
         assert_eq!(version_number("1.2.3-dev"), 1_02_03);
         assert_eq!(version_number("999.99.99"), 9_99_99_99);
     }
+
+    #[test]
+    fn report_size_in_pixels_and_chars() {
```
chrisduerr, line 2784: I'm not entirely sure about the stuff this test covers. If the goal is just to cover the dispatched method we shouldn't run a parser. If the goal is to test that the handler dispatches the correct method then we should move it to `ansi.rs` and don't test what the handler itself does.

So it seems like this would better be separated, to test only one specific thing rather than two. Most of the `ansi.rs` isn't covered but I think that's not too big of a deal, since the stuff is mostly plug and play and there's not a lot of functionality to cover with tests. There's no need to cover vte either way.

alacritty_terminal/src/ansi.rs
```diff
@@ -316,6 +316,12 @@ pub trait Handler {
 
     /// Pop the last title from the stack.
     fn pop_title(&mut self) {}
+
+    /// Report window size in pixels.
```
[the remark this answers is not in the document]
ayosec: I used “window size” because it is the terminology used in other terminal implementations.

It is changed to “text area size” in c76d20529d00ee4b31bb8a752d313822c16a31d8.

alacritty_terminal/src/term/mod.rs
ayosec: Initially I didn't want to copy the `SizeInfo` to reduce the final size of `Term`, but I guess that the difference is too small to be worth it. In 5581bf7a6ca31faaf2e41402764cbe9b9365c42b it is changed to keep a copy of `SizeInfo`.

alacritty_terminal/src/term/mod.rs
ayosec: I wanted a test to cover both the parser and the handler of the sequence. I removed it 34d57a289e9096605230bd585cce884ee2b7d847.

alacritty_terminal/src/term/mod.rs
```diff
@@ -745,6 +745,9 @@ pub struct Term<T> {
 
     /// Current forward and backward buffer search regexes.
     regex_search: Option<RegexSearch>,
+
+    /// Terminal size info.
```
chrisduerr, line 749:
```suggestion
    /// Information about window dimensions.
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -795,6 +798,7 @@ impl<T> Term<T> {
             title_stack: Vec::new(),
             selection: None,
             regex_search: None,
+            size: *size,
```
chrisduerr, line 801: If we need a copy anyways, just let the callee move the value by copy instead of reference and then change this to just `size,`.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1005,6 +1009,9 @@ impl<T> Term<T> {
 
         // Reset scrolling region.
         self.scroll_region = Line(0)..self.screen_lines();
+
+        // Save term size.
```
chrisduerr, line 1013:
```suggestion
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -2219,6 +2226,18 @@ impl<T: EventListener> Handler for Term<T> {
...
+    fn text_area_size_pixels<W: io::Write>(&mut self, writer: &mut W) {
+        let width = self.size.cell_width * self.grid.cols().0 as f32;
+        let height = self.size.cell_height * self.grid.screen_lines().0 as f32;
+        let _ = write!(writer, "\x1b[4;{};{}t", height, width);
+    }
+
+    #[inline]
+    fn text_area_size_chars<W: io::Write>(&mut self, writer: &mut W) {
+        let _ = write!(writer, "\x1b[8;{};{}t", self.grid.screen_lines(), self.grid.cols());
+    }
```
chrisduerr, line 2240: Just to clarify on my previous comment: I'd be perfectly fine with a test that just calls these methods directly, I think that would make some sense.

But since the other handler methods also aren't really covered and the functionality is relatively simple, I wouldn't require it here. But since you had a test previously anyways, we could add it here.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2219,6 +2226,18 @@ impl<T: EventListener> Handler for Term<T> {
             self.set_title(popped);
         }
     }
+
+    #[inline]
+    fn text_area_size_pixels<W: io::Write>(&mut self, writer: &mut W) {
+        let width = self.size.cell_width * self.grid.cols().0 as f32;
+        let height = self.size.cell_height * self.grid.screen_lines().0 as f32;
```
chrisduerr, line 2233:
```suggestion
        let width = self.size.cell_width * self.cols().0 as f32;
        let height = self.size.cell_height * self.screen_lines().0 as f32;
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -2219,6 +2226,18 @@ impl<T: EventListener> Handler for Term<T> {
...
+        let _ = write!(writer, "\x1b[4;{};{}t", height, width);
+    }
+
+    #[inline]
+    fn text_area_size_chars<W: io::Write>(&mut self, writer: &mut W) {
+        let _ = write!(writer, "\x1b[8;{};{}t", self.grid.screen_lines(), self.grid.cols());
```
chrisduerr, line 2239:
```suggestion
        let _ = write!(writer, "\x1b[8;{};{}t", self.screen_lines(), self.cols());
```

ayosec:
96a59b06ff104c54eda8183920b537ce9c6314a9 contains the latest suggestions.

alacritty_terminal/src/term/mod.rs
```diff
@@ -999,6 +1003,8 @@ impl<T> Term<T> {
         self.grid.resize(!is_alt, num_lines, num_cols);
         self.inactive_grid.resize(is_alt, num_lines, num_cols);
 
+        self.size = size;
```
chrisduerr, line 1006: This should be at the top, even before the num_cols/num_lines change. It's possible that the terminal resized but the number of columns/lines stays the same.

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
```diff
@@ -2219,6 +2225,18 @@ impl<T: EventListener> Handler for Term<T> {
             self.set_title(popped);
         }
     }
+
+    #[inline]
+    fn text_area_size_pixels<W: io::Write>(&mut self, writer: &mut W) {
+        let width = self.size.cell_width * self.cols().0 as f32;
+        let height = self.size.cell_height * self.screen_lines().0 as f32;
```
chrisduerr, line 2232: I think printing these as f32 here is potentially dangerous, since it would mangle the escape if these would actually be printed as floating point numbers. So the cell_width and cell_height should probably just be converted to usize.

```suggestion
        let width = self.size.cell_width as usize * self.cols().0;
        let height = self.size.cell_height as usize * self.screen_lines().0;
```

chrisduerr (changes requested):
I've tested the escapes and everything works as intended.

Just one last thing I've noticed to make sure no accidents happen, then things are good to go.

chrisduerr (approved):
LGTM!

PR #3639, opened 2020-04-25
Use embedded resource for window icon

davidhewitt:
Just a quick one to use `Icon::from_resource` to (on Windows) load the window icon from the resource embedded by the Windows manifest, rather than including it for a second time via `include_bytes`.

In my experience Windows does a slightly better job of scaling the window icon image in this case. It also leads to a binary size saving of 300KB (~6% of a release build as measured on my machine).

Before:

before

After:

after

(It's subtle, but to me lines are in general a bit smoother in the "after" shot.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Block cursor is no longer inverted at the start/end of a selection
 - Preserve selection on non-LMB or mouse mode clicks
 - Wayland client side decorations are now based on config colorscheme
+- Low resolution window decoration icon on Windows
 
 ### Fixed
 - Tabstops not being reset with `reset`
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -24,13 +24,18 @@ use glutin::event_loop::EventLoop;
 use glutin::platform::macos::{RequestUserAttentionType, WindowBuilderExtMacOS, WindowExtMacOS};
 #[cfg(not(any(target_os = "macos", windows)))]
 use glutin::platform::unix::{EventLoopWindowTargetExtUnix, WindowBuilderExtUnix, WindowExtUnix};
+#[cfg(windows)]
+use glutin::platform::windows::IconExtWindows;
 #[cfg(not(target_os = "macos"))]
 use glutin::window::Icon;
 use glutin::window::{CursorIcon, Fullscreen, Window as GlutinWindow, WindowBuilder, WindowId};
 use glutin::{self, ContextBuilder, PossiblyCurrent, WindowedContext};
-#[cfg(not(target_os = "macos"))]
+#[cfg(not(any(target_os = "macos", windows)))]
 use image::ImageFormat;
+#[cfg(not(any(target_os = "macos", windows)))]
 use log::error;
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Block cursor is no longer inverted at the start/end of a selection
 - Preserve selection on non-LMB or mouse mode clicks
 - Wayland client side decorations are now based on config colorscheme
+- Windows window icon is now loaded from the embedded resource
```
```
- Low resolution window decoration icon on Windows
```

I think this reads a bit better, since `Windows window` is a little awkward. I'd also move this to the fixed section, since from a user perspective it fixes an issue.

alacritty/src/window.rs
```diff
@@ -252,11 +255,7 @@ impl Window {
             _ => true,
         };
 
-        let image = image::load_from_memory_with_format(WINDOW_ICON, ImageFormat::Ico)
-            .expect("loading icon")
-            .to_rgba();
-        let (width, height) = image.dimensions();
-        let icon = Icon::from_rgba(image.into_raw(), width, height);
+        let icon = Icon::from_resource(0x101, None);
```
chrisduerr, line 258: What does 0x101 mean? Is there a constant we can use for that?

davidhewitt: It's the [same constant the icon is declared with in the manifest](https://github.com/alacritty/alacritty/blob/master/extra/windows/windows.rc#L1-L3)

kchibisov: could you give it a name and point from where it was taken?

kchibisov:
Will this PR work from just `cargo run`, I guess we won't have an icon here?

davidhewitt:
It's working for me with `cargo run`. Let me delete `target/` and retry just to be sure...

davidhewitt:
Yep, it does work with `cargo run`. (Pretty sure the resource embedding is done as part of `build.rs`.)

kchibisov:
Yeah, you're correct https://github.com/alacritty/alacritty/blob/f48c43f4d2fcac35fe2777a3f0f155c275d0204e/alacritty/build.rs#L40

alacritty/src/window.rs
```diff
@@ -47,9 +52,14 @@ use crate::gl;
 use crate::wayland_theme::AlacrittyWaylandTheme;
 
 // It's required to be in this directory due to the `windows.rc` file
-#[cfg(not(target_os = "macos"))]
+#[cfg(not(any(target_os = "macos", windows)))]
 static WINDOW_ICON: &[u8] = include_bytes!("../../extra/windows/alacritty.ico");
 
+// This should match the definition of IDI_ICON from `windows.rc`
+#[cfg(windows)]
+const IDI_ICON: WORD = 0x101;
```
kchibisov, line 60:
```suggestion
static IDI_ICON: WORD = 0x101;
```

davidhewitt: Why the preference for `static` here?

kchibisov: I forgot that we really need constexpr like thing here. nvm.

davidhewitt:
> 
> 
> rustfmt.

Done. (Darn it I forget that all the time! xD)

Issue #2764, opened 2019-08-29
Tabs not displayed when `cat`ing files


prometheanfire:
Gentoo, git master (as of 20190828-235959), Wayland.

This seems to be a thing supported in gn
I'm not sure if it's posix or not, but other terminal emulators display a 'tab' character which is useful for copying.  Alacritty does not support this (neither does kitty, but they don't care).  xfce4-terminal, gnome-terminal, urxvt supports it though.  The stackexchange link below has some useful tests for outputing tabs though.  When piping into wgetpaste it seems to work, but still doesn't work with when copying/pasting.

More info:

https://unix.stackexchange.com/questions/324676/output-tab-character-on-terminal-window/324679

https://bugzilla.gnome.org/show_bug.cgi?id=769316

prometheanfire:
https://gist.github.com/prometheanfire/02d9f52aea5cd27222e0e49540ca8244 shows as piped to `wgetpaste`.  when displaying in the terminal tabs are just spaces.  This makes it very hard to submit patches upstream (or copying anything with tabs...).

ECMA-48 is the standard that needs to be followed.

kchibisov:
@prometheanfire Do you mean drawing some "visual" char for tabs? If you cat file with tabs and then copy it, you'll copy everything with tabs. I've just wrote a file with tabs and then copy pasted it to vim and everything was fine (I'm also on Wayland, so we're using the same clipboard backend). The problem that you've faced probably, is that you didn't select all chars that forms a tab (default to 8).

prometheanfire:
you are correct, copying the text then pasting it pastes as tabs (just tested the same way, though I already had a patch file with tabs).

I guess what I'm talking about is the tabs do not show up as 'tabs'. I'm not sure what method the other terminals use, but when I highlight text in the terminals that I mentioned as 'working' it'll only highlight one 'block' as the tab, instead of the 8 spaces (or whatever is configured).  I imagine they are just using a hack to make a space extend or something (the stackexchange post mentioned that it was hackish), but it is useful.

Also, does alacritty do something when text is copied out to preserve the tabs?  I noticed that kitty does not preserve tabs on copy.

All in all this is much better than I imagined :D

kchibisov:
We do handle tabs explicitly, this was added here 0c3e2861 (related PR #991). The behavior you're referring to is related to how our selection works.

[...]

PR #3647, opened 2020-04-26
Automatically expands tab selection to nearest tab border

Dav1dde:
Fixes: https://github.com/alacritty/alacritty/issues/2764

Things to do:
- [x] Add `CHANGELOG.md` entry
- [x] Expand visual selection
- [x] Expand clipboard selection

Potential issues:
 - Changing tab width retroactively affects printed lines with an non-contained `\t`, e.g.: 
```
$ echo '12345678\t'
$ tabs 1,40,50
```

alactab4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Tabstops not being reset with `reset`
 - Selection not cleared when switching between main and alt grid
 - Fallback to `LC_CTYPE=UTF-8` on macOS without valid system locale
+- Expands selection, including a partial tab, to the nearest tab border
 
 ## 0.4.2
 
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -209,6 +209,7 @@ pub struct CursorKey {
 pub struct RenderableCellsIter<'a, C> {
     inner: DisplayIter<'a, Cell>,
     grid: &'a Grid<Cell>,
+    tabs: &'a TabStops,
     cursor: RenderableCursor,
     config: &'a Config<C>,
     colors: &'a color::List,
@@ -257,6 +258,7 @@ impl<'a, C> RenderableCellsIter<'a, C> {
         RenderableCellsIter {
             cursor: term.renderable_cursor(config),
             grid,
+            tabs: &term.tabs,
             inner,
             selection: selection_range,
             config,
@@ -290,6 +292,13 @@ impl<'a, C> RenderableCellsIter<'a, C> {
 
         let num_cols = self.grid.num_cols().0;
         let cell = self.grid[&point];
+
+        // highlight this cell if it is part of a partially selected tab
+        if let Some(mut tab_range) = self.tabs.tab_range(self.grid, point) {
+            if tab_range.any(|col| selection.contains(col, point.line)) {
+                return true;
+            }
+        }
 
         // Check if wide char's spacers are selected
         if cell.flags.contains(Flags::WIDE_CHAR) {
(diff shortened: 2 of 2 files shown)


Dav1dde:
Clippy currently fails on Travis:

```
error: an inclusive range would be more readable
    --> alacritty_terminal/src/term/mod.rs:2286:37
     |
2286 |         for col in IndexRange::from(Column(0)..point.col + 1).rev() {
     |                                     ^^^^^^^^^^^^^^^^^^^^^^^^ help: use: `Column(0)..=point.col`
     |
note: lint level defined here
    --> alacritty_terminal/src/lib.rs:16:9
     |
16   | #![deny(clippy::all, clippy::if_not_else, clippy::enum_glob_use, clippy::wrong_pub_self_convention)]
     |         ^^^^^^^^^^^
     = note: `#[deny(clippy::range_plus_one)]` implied by `#[deny(clippy::all)]`
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#range_plus_one
error: an inclusive range would be more readable
    --> alacritty_terminal/src/term/mod.rs:2311:42
     |
2311 |             return Some(IndexRange::from(tab_begin..grid.num_cols() + 1));
     |                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use: `tab_begin..=grid.num_cols()`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#range_plus_one
```

Strangely enough it does not for me locally.

Anyways not sure how to fix this, I cannot change it to an inclusive range because `IndexRange` is not compatible with the `RangeInclusive` trait. Should I just clippy allow it?

alacritty_terminal/src/term/mod.rs
```diff
@@ -2264,6 +2280,41 @@ impl TabStops {
             is_tabstop
         });
     }
+
+    fn tab_range(&self, grid: &Grid<Cell>, point: Point) -> Option<IndexRange<Column>> {
```
Dav1dde, line 2284: I was not sure where to put this function, it feels wrong here, but also not completely out of place.

Also not sure if it should take `Point<usize>` instead

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Tabstops not being reset with `reset`
 - Selection not cleared when switching between main and alt grid
 - Fallback to `LC_CTYPE=UTF-8` on macOS without valid system locale
+- Expands selection, including a partial tab, to the nearest tab border
```
```suggestion
- Prevent partial selection of tab characters
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -291,6 +293,13 @@ impl<'a, C> RenderableCellsIter<'a, C> {
         let num_cols = self.grid.num_cols().0;
         let cell = self.grid[&point];
 
+        // highlight this cell if it is part of a partially selected tab
```
chrisduerr, line 296:
```suggestion
        // Highlight this cell if it is part of a partially selected tab.
```

alacritty_terminal/src/term/mod.rs
chrisduerr: Is the point of this function to find the start and end of the tab the `point` is in? This seems a bit complicated.

Also isn't it pointless to get the entire range of the tab, when the only goal is to determine if a point lies within a selected tab?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1051,8 +1060,15 @@ impl<T> Term<T> {
             cols.start -= 1;
         }
 
+        let begin_point = self.grid.clamp_buffer_to_visible(Point::new(line, cols.start));
+        let begin_tab_range = self.tabs.tab_range(&self.grid, begin_point);
+
+        // if the range begins in the middle of a tab, move it to the beginning of the tab,
+        // this will include the tab character in the resulting string
+        let col_start = begin_tab_range.map(|range| range.0.start).unwrap_or(cols.start);
```
chrisduerr, line 1068: Is the entire reason for this just to prepend a `\t`? Seems like that could be done a little easier.

alacritty_terminal/src/term/mod.rs
Dav1dde: But you need to first figure out if the tab is selected.

Correct me if I am wrong here:

Every cell gets checked if it is selected, so in order to know if the current cell is selected, we have to check if the cell is part of a tab (go to the left and find a \t) and if the current cell is a tab, we have to find out if any cell of the tab is already selected.

E.g.:
* x = current cell
* y = selected
* \t = tab begin

`\t.....x......y` 

* part of a tab: to the right of a \t without a tab border
* any cell of the tab is selected (y)

Now if there is a higher level way of marking a tab and all of its cells as selected, this would obviously be easier. I've seen there are modifiers as used for wide chars, but it didnt seem like they are meant for this usecase.

alacritty_terminal/src/term/mod.rs
Dav1dde: The reason for this is to readjust the start of the selection to the actual beginning of the tab, in order to have the already existing logic for tabs add a `\t` to the copied text.

PR #3648, opened 2020-04-26
Update help for class option to match man pages

### LINKED ISSUE #3615: Document setting instance and base class in --help output
When opening Alacritty with `--class` option it changes the window instance instead the class.
For example, if I do `alacritty --class foo` and I check the window properties I got:
```
Instance: foo
Class: Alacritty
Title: Alacritty
```
`--title` option is working as expected, although.

### Infos

OS: Funtoo Linux
Window Manager: dwm
Alacritty: alacritty 0.5.0-dev (03bce99)

jeffwindsor:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dim escape (`CSI 2 m`) support for truecolor text
 - Incorrectly deleted lines when increasing width with a prompt wrapped using spaces
 - Docs for class in --help missing information on setting general class
+- Documentation for class in `--help` missing information on setting general class
 
 ## 0.4.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Scroll down escape (`CSI Ps T`) incorrectly pulling lines from history
 - Dim escape (`CSI 2 m`) support for truecolor text
 - Incorrectly deleted lines when increasing width with a prompt wrapped using spaces
+- Docs for class in --help missing information on setting general class
 
 ## 0.4.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Scroll down escape (`CSI Ps T`) incorrectly pulling lines from history
 - Dim escape (`CSI 2 m`) support for truecolor text
 - Incorrectly deleted lines when increasing width with a prompt wrapped using spaces
-- Docs for class in --help missing information on setting general class
 - Documentation for class in `--help` missing information on setting general class
 
 ## 0.4.3
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -24,7 +24,8 @@ pub struct Options {
     pub dimensions: Option<Dimensions>,
     pub position: Option<Delta<i32>>,
     pub title: Option<String>,
-    pub class: Option<String>,
+    pub class_instance: Option<String>,
+    pub class_general: Option<String>,
     pub embed: Option<String>,
     pub log_level: LevelFilter,
     pub command: Option<Program>,
@@ -43,7 +44,8 @@ impl Default for Options {
             dimensions: None,
             position: None,
             title: None,
-            class: None,
+            class_instance: None,
+            class_general: None,
             embed: None,
             log_level: LevelFilter::Warn,
             command: None,
@@ -122,7 +124,9 @@ impl Options {
             .arg(
                 Arg::with_name("class")
                     .long("class")
+                    .value_name("instance> | <instance>,<general")
                     .takes_value(true)
+                    .use_delimiter(true)
                     .help(&format!("Defines window class on Linux [default: {}]", DEFAULT_NAME)),
             )
             .arg(
@@ -200,7 +204,11 @@ impl Options {
             }
         }
 
-        options.class = matches.value_of("class").map(ToOwned::to_owned);
+        if let Some(mut class) = matches.values_of("class") {
+            options.class_instance = class.next().map(|instance| instance.to_owned());
+            options.class_general = class.next().map(|general| general.to_owned());
+        }
+
         options.title = matches.value_of("title").map(ToOwned::to_owned);
         options.embed = matches.value_of("embed").map(ToOwned::to_owned);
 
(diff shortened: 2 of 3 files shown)


jeffwindsor:
fixes #3615

alacritty/src/cli.rs
```diff
@@ -137,7 +137,8 @@ impl Options {
                 Arg::with_name("class")
                     .long("class")
                     .takes_value(true)
-                    .help(&format!("Defines window class on Linux [default: {}]", DEFAULT_NAME)),
+                    .value_names(&["instance", "general"])
+                    .help(&format!("Defines window class hint on Linux [default: {},{}]", DEFAULT_NAME, DEFAULT_NAME)),
```
chrisduerr, line 141: It looks like this only offers instance **and** general class as suggestions, however it should be clear that it is possible to specify just one of them.

The help output also instructs the user to use it as `--class instance general`, however it should be used as `--class instance,general`.

jeffwindsor: Got it, let me look how to do that and update the PR.  BTW Glad to help

jeffwindsor: @chrisduerr I was able to make the change to show
```
OPTIONS:
        --class <instance>,<general>               Defines window class hint on Linux [default: Alacritty,Alacritty]
```
is this correct or are we looking for below (matching man page)
```
OPTIONS:
        --class [ <instance> | <instance>,<general> ]               Defines window class hint on Linux [default: Alacritty,Alacritty]
```

chrisduerr: You've fixed 50% of the existing problems. We want to match the manpage.

alacritty.yml
```diff
@@ -83,8 +83,7 @@
   # Specifying '0' will disable scrolling.
   #history: 10000
 
-  # Number of lines the viewport will move for every line scrolled when
-  # scrollback is enabled (history > 0).
+  # Acceleration multiplier for mouse scrolling when scrollback is enabled (history > 0).
   #multiplier: 3
```
chrisduerr, line 87: This change seems unrelated?

jeffwindsor: man bad guys will pull that back out

alacritty/src/cli.rs
jeffwindsor: ok one min will try to get the other 50%

jeffwindsor:
@chrisduerr, I am very new to rust so maybe I am not testing it in the same way.  I tested it manually by running `cargo run -- --help` on my branch and that produced
```
OPTIONS:
        --class <instance>|<instance>,<general>    Defines window class hint on Linux [default: Alacritty,Alacritty]
    -e, --command <command>...                     Command and args to execute (must be last argument)
```
And yes, I agree I am definitely bending the clap usage.  I was not able to find a way to use clap to create documentation for  instance OR instance, general like in the man page.

Any ideas would be appreciated.

chrisduerr:
>  I tested it manually by running cargo run -- --help on my branch and that produced

You should test that it works too, not only that the help is correct.

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Regression in font rendering on macOS
 - Scroll down escape (`CSI Ps T`) incorrectly pulling lines from history
 - Dim escape (`CSI 2 m`) support for truecolor text
+- Docs for class in --help missing information on setting general class
```
kchibisov, line 55:
```suggestion
- Documentation for class in `--help` missing information on setting general class
```

extra/alacritty.man
```diff
@@ -42,8 +42,8 @@ Increases the level of verbosity (the max level is \fB\-vvv\fR)
 Prints version information
 .SH "OPTIONS"
 .TP
-\fB\-\-class\fR [ <instance> | <instance>,<general> ]
-Defines the window class hint on Linux [default: Alacritty,Alacritty ]
+\fB\-\-class\fR <instance> | <instance>,<general>
+Defines the window class hint on Linux [default: Alacritty,Alacritty]
```
kchibisov, line 46: I feel like we should mention that on Wayland it sets `app_id` and in `--help` as well, since it's not something obvious and could make it clear for folks. Or even add `--app_id` cli argument(not sure that we really should though). What do you think?

alacritty/src/cli.rs
```diff
@@ -122,7 +124,9 @@ impl Options {
             .arg(
                 Arg::with_name("class")
                     .long("class")
+                    .value_name("instance> | <instance>,<general")
```
kchibisov, line 127:
```suggestion
                    .value_name("instance> | <instance>,<general>")
```

extra/alacritty.man
chrisduerr: How does that work on Wayland? Only instance is supported and instance is app_id? General does nothing?

alacritty/src/cli.rs
chrisduerr: This is intentional, it's necessary since clap adds `<...>` around the value by itself with no way to turn that off as far as I'm aware. This is the best solution I could find (it's also why the `<` is missing at the start).

extra/alacritty.man
kchibisov: I mean it's `app+id`, but yeah, only one thing exists, nothing more.

CHANGELOG.md
```diff
@@ -53,6 +53,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Scroll down escape (`CSI Ps T`) incorrectly pulling lines from history
 - Dim escape (`CSI 2 m`) support for truecolor text
 - Incorrectly deleted lines when increasing width with a prompt wrapped using spaces
+- Docs for class in --help missing information on setting general class
+- Documentation for class in `--help` missing information on setting general class
```
kchibisov, line 57: borken merge/rebase,etc.

extra/alacritty.man
```diff
@@ -42,8 +42,10 @@ Increases the level of verbosity (the max level is \fB\-vvv\fR)
...
+
+On Wayland the instance class sets the `app+id`, while the general class is ignored.
```
kchibisov, line 48: I'm sorry, I've pressed some magic button on my keyboard, it's `app_id`.

alacritty/src/cli.rs
```diff
@@ -122,8 +124,10 @@ impl Options {
...
                     .takes_value(true)
-                    .help(&format!("Defines window class on Linux [default: {}]", DEFAULT_NAME)),
+                    .use_delimiter(true)
+                    .help(&format!("Defines window class or app+id on Linux [default: {}]", DEFAULT_NAME)),
```
kchibisov, line 130: same here `app_id`, sorry.

PR #3649, opened 2020-04-26
scrolling:multiplier doc clarification. resolves #3189

### LINKED ISSUE #3189: Comment in default config about scrolling is incorrect
The default configuration file says the following:
```
# Number of lines the viewport will move for every line scrolled when
# scrollback is enabled (history > 0).
#multiplier: 3

```

As you can see it's a number of `lines`, while it's accurate when you're using a mouse, on a `touchpad` you just changing the speed of accumulating events, and you'll still be able to scroll **exactly** by one line, even having `multiplier: 5`, while on mouse it's impossible.

So I see two possible solutions here: make touchpad to match a mouse behavior, so you should accumulate scrolling as always, but multiply afterwards, or we should update a comment in default configuration file. To say, that it's a multiplier for scrolling events, and remove the parts about lines.

jeffwindsor:

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -83,7 +83,7 @@
   # Specifying '0' will disable scrolling.
   #history: 10000
 
-  # Modifier for multiplying the mouse scrolling distance.
+  # Scrolling distance multiplier.
   #multiplier: 3
 
 # Font configuration
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -83,8 +83,7 @@
   # Specifying '0' will disable scrolling.
   #history: 10000
 
-  # Acceleration for mouse scrolling, number of lines the viewport will move for
-  # every line scrolled when scrollback is enabled (history > 0).
+  # Modifier for multiplying the mouse scrolling distance.
   #multiplier: 3
 
 # Font configuration
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -83,8 +83,8 @@
   # Specifying '0' will disable scrolling.
   #history: 10000
 
-  # Number of lines the viewport will move for every line scrolled when
-  # scrollback is enabled (history > 0).
+  # Acceleration for mouse scrolling, number of lines the viewport will move for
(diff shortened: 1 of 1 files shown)


alacritty.yml
```diff
@@ -83,8 +83,8 @@
   # Specifying '0' will disable scrolling.
   #history: 10000
 
-  # Number of lines the viewport will move for every line scrolled when
-  # scrollback is enabled (history > 0).
+  # Acceleration for mouse scrolling, number of lines the viewport will move for
+  # every line scrolled when scrollback is enabled (history > 0).
```
kchibisov, line 87: This is still not entirely true, I can move by 1 line with `multiplier: 10`, it's acceleration for mouse scrolling events in short, but I don't have good wording here.

jeffwindsor: How about just:
"Mouse scrolling acceleration"

kchibisov: I guess it's better, but I still think that it could be strange wording. what do you think @chrisduerr ?

chrisduerr: Not a huge fan, since this doesn't correctly represent that this is more than just an abstract number. Acceleration also sounds like mouse acceleration, so bigger movements, bigger outcome. But thankfully that's not the case (at least on Linux/Windows).

I'd suggest maybe something like the following?

```
Modifier for multiplying the mouse scrolling distance.
```

I'd definitely stay away from the word acceleration at least.

alacritty.yml
```diff
@@ -83,8 +83,7 @@
   # Specifying '0' will disable scrolling.
   #history: 10000
 
-  # Number of lines the viewport will move for every line scrolled when
-  # scrollback is enabled (history > 0).
+  # Modifier for multiplying the mouse scrolling distance.
```
chrisduerr, line 86:
```suggestion
  # Scrolling distance multiplier.
```

I've conferred with @kchibisov on IRC and we think just keeping it simple is best.

The `mouse` in my suggestion was incorrect, so it was no good since this obviously applies to the touchpad too.

PR #3660, opened 2020-04-30
Call glFinish right after swap_buffers on X11

### LINKED ISSUE #3061: Weird resize lag when launching Alacritty in tiling mode
Starting version 0.4.0 I seem to have this weird resize lag issue when launching Alacritty in tiling mode. It seems that when I try to launch it, it automatically tries to use the configured `lines` and `columns` for the window size. Then resizes itself to fit the whole tiled window (which is often obviously larger than the configured `lines` and `column` sizes).

This creates the issue that when you launch the program, you first see it sized in its configured `lines` and `columns`, then you see it resize itself momentarily. In previous versions this wasn't the case. It sorta just used whatever the window manager tells it to. And pops right into existence with the correct window dimensions (despite having configured `lines` and `columns`).

In previous versions, the `lines` and `columns` seems to be used only when the window manager tells it that its floating. But now in 0.4.0, it seems like it insists on using those two settings on startup even when tiled (but then just resizes itself moments later to adhere to the window manager calls).

Here's a video recording of the issue for reference: [https://imgur.com/a/VSzK69z](https://imgur.com/a/VSzK69z)

You can ignore the tearing on the video as that seems to be part of the recording. What you should notice is how in the previous version, the terminal sorta just pops right in the screen. Whereas now in 0.4.0, it's now a small little window first (or whatever size you give it), then it resizes itself.

That's not really an issue when you just open one or two terminals, then use it for the whole day. But when you have to open like 50 terminals throughout the day for the stuff you do, it becomes really annoying.

I'm running my X server with xf86-video-intel, and testing against the modesetting driver, the issues seems to persist. Telling the xf86-video-intel driver to use DRI2 (instead of DRI3) seems to make the issue go away. Though I was hoping that I wouldn't have switch to DRI2 for my X server, because it breaks a bunch of my mpv settings. This is all running on a ThinkPad X220 with Intel HD 3000 graphics.

# System

| | |
|------------------------|---------|
| Operating System | Linux |
| Alacritty Version | alacritty 0.4.0 |
| Display Server | X11 |
| Window Manager | bspwm |
| Compositor | picom |

# Logs

Some probably relevant startup logs:

| Command | Link |
|---------------------------|------------------------------------------------------------------------------------|
| `alacritty -vv` (0.3.3) | [https://pastebin.com/yTwV8wSy](https://pastebin.com/yTwV8wSy) |
| `alacritty -vv` (0.4.0) | [https://pastebin.com/uHHPCKxu](https://pastebin.com/uHHPCKxu) |

kchibisov:
It was find out that on X11 swap_buffers doesn't block with vsync on most
systems, so the block will be on the next gl command (glClear on the next
event loop tick for us), which introduces a permament one frame of delay

Fixes #3061.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Tabstops not being reset with `reset`
 - Selection not cleared when switching between main and alt grid
 - Fallback to `LC_CTYPE=UTF-8` on macOS without valid system locale
+- Resize lag on launch under some X11 wms
 
 ## 0.4.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Selection not cleared when switching between main and alt grid
 - Fallback to `LC_CTYPE=UTF-8` on macOS without valid system locale
 - Resize lag on launch under some X11 wms
+- Increased input latency due to vsync behavior on X11
 
 ## 0.4.2
 
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -513,9 +513,9 @@ impl Display {
         #[cfg(not(any(target_os = "macos", windows)))]
         {
             if self.is_x11 {
-                // It was find out that on X11 swap_buffers doesn't block with vsync on most
-                // systems, so the block will be on the next gl command (glClear on the next event
-                // loop tick for us), which introduces a permament one frame of delay
+                // On X11 `swap_buffers` does not block for vsync. However the next OpenGl command
+                // will block to synchronize (this is `glClear` in Alacritty), which causes a
+                // permanent one frame delay.
                 self.renderer.with_api(&config, &size_info, |api| {
                     api.finish();
                 });
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -118,6 +118,8 @@ pub struct Display {
     renderer: QuadRenderer,
     glyph_cache: GlyphCache,
     meter: Meter,
+    #[cfg(not(any(target_os = "macos", windows)))]
+    is_x11: bool,
 }
 
 impl Display {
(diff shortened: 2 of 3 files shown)


kchibisov:
This thing isn't required on Wayland at all, and likely on all other platforms, however it'll still be interesting to test swap_buffers block  on them, and where we actually could block in `fn draw`.

chrisduerr:
I'd definitely like to test this on Windows/macOS before merging. I can test macOS unless @casperstorm wants to help out. Maybe @davidhewitt / @cole-h could test on Windows?

chrisduerr:
As correctly noted by @kchibisov, what should be tested isn't this patch but the following one (on top of master):

```diff
diff --git a/alacritty/src/display.rs b/alacritty/src/display.rs
index eafab8c..17a3235 100644
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -498,7 +498,9 @@ impl Display {
             });
         }
 
+        let start = std::time::Instant::now();
         self.window.swap_buffers();
+        println!("BUFFER SWAP TIME: {:?}", std::time::Instant::now().duration_since(start));
     }
 }
 ```

The question is if this `swap_buffers` call blocks on vsync or not, so on a 60Hz monitor it should be around 16ms. If you run your monitor at a different refresh rate and want to test, please also tell me your refresh rate.

**Comment by chrisduerr** [comment]
Thanks to @cole-h and @casperstorm we were able to determine that this issue is only present on X11.

File: CHANGELOG.md [hunk 1]
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Tabstops not being reset with `reset`
 - Selection not cleared when switching between main and alt grid
+- Resize lag on launch under some X11 wms
```
chrisduerr, line 31: Might be good to add another comment about input latency improvement by one frame on X11?

alacritty/src/display.rs
```diff
@@ -499,6 +502,18 @@ impl Display {
         }
 
         self.window.swap_buffers();
+
+        #[cfg(not(any(target_os = "macos", windows)))]
+        {
+            if self.window.x11_window_id().is_some() {
+                // It was find out that on X11 swap_buffers doesn't block with vsync on most
+                // systems, so the block will be on the next gl command (glClear on the next event
+                // loop tick for us), which introduces a permament one frame of delay
```
chrisduerr, line 511:
```suggestion
                // On X11 `swap_buffers` does not block for vsync. However the next OpenGl command
                // will block to synchronize (this is `glClear` in Alacritty), which causes a
                // permanent one frame delay.
```

CHANGELOG.md
chrisduerr: Maybe something like this would work:

```
 - Increased input latency due to vsync behavior on X11
```

alacritty/src/display.rs
chrisduerr: This is still relevant.

Issue #2761, opened 2019-08-27
Don't hide cursor on modifiers press


kchibisov:
We generally want to avoid hiding cursor on modifiers press, since most of the time, it's continuation of mouse interaction like pressing URL modifier when mouse cursor is already over URL or pressing shift in mouse mode apps.

chrisduerr:
Generally I do think this does make some sense, however it's tough to say which modifier is actually a modifier change. Pressing lshift after rshift does not actually change modifier state for example, so we would make some assumptions here.

Generally I think it should be fine not to hide the cursor whenever a key is pressed that does not send any characters though.

PR #3661, opened 2020-04-30
Don't hide cursor on modifier press

Dav1dde:
Fixes: #2761.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Wayland client side decorations are now based on config colorscheme
 - Low resolution window decoration icon on Windows
 - Mouse bindings for additional buttons need to be specified as a number not a string
+- Don't hide cursor on modifier press with `mouse.hide_when_typing` enabled
 
 ### Fixed
 
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -590,12 +590,6 @@ impl<N: Notify + OnResize> Processor<N> {
                     },
                     WindowEvent::KeyboardInput { input, is_synthetic: false, .. } => {
                         processor.key_input(input);
-                        if input.state == ElementState::Pressed {
-                            // Hide cursor while typing.
-                            if processor.ctx.config.ui_config.mouse.hide_when_typing {
-                                processor.ctx.window.set_mouse_visible(false);
-                            }
-                        }
                     },
                     WindowEvent::ReceivedCharacter(c) => processor.received_char(c),
                     WindowEvent::MouseInput { state, button, .. } => {
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -181,7 +181,7 @@ impl<T: EventListener> Execute<T> for Action {
                 }
 
                 ctx.terminal_mut().vi_motion(motion)
-            }
+            },
             Action::ToggleFullscreen => ctx.window_mut().toggle_fullscreen(),
             #[cfg(target_os = "macos")]
             Action::ToggleSimpleFullscreen => ctx.window_mut().toggle_simple_fullscreen(),
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -128,6 +128,10 @@ impl<T: EventListener> Execute<T> for Action {
     fn execute<A: ActionContext<T>>(&self, ctx: &mut A) {
         match *self {
             Action::Esc(ref s) => {
+                if ctx.config().ui_config.mouse.hide_when_typing {
+                    ctx.window_mut().set_mouse_visible(false);
+                }
+
                 ctx.clear_selection();
                 ctx.scroll(Scroll::Bottom);
                 ctx.write_to_pty(s.clone().into_bytes())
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Preserve selection on non-LMB or mouse mode clicks
 - Wayland client side decorations are now based on config colorscheme
 - Low resolution window decoration icon on Windows
+- Don't hide cursor on modifier press (mouse.hide_when_typing setting)
```
chrisduerr, line 27:
```suggestion
- Don't hide cursor on modifier press with `mouse.hide_when_typing` enabled
```

This is about the same length but reads a bit nicer I think.

alacritty/src/event.rs
```diff
@@ -554,14 +554,23 @@ impl<N: Notify + OnResize> Processor<N> {
...
+                        }
+                    },
+                    WindowEvent::ReceivedCharacter(c) => {
+                        processor.received_char(c);
+                        // Hide cursor while typing
+                        if processor.ctx.config.ui_config.mouse.hide_when_typing
+                            && !*processor.ctx.suppress_chars()
+                        {
+                            processor.ctx.window.set_mouse_visible(false);
                         }
```
chrisduerr, line 572: Why do we do this here instead of the `received_char` method? Seems like this would just duplicate the `suppress_chars` check, which seems inefficient.

alacritty/src/event.rs
```diff
@@ -554,14 +554,23 @@ impl<N: Notify + OnResize> Processor<N> {
...
                         processor.key_input(input);
-                        if input.state == ElementState::Pressed {
-                            // Hide cursor while typing
-                            if processor.ctx.config.ui_config.mouse.hide_when_typing {
-                                processor.ctx.window.set_mouse_visible(false);
-                            }
+                        // Hide cursor while typing, when triggering a keybind that writes to pty
+                        if input.state == ElementState::Pressed
+                            && processor.ctx.config.ui_config.mouse.hide_when_typing
+                            && *processor.ctx.received_count() > 0
```
chrisduerr, line 560: This is not what `received_count` is for and it will break the existing `received_char` method.

Why is this not done directly in `input.rs`?

Dav1dde:
@chrisduerr thanks for the pointers, I moved the code over to `input.rs`. 

Maybe we should check for more actions than just `Action::Esc`, e.g. pasting does not hide the cursor (which feels okay for me).

I also split the initial check for suppress and vi mode in `received_char`, so the cursor is hidden if we receive a char in vi mode, I think that makes sense?

chrisduerr:
> Should we hide cursor on Vi mode actions, like moving around, etc.

We probably should.

Dav1dde:
> > Should we hide cursor on Vi mode actions, like moving around, etc.
> 
> We probably should.

The current implementation does that, or am I missing something?

kchibisov:
When I've checked it wasn't doing that for me, I can recheck though.

alacritty/src/input.rs
```diff
@@ -710,7 +710,15 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
-        if *self.ctx.suppress_chars() || self.ctx.terminal().mode().contains(TermMode::VI) {
+        if *self.ctx.suppress_chars() {
+            return;
+        }
+
+        if self.ctx.config().ui_config.mouse.hide_when_typing {
+            self.ctx.window_mut().set_mouse_visible(false);
+        }
+
+        if self.ctx.terminal().mode().contains(TermMode::VI) {
```
chrisduerr, line 721: I'm not a huge fan of this approach, it's kinda messy. 

I think it would be better to explicitly hide the mouse courser when the `ViMotion` command is triggered.

alacritty/src/input.rs
```diff
@@ -768,6 +776,19 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
+                    let hide_cursor = match binding.action {
+                        Action::Esc(_) => true,
+                        Action::ViMotion(_) => true,
+                        _ => false
+                    };
+
+                    if hide_cursor {
+                        self.ctx.window_mut().set_mouse_visible(false);
+                    }
+                }
```
chrisduerr, line 791: Why does this need to be done here instead of the action handlers?

Dav1dde: Good point, moved

alacritty/src/input.rs
Dav1dde: This here exists for keys pressed that are *not* bound to an action. If an action is triggered `suppress_chars` is `true`. But if there is no action and we're in vi mode the cursor gets hidden, if that behaviour is not desired, I can "merge" the `suppress_chars` and vi mode check back together.

Dav1dde:
Test failure seems unrelated:

```
The command "ci/script.sh" exited with 101.
error: failed to run custom build command for `winpty v0.2.0`
```

alacritty/src/input.rs
kchibisov: I think we should, it'll also simplify Vi mode cursor hide handling.

Dav1dde: Merged

PR #3665, opened 2020-05-01
Render colored emojis' true colors

### LINKED ISSUE #1864: Always render emojis with default color

lbonn:
As they have an alpha channel, we can use it directly for blending
instead of applying the normal text fragment shader that blends it with
the text color and ends up mixing with background as well (particularly
bad on bright background).

Instead of branching on data in the shader, it and storing an alpha
channel for the complete Atlas, it might be wiser to have a separate
Atlas and rendering phase for these glyphs.

Fixes #1864.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Resize lag on launch under some X11 wms
 - Increased input latency due to vsync behavior on X11
 - Freeze when application is invisible on Wayland
+- Emoji colors blending with terminal background
 
 ## 0.4.2
 
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -418,6 +418,8 @@ struct InstanceData {
     bg_g: f32,
     bg_b: f32,
     bg_a: f32,
+    // boolean set to true if the glyph already is colored (typically RGBA emojis)
+    pre_colored: u8,
 }
 
 #[derive(Debug)]
@@ -465,18 +467,11 @@ impl Batch {
         Self { tex: 0, instances: Vec::with_capacity(BATCH_MAX) }
     }
 
-    pub fn add_item(&mut self, mut cell: RenderableCell, glyph: &Glyph) {
+    pub fn add_item(&mut self, cell: RenderableCell, glyph: &Glyph) {
         if self.is_empty() {
             self.tex = glyph.tex_id;
         }
 
-        if glyph.colored {
-            // XXX Temporary workaround to prevent emojis being rendered with a wrong colors on, at
-            // least, dark backgrounds. For more info see #1864.
-            cell.fg.r = 255;
-            cell.fg.g = 255;
-            cell.fg.b = 255;
-        }
 
         self.instances.push(InstanceData {
             col: cell.column.0 as f32,
@@ -500,6 +495,7 @@ impl Batch {
             bg_g: f32::from(cell.bg.g),
             bg_b: f32::from(cell.bg.b),
             bg_a: cell.bg_alpha,
+            pre_colored: glyph.colored as u8,
         });
     }
 
@@ -643,6 +639,17 @@ impl QuadRenderer {
             );
             gl::EnableVertexAttribArray(4);
             gl::VertexAttribDivisor(4, 1);
+            // preColoredGlyph attribute
+            gl::VertexAttribPointer(
+                5,
+                1,
+                gl::BYTE,
+                gl::FALSE,
+                size_of::<InstanceData>() as i32,
+                (17 * size_of::<f32>()) as *const _,
+            );
+            gl::EnableVertexAttribArray(5);
+            gl::VertexAttribDivisor(5, 1);
 
             // Rectangle setup.
             gl::GenVertexArrays(1, &mut rect_vao);
@@ -1513,14 +1520,17 @@ impl Atlas {
             gl::PixelStorei(gl::UNPACK_ALIGNMENT, 1);
             gl::GenTextures(1, &mut id);
             gl::BindTexture(gl::TEXTURE_2D, id);
+            // We use an RGBA atlas that can handle colored glyphs.
+            // As they usually are outnumbered by regular text glyphs, this is slightly wasteful.
+            // Another way would be to have a separate atlas for RGBA and render in two passes.
             gl::TexImage2D(
                 gl::TEXTURE_2D,
                 0,
-                gl::RGB as i32,
+                gl::RGBA as i32,
                 size,
                 size,
                 0,
-                gl::RGB,
+                gl::RGBA,
                 gl::UNSIGNED_BYTE,
                 ptr::null(),
             );
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -418,7 +418,7 @@ struct InstanceData {
     bg_g: f32,
     bg_b: f32,
     bg_a: f32,
-    // boolean set to true if the glyph already is colored (typically RGBA emojis)
+    // Colored glyph indicator.
     colored: u8,
 }
 
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -472,7 +472,6 @@ impl Batch {
             self.tex = glyph.tex_id;
         }
 
-
         self.instances.push(InstanceData {
             col: cell.column.0 as f32,
             row: cell.line.0 as f32,
@@ -1520,7 +1519,8 @@ impl Atlas {
             gl::PixelStorei(gl::UNPACK_ALIGNMENT, 1);
             gl::GenTextures(1, &mut id);
             gl::BindTexture(gl::TEXTURE_2D, id);
-            // Use RGBA texture for both normal and emoji glyphs, since it has no performance impact.
+            // Use RGBA texture for both normal and emoji glyphs, since it has no performance
+            // impact.
             gl::TexImage2D(
                 gl::TEXTURE_2D,
                 0,
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -419,7 +419,7 @@ struct InstanceData {
     bg_b: f32,
     bg_a: f32,
     // boolean set to true if the glyph already is colored (typically RGBA emojis)
-    pre_colored: u8,
+    colored: u8,
 }
 
 #[derive(Debug)]
@@ -495,7 +495,7 @@ impl Batch {
             bg_g: f32::from(cell.bg.g),
             bg_b: f32::from(cell.bg.b),
             bg_a: cell.bg_alpha,
-            pre_colored: glyph.colored as u8,
+            colored: glyph.colored as u8,
         });
     }
 
@@ -639,7 +639,7 @@ impl QuadRenderer {
             );
             gl::EnableVertexAttribArray(4);
             gl::VertexAttribDivisor(4, 1);
-            // preColoredGlyph attribute
+            // coloredGlyph attribute
             gl::VertexAttribPointer(
                 5,
                 1,
@@ -1521,8 +1521,9 @@ impl Atlas {
             gl::GenTextures(1, &mut id);
             gl::BindTexture(gl::TEXTURE_2D, id);
             // We use an RGBA atlas that can handle colored glyphs.
-            // As they usually are outnumbered by regular text glyphs, this is slightly wasteful.
-            // Another way would be to have a separate atlas for RGBA and render in two passes.
+            // Most of the text glyphs don't need this alpha channel but it has a negligible impact in practice.
+            // For more details and comparative benchmarks with a dual atlas approach, see:
+            // https://github.com/alacritty/alacritty/pull/3665
             gl::TexImage2D(
                 gl::TEXTURE_2D,
                 0,
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -418,8 +418,8 @@ struct InstanceData {
     bg_g: f32,
     bg_b: f32,
     bg_a: f32,
-    // Colored glyph indicator.
-    colored: u8,
+    // Flag indicating that glyph uses multiple colors, like an Emoji.
+    multicolor: u8,
 }
 
 #[derive(Debug)]
@@ -495,7 +495,7 @@ impl Batch {
             bg_g: f32::from(cell.bg.g),
             bg_b: f32::from(cell.bg.b),
             bg_a: cell.bg_alpha,
-            colored: glyph.colored as u8,
+            multicolor: glyph.colored as u8,
         });
     }
 
@@ -639,7 +639,7 @@ impl QuadRenderer {
             );
             gl::EnableVertexAttribArray(4);
             gl::VertexAttribDivisor(4, 1);
-            // coloredGlyph attribute
+            // Multicolor flag.
             gl::VertexAttribPointer(
                 5,
                 1,
@@ -1520,10 +1520,7 @@ impl Atlas {
             gl::PixelStorei(gl::UNPACK_ALIGNMENT, 1);
             gl::GenTextures(1, &mut id);
             gl::BindTexture(gl::TEXTURE_2D, id);
-            // We use an RGBA atlas that can handle colored glyphs.
-            // Most of the text glyphs don't need this alpha channel but it has a negligible impact in practice.
-            // For more details and comparative benchmarks with a dual atlas approach, see:
-            // https://github.com/alacritty/alacritty/pull/3665
+            // Use RGBA texture for both normal and emoji glyphs, since it has no performance impact.
             gl::TexImage2D(
                 gl::TEXTURE_2D,
                 0,
--- a/font/src/lib.rs
+++ b/font/src/lib.rs
@@ -173,7 +173,7 @@ pub struct RasterizedGlyph {
 
 #[derive(Clone, Debug)]
 pub enum BitmapBuffer {
-    /// RGB pixels.
+    /// RGB alphamask.
     RGB(Vec<u8>),
 
     /// RGBA pixels with premultiplied alpha.
--- a/font/src/lib.rs
+++ b/font/src/lib.rs
@@ -173,7 +173,10 @@ pub struct RasterizedGlyph {
 
 #[derive(Clone, Debug)]
 pub enum BitmapBuffer {
+    /// RGB pixels.
     RGB(Vec<u8>),
+
+    /// RGBA pixels with premultiplied alpha.
     RGBA(Vec<u8>),
 }
 
--- a/res/text.f.glsl
+++ b/res/text.f.glsl
@@ -15,6 +15,7 @@
 in vec2 TexCoords;
 flat in vec3 fg;
 flat in vec4 bg;
+flat in int preColored;
 uniform int backgroundPass;
 
 layout(location = 0, index = 0) out vec4 color;
@@ -31,8 +32,14 @@ void main()
         alphaMask = vec4(1.0);
         color = vec4(bg.rgb, 1.0);
     } else {
-        vec3 textColor = texture(mask, TexCoords).rgb;
-        alphaMask = vec4(textColor, textColor.r);
-        color = vec4(fg, 1.0);
+        if (preColored != 0) {
+            vec4 c = texture(mask, TexCoords);
+            alphaMask = vec4(c.a);
+            color = c;
+        } else {
+            vec3 textColor = texture(mask, TexCoords).rgb;
+            alphaMask = vec4(textColor, textColor.r);
+            color = vec4(fg, 1.0);
+        }
     }
 }
--- a/res/text.f.glsl
+++ b/res/text.f.glsl
@@ -36,7 +36,9 @@ void main()
             // Color glyphs (such as emojis).
             vec4 glyphColor = texture(mask, TexCoords);
             alphaMask = vec4(glyphColor.a);
-            color = vec4(glyphColor.rgb, 1.0);
+            // Blend with background color used in a pass.
+            vec3 blendedColor = glyphColor.rgb + bg.rgb * (1.0 -  glyphColor.a);
+            color = vec4(blendedColor, 1.0);
         } else {
             // Regular text glyphs.
             vec3 textColor = texture(mask, TexCoords).rgb;
--- a/res/text.f.glsl
+++ b/res/text.f.glsl
@@ -37,11 +37,10 @@ void main()
             vec4 glyphColor = texture(mask, TexCoords);
             alphaMask = vec4(glyphColor.a);
 
-            // Handle premultiplied alpha.
-            //
-            // NOTE - we won't show garbage from division by zero on a screen,
-            // due to alpha being 0, so it's perfrectly fine to do so.
-            glyphColor.rgb = vec3(glyphColor.rgb / glyphColor.a);
+            // Revert alpha premultiplication.
+            if (glyphColor.a != 0) {
+                glyphColor.rgb = vec3(glyphColor.rgb / glyphColor.a);
+            }
 
             color = vec4(glyphColor.rgb, 1.0);
         } else {
--- a/res/text.f.glsl
+++ b/res/text.f.glsl
@@ -15,7 +15,7 @@
 in vec2 TexCoords;
 flat in vec3 fg;
 flat in vec4 bg;
-flat in int preColored;
+flat in int colored;
 uniform int backgroundPass;
 
 layout(location = 0, index = 0) out vec4 color;
@@ -32,11 +32,13 @@ void main()
         alphaMask = vec4(1.0);
         color = vec4(bg.rgb, 1.0);
     } else {
-        if (preColored != 0) {
-            vec4 c = texture(mask, TexCoords);
-            alphaMask = vec4(c.a);
-            color = c;
+        if (colored != 0) {
+            // Color glyphs (such as emojis).
+            vec4 glyphColor = texture(mask, TexCoords);
+            alphaMask = vec4(glyphColor.a);
+            color = vec4(glyphColor.rgb, 1.0);
         } else {
+            // Regular text glyphs.
             vec3 textColor = texture(mask, TexCoords).rgb;
             alphaMask = vec4(textColor, textColor.r);
             color = vec4(fg, 1.0);
--- a/res/text.f.glsl
+++ b/res/text.f.glsl
@@ -33,12 +33,17 @@ void main()
         color = vec4(bg.rgb, 1.0);
     } else {
         if (colored != 0) {
-            // Color glyphs (such as emojis).
+            // Color glyphs, like emojis.
             vec4 glyphColor = texture(mask, TexCoords);
             alphaMask = vec4(glyphColor.a);
-            // Blend with background color used in a pass.
-            vec3 blendedColor = glyphColor.rgb + bg.rgb * (1.0 -  glyphColor.a);
-            color = vec4(blendedColor, 1.0);
+
+            // Handle premultiplied alpha.
+            //
+            // NOTE - we won't show garbage from division by zero on a screen,
+            // due to alpha being 0, so it's perfrectly fine to do so.
+            glyphColor.rgb = vec3(glyphColor.rgb / glyphColor.a);
+
+            color = vec4(glyphColor.rgb, 1.0);
         } else {
             // Regular text glyphs.
             vec3 textColor = texture(mask, TexCoords).rgb;
--- a/res/text.v.glsl
+++ b/res/text.v.glsl
@@ -27,9 +27,12 @@ layout (location = 3) in vec3 textColor;
 // Background color
 layout (location = 4) in vec4 backgroundColor;
 
+layout (location = 5) in int preColoredGlyph;
+
 out vec2 TexCoords;
 flat out vec3 fg;
 flat out vec4 bg;
+flat out int preColored;
 
 // Terminal properties
 uniform vec2 cellDim;
@@ -71,4 +74,5 @@ void main()
 
     bg = vec4(backgroundColor.rgb / 255.0, backgroundColor.a);
     fg = textColor / vec3(255.0, 255.0, 255.0);
+    preColored = preColoredGlyph;
 }
--- a/res/text.v.glsl
+++ b/res/text.v.glsl
@@ -27,12 +27,13 @@ layout (location = 3) in vec3 textColor;
 // Background color
 layout (location = 4) in vec4 backgroundColor;
 
-layout (location = 5) in int preColoredGlyph;
+// Set to 1 if the glyph colors should be kept
+layout (location = 5) in int coloredGlyph;
 
 out vec2 TexCoords;
 flat out vec3 fg;
 flat out vec4 bg;
-flat out int preColored;
+flat out int colored;
 
 // Terminal properties
 uniform vec2 cellDim;
@@ -74,5 +75,5 @@ void main()
 
     bg = vec4(backgroundColor.rgb / 255.0, backgroundColor.a);
     fg = textColor / vec3(255.0, 255.0, 255.0);
-    preColored = preColoredGlyph;
+    colored = coloredGlyph;
 }
--- a/res/text.v.glsl
+++ b/res/text.v.glsl
@@ -15,13 +15,13 @@
 // Cell properties.
 layout (location = 0) in vec2 gridCoords;
 
-// glyph properties.
+// Glyph properties.
 layout (location = 1) in vec4 glyph;
 
 // uv mapping.
 layout (location = 2) in vec4 uv;
 
-// text fg color.
+// Text fg color.
 layout (location = 3) in vec3 textColor;
 
 // Background color.
--- a/res/text.v.glsl
+++ b/res/text.v.glsl
@@ -12,22 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 #version 330 core
-// Cell properties
+// Cell properties.
 layout (location = 0) in vec2 gridCoords;
 
-// glyph properties
+// glyph properties.
 layout (location = 1) in vec4 glyph;
 
-// uv mapping
+// uv mapping.
 layout (location = 2) in vec4 uv;
 
-// text fg color
+// text fg color.
 layout (location = 3) in vec3 textColor;
 
-// Background color
+// Background color.
 layout (location = 4) in vec4 backgroundColor;
 
-// Set to 1 if the glyph colors should be kept
+// Set to 1 if the glyph colors should be kept.
 layout (location = 5) in int coloredGlyph;
 
 out vec2 TexCoords;
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -488,8 +488,8 @@ impl Font {
             kCGImageAlphaPremultipliedFirst | kCGBitmapByteOrder32Host,
         );
 
-        // Give the context an opaque, black background.
-        cg_context.set_rgb_fill_color(0.0, 0.0, 0.0, 1.0);
+        // Set background color for graphics context.
+        cg_context.set_rgb_fill_color(0.0, 0.0, 0.0, 0.0);
         let context_rect = CGRect::new(
             &CGPoint::new(0.0, 0.0),
             &CGSize::new(f64::from(rasterized_width), f64::from(rasterized_height)),


alacritty/src/renderer/mod.rs
```diff
@@ -417,6 +417,7 @@ struct InstanceData {
     bg_g: f32,
     bg_b: f32,
     bg_a: f32,
+    colored: i32,
```
kchibisov, line 420: Can we use something lower on memory for boolean flag? I feel like sending i32 is a bit costly, right? However I remember that it wasn't the case on macOS for f16 types somehow in the past.

alacritty/src/renderer/mod.rs
```diff
@@ -1506,14 +1512,17 @@ impl Atlas {
             gl::PixelStorei(gl::UNPACK_ALIGNMENT, 1);
             gl::GenTextures(1, &mut id);
             gl::BindTexture(gl::TEXTURE_2D, id);
+            // We use an RGBA atlas that can handle colored glyphs.
+            // As they usually are outnumbered by regular text glyphs, this is slightly wasteful.
+            // Another way would be to have a separate atlas for RGBA and render in two passes.
```
kchibisov, line 1517: That was my main concern when I was adding RGBA information to solve this issue, so we should at least measure how it affects here to have RGBA instead of RGB, knowing that RGBA is very seldom thing. One more atlas seems like a good idea tbqh.

kchibisov: So I want at least know if one more pass + atlas is better than RGBA for all the thing.

kchibisov: One more `pro` for special atlas, is that colored glyphs require manual downsampling on Linux/BSD, so it saves perf if they will be live in there own atlas, however all of these require benchmarks and reports from GPU.

alacritty/src/renderer/mod.rs
lbonn: I'm quite new with shaders actually. Also was unhappy with that but AFAIK, the choice is between a bool and  int in the glsl but passing booleans this way is apparently forbidden (probably meant to be used with uniform only).

But it looks like changing this field to u8 and keeping `int` in the shader works fine, so I'll switch to that for now.

kchibisov: Yeah, bool is just u8, I was not sure about sending a bit more data.

alacritty/src/renderer/mod.rs
lbonn: Yes, I suppose benchmarking against master and a two-pass version is the next step.

Any pointer on what/how to benchmark would be appreciated. I suppose memory usage and some vtebench thing?

kchibisov: Yes. `vtebench` has all information you need + render timer. But VRAM usage is likely the most interesting here. From what I've seen `vtebench` and `render_timer` were +- the same, however I'm not sure about memory and it's a bit harder to measure it.

lbonn:
I've run a valgrind/massif test with master and the current PR. Heap memory moves from 29.6MB to 29.9MB during a test printing successively a text version of this page https://eosrei.github.io/emojione-color-font/full-demo.html and this text file https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt.

Will try to write a two-pass version next to compare with.

kchibisov:
The thing is that it's VRAM and not regular RAM, I guess they(valgrind/massif) are not looking into it here, right?

lbonn:
Ah yes I see, my bad.

My setup is an Haswell gpu on Linux. After a quick search, looks like there is not a lot of options for benchmarking. intel_gpu_top doesn't show memory information and gputop doesn't show anything about my gpu (might give it another try)

There is also a new lightly document tool from intel https://gitlab.freedesktop.org/drm/igt-gpu-tools/-/tree/master/tools/i915-perf
I've made a recording but it just spins forever when trying to read it.

chrisduerr:
I know Mesa's HUD can show VRAM usage of an application. Something like `GALLIUM_HUD="requested-VRAM+VRAM-usage" alacritty` should work?

kchibisov:
It seems like it's identical to master on memory usage according to the HUD?

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Tabstops not being reset with `reset`
 - Selection not cleared when switching between main and alt grid
 - Fallback to `LC_CTYPE=UTF-8` on macOS without valid system locale
+- Emoji colors do not blend with background and are displayed in their true colors
```
```suggestion
- Emoji colors blending with terminal background
```

We should state what was fixed, the way it's currently written it sounds a bit like it was fixed that they do not blend, but we fixed that it was blending incorrectly.

alacritty/src/renderer/mod.rs
```diff
@@ -417,6 +417,7 @@ struct InstanceData {
     bg_g: f32,
     bg_b: f32,
     bg_a: f32,
+    colored: u8,
```
chrisduerr, line 420: Adding a comment above this for uniformity would probably be a good idea.

alacritty/src/renderer/mod.rs
```diff
@@ -499,6 +493,7 @@ impl Batch {
             bg_g: f32::from(cell.bg.g),
             bg_b: f32::from(cell.bg.b),
             bg_a: cell.bg_alpha,
+            colored: glyph.colored as u8,
```
chrisduerr, line 496: I'm wondering if `colored` is the correct name for this. Since every glyph is technically colored. The only difference is that some glyphs are multi-colored.

kchibisov:
@lbonn If you decide to push colored render altas, open one more PR, it seems like we can just include this version since it's simple and works the same as the current master. We'll test macOS a bit more Linux cards, but it's working the same on AMDGPU on Linux already.

lbonn:
@kchibisov nice! Thanks for reviewing this so quickly.

I'll try to see if he can write the other version without to much headache. And maybe find a machine where I can actually run benchmarks :).

alacritty/src/renderer/mod.rs
lbonn: Good point. Maybe pre_colored, as in pre-colored glyph?

CHANGELOG.md
lbonn: Applied

alacritty/src/renderer/mod.rs
lbonn: Done, should also be more explicit with the new field name.

kchibisov:
>And maybe find a machine where I can actually run benchmarks :).

Don't worry much about benchmarks, since we'll run them anyway here ourselves before merge.

chrisduerr:
> Don't worry much about benchmarks, since we'll run them anyway here ourselves before merge.

I'm not sure if I would agree with this, as far as I can tell performance is basically the only main remaining issue?

kchibisov:
>I'm not sure if I would agree with this, as far as I can tell performance is basically the only main remaining issue?

I'm that I'm fine to benchmark on this PR, if setting up benchmarks is not something reliable.

The only thing this PR is requires is "Colored Atlas" approach, since this PR has a VRAM concerns now on some systems.

lbonn:
> The only thing this PR is requires is "Colored Atlas" approach, since the VRAM usage is a bit high here on some systems.

Ah I understood that the numbers were ok and that this could be merged as is.
What are the systems where VRAM usage goes up?

kchibisov:
@chrisduerr had VRAM issues, also macOS was a bit slower on your branch, so we need a colored atlas approach to compare and make a decision.

lbonn:
@kchibisov Ok I'll have another try soon then.

Also, I realized that if we don't mix regular glyphs and RGBA, we could also store glyphs and atlas textures using a single color component (GL_RED). This could help even more with ram usage.

lbonn:
Here is the more complicated version. Some details are still a bit dirty, but it should be enough to run benchmarks.

Some run on wayland and integrated intel gpu:

```
# master
 Performance counter stats for 'cat target/benchmarks/alt-screen-random-write.vte' (10 runs):

          1,562.90 msec task-clock                #    0.278 CPUs utilized            ( +-  0.74% )
            86,137      context-switches          #    0.055 M/sec                    ( +-  0.04% )
               819      cpu-migrations            #    0.524 K/sec                    ( +-  1.82% )
                96      page-faults               #    0.062 K/sec                    ( +-  0.41% )
     4,621,115,044      cycles                    #    2.957 GHz                      ( +-  0.68% )
     7,837,805,516      instructions              #    1.70  insn per cycle           ( +-  0.01% )
     3,051,755,024      branches                  # 1952.626 M/sec                    ( +-  0.00% )
        13,156,721      branch-misses             #    0.43% of all branches          ( +-  0.06% )

            5.6164 +- 0.0206 seconds time elapsed  ( +-  0.37% )

# correct color simple
 Performance counter stats for 'cat target/benchmarks/alt-screen-random-write.vte' (10 runs):

          1,558.32 msec task-clock                #    0.276 CPUs utilized            ( +-  0.50% )
            86,148      context-switches          #    0.055 M/sec                    ( +-  0.01% )
               711      cpu-migrations            #    0.456 K/sec                    ( +-  2.77% )
                95      page-faults               #    0.061 K/sec                    ( +-  0.31% )
     4,624,182,592      cycles                    #    2.967 GHz                      ( +-  0.49% )
     7,837,249,880      instructions              #    1.69  insn per cycle           ( +-  0.01% )
     3,051,669,303      branches                  # 1958.312 M/sec                    ( +-  0.00% )
        13,155,453      branch-misses             #    0.43% of all branches          ( +-  0.07% )

            5.6428 +- 0.0219 seconds time elapsed  ( +-  0.39% )

# double pass glyphs
Performance counter stats for 'cat target/benchmarks/alt-screen-random-write.vte' (10 runs):

          1,546.39 msec task-clock                #    0.278 CPUs utilized            ( +-  0.48% )
            86,091      context-switches          #    0.056 M/sec                    ( +-  0.01% )
               801      cpu-migrations            #    0.518 K/sec                    ( +-  2.47% )
                96      page-faults               #    0.062 K/sec                    ( +-  0.81% )
     4,567,975,640      cycles                    #    2.954 GHz                      ( +-  0.41% )
     7,836,852,206      instructions              #    1.72  insn per cycle           ( +-  0.01% )
     3,051,685,241      branches                  # 1973.426 M/sec                    ( +-  0.00% )
        13,185,732      branch-misses             #    0.43% of all branches          ( +-  0.09% )

            5.5665 +- 0.0281 seconds time elapsed  ( +-  0.50% )
```

I also have another laptop running nouveau. there is no vmem metric in my gallium hud but it can display `inst_executed` (sounds relevant).

On a scene full of emoji:

* master: 1.086M
* simple: 1.393M
* double-pass: 1.182M

xxd /dev/urandom:

* master: 515.3k
* simple: 615k
* double-pass: 572k

alacritty/src/renderer/mod.rs
```diff
@@ -535,7 +544,7 @@ impl Batch {
 }
 
 /// Maximum items to be drawn in a batch.
-const BATCH_MAX: usize = 0x1_0000;
+const BATCH_MAX: usize = 0x1000;
```
kchibisov, line 547: Why the batch size was changed, please revert it back?

alacritty/src/renderer/mod.rs
```diff
@@ -695,24 +704,24 @@ impl QuadRenderer {
             });
         }
 
-        let mut renderer = Self {
+        let mut layers = Vec::new();
```
kchibisov, line 707: Just use `vec!` macro.

alacritty/src/renderer/mod.rs
```diff
@@ -118,8 +118,11 @@ pub struct TextShaderProgram {
 
     /// Background pass flag.
     ///
```
kchibisov, line 120: This comment should be reworked.

res/text.f.glsl
```diff
@@ -24,15 +24,22 @@ uniform sampler2D mask;
 
 void main()
 {
-    if (backgroundPass != 0) {
+    if (pass == 0) {
+        // background
```
kchibisov, line 28: Comments should start with uppercased later and end with a period.

alacritty/src/renderer/mod.rs
lbonn: One batch takes around 4.3M on my setup, so these changes will increase global memory consumption from that amount. I figured that BATCH_MAX was quite big but I then realized that it's probably designed to hold a full screen at once.

I'll change it back.

kchibisov: The thing is that lower batch size will result in massive flickering all over the place, and 4MiB isn't something that bad.

res/text.f.glsl
```diff
@@ -24,15 +24,22 @@ uniform sampler2D mask;
...
         color = vec4(bg.rgb, 1.0);
-    } else {
+    } else if (pass == 1) {
+        // Regular text glyphs.
         vec3 textColor = texture(mask, TexCoords).rgb;
         alphaMask = vec4(textColor, textColor.r);
         color = vec4(fg, 1.0);
+    } else {
+        // Colored glyphs.
+        vec4 c = texture(mask, TexCoords);
```
kchibisov, line 41: I'd be in favor of using name that makes a bit more sense than just `c`.

lbonn: Actually, we can remove this variable entirely (done in last version)

kchibisov:
Also, can you run `rustfmt` on your PR? Just `cargo +nightly fmt` should work.

alacritty/src/renderer/mod.rs
```diff
@@ -420,6 +423,27 @@ struct InstanceData {
...
+    batch: Batch,
+    atlas: Vec<Atlas>,
+}
+
+impl Layer {
+    #[inline]
+    pub fn new() -> Self {
+        Self { current_atlas: 0, active_tex: 0, batch: Batch::new(), atlas: vec![] }
+    }
+}
```
chrisduerr, line 445: This looks like it could be replaced with a default implementation (should auto-derive)?

Could still have new, but that should probably then just call default.

alacritty/src/renderer/mod.rs
```diff
@@ -930,44 +932,50 @@ impl<'a, C> RenderApi<'a, C> {
...
                 gl::UNSIGNED_INT,
                 ptr::null(),
-                self.batch.len() as GLsizei,
+                layer.batch.len() as GLsizei,
             );
-            self.program.set_background_pass(false);
+            self.program.set_pass(match &layer_type {
+                LayerType::ColoredGlyphs => 2,
+                _ => 1,
+            });
```
chrisduerr, line 968: Maybe set_pass should take a `LayerType`? Creating a Background layer is probably a good idea anyways. Then just assign the fixed numbers to the enum and convert it to a number so we don't need to match it.

alacritty/src/renderer/mod.rs
```diff
@@ -1007,16 +1015,22 @@ impl<'a, C> RenderApi<'a, C> {
 
     #[inline]
     fn add_render_item(&mut self, cell: RenderableCell, glyph: &Glyph) {
+        let layer_id =
+            if glyph.colored { LayerType::ColoredGlyphs } else { LayerType::RegularGlyphs };
```
chrisduerr, line 1019:
```suggestion
        let layer_type =
            if glyph.colored { LayerType::ColoredGlyphs } else { LayerType::RegularGlyphs };
```

Seems like layer_type would be the correct name here? Since it's not an ID?

alacritty/src/renderer/mod.rs
```diff
@@ -1007,16 +1015,22 @@ impl<'a, C> RenderApi<'a, C> {
...
+
         // Flush batch if tex changing.
-        if !self.batch.is_empty() && self.batch.tex != glyph.tex_id {
-            self.render_batch();
+        if !batch.is_empty() && batch.tex != glyph.tex_id {
+            self.render_batch(layer_id);
         }
 
-        self.batch.add_item(cell, glyph);
+        let batch = &mut self.layers[layer_id as usize].batch;
```
chrisduerr, line 1027: This seems a bit unfortunate? Are you accessing things this way because self.render_batch complains when batch is made a mutable reference in the first place?

alacritty/src/renderer/mod.rs
```diff
@@ -1088,24 +1102,25 @@ impl<'a, C> RenderApi<'a, C> {
...
-    atlas: &mut Vec<Atlas>,
-    current_atlas: &mut usize,
-    rasterized: &RasterizedGlyph,
-) -> Glyph {
-    // At least one atlas is guaranteed to be in the `self.atlas` list; thus
-    // the unwrap.
+fn load_glyph(layers: &mut Vec<Layer>, rasterized: &RasterizedGlyph) -> Glyph {
+    let (layer, atlas_type) = &mut match rasterized.buf {
+        BitmapBuffer::RGB(_) => (&mut layers[0], gl::RGB),
+        BitmapBuffer::RGBA(_) => (&mut layers[1], gl::RGBA),
```
chrisduerr, line 1108: Hardcoding which atlas type uses which color format here seems like it could cause troubles in future refactorings.

Should the colorspace maybe be stored in the layer? Or/And the layer should maybe be indexed by layer type or colorspace?

alacritty/src/renderer/mod.rs
```diff
@@ -1123,37 +1138,42 @@ fn load_glyph(
...
 impl<'a, C> Drop for RenderApi<'a, C> {
     fn drop(&mut self) {
-        if !self.batch.is_empty() {
-            self.render_batch();
+        if !self.layers[0].batch.is_empty() {
+            self.render_batch(LayerType::RegularGlyphs);
+        }
+        if !self.layers[1].batch.is_empty() {
+            self.render_batch(LayerType::ColoredGlyphs);
         }
```
chrisduerr, line 1177: This doesn't seem very extensible to me. We should iterate over the layers here in a generic fashion rather than having to hard-code all layers manually.

res/text.v.glsl
```diff
@@ -51,7 +51,7 @@ void main()
     // Position of cell from top-left
     vec2 cellPosition = cellDim * gridCoords;
 
-    if (backgroundPass != 0) {
+    if (pass == 0) {
```
chrisduerr, line 54: This could need a comment too, just like the fragment shader.

lbonn:
> I think it might be necessary to write a new test for this. Currently we render everything in batches with the assumption that most things are probably in the first batch. We then commit and switch batches when the texture ID changes.

Adding new tests is probably a good idea.

> However, the string "..." would probably switch batches for every single character and I'm not sure if that's great for performance. So it should definitely be investigated before merging this.

It switches batches only internally, both batches are actually filling up and only sent to opengl when necessary.

I have added a simple trace to track the `render_batch` calls when running this snippet:

```
for k in $(seq 0 $(( ($(tput lines) + 1) * $(tput cols) / 2 ))); do printf '😀a'; done; printf '\n'
```

And I get (when switching to a window full of the pattern.

```
draw batch of size 17077, type RegularGlyphs
draw batch of size 8468, type ColoredGlyphs
```

Not really sure why one size is double the other though.

chrisduerr:
> It switches batches only internally, both batches are actually filling up and only sent through opengl when necessary.

Oh, you're absolutely right of course I forgot about that. In fact, this should actually reduce the number of batch switches required since we have two atlases that are active at the same time and each of them has an active texture.

The batch is only committed when the batch texture id changes ***within*** a layer. So the glyph's texture ID is checked against the batches' texture ID within the layer, not some global active texture ID. So this basically means batches for colored and uncolored are completely separate and colored emojis filling up the main atlas doesn't lead to unnecessary batch rendering with only ascii text. Though in practice that probably doesn't make a difference, since all ascii is always part of the same batch anyways and other unicode is usually a lot rarer and bottlenecked on other things like loading.

chrisduerr:
I'd also like to point out some more things I've noticed while reviewing this PR's performance.

On macOS I evaluated the render time and asrw performance and I wasn't able to measure any difference between any of this PR's commits and the master branch.

On Linux I've tested the latest commit and there were some regressions. The render time seems to be mostly consistent, so there was no change to that. As expected the VRAM usage increased by 4MB. One thing I've noticed though is that both RGB and RGBA atlases both take the same amount of VRAM on my system.

The performance of the asrw benchmark however has been heavily impacted by this. I was able to measure ~10% drop in performance consistently with going from around 5.5s down to 6s. I'm not certain what caused this slowdown, however I can look into it with `perf` if you cannot reproduce it.

kchibisov:
>The performance of the asrw benchmark however has been heavily impacted by this. I was able to measure ~10% drop in performance consistently with going from around 5.5s down to 6s. I'm not certain what caused this slowdown, however I can look into it with perf if you cannot reproduce it.

I can't repro this as well, and I recommend to try different compilers as always. This change can't affect ASRW, because PTY reader and rendering are running on separate threads, rendering can't affect times at all, and you'll process the same amount of events in total anyway. The only thing that could be different on this PR is `render_timer`. So all impact of this PR on ASRW is just rustc things, I guess.

lbonn:
Ok I've rerun some tests with perf (master, HEAD and HEAD^) and the fastest version for me is actually HEAD^ but everything is quite close. HEAD is the slowest.

In my opinion, merging HEAD^ shouldn't hurt much if we could not measure difference in GPU load, even more so that this PR is supposed to fix a bug, not introduce a new feature.

The last version feels actually overkill as it increases code complexity, memory usage and does not yield such impressive benchmark results.

kchibisov:
>In my opinion, merging HEAD^ shouldn't hurt much if we could not measure difference in GPU load, even more so that this PR is supposed to fix a bug, not introduce a new feature.

I kind of agree here, since on some systems it was the same VRAM (like mine).

Also, @chrisduerr had issues with VRAM on initial version of this PR, so that's why the last version actually exists.

chrisduerr:
> I can't repro this as well, and I recommend to try different compilers as always.

Fyi I've tested both stable and nightly and it was consintent on both. However I've tested it a couple of times because I agree with the assessment that this shouldn't be the case and it was resolved after restarting the terminal instances a couple of times. However it was consistent for some time, maybe I hit some bad cores?

I've benchmarked it again a couple of times and while this branch always seemed to be the slowest when there was a difference, there were a couple of instances where they performed identically. So my cores are probably just a bit biased.

> Also, @chrisduerr had issues with VRAM on initial version of this PR, so that's why the last version actually exists.

I've checked this again and the VRAM of the latest commit is definitely higher than the previous one, which makes sense since it creates another atlas. So there's really no way this commit could have improved memory usage unless the colored atlas wasn't spawned until it was needed. But after letting things settle down the previous commit uses about the same memory as master.

However this commit is consistently about 50-100usec faster than the previous one, which is about 3-6% out of 1600usec render time on average.

@casperstorm you've tested the previous version on macOS, right? Would you be willing to give this another test, just as a comparison?

casperstorm:
> @casperstorm you've tested the previous version on macOS, right? Would you be willing to give this another test, just as a comparison?

I’ll test in a few hours an update this reply. This is just a confirmation that I’ll do it.

kchibisov:
>I've checked this again and the VRAM of the latest commit is definitely higher than the previous one, which makes sense since it creates another atlas.

Ehm, ok. IIRC you've mentioned on IRC that the original commit had more VRAM usage, etc.

>However this commit is consistently about 50-100usec faster than the previous one, which is about 3-6% out of 1600usec render time on average.

Nah, can't really repro this. For me it's identical.

Honestly,  I'm in favor of using the first version, which just used `RGBA` for everything, since we should remove our background pass logic anyway at some point, due to bugs coming from it. Adding more `layers` could be a bit hard to achieve that, and knowing that normal users won't ever  overflow even one  atlas I feel like we can just go with first version. I'm not sure how it could change render timer if shaders on both version are very similar, however it does less branching during rendering.

kchibisov:
>However this commit is consistently about 50-100usec faster than the previous one, which is about 3-6% out of 1600usec render time on average.

And how it relates to master's render_times for you? I really don't see much difference, well for me latest commit is 100usec slower on average then the first commit here, and the first commit is entirely identical to master for me.

kchibisov:
After some testing, it turned out that it somehow leaves opaque border around emojis, so it blends with app behind.

picture

You can notice that pink border around the ball, and here for example, when I have green app behind.

picture

Both commits are affected here.

res/text.f.glsl
```diff
@@ -24,15 +24,21 @@ uniform sampler2D mask;
...
         color = vec4(bg.rgb, 1.0);
-    } else {
+    } else if (pass == 1) {
+        // Regular text glyphs.
         vec3 textColor = texture(mask, TexCoords).rgb;
         alphaMask = vec4(textColor, textColor.r);
         color = vec4(fg, 1.0);
+    } else {
+        // Colored glyphs.
+        color = texture(mask, TexCoords);
```
kchibisov, line 41: You should only use the alpha for `alphaMask`, but in the same time you should return `color = color(color.rgb, 1.0)` to solve mentioned issue. So it should look something like that, I guess.
```glsl
vec4 glyphColor = texture(mask, TexCoords);
alphaMask = vec4(color.a);
color = vec4(glyphColor.rgb, 1.0);
```

chrisduerr:
> Adding more layers could be a bit hard to achieve that, and knowing that normal users won't ever overflow even one atlas I feel like we can just go with first version. I'm not sure how it could change render timer if shaders on both version are very similar, however it does less branching during rendering.

As I've originally stated on IRC when I tested the first branch and there were some performance differences, the problem is not about VRAM usage but about render time. The issue is that we're transmitting more data to the renderer, since our instance data is bigger. The more data has to be transmitted for every glyph to be rendered, the slower it will be.

alacritty/src/renderer/mod.rs
```diff
@@ -418,6 +418,8 @@ struct InstanceData {
     bg_g: f32,
     bg_b: f32,
     bg_a: f32,
+    // boolean set to true if the glyph already is colored (typically RGBA emojis)
+    pre_colored: u8,
```
kchibisov: Could you just name it colored?

alacritty/src/renderer/mod.rs
```diff
@@ -1513,14 +1520,17 @@ impl Atlas {
             gl::PixelStorei(gl::UNPACK_ALIGNMENT, 1);
             gl::GenTextures(1, &mut id);
             gl::BindTexture(gl::TEXTURE_2D, id);
+            // We use an RGBA atlas that can handle colored glyphs.
+            // As they usually are outnumbered by regular text glyphs, this is slightly wasteful.
+            // Another way would be to have a separate atlas for RGBA and render in two passes.
```
kchibisov: Could add add explanation there, that one more Altlas approach doesn't give us much benefits over this and link to that PR?

res/text.f.glsl
```diff
@@ -31,8 +32,14 @@ void main()
         alphaMask = vec4(1.0);
         color = vec4(bg.rgb, 1.0);
     } else {
-        vec3 textColor = texture(mask, TexCoords).rgb;
-        alphaMask = vec4(textColor, textColor.r);
-        color = vec4(fg, 1.0);
+        if (preColored != 0) {
+            vec4 c = texture(mask, TexCoords);
+            alphaMask = vec4(c.a);
+            color = c;
```
kchibisov: You should apply the fix I've suggested for the second commit here, and also, don't use one char name here.

res/text.f.glsl
lbonn: Wow, good catch. Thanks!

lbonn:
Ok I've applied the suggested changes but was not able to test them right now (only that it compiles). Will confirm tomorrow morning.

alacritty/src/renderer/mod.rs
```diff
@@ -418,6 +418,8 @@ struct InstanceData {
     bg_g: f32,
     bg_b: f32,
     bg_a: f32,
+    // boolean set to true if the glyph already is colored (typically RGBA emojis)
```
kchibisov, line 421: I feel like it's too long tbh, maybe something like that

```suggestion
    // Colored glyph indicator.
```

res/text.v.glsl
```diff
@@ -27,9 +27,13 @@ layout (location = 3) in vec3 textColor;
 // Background color
 layout (location = 4) in vec4 backgroundColor;
 
+// Set to 1 if the glyph colors should be kept
```
kchibisov, line 30:
```suggestion
// Set to 1 if the glyph colors should be kept.
```

alacritty/src/renderer/mod.rs
kchibisov: Glyphs are monochrome or colored, and you color them in rendering, so technically, `colored` is a fine.

lbonn:
@kchibisov I've tested the blending on both versions, and it does not look perfect either.

blending

The emoji on the left was rendered with my original shader, and on the right with your changes, both of them with a violet window in the background.

It looks like the emoji on the right is blending with some dark color (black from the glyph data?) instead of the bright terminal background.

kchibisov:
@lbonn Could you repro on a very big emoji, when it stops increasing, since I can't do that on it.

kchibisov:
Actually it's being blended correctly, if you disable blend function you'll get the following. So it seems like it's working fine. However the issue could be somewhere in a freetype and our downsampling approach, since we downsample them manually.

picture

lbonn:
> @lbonn Could you repro on a very big emoji, when it stops increasing, since I can't do that on it.

@kchibisov Sorry, I did not quite get what you mean. Getting a screenshot with a terminal zoomed at the maximum? That's roughly what I did + 4x with no interpolation in gimp.

Here is the same glyph with the same background color on kitty for reference. I think it looks better and actually blends with the bright color:

kitty

kchibisov:
Can you check this?

```diff
diff --git a/res/text.f.glsl b/res/text.f.glsl
index 2fcc2b9..78d8f4e 100644
--- a/res/text.f.glsl
+++ b/res/text.f.glsl
@@ -36,7 +36,9 @@ void main()
             // Color glyphs (such as emojis).
             vec4 glyphColor = texture(mask, TexCoords);
             alphaMask = vec4(glyphColor.a);
-            color = vec4(glyphColor.rgb, 1.0);
+            // Blend with background color used in a pass.
+            vec3 blendedColor = glyphColor.rgb + bg.rgb * (1.0 -  glyphColor.a);
+            color = vec4(blendedColor, 1.0);
         } else {
             // Regular text glyphs.
             vec3 textColor = texture(mask, TexCoords).rgb;

```

chrisduerr:
Looking at the previous and current image, it seems like both take the wrong approach.

It seems like the initial approach replaced the pixel instead of blending with it, leading to the background becoming visible.

@kchibisov's patch fixed that, but it now seems to blend with black instead of the cell's actual background color?

lbonn:
The result looks good with @kchibisov last proposal!

kchibisov:
@lbonn could you revert the last commit, it's wrong and works only by luck. I'll send a proper one later.

kchibisov:
Apply this on top of your PR.

```diff
diff --git a/font/src/darwin/mod.rs b/font/src/darwin/mod.rs
index 26e98e5..19371c9 100644
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -488,8 +488,8 @@ impl Font {
             kCGImageAlphaPremultipliedFirst | kCGBitmapByteOrder32Host,
         );

-        // Give the context an opaque, black background.
-        cg_context.set_rgb_fill_color(0.0, 0.0, 0.0, 1.0);
+        // Set background color for graphics context.
+        cg_context.set_rgb_fill_color(0.0, 0.0, 0.0, 0.0);
         let context_rect = CGRect::new(
             &CGPoint::new(0.0, 0.0),
             &CGSize::new(f64::from(rasterized_width), f64::from(rasterized_height)),
diff --git a/font/src/lib.rs b/font/src/lib.rs
index 2a19f62..59088e0 100644
--- a/font/src/lib.rs
+++ b/font/src/lib.rs
@@ -173,7 +173,10 @@ pub struct RasterizedGlyph {

 #[derive(Clone, Debug)]
 pub enum BitmapBuffer {
+    /// RGB pixels.
     RGB(Vec<u8>),
+
+    /// RGBA pixels with premultiplied alpha.
     RGBA(Vec<u8>),
 }

diff --git a/res/text.f.glsl b/res/text.f.glsl
index 78d8f4e..0f51b78 100644
--- a/res/text.f.glsl
+++ b/res/text.f.glsl
@@ -33,12 +33,17 @@ void main()
         color = vec4(bg.rgb, 1.0);
     } else {
         if (colored != 0) {
-            // Color glyphs (such as emojis).
+            // Color glyphs, like emojis.
             vec4 glyphColor = texture(mask, TexCoords);
             alphaMask = vec4(glyphColor.a);
-            // Blend with background color used in a pass.
-            vec3 blendedColor = glyphColor.rgb + bg.rgb * (1.0 -  glyphColor.a);
-            color = vec4(blendedColor, 1.0);
+
+            // Handle premultiplied alpha.
+            //
+            // NOTE - we won't show garbage from division by zero on a screen,
+            // due to alpha being 0, so it's perfrectly fine to do so.
+            glyphColor.rgb = vec3(glyphColor.rgb / glyphColor.a);
+
+            color = vec4(glyphColor.rgb, 1.0);
         } else {
             // Regular text glyphs.
             vec3 textColor = texture(mask, TexCoords).rgb;
diff --git a/res/text.v.glsl b/res/text.v.glsl
index 51a2ff6..8978c11 100644
--- a/res/text.v.glsl
+++ b/res/text.v.glsl
@@ -15,13 +15,13 @@
 // Cell properties.
 layout (location = 0) in vec2 gridCoords;

-// glyph properties.
+// Glyph properties.
 layout (location = 1) in vec4 glyph;

 // uv mapping.
 layout (location = 2) in vec4 uv;

-// text fg color.
+// Text fg color.
 layout (location = 3) in vec3 textColor;

 // Background color.

```

alacritty/src/renderer/mod.rs
```diff
@@ -418,6 +418,8 @@ struct InstanceData {
     bg_g: f32,
     bg_b: f32,
     bg_a: f32,
+    // Colored glyph indicator.
+    colored: u8,
```
chrisduerr, line 422:
```suggestion
    // Flag indicating that glyph uses multiple colors, like an Emoji.
    multicolor: u8,
```

alacritty/src/renderer/mod.rs
```diff
@@ -643,6 +639,17 @@ impl QuadRenderer {
             );
             gl::EnableVertexAttribArray(4);
             gl::VertexAttribDivisor(4, 1);
+            // coloredGlyph attribute
```
chrisduerr, line 642: This comment's text should match the variable name and be terminated with a period.

font/src/lib.rs
```diff
@@ -173,7 +173,10 @@ pub struct RasterizedGlyph {
 
 #[derive(Clone, Debug)]
 pub enum BitmapBuffer {
+    /// RGB pixels.
     RGB(Vec<u8>),
```
chrisduerr, line 177: If we comment it, we might as well say that this isn't really RGB pixels but an RGB alphamask?

Effectively, this is also premultiplied, just with on alpha.

res/text.f.glsl
```diff
@@ -31,8 +32,23 @@ void main()
...
+        if (colored != 0) {
+            // Color glyphs, like emojis.
+            vec4 glyphColor = texture(mask, TexCoords);
+            alphaMask = vec4(glyphColor.a);
+
+            // Handle premultiplied alpha.
+            //
+            // NOTE - we won't show garbage from division by zero on a screen,
+            // due to alpha being 0, so it's perfrectly fine to do so.
+            glyphColor.rgb = vec3(glyphColor.rgb / glyphColor.a);
```
chrisduerr, line 44:
```suggestion
            // Revert alpha premultiplication.
            if (glyphColor.a != 0) {
                glyphColor.rgb = vec3(glyphColor.rgb / glyphColor.a);
            }
```

alacritty/src/renderer/mod.rs
```diff
@@ -1513,14 +1520,18 @@ impl Atlas {
             gl::PixelStorei(gl::UNPACK_ALIGNMENT, 1);
             gl::GenTextures(1, &mut id);
             gl::BindTexture(gl::TEXTURE_2D, id);
+            // We use an RGBA atlas that can handle colored glyphs.
+            // Most of the text glyphs don't need this alpha channel but it has a negligible impact in practice.
+            // For more details and comparative benchmarks with a dual atlas approach, see:
+            // https://github.com/alacritty/alacritty/pull/3665
```
chrisduerr, line 1526:
```suggestion
            // Use RGBA texture for both normal and emoji glyphs, since it has no performance impact.
```

res/text.f.glsl
kchibisov: Don't you like fun stuff?

font/src/lib.rs
kchibisov: Maybe we should change it...

res/text.f.glsl
chrisduerr: No fun allowed in Rust unfortunately. Especially because so many platforms out there.

font/src/lib.rs
lbonn: I've changed it to `RGB alphamask.` in the last revision.

PR #3666, opened 2020-05-01
clippy fixes

matthiaskrgr:


alacritty/src/window.rs
```diff
@@ -345,7 +345,7 @@ impl Window {
     }
 
     pub fn set_fullscreen(&mut self, fullscreen: bool) {
-        #[cfg(macos)]
+        #[cfg(target_os = "macos")]
         {
             if self.window().simple_fullscreen() {
```
matthiaskrgr, line 350: this looks like it was `macos` by mistake since all `target_os =` is used otherwise

chrisduerr: Turns out this is already done by winit. Would you mind removing this entirely?

We should do the same for `set_simple_fullscreen`.

matthiaskrgr: You mean the attribute + conditional code, right?

matthiaskrgr: &winit::window::Window might not have a `set_simple_fullscreen` function outside of `macos`...?

I'll push a commit and to see if CI passes and squash later if that works..

chrisduerr: Exactly. The dependency already makes sure that it is a no-op to call this when the other fullscreen mode is already active. So we don't need to do it ourselves.

alacritty/src/window.rs
```diff
@@ -345,13 +345,6 @@ impl Window {
     }
 
     pub fn set_fullscreen(&mut self, fullscreen: bool) {
-        #[cfg(macos)]
-        {
-            if self.window().simple_fullscreen() {
-                return;
-            }
-        }
-
```
chrisduerr, line 354: The same should be done for `set_simple_fullscreen`.

matthiaskrgr: Oh I was looking for a `cfg(..)` inside set_simple_fullscreen :sweat_smile:  sorry....

Issue #2861, opened 2019-10-09
mouse back forward mapping


unphased:
I'm able to see when launching with --print-events that this event is seen: 

```
[2019-10-09 10:27] [INFO] glutin event: DeviceEvent { device_id: DeviceId(X(DeviceId(10))), event: Button { button: 9, state: Released } }
[2019-10-09 10:27] [INFO] glutin event: WindowEvent { window_id: WindowId(X(WindowId(98566149))), event: MouseInput { device_id: DeviceId(X(DeviceId(2))), state: Released, button: Other(9), modifiers: ModifiersState { shift: false, ctrl: false, alt: false, logo: false } } }
```

It looks like the back button is Other(9) and forward is Other(8).

How can I bind these to something so I can use it for e.g. my jumplist in vim?

unphased:
I've tried to set the binding in mouse_bindings, but get an error 
```
Problem with config: invalid type: integer `8`, expected Left, Right, Middle, or a number; using default value
```

unphased:
I figured it out, a string number `"8"` in the config did the trick.

chrisduerr:
I honestly didn't even know myself that this works. Thanks for figuring this out!

PR #3668, opened 2020-05-01
Allows u8 as mouse binding value

Dav1dde:
Fixes: #2861

I did not remove "5" for compatibility reasons

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Preserve selection on non-LMB or mouse mode clicks
 - Wayland client side decorations are now based on config colorscheme
 - Low resolution window decoration icon on Windows
+- Mouse bindings for additional buttons need to be specified as a number not a string
 
 ### Fixed
 
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -14,7 +14,6 @@
 #![allow(clippy::enum_glob_use)]
 
 use std::fmt::{self, Debug, Display};
-use std::str::FromStr;
 
 use glutin::event::VirtualKeyCode::*;
 use glutin::event::{ModifiersState, MouseButton, VirtualKeyCode};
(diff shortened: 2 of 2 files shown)


kchibisov:
>I did not remove "5" for compatibility reasons

Nah, I feel like we should remove this syntax, there's no need to keep bug around and there's no place for compatibility for such bugs before `1.0.0` in my opinion, since you'll throw a valid error.

Dav1dde:
It's gone

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Tabstops not being reset with `reset`
 - Selection not cleared when switching between main and alt grid
 - Fallback to `LC_CTYPE=UTF-8` on macOS without valid system locale
+- Numbers are now allowed as mouse binding value
```
chrisduerr, line 33: If this is a new feature, it should be part of `Added`. If it's a fix for an old issue it should be properly documented.

I think I'd move it to changed section and go into more detail that mouse bindings are accepted again, but only as numbers, not as strings.

alacritty/src/config/bindings.rs
```diff
@@ -621,7 +621,18 @@ impl<'a> Deserialize<'a> for MouseButtonWrapper {
             type Value = MouseButtonWrapper;
 
             fn expecting(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-                f.write_str("Left, Right, Middle, or a number")
+                f.write_str("Left, Right, Middle, or a number (0 <= x <= 255")
```
chrisduerr, line 624:
```suggestion
                f.write_str("Left, Right, Middle, or a number from 0 to 255")
```

alacritty/src/config/bindings.rs
```diff
@@ -621,7 +621,18 @@ impl<'a> Deserialize<'a> for MouseButtonWrapper {
...
+
+            fn visit_u64<E>(self, value: u64) -> Result<MouseButtonWrapper, E>
+            where
+                E: de::Error,
+            {
+                if let Ok(other) = u8::try_from(value) {
+                    Ok(MouseButtonWrapper(MouseButton::Other(other)))
+                } else {
+                    Err(E::invalid_value(Unexpected::Unsigned(value), &self))
+                }
```
chrisduerr, line 635: Should probably use a match here.

kchibisov:
@chrisduerr I guess it's already there https://github.com/alacritty/alacritty/blob/6b45780f3afa23e9e4277bf43e6bf381172fa1ea/alacritty.yml#L410

chrisduerr:
> @chrisduerr I guess it's already there 

I just looked at the diff and you're completely right. My bad.

PR #3689, opened 2020-05-06
feat: subpixel anti aliasing for macOS

casperstorm:
This is a draft PR. As agreed with @kchibisov, we will try to finish it in collaboration. The reason being is that I am still very new to the language (and the code base).

The goal of this PR is to fix:
```
# Thin stroke font rendering (macOS only)
  #
  # Thin strokes are suitable for retina displays, but for non-retina screens
  # it is recommended to set `use_thin_strokes` to `false`
  #
  # macOS >= 10.14.x:
  #
  # If the font quality on non-retina display looks bad then set
  # `use_thin_strokes` to `true` and enable font smoothing by running the
  # following command:
  #   `defaults write -g CGFontRenderingFontSmoothingDisabled -bool NO`
```
This will overwrite all apps on the system, which is not ideal. This PR aims to do it automatically, and only for Alacritty.


alacritty/Cargo.toml
```diff
@@ -33,6 +33,9 @@ rustc_tools_util = "0.2.0"
 [target.'cfg(not(windows))'.dependencies]
 xdg = "2"
 
+[target.'cfg(target_os = "macos")'.dependencies]
+cocoa = { git = "https://github.com/servo/core-foundation-rs", branch = "master" }
```
casperstorm, line 37: This obvious needs to change, once they tag and release the new version which has the patch i've commited.

alacritty/src/display.rs
```diff
@@ -198,6 +202,11 @@ impl Display {
             api.clear(background_color);
         });
 
+        // Enable subpixel anti-aliasing.
+        #[cfg(target_os = "macos")]
+        let use_thin_strokes = config.font.use_thin_strokes();
+        Self::enable_font_smoothing(use_thin_strokes);
```
casperstorm, line 208: The idea behind calling this function here is to run the code initially when Alacritty starts.

alacritty/src/display.rs
```diff
@@ -351,6 +372,11 @@ impl Display {
         let physical = PhysicalSize::new(self.size_info.width as u32, self.size_info.height as u32);
         self.window.resize(physical);
         self.renderer.resize(&self.size_info);
+
+        // Enable subpixel anti-aliasing.
+        #[cfg(target_os = "macos")]
+        let use_thin_strokes = config.font.use_thin_strokes();
+        Self::enable_font_smoothing(use_thin_strokes);
```
casperstorm, line 379: The idea behind calling this function here is that when we do changes to the config we want to make sure if it affects the `use_thin_strokes` we update that to the new value. Otherwise we would have to restart Alacritty twice in order to make it apply.

casperstorm:
One thing which I would like to do is add a flag to ensure we only do this on ` > 10.14.x` however I am not sure what the best way to properly get that information.
I was looking at `uname()` but that seems only to get the darwin kernel version, which I most certainly can use (it just needs to be [above 18.0.0](https://en.wikipedia.org/wiki/Darwin_(operating_system))) but, it would be more readable with the actually macOS version.

kchibisov:
>One thing which I would like to do is add a flag to ensure we only do this on > 10.14.x however I am not sure what the best way to properly get that information.
I was looking at uname() but that seems only to get the darwin kernel version, which I most certainly can use (it just needs to be above 18.0.0) but, it would be more readable with the actually macOS version.

You should use this to check versions
https://developer.apple.com/documentation/foundation/nsoperatingsystemversion

casperstorm:
> You should use this to check versions
> https://developer.apple.com/documentation/foundation/nsoperatingsystemversion

Nice, and it seems it's [implemented as well ](https://github.com/servo/core-foundation-rs/blob/16ec92e7c43da3c12f24495436bb528ab1b3bfb6/cocoa/src/foundation.rs#L206)

alacritty/src/display.rs
```diff
@@ -20,6 +20,10 @@ use std::fmt::{self, Formatter};
 use std::sync::atomic::Ordering;
 use std::time::Instant;
 
+#[cfg(target_os = "macos")]
+use cocoa::base::{id, nil, NO};
+use cocoa::foundation::{NSString, NSUserDefaults};
```
chrisduerr, line 25: Attributes on Rust only refer to the next line, so you need to add this attribute twice.

```suggestion
#[cfg(target_os = "macos")]
use cocoa::base::{id, nil, NO};
#[cfg(target_os = "macos")]
use cocoa::foundation::{NSString, NSUserDefaults};
```

alacritty/src/display.rs
```diff
@@ -329,6 +340,18 @@ impl Display {
...
+    fn enable_font_smoothing(enable: bool) {
+        unsafe {
+            let key = NSString::alloc(nil).init_str("CGFontRenderingFontSmoothingDisabled");
+            if enable {
+                id::standardUserDefaults().setBool_forKey_(NO, key);
+            } else {
+                id::standardUserDefaults().removeObject_forKey_(key);
+            }
+        }
+    }
```
chrisduerr, line 353: I'd like to move this part to the font crate, since this is where the font smoothing is configured and used.

alacritty/src/display.rs
```diff
@@ -329,6 +340,18 @@ impl Display {
         });
     }
 
+    /// Enable subpixel anti-aliasing
+    fn enable_font_smoothing(enable: bool) {
```
chrisduerr, line 344: The function name is incorrect, it does not enable font smoothing but instead sets the font smoothing value based on the parameter.

casperstorm: Correct, that's a horrible name. I'll fix that.

alacritty/src/display.rs
casperstorm: Into the `font/src/darwin/mod.rs` file, eg?

alacritty/src/display.rs
```diff
@@ -329,6 +343,24 @@ impl Display {
...
+            let min_macos_version = NSOperatingSystemVersion::new(10, 14, 0);
+            if NSProcessInfo::processInfo(nil).isOperatingSystemAtLeastVersion(min_macos_version) {
+                let key = NSString::alloc(nil).init_str("CGFontRenderingFontSmoothingDisabled");
+                if enable {
+                    id::standardUserDefaults().setBool_forKey_(NO, key);
+                } else {
+                    id::standardUserDefaults().removeObject_forKey_(key);
+                }
+            }
+        }
```
casperstorm, line 361: I've added a check now to see if we are running `10.14.0`.

alacritty/src/display.rs
```diff
@@ -329,6 +343,24 @@ impl Display {
         });
     }
 
+    /// Set subpixel anti-aliasing
+    fn set_font_smoothing(enable: bool) {
+        unsafe {
+            // MacOS Mojave 10.14.x disables subpixel anti-aliasing.
+            // To re-enable it, on an application level, we want to ensure
+            // we are running at least 10.14.0.
```
chrisduerr, line 351:
```suggestion
            // MacOS Mojave 10.14.x disables subpixel anti-aliasing,
            // to re-enable it on an application level, we want to ensure
            // we are running at least 10.14.0.
```

alacritty/src/display.rs
chrisduerr: Yeah. There are a few approaches we could take, either we check whenever a font is loaded or we provide a method as part of its public API.

alacritty/src/display.rs
```diff
@@ -31,7 +31,7 @@ use parking_lot::MutexGuard;
 #[cfg(not(any(target_os = "macos", windows)))]
 use wayland_client::{Display as WaylandDisplay, EventQueue};
 
-use font::{self, Rasterize};
+use font::{self, Rasterize, set_font_smoothing};
```
casperstorm, line 34: @chrisduerr I've moved the `fn` to the font crate now. How do we like this?

font/src/darwin/mod.rs
```diff
@@ -278,6 +281,25 @@ pub struct Font {
 
 unsafe impl Send for Font {}
 
+
+/// Set subpixel anti-aliasing on MacOS.
```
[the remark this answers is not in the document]
kchibisov: We should state that this option requires restart. We should also cleanup alacritty.yml near this setting.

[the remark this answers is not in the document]
chrisduerr: Oh right it does, so no point in calling it on config reload either.

alacritty/src/display.rs
```diff
@@ -386,6 +393,13 @@ impl Display {
         let physical = PhysicalSize::new(self.size_info.width as u32, self.size_info.height as u32);
         self.window.resize(physical);
         self.renderer.resize(&self.size_info);
+
+        // Set subpixel anti-aliasing.
+        #[cfg(target_os = "macos")]
+        {
+            let use_thin_strokes = config.font.use_thin_strokes();
+            set_font_smoothing(use_thin_strokes);
+        }
```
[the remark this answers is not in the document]
chrisduerr: As @kchibisov has stated, this setting requires a restart, so there's no point in trying to 'live reload' it.

We should probably still set it so that the user doesn't have to restart twice, but no need to try and refresh the font.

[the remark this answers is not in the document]
kchibisov: We still should handle reload of this setting, since we should prevent double restart to apply, so putting it in `reload_config` is a way to go anyway. However we should toggle it on start as well somewhere.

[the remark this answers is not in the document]
chrisduerr: Yeah exactly, just without reloading the font itself.

font/src/darwin/mod.rs
```diff
@@ -278,6 +281,25 @@ pub struct Font {
...
+pub fn set_font_smoothing(enable: bool) {
+    unsafe {
```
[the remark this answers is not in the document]
casperstorm: Yup, moving it out of the `unsafe` block works just fine.

font/src/darwin/mod.rs
```diff
@@ -278,6 +281,25 @@ pub struct Font {
...
+        // MacOS Mojave 10.14.x disables subpixel anti-aliasing,
+        // to re-enable it on an application level, we want to ensure
+        // we are running at least 10.14.0.
+        let min_macos_version = NSOperatingSystemVersion::new(10, 14, 0);
+        if NSProcessInfo::processInfo(nil).isOperatingSystemAtLeastVersion(min_macos_version) {
```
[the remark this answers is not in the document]
casperstorm: But in the case `enable` is false there is two scenarios: 

1. `CGFontRenderingFontSmoothingDisabled` is already set to `true` (default) because it has never been set to `false`.
2. `CGFontRenderingFontSmoothingDisabled` is set to `false` and in this case we want to remove our value (`false`) so it returns to `true`.

alacritty/src/display.rs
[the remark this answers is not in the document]
casperstorm: Just to sum up so we can agree (or disagree):
I'll call it on `reload_config` to ensure we don't get into a double-restart in order to enable it. I'll also call it initially to check if it was changed while Alacritty is closed.

Is the two places im calling it right now ok, or how do I move from here? I'll update the `alacritty.yml` file meanwhile as well.

font/src/darwin/mod.rs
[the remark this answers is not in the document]
chrisduerr: No, in case the version isn't correct, we never want to do anything. Why would we try to remove a value we've never set ourselves?

alacritty/src/display.rs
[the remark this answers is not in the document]
chrisduerr: They're not ok. As previously mentioned, `handle_update` is not always called when the config is changed. It should be called from `reload_config`.

alacritty/src/display.rs
chrisduerr: Seems much better, thanks.

font/src/darwin/mod.rs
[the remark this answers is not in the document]
chrisduerr: No, in case the version isn't correct, we never want to do anything. Why would we try to remove a value we've never set ourselves?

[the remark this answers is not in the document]
casperstorm: Ah, alrite, I misunderstood you. This has been fixed.

alacritty.yml
```diff
@@ -153,17 +153,7 @@
...
-  # macOS >= 10.14.x:
-  #
-  # If the font quality on non-retina display looks bad then set
-  # `use_thin_strokes` to `true` and enable font smoothing by running the
-  # following command:
-  #   `defaults write -g CGFontRenderingFontSmoothingDisabled -bool NO`
-  #
-  # This is a global setting and will require a log out or restart to take
-  # effect.
+  # it is recommended to set `use_thin_strokes` to `false`.
```
casperstorm, line 156: @chrisduerr I moved the section about overwriting `CGFontRenderingFontSmoothingDisabled`. We won't need that anymore.

font/src/darwin/mod.rs
```diff
@@ -278,6 +281,31 @@ pub struct Font {
...
+/// to re-enable it.
+///
+/// This is a no-op on systems running High Sierra or earlier (< 10.14.0).
+pub fn set_font_smoothing(enable: bool) {
+    let min_macos_version = NSOperatingSystemVersion::new(10, 14, 0);
+    unsafe {
+        // Check that we're running at least Mojave (10.14.0+).
+        if !NSProcessInfo::processInfo(nil).isOperatingSystemAtLeastVersion(min_macos_version) {
+            return
+        }
```
casperstorm, line 298: @chrisduerr whats your/our preferred way here? `!` or ` == false`?

chrisduerr: No point to ever do `== false` really.

PR #3723, opened 2020-05-14
updated alacritty.yml to add clarity to url launcher comments

tsherrard:

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -375,7 +375,7 @@
     # This program is executed when clicking on a text which is recognized as a URL.
     # The URL is always added to the command as the last parameter.
     #
-    # When launcher is set to None, URL launching will be disabled completely.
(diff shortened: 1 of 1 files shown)


alacritty.yml
```diff
@@ -375,7 +375,7 @@
     # This program is executed when clicking on a text which is recognized as a URL.
     # The URL is always added to the command as the last parameter.
     #
-    # When set to `None`, URL launching will be disabled completely.
+    # When launcher is set to None, URL launching will be disabled completely.
```
chrisduerr, line 378: I think it might be a bit better to state the option even more explicitly, to make sure **everyone** gets it:

```suggestion
    # When set to `launcher: None`, URL launching will be disabled completely.
```

tsherrard: Great idea!

PR #3735, opened 2020-05-17
Fix crash when writing wide char in last column

chrisduerr:
This resolves an issue where trying to write a fullwidth character in
the last column would crash Alacritty, if linewrapping was disabled.

Instead of assuming that the linewrap put after the linewrapping spacer
was successful, the character writing is now skipped completely when
trying to put a wide character in the last column.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix escapes prematurely terminated by terminators in unicode glyphs
 - Incorrect location when clicking inside an unfocused window on macOS
 - Startup mode `Maximized` on Windows
+- Crash when trying to write a fullwidth character in the last column
 
 ## 0.4.2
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix escapes prematurely terminated by terminators in unicode glyphs
 - Incorrect location when clicking inside an unfocused window on macOS
 - Startup mode `Maximized` on Windows
+- Crash when trying to write a fullwidth character in the last column
```
kchibisov, line 46: Should we state that it was with disabled line wrapping?

chrisduerr:
This can be reproduced with the following script:

```
#!/bin/bash

fish_len=7
cols=$(tput cols)

echo -ne "\e[?25l\e[?7l"

for c in $(seq $cols -1 $((fish_len * -1 + 2))); do
    if [[ $((c % 2)) == 0 ]]; then
        fish="<°)))彡 "
    else
        fish="<°)))ミ "
    fi

    chop_tail=$((cols - c + 1))
    if [[ $chop_tail -gt 0 ]]; then
        fish=${fish:0:$chop_tail}
    fi

    chop_head=$((c * -1 + 1))
    if [[ $chop_head -gt 0 ]]; then
        fish=${fish:$chop_head}
        c=0
    fi

    echo -ne "\e[${c}G$fish"

    sleep .05
done

echo -e "\e[?25h\e[?7h"
```

CHANGELOG.md
chrisduerr: Should be good now?

PR #3736, opened 2020-05-17
Perform glClear() on-screen on X11

### LINKED ISSUE #3061: Weird resize lag when launching Alacritty in tiling mode

arvl130:
Fixes #3061.

On X11, `glClear()` relies on dimensions from the WM to know what portions of the screen it should color on.

These dimensions may be different for when the window is shown, and when the window is hidden. But the former tends to be more reliable (this is especially the case on tiling WMs which remap windows after they are shown).

So we only call `glClear()` when the window is on-screen.

This PR makes Alacritty call `set_visible()` earlier during init on X11 (the same behavior prior to 0815774cbfb9d1421b8077d5d9d641faf5f818c1), because this seems to be the only way I can get the window to init at the correct dimensions.

Other than that, it also makes Alacritty call `glClear()` earlier on X11 (right after `set_visible()`). This is to avoid having the window from being shown with an uninitialized surface (what 0815774cbfb9d1421b8077d5d9d641faf5f818c1 fixes) during init.

### DIFF
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -219,25 +219,27 @@ impl Display {
         // Update OpenGL projection.
         renderer.resize(&size_info);
 
-        // Clear screen.
-        let background_color = config.colors.primary.background;
-        renderer.with_api(&config, &size_info, |api| {
-            api.clear(background_color);
-        });
-
         #[cfg(not(any(target_os = "macos", windows)))]
         let is_x11 = event_loop.is_x11();
 
         #[cfg(not(any(target_os = "macos", windows)))]
         {
-            // On Wayland we can safely ignore this call, since the window isn't visible until you
-            // actually draw something into it and commit those changes.
-            if is_x11 {
-                window.swap_buffers();
+            // Clear screen on Wayland.
+            if !is_x11 {
+                let background_color = config.colors.primary.background;
                 renderer.with_api(&config, &size_info, |api| {
-                    api.finish();
+                    api.clear(background_color);
                 });
             }
+        }
+
+        #[cfg(any(target_os = "macos", windows))]
+        {
+            // Clear screen on Windows and MacOS.
+            let background_color = config.colors.primary.background;
+            renderer.with_api(&config, &size_info, |api| {
+                api.clear(background_color);
+            });
         }
 
         window.set_visible(true);
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -178,6 +178,18 @@ impl Display {
         // Create renderer.
         let mut renderer = QuadRenderer::new()?;
 
+        #[cfg(not(any(target_os = "macos", windows)))]
+        let is_x11 = event_loop.is_x11();
+
+        #[cfg(not(any(target_os = "macos", windows)))]
+        {
+            if is_x11 {
+                // On X11, show the window early so glClear()
+                // can color over it at the right size.
+                window.set_visible(true);
+            }
+        }
+
         let (glyph_cache, cell_width, cell_height) =
             Self::new_glyph_cache(dpr, &mut renderer, config)?;
 
@@ -219,27 +231,22 @@ impl Display {
         // Update OpenGL projection.
         renderer.resize(&size_info);
 
-        #[cfg(not(any(target_os = "macos", windows)))]
-        let is_x11 = event_loop.is_x11();
+        // Clear screen.
+        let background_color = config.colors.primary.background;
+        renderer.with_api(&config, &size_info, |api| {
+            api.clear(background_color);
+        });
 
         #[cfg(not(any(target_os = "macos", windows)))]
         {
-            // Clear screen on Wayland.
-            if !is_x11 {
-                let background_color = config.colors.primary.background;
+            // On Wayland we can safely ignore this call, since the window isn't visible until you
+            // actually draw something into it and commit those changes.
+            if is_x11 {
+                window.swap_buffers();
                 renderer.with_api(&config, &size_info, |api| {
-                    api.clear(background_color);
+                    api.finish();
                 });
             }
-        }
-
-        #[cfg(any(target_os = "macos", windows))]
-        {
-            // Clear screen on Windows and MacOS.
-            let background_color = config.colors.primary.background;
-            renderer.with_api(&config, &size_info, |api| {
-                api.clear(background_color);
-            });
         }
 
         window.set_visible(true);
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -203,14 +203,14 @@ impl Window {
                 if let Some(parent_window_id) = config.window.embed {
                     x_embed_window(windowed_context.window(), parent_window_id);
                 }
-                // Perform glClear() on-screen to avoid drawing color over on the window
-                // at incorrect sizes.
+                // Perform glClear() on-screen to avoid drawing color over the window
+                // at an incorrect size.
+                // We make GL calls here directly, instead of going through QuadRenderer,
+                // to avoid having to initialize SizeInfo.
                 windowed_context.window().set_visible(true);
                 gl_clear(config.colors.primary.background, config.background_opacity());
-                // We make these GL calls here directly, instead of going through QuadRenderer,
-                // to avoid having to initialize SizeInfo. Also, swap buffers to avoid glitches.
+                // Swap buffers to commit glClear() color into the screen.
                 windowed_context.swap_buffers().expect("swap buffers");
-                gl_finish();
             } else {
                 // Apply client side decorations theme.
                 let theme = AlacrittyWaylandTheme::new(&config.colors);
@@ -493,14 +493,6 @@ unsafe extern "C" fn xembed_error_handler(_: *mut XDisplay, _: *mut XErrorEvent)
 
 #[inline]
 #[cfg(not(any(target_os = "macos", windows)))]
-fn gl_finish() {
-    unsafe {
-        gl::Finish();
-    }
-}
-
-#[inline]
-#[cfg(not(any(target_os = "macos", windows)))]
 fn gl_clear(color: Rgb, alpha: f32) {
     unsafe {
         gl::ClearColor(
@@ -510,5 +502,7 @@ fn gl_clear(color: Rgb, alpha: f32) {
             alpha,
         );
         gl::Clear(gl::COLOR_BUFFER_BIT);
+        // Block until glClear() finishes.
+        gl::Finish();
     }
 }
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -47,7 +47,6 @@ use x11_dl::xlib::{Display as XDisplay, PropModeReplace, XErrorEvent, Xlib};
 use alacritty_terminal::config::Colors;
 use alacritty_terminal::config::{Decorations, StartupMode, WindowConfig};
 use alacritty_terminal::event::Event;
-use alacritty_terminal::term::color::Rgb;
 #[cfg(not(windows))]
 use alacritty_terminal::term::{SizeInfo, Term};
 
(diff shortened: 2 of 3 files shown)


alacritty/src/window.rs
```diff
@@ -510,5 +502,7 @@ fn gl_clear(color: Rgb, alpha: f32) {
             alpha,
         );
         gl::Clear(gl::COLOR_BUFFER_BIT);
+        // Block until glClear() finishes.
+        gl::Finish();
```
kchibisov, line 506: Move that back to its original place.

arvl130: Where exactly?

kchibisov: Just remove this call.

arvl130: Okay, then.

alacritty/src/window.rs
```diff
@@ -481,3 +490,17 @@ unsafe extern "C" fn xembed_error_handler(_: *mut XDisplay, _: *mut XErrorEvent)
...
+    unsafe {
+        gl::ClearColor(
+            (f32::from(color.r) / 255.0).min(1.0) * alpha,
+            (f32::from(color.g) / 255.0).min(1.0) * alpha,
+            (f32::from(color.b) / 255.0).min(1.0) * alpha,
+            alpha,
+        );
+        gl::Clear(gl::COLOR_BUFFER_BIT);
+    }
+}
```
chrisduerr, line 506: We should never just copy-paste functions like this. If the old API doesn't work, it needs to be changed, not duplicated.

chrisduerr:
Oh, another thing. I'm pretty sure on some X11 systems if you run without compositor and call clear *after* the window is shown, you'll see the uninitialized surface for a short time.

So I'm relatively certain that's why we need to clear *before* showing the window.

alacritty/src/display.rs
```diff
@@ -219,27 +219,29 @@ impl Display {
...
         let is_x11 = event_loop.is_x11();
 
         #[cfg(not(any(target_os = "macos", windows)))]
         {
-            // On Wayland we can safely ignore this call, since the window isn't visible until you
-            // actually draw something into it and commit those changes.
-            if is_x11 {
-                window.swap_buffers();
+            // Clear screen on Wayland.
+            if !is_x11 {
```
kchibisov, line 228: Why, it was changed that way?

alacritty/src/display.rs
```diff
@@ -219,27 +219,29 @@ impl Display {
...
         }
 
+        #[cfg(any(target_os = "macos", windows))]
+        {
+            // Clear screen on Windows and MacOS.
+            let background_color = config.colors.primary.background;
+            renderer.with_api(&config, &size_info, |api| {
+                api.clear(background_color);
+            });
+        }
```
kchibisov, line 243: Why that?

alacritty/src/window.rs
```diff
@@ -202,6 +203,14 @@ impl Window {
...
                     x_embed_window(windowed_context.window(), parent_window_id);
                 }
+                // Perform glClear() on-screen to avoid drawing color over the window
+                // at an incorrect size.
+                // We make GL calls here directly, instead of going through QuadRenderer,
+                // to avoid having to initialize SizeInfo.
+                windowed_context.window().set_visible(true);
+                gl_clear(config.colors.primary.background, config.background_opacity());
+                // Swap buffers to commit glClear() color into the screen.
+                windowed_context.swap_buffers().expect("swap buffers");
```
kchibisov, line 213: You can safely do that later in display, I don't see any point in doing that during window creation?

alacritty/src/window.rs
kchibisov: I agree, you can do that with old API just fine.

arvl130:
Yeah, I did a lot of weird things in there in my initial commit (and the few commits after). Perhaps, I should have marked it as a draft, as this was really just me trying to just get stuff somewhat "fixed" or "working" on my end. ( :

My intention was to get ideas on how I could do things better, because I do realize, just copy and pasting functions from the API, and moving things around to weird places, is definitely a big no-no. (and I'm also not really much of a programmer, so I guess that gives)

For now, I've reverted most of the changes I did and am just adding a section that shows the window earlier so that Alacritty can "Perform glClear() on-screen on X11".

My reasoning for doing this, as was said in the PR description, is to allow glClear() to color over the portions of the screen that the WM actually wants it to.

Because on some WMs, specifically tiling ones, the dimensions it will color over will be different for when its hidden, and when it is shown.

I realize that what @chrisduerr says is true. Doing glClear() on-screen causes it to there be a short period of time that the surface is uninitialized. My testing though, shows that this only becomes noticeable when we do api.clear() *way* late after the window has been shown.

I haven't tested yet up to what point is "way late". But as in the the latest commit. It seems to be any point wherein set_visible() and api.clear() do not follow each other.

I have added set_visible() above new_glyph_cache() for now, but the better place would have probably been just above api.clear() (so that they follow each other).

I have to do this because for some reason, because glClear() no longer picks up the dimensions from the WM if I have the set_visible() anywhere after new_glyph_cache().

So for now, they have to be separate.

I still have no idea why this is, so I was hopping some of you might have ideas on this.

chrisduerr:
It seems like swapping buffers right after showing the window is likely just a question of timing for people to be able to reproduce the issues fixed by @kchibisov in https://github.com/alacritty/alacritty/pull/2580.

arvl130:
I'd be interested in seeing how noticeable the uninitialized "transparent" surface would still be for people if they test 9bcb66291419e1f8025de37fb41d7a109d786bde. The reason for my particularly messy code up to that commit was trying to get `set_visible()`, `clear()`, and `swap_buffers()` to run right after each other.

On my systems, having them run like that essentially makes the uninitialized surface non-existent. Since we show the window (`set_visible`), render the color (`clear`), then commit (`swap_buffers`) to the screen, as fast as possible.

kchibisov:
>On my systems, having them run like that essentially makes the uninitialized surface non-existent. Since we show the window (set_visible), render the color (clear), then commit (swap_buffers) to the screen, as fast as possible.

That could work, but can you record `alacritty --print-events` and provide slowdowned screencast of you opening alacritty?

arvl130:
Sorry for the delays! [Here's](https://i.imgur.com/SA6ror5.mp4) the startup screencast for 9bcb66291419e1f8025de37fb41d7a109d786bde, and [here's](https://gist.github.com/arvl130/145d194c57f77676c77ee417225c859f) is the print events log.

At this point though, I don't think I will be going forward with this PR anymore.

Because as [you'll see](https://i.imgur.com/2CXTsIY.jpg) in the screencast, having `set_visible()`, `clear()`, and `swap_buffers` run consecutively only fixes the issue when launching Alacritty on an empty workspace somehow. (I didn't notice this sooner since I always launch Alacritty in its own workspace)

When there's a window already present in the workspace you're launching. The issue still pops up here and there. Making the fix *kinda* useless.

That said, I am looking into a different solution now. Which doesn't require having to show our window early in the `Display` creation.

The gist of it is, since we know `glClear()` seems to like using the window size we give it when initializing the `Window` class for sizing it's rectangle. We always initialize the `Window` class with a `size` that we're sure will always fill up the entire window.

The size I've settled with in my personal build is the `size()` for largest monitor among the monitors in `event_loop.available_monitors()`.

If the dimensions estimated by `calculate_dimensions` are less then the dimensions in our minimum dimensions. We use our minimum. Otherwise, we create it at whatever the size that's configured.

This accounts for people who want to create windows at a size larger than any of the active monitors. Since we don't want to shrink the initialization `size` for `Window` in these cases. We only want to resize when we think the final window could possibly be larger than our estimated size.

If the user ever decides to create a Alacritty at such a size that's bigger than our minimum size (either through columns, lines or paddings). We assume the window will be floating anyway, so #3061 will be a non-issue. If it's going to be tiled. Tiling WMs will always tile a window in such a size that's smaller or equal to the largest monitor. So once again, this becomes a non-issue.

The only edge-case where #3061 will still appear. Is when you specifically tell your WM to float your Alacritty window, AND launch it at a size larger than your largest monitor AND have Alacritty configured in such a way that's less than your

The bottomline is. `glClear()` *really* isn't too friendly with WMs sizing your windows arbitrarily. Since when we run this call, it believes the current window size (specifically the size its context it was spawned in) will always be the size it will be given.

But then later, when we actually hand it to the WM to show. This isn't always going to be the case. So we weird get issues like [this](https://i.imgur.com/2CXTsIY.jpg).

For now, I'm sticking with largest available_monitor as a minimum size for initializing `Window`. If you have better ideas for a minimum size than this though, I'd be interested in hearing them. ( :

One question I have though is: should I ever want to submit these changes upstream again, are there other places I could show it for preliminary checking, or is this already the best place to have it reviewed ?

kchibisov:
Have you tried showing the window closer to event loop creation? Like right before `run_return` call, maybe that'll work better?

chrisduerr:
> We always initialize the Window class with a size that we're sure will always fill up the entire window.

This sounds like a really bad idea that's just bound to cause more issues than it solves.

> One question I have though is: should I ever want to submit these changes upstream again, are there other places I could show it for preliminary checking, or is this already the best place to have it reviewed?

Github's the correct place for getting reviews.

arvl130:
@kchibisov Nope, that still runs `glClear()` off-screen with the hidden window size. So the issue still shows up for me when testing this.

I tried doing a switcheroo on your suggestion though, and instead tried moving `api.clear()`, directly before `run_return`. That seems to work a lot better for me.

arvl130:
@kchibisov I've force-pushed the new changes (e4b2bb1f717aeaee5c27d93d06978f8854b0ce2d) based on your suggestion, now this one actually works.

I've done my best to comment every change I've made.  Could you please give it another look?

The only things that should stand out, is that `renderer` and `is_x11` is now public (this allows us to call `api.clear()` from Processor, right before we call `run_return`).

I've also added multiple conditionals. So we can keep the previous behavior on unaffected systems (Unix with Wayland, MacOS, and Windows). This was also something I did in my earlier commits, but I failed to explain why. I hope it's more clear now why I added these changes.

Here's the [patch](https://gist.github.com/arvl130/d36b5e930c0bfc152df44785d81e0adb), the [logs](https://gist.github.com/arvl130/3c7fd8eaf5c7bd3d6ce6cb05f13bb716), and a video [before](https://i.imgur.com/BzW5ABW.mp4) and [after](https://i.imgur.com/ztWVpQG.mp4) this change is made based on master.

If there are still things I've overlooked, please tell me.

PR #3747, opened 2020-05-18
Make cursor thicknesses individually configurable

LawAbidingCactus:
This makes each of the cursor types' thicknesses individually configurable; a thicker beam is nice for cursor visibility in vim, but the same isn't true for the underline (a thicker underline looks a bit strange in other programs). Three configuration variables are exposed: `cursor.thickness.{beam,underline,hollow}`. This PR includes miscellaneous formatting changes made by `cargo fmt`, apparently needed for CI to pass (I can revert the ones not affecting files relevant to these changes, if that's needed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Default Command+N keybinding for SpawnNewInstance on macOS
 - Vi mode for copying text and opening links
 - `CopySelection` action which copies into selection buffer on Linux/BSD
-- Option `cursor.thickness` to set terminal cursor thickness
+- Options `cursor.thickness.{beam, underline, hollow}` to set terminal cursor thickness
 - Font fallback on Windows
 - Support for Fontconfig embolden and matrix options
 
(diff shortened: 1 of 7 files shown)


kchibisov:
This was rejected when I've brought this option and I'm not sure we should justify that complexity.

Should we reiterate that @chrisduerr ?

LawAbidingCactus:
@kchibisov Oh, hmm, I just found that issue. I've found other people wanting something like this (and I've generally been able to configure this with other minimal terminal emulators, like `st`), so I think the complexity can be justified somewhat. Most of the LoC added seem to be some form of functionality duplication, anyway.

chrisduerr:
The biggest problem is absolutely unnecessary configuration bloat.

I was already skeptical of the original implementation, because I think the default thickness should just be using the appropriate heuristics to make the cursors easily visible. However the lack of a clear ideal approach with different systems using both high and low DPI monitors made it clear that it's probably easiest to allow altering the cursor width independently from the font.

Those options should already sufficiently allow for making sure everyone can configure cursors that are clearly visible, which should be enough. So I don't see a reason to allow even more fine grained changes. The preference against the default cursor widths is already quite fringe, desiring completely different widths for every individual one is just completely unnecessary complexity for almost every user, for the benefit of very few who could likely easily get used to the underline cursor being a little too thick for their preference.

PR #3753, opened 2020-05-20
Update copypasta to v0.7.0

### LINKED ISSUE #3592: [wayland] Clipboard amnesia with alacritty and geany
Fedora 31 AND Void (musl) [different hardware]
0.4.2
wayland gnome and wayland sway

The problem manifests solely with the geany editor for me but perhaps there are other applications.

To reproduce, have an alacritty terminal open, open the geany editor.
Select some text in geany and Copy, verify that Paste in geany works as intended. It should.
Attempt Paste in alacritty terminal. Nothing is produced.
Attempt Paste in any other application. Should work as intended.

Repeat the attempt in firefox (or some other application), Paste works as intended in alacritty.

Now attempt Copy in alacritty after selecting some text. Paste works in alacritty and firefox.
Attempting Paste in geany produces nothing.

geany seems to be using xwayland but the problem is absent for other xwayland applications.

**In Fedora 31, When using gnome-terminal instead of alacritty, the issue is entirely absent.**

**In Void sway, When using sakura instead of alacritty, the issue is entirely absent.**

On Fedora the alacritty config is entirely default.
On void the alacritty config has been taken from someone's dotfiles and is customized.

kchibisov:
Fixes #3592.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Incorrect location when clicking inside an unfocused window on macOS
 - Startup mode `Maximized` on Windows
 - Crash when writing a fullwidth character in the last column with auto-wrap mode disabled
+- Paste from some apps on Wayland
 
 ## 0.4.2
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Incorrect location when clicking inside an unfocused window on macOS
 - Startup mode `Maximized` on Windows
 - Crash when writing a fullwidth character in the last column with auto-wrap mode disabled
+- Paste from some apps on Wayland
```
kchibisov, line 47: The problem was with `UTF8_STRING` mime type, not sure that it should be mentioned in that way...

Issue #3737, opened 2020-05-17
Char ₿ not being rendered


ademar111190:
The char ₿ [("\u20BF", "\xE2\x82\xBF")](https://en.wikipedia.org/wiki/Currency_Symbols_(Unicode_block)) is not being rendered on alacritty, but it is being rendered on iterm and terminal.

I'm using the same font, [Meslo LG S for Powerline](https://github.com/powerline/fonts/tree/master/Meslo%20Dotted) for all 3 terminal emulators.

The following table has a print of each app, I'm adding another unusual char and an emoji both correctly displayed to show that the issue does not seem to be related to the font.

|alacritty|iterm|terminal|
|---|---|---|
|![alacritty](https://user-images.githubusercontent.com/1225438/82137370-ad5fc700-97ed-11ea-9d56-c81c2ab995b4.png)|![iterm](https://user-images.githubusercontent.com/1225438/82137302-31fe1580-97ed-11ea-8a90-8328758f079b.png)|![terminal](https://user-images.githubusercontent.com/1225438/82137301-31657f00-97ed-11ea-9011-b4ebeca54bc1.png)|

It is my first time using this app, so I'm not sure if I'm doing some mistake, so I'm sharing my config font section:

```
font:
    normal:
        family: Meslo LG S for Powerline
        style: Regular
    bold:
        family: Meslo LG S for Powerline
        style: Bold
    italic:
        family: Meslo LG S for Powerline
        style: Italic
    bold_italic:
        family: Meslo LG S for Powerline
        style: Bold Italic
    size: 12
    use_thin_strokes: true
```

### System

OS: macOS
Version: alacritty 0.4.2 (f68de37)

chrisduerr:
Neither terminal nor iterm seem to be showing the same glyph, so it might just be some internal font they use. Do you have the same font configured in all applications when comparing them? It looks like iTerm at least has a different weight configured.

ademar111190:
It is a good point @chrisduerr maybe those terminals use some internal font as a fallback.

I did change the fonts to Meslo but I never seen any other option regarding to fonts, I'm attaching my config.


iterm:
![iterm](https://user-images.githubusercontent.com/1225438/82156767-889e3a80-9853-11ea-9b86-d2f374bccec2.png)

terminal
![terminal](https://user-images.githubusercontent.com/1225438/82156769-8a67fe00-9853-11ea-987e-c20d56566e51.png)

chrisduerr:
[...]

PR #3761, opened 2020-05-23
Fixes the missing ₿ character

ademar111190:
I did face a problem on Alacritty on macOS, it was not showing the `₿` character, all the issue evolution can be found here: https://github.com/alacritty/alacritty/issues/3737

With help on that issue, I manage to find a solution that is working fine for me, the solution is to change the fallback fonts to use the fonts already installed on the machine.

So I'm opening this MR in a hope that those changes can be benefic for the project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Startup mode `Maximized` on Windows
 - Crash when writing a fullwidth character in the last column with auto-wrap mode disabled
 - Paste from some apps on Wayland
+- Showing the `₿` character on macOS through `Rockwell` font
 
 ## 0.4.2
 
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -389,6 +389,12 @@ impl Descriptor {
                         fallbacks.push(descriptor.to_font(size, false))
                     };
 
+                    if let Some(descriptor) =
+                        descriptors_for_family("Rockwell").into_iter().next()
+                    {
+                        fallbacks.push(descriptor.to_font(size, false))
+                    };
+
                     // Include Menlo in the fallback list as well.
                     fallbacks.insert(0, Font {
                         cg_font: menlo.copy_to_CGFont(),


ademar111190:
> Thanks for looking into this issue.

It was a pleasure, what you see is the very first Rust code I ever wrote, it was really enjoyable :)

>This looks a bit like replacing one hack that works somewhat okay with something a lot worse. As far as I can tell, your code lacks some structure, what is the thought process behind this implementation? It seems mostly like a collection of some heuristics thrown together to solve your specific issue.

Makes sense, I'm not familiar with the project internals, and Rust language too, so maybe I'm doing something dumb, hopefully answering the following question may help it.

> Why did you use descriptors_for_family rather than cascade_list_for_languages for font fallback?

I was trying to get all installed families fonts, independently of language, because I thought that could help with some characters of foreign languages, so if a weird text, let's say `𐤓‎‎𐤀‎𐤌‎𐤄‎‎𐤃‎𐤀` appears, maybe having the fonts filtered by a specific language could remove the possibility to render it, does it makes sense?

>Why are you trying to manually override the priority of regular fonts?
> Why do you get a list of all fonts and put Menlo at the top, rather then using the font fallback for the specified font or Menlo?

It is an interesting question, the previous code was using Menlo-Regular and Apple Symbols as fallback being Menlo inserted at the first position of the list. So I thought that Menlo was chosen for some reason and should have preference over the other fonts to keep the previous behaviour and Regular should keep preference over the variety for the same reason.

By writing this answer to the question I guess that I understood what you say about structure, I can pick Menlo and add the other fonts as a fallback for it, "a fallback list for a fallback font", so there is no sort… I haven't thought about it before.

I would be happy by changing that, so let me ask, should I close this PR and open another with the changes or should I keep this PR open meanwhile I do the changes, or should I close anyway because even doing the changes it is not interesting to the project?

chrisduerr:
> I was trying to get all installed families fonts, independently of language, because I thought that could help with some characters of foreign languages, so if a weird text, let's say 𐤓‎‎𐤀‎𐤌‎𐤄‎‎𐤃‎𐤀 appears, maybe having the fonts filtered by a specific language could remove the possibility to render it, does it makes sense?

The problem is that font fallback usually isn't performed on a global fallback list. The fallbacks are performed based on the first specified font and every font has a different fallback list. There is no one global font fallback list.

> I would be happy by changing that, so let me ask, should I close this PR and open another with the changes or should I keep this PR open meanwhile I do the changes, or should I close anyway because even doing the changes it is not interesting to the project?

If you are interested in making changes, doing so in this PR would be fine. But I'm not familiar with macOS font fallback myself, so I can't give you any great advice on how to do that without basically implementing it myself.

I'd recommend looking at other applications to check how they perform font fallback and using that to improve the existing Alacritty implementation. I'd start with iTerm for example, since its open source and is likely to follow macOS best practices. But that's a bit of work of course, so you'd have to be willing to do that research to get this done.

ademar111190:
Thanks for the time explaining, I misunderstood how the font works.

There is one thing interesting about the behaviour of `cascade_list_for_languages`, the doc says that it returns a list that matches the original font's style, weight, and width, but somehow this method is not matching `Rockwell` when passing `Menlo` as an argument but if you revert the arguments and pass `Rockwell` as an argument the `cascade_list_for_languages` returns `Menlo`…

`Rockwell` is a system font that contains the `₿` glyph, so manually adding `Rockwell` in the same way `Apple Symbols` was added fixes the issue.

chrisduerr:
> Rockwell is a system font that contains the ₿ glyph, so manually adding Rockwell in the same way Apple Symbols was added fixes the issue.

That makes sense, of course manually adding it will make it work. Though that's not really a solution. I didn't write the original macOS code but I would already treat the `Menlo` and emoji fonts that are manually set as hacks on macOS. Intuitively I'd assume that we should get the fallback for the specified font, not `Menlo` and then we shouldn't add any fallbacks manually. But I'd assume that provides even worse results.

To figure out how to achieve *better* results, I'd have to look at how other applications are doing this. Though testing out how fallback behaves if you use your actual font instead of `Menlo` might be worth a shot, but it obviously won't work for everyone since some might have `Menlo` as primary font.

PR #3771, opened 2020-05-26
Prevent SHELL clobbering on not macos.

Mattbazooka:


alacritty_terminal/src/tty/unix.rs
```diff
@@ -174,7 +178,11 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, window_id: Option<usize>) ->
     // Setup shell environment.
     builder.env("LOGNAME", pw.name);
     builder.env("USER", pw.name);
-    builder.env("SHELL", pw.shell);
+    if cfg!(target_os = "macos") {
```
chrisduerr, line 181: This shouldn't be a runtime cfg.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -174,7 +178,11 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, window_id: Option<usize>) ->
...
+        builder.env("SHELL", pw.shell);
+    } else {
+        builder.env("SHELL", &*shell.program);
+    }
```
chrisduerr, line 185: I think putting this block at the end of the shell env setup with a newline separating it and a comment on why this is done might be a little cleaner.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -154,7 +155,10 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, window_id: Option<usize>) ->
 
         Shell::new_with_args("/bin/bash", argv)
     } else {
-        Shell::new(pw.shell)
+        match var("SHELL"){
```
chrisduerr, line 158: The `$SHELL` variable is a thing on macOS too, however Terminal.app isn't interested in it.

Should we not support this on macOS? Is there any official resource telling us we shouldn't? Otherwise we could just support this for all unix platforms.

Mattbazooka: I can't find any resources on the matter, and don't have access too macOS for testing as we may need to implement the "ugly hack" from #1856.

chrisduerr: As you can see in that, we still use `pw.shell` with the macos workaround:

https://github.com/alacritty/alacritty/pull/1856/files#diff-0655ced7de6892bafa1f0000619c718eR228

So we'd just need to define a `shell_program` variable or something similar before this if/else and then use that instead of the `pw.shell`.

PR #3776, opened 2020-05-27
Perform glClear() on-screen on X11

arvl130:
This is my attempt #⁠2 at getting a fix for #3061 merged into upstream. 😅

The way we fix it is still the same. So I've chosen to reuse the title. But there's a few differences on how we do things now.

I don't think it's worth your time me explaining what I changed since then. The only things that should stand out now is that:

1) We now call `api.clear()` much later in the process (only in X11).

The reason for doing this is still as the title implies. Though, the decision to do it this way was based on a suggestion by @kchibisov in the previous thread. It was suggested that calling things nearer to `run_return` might help make the previous fix work better. And in fact, it actually does, with some tweaking. So I'm submitting it here now for further review.

2) We now do `api.clear()` behind ifs and conditionals.

The reason for doing this is to retain our previous behavior in other systems (MacOS, Windows, Unix with Wayland) where this bug doesn't exist. I'm not sure if there are better ways to do this yet. If so, please tell me.

[This was something I actually also did in my previous PR, but I failed to explain why. Hope that clears it up now why there are new ifs and conditionals everywhere]

I've done my best to comment on every change I made, but there's probably a few things I still left here and there. If so, kindly please tell me how I could improve on that. Thank you for your time.

\[[master](https://i.imgur.com/dspmolA.mp4)\]
\[[patched](https://i.imgur.com/lBQZkh8.mp4)\]

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -12,6 +12,7 @@ use std::path::PathBuf;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::time::Instant;
+use std::{thread, time};
 
 use glutin::dpi::PhysicalSize;
 use glutin::event::{ElementState, Event as GlutinEvent, ModifiersState, WindowEvent};
@@ -419,19 +420,18 @@ impl<N: Notify + OnResize> Processor<N> {
     where
         T: EventListener,
     {
+        // Show the window.
+        self.display.window.set_visible(true);
 
-        // Clear on-screen on Unix (X11).
+        // Sleep for 5ms.
         //
-        // We only perform clearing here to avoid
-        // clearing at incorrect window dimensions.
-        // NOTE: We put this line behind a conditional and
-        // an `if` to avoid clearing twice on other systems.
-        #[cfg(not(any(target_os = "macos", windows)))]
-        {
-            if event_loop.is_x11() {
-                self.display.clear_x11(&self.config);
-            }
-        }
+        // `set_visible` does not block until the window is shown, so we
+        // sleep here to avoid clearing while the window is still hidden.
+        thread::sleep(time::Duration::from_millis(5));
+
+        // Clear screen.
+        debug!("Inner size before clearing: {:?}", self.display.window.inner_size());
+        self.display.clear(&self.config);
 
         event_loop.run_return(|event, event_loop, control_flow| {
             if self.config.debug.print_events {
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -419,6 +419,20 @@ impl<N: Notify + OnResize> Processor<N> {
     where
         T: EventListener,
     {
+
+        // Clear on-screen on Unix (X11).
+        //
+        // We only perform clearing here to avoid
+        // clearing at incorrect window dimensions.
+        // NOTE: We put this line behind a conditional and
+        // an `if` to avoid clearing twice on other systems.
+        #[cfg(not(any(target_os = "macos", windows)))]
+        {
+            if event_loop.is_x11() {
+                self.display.clear_x11(&self.config);
+            }
+        }
+
         event_loop.run_return(|event, event_loop, control_flow| {
             if self.config.debug.print_events {
                 info!("glutin event: {:?}", event);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Startup mode `Maximized` on Windows
 - Crash when writing a fullwidth character in the last column with auto-wrap mode disabled
 - Paste from some apps on Wayland
+- Incorrect window dimensions when clearing on X11
 
 ## 0.4.2
 
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -221,28 +221,32 @@ impl Display {
         // Update OpenGL projection.
         renderer.resize(&size_info);
 
-        // Clear screen.
-        let background_color = config.colors.primary.background;
-        renderer.with_api(&config, &size_info, |api| {
-            api.clear(background_color);
-        });
-
         // Set subpixel anti-aliasing.
         #[cfg(target_os = "macos")]
         set_font_smoothing(config.font.use_thin_strokes());
 
+        // Clear off-screen on Windows and MacOS.
+        //
+        // These systems do not suffer from issue #3061,
+        // so we can safely clear here as normal.
+        // NOTE: We put this line behind a conditional
+        // to avoid clearing twice on other systems.
+        #[cfg(any(target_os = "macos", windows))]
+        Self::clear_screen(config, &mut renderer, &size_info);
+
         #[cfg(not(any(target_os = "macos", windows)))]
         let is_x11 = event_loop.is_x11();
 
+        // Clear off-screen on Unix (non-X11).
+        //
+        // Wayland does not suffer from issue #3061,
+        // so we can safely clear here as normal.
+        // NOTE: We put this line behind a conditional and
+        // an `if` to avoid clearing twice on other systems.
         #[cfg(not(any(target_os = "macos", windows)))]
         {
-            // On Wayland we can safely ignore this call, since the window isn't visible until you
-            // actually draw something into it and commit those changes.
-            if is_x11 {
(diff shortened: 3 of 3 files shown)


arvl130:
>what's the point of separating macos/windows/Wayland into separate blocks, when they're all doing the same thing?
It's the best way I could think of having Windows/MacOS and Wayland retain off-screen `api.clear()`-ing. The Wayland block relies on variables which only exist in "not macos and windows" conditionals. Whereas MacOS and Windows do not rely on these, so I've had put them in separate blocks. Is there something I'm missing here? Are there better ways of doing this?

>How does this address the issue with uninitialized surfaces being shown when running without a compositor?
Despite the current arrangement, doing it this way, we can actually have `set_visible`, `api.clear()`, and `swap_buffers` run right after each other to combat this. I haven't decided on committing this yet though, because I found that `set_visible` doesn't actually block until the window is actually shown. So for some time after calling `set_visible`, the window is still hidden actually (for like 5ms).

For now sleeping works. But I'm not sure how viable that would be submitting.

kchibisov:
>It's the best way I could think of having Windows/MacOS and Wayland retain off-screen api.clear()-ing. The Wayland block relies on variables which only exist in "not macos and windows" conditionals. Whereas MacOS and Windows do not rely on these, so I've had put them in separate blocks. Is there something I'm missing here? Are there better ways of doing this?

Wayland is handled is completely different, first of all, the window on Wayland is shown in `fn draw` due to how Wayland works. Don't add any logic to it there if you're not sure why Wayland isn't handled at all on master. There's even a comment explaining it somewhere, I suppose, so clearing it doesn't really required actually, but it's done for consistency with the rest. I'd propose to move the entire window show + clear thing right before `run_return` for all platforms. If you feel uncomfortable with doing so, I can take a look myself, when I'll have time.

arvl130:
>I'd propose to move the entire window show + clear thing right before run_return for all platforms. If you feel uncomfortable with doing so, I can take a look myself, when I'll have time.

@kchibisov I wouldn't mind doing so. The only reason I had avoided doing that was just to try not to change stuff in other platforms, if they were already working anyways.

Though, that still leaves the problem of latency between showing and clearing. Ideally, `set_visible` should block until the window is actually shown, so we don't have to sleep and wait before running our next call.

chrisduerr:
> Despite the current arrangement, doing it this way, we can actually have set_visible, api.clear(), and swap_buffers run right after each other to combat this. I haven't decided on committing this yet though, because I found that set_visible doesn't actually block until the window is actually shown.

Doing this twice will likely delay startup time unnecessarily.

> I'd propose to move the entire window show + clear thing right before run_return for all platforms.

You mean just showing the window later? What's the benefit of this, what's the goal? It seems like this would just delay the time until the window is shown?

kchibisov:
>You mean just showing the window later? What's the benefit of this, what's the goal? It seems like this would just delay the time until the window is shown?

The benefit is to process resize on start right away, since it's send to a visible window afaics. It won't delay startup, since you're still processing resize in `run_return` and resize to a usable size.

chrisduerr:
> The benefit is to process resize on start right away, since it's send to a visible window afaics.

If you move both `set_visible` and the clearing, that won't help though, right? I'd assume that what you mean by `the entire window show + clear thing`.

> It won't delay startup

It might not delay the time it takes to run `alacritty -e true` (though shifting things around could), but it still renders the initial window state at a later stage. Of course the time it takes for the first significant rendering to show up should be the same, but the time it takes from the command to run until the window shows up at all would be different.

That said, it wouldn't necessarily be bad. In fact we could even delay rendering until we have the first actual content to render or until we're sure everything is initialized. Since at least on X11 no space is reserved until the window is shown, there wouldn't be additional time of empty reserved space with tiling WMs like i3. But it might delay the time it takes for Alacritty to be responsive to keyboard input.

arvl130:
We could do the whole `show + clear` thing in `Display` creation, or we could do it in `Processor` (before `run_return`). Both should work. So which won't do I go for ?

>Doing this twice will likely delay startup time unnecessarily.

I'm not sure I understand this entirely, since I've set up `if`s  and conditionals just to avoid clearing twice (if that's what you mean). If there really needs to be a more generic way of clearing among all platforms (that is, we shouldn't be clearing in two different ways--even behind conditionals), shouldn't @kchibisov suggestion to apply this to all platforms be sufficient?

kchibisov:
just to clarify. I mean to move entire making window visible procedure right before run_return for all platforms. The same amount of code, just in different place, closer to event processing, so we will handle resize events asap.

arvl130:
@kchibisov Is that correct?

alacritty/src/event.rs
```diff
@@ -419,6 +420,19 @@ impl<N: Notify + OnResize> Processor<N> {
...
         T: EventListener,
     {
+        // Show the window.
+        self.display.window.set_visible(true);
+
+        // Sleep for 5ms.
+        //
+        // `set_visible` does not block until the window is shown, so we
+        // sleep here to avoid clearing while the window is still hidden.
+        thread::sleep(time::Duration::from_millis(5));
```
kchibisov, line 430: remove that?

alacritty/src/event.rs
```diff
@@ -419,6 +420,19 @@ impl<N: Notify + OnResize> Processor<N> {
...
+
+        // Clear screen.
+        debug!("Inner size before clearing: {:?}", self.display.window.inner_size());
```
kchibisov, line 433: And I guess that.

alacritty/src/event.rs
```diff
@@ -419,6 +420,19 @@ impl<N: Notify + OnResize> Processor<N> {
...
+        self.display.clear(&self.config);
```
kchibisov, line 434: And move it before `set_visible(true)`.

arvl130:
@kchibisov I see, so we simply move the off-screen clearing + show, closer to resize. So that the amount of time between clear is visible, and resize, is shorter ?

This should make the clear rectangle visible less frequently. But the bug is still there, we just see it less often.

In my testing though, it isn't too different with what's on master. It's just happens somewhat less often.

kchibisov:
Well, the proper solution will be to understand why do we have multiple resizes during startup and we don't have a proper size right away in tiling wm scenarios, I guess X11 is capable of doing so, right?

arvl130:
The way I understand it. Resizes happen because during startup, we're already anticipating our expected sizes for the window before it's actually shown.

arvl130:
Note: I decided to delete the message before this because I accidently hit Enter before the message's actually finished. The entirety of the comment is in here.

@kchibisov The way I understand it. The reason why don't have a proper size immediately at window initialization, is because we only rely on predetermined sizes in the config (paddings, columns, lines). Or if none of those are set, just the hints that X gives us (from what I've seen, this is *always just* 800x600).

Window managers only actually give us sizes once the window is shown, so I'm not sure if there are ways we can rightly anticipate the correct size without actually showing.

kchibisov:
>Window managers only actually give us sizes once the window is shown, so I'm not sure if there are ways we can rightly anticipate the correct size without actually showing.

This is just wrong tbh if it's true. Can't window manager advertise the size it wants from us so we can use it? How other apps fixes blinking on start? Does it happen under compositor? I don't have X11 setup myself to try all of that, but it just sounds strange if it's actually true and the only way to get size is to show the window. I'd like to see some X11 docs/specs/recommendations on how to handle such issues. So if you're willing to find them or explain how other opengl apps handle that it'll be nice, I guess.

arvl130:
>This is just wrong tbh if it's true. Can't window manager advertise the size it wants from us so we can use it?

I guess I should have prefaced that part. I'm not knowledgeable enough in X11 to know if it is possible in general to anticipated sizes before showing applications. That's just how I saw things work in Alacritty.

>I'd like to see some X11 docs/specs/recommendations on how to handle such issues. So if you're willing to find them or explain how other opengl apps handle that it'll be nice, I guess.

I guess I could try and see how things are done in other OpenGL terminals. Kitty is the only other terminal I know that is like this. If you have other applications in mind I could look into, that would be nice.

chrisduerr:
Whatever application you use, it's probably a good idea to make sure transparency is enabled or the background is white. Since you wouldn't see this problem if your background is configured to be entirely black and transparency in OpenGL is disabled.

arvl130:
Thanks! I'll keep that in mind. Will post here once I have more info.

PR #3777, opened 2020-05-27
Add documentation about patch releases

chrisduerr:


CONTRIBUTING.md
```diff
@@ -116,23 +116,35 @@ and the final version are only comitted and tagged in this branch. The master br
...
+ 10. The new commit in the branch is tagged as `v0.2.0`
+ 11. A GitHub release is created for the `v0.2.0` tag
+ 12. The changelog since the last stable release (**not** RC) is added to the GitHub release
+     description
+ 13. The `-dev` is stripped from the `0.2.0-dev` changelog entries on master
+
+On the master and with new planned releases, the patch version is never bumped, but only the minor
+is increased. This allows for creating releases to fix bugs which build on previous minor releases
+and just increment the patch version, without having to decrement or advance the version based on
+these unplanned releases.
```
cole-h, line 139:
```suggestion
On master and with new planned releases, the patch version is never bumped; only the minor version
is increased. This allows for creating releases to fix bugs which build on top of previous minor releases
and just increment the patch version.
```

I think the last part I removed is unnecessary. The preceding portion of the sentence already gives enough context on why this should be done this way, there's no reason to bring in decrementing versions, etc., IMHO.

(Sorry, you'll have to reflow yourself if you accept this...)

CONTRIBUTING.md
```diff
@@ -116,23 +116,35 @@ and the final version are only comitted and tagged in this branch. The master br
...
+
+The exact steps for an exemplary `0.2.3` release might look like this:
+ 1. Initially, the version on the latest master is `0.3.0-dev` and the latest release was `0.2.2`
+ 2. A new `v0.2.3` branch is forked from the `v0.2.2` branch
+ 4. All bug fixes are cherry-picked from master into the `v0.2.3` branch
+ 5. The version is bumped to `v0.2.3-rc1` and the changelog's `0.3.0` entries are moved to `0.2.3`
+ 6. Steps 5-12 for regular releases are identical
```
kchibisov, line 146: This seems a bit odd. I don't get the `Steps 5-12`, it's a different list with its own `5` tbw.

CONTRIBUTING.md
```diff
@@ -116,23 +116,35 @@ and the final version are only comitted and tagged in this branch. The master br
...
+On the master and with new planned releases, the patch version is never bumped, but only the minor
+is increased. This allows for creating releases to fix bugs which build on previous minor releases
+and just increment the patch version, without having to decrement or advance the version based on
+these unplanned releases.
+
+The exact steps for an exemplary `0.2.3` release might look like this:
+ 1. Initially, the version on the latest master is `0.3.0-dev` and the latest release was `0.2.2`
+ 2. A new `v0.2.3` branch is forked from the `v0.2.2` branch
+ 4. All bug fixes are cherry-picked from master into the `v0.2.3` branch
+ 5. The version is bumped to `v0.2.3-rc1` and the changelog's `0.3.0` entries are moved to `0.2.3`
```
cole-h, line 145: Step 5 is somewhat unclear to me. Why are 0.3.0's changelog entries moved to 0.2.3? 0.2.3 was forked from 0.2.2, and only bugfixes are being backported, so why does 0.3.0 need to be involved here?

kchibisov: it's `0.3.0-dev` are moved to `v0.2.3` branch.

>Why are 0.3.0's changelog entries moved to 0.2.3? 0.2.3 was forked from 0.2.2, and only bugfixes are being backported, so why does 0.3.0 need to be involved here?

Bugfixes are from `0.3.0-dev`, not from `0.2.2`, etc. So changelog for bugfixes is on `0.3.0-dev`, so they should be backported, and once release is done you add new entry for `0.2.3` and remove entires from `0.3.0-dev`

cole-h: Ah, right; obvious now that you mention it. Thanks for clearing that up.

CONTRIBUTING.md
chrisduerr: The last part is not irrelevant though, because you could build on top of previous minor releases and just increment the patch version even if you bumped the patch on master too.

The difference is just that you do not have to deal with incrementing or decrementing the master, which is why I explicitly pointed this out.

CONTRIBUTING.md
chrisduerr: Because that's where the bugfixes are ported from.

CONTRIBUTING.md
```diff
@@ -133,17 +133,16 @@ The exact steps for an exemplary `0.2.0` release might look like this:
      description
  13. The `-dev` is stripped from the `0.2.0-dev` changelog entries on master
 
-On the master and with new planned releases, the patch version is never bumped, but only the minor
-is increased. This allows for creating releases to fix bugs which build on previous minor releases
-and just increment the patch version, without having to decrement or advance the version based on
-these unplanned releases.
+On master and with new planned releases, only the minor version is bumped. This makes it possible to
+create bug fix releases by incrementing the patch version of previous minor releases, without having
```
cole-h, line 137:
```suggestion
create bug fix releases by incrementing the patch version of {a/the} previous minor release, without having
```

PR #3802, opened 2020-06-01
Remove winpty reference from INSTALL.md

BYK:
Windows instructions in `INSTALL.md` still refers to WinPTY which is no longer automatically built. This patch removes the reference.


BYK:
Oh wait, I guess this was brought back. Sorry, closing.

PR #3805, opened 2020-06-01
Add feature gate for WinPTY

davidhewitt:
A feature gate for WinPTY support. This should make it easier for those who want to contribute and have Windows as a primary OS.

(`winpty-sys` crate is slow and painful to build; it needs a lot of C/C++ packages installed on your machine, including llvm, before it will compile.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Option `cursor.thickness` to set terminal cursor thickness
 - Font fallback on Windows
 - Support for Fontconfig embolden and matrix options
+- Feature gate for WinPTY support
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Option `cursor.thickness` to set terminal cursor thickness
 - Font fallback on Windows
 - Support for Fontconfig embolden and matrix options
-- Feature gate for WinPTY support
+- Opt-out compilation flag `winpty` to disable WinPTY support
 
 ### Changed
 
--- a/alacritty/Cargo.toml
+++ b/alacritty/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2018"
 [dependencies]
 alacritty_terminal = { path = "../alacritty_terminal", default-features = false }
 clap = "2"
-log = "0.4"
+log = { version = "0.4", features = ["std"] }
 time = "0.1.40"
 fnv = "1"
 serde = { version = "1", features = ["derive"] }
@@ -47,9 +47,10 @@ winapi = { version = "0.3.7", features = ["impl-default", "wincon"]}
 embed-resource = "1.3"
 
 [features]
-default = ["wayland", "x11"]
+default = ["wayland", "x11", "winpty"]
 x11 = ["alacritty_terminal/x11"]
 wayland = ["alacritty_terminal/wayland"]
+winpty = ["alacritty_terminal/winpty"]
 # Enabling this feature makes shaders automatically reload when changed
 live-shader-reload = []
 nightly = []
--- a/alacritty_terminal/src/tty/windows/mod.rs
+++ b/alacritty_terminal/src/tty/windows/mod.rs
@@ -24,20 +24,20 @@ use crate::term::SizeInfo;
 use crate::tty::windows::child::ChildExitWatcher;
 use crate::tty::{ChildEvent, EventedPty, EventedReadWrite};
 
+#[cfg(feature = "winpty")]
+mod automatic_backend;
 mod child;
 mod conpty;
-#[cfg(feature = "winpty")]
-mod switchable_backends;
 #[cfg(feature = "winpty")]
 mod winpty;
 
 #[cfg(not(feature = "winpty"))]
-pub use conpty::Conpty as Backend;
+use conpty::Conpty as Backend;
 #[cfg(not(feature = "winpty"))]
-pub use mio_anonymous_pipes::{EventedAnonRead as ReadPipe, EventedAnonWrite as WritePipe};
+use mio_anonymous_pipes::{EventedAnonRead as ReadPipe, EventedAnonWrite as WritePipe};
 
 #[cfg(feature = "winpty")]
-pub use switchable_backends::{
+use automatic_backend::{
     EventedReadablePipe as ReadPipe, EventedWritablePipe as WritePipe, PtyBackend as Backend,
 };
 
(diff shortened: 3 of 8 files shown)


alacritty/Cargo.toml
```diff
@@ -11,7 +11,7 @@ edition = "2018"
 [dependencies]
 alacritty_terminal = { path = "../alacritty_terminal", default-features = false }
 clap = "2"
-log = "0.4"
+log = { version = "0.4", features = ["std"] }
```
davidhewitt, line 14: This was necessary for builds with `--no-default-features` to work.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Option `cursor.thickness` to set terminal cursor thickness
 - Font fallback on Windows
 - Support for Fontconfig embolden and matrix options
+- Feature gate for WinPTY support
```
```suggestion
- Opt-out compilation flag `winpty` to disable WinPTY support
```

alacritty/Cargo.toml
chrisduerr: That seems strange to me? `--no-default-features` should only affect the root crate, not its dependencies, right?

On the latest master it seems to work just fine?

alacritty_terminal/src/tty/windows/switchable_backends.rs
```diff
@@ -0,0 +1,188 @@
+/// Types to wrap either Conpty or Winpty backends.
```
chrisduerr, line 1: The name `switchable_backends` sounds strange to me. Why did you choose this name?

alacritty/Cargo.toml
davidhewitt: Yeah hmm, it does indeed work on latest master with `--no-default-features`. On this branch, without this change, I get:

```
> cargo run --no-default-features
   Compiling alacritty v0.5.0-dev (C:\Users\david\dev\alacritty\alacritty)
error[E0425]: cannot find function `set_boxed_logger` in crate `log`
  --> alacritty\src\logging.rs:47:10
   |
47 |     log::set_boxed_logger(Box::new(logger))?;
   |          ^^^^^^^^^^^^^^^^ not found in `log`
```

alacritty_terminal/src/tty/windows/switchable_backends.rs
davidhewitt: Agreed, I don't love the name either. I considered alternatives like `dynamic_backend` but it doesn't use dynamic dispatch so didn't really like that much either.

Another option is to just put all of this stuff in the `winpty.rs` file, as it's only needed when the `winpty` feature is set.

alacritty/Cargo.toml
chrisduerr: Which directory are you running that command from?

davidhewitt: The `alacritty` subdirectory.

alacritty_terminal/src/tty/windows/switchable_backends.rs
chrisduerr: Why did you change stuff around so much? To prevent dead code warnings?

davidhewitt: Yes. An alternative would be to undo this move (put it back in `mod.rs`), and just add `#[cfg(feature = "winpty")]` to a lot more of the code there instead. Would that be preferable?

alacritty/Cargo.toml
chrisduerr: I've just checked and the `log` crate does not have any default features (see https://github.com/rust-lang/log/blob/master/Cargo.toml). Could you confirm again this actually prevents building?

alacritty_terminal/src/tty/windows/winpty.rs
```diff
@@ -71,9 +71,9 @@ pub fn new<C>(config: &Config<C>, size: &SizeInfo, _window_id: Option<usize>) ->
     let child_watcher = ChildExitWatcher::new(agent.raw_handle()).unwrap();
 
     Pty {
-        backend: super::PtyBackend::Winpty(agent),
-        conout: super::EventedReadablePipe::Named(conout_pipe),
-        conin: super::EventedWritablePipe::Named(conin_pipe),
+        backend: super::Backend::from(agent),
+        conout: super::ReadPipe::from(conout_pipe),
+        conin: super::WritePipe::from(conin_pipe),
```
chrisduerr, line 76: Is there any reason to use the renamed values here? Seems like that just increases complexity.

alacritty_terminal/src/tty/windows/mod.rs
```diff
@@ -33,161 +26,46 @@ use crate::tty::{ChildEvent, EventedPty, EventedReadWrite};
...
-}
-
-enum PtyBackend {
-    Winpty(winpty::Agent),
-    Conpty(conpty::Conpty),
-}
+#[cfg(feature = "winpty")]
+pub use switchable_backends::{
+    EventedReadablePipe as ReadPipe, EventedWritablePipe as WritePipe, PtyBackend as Backend,
+};
```
chrisduerr, line 42: I'm not a fan of `pub use`'ing a bunch of renamed stuff here. It makes it incredibly difficult to see what types actually are and just sends people on a wild goose hunt to figure out what the types actually are.

Things should be re-imported as appropriate.

alacritty_terminal/src/tty/windows/mod.rs
```diff
@@ -33,161 +26,46 @@ use crate::tty::{ChildEvent, EventedPty, EventedReadWrite};
 
 mod child;
 mod conpty;
+#[cfg(feature = "winpty")]
+mod switchable_backends;
```
chrisduerr, line 30: Please rename this to `automatic_backend`.

alacritty_terminal/src/tty/windows/switchable_backends.rs
chrisduerr, line 1:
```suggestion
/// Types to determine the appropriate PTY backend at runtime.
```

alacritty_terminal/src/tty/windows/switchable_backends.rs
```diff
@@ -0,0 +1,188 @@
...
+///
+/// When the `winpty` feature is enabled, these will be used in the Pty
+/// struct instead of the Conpty types.
```
chrisduerr, line 4: I don't think it's important where these will be used, that's documented where they are used. What should be documented is what they do, not how they are used.

Unless the `winpty` feature is disabled, the PTY backend will automatically fall back to WinPTY when the newer ConPTY API is not supported, as long as the user hasn't explicitly opted into the WinPTY config option.

Might be a bit of a long sentence, but I think that should get better across what *this* module does, not what other modules might do with this module.

alacritty_terminal/src/tty/windows/mod.rs
davidhewitt: That's a nice name!

alacritty_terminal/src/tty/windows/winpty.rs
davidhewitt: I was going for consistency with `conpty.rs`, where the renamed values will map to different concrete types depending on whether the `winpty` feature is enabled.

I'll change back to using the concrete types.

alacritty/Cargo.toml
davidhewitt: Hmm yes, it's definitely still an issue for me on this branch. Technically it's not wrong, as docs for `set_boxed_logger` state that `std` feature needs to be present, but it's weird that it's not an issue on master...

chrisduerr: Hmm, this might be due to a different dependency opting into that feature. Cargo's a bit special about that. Should be good, thanks.

alacritty_terminal/src/tty/windows/mod.rs
davidhewitt: Changed to just `use` instead of `pub use` for the renamed imports. They're now only used in this file, and I've added a private function `Pty::new` which takes args `impl Into<Backend>` etc. so that `conpty.rs` and `winpty.rs` don't need to import the concrete types.

davidhewitt:
Thanks for the feedback. That's feeling much cleaner!

Issue #3767, opened 2020-05-25
Multiple files dragging.


mysliwy112:
Dropping multiple files into terminal, doesn't generate spaces between filenames, rendering this function kinda useless.

https://github.com/alacritty/alacritty/issues/1335#issuecomment-397305237
In this usage case it would probably be more useful to use quotation marks for names, because inserting them between multiple names would be time-taking and misleading.

chrisduerr:
Could you share the output of `alacritty --print-events | grep `Dropped`?

mysliwy112:
Eh I should probably add this too:

OS: Windows 7
Backend: WinPTY
Alacritty: 0.4.2

Log:
```
[2020-05-26 09:39:57.818004000] [INFO] glutin event: WindowEvent { window_id: WindowId(WindowId(0xa03c0)), event: DroppedFile("E:\\a - b.txt") }
[2020-05-26 09:39:57.819004000] [INFO] glutin event: WindowEvent { window_id: WindowId(WindowId(0xa03c0)), event: DroppedFile("E:\\b - c.txt") }
[2020-05-26 09:39:57.819004000] [INFO] glutin event: WindowEvent { window_id: WindowId(WindowId(0xa03c0)), event: DroppedFile("E:\\a.txt") }
```

Using bash for this one, but problem works identical with cmd and powershell.
Alacritty window:
`$ E:\a - b.txtE:\b - c.txtE:\a.txt`

chrisduerr:
This isn't windows specific. Fundamentally the problem is just that every file dropped is completely separate. So we don't get an event that three files were dropped but three events that one file was dropped.

So we can't just easily add a space between these, unless we want to add a space after every dropped file. Though I think that might make some sense? What do you think @kchibisov?

kchibisov:
>So we can't just easily add a space between these, unless we want to add a space after every dropped file. Though I think that might make some sense? What do you think @kchibisov?

we can. we also can record a timestamp and try to insert space before dragged file, if one more event appeared to soon.

chrisduerr:
Making it timing based makes little sense, if anything it should be based on input being received between two files. Though an upper limit might be useful.

kchibisov:
I feel like simply adding space in the end should work just fine tbqh.

chrisduerr:
Yeah, I agree. Would be trivial to fix too if it's not desired by simply backspacing once.

PR #3808, opened 2020-06-02
Fix dropping of multiple files

FernandoPicazo:
This pull request is a fix for the problem occurring in #3767 where dragging multiple files into the terminal doesn't automatically separate the file names. This just appends a space to the string being written to the pty.

Apologies if the code style isn't write this is my first contribution to a Rust project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Shift + Backspace` now sends `^?` instead of `^H`
 - Default color scheme is now `Tomorrow Night` with the bright colors of `Tomorrow Night Bright`
 - Set IUTF8 termios flag for improved UTF8 input support
+- Dragging multiple files into terminal now adds a space after each file
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Shift + Backspace` now sends `^?` instead of `^H`
 - Default color scheme is now `Tomorrow Night` with the bright colors of `Tomorrow Night Bright`
 - Set IUTF8 termios flag for improved UTF8 input support
+- Dragging multiple files into terminal now adds a space after each file
```
kchibisov, line 38: This is not entirely related to multiple files, just say that space is added after dragging file, I guess.

FernandoPicazo: You got. I’ve changed it to reflect that it is after each file name. Not just multiple.

alacritty/src/event.rs
```diff
@@ -629,8 +629,9 @@ impl<N: Notify + OnResize> Processor<N> {
                         }
                     },
                     WindowEvent::DroppedFile(path) => {
+                        const PATH_SEPARATOR: &str = " ";
```
chrisduerr, line 632: No point in defining this as a constant here locally. We can just inline it.

FernandoPicazo: I've removed the constant definition and just used the string literal.

PR #3811, opened 2020-06-03
Fix font regression on macOS

### LINKED ISSUE #3809: Regression in font rendering in 0.4.3 vs 0.4.2
### System:

OS: macOS 10.14.6

### Settings:

- `use_thin_strokes: false`
- `defaults write -g CGFontRenderingFontSmoothingDisabled -bool NO`

There is already some discussion of this in the comments of #3804, but I'm opening a separate issue specifically for this issue.

With `use_thin_strokes: false` set, font rendering appears to be worse in 0.4.3 than it was in 0.4.2:

0.4.2:
image

0.4.3:
image

The change is quite subtle, but it definitely looks less smooth in 0.4.3. For what it's worth, this (0.4.3) looks similar to other font rendering on my system when `defaults write -g CGFontRenderingFontSmoothingDisabled -bool NO` is not set (i.e. without subpixel anti-aliasing).

thornjad:
The `0.0` alpha background works great for emoji, but normal text does not render well on macOS. Instead, use `1.0` for non-pre-colored glyphs.

Fixes #3809

### DIFF
--- a/font/src/darwin/mod.rs
+++ b/font/src/darwin/mod.rs
@@ -486,7 +486,6 @@ impl Font {
             self.glyph_index(character).ok_or_else(|| Error::MissingGlyph(character))?;
 
         let bounds = self.bounding_rect_for_glyph(Default::default(), glyph_index);
-        let is_colored = self.is_colored();
 
         let rasterized_left = bounds.origin.x.floor() as i32;
         let rasterized_width =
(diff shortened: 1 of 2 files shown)


font/src/darwin/mod.rs
```diff
@@ -486,6 +486,7 @@ impl Font {
             self.glyph_index(character).ok_or_else(|| Error::MissingGlyph(character))?;
 
         let bounds = self.bounding_rect_for_glyph(Default::default(), glyph_index);
+        let is_colored = self.is_colored();
```
chrisduerr, line 489: Could you move this further down to where it is first used?

PR #3825, opened 2020-06-06
Simplify internal DPRChanged event

kchibisov:
Since winit guarantees to send resize after ScaleFactorChange we
don't need to manually update size in DPRChanged event.

The bug mentioned in a comment on DPRChanged is fixed in winit for some time already.

### DIFF
--- a/alacritty_terminal/src/event.rs
+++ b/alacritty_terminal/src/event.rs
@@ -6,7 +6,7 @@ use crate::term::SizeInfo;
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum Event {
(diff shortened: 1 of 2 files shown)


alacritty_terminal/src/event.rs
```diff
@@ -6,7 +6,7 @@ use crate::term::SizeInfo;
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum Event {
-    DPRChanged(f64, (u32, u32)),
+    DPRChanged(f64),
```
chrisduerr, line 9: Could we move this to below `Message(Message)`? For no good reason other than it looking better. :)

PR #3829, opened 2020-06-06
Update dependencies

kchibisov:
basically updates base64 and some other deps together.


kchibisov:
Though, I'm not entirely sure about the bump of base64, since old version is still in the tree.

chrisduerr:
> Though, I'm not entirely sure about the bump of base64, since old version is still in the tree.

This should be fixed with the following redox PR:
https://gitlab.redox-os.org/redox-os/users/-/merge_requests/34

PR #3834, opened 2020-06-08
Cleanup window.rs imports

chrisduerr:
The window.rs imports have been extremely messy due to heavy
platform-specific import usage. While some of them are just stray
imports for a single platform, a lot of these are specific to linux/bsd.

To make these a little easier to read and maintain, the linux/bsd
specific imports were grouped together, since this allows us to just
have a single platform annotation.

Most of our imports follow the order of how relevant and foreign these
imports are to Alacritty, putting our own frequently used imports at the
far bottom and STD with things people usually don't need to care about
at the top. Since platform specific imports are generally important to
fewer people, I've decided to put them before all the other imports
rather than behind them.

### DIFF
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -1,37 +1,38 @@
-use std::convert::From;
+#[rustfmt::skip]
 #[cfg(not(any(target_os = "macos", windows)))]
-use std::ffi::c_void;
+use {
+    std::ffi::c_void,
+    std::os::raw::c_ulong,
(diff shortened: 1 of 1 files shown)


alacritty/src/window.rs
```diff
@@ -1,52 +1,46 @@
-use std::convert::From;
+#[rustfmt::skip]
```
chrisduerr, line 1: This is required to preserve the order of elements and white space between them.

PR #3846, opened 2020-06-14
Improve documentation for macos install with brew

sdondley:
There is possible confusion with users who don't read entire INSTALL.md file linked to when installing with macos install with brew. This adds an instruction to clone the alacritty repo before trying to install the manual.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Improve documentation for macos install with brew
 - Block cursor is no longer inverted at the start/end of a selection
 - Preserve selection on non-LMB or mouse mode clicks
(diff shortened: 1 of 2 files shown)


chrisduerr:
I've updated the documentation to just reference the GitHub releases page. I think everyone installing a terminal should be capable to replace the path with that information added.

Thanks for your contribution!

kchibisov:
I see what you mean, but I guess the instructions are bit mismatches tbqh. Like on a link it requires you to compress and install man page, however on our release page manpage is already compressed so it'll require just copy.

chrisduerr:
That's a good point, it should probably just be changed to say that the auxiliary files should be copied to the appropriate locations.

kchibisov:
>That's a good point, it should probably just be changed to say that the auxiliary files should be copied to the appropriate locations.

But terminfo should be complied, so some of them should be copied and some of them handled like in INSTALL.md

kchibisov:
Thanks.

PR #3847, opened 2020-06-14
Override default bindings when notmodes intersect

### LINKED ISSUE #3476: #3315 breaks arrow key navigation in Neovim
With #3315 merged, navigation inside of Neovim with arrow keys is broken.

What happens is that the key press seems to trigger twice.
So if I hit up arrow, the cursor will jump up to rows instead of one and so on.
This happens in drop down prompts and other UI elements as well if navigated with the arrow keys.

If I rollback to the commit before #3315 (64db7d3), everything works fine again.

### System

OS: Gentoo Linux
Version: alacritty 0.5.0-dev (1a8cd17)
WM: Sway (Wayland)

kchibisov:
Make notmodes follow modes in terms of override behavior, so
default binding won't be pruned if and only if its modes intersect
with new binding notmodes and vise-versa.

Fixes #3476

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -49,9 +49,9 @@ impl<T: Eq> Binding<T> {
         // the most likely item to fail so prioritizing it here allows more
         // checks to be short circuited.
         self.trigger == *input
+            && self.mods == mods
             && mode.contains(self.mode)
             && !mode.intersects(self.notmode)
-            && (self.mods == mods)
     }
 
     #[inline]
@@ -61,16 +61,20 @@ impl<T: Eq> Binding<T> {
             return false;
         }
 
-        // Completely empty modes match all modes.
-        if (self.mode.is_empty() && self.notmode.is_empty())
-            || (binding.mode.is_empty() && binding.notmode.is_empty())
-        {
-            return true;
+        let selfmode = if self.mode.is_empty() { TermMode::ANY } else { self.mode };
+        let bindingmode = if binding.mode.is_empty() { TermMode::ANY } else { binding.mode };
+
+        if !selfmode.intersects(bindingmode) {
+            return false;
         }
 
-        // Check for intersection (equality is required since empty does not intersect itself).
-        (self.mode == binding.mode || self.mode.intersects(binding.mode))
-            && (self.notmode == binding.notmode || self.notmode.intersects(binding.notmode))
+        // The bindings are never active at the same time when the required modes of one binding
+        // are part of the forbidden bindings of the other.
+        if self.mode.intersects(binding.notmode) || binding.mode.intersects(self.notmode) {
+            return false;
+        }
+
+        true
     }
 }
 
@@ -1023,6 +1027,7 @@ mod tests {
         b2.mode = TermMode::ALT_SCREEN;
 
         assert!(b1.triggers_match(&b2));
+        assert!(b2.triggers_match(&b1));
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


alacritty/src/config/bindings.rs
```diff
@@ -1023,6 +1028,7 @@ mod tests {
         b2.mode = TermMode::ALT_SCREEN;
 
         assert!(b1.triggers_match(&b2));
+        assert!(b2.triggers_match(&b1));
```
kchibisov, line 1031: Done on purpose to make every test symmetric, so it won't get borken in the future.

alacritty/src/config/bindings.rs
```diff
@@ -61,16 +61,21 @@ impl<T: Eq> Binding<T> {
...
+        let selfmode = if self.mode.is_empty() { TermMode::ANY } else { self.mode };
+
+        let bindingmode = if binding.mode.is_empty() { TermMode::ANY } else { binding.mode };
+
+        if !selfmode.intersects(bindingmode) {
+            return false;
+        }
+
+        // If the binding requires a mode while the other binding is never bound, they are not in
+        // conflict.
```
chrisduerr, line 73:
```suggestion
        // The bindings are never active at the same time when the required modes of one binding are part of the forbidden bindings of the other.
```

I think that's a little easier to understand.

alacritty/src/config/bindings.rs
```diff
@@ -61,16 +61,21 @@ impl<T: Eq> Binding<T> {
...
-        // Completely empty modes match all modes.
-        if (self.mode.is_empty() && self.notmode.is_empty())
-            || (binding.mode.is_empty() && binding.notmode.is_empty())
-        {
-            return true;
+        let selfmode = if self.mode.is_empty() { TermMode::ANY } else { self.mode };
+
+        let bindingmode = if binding.mode.is_empty() { TermMode::ANY } else { binding.mode };
+
+        if !selfmode.intersects(bindingmode) {
```
chrisduerr, line 68:
```suggestion
        // If none of the required modes intersect, we can be confident that the bindings are not in conflict, since they will never be active at the same time.
        if !selfmode.intersects(bindingmode) {
```

alacritty/src/config/bindings.rs
```diff
@@ -61,16 +61,21 @@ impl<T: Eq> Binding<T> {
             return false;
         }
 
-        // Completely empty modes match all modes.
-        if (self.mode.is_empty() && self.notmode.is_empty())
-            || (binding.mode.is_empty() && binding.notmode.is_empty())
-        {
-            return true;
+        let selfmode = if self.mode.is_empty() { TermMode::ANY } else { self.mode };
+
```
chrisduerr, line 65:
```suggestion
```

alacritty/src/config/bindings.rs
```diff
@@ -61,16 +61,22 @@ impl<T: Eq> Binding<T> {
...
-        // Completely empty modes match all modes.
-        if (self.mode.is_empty() && self.notmode.is_empty())
-            || (binding.mode.is_empty() && binding.notmode.is_empty())
-        {
-            return true;
+        let selfmode = if self.mode.is_empty() { TermMode::ANY } else { self.mode };
+        let bindingmode = if binding.mode.is_empty() { TermMode::ANY } else { binding.mode };
+
+        // If none of the required modes intersect, we can be confident that the bindings
+        // are not in conflict, since they will never be active at the same time.
```
chrisduerr, line 68:
```suggestion
```

This is wrong.

PR #3848, opened 2020-06-14
Add option to run command on bell

### LINKED ISSUE #1528: Add support for audio bell

kchibisov:
Fixes #1528.


kchibisov:
The visual bell in `term/mod.rs` was kept that way intentionally, since it's actually a visual bell.

kchibisov:
Configuration follows https://github.com/alacritty/alacritty/pull/2298#pullrequestreview-225782173

alacritty.yml
```diff
@@ -245,31 +245,46 @@
...
+  # This program is executed whenever the bell is rung.
+  #
+  # When set to `command: None`, bell command will be disabled completely.
+  #
+  # Example:
+  #   command:
+  #     program: notify-send
+  #     args: ["Hello, World!"]
+  #
+  #command: None
```
chrisduerr, line 287:
```suggestion
  #program: None
```

alacritty_terminal/src/config/mod.rs
```diff
@@ -115,6 +115,10 @@ pub struct Config<T> {
     #[serde(skip)]
     pub hold: bool,
 
+    // TODO: DEPRECATED
+    #[serde(default, deserialize_with = "failure_default")]
+    pub visual_bell: Option<BellConfig>,
```
chrisduerr, line 120:
```suggestion
    visual_bell: Option<BellConfig>,
```

alacritty_terminal/src/config/visual_bell.rs
```diff
@@ -2,47 +2,73 @@ use std::time::Duration;
```
chrisduerr, line 2: This file needs to be deleted I think?

alacritty_terminal/src/term/mod.rs
```diff
@@ -553,10 +553,10 @@ fn cubic_bezier(p0: f64, p1: f64, p2: f64, p3: f64, x: f64) -> f64 {
 
 impl VisualBell {
     pub fn new<C>(config: &Config<C>) -> VisualBell {
-        let visual_bell_config = &config.visual_bell;
+        let bell_config = config.bell();
```
chrisduerr, line 556: Could we maybe implement `From<Bell> for VisualBell`?

alacritty_terminal/src/config/bell.rs
```diff
@@ -0,0 +1,105 @@
...
+    #[serde(deserialize_with = "failure_default")]
+    pub animation: BellAnimation,
+
+    /// Visual bell duration in milliseconds.
+    #[serde(deserialize_with = "failure_default")]
+    duration: u16,
+
+    // FIXME RGB deserialization always successedes and uses #000000 color. However it should
+    // fallback to a proper color. So it should be reworked either with
+    // https://github.com/alacritty/alacritty/pull/3507 or after that PR will land.
```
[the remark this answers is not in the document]
kchibisov: I mean that you'll start doing manual deserialization, right? So it'll be possible to do so as well here, since Rgb will be free of hardcode to `#000000`. 

**NOTE** I don't introduce new bug here, since it was like that already.

alacritty_terminal/src/config/bell.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Visual bell flash color.
+    #[serde(deserialize_with = "failure_default")]
+    pub color: Rgb,
+
+    /// Command to run on bell.
+    #[serde(deserialize_with = "deserialize_command")]
+    pub command: Option<Program>,
```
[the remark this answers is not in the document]
kchibisov: As discussed on IRC we'll keep `command` since it's used in bindings already, so it fits default config more.

alacritty_terminal/src/config/mod.rs
kchibisov: It's required to throw deprecation warning.

alacritty_terminal/src/config/bell.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    // FIXME RGB deserialization always successedes and uses #000000 color. However it should
+    // fallback to a proper color. So it should be reworked either with
+    // https://github.com/alacritty/alacritty/pull/3507 or after that PR will land.
+    /// Visual bell flash color.
+    #[serde(deserialize_with = "failure_default")]
+    pub color: Rgb,
+
+    /// Command to run on bell.
+    #[serde(deserialize_with = "deserialize_command")]
```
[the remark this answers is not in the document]
kchibisov: `Program` doesn't have `Default` and I can't impl it, before there's none logically, so can't do that.

alacritty_terminal/src/config/bell.rs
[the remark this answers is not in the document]
chrisduerr: Not for all fields, that seems to be pretty much impossible since we have 16 colors. So having 16 different deserialization functions just doesn't make any sense.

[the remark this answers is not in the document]
kchibisov: Hm, so how should I handle deserialization of Rgb then?

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Font fallback on Windows
 - Support for Fontconfig embolden and matrix options
 - Opt-out compilation flag `winpty` to disable WinPTY support
+- Option to run command on bell which can be set in `bell.command`
```
chrisduerr, line 27: We should also add a change log entry for moving visual_bell option to bell.

alacritty/src/event.rs
```diff
@@ -758,6 +759,15 @@ impl<N: Notify + OnResize> Processor<N> {
         processor.ctx.terminal.dirty = true;
     }
 
+    fn start_program(program: &Program) {
+        let args = program.args().to_vec();
+
+        match start_daemon(program.program(), &args) {
+            Ok(_) => debug!("Launched {} with args {:?}", program.program(), args),
+            Err(_) => warn!("Unable to launch {} with args {:?}", program.program(), args),
+        }
```
chrisduerr, line 768: We have this construct in a lot of places, should it maybe be moved to `start_daemon`?

If it's used just for visual bell, it probably shouldn't be `start_program`, but I see nothing particularly visual bell specific here.

alacritty_terminal/src/config/bell.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub animation: BellAnimation,
+
+    /// Visual bell duration in milliseconds.
+    #[serde(deserialize_with = "failure_default")]
+    duration: u16,
+
+    // FIXME RGB deserialization always successedes and uses #000000 color. However it should
+    // fallback to a proper color. So it should be reworked either with
+    // https://github.com/alacritty/alacritty/pull/3507 or after that PR will land, since it'll
+    // allow to impl manual deserialization properly.
```
chrisduerr, line 26: Adding a reminder that this needs to be done before the PR can be merged.

alacritty_terminal/src/config/bell.rs
```diff
@@ -0,0 +1,106 @@
...
+{
+    // Deserialize to generic value.
+    let val = Value::deserialize(deserializer)?;
+
+    // Accept `None` to disable the bell command.
+    if val.as_str().filter(|v| v.to_lowercase() == "none").is_some() {
+        return Ok(None);
+    }
+
+    match <Option<Program>>::deserialize(val) {
```
chrisduerr, line 67:
```suggestion
    match Program::deserialize(val) {
```

alacritty_terminal/src/config/bell.rs
```diff
@@ -0,0 +1,106 @@
...
+
+impl BellConfig {
+    /// Visual bell duration in milliseconds.
+    #[inline]
+    pub fn duration(&self) -> Duration {
+        Duration::from_millis(u64::from(self.duration))
+    }
+}
+
+fn deserialize_command<'a, D>(deserializer: D) -> std::result::Result<Option<Program>, D::Error>
```
chrisduerr, line 55:
```suggestion
fn deserialize_bell_command<'a, D>(deserializer: D) -> Result<Option<Program>, D::Error>
```

alacritty_terminal/src/config/bell.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    // Accept `None` to disable the bell command.
+    if val.as_str().filter(|v| v.to_lowercase() == "none").is_some() {
+        return Ok(None);
+    }
+
+    match <Option<Program>>::deserialize(val) {
+        Ok(command) => Ok(command),
+        Err(err) => {
+            error!(target: LOG_TARGET_CONFIG, "Problem with config: {}; not using anything", err);
```
chrisduerr, line 70:
```suggestion
            error!(target: LOG_TARGET_CONFIG, "Problem with config: {}; ignoring field", err);
```

alacritty/src/event.rs
kchibisov: We can't abstract, since messages are different everywhere, so you may move, but you should supply the messages, so I feel like it's fine the way it is now.

alacritty_terminal/src/config/bell.rs
kchibisov: I mean it's not a bug from that PR so I'm not sure how it's related and prevents it from being merged, I've just highlighted a very old issue.

alacritty_terminal/src/config/bell.rs
kchibisov: The field is an option itself, so either do that or manually using `Ok(Some(command))`, so there's really not much difference, and I'd stick with the current approach, since we're using it in launcher deser already.

kchibisov:
Blocked by #3507

kchibisov:
I've rebased pr and solved some the remaining issue on that PR. The first commit is just rebase, and the last one is resolution of that one issue we had.

alacritty/src/event.rs
```diff
@@ -839,8 +839,15 @@ impl<N: Notify + OnResize> Processor<N> {
...
-                    TerminalEvent::Urgent => {
-                        processor.ctx.window.set_urgent(!processor.ctx.terminal.is_focused)
+                    TerminalEvent::Bell => {
+                        let _ = processor
+                            .ctx
+                            .config
+                            .bell()
+                            .command
+                            .as_ref()
+                            .map(Self::start_bell_program);
```
chrisduerr, line 849: This formatting seems very unfortunate. Could we maybe extract a `let bell_config = processor.ctx.config.bell();` to make this use only two lines? I don't care how it is achieved, but this seems very wasteful. `if let Some(xxx)` would be another option I suppose, since mapping just to execute is kinda meh anyways.

alacritty/src/event.rs
chrisduerr: Do the messages need to be different? Or would we be fine with just having a uniform message for launching processes and having that  fail?

I feel like it's probably not super necessary to be able to have a custom message for every different way we want to spawn a program?

alacritty_terminal/src/config/bell.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub command: Option<Program>,
+}
+
+impl Default for BellConfig {
+    fn default() -> Self {
+        Self {
+            animation: Default::default(),
+            duration: Default::default(),
+            color: DEFAULT_BELL_COLOR,
+            command: Default::default(),
```
chrisduerr, line 38:
```suggestion
            command: Default::default(),
            color: DEFAULT_BELL_COLOR,
```

Doesn't matter at all, but this would look so much better. :)

alacritty_terminal/src/config/bell.rs
chrisduerr: I feel like that probably adds unnecessary overhead though and it doesn't clearly communicate that this cannot fail. While just using `Ok(Some(command))` makes it clear that it can never be `None` at this point.

I feel like `Program::deserialize` is the better choice here.

alacritty/src/event.rs
```diff
@@ -1078,6 +1079,17 @@ impl<N: Notify + OnResize> Processor<N> {
     }
 }
 
+fn start_program<I, S>(program: &str, args: I)
+where
+    I: IntoIterator<Item = S> + Debug + Copy,
+    S: AsRef<OsStr>,
+{
+    match start_daemon(program, args) {
```
chrisduerr, line 1087: Do we not want to provide this functionality globally in `start_daemon`? Are there instances where we do not want to print a warning?

alacritty/src/event.rs
```diff
@@ -839,8 +835,13 @@ impl<N: Notify + OnResize> Processor<N> {
...
                     TerminalEvent::Title(title) => processor.ctx.window.set_title(&title),
                     TerminalEvent::Wakeup => processor.ctx.terminal.dirty = true,
-                    TerminalEvent::Urgent => {
-                        processor.ctx.window.set_urgent(!processor.ctx.terminal.is_focused)
+                    TerminalEvent::Bell => {
+                        let bell_config = processor.ctx.config.bell();
+                        let _ = bell_config
+                            .command
+                            .as_ref()
+                            .map(|program| start_program(program.program(), program.args()));
```
chrisduerr, line 843:
```suggestion
                        let bell_command = processor.ctx.config.bell().command.as_ref();
                        let _ = bell_command.map(|cmd| start_program(cmd.program(), cmd.args()));
```

alacritty_terminal/src/config/bell.rs
```diff
@@ -0,0 +1,122 @@
...
+            );
+
+            Ok(DEFAULT_BELL_COLOR)
+        },
+    }
+}
+
+fn deserialize_bell_command<'a, D>(
+    deserializer: D,
+) -> std::result::Result<Option<Program>, D::Error>
```
chrisduerr, line 71:
```suggestion
) -> Result<Option<Program>, D::Error>
```

alacritty/src/event.rs
kchibisov: I'm not sure that we should do that in `start_daemon`. It does the one thing, and have a special impl on each platform, so I'm not sure if there's any value in moving it there. Hiding errors from library users isn't something I really like tbqh.

chrisduerr: Could you move it out of `alacritty_terminal` then maybe? There's no reason for it to be there.

alacritty/src/daemon.rs
```diff
@@ -1,77 +1,64 @@
...
-where
-    I: IntoIterator<Item = S>,
-    S: AsRef<OsStr>,
-{
     // Setting all the I/O handles to null and setting the
     // CREATE_NEW_PROCESS_GROUP and CREATE_NO_WINDOW has the effect
     // that console applications will run without opening a new
     // console window.
-    Command::new(program)
+    #[cfg(windows)]
```
chrisduerr, line 50: Feels like it would be easier to have these as separate methods, even if they're not exposed publicly? Like this the annotations are quite complicated to keep track of since they're relatively big.

kchibisov: ok.

alacritty/src/daemon.rs
```diff
@@ -1,34 +1,53 @@
...
+/// Start the daemon and log error on failure.
+pub fn start_daemon<I, S>(program: &str, args: I)
+where
+    I: IntoIterator<Item = S> + Debug + Copy,
+    S: AsRef<OsStr>,
+{
+    match start_daemon_impl(program, args) {
+        Ok(_) => debug!("Launched {} with args {:?}", program, args),
+        Err(_) => warn!("Unable to launch {} with args {:?}", program, args),
+    };
```
chrisduerr, line 24:
```suggestion
    }
```

alacritty/src/daemon.rs
```diff
@@ -1,34 +1,53 @@
...
+        .spawn()
+        .map(|_| ())
 }
 
 #[cfg(not(windows))]
-pub fn start_daemon<I, S>(program: &str, args: I) -> io::Result<()>
+fn start_daemon_impl<I, S>(program: &str, args: I) -> io::Result<()>
 where
-    I: IntoIterator<Item = S>,
+    I: IntoIterator<Item = S> + Copy,
```
chrisduerr, line 50: Why did you add the `Copy` constraint?

alacritty/src/daemon.rs
```diff
@@ -1,34 +1,53 @@
...
+{
+    match start_daemon_impl(program, args) {
+        Ok(_) => debug!("Launched {} with args {:?}", program, args),
+        Err(_) => warn!("Unable to launch {} with args {:?}", program, args),
+    };
+}
 
-    pub use std::thread::*;
+#[cfg(windows)]
+fn start_daemon_impl<I, S>(program: &str, args: I) -> io::Result<()>
```
chrisduerr, line 28:
```suggestion
fn spawn_daemon<I, S>(program: &str, args: I) -> io::Result<()>
```

I think adding `_impl` is a bit strange. A similar but not identical name should do.

alacritty_terminal/src/config/bell.rs
chrisduerr: +1

alacritty/src/daemon.rs
kchibisov: Because you should copy a reference? Otherwise it's not working, since you should print it as well and to pass it somewhere else?

PR #3852, opened 2020-06-15
Update TERMINFO in INSTALL.md

btannous:

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -329,5 +329,15 @@ instead.
 To install alacritty's terminfo entry globally:
 
 ```sh
-sudo tic -xe alacritty,alacritty-direct extra/alacritty.info
+curl -LO https://raw.githubusercontent.com/alacritty/alacritty/master/extra/alacritty.info
+sudo tic -xe alacritty,alacritty-direct alacritty.info
+rm alacritty.info
+```
+
+To install alacritty's terminfo locally:
+
+```sh
+curl -LO https://raw.githubusercontent.com/alacritty/alacritty/master/extra/alacritty.info
+tic -xe alacritty,alacritty-direct alacritty.info
+rm alacritty.info
 ```


chrisduerr:
This is not the recommended way to install the terminfo. You should follow the instructions completely including the git clone and building the specific version from source.

btannous:
would it be better to look at the releases page, and grab the right alacritty.info from there that matches the version you are using?

https://github.com/alacritty/alacritty/releases

btannous:
it doesn’t look like the file gets updated too often. 

https://github.com/alacritty/alacritty/commits/master/extra/alacritty.info

i’m just thinking it would be nice to make it easier to use when sshing into a server that doesn’t know what alacritty is and you have messed up tab completion and arrow key movement and other thing that happen when TERM is unknown. 

sure it could be easy to do a export TERM=xterm

but those servers might not have git installed at that point.

chrisduerr:
> would it be better to look at the releases page, and grab the right alacritty.info from there that matches the version you are using?

Yes, that's the correct process. When building from source you'll already have the correct version in the directory though, which is what the INSTALL.md is for.

> it doesn’t look like the file gets updated too often.

Not getting updated often doesn't mean it doesn't get updated. There's no reason to recommend something that might cause serious troubles for users.

> i’m just thinking it would be nice to make it easier to use when sshing into a server that doesn’t know what alacritty is

That's a very specific usecase that is not what the INSTALL.md is for. There are also a couple of different solutions, I'd personally probably recommend just copying your local terminfo to the remote. That way it's actually the correct one and you can still easily do that in one command. That should work unless you ssh into a ton of different servers, in which case aliasing ssh to `TERM=xterm-256color ssh` is probably your only option.

btannous:
Good points @chrisduerr !

I didn't think about the copying from local to remote! That sounds like the best method.  I mean INSTALL.md is for building, I get it.

A search of `alacritty terminfo` has the INSTALL.md as a top hit, along with many other issues in the repo that talk about terminfo (maybe because others are running into this issue, sshing into a server w/o alacritty's terminfo installed).

Maybe you would be cool w/ me adding in a TERMINFO.md page to the repo (instead of this) with these suggestions (or one of them based on your guidance?)
- sftp terminfo local->remote -> tic ...
- grabbing alacritty version -> github releases terminfo -> tic...
- and/or suggesting the alias you pointed out (if working on many different servers)

If not, that's cool too, I think the info is here now for people - and I'm set on a good practice now.  Just trying to think of how to be helpful to the community if they are using alacritty and are not too familiar with a lack of a specific terminfo causing their terminal to act odd.

chrisduerr:
> Maybe you would be cool w/ me adding in a TERMINFO.md page to the repo (instead of this) with these suggestions (or one of them based on your guidance?)

If you're interested in helping out with improving documentation around this, which would be much appreciated, I think the best place would be our FAQ section in the README. It's definitely a frequently asked question so that seems fitting.

Alternatively a wiki entry would also be a good place. That's generally the recommended place for more of an exchange of information between users. Since the issue is not Alacritty specific and the solution are all just commonly known and used tools, that would be a better place.

Of course a combination would also work, if the documentation gets too long for the FAQ, we can just add a link to the wiki from the FAQ.

PR #3854, opened 2020-06-15
Clear selection on clear line/screen escapes

kchibisov:
Selection is now cleared if clear line or clear screen escape
sequences are clearing content behind it. In addition, selection
won't be cleared on resize if the amount of columns remain the same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Set IUTF8 termios flag for improved UTF8 input support
 - Dragging files into terminal now adds a space after each path
 - Default binding replacement conditions
+- Adjusted selection clearing granularity to more accurately match content
 
 ### Fixed
 
--- a/alacritty_terminal/src/selection.rs
+++ b/alacritty_terminal/src/selection.rs
@@ -195,12 +195,12 @@ impl Selection {
 
     /// Check whether selection contains any point in a given range.
     pub fn intersects_range(&self, range: RangeInclusive<usize>) -> bool {
-        let (mut start, mut end) = (self.region.start, self.region.end);
-        if Self::points_need_swap(start.point, end.point) {
+        let mut start = self.region.start.point.line;
+        let mut end = self.region.end.point.line;
+
+        if Self::points_need_swap(self.region.start.point, self.region.end.point) {
             mem::swap(&mut start, &mut end);
         }
-
-        let (start, end) = (start.point.line, end.point.line);
 
         let (range_start, range_end) = range.into_inner();
 
@@ -233,7 +233,9 @@ impl Selection {
         let num_cols = grid.num_cols();
 
         // Order start above the end.
-        let (mut start, mut end) = (self.region.start, self.region.end);
+        let mut start = self.region.start;
+        let mut end = self.region.end;
+
         if Self::points_need_swap(start.point, end.point) {
             mem::swap(&mut start, &mut end);
         }
--- a/alacritty_terminal/src/selection.rs
+++ b/alacritty_terminal/src/selection.rs
@@ -204,7 +204,7 @@ impl Selection {
 
         let (range_start, range_end) = range.into_inner();
 
-        range.start <= selection.start && range.end >= selection.end
+        range_start <= start && range_end >= end
     }
 
     /// Expand selection sides to include all cells.
@@ -658,5 +658,19 @@ mod tests {
             end: Point::new(6, Column(3)),
             is_block: true,
         });
+    }
+
+    #[test]
+    fn range_intersection() {
+        let mut selection =
+            Selection::new(SelectionType::Lines, Point::new(6, Column(1)), Side::Left);
+        selection.update(Point::new(3, Column(1)), Side::Right);
+
+        assert!(selection.intersects_range(4..=5));
+        assert!(selection.intersects_range(5..=7));
+        assert!(selection.intersects_range(2..=4));
+        assert!(selection.intersects_range(2..=7));
+        assert!(!selection.intersects_range(7..=8));
+        assert!(!selection.intersects_range(1..=2));
     }
 }
--- a/alacritty_terminal/src/selection.rs
+++ b/alacritty_terminal/src/selection.rs
@@ -7,7 +7,7 @@
 
 use std::convert::TryFrom;
 use std::mem;
-use std::ops::Range;
+use std::ops::{Range, RangeInclusive};
 
 use crate::index::{Column, Line, Point, Side};
 use crate::term::{Search, Term};
@@ -191,6 +191,22 @@ impl Selection {
             },
             SelectionType::Semantic | SelectionType::Lines => false,
         }
+    }
+
+    /// Check whether selection contains any point in a given range.
+    pub fn intersects_range(&self, range: RangeInclusive<usize>) -> bool {
+        let (mut start, mut end) = (self.region.start, self.region.end);
+        if Self::points_need_swap(start.point, end.point) {
+            mem::swap(&mut start, &mut end);
+        }
+
+        let (start, end) = (start.point.line, end.point.line);
+
+        let (range_start, range_end) = range.into_inner();
+
+        (start >= range_end && end <= range_end)
+            || (start >= range_start && end <= range_start)
+            || (start >= range_start && start <= range_end)
     }
 
     /// Expand selection sides to include all cells.
--- a/alacritty_terminal/src/selection.rs
+++ b/alacritty_terminal/src/selection.rs
@@ -7,7 +7,7 @@
 
 use std::convert::TryFrom;
 use std::mem;
-use std::ops::{Range, RangeInclusive};
+use std::ops::{Bound, Range, RangeBounds};
 
 use crate::index::{Column, Line, Point, Side};
 use crate::term::{Search, Term};
@@ -194,7 +194,7 @@ impl Selection {
     }
 
     /// Check whether selection contains any point in a given range.
-    pub fn intersects_range(&self, range: RangeInclusive<usize>) -> bool {
+    pub fn intersects_range<R: RangeBounds<usize>>(&self, range: R) -> bool {
         let mut start = self.region.start.point.line;
         let mut end = self.region.end.point.line;
 
@@ -202,7 +202,17 @@ impl Selection {
             mem::swap(&mut start, &mut end);
         }
 
-        let (range_start, range_end) = range.into_inner();
+        let range_start = match range.start_bound() {
+            Bound::Included(&range_start) => range_start,
+            Bound::Excluded(&range_start) => range_start.saturating_add(1),
+            Bound::Unbounded => 0,
+        };
+
+        let range_end = match range.end_bound() {
+            Bound::Included(&range_end) => range_end,
+            Bound::Excluded(&range_end) => range_end.saturating_sub(1),
+            Bound::Unbounded => usize::max_value(),
+        };
 
         range_start <= start && range_end >= end
     }
@@ -666,11 +676,14 @@ mod tests {
             Selection::new(SelectionType::Lines, Point::new(6, Column(1)), Side::Left);
         selection.update(Point::new(3, Column(1)), Side::Right);
 
-        assert!(selection.intersects_range(4..=5));
-        assert!(selection.intersects_range(5..=7));
+        assert!(selection.intersects_range(..));
+        assert!(selection.intersects_range(2..));
         assert!(selection.intersects_range(2..=4));
         assert!(selection.intersects_range(2..=7));
+        assert!(selection.intersects_range(4..=5));
+        assert!(selection.intersects_range(5..8));
+
+        assert!(!selection.intersects_range(..=2));
         assert!(!selection.intersects_range(7..=8));
-        assert!(!selection.intersects_range(1..=2));
     }
 }
--- a/alacritty_terminal/src/selection.rs
+++ b/alacritty_terminal/src/selection.rs
@@ -204,9 +204,7 @@ impl Selection {
 
         let (range_start, range_end) = range.into_inner();
 
-        (start >= range_end && end <= range_end)
-            || (start >= range_start && end <= range_start)
-            || (start >= range_start && start <= range_end)
+        range.start <= selection.start && range.end >= selection.end
     }
 
     /// Expand selection sides to include all cells.
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1050,16 +1050,12 @@ impl<T> Term<T> {
         self.vi_mode_cursor.point.col = min(self.vi_mode_cursor.point.col, num_cols - 1);
         self.vi_mode_cursor.point.line = min(self.vi_mode_cursor.point.line, num_lines - 1);
 
-        // Reset scrolling region.
+        // Recreate tabs list.
+        self.tabs.resize(self.grid.num_cols());
+
+        // Reset scrolling region and selection.
         self.scroll_region = Line(0)..self.grid.num_lines();
-
-        // Invalidate selection and tabs only when necessary.
-        if old_cols != num_cols {
-            self.selection = None;
-
-            // Recreate tabs list.
-            self.tabs.resize(self.grid.num_cols());
-        }
+        self.selection = None;
     }
 
     #[inline]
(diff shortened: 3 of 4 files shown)


kchibisov:
The only missing part here are tests, which I'll add in a meantime, but the code should behave in +- correct way.

kchibisov:
Ok, I was a bit optimistic about working here.

The issue I have right now, is that `SelectionRange` is by default in buffer coordinates, however `contains` is in visible coordinates (which confuses me a bunch of times). However our iteration is in visible coordinates, so that's why contains is done that way.

So to clear selection properly I need either one more `contains` function on `SelectionRange` or I need a selection in visible coordinates, so basically adding `to_visible_range` on selection with impl from https://github.com/alacritty/alacritty/blob/d526649ee67237a1bedad8228f1f017035ddad3f/alacritty_terminal/src/term/mod.rs#L224 , however it's a bit more work for conversion(not sure if it matters though). But in general handling should be a bit simpler, since `clear_screen` and `clear_line` are in visible already.

What do you think @chrisduerr  ?

chrisduerr:
Can't you just convert the point you're trying to check for to viewport coordinates? Clearing is always performed relative to the viewport anyways, right?

kchibisov:
I've updated PR, so it now has 2 commits (the first one is impl of clear behavior). The second is caching for selection.to_range conversion. I don't really like the second commit here, but first should +- work and clear selection properly.

alacritty_terminal/src/selection.rs
```diff
@@ -50,6 +50,36 @@ impl<L> SelectionRange<L> {
             && (self.start.col <= col || (self.start.line != line && !self.is_block))
             && (self.end.col >= col || (self.end.line != line && !self.is_block))
     }
+
+    pub fn is_above_buffer_line(&self, line: L, num_lines: L) -> bool
+    where
+        L: PartialEq + PartialOrd,
+    {
+        self.start.line > line || (self.end.line > line && self.end.line < num_lines)
```
chrisduerr, line 58: Why does it matter if self.end.line is smaller than num_lines?

alacritty_terminal/src/selection.rs
```diff
@@ -50,6 +50,36 @@ impl<L> SelectionRange<L> {
...
+
+    pub fn contains_cols_on_buffer_line(&self, line: L, cols: Range<Column>) -> bool
+    where
+        L: PartialEq + PartialOrd + Clone + Copy,
+    {
+        for col in IndexRange::from(cols.start..cols.end) {
+            if self.contains_buffer_point(col, line) {
+                return true;
+            }
+        }
```
chrisduerr, line 69: That just looks like the `any` method on iterator?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1743,25 +1745,42 @@ impl<T: EventListener> Handler for Term<T> {
...
+        let num_cols = self.grid.num_cols();
+        let selection_range = self.selection.as_ref().and_then(|s| s.to_range(&self));
+        let buffer_cursor = self.grid.buffer_cursor();
+
+        let cols = match mode {
             ansi::LineClearMode::Right => {
                 let row = &mut self.grid[cursor.point.line];
-                for cell in &mut row[cursor.point.col..] {
+                let cols = cursor.point.col..num_cols;
+                for cell in &mut row[cols.clone()] {
```
chrisduerr, line 1788: From here on out all branches basically do the same thing. So if we need to compute the cols anyways and return them, we can probably extract everything other than just the cols computation to after the match?

If `contains_cols_on_buffer_line` takes a reference to a range, we should also be able to get rid of the `clone`s? Even though the clones are extremely cheap of course.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1838,45 +1857,86 @@ impl<T: EventListener> Handler for Term<T> {
...
+                    if let Some(true) = selection_range.as_ref().map(|range| {
+                        !range.is_above_buffer_line(buffer_cursor.line - 1, num_lines.0)
+                    }) {
+                        self.selection = None;
+                    }
                 }
             },
             ansi::ClearMode::All => {
+                // Clearing all, so should clear selection as well.
+                self.selection = None;
```
chrisduerr: Should we? Selection doesn't need to be visible.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1838,45 +1857,86 @@ impl<T: EventListener> Handler for Term<T> {
...
                     self.grid.region_mut(..).each(|c| c.reset(&template));
                 } else {
                     let template = Cell { bg: template.bg, ..Cell::default() };
                     self.grid.clear_viewport(template);
                 }
             },
-            ansi::ClearMode::Saved => self.grid.clear_history(),
+            ansi::ClearMode::Saved => {
+                // Clear all, since we're clearing history already.
+                self.selection = None;
```
chrisduerr: Should we? We aren't necessarily clearing the viewport.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1838,45 +1857,86 @@ impl<T: EventListener> Handler for Term<T> {
...
+        let template = self.grid.cursor.template;
+        let num_cols = self.grid.num_cols();
+        let num_lines = self.grid.num_lines();
+        let cursor = self.grid.cursor.point;
+        let buffer_cursor = self.grid.buffer_cursor();
 
         match mode {
             ansi::ClearMode::Above => {
-                let cursor = self.grid.cursor.point;
+                let selection_range = self.selection.as_ref().and_then(|s| s.to_range(&self));
```
chrisduerr: If we accept to only do line-based granularity instead of cell-based, I think we should be fine with just using the selection without expansion to range?

Since the expansion is done on rendering anyways, it will just expand to the newly written/cleared content which doesn't seem that incorrect to me compared to what we currently do.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Set IUTF8 termios flag for improved UTF8 input support
 - Dragging files into terminal now adds a space after each path
 - Default binding replacement conditions
+- Selection is now cleared if clear line or clear screen escapes are clearing the content behind it
```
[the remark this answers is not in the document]
kchibisov: That's actually not entirely true, the original problem was that selection was persistent if you clear_line behind it, so the wording could be a bit hard, since in one case it was always cleared, but in the other it was never cleared, while now it's cleared in both cases, if something was changed behind it.

alacritty_terminal/src/selection.rs
kchibisov: To not clear selection which is out side of viewport, since you're technically not touching it?

kchibisov:
I've changed granularity to just lines and also now it works on non-expanded selection. Let me know what you think.

alacritty_terminal/src/selection.rs
```diff
@@ -193,6 +193,22 @@ impl Selection {
...
+            mem::swap(&mut start, &mut end);
+        }
+
+        let (start, end) = (start.point.line, end.point.line);
+
+        let (range_start, range_end) = range.into_inner();
+
+        (start >= range_end && end <= range_end)
+            || (start >= range_start && end <= range_start)
+            || (start >= range_start && start <= range_end)
```
chrisduerr, line 209: I'm gonna sound a bit silly if this is not true, but wouldn't the following condition be sufficient:

```
range.start >= selection.end && range.end <= selection.start
```

My thinking here is that we know for a fact that our selection is well structured. So we know the start is above the end when we swap points. So by knowing that the range end is below the range start, and knowing that the range start is above the selection end and the range end is below the selection start, we know that they must intersect?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1763,6 +1767,12 @@ impl<T: EventListener> Handler for Term<T> {
                 }
             },
         }
+
+        let buffer_cursor_line = self.grid.buffer_cursor_point().line;
+        self.selection = self
+            .selection
+            .take()
+            .filter(|s| !s.intersects_range(buffer_cursor_line..=buffer_cursor_line));
```
chrisduerr, line 1775: In theory an `intersects_line` function like you had previously would make this only a single line.

My main problem was with duplication of code, do you think it's worth it to have a method that basically just calls `self.intersects_range(line..=line)` internally?

Formatting is maybe not the best reason to do so, just an idea.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1840,8 +1850,8 @@ impl<T: EventListener> Handler for Term<T> {
         trace!("Clearing screen: {:?}", mode);
         let template = self.grid.cursor.template;
 
-        // Remove active selections.
-        self.selection = None;
+        let num_lines = self.grid.num_lines().0 - 1;
```
chrisduerr, line 1853:
```suggestion
        let last_line = self.num_lines().0 - 1;
```

We shouldn't name things `num_lines` when we subtract one to represent the last line.

Alternatively `last_visible_line` would be even more clear.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1875,8 +1894,16 @@ impl<T: EventListener> Handler for Term<T> {
...
                 }
+
+                self.selection =
+                    self.selection.take().filter(|s| !s.intersects_range(0..=num_lines));
+            },
+            ansi::ClearMode::Saved => {
+                self.grid.clear_history();
+
+                self.selection =
+                    self.selection.take().filter(|s| s.intersects_range(0..=num_lines));
```
chrisduerr, line 1905: This isn't correct, right? It should instead be `num_lines + 1 ..= grid.len() - 1`?

kchibisov: it's indeed more correct to use what you've wrote + inverse the condition.

alacritty_terminal/src/selection.rs
kchibisov: I don't get what are you talking about here. You have 3 patterns

- selection starts above, ends below range_end
- selection starts below and ends above range_start
- selection between range start and end

I'm not sure that i get your suggestion here, since it's not correct for case.

`range.start` is below `selection.end` and `selection.start` is above `range.end`(range inside selection).

alacritty_terminal/src/term/mod.rs
kchibisov: maybe, but I don't see how it could improve anything, since 1..=1 is still a valid range, so I don't see any issue. So, I'd prefer to keep it that way.

alacritty_terminal/src/selection.rs
chrisduerr: I think my start/end are swapped around. In my example the start is the top of the range and selection and the end is the bottom of the range and selection.

For the case you've mentioned, it would work if that's the case. You'd just have to swap them around to match that.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1875,8 +1894,20 @@ impl<T: EventListener> Handler for Term<T> {
                     let template = Cell { bg: template.bg, ..Cell::default() };
                     self.grid.clear_viewport(template);
                 }
+
+                self.selection =
+                    self.selection.take().filter(|s| !s.intersects_range(0..=last_line));
+            },
+            ansi::ClearMode::Saved => {
+                let last_scrollback_line = max(self.grid.len() - 1, last_line + 1);
```
chrisduerr, line 1902: Why are you `max`ing things here? Just in case there is no history?

Can we maybe just add a conditional to check that and skip this branch completely in that case? That way we don't have to do any expensive clearing operations at all.

alacritty_terminal/src/selection.rs
```diff
@@ -193,6 +193,20 @@ impl Selection {
...
+        let mut start = self.region.start.point.line;
+        let mut end = self.region.end.point.line;
+
+        if Self::points_need_swap(self.region.start.point, self.region.end.point) {
+            mem::swap(&mut start, &mut end);
+        }
+
+        let (range_start, range_end) = range.into_inner();
+
+        range.start <= selection.end && range.end >= selection.start
```
chrisduerr, line 207:
```suggestion
        range_end >= start && range_start <= end
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -1875,8 +1890,22 @@ impl<T: EventListener> Handler for Term<T> {
...
+                let last_scrollback_line = self.grid.len() - 1;
+
+                self.grid.clear_history();
+
+                if last_scrollback_line > last_line {
+                    self.selection = self
+                        .selection
+                        .take()
+                        .filter(|s| !s.intersects_range(last_line + 1..=last_scrollback_line));
+                }
```
chrisduerr, line 1907: Why the conditional here instead of the match arm? We can completely skip this entire thing if there is no history, right?

kchibisov: >Why the conditional here instead of the match arm? We can completely skip this entire thing if there is no history, right?

RIght, but then you have non exhaustive match, which looks kind of redundant and could be confusing and error prone sometimes(since you'll basically create a rule `_ => ()`). If anything we can guard in arm at the start.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1875,8 +1890,23 @@ impl<T: EventListener> Handler for Term<T> {
...
                 }
+
+                self.selection =
+                    self.selection.take().filter(|s| !s.intersects_range(0..=last_line));
+            },
+            ansi::ClearMode::Saved => {
+                let last_scrollback_line = self.grid.len() - 1;
+                if last_scrollback_line == last_line {
+                    return;
+                }
```
chrisduerr, line 1901: Imo the early return here is much, much more error prone and complex than having a `_ => ()` match arm.

You can also just do `ClearMode::Saved => ()`, if you're concerned about adding new variants.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1894,19 +1894,16 @@ impl<T: EventListener> Handler for Term<T> {
                 self.selection =
                     self.selection.take().filter(|s| !s.intersects_range(0..=last_line));
             },
-            ansi::ClearMode::Saved => {
-                let last_scrollback_line = self.grid.len() - 1;
-                if last_scrollback_line == last_line {
-                    return;
-                }
-
+            ansi::ClearMode::Saved if self.grid.len() - 1 > last_line => {
```
chrisduerr, line 1897: Why this cryptic check rather than `self.grid.history_size() > 0`? That seems to be more clear on intent.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1840,8 +1846,8 @@ impl<T: EventListener> Handler for Term<T> {
         trace!("Clearing screen: {:?}", mode);
         let template = self.grid.cursor.template;
 
-        // Remove active selections.
-        self.selection = None;
+        let last_line = self.grid.num_lines().0 - 1;
```
chrisduerr, line 1849: I'm really conflicted about this variable, can we just go with num_lines here and add the `- 1` wherever it's needed.

That way we don't need to do silly things like `last_line + 1` either, which just evaluates to `num_lines - 1 + 1`.

I've thought about just proposing `last_visible_line`, but even then the indexing direction doesn't make it clear if this is just always `0` or not.

Once that's done this is good to go for me.

kchibisov: >I'm really conflicted about this variable, can we just go with num_lines here and add the - 1 wherever it's needed.

It's needed everywhere in this function except one place, that's why it's +1 on when needed, since the places where `+1` is required is basically one, compared to like 5-6 where `-1` is needed. If it was the other way around then sure.

>That way we don't need to do silly things like last_line + 1 either, which just evaluates to num_lines - 1 + 1

But then you have `-1` all over the place.

>I've thought about just proposing last_visible_line, but even then the indexing direction doesn't make it clear if this is just always 0 or not.

It's technically not last visible line, it's last line where cursor can move to, but to name it that way is hard.

chrisduerr: "Everywhere" is a strong word for three instances of using this versus one of using num_lines. 5-6 is a **strong** exaggeration, that's basically double of the reality. To put it in other words: In 25% of cases you're doing something that makes no sense at all.

> But then you have -1 all over the place.

So? num_lines - 1 is extremely clear on what it means in Alacritty. last_line or last_line + 1 is not at all clear in any way whatsoever.

> It's technically not last visible line, it's last line where cursor can move to, but to name it that way is hard.

It's the last visible line index.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1891,8 +1892,11 @@ impl<T: EventListener> Handler for Term<T> {
                     self.grid.clear_viewport(template);
                 }
 
-                self.selection =
-                    self.selection.take().filter(|s| !s.intersects_range(0..=last_line));
+                let last_visible_line_index = num_lines - 1;
```
chrisduerr, line 1895: The variable name and definition is longer than the variable itself. What's the point?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1865,18 +1865,19 @@ impl<T: EventListener> Handler for Term<T> {
                     cell.reset(&template);
                 }
 
+                let last_visible_line_index = num_lines - 1;
```
chrisduerr, line 1868: The variable name and definition is longer than the variable itself. What's the point?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1875,8 +1891,24 @@ impl<T: EventListener> Handler for Term<T> {
                     let template = Cell { bg: template.bg, ..Cell::default() };
                     self.grid.clear_viewport(template);
                 }
+
+                let last_visible_line_index = num_lines - 1;
```
kchibisov, line 1895: You can't inline those, since `clippy` bugs.

chrisduerr: It's not a bug. It's just that you probably should implement the `intersects_range` method properly.

PR #3856, opened 2020-06-16
Add automatic scrolling during selection

### LINKED ISSUE #1173: When selecting to the top of the screen, alacritty should scroll
When the user makes a selection and the mouse reaches the top of the screen, the terminal should automatically scroll up to allow selecting lines outside of the current viewport.

chrisduerr:
Fixes #1173.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -30,7 +30,7 @@ use font::{self, Size};
 use alacritty_terminal::config::Font;
 use alacritty_terminal::config::LOG_TARGET_CONFIG;
 use alacritty_terminal::event::OnResize;
-use alacritty_terminal::event::{Event, EventListener, Notify};
+use alacritty_terminal::event::{Event as TerminalEvent, EventListener, Notify};
 use alacritty_terminal::grid::Scroll;
 use alacritty_terminal::index::{Column, Line, Point, Side};
 use alacritty_terminal::message_bar::{Message, MessageBuffer};
@@ -40,7 +40,7 @@ use alacritty_terminal::term::cell::Cell;
 use alacritty_terminal::term::{ClipboardType, SizeInfo, Term, TermMode};
 #[cfg(not(windows))]
 use alacritty_terminal::tty;
-use alacritty_terminal::util::{limit, start_daemon};
+use alacritty_terminal::util::start_daemon;
 
 use crate::cli::Options;
 use crate::clipboard::Clipboard;
@@ -48,8 +48,31 @@ use crate::config;
 use crate::config::Config;
 use crate::display::Display;
 use crate::input::{self, ActionContext as _, FONT_SIZE_STEP};
+use crate::scheduler::Scheduler;
 use crate::url::{Url, Urls};
 use crate::window::Window;
+
+/// Events dispatched through the UI event loop.
+#[derive(Debug, Clone)]
+pub enum Event {
+    TerminalEvent(TerminalEvent),
+    DPRChanged(f64, (u32, u32)),
+    Scroll(Scroll),
+    ConfigReload(PathBuf),
+    Message(Message),
+}
+
+impl From<Event> for GlutinEvent<'_, Event> {
+    fn from(event: Event) -> Self {
+        GlutinEvent::UserEvent(event)
+    }
+}
+
+impl From<TerminalEvent> for Event {
+    fn from(event: TerminalEvent) -> Self {
+        Event::TerminalEvent(event)
+    }
+}
 
 #[derive(Default, Clone, Debug, PartialEq)]
 pub struct DisplayUpdate {
@@ -80,6 +103,7 @@ pub struct ActionContext<'a, N, T> {
     pub config: &'a mut Config,
     pub event_loop: &'a EventLoopWindowTarget<Event>,
     pub urls: &'a Urls,
+    pub scheduler: &'a mut Scheduler,
     font_size: &'a mut Size,
 }
 
@@ -248,6 +272,25 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         }
     }
 
+    /// Spawn URL launcher when clicking on URLs.
+    fn launch_url(&self, url: Url) {
+        if self.mouse.block_url_launcher {
+            return;
+        }
+
+        if let Some(ref launcher) = self.config.ui_config.mouse.url.launcher {
+            let mut args = launcher.args().to_vec();
+            let start = self.terminal.visible_to_buffer(url.start());
+            let end = self.terminal.visible_to_buffer(url.end());
+            args.push(self.terminal.bounds_to_string(start, end));
+
+            match start_daemon(launcher.program(), &args) {
+                Ok(_) => debug!("Launched {} with args {:?}", launcher.program(), args),
+                Err(_) => warn!("Unable to launch {} with args {:?}", launcher.program(), args),
+            }
+        }
+    }
+
     fn change_font_size(&mut self, delta: f32) {
         *self.font_size = max(*self.font_size + delta, Size::new(FONT_SIZE_STEP));
         let font = self.config.font.clone().with_size(*self.font_size);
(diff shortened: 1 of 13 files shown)


alacritty/src/input.rs
```diff
@@ -33,13 +34,23 @@ use alacritty_terminal::vi_mode::ViMotion;
...
+use crate::event::{ClickState, Event, Mouse};
+use crate::scheduler::{Scheduler, SELECTION_SCROLLING_EVENT};
 use crate::url::{Url, Urls};
 use crate::window::Window;
 
 /// Font size change interval.
 pub const FONT_SIZE_STEP: f32 = 0.5;
 
+/// Interval for mouse scrolling during selection outside of the boundaries.
+const SELECTION_SCROLLING_INTERVAL: Duration = Duration::from_millis(25);
```
kchibisov, line 46: Could we increase tick rate, otherwise it feel really slow compared to what it was before. Especially when in fullscreen or when you don't really have gaps on bottom.

alacritty/src/event.rs
```diff
@@ -286,23 +329,8 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
-            let start = self.terminal.visible_to_buffer(url.start());
-            let end = self.terminal.visible_to_buffer(url.end());
-            args.push(self.terminal.bounds_to_string(start, end));
-
-            match start_daemon(launcher.program(), &args) {
-                Ok(_) => debug!("Launched {} with args {:?}", launcher.program(), args),
-                Err(_) => warn!("Unable to launch {} with args {:?}", launcher.program(), args),
-            }
-        }
+    fn scheduler(&mut self) -> &mut Scheduler {
```
kchibisov, line 332: Could you rename it to `scheduler_mut`? you can also update `clipboard` to `clipboard_mut` as well.

alacritty/src/input.rs
```diff
@@ -33,13 +34,23 @@ use alacritty_terminal::vi_mode::ViMotion;
...
+
+/// Minimum number of pixels at the bottom/top where selection scrolling is performed.
+const MIN_SELECTION_SCROLLING_HEIGHT: i32 = 5;
```
kchibisov, line 49: It's not DPI scaled so the amount could actually be very low on HiDPI monitors.

alacritty/src/input.rs
```diff
@@ -33,13 +34,23 @@ use alacritty_terminal::vi_mode::ViMotion;
...
+
+/// Number of pixels for increasing the selection scrolling speed factor by one.
+const SELECTION_SCROLLING_STEP: i32 = 25;
```
kchibisov, line 52: I'd lower this value tbqh, also, it could make sense to scale it with DPI as well, but I'm not sure about that.

alacritty/src/scheduler.rs
```diff
@@ -0,0 +1,114 @@
...
+
+use std::collections::VecDeque;
+use std::time::{Duration, Instant};
+
+use glutin::event::Event as GlutinEvent;
+
+use crate::event::Event as AlacrittyEvent;
+
+/// Event ID for scrolling during selection.
+pub const SELECTION_SCROLLING_EVENT: u64 = 0;
```
kchibisov, line 11: It's not event, it's a timer id. So you have some predefined timer ids and arbitrary timer. But if it's a predefined timer, how one could handle multiwindow? I'd assume no problem for selection, but some other things won't work. Anyway, the way it works right now kind of makes sense, however now I wonder what's the idea of `next_id` on `Scheduler` if we can just have a set of predefined timers, and so you never touch `next_id` anyway.

Also, we all is about predefined times, then it should be enum, I guess.

alacritty/src/event.rs
chrisduerr: I'm not sure if that's necessary if there is just a single way to access it?

alacritty/src/input.rs
chrisduerr: I think 25 is a good value to prevent it from accidentally just shooting the user way up in history. Especially because it's sometimes harder to get back since you might have more space on one side than the other.

alacritty/src/scheduler.rs
chrisduerr: Good point, since we do not need next_id yet, I'll convert it to an enum. If we decide that we need it later, we can always just change it.

alacritty/src/input.rs
chrisduerr: I think the current tick rate is a great minimum value, otherwise accurately selecting stuff becomes a hassle.

The primary goal isn't to move through the scrollback as quickly as possible.

alacritty/src/input.rs
chrisduerr: Seems like the easiest thing is just to scale it by DPI.

alacritty/src/event.rs
kchibisov: The standard library does that even where there's one method with name X, it names it X_mut https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut (it doesn't have get, just get_mut).  Also, it increases the clarity and a bit more consistent with nameing that we have. And you may want to add normal getters at some point as well without changing the API.

P.s. you even used `get_mut` over `get` in scheduler API as well.

alacritty/src/input.rs
kchibisov: Ehm, I don't really like it because it's very slow and I think doing a bit faster is reasonable. I'd say  `12` ms is what it should be, so the selection will be a bit smoother as well, while the rate is still normal.

>accurately selecting stuff becomes a hassle.

Or you're waiting ages to scroll. You're fine to re select a bit, since you're not scrolling that fast with `12ms`, and `25ms` doesn't protect you from inaccuracy, since you always select something that was invisible a frame before, so the chance to reselect even with 25ms is high already.

alacritty/src/input.rs
```diff
@@ -870,32 +891,61 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
+            end_top - mouse_y + step
+        } else if mouse_y >= start_bottom {
+            start_bottom - mouse_y - step
+        } else {
+            scheduler.unschedule(TimerId::SelectionScrolling);
+            return;
+        };
+
+        // Scale number of lines scrolled based on distance to boundary.
+        let delta = delta as isize / SELECTION_SCROLLING_STEP as isize;
```
kchibisov, line 920: isn't it `delta / step` or it's intentional that way? If it's intentional, leave a comment why.

alacritty/src/event.rs
chrisduerr: Get is different though, especially because `get` is always immutable and `get_mut` is always mutable. This method isn't called `get`, it's called `scheduler`.

Fact is that every single instance on this trait that only provides mutable access drops the _mut suffix. So I'm just following the existing style not introducing something new here.

alacritty/src/input.rs
chrisduerr: > Ehm, I don't really like it because it's very slow and I think doing a bit faster is reasonable. I'd say 12 ms is what it should be, so the selection will be a bit smoother as well, while the rate is still normal.

12 ms is far too quick for me, it's also higher than most monitor's refresh rate, so then the speed will depend on your monitor which isn't a great idea.

If your goal is to cover a huge distance as quickly as possible, use a feature that doesn't have the goal to still accurately preserve your mouse pointer position or make sure there's plenty of space around your window.

alacritty/src/input.rs
chrisduerr: Correct, thanks.

alacritty/src/input.rs
chrisduerr: I've slightly tightened up the values to 20 - 20. Maybe that's a good middle ground? Let me know what you think.

nawgz:
This is a heroic fix. The lack of this feature was preventing me from using this, but now that it works I am overjoyed to have the snappiest terminal. Cheers to you and all the other contributors

PR #3857, opened 2020-06-17
Escape Windows shell command line arguments

### LINKED ISSUE #3552: Improve quoted shell parameter handling on Windows
When launching Alacritty with a custom command via `alacritty.exe -e`, it seems that Alacritty does not preserve the quoting of its arguments. E.g.,

```
alacritty.exe -e wsl zsh -c 'cd ~ && pwd && zsh -i'
```

On my system, this opens a new window and prints `/mnt/c/Users/cbouchard`, my Windows home directory, which is the current directory of the shell running the command. So seemingly the `cd` did nothing. However,

```
alacritty.exe -e wsl zsh -c "'cd ~ && pwd && zsh -i'"
```

(Note the double-quoting.) This opens an new window and prints `/home/cbouchard`, which is my WSL home directory, i.e., `~`. So the `cd` worked as desired.

My best guess is that Alacritty is dropping the quoting from my positional parameters, so it's interpreting the first command as

```
wsl zsh -c cd ~ && pwd && zsh -i
```

So it runs the `cd` in a subshell, then runs `pwd` and `zsh -i`.

I feel like, if Alacritty will accept positional parameters as separate parameters, it should preserve the quoting/spaces of those parameters.

### System

OS: Windows
Version: alacritty 0.4.2 (f68de37)
Linux/BSD: X11/Wayland, DE or Compositor and WM
Windows: ConPTY

chrisbouchard:
Change the Windows command line argument handling to escape arguments according to [the rules in this article][parsing-args]. (Note the article is written in terms of parsing, but we're interested in escaping.)

[parsing-args]: https://docs.microsoft.com/en-us/cpp/c-language/parsing-c-command-line-arguments

The escaping rules are a bit insane. Each quote must be escaped with a leading backslash. Backslashes must *only* be escaped if they are in a run of backslashes preceding a quote. In that case, each backslash in the run must be escaped with its own backslash. All other backslashes must *not* be escaped and will be interpreted literally.

## Concerns

As mentioned [in this comment][concerns], it's not clear that escaping like this is *always* the correct thing to do. It depends on the shell program, and is not something we can know. I've marked this pull request WIP until we settle that.

I have added a check so that arguments are only escaped when they are required to be. This should make it safer to pass arguments to programs using &ldquo;odd&rdquo; parsing rules (e.g., `cmd.exe`), as long as those arguments are simple.

[concerns]: https://github.com/alacritty/alacritty/issues/3552#issuecomment-610710931

Fixes #3552

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Set IUTF8 termios flag for improved UTF8 input support
 - Dragging files into terminal now adds a space after each path
 - Default binding replacement conditions
+- Escape Windows shell command line arguments
 
 ### Fixed
 
--- a/alacritty_terminal/src/tty/windows/mod.rs
+++ b/alacritty_terminal/src/tty/windows/mod.rs
@@ -226,24 +226,25 @@ fn cmdline<C>(config: &Config<C>) -> String {
 /// between those delimiters are input to or output from the function.
 ///
 /// ```rust
-/// assert_eq!(quote_argument(r#"a b c"#), r#""a b c""#)
-/// assert_eq!(quote_argument(r#"a"bc"#), r#""a\"bc""#)
-/// assert_eq!(quote_argument(r#"a\"bc"#), r#""a\\\"bc""#)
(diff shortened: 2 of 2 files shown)


chrisduerr:
Closing since it has been stale for over a year.

PR #3879, opened 2020-06-23
Fix live right-click expansion

chrisduerr:
While the commit 43c0ad6ea9d2467ccf867a310c4f1e30f5b627c6 introduced
right click as a way to expand the active selection, it did not allow
for holding right click to continuously do so.

This commit remedies that problem by allowing live expansion with while
holding the right mouse button.


kchibisov:
That works great, however I found one more edge case where our expansion is imperfect. In Vi mode it should likely change Vi cursor position, so you may add one more commit here.

chrisduerr:
> That works great, however I found one more edge case where our expansion is imperfect. In Vi mode it should likely change Vi cursor position, so you may add one more commit here.

I actually skipped out on this intentionally, thinking that we shouldn't do this. But of course we should, since we do modify the selection we don't only click around.

chrisduerr:
Should be good now @kchibisov.

PR #3896, opened 2020-06-28
Rename alt_grid to inactive_grid

### LINKED ISSUE #3504: Rename alt_grid in Term struct

chrisduerr:
Since the alt_grid is not always the alternate screen buffer, the name
inactive_grid should fit much better.

Fixes #3504.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -738,9 +738,6 @@ pub struct Term<T> {
 
     pub selection: Option<Selection>,
 
-    /// The grid.
-    grid: Grid<Cell>,
-
     /// Tracks if the next call to input will need to first handle wrapping.
     /// This is true after the last column is set with the input function. Any function that
     /// implicitly sets the line or column needs to set this to false to avoid wrapping twice.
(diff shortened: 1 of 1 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -1063,15 +1065,21 @@ impl<T> Term<T> {
...
-        mem::swap(&mut self.grid, &mut self.alt_grid);
+            self.inactive_grid.cursor = self.inactive_grid.saved_cursor;
+            self.grid.cursor = self.grid.saved_cursor;
+        } else {
+            self.inactive_grid.saved_cursor = self.inactive_grid.cursor;
+            self.grid.saved_cursor = self.grid.cursor;
+        }
 
+        mem::swap(&mut self.grid, &mut self.inactive_grid);
+        self.mode ^= TermMode::ALT_SCREEN;
```
kchibisov, line 1082: I kind of liked that it was in `set` and `unset` methods, maybe we should keep it there? However, knowing that this function is being used only by those methods, maybe it's fine here.

alacritty_terminal/src/term/mod.rs
```diff
@@ -835,8 +838,7 @@ impl<T> Term<T> {
             visual_bell: VisualBell::new(config),
             input_needs_wrap: false,
             grid,
-            alt_grid: alt,
-            alt: false,
+            inactive_grid: alt,
```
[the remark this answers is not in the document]
chrisduerr: Should we? In this case it is absolutely correct.

alacritty_terminal/src/term/mod.rs
chrisduerr: What did you like about it? Seems better to have it just in one place instead of duplicated.

PR #3898, opened 2020-06-28
Document supported escape sequences

### LINKED ISSUE #3440: Document escape sequence support

chrisduerr:
Fixes #3440.

### DIFF
--- a/docs/escape_support.md
+++ b/docs/escape_support.md
@@ -81,8 +81,8 @@ brevity.
 
 | ESCAPE    | STATUS      | NOTE                                               |
 | --------- | ----------- | -------------------------------------------------- |
-| `OSC 0`   | IMPLEMENTED | Tabs are not supported                             |
-| `OSC 1`   | REJECTED    | Tabs are not supported                             |
+| `OSC 0`   | IMPLEMENTED | Icon names are not supported                       |
+| `OSC 1`   | REJECTED    | Icon names are not supported                       |
 | `OSC 2`   | IMPLEMENTED |                                                    |
 | `OSC 4`   | IMPLEMENTED |                                                    |
 | `OSC 10`  | IMPLEMENTED |                                                    |
--- a/docs/escape_support.md
+++ b/docs/escape_support.md
@@ -81,7 +81,7 @@ brevity.
 
 | ESCAPE    | STATUS      | NOTE                                               |
 | --------- | ----------- | -------------------------------------------------- |
-| `OSC 0`   | IMPLEMENTED |                                                    |
+| `OSC 0`   | IMPLEMENTED | Tabs are not supported                             |
 | `OSC 1`   | REJECTED    | Tabs are not supported                             |
 | `OSC 2`   | IMPLEMENTED |                                                    |
 | `OSC 4`   | IMPLEMENTED |                                                    |
@@ -90,7 +90,7 @@ brevity.
 | `OSC 12`  | IMPLEMENTED |                                                    |
 | `OSC 50`  | IMPLEMENTED | Only `CursorShape` is supported                    |
 | `OSC 52`  | IMPLEMENTED | Only Clipboard and primary selection supported     |
-| `OSC 104` | IMPLEMENTED | Only Clipboard and primary selection supported     |
+| `OSC 104` | IMPLEMENTED |                                                    |
 | `OSC 110` | IMPLEMENTED |                                                    |
 | `OSC 111` | IMPLEMENTED |                                                    |
 | `OSC 112` | IMPLEMENTED |                                                    |
--- a/docs/escape_support.md
+++ b/docs/escape_support.md
@@ -1,0 +1,100 @@
+# Escape Sequence Support
+
+This list includes all escape sequences Alacritty currently supports.
+
+### Legend
+
+The available statuses are `PARTIAL`, `IMPLEMENTED` and `REJECTED`. While a
+status of `PARTIAL` means there is still work left to be done, a status of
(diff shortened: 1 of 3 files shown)


kchibisov:
The file should be named a bit differently, I'd say. Like escape_support.md

docs/escape_support.md
```diff
@@ -0,0 +1,112 @@
...
+|            |             |   `12`                                            |
+|            |             |   `25`                                            |
+|            |             |   `1000`                                          |
+|            |             |   `1002`                                          |
+|            |             |   `1004`                                          |
+|            |             |   `1005`                                          |
+|            |             |   `1006`                                          |
+|            |             |   `1007`                                          |
+|            |             |   `1049`                                          |
+|            |             |   `2004`                                          |
```
kchibisov, line 71: I'm not entirely sure how this should be formatted, but it seems a bit weird that it looks like that in a table.

chrisduerr: Have you checked the output? I think it looks fine. I agree with you though that I'm not entirely sure if this is the best format, since it uses up a lot of vertical space while not utilizing horizontal space at all.

Do you have any suggestions?

docs/escape_support.md
```diff
@@ -0,0 +1,112 @@
...
+| `OSC 4`   | IMPLEMENTED |                                                    |
+| `OSC 10`  | IMPLEMENTED |                                                    |
+| `OSC 11`  | IMPLEMENTED |                                                    |
+| `OSC 12`  | IMPLEMENTED |                                                    |
+| `OSC 50`  | IMPLEMENTED | Only `CursorShape` is supported                    |
+| `OSC 52`  | IMPLEMENTED | Only Clipboard and primary selection supported     |
+| `OSC 104` | IMPLEMENTED | Only Clipboard and primary selection supported     |
+| `OSC 110` | IMPLEMENTED |                                                    |
+| `OSC 111` | IMPLEMENTED |                                                    |
+| `OSC 112` | IMPLEMENTED |                                                    |
```
kchibisov, line 108: In general, I guess it's nice to point to a things where those OSC came from and some references where to look them up. Maybe some links to xterm docs, etc?

docs/escape_support.md
kchibisov: I was looking on rendered md file, yeah. maybe just collon separated list is fine instead of virtual layout.

kchibisov:
Also, it could be nice to update PR template to mention that file. Like 'if you're adding support for some escape sequences make sure to update escape_sequence.md` or something like that.

docs/escape_support.md
chrisduerr: I'm not sure if we want to link directly to just XTerm docs, since in many cases they're also just an aggregation of other sources? In our wiki we have a list with a bunch of different resources, that's probably what people should be looking at.

But I'm not sure if we want to link to any of that? Since it's not really the purpose of this document to explain what these escapes are at all, just which ones are supported.

chrisduerr:
> Also, it could be nice to update PR template to mention that file. Like 'if you're adding support for some escape sequences make sure to update escape_sequence.md` or something like that.

That's a great point, yeah.

docs/escape_support.md
chrisduerr: Colon separated? Why colon? I'd imagine if anything it should be comma separated?

kchibisov: yeah, `comma` one.

chrisduerr: Does it look better to you now?

docs/escape_support.md
```diff
@@ -0,0 +1,100 @@
...
+| `OSC 0`   | IMPLEMENTED |                                                    |
+| `OSC 1`   | REJECTED    | Tabs are not supported                             |
+| `OSC 2`   | IMPLEMENTED |                                                    |
+| `OSC 4`   | IMPLEMENTED |                                                    |
+| `OSC 10`  | IMPLEMENTED |                                                    |
+| `OSC 11`  | IMPLEMENTED |                                                    |
+| `OSC 12`  | IMPLEMENTED |                                                    |
+| `OSC 50`  | IMPLEMENTED | Only `CursorShape` is supported                    |
+| `OSC 52`  | IMPLEMENTED | Only Clipboard and primary selection supported     |
+| `OSC 104` | IMPLEMENTED | Only Clipboard and primary selection supported     |
```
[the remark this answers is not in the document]
chrisduerr: Yep, just copy-paste error, thanks!

docs/escape_support.md
```diff
@@ -0,0 +1,100 @@
...
+| `CSI t`    | PARTIAL     | Only parameters `22` and `23` are supported       |
+| `CSI u`    | IMPLEMENTED |                                                   |
+| `CSI X`    | IMPLEMENTED |                                                   |
+| `CSI Z`    | IMPLEMENTED |                                                   |
+
+### OSC (Operating System Command) - `ESC ]`
+
+| ESCAPE    | STATUS      | NOTE                                               |
+| --------- | ----------- | -------------------------------------------------- |
+| `OSC 0`   | IMPLEMENTED | Tabs are not supported                             |
```
kchibisov, line 84: It has nothing to do with tabs, it's just a window icon, aka _NET_WM_ICON.

PR #3900, opened 2020-06-28
Dim foreground color for truecolor text

### LINKED ISSUE #3766: Dim foreground color for truecolored text as well.
At least vte/xterm are doing so.

carloabelli:
Fixes #3766

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Display not scrolling when printing new lines while scrolled in history
 - Regression in font rendering on macOS
 - Scroll down escape (`CSI Ps T`) incorrectly pulling lines from history
+- Dim escape (`CSI 2 m`) support for truecolor text
 
 ## 0.4.3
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Display not scrolling when printing new lines while scrolled in history
 - Regression in font rendering on macOS
 - Scroll down escape (`CSI Ps T`) incorrectly pulling lines from history
+- Dim truecolor foreground
```
chrisduerr, line 54: This almost makes it sound like the truecolor foreground is now always dimmed, I think we can be a little more specific:

```suggestion
- Dim escape (`CSI 2 m`) support for truecolor text
```

kchibisov: Shouldn't that be in `Added` section or it's fine to be in `Fixed`?

chrisduerr: I think fixed is fine, since it's kind of a bug that it hasn't been done until now.

benknoble:
This fix doesn't seem to be quite working for me in vim: two highlight groups with the same color (but one underlined) show up very differently (one is dimmed, which is why I suspect alacritty and not vim; I don't have another true-color compatible term to test, but MacVim doesn't do this).

Screenshot:

Capture d’écran 2020-10-07 à 12 54 27

PR #3906, opened 2020-07-01
Add missing changelog entry

chrisduerr:
This adds a missing changelog entry for a bug fixed in
bc60782e424359d818a22ffa2ffa3c2dc88ab6fe.


chrisduerr:
While this bug was present on master, it is not present on 0.4.3. Thus a changelog entry is not necessary.

Issue #3583, opened 2020-04-13
Reflow deleting history


chrisduerr:
When holding <kbd>space</kbd> until the cursor is on the next line of the prompt, then growing the shell width, content above the cursor will be deleted and the cursor is moved upwards.

Repro:
 * run `bash` or `zsh` (only ones tested, others probably also affected)
 * `ls -lah` to write some content
 * <kbd>space</kbd> until on the next line
 * grow terminal width
   -> The `ls` content is deleted

PR #3910, opened 2020-07-03
Fix cursor reflow

### LINKED ISSUE #3584: Reflow not wrapping cursor correctly
When the grid width is changed and content reflows, the cursor should reflow with the content. This is necessary to make sure that writes will be correct when occuring during a resize.

Repro:
 * Run `echo -n "[LONG LINE]" && sleep 10 && echo -n "xxx"`
 * Shrink terminal to wrap the long line, before sleep is done
   -> The `xxx` is written at the wrong location, it should be right after `[LONG LINE]`

chrisduerr:
To make sure that output is consistent even while resizing the window,
the cursor will now reflow with the content whenever the window size is
changed.

Since the saved cursor is more likely to represent a position in the
grid rather than a reference to the content below it and handling of
resize before jumping back to it is more likely than with the primary
cursor, no reflow is performed for the saved cursor

Fixes #3584.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Documentation for class in `--help` missing information on setting general class
 - Linewrap tracking when switching between primary and alternate screen buffer
 - Preservation of the alternate screen's saved cursor when swapping to primary screen and back
+- Reflow of cursor during resize
 
 ## 0.4.3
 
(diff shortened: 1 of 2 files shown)


kchibisov:
It's still a bit buggy, but much better than before.

For example try `for i in `seq 0 10000`; do echo -ne "1234567890"; sleep 0.1 ; done` and start resizing it by moving your mouse out side of the window clockwise. Do it for a bit and then introspect the output. You may notice that there's incomplete sequences of `1234567890`, like `123456789` or even shorter.

On the other side vte handles that correctly and you only have `1234567890` sequences in the output.

kchibisov:
This PR reintroduces https://github.com/alacritty/alacritty/issues/3583 for me, the commit from where it was branched works fine.

kchibisov:
The line deletion I'm talking about.

record

on master the same looks like that.

record

PR #3916, opened 2020-07-04
Adds feature to select random colors on startup

cr0ax:
Since Alacritty watches the `alacritty.yml` file it is not possible to
modify that file on-the-fly to produce randomized colors when starting
the terminal. Additionally, it just feels kind of hacky to do it like
that.

With this commit, we preserve existing behavior for people who have not added
the random_colors option, while enabling it for those who have.

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -256,8 +256,8 @@ impl Options {
         match &config.random_colors {
             Some(colors) => {
                 info!("Available colors {:?}", colors);
-                config.colors = colors[rand::random::<usize>()%colors.len()].clone();
-            }
+                config.colors = colors[rand::random::<usize>() % colors.len()].clone();
+            },
             None => info!("No random colors found"),
         }
 
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -2,10 +2,12 @@ use std::cmp::max;
 use std::path::PathBuf;
 
 use clap::{crate_authors, crate_description, crate_name, crate_version, App, Arg};
-use log::{self, error, LevelFilter};
+use log::{self, error, info, LevelFilter};
 
 use alacritty_terminal::config::{Delta, Dimensions, Program, DEFAULT_NAME};
 use alacritty_terminal::index::{Column, Line};
+
+use rand;
 
 use crate::config::Config;
 
@@ -249,6 +251,14 @@ impl Options {
         match self.working_dir.or_else(|| config.working_directory.take()) {
             Some(ref wd) if !wd.is_dir() => error!("Unable to set working directory to {:?}", wd),
             wd => config.working_directory = wd,
+        }
+
+        match &config.random_colors {
+            Some(colors) => {
+                info!("Available colors {:?}", colors);
+                config.colors = colors[rand::random::<usize>()%colors.len()].clone();
+            }
+            None => info!("No random colors found"),
         }
 
         if let Some(lcr) = self.live_config_reload {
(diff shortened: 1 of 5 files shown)


alacritty/src/cli.rs
```diff
@@ -251,6 +253,14 @@ impl Options {
...
         }
 
+        match &config.random_colors {
+            Some(colors) => {
+                info!("Available colors {:?}", colors);
+                config.colors = colors[rand::random::<usize>() % colors.len()].clone();
+            },
+            None => info!("No random colors found"),
+        }
+
```
cr0ax, line 263: Apologies if this is not the place to do this.


I needed a mutable `config` and this looked like a good spot. Happy to move it if not appropriate.

kchibisov:
Thanks for the PR, but this is not something that suits upstream. The random colors could be achieved by executing a shell script, which sets colors and you can put it in bashrc(or whatever shell you're using) if you want to. However those colors could be reset, so if someone wants for those to be persistent, they should implement `OSC 5` into alacritty, which is welcoming, ofc.

cr0ax:
@kchibisov

What is `OSC5`?

Why does this terminal have any color support in the config if it can all be a accomplished in an rc?

cr0ax:
Also I want to point out this is to replace themes

As in https://github.com/eendroroy/alacritty-theme. Not single colors.

An rc cannot change the background color of the terminal, can it?

kchibisov:
background - `echo -e "\e]11;#ff00ff\007"` 
foreground - `echo -e "\e]10;#ff00ff\007"`
cursor - `echo -e "\e]12;#ff00ff\007"`

The thing is that those things are reset by `reset`, so that's why you set them in a terminal config, since you should reset to some default value.

OSCs https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands

cr0ax:
I can't find any documentation anywhere for

```
background - echo -e "\e]11;#ff00ff\007"
foreground - echo -e "\e]10;#ff00ff\007"
cursor - echo -e "\e]12;#ff00ff\007"
```

could you link a reference?

I've been all over but I can't find anywhere where 10-12 are referenced as those things (they work though!)

Also the `\007` doesn't seem to do anything

kchibisov:
On the link I've posted before, scroll a bit until
```
Ps = 1 0  ⇒  Change VT100 text foreground color to Pt.
Ps = 1 1  ⇒  Change VT100 text background color to Pt.
Ps = 1 2  ⇒  Change text cursor color to Pt.
```

They are all `OSC` sequences.

kchibisov:
There's also an `OSC 17` and `OSC 19` to change selection color, however it's not implement in alacritty, as well as some other obscure sequences in that doc.

cr0ax:
Ok, and I'm guessing

```
Ps = 1 0 4 ; c ⇒  Reset Color Number c.  It is reset to the
          color specified by the corresponding X resource.  Any number
          of c parameters may be given.  These parameters correspond to
          the ANSI colors 0-7, their bright versions 8-15, and if sup-
          ported, the remainder of the 88-color or 256-color table.  If
          no parameters are given, the entire table will be reset.
```

corresponds to

```
normal:
    black:   '#111111'
    red:     '#aa4450'
    green:   '#719611'
    yellow:  '#cc8800'
    blue:    '#6688aa'
    magenta: '#8f6f8f'
    cyan:    '#528b8b'
    white:   '#d3d3d3'
```

etc

kchibisov:
yeah, you do something like `echo -e "\e]4;1;#ff00ff\007"` to set red(1) and then you can `echo -e "\e]104;1;\007"` to reset it.

cr0ax:
Got it :100: thanks

Still don't know what the `\007` is for but it seems to be working

thanks again!

PR #3917, opened 2020-07-04
Fix compilation on 32bit targets

### LINKED ISSUE #3915: Build error on armv7l, font package.
I am trying to compile alacritty master (tried -ligatures and -git) (0.5.0) on my Raspberry pi 4 running Arch arm. The package availiable in the community repo is 0.4.3-1 and will not run with OpenGL driver enabled. (GLSL 3.30 not supported, see currently open issue by someone else).
I tried compiling the latest version of fonts but got the same error. I have an issue on the author of font github as well. But as you are using an old package for alacritty i tought id make one here.
I've tried current master as well as the AUR packages. Obviously they all use the same font package...

See build errors below.

### System

OS: Arch Linux arm/aarch64
Version: master
Linux: X11 DWM and Xfce4

### Logs

Compile time error in included package "font".
src/ft/mod.rs:339:26-38: error E0308: Mismatched types.
Compiler help message: "you can convert an `i64` to `i32` and panic if the converted value wouldn't fit.
rustc --explain E0308

kchibisov:
Fixes #3915.

### DIFF
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -8,7 +8,7 @@ use std::rc::Rc;
 use freetype::tt_os2::TrueTypeOS2Table;
 use freetype::{self, Library, Matrix};
 use freetype::{freetype_sys, Face as FTFace};
-use libc::c_uint;
+use libc::{c_long, c_uint};
 use log::{debug, trace};
 
 pub mod fc;
@@ -89,8 +89,8 @@ fn to_freetype_26_6(f: f32) -> isize {
 }
 
 #[inline]
-fn to_fixedpoint_16_6(f: f64) -> i64 {
-    (f * 65536.0) as i64
+fn to_fixedpoint_16_6(f: f64) -> c_long {
+    (f * 65536.0) as c_long
 }
 
 impl Rasterize for FreeTypeRasterizer {


font/src/ft/mod.rs
```diff
@@ -89,8 +89,8 @@ fn to_freetype_26_6(f: f32) -> isize {
 }
 
 #[inline]
-fn to_fixedpoint_16_6(f: f64) -> i64 {
-    (f * 65536.0) as i64
+fn to_fixedpoint_16_6(f: f64) -> c_long {
+    (f * 65536.0) as c_long
```
chrisduerr, line 93: Can't we just use an `isize` here?

kchibisov: Is it a thing that `clong` matches `isize` on any combination of arches + bitness? I'm not sure that it's true and it's not stated anywhere, so I'd like to avoid that, so I've picked `clong` for that to void any potential issues in any form.

chrisduerr: I assumed that was your reasoning, which makes sense, yeah. Matching the external interface is usually the better choice.

But would you mind changing `to_freetype_26_6` too then? Just for consistency.

kchibisov: The other API is in isize.

PR #3918, opened 2020-07-04
Fix compilation on 32bit targets

### LINKED ISSUE #3915: Build error on armv7l, font package.

chrisduerr:
This works around libc's issue of setting `c_long` to a fixed value
rather than making it depend on the architecture, which caused the patch
92ea355eeea538bd868eaebdc03469630aba281c to fail.

Fixes #3915.

### DIFF
--- a/font/src/ft/mod.rs
+++ b/font/src/ft/mod.rs
@@ -8,7 +8,7 @@ use std::rc::Rc;
 use freetype::tt_os2::TrueTypeOS2Table;
 use freetype::{self, Library, Matrix};
 use freetype::{freetype_sys, Face as FTFace};
-use libc::{c_long, c_uint};
+use libc::c_uint;
 use log::{debug, trace};
 
 pub mod fc;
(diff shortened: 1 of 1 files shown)


kchibisov:
There's not need to change `c_long`, everything is working as it should.

PR #3921, opened 2020-07-05
Fix saved cursor handling

chrisduerr:
This resolves several problems with handling of the saved cursor when
switching between primary and alternate screen. Additionally ref-tests
are also added for all common interactions to make sure the behavior
does not regress.

The behavior is based on XTerm's behavior except for interaction with
`reset`. XTerm does not reset the alternate screen's saved cursor on
`reset`, but VTE does. Since a `reset` should reset as much as possible,
Alacritty copies VTE here instead of XTerm.


CHANGELOG.md
```diff
@@ -55,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dim escape (`CSI 2 m`) support for truecolor text
 - Incorrectly deleted lines when increasing width with a prompt wrapped using spaces
 - Documentation for class in `--help` missing information on setting general class
+- Preservation of the alternate screen's saved cursor when swapping to primary screen and back
```
chrisduerr, line 58: The changelog entry solely describes the changes to the 0.4.3 version.

Alacritty's behavior strongly regressed on master when the cursor was moved to the grid, which has also been fixed but does not warrant a changelog entry. However the behavior on 0.4.3 wasn't correct either.

The behavior now matches XTerm except for `reset` as mentioned in the commit message.

PR #3922, opened 2020-07-05
Preserve wrapline flag across alt screen switches

chrisduerr:
While neither VTE, URxvt nor Kitty handle this, preserving the wrapline
flag across alternate screen switches seems like the correct thing to
do. XTerm also does handle this correctly, which indicates that it is a
bug and not a feature.


kchibisov:
Btw, should be add a changelog entry for that 'fix'? i guess we can do that.

PR #3931, opened 2020-07-08
Implement cursor blinking

### LINKED ISSUE #302: Cursor Not Blinking
Is there an option to cause the cursor to blink? I find it to be helpful confirmation that the terminal has (or doesn't have) focus.

Dettorer:
> Please make sure to document all user-facing changes in the
> `CHANGELOG.md` file.

Done

> If support for a new escape sequence was added, it should be documented
> in `./docs/escape_support.md`.

Done
note: `CSI ? 12 h` and `CSI ? 12 l` were already marked as supported.

The default blinking rate is 500ms hide -> 500ms show -> …

TODO:
- [x] wether to blink or not by default
- [x] start blinking at startup if configuration says so
- [x] start or stop blinking at vi/normal mode change according to configuration
- [x] handle `CSI 1/3/5 SP q` (set cursor style to blinking block/underline/bar)
- [x] make the "blinking cursor" an option on each cursor style instead of separate new styles
- [x] make the vi cursor blink too

Optionally:
- [x] the blinking rate
- [x] completely disable blinking (ignore escape sequences), possibly by setting the rate to `0`
- [x] reinitialize the blinking cycle on user input
- (?) allow different blinking rate for the default cursor and vi cursor

Fixes #302

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -359,7 +359,7 @@
   #vi_mode_style: None
 
   # Cursor blinking interval in milliseconds. Using `0` will disable blinking.
-  #blink_rate: 750
+  #blink_rate: 666
 
   # If this is `true`, the cursor will be rendered as a hollow box when the
   # window is not focused.
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -359,7 +359,7 @@
   #vi_mode_style: None
 
   # Cursor blinking interval in milliseconds. Using `0` will disable blinking.
-  #blink_rate: 666
+  #blink_rate: 500
 
   # If this is `true`, the cursor will be rendered as a hollow box when the
   # window is not focused.
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -358,8 +358,8 @@
   # See `cursor.style` for available options.
   #vi_mode_style: None
 
-  # Cursor blinking interval in milliseconds. Using `0` will disable blinking.
-  #blink_rate: 500
+  # Cursor blinking interval in milliseconds.
+  #blink_interval: 500
 
   # If this is `true`, the cursor will be rendered as a hollow box when the
   # window is not focused.
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -339,22 +339,22 @@
 #cursor:
   # Cursor style
   #style:
-      # Cursor shape
-      #
-      # Values for `shape`:
-      #   - ▇ Block
-      #   - _ Underline
-      #   - | Beam
-      #shape: Block
+    # Cursor shape
+    #
+    # Values for `shape`:
+    #   - ▇ Block
+    #   - _ Underline
+    #   - | Beam
+    #shape: Block
 
-      # Cursor blinking state
-      #
-      # Values for `blinking`:
-      #   - Never: Prevent the cursor from ever blinking
-      #   - Off: Disable blinking by default
-      #   - On: Enable blinking by default
-      #   - Always: Force the cursor to always blink
-      #blinking: Off
+    # Cursor blinking state
+    #
+    # Values for `blinking`:
+    #   - Never: Prevent the cursor from ever blinking
+    #   - Off: Disable blinking by default
+    #   - On: Enable blinking by default
+    #   - Always: Force the cursor to always blink
+    #blinking: Off
 
   # Vi mode cursor style
   #
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -338,12 +338,17 @@
 
 #cursor:
   # Cursor style
-  #
-  # Values for `style`:
-  #   - ▇ Block
-  #   - _ Underline
-  #   - | Beam
-  #style: Block
+  #style:
+      # The shape of the cursor
+      #
+      # Values for `shape`:
+      #   - ▇ Block
+      #   - _ Underline
+      #   - | Beam
+      #shape: Block
+
+      # Whether the cursor should blink by default or not.
+      #blinking: false
 
   # Vi mode cursor style
   #
@@ -352,6 +357,11 @@
   #
   # See `cursor.style` for available options.
   #vi_mode_style: None
+
+  # Cursor blinking interval
+  #
+  # Rate in milliseconds used for cursor blinking when a blinking cursor style is active. A `blink_rate` of `0` will completely disable cursor blinking.
+  #blink_rate: 500
 
   # If this is `true`, the cursor will be rendered as a hollow box when the
   # window is not focused.
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -339,7 +339,7 @@
 #cursor:
   # Cursor style
   #style:
-      # The shape of the cursor
+      # Cursor shape
       #
       # Values for `shape`:
       #   - ▇ Block
@@ -347,7 +347,7 @@
       #   - | Beam
       #shape: Block
 
-      # Whether the cursor should blink by default or not.
+      # Default cursor blinking state.
       #blinking: false
 
   # Vi mode cursor style
@@ -358,10 +358,8 @@
   # See `cursor.style` for available options.
   #vi_mode_style: None
 
-  # Cursor blinking interval
-  #
-  # Rate in milliseconds used for cursor blinking when a blinking cursor style is active. A `blink_rate` of `0` will completely disable cursor blinking.
-  #blink_rate: 500
+  # Cursor blinking interval in milliseconds. Using `0` will disable blinking.
+  #blink_rate: 750
 
   # If this is `true`, the cursor will be rendered as a hollow box when the
   # window is not focused.
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -347,8 +347,14 @@
       #   - | Beam
       #shape: Block
 
-      # Default cursor blinking state.
-      #blinking: false
+      # Cursor blinking state
+      #
+      # Values for `blinking`:
+      #   - Never: Prevent the cursor from ever blinking
+      #   - Off: Disable blinking by default
+      #   - On: Enable blinking by default
+      #   - Always: Force the cursor to always blink
+      #blinking: Off
 
   # Vi mode cursor style
   #
@@ -359,7 +365,7 @@
   #vi_mode_style: None
 
   # Cursor blinking interval in milliseconds.
-  #blink_interval: 500
+  #blink_interval: 750
 
   # If this is `true`, the cursor will be rendered as a hollow box when the
   # window is not focused.
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -507,26 +507,44 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         };
 
         // Check terminal cursor style.
-        let blinking = cursor_style.blinking(self.terminal.cursor_style().blinking);
+        let terminal_blinking = self.terminal.cursor_style().blinking;
+        let blinking = cursor_style.blinking_override().unwrap_or(terminal_blinking);
 
         // Update cursor blinking state.
-        if !blinking || self.config.cursor.blink_rate == 0 || !self.terminal.is_focused {
-            self.scheduler.unschedule(TimerId::BlinkCursor);
-            *self.cursor_hidden = false;
-            self.terminal.dirty = true;
-        } else if !self.scheduler.scheduled(TimerId::BlinkCursor) {
+        self.scheduler.unschedule(TimerId::BlinkCursor);
+        if blinking && self.terminal.is_focused {
             self.scheduler.schedule(
                 GlutinEvent::UserEvent(Event::BlinkCursor),
                 Duration::from_millis(self.config.cursor.blink_rate),
                 true,
                 TimerId::BlinkCursor,
             )
+        } else {
+            *self.cursor_hidden = false;
+            self.terminal.dirty = true;
         }
     }
 
+    /// Enable input mode.
+    ///
+    /// Putting the user in input mode will temporarily disable some features like terminal cursor
+    /// blinking or the mouse cursor.
+    ///
+    /// This mode will disable itself automatically.
     #[inline]
-    fn show_cursor(&mut self) {
-        *self.cursor_hidden = false;
+    fn enable_input_mode(&mut self) {
+        // Disable cursor blinking.
+        let cursor_blink_rate = self.config.cursor.blink_rate;
+        if let Some(timer) = self.scheduler.get_mut(TimerId::BlinkCursor) {
+            timer.deadline = Instant::now() + Duration::from_millis(cursor_blink_rate);
+            *self.cursor_hidden = false;
+            self.terminal.dirty = true;
+        }
+
+        // Hide mouse cursor.
+        if self.config.ui_config.mouse.hide_when_typing {
+            self.window.set_mouse_visible(false);
+        }
     }
 
     #[inline]
@@ -1158,9 +1176,6 @@ impl<N: Notify + OnResize> Processor<N> {
 
         processor.ctx.terminal.update_config(&config);
 
-        // Update cursor blinking.
-        processor.ctx.update_cursor_blinking();
-
         // Reload cursor if its thickness has changed.
         if (processor.ctx.config.cursor.thickness() - config.cursor.thickness()).abs()
             > std::f64::EPSILON
@@ -1203,6 +1218,9 @@ impl<N: Notify + OnResize> Processor<N> {
         set_font_smoothing(config.ui_config.font.use_thin_strokes());
 
         *processor.ctx.config = config;
+
+        // Update cursor blinking.
+        processor.ctx.update_cursor_blinking();
 
         processor.ctx.terminal.dirty = true;
     }
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1081,7 +1081,10 @@ impl<N: Notify + OnResize> Processor<N> {
                                 // Reenable blinking cursor based on current mode.
                                 let config = &processor.ctx.config;
                                 let cursor_style = if terminal.mode().contains(TermMode::VI) {
-                                    config.cursor.vi_mode_style().unwrap_or_else(|| config.cursor.style())
+                                    config
+                                        .cursor
+                                        .vi_mode_style()
+                                        .unwrap_or_else(|| config.cursor.style())
                                 } else {
                                     config.cursor.style()
                                 };
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -506,7 +506,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     #[inline]
     fn on_typing_start(&mut self) {
         // Disable cursor blinking.
-        let blink_interval = self.config.cursor.blink_interval;
+        let blink_interval = self.config.cursor.blink_interval();
         if let Some(timer) = self.scheduler.get_mut(TimerId::BlinkCursor) {
             timer.deadline = Instant::now() + Duration::from_millis(blink_interval);
             *self.cursor_hidden = false;
@@ -709,7 +709,7 @@ impl<'a, N: Notify + 'a, T: EventListener> ActionContext<'a, N, T> {
         if blinking && self.terminal.is_focused {
             self.scheduler.schedule(
                 GlutinEvent::UserEvent(Event::BlinkCursor),
-                Duration::from_millis(self.config.cursor.blink_interval),
+                Duration::from_millis(self.config.cursor.blink_interval()),
                 true,
                 TimerId::BlinkCursor,
             )
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1081,10 +1081,8 @@ impl<N: Notify + OnResize> Processor<N> {
                                 // Reenable blinking cursor based on current mode.
                                 let config = &processor.ctx.config;
                                 let cursor_style = if terminal.mode().contains(TermMode::VI) {
-                                    config
-                                        .cursor
-                                        .vi_mode_style()
-                                        .unwrap_or_else(|| config.cursor.style())
+                                    let style = config.cursor.style();
+                                    config.cursor.vi_mode_style().unwrap_or(style)
                                 } else {
                                     config.cursor.style()
                                 };
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -515,7 +515,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         if blinking && self.terminal.is_focused {
             self.scheduler.schedule(
                 GlutinEvent::UserEvent(Event::BlinkCursor),
-                Duration::from_millis(self.config.cursor.blink_rate),
+                Duration::from_millis(self.config.cursor.blink_interval),
                 true,
                 TimerId::BlinkCursor,
             )
@@ -534,9 +534,9 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     #[inline]
     fn enable_input_mode(&mut self) {
         // Disable cursor blinking.
-        let cursor_blink_rate = self.config.cursor.blink_rate;
+        let blink_interval = self.config.cursor.blink_interval;
         if let Some(timer) = self.scheduler.get_mut(TimerId::BlinkCursor) {
-            timer.deadline = Instant::now() + Duration::from_millis(cursor_blink_rate);
+            timer.deadline = Instant::now() + Duration::from_millis(blink_interval);
             *self.cursor_hidden = false;
             self.terminal.dirty = true;
         }
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -768,7 +768,6 @@ pub struct Processor<N> {
     event_queue: Vec<GlutinEvent<'static, Event>>,
     search_state: SearchState,
     cli_options: CLIOptions,
-    cursor_hidden: bool,
 }
 
 impl<N: Notify + OnResize> Processor<N> {
@@ -801,7 +800,6 @@ impl<N: Notify + OnResize> Processor<N> {
             clipboard,
             search_state: SearchState::new(),
             cli_options,
-            cursor_hidden: false,
         }
     }
 
@@ -911,7 +909,7 @@ impl<N: Notify + OnResize> Processor<N> {
                 scheduler: &mut scheduler,
                 search_state: &mut self.search_state,
                 cli_options: &self.cli_options,
-                cursor_hidden: &mut self.cursor_hidden,
+                cursor_hidden: &mut self.display.cursor_hidden,
                 event_loop,
             };
             let mut processor = input::Processor::new(context, &self.display.highlighted_url);
@@ -950,7 +948,6 @@ impl<N: Notify + OnResize> Processor<N> {
                     &self.mouse,
                     self.modifiers,
                     &self.search_state,
-                    self.cursor_hidden,
                 );
             }
         });
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -22,7 +22,7 @@ use glutin::event_loop::{ControlFlow, EventLoop, EventLoopProxy, EventLoopWindow
 use glutin::platform::desktop::EventLoopExtDesktop;
 #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
 use glutin::platform::unix::EventLoopWindowTargetExtUnix;
-use log::info;
+use log::{info, warn};
 use serde_json as json;
 
 #[cfg(target_os = "macos")]
@@ -804,6 +804,15 @@ impl<N: Notify + OnResize> Processor<N> {
     {
         let mut scheduler = Scheduler::new();
 
+        if self.config.cursor.style.blinking {
+            if let Err(err) = event_loop
+                .create_proxy()
+                .send_event(Event::TerminalEvent(TerminalEvent::CursorStartBlinking))
+            {
+                warn!("Could not make the cursor blink at startup: {}", err);
+            }
+        }
+
         event_loop.run_return(|event, event_loop, control_flow| {
             if self.config.ui_config.debug.print_events {
                 info!("glutin event: {:?}", event);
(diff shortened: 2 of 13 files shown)


chrisduerr:
Hey @Dettorer, do you still plan to work on this? If you need some help please let me know.

Dettorer:
Yep, vacation got in the way but I still plan to work on this.

alacritty.yml
```diff
@@ -325,9 +325,9 @@
   # Cursor style
   #
   # Values for `style`:
-  #   - ▇ Block
-  #   - _ Underline
-  #   - | Beam
+  #   - ▇ Block or BlinkingBlock
+  #   - _ Underline or BlinkingUnderline
+  #   - | Beam or BlinkingBeam
```
stertingen, line 330: I'd prefer to have a separate option for this, like `blink: true`. Here, we mix two different things in a single configuration option.

Maybe in the future, support for custom cursors may be added and someone is crazy enough to want a blinking emoji. That would be easier if I could set something like: `style: '★'` and set the blink option elsewhere.

alacritty.yml
```diff
@@ -338,6 +338,13 @@
   # See `cursor.style` for available options.
   #vi_mode_style: None
 
+  # Blinking cursors blink rate in milliseconds
+  #
+  # If a cursor is blinking, defines at what rate. This concerns the "Blinking*"
+  # cursor style above as well as escape sequences that makes the cursor start
+  # blinking.
+  #blink_rate: 500
```
stertingen, line 346: Or, instead of adding a separate option, set `blink_rate` to `0` to disable blinking.

One more thing: The unit of the value is ms, right? I'd suggest to rename the config option to `blink_period` or make this a proper rate where the default value is 2 Hz. Or, at least, mention somewhere in the comment that the value has the unit ms.

alacritty.yml
Dettorer: That would work too yes, I went for this kind of settings because of the escape sequences (not handled yet in the current state of that PR) that instruct the terminal to switch to "Block" or "BlinkingBlock", but that's not a blocker.

alacritty.yml
Dettorer: Yes, setting `blink_rate` to 0 does disable blinking, I just forgot to say it in the comment, thanks for the reminder.

The fact that it's in ms is said on the first line of that comment. I personally find a value of ms more common than a frequency in Hz, I rarely see tools that have configuration options in Hz.

stertingen: Oh yeah, my bad. I was too focused on the paragraph, nevermind then.

alacritty.yml
chrisduerr: As mentioned previously by @kchibisov, you can actually combine this with the blink rate option.

chrisduerr: > Maybe in the future, support for custom cursors may be added and someone is crazy enough to want a blinking emoji. That would be easier if I could set something like: style: '★' and set the blink option elsewhere.

That's certainly not desired, but generally you are right. This should just be a separate boolean flag.

Dettorer: I can make the `CursorStyle` enum a two-fields struct: `shape` (current enum) and `blinking` (bool).
Or keep the current enum with `Blinking*` variants until alacritty supports more shapes, what would you think is more reasonable @chrisduerr?

chrisduerr: You should probably have a `CursorShape` enum and a `CursorStyle` struct that has a boolean option if the cursor should be blinking or not.

alacritty.yml
```diff
@@ -321,13 +321,18 @@
...
 #cursor:
-  # Cursor style
-  #
-  # Values for `style`:
-  #   - ▇ Block
-  #   - _ Underline
-  #   - | Beam
-  #style: Block
+  # Cursor style, defines its shape and attributes
+  style:
```
chrisduerr, line 325:
```suggestion
  #style:
```

alacritty.yml
```diff
@@ -321,13 +321,18 @@
...
 
 #cursor:
-  # Cursor style
-  #
-  # Values for `style`:
-  #   - ▇ Block
-  #   - _ Underline
-  #   - | Beam
-  #style: Block
+  # Cursor style, defines its shape and attributes
```
chrisduerr, line 324:
```suggestion
  # Cursor style
```

alacritty.yml
```diff
@@ -321,13 +321,18 @@
...
+  style:
+      # The displayed shape of the cursor
```
chrisduerr, line 326:
```suggestion
      # The shape of the cursor
```

alacritty.yml
```diff
@@ -321,13 +321,18 @@
...
+      #
+      # Values for `shape`:
+      #   - ▇ Block
+      #   - _ Underline
+      #   - | Beam
+      #shape: Block
+
+      # Whether the cursor should blink by default or not
```
chrisduerr, line 334:
```suggestion
      # Whether the cursor should blink by default or not.
```

alacritty.yml
```diff
@@ -337,6 +342,16 @@
   # See `cursor.style` for available options.
   #vi_mode_style: None
 
+  # Blinking cursors blink rate in milliseconds
```
chrisduerr, line 345:
```suggestion
  # Cursor blinking interval
```

alacritty.yml
```diff
@@ -337,6 +342,16 @@
...
+  #
+  # If a cursor is blinking, defines at what rate. This concerns the "Blinking*"
+  # cursor style above as well as escape sequences that make the cursor start
+  # blinking.
+  #
+  # Setting this to 0 completely disables blinking, even escape sequences
+  # related to blinking will be ignored in that case.
```
chrisduerr, line 352:
```suggestion
  # Rate in milliseconds used for cursor blinking when a blinking cursor style is active. A `blink_rate` of `0` will completely disable cursor blinking.
```

alacritty/src/event.rs
```diff
@@ -749,6 +749,15 @@ impl<N: Notify + OnResize> Processor<N> {
...
         let mut scheduler = Scheduler::new();
 
+        if self.config.cursor.style.blinking {
+            if let Err(err) = event_loop
+                .create_proxy()
+                .send_event(Event::TerminalEvent(TerminalEvent::CursorStartBlinking))
+            {
+                warn!("Could not make the cursor blink at startup: {}", err);
+            }
+        }
```
chrisduerr, line 759: The send_event here doesn't need any error handling, just how it's done for other channels in Alacritty.

It also seems pointless to send an event to the event loop, when we are literally right in the event handler, that just seems like a pointless indirection. Can't we just start the blinking ourselves?

alacritty/src/event.rs
```diff
@@ -923,6 +932,29 @@ impl<N: Notify + OnResize> Processor<N> {
                     },
                     TerminalEvent::MouseCursorDirty => processor.reset_mouse_cursor(),
                     TerminalEvent::Exit => (),
+                    TerminalEvent::CursorStartBlinking => {
+                        let rate = processor.ctx.config.cursor.blink_rate;
+                        let scheduler = &mut processor.ctx.scheduler;
+                        if rate != 0 && scheduler.get_mut(TimerId::CursorBlinking).is_none() {
+                            processor.ctx.scheduler.schedule(
```
chrisduerr, line 939: Why define a shorthand for scheduler, but then not use it?

alacritty/src/event.rs
```diff
@@ -923,6 +932,29 @@ impl<N: Notify + OnResize> Processor<N> {
                     },
                     TerminalEvent::MouseCursorDirty => processor.reset_mouse_cursor(),
                     TerminalEvent::Exit => (),
+                    TerminalEvent::CursorStartBlinking => {
+                        let rate = processor.ctx.config.cursor.blink_rate;
+                        let scheduler = &mut processor.ctx.scheduler;
+                        if rate != 0 && scheduler.get_mut(TimerId::CursorBlinking).is_none() {
```
chrisduerr, line 938: What if the configuration file is changed to disable blinking while Alacritty is active? We should support live config reload for this.

alacritty/src/event.rs
```diff
@@ -923,6 +932,29 @@ impl<N: Notify + OnResize> Processor<N> {
...
+                                    TerminalEvent::CursorBlink,
+                                )),
+                                Duration::from_millis(rate),
+                                true,
+                                TimerId::CursorBlinking,
+                            )
+                        }
+                    },
+                    TerminalEvent::CursorBlink => {
+                        processor.ctx.terminal.cursor_blinking_out ^= true; // flip boolean value
```
chrisduerr, line 950:
```suggestion
                        processor.ctx.terminal.cursor_blinking_out ^= true;
```

This comment is pointless.

alacritty/src/input.rs
```diff
@@ -864,6 +867,12 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
             },
             ElementState::Released => *self.ctx.suppress_chars() = false,
         }
+
+        // Reset the cursor blinking cycle when the user types
```
chrisduerr, line 871:
```suggestion
        // Disable cursor blinking while the user is typing.
```

alacritty/src/input.rs
```diff
@@ -864,6 +867,12 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
...
+        if let Some(timer) = self.ctx.scheduler_mut().get_mut(TimerId::CursorBlinking) {
+            timer.deadline = Instant::now() + Duration::from_millis(500);
+            self.ctx.terminal_mut().cursor_blinking_out = false;
+        }
```
chrisduerr, line 875: Should this really always be done when any key is pressed, or only when a character is received? I'd argue that just pressing Control shouldn't disable cursor blinking.

alacritty/src/input.rs
```diff
@@ -864,6 +867,12 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
             },
             ElementState::Released => *self.ctx.suppress_chars() = false,
         }
+
+        // Reset the cursor blinking cycle when the user types
+        if let Some(timer) = self.ctx.scheduler_mut().get_mut(TimerId::CursorBlinking) {
+            timer.deadline = Instant::now() + Duration::from_millis(500);
```
chrisduerr, line 873: Where is this 500 coming from?

alacritty/src/renderer/mod.rs
```diff
@@ -1034,6 +1034,7 @@ impl<'a> RenderApi<'a> {
 
     pub fn render_cell(&mut self, cell: RenderableCell, glyph_cache: &mut GlyphCache) {
         let chars = match cell.inner {
+            RenderableCellContent::Cursor(cursor_key) if cursor_key.blinking_out => return,
```
chrisduerr, line 1037: Isn't this a bit late? We could save ourselves a lot of work if we just filtered it out much earlier.

alacritty_terminal/src/ansi.rs
```diff
@@ -338,9 +338,22 @@ pub enum CursorStyle {
     Hidden,
 }
 
+impl Default for CursorShape {
+    fn default() -> CursorShape {
+        CursorShape::Block
+    }
+}
+
+/// Describes style of cursor (shape + blinking).
```
chrisduerr, line 347: This comment doesn't really provide any value. `Describes the style of cursor` is the same as `CursorStyle` and the comment of `shape + blinking` is just all the fields. We shouldn't document fields because those might change, so that's actually actively harmful. I think it would be better to just delete this rather than using it as-is.

alacritty_terminal/src/ansi.rs
```diff
@@ -1076,11 +1090,12 @@ where
...
-                let style = match arg_or_default!(idx: 0, default: 0) {
+                let ps = arg_or_default!(idx: 0, default: 0);
+                let style = match ps {
                     0 => None,
-                    1 | 2 => Some(CursorStyle::Block),
-                    3 | 4 => Some(CursorStyle::Underline),
-                    5 | 6 => Some(CursorStyle::Beam),
+                    1 | 2 => Some(CursorStyle { shape: CursorShape::Block, blinking: ps == 1 }),
+                    3 | 4 => Some(CursorStyle { shape: CursorShape::Underline, blinking: ps == 3 }),
+                    5 | 6 => Some(CursorStyle { shape: CursorShape::Beam, blinking: ps == 5 }),
```
chrisduerr, line 1098: I feel like the additional conditional is just unnecessarily complicated. Expanding this to struct to have separate entries for blinking/non-blinking probably makes it a ton easier to just read without having to think about what is going on.

alacritty_terminal/src/config/mod.rs
```diff
@@ -128,6 +128,8 @@ pub struct Cursor {
     pub style: CursorStyle,
```
chrisduerr, line 128: We can't just change things, we need to still support the old format and print deprecation warnings.

alacritty_terminal/src/event.rs
```diff
@@ -14,6 +14,9 @@ pub enum Event {
     Wakeup,
     Bell,
     Exit,
+    CursorStartBlinking,
+    CursorBlink,
```
chrisduerr, line 18: I don't think this is a good name, since `Blink` has no clear indication of what it should do and thus would make it difficult to implement. It should instead be something like `ToggleCursorVisibility`.

alacritty_terminal/src/term/mod.rs
```diff
@@ -58,6 +58,8 @@ struct RenderableCursor {
 pub struct CursorKey {
     pub style: CursorStyle,
     pub is_wide: bool,
+    /// For Blinking* cursors, true if the cursor is currently hidden
+    pub blinking_out: bool,
```
chrisduerr, line 62: This variable name is incredibly confusing, could we go with something more descriptive. A cursor being "out" is not something I would understand, so other people would probably have even more of a problem.

alacritty_terminal/src/term/mod.rs
```diff
@@ -687,6 +689,9 @@ pub struct Term<T> {
 
     pub selection: Option<Selection>,
 
+    /// Blinking state of the current cursor, true is hidden (used only blinking cursor styles)
+    pub cursor_blinking_out: bool,
```
chrisduerr, line 693: This variable name should be changed.

alacritty_terminal/src/term/mod.rs
```diff
@@ -830,6 +836,29 @@ impl<T> Term<T> {
...
         }
+
+        self.update_blinking_cursor_context();
+    }
+
+    /// Depending on current mode and cursor configuration, decide whether to start or stop
+    /// blinking the cursor.
+    ///
+    /// Typically used after a configuration update or vi_mode toggle.
+    pub fn update_blinking_cursor_context(&self)
```
chrisduerr, line 847: I'm questioning the API ergonomics of how this is supposed to work. The terminal instructs the UI when blinking should start or end, then the UI tells the terminal to tell the UI to update it about how things should work? That seems insanely convoluted.

Why can't the UI just figure this out itself?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1293,9 +1322,9 @@ impl<T> Term<T> {
             !self.mode.contains(TermMode::SHOW_CURSOR) || point.line >= self.screen_lines();
         let cursor_style = if hidden && !vi_mode {
             point.line = Line(0);
-            CursorStyle::Hidden
+            CursorStyle { shape: CursorShape::Hidden, blinking: false }
         } else if !self.is_focused && config.cursor.unfocused_hollow() {
-            CursorStyle::HollowBlock
+            CursorStyle { shape: CursorShape::HollowBlock, blinking: false }
```
chrisduerr, line 1327: The unfocused hollow block cursor never blinks, but it seems like other cursors still blink even without focus.

I think if Alacritty isn't focused, the cursor shouldn't blink. I wouldn't even make this configurable, since it's just a waste of resources really.

Dettorer: Indeed, disabling blinking while unfocused also is the behavior of XTerm and rxvt, I'll try to make it work too.

alacritty_terminal/src/term/mod.rs
Dettorer: I didn't find a better way to do that, but that's my first contribution so if you think that's convoluted there must be an other way. I'll look into it and ask question on IRC if needed.

alacritty_terminal/src/term/mod.rs
chrisduerr: It might be a bit of an indication that the way this is done isn't ideal.

I don't quite remember all our previous conversations, but have you tried storing the cursor visibility for blinking outside of alacritty_terminal? Currently I don't see why it's needed inside of it.

alacritty/src/event.rs
chrisduerr: That sounds like a good idea anyways, since we want to keep our event handler functionally simple and delegate to methods for anything even slightly complex.

alacritty_terminal/src/ansi.rs
chrisduerr: We can keep it at `Describes the style of the cursor.` I don't really care either way. But I wouldn't remove existing documentation for this.

alacritty_terminal/src/ansi.rs
chrisduerr: Instead of `1 | 2 => ps == 1`, you'd just do `1 => true, 2 => false`. Highly simplified, but hopefully you get the gist.

alacritty_terminal/src/event.rs
chrisduerr: Sure, that sounds like a good idea.

yuri1969:
I got feedback using 1adcb53 - when you launch an interactive tool (such as `vi` or `ranger`) the cursor stops blinking and remains so after you return form the app.

alacritty_terminal/src/ansi.rs
nixpulvis: Something else is wrong around this, since I get the error: "[ERROR] ... Problem with config: missing field \`shape\`; using default" when I insert the following configuration I would expect to work:

```yaml
cursor:
  style:
    blinking: true
```

chrisduerr:
@Dettorer Please let me know when you need another review, that way I don't have to check every time if all the comments have been addressed. Just to make sure you're not waiting for a review while I'm waiting for you to change things.

Dettorer:
Yeah, sorry, I'm planning on marking the opened conversations as resolved as I go fixing or implementing them. I'll ask again explicitly for a review once it's done.

Although I should mention that I might not be able to work on this for a month or two (my job is about to get very time consuming). If someone wants/needs to take over I'm totally ok with it, but I'll also be happy to finish this PR once things calm down on my side.

jasonjckn:
looks like the work here is nearly done? move it to the finish line?

Dettorer:
Feel free to do so if you want to, but I should be able to start working on this again pretty soon.

PouriDev:
Any progress on this PR?

Dettorer:
Ok sorry, the sanitary situation made my job way more time consuming than I anticipated and I keep forgetting to give updates here. I should stop saying "I should be available soon", it's becoming a pattern.

I suggest you consider me out of this for the foreseeable future. I'll still be happy to continue my work on this when I have the time if nobody had a go at it, but I can't say when that'll be.

Again, I should be able to answer some questions from time to time though if someone wants to understand my drafts.

kchibisov:
I think @chrisduerr was willing to pick up this after we do 0.6.0 release, if not I can pick up it.

chrisduerr:
> I think @chrisduerr was willing to pick up this after we do 0.6.0 release, if not I can pick up it.

Will be absolutely trivial to do at the beginning of 0.7.0, yes. Especially since we already have cursor hiding logic in place now.

alacritty_terminal/src/ansi.rs
```diff
@@ -868,18 +875,22 @@ where
                 unhandled(params);
             },
 
+            // TODO: 50 still supported? https://iterm2.com/documentation-escape-codes.html
```
chrisduerr, line 878: We should probably keep support for this I'd assume, even though 1337 should be used instead? I'm not sure which application uses this escape.

kchibisov: 1337 sounds like iterm

chrisduerr:
@kchibisov I think this should be ready for review.

chrisduerr:
One thing I've noticed: It seems like a `reset` after blinking has been disabled with an escape sequence doesn't restore the blinking. That still needs to be investigated.

alacritty/src/event.rs
```diff
@@ -1024,13 +1068,26 @@ impl<N: Notify + OnResize> Processor<N> {
...
+                                    let style = config.cursor.style();
+                                    config.cursor.vi_mode_style().unwrap_or(style)
+                                } else {
+                                    config.cursor.style()
+                                };
+                                processor.ctx.update_cursor_blinking(cursor_style.blinking);
                             } else {
                                 processor.ctx.window.set_mouse_visible(true);
+
+                                processor.ctx.update_cursor_blinking(false);
```
kchibisov, line 1090: You unconditionally enable blinking even if the blinking cursor mode was unset before, on focus change, this is noticeable in e.g. weechat which disables blinking completely, and after refocus the cursor starts blinking, which shouldn't be. You should track whether the current mode/cursor allows the cursor blinking in a first place, to handle unset of the mode properly.

alacritty/src/input.rs
```diff
@@ -890,6 +898,13 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
             return;
         }
 
+        // Disable cursor blinking while the user is typing.
+        let cursor_blink_rate = self.ctx.config().cursor.blink_rate;
+        if let Some(timer) = self.ctx.scheduler_mut().get_mut(TimerId::BlinkCursor) {
+            timer.deadline = Instant::now() + Duration::from_millis(cursor_blink_rate);
+            self.ctx.show_cursor();
+        }
```
kchibisov, line 906: Need something like that for Vi motions, since right now it's not really usable in it, since cursor constantly blinking while you're moving and you don't see the cursor.

kchibisov (changes requested):
I've mentioned on IRC, that I think it makes sense to provide an option to disable blinking or disable unsetting blinking cursor(all ncurses apps and some applications don't provide an option to have blinking cursor in them). However I'm not sure how many terminals allow those things, so requires checking some of them.

alacritty_terminal/src/config/mod.rs
```diff
@@ -149,6 +161,7 @@ impl Default for Cursor {
             vi_mode_style: Default::default(),
             thickness: Percentage::new(DEFAULT_CURSOR_THICKNESS),
             unfocused_hollow: Default::default(),
+            blink_rate: 666,
```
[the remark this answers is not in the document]
chrisduerr: I personally find neither 750 nor 500 ideal. So I picked the next best thing in between. Though I might have just been biased based on the other terminals I tested.

alacritty.yml
```diff
@@ -338,12 +338,17 @@
...
+      # Cursor shape
+      #
+      # Values for `shape`:
+      #   - ▇ Block
+      #   - _ Underline
+      #   - | Beam
+      #shape: Block
+
+      # Default cursor blinking state.
+      #blinking: false
```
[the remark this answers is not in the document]
chrisduerr: It's already documented in the comment above vi mode?

alacritty_terminal/src/config/mod.rs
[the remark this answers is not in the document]
kchibisov: I think it's nice, since I don't even like the default you've picked, and I'd image that it'll be hard to pick a reasonable one, since you like 666, but I don't find it good. Also, even gtk allows changing it.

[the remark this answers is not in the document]
chrisduerr: Fair enough. Which would be your preferred blink rate? I don't really like blinking to begin with.

alacritty/src/event.rs
```diff
@@ -495,6 +497,38 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
+    #[inline]
+    fn update_cursor_blinking(&mut self) {
+        // Get config cursor style.
+        let mut cursor_style = self.config.cursor.style;
+        if self.terminal.mode().contains(TermMode::VI) {
+            cursor_style = self.config.cursor.vi_mode_style.unwrap_or(cursor_style);
+        };
+
+        // Check terminal cursor style.
+        let blinking = cursor_style.blinking(self.terminal.cursor_style().blinking);
```
chrisduerr, line 510: I'm not sure if I like this API. The `cursor_style.blinking(current_state)` thing is a bit weird, but that's the easiest API for us in Alacritty.

I think the other alternative would be a bit nicer, which would be `cursor_style.blinking_override().unwrap_or(terminal_state)`, but it's not as concise for our usage.

What do you think @kchibisov?

chrisduerr:
@kchibisov I think all your suggestions should be integrated now. The only question left is if we want to do anything about the `blink_interval` I suppose. It should be trivial to make that cursor-specific, but I'm not sure if there's any benefit to that?

alacritty.yml
```diff
@@ -353,6 +358,9 @@
   # See `cursor.style` for available options.
   #vi_mode_style: None
 
+  # Cursor blinking interval in milliseconds.
+  #blink_interval: 500
```
kchibisov, line 362: I think 700 or 650 should be a bit better for default, since they are no that annoying like 500.

alacritty/src/event.rs
```diff
@@ -495,6 +497,56 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
+    }
+
+    /// Enable input mode.
+    ///
+    /// Putting the user in input mode will temporarily disable some features like terminal cursor
+    /// blinking or the mouse cursor.
+    ///
+    /// This mode will disable itself automatically.
+    #[inline]
+    fn enable_input_mode(&mut self) {
```
kchibisov, line 535: I'm not sure if I like `mode` here, maybe we should use something like `fn started_typing(&mut self)`. Since each time I see input mode I think it's IME stuff.

docs/escape_support.md
```diff
@@ -81,21 +81,21 @@ brevity.
...
+| `OSC 4`    | IMPLEMENTED |                                                    |
+| `OSC 10`   | IMPLEMENTED |                                                    |
+| `OSC 11`   | IMPLEMENTED |                                                    |
+| `OSC 12`   | IMPLEMENTED |                                                    |
+| `OSC 50`   | IMPLEMENTED | Only `CursorShape` is supported                    |
+| `OSC 52`   | IMPLEMENTED | Only Clipboard and primary selection supported     |
+| `OSC 104`  | IMPLEMENTED |                                                    |
+| `OSC 110`  | IMPLEMENTED |                                                    |
+| `OSC 111`  | IMPLEMENTED |                                                    |
+| `OSC 112`  | IMPLEMENTED |                                                    |
```
kchibisov, line 98: Seems like alignment wasn't reverted after you've removed OSC 1337.

alacritty/src/event.rs
```diff
@@ -495,6 +497,56 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
+        self.scheduler.unschedule(TimerId::BlinkCursor);
+        if blinking && self.terminal.is_focused {
+            self.scheduler.schedule(
+                GlutinEvent::UserEvent(Event::BlinkCursor),
+                Duration::from_millis(self.config.cursor.blink_interval),
+                true,
+                TimerId::BlinkCursor,
+            )
+        } else {
+            *self.cursor_hidden = false;
```
kchibisov, line 523: If you have blinking `On` and then you do `printf "\e[?12l\a"`, it's still blinking, but it shouldn't. However it seems like vte literally does nothing on Blinking cursor mode, so maybe it's fine? Like it neither disables it nor enables it, and blinking can only be changed via `CSI Ps SP q`, maybe we should follow it as well? Have you checked how it works in xterm/urxvt?

kchibisov (changes requested):

alacritty_terminal/src/term/mod.rs
```diff
@@ -2095,6 +2109,8 @@ impl<T: EventListener> Handler for Term<T> {
         self.title = None;
         self.selection = None;
         self.regex_search = None;
+
```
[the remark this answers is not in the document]
chrisduerr: Separate PR, has nothing to do with this.

alacritty/src/event.rs
chrisduerr: Just a bug.

alacritty_terminal/src/event.rs
```diff
@@ -11,6 +11,8 @@ pub enum Event {
     ResetTitle,
     ClipboardStore(ClipboardType, String),
     ClipboardLoad(ClipboardType, Arc<dyn Fn(&str) -> String + Sync + Send + 'static>),
+    // TODO: This really isn't a great name?
+    CursorBlinking(bool),
```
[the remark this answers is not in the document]
chrisduerr: I think `CursorBlinkingChange` should be a bit better.

alacritty/src/event.rs
chrisduerr: I feel like a `started_typing` method should return a `bool` or something? I don't like that name since it doesn't really help me understand that it does.

alacritty_terminal/src/config/mod.rs
```diff
@@ -121,9 +121,11 @@ impl Default for EscapeChars {
 #[derive(Deserialize, Copy, Clone, Debug, PartialEq)]
 pub struct Cursor {
     #[serde(deserialize_with = "failure_default")]
-    pub style: CursorStyle,
+    pub style: ConfigCursorStyle,
     #[serde(deserialize_with = "option_explicit_none")]
-    pub vi_mode_style: Option<CursorStyle>,
+    pub vi_mode_style: Option<ConfigCursorStyle>,
+    #[serde(deserialize_with = "failure_default")]
+    pub blink_interval: u64,
```
kchibisov, line 128: Do we really want to keep `u64` here? I don't see any reason to use it over something like `u32` or `u16` for blinking.

kchibisov (changes requested):

alacritty/src/event.rs
```diff
@@ -1024,15 +1087,17 @@ impl<N: Notify + OnResize> Processor<N> {
                     },
                     WindowEvent::Focused(is_focused) => {
                         if window_id == processor.ctx.window.window_id() {
-                            processor.ctx.terminal.is_focused = is_focused;
-                            processor.ctx.terminal.dirty = true;
+                            let terminal = &mut processor.ctx.terminal;
+                            terminal.is_focused = is_focused;
+                            terminal.dirty = true;
```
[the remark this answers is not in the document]
chrisduerr: Yes, still a leftover from the previous commits.

alacritty.yml
```diff
@@ -353,6 +364,9 @@
   # See `cursor.style` for available options.
   #vi_mode_style: None
 
+  # Cursor blinking interval in milliseconds.
+  #blink_interval: 750
```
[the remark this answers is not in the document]
chrisduerr: You mean force clamp it to be at least 50ms? I'm not sure if that actually provides any real benefit?

It's not like your computer suddenly explodes when you put 0 as the interval. Is this realistically something that can be messed up by accident?

alacritty_terminal/src/config/mod.rs
chrisduerr: There is a reason: The duration conversion takes a u64. So one way or another we need to convert it to u64 anyways at the only place we use it. So the best option is to just define it as u64 to begin with.

I see no benefit to define it as u16 when we then convert it later anyways. It's not like it's saving us a measurable amount of memory.

alacritty.yml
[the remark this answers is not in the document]
kchibisov: I mean I'm not sure what to expect from interval 0, from what I can see it shouldn't blink at all, however it blink, since we theoretically can't provide an interval of zero, so I was thinking to clamping it to a value that we can handle, like 50ms or something. We can handle 4-5ms though I guess.

But we can keep it as 0 since I don't think it could mess up anything.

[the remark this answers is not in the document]
chrisduerr: I suppose it might be helpful to limit it just for debugging purposes. Since a rate of `0` ms might flood the event queue and cause problems. So I think I'll just limit it to something like `10ms` to make sure other than what is visible to the user, we don't run into any internal issues.

kchibisov (approved):

chrisduerr (approved):

PR #3935, opened 2020-07-09
Hard require system's Fontconfig on Linux/BSD

kchibisov:
`CHANGELOG.md` file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Minimum Rust version has been bumped to 1.41.0
 - Prebuilt Linux binaries have been removed
 - Added manpage, terminfo, and completions to macOS application bundle
+- On Linux/BSD the build will fail without Fontconfig installed, instead of building it from source
 
 ### Removed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Minimum Rust version has been bumped to 1.41.0
 - Prebuilt Linux binaries have been removed
 - Added manpage, terminfo, and completions to macOS application bundle
+- The system's Fontconfig is a hard requirement now on Linux/BSD
```
chrisduerr, line 14: Is it any fontconfig version, or do we require a specific minimum version? If so, it might be nice to state that.

Also I'd slightly reword it:

```
- On Linux/BSD the build will fail without fontconfig installed, instead of building it from source

kchibisov: >Is it any fontconfig version, or do we require a specific minimum version?

I'm not sure what we require actually, so let's just say, any for now(which is likely not true in general). But we don't use any new(??) fancy features anyway. I guess it'll make sense to just update it on demand honestly.

chrisduerr: Sure, sounds good to me. If there are any reports that fontconfig is present but too outdated, we can still be more specific.

PR #3936, opened 2020-07-09
Bump urlocator to 0.1.4

chrisduerr:


chrisduerr:
Bumped it and switched up the commit message to be more clear about the relevant updates that have an effect on us.

PR #3952, opened 2020-07-11
Add support for searching without vi mode

### LINKED ISSUE #3937: Support interaction with search without vi mode
Currently it is required to enter vi mode to interact with the search to jump to previous/next matches.
However, users might not want to ever touch vi mode but still interact with search.

If search is started without vi mode active, the "active" search match should always be selected so it can be easily copied.
Then whenever the SearchNext/SearchPrevious actions are executed after search is confirmed,
the selection should move to the next/previous match and the viewport should be moved.

chrisduerr:
This implements search without vi mode by using the selection to track
the active search match and advancing it on user input. The keys to go
to the next or previous match are not configurable and are bound to
enter and shift enter based on Firefox's behavior.

Fixes #3937.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -671,6 +671,7 @@
   #- { key: W,      mods: Command,            action: Quit             }
   #- { key: N,      mods: Command,            action: SpawnNewInstance }
   #- { key: F,      mods: Command|Control,    action: ToggleFullscreen }
+  #- { key: F,      mods: Command,            action: Search           }
 
 #debug:
   # Display the time it takes to redraw each frame.
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -480,8 +480,8 @@
 # - `action`: Execute a predefined action
 #
 #   - ToggleViMode
-#   - Search
-#   - SearchReverse
+#   - SearchForwards
+#   - SearchBackwards
 #   - Copy
 #   - Paste
 #   - PasteSelection
@@ -636,14 +636,16 @@
   #- { key: W,      mods: Shift,         mode: Vi, action: WordRight               }
   #- { key: E,      mods: Shift,         mode: Vi, action: WordRightEnd            }
   #- { key: Key5,   mods: Shift,         mode: Vi, action: Bracket                 }
-  #- { key: Slash,                       mode: Vi, action: Search                  }
-  #- { key: Slash,  mods: Shift,         mode: Vi, action: SearchReverse           }
+  #- { key: Slash,                       mode: Vi, action: SearchForwards          }
+  #- { key: Slash,  mods: Shift,         mode: Vi, action: SearchBackwards         }
   #- { key: N,                           mode: Vi, action: SearchNext              }
   #- { key: N,      mods: Shift,         mode: Vi, action: SearchPrevious          }
 
   # (Windows, Linux, and BSD only)
   #- { key: V,        mods: Control|Shift,           action: Paste            }
   #- { key: C,        mods: Control|Shift,           action: Copy             }
+  #- { key: F,        mods: Control|Shift,           action: SearchForwards   }
+  #- { key: B,        mods: Control|Shift,           action: SearchBackwards  }
   #- { key: C,        mods: Control|Shift, mode: Vi, action: ClearSelection   }
   #- { key: Insert,   mods: Shift,                   action: PasteSelection   }
   #- { key: Key0,     mods: Control,                 action: ResetFontSize    }
@@ -671,7 +673,8 @@
   #- { key: W,      mods: Command,            action: Quit             }
   #- { key: N,      mods: Command,            action: SpawnNewInstance }
   #- { key: F,      mods: Command|Control,    action: ToggleFullscreen }
-  #- { key: F,      mods: Command,            action: Search           }
+  #- { key: F,      mods: Command,            action: SearchForwards   }
+  #- { key: B,      mods: Command,            action: SearchBackwards  }
 
 #debug:
   # Display the time it takes to redraw each frame.
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -145,7 +145,7 @@
 
   # Glyph offset determines the locations of the glyphs within their cells with
   # the default being at the bottom. Increasing `x` moves the glyph to the right,
-  # increasing `y` moves the glyph upwards.
+  # increasing `y` moves the glyph upward.
   #glyph_offset:
   #  x: 0
   #  y: 0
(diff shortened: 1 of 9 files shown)


alacritty/src/event.rs
```diff
@@ -91,8 +91,8 @@ pub struct SearchState {
     /// Change in display offset since the beginning of the search.
     display_offset_delta: isize,
 
-    /// Vi cursor position before search.
-    vi_cursor_point: Point,
+    /// Search origin in viewport coordinates relative to original display offset.
+    origin: Point,
```
chrisduerr, line 95: I've decided to rename this from the `vi_cursor_point` to just `origin` since it never really tracked the vi point to begin with. Even before this PR it was tracking the origin of the search which was unrelated to the vi cursor position. But since this PR changes search around anyways I've decided to include it here.

kchibisov:
Reverse search doesn't work. By that I mean that it seems like there's no difference between reverse and normal search.

kchibisov:
If you scroll up, the search will continue from visual point in a buffer, the original point was. However I'm expecting that it should continue from the top(if you scrolled down) or bottom (if you scolled up)?

It's also unclear how RMB selection works and whether normal selection should be allowed here. It seems like RMB work unreliable and you can't select, however I'd expect the use case when someone decided to scroll search for something and select things around, which is not something possible now? I feel like just selecting stuff should still work, but since you're using selection to track active match it could be problematic to make it work.

chrisduerr:
> If you scroll up, the search will continue from visual point in a buffer, the original point was. However I'm expecting that it should continue from the top(if you scrolled down) or bottom (if you scolled up)?

If I understand you correctly, that is actually a feature. It works the same way in Firefox and I think it makes sense to keep track of the search origin like this.

> It's also unclear how RMB selection works and whether normal selection should be allowed here. It seems like RMB work unreliable and you can't select, however I'd expect the use case when someone decided to scroll search for something and select things around, which is not something possible now? I feel like just selecting stuff should still work, but since you're using selection to track active match it could be problematic to make it work.

I don't think that makes a ton of sense, or is something we need to support. Searching should focus on searching a thing, not searching one thing then selecting another. If the user wants to manually select something then they can just leave search and do it after that's done?

chrisduerr:
@kchibisov I've fixed both issues and things should be working properly now.

kchibisov:
Should we add some documentation about Vi less search and implicit bindings it has? I guess we should, since it's not something obvious, and I'd really like to bind it to something. We can go for `Ctrl + Shift + F` or `Ctrl + Shift + H`.

chrisduerr:
Where would you like this documentation to be?

kchibisov:
in key bindings section, mentioning that if you have `Search` or `ReverseSearch` active without vi mode, you'll have additional bindings on `Enter`/`Shift + Enter`.

chrisduerr:
Should things be documented in the key bindings section if they're not keybindings that can be changed? I'm not sure that's the right place and I don't see how people would find it.

alacritty.yml
```diff
@@ -671,6 +671,7 @@
   #- { key: W,      mods: Command,            action: Quit             }
   #- { key: N,      mods: Command,            action: SpawnNewInstance }
   #- { key: F,      mods: Command|Control,    action: ToggleFullscreen }
+  #- { key: F,      mods: Command,            action: Search           }
```
chrisduerr, line 674: I still need to test this on macOS hardware, just to make sure it doesn't cause any problems.

I don't think there's any obvious default binding for reverse search.

kchibisov:
>Should things be documented in the key bindings section if they're not keybindings that can be changed? I'm not sure that's the right place and I don't see how people would find it.

Not sure either, but on the other those are key bindings to some point, just builtin.

alacritty.yml
talios: Just tested this on my OS/X alacritty install - cmd-f seems to enter search mode, tho I can't seem to find anyway OUT of search mode?   up/down seems to cycle thru my shell history, but <enter> doesn't seem to process/execute anything - have to restort to closing the window...

talios: Actually - this was tested using a `master` build, not this PR.

chrisduerr: Escape should leave the vi mode. Command + F is identical to what Firefox uses on macOS, correct? Are you aware of any standard binding for searching in the opposite order? I know ctrl + command + F is already used for fullscreen, is command + shift + F already in use?

chrisduerr:
> Not sure either, but on the other those are key bindings to some point, just builtin.

Yes, but I just don't think anyone would find it there anyways. So I'm not sure if it makes sense to document it there. Especially because it might be a relatively big block.

kchibisov:
>Yes, but I just don't think anyone would find it there anyways. So I'm not sure if it makes sense to document it there. Especially because it might be a relatively big block.

I'd assume that the right place for that is somewhere else and not in a config.

alacritty.yml
talios: I just pulled the viless PR/branch which I didn't have locally and it seems to work fine for me.

I'd added the binding to my own alacritty.yaml file.  There's no current binding for command+shift+F - I see above theres a reference to `ReverseSearch` which doesn't seem to exist?

alacritty-viless-search

talios: (Removed my custom binding and the default comes along fine.)

chrisduerr: I haven't added a binding to command+shift+f, because I'm not sure if it is used by any other application. I don't want to add something that is no-standard.

talios: I get you!  iTerm2 and the standard OS/X terminal use ctrl-f for find, ctrl-g for find-next and ctrl-shift-g for find previous - which seems to be mapped to enter/shift-enter in this patch?  I think I kinda like enter/shift-enter better - much faster to use anyway.

chrisduerr: Yes, Enter/Shift+Enter is taken from Firefox's bindings. I think that makes sense to follow.

talios: I just want to say - damnit, I've used this new search so much today I'm wondering how I survived without it. Look forward to this hitting the mainline.

alacritty.yml
```diff
@@ -480,8 +480,8 @@
 # - `action`: Execute a predefined action
 #
 #   - ToggleViMode
-#   - Search
-#   - SearchReverse
+#   - SearchForwards
+#   - SearchBackwards
```
kchibisov, line 484: I guess they should be without `s` on the end? Just `SearchForward` and `SearchBackward`.

PR #3955, opened 2020-07-12
Add a config option to disable vsync

aruhier:
Disabling vsync can be useful to improve input latency.  In some cases, having vsync in alacritty can be redundant, for example when using a compositor that already handles vsync.

Default is to let the vsync on, but can be explicitely disabled by the user in the config via the `window.disable_vsync` option.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -78,6 +78,13 @@
   # Set this to `None` to use the default theme variant.
   #gtk_theme_variant: None
 
+  # Disable vsync (changes require restart)
+  #
+  # Force to disable vsync, what can improve the input latency but can also add tearing.
+  # A worse throughput performance is also expected by disabling the vsync (about 35% slower).
+  # On Wayland, vsync is always disabled, setting this option has no effect.
+  #disable_vsync: false
+
 #scrolling:
   # Maximum number of lines in the scrollback buffer.
   # Specifying '0' will disable scrolling.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for `gopher` and `gemini` URLs
 - Unicode 13 support
 - Option to run command on bell which can be set in `bell.command`
+- Option `window.disable_vsync` to explicitly disable vsync
 
 ### Changed
 
--- a/alacritty/src/config/window.rs
+++ b/alacritty/src/config/window.rs
@@ -58,6 +58,10 @@ pub struct WindowConfig {
     /// Use dynamic title.
     #[serde(default, deserialize_with = "failure_default")]
     dynamic_title: DefaultTrueBool,
+
+    /// Disable vsync.
+    #[serde(deserialize_with = "failure_default")]
+    pub disable_vsync: bool,
 }
 
 pub fn default_title() -> String {
@@ -90,6 +94,7 @@ impl Default for WindowConfig {
             gtk_theme_variant: Default::default(),
             title: default_title(),
             dynamic_title: Default::default(),
+            disable_vsync: Default::default(),
         }
     }
 }
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -604,7 +604,7 @@ impl Display {
 
         #[cfg(not(any(target_os = "macos", windows)))]
         {
-            if self.is_x11 {
+            if self.is_x11 && !config.ui_config.window.disable_vsync {
                 // On X11 `swap_buffers` does not block for vsync. However the next OpenGl command
                 // will block to synchronize (this is `glClear` in Alacritty), which causes a
                 // permanent one frame delay.
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -159,9 +159,9 @@ impl Window {
 
         // Disable vsync on Wayland.
         #[cfg(not(any(target_os = "macos", windows)))]
-        let vsync = !event_loop.is_wayland();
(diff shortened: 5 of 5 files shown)


chrisduerr:
Disabling vsync would mainly waste a lot of system resources. What exactly do you mean by compositor handling vsync, are you talking about X11 or Sway? Our vsync synchronizes with the compositor, it is not completely separate.

aruhier:
I'm talking about X11. For sway and other wayland compositors, I understand that alacritty already handles that well, and it has way less visible input lag for me on wayland that on X11.

On Xorg, when you have a compositor like compton/picom with vsync enabled, isn't it doing a redundant job with alacritty? Both will buffer frames and if I'm not wrong, they don't know about the vsync capability of each other.

Kitty allows to disable vsync for example (their reason is to lower the input lag on high refresh rate monitors, but I don't have any benchmark), although they also have a framerate limit. Personally I didn't see a high CPU nor GPU usage by alacritty when printing a large number of lines with vsync disabled, but I have a 144Hz display.

If I'm wrong of if it's not worth it, please feel free to close this PR if you have a better solution in mind. Also a benchmark could be useful to measure if the trade-off is a good one or totally garbage, I'll look into that.

aruhier:
I used typometer to run some benchmarks on X11: https://up2sha.re/file?f=QkTs8hqNCXckLuGdJM

I did multiple tests on a 144Hz monitor then on 60Hz:
* Without vsync (and compositor without vsync)
* With vsync (and compositor without vsync)
* With vsync without compositor
* With vsync without compositor

I cannot use the vsync in my compositor (I use picom, dotfiles here: https://github.com/aruhier/dotfiles/blob/tour-anthony/compton/compton.conf) as it's broken on my setup. If I enable it, it uses a 60Hz refreshrate for everything instead of 144Hz.

Here are the results:
```
# Title	Min	Max	Avg	SD
1 Without vsync                     11.0    62.3    28.5    9.0
2 Without vsync                     13.1    55.1    25.8    8.3
3 Without vsync                     8.3     52.1    24.7    7.4
4 With vsync                        17.9    53.4    30.6    6.6
5 With vsync                        18.8    58.3    31.8    7.2
6 With vsync                        20.3    77.9    35.4    8.5
7 With vsync no compositor          13.9    39.5    23.9    4.4
8 With vsync no compositor          18.4    36.2    25.1    4.4
9 Without vsync no compositor       7.2     28.7    13.2    4.3
10 Without vsync no compositor      6.9     29.7    14.7    5.1
11 Without vsync 60Hz               13.9    58.8    27.5    8.6
12 With vsync 60Hz                  33.4    74.8    50.3    7.6
13 With vsync 60Hz                  30.1    77.1    49.4    8.2
14 With vsync 60Hz no compositor    29.2    62.3    40.3    8.3
15 With vsync 60Hz no compositor    29.9    49.4    39.5    8.2
```

It's not that much of a gain on 144Hz, but the input lag seems to be way lower on a 60Hz display (which makes sense). The best results are anyway on without vsync and without compositor.

alacritty.yml
```diff
@@ -78,6 +78,12 @@
   # Set this to `None` to use the default theme variant.
   #gtk_theme_variant: None
 
+  # Disable vsync (changes require restart)
+  #
+  # Force to disable vsync, what can improve the input latency but can also add tearing.
+  # On Wayland, vsync is always disabled, setting this option has no effect.
+  #disable_vsync: false
```
chrisduerr, line 85: I think having this option as just `vsync` would make more sense.

alacritty.yml
```diff
@@ -78,6 +78,12 @@
   # Set this to `None` to use the default theme variant.
   #gtk_theme_variant: None
 
+  # Disable vsync (changes require restart)
+  #
+  # Force to disable vsync, what can improve the input latency but can also add tearing.
+  # On Wayland, vsync is always disabled, setting this option has no effect.
```
chrisduerr, line 84: We should state that this will decrease performance significantly. We're talking like ~35% in some basic benchmarks.

kchibisov:
In general I don't like that option and solution. The thing I'd consider is frame scheduling, so you draw once, but closer to vblank. On Wayland there's an explicit mechanism for that, on other platforms it should be done manually.

alacritty.yml
aruhier: I also wanted to call it `vsync`, but it was a bit misleading, because on Wayland for example, even if `vsync` is set to `true`, it's not gonna be forced.
That's why I called it `disable_vsync`.

alacritty.yml
aruhier: Done.

aruhier:
@kchibisov: thanks for your comment.

I'm not sure I will be able to do that, I don't have any sufficient knowledge in OpenGL, but I'll try to have a look at it.
I think we can close this PR then.

chrisduerr:
I don't think there's any OpenGL involved. It's just a matter of getting the refresh rate of the monitor and syncing with that manually, rather than blocking on OpenGL calls. So it should involve *less* OpenGL. Getting the refresh rate wouldn't be a problem, that's already available. The thing one would have to figure out is just when exactly the next vblank happens. Of course one could also just time it from the last vblank. After that it's just a question of setting a reasonable timeout before the vblank since we need some time to render ourselves and the scheduling should then work independently.

kchibisov:
All browsers should have infra to do that, some of them are not perfect, but you can get an inspiration from them. They should do that to handle https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame nicely(looks very similar to Wayland's frame callbacks mechanism).

chrisduerr:
I've opened https://github.com/alacritty/alacritty/issues/3972 to track this.

PR #3982, opened 2020-07-16
Add different search label for backward search

chrisduerr:


alacritty/src/event.rs
```diff
@@ -83,10 +83,10 @@ impl From<TerminalEvent> for Event {
 /// Regex search state.
 pub struct SearchState {
     /// Search string regex.
-    regex: Option<String>,
+    pub regex: Option<String>,
 
     /// Search direction.
-    direction: Direction,
+    pub direction: Direction,
```
chrisduerr, line 89: Having the search state public and modifiable doesn't make a ton of sense.

alacritty/src/display.rs
```diff
@@ -639,14 +645,15 @@ impl Display {
         size_info: &SizeInfo,
         message_bar_lines: usize,
         search_regex: Option<String>,
+        search_label: &str,
```
chrisduerr, line 648: Two parameters for basically the same thing is complicated.

alacritty/src/display.rs
```diff
@@ -639,14 +645,15 @@ impl Display {
...
     ) {
         let search_regex = match search_regex {
             Some(search_regex) => search_regex,
             None => return,
         };
         let glyph_cache = &mut self.glyph_cache;
 
-        let label_len = SEARCH_LABEL.len();
+        let label_len = search_label.len();
```
chrisduerr, line 656: The label's len is previously used for the IME already. Computing it again here is wasteful.

alacritty/src/display.rs
```diff
@@ -474,12 +474,18 @@ impl Display {
             None
         };
 
+        let search_regex = search_state.regex.as_ref().map(|regex| Self::format_search(&regex));
+        let search_label = match search_state.direction {
+            Direction::Right => FORWARD_SEARCH_LABEL,
+            Direction::Left => BACKWARD_SEARCH_LABEL,
+        };
```
chrisduerr, line 481: Computing this even if search isn't active is wasteful.

alacritty/src/display.rs
```diff
@@ -474,12 +474,18 @@ impl Display {
...
+
         // Update IME position.
         #[cfg(not(windows))]
         {
             let point = match &search_regex {
                 Some(regex) => {
-                    let column = min(regex.len() + SEARCH_LABEL.len() - 1, terminal.cols().0 - 1);
+                    let column = min(regex.len() + search_label.len() - 1, terminal.cols().0 - 1);
                     Point::new(terminal.screen_lines() - 1, Column(column))
                 },
```
chrisduerr, line 490: I'd like to move this to a different location maybe. It would be possible to just store the grid's cursor point, drop the terminal and then update the IME at a later point. Since the cursor is cheap to copy that shouldn't be a problem.

That way we could then maybe update the IME in a branch together with search, so we wouldn't have to check that search is enabled in draw_search but would just pass the desugared search `String` (or `&str`) without `Option` directly to `draw_search`. This could also make it possible to maybe already get rid of the search label, but that might be tough.

alacritty/src/display.rs
```diff
@@ -592,9 +577,28 @@ impl Display {
...
+                let search_label = match search_state.direction() {
+                    Direction::Right => FORWARD_SEARCH_LABEL,
+                    Direction::Left => BACKWARD_SEARCH_LABEL,
+                };
+
+                let regex = Self::format_search(&size_info, regex, search_label);
+
+                self.draw_search(config, &size_info, message_bar_lines, &regex);
+
+                Point::new(size_info.lines() - 1, Column(regex.len() - 1))
```
chrisduerr, line 594: In theory it would be possible to do all this in the `draw_search` method, but I don't like returning the ime position from the `draw_search` function, that doesn't make a lot of sense to me. And doing more than just compute the ime position in a function that is meant to compute it wouldn't be a great idea either. So I think this is the best approach.

alacritty/src/display.rs
```diff
@@ -638,25 +653,13 @@ impl Display {
         config: &Config,
         size_info: &SizeInfo,
         message_bar_lines: usize,
-        search_regex: Option<String>,
+        search_regex: &str,
```
kchibisov, line 656: This is also not a search regex anymore, it's just a search bar's content.

alacritty/src/display.rs
```diff
@@ -579,18 +579,20 @@ impl Display {
 
         self.draw_render_timer(config, &size_info);
 
-        // Handle search bar rendering.
+        // Handle regex search.
```
kchibisov, line 582:
```suggestion
        // Handle regex search and compute IME position.
```

I feel like it reflects things a bit more?

PR #3983, opened 2020-07-16
Remove Ubuntu PPA installation instructions

### LINKED ISSUE #3863: Update Ubuntu PPA
It seems like the Ubuntu PPA hasn't been updated in a while. While on the system76 ppa it is up to date, unfortunately the one we link to is still on 0.4.1.

@mmstick are you still updating the popos one? IIRC yours came before that, right? If not, should our documentation be updated? It has been a while since I've used a ppa or Ubuntu at all, but I feel like adding the system76 ppa might create troubles since it provides a ton of other packages which will probably be picked over the official ones available from Ubuntu.

chrisduerr:
Since the Ubuntu PPA from mmstick is apparently not maintained anymore,
people should no longer be instructed to install the outdated version
that is available from the PPA.

The System76 PPA repository still contains an up to date version of
Alacritty, but since it also comes with several other packages we cannot
recommend users to install it outside of Pop!_OS.

Fixes #3863.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,9 +97,6 @@ zypper in alacritty
 
 ### Pop!\_OS
 
-> **Since this requires a Pop!\_OS specific PPA, it will not work on other Ubuntu
-> distributions.**
-
 ```sh
 apt install alacritty
 ```
--- a/README.md
+++ b/README.md
@@ -95,13 +95,10 @@ nix-env -iA nixos.alacritty
 zypper in alacritty
 ```
 
-### Pop!\_OS / Ubuntu
+### Pop!\_OS
 
-> If you're not running Pop!_OS, you'll have to add a third party repository first:
->
-> ```sh
-> add-apt-repository ppa:mmstick76/alacritty
-> ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -95,13 +95,10 @@ nix-env -iA nixos.alacritty
...
 
-### Pop!\_OS / Ubuntu
+### Pop!\_OS
 
-> If you're not running Pop!_OS, you'll have to add a third party repository first:
->
-> ```sh
-> add-apt-repository ppa:mmstick76/alacritty
-> ```
+> **Since this requires a Pop!\_OS specific PPA, it will not work on other Ubuntu
```
kchibisov, line 100: Not sure that it's even worth mentioning Ubuntu here, since it's just for pop os already. I'd rather remove that part about Ubuntu, since there're many distros using `apt`.

chrisduerr: I've had it removed initially, but then added it because I thought that it might prevent some complaints.

But you're right that it should already be implicit by the removal of the `Ubuntu` from the header. If in the future people should frequently complain about that, we can still change it again to include this notice. Until then, better keep it short.

PR #3986, opened 2020-07-16
Fix selection update in viless search

chrisduerr:


alacritty/src/event.rs
```diff
@@ -402,6 +402,11 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     #[inline]
     fn push_search(&mut self, c: char) {
         if let Some(regex) = self.search_state.regex.as_mut() {
+            // When searching without vi mode, we need to clear selection to invalidate it.
```
kchibisov, line 405: This comment sounds a bit strange to me. Maybe just
```suggestion
            // When searching without vi mode, we need to invalidate selection.
```

it's not really clear what is `it`, if there's some other intent.

chrisduerr: I don't really like your proposal at all, but I've changed it to something that might be a little easier to understand.

PR #3990, opened 2020-07-17
Fix crashes during search

### LINKED ISSUE #3987: Crash during search with too narrow window
If the window's width is smaller than the search bar's label length and the search regex is changed, Alacritty will crash.

chrisduerr:
This fixes two crashes related to search. The first crash was due to
variable shadowing computing the incorrect length of a search as soon as
it was started, causing an overflow after attempting a subtraction.
Since overflows are only debug mode checks, in release this would lead
to an IME positioned far, far beyond the window's boundaries (likely
offscreen).

The other bug was that the search was always started from the last
column in reverse search without vi mode, but after shrinking the
terminal width that origin point is no longer a valid part of the grid.
As a result it would cause an index out of bounds. The line was already
clamped to the grid, but now the same is also done for the column.

Fixes #3987.

### DIFF
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -592,7 +592,7 @@ impl Display {
                 self.draw_search(config, &size_info, message_bar_lines, &search_text);
 
                 // Compute IME position.
-                Point::new(size_info.lines() - 1, Column(regex.len() - 1))
+                Point::new(size_info.lines() - 1, Column(search_text.len() - 1))
             },
             None => cursor_point,
         };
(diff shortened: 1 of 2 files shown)


alacritty/src/display.rs
```diff
@@ -627,25 +627,31 @@ impl Display {
         search_label: &str,
     ) -> String {
         // Add spacers for wide chars.
-        let mut text = String::with_capacity(search_regex.len());
+        let mut formatted_regex = String::with_capacity(search_regex.len());
```
chrisduerr, line 630: I've renamed all of the variables here to make them a little bit more clear.

The goal is that this will prevent accidentally using the incorrect variable in a follow up step or when renaming things and relying on shadowing. Now it should be very clear that `truncated_regex` for example is the search regex after truncation and before the label is prepended.

alacritty/src/display.rs
```diff
@@ -592,7 +592,7 @@ impl Display {
                 self.draw_search(config, &size_info, message_bar_lines, &search_text);
 
                 // Compute IME position.
-                Point::new(size_info.lines() - 1, Column(regex.len() - 1))
+                Point::new(size_info.lines() - 1, Column(search_text.len() - 1))
```
chrisduerr, line 595: Since we renamed `regex` to `search_text`, this broke. Unfortunately Rust's variable shadowing did not do us any favors here.

PR #3991, opened 2020-07-17
Rename font crate to crossfont

chrisduerr:
I have not yet moved this to a separate repository, but I think it might be a good idea.
What do you think @nixpulvis / @kchibisov?

We need to rename it one way or another so we can just rename it now and potentially move it later.
But if we know that we are going to move it, we might as well just do it all in one go.


chrisduerr:
I've moved the font subcrate to https://github.com/alacritty/crossfont and published it on crates.io. That should make it easier if other people want to rely on it and allow for separate issue tracking.

PR #3993, opened 2020-07-18
Bump alacritty_terminal version

chrisduerr:
This change bumps the version of the `alacritty_terminal` crate to force
it out of sync with the Alacritty application. Since it is a library
that will be published on crates.io, it should be following semver
rather than our binary's release version.

While it would be possible to just keep it at 0.5.0 and remove the
`-dev` suffix, explicitly disconnecting it from the Alacritty version
should give it a clean starting-off point and prevent unnecessary
confusion. Bumping it to 0.10.0 instead of something like 0.6.0 should
allow for a nice gap between the different versions without being too
arbitrary. Changing the version to 0.1.0 or something similar is not an
option since decreasing semver version would likely cause a lot of
problems.

While it would be possible to further separate the Alacritty terminal
library from the Alacritty terminal emulator, by renaming it from
`alacritty_terminal` to something different, I don't think that is
necessary or a good idea at the moment. While the `alacritty_terminal`
library *can* be used for other terminals, its primary goal is still to
support the Alacritty terminal emulator and I don't see that changing.
So making that clear in its name seems like a good idea. Since there is
no plan to maintain this library separately from Alacritty, but to keep
both tied together, the naming should reflect this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,8 +45,8 @@ issues.
 Please note that the minimum supported version of Alacritty is Rust 1.41.0. All patches are expected
 to work with the minimum supported version.
 
-If you have made any breaking changes to `alacritty_terminal`, make sure to bump its version
-according to semver.
+Since `alacritty_terminal`'s version always tracks the next release, make sure that the version is
+bumped according to semver when necessary.
 
 ### Testing
 
@@ -131,10 +131,11 @@ The exact steps for an exemplary `0.2.0` release might look like this:
      major issues are found in the release candidates
   9. In the branch, the version is bumped to `0.2.0`
  10. The new commit in the branch is tagged as `v0.2.0`
- 11. A GitHub release is created for the `v0.2.0` tag
- 12. The changelog since the last stable release (**not** RC) is added to the GitHub release
+ 11. The new version is published to crates.io
+ 12. A GitHub release is created for the `v0.2.0` tag
+ 13. The changelog since the last stable release (**not** RC) is added to the GitHub release
      description
- 13. The `-dev` is stripped from the `0.2.0-dev` changelog entries on master
+ 14. The `-dev` is stripped from the `0.2.0-dev` changelog entries on master
 
 On master and with new planned releases, only the minor version is bumped. This makes it possible to
 create bug fix releases by incrementing the patch version of a previous minor release, without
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -45,6 +45,9 @@ issues.
 Please note that the minimum supported version of Alacritty is Rust 1.41.0. All patches are expected
 to work with the minimum supported version.
 
+If you have made any breaking changes to `alacritty_terminal`, make sure to bump its version
```
kchibisov, line 48: This part seems out of place, knowing that you have literally the same message later.

chrisduerr: I've considered the same, but this section is about contributions to Alacritty while the other section is about releases.

I think it's relevant in both sections, so I thought it makes sense to state it twice even if there is a slight overlap.

kchibisov: It just sounds to me like. If you have made any breaking changes to `alacritty_terminal` from your patch, then update version according to semver as well. My main concern here is not that we have overlap, but because it's really confusing the way it is right now. It's pretty clear in `Release` section, since you're doing release, and since versions are synced, it tells you what to do, and how to bump. The `Release` applies even to contributors, don't you think?

chrisduerr: Say the current state of alacritty_terminal is 0.10.0-dev, then a release would bump it to 0.10.1-dev, since there obviously haven't been any breaking changes after 0.10.0 (there haven't been any changes at all!).

However, now a user writes a patch and sends a PR that changes the public interface of Alacritty. At that point keeping the version at 0.10.1-dev would be incorrect, since we cannot release it as 0.10.1-dev anymore. So the version needs to be bumped to 0.11.0.

The section at the top instructs contributors to bump the version according to semver, the section at the bottom instructs publishers to bump just the patch version after a release. They aren't the same and they have different purposes.

kchibisov: >However, now a user writes a patch and sends a PR that changes the public interface of Alacritty. At that point keeping the version at 0.10.1-dev would be incorrect, since we cannot release it as 0.10.1-dev anymore. So the version needs to be bumped to 0.11.0.

Oh, that way, then you should add it to PR template as well?

chrisduerr: Probably a good idea.

PR #4011, opened 2020-07-22
Add secondary DA support

### LINKED ISSUE #3100: Implement "Send Device Attributes",  Secondary and Tertiary DA
Implement `CSI > Ps c` and `CSI = Ps c`:

> ```
> CSI > Ps c
> Send Device Attributes (Secondary DA).
>             Ps = 0  or omitted ⇒  request the terminal's identification
>           code.  The response depends on the decTerminalID resource set-
>           ting.  It should apply only to VT220 and up, but xterm extends
>           this to VT100.
>             ⇒  CSI  > Pp ; Pv ; Pc c
>           where Pp denotes the terminal type
>             Pp = 0  ⇒  "VT100".
>             Pp = 1  ⇒  "VT220".
>             Pp = 2  ⇒  "VT240".
>             Pp = 1 8  ⇒  "VT330".
>             Pp = 1 9  ⇒  "VT340".
>             Pp = 2 4  ⇒  "VT320".
>             Pp = 4 1  ⇒  "VT420".
>             Pp = 6 1  ⇒  "VT510".
>             Pp = 6 4  ⇒  "VT520".
>             Pp = 6 5  ⇒  "VT525".
> 
>           and Pv is the firmware version (for xterm, this was originally
>           the XFree86 patch number, starting with 95).  In a DEC termi-
>           nal, Pc indicates the ROM cartridge registration number and is
>           always zero.
> ```

> ```
> CSI = Ps c
> Send Device Attributes (Tertiary DA).
>             Ps = 0  ⇒  report Terminal Unit ID (default), VT400.  XTerm
>           uses zeros for the site code and serial number in its DECRPTUI
>           response.
> ```
Broken out from https://github.com/jwilm/alacritty/issues/3091#issuecomment-565549380
@AmaiKinono

chrisduerr:
This adds support for the secondary DA escape sequence response.
Alacritty's version is formatted allowing for up to 99 minor and patch
versions, which should be sufficient.

The tertiary DA is intentionally not implemented and marked as rejected
in the documentation, since a lot of terminals do not support it, or
report useless data (XTerm/URxvt/Kitty).

Fixes #3100.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1499,14 +1499,14 @@ impl<T: EventListener> Handler for Term<T> {
     fn identify_terminal<W: io::Write>(&mut self, writer: &mut W, intermediate: Option<char>) {
         match intermediate {
             None => {
-                println!("Reporting primary device attributes");
+                trace!("Reporting primary device attributes");
                 let _ = writer.write_all(b"\x1b[?6c");
             },
             Some('>') => {
-                println!("Reporting secondary device attributes");
+                trace!("Reporting secondary device attributes");
                 let _ = writer.write_all(format!("\x1b[>0;{};1c", version_number()).as_bytes());
             },
-            _ => println!("Unsupported device attributes intermediate"),
+            _ => debug!("Unsupported device attributes intermediate"),
         }
     }
 
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -5,7 +5,7 @@ use std::iter::Peekable;
 use std::ops::{Index, IndexMut, Range, RangeInclusive};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
-use std::{io, iter, mem, ptr, str};
+use std::{env, io, iter, mem, ptr, str};
 
 use log::{debug, trace};
 use serde::{Deserialize, Serialize};
@@ -1496,9 +1496,18 @@ impl<T: EventListener> Handler for Term<T> {
     }
 
     #[inline]
-    fn identify_terminal<W: io::Write>(&mut self, writer: &mut W) {
-        trace!("Reporting terminal identity");
-        let _ = writer.write_all(b"\x1b[?6c");
+    fn identify_terminal<W: io::Write>(&mut self, writer: &mut W, intermediate: Option<char>) {
+        match intermediate {
+            None => {
+                println!("Reporting primary device attributes");
+                let _ = writer.write_all(b"\x1b[?6c");
+            },
+            Some('>') => {
+                println!("Reporting secondary device attributes");
+                let _ = writer.write_all(format!("\x1b[>0;{};1c", version_number()).as_bytes());
+            },
+            _ => println!("Unsupported device attributes intermediate"),
+        }
     }
 
     #[inline]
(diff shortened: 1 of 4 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -1496,9 +1496,18 @@ impl<T: EventListener> Handler for Term<T> {
...
-        trace!("Reporting terminal identity");
-        let _ = writer.write_all(b"\x1b[?6c");
+    fn identify_terminal<W: io::Write>(&mut self, writer: &mut W, intermediate: Option<char>) {
+        match intermediate {
+            None => {
+                println!("Reporting primary device attributes");
+                let _ = writer.write_all(b"\x1b[?6c");
+            },
+            Some('>') => {
+                println!("Reporting secondary device attributes");
```
kchibisov, line 1506: s/println/trace/

alacritty_terminal/src/term/mod.rs
```diff
@@ -1496,9 +1496,18 @@ impl<T: EventListener> Handler for Term<T> {
...
+                let _ = writer.write_all(format!("\x1b[>0;{};1c", version_number()).as_bytes());
+            },
+            _ => println!("Unsupported device attributes intermediate"),
```
kchibisov, line 1509: s/println/debug/ ?

alacritty_terminal/src/term/mod.rs
```diff
@@ -2184,6 +2193,28 @@ impl<T: EventListener> Handler for Term<T> {
     }
 }
 
+/// Terminal version for escape sequence reports.
+///
+/// This reports Alacritty's current version as a unique number based on its semver version. The
```
kchibisov, line 2198: This should be about `alacritty_terminal` crate version and not Alacritty in general, right?

kchibisov:
You should also add changelog entry.

alacritty_terminal/src/term/mod.rs
chrisduerr: Incorrect, the crate version depends on the application `alacritty_terminal` is included in.

kchibisov: I might word it strangely, but I mean that it has nothing to do with alacritty, other users likely has different names of their front end.

chrisduerr: I mean it's internal documentation anyways, not public for library consumers. For all intents and purposes, alacritty_terminals is also part of Alacritty.

kchibisov: Well, I'd may be phrase it in a following way tbqh. What do you think?

```suggestion
/// This reports terminal's current version as a unique number based on its semver version. The
```

chrisduerr: I've considered it, on the other hand if the terminal doesn't use semver this might not actually report the correct thing at all.

I feel like in general making statements about Alacritty in `alacritty_terminal` shouldn't be prohibited, so I'm not sure how I feel about trying to be generic here. Especially with internal-only documentation like this.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2184,6 +2193,28 @@ impl<T: EventListener> Handler for Term<T> {
...
+/// different versions are padded to ensure that a higher semver version will always report a
+/// higher version number.
+fn version_number() -> usize {
+    let mut version = env::var("CARGO_PKG_VERSION").unwrap_or_default();
```
kchibisov, line 2202: This is not a correct way to so. You have this variable if you run alacritty from `cargo, e.g. `cargo run`. However this variable is not presented when you use `alacritty` just by executable from your `PATH`. So the version should be defined during compile time and not runtime.

PR #4013, opened 2020-07-22
docs: improve docs related to opening URLs

markstos:
Before, if you searched in alacritty.yml for URL,
it was easy to miss details which are closely related to it,
like semantic_escape_chars and the "Open" action.

The updated docs explain what a semantic selection is and mention
"URL" in a couple places to make these features more discoverable.
Before, if you searched in alacritty.yml for URL,
it was easy to miss details which are closely related to it,
like semantic_escape_chars and the "Open" action.

The updated docs explain what a semantic selection is and mention
"URL" in a couple places to make these features more discoverable.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -315,6 +315,7 @@
 #background_opacity: 1.0
 
 #selection:
+  # This string contains all characters that are used as separators for "semantic words" in Alacritty.
   #semantic_escape_chars: ",│`|:\"' ()[]{}<>\t"
 
   # When set to `true`, selected text will be copied to the primary clipboard.
@@ -480,10 +481,11 @@
 #
 #   - ToggleViMode
 #   - SearchForward
+#       Start searching toward the right of the search origin.
 #   - SearchBackward
+#       Start searching toward the left of the search origin.
 #   - Copy
 #   - Paste
-#   - PasteSelection
 #   - IncreaseFontSize
 #   - DecreaseFontSize
 #   - ResetFontSize
@@ -496,51 +498,88 @@
 #   - ScrollToTop
 #   - ScrollToBottom
 #   - ClearHistory
+#       Remove the terminal's scrollback history.
 #   - Hide
+#       Hide the Alacritty window.
 #   - Minimize
+#       Minimize the Alacritty window.
 #   - Quit
+#       Quit Alacritty.
 #   - ToggleFullscreen
 #   - SpawnNewInstance
+#       Spawn a new instance of Alacritty.
 #   - ClearLogNotice
+#       Clear Alacritty's UI warning and error notice.
 #   - ClearSelection
+#       Remove the active selection.
 #   - ReceiveChar
 #   - None
 #
 #   (`mode: Vi` only):
 #   - Open
+#       Open URLs at the cursor location with the launcher configured in `url.launcher`.
 #   - Up
+#       Move the vi mode cursor up by one line.
 #   - Down
+#       Move the vi mode cursor down by one linee.
 #   - Left
+#       Move the vi mode cursor left by one character.
 #   - Right
+#       Move the vi mode cursor right by one character.
 #   - First
+#       Move the vi mode cursor to the start of the line, searching across newlines if it is already at the beginning.
 #   - Last
+#       Move the vi mode cursor to the end of the line, searching across newlines if it is already at the end.
 #   - FirstOccupied
+#       Move the vi mode cursor to the first non-empty cell in this line, searching across newlines if it is already at the first one.
 #   - High
+#       Move the vi mode cursor to the top of the screen.
 #   - Middle
+#       Move the vi mode cursor to the middle of the screen.
 #   - Low
+#       Move the vi mode cursor to the bottom of the screen.
 #   - SemanticLeft
+#       Move the vi mode cursor to the start of the previous semantically separated word.
 #   - SemanticRight
+#       Move the vi mode cursor to the start of the next semantically separated word.
 #   - SemanticLeftEnd
+#       Move the vi mode cursor to the end of the previous semantically separated word.
 #   - SemanticRightEnd
-#   - WordRight
+#       Move the vi mode cursor to the end of the next semantically separated word.
 #   - WordLeft
-#   - WordRightEnd
+#       Move the vi mode cursor to the start of the previous whitespace separated word.
+#   - WordRight
+#       Move the vi mode cursor to the start of the next whitespace separated word.
 #   - WordLeftEnd
+#       Move the vi mode cursor to the end of the previous whitespace separated word.
+#   - WordRightEnd
+#       Move the vi mode cursor to the end of the next whitespace separated word.
 #   - Bracket
+#       Move the vi mode cursor to the next character that matches the bracket at the cursor's
+#       current location.
 #   - ToggleNormalSelection
 #   - ToggleLineSelection
 #   - ToggleBlockSelection
 #   - ToggleSemanticSelection
+#       Toggle semantic selection based on `selection.semantic_escape_chars`.
 #   - SearchNext
+#       Jump to the beginning of the next match.
 #   - SearchPrevious
+#       Jump to the beginning of the previous match.
 #   - SearchStart
+#       Jump to the next start of a match to the left of the vi mode cursor.
 #   - SearchEnd
+#       Jump to the next end of a match to the left of the vi mode cursor.
 #
 #   (macOS only):
-#   - ToggleSimpleFullscreen: Enters fullscreen without occupying another space
+#   - ToggleSimpleFullscreen
+#       Enter fullscreen without occupying another space.
 #
 #   (Linux/BSD only):
-#   - CopySelection: Copies into selection buffer
+#   - CopySelection
+#       Copy from the selection buffer.
+#   - PasteSelection
+#       Paste from the selection buffer.
 #
 # - `command`: Fork and execute a specified command plus arguments
 #


alacritty.yml
```diff
@@ -313,7 +313,9 @@
 #background_opacity: 1.0
 
 #selection:
-  #semantic_escape_chars: ",│`|:\"' ()[]{}<>\t"
+  # This characters for the boundaries when quickly selecting words
+  # Remove ':' if you frequently select URLs so the whole URL matches.
```
chrisduerr, line 317: I don't think we should recommend removing `:`. If people don't like these, they can decide on their own what they want to remove.

alacritty.yml
```diff
@@ -613,7 +615,11 @@
   #- { key: V,                           mode: Vi, action: ToggleNormalSelection   }
   #- { key: V,      mods: Shift,         mode: Vi, action: ToggleLineSelection     }
   #- { key: V,      mods: Control,       mode: Vi, action: ToggleBlockSelection    }
+  #
+  # select the "word" under the cursor. See semantic_escape_chars to change boundary detection
```
chrisduerr, line 619: I don't think it makes sense to document the bindings, if anything we should document the actions.

alacritty.yml
markstos: OK. I added it thinking selecting URLs would be a common use-case for the feature, as the various other approaches for selecting URLs with the keyboard have not been implemented.

alacritty.yml
markstos: I'm not sure what you mean. I  _was_ documenting what `ToggleSemanticSelection` does.

Are you suggesting that there should be a separate location where actions are documented outside of the array of key bindings?

In the README the docs say _You can find the default configuration file with documentation for all available fields_ 

This gives the impression all things found in `alacritty.yml` would be documented, but the actions are not. Some are clear enough. Others are not. For example, it's not clear what "FirstOccupied" does.

chrisduerr: > Are you suggesting that there should be a separate location where actions are documented outside of the array of key bindings?

There already is. All existing actions are documented, you're just adding documentation about an action to a binding, which doesn't make much sense.

If we wanted to document the action, we should document it where the action is documented, not the binding.

markstos: Where is ToggleSemanticSelection documented? 

I just searched the project to double check if I missed something:

https://github.com/alacritty/alacritty/search?q=ToggleSemanticSelection&unscoped_q=ToggleSemanticSelection

I see that `alacritty.yml` documents that it exists, but it doesn't explain what it does.

Are you referring to the comments in the source code for each action?
https://github.com/alacritty/alacritty/blob/142f84efb955a9fa59f4205ec3a6db3964c5f433/alacritty/src/config/bindings.rs#L82

The comment for `ToggleSemanticSelection` says  just "/// Toggle semantic vi selection."
That's really just repeating the words in the variable name. A more helpful comment would explain how "semantic" is implemented or provide a link for more information about the semantic matching system. That's the kind of explanation I was trying to provide. 

For example, I thought maybe it was going to match from one paren to another paren or from one brace to another brace. It wasn't clear from the name that was going to match "words".

chrisduerr: https://github.com/alacritty/alacritty/blob/master/alacritty.yml#L534

> I see that alacritty.yml documents that it exists, but it doesn't explain what it does.

Correct, but that's what I'm saying. If it should be documented, the action should be documented not the binding.

The binding doesn't do anything, it just ties a key to an action. You can have an action bound an arbitrary amount of times, but it still always does the same thing. So what should be documented is the action, not the binding.

> For example, I thought maybe it was going to match from one paren to another paren or from one brace to another brace. It wasn't clear from the name that was going to match "words".

It doesn't match words. It matches semantically, based on the semantic escape characters.

nixpulvis: I'm 100% in favor of documenting the action not the binding. The question remains, _how_ exactly to document the action?

I would propose adding them below the item (possibly with an additional newline as shown) like the following:

```
# - Open
#   Attempt to open the element under the cursor, like a URL. See also url.launcher.
#
# - ToggleSemanticSelection
#   Select the "word" under the cursor. See `semantic_escape_chars` to change boundary detection.
# - ...
```

FWIW, if this were rendered markdown it would look like this:

- Open
  Attempt to open the element under the cursor, like a URL. See also url.launcher.

- ToggleSemanticSelection
  Select the "word" under the cursor. See `semantic_escape_chars` to change boundary detection.
- ...

markstos: I can also get behind documenting the actions and not the bindings, and @nixpulvis proposal for how to format them seems fine with me. They are in a YAML comment, for formatting them as valid Markdown does invite future automation or reformatting to further surface the docs. 

Looks like someone could take a pass through the source code to gather all the code comments describing the actions and turn those into docs.

https://github.com/alacritty/alacritty/blob/142f84efb955a9fa59f4205ec3a6db3964c5f433/alacritty/src/config/bindings.rs#L82

chrisduerr: I personally prefer not having spaces between the different actions, since that would waste a lot of space and break the list apart. I think something like this is better:

```
#   - Up
#       Move the vi mode cursor up
#   - Down
#       Move the vi mode cursor down
```

Personally I don't think there's much value in taking code comments into user-facing documentation. These two have completely different purposes and should not just copy one another.

alacritty.yml
markstos: I've removed this line in my updated commit.

alacritty.yml
markstos: My updated commit provides docs in this format.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - ScrollLineUp
+#       Scroll one line up
 #   - ScrollLineDown
+#       Scroll one line down
 #   - ScrollToTop
+#       Scroll all the way to the top
 #   - ScrollToBottom
+#       Scroll all the way to the bottom
 #   - ClearHistory
+#       Clear the display buffer(s) to remove history
```
chrisduerr, line 518: Are you aware of what this does? Otherwise documenting it might be a bit difficult. The statement "clear the display buffer(s)" seems a bit strange to me.

There's two buffers we have in Alacritty, the primary and alternate screen buffer. The alternate screen buffer never has any history so it's irrelevant here. So talking about buffer***s*** makes no sense in this case. So all this does is clear the **history** of the primary screen buffer. This has no effect on the visible content.

```suggestion
#       Remove the terminal's scrollback history
```

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - Hide
+#       Hide the Alacritty window
 #   - Minimize
+#       Minimize the Alacritty window
 #   - Quit
+#       Quit Alacritty
 #   - ToggleFullscreen
+#       Toggle full screen
```
chrisduerr, line 526: Are these comments of value? I'd imagine they're self-descriptive enough.

I don't think we need to document every action. If a comment cannot contribute enough value, it's just wasting space. So in that case omitting it is fine.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - SpawnNewInstance
+#       Spawn a new instance of Alacritty
```
chrisduerr, line 528: We should document why people might want to use this over just spawning a new Alacritty instance using a command binding. The main difference here is that SpawnNewInstance will try to preserve the path. Currently this only works on Linux and some BSD systems though.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - ClearLogNotice
+#       Clear warning and error notices
```
chrisduerr, line 530: Should we maybe go into more detail on what this means? It might make sense to clarify that this talks about Alacritty's UI error messages.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - ClearSelection
+#       Clear active selection
```
chrisduerr, line 532: I think we can do better in a description here. Rather than reusing the word `clear`, by replacing it with `Remove the active selection` we should be able to contribute to the name itself and help understand what this actually does.

So people know this doesn't just clear the selection clipboard or something like that.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - ReceiveChar
+#       Allow receiving character input
```
chrisduerr, line 534: I think this can be documented a bit better, especially w.r.t. why someone would want to use it.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - None
+#       No action
```
chrisduerr, line 536: Are these comments of value? I'd imagine they're self-descriptive enough.

I don't think we need to document every action. If a comment cannot contribute enough value, it's just wasting space. So in that case omitting it is fine.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #
 #   (`mode: Vi` only):
 #   - Open
+#       Attempt to open the element under the cursor, like a URL. See also url.launcher.
```
chrisduerr, line 540: Currently this only opens URLs, so we should be clear and state that. If that changes in the future, we can change our documentation.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - Up
+#       Move one character up
```
chrisduerr, line 542: This applies to all vi motions, but I'm going to put it just here to make the review a little less "spammy" than it already is.

I think we should clarify what is moved here. Just stating `Move one character up` might also just mean that the viewport is moved.

I'd suggest stating that the `vi mode cursor` is moved in all these motion documentations. In this case specifically, I'd suggest something like this:

```suggestion
#   - Up
#       Move the vi mode cursor up by one line
```

I think it makes sense to replace "one character" with "one line", since the idea of moving up by "one character" is a bit strange in a line based environment. For left/right motion it does make sense though (since a character might span over multiple columns).

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - Left
+#       Move one character left
 #   - Right
+#       Move one character right
 #   - First
+#       Move to the start of line
 #   - Last
+#       Move to the end of line
 #   - FirstOccupied
+#       Move to the first non-empty cell
```
chrisduerr, line 554: These bindings are actually a bit more complicated than they might seem at first. They first move to the specified character in the current line and then move to the first character across wrapped lines. This is quite a lot of detail and I'm not sure if we want to document it, but I think it might be useful to know about?

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - High
+#       Move to the top of the screen
 #   - Middle
+#       Move to the middle of the screen
 #   - Low
+#       Move to the bottom of the screen
 #   - SemanticLeft
+#       Move to start of semantically separated word
```
chrisduerr, line 562: With all these word based movements, you omitted an `a` before the adjective that makes things a bit awkward to read here. So saying `of a semantically separated word` sounds a lot better.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - SemanticRight
+#       Move to start of next semantically separated word.
```
chrisduerr, line 564: You started adding punctuation here. I'd assume that it was by mistake?

Generally I think it would be in line with Alacritty's other internal and external documentation to add periods to all of these.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
-#   - WordRightEnd
+#       Move to start of whitespace separated word.
+#   - WordRight
+#       Move to start of next whitespace separated word.
 #   - WordLeftEnd
+#       Move to end of previous whitespace separated word.
+#   - WordRightEnd
+#       Move to end of whitespace separated word.
 #   - Bracket
+#       Move to opposing bracket.
```
chrisduerr, line 578: I think we can clarify this a bit better. Something like this maybe:

```suggestion
#       Move the vi mode cursor to the next character that matches the bracket at the cursor's current location
```

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - ToggleNormalSelection
+#       Toggle normal vi selection
 #   - ToggleLineSelection
+#       Toggle line vi selection
 #   - ToggleBlockSelection
+#       Toggle block vi selection
```
chrisduerr, line 584: If we just state the same thing the action says, I don't think that's of much value. But maybe there's a better description here? It might be useful to state that one can toggle between these modes even if there currently already is a selection active, which will lead to a conversion of the active selection.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - ToggleSemanticSelection
+#       Toggle selecting the "word" under the cursor.
+#       See semantic_escape_chars to change boundary detection
 #   - SearchNext
+#       Jump to the beginning of the next match.
```
chrisduerr, line 589: We should clarify how this interacts with search. So we need to make it clear that the user first searches for something and once a search is completed, this action will allow you to jump between the highlighted search match results.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
 # - `action`: Execute a predefined action
 #
 #   - ToggleViMode
+#       Enable or Disable Using Vi key bindings
```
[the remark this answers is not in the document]
markstos: I'll remove this comment. The name is self-evident enough.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - SearchForward
+#       Start a forward buffer search
 #   - SearchBackward
+#       Start a backward buffer search
 #   - Copy
+#       Copy the selection to the clipboard
```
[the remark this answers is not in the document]
markstos: I'll remove this one.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - Paste
+#       Paste from the system clipboard
```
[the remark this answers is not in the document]
markstos: I'll remove this comment.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - PasteSelection
+#       Paste from the selection buffer
```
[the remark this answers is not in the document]
markstos: Advice accepted.

alacritty.yml
```diff
@@ -480,62 +481,118 @@
...
 #   - ScrollHalfPageDown
+#       Scroll a half page down
 #   - ScrollLineUp
+#       Scroll one line up
 #   - ScrollLineDown
+#       Scroll one line down
 #   - ScrollToTop
+#       Scroll all the way to the top
 #   - ScrollToBottom
+#       Scroll all the way to the bottom
```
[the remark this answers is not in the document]
markstos: Fair enough. Removing.

alacritty.yml
markstos: I think documenting Hide, Minimize and Quit are are all valuable because _what_ they hide, minimize and quit is clarified. For example, "quit" alone might refer to quitting a mode. I'll remove the "toggle full screen" comment.

alacritty.yml
markstos: Advice taken.

alacritty.yml
chrisduerr: What might "minimize" refer to? I think especially with the way these are grouped, they are fairly obvious. I could see your argument for Quit, but I'm not sure how it would apply to Hide/Minimize.

alacritty.yml
markstos: I'm not clear what it does, so I'm removing the docs for it. You can add something back if you like.

alacritty.yml
markstos: Removing.

alacritty.yml
markstos: Updating to clarify it only opens URLs.

alacritty.yml
markstos: I'm updating `First` and `Last` with phrasing like _Move to the start of line, including wrapped lines_

Sinc `FirstOccupied` doesn't clearly imply that it is a line-based action, I think we can leave off the note about line wrapping there.

chrisduerr: > Sinc FirstOccupied doesn't clearly imply that it is a line-based action, I think we can leave off the note about line wrapping there.

It still is though. All these actions move to the first instance in the line first and then move to the first instance across wrapped lines when already at the first instance within the line. They do not move across wrapped lines immediately.

alacritty.yml
markstos: Adding.

alacritty.yml
markstos: Suggestion accepted.

alacritty.yml
markstos: I'm removing the descriptions here.

alacritty.yml
markstos: Updated to clarify when they apply.

markstos:
@chrisduerr Thanks for the prompt and detailed feedback. I've pushed a commit to reflect refinements based on your feedback.

alacritty.yml
```diff
@@ -481,10 +482,13 @@
 #
 #   - ToggleViMode
 #   - SearchForward
+#       Start searching to the right of the search origin.
 #   - SearchBackward
+#       Start searching to the left of the search origin.
 #   - Copy
 #   - Paste
 #   - PasteSelection
+#       Paste from the selection clipboard buffer. Only applies to Wayland/X11.
```
chrisduerr, line 491:
```suggestion
#       Paste from the selection clipboard buffer on Wayland/X11.
```

chrisduerr: I think this reads a little nicer.

alacritty.yml
```diff
@@ -497,45 +501,79 @@
 #   - ScrollToTop
 #   - ScrollToBottom
 #   - ClearHistory
+#       Remove the terminal's scrollback history.
 #   - Hide
+#       Hide the Alacritty window.
 #   - Minimize
+#       Minimize the Alacritty window.
 #   - Quit
+#       Quit Alacritty
```
chrisduerr, line 510:
```suggestion
#       Quit Alacritty.
```

alacritty.yml
```diff
@@ -497,45 +501,79 @@
...
 #   - ToggleFullscreen
 #   - SpawnNewInstance
+#       Spawn a new instance of Alacritty that preserves the path. Only applies to Linux/BSD.
```
chrisduerr, line 513: It's not really true that this action only applies to Linux/BSD, but only on Linux/BSD do we currently try to open it in the shell's current working directory.

It's also not guaranteed to work, for several technical reasons.

alacritty.yml
```diff
@@ -497,45 +501,79 @@
...
 #   - ClearLogNotice
+#       Clear Alacritty's UI warning and error notices.
```
chrisduerr, line 515:
```suggestion
#       Clear Alacritty's UI warning and error notice.
```

While there can be multiple errors, there's only one warning/error message bar that this clears. So I think that makes it a bit more clear that we just refer to this one thing here.

alacritty.yml
```diff
@@ -497,45 +501,79 @@
...
 #   - ClearSelection
+#       Remove the active selection.
 #   - ReceiveChar
 #   - None
 #
 #   (`mode: Vi` only):
 #   - Open
+#       Attempt to open URL under the cursor. See also url.launcher.
```
chrisduerr, line 523: If there is no URL under the cursor, this does nothing. We don't even try to open it as a URL. So we should be clear that this opens the URL if there is one.

```suggestion
#       Opens URLs at the cursor location with the launcher configured in `url.launcher`.
```

alacritty.yml
```diff
@@ -497,45 +501,79 @@
...
 #   - Up
+#       Move the vi mode cursor up by one line.
 #   - Down
+#       Move one character down.
```
chrisduerr, line 527: As mentioned in the comment about the `Up` action, it applied to all motions. So please change these too.

alacritty.yml
```diff
@@ -497,45 +501,79 @@
...
 #   - FirstOccupied
+#       Move to the first non-empty cell.
 #   - High
+#       Move to the top of the screen.
 #   - Middle
+#       Move to the middle of the screen.
 #   - Low
+#       Move to the bottom of the screen.
 #   - SemanticLeft
+#       Move to start of a semantically separated word.
```
chrisduerr, line 545:
```suggestion
#       Move to start of the previous semantically separated word.
```

alacritty.yml
```diff
@@ -497,45 +501,79 @@
...
+#       Move to end of the next semantically separated word.
 #   - WordLeft
-#   - WordRightEnd
+#       Move to start of the previous whitespace separated word.
+#   - WordRight
+#       Move to start of the next whitespace separated word.
 #   - WordLeftEnd
+#       Move to end of the previous whitespace separated word.
+#   - WordRightEnd
+#       Move to end of whitespace separated word.
```
chrisduerr, line 559:
```suggestion
#       Move to end of the next whitespace separated word.
```

alacritty.yml
markstos: In the last draft, I tried to consistently add punctuation.

alacritty.yml
nixpulvis: Personally I'd avoid mentioning the path at all here. It's unreliable across systems. Perhaps if, down the road, launching in the same working directory works more consistently, then we could mention it here.

alacritty.yml
```diff
@@ -481,10 +482,13 @@
 #
 #   - ToggleViMode
 #   - SearchForward
+#       Start searching to the right of the search origin.
 #   - SearchBackward
+#       Start searching to the left of the search origin.
```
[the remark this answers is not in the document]
markstos: I see what you are saying. You saying that just because an object is the "to the left" of another, it does not mean the object is "heading in the left direction".

I'll replace "to the left" with "towards the left". The words `leftward` and `rightward` would also work here, but `towards the left` is simpler English, so I'll stick with that.

alacritty.yml
markstos: @nixpulvis Does it work reliably on Linux? For those of us who use Linux exclusively, that's interesting to us. It seems fair to document features that only work on one platform or environment.

In the interest of getting this merged, though I'll drop the comment since both of you spoke against it.

alacritty.yml
markstos: Thanks for catching that.

I missed those parallel changes last time. Fixed in my next commit.

markstos:
The last round of feedback has been addressed in the latest commit.

alacritty.yml
nixpulvis: TBH I haven't tested it in a while, I setup an X11 dependent script a while back to use as a launcher, and never really needed to replace it (though it would be nice to remove it).

FWIW, here's my script:

```sh
ACTIVE_WINDOW=$(xdotool getactivewindow)
ACTIVE_WM_CLASS=$(xprop -id $ACTIVE_WINDOW | grep WM_CLASS)

if [[ $ACTIVE_WM_CLASS == *"Alacritty"* ]]
then
    # Get PID. If _NET_WM_PID isn't set, bail.
    PID=$(xprop -id $ACTIVE_WINDOW | grep _NET_WM_PID | grep -oP "\d+")
    if [[ "$PID" == "" ]]
    then
        exit 1
    fi

    # Get first child of terminal
    CHILD_PID=$(pgrep -P $PID)
    if [[ "$PID" == "" ]]
    then
        exit 1
    fi

    # Get current directory of child. The first child should be the shell.
    pushd "/proc/${CHILD_PID}/cwd"
    SHELL_CWD=$(pwd -P)
    popd

    # Start alacritty with the working directory
    alacritty --working-directory $SHELL_CWD $@
else
    exit 1
fi
```

alacritty.yml
```diff
@@ -497,45 +501,79 @@
...
+#       Move the vi mode cursor to the next character that matches the bracket at the cursor's
+#       current location.
 #   - ToggleNormalSelection
 #   - ToggleLineSelection
 #   - ToggleBlockSelection
 #   - ToggleSemanticSelection
+#       Toggle selecting the semantic string under the cursor.
+#       See semantic_escape_chars to change boundary detection.
 #   - SearchNext
+#       Once a search has completed, jump to the beginning of the next match.
```
chrisduerr, line 570: I don't like the wording of this, since the search is still ongoing. Just the search regex input has been completed.
Though I'm not entirely sure how to word it any better.

alacritty.yml
markstos: Thanks for the tip. I'm marking this as resolved because I removed the statement about the path for now.

alacritty.yml
markstos: I'll try "Once a search term has been entered,"

markstos:
I've accepted all the suggested edits and made one minor refinement.

alacritty.yml
```diff
@@ -497,45 +501,78 @@
...
 #   - ToggleSemanticSelection
+#       Toggle semantic selection based on `selection.semantic_escape_chars`.
 #   - SearchNext
+#       Once a search term has been entered, jump to the beginning of the next match.
 #   - SearchPrevious
+#       Once a search term has been entered, jump to the beginning of the previous match.
 #   - SearchStart
+#       Jump to the next start of a match to the left of the origin.
 #   - SearchEnd
+#       Jump to the next end of a match to the right of the origin.
```
chrisduerr, line 575:
```suggestion
#       Jump to the next start of a match to the left of the vi mode cursor.
#   - SearchEnd
#       Jump to the next end of a match to the left of the vi mode cursor.
```

alacritty.yml
```diff
@@ -497,45 +501,78 @@
...
+#       current location.
 #   - ToggleNormalSelection
 #   - ToggleLineSelection
 #   - ToggleBlockSelection
 #   - ToggleSemanticSelection
+#       Toggle semantic selection based on `selection.semantic_escape_chars`.
 #   - SearchNext
+#       Once a search term has been entered, jump to the beginning of the next match.
 #   - SearchPrevious
+#       Once a search term has been entered, jump to the beginning of the previous match.
```
chrisduerr, line 571: Honestly I think whatever we come up with here, it's just not going to be any more useful than just the description of the action itself without mentioning the relation to the search action.

So let's just go with the same thing we do for SearchStart/SearchEnd and let people figure out on their own that this is only active after the search has been confirmed.

```suggestion
#       Jump to the beginning of the next match.
#   - SearchPrevious
#       Jump to the beginning of the previous match.
```

markstos:
Suggestions accepted. I believe this branch is ready to squash-and-merge now.

alacritty.yml
```diff
@@ -313,7 +313,8 @@
 #background_opacity: 1.0
 
 #selection:
-  #semantic_escape_chars: ",│`|:\"' ()[]{}<>\t"
+  # This string contains all characters that are used as separators for "semantic words" in Alacritty.
+  # semantic_escape_chars: ",│`|:\"' ()[]{}<>\t"
```
kchibisov, line 317:
```suggestion
  #semantic_escape_chars: ",│`|:\"' ()[]{}<>\t"
```

alacritty.yml
```diff
@@ -497,45 +501,78 @@
...
+#       Jump to the beginning of the next match.
 #   - SearchPrevious
+#       Jump to the beginning of the previous match.
 #   - SearchStart
+#       Jump to the next start of a match to the left of the vi mode cursor.
 #   - SearchEnd
+#       Jump to the next end of a match to the left of the vi mode cursor.
 #
 #   (macOS only):
 #   - ToggleSimpleFullscreen: Enters fullscreen without occupying another space
```
chrisduerr, line 578: One other thing I've noticed right here. It looks like we already have documentation on some of these actions, but this isn't formatted the same way as the new documentation you've added.

Since I prefer the new formatting, could you apply it to the `ToggleSimpleFullscreen` and `CopySelection` action too?

The description should be changed to `Copy from the selection clipboard buffer.` for the copy stuff and to `Enter fullscreen without occupying another space.` for the toggle simple fullscreen action.

alacritty.yml
```diff
@@ -481,10 +482,13 @@
 #
 #   - ToggleViMode
 #   - SearchForward
+#       Start searching toward the right of the search origin.
 #   - SearchBackward
+#       Start searching toward the left of the search origin.
 #   - Copy
 #   - Paste
 #   - PasteSelection
+#       Paste from the selection clipboard buffer on Wayland/X11.
```
chrisduerr, line 491: Can we move this down to `CopySelection`? This already states that it's Linux/BSD only, so we can then change it to `Paste from the selection clipboard buffer.`

alacritty.yml
markstos: Fixed in next commit.

alacritty.yml
markstos: Fixed in next commit. I'm using the term "selection buffer" for both, rather than "selection clipboard buffer". If a Linux user is reading alacritty.yml file to set up custom key bindings, they likely understand what the selection buffer is.

chrisduerr: Sounds good to me.

alacritty.yml
chrisduerr: Make sure to push it and rebase this PR on top of the latest master. Then it can be merged once CI is done.

markstos:
latest round of feedback addressed.

markstos:
It's a small Thank You for a great tool that I use every day.

I have particularly been enjoying the new Vi mode. I use it push a branch to grab the URL to "create pull request" from the output to easily open that page.

markstos:
I've now pushed one squashed and rebased commit.

PR #4018, opened 2020-07-22
Improve selection expansion changelog entry

chrisduerr:


markstos:
Looks Good to Me.

It'd be nice if Vi Mode selections could also be expanded in a similar way.

I tried using the "Sematic Select" action then quickly repeating it to mimic a "double click", but that didn't work.

chrisduerr:
> I tried using the "Sematic Select" action then quickly repeating it to mimic a "double click", but that didn't work.

What do you mean by this? The semantic selection should just start right away, right?

markstos:
Yes, it immediately selected a word.

I was curious if I could emulate the mouse behavior of repeatedly clicking to expand the selection. I tried repeatedly using the keybinding for semantic selection to see if it worked like and it didn't.

chrisduerr:
I'm not going to change the vi mode entry for now, since it shouldn't be any surprise to someone who has used vi before. So I'd rather just keep that a bit shorter for now. Though documenting the actions in general would make sense.

markstos:
I've used Vi, Vim and Neovim for a couple decades and it was a surprise me. I thought maybe I missed something, so typed `:help semantic` in Neovim, but "Sorry, no help for semantic".

Still thinking maybe I should have already known about the term "semantic selection" I also did a web search:

https://duckduckgo.com/?q=vim+semantic+selection&t=canonical&ia=web

Still nothing obvious in the first page of results there about Vim semantic selections.  (At least not by that name).

I think having some clear docs about this now could save the development team some support time later.

chrisduerr:
> I've used Vi, Vim and Neovim for a couple decades and it was a surprise me. I thought maybe I missed something, so typed :help semantic in Neovim, but "Sorry, no help for semantic".

It works the same way every selection in vim does. Just like line selection and block selection. None of these require pressing keys multiple times. In fact that's not how any binding in vi works, they don't toggle between modes based on the number of presses.

The term semantic selection just refers to the semantic escape characters defined in Alacritty's configuration and the existing semantic selection used for the mouse double click.

markstos:
Every vi-like interface has some differences and that was one place I was exploring to see if it might be different. Without documentation for actions, expect users to explore and experiment. 

Also, thanks for your work on Alacritty and the Vi selection mode feature. I am grateful for both and was attempting to contribute my own small improvement by updating the docs a bit to help another new user struggle less where I spent a little time to figure out how it worked.

chrisduerr:
> Without documentation for actions, expect users to explore and experiment.

I mean the semantic selection in vi mode action's binding is documented. It states exactly what you need to do to activate it in the default configuration file.

PR #4019, opened 2020-07-23
Document Alacritty's UI features

chrisduerr:


docs/features.md
```diff
@@ -0,0 +1,64 @@
...
+This document gives an overview over Alacritty's features beyond its terminal
+emulation capabilities. To get a list with supported control sequences take a
+look at [Alacritty's escape sequence support](./escape_support.md).
+
+## Vi Mode
+
+The vi mode allows moving around Alacritty's viewport and scrollback using the
+keyboard. It also serves as a jump-off point for other features like search and
+opening URLs with the keyboard. By default you can launch it using
+<kbd>Ctrl</kbd> <kbd>Shift</kbd> <kbd>Space</kbd>.
```
kchibisov, line 12: There's a way to add a `plus` sign in between for it to look more homogeneous.

docs/features.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Motion
+
+The cursor motions are setup by default to mimic vi, however they are fully
+configurable. If you don't like vi's bindings, take a look at the [configuration
+file] to change the various movements.
+
+### Selection
+
+One useful feature of vi mode is the ability to make selections and copy text to
```
kchibisov, line 22: You have one selection.

```suggestion
One useful feature of vi mode is the ability to select and copy text to
```

docs/features.md
```diff
@@ -0,0 +1,64 @@
...
+the clipboard. By default you can start a selection using <kbd>v</kbd> and copy
+it using <kbd>y</kbd>. All selection modes that are available with the mouse can
+be accessed from vi mode, including the semantic (<kbd>Alt</kbd> <kbd>v</kbd>),
+line (<kbd>Shift</kbd> <kbd>v</kbd>) and block selection (<kbd>Ctrl</kbd>
```
kchibisov, line 26: Same thing with `plus` sign.

docs/features.md
```diff
@@ -0,0 +1,64 @@
...
+
+Search allows you to find anything in Alacritty's scrollback buffer, it can be
+launched either directly (<kbd>Ctrl</kbd> <kbd>Shift</kbd> <kbd>f</kbd>), or
+from vi mode (<kbd>/</kbd>).
+
+### Vi Search
+
+When using search while the vi mode is active, it can be used to quickly move
+around the scrollback buffer. The bindings <kbd>Ctrl</kbd> <kbd>n</kbd> and
+<kbd>Ctrl</kbd> <kbd>Shift</kbd> <kbd>n</kbd> allow you to navigate within
```
kchibisov, line 47: Those are just `n` and `shift + n`, no ctrl is involved here. Also, `plus`.

docs/features.md
```diff
@@ -0,0 +1,64 @@
...
+matches, which make them useful tools for selecting your matches.
+
+### Normal Search
+
+During normal search you don't have the opportunity to move around freely, but
+you can still jump between matches using <kbd>Enter</kbd> and <kbd>Shift</kbd>
+<kbd>Enter</kbd>. After leaving search with <kbd>Escape</kbd> your active match
+stays selected, allowing you to easily copy it.
```
kchibisov, line 55: it's not true that it stays selected(it'll be cleared in alt-screen). But I guess the current wording is fine.

docs/features.md
chrisduerr: I'm not sure what I like more tbh. I feel like without the plus formats a little nicer, the `+` looks a little out of place:

<kbd>Ctrl</kbd>+<kbd>c</kbd>
<kbd>Ctrl</kbd> <kbd>c</kbd>

docs/features.md
chrisduerr: You can perform it multiple times though. So making selections is still correct? I can change it though to make it more clear that the only purpose of selections is to copy text?

docs/features.md
chrisduerr: That is incorrect, I'm explicitly talking about motions **within** the search here.

docs/features.md
chrisduerr: That's not correct. The selection stays active both in the primary and alternative screen buffer.

Now if your content is resized and that clears the content below it, that's a completely different thing.

docs/features.md
kchibisov: I mean we don't have `Ctrl + Shift + N` or `Ctrl + n` in a alacritty.yml or implicitly in terminal, I'm not sure what is incorrect here.

docs/features.md
kchibisov: I do like it in a way `Ctrl + c` tbqh, but it uses a bit different formatting, than `kbd`.

chrisduerr: I personally prefer <kbd>Ctrl</kbd> <kbd>c</kbd> out of all of them. It doesn't have an out of place looking connecting character, is still clear and makes it obvious that this is about a keybinding.

docs/features.md
chrisduerr: Oh right, these aren't bound to anything by default. I completely forgot about that.

Do you think it would make sense to bind these by default? Do you have them bound to something?

chrisduerr:
> We should link to this document from the readme, so it'll be easy to discover.

Completely agree with that, I wanted to do that actually but forgot about it. Great idea.

> As for the features you've listed itself. Should we mention url opening in normal mode? Also, should we mention that you should press shift to select in certain cases, and that you can figure that out from the current mouse cursor state?

I'm a bit conflicted on this one. It makes some sense, but on the other hand it's not that crazy of a feature. It feels like something that wouldn't need to be documented? But it might be useful for people that are trying to compare Alacritty without actually running it. I suppose it doesn't hurt to add it.

docs/features.md
kchibisov: >Do you think it would make sense to bind these by default? Do you have them bound to something?

I don't think that we should.

kchibisov: This is still not addressed. Just removing the sentence about `ctrl + shift + n` should solve it.

chrisduerr: Should we then document that these bindings can be used to jump within matches? I think it's an important detail of vi search.

kchibisov: >Should we then document that these bindings can be used to jump within matches? I think it's an important detail of vi search.

You mean `Actions`? we don't have the bindings you're talking about.

chrisduerr: No, I mean that people can create these bindings to jump around within matches.

kchibisov: people can create whatever bindings they want. They could bind certain actions to have an ability to move within the search, however actions that we have to navigate around, are not bound to a 'current' match, they just traverse the matches in a defined order. So I do feel like there's no value in that sentence at all.

chrisduerr: How is there no value in documenting that a binding exists that is not bound by default which would likely be very useful to a lot of people?

kchibisov: Well, we have other bindings not bound by default, which also has a value to begin with.

>How is there no value in documenting that an action exists that is not bound by default which would likely be very useful to a lot of people

You can say that there're other ways to move Vi cursor other than default `n` and `N` ans just list the actions folks can use, describing what they should do. However if you want to describe a key binding that may be of use maybe use config's key binding section? For me features file is more like, 'things you can try right away without touching config at all'.

chrisduerr: Using the configuration file's key binding section would likely make this harder to discover though.

chrisduerr: I've reworded things a little bit, let me know what you think.

docs/features.md
```diff
@@ -0,0 +1,72 @@
...
+Search allows you to find anything in Alacritty's scrollback buffer. You can
+search forward using <kbd>Ctrl</kbd> <kbd>Shift</kbd> <kbd>f</kbd> and
+backward using <kbd>Ctrl</kbd> <kbd>Shift</kbd> <kbd>b</kbd>.
+
+### Vi Search
+
+In vi mode the search is bound to <kbd>/</kbd> for forward and <kbd>?</kbd> for
+backward search. This allows you to move around quickly and help with selecting
+content. The `SearchStart` and `SearchEnd` keybinding actions can be bound if
+you're looking for a way to jump to either end of a match.
```
kchibisov, line 48: The sentence is unfinished. 

```suggestion
you're looking for a way to jump to either start of the match or end.
```

chrisduerr: It is finished, you can jump to either end of the search match with these bindings.

PR #4027, opened 2020-07-24
Remove package installation instructions in README

vancanhuit:
Add TOC in README.


kchibisov:
I don't like the idea to add ToC. README should give a short overview of the project, and not be a book in the end. I do agree that moving all the official packing to some other place will likely make sense. In general a lot of package managers do have a way to search, so folks can really figure that out themselves?

chrisduerr:
@vancanhuit Would you remind removing the ToC and the package install instructions? Maybe the only notice we need to keep is the one about the binaries available on the releases page.

vancanhuit:
I've moved the OS package installation into a separate file called `OS-PKGS-INSTALL.md`.

vancanhuit:
@chrisduerr Why should we remove the OS package installation instructions?

chrisduerr:
> Why should we remove the OS package installation instructions?

Because they're pointless. People should be able to search their package repos for Alacritty. That should be the natural first step **before** checking Alacritty's readme anyways. And it's not like there's going to be any naming conflicts.

vancanhuit:
Ok, I see. I've removed it from the README.

README.md
```diff
@@ -13,6 +13,21 @@ Alacritty currently supports macOS, Linux, BSD, and Windows.
...
+  - [Manual installation](#manual-installation)
+  - [Prebuilt binaries](#prebuilt-binaries)
+- [Configuration](#configuration)
+  - [Windows](#windows)
+- [Contributing](#contributing)
+- [Issues (known, unknown, feature requests, etc.)](#issues-known-unknown-feature-requests-etc)
+- [FAQ](#faq)
+- [IRC](#irc)
+- [Wayland](#wayland)
+- [License](#license)
```
chrisduerr, line 29: I don't think a ToC is necessary at this point anymore.

README.md
```diff
@@ -41,120 +56,13 @@ You can find an overview over the features available in Alacritty [here](./docs/
...
-
-### Windows
-
-#### Via [Chocolatey](https://chocolatey.org)
-
-```batch
-choco install alacritty
-```
+### Manual installation
+For everyone else the instructions to build Alacritty from source can be found [here](INSTALL.md).
```
chrisduerr, line 60: This wording doesn't make sense anymore. I also don't think there's a point to have two subsections for essentially just two paragraphs.

So we can just have one sentence about prebuilt binaries and then a second paragraph talk about building from source. In the prebuilt binary section we should also mention that Alacritty is available on various package managers on Linux, BSD, macOS and Windows. That should encourage people to search their respective repos.

vancanhuit: Okay, I will update later on.

README.md
```diff
@@ -164,6 +54,8 @@ requires winpty to emulate UNIX's PTY API. The agent is a single binary
 executable and is available through the
 [GitHub releases page](https://github.com/alacritty/alacritty/releases).
```
chrisduerr, line 55: I suggest changing this to:

```
On Windows, Alacritty requires winpty for versions older than Windows 10
(October 2018 Update) and Microsoft's VC++ redistributable. The winpty agent
(`winpty-agent.exe`) **must** be in the same directory as the Alacritty
executable and is available through the
[GitHub releases page](https://github.com/alacritty/alacritty/releases).
```

That way we don't have things split into a million different paragraphs.

README.md
```diff
@@ -164,6 +54,8 @@ requires winpty to emulate UNIX's PTY API. The agent is a single binary
...
 
+For everyone else the instructions to build Alacritty from source can be found [here](INSTALL.md).
```
chrisduerr, line 57:
```suggestion
For everyone else, the instructions to build Alacritty from source can be found [here](INSTALL.md).
```

nixpulvis:
I've thought about opening this kind of PR before myself. These instructions are very generic and redundant. People should be comfertable enough with their OS to know where to find the packages.

I would be curious if leaving a list (ideally with links) of packages in `apt`, `brew`, `pacman`, etc. would be a good idea.

chrisduerr:
> I would be curious if leaving a list (ideally with links) of packages in apt, brew, pacman, etc. would be a good idea.

I don't think it's useful to maintain these, especially because it requires constant work from our part. Pointing out that there are package managers for macOS, Windows, BSD and Linux should already give a hint to people of every platform that there are package managers that support it.

chrisduerr:
@vancanhuit could you please rebase your PR instead of cherry-picking the commits? Otherwise the diff will be messed up.

vancanhuit:
@chrisduerr How to rebase the PR? This is the first time I face this issue so I'm confused about how to do this.

README.md
```diff
@@ -41,128 +41,17 @@ You can find an overview over the features available in Alacritty [here](./docs/
...
-
-### Fedora
-
-Unofficial builds of stable tags can be found in Fedora Copr:
-[pschyska/alacritty](https://copr.fedorainfracloud.org/coprs/pschyska/alacritty/).
-
-``` sh
-dnf copr enable pschyska/alacritty
-dnf install alacritty
-```
```
kchibisov, line 67: We should likely keep those somewhere in INSTALL.md, since this is a know third party repo for fedora. And so, you can't find it on your own with package manager.

INSTALL.md
```diff
@@ -213,62 +237,76 @@ filling in this section of the README.
...
 ```
 
 If all goes well, this should place a binary at `target/release/alacritty`.
 On Windows this directory should also contain the `winpty-agent.exe`.
 
-#### Desktop Entry
-
-Many linux distributions support desktop entries for adding applications to
-system menus. To install the desktop entry for Alacritty, run
+### MacOS
```
nixpulvis, line 247:
```suggestion
### macOS
```

nixpulvis: I'm also curious why macOS gets a full build (not just the binary) while the other platforms have additional components through the Post Build. I feel like the whole Building section could remove the platform subsections, altogether.

INSTALL.md
```diff
@@ -213,62 +237,76 @@ filling in this section of the README.
...
 ```
 
-You can find a prerendered SVG logo as well as simplified versions of the SVG in
-the `extra/logo/compat` directory.
+## Post Build
 
-### MacOS
+There are some extra things you might want to set up after installing Alacritty.
+All the post build instruction assume you're still inside the Alacritty
+repository.
```
nixpulvis, line 258: This is a tiny point, but I would think, ... inside the `alacritty/` directory is a bit more clear.

INSTALL.md
chrisduerr: I'm not sure what you mean?

INSTALL.md
chrisduerr: No, that would just be more confusing because there's an `alacritty` directory in the repo and you can name the clone whatever you want. So saying you should be inside the repo makes most sense.

INSTALL.md
nixpulvis: Like an "App" build should be self-contained (or at least to the best of it's ability), which includes things like the icon. Futhermore, `make app` itself is more or less the macOS equivalent of the whole Post Build section, with the exception of not actually installing things. It would be nice if the man pages, and terminfo files could be served directly from the App bundle itself, but to my knowledge, that's not really an option.

So I just think the process of building on macOS is basically the same as Linux/Windows, it's the Post Build where whole `Makefile` seems most relevant.

chrisduerr: But you don't want to build like on linux? You want to `make app`? Why would you `cargo build --release` just to `make app` which calls `cargo build --release` again. I don't understand what your goal is.

nixpulvis: I guess this is all in the "Manual Installation" section, so I can see why it makes sense to just take the shortest path to that end for macOS. That said, `make app` does not install the man page for example. So I guess macOS is just in a bit of a strange situation here. It should all be fine though, since the Post Build instructions should still apply, and the files packaged into the App bundle are mostly there in case someone doesn't have access to the source, which is where this INSTALL.md is located.

Basically, perhaps it could be made more clear, but I see no issue with this now.

chrisduerr: > That said, make app does not install the man page for example. So I guess macOS is just in a bit of a strange situation here.

Yes, that's why I moved the post build instructions to a completely separate step. Getting the auxiliary files from the app bundle instead of the repository when you're already inside it is just more work than it's worth. So easier to just document it the same for all platforms.

PR #4031, opened 2020-07-25
Invert fixed color cursor if it matches cell bg

### LINKED ISSUE #4016: Cursor should be inverted when placed on same foreground color
### System
OS: Linux (but probably all)
Version: alacritty 0.5.0-dev (87e5b1aa)
Linux/BSD: X11

Here is an example in urxvt: 
image

The same behavior in alacritty:
image

I think usability is greatly improved with this behavior.

chrisduerr:
This should reduce the number of times people with fixed cursor colors
run into troubles when existing text is already colored.

Using just the background color as a metric instead of both background
and foreground color should ensure that the cursor still has a clear
shape, since just changing the foreground color for a cursor might be
difficult to see. Always inverting the entire cursor instead of keeping
the fixed foreground color is important to make sure the contrast isn't
messed up.

Fixes #4016.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Cursors are now inverted when their fixed color matches the cell's background
+- Cursors are now inverted when their fixed color is similar to the cell's background
 
 ## 0.5.0-dev
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Secondary device attributes escape (`CSI > 0 c`)
 
+### Changed
+
+- Cursors are now inverted when their fixed color matches the cell's background
+
 ## 0.5.0-dev
 
 ### Packaging
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -31,6 +31,9 @@ mod search;
 /// Max size of the window title stack.
 const TITLE_STACK_MAX_DEPTH: usize = 4096;
 
+/// Minimum contrast between a fixed cursor color and the cell's background.
+const MIN_CURSOR_CONTRAST: f64 = 1.5;
+
 /// Maximum number of linewraps followed outside of the viewport during search highlighting.
 const MAX_SEARCH_LINES: usize = 100;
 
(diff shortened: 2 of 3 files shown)


kchibisov:
I do like the idea in general, but I'd like to see some error range for those checks actually, since if colors different on `1` in some component you won't see a difference anyway, but the color won't be inverted.

I can't suggest a range that we can use, but in general it's something I'd like to explore. Though, I don't want something very complex here, and so it should be done in RGB colors and without much math.

chrisduerr:
> I do like the idea in general, but I'd like to see some error range for those checks actually

I have considered this but I find this **far** to complicated. This is not about fixing things with truecolor, but mostly about conflicts within your own colorscheme. There's no good way to fix this with truecolor anyways, so I don't think it's worth even trying. Having to compute the "distance" just seems like far too much trouble.

kchibisov:
Yeah, but your colorscheme colors could also differ just a bit, like white and bright white for example, or your cursor won't much any color of from them, and so you won't solve that issue within your colorscheme, unless you use exactly the same colors.

Though, I guess the way it's right now is fine, but I guess checking distance between colors, assuming rgb vectors, isn't something that won't work for certain cases? Though, I guess distance won't work well, and so it should be something like max difference among color components, to make it work nicer, which isn't something that complex tbqh.

chrisduerr:
> Though, I guess the way it's right now is fine, but I guess checking distance between colors, assuming rgb vectors, isn't something that won't work for certain cases? Though, I guess distance won't work well, and so it should be something like max difference among color components, to make it work nicer, which isn't something that complex tbqh.

It should probably be something like color contrast, but some colorschemes have a relatively small contrast between things to begin with and we wouldn't want to always invert it in those cases. Computing contrast also is a bit annoying. It's definitely not "simple".

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Secondary device attributes escape (`CSI > 0 c`)
 
+### Changed
+
+- Cursors are now inverted when their fixed color matches the cell's background
```
kchibisov, line 15: We should mention that it's now contrast based.

alacritty_terminal/src/term/mod.rs
```diff
@@ -389,13 +392,19 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
...
             {
-                // Handle cell below cursor.
                 if self.cursor.rendered {
+                    // Handle cell below cursor.
                     let cell = self.inner.next()?;
                     let mut cell = RenderableCell::new(self, cell);
 
                     if self.cursor.key.style == CursorStyle::Block {
-                        cell.fg = self.cursor.text_color.color(cell.fg, cell.bg);
+                        // Invert cursor if static background matches the cell's background.
```
kchibisov, line 401: Actually, both comments are a bit out dated, since it's not really a match, it's similar to.

alacritty_terminal/src/term/mod.rs
```diff
@@ -412,7 +421,13 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
 
                     let mut cell = RenderableCell::new(self, cell);
                     cell.inner = RenderableCellContent::Cursor(self.cursor.key);
-                    cell.fg = self.cursor.cursor_color.color(cell.fg, cell.bg);
+
+                    // Only apply static color if it doesn't match the cell's current background.
```
kchibisov, line 425: same here.

PR #4036, opened 2020-07-26
Fix CellForeground and CellBackground cursor colors

kchibisov:
This commit fixes regression introduced in
bedf5f3004e8f33011925ca471be02ead96f4581, when setting
certain CellForeground/CellBackground combinations stoppped working.

For example setting CellForeground for both foreground and background.

--
I do remember mentioning it on IRC, but I fall a sleep before commenting on PR and so it got merged, I guess.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -399,8 +399,8 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
 
                     if self.cursor.key.style == CursorStyle::Block {
                         cell.fg = match self.cursor.cursor_color {
-                            // Invert cursor if static background is close to the cell's
-                            // background.
+                            // Apply cursor color, or invert the cursor if it has a fixed background
+                            // close to the cell's background.
                             CellRgb::Rgb(col) if col.contrast(cell.bg) < MIN_CURSOR_CONTRAST => {
                                 cell.bg
                             },
@@ -423,8 +423,8 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
                     let mut cell = RenderableCell::new(self, cell);
                     cell.inner = RenderableCellContent::Cursor(self.cursor.key);
 
-                    // Only apply static color if it isn't close to the cell's current
-                    // background or refers to cell colors.
+                    // Apply cursor color, or invert the cursor if it has a fixed background close
+                    // to the cell's background.
                     match self.cursor.cursor_color {
                         CellRgb::Rgb(color) if color.contrast(cell.bg) < MIN_CURSOR_CONTRAST => (),
                         _ => cell.fg = self.cursor.cursor_color.color(cell.fg, cell.bg),
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -423,14 +423,12 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
                     let mut cell = RenderableCell::new(self, cell);
                     cell.inner = RenderableCellContent::Cursor(self.cursor.key);
 
-                    cell.fg = match self.cursor.cursor_color {
-                        // Only apply static color if it isn't close to the cell's current
-                        // background.
-                        CellRgb::Rgb(color) if color.contrast(cell.bg) < MIN_CURSOR_CONTRAST => {
-                            cell.fg
-                        },
-                        _ => self.cursor.cursor_color.color(cell.fg, cell.bg),
-                    };
+                    // Only apply static color if it isn't close to the cell's current
+                    // background or refers to cell colors.
+                    match self.cursor.cursor_color {
+                        CellRgb::Rgb(color) if color.contrast(cell.bg) < MIN_CURSOR_CONTRAST => (),
+                        _ => cell.fg = self.cursor.cursor_color.color(cell.fg, cell.bg),
+                    }
 
                     return Some(cell);
                 }
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -398,14 +398,19 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
                     let mut cell = RenderableCell::new(self, cell);
 
                     if self.cursor.key.style == CursorStyle::Block {
-                        // Invert cursor if static background is close to the cell's background.
-                        match self.cursor.cursor_color {
-                            CellRgb::Rgb(col) if col.contrast(cell.bg) >= MIN_CURSOR_CONTRAST => {
-                                cell.fg = self.cursor.text_color.color(cell.fg, cell.bg);
+                        cell.fg = match self.cursor.cursor_color {
+                            CellRgb::Rgb(col) => {
+                                // Invert cursor if static background is close to the cell's
+                                // background.
+                                if col.contrast(cell.bg) >= MIN_CURSOR_CONTRAST {
+                                    self.cursor.text_color.color(cell.fg, cell.bg)
+                                } else {
+                                    cell.bg
+                                }
                             },
-                            CellRgb::Rgb(_) => cell.fg = cell.bg,
-                            _ => cell.fg = self.cursor.text_color.color(cell.fg, cell.bg),
-                        }
+                            CellRgb::CellForeground => cell.fg,
+                            CellRgb::CellBackground => cell.bg,
+                        };
                     }
 
                     return Some(cell);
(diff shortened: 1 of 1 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -403,7 +403,8 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
                             CellRgb::Rgb(col) if col.contrast(cell.bg) >= MIN_CURSOR_CONTRAST => {
                                 cell.fg = self.cursor.text_color.color(cell.fg, cell.bg);
                             },
-                            _ => cell.fg = cell.bg,
+                            CellRgb::Rgb(_) => cell.fg = cell.bg,
+                            _ => cell.fg = self.cursor.text_color.color(cell.fg, cell.bg),
```
chrisduerr, line 407: Just use `< MIN_CURSOR_CONTRAST` here.

alacritty_terminal/src/term/mod.rs
```diff
@@ -427,6 +428,8 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
                         if color.contrast(cell.bg) >= MIN_CURSOR_CONTRAST {
                             cell.fg = self.cursor.cursor_color.color(cell.fg, cell.bg);
                         }
+                    } else {
+                        cell.fg = self.cursor.cursor_color.color(cell.fg, cell.bg);
                     }
```
chrisduerr, line 433: This should be transformed into a match statement matching the other one.

alacritty_terminal/src/term/mod.rs
```diff
@@ -398,13 +398,19 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
...
+                                    self.cursor.text_color.color(cell.fg, cell.bg)
+                                } else {
+                                    cell.bg
+                                }
                             },
-                            _ => cell.fg = cell.bg,
-                        }
+                            CellRgb::CellForeground => cell.fg,
+                            CellRgb::CellBackground => cell.bg,
+                        };
```
chrisduerr, line 413: Can you explain why you did not follow my suggestion here?

kchibisov: I don't get your idea to translate to match statement. In what way do you want it to be? Like that?

```rust
match self.cursor.cursor_color {
  CellRgb::Rgb(col) => { /* something */ },
  _ => /* reset */,
}
```

chrisduerr:
```
match self.cursor.cursor_color {
  CellRgb::Rgb(col) if col < MIN_CONTRAST => { /* do special thing */ },
  _ => /* do thing we've always done */,
}
```

chrisduerr: The only difference should be what happens when it's Rgb and the contrast is smaller than the minimum, all other cases just do the same thing as before.

alacritty_terminal/src/term/mod.rs
```diff
@@ -422,12 +423,14 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
...
 
-                    // Only apply static color if it isn't close to the cell's current background.
-                    if let CellRgb::Rgb(color) = self.cursor.cursor_color {
-                        if color.contrast(cell.bg) >= MIN_CURSOR_CONTRAST {
-                            cell.fg = self.cursor.cursor_color.color(cell.fg, cell.bg);
-                        }
-                    }
+                    cell.fg = match self.cursor.cursor_color {
+                        // Only apply static color if it isn't close to the cell's current
+                        // background.
```
chrisduerr, line 428: This comment isn't accurate anymore.

alacritty_terminal/src/term/mod.rs
```diff
@@ -422,12 +423,14 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
...
+                        CellRgb::Rgb(color) if color.contrast(cell.bg) < MIN_CURSOR_CONTRAST => {
+                            cell.fg
```
chrisduerr, line 430: Why would we set the cell.fg to cell.fg?

kchibisov: I did it just for consistency with other `match` I've added, but there's no need to do it that way here, I guess.

alacritty_terminal/src/term/mod.rs
```diff
@@ -422,11 +423,11 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
                     let mut cell = RenderableCell::new(self, cell);
                     cell.inner = RenderableCellContent::Cursor(self.cursor.key);
 
-                    // Only apply static color if it isn't close to the cell's current background.
-                    if let CellRgb::Rgb(color) = self.cursor.cursor_color {
-                        if color.contrast(cell.bg) >= MIN_CURSOR_CONTRAST {
-                            cell.fg = self.cursor.cursor_color.color(cell.fg, cell.bg);
-                        }
+                    // Only apply static color if it isn't close to the cell's current
+                    // background or refers to cell colors.
```
chrisduerr, line 427: The latter part is already implied by `static`.

Maybe something like this is more clear:

```suggestion
                    // Apply cursor color, or invert the cursor if it has a fixed background close to the cell's background.
```

I'm not sure which of these I like more. The one describes the purpose of this block better, but describes something that isn't done, while the other isn't entirely clear

alacritty_terminal/src/term/mod.rs
```diff
@@ -398,13 +398,14 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
                     let mut cell = RenderableCell::new(self, cell);
 
                     if self.cursor.key.style == CursorStyle::Block {
-                        // Invert cursor if static background is close to the cell's background.
-                        match self.cursor.cursor_color {
-                            CellRgb::Rgb(col) if col.contrast(cell.bg) >= MIN_CURSOR_CONTRAST => {
-                                cell.fg = self.cursor.text_color.color(cell.fg, cell.bg);
+                        cell.fg = match self.cursor.cursor_color {
+                            // Invert cursor if static background is close to the cell's
+                            // background.
```
chrisduerr, line 403:
```suggestion
                            // Apply cursor color, or invert the cursor if it has a fixed background close to the cell's background.
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -422,11 +423,11 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
...
-                    // Only apply static color if it isn't close to the cell's current background.
-                    if let CellRgb::Rgb(color) = self.cursor.cursor_color {
-                        if color.contrast(cell.bg) >= MIN_CURSOR_CONTRAST {
-                            cell.fg = self.cursor.cursor_color.color(cell.fg, cell.bg);
-                        }
+                    // Only apply static color if it isn't close to the cell's current
+                    // background or refers to cell colors.
+                    match self.cursor.cursor_color {
+                        CellRgb::Rgb(color) if color.contrast(cell.bg) < MIN_CURSOR_CONTRAST => (),
+                        _ => cell.fg = self.cursor.cursor_color.color(cell.fg, cell.bg),
```
chrisduerr, line 430: I think the following would be better, it requires 1.42.0 though. But chances are that's fine for 0.6.0?

```suggestion
                    if !matches(self.cursor.cursor_color, CellRgb::Rgb(color) if color.contrast(cell.bg) < MIN_CURSOR_CONTRAST) {
                        cell.fg = self.cursor.cursor_color.color(cell.fg, cell.bg);
```

kchibisov: I'd bump straight to `v0.4.3` if anything in a separate PR, are you fine with that?

chrisduerr: 0.4.3?

chrisduerr: In general I'm opposed to bumping minimum Rust version separately, since that makes it harder to track down the original reason for the bump.

kchibisov: >0.4.3?

Yeah, I want `0.4.3` for cfg before if.

chrisduerr: You mean 1.43.0?

chrisduerr: Also I've looked at the other PR and if you want to bump it for multiple different reasons anyways, I'd recommend merging this *without* the `matches!` macro and then fixing it immediately in the rust version bump PR.

kchibisov: ok.

PR #4037, opened 2020-07-26
Document rejected window ops escapes

### LINKED ISSUE #4035: Alacritty doesn't support the resize window CSI sequence
### System

OS: Linux (EndeavorOS)
Version: alacritty 0.4.3
Linux/BSD: Xfce4

### Alacritty doesn't support the resize window CSI sequence

According to the [XTerm documentation](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h4-Functions-using-CSI-_-ordered-by-the-final-character-lparen-s-rparen:CSI-Ps;Ps;Ps-t:Ps-=-8;height;width.2519), the CSI sequences listed bellow should resize Alacritty into given size.

* `echo -n "\e[8;20;80t"` should set form size to height=20, width=80
* `echo -n "\e[8;30t"` should set form size to height=30
* `echo -n "\e[8;;90t"` should set form size to width=90
* `echo -n "\e[8;0;0t"` for xterm, the form will be maximized

chrisduerr:
Fixes #4035.

### DIFF
--- a/docs/escape_support.md
+++ b/docs/escape_support.md
@@ -73,6 +73,7 @@ brevity.
 | `CSI s`    | IMPLEMENTED |                                                   |
 | `CSI T`    | IMPLEMENTED |                                                   |
 | `CSI t`    | PARTIAL     | Only parameters `22` and `23` are supported       |
+|            |             | Rejected: `1`-`13`, `15`, `19`-`21`, `24`         |
 | `CSI u`    | IMPLEMENTED |                                                   |
 | `CSI X`    | IMPLEMENTED |                                                   |
 | `CSI Z`    | IMPLEMENTED |                                                   |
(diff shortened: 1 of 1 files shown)


docs/escape_support.md
```diff
@@ -73,6 +73,7 @@ brevity.
 | `CSI s`    | IMPLEMENTED |                                                   |
 | `CSI T`    | IMPLEMENTED |                                                   |
 | `CSI t`    | PARTIAL     | Only parameters `22` and `23` are supported       |
+|            |             | Rejected: `1`-`13`, `15`, `19`-`21`, `24`         |
```
kchibisov, line 76: Maybe we can move rejected to a left column?

chrisduerr: Not a bad idea, sure!

Issue #4038, opened 2020-07-27
Sometimes cjk character can only see the left half


flw-cn:
Sometimes the cjk character and the spacer after it are assigned to a different texture atlas, resulting in only the left half of the character being drawn.

<img width="884" alt="image" src="https://user-images.githubusercontent.com/5546718/88493741-5d6b6080-cfe5-11ea-9773-9228a8602a9c.png">

The chinese character <img width="28" alt="image" src="https://user-images.githubusercontent.com/5546718/88494117-18482e00-cfe7-11ea-8da9-514a1bdc4f3e.png"> is drawn as <img width="35" alt="image" src="https://user-images.githubusercontent.com/5546718/88494059-d61eec80-cfe6-11ea-93a2-710d8365e0f2.png">

### System

OS: macOS
Version: alacritty 0.6.0-dev (dd32447)

### Logs

```
[2020-07-27 08:41:02.638338000] [DEBUG] Using environment locale: en_US.UTF-8/UTF-8/en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/en_US.U
TF-8
[2020-07-27 08:41:02.646170000] [INFO] Welcome to Alacritty
[2020-07-27 08:41:02.646357000] [INFO] Configuration loaded from "/Users/flw/.config/alacritty/alacritty.yml"
[2020-07-27 08:41:02.767907000] [DEBUG] Estimated DPR: 2
[2020-07-27 08:41:02.767989000] [DEBUG] Estimated Cell Size: 21 x 45
[2020-07-27 08:41:02.768034000] [DEBUG] Estimated Dimensions: None
[2020-07-27 08:41:02.958046000] [INFO] Device pixel ratio: 2
[2020-07-27 08:41:02.961405000] [INFO] Initializing glyph cache...
[2020-07-27 08:41:03.041121000] [INFO] ... finished initializing glyph cache in 0.079594508s
[2020-07-27 08:41:03.041218000] [INFO] Cell Size: 21 x 45
[2020-07-27 08:41:03.041248000] [INFO] Padding: 0 x 0
[2020-07-27 08:41:03.041280000] [INFO] Width: 1600, Height: 1200
[2020-07-27 08:41:03.043164000] [INFO] PTY dimensions: Line(26) x Column(76)
[2020-07-27 08:41:03.055212000] [INFO] Initialisation complete
[2020-07-27 08:41:03.557021000] [DEBUG] New num_cols is 121 and num_lines is 35
[...]

PR #4045, opened 2020-07-28
Render wide characters correctly to avoid being cut off

### LINKED ISSUE #791: Glyphs sometimes cut off (need to draw background in single pass)

flw-cn:
The commit contains a realistically tested workaround with no performance impact.

I made a few necessary changes to the shaders in the GLSL code, fixing the
background width of the wide characters and skipping everything about spacer
handling (in fact, they shouldn't be displayed on the screen at all, either
foreground or background).

This should close #791, cc #4038.

### DIFF
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -421,6 +421,8 @@ struct InstanceData {
     bg_a: f32,
     // Flag indicating that glyph uses multiple colors, like an Emoji.
     multicolor: u8,
+    // Set to 1 if glyph is a WIDE_CHAR.
+    is_wide_char: u8,
 }
 
 #[derive(Debug)]
@@ -497,6 +499,7 @@ impl Batch {
             bg_b: f32::from(cell.bg.b),
             bg_a: cell.bg_alpha,
             multicolor: glyph.colored as u8,
+            is_wide_char: cell.flags.contains(Flags::WIDE_CHAR) as u8,
         });
     }
 
@@ -651,6 +654,17 @@ impl QuadRenderer {
             );
             gl::EnableVertexAttribArray(5);
             gl::VertexAttribDivisor(5, 1);
+            // WideChar flag.
+            gl::VertexAttribPointer(
+                6,
+                1,
+                gl::BYTE,
+                gl::FALSE,
+                size_of::<InstanceData>() as i32,
+                (17 * size_of::<f32>() + 1) as *const _,
+            );
+            gl::EnableVertexAttribArray(6);
+            gl::VertexAttribDivisor(6, 1);
 
             // Rectangle setup.
             gl::GenVertexArrays(1, &mut rect_vao);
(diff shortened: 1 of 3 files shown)


res/text.v.glsl
```diff
@@ -44,6 +51,11 @@ uniform int backgroundPass;
 
 void main()
 {
+    skipSpacer = isSpacer;
+    if (skipSpacer != 0) {
+        return;
+    }
```
chrisduerr, line 57: I don't think it makes sense to send data to the shader, just to always discard it?

alacritty/src/renderer/mod.rs
```diff
@@ -421,6 +421,10 @@ struct InstanceData {
     bg_a: f32,
     // Flag indicating that glyph uses multiple colors, like an Emoji.
     multicolor: u8,
+    // Set to 1 if glyph is a WIDE_CHAR.
+    is_wide_char: u8,
+    // Set to 1 if glyph is a WIDE_CHAR_SPACER.
+    is_spacer: u8,
```
kchibisov, line 427: I guess we can reuse multicolor value and call it like cell_flags and just pass all three values in it. Having 3 boolean flags to pass to GPU is wasteful to me.

flw-cn: Ok, let me have a try.

res/text.v.glsl
flw-cn: Are you saying we should just remove the spacer rendering operation from the Rust code?
Let me have a try.

chrisduerr: Yes. We are already filtering out whitespace characters. So if the spacer is not needed for rendering, we should filter it out. Though make sure to filter both the leading and trailing spacers.

flw-cn:
I've discarded the spacer rendering action.

But when I try to reduce the flags I run into a lot of trouble.

* The smallest GLSL data types are 32-bit, but flags use `u8` to save space (if we use `u32`, it might be simpler to use four `u8` instead).
* After I pass the flags of u8, the bit is not easily and intuitively retrieved in the GLSL script, in fact, there is a problem with the bit-order. For example, if I pass `flags = 0x02`, I will receive `0x4000_0000` in GLSL.
* The FLAG_NAME/VALUE in the GLSL script must also correspond to the one in Rust, but this can easily go wrong during maintenance late on.
* Even if I'm lucky enough to do it, I don't have the energy to test all other platforms to make sure it's universally correct.

In summary, I'm not going to merge the flags, and I've kept a separate `WIDE_CHAR` flag.

If you have a better suggestion on what to do, please let me know. @chrisduerr @kchibisov

alacritty/src/renderer/mod.rs
```diff
@@ -1023,6 +1037,12 @@ impl<'a> RenderApi<'a> {
 
     #[inline]
     fn add_render_item(&mut self, cell: RenderableCell, glyph: &Glyph) {
+        if cell.flags.contains(Flags::WIDE_CHAR_SPACER)
+            && !cell.flags.contains(Flags::LEADING_WIDE_CHAR_SPACER)
+        {
+            return;
+        }
```
chrisduerr, line 1044: Things should be filtered out earlier, not in the renderer.

flw-cn:
Anyway, I did my best. I hope my fork can give you some inspiration. Others are also welcome to follow up to resolve this issue once and for all.

chrisduerr:
Closing since @flw-cn does not seem interested in finishing up this PR.

PR #4052, opened 2020-07-30
Fix 0.5.0 changelog

chrisduerr:


CHANGELOG.md
```diff
@@ -30,6 +30,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 - Environment variable `RUST_LOG` for selecting the log level
+- Deprecated `window.start_maximized` config field
+- Deprecated `render_timer` config field
+- Deprecated `persistent_logging` config field
```
kchibisov, line 35: Could you move that section after the `Fixed` one? Just for consistency with the other changelog, entries, I guess that's why I've put it there initially, and so we end up with 2 identical sections.

chrisduerr: I think it's more appropriate up there? Since it's basically one of the most relevant sections, so it should be at the top?

There is no consistent order in the past iirc.

PR #4057, opened 2020-08-01
Add notes about version updates on master

kchibisov:
All version updates on master branch don't represent stable releases
in anyway, so add note that it should be mentioned in the commits, which
refer to releases.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -137,6 +137,9 @@ The exact steps for an exemplary `0.2.0` release might look like this:
      description
  14. The `-dev` is stripped from the `0.2.0-dev` changelog entries on master
 
+All commits to master branch, which speculates with version number, like on
+steps `3` and `14`, should state that they don't represent a stable release.
+
 On master and with new planned releases, only the minor version is bumped. This makes it possible to
 create bug fix releases by incrementing the patch version of a previous minor release, without
 having to adjust the next planned release's version number.


CONTRIBUTING.md
```diff
@@ -137,6 +137,9 @@ The exact steps for an exemplary `0.2.0` release might look like this:
      description
  14. The `-dev` is stripped from the `0.2.0-dev` changelog entries on master
 
+All commits to master branch, which speculates with version number, like on
+steps `3` and `14`, should state that they don't represent a stable release.
```
chrisduerr, line 141: This doesn't really make sense to me, could you reword it a bit? Also I'm not sure if we explicitly need to point this out since this section is getting quite complicated? So we should omit everything that isn't absolutely necessary.

kchibisov: Well, it seems like it's necessary, don't you think? Especially since you basically require it to be presented. I do agree that it's a bit too much at this point, and I had some troubles even phrasing it initially, so I'm not entirely sure how I'd rephrase it.

chrisduerr: I mean it's not really necessary, it's just a precaution. There's nothing really terrible that would happen, it's just to prevent issues from popping up before they even exist. If we don't document this and it's omitted, the worst that could happen is that people would figure it out on their own, right?

kchibisov: that's fair.

You can close that PR, since I don't really like it myself.

chrisduerr: I'll close this for now, with the note that the entire section can probably be streamlined a bit in the future.

PR #4058, opened 2020-08-01
Correct various config options

joshukraine:
This resolves the following deprecation warning: "Config dynamic_title is deprecated; please use window.dynamic_title instead"

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -65,6 +65,9 @@
   # Window title
   #title: Alacritty
 
+  # Allow terminal applications to change Alacritty's window title.
+  #dynamic_title: true
+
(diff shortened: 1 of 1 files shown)


kchibisov:
Could you place it after window.title option?

joshukraine:
> Could you place it after window.title option?

Done! ✅

joshukraine:
> Thanks a lot for your contribution!
> 
> I've actually went through all the config options to see if there's anything else we forgot to move around in our library rework and there's one other thing I've noticed.
> 
> The config options `colors.search.bar` do not work when uncommented. The idea isn't wrong, but the format simply isn't supported at this location. Can you change that too please?

Done! ✅

PR #4062, opened 2020-08-01
Add note regarding release tags

adrian5:
I think this is worth mentioning, now that more Linux users may decide to build from source themselves. They likely want to pick a stable release and not the current state of `master`.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -43,6 +43,13 @@ git clone https://github.com/alacritty/alacritty.git
 cd alacritty
 ```
 
+You may then also want to pick a specific (stable) release to compile, e.g.:
+
+```sh
+git tag
+git checkout v0.5.0
+```
+
 ### Install the Rust compiler with `rustup`
 
 1. Install [`rustup.rs`](https://rustup.rs/).


adrian5:
My thinking was that users of popular distros like Ubuntu and Mint will realize there's no package for Alacritty (anymore), and end up at [README.md](https://github.com/alacritty/alacritty#installation) → [INSTALL.md](https://github.com/alacritty/alacritty/blob/master/INSTALL.md#prerequisites) which, if followed verbatim, doesn't build the latest stable release.

chrisduerr:
> there's no package for Alacritty (anymore)

There's no change to the Ubuntu/Mint packaging situation. If you're talking about the PPA being removed, that's solely a temporary measure to ensure the maintenance status of it can keep up. It already is updated to 0.5.0.

I'm going to close this PR for now, since I haven't yet heard any reports of people that ran into problems because they were running the latest master instead of stable version. So as far as I can tell this isn't really a problem. Should this turn out to be a problem, we can easily fix this in the future, but generally when installing from source the latest master is what you want. And someone capable of installing from source should be able to pick a release tag. That way we don't bloat up our documentation with unnecessary detail, which could make other things harder to find.

adrian5:
Ok, fair enough.

PR #4078, opened 2020-08-03
Add support for colon separated SGR parameters

### LINKED ISSUE #1485: bring 24-bit (aka truecolor) support up to date
Alacritty implements 24-bit colors through a semicolon-separated escape sequence (38;2;...). However, the preferred escape sequence is now colon-separated, with an optional color space argument prefixing the rgb arguments.

This is perhaps best described in the comment for the parse_extended_colors() function in xterm-333/charproc.c:

Note that xterm actually supports *three* variants: semicolor-separated, colon-separated, and a mixed version. It's unclear to me why it supports the third variant.

> Some background -
> 
> Todd Larason provided the initial changes to support 256-colors in July 1999.
> I pointed out that the description of SGR 38/48 in ECMA-48 was vague, and
> was unsure if there would be some standard using those codes.  His response
> was that this was documented (it turns out, in equally vague terms) in ITU
> T.416
> 
> Discussing this with Todd Larason in mid-1999, my point was that given the
> high cost of obtaining ITU T.416 (ISO-8613-6), the standard was not going
> to be effective (more than $100 then, more than $200 in 2012)
> 
> We overlooked the detail about ":" as a subparameter delimiter (documented
> in 5.4.t2 in ECMA-48).  Some discussion in KDE in mid-2006 led Lars Doelle
> to discuss the issue with me.  Lars' initial concern dealt with the fact
> that a sequence such as
>  CSI 38 ; 5 ; 1 m
> violated the principle that SGR parameters could be used in any order.
> Further discussion (see KDE #107487) resolved that the standard expected
> that the sequence would look like
>  CSI 38 ; 5 : 1 m
> which still violates that principle, since the "5:1" parameter has to
> follow the "38" to be useful.
> 
> This function accepts either format (per request by Paul Leonerd Evans).
> It also accepts
>  CSI 38 : 5 : 1 m
> according to Lars' original assumption.  While implementing that, I added
> support for Konsole's interpretation of "CSI 38 : 2" as a 24-bit RGB value.
> ISO-8613-6 documents that as "direct color".
> 
> At the time in 2012, no one noticed (or commented) regarding ISO-8613-6's
> quirk in the description of direct color:  it mentions a color space
> identifier parameter which should follow the "2" (as parameter 1).  In the
> same section, ISO-8613-6 mentions a parameter 6 which can be ignored, as
> well as parameters 7 and 8.  Like parameter 1, parameters 7 and 8 are not
> defined clearly in the standard, and a close reading indicates they are
> optional, saying they "may be used".  This implementation ignores parameters
> 6 (and above), and provides for the color space identifier by checking the
> number of parameters:
>  3 after "2" (no color space identifier)
>  4 or more after "2" (color space identifier)
> 
> By the way - all of the parameters are decimal integers, and missing
> parameters represent a default value.  ISO-8613-6 is clear about that.

Furthermore, as of ncurses 6.1, the correct way to tell applications that the terminal supports 24-bit colors is by setting the terminfo setaf/setab capabilities. The Tc option is, as far as I can tell, obsolete. It is not used at all in any terminfo entries in ncurses 6.1.

The terminfo database provided with ncurses 6.1 defines a couple of "features" related to this: *xterm+direct*, *xterm+direct2* and *xterm+indirect*.

*indirect* corresponds to the escape sequence implemented by alacritty today, i.e. semicolon-separated.

*direct2* corresponds to the colon-separated version *without* the color space argument, and *direct* corresponds to the colon-separated version *with* the (optional) color space argument.

One can "use" these features to get the correct setaf/setab capabilities.

The alacritty.info file could/should be updated to use one of these definitions. As a first step, use the indirect version, and switch to the direct one if/when the colon-separated escape sequence is implemented.

FYI, I find the current alacritty.info to be a bit weird, in that it uses xterm-256color in the alacritty definition. If I set TERM to a non-256color entry, it's because _I don't want_ 256 colors...

I've been using the following replacement for some time now, with success. 24-bit support works in both vim and emacs.

```
alacritty| alacritty,
	use=xterm,

alacritty-256color| alacritty with 256 colors,
	use=xterm+256color, use=alacritty,

alacritty-direct| alacritty with direct-color indexing,
	use=xterm+indirect, use=alacritty,
```

chrisduerr:
This implements the colon separated form of SGR 38 and 48.

Fixes #1485.


kchibisov:
>Could you benchmark this? There's no performance difference on my system.

On my neither, the perf in a margin of error here. I was comparing master with the git version of vte fwiw, since it's more fair.

```
master: 4.6125 +- 0.0432 seconds time elapsed  ( +-  0.94% )
#4078: 4.61421 +- 0.00997 seconds time elapsed  ( +-  0.22% )
```

So basically the same.

alacritty/src/event.rs
```diff
@@ -1013,18 +1013,18 @@ impl<N: Notify + OnResize> Processor<N> {
...
-            },
+                    | WindowEvent::TouchpadPressure { .. }
+                    | WindowEvent::CursorEntered { .. }
+                    | WindowEvent::AxisMotion { .. }
+                    | WindowEvent::HoveredFileCancelled
+                    | WindowEvent::Destroyed
+                    | WindowEvent::HoveredFile(_)
+                    | WindowEvent::Touch(_)
+                    | WindowEvent::Moved(_)
+            ),
```
chrisduerr, line 1027: This is a new clippy suggestion.

alacritty/src/window.rs
```diff
@@ -261,7 +256,7 @@ impl Window {
             .with_title(title)
             .with_visible(false)
             .with_transparent(true)
-            .with_decorations(decorations)
+            .with_decorations(window_config.decorations != Decorations::None)
```
chrisduerr, line 259: It was suggested by clippy to use `matches!`, but I think PartialEq is better here.

alacritty_terminal/Cargo.toml
```diff
@@ -14,7 +14,7 @@ bitflags = "1"
 parking_lot = "0.10.2"
 serde = { version = "1", features = ["derive"] }
 serde_yaml = "0.8"
-vte = { version = "0.8.0", default-features = false }
+vte = { git = "https://github.com/alacritty/vte", default-features = false }
```
chrisduerr, line 17: I'd merge this against the git branch for now, then we can test the new library ourselves before making a release.

kchibisov: I don't like having git version here in general. Maybe we should create a vte beta release, I guess it'll make more sense?

alacritty/src/event.rs
kchibisov: If those things are new clippy suggestions make them in a separate PR or commit please.

alacritty/src/window.rs
kchibisov: Same here, it has nothing to do with the PR.

alacritty/src/event.rs
chrisduerr: Done https://github.com/alacritty/alacritty/pull/4088.

alacritty_terminal/Cargo.toml
chrisduerr: Good point, this is going to make bisecting in the future impossible, I didn't think about that. If we use a git branch we need to specify the revision.

I feel like pinning to a commit is probably better than just uploading every random commit to crates.io though? Assuming there might be other changes to vte.

I think it might be a nice thing to think about in general, with stuff like crossfont too.

kchibisov: Assuming that commit won't go away, then maybe it's fine to do so, but using versions from crates.io are more safe in general.

chrisduerr: I mean commits should never go away, so I think that's pretty safe.

I just don't want to get into a situation where every time we want to test a development version of one of our crates, we need to start jumping through hoops and release new versions.

If we instead just linked to a specific commit, that should allow us to implement new stuff without having to put in any work.

But of course crates.io is less likely to go away than our github repository.

kchibisov: let's stick to specific hash then and try how it works.

chrisduerr:
> Once you merge this PR, please open a follow up issue to remove git dependency on vte, when the new release will be out, so we won't forget to update Cargo.toml properly for the next release.

I mean we can't publish it with a git dependency, so it's hard to forget completely, but yeah it makes sense to just have it in the milestone.

PR #4086, opened 2020-08-05
Use `tcgetpgrp` to get PID for SpawnNewInstance

### LINKED ISSUE #4082: Determine active process for SpawnNewInstace using tcgetpgrp system call
Using the `tcgetpgrp` system call on the terminal's fd makes it possible to easily get a group of active terminal processes, which allows for a relatively reliable way to determine the active "shell" program. While this returns a group of processes, using any of the processes in the group will likely produce good results.

chrisduerr:
Fixes #4082.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -299,7 +299,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         let args = {
             // Use working directory of controlling process, or fallback to initial shell.
             let mut pid = unsafe { libc::tcgetpgrp(tty::master_fd()) };
-            if pid == -1 {
+            if pid < 0 {
                 pid = tty::child_pid();
             }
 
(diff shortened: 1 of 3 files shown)


alacritty/src/event.rs
```diff
@@ -297,16 +297,20 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
 
         #[cfg(unix)]
         let args = {
+            // Use working directory of controlling process, or fallback to initial shell.
+            let mut pid = unsafe { libc::tcgetpgrp(tty::master_fd()) };
+            if pid == -1 {
```
kchibisov, line 302:
```suggestion
            if pid < 0 {
```

chrisduerr: Should never happen, but better safe than sorry.

kchibisov: it's just a matter of proper error handling.

Issue #4073, opened 2020-08-03
Keybinding Specific Search


adam-kerrigan:
Is it possible to keybind a specific search/keybind specific keystrokes? I've tried passing through characters with a chained keybinding but this pastes the text on the screen rather than into the search box. I wouldn't be sure how to send the return key this way either.

key_bindings:
  - { key: C, mods: Control|Shift,           action: ToggleViMode }
  - { key: C, mods: Control|Shift, mode: Vi, action: SearchBackward}
  - { key: C, mods: Control|Shift, mode: Vi, chars: '[0-9]: \\[.\*:.*\\\] \\$' }

I wanted to be able to copy the output of the last command by selecting between PS1's or copy the most recent occurrence of https.

### System

OS: macOS
Version: alacritty 0.5.0 (a6681e3)

chrisduerr:
You mean for doing stuff like searching for `https://.*` immediately for example?

It's not possible right now, but I've considered this myself as an option to make opening URLs easier.

kerrigoon:
I think both of these options would be good but personally I'd prefer the first as you could pass movements to Vi mode. So then for the https example you could do something like:
1) Toggle Vi mode
2) Search for https
3) Turn on semantic selection
4) pass 'E' to Vi mode
5) open/copy link
and you wouldn't have to worry about the ending of the url, if it was the last thing on the line or whether it wrapped around to the next line

chrisduerr:
You don't and shouldn't pass `E` to vi mode, that's just a binding you can bind yourself. So this usecase would work with both solutions anyways.

kerrigoon:
I've tried to have a fiddle with this and got something working using an extra field in the bindings enum. However I have an issue that I can't find where to turn off the highlighting without calling cancel_search, and cancel_search looses the position of the cursor in normal mode. Do you know how this is done?

chrisduerr:
> cancel_search looses the position of the cursor in normal mode. Do you know how this is done

You cannot change the position of the cursor in normal mode through Alacritty. That is not possible and that's good the way it is.

kerrigoon:
Sorry not the cursor the but the selection of the search. This is in both modes I just mentioned normal because the highlighting isn't an issue in Vi mode as it disappears on exit.

nixpulvis:
[...]

PR #4087, opened 2020-08-05
Added search field to key-bindings

kerrigoon:
Added new field to key-bindings, 'search', to go with 'action',
'command' and 'char'.
Field takes either a String or a map of { string: 'regex string', direction: 'forward' |     'backward' }
Changed the start_search function to take an additional String argument with current acti    ons SearchForward & SearchBackward passing an empty string.

This is usefull for keybindings to open URLs or copy the output from the
last command, for example:

```
key_bindings:
  - { key: G, mods: Command, mode: ~Vi, action: ToggleViMode }
  - { key: G, mods: Command, mode: Vi, search: {string: 'http.?://', direction: "backward"} }
  - { key: G, mods: Command, mode: Vi, action: Open }
  - { key: G, mods: Command, mode: Vi, action: ClearSelection }
  - { key: G, mods: Command, mode: Vi, action: ToggleViMode }
```

Linked Issue: #4073 

> Please make sure to document all user-facing changes in the
> `CHANGELOG.md` file.

Done

> If support for a new escape sequence was added, it should be documented
> in `./docs/escape_support.md`.

N/A

> Since `alacritty_terminal`'s version always tracks the next release, make sure
> that the version is bumped according to semver when necessary.

I haven't changed any versions

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -587,6 +587,17 @@
 #    `args` array of command line parameter strings. For example:
 #       `{ program: "alacritty", args: ["-e", "vttest"] }`
 #
+# - 'search': Search the buffer either forwards (default) or backwards for a string
+#
+#   The search field can be either a string or a map containing a regex `string` and
+#   `direction` string specifying one of "Right" | "Forward" | "Left" | "Backward". For example:
+#       `"http.?:"` or
+#       `{ string: "http.?", direction: "Backward" }`
+#
+#   If the string is empty a prompt is triggered for input. For example:
+#       `""` or
+#       `{ string: "", direction: "Backward" }`
+#
 # And optionally:
 #
 # - `mods`: Key modifiers to filter binding actions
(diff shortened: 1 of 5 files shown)


kerrigoon:
I thought moving the search outside of actions and into its own category would simplify the issue of parameterising and create more flexibility. With this approach the search actions could be removed entirely and all searches just bound through this instead, as you can specify a direction and/or leave the string blank to open the search prompt. 

Parameterising it seemed more complicated to me as if you cold find a nice way to write it in the config file I think it still ends up with multiple search actions. The other actions that take arguments have both been implemented this way, Command and Esc, so I just thought it made sense.

But I agree, if adding other parameterised actions is something you want then having them processed the same would be ideal and creating new fields for each one would just get messy.

chrisduerr:
> I thought moving the search outside of actions and into its own category would simplify the issue of parameterising and create more flexibility.

It doesn't create any flexibility, it just creates a completely different category of things one has to handle and maintain. Making it incredibly difficult to change things in the future. And that makes things incredibly complicated once we want to add similar functionality for other actions.

kerrigoon:
I meant flexibility regarding what you have available to pass to the search function, but I see your point and agree.
Being able to quickly copy things from the keyboard is a feature I really enjoy in terminals so I look forward to it being added.
I haven't got a clue how to do it like that though so I'll close this.

PR #4091, opened 2020-08-06
Add ^C binding to cancel search

### LINKED ISSUE #4089: Exit search using control-c (or control-g) by default
Would it be possible to add control-c and/or control-g as a default key binding to exit the search?

The reason I'm asking is because it was not obvious to me that I had to press escape until I read the docs. This happened a couple of times, and each time I was convinced it should have been an Alacritty bug that made the terminal hang somehow. Users are used to having control-c available for cancelling stuff.
So, for me not really an issue anymore, but it will help new users.

Thanks!

chrisduerr:
Fixes #4089.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Secondary device attributes escape (`CSI > 0 c`)
 - Support for colon separated SGR 38/48
 - New Ctrl+C binding to cancel search
+- New Ctrl+C binding to cancel search and leave vi mode
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Secondary device attributes escape (`CSI > 0 c`)
 - Support for colon separated SGR 38/48
+- New Ctrl+C binding to cancel search
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


kchibisov:
Could we add this binding to exit Vi mode as well? IIRC you can press `ctrl + c` to exit insert mode, so it could make some sense, what do you think?

chrisduerr:
> Could we add this binding to exit Vi mode as well? IIRC you can press ctrl + c to exit insert mode, so it could make some sense, what do you think?

Hmm, I guess I see no reason why not and it could help some users getting out of vi mode. So sure, let's do it.

alacritty/src/config/bindings.rs
```diff
@@ -368,6 +368,7 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
         Escape,                        +TermMode::VI; Action::ClearSelection;
         I,                             +TermMode::VI; Action::ScrollToBottom;
         I,                             +TermMode::VI; Action::ToggleViMode;
+        C,      ModifiersState::CTRL,  +TermMode::VI; Action::ToggleViMode;
```
kchibisov, line 371: You should add it in config as well.

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Secondary device attributes escape (`CSI > 0 c`)
 - Support for colon separated SGR 38/48
+- New Ctrl+C binding to cancel search
+- New Ctrl+C binding to cancel search and leave vi mode
```
kchibisov, line 21: That sounds a bit strange, seems like rebase issue?

chrisduerr: Correct, let me fix that.

PR #4097, opened 2020-08-07
Fix characters swallowed during search

chrisduerr:
This resolves a bug where characters get swallowed when pressing them
after pressing backspace before the backspace key is released.


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Incorrect window location with negative `window.position` config options
 - Slow rendering performance with HiDPI displays, especially on macOS
+- Keys swallowed during search when pressing them right before releasing backspace
```
chrisduerr, line 28: I've documented the case that users are most likely to run into. In theory this can be the case with any key bindings. But I think since this is the most likely people that have run into this can probably extrapolate from it if they really need to.

PR #4098, opened 2020-08-08
Update spec file

grumpey:
Bump rust version to 1.43
Add ncurses and gcc-c++ to build requires so alacritty will build on rawhide (Fedora 33) 
~Switch to pkgconfig for freetype, fontconfig and xcb.~

COPR Build here: https://copr.fedorainfracloud.org/coprs/grumpey/sway-testing/build/1599704/

Thanks


extra/linux/redhat/alacritty.spec
```diff
@@ -7,14 +7,16 @@ URL:           https://github.com/alacritty/alacritty
 VCS:           https://github.com/alacritty/alacritty.git
 Source:        alacritty-%{version}.tar
 
-BuildRequires: rust >= 1.41.0
+BuildRequires: rust >= 1.43.0
```
chrisduerr, line 10: Could you fix this in the CONTRIBUTING.md too?

extra/linux/redhat/alacritty.spec
```diff
@@ -7,14 +7,16 @@ URL:           https://github.com/alacritty/alacritty
...
-BuildRequires: fontconfig-devel
-BuildRequires: libxcb-devel
+BuildRequires: gcc-c++
+BuildRequires: python3
+BuildRequires: pkgconfig(freetype2)
+BuildRequires: pkgconfig(fontconfig)
+BuildRequires: pkgconfig(xcb)
 BuildRequires: desktop-file-utils
-BuildRequires: python36
+BuildRequires: ncurses
```
chrisduerr, line 19: Does Fedora not have a "base" package that is always expected to be installed with things like ncurses? Not much you can do on linux without ncurses installed.

extra/linux/redhat/alacritty.spec
```diff
@@ -7,14 +7,16 @@ URL:           https://github.com/alacritty/alacritty
...
 BuildRequires: cargo
 BuildRequires: cmake
-BuildRequires: freetype-devel
-BuildRequires: fontconfig-devel
-BuildRequires: libxcb-devel
+BuildRequires: gcc-c++
+BuildRequires: python3
+BuildRequires: pkgconfig(freetype2)
+BuildRequires: pkgconfig(fontconfig)
+BuildRequires: pkgconfig(xcb)
```
chrisduerr, line 17: What does this `pkgconfig` annotation mean? The `-devel` seems like the normal way to show dev dependencies.

extra/linux/redhat/alacritty.spec
```diff
@@ -7,14 +7,16 @@ URL:           https://github.com/alacritty/alacritty
...
 Source:        alacritty-%{version}.tar
 
-BuildRequires: rust >= 1.41.0
+BuildRequires: rust >= 1.43.0
 BuildRequires: cargo
 BuildRequires: cmake
-BuildRequires: freetype-devel
-BuildRequires: fontconfig-devel
-BuildRequires: libxcb-devel
+BuildRequires: gcc-c++
```
chrisduerr, line 13: Which crate depedency requires g++?

extra/linux/redhat/alacritty.spec
grumpey: For F32 it looks like it was pulled in by the coreutils package, for F33 that package was updated and ncurses was made optional so it needed to be added to the build requires. 

ref: https://src.fedoraproject.org/rpms/coreutils/c/5d1216301214534571714883927c09b5771fcf7d?branch=master

extra/linux/redhat/alacritty.spec
grumpey: expat-sys

extra/linux/redhat/alacritty.spec
grumpey: The library has provided a pkg-config file (pc) that can is used by rpm to meet the dependency. 
  
Fedora recommends for libraries that use pkg-config that the build dependency be done that way.

ref[1]: https://docs.fedoraproject.org/en-US/packaging-guidelines/PkgConfigBuildRequires/
ref[2]: https://people.freedesktop.org/~dbn/pkg-config-guide.html
ref[3]: https://rusnak.io/why-is-pkg-config-the-best-thing-since-sliced-bread/

chrisduerr: How do you know that these are all using pkgconfig? I'm not sure that `xcb` is...

grumpey: It can be verified with DNF, for libxcb-devel: https://paste.centos.org/view/206ab29

Thanks

chrisduerr: You're saying that these are the pkgconfig things that pull libxcb-devel, right? But I'm saying that I'm not sure if the xcb crate uses pkgconfig. In fact I don't think it is.

grumpey: > You're saying that these are the pkgconfig things that pull libxcb-devel, right? But I'm saying that I'm not sure if the xcb crate uses pkgconfig. In fact I don't think it is.

BuildRequires tells rpmbuild what is needed at build time.
Requires is what's needed at runtime. 

It's not for the crate, it's to meet the build dependencies.

chrisduerr: I'm aware. I'm just saying that the `xcb` crate is not using `pkgconfig` to link against dependencies at build time. The Freetype/Fontconfig stuff is, but xcb is not.

I'm fairly certain that `xcb` can be compiled without `pkgconfig` installed at all, thus specifying it as a dependency that links against `xcb` using `pkgconfig` is false.

grumpey: Switched back to -devel, thanks.

PR #4104, opened 2020-08-10
Remove markdownish markup from `--help` output

kchibisov:


alacritty/src/cli.rs
```diff
@@ -130,7 +130,7 @@ impl Options {
                     .takes_value(true)
                     .use_delimiter(true)
                     .help(&format!(
-                        "Defines window class or `app_id` on Linux [default: {}]",
+                        "Defines window class or app_id on X11/Wayland [default: {}]",
```
chrisduerr, line 133:
```suggestion
                        "Defines window class/app_id on X11/Wayland [default: {}]",
```

This is a bit more uniform and should make it clear that class is for X11 and app_id is for Wayland.

PR #4107, opened 2020-08-11
Add support for double underline

chrisduerr:
This adds support for double underlines using the colon separated escape
sequence `CSI 4 : 2 m`.

Alacritty will now also always fallback to the normal underline in case
any of the other underlines like the undercurl are specified. The escape
sequence `CSI 4 : 0 m` can now be used to clear all underlines.

Some terminals support `CSI 21 m` for double underline, but since
Alacritty already uses that as cancel bold which is a little more
consistent, that behavior has not changed. So the colon separated
variant must be used.


alacritty/src/renderer/rects.rs
```diff
@@ -106,31 +146,36 @@ impl RenderLines {
 
     /// Update the stored lines with the next cell info.
     pub fn update(&mut self, cell: RenderableCell) {
```
kchibisov, line 148: you may want to mark it `#[inline]`, I guess.

chrisduerr:
> The following doesn't really work and create scuffed triple underline. I think that underline should reset double underline(and vise-versa), and that's how it's done in kitty and vte(which makes sense).

That's a great point which I didn't actually consider for a test.

> Also, have you checked how other apps(not terminals) position double underlines, I've tested kitty/vte and they all position them slightly different with the same font size.

No and I honestly don't *really* care. The way I've done it in this PR is the only reasonable one if you ask me, given how tight the space is.

chrisduerr:
Great catch, thanks!

PR #4118, opened 2020-08-13
Fix message bar color being picked from background

### LINKED ISSUE #4116: Message bar text unreadable
Message bar has harcoded yellow or red background but foreground color is from `colors.primary.background`. It is easy to see that it cannot work both with light or dark themes.

I think either message bar’s background color could be get from `colors.normal.{yellow,red}` or use also hardcoded (black) foreground color for messages.

### System

Version: `alacritty 0.5.0`

kchibisov:
This commit reverts the message bar coloring to an old one, so the text
is always colored with #000000 over predefined red and yellow colors.

Fixes #4116.

--

I'd not that I don't really like how it looks, as an alternative solution I'd suggest to use foreground over yellow/red colors, but having predefined colors, like we had before seems also fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Slow rendering performance with HiDPI displays, especially on macOS
 - Keys swallowed during search when pressing them right before releasing backspace
 - Crash when a wrapped line is rotated into the last line
+- Yellow/Red colors for message bar not being picked from user's config
 
 ## 0.5.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Use working directory of active process instead of shell for SpawnNewInstance action
 - Fallback to normal underline for unsupported underline types in `CSI 4 : ? m` escapes
 - The user's background color is now used as the foreground for the render timer
+- Use yellow/red from the config for error and warning messages instead of fixed colors
 
 ### Fixed
 
@@ -33,7 +34,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Slow rendering performance with HiDPI displays, especially on macOS
 - Keys swallowed during search when pressing them right before releasing backspace
 - Crash when a wrapped line is rotated into the last line
-- Yellow/Red colors for message bar not being picked from user's config
 
 ## 0.5.0
 
--- a/alacritty/src/logging.rs
+++ b/alacritty/src/logging.rs
@@ -15,11 +15,9 @@ use std::sync::{Arc, Mutex};
 use glutin::event_loop::EventLoopProxy;
 use log::{self, Level};
 
-use alacritty_terminal::term::color;
-
 use crate::cli::Options;
 use crate::event::Event;
-use crate::message_bar::Message;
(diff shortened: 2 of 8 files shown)


kchibisov:
I've updated commit as well as commit message to reflect the changes. I've decided to use background color for text, and normal.red/normal.yellow for background of the message bar.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Slow rendering performance with HiDPI displays, especially on macOS
 - Keys swallowed during search when pressing them right before releasing backspace
 - Crash when a wrapped line is rotated into the last line
+- Yellow/Red colors for message bar not being picked from user's config
```
chrisduerr, line 35: Maybe this should be in the `Changed` section, since this was never intended to be this way.

alacritty/src/message_bar.rs
```diff
@@ -12,14 +11,24 @@ const TRUNCATED_MESSAGE: &str = "[MESSAGE TRUNCATED]";
 #[derive(Debug, Eq, PartialEq, Clone)]
 pub struct Message {
     text: String,
-    color: Rgb,
+    ty: MessageType,
     target: Option<String>,
 }
 
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+/// Message's contents hint.
```
chrisduerr, line 19:
```suggestion
/// Purpose of the message.
#[derive(Debug, Eq, PartialEq, Clone, Copy)]
```

alacritty/src/logging.rs
```diff
@@ -97,14 +95,14 @@ impl log::Log for Logger {
                         env_var,
                         record.args(),
                     );
-                    let color = match record.level() {
-                        Level::Error => color::RED,
-                        Level::Warn => color::YELLOW,
+                    let message_type = match record.level() {
+                        Level::Error => MessageType::Error,
+                        Level::Warn => MessageType::Warning,
                         _ => unreachable!(),
```
chrisduerr, line 101: You don't have to do it in this PR, but I think we can significantly improve the structure of this method and avoid this `unreachable!` completely.

If we move the logfile writing/stdout/message bar to different methods, we might be able to clean things up a lot easier with match statements and early returns.

I can also send a separate PR to propose some cleanup on this. That might make things a little more clear so the refactoring isn't conflated with the bugfix.

CHANGELOG.md
kchibisov: I guess it could be there.

alacritty/src/logging.rs
kchibisov: yeah, but I'd rather do it separately tbqh.

chrisduerr: Sounds good, yeah.

CHANGELOG.md
```diff
@@ -25,14 +25,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Cursors are now inverted when their fixed color is similar to the cell's background
 - Use working directory of active process instead of shell for SpawnNewInstance action
 - Fallback to normal underline for unsupported underline types in `CSI 4 : ? m` escapes
+- Yellow/Red colors for message bar not being picked from user's config
```
chrisduerr, line 28:
```suggestion
- Use yellow/red from the config for error and warning messages instead of fixed colors
```

We should rephrase this, since it's a change not a fix.

kchibisov:
> We should rephrase this, since it's a change not a fix.

Oh, right. Thx.

PR #4121, opened 2020-08-13
Use user's background color for render timer

kchibisov:
Using user's background color in pair with red color
should provide better contrast.


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Cursors are now inverted when their fixed color is similar to the cell's background
 - Use working directory of active process instead of shell for SpawnNewInstance action
 - Fallback to normal underline for unsupported underline types in `CSI 4 : ? m` escapes
+- The foreground color for debug.render_timer is now matches user's background
```
chrisduerr, line 28:
```suggestion
- The user's background color is now used as the foreground for the render timer
```

There's actually a typo in this commit message `is now matches`. I think rephrasing it a bit is beneficial.

PR #4122, opened 2020-08-13
Close the message bar on release of button press

nixpulvis:
This is a tiny fix for the expected UX of our message bar's button. I was a little startled when the button triggered on press, so I've updated it to trigger on release.

### TODO

- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fallback to normal underline for unsupported underline types in `CSI 4 : ? m` escapes
 - The user's background color is now used as the foreground for the render timer
 - Use yellow/red from the config for error and warning messages instead of fixed colors
+- The message bar close button now activates on release instead of inital press
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


chrisduerr (changes requested):
All this does is make sure the cursor is released above the [X] button, however it does not say anything about the button press location. I'd say that's even more against how things usually work with buttons than the original behavior, so we'd have to fix that.

Our url launcher is blocked when the mouse cursor moves, but that behavior is based on cell motion. So when moving from left side of a cell to the right side, or moving a single cell, the launcher is blocked. However [X] is three cells, or six halves wide. So we can't just use that code.

Calculating the flag differently and updating that whenever the cursor moves is also a rather ugly solution, leading to semi-complicated calculations during cursor motion for things that should be checked during cursor release.

My best suggestion to solve this is to add a last press location, store that, then do all the necessary calculations during release. So basically `down_at_message_close && up_at_message_close`. That should then of course also be used for the url launcher, to unify things properly.

It shouldn't be too difficult, but it requires a good chunk more work than the current single line change.

chrisduerr:
Closing since this PR has been stale for over a year and @nixpulvis has moved on to brighter pastures.

PR #4125, opened 2020-08-13
Add logging for crossfont

chrisduerr:
This refactors the logging function to clean things up a little bit and
also adds the ability to see log messages for the `crossfont` library.

### DIFF
--- a/alacritty/src/logging.rs
+++ b/alacritty/src/logging.rs
@@ -19,7 +19,11 @@ use crate::cli::Options;
 use crate::event::Event;
 use crate::message_bar::{Message, MessageType};
 
+/// Name for the environment variable containing the log file's path.
 const ALACRITTY_LOG_ENV: &str = "ALACRITTY_LOG";
+/// List of targets which will be logged by Alacritty.
+const ALLOWED_TARGETS: [&str; 4] =
+    ["alacritty_terminal", "alacritty_config", "alacritty", "crossfont"];
 
 pub fn initialize(
     options: &Options,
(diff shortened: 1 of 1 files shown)


alacritty/src/logging.rs
```diff
@@ -63,55 +94,32 @@ impl log::Log for Logger {
...
+        // Get target crate.
+        let index = record.target().find(':').unwrap_or_else(|| record.target().len());
+        let target = &record.target()[..index];
+
+        // Only log our own crates.
+        if !self.enabled(record.metadata())
+            || !matches!(target, "alacritty_terminal" | "alacritty" | "crossfont")
+        {
+            return;
+        }
```
kchibisov, line 106: This check in inaccurate. We have `alacritty_config` as a target for config errors, etc, so you should likely add it here as well, since `target` is not a crate name.

Right now it leads to config errors not being displayed in the message bar, you can test with `alt_send_esc: true1`.

Also, I'm not sure whether crossfont errors should reach message bar tbqh. I'm fine with seeing them in a log output of `-vvv`, but I don't want them to populate in message bar, I guess, since I feel like it could spam message a lot, and since we don't have close all functionality it's hard to delete things in it.

So, I'd say `alacritty_config` and `alacritty` should be the only target we could show in a message bar. What do you think?

chrisduerr: I've considered if we should have something like crates which log to the logfile vs crates allowed to log to the message bar. For something like winit it might be useful but since we have control over crossfont I think we should try to make messages warning/errors only if they are critical enough that a user might want to see them.

I've checked out the code of crossfont when I wrote this and the only warning+ it uses is when a font cannot be found on macOS, which would be something that could be helpful for the user for example.

kchibisov: Then it's likely fine, but I feel like we log error in alacritty as well if we fail to load something(you've changed that code recently)?

chrisduerr: I've mentioned this on IRC but just to document this: Hitting the only `warn!` message in `crossfont` does not cause any error propagation. There are no `error!` logs. So there should not be any duplicate messages.

PR #4129, opened 2020-08-14
Update correct location for Windows configuration

yesudeep:
alacritty did not load the configuration for me from `%APPDATA%\alacritty\alacritty.yml` but 
did from `%APPDATA%\Roaming\alacritty\alacritty.yml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ following locations:
 
 On Windows, the config file should be located at:
 
-`%APPDATA%\alacritty\alacritty.yml`
+`%APPDATA%\Roaming\alacritty\alacritty.yml`
 
 ## Contributing
 


kchibisov:
it's not `AppData` it's `%AppData%`, we don't have `Roaming` in a code at all, we use the exact location we're using in a README. `%AppData%` is an environment variable, so it could be expanded to `AppData\Roaming\alacritty\alacritty.yml`.

kchibisov:
The exact path we're using in the code https://github.com/alacritty/alacritty/blob/b630bb45df4e53c01d24251dafa889a7987d8b33/alacritty/src/cli.rs#L17 . As you can see it exactly matches the one we have in a README.

PR #4133, opened 2020-08-15
Add power management config

### LINKED ISSUE #3587: Alacritty Does Not Use GPU In Hybrid Graphics Mode
I'm not sure if I would really describe this as a bug, however when my laptop is running in hybrid graphics mode, alacritty doesn't seem to acquire use of the GPU, and instead will use the CPU for rendering.

Here is the output from `nvidia-smi`
```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 440.64       Driver Version: 440.64       CUDA Version: 10.2     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  GeForce GTX 1650    Off  | 00000000:01:00.0 Off |                  N/A |
| N/A   37C    P8     7W /  N/A |     31MiB /  3911MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0      1726      G   /usr/lib/xorg/Xorg                            14MiB |
|    0      2225      G   /usr/lib/xorg/Xorg                            14MiB |
+-----------------------------------------------------------------------------+
```

Notice that under the processes, it's just Xorg, and the alacrity process isn't listed there

### System

OS: Linux
Version: solidity@gray ~/C/R/Temporal> 
Linux/BSD:Pop!_OS (pretty sure DE)

### Logs

Crashes: not applicable
Font/Terminal size:
```
Created log file at "/tmp/Alacritty-4360.log"
[2020-04-13 16:47] [INFO] Welcome to Alacritty
[2020-04-13 16:47] [INFO] Configuration loaded from "/home/solidity/.alacritty.yml"
[2020-04-13 16:47] [DEBUG] Estimated DPR: 1
[2020-04-13 16:47] [DEBUG] Estimated Cell Size: 9 x 17
[2020-04-13 16:47] [DEBUG] Estimated Dimensions: None
[2020-04-13 16:47] [INFO] Device pixel ratio: 1
[2020-04-13 16:47] [INFO] Initializing glyph cache...
[2020-04-13 16:47] [INFO] ... finished initializing glyph cache in 0.004614572s
[2020-04-13 16:47] [INFO] Cell Size: 9 x 17
[2020-04-13 16:47] [INFO] Padding: 0 x 0
[2020-04-13 16:47] [INFO] Width: 800, Height: 600
[2020-04-13 16:47] [INFO] PTY Dimensions: Line(35) x Column(88)
[2020-04-13 16:47] [INFO] Initialisation complete
[2020-04-13 16:47] [DEBUG] Term::resize dimensions unchanged
[2020-04-13 16:47] [INFO] Width: 800, Height: 600
[2020-04-13 16:47] [DEBUG] [unhandled osc_dispatch]: [['7',],['f','i','l','e',':','/','/','g','r','a','y','/','h','o','m','e','/','s','o','l','i','d','i','t','y','/','C','o','d','e','/','R','T','r','a','d','e','L','t','d','/','T','e','m','p','o','r','a','l',],] at line 758
```
Keyboard and bindings: ive attached as file
[Alacritty-4441.log](https://github.com/alacritty/alacritty/files/4472400/Alacritty-4441.log)

chrisduerr:
The new power management configuration option `window.energy` should
allow users to select if they want to use the dedicated or integrated
GPU.

Until now it was not possible for people to force-enable their dedicated
GPUs when not on power on macOS. This should allow them to make the
choice themselves, allowing for better performance when temporarily
without power.

Fixes #3587.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -81,6 +81,18 @@
   # Set this to `None` to use the default theme variant.
   #gtk_theme_variant: None
 
+  # Power management mode
+  #
+  # This will control things like the GPU used to render Alacritty. When set to
+  # `Auto`, Alacritty will automatically switch to low power mode when on
+  # battery.
+  #
+  # Values for `energy`:
+  #   - Performance
+  #   - Powersave
+  #   - Auto
+  #energy: Auto
+
 #scrolling:
   # Maximum number of lines in the scrollback buffer.
   # Specifying '0' will disable scrolling.
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -36,7 +36,7 @@ use winapi::shared::minwindef::WORD;
 use alacritty_terminal::index::Point;
 use alacritty_terminal::term::SizeInfo;
 
-use crate::config::window::{Decorations, StartupMode, WindowConfig};
+use crate::config::window::{Energy, Decorations, StartupMode, WindowConfig};
 use crate::config::Config;
 use crate::gl;
 
@@ -105,6 +105,7 @@ impl From<crossfont::Error> for Error {
 fn create_gl_window<E>(
     mut window: WindowBuilder,
     event_loop: &EventLoop<E>,
+    window_config: &WindowConfig,
     srgb: bool,
     vsync: bool,
     dimensions: Option<PhysicalSize<u32>>,
@@ -113,10 +114,16 @@ fn create_gl_window<E>(
         window = window.with_inner_size(dimensions);
     }
 
+    let hardware_acceleration = match window_config.energy {
+        Energy::Performance => Some(true),
+        Energy::Powersave => Some(false),
+        Energy::Auto => None,
+    };
+
     let windowed_context = ContextBuilder::new()
         .with_srgb(srgb)
         .with_vsync(vsync)
-        .with_hardware_acceleration(None)
+        .with_hardware_acceleration(hardware_acceleration)
         .build_windowed(window, event_loop)?;
 
     // Make the context current so OpenGL operations can run.
@@ -153,7 +160,7 @@ impl Window {
         #[cfg(not(any(target_os = "macos", windows)))] wayland_event_queue: Option<&EventQueue>,
     ) -> Result<Window> {
         let window_config = &config.ui_config.window;
-        let window_builder = Window::get_platform_window(&window_config.title, &window_config);
+        let window_builder = Window::get_platform_window(&window_config);
 
         // Disable vsync on Wayland.
         #[cfg(not(any(target_os = "macos", windows)))]
@@ -161,9 +168,17 @@ impl Window {
         #[cfg(any(target_os = "macos", windows))]
         let vsync = true;
 
-        let windowed_context =
-            create_gl_window(window_builder.clone(), &event_loop, false, vsync, size)
-                .or_else(|_| create_gl_window(window_builder, &event_loop, true, vsync, size))?;
+        let windowed_context = create_gl_window(
+            window_builder.clone(),
+            &event_loop,
+            &window_config,
+            false,
+            vsync,
+            size,
+        )
+        .or_else(|_| {
+            create_gl_window(window_builder, &event_loop, &window_config, true, vsync, size)
+        })?;
 
         // Text cursor.
         let current_mouse_cursor = CursorIcon::Text;
@@ -243,7 +258,7 @@ impl Window {
     }
 
     #[cfg(not(any(target_os = "macos", windows)))]
-    pub fn get_platform_window(title: &str, window_config: &WindowConfig) -> WindowBuilder {
+    pub fn get_platform_window(window_config: &WindowConfig) -> WindowBuilder {
         let image = image::load_from_memory_with_format(WINDOW_ICON, ImageFormat::Ico)
             .expect("loading icon")
             .to_rgba();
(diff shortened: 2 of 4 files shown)


alacritty/src/window.rs
```diff
@@ -113,10 +114,16 @@ fn create_gl_window<E>(
         window = window.with_inner_size(dimensions);
     }
 
+    let hardware_acceleration = match window_config.energy {
+        Energy::Performance => Some(true),
+        Energy::Powersave => Some(false),
```
kchibisov, line 119: Have you verified that on macOS it'll use integrated GPU even on AC? The same question for `Some(true)`, will it use discrete GPU even on battery?

alacritty.yml
```diff
@@ -81,6 +81,18 @@
...
+  #
+  # This will control things like the GPU used to render Alacritty. When set to
+  # `Auto`, Alacritty will automatically switch to low power mode when on
+  # battery.
+  #
+  # Values for `energy`:
+  #   - Performance
+  #   - Powersave
+  #   - Auto
+  #energy: Auto
```
kchibisov, line 94: I don't really like the name `energy`, since the option is purely about picking GPU, maybe I'd call it `preferred_gpu`, with variants `Integrated`, `Dedicated`, and `Auto`, or `performance_mode` with the variants you have now.

chrisduerr: > I don't really like the name energy, since the option is purely about picking GPU

I'm not a big fan of the name `energy` either, but I don't see why we would make an option exclusive to GPUs when there could be different purposes for it in the future. The goal is not to pick a specific GPU, but to select how power hungry Alacritty is. Thus it seems smarter to represent the goal of the user rather than the means for achieving them. So `preferred_gpu` would be a bad idea imho.

> or performance_mode with the variants you have now.

I've considered something like this, but `performance_mode: performance` doesn't make much sense to me. It's the power management mode Alacritty will be running in, that's the goal. The *performance* mode should always be performance, this is about energy not performance.

alacritty/src/window.rs
chrisduerr: That's a great question. I completely forgot to mention that this PR is not sufficiently reviewed yet when it comes to the feature itself. While #3587 states that `with_hardware_acceleration(Some(true))` enables the dedicated GPU for people, on my system the dedicated GPU is never used for Alacritty according to the system manager.

So I'll have to look into why that is, since other applications like iTerm seem to use it just fine.

However that won't happen until I have my macbook around for a longer time on monday.

alacritty.yml
kchibisov: you may call it powersaving_mode or something like that, but `energy: Performance` doesn't sound right to me.

chrisduerr: I've considered that option too, but I like none of the suggestions tbh. I don't like `energy` either, but `powersaving_mode: Performance` just doesn't make a ton of sense to me. Seems like that would be more of a boolean option.

kchibisov: `hardware_acceleration_mode`, since it's basically what it's doing. Or maybe without `hardware`.

chrisduerr:
I'm going to close this for now since I can't figure out how macOS works and I'm not going to waste my time with their OpenGL crap at this point.

I'm not sure why it was reported that `with_hardware_acceleration` works to force the use of the dedicated GPU, but it doesn't work for me.

kchibisov:
>I'm not sure why it was reported that with_hardware_acceleration works to force the use of the dedicated GPU, but it doesn't work for me.

Because is was on Linux/X11 with nvidia gpu? Like there wasn't anything wrt macOS.

chrisduerr:
Oh I didn't realize that @kchibisov, thanks for reminding me. You don't have a system with switchable GPUs, right?

I think @nixpulvis has a nvidia gpu, but I'm not sure if that's just a desktop without integrated? If not, could you test this?

kchibisov:
>You don't have a system with switchable GPUs, right?

Nah, I don't have such thing.

nixpulvis:
@chrisduerr As a matter of fact, I recently had to get a new motherboard, which meant I ended up needing a new CPU as well. And, since I plan to switch to AMD as soon as I can justify buying a whole new rig, I decided I'd get a CPU which wouldn't need a dedicated GPU so it could serve in it's afterlife as a alternate computer, or sever for me.

Long story short. I should be able to test Linux/Windows on a machine which has both integrated and dedicated GPUs. However, I'm still unboxing everything, and may need to figure out the configurations and feedback for this, since I've never played with it before.

chrisduerr:
Running a desktop with a GPU and iGPU might be a bit different than a laptop with automatic graphics switching unfortunately, so I'm not sure that'll be helpful.

jorgebef:
> Running a desktop with a GPU and iGPU might be a bit different than a laptop with automatic graphics switching unfortunately, so I'm not sure that'll be helpful.

I posted on another related issue that is open, I happen to have a Macbook that has switchable GPUs and would love to help out with this.
If there is anything I can help with please let me know.

chrisduerr:
> I posted on another related issue that is open, I happen to have a Macbook that has switchable GPUs and would love to help out with this.

Switchable graphics on macOS doesn't seem to work. So you'd have to run Linux to test this. I haven't seen any application that manages to properly switch on macOS unfortunately. You just have to set it in the Info.plist.

chrisduerr:
Closing since nobody was able to get any benefit from this.

PR #4143, opened 2020-08-18
Add a cursor inspired by the Plan 9 terminal

laumann:
This adds another glyph that is is similar to the beam cursor, except
both ends are slightly thicker. Like an I-beam, but the "beams" are
square.

I haven't updated any documentation or versions, because I'm not sure this is something that anyone wants, I just thought it could be fun to add this as an option.

How it looks with my configuration:

image

with the default configuration:

image

Screenshot from Plan 9 for reference:

image

TODO:
 - [ ] Update `CHANGELOG.md`
 - [ ] Update version of `alacritty_terminal` (?)


laumann:
> Thank you for the suggestion, but I have no interest in adding other cursor shapes. The Beam/Block/Underline cursors are unified cursor shapes supported by escape sequences, each with its own purpose. The goal isn't making things interesting visually.

I understand, thanks for looking at it :-)

PR #4152, opened 2020-08-20
added ResetAndClear action

### LINKED ISSUE #3997: Sending ANSI Escape Code "clear" does not work
I am trying to configure a new keybinding (`Ctrl+k`) to send a clear sequence to alacritty, I have tried:

```
  - { key: K,          mods: Control, chars: "\e[H\e[2J",     mode: ~Alt       }
or
  - { key: K,          mods: Control, chars: "\x1b[2J",       mode: ~Alt       }
```

and a few other variants, but none of them seem to work.

However I simply: `echo '\e[H\e[2J'`, then the terminal is cleared. 

What is even weirder is that the chars seem to be sent to the terminal:

```
> showkey -a

Press any keys - Ctrl-D will terminate this program

^[[H^[[2J 	 27 0033 0x1b
 	 91 0133 0x5b
 	 72 0110 0x48
 	 27 0033 0x1b
 	 91 0133 0x5b
 	 50 0062 0x32
 	 74 0112 0x4a
```

### System

OS: Linux
Version: alacritty 0.4.3
Linux/BSD: Wayland with Sway sway version v1.5-rc2-1d149230 (Jul  5 2020, branch 'master')

### Logs

No Crash
See the `showkey` logs

I have looked at #1047 but none of the suggestions have worked for me (running Arch Linux).

I could use the default `ClearLogNotice` binding (`Ctrl+l`) but it has limitations. For example it doesn't work when piping `tree / | more`, or when logging: `journalctl -b -f`.

My research before posting this:

- [Alacritty Docs](https://github.com/alacritty/alacritty/tree/master/docs)
- [Stack Overflow](https://stackoverflow.com/questions/37774983/clearing-the-screen-by-printing-a-character)
- [vt100](https://invisible-island.net/ncurses/terminfo.src.html#tic-vt100)

markwellis:
Resets the terminal, same as
- "Clear Scrollback and Reset" in xcfe4-terminal
- "Clear Scrollback and Reset" in konsole
- "Reset and Clear" in gnome-terminal

Fixes #3997

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support for colon separated SGR 38/48
 - New Ctrl+C binding to cancel search and leave vi mode
 - Escapes for double underlines (`CSI 4 : 2 m`) and underline reset (`CSI 4 : 0 m`)
+- ResetAndClear action which clears scrollback and resets
 
 ### Changed
 
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -145,6 +145,9 @@ pub enum Action {
     /// Clear the display buffer(s) to remove history.
     ClearHistory,
 
+    /// Reset and clear the display
+    ResetAndClear,
+
     /// Hide the Alacritty window.
     Hide,
 
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -311,6 +311,11 @@ impl<T: EventListener> Execute<T> for Action {
                 term.vi_motion(ViMotion::FirstOccupied);
             },
             Action::ClearHistory => ctx.terminal_mut().clear_screen(ClearMode::Saved),
+            Action::ResetAndClear => {
+                ctx.terminal_mut().clear_screen(ClearMode::All);
+                ctx.terminal_mut().reset_state();
+                ctx.terminal_mut().dirty = true;
+            },
             Action::ClearLogNotice => ctx.pop_message(),
             Action::SpawnNewInstance => ctx.spawn_new_instance(),
             Action::ReceiveChar | Action::None => (),


chrisduerr (changes requested):
Closing for the same reasons that were already stated in #3997.

markwellis:
> Closing for the same reasons that were already stated in #3997.

The reasons in #3997 are that it might leave applications in a strange state. In my experience this is not the case except if done in something like vim. And if the terminal gets in a weird state running it again fixes it.

alacritty

This is something most other terminals do and it really is very useful to me, e.g. when tailing logs I can clear mid way through to remove the parts I'm not interested in, then I can let it scroll and jump back to the top later and not have to re-find the interesting stuff.

Please could you reconsider, this is a very useful feature to me

markwellis:
Hopefully this will show my usecase better than I'm explaining:
alacritty_tail

I tail a log, then clear the screen, read the stuff i want, then clear it again.

If I can't clear the screen the best I can do is press return a bunch of times and look for the gaps, but it's not as nice as just removing the stuff I'm not interested in.

Thanks

jtheoof:
@chrisduerr  I do not understand the reluctance to not have this in Alacritty. Yes it can break the state of the app, but that's the purpose of it, especially in the use cases listed here and in #3997. Letting the users decide through proper use of this action seems pretty reasonable to me.

kchibisov:
The point is that once you add a hacky feature, you'll deal with reports from users like 'my editor got borken after ResetAndClear'. Having a feature that doesn't really work doesn't make any sense. If you want something like that for `tail`, maybe add a binding to tail to perform log clear? You can even write escape sequence to a PTY where `tail` is running, which will result in exactly the same thing, and even add a binding to write to a pty a clear escape(yes, it's possible with a bunch of hacks). Adding hacks on purpose to alacritty doesn't make any sense to me.

jtheoof:
In my opinion this more of a feature than a hack if properly documented as to what this action does.

kchibisov:
I mean, what the point of adding it if you can already hack it around? I don't really like an idea of having such thing at all, knowing that it can't work as intended.

graemel:
Why not make it a config option to enable it, with it defaulted to off? That way people can decide if they want it enabled.

kchibisov:
That just creates even more of a mess.

kchibisov:
If someone don't understand about what I'm talking about in a 'you can do it right now in a bit hacky way', here's a script, which you can bind on any keybinding.

```sh
#!/bin/sh

if [ "${ALACRITTY_LOG+set}" = set ]; then
    alacritty_log_file="$(echo $ALACRITTY_LOG | grep -o '[0-9]*')"
    export ALACRITTY_PID="$alacritty_log_file"
fi

tty=$(ps o tty --ppid $ALACRITTY_PID  | grep p)

echo -en "\e[2J\a\e[3J\a\e[1H\a" > /dev/$tty

```

I'm not encouraging to use such thing, I'm just saying that there's a way to add whatever hacks you want, due to how communication over pty interfaces.

markwellis:
I get it's your project and you're to merge or not merge whatever code you like. Makes sense, it's your work I have no problem with that.

However, calling a feature of pretty much every terminal I've used a hack seems weird. Using it with tail was just an example, I do this with many other programs, compiling, running things in debug mode, general logging from programs.

As for saying it doesn't work with a binding, it works exactly as I expect it to, exactly as it does in every other terminal. It clears the terminal in the middle of a running app, that's exactly what I want to happen.

To stop people reporting it causing problems, would it be acceptable to call it something like `HardRestetTerminal` or something clearer?

The script you posted above, how can I run that whilst something else is still running? That's the aim here, clear the terminal whilst something is still running without pausing it or exiting.

chrisduerr:
> In my opinion this more of a feature than a hack if properly documented as to what this action does.

That would only be the case if it wouldn't be utterly destructive to the terminal state, making it possible for all kinds of stuff to just break.

The only purpose for such an action is to reset the terminal state when reaching a point of no return and having a seemingly corrupted terminal state. Though I'd expect Alacritty users to be familiar with the `reset` keyword and soft reset, which basically allows you to recover without a special binding. Fancy GUI terminal emulators just bind it to some nice action because people might not be aware of how else to do it. It doesn't make sense as a terminal feature, it's a recovery option.

markwellis:
problem is I can't run reset whilst something is running. That's what I want, to clear the terminal whilst something is running. It's not for recovery, it's for clearing the screen in the middle of something running.

kchibisov:
>problem is I can't run reset whilst something is running. That's what I want, to clear the terminal whilst something is running. It's not for recovery, it's for clearing the screen in the middle of something running.

Have you actually tried binding it? it'll work if something is running without any issue.

chrisduerr:
> problem is I can't run reset whilst something is running.

Because you're not supposed to. Resetting the terminal state while something is running perfectly fine makes no sense.

That said, as @kchibisov has stated already it will obviously work with escape sequences like the soft reset.

markwellis:
Yes that's what the above gif is, me using it whilst tail is running. I've been using it all day with tail, vagrant output and ssh. Obviously if I use it in vim all hell breaks loose, but I don't do that.

image

To be honest, I'd never even considered it's there to reset the terminal when it's gets corrupted, I always just type reset for that. To me, it's just a way to clear the screen whilst something is running.

Maybe it's a weird way to work, I'd never thought about it before tbh. It's just something that's there and does what I wanted. Before that I used to press return a lot and look for the gaps, which sucks.

markwellis:
Is there a way I can bind escape sequences to a key? I couldn't get it to work with `chars`

kchibisov:
Bind my script to a hotkey.

chrisduerr:
> To be honest, I'd never even considered it's there to reset the terminal when it's gets corrupted, I always just type reset for that. To me, it's just a way to clear the screen whilst something is running.

You're saying that we should just document what it does, but you're not sure yourself what it is used for and using it for something different than it is intended for. That should already show that implementing this will only lead to trouble and abuse.

markwellis:
> > To be honest, I'd never even considered it's there to reset the terminal when it's gets corrupted, I always just type reset for that. To me, it's just a way to clear the screen whilst something is running.
> 
> You're saying that we should just document what it does, but you're not sure yourself what it is used for and using it for something different than it is intended for. That should already show that implementing this will only lead to trouble and abuse.

But how do you know what it's intended use is if you'd never heard of it before #3997 ? Who's to say what's right and wrong; what's abuse and what's not?

Anyway as I said, it's your project you don't have to accept the patch, no problem. But it is a feature of every other terminal, and it's useful to me for how I use it. I'm sure there are others.

> Bind my script to a hotkey, and no, you can't trigger escape processing if you write youself with `chars`, someone else should do that.

What do you mean hotkey? something external to alacritty? I don't understand sorry.

kchibisov:
>What do you mean hotkey? something external to alacritty? I don't understand sorry.

I mean that you can define a binding in alacritty.yml to run an arbitrary command. For example.

- { key: K,  mods: Alt,  command: path_to_script }

markwellis:
> > What do you mean hotkey? something external to alacritty? I don't understand sorry.
> 
> I mean that you can define a binding in alacritty.yml to run an arbitrary command. For example.
> 
> ```
> - { key: K,  mods: Alt,  command: path_to_script }
> ```

ok this I can work with.

Would you be open to a PR to support to send escape codes via key bindings?

chrisduerr:
No.

mitar:
Hm. I think there is a bit of confusion here. I think in most cases people just want "clear", not "reset and clear". If you are piping logs you need just "clear". And alacritty already supports that. On Mac command-K standard shortcut does that.

Reset is when things get into a strange state. For that I opened: https://github.com/alacritty/alacritty/issues/4199

nixpulvis:
@mitar I believe the request is for a clear which clears both the grid, and the scrollback.

> e.g. when tailing logs I can clear mid way through to remove the parts I'm not interested in, then I can let it scroll and jump back to the top later and not have to re-find the interesting stuff.

This means that the current `ClearHistory` action is probably not what people want with this issue, however I do think the title/description makes this confusing.

It's the position of the Alacritty developers that we would like to avoid breaking applications as much as possible, which means allowing arbitrary clears from keybinding isn't allowed. I personally might find this overly restrictive, since it's not a default, and the user provided an input to trigger it, thus knowing it's being done, but I can also see the other point of view.

mitar:
> I believe the request is for a clear which clears both the grid, and the scrollback.

Sure, but you can bind two actions to one keyboard shortcut. So the only missing piece is then `ResetTerminal` action. Because clearing the scrollback already has an action. We do not need a special action for doing both.

alexdavid:
> Sure, but you can bind two actions to one keyboard shortcut.

Right, but there is no `ClearGrid` action, there is only the `ClearHistory` action which clears scrollback.

I understand where the devs are coming from not wanting to break applications, and that it's not the responsibility of the terminal emulator to allow binding a clear grid shortcut, and instead should be the responsability of a pager. But I could make the same argument about having scrollback, why not defer that responsibility to a pager as well?

I do want to say that I prefer to use projects where the devs aggressively turn down features as it means the project can stay lean, and focus on stability (and as @kchibisov mentioned there are even workarounds). I just found it a little strange that it was missing since every other terminal emulator I've run has had a clear grid shortcut in some form or another.

Even kitty, who's developer [prefers to build a terminal emulator that does the right thing even if it goes against what most other terminal emulators are doing](https://github.com/kovidgoyal/kitty/issues/197#issuecomment-348062665) has a [default shortcut to reset the terminal](https://sw.kovidgoyal.net/kitty/#id12).

chrisduerr:
> But I could make the same argument about having scrollback, why not defer that responsibility to a pager as well?

That's exactly how it works actually. Alacritty doesn't track the scrollback of pagers.

alexdavid:
> That's exactly how it works actually. Alacritty doesn't track the scrollback of pagers.

My point is that if your stance is "if you want to be able to clear in the middle of a long runnign program output, you should run it through a pager that supports \x0c", I could make the same argument for removing scrollback entirely with "if you want to be able to see more than a single page of output from a program you should run it through a page that supports scrollback"

Again, not saying that this should be implemented, just throwing in my 2 cents.

PR #4169, opened 2020-08-25
Update example for -o flag to be more apparent.

nixpulvis:
Just a small tweak to the `--help`/man page example for the newly added `--option` flag to help make the behavior more obvious to a user who simply copies and pastes the example.

P.S. I'm wondering if it would be a good idea to write our man page or CLI help text once, and generate the other. This might now be too hard with something like [rust-cli/man](https://github.com/rust-cli/man). Thoughts?

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -182,7 +182,7 @@ impl Options {
                     .short("o")
                     .multiple(true)
                     .takes_value(true)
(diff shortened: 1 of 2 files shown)


chrisduerr (changes requested):
Could you rebase this?

chrisduerr (approved):

PR #4171, opened 2020-08-26
Add openSUSE installation instructions

hennevogel:


chrisduerr (changes requested):
We don't include official installation instructions, since people should be able to find official packages for their distros themselves without any issue.

PR #4190, opened 2020-08-31
Fix compilation for windows-gnu with default features 

mati865:
WinPTY crate supports only windows-msvc right now so leaving it enabled for windows-gnu results in build errors unless `--no-default-features` is added.

### DIFF
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -36,7 +36,7 @@ winapi = { version = "0.3.7", features = [
 ]}
 mio-anonymous-pipes = "0.1"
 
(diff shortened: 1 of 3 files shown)


alacritty_terminal/Cargo.toml
```diff
@@ -37,6 +36,10 @@ winapi = { version = "0.3.7", features = [
 ]}
 mio-anonymous-pipes = "0.1"
 
+# Winpty crate supports only MSVC
```
chrisduerr, line 39:
```suggestion
# Winpty crate supports only MSVC.
```

chrisduerr (approved):

Issue #4193, opened 2020-09-01
Feature: A way to replace Logo and Ctrl keys for all input on Linux


mitar:
I am trying to configure my Ubuntu 20.04 Linux Gnome shell to behave similar to how Mac behaves about keyboard shortcuts: super key should be used for all UI keyboard shortcuts (like copying to the clipboard), while ctrl should be used for the terminal emulators, for example for ctrl-c. [See this post for more information about my setup](https://unix.stackexchange.com/questions/607113/how-to-configure-the-keyboard-to-behave-like-macs). The issue I am having is that for things to work this way, I change modifier keys system-wide and would need for terminal emulators to have different modifier keys, in particular, logo and ctrl keys should be swapped. Looking around it seems this is not really possible, unless the terminal emulator itself does the swapping. I have managed to patch Gnome Terminal to behave like this by simplify swapping modifiers on the input, but for Alacritty I was unable to find a way to do so, so I am opening an issue to see if a) this is something somebody else would know how to do b) this is something others would also find useful and would be something we add to the Alacritty itself.

So what is needed is to swap Logo and Ctrl keys around for all inputs. This is easy for key presses as I can just swap them around in `modifiers_input`:

```rust
        let mut modifiers = modifiers.clone();
        if modifiers.ctrl() != modifiers.logo() {
            modifiers.toggle(ModifiersState::LOGO);
            modifiers.toggle(ModifiersState::CTRL);
        }
        *self.ctx.modifiers() = modifiers;
```

But the problem is in `received_char` because winit calls [`wt.xconn.lookup_utf8`](https://docs.rs/winit/0.22.2/src/winit/platform_impl/linux/x11/event_processor.rs.html#597) to obtain the character pressed, which in turn [modifies it to control codes if `ctrl` was pressed during that time](https://github.com/mirror/libX11/blob/fb65b67169105c82c243f265e0f993cf72efdacd/src/KeyBind.c#L867-L874). But that is the wrong `ctrl` and not swapped. So I tried to reverse this change and obtain the original character, but there is not enough information available to do that well, to my understanding. So I am stuck here.

chrisduerr:
There is https://github.com/alacritty/alacritty/issues/62 which will solve this for macOS, since the modifiers on macOS are completely backwards and broken. However I don't expect it to make much of a difference on Linux.

[...]

PR #4195, opened 2020-09-01
Make logo key convert characters to control characters, too

mitar:
So I figured out this is the only thing which I need to allow me to use ctrl key everywhere around my system for the UI keyboard shortcut key, so `ctrl-c` then copies to the clipboard in all apps, including alacritty. I then configure keyboard shortcut in alacritty the same as I have them system wide. And then I can use the super key to send control characters to the terminal. With this configuration option enabled, both ctrl and super keys send control characters to the terminal, but you can override them with alacritty shortcuts which then suppress that particular combination.

I in fact configure system-wide modifier keys to be in a different order (super, fn, alt, ctrl) so that the key closest to the space key behaves like Mac's command key. But this is unrelated to alacritty. For some more background on this see: https://github.com/alacritty/alacritty/issues/4193

### DIFF
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -868,6 +868,30 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
         self.ctx.window_mut().set_mouse_cursor(mouse_state.into());
     }
 
+    // Same logic to map to control characters as found in
+    // libX11's _XTranslateKeySym function.
+    fn set_control(&mut self, bytes: &mut [u8]) {
+        if bytes.len() != 1 {
+            return;
+        }
+
+        let mut c = bytes[0];
+
+        if (c >= '@' as u8 && c < 0o177) || c == ' ' as u8 {
+            c &= 0x1F
+        } else if c == '2' as u8 {
+            c = 0o000
+        } else if c >= '3' as u8 && c <= '7' as u8 {
+            c -= '3' as u8 - 0o033
+        } else if c == '8' as u8 {
+            c = 0o177
+        } else if c == '/' as u8 {
+            c = '_' as u8 & 0x1F
+        };
+
+        bytes[0] = c;
+    }
+
     /// Process a received character.
     pub fn received_char(&mut self, c: char) {
         let suppress_chars = *self.ctx.suppress_chars();
@@ -902,6 +926,12 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
             c.encode_utf8(&mut bytes[..]);
         }
 
+        if self.ctx.config().ui_config.logo_as_ctrl()
+            && self.ctx.modifiers().logo()
+            && !self.ctx.modifiers().ctrl()
+        {
+            self.set_control(bytes.as_mut_slice());
+        }
         if self.ctx.config().ui_config.alt_send_esc()
             && *self.ctx.received_count() == 0
             && self.ctx.modifiers().alt()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Configuration file option for sourcing other files (`import`)
 - CLI parameter `--option`/`-o` to override any configuration field
 - Escape sequences to report text area size in pixels (`CSI 14 t`) and in characters (`CSI 18 t`)
+- New `logo_as_ctrl` option for controlling if logo key should convert written
+  characters to control characters, like the ctrl key does. On X11 only.
 
 ### Changed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -390,6 +390,10 @@
 # Send ESC (\x1b) before characters when alt is pressed.
 #alt_send_esc: true
 
+# Make logo key convert characters written to the terminal to
+# control characters, like the ctrl key does. On X11 only.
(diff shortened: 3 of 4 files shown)


alacritty/src/input.rs
```diff
@@ -868,6 +868,30 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
         self.ctx.window_mut().set_mouse_cursor(mouse_state.into());
     }
 
+    // Same logic to map to control characters as found in
+    // libX11's _XTranslateKeySym function.
```
nixpulvis, line 872: I'm still trying to understand all this, but in the process found this useful message on the xorg mailing list: https://www.mail-archive.com/xorg@lists.x.org/msg04434.html

nixpulvis:
Let me see if I'm understanding this correctly... You have the following global remapping (from left to right, on the left side of the space bar): <kbd>Ctrl</kbd>, <kbd>Alt</kbd>, <kbd>Fn</kbd>, <kbd>Super</kbd> (aka logo). You want to have the right-most key + <kbd>C</kbd> perform a copy. This seems possibly today, just bind <kbd>Super</kbd> + <kbd>C</kbd> in `alacritty.yml`. If my understanding is correct, the issue lies in all the other applications who bind <kbd>Ctrl</kbd>+<kbd>C</kbd> to copy, which I am somewhat understanding is a pain to remap.

Is this correct?

chrisduerr (changes requested):
As stated in the issue, I have no desire to support this in Alacritty.

mitar:
> I mean, we currently do all those 

Sorry, which issue? #4193? In #4193 I was asking for something a bit different. This now is much simpler.

alacritty/src/input.rs
mitar: I mean, we currently do all those changes internally inside libX11, so we should probably keep things the same, just to not have a difference between both keys.

mitar:
> If my understanding is correct, the issue lies in all the other applications who bind Ctrl+C to copy, which I am somewhat understanding is a pain to remap.

Yes, so it is easier to use Ctrl-C in all other apps, and change the interrupt to Super-C in Alacritty. So this is what this PR provides. A way for you to send control codes using Super key so that you can keep using Ctrl key for system-wide UI shortcuts. So in reality my keys are: super, alt, fn, ctrl in Linux, but on the keyboard they look like ctrl, alt, fn, and super. Super confusing. But super intuitive at the end and super powerful.

chrisduerr:
#4193, correct.

mitar:
I do not understand your comment there:

> It sounds to me like you've just put yourself in a bad spot by trying to make something happen that isn't supported. So you really want to fight that uphill battle, you're gonna have to do it on your own.

I mean, this PR sidesteps that issue and makes it an easy win. I thought the whole idea of open source is to work together to make things which are not yet supported, supported. I managed to do with a pretty small change. So not sure what "uphill battle" you are talking about here. This literally addresses the issue I was having and make opens one to have same shortcuts in alacritty on both Mac and Linux. I think this is a great feature. Especially if you have to work on both machines.

kchibisov:
>Yes, so it is easier to use Ctrl-C in all other apps

If the goal is to use `ctrl + c` in terminal, just map `Copy` to it, and send interrupt and other things via `ctrl + shift + c`, since `ctrl + c` and `ctrl + shift + c` are the same when it comes to chars those are sending. You can do the exact same thing for `ctrl + v`.

```
- { key: C,       mods: Control, action: Copy                                }
- { key: C,       mods: Control|Shift, action: ReceiveChar                   }
- { key: V,       mods: Control, mode: ~Vi, action: Paste                    }
- { key: V,       mods: Control|Shift, action: ReceiveChar                   }
```

mitar:
The goal is to have the same muscle memory between Mac and Linux. On Mac I do not have to press extra shift to send interrupt. It is also just faster to press just one key plus c.

I understand that other people have other preferences here. This is fine. But I think that for people who are used to Mac, or use both Mac and Linux, it would be great if you could configure things to be the same. With this patch this is possible.

nixpulvis:
> I managed to do with a pretty small change.

I would argue that a whole new configuration option, with a mapping that frankly I'd be surprised if anyone truly understands behind it, is very much **not** a pretty small change. Again, having myself once converted from macOS as my default machine to a Linux setup, I'm somewhat sympathetic to wanting the same exact bindings, but I think it's pretty clear to me that the burden is on other applications.

> I thought the whole idea of open source is to work together to make things which are not yet supported, supported.

Open source is about empowerment, not acceptance. There is nothing stopping you from, creating a fork, and following upstream for changes. This is what many of us do (with or without divergence from `upstream/master`), and as long as your patch remains conflict free, then this should be trivial to maintain. We can't promise the fork will be trivial forever, but as you said, it's a small patch (in some respects).

All that being said... I do wish to bring up the ability to send escapes through keybindings. This may allow setting a custom `stty intr <char>` option, which you trigger with a mapped keybinding. It's still a bit of a hack, but would possibly allow avoiding a fork, since this is a feature that makes sense (to me anyway).

chrisduerr:
> All that being said... @chrisduerr I do wish to bring up the ability to send escapes through keybindings. This may allow setting a custom stty intr <char> option, which you trigger with a mapped keybinding. It's still a bit of a hack, but would possibly allow avoiding a fork, since this is a feature that makes sense (to me anyway).

There's no good way to do this, but that's a different issue. Sending "escapes" through keybindings is just a recipe for disaster.

Issue #1180, opened 2018-03-12
tmux background colors cause windows to become completely opaque


fallwith:
macOS 10.13.3 with tmux 2.6

I have Alacritty's opacity level set to 80%:

```
background_opacity: 0.8
```

The resulting Alacritty window is partially opaque as desired. Then I launch tmux and the window becomes completely opaque.

I have discovered that if I stop specifying background colors in tmux's configuration, the Alacritty window remains correctly at the desired 80% opacity.

```
# original tmux configuration
set -g window-style 'fg=colour250,bg=colour233'
set -g window-active-style 'fg=colour255,bg=black'

# workaround for Alacritty opacity issue
set -g window-style 'fg=colour250'
set -g window-active-style 'fg=colour255'
```

My expectation is that a background color choice from a terminal application should not have an impact on that terminal's window opacity. If there is anything I can do currently in either Alacritty or tmux to leverage both background colors and transparency, please let me know.

jwilm:
Does this not show up in other terminals? From what I've looked at with eg urxvt, background is always filled in at 100% opacity *unless* it's the default background color.

chrisduerr:
I guess it's a question of how you want to look at it. Our global `opacity` setting should probably always take place and just make every background transparent.

Then alternatively the background color of the terminal could be set to something like `0xff00ff22` to have transparency just  for the default background color.

I can see why you'd expect everything to be transparent when setting a global opacity to a certain value.

Allowing RGBA notation is probably a bit too much, but I can see where the confusion is coming from.

maximbaz:
For reference: https://github.com/jwilm/alacritty/pull/847

fallwith:
@jwilm the other terminals I've tested recently under macOS are Terminal and iTerm 2 and both of those retain transparency despite the presence of a background color. It's a fair point about the current behavior possibly matching the expected behavior of most terminals (Linux ones especially). I'd be ok with the default behavior remaining as is if that's the case.

fallwith:
@chrisduerr You've summarized the situation perfectly and greatly helped with my understanding of why things are the way they are now. Thank you. Since opening this issue I've realized that it is not tmux specific (can reproduce with a Vim background color as well outside of tmux). Your comment explaining the background color behavior is quite helpful.

sodiumjoe:
I think this is a dupe of #1082

chrisduerr:
Closing as a duplicate of #1082. Thanks for the heads-up joe.

[...]

Issue #1082, opened 2018-02-02
background transparency goes away when I enter vim


llimllib:
Which operating system does the issue occur on? Mac OS X 10.13.3, alacritty `2466f81`

To reproduce:

1. Set up base16 colors in your terminal: https://github.com/chriskempson/base16-shell
2. Load up a base16 color scheme in vim: https://github.com/chriskempson/base16-vim/blob/master/colors/base16-twilight.vim is the particular one I use
3. notice that your vim background is no longer transparent. If you quit and return to the shell, the transparency returns
  a. in iterm2, with the same vimrc and bashrc, vim remains transparent

jwilm:
What happens if you change

```viml
let s:gui00        = "1e1e1e"
let g:base16_gui00 = "1e1e1e"
```

to be

```viml
let s:gui00        = ""
let g:base16_gui00 = ""
```

llimllib:
No change, unfortunately

severen:
I'm also running into this issue with the Ayu theme on Neovim. Using `03f9e0c`.

maximbaz:
I believe this was caused by https://github.com/jwilm/alacritty/pull/847, however that was a good fix for powerline icons. 

Also, adding `hi! Normal ctermbg=NONE guibg=NONE` after `colorscheme ...` seems to fix the issue for me.

laanwj:
Would be somewhat nice if it was possible to make other background colors transparent as well, say, a green background becomes a transparent green background. I don't know of any terminal that does this, though.

kchibisov:
Your vim colorscheme could set background color to be a specific value(e.g. truecolor or indexed color) in which case alacritty **should** draw it without opacity. To make opacity work in vim, you should use set background in vim to `None`.

Not sure how it helps but something like this could work for arbitrary themes.
```
hi! Normal ctermbg=NONE guibg=NONE
hi! NonText ctermbg=NONE guibg=NONE guifg=NONE ctermfg=NONE
```

bjce:
I have neovim `nvim 0.4.3` and `alacritty 0.7.2 (5ac8060)`

I tried (sequentially) the following lines after `colorscheme` in the `.vimrc` without success: 
```
hi Normal guibg=NONE ctermbg=NONE
hi Normal ctermbg=NONE guibg=NONE
hi! NonText ctermbg=NONE guibg=NONE guifg=NONE ctermfg=NONE
```

[...]

PR #847, opened 2017-10-17
Fix solid background color opacity

jwilm:
Since landing the patch adding transparency support to Alacritty,
there's been an issue where othewise solid background cells were also
being rendered partially transparent. Now, all filled background cells
are rendered fully opaque.

Some logic was added to support discarding filled backgrounds which had
the same color as the default background. This means that, if the
default background is #000 and a cell has that background, it will never
be rendered opaque. This may not be correct.

Note that many truecolor vim color schemes print spaces for default
colored background cells. Performance can be dramatically improved by
using ctermbg=NONE guibg=NONE to skip rendering those cells.

Issue #807, opened 2017-10-02
Transparency affects all background colors


haarp:
Linux, X11

Hey! When transparency is enabled, it affects every background color. Other terminal emulators have it apply only to the "background" of the terminal, but have custom background colors appear at 100% still.

See screenshots. How it should be (transparency disabled entirely):
![screenshot_2017-10-02_21-46-23](https://user-images.githubusercontent.com/781030/31096217-b32d6fe0-a7bb-11e7-845d-f2f3d9c9726c.png)

What alacritty does with transparency enabled:
![screenshot_2017-10-02_21-49-55](https://user-images.githubusercontent.com/781030/31096234-c0c82352-a7bb-11e7-90d1-88ff04dcbcc0.png)

Javyre:
I'm pretty sure this is now fixed in #847 (merged yesterday)

haarp:
Yes, indeed. Thanks!

PR #4196, opened 2020-09-01
Allow to apply opacity to all background colors

### LINKED ISSUE #741: Transparency improvements
Based on discussion in #331, we would like to do one or both of the following:

- Add an option to make non-default background colors opaque
- Add an option to fonts to customize their opacity.

satlank:
Alacritty currently only applies the value of `background_opacity` to
the default background and not to all (e.g. #807, addressed in #847).
However this is not necessarily always desired (e.g. #1082, #1180).

This patch adds a configuration option and logic (by partly re-instating
code removed in #847) to toggle the behavior between either applying
only to the default background color, or all background colors.  For
example iterm2 also has this option and the naming of the toggle is
borrowed from there.

Fixes #741.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -93,6 +93,12 @@
   # and `light`. Set this to `None` to use the default theme variant.
   #gtk_theme_variant: None
 
+  # Background opacity
+  #
+  # Window opacity as a floating point number from `0.0` to `1.0`.
+  # The value `0.0` is completely transparent and `1.0` is opaque.
+  #background_opacity: 1.0
+
 #scrolling:
   # Maximum number of lines in the scrollback buffer.
   # Specifying '0' will disable scrolling.
@@ -284,6 +290,13 @@
   #
   #indexed_colors: []
 
+  # Opaque backgrounds with transparency enabled
+  #
+  # Whether or not `background_opacity` applies to all backgrounds or only the
+  # default background. When set to `false` all cells will be transparent
+  # regardless of their background color.
+  #opaque_background_colors: true
+
 # Bell
 #
 # The bell is rung every time the BEL control character is received.
@@ -324,12 +337,6 @@
   #     args: ["Hello, World!"]
   #
   #command: None
-
-# Background opacity
-#
-# Window opacity as a floating point number from `0.0` to `1.0`.
-# The value `0.0` is completely transparent and `1.0` is opaque.
-#background_opacity: 1.0
 
 #selection:
   # This string contains all characters that are used as separators for
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -292,7 +292,7 @@
 
   # Opaque backgrounds with transparency enabled
   #
-  # Whether or not `background_opacity` applies to all backgrounds or only the
+  # Whether or not `background_opacity` applies to all backgrounds or only to the
   # default background. When set to `false` all cells will be transparent
   # regardless of their background color.
   #opaque_background_colors: true
--- a/alacritty/res/text.f.glsl
+++ b/alacritty/res/text.f.glsl
@@ -18,7 +18,7 @@ void main() {
         }
 
         alphaMask = vec4(1.0);
-        color = vec4(bg.rgb, 1.0);
+        color = vec4(bg.rgb, 1.0) * bg.a;
     } else if ((int(fg.a) & COLORED) != 0) {
         // Color glyphs, like emojis.
         vec4 glyphColor = texture(mask, TexCoords);
--- a/alacritty/res/text.f.glsl
+++ b/alacritty/res/text.f.glsl
@@ -18,7 +18,7 @@ void main() {
         }
 
         alphaMask = vec4(1.0);
-        color = vec4(bg.rgb, 1.0) * bg.a;
+        color = vec4(bg.rgb, bg.a);
     } else if ((int(fg.a) & COLORED) != 0) {
         // Color glyphs, like emojis.
         vec4 glyphColor = texture(mask, TexCoords);
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -65,10 +65,7 @@ impl Default for UIConfig {
 impl UIConfig {
     #[inline]
     pub fn background_opacity(&self) -> f32 {
-        match self.background_opacity {
-            Some(x) => x.as_f32(),
-            None => self.window.background_opacity(),
-        }
+        self.background_opacity.map(Percentage::as_f32).unwrap_or_else(|| self.window.background_opacity())
     }
 
     #[inline]
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -41,8 +41,9 @@ pub struct UIConfig {
     /// Bindings for the mouse.
     mouse_bindings: MouseBindings,
 
-    /// Background opacity from 0.0 to 1.0.
-    background_opacity: Percentage,
+    /// TODO: Deprcated
+    #[config(deprecated = "use `window.background_opacity` instead")]
+    pub background_opacity: Option<Percentage>,
 }
 
 impl Default for UIConfig {
@@ -65,7 +66,10 @@ impl Default for UIConfig {
 impl UIConfig {
     #[inline]
     pub fn background_opacity(&self) -> f32 {
-        self.background_opacity.as_f32()
+        match self.background_opacity {
+            Some(x) => x.as_f32(),
+            None => self.window.background_opacity(),
+        }
     }
 
     #[inline]
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -65,7 +65,9 @@ impl Default for UIConfig {
 impl UIConfig {
     #[inline]
     pub fn background_opacity(&self) -> f32 {
-        self.background_opacity.map(Percentage::as_f32).unwrap_or_else(|| self.window.background_opacity())
+        self.background_opacity
+            .map(Percentage::as_f32)
+            .unwrap_or_else(|| self.window.background_opacity())
     }
 
     #[inline]
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -41,7 +41,6 @@ pub struct UIConfig {
     /// Bindings for the mouse.
     mouse_bindings: MouseBindings,
 
-    /// TODO: Deprcated
     #[config(deprecated = "use `window.background_opacity` instead")]
     pub background_opacity: Option<Percentage>,
 }
--- a/alacritty/src/config/window.rs
+++ b/alacritty/src/config/window.rs
@@ -7,7 +7,7 @@ use serde::de::{self, MapAccess, Visitor};
 use serde::{Deserialize, Deserializer};
 
 use alacritty_config_derive::ConfigDeserialize;
-use alacritty_terminal::config::LOG_TARGET_CONFIG;
+use alacritty_terminal::config::{Percentage, LOG_TARGET_CONFIG};
 use alacritty_terminal::index::{Column, Line};
 
 use crate::config::ui_config::Delta;
@@ -15,7 +15,7 @@ use crate::config::ui_config::Delta;
 /// Default Alacritty name, used for window title and class.
 pub const DEFAULT_NAME: &str = "Alacritty";
 
-#[derive(ConfigDeserialize, Debug, Clone, PartialEq, Eq)]
+#[derive(ConfigDeserialize, Debug, Clone, PartialEq)]
 pub struct WindowConfig {
     /// Initial position.
     pub position: Option<Delta<i32>>,
@@ -50,6 +50,9 @@ pub struct WindowConfig {
 
     /// Initial dimensions.
     dimensions: Dimensions,
+
+    /// Background opacity from 0.0 to 1.0.
+    background_opacity: Percentage,
 }
 
 impl Default for WindowConfig {
@@ -66,6 +69,7 @@ impl Default for WindowConfig {
             class: Default::default(),
             padding: Default::default(),
             dimensions: Default::default(),
+            background_opacity: Default::default(),
         }
     }
 }
@@ -102,6 +106,11 @@ impl WindowConfig {
     #[inline]
     pub fn maximized(&self) -> bool {
         self.startup_mode == StartupMode::Maximized
+    }
+
+    #[inline]
+    pub fn background_opacity(&self) -> f32 {
+        self.background_opacity.as_f32()
     }
 }
 
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -502,7 +502,7 @@ impl Batch {
             bg_r: cell.bg.r,
             bg_g: cell.bg.g,
             bg_b: cell.bg.b,
-            bg_a: cell.bg_alpha as u8,
+            bg_a: (cell.bg_alpha * 255.0) as u8,
         });
     }
 
@@ -851,7 +851,6 @@ impl<'a> RenderApi<'a> {
 
     #[inline]
     fn add_render_item(&mut self, cell: &mut RenderableCell, glyph: &Glyph) {
-        cell.bg_alpha *= 255.;
         if !self.config.colors.opaque_background_colors {
             cell.bg_alpha *= self.config.ui_config.background_opacity();
         }
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -760,9 +760,9 @@ impl<'a> RenderApi<'a> {
         unsafe {
             let alpha = self.config.ui_config.background_opacity();
             gl::ClearColor(
-                (f32::from(color.r) / 255.0).min(1.0) * alpha,
-                (f32::from(color.g) / 255.0).min(1.0) * alpha,
-                (f32::from(color.b) / 255.0).min(1.0) * alpha,
+                (f32::from(color.r) / 255.0).min(1.0),
+                (f32::from(color.g) / 255.0).min(1.0),
+                (f32::from(color.b) / 255.0).min(1.0),
                 alpha,
             );
             gl::Clear(gl::COLOR_BUFFER_BIT);
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -20,7 +20,8 @@ use alacritty_terminal::term::color::Rgb;
 use alacritty_terminal::term::{CursorKey, RenderableCell, RenderableCellContent, SizeInfo};
 
 use crate::config::font::{Font, FontDescription};
-use crate::config::ui_config::{Delta, UIConfig};
+use crate::config::ui_config::Delta;
+use crate::config::Config;
 use crate::cursor;
 use crate::gl;
 use crate::gl::types::*;
@@ -399,7 +400,7 @@ pub struct RenderApi<'a> {
     atlas: &'a mut Vec<Atlas>,
     current_atlas: &'a mut usize,
     program: &'a mut TextShaderProgram,
-    config: &'a UIConfig,
+    config: &'a Config,
     cursor_config: Cursor,
 }
 
@@ -453,7 +454,7 @@ impl Batch {
             bg_r: cell.bg.r,
             bg_g: cell.bg.g,
             bg_b: cell.bg.b,
-            bg_a: (cell.bg_alpha * 255.0) as u8,
+            bg_a: cell.bg_alpha as u8,
         });
     }
 
@@ -636,7 +637,7 @@ impl QuadRenderer {
 
     pub fn with_api<F, T>(
         &mut self,
-        config: &UIConfig,
+        config: &Config,
         cursor_config: Cursor,
         props: &SizeInfo,
         func: F,
@@ -716,7 +717,7 @@ impl QuadRenderer {
 impl<'a> RenderApi<'a> {
     pub fn clear(&self, color: Rgb) {
         unsafe {
-            let alpha = self.config.background_opacity();
+            let alpha = self.config.ui_config.background_opacity();
             gl::ClearColor(
                 (f32::from(color.r) / 255.0).min(1.0) * alpha,
                 (f32::from(color.g) / 255.0).min(1.0) * alpha,
@@ -807,7 +808,11 @@ impl<'a> RenderApi<'a> {
     }
 
     #[inline]
-    fn add_render_item(&mut self, cell: &RenderableCell, glyph: &Glyph) {
+    fn add_render_item(&mut self, cell: &mut RenderableCell, glyph: &Glyph) {
+        cell.bg_alpha *= 255.;
+        if !self.config.colors.opaque_background_colors {
+            cell.bg_alpha *= self.config.ui_config.background_opacity();
+        }
         // Flush batch if tex changing.
         if !self.batch.is_empty() && self.batch.tex != glyph.tex_id {
             self.render_batch();
@@ -830,13 +835,13 @@ impl<'a> RenderApi<'a> {
                     self.load_glyph(&cursor::get_cursor_glyph(
                         cursor_key.shape,
                         metrics,
-                        self.config.font.offset.x,
-                        self.config.font.offset.y,
+                        self.config.ui_config.font.offset.x,
+                        self.config.ui_config.font.offset.y,
                         cursor_key.is_wide,
                         self.cursor_config.thickness(),
                     ))
                 });
-                self.add_render_item(&cell, glyph);
+                self.add_render_item(&mut cell, glyph);
                 return;
             },
             RenderableCellContent::Chars((c, ref mut zerowidth)) => (c, zerowidth.take()),
@@ -860,7 +865,7 @@ impl<'a> RenderApi<'a> {
 
         // Add cell to batch.
         let glyph = glyph_cache.get(glyph_key, self);
-        self.add_render_item(&cell, glyph);
+        self.add_render_item(&mut cell, glyph);
 
         // Render visible zero-width characters.
         if let Some(zerowidth) = zerowidth.filter(|_| !hidden) {
@@ -875,7 +880,7 @@ impl<'a> RenderApi<'a> {
                 // anchor has been moved to the right by one cell.
                 glyph.left += glyph_cache.metrics.average_advance as i16;
 
-                self.add_render_item(&cell, &glyph);
+                self.add_render_item(&mut cell, &glyph);
             }
         }
     }
--- a/alacritty_terminal/src/config/colors.rs
+++ b/alacritty_terminal/src/config/colors.rs
@@ -5,7 +5,7 @@ use alacritty_config_derive::ConfigDeserialize;
 
 use crate::term::color::{CellRgb, Rgb};
 
-#[derive(ConfigDeserialize, Clone, Debug, Default, PartialEq, Eq)]
+#[derive(ConfigDeserialize, Clone, Debug, PartialEq, Eq)]
 pub struct Colors {
     pub primary: PrimaryColors,
     pub cursor: InvertedCellColors,
@@ -16,6 +16,24 @@ pub struct Colors {
     pub dim: Option<DimColors>,
     pub indexed_colors: Vec<IndexedColor>,
     pub search: SearchColors,
+    pub opaque_background_colors: bool,
+}
+
+impl Default for Colors {
+    fn default() -> Self {
+        Self {
+            primary: Default::default(),
+            cursor: Default::default(),
+            vi_mode_cursor: Default::default(),
+            selection: Default::default(),
+            normal: Default::default(),
+            bright: Default::default(),
+            dim: Default::default(),
+            indexed_colors: Default::default(),
+            search: Default::default(),
+            opaque_background_colors: true,
+        }
+    }
 }
 
 impl Colors {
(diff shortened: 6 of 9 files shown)


kchibisov:
I'm not entirely oposed to that feature knowing that certain terminals indeed apply opacity to everything. However I don't like the implementation, you don't need to change anything in rendering part at all. The thing you want to change is to return 0 in https://github.com/alacritty/alacritty/blob/5ee7ae8a274fcc58754aa5a91a14b1dedbb87583/alacritty_terminal/src/term/mod.rs#L364 when the config option is activated.

satlank:
Thanks for looking at the PR and the suggestion (which sounds much simpler). I may be missing something, but it appears to be more complicated than just returning 0, as that seems to make all backgrounds to disappear. E.g. in the attached image the left terminal is with the patch in this PR, the right one from master with the shown diff.

image

Any suggestion where I should be looking?

kchibisov:
Ah, yeah, you need to return `background_opacity()` from `compute_bg_alpha()` and do in text.f.glsl `color = vec4(bg.rgb, 1.0) * bg.a` in a place where you have added multiply by the value from uniform in your patch.

satlank:
Right, okay, that seems to work, albeit it needs to be `color = vec4(bg.rgb, 1.0) * bg.a / 255.0` in `text.f.glsl`, it appears `bg.a` gets scaled to [0,255] somewhere in between returning a float [0,1.0] from `compute_bg_alpha()` and using it in the GSLS program. I am not sure about performance impacts at that point, naively a division instead of a simple multiply seems undesirable and maybe injecting the relevant multiplier directly as done originally is simpler?

It is going to be a bit messy with the config settings: I had added the `keep_background_colors_opaque` to `struct UIConfig` to keep it with `background_opacity`, but that isn't (intentionally, I suppose) accessible from `alacritty_terminal`. However if I move it to `struct Config<T>` in `alacritty_terminal/src/config/mod.rs` the background opacity isn't accessible for use by `RenderApi.clear()` in `alacritty/src/renderer/mod.rs`.

Any thoughts on how this should be structured?

kchibisov:
>gets scaled to [0,255] somewhere in between returning a float [0,1.0] from

it get scaled when pushing instance data in render iirc.

> naively a division instead of a simple multiply seems undesirable and maybe injecting the relevant multiplier directly as done originally is simpler?

well, you need to adjust vertex shader to make division, forgot about that part. it's already there, just need to include alpha in it as well.

 
>It is going to be a bit messy with the config settings: I had added the keep_background_colors_opaque to struct UIConfig to keep it with background_opacity, but that isn't (intentionally, I suppose) accessible from alacritty_terminal. However if I move it to struct Config<T> in alacritty_terminal/src/config/mod.rs the background opacity isn't accessible for use by RenderApi.clear() in alacritty/src/renderer/mod.rs.

that's the most annoying part here tbqh. In general, I'd like to see splitting out the Renderable iterator from alacritty_terminal, since rgb conversions must be done by alacritty, since it makes it really hard to use the lib in cases like embedding in other applications, etc,  however it's not that simple to approach, but I can't think of some simple solution to that problem of passing opacity around.

@chrisduerr do you have any opinion on 'extracting iterator' from `alacritty_terminal` or changing things that it returns an iterator over visible chars(with touching them) and you in alacritty builds another iterator out of that or maps its elements?

The original solution with uniform isn't that bad, but I just feel like it's too much hasstle for opacity option, knowing that we already have one and making all things transparent is subjecive.

chrisduerr:
> @chrisduerr do you have any opinion on 'extracting iterator' from alacritty_terminal or changing things that it returns an iterator over visible chars(with touching them) and you in alacritty builds another iterator out of that or maps its elements?

I feel like any suggestion in this direction would be out of scope for this issue? If we want to do that, it should be done separately. But in general nothing stops you from having unmapped colors returned by the cell iterator, right?

kchibisov:
>I feel like any suggestion in this direction would be out of scope for this issue?

Yeah, it'd a bit out of scope for that issue.

>But in general nothing stops you from having unmapped colors returned by the cell iterator, right?

Right.

satlank:
Ok, I am not sure I am following the intricacies of where cell iterators ought to be (and the finer points of what belongs into `alacritty` and what into `alacritty_terminal`). I does sound like the opacity handling should all happen in `alacritty`, but it would require refactoring to move the relevant bits from `alacritty_terminal` to `alacritty` first?

Looking at the configuration conundrum again, there is a way to leave `background_opacity` in `UIConfig`: Instead of introducing a toggle (`keep_background_colors_opaque`) I can add another opacity value `non_default_background_opacity` (needs a better name, naming is hard, bear with me) which would be attached to `Config<T>`, defaulted to 1.0 and used instead of the hard-coded 1.0 in `compute_bg_alpha()` - it should be set to the same value as `background_opacity` or 1.0 to recover the desired behaviour. But that doesn't look particularly appealing either.

Maybe it's best to not merge this feature then and revisit it when the separation is cleaner?

chrisduerr:
> It does sound like the opacity handling should all happen in alacritty, but it would require refactoring to move the relevant bits from alacritty_terminal to alacritty first?

Yes. Currently it is not done ***entirely*** in `alacritty`, which it should.

> Looking at the configuration conundrum again, there is a way to leave background_opacity in UIConfig

Just to be clear, when it comes to configuration options there is no other choice. It must be in the UIConfig, we can't put UI config stuff in the `alacritty_terminal` config. Since moving it after the fact is just going to be a massive pain. In fact the `background_opacity` option should be on `window`.

> Maybe it's best to not merge this feature then and revisit it when the separation is cleaner?

I'm open to PRs that separate the code. It's likely the easiest solution. I'm just saying that it shouldn't be done in this PR.

kchibisov:
Well, as an alternative solution we can do something like
```diff
diff --git a/alacritty/res/text.f.glsl b/alacritty/res/text.f.glsl
index c0c1bd1b..de0e9fa9 100644
--- a/alacritty/res/text.f.glsl
+++ b/alacritty/res/text.f.glsl
@@ -16,7 +16,7 @@ void main()
             discard;

         alphaMask = vec4(1.0);
-        color = vec4(bg.rgb, 1.0);
+        color = vec4(bg.rgb, 1.0) * bg.a;
     } else {
         if (fg.a != 0.0) {
             // Color glyphs, like emojis.
diff --git a/alacritty/res/text.v.glsl b/alacritty/res/text.v.glsl
index 0f5fd3f9..bff3f42d 100644
--- a/alacritty/res/text.v.glsl
+++ b/alacritty/res/text.v.glsl
@@ -56,6 +56,6 @@ void main()
         TexCoords = uvOffset + position * uvSize;
     }

-    bg = vec4(backgroundColor.rgb / 255.0, backgroundColor.a);
+    bg = backgroundColor / 255.;
     fg = vec4(textColor.rgb / 255.0, textColor.a);
 }
diff --git a/alacritty/src/renderer/mod.rs b/alacritty/src/renderer/mod.rs
index 6b27a837..3d66b96c 100644
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -491,7 +491,7 @@ impl Batch {
             bg_r: cell.bg.r,
             bg_g: cell.bg.g,
             bg_b: cell.bg.b,
-            bg_a: (cell.bg_alpha * 255.0) as u8,
+            bg_a: cell.bg_alpha as u8,
             multicolor: glyph.multicolor,
         });
     }
@@ -995,7 +995,13 @@ impl<'a> RenderApi<'a> {
     }

     #[inline]
-    fn add_render_item(&mut self, cell: RenderableCell, glyph: &Glyph) {
+    fn add_render_item(&mut self, mut cell: RenderableCell, glyph: &Glyph) {
+        if self.config.keep_background_colors_opaque {
+            cell.bg_alpha *= 255.;
+        } else {
+            cell.bg_alpha *= 255. * self.config.background_opacity();
+        }
+
         // Flush batch if tex changing.
         if !self.batch.is_empty() && self.batch.tex != glyph.tex_id {
             self.render_batch();

```

satlank:
Thanks for the clarifications! I'll try to update the PR tonight, I am on the wrong machine for most of the day, your alternative solution looks like a good middle ground.

satlank:
Ok, all updated. It certainly is a neater diff now, thanks again for the improvement suggestions.

alacritty.yml
```diff
@@ -321,6 +321,11 @@
 # Window opacity as a floating point number from `0.0` to `1.0`.
 # The value `0.0` is completely transparent and `1.0` is opaque.
 #background_opacity: 1.0
+#
+# Whether or not `background_opacity` applies to all backgrounds or only the
+# default background.  When set to `true` it will only apply to the default
+# background and keep other background colors opaque.
+#keep_background_colors_opaque: true
```
chrisduerr, line 328: I think the naming of this option isn't ideal, since it says little about what it does in its name and I'd advocate for moving this option to `window` instead of keeping it at the top level. Together with `background_opacity` probably.

alacritty/src/renderer/mod.rs
```diff
@@ -995,7 +995,13 @@ impl<'a> RenderApi<'a> {
     }
 
     #[inline]
-    fn add_render_item(&mut self, cell: RenderableCell, glyph: &Glyph) {
+    fn add_render_item(&mut self, mut cell: RenderableCell, glyph: &Glyph) {
+        if self.config.keep_background_colors_opaque() {
+            cell.bg_alpha *= 255.;
+        } else {
+            cell.bg_alpha *= 255. * self.config.background_opacity();
+        }
```
chrisduerr, line 1003:
```suggestion
        cell.bg_alpha *= 255.;
        if !self.config.keep_background_colors_opaque() {
            cell.bg_alpha *= self.config.background_opacity();
        }
```

chrisduerr (changes requested):

alacritty/src/renderer/mod.rs
satlank: Oh, yes, will do that.

alacritty.yml
satlank: I am open to suggestions for a better name. My 'inspiration' for naming it that way came from iterm (not that that is a strong argument):
image

Happy to move it to the window section, do you want me to move only the new option for now so that it is in the right place to start with or also `background_opacity`? If the latter what is the procedure to deprecate configuration options or is a breaking change fine?

alacritty/src/renderer/mod.rs
satlank: That's done.

alacritty.yml
chrisduerr: > My 'inspiration' for naming it that way came from iterm (not that that is a strong argument):

I mean that's definitely a solid argument, thanks for letting me know. I'm struggling to suggest a much better alternative though. My main problem is that the `keep_` prefix doesn't really fit in well with our configuration options. Maybe something more like `opaque_colored_backgrounds`?

If you have some other suggestions, please do let me know.

> Happy to move it to the window section, do you want me to move only the new option for now so that it is in the right place to start with or also background_opacity? If the latter what is the procedure to deprecate configuration options or is a breaking change fine?

I think I'd move both, just to have it all in place. Though I'm not entirely sure if this should be under `colors` or `window`. @kchibisov any thoughts?

kchibisov: it should be under window, I'd say.

kchibisov: >I mean that's definitely a solid argument, thanks for letting me know. I'm struggling to suggest a much better alternative though. My main problem is that the keep_ prefix doesn't really fit in well with our configuration options. Maybe something more like opaque_colored_backgrounds?

We can have something like `apply_opacity_to_background_colors: false` setting, but I don't really mind original option.

kchibisov (changes requested):
Could you move your new config option to `window` section, and also `background_opacity` (with adding deprecation for that one)?

alacritty.yml
chrisduerr: That's a tad long.

chrisduerr:
Okay, I've taken a bit of a look at this myself and I think the following is what I would recommend.

<details>

```diff
diff --git a/alacritty.yml b/alacritty.yml
index ac4477f..8a02ca6 100644
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -21,6 +21,12 @@
   #TERM: alacritty
 
 #window:
+  # Background opacity
+  #
+  # Window opacity as a floating point number from `0.0` to `1.0`.
+  # The value `0.0` is completely transparent and `1.0` is opaque.
+  #background_opacity: 1.0
+
   # Window dimensions (changes require restart)
   #
   # Specified in number of columns/lines, not pixels.
@@ -172,6 +178,13 @@
 
 # Colors (Tomorrow Night)
 #colors:
+  # Opaque backgrounds with transparency enabled
+  #
+  # Whether or not `background_opacity` applies to all backgrounds or only the
+  # default background. When set to `false` all cells will be transparent
+  # regardless of their background color.
+  #opaque_background_colors: true
+
   # Default colors
   #primary:
   #  background: '#1d1f21'
@@ -316,12 +329,6 @@
   #
   #command: None
 
-# Background opacity
-#
-# Window opacity as a floating point number from `0.0` to `1.0`.
-# The value `0.0` is completely transparent and `1.0` is opaque.
-#background_opacity: 1.0
-
 #selection:
   # This string contains all characters that are used as separators for "semantic words" in Alacritty.
   #semantic_escape_chars: ",│`|:\"' ()[]{}<>\t"
```

</details>

This should be a nice descriptive name and simple to understand documentation. Having one on color while the other on window I'd explain since things like blur/transparency are usually understood on a window level, but the transparency of individual colors is not really related to the window.

Let me know what you think.

satlank:
> This should be a nice descriptive name and simple to understand documentation. Having one on color while the other on window I'd explain since things like blur/transparency are usually understood on a window level, but the transparency of individual colors is not really related to the window.
> 
> Let me know what you think.

I like the new naming, so did that. I have left both options together in the config though for two reasons, a) it makes more sense to me to have that options with the opacity settings as it toggles how opacity is applied, and b) when moved to colors it isn't accessible in the render (that only has access to the UIConfig). Can change the logic on where it is applied to address b) I suppose.

alacritty.yml
satlank: I've changed it your suggestion.

satlank:
> Could you move your new config option to window section, and also background_opacity (with adding deprecation for that one)?

That's done.

alacritty/src/config/window.rs
```diff
@@ -13,7 +15,7 @@ use crate::config::ui_config::{DefaultTrueBool, Delta};
 pub const DEFAULT_NAME: &str = "Alacritty";
 
 #[serde(default)]
-#[derive(Deserialize, Debug, Clone, PartialEq, Eq)]
```
satlank, line 16: Needed to drop `Eq` as `Percentage` is effectively a `f32` and thus doesn't have `Eq`.

alacritty/src/config/ui_config.rs
```diff
@@ -44,9 +44,9 @@ pub struct UIConfig {
     #[serde(default, deserialize_with = "failure_default")]
     live_config_reload: DefaultTrueBool,
 
-    /// Background opacity from 0.0 to 1.0.
+    // TODO: DEPRECATED
     #[serde(default, deserialize_with = "failure_default")]
-    background_opacity: Percentage,
+    pub background_opacity: Option<Percentage>,
```
satlank, line 49: `pub` because otherwise the deprecation check can't access it.

chrisduerr:
> it makes more sense to me to have that options with the opacity settings as it toggles how opacity is applied

It has nothing to do with the window though.

> when moved to colors it isn't accessible in the render (that only has access to the UIConfig). Can change the logic on where it is applied to address b) I suppose.

That's something that can be easily changed.

kchibisov:
>That's something that can be easily changed.

Oh, but colors is a part of a alacritty_terminal config, so and moving opacity there is a bit of a problem tbqh, since it has nothing to do with a terminal. In general colors should be in UIConfig, but it's not that simple to do.

What do you think about having them at a top level for a while? And once we move colors to UIConfig we can move opacity things as well?

Alternatively we can go to a thing we've used in a past with `Config<UIConfig>` in a renderer, so not really a big deal, and it's something simple to change.

chrisduerr:
> Oh, but colors is a part of a alacritty_terminal config

I mean that'll change in the future anyways.

> What do you think about having them at a top level for a while?

I think that introducing a config option knowing fully well it won't stay where it is will only introduce unnecessary churn for users.

> Alternatively we can go to a thing we've used in a past with Config<UIConfig> in a renderer, so not really a big deal, and it's something simple to change.

There's absolutely nothing wrong with passing the full config to the renderer, it changes nothing. The only thing we cannot do is pass the UIConfig to the terminal.

satlank:
Sorry, have been snowed under the last couple of days. I'll try to get back to this tomorrow.

caseyavila:
Any updates/progress for this feature?

satlank:
> Any updates/progress for this feature?

Yeah, sorry, half-forgot about this. I'll try to get back to this over the holidays, need to update my local binaries with master anyway.

ric2work:
Hi

satlank:
Ok, I think this has addressed the changes requested (and I've also rebased onto master dealing with the merge conflicts).

@chrisduerr, apologies for the delay in getting back to this. Is there anything else that needs changing?

alacritty.yml
```diff
@@ -284,6 +290,13 @@
   #
   #indexed_colors: []
 
+  # Opaque backgrounds with transparency enabled
+  #
+  # Whether or not `background_opacity` applies to all backgrounds or only the
+  # default background. When set to `false` all cells will be transparent
+  # regardless of their background color.
```
chrisduerr, line 297:
```suggestion
  # Whether or not `background_opacity` applies to all backgrounds or only to the
  # default background. When set to `false` all cells will be transparent
  # regardless of their background color.
```

alacritty.yml
```diff
@@ -284,6 +290,13 @@
...
+  #opaque_background_colors: true
```
chrisduerr, line 298: It has been a while and I don't completely remember our previous discussions, but `opaque_colored_backgrounds` might be an alternative name for this that is a little more clear? I'm not sure though, I think both would probably be fine.

alacritty/res/text.f.glsl
```diff
@@ -18,7 +18,7 @@ void main() {
         }
 
         alphaMask = vec4(1.0);
-        color = vec4(bg.rgb, 1.0);
+        color = vec4(bg.rgb, 1.0) * bg.a;
```
chrisduerr, line 21: I'm not sure this logic is correct? Just looking at the results, it seems like `color = vec4(bg.rgb, bg.a);` gives me results closer to what I would expect?

Neither of them seem to actually be correct though, since it seems like they aren't independent from the actual background color? I'd expect a transparent cell background with the same hex color as the background to have the exact same color, but it doesn't. This is likely because the cell's background is getting merged with the background of the terminal.

alacritty/src/config/ui_config.rs
```diff
@@ -41,8 +41,9 @@ pub struct UIConfig {
     /// Bindings for the mouse.
     mouse_bindings: MouseBindings,
 
-    /// Background opacity from 0.0 to 1.0.
-    background_opacity: Percentage,
+    /// TODO: Deprcated
```
chrisduerr, line 44: This annotation isn't necessary anymore, the new `#[config(deprecated)]` annotation is sufficient to fiend deprecated fields.

alacritty/src/config/ui_config.rs
```diff
@@ -65,7 +66,10 @@ impl Default for UIConfig {
 impl UIConfig {
     #[inline]
     pub fn background_opacity(&self) -> f32 {
-        self.background_opacity.as_f32()
+        match self.background_opacity {
+            Some(x) => x.as_f32(),
+            None => self.window.background_opacity(),
+        }
```
chrisduerr, line 72: This might work:

```rust
self.background_opacity.map(Percentage::as_f32).unwrap_or_else(|| self.window.background_opacity())
```

alacritty/src/renderer/mod.rs
```diff
@@ -807,7 +808,11 @@ impl<'a> RenderApi<'a> {
     }
 
     #[inline]
-    fn add_render_item(&mut self, cell: &RenderableCell, glyph: &Glyph) {
+    fn add_render_item(&mut self, cell: &mut RenderableCell, glyph: &Glyph) {
+        cell.bg_alpha *= 255.;
```
chrisduerr, line 812: This can probably stay where it was? Doesn't matter by which value we multiply first, so we can keep it in the 1.0 to 0.0 range for now.

alacritty/src/renderer/mod.rs
```diff
@@ -807,7 +808,11 @@ impl<'a> RenderApi<'a> {
...
+        if !self.config.colors.opaque_background_colors {
+            cell.bg_alpha *= self.config.ui_config.background_opacity();
+        }
```
chrisduerr, line 815: Is this the right place for this? If `cell` already has a `bg_alpha` field, shouldn't this property be applied when this field is initialized? Doing it right here seems somewhat arbitrary?

alacritty_terminal/src/config/colors.rs
```diff
@@ -16,6 +16,24 @@ pub struct Colors {
...
+            primary: Default::default(),
+            cursor: Default::default(),
+            vi_mode_cursor: Default::default(),
+            selection: Default::default(),
+            normal: Default::default(),
+            bright: Default::default(),
+            dim: Default::default(),
+            indexed_colors: Default::default(),
+            search: Default::default(),
+            opaque_background_colors: true,
```
chrisduerr, line 34:
```suggestion
            opaque_background_colors: true,
            primary: Default::default(),
            cursor: Default::default(),
            vi_mode_cursor: Default::default(),
            selection: Default::default(),
            normal: Default::default(),
            bright: Default::default(),
            dim: Default::default(),
            indexed_colors: Default::default(),
            search: Default::default(),
```

Putting non-default fields at the top makes it easy to see why there is a custom `Default` implementation.

chrisduerr (changes requested):
Thanks for picking this back up.

alacritty.yml
satlank: I don't have strong feelings either way. Whatever you prefer.

alacritty/src/config/ui_config.rs
satlank: Ah, neat, also removes my typo o_O..

alacritty/src/config/ui_config.rs
satlank: Yes, looks like it would after some mental gymnastics. Not too used to the unwrap-chain bits..  Is that any clearer though?

alacritty.yml
chrisduerr: Let's just leave it as-is for now. I don't really think it's a significant improvement.

alacritty/src/renderer/mod.rs
satlank: I think there was a reason to do this, but I don't recall. I'll try whether I can figure out again why I changed that here.

alacritty/src/config/ui_config.rs
chrisduerr: Yeah, I'd recommend this over the existing code.

alacritty/res/text.f.glsl
satlank: Hm. I vaguely remember thinking about this and preferring it to darken the colours as well when applying the opacity. I'll go through this again and see if I can retrace my steps.

alacritty/src/renderer/mod.rs
satlank: Well, I think I basically followed https://github.com/alacritty/alacritty/pull/4196#issuecomment-686057677 with where to put the scaling. Maybe there is a better way to put it, since I am anyway going to retrace my steps to address the other options I'll bear this is in mind an see if there is a more logical place.

chrisduerr: Ah, I've previously said the following:



> > It does sound like the opacity handling should all happen in alacritty, but it would require refactoring to move the relevant bits from alacritty_terminal to alacritty first?
>
> Yes. Currently it is not done entirely in alacritty, which it should.

This is technically correct, but fact is that opacity handling happens in `alacritty_terminal` already. So if we want to move it outside of `alacritty_terminal`, it should probably all be moved. I believe that's probably what I was thinking of back then too. Generally though that's a lot of work, so I'd be fine with putting it in `alacritty_terminal` for now, you should be able to find the other alpha stuff in `RenderableCell::new` I think.

alacritty/res/text.f.glsl
chrisduerr: As I've stated, I think both are incorrect. Generally the ideal solution would be to do this *correctly*, but I haven't looked into that so I can't show you a solution for that.

But at least `vec4(bg.rgb, bg.a)` looks like it is a little more correct at least on a dark background.

satlank: Okay, I think the reason I went with the multiplier on all instead of just setting alpha was that the latter messed up the background more visibly when entering tmux (it's a bit difficult to tell, but the issue is the more whiteish background in tmux compared to the non-tmux in the last two screenshots below - the opacity is set to 0.4 to have a more extreme impact, I normally run with 0.9).

I think you are right that there is something fishy going on, it does look like that tmux/no-tmux does look slightly different in my implementation as well. But other than disecting the complete rendering path I don't know how to do it properly - so certainly won't come up with a solution today and possibly not this year either. I'll poke around as I get time over the next days. I *think* that my original implementation (which just backed out changes that switch the original opacity-to-all to the now-in-master opacity-only-to-default-bg) didn't have this problem, so maybe that's a place to start. 

master:
master
master tmux:
master tmux
this branch:
branch
this branch tmux:
branch-tmux
this branch only alpha:
only-alpha
this branch only alpha tmux:
only-alpha-tmux

alacritty/src/config/ui_config.rs
satlank: Okay, that's done.

alacritty/src/config/ui_config.rs
satlank: Dropped.

alacritty/res/text.f.glsl
satlank: Ah, the difference in bg between tmux/non-tmux case is because of how the background opacity is currently done in master (https://github.com/alacritty/alacritty/blob/master/alacritty/src/renderer/mod.rs#L761-L766), it doesn't just set the alpha channel, it also scales the RGB channels. I don't have an opinion on which is preferable, but changing both `clear` to only set the alpha channel and `color = vec4(bg.rgb, bg.a)` I don't see a difference in bg colour between my tmux/non-tmux cases.
So if you are happy to change how the opacity is applied also in the 'only for the default bg' case, then it works to just apply it to the alpha channel.
For how I am using the opacity setting it doesn't make much of a difference. But for a smaller opacity setting the difference is much more notable, though I find it hard to tell which version 'looks better'.

chrisduerr: I have no problem with changing things as long as it works properly. There shouldn't be much "opinion" involved. 50% opacity should be a pretty clear instruction usually.

satlank: ¯\_(ツ)_/¯
Changed to only apply the opacity to the alpha channel.

alacritty/src/renderer/mod.rs
satlank: I think I see where to do that in alacritty_terminal. However while I can get to `config.colors.opaque_background_color` I am not sure that `config.ui_config.background_opacity()` is accessible from there - at least the compiler keeps shouting at me? I suppose I could move the `background_opacity` to `color`, but given that you were keen on moving it to `window` I am not sure that is desirable (plus I am not sure how to support the previous location of that field)?

chrisduerr: Fair enough, I'd hold on this for now then I suppose.

alacritty/src/renderer/mod.rs
satlank: Not sure why that got changed. Seems to be fine to leave that scaling in the batch so moved it back there.

alacritty/res/text.f.glsl
chrisduerr: You just need to watch out, because the glyph data is not in RGB format.

dankamongmen:
If you're going to do this, I'd appreciate you making it discoverable from clients. Something might be relying on that background color. Overall I think this is a pretty bad idea. What if my program is indicating an error state by throwing a red background on some glyphs? With this patch, as I understand things, you'd throw away that important indicator for...what, exactly?

Kitty does this with one particular color, and I already have to deal with that in Notcurses. Please see this bug:

 https://github.com/dankamongmen/notcurses/issues/1117

and see how this can have unexpected effects (you break any kind of plot or other non-text being drawn with block characters, for instance -- imagine that i'm drawing with half blocks. One half of the cell is my foreground color, one is my background color. You just mapped half of my "pixels" to the same color).

I didn't know this about iTerm2, but I'm going to go file a feature request against them. In the meantime, I feel such a feature is incomplete without:

* a way to discover whether it is enabled, or at least push the current setting onto a stack, and
* a way to disable/enable this behavior at runtime, ideally expressed as a terminfo extension capability

otherwise you're introducing a feature with ramifications the user might not foresee, with no way to work around it in clients that know they want to. please don't, @satlank .

caseyavila:
Well this is set up to be configured in the config file, so not only is it optional, but I suppose one could read the configuration file to figure out whether the feature is toggled.

chrisduerr:
https://github.com/alacritty/alacritty/pull/5297

dankamongmen:
> Well this is set up to be configured in the config file, so not only is it optional, but I suppose one could read the configuration file to figure out whether the feature is toggled.

no other capabilities are discovered in such a manner, for good reason. what if there is no config file? i'd have to know/track the default. what if they're using a config file specified via `--config-file`? how am i supposed to know their `XDG_CONFIG_HOME`? was this a serious suggestion?

Issue #4187, opened 2020-08-31
Spawning a new instance does not retain the --config-file command-line option 


marioroy:
Thank you for the `--config-file` command-line option. Unfortunately, spawning a new window instance via the keyboard shortcut `SpawnNewInstance` loads the default configuration and not the configuration specified for the initial window.

My default configuration `%APPDATA%/alacritty/alacritty.yml` is configured for WSL 2 Ubuntu 20.04.
```text
window:
  dynamic_title: false
  title: Ubuntu

# WSL 2 Ubuntu 20.04
shell:
  program: C:\Windows\System32\bash
  args:
    - --login

working_directory: C:\Users\mario
```

I have another config for Cygwin `%APPDATA%/alacritty/alacritty_cygwin.yml`.
```text
window:
  dynamic_title: false
  title: Cygwin

shell:
  program: C:\cygwin64\bin\bash
  args:
    - --login

working_directory: C:\cygwin64\home\mario
```

I have two desktop icons `alacritty - ubuntu` and `alacritty - cygwin`. The one for Cygwin is called with the option `--config-file %APPDATA%\alacritty\alacritty_cygwin.yml`. Spawning a new instance window does not check if the config-file option was specified initially so that to use the same config when spawning a new window.

Regards.

chrisduerr:
I think it makes sense to persist CLI arguments. It should also be fairly easy to do so.

marioroy:
Today, I learned of being able to specify a command and arguments. Thank you.
```text
key_bindings:
  - { key: K,        mods: Command, mode: ~Vi, chars: "\x0c" }
  - { key: L,        mods: Command, mode: ~Vi, chars: "\x0c" }
  - { key: C,        mods: Command, action: Copy             }
  - { key: V,        mods: Command, action: Paste            }
  - { key: Q,        mods: Command, action: Quit             }
  - { key: W,        mods: Command, action: Quit             }
  - { key: N,        mods: Command, command: { program: 'C:\Users\mario\Downloads\Alacritty-v0.5.0-windows-portable\alacritty.exe', args: ['--config-file', 'C:\Users\mario\AppData\Roaming\alacritty\alacritty_cygwin.yml'] } }
  - { key: Key0,     mods: Command, action: ResetFontSize    }
  - { key: Equals,   mods: Command, action: IncreaseFontSize }
  - { key: Add,      mods: Command, action: IncreaseFontSize }
[...]

PR #4197, opened 2020-09-01
Persist Config file while spawning new instance

RohanPoojary:
PR to address issue: [4187](https://github.com/alacritty/alacritty/issues/4187).
This is my first contribution, please let me know if this isn't the expected solution.

I have tested it on my macOS(Mojave) and the new window uses the expected config file.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -302,8 +302,10 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         let mut env_args = env::args();
         let alacritty = env_args.next().unwrap();
 
+        let working_directory_set;
+
         #[cfg(unix)]
-        let args = {
+        let mut args = {
             // Use working directory of controlling process, or fallback to initial shell.
             let mut pid = unsafe { libc::tcgetpgrp(tty::master_fd()) };
             if pid < 0 {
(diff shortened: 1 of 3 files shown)


alacritty/src/event.rs
```diff
@@ -314,9 +314,15 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
 
-            fs::read_link(link_path)
+            let mut args = fs::read_link(link_path)
                 .map(|path| vec!["--working-directory".into(), path])
-                .unwrap_or_default()
+                .unwrap_or_default();
+
+            // Adding currently loaded config-file in arguments.
+            args.push("--config-file".into());
+            args.extend(self.config.ui_config.config_paths.clone());
```
chrisduerr, line 323: We'd want to load all current arguments, not just config file. So we should lookup the current cmdline and just append it. That could be a bit of a problem if the working directory parameter already exists, but in general it is the most sensible solution.

chrisduerr (changes requested):

alacritty/src/event.rs
RohanPoojary: Agreed. Will append the arguments by reading it from command line arguments.
Also for windows the args variable expects it to be `Vec<String>` while on Unix its `Vec<OsStr>` is the difference expected?

RohanPoojary: @chrisduerr any remarks?

chrisduerr (approved):
Since I wasn't entirely sure how to best approach this, I've decided to implement the remaining stuff myself. Existing working directories should now be filtered out and all other CLI parameters are still passed correctly.

Since I've wrote a chunk of this code, @kchibisov should probably review it.

kchibisov (changes requested):
It's not implemented for Windows however nothing stops us from doing so.

alacritty/src/event.rs
```diff
@@ -299,10 +299,13 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     }
 
     fn spawn_new_instance(&mut self) {
-        let alacritty = env::args().next().unwrap();
+        let mut env_args = env::args();
+        let alacritty = env_args.next().unwrap();
+
+        let working_directory_set;
```
kchibisov, line 305: The changes look fine, however I wonder whether we can return set working directory by returning a tuple in a place where you set `args`. so it'll be `let (mut args, working_directory_set) = { (args, false) };` That should save a bunch of lines.

chrisduerr: Actually there's a far simpler solution, I was just being silly.

alacritty/src/event.rs
```diff
@@ -314,12 +315,29 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
         };
+
         #[cfg(not(unix))]
-        let args: Vec<String> = Vec::new();
+        let mut args: Vec<PathBuf> = Vec::new();
+
+        let working_directory_set = !args.is_empty();
+
+        // Reuse the arguments passed to Alacritty for the new instance.
+        while let Some(arg) = env_args.next() {
```
RohanPoojary, line 332: I think we can just use `filter` method here.
`
        args.extend(
            env_args
                .filter(|x| !working_directory_set || x != "--working-directory")
                .map(|x| x.into()),
        );

`

chrisduerr: We can't, look at what the code does. It doesn't just filter one element.

RohanPoojary: Alright, So this will be a place to filter out other arguments as well?

RohanPoojary: Oh right, my bad

kchibisov (approved):

PR #778, opened 2017-09-08
Change cursor from pointer to text

DSpeckhals:
It seems that (as a rule) terminal emulators use the text mouse cursor rather
than the pointer that is used now. This commit changes the cursor to
using winit's built-in configuration function.

PR #4216, opened 2020-09-09
Add config option to use "default" mouse cursor style instead of "text"

NicholasAsimov:
Since I've upgraded from Alacritty v0.3.3 I've noticed that the [bug was fixed](https://github.com/alacritty/alacritty/commit/4dd327f0aec53d4501601281b796b1d7dda9a5d3) making the "text selection" mouse cursor an actual default.

I would really like an option to get back to how it was before and be able to choose whether "text" or "default" cursor style should be used.

I understand that adding a config option just for this might be excessive and I would love to hear any better/simpler ways to provide this choice for the user. Even though the "text" cursor [is a default in many terminals](https://github.com/alacritty/alacritty/pull/778) it still varies in my experience.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Configuration file option for sourcing other files (`import`)
 - CLI parameter `--option`/`-o` to override any configuration field
 - Escape sequences to report text area size in pixels (`CSI 14 t`) and in characters (`CSI 18 t`)
+- Option `mouse.force_default_cursor` to force the use of "default" mouse cursor style instead of "text"
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


kchibisov:
Thanks for the patch, but I don't see a reason for that.

It was always in a way that we have `Text` cursor and I don't see any intention to change it to `left_ptr`. On some systems it was in a way that it wasn't text by default, since the cursor setting was broken in winit(the library we're using) on those backends.

The `text` cursor shape tells you that you can select a text. We're also using `left_ptr` when you can't select a text, and a `hand` when you can click on something, like url.

For example if you enter `htop` you'll likely see a `left_ptr` cursor, since you can't select. The same applies to a `tmux` with enabled mouse mode, since mouse is captured by it, meaning that you can't select with our selection. However in `htop`/`tmux`(with mouse mode) if you press `Shift` the cursor will be changed to `Text` giving you a clear **indication** that you can select text.

PR #4226, opened 2020-09-13
Add changelog entry for FreeType version bump

chrisduerr:


chrisduerr (commented):
This is quite relevant information for package maintainers on Linux, so it's probably a good idea to make this very clear.

kchibisov:
>This is quite relevant information for package maintainers on Linux, so it's probably a good idea to make this very clear.

The problem is that it's in 0.5.0, so the packagers should check that on 0.5.0, however the minimum version is required for source based distros(assuming a lot of version of freetype) tbqh, since in binary distros you'll get a build failure or something early on. Maybe we can communicate it a bit better or let's say fix that in cross-font, since it's pretty trivial to fix if you think so.

chrisduerr:
I don't think this is easy to fix in crossfont. But would you add it to 0.5.0 instead? I forgot this is part of 0.5.0, I must have misremembered.

kchibisov:
Well, we can fix it for master and then backport to 0.5.1, since 0.5.0 is already out.

chrisduerr:
How do you suggest we "fix" it though? Adding a feature flag doesn't fix the issue.

kchibisov:
>How do you suggest we "fix" it though? Adding a feature flag doesn't fix the issue.

I'd send a fix for cross font when I get time.

chrisduerr:
How do you plan to fix it though?

kchibisov:
>How do you plan to fix it though?

Check that function exists in build.rs and pass cfg to rustc? that's how you do conditional compilation.

chrisduerr:
Ah, so you do want to verify the installed freetype version. That's okay I guess.

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Prebuilt Linux binaries have been removed
 - Added manpage, terminfo, and completions to macOS application bundle
 - On Linux/BSD the build will fail without Fontconfig installed, instead of building it from source
+- Minimum FreeType version has been bumped to 2.8 on Linux
```
kchibisov, line 55:
```suggestion
- Minimum FreeType version has been bumped to 2.8 on Linux/BSD
```

kchibisov (changes requested):

kchibisov (approved):

PR #4232, opened 2020-09-14
Update cargo lockfile

chrisduerr:
This updates the lockfile to the new format introduced in Rust 1.41.0.
Since the lockfile has been completely regenerated, all depedencies were
updated.


kchibisov (approved):
Thanks.

PR #4245, opened 2020-09-23
Log new padding and cell size on 'SizeInfo' update

kchibisov:
This makes our logging of 'SizeInfo' changes more consistent
with the one we're using when creating a new 'Display'.

### DIFF
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -435,10 +435,10 @@ impl Display {
         let physical = PhysicalSize::new(self.size_info.width as u32, self.size_info.height as u32);
         self.window.resize(physical);
 
+        self.renderer.resize(&self.size_info);
+        
         info!("Padding: {} x {}", self.size_info.padding_x, self.size_info.padding_y);
         info!("Width: {}, Height: {}", self.size_info.width, self.size_info.height);
-
-        self.renderer.resize(&self.size_info);
     }
 
     /// Draw the screen.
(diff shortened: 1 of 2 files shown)


alacritty/src/display.rs
```diff
@@ -432,6 +434,10 @@ impl Display {
         // Resize renderer.
         let physical = PhysicalSize::new(self.size_info.width as u32, self.size_info.height as u32);
         self.window.resize(physical);
+
+        info!("Padding: {} x {}", self.size_info.padding_x, self.size_info.padding_y);
+        info!("Width: {}, Height: {}", self.size_info.width, self.size_info.height);
+
         self.renderer.resize(&self.size_info);
```
chrisduerr, line 441:
```suggestion

        self.renderer.resize(&self.size_info);
        
        info!("Padding: {} x {}", self.size_info.padding_x, self.size_info.padding_y);
        info!("Width: {}, Height: {}", self.size_info.width, self.size_info.height);
```

chrisduerr (changes requested):

alacritty/src/display.rs
```diff
@@ -432,7 +434,11 @@ impl Display {
         // Resize renderer.
         let physical = PhysicalSize::new(self.size_info.width as u32, self.size_info.height as u32);
         self.window.resize(physical);
+
```
chrisduerr, line 437:
```suggestion
```

chrisduerr (changes requested):

alacritty/src/display.rs
kchibisov: lol.

chrisduerr (approved):

PR #4248, opened 2020-09-24
Remove Unused Linear Type

nixpulvis:
Just noticed that this new type isn't used anywhere, and I frankly can't imagine what it's purpose could be. Perhaps this is motivated by my long standing desire to remove `Line` and `Column` as well... but at least this is a smaller, and hopefully less contentious removal.

### DIFF
--- a/alacritty_terminal/src/index.rs
+++ b/alacritty_terminal/src/index.rs
@@ -220,28 +220,6 @@ impl fmt::Display for Column {
     }
 }
 
-/// A linear index.
-///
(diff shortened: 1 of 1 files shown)


chrisduerr (approved):
> Just noticed that this new type isn't used anywhere

New type? More like ancient type. This was originally used for selection I believe to store the position of its start/end as linear points combining both line and column into a single value. Though realistically that just made things more difficult, not easier, which is why I replaced it.

I didn't realize this was still a thing, thanks for removing it.

PR #4259, opened 2020-09-26
Fix selection incorrectly expanding when scrolled in history

### LINKED ISSUE #4257: Selection selects by 2 chars after catting file with unicode
If you cat a file like https://github.com/alacritty/vtebench/blob/master/benchmarks/unicode/symbols

scroll to a point where you catted it, and try to select something, selection will always select by 2 chars.

record

kchibisov:
When doing selection expansion we were checking for wide char flags
on a cells from the bottom of the terminal instead of in a current
viewport when scrolled up in history, which was leading to expanding
more than needed if we had wide chars on the same viewport cell,
but in the bottom of the terminal.

Fixes #4257.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Crash when copying/pasting with neither pointer nor keyboard focus on Wayland
 - Crash due to fd leak on Wayland
 - IME window position with fullwidth characters in the search bar
+- Selection sometimes expanding over 2 chars when scrolled up in history
 
 ## 0.5.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Crash when copying/pasting with neither pointer nor keyboard focus on Wayland
 - Crash due to fd leak on Wayland
 - IME window position with fullwidth characters in the search bar
-- Selection sometimes expanding over 2 chars when scrolled up in history
+- Selection expanding over 2 characters when scrolled in history with fullwidth characters in use
 
 ## 0.5.0
 
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -211,8 +211,7 @@ impl<'a, C> RenderableCellsIter<'a, C> {
 
         let num_cols = self.grid.cols();
 
-        // Since we can have non zero display offset, we must take points from a grid in buffer
-        // coordinates.
+        // Convert to absolute coordinates to adjust for the display offset.
         let buffer_point = self.grid.visible_to_buffer(point);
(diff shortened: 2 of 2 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -210,23 +210,29 @@ impl<'a, C> RenderableCellsIter<'a, C> {
         }
 
         let num_cols = self.grid.cols();
-        let cell = self.grid[&point];
+
+        // Since we can have non zero display offset, we must take points from a grid in buffer
+        // coordinates.
```
chrisduerr, line 215:
```suggestion
        // Convert to absolute coordinates to adjust for the display offset.
```

Just a bit shorter and straight to the point.

alacritty_terminal/src/term/mod.rs
```diff
@@ -210,23 +210,29 @@ impl<'a, C> RenderableCellsIter<'a, C> {
...
+        let buffer_point = self.grid.visible_to_buffer(point);
+        let cell = self.grid[buffer_point];
 
         // Check if wide char's spacers are selected.
         if cell.flags.contains(Flags::WIDE_CHAR) {
             let prev = point.sub(num_cols, 1);
```
chrisduerr, line 221: Why do we keep using the viewport point here, instead of using `sub_absolute`?

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Crash when copying/pasting with neither pointer nor keyboard focus on Wayland
 - Crash due to fd leak on Wayland
 - IME window position with fullwidth characters in the search bar
+- Selection sometimes expanding over 2 chars when scrolled up in history
```
chrisduerr, line 47:
```suggestion
- Selection expanding over 2 characters when scrolled in history with fullwidth characters in use
```

I don't like being so vague here, since we know exactly what the issue is.

chrisduerr (changes requested):

CHANGELOG.md
kchibisov: I thought about it, but it was really too long, but if it's fine I'll use it.

alacritty_terminal/src/term/mod.rs
kchibisov: The checks for selection are in visible coordinates. Also will `add_absolute`/`sub_absolute` clamp to viewport, I'm not sure that they will? Also, the point in `is_selected` is also in viewport, so you need to convert to viewport anyway.

CHANGELOG.md
chrisduerr: This is still under the 100 col limit we usually have for changelog entries?

alacritty_terminal/src/term/mod.rs
chrisduerr: Sure, sounds good.

chrisduerr (changes requested):

chrisduerr (approved):

PR #4279, opened 2020-10-02
Fix handling of Unicode characters in Message bar

### LINKED ISSUE #4250: Message bar text shouldn't be computed based on byte length
There are some string length checks for the message bar which use the byte length (see [here](https://github.com/alacritty/alacritty/blob/62c300c14ce7a96373e0094763e9fd035b6b9b6d/alacritty/src/display.rs#L419)). Generally I don't think unicode and fullwidth characters have been considered much for the message bar. However since it's possible for user config to show up inside the message bar, it's possible to have these unicode symbols present here.

alessandroasm:
When formating text for displaying in Message bar, allacritty was using the byte
length of the text instead of character length; this generated unnecessary blank
space at the end of lines.

Fixes #4250.

### DIFF
--- a/alacritty/src/message_bar.rs
+++ b/alacritty/src/message_bar.rs
@@ -43,12 +43,13 @@ impl Message {
         let mut lines = Vec::new();
         let mut line = String::new();
         for c in self.text.trim().chars() {
+            let line_len = line.chars().count();
             if c == '\n'
-                || line.len() == num_cols
+                || line_len == num_cols
                 // Keep space in first line for button.
                 || (lines.is_empty()
                     && num_cols >= button_len
-                    && line.len() == num_cols.saturating_sub(button_len + CLOSE_BUTTON_PADDING))
+                    && line_len == num_cols.saturating_sub(button_len + CLOSE_BUTTON_PADDING))
             {
                 // Attempt to wrap on word boundaries.
                 if let (Some(index), true) = (line.rfind(char::is_whitespace), c != '\n') {
--- a/alacritty/src/message_bar.rs
+++ b/alacritty/src/message_bar.rs
@@ -38,7 +38,7 @@ impl Message {
         let total_lines =
             (size_info.height() - 2. * size_info.padding_y()) / size_info.cell_height();
         let max_lines = (total_lines as usize).saturating_sub(MIN_FREE_LINES);
-        let button_len = CLOSE_BUTTON_TEXT.width();
+        let button_len = CLOSE_BUTTON_TEXT.chars().count();
 
         // Split line to fit the screen.
         let mut lines = Vec::new();
--- a/alacritty/src/message_bar.rs
+++ b/alacritty/src/message_bar.rs
@@ -1,7 +1,7 @@
 use std::collections::VecDeque;
 
 use alacritty_terminal::term::SizeInfo;
-use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
+use unicode_width::UnicodeWidthChar;
 
 pub const CLOSE_BUTTON_TEXT: &str = "[X]";
 const CLOSE_BUTTON_PADDING: usize = 1;
--- a/alacritty/src/message_bar.rs
+++ b/alacritty/src/message_bar.rs
@@ -1,7 +1,7 @@
 use std::collections::VecDeque;
 
-use unicode_width::{ UnicodeWidthChar, UnicodeWidthStr };
 use alacritty_terminal::term::SizeInfo;
+use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
 
 pub const CLOSE_BUTTON_TEXT: &str = "[X]";
 const CLOSE_BUTTON_PADDING: usize = 1;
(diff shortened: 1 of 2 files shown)


chrisduerr (changes requested):
This needs a changelog entry.

The `pad_text` method also has the same problem in `message_bar.rs`.

I've checked every single instance of `.len()` in Alacritty and I think that should cover all of them.

alessandroasm:
I just pushed the changes requested.

I'm glad to help :+1:

alacritty/src/message_bar.rs
```diff
@@ -110,7 +111,8 @@ impl Message {
     /// Right-pad text to fit a specific number of columns.
     #[inline]
     fn pad_text(mut text: String, num_cols: usize) -> String {
-        let padding_len = num_cols.saturating_sub(text.len());
+        let text_len = text.chars().count();
+        let padding_len = num_cols.saturating_sub(text_len);
```
chrisduerr, line 115: We can inline this.

chrisduerr (changes requested):

alacritty/src/message_bar.rs
alessandroasm: Done!

chrisduerr (changes requested):
Are you aware of how to test this? I've tested it and it seems like this does not correctly fix it. You'll likely have to use `unicode-width` instead to determine the width of the string.

alessandroasm:
You are right, I just did a more thorough test and I was missing some cases, like Japanese characters and emojis. I'll push a new commit soon using `unicode-width`.

While testing with the emojis, I found an issue with ligature rendering:
image

How it should look like:
image

This will affect the current issue, since `unicode-width` returns 4 for 👩‍🔬. I'll check more on this later.

chrisduerr:
There is no issue with ligature rendering. That's working as inteded. `unicode-width` is also working as intended and gives back the correct size of the graphemes.

alacritty/src/message_bar.rs
```diff
@@ -55,16 +56,24 @@ impl Message {
...
+                    line_len = 0;
                 }
             }
 
             if c != '\n' {
                 line.push(c);
+
+                if let Some(char_width) = c.width() {
+                    line_len += char_width;
+                }
```
chrisduerr, line 76: Should just use `.unwrap_or(0)`. Might have to use `.unwrap_or_default()` if clippy complains, but `unwrap_or(0)` is more descriptive and shorter so I think it's better in this case.

alacritty/src/message_bar.rs
```diff
@@ -341,6 +350,23 @@ mod tests {
...
+
+        let lines = message_buffer.message().unwrap().text(&size);
+
+        assert_eq!(lines, vec![
+            String::from("ab  [X]"),
+            // the "e" would fit in this line but 👩 width is 2
+            String::from("c 👩d  "),
+            String::from("e fgh  ")
+        ]);
+    }
```
chrisduerr, line 368: Very good idea to test this, thanks for adding one.

alacritty/src/message_bar.rs
```diff
@@ -341,6 +350,23 @@ mod tests {
...
+        let input = "ab\nc 👩d e fgh";
+        let mut message_buffer = MessageBuffer::new();
+        message_buffer.push(Message::new(input.into(), MessageType::Error));
+        let size = SizeInfo::new(7., 10., 1., 1., 0., 0., false);
+
+        let lines = message_buffer.message().unwrap().text(&size);
+
+        assert_eq!(lines, vec![
+            String::from("ab  [X]"),
+            // the "e" would fit in this line but 👩 width is 2
```
chrisduerr, line 364: I don't think it's necessary to document this. It should be very clear because your editor shows you exactly how big it is. It's also a bit incorrect, because it wouldn't fit and it doesn't because this isn't just a single width character.

chrisduerr (changes requested):

alessandroasm:
I just pushed all the changes suggested.

Thanks for reviewing all those changes so quickly!

chrisduerr (approved):
Since I've noticed some more necessary changes, I've decided to look into it myself and have made some minor tweaks to this patch. All changes should be explained in my commit message.

I think this should be good to go now, thanks for the PR.

alacritty/src/message_bar.rs
```diff
@@ -1,6 +1,7 @@
 use std::collections::VecDeque;
 
 use alacritty_terminal::term::SizeInfo;
+use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
```
kchibisov, line 4:
```suggestion
use unicode_width::UnicodeWidthChar;
```

kchibisov (changes requested):

kchibisov (approved):

alessandroasm:
@chrisduerr Hi! If it's not much trouble could you tag this PR with "hacktoberfest-accepted"?
Thanks in advance!

chrisduerr:
I'd rather not add additional labels to this repository. Though you can make pull requests to your own repositories, which is probably easier and doesn't mean any burden for us.

Issue #4157, opened 2020-08-21
Support Path/Env Expansion in Config File(s)


nixpulvis:
Now that #4138 has landed, the need to write things like `~/.config/alacritty/colors.yml` (or `$HOME`) is desired. Ideally this expansion would happen for all other paths (or strings) in the config file.

Furthermore, for paths, it would be nice to allow glob imports like `~/.config/alacritty/bindings/*.yml`, which could allow easy organization of various settings from a directory. However, this feature is much less needed than simple `~` or ENV variables.

gempir:
I think it should not work with environment values, because there are to many potential problems there and confusing behaviour to the user.

But there should be a way to reference the home directory. Using `~` would probably least confusing for users and what most would expect to work.

EdenEast:
Paths beginning with `~` or `~/` depending on if you want to support the `~undeadleech/alacritty.yml` case should be
expanded to HOME based:

- `HOME` env variable (they might have defined something different then standard)
- Fallback to platforms define defaults: (`dirs::home_dir()`)
  - Linux:   `/home/{username}`
  - Mac:     `/Users/{username}`
  - Windows: `C:\Users\{Username}`

If a user like me that works on multiple machine and multiple platforms (linux, windows), this would allow for one
alacritty config to be in all cases. The documentation for import would explain to user the order of expansion for `~`.

EdenEast:
Another solution to @nixpulvis problem would be to support relative paths for the import section. Paths would be relative to the file they are defined in. 

For example, you are using the default configuration location `XDG_CONFIG_HOME/alacritty`. You have broken up your configuration into categories. You directory structure might look something like this:

```
.config/alacritty ➜ tree .
.
|-- alacritty.yml
|-- config
    |-- functional.yml
    |-- input.yml
    |-- theme.yml
    `-- visual.yml
````

In your entry point of your configuration your import section would look something like this:

```yml
imports:
  - config/functional.yml
  - config/input.yml
  - config/theme.yml
  - config/visual.yml
```

As these are relative paths they are relative to the entry file that they were defined in (`~/.config/alacritity/alacritty.yml`) in this example.

@nixpulvis would this help in the mean time while Path/Env expansion is figured out. @chrisduerr your thoughts as well? Also would this be a separate issue?

chrisduerr:
> @nixpulvis would this help in the mean time while Path/Env expansion is figured out. @chrisduerr your thoughts as well? Also would this be a separate issue?

[...]

PR #4299, opened 2020-10-09
Spawn new alacritty process in CWD on macOS

### LINKED ISSUE #808: Open new window in same directory
In other terminals, this is commonly bound to Ctrl-Shift-T.

Possibly a duplicate of https://github.com/jwilm/alacritty/issues/607 but I'm not certain.

### LINKED ISSUE #1979: Spawn new Alacritty instance in working directory on macOS using proc_pidinfo

nixpulvis:
fixes #1979
fixes #808 (for macOS)


alacritty/Cargo.toml
```diff
@@ -42,6 +42,7 @@ image = { version = "0.23.3", default-features = false, features = ["ico"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 objc = "0.2.2"
+darwin-libproc = "0.2"
```
chrisduerr, line 45: My biggest concern with this PR are the two new dependencies pulled in here. Unlike libc, these are basically unused and seem to be passively maintained by just a single person. So I think if we could circumvent these it would be beneficial.

chrisduerr (changes requested):

alacritty/Cargo.toml
nixpulvis: @chrisduerr I'm extremely conflicted about this personally...

Before I responded I had a bit of a walk through of the code inside `darwin-libproc`, and found that they aren't building the binding code in a build script, but instead using a fixed run of `bindgen`. So either way I've opened heim-rs/darwin-libproc#2. So I guess I'll see how that goes...

Assuming there's not a horrible issue with that process, I feel like working with the community here is better than writing our own logic, even if it's for a very small application. I personally don't want to maintain this logic very much, and I expect the users of `darwin-libproc` and `heim` in general will spot issues in addition to ourselves. There does seem to be some macOS version dependence, but I'm not sure why we'd want to handle that ourselves any differently than anyone else. I'm also not sure if there's any versions of macOS we care about without the features we need.

Having read over the implementation of [`pid_info`](https://github.com/heim-rs/darwin-libproc/blob/dcc1a4ba7d22164cf7e1656a64e814440bbe4742/darwin-libproc/src/pid_info.rs#L4-L30) now, I'm not sure if there's anything I would do differently. I couldn't find any documentation from Apple for `libproc` past the header itself really, so I just have to go on sane assumptions and the testing I did locally on my one Mac laptop I have lying around.

Everything seems to work, so I'm not sure if this should be merged, or if we should avoid this issue altogether. But that's kinda where I'm at right now.

chrisduerr: > I feel like working with the community here is better than writing our own logic

There is no "community". It's just some random person I do not know or trust at all.

> I expect the users of darwin-libproc and heim in general will spot issues

It seems like that has already failed, with them using `bindgen` improperly.

> I'm also not sure if there's any versions of macOS we care about without the features we need.

That should be very clear and decided upon independent from this issue. Currently at least 10.11+ should be supported.

nixpulvis: I'd like to wait to see what they say about my PR for at least a couple days. If there's no response, or I'm unsatisfied, we can move on. I'm curious if you'd rather see a hand written binding, or our own use of `bindgen` though?

> That should be very clear and decided upon independent from this issue. Currently at least 10.11+ should be supported.

Seems like the value for the "flavor" argument of this function call hasn't changed since [10.11](https://git.privacyone.io/useful/macos-sdk/-/blob/master/MacOSX10.11.sdk/usr/include/sys/proc_info.h), since it's still 9 on my machine.

chrisduerr: > I'm curious if you'd rather see a hand written binding, or our own use of bindgen though?

I don't care particularly much about this detail. I feel like for a single function a handwritten one is probably going to be easier?

nixpulvis: > I feel like for a single function a handwritten one is probably going to be easier?

I don't think so... `bindgen` works **very** well in my experience, and makes all the needed values exposed in a reasonable fashion. Whereas otherwise, I'd need to define a number of constants myself, and ensure they stay consistent with the header.

chrisduerr: The thing is all we need is the single function, nothing else.

nixpulvis: Unused code optimizations should make it so we're not bloating our binaries... unless something is wrong with that, I don't see why I should have to maintain these bindings in a more cumbersome way. Fact is, single function or not, `bindgen` is the best way to do this I believe.

chrisduerr: Pulling in a whole swath of code for just a single function is a bit ridiculous and just unnecessarily complicated. Shoving this work on the code optimizer but having to deal with constantly updating a load of completely unnecessary functionality is not productive.

nixpulvis: I mean, we already require `bindgen` for `winpty-sys` so I see no harm reusing it here.

nixpulvis: I suppose technically it's optional. But then again, this is only required on macOS, so it's somewhat similar. Perhaps I need to fix the current state of the `Cargo.lock`?

chrisduerr: > I mean, we already require bindgen for winpty-sys so I see no harm reusing it here.

You mean the thing that is getting removed in ~10 weeks? My problem is not with a dependency on bindgen, it's just that I don't see how bindgen makes anything easier for us here. It's just a single function definition without a dependency on any other types.

I'm not sure if it might require one or two additional calls for interpreting the `proc_pidinfo` call, but it certainly isn't extensive.

nixpulvis: I'm mostly talking about the `#define PROC_PIDVNODEPATHINFO		9` definition. Perhaps adding our own constant isn't the end of the world, but I still feel like I'd rather use `bindgen`.

chrisduerr: A single constant certainly shouldn't be the end of the world.

nixpulvis: OK, I'll look into making a `macos` specific module in `alacritty` for this definition.

chrisduerr: Sure. Fyi we already do this for Windows PTY stuff. So it's certainly not the first time.

nixpulvis:
Updated the code to inline everything to `alacritty/macos/proc.rs`.

alacritty/src/main.rs
```diff
@@ -36,8 +36,11 @@ mod daemon;
 mod display;
 mod event;
 mod input;
+// TODO: Move into macos.
 #[cfg(target_os = "macos")]
 mod locale;
```
nixpulvis, line 41: I'm happy to do this now, or immediately (before 0.6) in a follow-up PR.

chrisduerr: My policy is to always allow style improvements in PRs. If we can make our code better by restructuring things a bit, that sounds good to me, no need to be overly pedantic about separation there.

So feel free to go ahead and move this into the macos module.

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,104 @@
...
+
+extern "C" {
+    pub fn proc_pidinfo(
+        pid: raw::c_int,
+        flavor: raw::c_int,
+        arg: u64,
+        buffer: *mut raw::c_void,
+        buffersize: raw::c_int,
+    ) -> raw::c_int;
+}
```
chrisduerr, line 104: This might be just me, but I feel like "hollow" API definitions like these are better fit at the top of the file. So once everything is defined, you can then use it.

There is some precedence with the definition of the ConptyApi in the windows backend. So I think it should be good to do so.

I'd assume your idea was to "hide away" the things most people shouldn't care much about?

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,104 @@
...
+use std::io;
+use std::mem;
+use std::os::raw;
+use std::os::unix::ffi::OsStrExt;
+use std::path::PathBuf;
+use std::slice;
+use memchr::memchr;
+
+pub fn cwd(pid: raw::c_int) -> io::Result<PathBuf> {
+    let flavor = PROC_PIDVNODEPATHINFO as raw::c_int;
```
chrisduerr, line 11: Can you not just define `PROC_PIDVNODEPATHINFO` as a c_int?

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,104 @@
...
+    let mut info = mem::MaybeUninit::<proc_vnodepathinfo>::uninit();
+    let size = mem::size_of::<proc_vnodepathinfo>() as raw::c_int;
+    let result = unsafe {
+        proc_pidinfo(pid, flavor, 0, info.as_mut_ptr() as *mut raw::c_void, size)
+    };
+    match result {
+        c if c <= 0    => return Err(io::Error::last_os_error()),
```
chrisduerr, line 18: Are you sure that `0` is an error value?

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,104 @@
...
+        s if s != size => return Err(io::Error::new(io::ErrorKind::Other, "invalid return size")),
+        _ => {}
```
chrisduerr, line 20:
```suggestion
        _ => (),
```

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+    let raw_path = unsafe {
+        let info = info.assume_init();
+        slice::from_raw_parts(info.pvi_cdir.vip_path.as_ptr() as *const u8,
+                              info.pvi_cdir.vip_path.len())
+    };
+    let null_idx = memchr(0x00, &raw_path).unwrap_or(0);
```
chrisduerr, line 27:
```suggestion
    let null_idx = memchr(0, &raw_path).unwrap_or(0);
```

chrisduerr (changes requested):

alacritty/src/macos/proc.rs
nixpulvis: > I'd assume your idea was to "hide away" the things most people shouldn't care much about?

Yea, since I was basically only wanting to expose `cwd`. Perhaps I should just make everything else in here not `pub`. Not sure why I didn't already do that TBH.

alacritty/src/macos/proc.rs
nixpulvis: How about `'\0'`?

chrisduerr: I'd be fine with '\0', that would make it pretty clear what's going on.

Cargo.lock
```diff
@@ -37,6 +37,7 @@ dependencies = [
  "image",
  "libc",
  "log",
+ "memchr",
```
nixpulvis, line 40: hmmm, IDK why this didn't go away with my last commit. I'll update is a sec.

alacritty/src/macos/proc.rs
nixpulvis: Sure could! I started with the output of `bindgen` and didn't catch all the transforms as you can tell :wink:

alacritty/src/macos/proc.rs
nixpulvis: Good question. I am quite confident it is *some* kind of error, since we simply cannot build a `proc_vnodepathinfo` without some bytes returned. That being said, I don't know for sure if it's reliable to use `last_os_error()` here, or if the other error makes more sense.

I tried looking at some code I found for the implementation of `proc_pidinfo` online, but it's not consistent with the interface we're using (it's also horrible code). Perhaps I can find the latest version somewhere...

alacritty/src/macos/proc.rs
chrisduerr: So stop casting it then?

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,99 @@
+use std::ffi::{CString, CStr};
+use std::io;
+use std::mem;
+use std::os::raw;
+use std::path::PathBuf;
+
+pub fn cwd(pid: raw::c_int) -> io::Result<PathBuf> {
+    let flavor = PROC_PIDVNODEPATHINFO as raw::c_int;
+    let mut info = mem::MaybeUninit::<proc_vnodepathinfo>::uninit();
```
chrisduerr, line 9:
```suggestion
    let mut info = MaybeUninit::<proc_vnodepathinfo>::uninit();
```

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,99 @@
...
+            s if s != size => return Err(io::Error::new(io::ErrorKind::Other, "invalid return size")),
+            _ => {}
+        };
+        CStr::from_ptr(info.assume_init().pvi_cdir.vip_path.as_ptr())
+    };
+
+    match CString::from(c_str).into_string() {
+        Err(e) => Err(io::Error::new(io::ErrorKind::Other, e)),
+        Ok(s) => Ok(PathBuf::from(s)),
+    }
```
chrisduerr, line 24:
```suggestion
    CString::from(c_str).into_string()
        .map(PathBuf::from)
        .map_err(|err| IOError::new(IOErrorKind::Other, err));
    }
```

I think both would be fine here, but I feel like since we just remap both it might be nicer to do in by mapping?

alacritty/src/main.rs
```diff
@@ -96,7 +96,7 @@ fn main() {
     env::set_current_dir(dirs::home_dir().unwrap()).unwrap();
     // Set locale.
     #[cfg(target_os = "macos")]
-    locale::set_locale_environment();
+    macos::locale::set_locale_environment();
```
chrisduerr, line 99: These imports are too deep, just local should be sufficient.

chrisduerr (changes requested):

alacritty/src/event.rs
```diff
@@ -4,7 +4,7 @@ use std::borrow::Cow;
 use std::cmp::{max, min};
 use std::env;
 use std::fmt::Debug;
-#[cfg(unix)]
+#[cfg(all(unix, not(target_os = "macos")))]
```
kchibisov, line 7: For such cases we use `#[cfg(not(any(target_os = "macos", windows)))]`

CHANGELOG.md
```diff
@@ -67,6 +67,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `Add` and `Subtract` keys are now named `NumpadAdd` and `NumpadSubtract` respectively
 - Feature checking when cross compiling between different operating systems
 - Crash when writing to the clipboard fails on Wayland
+- `SpawnNewInstance` will now remain in the same working directory on macOS
```
kchibisov, line 70: I don't really like the way it's written here.

Maybe something like
```
`SpawnNewInstance` not using working directory of active process on macOS
```

kchibisov (changes requested):

CHANGELOG.md
chrisduerr: The way you rephrased it is a bugfix, which it isn't. This should actually be in the changed or added section.

Besides that, "will now remain" is a bit misleading, since it doesn't really remain anywhere. It changes into that directory.

alacritty/src/macos/proc.rs
nixpulvis: :sweat_smile:

CHANGELOG.md
nixpulvis: I put the change log entry in the "Fixed" section because it's clearly stated behaviors wasn't being respected in from the configuration:

https://github.com/alacritty/alacritty/blob/80688e082d13197b6661cfb6cfa60734034270cf/alacritty.yml#L375-L379

That said, I'll move this to the changed section, since that's applicable as well, and the remaining open issue (#1979) is marked as an enhancement.

chrisduerr: Thanks for the clarification, but at that point I'd say it's more of a bug in our documentation than anything else. This definitely was planned to work exactly the way it does right now on macOS. Generally unless it's actually something that should be behaving differently, I wouldn't consider it a bug.

Having things fall in the `Fixed` section is also usually just the "goto" place for everything that isn't a new feature. So if there's any reason to move it somewhere else it's probably a good idea to do so.

alacritty/src/main.rs
```diff
@@ -28,6 +28,9 @@ use alacritty_terminal::sync::FairMutex;
 use alacritty_terminal::term::Term;
 use alacritty_terminal::tty;
 
+#[cfg(target_os = "macos")]
+use crate::macos::locale;
```
chrisduerr, line 32: Imports from modules inside of Alacritty are usually *below* those module imports.

You can see that in this file already.

chrisduerr (changes requested):

alacritty/src/main.rs
nixpulvis: I'll move it, I just generally like to keep all my imports together.

nixpulvis:
So I have a question about the expected interaction between this feature of `SpawnNewInstance` and our `working_directory` configuration. I think I was expecting that configuration on the parent to override this behavior, correctly setting the value to it's own configuration's default. But I understand that other terminal emulators provide an option for this case, for example `spawn_new_working_directory: Default | Current`, but I'm not convinced there isn't a better way to do this.

Furthermore, I don't think it would be fair to claim that the "directory of the parent process will be used", since launching `alacritty` from `xterm`, for example, puts me in our default working directory, not following the logic in `spawn_new_instance` as described.

I'm totally open to the idea of a new or modified configuration here, but first, how would others expect this to work? Can we unify the logic in `spawn_new_instance` in some way?

chrisduerr:
> So I have a question about the expected interaction between this feature of SpawnNewInstance and our working_directory configuration. I think I was expecting that configuration on the parent to override this behavior, correctly setting the value to it's own configuration's default. But I understand that other terminal emulators provide an option for this case, for example spawn_new_working_directory: Default | Current, but I'm not convinced there isn't a better way to do this.

I don't see a need to preserve the original working directory. We don't do this for no CLI parameter, so I don't see why we would do this for a CLI overridden one.

> Furthermore, I don't think it would be fair to claim that the "directory of the parent process will be used", since launching alacritty from xterm, for example, puts me in our default working directory, not following the logic in spawn_new_instance as described.

I don't see what you mean? What does SpawnNewInstance have to do with launching things from XTerm?

> I'm totally open to the idea of a new or modified configuration here, but first, how would others expect this to work? Can we unify the logic in spawn_new_instance in some way?

I don't see why we would change **anything** here, that discussion seems especially out of place for this PR. If anything I'd create a new issue, but I don't see what you would like to do.

nixpulvis:
> I don't see what you mean? What does SpawnNewInstance have to do with launching things from XTerm?

Ignore my statement about `xterm` I was getting things a bit mixed up.

https://github.com/alacritty/alacritty/blob/80688e082d13197b6661cfb6cfa60734034270cf/alacritty.yml#L375-L379

I'm reading the documentation for `working_directory` (linked again), and wondering how the behavior should work in the case it's `None` or unset. Further I would like to understand if we can make this consistent with `SpawnNewInstance` moving forward.

1. The documentation today states that "the working directory of the parent process will be used" which is not the case on macOS as you can see in the culmination of #1007. This is because running Alacritty.app would otherwise start in `/`. It would be really nice to be able to set this path from the `Info.plist` (or somewhere in the app bundle), so that launching `alacritty` from the CLI works just as it does on other platforms. After some searching around online however, I wasn't able to figure out a way macOS lets you do this. Really macOS?!

2. It stands to reason, if I set my configuration's `working_directory` I might want that to be the same for `SpawnNewInstance` calls as well. This is why, even ignoring macOS, I would like to see these behaviors unified in some way now. Basically the working directory would always be set with the follow precedence for both external executions and `SpawnNewInstance` actions: CLI option (`--working-directory ...` / `-o working_directory ...`) > Configuration (`working_directory: ...`) > Parent's CWD.

Points 1 and 2 here are related, since today on macOS the initial launch of Alacritty.app behaves as if `working_directory: $HOME` was set. However, it's not currently possible to set environment variables from the configuration file (see #4157). Assuming we could set `working_directory: ~` to replace the hard coded working directory for macOS, this still does not help the user who would like the initial launch of Alacritty.app to start in `$HOME` but future calls to `SpawnNewInstance` to retain the CWD.

Basically, things are a bit confusing currently, and I would like to see how we can make the way launching Alacritty works across platforms/actions more consistent.

chrisduerr:
> The documentation today states that "the working directory of the parent process will be used" which is not the case on macOS as you can see in the culmination of #1007.

That's solely a macOS system problem and I do not care about that at all. If someone wants to fix that, talk to Apple.

> It stands to reason, if I set my configuration's working_directory I might want that to be the same for SpawnNewInstance calls as well.

**No.** The entire reason why `SpawnNewInstance` exists at all is to persist the working directory. This will absolutely **not** be changed.

> Basically, things are a bit confusing currently

I don't see how they're confusing at all. And no matter what, nothing is going to be changed about the behavior in this PR.

alacritty/src/macos/proc.rs
nixpulvis: I'm going to move the 0 case to the `ErrorKind::Other` since I'm not at all convinced `last_os_error` will return anything meaningful, and "invalid return size" is absolutly a correct assessment of this error.

nixpulvis:
It's probably not worth noting anywhere else, since macOS users should hopefully know how launching Alacritty works on their platform, however since I needed to remember how this all worked...

Here is a configuration to override the default <kbd>⌘</kbd> + <kbd>N</kbd> binding to the old behavior, where `SpawnNewInstance` always opens in the default directory:

```yaml
key_bindings:
  - { key: N, mods: Command, command: { program: "open", args: ["-na", "Alacritty.app"] }}
```

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,119 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_cwd() {
+        let pid = std::process::id();
+        cwd(pid as i32)
+            .map(|path| assert!(!path.to_str().expect("UTF-8 encoding").is_empty()))
```
chrisduerr, line 42: Performing an assertion inline like this I would usually consider pretty bad practice. Generally `.map` should not be used as a `.for_each` kinda thing, but instead only be used when the goal is to transform one value into another.

So just doing `let cwd = cwd(pid).expect(...); assert!` should be much better.

It's also possible to just get the current working directory of the test process itself using `std::env::current_dir`, so I feel like we should be able to write a much better test with that instead? Then the whole expect stuff wouldn't even be necessary.

chrisduerr (changes requested):

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,119 @@
+#![allow(non_camel_case_types)]
```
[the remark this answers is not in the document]
nixpulvis: Oh, I would be happy to do that. I'm not sure where that comment was, sorry I don't recall it.

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,119 @@
...
+    let mut info = MaybeUninit::<proc_vnodepathinfo>::uninit();
+    let size = mem::size_of::<proc_vnodepathinfo>() as raw::c_int;
+    let c_str = unsafe {
+        match proc_pidinfo(
+            pid,
+            PROC_PIDVNODEPATHINFO,
+            0,
+            info.as_mut_ptr() as *mut raw::c_void,
+            size,
+        ) {
```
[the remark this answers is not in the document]
nixpulvis: Yea I can understand that. I'll update this.

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,119 @@
...
+    use super::*;
+
+    #[test]
+    fn test_cwd() {
+        let pid = std::process::id();
+        cwd(pid as i32)
+            .map(|path| assert!(!path.to_str().expect("UTF-8 encoding").is_empty()))
+            .expect("pid must be accessible");
+    }
+}
```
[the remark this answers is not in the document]
nixpulvis: Yea, I was just treating the rest of the code as a separate file in my head, and given that I really only plan to add this one test I figured I'd put it here. But consistency with everything else makes sense.

alacritty/src/macos/proc.rs
nixpulvis: Sadly `std::io::Error` doesn't implement `==`.
https://users.rust-lang.org/t/help-understanding-io-error-and-lack-of-partialeq/13212/9

alacritty/src/macos/proc.rs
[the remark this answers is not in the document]
chrisduerr: That's all there was to it. So just move it into a module so it's a bit "separated".

alacritty/src/macos/proc.rs
chrisduerr: It should never error anyways, right? Might as well just map it or something. Though unwrapping is perfectly fine in tests too if it generates the simpler code. The errors for unwraps in test code are extremely easy to debug, so much so that I probably wouldn't even bother with `expect` unless there are multiple errors on the same line.

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Use yellow/red from the config for error and warning messages instead of fixed colors
 - Existing CLI parameters are now passed to instances spawned using `SpawnNewInstance`
 - Wayland's Client side decorations now use the search bar colors
+- `SpawnNewInstance` now starts in the same working directory on macOS
```
kchibisov, line 39: This still has strange message, I'd suggest to go through the previous comments wrt that entry, and update it.

kchibisov (changes requested):

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,117 @@
+use std::ffi::{CStr, CString};
+use std::io;
+use std::mem::{self, MaybeUninit};
+use std::os::raw;
+use std::path::PathBuf;
+
+pub fn cwd(pid: raw::c_int) -> io::Result<PathBuf> {
```
chrisduerr, line 7: I've been considering this for a while and I think this definitely is something that should be discussed:

The return value of `io::Result` seems kinda misleading here. We have multiple places where we us errors that are clearly not io errors and turn them into `io::Errorkind::Other` just so we can return them here.

The only reason for doing this as far as I can tell would be laziness, since it's just "easier" to use `io::Error` for everything rather than providing separate errors depending on failure case. I think we should probably have higher standards than this.

Either we want to offer good errors and define a custom error enum that implements `From<io::Error>`, or we just say the error doesn't matter at all and we return an `Option<PathBuf>`. But the "excessive" use of `io::ErrorKind::Other` here is kinda rubbing me the wrong way.

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,117 @@
...
+        }
+    };
+
+    CString::from(c_str)
+        .into_string()
+        .map(PathBuf::from)
+        .map_err(|e| io::Error::new(io::ErrorKind::Other, e))
+}
+
+// Cleaned up `bindgen` output below
```
chrisduerr, line 29: What purpose does this comment have? If it actually is useful, which I cannot see at the moment, we should add punctuation at least.

If we actually want to write useful documentation, this should probably be a doc comment explaining that this is a binding to the macos proc system API. Generally internal doc comments are preferred in that case.

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,117 @@
...
+            buffersize: raw::c_int,
+        ) -> raw::c_int;
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::env;
+    use std::process;
```
chrisduerr, line 111:
```suggestion
    use super::*;
    
    use std::env;
    use std::process;
```

alacritty/src/macos/proc.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    #[test]
+    fn test_cwd() {
```
chrisduerr, line 114: We don't need to call the test a test in the method name. Either we call this just `fn cwd` or we actually describe what about `cwd` it tests. Like `cwd_matches_std` or something like that.

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Wayland's Client side decorations now use the search bar colors
 - Reduce memory usage by up to at least 30% with a full scrollback buffer
 - The number of zerowidth characters per cell is no longer limited to 5
+- `SpawnNewInstance` is now using working directory of active process on macOS
```
chrisduerr, line 41:
```suggestion
- `SpawnNewInstance` is now using the working directory of the terminal foreground process on macOS
```

"Active process" is honestly kind of a poor name for it. I know that the other changelog entry uses this wording already but that's just a bit misleading. The `terminal` is kinda optional, but I think it's a useful addition.

I'm open to changing the other changelog entry too.

chrisduerr (changes requested):

alacritty/src/macos/proc.rs
kchibisov: Well, the point is to avoid name collision afaics, but I can rename test.

chrisduerr: It's a separate module, there is no collision.

kchibisov: No, the function from top module is called `cwd`, then they do `super::*`, which brings cwd, here, and then you call a test `cwd`, there's a collision due to the way it's used. So to avoid that you should either rename test or use full path for function, like `super::cwd`.

alacritty/src/macos/proc.rs
kchibisov: We return last_os_error in some cases though, so the usage of os error seems legit?

alacritty/src/macos/proc.rs
nixpulvis: I'd be happy to expand this comment to include more information about the bindgen process, mainly the command used to create the seed of this file itself. Given that I needed to strip a lot of things manually, I figured this was enough however.

alacritty/src/macos/proc.rs
nixpulvis: yea IIRC I started with this in the same module. Renaming the test just `cwd` would even be fine with me.

alacritty/src/macos/proc.rs
kchibisov: I don't think that we should add more docs about a function we're using, we should just refer from where we're using it, since they should have docs already and repeating them doesn't give any benefit.

alacritty/src/macos/proc.rs
chrisduerr: Just to be clear, even when `super::*` is used, there is no problem with conflicts. But of course @nixpulvis suggestion is not unlikely, if it wasn't in a separate module it would be problematic.

alacritty/src/macos/proc.rs
chrisduerr: I'm not sure why the bindgen process needs to be documented at all. As mentioned earlier how this code is generated should be irrelevant.

If anything, the module itself should be documented, not what environment/editor/generator you used to create it.

alacritty/src/macos/proc.rs
chrisduerr: Always returning `io::Error` is only legit if all your errors are IO errors. Some errors being io errors doesn't justify converting all your errors to io errors.

kchibisov: Well, most of your errors are IO errors, and you have one case when you decided to fallback. Converting originally IO errors to some other errors sounds stranger than using `io::Error`.

kchibisov: Though, I can remove one io error that doesn't make sense.

chrisduerr: You're not converting IO errors to anything. This is the same standard error handling we do everywhere else:

https://github.com/alacritty/alacritty/blob/539855cdb92c44dc0f755e6ea6debcf13a77f049/alacritty/src/config/mod.rs#L38-L50

CHANGELOG.md
```diff
@@ -30,15 +30,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
-- Use working directory of active process instead of shell for SpawnNewInstance action
 - Fallback to normal underline for unsupported underline types in `CSI 4 : ? m` escapes
 - The user's background color is now used as the foreground for the render timer
 - Use yellow/red from the config for error and warning messages instead of fixed colors
 - Existing CLI parameters are now passed to instances spawned using `SpawnNewInstance`
 - Wayland's Client side decorations now use the search bar colors
 - Reduce memory usage by up to at least 30% with a full scrollback buffer
 - The number of zerowidth characters per cell is no longer limited to 5
-- `SpawnNewInstance` is now using working directory of active process on macOS
+- `SpawnNewInstance` is now using working directory of the terminal foreground process on macOS/Linux
```
chrisduerr, line 40: Adding the /Linux to the end made things go beyond our 100 character limit per line by 1.

It's also not entirely correct, since it applies to BSD too.

alacritty/src/macos/proc.rs
```diff
@@ -1,41 +1,84 @@
...
+            Error::Io(err) => err.source(),
+            Error::IntoString(err) => err.source(),
+        }
+    }
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::InvalidSize => write!(f, "Invalid proc_pidinfo return size."),
```
chrisduerr, line 34:
```suggestion
            Error::InvalidSize => write!(f, "Invalid proc_pidinfo return size"),
```

alacritty/src/macos/proc.rs
```diff
@@ -1,41 +1,84 @@
...
 
-    CString::from(c_str)
-        .into_string()
-        .map(PathBuf::from)
-        .map_err(|e| io::Error::new(io::ErrorKind::Other, e))
+    Ok(CString::from(c_str).into_string().map(PathBuf::from)?)
 }
 
-// Cleaned up `bindgen` output below
+/// Bindings for libproc `proc_pidinfo` function.
```
chrisduerr, line 72:
```suggestion
/// Bindings for libproc.
```

While currently our goal is just the one function, they certainly provide more than just that.

alacritty/src/macos/proc.rs
```diff
@@ -1,41 +1,84 @@
...
+use std::ffi::{CStr, CString, IntoStringError};
+use std::fmt::{self, Display, Formatter};
 use std::io;
 use std::mem::{self, MaybeUninit};
-use std::os::raw;
+use std::os::raw::{c_int, c_void};
 use std::path::PathBuf;
 
-pub fn cwd(pid: raw::c_int) -> io::Result<PathBuf> {
+/// Error occuring during getting cwd from `proc_pidinfo`.
```
chrisduerr, line 8: Our errors shouldn't leak implementation details like the API we're using to get the result we want. In theory since this is technically `proc::Error`, this shouldn't even be specific to the `cwd` function. But I'm willing to make that concession for now.

```suggestion
/// Error during working directory retrieval.
```

alacritty/src/macos/proc.rs
```diff
@@ -1,41 +1,84 @@
...
+#[derive(Debug)]
+pub enum Error {
+    /// IO error using `proc_pidinfo`.
```
chrisduerr, line 11: Probably best to just say this is an IO Error. At which point the comment becomes rather useless.

alacritty/src/macos/proc.rs
```diff
@@ -1,41 +1,84 @@
...
+    Io(io::Error),
+
+    /// Error converting into utf8 string.
+    IntoString(IntoStringError),
+
+    /// Invalid size.
```
chrisduerr, line 17: If we add a comment, it should be better at describing the actual problem:

```suggestion
    /// Expected struct size did not match libproc's.
```

The use of the word `struct` could be contested, just using `return size` would be fine too I think.

chrisduerr (changes requested):

CHANGELOG.md
```diff
@@ -30,14 +30,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - Cursors are now inverted when their fixed color is similar to the cell's background
+- Use working directory of the terminal foreground process instead of shell for SpawnNewInstance action
```
```suggestion
- Use the working directory of the terminal foreground process, instead of the shell's working
    directory, for the `SpawnNewInstance` action
```

CHANGELOG.md
```diff
@@ -30,14 +30,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 - Fallback to normal underline for unsupported underline types in `CSI 4 : ? m` escapes
 - The user's background color is now used as the foreground for the render timer
 - Use yellow/red from the config for error and warning messages instead of fixed colors
 - Existing CLI parameters are now passed to instances spawned using `SpawnNewInstance`
 - Wayland's Client side decorations now use the search bar colors
 - Reduce memory usage by up to at least 30% with a full scrollback buffer
 - The number of zerowidth characters per cell is no longer limited to 5
-- `SpawnNewInstance` is now using working directory of the terminal foreground process on macOS/Linux
+- `SpawnNewInstance` is now using working directory of the terminal foreground process on macOS
```
chrisduerr, line 41:
```suggestion
 - `SpawnNewInstance` is now using the working directory of the terminal foreground process on macOS
```

chrisduerr (changes requested):

kchibisov (approved):

chrisduerr (approved):

PR #4302, opened 2020-10-09
Make use of glutin wayland/x11 features

kchibisov:
This should allow users that are not using Wayland/X11
to reduce the resulted binary size and compilation times.

This honestly looks nicer than I expected, so I feel like it should be fine to merge it.

For reference striped binary sizes:
```
x11: 4.6M
wayland: 5.9M
both: 6.4M
```

### DIFF
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -4,7 +4,7 @@
 use std::cmp::min;
 use std::f64;
 use std::fmt::{self, Formatter};
-#[cfg(not(any(target_os = "macos", windows)))]
+#[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
 use std::sync::atomic::Ordering;
 use std::time::Instant;
 
@@ -17,7 +17,7 @@ use glutin::window::CursorIcon;
 use log::{debug, info};
 use parking_lot::MutexGuard;
 use unicode_width::UnicodeWidthChar;
-#[cfg(not(any(target_os = "macos", windows)))]
+#[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
 use wayland_client::{Display as WaylandDisplay, EventQueue};
 
 #[cfg(target_os = "macos")]
@@ -154,14 +154,15 @@ pub struct Display {
     /// Currently highlighted URL.
     pub highlighted_url: Option<Url>,
 
-    #[cfg(not(any(target_os = "macos", windows)))]
+    #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
     pub wayland_event_queue: Option<EventQueue>,
+
+    #[cfg(not(any(target_os = "macos", windows)))]
+    pub is_x11: bool,
 
     renderer: QuadRenderer,
     glyph_cache: GlyphCache,
     meter: Meter,
-    #[cfg(not(any(target_os = "macos", windows)))]
-    is_x11: bool,
 }
 
 impl Display {
@@ -184,11 +185,11 @@ impl Display {
         debug!("Estimated window size: {:?}", estimated_size);
         debug!("Estimated cell size: {} x {}", cell_width, cell_height);
 
-        #[cfg(not(any(target_os = "macos", windows)))]
+        #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
         let mut wayland_event_queue = None;
 
         // Initialize Wayland event queue, to handle Wayland callbacks.
-        #[cfg(not(any(target_os = "macos", windows)))]
+        #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
         if let Some(display) = event_loop.wayland_display() {
             let display = unsafe { WaylandDisplay::from_external_display(display as _) };
             wayland_event_queue = Some(display.create_event_queue());
@@ -199,7 +200,7 @@ impl Display {
             event_loop,
             &config,
             estimated_size,
-            #[cfg(not(any(target_os = "macos", windows)))]
+            #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
             wayland_event_queue.as_ref(),
         )?;
 
@@ -253,8 +254,10 @@ impl Display {
         #[cfg(target_os = "macos")]
         set_font_smoothing(config.ui_config.font.use_thin_strokes());
 
-        #[cfg(not(any(target_os = "macos", windows)))]
+        #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
         let is_x11 = event_loop.is_x11();
+        #[cfg(not(any(feature = "x11", target_os = "macos", windows)))]
+        let is_x11 = false;
 
         // On Wayland we can safely ignore this call, since the window isn't visible until you
         // actually draw something into it and commit those changes.
@@ -296,7 +299,7 @@ impl Display {
             highlighted_url: None,
             #[cfg(not(any(target_os = "macos", windows)))]
             is_x11,
-            #[cfg(not(any(target_os = "macos", windows)))]
+            #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
             wayland_event_queue,
         })
     }
@@ -582,12 +585,12 @@ impl Display {
 
         // Frame event should be requested before swaping buffers, since it requires surface
         // `commit`, which is done by swap buffers under the hood.
-        #[cfg(not(any(target_os = "macos", windows)))]
+        #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
         self.request_frame(&self.window);
 
         self.window.swap_buffers();
 
-        #[cfg(not(any(target_os = "macos", windows)))]
+        #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
         if self.is_x11 {
             // On X11 `swap_buffers` does not block for vsync. However the next OpenGl command
             // will block to synchronize (this is `glClear` in Alacritty), which causes a
@@ -662,7 +665,7 @@ impl Display {
 
     /// Requst a new frame for a window on Wayland.
     #[inline]
-    #[cfg(not(any(target_os = "macos", windows)))]
+    #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
     fn request_frame(&self, window: &Window) {
         let surface = match window.wayland_surface() {
             Some(surface) => surface,
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -726,14 +726,7 @@ impl<N: Notify + OnResize> Processor<N> {
         cli_options: CLIOptions,
     ) -> Processor<N> {
         #[cfg(not(any(target_os = "macos", windows)))]
-        let clipboard = {
-            #[cfg(all(feature = "x11", not(feature = "wayland")))]
-            let wayland_display = None;
-            #[cfg(feature = "wayland")]
-            let wayland_display = display.window.wayland_display();
-
-            Clipboard::new(wayland_display)
-        };
+        let clipboard = Clipboard::new(display.window.wayland_display());
         #[cfg(any(target_os = "macos", windows))]
         let clipboard = Clipboard::new();
 
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -19,7 +19,7 @@ use glutin::dpi::PhysicalSize;
 use glutin::event::{ElementState, Event as GlutinEvent, ModifiersState, MouseButton, WindowEvent};
 use glutin::event_loop::{ControlFlow, EventLoop, EventLoopProxy, EventLoopWindowTarget};
 use glutin::platform::desktop::EventLoopExtDesktop;
-#[cfg(not(any(target_os = "macos", windows)))]
+#[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
 use glutin::platform::unix::EventLoopWindowTargetExtUnix;
 use log::info;
 use serde_json as json;
@@ -726,7 +726,14 @@ impl<N: Notify + OnResize> Processor<N> {
         cli_options: CLIOptions,
     ) -> Processor<N> {
         #[cfg(not(any(target_os = "macos", windows)))]
-        let clipboard = Clipboard::new(display.window.wayland_display());
+        let clipboard = {
+            #[cfg(all(feature = "x11", not(feature = "wayland")))]
+            let wayland_display = None;
+            #[cfg(feature = "wayland")]
+            let wayland_display = display.window.wayland_display();
+
+            Clipboard::new(wayland_display)
+        };
         #[cfg(any(target_os = "macos", windows))]
         let clipboard = Clipboard::new();
 
@@ -749,7 +756,7 @@ impl<N: Notify + OnResize> Processor<N> {
 
     /// Return `true` if `event_queue` is empty, `false` otherwise.
     #[inline]
-    #[cfg(not(any(target_os = "macos", windows)))]
+    #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
     fn event_queue_empty(&mut self) -> bool {
         let wayland_event_queue = match self.display.wayland_event_queue.as_mut() {
             Some(wayland_event_queue) => wayland_event_queue,
@@ -767,7 +774,7 @@ impl<N: Notify + OnResize> Processor<N> {
 
     /// Return `true` if `event_queue` is empty, `false` otherwise.
     #[inline]
-    #[cfg(any(target_os = "macos", windows))]
+    #[cfg(any(target_os = "macos", windows, not(feature = "wayland")))]
     fn event_queue_empty(&mut self) -> bool {
         self.event_queue.is_empty()
     }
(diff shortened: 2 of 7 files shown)


kchibisov:
One more thing that I haven't properly addressed.

We should likely check both features on CI, so right now I've only added jobs for bsd, since I understand sr.ht's format, unlike the one travis is using. Do we really want to check x11/wayland features on linux on travis or not? I feel like it's fine to check them only on bsd.

alacritty/src/display.rs
```diff
@@ -253,8 +254,10 @@ impl Display {
         #[cfg(target_os = "macos")]
         set_font_smoothing(config.ui_config.font.use_thin_strokes());
 
-        #[cfg(not(any(target_os = "macos", windows)))]
+        #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
         let is_x11 = event_loop.is_x11();
+        #[cfg(not(any(feature = "x11", target_os = "macos", windows)))]
+        let is_x11 = false;
```
chrisduerr, line 260: Does this call not work without the backend enabled in glutin? Otherwise it's probably not worth it gating it just for the negligible performance difference.

alacritty/src/event.rs
```diff
@@ -726,7 +726,14 @@ impl<N: Notify + OnResize> Processor<N> {
         cli_options: CLIOptions,
     ) -> Processor<N> {
         #[cfg(not(any(target_os = "macos", windows)))]
-        let clipboard = Clipboard::new(display.window.wayland_display());
+        let clipboard = {
+            #[cfg(all(feature = "x11", not(feature = "wayland")))]
+            let wayland_display = None;
+            #[cfg(feature = "wayland")]
+            let wayland_display = display.window.wayland_display();
```
chrisduerr, line 733: What happens when we call this without Wayland feature enabled? Is it just not available at all?

It would probably be a bit easier and shorter to just have three statements with three annotations using `Clipboard::new(None)` and `Clipboard::new(display.window.wayland_display())` instead?

alacritty/src/window.rs
```diff
@@ -254,22 +262,28 @@ impl Window {
...
-            builder = builder.with_gtk_theme_variant(val.clone())
-        }
+        // Wayland.
+        #[cfg(feature = "wayland")]
+        let builder = builder.with_app_id(class.instance.clone());
+
+        #[cfg(feature = "x11")]
+        let builder = match &window_config.gtk_theme_variant {
+            Some(val) => {
+                let builder = builder.with_class(class.instance.clone(), class.general.clone());
```
chrisduerr, line 281: Why is class only set when gtk_theme_variant is present?

alacritty/src/window.rs
```diff
@@ -254,22 +262,28 @@ impl Window {
...
             .with_visible(false)
             .with_transparent(true)
             .with_decorations(window_config.decorations != Decorations::None)
             .with_maximized(window_config.maximized())
             .with_fullscreen(window_config.fullscreen())
-            .with_window_icon(icon.ok())
-            // X11.
-            .with_class(class.instance.clone(), class.general.clone())
-            // Wayland.
-            .with_app_id(class.instance.clone());
```
chrisduerr, line 268: Honestly I feel like this is almost a mistake in winit to expose these things differently like that. Would it maybe make sense to think about unifying these upstream to have a `with_class` that sets `app_id` on Wayland?

alacritty/src/window.rs
```diff
@@ -313,8 +327,9 @@ impl Window {
     }
 
     #[cfg(not(any(target_os = "macos", windows)))]
-    pub fn set_urgent(&self, is_urgent: bool) {
-        self.window().set_urgent(is_urgent);
+    pub fn set_urgent(&self, _is_urgent: bool) {
+        #[cfg(feature = "x11")]
```
chrisduerr, line 331: These cfgs look very strange. I feel like it should cfg the entire function? Just use the windows `set_urgent` for Wayland?

alacritty/src/window.rs
```diff
@@ -335,7 +350,10 @@ impl Window {
 
     #[cfg(not(any(target_os = "macos", windows)))]
     pub fn x11_window_id(&self) -> Option<usize> {
-        self.window().xlib_window().map(|xlib_window| xlib_window as usize)
+        #[cfg(feature = "x11")]
+        return self.window().xlib_window().map(|xlib_window| xlib_window as usize);
+        #[cfg(not(feature = "x11"))]
+        return None;
     }
```
chrisduerr, line 357: I don't think we need to define this for Wayland. Since we have a config in main.rs that just uses `None` for Windows/macOS anyways. So that should be consistent across cfgs here.

I'd be fine with just having a separate `x11_window_id` method for Windows/macOS/Wayland though if your goal is to clean up the callee a bit.

alacritty/src/window.rs
```diff
@@ -449,8 +467,8 @@ fn x_embed_window(window: &GlutinWindow, parent_id: c_ulong) {
...
 }
 
-#[cfg(not(any(target_os = "macos", windows)))]
+#[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
 unsafe extern "C" fn xembed_error_handler(_: *mut XDisplay, _: *mut XErrorEvent) -> i32 {
-    error!("Could not embed into specified window.");
+    log::error!("Could not embed into specified window.");
```
chrisduerr, line 472: This is kinda inconsistent with every other place where we log. I'd assume you're trying to save yourself the cfg'ed import?

It's probably okay right here, we do similar stuff for `exit`.

chrisduerr (changes requested):
> We should likely check both features on CI, so right now I've only added jobs for bsd, since I understand sr.ht's format, unlike the one travis is using. Do we really want to check x11/wayland features on linux on travis or not? I feel like it's fine to check them only on bsd.

This is one of the bigger problems which will have impact not only on CI complexity, but also on CI runtime. But if we add these features we must add this to all CI platforms.

alacritty/src/display.rs
kchibisov: it doesn't work, there's `is_wayland` though.

alacritty/src/event.rs
kchibisov: I can return `None` form window.rs I guess, should be the same, but in a bit different place. If `Wayland` feature is disable you can't get `wayland_display` since no Wayland.

alacritty/src/window.rs
kchibisov: Ah, I haven't noticed that, will try to move somewhere else.

alacritty/src/window.rs
kchibisov: I can bring back old annotations should be that complex, if you want them.

kchibisov:
>But if we add these features we must add this to all CI platforms.

We already had them though, just wasn't testing at all. I'm not sure that we need to check those on Linux, since winit's API about wayland/non-wayland is just for unix platforms, so I think just running them on bsd is fine, don't you think?

alacritty/src/event.rs
kchibisov: Though, I'd refactor to use 3 statements that just create a clipboard.

alacritty/src/window.rs
kchibisov: Nice catch.

alacritty/src/display.rs
chrisduerr: And that always works? I don't mind either way since it wouldn't change anything for us. Or are you saying we could call is_wayland instead? Which would be pointless.

alacritty/src/event.rs
chrisduerr: I think it might be nice to just handle these kind of things in window.rs when I think about it tbh. That way our callee never has to worry about a thing.

alacritty/src/window.rs
kchibisov: Maybe it can accept some kind of enum, and have them in one call, but the main issue is that those things have different amount of arguments and don't really relate to each other.

alacritty/src/window.rs
chrisduerr: I think this is fine, just wanted to make sure.

alacritty/src/display.rs
kchibisov: No, it doesn't always work, it just that knowing that it's a case when you have wayland feature we can call this function, but you can make one statement out of this.

chrisduerr:
> We already had them though, just wasn't testing at all. I'm not sure that we need to check those on Linux, since winit's API about wayland/non-wayland is just for unix platforms, so I think just running them on bsd is fine, don't you think?

That's a good point. I'm just not sure if we shouldn't at least run them on Linux? Since BSD/Linux can definitely make a difference, though I suppose unlikely where X11/Wayland features are used? It's probably fine, we can go with BSD first and then just see how it works out I suppose.

alacritty/src/window.rs
kchibisov: I feel like I've done it like that to not comment call out on caller side.

kchibisov:
>That's a good point. I'm just not sure if we shouldn't at least run them on Linux? Since BSD/Linux can definitely make a difference, though I suppose unlikely where X11/Wayland features are used? It's probably fine, we can go with BSD first and then just see how it works out I suppose.

Yeah, just bsd is fine ( I just don't want to mess up with travis, since it's way to complex compared to shell scripts on sr.ht. )

alacritty/src/display.rs
chrisduerr: How could we make one statement out of it? We'd need another one for `is_x11 = true` then wouldn't we?

kchibisov: You can't, typo.

alacritty/src/window.rs
chrisduerr: How do they not relate to each other? Isn't `app_id` basically the Wayland class equivalent? And I think we can resolve the different parameter issue with thorough documentation.

As you can see we need to pass both anyways since we want to work on X11 and winit isn't a toolkit that should be used to build applications that work just one Wayland. So there should be abstractions and I think this abstraction would make sense.

Is there any scenario where you would want to set the class instance to something different than the `app_id`?

alacritty/src/window.rs
chrisduerr: If you merge it with the windows cfg the callsite will not change at all.

alacritty/src/window.rs
kchibisov: I guess you can call it as alternative, I'm just saying that there's no instance thing. You can propose unification upstream.

chrisduerr: I will, thanks.

alacritty/src/window.rs
```diff
@@ -153,15 +157,16 @@ impl Window {
...
         let window_config = &config.ui_config.window;
         let window_builder = Window::get_platform_window(&window_config.title, &window_config);
 
         // Disable vsync on Wayland.
-        #[cfg(not(any(target_os = "macos", windows)))]
+        #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
         let vsync = !event_loop.is_wayland();
-        #[cfg(any(target_os = "macos", windows))]
+        #[cfg(any(target_os = "macos", windows, not(feature = "wayland")))]
         let vsync = true;
```
chrisduerr, line 170: How expensive is the Wayland check? We perform an is_wayland check here and later do an `is_x11` check. I think in theory we could just store this to a variable named x11 (since that seems our common name for this) and then use that for vsync and the checks later.

Would probably also help the optimizer.

alacritty/src/window.rs
```diff
@@ -254,22 +262,26 @@ impl Window {
...
-            // X11.
-            .with_class(class.instance.clone(), class.general.clone())
-            // Wayland.
-            .with_app_id(class.instance.clone());
+            .with_window_icon(icon.ok());
 
-        if let Some(ref val) = window_config.gtk_theme_variant {
-            builder = builder.with_gtk_theme_variant(val.clone())
-        }
+        // Wayland.
```
chrisduerr, line 274:
```suggestion
```

Since this block handles both Wayland/X11 the docs here are a bit confusing. Easier to just read the cfgs.

alacritty/src/window.rs
```diff
@@ -254,22 +262,26 @@ impl Window {
...
+        #[cfg(feature = "wayland")]
+        let builder = builder.with_app_id(class.instance.clone());
+        #[cfg(feature = "x11")]
+        let builder = builder.with_class(class.instance.clone(), class.general.clone());
+
+        #[cfg(feature = "x11")]
+        let builder = match &window_config.gtk_theme_variant {
+            Some(val) => builder.with_gtk_theme_variant(val.clone()),
+            None => builder,
+        };
```
chrisduerr, line 284: I think the `if let Some` used previously is a bit nicer here?

alacritty/src/window.rs
```diff
@@ -254,22 +262,26 @@ impl Window {
 
         let class = &window_config.class;
 
-        let mut builder = WindowBuilder::new()
+        let builder = WindowBuilder::new()
```
chrisduerr, line 265: Why did you remove the mutability and then later just redefine the variable? Wouldn't it be easier to just mutate it?

alacritty/src/window.rs
```diff
@@ -312,9 +324,9 @@ impl Window {
         }
     }
 
-    #[cfg(not(any(target_os = "macos", windows)))]
-    pub fn set_urgent(&self, is_urgent: bool) {
-        self.window().set_urgent(is_urgent);
+    #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
+    pub fn set_urgent(&self, _is_urgent: bool) {
+        self.window().set_urgent(_is_urgent);
```
chrisduerr, line 329:
```suggestion
    pub fn set_urgent(&self, is_urgent: bool) {
        self.window().set_urgent(is_urgent);
```

alacritty/src/window.rs
```diff
@@ -326,16 +338,18 @@ impl Window {
...
         self.window().set_outer_position(pos);
     }
 
-    #[cfg(not(any(target_os = "macos", windows)))]
     pub fn x11_window_id(&self) -> Option<usize> {
-        self.window().xlib_window().map(|xlib_window| xlib_window as usize)
+        #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
+        return self.window().xlib_window().map(|xlib_window| xlib_window as usize);
+        #[cfg(any(target_os = "macos", windows, not(feature = "x11")))]
+        return None;
```
chrisduerr, line 352: Do we maybe want to make this and `wayland_display` into separate methods for `None`? I think that might be a bit cleaner. Either way it should be uniform and currently it's a bit of a mix between the two since `set_urgent` uses separate methods for example.

chrisduerr (changes requested):

alacritty/src/window.rs
kchibisov: Rust doesn't allow putting cfg before assignment.

alacritty/src/window.rs
kchibisov: We can't use `mut`, since we can't put cfg above `builder = builder`, so `match` is shorter than `if let`.

alacritty/src/window.rs
kchibisov: Well, there's no x11 on macOS/Windows and winit check is free, since it just checks enum variant and returns a constant.

alacritty/src/window.rs
chrisduerr: Huh, you're right. TIL.

alacritty/src/window.rs
chrisduerr: All we want to check is if it's Wayland or not, right? For everything else it should always use vsync? So I don't see what macOS/Windows have to do with it since we could just always define `x11 = false` in that case?

kchibisov: updated.

alacritty/src/window.rs
```diff
@@ -153,20 +157,23 @@ impl Window {
...
         config: &Config,
         size: Option<PhysicalSize<u32>>,
-        #[cfg(not(any(target_os = "macos", windows)))] wayland_event_queue: Option<&EventQueue>,
+        #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
+        wayland_event_queue: Option<&EventQueue>,
     ) -> Result<Window> {
         let window_config = &config.ui_config.window;
         let window_builder = Window::get_platform_window(&window_config.title, &window_config);
 
         // Disable vsync on Wayland.
```
chrisduerr, line 166: This comment isn't entirely accurate anymore. I'd go more for something like this:

```suggestion
        // Check if we're running Wayland to disable vsync.
```

That way we still have the vsync thing in here, but it doesn't act like this is directly disabling it. Though I'm open for suggestions.

alacritty/src/window.rs
```diff
@@ -374,17 +393,22 @@ impl Window {
...
-    pub fn wayland_display(&self) -> Option<*mut c_void> {
+    #[cfg(any(not(feature = "wayland"), any(target_os = "macos", windows)))]
+    pub fn wayland_display(&self) -> Option<*mut std::ffi::c_void> {
+        None
+    }
+
+    #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
+    pub fn wayland_display(&self) -> Option<*mut std::ffi::c_void> {
         self.window().wayland_display()
     }
```
chrisduerr, line 404: Could we order these methods uniformly to have the simple "does nothing" version always at the end? I think that makes things a bit easier to understand when reading through things from top to bottom.

chrisduerr (changes requested):

chrisduerr (approved):
LGTM

joshuarli:
This is wonderful, thank you @kchibisov for caring!

Issue #2950, opened 2019-11-04
Alacritty should provide option to completely turn off icon bouncing.


chakrit:
**Which operating system does the issue occur on?**
OS X

**If on linux, are you using X11 or Wayland?**
(n/a)

**Description**
This is somewhat related to #2492 but I think this is more Alacritty's job to provide a knob to turn off any kind of notification altogether. Think of this as having apps requesting your permission to send you notifications. Alacritty shouldn't just let any shell program cause the icon to bounce, it should be the one in-control (and thus let the user choose).

It is quite annoying to have to configure and re-configure each piece of software to handle signals correctly every time I re-install it or move between OSes/shells.

Something like this would be perfect:

```yaml
bell:
  enabled: false
  visual_bell:
    # maybe move visual bell option here?
```

kchibisov:
>Alacritty shouldn't just let any shell program cause the icon to bounce, it should be the one in-control (and thus let the user choose).

We're just getting escapes, we don't know who is sending them etc, so it's nearly impossible to control things in the suggested way.

The question is whether we should allow disabling setting urgency flag on a window through the bell.

I personally don't mind adding something to toggle this kind of things, however I wonder who is responsible for silencing urgency requests in general. I think if you want to "disable icon jumping" for, let's say, `N` apps, you probably want to disable this behavior in one specific place (e.g. your WM settings), and not changing all application to fit your needs.

I also wondering about how other macOS terminal emulators behave in the mentioned use case (I don't have macOS machine to test things myself), but I feel like no one provides a toggle to alter this behavior (they're not setting urgency at all or setting it all the time like `alacritty`).

chrisduerr:
There are two escape sequences which already exist that Alacritty currently does not implement:
 - DEC 1042 -> Enable Urgency window manager hint on bell
 - DEC 1043 -> Enable raising of the window on bell

nixpulvis:
I'm copying the descriptions for these escapes below:

**`bellIsUrgent`**
> Ps = 1 0 4 2  ⇒  Enable Urgency window manager hint when Control-G is received, xterm. This enables the bellIsUrgent resource.
>
> bellIsUrgent (class BellIsUrgent)
Specifies whether to set the Urgency hint for the window manager when making a bell sound.  The default is "false".

**`popOnBell`**
> Ps = 1 0 4 3  ⇒  Enable raising of the window when Control-G is received, xterm.  This enables the popOnBell resource.
>
> popOnBell (class PopOnBell)
[...]

PR #4304, opened 2020-10-10
alacritty: support enabling/disabling urgency hints

ttaylorr:
On macOS, setting the window to urgent causes Alacritty's dock icon to
bounce until the window is focused again (which removes the window's
urgency bit). This is a behavior that users have requested to be able to
turn off \[1\].

Teach Alacritty to stop setting the window as urgent upon a bell by
emulating xterm's 'bellIsUrgent' resource and relevant CSIs. When this
resource is enabled (with 'CSI ? 1042 h'), a bell event causes the
window to be marked as urgent. When the resource is disabled (with 'CSI
? 1042 l'), the window is not marked urgent in the event of a bell.

There are two wrinkles worth noting here:

  - The 'TermMode::URGENCY_HINTS' does _not_ affect the terminal's
    configured bell command, since we only want to control whether or
    not the window is marked as urgent, not anything else.

  - In xterm, the 'bellIsUrgent' resource is _disabled_ by default.
    Since bouncing the dock icon has been the default in Alacritty on
    macOS thus far, do not make an effort to change that in this patch.

This allows users to write "\e[?1042l" in their prompt and disable
bouncing the dock icon, if they so desire.

\[1\]: https://github.com/alacritty/alacritty/issues/2950

Signed-off-by: Taylor Blau \<me@ttaylorr.com\>

Closes: https://github.com/alacritty/alacritty/issues/2950

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Escape sequences to report text area size in pixels (`CSI 14 t`) and in characters (`CSI 18 t`)
 - Support for single line terminals dimensions
 - Right clicking on Wayland's client side decorations will show application menu
+- Escape sequences to enable and disable window urgency hints (`CSI ? 1042 h`, `CSI ? 1042 l`)
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


chrisduerr (approved):
Looks and works great, thanks for your contribution.

PR #4309, opened 2020-10-12
Fix incorrect X11 feature check

chrisduerr:
Thanks to @aruhier for spotting this so quickly! (https://github.com/alacritty/alacritty/commit/721f789b5f98ab5c47e6e817c3c2228636ca0a1a#r43183826)

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -767,7 +767,7 @@ impl<N: Notify + OnResize> Processor<N> {
 
     /// Return `true` if `event_queue` is empty, `false` otherwise.
     #[inline]
-    #[cfg(any(target_os = "macos", windows, not(feature = "wayland")))]
+    #[cfg(any(not(feature = "wayland"), target_os = "macos", windows))]
     fn event_queue_empty(&mut self) -> bool {
         self.event_queue.is_empty()
     }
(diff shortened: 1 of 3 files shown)


alacritty/src/event.rs
```diff
@@ -767,7 +767,7 @@ impl<N: Notify + OnResize> Processor<N> {
 
     /// Return `true` if `event_queue` is empty, `false` otherwise.
     #[inline]
-    #[cfg(any(target_os = "macos", windows, not(feature = "wayland")))]
+    #[cfg(any(not(feature = "wayland"), target_os = "macos", windows))]
```
chrisduerr, line 770: I've changed this to move the wayland feature **before** the platform checks. This is for uniformity to make it easier and quicker to read these `cfg` annotations, having a uniform structure everywhere.

kchibisov (approved):

alacritty/src/window.rs
```diff
@@ -193,6 +190,8 @@ impl Window {
             }
         }
 
+        #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
+        let mut wayland_surface = None;
```
chrisduerr, line 194: Moved this closer to where it is used, just to reduce code complexity a bit.

aruhier:
Thanks a lot for the patch!

kchibisov (approved):

PR #4311, opened 2020-10-13
Follow xterm and preserve foreground color on reset

### LINKED ISSUE #4264: Black cursor in neovim with dark colorscheme
For bug reports, the following information can help speed up the process. Please describe the bug that you have found and what you would expect to happen instead.

### System

OS: Arch Linux
Version: alacritty 0.5.0
Sway WM, Wayland.
Alacritty Config: https://github.com/alexryndin/dotfiles/blob/7cc014348e758edc56b6e7c4e37213ec89a61171/.config/alacritty/alacritty.yml

I prefer white and black colorscheme in alacritty, but I use dark scheme (gruvbox) in vim. It looks good outside vim:
grim-27 09 2020-180138
But cursor color fails to match vim's colorscheme (cursor is on line 2):
grim-27 09 2020-180400
Compare to gnome-terminal:
grim-27 09 2020-180511

kchibisov:
Xterm preserves foreground color on reset, but those cells are
considered empty in it, thus we change our 'is_empty' and 'reset' logic
to match it.

Fixes #4264.

### DIFF
--- a/alacritty_terminal/src/term/cell.rs
+++ b/alacritty_terminal/src/term/cell.rs
@@ -55,7 +55,6 @@ impl GridCell for Cell {
         (self.c == ' ' || self.c == '\t')
             && self.extra[0] == ' '
             && self.bg == Color::Named(NamedColor::Background)
-            && self.fg == Color::Named(NamedColor::Foreground)
             && !self.flags.intersects(
                 Flags::INVERSE
                     | Flags::UNDERLINE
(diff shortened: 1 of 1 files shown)


alacritty_terminal/src/term/cell.rs
```diff
@@ -131,7 +130,7 @@ impl Cell {
     #[inline]
     pub fn reset(&mut self, template: &Cell) {
         // memcpy template to self.
```
nixpulvis, line 132: I feel like this comment should be updated or removed.

chrisduerr: No worries, chances are the non-copy cells PR will do just that.

PR #4313, opened 2020-10-13
Adding Ref Test Diffs

nixpulvis:
Show diffs on ref tests with the `pretty_assertions` crate. The one thing I'm not really sure about with this is if it's better to just fail immediately on the first non-match, or to keep things more like they were before as I've done here.

screenshot-2020-10-13T18:10:30Z

vs before:

screenshot-2020-10-13T18:22:46Z

### DIFF
--- a/alacritty_terminal/tests/ref.rs
+++ b/alacritty_terminal/tests/ref.rs
@@ -1,3 +1,4 @@
+use pretty_assertions::assert_eq;
 use serde::Deserialize;
 use serde_json as json;
 
@@ -119,20 +120,10 @@ fn ref_test(dir: &Path) {
             for j in 0..grid.cols().0 {
                 let cell = term_grid[i][Column(j)];
                 let original_cell = grid[i][Column(j)];
-                if original_cell != cell {
-                    println!(
-                        "[{i}][{j}] {original:?} => {now:?}",
-                        i = i,
-                        j = j,
-                        original = original_cell,
-                        now = cell,
-                    );
-                }
+                println!("{}", pretty_assertions::Comparison::new(&cell, &original_cell));
             }
         }
 
-        panic!("Ref test failed; grid doesn't match");
+        panic!("ref test failed; grid doesn't match");
     }
-
-    assert_eq!(grid, term_grid);
 }
--- a/alacritty_terminal/tests/ref.rs
+++ b/alacritty_terminal/tests/ref.rs
@@ -1,4 +1,4 @@
-use pretty_assertions::assert_eq;
+use pretty_assertions::Comparison;
 use serde::Deserialize;
 use serde_json as json;
 
@@ -120,7 +120,7 @@ fn ref_test(dir: &Path) {
             for j in 0..grid.cols().0 {
                 let cell = term_grid[i][Column(j)];
                 let original_cell = grid[i][Column(j)];
-                println!("{}", pretty_assertions::Comparison::new(&cell, &original_cell));
+                println!("{}", Comparison::new(&original_cell, &cell));
             }
         }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,6 +68,7 @@ dependencies = [
  "miow 0.3.5",
  "nix",
  "parking_lot",
+ "pretty_assertions",
  "regex-automata",
  "serde",
  "serde_json",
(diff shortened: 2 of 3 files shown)


alacritty_terminal/tests/ref.rs
```diff
@@ -119,20 +120,10 @@ fn ref_test(dir: &Path) {
...
-                }
+                println!("{}", Comparison::new(&original_cell, &cell));
             }
         }
 
-        panic!("Ref test failed; grid doesn't match");
+        panic!("ref test failed; grid doesn't match");
     }
-
-    assert_eq!(grid, term_grid);
```
chrisduerr, line 137: Why did you remove this? It's necessary.

chrisduerr (changes requested):
I doubt that this is useful at all. In my experience the diffs the reftest usually create are completely different from the kind of things that are shown in your example.

How useful is the output of the diff when **every single cell** has different flags?

alacritty_terminal/tests/ref.rs
nixpulvis: Oh indeed, I got a bit hasty with my editing.

nixpulvis:
@chrisduerr indeed, this crate seems to do nothing useful at all with a failing cell on `flags`. I suppose there are cases where something like this would be nice, for example the attached images of the foreground color diff. But there are still many cases where it will only further complicate the output.

I'm curious if it's worth it to try and identify the cases where it's helpful, and apply the coloring there, otherwise revert to something simpler like we have today?

chrisduerr:
> I suppose there are cases where something like this would be nice, for example the attached images of the foreground color diff. But there are still many cases where it will only further complicate the output.

So why not do the exact same thing we do right now and put it through `diff` if you want diffs? Why do we need to bundle everything into one thing?

> I'm curious if it's worth it to try and identify the cases where it's helpful, and apply the coloring there, otherwise revert to something simpler like we have today?

I do not think that is useful. If you want a diff, just run the output through a diff tool. It's far more composable, allowing you to use whatever tool you want and you still get what you actually need - the raw output - from the tests themselves.

99% of the time, I do not get something that works nicely as a formatted diff like in your example. That certainly is the (very rare) exception.

nixpulvis:
> So why not do the exact same thing we do right now and put it through diff if you want diffs? Why do we need to bundle everything into one thing?

Our output isn't really diff compatible... it does however already show the before and after, it's just kinda hard to make out.

That said, it's not too hard to save the output from the test and run that through diff on the original, so we could just do that, but I was hoping this might help people out. We just need to save the new grid.json on failure somewhere.

chrisduerr:
> Our output isn't really diff compatible

I've run it through diff tools a lot in the past without any troubles at all. How is it not "diff compatible"? In fact what does that even mean?

nixpulvis:
> How is it not "diff compatible"?

If you look at the output of the test, it's already itself a diff. It's just not highlighting the changes as I was hoping to.
```
[29][103] Cell { c: ' ', fg: Named(Foreground), bg: Named(Background), flags: (empty), extra: [' ', ' ', ' ', ' ', ' '] } => Cell { c: ' ', fg: Named(Red), bg: Named(Background), flags: (empty), extra: [' ', ' ', ' ', ' ', ' '] }
```

> I've run it through diff tools a lot in the past without any troubles at all.

For me to use a standard diff tool, I'd need a way to save the output of the tested `term_grid`, which isn't all that hard to implement, but is not provided in the ref tests today.

I feel like we're talking about diffing different things. Either that or I'm not understanding you.

nixpulvis:
I feel like I'm going crazy...

How is it I'm supposed to take the output and pass it to `diff` exactly? I would need to separate the part before and after the `=>` into two separate parts, then pass them each to `diff`, no? This seems like it could be made simpler to check.

chrisduerr:
> How is it I'm supposed to take the output and pass it to diff exactly?

You enter vi mode, start copy, go to the beginning of the line and done. Then save that to a file or do whatever you want with it and same goes for the "left" side of the comparison.

Especially with vi mode this really isn't a difficult task.

nixpulvis:
OK, it's true it's not horribly difficult. I was just thinking it could be nice to try and make the output of the test more directly informative / viable. But if that's not the working model, I'm happy to just close this.

chrisduerr:
Based on my experience, I don't think this would improve much.

PR #4317, opened 2020-10-18
Handling CompositionEvent

Riey:
## Summary

Support `CompositionEvent` implemented on winit's [sub branch](https://github.com/rust-windowing/winit/tree/composition-event) for IME preedit support

I check work on my Arch Linux + uim(byeoru)

For now this only work on X11 because winit doesn't have implementation for other platforms

## Blocking upstream issue

https://github.com/rust-windowing/winit/issues/1497

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,6 +49,7 @@ dependencies = [
  "urlocator",
  "wayland-client",
  "winapi 0.3.9",
+ "winit",
  "x11-dl",
  "xdg",
 ]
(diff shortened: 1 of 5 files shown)


chrisduerr (commented):
> For now this only work on X11 because winit doesn't have implementation for other platforms

What's the reason for creating a PR for Alacritty when it is not even done being implemented in winit?

Riey:
> What's the reason for creating a PR for Alacritty when it is not even done being implemented in winit?

I think it could be a draft PR cause this is marked as WIP if you don't think about like that I'll close this and create later

chrisduerr:
> I'll close this and create later

Much appreciated, having PRs around for a long time with nobody working on it just makes it a lot harder to see which PRs actually are progressing.

Issue #2237, opened 2019-04-01
Live config reload doesn't notice when symlink is replaced


Nemo157:
I use a config file manager that symlinks in my alacritty config file and replaces the symlink when updated. Because alacritty [canonicalizes the config filename](https://github.com/jwilm/alacritty/blob/a9c6e2807f3a205ee1adc8adb732aac9215ea6cb/src/config/mod.rs#L2298) it ends up watching the old config file and doesn't notice when the symlink is updated to point at the new config file (at least, I'm assuming this is the issue, I'll try to get around to adding some debugging and checking exactly what alacritty does soon).

It seems like accurately watching a symlinked file is going to be difficult, you need to notice when the symlink _or_ the targeted file changes, and there could be many layers of symlinks to watch (include having directory components symlinked and needing to notice when they update). If you feel this is too complex to be worth it I'm fine with marking this as won't fix.

(Linux 5.0.4-arch1-1-ARCH x86_64 GNU/Linux, X11)

chrisduerr:
I think we could potentially make this a little better by checking if the real path of the file is different than the file itself and then watch both directories if that's the case. With nested symlink levels that would still only cover changing the file itself and the source file, but I think that's the only sane solution that doesn't involve a change on the user's system.

I'm not sure how watching symlinks works out with the library we're using though. This would need to be investigated first and we might need to consider that different kernels (especially linux/macos/bsd in this regard) might have different APIs to watch files used by our dependency.

Then the alternative is to change the user's system. Of course we can't really control this, but looking at your usecase, I feel like maybe hardlinks might help? The advantage of hardlinks is that they still get updated when either file is updated, however the real path does not resolve to the 'original file' (since there really is none). I think these should also work correctly with the library we're using and hopefully resolve all your problems?

Nemo157:
I've actually investigated this slightly more and found the issue is slightly different to what I thought. The symlink is not being replaced, but a grandparent directory of the file it symlinks to is. The [tool I use](https://github.com/freshshell/fresh) creates a new directory, builds new versions of config files into it, then replaces a the grandparent with the new directory.

Adding some debugging it appears that alacritty only sees the removal of the config file and its direct parent once the new directory is in place, since it is not watching the grandparent directory (`/home/wim/.fresh/build`) that gets replaced:

```
[src/config/mod.rs:2310] rx.recv().expect("watcher event") = Remove(
    "/home/wim/.fresh/build/config-alacritty/alacritty.yml"
)
[...]

PR #4322, opened 2020-10-20
Add polling for a new config to watch when config is deleted

oknozor:
Hello , this is my first PR to alacritty so I have a few questions : 
- do this needs to be documented in changelog.md  ? 
- shall I bump the version (again I am not sure) ? 

This addresses #2237. 
Regarding the implementation I am not sure what I wrote is a 100% correct : 

- Is it safe to not resolve symlinks ? 
  ```rust
              if path.symlink_metadata().is_ok() {
                  Symlink(path.clone())
              } else {
                  ConfigPath::Path(path.clone())
              }
  ```   
  The idea here is to keep symlinks in place instead of following them. This way when config is unlinked we can wait for it to be re-linked. Symlink paths are not canonicalized in the next step, instead they are checked with `Path::exist`, could this have potential side effects ? 

- How long should we wait for config to be recreated ? 
  I have set a timeout of 6s with 100ms steps, is this appropriate ?

Let me now if anything needs to be changed.

### DIFF
--- a/alacritty/src/config/monitor.rs
+++ b/alacritty/src/config/monitor.rs
@@ -10,26 +10,42 @@ use alacritty_terminal::thread;
 
 use crate::event::{Event, EventProxy};
 
-pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventProxy) {
-    // Canonicalize all paths, filtering out the ones that do not exist.
-    paths = paths
-        .drain(..)
-        .filter_map(|path| match fs::canonicalize(&path) {
-            Ok(path) => Some(path),
+struct ConfigPath {
+    path: PathBuf,
+    symlink: Option<PathBuf>,
+}
+
+impl ConfigPath {
+    fn get(path: PathBuf) -> Option<Self> {
+        // Canonicalize all paths, filtering out the ones that do not exist.
+        match fs::canonicalize(&path) {
+            Ok(path) => {
+                // Keep the symlink in case it is recreated
+                let symlink =
+                    if path.symlink_metadata().is_ok() { Some(path.clone()) } else { None };
+
+                Some(ConfigPath { path, symlink })
+            },
             Err(err) => {
                 error!("Unable to canonicalize config path {:?}: {}", path, err);
                 None
             },
-        })
-        .collect();
+        }
+    }
+}
+
+pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventProxy) {
+    // Canonicalize paths and eventually get symlinks
+    let config_paths: Vec<ConfigPath> = paths.drain(..).filter_map(ConfigPath::get).collect();
 
     // Don't monitor config if there is no path to watch.
-    if paths.is_empty() {
+    if config_paths.is_empty() {
         return;
     }
 
     // The Duration argument is a debouncing period.
     let (tx, rx) = mpsc::channel();
+
     let mut watcher = match watcher(tx, Duration::from_millis(10)) {
         Ok(watcher) => watcher,
         Err(err) => {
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
I don't see why we would jump through so many hoops just to detect if a config might have been replaced or not. Especially because this still doesn't make symlinks work properly.

Also that polling rate is way too high anyways, at that point the user's system basically under load when doing nothing.

So why special case symlinks when it doesn't help with symlinks anyways? At that point we're just always watching the parent directory which we could have been doing from the beginning then (but we obviously don't want to).

Also why do you start watching all parent directories after having figured out that one of the files must have been replaced? At that point we'd already have our file.

oknozor:
@chrisduerr 

>  polling rate is way too high anyways

Ok I will change this, what would be a correct polling rate then ?


>  point we'd already have our file

I will also make the required changes to watch only the removed file.

>  So why special case symlinks when it doesn't help with symlinks 

Regarding this one I am not sure to understand why this doesn't help with symlinks. 
This I what I understand so far :  
 - `std::fs::canonicalize` resolve symlinks. 
  ```rust
      let path = std::fs::canonicalize("/home/okno/.config/alacritty/alacritty.yml").unwrap();
      assert_eq!(path, PathBuf::from("/home/okno/dotfiles/.dots/alacritty/alacritty.yml"));
  ```

By having a special case for them we can directly watch for the symlink and not the referenced file : 

- Given alacritty started with the following config symlinked.
  `.config/alacritty -> /home/okno/dotfiles/.dots/alacritty`
- And the config is unlinked then relinked to another config.
  `alacritty -> /home/okno/dotfiles/.dots/alacritty2`

If we are watching for changes in `/home/okno/dotfiles/.dots/alacritty`, we can never find if the symlink is now referencing another file. 
 
That's why I added this special case. Could you explain why this is  a problem ?

chrisduerr:
> Regarding this one I am not sure to understand why this doesn't help with symlinks.

Have you tested your code?

oknozor:
Yes I works like a charm for me, I am able to unlink/relink config and it reloads as expected.

chrisduerr:
> Yes I works like a charm for me, I am able to unlink/relink config and it reloads as expected.

It works like a charm? It does not reload the configuration file anymore when you edit the symlinked file. It performs worse for symlinks than the latest master.

oknozor:
Relinking works but editing the symlink does not update indeed, sorry I missed this. I'll try to fix it.

kchibisov (changes requested):
I'm against of adding a polling approach and approach that adds more complexity to supervising. If anything we can add explicit binding to force config reload and research, and if it notices that the previous config no longer accessible via default path, or just reload if it's still there.

oknozor:
@chrisduerr  I have updated to make the config reload correctly when relinking or editing the file. 


> If anything we can add explicit binding to force config reload and research

@kchibisov  
Where and how should that be implemented ?

kchibisov:
>Where and how should that be implemented ?

I suggest to add an `ReloadConfig` binding for `key_bindings`, and leave the current polling alone, adding more complexity to it isn't what everyone wants to maintain for very small amount of users.

So to implement it you'd need to add that binding to `alacritty/src/config/bindings.rs`.

Then you should add handling into `alacritty/src/input.rs`. The implementation of that binding could be done in a following way:


- Get current config logical path (not sure if it's stored anywhere, may require storing)`
- Try to load config from the current logical path, if the physical path to that config differs to the current config physical path start watching new config and drop watching of the old one
- Reload config, could be done via issuing `Event::ConfigReload` or we may call to `reload_config` function from event.rs directly, though it could require moving it a bit.

The most complicated part here is recreating or updating the config watcher thread. With the current model it seems like recreating is what should be performed. I feel like the less complex is recreating config watcher, since for informing thing you'd need something like `mio`, though we already depend on it...

I'm not sure that the name should be just `ReloadConfig`, but I guess we can figure that out later, maybe `ForceConfigReload`.

chrisduerr:
There has been some work to better support symlinks and a polling approach has never been a particularly great option. So closing as resolved. For even more complex scenarios I'd recommend something like #5313.

Issue #4327, opened 2020-10-21
Incorrect DPR when launched in a second monitor


noritada:
Hi,

I encountered phenomena similar to #71.

When I launch the app from a second monitor with DPR 1, it displays the half-sized display in a correct size window. When I move the window to the main monitor with DPR 2, then it displays correctly, and when I move it again to the second monitor, it also displays correctly, although the log reports that it changed DPR 1 -> DPR 2 -> DPR 1.

On the other hand, when I launch alacritty from a primary builtin monitor with DPR 2, it works fine.

(It seems that "DPR 1" has already been detected before the display is drawn, but the display does not reflect it. I have checked the code and tried some changes, but cannot figure out how to run something like `terminal.resize(size_info)` inside the initialization process. What changes shall I try?)

| Action | Monitor   | Monitor DPR | Logged DPR | Expected DPR | Actual DPR |
|--------|-----------|---|---|---|---|
| Launch | secondary | 1 | 1 | 1 | 2? |
| Move   | primary   | 2 | 2 | 2 | 2  |
| Move   | secondary | 1 | 1 | 1 | 1  |

| Action | Monitor   | Monitor DPR | Logged DPR | Expected DPR | Actual DPR |
|--------|-----------|---|---|---|---|
| Launch | primary   | 2 | 2 | 2 | 2  |
| Move   | secondary | 1 | 1 | 1 | 1  |
| Move   | primary   | 2 | 2 | 2 | 2  |

![ss 2020-10-21 23 02 21（2）](https://user-images.githubusercontent.com/20850/96739281-922b6000-13fa-11eb-873a-72723fc9e35c.png)

### System

OS: macOS High Sierra (10.13.6)
Version: alacritty `0.6.0-dev (21eb84f)` (built from commit 21eb84fc5e0e649f53e3785957fa127ee9a78688)

### Logs

secondary (1) -> primary (2) -> secondary (1):
```
% /Applications/Alacritty.app/Contents/MacOS/alacritty -vv
[2020-10-21 23:02:18.127642000] [DEBUG] [alacritty] Using environment locale: ja_JP.UTF-8
Created log file at "/var/folders/wp/x1kk7wts6v3g_g1fy3zwp5c40000gn/T/Alacritty-13668.log"
[2020-10-21 23:02:18.128555000] [INFO ] [alacritty] Welcome to Alacritty
[2020-10-21 23:02:18.128596000] [INFO ] [alacritty] Configuration files loaded from:
[...]

PR #4328, opened 2020-10-21
Check viewport size after initial resize

chrisduerr:
Since the `set_inner_size` call might immediately alter the window size,
the query for the window's dimensions should be performed after the
resize has been requested.

### DIFF
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -206,9 +206,6 @@ impl Display {
 
         info!("Device pixel ratio: {}", window.dpr);
 
-        // get window properties for initializing the other subsystems.
-        let viewport_size = window.inner_size();
-
         // Create renderer.
         let mut renderer = QuadRenderer::new()?;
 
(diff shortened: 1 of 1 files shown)


alacritty/src/display.rs
```diff
@@ -240,6 +238,7 @@ impl Display {
 
         info!("Cell size: {} x {}", cell_width, cell_height);
         info!("Padding: {} x {}", size_info.padding_x(), size_info.padding_y());
+        info!("Width: {}, Height: {}", size_info.width(), size_info.height());
```
chrisduerr, line 241: We previously haven't logged this at startup, but I think it is very useful especially because we log it on resize.

chrisduerr (commented):
I'd like to wait for #4327, since it might be possible to repurpose this PR to include a fix for that if the issue becomes obvious in the near future.

kchibisov (approved):

PR #4331, opened 2020-10-22
Remove unneeded `collect()` in `RenderLines.rects`

mlodato517:
Just something I noticed while looking around and it seemed easy/safe to improve. I ran the benchmarks a couple times and it wasn't _obviously_ faster but I was just looking number-by-number in my editor because, for whatever reason, when I run `./gnuplot/summary.sh *.dat output.svg` I get `line 0: undefined variable plot` 🤷 But, if someone else can help me get the plotter set up I can confirm if this is indeed an improvement.

**This Commit**
Uses `move` semantics to copy a `Flags` instance instead of allocating a
`Vec<RenderRect>`.

**Why?**
I expect the performance is better here - `Flags` is a `u16` so making
copies of that should be very cheap while heap-allocating a
`Vec<RenderRect>` is probably more costly. Although, of course, the
compiler could be doing all sorts of witchcraft to optimize away an
unnecessary allocation.

I also used `flat_map` which
seemed more idiomatic than `.map(...).flatten()` but I believe they're
functionally equivalent.

### DIFF
--- a/alacritty/src/renderer/rects.rs
+++ b/alacritty/src/renderer/rects.rs
@@ -142,10 +142,9 @@ impl RenderLines {
     pub fn rects(&self, metrics: &Metrics, size: &SizeInfo) -> Vec<RenderRect> {
         self.inner
             .iter()
-            .map(|(flag, lines)| -> Vec<RenderRect> {
-                lines.iter().map(|line| line.rects(*flag, metrics, size)).flatten().collect()
+            .flat_map(|(flag, lines)| {
+                lines.iter().flat_map(move |line| line.rects(*flag, metrics, size))
             })
-            .flatten()
             .collect()
     }
 


chrisduerr:
> when I run ./gnuplot/summary.sh *.dat output.svg I get line 0: undefined variable plot

Do you have bash and gnuplot installed? I don't see how you could get that error with the latest master of vtebench, there is no variable plot.

mlodato517:
Aha! This was my problem - https://superuser.com/questions/1033273/bash-4-3-substring-negative-length-on-os-x. I edited the shell script and it worked:

gnuplot-alacritty

So it might be slower so maybe forget this

chrisduerr (approved):
Code looks good, thanks.

chrisduerr:
> So it might be slower so maybe forget this

I'll run the benchmarks on my benchmark gig. I don't see why this would be any slower.

chrisduerr:
So just for transparency, in my first run this did actually seem to perform slightly worse in the `scrolling` benchmark. However since that seemed a bit unusual to me, I've run it two more times and would have to conclude that it's just a bit inconsistent.

Here are the full results:

tmp

tmp

tmp

Since all the "regressions" of the first benchmark are within the margin of error, I feel fairly confident in saying that this shouldn't be a problem.

mlodato517:
Huh ... fascinating. So it seems like it might not be a regression but it also seems like it might not be an improvement. Well, it was worth a shot! I appreciate you checking those benchmarks and I leave it in your hands to merge or close 🙇

chrisduerr:
> Huh ... fascinating. So it seems like it might not be a regression but it also seems like it might not be an improvement.

It seems very unlikely to me that changes to this part of the code would have a significant performance impact, especially in these benchmarks. Since the terminal lock is already dropped at this point and it shouldn't cause the PTY to slow down at all. The main thing this might change is the time it takes to render and we have a fair bit of capacity left over in all our benchmarks for rendering.

Issue #4330, opened 2020-10-22
Add optional to imports


EdenEast:
First off I would like to say thanks to @chrisduerr for pr #4138 (was very excited for this)

I started to look at converting my configuration in my dotfiles to use this new feature. I would like to have a base
configuration saved in my dotfiles so that all my different machines have a similar configuration. Although trying to
setup machine specific configuration I ran into a little issue.

In my dotfiles I have something like this in my main configuration entry point `alacritty.yml`:

```yaml
import:
  -  ~/.config/alacritty/visual.yml
  -  ~/.config/alacritty/binding.yml
  -  ~/.config/alacritty/operation.yml

  # Source local machine specific configuration if required
  -  ~/.local/share/alacritty/config.yml
```

On some machines this file may not exist causing an error. I would like to mark this import as `optional`. If the import
does not exist and the import is marked as optional then it would skip trying to source it.

An example of what that would look like:

```yaml
import:
  -  ~/.config/alacritty/visual.yml
  -  ~/.config/alacritty/binding.yml
  -  ~/.config/alacritty/operation.yml
  -  path: ~/.local/share/alacritty/config.yml
     opt: true
```

This would allow for two different formats for `import`. If an import is not explicitly marked as `opt: true` then it
will be assumed false and functionality will remain the same.

I have been moking up this feature in a branch and would be interested in contributing a pr for this issue. Please let me
know your thoughts.

chrisduerr:
What are your arguments for making certain paths optional, rather than just making them all optional?

Also why should this be fixed inside of Alacritty instead of just touching the file itself to make sure every file is present on all systems?

EdenEast:
Being able to distinguish between optional and required imports would declare indent. I am defining what imports I expect
to be imported and others that I have defined as optional and not required. If a required import does not exist I would
like to be informed about it. An example would be I messed up a rebase in my configuration and there are missing files
that I expect to be there.

As for touching a file that would also work. The ability to clone and launch without extra steps would be nice to have.

chrisduerr:
> Being able to distinguish between optional and required imports would declare indent. I am defining what imports I expect
to be imported and others that I have defined as optional and not required. If a required import does not exist I would
like to be informed about it.

This makes sense, but it could probably make it significantly harder to understand this configuration. I'd like to see how this nicely documented to make sure people are aware of how it works.

> As for touching a file that would also work. The ability to clone and launch without extra steps would be nice to have.

[...]

PR #4332, opened 2020-10-22
Add optional configuration imports

EdenEast:
# Optional imports

This adds the ability to mark imports as `optional`. Optional imports
will be skip loading if they do not exist. If an import is not
explicitly marked as `optional`, then there will be an error raised like
normal.

An optional import is defined as a `path` and `optional` keys. Example:

```yaml
import:
  - /normal/path/to/config.yml
  - path: /path/to/config.yml
    optional: true
```

If both these files do not exist alacritty will raise an error on the
/normal/path/to/config.yml file.

Fixes: #4330

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.7.0-dev
 
+### Changed
+
+- Providing non existing paths in `import` won't result in error. Importing from these configs will be skipped
+
 ### Fixed
 
 - Wide characters sometimes being cut off
@@ -47,7 +51,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Reduce memory usage by up to at least 30% with a full scrollback buffer
 - The number of zerowidth characters per cell is no longer limited to 5
 - `SpawnNewInstance` is now using the working directory of the terminal foreground process on macOS
-- Configuration `import` paths are now optional and will skip importing if path does not exist.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Reduce memory usage by up to at least 30% with a full scrollback buffer
 - The number of zerowidth characters per cell is no longer limited to 5
 - `SpawnNewInstance` is now using the working directory of the terminal foreground process on macOS
+- Configuration `import` paths are now optional and will skip importing if path does not exist.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Providing non existing paths in `import` won't result in error. Importing from these configs will be skipped
+- Nonexistent config imports are ignored instead of raising an error
 
 ### Fixed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -4,7 +4,8 @@
 #
 # These configuration files will be loaded in order, replacing values in files
 # loaded earlier with those loaded later in the chain. The file itself will
-# always be loaded last.
+# always be loaded last. If an import path cannot be found it will be skipped.
+# Skipped files will be logged at info level.
 #import:
 #  - /path/to/alacritty.yml
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -5,7 +5,6 @@
 # These configuration files will be loaded in order, replacing values in files
 # loaded earlier with those loaded later in the chain. The file itself will
 # always be loaded last. If an import path cannot be found it will be skipped.
-# Skipped files will be logged at info level.
 #import:
 #  - /path/to/alacritty.yml
 
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -66,7 +66,7 @@ impl Display for Error {
             Error::NotFound => write!(f, "Unable to locate config file"),
             Error::ReadingEnvHome(err) => {
                 write!(f, "Unable to read $HOME environment variable: {}", err)
-            },
+            }
             Error::Io(err) => write!(f, "Error reading config file: {}", err),
             Error::Yaml(err) => write!(f, "Problem with config: {}", err),
         }
@@ -137,7 +137,7 @@ fn load_from(path: &PathBuf, cli_config: Value) -> Result<Config> {
         Err(err) => {
             error!(target: LOG_TARGET_CONFIG, "Unable to load config {:?}: {}", path, err);
             Err(err)
-        },
+        }
     }
 }
 
@@ -183,7 +183,7 @@ fn parse_config(
             } else {
                 return Err(Error::Yaml(error));
             }
-        },
+        }
     };
 
     // Merge config with imports.
@@ -198,7 +198,7 @@ fn load_imports(config: &Value, config_paths: &mut Vec<PathBuf>, recursion_limit
         Some(_) => {
             error!(target: LOG_TARGET_CONFIG, "Invalid import type: expected a sequence");
             return Value::Null;
-        },
+        }
         None => return Value::Null,
     };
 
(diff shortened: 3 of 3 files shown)


chrisduerr:
Before doing any sort of review, I'd be curious what @nixpulvis / @kchibisov think about this (see more discussion in the issue #4330).

I generally don't see too much of a reason why we should outright block it, but there might be alternative solutions to the problem that should also be evaluated.

EdenEast:
Apologies if I was too quick with making this PR. I made this pr as I had the changes in a fork and made it as an
example implementation for issue #4330. Yes I agree this should not be reviewed until further discussion on the issue.

chrisduerr:
> Apologies if I was too quick with making this PR

No problem, it's always good to have some reference code to look at. Of course it comes with a risk of wasting time when big changes are requested, but I'd assume you were aware of that before creating the PR.

kchibisov:
>Before doing any sort of review, I'd be curious what @nixpulvis / @kchibisov think about this (see more discussion in the issue #4330).

IIRC I was suggesting to add optional for everything when this feature was implemented, I'm not sure if it's recorded anywhere though. I do think that everything as optional is a good idea and it's simpler. If your config is missing unintentionally, I bet you can notice it without throwing a warning/error on the users face.

kchibisov:
@EdenEast Just to clarify here. The way forward here would be to make all imports optional, and maybe log missing ones into info log or debug log level.

EdenEast:
Alright I can update this pr to reflect this.

alacritty/src/config/mod.rs
```diff
@@ -222,6 +222,11 @@ fn load_imports(config: &Value, config_paths: &mut Vec<PathBuf>, recursion_limit
             },
         };
 
+        if !path.exists() {
+            info!(target: LOG_TARGET_CONFIG, "Skip importing: '{}' not found", path.display());
```
kchibisov, line 226: I'm not sure whether it should be filtered here, maybe we should do it earlier? I'm not sure if there's a reason to watch for config file that can appear after some period of time.

Also, I don't quite like the way this `info` message looks, maybe it should follow our message about loaded config and be multiline, I'd suggest something like that:

```rust
info!("Skipping importing config; not found:");
info!("  {:?}", path);
```

alacritty.yml
```diff
@@ -4,7 +4,8 @@
 #
 # These configuration files will be loaded in order, replacing values in files
 # loaded earlier with those loaded later in the chain. The file itself will
-# always be loaded last.
+# always be loaded last. If an import path cannot be found it will be skipped.
+# Skipped files will be logged at info level.
```
kchibisov, line 8: I'm that we should mention such that something will be printed into logs, I think I'd remove this part.

kchibisov (changes requested):
You'll also need a changelog entry.

Also, the place where you log about potential failed imports is a bit wrong, you either do that before updating `config_paths` vector or filter out patch you print that you've imported, since you haven't imported them, if that makes sense.

Also, you should explicitly request a review, so we won't forget, since github is acting weirdly sometimes and not notifies via mails(at least me) about force pushes.

alacritty.yml
EdenEast: Ya I was not sure about the log info part as well. I added it to be extra explicit to the user. It will be removed.

alacritty/src/config/mod.rs
EdenEast: I filtered paths here as this is right after they are parsed from config and before they are added to the config list. If we would want to support watching these files then I would move the check to after the path is added to the config_paths list and before attempting to read its file contents. 

Do we want to support watching for files that do not exist? I played with the idea in `monitor::watch()`. Bypassing the `fs::canonicalize` for testing it did work when an import file was touched and caught with the notify watcher. Apparently renames are currently not caught at least on windows (which I was testing). [165](https://github.com/notify-rs/notify/issues/165) and [166](https://github.com/notify-rs/notify/pull/166).

The check for path exist would depend on how config watchers should be handled for files that don't exist yet.

My option, it would be interesting to support but not worth the change. You can also touch/save any other alacrity config file and it will reload the config the new file will be caught. 

As for the info message I can change it to your suggestion.

kchibisov: Ah, yeah, you're right, it's in the current place. So just changing a message should be fine. I forgot that `parse_config` calls `load_imports`, hence recursion.

kchibisov (changes requested):

alacritty/src/config/mod.rs
```diff
@@ -222,6 +222,12 @@ fn load_imports(config: &Value, config_paths: &mut Vec<PathBuf>, recursion_limit
             },
         };
 
+        if !path.exists() {
+            info!(target: LOG_TARGET_CONFIG, "Skipping importing config; not found:");
+            info!(target: LOG_TARGET_CONFIG, "  {}", path.display());
```
kchibisov, line 227:
```suggestion
            info!(target: LOG_TARGET_CONFIG, "  {:?}", path.display());
```

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Reduce memory usage by up to at least 30% with a full scrollback buffer
 - The number of zerowidth characters per cell is no longer limited to 5
 - `SpawnNewInstance` is now using the working directory of the terminal foreground process on macOS
+- Configuration `import` paths are now optional and will skip importing if path does not exist.
```
kchibisov, line 50: The changelog entry should be in `0.7.0-dev` section. Also maybe something like that will work better.

```suggestion
- Providing non existing paths in `import` won't result in error and importing from these configs will be skipped
```

kchibisov (changes requested):

alacritty/src/config/mod.rs
EdenEast: `path.display()` returns an object that Implements the `Display` trait. Since it implements the display trait shouldn't we be calling the display format operator?

kchibisov: The point is to automatically quote it, we do the same for regular config printing, so it's consistent.

EdenEast: Ah ok I understand will do.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.7.0-dev
 
+### Changed
+
+- Providing non existing paths in `import` won't result in error. Importing from these configs will be skipped
```
kchibisov, line 12: I don't really like splitting sentence into 2 in changelog, and we never had something like that before. 

```suggestion
- Providing non existing paths in `import` won't result in error, instead importing configs from them will be skipped
```

@chrisduerr do you any suggestion on how to form this entry?

kchibisov (changes requested):

CHANGELOG.md
chrisduerr: We don't need a changelog entry, this was never released?

kchibisov: We need, it'll go into 0.7.0, however imports will be in 0.6.0.

chrisduerr: Oh that's a good point. I'm not sure why I was just assuming this would be included in 0.6.0.

But I suppose it's not an issue with the existing imports system, so not having them optional should work perfectly fine.

Here would be my suggestion then:

```
 - Nonexistent config imports are skipped instead of raising an error
```

kchibisov: We can cherry pick it into 0.6.0 if it's fine, since import thing is a new, so a change for it makes sense for backporting, I guess.

chrisduerr: Fine either way, probably easier to just get 0.6.0 out there. We'll need a follow up on imports because of home expansion anyways before it becomes **really** useful. So bundling that into 0.7.0 seems fine.

kchibisov: Ok, let's move it into 0.7.0.

alacritty/src/config/mod.rs
```diff
@@ -183,7 +183,7 @@ fn parse_config(
             } else {
                 return Err(Error::Yaml(error));
             }
-        },
+        }
```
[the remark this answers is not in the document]
EdenEast: Ah my other machine is pointing to stable rustfmt will fix this.

kchibisov (approved):

EdenEast:
Thank you @kchibisov and @chrisduerr for you help. Now looking forward to resolving the home expansion to make imports awesome.

PR #4345, opened 2020-10-24
Remove rustc_tools_util dependency

chrisduerr:
Since our usage of the rustc_tools_util crate is so trivial, it seems
like we should be able to just inline it directly into Alacritty.

It's a very well trusted crate, being hosted directly by rust-lang and
it does not pull in any other dependencies, but having a dependency for
just 6 lines of code seems a bit extreme.


kchibisov (approved):
Thanks.

Issue #4237, opened 2020-09-16
extending double-clicked selection with shift+RMB deselects part of original selection


p1otr:
Hi,

I'm testing Alacritty as an alternative terminal for URxvt (which is not Wayland ready yet) and I like it so far. Thanks!

There's one¹ annoyance, though: when I double or triple click on some text and later want to extend selection with a right mouse button click - original selection starts on the character I double/triple clicked instead of the whole word.

F.e. if I double click on "a" from "bar" in "foo bar baz" - "bar" is selected as expected.
When I right-click on "f" - "foo ba" is selected instead of expected "foo bar".
Right-clicking later on "z" selects "r baz" instead of extending previous extension (i.e. I expect to have "foo bar baz" selected after second click).

Please let me know if it's configurable and just didn't find the right switch.
Thanks again for writing Alacritty!

[¹] well… two, but from what I know there's nothing you can do about bell (`echo -e '\a'`) not setting urgent flag for windows on Wayland - as there no such thing there yet

### System

OS: Linux (Debian Sid)
Version: 0.6.0-dev (1bf40a8)
Linux/BSD: Wayland, Sway

chrisduerr:
If you want to semantically extend the selection, you need to double right click. If you extend it with just a single click, it will be converted to a normal selection.

p1otr:
that's better than what I was doing before, but when I extend selection - I usually want to fine tune it so another double click is not always optimal (it selects more than I want). Thanks for the tip, though

kchibisov:
I see. I'm not sure if that's behavior we want to have, since the goal is to change selections on a fly wrt original selection. So the span from triple click selection will include the entire line, so single right click selection will include that line + some minor part of it. What do you think @chrisduerr ?

chrisduerr:
> What do you think @chrisduerr ?

I think that if you don't want a semantic selection, you probably shouldn't start with one?

kchibisov:
>I think that if you don't want a semantic selection, you probably shouldn't start with one?

I guess, but I feel like the way our expansion works isn't very intuitive, since you can change originally selected things, which isn't desired and may require users some time to understand how it works? So the way expanding part works is fine, however the selected region you've done at first should be stored in expanded mode? On the other side the way it's done that way right now is due to Vi mode, where those changes are following vim.

freshprince:
[...]

PR #4346, opened 2020-10-25
Add touchscreen support

### LINKED ISSUE #3671: Add touchscreen support
It shouldn't be that hard to use winit's touchscreen events and translate them to our inner `mouse` struct, at least some basic interaction. I'm not sure how to emulate RMB/MiddleClick for mouse mode though.

**Edit:** Apparently for proper touchscreen support it would also be necessary to enable using it with Wayland CSDs.

ch-sy:
Fixes #3671.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Touchscreen support for scrolling, zooming and clicking
+- Touchscreen support for scrolling, zooming, selecting and clicking
 - Secondary device attributes escape (`CSI > 0 c`)
 - Support for colon separated SGR 38/48
 - New Ctrl+C binding to cancel search and leave vi mode
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Touchscreen support for scrolling, zooming and clicking
 - Secondary device attributes escape (`CSI > 0 c`)
 - Support for colon separated SGR 38/48
 - New Ctrl+C binding to cancel search and leave vi mode
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -15,7 +15,7 @@ use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
-use glutin::dpi::PhysicalSize;
+use glutin::dpi::{PhysicalSize, PhysicalPosition};
 use glutin::event::{ElementState, Event as GlutinEvent, ModifiersState, MouseButton, WindowEvent};
 use glutin::event_loop::{ControlFlow, EventLoop, EventLoopProxy, EventLoopWindowTarget};
 use glutin::platform::desktop::EventLoopExtDesktop;
@@ -664,7 +664,7 @@ pub enum ClickState {
     TripleClick,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Copy, Clone)]
 pub struct TouchFinger {
     pub start_x: f64,
     pub start_y: f64,
@@ -695,15 +695,16 @@ impl Default for Touchscreen {
 }
 
 impl Touchscreen {
-    pub fn mean_finger_position(&self) -> (f64, f64) {
+    pub fn mean_finger_position(&self) -> PhysicalPosition<f64> {
         let mut mean_position = (0.0, 0.0);
         for finger in self.fingers.values() {
             mean_position.0 += finger.x;
             mean_position.1 += finger.y;
         }
         let finger_count = self.fingers.len() as f64;
-        (mean_position.0 / finger_count, mean_position.1 / finger_count)
+        (mean_position.0 / finger_count, mean_position.1 / finger_count).into()
     }
+
     pub fn mean_finger_distance(&self) -> f64 {
         let mut mean_distance = 0.0;
         let mut pair_count = 0;
@@ -715,13 +716,38 @@ impl Touchscreen {
         }
         mean_distance / pair_count as f64
     }
-    pub fn is_zooming(&self) -> bool {
-        let min_finger_count_to_zoom = 2;
-        self.fingers.len() >= min_finger_count_to_zoom
+
+    pub fn check_for_gestures(&mut self, finger_id: u64) {
+        if let Some(&finger) = self.fingers.get(&finger_id) {
+            // Allow a click despite wobbling fingers.
+            const TOLERANCE_RADIUS: f64 = 16.0;
+            let distance_to_start = (finger.x - finger.start_x).hypot(finger.y - finger.start_y);
+            if distance_to_start > TOLERANCE_RADIUS {
+                 self.is_gesture = true;
+            }
+        }
     }
+
+    pub fn is_scrolling(&self) -> bool {
+        // Set it to `self.fingers.len() >= 1` to allow scrolling while zooming
+        self.fingers.len() == 1
+    }
+
+    pub fn is_zooming(&self) -> bool {
+        const MIN_FINGER_COUNT_TO_ZOOM: usize = 2;
+        self.fingers.len() >= MIN_FINGER_COUNT_TO_ZOOM
+    }
+
     pub fn init_zooming(&mut self) {
         self.start_finger_distance = self.mean_finger_distance();
         self.relative_zoom_level = 0;
+    }
+
+    pub fn set_finger_position(&mut self, id: u64, location: &PhysicalPosition<f64>) {
+        if let Some(mut finger) = self.fingers.get_mut(&id) {
+            finger.x = location.x;
+            finger.y = location.y;
+        }
     }
 }
 
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -683,7 +683,7 @@ pub struct Mouse {
     pub touch_finger: HashMap<u64, TouchFinger>,
     pub touch_mean_y: f64,
     pub touch_start_finger_distance: f64,
-    pub touch_relative_zoom_level: f64,
+    pub touch_relative_zoom_level: i64,
     pub touch_is_gesture: bool,
 }
 
@@ -708,7 +708,7 @@ impl Default for Mouse {
             touch_finger: HashMap::new(),
             touch_mean_y: 0.0,
             touch_start_finger_distance: 1.0,
-            touch_relative_zoom_level: 0.0,
+            touch_relative_zoom_level: 0,
             touch_is_gesture: false,
         }
     }
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -15,7 +15,7 @@ use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
-use glutin::dpi::{PhysicalSize, PhysicalPosition};
+use glutin::dpi::{PhysicalPosition, PhysicalSize};
 use glutin::event::{ElementState, Event as GlutinEvent, ModifiersState, MouseButton, WindowEvent};
 use glutin::event_loop::{ControlFlow, EventLoop, EventLoopProxy, EventLoopWindowTarget};
 use glutin::platform::desktop::EventLoopExtDesktop;
@@ -680,10 +680,7 @@ pub enum Gesture {
     None,
     Clicking,
     Scrolling,
-    Zooming {
-        start_finger_distance: f64,
-        old_zoom: i64,
-    },
+    Zooming { start_finger_distance: f64, old_zoom: i64 },
     Selecting,
 }
 
@@ -695,10 +692,7 @@ pub struct Touchscreen {
 
 impl Default for Touchscreen {
     fn default() -> Touchscreen {
-        Touchscreen {
-            fingers: HashMap::new(),
-            gesture: Gesture::None,
-        }
+        Touchscreen { fingers: HashMap::new(), gesture: Gesture::None }
     }
 }
 
@@ -716,24 +710,28 @@ impl Touchscreen {
     }
 
     pub fn set_finger(&mut self, id: u64, location: &PhysicalPosition<f64>) -> TouchFinger {
-        *(self.fingers.entry(id).and_modify(|finger| {
-            finger.delta_x = location.x - finger.x;
-            finger.delta_y = location.y - finger.y;
-            finger.x = location.x;
-            finger.y = location.y;
-        }).or_insert(TouchFinger {
-            start_x: location.x,
-            start_y: location.y,
-            start_timestamp: Instant::now(),
-            delta_x: 0.0,
-            delta_y: 0.0,
-            x: location.x,
-            y: location.y,
-        }))
+        *(self
+            .fingers
+            .entry(id)
+            .and_modify(|finger| {
+                finger.delta_x = location.x - finger.x;
+                finger.delta_y = location.y - finger.y;
+                finger.x = location.x;
+                finger.y = location.y;
+            })
+            .or_insert(TouchFinger {
+                start_x: location.x,
+                start_y: location.y,
+                start_timestamp: Instant::now(),
+                delta_x: 0.0,
+                delta_y: 0.0,
+                x: location.x,
+                y: location.y,
+            }))
     }
 
     pub fn new_zoom_gesture(&self) -> Gesture {
-        Gesture::Zooming {start_finger_distance: self.mean_finger_distance(), old_zoom: 0}
+        Gesture::Zooming { start_finger_distance: self.mean_finger_distance(), old_zoom: 0 }
     }
 }
 
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -668,6 +668,7 @@ pub enum ClickState {
 pub struct TouchFinger {
     pub start_x: f64,
     pub start_y: f64,
+    pub start_timestamp: Instant,
     pub x: f64,
     pub y: f64,
 }
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -669,47 +669,45 @@ pub struct TouchFinger {
     pub start_x: f64,
     pub start_y: f64,
     pub start_timestamp: Instant,
+    pub delta_x: f64,
+    pub delta_y: f64,
     pub x: f64,
     pub y: f64,
+}
+
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum Gesture {
+    None,
+    Clicking,
+    Scrolling,
+    Zooming {
+        start_finger_distance: f64,
+        old_zoom: i64,
+    },
+    Selecting,
 }
 
 #[derive(Debug)]
 pub struct Touchscreen {
     pub fingers: HashMap<u64, TouchFinger>,
-    pub mean_y: f64,
-    pub start_finger_distance: f64,
-    pub relative_zoom_level: i64,
-    pub is_gesture: bool,
+    pub gesture: Gesture,
 }
 
 impl Default for Touchscreen {
     fn default() -> Touchscreen {
         Touchscreen {
             fingers: HashMap::new(),
-            mean_y: 0.0,
-            start_finger_distance: 1.0,
-            relative_zoom_level: 0,
-            is_gesture: false,
+            gesture: Gesture::None,
         }
     }
 }
 
 impl Touchscreen {
-    pub fn mean_finger_position(&self) -> PhysicalPosition<f64> {
-        let mut mean_position = (0.0, 0.0);
-        for finger in self.fingers.values() {
-            mean_position.0 += finger.x;
-            mean_position.1 += finger.y;
-        }
-        let finger_count = self.fingers.len() as f64;
-        (mean_position.0 / finger_count, mean_position.1 / finger_count).into()
-    }
-
     pub fn mean_finger_distance(&self) -> f64 {
         let mut mean_distance = 0.0;
         let mut pair_count = 0;
         for (i, finger_a) in self.fingers.values().enumerate() {
-            for finger_b in self.fingers.values().skip(i+1) {
+            for finger_b in self.fingers.values().skip(i + 1) {
                 pair_count += 1;
                 mean_distance += (finger_a.x - finger_b.x).hypot(finger_a.y - finger_b.y);
             }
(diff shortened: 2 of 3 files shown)


chrisduerr (changes requested):

chrisduerr:
@kchibisov / @nixpulvis does anyone of you have touch devices and is willing to maintain this in the future? I do not have anything and am definitely not going to change that any time soon. Unless we have someone capable of maintaining this, it cannot be merged.

kchibisov:
>@kchibisov / @nixpulvis does anyone of you have touch devices and is willing to maintain this in the future? I do not have anything and am definitely not going to change that any time soon. Unless we have someone capable of maintaining this, it cannot be merged.

I don't have anything, so can't really test. Maybe at some point there will be some tool that emulates touchscreen for wayland, but there's nothing I'm aware of right now afaics.

ch-sy:
I am currently making changes based on your comments. I hope it's okay to improve the user experience as it's still a bit experimental. I couldn't find a terminal with good touchscreen support for reference.

> > @kchibisov / @nixpulvis does anyone of you have touch devices and is willing to maintain this in the future? I do not have anything and am definitely not going to change that any time soon. Unless we have someone capable of maintaining this, it cannot be merged.
> 
> I don't have anything, so can't really test. Maybe at some point there will be some tool that emulates touchscreen for wayland, but there's nothing I'm aware of right now afaics.

You could replace the corresponding mouse events so that it calls the touch function. Should be enough for testing the basic stuff, but unfortunately not for things like the zoom.

For example:
```rust
WindowEvent::MouseInput { state, device_id,.. } => {
    processor.ctx.window.set_mouse_visible(true);
    match state {
        ElementState::Pressed => processor.touch_input(Touch {
            device_id,
            phase: glutin::event::TouchPhase::Started,
            location: (processor.ctx.mouse.x as f64, processor.ctx.mouse.y as f64).into(),
            force: None,
            id: 0
        }),
        ElementState::Released => processor.touch_input(Touch {
            device_id,
            phase: glutin::event::TouchPhase::Ended,
            location: (processor.ctx.mouse.x as f64, processor.ctx.mouse.y as f64).into(),
            force: None,
            id: 0
        }),
    };
},
```
and
```rust
WindowEvent::CursorMoved { position, device_id,.. } => {
    processor.ctx.mouse.x = position.x as usize;
    processor.ctx.mouse.y = position.y as usize;
    processor.touch_input(Touch {
        device_id,
        phase: glutin::event::TouchPhase::Moved,
        location: position,
        force: None,
        id: 0
    })
},
```
About the `scroll_terminal` function. Which unit is used here? According to the name, it would be pixels. However, it is three times the size, which is also inconsistent with the DPI.

chrisduerr:
> Should be enough for testing the basic stuff, but unfortunately not for things like the zoom.

Exactly, which makes it impossible for me to fully test that this PR behaves reasonably at all. It also makes it impossible for me to make sure it works in the future or fix things when they break.

I'm not against adding this, the idea of having a terminal that works with touchscreens with just a relatively small patch seems like a good opportunity, but I cannot merge something nobody can test or maintain.

> About the scroll_terminal function. Which unit is used here? According to the name, it would be pixels. However, it is three times the size, which is also inconsistent with the DPI.

It is in pixels, but the value is multiplied by the scrolling multiplier set in the config (you can see that by reading the function, I think it should be relatively clear).

nixpulvis:
> @kchibisov / @nixpulvis does anyone of you have touch devices and is willing to maintain this in the future? I do not have anything and am definitely not going to change that any time soon. Unless we have someone capable of maintaining this, it cannot be merged.

I've got a Thinkpad X1 Tablet (not the strange folding one) lying around here I'm happy to test with.

I've built this now and taken a very quick first look at it. It seems to work for clicking URLs, scrolling and zooming as designed. There's perhaps some refinement to get the scrolling distance to match my finger's, but I frankly don't mind it being a bit faster, especially given that I don't expect to have inertial scrolling implemented.

At some point this coming week I'll make a pass through the code and test each new functionality.

chrisduerr:
> At some point this coming week I'll make a pass through the code and test each new functionality.

Thanks. If you can verify functionality, I can check the code and make sure everything's good in that regard. If I have anything specific for you to test I'll let you know.

chrisduerr:
@ch-sy Please request a review from me once everything should be in a mergeable state.

ch-sy:
I am satisfied with the code for my part. If inertial scrolling is implemented for the touchpad later, then it should be also relatively easy to implement it for the touchscreen, by using the same infrastructure. I have now also implemented the selection of text.

ch-sy:
I am grateful for your feedback. Discussions are essential for high quality, so I have no problem with them and I look forward to learn new things.

>There is precedence for mapping Scrolling to a two finger drag, since that's how it works on most trackpads, but I agree that a single finger drag is most common in other applications

While one-finger gestures on the trackpad are necessary for moving the mouse cursor, the finger position on the touchscreen can be directly mapped as the mouse position and thus enables more convenient control options. This is the reason why only one finger is required to scroll with the touchscreen and I think scrolling with the thumb should not be prevented as it is standard in almost every application with touchscreen control and is also my preference.

Zooming and scrolling with two or more fingers at the same time would theoretically not be mutually exclusive and would also be an option. This is the behavior in all major browsers and was possible in my first commit. The advantage of allowing both at the same time is the more diverse control options and it doesn't matter whether the user prefers scrolling with one, two or even more fingers. The downside is that it can wobble a bit when zooming or scrolling and the fact that the movements are step by step can also be irritating.

>I'm wondering if the implementation of selection is correct. It seems a bit strange to use single finger single touch movement for both scrolling and selection. Applications like Firefox seem to perform selection on double tap and drag, while others somewhat strangely use two finger drag. I'm not at all sure what is most standard or best, but I don't think this kind of overloading as implemented today is what we should do. I particularly don't like how selection works only upon the X axis to start with. And ideally, we'd allow selections of all four kinds: Simple, Semantic, Lines, and even Block.

The horizontal movement of a finger is treated differently in each application. Most often nothing. For example:
* Firefox: Nothing or horizontal scrolling if applicable.
* Microsoft Edge: Previous or next page
* Notepad: Selecting
* Microsofts command Line: Selecting

In my opinion selecting makes here the most sense, as there is no horizontal scrolling in a terminal emulator and in some terminal applications it can then be interpreted as dragging things for example. In the current implementation, scrolling gestures are preferred registered because they are used more frequently and with less precision, and text is therefore marked less accidentally. What I might also have to say is that if you start the selection with a horizontal movement and add another finger, it results in undefined behavior. Depending on the operating system you will be surprised what happens. Here we should determine what should happen in such a case.
>For example, I expect a user may come from iOS expecting some kind of copy/paste interface. We have a good answer for copy, since simply creating a selection updates a clipboard, however I wouldn't have a touch-based solution for paste in mind. I'm personally not bothered by this, but it is noticeable lacking. Since I wouldn't want to allow popups or modals of any kind here, there are many existing implementations which are already incompatible (like iOS's). If we find a clever, and elegant way to support things without the clutter of more visual panels, I'd be interested in what other features we can support are as well.

Right clicking for example is currently handled by holding more than 300ms. That should perhaps be a config entry. I was surprised that the behaviour of right clicking is selecting the text from the position of the last left click to the position of the right click.
Perhaps this could be replaced by inserting, since otherwise there are several options for selecting but none for inserting. This would be the same behavior as in the Windows command line, which has proven to be the most touch-friendly so far.
The users could possibly be asked here how often they use the right-click to mark and whether a paste function would not make more sense here. Shift and click could then be used for selecting without holding a mouse/trackpad button like most applications.

>I think it's pretty exciting to think about the possibilities for touch input in Alacritty, but from what I've seen in iOS specifically, I  want to be very careful to avoid adding things which don't feel right. We should be conservative in how we approach this, since these will become learned fundamental interactions with our application.

I am open to suggestions with further gestures and can also implement them, but at the moment I cannot think of any others that really make sense. I hope we can agree that it shouldn't be a magic show. 😅

With `Option<Gesture>` I agree with you that this is more understandable and I will change that.

chrisduerr:
> Zooming and scrolling with two or more fingers at the same time would theoretically not be mutually exclusive and would also be an option. This is the behavior in all major browsers and was possible in my first commit. The advantage of allowing both at the same time is the more diverse control options and it doesn't matter whether the user prefers scrolling with one, two or even more fingers. The downside is that it can wobble a bit when zooming or scrolling and the fact that the movements are step by step can also be irritating.

I think the complexity combined with a little bit of a strange behavior make this not worth it. Giving the user the option to scroll with 19 fingers is all nice and well, but we really don't need to allow that. One consistent and sensible way to do so is completely sufficient.

> In my opinion selecting makes here the most sense, as there is no horizontal scrolling in a terminal emulator and in some terminal applications it can then be interpreted as dragging things for example.

I strongly disagree with this. Having horizontal motion map to selection but vertical movement map to scrolling is incredibly inconsistent and borderline unusable. Being able to select with a single finger horizontally but not vertically is absolutely unacceptable.

> Right clicking for example is currently handled by holding more than 300ms.

There's no point in doing this. Nobody will ever figure out this is the case and people accidentally running into it are just gonna get annoyed. It's better to not have any right click at all than strange left right click transitions like this after some "magical" interval with zero feedback to the user wrt which click he just performed.

>  Perhaps this could be replaced by inserting, since otherwise there are several options for selecting but none for inserting. This would be the same behavior as in the Windows command line, which has proven to be the most touch-friendly so far.

Just no.

ch-sy:
Thank you for your feedback.

> > Zooming and scrolling with two or more fingers at the same time would theoretically not be mutually exclusive and would also be an option. This is the behavior in all major browsers and was possible in my first commit. The advantage of allowing both at the same time is the more diverse control options and it doesn't matter whether the user prefers scrolling with one, two or even more fingers. The downside is that it can wobble a bit when zooming or scrolling and the fact that the movements are step by step can also be irritating.
> 
> I think the complexity combined with a little bit of a strange behavior make this not worth it. Giving the user the option to scroll with 19 fingers is all nice and well, but we really don't need to allow that. One consistent and sensible way to do so is completely sufficient.

Since it would reduce the constraints, the code complexity stays pretty much the same, so the user experience is more important here. In my opinion, limitations can confuse users. But I can get used to it how it is at the moment.

> > In my opinion selecting makes here the most sense, as there is no horizontal scrolling in a terminal emulator and in some terminal applications it can then be interpreted as dragging things for example.
> 
> I strongly disagree with this. Having horizontal motion map to selection but vertical movement map to scrolling is incredibly inconsistent and borderline unusable. Being able to select with a single finger horizontally but not vertically is absolutely unacceptable.

I'm not sure you fully understood how it works, but it is wrong that you cannot select vertically with this method. Please test it with your technical possibilities. It is not uncommon at all that the axes are used for different functions and it should be intuitive enough to get it in a few seconds. It's a tradeoff with scrolling, which is far more important for the touchscreen.

> > Right clicking for example is currently handled by holding more than 300ms.
> 
> There's no point in doing this. Nobody will ever figure out this is the case and people accidentally running into it are just gonna get annoyed. It's better to not have any right click at all than strange left right click transitions like this after some "magical" interval with zero feedback to the user wrt which click he just performed.
> 
> > Perhaps this could be replaced by inserting, since otherwise there are several options for selecting but none for inserting. This would be the same behavior as in the Windows command line, which has proven to be the most touch-friendly so far.
> 
> Just no.

Yes, right-click is pretty useless for me at the moment, as it has an insignificant function assigned. That's the feedback I can say about it. If it is needed as it is at the moment, then I accept that, but will then make my own version.

chrisduerr:
> It is not uncommon at all that the axes are used for different functions and it should be intuitive enough to get it in a few seconds.

That's not common at all and incredibly unintuitive.

ch-sy:
>That's not common at all and incredibly unintuitive.

I think then it makes sense for me to develop it for myself as we have different ideas about good touch controls. 
I wish you all the best!

nixpulvis:
@ch-sy / @chrisduerr 

We're really giving up on this just because we can't agree on a better implementation for selection and scrolling together?

I agree 100% that scrolling is the most important added feature, and I'd love to see this work merged. Why would two fingers for scrolling be a horrible idea? I think that would allow solving all my concerns. Or could you think of a good way to make this configurable, since users will absolutely have some preconceived ideas about how they want this to work?

Getting things right often takes discussion and the arguments need to be sound. We are pretty conservative here about adding new things, so I understand it sounds like we're being overly resistive or obstinate, but as I said in my initial response, once we set things, we really don't want to be forced to change them later.

I hope we can take the time to continue to think this through. But if not, thanks for at least getting the ball moving. Perhaps someone else will come along and use this as the starting point down the road.

chrisduerr:
> Perhaps someone else will come along and use this as the starting point down the road.

That's certainly an option. Generally it's just not a good idea to do something in a way completely different from any other application just because it fits the usecase of one person. Touchscreen interfaces with mobile phones have been fleshed out pretty well over the years, there's nothing we need to reinvent here.

nixpulvis:
So then selection should be a press and hold. I'm completely OK with this, but I'm not sure how it would be possible to get the other kinds of selections with that scheme. 

I'll also say that mobile touch interfaces have grown very unsatisfying in a few ways. Most notably for me at the moment is the conflation of "tap to select and edit red underlined word" and "tap to move cursor". But I don't expect we should need the former to ever appear in Alacritty. It is also a bit strange that tap and hold selects a word, but then adjustments to the selection are by character.

chrisduerr:
> So then selection should be a press and hold. I'm completely OK with this, but I'm not sure how it would be possible to get the other kinds of selections with that scheme.

Press and hold, or double press. Once the selection exists, you can drag the ends. Clicking anywhere else will cancel the selection. 

> I'll also say that mobile touch interfaces have grown very unsatisfying in a few ways. Most notably for me at the moment is the conflation of "tap to select and edit red underlined word" and "tap to move cursor". But I don't expect we should need the former to ever appear in Alacritty. It is also a bit strange that tap and hold selects a word, but then adjustments to the selection are by character.

Not familiar with what you're talking about here.

nixpulvis:
I like press and hold personally. I can even imagine two finger press and hold performing a box selection, since you can make neat rectangles with your two fingers. That leaves a three finger press for line selection, which feels really natural to me (it's generally going to be in a line, and triple click is already line selection with the pointer).

Exactly how to handle simple and semantic together is a bit of a quandary to me. I think it makes sense to expand the semantic word under the initial press and hold, like on iOS, but after that it's less clear what we can do. This is somewhat related to #4237.

chrisduerr:
> I can even imagine two finger press and hold performing a box selection, since you can make neat rectangles with your two fingers. That leaves a three finger press for line selection, which feels really natural to me (it's generally going to be in a line, and triple click is already line selection with the pointer).

No need to overcomplicate it. We should go with the basics first if we ever want to support this.

> Exactly how to handle simple and semantic together is a bit of a quandary to me. I think it makes sense to expand the semantic word under the initial press and hold, like on iOS, but after that it's less clear what we can do. This is somewhat related to #4237.

I wouldn't worry too much about semantic/line/box selection. Those generally are not super standard on mobile/touch devices opposed to mouse where they basically always are supported. It's usually double and long press semantically select the current word and then you can extend on a character level from there.

ch-sy:
Ok, I'll try to unravel it after going through the arguments again and getting feedback from others.

The right click can be removed as it is confusing with no visual feedback and has no important function.

Scrolling with one finger is essential. The people I interviewed had preferences for scrolling with one as well as with two or more fingers. They also agreed that you should be given a choice. There are also other problems with selecting with one finger:
- Scrolling with two fingers is quickly registered as a selection. The fingers almost never come to the screen at the same time. So while the first finger is already moving on the screen, you have no certainty about another finger. If scrolling is accepted afterwards, something has already been selected incorrectly. To fix that and the movement, you definitely need a lot of code here. The reverse case is also unclear if a finger is removed while scrolling.
- Scrolling with two fingers is in some cases much less comfortable. Scrolling with just the thumb is not possible. If you are sitting in the laptop position, you have to bring your hand very close to your body, which makes it difficult to steer with two fingers. With the thumb it becomes possible to grasp the screen and thus to keep the hand further away from the body. This allows control of the main touch screen scrolling function, in the best posture.
- I don't know of any app that allows you to select with one finger and scroll with two fingers.

I already mentioned the Windows terminal. This is a good reference as Microsoft wants to make their products as touch friendly as possible in order to sell their convertibles to thousands of people. Except for the ability to scroll with two fingers, it is the same in terms of functionality.
Starting the selection using a horizontal movement may not be widespread. But it does not stand in the way of other important functions and would basically allow selection. Even if it's not intuitive, the nature of the terminals are basically not intuitive at all. If you don't know how they work, it takes archeological work to understand them. They should primarily be efficient and not intuitive. The user base should be considered here.

Configurations are always an option and can make everyone happy. I would, however, limit it to what we cannot agree on in order to keep the complexity down.

A discussion only makes sense if the counter-arguments are dealt with in detail. Would arguments with only "Just no" or "Not at all" convince you? If you can convince me with good arguments as you have already done with some things, I'm also willing to change my mind, otherwise it's just a waste of energy and I stick to what satisfies me the most.

In general, I would be interested in a friendly and objective discussion. If you want to develop solid software, all concerns must be taken into account and must not be dismissed as individual cases, because you cannot know such a thing without major surveys. Each of us represents a part of the user base.

chrisduerr:
> Starting the selection using a horizontal movement may not be widespread. But it does not stand in the way of other important functions and would basically allow selection. Even if it's not intuitive, the nature of the terminals are basically not intuitive at all. If you don't know how they work, it takes archeological work to understand them. They should primarily be efficient and not intuitive. The user base should be considered here.

This is not a good argument for anything at all. "It's not intuitive but it doesn't matter" makes no sense unless you say why it doesn't.

> Would arguments with only "Just no" or "Not at all" convince you?

There's no need to convince anyone. There are issues where it makes sense to discuss and others which are a deal breaker which would simply be a waste of time to even talk about.

ch-sy:
> This is not a good argument for anything at all. "It's not intuitive but it doesn't matter" makes no sense unless you say why it doesn't.

I am not saying that it is not important, but that it is not the most important in the context of other aspects. Effectiveness is more important than intuitiveness in a terminal application.

> There's no need to convince anyone. There are issues where it makes sense to discuss and others which are a deal breaker which would simply be a waste of time to even talk about.

Exactly. Scrolling with at least two or more fingers is a deal breaker for me and other people. You try to downplay it by simply ignoring a lot of arguments or claiming that there are no such people. That poisons any discussion.

The question is, do you just want someone to implement it exactly the way you want it or someone who is deeply involved in how best to implement it?

chrisduerr:
> Exactly. Scrolling with at least two or more fingers is a deal breaker for me and other people. You try to downplay it by simply ignoring a lot of arguments or claiming that there are no such people. That poisons any discussion.

You mean it's a deal breaker for you if it's possible to scroll with just one finger? I never suggested that scrolling should require at least two fingers, quite the opposite.

> The question is, do you just want someone to implement it exactly the way you want it or someone who is deeply involved in how best to implement it?

I want someone to implement it exactly the way I want it of course. I have full responsibility for all code that is merged into Alacritty and will have to maintain it now and in the future. That doesn't mean you can't convince me that some things might be good to consider, but at the end of the day I will never merge anything I don't believe in.

ch-sy:
> You mean it's a deal breaker for you if it's possible to scroll with just one finger? I never suggested that scrolling should require at least two fingers, quite the opposite.

Let us know if you would like to agree with any arguments so that we can have clues to your point of view.

Due to the feedback I am now convinced that it is necessary to be able to scroll with one or more fingers, as this is preferred equally.

I personally do not use the selection function because it is too imprecise for me. However, I would like to give users a way to select text and the current method doesn't get in the way like the other. If you know a better way to select text without getting in the way of scrolling, please suggest it.

> I want someone to implement it exactly the way I want it of course. I have full responsibility for all code that is merged into Alacritty and will have to maintain it now and in the future. That doesn't mean you can't convince me that some things might be good to consider, but at the end of the day I will never merge anything I don't believe in.

OK, one thing leads to another.
If you want to take full responsibility, then you should also take complaints seriously. Otherwise it's pointless.

chrisduerr:
> I personally do not use the selection function because it is too imprecise for me. However, I would like to give users a way to select text and the current method doesn't get in the way like the other. If you know a better way to select text without getting in the way of scrolling, please suggest it.

Both @nixpulvis and I have already suggested other solutions. Holding your finger down on top of a word or double clicking it should start a selection. That's the way it's done in most other applications I've used so far, so it makes sense for us to do the same. There's no conflict with scrolling.

ch-sy:
> Both @nixpulvis and I have already suggested other solutions. Holding your finger down on top of a word or double clicking it should start a selection. That's the way it's done in most other applications I've used so far, so it makes sense for us to do the same. There's no conflict with scrolling.

You are aware that this requires quite a bit of logic? 
In this case, the mouse cannot simply be emulated.

Should the corresponding balloons then also be drawn? 
They indicate that the selection can be dragged and also make it possible to make a selection without covering it with your finger.

Anyway, I would be happy with this method.

Dreomite:
I've been a full-time touchscreen Linux user for 2 years with several years of touchscreen Windows prior to that. I've tested this PR and have read the following conversation as well. Here are some thoughts.

- One-fingered vertical scroll and horizontal selection feels ***awesome***. This has always frustrated me that Termux decided to follow the established Android pattern of selecting text with a long press. Which made selecting text a tedious process compared to what I've been doing on my laptop: quickly selecting text with just 1 finger. @ch-sy's approach feels very intuitive to a person working with terminals on a daily basis.
  
  The established patterns should be followed when it makes sense. Terminals cannot be scrolled horizontally, unlike other applications working with text. And leaving the end users without this dimension of interaction doesn't make sense.

  Double-taps are great for selecting strings with whole words, but it doesn't make them a substitute for proper selection, because they are imprecise. It can be partially fixed in the future, if, theoretically, Alacritty will introduce cursor bubbles, but it will never compete with normal selection in terms of speed. *[Termux flashbacks]*

  I understand the unease about going against some established trends, but I guarantee, the absolute majority of touch screen users will like it, and other projects will follow you as well.
  
- Double touch should select the word right after the finger is pressed, not after it's released. As if you would use your mouse. Then, if the finger is still touching the screen, you can adjust your selection following the finger. Currently, you have to press and immediately move your finger in order to initialize proper selection mode. If you don't do this fast enough, and move the finger sideways, it will do a regular selection. If you move it up, it will scroll the terminal.

- The scrolling experience feels off, because the content's moving speed is accelerated and is constantly displaced relatively to the finger. Perfectly, there should be a smooth scrolling with inertia. But maybe this is out of scope of this PR.

- The only thing I'm not sure about is long press. If Alacritty is to have full-fledged touch support, then it should be reserved for context menu in the future (with paste option) and the delay should be increased to 500ms (Android defaults). Making it paste from the buffer is a bad idea: would be really annoying to accidentally shove unrelated stuff into the command line. Otherwise, long tap can be a substitute for double tap and select whole words. It can also be left out of this PR and do nothing, until it is decided what to do next.

I've seen lots of other things mentioned in this thread, like line selection or selecting things with 2 fingers, etc. But they do not conflict with the current implementation and are out of scope of this pull request. Before getting to these things, some basics should be implemented, and then you can gradually build upon it.

*Gently poking @nixpulvis because they asked for end-user perspective.*

PR #4348, opened 2020-10-25
Switch to dirs-next

lespea:
Upgrade to `dirs-next` since `dirs` is no longer supported.  [This comment](https://github.com/sharkdp/fd/pull/664#issuecomment-708662203) comment has more details.  I assume this doesn't need a changelog entry?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,7 +29,7 @@ dependencies = [
  "clap",
  "copypasta",
  "crossfont",
- "dirs",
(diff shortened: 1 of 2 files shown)


chrisduerr (changes requested):
Currently I don't see how this improves anything other than adding another dependency.

If `terminfo` is updated to use `dirs-next`, then it would make some sense to stay consistent. But it's not like we need `dirs` to be actively maintained.

PR #4349, opened 2020-10-25
Update to time 0.2

lespea:
Upgrade to a supported version time, `0.2`.  Needed a slight change in the formatting syntax but the output should be the same.  I believe using `lazy_format` saves allocating a string, not that it really matters for this code but still a nice benefit.  I didn't update the changelog as there should be no change in functionality.

More details: https://github.com/time-rs/time/issues/190

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,7 +43,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
- "time",
+ "time 0.2.22",
  "unicode-width",
  "urlocator",
  "wayland-client",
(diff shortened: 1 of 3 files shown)


chrisduerr (changes requested):
Please do not update the entire lockfile.

lespea:
Okay I ran `cargo update -p time` instead, hopefully that works?

chrisduerr (changes requested):
This adds 21 new dependencies without any benefit. I don't see a reason for us to pull in so many dependencies for nothing.

PR #4354, opened 2020-10-27
Error on dead code when running clippy

kchibisov:
It should simplify tracking of dead code on CI builds
and when cross checking.

### DIFF
--- a/alacritty/src/main.rs
+++ b/alacritty/src/main.rs
@@ -1,6 +1,8 @@
 //! Alacritty - The GPU Enhanced Terminal.
 
+#![warn(rust_2018_idioms, future_incompatible)]
 #![deny(clippy::all, clippy::if_not_else, clippy::enum_glob_use, clippy::wrong_pub_self_convention)]
+#![cfg_attr(feature = "cargo-clippy", deny(warnings))]
 #![cfg_attr(all(test, feature = "bench"), feature(test))]
 // With the default subsystem, 'console', windows creates an additional console
 // window for the program.
--- a/alacritty_terminal/src/lib.rs
+++ b/alacritty_terminal/src/lib.rs
@@ -1,6 +1,8 @@
 //! Alacritty - The GPU Enhanced Terminal.
 
+#![warn(rust_2018_idioms, future_incompatible)]
 #![deny(clippy::all, clippy::if_not_else, clippy::enum_glob_use, clippy::wrong_pub_self_convention)]
+#![cfg_attr(feature = "cargo-clippy", deny(warnings))]
 #![cfg_attr(all(test, feature = "bench"), feature(test))]
 
 pub mod ansi;
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -898,8 +898,8 @@ impl<N: Notify + OnResize> Processor<N> {
     ///
     /// Doesn't take self mutably due to borrow checking.
     fn handle_event<T>(
-        event: GlutinEvent<Event>,
-        processor: &mut input::Processor<T, ActionContext<N, T>>,
+        event: GlutinEvent<'_, Event>,
+        processor: &mut input::Processor<'_, T, ActionContext<'_, N, T>>,
     ) where
         T: EventListener,
     {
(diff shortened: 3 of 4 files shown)


alacritty/src/main.rs
```diff
@@ -1,6 +1,7 @@
 //! Alacritty - The GPU Enhanced Terminal.
 
 #![deny(clippy::all, clippy::if_not_else, clippy::enum_glob_use, clippy::wrong_pub_self_convention)]
+#![cfg_attr(feature =  "cargo-clippy", deny(dead_code))]
```
chrisduerr, line 4:
```suggestion
#![cfg_attr(feature =  "cargo-clippy", deny(warnings))]
```

alacritty_terminal/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 //! Alacritty - The GPU Enhanced Terminal.
 
 #![deny(clippy::all, clippy::if_not_else, clippy::enum_glob_use, clippy::wrong_pub_self_convention)]
+#![cfg_attr(feature =  "cargo-clippy", deny(dead_code))]
```
chrisduerr, line 4:
```suggestion
#![cfg_attr(feature =  "cargo-clippy", deny(warnings))]
```

chrisduerr (changes requested):
Let's be even more hardcore here. I don't ever want to see a rustc warning on master.

chrisduerr:
Also could we maybe go ahead and all even more lint errors that are currently not enforced? I think both `rust_2018_idioms` and `future-incompatible` would be good options. I'm not sure if the latter is already included in `warning` but the former certainly isn't.

alacritty/src/main.rs
```diff
@@ -1,6 +1,8 @@
 //! Alacritty - The GPU Enhanced Terminal.
 
+#![warn(rust_2018_idioms, future_incompatible)]
```
chrisduerr, line 3: Why warn and not deny?

chrisduerr (changes requested):

alacritty/src/main.rs
kchibisov: Well, they are warnings by their nature, so I decided to nor abort on them for normal compilations, since if future rust will add more stuff into those checks and they trigger on more code, it'll be impossible to build alacritty, thus breaking packaging, since new rust won't be able to build old alacritty, since rust fine tuned warnings. However since we error on warnings when running clippy, we will error on them as well, and in the end there's no way we can miss anything from those warnings on CI.

chrisduerr: > However since we error on warnings when running clippy, we will error on them as well

Those are rust warnings though, not clippy warnings? Have you confirmed this is actually the case? After all it wasn't erroring on the other warnings you've set to deny in this PR either. Since `#[deny(warnings)]` does not disable all warnings but instead just disables everything that is in the "warnings" group by default.

But I'm perfectly fine with putting that behind a cfg and keeping it as only a warning for normal compilation. That way people will see it and at the same time it will fail CI.

kchibisov: >Those are rust warnings though, not clippy warnings? Have you confirmed this is actually the case? After all it wasn't erroring on the other warnings you've set to deny in this PR either. Since #[deny(warnings)] does not disable all warnings but instead just disables everything that is in the "warnings" group by default.

Yes, but we start error on rustc warnings when running clippy on that PR, so everything will be aborted as you want with `deny`.

kchibisov: Hm, and I think keeping those as warnings for normal builds instead of just adding them to `deny` when running clippy is better, since it's easier to notice them.

chrisduerr: > Yes, but we start error on rustc warnings when running clippy on that PR, so everything will be aborted as you want with deny.

We don't error on rustc warnings, we error on rustc lints that are usually warnings. Your current version will just spit out the warnings in clippy mode and keep going.

chrisduerr: Should be good, I've just confirmed myself that `deny(warnings)` actually includes things that are manually changed into warnings too.

chrisduerr (approved):

alacritty/src/main.rs
kchibisov: >Should be good, I've just confirmed myself that deny(warnings) actually includes things that are manually changed into warnings too.

Yes, and those a lints, as well as dead code ones, however they are not enabled by default, so that could be confusing. The dead code and rust idioms one are the same things when rust blows up on them with error afaics.

PR #4355, opened 2020-10-27
Fix build failure on macOS without features

### LINKED ISSUE #4353: Compilation failure on macOS when --no-default-features is passed
Compiling on MacOS, I would not expect to need the default features of 'x11', 'wayland', and 'winpty'. But the if defined statements in `alacritty/src/window.rs` do not take that case into account, so I get the following build failure:

```
error[E0201]: duplicate definitions with name `set_urgent`:
   --> alacritty/src/window.rs:343:5
    |
334 | /     pub fn set_urgent(&self, is_urgent: bool) {
335 | |         if !is_urgent {
336 | |             return;
337 | |         }
338 | |
339 | |         self.window().request_user_attention(RequestUserAttentionType::Critical);
340 | |     }
    | |_____- previous definition of `set_urgent` here
...
343 |       pub fn set_urgent(&self, _is_urgent: bool) {}
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ duplicate definition

error: aborting due to previous error
```

However, if I leave them defined I end up with dead_code warnings on compilation:

```
warning: associated function is never used: `wayland_display`
   --> alacritty/src/window.rs:401:12
    |
401 |     pub fn wayland_display(&self) -> Option<*mut std::ffi::c_void> {
    |            ^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default
```

which is also not ideal.

### System

OS: macOS

chrisduerr:
Fixes #4353.

### DIFF
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -339,7 +339,7 @@ impl Window {
         self.window().request_user_attention(RequestUserAttentionType::Critical);
     }
 
-    #[cfg(any(not(feature = "x11"), windows))]
+    #[cfg(any(windows, not(any(feature = "x11", target_os = "macos"))))]
     pub fn set_urgent(&self, _is_urgent: bool) {}
 
     pub fn set_outer_position(&self, pos: PhysicalPosition<i32>) {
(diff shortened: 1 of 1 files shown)


nixpulvis (changes requested):
Working on macOS with and without default features.

However on X11 the I get the following error:

```
cargo -Zpackage-features run --no-default-features --features='alacritty/x11'

error[E0599]: no method named `wayland_display` found for struct `window::Window` in the current scope
```

chrisduerr:
Good point, should be good now @nixpulvis.

nixpulvis:
I was momentarily very happy to see the dummy `wayland_display` function deleted, but it was sadly there for a reason.

While you're at it, you might as well remove this warning too:

```
warning: unused import: `EventLoopWindowTargetExtUnix`
```

chrisduerr:
> While you're at it, you might as well remove this warning too

Good point, there should be no more warnings with any feature combination on Linux now.

nixpulvis (approved):
Working as intended. Thanks

PR #4366, opened 2020-10-29
Set ALACRITTY_PID environment variable

zqad:
Useful in scripts and daemons when trying to figure out the parent terminal.

I used this to solve the issue highlighted in this comment:
https://github.com/alacritty/alacritty/issues/1072#issuecomment-674310366

With the environment variable available, I just drop in a short shell script containing the following, and point to it with bell -> command -> program:
```
#!/bin/sh
swaymsg "[pid=$ALACRITTY_PID] urgent enable"
```
Or just use sh directly:
```
bell:
  command:
    program: sh
    args: ["-c", 'swaymsg "[pid=$ALACRITTY_PID] urgent enable"']
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Use yellow/red from the config for error and warning messages instead of fixed colors
 - Existing CLI parameters are now passed to instances spawned using `SpawnNewInstance`
 - Wayland's Client side decorations now use the search bar colors
+- Set `$ALACRITTY_PID` to the terminal PID on start
 
 ### Fixed
 
--- a/alacritty/src/main.rs
+++ b/alacritty/src/main.rs
@@ -135,6 +135,7 @@ fn run(
 
     // Set environment variables.
     tty::setup_env(&config);
+    std::env::set_var("ALACRITTY_PID", &format!("{}", std::process::id()));
 
     let event_proxy = EventProxy::new(window_event_loop.create_proxy());
 


kchibisov:
Well, we kind of have pid in `ALACRITTY_LOG` env variable already, just requires grepping for it. So I'm kind of against adding one more variable just for pid, since it doens't provide any value, other than hackish scripts.

zqad:
I see that point, but I would recommend against pointing to the `ALACRITTY_LOG` variable, as that would mean that the log path becomes an implicit interface if people starts to parse it. Plus I assume the path differs between platforms. And someone might decide to move it to say /run/user/<pid>/alacritty at which time there might be some breakage.. Saying this makes me want to point out that having an explicit variable should probably make for less hackish scripts overall ;-)

Other uses for the PID might be to raise windows programmatically from scripts, or interacting with the compositor in other ways, so I would say that it opens up quite a few doors to have it readily available. Just looking at the comment above and talking to others, it seems that urgency is the top missing bit in my proximity, but there are probably more things that can be done using this info.

kchibisov:
>I see that point, but I would recommend against pointing to the ALACRITTY_LOG variable, as that would mean that the log path becomes an implicit interface if people starts to parse it. Plus I assume the path differs between platforms. And someone might decide to move it to say /run/user/(pid)/alacritty at which time there might be some breakage.. Saying this makes me want to point out that having an explicit variable should probably make for less hackish scripts overall ;-)

It just always suffixed with pid, I don't see a point of different platforms, you're saying that way, only because you need pid for Wayland, just because gnome can't make their shit together and add proper urgency, since they want dbus for it. Your issues from compositors interactions from scripts are biased torwards Wayland, also, the compositor knows the pid of the currently focused application or some other other things in general. Just having random variable that sets pid for literally no reason, since it's useful only in scripts you're running with bindings, not in your shell, since you can get `$ppid` there just fine.

>it seems that urgency is the top missing bit in my proximity, but there are probably more things that can be done using this info.

Yeah, because of GNOME, not because we don't **directly** expose pid, it just happened that sway has criteria for pid (they were added also because of GNOME).

kchibisov:
if @chrisduerr don't mind setting pid for whatever reason, it'll save me like one `grep`...

chrisduerr:
I'd be curious how many other terminal emulators export their PID in an environment variable. Generally I'm against giving user an easy interface for hacky scripts.

nixpulvis:
As a currently happy X11 user, I'm struggling to think up a use case for this variable at all...

kchibisov:
>As a currently happy X11 user, I'm struggling to think up a use case for this variable at all...

Arbitrary write to pty? Like the whole purpose for the PID thing in general, and the reason why it's 'hacky', the Wayland thing is just coincidence that went too far.

chrisduerr:
I just tested `termite`, `xterm` and `urxvt` and none of them set the PID either. So I really don't see a reason why we would.

As @kchibisov and @nixpulvis also don't seem massively in favor of this, I'm going to close this PR.

nixpulvis:
> Arbitrary write to pty?

I'm personally still all for this functionality, but would much rather see it as a feature of Alacritty than a hacky thing that writes to the PTY from another process.

chrisduerr:
> than a hacky thing that writes to the PTY from another process.

Alacritty itself would still write to the PTY from another process.

zqad:
Fair enough, but I do think that having the possibility of running commands with BEL and dismissing this as the opening to hacky scripts is not completely fair :-) My approach with this change was just to make workarounds more convenient, as I assume the command addition to bell was also added to enable workarounds. And I do agree about the root cause, Wayland should really have this interface. And I do know the sad history about Gnome's blocking stance. In the end, I would much rather not need to do any hacks in this area, but here we are.

Anyway, this is my current hack (derived from one I got from @plundra), in case someone else ends up here from the google machine:
```bell:
  command:
    program: sh
    args: ["-c", 'apid=${ALACRITTY_LOG#*/Alacritty-};swaymsg "[pid=${apid%.log}] urgent enable"']
```

Also, thanks for the work being poured into this. I can't remember the last time I was this happy with a piece of software of the bat so you are all obviously doing some great work here.

Cheers,
/Jonas

chrisduerr:
> as I assume the command addition to bell was also added to enable workarounds

No it was not. Ringing the audio bell with a command is not what I would consider a "hack". That was the primary goal of it.

PR #4380, opened 2020-11-05
Add editorconfig

### LINKED ISSUE #4378: Consider setting up editorconfig
While rustfmt prevents us from spending our time on catching things like wrong indention or mixed indention, it's unclear what we're expecting for our shaders, etc, since rustfmt doesn't work on them, thus I think it'll be nice to have `.editorconfig` which could cover them, as well as rust itself. A lot of editors, like vim (required plugging though) can automatically pick them up and apply respectful settings and its syntax is pretty clear.

kchibisov:
This commit also purges .gitignore, since those files should be handled
by global ignore file.

Fixes #4378.


.editorconfig
```diff
@@ -0,0 +1,20 @@
...
+trim_trailing-whitespace = true
+
+[*.yml]
+indent_style = space
+trim_trailing-whitespace = true
+indent_size = 2
+
+[Makefile]
+indent_style = tab
+tab_width = 8
```
kchibisov, line 19: If it matters github supports editorconfig for diffs and file vies, so if you want different tab_width and indent_size let me know for Makefile let me know.

Makefile
```diff
@@ -5,15 +5,15 @@ RELEASE_DIR = target/release
 MANPAGE = $(ASSETS_DIR)/alacritty.man
 TERMINFO = $(ASSETS_DIR)/alacritty.info
 COMPLETIONS_DIR = $(ASSETS_DIR)/completions
-COMPLETIONS = $(COMPLETIONS_DIR)/_alacritty \
-			  $(COMPLETIONS_DIR)/alacritty.bash \
-			  $(COMPLETIONS_DIR)/alacritty.fish
+COMPLETIONS =   $(COMPLETIONS_DIR)/_alacritty \
```
chrisduerr, line 8: Are you trying to mess with me now? :smiley:

Makefile
```diff
@@ -5,15 +5,15 @@ RELEASE_DIR = target/release
...
+		$(COMPLETIONS_DIR)/alacritty.bash \
+		$(COMPLETIONS_DIR)/alacritty.fish
```
chrisduerr, line 10: Generally I'd be fine with aligning these tbh. EIther way is fine to me so this doesn't really need to be changed if you ask me.

Makefile
```diff
@@ -51,10 +51,10 @@ $(DMG_NAME): $(APP_NAME)
...
 	@ln -sf /Applications $(DMG_DIR)/Applications
 	@hdiutil create $(DMG_DIR)/$(DMG_NAME) \
-		-volname "Alacritty" \
-		-fs HFS+ \
-		-srcfolder $(APP_DIR) \
-		-ov -format UDZO
+			-volname "Alacritty" \
+			-fs HFS+ \
+			-srcfolder $(APP_DIR) \
+			-ov -format UDZO
```
chrisduerr, line 57: Why go two levels deep here?

.editorconfig
```diff
@@ -0,0 +1,19 @@
...
+indent_style = space
+indent_size = 4
+
+[*.yml]
+indent_style = space
+indent_size = 2
+
+[Makefile]
+indent_style = tab
+indent_size = 8
```
chrisduerr, line 19: What's your argument for using such a high indent size for makefiles? I'd stick to 4.

.gitignore
```diff
@@ -1,36 +1 @@
-# Rust build directory
 **/target
```
chrisduerr, line 1: Since we're in a workspace now, just the root should be sufficient, right?

chrisduerr (changes requested):

.gitignore
kchibisov: you mean, just `target`?

Makefile
```diff
@@ -6,14 +6,14 @@ MANPAGE = $(ASSETS_DIR)/alacritty.man
 TERMINFO = $(ASSETS_DIR)/alacritty.info
 COMPLETIONS_DIR = $(ASSETS_DIR)/completions
 COMPLETIONS = $(COMPLETIONS_DIR)/_alacritty \
-			  $(COMPLETIONS_DIR)/alacritty.bash \
-			  $(COMPLETIONS_DIR)/alacritty.fish
+			$(COMPLETIONS_DIR)/alacritty.bash \
+			$(COMPLETIONS_DIR)/alacritty.fish
```
chrisduerr, line 10: Tabs always make this unnecessarily difficult to see, but are those three tabs of indentation? Why?

You'd think there should be one tab of indentation at most beyond the previous element, past that we're talking about alignment which should be done with spaces?

Makefile
chrisduerr: Tabs are not for alignment.

chrisduerr (changes requested):

.editorconfig
```diff
@@ -0,0 +1,19 @@
...
+indent_style = space
+indent_size = 4
+
+[*.yml]
+indent_style = space
+indent_size = 2
+
+[Makefile]
+indent_style = tab
+indent_size = 4
```
chrisduerr, line 19: Actually, I see no reason why we should add the `indent_size` at all if it's about tabs. With spaces it makes sense but let people decide on their own for something that uses tabs anyways?

kchibisov: it'll render 4 spaces for tab on github UI instead of 8 though.

chrisduerr: I mean that's something anyone who actually cares can override too. I don't particularly care about what github does, I don't write my code for github.

kchibisov: yeah, that's fair.

chrisduerr (approved):

nixpulvis:
Personally, I find the removal of Rust specific git ignores a little unfortunate. I generally don't configure my global `.gitignore` for any language/framework in particular.

kchibisov:
>Personally, I find the removal of Rust specific git ignores a little unfortunate. I generally don't configure my global .gitignore for any language/framework in particular.

We have the exact same gitignore rust produces by default.

chrisduerr:
Yeah, all the Rust stuff (`**/Cargo.lock`, `**/target`) were back from when cargo workspaces didn't exist, or weren't used at least. Back then everything just produced their own lockfiles and target directories.

If you build anything in a workspace, the root target and lockfile will be used. So it's simply not necessary anymore. All crates we provide in our repo are part of this workspace.

PR #4382, opened 2020-11-06
Deprecate the WinPTY backend

chrisduerr:


alacritty/src/config/mod.rs
```diff
@@ -312,6 +312,14 @@ fn print_deprecation_warnings(config: &Config) {
             "Config dynamic_title is deprecated; please use window.dynamic_title instead",
         )
     }
+
+    #[cfg(windows)]
+    if config.winpty_backend {
```
kchibisov, line 317: This warning is kind of useless when it's built without winpty backend? I think we should mention in that case, that winpty backend is disabled and conpty is being used, and if they want winpty, they should compile stuff manually.

chrisduerr: I mean the goal is for them to delete the config option, does it matter if it is enabled or not?

If it's built without the WinPTY backend, there should still be a warning that tells the user that they should delete the option?

kchibisov: I'm not sure if I'd always warn tbqh, since on some systems it's desired to use it, that being sad, I'd only warn if winpty feature is enabled, and this option is there.

chrisduerr: I mean even if it's desired to use it, people should migrate away from it. If it's gonna be removed in two months, they're going to have to accept it anyways.

Warning even though the WinPTY feature isn't enabled allows us to tell users to actually clean up their config file, then we can just remove the variable when the feature is removed. Otherwise we'd have to warn then, which seems suboptimal?

kchibisov: I guess...

kchibisov (approved):

alacritty/src/config/mod.rs
chrisduerr: I mean if anything, shouldn't it be the other way around? Warn only when it was built **without** the WinPTY backend? If people intentionally opt into the feature then they probably know what they're doing and are aware that the feature is deprecated.

That could make some sense I think.

kchibisov: Ah, yeah, the other way around, but it really doesn't matter, knowing that we'll drop it in like 2 month.

alacritty/src/config/mod.rs
```diff
@@ -313,11 +313,12 @@ fn print_deprecation_warnings(config: &Config) {
         )
     }
 
-    #[cfg(windows)]
+    #[cfg(all(windows, not(feature = "winpty")))]
     if config.winpty_backend {
         warn!(
             target: LOG_TARGET_CONFIG,
-            "Config winpty_backend is deprecated; please use the ConPTY backend instead",
+            "Config winpty_backend is deprecated and requires a compilation flag; please use the \
```
kchibisov, line 320: In that form it's not like you should use ConPTY backend, it's like you have no choice and the only thing you can do is to remove it, since you're already using ConPTy backend. So, I'd removed the last part of that sentence, or reworded it.

kchibisov (changes requested):

alacritty/src/config/mod.rs
chrisduerr: How would you reword it? We want the user to use the ConPTY backend, instead of recompiling with WinPTY. So that's why I've worded it this way.

kchibisov: Maybe something like that?

 "Config winpty_backend is deprecated and requires a compilation flag; it can be removed from the config"

chrisduerr: But at that point, doesn't it sound like the compilation flag should be enough? Which it is not, you'd still need the configuration option.

kchibisov: I just don't like `Please use ConPTY`, because you can't really not use it.

chrisduerr: I think it should be fine now. If someone can compile it from source, they should be able to figure out that the option is still required.

kchibisov (approved):

PR #4387, opened 2020-11-06
Migrate from Travis CI to GitHub Actions

chrisduerr:
This removes all CI builds from travis-ci, due to their recent changes
in policy and harsh limitations on builds. With build times over 2
hours, it was a significant hindrance to development.

Instead of Travis CI, the CI is now split on Sourcehut and GitHub. Since
Sourcehut only supports Linux/BSD, all builds on those operating systems
are executed there. The GitHub Actions CI is used to build for
Windows/macOS, which are not available on Sourcehut.

Since asset deployment for releases requires builds on all platforms,
this is also done on GitHub actions. Though the new `upload_asset.sh`
script makes sure that migration in the future is fairly simple and we
do not tie ourselves to the overly complicated GitHub Actions ecosystem.


chrisduerr:
The available test results can be found here:
 - Assets: https://github.com/chrisduerr/alacritty/releases/tag/v9.6.4
 - Release: https://github.com/chrisduerr/alacritty/actions/runs/350022919
 - Build: https://github.com/chrisduerr/alacritty/actions/runs/350023112

.builds/linux.yml
```diff
@@ -0,0 +1,36 @@
...
+  - make
+  - freetype2
+  - fontconfig
+  - libxcb
+
+sources:
+  - https://github.com/alacritty/alacritty
+
+environment:
+  PATH: /home/build/.cargo/bin:/usr/bin/
```
kchibisov, line 15: If you want to update path, you may try updating freebsd job as well?

.builds/linux.yml
```diff
@@ -0,0 +1,36 @@
...
+
+tasks:
+  - rustup: |
+      curl https://sh.rustup.rs -sSf | sh -s -- -y --profile minimal --default-toolchain stable
+  - test: |
+      cd alacritty
+      cargo build
```
kchibisov, line 22: what's the point of 'cargo build', just 'cargo test' should be sufficient?

.builds/linux.yml
```diff
@@ -0,0 +1,36 @@
...
+      rustup toolchain install nightly -c rustfmt
+      cargo +nightly fmt -- --check
+  - clippy: |
+      cd alacritty
+      rustup component add clippy
+      cargo clippy --all-targets
+  - oldstable: |
+      cd alacritty
+      rustup default 1.43.1
+      cargo build
```
kchibisov, line 35: same as above.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Rustup
+        run: rustup default ${{ matrix.rust_version }}
+      - name: Build
+        run: cargo build
+      - name: Test
+        run: cargo test
```
kchibisov, line 24: Why not just `test`?

scripts/upload_asset.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
```
kchibisov, line 1: Maybe it should be in `ci` folder like it was before? `scripts` sounds like something users can use or distros think that they should package, however it's only useful and will be used by our CI job.

scripts/upload_asset.sh
```diff
@@ -0,0 +1,93 @@
...
+
+# Assure parameters are correct.
+if [ $# -lt 2 ]; then
+    echo "Usage: upload_asset.sh <FILE> <TOKEN>"
+    exit 1
+fi
+
+repo="chrisduerr/alacritty"
```
kchibisov: Wrong repo.

scripts/upload_asset.sh
```diff
@@ -0,0 +1,93 @@
...
+file_path=$1
+bearer=$2
+
+echo "Starting asset upload from $file_path to $repo."
```
kchibisov, line 13:
```suggestion
printf "Starting asset upload from $file_path to $repo.\n"
```

scripts/upload_asset.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
+
+# Assure parameters are correct.
+if [ $# -lt 2 ]; then
+    echo "Usage: upload_asset.sh <FILE> <TOKEN>"
```
kchibisov, line 5:
```suggestion
    printf "Usage: upload_asset.sh <FILE> <TOKEN>\n"
```

scripts/upload_asset.sh
```diff
@@ -0,0 +1,93 @@
...
+# Get the release for this tag.
+tag="$(git describe --tags --abbrev=0)"
+
+# Make sure the git tag could be determined.
+if [ -z "$tag" ]; then
+    printf "\e[31mError: Unable to find git tag\e[0m\n"
+    exit 1
+fi
+
+echo "Git tag: $tag"
```
kchibisov, line 24:
```suggestion
printf "Git tag: $tag\n"
```

scripts/upload_asset.sh
```diff
@@ -0,0 +1,93 @@
...
+
+# Get the upload URL for the current tag.
+#
+# Since this might be a draft release, we can't just use the /releases/tags/:tag
+# endpoint which only shows published releases.
+echo "Checking for existing release..."
```
kchibisov, line 30:
```suggestion
printf "Checking for existing release...\n"
```

scripts/upload_asset.sh
```diff
@@ -0,0 +1,93 @@
...
+    | paste - - \
+    | grep -e "tag_name\": \"$tag" \
+    | head -n 1 \
+    | sed 's/.*\(https.*assets\).*/\1/' \
+)
+
+# Create a new release if we didn't find one for this tag.
+if [ -z "$upload_url" ]; then
+    echo "No release found."
+    echo "Creating new release..."
```
kchibisov, line 46: printf for them as well. Maybe even merge into one statement.

scripts/upload_asset.sh
```diff
@@ -0,0 +1,93 @@
...
+
+    # Abort if the release could not be created.
+    if [ $? -ne 0 ]; then
+        printf "\e[31mError: Unable to create new release.\e[0m\n"
+        exit 1;
+    fi
+
+    # Extract upload URL from new release.
+    upload_url=$(\
+        echo "$response" \
```
kchibisov, line 66: printf

scripts/upload_asset.sh
```diff
@@ -0,0 +1,93 @@
...
+
+# Propagate error if no URL for asset upload could be found.
+if [ -z "$upload_url" ]; then
+    printf "\e[31mError: Unable to find release upload url.\e[0m\n"
+    exit 2
+fi
+
+# Upload the file to the tag's release.
+file_name=${file_path##*/}
+echo "Uploading asset $file_name to $upload_url..."
```
kchibisov, line 80: printf

.github/workflows/release.yml
```diff
@@ -0,0 +1,80 @@
+name: Release
+
+on:
+  push:
+    branches: [master]
```
kchibisov, line 5: releases are usually on separate branches, how it'll work with them?

.github/workflows/release.yml
```diff
@@ -0,0 +1,80 @@
...
+          sudo apt-get install cmake pkg-config libfreetype6-dev libfontconfig1-dev \
+            libxcb-xfixes0-dev python3
+      - name: Test
+        run: cargo test --release
+      - name: Gzip manpage
+        run: gzip -c "./extra/alacritty.man" > "./alacritty.1.gz"
+      - name: Upload Assets
+        run: |
+          cp ./extra/logo/alacritty-term.svg ./Alacritty.svg
+          ./scripts/upload_asset.sh ./Alacritty.svg $GITHUB_TOKEN
```
kchibisov, line 73: Why `cp`?

kchibisov (changes requested):

.builds/linux.yml
chrisduerr: I think it's useful to do both, since one will use the previous artifacts anyways and you know if just the tests fail, or if the actual build fails already.

scripts/upload_asset.sh
chrisduerr: Having a ci folder just for this one script seems a bit much. Why do you think that `scripts` is for user scripts?

scripts/upload_asset.sh
chrisduerr: I have no interest in using only echo in this script.

.github/workflows/release.yml
chrisduerr: Tbh I did not consider that. The branch restriction should probably be removed.

scripts/upload_asset.sh
kchibisov: I don't really like mixing 2 things.

scripts/upload_asset.sh
kchibisov: Well, how distribution will understand that this shouldn't be packaged? It seems like some generic script, well, they can look into it, but having `ci` will clarify things for folks that are looking at the repo first time.

.github/workflows/release.yml
```diff
@@ -69,7 +68,7 @@ jobs:
         run: gzip -c "./extra/alacritty.man" > "./alacritty.1.gz"
       - name: Upload Assets
         run: |
-          cp ./extra/logo/alacritty-term.svg ./Alacritty.svg
+          mv ./extra/logo/alacritty-term.svg ./Alacritty.svg
```
kchibisov, line 71: I mean, why should we even move it before upload? Can't we use full path like you do for everything else?

.builds/freebsd.yml
```diff
@@ -12,19 +12,22 @@ packages:
 sources:
   - https://github.com/alacritty/alacritty
 
+environment:
+  PATH: /home/build/.cargo/bin:/usr/bin/
```
kchibisov, line 16: Would you mind adding `/usr/local/bin` before `/usr/bin`, since it's where things goes from packges on  freebsd iirc.

kchibisov (changes requested):

.builds/linux.yml
kchibisov: That not entirely true, it'll relink iirc, so it'll add more build time than it should. Though, maybe having separate status justifies building more.

.github/workflows/release.yml
chrisduerr: To rename it?

kchibisov: Oh, yeah, forgot that it's different name..

scripts/upload_asset.sh
chrisduerr: Why would a distribution start packaging random scripts?

scripts/upload_asset.sh
chrisduerr: I do though. There's nothing wrong with mixing these two for the things they're better at. Using printf just to not use echo when echo is just simpler makes little sense.

scripts/upload_asset.sh
kchibisov: I just know that certain distros are packaging our scripts already, so they'll just package this one as well...

chrisduerr: They are? Do you know which?

kchibisov: https://gitweb.gentoo.org/repo/gentoo.git/tree/x11-terms/alacritty/alacritty-0.5.0.ebuild#n360

kchibisov (changes requested):

.builds/freebsd.yml
```diff
@@ -7,21 +8,30 @@ packages:
   - x11-fonts/fontconfig
   - x11-fonts/dejavu
   - x11/libxcb
+
 sources:
   - https://github.com/alacritty/alacritty
+
+environment:
+  PATH: /home/build/.cargo/bin:/bin:/usr/bin:/usr/local/bin
```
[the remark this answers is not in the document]
chrisduerr: As long as it works, I don't think it matters?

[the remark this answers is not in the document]
chrisduerr: Different versions of applications aren't generally shipped multiple times in these locations on FreeBSD, right?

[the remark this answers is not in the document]
kchibisov: >Different versions of applications aren't generally shipped multiple times in these locations on FreeBSD, right?

No, `/usr/bin` have some as well as `/bin` has some common common utils, and they also link to system libraries  from `/usr/lib`, everything from `/usr/local/bin` are linked with libraries under `/usr/local/lib`. So I'd generally make local path before everything else, just because everything is goes there, and you want to always use a tool you're installing, not the one system provides (They even has duplicated libs, since you have ncurses in local and ncurses in /usr/lib, and they're completely different).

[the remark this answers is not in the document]
chrisduerr: > and you want to always use a tool you're installing, not the one system provides

We're not talking about a user system though, but CI. What practical difference does this make?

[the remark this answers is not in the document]
chrisduerr: Also fyi, this is the default `PATH`: `/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin:/usr/local/bin:/home/build/bin`.

So this matches what I have.

kchibisov (approved):

PR #4388, opened 2020-11-07
Fix config file formatting

chrisduerr:
The configuration file, especially the section documenting the different
binding actions, had some extremely long lines. All the text is now
reformatted to be at most 80 columns wide.

The only lines remaining which are beyond 80 columns wide are the
configuration bindings, which would significantly suffer from linebreaks
and are not plain text.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -57,7 +57,7 @@
   #
   # Values for `decorations` (macOS only):
   #     - transparent: Title bar, transparent background and title bar buttons
-  #     - buttonless: Title bar, transparent background, but no title bar buttons
+  #     - buttonless: Title bar, transparent background and no title bar buttons
   #decorations: full
 
   # Startup Mode (changes require restart)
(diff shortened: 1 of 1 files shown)


alacritty.yml
```diff
@@ -524,67 +530,73 @@
...
-#       Open URLs at the cursor location with the launcher configured in `url.launcher`.
+#       Open URLs at the cursor location with the launcher configured in
+#       `url.launcher`.
+#   - ToggleNormalSelection
+#   - ToggleLineSelection
+#   - ToggleBlockSelection
+#   - ToggleSemanticSelection
+#       Toggle semantic selection based on `selection.semantic_escape_chars`.
+#
+# - Vi mode exclusive cursor motion actions:
```
chrisduerr, line 544: I've made the choice of separating different actions a bit more, to make it possible to shorten the descriptions of the actions dramatically. That should hopefully make things a little quicker to read.

I personally find that the different sections of the bindings are a bit difficult to see though, so maybe there's a better way to do that.

Maybe we should use markdown style `# Headings` instead of the `- Sections:` we have right now?

kchibisov: I think it could confuse a bit, since `#` is a start of comment in yaml? Though, I'm not completely against `#` for certain things... Up to you here.

chrisduerr: Yeah I'm not sure if I want to change anything about that at this point. Just an idea for the future.

kchibisov (approved):

PR #4390, opened 2020-11-08
Add ability to select text during search

chrisduerr:
This removes the restriction of not being able to select text while the
search is active, making it a bit less jarring of a UX when the user
tries to interact with the terminal during search.

Since the selection was used during vi-less search to highlight the
focused match, the normal match color is now inverted instead to allow
the user to still recognize the it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Unicode 13 support
 - Option to run command on bell which can be set in `bell.command`
 - Fallback to program specified in `$SHELL` variable on Linux/BSD if it is present
+- Ability to make selections while search is active
 
 ### Changed
 
@@ -120,6 +121,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - URLs are no longer highlighted without a clearly delimited scheme
 - Renamed config option `visual_bell` to `bell`
 - Moved config option `dynamic_title` to `window.dynamic_title`
+- When searching without vi mode, matches are only selected once search is cancelled
 
 ### Fixed
 
@@ -141,6 +143,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Ingoring of default FreeType properties
 - Alacritty crashing at startup when the configured font does not exist
 - Font size rounding error
+- Opening URLs while search is active
 
 ### Removed
 
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -480,7 +480,7 @@ impl Display {
                 for mut cell in grid_cells {
                     // Invert the active match in vi-less search.
                     let cell_point = Point::new(cell.line, cell.column);
-                    if !cell.selected
+                    if cell.is_match
                         && viewport_match
                             .as_ref()
                             .map_or(false, |viewport_match| viewport_match.contains(&cell_point))
--- a/alacritty/src/display.rs
+++ b/alacritty/src/display.rs
@@ -27,7 +27,7 @@ use crossfont::{self, Rasterize, Rasterizer};
 use alacritty_terminal::event::{EventListener, OnResize};
 use alacritty_terminal::index::{Column, Direction, Point};
 use alacritty_terminal::selection::Selection;
-use alacritty_terminal::term::{RenderableCell, SizeInfo, Term, TermMode};
+use alacritty_terminal::term::{SizeInfo, Term, TermMode};
 use alacritty_terminal::term::{MIN_COLS, MIN_SCREEN_LINES};
 
 use crate::config::font::Font;
@@ -437,7 +437,13 @@ impl Display {
         mods: ModifiersState,
         search_state: &SearchState,
     ) {
-        let grid_cells: Vec<RenderableCell> = terminal.renderable_cells(config).collect();
+        // Convert search match from viewport to absolute indexing.
+        let search_active = search_state.regex().is_some();
+        let viewport_match = search_state
+            .focused_match()
+            .and_then(|focused_match| terminal.grid().clamp_buffer_range_to_visible(focused_match));
+
+        let grid_cells = terminal.renderable_cells(config, !search_active).collect::<Vec<_>>();
         let visual_bell_intensity = terminal.visual_bell.intensity();
         let background_color = terminal.background_color();
         let cursor_point = terminal.grid().cursor.point;
@@ -471,7 +477,21 @@ impl Display {
 
             self.renderer.with_api(&config.ui_config, config.cursor, &size_info, |mut api| {
                 // Iterate over all non-empty cells in the grid.
-                for cell in grid_cells {
+                for mut cell in grid_cells {
+                    // Invert the active match in vi-less search.
+                    let cell_point = Point::new(cell.line, cell.column);
+                    if !cell.selected
+                        && viewport_match
+                            .as_ref()
+                            .map_or(false, |viewport_match| viewport_match.contains(&cell_point))
+                    {
+                        let colors = config.colors.search.focused_match;
+                        let match_fg = colors.foreground().color(cell.fg, cell.bg);
+                        cell.bg = colors.background().color(cell.fg, cell.bg);
+                        cell.fg = match_fg;
+                        cell.bg_alpha = 1.0;
+                    }
+
                     // Update URL underlines.
                     urls.update(size_info.cols(), &cell);
 
@@ -525,7 +545,7 @@ impl Display {
         }
 
         if let Some(message) = message_buffer.message() {
-            let search_offset = if search_state.regex().is_some() { 1 } else { 0 };
+            let search_offset = if search_active { 1 } else { 0 };
             let text = message.text(&size_info);
 
             // Create a new rectangle for the background.
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -9,6 +9,7 @@ use std::fs;
 use std::fs::File;
 use std::io::Write;
 use std::mem;
+use std::ops::RangeInclusive;
 use std::path::PathBuf;
 #[cfg(not(any(target_os = "macos", windows)))]
 use std::sync::atomic::Ordering;
@@ -94,6 +95,9 @@ pub struct SearchState {
 
     /// Search origin in viewport coordinates relative to original display offset.
     origin: Point,
+
+    /// Focused match during active search.
+    focused_match: Option<RangeInclusive<Point<usize>>>,
 }
 
 impl SearchState {
@@ -110,6 +114,11 @@ impl SearchState {
     pub fn direction(&self) -> Direction {
         self.direction
     }
+
+    /// Focused match during vi-less search.
+    pub fn focused_match(&self) -> Option<&RangeInclusive<Point<usize>>> {
+        self.focused_match.as_ref()
+    }
 }
 
 impl Default for SearchState {
@@ -118,6 +127,7 @@ impl Default for SearchState {
             direction: Direction::Right,
             display_offset_delta: 0,
             origin: Point::default(),
+            focused_match: None,
             regex: None,
         }
     }
@@ -209,7 +219,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         selection.update(absolute_point, side);
 
         // Move vi cursor and expand selection.
-        if self.terminal.mode().contains(TermMode::VI) {
+        if self.terminal.mode().contains(TermMode::VI) && !self.search_active() {
             self.terminal.vi_mode_cursor.point = point;
             selection.include_all();
         }
@@ -385,6 +395,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         let num_lines = self.terminal.screen_lines();
         let num_cols = self.terminal.cols();
 
+        self.search_state.focused_match = None;
         self.search_state.regex = Some(String::new());
         self.search_state.direction = direction;
 
@@ -393,9 +404,6 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         self.search_state.origin = if self.terminal.mode().contains(TermMode::VI) {
             self.terminal.vi_mode_cursor.point
         } else {
-            // Clear search, since it is used as the active match.
-            self.terminal.selection = None;
-
             match direction {
                 Direction::Right => Point::new(Line(0), Column(0)),
                 Direction::Left => Point::new(num_lines - 2, num_cols - 1),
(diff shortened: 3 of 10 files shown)


chrisduerr (commented):
I'd love to hear what @nixpulvis has to say on this, since he's had some feedback on vi-less search before.

I'm especially curious what the general thoughts on the selection during the search are. We have the ability to highlight the focused match independent from the selection now, but that doesn't mean we can't still use the selection.

So should the selection be moved every time the match is switched, or should it just be moved to the final search match once the selection is canceled?

alacritty/src/input.rs
```diff
@@ -791,7 +788,7 @@ impl<'a, T: EventListener, A: ActionContext<T>> Processor<'a, T, A> {
             };
 
             self.ctx.window_mut().set_mouse_cursor(new_icon);
-        } else if !self.ctx.search_active() {
+        } else {
```
kchibisov, line 791: If I remember that code correctly, you've enabled certain mouse interactions you shouldn't enable, like Paste on middle click, and since your PR just adds ability to select, you should restrict `PasteSelection`.

alacritty/src/display.rs
```diff
@@ -4,6 +4,7 @@
 use std::cmp::min;
 use std::f64;
 use std::fmt::{self, Formatter};
+use std::mem;
```
kchibisov, line 7: Since the function you're using from it is used like once, maybe we can save us one line and use `std::mem::swap`? Though, I don't really mind the `use`.

kchibisov (changes requested):
I like this change, though, I'd like to see active match for Vi mode while I'm typing search as well, iirc you had something working for it already.

alacritty/src/display.rs
```diff
@@ -471,7 +473,18 @@ impl Display {
 
             self.renderer.with_api(&config.ui_config, config.cursor, &size_info, |mut api| {
                 // Iterate over all non-empty cells in the grid.
-                for cell in grid_cells {
+                for mut cell in grid_cells {
+                    // Invert the active match in vi-less search.
+                    if search_state.regex().is_some()
+                        && search_state.focused_match().contains(&Point::new(
+                            display_offset + size_info.screen_lines().0 - cell.line.0 - 1,
```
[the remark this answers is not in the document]
chrisduerr: The problem is that we don't have access to grid or terminal here, so we have no way to use their methods for conversion.

The only option would be to convert the `search_state.focused_match()` range to a `Point` by using the `clamp_buffer_to_visible` method of the grid, but that would require deconstructing and reconstructing the range which would be rather ugly too.

Would look something like this, do you prefer that?

```rust
let start = term.grid().clamp_buffer_to_visible(search_state.focused_match().start());
let end = term.grid().clamp_buffer_to_visible(search_stat.focused_match().end());
let viewport_match = start..=end;

// ...

if search_state.regex().is_some() && viewport_search.contains(&Poin::new(cell.line, cell.column)) {
```

alacritty/src/display.rs
```diff
@@ -471,7 +473,18 @@ impl Display {
 
             self.renderer.with_api(&config.ui_config, config.cursor, &size_info, |mut api| {
                 // Iterate over all non-empty cells in the grid.
-                for cell in grid_cells {
+                for mut cell in grid_cells {
+                    // Invert the active match in vi-less search.
```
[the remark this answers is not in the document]
chrisduerr: Out of curiosity, do you have selection configured to invert too?

alacritty/src/input.rs
chrisduerr: That's a great point, I'll have to look into that. All we want to allow is just selection I think.

alacritty/src/display.rs
chrisduerr: Generally we always import things even if it's just used once, which I think is a good idea because that means you don't have to go hunting for it the next time you add something from `mem` (you'll see immediately that there's an import already.

So I really wouldn't want to remove the import here unless there's a clear advantage to using the full path (like saving yourself a #[cfg()] for example if it's in a method that only exists on one platform).

alacritty/src/display.rs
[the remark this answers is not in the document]
kchibisov: No, actually, but I don't remember why `No`, since I sometimes inver, sometimes don't. You can't use selection as a fallback, since it could confuse users.

alacritty/src/display.rs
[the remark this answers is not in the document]
kchibisov: Yes, I'll prefer it, since you don't reconstruct terminal's internal function implementation, meaning you'll have less chance to break this code if internal function will change.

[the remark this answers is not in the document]
chrisduerr: Yeah, makes sense. This way it just duplicates a bunch of glue code, but no logic.

alacritty/src/event.rs
```diff
@@ -94,6 +95,9 @@ pub struct SearchState {
 
     /// Search origin in viewport coordinates relative to original display offset.
     origin: Point,
+
+    /// Focused match during vi-less search.
+    focused_match: RangeInclusive<Point<usize>>,
```
kchibisov, line 100: Could we make it `Option<RangeInclusive<Point<usize>>>` , since the logic `self.search_state.focused_match != (Point::default()..=Point::default())` looks just bad to me, and we already have a bug from that logic where you could have top left cell highlighted where it shouldn't? Just error prone tbqh.

It's also more logical to have `Option`, since if you have search active, but haven't typed anything you won't have any match, however right now I have an active match for normal search if I don't even type on the bottom left corner. To repro requires having text in the bottom left corner of the terminal(not viewport) and scrolling down one line, so you can see it.

picture

alacritty/src/display.rs
```diff
@@ -471,7 +478,17 @@ impl Display {
...
+                    if search_active && viewport_match.contains(&Point::new(cell.line, cell.column))
+                    {
+                        let colors = config.colors.search.focused_match;
+                        let match_fg = colors.foreground().color(cell.fg, cell.bg);
+                        cell.bg = colors.background().color(cell.fg, cell.bg);
+                        cell.fg = match_fg;
+                        cell.bg_alpha = 1.0;
+                    }
```
kchibisov, line 490: Since we're applying that logic after selection is done you hide selection with active match, which I'm not really a fan, however doing it the right way will mean doing it in renderable cells iter, which is becoming messy, however I'm not sure how to resolve this issue besides applying active match color before selection.

alacritty/src/event.rs
```diff
@@ -110,6 +114,11 @@ impl SearchState {
     pub fn direction(&self) -> Direction {
         self.direction
     }
+
+    /// Focused match during vi-less search.
```
kchibisov, line 118: It's not only Vi-less mode afaics.

CHANGELOG.md
```diff
@@ -100,6 +100,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Unicode 13 support
 - Option to run command on bell which can be set in `bell.command`
 - Fallback to program specified in `$SHELL` variable on Linux/BSD if it is present
+- Ability to make selections while search is active
```
kchibisov, line 103: I've also noticed that you've fixed one minor issue just by enabling our common mouse interactions while search is active.

We were highlighting url in search mode, but we were blocking the opening, however now it opens as expected, so I guess we should mention that, since if blocking url opening was expected we shouldn't highlight it in a first place.

alacritty/src/display.rs
```diff
@@ -437,7 +437,14 @@ impl Display {
         mods: ModifiersState,
         search_state: &SearchState,
     ) {
-        let grid_cells: Vec<RenderableCell> = terminal.renderable_cells(config).collect();
+        // Convert search match from viewport to absolute indexing.
+        let search_active = search_state.regex().is_some();
+        let focused_match = search_state.focused_match();
+        let focused_match_start = terminal.grid().clamp_buffer_to_visible(*focused_match.start());
+        let focused_match_end = terminal.grid().clamp_buffer_to_visible(*focused_match.end());
+        let viewport_match = focused_match_start..=focused_match_end;
```
kchibisov, line 445: If you have scrollback, and you've matched something on top most line(shouldn't be top most though) and then scroll down, so this focused match will go offscreen, you'll have a top left cell with active match colors.

picture

Also, if you don't have scrollback and resize terminal, you'll match top left cell as well, however in that case it's required that you don't have scrollback at all.

resize_top_selected

alacritty_terminal/src/term/mod.rs
```diff
@@ -136,12 +136,44 @@ pub struct RenderableCellsIter<'a, C> {
...
 }
 
+impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
+    type Item = RenderableCell;
+
+    /// Gets the next renderable cell.
+    ///
+    /// Skips empty (background) cells and applies any flags to the cell state
+    /// (eg. invert fg and bg colors).
+    // #[inline]
```
kchibisov, line 153: You should either uncomment this `#[inline]` or remove it.

alacritty_terminal/src/term/mod.rs
```diff
@@ -136,12 +136,44 @@ pub struct RenderableCellsIter<'a, C> {
...
+                // Handle non-cursor cells.
+                let cell = self.inner.next()?;
+                let cell = RenderableCell::new(self, cell);
+
+                // Skip empty cells.
+                if !cell.is_empty() {
+                    return Some(cell);
+                }
+            }
+        }
```
kchibisov, line 173: Nice refactoring of this function.

kchibisov (commented):
Works mostly as I'd expect, maybe except selection precedence, but fixing it is also not really trivial and I'm not sure how it matters.

kchibisov (changes requested):
See above.

alacritty/src/event.rs
chrisduerr: > we already have a bug from that logic where you could have top left cell highlighted where it shouldn't?

I'm not sure that's related to this Default..=Default is not a range that includes anything.

But sure, I'll look into it.

alacritty_terminal/src/term/mod.rs
chrisduerr: Accident, good catch.

alacritty_terminal/src/term/mod.rs
chrisduerr: Yeah it got a tad messy. It's unfortunate that I can't throw the functions in the same block, but goto def should help with that.

alacritty/src/display.rs
```diff
@@ -471,7 +478,22 @@ impl Display {
...
                 // Iterate over all non-empty cells in the grid.
-                for cell in grid_cells {
+                for mut cell in grid_cells {
+                    // Invert the active match in vi-less search.
+                    let cell_point = Point::new(cell.line, cell.column);
+                    if viewport_match
+                        .as_ref()
+                        .map_or(false, |viewport_match| viewport_match.contains(&cell_point))
+                        && !visible_selection
+                            .map_or(false, |selection| selection.contains(cell.column, cell.line))
```
kchibisov, line 488: I thin it shouldn't work with expanded selection, like wide chars.

chrisduerr: You're probably right, yes.

alacritty/src/event.rs
```diff
@@ -420,22 +428,26 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
+        } else if let Some(focused_match) = &self.search_state.focused_match {
+            // Create a selection for the focused match.
+            let start = self.terminal.grid().clamp_buffer_to_visible(*focused_match.start());
+            let end = self.terminal.grid().clamp_buffer_to_visible(*focused_match.end());
+            self.start_selection(SelectionType::Simple, start, Side::Left);
+            self.update_selection(end, Side::Right);
         }
 
+        // Clear focused match.
+        self.search_state.focused_match = None;
```
kchibisov, line 443: Could we move it to `exit_search` function? Right now you have a minor bug in Vi mode when you persist active match after search was 'confirmed', so resetting it there should prevent from such issues?

alacritty/src/event.rs
```diff
@@ -393,9 +404,6 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         self.search_state.origin = if self.terminal.mode().contains(TermMode::VI) {
             self.terminal.vi_mode_cursor.point
         } else {
-            // Clear search, since it is used as the active match.
-            self.terminal.selection = None;
-
```
kchibisov, line 398: I'm not sure if we should keep it tbqh, I like always clearing a selection, since selection could interfere with focused match and you won't see a focused match when you start searching due to selection.

alacritty/src/event.rs
```diff
@@ -393,9 +404,6 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         self.search_state.origin = if self.terminal.mode().contains(TermMode::VI) {
             self.terminal.vi_mode_cursor.point
         } else {
-            // Clear search, since it is used as the active match.
-            self.terminal.selection = None;
```
kchibisov, line 397: I think we should clear selection, so it won't obscure your new focused match? Though we shouldn't do that for Vi mode, since you can move selection via Vi cursor and it's intentional most of the time that way.

kchibisov (changes requested):
Besides that failing test and some minor issues it looks good.

alacritty/src/event.rs
chrisduerr: I'm not sure what you mean?

kchibisov: Start normal search, type something, leave normal search, start search again, starting typing the same request, you should have focused match obscured by selection.

chrisduerr: Hmmm, I do now understand your issue, but I'm not sure if we want to fix that or not. Generally it's trivial either way, but I'm not sure what's the better behavior here. I'll follow up after work today.

alacritty/src/event.rs
```diff
@@ -431,10 +445,8 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     #[inline]
     fn push_search(&mut self, c: char) {
         if let Some(regex) = self.search_state.regex.as_mut() {
-            // Prevent previous search selections from sticking around when not in vi mode.
-            if !self.terminal.mode().contains(TermMode::VI) {
-                self.terminal.selection = None;
-            }
+            // Clear selection so we do not obstruct any matches.
+            self.terminal.selection = None;
```
chrisduerr, line 449: I've decided to do this in `push_search` since this is where it actually matters. There's no need to clear the selection while it is still empty, there's also the possibility that you have something selected after search started and then you don't see it anymore. 

So doing it here makes sure that whenever you type something new, the selection will never obstruct anything.

kchibisov, line 449: We should clear it only for normal search, and keep it in Vi mode, so you can expand selection by searching, like you can in vim.

kchibisov (changes requested):

alacritty/src/event.rs
chrisduerr: Hmm, that's a good point and something I definitely want to be able to do. But the active match problematic is the same for vi mode, isn't it?

Might just better to never reset the selection then, so it behaves uniformly at least?

kchibisov: Or to keep Vi cursor over active match while typing?

chrisduerr: What would that achieve?

kchibisov: Nothing.

Anyway, I'm not sure if we want to have it for normal mode, for Vi mode I think it's not really a problem since your selection updates while you're typing, so it's more clear what is active, etc, and it seems to match what Vi does, since it keeps selection.

kchibisov (approved):

kchibisov (approved):

PR #4396, opened 2020-11-09
Fix github actions clippy job

chrisduerr:


.builds/freebsd.yml
```diff
@@ -24,7 +24,7 @@ tasks:
   - clippy: |
       cd alacritty
       rustup component add clippy
-      cargo clippy --all-targets
+      cargo clippy
```
chrisduerr, line 27: I'm not aware of any benefit of passing `--all-targets` in our current version of Alacritty. Since everything is in the workspace clippy just warns about everything by default as far as I can tell.

chrisduerr: Hold on this please, I think I know why I am using this.

chrisduerr: Correction, this is necessary for running clippy against tests.

kchibisov (approved):

PR #4397, opened 2020-11-10
Fix use after free when dropping extra cell's data

kchibisov:
Commit ec42b42ce601808070462111c0c28edb0e89babb added for
each cell a posibility to carry a pointer on it, thus
old code that was optimizing copying with 'ptr::copy' was
duplicating "unique" pointers ('Box'), which was resulting
in use after free, when trying to perform double free.

This commit fixes this issue by properly cloning cells
and also rewrites affected unsafe blocks in safe rust.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1486,10 +1486,10 @@ impl<T: EventListener> Handler for Term<T> {
         if self.mode.contains(TermMode::INSERT) && self.grid.cursor.point.col + width < num_cols {
             let line = self.grid.cursor.point.line;
             let col = self.grid.cursor.point.col;
-            let line = &mut self.grid[line];
+            let line = &mut self.grid[line][..];
 
-            for col in IndexRange::from(col..(num_cols - width)).rev() {
-                line[col + width] = line[col].clone();
+            for col in (col.0..(num_cols - width).0).rev() {
+                line.swap(col + width, col);
             }
         }
 
@@ -1572,15 +1572,15 @@ impl<T: EventListener> Handler for Term<T> {
         let num_cells = (self.cols() - destination).0;
 
         let line = cursor.point.line;
-        let line = &mut self.grid[line];
+        let line = &mut self.grid[line][..];
 
         for offset in (0..num_cells).rev() {
-            line[destination + offset] = line[source + offset].clone();
+            line.swap(destination.0 + offset, source.0 + offset);
         }
 
         // Cells were just moved out toward the end of the line;
         // fill in between source and dest with blanks.
-        for cell in &mut line[source..destination] {
+        for cell in &mut line[source.0..destination.0] {
             *cell = bg.into();
         }
     }
@@ -1839,15 +1839,15 @@ impl<T: EventListener> Handler for Term<T> {
         let num_cells = (cols - end).0;
 
         let line = cursor.point.line;
-        let line = &mut self.grid[line];
+        let line = &mut self.grid[line][..];
 
         for offset in 0..num_cells {
-            line[start + offset] = line[end + offset].clone();
+            line.swap(start.0 + offset, end.0 + offset);
         }
 
         // Clear last `count` cells in the row. If deleting 1 char, need to delete
         // 1 cell.
-        let end = cols - count;
+        let end = (cols - count).0;
         for cell in &mut line[end..] {
             *cell = bg.into();
         }
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1486,10 +1486,10 @@ impl<T: EventListener> Handler for Term<T> {
         if self.mode.contains(TermMode::INSERT) && self.grid.cursor.point.col + width < num_cols {
             let line = self.grid.cursor.point.line;
             let col = self.grid.cursor.point.col;
-            let line = &mut self.grid[line][..];
+            let row = &mut self.grid[line][..];
 
             for col in (col.0..(num_cols - width).0).rev() {
-                line.swap(col + width, col);
+                row.swap(col + width, col);
             }
         }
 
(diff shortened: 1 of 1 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -1574,18 +1572,15 @@ impl<T: EventListener> Handler for Term<T> {
...
-        let row = &mut self.grid[line];
-
-        unsafe {
-            let src = row[source..].as_ptr();
-            let dst = row[destination..].as_mut_ptr();
+        let line = &mut self.grid[line];
 
-            ptr::copy(src, dst, num_cells);
+        for offset in (0..num_cells).rev() {
+            line[destination + offset] = line[source + offset].clone();
```
chrisduerr, line 1578: Same thing here, this just seems unnecessarily inefficient even for safe Rust. Cloning cells is just extremely wasteful when there's no reason why we would ever need to clone it.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1841,22 +1836,19 @@ impl<T: EventListener> Handler for Term<T> {
...
-        let row = &mut self.grid[line];
-
-        unsafe {
-            let src = row[end..].as_ptr();
-            let dst = row[start..].as_mut_ptr();
+        let line = &mut self.grid[line];
 
-            ptr::copy(src, dst, n);
+        for offset in 0..num_cells {
+            line[start + offset] = line[end + offset].clone();
```
chrisduerr, line 1845: Same thing here.

chrisduerr (changes requested):
This kinda feels a bit lazy to me. I get that we might not want to use unsafe, but that's no reason to ignore performance of the implementation completely.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1488,10 +1488,8 @@ impl<T: EventListener> Handler for Term<T> {
             let col = self.grid.cursor.point.col;
             let line = &mut self.grid[line];
 
-            let src = line[col..].as_ptr();
-            let dst = line[(col + width)..].as_mut_ptr();
-            unsafe {
-                ptr::copy(src, dst, (num_cols - col - width).0);
+            for col in IndexRange::from(col..(num_cols - width)).rev() {
+                line[col + width] = line[col].clone();
```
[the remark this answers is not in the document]
kchibisov: I'm not sure how you'd use swap here and what are you going to swap, here. Also the regions may overlap, so just swap is incorrect at all.

alacritty_terminal/src/term/mod.rs
kchibisov: We can't really swap them because of potentially overlapping regions, the only way we can use swap here is if we do that only for non overlapping region that will be cleared later on, and after we've passed overlapping region we should go back to clone, since we literally cloning after that.

alacritty_terminal/src/term/mod.rs
kchibisov: Same reply here.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1486,10 +1486,10 @@ impl<T: EventListener> Handler for Term<T> {
...
             let line = self.grid.cursor.point.line;
             let col = self.grid.cursor.point.col;
-            let line = &mut self.grid[line];
+            let line = &mut self.grid[line][..];
 
-            for col in IndexRange::from(col..(num_cols - width)).rev() {
-                line[col + width] = line[col].clone();
+            for col in (col.0..(num_cols - width).0).rev() {
+                line.swap(col + width, col);
             }
```
chrisduerr: Honestly I kinda like naming the line number `line` while naming the row itself `row`. I think that separates the two a bit nicer for these instances where both are present. This applies to all changes.

Shadowing the line number with the line slice just seems a bit weird to me?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1572,15 +1572,15 @@ impl<T: EventListener> Handler for Term<T> {
...
-            line[destination + offset] = line[source + offset].clone();
+            line.swap(destination.0 + offset, source.0 + offset);
         }
 
         // Cells were just moved out toward the end of the line;
         // fill in between source and dest with blanks.
-        for cell in &mut line[source..destination] {
+        for cell in &mut line[source.0..destination.0] {
             *cell = bg.into();
         }
```
chrisduerr: Can't these two be combined? It seems like a shame to iterate twice just to insert the blanks when really we already swap them into place.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1839,15 +1839,15 @@ impl<T: EventListener> Handler for Term<T> {
...
+            line.swap(start.0 + offset, end.0 + offset);
         }
 
         // Clear last `count` cells in the row. If deleting 1 char, need to delete
         // 1 cell.
-        let end = cols - count;
+        let end = (cols - count).0;
         for cell in &mut line[end..] {
             *cell = bg.into();
         }
```
chrisduerr: These could probably also be combined maybe?

chrisduerr (changes requested):
Functionality of everything looks good, accepts wide characters and normal characters and can deal with massive and tiny values without any panics in my tests. Behavior also seems to match master which is good, with the exception that it's actually safe.

alacritty_terminal/src/term/mod.rs
kchibisov: I've tried but it looked ugly a bit, since you should check for both boundaries in a loop, while here you check for them in a loop condition...

alacritty_terminal/src/term/mod.rs
kchibisov: Only if you start dynamically checking in a loop for boundaries, since those loops iterate for difference amount of cell.

alacritty_terminal/src/term/mod.rs
chrisduerr: Fair enough, it's worked well so far so that should be fine I suppose. Though I'd imagine having a check inside the loop would save us the increment/decrement of the counter we need to use with two loops, but compared to how expensive these loops are with the `swap` that is completely insignificant compared to possible code style advantages.

chrisduerr (approved):

chrisduerr:
Needs a rebase but should be good to go.

Issue #4404, opened 2020-11-11
Cursor vanishes when there're a lot of altases around


kchibisov:
To repro you can 'cat' symbols file from vtebench unicode benchmark, enter vi mode, and then trying to enable selection, the cursor will vanish on certain glyphs, but will wok on other glyphs.

![picture](https://user-images.githubusercontent.com/27620401/98787870-1dec4700-2411-11eb-90df-d66a67ac747d.png)

PR #4405, opened 2020-11-11
Fix wide characters being cut off

### LINKED ISSUE #791: Glyphs sometimes cut off (need to draw background in single pass)

kchibisov:
Fixes #791.

### DIFF
--- a/alacritty/res/text.f.glsl
+++ b/alacritty/res/text.f.glsl
@@ -9,31 +9,29 @@ layout(location = 0, index = 1) out vec4 alphaMask;
 
 uniform sampler2D mask;
 
-void main()
-{
+void main() {
     if (backgroundPass != 0) {
-        if (bg.a == 0.0)
+        if (bg.a == 0.0) {
             discard;
+        }
 
         alphaMask = vec4(1.0);
         color = vec4(bg.rgb, 1.0);
-    } else {
-        if (fg.a != 0.0) {
-            // Color glyphs, like emojis.
-            vec4 glyphColor = texture(mask, TexCoords);
-            alphaMask = vec4(glyphColor.a);
+    } else if (int(fg.a) > 1) {
+        // Color glyphs, like emojis.
+        vec4 glyphColor = texture(mask, TexCoords);
+        alphaMask = vec4(glyphColor.a);
 
-            // Revert alpha premultiplication.
-            if (glyphColor.a != 0) {
-                glyphColor.rgb = vec3(glyphColor.rgb / glyphColor.a);
-            }
-
-            color = vec4(glyphColor.rgb, 1.0);
-        } else {
-            // Regular text glyphs.
-            vec3 textColor = texture(mask, TexCoords).rgb;
-            alphaMask = vec4(textColor, textColor.r);
-            color = vec4(fg.rgb, 1.0);
+        // Revert alpha premultiplication.
+        if (glyphColor.a != 0) {
+            glyphColor.rgb = vec3(glyphColor.rgb / glyphColor.a);
         }
+
+        color = vec4(glyphColor.rgb, 1.0);
+    } else {
+        // Regular text glyphs.
+        vec3 textColor = texture(mask, TexCoords).rgb;
+        alphaMask = vec4(textColor, textColor.r);
+        color = vec4(fg.rgb, 1.0);
     }
 }
--- a/alacritty/res/text.f.glsl
+++ b/alacritty/res/text.f.glsl
@@ -9,6 +9,8 @@ layout(location = 0, index = 1) out vec4 alphaMask;
 
 uniform sampler2D mask;
 
+#define COLORED 2
+
 void main() {
     if (backgroundPass != 0) {
         if (bg.a == 0.0) {
@@ -17,7 +19,7 @@ void main() {
 
         alphaMask = vec4(1.0);
         color = vec4(bg.rgb, 1.0);
-    } else if (int(fg.a) > 1) {
+    } else if ((int(fg.a) & COLORED) != 0) {
         // Color glyphs, like emojis.
         vec4 glyphColor = texture(mask, TexCoords);
         alphaMask = vec4(glyphColor.a);
--- a/alacritty/res/text.v.glsl
+++ b/alacritty/res/text.v.glsl
@@ -1,18 +1,24 @@
 #version 330 core
 // Cell properties.
-layout (location = 0) in vec2 gridCoords;
+layout(location = 0) in vec2 gridCoords;
 
 // Glyph properties.
-layout (location = 1) in vec4 glyph;
+layout(location = 1) in vec4 glyph;
 
 // uv mapping.
-layout (location = 2) in vec4 uv;
+layout(location = 2) in vec4 uv;
 
-// Text foreground rgb packed together with multicolor flag.
-layout (location = 3) in vec4 textColor;
+// Text foreground rgb packed together with cell flags. textColor.a
+// is a bitflag with the following values.
+//
+// 0 - empty
+// 1 - wide char
+// 2 - colored glyph
+//
+layout(location = 3) in vec4 textColor;
 
 // Background color.
-layout (location = 4) in vec4 backgroundColor;
+layout(location = 4) in vec4 backgroundColor;
 
 out vec2 TexCoords;
 flat out vec4 fg;
@@ -24,9 +30,9 @@ uniform vec4 projection;
 
 uniform int backgroundPass;
 
+#define WIDE_CHAR 1
 
-void main()
-{
+void main() {
     vec2 projectionOffset = projection.xy;
     vec2 projectionScale = projection.zw;
 
@@ -39,8 +45,15 @@ void main()
     vec2 cellPosition = cellDim * gridCoords;
 
     if (backgroundPass != 0) {
-        vec2 finalPosition = cellPosition + cellDim * position;
-        gl_Position = vec4(projectionOffset + projectionScale * finalPosition, 0.0, 1.0);
+        vec2 backgroundDim = cellDim;
+        // We use 1 to identify wide chars.
+        if ((int(textColor.a) & WIDE_CHAR) != 0) {
+            // Update wide char x dimension so it'll cover following spacer.
+            backgroundDim.x *= 2;
+        }
+        vec2 finalPosition = cellPosition + backgroundDim * position;
+        gl_Position =
+            vec4(projectionOffset + projectionScale * finalPosition, 0.0, 1.0);
 
         TexCoords = vec2(0, 0);
     } else {
@@ -49,7 +62,8 @@ void main()
         glyphOffset.y = cellDim.y - glyphOffset.y;
 
         vec2 finalPosition = cellPosition + glyphSize * position + glyphOffset;
-        gl_Position = vec4(projectionOffset + projectionScale * finalPosition, 0.0, 1.0);
+        gl_Position =
+            vec4(projectionOffset + projectionScale * finalPosition, 0.0, 1.0);
 
         vec2 uvOffset = uv.xy;
         vec2 uvSize = uv.zw;
--- a/alacritty/res/text.v.glsl
+++ b/alacritty/res/text.v.glsl
@@ -9,8 +9,8 @@ layout(location = 1) in vec4 glyph;
 layout(location = 2) in vec4 uv;
 
 // Text foreground rgb packed together with cell flags. textColor.a
-// is a bitflag. For the actual possible values of bitflags consult
-// RenderingGlyphFlags in renderer/mod.rs.
+// are the bitflags; consult RenderingGlyphFlags in renderer/mod.rs
+// for the possible values.
 layout(location = 3) in vec4 textColor;
 
 // Background color.
@@ -43,7 +43,7 @@ void main() {
     if (backgroundPass != 0) {
         vec2 backgroundDim = cellDim;
         if ((int(textColor.a) & WIDE_CHAR) != 0) {
-            // Update wide char x dimension so it'll cover following spacer.
+            // Update wide char x dimension so it'll cover the following spacer.
             backgroundDim.x *= 2;
         }
         vec2 finalPosition = cellPosition + backgroundDim * position;
--- a/alacritty/res/text.v.glsl
+++ b/alacritty/res/text.v.glsl
@@ -9,12 +9,8 @@ layout(location = 1) in vec4 glyph;
 layout(location = 2) in vec4 uv;
 
 // Text foreground rgb packed together with cell flags. textColor.a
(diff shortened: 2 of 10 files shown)


kchibisov:
@chrisduerr I've addressed your comment from irc, would you mind taking a look?

alacritty_terminal/src/term/mod.rs
```diff
@@ -165,8 +165,8 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
                 let cell = self.inner.next()?;
                 let cell = RenderableCell::new(self, cell);
 
-                // Skip empty cells.
-                if !cell.is_empty() {
+                // Skip empty cells and wide chars spacers.
```
chrisduerr, line 168:
```suggestion
                // Skip empty cells and wide char spacers.
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -165,8 +165,8 @@ impl<'a, C> Iterator for RenderableCellsIter<'a, C> {
                 let cell = self.inner.next()?;
                 let cell = RenderableCell::new(self, cell);
```
chrisduerr, line 166: Since this doesn't affect our cursor path, does this mean we still need all our special selection logic solely to make the block cursor work?

alacritty/src/renderer/rects.rs
```diff
@@ -184,3 +184,13 @@ impl RenderLines {
...
+
+#[inline]
+fn into_cell_end(cell: &RenderableCell) -> Point {
+    let mut end: Point = cell.into();
+    if cell.flags.contains(Flags::WIDE_CHAR) {
+        end.col += 1;
+    }
+
+    end
+}
```
chrisduerr, line 196: One way or another this gets invoked anyways. Can't we just calculate the point after the `!cell.flags.contains(flag)` check and then always use that? Just as efficient and sounds much cleaner.

alacritty/src/renderer/mod.rs
```diff
@@ -441,6 +458,15 @@ impl Batch {
...
         }
 
+        let mut cell_flags = RenderingGlyphFlags::empty();
+        if glyph.multicolor != 0 {
+            cell_flags |= RenderingGlyphFlags::COLORED;
+        }
+
+        if cell.flags.contains(Flags::WIDE_CHAR) {
+            cell_flags |= RenderingGlyphFlags::WIDE_CHAR;
+        }
```
chrisduerr, line 468: Can we use the methods provided by bitflags? They should be easier to understand and thus maintain.

Even better if we just do `cell_flags.set(COLORED, glyph.multicolor != 0)`. Then it's also shorter!

alacritty/src/renderer/mod.rs
```diff
@@ -359,30 +360,46 @@ impl GlyphCache {
...
     uv_height: f32,
+
     // Color.
     r: u8,
     g: u8,
     b: u8,
-    // Flag indicating that a glyph uses multiple colors; like an Emoji.
-    multicolor: u8,
+
+    // Cell flags.
```
chrisduerr, line 400: This comment is useless, if we want to add a comment here, we should provide additional value over just the variable name. An example for what is stored in the flags at the moment could be helpful (Cell flags like multicolor or fullwidth character). Otherwise we might as well just remove it.

alacritty/src/renderer/mod.rs
```diff
@@ -359,30 +360,46 @@ impl GlyphCache {
     }
 }
 
+// XXX these flags must be in sync with their usage in text.*.glsl shaders.
```
chrisduerr, line 363:
```suggestion
// NOTE: These flags must be in sync with their usage in the text.*.glsl shaders.
```

I find `NOTE` a bit nicer than the unspecific `XXX`, should be highlighted just the same in vim at least.

alacritty/res/text.v.glsl
```diff
@@ -39,8 +45,15 @@ void main()
     vec2 cellPosition = cellDim * gridCoords;
 
     if (backgroundPass != 0) {
-        vec2 finalPosition = cellPosition + cellDim * position;
-        gl_Position = vec4(projectionOffset + projectionScale * finalPosition, 0.0, 1.0);
+        vec2 backgroundDim = cellDim;
+        // We use 1 to identify wide chars.
```
chrisduerr, line 49:
```suggestion
```

The constant tells us why, the `1` this comment talks about is nowhere to be found.

alacritty/res/text.v.glsl
```diff
@@ -1,18 +1,24 @@
...
 
-// Text foreground rgb packed together with multicolor flag.
-layout (location = 3) in vec4 textColor;
+// Text foreground rgb packed together with cell flags. textColor.a
+// is a bitflag with the following values.
+//
+// 0 - empty
+// 1 - wide char
+// 2 - colored glyph
+//
```
chrisduerr, line 17:
```suggestion
// Text foreground rgb packed together with cell flags. textColor.a
// is a bitflag with the following values.
//
// 0 - empty
// 1 - wide char
// 2 - colored glyph
```

We usually don't add spaces after our comments.

Also I'm not sure if we want to document the actual *values* of the cell flags, that seems a bit strange to me for bitflags. Instead we should document the individual bits? Or how would this list continue, I'd assume with `4 - whatever`?

alacritty/res/text.f.glsl
```diff
@@ -9,31 +9,29 @@ layout(location = 0, index = 1) out vec4 alphaMask;
...
-        if (fg.a != 0.0) {
-            // Color glyphs, like emojis.
-            vec4 glyphColor = texture(mask, TexCoords);
-            alphaMask = vec4(glyphColor.a);
-
-            // Revert alpha premultiplication.
-            if (glyphColor.a != 0) {
-                glyphColor.rgb = vec3(glyphColor.rgb / glyphColor.a);
-            }
+    } else if (int(fg.a) > 1) {
```
chrisduerr, line 20: This breaks as soon as another bitflag is added? Can we do the same we did for wide chars and use a constant and a bitwise and?

chrisduerr (changes requested):

alacritty/src/renderer/mod.rs
kchibisov: I've just followed other useless comments here.

alacritty_terminal/src/term/mod.rs
kchibisov: I'm not quite sure about what are you talking about here, unfortunately. What do you mean by 'work', like expand over normal chars? If that's the case, I think we may apply all coloring where block cursor is to the cell below it? And then in selection just not mark this cell as selected?

chrisduerr: What is the reason why the logic for checking if a wide char spacer is selected is still in this code? It shouldn't be necessary anymore.

alacritty/src/renderer/mod.rs
```diff
@@ -441,6 +458,10 @@ impl Batch {
             self.tex = glyph.tex_id;
         }
 
+        let mut cell_flags = RenderingGlyphFlags::empty();
+        cell_flags.set(RenderingGlyphFlags::COLORED, glyph.multicolor != 0);
```
chrisduerr, line 462: Does `multicolor` actually need to be stored in the `Glyph` as a u8? Glyphs itself aren't passed to OpenGL, right?

alacritty/res/text.v.glsl
```diff
@@ -39,8 +41,14 @@ void main()
     vec2 cellPosition = cellDim * gridCoords;
 
     if (backgroundPass != 0) {
-        vec2 finalPosition = cellPosition + cellDim * position;
-        gl_Position = vec4(projectionOffset + projectionScale * finalPosition, 0.0, 1.0);
+        vec2 backgroundDim = cellDim;
+        if ((int(textColor.a) & WIDE_CHAR) != 0) {
+            // Update wide char x dimension so it'll cover following spacer.
```
chrisduerr, line 46:
```suggestion
            // Update wide char x dimension so it'll cover the following spacer.
```

alacritty/res/text.v.glsl
```diff
@@ -1,18 +1,20 @@
...
 
 // uv mapping.
-layout (location = 2) in vec4 uv;
+layout(location = 2) in vec4 uv;
 
-// Text foreground rgb packed together with multicolor flag.
-layout (location = 3) in vec4 textColor;
+// Text foreground rgb packed together with cell flags. textColor.a
+// is a bitflag. For the actual possible values of bitflags consult
+// RenderingGlyphFlags in renderer/mod.rs.
```
chrisduerr, line 13:
```suggestion
// Text foreground rgb packed together with cell flags. textColor.a
// are the bitflags; consult RenderingGlyphFlags in renderer/mod.rs for the possible values.
```

chrisduerr (changes requested):

alacritty/src/renderer/mod.rs
kchibisov: Well, in the current state of things it should be stored there, since that's what we load from rasterizer and the only place where we know that bufffer is RGBA/RGB. They are not passed, but we use them to generate instance data, as you can see here.

alacritty_terminal/src/term/mod.rs
kchibisov: Well, you can't really just remove it, it'll require a slight rewrite, I'll look into it once I have time.

alacritty_terminal/src/term/mod.rs
```diff
@@ -282,18 +282,6 @@ impl<'a, C> RenderableCellsIter<'a, C> {
                 // Check line-wrapping, leading spacer.
```
chrisduerr, line 282: Why is the trailing spacer check still made here?

alacritty_terminal/src/term/mod.rs
```diff
@@ -282,18 +282,6 @@ impl<'a, C> RenderableCellsIter<'a, C> {
...
                 || (self.grid[buffer_prev].flags.contains(Flags::LEADING_WIDE_CHAR_SPACER)
                     && selection.contains(prev.col, prev.line))
-        } else if cell.flags.contains(Flags::WIDE_CHAR_SPACER) {
```
chrisduerr, line 285: Does this work if only the wide char spacer is selected? I thought this might be the part we might need to keep, since we still need to mark the wide char cell when the spacer is selected.

alacritty/src/renderer/mod.rs
chrisduerr: Yeah, but why not convert them from bool to u8 here? It seems strange that it is stored as u8 on the Glyph?

I'm suggesting this:

```diff
diff --git a/alacritty/src/renderer/mod.rs b/alacritty/src/renderer/mod.rs
index b347556..c66e188 100644
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -123,7 +123,7 @@ pub struct RectShaderProgram {
 #[derive(Copy, Debug, Clone)]
 pub struct Glyph {
     tex_id: GLuint,
-    multicolor: u8,
+    multicolor: bool,
     top: i16,
     left: i16,
     width: i16,
@@ -463,7 +463,7 @@ impl Batch {
             bg_g: cell.bg.g,
             bg_b: cell.bg.b,
             bg_a: (cell.bg_alpha * 255.0) as u8,
-            multicolor: glyph.multicolor,
+            multicolor: glyph.multicolor as u8,
         });
     }
 
@@ -1067,7 +1067,7 @@ fn load_glyph(
         },
         Err(AtlasInsertError::GlyphTooLarge) => Glyph {
             tex_id: atlas[*current_atlas].id,
-            multicolor: 0,
+            multicolor: false,
             top: 0,
             left: 0,
             width: 0,
@@ -1581,7 +1581,7 @@ impl Atlas {
 
         Glyph {
             tex_id: self.id,
-            multicolor: multicolor as u8,
+            multicolor: multicolor,
             top: glyph.top as i16,
             left: glyph.left as i16,
             width: width as i16,
```

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
kchibisov: it's actually what it does, if spacer is selected it selects wide char, the branch I've removed was when wide char is selected -> select spacer, which should automatically work anyway. If you've reached this `if` it means that your wide char isn't selected, thus you check trailing spacer to select a wide char, so everything should be kept this way.

OJFord:
Thanks for working on this. I checked it out and cargo built to try, but I'm still seeing (running `target/debug/alacritty`):
image


Is there something else I need to do, or maybe I jumped the gun and it's not expected to be ready yet?

Cheers.

kchibisov:
The issue isn't about overlapping but about wide characters being cut off after some point. You may see how wide chars look at https://github.com/alacritty/alacritty/issues/4404#issue-740572459 when Korean glyphs are missing half of it, so this PR aims to solve it.

Your issue is that multidot char is from some non monospace font, since it has a width of 1, however it occupies 2 cells, thus it's not a wide char per se( since those have a width of 2). I'd suggest to use monospace font for it.

alacritty/src/renderer/mod.rs
```diff
@@ -359,30 +360,46 @@ impl GlyphCache {
...
     uv_height: f32,
+
     // Color.
     r: u8,
     g: u8,
     b: u8,
-    // Flag indicating that a glyph uses multiple colors; like an Emoji.
-    multicolor: u8,
+
+    // Cell flags like multicolor of fullwidth character.
```
chrisduerr, line 400:
```suggestion
    // Cell flags like multicolor or fullwidth character.
```

chrisduerr (changes requested):

OJFord:
Thanks, sorry for the noise.

It is (supposed to be?) a monospace font - I rebuild my Iosevka font at the weekend to remove nerd font patching (and try using a fallback font instead), so I assumed this glyph was from the latter and it was a rendering issue that hadn't occurred when it was patched. But it turns out it's actually from (unpatched) Iosevka anyway, not the symbol fallback font, so something wrong with my build maybe. (I was using (patched) Iosevka before and this was fine, hence my assumption that it was the other font causing the problem.)

Cheers.

chrisduerr (approved):
LGTM.

chrisduerr:
Can confirm that the render timer seems to be identical with a dGPU on macOS. Feel free to rebase and merge @kchibisov.

PR #4414, opened 2020-11-13
Bump development version to 0.7.0-dev

chrisduerr:


CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 0.6.0-dev
+## 0.7.0-dev
+
+## 0.6.0
```
chrisduerr, line 10: This change is mostly just a suggestion to simplify our release process. That way there should be less commits on master that could be confused with a release.

kchibisov: Well, if we add cherry picks, we will have the same amount of commits on master?

kchibisov (changes requested):
Cargo.lock isn't updated.

kchibisov (changes requested):
I'm still not sure if we should change anything about release process, since you only save commits in a best case, if we backport changelog entries you won't save anything.

chrisduerr:
> I'm still not sure if we should change anything about release process, since you only save commits in a best case, if we backport changelog entries you won't save anything.

Why wouldn't we save anything when we backport fixes? Those would just add a single line to the previous release? That should be unrelated, right?

kchibisov:
>Why wouldn't we save anything when we backport fixes? Those would just add a single line to the previous release? That should be unrelated, right?

I mean that if the commit that backports fixes, is the one that removes `-dev`, so instead we'll have a commit that just backports fixes, so in total the same amount of them.

it's wrt 
>That way there should be less commits on master

As for 

>confused with a release.

The actual commit that confuses with release, is the one that bumps to `0.x.0-dev` version.

chrisduerr:
> I mean that if the commit that backports fixes, is the one that removes -dev, so instead we'll have a commit that just backports fixes, so in total the same amount of them.

But fixes are never ported from the branch to the master, but always ported from master to the branch? I'm not sure what "backporting" you're talking about? There is nothing that ever flows from the release branch back to the master branch.

kchibisov:
But fixes are never ported from the branch to the master, but always ported from master to the branch? I'm not sure what "backporting" you're talking about? There is nothing that ever flows from the release branch back to the master branch.

changelog entries are backported, like they move from -dev to release, if you cherry pick fixes.

CONTRIBUTING.md
```diff
@@ -123,6 +123,7 @@ The exact steps for an exemplary `0.2.0` release might look like this:
   1. Initially, the version on the latest master is `0.2.0-dev`
   2. A new `v0.2.0` branch is created for the release
   3. On master, the version is bumped to `0.3.0-dev`
+     and the `-dev` suffix is removed from the `0.2.0` changelog
```
kchibisov, line 126: Should it be `0.2.0-dev`?

chrisduerr: Either way would be fine, but we can clarify, sure. This is more talking about the "changelog of the 0.2.0 version".

kchibisov (approved):

PR #4417, opened 2020-11-14
Draw cursor with rect renderer

### LINKED ISSUE #3471: Underline cursor cannot be seen on underlined text.
Hello, 
I am using emacs in the cli and it uses underline as the highlight (for linting and such). When I place the cursor on the underlined text the cursor is hidden by the underline.

Is there a way to somehow render the cursor in a different color to enable it to be seen? I tried the different settings and checked the issues but I could not find the answer.

Maybe the simplest solution would be to set the underline color to the cursor color with priority. Or just reorder rendering and firstly draw underline and override it with the cursor.

## Step to reproduce
I set the cursor color to pink to better visualize it.
image

Now when I go into tmux `copy-mode` and move the cursor to the first letter of the word "[u]nderline"

image

### System

OS: Linux
Version: alacritty 0.4.1
Linux: Wayland, sway

### Logs

Font/Terminal size: 
```
[2020-03-16 21:13] [INFO] Welcome to Alacritty
[2020-03-16 21:13] [INFO] Configuration loaded from "<home_dir>/.config/alacritty/alacritty.yml"
[2020-03-16 21:13] [DEBUG] Estimated DPR: 1
[2020-03-16 21:13] [DEBUG] Estimated Cell Size: 11 x 20
[2020-03-16 21:13] [DEBUG] Estimated Dimensions: None
[2020-03-16 21:13] [INFO] Device pixel ratio: 1
[2020-03-16 21:13] [INFO] Initializing glyph cache...
[2020-03-16 21:13] [INFO] ... finished initializing glyph cache in 0.010245541s
[2020-03-16 21:13] [INFO] Cell Size: 11 x 20
[2020-03-16 21:13] [INFO] Padding: 0 x 0
[2020-03-16 21:13] [INFO] Width: 800, Height: 600
[2020-03-16 21:13] [INFO] PTY Dimensions: Line(30) x Column(72)
[2020-03-16 21:13] [INFO] Initialisation complete
[2020-03-16 21:13] [DEBUG] New num_cols is 97 and num_lines is 47
[2020-03-16 21:13] [INFO] Width: 1076, Height: 944
```

### LINKED ISSUE #4404: Cursor vanishes when there're a lot of altases around

kchibisov:
This commit makes cursors being drawn via rects, thus it's always above
underlines/strikeouts. Also, since cursor isn't a glyph anymore, it
can't be obscured due to atlas switching during glyphs drawing.

Fixes #4404.
Fixes #3471.

### DIFF
--- a/alacritty/src/cursor.rs
+++ b/alacritty/src/cursor.rs
@@ -1,112 +1,92 @@
-//! Helpers for creating different cursor glyphs from font metrics.
-
-use crossfont::{BitmapBuffer, Metrics, RasterizedGlyph};
+//! Convert a cursor into an iterator of rects.
 
 use alacritty_terminal::ansi::CursorShape;
+use alacritty_terminal::term::color::Rgb;
+use alacritty_terminal::term::render::RenderableCursor;
+use alacritty_terminal::term::SizeInfo;
 
-pub fn get_cursor_glyph(
-    cursor: CursorShape,
-    metrics: Metrics,
-    offset_x: i8,
-    offset_y: i8,
-    is_wide: bool,
-    cursor_thickness: f32,
-) -> RasterizedGlyph {
-    // Calculate the cell metrics.
-    //
-    // NOTE: With Rust 1.47+ `f64 as usize` is defined to clamp automatically:
-    // https://github.com/rust-lang/rust/commit/14d608f1d8a0b84da5f3bccecb3efb3d35f980dc
-    let height = (metrics.line_height + f64::from(offset_y)).max(1.) as usize;
-    let mut width = (metrics.average_advance + f64::from(offset_x)).max(1.) as usize;
-    let line_width = (cursor_thickness * width as f32).round().max(1.) as usize;
+use crate::renderer::rects::RenderRect;
 
-    // Double the cursor width if it's above a double-width glyph.
-    if is_wide {
-        width *= 2;
-    }
-
-    match cursor {
-        CursorShape::HollowBlock => get_box_cursor_glyph(height, width, line_width),
-        CursorShape::Underline => get_underline_cursor_glyph(width, line_width),
-        CursorShape::Beam => get_beam_cursor_glyph(height, line_width),
-        CursorShape::Block => get_block_cursor_glyph(height, width),
-        CursorShape::Hidden => RasterizedGlyph::default(),
-    }
+/// Trait for conversion into the iterator.
+pub trait IntoRects {
+    /// Consume the cursor for an iterator of rects.
+    fn rects(self, size_info: &SizeInfo, thickness: f32) -> CursorRects;
 }
 
-/// Return a custom underline cursor character.
-pub fn get_underline_cursor_glyph(width: usize, line_width: usize) -> RasterizedGlyph {
-    // Create a new rectangle, the height is relative to the font width.
-    let buffer = BitmapBuffer::RGB(vec![255u8; width * line_width * 3]);
+impl IntoRects for RenderableCursor {
+    fn rects(self, size_info: &SizeInfo, thickness: f32) -> CursorRects {
+        let point = self.point();
+        let x = point.col.0 as f32 * size_info.cell_width() + size_info.padding_x();
+        let y = point.line.0 as f32 * size_info.cell_height() + size_info.padding_y();
 
-    // Create a custom glyph with the rectangle data attached to it.
-    RasterizedGlyph {
-        character: ' ',
-        top: line_width as i32,
-        left: 0,
-        height: line_width as i32,
-        width: width as i32,
-        buffer,
-    }
-}
+        let mut width = size_info.cell_width();
+        let height = size_info.cell_height();
 
-/// Return a custom beam cursor character.
-pub fn get_beam_cursor_glyph(height: usize, line_width: usize) -> RasterizedGlyph {
-    // Create a new rectangle that is at least one pixel wide
-    let buffer = BitmapBuffer::RGB(vec![255u8; line_width * height * 3]);
+        if self.is_wide() {
+            width *= 2.;
+        }
 
-    // Create a custom glyph with the rectangle data attached to it
-    RasterizedGlyph {
-        character: ' ',
-        top: height as i32,
-        left: 0,
-        height: height as i32,
-        width: line_width as i32,
-        buffer,
-    }
-}
+        let thickness = (thickness * width as f32).round().max(1.);
 
-/// Returns a custom box cursor character.
-pub fn get_box_cursor_glyph(height: usize, width: usize, line_width: usize) -> RasterizedGlyph {
-    // Create a new box outline rectangle.
-    let mut buffer = Vec::with_capacity(width * height * 3);
-    for y in 0..height {
-        for x in 0..width {
-            if y < line_width
-                || y >= height - line_width
-                || x < line_width
-                || x >= width - line_width
-            {
-                buffer.append(&mut vec![255u8; 3]);
-            } else {
-                buffer.append(&mut vec![0u8; 3]);
-            }
+        match self.shape() {
+            CursorShape::Beam => beam(x, y, height, thickness, self.color()),
+            CursorShape::Underline => underline(x, y, width, height, thickness, self.color()),
+            CursorShape::HollowBlock => hollow(x, y, width, height, thickness, self.color()),
+            _ => CursorRects::default(),
         }
     }
+}
 
-    // Create a custom glyph with the rectangle data attached to it.
-    RasterizedGlyph {
-        character: ' ',
-        top: height as i32,
-        left: 0,
-        height: height as i32,
-        width: width as i32,
-        buffer: BitmapBuffer::RGB(buffer),
+/// Cursor rect iterator.
+#[derive(Default)]
+pub struct CursorRects {
+    rects: [Option<RenderRect>; 4],
+    index: usize,
+}
+
+impl From<RenderRect> for CursorRects {
+    fn from(rect: RenderRect) -> Self {
+        Self { rects: [Some(rect), None, None, None], index: 0 }
     }
 }
 
-/// Return a custom block cursor character.
-pub fn get_block_cursor_glyph(height: usize, width: usize) -> RasterizedGlyph {
-    // Create a completely filled glyph.
-    let buffer = BitmapBuffer::RGB(vec![255u8; width * height * 3]);
+impl Iterator for CursorRects {
+    type Item = RenderRect;
 
-    // Create a custom glyph with the rectangle data attached to it.
-    RasterizedGlyph {
-        character: ' ',
-        top: height as i32,
-        left: 0,
-        height: height as i32,
-        width: width as i32,
-        buffer,
+    fn next(&mut self) -> Option<Self::Item> {
+        let rect = self.rects.get_mut(self.index)?;
+        self.index += 1;
+        rect.take()
+    }
+}
+
+/// Create an iterator yielding a single beam rect.
+fn beam(x: f32, y: f32, height: f32, thickness: f32, color: Rgb) -> CursorRects {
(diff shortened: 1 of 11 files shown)


chrisduerr:
Just applying inverse to the cell's below the cursor and having a different method to get the cursor glyphs seems fine to me.

kchibisov:
I've updated everything, but I'd like to merge it after #4452, so it'd draw things faster.

chrisduerr:
> I've updated everything, but I'd like to merge it after #4452, so it'd draw things faster.

I'm sorry, but I don't understand. Even if this is merged before #4452, this would still become faster once #4452 is merged, right? Why should this wait until a PR opened after it is merged? Are there some changes in #4452 that you'd like to explicitly make use of, that wouldn't have to be made just by rebasing #4452 on top of this?

kchibisov:
>I'm sorry, but I don't understand. Even if this is merged before #4452, this would still become faster once #4452 is merged, right? Why should this wait until a PR opened after it is merged? Are there some changes in #4452 that you'd like to explicitly make use of, that wouldn't have to be made just by rebasing #4452 on top of this?

I just want things a bit faster and don't want to invoke rect renderer without any rect being present.

kchibisov:
>I've updated everything, but I'd like to merge it after #4452, so it'd draw things faster.

this one is merged.

alacritty/src/cursor.rs
```diff
@@ -1,112 +1,66 @@
...
+use alacritty_terminal::ansi::CursorShape;
+use alacritty_terminal::term::color::Rgb;
+use alacritty_terminal::term::{CursorKey, SizeInfo};
 
-use crossfont::{BitmapBuffer, Metrics, RasterizedGlyph};
+use crate::renderer::rects::RenderRect;
 
-use alacritty_terminal::ansi::CursorShape;
+pub struct Cursor {
+    /// Coordinates of top left corner of the cursor cell.
```
chrisduerr, line 9:
```suggestion
    /// Coordinates of the top left corner of the cursor cell.
```

alacritty/src/cursor.rs
```diff
@@ -1,112 +1,66 @@
...
-            if y < line_width
-                || y >= height - line_width
-                || x < line_width
-                || x >= width - line_width
-            {
-                buf.append(&mut vec![255u8; 3]);
-            } else {
-                buf.append(&mut vec![0u8; 3]);
-            }
+    pub fn into_rects(self, size_info: &SizeInfo) -> Vec<RenderRect> {
```
chrisduerr, line 25: Multiple rects for the box cursor, correct?

alacritty/src/cursor.rs
```diff
@@ -1,112 +1,66 @@
...
+        let height = size_info.cell_height();
+        let line_width = (self.thickness * width as f64).round().max(1.) as f32;
+
+        if self.key.is_wide {
+            width *= 2.;
+        }
+
+        match self.key.shape {
+            CursorShape::Beam => self.beam(height, line_width),
+            CursorShape::Block | CursorShape::Hidden => vec![],
```
chrisduerr, line 36:
```suggestion
            CursorShape::Block | CursorShape::Hidden => Vec::new(),
```

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
```diff
@@ -201,12 +201,20 @@ impl<'a, C> RenderableCellsIter<'a, C> {
...
+                _ => {
+                    let fg = cell.fg;
+                    cell.fg = self.cursor.text_color.color(fg, cell.bg);
+                    cell.bg = self.cursor.cursor_color.color(fg, cell.bg);
+                },
+            }
+
+            // Since we draw Block cursor by drawing cell below it with a proper color,
+            // we must adjust alpha to make it visible.
+            cell.bg_alpha = 1.;
```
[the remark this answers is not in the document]
kchibisov: next_cursor_cell is the cell below cursor, and next_cursor is a cursor cell, so I modify the cell below it? I don't see a problem here.

alacritty/src/cursor.rs
kchibisov: Yes, however what do you think about passing a mutable reference to a vector? I think it should be more clear, and especially since reduce allocations on each draw.

alacritty/src/renderer/mod.rs
```diff
@@ -823,21 +808,10 @@ impl<'a> RenderApi<'a> {
...
-                        cursor_key.is_wide,
-                        self.cursor_config.thickness(),
-                    ))
-                });
-                self.add_render_item(&cell, glyph);
-                return;
+            RenderableCellContent::Cursor(_) => {
+                // TODO maybe yield cells that can't be rendered and not pattern match in
+                // a loop inside display.rs.
+                unreachable!("Trying to draw cursor from cell renderer.")
```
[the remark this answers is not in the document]
kchibisov: I'll try to refactor it so the cursor will be returned somewhere else.

alacritty/src/cursor.rs
chrisduerr: I'm not sure what you mean, why would it need to be mutable?

And the Vecs are created in this function, right? So that would create lifetime problems. Or do you plan to store the vector on the `Cursor`?

alacritty_terminal/src/term/mod.rs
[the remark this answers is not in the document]
chrisduerr: Huh, I'm not sure what I've been thinking. I must have confused myself. You're right of course.

kchibisov:
I've refactored things a bit, however I'm not sure if it's the right way to do that, but it actually works...

alacritty_terminal/src/term/mod.rs
```diff
@@ -124,6 +138,16 @@ impl<'a> RenderableSearch<'a> {
     }
 }
 
+/// Common rendering primitives stored in one structure to simplify passing them around.
+struct RenderableContentInner<'a, T, C> {
```
kchibisov, line 142: I have no clue how to name it, maybe `RenderableContext`.

alacritty_terminal/src/term/mod.rs
```diff
@@ -124,6 +138,16 @@ impl<'a> RenderableSearch<'a> {
...
+    term: &'a Term<T>,
+    selection: Option<SelectionRange<Line>>,
+    search: RenderableSearch<'a>,
+    config: &'a Config<C>,
+    cursor_point: Point,
+    cursor_shape: CursorShape,
```
kchibisov, line 148: We don't pass cursor here, since when getting rendering cursor we need to get cell first before getting color. I also decided to not create like 2 different renderable cursor and decided to stick with one.

alacritty_terminal/src/term/mod.rs
```diff
@@ -938,6 +888,58 @@ impl<T> Term<T> {
         res
     }
 
+    /// Check selection state of a cell.
+    pub fn is_selected(
```
kchibisov, line 892: since it's on `Term` now maybe we can use it somewhere else, but that cursor stuff annoys me a bit. Passing selection also sounds questionable, but I think it's fine, if we hide this function for now...

alacritty_terminal/src/term/mod.rs
```diff
@@ -50,12 +50,26 @@ pub const MIN_SCREEN_LINES: usize = 1;
 
 /// Cursor storing all information relevant for rendering.
 #[derive(Debug, Eq, PartialEq, Copy, Clone, Deserialize)]
-struct RenderableCursor {
+pub struct RenderableCursor {
+    color: Rgb,
```
kchibisov, line 54: Decided to store here just to avoid creating one more struct, but it makes certain things a bit more uglier, I guess.

alacritty/src/cursor.rs
```diff
@@ -1,112 +1,79 @@
...
-    // Create a custom glyph with the rectangle data attached to it.
-    RasterizedGlyph {
-        c: ' ',
-        top: line_width as i32,
-        left: 0,
-        height: line_width as i32,
-        width: width as i32,
-        buf: BitmapBuffer::RGB(buf),
+impl Cursor {
+    fn new(cursor: RenderableCursor, size_info: &SizeInfo, thickness: f64) -> Self {
```
chrisduerr, line 21: I find this API a bit strange, how you create a cursor first and then create a cursor rect from it.

You should know what kind of cursor you have from the beginning, right? So why not just create one immediately?

You also do strange stuff like creating a cursor, passing `size_info` to it, but then getting the `width` out of size info, multiplying it by the width the created cursor gives you, just to pass it back to the cursor. After you passed the `size_info` to the cursor it had all the information it ever needed, why juggle around with the `width` outside of it?

alacritty_terminal/src/term/mod.rs
chrisduerr: `RenderContext` maybe? `RenderableContext` would mean it's a context that can be rendered.

alacritty_terminal/src/term/mod.rs
chrisduerr: Why move it on `Term`, it seems out of place. Passing `selection` to something on `Term` doesn't make a ton of sense, because it already has selection stored on it. So that makes all of this rather questionable.

Also passing a cursor to a function that checks if a cell is selected doesn't really make sense to me for a terminal selection check?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1440,22 +1442,44 @@ impl<T> Term<T> {
...
+            cursor_shape,
+            search: RenderableSearch::new(self),
+        };
+
+        let cell = Indexed {
+            inner: &self.grid[buffer_point.line][buffer_point.col],
+            column: cursor_point.col,
+            line: cursor_point.line,
+        };
+        let mut cell = RenderableCell::new(&mut inner, cell);
```
chrisduerr, line 1470: Do we create an entire `RenderableContentInner` including selection, config, term and search, then create a cell from it, just to check which color our cursor is supposed to be?

It feels like there's got to be an easier way to figure out a cursor's color than that.

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
kchibisov: Because I can't create a renderable cell otherwise? I need selection and searcher to perform that. I'd like to not do so, but then I need to change the logic entirely, though, maybe selection logic doesn't really matter for is_selected. The main issue is that it originates from renderable cell constructor, and i need it to create a cursor...

alacritty_terminal/src/term/mod.rs
kchibisov: Well, all coloring stuff is handles by renderable cell. And we have a lot of fancy logic to get colors, etc, etc. Maybe we can craft something simple, but I'm not sure.

chrisduerr: Yeah, but you don't need stuff like `bg_alpha`, `column/line`, `zerowidth`, `flags` or `match` detection. Sure the color logic is here, but it just seems like there's also a lot of other logic that we don't really care about here?

I think the biggest problem is just that we kinda construct a full renderable cell here, when we know that the renderable cell will already be created in a different place. It would be nice if we could just access the information from that cell instead of having to build it twice.

kchibisov: But the information will be there only when iteration is done, we could pass a mutable reference to a renderable cursor and finish initialization there.

chrisduerr: At that point we could just return a struct from `renderable_cells` which contains cursor and cells.

chrisduerr (approved):

PR #4418, opened 2020-11-14
Load only necessary gl functions

kchibisov:
This commit make 'gl::load_with' only load
symbols alacritty will actually use. Doing so
slightly improves compilation times, since the gl_bindings.rs
file size is reduced from 21K LoC to 2K LoC, and also the actual
loading by up to 20x, though the loading is usually sub millisecond
anyway.

I'm not sure if this one needs a changelog entry, since it's not really noticeable for the user.

### DIFF
--- a/alacritty/build.rs
+++ b/alacritty/build.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 use std::env;
 use std::fs::File;
 use std::path::Path;
@@ -11,7 +12,7 @@ fn main() {
     let dest = env::var("OUT_DIR").unwrap();
     let mut file = File::create(&Path::new(&dest).join("gl_bindings.rs")).unwrap();
 
-    Registry::new(Api::Gl, (4, 5), Profile::Core, Fallbacks::All, ["GL_ARB_blend_func_extended"])
+    Registry::new(Api::Gl, (3, 3), Profile::Core, Fallbacks::All, ["GL_ARB_blend_func_extended"])
         .write_bindings(GlobalGenerator, &mut file)
         .unwrap();
 


alacritty/build.rs
```diff
@@ -11,9 +12,71 @@ fn main() {
...
+        "Uniform4f",
+        "UseProgram",
+        "VertexArrays",
+        "VertexAttribDivisor",
+        "VertexAttribPointer",
+        "Viewport",
+    ]
+    .iter()
+    .cloned()
+    .collect();
```
chrisduerr, line 66: This is absolutely massive and will have to be changed every single time we want to use a new function. We also have no good way to remove things we don't need anymore and code completion will certainly suffer under it.

The reduction in code size seems great, but is it really worth it? How much does it improve compile times, I'd imagine the difference is basically nothing.

chrisduerr (changes requested):

alacritty/build.rs
kchibisov: The size reduction is closer to 100KiB from that. And symbols loading is 20usecs, instead of 400usecs.

Also, we touch our rendererer functions like once a year, so if someone really wants to add more things they can disable filtering.

Also, there's a docs.gl which is way better for OpenGL docs than auto-completion. 

The difference is like 3-4 seconds for debug build on my system (30 instead of 33/34).

chrisduerr: Those arguments aren't sufficient for me to justify the development overhead introduced by this addition. I never want to put us into a situation where certain parts of our code are unmaintainable for marginal compile time improvements of the rest of it.

kchibisov: I mean, it's -150KiB in binary size from just removing dead code. I'm not sure how things are unmaintainable, you mean, if we remove something and won't update this list? It could be the case where we'd need automation...

We may still reduce code size by 70KiB from just using 3.3 profile iirc(it'll be like 11K LoC).

chrisduerr: I don't see how this would be sustainable without automation. Just me reviewing this PR is a great example. The only way for me to properly verify that there are no unnecessary imports is to try and remove every single one and then verify that compilation fails. That's just a massive amount of time that has to be invested just to check for something like this.

If we assume that this reduces the size by 80KiB over just using 3.3 profile, that's only a 1.3% reduction in binary size. Fairly insignificant. Even if we count the whole 150KiB that's only 2.5%. If I can get that size reduction for free, I'm all for it. But taking a hit on maintainability for a 2.5% reduction in binary size? That doesn't seem worth it.

chrisduerr (approved):

PR #4421, opened 2020-11-14
Feature gate 'image' when building without x11 feature

kchibisov:
On Wayland there's no way to embed icon into the window,
thus there's no point in loading it when x11 feature is disabled.

I'd like to include into v0.6.0 to include with the massive feature gating, that was done for it.


chrisduerr (approved):
I'm not 100% sure about pushing a new RC for this, but I think I'm generally in favor.

kchibisov:
I mean not just for this, since maybe we want to include other minor stuff I've send today.

PR #4431, opened 2020-11-16
Add Alpine Linux build prereqs

ChandlerSwift:
This PR adds a list of dependencies to build Alacritty on Alpine Linux.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -42,8 +42,9 @@ cargo install alacritty
         12. [Gentoo](#gentoo)
         13. [Clear Linux](#clear-linux)
         14. [GNU Guix](#gnu-guix)
-        15. [Windows](#windows)
-        16. [Other](#other)
+        15. [Alpine Linux](#alpine-linux)
+        16. [Windows](#windows)
+        17. [Other](#other)
 2. [Building](#building)
     1. [Linux/Windows](#linux--windows)
     2. [macOS](#macos)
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -214,6 +215,16 @@ to be missing, please open an issue.
...
 ```
 
+#### Alpine Linux
+
+On Alpine Linux, you need a few extra libraries to build Alacritty. Here's an
+`apk` command that should install all of them. If something is still found to
+be missing, please open an issue.
+
+```sh
+sudo apk add cmake pkgconf fontconfig-dev python3 libxcb-dev
```
kchibisov, line 225: You need freetype here as well.

INSTALL.md
```diff
@@ -42,8 +42,9 @@ cargo install alacritty
         12. [Gentoo](#gentoo)
         13. [Clear Linux](#clear-linux)
         14. [GNU Guix](#gnu-guix)
-        15. [Windows](#windows)
-        16. [Other](#other)
+        15. [Alpine Linux](#alpine-linux)
```
chrisduerr, line 45: You've put it before GNU Guix in the actual docs, but it's after Guix here. So please change one of them.

chrisduerr (changes requested):

INSTALL.md
ChandlerSwift: I'd missed this because `fontconfig-dev` has `freetype-dev` as a dependency. However, in hindsight, it also has `pkgconf` as a dependency---now we should be consistent with all the dependencies explicitly listed.

INSTALL.md
ChandlerSwift: Fixed!

kchibisov (approved):

chrisduerr (approved):

Issue #3926, opened 2020-07-07
Alacritty icon concept for macOS Big Sur


jdsimcoe:
This is only a rough concept I put together in Figma, but I wanted to explore what the Alacritty icon would/could look like in the macOS Big Sur container style:

![icon-preview](https://user-images.githubusercontent.com/275617/86828988-c2681200-c048-11ea-80eb-a5b2ed9f43df.png)

Definitely not trying to step on any toes here or take this over, but just wanted to explore and idea and toss it out there for consideration. The other option I explored had the red-to-yellow gradient in a smaller container scaled back behind the terminal view (more like the current icon):

![icon-preview](https://user-images.githubusercontent.com/275617/86829704-9305d500-c049-11ea-8534-991ab09e465b.png)

@jwilm would love any thoughts or feedback if/when you have time.

chrisduerr:
Thanks for sharing your work, but I don't think there's any interest in changing Alacritty's icon just because of some macOS redesign.

alexanderniebuhr:
@jdsimcoe could you supply icon files?
Would love to manually change the icon :)

jdsimcoe:
@chrisduerr I understand that... since macOS is permanently changing its icon container shape I wondered what it would look like to make it feel consistent. I get not wanting to change it around.

@alexanderniebuhr here is a link to the Figma file if you want to play with it: https://www.figma.com/file/gN4EZ0QUHWF9JecskKULEB/Alacritty-Big-Sur-icon/duplicate

Also here is a Dropbox link to the `.icns` icon file: https://www.dropbox.com/s/0i4ez0el7paksg3/Alacritty.icns?dl=0.

mambusskruj:
I like this icon representation for Big Sur. Thank you. 

BTW can't understand what is the problem to put slightly different icon for MacOS while building `.dmg`.

kchibisov:
@jdsimcoe do you have an svg for this icon? We may put it in https://github.com/alacritty/alacritty/tree/master/extra/logo so the users may use it if they want.

jdsimcoe:
Here is the SVG: https://gist.github.com/jdsimcoe/8760878212c4001216a738d2111dcb33. Also here is a link to the icon in Figma if this is helpful for `.icns` export: https://www.figma.com/file/CI8Keod1I0OBlslWjMSiVE/Alacritty-Icon-for-macOS-Big-Sur/duplicate.

kchibisov:
[...]

PR #4435, opened 2020-11-16
Create alacritty-term-bigsur.svg

jdsimcoe:
Adding a macOS Big Sur-style Alacritty icon.

`CHANGELOG.md` file.

### DIFF
--- a/extra/logo/alacritty-term-bigsur.svg
+++ b/extra/logo/alacritty-term-bigsur.svg
@@ -1,0 +1,143 @@
+<svg width="1024" height="1024" viewBox="0 0 1024 1024" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M408.787 107.93H615.339C663.15 107.93 696.126 107.967 721.721 110.059C746.757 112.104 760.482 115.873 770.558 121.008C793.135 132.511 811.49 150.866 822.993 173.442C828.127 183.519 831.897 197.244 833.942 222.28C836.033 247.875 836.071 280.851 836.071 328.662V532.338C836.071 580.149 836.033 613.125 833.942 638.72C831.897 663.756 828.127 677.481 822.993 687.558C811.49 710.134 793.135 728.489 770.558 739.992C760.482 745.127 746.757 748.896 721.721 750.941C696.126 753.033 663.15 753.07 615.339 753.07H408.787C360.976 753.07 328 753.033 302.405 750.941C277.369 748.896 263.644 745.127 253.567 739.992C230.991 728.489 212.636 710.134 201.133 687.558C195.998 677.481 192.229 663.756 190.184 638.72C188.092 613.125 188.055 580.149 188.055 532.338V328.662C188.055 280.851 188.092 247.875 190.184 222.28C192.229 197.244 195.998 183.519 201.133 173.442C212.636 150.866 230.991 132.511 253.567 121.008C263.644 115.873 277.369 112.104 302.405 110.059C328 107.967 360.976 107.93 408.787 107.93Z" stroke="url(#paint0_linear)" stroke-width="95.8603"/>
+<g filter="url(#filter0_d)">
+<path d="M100 397.6C100 293.43 100 241.345 120.273 201.558C138.105 166.56 166.56 138.105 201.558 120.273C241.345 100 293.43 100 397.6 100H626.4C730.57 100 782.655 100 822.442 120.273C857.44 138.105 885.895 166.56 903.727 201.558C924 241.345 924 293.43 924 397.6V626.4C924 730.57 924 782.655 903.727 822.442C885.895 857.44 857.44 885.895 822.442 903.727C782.655 924 730.57 924 626.4 924H397.6C293.43 924 241.345 924 201.558 903.727C166.56 885.895 138.105 857.44 120.273 822.442C100 782.655 100 730.57 100 626.4V397.6Z" fill="#14222A"/>
+</g>
+<mask id="mask0" mask-type="alpha" maskUnits="userSpaceOnUse" x="100" y="100" width="824" height="824">
+<path d="M100 397.6C100 293.43 100 241.345 120.273 201.558C138.105 166.56 166.56 138.105 201.558 120.273C241.345 100 293.43 100 397.6 100H626.4C730.57 100 782.655 100 822.442 120.273C857.44 138.105 885.895 166.56 903.727 201.558C924 241.345 924 293.43 924 397.6V626.4C924 730.57 924 782.655 903.727 822.442C885.895 857.44 857.44 885.895 822.442 903.727C782.655 924 730.57 924 626.4 924H397.6C293.43 924 241.345 924 201.558 903.727C166.56 885.895 138.105 857.44 120.273 822.442C100 782.655 100 730.57 100 626.4V397.6Z" fill="#14222A"/>
+</mask>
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
As stated [here](https://github.com/alacritty/alacritty/issues/3926#issuecomment-728295412) I am **very strongly** against adding this.

kchibisov:
I personally don''t see anything wrong with having compat icon for macOS, we already provide compat icons for KDE, because of kde/qt bugs. Also, it's not like we can maintain any of the icon anyway, so they are just laying around. We also had an icon specially for macOS/Windows in the past iirc.

chrisduerr:
> I personally don''t see anything wrong with having compat icon for macOS

There is no compatibility problem though. The only reason why the KDE icon exists is because our main icon does not technically render correctly **at all** with the minimal SVG spec. That means our icon is completely broken with a valid small SVG implementation, of course it makes sense to add a compatibility icon for that.

The reason for adding this icon would solely fall down to user preference, which is not an issue of compatibility at all.

> We also had an icon specially for macOS/Windows in the past iirc.

That is incorrect. Having an existing icon packaged for a platform is completely different from having multiple different icons.

kchibisov:
>There is no compatibility problem though. The only reason why the KDE icon exists is because our main icon does not technically render correctly at all

That's fair. 

I don't  see any problem with proving additional icon for big sur. We already provide different variants of the same icon term and term-scanlines. This one looks like term, but  more rounded. If it was a completely different icon, I'd not be inclined to add it.

mambusskruj:
I can only say here that for me this conversation is not about different icon, like with different elements or colors. It's more about different representation of existed icon. PR icon has same design with same elements except of different border shape. And it's just applies Apple HIG so it sees beautiful within OS UI. So the icon can exist as an option to use. 
P.S. IMO without any tension.

chrisduerr:
> I don't see any problem with proving additional icon for big sur. We already provide different variants of the same icon term and term-scanlines. This one looks like term, but more rounded. If it was a completely different icon, I'd not be inclined to add it.

The only reason why there are icons with and without scanlines is to make it behave correctly at different sizes. None of this is about user preference, 100% of this "Big Sur Icon" is about user preference.

There is zero reason to add this and then deny a GNOME icon, a KDE icon, a Window icon, a Sway icon. It makes no sense at all.

Alacritty provides the official Alacritty icon. If you don't like that icon and would like to use a community made icon instead that's perfectly fine, but that will **not** be shipped with the Alacritty source code. It has **nothing** to do with Alacritty. It's purely an icon built by the community to satisfy the preferences of *some* users.

kchibisov:
>There is zero reason to add this and then deny a GNOME icon, a KDE icon, a Window icon, a Sway icon. It makes no sense at all.

The icons on linux are configured by icon theme, which you set globally, since such things are done externally. Windows doesn't really have a guidelines for icons afaics, and macOS provides them.  We may keep it under contrib folder so it won't get lost, I don't see any issue with that, since contrib folder isn't something that associated with project and just lives on its own.

chrisduerr:
> The icons on linux are configured by icon theme, which you set globally, since such things are done externally.

Just like icon themes can be replaced on Linux, everyone can *easily* replace their icon in their application bundle. That's completely trivial.

> We may keep it under contrib folder so it won't get lost

So what if it gets lost? It's not our official icon. If people like it and share it, it will stay alive. If people don't care, it'll die. That's perfectly fine.

jdsimcoe:
@chrisduerr @kchibisov I'm fine with whatever you decide. Most macOS apps are adopting Big Sur icon container shapes since the design is forward-compatible with the new system and it still looks great on older platforms. iTerm 2 has done this, and I still think having the "official" macOS icon be the Big Sur shape is the best course of action, but that isn't for me to decide.

JesseFarebro:
@chrisduerr What you outline seems to be a non-issue. If you’re going to support a platform why shouldn’t an icon come along with that support? Some platforms impose this to some extent, e.g., iOS. As others have stated this is fundamentally the same icon just with some minor tweaks to the borders.

  What is the added cost of maintaining an icon for GNOME, macOS, Windows if that’s what the community wants and they’ve contributed the icons? Maybe a couple of lines in your CI / build scripts? The added cost to the end-user to change this with every release seems much higher.  

The churn this adds seems to be minimal and in my opinion, Alacritty feeling more native on these platforms is a big win. You may not use macOS but people who do are the kind of people who care about this kind of thing. I think following the human interface guidelines on any platform is a welcomed contribution that leads to a better user experience.

chrisduerr:
> I'm fine with whatever you decide. Most macOS apps are adopting Big Sur icon container shapes since the design is forward-compatible with the new system and it still looks great on older platforms. iTerm 2 has done this, and I still think having the "official" macOS icon be the Big Sur shape is the best course of action, but that isn't for me to decide.

Apple changing their icon theme every few years is no reason for Alacritty to change. Our old icon didn't comply with their standards and there's no reason why a new one would have to.

> What you outline seems to be a non-issue. If you’re going to support a platform why shouldn’t an icon come along with that support? 

If building an icon is necessary for a platform, I'm happy to not support that platform.

>   What is the added cost of maintaining an icon for GNOME, macOS, Windows if that’s what the community wants and they’ve contributed the icons? Maybe a couple of lines in your CI / build scripts? The added cost to the end-user to change this with every release seems much higher.  

There is exactly zero benefit, it's completely unrelated to Alacritty and you're already causing a maintenance overhead of creating issues and sending PRs. Opening the floodgates to that will just make it even worse.

> Alacritty feeling more native on these platforms is a big win

I have zero interest in this. If Apple imposes restrictions that make Alacritty feel out of place, that's the fault of the platform, not Alacritty's. macOS is already extremely bad when it comes to this, I don't care one bit about trying to comply with their misguided UI/UX guidelines.

Closing this PR since I do not see this going anywhere productive.

JesseFarebro:
> Apple changing their icon theme every few years is no reason for Alacritty to change. Our old icon didn't comply with their standards and there's no reason why a new one would have to.
> If building an icon is necessary for a platform, I'm happy to not support that platform.

This is obviously a community maintained project, no one is saying you **must** do this for each platform, the whole point of this PR is that the community contributed an icon adhering to the guidelines for a popular platform and you're unwilling to give it the time of day.

> There is exactly zero benefit, it's completely unrelated to Alacritty and you're already causing a maintenance overhead of creating issues and sending PRs. Opening the floodgates to that will just make it even worse.

That's your opinion, I would argue that adhering to the human interface guidelines has a non-zero benefit and is entirely related to Alacritty. Your floodgate argument is a poor one, the stance could easily be: if there's a clear human interface guideline in place that's actively supported by other apps on said platform then we'll consider a **contribution**.

> I have zero interest in this. If Apple imposes restrictions that make Alacritty feel out of place, that's the fault of the platform, not Alacritty's. macOS is already extremely bad when it comes to this, I don't care one bit about trying to comply with their misguided UI/UX guidelines.
> Closing this PR since I do not see this going anywhere productive.

You seem to have quite the vendetta for macOS. You as many of us volunteer our time to open source projects, we all appreciate it, but at the end of the day perhaps you shouldn't let your personal biases affect the project, frankly, it just comes off as hostile but perhaps that's why you're an effective maintainer.

I don't see anything wrong with this type of contribution and it seems unlikely you can come up with a convincing argument at this point beyond: I don't like macOS, I won't use it, and I don't care.

PR #4441, opened 2020-11-18
Fix vi mode terminal reset

chrisduerr:
Since the vi mode is unrelated to the terminal emulation itself, it
should not be reset during a `reset` to prevent unnecessary confusion.

This also prevents the search from switching from vi mode to vi-less
search without any indication to the user.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - Wide characters sometimes being cut off
+- Preserve vi mode across terminal `reset`
 
 ## 0.6.0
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - Wide characters sometimes being cut off
+- Preserve vi mode across terminal `reset`
```
kchibisov, line 13: Since we use it with capital V in config, we should use capital here as well?

```suggestion
- Preserve Vi mode across terminal `reset`
```

chrisduerr: In the past we don't usually have it capitalized? I think officially vi isn't capitalized either?

PR #4445, opened 2020-11-19
Fix config import log calls

chrisduerr:
This removes the two separate missing config import log calls and
combines them into one. Since each log call acquires a separate log,
having two separate calls can cause problems with multithreading.
Handling everything in just one should prevent any other log from
showing up in between.

### DIFF
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -223,8 +223,7 @@ fn load_imports(config: &Value, config_paths: &mut Vec<PathBuf>, recursion_limit
         };
 
         if !path.exists() {
-            info!(target: LOG_TARGET_CONFIG, "Skipping importing config; not found:");
-            info!(target: LOG_TARGET_CONFIG, "  {:?}", path.display());
+            info!(target: LOG_TARGET_CONFIG, "Config import not found: {:?}", path);
             continue;
         }
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -4,7 +4,7 @@
 #
 # These configuration files will be loaded in order, replacing values in files
 # loaded earlier with those loaded later in the chain. The file itself will
-# always be loaded last. If an import path cannot be found it will be skipped.
+# always be loaded last. If an import path cannot be found, it will be skipped.
 #import:
 #  - /path/to/alacritty.yml
 


alacritty/src/config/mod.rs
```diff
@@ -223,8 +223,7 @@ fn load_imports(config: &Value, config_paths: &mut Vec<PathBuf>, recursion_limit
         };
 
         if !path.exists() {
-            info!(target: LOG_TARGET_CONFIG, "Skipping importing config; not found:");
-            info!(target: LOG_TARGET_CONFIG, "  {:?}", path.display());
+            info!(target: LOG_TARGET_CONFIG, "Config import not found: {:?}", path);
```
kchibisov, line 226: I don't see a reason to change it, if you're talking about the message from created log, it's due to on demand creation of the logger. We have the exact same logging for our config path, so if you want to protect against such things, change it as well.

kchibisov (changes requested):

alacritty/src/config/mod.rs
kchibisov: Or change the way on demand logger works, so the it'll log its path before the first log message.

chrisduerr: The reason it's done this way for the configuration files is just because there are multiple files logged at the same time and it's significantly more complicated to do this in a single statement.

Right here it's no more complicated at all to combine it into a single statement, if anything it's easier. So there's no reason why we should ever allow a different log message to get in between.

kchibisov: Yeah, but then you have overly long path, why I've suggest to make 2 calls, is because the log entry from a missing config file, is sooooo long that it's hard to look at.

Again the split only happens due to logger logic, we may adjust it to log about log file before.

chrisduerr: Also just to be clear, this problem will show up for the user at the moment:

tmp

I'll send a separate PR for improving the position of the log file creation message.

kchibisov: I've already send #4446

kchibisov: Just to be clear, I don't really like overly long lines in a log, since they are hard to read, since our path are very long, hence they are absolute. If you have any idea how to log them without taking 100+ cells in a terminal, I'm all ears.

kchibisov: Also, they can't be split, since the terminal is locked when the processing of the events is done, thus it's single threaded.

chrisduerr: A lock of our terminal has nothing to do with stdout?

kchibisov: This message can be printed only from 2 paths afaics, from during startup and config load(single threaded here), and from config reload, (also kind of single threaded, since we're looking terminal). So I don't see how you can split it from other threads?

chrisduerr: A function should not require any knowledge about where it is called from? Logging with multiple writes to stdout in a multithreaded logger is just bad practice and you should always assume that this will add the possibility to mess up your output.

I see no reason why we wouldn't use a single log call here, these are not two separate log messages. This is a single message, sent at a single point in time, expressing a singular issue.

kchibisov: As I said, it's getting long with a single call.

chrisduerr: It's a log message, such is life. Putting it in a new line also only saves you 26 columns, that's not a massive difference and certainly doesn't justify this.

Calling logging twice for things that should be connected is just unnecessarily fragile and bad style.

kchibisov: But we already have this style in `fn run()`, so you can split those paths as well.

chrisduerr: As stated previously, `fn run()` is completely different. If you have a simple suggestion to rewrite this in a single log call, please do let me know.

kchibisov: What if we introspect the message we're about to log in the logger for having `\n` `\r` inside of it, and generate a proper log entry for each of them and then write everything in a single write_all call? So we can use `\n` in our logging function?

chrisduerr: What do you mean by "proper log entry"? Why should we split log messages up?

kchibisov: I'll send a patch to explain what I mean, since it's simpler.

kchibisov:
I think it's obsolete now?

PR #4447, opened 2020-11-19
Fix stdout log message order

chrisduerr:
This patch makes sure that the message for the creation of a log file is
always the first entry, before any other log file messages.

Since we initialize our log file dynamically, the message is printed as
soon as something is written to it. By making sure that we always write
to a file first and then try stdout, we can ensure that the log file is
always initialized before ever writing log messages to stdout.


kchibisov:
Duplicate of #4446.

PR #4452, opened 2020-11-21
Render underline and strikeout rects in batches

w23:
tl;dr this makes `dense_cells` run 5x faster (from 40-50fps to 270fps on my machine). 

# Why
Current master renderer renders each rect in a separate draw call, which is fine if there's just a few of them. However, if there are *many*, as in e.g. `vtebench -b benchmarks/dense_cells`, it starts to get ugly.

Thankfully, rects conveniently come in bulk, so we might just render them in a single draw call, as this PR suggests.

# How
`RectRenderer` is introduced which encapsulates rect batch building and rendering. It has `draw` method to draw a vector of rects. For each rect in the vector two triangles and six total vertices are produced, each with its own copy of rect color. These vertices are written into `vertices` accumulation vector. When all rects have been processed, vertices data is uploaded to VBO and is rendered with `glDrawArrays`.
It is assumed that there won't be many calls to `RectRenderer::draw`, so it does not batch rects across `draw` calls. Each call will clear `vertices` buffer.
The buffer is not deallocated ever. Not touching allocator gives up to 3% perf improvement, while worst reasonable case will require just a few hundred kb of memory.

# Alternatives
- instancing. While it would certainly be more memory efficient (20 bytes vs 48 bytes per rect; these can be individually improved ~1.5x, though), it moves alacritty further away from being OpenGL ES 2.0-friendly.
- store rect data in texture(s) and sample it in vertex shader. This would make it almost as memory efficient as instancing, with the cost being in code complexity.


alacritty/src/main.rs
```diff
@@ -67,6 +67,9 @@ use crate::event::{Event, EventProxy, Processor};
 use crate::macos::locale;
 use crate::message_bar::MessageBuffer;
 
+#[macro_use]
+extern crate memoffset;
```
chrisduerr, line 71: Macros should be imported explicitly.

alacritty/src/renderer/mod.rs
```diff
@@ -691,37 +672,7 @@ impl QuadRenderer {
...
-            gl::EnableVertexAttribArray(0);
-        }
-
-        // Draw all the rects.
-        for rect in rects {
-            self.render_rect(&rect, props);
-        }
+        self.rect_renderer.draw(&self.rect_program, props, rects);
 
         // Deactivate rectangle program again.
```
chrisduerr, line 677: If the program is activated inside the rect renderer, shouldn't the reset be performed inside it as well?

I'm just talking about stuff like `gl::UseProgram(0)`. Being enabled inside the draw call and enabled outside seems a bit strange, but I can see why you wouldn't want to reset things that the rect renderer technically has no knowledge of.

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+
+use crate::gl;
+use crate::gl::types::*;
+use crate::renderer::rects::RenderRect;
+use crate::renderer::RectShaderProgram;
+
+use std::mem::size_of;
+use std::ptr;
+
+const MAX_U16_INDICES: usize = 65536;
```
chrisduerr, line 11: This seems like a strange constant. Wouldn't you just use `u16::max_value()` instead? Though I don't see how the maximum u16 is relevant here, this is just an arbitrary number of vertices, right?

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    pub fn draw(
+        &mut self,
+        program: &RectShaderProgram,
+        size_info: &SizeInfo,
+        rects: Vec<RenderRect>,
+    ) {
+        if rects.is_empty() {
+            return;
+        }
```
chrisduerr, line 94: I'd assume this is just an optimization? Does it make sense to perform that here? We wouldn't want it to do more harm that good and chances are checking at a higher level might make sense?

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+            self.append_rect(center_x, center_y, rect);
+        }
+
+        self.draw_accumulated();
+    }
+
+    fn append_rect(&mut self, center_x: f32, center_y: f32, rect: &RenderRect) {
+        if rect.alpha <= 0. {
+            return;
+        }
```
chrisduerr, line 124: Just to be clear, this should be completely unreachable.

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        assert!(self.vertices.len() <= MAX_U16_INDICES - 4);
```
chrisduerr, line 126: I'd assume the `- 4` is because each new rect requires 4 new vertices?

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+            b: rect.color.b,
+            a: (rect.alpha * 255.) as u8,
+        };
+
+        self.vertices.extend_from_slice(&[
+            Vertex { x, y, color },
+            Vertex { x, y: y - height, color },
+            Vertex { x: x + width, y, color },
+            Vertex { x: x + width, y: y - height, color },
+        ]);
```
chrisduerr, line 145: This is a bit of a strange way to push four values to a vector. I'd hope that the optimizer cleans this up for you, but it's still a bit weird.

I think a better idea would be to just push four times. I'd assume you might have done it this way to prevent it from reallocating more than once? Could maybe just `self.vertices.reserve(4);` be good enough to solve that problem?

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        if self.vertices.len() == MAX_U16_INDICES {
+            self.draw_accumulated();
+        }
```
chrisduerr, line 149: What's the reason behind limiting this at all? Would it impact performance due to the payload being too big, or is it solely about memory usage?

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    fn draw_accumulated(&mut self) {
+        if self.vertices.is_empty() {
+            return;
+        }
+
+        // Generate new indices in index buffer on-demand
+        assert!(self.indices.len() % 6 == 0);
```
chrisduerr, line 158: Wait, how do `indices` work. You store this globally on the rect renderer and they're initialized once when the renderer is created (which happens once at launch iirc).

Then you push to indices whenever an accumulated draw is performed.

~~Isn't this leaking memory? Where are the indices cleared up? Shouldn't they be re-created every time `draw_accumulated` is called?~~

~~If the goal is to prevent reallocation every time and keep a bit of memory around for the indices, one could call `clear` on the vec before drawing, but I feel like it would probably be fine to just recreate the Vec?~~

**Edit:** You do indeed seem to be calling `clear` at least. So this isn't a memory leak. But I'm still not sure if it isn't much easier to just re-create the vec in `draw_accumulated` and would be curious what the performance impact of that is.

Also as a sidenote: Not a massive fan from shipping assertions in production code. Usually those are just pointless branches that should never be hit. Generally we only have those enabled in debug mode.

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+        for index in generated_quads..need_quads {
+            let index = index * 4;
+            self.indices.extend_from_slice(&[
+                index,
+                index + 1,
+                index + 2,
+                index + 2,
+                index + 3,
+                index + 1,
+            ]);
```
chrisduerr, line 170: Same question as the vertices extension applies here.

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        // Upload accumulated buffers
```
chrisduerr, line 173:
```suggestion
        // Upload accumulated buffers.
```

Generally comments should always end with a period. This applies to all comments in this PR.

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+        unsafe {
+            gl::BufferData(
+                gl::ARRAY_BUFFER,
+                (self.vertices.len() * std::mem::size_of::<Vertex>()) as isize,
+                self.vertices.as_ptr() as *const _,
+                gl::STREAM_DRAW,
+            );
+
+            // If we need more indices than have been already uploaded
+            if self.uploaded_indices < self.indices.len() {
```
chrisduerr, line 183: Didn't we make sure at the beginning of this method that vertices isn't empty, then push the new vertices to the indices? How would uploaded_indices ever be smaller when we *just* increased the size of indices?

The only point for `uploaded_indices` to exist seems to be to support the memory leak that is the constantly growing indices list. Even if one would want to do that there'd be no reason to keep `uploaded_indices` stored on the program though, one could just check when entering this function.

chrisduerr (changes requested):

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+
+            // Color
+            gl::VertexAttribPointer(
+                1,
+                4,
+                gl::UNSIGNED_BYTE,
+                gl::TRUE,
+                (size_of::<Vertex>()) as _,
+                offset_of!(Vertex, color) as *const _,
+            );
```
kchibisov, line 76: I think you can look into how we're handing `VertexAttribPointer` without a `offset_of` in `render/mod.rs`, since it seems to work fine for quite a long time already.

alacritty/Cargo.toml
```diff
@@ -30,6 +30,7 @@ copypasta = { version = "0.7.0", default-features = false }
 libc = "0.2"
 unicode-width = "0.1"
 bitflags = "1"
+memoffset = "0.5"
```
kchibisov, line 33: See my comment about `offset_of!` we're already handling attrib pointers with it, so I think you can try what we're doing.

alacritty/src/renderer/mod.rs
```diff
@@ -410,17 +410,17 @@ struct InstanceData {
...
-    rect_vao: GLuint,
-    rect_vbo: GLuint,
     atlas: Vec<Atlas>,
     current_atlas: usize,
     active_tex: GLuint,
     batch: Batch,
     rx: mpsc::Receiver<Msg>,
+
+    rect_renderer: SolidRectRenderer,
+    rect_program: RectShaderProgram,
```
kchibisov, line 423: I think we can move `rect_program` into `rect_renderer`? I don't see any reason to not have it there. I also question whether we should have something like and split-out quad renderer into a separate file, since `renderer/mod.rs` is getting big tbqh? Can do it separately though.

```rust
struct Renderer {
    quad_renderer: QuadRenderer,
    rect_renderer: RectRenderer,

    rx: mspc::Receiver<Msg>,
}
```

What do you think @chrisduerr ?

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+    b: u8,
+    a: u8,
+}
+
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
+struct Vertex {
+    // TODO these can certainly be i16
+    x: f32,
+    y: f32,
```
kchibisov, line 27: I think macOS doesn't like i16 in general, so I'd go for f32, though, we can test  that it works on macOS.

alacritty/src/renderer/solidrect.rs
```diff
@@ -0,0 +1,200 @@
...
+    color: Rgba,
```
kchibisov, line 28: In general, we tend to inline `Rgba` into such things, so just having `r: u8`, `g: u8`, etc, like we have on instance data to pass the same vec4 seems more reasonable, otherwise we should refactor the way we  store such things on `InstanceData` to use new `Rgba` struct, and not just in `RectRenderer`.

alacritty/src/renderer/solidrect.rs
kchibisov: I'd add a comment either via `-4` is needed or name it properly to reflect that.

kchibisov (changes requested):

kchibisov:
I've tested this branch on my machine, and it's about 6x perf improvement on dense_cells (from 40 to 240-250 fps jump).

alacritty/src/renderer/solidrect.rs
w23: Honestly, I don't know the best way to do it. I'll try to investigate it separately when I have time.

My mental model here is `std::vector` from C++. For vector it is more optimal than pushing items one by one, and definitely faster than reserving each time. Reserving messes with growth strategy: it makes sure that there's an allocation **each time** an element is added, while default growth in push_back will allocate only occasionally by pre-allocating more storage.

alacritty/src/renderer/solidrect.rs
w23: GLES2 spec only allows 16-bit indices, and I still hope to be able to support that.

alacritty/src/renderer/solidrect.rs
w23: I think I should just pre-allocate and pre-upload indices upfront. That would be the simplest way :).

Asserts are intended for debug mode only and are there to provide visibility into invariants expected. I can certainly remove them if you feel that they provide no value.

alacritty/src/renderer/mod.rs
w23: A avoided moving `rect_program` into SolidRectRender because program reloading is coupled with toplevel QuadRenderer. I will try to make it work.

alacritty/src/renderer/solidrect.rs
w23: There's also a problem where rect coordinates come with non-integer pixel coordinates. I looked at it briefly and it has to do with non-integer thickness, did not investigate further.

alacritty/src/renderer/solidrect.rs
kchibisov: There's no `emplace_back` in rust, however I'm not really sure that push is better than what you're doing when it comes to copy members. One can benchmark it, but that's about it. Though, push is more common in such cases afaics.

alacritty/src/renderer/solidrect.rs
kchibisov: >Asserts are intended for debug mode only and are there to provide visibility into invariants expected. I can certainly remove them if you feel that they provide no value.

if you want them for debug, use `debug_assert`.

alacritty/src/renderer/mod.rs
kchibisov: >I avoided moving rect_program into SolidRectRender because program reloading is coupled with toplevel QuadRenderer. I will try to make it work.

I've explicitly checked that it'll work fine, so you just access a field on a rect render, which is basically the same as accessing anything on quad renderer, since rect one is already there.

alacritty/src/renderer/mod.rs
w23: I've moved state strictly related to rects (buffers, program) inside the rect_renderer, and all other reasonably-common state (viewport, blend mode) to outside.

alacritty/Cargo.toml
w23: I have trouble figuring out how to amend that macro to be valid and available for both mod.rs and solidrect.rs

Moreover, I think that doing it the way I do it is better: it more clearly states that this particular field is supposed to be this attribute.

alacritty/src/renderer/solidrect.rs
w23: clippy doesn't like this: `error: 6 bindings with single-character names in scope` :(
```
let x = (rect.x - center_x) / center_x;
let y = -(rect.y - center_y) / center_y;
let width = rect.width / center_x;
let height = rect.height / center_y;
let r = rect.color.r;
let g = rect.color.g;
let b = rect.color.b;
let a = (rect.alpha * 255.) as u8;

// Append rect vertices.
self.vertices.extend_from_slice(&[
    Vertex { x, y: y - height, r, g, b, a },
    Vertex { x: x + width, y, r, g, b, a },
    Vertex { x: x + width, y: y - height, r, g, b, a },
]);
```

alacritty/src/renderer/solidrect.rs
w23: The original code does this :shrug:, but I trust you and have removed it.

alacritty/src/renderer/solidrect.rs
w23: Commented

alacritty/src/renderer/solidrect.rs
w23: I left it as is for now. Can change to individual pushes if you insist. Or if I find time to benchmark it somehow and have data to back either approach.

alacritty/src/renderer/solidrect.rs
w23: Replaced a single remaining assert with debug_assert

alacritty/src/renderer/solidrect.rs
w23: Fixed

alacritty/src/renderer/solidrect.rs
w23: Removed this on-demand uploading entirely.

alacritty/src/renderer/mod.rs
w23: Done

alacritty/src/renderer/solidrect.rs
chrisduerr: To be clear, my suggestion was not focused on performance but on code clarity.

I'm relatively certain that `extend_from_slice` is decent for performance especially with small datatypes. However it does invoke `clone` which **requires** you to be aware of these datatypes and make sure they actually stay small and easy to clone around.

The assumption here is that you **do not** have ownership over the slice, which is why it's not a mutable borrow. So the code doesn't really express our intentions very well, because cloning an existing slice certainly isn't our goal here, but our goal is to push new elements, that were just created, onto that vector.

So unless there's a good reason to do it this way (as in: significant performance benefit), our code should focus on being clear on what it's trying to do and help with maintenance in the future. Currently this code doesn't do that very well.

alacritty/src/renderer/solidrect.rs
chrisduerr: Locally disabling it here should be an option. I don't see a reason for this lint, but I'd have to check it out myself since I've never seen it before.

alacritty/Cargo.toml
kchibisov: Yeah, but it's 3 build deps and memoffset testing their crate in a way we compute address. The thing is that it works quite good only for `repr(C)` structs. The real solution to what we're doing is when rust provides a **reliable** way to take an offset into struct field.

Example of the patch that gets rid of `memoffset`.

```diff
diff --git a/Cargo.lock b/Cargo.lock
index 1064cb2..1eddd10 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,7 +38,6 @@ dependencies = [
  "image",
  "libc",
  "log",
- "memoffset",
  "notify",
  "objc",
  "parking_lot",
@@ -1141,15 +1140,6 @@ dependencies = [
  "winapi 0.3.9",
 ]
 
-[[package]]
-name = "memoffset"
-version = "0.5.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "043175f069eda7b85febe4a74abbaeff828d9f8b448515d3151a14a3542811aa"
-dependencies = [
- "autocfg",
-]
-
 [[package]]
 name = "miniz_oxide"
 version = "0.3.7"
diff --git a/alacritty/Cargo.toml b/alacritty/Cargo.toml
index 9069b93..081ea4d 100644
--- a/alacritty/Cargo.toml
+++ b/alacritty/Cargo.toml
@@ -30,7 +30,6 @@ copypasta = { version = "0.7.0", default-features = false }
 libc = "0.2"
 unicode-width = "0.1"
 bitflags = "1"
-memoffset = "0.5"
 
 [build-dependencies]
 gl_generator = "0.14.0"
diff --git a/alacritty/src/renderer/rectrenderer.rs b/alacritty/src/renderer/rectrenderer.rs
index f2c72af..eb5f9d4 100644
--- a/alacritty/src/renderer/rectrenderer.rs
+++ b/alacritty/src/renderer/rectrenderer.rs
@@ -1,8 +1,6 @@
 use std::mem::size_of;
 use std::ptr;
 
-use memoffset::offset_of;
-
 use alacritty_terminal::term::SizeInfo;
 
 use crate::gl;
@@ -104,27 +102,34 @@ impl RectRenderer {
             // VBO binding is not part ot VAO, but attributes are.
             gl::BindBuffer(gl::ARRAY_BUFFER, vbo);
 
+            let mut size = 0;
+            let mut index = 0;
+
+            macro_rules! add_attr {
+                ($count:expr, $gl_type:expr, $gl_normalized:expr, $type:ty) => {
+                    gl::VertexAttribPointer(
+                        index,
+                        $count,
+                        $gl_type,
+                        $gl_normalized,
+                        size_of::<Vertex>() as i32,
+                        size as *const _,
+                    );
+                    gl::EnableVertexAttribArray(index);
+
+                    #[allow(unused_assignments)]
+                    {
+                        size += $count * size_of::<$type>();
+                        index += 1;
+                    }
+                };
+            }
+
             // Position.
-            gl::VertexAttribPointer(
-                0,
-                2,
-                gl::FLOAT,
-                gl::FALSE,
-                (size_of::<Vertex>()) as _,
-                ptr::null(),
-            );
-            gl::EnableVertexAttribArray(0);
+            add_attr!(2, gl::FLOAT, gl::FALSE, f32);
 
             // Color.
-            gl::VertexAttribPointer(
-                1,
-                4,
-                gl::UNSIGNED_BYTE,
-                gl::TRUE,
-                (size_of::<Vertex>()) as _,
-                offset_of!(Vertex, color) as *const _,
-            );
-            gl::EnableVertexAttribArray(1);
+            add_attr!(4, gl::UNSIGNED_BYTE, gl::TRUE, u8);
 
             // Reset buffer bindings.
             gl::BindVertexArray(0);

```

Can go futher and refactor the macro in quad render to share this macro in rect render.

alacritty/src/renderer/mod.rs
kchibisov: Btw, since we've mentioned splitting QuadRenderer into quad_renderer.rs are you fine with doing so or will do that ourselves?

Also, the file should be `rect_renderer.rs`.

alacritty/src/renderer/solidrect.rs
kchibisov: That sounds strange, but I think possible, it's not like you have a half a pixel on the screen.

alacritty/src/renderer/mod.rs
chrisduerr: Could also just do `renderer/quad.rs` and `renderer/rect.rs`. Of course `rect.rs` already exists, but that would probably be okay since these two things are related.

alacritty/src/renderer/solidrect.rs
chrisduerr: > it's not like you have a half a pixel on the screen.

I mean technically, subpixel positioning is a thing sooo.. :)

alacritty/src/renderer/solidrect.rs
w23: I see what you mean. Will change to pushes.

alacritty/Cargo.toml
w23: Yeah I tried making it common and ran into a wall as I still don't understand how to do macros in rust properly (and it modules `use` model drives me nuts).

alacritty/src/renderer/mod.rs
w23: It's `rects.rs`, so we can have both!

If you feel like it should be a part of this PR, I can try to make it.

alacritty/src/renderer/rectrenderer.rs
```diff
@@ -0,0 +1,258 @@
...
+        self.vertices.extend_from_slice(&[
+            Vertex { x, y, color },
+            Vertex { x, y: y - height, color },
+            Vertex { x: x + width, y, color },
+            Vertex { x: x + width, y: y - height, color },
+        ]);
+    }
+
+    fn draw_accumulated(&mut self) {
+        unsafe {
```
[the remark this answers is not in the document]
w23: That would also put vertices.clear() under unsafe, which i'm not sure i understand implications of

w23:
Made and pushed all the suggestions I agree with.
Remaining things from the comments:
1. whether to up `draw_accumulated()` under `unsafe`.
2. whether to replace `Rgba` with just individual components (and how to silence clippy on that)
3. whether to move instanced glyph renderer into a separate file.

I'm not sure what's the culture here around PR, but I would appreciate if you mark all addressed threads as resolved, as it's getting hard to navigate :D.

alacritty/src/renderer/rect.rs
```diff
@@ -0,0 +1,259 @@
...
+            let mut index = 0;
+            let mut size = 0;
+
+            macro_rules! add_attr {
+                ($count:expr, $gl_type:expr, $type:ty) => {
+                    gl::VertexAttribPointer(
+                        index,
+                        $count,
+                        $gl_type,
+                        gl::FALSE,
```
kchibisov, line 112: You actually need to pass gl::TRUE for normalized in `color` case, since otherwise it's always white or similar.

kchibisov (changes requested):

kchibisov:
>I'm not sure what's the culture here around PR, but I would appreciate if you mark all addressed threads as resolved, as it's getting hard to navigate :D.

Can't you mark them as resolved yourself if you feel like it's resolved?

alacritty/src/renderer/rect.rs
w23: KEKW true

alacritty/src/renderer/rectrenderer.rs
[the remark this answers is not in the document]
kchibisov: In general, even if the entire code inside a function is safe it could still be unsafe, due to some external implications, so unsafe clearly exposes that to the user of the function, thus they'll be more carefully when using it. I'm not entirely sure if we should try to prevent with unsafe here, since the function is internal anyway...

alacritty/src/renderer/solidrect.rs
w23: Yeah, vertex coordinates have subpixel precision even without fancy multisampling stuff. If they hadn't, we'd all be looking at jittery 3D graphics from 90s.

kchibisov:
We may refactor our macros stuff like

<Details>

```diff
diff --git a/alacritty/src/renderer/mod.rs b/alacritty/src/renderer/mod.rs
index db845da..61aba17 100644
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -33,6 +33,38 @@ use crate::gl::types::*;
 use crate::renderer::rect::{RectRenderer, RectShaderProgram};
 use crate::renderer::rects::RenderRect;
 
+#[macro_use]
+macro_rules! post_vertex_attr_enable {
+    ($index:ident, InstanceData) => {
+        gl::EnableVertexAttribArray($index);
+        gl::VertexAttribDivisor($index, 1);
+    };
+    ($index:ident, $stride:tt) => {
+        gl::EnableVertexAttribArray($index);
+    };
+}
+
+#[macro_use]
+macro_rules! add_vertex_attr {
+    ($stride:tt, $index:ident, $size:ident, $count:expr, $gl_type:expr, $norm:expr, $ty:ty) => {
+        gl::VertexAttribPointer(
+            $index,
+            $count,
+            $gl_type,
+            $norm,
+            size_of::<$stride>() as i32,
+            $size as *const _,
+        );
+        post_vertex_attr_enable!($index, $stride);
+
+        #[allow(unused_assignments)]
+        {
+            $size += $count * size_of::<$ty>();
+            $index += 1;
+        }
+    };
+}
+
 mod rect;
 pub mod rects;
 
@@ -558,45 +590,24 @@ impl QuadRenderer {
             let mut index = 0;
             let mut size = 0;
 
-            macro_rules! add_attr {
-                ($count:expr, $gl_type:expr, $type:ty) => {
-                    gl::VertexAttribPointer(
-                        index,
-                        $count,
-                        $gl_type,
-                        gl::FALSE,
-                        size_of::<InstanceData>() as i32,
-                        size as *const _,
-                    );
-                    gl::EnableVertexAttribArray(index);
-                    gl::VertexAttribDivisor(index, 1);
-
-                    #[allow(unused_assignments)]
-                    {
-                        size += $count * size_of::<$type>();
-                        index += 1;
-                    }
-                };
-            }
-
             // Coords.
-            add_attr!(2, gl::UNSIGNED_SHORT, u16);
+            add_vertex_attr!(InstanceData, index, size, 2, gl::UNSIGNED_SHORT, gl::FALSE, u16);
 
             // Glyph offset and size.
-            add_attr!(4, gl::SHORT, i16);
+            add_vertex_attr!(InstanceData, index, size, 4, gl::SHORT, gl::FALSE, i16);
 
             // UV offset.
-            add_attr!(4, gl::FLOAT, f32);
+            add_vertex_attr!(InstanceData, index, size, 4, gl::FLOAT, gl::FALSE, f32);
 
             // Color and cell flags.
             //
             // These are packed together because of an OpenGL driver issue on macOS, which caused a
             // `vec3(u8)` text color and a `u8` cell flags to increase the rendering time by a
             // huge margin.
-            add_attr!(4, gl::UNSIGNED_BYTE, u8);
+            add_vertex_attr!(InstanceData, index, size, 4, gl::UNSIGNED_BYTE, gl::FALSE, u8);
 
             // Background color.
-            add_attr!(4, gl::UNSIGNED_BYTE, u8);
+            add_vertex_attr!(InstanceData, index, size, 4, gl::UNSIGNED_BYTE, gl::FALSE, u8);
 
             // Cleanup.
             gl::BindVertexArray(0);
diff --git a/alacritty/src/renderer/rect.rs b/alacritty/src/renderer/rect.rs
index be4895a..52b04ae 100644
--- a/alacritty/src/renderer/rect.rs
+++ b/alacritty/src/renderer/rect.rs
@@ -103,31 +103,11 @@ impl RectRenderer {
             let mut index = 0;
             let mut size = 0;
 
-            macro_rules! add_attr {
-                ($count:expr, $gl_type:expr, $normalize:expr, $type:ty) => {
-                    gl::VertexAttribPointer(
-                        index,
-                        $count,
-                        $gl_type,
-                        $normalize,
-                        size_of::<Vertex>() as i32,
-                        size as *const _,
-                    );
-                    gl::EnableVertexAttribArray(index);
-
-                    #[allow(unused_assignments)]
-                    {
-                        size += $count * size_of::<$type>();
-                        index += 1;
-                    }
-                };
-            }
-
             // Position.
-            add_attr!(2, gl::FLOAT, gl::FALSE, f32);
+            add_vertex_attr!(Vertex, index, size, 2, gl::FLOAT, gl::FALSE, u16);
 
             // Color.
-            add_attr!(4, gl::UNSIGNED_BYTE, gl::TRUE, u8);
+            add_vertex_attr!(Vertex, index, size, 4, gl::UNSIGNED_BYTE, gl::TRUE, u8);
 
             // Reset buffer bindings.
             gl::BindVertexArray(0);
```

</Details>

But I'm not entirely sure that I like it.

w23:
Please let me know if there's anything else I should do.

chrisduerr (changes requested):

alacritty/src/renderer/rect.rs
```diff
@@ -0,0 +1,259 @@
...
+                    );
+                    gl::EnableVertexAttribArray(index);
+
+                    #[allow(unused_assignments)]
+                    {
+                        size += $count * size_of::<$type>();
+                        index += 1;
+                    }
+                };
+            }
```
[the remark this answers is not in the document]
kchibisov: I've posted a potential macro that we could use, but it's way more complicated than what we have, since macros are not equal.

https://github.com/alacritty/alacritty/pull/4452#issuecomment-731910384

alacritty/src/renderer/rect.rs
```diff
@@ -0,0 +1,259 @@
...
+
+use alacritty_terminal::term::SizeInfo;
+
+use crate::gl;
+use crate::gl::types::*;
+use crate::renderer::rects::RenderRect;
+use crate::renderer::{create_program, create_shader, Error, ShaderCreationError};
+
+/// Maxmimum number of rect vertices per batch. Limited by 16-bit integer indices.
+const MAX_VERTICES: usize = u16::max_value() as usize + 1;
```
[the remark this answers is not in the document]
kchibisov: Can't we basically compute it in a way that it's `mod 4 == 0`, since rust allows constepxr stuff?

alacritty/src/renderer/rect.rs
```diff
@@ -0,0 +1,259 @@
...
+        for rect in &rects {
+            self.append_rect(center_x, center_y, rect);
+            if self.vertices.len() == MAX_VERTICES {
+                self.draw_accumulated();
+            }
+        }
+
+        if !self.vertices.is_empty() {
+            self.draw_accumulated();
+        }
```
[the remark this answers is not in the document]
kchibisov: Use https://doc.rust-lang.org/std/slice/struct.Chunks.html and invoke closure that appends & draws for each chunk.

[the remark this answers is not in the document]
chrisduerr: Yeah, I was thinking something like that. Though taking chunks might make calling `append_rect` a tiny bit more annoying. It would probably still be better than the current version though.

alacritty/src/renderer/mod.rs
```diff
@@ -30,21 +30,19 @@ use crate::config::ui_config::{Delta, UIConfig};
 use crate::cursor;
 use crate::gl;
 use crate::gl::types::*;
+use crate::renderer::rect::{RectRenderer, RectShaderProgram};
 use crate::renderer::rects::RenderRect;
```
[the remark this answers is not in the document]
w23: `rects` feel like a publicly-visible module that is itself "logic only", without any rendering.
I can (1) rename `rect.rs` to `rect_render.rs` or something. Or I can just (2) dump renderer into `rects.rs`.
Or is there a better option?

alacritty/src/renderer/mod.rs
```diff
@@ -670,12 +637,10 @@ impl QuadRenderer {
```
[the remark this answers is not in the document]
w23: Yeah, shader reloading for rects is implemented, but file watching isn't. Honestly, I don't like the watching-reloading shader story here that much (hence `filewatch.rs` in the grid rewrite pr).
Here I can add file monitoring for rect shader for consistency, it should be straightforward. (although it will still mean that all shaders will be reloaded even if only a single one of them changed, but I feel that it isn't really a problem)

alacritty/src/renderer/mod.rs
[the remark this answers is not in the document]
chrisduerr: You can always rearrange things to make more sense logically. We just shouldn't have two modules with nearly identical names. Even `rect.rs` and `rect_render.rs` are so close together that it raises the question why those are not combined.

There's also nothing wrong with having part of a module for public use. But if the two things don't make sense together, it's possible to restructure things. Some stuff might be better in a `display/` module maybe. But we can also delay that until the rest of this PR is figured out. Since this is mostly an organizational issue.

alacritty/src/renderer/mod.rs
```diff
@@ -832,7 +774,7 @@ impl QuadRenderer {
 
         self.active_tex = 0;
         self.program = program;
-        self.rect_program = rect_program;
+        self.rect_renderer.set_program(rect_program);
```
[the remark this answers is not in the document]
w23: It more clearly states the intent that you're not supposed to access it.
But I don't have strong preference and can make it `pub` if you insist.

alacritty/src/renderer/rect.rs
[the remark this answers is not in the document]
w23: +1 here is not for divisibility by four. It's there because max value for u16 is 65535, but max number of supported vertices is 65536 (65536th vertex indexed in base-0 will be 65535).

alacritty/src/renderer/rect.rs
```diff
@@ -0,0 +1,259 @@
...
+    b: u8,
+    a: u8,
+}
+
+/// Vertex to store solid-color rectangle data.
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
+struct Vertex {
+    // Normalized screen coordinates.
+    // TODO these can certainly be i16.
```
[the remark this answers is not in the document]
w23: It requires investigation of why thickness is non-integer sometimes, and what should we do with non-integer coordinates in such case.

alacritty/src/renderer/rect.rs
```diff
@@ -0,0 +1,259 @@
...
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
+struct Rgba {
+    r: u8,
+    g: u8,
+    b: u8,
+    a: u8,
+}
+
+/// Vertex to store solid-color rectangle data.
```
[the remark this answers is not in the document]
w23: Solid color as an opposite to textured glyph quads/rects. Alpha is "transparent" vs "opaque".

Anyway, my English is hardly any good, and I'd appreciate it a lot if you pointed to any other language misuse I happen to produce should you spot some.

alacritty/src/renderer/rect.rs
```diff
@@ -0,0 +1,259 @@
...
+    y: f32,
+
+    // Color.
+    color: Rgba,
+}
+
+/// Structure to store and group together rect-related vertices data and GL objects.
+#[derive(Debug)]
+pub struct RectRenderer {
+    // GL buffer objects. VAO stores attribute and ebo bindings.
```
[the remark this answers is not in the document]
w23: To point out why attributes and EBO bindings happen only once on init. And why we need to bind VBO for vertex data upload.

alacritty/src/renderer/rect.rs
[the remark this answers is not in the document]
w23: This macro is different from the one in `render/mod.rs` in non-trivial ways. @kchibisov suggested a more common macro, but it's even more complex.

The proper way to do this is to use `offsetof`, of course ;).

alacritty/src/renderer/rect.rs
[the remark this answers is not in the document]
w23: Chunks is a nice suggestion, I didn't know it existed. Thanks! I'll try to apply it.

alacritty/src/renderer/rect.rs
```diff
@@ -0,0 +1,259 @@
...
+    vao: GLuint,
+    vbo: GLuint,
+    ebo: GLuint,
+
+    program: RectShaderProgram,
+
+    // Accumulated vertices data. Allocated space reused between frames.
+    vertices: Vec<Vertex>,
```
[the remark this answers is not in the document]
w23: This is getting kind of philosophical :).
Defining a Vec where you need it is only simple if you don't care about externalities, which is introduction of allocator activity, which is rather complex. It may not show up in regular profiling, but sometime somewhere in one frame out of 1000 you'd get a spike.
This is one of the reasons why many mature game engines preallocate pools for everything, have custom container implementations, and have "no allocations in renderloop" policy.
Weeding out allocations from codebase that wasn't allocator-aware previously (or even discovering that you have this death-by-1000-cuts problem) is often not trivial.

Here the worst case mem cost of that Vec is 800Kb (+800Kb in GL buffer), and even if it is unused and memory pressure scenario arises, the OS will happily swap it out. This worst-case wasted space can be lowered by having lower value for MAX_VERTICES.

alacritty/src/renderer/rect.rs
```diff
@@ -0,0 +1,259 @@
...
+        indices.reserve(max_rects * 6);
+        for index in 0..max_rects {
+            let index = index as u16 * 4;
+            indices.push(index);
+            indices.push(index + 1);
+            indices.push(index + 2);
+            indices.push(index + 2);
+            indices.push(index + 3);
+            indices.push(index + 1);
+        }
```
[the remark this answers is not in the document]
w23: Because this method needs index data, the data is constant, and generating it on-demand requires (a) bookkeeping, (b) reuploading this data to GL.

Alternatively, we could get rid of indices completely. Just generate 6 vertices per rect, vertex shader is trivial anyway. This is ~20% less memory efficient, but also allows to render all quads at once regardless of any limits.

There are more complex alternatives which involve baking rect data into textures, using gl_VertexID (non-GLES-friendly), etc.

w23:
An update:
- completely removed indices, EBO and friends. This allows to render all the rects at once, no 64k chunking necessary. No visible perf impact on my machine, even if rendering >70k rects.
- removed vertices buffer caching (profiling this properly requires adding quite a bit of scaffolding).
- comments are updated accordingly, and some other minor tweaks applied.
- moved all rect rendering to `rects.rs`. Now there's this not at all confusing beauty: `use crate::renderer::rects::{RectRenderer, RectShaderProgram, RenderRect};`

w23:
Also, I feel like no-vsync rendering should be a feature. Is there anything else that needs to be done besides this?
```
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -886,11 +886,11 @@ impl<N: Notify + OnResize> Processor<N> {
                 self.submit_display_update(&mut terminal, old_is_searching, display_update_pending);
             }

-            // Skip rendering on Wayland until we get frame event from compositor.
-            #[cfg(not(any(target_os = "macos", windows)))]
-            if !self.display.is_x11 && !self.display.window.should_draw.load(Ordering::Relaxed) {
-                return;
-            }
+            // // Skip rendering on Wayland until we get frame event from compositor.
+            // #[cfg(not(any(target_os = "macos", windows)))]
+            // if !self.display.is_x11 && !self.display.window.should_draw.load(Ordering::Relaxed) {
+            //     return;
+            // }

             if terminal.dirty {
                 terminal.dirty = false;
diff --git a/alacritty/src/window.rs b/alacritty/src/window.rs
index 953fffd..ff8d97a 100644
--- a/alacritty/src/window.rs
+++ b/alacritty/src/window.rs
@@ -117,7 +117,7 @@ fn create_gl_window<E>(

     let windowed_context = ContextBuilder::new()
         .with_srgb(srgb)
-        .with_vsync(vsync)
+        .with_vsync(false)
         .with_hardware_acceleration(None)
         .build_windowed(window, event_loop)?;
```

kchibisov:
>Also, I feel like no-vsync rendering should be a feature. Is there anything else that needs to be done besides this?

I don't think that it should tbqh, but it's all you should do to disable 'vsync' afaics. And there's no vsync on Wayland at all anyway...

w23:
> I don't think that it should tbqh

I mean, it should be close to `live-shader-reload` in terms of how often anyone needs it. And its usefulness is not limited to rendering exclusively.

alacritty/src/renderer/rects.rs
```diff
@@ -190,3 +195,231 @@ impl RenderLines {
...
+
+        // Build rect vertices vector.
+        let mut vertices = RectVertices::new(rects.len());
+        for rect in &rects {
+            vertices.add_rect(center_x, center_y, rect);
+        }
+
+        unsafe {
+            // Upload and render accumulated vertices.
+            vertices.draw();
```
kchibisov, line 317: I've checked the discussion wrt where to put vector with vertices, and in the current implementation without indices I think it makes sense to store it on a Renderer(since Vertices can't draw themself, renderer can draw something with vertices).

So the change would be simple.

- Store `RectVertices` (or just `Vec`) on the `RectRenderer`
- When starting drawing you call `self.vertices.clear()`.
- Iterate over `rects` and add new rects
- Inline what `draw` does on the `RectVertices`

We'd have rectangles for each frame anyway so allocating vector for 1 element will be just strange.

Doing minimum vsync amount of allocations each frame seems wasteful, especially with slow system allocators.

alacritty/src/renderer/rects.rs
```diff
@@ -190,3 +195,231 @@ impl RenderLines {
...
+        // Calculate rectangle position.
+        let x = (rect.x - center_x) / center_x;
+        let y = -(rect.y - center_y) / center_y;
+        let width = rect.width / center_x;
+        let height = rect.height / center_y;
+        let color = Rgba {
+            r: rect.color.r,
+            g: rect.color.g,
+            b: rect.color.b,
+            a: (rect.alpha * 255.) as u8,
```
kchibisov, line 351: Could we pass it as not requiring normalization and then normalize in shader like we do with text rendering? Should be the same as for perf, right? This should simplify vertex attrib pointer creation, since everything won't require normalization.

kchibisov (changes requested):

chrisduerr:
> I mean, it should be close to live-shader-reload in terms of how often anyone needs it. And its usefulness is not limited to rendering exclusively.

Just to be clear on this: My personal take on `live-shader-reload` is that it's a broken mess nobody has ever had any need for. I've personally never needed it and I believe the only purpose if it was to help with development in the past. I'm personally all for nuking that feature since I've never *needed* it in years.

alacritty/src/renderer/rects.rs
w23: This would require multiplying `rect.color.{rgb}` by 255 instead. Normalization is the preferred way.

alacritty/src/renderer/rects.rs
```diff
@@ -190,3 +195,231 @@ impl RenderLines {
...
+                        index += 1;
+                    }
+                };
+            }
+
+            // Position.
+            add_attr!(2, gl::FLOAT, gl::FALSE, f32);
+
+            // Color.
+            add_attr!(4, gl::UNSIGNED_BYTE, gl::TRUE, u8);
```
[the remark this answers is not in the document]
w23: Honestly, I don't see any value in macroing this at all. I can relate to avoiding another dependency, but then just hard-coding attributes offsets and making all the `glVertexAttribPointer` calls manually seems fine. It's how it's usually done.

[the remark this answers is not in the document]
kchibisov: The point is to reduce the number of lines from the call, we can remove macro, I guess. Up to @chrisduerr , since he added it in the past.

alacritty/src/renderer/rects.rs
```diff
@@ -190,3 +195,216 @@ impl RenderLines {
...
+    b: u8,
+    a: u8,
+}
+
+/// Struct that stores vertex 2D coordinates and color for rect rendering.
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
+struct Vertex {
+    // Normalized screen coordinates.
+    // TODO these can certainly be i16.
```
kchibisov, line 219: I think we've discussed and agreed that it can't really be i16 due to subpixel precision, etc? I'm not sure if we should care about it for plain rectangles, but whatever.

alacritty/src/renderer/rects.rs
kchibisov: I mean, it'll be something like that.

```diff
diff --git a/alacritty/res/rect.f.glsl b/alacritty/res/rect.f.glsl
index 945eaf2..3810e7c 100644
--- a/alacritty/res/rect.f.glsl
+++ b/alacritty/res/rect.f.glsl
@@ -6,5 +6,5 @@ out vec4 FragColor;

 void main()
 {
-    FragColor = color;
+    FragColor = color / 255.;
 }
diff --git a/alacritty/src/renderer/rects.rs b/alacritty/src/renderer/rects.rs
index b415ad1..546f9c0 100644
--- a/alacritty/src/renderer/rects.rs
+++ b/alacritty/src/renderer/rects.rs
@@ -284,7 +284,7 @@ impl RectRenderer {
             add_attr!(2, gl::FLOAT, gl::FALSE, f32);

             // Color.
-            add_attr!(4, gl::UNSIGNED_BYTE, gl::TRUE, u8);
+            add_attr!(4, gl::UNSIGNED_BYTE, gl::FALSE, u8);

             // Reset buffer bindings.
             gl::BindVertexArray(0);

```

Though, we decide to not use the macro, we'd be fine with asking for normalization.

alacritty/src/renderer/rects.rs
```diff
@@ -190,3 +195,216 @@ impl RenderLines {
...
+
+        // Build rect vertices vector.
+        self.vertices.clear();
+        for rect in &rects {
+            self.add_rect(center_x, center_y, rect);
+        }
+
+        unsafe {
+            // Upload and render accumulated vertices.
+            // Upload accumulated vertices.
```
kchibisov, line 319: Seems like a rebase issue or something similar, or a bad positioning of two comments. Removing the top one will be fine.

alacritty/src/renderer/rects.rs
```diff
@@ -190,3 +195,216 @@ impl RenderLines {
...
+    // Normalized screen coordinates.
+    // TODO these can certainly be i16.
+    x: f32,
+    y: f32,
+
+    // Color.
+    color: Rgba,
+}
+
+/// Struct to group together rect-related GL objects and rendering functionality.
```
kchibisov, line 227: I find this comments starting with `struct` a bit weird to read tbqh, it's not really a C, since we have `impl` blocks for the sturcts, etc. I think it'd be nice to not mention `struct` at all, since it's already marked as such in docs. This applies to other struct you have here (Can go without comment for it tbqh).

kchibisov (changes requested):
Besides some minor comments, that's look good.

alacritty/src/renderer/rects.rs
w23: Removed comments

alacritty/src/renderer/rects.rs
w23: I think the no-macro version is much more readable than trying to hide all this stuff.

alacritty/src/renderer/rects.rs
w23: I'd say making it i16 is possible, as all this subpixel stuff should not affect anything that much here. But it would also be helpful to also revise non-rendering portion of it also. Let's do this separately.

alacritty/src/renderer/rects.rs
w23: I can't see this 2-3% difference (too much noise here). But I moved allocation back to the parent RectRenderer just in case.

alacritty/src/renderer/rects.rs
```diff
@@ -190,3 +195,205 @@ impl RenderLines {
...
+            gl::EnableVertexAttribArray(0);
+
+            // Color.
+            gl::VertexAttribPointer(
+                1,                          // Attribute location.
+                4,                          // Component count: 4. r, g, b, a.
+                gl::UNSIGNED_BYTE,          // Attribute type: u8.
+                gl::TRUE,                   // Normalized: 0..255 is mapped to 0..1.
+                size_of::<Vertex>() as i32, // Stride.
+                8 as *const _,              // Offset of Vertex::color.
```
kchibisov, line 275: Could you apply the logic with `size` we have inside a macro in quad renderer? So it's a bit more clear what it's going on than `8`?

kchibisov (changes requested):
Could you also add an entry into CHANGELOG into `Fixed` section wording similar to `Slow rendering performance with a lot of cells with underline/strikeout attributes`

alacritty/src/renderer/rects.rs
```diff
@@ -190,3 +195,207 @@ impl RenderLines {
...
+            // Allocate buffers.
+            gl::GenVertexArrays(1, &mut vao);
+            gl::GenBuffers(1, &mut vbo);
+
+            gl::BindVertexArray(vao);
+
+            // VBO binding is not part of VAO itself, but VBO binding is stored in attributes.
+            gl::BindBuffer(gl::ARRAY_BUFFER, vbo);
+
+            let attribute_offset = 0;
```
kchibisov, line 257: Can't we make it `mut` and then do `+=` instead of shadowing?

kchibisov (changes requested):

kchibisov (approved):

w23:
@chrisduerr does it look good to you, or should I also remove `live-shader-reload` completely in this change? I agree with your argument that it's a broken mess, and I haven't actually used it in a meaningful way while experimenting with the renderer.

kchibisov:
>@chrisduerr does it look good to you, or should I also remove live-shader-reload completely in this change? I agree with your argument that it's a broken mess, and I haven't actually used it in a meaningful way while experimenting with the renderer.

We may remove it separately tbqh.

chrisduerr:
I'll review the whole thing again, I haven't looked at it since my last review.

I think unless there's a concrete benefit of removing it in this PR, it makes sense to separate changes.

alacritty/src/renderer/rects.rs
```diff
@@ -190,3 +195,208 @@ impl RenderLines {
...
+static RECT_SHADER_V: &str = include_str!("../../res/rect.v.glsl");
+
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
+struct Rgba {
+    r: u8,
+    g: u8,
+    b: u8,
+    a: u8,
+}
```
chrisduerr, line 212: Would it make sense to share this between the normal and the rect renderer? I know @kchibisov has mentioned something like this before, but it would be nice if things behaved uniformly.

It's also probably something that is useful in more places than just right here. Alternatively using `Rgb` from `alacritty_terminal` together with alpha would be another option.

alacritty/src/renderer/rects.rs
```diff
@@ -190,3 +195,208 @@ impl RenderLines {
...
+            let mut attribute_offset = 0;
+
+            // Position.
+            gl::VertexAttribPointer(
+                0,                            // Attribute location.
+                2,                            // Component count: 2. x and y.
+                gl::FLOAT,                    // Attribute type: f32
+                gl::FALSE,                    // Not normalized.
+                size_of::<Vertex>() as i32,   // Stride.
+                attribute_offset as *const _, // Offset of Vertex::x is zero.
```
chrisduerr, line 266: I find the inline documentation here a bit strange. What is the purpose of this?

The `glVertexAttribPointer` function should have the different inputs documented already, correct? So if we need to document anything, it would only be things that are relevant to our individual usecase. Like x/y, r/g/b/a for example.

alacritty/src/renderer/rects.rs
```diff
@@ -190,3 +195,208 @@ impl RenderLines {
...
+        self.vertices.clear();
+        for rect in &rects {
+            self.add_rect(center_x, center_y, rect);
+        }
+
+        unsafe {
+            // Upload accumulated vertices.
+            gl::BufferData(
+                gl::ARRAY_BUFFER,
+                (self.vertices.len() * std::mem::size_of::<Vertex>()) as isize,
```
chrisduerr, line 314: In renderer/mod.rs we import `std::mem::size_of` completely, usually I'd discourage that in Alacritty for consistency. But just an import of `std::mem` at the top should help shorten things a little bit.

alacritty/src/renderer/rects.rs
```diff
@@ -190,3 +195,208 @@ impl RenderLines {
...
+            gl::BindVertexArray(0);
+        }
+    }
+
+    fn add_rect(&mut self, center_x: f32, center_y: f32, rect: &RenderRect) {
+        // Calculate rectangle position.
+        let x = (rect.x - center_x) / center_x;
+        let y = -(rect.y - center_y) / center_y;
+        let width = rect.width / center_x;
+        let height = rect.height / center_y;
```
chrisduerr, line 336: As someone unfamiliar with OpenGL the purpose of these calculations isn't immediately obvious to me. Based on some rough guesses I'd assume you're constructing a normalized rectangle ranging with values from -1 to +1?

If that is the case, I think it would make sense to just state that nice and shortly, if it isn't please let me know what it does.

My suggestion would be something like this:

```
    // Calculate the normalized rectangle positions. These values range from -1 to +1.
```

chrisduerr (changes requested):
Things are looking pretty good, thanks for your work.

Just some minor tweaks and it should be good to go.

alacritty/src/renderer/rects.rs
kchibisov: Just to clarify, this logic is straight from the old code. And it was without docs even there, so that's why I haven't mentioned during review.

alacritty/src/renderer/rects.rs
kchibisov: I also found them a bit strange, though, it's pretty clear what everything is doing. But at this point you can consult docs.gl, if it matters.

alacritty/src/renderer/rects.rs
kchibisov: I think the problem is that you want a `repr(C)`, however alacritty_terminal structure isn't `repr(C)`, thus you can't really say that it'll work the way you want. Maybe rust will make it `repr(C)` if you start nesting, idk.

alacritty/src/renderer/rects.rs
w23: I got the impression that one of the reasons behind various incarnations of macros which were discussed here at length was these gl calls being hard to follow. The comments are here so that anyone reading this code would be immediately aware of arguments meaning without consulting docs or anything.

If you still feel that they should not be there, I can remove them.

alacritty/src/renderer/rects.rs
chrisduerr: I mean we're free to add whatever attributes we want. It's not like `repr(C)` on a struct with 4 elements is going to result in any weird inefficient behavior.

Though generally I'd say reusing `Rgb` from `alacritty_terminal` is probably the worst part of the suggestion, since we obviously need more than just `Rgb`.

alacritty/src/renderer/rects.rs
chrisduerr: > I got the impression that one of the reasons behind various incarnations of macros which were discussed here at length was these gl calls being hard to follow.

This is not the case, instead the reason is to make sure that ordering of functions is guranteed to be correct and work right. Having just a ton of separate functions one after each other makes it harder to grasp which things are connected and need to work together.

> The comments are here so that anyone reading this code would be immediately aware of arguments meaning without consulting docs or anything.

The docs are exactly what this kind of stuff is there for, ideally the function in Rust would make that immediately obvious but unfortunately through FFI that is often lost. But it's not our job here to document OpenGL functions.

> If you still feel that they should not be there, I can remove them.

I think other than "magic numbers" like passing `2`, they shouldn't be there. Extracting them into variables might help with that. But documenting just what function parameters do on the OpenGL side doesn't make much sense.

alacritty/src/renderer/rects.rs
chrisduerr: Oh, I missed that. The point still stands though so I think with my suggestion that would be a trivial improvement.

alacritty/src/renderer/rects.rs
w23: It's not clear to me what should I do here exactly.

This `Rgba` thing is maybe not great in greater context, but is locally-optimal, given clippy complaints and such (IMHO).

chrisduerr: It's mostly about consistency. Having it as fields in one struct and `Rgba` in another doesn't make much sense.

We should decide to do either one **or** the other. And once we've made that decision, I don't care what clippy says. We can ignore that warning if it doesn't make sense.

I personally don't care much about which one we choose, usually structs are nicer to work with but if raw types are easier for OpenGL interop then that's perfectly fine with me too.

alacritty/src/renderer/rects.rs
```diff
@@ -328,26 +322,25 @@ impl RectRenderer {
...
         let height = rect.height / half_height;
-        let color = Rgba {
-            r: rect.color.r,
-            g: rect.color.g,
-            b: rect.color.b,
-            a: (rect.alpha * 255.) as u8,
-        };
+        let r = rect.color.r;
+        let g = rect.color.g;
+        let b = rect.color.b;
```
chrisduerr, line 335:
```suggestion
        let Rgb { r, g, b } = rect.color;
```

chrisduerr (changes requested):

alacritty/src/renderer/rects.rs
w23: whoa, i did not know this is possible. thanks!

chrisduerr (approved):

PR #4455, opened 2020-11-21
Print multiline logs in one write call

kchibisov:
Fixes potential split of multiline logs in the multithreaded context
by printing them in one write call.

Example of how it looks:

```
Created log file at "/tmp/Alacritty-23566.log"
[2020-11-22 00:57:51.056463821] [INFO ] [alacritty] Welcome to Alacritty
[2020-11-22 00:57:51.056571680] [INFO ] [alacritty] Configuration files loaded from:
[2020-11-22 00:57:51.056571680] [INFO ]               "/home/kchibisov/alacritty.yml"
[2020-11-22 00:57:51.056571680] [INFO ]               "/home/kchibisov/.config/alacritty/alacritty.yml"
[2020-11-22 00:57:51.056571680] [INFO ]               "/dev/null"
[2020-11-22 00:57:51.142588963] [INFO ] [alacritty] Device pixel ratio: 1
[2020-11-22 00:57:51.145890991] [INFO ] [alacritty] Initializing glyph cache...
[2020-11-22 00:57:51.176849642] [INFO ] [alacritty] ... finished initializing glyph cache in 0.030898353s
[2020-11-22 00:57:51.176949109] [INFO ] [alacritty] Cell size: 8 x 18
[2020-11-22 00:57:51.176981554] [INFO ] [alacritty] Padding: 4 x 3
[2020-11-22 00:57:51.177008465] [INFO ] [alacritty] Width: 800, Height: 600
[2020-11-22 00:57:51.178652058] [INFO ] [alacritty] PTY dimensions: Line(33) x Column(99)
[2020-11-22 00:57:51.180553903] [INFO ] [alacritty] Initialisation complete
[2020-11-22 00:57:51.197011645] [INFO ] [alacritty] Padding: 6 x 8
[2020-11-22 00:57:51.197082528] [INFO ] [alacritty] Width: 1268, Height: 1402
```

### DIFF
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -228,8 +228,7 @@ fn load_imports(config: &Value, config_paths: &mut Vec<PathBuf>, recursion_limit
         }
 
         if !path.exists() {
-            info!(target: LOG_TARGET_CONFIG, "Skipping importing config; not found:");
-            info!(target: LOG_TARGET_CONFIG, "  {:?}", path.display());
+            info!(target: LOG_TARGET_CONFIG, "Config import not found:\n  {:?}", path.display());
             continue;
         }
 
--- a/alacritty/src/logging.rs
+++ b/alacritty/src/logging.rs
@@ -108,7 +108,29 @@ impl log::Log for Logger {
         }
 
         let now = time::strftime("%F %T.%f", &time::now()).unwrap();
-        let msg = format!("[{}] [{:<5}] [{}] {}\n", now, record.level(), target, record.args());
+
+        let lines = record.args().to_string();
+        let mut lines = lines.split('\n').peekable();
+
+        let mut msg = format!("[{}] [{:<5}] [{}] ", now, record.level(), target);
+
+        // Alignment for the lines after the first new line character in the payload.
+        let alignment = msg.chars().count();
+
+        // The first line must be on the same line as `target` in the `msg`.
+        if let Some(line) = lines.next() {
+            msg.push_str(&line);
+        }
+
+        // Push the rest of the lines prefixed with `alignment` to make the multiline output look
+        // uniform.
+        for line in lines {
+            let line = format!("\n{:width$}{}", "", line, width = alignment);
+            msg.push_str(&line);
+        }
+
+        // Push the trailing new line character, so the log will reach the output.
+        msg.push('\n');
 
         if let Ok(mut logfile) = self.logfile.lock() {
             // Write to logfile.
--- a/alacritty/src/logging.rs
+++ b/alacritty/src/logging.rs
@@ -110,12 +110,9 @@ impl log::Log for Logger {
         // Create log message for the given `record` and `target`.
         let msg = create_log_message(record, &target);
 
-        // Trim extra spaces if there're any.
-        let msg = msg.trim_end_matches(' ').as_ref();
-
         if let Ok(mut logfile) = self.logfile.lock() {
             // Write to logfile.
-            let _ = logfile.write_all(msg);
+            let _ = logfile.write_all(msg.as_ref());
 
             // Write to message bar.
             if record.level() <= Level::Warn {
@@ -125,7 +122,7 @@ impl log::Log for Logger {
 
         // Write to stdout.
         if let Ok(mut stdout) = self.stdout.lock() {
-            let _ = stdout.write_all(msg);
+            let _ = stdout.write_all(msg.as_ref());
         }
     }
 
@@ -140,13 +137,15 @@ fn create_log_message(record: &log::Record<'_>, target: &str) -> String {
     // with fullwidth/unicode chars here, so just `msg.len()` will work fine.
     let alignment = msg.len();
 
-    // Push lines with adding extra padding into the next line, we'll trim it later on.
+    // Push lines with adding extra padding on the next line, which is trimmed later.
     let lines = record.args().to_string();
     for line in lines.split('\n') {
         let line = format!("{}\n{:width$}", line, "", width = alignment);
         msg.push_str(&line);
     }
 
+    // Drop extra trailing alignment.
+    msg.truncate(msg.len() - alignment);
     msg
 }
 
--- a/alacritty/src/logging.rs
+++ b/alacritty/src/logging.rs
@@ -72,7 +72,7 @@ impl Logger {
         #[cfg(windows)]
         let env_var = format!("%{}%", ALACRITTY_LOG_ENV);
 
-        let msg = format!(
+        let message = format!(
             "[{}] See log at {} ({}):\n{}",
             record.level(),
             logfile_path,
@@ -85,7 +85,7 @@ impl Logger {
             _ => unreachable!(),
         };
 
-        let mut message = Message::new(msg, message_type);
+        let mut message = Message::new(message, message_type);
         message.set_target(record.target().to_owned());
 
         let _ = event_proxy.send_event(Event::Message(message));
@@ -108,11 +108,11 @@ impl log::Log for Logger {
         }
 
         // Create log message for the given `record` and `target`.
-        let msg = create_log_message(record, &target);
+        let message = create_log_message(record, &target);
 
         if let Ok(mut logfile) = self.logfile.lock() {
             // Write to logfile.
-            let _ = logfile.write_all(msg.as_ref());
+            let _ = logfile.write_all(message.as_ref());
 
             // Write to message bar.
             if record.level() <= Level::Warn {
@@ -122,7 +122,7 @@ impl log::Log for Logger {
 
         // Write to stdout.
         if let Ok(mut stdout) = self.stdout.lock() {
-            let _ = stdout.write_all(msg.as_ref());
+            let _ = stdout.write_all(message.as_ref());
         }
     }
 
@@ -131,22 +131,22 @@ impl log::Log for Logger {
 
 fn create_log_message(record: &log::Record<'_>, target: &str) -> String {
     let now = time::strftime("%F %T.%f", &time::now()).unwrap();
-    let mut msg = format!("[{}] [{:<5}] [{}] ", now, record.level(), target);
+    let mut message = format!("[{}] [{:<5}] [{}] ", now, record.level(), target);
 
     // Alignment for the lines after the first new line character in the payload. We don't deal
-    // with fullwidth/unicode chars here, so just `msg.len()` will work fine.
-    let alignment = msg.len();
+    // with fullwidth/unicode chars here, so just `message.len()` is sufficient.
+    let alignment = message.len();
 
-    // Push lines with adding extra padding on the next line, which is trimmed later.
+    // Push lines with added extra padding on the next line, which is trimmed later.
     let lines = record.args().to_string();
     for line in lines.split('\n') {
         let line = format!("{}\n{:width$}", line, "", width = alignment);
-        msg.push_str(&line);
+        message.push_str(&line);
     }
 
     // Drop extra trailing alignment.
-    msg.truncate(msg.len() - alignment);
-    msg
+    message.truncate(message.len() - alignment);
+    message
 }
 
 struct OnDemandLogFile {
--- a/alacritty/src/logging.rs
+++ b/alacritty/src/logging.rs
@@ -107,9 +107,11 @@ impl log::Log for Logger {
             return;
         }
 
-        let now = time::strftime("%F %T.%f", &time::now()).unwrap();
-        let mut msg = format!("[{}] [{:<5}] [{}] ", now, record.level(), target);
-        let msg = append_log_message(&mut msg, record);
+        // Create log message for the given `record` and `target`.
+        let msg = create_log_message(record, &target);
+
+        // Trim extra spaces if there're any.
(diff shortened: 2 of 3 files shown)


alacritty/src/config/mod.rs
```diff
@@ -223,8 +223,11 @@ fn load_imports(config: &Value, config_paths: &mut Vec<PathBuf>, recursion_limit
         };
 
         if !path.exists() {
-            info!(target: LOG_TARGET_CONFIG, "Skipping importing config; not found:");
-            info!(target: LOG_TARGET_CONFIG, "  {:?}", path.display());
+            info!(
+                target: LOG_TARGET_CONFIG,
+                "Skipping importing config; not found:\n  {:?}",
```
chrisduerr, line 228:
```suggestion
                "Config import not found:\n  {:?}",
```

`Skipping importing` just sounds strange. I think it's better to change this up a bit.

Though I personally still find it very strange that a single, self-contained import statement is logged on two lines. That just doesn't make a ton of sense to me.

alacritty/src/main.rs
```diff
@@ -130,10 +130,16 @@ fn run(
...
-    }
+    info!(
+        "{}",
+        &config.ui_config.config_paths.iter().fold(
+            String::from("Configuration files loaded from:"),
+            |mut msg, path| {
+                msg.push_str(&format!("\n  {:?}", path.display()));
+                msg
+            },
+        )
```
chrisduerr, line 141: This is getting a bit messy. It should at least be extracted into a separate statement.

alacritty/src/logging.rs
```diff
@@ -108,7 +108,22 @@ impl log::Log for Logger {
         }
 
         let now = time::strftime("%F %T.%f", &time::now()).unwrap();
-        let msg = format!("[{}] [{:<5}] [{}] {}\n", now, record.level(), target, record.args());
+
+        let msg_prefix = format!("[{}] [{:<5}]", now, record.level());
+        let lines = record.args().to_string();
+        let mut lines = lines.split('\n');
+        let msg = format!("{} [{}] {}\n", msg_prefix, target, lines.next().unwrap_or(""));
```
chrisduerr, line 115:
```suggestion
        let msg = format!("{} [{}] {}\n", msg_prefix, target, lines.next().unwrap_or_default());
```

alacritty/src/logging.rs
```diff
@@ -108,7 +108,22 @@ impl log::Log for Logger {
...
+        let msg = lines.fold(msg, |mut msg, line| {
+            msg.push_str(&format!(
+                "{} {:width$} {}\n",
+                msg_prefix,
+                "",
+                line,
+                width = target.len() + 2,
+            ));
+            msg
+        });
```
chrisduerr, line 126: What's the point of folding here? you got the message already, you're just mutating it anyways? Just use a loop?

chrisduerr (changes requested):
I **really** don't like the way this is formatted with pretending every line is an independent log message. That just makes it unnecessarily complicated to quickly see which lines actually contain new log messages and which lines are just the last log message but extended.

As far as I can tell, repeating information pointlessly just makes my life more difficult here.

kchibisov:
>I really don't like the way this is formatted with pretending every line is an independent log message. That just makes it unnecessarily complicated to quickly see which lines actually contain new log messages and which lines are just the last log message but extended.

I mean it's pretty clear that they are from the same log, since I don't duplicate the log target, thus the message without the log target is connected to the one that had it. I don't like it without any prefix at all, since it seems like it came from some other source other than our logger. Maybe at least keep a time?

chrisduerr:
> I mean it's pretty clear that they are from the same log, since I don't duplicate the log target, thus the message without the log target is connected to the one that had it.

That's hardly a very clear indication, when one out of 3 metadata fields half way through the line is missing. In fact that's quite the opposite of "pretty clear".

kchibisov:
Do we want to align it with just spaces then?

chrisduerr:
> Do we want to align it with just spaces then?

I'd personally prefer it significantly.

If @nixpulvis is around, maybe he could weigh in? Would be good to have a third opinion I suppose.

alacritty/src/logging.rs
```diff
@@ -108,7 +108,29 @@ impl log::Log for Logger {
         }
 
         let now = time::strftime("%F %T.%f", &time::now()).unwrap();
-        let msg = format!("[{}] [{:<5}] [{}] {}\n", now, record.level(), target, record.args());
+
+        let lines = record.args().to_string();
+        let mut lines = lines.split('\n').peekable();
```
chrisduerr, line 113: Why peekable? You don't peek anywhere, right?

alacritty/src/logging.rs
```diff
@@ -108,7 +108,29 @@ impl log::Log for Logger {
...
+
+        let mut msg = format!("[{}] [{:<5}] [{}] ", now, record.level(), target);
+
+        // Alignment for the lines after the first new line character in the payload.
+        let alignment = msg.chars().count();
```
chrisduerr, line 118: Just to be clear, this should be fine since we don't have to deal with fullwidth/unicode here. In fact I think `msg.len()` should probably work too and be a little easier? We're only dealing with ascii here. Technically `msg.chars().count()` is slightly more robust, but if we want to be able to handle non-unicode stuff we'd need to consider fullwidth characters?

alacritty/src/logging.rs
```diff
@@ -108,7 +108,29 @@ impl log::Log for Logger {
...
+        if let Some(line) = lines.next() {
+            msg.push_str(&line);
+        }
+
+        // Push the rest of the lines prefixed with `alignment` to make the multiline output look
+        // uniform.
+        for line in lines {
+            let line = format!("\n{:width$}{}", "", line, width = alignment);
+            msg.push_str(&line);
+        }
```
chrisduerr, line 130: We could in theory also push the padding after each line and then run `trim_end` I suppose? Not sure if that's any cleaner, but it probably would be slightly shorter. By using `trim_end_matches(' ')` we would even skip the `msg.push('\n')`.

alacritty/src/logging.rs
```diff
@@ -108,7 +108,29 @@ impl log::Log for Logger {
         }
 
         let now = time::strftime("%F %T.%f", &time::now()).unwrap();
-        let msg = format!("[{}] [{:<5}] [{}] {}\n", now, record.level(), target, record.args());
+
```
chrisduerr, line 111: The stuff in logging.rs is getting kinda long, do you think it would be worth it to move the creation of the log text into a separate function? There shouldn't be too many arguments (only record?) and the output is a clean `String`.

alacritty/src/main.rs
```diff
@@ -130,9 +130,14 @@ fn run(
 ) -> Result<(), Box<dyn Error>> {
     info!("Welcome to Alacritty");
 
-    info!("Configuration files loaded from:");
-    for path in &config.ui_config.config_paths {
-        info!("  \"{}\"", path.display());
+    // Log the configuration paths.
+    {
```
chrisduerr, line 134: What is this scope for?

chrisduerr (changes requested):

alacritty/src/main.rs
kchibisov: To drop things right away and don't keep them for the entire left time of the program.

chrisduerr: Should we maybe move this into a method then? If we need separation, might as well do it completely?

alacritty/src/logging.rs
```diff
@@ -122,13 +123,29 @@ impl log::Log for Logger {
...
         if let Ok(mut stdout) = self.stdout.lock() {
-            let _ = stdout.write_all(msg.as_ref());
+            let _ = stdout.write_all(msg);
         }
     }
 
     fn flush(&self) {}
 }
 
+fn append_log_message<'a>(msg: &'a mut String, record: &log::Record<'_>) -> &'a [u8] {
```
chrisduerr, line 133: Why isn't the time and format moved into this function? Wouldn't it be easier to just return a String here?

chrisduerr (changes requested):

alacritty/src/logging.rs
```diff
@@ -107,12 +107,15 @@ impl log::Log for Logger {
             return;
         }
 
-        let now = time::strftime("%F %T.%f", &time::now()).unwrap();
-        let msg = format!("[{}] [{:<5}] [{}] {}\n", now, record.level(), target, record.args());
+        // Create log message for the given `record` and `target`.
+        let msg = create_log_message(record, &target);
+
+        // Trim extra spaces if there're any.
+        let msg = msg.trim_end_matches(' ').as_ref();
```
chrisduerr, line 114: Why is this not part of `create_log_message`? Because it returns a reference? Could probably just convert it to owned or something?

alacritty/src/logging.rs
```diff
@@ -122,13 +125,31 @@ impl log::Log for Logger {
...
 
+fn create_log_message(record: &log::Record<'_>, target: &str) -> String {
+    let now = time::strftime("%F %T.%f", &time::now()).unwrap();
+    let mut msg = format!("[{}] [{:<5}] [{}] ", now, record.level(), target);
+
+    // Alignment for the lines after the first new line character in the payload. We don't deal
+    // with fullwidth/unicode chars here, so just `msg.len()` will work fine.
+    let alignment = msg.len();
+
+    // Push lines with adding extra padding into the next line, we'll trim it later on.
```
chrisduerr, line 143:
```suggestion
    // Push lines with extra padding on the next line, which is trimmed later.
```

Commenting this on the trimming function might also be an option.

chrisduerr (changes requested):

alacritty/src/logging.rs
kchibisov: Yeah, but it's one more allocation for nothing, so I've decided to do that separately.

chrisduerr: I mean if that's the only concern, maybe use `rposition` and `truncate`?

kchibisov: I think just `truncate` should be fine?

alacritty/src/logging.rs
```diff
@@ -107,8 +107,8 @@ impl log::Log for Logger {
             return;
         }
 
-        let now = time::strftime("%F %T.%f", &time::now()).unwrap();
-        let msg = format!("[{}] [{:<5}] [{}] {}\n", now, record.level(), target, record.args());
```
chrisduerr, line 111: Could rename this to `message`, but I don't really care.

alacritty/src/logging.rs
```diff
@@ -129,6 +129,26 @@ impl log::Log for Logger {
     fn flush(&self) {}
 }
 
+fn create_log_message(record: &log::Record<'_>, target: &str) -> String {
+    let now = time::strftime("%F %T.%f", &time::now()).unwrap();
+    let mut msg = format!("[{}] [{:<5}] [{}] ", now, record.level(), target);
+
+    // Alignment for the lines after the first new line character in the payload. We don't deal
+    // with fullwidth/unicode chars here, so just `msg.len()` will work fine.
```
chrisduerr, line 137:
```suggestion
    // with fullwidth/unicode chars here, so just `msg.len()` is sufficient.
```

I don't like the word "fine", it implies that it's not great. But we've already determined that this should actually be the best solution here, so I'd rather make that clear.

alacritty/src/logging.rs
```diff
@@ -129,6 +129,26 @@ impl log::Log for Logger {
...
+    let alignment = msg.len();
+
+    // Push lines with adding extra padding on the next line, which is trimmed later.
```
chrisduerr, line 140:
```suggestion
    // Push lines with extra padding added on the next line, which is trimmed later.
```

alacritty/src/logging.rs
```diff
@@ -129,6 +129,26 @@ impl log::Log for Logger {
...
+    let lines = record.args().to_string();
+    for line in lines.split('\n') {
+        let line = format!("{}\n{:width$}", line, "", width = alignment);
+        msg.push_str(&line);
+    }
+
+    // Drop extra trailing alignment.
+    msg.truncate(msg.len() - alignment);
```
chrisduerr, line 148: Wouldn't this still include `\n`? Since alignment is just the whitespace?

chrisduerr (changes requested):

alacritty/src/logging.rs
kchibisov: it's a lot of whitespaces though.

chrisduerr: Unless I'm missing an off by one somewhere, it seems to me like it should though: 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=149234e85155df110bd356d7a7fa2e44

chrisduerr: Oh, just double-checked the old code.

Messages are expected to end in a newline, so it works out. Or at the very least it doesn't seem to matter.

kchibisov: They must end with a new line, otherwise they won't be printed at all or something like that. The only goal is to remove trailing whitespaces here, though, you could use `rposition` to find `\n`, but doesn't really matter...

chrisduerr (approved):

PR #4459, opened 2020-11-22
Fix scrolling region performance with fixed lines at top and bottom

chrisduerr:


chrisduerr:
Performance results:

tmp

alacritty_terminal/src/grid/mod.rs
```diff
@@ -263,46 +287,57 @@ impl<T: GridCell + Default + PartialEq + Clone> Grid<T> {
...
-        } else {
-            // Subregion rotation.
-            for line in IndexRange(region.start..(region.end - positions)) {
-                self.raw.swap_lines(line, line + positions);
-            }
+        // TODO: Technically the rotation here will rotate dirty fixed lines at the top of the
+        // viewport into history. Should these be cleared? This is only a problem when rotating
+        // more than a single line.
+        //  -> I say it doesn't matter, since you shouldn't use a scrolling region like that
+        //  outside of the alt screen anyways.
```
chrisduerr, line 327: @kchibisov This comment talks about the same thing we've discussed on IRC. Generally I'd say that the performance is significantly better for the purpose of this PR, while not touching the performance in other areas much.

Do you think there's some other approach that should be evaluated? If so, what would you suggest? I've tried to create a single thing that fixes everything, but if we wanted to implement multiple different variations we'd probably need at least three. One optimized for primary screen scrolling, one for alt screen with the fixed lines and then one fallback that can handle everything but is slow.

kchibisov (changes requested):

kchibisov (approved):

PR #4475, opened 2020-11-25
Fix CI upload assets tag matching

chrisduerr:
The previous version would search for the last tag by matching the
beginning of the tag name. By explicitly searching for the trailing `"`
with grep, an exact tag match is now enforced.

Since releases like v1.2.3 always match the beginning of their RCs
(v1.2.3-rc4), this makes sure that the assets aren't pushed to the
previous release.


chrisduerr:
For the 0.6.0 release this shouldn't be necessary, since I've decided to just draft a release myself. That way it should hopefully pick the latest one correctly.

kchibisov (approved):

PR #4476, opened 2020-11-25
Draw selection below Vi cursor when it's hidden

kchibisov:


alacritty_terminal/src/term/mod.rs
```diff
@@ -1400,7 +1398,7 @@ impl<T> Term<T> {
     }
 
     /// Get rendering information about the active cursor.
-    fn renderable_cursor<C>(&self, config: &Config<C>) -> RenderableCursor {
+    fn renderable_cursor<C>(&self, config: &Config<C>, show_cursor: bool) -> RenderableCursor {
```
chrisduerr, line 1401: Hiding it here rather than the iterator actually makes a lot more sense anyways.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1413,10 +1411,13 @@ impl<T> Term<T> {
...
-        let cursor_shape = if hidden && !vi_mode {
-            point.line = Line(0);
+        let hidden = !show_cursor
+            || !self.mode.contains(TermMode::SHOW_CURSOR)
+            || point.line >= self.screen_lines();
+        let cursor_shape = if hidden {
+            if !vi_mode {
+                point.line = Line(0);
+            }
             CursorShape::Hidden
```
chrisduerr, line 1421: This isn't correct though, is it? If `SHOW_CURSOR` isn't set, the vi mode cursor must not be hidden. That's at least one of the reason for the `!vi_mode` check and it looks like you break that here?

chrisduerr (changes requested):

chrisduerr (approved):
LGTM.

rudyscat:
Lo creo

PR #4486, opened 2020-11-26
Fix DoS caused by excessive CSI parameter values

chrisduerr:


alacritty_terminal/src/ansi.rs
```diff
@@ -31,9 +31,13 @@ fn parse_rgb_color(color: &[u8]) -> Option<Rgb> {
 
     // Scale values instead of filling with `0`s.
     let scale = |input: &str| {
-        let max = u32::pow(16, input.len() as u32) - 1;
-        let value = u32::from_str_radix(input, 16).ok()?;
-        Some((255 * value / max) as u8)
+        if input.len() > 4 {
```
kchibisov, line 34: Could you explain why this check appeared?

chrisduerr: Oh yeah, sure. This causes an overflow in debug mode. If your rgba parameter is excessively long (8 bytes or more), then the `u32::pow(16, len)` will overflow. So limiting it to the maximum XParseColor value (4) will prevent that from happening.

In practice this isn't a crash, but obviously we don't want some random overflow from messing up the color here, it's better to just ignore it at that point.

kchibisov (approved):

PR #4492, opened 2020-11-27
Add "PrefersNonDefaultGPU" desktop entry key

dvergeylen:
A new Desktop entry key has been added recently (≃ May 2020) to ask for an application to run on a more powerful GPU, if possible.
See: https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html.

A per the doc:
"If true, the application prefers to be run on a more powerful discrete GPU if available, which we describe as “a GPU other than the default one” in this spec to avoid the need to define what a discrete GPU is and in which cases it might be considered more powerful than the default GPU. This key is only a hint and support might not be present depending on the implementation."

This avoids the necessity to add `DRI_PRIME=1` in the Exec command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - New `cursor.style.blinking` option to set the default blinking state
 - New `cursor.blink_interval` option to configure the blinking frequency
 - Support for cursor blinking escapes (`CSI ? 12 h`, `CSI ? 12 l` and `CSI Ps SP q`)
+- `PrefersNonDefaultGPU` entry to Linux .desktop file
 
 ### Changed
 
--- a/extra/linux/Alacritty.desktop
+++ b/extra/linux/Alacritty.desktop
@@ -5,6 +5,7 @@ Exec=alacritty
 Icon=Alacritty
 Terminal=false
 Categories=System;TerminalEmulator;
+PrefersNonDefaultGPU=true
 
 Name=Alacritty
 GenericName=Terminal


kchibisov:
I don't understand why it should be done that way and it's not something that is working in general. Running from integrated GPU is fine and I don't see much difference on my systems when it comes to rendering perf.

Also, this setting assume that you're using some environment that handles such things, so will be useless most of the time.

The alternative solution to this is https://github.com/alacritty/alacritty/pull/4133 , which is more reliable than using this setting in desktop file...

dvergeylen:
Thanks for the feedback @kchibisov 🙂

> I don't understand why it should be done that way and it's not something that is working in general. Running from integrated GPU is fine and I don't see much difference on my systems when it comes to rendering perf.

I didn't neither when I was under X11 + nividia proprietary drivers. But I recently migrated to Gnome + Wayland + open source drivers and not assigning this makes a noticeable difference. The latency is very high and CPU fan becomes crazy.

> 
> Also, this setting assume that you're using some environment that handles such things, so will be useless most of the time.

It will indeed become useful where this is implemented. As stated in the key entry documentation "This key is only a hint and support might not be present depending on the implementation."

> 
> The alternative solution to this is #4133 , which is more reliable than using this setting in desktop file...

Oh, I wasn't aware of this! Yes this might be more relevant / reliable. Feel free to close then. 🤷

chrisduerr:
> I didn't neither when I was under X11 + nividia proprietary drivers. But I recently migrated to Gnome + Wayland + open source drivers and not assigning this makes a noticeable difference. The latency is very high and CPU fan becomes crazy (integrated GPU).

> Oh, I wasn't aware of this! Yes this might be more relevant / reliable. Feel free to close then. shrug

Could you actually test #4133 and confirm that it is working? I've written it without any device that is capable of graphics switching and confirmation that this does actually do something on Linux is the last thing it is waiting on.

nogamike76 (commented):
What are we trying to use this for?

chrisduerr:
Closing since there has been no follow-up. I'd still appreciate a test of #4133, but that should solve all problems.

Issue #4199, opened 2020-09-01
Feature: ResetTerminal action


mitar:
I think it would be useful to have a ResetTerminal action, so that I could tie it to a keyboard shortcut to reinitialize the terminal. This could be useful when I break it so much that typing `reset` it not viable, or if I am not inside a shell but some other program (although I am not sure if it would work there).

It is not something I use often in other terminal emulators, but sometimes I do and sometimes it fixes the issue (and sometimes I just restart the terminal emulator app).

nixpulvis:
I expect this will have the same issue as #4152.

Again, as @kchibisov has mentioned in the other thread, there's really nothing stopping you from finding the `pts/<N>` of the terminal that needs reset and sending it `echo -e "\ec"`.

However, an easier way to get this functionality would still be nice in my eyes. I understand I'm arguing for something we've explictly stated we don't want, but given that the ability to push arbitrary escapes through `/dev/pts` exists, and terminal applications are not all going to be horribly broken by this, I feel that if anything would be allowed, perhaps it would be a reset action. The use case for this is clear, something gets messed up, and the user cannot currently send escapes any other way (killing a running program is not always an option).

chrisduerr:
This is indeed the exact same thing as #4152 and #3997.

In general, people that know what this binding actually does and when it should be used, will have no problem doing the same thing without it. The only thing that could be brought up is making it simpler, but I really doubt you find yourself in a locked up terminal without being able to type `reset` that much. In general typing `reset` is available and I'm not aware of the last time it wasn't for me.

> The use case for this is clear, something gets messed up, and the user cannot currently send escapes any other way (killing a running program is not always an option).

Do you have any examples of where "something gets messed up" with a running program the user cannot interrupt? I'd struggle even intentionally putting myself into that scenario.

> Rather than implement this action as a one off, why not allow sending arbitrary escapes

That has already been discussed before and the arguments really haven't changed. I'm not aware of any way this could be used which wouldn't be abuse of the feature.

kchibisov:
>I think it would be useful to have a ResetTerminal action, so that I could tie it to a keyboard shortcut to reinitialize the terminal. This could be useful when I break it so much that typing reset it not viable, or if I am not inside a shell but some other program (although I am not sure if it would work there).

[...]

PR #4494, opened 2020-11-27
Add 'Reset' action to clear scrollback buffer and reset the terminal

berkeleyinc:
The missing ability to reset the screen while another program is running inside the terminal was bugging me too.

See https://github.com/alacritty/alacritty/issues/3997
and https://github.com/alacritty/alacritty/issues/4199

This adds the Action "Reset".

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -142,6 +142,9 @@ pub enum Action {
     /// Scroll all the way to the bottom.
     ScrollToBottom,
 
+    /// Clears scrollback buffer and resets the terminal
+    Reset,
+
     /// Clear the display buffer(s) to remove history.
     ClearHistory,
 
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -308,6 +308,7 @@ impl<T: EventListener> Execute<T> for Action {
             },
             Action::ClearHistory => ctx.terminal_mut().clear_screen(ClearMode::Saved),
             Action::ClearLogNotice => ctx.pop_message(),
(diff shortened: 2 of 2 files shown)


chrisduerr:
See #3997, #4152 and #4199.

berkeleyinc:
My reason for which I need it:
I'm developing an application and the  program outputs a lot of data.
Using the scrollback buffer, I'm grepping for specific strings in the output.
In the scrollback buffer, there is also the output of the previous run and I don't want that to be shown in my vi-mode searches.
Of course, I could manually "reset", but sometimes I'm deciding whether I need the previous output case by case. 
Having to add or remove a "reset" statement is a lot of hassle and disturbs the programming flow.
This is just one example of the added comfort, the Reset action would bring.
I've recently came back from termite to alacritty but this issue keeps driving me away.

That's why I'd kindly ask to take it into discussion again. The other github issues show that it's a needed feature and many other major terminals have this included as well.

chrisduerr:
I don't see how a reset action would be necessary for this. Again this seems to indicate that people using these actions are not familiar with what the action does.

berkeleyinc:
Is there another way to clear the scrollback buffer?

chrisduerr:
Yeah, the `ClearHistory` action.

berkeleyinc:
Well, that's exactly what I looked for. 
I tried the option before but I guess I was used to the clearing of the screen.

So it's all fine then.
Thank you for the quick answer!

s-valent:
> Yeah, the `ClearHistory` action.

It does not work in all cases, because not all terminal programs (like docker, for example) handle ^L

PR #4517, opened 2020-12-04
Add Apple Silicon support

samdoiron:
Updates alacritty to be compatible with the latest version
of winit, via the parallel pull in glutin:

https://github.com/rust-windowing/glutin/pull/1341

Of course, the dependency changes won't go in as-is. I'm
including them here in case anyone else wants to build
on an M1 in the meantime.

`CHANGELOG.md` file.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,3 +8,6 @@ members = [
 lto = true
 debug = 1
 incremental = false
+
+[patch.crates-io]
+glutin = { git = "https://github.com/samdoiron/glutin" }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -8,3 +8,6 @@ members = [
 lto = true
 debug = 1
 incremental = false
+
+[patch.crates-io]
+glutin = { git = "https://github.com/samdoiron/glutin" }
```
samdoiron, line 13: Obviously temporary. Will switch to the released version once/if [this PR is merged](https://github.com/rust-windowing/glutin/pull/1341)

chrisduerr (changes requested):
Please do not update dependencies unless a new version is actually released. That's not conductive and just wastes a lot of time for reviewers.

PR #4522, opened 2020-12-06
Non-VI mode selections

fxdave:
# I added several new actions to non-vi mode:

- **SelectAddNextLine**
       Extend selection with the line above
- **SelectAddPreviousLine**
       Extend selection with the line below
- **SelectNextLine**
       Select line below the selection or below the cursor when there is no selection
- **SelectPreviousLine**
       Select line above the selection or below the cursor when there is no selection

The main problem that this PR solves: copying the example code from git push. 
image

Yes, I can do this currently in VI mode, but it is easier to use my mouse, which I wanted to avoid.

**Any ideas about the default shortcuts?**
I wanted Ctrl+Alt+{Up, Down}, Ctrl+Alt+Shift+{Up, Down} but that didn't work, unfortunately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - New `cursor.style.blinking` option to set the default blinking state
 - New `cursor.blink_interval` option to configure the blinking frequency
 - Support for cursor blinking escapes (`CSI ? 12 h`, `CSI ? 12 l` and `CSI Ps SP q`)
+- SelectNextLine, SelectPreviousLine, SelectAddPreviousLine, SelectAddNextLine actions.
 
 ### Changed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -545,6 +545,14 @@
 #       Spawn a new instance of Alacritty.
 #   - ClearLogNotice
 #       Clear Alacritty's UI warning and error notice.
+#   - SelectAddNextLine,
+#       Extend selection with the line above
+#   - SelectAddPreviousLine,
+#       Extend selection with the line below
+#   - SelectNextLine,
+#       Select line below the selection or below the cursor when there is no selection
+#   - SelectPreviousLine,
+#       Select line above the selection or below the cursor when there is no selection
 #   - ClearSelection
 #       Remove the active selection.
 #   - ReceiveChar
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -167,6 +167,18 @@ pub enum Action {
     #[cfg(target_os = "macos")]
     ToggleSimpleFullscreen,
 
+    // Extend selection with the line above
+    SelectAddNextLine,
+
+    // Extend selection with the line below
+    SelectAddPreviousLine,
+
+    // Select line below the selection or below the cursor when there is no selection
+    SelectNextLine,
+
+    // Select line above the selection or below the cursor when there is no selection
+    SelectPreviousLine,
+
     /// Clear active selection.
     ClearSelection,
 
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -132,6 +132,87 @@ impl Action {
             selection.include_all();
         }
     }
+
+    fn select_next_line<T, A>(ctx: &mut A, add: bool) where T: EventListener, A: ActionContext<T> {
+        let next_line_number = {
+            let screen_lines = ctx.size_info().screen_lines();
+            let max_line = if screen_lines > Line(1) { screen_lines - Line(1) } else { screen_lines };
+            let actual_line_number = Self::get_actual_selection_or_cursor_line_number(ctx,false);
+            if actual_line_number < max_line {
+                actual_line_number + Line(1)
+            } else {
+                actual_line_number
+            }
+        };
+        if add {
+            ctx.update_selection(
+                Point::new( next_line_number, Column::from(0)), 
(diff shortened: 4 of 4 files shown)


chrisduerr (changes requested):
> Yes, I can do this currently in VI mode, but it is easier to use my mouse, which I wanted to avoid.

You say you don't want to use vi mode, but use the mouse instead. But then you add new bindings to avoid using the mouse and use the keyboard instead. That doesn't quite compute for me?

If you are using the mouse anyways, keep going with the mouse. If you're using the keyboard, why not use vi mode? I fail to see the motivation behind this.

Adding new bindings just because it's simple isn't really reason enough to do it.

fxdave:
I want to use my keyboard; that's why I made this because this reduces the required steps hugely so that I don't need to use my mouse.
And not everybody wants VI shortcuts.
Personally, for e.g., I would use CTRL+S better than `:` and `s` and `<ENTER>`

chrisduerr:
>  And not everybody wants VI shortcuts.

So the problem's just with the vi keyboard mode having vi bindings?

> I would use CTRL+S better than : and s and <ENTER>

Binding something to ctrl+s is probably not the best idea. If anything that should probably be ctrl+shift+s.

fxdave:
> Binding something to ctrl+s is probably not the best idea. If anything that should probably be ctrl+shift+s.

It works on all of my apps.

Anyway, let's be numerical:

VI mode:
- ToggleViMode
- Up
- Up
- ToggleLineSelection
- Copy
- ToggleViMode
- Paste

Mine:
- SelectPreviousLine
- SelectPreviousLine
- Copy
- Paste

In my case, it's less step, fewer shortcuts to remember.

fxdave:
As I see there are two types of keyboard-oriented UIs: 
VI and Common.

If this will be the terminal emulator of the future because this can be, I think It shouldn't force people to use it like VI. Because those people who use VSCode, like me, the mouse would be faster and easier to use, but still painful.

I also switched from NeoVIM to Micro for editing in the terminal.

chrisduerr:
> In my case, it's less step, fewer shortcuts to remember.

You're free to combine multiple vi shortcuts if you'd like to speed things up.

> If this will be the terminal emulator of the future because this can be, I think It shouldn't force people to use it like VI. Because those people who use VSCode, like me, the mouse would be faster and easier to use, but still painful.

Alacritty is **not** going to support two different ways to do the same thing. If you don't like the bindings, you're free to remap them. But the modal editing isn't going anywhere.

Instead of trying to introduce some other features that do not integrate with the existing code at all, I suggest trying to work on making what exists already easier to use. I have no problem with improving vi mode if you have good suggestions on how to improve it, but I'm not going to introduce a separate way to handle keyboard interactions just because some people get scared when they read the word `vi`.

fxdave:
I partly understand your decision.

> You're free to combine multiple vi shortcuts if you'd like to speed things up.

Currently, it seems like key bindings only accept one Action. Or did you write this because I'm allowed to imrove this, or I don't know a feature?

kchibisov:
No, you can bind a lot of actions to the same keys. e.g.

```yaml
  - { key: E,       mods: Control,       action: ScrollLineDown,    mode: Vi,  }
  - { key: E,       mods: Control,       action: ScrollLineDown,    mode: Vi,  }
  - { key: E,       mods: Control,       action: ScrollLineDown,    mode: Vi,  }
  - { key: E,       mods: Control,       action: ScrollLineDown,    mode: Vi,  }
  - { key: E,       mods: Control,       action: ScrollLineDown,    mode: Vi,  }
```

That scroll five lines.

fxdave:
I tried:
```yml
  # Select previous line:
  - { key: U, mods: Shift, mode: ~Vi, action: ToggleViMode}
  - { key: U, mods: Shift, mode: Vi, action: Up}
  - { key: U, mods: Shift, mode: Vi, action: ToggleLineSelection}
  - { key: U, mods: Shift, mode: Vi, action: ToggleViMode}
```

However, As I see, ToggleViMode clears the selection, and there are other limitations too.

It also looks like Vi mode shouldn't be implemented in the code, but the bindings should be stateful instead. For example:
```yml
key_bindings:
    - mode: ~SomeCustomModeName,
      actions:
           - { key: U, mods: Shift, action: EnterMode(SomeCustomModeName)}
    - mode: SomeCustomModeName,
      actions:
           - { key: U, mods: Shift, action: LeaveMode(SomeCustomModeName)}
           - { key: Space,  mods: Shift|Control, action: ScrollToBottom }
           - ...
```
This also makes limitations that can be refined, but we could make it up to the user to configure Vi if he likes it (with examples).

And having default keybindings instead of relying on the configuration seems like a bad design to me. This also can be improved by generating a default configuration file when there isn't any.

chrisduerr:
> However, As I see, ToggleViMode clears the selection, and there are other limitations too.

If you have any suggestions that would allow the selection to stay around, I'm open for ideas. But currently this is not done because there's a conflict between the vi cursor and selection. It is not clear where the vi cursor should appear when the selection stays around.

>  It also looks like Vi mode shouldn't be implemented in the code, but the bindings should be stateful instead. For example:

There's no way to not define the vi mode in code. Entering some arbitrary mode doesn't get you anywhere unless that mode is actually backed by code that does something for you.

> And having default keybindings instead of relying on the configuration seems like a bad design to me. 

**Nobody** wants to run with a configuration file that doesn't have any default bindings, believe me. There's a lot of stuff that doesn't work with just raw keyboard characters.

> This also can be improved by generating a default configuration file when there isn't any.

That's not an improvement. Generating files on the user's system when there is no need to is just bad.

PR #4525, opened 2020-12-06
support paste/paste_selection to search regex.

brianjcj:
`CHANGELOG.md` file.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -458,6 +458,19 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     }
 
     #[inline]
+    fn push_string_search(&mut self, s: &str) {
+        if let Some(regex) = self.search_state.regex.as_mut() {
+            if !self.terminal.mode().contains(TermMode::VI) {
+                // Clear selection so we do not obstruct any matches.
+                self.terminal.selection = None;
+            }
+
+            regex.push_str(s);
+            self.update_search();
+        }
+    }
+
+    #[inline]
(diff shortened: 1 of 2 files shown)


chrisduerr (changes requested):
While this is something planned to be supported in Alacritty, I don't think the way it is implemented here is beneficial to the maintenance of vi mode.

My goal for implementing this would be to have proper bindings for the vi mode with all currently hardcoded options being set from the configuration file. Once that's the case, we can get rid of all the existing bindings and clean this up.

Realistically this will probably mean we'll have to introduce a new `mode` to the config, which represents the active search. However this mode obviously can't be added to `TermMode`, which makes things a bit complicated.

If you're willing to look into this @brianjcj, please do let me know. Otherwise I probably would have implemented this before the next release anyways.

brianjcj:
> While this is something planned to be supported in Alacritty, I don't think the way it is implemented here is beneficial to the maintenance of vi mode.
> 
> My goal for implementing this would be to have proper bindings for the vi mode with all currently hardcoded options being set from the configuration file. Once that's the case, we can get rid of all the existing bindings and clean this up.
> 
> Realistically this will probably mean we'll have to introduce a new `mode` to the config, which represents the active search. However this mode obviously can't be added to `TermMode`, which makes things a bit complicated.
> 
> If you're willing to look into this @brianjcj, please do let me know. Otherwise I probably would have implemented this before the next release anyways.

Looking forward to your implementation! I just can't wait for it and made a quick hack for it :-)

seperman:
Hello, any updates on your implementation @chrisduerr ?

PR #4536, opened 2020-12-08
Man page: add path to Linux example config file

marcospb19:
Refer to `/usr/share/doc/alacritty/example/alacritty.yml` in man page.

### DIFF
--- a/extra/alacritty.man
+++ b/extra/alacritty.man
@@ -51,6 +51,7 @@ Alacritty looks for the configuration file at the following paths:
     4. $HOME/.alacritty.yml
 
 On Windows, the configuration file is located at %APPDATA%\\alacritty\\alacritty.yml.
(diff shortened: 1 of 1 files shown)


marcospb19:
Thanks for the great software!

I assumed an guaranteed `/usr/share/doc/alacritty/example/alacritty.yml` path for every `Linux` installation of `alacritty`, is this correct?

marcospb19:
Looks like CI failed for unrelated reasons, but i'll try to retrigger anyways.

chrisduerr (changes requested):
> I assumed an guaranteed /usr/share/doc/alacritty/example/alacritty.yml path for every Linux installation of alacritty, is this correct?

Thank you for the contribution, but this is not correct. Shipping the default configuration file is completely up to the distribution and not something we have control over at all. So unfortunately we can't provide a filesystem link to it.

PR #4538, opened 2020-12-08
Remove alacritty.spec

chrisduerr:
Since Alacritty is now officially packaged by Fedora, it is no longer
necessary to provide this in the official repository.


pschyska:
@chrisduerr I think you can remove `.copr/Makefile` as well. Cheers

chrisduerr:
@pschyska Thanks, I didn't see that!

kchibisov:
Should we add a note in packaging section that we've removed copr stuff?

pschyska:
> Should we add a note in packaging section that we've removed copr stuff?

It wouldn't hurt, I guess? I've spoken to a number of copr users that didn't know it's now upstream.

chrisduerr:
> Should we add a note in packaging section that we've removed copr stuff?

And who would that be for? The maintainer of the copr already knows about it. The packaging section is *for packagers*, not to inform users about packaging.

kchibisov:
I was thinking about the ones who might used our copr stuff for their own copr jobs or something like that, besides @pschyska . But I think it's fine without mentioning it.

Issue #4518, opened 2020-12-05
Unable to update to alacritty-0.6.0-1 in Fedora 33


oswcab:
OS: Fedora Linux 33
Version: alacritty 0.5.0
Linux/BSD: X11/Gnome

 Package         Architecture             Version           Repository           Size
Upgrading:
 alacritty          x86_64                     0.6.0-1.fc33    copr:copr.fedorainfracloud.org:pschyska:alacritty                                2.0 M

Transaction Summary
Upgrade  1 Package

Total download size: 2.0 M
Is this ok [y/N]: y
Downloading Packages:
alacritty-0.6.0-1.fc33.x86_64.rpm                                                                                                                                                    2.0 MB     00:06
Running transaction check
Transaction check succeeded.
Running transaction test
The downloaded packages were saved in cache until the next successful transaction.
You can remove cached packages by executing 'dnf clean packages'.
**Error: Transaction test error:
  file /usr/share/terminfo/a/alacritty from install of alacritty-0.6.0-1.fc33.x86_64 conflicts with file from package ncurses-base-6.2-3.20200222.fc33.noarch**

chrisduerr:
Please take up maintenance issues with the package maintainers.

mikeslattery:
@chrisduerr, I respectfully disagree that this should have been closed.  The files to create the copr package are maintained here, in the alacritty/alacritty project.

The [copr package](https://copr.fedorainfracloud.org/oprs/pschyska/alacritty/) uses the source from this project to build the rpm (I think).  Changes in this source probably will fix the copr.

I believe the root issue was caused by [adding the terminfo file in ncurses-base](https://bugzilla.redhat.com/show_bug.cgi?id=1849974).   The alacritty copr package should remove the terminfo file.

The fix is:
* Remove (2) references to terminfo in [extra/linux/redhat/alacritty.spec](../blob/master/extra/linux/redhat/alacritty.spec#39)
* Delete [extra/alacritty.info](../blob/master/extra/alacritty.info), the source file for the conflicting terminfo file. 
* Test [.copr/Makefile](../blob/master/.copr/Makefile)

pschyska:
@mikeslattery I think you shouldn't drop extra/alacritty.info . This is probably used by some non-Fedora packages.

pschyska:
I'm making a fix for that today on my personal branch for the time being and will re-trigger the build on copr. I heard alacritty is upstream now and we can drop the copr at some point.

[...]

PR #4540, opened 2020-12-08
Don't install custom terminfo on f33 or later

pschyska:
This fixes the rpmbuild for Fedora >=33. It uses the upstream ncurses terminfo (which added alacritty) instead of compiling with tic to prevent a file-conflict.
I also added Run-time-dependencies: freetype,fontconfig,libxcb, and fixed an issue with a directory remaining after package removal.

Arguably, the .copr and rpmspec could be removed here, as Fedora packages alacritty themselves now. I don't really see a reason to carry the copr, or the rpmspec around (other than 0.6 is not packaged yet, and 0.5.0 is broken in wayland), but that's for you to decide.

cc: #4518


chrisduerr (changes requested):
See https://github.com/alacritty/alacritty/pull/4538.

PR #4544, opened 2020-12-09
Fix a deprecation warning discovered in the build

erikh:
the image library has changed to_rgba to to_rgba8

hope this is at least a little helpful!

-Erik


kchibisov:
Have you noticed when doing 'cargo install'? If so cargo used latest versions of dependencies, thus not the one we're actually using, if you want to update dependencies update them properly.

erikh:
My apologies.

PR #4546, opened 2020-12-10
Don't draw missing glyphs for zerowidth characters

kchibisov:
This should prevent missing glyph symbols from zerowidth characters
obscuring normal glyphs.

### DIFF
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -7,8 +7,8 @@ use std::ptr;
 
 use bitflags::bitflags;
 use crossfont::{
-    BitmapBuffer, FontDesc, FontKey, GlyphKey, Rasterize, RasterizedGlyph, Rasterizer, Size, Slant,
-    Style, Weight,
+    BitmapBuffer, Error as RasterizerError, FontDesc, FontKey, GlyphKey, Rasterize,
+    RasterizedGlyph, Rasterizer, Size, Slant, Style, Weight,
 };
 use fnv::FnvHasher;
 use log::{error, info};
@@ -92,10 +92,17 @@ pub struct TextShaderProgram {
     u_background: GLint,
 }
 
+bitflags! {
+    struct GlyphFlags : u8 {
+        const COLORED = 0b0000_0001;
+        const MISSING_GLYPH = 0b0000_0010;
+    }
+}
+
 #[derive(Copy, Debug, Clone)]
 pub struct Glyph {
     tex_id: GLuint,
-    multicolor: bool,
+    flags: GlyphFlags,
     top: i16,
     left: i16,
     width: i16,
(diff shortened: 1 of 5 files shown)


alacritty/src/renderer/mod.rs
```diff
@@ -287,14 +295,27 @@ impl GlyphCache {
...
             rasterized.left += i32::from(glyph_offset.x);
             rasterized.top += i32::from(glyph_offset.y);
             rasterized.top -= metrics.descent as i32;
 
-            loader.load_glyph(&rasterized)
+            let mut glyph = loader.load_glyph(&rasterized);
+
+            // Mark glyph as missing.
+            if is_missing {
+                glyph.flags.insert(GlyphFlags::MISSING_GLYPH);
```
kchibisov, line 315: I don't really have an idea how to do it nicer than that. The whole purpose is to avoid going into crossfont and performing the entire fallback logic on each redraw when zerowidth char is missing in the font, since it could only work if we add missing glyph for that char into the glyph cache, but not drawn on screen.

If you have any idea how to do that better, I'd like to hear, however everything I can think of will require rework of the loading logic entirely, and not sure if it'll be any better than what we have now...

alacritty/src/renderer/mod.rs
```diff
@@ -209,8 +216,9 @@ impl GlyphCache {
 
     fn load_glyphs_for_font<L: LoadGlyph>(&mut self, font: FontKey, loader: &mut L) {
         let size = self.font_size;
-        for i in 32u8..=126u8 {
-            self.get(GlyphKey { font_key: font, c: i as char, size }, loader);
+        for character in 32u8..=126u8 {
+            let character = character as char;
+            self.get(GlyphKey { font_key: font, character, size }, loader);
```
chrisduerr, line 221: I don't think this is an improvement at all? `32u8..=126u8` doesn't return a character, so `i` is more appropriate.

alacritty/src/renderer/mod.rs
chrisduerr: I don't like adding this information to the glyph flags at all, logically that doesn't make much sense to me.

Maybe an enum could help out? So we cache either `Glyph { fields }` or `Missing`? The goal is just to get the information that the glyph is missing out of `self.cache.get(key)`, right? Also another more pragmatic solution is to just store the glyph as completely empty, if we don't render anything, there's no need for us to care?

One thing I certainly wouldn't want to do is reserve space in our atlas for missing glyphs, or have duplicate instances of the missing glyph texture in it. So ideally we'd have zero atlas space occupied by missing zerowidth and a single cell for the normal one.

chrisduerr (changes requested):

chrisduerr (approved):

chrisduerr:
I ran this against master and the render timer for missing glyphs (see https://github.com/alacritty/vtebench/pull/30) went from ~1200 usec to ~1000 usec. So this should also give a slight performance improvement for missing glyph characters.

PR #4550, opened 2020-12-10
Allow for Commands to take visible or all text as input (Revived)

### LINKED ISSUE #1615: Piping visible text to external programs

Axelen123:
Fixes #1615.
See also #3709.

I adapted the original PR to avoid merge conflicts.
I also added basic parallelization using rayon as suggested by Chris [here](https://github.com/alacritty/alacritty/pull/3709#issuecomment-671103599). After that, I documented the code and made sure that the tests still pass.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -681,9 +681,16 @@
 #
 # - `command`: Fork and execute a specified command plus arguments
 #
-#    The `command` field must be a map containing a `program` string and an
-#    `args` array of command line parameter strings. For example:
-#       `{ program: "alacritty", args: ["-e", "vttest"] }`
+#    The `command` field can be either only a string specifying the program, or
+#    a map of:
+#
+#    - `program` (required), string specifying the program
+#    - `args` (optional), array of command line parameter strings
+#    - `input` (optional), either VisibleText or AllText. If specified, then
+#      respectively either the visible text in the terminal or all the text
+#      including the scrollback is piped to the command's stdin.
+#
+#    For example: `{ program: "alacritty", args: ["-e", "vttest"] }`
 #
 # And optionally:
 #
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -159,6 +159,7 @@ impl Default for SearchState {
 pub struct ActionContext<'a, N, T> {
     pub notifier: &'a mut N,
     pub terminal: &'a mut Term<T>,
+    pub terminal_ptr: &'a Arc<FairMutex<Term<T>>>,
     pub clipboard: &'a mut Clipboard,
     pub mouse: &'a mut Mouse,
     pub received_count: &'a mut usize,
@@ -325,6 +326,11 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     }
 
     #[inline]
+    fn terminal_ptr(&self) -> &Arc<FairMutex<Term<T>>> {
+        self.terminal_ptr
+    }
+
+    #[inline]
     fn terminal_mut(&mut self) -> &mut Term<T> {
         self.terminal
     }
@@ -370,7 +376,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             args.push(arg.into());
         }
 
-        start_daemon(&alacritty, &args);
+        start_daemon(&alacritty, &args, None);
     }
 
     /// Spawn URL launcher when clicking on URLs.
@@ -387,9 +393,11 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             let end = Point::new(Line(end.line as i32 - display_offset as i32), end.column);
 
             let mut args = launcher.args().to_vec();
-            args.push(self.terminal.bounds_to_string(start, end));
+            let mut s = String::new();
+            self.terminal.write_bounds_to_string(&mut s, start, end);
+            args.push(s);
 
-            start_daemon(launcher.program(), &args);
+            start_daemon(launcher.program(), &args, None);
         }
     }
 
@@ -947,8 +955,11 @@ impl<N: Notify + OnResize> Processor<N> {
     }
 
     /// Run the event loop.
-    pub fn run<T>(&mut self, terminal: Arc<FairMutex<Term<T>>>, mut event_loop: EventLoop<Event>)
-    where
+    pub fn run<T>(
+        &mut self,
+        terminal_ptr: Arc<FairMutex<Term<T>>>,
+        mut event_loop: EventLoop<Event>,
+    ) where
         T: EventListener,
     {
         let mut scheduler = Scheduler::new();
@@ -1011,13 +1022,14 @@ impl<N: Notify + OnResize> Processor<N> {
                 },
             }
 
-            let mut terminal = terminal.lock();
+            let mut terminal = terminal_ptr.lock();
 
             let mut display_update_pending = DisplayUpdate::default();
             let old_is_searching = self.search_state.history_index.is_some();
 
             let context = ActionContext {
                 terminal: &mut terminal,
+                terminal_ptr: &terminal_ptr,
                 notifier: &mut self.notifier,
                 mouse: &mut self.mouse,
                 clipboard: &mut clipboard,
@@ -1077,7 +1089,7 @@ impl<N: Notify + OnResize> Processor<N> {
 
         // Write ref tests to disk.
         if self.config.ui_config.debug.ref_test {
-            self.write_ref_test_results(&terminal.lock());
+            self.write_ref_test_results(&terminal_ptr.lock());
         }
     }
 
@@ -1143,7 +1155,7 @@ impl<N: Notify + OnResize> Processor<N> {
 
                         // Execute bell command.
                         if let Some(bell_command) = &processor.ctx.config.ui_config.bell.command {
-                            start_daemon(bell_command.program(), bell_command.args());
+                            start_daemon(bell_command.program(), bell_command.args(), None);
                         }
                     },
                     TerminalEvent::ClipboardStore(clipboard_type, content) => {
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -8,6 +8,8 @@
 use std::borrow::Cow;
 use std::cmp::{max, min, Ordering};
 use std::marker::PhantomData;
+use std::mem;
+use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use glutin::dpi::PhysicalPosition;
@@ -20,12 +22,15 @@ use glutin::platform::macos::EventLoopWindowTargetExtMacOS;
 use glutin::window::CursorIcon;
 
 use alacritty_terminal::ansi::{ClearMode, Handler};
+use alacritty_terminal::config::CommandInput;
 use alacritty_terminal::event::EventListener;
 use alacritty_terminal::grid::{Dimensions, Scroll};
 use alacritty_terminal::index::{Boundary, Column, Direction, Line, Point, Side};
 use alacritty_terminal::selection::SelectionType;
+use alacritty_terminal::sync::FairMutex;
 use alacritty_terminal::term::search::Match;
 use alacritty_terminal::term::{ClipboardType, SizeInfo, Term, TermMode};
+use alacritty_terminal::thread;
 use alacritty_terminal::vi_mode::ViMotion;
 
 use crate::clipboard::Clipboard;
@@ -78,6 +83,7 @@ pub trait ActionContext<T: EventListener> {
     fn window(&self) -> &Window;
     fn window_mut(&mut self) -> &mut Window;
     fn terminal(&self) -> &Term<T>;
+    fn terminal_ptr(&self) -> &Arc<FairMutex<Term<T>>>;
     fn terminal_mut(&mut self) -> &mut Term<T>;
     fn spawn_new_instance(&mut self) {}
     fn change_font_size(&mut self, _delta: f32) {}
@@ -139,7 +145,61 @@ impl<T: EventListener> Execute<T> for Action {
                 ctx.scroll(Scroll::Bottom);
                 ctx.write_to_pty(s.clone().into_bytes())
             },
-            Action::Command(program) => start_daemon(program.program(), program.args()),
+            Action::Command(ref cmd) => {
+                let args = cmd.args().to_owned();
+                let input = cmd.input();
+                let program = cmd.program().to_string();
+
+                let ptr = ctx.terminal_ptr().clone();
+                thread::spawn_named("command", move || {
+                    let input = input.map(|i| {
+                        let term = ptr.lock();
+                        let cols = term.columns();
+                        let screen_lines = term.screen_lines();
+                        let bottommost_line = term.bottommost_line();
+
+                        match i {
+                            CommandInput::VisibleText => {
+                                let viewport_start = Line(-(term.grid().display_offset() as i32));
+                                let viewport_end = viewport_start + bottommost_line;
+
+                                let mut s = String::with_capacity(screen_lines * cols);
+                                term.write_bounds_to_string(
+                                    &mut s,
+                                    Point::new(viewport_start, Column(0)),
+                                    Point::new(viewport_end, term.last_column()),
+                                );
+                                s
+                            },
+                            CommandInput::AllText => {
+                                let mut s = String::with_capacity(term.total_lines() * cols);
+                                let mut y = term.topmost_line();
+
+                                mem::drop(term);
+
+                                while y <= bottommost_line {
+                                    let term = ptr.lock();
+
+                                    term.write_bounds_to_string(
+                                        &mut s,
+                                        Point::new(y, Column(0)),
+                                        Point::new(
+                                            min(y + screen_lines - 1, bottommost_line),
+                                            term.last_column(),
+                                        ),
+                                    );
+                                    y += screen_lines;
+
+                                    mem::drop(term);
+                                }
+                                s
+                            },
+                        }
+                    });
+
+                    start_daemon(&program, &args, input.as_deref());
+                });
+            },
             Action::Hint(hint) => {
                 ctx.hint_state().start(hint.clone());
                 ctx.mark_dirty();
@@ -1111,6 +1171,10 @@ mod tests {
 
         fn terminal(&self) -> &Term<T> {
             &self.terminal
+        }
+
+        fn terminal_ptr(&self) -> &Arc<FairMutex<Term<T>>> {
+            unimplemented!();
         }
 
         fn terminal_mut(&mut self) -> &mut Term<T> {
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -174,6 +174,12 @@ impl From<CursorBlinking> for bool {
     }
 }
 
+#[derive(ConfigDeserialize, Debug, Copy, Clone, PartialEq, Eq)]
+pub enum CommandInput {
+    VisibleText,
+    AllText,
+}
+
 #[derive(Deserialize, Debug, Clone, PartialEq, Eq)]
 #[serde(untagged)]
 pub enum Program {
@@ -182,6 +188,8 @@ pub enum Program {
         program: String,
         #[serde(default)]
         args: Vec<String>,
+        #[serde(default)]
+        input: Option<CommandInput>,
     },
 }
 
@@ -197,6 +205,13 @@ impl Program {
         match self {
             Program::Just(_) => &[],
             Program::WithArgs { args, .. } => args,
+        }
+    }
+
+    pub fn input(&self) -> Option<CommandInput> {
+        match self {
+            Program::Just(_) => None,
+            Program::WithArgs { input, .. } => *input,
         }
     }
 }
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -354,44 +354,44 @@ impl<T> Term<T> {
 
         if is_block {
             for line in (start.line.0..end.line.0).map(Line::from) {
-                res += &self.line_to_string(line, start.column..end.column, start.column.0 != 0);
+                self.write_line_to_string(
+                    &mut res,
+                    line,
+                    start.column..end.column,
+                    start.column.0 != 0,
+                );
 
                 // If the last column is included, newline is appended automatically.
                 if end.column != self.columns() - 1 {
                     res += "\n";
                 }
             }
-            res += &self.line_to_string(end.line, start.column..end.column, true);
+            self.write_line_to_string(&mut res, end.line, start.column..end.column, true);
         } else {
-            res = self.bounds_to_string(start, end);
+            self.write_bounds_to_string(&mut res, start, end);
         }
 
         Some(res)
     }
 
-    /// Convert range between two points to a String.
-    pub fn bounds_to_string(&self, start: Point, end: Point) -> String {
-        let mut res = String::new();
-
+    /// Write the range between two points to a String.
+    pub fn write_bounds_to_string(&self, buf: &mut String, start: Point, end: Point) {
         for line in (start.line.0..=end.line.0).map(Line::from) {
             let start_col = if line == start.line { start.column } else { Column(0) };
             let end_col = if line == end.line { end.column } else { self.last_column() };
 
-            res += &self.line_to_string(line, start_col..end_col, line == end.line);
+            self.write_line_to_string(buf, line, start_col..end_col, line == end.line);
         }
-
-        res
     }
 
-    /// Convert a single line in the grid to a String.
-    fn line_to_string(
+    /// Write a single line in the grid to a String.
+    fn write_line_to_string(
         &self,
+        buf: &mut String,
         line: Line,
         mut cols: Range<Column>,
         include_wrapped_wide: bool,
-    ) -> String {
-        let mut text = String::new();
-
+    ) {
         let grid_line = &self.grid[line];
         let line_length = min(grid_line.line_length(), cols.end + 1);
 
@@ -419,11 +419,11 @@ impl<T> Term<T> {
 
             if !cell.flags.intersects(Flags::WIDE_CHAR_SPACER | Flags::LEADING_WIDE_CHAR_SPACER) {
                 // Push cells primary character.
-                text.push(cell.c);
+                buf.push(cell.c);
 
                 // Push zero-width characters.
                 for c in cell.zerowidth().into_iter().flatten() {
-                    text.push(*c);
+                    buf.push(*c);
                 }
             }
         }
@@ -432,7 +432,7 @@ impl<T> Term<T> {
             && (line_length.0 == 0
                 || !self.grid[line][line_length - 1].flags.contains(Flags::WRAPLINE))
         {
-            text.push('\n');
+            buf.push('\n');
         }
 
         // If wide char is not part of the selection, but leading spacer is, include it.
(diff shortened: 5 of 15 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -1457,6 +1458,36 @@ impl<T> Term<T> {
             rendered: false,
         }
     }
+
+    fn grid_to_string_between(&self, y0: Line, y1: Line) -> String {
+        let grid = &self.grid;
+        // Use rayon to iterate over each row.
+        (*y0..*y1).into_par_iter().rev().map(|y| {
+            // Create a String and fill it with the contents of the current row.
+            let mut s = String::with_capacity(*grid.cols());
```
chrisduerr, line 1467: Allocating a string for every row is probable rather inefficient.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1457,6 +1458,36 @@ impl<T> Term<T> {
...
+            let row = &grid[y];
+            for x in 0..*grid.cols() {
+                let cell = &row[Column(x)];
+                s.push(cell.c);
+            }
```
chrisduerr, line 1472: This doesn't create the correct output string but instead will lead to a lot of trailing whitespace.

chrisduerr (changes requested):
Have you actually run any benchmarks to verify your `rayon` implementation improves the performance at all? Unless we have a benchmark for this conversion, it's kinda pointless to try and optimize it.

Axelen123:
There.
New version should create the correct output string and be more efficient.
I am going to write a benchmark now to verify that it improves performance.

Axelen123:
It appears that with [this](https://gist.github.com/Axelen123/8bfef130d9da9c06ace330e6514c6157) benchmark, running it in parallel is alot slower than serial:

Table of results (ns/iter):

|   -   | Parallel | Serial        |
|-----|----------|----------------|
| Old | 33,559   | 19,361   |
| New | 329,58   | 30,463 |

The new version is also slightly slower than the old version.
The new version is alot slower when ran in parallel.

My guess is that it gets alot harder for the CPU and/or compiler to optimize it (could of course also be that I made a mistake in the code or benchmark).

I am going to remove rayon (for now atleast). The new one is still pretty fast when you run it in serial.

chrisduerr:
At this point I'm not exactly sure what benefit this provides over #1615, most of the points mentioned there still stand.

Axelen123:
> At this point I'm not exactly sure what benefit this provides over #1615, most of the points mentioned there still stand.

One improvement I made [here](https://github.com/Axelen123/alacritty/blob/c11ded0d22c9f62bc8a75be512f009da67d7817b/alacritty/src/input.rs#L174) is that it calls `start_daemon()` in another thread. This means that it will not write to the program's stdin on the main thread.

alacritty_terminal/src/term/mod.rs
Axelen123: Marking this one as resolved because the new version does not do that anymore.

alacritty_terminal/src/term/mod.rs
Axelen123: It now iterates over each row using `row.occ`. It reduces the amount of trailing whitespace unless something like `.last_mut()`  was called on the row. Now it also calls `.trim_end()` on the output in a different thread.

Axelen123:
I think I will work on it a bit more and maybe clean up the commit history. After that I will request another review.

Axelen123:
I am ready to mark this PR as ready for review. I have made several improvements.

Changelog:
- Fixed the trailing whitespace issue (for real)
- Output is now exactly the same as st (apart from how st handles chars typed by the user)
- It is faster (tested it with the same benchmark as before)
- I have also changed the tests
- Commit history is cleaner

Axelen123:
If you want to know how st does it, then you should check out the `tlinelen()` function inside [st.c](https://git.suckless.org/st/file/st.c.html#l413) and the [externalpipe](https://st.suckless.org/patches/externalpipe/st-externalpipe-0.8.4.diff) patch.

alacritty/src/event.rs
```diff
@@ -552,6 +552,14 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     fn scheduler_mut(&mut self) -> &mut Scheduler {
         self.scheduler
     }
+
+    fn to_string(&self) -> String {
+        self.terminal.grid_to_string()
+    }
```
chrisduerr, line 558: The `to_string` method should never be manually implemented like this. The `Display` trait should be used instead.

alacritty/src/input.rs
```diff
@@ -156,12 +160,20 @@ impl<T: EventListener> Execute<T> for Action {
...
 
-                start_daemon(program, args);
+                let input_str = input.map(|i| match i {
+                    CommandInput::VisibleText => ctx.to_string_only_visible(),
+                    CommandInput::AllText => ctx.to_string(),
+                });
+
+                spawn(move || {
+                    start_daemon(&program, &args, input_str.as_deref());
+                });
```
chrisduerr, line 176: Threads should ideally be spawned with a name and functions shouldn't be directly imported.

alacritty/src/input.rs
```diff
@@ -1267,6 +1279,14 @@ mod tests {
...
             unimplemented!();
         }
+
+        fn to_string(&self) -> String {
+            self.terminal.grid_to_string()
+        }
+
+        fn to_string_only_visible(&self) -> String {
+            self.terminal.grid_to_string_only_visible()
+        }
```
chrisduerr, line 1289: Why make this a function on the grid and terminal, when we can just non-mutably access the grid from the terminal using `self.terminal.grid()`?

alacritty_terminal/src/config/mod.rs
```diff
@@ -271,6 +271,12 @@ impl Into<bool> for CursorBlinking {
     }
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize)]
```
chrisduerr, line 274:
```suggestion
#[derive(Deserialize, Debug, Copy, Clone, PartialEq, Eq)]
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -1457,6 +1457,57 @@ impl<T> Term<T> {
             rendered: false,
         }
     }
+
+    fn grid_to_string_between(&self, y0: Line, y1: Line) -> String {
+        let grid = &self.grid;
+
+        let mut s = String::with_capacity(*(y1 - y0) * *grid.cols());
```
chrisduerr, line 1464: This will cause problems when y0 is bigger than y1.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1457,6 +1457,57 @@ impl<T> Term<T> {
...
+            // Calculate the line length. We need to do this to avoid trailing spaces.
+            if wrapline {
+                line_len = *grid.cols();
+            } else {
+                for n in (0..*grid.cols()).rev() {
+                    if row[Column(n)].c != ' ' {
+                        line_len = n + 1;
+                        break;
+                    }
+                }
```
chrisduerr, line 1480: This is not the correct way to calculate line length/emptiness.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1457,6 +1457,57 @@ impl<T> Term<T> {
...
+
+            if !wrapline {
+                s.push('\n');
+            }
+        }
+
+        // Ensure that the string ends with a newline.
+        if !s.ends_with('\n') {
+            s.push('\n');
+        }
```
chrisduerr, line 1496: The last line cannot be `wrapline` iirc, so I don't think this branching is necessary?

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
Axelen123: While that is probably true for the last line in the grid, the last **visible** line can be wrapline if you scroll up.

alacritty/src/input.rs
Axelen123: Fixed it

alacritty_terminal/src/term/mod.rs
Axelen123: Oh. I found the `LineLength` trait that `grid::Row<Cell>` implements. I am assuming that is the correct way? Please tell me if that is not the case.

alacritty_terminal/src/term/mod.rs
Axelen123: I'm not sure if it could actually happen.
`grid_to_string_between()` is a private function.
Adding a few assertions might still be a good idea.

alacritty/src/input.rs
Axelen123: It is implemented on the terminal and action context but not on the grid.

alacritty_terminal/src/term/mod.rs
chrisduerr: It is not, sorry for not being more specific here. But checking for cells matching `' '` will cut off whitespace at the end. You should iterote over the cells in reverse and check `!cell.is_empty()`.

alacritty_terminal/src/term/mod.rs
chrisduerr: Yeah, I'd recommend to at the very least add an assertion for something that might cause seriously wrong behavior at runtime.

If it can't happen, it should at least explode with a nice error.

alacritty/src/input.rs
chrisduerr: Ah, it's not implemented on the grid because that doesn't know about its cell's contents, right?

Axelen123: That is correct. It is implemented on `Term<T>` because the grid is always `Grid<Cell>`.

chrisduerr:
@kchibisov Do you think we should provide the ability to pass colored text? It would probably be useful though with 100k scrollback it would take a second.

ayosec:
> provide the ability to pass colored text?

IMO, it is very useful to add an option to send escape sequences to replicate the colors in the screen. Thus, users can add extra features without changing the code in Alacritty.

For example, to add something like *Export to HTML*:

```yaml
command:
  input: AllTextWithEscSeqs
  program: sh
  args:
    - -c
    - |
      # Convert to HTML with https://github.com/theZiz/aha
      aha | xsel -b
```

Another feature that can be implemented is to take a snapshot of the current terminal output, so we can keep it to compare command outputs.

```yaml
command:
  input: AllTextWithEscSeqs
  program: bash
  args:
    - -c
    - |
      OUTPUT=$(mktemp)
      cat > "$OUTPUT"
      alacritty -e less +G -R "$OUTPUT"
```

kchibisov:
>@kchibisov Do you think we should provide the ability to pass colored text? It would probably be useful though with 100k scrollback it would take a second.

I mean we should, but I think we should make this async, so the block won't be critical. However I can't really think of a way to make it async, since you basically should clone the entire terminal, but cloning it alone will take a lot of time...

However if we have something that could run for seconds and literally freeze the app, we should provide some feedback on what we're doing, like continuing to process events, but not releasing the terminal lock, etc.

alacritty_terminal/src/term/mod.rs
Axelen123: I think it is correct now.

alacritty_terminal/src/term/mod.rs
Axelen123: I have added them. I don't know if we need any more assertions here.

Axelen123:
I believe we should support colored text.

> I mean we should, but I think we should make this async, so the block won't be critical. However I can't really think of a way to make it async, since you basically should clone the entire terminal, but cloning it alone will take a lot of time...

I made a new branch [here](https://github.com/Axelen123/alacritty/tree/externalpipe-poc) that runs `grid_to_string_between()` in another thread. It is very hacky but it uses scoped threads to get around some of the lifetime issues. My idea is that maybe we could lock the terminal, do a few lines and then release the lock. I could be wrong here but I think it would sort of make it async and we don't have to clone the entire terminal. The only problems are:
- The result will probably be incorrect if the grid changes after we unlock it.
- `Term<T>` does not require that `T` is `Send`.

I looked at the way @chrisduerr implemented color support and the diff [is pretty big](https://github.com/alacritty/alacritty/compare/master...chrisduerr:pager) so maybe it deserves it's own PR?

chrisduerr:
I don't think it makes sense to put color in a different PR, since that fundamentally affects the way the grid is transformed into characters. So just doing it one way first and then having to rework it completely just doesn't make much sense. I wouldn't want to get into a position where the threshold to implementing it properly is higher than before it was supported at all.

Axelen123:
There. Most of it was copied from [here](https://github.com/chrisduerr/alacritty/tree/pager).
Still need to resolve some of those TODOs.

alacritty/src/event.rs
Axelen123: That is correct but I don't think we should do it anymore because the signature of `to_string()` has changed.

Axelen123: Honestly, I think it is better to just remove the functions from the action context and use `grid_to_string()` directly.

chrisduerr:
>  Is there a good ref test recording with colors that I can use to benchmark it?

I'm not aware of any, but just 100k of anything should work pretty well. The dense cell benchmark of vtebench is probably a good one, though you need to change that to actually use scrollback. I'm pretty sure I've just written my own benchmark for my pager branch.

Axelen123:
I have added async support.
There are still quite a few TODOs but it is almost done.

I believe that the best option is to give the user two choices:
# Sync
If `lines_per_run` is zero then it is effectively synchronous.
The result is guaranteed to be correct but Alacritty will freeze until it is done.
IMO this option is better if you only care about visible text or if you don't need colors.

# Async
If `lines_per_run` is greater than zero then it is asynchronous.
Alacritty won't freeze but the result will not be correct if the program changes any previous line that has already been processed.
This can be mitigated in most cases by setting `lines_per_run` to the amount of lines shown on the screen as the program is unlikely to change any lines in the scrollback (I am not even sure if you can).
If `lines_per_run` is set to a high value, the result is more likely to be correct but it will be less responsive.

chrisduerr:
I don't think this choice should be put on the user. We should know best which approach gives the best results, putting that on the user to decide is just a bit lazy. If our approach isn't fast enough to please everyone, we need to make it faster.

shixianli158:
like you same

shixianli158:
hh

Axelen123:
> I don't think this choice should be put on the user. We should know best which approach gives the best results, putting that on the user to decide is just a bit lazy. If our approach isn't fast enough to please everyone, we need to make it faster.

Yes. I was not planning on getting it merged without good defaults. I thought locking and unlocking the mutex inside the while loop would be more expensive, which would have made it beneficial to give the user control over `lines_per_run`. Since that is not the case, I believe that the best option is to set `lines_per_run` to `grid.screen_lines()` for correctness. I added a configuration value called `pipe_responsive_mode` that sets `lines_per_run` to 10. This will probably be useful for devices that have very weak CPUs. It should still be correct in almost all situations even with `pipe_responsive_mode` enabled.
I still need to add more documentation. There is probably also a better place and name for `pipe_responsive_mode`.

chrisduerr:
I'll be very clear here: I don't want to see any configuration option for this in the config file.

Axelen123:
It is now a cargo feature instead of a config file option. I don't think this will be a permanent solution. I am still working on it.

chrisduerr:
If we want to merge this, we should find a solution that always works without any configuration.

Axelen123:
Do you remember anything about the comment that you wrote [here](https://github.com/chrisduerr/alacritty/blob/8ce1e06b4b4827a4eb043cf7aaa5fc78ddda9026/alacritty_terminal/src/ansi.rs#L628)?

chrisduerr:
Yes, the comment is exactly what it says. The check has a cost itself and in my tests it did not reliably improve performance. So more tests would be required.

But most of my code was just prototyping really. None of that was in a mergeable state.

alacritty.yml
```diff
@@ -640,9 +640,18 @@
...
+#    The `command` field can be either only a string specifying the program, or
+#    a map of:
+#
+#    - `program` (required), string specifying the program
+#    - `args` (optional), array of command line parameter strings
+#    - `input` (optional), either VisibleText or AllText. If specified, then
+#      respectively either the visible text in the terminal or all the text
+#      including the scrollback is piped to the command's stdin.
+#    - `pipe_escape_sequences` (optional), if set to `true`, then Alacritty will also
+#      pipe ANSI escape sequences. This is slow so it should be left disabled when possible.
```
chrisduerr, line 652: This seems like a dangerous way to do configuration. What do we expect to happen when input isn't present but `pipe_escape_sequences` is?

Why was this split into two different fields when they're so tightly coupled together?

alacritty/src/input.rs
```diff
@@ -153,12 +159,38 @@ impl<T: EventListener> Execute<T> for Action {
...
+                    let input = input.map(|mut collector| {
+                        let mut done = false;
+                        while !done {
+                            let term = term.lock();
+                            done = collector.execute(term.grid());
+                            mem::drop(term);
+                        }
+
+                        collector.finish()
+                    });
```
chrisduerr, line 190: Do you spawn your stuff in a thread here because it's so slow?

alacritty_terminal/src/ansi.rs
```diff
@@ -1274,6 +1335,37 @@ fn parse_sgr_color(params: &mut dyn Iterator<Item = u16>) -> Option<Color> {
...
+    "150", "151", "152", "153", "154", "155", "156", "157", "158", "159", "160", "161", "162",
+    "163", "164", "165", "166", "167", "168", "169", "170", "171", "172", "173", "174", "175",
+    "176", "177", "178", "179", "180", "181", "182", "183", "184", "185", "186", "187", "188",
+    "189", "190", "191", "192", "193", "194", "195", "196", "197", "198", "199", "200", "201",
+    "202", "203", "204", "205", "206", "207", "208", "209", "210", "211", "212", "213", "214",
+    "215", "216", "217", "218", "219", "220", "221", "222", "223", "224", "225", "226", "227",
+    "228", "229", "230", "231", "232", "233", "234", "235", "236", "237", "238", "239", "240",
+    "241", "242", "243", "244", "245", "246", "247", "248", "249", "250", "251", "252", "253",
+    "254", "255"
+];
```
chrisduerr, line 1367: These implementation details you took from my personal tests are not appropriate for production use. Their purpose was solely for trying to figure out what the optimal possible performance of this implementation might be.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1743,6 +1762,147 @@ impl<T: EventListener> Handler for Term<T> {
...
+    end: Line,
+    start: Line,
+    cols: Column,
+}
+
+impl GridCollector {
+    fn new<T>(term: &Term<T>, start: Line, end: Line, mode: CollectorMode, esc_seqs: bool) -> Self {
+        assert!(*start < term.grid.total_lines());
+        assert!(*end <= term.grid.total_lines());
+        assert!(start < end);
```
chrisduerr, line 1796: I don't want any runtime crashes please.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1743,6 +1762,147 @@ impl<T: EventListener> Handler for Term<T> {
...
+impl From<usize> for CollectorMode {
+    fn from(value: usize) -> Self {
+        match NonZeroUsize::new(value) {
+            Some(value) => Self::Async(value),
+            None => Self::Sync,
+        }
+    }
+}
+
+pub struct GridCollector {
```
chrisduerr, line 1780: Why do we need a completely different structure for iterating over the grid and converting it to a string, instead of using what already exists?

This entire struct and its methods seem like a lot of complexity that is duplicated from other places and will easily haunt us in the future.

chrisduerr (changes requested):
This is just a surface-level review, but I think there's a lot of code quality improvements left to be made.

asdf8dfafjk:
If you don't mind a bystander's comment, would you consider replacing `/bin/bash` with `/usr/bin/env bash`?

chrisduerr:
> If you don't mind a bystander's comment, would you consider replacing /bin/bash with /usr/bin/env bash?

Please don't comment on things you do not fully understand.

alacritty.yml
michaeldel: Would using a pipe to combine them be more suitable? Like
```yml
input: VisibleText | EscapeSequences
```
This syntax is already used for keybindings and we could require having at least `VisibleText` or `AllText` present for `EscapeSequences` to be usable in the field.

alacritty/src/input.rs
Axelen123: No. I spawn it in a thread because it allows me to make it asynchronous.

alacritty/src/event.rs
```diff
@@ -999,8 +1007,11 @@ impl<N: Notify + OnResize> Processor<N> {
     }
 
     /// Run the event loop.
-    pub fn run<T>(&mut self, terminal: Arc<FairMutex<Term<T>>>, mut event_loop: EventLoop<Event>)
-    where
+    pub fn run<T>(
+        &mut self,
+        terminal_ptr: Arc<FairMutex<Term<T>>>,
```
chrisduerr, line 1012: What is going on here? Why did you rename these things?

alacritty_terminal/build.rs
```diff
@@ -0,0 +1,15 @@
...
+
+fn main() {
+    let dest = env::var("OUT_DIR").unwrap();
+    let mut file = File::create(&Path::new(&dest).join("ansi_array.rs")).unwrap();
+
+    file.write_all(b"pub const U8_TO_STR: [&str; 256] = [").unwrap();
+    for i in 0..=u8::MAX {
+        write!(file, r#""{};","#, i).unwrap();
+    }
+    file.write_all(b"];").unwrap();
```
chrisduerr, line 14: No, this is not how to do this.

chrisduerr (changes requested):

alacritty_terminal/build.rs
Axelen123: How are you supposed to do it?

chrisduerr: You should use a fast but readable implementation.

Axelen123: I changed it. Is it better now?

alacritty_terminal/src/term/mod.rs
```diff
@@ -418,12 +426,17 @@ impl<T> Term<T> {
             }
 
             if !cell.flags.intersects(Flags::WIDE_CHAR_SPACER | Flags::LEADING_WIDE_CHAR_SPACER) {
+                if esc_seqs {
+                    cell.as_escape(buf, last_cell);
```
nixpulvis, line 430: I'm wondering if it wouldn't makes sense to split up this PR into one for the basic functionality and then a follow-on for adding back in various escapes. Personally, I think this should be two separate commands users can choose between anyways, much like how copy/paste today work without this. If done properly, perhaps we could think about adding a new kind of `Copy` action as well.

chrisduerr: With escape sequences things might get slowed down significantly though, so an approach that works for plain text might not be workable for colored text.

nixpulvis: A lot of the code is going to be the same. Spawning the process, modifying the config, etc.

alacritty/src/event.rs
Axelen123: I renamed it because I need the `Arc` pointer for the action context (used by the `Action::Command` handler later) but [this line](https://github.com/alacritty/alacritty/blob/17923efccfb7e6bbcdeb5f748796d9ee1ea70149/alacritty/src/event.rs#L1066) redefines the variable.

alacritty_terminal/src/term/mod.rs
Axelen123: > A lot of the code is going to be the same. Spawning the process, modifying the config, etc.

That is mostly true as long as we keep it asynchronous.

Axelen123:
I am requesting another review because no one has responded in over a week.

alacritty_terminal/src/term/mod.rs
Axelen123: It is done

alacritty_terminal/build.rs
Axelen123: Update:
This is no longer applicable right now but I still considered it to be solved by my update.

Axelen123:
I am no longer interested in working on this due to a lack of communication, motivation, and time.

asdf8dfafjk:
Sorry to hear, thanks for all the hard work @Axelen123 

Personally I'm looking forward to try wezterm, less blowhardy, probably better too :)

methuselah-0:
I was looking for this functionality, and instead made a workaround for this with ydotool, wl-clipboard and a little shellscript but as wish for myself that I had found something already documented, here's how-to for others until this PR/functionality gets revived:
```
#!/bin/sh
[ "$1" = -t ] && {
    # Ctrl+Shift+Space
    sudo ydotool key 29:1 42:1 57:1 57:0 42:0 29:0
    # sleep 0.1

    # ggVG
    sudo ydotool key 34:1 34:0 34:1 34:0 42:1 47:1 47:0 34:1 34:0 42:0
    # sleep 0.1

    # Ctrl+Shift+C
    sudo ydotool key 29:1 42:1 46:1 46:0 57:0 42:0 29:0
    # sleep 0.1

    # Ctrl+Shift+Space
    sudo ydotool key 29:1 42:1 57:1 57:0 42:0 29:0
}

TMP="$(mktemp)" && wl-paste > "$TMP" && emacsclient -t -r "$TMP" ; rm "$TMP"
```

PR #4556, opened 2020-12-11
Replace serde's derive with custom proc macro

chrisduerr:
This replaces the existing `Deserialize` derive from serde with a
`ConfigDeserialize` derive. The goal of this new proc macro is to allow
a more error-friendly deserialization for the Alacritty configuration
file without having to manage a lot of boilerplate code inside the
configuration modules.

The first part of the derive macro is for struct deserialization. This
takes structs which have `Default` implemented and will only replace
fields which can be successfully deserialized. Otherwise the `log` crate
is used for printing errors. Since this deserialization takes the
default value from the struct instead of the value, it removes the
necessity for creating new types just to implement `Default` on them for
deserialization.

Additionally, the struct deserialization also checks for `Option` values
and makes sure that explicitly specifying `none` as text literal is
allowed for all options.

The other part of the derive macro is responsible for deserializing
enums. While only enums with Unit variants are supported, it will
automatically implement a deserializer for these enums which accepts any
form of capitalization.

Since this custom derive prevents us from using serde's attributes on
fields, some of the attributes have been reimplemented for
`ConfigDeserialize`. These include `#[config(flatten)]`,
`#[config(skip)]` and `#[config(alias = "alias)]`. The flatten attribute
is currently limited to at most one per struct.

Additionally the `#[config(deprecated = "optional message")]` attribute
allows easily defining uniform deprecation messages for fields on
structs.


alacritty_terminal/src/config/mod.rs
```diff
@@ -17,132 +16,103 @@ pub use crate::config::bell::{BellAnimation, BellConfig};
 pub use crate::config::colors::Colors;
 pub use crate::config::scrolling::Scrolling;
 
-pub const LOG_TARGET_CONFIG: &str = "alacritty_config";
-const DEFAULT_CURSOR_THICKNESS: f32 = 0.15;
-const MAX_SCROLLBACK_LINES: u32 = 100_000;
+pub const LOG_TARGET_CONFIG: &str = "alacritty_config_derive";
```
chrisduerr, line 19: Currently this is hardcoded, I'm not sure there's any good way to share this constant with `alacritty_config_derive`. The only option would be to import it from `alacritty_config_derive`, but that would create a cyclic dependency.

Unfortunately it's not possible to export constants from proc macros.

alacritty/src/config/window.rs
```diff
@@ -108,27 +99,44 @@ impl WindowConfig {
...
+    #[inline]
+    pub fn set_general(&mut self, general: String) {
+        match &mut self.class {
+            Class::Just(instance) => {
+                let instance = instance.clone();
+                self.class = Class::WithGeneral { instance, general };
+            },
+            Class::WithGeneral { general: g, .. } => *g = general,
         }
     }
```
chrisduerr, line 136: I'd say this is probably the worst "regression" when it comes to code quality with the new deserializer.

Unfortunately I'm not sure about a good way to abstract over untagged enums with struct and tuple variants. So I'm currently uncertain how I would improve this specific part.

alacritty_config_derive/src/de_struct.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        // Create token stream for deserialization and error handling.
+        field_streams.match_assignments.extend(quote! {
+            #(#literals)|* => { #match_assignment_stream },
+        });
+    }
+
+    field_streams
+}
```
chrisduerr, line 171: This is an absolutely massive function and I've tried to refactor it a few times to see if there's maybe a more elegant way to do this, but I couldn't find anything that would actually make this easier to read.

Since there's a lot of references to shared data over the entire lifetime of that method, it makes it a bit difficult to move things into functions without creating unreasonable control flow or massive parameter lists.

chrisduerr (commented):
These are the last remaining style questions I'm not 100% certain about.

alacritty_config_derive/src/de_enum.rs
```diff
@@ -0,0 +1,64 @@
...
+        let variant_str = variant_ident.to_string();
+        available_values = format!("{}`{}`, ", available_values, variant_str);
+
+        let literal = variant_str.to_lowercase();
+
+        match_arms_stream.extend(quote! {
+            #literal => Ok(#ident :: #variant_ident),
+        });
+    }
+    available_values.truncate(available_values.len().saturating_sub(2));
```
kchibisov, line 28: This subtract isn't very clear(why 2), would suggest to add a comment.

alacritty_config_derive/src/de_struct.rs
```diff
@@ -0,0 +1,218 @@
...
+use quote::{format_ident, quote};
+use syn::parse::{self, Parse, ParseStream};
+use syn::punctuated::Punctuated;
+use syn::spanned::Spanned;
+use syn::{Error, Field, GenericParam, Generics, Ident, LitStr, Token, Type, TypeParam};
+
+/// Error message when attempting to flatten multiple fields.
+const MULTIPLE_FLATTEN_ERROR: &str = "At most one instance of #[config(flatten)] is supported";
+/// Use this crate's name as log target.
+const LOG_TARGET: &str = env!("CARGO_PKG_NAME");
```
kchibisov, line 12: If you have concerns wrt using target log name, etc in proc macros. You may look into wayland-scanner magic, since it works by assuming that you 'use' certain things, and it'll generate either server or client code for that.

The wayland-scanner usage of quote is a bit complicated, but you have an option to provide custom target if that's what you want.

Since you had a problem of exporting log target info, but you can actually do that the other way around and bring it in(not sure if you can do that conditionally though).

kchibisov (commented):
Just some minor comments, can't look more into it unfortunately.

alacritty/src/config/debug.rs
```diff
@@ -1,64 +1,33 @@
...
-use alacritty_terminal::config::{failure_default, LOG_TARGET_CONFIG};
+use alacritty_config_derive::ConfigDeserialize;
 
 /// Debugging options.
-#[serde(default)]
-#[derive(Deserialize, Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(ConfigDeserialize, Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Debug {
-    #[serde(default = "default_log_level", deserialize_with = "deserialize_log_level")]
     pub log_level: LevelFilter,
```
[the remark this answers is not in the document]
chrisduerr: We can definitely update our docs for that.

alacritty/src/config/font.rs
```diff
@@ -17,62 +14,41 @@ use crate::config::ui_config::Delta;
...
-#[derive(Debug, Deserialize, Clone, PartialEq, Eq)]
+#[derive(ConfigDeserialize, Default, Debug, Clone, PartialEq, Eq)]
 pub struct Font {
+    /// Extra spacing per character.
+    pub offset: Delta<i8>,
+
+    /// Glyph offset within character cell.
+    pub glyph_offset: Delta<i8>,
+
+    pub use_thin_strokes: bool,
```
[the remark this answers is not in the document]
chrisduerr: Because it's pointless. We have a method just for non-macos to always return false anyways and we have it documented in our config either way. If we wanted to warn on missing fields we wouldn't want to warn on this. So why have **more** code to remove it on a platform when there's no benefit to it at all?

alacritty/src/config/mod.rs
```diff
@@ -283,56 +281,6 @@ fn installed_config() -> Option<PathBuf> {
...
-    #[cfg(all(windows, not(feature = "winpty")))]
-    if config.winpty_backend {
-        warn!(
-            target: LOG_TARGET_CONFIG,
-            "Config winpty_backend is deprecated and requires a compilation flag; it should be \
-             removed from the config",
-        )
-    }
-}
-
```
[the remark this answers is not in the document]
chrisduerr: The messages aren't gone, they have been moved to attributes on the specific field. That's one of the features of the library I've written.

alacritty_config_derive/src/de_struct.rs
chrisduerr: I'm not sure what you mean? Import the target from the outside?

That's something I cannot do because it is completely opaque to the user of the library and there's no way to "detect" the imports from the proc macro.

alacritty/src/config/debug.rs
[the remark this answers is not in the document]
chrisduerr: I've documented the field as `Off`, that should be sufficient. I don't think there's any need to support `None` here.

I've tried it with all things capitalized but I don't really like the way that looks, I personally don't think it should be much of an issue that it prints everything capitalized when both work just fine.

alacritty_terminal/src/config/scrolling.rs
```diff
@@ -1,54 +1,36 @@
-use log::error;
+use serde::de::Error as SerdeError;
 use serde::{Deserialize, Deserializer};
 
-use crate::config::{failure_default, LOG_TARGET_CONFIG, MAX_SCROLLBACK_LINES};
+use alacritty_config_derive::ConfigDeserialize;
+
+/// Maximum scrollback amount configurable.
+const MAX_SCROLLBACK_LINES: u32 = 100_000;
```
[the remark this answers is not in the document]
chrisduerr: It does not. Separating by thousands is not a problem.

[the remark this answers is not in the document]
kchibisov: Huh, ok.

alacritty_config_derive/src/de_struct.rs
kchibisov: I got confused by different rust macros and how they work. Nothing will help from that I've mentioned.

PR #4562, opened 2020-12-12
Fix live reload with broken yaml on start

### LINKED ISSUE #4561: Config reload does not work if config is broken on launch
When using a broken configuration file to launch Alacritty, fixing the error will not remove the error from Alacritty. Instead live reload is completely disabled if the config file has any errors.

It's not necessary to live reload config files that are missing or have invalid permissions, but for configuration files which can be read without any problem the live config reload should always be enabled.

I'm not certain this has behaved different previously, but I think this might be a regression introduced by CLI config overrides.

chrisduerr:
Since the current behavior would just load the default configuration
file whenever the configuration file couldn't be loaded, the path was
not set to any value. As a result however, the live config reload
feature would not work with a broken yaml (one which cannot be
deserialized, not one with warnings).

If a configuration file has been specified, but the deserialization
still failed, the path is now preserved on the default configuration
file to make it possible to live reload a fix for the issue.

Fixes #4561.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Window not being completely opaque on Windows
 - Window being always on top during alt-tab on Windows
 - Cursor position not reported to apps when mouse is moved with button held outside of window
+- No live config update when starting Alacritty with a broken configuration file
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


chrisduerr:
I've noted in the original issue that permissions might justify not creating a config watcher, but based on my testing the configuration file watcher works great even without permissions to read the file. As soon as the permission issue is fixed, the configuration file is loaded and ready to go.

kchibisov (approved):

PR #4567, opened 2020-12-13
Fix cursor reports with mouse outside of window

### LINKED ISSUE #4566: Mouse mode motions are not reported when button is pressed and the pointer is outside of the terminal
Other terminals seem to report that and I'm not sure why we're not doing so.  The reported values should be clamped to viewport.

chrisduerr:
Previously Alacritty would not report cursor escapes to the application
when a mouse button was held down and the mouse was moved. This prevents
applications like tmux from updating their selection.

Similarly to how windowing libraries keep reporting mouse events when
the left mouse button is held down over the window, the escape sequences
are now clamped to within the grid and reported to applications.

Fixes #4566.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Compilation when targetting aarch64-apple-darwin
 - Window not being completely opaque on Windows
 - Window being always on top during alt-tab on Windows
-- No live config update when starting Alacritty with a broken configuration file
 - Cursor position not reported to apps when mouse is moved with button held outside of window
 
 ### Removed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Compilation when targetting aarch64-apple-darwin
 - Window not being completely opaque on Windows
 - Window being always on top during alt-tab on Windows
+- No live config update when starting Alacritty with a broken configuration file
```
kchibisov, line 36:
```suggestion
```

kchibisov (changes requested):

kchibisov (approved):

PR #4580, opened 2020-12-16
Fix invalid ESC escape sequence parsing

chrisduerr:
This strictens the ESC escape sequence parser to prevent invalid
intermediates from being ignored. Previously the parser would just look
at the first intermediate without validating that the rest of them is
empty.

If an escape like `\e(#0` is used now, it will no longer be accepted as
`\e(0` since the intermediate `#` is also present.


Imrandanish951:
Unsubscribe

chrisduerr:
Definitely 5.

kchibisov (approved):

PR #4581, opened 2020-12-16
Increase file watcher delay on BSD

### LINKED ISSUE #3871: High load while IDLE under FreeBSD
On my system, alacritty is causing about 2.5-3.0 CPU load (according to htop) while only showing a prompt. Other terminals, st for example, show a load of 0.0 when similary idle...

Via dtrace(1) i gathered the syscalls done by the system, and this showed me, that alacritty generates 2774 syscalls in 3.4s.

Those are the syscalls generated by alacritty during idle:
```
  compat11.kevent                                                  10
  recvmsg                                                          50
  close                                                           639
  compat11.stat                                                   639
  fstatfs                                                         639
  nanosleep                                                       639
  open                                                            639
  getdirentries                                                  1278
  compat11.lstat                                                 1917
```
truss(1) showed, that those open/getdirentries calls were done on $HOME/.config/alacritty and the config file it is containing.

To me it seems that alacritty is really busy querying its XDG_CONFIG dir and file (at least on FreeBSD).

### System
OS: FreeBSD 12.1p6
Version: alacritty 0.4.3
Linux/BSD: X11, Openbox

chrisduerr:
Since BSD does not support inotify polling in the `notify` crate and
instead relies on manual filesystem polling, this would cause a high CPU
usage at 100 polls a second.

A separate polling rate of once per second is now used for platforms
which do not support filesystem polling, allowing users to still make
use of live config reload on BSD.

Fixes #3871.

### DIFF
--- a/alacritty/src/config/monitor.rs
+++ b/alacritty/src/config/monitor.rs
@@ -10,6 +10,11 @@ use alacritty_terminal::thread;
 
 use crate::event::{Event, EventProxy};
 
+#[cfg(any(target_os = "linux", target_os = "macos", target_os = "windows"))]
+const DEBOUNCE_DELAY: Duration = Duration::from_millis(10);
+#[cfg(not(any(target_os = "linux", target_os = "macos", target_os = "windows")))]
+const DEBOUNCE_DELAY: Duration = Duration::from_millis(1000);
+
 pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventProxy) {
     // Canonicalize all paths, filtering out the ones that do not exist.
     paths = paths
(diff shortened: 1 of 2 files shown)


chrisduerr:
We could in theory wait for a fix in `notify`, but I think even if it gets fixed on some BSD platforms it will probably take a while until all BSD platforms will support inotify-like functionality with the `notify` crate. So this should be good as a fallback for these platforms.

alacritty/src/config/monitor.rs
```diff
@@ -10,6 +10,11 @@ use alacritty_terminal::thread;
 
 use crate::event::{Event, EventProxy};
 
+#[cfg(any(target_os = "linux", target_os = "macos", target_os = "windows"))]
+const DEBOUNCE_DELAY: Duration = Duration::from_millis(10);
```
kchibisov, line 14: Does this delay rely matter for linux/macos/windows? Could just have 1 sec for everything.

chrisduerr: Yes, the debounce period applies to linux/macos/windows too. With the file watcher it's the frequency for which the file is polled, with inotify it's the time for which an event is held before it is dispatched, to make sure operations that happen together don't unnecessarily trigger things twice.

kchibisov (approved):

PR #4585, opened 2020-12-17
Fix crash due to clipboard not being properly released on Wayland

kchibisov:


chrisduerr (approved):
Looks good, I'd assume that "released" in this case just means letting go of a handle to it in a certain scenario? In that case that should probably be fine.

kchibisov:
Yeah, it's basically when some error occurred and we had to manually drop it or something.  Not sure if I can even repro that with alacritty.

In a wild, you need to drop clipboard in the middle of handling stuff or something like that afaics.

PR #4600, opened 2020-12-21
Fix artifacts on macOS with transparent windows

### LINKED ISSUE #889: Graphical artifacts with window transparency

chrisduerr:
Due to the way macOS draws shadows for transparent windows, resizing
them will lead to text artifacts remaining present after a window has
been resized.

The `invalidateShadow` call is used whenever the opacity isn't `1.0` to
make sure these shadows are cleared before redrawing, so no artifacts
remain when resizing transparent windows.

Fixes #889.


ollyde:
How do we call this? I'm getting artifacts V2. @chrisduerr 

```
#[command]
pub async fn invalidate_window_shadow(window: tauri::Window) -> Result<(), String> {
    #[cfg(target_os = "macos")]
    {
        window.invalidate_shadow().map_err(|e| e.to_string())?;
    }

    Ok(())
}
```

no method named `invalidate_shadow` found for struct `tauri::Window` in the current scope
method not found in `Window`

PR #4602, opened 2020-12-21
Remove terminfo dependency

### LINKED ISSUE #4597: Remove `terminfo` dependency
Currently terminfo is used only to check if the Alacritty terminfo is
installed, so the fallback `xterm-256color` can be used if it is not.

However the terminfo crate does a few things and comes with some
unnecessary dependencies for Alacritty's usecase, like `phf`.

Since Alacritty's usage is so basic, it should be trivial to move this
terminfo functionality into Alacritty instead, reducing the dependency
on these external projects.

brightly-salty:
Fixes #4597

### DIFF
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -25,7 +25,7 @@ log = "0.4"
 unicode-width = "0.1"
 base64 = "0.12.0"
 regex-automata = "0.1.9"
-dirs = "3.0.1"
+dirs = "2.0.2"
 
 [target.'cfg(unix)'.dependencies]
 nix = "0.18.0"
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -24,8 +24,8 @@ mio-extras = "2"
 log = "0.4"
 unicode-width = "0.1"
 base64 = "0.12.0"
-terminfo = "0.7.1"
 regex-automata = "0.1.9"
+dirs = "3.0.1"
 
 [target.'cfg(unix)'.dependencies]
 nix = "0.18.0"
--- a/alacritty_terminal/src/tty/mod.rs
+++ b/alacritty_terminal/src/tty/mod.rs
@@ -1,6 +1,6 @@
 //! TTY related functionality.
 
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::{env, io};
 
 use crate::config::Config;
@@ -64,7 +64,8 @@ pub fn setup_env<C>(config: &Config<C>) {
     // Default to 'alacritty' terminfo if it is available, otherwise
     // default to 'xterm-256color'. May be overridden by user's config
     // below.
-    env::set_var("TERM", if is_terminfo_available() { "alacritty" } else { "xterm-256color" });
+    let terminfo = if terminfo_exists("alacritty") { "alacritty" } else { "xterm-256color" };
+    env::set_var("TERM", terminfo);
 
     // Advertise 24-bit color support.
     env::set_var("COLORTERM", "truecolor");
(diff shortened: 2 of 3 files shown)


alacritty_terminal/Cargo.toml
```diff
@@ -24,8 +24,8 @@ mio-extras = "2"
 log = "0.4"
 unicode-width = "0.1"
 base64 = "0.12.0"
-terminfo = "0.7.1"
 regex-automata = "0.1.9"
+dirs = "3.0.1"
```
chrisduerr, line 28: Please use the same version of `dirs` that is already in use by Alacritty.

alacritty_terminal/src/tty/mod.rs
```diff
@@ -81,3 +79,57 @@ pub fn setup_env<C>(config: &Config<C>) {
         env::set_var(key, value);
     }
 }
+
+fn is_alacritty_terminfo_available() -> bool {
+    use std::fs;
+    use std::path::{Path, PathBuf};
```
chrisduerr, line 85: Imports should be at the top of the file.

alacritty_terminal/src/tty/mod.rs
```diff
@@ -81,3 +79,57 @@ pub fn setup_env<C>(config: &Config<C>) {
         env::set_var(key, value);
     }
 }
+
+fn is_alacritty_terminfo_available() -> bool {
```
chrisduerr, line 83: I think this function would make more sense as `is_terminfo_available`.

alacritty_terminal/src/tty/mod.rs
```diff
@@ -81,3 +79,57 @@ pub fn setup_env<C>(config: &Config<C>) {
...
+    if let Ok(prefix) = env::var("PREFIX") {
+        let path = Path::new(&prefix);
+        search.push(path.join("etc/terminfo"));
+        search.push(path.join("lib/terminfo"));
+        search.push(path.join("share/terminfo"));
+    }
+    search.push("/etc/terminfo".into());
+    search.push("/lib/terminfo".into());
+    search.push("/usr/share/terminfo".into());
+    search.push("/boot/system/data/terminfo".into());
```
chrisduerr, line 110: I think an explicit `PathBuf::new` is better for initializing new values than `.into()`.

alacritty_terminal/src/tty/mod.rs
```diff
@@ -81,3 +79,57 @@ pub fn setup_env<C>(config: &Config<C>) {
...
+    for path in search {
+        if fs::metadata(&path).is_ok() {
```
chrisduerr, line 112: Should be a filter for the iterator.

alacritty_terminal/src/tty/mod.rs
```diff
@@ -81,3 +79,57 @@ pub fn setup_env<C>(config: &Config<C>) {
...
+
+            {
+                let mut path = path.clone();
+                path.push(format!("{:x}", first as usize));
+                path.push(name);
+
+                if fs::metadata(&path).is_ok() {
+                    return true;
+                }
+            }
```
chrisduerr, line 131: Why these scoped blocks?

alacritty_terminal/src/tty/mod.rs
```diff
@@ -81,3 +79,57 @@ pub fn setup_env<C>(config: &Config<C>) {
...
+    search.push("/etc/terminfo".into());
+    search.push("/lib/terminfo".into());
+    search.push("/usr/share/terminfo".into());
+    search.push("/boot/system/data/terminfo".into());
+    for path in search {
+        if fs::metadata(&path).is_ok() {
+            {
+                let mut path = path.clone();
+                path.push(first.to_string());
+                path.push(name);
```
chrisduerr, line 116: It might be possible to just change the way these paths are pushed to push the correct paths immediately using maybe a function as abstraction? That way we only need to do `paths.any(PathBuf::exists)` basically.

chrisduerr (changes requested):

chrisduerr (changes requested):
I don't think you quite understood my suggestions.

The `is_terminfo_available` function should be generic and take the terminfo name as parameter.

The iterator shouldn't take a function that just does the exact same thing the loop did previously, the paths should be pushed into it in their full form.

brightly-salty:
> I don't think you quite understood my suggestions.
> 
> The `is_terminfo_available` function should be generic and take the terminfo name as parameter.
> 
> The iterator shouldn't take a function that just does the exact same thing the loop did previously, the paths should be pushed into it in their full form.

I don't think I did either. I can easily make terminfo generic, but each path to be checked must be checked in two variants (`/a/alacritty` and `/61/alacritty`) (or that is at least the way terminfo does it) so we would have to duplicate each path individually. Are you sure this is what is wanted?

chrisduerr:
> so we would have to duplicate each path individually. Are you sure this is what is wanted?

That's why you just create a generic method to add paths which automatically appends everything necessary.

brightly-salty:
> That's why you just create a generic method to add paths which automatically appends everything necessary.

Something like `add_path(path: PathBuf, search: &mut Vec<PathBuf>, terminfo: &str)` that we call within `is_terminfo_available`? 

(Sorry if I'm needing too much mentoring throughout this, I just want to get this right. I love Alacritty and have been using it for the past year or two, by the way. Thanks for your work on it!)

chrisduerr:
> Something like add_path(path: PathBuf, search: &mut Vec<PathBuf>, terminfo: &str) that we call within is_terminfo_available?

Yes, it's probably a lot easier to just make it a struct to clean up the structure a bit.

brightly-salty:
> Yes, it's probably a lot easier to just make it a struct to clean up the structure a bit.

Is this what you were thinking? I can push a commit now if it is, otherwise I'll work on it some more?
```
struct TerminfoPaths {
    paths: Vec<PathBuf>,
    terminfo_string: String,
    terminfo_char: char,
}

impl TerminfoPaths {
    fn new(terminfo: &str) -> Self {
        let paths = Vec::new();
        let terminfo_string = terminfo.to_owned();
        let terminfo_char = terminfo.chars().next().unwrap();
        Self { paths, terminfo_string, terminfo_char }
    }

    fn create_path(&mut self, weird: bool, path: PathBuf) {
        let mut path_buf = path;
        if weird {
            path_buf.push(format!("{:x}", self.terminfo_char as usize));
        } else {
            path_buf.push(self.terminfo_char.to_string());
        }
        path_buf.push(self.terminfo_string.clone());
        if path_buf.exists() {
            self.paths.push(path_buf);
            return;
        }
    }

    fn add_path<P: Into<PathBuf>>(&mut self, into_path: P) {
        let path: PathBuf = into_path.into();
        if path.exists() {
            self.create_path(false, path.clone());
            self.create_path(true, path);
        }
    }

    fn load_paths(&mut self) {
        if let Some(dir) = env::var_os("TERMINFO") {
            self.add_path(dir);
        } else if let Some(mut home) = dirs::home_dir() {
            home.push(".terminfo");
            self.add_path(home);
        }
        if let Ok(dirs) = env::var("TERMINFO_DIRS") {
            for dir in dirs.split(':') {
                self.add_path(dir);
            }
        }
        if let Ok(prefix) = env::var("PREFIX") {
            let path = Path::new(&prefix);
            self.add_path(path.join("etc/terminfo"));
            self.add_path(path.join("lib/terminfo"));
            self.add_path(path.join("share/terminfo"));
        }
        self.add_path("/etc/terminfo");
        self.add_path("/lib/terminfo");
        self.add_path("/usr/share/terminfo");
        self.add_path("/boot/system/data/terminfo");
    }

    fn path_exists(&self) -> bool {
        !self.paths.is_empty()
    }
}
```

chrisduerr:
It needs work, but something like that maybe.

chrisduerr:
@brightly-salty I'll refactor the code a bit myself, I think that's the easiest solution. Then you can check out if that makes sense to you.

alacritty_terminal/src/tty/mod.rs
```diff
@@ -81,3 +78,48 @@ pub fn setup_env<C>(config: &Config<C>) {
...
+    // Return true if the terminfo file exists at the specified location.
+    macro_rules! check_path {
+        ($path:expr) => {
+            if $path.join(first).join(terminfo).exists()
+                || $path.join(&first_hex).join(terminfo).exists()
+            {
+                return true;
+            }
+        };
+    }
```
chrisduerr, line 97: I've decided to go with a macro here, since this completely avoids the allocation for the `Vec`, which is entirely unnecessary here.

So we can just easily check each path immediately and abort without pushing a bunch of useless paths to a vec that we never read anyways.

chrisduerr (approved):

brightly-salty (commented):
Looks good to me! That makes it a lot simpler.

PR #4620, opened 2020-12-29
Update Alacritty's description

chrisduerr:
Alacritty's description in the readme and manpage has always been a bit
overly aggressive in its marketing. This new updated text should more
accurately describe Alacritty's current state without pointlessly
agitating people.


README.md
```diff
@@ -1,28 +1,24 @@
...
 </p>
 
-Alacritty is the fastest terminal emulator in existence. Using the GPU for
-rendering enables optimizations that simply aren't possible without it.
-Alacritty currently supports macOS, Linux, BSD, and Windows.
+<h1 align="center">Alacritty - A fast, cross-platform, OpenGL terminal emulator</h1>
 
 <p align="center">
   <img width="600"
        alt="Alacritty running vim inside tmux"
```
cole-h, line 9: Really minor nit, but is this still "vim inside tmux"? I guess you could be hiding the tab bar thing, but it doesn't seem necessary to call out the tmux part when it's not even visible.

chrisduerr: It's not.

chrisduerr: https://github.com/alacritty/alacritty/pull/4623 should fix it.

PR #4660, opened 2021-01-04
Add colored underline support

### LINKED ISSUE #4142: Support for setting individual color for underline (<ESC>[58...m)
It would be nice if alacritty learns how to set the color of the underline independently of the text color. This is in particularly useful for marking spelling mistakes in a (console) text editor and still keep other syntax highlighting intact.

According to Wikipedia [0] the (semi-standard) CSI / SGR / ANSI escape code / whatever sequence is `<ESC>[58...m`

Other terminals which support this are for example Kitty, VTE, mintty, and iTerm2.

[0] https://en.wikipedia.org/wiki/Control_Sequence_Introducer

Jesse-Bakker:
Fixes #4142

Support colored underline, according to [kitty's protocol extension](https://sw.kovidgoyal.net/kitty/protocol-extensions.html#colored-and-styled-underlines). When a different underline color is specified, it stays the same under color inversion when selecting, but it follows the foreground color when it is not.


chrisduerr (changes requested):
Underline color is basically never needed, it should be allocated dynamically.

Jesse-Bakker:
It is now allocated dynamically

chrisduerr:
No it isn't.

alacritty/src/renderer/rects.rs
```diff
@@ -174,11 +174,17 @@ impl RenderLines {
             end.col += 1;
         }
 
+        let color = if flag & Flags::UNDERLINE == Flags::UNDERLINE
+            || flag & Flags::DOUBLE_UNDERLINE == Flags::DOUBLE_UNDERLINE
```
chrisduerr, line 178: You should use `flag.contains(Flags::STUFF)`.

alacritty_terminal/src/term/cell.rs
```diff
@@ -92,6 +93,20 @@ impl Cell {
         self.extra.get_or_insert_with(Default::default).zerowidth.push(c);
     }
 
+    #[inline]
+    pub fn ul(&self) -> Option<Color> {
+        self.extra.as_ref().map(|extra| extra.ul)?
```
chrisduerr, line 98: Looks like you want to use .and_then?

alacritty_terminal/src/term/cell.rs
```diff
@@ -92,6 +93,20 @@ impl Cell {
         self.extra.get_or_insert_with(Default::default).zerowidth.push(c);
     }
 
+    #[inline]
+    pub fn ul(&self) -> Option<Color> {
```
chrisduerr, line 97: Please don't use abbreviations like `ul` and instead spell it out. This applies to all changes in this PR.

alacritty_terminal/src/term/render.rs
```diff
@@ -211,6 +212,12 @@ impl RenderableCell {
             Self::compute_bg_alpha(cell.bg)
         };
 
+        let mut ul_rgb = if let Some(ul) = cell.ul() {
+            Self::compute_fg_rgb(content.config, &content.term.colors, ul, cell.flags)
+        } else {
+            fg_rgb
+        };
```
chrisduerr, line 219: .map().unwrap_or() seems more appropriate here.

alacritty_terminal/src/term/render.rs
```diff
@@ -223,6 +230,10 @@ impl RenderableCell {
             bg_rgb = config_bg.color(fg_rgb, bg_rgb);
             fg_rgb = selected_fg;
 
+            if cell.ul().is_none() {
+                ul_rgb = fg_rgb;
+            }
```
chrisduerr, line 235: Why not just put the entire underline logic below these conditionals? That way it shouldn't be necessary to mutate `ul_rgb` after the fact.

chrisduerr (changes requested):
Have you benchmarked this with vtebench? If so I'd be interested in seeing the results compared to master.

alacritty_terminal/tests/ref/colored_underline/alacritty.recording
```diff
@@ -0,0 +1,3 @@
+[1m[7m%[27m[1m[0m                                                                                [0m[27m[24m[Jpohjala% [K[?2004h[0m[27m[24m[Jpohjala% [7mecho "\u1b[58;5;2;38;5;3mNO\u1b[4mUND\u1b[58;5;6mERLI\u1b[59mNE\u1b[0mD"[27m[72D[27me[27mc[27mh[27mo[27m [27m"[27m\[27mu[27m1[27mb[27m[[27m5[27m8[27m;[27m5[27m;[27m2[27m;[27m3[27m8[27m;[27m5[27m;[27m3[27mm[27mN[27mO[27m\[27mu[27m1[27mb[27m[[27m4[27mm[27mU[27mN[27mD[27m\[27mu[27m1[27mb[27m[[27m5[27m8[27m;[27m5[27m;[27m6[27mm[27mE[27mR[27mL[27mI[27m\[27mu[27m1[27mb[27m[[27m5[27m9[27mm[27mN[27mE[27m\[27mu[27m1[27mb[27m[[27m0[27mm[27mD[27m"[?2004l
```
chrisduerr, line 1: This ref test is rather big and I see no reason why it couldn't be combined with the underline ref test. Would you mind re-recording it to include the tests of both so the size is a bit smaller? Minimizing window size will also help with reducing the output grid.json a bit.

alacritty_terminal/src/term/cell.rs
Jesse-Bakker: I followed the convention set by `fg` and `bg` here. I guess `ul` is less obious though

chrisduerr: Yeah, fg/bg could be renamed too, but they're **really** obvious. Nobody knows what `ul` is. Generally I would encourage using the full length as much as possible.

Jesse-Bakker: Yes, I agree

Jesse-Bakker:
Benchmark:
summary

chrisduerr:
> Benchmark shows it's a bit slower, as expected:

It shouldn't be in any of these benchmarks.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.8.0-dev
 
+### Added
+
+- Support for colored underlines
```
chrisduerr, line 12: I think this entry isn't entirely clear. You can already have colored underlines in Alacritty, you just can't change it from the text color.

Maybe something like this is a little less confusing:


```
- Underline color escape (`CSI 58 : 2 : Ps : Ps : Ps m`/`CSI 58 : 5 : Ps m`)
```

alacritty/src/renderer/rects.rs
```diff
@@ -174,11 +174,15 @@ impl RenderLines {
             end.col += 1;
         }
 
+        let color = if flag.contains(Flags::UNDERLINE) || flag.contains(Flags::DOUBLE_UNDERLINE) {
+            cell.underline_color
+        } else {
+            cell.fg
+        };
```
chrisduerr, line 181: Is this done to prevent strikethrough from using this color? That seems quite unfortunate to me, are you aware of why this escape sequence doesn't apply to all lines?

I suppose it's because both strikethrough and underline might be active at the same time?

Generally this just seems concerning because it might cause further escape sequence to complicate things even further, adding yet another color for strikethrough, but I suppose while a red underline for typos is quite common, the strikethrough might not ever have a reason to be a different color than the text itself.

Generally it might be a good idea to add a comment to this stating that strikethrough always uses the cell's color.

alacritty_terminal/src/term/cell.rs
```diff
@@ -92,6 +93,20 @@ impl Cell {
         self.extra.get_or_insert_with(Default::default).zerowidth.push(c);
     }
 
+    #[inline]
+    pub fn underline_color(&self) -> Option<Color> {
+        self.extra.as_ref().and_then(|extra| extra.underline_color)
+    }
```
chrisduerr, line 99: I'm not sure how much sense it makes to add a separate method for every single field in `extra`. Generally accessing this data is likely very performance sensitive, so small things can make big differences.

Having a single `pub fn extra(&self) -> Option<&CellExtra>` might make sense, because at that point the consumer of that method can probably do the exact same things as before but now the check for `None` on `extra` only needs to happen once, not multiple times.

alacritty_terminal/src/term/cell.rs
```diff
@@ -92,6 +93,20 @@ impl Cell {
...
+
+    #[inline]
+    pub fn set_underline_color(&mut self, color: Option<Color>) {
```
chrisduerr, line 102: We might want to document how this function works. Passing `None` to reset might seem obvious once you look into it, but it could confuse a first reader.

alacritty_terminal/src/term/mod.rs
```diff
@@ -995,6 +996,7 @@ impl<T> Term<T> {
         cursor_cell.c = c;
         cursor_cell.fg = fg;
         cursor_cell.bg = bg;
+        cursor_cell.set_underline_color(underline_color);
```
chrisduerr, line 999: I haven't benchmarked things yet, but this might have a big impact on performance.

The `write_at_cursor` function is **extremely** hot, so even a simple branching like in the `set_underline_color` might be problematic.

I'll just leave this comment here for now, but this might be a problem. Feel free to ignore it until everything else is done and I have run my own benchmarks though.

alacritty_terminal/tests/ref/underline/alacritty.recording
```diff
@@ -1,13 +1,15 @@
...
+[?2004l
[4;4:2mUNDERLINED[0m
+]0;jesse@pohjala:~/dev/oss/alacritty[?2004h$ [7mecho -e "\e[4:2;4:1mUNDERLINED\e[0m"[27m
[C[Cecho -e "\e[4:2;4:1mUNDERLINED\e[0m"
+[?2004l
[4:2;4:1mUNDERLINED[0m
+]0;jesse@pohjala:~/dev/oss/alacritty[?2004h$ [7mecho -e "\e[58;5;2;38;5;3mNO\e[4mUND\e[58;5;6mERLI\e[59mNE\e[0m"[27m
[C[Cecho -e "\e[58;5;2;38;5;3mNO\e[4mUND\e[58;5;6mERLI\e[59mNE\e[0m"
+[?2004l
[58;5;2;38;5;3mNO[4mUND[58;5;6mERLI[59mNE[0m
```
chrisduerr, line 14: You only ever use the semicolon separated form here, however that's technically the legacy version and the preferred one should be the colon separated one.

So at the very least we should have a test that verifies both the `5` and `2` forms work properly with the colon separated one.

alacritty_terminal/tests/ref/underline/size.json
```diff
@@ -1 +1 @@
-{"width":939.0,"height":503.0,"cell_width":8.0,"cell_height":16.0,"padding_x":5.0,"padding_y":3.0,"cols":116,"screen_lines":31}
+{"width":620.0,"height":500.0,"cell_width":8.0,"cell_height":20.0,"padding_x":0.0,"padding_y":0.0,"screen_lines":25,"cols":77}
```
chrisduerr, line 1: The size is small and stuff is still rendered nicely without linebreaks or anything like that. That's exactly how it should be, thanks.

chrisduerr (changes requested):
Since this adds a new escape sequence, it should be documented in our `docs/escape_support.md` file.

alacritty/src/renderer/rects.rs
Jesse-Bakker: Yes, the underline color escape sequence should only affect (special) underline color (set with the `<ESC>[4..m` escape sequences), as in KiTTY, iTerm2, VTE.

alacritty_terminal/src/term/cell.rs
Jesse-Bakker: That could certainly be done. How would you imagine mutating the extra data? There could be a `pub fn extra_mut(&mut self) -> Option<&mut CellExtra>`, but that can only mutate existing extra data. A `pub fn set_extra(&mut self, extra: CellExtra)` would do unnecessary moves if there is already data there. Maybe we should then have both? Or `pub fn ensure_extra(&mut self)` which does the allocation?

chrisduerr: Currently I don't think there's any necessity to change tho way the data is mutated. The `push_zerowidth` function is in a different codepath that wouldn't have access to the mutable borrow of `extra_mut` anyways. So having two separate functions which both have to initialize it if it doesn't exist yet should be fine.

While the read `None` path is hit very often, the `write` path is also only hit very rarely. So you would only benefit when trying to write *multiple* fields in the extra struct using a mutable reference and that's so rare that we don't need to worry about optimizing in that regard yet.

chrisduerr:
Performance results from my benchmarks:

tmp

There seems to be a consistent upwards trend with this PR. I'm not sure what the best way to solve this is, but I personally don't think adding colored underlines justifies a performance regression. I'd say usually they're rather useless.

alacritty_terminal/src/term/mod.rs
chrisduerr: I'd appreciate if you could look into optimizing this part of the code. It's most likely where the performance regression is coming from and this is a rather unfortunate implementation.

Ideally this function shouldn't have to do anything different if the underline isn't colored, but that might be tough to realize.

I could look into this myself, but since I personally don't care about this feature that's going to be at the bottom of my priority list for now.

chrisduerr (changes requested):

DanielVoogsgerd:
Just wanted to leave a comment that there is a demand for this. For example, I use it for diagnostics in my editor of choice, but many use cases can be thought of.

If we come across this PR and also like this feature, I'd recommend upvoting the original comment by @Jesse-Bakker 

My apologies if this comes across as spamming, just wanting to tell people that upvoting on these kinds of issues matters.

alacritty_terminal/src/term/mod.rs
ayosec: If performance is more important than memory maybe it is better to put the `underline_color` field in `Cell` instead of `CellExtra`.

I did a quick test in https://github.com/ayosec/alacritty/commit/bbf64848520c7b3b53614869f88f3195ca955c51 and then launched `vtebench`. In my machine, the performance after the change is better is most cases:

output

chrisduerr:
> just wanting to tell people that upvoting on these kinds of issues matters.

It doesn't. Neither does endless commenting. Please keep things productive, **especially** on PRs and not just issues.

alacritty_terminal/src/term/mod.rs
Jesse-Bakker: I got rid of 1 branch in the common case: there used to be a check for the existence of `extra` when setting underline_color to `None`, but as `extra` is dropped in `write_at_cursor`, we know this will be `None`, so we can get rid of the branch.

chrisduerr: Well is that enough to match master's performance?

Jesse-Bakker: According to my benchmarks, which are not very consistent as you mentioned, it is only very slightly slower than master. As far as I can tell, this will only add one branch (I tried optimizing it further to only have one branch in the rust code, but the generated asm did not change), which is probably as fast as it can get.

chrisduerr: > which is probably as fast as it can get

"probably as fast as it can get" is not very reassuring. It doesn't just matter how often it branches, it also matters how it branches and if the assembly it generates makes sense or not.

That information is necessary to know if it's possible to store this in the dynamic data, or if maybe it has to be stored in every cell and bloat Alacritty's memory size.

There are also various tips on improving benchmarking consistency in here: https://llvm.org/docs/Benchmarking.html

Generating some alternative benchmarks for faster validation and running everything through `perf` to verify its performance probably makes sense.

nixpulvis:
I just started looking at this PR today. First of all, I'm excited about this personally for my `vim` configuration's spell checker. I may find other uses afterwards. I know the linked kitty extension talks about other "styles" of underline as well, but I'm personally less excited about that, so this seems like a great place to start.

Anyway, I think I've found a small bug in my initial testing. At least a behavior that works differently than kitty's. Running these two commands in sequence should probably result in the first `TEST` underlined in red, followed by the next `TEST` underlined in the default foreground color.

```sh
echo -e "\e[58:5:1;4mTEST\e[0m"
echo -e "\e[58:5;4mTEST\e[0m"
```

I might also recommend that the ref test be updated. It would help me personally if instead of `NOUNDERLINE` it was something like `NOREDGREENBLUERESET` as well, but this is just a tiny suggestion if you want, while you're editing the ref test anyway.

chrisduerr:
> Is Su for styled underlines in general, or would it be OK to advertise this for just the colored features of CSI 58 ... m and CSI 59 ... m?

I'm not actually sure if anyone uses `Su`, but vim does use `Smulx` I believe which Alacritty already exports.

bluz71:
>  I'm not sure what the best way to solve this is, but I personally don't think adding colored underlines justifies a performance regression. I'd say usually they're rather useless.

The point of view with respect to coloured underlines (and eventually coloured undercurls) depends on one's workflow.

One man's useless feature can be another's most important feature.

For me this is my absolute **most desired** Alacritty enhancement, one that I would willing trade a few CPU cycles for. So I am saddened to hear that it may be turfed. I genuinely hope that is not the case.

Coloured underlines can be handy for visualising lint and spelling errors in the Vim and Neovim editors. Language Server Protocol is quickly gaining traction with the best servers providing diagnostic support (for example the TypeScript language server). Vim and Neovim, via LSP, can highlight the actual error position via coloured underline or undercurls. This brings IDE like error visualisation to standard text mode editors which I find very cool.

This screenshot highlights a usage of coloured undercurl to highlight spelling errors (via GNOME Terminal which does support coloured undercurls):

88470712-33f9f880-cf43-11ea-9bc8-553ccd446614

aignas (commented):
It seems that the `Term::new` contains access to the `config` passed from the `src/main.rs`.

It would be great to have a solution, where people can opt into the underline with color if they are willing to sacrifice those extra CPU cycles. This could at least give the people feature they want, whilst the feature itself can be optimized later.

nixpulvis:
> It would be great to have a solution, where people can opt into the underline with color if they are willing to sacrifice those extra CPU cycles. This could at least give the people feature they want, whilst the feature itself can be optimized later.

This kind of suggestion comes up a lot, however it misses two key points.

1. Just because something is behind a configuration flag doesn't remove it from the development burden.
2. There's no guarantee things will be optimized later.

Maintaining Alacritty is somewhat of a challenge in balancing features, performance, and legacy compatibility. Hopefully you'll understand why we'd rather get things right the first time here.

Without having run the benchmarks myself, just going off what's been posted here, I wouldn't think that it's very far off from being performant enough. But I'm not the final say on any of this.

chrisduerr:
> It would be great to have a solution, where people can opt into the underline with color if they are willing to sacrifice those extra CPU cycles.

Alacritty will not add configuration options just to make sure it stays technically "fast". Either everything Alacritty does is fast, or nothing is.

> Without having run the benchmarks myself, just going off what's been posted here, I wouldn't think that it's very far off from being performant enough. But I'm not the final say on any of this.

I'm confident this can be made reasonably fast. The question is just how. In a worst-case scenario, it could be considered to just grow the cell size and include this directly. As mentioned in the original issue, that's something we could discuss since we've made space in the cell size partially to allow for additions like this.

But I think the current code still has room for improvement. Only because looking at `perf` output and shuffling things around can be a lot of work, doesn't mean that things should just be tossed in and merged. It just takes some time.

aignas:
> > It would be great to have a solution, where people can opt into the underline with color if they are willing to sacrifice those extra CPU cycles. This could at least give the people feature they want, whilst the feature itself can be optimized later.
> 
> This kind of suggestion comes up a lot, however it misses two key points.
> 
>  1. Just because something is behind a configuration flag doesn't remove it from the development burden. In fact configurations should generally be viewed as a larger burden in general, as they need to be respected if a user specifies them going forward, and deprecated when/if they are removed.
> 
>  2. There's no guarantee things will be optimized later.
> 
> Maintaining Alacritty is somewhat of a challenge in balancing features, performance, and legacy compatibility. Hopefully you'll understand why we'd rather get things right the first time here.

I agree with the points above and completely respect having this stance and just wanted to have this explicit here. Software is sometimes about the trade offs and wanted to see what trade offs you are leaning towards here. Thanks for the response. Since it is not my call to make, I'll gladly accept any outcome that you chose - at the end of the day, we have a terminal that renders fonts really well, is fast and works really well.

> Without having run the benchmarks myself, just going off what's been posted here, I wouldn't think that it's very far off from being performant enough. But I'm not the final say on any of this.

I have run the benchmarks on my ThinkPad T420s (which is almost 10 years old now) and the performance difference is within the error bars. That said, I may have not followed the LLVM benchmarking guideline to the letter, so my results may be skewed.

bluz71:
I built and tested @Jesse-Bakker's PR and I absolutely love it.

Current Alacritty:

Default

And the colored underline branch:

Colored

The above are spelling errors in Neovim. Same can apply to LSP diagnostics, which is actually my prime motivator for this wanting this PR.

I know performance is the most important attribute for Alacritty, but studying @chrisduerr's performance graph (from a few posts up) is the performance loss of this branch that bad? Looks about 1-2%'ish (and the *missing* tests were actually faster in the colored-underline branch).

Performance

Anyway, I hope this PR does not rust away into nothingness, it's certainly very useful for certain folks such as myself. Actually it is my most desired want for Alacritty.

Cheers.

Israel-Laguan (approved):
Just fix conflicts, besides that looks good to me

bluz71:
Question, if the conflicts are resolved then this will be merged to master? Or does it still require others to approve?

Just wanting to know if I should raise my hopes or not.

This is my number one most wanted feature, so I have a vested interest.

Israel-Laguan:
@bluz71 I would say after rebase and solve merge conflicts a regression test and @chrisduerr can consider merge.

chrisduerr:
@Israel-Laguan Please don't speak on the behalf of others when you have absolutely no clue of anything.

> if the conflicts are resolved then this will be merged to master

No.

gshpychka:
> @Israel-Laguan Please don't speak on the behalf of others when you have absolutely no clue of anything.

That's pretty rude.

Jesse-Bakker:
@Israel-Laguan The reason this PR isn't merged are not merge conflicts. When/if I have time and motivation to fix the actual issues, they will be solved. You are of course welcome to do some profiling and optimization yourself to move this along, but any comments that do not try to actively help just cause unnecessary burden on the maintainers and PR authors.

kchibisov:
Implemented in https://github.com/alacritty/alacritty/pull/5935.

PR #4671, opened 2021-01-07
Remove the GitHub pull request template

chrisduerr:
The template itself never provided much of a benefit and just causes
additional work for every single PR that is opened. That makes it easier
to just remind people after the PR was created than putting it in the
description so contributors always have to remove it.


cole-h:
FWIW, it is possible to surround the contents with HTML comment tags. Then it won't appear in the PR body, but still be visible upon opening a PR and "useful" for guidance.

chrisduerr:
@cole-h That's a good tip, thanks. But I think at least for me personally that would still make things unnecessarily complicated. It just becomes difficult to edit things because your text and the pull request template's text gets all mashed together, so you need to figure out where to add new text and remove existing stuff to get a nice formatting.

Though should it turn out that a lot of people start forgetting changelog entries for example, it might be worth exploring that in the future.

PR #4675, opened 2021-01-08
Add Oracle Linux 7 in install doc

rohchakr:
I have successfully built alacritty from source in my Oracle Linux 7 machine by following RHEL 7 instructions

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -104,7 +104,7 @@ missing, please open an issue.
 dnf install cmake freetype-devel fontconfig-devel libxcb-devel g++
 ```
 
-#### CentOS/RHEL 7
+#### CentOS/RHEL 7/Oracle Linux 7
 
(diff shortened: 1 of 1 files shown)


chrisduerr (commented):
Oracle Linux is just RHEL rebranded, right? I'm not sure there's any point in listing it if it's the same thing with a different name? I'd expect users to be aware of this?

Oracle Linux seems somewhat obscure, so it's a bit different than RHEL/CentOS or Debian/Ubuntu, which are very popular variations. It's a bit like listing KDE Neon/PopOS/Kubuntu/... all separately, that just be a bit repetitive?

rohchakr:
Thanks @chrisduerr 

Yes, I agree it would be repetitive and cluttered to cover obscure diastros which are not very different from their base diastros.

Certainly, Oracle Linux is built on top of RHEL and I am too naive to talk about the differences (never used RHEL).

Closing this PR

PR #4677, opened 2021-01-08
Fix jumping between matches in reverse vi search

### LINKED ISSUE #4676: Cannot jump between matches in reverse vi mode search
When trying to jump between matches in reverse search with vi mode, the vi
cursor gets stuck on matches without being able to jump between them.

This regression has been introduced in 08e111dfae15d2428d87c3a80c68d858b144d287.

chrisduerr:
Fixes #4676.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.8.0-dev
 
+## 0.7.1
+
+### Fixed
+
+- Jumping between matches in backward vi search
+
 ## 0.7.0
 
 ### Added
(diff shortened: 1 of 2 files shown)


cole-h:
Despite adding a 0.7.1 section to the changelog, it doesn't seem like a new 0.7.1 release was tagged. Unless this was on purpose and you're waiting to fix a few more bugs before releasing it?

chrisduerr:
> Despite adding a 0.7.1 section to the changelog, it doesn't seem like a new 0.7.1 release was tagged. Unless this was on purpose and you're waiting to fix a few more bugs before releasing it?

I've decided to immediately release a new version, to maybe get it out before major distributions package 0.7.0. However I'm still waiting for a reply to #4678 to see if that issue also requires a bugfix.

You're also on X11, right? Have you run into any issues?

cole-h:
On Wayland, actually. I haven't been "drinking from the tap" as of late, but I haven't run into any problems recently.

chrisduerr:
I'll wait another couple of hours, but it probably makes sense to push another version rather than waiting. I doubt this X issue is something that can be easily fixed. If there are more issues, we can always release a 0.7.2.

PR #4683, opened 2021-01-09
Cross compile macOS app for arm64 and x86_64

bouk:
Here I'm adding cross compilation for arm64 and x86_64, and combining them into a [universal binary](https://developer.apple.com/documentation/xcode/building_a_universal_macos_binary) with lipo.

I also added installation of the targets in the release workflow. With these changes the app will run natively on both Intel and M1 Macs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,11 +25,11 @@ jobs:
         run: |
           rustup default 1.43.1
           cargo test
-  build-macos-universal:
+  build-macos-apple-silicon:
     runs-on: macos-11.0
     steps:
       - uses: actions/checkout@v2
-      - name: Install targets
-        run: rustup update && rustup target add x86_64-apple-darwin aarch64-apple-darwin
-      - name: Make App
-        run: make universal
+      - name: Install target
+        run: rustup update && rustup target add aarch64-apple-darwin
+      - name: Ensure it builds for Apple Silicon
+        run: MACOSX_DEPLOYMENT_TARGET="10.11" cargo build --release --target=aarch64-apple-darwin
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,3 +25,11 @@ jobs:
         run: |
           rustup default 1.43.1
           cargo test
+  build-macos-universal:
+    runs-on: macos-11.0
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install targets
+        run: rustup update && rustup target add x86_64-apple-darwin aarch64-apple-darwin
+      - name: Make App
+        run: make universal
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,5 +31,5 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install target
         run: rustup update && rustup target add aarch64-apple-darwin
-      - name: Ensure it builds for Apple Silicon
+      - name: Check build
         run: cargo check --target=aarch64-apple-darwin
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,11 +25,11 @@ jobs:
         run: |
           rustup default 1.43.1
           cargo test
-  build-macos-apple-silicon:
+  check-macos-apple-silicon:
     runs-on: macos-11.0
     steps:
       - uses: actions/checkout@v2
       - name: Install target
         run: rustup update && rustup target add aarch64-apple-darwin
       - name: Ensure it builds for Apple Silicon
-        run: MACOSX_DEPLOYMENT_TARGET="10.11" cargo build --release --target=aarch64-apple-darwin
+        run: cargo check --target=aarch64-apple-darwin
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
         run: |
           rustup default 1.43.1
           cargo test
-  check-macos-apple-silicon:
+  check-macos-arm:
     runs-on: macos-11.0
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,12 +10,12 @@ env:
 
 jobs:
   macos:
-    runs-on: macos-latest
+    runs-on: macos-11.0
 
     steps:
       - uses: actions/checkout@v2
       - name: Install targets
-        run: rustup target add x86_64-apple-darwin aarch64-apple-darwin
+        run: rustup update && rustup target add x86_64-apple-darwin aarch64-apple-darwin
       - name: Test
         run: cargo test --release
       - name: Make App
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Test
         run: cargo test --release
       - name: Make App
-        run: make dmg "MULTI_ARCH=x86_64 aarch64"
+        run: make universal
       - name: Upload Application
         run: |
           mv ./target/release/osx/Alacritty.dmg ./Alacritty-${GITHUB_REF##*/}.dmg
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Test
         run: cargo test --release
       - name: Make App
-        run: make dmg
+        run: make dmg "MULTI_ARCH=x86_64 aarch64"
       - name: Upload Application
         run: |
           mv ./target/release/osx/Alacritty.dmg ./Alacritty-${GITHUB_REF##*/}.dmg
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,6 +14,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Install targets
+        run: rustup target add x86_64-apple-darwin aarch64-apple-darwin
       - name: Test
         run: cargo test --release
       - name: Make App
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Test
         run: cargo test --release
       - name: Make App
-        run: make universal
+        run: make dmg-universal
       - name: Upload Application
         run: |
           mv ./target/release/osx/Alacritty.dmg ./Alacritty-${GITHUB_REF##*/}.dmg
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,7 @@ $(APP_NAME)-%: $(TARGET)-%
 	@echo "Created '$(APP_NAME)' in '$(APP_DIR)'"
 
 dmg: | $(DMG_NAME)-native ## Pack Alacritty.app into .dmg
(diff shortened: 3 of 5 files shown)


chrisduerr (commented):
I have no way to test cross compilation on macOS.

What systems did you test this on? Have you actually confirmed that the cross compilation works, or just tested the compilation on the targets themselves?

bouk:
I'm running this on my M1 MacBook Air. I can run the one binary that's created with these commands (arch allows you to choose with architecture to run the binary with):

```bash
$ rustup target add x86_64-apple-darwin aarch64-apple-darwin
$ make dmg
$ arch -x86_64 target/release/osx/Alacritty.app/Contents/MacOS/alacritty &
$ arch -arm64 target/release/osx/Alacritty.app/Contents/MacOS/alacritty &
```

And confirm from Activity Monitor that I can run alacritty in both Architectures (the 'Intel' architecture is running through 
Rosetta 2 on my machine):

image

Here's the resulting package if anyone else wants to try running it on their (non-arm64) machine:
[Alacritty.dmg.zip](https://github.com/alacritty/alacritty/files/5792461/Alacritty.dmg.zip)

I have an Intel MacBook here that I'll try running the result on later.

chrisduerr:
> I have an Intel MacBook here that I'll try running the result on later. Let me know if you need any more info.

The interesting part would be verifying that something compiled on Intel works on ARM and Intel, since that's the actual architecture our CI is running on. So we should make sure the existing stuff keeps working and the ARM cross-compilation from Intel works.

bouk:
I [manually ran the workflow](https://github.com/bouk/alacritty/actions/runs/475773354) and confirmed that the resulting dmg (which is built on x86_64) is runnable on my M1 MacBook 🎉

I did have to make one small change to use the [macos-11.0](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11.0-Readme.md) builder and to update Rust to the latest version, but then the dmg is built correctly: https://github.com/bouk/alacritty/suites/1801767281/artifacts/34555560

So this PR is good to go I think.

chrisduerr:
> I did have to make one small change to use the macos-11.0 builder and to update Rust to the latest version

Having to update the rust version in the CI job seems rather unfortunate. It might make sense to wait until the github pipeline is updated to use the latest version. I haven't checked myself but I'd assume you had to update all the way to 1.49 for it to work? It seems like the current macOS runner uses Rust 1.48.0, so I'd assume they update quite frequently (see [docs](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11.0-Readme.md)).

> So this PR is good to go I think, although maybe we should add macos-11.0 to CI

I don't see any reason to build on multiple macOS versions, however I'm perfectly fine with updating to the latest available version (as long as it works of course).

bouk:
Yes we need Rust 1.49 for arm64 support, which `rustup update` gets us.

If we don't want to run it on both, shall I then change CI to use macos-11.0 also? Or is it fine like this?

Makefile
```diff
@@ -33,14 +35,18 @@ binary: | $(TARGET) ## Build release binary with cargo
 $(TARGET):
 	MACOSX_DEPLOYMENT_TARGET="10.11" cargo build --release
 
+target/%/release/$(TARGET): phony_explicit
+	MACOSX_DEPLOYMENT_TARGET="10.11" cargo build --release --target=$*
```
chrisduerr, line 39: This will cause the build to fail unless the user building it has both targets installed, correct? I'm not sure that's good, I think users should probably be able to build a binary that works just on their system if they don't intend to ship it anywhere. Otherwise we'd have to document this in our installation instructions.

chrisduerr (changes requested):

.github/workflows/release.yml
```diff
@@ -10,10 +10,12 @@ env:
 
 jobs:
   macos:
-    runs-on: macos-latest
+    runs-on: macos-11.0
 
     steps:
       - uses: actions/checkout@v2
+      - name: Install targets
+        run: rustup update && rustup target add x86_64-apple-darwin aarch64-apple-darwin
```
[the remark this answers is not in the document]
bouk: Unfortunately I don't think there's an apple ARM builder available in github actions, maybe they will provide it at some point. Although just building it in CI would be good in itself.

Makefile
bouk: That's correct, I'd be open to suggestions on how we can write this Makefile so it supports choosing whether you just want a native or a cross-platform build.

.github/workflows/release.yml
```diff
@@ -10,10 +10,12 @@ env:
 
 jobs:
   macos:
-    runs-on: macos-latest
+    runs-on: macos-11.0
```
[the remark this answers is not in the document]
bouk: I'll attempt building it with macos-latest to see if that works.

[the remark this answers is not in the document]
bouk: This fails unfortunately: https://github.com/bouk/alacritty/runs/1677172051?check_suite_focus=true

Seems it needs a newer version of the SDK

Makefile
bouk: I've updated the Makefile to add a condition that only builds for multiple architectures if you set the MULTI_ARCH option

.github/workflows/release.yml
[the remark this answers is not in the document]
chrisduerr: I'm talking about cross-testing from one platform to another. Just like you cross build the binary.

Makefile
chrisduerr: Maybe just a `make universal` instead of `make app`? Would probably be easier.

bouk: Modified the makefile :+1:

guyeu:
Is there anything preventing to merge this PR? This is very import to use terminal emulator without Rosetta2 because of homebrew cannot work correctly under Rosetta2.

Byron:
@guyeu Please note that [this comment](https://github.com/alacritty/alacritty/pull/4683#issuecomment-757450192) provides a link to a prebuilt universal binary.
If you need an ARM build, cloning this repository and running `make binary app` usually does the trick provided the Rust toolchain is already ARM itself. You would then find the application package in `target/release/osx`. Hope that helps.

zlj-zz:
I try run it in my computer, it's greate.
         -------------------------------
OS: macOS 11.1 20C69 arm64
Host: MacBookPro17,1
Kernel: 20.2.0
Shell: /bin/zsh 5.8

bouk:
@chrisduerr anything that is preventing this from being merged?

chrisduerr:
Yes, everything I've mentioned in my review.

bouk:
I added a universal makefile target, and added the universal build to CI. That addresses your comments

bouk:
I also rebased on master

bouk:
It seems GitHub is having issues with the macos build

.github/workflows/ci.yml
```diff
@@ -25,3 +25,11 @@ jobs:
...
           rustup default 1.43.1
           cargo test
+  build-macos-universal:
+    runs-on: macos-11.0
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install targets
+        run: rustup update && rustup target add x86_64-apple-darwin aarch64-apple-darwin
+      - name: Make App
+        run: make universal
```
chrisduerr, line 35: We don't need to build a universal image here, all we need to do is test the arm code. We're already testing x86, there's no point in doing it twice.

.github/workflows/release.yml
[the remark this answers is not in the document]
chrisduerr: This is not resolved, the build process during release still does not test all targets.

Makefile
```diff
@@ -44,23 +50,25 @@ $(APP_NAME): $(TARGET)
...
 
 install: $(DMG_NAME) ## Mount disk image
 	@open $(DMG_DIR)/$(DMG_NAME)
 
-.PHONY: app binary clean dmg install $(TARGET)
+.PHONY: app binary clean dmg install $(TARGET)-native $(TARGET)-universal
 
 clean: ## Remove all artifacts
-	-rm -rf $(APP_DIR)
+	-rm -rf target
```
chrisduerr, line 74: We should probably just run `cargo clean` rather than nuking the target directory. That also destroys everything inside it but it correctly follows overrides to the target directory. It might be located somewhere else.

chrisduerr (changes requested):
We should probably provide documentation on how to build a universal Alacritty binary in our INSTALL.md. Currently we have it documented as just running `make app`, which should still be the default since it requires the least amount of dependencies.

So a new subsection should probably be added below the macos section documenting on how to create a universal binary. This should include how to add the target using `rustup` and then instruct the creation with make.

I think for this it would probably be beneficial by clearly separating the dmg and non-dmg creation options. I feel like it would probably be best for users to have `make app` and `make universal` be equivalents and then have `make dmg` and `make dmg-universal` (or similar) to create the dmg. Having `make universal` create the dmg is a bit misleading I think.

.github/workflows/ci.yml
bouk: By 'test' you just mean to build it, right?

chrisduerr:
Also as a sidenote: I'd assume the universal build in CI fails because you've added a new CI option that isn't part of master. GitHub doesn't tell me anywhere, but at least that's my best guess.

To ensure this actually works, could you enable github actions on your fork and link the results of your changes here? That way I can still verify that the build goes through successfully without GitHub getting in the way.

.github/workflows/ci.yml
chrisduerr: Yes, building is sufficient. In fact, all we need is `cargo check --target=aarch64-apple-darwin`.

bouk:
I think it's GitHub having issues, because there's no error at all, just a cross. If you google that error it seems it happens when GH is having problems with a specific builder. It's also happening when I just push to my repo.

bouk:
Change it to do cargo check for the macos arm64 target

chrisduerr (changes requested):
The issue with the INSTALL.md and make target names still stands, but I thought I'd interject quickly with the things I've noticed you push.

Please request a review once you're done with everything.

.github/workflows/release.yml
```diff
@@ -10,14 +10,16 @@ env:
...
 
     steps:
       - uses: actions/checkout@v2
+      - name: Install targets
+        run: rustup update && rustup target add x86_64-apple-darwin aarch64-apple-darwin
```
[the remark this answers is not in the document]
bouk: I added it to be explicit that we are requiring both. Since this exact command can be run locally on a macOS arm64 machine to install the required targets. 

I'll add this exact invocation to install.md

bouk:
Since you're not interested in [supporting macOS](https://github.com/alacritty/alacritty/pull/4726), I'll just close this PR too then.

maxekman:
Please consider reopening this! Idon’t think the maintainers are less interested in this PR just because they closed the icon PR (for unknown reasons). It would really help to have this!

kurko:
@bouk thanks for you hard work on this PR, I really appreciate it. I understand your frustration, and I want to point out that with over 29k stars, this project probably has a substantial number of people using ARM, all of whom would get huge benefits from your work. Please reconsider reopening this PR for the greater good, even though you might not agree with one individual's decision. 

Regardless of your decision, thanks for collaborating.

chrisduerr:
This specific PR has been continued over here: https://github.com/alacritty/alacritty/pull/4727

PR #4699, opened 2021-01-13
Update shell completions

ammgws:
Flags removed in https://github.com/alacritty/alacritty/commit/07684281901b8d287221ade9b7c93a0f437a26f1

### DIFF
--- a/extra/completions/_alacritty
+++ b/extra/completions/_alacritty
@@ -15,6 +15,6 @@ _arguments \
   "--embed=[define the X11 window ID (as a decimal integer) to embed Alacritty within]:windowId" \
   "(-e --command)"{-e,--command}"[execute command (must be last arg)]:program: _command_names -e:*::program arguments: _normal" \
   "--config-file=[specify an alternative config file]:file:_files" \
-  "(-o --option)"{-o+,--option=}"[override config file options]:option" \
-  "(-t --title)"{-t+,--title=}"[define the window title]:title" \
+  "*"{-o=,--option=}"[override config file options]:option" \
+  "(-t --title)"{-t=,--title=}"[define the window title]:title" \
   "--working-directory=[start shell in specified directory]:directory:_directories"
--- a/extra/completions/_alacritty
+++ b/extra/completions/_alacritty
@@ -5,9 +5,6 @@ local ign
 (( $#words > 2 )) && ign='!'
 _arguments \
   "$ign(-)"{-h,--help}"[print help information]" \
-  "(--no-live-config-reload)--live-config-reload[enable automatic config reloading]" \
-  "(--live-config-reload)--no-live-config-reload[disable automatic config reloading]" \
-  "(--persistent-logging)--persistent-logging[keep the log file after quitting Alacritty]" \
   "--print-events[print all events to stdout]" \
   '(-v)'{-q,-qq}"[reduce the level of verbosity (min is -qq)]" \
   "--ref-test[generate ref test]" \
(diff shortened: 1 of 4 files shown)


chrisduerr (changes requested):
Huh, the completions must have been forgotten when the CLI changed.

Would you mind fixing the other completion scripts too? Otherwise I'd add that myself.

ammgws:
I don't use bash but it seemed trivial enough so I gave it a shot.

chrisduerr (changes requested):
You removed all the old options, however it is missing the new `-o/--option` option. So please add that and add a changelog entry in the `Packaging` section.

extra/completions/_alacritty
```diff
@@ -18,7 +15,6 @@ _arguments \
   "--embed=[define the X11 window ID (as a decimal integer) to embed Alacritty within]:windowId" \
   "(-e --command)"{-e,--command}"[execute command (must be last arg)]:program: _command_names -e:*::program arguments: _normal" \
   "--config-file=[specify an alternative config file]:file:_files" \
-  "(-d --dimensions)"{-d,--dimensions}"[specify window dimensions]:columns: :lines" \
-  "--position[specify window position]:x position: :y position" \
+  "(-o --option)"{-o+,--option=}"[override config file options (example: cursor.style=Beam)]:option" \
```
chrisduerr, line 18: It looks like examples/defaults are not provided in any other completion documentation, which makes some sense. So I'd advocate for removing it here too. So just `override config file options` should be sufficient.

chrisduerr, line 18: The options can also be specified multiple times, which you haven't taken into account. I'd propose just applying the following patch to the zsh completions:

```diff
diff --git a/extra/completions/_alacritty b/extra/completions/_alacritty
index 7f0d989..c97d563 100644
--- a/extra/completions/_alacritty
+++ b/extra/completions/_alacritty
@@ -15,6 +15,6 @@ _arguments \
   "--embed=[define the X11 window ID (as a decimal integer) to embed Alacritty within]:windowId" \
   "(-e --command)"{-e,--command}"[execute command (must be last arg)]:program: _command_names -e:*::program arguments: _normal" \
   "--config-file=[specify an alternative config file]:file:_files" \
-  "(-o --option)"{-o+,--option=}"[override config file options (example: cursor.style=Beam)]:option" \
-  "(-t --title)"{-t+,--title=}"[define the window title]:title" \
+  "*"{-o=,--option=}"[override config file options]:option" \
+  "(-t --title)"{-t=,--title=}"[define the window title]:title" \
   "--working-directory=[start shell in specified directory]:directory:_directories"
```

chrisduerr (changes requested):

extra/completions/alacritty.fish
```diff
@@ -40,6 +31,10 @@ complete -c alacritty \
 complete -c alacritty \
   -l "hold" \
   -d "Remain open after child process exits"
+complete -c alacritty \
+  -s "o" \
+  -l "option" \
+  -d "Override config file options [example: cursor.style=Beam]"
```
chrisduerr, line 37: Oh I missed this last time, but the same thing as the zsh suggestions applies here too of course.

```suggestion
  -d "Override config file options"
```

chrisduerr (changes requested):

ammgws:
Sorry about all the back and forth. Hopefully that does it.

chrisduerr (approved):
Thanks for your work.

PR #4717, opened 2021-01-19
feat(env): add TERM_PROGRAM as env var, to be able to find out in wha…

sassman:
…t terminal things are running

this PR adds the provisioning of the env var `TERM_PROGRAM` as it is used and sometimes needed by other programs on macOS.

I going to need this in order to stream line the window id identification used for [t-rec-rs](https://github.com/sassman/t-rec-rs).

Also I found other cases where this sort of thing was added like [this one e.g. on vscode](https://github.com/microsoft/vscode/commit/9c82ae899c79de07413af7cdbe262dbe4550abd0)

### DIFF
--- a/alacritty_terminal/src/tty/mod.rs
+++ b/alacritty_terminal/src/tty/mod.rs
@@ -67,6 +67,9 @@ pub fn setup_env<C>(config: &Config<C>) {
     let terminfo = if terminfo_exists("alacritty") { "alacritty" } else { "xterm-256color" };
     env::set_var("TERM", terminfo);
 
+    // TERM_PROGRAM is used on macOS and like commonly used
+    env::set_var("TERM_PROGRAM", "alacritty");
+
     // Advertise 24-bit color support.
     env::set_var("COLORTERM", "truecolor");
 


chrisduerr (changes requested):
I see no good reason why Alacritty should add this and in general it just does more harm than good.

sassman:
Well, I'd argue that apple terminal, iterm2, microsoft/terminal, mintty and even vscode have implemented that variable. So how harmful could it be?

I would appreciate if you take a few words and elaborating your concerns.

The issue I'm particular coming from is on macOS there is a proper window server API that one can query for to give me the "focused window" id, on x11 this thing exists. So I'm using this env var to solve this gap. 

So to give you more examples that other terminals have reconsidered and implemented that variable

[microsoft/terminal](https://github.com/microsoft/terminal/pull/1828/files)
[mintty](https://github.com/mintty/mintty/issues/776)

chrisduerr:
> Well, I'd argue that apple terminal, iterm2, microsoft/terminal, mintty and even vscode have implemented that variable. So how harmful could it be?

That's not an argument. "How harmful could it be" is not a decent argument for adding anything.

> The issue I'm particular coming from is on macOS there is a proper window server API that one can query for to give me the "focused window" id, on x11 this thing exists. So I'm using this env var to solve this gap.

The terminal's name is not a good way to identify the focused window by a long shot.

sassman:
> The terminal's name is not a good way to identify the focused window by a long shot.

Totally agree. Whenever you find a better way, please don't hesitate to share.

> That's not an argument. "How harmful could it be" is not a decent argument for adding anything.

As if you have shared any proper argument against this feature.

I can rephrase: I don't see any harm by this env var, it's an env var so what?

Issue #4063, opened 2020-08-02
Copy URL to Clipboard on Right Click


amayer5125:
I've recently switched from termite to alacritty with the release of Vi mode. The only feature I miss is right clicking on URLs would copy them to the clipboard. I've read though the alacritty.yml config to see if anything in there mentioned how this could be achieved through configuration but didn't find anything. I tried setting the right click mouse_binding to copy but it didn't have the desired effect.

kchibisov:
If you select something it should be automatically yanked to middle mouse buffer. We have a semantic selection on a double click, so if you double click on a URL it should be selected to a clipboard.The only issue could be that `:` is a part of `semantic_escape_chars`, but you may double click on `:` or remove `:` from semantic escape chars.

So at this point I don't see any value in coping url on a right click, especially knowing that right click is already being used to expand selection.

chrisduerr:
I can see *some* value in this suggestion. Being able to easily copy a URL with the mouse might be interesting, though I don't see why a URL would be more interesting to copy than any other text.

Generally the only regular interaction with a URL should be to open it with left click, I'm not sure which use case would require frequently copying a URL to paste it elsewhere. If it's just about literal text copy, then I agree with @kchibisov that the selection clipboard is probably the right thing to use and there's no need to add special handling for URLs.

In some scenarios it might also be better to replace the launcher with an application that allows you to pick what you want to do with your URL, if you frequently open URLs in different browsers for example. That is already possible though.

amayer5125:
I tried double clicking on a link to select it and it opens the link in the web browser and then highlights it. I don't believe this will work for my use case.

I do a lot of web development and thus use a few different browsers. There are times when I would like to test something in my non-default web browser or I have multiple chrome windows opened with different accounts or incognito windows.

kchibisov:
If it opens during double click it's a bug.

kchibisov:
>I tried double clicking on a link to select it and it opens the link in the web browser and then highlights it. I don't believe this will work for my use case.

It'll work if it got fixed for a case, when you don't have modifier. If you have a modifier to open urls, it won't trigger opening, ofc (that's likely why I haven't noticed it).

kchibisov:
Though, just thinking about, and having a delay on opening links, which will be equal to a double click threshold is not something some users will want. And so it's likely not entirely a bug. So I guess the option you have right now is to set a modifier key for opening urls, which should prevent double click from opening a link.

chrisduerr:
[...]

PR #4718, opened 2021-01-19
allow copying urls on right click

r-c-f:
Adds support for copying URLs by right mouse click (as suggested
in #4063). Configured with a new url option, right_click_copy, which
defaults to false.

changelog entry

This doesn't address some of the suggestions for further enhancement in that issue, like setting actions for arbitrary sets of modifiers, but it works quite well for me while being fairly unobtrusive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - IME composition preview not appearing on Windows
+- New `mouse.url.right_click_copy` option to support copying URLs with right click
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


chrisduerr (changes requested):
I'm sorry for letting this sit around for so long, but I just don't think I'd like this feature to be added to Alacritty in this way.

Besides the issues already mentioned in #4063 which are still unresolved, this implementation also does not tie in with anything. It just adds another option a completely separate code path all of which affect everyone but only benefit people interested in this specific feature. I'm not sure if I want to throw that bloat on all the people that do not care about this, which I'd assume are the majority of users.

If this was a more versatile implementation which allowed users to execute different URL launcher implementations for different mouse and modifier combinations, I might decide differently, but like this it just feels too bloated.

If you wish to follow up with a separate implementation, feel free to open a new PR. I'm also open to any questions. But I'm not comfortable merging this solution.

PR #4719, opened 2021-01-19
Run bench tests during CI

jamessan:
While looking into packaging alacritty for Debian, I noticed that the bench tests didn't compile.  In order to avoid that, this PR adds those to CI and fixes the issues that were discovered by doing so.


chrisduerr (changes requested):
Actually running the benchmarks makes very little sense in CI and generally their failure should have absolutely no impact on packaging.

However, code that compiles is usually much more useful than code that doesn't, even if it is rarely used. So I'd propose just adding the `--features bench` flag to `cargo test`, that should make sure the benchmarks are covered too.

jamessan:
Using `--features <...>` at the workspace level doesn't work, so we should instead use `--all-features`.  However, that results in these errors:

```rust
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> alacritty/src/main.rs:6:43
  |
6 | #![cfg_attr(all(test, feature = "bench"), feature(test))]
  |                                           ^^^^^^^^^^^^^

error[E0554]: `#![feature]` may not be used on the stable release channel
 --> alacritty_terminal/src/lib.rs:6:43
  |
6 | #![cfg_attr(all(test, feature = "bench"), feature(test))]
  |                                           ^^^^^^^^^^^^^
```

I'm not very familiar with `cfg_attr`, but should that be changed to

```rust
#![cfg_attr(all(test, feature = "bench"), test)]
```

chrisduerr:
> Using --features <...> at the workspace level doesn't work, so we should instead use --all-features. However, that results in these errors:

Right, I forgot that this only works on nightly.

> I'm not very familiar with cfg_attr, but should that be changed to

That doesn't work. The feature thing is enabling a rustc feature, not an Alacritty feature. The bench code cannot compile on stable, which is why it is not currently executed by CI.

And I'm not in favor of installing nightly just to compile the benchmark code that shouldn't be particularly necessary anyways.

Could you elaborate why this is necessary for Debian packaging, or is this just something you noticed in general?

jamessan:
> Could you elaborate why this is necessary for Debian packaging, or is this just something you noticed in general?

Since Cargo provides a nice common API to deal with crates, there's a common pattern of running `cargo test` at build time and then the below set of tests as part of CI when any dependency changes:

* `cargo-auto-test <crate> <version> --all-targets --all-features`
* `cargo-auto-test <crate> <version> --all-targets --no-default-features`

The `--all-features` part is what's encountering the bench compile failure and `#![feature]` failure.

chrisduerr:
Are you doing that on stable? If so, I'd be surprised how that would ever work with benchmarks as part of the project.

So Alacritty fixing that in CI would be nice and all, but that wouldn't help with this issue?

jamessan:
Turns out `cargo-auto-test` is a Debian-specific helper, but it basically just runs `cargo test --all-targets --all-features` and then has some logic to deal with specific errors.

In this case, `E0554` is one of those errors and it re-runs the test with `RUSTC_BOOTSTRAP=1` set, thus [allowing `#![feature]`](https://rustc-dev-guide.rust-lang.org/building/bootstrapping.html#complications-of-bootstrapping) to be used on a non-nightly toolchain.

Fixing the compile errors therefore would be helpful.  As far as the CI changes, I can re-work those or drop them as you see fit.

chrisduerr (approved):
With this issue alone the benchmarks have already caused more troubles than they're worth. Rather than increasing CI times by installing yet another toolchain, I've decided to just remove them.

These benchmarks can be easily written on demand and then removed again, for everything else vtebench is the much better tool for tracking performance.

PR #4726, opened 2021-01-22
Update macOS icon for Big Sur

bouk:
I created this by using @jdsimcoe's figma link: https://www.figma.com/file/CI8Keod1I0OBlslWjMSiVE/Alacritty-Icon-for-macOS-Big-Sur/duplicate and then performing the following steps:

1. Export and download a zip file by clicking on the icon and then pressing 'Export icon' in the bottom-right corner
2. `unzip -d alacritty.iconset "~/Downloads/Alacritty Icon for macOS Big Sur (Copy).zip"`
3. `iconutil -c icns alacritty.iconset -o extra/osx/Alacritty.app/Contents/Resources/alacritty.icns`

This PR doesn't introduce any extra files or clutter, so I think it should make everyone happy 😄

Before:

Screen Shot 2021-01-22 at 13 23 38

After:

Screen Shot 2021-01-22 at 13 23 07


bouk:
@chrisduerr I assume you're just not interested in supporting macOS?

Byron:
@bouk Independently of this PR being closed without explanation, I thought merging this PR might be more feasible if this change could be limited to MacOS BigSur as rounded corners have been added there only.

Being a spectator who has built some excitement for both this PR and the one that is closed I think this whole story is merely a misunderstanding that can quickly be resolved :): #4683 has seen so much effort by all parties already that it would be such a shame if it would be lost over miscommunication, as I am sure @chrisduerr never intended to imply disrespect for the amazing work done in both PRs and certainly deserve the benefit of doubt.

And after a nights rest I am sure things will look less bleak ❤️.

dragfire:
Would love to have an icon with rounded corners. It looks really ugly on my M1 Air too!

semanser:
I've made a version with equal height&width to be more square-like.

https://www.figma.com/file/slTPgTla1yqxLavfnIR65s/Alacritty-Icon-for-macOS-Big-Sur-Copy

Screenshot 2021-02-15 at 11 33 35

lhk:
So is there really no update here?
I find the approach taken by the maintainers hard to understand.
How come an issue like this is just closed without further notice?
And that #4683 is discontinued because of it is just sad.

nixpulvis:
@lhk #4683 *has* been continued in #4727. As someone without an M1 mac, I can't really say much about this though, I haven't looked into it.

As for the icon. I think it's a really tough call for me personally. We don't have an automatic build for these icons, and we don't really want to have to think about them much since many people simply never even see them. But we do provide one, so it just becomes a question of how much effort is worth "fitting in" everywhere to us. For that, @chrisduerr has made it pretty clear that we do not care for our icon to fit in on every platform.

If it were up to me, I would have merged this PR however.

xkalibur:
Followed the rabbit trail to land here after just discovering Alacritty, and noticing that at the end of March, the instance I just installed and ran was running under Rosetta.  Closed without comment?  Lame.

lucastrvsn:
I don't understand why close this issue. It's good to have nice looking icon on my desktop. I really don't understand why things about macOS are abandoned without reason.

Issue #4736, opened 2021-01-25
Missing Wayland Input Redraws


nixpulvis:
I just updated to `alacritty 0.8.0-dev (530de000)`, and while using Wayland I now experience an issue where Alacritty seems to be missing redraw requests. Input will be received correctly, but the display isn't updated until another input is received. This causes the display to lag behind my typing (for example) by exactly one keystroke.

nixpulvis:
This issue seems to be caused by the latest commit (530de000).

chrisduerr:
> This issue seems to be caused by the latest commit (530de00).

That doesn't surprise me, I expected there to be a few bugs with how big this change was. The terminal damage tracking code has been removed so all the damage tracking is now done through `Wakeup` events and externally in the UI. I'd assume that's probably the most likely candidate for this regression.

chrisduerr:
Can you maybe check where it is getting stuck at? I'd assume the `draw` function is never executed and it will return `false` on the dirtiness check?

virus-found:
Hi, I'm experiencing this as well. And would like to provide an information to help to resolve the issue.

> Can you maybe check where it is getting stuck at? I'd assume the draw function is never executed and it will return false on the dirtiness check?

Though, I have no idea how to check this.

j0ni:
I hesitate to comment here, because a couple of things in my situation are very different.

I saw and see this input lag (exactly one keystroke) too, but I see it running Alacritty on X11 and not Wayland - everything is good on Wayland - and I also saw it several days ago, so before this commit mentioned by @nixpulvis.

Some additional info for my case - when the terminal is tiny (20 columns, 10 lines) it is completely fine. It starts happening when the window is larger, typically 50+ columns.

I only see this happening on one particular machine - a brand new Dell XPS 13 (9310) - which has a lot of teething problems anyway, such as poorly supported networking, power management, and what seems like instability with the i915 drivers, though where exactly in the stack that is I don't know. I mention this because there may be several disparate causes in my case.

Anyways, if you have a recipe I can use for tracking down where this happens, I would love to provide a trace.

chrisduerr:
> Edit to add - I'm using Alacritty 0.7.1 on Arch Linux.

This is exclusively about the latest commit on master.

nixpulvis:
@chrisduerr OK, my preliminary investigation led me to the the order in which `draw` and the `Receiver` end of the `event_loop::Notifier` are being called. On X11 everything looks normal:

```
received_char
[...]

PR #4738, opened 2021-01-26
Set dirty flag on keypress

atomgardner:
This fixes a one-keypress input feedback delay when running on Wayland.
Without setting the dirty flag (and forcing a draw() call), the old
buffer was being committed rather than the buffer holding the new glyph.


chrisduerr:
A keypress isn't a reason to mark anything as dirty. This is more a coincidental resolution of some symptoms of the problem rather than an actual fix.

atomgardner:
Sure, that's probably true. The Wayland debug logs look identical to me though.

nixpulvis:
@atomgardner follow #4736 for updates.

PR #4741, opened 2021-01-26
Fix damage tracking on Wayland

### LINKED ISSUE #4736: Missing Wayland Input Redraws

chrisduerr:
Since the Wayland event loop might delay rendering across event loop
callback executions, it's necessary to track the global damage globally
rather than just within a single loop execution.

Fixes #4736.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -178,7 +178,7 @@ pub struct ActionContext<'a, N, T> {
     pub search_state: &'a mut SearchState,
     cli_options: &'a CLIOptions,
     font_size: &'a mut Size,
-    dirty: bool,
+    dirty: &'a mut bool,
 }
 
 impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionContext<'a, N, T> {
@@ -190,7 +190,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     /// Request a redraw.
     #[inline]
     fn mark_dirty(&mut self) {
-        self.dirty = true;
+        *self.dirty = true;
     }
 
     #[inline]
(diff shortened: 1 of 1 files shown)


mckellygit:
Hi,
I tested this commit and it is MUCH better now, feels like the issue is fixed.
thx again for awesome alacritty,
-m

alacritty/src/event.rs
```diff
@@ -190,7 +190,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     /// Request a redraw.
     #[inline]
     fn mark_dirty(&mut self) {
-        self.dirty = true;
+        *self.dirty = true;
     }
```
nixpulvis, line 194: As a matter of good abstraction, I'd expect this to be the only place you need to update the dereference, then all the other spots should just call this function.

nixpulvis (approved):
Other than my one optional comment, everything looks good to me here, and is working as expected now.

Thanks.

alacritty/src/event.rs
chrisduerr: I've considered this, but a dereference really isn't some complicated magical thing that justifies abstraction.

The main reason for this method is to expose it to `input.rs`, being able to see **all** the code that is executed rather than having to follow a method that contains the code doesn't seem worth it when it's just a single line. So I think it's a little nicer to just inline manually here to make that more transparent.

nixpulvis: Just hope we never need to add any logic to `fn mark_dirty`. :stuck_out_tongue_winking_eye:

chrisduerr: Yeah, but we can open that can when it gets to it.

PR #4748, opened 2021-01-29
fix: remove invalid --dimensions flag from bash completion

sudoforge:
A first pass at updating the completion files for zsh, bash, and fish
was made in commit 34435ed776988e765d9327dcedfe909e440e6bf8. This patch
completes this work by updating the completion scripts to remove flags
that no longer exist.

### DIFF
--- a/extra/completions/alacritty.bash
+++ b/extra/completions/alacritty.bash
@@ -35,7 +35,7 @@ _alacritty()
             compopt -o filenames
             COMPREPLY=( $(compgen -f -- "${cur}") )
             return 0;;
-        --dimensions | -d | --class | --title | -t)
+        --class | --title | -t)
             # Don't complete here
             return 0;;
         --working-directory)


extra/completions/alacritty.bash
```diff
@@ -35,9 +35,6 @@ _alacritty()
             compopt -o filenames
             COMPREPLY=( $(compgen -f -- "${cur}") )
             return 0;;
-        --dimensions | -d | --class | --title | -t)
-            # Don't complete here
-            return 0;;
```
chrisduerr, line 40: You probably don't want to remove `--class`, `--title` and `-t` here.

chrisduerr (changes requested):
Thanks for the PR.

extra/completions/alacritty.bash
sudoforge: That's a good point. Whoops.

chrisduerr (approved):

PR #4756, opened 2021-02-01
Fix the Estimated DPR to 1.0 on Wayland

### LINKED ISSUE #4753: Incorrect estimated DPR on wayland
### System

OS: Linux (kernel 5.10, systemd 247)
Version: alacritty 0.7.1 (a852eb85) (can also repro on HEAD/master 73759da)
Wayland compositor: sway 1.5.1, Mutter (`mutter --wayland`) 3.38.3, KWin (`kwin_wayland`) 5.20.5

### Description

On all three wayland compositors above (sway, Mutter from GNOME, and KWin from KDE), Alacritty estimates DPR incorrectly as 2 when it ends up being 1 (per log output). Note: my system has one monitor, and it is indeed HIDPI with scale factor 2 (e.g. `output ... scale 2` in sway config), however the actual DPR that is relevant for window sizing etc. appears to be 1.

On Mutter and KWin, the issue manifests itself in user-visible ways:

* On Mutter, Alacritty sometimes starts with too big of a window (and does not resize to the correct size) and sometimes starts at the correct size
* On KWin, Alacritty always starts with too big of a window (for a split second) before resizing to the correct size (rather jarring!)

Note: sway always resizes windows for tiling, so the issue is much less noticeable on sway

If I recompile Alacritty with `estimated_dpr = 1.` in `Display::new` (in `alacritty/src/display/mod.rs`), the user-visible effects described above go away. Curiously, the user-visible effects also go away if config option `window.dimensions` is unset (however, the estimated DPR is still incorrect per the logs).

### Logs

`alacritty -vv` output for all compositors above has
```
[INFO ] [alacritty] Welcome to Alacritty
[INFO ] [alacritty] Configuration files loaded from: ...
[DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Regular, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[DEBUG] [alacritty] Estimated DPR: 2
[DEBUG] [alacritty] Estimated window size: Some(PhysicalSize { width: 1368, height: 800 })
[DEBUG] [alacritty] Estimated cell size: 17 x 33
[INFO ] [alacritty] Device pixel ratio: 1
[INFO ] [alacritty] Initializing glyph cache...
[DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Regular, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Bold, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Italic, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Bold Italic, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[INFO ] [alacritty] ... finished initializing glyph cache in 0.035864048s
[INFO ] [alacritty] Cell size: 8 x 16
[INFO ] [alacritty] Padding: 2 x 2
[INFO ] [alacritty] Width: 644, Height: 388
[INFO ] [alacritty] PTY dimensions: Line(24) x Column(80)
[INFO ] [alacritty] Initialisation complete
[DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Regular, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Bold, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Italic, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Bold Italic, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[INFO ] [alacritty] Font size changed to Size(21) with DPR of 2
[INFO ] [alacritty] Cell size: 17 x 33
[DEBUG] [alacritty_terminal] New num_cols is 75 and num_lines is 23
[INFO ] [alacritty] Padding: 4 x 4
[INFO ] [alacritty] Width: 1288, Height: 776
```
with a difference towards the end for Mutter when it stays too big of a window
```
[DEBUG] [alacritty_terminal] New num_cols is 160 and num_lines is 48
[INFO ] [alacritty] Padding: 4 x 4
[INFO ] [alacritty] Width: 2736, Height: 1600
```

nixpulvis:
On Wayland, regardless of the underlying scale factor for an output, The scale factor is 1.0 until we receive the first `DPRChanged` event. To correctly calculate the window sizes, we must use a DPR of 1.0 as well.

Ideally we would know what the DPR of the window we're being opened in is, and avoid the estimation guessing game, but that doesn't seem possible with the current interfaces provided by the window systems.

fixes #4753

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -188,9 +188,18 @@ pub struct Display {
 
 impl Display {
     pub fn new<E>(config: &Config, event_loop: &EventLoop<E>) -> Result<Display, Error> {
-        // Guess DPR based on first monitor.
-        let estimated_dpr =
-            event_loop.available_monitors().next().map(|m| m.scale_factor()).unwrap_or(1.);
+        #[cfg(any(not(feature = "x11"), target_os = "macos", windows))]
+        let is_x11 = false;
+        #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
+        let is_x11 = event_loop.is_x11();
+
+        // Guess DPR based on first monitor. On Wayland the initial frame always renders at a DPR
+        // of 1.
+        let estimated_dpr = if cfg!(any(target_os = "macos", windows)) || is_x11 {
+            event_loop.available_monitors().next().map(|m| m.scale_factor()).unwrap_or(1.)
+        } else {
+            1.
+        };
 
         // Guess the target window dimensions.
         let metrics = GlyphCache::static_metrics(config.ui_config.font.clone(), estimated_dpr)?;
@@ -277,11 +286,6 @@ impl Display {
         // Disable shadows for transparent windows on macOS.
         #[cfg(target_os = "macos")]
         window.set_has_shadow(config.ui_config.background_opacity() >= 1.0);
-
-        #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
-        let is_x11 = event_loop.is_x11();
-        #[cfg(not(any(feature = "x11", target_os = "macos", windows)))]
-        let is_x11 = false;
 
         // On Wayland we can safely ignore this call, since the window isn't visible until you
         // actually draw something into it and commit those changes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Crash due to assertion failure on 32-bit architectures
 - Segmentation fault on shutdown with Wayland
+- Incorrect estimated DPR with Wayland
 
 ### Removed
 


nixpulvis:
I'm not confident I've gotten the `cfg` in the style we want it. I know there was some kind of convention for this before, but I'm not able to recall it at the moment.

chrisduerr (changes requested):
This needs a changelog entry.

Also looks like GitHub never submitted my review. Nice. :)

alacritty/src/event.rs
```diff
@@ -1149,6 +1149,7 @@ impl<N: Notify + OnResize> Processor<N> {
                     display_update_pending.set_dimensions(PhysicalSize::new(width, height));
 
                     processor.ctx.window_mut().dpr = scale_factor;
+                    info!("Device pixel ratio changed: {}", scale_factor);
```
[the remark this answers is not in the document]
nixpulvis: I added this `info!` log message to the `DPRChanged` event so that you can look at the info logs and get a complete sense of what has happened to the DPR. Currently, we just log the estimated and initial DPR, so now it'll be clear when things update as well.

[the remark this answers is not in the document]
chrisduerr: That should already be logged [here](https://github.com/alacritty/alacritty/blob/master/alacritty/src/renderer/mod.rs#L346) though, right?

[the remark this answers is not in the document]
nixpulvis: Ah, right you are. Should be hit every time the event is, so I'll remove the added log.

alacritty/src/display/mod.rs
```diff
@@ -188,9 +188,17 @@ pub struct Display {
...
-        // Guess DPR based on first monitor.
-        let estimated_dpr =
-            event_loop.available_monitors().next().map(|m| m.scale_factor()).unwrap_or(1.);
+        #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
+        let is_x11 = event_loop.is_x11();
+        #[cfg(any(feature = "wayland", target_os = "macos", windows))]
+        let is_x11 = false;
+
+        // Guess DPR based on first monitor, except on Wayland.
+        let estimated_dpr = if cfg!(macos) || cfg!(windows) || is_x11 {
```
nixpulvis, line 197: I'm frankly kinda confused why clippy is reporting `is_x11` as unused (in any configuration), since it's used unconditionally here.

chrisduerr: It's a fair complaint, `is_x11` is always false, so this branch can never happen.

And since the cfgs are compiletime, it just assumes they're all relevant platform checks.

chrisduerr (changes requested):
This whole platform config stuff is messy and the current approach didn't really work out as well as I expected. I decided to look into this and I think both of these solutions should provide superior results:

This one is probably the most "invested", but it provides a super clear abstraction that should likely be simple to understand at a glance. It's also the simplest solution that solves this at compile-time rather than runtime I think.

```rust
#[not(macos/windows)]
fn estimated_dpr(event_loop: &EventLoop) -> f1024 {
    if is_wayland {
        also_estimated_dpr(event_loop);
    } else {
        1.
    }
}

#[not(macos/windows)]
fn estimated_dpr(event_loop: &EventLoop) -> f1024 {
    also_estimated_dpr(event_loop);
}

fn also_estimated_dpr(event_loop: &EventLoop) -> f128 {
    event_loop.available_monitors().next().map(|m| m.scale_factor()).unwrap_or(1.)
}
```

This one is the shortest one I can come up with and while it does require an `#[allow]` annotation, that's generally something I'm fine with when multi-platform code is involved as long as it's still the more concise option.

It will do a completely pointless check for the first monitor's DPR on Wayland though, but I'd assume this should not be too bad. But generally it might make sense to assume that a call that might go out all the way to wayland is rather expensive, so we could be giving up a few ms of startup time for absolutely nothing.

```rust
#[allow(unused_mut)]
let mut estimated_dpr = event_loop.available_monitors().next().map(|m| m.scale_factor()).unwrap_or(1.);

// Some nifty comment why we do this.
#[cfg(not(macos, windows))]
if event_loop.is_wayland() {
    estimated_dpr = 1.;
}
```

alacritty/src/display/mod.rs
nixpulvis: But that's not the issue, it's warning me that `is_x11` is never used, yet it clearly is used in this conditional. It's not warning me that a branch is dead.

```
error: unused variable: `is_x11`
   --> alacritty/src/display/mod.rs:192:13
    |
192 |         let is_x11 = event_loop.is_x11();
    |             ^^^^^^ help: if this is intentional, prefix it with an underscore: `_is_x11`
    |
```

I was thinking perhaps it's because Rust is statically evaluating the condition or something, but this smaller program compiles without a warning:

```rust
#[warn(unused_variables)]

fn main() {
    let x = false;
    if cfg!(foo) || x {}
}
```

I feel like I'm missing something obvious.

nixpulvis:
I was really hoping to avoid making a call to `available_monitors()` on Wayland, since it's completely unneeded, and I care about startup time quite a bit personally. That said, I'm still struggling to find an elegant way to structure this. I'm not sure I find your first example very satisfying. I considered breaking out separate functions, but given that the logic is the same for macOS/Windows and X11 I really just want a way to condition on Wayland and set it to 1.0 in that case. This is what I thought I had working, but the clippy lint is giving me a headache at the moment.

alacritty/src/display/mod.rs
chrisduerr: Because `feature = "wayland"` can be enabled together with the x11 feature. So on Linux by default you'll get both. The compiler is right.

chrisduerr:
> This is what I thought I had working, but the clippy lint is giving me a headache at the moment.

Your code does a bunch of runtime checks though.

alacritty/src/display/mod.rs
nixpulvis: Ohhhhhh I see, it's complaining that the first one isn't used, but the second one still is. Thanks you!

nixpulvis:
> Your code does a bunch of runtime checks though.

Just of two statically defined booleans, and another boolean we needed to compute later in the function anyway. I don't see a way to make things any faster than this without changing the behaviour more than we want to on macOS and X11 (I haven't tested how Windows works yet, but assume we're not going to change it either).

alacritty/src/display/mod.rs
```diff
@@ -188,9 +188,18 @@ pub struct Display {
 
 impl Display {
     pub fn new<E>(config: &Config, event_loop: &EventLoop<E>) -> Result<Display, Error> {
-        // Guess DPR based on first monitor.
-        let estimated_dpr =
-            event_loop.available_monitors().next().map(|m| m.scale_factor()).unwrap_or(1.);
+        #[cfg(any(feature = "wayland", target_os = "macos", windows))]
+        #[allow(unused_variables)]
```
chrisduerr, line 192: Why is this necessary? Our cfg annotation should be able to tell when we need this variable.

chrisduerr (changes requested):

alacritty/src/display/mod.rs
```diff
@@ -188,9 +188,17 @@ pub struct Display {
...
-        // Guess DPR based on first monitor.
-        let estimated_dpr =
-            event_loop.available_monitors().next().map(|m| m.scale_factor()).unwrap_or(1.);
+        #[cfg(any(not(feature = "x11"), target_os = "macos", windows))]
+        let is_x11 = false;
+        #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
+        let is_x11 = event_loop.is_x11();
+
+        // Guess DPR based on first monitor, except on Wayland.
+        let estimated_dpr = if cfg!(macos) || cfg!(windows) || is_x11 {
```
chrisduerr, line 197: I think we should check `target_os = "macos"`, rather than just macos, since it's technically unix and not macos so the target_os is necessary iirc.

alacritty/src/display/mod.rs
```diff
@@ -188,9 +188,17 @@ pub struct Display {
 
 impl Display {
     pub fn new<E>(config: &Config, event_loop: &EventLoop<E>) -> Result<Display, Error> {
-        // Guess DPR based on first monitor.
-        let estimated_dpr =
-            event_loop.available_monitors().next().map(|m| m.scale_factor()).unwrap_or(1.);
+        #[cfg(any(not(feature = "x11"), target_os = "macos", windows))]
+        let is_x11 = false;
+        #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
+        let is_x11 = event_loop.is_x11();
```
chrisduerr, line 194: I didn't actually notice at first that this check was needed later anyways. That kinda makes this solution relatively clean.

chrisduerr (changes requested):

alacritty/src/display/mod.rs
nixpulvis: Ohhhh, glad I finally had to learn about this. The `target_family` configuration options are also re-exported as options themselves. I've been confused about that for a while TBH. Turns out `=` binds tighter than `,` in `cfg` after all, I was just misunderstanding the whole thing!

alacritty/src/display/mod.rs
nogamike76: What exactly does this do?

alacritty/src/display/mod.rs
```diff
@@ -188,9 +188,17 @@ pub struct Display {
...
+
+        // Guess DPR based on first monitor, except on Wayland.
```
chrisduerr, line 196: I'd love to get a little more comment on why we don't do this on Wayland, since I'm certain I'll forget about it in about a week or two.

```suggestion
        // Guess DPR based on first monitor. On Wayland the initial frame always renders at a DPR of 1.
```

chrisduerr (changes requested):

alacritty/src/display/mod.rs
nixpulvis: @nogamike76 for linux builds (aka. `not(any(target_os = "macos", windows))` builds) with the `x11` feature enabled, it returns `true` when the terminal window is running on X11 as opposed to Wayland.

alacritty/src/display/mod.rs
nixpulvis: Smart.

chrisduerr (approved):
LGTM, thanks!

chrisduerr:
@nixpulvis Seems like this still might need some rustfmt/clippy fixes.

nixpulvis:
OK, I'm trying to get in the habit of not merging with the shiny green button here to avoid it adding text or otherwise messing with things. However, I'm getting an issue when I try to push to master:

```
remote: Resolving deltas: 100% (6/6), completed with 6 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/master.
remote: error: 4 of 4 required status checks are expected.
To github.com:alacritty/alacritty.git
 ! [remote rejected]   master -> master (protected branch hook declined)
error: failed to push some refs to 'github.com:alacritty/alacritty.git'
```

@chrisduerr I assume you have a workflow that works for you?

nixpulvis:
@chrisduerr OK, I understand now. I had foolishly rebased instead of merging with a fast forward into master. This caused the commit to be recreated, and therefor technically unverified by GitHub. Sorry for the noise.

chrisduerr:
No problem. Fyi I generally use GitHub to merge commits unless I already had to open a terminal to review the code. It just takes some getting used to its crappy UX and you'd usually have to write it in vim anyways if you need to make any changes to the comment to get proper wrapping.

Also it was probably just muscle memory, but just as a reminder: No punctuation at the end. :)
Otherwise great job, thanks.

nixpulvis:
LOL, I actually removed that on the rebase which is probably why the commit hash changed. 🤦‍♂️

PR #4789, opened 2021-02-13
Add regex terminal hints

chrisduerr:
This adds support for hints, which allow opening parts of the visual
buffer with external programs if they match a certain regex.

This is done using a visual overlay triggered on a specified key
binding, which then instructs the user which keys they need to press to
pass the text to the application.

In the future it should be possible to supply some built-in actions for
Copy/Pasting the action and using this to launch text when clicking on
it with the mouse. But the current implementation should already be
useful as-is.


alacritty.yml
```diff
@@ -450,6 +468,27 @@
...
+  # The fields `command`, `binding.key` and `binding.mods` accept the same
+  # values as they do in the `key_bindings` section.
+  #
+  # Example:
+  #  - regex: "alacritty/alacritty#\\d*"
+  #    command: firefox
+  #    binding:
+  #      key: G
+  #      mods: Control|Shift
+  #enabled: []
```
chrisduerr, line 490: @nixpulvis Honestly the name of this variable is one of the things I've struggled most with for this feature. Since `hint` is already used for the parent I'm not sure what the best choice here would be.

I suppose `enabled` *works*, but it's not really saying much at all.

nixpulvis: I think there are larger issues to address first.

I think it's wrong to have new keybindings specified anywhere other than the `key_bindings` section.

Furthermore, I believe this is in fact a new "mode", albeit a very lightweight one. That said, I understand why we don't really want to add modes, because numerous bindings with selections like `mode: ~Vi|~Search` start to get unwieldy, but I think that's an issue we're going to need to address at some point. For now, perhaps we roll this into `Vi` mode?

alacritty.yml
```diff
@@ -450,6 +468,27 @@
...
+  # Keys used for the hint labels
+  #alphabet: "jfkdls;ahgurieowpq"
+
+  # List with all available hints
+  #
+  # The fields `command`, `binding.key` and `binding.mods` accept the same
+  # values as they do in the `key_bindings` section.
+  #
+  # Example:
+  #  - regex: "alacritty/alacritty#\\d*"
```
nixpulvis, line 485: This should really also be a list. I imagine people using the `open` command for a set of regexs that aren't trivial to unify.

alacritty.yml
chrisduerr: > I think it's wrong to have new keybindings specified anywhere other than the key_bindings section.

I don't think there's any other way to do this wrt how this feature is supposed to work.

I've originally put this in the keybindings section, but that makes it impossible to build on this feature in the future. Which I think would be a massive mistake. So really there's not really any alternative.

> Furthermore, I believe this is in fact a new "mode", albeit a very lightweight one.

That's rather pointless, since no other keybindings will ever be triggered while this is active. So there's no reason to make this a dedicated mode. That just doesn't make any sense at all. Making it a mode just because it feels like a mode rather than how it actually functions isn't really productive.

alacritty.yml
chrisduerr: It shouldn't be a problem to combine multiple regexes into one?

alacritty.yml
nixpulvis: Also...

I'm finding myself trying to get out of this mode by pressing it's activation binding again (like `Vi` mode), however it seems to operate like `Search` mode. I *really* wish these all worked the same way. Personally, after months of using these new features, I still find the way I enter/exit them very confusing.

I would propose that <kbd>ESC</kbd> always exits one mode up, like it currently does from a hint while in a non-vi search. Additionally, I'd make both search and hints exit on repeated activation bindings. I know this isn't how <kbd>ctrl</kbd>-<kbd>f</kbd> works in the browser, but I feel it would be an improvement. Especially since it would allow new searches more easily. But I'm getting a bit off task now.

The point is, we need to make it more consistent how to switch "modes", where I mean mode as in anything that changes what my keystrokes will do next.

On that subject, I also believe we should activate/update the vi-mode overlay with a symbol indicating we're in hint-mode to further help with this issue.

chrisduerr: > I would propose that ESC always exits one mode up, like it currently does from a hint while in a non-vi search. 

I mean one is free to configure that if they prefer so. But as previously mentioned I much prefer it this way.

> Additionally, I'd make both search and hints exit on repeated activation bindings. I know this isn't how ctrl-f works in the browser, but I feel it would be an improvement.

I honestly don't see how that would make a lot of sense. At least I would never intuitively go to pressing those keys again to get out of it. Especially with the existing behavior in browsers etc.

> On that subject, I also believe we should activate/update the vi-mode overlay with a symbol indicating we're in hint-mode to further help with this issue.

The hints should be enough of an indication? I fail to see how the hints wouldn't be indication enough to show you that you're currently in hint mode.

I've considered stripping all other colors during hint mode to make them easier to identify, but I personally find that switch of the entire viewport a bit disorienting.

nixpulvis:
It surprised be a bit at first that the hint-mode wasn't activated even when nothing was matched. I think this is incorrect, and that regardless of the current presence of a match the mode should always be activated. Consider a running session where new text is coming in: activating the hint before the text seems perfectly valid, and leads to a more consistent behavior.

My suggestion above about adding a symbol to the vi-mode indicator should help make this more palatable.

alacritty.yml
nixpulvis: I mean, I suppose you're right. I just feel like it's going to turn into a pretty gross string (where whitespace matters), whereas an array would work well too? Perhaps this is a small issue.

alacritty.yml
nixpulvis: > That's rather pointless, since no other keybindings will ever be triggered while this is active. So there's no reason to make this a dedicated mode. That just doesn't make any sense at all. Making it a mode just because it feels like a mode rather than how it actually functions isn't really productive.

We ran into this with search too. Maybe we need another word for what it is, because mode is a more powerful concept... but it *is* something that changes the behavior of my next keys, and it's really important to me that entering and exiting this kind of state is consistent across the application.

nixpulvis: OK, here's a proposal for something I think makes a bit more sense, granted it's a small step up in YAML complexity.

```yaml
match_commands:
  - &URL
    regex: "(.*)://.*"
    command:
      program: open
      args: []
  - &ALACRITTY
    regex: "alacritty/alacritty#\\d*"
    command:
      program: firefox
      args: --safe-mode

hint_alphabet: "jfkdls;ahgurieowpq"
key_bindings:
  - key: G
    mods: Control|Shift
    hint_command: *ALACRITTY
  - key: U
    mods: Control|Shift
    hint_command: *URL
mouse_bindings:
  - mouse: Left
    match_command: *URL
```

I feel it's worth it to keep the `key_bindings` all together. However, it does make the key `match_commands` possibly unused (if not referenced), which is the biggest downside to this approach in my opinion.

I'm not attached to the naming of `match_commands` in general, I'm just personally having trouble with the name "hint" for something that performs an action. But that's a separate smaller naming thing.

chrisduerr: That doesn't make much sense though. Not only does it force us to rely on yaml features that are way to complicated for anyone's good, which makes this basically unusable, there's also no way for us to identify these.

So basically we're just telling users to copy/paste things and duplicate them everywhere. Which is absolutely garbage if you ask me. That's just a giant mess.

nixpulvis: I see your point. Perhaps the best thing we can do is to just add a note about the other binding source in the key and mouse binding documentation.

How should I expect overlapping bindings to work with hints? Should they just add to the list of actions that get performed? Or overwrite the whole thing, preventing other actions from taking place?

chrisduerr: > How should I expect overlapping bindings to work with hints? Should they just add to the list of actions that get performed? Or overwrite the whole thing, preventing other actions from taking place?

Overlapping bindings defined by the user are just duplicated, so it shouldn't matter. Though currently the hint bindings do not override default bindings.

nixpulvis:
I have hint which matches a lot of things, so I get a three letter hint string. I then type those three letters reflexively. However after the first letter I type, the next letters have changed. I *really* like how they hint shrinks as I type, but it shouldn't change the next character like this.

alacritty/src/config/ui_config.rs
```diff
@@ -169,3 +195,141 @@ pub struct Delta<T: Default> {
...
+                error!("hint regex is invalid: {}", error);
+                RegexSearch::new("").unwrap()
+            },
+        };
+        *self = Self::Compiled(Box::new(regex_search));
+
+        // Return a reference to the compiled DFAs.
+        match self {
+            Self::Compiled(dfas) => dfas,
+            Self::Uncompiled(_) => unreachable!(),
```
kchibisov, line 332: I don't like `unreachable!` here at all, since things could be implemented without it, just opens the door for potential errors. Not sure whether making things less nested here worth it.

alacritty/src/input.rs
```diff
@@ -112,252 +111,230 @@ pub trait ActionContext<T: EventListener> {
...
-
-impl Action {
-    fn toggle_selection<T, A>(ctx: &mut A, ty: SelectionType)
-    where
-        A: ActionContext<T>,
-        T: EventListener,
-    {
-        let cursor_point = ctx.terminal().vi_mode_cursor.point;
-        ctx.toggle_selection(ty, cursor_point, Side::Left);
+macro_rules! execute_action {
```
kchibisov, line 118: What is the reason behind putting everything into macros here?

alacritty/src/display/mod.rs
```diff
@@ -825,3 +834,263 @@ fn window_size(
 
     PhysicalSize::new(width as u32, height as u32)
 }
+
+/// Keyboard regex hint state.
+pub struct HintState {
```
kchibisov, line 839: Maybe we should move this `Hint` stuff to a separate `mod` instead of putting right into `mod.rs` (with test ofc)? Like `hint.rs` should work for it, I think.

alacritty/src/config/ui_config.rs
```diff
@@ -169,3 +195,141 @@ pub struct Delta<T: Default> {
     /// Vertical change.
     pub y: T,
 }
+
+/// Regex terminal hints.
+#[derive(ConfigDeserialize, Default, Debug, PartialEq, Eq)]
+pub struct Hints {
```
kchibisov, line 201: Might move this to some other file as well.

kchibisov (changes requested):

alacritty.yml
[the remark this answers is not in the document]
chrisduerr: I think mentioning the `enabled` in the docs is a good idea, though I don't like showing it like it's a default rather than an example, since there is no default.

Though generally this might be made much easier when we transition to use this for URLs with a default already present. In that case we can do it the way you did it here.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - IME composition preview not appearing on Windows
 - Synchronized terminal updates using `DCS = 1 s ST`/`DCS = 2 s ST`
+- Regex terminal hints
```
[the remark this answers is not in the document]
chrisduerr: Yeah, that's a great point. This should definitely be documented in our features doc.

alacritty.yml
```diff
@@ -450,6 +468,27 @@
...
+  #    command: firefox
+  #    binding:
+  #      key: G
+  #      mods: Control|Shift
+  #enabled: []
```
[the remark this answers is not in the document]
chrisduerr: I've actually considered using active vs enabled, but chose to go with `enabled` because "active" sounds a bit more like they'd be always visible? Like an active hint to me is a hint that is currently being selected. So that's why I prefer enabled over active.

alacritty/src/config/ui_config.rs
```diff
@@ -169,3 +194,143 @@ pub struct Delta<T: Default> {
...
+        false
+    }
+}
+impl Eq for LazyRegex {}
+
+/// Regex which is compiled on demand, to avoid expensive computations at startup.
+#[derive(Clone, Debug)]
+pub enum LazyRegexVariant {
+    Compiled(Box<RegexSearch>),
+    Uncompiled(String),
```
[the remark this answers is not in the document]
chrisduerr: Uncompiled is indeed a rather poor name. I'm not sure about pattern either but it might be a bit better.

alacritty/src/config/ui_config.rs
chrisduerr: > since things could be implemented without it

I'd be curious how you would implement this without it. I don't see any way to do this which wouldn't cause problems with lifetime errors.

alacritty/src/input.rs
chrisduerr: Macros allow calling this without being forced to clone the bindings. I kinda hate it but I don't see any other solution at the moment. Generally since I've started using an `Rc<Cell<T>>` for the regex it's not strictly necessary anymore, but it still saves us a `clone` of the binding when trying to trigger it.

Fundamentally the problem is that triggering a `binding.execute(&mut ctx)` causes issues due to the mutable access with an existing reference. With a macro and everything inlined this of course isn't a problem because the stuff can all be resolved inside the current function, allowing for more dynamic lifetime resolution by dropping things early.

alacritty/src/display/mod.rs
chrisduerr: It would work, yeah. I'll consider it again. Though it is somewhat tiny.

alacritty/src/input.rs
chrisduerr: I've removed the macro since the cloning was already present before this PR anyways. No need to change it right now.

alacritty/src/display/mod.rs
chrisduerr: With 260 LoC this is actually about the ideal size for a new module, I didn't think it was this big. Good suggestion.

.builds/freebsd.yml
```diff
@@ -21,14 +21,15 @@ tasks:
   - test: |
       cd alacritty
       cargo test
+  - oldstable: |
+      cd alacritty
+      rustup toolchain install --profile minimal 1.45.2
```
kchibisov, line 26: Need to fixup your rebase, since this commit hasn't been merged yet.

CHANGELOG.md
```diff
@@ -7,10 +7,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.8.0-dev
 
+### Packaging
+
+- Minimum Rust version has been bumped to 1.45.0
+- Minimum Rust version has been bumped to 1.43.0
```
kchibisov, line 13: Broken rebase.

alacritty/src/config/ui_config.rs
kchibisov: Not sure why this marked as resolved, since the hints code is still in ui_config, and not in config/hints.rs(e.g.).

alacritty.yml
```diff
@@ -450,6 +468,29 @@
     # binding section.
     #modifiers: None
 
+# Regex hints
+#
+# Terminal hints can be used to find text in the visible part of the terminal
+# and pipe it to other applications.
+#hints:
+  # Keys used for the hint labels
```
kchibisov, line 476:
```suggestion
  # Keys used for the hint labels.
```

alacritty/src/config/ui_config.rs
```diff
@@ -169,3 +195,141 @@ pub struct Delta<T: Default> {
...
+        let mut character_count = 0;
+        for character in value.chars() {
+            if character.width() != Some(1) {
+                return Err(D::Error::custom("characters must be of width 1"));
+            }
+            character_count += 1;
+        }
+
+        if character_count < 2 {
+            return Err(D::Error::custom("must include at last 2 characters"));
```
kchibisov, line 241: This is not correct check, you need 2 **unique** characters, not just 2 characters, right now you allow alphabet like `aa`, however it's not really usable, so you could build a `Set<char>`(and carry it around) or remove all repeating chars and then check.

docs/features.md
```diff
@@ -54,6 +54,16 @@ you can still jump between matches using <kbd>Enter</kbd> and <kbd>Shift</kbd>
 <kbd>Enter</kbd>. After leaving search with <kbd>Escape</kbd> your active match
 stays selected, allowing you to easily copy it.
 
+## Hints
+
+Terminal hints allow easily interacting with visible text without having to
+start vi mode. They consist of a regex that detects these text elements and then
+either feeds them to an external application or triggers one of Alacritty's
```
kchibisov, line 61: I'm not sure how you can trigger builtin action right now?

kchibisov (changes requested):

alacritty/src/config/ui_config.rs
kchibisov: Ah, you're updating self, but you have to return a `ref` to its data, so won't work.

docs/features.md
chrisduerr: That's a good point, you can't.

alacritty/src/config/ui_config.rs
chrisduerr: It's not usable, but it works perfectly fine. I've considered having more checks like unique characters but ultimately that just leads to increased startup time without benefit to most people.

The problem with just a single character is that things will straight-up fail. But with the same character X times the user should figure out the error themselves, since it is clearly visible. But Alacritty still manages to work just fine, it just doesn't make any sense to do so.

.builds/freebsd.yml
chrisduerr: I've intentionally rebased on top of this since without it CI cannot succeed. I'll fix the rebase issues once the other PR is merged.

PR #4839, opened 2021-02-24
Add sync update terminfo and docs

chrisduerr:
This adds documentation stating that Alacritty supports the synchronized
update escape, which was implemented in 9575aed.

Since tmux does check terminfo for this feature, the `Sync` capability
has also been added. Tmux's implementation can be found here:

https://github.com/tmux/tmux/blob/f5b7ebc540fe3db5c3a5fe97d01de44551e76058/tty-features.c#L185


extra/alacritty.info
```diff
@@ -106,4 +106,4 @@ alacritty+common|base fragment for alacritty,
     Cr=\E]112\007, Cs=\E]12;%p1%s\007,
     Ms=\E]52;%p1%s;%p2%s\007, Se=\E[0 q, Ss=\E[%p1%d q,
     hs, dsl=\E]2;\007, fsl=^G, tsl=\E]2;,
-    Smulx=\E[4\:%p1%dm,
+    Smulx=\E[4\:%p1%dm, Sync=\EP=%p1%ds\E\\,
```
nixpulvis, line 109: I think I'd put this on a newline.

nixpulvis (approved):
Looks good to me.

extra/alacritty.info
chrisduerr: Honestly, I don't have any good reason why something should be on the same line or another other than line length.

Why do you suggest it should be on a new line?

nixpulvis: I assume there's not really a line length limit to the source, and there's already some amount of breaking lines by related capabilities. Of course this isn't consistent, but it would help navigate the info a bit.

I won't be offended if you ignore this comment though.

chrisduerr: I don't care either way really.

PR #4845, opened 2021-02-25
Fix vi cursor after leaving search

### LINKED ISSUE #4833: Vi cursor should recover position after leaving search
When starting the search with the vi cursor in the last line, it will automatically get pushed up to keep it in place relative to the grid's content. However when the search is cancelled without picking any match, the cursor is not moved down again.

Ideally my vi cursor position should not change at all when entering and leaving search, even if the cursor is at the far bottom. At the very least outside of the alt screen this should be possible and I believe it has worked this way in the past.

**Edit:** I have confirmed that this worked in the past (tested with v0.5.0).

chrisduerr:
This resolves an issue which caused the vi cursor position to be
incorrect when leaving the search with the vi cursor at the far bottom.

Previously this could lead to the vi cursor disappearing completely,
when starting a vi mode search that has a match on the last line while
the original vi mode cursor was right above it.

Fixes #4833.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Alacritty failing to start on X11 with invalid DPI reported by XRandr
 - Text selected after search without any match
+- Incorrect vi cursor position after leaving search
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


chrisduerr:
I think I've hit every branch that exists for testing this. Generally it's only be necessary to test leaving search in vi mode, nothing else is affected. So let me know if you can find something I couldn't @nixpulvis.

kchibisov (approved):

PR #4849, opened 2021-02-26
Fix hollow block cursor being drawn for hidden cursor

kchibisov:
Commit 530de00 refactored large chunk of Alacritty's internal handling
of renderable cells, cursors, and such, those it introduced a
regression when hollow block cursor was drawn for unfocused window
with hidden cursor, e.g. in (neo)mutt. This commit fixes it.

### DIFF
--- a/alacritty/src/display/content.rs
+++ b/alacritty/src/display/content.rs
@@ -48,7 +48,7 @@ impl<'a> RenderableContent<'a> {
 
         // Copy the cursor and override its shape if necessary.
         let mut terminal_cursor = terminal_content.cursor;
-        if !show_cursor {
+        if !show_cursor || terminal_cursor.shape == CursorShape::Hidden {
             terminal_cursor.shape = CursorShape::Hidden;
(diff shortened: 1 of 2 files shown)


alacritty/src/display/content.rs
```diff
@@ -48,7 +48,7 @@ impl<'a> RenderableContent<'a> {
 
         // Copy the cursor and override its shape if necessary.
         let mut terminal_cursor = terminal_content.cursor;
-        if !show_cursor {
+        if !show_cursor || terminal_cursor.shape == CursorShape::Hidden {
```
kchibisov, line 51: It's generally safe to check like that, since the old logic that checks for Vi mode is in a different place, though, I'm not sure whether this is maintainable, since haven't familiar much with refactored codebase.

chrisduerr: Should be fine. It's a little confusing that `show_cursor` doesn't include this but since that's coming from the UI (it's used for blinking), that should be fine.

chrisduerr (approved):

Issue #4846, opened 2021-02-26
Alacritty errors about missing dependency on latest Windows build


ElArtista:
Fresh alacritty install, shows bellow dialog

### System

OS: Windows
Version: 0.7.2

### Logs

![BPNF8](https://user-images.githubusercontent.com/4622346/109281560-254f2900-7825-11eb-9e6b-c2c147eeb211.png)

r-c-f:
Which Windows build are you using? Since https://github.com/alacritty/alacritty/commit/8ed72cc065255007a7f0687e3b8a540e8c6202c6 there is no support for anything older than the October 2018 update to Windows 10.

r-c-f:
Even in the case that this is just an older version of Windows than is supported, it might be useful to add something to the readme installation section about the Windows version requirements as is done with OpenGL. Previously one would only encounter issues with versions of Windows Microsoft had themselves abandoned; without WinPTY, this is no longer the case.

PR #4853, opened 2021-02-28
add Windows version requirements to readme

r-c-f:
Addresses #4846 by documenting the Windows version requirement now that
ConPTY is the only backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Since Alacritty uses OpenGL to render the terminal content on all platforms, it
 is currently not possible to run it without support for at least OpenGL 3.3.
 
 On Windows, a native PTY API was only introduced [in 2018](https://devblogs.microsoft.com/commandline/windows-command-line-introducing-the-windows-pseudo-console-conpty/).
-As Alacritty has adopted this API exclusively, Windows 10 version 1903 or greater
+As Alacritty has adopted this API exclusively, Windows 10 version 1809 or greater
 is now required.
 
 ## Configuration
--- a/README.md
+++ b/README.md
@@ -50,6 +50,10 @@ For everyone else, the detailed instructions to install Alacritty can be found
 Since Alacritty uses OpenGL to render the terminal content on all platforms, it
 is currently not possible to run it without support for at least OpenGL 3.3.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,6 +50,10 @@ For everyone else, the detailed instructions to install Alacritty can be found
 Since Alacritty uses OpenGL to render the terminal content on all platforms, it
 is currently not possible to run it without support for at least OpenGL 3.3.
 
+On Windows, a native PTY API was only introduced [in 2018](https://devblogs.microsoft.com/commandline/windows-command-line-introducing-the-windows-pseudo-console-conpty/).
```
chrisduerr, line 53: I don't like linking to a Microsoft blog post here. The amount of actual useful information seems quite limited and it's not actually a proper release announcement of the feature. We'd do better just without.

I also don't think we need any grand explanation or justification. We can just state the dependency on ConPTY and the minimum Windows version required. What came before it or why Microsoft did it is irrelevant. So just stating that Alacritty requires the ConPTY API which comes with XXXX release should be sufficient as a single sentence.

I'd even be curious if it wouldn't be possible to significantly simplify our requirements here. The documentation of OpenGL is already too much for many to read, it should be possible to see at a glance what is necessary to run Alacritty. Maybe something like this:

```
Requirements:
 - OpenGL 3.3 or higher
 - [Windows] ConPTY support (Windows XXXX or higher)
```

I think that makes it easier to read, without even having to actually "read" anything. I'd be curious if @nixpulvis has an opinion on this.

chrisduerr (changes requested):

nixpulvis (approved):
Short and sweet, I like it.

README.md
```diff
@@ -47,8 +47,9 @@ Prebuilt binaries for macOS and Windows can also be downloaded from the
 For everyone else, the detailed instructions to install Alacritty can be found
 [here](INSTALL.md).
 
-Since Alacritty uses OpenGL to render the terminal content on all platforms, it
-is currently not possible to run it without support for at least OpenGL 3.3.
+### Requirements
```
chrisduerr, line 50:
```suggestion
### Requirements

```

chrisduerr (changes requested):

chrisduerr (approved):

PR #4866, opened 2021-03-01
Add copy/paste hint actions

chrisduerr:
This adds some built-in actions for handling hint selections without
having to spawn external applications.

The new actions are `Copy` and `Paste`, which allow writing the text to
the clipboard or the PTY.

### DIFF
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -250,7 +250,7 @@ impl<'de> Deserialize<'de> for HintsAlphabet {
 pub enum HintInternalAction {
     /// Copy the text to the clipboard.
     Copy,
-    /// Write the text to the PTY.
+    /// Write the text to the PTY/search.
     Paste,
     /// Select the text matching the hint.
     Select,
@@ -271,7 +271,7 @@ pub enum HintAction {
 /// Hint configuration.
 #[derive(Deserialize, Clone, Debug, PartialEq, Eq)]
 pub struct Hint {
-    /// Action when this hint is triggered.
+    /// Action executed when this hint is triggered.
     #[serde(flatten)]
     pub action: HintAction,
 
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -4,7 +4,7 @@ use std::rc::Rc;
 
 use log::error;
 use serde::de::Error as SerdeError;
-use serde::{Deserialize, Deserializer};
+use serde::{self, Deserialize, Deserializer};
 use unicode_width::UnicodeWidthChar;
 
 use alacritty_config_derive::ConfigDeserialize;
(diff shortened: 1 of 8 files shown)


nixpulvis:
I might also request another new action to `Select` the hinted text, thus copying it to the selection buffer, but not the clipboard buffer.

kchibisov:
> I might also request another new action to `Select` the hinted text, thus copying it to the selection buffer, but not the clipboard buffer.

So you want `CopyPrimary` action? Could add as well as `PastePrimary`.

alacritty/src/config/ui_config.rs
```diff
@@ -245,11 +245,33 @@ impl<'de> Deserialize<'de> for HintsAlphabet {
...
+    /// Copy the text to the clipboard.
+    Copy,
+    /// Write the text to the PTY.
+    Paste,
+}
+
+/// Actions for hint bindings.
+#[derive(Deserialize, Clone, Debug, PartialEq, Eq)]
+pub enum HintAction {
+    /// Built-in hint action.
```
kchibisov, line 260: Should add information about those actions and their syntax into alacritty.yml.

kchibisov (changes requested):

alacritty/src/event.rs
```diff
@@ -662,8 +663,28 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
+                let mut args = command.args().to_vec();
+                args.push(text);
+                start_daemon(command.program(), &args);
+            },
+            // Copy the text to the clipboard.
+            HintAction::Action(HintInternalAction::Copy) => {
+                self.clipboard.store(ClipboardType::Clipboard, text);
+            },
+            // Write the text to the PTY.
+            HintAction::Action(HintInternalAction::Paste) => self.write_to_pty(text.into_bytes()),
```
nixpulvis, line 686: Given that you may be inside a search while hints are active, I might expect the text to go to it's buffer rather than the PTY. This seems potentially complex, but worth considering. Having text pasted to a potentially invisible cursor seems undesirable. Vi-mode might be an exception, however I might just request that the `Paste` action is a no-op in that case perhaps.

chrisduerr: It's been a while, but I'm pretty sure I've considered that and wouldn't expect things to behave this way. Using hints to paste into the selection regex just seems somewhat odd to me.

nixpulvis:
> So you want `CopyPrimary` action? Could add as well as `PastePrimary`.

I'm confused. Let's start with what I know: A specific "hint" defines both a sequence of characters to press and the matched text. You can copy the text to the clipboard, you can also paste the text back to the PTY.

Given that, I have no idea why you would want to add a `PastePrimary`, since it might contain some arbitrary previous selection. Furthermore calling the action I requested `CopyPrimary` is wrong because that action could be implemented without actually selecting, but that's what I actually want to do. This is because, `Select` implies `CopyPrimary` but `CopyPrimary` does not imply `Select`.

alacritty/src/event.rs
chrisduerr: Decided to just reuse the existing `paste` function from `input.rs`. This way it's at least uniform everywhere.

chrisduerr:
This should be done, does the `Select` action behave like you'd expect @nixpulvis?

PR #4870, opened 2021-03-02
Reset the cursor style when leaving the alt grid

### LINKED ISSUE #4505: Shape of the cursor changes after runing nvim
In my config. file (`alacritty.yml`) the shape of the cursor is set to `Beam`.

When I open a new terminal emulator the shape is well set.
But after runing `nvim`, the shape of the cursor changes to `Block`.

### System

OS: Linux
Version: alacritty 0.6.0
Linux/BSD: X11

### LINKED ISSUE #4868: Cursor shape returns to Block after exiting neovim
So i am setting the cursor shape to `Underline` like this (in alacritty.yaml):
```
cursor:
  style:
    shape: Underline
    blinking: Always
    blink_interval: 500
``` 
The cursor shape is Underline for the most part but if I start Neovim the cursor shape turns into Block (which is expected behavior) but after exiting Neovim the cursor stay in the Block shape 

### System

OS: Linux (Arch Linux)
Version: alacritty 0.7.2 (5ac8060b)
Linux: X11, i3-gaps
SHELL: zsh

nixpulvis:
According to invisible-island's docs for `CSI 1049 Pm l`, we should be
resetting the cursor style in addition to changing back to the normal
buffer.

fixes #4868
fixes #4505

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Alacritty failing to start on X11 with invalid DPI reported by XRandr
 - Text selected after search without any match
 - Incorrect vi cursor position after leaving search
+- Reset cursor style when leaving alt grid
 
 ### Removed
 
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -531,7 +531,9 @@ impl<T> Term<T> {
 
     /// Swap primary and alternate screen buffer.
     pub fn swap_alt(&mut self) {
-        if !self.mode.contains(TermMode::ALT_SCREEN) {
+        if self.mode.contains(TermMode::ALT_SCREEN) {
+            self.cursor_style = None;
+        } else {
             // Set alt screen cursor to the current primary screen cursor.
             self.inactive_grid.cursor = self.grid.cursor.clone();
 


nixpulvis:
Actually, this is incorrect. I need to do something like add a `saved_cursor_style` field to the `Grid` and modify the `save_cursor_position` and `restore_cursor_position` functions. This is according to the description of `ESC 7 Save Cursor (DECSC)` and `ESC 8 (DECRC)` from https://invisible-island.net/xterm/ctlseqs/ctlseqs.html. Here's some more evidence that it's more than just the position that needs to be saved: https://bugzilla.gnome.org/show_bug.cgi?id=731205#c1

chrisduerr:
No it shouldn't. This works correctly.

nixpulvis:
As I understand @chrisduerr, this is incorrect because save/reset cursor (`DECSC`/`DECRC`) are referring only to the cursor position. This is in conflict with gnome's `vte` issue I linked above, but we follow xterm so it's possible not relevant.

Either way, there was discussion for a way neovim could send a reset style in this open issue neovim/neovim#4396. I'd be interested if there's something we can do to help resolve this issue for people?

@justinmk left the following comment:

> Unless alacritty supports the feature described in this issue it's not possible to fix.

chrisduerr:
> This is in conflict with gnome's vte issue I linked above

That issue does not mention cursor shape at all. VTE behaves exactly like Alacritty.

PR #4892, opened 2021-03-12
fix clippy::inconsistent_struct_constructor

matthiaskrgr:
Make sure that struct fields in the instantiation are ordered the same as in the definition.

### DIFF
--- a/alacritty/src/display/wayland_theme.rs
+++ b/alacritty/src/display/wayland_theme.rs
@@ -8,8 +8,8 @@ const INACTIVE_OPACITY: u8 = 127;
 
 #[derive(Debug, Clone)]
 pub struct AlacrittyWaylandTheme {
-    pub background: ARGBColor,
     pub foreground: ARGBColor,
+    pub background: ARGBColor,
     pub dim_foreground: ARGBColor,
     pub hovered_close_icon: ARGBColor,
     pub hovered_maximize_icon: ARGBColor,
@@ -32,8 +32,8 @@ impl AlacrittyWaylandTheme {
             background,
             dim_foreground,
             hovered_close_icon,
-            hovered_minimize_icon,
             hovered_maximize_icon,
+            hovered_minimize_icon,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


chrisduerr:
Are these the only locations the lint complains about (so you rebuilt from clean and everything)? I feel like that lint should complain all over the place.

I think I'm generally **very much** against following this lint. I'm not sure about its exact definition but if it just always enforces the same order then that's definitely not something I'm interested in.

matthiaskrgr:
I checked with `clippy --all-targets --all-features -- --cap-lints warn ` and these were the only findings.. :thinking: 

The lint is supposed to catch patterns like this:
```rust
struct S {
  a: u8,
  b: u8,
}

let a = 1;
let b = 2;

let s1 = S { a, b };
let s2 = S { b, a };
````

One might think that s1 is `S { 1, 2 }` and s2 is `S { 2, 1 }` but that is in fact not the case, both are `{ 1, 2 }`

Note that the lint only warns about the short struct init.
If we explicitly write the fields down `let s3 = S { b: a, a: b };`, we can see that there was an (or was no) intend to swap the elements and there is no need to lint.

Sure, feel free to scratch the pr.

**Comment by chrisduerr** [comment]
If this is solely about the short form, where all fields are already initialized, then I'm fine with it. The lint description isn't entirely clear about that though (https://rust-lang.github.io/rust-clippy/master/index.html#inconsistent_struct_constructor).

File: alacritty/src/display/wayland_theme.rs [hunk 1]
```diff
@@ -28,12 +28,12 @@ impl AlacrittyWaylandTheme {
         dim_foreground.a = INACTIVE_OPACITY;
 
         Self {
-            foreground,
             background,
+            foreground,
```
chrisduerr, line 32: Can you please change the struct definition here? I don't like background before foreground.

chrisduerr (changes requested):

chrisduerr (approved):

matthiaskrgr:
I tried to improve the lint output and description a bit, I hope that's better: https://github.com/rust-lang/rust-clippy/pull/6892

chrisduerr:
I like it, thanks!

PR #4894, opened 2021-03-13
Fixed IME position being little earlier one in X11

moko256:
This fixes that IME position being little earlier one in X11.

I tested on X11 gnome (3.36.4-1ubuntu1~20.04.2) and mozc (2.23.2815.102+dfsg-8ubuntu1) on Ubuntu.

https://user-images.githubusercontent.com/16546008/111015999-c5dd4580-83ee-11eb-9f9f-001454a0a173.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Alacritty failing to start on X11 with invalid DPI reported by XRandr
 - Text selected after search without any match
 - Incorrect vi cursor position after leaving search
+- IME position being little earlier one on X11
 
 ### Removed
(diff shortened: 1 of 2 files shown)


chrisduerr:
What exactly is the goal of this PR? It seems like you're trying not to set the IME position when it has already been set, but I don't see how that helps? What is actually causing this issue?

nixpulvis:
Watching the video makes the goal clear to me at least, they're setting the IME overlay so it's closer to the cursor. However, I'm not sure about the code of this PR. I would think the `update_ime_position` function just needs changing, and that no new state on the `Window` is needed.

The IME's position should be based on the location of the cursor (both normal and search input).

chrisduerr:
Closing since author is unresponsive.

PR #4895, opened 2021-03-13
change windows default url_launcher

### LINKED ISSUE #2084: Click on https URLs wrongly opens File Explorer in Windows
I use Windows 10 x64 with the lates 0.2.7 release build.

Since 0.2.3 I regularily have the issue that a click on URLs starting with `https` sometimes will open the File Explorer instead of the Webbrowser. It might also have to do with tmux. Simple http URLs work, as long as they are on a single line (otherwise they get trimmed to the first linebreak) – when adressing this #1906 should be considered as well.

Since this clearly isn't intended behaviour I decided to post this as an issue.

This is relevant for  #1970

### LINKED ISSUE #4885: Change windows default url_launcher to cmd.exe /c start ''
When clicking a link on windows with the default url_launcher explorer opens in My Documents.

It would be better to use ``cmd.exe /c start "" $LINK/$PATH`` because it opens files and protocols with their default programs in windows (eg links with your browser and files with the program associated with that file ending).

The config for that would be:

```yaml
mouse:
  url:
    launcher:
      program: cmd.exe
      args: [ '/c', 'start', '']
```

I am not sure how to translate that into rust code.

fuzzmz:
When clicking on a link on windows we don't want to launch Explorer.

Using `cmd /c start "" $LINK/$PATH` opens files and protocols with
their default programs.

Fixes #4885.
Fixes #2084.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Alacritty failing to start on X11 with invalid DPI reported by XRandr
 - Text selected after search without any match
 - Incorrect vi cursor position after leaving search
+- Clicking on URLs on Windows incorrectly opens File Explorer
 
 ### Removed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -456,7 +456,7 @@
     # Default:
     #   - (macOS) open
     #   - (Linux/BSD) xdg-open
-    #   - (Windows) explorer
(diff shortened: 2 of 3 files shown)


2e0gzzdates (approved):

chrisduerr (changes requested):
This needs changes in the alacritty.yml and a changelog entry.

fuzzmz:
@chrisduerr, done.

chrisduerr (approved):

PR #4907, opened 2021-03-21
Switch to a uniform grid line indexing system

chrisduerr:
Previously Alacritty was using two different ways to reference lines in
the terminal. Either a usize, or a Line(usize). These indexing systems
both served different purposes, but made it difficult to reason about
logic involving these systems because of its inconsistency.

To resolve this issue, a single new Line(i32) type has been introduced.
All existing references to lines and points now rely on this definition
of a line.

The indexing starts at the top of the terminal region with the line 0,
which matches the line 1 used by escape sequences. Each line in the
history becomes increasingly negative and the bottommost line is equal
to the number of visible lines minus one.

Having a system which goes into the negatives allows following the
escape sequence's indexing system closely, while at the same time making
it trivial to implement `Ord` for points.

The Alacritty UI crate is the only place which has a different indexing
system, since rendering and input puts the zero line at the top of the
viewport, rather than the top of the terminal region.

All instances which refer to a number of lines/columns instead of just a
single Line/Column have also been changed to use a `usize` instead. This
way a Line/Column will always refer to a specific place in the grid and
no confusion is created by having a count of lines as a possible index
into the grid storage.


chrisduerr:
The benchmarks do not seem to indicate any performance regressions:

tmp

chrisduerr:
@nixpulvis Are you interested in testing this before merge?

nixpulvis:
@chrisduerr I can take a look over it a bit. I can't promise I'll be able to look at it all,

alacritty/src/display/content.rs
```diff
@@ -233,19 +237,21 @@ impl RenderableCell {
...
             let config_bg = colors.search.matches.background;
             Self::compute_cell_rgb(&mut fg_rgb, &mut bg_rgb, &mut bg_alpha, config_fg, config_bg);
-
-            is_match = true;
         }
 
+        // Convert cell point to viewport position.
+        let cell_point = cell.point;
+        let line = (cell_point.line + content.terminal_content.display_offset as i32).0 as usize;
+        let point = Point::new(line, cell_point.column);
```
nixpulvis, line 245: Can we make this into a helper function? I've read it twice already in the first 30 lines of this patch.

chrisduerr: I believe this exists twice in general. I've considered it but it really has no place in alacritty_terminal. And there's no other great spot that is shared between the two places where it is used.

PR #4919, opened 2021-03-25
Keep viewport in place when resizing

### LINKED ISSUE #4879: Resize does not keep viewport in place
It seems like there was a regression with resizing the viewport, where content would be incorrectly moved around when a long line was resized while inside of the scrollback history.

https://user-images.githubusercontent.com/8886672/110221220-d78c8d80-7ec2-11eb-9a8e-58885a582591.mp4

I have not yet bisected this, but I'm fairly sure this worked correctly previously without moving the viewport around.

st3iny:
Fixes #4879

First of all, I'd like to thank you for maintaining the most awesome terminal emulator out there. Keep it up :)

My code anchors the viewport at the last visible line. This corresponds to my intuition of how terminal emulators should work. If you want it anchored at the first visible line or have another approach at this issue feel free to leave a comment.

Your feedback is appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Incorrect vi cursor position after leaving search
 - Clicking on URLs on Windows incorrectly opens File Explorer
 - Incorrect underline cursor thickness on wide cell
+- Viewport moving around when resizing while scrolled into history
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


chrisduerr:
The code doesn't seem quite right and there are a lot of edgecases, but I'll try to take a look after #4907 is done.

chrisduerr (changes requested):
Could you rebase your PR on top of the latest master?

st3iny:
Done. Here is a little demonstration of my changes in action:


https://user-images.githubusercontent.com/1479486/113108898-c4d95000-9205-11eb-831b-5d3b4010e95f.mp4

chrisduerr (approved):
Looks good, thanks.

I've decided to fix a crash I've noticed due to my previous commit and added a changelog message, but the resizing itself worked just fine.

PR #4923, opened 2021-03-28
fixed the incorrect underline thickness on wide cell

### LINKED ISSUE #4922: Cursor thickness is increased incorrectly on wide cells
When a cursor like the underline is below a wide cell, its thickness is increased (I'd assume doubled).

stokhos:
fixes #4922

### DIFF
--- a/alacritty/src/display/cursor.rs
+++ b/alacritty/src/display/cursor.rs
@@ -22,11 +22,11 @@ impl IntoRects for RenderableCursor {
         let mut width = size_info.cell_width();
         let height = size_info.cell_height();
 
+        let thickness = (thickness * width as f32).round().max(1.);
+
         if self.is_wide() {
             width *= 2.;
         }
-
-        let thickness = (thickness * width as f32).round().max(1.);
 
         match self.shape() {
             CursorShape::Beam => beam(x, y, height, thickness, self.color()),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Text selected after search without any match
 - Incorrect vi cursor position after leaving search
 - Clicking on URLs on Windows incorrectly opens File Explorer
+- Incorrect underline cursor thickness on wide cell
 
 ### Removed
 


alacritty/src/display/cursor.rs
```diff
@@ -24,6 +24,7 @@ impl IntoRects for RenderableCursor {
 
         if self.is_wide() {
             width *= 2.;
+            thickness /= 2.;
```
chrisduerr, line 27: Why is the thickness doubled here in the first place? Looks like you're fixing a symptom here.

chrisduerr (changes requested):

alacritty/src/display/cursor.rs
```diff
@@ -26,7 +26,10 @@ impl IntoRects for RenderableCursor {
             width *= 2.;
         }
 
-        let thickness = (thickness * width as f32).round().max(1.);
+        let mut thickness = (thickness * width as f32).round().max(1.);
+        if self.is_wide() {
+            thickness /= 2.;
+        }
```
chrisduerr, line 32: I feel like there's a little bit of a lack of understanding here when it comes to this issue.

Try moving the line 29 above the `is_wide` check. I'm pretty sure just reordering these statements should resolve this issue.

Alternatively the `cell_width` could be used instead of the `width` variable, to make things a bit more explicit. But I personally think just reusing the existing `width` should be fine.

chrisduerr (changes requested):

alacritty/src/display/cursor.rs
stokhos: I see what you mean.

alacritty/src/display/cursor.rs
stokhos: > Try moving the line 29 above the is_wide check. I'm pretty sure just reordering these statements should resolve this issue.

Yes. This is the best way. :+1:

chrisduerr (changes requested):
Since I believe this is an issue with the latest release, could you please add a changelog entry too?

stokhos:
> Since I believe this is an issue with the latest release, could you please add a changelog entry too?

Sure, how can I add this changelog, or where should I put it. Cause this is my first commit in such a large project, I'm totally lost.

chrisduerr:
Correct, see https://github.com/alacritty/alacritty/blob/master/CONTRIBUTING.md#documentation.

chrisduerr (approved):

PR #4947, opened 2021-04-05
Fix automatic scrolling on resize

### LINKED ISSUE #4944: Automatic scrolling breaks on resize
Since 58cae8f2edf44e6bf0721543c61edb57bdf7d776 there has been an issue where the viewport would get locked into place when the last line is longer than the new viewport width and thus broken onto a new line.

Since the viewport is kept in place, the new line is put below the viewport increasing the display offset to 1. But since we've been following the terminal output we now suddenly won't do that anymore. To a user it might look like the terminal suddenly "froze".

I think we should add a special case where we keep everything stuck at the bottom of the viewport if we're already at the bottom, just so we keep following all new output.

@st3iny you wrote the original PR, are you interested in looking into this?

st3iny:
Fixes #4944 

Only rotate viewport if it isn't scrolled. Please give this one a good test.

### DIFF
--- a/alacritty_terminal/src/grid/resize.rs
+++ b/alacritty_terminal/src/grid/resize.rs
@@ -187,7 +187,7 @@ impl<T: GridCell + Default + PartialEq + Clone> Grid<T> {
 
                 cursor_line_delta += line_delta.0 as usize;
             } else if row.is_clear() {
-                if i <= self.display_offset {
+                if i < self.display_offset {
                     // Since we removed a line, rotate down the viewport.
                     self.display_offset = self.display_offset.saturating_sub(1);
                 }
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
I've played around with this a little bit and honestly I think it might be better to just do `i < self.display_offset`.

It just feels somewhat more natural to see the new line visible when breaking when it is at the bottom, otherwise you're almost unsure if it's working properly or not.

The same thing applies to `grow_columns` too.

st3iny:
Thanks for the feedback. Resizing looks less confusing now.

alacritty_terminal/src/grid/resize.rs
```diff
@@ -355,7 +355,7 @@ impl<T: GridCell + Default + PartialEq + Clone> Grid<T> {
                     }
                     row = Row::from_vec(wrapped, occ);
 
-                    if i <= self.display_offset {
```
chrisduerr, line 358: I don't think the zero check is necessary anymore with this? Since both are a usize it's impossible for i to ever be smaller than the display offset if it is zero.

chrisduerr (changes requested):

chrisduerr (approved):
Thanks.

PR #4962, opened 2021-04-11
Add vi/mouse hint highlighting support

chrisduerr:
This patch removes the old url highlighting code and replaces it with a
new implementation making use of hints as sources for finding matches in
the terminal.


chrisduerr:
tmp

Performance seems good.

cmonr:
Was there a particular reason that this change was needed?

PR #4967, opened 2021-04-14
Add hint action for moving the vi cursor

### LINKED ISSUE #4319: Feature: easymotion like jumps in vi mode
It would be great to have [easymotion](https://github.com/easymotion/vim-easymotion) like functionality in vi mode for quick jumps to place where v selection can start and later on end.

chrisduerr:
Fixes #4319.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -475,6 +475,8 @@
   #       Paste the hint's text to the terminal or search.
   #   - Select
   #       Select the hint's text.
+  #   - MoveViModeCursor
+  #       Move the vi mode cursor to the beginning of the hint.
   #enabled:
   # - regex: "(mailto:|gemini:|gopher:|https:|http:|news:|file:|git:|ssh:|ftp:)\
   #           [^\u0000-\u001F\u007F-\u009F<>\" {-}\\^⟨⟩`]+"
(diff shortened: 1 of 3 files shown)


chrisduerr:
This is trivial to support, so I feel like it definitely might help some, but it's questionable if people are willing to define completely separate hints just for motion in vi mode. Since currently it's not possible to have mode-dependent hint actions or trigger multiple actions for the same hint.

I'd be curious how you feel about this @nixpulvis.

nixpulvis:
I'm wondering if it doesn't make sense to enter vi-mode when this action is triggered. Since it's effectively a no-op outside of vi-mode otherwise. I'm also curious if adding a `mode` selector to the hints (akin to other bindings) makes sense.

chrisduerr:
> I'm wondering if it doesn't make sense to enter vi-mode when this action is triggered.

I like that idea. Seems like a good solution.

> I'm also curious if adding a mode selector to the hints (akin to other bindings) makes sense.

I'm kinda wondering the same thing, but at the same time I **really** don't want to introduce this complexity to hints too. It's just so much to wrap your head around, but I suppose it makes sense to do different things for hints while in vi mode? But I think initially I'd just want to suggest using different key bindings until there is some more feedback (so maybe post 0.8?).

nixpulvis:
In some ways it's actually less complexity, since bindings then seem more consistent in what you can specify across hints and normal bindings.

PR #4758, opened 2021-02-03
Add note about signing and notarization to README

nixpulvis:
Just figured I'd put this in the README to possible help our users understand our stance on the issue. Feel free to close if you'd rather not add anything.

fixes #1896
fixes #3448
fixes #4529
fixes #4673
fixes #4731

Issue #1896, opened 2018-12-13
macOS complains about an "unidentified developer"


jsatk:
Is there a way to mark this app as from an identified developer with macOS so I don't get the scary warning every time I launch this app after a restart of my computer?

chrisduerr:
I'm not familiar with this, but every release is signed with PGP so that shouldn't be an issue.

Is this an apple-specific thing that might need to be specified in the app bundle?

jsatk:
Yep!

Info here: https://stackoverflow.com/questions/30297665/how-to-sign-my-app-as-identified-apple-developer#30477278

szippy:
If you set yourself up as a nonprofit who distributes free software, Apple will allow you to apply for an exception according to this article. You have to apply for a DUNS number, but can do that if you're a sole-prop. Here's the article apple has on their requirements. 

It might be a good alternative to paying a bunch of cash to Apple :) 

https://developer.apple.com/support/membership-fee-waiver/

chrisduerr:
Well paying a bunch of cash to Apple is not an option. Setting up a nonprofit and applying is also not an option, since that's way too much work.

Apple users will just have to deal with the restrictions Apple imposes on them.

directionless:
I respect y'alls choice of _not_ signing things, it does come across as a bit hostile to osx platform users. Anyhow, I wanted to leave a couple notes about the process as I understand it today. If y'all wanted, I'd be happy to try to PR them into your Makefile (or similar).

I think the apple developer program is $99 USD/year.

You can codesign outside of the Xcode environment. You'd want to do something like:
```
codesign --force \
             --entitlements entitlements \
             --options runtime \
             --timestamp \
             -s "$CODESIGN_IDENTITY" \
             -v \
            target/release/osx/Alacritty.app
```

You may also need to notarize it (where apple is going for code signing). That's a similar `xcrun altool` but it's primarily for distribution packages. 

Anyhow, happy to write shell code to help. Though I'm not sure there's interest.

kchibisov:
> I think the apple developer program is $99 USD/year.

This is the main problem here, and as you can read in above comments. The problem is not 99$/year it's 99$ a year to apple and strange hassle. Thank you for your instructions though.

chrisduerr:
> it does come across as a bit hostile to osx platform users

I think it's a bit hostile of Apple to request a significant amount of money for developers to write free software for their platform.

> I think the apple developer program is $99 USD/year.

[...]

PR #4981, opened 2021-04-16
Note about working around macos signing requirements

iloveitaly:
I've read the [previous](https://github.com/alacritty/alacritty/pull/4758) [discussion](https://github.com/alacritty/alacritty/issues/1896) on this issue and agree that it is absolutely crazy for Apple to require a $99 subscription to distribute free software.

There's a simple workaround available to remove Alacritty from the MacOS security check. This will save MacOS users some time when they are installing Alacritty.

Thanks for the amazing work on this project!

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,14 @@ Prebuilt binaries for macOS and Windows can also be downloaded from the
 For everyone else, the detailed instructions to install Alacritty can be found
 [here](INSTALL.md).
 
+### MacOS
+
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
You're making the assumption here that the user installed the application with brew, which will just confuse people that didn't do that.

Generally I'd expect people to know how to install applications on their system. So I'm not sure how helpful this actually is.

The number of people questioning how to get around this is actually extremely slim, I don't remember anyone bringing this up within the last year or so. The only thing people complain about is that Alacritty isn't signed, not that they don't know how to run unsigned apps.

So in this form I think this does more harm than good.

PR #5004, opened 2021-04-25
Fix overwriting of fullwidth character

### LINKED ISSUE #3726: Properly overwrite wide chars
Currently when the cursor is positioned above a wide char, or a wide char spacer, the underlying cell is simply overwritten without considering the previous fullwidth character content.

This can be reproduced with the following two escapes:
```
echo " 🥱🥱\b\b\ba"
```
This writes a character in the spacer of the first full width character, leading to overlapping glyphs.

```
echo " 🥱🥱\b\b\b\ba"
```
This overwrites the first full width character, which first seems to render correctly, however selecting the leftover wide char spacer will lead to the selection expanding across the other full width character incorrectly.

I'm not sure if this is ever a problem in practice, since usually content is cleared before it is rewritten to ensure correct behavior, but Alacritty's current behavior is probably the farthest away from correct in any of the terminal emulators I've tested.

The proper solution would likely to overwrite the entire fullwidth glyph when a character is written in either of its two cells, but there is one terminal that behaves differently (Kitty), which just doesn't write the character to the grid at all if it is written to the location of the spacer.

A proper resolution would likely not be very complicated, however certainly have some negative performance impact since I'd expect it to require two additional conditionals in the `input` function whenever any character is written. If there's a solution that can circumvent these in the hot path that would probably be ideal.

chrisduerr:
Fixes #3726.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -798,6 +798,14 @@ impl<T> Term<T> {
         cursor_cell.bg = bg;
         cursor_cell.flags = flags;
     }
+
+    /// Write `c` at the cell with a temporary flag set for the template cell.
+    #[inline(always)]
+    fn write_at_cursor_with_flag(&mut self, c: char, flag: Flags) {
+        self.grid.cursor.template.flags.insert(flag);
+        self.write_at_cursor(c);
+        self.grid.cursor.template.flags.remove(flag);
+    }
 }
 
 impl<T> Dimensions for Term<T> {
@@ -830,18 +838,18 @@ impl<T: EventListener> Handler for Term<T> {
         // Handle zero-width characters.
         if width == 0 {
             // Get previous column.
-            let mut col = self.grid.cursor.point.column.0;
+            let mut column = self.grid.cursor.point.column;
             if !self.grid.cursor.input_needs_wrap {
-                col = col.saturating_sub(1);
+                column.0 = column.saturating_sub(1);
             }
 
             // Put zerowidth characters over first fullwidth character cell.
             let line = self.grid.cursor.point.line;
-            if self.grid[line][Column(col)].flags.contains(Flags::WIDE_CHAR_SPACER) {
-                col = col.saturating_sub(1);
+            if self.grid[line][column].flags.contains(Flags::WIDE_CHAR_SPACER) {
+                column.0 = column.saturating_sub(1);
             }
 
-            self.grid[line][Column(col)].push_zerowidth(c);
+            self.grid[line][column].push_zerowidth(c);
             return;
         }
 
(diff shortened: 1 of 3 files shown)


chrisduerr:
Performance of the current version is poor, the only version that doesn't introduce a significant regression I've found was applying only a minimal fix for deleting the characters without clearing any of the flags:

tmp

I'll have to investigate and maybe `perf` it, chances are there are some optimization oddities going on. In theory the intersection should be the same speed as the contains, but I've been getting mixed results.

chrisduerr:
This should basically be good now, I've managed to match master's performance:

tmp

The benchmarks for `write_with_flag_fun` are a simple style improvement which moves the three instances where we write widechar/spacers to a generic function that takes the `c` and `Flags` and then inserts the flag, calls `write_at_cursor` and removes the flag automatically.

While this is a style enhancement, I'm somewhat hesitant to merge it since it has some unreliable performance characteristics (see `missing` benchmark for example).

nixpulvis:
Here's what I'm seeing for performance on my ThinkPad X1 Carbon (5th)

summary

chrisduerr:
I'm not sure about the reliability of your platform, but that seems within what I would expect.

nixpulvis:
I could test the "fun" patch as well if it would be helpful to you?

chrisduerr:
Feel free, though I don't really think it's worth it tbh.

alacritty_terminal/src/term/mod.rs
```diff
@@ -860,14 +878,18 @@ impl<T: EventListener> Handler for Term<T> {
...
+            self.grid.cursor.template.flags.insert(Flags::WIDE_CHAR);
+            self.write_at_cursor(c);
+            self.grid.cursor.template.flags.remove(Flags::WIDE_CHAR);
 
             // Write spacer to cell following the wide glyph.
             self.grid.cursor.point.column += 1;
-            self.write_at_cursor(' ').flags.insert(Flags::WIDE_CHAR_SPACER);
+            self.grid.cursor.template.flags.insert(Flags::WIDE_CHAR_SPACER);
+            self.write_at_cursor(' ');
+            self.grid.cursor.template.flags.remove(Flags::WIDE_CHAR_SPACER);
```
nixpulvis, line 889: I don't completely understand why you moved this to the template, but I'm completely fine assuming it was done for a good reason. Let me know if you care to explain.

nixpulvis (approved):
I tested this out a bit, and it looks like it's working as expected.

alacritty_terminal/src/term/mod.rs
chrisduerr: The `write_at_cursor` function previously returned a mutable reference. That made borrowing semantics complicated since branching while returning a mutable reference prevents you from reasigning that variable since borrowck assumes that is incorrect (which it is not).

So I found this alternative solution which allows us to do the same amount of work (except for one more branch) in our hottest path.

PR #5009, opened 2021-04-26
Update vi-mode `Open` Config Docs for Hints

### LINKED ISSUE #5005: Url.Launcher Reference In Config File
[Here](https://github.com/alacritty/alacritty/blob/master/alacritty.yml#L596) in the example alacritty.yml there is a reference to url.launcher. Is this intended as it seems url.launcher was removed in a previous commit?

nixpulvis:
Fixes #5005

I'm not sure I love how the `Open` action for vi-mode is dependent on a configuration named `mouse.enabled`, but I'm also not sure what might make more sense. Fundamentally there are regex matches, which can be either A) highlighted by the mouse and activated with this `Open` command, and B) activated without highlighting by keyboard hint prefix input. To me, B is what really lines up with "hints", and A could be a more general matching feature. Still, I don't know if thinking of things like this would help.

Perhaps a better way to clear up some potential confusion would be to rename `Open` to `Click`, thus making it more clear how this works.

@chrisduerr thoughts?

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -228,7 +228,7 @@
     #  background: '#c5c8c6'
     #  foreground: '#1d1f21'
 
-  # Keyboard hints
+  # Keyboard regex hints
   #hints:
     # Fist character in the hint label
     #
@@ -592,8 +592,8 @@
 # - Vi mode exclusive actions:
 #
 #   - Open
-#       Open URLs at the cursor location with the launcher configured in
-#       `url.launcher`.
+#       Perform the action of the first matching hint under the vi mode cursor
+#       with `mouse.enabled` set to `true`.
 #   - ToggleNormalSelection
 #   - ToggleLineSelection
 #   - ToggleBlockSelection
--- a/alacritty/src/display/hint.rs
+++ b/alacritty/src/display/hint.rs
@@ -165,7 +165,7 @@ impl HintState {
 }
 
 /// Hint match which was selected by the user.
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 pub struct HintMatch {
     /// Action for handling the text.
     pub action: HintAction,


alacritty/src/display/hint.rs
```diff
@@ -165,7 +165,7 @@ impl HintState {
 }
 
 /// Hint match which was selected by the user.
-#[derive(Clone)]
+#[derive(Debug, Clone)]
```
chrisduerr, line 168: Did you actually need `Debug` on this for testing, or did you just add it for different reasons?

alacritty.yml
```diff
@@ -592,8 +592,8 @@
 # - Vi mode exclusive actions:
 #
 #   - Open
-#       Open URLs at the cursor location with the launcher configured in
-#       `url.launcher`.
+#       Performs the action of the first matching mouse enabled hint under the
+#       cursor, if any.
```
chrisduerr, line 596: I don't think the `if any` at the end contributes to this and I'm not entirely certain that `mouse enabled` should be used as a verb, since people might not be aware of this setting. These two issues are minor though.

What I'd definitely like to point out is that we're talking about the vi mode cursor here, not the mouse or terminal cursor. Especially if we keep it at `mouse.enabled` that could be confusing.

chrisduerr (changes requested):

alacritty/src/display/hint.rs
nixpulvis: Yea, I found myself trying to print this out to understand things a bit better. In general, I find having `Debug` implementations is a good idea.

Why wouldn't we have them on basically everything?

chrisduerr: > Yea, I found myself trying to print this out to understand things a bit better. In general, I find having Debug implementations is a good idea.

Fair enough, I'm just asking if you added it out of completeness or because you actually used it. Generally my motto with `Debug` derives (at least for internal stuff) is "add only". If I ever have a reason to add it, I keep it. So if you found a reason to print this out, I have no issue with adding this. But I'm against adding things to internal stuff nobody ever had any interest in debugging.

> Why wouldn't we have them on basically everything?

Just a matter of not introducing code nobody needs.

alacritty.yml
nixpulvis: Pushed up a rewording.

alacritty.yml
```diff
@@ -592,8 +592,8 @@
 # - Vi mode exclusive actions:
 #
 #   - Open
-#       Open URLs at the cursor location with the launcher configured in
-#       `url.launcher`.
+#       Perform the action of the first matching hint, with `mouse.enabled`,
+#       under the cursor.
```
chrisduerr, line 596: You've changed everything minor I didn't mind much and skipped the thing I was actually concerned about. :smile: 

Also I'm not sure just stating `with mouse.enabled` is sufficient, since technically that could be any of the two values. I'd maybe suggest the following:

```suggestion
#       Perform the action of the first matching hint under the vi mode cursor with `mouse.enabled` set to `true`.
```

Feel free to make suggestions yourself though please.

chrisduerr (changes requested):

alacritty.yml
nixpulvis: Oh god damn it... I had added it in one rewording, then managed to delete it in my frantic search for a way to write this.

nixpulvis: > Also I'm not sure just stating with `mouse.enabled` is sufficient, since technically that could be any of the two values.

I'm quite confident anyone can manage to understand that this implies a value of `true`, frankly.

chrisduerr: I'm mostly just trying to be cautious, since we're using up two lines already anyways. With yaml there always has to be a value associated, so just saying `mouse.enabled` has to be there technically doesn't document what that value is supposed to be.

nixpulvis:
OK, if this looks good I'll rebase and merge.

alacritty.yml
```diff
@@ -592,8 +592,7 @@
 # - Vi mode exclusive actions:
 #
 #   - Open
-#       Perform the action of the first matching hint, with `mouse.enabled`,
-#       under the cursor.
+#       Perform the action of the first matching hint under the vi mode cursor with `mouse.enabled` set to `true`.
```
chrisduerr: Could you wrap this for me? IIRC it should all be wrapped at `80`.

chrisduerr (changes requested):

alacritty.yml
nixpulvis: I just noticed that myself, and already fixed it.

chrisduerr (approved):
Feel free to merge at your own discretion.

PR #5016, opened 2021-04-28
Add terminal size popup when resizing the window

akmin04:
Shows a small popup with the window dimensions (columns x lines) in the center of the window when resizing, similar Ubuntu terminal.

# Changes
- Draw popup in the center of the display when a `WindowEvent::Resized` is received. The popup persists for `resize_popup_duration` after the user stops resizing.
- `resize_popup_duration` config under `window` specifying the duration of the popup (setting to 0 will disable the popup).

# Questions
- Should the popup be enabled/disabled by default? (currently set to enabled/500 ms)
- Any specific colors to use? (currently uses primary foreground/background swapped)

# Example

https://user-images.githubusercontent.com/26666500/116330083-99547180-a781-11eb-96e5-e2c22cb92cc3.mov

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -80,6 +80,10 @@
   # Allow terminal applications to change Alacritty's window title.
   #dynamic_title: true
 
+  # Duration of window resize popup in milliseconds.
+  # Specifying '0' will disable the popup.
+  #resize_popup_duration: 500
+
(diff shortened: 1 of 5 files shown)


chrisduerr (changes requested):
I'm not interested in adding this to Alacritty.

PR #5019, opened 2021-04-28
Clean up config files loaded log message

nixpulvis:
This is extremely minor, but I'm always a fan of cleaning up log messages to help reading through them. This removes a slightly confusing line in the case a user doesn't have a config file, and the default is used. In this case, another log message will print as follows:

```
[INFO ] [alacritty_config_derive] No config file found; using default
```

### DIFF
--- a/alacritty/src/main.rs
+++ b/alacritty/src/main.rs
@@ -235,10 +235,11 @@ fn run(
 }
 
 fn log_config_path(config: &Config) {
-    let mut msg = String::from("Configuration files loaded from:");
-    for path in &config.ui_config.config_paths {
-        msg.push_str(&format!("\n  {:?}", path.display()));
+    if !config.ui_config.config_paths.is_empty() {
+        let mut msg = String::from("Configuration files loaded from:");
+        for path in &config.ui_config.config_paths {
+            msg.push_str(&format!("\n- {}", path.display()));
+        }
+        info!("{}", msg);
     }
-
-    info!("{}", msg);
 }
--- a/alacritty/src/main.rs
+++ b/alacritty/src/main.rs
@@ -235,11 +235,14 @@ fn run(
 }
 
 fn log_config_path(config: &Config) {
-    if !config.ui_config.config_paths.is_empty() {
-        let mut msg = String::from("Configuration files loaded from:");
-        for path in &config.ui_config.config_paths {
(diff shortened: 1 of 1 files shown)


alacritty/src/main.rs
```diff
@@ -235,10 +235,11 @@ fn run(
 }
 
 fn log_config_path(config: &Config) {
-    let mut msg = String::from("Configuration files loaded from:");
-    for path in &config.ui_config.config_paths {
-        msg.push_str(&format!("\n  {:?}", path.display()));
+    if !config.ui_config.config_paths.is_empty() {
```
chrisduerr, line 238: I'd advocate for an early return here.

chrisduerr (commented):
Not sure if we should add a changelog message for this. It seems rather minor so we're probably good without it.

nixpulvis:
> Not sure if we should add a changelog message for this. It seems rather minor so we're probably good without it.

That was my feeling as well.

alacritty/src/main.rs
```diff
@@ -235,9 +235,13 @@ fn run(
...
 
 fn log_config_path(config: &Config) {
+    if config.ui_config.config_paths.is_empty() {
+        return
+    }
+
     let mut msg = String::from("Configuration files loaded from:");
     for path in &config.ui_config.config_paths {
-        msg.push_str(&format!("\n  {:?}", path.display()));
+        msg.push_str(&format!("\n- {}", path.display()));
```
chrisduerr, line 244: Why did you change this?

nixpulvis: > 2. Print out the paths with Display in a YAML-like format.

It was mostly stylistic, sorry if there was some history here I'm stepping on.

chrisduerr: By stripping the debug option it becomes more difficult to see things like excessive whitespace, so that's no good.

And I'm also not a fan of the `-` at the beginning. I much prefer things actually being indented when they belong to the previous statement, since that makes it much simpler to skip over them if you don't care about it.

nixpulvis: That's a fair point about irritating file paths with whitespace (I had a feeling you might bring this up). The indentation is already pretty deep due to the way `info!` messages work, so that's actually part of why I changed this.

chrisduerr: I don't see any problem with that.

PR #5025, opened 2021-04-29
Stop basing focused match color on matched colors

### LINKED ISSUE #5022: Search mode focused_match using incorrect CellBackground/CellForeground colors
### System

OS: Linux
Version: 0.7.2
Linux/BSD: X11, i3wm
Tested with the default config.
```
colors:
  search:
      matches:
        foreground: '#000000'
        background: '#ffffff'
      focused_match:
        foreground: CellBackground
        background: CellForeground
```
https://user-images.githubusercontent.com/33134232/116493381-798d7e00-a853-11eb-9a75-ed63188a5623.mp4

While in search mode, the focused match is expected to use the CellBackground/CellForeground colors like when selecting text. However, alacritty seems to be using the whatever colors are set under `search.matches` as the cell background/foreground.

nixpulvis:
Pending discussion and CHANGELOG entry.

fixes #5022

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -221,8 +221,8 @@
     #  foreground: '#000000'
     #  background: '#ffffff'
     #focused_match:
-    #  foreground: CellBackground
-    #  background: CellForeground
+    #  foreground: '#ffffff'
+    #  background: '#000000'
 
     #bar:
     #  background: '#c5c8c6'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - The vi mode cursor is now created in the top-left if the terminal cursor is invisible
+- Focused search match will use cell instead of match colors for CellForeground/CellBackground
 
 ### Fixed
 
--- a/alacritty/src/config/color.rs
+++ b/alacritty/src/config/color.rs
@@ -122,9 +122,24 @@ impl Default for InvertedCellColors {
 
 #[derive(ConfigDeserialize, Debug, Copy, Clone, Default, PartialEq, Eq)]
 pub struct SearchColors {
-    pub focused_match: InvertedCellColors,
+    pub focused_match: FocusedMatchColors,
     pub matches: MatchColors,
     bar: BarColors,
+}
+
+#[derive(ConfigDeserialize, Debug, Copy, Clone, PartialEq, Eq)]
+pub struct FocusedMatchColors {
+    pub foreground: CellRgb,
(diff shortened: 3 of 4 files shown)


chrisduerr (changes requested):
Fix comments, add changelog, change defaults of `focused_match` in yaml and code. Then it should be good to go.

nixpulvis:
I really don't like the way this looks by default now however:

screenshot-2021-04-29T17:00:36Z

Trying to determine which of these is focused is quite hard.

chrisduerr:
> I really don't like the way this looks by default now however:

As I've said, the defaults need to be changed they're currently incorrect. And if you're just talking about 'your defaults', that really doesn't matter.

nixpulvis:
I'm thinking this might be a good default, with the matched text background in red, and the focused match inverted so-as to be consistent with selections:

Colored text focused (first `foo`):
screenshot-2021-04-29T17:23:23Z

Default foreground text focused (3rd `foo`):
screenshot-2021-04-29T17:23:33Z

chrisduerr:
The focused match color should be the inverse of the normal text color.

nixpulvis:
> The focused match color should be the inverse of the normal text color.

Isn't it? Or do you mean always like in the second image? If so, why? Isn't it nice that it's consistent with the way selections look in general?

chrisduerr:
It's nice to keep things simple. There's no need to over-complicate this. The default config should behave the same way and the best way to do this is to just invert the normal match colors manually for the focused match.

alacritty.yml
```diff
@@ -218,11 +218,11 @@
     # Allowed values are CellForeground/CellBackground, which reference the
     # affected cell, or hexadecimal colors like #ff00ff.
     #matches:
-    #  foreground: '#000000'
+    #  foreground: '#ff0000'
```
chrisduerr, line 221: I've told you that I don't want to change this, didn't I? I'm not sure how to make myself more clear on how I'd suggest to resolve this.

chrisduerr (changes requested):

lockheadone:
https://github.com/alacritty/alacritty/pull/5025/files/70b42ab429dc103cb940809c530c0e04c38f3782

nixpulvis:
> https://github.com/alacritty/alacritty/pull/5025/files/70b42ab429dc103cb940809c530c0e04c38f3782

Was there a point to this link?

alacritty.yml
chrisduerr: There is zero reason to change the defaults in this PR. The requested change is completely unrelated to that. I'm not sure why you're derailing this PR with an unrelated change.

chrisduerr (approved):

PR #5026, opened 2021-04-29
Fix highlighting multiple hints in the same line

### LINKED ISSUE #5010: Mouse enabled hint only highlights first match
With this simple hint:

```yaml
hints:
  enabled:
    - regex: foo
      action: Select
      mouse:
        enabled: true
```

And the following text in the grid:

```
foo foo
```

Only  the first `foo` will be highlighted on mouseover.

This does not seem to be an issue for the keyboard hints.

chrisduerr:
Fixes #5010.

### DIFF
--- a/alacritty/src/display/hint.rs
+++ b/alacritty/src/display/hint.rs
@@ -3,8 +3,8 @@ use std::cmp::{max, min};
 use glutin::event::ModifiersState;
 
 use alacritty_terminal::grid::BidirectionalIterator;
-use alacritty_terminal::index::{Boundary, Point};
-use alacritty_terminal::term::search::{Match, RegexSearch};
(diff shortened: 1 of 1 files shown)


alacritty/src/display/hint.rs
```diff
@@ -266,11 +266,12 @@ pub fn highlighted_at<T>(
             let mut end = term.line_search_right(point);
             end.line = min(end.line, point.line + MAX_SEARCH_LINES);
 
-            // Function to verify if the specified point is inside the match.
-            let at_point = |rm: &Match| *rm.start() <= point && *rm.end() >= point;
+            // Function to verify that the specified point is inside the match.
+            let at_point = |rm: &Match| *rm.end() >= point && *rm.start() <= point;
```
chrisduerr, line 270: Swapping these should improve performance.

PR #5027, opened 2021-04-29
Add support for magnet URIs

rnestler:


chrisduerr (commented):
Which terminal application emits these URIs?

rnestler:
> Which terminal application emits these URIs?

My very own https://github.com/rnestler/rust-torrent-search/. I use it to search for torrents and then opening them in transmission.

chrisduerr (approved):

nixpulvis (approved):
Seems like a fine idea, I might just add this to my own regex as well.

PR #5034, opened 2021-05-01
Bump version to 0.8.0-rc1

chrisduerr:


alacritty/Cargo.toml
```diff
@@ -10,7 +10,7 @@ edition = "2018"
 
 [dependencies.alacritty_terminal]
 path = "../alacritty_terminal"
-version = "0.13.0-dev"
+version = "0.14.0-rc1"
```
chrisduerr, line 13: I noticed that we must have forgotten to bump the `alacritty_terminal` version while the 0.7.2 release was made. So I'd probably update this to 0.14.1 directly once the 0.8.0 release is made, since it's now too late anyways might as well wait.

Issue #5067, opened 2021-05-07
Crash when selecting text while selecting rapid text output


bradfitz:
Crash while selecting moving text.

### System

OS: macOS (Big Sur 11.2.3)
Version: alacritty 0.8.0-rc3 (706f476)

### Logs

```
$ ./Alacritty.app/Contents/MacOS/alacritty 
thread 'main' panicked at 'index out of bounds: the len is 10046 but the index is 10236', /Users/runner/work/alacritty/alacritty/alacritty_terminal/src/term/mod.rs:395:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Video

I ran `find . | perl -npe 's/\w/x/g'` and selected text while it spewed. It very reliably crashes:

https://user-images.githubusercontent.com/2621/117462177-3aba9080-af03-11eb-8099-a70a725a4c32.mov

chrisduerr:
Great find, thanks for testing the RC. I can reproduce, should probably be an easy fix.

PR #5071, opened 2021-05-07
Add Ctrl+c for exiting hint mode.

absorber:
Previously only ESC was used for quitting hint mode, but now Ctrl+c should work too.

### DIFF
--- a/alacritty/src/display/hint.rs
+++ b/alacritty/src/display/hint.rs
@@ -117,6 +117,8 @@ impl HintState {
             },
             // Cancel hint highlighting on ESC.
             '\x1b' => self.stop(),
+            // Do the same but with Ctrl+c
+            '\x3' => self.stop(),
             _ => (),
         }
 
--- a/alacritty/src/display/hint.rs
+++ b/alacritty/src/display/hint.rs
@@ -116,9 +116,7 @@ impl HintState {
                 self.keys.pop();
             },
             // Cancel hint highlighting on ESC/Ctrl+c.
-            '\x1b' | '\x03' => {
-                self.stop();
-            },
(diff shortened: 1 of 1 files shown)


alacritty/src/display/hint.rs
```diff
@@ -117,6 +117,8 @@ impl HintState {
             },
             // Cancel hint highlighting on ESC.
             '\x1b' => self.stop(),
+            // Do the same but with Ctrl+c
+            '\x3' => self.stop(),
```
chrisduerr, line 121: These two statements can be combined using `|`, see the lines just above this. The comment should then also be updated to say `ESC/Ctrl+C` of course.

Also the same as the backspace key (`\x08`), the hexadecimal needs to be padded with a `0`.

chrisduerr (changes requested):
No changelog message necessary, I'll cherry pick this into the next RC together with a fix for #5067.

alacritty/src/display/hint.rs
```diff
@@ -115,10 +115,10 @@ impl HintState {
...
                 self.keys.pop();
             },
-            // Cancel hint highlighting on ESC.
-            '\x1b' => self.stop(),
-            // Do the same but with Ctrl+c
-            '\x3' => self.stop(),
+            // Cancel hint highlighting on ESC/Ctrl+c.
+            '\x1b' | '\x03' => {
+                self.stop();
+            },
```
chrisduerr, line 121: Could you inline this? There's no need to have a dedicated block.

chrisduerr (changes requested):

chrisduerr (approved):
Looks good, thanks for the PR. Could you rebase this on top of the latest master?

Issue #567, opened 2017-05-13
intel gpu support


OliverLeitner:
is there any plans for supporting intel gpu's at this point?

it lets me compile it on my laptop with intellinux driver, but it doesnt run, missing ES, no support for GSLS 3.x...

`drm:intel_pipe_update_end [i915]] *ERROR* Atomic update failure on pipe A (start=9620131 end=9620132) time 618 us, min 763, max 767, scanline start 760, end 9`

maximbaz:
FYI the alacritty compiles and runs fine on my laptop with Intel HD Graphics 4400, so you probably should a be a bit more specific about your GPU.

OliverLeitner:
oh, thanks for the info, if it helps, heres some better info on the "card":

```
00:00.0 Host bridge [0600]: Intel Corporation Mobile GME965/GLE960 Memory Controller Hub [8086:2a10] (rev 0c)
00:02.0 VGA compatible controller [0300]: Intel Corporation Mobile GME965/GLE960 Integrated Graphics Controller [8086:2a12] (rev 0c)
00:02.1 Display controller [0380]: Intel Corporation Mobile GME965/GLE960 Integrated Graphics Controller [8086:2a13] (rev 0c)
```

the xorg driver im using is the intel "proprietary" one...

https://01.org/linuxgraphics/downloads

the kernel is a intel-drm development build from ubuntu:
http://kernel.ubuntu.com/~kernel-ppa/mainline/drm-intel-next/

`Linux lappy 4.11.0-997-lowlatency #201704182244 SMP PREEMPT Wed Apr 19 02:49:15 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`

the modules loaded:

```
i915                 1515520  3
i2c_algo_bit           16384  1 i915
drm_kms_helper        163840  1 i915
drm                   356352  5 i915,drm_kms_helper
video                  40960  1 i915
```
intel_gpu_top shows its being utilized

the xorg device section:

```
Section "Device"
   Identifier  "Intel Graphics"
   Driver      "intel"
Option      "AccelMethod" "sna"
Option      "TearFree" "true"
Option      "SwapbuffersWait" "true"
Option      "VSync" "true"
Option      "DRI" "true"
VideoRam        461632   # sometimes (before i updated the bios, it were useful)
EndSection
```

the laptop is a compaq 610 (compaq branded hp laptop...)

OliverLeitner:
ok, i think i found a way "around" the problem...
[...]

PR #5086, opened 2021-05-09
Enable alacritty to run on OpenGL 2.1 when needed extensions are present

FrostKiwi:
This PR lowers the GLSL version of Alacritty's shader files from #version 330 core to #version 120 and instead requires the needed extensions. There are two upsides to this

1. Allows Alacritty to run on OpenGL 2.1 machines, where the needed extensions are present.
2. If the needed extensions are not present, this gives a clearer error message, as compared to just "GLSL 330 is not supported"

In addition, the GLSL Version is checked. By default, the normal Shader files are used. However, if GLSL version 1.20 is detected, create_shader patches the default shaderfiles to be GLSL 1.20 complaint.

Now alacritty runs on my Thinkpad T500, with only OpenGL 2.1:
QsfIYDf

Thanks to the excellent work by Jason Ekstrand, GL_EXT_gpu_shader4 and GL_ARB_blend_func_extended were made available in a branch of Mesa for Intel iGPUs of Gen4 and Gen5. This is documented [in the Mesa-dev mailing list](https://lists.freedesktop.org/archives/mesa-dev/2021-May/225249.html). This includes the GM45 chipset / the GMA 4500 MHD. With that Mesa branch and this PR, Alacritty can run on a new range of devices, where it was previously not possible, like the still popular Thinkpad x200, T400, T500 etc. and would clear up issues like the ones seen here: https://github.com/alacritty/alacritty/issues/128#issuecomment-398056784, here: https://github.com/alacritty/alacritty/issues/128#issuecomment-439418891, here: https://github.com/alacritty/alacritty/issues/128#issuecomment-708902510 and here: https://github.com/alacritty/alacritty/issues/567#issuecomment-325201151. Hopefully these Gen4/Gen5 OpenGL extensions will get upstreamed to mainline Mesa...^^

Since my favourite terminal emulator [termite](https://github.com/thestinger/termite) deprecated, I'm looking forward to use a terminal emulator with a reputation as excellent as alacritty, on my trusty Thinkpad x200 and T500.


alacritty/res/text.f.glsl
```diff
@@ -21,7 +23,7 @@ void main() {
         color = vec4(bg.rgb, 1.0);
     } else if ((int(fg.a) & COLORED) != 0) {
         // Color glyphs, like emojis.
-        vec4 glyphColor = texture(mask, TexCoords);
+        vec4 glyphColor = texture2D(mask, TexCoords);
```
chrisduerr, line 26: Have you tested these new shaders on machines that support OpenGL 3.3+? Did you notice any regressions?

chrisduerr (commented):
Thanks for the PR. I can test this on macOS myself tomorrow.

alacritty/res/text.f.glsl
FrostKiwi: tested, no issues or regressions.
texture2D() and texture() are one and the same under the hood. With OpenGL 3+ more capabilities were added to what texture2D() can do, so it was renamed to texture().
The naming texture2D was deprecated and is not available in OpenGL 3.3 core any more, but is still available in all compatibility profiles.

chrisduerr: That's my understanding too, I'll still run a few framerate benchmarks once I've confirmed this works properly on macOS.

FrostKiwi: Actually, on my Nvidia Graphics card it fails to run, whilst on my Intel OpenGL 3+ Laptop it runs fine. Maybe the NVidia driver does not assume Compatibility by default?
I guess I'll add seperate Shaders based on GL_SHADING_LANGUAGE_VERSION after all...

FrostKiwi: I added a GLSL Version check. Now both my Thinkpad T500 and my GTX 1060 are happy.
Since the original GLSL files are used by default, there should be no change in either performance, or compatibility.

FrostKiwi:
> 
> 
> I have concerns about just duplicating shader files for different versions. That doesn't seem beneficial for maintainability. Especially since they're **largely** the same, but not entirely. And for most people it's probably impossible to verify changes to those shader files work properly.

Makes sense, will redo.

Rada8732 (approved):

FrostKiwi:
Removed the multiple shader files and reverted all changes to mod.rs and rect.rs
Instead just create_shader in mod.rs is being used to patch the shader files at run-time, if GLSL120 is detected. Kind like graphics settings in video games...

This limits the changes of this PR to just a single function and the default shader files are untouched.

FrostKiwi:
> 
> 
> > The change from texture( to texture2D( should be bulletproof already. With the open brace there should be no situation, where if would replace something else, but a function call.
> 
> What if a different function ends with `texture(`? There's **nothing** bulletproof about this.
> 
> > Lastly, keep in mind, this code path is only taken under one very specific case: The case of GLSL120, where the default GLSL files crash anyways.
> 
> That is not a good argument at all. Saying things don't matter because they're only part of the supported platforms is not a reasonable approach. If something is officially supported, it needs to work properly. If we can't provide this feature properly, it shouldn't be upstreamed.

I see what you mean. Will redo.

FrostKiwi:
Reworked create_shader() - No more String replacement and No multiple shader files for maintainability

In GLSL 1.20, texture() is undefined. So we can define a function Macro to enable backward compatibility:
> #define texture(a, b) texture2D(a, b)

As per [GLSL 1.20 spec page 64](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.1.20.pdf) and [GLSL 3.30 spec page 92,](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.3.30.pdf) texture2D() and texture() and have an optional third argument for the mip-map sample bias. _VA_ARGS  _ is not allowed in GLSL, the macro cannot cover for this case. However, since this a terminal application and mip-maps are not used, this should be fine.

Lastly, to keep in line with
>The #version directive must occur in a shader before anything else, except for comments and white space.

create_shader() checks for the existence of C-Style and CPP-Style comments, before placing the GLSL 1.20 patch. This is to be in line with the specification and the unlikely case, that a directive is placed in the comments. This was behavior was tested with the use of the following Shader file:
```
/* These comments contain #version 330 core to test shader cleanup */
/* These comments contain #version 330 core 
to test shader cleanup */
/* These comments contain
 #version 330 core to test shader cleanup */
// These comments contain #version 330 core to test shader cleanup
#version 330 core
layout (location = 0) in vec2 aPos;
layout (location = 1) in vec4 aColor;

flat out vec4 color;

void main()
{
    color = aColor;
    gl_Position = vec4(aPos.x, aPos.y, 0.0, 1.0);
}
```

chrisduerr (changes requested):
I'm sorry for the lack of communication, but I don't think this PR is going anywhere.

All the existing patches have had serious issues especially with regards to maintainability and I don't expect any future patches to change that. I could probably try and draft something up myself, but I'm not interested in investing time into this right now.

So unless you have any breakthroughs, I'd suggest maintaining this as a fork as long as no maintainable versions of this patch are available for upstreaming.

FrostKiwi:
@chrisduerr makes sense
My approach so far was not appropriate for alacritty's code base.
I think there is another way, inspired by how the OpenGL demo implementations of the GUI library Nuklear handle the backwards compatibility.
I'll close this, remake the patch from the ground up and create a new PR by the end of the week.

labmir:
I've been using Alacritty on my thinkpad x201 (OpenGL 2.1, sway, wayland) with the Alacritty patch (that lowers the GLSL version). It works fine, except when you set the terminal background color white, the text becomes very thin, but when you put terminal into fullscreen, the problem goes away.
image
image

FrostKiwi:
> I've been using Alacritty on my thinkpad x201 (OpenGL 2.1, sway, wayland) with the Alacritty patch (that lowers the GLSL version). It works fine, except when you set the terminal background color white, the text becomes very thin, but when you put terminal into fullscreen, the problem goes away.

FYI: Though the code change is the same, this PR is abondoned. I have opened a new one with cleaner code: https://github.com/alacritty/alacritty/pull/5168

Also Sway user here :]
I tried replicating those effects with your color scheme:
20210616_14h43m48s_grim
20210616_14h43m21s_grim

I saw no such fringing and no difference to fullscreen. Can you check, with "glxinfo | grep blend_func_extended" whether GL_ARB_blend_func_extended is found? This is the function that is used for Glyph blending. Though GLSL should not have compiled with it missing, though gotta check that first.
As descriebed above, in the case of my Laptops, the Thinkpad x200 and Thinkpad T500 that extension has to be enabled in Mesa first, before it is used for clean text blending...

labmir:
`Can you check, with "glxinfo | grep blend_func_extended" whether GL_ARB_blend_func_extended is found? `
Just instilled Mesa from git using [wip/gen4-gpu-shader4 branch](https://lists.freedesktop.org/archives/mesa-dev/2021-May/225249.html) and it all seem to work fine now. Thank you for all your work BTW, I can see you've done so much to get this working on older Intel cards!

labmir:
`Also Sway user here :]`
How did you make that title-bar, while having no borders? 
image

FrostKiwi:
> Thank you for all your work BTW, I can see you've done so [much](https://lists.freedesktop.org/archives/mesa-dev/2021-May/225249.html) to get this working on older Intel cards!

Thanks :] It's mainly the Mesa guys doing some incredible legacy support. To tidy up Mesa's codebase we are actually [getting a whole new 3D Driver](https://www.phoronix.com/scan.php?page=news_item&px=Crocus-WIP-MR-Mesa) for our old cards!

> How did you make that title-bar, while having no borders?

That comes from this Sway PR allowing resizeable titlebars: https://github.com/swaywm/sway/pull/6059  I always wanted that kind of border style, but it was not possible. You can misuse that titlebar height feature to get that look.

PR #5117, opened 2021-05-13
Replace tabs with spaces when converting lines to strings

### LINKED ISSUE #5084: Selection not copied after tab character
This morning, I had to come to the disappointing realization that... ehh.. that somebody is watching me. And it is not God. They are already knocking at my door, so I can only say that censorship has reached terminals too. Be careful!

```sh
$ printf 'I do not\rI\t like '$TERM
I do not like alacritty

$ # Selecting previous text, for example to send it to a friend...
$ xclip -o
I        like alacritty

$ # Do     panic.
$ rrn -rf /
```

---

Jokes aside, here is a minimal, reproducible example:
```sh
$ printf 'x X\r[\t]'
[ X     ]

$ # Selecting whole line using mouse/Vi-mode.
$ xclip -o|od -c
0000000   [  \t   ]
```

### System

OS: Arch Linux
Version: `alacritty 0.7.2 (5ac8060b)`
Linux/BSD: X11

PabloMarcendo:
Fixes #5084

The `line_to_string()` function used to skip over cells after a tab, but the example in the issue shows that there can be characters is those cells.

By replacing the tab characters with spaces and not skipping over cells, the clipboard value now looks the same as the copied selection.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -400,26 +400,17 @@ impl<T> Term<T> {
             cols.start -= 1;
         }
 
-        let mut tab_mode = false;
         for column in (cols.start.0..line_length.0).map(Column::from) {
             let cell = &grid_line[column];
 
-            // Skip over cells until next tab-stop once a tab was found.
-            if tab_mode {
-                if self.tabs[column] {
-                    tab_mode = false;
-                } else {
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
Tabs should be copied as tabs, not as spaces.

PR #5127, opened 2021-05-15
Replace tabs with spaces if there is text between the tab and the next tab-stop

PabloMarcendo:
Second attempt at fixing #5084.

Most tabs should still be copied as tabs.
The use of `\r` characters can cause the cells between the tab and the next tab-stop to contain non-whitespace characters. When calling the `line_to_string` method, the `\r` itself is no longer present in the terminal line. Trying to guess where it was and inserting it in the resulting string seems like a bad idea to me. So in this case I think we should replace the tab with a space.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -400,17 +400,37 @@ impl<T> Term<T> {
             cols.start -= 1;
         }
 
+        let mut tab_mode: Option<String> = None;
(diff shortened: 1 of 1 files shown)


chrisduerr:
This is **not** something that should be fixed on copy and I'm not sure why you seem to believe it should be. Fixing things on output rather than on input is absurd, we should make sure that our grid is in a reasonable state after writing to it.

PR #5129, opened 2021-05-15
Add a `--here` flag

Milo123459:
Hello!

First-time contributor here, looking to help out.
Something I do a lot from VSCode is typing out `alacritty --working-directory $PWD` which is very long and annoying after a while of typing, so I changed a few things:

Adds a short option for working-directory: `w`
Adds a new --here flag, opening alacrity in the current working directory.

If both flags are present (`--working-directory` and `--here`) the value of working directory will be used.

Thank you! Hope to be making some more PRs.

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -1,4 +1,5 @@
 use std::cmp::max;
+use std::env;
 use std::path::PathBuf;
 
 use clap::{crate_authors, crate_description, crate_name, crate_version, App, Arg};
@@ -114,6 +115,7 @@ impl Options {
             .arg(
                 Arg::with_name("working-directory")
                     .long("working-directory")
+                    .short("wd")
                     .takes_value(true)
                     .help("Start the shell in the specified working directory"),
             )
@@ -137,6 +139,11 @@ impl Options {
                     .multiple(true)
                     .takes_value(true)
                     .help("Override configuration file options [example: cursor.style=Beam]"),
+            )
+            .arg(
+                Arg::with_name("here")
+                    .long("here")
+                    .help("Open alacritty in the current working directory"),
             )
             .get_matches();
 
@@ -169,6 +176,10 @@ impl Options {
             _ => options.log_level = LevelFilter::Trace,
         }
 
+        if matches.value_of("here").is_some() {
+            options.working_directory = Some(env::current_dir().unwrap())
+        }
+
         if let Some(dir) = matches.value_of("working-directory") {
             options.working_directory = Some(PathBuf::from(dir.to_string()));
         }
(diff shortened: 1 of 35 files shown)


chrisduerr:
On any sane operating system that would be the default behavior anyways. And I don't see much problem in the way you do it right now.

Milo123459:
> On any sane operating system that would be the default behavior anyways. And I don't see much problem in the way you do it right now.

OK, I completely understand.

The only thing I wanted to achieve was saving a bit of time, but I understand. Maybe something so it default to the current working directory if no argument is provided to working directory? I don't know.

Sorry for any trouble

chrisduerr:
> The only thing I wanted to achieve was saving a bit of time, but I understand. Maybe something so it default to the current working directory if no argument is provided to working directory?

That's already what Alacritty does. That's how spawning processes works. It's inherited from the parent.

Milo123459:
Not for me at least. I just typed `alacritty --working-directory` and it says a value is required. Ideas?

chrisduerr:
Alacritty by default will inherit the working directory from its parent. No arguments required.

Milo123459:
> Alacritty by default will inherit the working directory from its parent. No arguments required.

Oh, I misunderstood you, sorry. 

From my understanding, if I'm cded into Desktop in VScode and I type alacrity, alacrity should open in the desktop directory? It doesn't, it just goes to my predefined working_directory in my config, which is only there so it doesn't spawn in the system32 folder on windows as it usually does without it.

Milo123459:
Anyway, sorry for wasting your time

brian6932:
@Milo123459 Would it be possible for you to rebase and resubmit this PR with just the short option for `--working-directory`? I think a `-w` or `-wd` alias for `--working-directory` would be pretty useful. It feels like too much typing as is, and I was kind of surprised that this was not a thing already.

Milo123459:
Sure! I'll work on it soon.

Milo123459:
Actually - there is no point. Chris already said no and I don't want to bother him with this again. I doubt changing this would change anything.

chrisduerr:
I mean the `-w` flag isn't used at the moment, so if people are actually interested I probably wouldn't mind adding that short option. But I do have to question what you're doing when typing out `--working-directory` fully instead of just using shell completions after `--w`. So ultimately I think for most people it just shouldn't really matter since it's just a single character difference basically.

Milo123459:
Well, the reason I wanted this change is because if you spawn Alacritty in lets say a VSCode terminal then close VSCode, alacrity closes along with it.

chrisduerr:
You can just disown the terminal from the terminal spawning it and that's not an issue anymore.

brian6932:
There are people who don't use that type of shell completion in order to use stuff like fzf or other tab invoked CLI programs, I do think these should be submitted as different PRs tho, if it isn't used, I don't see why not @chrisduerr

chrisduerr:
People shooting themselves in the foot probably shouldn't surprised when it bleeds.

PR #5141, opened 2021-05-17
feat: Add WINIT_X11_SCALE_FACTOR to config

kozachynskyi:
This change to standard config is aimed to decrease the number of new issues connected to wrong font scaling,


chrisduerr:
I would discourage people from setting this, so I'm not particularly interested in documenting this in the config file. It's exists solely to resolve bugs with xrandr.

PR #5152, opened 2021-05-19
chore(deps): force usage of serde_yaml >= 0.8.4

PythonCreator27:
Upgrade serde_yaml to >= 0.8.4. Nothing really changes, just for https://deps.rs.

NOTE: cargo parses `0.8.4` as `^0.8.4`, so nothing *should* have been flagged by deps.rs in the first place, but this just really settles it.


chrisduerr (changes requested):
I see no reason to make this change. While I personally prefer specifying the full semver version for depedencies, there's no reason to do it with just `serde_yaml`.

PR #5158, opened 2021-05-20
Add IPFS URL schemes support (ipfs: & ipns:)

pataquets:
As done in #5027 & #5149.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Regex terminal hints ([see features.md](./docs/features.md#hints))
 - macOS keybinding (cmd+alt+H) hiding all windows other than Alacritty
 - Support for `magnet` URLs
+- Support for `ipfs`/`ipns` URLs
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


pataquets:
I can't remember if any CLI app emits such URLs on its console output. However, I find them from time to time (under Alacritty), either while browsing with w3m/lynx/etc. or in textfiles or mail messages under Emacs.
For handling such URLs, I use [Brave browser](https://brave.com/ipfs-support/) which natively supports them when clicked (or opening with `xdg-open`, for instance).
More info: https://docs.ipfs.io/how-to/address-ipfs-on-web/#native-urls

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Regex terminal hints ([see features.md](./docs/features.md#hints))
 - macOS keybinding (cmd+alt+H) hiding all windows other than Alacritty
 - Support for `magnet` URLs
+- Support for `ipfs`/`ipns` URLs
```
chrisduerr, line 21: Could you add a new `0.9.0-dev` section at the top and add things there? And then remove the `-dev` from `0.8.0-dev`.

chrisduerr (changes requested):
I still question how useful support for this protocol would be, but I suppose it doesn't hurt much to add if it's in use by some people.

chrisduerr (changes requested):
This PR needs to be rebased on top of the latest master.

pataquets:
Done!

PR #5227, opened 2021-06-10
Fix wrap around buffer on search

a5ob7r:
# SUMMARY

## Problem
SearchNext/SearchPrevious actions doesn't work in almost cases.

## Solution
Stop regex search traversing only if no need to wrap around on scroll buffer and searching target point has already been at same point to end point or been beyond one.

# DETAIL

SearchNext/SearchPrevious actions doesn't work in almost cases. The actions try to search between head point of current line on vi mode cursor to tail point of a line of one level above the line. For example, when a start point is at (3, 0) and an end point position is determined as (2, 119) on a grid(80x120) for a searching, the searching doesn't find anything.

The reason why is that the start point has already been beyond the end point(`(2, 119) <= (3, 0)`). So It is recognized that the search traversing has already been completed and it exits immediately on starting.

So consider whether needs to wrap around to opposite side of scroll buffer at an scroll buffer side or not, in addition to searching start point and the end one, to detect search traversing completion.

### DIFF
--- a/alacritty_terminal/src/term/search.rs
+++ b/alacritty_terminal/src/term/search.rs
@@ -203,7 +203,7 @@ impl<T> Term<T> {
 
         let mut iter = self.grid.iter_from(start);
         let mut state = dfa.start_state();
-        let mut last_wrapped = false;
+        let mut last_line_wrapped = false;
         let mut regex_match = None;
         // Need to wrap around to opposite side of scroll buffer if 'end' point has already been
         // same at 'start' point or been beyond one.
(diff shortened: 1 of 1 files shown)


chrisduerr:
Could you please provide an example, Ideally with a video? I find your English very difficult to understand.

chrisduerr:
Huh, that is completely broken. I'm not sure about your patch though.

The first step should be to backtrace this regression and verify where it came from, my bet would be on 3e867a05. And at a glance your patch seems a bit peculiar when looking at those original changes, so I'm not sure that makes a lot of sense.

Could you verify this?

a5ob7r:
I verified it. I confirmed that your bet is right. Actually, this regression came from  3e867a0.

chrisduerr:
In that case the original problem is likely what needs solving, just in a different manner.

a5ob7r:
OK. I will try find the original problem. I close this PR instead.

@chrisduerr Thank you for your comments!

PR #5244, opened 2021-06-16
Fix regex search regression

a5ob7r:
The regression is that vi cursor doesn't jump to match in almost cases when invokes SearchNext/SearchPrevious actions. (The exception is that vi cursor is on topmost line of scrollback buffer or on bottompost and rightmost point of one.)

The original problem of this regression is that the regex searching skips one character follows wide character without wide character spacer unintentionally.

This pr is an alternative approach of #5227 which has already been closed.

### DIFF
--- a/alacritty_terminal/src/term/search.rs
+++ b/alacritty_terminal/src/term/search.rs
@@ -800,4 +800,29 @@ mod tests {
         let mut iter = RegexIter::new(start, end, Direction::Right, &term, &dfas);
         assert_eq!(iter.next(), None);
     }
+
+    #[test]
+    fn wrap_around_to_another_end() {
+        #[rustfmt::skip]
+        let term = mock_term("\
+            abc\r\n\
+            def\
+        ");
+
+        // bottom to top
+        let dfas = RegexSearch::new("abc").unwrap();
+        let start = Point::new(Line(1), Column(0));
+        let end = Point::new(Line(0), Column(2));
+        let match_start = Point::new(Line(0), Column(0));
+        let match_end = Point::new(Line(0), Column(2));
+        assert_eq!(term.regex_search_right(&dfas, start, end), Some(match_start..=match_end));
+
+        // top to bottom
+        let dfas = RegexSearch::new("def").unwrap();
+        let start = Point::new(Line(0), Column(2));
+        let end = Point::new(Line(1), Column(0));
+        let match_start = Point::new(Line(1), Column(0));
+        let match_end = Point::new(Line(1), Column(2));
+        assert_eq!(term.regex_search_left(&dfas, start, end), Some(match_start..=match_end));
+    }
 }
--- a/alacritty_terminal/src/term/search.rs
+++ b/alacritty_terminal/src/term/search.rs
@@ -809,7 +809,7 @@ mod tests {
             def\
         ");
 
-        // bottom to top
+        // Bottom to top.
         let dfas = RegexSearch::new("abc").unwrap();
         let start = Point::new(Line(1), Column(0));
         let end = Point::new(Line(0), Column(2));
@@ -817,7 +817,7 @@ mod tests {
         let match_end = Point::new(Line(0), Column(2));
         assert_eq!(term.regex_search_right(&dfas, start, end), Some(match_start..=match_end));
 
-        // top to bottom
+        // Top to bottom.
         let dfas = RegexSearch::new("def").unwrap();
         let start = Point::new(Line(0), Column(2));
         let end = Point::new(Line(1), Column(0));
--- a/alacritty_terminal/src/term/search.rs
+++ b/alacritty_terminal/src/term/search.rs
@@ -239,12 +239,7 @@ impl<T> Term<T> {
             }
 
             // Stop once we've reached the target point.
-            //
-            // We check beyond the point itself to account for skipped characters after wide chars
-            // without spacer.
-            if (direction == Direction::Right && point >= end)
-                || (direction == Direction::Left && point <= end)
-            {
+            if point == end {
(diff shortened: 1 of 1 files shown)


alacritty_terminal/src/term/search.rs
```diff
@@ -798,4 +800,29 @@ mod tests {
...
+
+    #[test]
+    fn wrap_around_to_another_end() {
+        #[rustfmt::skip]
+        let term = mock_term("\
+            abc\r\n\
+            def\
+        ");
+
+        // bottom to top
```
chrisduerr, line 812:
```suggestion
        // Bottom to top.
```

alacritty_terminal/src/term/search.rs
```diff
@@ -798,4 +800,29 @@ mod tests {
...
+        let dfas = RegexSearch::new("abc").unwrap();
+        let start = Point::new(Line(1), Column(0));
+        let end = Point::new(Line(0), Column(2));
+        let match_start = Point::new(Line(0), Column(0));
+        let match_end = Point::new(Line(0), Column(2));
+        assert_eq!(term.regex_search_right(&dfas, start, end), Some(match_start..=match_end));
+
+        // top to bottom
```
chrisduerr, line 820:
```suggestion
        // Top to bottom.
```

alacritty_terminal/src/term/search.rs
```diff
@@ -291,6 +286,13 @@ impl<T> Term<T> {
         match direction {
             Direction::Right if cell.flags.contains(Flags::WIDE_CHAR) => {
                 iter.next();
+
+                // If iterates `iter` when finds a wide character without wide character spacer,
+                // one character is skipped unintentionally. So returns `iter` to one previous
+                // position.
+                if !iter.cell().flags.contains(Flags::WIDE_CHAR_SPACER) {
+                    iter.prev();
+                }
```
chrisduerr, line 295: Couldn't we just not call `iter.next()` in this case? Calling next just to call prev immediately seems like a strange process.

chrisduerr (changes requested):
Thanks for the PR, sorry for the delay.

alacritty_terminal/src/term/search.rs
```diff
@@ -290,7 +285,16 @@ impl<T> Term<T> {
...
+                // If advances `iter` when finds a wide character without wide character spacer,
+                // one character is skipped unintentionally. So advances `iter` only if next
+                // item(cell) exists and it is wide character spacer.
+                let idxd = match iter.peek() {
+                    Some(idxd) => idxd,
+                    None => return
+                };
+                if idxd.cell.flags.contains(Flags::WIDE_CHAR_SPACER) {
+                    iter.next();
+                }
```
chrisduerr, line 297: I think you might be overcomplicating things, there's no need to do this peeking to determine if there's going to be a spacer or not, all we need to make sure of is that there's a cell after the wide char. So we just need to check the current cell's column.

chrisduerr (changes requested):

alacritty_terminal/src/term/search.rs
a5ob7r: Oh, sure!

a5ob7r:
@chrisduerr Done!

alacritty_terminal/src/term/search.rs
```diff
@@ -290,7 +285,11 @@ impl<T> Term<T> {
     ) {
         match direction {
             Direction::Right if cell.flags.contains(Flags::WIDE_CHAR) => {
-                iter.next();
+                // If advances `iter` at the end of column one character is skipped
+                // unintentionally.
+                if iter.point().column < self.columns() - 1 {
+                    iter.next();
+                };
```
chrisduerr, line 292: For the sake of consistency, the conditional should probably be moved into the match arm. Otherwise it's somewhat arbitrary.

alacritty_terminal/src/term/search.rs
```diff
@@ -290,7 +285,11 @@ impl<T> Term<T> {
     ) {
         match direction {
             Direction::Right if cell.flags.contains(Flags::WIDE_CHAR) => {
-                iter.next();
+                // If advances `iter` at the end of column one character is skipped
+                // unintentionally.
```
chrisduerr, line 289: Having a comment on this makes sense, but the comment itself doesn't. This is not about the "end of column", but the end of the row (i.e. the last column in the row).

Something more like this seems more appropriate:

```
// In the alternate screen buffer there might not be a wide char spacer, so we only advance the iterator when the wide char is not in the last column.
```

A bit verbose maybe, but I think that is a better description of the issue.

chrisduerr (changes requested):

alacritty_terminal/src/term/search.rs
```diff
@@ -289,9 +284,14 @@ impl<T> Term<T> {
         direction: Direction,
     ) {
         match direction {
-            Direction::Right if cell.flags.contains(Flags::WIDE_CHAR) => {
+            // In the alternate screen buffer there might not be a wide char spacer after a wide
+            // char, so we only advance the iterator when the wide char is not in the last column.
+            Direction::Right
+                if cell.flags.contains(Flags::WIDE_CHAR)
+                    && iter.point().column < self.columns() - 1 =>
```
chrisduerr, line 291:
```suggestion
                    && iter.point().column < self.last_column() =>
```

chrisduerr (changes requested):

chrisduerr (approved):
Finally regex search is working again. Thanks for your help.

a5ob7r:
Thank you as well for your many reviews!

PR #5257, opened 2021-06-18
Add modes to regex hint bindings

### LINKED ISSUE #5154: Allow specifying a mode for hint bindings
I'm trying out the new hints feature and hacked together a regex for file paths. I'd now like to use `F` in vi mode as hint binding. Am I missing something or is this currently not supported?

Relevent part of my config:

```yml
hints:
  enabled:
   - regex: "(/|~/)[[:print:]]+/([[:alnum:]]+[.])+[[:alnum:]]+[[:space:]]"
     command:
       program: emacsclient
       args: [--no-wait, --create-frame, --alternate-editor=""]
     post_processing: true
     mouse:
       enabled: true
       mods: None
     binding:
        key: F
        # mode: Vi|~Search

```

Version: `alacritty 0.8.0 (a1b13e68)`

sophrosyne97:
Closes #5154

Adds ModeWrapper to the HintBinding struct to allow users to specify which modes to allow/disallow for their hint binding.
Also implements the Default trait for ModeWrapper to allow users to omit the mode field like ModsWrapper ~~and adds the "None" pattern match for the deserialize implementation to make the mode field consistent with the mods field for the config~~.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -458,8 +458,8 @@
   # Each hint must have a `regex` and either an `action` or a `command` field.
   # The fields `mouse`, `binding` and `post_processing` are optional.
   #
-  # The fields `command`, `binding.key`, `binding.mods` and `mouse.mods` accept
-  # the same values as they do in the `key_bindings` section.
+  # The fields `command`, `binding.key`, `binding.mods`, `binding.mode` and 
+  # `mouse.mods` accept the same values as they do in the `key_bindings` section.
   #
   # The `mouse.enabled` field controls if the hint should be underlined while
   # the mouse with all `mouse.mods` keys held or the vi mode cursor is above it.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Support for `ipfs`/`ipns` URLs
+- Mode field for regex hint bindings
 
 ### Fixed
 
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -726,7 +726,8 @@ impl<'a> Deserialize<'a> for Key {
     }
 }
 
-struct ModeWrapper {
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct ModeWrapper {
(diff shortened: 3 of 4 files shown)


alacritty.yml
```diff
@@ -458,7 +458,7 @@
   # Each hint must have a `regex` and either an `action` or a `command` field.
   # The fields `mouse`, `binding` and `post_processing` are optional.
   #
-  # The fields `command`, `binding.key`, `binding.mods` and `mouse.mods` accept
+  # The fields `command`, `binding.key`, `binding.mods`, `binding.mode` and `mouse.mods` accept
```
chrisduerr, line 461: Config yaml doc is wrapped at 80.

alacritty.yml
```diff
@@ -488,6 +488,7 @@
   #   binding:
   #     key: U
   #     mods: Control|Shift
+  #     mode: None
```
chrisduerr, line 491: Adding a new mode would require adding it to the documentation in the bindings section, but I'm not sure it's a good idea to add `None` as a mode, since there is no "None" mode.

This is a list of modes and notmodes, so `~None` would be a thing too, which makes no sense really. If anything the default would just be an empty string.

chrisduerr (changes requested):

alacritty.yml
sophrosyne97: Yeah I added the `none` mode because I couldn't figure out a good way to show the new field in the configuration file while also having both of the `ModeWrapper`'s `BindingMode` members remain empty. That's why I only wrote `none` in instead of both `none` and `~none,` since it would purely mean to return the default `ModeWrapper` in its deserialize implementation. I can change this if you like though and just simply omit the field from the binding example in config and remove the `None` mode if you prefer.

chrisduerr: Just `mode: ""` would work. But since mode is somewhat of an "advanced" feature I'd be kinda okay with just leaving it out too tbh.

sophrosyne97: Alright sounds good to me.

alacritty.yml
sophrosyne97: fixed.

alacritty.yml
sophrosyne97: fixed.

chrisduerr (approved):
LGTM

PR #5265, opened 2021-06-21
Avoid rapidly flashing lights if bell is repeated frequently

AlexApps99:
An example of this is when backspace is held in bash:
Peek 2021-06-21 17-37


chrisduerr:
I see no problem with the current behavior. Why did you create this PR?

AlexApps99:
I thought that the flashing repeatedly was too irritating, especially as it happens so often
Feel free to close if you disagree

chrisduerr:
Changing the algorithm used to something that falls off a little slower will likely help a lot already if you're bothered by the rapid flashing.

But generally I'd recommend running a configuration that doesn't spam the bell that rapidly.

PR #5267, opened 2021-06-21
Fix vi cursor scroll to emulate vi/vim

a5ob7r:
This is an additional patch for 6f135d7.

This fixes the regression that vi cursor doesn't move to appropriate position to emulate vi/vim after invokes `SearchPage*`.

To emulate vi/vim the cursor should move up/down some lines if the viewport on topmost scrollback buffer or on bottommost one when invokes `SearchPage*` action. Otherwise the cursor should look like no movement relatively on viewport.

They are sample videos about this problem.

Steps in videos

1. Run `cat alacritty.yml`
2. Enter vi mode
3. Invoke `ScrollHalfPageUp` action several times
4. Invoke `Up` action several times
5. Go to `3.` step

Before fixing regression,

https://user-images.githubusercontent.com/12132068/122762915-c2eccd80-d2d8-11eb-8af9-8c6ff90353a9.mp4

After fixing regression,

https://user-images.githubusercontent.com/12132068/122763027-dd26ab80-d2d8-11eb-8fd0-cfaab28cee5c.mp4

BTW in vi,

Steps in the video
1. Open `alacritty.yml` using vi and go to bottom of the file
2. Type `Ctrl+U` several times
3. Type `k` or `j` several times
4. Go to `2.` step

https://user-images.githubusercontent.com/12132068/122763269-19f2a280-d2d9-11eb-9893-d65c2f3ed245.mp4

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1989,33 +1989,6 @@ mod tests {
     }
 
     #[test]
-    fn selection_with_scroll() {
-        let size = SizeInfo::new(5., 3., 1.0, 1.0, 0.0, 0.0, false);
-        let mut term = Term::new(&MockConfig::default(), size, ());
-
-        for _ in 0..2 {
-            term.grid_mut()[Line(2)][Column(0)].c = 'a';
-            term.grid_mut()[Line(2)][Column(1)].c = 'b';
-            term.grid_mut()[Line(2)][Column(2)].c = 'c';
-            term.clear_screen(ansi::ClearMode::All);
-        }
-        term.grid_mut()[Line(0)][Column(0)].c = 'd';
-
-        term.toggle_vi_mode();
-        term.selection =
-            Some(Selection::new(SelectionType::Lines, Point::new(Line(0), Column(0)), Side::Left));
-
-        term.scroll_display(Scroll::PageUp);
-        assert_eq!(term.selection_to_string(), Some(String::from("abc\nd\n")));
-
-        term.scroll_display(Scroll::PageUp);
-        assert_eq!(term.selection_to_string(), Some(String::from("abc\n\n\nabc\nd\n")));
-
-        term.scroll_display(Scroll::PageUp);
-        assert_eq!(term.selection_to_string(), Some(String::from("abc\n\n\nabc\nd\n")));
-    }
-
-    #[test]
     fn semantic_selection_works() {
         let size = SizeInfo::new(5., 3., 1.0, 1.0, 0.0, 0.0, false);
         let mut term = Term::new(&MockConfig::default(), size, ());
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -280,11 +280,6 @@ impl<T> Term<T> {
         self.grid.scroll_display(scroll);
         self.event_proxy.send_event(Event::MouseCursorDirty);
 
-        // Clamp vi mode cursor to the viewport.
-        let viewport_start = -(self.grid.display_offset() as i32);
-        let viewport_end = viewport_start + self.bottommost_line().0;
-        let vi_cursor_line = &mut self.vi_mode_cursor.point.line.0;
-        *vi_cursor_line = min(viewport_end, max(viewport_start, *vi_cursor_line));
         self.vi_mode_recompute_selection();
     }
 
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -280,6 +280,11 @@ impl<T> Term<T> {
         self.grid.scroll_display(scroll);
         self.event_proxy.send_event(Event::MouseCursorDirty);
 
+        // Clamp vi mode cursor to the viewport.
+        let viewport_start = -(self.grid.display_offset() as i32);
+        let viewport_end = viewport_start + self.bottommost_line().0;
+        let vi_cursor_line = &mut self.vi_mode_cursor.point.line.0;
+        *vi_cursor_line = min(viewport_end, max(viewport_start, *vi_cursor_line));
         self.vi_mode_recompute_selection();
     }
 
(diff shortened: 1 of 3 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -280,11 +280,6 @@ impl<T> Term<T> {
         self.grid.scroll_display(scroll);
         self.event_proxy.send_event(Event::MouseCursorDirty);
 
-        // Clamp vi mode cursor to the viewport.
-        let viewport_start = -(self.grid.display_offset() as i32);
-        let viewport_end = viewport_start + self.bottommost_line().0;
-        let vi_cursor_line = &mut self.vi_mode_cursor.point.line.0;
-        *vi_cursor_line = min(viewport_end, max(viewport_start, *vi_cursor_line));
```
chrisduerr, line 287: Why did you remove this? I believe this is still necessary in some scenarios.

alacritty_terminal/src/vi_mode.rs
```diff
@@ -756,4 +746,45 @@ mod tests {
...
+    fn history_term() -> Term<()> {
+        let mut term = term();
+
+        for i in 0..(term.screen_lines() as i32) {
+            term.grid_mut()[Line(i)][Column(0)].c = 'c';
+        }
+        term.grid_mut().clear_viewport();
+
+        term
+    }
```
chrisduerr, line 759: What's the purpose of this function?

chrisduerr (changes requested):
I'm not sure I prefer VI's behavior here, which is why in the last PR I've mentioned that we should probably stay in the same column instead of moving to the first.

I believe we've behaved this way in the past, I'm not sure why it was changed.

a5ob7r:
> I'm not sure I prefer VI's behavior here, which is why in the last PR I've mentioned that we should probably stay in the same column instead of moving to the first.

I see. Maybe preferable cursor movement is different between terminal and text editor.

> I believe we've behaved this way in the past, I'm not sure why it was changed.

`ScrollPage*` action's behavior has been changed since 3bd5ac2. Is it intentional? I think it is unintentional and is a regression which we should fix, because the changing isn't mentioned in the Pull Request/commit message.

chrisduerr:
> ScrollPage* action's behavior has been changed since 3bd5ac2. Is it intentional?

It's not. Since you've tested it already can you document here how exactly it worked before? I'm not entirely certain because I believe it has changed in the past.

a5ob7r:
Assume that the positions of viewport and vi cursor are indicated by new unified index type which is introduced since 3bd5ac2.

It is so simple. In the past a scroll action adds/subtracts same amount of lines to both viewport's position(display_offset) and vi cursor position. And it corrects (or clamps) viewport's position to fit in scrollback buffer and vi cursor position to fit in viewport if the positions go to out of bounds of them.

By the way, on current HEAD an scroll action adds/subtracts vi cursor position only if viewport position is on either end(topmost or bottommost) and scroll direction is the nearest end. On the other hand it always corrects vi cursor position to fit in viewport.

alacritty_terminal/src/term/mod.rs
a5ob7r: Can you show me an example? We should add it as a test if it exists.

alacritty_terminal/src/vi_mode.rs
a5ob7r: As this function name the purpose is to create a terminal which have scrollback buffer. But come to think of it, it is enough to call `clear_viewport()`.

alacritty_terminal/src/term/mod.rs
chrisduerr: Scrolling with the mouse. This is definitely still required.

alacritty_terminal/src/vi_mode.rs
```diff
@@ -756,4 +746,40 @@ mod tests {
         cursor = cursor.motion(&mut term, ViMotion::WordLeft);
         assert_eq!(cursor.point, Point::new(Line(0), Column(0)));
     }
+
+    fn history_term() -> Term<()> {
+        let mut term = term();
+        term.grid_mut().clear_viewport();
+        term
+    }
```
chrisduerr, line 754: Not sure this is helpful as a helper function. Probably easier to just inline it, since the behavior is a little more clear then.

alacritty_terminal/src/vi_mode.rs
```diff
@@ -756,4 +746,40 @@ mod tests {
...
+        let overlines = (term.total_lines() * 2) as i32;
+
+        let mut cursor = ViModeCursor::new(Point::new(Line(0), Column(0)));
+
+        cursor = cursor.scroll(&term, -overlines);
+        assert_eq!(cursor.point, Point::new(Line(term.screen_lines() as i32 - 1), Column(0)));
+
+        cursor = cursor.scroll(&term, overlines);
+        assert_eq!(cursor.point, Point::new(Line(-(term.history_size() as i32)), Column(0)));
+    }
```
chrisduerr, line 784: I'm not sure I like the whole `overlines` thing here. It makes stuff somewhat complicated. Especially because the cursor position is preserved between the two tests so one not only needs to make sure that the second test overshoots from the 0-0 point, but from the end position of the previous test. Making it possible for the second test to fail by running the first part just by picking the wrong `overlines` value.

I think being more explicit and inlining `overlines` to be specific to the thing we want to test right at that moment would clear things up a bit. It might also make sense to reset the cursor position between tests. That way by writing things out a bit more explicitly it should be easier to understand what we're trying to validate with the test.

chrisduerr (changes requested):
Sorry for the delay. Good work, looks fine mostly.

The behavior wrt the HalfPage motion seems accurate when comparing it to neovim at least.

alacritty_terminal/src/vi_mode.rs
a5ob7r: > Scrolling with the mouse. This is definitely still required.

I see. It is missing insight for me. I readd the code and add some tests to realize this regression.

> Not sure this is helpful as a helper function. Probably easier to just inline it, since the behavior is a little more clear then.

You're right. I remove this function and make it inline.

alacritty_terminal/src/vi_mode.rs
a5ob7r: OK. I rewrite this function and separate it into two functions to be explicit.

a5ob7r:
@chrisduerr Thank you for your reviews! I reflected them.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1930,6 +1930,83 @@ mod tests {
...
+        assert_eq!(term.vi_mode_cursor.point, Point::new(Line(-1), Column(0)));
+        assert_eq!(term.grid.display_offset(), 10);
+
+        term.scroll_display(Scroll::PageUp);
+        assert_eq!(term.vi_mode_cursor.point, Point::new(Line(-11), Column(0)));
+        assert_eq!(term.grid.display_offset(), 20);
+
+        term.scroll_display(Scroll::PageUp);
+        assert_eq!(term.vi_mode_cursor.point, Point::new(Line(-11), Column(0)));
+        assert_eq!(term.grid.display_offset(), 20);
```
chrisduerr, line 1953: This tests for a perfect match, where a jump ends up in exactly the right spot with a full page.

I feel like it might be useful to have the scrollback buffer uneven, just to test the case where it can only jump *less* than a full page until it lands at the top?

Same applies to `PageDown` this should be trivial to add.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1930,6 +1930,83 @@ mod tests {
...
+
+        term.scroll_display(Scroll::PageUp);
+        assert_eq!(term.selection_to_string(), Some(String::from("abc\nd\n")));
+
+        term.scroll_display(Scroll::PageUp);
+        assert_eq!(term.selection_to_string(), Some(String::from("abc\n\n\nabc\nd\n")));
+
+        term.scroll_display(Scroll::PageUp);
+        assert_eq!(term.selection_to_string(), Some(String::from("abc\n\n\nabc\nd\n")));
+    }
```
chrisduerr, line 2008: What was the reason for adding a selection test here?

I think testing the selection logic **and** vi cursor motion logic together is usually a bit strange. I'd rather have it broken down into smaller, simpler tests which should exist already. So I'm not sure why we'd test the conversion of the selection range to string in here.

alacritty_terminal/src/vi_mode.rs
```diff
@@ -756,4 +747,73 @@ mod tests {
         cursor = cursor.motion(&mut term, ViMotion::WordLeft);
         assert_eq!(cursor.point, Point::new(Line(0), Column(0)));
     }
+
+    #[test]
+    fn scroll_simple() {
+        let mut term = term();
+
+        // Create 1 lines of scrollback.
```
chrisduerr, line 755:
```suggestion
        // Create 1 line of scrollback.
```

chrisduerr (changes requested):
I think only the tests need some refinement here. Other than that everything looks good now.

alacritty_terminal/src/term/mod.rs
a5ob7r: The reason why is that `term.vi_mode_recompute_selection` is called in `term.scroll_display`. So I thought we should test whether correct re selection is completed or not. But I think your review is correct and will delete this test.

chrisduerr:
Could you rebase this on top of master? Just so it can be merged as soon as the review is done.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1932,6 +1932,62 @@ mod tests {
...
+    fn scroll_display_page_up() {
+        let size = SizeInfo::new(5., 10., 1.0, 1.0, 0.0, 0.0, false);
+        let mut term = Term::new(&MockConfig::default(), size, ());
+
+        // Create 11 lines of scrollback.
+        for _ in 0..20 {
+            term.newline();
+        }
+
+        // Scrollable amount to top is 11.
```
chrisduerr, line 1945: Love the comments here, makes things a lot more clear. Thanks.

chrisduerr (approved):
LGTM, thanks for the PR.

PR #5286, opened 2021-06-27
install.md: Suggest `nix develop`, too

Patryk27:
Newer versions of Nix deprecate `nix-shell`, providing in its place a new command called `nix develop`.

I've also s/NixOS/Nix, as using `nix-shell` / `nix develop` doesn't actually require NixOS (e.g. both should work fine on Ubuntu + Nix).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -176,10 +176,13 @@ eopkg install fontconfig-devel
 #### NixOS/Nixpkgs
 
 The following command can be used to get a shell with all development
-dependencies on [NixOS](https://nixos.org).
+dependencies using [Nix](https://nixos.org).
 
 ```sh
 nix-shell -A alacritty '<nixpkgs>'
+
+# or, if you're using Nix Flakes:
+nix develop nixpkgs#alacritty
 ```
 
 #### Gentoo


INSTALL.md
```diff
@@ -176,10 +176,13 @@ eopkg install fontconfig-devel
...
 
 The following command can be used to get a shell with all development
-dependencies on [NixOS](https://nixos.org).
+dependencies using [Nix](https://nixos.org).
 
 ```sh
 nix-shell -A alacritty '<nixpkgs>'
+
+# or, if you're using Nix Flakes:
+nix develop nixpkgs#alacritty
```
chrisduerr, line 185: We should suggest one or another, Alacritty isn't an instruction manual on how to use your package manager, if you have troubles you're expected to look into this yourself.

I'm not sure what the status of the deprecation is, but if the `nix develop` option is already supported widely the existing one should just be replaced.

chrisduerr (changes requested):

INSTALL.md
Patryk27: Got it - in this case let's stick with current `nix-shell`; `nix develop` is still experimental and might change. I'll re-create this pull request in the future, when the command gets stabilized 🙂

PR #5288, opened 2021-06-28
Remove .agignore

a5ob7r:
This file is a configuration file for The Silver Searcher which is useful but not related to Alacritty and the development directly.

Also this file has no effect in practice because the ignored target no longer exists in current repository. So nobody is affects by this removing.

### DIFF
--- a/.agignore
+++ b/.agignore
@@ -1,1 +1,0 @@
-tests/ref


chrisduerr:
Oh, I initially thought you added some stuff but looks like you just deleted the file. Definitely the right move, it's just been a waste of space until now. Nobody I know even uses ag.

chrisduerr (approved):

PR #5291, opened 2021-06-29
Add instructions to change default terminal to Alacritty

DanSGiesbrecht:
I found it useful to swap my default terminal emulator to Alacritty. This took a bit of time to figure out, and I think others would benefit from these instructions added to `INSTALL.md`.

I'm unfamiliar how broadly `update-alternatives` works in other distros - I'm solely familiar with Ubuntu and derivatives.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -290,6 +290,22 @@ If you are having problems with Alacritty's logo, you can replace it with
 prerendered PNGs and simplified SVGs available in the `extra/logo/compat`
 directory.
 
+### Default Terminal
+
+Some distributions support changing the default terminal emulator
+using `update-alternatives`:
+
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
> I'm unfamiliar how broadly update-alternatives works in other distros

It doesn't.

Generally setting default applications and similar operations are so highly distro/DE specific, that I'd expect someone to be familiar with how to do this on their own. If not, it's the job of the distro/DE to document how to do this, not Alacritty's.

So I don't think this is the correct place to document this.

DanSGiesbrecht:
@chrisduerr no problem at all. Since it's not a broadly applicable tool, I totally agree. Thanks for your time.

PR #5293, opened 2021-06-29
[WIP] Implement modifyOtherKeys=1

### LINKED ISSUE #3101: Implement "modifyOtherKeys"
In short, enabling _modifyOtherKeys_ lets the client recognize e.g. `Ctrl-.`; i.e. to see modifiers for keys where modifiers are normally ignored.

Note that in XTerm, the user can also enable this by default via XResources.

The enabling/disabling sequence is:
> ```
> CSI > Pp m
>           Set/reset key modifier options, xterm.  Set or reset resource-
>           values used by xterm to decide whether to construct escape
>           sequences holding information about the modifiers pressed with
>           a given key.
> 
>           The first parameter Pp identifies the resource to set/reset.
>           The second parameter Pv is the value to assign to the
>           resource.
> 
>           If the second parameter is omitted, the resource is reset to
>           its initial value.  Values 3  and 5  are reserved for keypad-
>           keys and string-keys.
> 
>             Pp = 0  ⇒  modifyKeyboard.
>             Pp = 1  ⇒  modifyCursorKeys.
>             Pp = 2  ⇒  modifyFunctionKeys.
>             Pp = 4  ⇒  modifyOtherKeys.
> 
>           If no parameters are given, all resources are reset to their
>           initial values.
> ```

The values that can be "assigned" for _modiyOtherKeys_ are:

> ```
> modifyOtherKeys (class ModifyOtherKeys)
>                Like modifyCursorKeys, tells xterm to construct an escape
>                sequence for ordinary (i.e., "other") keys (such as "2") when
>                modified by Shift-, Control-, Alt- or Meta-modifiers.  This
>                feature does not apply to special keys, i.e., cursor-, keypad-,
>                function- or control-keys which are labeled on your keyboard.
>                Those have key symbols which XKB identifies uniquely.
>
>                For example, this feature does not apply to special control-
>                keys (e.g., Escape, Tab, Enter, Backspace) Other control keys
>                (e.g., Control-I, Control-M, Control-H) may send escape
>                sequences when this feature is enabled.
> 
>                The default is "0":
> 
>                0    disables this feature.
> 
>                1    enables this feature for keys except for those with well-
>                     known behavior, e.g., Tab, Backarrow and some special
>                     control character cases which are built into the X11
>                     library, e.g., Control-Space to make a NUL, or Control-3
>                     to make an Escape character.
> 
>                     Except for those special cases built into the X11 library,
>                     the Shift- and Control- modifiers are treated normally.
>                     The Alt- and Meta- modifiers do not cause xterm to send
>                     escape sequences.  Those modifier keys are interpreted
>                     according to other resources, e.g., the metaSendsEscape
>                     resource.
> 
>                2    enables this feature for keys including the exceptions
>                     listed.  Xterm ignores the special cases built into the
>                     X11 library.  Any shifted (modified) ordinary key sends an
>                     escape sequence.  The Alt- and Meta- modifiers cause xterm
>                     to send escape sequences.
> 
>                The Xterm FAQ has an extended discussion of this feature, with
>                examples:
>                https://invisible-island.net/xterm/modified-keys.html
> ```

Broken out from https://github.com/jwilm/alacritty/issues/3091#issuecomment-565715265, and https://github.com/jwilm/alacritty/issues/2688#issuecomment-516987534
@AmaiKinono

Patryk27:
Current implementation's only partial and not thoroughly tested yet; though it already makes Emacs recognize Shift+Enter, which works wonders for my org-mode!

Anyway, I wanted to ask if I'm going the right route - classic: it'd be pity if I were to spend a few nights to hear that something could've been simplified from day zero 🙂 

cc @chrisduerr

---

Closes #3101

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -773,7 +773,8 @@ impl<'a> Deserialize<'a> for ModeWrapper {
 
             fn expecting(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
                 f.write_str(
-                    "a combination of AppCursor | AppKeypad | Alt | Vi, possibly with negation (~)",
+                    "a combination of appcursor | appkeypad | alt | vi | search, possibly with \
+                     negation (~)",
                 )
             }
 
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -325,6 +325,94 @@ impl<T: EventListener> Execute<T> for Action {
 }
 
 impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
+    fn xterm_modifiers(&mut self) -> u8 {
+        let mods = self.ctx.modifiers();
+        let mut n = 0;
+
+        if mods.shift() {
+            n |= 1;
+        }
+
+        if mods.alt() {
+            n |= 2;
+        }
+
+        if mods.ctrl() {
+            n |= 4;
+        }
+
+        n + 1
+    }
+
+    fn process_received_char(&mut self, c: char) {
+        let len = c.len_utf8();
+
+        if len == 1
+            && !self.ctx.modifiers().is_empty()
+            && self.ctx.terminal().mode().contains(TermMode::MODIFY_OTHER_KEYS1)
+        {
+            // xterm's definition of `other` in `modifyOtherKeys` is somewhat
+            // vague, mostly stemming out of xterm's inner design:
+            //
+            // https://github.com/ThomasDickey/xterm-snapshots/blob/f8a406455c18ec6f3de1cf2c642f7b7f0312d362/input.c#L547
+            //
+            // That's why our implementation in here is a mere approximation -
+            // it was done by analyzing "what xterm does" and trying to shoehorn
+            // it in here.
+            //
+            // To avoid duplicating comments, the implementation below follows
+            // as pure code, with comments extracted into tests at th bottom of
+            // this file.
+
+            let key = c as u8;
+            let xmods = self.xterm_modifiers();
+            let mods = self.ctx.modifiers();
+
+            let mut allowed = c == '-' || c == '=' || c == ',' || c == '.' || c.is_ascii_digit();
+
+            if key == 9 || key == 13 {
+                allowed |= !mods.alt();
+            }
+
+            if key == 32 {
+                allowed |= mods.ctrl() && mods.alt();
+            }
+
+            if key > 32 && key < 46 {
+                allowed |= mods.ctrl() || mods.alt();
+            }
+
+            // TODO check: ['!', '#', '$', '%', '&', '*', '(', ')', '_', '+', ':', '"', '<', '>',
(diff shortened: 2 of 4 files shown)


alacritty/src/config/bindings.rs
```diff
@@ -773,7 +773,8 @@ impl<'a> Deserialize<'a> for ModeWrapper {
 
             fn expecting(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
                 f.write_str(
-                    "a combination of AppCursor | AppKeypad | Alt | Vi, possibly with negation (~)",
+                    "a combination of appcursor | appkeypad | alt | vi | search, possibly with \
+                     negation (~)",
```
chrisduerr, line 777: This should be a separate PR.

alacritty/src/input.rs
```diff
@@ -324,6 +324,94 @@ impl<T: EventListener> Execute<T> for Action {
...
+            //
+            // To avoid duplicating comments, the implementation below follows
+            // as pure code, with comments extracted into tests at th bottom of
+            // this file.
+
+            let key = c as u8;
+            let xmods = self.xterm_modifiers();
+            let mods = self.ctx.modifiers();
+
+            let mut allowed = c == '-' || c == '=' || c == ',' || c == '.' || c.is_ascii_digit();
```
chrisduerr, line 371: This doesn't make much sense, does it? Most of these aren't "other" keys at all.

chrisduerr (changes requested):
I personally still question if this functionality is useful at all. It seems like an atrocious solution to the "problem" and only used by applications that really shouldn't be used (like terminal Emacs).

I'd be curious if there's any "good citizen" terminal application that actually has a useful application for this.

alacritty/src/input.rs
Patryk27: Well, to be honest what I've done is to mimic xterm behavior - and after activating modifyOtherKeys=1, xterm _does_ send new escape codes for those characters; but yeah, the definition of "other" is pretty vague here.

> I'd be curious if there's any "good citizen" terminal application that actually has a useful application for this.

vim? 😅 

Out of other solutions, there's also https://sw.kovidgoyal.net/kitty/keyboard-protocol.html - overall it's a better approach, but I'm not sure about its compatibility; I'll check later - if tmux and some other applications implement it, it might be more profitable one to implement than modifyOtherKeys.

Patryk27: Actually, `CSI u` seems to be supported quite nicely - I'll try implementing that one instead.

chrisduerr: `CSI u` is a much more reasonable approach to solving this and other problems.

Patryk27:
kk, I'll be implementing `CSI u` instead - so I'm closing this merge request for now; thanks for responding!

puremourning:
> I'd be curious if there's any "good citizen" terminal application that actually has a useful application for this.

Vim.

PR #5297, opened 2021-07-02
Add option to apply opacity to all background colors

### LINKED ISSUE #741: Transparency improvements

kchibisov:
In some cases it could be desired to apply 'background_opacity'
to all background colors instead of just 'colors.primary.background',
thus adding an 'colors.opaque_background_colors' option to control that.

Fixes #741.

--
Based on #4196.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.10.0-dev
 
+### Added
+
+- Option `colors.transparent_background_colors` to allow applying opacity to all background colors
+
 ### Changed
 
 - `ExpandSelection` is now a configurable mouse binding action
+
+### Removed
+
+- Config option `background_opacity`, you should use `window.opacity` instead
 
 ## 0.9.0
 
@@ -21,7 +29,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Support for `ipfs`/`ipns` URLs
 - Mode field for regex hint bindings
-- Option `colors.opaque_background_colors` to allow applying opacity to all background colors
 
 ### Fixed
 
@@ -41,10 +48,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Broken bitmap font rendering with FreeType 2.11+
 - Crash with non-utf8 font paths on Linux
 - Newly installed fonts not rendering until Alacritty restart
-
-### Removed
-
-- Config option `background_opacity`, you should use `window.background_opacity` instead
 
 ## 0.8.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Support for `ipfs`/`ipns` URLs
 - Mode field for regex hint bindings
+- Option `colors.opaque_background_colors` to allow applying opacity to all background colors
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Crash with non-utf8 font paths on Linux
 - Newly installed fonts not rendering until Alacritty restart
 
+### Removed
+
+- Config option `background_opacity`, you should use `window.background_opacity` instead
+
 ## 0.8.0
 
 ### Packaging
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - `ExpandSelection` is now a configurable mouse binding action
-
-### Removed
-
 - Config option `background_opacity`, you should use `window.opacity` instead
 
 ## 0.9.0
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -63,6 +63,13 @@
   #     - buttonless: Title bar, transparent background and no title bar buttons
   #decorations: full
 
+  # Background opacity
+  #
+  # Window opacity as a floating point number from `0.0` to `1.0`.
+  # The value `0.0` is completely transparent and `1.0` is opaque.
+  #background_opacity: 1.0
+
+
   # Startup Mode (changes require restart)
   #
   # Values for `startup_mode`:
@@ -359,12 +366,6 @@
   #     args: ["Hello, World!"]
   #
   #command: None
-
-# Background opacity
-#
-# Window opacity as a floating point number from `0.0` to `1.0`.
-# The value `0.0` is completely transparent and `1.0` is opaque.
-#background_opacity: 1.0
 
 #selection:
   # This string contains all characters that are used as separators for
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -69,7 +69,6 @@
   # The value `0.0` is completely transparent and `1.0` is opaque.
   #opacity: 1.0
 
-
   # Startup Mode (changes require restart)
   #
   # Values for `startup_mode`:
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -67,7 +67,7 @@
   #
   # Window opacity as a floating point number from `0.0` to `1.0`.
   # The value `0.0` is completely transparent and `1.0` is opaque.
-  #background_opacity: 1.0
+  #opacity: 1.0
 
 
   # Startup Mode (changes require restart)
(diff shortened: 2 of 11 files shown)


chrisduerr:
Is this basically a rewrite due to the changes made to the codebase since the original PR? Looks like it doesn't have a ton in common.

kchibisov:
Yeah, it's just a rewrite, since the codebase was changed in a way that `colors` are now in `ui_config`, which was the main complain back in the days.

kchibisov:
I wonder whether we should move `background_opacity` somewhere else in a config rather than keeping at the root. I think we can move it to `window.background_opacity`, what do you think of it?

chrisduerr:
> I wonder whether we should move background_opacity somewhere else in a config rather than keeping at the root. I think we can move it to window.background_opacity, what do you think of it?

I think this has been discussed in the past and I agree that `window` would be a good place to have this. Alternatively maybe in `colors`, but I think I prefer `window`. IIRC there was some reason why we didn't move it in the past, but that might have been resolved when the colors were moved out of `alacritty_terminal`.

kchibisov:
I've moved background_opacity to window, should be good now.

alacritty/src/config/color.rs
```diff
@@ -29,6 +30,25 @@ impl Colors {
...
+            vi_mode_cursor: Default::default(),
+            selection: Default::default(),
+            normal: Default::default(),
+            bright: Default::default(),
+            dim: Default::default(),
+            indexed_colors: Default::default(),
+            search: Default::default(),
+            line_indicator: Default::default(),
+            hints: Default::default(),
+            opaque_background_colors: true,
```
chrisduerr, line 47: I wonder if it wouldn't be worth it to change this to `transparent_background_colors`, just so the default is `false`.

I also think that people struggle more to understand 'opaque' compared to 'transparent'. Though our existing window opacity does already introduce this concept.

alacritty/src/display/content.rs
```diff
@@ -350,11 +350,13 @@ impl RenderableCell {
...
     #[inline]
-    fn compute_bg_alpha(bg: Color) -> f32 {
+    fn compute_bg_alpha(content: &mut RenderableContent<'_>, bg: Color) -> f32 {
         if bg == Color::Named(NamedColor::Background) {
             0.
-        } else {
+        } else if content.config.ui_config.colors.opaque_background_colors {
             1.
+        } else {
+            content.config.ui_config.background_opacity()
```
chrisduerr, line 359: In theory we could take a reference to ui_config as argument to this method, just to keep things minimal and reusable in theory. But it doesn't matter *that* much.

alacritty/src/display/mod.rs
```diff
@@ -484,12 +484,17 @@ impl Display {
         config: &Config,
         search_state: &SearchState,
     ) {
+        let mut grid_cells = {
+            let cols = self.size_info.columns();
+            let lines = self.size_info.screen_lines();
+            Vec::with_capacity(cols * lines)
+        };
```
chrisduerr, line 491: This seems like an optimization unrelated to this PR? I've benchmarked this particular change myself previously and it did not have any impact on performance. Have you made any other observations or was this just an unrelated test you were performing?

alacritty.yml
```diff
@@ -63,6 +63,13 @@
   #     - buttonless: Title bar, transparent background and no title bar buttons
   #decorations: full
 
+  # Background opacity
+  #
+  # Window opacity as a floating point number from `0.0` to `1.0`.
+  # The value `0.0` is completely transparent and `1.0` is opaque.
+  #background_opacity: 1.0
```
chrisduerr, line 70: I wonder if it's worth it to change this to just `opacity`? I think `window.opacity` should be pretty clear by itself, that way we could take this opportunity to make things a little more concise.

chrisduerr (changes requested):
Looks mostly good to go tbh. If you have the time let me know what you think about the remaining feedback. I'm happy to help out with resolving these issues.

alacritty/src/display/mod.rs
kchibisov: I just wrote that way, since it's always better not to do reallocs if you know upfront the size of allocation. It could show no difference for you, since your system if fast or your allocator is performing well, but you couldn't know how it'll work out with 'not smart enough' allocator or highly loaded system.

chrisduerr: > It could show no difference for you, since your system if fast or your allocator is performing well, but you couldn't know how it'll work out with 'not smart enough' allocator or highly loaded system.

That's a fair point, it's probably not going to be **worse**. Though this will definitely cause a bigger allocation than necessary since often the screen is likely not entirely full. Since this happens a bunch ideally the allocator won't have to ask the system anyway though.

CHANGELOG.md
```diff
@@ -7,10 +7,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+
+- Option `colors.transparent_background_colors` to allow applying opacity to all background colors
+
 ### Changed
 
 - `ExpandSelection` is now a configurable mouse binding action
 
+### Removed
+
+- Config option `background_opacity`, you should use `window.opacity` instead
```
chrisduerr, line 20: I don't think we added things to the `Removed` section in the past until it was actually removed? I think this should just be a `Changed` entry.

alacritty.yml
```diff
@@ -312,6 +319,13 @@
   #
   #indexed_colors: []
 
+  # Transparent backgrounds with transparency enabled
```
chrisduerr, line 322: This phrasing with the double "transparent" sounds a bit odd.

```suggestion
  # Transparent cell backgrounds
```

I'd suggest keeping the headline short like this to just encourage people to read the expanded explanation below instead of confusing them unnecessarily.

alacritty.yml
```diff
@@ -312,6 +319,13 @@
...
+  #
+  # Whether or not `window.opacity` applies to all backgrounds or only to the
+  # default background. When set to `false` all cells will be transparent
+  # regardless of their background color.
```
chrisduerr, line 326:
```suggestion
  # Whether or not `window.opacity` applies to all cell backgrounds or only to the
  # default background. When set to `false` all cells will be transparent
  # regardless of their background color.
```

I think explicitly pointing out cells here makes sense, since it is possible to change the terminal background in general. We don't just mean that a changed terminal background is transparent, but that each individual cell background is.

We could also change "default background" to "terminal background". I think many people probably don't see a "normal" cell as a cell with a default background but rather a cell with a transparent background.

The second part of the paragraph is also completely wrong. When set to `false`, cells that do not have the default background are **opaque**, not transparent.

```suggestion
  # Whether or not `window.opacity` applies to all backgrounds or only to the
  # default background. When set to `true` all cells will be transparent
  # regardless of their background color.
```

alacritty.yml
```diff
@@ -63,6 +63,13 @@
   #     - buttonless: Title bar, transparent background and no title bar buttons
   #decorations: full
 
+  # Background opacity
+  #
+  # Window opacity as a floating point number from `0.0` to `1.0`.
+  # The value `0.0` is completely transparent and `1.0` is opaque.
+  #opacity: 1.0
```
chrisduerr, line 70: Do you think we should leave this as `opacity` or change it to `transparency`? I personally think it's not worth it to change it since most Alacritty users will likely be used to the existing behavior already and inverting it will just cause unnecessary confusion. But one could argue that it's more consistent with the new `transparent_background_colors` option.

alacritty/src/display/content.rs
```diff
@@ -350,11 +350,13 @@ impl RenderableCell {
     /// using the named input color, rather than checking the RGB of the background after its color
     /// is computed.
     #[inline]
-    fn compute_bg_alpha(bg: Color) -> f32 {
+    fn compute_bg_alpha(config: &UiConfig, bg: Color) -> f32 {
         if bg == Color::Named(NamedColor::Background) {
             0.
-        } else {
+        } else if config.colors.transparent_background_colors {
             1.
```
chrisduerr, line 357: This is wrong, isn't it? Looks like you just changed the name from `opaque` to `transparent` but you didn't invert the behavior.

alacritty/src/display/mod.rs
chrisduerr: I really don't want to change this in this particular PR. I don't think it's uncontroversial enough to just slip it in.

chrisduerr (changes requested):
Should be good assuming that all my suggestions are fine with you.

alacritty.yml
chrisduerr: This is still unchanged?

alacritty.yml
chrisduerr: This is also still unchanged.

alacritty/src/display/mod.rs
```diff
@@ -490,6 +490,7 @@ impl Display {
         for cell in &mut content {
             grid_cells.push(cell);
         }
+
```
chrisduerr, line 493: I kinda like it without this newline, since it keeps all the `content.` stuff together. Not that big of a deal though.

chrisduerr (changes requested):

alacritty.yml
kchibisov: Yeah, I've discarded `alacritty.yml` changes at some point, since I was updating config to test things.

chrisduerr: So fix it again? I'm puzzled why you're re-requesting a review without addressing the comments.

kchibisov: The push failed I guess, should be fine now.

chrisduerr (approved):
LGTM

PR #5298, opened 2021-07-03
Add buffer for PTY reads during terminal lock

chrisduerr:
Before this patch, Alacritty's PTY reader would always try to read the
PTY into a buffer and then wait for the acquisition of the terminal lock
to process this data. Since locking for the terminal could take some
time, the PTY could fill up with the thread idling while doing so.

As a solution, this patch keeps reading to a buffer while the terminal
is locked in the renderer and starts processing all buffered data as
soon as the lock is released.

This has halfed the runtime of a simple `cat` benchmark from ~9 to ~4
seconds when the font size is set to `1`. Running this patch with
"normal" grid densities does not appear to make any significant
performance differences in either direction.

One possible memory optimization for the future would be to use this
buffer for synchronized updates, but since this currently uses a dynamic
buffer and would be a bit more cluttered, it has not been implemented in
this patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Regression in rendering performance with dense grids since 0.6.0
 - Crash/Freezes with partially visible fullwidth characters due to alt screen resize
 - Incorrect vi cursor position after invoking `ScrollPageHalfUp` action
+- Slow PTY read performance with extremely dense grids
 
 ## 0.8.0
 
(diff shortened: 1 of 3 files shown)


chrisduerr:
I think this should be good to go.

It's important to note that on my benchmark machine even with very dense grids this did not make any differences in our usual tests. Neither positive nor negative. These are the results for a 1000x250 grid:

tmp

So I'd assume that in practice there aren't a ton of examples where this is relevant. Which raised the question for me if it is worth it to include this patch. However there definitely are edge cases where this does become relevant, so since this does not regress anything else, a performance improvement in these edge cases seems worth it.

PR #5308, opened 2021-07-06
Fix no text selection clear when leave vi mode

a5ob7r:
This bug comes from 530de00049c2afcc562d36ccdb3e6afa2fe396a5. This causes unintentional text selection expansion and highlight from left end of the selection to bottom of viewport.

An sample video which is show the bug

https://user-images.githubusercontent.com/12132068/124625853-23733180-deb9-11eb-813d-2ec04bba8f17.mp4

Reproduction steps (with default keybinding)

1. Enter vi mode (by `Ctrl+Shift+Space`).
2. Move vi mode cursor arbitrarily or don't move.
3. Start text selection (by `v`).
4. Move vi mode cursor arbitrarily or don't move.
5. Leave vi mode (by `Ctrl+Shift+Space`).
6. Go to step `1.`.

PR #5310, opened 2021-07-06
Fix unintentional text selection range changing after leave vi mode

a5ob7r:
This bug comes from 530de00049c2afcc562d36ccdb3e6afa2fe396a5. The vi cursor movement changes text selection range when it is on vi mode. On the other hand the cursor movement doesn't change the range when it isn't on vi mode. So preserve text selection range by toggling vi mode early.

This is another approach of #5308


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Incorrect vi cursor position after invoking `ScrollPage*` action
 - Slow PTY read performance with extremely dense grids
 - Crash when resizing during vi mode
+- Unintentional text selection range changing after leave vi mode
```
chrisduerr, line 22:
```suggestion
- Unintentional text selection range change after leaving vi mode
```

chrisduerr (changes requested):
Changing the order of operations makes sense to me, thanks for the PR.

chrisduerr (approved):

PR #5313, opened 2021-07-07
Reload config on SIGUSR1

### LINKED ISSUE #4590: Feature: Reload ckon SIGUSR1

ncfavier:
Fixes #4590.

Seems to work well, I've had Alacritty freeze a couple times after getting SIGUSR1 but I can't reproduce it so it might have been related to something else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Support for `ipfs`/`ipns` URLs
 - Mode field for regex hint bindings
+- Reload configuration on SIGUSR1
 
 ### Fixed
 
--- a/alacritty/Cargo.toml
+++ b/alacritty/Cargo.toml
@@ -39,6 +39,7 @@ dirs = "3.0.1"
 gl_generator = "0.14.0"
 
 [target.'cfg(not(windows))'.dependencies]
+signal-hook = "0.3.9"
 xdg = "2"
 
 [target.'cfg(not(target_os = "macos"))'.dependencies]
--- a/extra/alacritty.man
+++ b/extra/alacritty.man
@@ -62,6 +62,8 @@ Defines the window title [default: Alacritty]
 .TP
 \fB\-\-working\-directory\fR <working\-directory>
 Start the shell in the specified working directory
+.SH "SIGNALS"
+\fBSIGUSR1\fR makes Alacritty reload its configuration.
 .SH "SEE ALSO"
 See the alacritty github repository at https://github.com/alacritty/alacritty for the full documentation.
 .SH "BUGS"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,6 +40,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
+ "signal-hook",
  "time",
  "unicode-width",
  "wayland-client",
@@ -81,6 +82,7 @@ dependencies = [
  "serde_json",
  "serde_yaml",
  "signal-hook",
+ "signal-hook-mio",
  "unicode-width",
  "vte",
  "winapi 0.3.9",
@@ -986,6 +988,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "mio-uds"
+version = "0.6.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "afcb699eb26d4332647cc848492bbc15eafb26f08d0304550d5aa1f612e066f0"
+dependencies = [
+ "iovec",
+ "libc",
+ "mio",
+]
+
+[[package]]
 name = "miow"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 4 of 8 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Support for `ipfs`/`ipns` URLs
 - Mode field for regex hint bindings
+- Reload configuration on SIGUSR1
```
chrisduerr, line 14: I'd be curious where else this feature can/needs to be advertised. I don't think we should just expect people to figure it out on their own.

Maybe we could document the live config reload in our features.md and then add the ability to manually trigger a reload to that?

alacritty/Cargo.toml
```diff
@@ -34,6 +34,7 @@ libc = "0.2"
 unicode-width = "0.1"
 bitflags = "1"
 dirs = "3.0.1"
+signal-hook = "0.3.9"
```
chrisduerr, line 37: Seems like there have been a lot of updates to `signal-hook`, we should update the version in `alacritty_terminal/Cargo.toml` to keep things consistent.

alacritty/src/main.rs
```diff
@@ -180,9 +181,20 @@ fn run(
...
+    // Reload configuration when receiving a SIGUSR1 signal.
+    let config_path = config.ui_config.config_paths[0].clone();
+    std::thread::spawn(move || {
+        let mut signals = Signals::new(&[sigconsts::SIGUSR1]).unwrap();
+        for sig in signals.forever() {
+            if sig == sigconsts::SIGUSR1 {
+                event_proxy.send_event(Event::ConfigReload(config_path.clone()));
+            }
+        }
+    });
```
chrisduerr, line 196: I think it would be nice to make this part of the monitor. Just so we have config reload watching all in one place.

I'd also always prefer spawning a named thread, though I personally think that a separate thread just for this is a bit unfortunate. Especially because we can expect that 99.9% of users will never use, nor even know about this feature, since live config reload exists.

extra/alacritty.man
```diff
@@ -62,6 +62,8 @@ Defines the window title [default: Alacritty]
 .TP
 \fB\-\-working\-directory\fR <working\-directory>
 Start the shell in the specified working directory
+.SH "SIGNALS"
+\fBSIGUSR1\fR makes Alacritty reload its configuration.
```
chrisduerr, line 66: Documenting things here seems like a smart idea.

chrisduerr (changes requested):
Alacritty freezing is a pretty serious issue that should definitely get investigated properly. Since `alacritty_terminal` does some signal handling I'd be curious if the two conflict in some way.

alacritty/src/main.rs
ncfavier: > I think it would be nice to make this part of the monitor. Just so we have config reload watching all in one place.

Sure.

> I'd also always prefer spawning a named thread, though I personally think that a separate thread just for this is a bit unfortunate. Especially because we can expect that 99.9% of users will never use, nor even know about this feature, since live config reload exists.

I agree, but I don't really know what other (safe) way there is. I couldn't find a way to plug signal events into the glutin event loop directly. We could make signal_hook set a flag, but then when do we read the flag?

chrisduerr: Receiving it through the event loop would be ideal, but I'd assume that's probably going to be problematic since it's not really known for its extensibility. At least not without spawning worker threads anyway.

Maybe things could at least be combined with the monitor thread in some way? We already got that one around anyway.

ncfavier: Yeah, I'm not sure how. The config watcher thread listens on a channel, and I don't see a straightforward way to have a signal handler wake that channel up.

alacritty/Cargo.toml
ncfavier: Done.

CHANGELOG.md
ncfavier: Done.

ncfavier:
> Alacritty freezing is a pretty serious issue that should definitely get investigated properly. Since `alacritty_terminal` does some signal handling I'd be curious if the two conflict in some way.

Looking at signal-hook's description, it seems to be able to handle such cases just fine.

I'll use my fork for a while and see if I can observe the behaviour again.

chrisduerr:
> Looking at signal-hook's description, it seems to be able to handle such cases just fine.

Generally signal-hook tries very hard to allow for this, yes. Signals are just always kinda funky and we are using a crate that builds on top of signal-hook to combine it with mio.

ncfavier:
OK, the freeze happens when I restart the ibus daemon twice in rapid succession. I could reproduce it on 0.8.0 so it has nothing to do with this PR.

chrisduerr (changes requested):
My original review feedback still stands.

ncfavier:
You mean the thread issue? I really don't think there's a way out of that, or that it's that much of a problem. (The config watcher already spawns two threads...)

I found the chan-signal crate which I thought could be helpful at first, but it's [deprecated](https://github.com/BurntSushi/chan-signal#this-crate-has-reached-its-end-of-life-and-is-now-deprecated) and as you can see they recommend the thread approach instead.

chrisduerr:
Well I'm not particularly interested in this issue right now. So if you're not willing to look more into this, I don't have any interest in doing the work myself.

ncfavier:
I have looked into this quite deeply. The main problem seems to be that `notify` uses an `std::sync::mpsc::channel` (see https://github.com/notify-rs/notify/issues/160), which doesn't have a `select`-like interface. So if a thread is listening for `notify` events, it can't do anything else.

This seems to be fixed in `notify` v5, but it doesn't have a stable release and doesn't support debounced events yet.

Here are the other alternatives I looked at:

- Using `signal-hook`'s other interfaces.
  - The `flag` module can set a boolean flag when a signal occurs. We could check that flag in the main event loop, but then there's no guarantee that the reload will be triggered right away. This also requires splitting the logic across several files.
  - `signal-hook-mio`: `glutin` uses `winit` which uses `mio` internally and only on X11, we can't possibly rely on that.
  - `low_level::register` allows registering signal handlers (in an `unsafe` block), but the things we can do in such a handler are [extremely limited](https://docs.rs/signal-hook/0.3.9/signal_hook/low_level/fn.register.html#safety). `signal-hook` does provide an async-signal-safe [Channel](https://docs.rs/signal-hook/0.3.9/signal_hook/low_level/channel/struct.Channel.html) type, but that channel "is not responsible for wakeups", so it won't help us. See [this blog post](https://vorner.github.io/2018/06/28/signal-hook.html) for more background.
- Using a different crate.
  - `chan-signal` is deprecated and [spawns a thread](https://github.com/BurntSushi/chan-signal/blob/master/src/lib.rs#L250).
  - `signal-msg` uses `simple-signal` which [spawns a thread](https://github.com/swizard0/rust-simple-signal/blob/master/src/lib.rs#L96).
  - `addy` [spawns a thread](https://github.com/xvrqt/addy/blob/master/src/lib.rs#L922).
  - `signal`only allows waiting for a signal synchronously, so we'd need to spawn a thread.

Keep in mind that threads are *cheap*, as opposed to processes. A sleeping thread barely consumes any cycles or memory.

ncfavier:
After e0f0cdfb8aed4c99819019fd3d515d3584e01a45 (presumably the update to winit 0.25.0), Alacritty crashes on SIGUSR1 with this message:

> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 4, kind: Interrupted, message: "Interrupted system call" }', /home/n/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.25.0/src/platform_impl/linux/x11/mod.rs:360:54

ncfavier:
Closing as https://github.com/alacritty/alacritty/pull/5399 removes my need for this.

PR #5314, opened 2021-07-08
Fix Void Linux link in the install guide

pbrkr:
While skimming the install docs I followed the Void Linux link but ended up at a blog with an expired certificate. Checking on [Distro Watch](https://distrowatch.com/table.php?distribution=void) shows that https://voidlinux.org is the correct link for Void Linux.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -131,7 +131,7 @@ Compiles out of the box for 14.2
 
 #### Void Linux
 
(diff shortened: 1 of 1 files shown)


chrisduerr (approved):
> While skimming the install docs I followed the Void Linux link but ended up at a blog with an expired certificate.

Must have confused that with Manjaro. Thanks for the patch.

PR #5335, opened 2021-07-16
Fix copying interrupted tab characters

### LINKED ISSUE #5084: Selection not copied after tab character

chrisduerr:
Fixes #5084.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Search without vi mode not starting at the correct location when scrolled into history
 - Crash when starting a vi mode search from the bottommost line
 - Original scroll position not restored after canceling search
+- Clipboard copy skipping non-empty cells when encountering an interrupted tab character
 
 ## 0.8.0
 
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -406,7 +406,7 @@ impl<T> Term<T> {
 
             // Skip over cells until next tab-stop once a tab was found.
             if tab_mode {
-                if self.tabs[column] {
+                if self.tabs[column] || cell.c != ' ' {
                     tab_mode = false;
                 } else {
                     continue;


chrisduerr:
After some discussion on IRC, the previous PR attempts by @PabloMarcendo turned out to be closer to the solution than I had initially assumed. Though this version should still perform better than the proposed alternatives.

I've decided to still copy the tab character as a tab, instead of spaces. I think that is the simplest solution and it represents the grid state most accurately. However there are some arguments for copying spaces instead.

For future reference the following IRC discussion should shed some insights on the trade-offs:

<details>
<summary>IRC Log</summary>

```
2021-06-07 00:07:43	@kchibisov	UndeadLeech: you've got why #5084 is happening, right?
2021-06-07 00:07:55	@UndeadLeech	Mostly, yeah.
2021-06-07 00:08:21	@UndeadLeech	It's similar to the issue with wide chars, not sure if you're familiar with that.
2021-06-07 00:09:03	@kchibisov	I mean, the issue is not really similar.
2021-06-07 00:09:09	@kchibisov	The problem is '\r' here.
2021-06-07 00:08:54	@UndeadLeech	\r is just a way to reposition the cursor.
2021-06-07 00:09:46	@kchibisov	Yeah, and it makes you do '\t', 'X', TAB_STOP
2021-06-07 00:10:00	@kchibisov	And we skip everything until the TAB_STOP in our selection logic.
2021-06-07 00:10:21	@kchibisov	Just to copy only '\t' and whitespaces from empty cells.
2021-06-07 00:11:01	@kchibisov	The question is what we except here as a result, since logically 'X' should be after TAB_STOP, but it's not.
2021-06-07 00:10:54	@UndeadLeech	Actually I'm not sure I should have closed the associated PRs now that I think about it.
2021-06-07 00:11:36	@kchibisov	it's not due to cursor repositioning.
2021-06-07 00:11:32	@UndeadLeech	What do you mean?
2021-06-07 00:12:07	@UndeadLeech	It's because of cursor repositioning, because without it there's no way to write inside of a tab.
2021-06-07 00:12:47	@kchibisov	yeah, exactly.
2021-06-07 00:12:52	@kchibisov	And the question is how we should handle it.
2021-06-07 00:13:20	@UndeadLeech	Probably instead of skipping until tabstop, skip until !whitespace || tabstop, right?
2021-06-07 00:13:28	@UndeadLeech	Or alternatively, remove the tab.
2021-06-07 00:14:12	@kchibisov	Yeah, but copying '\t' 'X' is incorrect here...
2021-06-07 00:14:05	@UndeadLeech	I'm not really familiar with your example.
2021-06-07 00:14:50	@kchibisov	'x X\r[\t ]'
2021-06-07 00:14:56	@kchibisov	printf 'x X\r[\t ]'
2021-06-07 00:14:58	@kchibisov	The one from the issue
2021-06-07 00:14:48	@UndeadLeech	Okay, so what would you expect to copy?
2021-06-07 00:15:07	@UndeadLeech	Whitespace instead of \t?
2021-06-07 00:15:43	@kchibisov	I guess so.
2021-06-07 00:15:40	@UndeadLeech	Yeah, I think that's actually what other terminals do.
2021-06-07 00:16:23	@kchibisov	That's what vte does at least from what I can see.
2021-06-07 00:15:55	@UndeadLeech	So as I've said, writing inside of a tab should remove the tab from the cell then.
2021-06-07 00:16:00	@UndeadLeech	Kitty and XTerm too.
2021-06-07 00:16:59	@kchibisov	But remove in what way, like really erase from the cell or just from the selection.
2021-06-07 00:17:25	@kchibisov	Since if we're going to remove it from the cell we should handle it early on and check whether we're currently in a tab...
2021-06-07 00:16:57	@UndeadLeech	Probably actually erasing it from the cell. Because at that point it's not a tab anymore.
2021-06-07 00:17:10	@UndeadLeech	That's why I initially said this is similar to wide chars.
2021-06-07 00:17:31	@UndeadLeech	Though that codepath is getting kinda icky.
2021-06-07 00:18:01	@UndeadLeech	One thing that might be interesting to test is writing whitespace, or writing an 'X' and then whitespace.
2021-06-07 00:18:19	@UndeadLeech	At that point some terminals might handle it as tab again? That would show if they actually nuke the cell or not.
2021-06-07 00:20:41	@kchibisov	I think they nuke it actually...
2021-06-07 00:20:42	@UndeadLeech	I don't think Kitty does. Though I haven't found a great testcase yet.
2021-06-07 00:21:00	@UndeadLeech	Just `printf 'x  \r[\t]'` will kinda work. But it's not great.
2021-06-07 00:21:42	@UndeadLeech	Also there's kinda two different scenarios: Someone writes a tab first, then overwrites it. Or someone writes something and then inserts a tab.
2021-06-07 00:23:24	@UndeadLeech	Oooh, that's interesting.
2021-06-07 00:23:52	@UndeadLeech	If you do tab first and then write into it, Kitty still copies the tab. But if you write first and then insert a tab, it doesn't.
2021-06-07 00:24:08	@UndeadLeech	Repro: `reset; printf "\e[0;0HA\tB\e[0;3HX"; sleep infinity`
2021-06-07 00:24:59	@UndeadLeech	The biggest difference to Alacritty being that even if it's copying the tab, it is not just skipping all characters.
2021-06-07 00:29:02	@kchibisov	Well, in that case we can rewrite put_tab logic to ignore putting a tab if we have 'something' already there...
2021-06-07 00:29:13	@kchibisov	But I'm still not sure if it's a 'correct' solution.
2021-06-07 00:30:24	@kchibisov	I mean, 'something' ~ something which is not eq to empty cell and in range from cursor position to next tab stop.
2021-06-07 00:32:17	@kchibisov	Though, the one thing I can say for sure, is that the user excepts to copy whay they see on the screen.
2021-06-07 00:32:10	@UndeadLeech	The `put_tab` logic would solve the issue itself, but not my example.
2021-06-07 00:32:45	@kchibisov	So changing only 'selection' logic is some kind of compromise here.
2021-06-07 00:32:42	@UndeadLeech	Yeah, but tabs can't really be seen.
2021-06-07 00:33:35	@kchibisov	Yeah, but 'X' can be seen here.
2021-06-07 00:33:48	@kchibisov	So removing 'tab' from selection is fine in that case.
2021-06-07 00:33:18	@UndeadLeech	Yeah, but it could still be a tab in front of it, not whitespace.
2021-06-07 00:34:25	@kchibisov	Well, it should be on a tab stop then, if it's not on a tab stop, it means that 'tab' isn't working, so it's just a whitespace.
2021-06-07 00:34:13	@UndeadLeech	Dunno.
2021-06-07 00:34:55	@UndeadLeech	I feel like most of these probably work fine. Might just be a matter of choosing what's simplest.
2021-06-07 00:35:36	@kchibisov	The selection thing is simpler.
2021-06-07 00:35:33	@UndeadLeech	Which selection thing? But yeah, I think modifying the grid state itself is trouble.
2021-06-07 00:36:32	@kchibisov	I mean, that removing tabs only in selected text and not on a grid, since touching grid is costly.
2021-06-07 00:36:17	@UndeadLeech	That's kinda sketchy though. 
2021-06-07 00:37:00	@kchibisov	I mean, the situation is sketchy already.
2021-06-07 00:36:34	@UndeadLeech	At that point we basically admit that our grid state's fucked, but we don't fix it.
2021-06-07 00:36:57	@UndeadLeech	And you'll get weird stuff like the tab suddenly popping back up, after writing whitespace again.
2021-06-07 00:38:08	@kchibisov	Well, if you move everything to '\0' or some other identification of empty cell other than 'whitespace' it shouldn't be an issue.
2021-06-07 00:38:00	@UndeadLeech	Writing whitespace isn't the only way one can delete content again.
2021-06-07 00:38:10	@UndeadLeech	"Erase in line right" would write \0s
2021-06-07 00:39:12	@kchibisov	yeah, that's true, and putting extra logic for tabs there is messy at least.
2021-06-07 00:40:03	@UndeadLeech	That's why I thought just keeping the tab in the grid and copying it might be the simplest solution.
2021-06-07 00:41:15	@kchibisov	Yeah, but the text you copy won't match the text you paste visually.
2021-06-07 00:41:03	@UndeadLeech	Again, visually a tab often just looks like a single whitespace.
2021-06-07 00:41:08	@UndeadLeech	Like in the prompt for example.
2021-06-07 00:41:54	@kchibisov	I mean it's usually 8 whitespaces.
2021-06-07 00:41:49	@UndeadLeech	Yeah, I'm aware. But the prompt is one example where it cannot be reliably identified.
2021-06-07 00:43:39	@UndeadLeech	XTerm copies just a single whitespace in both examples btw.
2021-06-07 00:43:49	@UndeadLeech	So it's more consistent than Kitty at least.
2021-06-07 00:44:15	@UndeadLeech	Oh actually XTerm can't copy tabs at all.
2021-06-07 00:44:21	@UndeadLeech	So it's probably worthless. 
2021-06-07 00:45:02	@kchibisov	Maybe it has a setting to copy tabs...
2021-06-07 00:44:46	@UndeadLeech	If you try to copy tabs, it just copies every single cell as whitespace.
2021-06-07 00:44:57	@UndeadLeech	So a tab is basically just interpreted as an escape, not as an actual character.
2021-06-07 00:45:10	@UndeadLeech	I can test VTE, gimme a sec.
2021-06-07 00:50:29	@UndeadLeech	Huh, that's strange.
2021-06-07 00:51:05	@UndeadLeech	In the second example, where I write the `X` inside the tab between `A` and `B`, the resulting text in VTE is actually `A X\tB`.
2021-06-07 00:52:27	@UndeadLeech	I hate everything lol.
```

</details>

PR #5358, opened 2021-07-21
Add support for `CSI u` bindings

### LINKED ISSUE #3101: Implement "modifyOtherKeys"

Patryk27:
Closes #3101<sup>1</sup>

The implementation is based on:
- http://www.leonerd.org.uk/hacks/fixterms/
- https://github.com/alexherbo2/alacritty-extended-keys/blob/master/keys.yml<br>(there are many similar keybinding-repos for Alacritty, and there's nothing special about this one - just happened to be the first one that I stumbled upon, and I considered worth mentioning it.)

I've checked the keycodes we yield using `showkey -a` and [`emacs -nw`](https://gist.github.com/gnachman/b4fb1e643e7e82a546bc9f86f30360e4), and most of them seem to work correctly (and for the rest I've left some comments in the code).

Please lemme know what you think and if my approach is even valid in the first place - terminals are startlingly complicated piece of software 😅 

## Different approaches

I've also considered implementing [the neat thing that Kitty supports](https://sw.kovidgoyal.net/kitty/keyboard-protocol/), but for my taste it's an investment not worth that much (to give a peek - my adventure with `CSI u` started mainly because I wanted for terminal Emacs to recognize Ctrl+Enter 🙃).

<sup>1</sup> disclaimer: this pull request doesn't implement support for `modifyOtherKeys`, but - [as discussed](https://github.com/alacritty/alacritty/pull/5293) - going with `CSI u` is probably a better solution

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -531,6 +531,18 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
             SearchAction::SearchFocusPrevious;
     );
 
+    bindings.extend(vt_function_key_bindings());
+    bindings.extend(platform_key_bindings());
+
+    let csiu_bindings = csiu_key_bindings(&bindings);
+    bindings.extend(csiu_bindings);
+
+    bindings
+}
+
+fn vt_function_key_bindings() -> Vec<KeyBinding> {
+    let mut bindings = vec![];
+
     //   Code     Modifiers
     // ---------+---------------------------
     //    2     | Shift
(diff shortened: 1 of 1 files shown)


Patryk27:
Also, if this binding-based approach is correct, I'd gladly extend it to more modifiers - I've started with just a few to get a sneak peek.

chrisduerr (changes requested):
I doubt a keybinding-based solution to this could ever provide any meaningful solution.

Limiting things to just 0-9/A-Z just isn't sufficient, that barely helps with Ctrl+Shift but that's about it. I don't think it would be in the spirit of the protocol to half-ass it like this.

I'd have to check some other implementations to get some insights on the common limitations/approaches taken, but I doubt they usually define all the bindings manually. That would certainly create a massive matrix of bindings which quickly becomes a problem.

listx:
Hello, I just wanted to ask a question here because I am interested in native `CSI u` support for Alacritty.

> I'd have to check some other implementations to get some insights on the common limitations/approaches taken, but I doubt they usually define all the bindings manually. That would certainly create a massive matrix of bindings which quickly becomes a problem.

We just want some functionality that converts the recognized keypresses (with or without modifiers) and sends the `CSI u`-based encodings instead of whatever we're doing today, correct?

The "recognize keypresses with or without modifiers" bit seems to be in around https://github.com/alacritty/alacritty/blob/ec4dc32688cfc71686c24a4edc59fd7427a6bf33/alacritty/src/input.rs#L799-L807, so it looks like we just need to add an if-block somewhere around there like this:

```
if csi_u_flag_is_enabled_in_config {
  // encode using CSI u encoding scheme and add/replace into `bytes`
}
```

Patryk27:
@listx I've already checked this approach [while trying to implement `modifyOtherKeys` some time ago](https://github.com/alacritty/alacritty/pull/5293/files#diff-0fd7077a384dd164d220e2b11ab393e3e34862d97762b6682dcb52219804f736R327), and while `CSI u` is conceptually a bit simpler than `MOK`, there - as far as I remember - remained a core issue: what should happen to keystrokes that already are encoded via ANSI - i.e. whether `Ctrl+C` should be sent as-is, or using `CSI`.

For instance, [Kitty's approach](https://sw.kovidgoyal.net/kitty/keyboard-protocol/) is to have various levels of "what should be sent via CSI" - from "some selected keystrokes", up to "everything"; but, if I understand it correctly, it still requires keeping some hard-coded keymap of "for the most basic level of `CSI u`, we don't encode Ctrl+A, Ctrl+B, ...".

So a tl;dr would be: a flag such as `csi_u_flag_is_enabled_in_config` would work, but we probably wouldn't like to just encode all of the keystrokes at once, and imho implementing Kitty's approach might be better (but harder, too).

huyz:
> I had to get around this problem by making Alacritty send different custom sequence for Ctrl+I, to be parsed in tmux separately.

@listx I'm a noob to tmux; do you have sample config lines for this tmux parsing to bring back the `Ctrl-I` sequence?

listx:
> > I had to get around this problem by making Alacritty send different custom sequence for Ctrl+I, to be parsed in tmux separately.
> 
> @listx I'm a noob to tmux; do you have sample config lines for this tmux parsing to bring back the `Ctrl-I` sequence?

Sure, have a look at https://github.com/listx/syscfg/blob/ec89be49694f1befeb8a2dedaa46001632e41360/tmux/.tmux.conf#L348.

PR #5376, opened 2021-07-28
Draft: Feature feedoutput (pipeing commandoutput back to the pty in keybindings)

noctux:
Hey, thank you for alacritty. This PR is mostly intended to scratch a small personal itch of mine with the software. Currently it's in alpha state and only contains the minmum amount of work to obtain a working sample.

What this commit does is to allow `commands` that are attached to keybindings to feed back output from stdout into the running terminal session.

Exemplary usecase/Rational: Due to wayland's gui isolation on linux (and potentially also on platforms such as windows/osx), it is not that easy any more to type in passwords into a given window automatically, at least not in the way it was possible with X11 using xdotools; so my password management solution (build around pass, rofi/dmenu and xdotools) broke. The easiest (and probably only secure) way to fix this is by integrating an "insert-password" funcitonality into the respective applications - such as alacritty - themselves. Currently it looks like this (where [rofi-pass](https://github.com/carnager/rofi-pass) opens a graphical fuzzy finder across my passwords and outputs the respective password on stdout):

```yaml
key_bindings:
  - { key: P,         mods: Control,                    command: { program: "rofi-pass", args: [], feedoutput: true } }
```

What is nice that nearly all functionality is already there, and therefore the additional code, and probably also maintainability burden, are thus quite minimal. As a side effect, with minimal additional code, we could also attach this to `hint`-handles as well, even though I yet do not see immediate usecases for that one:

```yaml
hints:
  enabled:
   - regex: "(ipfs:|ipns:|magnet:|mailto:|gemini:|gopher:|https:|http:|news:|file:|git:|ssh:|ftp:)\
             [^\u0000-\u001F\u007F-\u009F<>\"\\s{-}\\^⟨⟩`]+"
     # Input the curled sourcecode into the terminal. The integrated way to do "curl | bash" (yes, I'm kidding)
     command:  { program: "curl", args: [], feedoutput: true }
     binding:
       key: U
       mods: Control|Shift
```

Currently, the PR is in draft state, and definitly needs more work (which I'm willing to do, but I would just like some early feedback and guidance to avoid work going into wrong directions). Areas that I will definitely improve when the concept is ironed out:
1. Write documentation for the feature
2. Write tests (if possible, have to look into it)
3. Validate that the option is not present at unsuitable places (e.g.: Bell program)

Additionally, I would value input on:
- naming of the option: You probably know the drill: There are only 2 hard problems in CS: cache-invalidation, naming things, and off-by-one errors. I'm not really happy with `feedoutput`, but all alternatives I came up with (`passthrough`, `typeback`, `feedback`, `redirect`, `to_pty`) did not seem too natural either
- Should it work for hints? Or does it seem out of place?
- Is deamon.rs really the correct place for `capture_output`?
- Is `write_to_pty` the correct function to input into the session?

Thank you for reading this far. As said, I'd be glad to make any changes if you actually consider adding this feature to alacritty.

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -183,7 +183,7 @@ impl Options {
             // `Arg::min_values(1)` is set.
             let program = String::from(args.next().unwrap());
             let args = args.map(String::from).collect();
-            options.command = Some(Program::WithArgs { program, args });
+            options.command = Some(Program::WithArgs { program, args, feedoutput: false });
         }
 
         if matches.is_present("hold") {
--- a/alacritty/src/daemon.rs
+++ b/alacritty/src/daemon.rs
@@ -74,3 +74,22 @@ where
             .map(|_| ())
     }
 }
+
+#[cfg(not(windows))]
+pub fn capture_output<I, S>(program: &str, args: I) -> Option<Vec<u8>>
+where
+    I: IntoIterator<Item = S> + Copy,
+    S: AsRef<OsStr>,
+{
+    let output = Command::new(program)
+        .args(args)
+        .stdin(Stdio::null())
+        .stderr(Stdio::null())
+        .output().ok()?;
+
+    if output.status.success() {
+        Some(output.stdout)
+    } else {
+        None
+    }
+}
--- a/alacritty/src/daemon.rs
+++ b/alacritty/src/daemon.rs
@@ -85,7 +85,8 @@ where
         .args(args)
         .stdin(Stdio::null())
         .stderr(Stdio::null())
-        .output().ok()?;
+        .output()
+        .ok()?;
 
     if output.status.success() {
         Some(output.stdout)
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -43,7 +43,7 @@ use crate::cli::Options as CLIOptions;
 use crate::clipboard::Clipboard;
 use crate::config::ui_config::{HintAction, HintInternalAction};
 use crate::config::{self, Config};
-use crate::daemon::start_daemon;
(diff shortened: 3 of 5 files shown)


chrisduerr:
Is there any usecase for this other than passwords? Because realistically that is a terrible one. Integrating that into every single application makes no sense at all when wayland is fully capable of allowing certain applications to just act as virtual keyboards.

noctux:
Hmm good question. Of course you can extend this approach to various similar usecases that can then again be partly discharged to other, application specific usecases (e.g.: chooser to insert utf8-characters by human readable name, opening an actual graphical file-selection dialogue via zenity and getting the filepath into terminal applications, etc.pp.), but given the nature of this feature, all those solutions could of course be fullfilled by the combination of a tool+virtual keyboard solution, if a proper and secure one can be found for a particular platform.

At the same time, I believe passwords to actually be a good usecase :) The solutions I know, such as ydootol, actually sacrifice a lot of the security benefits one gets from using wayland in the first place by allowing all applications of a given user to write to uinput. Of course I could try to instead apply this selectivly to just my password-manager-binary, but in both cases, the solution would not work on shared systems where I'm not root... At the same time, all autotype based solution always carry the risk of "misfocusing/focus-stealing" when the typing starts (another window gets into foregreground, e.g. an active chatwindow with a new message...). Having this directly within an application mitigates that risk.

Re "patching every application": There was this moment when I've realized that there is a browser, and a lot of terminals running, and nothing else. So for me personally, having that feature in alacritty itself would be sufficient. But of course, this is a quite personal thing, and if neither you nor any other human being see any value in this feature, there probably is no value of upstreaming it :)

Issue #5387, opened 2021-08-02
Allow logging from dependencies when logging at trace level


sbosnick:
OS: Linux
Version: 0.8.0
Linux/BSD: Wayland, Gnome

Currently `alacritty` filters the log message to exclude all messages that come from its dependencies
such as [winit](https://crates.io/crates/winit) or (with a Wayland compositor) 
[smithay-client-toolkit](http://crates.io/crates/smithay-client-toolkit/0.12.3). I propose that
this should change to allow Error and Warn level messages to come through from such crates
when `alacritty` is run with the `-vvv` option.

The `-vvv` option enables Trace level log messages so additional detail from dependant crates
won't be completly out of place.

This additional debugging information would allow for addtional options for diagnosing errors.
For example, if this had been in place then it would have shown an Error level log messages from
`smithay-client-toolkit` for #4780. The "fail silently" part of my comment on that issue is
in part an `alacritty` issue, not just a `winit` issue.

PR #5388, opened 2021-08-02
Allow logging from dependencies at trace level

sbosnick:
This change allows the logging of Error and Warn level messages from dependencies when
the overall logging level is set to Trace (though the "-vvv" option).

Fixes: #5387﻿

### DIFF
--- a/alacritty/src/logging.rs
+++ b/alacritty/src/logging.rs
@@ -13,7 +13,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 
 use glutin::event_loop::EventLoopProxy;
-use log::{self, Level};
+use log::{self, Level, LevelFilter};
 
 use crate::cli::Options;
 use crate::event::Event;
@@ -103,7 +103,7 @@ impl log::Log for Logger {
         let target = &record.target()[..index];
 
         // Only log our own crates.
-        if !self.enabled(record.metadata()) || !ALLOWED_TARGETS.contains(&target) {
+        if !self.enabled(record.metadata()) || !is_allowed_target(record.level(), target) {
             return;
         }
 
@@ -147,6 +147,16 @@ fn create_log_message(record: &log::Record<'_>, target: &str) -> String {
     // Drop extra trailing alignment.
     message.truncate(message.len() - alignment);
     message
+}
+
+fn is_allowed_target(level: Level, target: &str) -> bool {
+    use Level::{Error, Warn};
+    use LevelFilter::Trace;
+
+    match (level, log::max_level()) {
+        (Error, Trace) | (Warn, Trace) => true,
+        _ => ALLOWED_TARGETS.contains(&target),
+    }
 }
 
 struct OnDemandLogFile {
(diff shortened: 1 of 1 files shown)


alacritty/src/logging.rs
```diff
@@ -149,6 +149,16 @@ fn create_log_message(record: &log::Record<'_>, target: &str) -> String {
     message
 }
 
+fn is_allowed_target(level: Level, target: &str) -> bool {
+    use Level::{Error, Warn};
+    use LevelFilter::Trace;
```
chrisduerr, line 154: Please don't put imports inside functions.

alacritty/src/logging.rs
```diff
@@ -149,6 +149,16 @@ fn create_log_message(record: &log::Record<'_>, target: &str) -> String {
     message
 }
 
+fn is_allowed_target(level: Level, target: &str) -> bool {
```
chrisduerr, line 152:
```suggestion
/// Check if log messages from a crate should be logged.
fn is_allowed_target(level: Level, target: &str) -> bool {
```

Wording might be a bit awkward, but I think a little comment would be nice in case it's not immediately obvious what "allowed target" means in this context.

chrisduerr (changes requested):
Thanks for sending a PR.

alacritty/src/logging.rs
sbosnick: Changed in latest commit.

alacritty/src/logging.rs
sbosnick: Added comment to latest commit.

alacritty/src/logging.rs
```diff
@@ -102,8 +102,8 @@ impl log::Log for Logger {
         let index = record.target().find(':').unwrap_or_else(|| record.target().len());
         let target = &record.target()[..index];
 
-        // Only log our own crates.
-        if !self.enabled(record.metadata()) || !ALLOWED_TARGETS.contains(&target) {
+        // Only log our own crates (except when logging at Level::Trace).
```
chrisduerr, line 105:
```suggestion
        // Only log our own crates, except when logging at Level::Trace.
```

alacritty/src/logging.rs
```diff
@@ -149,6 +149,14 @@ fn create_log_message(record: &log::Record<'_>, target: &str) -> String {
     message
 }
 
+/// Check if log messages from a crate should be logged.
+fn is_allowed_target(level: Level, target: &str) -> bool {
+    match (level, log::max_level()) {
+        (Level::Error, LevelFilter::Trace) | (Level::Warn, LevelFilter::Trace) => true,
```
chrisduerr, line 155: I think on the latest Rust version it should be possible to do `Level::Warn | Level::Error` inside the tuple. I'm not sure when that was introduced, but our CI should be able to verify that.

chrisduerr (changes requested):

alacritty/src/logging.rs
sbosnick: I made this change.

alacritty/src/logging.rs
sbosnick: I have made this change and it compiles on my machine (Rust version 1.54.0). I will wait until this passes CI to resolve this conversation.

chrisduerr (approved):
Thank you for the PR.

alacritty/src/logging.rs
chrisduerr: It's newer than 1.50.0, so we can't use this unfortunately. I've fixed it myself.

PR #5399, opened 2021-08-05
Watch base config paths in addition to canonicalized paths

ncfavier:
Implements the suggestion at https://github.com/alacritty/alacritty/issues/2237#issuecomment-479242735 to watch both the base config paths and their canonicalised versions.

Tested to make sure both things work.


alacritty/src/config/monitor.rs
```diff
@@ -16,17 +16,20 @@ const DEBOUNCE_DELAY: Duration = Duration::from_millis(10);
...
-                None
-            },
-        })
-        .collect();
+    // Also watch the targets of symbolic links
+    paths.append(
+        &mut paths
+            .iter()
+            .filter_map(|path| match fs::canonicalize(&path) {
+                Ok(canonical_path) if canonical_path != *path => Some(canonical_path),
```
chrisduerr, line 24: We maybe should make sure this is actually a symlink. This way it would also add a path twice if it contains segments like `/../`, which is unnecessary. So to ensure the number of watched paths is as small as possible, we should only add symlinks twice.

chrisduerr (changes requested):
This still needs a changelog entry.

Which systems did you test this on?

alacritty/src/config/monitor.rs
ncfavier: Yes, and the current version is wrong anyway: we're not filtering the base paths anymore so we might end up watching a non-existent path.

ncfavier:
Tested on NixOS on x86_64-linux.

ncfavier:
I made it not ignore `Rename` events because I couldn't find a reason for that in the first place.

Tested on the same platform.

chrisduerr:
> I made it not ignore Rename events because I couldn't find a reason for that in the first place.

Because renaming a file does not modify the configuration file in any relevant way.

ncfavier:
It does if you're renaming `~/.config/alacritty/foo` to `~/.config/alacritty/alacritty.yml` (an unusual scenario, but still no reason to ignore it).

chrisduerr:
No it doesn't, because the content of the configuration file doesn't change.

ncfavier:
1. move `alacritty.yml` to `foo`
2. edit `foo`
3. move `foo` back to `alacritty.yml`

(don't do these too fast or it might get debounced into a single `Write` event)

I've tested this: 0.9.0 doesn't reload, this PR reloads.

chrisduerr:
If you want to make life hard on yourself, I guess you could do that. I suppose it doesn't hurt.

alacritty/src/config/monitor.rs
```diff
@@ -73,17 +88,15 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventProxy) {
...
-                        continue;
-                    }
-
+                | DebouncedEvent::Chmod(path)
+                    if paths.contains(&path) =>
+                {
                     // Always reload the primary configuration file.
                     event_proxy.send_event(Event::ConfigReload(paths[0].clone()));
-                },
+                }
```
chrisduerr, line 99:
```suggestion
                },
```

alacritty/src/config/monitor.rs
```diff
@@ -15,18 +14,34 @@ const DEBOUNCE_DELAY: Duration = Duration::from_millis(10);
...
-                error!("Unable to canonicalize config path {:?}: {}", path, err);
-                None
-            },
-        })
-        .collect();
+pub fn watch(paths: Vec<PathBuf>, event_proxy: EventProxy) {
+    let paths = {
+        let mut paths_tmp = Vec::new();
+        for path in paths {
+            if path.exists() {
```
chrisduerr, line 21: Is checking for the existence of a path really necessary? In theory files can be deleted at any time and iirc paths are only passed to the monitor when the config could be loaded. So effectively this only covers for the case that config files have been deleted in the nanoseconds between parsing and monitor creation? Correct me if I'm wrong on this.

alacritty/src/config/monitor.rs
```diff
@@ -15,18 +14,34 @@ const DEBOUNCE_DELAY: Duration = Duration::from_millis(10);
...
-            Ok(path) => Some(path),
-            Err(err) => {
-                error!("Unable to canonicalize config path {:?}: {}", path, err);
-                None
-            },
-        })
-        .collect();
+pub fn watch(paths: Vec<PathBuf>, event_proxy: EventProxy) {
+    let paths = {
+        let mut paths_tmp = Vec::new();
```
chrisduerr, line 19: The unfortunate naming of this variable indicates that this is probably not the greatest way to solve this problem. I don't think it should be necessary to create a separate vector.

alacritty/src/config/monitor.rs
```diff
@@ -15,18 +14,34 @@ const DEBOUNCE_DELAY: Duration = Duration::from_millis(10);
...
+        for path in paths {
+            if path.exists() {
+                paths_tmp.push(path.clone());
+
+                // Also watch the targets of symbolic links
+                match path.symlink_metadata() {
+                    Ok(metadata) if metadata.file_type().is_symlink() => {
+                        match path.canonicalize() {
```
chrisduerr, line 27: By only canonicalizing paths that are symlinks, we're leaving us open for paths to be watched multiple times even though they're not symlinks.

Take the following example:
 - /tmp/../alacritty.yml
 - /alacritty.yml

Both are the same file, but only one of them needs to be watched.

Ultimately this should never happen without user error, but doing it "right" is basically the same effort as not, so I see no reason why we wouldn't do it properly.

alacritty/src/config/monitor.rs
```diff
@@ -15,18 +14,34 @@ const DEBOUNCE_DELAY: Duration = Duration::from_millis(10);
...
+pub fn watch(paths: Vec<PathBuf>, event_proxy: EventProxy) {
+    let paths = {
+        let mut paths_tmp = Vec::new();
+        for path in paths {
+            if path.exists() {
+                paths_tmp.push(path.clone());
+
+                // Also watch the targets of symbolic links
+                match path.symlink_metadata() {
+                    Ok(metadata) if metadata.file_type().is_symlink() => {
```
chrisduerr, line 26: These operations should be chained. We shouldn't have three levels of conditionals (match -> if -> match) when all of that can be done in a single map/filter/whatever chain.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - `ExpandSelection` is now a configurable mouse binding action
+- The config watcher now watches the base config paths in addition to symbolic link targets
```
chrisduerr, line 13: I find this phrasing a bit difficult. "Base config path" for me sounds canonicalized, but in this case means the symlink itself. Maybe making it a bit simpler would make things easier to understand.

I'd propose something like this maybe:
```suggestion
- Reload configuration files when their symbolic link is replaced
```

I think that this documentation should be easier to understand, more directly correlated to the changes a user would expect and more concise. The wording "their symbolic link is replaced" is still a bit imprecise and awkward, though it's probably fine.

chrisduerr (changes requested):

alacritty/src/config/monitor.rs
ncfavier: I couldn't find a function in the standard library that canonicalises paths without resolving symlinks. What are you thinking?

chrisduerr: We can still check if the file is a symlink, that is no problem. But we should **always** push the canonicalized path and only push the non-canonicalized one if the file is a symlink. Basically exactly the opposite way of what you've done right now.

ncfavier: Indeed, that makes more sense.

alacritty/src/config/monitor.rs
ncfavier: You're right, that's a useless check.

alacritty/src/config/monitor.rs
ncfavier: The problem with a more functional approach is that I'd need a `flat_map` along the way (because each input path can yield 0, 1 or 2 output paths), and Rust doesn't have a `yield` keyword so it seems like a pain in the behind to generate iterators ad hoc.

I could however make the code itself flatter.

chrisduerr: In this case itself, there's only ever one path being yielded. I'm not talking about the loop in general, just about everything past the symlink metadata check.

ncfavier: Oh, ok. I just pushed code implementing the new logic, I don't think this discussion applies anymore.

alacritty/src/config/monitor.rs
ncfavier: My Rust knowledge is probably at fault here, but the alternatives I see are:
- using iterators, which runs into the problem I was describing
- modifying the vector in place, which sounds a lot more cumbersome and inefficient than creating a new one

alacritty/src/config/monitor.rs
ncfavier: It seems like the formatter is [complaining](https://builds.sr.ht/~undeadleech/job/568957#task-rustfmt-77) about this

alacritty/src/config/monitor.rs
chrisduerr: Just to clarify: I'll try and play around with a few options myself when I got the time. I just need a little time until I can get around to it.

alacritty/src/config/monitor.rs
chrisduerr: You're right. I'm really surprised by this since we have `match_block_trailing_comma = true` in our config.

I feel like this might be somewhat of a bug, because it doesn't happen without the `if`.

PR #5401, opened 2021-08-06
Apply 'font.glyph_offset' for underline/strikeout

### LINKED ISSUE #5304: Apply glyph offset to underline and strikeout

kchibisov:
Fixes #5304.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - `ExpandSelection` is now a configurable mouse binding action
+- `font.glyph_offset` is now applied to underline/strikeout
 
 ## 0.9.0
 
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -538,7 +538,7 @@ impl Display {
             });
         }
 
-        let mut rects = lines.rects(&metrics, &size_info);
+        let mut rects = lines.rects(&metrics, &size_info, &config.ui_config.font.glyph_offset);
 
         if let Some(vi_mode_cursor) = vi_mode_cursor {
             // Indicate vi mode by showing the cursor's position in the top right corner.
--- a/alacritty/src/renderer/rects.rs
+++ b/alacritty/src/renderer/rects.rs
@@ -9,6 +9,7 @@ use alacritty_terminal::term::cell::Flags;
 use alacritty_terminal::term::color::Rgb;
 use alacritty_terminal::term::SizeInfo;
 
+use crate::config::ui_config::Delta;
 use crate::display::content::RenderableCell;
 use crate::gl;
 use crate::gl::types::*;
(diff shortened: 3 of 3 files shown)


chrisduerr (changes requested):
I question the correctness of this, especially when it comes to the X offset.

Fundamentally the idea behind these offsets should be to better align the content within the cells, so why would we offset the X position? Our calculations should always position the X position correctly relative to the cell, so I see no reason to ever apply the offset here.

The Y position is based on metrics, so it's more likely that there is an actual issue that would have to be fixed, but there's no concrete correlation between glyph position and underline metrics, so there's no guarantee that a single offset would work for both.

Is there any concrete example for an issue this patch fixes for you?

kchibisov:
I think it's pretty straight forward to test this, you just run alacritty with `alacritty -o font.glyph_offset.x=10 font.glyph_offset.y=10` and look on underline/strikeout text with and without a patch.
picture

chrisduerr:
The assumptions you're making are flawed though. You're using the offset on a font that has already correctly positioned glyphs, effectively moving them to an incorrect location. When realistically the goal should be to move incorrectly positioned glyphs to a correct location.

As you can see with your cursor, your usage of the glyph offset breaks the terminal. The underlines are aligned properly with the grid.

kchibisov:
Yeah, but underlines are attached to glyphs, since underlines/strikeout metrics are being taken from the glyphs, and if you move glyphs you should also move underlines/strikeouts with them. It breaks terminal with big glyph offset, but if you're using small glyph offset with `font.offset` it doesn't break anything, but puts strikeout in particular in a correct (so it crosses the glyphs) position.

chrisduerr:
> Yeah, but underlines are attached to glyphs, since underlines/strikeout metrics are being taken from the glyphs

Not the X position though. The X position is solely based on the cell position, not on the glyph.

The glyphs aren't moved *out of* place, they're moved *into* place. Talking just about X position the goal of the offset is to place glyphs correctly within the cell, while underlines will **always** be placed at the correct location since they always go from the start to the end of the cell. So what argument is there to apply an offset here?

kchibisov:
As it was discussed on IRC we can only apply `glyph_offset` to `y` coordinate, so in a cases where `offset.y` is negative underlines and strikeouts "wouldn't be out of place".

chrisduerr:
Appling the y offset could *sometimes* be correct, but certainly not always. I'm not sure if it's more commonly required or not.

If people usually use the glyph_offset to adjust for additional space created by the offset, then this would be a reasonable patch. But if people primarily use the glyph_offset to fix positioning of the glyphs themselves then I'm not sure.

Have you ever run into any fonts where adjusting the glyph_offset was necessary without adjusting the offset?

kchibisov:
>Have you ever run into any fonts where adjusting the glyph_offset was necessary without adjusting the offset?

I think yes, and it was due to hinting or fractinal font sizes or something like that, or I just wanted to 'center' glyphs in the cell sometimes, since sometimes some letters too close to the top of the cell.

chrisduerr:
Really don't know how to feel about this, but I guess we can just go with applying the offset just to `y` and rolling with that until somebody complains. If nobody does it probably works fine.

kchibisov:
nah, don't want this anymore.

PR #5414, opened 2021-08-13
Touch-to-click

bytbox:
This is a baby step towards touch support: tapping will click, but no other touch functionality is available. It does most of what *I* want, but of course I can't vouch for anyone else...

Because of the simplicity of this change, there are a few strange behaviors. None of them seem to be "biting" or potentially harmful, but here's the ones I know of:

- The message bar is not handled correctly. If you tap on the message bar, that gets interpreted as a click behind the message bar.
- Dragging is not detected. That doesn't just mean you can't scroll; it also means that if you drag, it interprets it as a click at the last place your finger was.
- Multitouch is totally ignored. If you put one finger down, then another, then lift the first, it interprets that as a "click". When you lift the second, that's another "click".

Note (as mentioned in the commits) that this code essentially contains a workaround for winit issue 1996. The main part of that workaround is that it necessitates keeping track of a `TouchState` object --- that'll be necessary eventually anyway, so I don't feel guilty.

I have added no tests, expecting that touch behavior will change substantially in the future, and those tests would need to be immediately re-written. Tests have been modified only to keep them compiling (necessary cuz of the additional variable tracking TouchState).

Finally, a question: to detect the difference between a drag (to be ignored, for now) and a tap, it's necessary to have some sort of threshold, probably in distance moved. Is this worthy of adding a configuration option to alacritty.yml? I think the only real alternative is a hidden constant somewhere.

Thanks,
Scott

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -161,6 +161,7 @@ pub struct ActionContext<'a, N, T> {
     pub terminal: &'a mut Term<T>,
     pub clipboard: &'a mut Clipboard,
     pub mouse: &'a mut Mouse,
+    pub touch: &'a mut TouchState,
     pub received_count: &'a mut usize,
     pub suppress_chars: &'a mut bool,
     pub modifiers: &'a mut ModifiersState,
@@ -301,6 +302,16 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     #[inline]
     fn mouse(&self) -> &Mouse {
         self.mouse
+    }
+
+    #[inline]
+    fn touch_mut(&mut self) -> &mut TouchState {
+        self.touch
+    }
+
+    #[inline]
+    fn touch(&self) -> &TouchState {
+        self.touch
     }
 
     #[inline]
@@ -972,6 +983,19 @@ impl Mouse {
     }
 }
 
+/// State of touch.
+#[derive(Debug)]
+pub struct TouchState {
+    pub x: f64,
+    pub y: f64,
+}
+
+impl Default for TouchState {
+    fn default() -> TouchState {
+        TouchState { x: Default::default(), y: Default::default() }
+    }
+}
+
 /// The event processor.
 ///
 /// Stores some state from received events and dispatches actions when they are
@@ -979,6 +1003,7 @@ impl Mouse {
 pub struct Processor<N> {
     notifier: N,
     mouse: Mouse,
+    touch: TouchState,
     received_count: usize,
     suppress_chars: bool,
     modifiers: ModifiersState,
@@ -1016,6 +1041,7 @@ impl<N: Notify + OnResize> Processor<N> {
             event_queue: Default::default(),
             modifiers: Default::default(),
             mouse: Default::default(),
+            touch: Default::default(),
             dirty: Default::default(),
         }
     }
@@ -1119,6 +1145,7 @@ impl<N: Notify + OnResize> Processor<N> {
                 terminal: &mut terminal,
                 notifier: &mut self.notifier,
                 mouse: &mut self.mouse,
+                touch: &mut self.touch,
                 clipboard: &mut clipboard,
                 received_count: &mut self.received_count,
                 suppress_chars: &mut self.suppress_chars,
@@ -1303,9 +1330,7 @@ impl<N: Notify + OnResize> Processor<N> {
                         processor.ctx.window().set_mouse_visible(true);
                         processor.mouse_wheel_input(delta, phase);
                     },
-                    WindowEvent::Touch(touch) => {
-                        processor.touch_input(touch)
-                    },
+                    WindowEvent::Touch(touch) => processor.touch_input(touch),
                     WindowEvent::Focused(is_focused) => {
                         if window_id == processor.ctx.window().window_id() {
                             processor.ctx.terminal.is_focused = is_focused;
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -986,13 +986,27 @@ impl Mouse {
 /// State of touch.
 #[derive(Debug)]
 pub struct TouchState {
+    pub start_x: f64,
+    pub start_y: f64,
     pub x: f64,
     pub y: f64,
 }
 
 impl Default for TouchState {
     fn default() -> TouchState {
-        TouchState { x: Default::default(), y: Default::default() }
+        TouchState {
+            start_x: Default::default(),
+            start_y: Default::default(),
+            x: Default::default(),
+            y: Default::default(),
+        }
+    }
+}
+
+impl TouchState {
+    /// Returns square of the distance travelled since the touch began.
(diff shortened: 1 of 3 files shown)


chrisduerr:
> The message bar is not handled correctly. If you tap on the message bar, that gets interpreted as a click behind the message bar.

This should probably be similar to the mouse interaction with the message bar. Clicking on it is clamped to the grid, but clicking on the "button" works to dismiss it. Ignoring it would be fine too, but I think the button should work (I have no way to test touch support myself).

> Dragging is not detected. That doesn't just mean you can't scroll; it also means that if you drag, it interprets it as a click at the last place your finger was.

This shouldn't be the case. While we don't need drag support, we need to be able to distinguish a click from a drag. That means at the very least we should probably figure out if the finger position has changed between press and release. If the finger was moved in between doesn't matter super much to me at this point, but might be worth considering too.

> Multitouch is totally ignored. If you put one finger down, then another, then lift the first, it interprets that as a "click". When you lift the second, that's another "click".

That sounds okay.

> Finally, a question: to detect the difference between a drag (to be ignored, for now) and a tap, it's necessary to have some sort of threshold, probably in distance moved. Is this worthy of adding a configuration option to alacritty.yml? I think the only real alternative is a hidden constant somewhere.

There is absolutely no way a configuration option will be added for such a technical and specific thing. It's our responsibility to find a good value that works reliably to remove that complexity from the user, just shoving everything one might not be certain about into configuration is just a lazy way to avoid finding a good solution really. Adding a constant is no problem and there's no reason to "hide" it, as it is done with many other things already in Alacritty's code.

chrisduerr (changes requested):
See my comment.

bytbox:
(I'm not done yet --- tomorrow I intend to get the button in the message bar to behave correctly.)

chrisduerr:
> I'm not done yet

Just request a review when you are.

bytbox:
We now handle message bar clicks just as the mouse does.

alacritty/src/event.rs
```diff
@@ -972,13 +983,41 @@ impl Mouse {
...
 }
 
+/// State of touch.
+#[derive(Debug)]
+pub struct TouchState {
+    pub start_x: f64,
+    pub start_y: f64,
+    pub x: f64,
+    pub y: f64,
+}
```
chrisduerr, line 993: Is it really necessary to separate this from what is already tracked in the `Mouse` struct? Just seems like a waste to me.

alacritty/src/event.rs
```diff
@@ -972,13 +983,41 @@ impl Mouse {
...
+            y: Default::default(),
+        }
+    }
+}
+
+impl TouchState {
+    /// Returns square of the distance travelled since the touch began.
+    pub fn dist2(&self) -> f64 {
+        (self.start_x - self.x).powi(2) + (self.start_y - self.y).powi(2)
+    }
```
chrisduerr, line 1010: Square distance should be irrelevant for this usecase. What matters is if the same cell is pressed or not.

alacritty/src/input.rs
```diff
@@ -674,6 +708,32 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
         }
     }
 
+    fn on_message_bar_click(&mut self) {
```
chrisduerr, line 711: The function name should describe what this method actually does. It's not a click handler for the message bar.

chrisduerr (changes requested):

alacritty/src/input.rs
bytbox: Right, it's a click to the box in the message bar. Appropriately renamed.

alacritty/src/event.rs
bytbox: I spent a while playing with different settings for the distance. The difficulty is: on my screen, it's very easy for a sloppy tap to start in one cell and jump to another. Some of this is just that I'm a klutz using a small font, but then, many users are klutzes! Having to fine-tune taps because they're not registering is extraordinarily irritating --- the whole point of touch is that it's faster and requires much less thought.

alacritty/src/event.rs
bytbox: I don't see a way to do without tracking `start_x` and `start_y` separately, since I see nothing analogous in `Mouse`. We could always place them in `Mouse` and call them `touch_start_x` and `touch_start_y`, but that's horrible. There are many other options (I won't make a list), but all feel similarly unnatural.

There are a few additional considerations:

- It would be nice if attempting to use the mouse and touch screen simultaneously didn't result in horrifically confusing misbehavior. This is made less likely by keeping the state info as separate as possible.
- Future features (I do intend to implement drag-to-scroll eventually, just not this month and not for this PR) are even more likely to need a dedicated `TouchState` object.

I also view `Mouse.x` and `TouchState.x` as conceptually very different, and therefore it's just "right" to keep them separate, but that's just some vague intuition that I can't possibly justify.

alacritty/src/event.rs
chrisduerr: The same could be argued about the mouse though and I'm not aware of anyone having trouble with it here. So I'm not sure why the touchscreen would be any different?

alacritty/src/event.rs
chrisduerr: > It would be nice if attempting to use the mouse and touch screen simultaneously didn't result in horrifically confusing misbehavior.

I really don't care much about this. Having the touchscreen solely as mouse emulation seems fine to me. There's absolutely no need to support both at the same time.

> I also view Mouse.x and TouchState.x as conceptually very different, and therefore it's just "right" to keep them separate, but that's just some vague intuition that I can't possibly justify.

I don't see how they're different at all. Both present the last pointer X position.

> I don't see a way to do without tracking start_x and start_y separately, since I see nothing analogous in Mouse.

There is. We have `block_hint_launcher` already which is set to `true` in `mouse_moved` whenever the mouse moved by one cell. This is used to prevent the user from opening a URL when trying to select it. Touch click recognition should behave similarly.

Though of course if we actually wanted mouse clicking to work **properly**, we shouldn't just emit a single click at some point in time. Realistically even with a touch click the press and release should be reported separately, otherwise applications will not work properly. So most of the time it really shouldn't be alacritty that determines if the mouse has moved or not, it should be the application handling the click. It's just that in some scenarios we're handling the click in Alacritty.

alacritty/src/event.rs
bytbox: At the end of the day, I can only report that it's empirically different for me. I realize that's not helpful.

bytbox:
Thanks for the input. I don't think we're going to converge, in particular

> Having the touchscreen solely as mouse emulation seems fine to me.

makes little sense to me, since both the types of inputs they accept and the desired behavior are substantially different.

I guess there's no reason for you to accept code you can't test and won't use. I'm perfectly happy, since I have a version that I can use. (I'm looking forward to adding scroll when I get a chance) I'll close this PR now.

Thanks for maintaining alacritty.

bytbox:
(For completeness of the link web: #3671.)

Issue #1702, opened 2018-10-23
Alacritty does not report desired window (re)size


mitchellwrosen:
Hello! Using a tiling window manager, I've found that my "fixed column" layout does not work properly. I did enough printf-debugging, comparing to `urxvt`, to discover that some X11 calls involving [size hints](https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/wm-normal-hints.html) are not returning any useful information in alacritty.

Does this make sense? Please let me know if I've provided enough information, I'm happy to keep drilling.

> Which operating system does the issue occur on?

Arch linux

> If on linux, are you using X11 or Wayland?

X11

chrisduerr:
Which window manager are you using specifically? In general Alacritty should work just fine with tiling window managers.

mitchellwrosen:
I'm using xmonad. It does work, but _not_ for the fixed column layout, which queries the application for how wide a "column" should be (which, for a terminal, depends on the font size).

chrisduerr:
The problem I see with this is that Alacritty doesn't actually take care of the window itself, so it's required that the library we use makes setting this possible.

It sounds like an extremely specific thing though, so that might take some time before someone can take a look at it. I'm also not certain how exactly this works in xmonad, does each Application get the size it requests or is this a minimum size requirement?

mitchellwrosen:
I'm not exactly sure myself, here is the relevant code from [xmonad-contrib](https://hackage.haskell.org/package/xmonad-contrib-0.15/docs/XMonad-Layout-FixedColumn.html):

```haskell
-- | Determine the width of @w@ given that we would like it to be @n@
--   columns wide, using @inc@ as a resize increment for windows that
--   don't have one
widthCols :: Int -> Int -> Window -> X Int
widthCols inc n w = withDisplay $ \d -> io $ do
    sh <- getWMNormalHints d w
    bw <- fmap (fromIntegral . wa_border_width) $ getWindowAttributes d w
    let widthHint f = f sh >>= return . fromIntegral . fst
        oneCol      = fromMaybe inc $ widthHint sh_resize_inc
        base        = fromMaybe 0 $ widthHint sh_base_size
    return $ 2 * bw + base + n * oneCol
```

So we call `getWMNormalHints`, which returns a `SizeHints`:

```haskell
data SizeHints = SizeHints
                   { sh_min_size     :: Maybe (Dimension, Dimension)
                   , sh_max_size     :: Maybe (Dimension, Dimension)
[...]

PR #5417, opened 2021-08-14
Add an option to ask X11/MacOS to use window resize increments (cell-precision resize)

### LINKED ISSUE #388: When resizing, use row/column precision, not pixel precision.

necauqua:
Hello
This is a feature that I really miss, and I knew about X11 EWMHs, so I've decided to implement it.

Tested it on AwesomeWM (which does support (re)size hints) and on macOS Big Sur - where I had the issue with logical pixels, but if you double the increment it works fine, although obviously it snaps by two cells at a time.

I am open to review comments on this, I am not sure I got the #[cfg]s right or about should the macOS part even be included and so on, and I am willing for this feature to finally be part of the program

Partially closes #388 and #1702, Windows has none of that, to my knowledge

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -626,6 +626,10 @@ impl Display {
             config.window.dynamic_padding,
         );
 
+        self.window.set_resize_increments(config.window.resize_increments.then(|| {
+            PhysicalSize::from((self.size_info.cell_width(), self.size_info.cell_height()))
+        }));
+
         // Update number of column/lines in the viewport.
         let message_bar_lines =
             message_buffer.message().map_or(0, |m| m.text(&self.size_info).len());
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -330,6 +330,7 @@ impl Window {
         };
 
         let builder = WindowBuilder::new()
+            .with_name(&identity.class.general, &identity.class.instance)
             .with_title(&identity.title)
             .with_visible(false)
             .with_transparent(true)
@@ -339,13 +340,6 @@ impl Window {
 
         #[cfg(feature = "x11")]
         let builder = builder.with_window_icon(icon.ok());
-
-        #[cfg(feature = "wayland")]
-        let builder = builder.with_app_id(identity.class.instance.to_owned());
-
-        #[cfg(feature = "x11")]
-        let builder = builder
-            .with_class(identity.class.instance.to_owned(), identity.class.general.to_owned());
 
         #[cfg(feature = "x11")]
         let builder = match &window_config.gtk_theme_variant {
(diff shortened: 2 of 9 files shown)


alacritty/src/display/mod.rs
```diff
@@ -406,6 +411,18 @@ impl Display {
...
+    fn update_resize_increments(&self, config: &Config) {
+        self.window.update_resize_increments(if config.ui_config.window.resize_increments {
+            let cell_size = PhysicalSize::<f32>::from((
+                self.size_info.cell_width(),
+                self.size_info.cell_height(),
+            ));
+            Some(cell_size.to_logical(self.window.dpr))
+        } else {
+            None
+        });
```
chrisduerr, line 423: Please never do this. The goal of patches should be to produce readable, not compressed code. Please extract these conditionals into separate statements.

alacritty/src/display/window.rs
```diff
@@ -436,6 +440,49 @@ impl Window {
...
 
+    #[cfg(all(feature = "x11", not(windows), not(target_os = "macos")))]
+    pub fn update_resize_increments(&self, increments: Option<LogicalSize<f32>>) {
+        // winit does not provide a proper public API to set X window hints at runtime
+        // without using the doc(hidden) xlib_connection or being cross-platform.
+        //
+        // It *does* have WindowBuilderExtUnix.with_resize_increments, however it
+        // only applies on window creation and you cannot change increments later
+        //
+        // So we have to do it manually instead
```
chrisduerr, line 451: No we don't. What we should do is sending a patch upstream to add this functionality to winit.

chrisduerr (changes requested):

alacritty/src/display/mod.rs
necauqua: I actually never realized that `bool::then` was stabilized, I'll refactor this

alacritty/src/display/window.rs
necauqua: You're right, I never thought about that, I'll try to add this to winit

Although it will delay this

alacritty/src/display/mod.rs
chrisduerr: `bool::then` is not an option due to our MSRV.

necauqua: Now, a year later, it is :)

alacritty/src/display/window.rs
necauqua: While I did what you asked for - now waiting for them to notice->review->hopefully merge->release - I want to object as in this very file I've seen a bunch of similar stuff - directly (and unsafely) talking to x11 or cocoa or windows - so this wouldn't have been an odd one out ¯\\\_(ツ)\_/¯

necauqua:
This is (now) a draft where I overrode winit to my branch (https://github.com/rust-windowing/winit/pull/2411) to test/demo what it would look like when/if it gets merged.

This also required to update glutin and fix a couple of places that broke because of that and I suspect those macos CI failures are related to that.

leaxoy:
Any progress on this?

necauqua:
@leaxoy waiting for winit 0.28 release which will have my changes for resize increments

kchibisov:
Closing in favor of #6675

PR #5423, opened 2021-08-16
Add check for CARGO_CFG_TARGET_OS, rustc-link-search for OpenBSD [ close #5422 ]

### LINKED ISSUE #5422: alacritty 0.9.0 / v0.10.0-dev fails to compile on OpenBSD 6.9-current
After installing as written in the `INSTALL.md`, and running `cargo build --release` at last step of compilation it fails to compile the program. At first glance it looks like missing something in the `build.rs` but I'm not a Rust expert to say this.

The same happens with `cargo install alacritty`.

### System

OS: OpenBSD 6.9-current
Version: failed to compile `alacritty v0.9.0` / `alacritty v0.10.0-dev` the error is the same
Linux/BSD: Xenocara, i3

### Logs

```
Building [=======================> ] 188/189: alacritty(bin)

error: linking with `cc` failed: exit status: 1
  |
  = note: [annoying stuff]
  = note: ld: error: unable to find library -lxkbcommon
          cc: error: linker command failed with exit code 1 (use -v to see invocation)

error: aborting due to previous error

error: could not compile `alacritty`

To learn more, run the command again with --verbose.
```

But I have `libxkbcommon`:
```
$ pkg_info -Q libxkbcommon
debug-libxkbcommon-1.3.0
libxkbcommon-1.3.0 (installed)
```

So I think I should probably say somewhere to get also the `/usr/local/lib/`, but this looks awkward as it could have failed somewhere else.

# Solution

While this get fixed, `RUSTFLAGS='-L/usr/local/lib' cargo build --release`.

In `alacritty/build.rs`, in the `main` function add the following:
- [x] check for target "openbsd"
- [x] add `println!(r"cargo:rustc-link-search=/usr/local/lib")`

1dotd4:
OpenBSD has libxkbcommon in /usr/local/lib which is not pointed from
rust. I've followed the following blog post for the CARGO_CFG_TARGE_OS.

https://kazlauskas.me/entries/writing-proper-buildrs-scripts.html

### DIFF
--- a/alacritty/build.rs
+++ b/alacritty/build.rs
@@ -15,6 +15,17 @@ fn main() {
         .write_bindings(GlobalGenerator, &mut file)
         .unwrap();
 
+    let target_os = env::var("CARGO_CFG_TARGET_OS");
+    match target_os.as_ref().map(|x| &**x) {
+        Ok("linux") | Ok("freebsd") => {},
+        Ok("openbsd") => {
+            println!(r"cargo:rustc-link-search=/usr/local/lib")
+        },
+        Ok("windows") => {},
+        _ => {}
+    }
+
+
     #[cfg(windows)]
     embed_resource::compile("./windows/windows.rc");
 }


chrisduerr (changes requested):
Adding platform-specific linker overrides to the build script of Alacritty is **not** something I'm interested in.

If people use obscure platforms that can't even manage to provide linkable libraries, then that should be worked around outside of Alacritty's upstream.

PR #5443, opened 2021-08-23
Fix background_opacity deprecation

chrisduerr:
During the deprecation of the background_opacity field, it was
incorrectly renamed to window_opacity. This changes that back to ensure
the old field still works and a warning is emitted accordingly.

See the original regression here:
https://github.com/alacritty/alacritty/commit/c24d7dfd0d2d8849f0398d7cb1a65d6562ee7a0d#diff-f92f0b1ad70a7b75b7266f3c9e569025e023d186814aa88c2b690800850ccb78L72-R73

### DIFF
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -70,7 +70,7 @@ pub struct UiConfig {
 
     /// Background opacity from 0.0 to 1.0.
     #[config(deprecated = "use window.opacity instead")]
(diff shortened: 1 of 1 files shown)


chrisduerr:
Waiting for a review of you here @kchibisov, since you submitted the original patch.

kchibisov (approved):
yeah, I've run sed on old config option as well...

PR #5476, opened 2021-09-06
Add more funky unicode chars to the default URL regex

unrelentingtech:
These are often used for [vim listchars](https://www.reddit.com/r/vim/comments/4hoa6e/what_do_you_use_for_your_listchars/) and similar, so any time a URL would be shown in an editor with e.g. a line break after it, that character would be caught in the regex…


chrisduerr:
These are perfectly valid, your specific usecase for URLs has no relevance for Alacritty's default URL detection.

PR #5480, opened 2021-09-07
Implement hyperlink terminal escapes 

### LINKED ISSUE #922: Support hyperlinks terminal escapes
See this gist: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

This is supported in iTerm2 and Gnome Terminal.

```
echo -e '\e]8;;http://example.com\aThis is a link\e]8;;\a'
```

displays `This is a link` with some formatting (in gnome term it's a dotted underline) and is clickable to open the given hyperlink.

Tmux currently has no support, so until that happens it's likely not going to be too useful in alacritty, but still a cool feature nonetheless.

oxalica:
Fix #922 

The definition of this escape sequence follows https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

**What are done in this PR:**
- Hyperlink escape handling
- ~~Customizable inactive and active styles.~~
  - ~~Default: Underline for inactive links, DoubleUnderline for active (hovered) links.~~
- Mouse hover icon and click to open, with customizable launcher commands (just like `hints.enabled.command`).
- Optional modifier keys required for activate and click action (just like `hints.enabled.mouse.mods`)
  - Use the same config as regex hint.
- Preview for link target URI at the bottom, with customizable fg/bg color.

**Test commands:**
```bash
# Simple
printf '\e]8;;http://example.com\aThis is a link\e]8;;\a\n'
# Complex (with hyperlink id)
printf '\e]8;id=ddg;https://ddg.gg\e\\[Go du]\e]8;;\e\\\e]8;;https://example.com\e\\Go example\e]8;;\e\\\e]8;;https://github.com\e\\Go github\e]8;;\e\\\n\e]8;id=ddg;https://ddg.gg\e\\[cking]\e]8;;\e\\\n'
```

**Screenshot: (updated)**

https://user-images.githubusercontent.com/14816024/140744154-42e03a87-e469-4d94-a7f8-c661003854d3.mp4


oxalica:
vtebench shows no significant difference.

output

chrisduerr:
Thanks for the PR, I'll take a closer look in a bit but there are a couple thoughts I had right off the bat:

First of all for such a simple feature this is a big patch. The amount of code useful solely for this single feature that doesn't have any other value is pretty big (pretty much all of this PR). I personally don't see myself ever making much use of this feature so it's questionable how beneficial this is.

A separate configuration to hints is also somewhat annoying. We just got the hints and URLs unified to have a more straight-forward and easy to understand configuration. Now there's a separate configuration again and people are without a doubt going to question why their URL highlighting doesn't work as expected when changing just one of these rather than both.

I'm also not sure why they should be styled when they're not active. Alacritty has never styled inactive (non-hovered) URLs and I don't see how this feature is any different. Why add a style to a cell that hasn't been specified for URLs, but not for hints? Then use a separate style for hovered URLs that isn't used for hints. That just seems inconsistent.

It's good that you have a preview for the highlighted links included off the bat, as I've mentioned before there's no way this feature would be usable without that. But the `render_string` facility isn't exactly ideal for doing so. I'd expect URIs to be cut-off at the right side of the terminal when they're too long? If this was just a visual thing I wouldn't care super much about it, usually with massive URLs people don't need to see all of that garbled mess. But since this is an essential part of the security of this feature I'm not sure if that suffices. One could certainly argue that once people are aware that there's more to a URL, they should just expand their terminal until they can see all of it or not click it, but even to do just that they'd need to **know** that there is more and the URL doesn't just end right at the screen border. So at the **minimum** we need to add some indicator that the URL cannot be displayed completely.

The code itself seems better than I initially would have expected. It makes use of the cell's heap storage to avoid filling up ram and there's no unnecessary branching when writing cells without hyperlinks. The biggest problem is probably the `Hyperlink` struct which is an absolute mess. I'd be curious why you went through the trouble of using Arc<str>s that combine IDs and URIs, then throw in an isize which has different meanings for positive and negative. That just seems like extremely unmaintainable code to me, so I'd assume that there were significant performance reasons for doing so. Did you run into performance issues with any tests before going for this approach?

oxalica:
> First of all for such a simple feature this is a big patch. 

Actually yes. This PR is implemented as Proof of Concept for the long-living issue. I just implement features what I want to have straightforward and try to make them *fast first*. I know that you are unhappy with PRs slowing down performance, so I think we could first make it *possible* then make it clean.

If you feel happy with the benchmark result, I can clean up codes later. Some code branches are merge-able (as you say, with `hints`) and can be simplified.

> A separate configuration to hints is also somewhat annoying.

Yes. Hover effect can be moved into `hints`, but inactive hyperlink style still doesn't fit in there. (see below for necessarily)

> Why add a style to a cell that hasn't been specified for URLs, but not for hints?

Inactive link styles are **necessary**. Hyperlink is an attribute specified by inner program, binding to some text, like color. And user cannot even know what is clickable without it. No one will scanning cursor to "discover" all links printed by like clang errors. Typically, this should be a special or unique style like dotted underline in `gnome-termimal` to easily distinguish from other styled text.

On the other hand, hints are related to your cursor, it's an interpretation (defined by you!) of some text maybe without special meaning for inner program. Hints should not be styled unless hovered.

> Then use a separate style for hovered URLs that isn't used for hints. That just seems inconsistent.

I agree. We should use the same style for hints and hyperlinks when hovered.
Originally I want to set inactive links to be dotted underline, and underline for hovering (same as `hints`). But unfortunately dotted underline is not implemented by us. I don't know what can be a better default style for inactive links. So I just randomly choose underline for inactive and double underline for active to make them different.

> But the render_string facility isn't exactly ideal for doing so. I'd expect URIs to be cut-off at the right side of the terminal when they're too long?

Currently it just cut-off at the right side. But we can render ellipsis at the end (`foobar…`) or in the middle (`foo…bar`).
I don't think multi-line is necessary and I prefer ellipsis at the end. Because for URL/URI, the most important part is in the front, like schema and domain name. If user can confirm that schema and domain name is secure, they don't want to see the whole, maybe hundreds characters long query parameters or hashes (SPA usually have very long hashes and with tokens. But you can simply check domain name to avoid sending tokens to wrong server).

> The biggest problem is probably the Hyperlink struct which is an absolute mess.

Yes. It's a premature optimization for both memory and performance when printing and handling hyperlinks. But I want to confirm if there is any performance issue now first. If it's fine or unrelated, then we can swap in an simple impl.
But `Arc` should be necessary, or we'll copy the whole link string for every characters printed after the starting escape.

chrisduerr:
> But unfortunately dotted underline is not implemented by us.

Just to clarify, this should probably be something that is trivial to support in a separate PR. So it might be an option in the future. At least it's much cleaner to implement than something like curved underlines where you have to bother with aliasing.

> Currently it just cut-off at the right side. But we can render ellipsis at the end (foobar…) or in the middle (foo…bar).
I don't think multi-line is necessary and I prefer ellipsis at the end.

I pretty much agree with this. Should also be trivial to solve.

> Yes. It's a premature optimization for both memory and performance when printing and handling hyperlinks. But I want to confirm if there is any performance issue now first.

In that case part of this PR should probably be to come up with some benchmarks that stress the hyperlink codepaths, not just our existing benchmarks. Vtebench should make this easy, anything mouse related might require more manual testing but even that can be done during vtebench runs. This can then both be compared to versions where hyperlinks are no-op and alternative potentially slower implementations. But all optimizations that aren't completely obvious or impact maintainability should be founded on some form of benchmarks.

> But Arc should be necessary, or we'll copy the whole link string for every characters printed after the starting escape.

Arc is one option for this, yes. One could also have the text stored separately but at that point one would have to do the garbage collection manually and basically create a custom RC for determining when a URI text can be dropped.

oxalica:
I refactored most things and now hyperlinks are considered as hints souce rather than `regex`. So they can share most of code for like rendering and action.

Inactive link styles are currently removed since it can be implemented in a separated PR if required later. Especially after we have dotted underline or other special styles implemented.

There are still some work not done including:
- URL preview bar rendering truncation.

  I just notice that `render_string` doesn't support unicode wide character. They overlap with each other and cause a mess.
  I'm not sure if we have other existing way to correctly handle them. Or we need to calculate character widths and adjust the position in `render_string`, which seems reinventing the wheel to me.

- Navigate hints with keyboard (Control-Shift-U) for both regex and hyperlink hints.

  The current implementation seems tightly coupled with regex matching. It still need a lot of work.

chrisduerr:
> I just notice that render_string doesn't support unicode wide character. They overlap with each other and cause a mess.
I'm not sure if we have other existing way to correctly handle them. Or we need to calculate character widths and adjust the position in render_string, which seems reinventing the wheel to me.

In other places we actually go through the string and insert whitespace, you should be able to see that in the message bar rendering code. That's currently the only place we need it so it's probably rather ugly, so I'd suggest doing it "properly" in the `render_string` or a separate helper method but following the same principle. Don't think there's any need to do complex compositing of text strings, just adding whitespace should work without any issues.

> Navigate hints with keyboard (Control-Shift-U) for both regex and hyperlink hints.

I think I'd want this to be part of the initial PR, I think otherwise people would get really confused why they can open some links with the keyboard but not others. I suppose that's also a great argument for combining hints and the url escape because that should make it sensible to provide keyboard URL selection.

oxalica:
> so I'd suggest doing it "properly" in the render_string or a separate helper method but following the same principle

Currently every `render_string` caller has slightly different layout. Message bar have multiple lines with a `[x]` on the right side, render timer just simply prints numbers, search bar has a "scrolling" behavior with the `Search: ` text always on the front, and URL preview has truncation at last. It's hard to unify them to a single function which each has kind of pre-processing besides unicode width handling. Maybe they need a further refactor but I just don't touch `render_string` and implement truncation specially for URL preview now.

> Navigate hints with keyboard (Control-Shift-U) for both regex and hyperlink hints.

It's now implemented. Hyperlinks with the same `id` are considered the same and only the first part is navigate-able.

alacritty-search-hints

chrisduerr:
> Currently every render_string caller has slightly different layout. Message bar have multiple lines with a [x] on the right side, render timer just simply prints numbers, search bar has a "scrolling" behavior with the Search: text always on the front, and URL preview has truncation at last. It's hard to unify them to a single function which each has kind of pre-processing besides unicode width handling.

Beautiful, isn't it? Every thing working slightly differently.

Just having a method with a `String -> String` signature that inserts spaces so that `chars().count()` is the same as its width might be sufficient and generic enough to be used in multiple places?

>  Hyperlinks with the same id are considered the same and only the first part is navigate-able.

That sounds like it's pretty much the only option, right? Or why do you think this is important to point out explicitly?

I'd hope that if only part of the fragmented URI is visible on screen, the hints are still displayed correctly only on the first visible part of the hyperlink?

oxalica:
> Just having a method with a String -> String signature that inserts spaces so that chars().count() is the same as its width might be sufficient and generic enough to be used in multiple places?

No. If we don't do truncation (or wrapping, for message bar) at the same time of space insertion, it would cut wide characters. Extracting the space insertion seems to make other things more complicated. So I don't want to touch existing code now.

> I'd hope that if only part of the fragmented URI is visible on screen, the hints are still displayed correctly only on the first visible part of the hyperlink?

Yes. The first visible part shows the keyboard hint as expected. 
It is done be scanning the visible grid and dedup using `HashSet` [here](https://github.com/alacritty/alacritty/pull/5480/files#diff-c594944518dcab50be2c9f6a148f324d5a4f663cd3f9be99593c1f50872ea61bR110).
Since we use `Arc` to get rid of garbage collection hyperlinks, there is no easy way other than traversing the (visible) grid to enumerate all links. The performance should be acceptable though.

alacritty.yml
```diff
@@ -234,6 +234,13 @@
     #  background: '#c5c8c6'
     #  foreground: '#1d1f21'
 
+  # Hyperlink URL preview bar
+  #
+  # Default to be the inverse of normal foreground/background.
+  #hyperlink_preview_bar:
+  #  background: '#c5c8c6'
+  #  foreground: '#1d1f21'
+
```
chrisduerr, line 243: Realistically is there any reason why the search and hyprlink preview bar should have different colors? I think unification might make sense.

alacritty.yml
```diff
@@ -452,18 +459,32 @@
...
+  #   Note that target URL of a hyperlink may differ from its text displayed.
+  #
+  #   `hyperlink` takes precedence over `regex` when they both match on the
+  #   same text.
+  #
+  # - Either an `action` or a `command` field, specifying the action when
+  #   triggered.
+  #
+  # - Zero or more optional fields, including `mouse`, `binding` and
+  #   `post_processing`.
```
chrisduerr, line 487: This is too verbose, with this much text nobody will read through the hints documentation anymore. The existing documentation was already **a lot** and this almost doubles its size.

I also find it questionable if regex and hint should be combined in the same hint. It seems somewhat strange and clunky.

And if we decide to support multiple things in the same hint, should we also allow multiple regexes?

alacritty/src/display/content.rs
```diff
@@ -257,6 +259,7 @@ impl RenderableCell {
 
         RenderableCell {
             zerowidth: cell.zerowidth().map(|zerowidth| zerowidth.to_vec()),
+            hyperlink: cell.hyperlink().cloned(),
```
chrisduerr, line 262: Haven't looked at the full PR yet, but do we really need to clone this for every single cell part of the hyperlink? If it's only about rendering wouldn't it suffice to make this into a bool?

alacritty/src/display/content.rs
```diff
@@ -407,31 +410,31 @@ impl<'a> Hint<'a> {
...
             return None;
         }
 
         // Match starting position on this line; linebreaks interrupt the hint labels.
         let start = self
-            .regex
             .matches
-            .get(self.regex.index)
-            .map(|regex_match| max(*regex_match.start(), viewport_start))
+            .matches
```
chrisduerr, line 420: This seems like pretty bad naming. `matches.matches` is just strange.

alacritty/src/display/content.rs
```diff
@@ -407,31 +410,31 @@ impl<'a> Hint<'a> {
...
+            .get(self.matches.index)
+            .map(|matched| max(*matched.bounds().start(), viewport_start))
```
chrisduerr, line 422:
```suggestion
            .map(|match| max(*match.bounds().start(), viewport_start))
```

"Matched" as a noun here seems like a strange choice.

alacritty/src/display/content.rs
```diff
@@ -478,31 +481,33 @@ impl DerefMut for RegexMatches {
...
-impl<'a> Regex<'a> {
-    /// Create a new renderable regex iterator.
+impl<'a> TrackedHintMatches<'a> {
+    /// Create a new renderable iterator.
     fn new<T>(term: &Term<T>, dfas: &RegexSearch) -> Self {
-        let matches = Cow::Owned(RegexMatches::new(term, dfas));
-        Self { index: 0, matches }
+        let matches = RegexMatches::new(term, dfas);
+        let matches =
+            matches.0.into_iter().map(|bounds| HintMatchLocation::Regex { bounds }).collect();
```
chrisduerr, line 499: This seems rather ugly, mapping everything and recollecting just to convert it to the right wrapper type.

alacritty/src/display/content.rs
```diff
@@ -478,31 +481,33 @@ impl DerefMut for RegexMatches {
...
 
-    /// Advance the regex tracker to the next point.
+    /// Advance the tracker to the next point.
     ///
-    /// This will return `true` if the point passed is part of a regex match.
+    /// This will return `true` if the point passed is part of a hint match.
     fn advance(&mut self, point: Point) -> bool {
-        while let Some(regex_match) = self.matches.get(self.index) {
-            if regex_match.start() > &point {
+        while let Some(matched) = self.matches.get(self.index) {
```
chrisduerr, line 507:
```suggestion
        while let Some(match) = self.matches.get(self.index) {
```

Same thing again, `matched` as a noun doesn't seem like a great variable name when just "match" is available.

alacritty/src/display/hint.rs
```diff
@@ -33,6 +35,27 @@ pub struct HintState {
     keys: Vec<char>,
 }
 
+#[derive(Debug, Clone)]
+pub enum HintMatchLocation {
+    Regex { bounds: Match },
```
chrisduerr, line 40: Usually for a single parameter just `Regex(Match)` would be easier to use.

alacritty/src/display/hint.rs
```diff
@@ -33,6 +35,27 @@ pub struct HintState {
...
+            Self::Regex { bounds } | Self::Hyperlink { bounds, .. } => bounds.clone(),
+        }
+    }
+
+    fn hyperlink(&self) -> Option<&Hyperlink> {
+        match self {
+            Self::Hyperlink { hyperlink, .. } => Some(hyperlink),
+            _ => None,
+        }
+    }
```
chrisduerr, line 56: This method seems a little strange. Not the end of the world but I wouldn't expect a method exclusive to a single of the enum variants on the entire enum that just returns None otherwise. Seems like at that point you'd rather just do the logic externally?

alacritty/src/display/hint.rs
```diff
@@ -70,21 +93,83 @@ impl HintState {
...
-            // Apply post-processing and search for sub-matches if necessary.
-            if hint.post_processing {
-                matches
-                    .drain(..)
-                    .map(|rm| HintPostProcessor::new(term, regex, rm).collect::<Vec<_>>())
-                    .flatten()
-                    .collect()
-            } else {
-                matches.0
+        let mut hint_matches: Vec<HintMatchLocation> = Vec::new();
```
chrisduerr, line 96: This method is an absolute massive mess and I'm going to hold off on a review until it's cleaned up.

alacritty/src/display/hint.rs
```diff
@@ -259,32 +371,51 @@ pub fn highlighted_at<T>(
...
+        }
 
-            // Apply post-processing and search for sub-matches if necessary.
-            let regex_match = if hint.post_processing {
-                HintPostProcessor::new(term, regex, regex_match).find(at_point)
-            } else {
-                Some(regex_match)
-            };
+        if let Some(regex) = &hint.regex {
+            if let Some(matched) =
```
chrisduerr, line 381: Should never have two nested conditionals like that, it just unnecessarily increases indentation. Since the outside option isn't needed in the final scope, we can just and_then things.

alacritty/src/display/mod.rs
```diff
@@ -518,13 +518,19 @@ impl Display {
...
                     let point = viewport_to_point(display_offset, cell.point);
-                    if highlighted_hint.as_ref().map_or(false, |h| h.bounds.contains(&point))
-                        || vi_highlighted_hint.as_ref().map_or(false, |h| h.bounds.contains(&point))
+                    let hyperlink = cell.hyperlink.as_ref();
+                    if (highlighted_hint
+                        .as_ref()
+                        .map_or(false, |h| h.should_highlight(point, hyperlink)))
+                        || vi_highlighted_hint
+                            .as_ref()
+                            .map_or(false, |h| h.should_highlight(point, hyperlink))
```
chrisduerr, line 533: This is somewhat unfortunately messy. Though not the highest priority.

alacritty/src/display/mod.rs
```diff
@@ -771,6 +778,46 @@ impl Display {
         });
     }
 
+    /// Draw the hyperlink target URI preview of which are currently active.
+    fn draw_active_hyperlink_preview(&mut self, config: &Config, size_info: &SizeInfo) {
+        // Is there really anyone hovering two links simontinously?
```
chrisduerr, line 783: Well is it possible? Each cell should only have one hyperlink, so we shouldn't have to consider this.

alacritty/src/event.rs
```diff
@@ -635,38 +635,46 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
             HintAction::Action(HintInternalAction::Copy) => {
-                let text = self.terminal.bounds_to_string(*hint.bounds.start(), *hint.bounds.end());
-                self.clipboard.store(ClipboardType::Clipboard, text);
+                self.clipboard.store(ClipboardType::Clipboard, target);
             },
             // Write the text to the PTY/search.
             HintAction::Action(HintInternalAction::Paste) => {
-                let text = self.terminal.bounds_to_string(*hint.bounds.start(), *hint.bounds.end());
-                self.paste(&text);
+                self.paste(&target);
```
chrisduerr, line 658: If possible we should remove the `=> { .. }` and inline stuff just to make things shorter. I think that would be possible here.

alacritty/src/event.rs
```diff
@@ -635,38 +635,46 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
             // Select the text.
             HintAction::Action(HintInternalAction::Select) => {
-                self.start_selection(SelectionType::Simple, *hint.bounds.start(), Side::Left);
-                self.update_selection(*hint.bounds.end(), Side::Right);
-                self.copy_selection(ClipboardType::Selection);
+                if let Some(bounds) = bounds {
+                    self.start_selection(SelectionType::Simple, *bounds.start(), Side::Left);
+                    self.update_selection(*bounds.end(), Side::Right);
+                    self.copy_selection(ClipboardType::Selection);
+                }
```
chrisduerr, line 666: I think it's unfortunate that hyperlinks can't be selected. I'd assume the primary reason for this is that a hyperlink might not be continuous?

alacritty_terminal/src/ansi.rs
```diff
@@ -983,6 +989,30 @@ where
                 unhandled(params);
             },
 
+            // Hyperlink.
+            b"8" => {
+                if params.len() >= 3 {
```
chrisduerr, line 994: This seems like a good candidate for an early return rather than indenting everything?

alacritty_terminal/src/ansi.rs
```diff
@@ -983,6 +989,30 @@ where
...
+                        return;
+                    }
+
+                    // Link parameters are in format of `key1=value1:key2=value2`.
+                    // Currently only key `id` is defined.
+                    let id = link_params
+                        .split(|&b| b == b':')
+                        .find(|kv| kv.starts_with(b"id="))
+                        .and_then(|kv| str::from_utf8(&kv[3..]).ok());
+                    // Is empty URI a sane default?
```
chrisduerr, line 1008: Not starting a hyperlink if we don't understand the URI is probably a saner default.

alacritty_terminal/src/term/cell.rs
```diff
@@ -104,8 +106,34 @@ impl Cell {
     #[inline(never)]
     pub fn clear_wide(&mut self) {
         self.flags.remove(Flags::WIDE_CHAR);
-        self.drop_extra();
         self.c = ' ';
+        if self.extra.as_ref().map_or(false, |extra| extra.hyperlink.is_none()) {
+            self.drop_extra();
+        }
```
chrisduerr, line 112: The zerowidth character should still be cleared. Pretty sure this will lead to incorrect behavior with zerowidth characters present.

chrisduerr (changes requested):
Just a super rushed rough first review of the code so you're aware of some of the issues.

Please let me know if you have any questions.

alacritty.yml
oxalica: Simply because they are not the same thing (input bar v.s. output bar) and search bar color is under `search` field.
If we want to unify them into another field, I'm not sure what could be the suitable name. Just `bar_color` is a bad name.

alacritty/src/display/content.rs
oxalica: There's a keyword `match`. I use `matched` everywhere just to workaround it.

alacritty/src/event.rs
oxalica: Yes. But we could save their location when scanning hyperlinks in `HintState::update`, which makes the first part selectable. It should be fine for most cases, yes?

alacritty/src/display/mod.rs
oxalica: Actually yes. You can go vi mode and put your cursor over one, and move your mouse to point another one.

alacritty/src/display/content.rs
oxalica: I also don't want this field be a part of `RenderableCell` since it doesn't affect rendering.
It's here because in `Display::draw` we convert the grid to `RenderableCell`, then drop the grid early to free the lock, before traversing cells here: https://github.com/alacritty/alacritty/blob/770128c16d1baea1d8f1b528937f38226ca58c02/alacritty/src/display/mod.rs#L530

If it's not in `RenderableCell`, we should move the highlight checking logic to here, when holding the lock:
https://github.com/alacritty/alacritty/blob/770128c16d1baea1d8f1b528937f38226ca58c02/alacritty/src/display/mod.rs#L491

alacritty.yml
oxalica: > I also find it questionable if regex and hint should be combined in the same hint. 

If they are not, we need to use different key bindings for them, instead of Control-Shift-U to show both regex matched and hyperlink escaped text. I don't think it provide a good user experience.

alacritty/src/event.rs
chrisduerr: Not sure if we should offer it if it doesn't reliably make sense. Maybe selecting from the start of the first part to the end of the last part makes sense? But that could involve stuff like tmux panes and cause strange issues.

alacritty/src/display/mod.rs
chrisduerr: That's definitely possible and should already work with normal hints. The both should be treated independently without any issues though. "Hovering" as in hovering above with the mouse should only be possible with one.

alacritty/src/display/content.rs
chrisduerr: Ah right, that's why I was using `regex_match`, really don't like `matched` though it's not a great name.

alacritty/src/display/content.rs
chrisduerr: That only needs the bounds though, right? We don't need to clone the arc?

alacritty.yml
chrisduerr: The name is a different question, but I definitely think combination and deprecation of the old search bar colors makes sense.

alacritty/src/display/content.rs
oxalica: > That only needs the bounds though

Yes but we cannot get bounds from `Hyperlink` without traversing the grid.
Traversing grid and comparing them with highlighted one when holding the term lock seems quite wrong.

Another idea is to traverse the grid in `highlighted_at`, so it only occurs when mouse hovering a hyperlink. And save *ALL* bounds instead of the hovered one to `HintMatch`. Then we'll have bounds when rendering.

alacritty/src/display/hint.rs
oxalica: The traversing part is separated to `traverse_hyperlinks` now.

alacritty/src/event.rs
oxalica: It's now supported.

oxalica:
I refactor and simplify some code. `HintMatchLocation` is removed now. 

Also that bounds calculation of hyperlinks are now handled for both mouse (`highlighted_at`) and keyboard navigation (`HintState`). When set to hint action to `Select`, mouse will select the contiguous part under the cursor point (all parts can be selected individually), and Control-Shift-U will only print hint labels on the first part (can only select the first part).

alacritty/src/display/content.rs
chrisduerr: I'm not sure what you mean. This is just stored here to render the underline, right? Why would we need to store the URL's **text** to render that? I don't see any reason why we'd need that information.

Storing IDs to compare to the hovered one to distinguish between hovered or not is one thing, but that doesn't require any text storage.

oxalica: > Storing IDs to compare to the hovered one

I got it. But id can also be `Arc<str>` in some cases. You still cannot get rid of string or Arc cloning. I don't think replacing `Hyperlink` with `Id` here really changes anything.

chrisduerr: Well dragging around an Arc is definitely different from dragging around an integer. It just seems a bit wasteful to clone everything for every single cell, especially if there's a lot of URIs present.

oxalica:
A question: how to deal with links with same id but different target URL? Currently we compare both id and URL for equivalence. If we want to compare only id, we'll underline all parts with same id when hovering one, but clicking different part may goto different target URL. Seems like another security issue?

chrisduerr:
> A question: how to deal with links with same id but different target URL? 

It's been a long time since I've looked at the proposal, so it would depend on how that's suggested to be handled there. Is same ID but different URI a thing?

oxalica:
Okay it's mentioned [here](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#hover-underlining-and-the-id-parameter)
> Character cells that have the same target URI and the same nonempty id are always underlined together on mouseover.

So we cannot simply rely on only `id` string. Thus we need another struct like `Either<usize, (URI, id)>` to actually represent "id" as you said for `RenderableCell`. It's the same size as `Hyperlink` and can only reduce one `Arc::clone` call when creating `RenderableCell`. Nothing more is affected since we never clone `RednerableCell`.
Since `Arc::clone` is already quite cheap, does adding a new struct and more conversion code to reduce a single `Arc::clone` really worth it?

alacritty.yml
oxalica: The docs are simplified.
But I keep hyperlink and regex NOT exclusive since I prefer navigate both with one key shortcut. Also users can still choose to use different ones if they want.

> should we also allow multiple regexes

No. Regex can already be naturally joined by `|`.

alacritty.yml
oxalica: Merged into `information_bar`

chrisduerr:
> Character cells that have the same target URI and the same nonempty id are always underlined together on mouseover.

That's quite interesting. How does that work with multiple of the same URIs visible? Feels a bit weird. I suppose I'd prefer just a forced ID for every URI but I guess not much we can do.

oxalica:
> How does that work with multiple of the same URIs visible? Feels a bit weird. 

You can see my test example in the issue description. The left two part `[Go du]` `[cking]` are marked as the same id so they are highlighted together. Another two part `Go example` `Go github` have no id so they get new unique numeric id every time.

If two links have same string id provided by program but different URI. I currently make them different, as the spec said, by making `Hyperlink` comparing both string id and URI for equivalence. You cannot trust program provided id.

In brief, the current implementation follows the spec and works intuitively.
I just want to argue that `Hyperlink` in `RenderableCell` should be fine. You cannot simply use `usize` there.

Also this PR is currently ready for review.

chrisduerr:
@perfbot

alacritty/src/daemon.rs
```diff
@@ -15,10 +15,10 @@ use winapi::um::winbase::{CREATE_NEW_PROCESS_GROUP, CREATE_NO_WINDOW};
 /// Start the daemon and log error on failure.
 pub fn start_daemon<I, S>(program: &str, args: I)
 where
-    I: IntoIterator<Item = S> + Debug + Copy,
+    I: IntoIterator<Item = S> + Debug + Clone,
     S: AsRef<OsStr>,
 {
-    match spawn_daemon(program, args) {
+    match spawn_daemon(program, args.clone()) {
```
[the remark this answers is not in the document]
oxalica: The args is used below for logging. I want to pass iterator `args.chain(iter::once(target))` to avoid `Vec`, but unfortunately chained iterator is not `Copy`. 
To reduce complexity, I revert this part back and keep the code of collecting `Vec`.

alacritty/src/display/content.rs
```diff
@@ -478,31 +480,31 @@ impl DerefMut for RegexMatches {
...
 
-impl<'a> Regex<'a> {
-    /// Create a new renderable regex iterator.
+impl<'a> TrackedHintMatches<'a> {
+    /// Create a new renderable iterator.
     fn new<T>(term: &Term<T>, dfas: &RegexSearch) -> Self {
-        let matches = Cow::Owned(RegexMatches::new(term, dfas));
-        Self { index: 0, matches }
+        let matches = RegexMatches::new(term, dfas);
+        Self { index: 0, matches: Cow::Owned(matches.0) }
```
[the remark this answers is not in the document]
oxalica: Functionality of `RegexMatches` is moved to `visible_regex_match_iter` now.

alacritty/src/display/hint.rs
```diff
@@ -70,21 +73,57 @@ impl HintState {
...
-                    .collect()
-            } else {
-                matches.0
-            }
-        });
+        let mut hint_matches: Vec<Match> = Vec::new();
+
+        // Find hyperlinks.
+        if hint.hyperlink {
+            // FIXME: These code dups with `RegexMatches::new`.
```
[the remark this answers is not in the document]
oxalica: Fixed

alacritty/src/display/hint.rs
```diff
@@ -70,21 +73,57 @@ impl HintState {
...
+                    );
+                } else {
+                    hint_matches.extend(matches.0);
+                }
+            });
+        }
+
+        // Sort and dedup ranges. Currently overlapped but not exactly same ranges are kept.
+        hint_matches.sort_by_key(|bounds| (*bounds.start(), Reverse(*bounds.end())));
+        hint_matches.dedup_by_key(|bounds| *bounds.start());
```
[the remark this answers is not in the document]
oxalica: I just use the most simple way -- sort and dedup. I'm not sure what's the better approach here when part of link text is URI, or part of URI-like text is a link. I think this mostly happen when a hyperlink is wrapped in tmux or something. In this case, the first part will be both regex-match and hyperlink-match, and we should prefer the latter, since regex-match get the wrong result. 
Complex overlapping case is very rare to happen in real world.

alacritty/src/display/hint.rs
```diff
@@ -239,6 +311,52 @@ impl HintLabels {
     }
 }
 
+/// Traverse a range of grid for all hyperlinks, with adjacent ones joined into a range.
+fn traverse_hyperlinks<T>(
+    term: &Term<T>,
+    start: Point,
+    end: Point,
+    mut callback: impl FnMut(Hyperlink, Match),
```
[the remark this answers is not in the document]
oxalica: It's now refactored to `visible_hyperlink_iter`, a function returning Iterator.

alacritty_terminal/src/index.rs
```diff
@@ -56,31 +56,51 @@ impl<L, C> Point<L, C> {
...
+        D: Dimensions,
+    {
+        self.sub_unbounded(dimensions, rhs).grid_clamp(dimensions, boundary)
+    }
+
+    /// Add a number of columns from a point without clamp on boundary.
     #[inline]
     #[must_use = "this returns the result of the operation, without modifying the original"]
-    pub fn add<D>(mut self, dimensions: &D, boundary: Boundary, rhs: usize) -> Self
+    pub fn add_unbounded<D>(mut self, dimensions: &D, rhs: usize) -> Self
```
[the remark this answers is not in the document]
oxalica: I refactored the grid traversal code and now these are removed.

oxalica:
Finally pick it up again.
I refactor most of the grid traversal and regex matching part to make them more simple.
The messy commit history is also cleaned up. Hope it's easier to review.

chrisduerr:
@oxalica If there's nothing else you want to do, please re-request a review from me.

chrisduerr:
I'm sorry that I haven't gotten around to another review of this PR, I'd just like to let you know that it is still on my radar.

alacritty_terminal/src/grid/mod.rs
```diff
@@ -397,11 +397,18 @@ impl<T> Grid<T> {
         self.raw.truncate();
     }
 
-    /// Iterate over all cells in the grid starting at a specific point.
+    /// Iterate over all cells in the grid from a starting point (exclusive) to an ending point
+    /// (inclusive).
+    #[inline]
+    pub fn iter_from_to(&self, start: Point, end: Point) -> GridIterator<'_, T> {
+        GridIterator { grid: self, point: start, end }
+    }
```
chrisduerr, line 405: Why move this to a new method? It's not used anywhere.

alacritty/src/event.rs
```diff
@@ -659,28 +659,31 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             return;
         }
 
-        match &hint.action {
+        let bounds = hint.bounds();
+        let target = match hint {
```
chrisduerr, line 663: Target as a variable name here is kinda poor.

alacritty/src/config/color.rs
```diff
@@ -18,15 +18,26 @@ pub struct Colors {
...
+            .bar
+            .foreground
+            .or(self.information_bar.foreground)
+            .unwrap_or(self.primary.background)
     }
 
-    pub fn search_bar_background(&self) -> Rgb {
-        self.search.bar.background.unwrap_or(self.primary.foreground)
+    pub fn information_bar_background(&self) -> Rgb {
+        // Legacy field first.
```
chrisduerr, line 35: The comment isn't really necessary, that should be a familiar thing within the codebase so no need to point it out every time.

alacritty/src/config/color.rs
```diff
@@ -18,15 +18,26 @@ pub struct Colors {
...
     pub hints: HintColors,
     pub transparent_background_colors: bool,
+    pub information_bar: BarColors,
 }
 
 impl Colors {
-    pub fn search_bar_foreground(&self) -> Rgb {
-        self.search.bar.foreground.unwrap_or(self.primary.background)
+    pub fn information_bar_foreground(&self) -> Rgb {
+        // Legacy field first.
```
chrisduerr, line 26: The comment isn't really necessary, that should be a familiar thing within the codebase so no need to point it out every time.

alacritty.yml
```diff
@@ -455,18 +461,23 @@
...
   #alphabet: "jfkdls;ahgurieowpq"
 
   # List with all available hints
   #
-  # Each hint must have a `regex` and either an `action` or a `command` field.
-  # The fields `mouse`, `binding` and `post_processing` are optional.
+  # Each hint must have:
+  # 1. At least one of a `regex` or a `hyperlinks` field. It marks what content
+  #    this hint setting appies to. The former takes the matching regex,
+  #    and the latter takes `true` to match program marked hyperlinks.
```
chrisduerr, line 477:
```suggestion
  # 1. At least one of a `regex` or a `hyperlinks` field. It marks what content
  #    this hint setting applies to. Setting `hyperlinks` to `true` will match the hyperlink escape sequence.
```

Formatting isn't right, but I think this should be better. No need to point out that regex is a regex, the name kinda makes that obvious. Pointing out that this is about the content isn't particularly insightful either I think. So just keeping it short here.

alacritty.yml
```diff
@@ -262,6 +258,16 @@
   #  foreground: None
   #  background: None
 
+  # Information bar
```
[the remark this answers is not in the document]
chrisduerr: Still not a fan of `information bar`. Especially because for things like search it's used as an input not as an output for information.

Maybe just use the location of this "bar" as a name? So something like `footer bar`? The concepts of footers should be familiar to people already so they know where to look for it. Could also go with just `footer` but I think people would struggle with associating that directly to the bar at the bottom.

alacritty/src/config/ui_config.rs
```diff
@@ -324,8 +325,13 @@ pub enum HintAction {
...
 pub struct Hint {
-    /// Regex for finding matches.
-    pub regex: LazyRegex,
+    /// Apply this hint to regex matches.
+    #[serde(default)]
+    pub regex: Option<LazyRegex>,
+
+    /// Apply this hint to hyperlinks.
+    #[serde(default)]
+    pub hyperlinks: bool,
```
chrisduerr, line 334: By making both of these optional without any logic for checking that one of them is present, you're allowing users to specify hints with neither `regex` nor `hyperlinks`. I think that's probably the biggest functional issue with the PR in its current state.

If a config is invalid, we need to point that out. Especially because typos like `regxe` can be easily missed and users will start complaining that their stuff doesn't work. So you should ensure that at least one is present and throw an error otherwise.

alacritty/src/display/content.rs
```diff
@@ -408,102 +411,63 @@ impl<'a> Hint<'a> {
...
-}
-
-impl DerefMut for RegexMatches {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.0
-    }
-}
-
-/// Visible regex match tracking.
+/// Visible hint matches with current index, also used in VI mode search.
```
chrisduerr, line 441: The phrase "Visible hint matches with current index" implies that these are only the hint matches visible with some arbitrary index taken into account, but the `index` is unrelated to that and just an index into these matches. So the comment should make that more clear.

Also calling something `hint matches` when it is also used for search is super misleading. The "also used in VI mode search" comment is not sufficient here. The name shouldn't imply a certain feature when it's also used in a completely different context.

alacritty/src/display/content.rs
```diff
@@ -408,102 +411,63 @@ impl<'a> Hint<'a> {
...
     index: usize,
 }
 
-impl<'a> Regex<'a> {
-    /// Create a new renderable regex iterator.
-    fn new<T>(term: &Term<T>, dfas: &RegexSearch) -> Self {
-        let matches = Cow::Owned(RegexMatches::new(term, dfas));
-        Self { index: 0, matches }
+impl<'a> HintMatches<'a> {
+    /// Create from given matches with index 0.
```
chrisduerr, line 452: The consumer should likely be completely unaware of the entire `index` logic. They just want to have some matches they can "iterate" over. No need to expose internals like that, even in comments.

alacritty/src/display/content.rs
```diff
@@ -408,102 +411,63 @@ impl<'a> Hint<'a> {
...
+    fn new(matches: impl Into<Cow<'a, [Match]>>) -> Self {
+        Self { matches: matches.into(), index: 0 }
+    }
+
+    /// Create from regex matches on term visable part, with index 0.
```
chrisduerr, line 457: Same comment about the index applies here too.

Also the sentence `Create from regex matches on term visable part` sounds somewhat broken, I'm not sure what you're trying to say here exactly.

alacritty/src/display/mod.rs
```diff
@@ -778,6 +779,47 @@ impl Display {
         });
     }
 
+    /// Draw the hyperlink target URI preview of which are currently active.
```
chrisduerr, line 782:
```suggestion
    /// Draw the focused hyperlink escape target URI preview.
```

alacritty/src/display/mod.rs
```diff
@@ -778,6 +779,47 @@ impl Display {
...
+    fn draw_active_hyperlink_preview(&mut self, config: &Config, size_info: &SizeInfo) {
+        // Currently we simply stack vi-highlighted and cursor-pointed URL if both exist.
```
chrisduerr, line 784: This phrasing implies we want to change that in the future, why didn't you change that immediately?

Fundamentally I don't think it's too big of a deal to draw overlap if the two URIs should overlap, but chances are we might be able to filter for full equivalency and then just save ourselves the hassle? There's probably not a lot of scenarios where they overlap but aren't exactly the same, right? Not sure if that's possible at all.

alacritty/src/display/mod.rs
```diff
@@ -778,6 +779,47 @@ impl Display {
...
+        {
+            let glyph_cache = &mut self.glyph_cache;
+
+            let point = Point::new(size_info.screen_lines().saturating_sub(1 + idx), Column(0));
+            let fg = config.ui_config.colors.information_bar_foreground();
+            let bg = config.ui_config.colors.information_bar_background();
+
+            let mut text = String::with_capacity(uri.len());
+            let mut text_len = 0usize;
+            // Leave room for the ellipsis character.
```
chrisduerr, line 803: Curious if we should maybe just define this ellipsis character as a constant and then use `unicode_width` or `len` or something to build the logic on dynamically. In theory if someone would want to change the character for some reason that would make it much easier then (to `...` for example).

We do the same thing with search I believe.

alacritty/src/display/mod.rs
```diff
@@ -778,6 +779,47 @@ impl Display {
...
+            let max_len = size_info.columns().saturating_sub(2);
+            for c in uri.chars() {
+                let width = c.width().unwrap_or(0);
+                if text_len + width > max_len {
+                    text.push_str("… ");
```
chrisduerr, line 808: Why is there a whitespace after the ellipsis? Should be in the last column, right? Is this because the ellipsis might be in the second to last column when dealing with fullwidth characters?

alacritty/src/display/hint.rs
```diff
@@ -70,21 +73,40 @@ impl HintState {
...
-                    .collect()
-            } else {
-                matches.0
+        let mut hint_matches: Vec<Match> = Vec::new();
+
+        // Find hyperlinks.
+        if hint.hyperlinks {
+            let mut visited = HashSet::new();
+            for (hyperlink, bounds) in visible_hyperlink_iter(term) {
+                // Only make the first part of a hyperlink is navigate-able, if there are many.
```
chrisduerr, line 82: Don't quite understand this comment? Do you mean clickable by "navigate-able"?

alacritty/src/display/hint.rs
```diff
@@ -239,6 +294,34 @@ impl HintLabels {
     }
 }
 
+/// Iterate all visible hyperlinks.
+/// Multiple adjacent cells from the same hyperlink are joined into a single range.
+fn visible_hyperlink_iter<T>(term: &Term<T>) -> impl Iterator<Item = (Hyperlink, Match)> + '_ {
```
chrisduerr, line 299:
```suggestion
fn visible_hyperlinks<T>(term: &Term<T>) -> impl Iterator<Item = (Hyperlink, Match)> + '_ {
```

alacritty/src/display/hint.rs
```diff
@@ -70,21 +73,40 @@ impl HintState {
...
-                    .map(|rm| HintPostProcessor::new(term, regex, rm).collect::<Vec<_>>())
-                    .flatten()
-                    .collect()
-            } else {
-                matches.0
+        let mut hint_matches: Vec<Match> = Vec::new();
+
+        // Find hyperlinks.
+        if hint.hyperlinks {
+            let mut visited = HashSet::new();
```
chrisduerr, line 80: Not sure what you mean by "visited" here? There is no tracking of hyperlinks that have been clicked, the concept of "visited" seems misleading.

alacritty/src/display/hint.rs
```diff
@@ -259,30 +342,65 @@ pub fn highlighted_at<T>(
...
+}
 
-            // Apply post-processing and search for sub-matches if necessary.
-            let regex_match = if hint.post_processing {
-                HintPostProcessor::new(term, regex, regex_match).find(at_point)
-            } else {
-                Some(regex_match)
-            };
+/// Iterate over all visible regex matches.
+pub fn visible_regex_match_iter<'a, T>(
```
chrisduerr, line 373:
```suggestion
pub fn visible_regex_matches<'a, T>(
```

alacritty/src/display/hint.rs
```diff
@@ -239,6 +294,34 @@ impl HintLabels {
...
+/// Multiple adjacent cells from the same hyperlink are joined into a single range.
+fn visible_hyperlink_iter<T>(term: &Term<T>) -> impl Iterator<Item = (Hyperlink, Match)> + '_ {
+    let mut grid_iter = term
+        .grid()
+        .display_iter()
+        // Keep None here. It should stop the `while` below since non-links should cut links.
+        .map(|cell| Some((cell.point, cell.hyperlink().cloned()?)))
+        .peekable();
+
+    std::iter::from_fn(move || {
```
chrisduerr, line 307:
```suggestion
    iter::from_fn(move || {
```

alacritty/src/display/hint.rs
```diff
@@ -239,6 +294,34 @@ impl HintLabels {
     }
 }
 
+/// Iterate all visible hyperlinks.
+/// Multiple adjacent cells from the same hyperlink are joined into a single range.
+fn visible_hyperlink_iter<T>(term: &Term<T>) -> impl Iterator<Item = (Hyperlink, Match)> + '_ {
+    let mut grid_iter = term
+        .grid()
+        .display_iter()
+        // Keep None here. It should stop the `while` below since non-links should cut links.
```
chrisduerr, line 303: What do you mean with this comment?

alacritty/src/display/hint.rs
```diff
@@ -239,6 +294,34 @@ impl HintLabels {
     }
 }
 
+/// Iterate all visible hyperlinks.
```
chrisduerr, line 297:
```suggestion
/// Iterate over all visible hyperlinks.
```

alacritty/src/display/hint.rs
```diff
@@ -239,6 +294,34 @@ impl HintLabels {
...
+/// Multiple adjacent cells from the same hyperlink are joined into a single range.
```
chrisduerr, line 298:
```suggestion
///
/// Multiple adjacent cells from the same hyperlink are joined into a single range.
```

alacritty/src/display/hint.rs
```diff
@@ -239,6 +294,34 @@ impl HintLabels {
...
+        .grid()
+        .display_iter()
+        // Keep None here. It should stop the `while` below since non-links should cut links.
+        .map(|cell| Some((cell.point, cell.hyperlink().cloned()?)))
+        .peekable();
+
+    std::iter::from_fn(move || {
+        // Find the next hyperlink.
+        let (start_point, hyperlink) = grid_iter.find_map(|cell| cell)?;
+        let mut end_point = start_point;
```
chrisduerr, line 310:
```suggestion
        // Find the next hyperlink.
        let (start_point, hyperlink) = grid_iter.find_map(|cell| cell)?;
        let mut end_point = start_point;
        
```

alacritty/src/display/hint.rs
```diff
@@ -239,6 +294,34 @@ impl HintLabels {
...
+        // Extend until another link or a non-link cell.
```
chrisduerr, line 311:
```suggestion
        // Extend hyperlink until it is terminated or a new link is encountered.
```

alacritty/src/display/hint.rs
```diff
@@ -259,30 +342,65 @@ pub fn highlighted_at<T>(
             return None;
         }
 
-        hint.regex.with_compiled(|regex| {
-            // Setup search boundaries.
-            let mut start = term.line_search_left(point);
-            start.line = max(start.line, point.line - MAX_SEARCH_LINES);
-            let mut end = term.line_search_right(point);
-            end.line = min(end.line, point.line + MAX_SEARCH_LINES);
+        // TODO: Use `bool::then` instead of double-if when MSRV >= 1.50.0
```
chrisduerr, line 345:
```suggestion
        // TODO: Use `bool::then` instead of double-if with Rust version (MSRV) >= 1.50.0.
```

Comment should include `Rust` and `version` substrings to make it easier to find.

alacritty/src/display/hint.rs
```diff
@@ -259,30 +342,65 @@ pub fn highlighted_at<T>(
...
+        }
+
+        None
+    })
+}
 
-            // Check if there's any match at the specified point.
-            let mut iter = RegexIter::new(start, end, Direction::Right, term, regex);
-            let regex_match = iter.find(at_point)?;
+/// Retrive the hyperlink with its range, if there is one at the specified point.
```
chrisduerr, line 366:
```suggestion
/// Retrieve the hyperlink with its range, if there is one at the specified point.
```

alacritty/src/display/hint.rs
```diff
@@ -259,30 +342,65 @@ pub fn highlighted_at<T>(
...
+fn hyperlink_at<T>(term: &Term<T>, point: Point) -> Option<(Hyperlink, Match)> {
+    let _ = term.hyperlink_at(point)?;
```
chrisduerr, line 368:
```suggestion
    term.hyperlink_at(point)?;
```

I don't think it should be necessary to bind to `_` here?

Also the logic in this method seems extremely strange to me. First we check the terminal if there's just a simple hyperlink, it tells us there is, then we go on and iterate over **all** visible hyperlinks to find it. Seems inefficient?

chrisduerr (changes requested):

kchibisov:
Closing in favor of #6139

PR #5493, opened 2021-09-16
Clean up and abstract shader creation code

oxalica:
Shader creation related codes are moved to a separated file as a lifetime-managed `ShaderProgram` in order to reduce complexity and redundancy in both rectangle and text renderer.

This also makes creating new shaders much easier if needed in the future.

### DIFF
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -1,8 +1,7 @@
 use std::collections::HashMap;
-use std::fmt::{self, Display, Formatter};
 use std::hash::BuildHasherDefault;
 use std::mem::size_of;
-use std::{io, ptr};
+use std::{fmt, ptr};
 
 use bitflags::bitflags;
 use crossfont::{
@@ -24,9 +23,18 @@ use crate::display::content::RenderableCell;
 use crate::gl;
 use crate::gl::types::*;
 use crate::renderer::rects::{RectRenderer, RenderRect};
+use crate::renderer::shader::{ShaderError, ShaderProgram};
 
 pub mod builtin_font;
 pub mod rects;
+mod shader;
+
+macro_rules! cstr {
+    ($s:literal) => {
+        // This can be optimized into an no-op with pre-allocated NUL-terminated bytes.
+        unsafe { std::ffi::CStr::from_ptr(concat!($s, "\0").as_ptr().cast()) }
+    };
+}
 
 // Shader source.
 static TEXT_SHADER_F: &str = include_str!("../../res/text.f.glsl");
@@ -45,30 +53,31 @@ pub trait LoadGlyph {
 
 #[derive(Debug)]
 pub enum Error {
-    ShaderCreation(ShaderCreationError),
+    /// Shader error.
+    Shader(ShaderError),
 }
 
 impl std::error::Error for Error {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         match self {
-            Error::ShaderCreation(err) => err.source(),
+            Error::Shader(err) => err.source(),
         }
     }
 }
 
-impl Display for Error {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Error::ShaderCreation(err) => {
+            Error::Shader(err) => {
                 write!(f, "There was an error initializing the shaders: {}", err)
             },
         }
     }
 }
 
-impl From<ShaderCreationError> for Error {
-    fn from(val: ShaderCreationError) -> Self {
-        Error::ShaderCreation(val)
+impl From<ShaderError> for Error {
+    fn from(val: ShaderError) -> Self {
+        Error::Shader(val)
     }
 }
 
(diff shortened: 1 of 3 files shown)


chrisduerr (changes requested):
Thanks for the PR.

Is there a concrete motivation behind these changes?

oxalica:
> Is there a concrete motivation behind these changes?

1. Just because these code duplicates. I think the question should be why not abstract them.
2. I'm planning some more modification later. it makes things easier.

alacritty/Cargo.toml
```diff
@@ -34,6 +34,8 @@ libc = "0.2"
 unicode-width = "0.1"
 bitflags = "1"
 dirs = "3.0.1"
+cstr = "0.2.8"
+thiserror = "1.0.29"
```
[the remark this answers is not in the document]
oxalica: `cstr` is removed now.

For `thiserror`, it's widely used and already a transient dependency of us in master. It saves code for `Display` and `Error` impl and makes them easier to maintain. It definitely worth it.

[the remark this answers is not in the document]
chrisduerr: I disagree, I don't think `thiserror` provides value other than artificially creating conciseness by using an unintuitive macro DSL. So I have no interest in using it inside of Alacritty.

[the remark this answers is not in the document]
oxalica: Okay, okay. It's now removed.

alacritty/src/renderer/mod.rs
```diff
@@ -45,30 +53,32 @@ pub trait LoadGlyph {
 
 #[derive(Debug)]
 pub enum Error {
-    ShaderCreation(ShaderCreationError),
+    /// Shader error.
+    /// Currently it must be about initialization.
```
chrisduerr, line 57:
```suggestion
```

alacritty/src/renderer/mod.rs
```diff
@@ -25,8 +24,17 @@ use crate::display::content::RenderableCell;
 use crate::gl;
 use crate::gl::types::*;
 use crate::renderer::rects::{RectRenderer, RenderRect};
+use crate::renderer::shader::{ShaderError, ShaderProgram};
 
 pub mod rects;
+pub mod shader;
```
chrisduerr, line 30: Does it need to be pub?

alacritty/src/renderer/mod.rs
```diff
@@ -687,7 +697,7 @@ impl QuadRenderer {
         F: FnOnce(RenderApi<'_>) -> T,
     {
         unsafe {
-            gl::UseProgram(self.program.id);
+            gl::UseProgram(self.program.id());
```
chrisduerr, line 700: Curious what you'd think of a `self.program.use()` method or something like that.

I feel like if we're already creating an object for the program that provides fields and methods like `id`, we could also use that to interact with the OpenGL program in a more object-y way.

chrisduerr (changes requested):

alacritty/src/renderer/mod.rs
oxalica: Pair of `program.use()` and `clear_program()` doesn't really differ from pair of `gl::UseProgram`.
If we want more abstraction, they should be `self.program.with_program(|| {})`. But currently we manually control OpenGL state everywhere. I won't refactor all of these OpenGL calls entirely. Instead I want to keep the modification of this PR to be small and only related to shader creation.

BTW: `gl` itself is a low level API binding crate. If we want to make these calls more clean, we would looking for some high-level crate first. But it's out of the scope of this PR.

chrisduerr: > Pair of program.use() and clear_program() doesn't really differ from pair of gl::UseProgram.
If we want more abstraction, they should be self.program.with_program(|| {}).

Only because an even higher level of abstraction exists that isn't worth investing time into, doesn't mean the intermediate step isn't worth it. For consumer's it's certainly easier to see that you can just `.use()` a program, rather than having to figure out that the program's ID is what you need to pass to `gl::UseProgram`. I personally don't care *that* much, but I think your argument is pretty weak.

> we would looking for some high-level crate first

We shouldn't. There's not really a much higher level API that is suitable for Alacritty.

chrisduerr: All instances of `useProgram` I could find were directly after explicitly using a program with `gl::UseProgram(self.program.id())`. So it would make sense to use that `self.program` to "un-use" it again too.

This could either be done with a method on program (like `self.program.clear_program()`), or by doing both automatically using `self.program.with_program(|| { … })`.

chrisduerr (changes requested):

chrisduerr (approved):
This PR got somewhat lost in my notifications. I think @oxalica is right and it's best to just stick with consistently using `gl::` functions directly everywhere.

alacritty/src/renderer/shader.rs
```diff
@@ -0,0 +1,161 @@
...
+        gl::GetShaderInfoLog(shader, max_length, &mut actual_length, buf.as_mut_ptr() as *mut _);
+    }
+
+    // Build a string.
+    unsafe {
+        buf.set_len(actual_length as usize);
+    }
+
+    // XXX should we expect OpenGL to return garbage?
+    String::from_utf8(buf).unwrap()
```
kchibisov, line 136: I was looking into that, and can't we use lossy convertion just to remove this unwrap?

chrisduerr: Pretty sure I didn't mention that because it also exists in the old version, but I don't see why we wouldn't.

chrisduerr: Done.

kchibisov (approved):

PR #5494, opened 2021-09-17
Update INSTALL.md

hxn-dev:
add python3 to arch dependencies list


chrisduerr (changes requested):
That's not even a valid package on archlinux.

Issue #4703, opened 2021-01-14
Alacritty fails to start in 10bit color mode


pointshader:
I followed the following guide to enable 10bit color on my machine. 10bit color appears to working for other programs, but alacritty fails to start.

 ```
Creating a /etc/X11/xorg.conf.d/30-screensetup.conf with the following
content will give you 10-bit color:

Section "Screen"
  Identifier    "Default Screen"
  Monitor        "Configured Monitor"
  Device         "Configured Video Device"
  #               24 for 8-bit or 30 for 10-bit
  DefaultDepth    30
EndSection

And that's it. DefaultDepth 30 is the crucial line. The reason 30 enables 10-bit color is that R, G, B = 3 colors * 10 bits = 30. 
```
(Guide from https://linuxreviews.org/HOWTO_enable_10-bit_color_on_Linux)

### System

OS: Arch Linux
Version: alacritty 0.7.1 (a852eb85)
Linux/BSD: X11 & i3
Nvidia 3090 with driver 460.32.03

chrisduerr:
How does it fail to start? What error do you get?

pointshader:
### Error
[2021-01-14 05:05:26.680909269] [ERROR] [alacritty] Alacritty encountered an unrecoverable error:
                                                    
                                                    	Error creating GL context; Received multiple errors. Errors: `[NoAvailablePixelFormat, NoAvailablePixelFormat]`

pointshader:
(I was rebooting to go back into 10bit mode as you were responding, @chrisduerr !
I'm not an x11 expert, but what's interesting is that it can't get a PixelFormat, yet in nvidia-smi, `X Screen 0 depth` is recognized as `30`. This is despite the fact that I can't set the gpu to 10bcc in the nvidia-smi interface, as the drop down only support depths of Auto, 6 bpc, and 8 bpc. This is using a 10bit monitor, verified hardware setup in Windows.

chrisduerr:
Can you start other OpenGL programs like glxinfo?

pointshader:
And for clarification, removing the conf file and rebooting resets the reported depth to `24`, and alacritty starts normally. I encountered some sluggishness with other programs running in 10bit mode, so I'll be disabling 10bit for now. So personally, no rush for this issue. But long term, I imagine more users will run into the same problem. 

Shoutouts to all the devs for making such a great terminal emulator!

pointshader:
[...]

PR #5495, opened 2021-09-17
Try running in 30bpp (10-bit color) if default fails

abdelq:
Related to #4703. Tries setting the color bits to 30 before failing in case 10-bit color is set.

Not the biggest fan of that chain of `or_else`, but it's not that bad. If you have any idea for improvement, let me know.

### DIFF
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -127,17 +127,21 @@ fn create_gl_window<E>(
     event_loop: &EventLoop<E>,
     srgb: bool,
     vsync: bool,
+    deep_color: bool,
     dimensions: Option<PhysicalSize<u32>>,
 ) -> Result<WindowedContext<PossiblyCurrent>> {
     if let Some(dimensions) = dimensions {
         window = window.with_inner_size(dimensions);
     }
 
-    let windowed_context = ContextBuilder::new()
-        .with_srgb(srgb)
-        .with_vsync(vsync)
-        .with_hardware_acceleration(None)
-        .build_windowed(window, event_loop)?;
+    let mut windowed_context =
+        ContextBuilder::new().with_srgb(srgb).with_vsync(vsync).with_hardware_acceleration(None);
+
+    if deep_color {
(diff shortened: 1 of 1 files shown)


chrisduerr:
Have you verified that this fixes a crash when trying to start Alacritty on a 10-bit system?

alacritty/src/display/window.rs
```diff
@@ -186,9 +190,29 @@ impl Window {
...
+                })
+                .or_else(|_| {
+                    create_gl_window(
+                        window_builder.clone(),
+                        event_loop,
+                        false,
+                        !is_wayland,
+                        true,
+                        size,
+                    )
```
chrisduerr, line 212: The entire point of this is to act as a fallback in case one doesn't work, combining both srgb and 10-bit mode in the fallback doesn't sound like a good idea. We're having two branches here for a possibility of 4 different setups.

abdelq: Hence my comment about looking for ideas on how to do it better.

abdelq:
> Have you verified that this fixes a crash when trying to start Alacritty on a 10-bit system?

Yes, it works.

alacritty/src/display/window.rs
chrisduerr: With the color format the issue is that it might be different for every monitor. I'm not sure if depth can be defined on a per-monitor level but generally we should try to figure this out in advance.

Connor-GH:
This makes the app launch when 10-bit color is on. I have tested with the PR's compiled src

alacritty/src/display/window.rs
kchibisov: Glutin doesn't allow to do this now unfortunately. I'd suggest to fallback to srgb + 10 bit in the end.

kchibisov (changes requested):
Given that it fixes issue I'm not against it, however could we cache state we've picked so multi window won't probe everything once again?

I think we can use atomic globals (local to window.rs) for that or something like that?

PR #5496, opened 2021-09-18
Unix: Hide TTY close EIO in Rust 1.55+

ilya-bobyr:
`ErrorKind::Other` no longer includes `EIO` since Rust 1.55:

    https://blog.rust-lang.org/2021/09/09/Rust-1.55.0.html#stdioerrorkind-variants-updated

It was not precise enough from the very beginning, as the comment says
that only EIO should be hidden, while the code was any uncategorised
errors.

### DIFF
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -394,7 +394,7 @@ where
                                     // This sucks, but checking the process is either racy or
                                     // blocking.
                                     #[cfg(target_os = "linux")]
-                                    if err.kind() == ErrorKind::Other {
+                                    if err.raw_os_error() == Some(libc::EIO) {
                                         continue;
                                     }
 


chrisduerr (commented):
What is the actual errorkind? Uncategorized?

ilya-bobyr:
> What is the actual errorkind? Uncategorized?

It is.  But the doc says that it should not be used.
Besides, there could be other errors in this category.
And the comment says that we only want to filter EIO.

Here is the code in the standard library that is classifying Unix errno into categories:
https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/src/std/sys/unix/mod.rs.html#138-185

`ErrorKind::Uncategorized` is both `hidden` and `unstable`:
https://doc.rust-lang.org/stable/src/std/io/error.rs.html#297-304

chrisduerr (approved):

PR #5515, opened 2021-09-27
Fix vi indicator obstructing vi mode cursor

### LINKED ISSUE #5504: Vi mode line indicator not hidden when hovering it with display offset
If vi mode is active and the cursor is in the top-right position, the vi mode line indicator is only hidden when the display offset is 0. If the display offset is 1 or higher, the cursor is hidden by the line number indicator.

The indicator should be hidden regardless of the display offset when the vi mode cursor is below it.

chrisduerr:
Fixes #5504.


chrisduerr:
@perfbot

PR #5517, opened 2021-09-28
Fix incorrect vi mode search origin

### LINKED ISSUE #5460: Vi mode search does not start at cursor
When a vi mode search is started while the bottommost line on the buffer is filled, a search will move the vi cursor causing the search to start one line below the original vi cursor position. This can cause the search to wrap around the buffer, confusing the user (me in this case).

chrisduerr:
Fixes #5460.


chrisduerr:
This regression was introduced in 3bd5ac221ab3b122962063edd1f4c10f9f2d117f.

PR #5527, opened 2021-10-03
Fix invalid repetition operator in help target regex

afh:
When using gmake 4.3 with grep 0.9 on OpenBSD 6.9, running `gmake help` results in the following error: `grep: repetition-operator operand invalid`.

Use of grep(1) can be omitted by use of an pattern-action in awk(1).

This change was also tested on macos 11.6

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ vpath $(DMG_NAME) $(APP_DIR)
 all: help
 
 help: ## Print this help message
-	@grep -E '^[a-zA-Z._-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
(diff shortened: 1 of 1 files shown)


chrisduerr:
I much prefer the previous version and the makefile isn't meant for OpenBSD anyway. Is there a reason why you want to run the makefile on OpenBSD?

afh:
I'm rather unfamiliar with rust, so the Makefile is the only way I knew to compile alacritty on OpenBSD, but it seems `cargo b` works just as well, so I'm going to close this PR. Thanks for considering and reviewing.

While I have your attention: Currently I do have to invoke `cargo b` with `RUSTFLAGS=-L/usr/local/lib` as that's where `libxkbcommon.so` is installed. Is it possible to specify the additional library search path for OpenBSD in Cargo.toml? If yes how?

chrisduerr:
> Is it possible to specify the additional library search path for OpenBSD in Cargo.toml?

Not that I'm aware of.

afh:
Alright, `RUSTFLAGS=-L/usr/local/lib` it is then :smile:

PR #5528, opened 2021-10-03
Add option to escape a dropped file's path

dsmatter:
Currently, drag & dropping a file into an alacritty window inserts its path unescaped with a trailing space. Especially when dragging multiple files with spaces (or shell control characters) in their names as arguments to a shell command, manually escaping the file paths can become pretty tedious. Other terminal emulators (e.g. macOS Terminal or iTerm) hence insert the file paths in an escaped form suitable for most shells.

This PR proposes the new config option `mouse.escape_dropped_path` which lets users choose from these escape modes:

- `None`: don't escape the file path (default and current behaviour)
- `SingleQuotes`: wrap the file path in single quotes (and escape single quotes in the file path)
- `Backslashes`: bachslash-escape spaces and common shell control characters

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Option `colors.transparent_background_colors` to allow applying opacity to all background colors
+- Option `mouse.escape_dropped_path` to control escaping of a dropped file's path
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


chrisduerr:
As previously discussed, this is not something I'm interested in adding.

kchibisov:
Maybe we can not escape but rather quote, though you can have quotes in a filenames.

chrisduerr:
Quoting was discussed, it's much easier to add quotes yourself than to remove them. Alacritty really doesn't need to support 6 different ways to drop files. The only reason why it was added in the first place is that it was just 4 LoCs due to upstream implementing all the logic. Blowing that completely out of proportion with a million config options is not worth it.

PR #5531, opened 2021-10-06
Update INSTALL.md

TQMatvey:
needed cargo for building

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -84,7 +84,7 @@ to build Alacritty. Here's an apt command that should install all of them. If
 something is still found to be missing, please open an issue.
 
 ```sh
-apt-get install cmake pkg-config libfreetype6-dev libfontconfig1-dev libxcb-xfixes0-dev libxkbcommon-dev python3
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -84,7 +84,7 @@ to build Alacritty. Here's an apt command that should install all of them. If
 something is still found to be missing, please open an issue.
 
 ```sh
-apt-get install cmake pkg-config libfreetype6-dev libfontconfig1-dev libxcb-xfixes0-dev libxkbcommon-dev python3
+apt-get install cmake pkg-config libfreetype6-dev libfontconfig1-dev libxcb-xfixes0-dev libxkbcommon-dev python3 cargo
```
kchibisov, line 87: I wonder if we should explicitly mention `cargo` in deps here since we ask to install rust via `rustup` in a section above.

We can remove this section or move it and also add cargo to every install command.

kchibisov (changes requested):

INSTALL.md
chrisduerr: We shouldn't, that's why we don't have it in any other of the installation instructions. 

Of course Rust could be installed through the package manager and chances are that might work okay for some of the distros. But generally I see no reason to change this now and with distros like debian that's just going to cause issues.

PR #5533, opened 2021-10-06
Fixing hint post processing close brackets.

### LINKED ISSUE #5492: Hint post processing crash
Alacritty with the following config crashes when hovering over the `)` character.

```yaml
hints:
  enabled:
   - regex: "[^/ ]"
     action: Paste
     post_processing: true
     mouse:
       enabled: true
       mods: None
```

It only crashes with post processing enabled and I get no relevant output with `-vvv` or `--print-events`.

`alacritty 0.9.0 (fed349aa)` on Linux + Wayland (sway)

gsfreema:
Fixes #5492 

When the match is as single character, don't exclude any character from the range during hint post processing. Otherwise, the HintPostProcessor can end up in an infinite loop it is iterated over.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Line indicator obstructing vi mode cursor when scrolled into history
 - Vi mode search starting in the line below the vi cursor
+- Invisible cursor with matching foreground/background colors
 
 ## 0.9.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Option `colors.transparent_background_colors` to allow applying opacity to all background colors
+- Support for running multiple windows from a single Alacritty instance (see docs/features.md)
 
 ### Changed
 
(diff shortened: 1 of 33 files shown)


gsfreema:
It does make sense that the function should function the same regardless of the match length, but that does mean that it will return an invalid RangeInclusive<Point> since the end of the range will be before the start of the range which is what the current functionality does and why it's resulting in an infinite loop when iterating over it.

gsfreema:
No worries. I can investigate some of those options. I had thought about an exclusive range but so many other pieces expected an inclusive range that I felt that was best to work with. According to the inclusive range docs, an end that is before the start is considered an empty range which is why I thought it might be ok to return.

chrisduerr:
>  I had thought about an exclusive range but so many other pieces expected an inclusive range that I felt that was best to work with.

Yeah with an exclusive range you'd likely be checking if it's empty or not and then convert it immediately anyway. So at that point you might as well just return an option.

chrisduerr:
Could you please rebase this on top of the latest master? It doesn't rebase cleanly and is not reviewable in its current state.

roland-rollo:
Why do you close this?

gsfreema:
My limited git knowledge left me with 32 files changes every time I tried rebasing on upstream/master. I don't really have the time to figure out why that kept happening so it was far easier and faster for me to update the single file necessary for a fix in a different branch and create a new PR.

PR #5536, opened 2021-10-10
Add `min_dpr` and `max_dpr` config values + minor fix

JOT85:
## `min_dpr` and `max_dpr`

I have differently sized displays with the same resolution - one a laptop screen, another a larger monitor. They both have different scale factors and I like how alacritty scales between them. However I'd like to have the font slightly smaller on my laptop, but then it becomes slightly too pixelated to properly read on my larger display.

I wouldn't like to set a fixed dpr, since that would disable scaling, and I don't want to update my display's scale factor as the scaling is perfect for everything else!

I've added `min_dpr` and `max_dpr` values to the config, which allow you set a minimum and maximum value for the dpr scaling. In my use case, I can set the maximum scaling to 1.3 and then set my font size accordingly and it doesn't get too large on my laptop :)

This could also be used, as described in the config comment, to disable scaling by setting the minimum and maximum both to 1.

## Minor fix

Also, there was a check for X11 displays, which reset the dpr if the value was too high do to a configuration error. This check wasn't performed when a window was switched between monitors though. So the error isn't detected when moved onto another screen with this problem.

I've moved all updates to the cached dpr value into the same method, meaning that the same calculations are always performed, including this check. The `dpr` field is now private to avoid accidentally mutating it directly.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1462,6 +1462,7 @@ impl<N: Notify + OnResize> Processor<N> {
             processor.ctx.window().set_title(&config.ui_config.window.title);
         }
 
+        #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
         if is_wayland {
             processor.ctx.window().set_wayland_theme(&config.ui_config.colors);
         }


JOT85:
As mentioned in the original comment, I didn't want to explicitly set the environment variable, as that would disable scaling entirely. I still want there to be scaling, but limited by these values - so there would be a difference between screens, but not as large.

JOT85:
@chrisduerr it's obviously up to you, and you might have already made your decision, but before this is dismissed, I'd like to make my problem really clear!

Here's how my terminal looks on my laptop:

Example0

It looks great! The font size is perfect and it's readable.

Here's how it looks when it's on my monitor side by side:

Example1

It's scaled down, which makes sense since the monitor is larger. All of the sizes and DPI values are reported correctly.

But, the problem is, that's only a 1080p monitor, so if you look at the text, it's not very readable:

Example2

The scaling isn't wrong! It's just that my font size is too small for my 24" monitor to render it clearly. If I make the font size any larger, however, it becomes too large for the laptop display!

## Alternative Solutions
### Disable Scaling
As you suggested, I could set the environment variable to fix the font scaling, I could also set the scaling to a constant elsewhere, but then the font does not scale down from my laptop to my monitor and is too large! The same font size isn't nice on both my laptop and my monitor, so I either need scaling or to manually change the zoom when I move between them, which isn't nice. But I can't use the default scaling because I can't read the text!
### Falsifying the screen sizes...
This seems simple. I thought it would be easy. It is not... In fact, trying to lie about the physical screen size is actually really hard!

It's also not ideal since I don't need to lie to every program about my screen size.

## This solution
This solution allows people to limit how small the font gets scaled, preventing it from becoming unreadable! I don't think there's a proper alternative?

## Also the bug fix...
Whether or not the config options are added, it might still be worth fixing the bug where the X11 bound is not checked when the window is moved between screens?

JOT85:
Thanks for the response. I understand your stance! I'll see what I can get working on Wayland.

I'll submit a PR for the bugfix alone tomorrow.

PR #5537, opened 2021-10-10
Change order of checks in `.builds` files

JOT85:
This is a really minor thing, it's ok if you disagree with my reasoning for it!

When I submitted a PR, I used the stable version of rustfmt instead of nightly, which cause the rustfmt check to fail. Since rustfmt was after the main `cargo test`, it took a long time to fail. If rustfmt was first, I'd have noticed much sooner and it would have saved a bit of time! The formatting checks are really fast, too, so it passes quickly and doesn't make much of a difference to the time until the later tests are run.


chrisduerr:
No, the order is as-is for a reason.

PR #5539, opened 2021-10-11
Fix: Always check X11 dpr is sane

JOT85:
Following discussion of #5536, this only contains the fix to the X11 bug.

Only update and receive dpr values through `update_dpr` and `get_dpr`, `update_dpr` method always perform the X11 dpr sanity check. Previsouly, this check was performed on startup but not when moving between screens. This means that when moving to a screen with an insane dpr value, it would be used, but when started on that screen, it would not.


kchibisov (changes requested):
I wonder why we're checking for dpi here and not in winit? I think it could and **should** be easily done there.

kchibisov:
See https://github.com/rust-windowing/winit/pull/1403

JOT85:
@kchibisov we should probably remove the check that was there in the first place since winit does it now.

kchibisov:
we should, but with winit update, since check isn't released in winit yet.

PR #5545, opened 2021-10-16
Remove note about larger datasize-cur requirement

afh:
I'm able to build alacritty with the default `datasize-cur` of `1536M` on the latest version of OpenBSD, i.e. 7.0, so it seems increasing `datasize-cur` is no longer a requirement.

In case there is worry that this may come up in the future again, I'm happy to reword this, e.g.: "The defaults user limits _may_ be insufficient to build Alacrity, in that case [...]".

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -153,7 +153,7 @@ pkg install cmake freetype2 fontconfig pkgconf python3
 
 #### OpenBSD
 
-On OpenBSD 6.5, you need [Xenocara](https://xenocara.org) and Rust to build
+On OpenBSD you need [Xenocara](https://xenocara.org) and Rust to build
 Alacritty, plus Python 3 to build its XCB dependency. If something is still
 found to be missing, please open an issue.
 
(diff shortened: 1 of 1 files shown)


j3lli21:
@jelli215

afh:
Thanks for the insight @chrisduerr, much appreciated!

If find the recommendation to change `datasize-cur` in `/etc/login.conf` too intrusive and permanent for my taste, so I wanted to find another solution.
After some digging around I learned that OpenBSD provides the `_pbuild` user which also has the [`pbuild` login class](https://github.com/openbsd/src/search?q=*pbuild+filename%3Alogin.conf); it has a much more reasonable datasize set (for me on amd64 it is 8G!).

The `_pbuild` user is used by the distributed ports builder (see [dpb(1)](https://man.openbsd.org/dpb)), so it seems to be well suited for building Alacritty (and possibly other software with more demanding user limit requirements).

I have updated the Alacritty installation (or rather building) instructions on OpenBSD, so that the `_pbuild` user is being used without the need to change system files :slightly_smiling_face: 

What are your thoughts on this?

:information_source: I've tried and tested this using the recently released OpenBSD 7.0 :blowfish:

PR #5575, opened 2021-10-31
Implement passing options via IPC

### LINKED ISSUE #5560: Allow passing arguments to `alacritty msg create-window`
The `alacritty msg create-window` command currently takes no parameters.

Ideally it should be possible to have the same Alacritty instance running, but spawn windows using different commands using the `-e` parameter. Passing other arguments like `--working-directory` might also make sense.

### LINKED ISSUE #5561: `alacritty msg create-window` copies shell
#5560 already tracks the option to specify a separate shell for Alacritty's `create-window` command, however by default the command of the first executable is still copied. This doesn't make much sense and can lead to confusing behavior.

Ideally the parameters of the initial window should be largely ignored and replaced with the default. Then new parameters can be passed through the command when #5560 is resolved.

Repro:
```sh
> unset ALACRITTY_SOCKET
> alacritty -e sh -c "while true; do date; sleep 1; done" &
> alacritty msg create-window
# => Should show two windows doing the same date loop
```

### LINKED ISSUE #5562: CreateNewWindow command should copy working directory
Since the CreateNewWindow command is always associated to one specific Alacritty window, this window's working directory should be copied when spawning new windows.

kchibisov:
This commit implements passing arguments via CLI, preserving working directory, and also adds handling when spawning process for pty fails.

Fixes #5562.
Fixes #5561.
Fixes #5560.


alacritty/src/ipc.rs
```diff
@@ -1,25 +1,30 @@
...
-
 /// Environment variable name for the IPC socket path.
 const ALACRITTY_SOCKET_ENV: &str = "ALACRITTY_SOCKET";
 
+const JSON_SEPARATOR: u8 = b'\n';
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub enum IPCMessage {
+    CreateWindow(TerminalOptions),
+}
```
chrisduerr, line 26: Some docs on this enum would be nice just to know what it's used for.

Also in Rust the correct capitalization is `IpcMessage`.

alacritty/src/ipc.rs
```diff
@@ -40,13 +45,39 @@ pub fn spawn_ipc_socket(options: &Options, event_proxy: EventLoopProxy<Event>) -
...
-                let _ = event_proxy.send_event(Event::new(EventType::CreateWindow, None));
+        let mut data = Vec::new();
+        for stream in listener.incoming() {
+            let stream = match stream {
+                Ok(stream) => stream,
+                Err(err) => {
+                    warn!("Failed to get incomming stream! {}", err);
+                    continue;
+                },
+            };
```
chrisduerr, line 56: When does this error? I'd assume when the listener is killed or something?

I feel like we can either just `.filter` on `incoming()` or the error is actually an issue and we `break`?

alacritty/src/ipc.rs
```diff
@@ -40,13 +45,39 @@ pub fn spawn_ipc_socket(options: &Options, event_proxy: EventLoopProxy<Event>) -
...
+
+            data.clear();
+            let mut stream = BufReader::new(stream);
+
+            match stream.read_until(JSON_SEPARATOR, &mut data) {
+                Ok(0) | Err(_) => continue,
+                Ok(bytes_read) => bytes_read,
+            };
```
chrisduerr, line 64: I don't think `\n` as `JSON_SEPARATOR` makes a lot of sense. Probably don't need any constant at all but if we do it should probably be `MESSAGE_SEPARATOR` or something since it's not actually related to the format at all.

More importantly though, this method looks a lot like `read_line`. Do we really need to store things in a `Vec` for serde, can't we just use a `String` instead and sidestep this issue of actually passing a separator?

alacritty/src/ipc.rs
```diff
@@ -40,13 +45,39 @@ pub fn spawn_ipc_socket(options: &Options, event_proxy: EventLoopProxy<Event>) -
...
+                    continue;
+                },
+            };
+
+            data.clear();
+            let mut stream = BufReader::new(stream);
+
+            match stream.read_until(JSON_SEPARATOR, &mut data) {
+                Ok(0) | Err(_) => continue,
+                Ok(bytes_read) => bytes_read,
```
chrisduerr, line 63:
```suggestion
                Ok(_) => (),
```

alacritty/src/ipc.rs
```diff
@@ -40,13 +45,39 @@ pub fn spawn_ipc_socket(options: &Options, event_proxy: EventLoopProxy<Event>) -
...
+            };
+
+            // Read pending events on socket.
+            let message: IPCMessage = match serde_json::from_slice(&data) {
+                Ok(message) => message,
+                Err(err) => {
+                    warn!("Fail to get data from socket: {}", err);
```
chrisduerr, line 70:
```suggestion
                    warn!("Fail to convert data from socket: {}", err);
```

I think convert is better than `get` here, since we were able to get the data from the socket the only thing that didn't work is parsing it.

alacritty/src/ipc.rs
```diff
@@ -55,9 +86,13 @@ pub fn spawn_ipc_socket(options: &Options, event_proxy: EventLoopProxy<Event>) -
...
 /// Send a message to the active Alacritty socket.
-pub fn send_message(socket: Option<PathBuf>, message: &[u8]) -> IoResult<()> {
-    let socket = find_socket(socket)?;
-    socket.send(message)?;
+pub fn send_message(socket: Option<PathBuf>, message: IPCMessage) -> IoResult<()> {
+    let mut socket = find_socket(socket)?;
+
+    let mut message = serde_json::to_string(&message)?;
+    message.push(JSON_SEPARATOR as char);
+    socket.write_all(message[..].as_bytes())?;
```
chrisduerr, line 94: Is it actually necessary to push the `JSON_SEPARATOR` here? Seems somewhat strange from an API perspective for Rust, doesn't `flush` work?

alacritty/src/ipc.rs
```diff
@@ -55,9 +86,13 @@ pub fn spawn_ipc_socket(options: &Options, event_proxy: EventLoopProxy<Event>) -
 }
 
 /// Send a message to the active Alacritty socket.
-pub fn send_message(socket: Option<PathBuf>, message: &[u8]) -> IoResult<()> {
-    let socket = find_socket(socket)?;
-    socket.send(message)?;
+pub fn send_message(socket: Option<PathBuf>, message: IPCMessage) -> IoResult<()> {
+    let mut socket = find_socket(socket)?;
+
+    let mut message = serde_json::to_string(&message)?;
```
chrisduerr, line 92: I just wanted to point out that back when I looked at this, I was considering a binary exchange format. I think in theory that would be pretty nice because the data sent over the wire would be much more efficiently encoded.

But from what I've seen that would require pulling another big dependency just for doing this, so I think just using what we have already is probably smart.

alacritty/src/ipc.rs
```diff
@@ -1,25 +1,30 @@
...
-pub const SOCKET_MESSAGE_CREATE_WINDOW: [u8; 1] = [1];
-
 /// Environment variable name for the IPC socket path.
 const ALACRITTY_SOCKET_ENV: &str = "ALACRITTY_SOCKET";
 
+const JSON_SEPARATOR: u8 = b'\n';
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub enum IPCMessage {
+    CreateWindow(TerminalOptions),
```
chrisduerr, line 25: Do we want to somehow version the messages sent through the protocol? I think it should be fine without it since we just serde deserialize everything anyway and that will return an error if things don't match, but I'm curious if you've considered this.

Unfortunately the only interesting part for versioning would be forward-compatibility, since you'll rarely have a server that is newer than the client connecting to the socket. And that can't really be done anyway, so probably pointless.

chrisduerr (changes requested):
Dope, thanks for the PR.

Looked through everything and added comments at all the relevant places, generally looks really good. Haven't tested it yet but I'd assume it just works.

alacritty/src/cli.rs
```diff
@@ -34,10 +35,6 @@ pub struct Options {
     #[structopt(long)]
     pub embed: Option<String>,
```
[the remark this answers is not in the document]
kchibisov: Hm, that makes sense, just added something basic for now. Will look into them as well.

alacritty/src/event.rs
```diff
@@ -1302,8 +1295,12 @@ impl Processor {
                     }
                 },
                 // Create a new terminal window.
-                GlutinEvent::UserEvent(Event { payload: EventType::CreateWindow, .. }) => {
-                    if let Err(err) = self.create_window(event_loop, proxy.clone()) {
+                GlutinEvent::UserEvent(Event {
+                    payload: EventType::CreateWindow(options), ..
+                }) => {
+                    let options = if options.is_empty() { Some(options) } else { None };
```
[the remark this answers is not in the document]
kchibisov: I initially was transferring `Option` but then decided to add empty check. Could probably move it back to option or something else.

alacritty/src/config/mod.rs
```diff
@@ -32,7 +32,7 @@ use crate::config::ui_config::UiConfig;
 /// Maximum number of depth for the configuration file imports.
 const IMPORT_RECURSION_LIMIT: usize = 5;
 
-pub type Config = TermConfig<UiConfig>;
+pub type Config = UiConfig;
```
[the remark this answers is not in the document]
kchibisov: Yeah, I think to just call it `Config` and that's about it.

alacritty/src/daemon.rs
```diff
@@ -24,6 +30,27 @@ where
     }
 }
 
+#[cfg(not(windows))]
+pub fn foreground_process_path() -> Result<PathBuf, Box<dyn Error>> {
```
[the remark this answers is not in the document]
kchibisov: Yeah, makes sense.

alacritty/src/window_context.rs
```diff
@@ -81,15 +84,26 @@ impl WindowContext {
...
-        let pty = tty::new(config, &display.size_info, display.window.x11_window_id());
+        let mut new_pty_config = PtyConfig::new();
+        let pty_config = match options {
+            Some(mut options) => {
+                new_pty_config.working_directory = options.working_directory.take();
+                new_pty_config.shell = options.command();
+                new_pty_config.hold = options.hold;
+                &new_pty_config
+            },
+            None => &config.terminal_config.pty_config,
```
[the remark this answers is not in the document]
kchibisov: I agree, but I'm not sure where to put it. will try moving things around, maybe something works out.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -143,7 +144,7 @@ fn default_shell(pw: &Passwd<'_>) -> Program {
 }
 
 /// Create a new TTY and return a handle to interact with it.
-pub fn new<C>(config: &Config<C>, size: &SizeInfo, window_id: Option<usize>) -> Pty {
+pub fn new(config: &PtyConfig, size: &SizeInfo, window_id: Option<usize>) -> Result<Pty> {
```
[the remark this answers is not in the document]
kchibisov: It's required to not crash entire process when you do something like `alacritty msg create-window -e not_found`, since without it you'd bring entire process down instead of just a failure to create window.

alacritty/src/ipc.rs
kchibisov: we continue, just log a warning that some client failed to connect to listener, which is our main application.

alacritty/src/ipc.rs
kchibisov: I wasn't entirely sure how data is separated on the wire here, so I just wrote it in a way I'd know it'll work for sure. Will try to optimize it.

alacritty/src/ipc.rs
kchibisov: I was also thinking about using protobuf, msgpack or some other fancy thing, but then I see serde and unix sockets in rust stdlib, so I decided to pick whatever I have, since the bandwidth is low anyway and json is fine here...

alacritty/src/ipc.rs
kchibisov: Not sure about this one.

chrisduerr: I personally think just not doing any versioning should be perfectly fine. Do you have any arguments for why it might be smart to do it?

alacritty/src/ipc.rs
chrisduerr: You know of any example this can be reproduced? Is this generally just an issue with the client and safe to skip as a server?

alacritty_terminal/src/tty/unix.rs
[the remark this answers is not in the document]
chrisduerr: Makes sense anyway, better do it correct right away. Great.

alacritty/src/config/mod.rs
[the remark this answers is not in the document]
chrisduerr: Yeah either that or use `UIConfig` everywhere. Either way a type alias isn't necessary.

alacritty/src/event.rs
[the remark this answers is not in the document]
chrisduerr: I think it makes sense to just always have it around. Even if it's stored somewhere else at least make people pass that explicitly. If you then use `Into` for the argument you might even get it to be dynamic where it doesn't matter if you pass `TerminalOption` or `PtyOption` or whatever.

alacritty/src/cli.rs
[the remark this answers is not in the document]
chrisduerr: Yeah, once we got the basics figured out it shouldn't be a problem to do other stuff later. Title/Class are just the main ones that would likely be interesting.

alacritty/src/ipc.rs
kchibisov: I'm fine without versioning for now, since feature isn't even released, could think about it close to release.

alacritty/src/ipc.rs
kchibisov: agree.

kchibisov:
@chrisduerr I've fixed most code style issues, I think only config issue is left.

chrisduerr:
Sweet, that was quick. I'll take a look tomorrow.

alacritty/src/cli.rs
```diff
@@ -214,6 +201,38 @@ fn parse_class(input: &str) -> Result<Class, String> {
...
 }
 
+/// Terminal specific cli options which can be passed to new windows via IPC.
+#[derive(Serialize, Deserialize, StructOpt, Default, Debug, Clone, PartialEq)]
+pub struct TerminalOptions {
+    /// Start the shell in the specified working directory.
+    #[structopt(long)]
+    pub working_directory: Option<PathBuf>,
+
+    /// Remain open after child process exits.
```
chrisduerr, line 211:
```suggestion
    /// Remain open after child process exit.
```

Either that or `after the child process exits`. Just to be pedantic.

alacritty/src/config/mod.rs
```diff
@@ -27,20 +27,18 @@ pub use crate::config::bindings::{
...
-pub type Config = TermConfig<UiConfig>;
-
 /// Result from config loading.
 pub type Result<T> = std::result::Result<T, Error>;
 
 /// Errors occurring during config loading.
 #[derive(Debug)]
 pub enum Error {
-    /// Config file not found.
+    /// UiConfig file not found.
```
chrisduerr, line 41: Search/Replace error or was this intentional?

alacritty/src/config/mod.rs
```diff
@@ -72,7 +70,7 @@ impl Display for Error {
                 write!(f, "Unable to read $HOME environment variable: {}", err)
             },
             Error::Io(err) => write!(f, "Error reading config file: {}", err),
-            Error::Yaml(err) => write!(f, "Config error: {}", err),
+            Error::Yaml(err) => write!(f, "UiConfig error: {}", err),
```
chrisduerr, line 73: Search/Replace error or was this intentional?

alacritty/src/config/mod.rs
```diff
@@ -100,21 +98,21 @@ impl From<serde_yaml::Error> for Error {
...
 
 /// Load the configuration file.
-pub fn load(options: &Options) -> Config {
+pub fn load(options: &Options) -> UiConfig {
     let config_options = options.config_options.clone();
     let config_path = options.config_file.clone().or_else(installed_config);
 
     // Load the config using the following fallback behavior:
-    //  - Config path + CLI overrides
+    //  - UiConfig path + CLI overrides
```
chrisduerr, line 106: Search/Replace error or was this intentional?

alacritty/src/config/mod.rs
```diff
@@ -241,7 +239,7 @@ fn load_imports(config: &Value, config_paths: &mut Vec<PathBuf>, recursion_limit
         }
 
         if !path.exists() {
-            info!(target: LOG_TARGET_CONFIG, "Config import not found:\n  {:?}", path.display());
+            info!(target: LOG_TARGET_CONFIG, "UiConfig import not found:\n  {:?}", path.display());
```
chrisduerr, line 242: Search/Replace error or was this intentional?

alacritty/src/event.rs
```diff
@@ -35,10 +33,10 @@ use alacritty_terminal::term::{ClipboardType, SizeInfo, Term, TermMode};
 #[cfg(not(windows))]
 use alacritty_terminal::tty;
 
-use crate::cli::Options as CliOptions;
+use crate::cli::{Options as CliOptions, TerminalOptions as TerminalCLIOptions};
```
chrisduerr, line 36:
```suggestion
use crate::cli::{Options as CliOptions, TerminalOptions as TerminalCliOptions};
```

Correct capitalization in Rust is camelcase even for acronyms.

alacritty/src/window_context.rs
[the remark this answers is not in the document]
chrisduerr: I'm still not sure about this, passing `Option<TerminalOptions>` around, especially to WindowContext when it always should be defined as *something* seems strange.

I'm writing this comment after finding that the `CreateWindow` event takes an Option, but I don't think it should.

alacritty/src/ipc.rs
```diff
@@ -40,13 +37,31 @@ pub fn spawn_ipc_socket(options: &Options, event_proxy: EventLoopProxy<Event>) -
...
     thread::spawn_named("socket listener", move || {
-        // Accept up to 2 bytes to ensure only one byte is received.
-        // This ensures forward-compatibility.
-        let mut buf = [0; 2];
-
-        while let Ok(received) = socket.recv(&mut buf) {
-            if buf[..received] == SOCKET_MESSAGE_CREATE_WINDOW {
-                let _ = event_proxy.send_event(Event::new(EventType::CreateWindow, None));
+        let mut data = String::new();
+        for stream in listener.incoming().filter_map(|stream| stream.ok()) {
```
chrisduerr, line 41:
```suggestion
        for stream in listener.incoming().filter_map(Result::ok) {
```

50% shorter!

alacritty/src/ipc.rs
```diff
@@ -40,13 +37,31 @@ pub fn spawn_ipc_socket(options: &Options, event_proxy: EventLoopProxy<Event>) -
...
+            match stream.read_line(&mut data) {
+                Ok(0) | Err(_) => continue,
+                Ok(_) => (),
+            };
+
+            // Read pending events on socket.
+            let message: SocketMessage = match serde_json::from_str(&data) {
+                Ok(message) => message,
+                Err(err) => {
+                    warn!("Fail to convert data from socket: {}", err);
```
chrisduerr, line 54:
```suggestion
                    warn!("Failed to convert data from socket: {}", err);
```

alacritty/src/main.rs
```diff
@@ -217,13 +215,13 @@ fn alacritty(options: Options) -> Result<(), String> {
...
-fn log_config_path(config: &Config) {
-    if config.ui_config.config_paths.is_empty() {
+fn log_config_path(config: &UiConfig) {
+    if config.config_paths.is_empty() {
         return;
     }
 
-    let mut msg = String::from("Configuration files loaded from:");
-    for path in &config.ui_config.config_paths {
+    let mut msg = String::from("UiConfiguration files loaded from:");
```
chrisduerr, line 223: Search & Replace screw-up?

alacritty_terminal/src/config/mod.rs
```diff
@@ -15,37 +15,43 @@ pub use crate::config::scrolling::Scrolling;
...
     /// Remain open after child process exits.
     #[config(skip)]
     pub hold: bool,
 }
 
+impl PtyConfig {
+    pub fn new() -> Self {
+        Default::default()
+    }
+}
```
chrisduerr, line 53: If you implement `From`/`Into` for `TerminalOptions` -> `PtyConfig` you probably don't need `PtyConfig::new` anymore.

alacritty/src/window_context.rs
```diff
@@ -81,15 +84,26 @@ impl WindowContext {
...
         // pseudoterminal. A file descriptor for the master side is retained for
         // reading/writing to the shell.
-        let pty = tty::new(config, &display.size_info, display.window.x11_window_id());
+        let mut new_pty_config = PtyConfig::new();
+        let pty_config = match options {
+            Some(mut options) => {
+                new_pty_config.working_directory = options.working_directory.take();
+                new_pty_config.shell = options.command();
+                new_pty_config.hold = options.hold;
+                &new_pty_config
```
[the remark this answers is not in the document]
chrisduerr: Bump.

chrisduerr (changes requested):
Turns out "tomorrow" took a tad longer than I thought. Better late than never I suppose.

chrisduerr:
Before I forget: Since this changes Alacritty's CLI, we'll also need to adjust our completions and manpage(s).

alacritty/src/window_context.rs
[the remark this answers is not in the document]
kchibisov: The reason behind is to not overwrite config when we shouldn't, like in a case where default options was `hold=true` and then we pass `hold=false` to disable holding for specific windows. However one could make all fields optional, which is ugly.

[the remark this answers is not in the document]
chrisduerr: The fields are always initialized to *something* though. So I'm just questioning whether this location is the right place for this logic.

alacritty_terminal/src/config/mod.rs
chrisduerr: This is technically unused now, just no warning because it's a public API.

Do you want to keep this around just to make the shorter `::new` available in our public API and make it easier to understand how this can be initialized when looking at rustdoc, or should we remove this?

chrisduerr (changes requested):
Looks pretty good now I'd say.

Only big thing missing are the updates to completion and manpages.

alacritty_terminal/src/config/mod.rs
kchibisov: I just like to have `new` constructor in general for such types. We can remove it, but having I don't see a reason why.

kchibisov:
>Only big thing missing are the updates to completion and manpages.

I've updated man pages, but as for completions I'm not sure how to update them, since docs are pretty unclear (and I'm not good with shell scripting).

chrisduerr:
> I've updated man pages, but as for completions I'm not sure how to update them, since docs are pretty unclear (and I'm not good with shell scripting).

Fair point, one person having to waste their time on learning how to write completions for three different shells is probably enough. I'll take a look.

chrisduerr:
I've written most of the completions by hand but then decided that it's probably not worth spending hours on that every time. So I've opened https://github.com/alacritty/alacritty/pull/5619 to fix this. After that this PR can be updated.

alacritty/src/window_context.rs
```diff
@@ -81,15 +83,24 @@ impl WindowContext {
...
+        let new_pty_config;
+        let pty_config = match options {
+            Some(options) => {
+                new_pty_config = options.into();
+                &new_pty_config
+            },
+            None => &config.terminal_config.pty_config,
+        };
+
+        let pty = tty::new(pty_config, &display.size_info, display.window.x11_window_id())?;
```
chrisduerr, line 103: After the completions PR, this should be pretty much good to go. Only thing left would be my suggestion to maybe try this code with map/unwrap_or?

Seems like you've got the whole `new_pty_config` temporary variable to work around borrowing issues. Sounds a lot like a `Cow` usecase to me?

chrisduerr (changes requested):

alacritty/src/window_context.rs
kchibisov: It's not about borrowing, it's more about picking right reference to variable, not sure how `Cow` (copy on write) relates to that? I mean, I could in theory store Cow then clone it and then `if Some(..) = options` update its variable, but I don't like that approach.

chrisduerr: It seems like you're declaring the `new_pty_config` variable to extend the lifetime of the `options.into()` since you expect to store a reference because you can't move out of `config.terminal_config`. If you instead return a `Cow` from the match which stores `Owned` in the `Some` case and `Borrowed` in the `None` case you'll avoid having to declare the `new_pty_config` environment variable for creating a temporary borrowed value.

extra/alacritty-msg.man
```diff
@@ -14,6 +14,16 @@ Path for IPC socket creation
...
+.SH "CREATE-WINDOW OPTIONS"
+.TP
+\fB\-\-hold\fR
+Remain open after child process exits
+.TP
+\fB\-\-working\-directory\fR <working\-directory>
+Start the shell in the specified working directory
+.TP
+\fB\-e\fR, \fB\-\-command\fR <command>...
+Command and args to execute (must be last argument)
```
chrisduerr, line 26: Currently this is indented on the toplevel, how do you feel about indenting this below the `create-window` message and just calling it `OPTIONS`.

Should make it a little easier to understand without having to cross-reference the title.

Also the `SYNOPSIS` might need to be changed to document that after `[MESSAGES]` there's also another `[OPTIONS]`, or maybe `[MESSAGE_OPTIONS]`. Also I'd say `[MESSAGES]` should probably be `[MESSAGE]` because there can only be one.

chrisduerr (changes requested):

extra/alacritty-msg.man
```diff
@@ -2,28 +2,27 @@
 .SH NAME
 alacritty-msg \- Send messages to Alacritty
 .SH "SYNOPSIS"
-alacritty msg [OPTIONS] [MESSAGES]
+alacritty msg [OPTIONS] [MESSAGE] [MESSAGE_OPTIONS]
```
chrisduerr, line 5:
```suggestion
alacritty msg [OPTIONS] <MESSAGE> [MESSAGE_OPTIONS]
```

This has absolutely nothing to do with your changes, but thinking about the usage of `alacritty msg` I think `[MESSAGE]` should be `<MESSAGE>` instead. Since it's not optional; we always want *some* message sent.

chrisduerr (changes requested):
Looking great now, just one minor thing I noticed and we should be good to go with this.

extra/alacritty-msg.man
```diff
@@ -2,18 +2,27 @@
...
 .SH "OPTIONS"
-.TP
 \fB\-s\fR, \fB\-\-socket\fR <socket>
 Path for IPC socket creation
 .SH "MESSAGES"
-.TP
 \fBcreate-window\fR
 Create a new window in the same Alacritty process
+.TP
+.SH "OPTIONS"
```
chrisduerr, line 16: Actually one more thing: This is on the same level as `create-window` at the moment. I think it would be better to indent it one more level? That way nobody thinks that `OPTIONS` is a valid message to send?

chrisduerr (approved):
LGTM

PR #5581, opened 2021-11-01
Updating hit processing to return None when empty.

### LINKED ISSUE #5492: Hint post processing crash

gsfreema:
Fixes #5492 

Looping through processed matches for a result.
This is necessary since returning None stops an iterator.

### DIFF
--- a/alacritty/src/display/hint.rs
+++ b/alacritty/src/display/hint.rs
@@ -315,7 +315,7 @@ impl<'a, T> HintPostProcessor<'a, T> {
         };
 
         // Post-process the first hint match.
-        post_processor.iterate_to_next_processed_match(regex_match);
+        post_processor.find_next_match(regex_match);
 
         post_processor
     }
@@ -388,24 +388,19 @@ impl<'a, T> HintPostProcessor<'a, T> {
         }
     }
 
-    /// Loops over processed matches until a match is found or the
-    /// range is exhausted.
-    fn iterate_to_next_processed_match(&mut self, mut regex_match: Match) {
-        self.next_match = None;
+    fn find_next_match(&mut self, regex_match: Match) {
+        self.next_match = self.hint_post_processing(&regex_match);
 
-        while regex_match.start() <= regex_match.end() {
-            self.next_match = self.hint_post_processing(&regex_match);
-
-            regex_match = match self.next_match {
-                Some(_) => break,
-                None => regex_match.start().add(self.term, Boundary::Grid, 1)..=*regex_match.end(),
-            }
+        if self.next_match.is_none() {
+            self.next_match =
+                self.term.regex_search_right(self.regex, *regex_match.start(), *regex_match.end())
         }
 
-        self.start = self
-            .next_match
-            .as_ref()
-            .map_or(*regex_match.start(), |nm| nm.end().add(self.term, Boundary::Grid, 1));
+        self.start = self.next_match.as_ref().map_or(regex_match.end(), |nm| nm.end()).add(
+            self.term,
+            Boundary::Grid,
+            1,
+        );
     }
 }
 
@@ -417,7 +412,7 @@ impl<'a, T> Iterator for HintPostProcessor<'a, T> {
 
         if self.start <= self.end {
             if let Some(rm) = self.term.regex_search_right(self.regex, self.start, self.end) {
-                self.iterate_to_next_processed_match(rm);
+                self.find_next_match(rm);
             }
         }
 
@@ -482,9 +477,9 @@ mod tests {
             &search,
             Point::new(Line(0), Column(1))..=Point::new(Line(0), Column(1)),
         )
-        .take(1)
+        .take(2)
         .count();
 
-        assert_eq!(count, 0);
+        assert_eq!(count, 1);
     }
 }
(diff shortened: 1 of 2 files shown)


alacritty/src/display/hint.rs
```diff
@@ -378,7 +381,31 @@ impl<'a, T> HintPostProcessor<'a, T> {
...
+            regex_match = match self.next_match {
+                Some(_) => break,
+                None => regex_match.start().add(self.term, Boundary::Grid, 1)..=*regex_match.end(),
+            }
+        }
+
+        self.start = self
+            .next_match
+            .as_ref()
+            .map_or(*regex_match.start(), |nm| nm.end().add(self.term, Boundary::Grid, 1));
```
chrisduerr, line 408: Why do you use `regex_match.start()` as fallback here? Because this means that we've mutated `regex_start` to actually be bigger than `regex_match.end()` and we now have to use the end instead of the start to determine the farthest point? Seems a bit confusing.

chrisduerr (changes requested):
Thanks for the update. The logic looks like it *should* behave as I'd intuitively expect now, though I haven't tested it myself.

alacritty/src/display/hint.rs
```diff
@@ -378,7 +381,31 @@ impl<'a, T> HintPostProcessor<'a, T> {
...
+    /// range is exhausted.
+    fn iterate_to_next_processed_match(&mut self, mut regex_match: Match) {
+        self.next_match = None;
+
+        while regex_match.start() <= regex_match.end() {
+            self.next_match = self.hint_post_processing(&regex_match);
+
+            regex_match = match self.next_match {
+                Some(_) => break,
+                None => regex_match.start().add(self.term, Boundary::Grid, 1)..=*regex_match.end(),
```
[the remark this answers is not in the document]
gsfreema: Are you referring to mutating the start of regex_match? As far as I can tell, there isn't a function to mutate start and end and they are both marked as pub(crate).

alacritty/src/display/hint.rs
gsfreema: I don't believe I understand the problem with using `regex_match.start()` as the fallback. Isn't it expected that `start` eventually moves past `end` as part of the Iterator implementation?

alacritty/src/display/hint.rs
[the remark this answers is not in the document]
chrisduerr: My bad, I always forget that's only available on `Range`, not on `RangeInclusive`.

alacritty/src/display/hint.rs
chrisduerr: The problem is that it relies heavily on the other code putting you in a state that you don't really want to be in to begin with.

alacritty/src/display/hint.rs
```diff
@@ -378,7 +381,31 @@ impl<'a, T> HintPostProcessor<'a, T> {
...
+            }
```
chrisduerr, line 402: I'm not sure if this contradicts something I've said previously, but I don't think overlapping hint matches should be post-processed multiple times. The `regex_search_right` method is already responsible for handling potential overlaps.

So really if `hint_post_processing` doesn't return a match, the next match must be determined by regex search not by post processing.

chrisduerr (changes requested):

alacritty/src/display/hint.rs
```diff
@@ -306,13 +306,16 @@ struct HintPostProcessor<'a, T> {
...
+            end: *regex_match.end(),
+            term,
+            regex,
+        };
 
         // Post-process the first hint match.
-        let next_match = post_processor.hint_post_processing(&regex_match);
-        post_processor.start = next_match.end().add(term, Boundary::Grid, 1);
-        post_processor.next_match = Some(next_match);
+        post_processor.find_next_match(regex_match);
```
chrisduerr, line 318: The hint post processor is an iterator over all processed matches, so if the post processing results in an empty match, there should not be any other match returned by the post-processor. It's not a regex match iterator.

gsfreema: OK, if I'm reading this correctly, then if post-processing returns None, there shouldn't be any other results from this iterator. That has me thinking that the other comment of
> The next match would still need to be post-processed too.

doesn't really apply because there shouldn't be anymore results since post-processing resulted in `None`.

chrisduerr: The post-processing resulted in `None`, but when you call the post-processor from inside the regex match iterator you still have to go to the next match and post-process that. You can't just stop without any further matches just because your first match is empty.

Now you might not need to do anything to achieve this, but you can't ever just call `regex_search_right` directly without post-processing it first since every match returned by `regex_search_right` might itself contain multiple other matches.

alacritty/src/display/hint.rs
```diff
@@ -378,7 +381,31 @@ impl<'a, T> HintPostProcessor<'a, T> {
...
+            }
+
+            self.start = regex_match.start().add(self.term, Boundary::Grid, 1);
+            if self.start > self.end {
+                return;
+            }
+
+            match self.term.regex_search_right(self.regex, self.start, self.end) {
+                Some(rm) => regex_match = rm,
+                None => return,
```
chrisduerr, line 406: By using an early return here instead of a `break`, we don't need to worry about the state of `self.start`, since `self.next_match` is none and thus `self.next()` will abort immediately before ever doing anything with `self.start`.

chrisduerr (approved):
I've made some minor style changes myself and added a changelog entry. I think things should be good now.

PR #5582, opened 2021-11-02
Add font-size presets and bindings

djpohly:
This patch adds a "font.preset_size" item to the configuration file. This item holds an array of preset font sizes which can be selected with the new FontSize1, FontSize2, and FontSize3 commands.  Bindings for these commands default to Ctrl(/Cmd)-1/2/3.

At the moment, this is only a proof of concept to demonstrate a feature I would find useful - more an idea than a real request.  Ideally, the preset sizes would be placed directly in the bindings, obviating the extra array. This would require a parameterized action, but it was not obvious how to set one up (if it is currently possible at all), hence the much-less-than-ideal design.

Any ideas how this could be implemented in a more robust fashion, or if it's a feature that would be wanted at all? I'm not very experienced with Rust, so go easy on me. :)

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -158,6 +158,12 @@
   # Point size
   #size: 11.0
 
+  # Predefined alternative sizes
+  #preset_size
+  #  - 14.0
+  #  - 18.0
+  #  - 24.0
+
   # Offset is the extra space around each character. `offset.y` can be thought
   # of as modifying the line spacing, and `offset.x` as modifying the letter
   # spacing.
@@ -579,6 +585,9 @@
 #   - IncreaseFontSize
 #   - DecreaseFontSize
 #   - ResetFontSize
+#   - FontSize1
+#   - FontSize2
+#   - FontSize3
 #   - ScrollPageUp
 #   - ScrollPageDown
 #   - ScrollHalfPageUp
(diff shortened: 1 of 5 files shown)


chrisduerr:
> if it's a feature that would be wanted at all?

Not interested in adding this to Alacritty.

Parameterized actions/Generically overriding specific config options for a single option at runtime might be things desired in general, but really that's far from anything this PR does and more complicated.

djpohly:
Okay, thanks for the feedback.

lefuturiste:
Hello, sorry to speak here, but I'm interested in such a feature or similar, the goal for me is being able to quickly change font-sizes of terminal without manually hitting many times the Zoom more or Zoom less action. I found this concept to be useful. 
Is there any way to do this (a particular config) as a workaround in the current state of alacritty? Thanks for your work.

kchibisov:
`alacritty msg config font.size=20` and to reset `alacritty msg config -r`.

lefuturiste:
> `alacritty msg config font.size=20` and to reset `alacritty msg config -r`.

Not working
```
Ignoring invalid CLI option 'font.size': TOML parse error at line 1, column 10
  |
1 | font.size
  |          ^
expected `.`, `=`

Ignoring invalid CLI option '20': TOML parse error at line 1, column 3
  |
1 | 20
  |   ^
expected `.`, `=`
```

PR #5599, opened 2021-11-09
Handle option as meta key on macos

theryangeary:
related to #62 - provides support for handling the Mac option key as the meta key, sending the `\x1b` control sequence before the key. It looks like a lot of people have taken to manually listing _every_ alt+key combo they might use, it might be helpful to just automatically support that in the way it works out of the box on Linux.

The character resolution code is.... ugly. Not sure if there is a better way to handle that, would be happy to hear input.

### DIFF
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -13,6 +13,7 @@ use std::time::{Duration, Instant};
 use glutin::dpi::PhysicalPosition;
 use glutin::event::{
     ElementState, KeyboardInput, ModifiersState, MouseButton, MouseScrollDelta, TouchPhase,
+    VirtualKeyCode,
 };
 use glutin::event_loop::EventLoopWindowTarget;
 #[cfg(target_os = "macos")]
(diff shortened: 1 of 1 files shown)


chrisduerr:
> The character resolution code is.... ugly. Not sure if there is a better way to handle that, would be happy to hear input.

There is, upstream. Where this issue has already been reported and people have worked on it before.

theryangeary:
So if this has already been worked on upstream, how do we make use of it in alacrity? As far as I can find the only solution thus far is to remap every key in the config file.

chrisduerr:
It's been worked on, but nothing was ever merged.

kevinsjoberg:
Then why was this closed? Not interested in supporting it? Bad implementation? Something else?

DavSanchez:
Without entering in a discussion about macOS support in Alacritty I want to ask, if this was merged, would this change affect both Alt keys in the keyboard? I have a non-english keyboard layout and I would still need access to some special character combinations. Is there any way for this to affect only the left Alt key for example?

Thanks

PR #5607, opened 2021-11-13
Fix crash when clears the viewport and the vi cursor on scrollback

### LINKED ISSUE #5544: Alacritty crashes when in vi mode and terminal clears
Alacritty panics when receiving any clear ansi escape code (tested with `^[[2J` and `^[c`) when in vi mode near the top of a long buffer with the following message:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', alacritty/src/display/content.rs:66:85
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Steps to reproduce
```shell
$ yes # output a few thousand lines of output (seems to only happen with large scrollback buffer)
$ sleep 5; clear # Delay clearing long enough to enter vi mode and jump to top of scrollback
# ctrl+space, g
# wait for clear to execute
```

### System

OS: Linux 5.14.11-arch1-1
Version: alacritty 0.9.0 (fed349aa), alacritty 0.10.0-dev (f90dd12e)
Linux/BSD: Swaywm

### Logs

```
[2021-10-14 17:39:13.785011895] [INFO ] [alacritty] Welcome to Alacritty
[2021-10-14 17:39:13.785059625] [INFO ] [alacritty] Configuration files loaded from:
                                                      "/home/alex/.config/alacritty/alacritty.yml"
[2021-10-14 17:39:13.790835891] [DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Book, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[2021-10-14 17:39:13.791264364] [DEBUG] [alacritty] Estimated DPR: 1
[2021-10-14 17:39:13.791274062] [DEBUG] [alacritty] Estimated window size: None
[2021-10-14 17:39:13.791277589] [DEBUG] [alacritty] Estimated cell size: 9 x 18
[2021-10-14 17:39:13.811619986] [INFO ] [alacritty] Device pixel ratio: 1
[2021-10-14 17:39:13.814423501] [INFO ] [alacritty] Initializing glyph cache...
[2021-10-14 17:39:13.814771643] [DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Book, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[2021-10-14 17:39:13.815199295] [DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Bold, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[2021-10-14 17:39:13.815601028] [DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Oblique, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[2021-10-14 17:39:13.816009643] [DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Bold Oblique, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[2021-10-14 17:39:13.819742710] [INFO ] [alacritty] ... finished initializing glyph cache in 0.005309871s
[2021-10-14 17:39:13.819753019] [INFO ] [alacritty] Cell size: 9 x 18
[2021-10-14 17:39:13.819757918] [INFO ] [alacritty] Padding: 0 x 0
[2021-10-14 17:39:13.819762146] [INFO ] [alacritty] Width: 800, Height: 600
[2021-10-14 17:39:13.819959466] [INFO ] [alacritty] PTY dimensions: 33 x 88
[2021-10-14 17:39:13.821683588] [INFO ] [alacritty] Initialisation complete
[2021-10-14 17:39:13.831748993] [DEBUG] [alacritty_terminal] New num_cols is 210 and num_lines is 28
[2021-10-14 17:39:13.831882413] [INFO ] [alacritty] Padding: 0 x 0
[2021-10-14 17:39:13.831894346] [INFO ] [alacritty] Width: 1897, Height: 511
[2021-10-14 17:39:13.851039870] [DEBUG] [crossfont] Loaded Face Face { ft_face: Font Face: Regular, load_flags: TARGET_LIGHT, render_mode: "Normal", lcd_filter: 1 }
[2021-10-14 17:39:25.154252793] [DEBUG] [alacritty_terminal] Term::resize dimensions unchanged
[2021-10-14 17:39:25.154300102] [INFO ] [alacritty] Padding: 0 x 0
[2021-10-14 17:39:25.154312765] [INFO ] [alacritty] Width: 1897, Height: 511
thread 'main' panicked at 'index out of bounds: the len is 28 but the index is 9999', /build/alacritty/src/alacritty/alacritty_terminal/src/term/search.rs:422:16
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

a5ob7r:
Fixes #5544.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1423,6 +1423,11 @@ impl<T: EventListener> Handler for Term<T> {
                 if self.mode.contains(TermMode::ALT_SCREEN) {
                     self.grid.reset_region(..);
                 } else {
+                    // Change the absolute vi cursor position into inside the viewport, but
+                    // preserve the relative one in the display area.
+                    let line = self.vi_mode_cursor.point.line + self.grid.display_offset();
+                    self.vi_mode_cursor.point.line = line.grid_clamp(self, Boundary::Cursor);
+
                     self.grid.clear_viewport();
                 }
 
@@ -2116,6 +2121,36 @@ mod tests {
         term.input('a');
 
         assert_eq!(term.grid()[cursor].c, '▒');
+    }
+
+    #[test]
+    fn clear_viewport_updates_vi_mode_cursor_pos() {
+        let size = SizeInfo::new(10.0, 20.0, 1.0, 1.0, 0.0, 0.0, false);
+        let mut term = Term::new(&Config::default(), size, ());
+
+        // Create 10 lines of scrollback.
+        for _ in 0..29 {
+            term.newline();
+        }
+
+        // Emulate vi cursor motions on default configurations.
+        // gg
+        term.scroll_display(Scroll::Top);
+        term.vi_mode_cursor = term.vi_mode_cursor.scroll(&term, 10);
+        // jjjjj
+        for _ in 0..5 {
+            term.vi_mode_cursor = term.vi_mode_cursor.motion(&mut term, ViMotion::Down);
+        }
+        // lll
+        for _ in 0..3 {
+            term.vi_mode_cursor = term.vi_mode_cursor.motion(&mut term, ViMotion::Right);
+        }
+
+        // Clear the viewport.
+        term.clear_screen(ansi::ClearMode::All);
+
+        assert_eq!(term.vi_mode_cursor.point.line.0, 5);
+        assert_eq!(term.vi_mode_cursor.point.column.0, 3);
     }
 
     #[test]
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1423,21 +1423,19 @@ impl<T: EventListener> Handler for Term<T> {
                 if self.mode.contains(TermMode::ALT_SCREEN) {
                     self.grid.reset_region(..);
                 } else {
-                    // Change the absolute vi cursor position into inside the viewport, but
-                    // preserve the relative one in the display area.
-                    let line = self.vi_mode_cursor.point.line + self.grid.display_offset();
-                    self.vi_mode_cursor.point.line = line.grid_clamp(self, Boundary::Cursor);
-
                     self.grid.clear_viewport();
+
+                    self.vi_mode_cursor.point.line =
+                        self.vi_mode_cursor.point.line.grid_clamp(self, Boundary::Cursor);
                 }
 
                 self.selection = None;
             },
             ansi::ClearMode::Saved if self.history_size() > 0 => {
-                let line = self.vi_mode_cursor.point.line + self.grid.display_offset();
-                self.vi_mode_cursor.point.line = line.grid_clamp(self, Boundary::Cursor);
-
                 self.grid.clear_history();
+
+                self.vi_mode_cursor.point.line =
+                    self.vi_mode_cursor.point.line.grid_clamp(self, Boundary::Cursor);
 
                 self.selection = self.selection.take().filter(|s| !s.intersects_range(..Line(0)));
             },
@@ -2127,7 +2125,7 @@ mod tests {
     }
 
     #[test]
-    fn clear_viewport_updates_vi_mode_cursor_pos() {
+    fn clear_viewport_set_vi_cursor_into_viewport() {
         let size = SizeInfo::new(10.0, 20.0, 1.0, 1.0, 0.0, 0.0, false);
         let mut term = Term::new(&Config::default(), size, ());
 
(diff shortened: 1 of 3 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -2118,6 +2123,36 @@ mod tests {
...
+        term.scroll_display(Scroll::Top);
+        term.vi_mode_cursor = term.vi_mode_cursor.scroll(&term, 10);
+        // jjjjj
+        for _ in 0..5 {
+            term.vi_mode_cursor = term.vi_mode_cursor.motion(&mut term, ViMotion::Down);
+        }
+        // lll
+        for _ in 0..3 {
+            term.vi_mode_cursor = term.vi_mode_cursor.motion(&mut term, ViMotion::Right);
+        }
```
chrisduerr, line 2147:
```suggestion
        term.vi_mode_cursor.point = Point::new(term.topmost_line() + 5, 3);
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -2118,6 +2123,36 @@ mod tests {
...
+        let size = SizeInfo::new(10.0, 20.0, 1.0, 1.0, 0.0, 0.0, false);
+        let mut term = Term::new(&MockConfig::default(), size, ());
+
+        // Create 10 lines of scrollback.
+        for _ in 0..29 {
+            term.newline();
+        }
+
+        // Emulate vi cursor motions on default configurations.
+        // gg
```
chrisduerr, line 2137:
```suggestion
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -2118,6 +2123,36 @@ mod tests {
...
+        for _ in 0..5 {
+            term.vi_mode_cursor = term.vi_mode_cursor.motion(&mut term, ViMotion::Down);
+        }
+        // lll
+        for _ in 0..3 {
+            term.vi_mode_cursor = term.vi_mode_cursor.motion(&mut term, ViMotion::Right);
+        }
+
+        // Clear the viewport.
+        term.clear_screen(ansi::ClearMode::All);
```
chrisduerr, line 2150:
```suggestion
        // Clear the viewport and scrollback.
        term.clear_screen(ansi::ClearMode::All);
```

Kinda the point that we're not just clearing the viewport.

chrisduerr (changes requested):
I'd be curious to know why you've decided to go for keeping the vi mode cursor in the same viewport-relative position, rather than clamping it to the top of the screen.

a5ob7r:
It's not my favorite that force cursor relative movements in the display area caused by buffer scroll. So the reason is so personal. I honestly don't mind either way, so I can switch to another way if needed. But I prefer the current way if I had to choose one of the two ways.

alacritty_terminal/src/term/mod.rs
a5ob7r: Really? The mode name, `ansi::ClearMode::All` , looks like to clear both. But it only clears the viewport according to the code. In fact, `printf '\033[2J'` clears only the viewport but not the scrollback buffer on Alacritty. Is this my misunderstanding?

a5ob7r: This situation is caused by `clear -x` , which returns `\033[H\033[2J` . On Alacritty, this changes the cursor position to (1, 1) and clears only the viewport. And, `clear` returns `\033[H\033[2J\033[3J` , which also clears the scrollback buffer, in addition to the previously mentioned behavior.

a5ob7r:
After thinking again, I also think that another way, which causes vi mode cursor clamping to the upper side of the display area, may make more sense because the contents at the display area are only near the upper side if we clear the display area. But this relies on the `clear` command which returns `\033[H` in addition to `\033[2J` . If someone uses other sequences like `\033[30H` instead of `\033[H`, maybe the contents at the display aren't on the upper side of the display area.

a5ob7r:
This is likely a very rare case and overthinking. :smile:

alacritty_terminal/src/term/mod.rs
chrisduerr: Oh yeah, I'm not sure what I was thinking. We explicitly have `ClearMode::Saved` for the scrollback and the `clear_viewport` method even clearly states that it's only clearing the viewport.

I was just caught up on the bug I suppose: The problem only occurs when the vi mode cursor is in the scrollback and that is "cut off". So I'm not sure your bugfix solves the problem when **only** the saved lines (as in history) is deleted while scrolled up?

chrisduerr (changes requested):
> It's not my favorite that force cursor relative movements in the display area caused by buffer scroll. So the reason is so personal. I honestly don't mind either way, so I can switch to another way if needed. But I prefer the current way if I had to choose one of the two ways.

Just to be clear, I was mainly looking for your argument as to *why* you prefer this, not saying it's a wrong approach. I don't think there's any perfect solution so it's good to consider the benefits of both approaches.

One thing to consider might be if the display is offset, but the cursor is at the far bottom so even after nuking the entire history the old cursor position is still visible. I'm not sure how your current solution behaves in that instance, but I think it would make sense for the cursor to stay on the content in that case, rather than staying in the same relative viewport position.

a5ob7r:
Sorry for my so slow response.

a5ob7r:
> One thing to consider might be if the display is offset, but the cursor is at the far bottom so even after nuking the entire history the old cursor position is still visible. I'm not sure how your current solution behaves in that instance, but I think it would make sense for the cursor to stay on the content in that case, rather than staying in the same relative viewport position.

Oh. It's my missing insight. I also think that the behavior makes sense. It may cause not intuitive or not consistent behavior, which depends on whether the vi cursor is near the topmost viewport and the display offset value if we accept both my suggestion and the behavior. On the other hand, simple vi cursor clamping to the viewport doesn't cause the problem and achieves the behavior automatically. So I'll choose the latter.

alacritty_terminal/src/term/mod.rs
a5ob7r: It's a good point. This problem is caused when clears the scrollback buffer, but it slipped in my mind. I'll fix it later.

chrisduerr:
> On the other hand, simple vi cursor clamping to the viewport doesn't cause the problem and achieves the behavior automatically. So I'll choose the latter.

Seems simpler and realistically this is somewhat of an edgecase anyway. So people probably won't even notice one way or another.

alacritty_terminal/src/term/mod.rs
a5ob7r: It causes a freeze, not a crash. This can be fixed by the same way.

a5ob7r:
I have done to fix some problems above and to rebase this branch with origin/master.

chrisduerr (approved):
Good changelog entries, simple fix, all tested. Unfortunately I have nothing to complain about :smile:.

PR #5609, opened 2021-11-14
Update signal-hook to 0.3

jamessan:
This adds an additional dependency on signal-hook-mio, since signal-hook has split it into a separate crate.

---
My impetus for creating this PR is to aid in Debian's packaging of alacritty.  We've recently bumped from futures 0.1 to 0.3 which, bubbling up through various dependencies, requires updating signal-hook to a newer version.

### DIFF
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -29,7 +29,8 @@ dirs = "3.0.1"
 
 [target.'cfg(unix)'.dependencies]
 nix = "0.22.0"
-signal-hook = { version = "0.1", features = ["mio-support"] }
+signal-hook = "0.3.10"
+signal-hook-mio = { version = "0.2.1", features = ["support-v0_6"] }
 
 [target.'cfg(windows)'.dependencies]
 miow = "0.3"
--- a/alacritty_terminal/src/tty/unix.rs
+++ b/alacritty_terminal/src/tty/unix.rs
@@ -18,8 +18,8 @@ use mio::unix::EventedFd;
 use nix::pty::openpty;
 #[cfg(any(target_os = "linux", target_os = "macos"))]
 use nix::sys::termios::{self, InputFlags, SetArg};
-use signal_hook as sighook;
-use signal_hook::iterator::Signals;
+use signal_hook::consts as sigconsts;
+use signal_hook_mio::v0_6::Signals;
 
 use crate::config::{Config, Program};
 use crate::event::OnResize;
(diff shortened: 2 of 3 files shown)


alacritty_terminal/Cargo.toml
```diff
@@ -29,7 +29,8 @@ dirs = "3.0.1"
 
 [target.'cfg(unix)'.dependencies]
 nix = "0.22.0"
-signal-hook = { version = "0.1", features = ["mio-support"] }
+signal-hook = "0.3"
+signal-hook-mio = { version = "0.2", features = ["support-v0_6"] }
```
chrisduerr, line 33: Please specify the full latest versions (including fixup) of these crates unless you've explicitly verified that 0.3.0 and 0.2.0 work too (in which case this should be signified by explicitly specifying that.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -328,7 +329,7 @@ impl EventedPty for Pty {
     #[inline]
     fn next_child_event(&mut self) -> Option<ChildEvent> {
         self.signals.pending().next().and_then(|signal| {
-            if signal != sighook::SIGCHLD {
+            if signal != sighook::consts::SIGCHLD {
```
chrisduerr, line 332: Might be overly pedantic, but having two module import depth here with the root module being a rename of an import seems unnecessarily complicated.

I think I'd prefer `consts::SIGCHLD`, just `SIGCHLD` (assuming we aren't using any signals from libc here), or an alias on the lower level like `sigconsts::SIGCHLD`. I'd say all of these are superior to the current solution and would be acceptable.

chrisduerr (changes requested):
Is the goal to make Alacritty build without any duplicate dependencies? While we had one version in the past I think unfortunately due to mio that's not gonna be possible for a long time.

Though contributions to get us back towards that are always welcome of course.

alacritty_terminal/src/tty/unix.rs
jamessan: There are uses of libc signals, but they're all namespaced (`libc::SIG...`).  I think `sigconsts` aligns nicely with the existing alias of `sighook`.

jamessan:
> Is the goal to make Alacritty build without any duplicate dependencies? While we had one version in the past I think unfortunately due to mio that's not gonna be possible for a long time.

I proposed these changes because they were pretty straight forward (which is also why I stuck with 0.6 mio support) and simplify things on the packaging side.  We do have support for multiple versions of crates being packaged, but try to avoid it when possible.

chrisduerr:
> We do have support for multiple versions of crates being packaged, but try to avoid it when possible.

I think minimizing duplicate dependencies while having *some* duplicates is the most realistic approach to actually ship Rust applications. So that makes sense.

chrisduerr (approved):
Thanks for the PR.

PR #5615, opened 2021-11-18
Support ligatures using allsorts

fee1-dead:
See also https://github.com/alacritty/crossfont/pull/35/. Current PR includes a patch to Cargo.toml using my fork. Will change after the crossfont PR gets merged.

### DIFF
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -24,8 +24,8 @@ tasks:
       cargo test
   - oldstable: |
       cd alacritty
-      rustup toolchain install --profile minimal 1.46.0
-      rustup default 1.46.0
+      rustup toolchain install --profile minimal 1.51.0
+      rustup default 1.51.0
       cargo test
   - clippy: |
       cd alacritty
--- a/.builds/linux.yml
+++ b/.builds/linux.yml
@@ -27,8 +27,8 @@ tasks:
       cargo +nightly fmt -- --check
   - oldstable: |
       cd alacritty
-      rustup toolchain install --profile minimal 1.46.0
-      rustup default 1.46.0
+      rustup toolchain install --profile minimal 1.51.0
+      rustup default 1.51.0
       cargo test
   - clippy: |
       cd alacritty
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
         run: cargo test
       - name: Oldstable
         run: |
-          rustup default 1.46.0
+          rustup default 1.51.0
           cargo test
       - name: Clippy
         run: |
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,7 +42,7 @@ and
 [easy](https://github.com/alacritty/alacritty/issues?q=is%3Aopen+is%3Aissue+label%3A%22D+-+easy%22)
 issues.
 
-Please note that the minimum supported version of Alacritty is Rust 1.46.0. All patches are expected
+Please note that the minimum supported version of Alacritty is Rust 1.51.0. All patches are expected
 to work with the minimum supported version.
 
 Since `alacritty_terminal`'s version always tracks the next release, make sure that the version is
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,22 @@
 version = 3
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+
+[[package]]
 name = "ab_glyph_rasterizer"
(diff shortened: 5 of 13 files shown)


AlexApps99:
Image of it in action:
image

chrisduerr (changes requested):
> From the CI logs it looks like allsorts does not support the MSRV of alacritty. What do you think is the best action here @chrisduerr?

MSRV can be bumped up to the latest stable Rust version available in the Ubuntu repos. If that isn't enough, allsorts is unsuitable.

Why did you pick allsorts over a library that is actually reliably tested?

chrisduerr:
@perfbot

fee1-dead:
> Why did you pick allsorts over a library that is actually reliably tested?

Because allsorts is rust native. It removes the need of a third party FFI crate, so no FFI overhead and it is cross platform. If you would prefer using Harfbuzz, then I can switch over to Harfbuzz, but I am no expert in coretext and uniscribe to be able to easily come up with a windows/mac support.

Would you like me to open another PR that uses harfbuzz to compare performance?

chrisduerr:
> Would you like me to open another PR that uses harfbuzz to compare their performance?

I think as the person proposing this change, you should test the performance implications of different solutions. And I'd like to point out that the previous test only verified potential PTY lock contention issues, while the change itself will most likely impact framerate. So to actually verify the performance of these benchmarks one **must** record the rendering framerate.

> Because allsorts is rust native.

That seems more like an argument for laziness rather than picking the correct solution. As far as I'm aware there's basically nobody using allsorts; it's a new library without a lot of development time behind it that has largely dried up over the past 4 months. So all things considered the argument really should be made why **not** to choose the reliable standard Harfbuzz that has been tweaked wrt API stability and performance for decades.

fee1-dead:
Oops, I retract that comment. I forgot to build with release mode.

chrisduerr:
> EDIT: to be honest there is not much of a difference to me. I couldn't tell the difference between allsorts and harfbuzz and no shaping

Define "tell the difference". You shouldn't have to try and tell anything, benchmarks should always be measured in metrics that can be tracked and compared against each other. So either there is a framerate difference or there isn't. Also of course since vsync might limit the framerate, one should ensure to decrease the font size until Alacritty starts struggling since ultimately that's where this starts to matter.

fee1-dead:
Hello, I'm back and I tried to benchmark using font size 0.1 with `cat /dev/urandom`.

Here are the results (sorry for low quality recording):

https://user-images.githubusercontent.com/43851243/146580346-8fd25145-9284-4b55-9432-0222dfdb80a3.mp4

https://user-images.githubusercontent.com/43851243/146580351-4945610b-cf66-4dfc-aa94-54b8133f0c70.mp4

https://user-images.githubusercontent.com/43851243/146580348-34fba9b2-2762-4969-9d06-cac1ccb38f84.mp4

From human eye I think harfbuzz is better for this benchmark. I am closing this and reopening another that uses the harfbuzz library.

chrisduerr:
That is not a benchmark lol. Calling it a rough estimate would probably still give it too much credit.

fee1-dead:
I still have the binaries. I am not really familiar with measuring performance of terminals. Could you please provide with some instructions or tools to generate data with? Thank you.

chrisduerr:
Benchmarking workloads are part of Alacritty's contribution documentation, measuring framerates depends both on your hardware and software so you'll have to figure that out on your own (though it's likely the more relevant piece of information here).

fee1-dead:
Hello, I'm back. 

# typometer

Screenshot from 2021-12-18 19-41-45

# vtebench

output

# framerate

I used mangohud to observe the framerate of `cat /dev/urandom` on wayland with font size set to 1. The MangoHUD program does have a logging option but it is unavailable under wayland. Running `cat /dev/urandom` on Xorg unfortunately freezes the X server so I had to observe the average via human eye.

Baseline (no shaping, commit before either change): ~17.1 ms, 58 fps on average

image

Shaping with allsorts: ~32.1 ms, 31 fps on average

image

Shaping with harfbuzz: ~23.4 ms, 42 fps on average

image

# binary size

```
ls -al alacritty_* | awk '{ print $(NF)"  "($(NF-4))" bytes" }'
alacritty_allsorts  35448344 bytes
alacritty_baseline  32224528 bytes
alacritty_harfbuzz  32291456 bytes
```

I think the overall assessment is that harfbuzz is superior.

chrisduerr:
That rendering performance seems awful for both harfbuzz and allsorts, considering that the baseline is probably only at 60 because it's limited by vsync. And that's with a trivially slow benchmark like `cat /dev/urandom`.

fee1-dead:
Well, at least we know which library to choose. I'll try to improve performance with harfbuzz.

fee1-dead:
Also, note that `cat /dev/urandom` is probably biased against shaping engines because they output a lot of characters that shaping engines don't recognize. That means fallback paths will be executed more often than a daily use scenario.

Issue #5616, opened 2021-11-18
[Feature request] Pass current working directory to hint commands


carrascomj:
### Problem
I have a hint for `git status`, cargo tracebacks and python tracebacks that macthes the files in the visible text (and the line in the two former cases) and opens an editor on that file.

The problem is that, for instance, `git status` or `cargo test` show paths relative to the working directory:

![image](https://user-images.githubusercontent.com/46683255/142457548-054f6815-711c-47f9-bec7-67b3b4713578.png)

but the PWD is not shared with the child process launched by the `HintAction`, so the command cannot know where the file is to open it.

### Feature request
Allow to pass the current working directory to the `HintCommand`. One way would be to replace a special argument (for instance "PWD") with the working directory of the tty (although I believe it is not the best possible design it is easy to implement).

PR #5617, opened 2021-11-18
Pass current working directory to Commands

carrascomj:
Would implement #5616 

### Description
Commands are spawned setting the current working directory to the one of the parent alacritty instance.

### Implementation
Pass the current working directory of the running alacritty instance to [Command::current_dir](https://doc.rust-lang.org/std/process/struct.Command.html#method.current_dir). Only works on Unix-like systems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Reload configuration files when their symbolic link is replaced
 - Strip trailing whitespaces when yanking from a block selection
 - Display area keeps history position when viewport is cleared
+- Commands spawn from the current directory of the parent alacritty instance in Unix-like systems
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Reload configuration files when their symbolic link is replaced
 - Strip trailing whitespaces when yanking from a block selection
 - Display area keeps history position when viewport is cleared
-- Commands spawn from the current directory of the parent alacritty instance in Unix-like systems
+- Commands spawn from the current directory of the foreground shell in Unix-like systems
 
 ### Fixed
 
--- a/alacritty/src/daemon.rs
+++ b/alacritty/src/daemon.rs
@@ -1,6 +1,10 @@
+#[cfg(not(windows))]
+use alacritty_terminal::tty;
 use std::ffi::OsStr;
 use std::fmt::Debug;
 use std::io;
+#[cfg(not(any(target_os = "macos", windows)))]
+use std::fs;
 #[cfg(not(windows))]
 use std::os::unix::process::CommandExt;
 #[cfg(windows)]
@@ -43,6 +47,26 @@ where
         .map(|_| ())
 }
 
+// get working directory of controlling process
+#[cfg(not(windows))]
+pub fn foreground_process_path() -> io::Result<std::path::PathBuf> {
+    let mut pid = unsafe { libc::tcgetpgrp(tty::master_fd()) };
+    if pid < 0 {
+        pid = tty::child_pid();
+    }
+
+    #[cfg(not(any(target_os = "macos", target_os = "freebsd")))]
+    let link_path = format!("/proc/{}/cwd", pid);
+    #[cfg(target_os = "freebsd")]
+    let link_path = format!("/compat/linux/proc/{}/cwd", pid);
+    #[cfg(not(target_os = "macos"))]
+    let cwd = fs::read_link(link_path);
+    #[cfg(target_os = "macos")]
+    let cwd = macos::proc::cwd(pid);
+
+    cwd
+}
+
 #[cfg(not(windows))]
 fn spawn_daemon<I, S>(program: &str, args: I) -> io::Result<()>
 where
@@ -55,6 +79,7 @@ where
             .stdin(Stdio::null())
             .stdout(Stdio::null())
             .stderr(Stdio::null())
+            .current_dir(foreground_process_path().unwrap_or_default())
             .pre_exec(|| {
                 match libc::fork() {
                     -1 => return Err(io::Error::last_os_error()),
--- a/alacritty/src/daemon.rs
+++ b/alacritty/src/daemon.rs
@@ -1,6 +1,4 @@
 #[cfg(not(windows))]
-use alacritty_terminal::tty;
-#[cfg(not(windows))]
 use std::error::Error;
 use std::ffi::OsStr;
 use std::fmt::Debug;
@@ -17,6 +15,12 @@ use std::process::{Command, Stdio};
 use log::{debug, warn};
 #[cfg(windows)]
 use winapi::um::winbase::{CREATE_NEW_PROCESS_GROUP, CREATE_NO_WINDOW};
+
+#[cfg(not(windows))]
+use alacritty_terminal::tty;
+
+#[cfg(target_os = "macos")]
+use crate::macos;
 
 /// Start the daemon and log error on failure.
 pub fn start_daemon<I, S>(program: &str, args: I)
--- a/alacritty/src/daemon.rs
+++ b/alacritty/src/daemon.rs
@@ -2,9 +2,9 @@
 use alacritty_terminal::tty;
 use std::ffi::OsStr;
 use std::fmt::Debug;
-use std::io;
 #[cfg(not(any(target_os = "macos", windows)))]
 use std::fs;
+use std::io;
 #[cfg(not(windows))]
 use std::os::unix::process::CommandExt;
 #[cfg(windows)]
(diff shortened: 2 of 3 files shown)


chrisduerr:
I believe it has been discussed before that child processes spawned in any way (same for keybindings for example) should be spawned in the current working directory, simply because when it matters that's usually what you want and the location your Alacritty was started in basically never matters.

So I'm open to this change in general, but I think it should be consistent across the board and the implementation should not rely on replacing some random `PWD` argument, the working directory should just be set for the new process and when an application is interested in reading it, they just need to check their cwd.

chrisduerr (changes requested):
See my comment.

carrascomj:
Sorry, I didn´t see that discussion but it makes sense.

It only works in Unix as of now. My excuse is that's the way it was done for [event::spawn_new_instance](https://github.com/alacritty/alacritty/blob/master/alacritty/src/event.rs#L354), but I don't know how to do it on windows nor I have access to a windows machine right now.

alacritty/src/daemon.rs
```diff
@@ -55,6 +79,7 @@ where
             .stdin(Stdio::null())
             .stdout(Stdio::null())
             .stderr(Stdio::null())
+            .current_dir(foreground_process_path().unwrap_or_default())
```
chrisduerr, line 82: This is not a good idea, we should *conditionally* add the current dir, not fallback to an empty path that doesn't exist. Your `unwrap_or_default` here will actually prevent the subprocess from starting at all.

alacritty/src/event.rs
```diff
@@ -354,14 +352,6 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
-        // as working-directory parameter.
-        #[cfg(unix)]
-        let mut args = foreground_process_path()
-            .map(|path| vec!["--working-directory".into(), path])
-            .unwrap_or_default();
-
-        #[cfg(not(unix))]
         let mut args: Vec<PathBuf> = Vec::new();
 
         let working_directory_set = !args.is_empty();
```
chrisduerr, line 357: You're breaking the behavior this boolean is used for by removing the `args` assignment on unix. See the following lines of code here.

chrisduerr (changes requested):
Besides the review comments, this is also missing an entry in the changelog.

alacritty/src/daemon.rs
carrascomj: Right, I thought that `PathBuf::default` pointed to a reasonable directory (which in hindsight does not make sense). I changed that and made `foreground_process_path` return a trait object error (like it is in master), because the Error might not be an `io::Error` for macos.

alacritty/src/event.rs
carrascomj: Thanks, I think it is fixed (simply skipping `--working-directory arg` now).

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `ExpandSelection` is now a configurable mouse binding action
 - Config option `background_opacity`, you should use `window.opacity` instead
 - Reload configuration files when their symbolic link is replaced
+- Commands spawn from the current directory of the parent alacritty instance in Unix-like systems
```
chrisduerr, line 26: This comment isn't accurate. They spawn in the current directory of the foreground shell, not in Alacritty's.

alacritty/src/daemon.rs
```diff
@@ -43,18 +49,44 @@ where
...
+
+    #[cfg(not(any(target_os = "macos", target_os = "freebsd")))]
+    let link_path = format!("/proc/{}/cwd", pid);
+    #[cfg(target_os = "freebsd")]
+    let link_path = format!("/compat/linux/proc/{}/cwd", pid);
+
+    #[cfg(not(target_os = "macos"))]
+    let cwd = fs::read_link(link_path)?;
+
+    // this might not be a io::Result, so a Boxed dyn Error is needed
```
chrisduerr, line 68: You don't need to explain type signatures in your comments.

alacritty/src/daemon.rs
```diff
@@ -43,18 +49,44 @@ where
         .map(|_| ())
 }
 
+// get working directory of controlling process
```
chrisduerr, line 52: Please follow the comment guidelines.

alacritty/src/event.rs
```diff
@@ -354,22 +352,13 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
 
-        let working_directory_set = !args.is_empty();
-
         // Reuse the arguments passed to Alacritty for the new instance.
         while let Some(arg) = env_args.next() {
-            // Drop working directory from existing parameters.
-            if working_directory_set && arg == "--working-directory" {
+            // Drop working directory from existing parameters; start_daemon()
+            // will set it to the parent alacritty instance current directory
+            if arg == "--working-directory" {
```
chrisduerr, line 361: What if we're on Windows and the working directory was set? Now you're just dropping it instead of passing it.

alacritty/src/daemon.rs
```diff
@@ -43,18 +49,44 @@ where
...
 fn spawn_daemon<I, S>(program: &str, args: I) -> io::Result<()>
 where
     I: IntoIterator<Item = S> + Copy,
     S: AsRef<OsStr>,
 {
+    let mut command = Command::new(program);
+    let mut command_builder = match foreground_process_path() {
+        Ok(cwd) => command.args(args).current_dir(cwd),
+        _ => command.args(args),
+    };
```
chrisduerr, line 85: `command.args(args)` is the same in both paths here? Can't you just move that out and use `if let` instead?

chrisduerr (changes requested):

alacritty/src/event.rs
carrascomj: I am new on cross-platform in Rust, I am making a lot of mistakes like this one, thanks for the help.

CHANGELOG.md
carrascomj: Corrected.

alacritty/src/daemon.rs
carrascomj: Refactored.

alacritty/src/daemon.rs
carrascomj: Ok, I removed it. I felt like commenting it because at first glance (without looking into the macos part) it might seem like an `io::Result` would be the right choice.

alacritty/src/daemon.rs
```diff
@@ -1,10 +1,16 @@
...
 use std::ffi::OsStr;
 use std::fmt::Debug;
+#[cfg(not(any(target_os = "macos", windows)))]
+use std::fs;
 use std::io;
 #[cfg(not(windows))]
 use std::os::unix::process::CommandExt;
 #[cfg(windows)]
 use std::os::windows::process::CommandExt;
+use std::path::PathBuf;
```
chrisduerr, line 13: This seems to be only used on `#[cfg(not(windows))]`?

alacritty/src/daemon.rs
```diff
@@ -1,10 +1,16 @@
+#[cfg(not(windows))]
+use alacritty_terminal::tty;
+use std::error::Error;
```
chrisduerr, line 3: This seems to be only used on #[cfg(not(windows))]?

alacritty/src/daemon.rs
```diff
@@ -43,18 +49,43 @@ where
...
     I: IntoIterator<Item = S> + Copy,
     S: AsRef<OsStr>,
 {
+    let mut command = Command::new(program);
+    let mut command_builder = command.args(args);
+    if let Ok(cwd) = foreground_process_path() {
+        command_builder.current_dir(cwd);
+    }
+    command_builder =
+        command_builder.stdin(Stdio::null()).stdout(Stdio::null()).stderr(Stdio::null());
```
chrisduerr, line 86: I think we can just rename this variable to `builder` without any problems. That way we get rid of the ugly linebreak.

alacritty/src/event.rs
```diff
@@ -664,6 +654,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             HintAction::Command(command) => {
                 let text = self.terminal.bounds_to_string(*hint.bounds.start(), *hint.bounds.end());
                 let mut args = command.args().to_vec();
+
```
chrisduerr, line 657:
```suggestion
```

alacritty/src/event.rs
```diff
@@ -354,22 +352,14 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
         let mut args: Vec<PathBuf> = Vec::new();
 
-        let working_directory_set = !args.is_empty();
-
         // Reuse the arguments passed to Alacritty for the new instance.
         while let Some(arg) = env_args.next() {
-            // Drop working directory from existing parameters.
-            if working_directory_set && arg == "--working-directory" {
+            // Drop working directory from existing parameters; start_daemon()
+            // will set it to the parent alacritty instance current directory
```
chrisduerr, line 360: `start_daemon()` technically doesn't set the `--working-directory` parameter to anything.

```suggestion
            // On Unix, the working directory of the foreground shell is used by `start_daemon`.
```

chrisduerr (changes requested):

chrisduerr (approved):

PR #5619, opened 2021-11-18
Switch to clap-generated completions

chrisduerr:
The current completions required a lot of domain-specific knowledge
about each individual shell and their completion functionality. Much of
which is sparsely documented.

While clap does not generate perfect completions, since parameters like
`-e` are missing completions, it does a reasonable job while requiring
no work on writing these completions.

Since access to `cli.rs` isn't possible from the `build.rs`, these
completions aren't always generated on build. Instead a test verifies
that there has been no changes to these completions and provides a
simple code sample for re-generating them. This should provide a simple
solution with minimal overhead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,6 +21,7 @@ dependencies = [
  "alacritty_config_derive",
  "alacritty_terminal",
  "bitflags",
+ "clap",
  "cocoa 0.24.0",
  "copypasta",
  "crossfont",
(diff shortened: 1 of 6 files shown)


chrisduerr:
@kchibisov Kinda curious how you feel about this. Let me know what you think both about generating completions with clap and the way it is done.

kchibisov (approved):
Looks good to me.

PR #5627, opened 2021-11-22
Implement just_frame visual bell mode

Setadokalo:
Having the entire screen flash for the visual bell can be very distracting, but clearly tells the user when a bell event occurs. I think flashing the outermost ring of characters gets the best of both worlds, being both clear and not nearly as visually distracting.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -346,6 +346,9 @@
   #   - Linear
   #animation: EaseOutExpo
 
+  # Whether the visual bell should flash the entire window, or just the outer edges.
+  #just_frame: false
+
   # Duration of the visual bell flash in milliseconds. A `duration` of `0` will
   # disable the visual bell animation.
   #duration: 0
--- a/alacritty/src/config/bell.rs
+++ b/alacritty/src/config/bell.rs
@@ -10,6 +10,9 @@ pub struct BellConfig {
     /// Visual bell animation function.
     pub animation: BellAnimation,
 
+    /// Flash entire window, or just the outer frame.
+    pub just_frame: bool,
+
     /// Command to run on bell.
     pub command: Option<Program>,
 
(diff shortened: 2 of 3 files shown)


chrisduerr (changes requested):
Why bother with a visual bell at all at that point?

Setadokalo:
The visual bell is still clearly visible, and without occluding *all text on the screen* for some portion of a second. The full screen visual bell can be very distracting, especially when programs are very aggressive with the bell, like zsh sending it anytime you try to tab complete something with multiple possible resolutions.
Flashing just the edges isn't a unique idea: it's also done in [the visualbell2 patch for `st`](https://st.suckless.org/patches/visualbell2/).

chrisduerr (changes requested):
I don't see the point. If your zsh is flashing too often, why not just change your zsh configuration? If the flash is too aggressive, why not just change the time/bezier curve? This just seems like another pointless configuration option for a feature most people disable right away anyway.

Setadokalo:
changing the timer and curve doesn't prevent it from occluding the entire screen. For a moment, everything on screen is impossible to read. I don't mind zsh sending bells on tab completion, but having the entire screen become unreadable breaks focus, distracting me and slowing me down.
Also, visual bell isn't 'disabled by most people right away', it's *disabled by default* until they turn it *on.* And I'd imagine many of those who *do* know about it don't turn it on, because it's so intense.

chrisduerr (changes requested):
I don't think it's a good idea to implement multiple different kinds of visual bell. If the majority of people prefer one implementation over the other I'd be willing to change it, but I'm not willing to offer both.

Based on the number of complaints about the visual bell so far, I'd assume that most people prefer things the way they are right now. So I'd recommend just patching your Alacritty if you have such specific demands.

Setadokalo:
For anyone who does want this, here's [a patch](https://github.com/alacritty/alacritty/files/7604933/0001-Add-border-only-visual-bell-mode.patch.gz) you can apply to add it.

PR #5646, opened 2021-11-30
Added option to clip low line scroll deltas to 1.0

soooch:
I've called this `notched_scroll_wheel` for now, but hopefully a better name will come up.

This option would enable the user to force each "step" from a scroll wheel to scroll the terminal at least one line.

This is useful for machines with fractional line scroll deltas which can sometimes be below 1.0. Although Alacritty already accumulates fractional scrolls, some may prefer to simply coerce low scroll deltas up to a line.

This option approximates the behavior of Terminal.app and iTerm2 on Mac.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -107,6 +107,10 @@
   # Scrolling distance multiplier.
   #multiplier: 3
 
+  # If this is `true`, each scroll wheel "step" will scroll at least 1 line.
+  # Useful if a system uses fractional scrolling.
+  #notched_scroll_wheel: false
+
 # Font configuration
 #font:
   # Normal (roman) font face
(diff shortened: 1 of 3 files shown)


chrisduerr:
This was how Alacritty behaved originally on macOS I believe and it was absolutely terrible. Why would we want to introduce this option at all? It seems excessively specific as a configuration option and actively goes **against** the hardware-reported scroll values, basically intentionally producing **incorrect** behavior.

So I don't see a strong argument to add this. If you're interested in having this specific incorrect behavior, I think maintaining your own patch is the best path forward.

soooch:
> This was how Alacritty behaved originally on macOS I believe and it was absolutely terrible. Why would we want to introduce this option at all? It seems excessively specific as a configuration option and actively goes **against** the hardware-reported scroll values, basically intentionally producing **incorrect** behavior.
> 
> So I don't see a strong argument to add this. If you're interested in having this specific incorrect behavior, I think maintaining your own patch is the best path forward.

Can this really be considered incorrect behavior when the first party Apple terminal app does it? macOS's scrolling behavior is clearly not designed for terminal environments.

Regardless, I understand not wanting to add additional platform specific complexity.

PR #5648, opened 2021-11-30
Change default macOS Command+N binding to CreateNewWindow

camsteffen:


chrisduerr (changes requested):
I'm against changing this until after the first stable release with the multiwindow functionality.

PR #5661, opened 2021-12-05
Display area keeps history position when viewport is cleared

a5ob7r:
## Overview

This patch resolves a behavior that a vi cursor doesn't keep track of content in a scrollback buffer due to clamping the vi cursor into the viewport when viewport clear is invoked.

This is similar to #5341, but this problem is caused by viewport clear instead of new outputs to the viewport.

## Reproduction steps

Assume to run Alacritty with default configurations.

1. Create scrollback buffer(history).

```sh
$ seq $LINES
```

2. Invoke viewport clear with a little delay.

```sh
$ sleep 5; clear -x
```

3. Activate vi mode by typing `Ctrl+Shift+Space` and go to the topmost of scrollback buffer(history) by typing `g`.

4. Wait until delayed viewport clear is invoked.

Before this fix, vi cursor is clamped into a viewport.

https://user-images.githubusercontent.com/12132068/144743901-b0a78ee9-c944-4cad-bfea-397c1f60909b.mp4

After this fix, vi cursor keeps track of a line in a history.

https://user-images.githubusercontent.com/12132068/144743925-6fc3cf7a-43c9-4eda-9cff-71e058dca039.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Config option `background_opacity`, you should use `window.opacity` instead
 - Reload configuration files when their symbolic link is replaced
 - Strip trailing whitespaces when yanking from a block selection
+- Display area keeps history position when viewport is cleared
 
 ### Fixed
 
--- a/alacritty_terminal/src/grid/mod.rs
+++ b/alacritty_terminal/src/grid/mod.rs
@@ -313,9 +313,6 @@ impl<T: GridCell + Default + PartialEq + Clone> Grid<T> {
         let positions = (iter.point.line.0 + 1) as usize;
         let region = Line(0)..Line(self.lines as i32);
 
-        // Reset display offset.
-        self.display_offset = 0;
-
         // Clear the viewport.
         self.scroll_up(&region, positions);
 
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1424,10 +1424,15 @@ impl<T: EventListener> Handler for Term<T> {
                 if self.mode.contains(TermMode::ALT_SCREEN) {
                     self.grid.reset_region(..);
                 } else {
+                    let old_offset = self.grid.display_offset();
+
                     self.grid.clear_viewport();
 
+                    // Compute number of lines scrolled by clearing the viewport.
+                    let lines = self.grid.display_offset().saturating_sub(old_offset);
+
                     self.vi_mode_cursor.point.line =
-                        self.vi_mode_cursor.point.line.grid_clamp(self, Boundary::Cursor);
+                        (self.vi_mode_cursor.point.line - lines).grid_clamp(self, Boundary::Grid);
                 }
 
                 self.selection = None;
@@ -2126,7 +2131,7 @@ mod tests {
     }
 
     #[test]
-    fn clear_viewport_set_vi_cursor_into_viewport() {
+    fn clearing_viewport_keeps_history_position() {
         let size = SizeInfo::new(10.0, 20.0, 1.0, 1.0, 0.0, 0.0, false);
         let mut term = Term::new(&Config::default(), size, ());
 
@@ -2135,20 +2140,19 @@ mod tests {
             term.newline();
         }
 
-        // Change the display area and the vi cursor position.
+        // Change the display area.
         term.scroll_display(Scroll::Top);
-        term.vi_mode_cursor.point = Point::new(Line(-5), Column(3));
+
+        assert_eq!(term.grid.display_offset(), 10);
 
         // Clear the viewport.
         term.clear_screen(ansi::ClearMode::All);
 
-        assert_eq!(term.grid.display_offset(), 0);
-        assert_eq!(term.vi_mode_cursor.point.line.0, 0);
-        assert_eq!(term.vi_mode_cursor.point.column.0, 3);
+        assert_eq!(term.grid.display_offset(), 10);
     }
 
     #[test]
-    fn clear_scrollback_set_vi_cursor_into_viewport() {
+    fn clearing_viewport_with_vi_mode_keeps_history_position() {
         let size = SizeInfo::new(10.0, 20.0, 1.0, 1.0, 0.0, 0.0, false);
         let mut term = Term::new(&Config::default(), size, ());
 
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `ExpandSelection` is now a configurable mouse binding action
 - Config option `background_opacity`, you should use `window.opacity` instead
 - Reload configuration files when their symbolic link is replaced
+- Vi cursor keeps track the history when viewport clear is invoked
```
```suggestion
- Vi cursor keeps history position when viewport is cleared
```

alacritty_terminal/src/grid/mod.rs
```diff
@@ -426,6 +420,11 @@ impl<T> Grid<T> {
         self.display_offset
     }
 
+    #[inline]
+    pub fn reset_display_offset(&mut self) {
+        self.display_offset = 0;
+    }
```
chrisduerr, line 426: Why is this a method? Just to have it accessibly publicly? Why not use the `scroll` method?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1421,17 +1421,32 @@ impl<T: EventListener> Handler for Term<T> {
...
+                    // Clamp the vi cursor into the display area.
+                    let display_topmost = -(self.grid.display_offset() as i32);
+                    let display_bottommost = display_topmost + (self.screen_lines() as i32) - 1;
+                    self.vi_mode_cursor.point.line = Line(max(
+                        min(self.vi_mode_cursor.point.line.0, display_bottommost),
+                        display_topmost,
+                    ));
+                } else {
+                    self.grid.clear_viewport();
+                    self.grid.reset_display_offset();
```
chrisduerr, line 1442: Why do we reset the display offset conditionally? That just seems inconsistent. Either we always go back to the bottom, or we never do. The vi cursor's presence should have no impact on this.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1421,17 +1421,32 @@ impl<T: EventListener> Handler for Term<T> {
...
+                    let lines = self.grid.display_offset().saturating_sub(old_offset);
+                    self.vi_mode_cursor.point.line.0 -= lines as i32;
+
+                    // Clamp the vi cursor into the display area.
+                    let display_topmost = -(self.grid.display_offset() as i32);
+                    let display_bottommost = display_topmost + (self.screen_lines() as i32) - 1;
+                    self.vi_mode_cursor.point.line = Line(max(
+                        min(self.vi_mode_cursor.point.line.0, display_bottommost),
+                        display_topmost,
+                    ));
```
chrisduerr, line 1439: Why do we need to clamp?

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
a5ob7r: The reason is to keep Alacritty's previous one on vi-mode when viewport clear is invoked. I'm interested in the problem on vi-mode, and I don't know an appropriate behavior on not vi-mode when viewport clear is invoked. For example, Xterm, which is a defacto standard terminal emulator, and urxvt reset display offset when that is done, however many other terminal emulators don't do that. I may prefer the latter's style if anything. Additionally, current Alacritty changes display offset for every new output to the viewport. The latter's style may be consistent with that.

alacritty_terminal/src/grid/mod.rs
a5ob7r: > Why is this a method? Just to have it accessibly publicly?

I'm not sure that it should be public... In fact we just want to reset display offset to 0 but not change it to any value. So I create the method instead of being it publicity.

a5ob7r: But if we don't reset display offset regardless of whether or not vi-mode is enabled, I think we may need not to be it publicity or to create the method.

a5ob7r: > Why not use the scroll method?

Oh, I didn't notice that...

chrisduerr: That's what I'm here for. :smile:

alacritty_terminal/src/term/mod.rs
chrisduerr: Most terminals probably just scroll down if there's any "write" to the viewport, including clears (XTerm and URxvt certainly do). Which isn't something Alacritty does and I generally disagree with the behavior.

I'm not sure I care particularly much either way. But it **must** be consistent regardless of vi mode. So if we scroll down without it, we need to scroll down with it. And to be consistent with our other logic (i.e. not scrolling down on new output), I think keeping the offset makes sense.

a5ob7r: > I'm not sure I care particularly much either way. But it must be consistent regardless of vi mode. So if we scroll down without it, we need to scroll down with it. And to be consistent with our other logic (i.e. not scrolling down on new output), I think keeping the offset makes sense.

I got it. I'll fix to keep the display offset whether or not vi mode is active. Thanks for your comment!

alacritty_terminal/src/grid/mod.rs
```diff
@@ -335,7 +332,6 @@ impl<T: GridCell + Default + PartialEq + Clone> Grid<T> {
 
         self.saved_cursor = Cursor::default();
         self.cursor = Cursor::default();
-        self.display_offset = 0;
```
chrisduerr, line 338: Just like to point out that this is still needed here, but `clear_history` does it for us (as it should).

alacritty_terminal/src/term/mod.rs
```diff
@@ -1422,10 +1422,15 @@ impl<T: EventListener> Handler for Term<T> {
                 if self.mode.contains(TermMode::ALT_SCREEN) {
                     self.grid.reset_region(..);
                 } else {
+                    let old_offset = self.grid.display_offset();
+
                     self.grid.clear_viewport();
 
+                    // How many lines has been scrolled by clearing viewport.
```
chrisduerr, line 1429: I don't like comments which are phrased as questions rather than statements. I think it's more appropriate to have something like `Compute number of lines scrolled by clearing the viewport`.

chrisduerr (changes requested):
You only have tests for display offset staying in place or for display offset being reset completely, but a significant part of the new logic is around what happens in-between.

Are you interested in writing another test to verify this behavior is correct too?

a5ob7r:
> Are you interested in writing another test to verify this behavior is correct too?

Yes. And I have fixed problems above included it.

chrisduerr (approved):
Works as expected, thanks.

PR #5671, opened 2021-12-09
Remove trailing whitespaces when using block selection

### LINKED ISSUE #5638: Strip trailing whitespaces in rectangular select
One of the most common reasons to use rectangular select is when you're in a ncurses app with a vertical split (e.g vim, weechat, tmux...) and you want to select some part of one of the panes. If you use a normal selection, you will select the content of the other vertical panes:

image

Instead, by using a rectangular select, we can select the content from a single pane:

image

However, doing this in Alacritty doesn't remove trailing whitespaces, so the output will be a rectangle filled with whitespaces at the end of each line. Here is the output, in my editor which shows trailing whitespaces:

image

Urxvt strips all trailing whitespaces from rectangular select. I think this would be a good idea to implement in Alacritty. The UI also indicates this by having a different selection display for the trailing whitespaces that will be stripped:

image

jeremycostanzo:
Resolves #5638

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `ExpandSelection` is now a configurable mouse binding action
 - Config option `background_opacity`, you should use `window.opacity` instead
 - Reload configuration files when their symbolic link is replaced
+- Strip trailing whitespaces when yanking from a block selection
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


chrisduerr (changes requested):
Pretty sure this is the same I had in mind when looking into this. So sounds good to me.

Only thing missing is a changelog entry.

jeremycostanzo:
> 

Thank you, I added it.

kchibisov (approved):

PR #5688, opened 2021-12-15
Fix scrolling variables type documentation

jpic:
This fixes the following error when the buffer line is uncommented:
```
    [ERROR] See log at /tmp/Alacritty-79411.log ($ALACRITTY_LOG):
    Config error: env: invalid type: integer `100000`, expected a string
```


chrisduerr (changes requested):
This is just wrong.

chrisduerr:
There is no issue.

PR #5692, opened 2021-12-17
Add ability to catch unhandled vte calls

grant0417:
Currently there is no way to handle vte calls that are not handled by `alacritty_terminal::ansi::Processor`. This would allow for custom logic by a `alacritty_terminal::ansi::Handler` for calls that are not handle by `alacritty_terminal::ansi::Processor`.

### DIFF
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -263,6 +263,13 @@ impl<'a, H: Handler + 'a> Performer<'a, H> {
     }
 }
 
+/// Indicates if a handler has handled an action
+#[derive(Debug, Eq, PartialEq, Copy, Clone)]
+pub enum HandledStatus {
+    Handled,
+    Unhandled,
+}
+
 /// Type that handles actions from the parser.
 ///
 /// XXX Should probably not provide default impls for everything, but it makes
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
> Currently there is no way to handle vte calls that are not handled by alacritty_terminal::ansi::Processor. This would allow for custom logic by a alacritty_terminal::ansi::Handler for calls that are not handle by alacritty_terminal::ansi::Processor.

Why would you ever want that? Are you trying to use some proprietary custom escape sequences? Escape sequences that are standardized should be added to `alacritty_terminal` directly.

grant0417:
Hey Chris, yeah I quite like the alacritty_terminal ansi parser but want to be able to support custom/non-standard/proprietary escape sequences.

This is really just a fallback and should not be used for standard sequences which would definitely be upstreamed. An example of the non-standard codes I want to support is OSC 133 https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md which helps with determining where the prompt is or OSC 8 https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda which adds support for URL handling.

chrisduerr:
And how do you plan to consume this API? The `alacritty_terminal` library intentionally obstructs this. This really doesn't seem like you're using things the way they were meant to be used…

grant0417:
I want to use `ansi::Parser` to parse the ansi escape codes from my terminal, but I want to be able to parse any non-standard escape codes I am using. What part is intentionally obstructed, because currently it is working as I want, it allows me to handle cursor movement, scrolling, printing, etc, but not any non-standard escape codes.

chrisduerr (changes requested):
If you're implementing your own terminal, I recommend depending on `vte` and not `alacritty_terminal`. The `alacritty_terminal` crate is built to be a full terminal emulator and as such modifications to use it just as an ansi parser are inappropriate.

While there have been attempts to move the parsing into the `vte` crate, none of them have been fruitful due to a lack of interest. So honestly I think the best way for anyone to use this code is just to take what you can use and extend it as needed. It's just ~1500LoC, most of which are random enum variants, really not worth the hassle to try to bend some crate surface to your needs when you can just take it as-is and extend on your own.

Issue #4025, opened 2020-07-24
Country flag emojis are not displayed correctly


dashezup:
### System

OS: Void Linux
Version: alacritty 0.4.3
Linux/BSD: Wayland, Swaywm
Font: Fira Code, [Apple Color Emoji](https://github.com/samuelngs/apple-emoji-linux)

### fontconfig

~/.config/fontconfig/fonts.conf

```
<?xml version="1.0"?>
<!DOCTYPE fontconfig SYSTEM "fonts.dtd">
<fontconfig>
  <alias>
    <family>foobar</family>
    <prefer>
      <family>Fira Code</family>
      <family>Apple Color Emoji</family>
     </prefer>
  </alias>
</fontconfig>
```

I have set font family to foobar in alacritty.yml, most of other emojis are displayed correctly except country flags emoji, I have also tried with Liberation Mono / DejaVu Sans + Noto Color Emoji, both can't display country flag character correctly.

Seems country flags has two characters, I guess this could be related, some flags such like Crossed Flags are rendered correctly.

List of flags emoji on [Emojipedia](https://emojipedia.org/flags/)

### Screenshots

![image_2020-07-24_11-13-04](https://user-images.githubusercontent.com/67376103/88386413-da0cfd80-cd9f-11ea-8e55-b67315cc3d93.png)

I have also tried alacritty on macOS Catalina, installed by using brew, got similar result.

![tg_image_929287869](https://user-images.githubusercontent.com/67376103/88386417-db3e2a80-cd9f-11ea-9123-c8e240119e14.jpeg)

chrisduerr:
This is basically a duplicate of https://github.com/alacritty/alacritty/issues/50. Though I think the interest for grapheme clusters is even lower than that for ligatures.

Issue #3975, opened 2020-07-15
Support For Emoji Modifiers


carloabelli:
Now that there is emoji support, support for emoji modifiers would be nice as well. For example the `👋🏻` emoji does not render correctly (instead renders as its component parts):

![2020-07-15T14:46:46,290402768+00:00](https://user-images.githubusercontent.com/3083317/87559614-83f5d700-c688-11ea-8767-780c2ec59c9d.png)

### System

OS: Linux
Version: alacritty 0.4.3
Linux: wayland (sway)

carloabelli:
It seems that at least some terminals support this behavior. See https://github.com/neovim/neovim/issues/7151. I'm not exactly sure how they do it, but it does not appear to be hack you described.

kchibisov:
The only terminal that supports such things, I'm aware of is kitty, and it's broken in unusable way, when it comes to such things. The thing is that such terminals do combine them without preserving the width, and so  break the applications you're using.

For example in kitty.

![picture](https://user-images.githubusercontent.com/27620401/87575424-90098500-c6d8-11ea-8d91-4bd9593a79bb.png)

I've moved my cursor the the start of the input in a prompt with `ctrl + a`, and you may see that it's over prompt itself, which is unusable.

On the other hand in alacritty the emojis are not combined, but you have a usable state.

![picture](https://user-images.githubusercontent.com/27620401/87575625-da8b0180-c6d8-11ea-99f2-a3c04dc52f02.png)

That's more noticeable in application like `weechat`, since once someone posted an such emoji into it everything goes wild, and content is not being aligned anymore.

The thing way it's happening, because applications rely on `wcwidth` like functions, and so using `wcswidth` like functions will break implicit synchronization. It could also be broken by mismatches of `wcwidth` but it happens way less frequent than `wcswidth` and `wcwidth` mismatches, and the usage of `wcwidth` is way less simpler and performant.

christianparpart:
`wc(s)width` MUST die! If you look at a single codepoints only then you are doomed to not support unicode. Unicode (user perceived) characters can be more than one codepoint. And kitty is doing a decent job at finding out what Codepoint belong to which grid cell.

The problem is that many apps rely on `wcwidth` and this API is broken by definition.
[...]

Issue #936, opened 2017-12-06
Hangul / Multi grapheme input is broken


azah:
Korean characters are made up of at least two graphemes, max 3-4. Normal input waits for a space or the maximum graphemes before moving onto the next character.

Intended behavior: Typing ㄲ ㅗ ㄱ should produce 꼭, ㄷ ㅏ ㄹ ㄱ should produce 닭, etc.

Alacritty behavior: Typing ㄲ ㅗ ㄱ stays as the individual graphemes. It doesn't appear to enter the mode of character composition.

System: OSX
Tested with iTerm2, no issues there on my setup.

azah:
I asked on #rust and was pointed to this library: https://crates.io/crates/unicode-segmentation

jwilm:
Thanks for the issue! This *may* be an issue in `winit`. When you are composing a character, do the individual characters normally show up until the grapheme is finished and then disappear?

Regarding input, unicode segmentation is relevant there, and so is unicode-width. harfbuzz is certainly a render-time tool though. And actually, we don't properly support multi-codepoint grapheme clusters yet; cells are limited to single code points for now.

azah:
Here is a gif: https://gfycat.com/gifs/detail/SaneEdibleIrishsetter

Essentially it builds in place until you press space, or reach the hangul limit (3 or 4 depending).
At least for Korean. I can't speak for Mandarin or Arabic which probably work differently.

jwilm:
And it works similarly in iTerm?

azah:
Yep! iTerm2 and Terminal.app work the same as that gif, and have no problem switching input sources.

jwilm:
In that case, this sounds like a duplicate of #306. Generally speaking, full unicode support is a bit tricky since it requires changes to how cell data is stored in the grid. Our current implementation only supports having a single code point per cell (a Rust `char` type) and not arbitrary UTF-8 sequences. Supporting arbitrary UTF-8 requires running unicode-segmentation on the input we get, storing this efficiently for each cell, and then using harfbuzz to shape text when rendering.

What's really needed to get this moving forward is someone who is familiar with these problems to propose an implementation for Alacritty. The *really hard* part is doing all of this and not leaving performance on the table.

mbrubeck:
[...]

PR #5696, opened 2021-12-17
Support for Font Ligatures using harfbuzz

### LINKED ISSUE #50: Support for ligatures

fee1-dead:
NOTE: https://github.com/alacritty/crossfont/pull/35/ has to be merged first. The current change modifies the dependency to point at my fork. It should point to a released version instead!

Closes #50. We should be able to statically link to harfbuzz on every supported platform, so not gating this behind `cfg(unix)`.

Supersedes #2677 and #5615.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,6 +24,7 @@ dependencies = [
  "fnv",
  "gl_generator",
  "glutin",
+ "harfbuzz_rs",
  "libc",
  "log",
  "notify",
@@ -359,9 +360,8 @@ dependencies = [
 
 [[package]]
 name = "crossfont"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1299695a4c6417b7e4a6957bd963478406e148b7b0351e2f2ce31296b0518251"
+version = "0.3.1"
+source = "git+https://github.com/fee1-dead/crossfont#dbf0fc720f75918e56c7d8b14fd42bf40b159ede"
 dependencies = [
  "cocoa",
  "core-foundation 0.9.2",
@@ -573,6 +573,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7684cf33bb7f28497939e8c7cf17e3e4e3b8d9a0080ffa4f8ae2f515442ee855"
 
 [[package]]
+name = "freetype"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bee38378a9e3db1cc693b4f88d166ae375338a0ff75cb8263e1c601d51f35dc6"
+dependencies = [
+ "freetype-sys",
+ "libc",
+]
+
+[[package]]
 name = "freetype-rs"
 version = "0.26.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -721,6 +731,30 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3da5951a1569dbab865c6f2a863efafff193a93caf05538d193e9e3816d21696"
 dependencies = [
  "gl_generator",
+]
+
+[[package]]
+name = "harfbuzz-sys"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bf8c27ca13930dc4ffe474880040fe9e0f03c2121600dc9c95423624cab3e467"
+dependencies = [
+ "cc",
+ "core-graphics 0.22.3",
+ "core-text",
+ "foreign-types 0.3.2",
+ "freetype",
+ "pkg-config",
+]
+
+[[package]]
+name = "harfbuzz_rs"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "58c18cfc23920d010f7637b86240a5f8892dfdbcb4f0e60828f8a9dc6ff5521d"
+dependencies = [
+ "bitflags",
+ "harfbuzz-sys",
 ]
 
 [[package]]
--- a/alacritty/Cargo.toml
+++ b/alacritty/Cargo.toml
@@ -28,12 +28,13 @@ serde_json = "1"
 glutin = { version = "0.28.0", default-features = false, features = ["serde"] }
 notify = "4"
 parking_lot = "0.11.0"
-crossfont = { version = "0.3.1", features = ["force_system_fontconfig"] }
+crossfont = { git = "https://github.com/fee1-dead/crossfont", features = ["force_system_fontconfig"] }
 copypasta = { version = "0.7.0", default-features = false }
 libc = "0.2"
 unicode-width = "0.1"
 bitflags = "1"
 dirs = "3.0.1"
+harfbuzz_rs = "2.0.1"
 
 [dev-dependencies]
 clap = "2.33.3"
--- a/alacritty/src/config/font.rs
+++ b/alacritty/src/config/font.rs
@@ -14,7 +14,7 @@ use crate::config::ui_config::Delta;
 /// field in this struct. It might be nice in the future to have defaults for
 /// each value independently. Alternatively, maybe erroring when the user
 /// doesn't provide complete config is Ok.
-#[derive(ConfigDeserialize, Default, Debug, Clone, PartialEq, Eq)]
+#[derive(ConfigDeserialize, Debug, Clone, PartialEq, Eq)]
 pub struct Font {
     /// Extra spacing per character.
     pub offset: Delta<i8>,
@@ -23,6 +23,8 @@ pub struct Font {
     pub glyph_offset: Delta<i8>,
 
     pub use_thin_strokes: bool,
+
+    pub ligatures: bool,
 
     /// Normal font face.
     normal: FontDescription,
@@ -69,6 +71,22 @@ impl Font {
     /// Get bold italic font description.
     pub fn bold_italic(&self) -> FontDescription {
         self.bold_italic.desc(&self.normal)
+    }
+}
+
+impl Default for Font {
+    fn default() -> Self {
+        Self {
+            offset: Default::default(),
+            glyph_offset: Default::default(),
+            use_thin_strokes: false,
+            ligatures: true,
+            normal: Default::default(),
+            bold: Default::default(),
+            italic: Default::default(),
+            bold_italic: Default::default(),
+            size: Default::default(),
+        }
     }
 }
 
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -45,8 +45,9 @@ use crate::display::meter::Meter;
 use crate::display::window::Window;
 use crate::event::{Mouse, SearchState};
 use crate::message_bar::{MessageBuffer, MessageType};
-use crate::renderer::rects::{RenderLines, RenderRect};
+use crate::renderer::rects::RenderRect;
 use crate::renderer::{self, GlyphCache, QuadRenderer};
+use crate::text_run::{TextRun, TextRunIter};
 
 pub mod content;
 pub mod cursor;
@@ -357,8 +358,8 @@ impl Display {
             info!("Initializing glyph cache...");
             let init_start = Instant::now();
 
-            let cache =
-                renderer.with_loader(|mut api| GlyphCache::new(rasterizer, &font, &mut api))?;
+            let cache = renderer
+                .with_loader(|mut api| GlyphCache::new(rasterizer, &font, &mut api, config))?;
 
             let stop = init_start.elapsed();
             let stop_f = stop.as_secs() as f64 + f64::from(stop.subsec_nanos()) / 1_000_000_000f64;
@@ -479,14 +480,16 @@ impl Display {
         config: &UiConfig,
         search_state: &SearchState,
     ) {
+        let highlighted_hint = self.highlighted_hint.as_ref().map(|m| m.bounds.clone());
+        let vi_highlighted_hint = self.vi_highlighted_hint.as_ref().map(|m| m.bounds.clone());
         // Collect renderable content before the terminal is dropped.
         let mut content = RenderableContent::new(config, self, &terminal, search_state);
-        let mut grid_cells = Vec::new();
-        for cell in &mut content {
-            grid_cells.push(cell);
-        }
         let background_color = content.color(NamedColor::Background as usize);
         let display_offset = content.display_offset();
+        let grid_text_runs: Vec<TextRun> = {
+            TextRunIter::<()>::from_content(&mut content, highlighted_hint, vi_highlighted_hint)
+                .collect()
+        };
         let cursor = content.cursor();
 
         let cursor_point = terminal.grid().cursor.point;
@@ -507,7 +510,7 @@ impl Display {
             api.clear(background_color);
         });
 
-        let mut lines = RenderLines::new();
+        let mut rects: Vec<RenderRect> = Vec::new();
 
         // Draw grid.
         {
@@ -518,29 +521,37 @@ impl Display {
             self.renderer.set_viewport(&size_info);
 
             let glyph_cache = &mut self.glyph_cache;
-            let highlighted_hint = &self.highlighted_hint;
-            let vi_highlighted_hint = &self.vi_highlighted_hint;
+
             self.renderer.with_api(config, &size_info, |mut api| {
-                // Iterate over all non-empty cells in the grid.
-                for mut cell in grid_cells {
-                    // Underline hints hovered by mouse or vi mode cursor.
-                    let point = viewport_to_point(display_offset, cell.point);
-                    if highlighted_hint.as_ref().map_or(false, |h| h.bounds.contains(&point))
-                        || vi_highlighted_hint.as_ref().map_or(false, |h| h.bounds.contains(&point))
-                    {
-                        cell.flags.insert(Flags::UNDERLINE);
+                for text_run in grid_text_runs {
+                    if text_run.flags.contains(Flags::UNDERLINE) {
+                        let underline_metrics = (
+                            metrics.descent,
+                            metrics.underline_position,
+                            metrics.underline_thickness,
+                        );
+                        rects.push(RenderRect::from_text_run(
+                            &text_run,
+                            underline_metrics,
+                            &size_info,
+                        ));
                     }
-
-                    // Update underline/strikeout.
-                    lines.update(&cell);
-
-                    // Draw the cell.
-                    api.render_cell(cell, glyph_cache);
+                    if text_run.flags.contains(Flags::STRIKEOUT) {
+                        let strikeout_metrics = (
+                            metrics.descent,
+                            metrics.strikeout_position,
+                            metrics.strikeout_thickness,
+                        );
+                        rects.push(RenderRect::from_text_run(
+                            &text_run,
+                            strikeout_metrics,
+                            &size_info,
+                        ));
+                    }
+                    api.render_text_run(text_run, glyph_cache);
                 }
             });
         }
-
-        let mut rects = lines.rects(&metrics, &size_info);
 
         if let Some(vi_mode_cursor) = vi_mode_cursor {
             // Indicate vi mode by showing the cursor's position in the top right corner.
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -487,8 +487,13 @@ impl Display {
         let background_color = content.color(NamedColor::Background as usize);
         let display_offset = content.display_offset();
         let grid_text_runs: Vec<TextRun> = {
-            TextRunIter::<()>::from_content(&mut content, highlighted_hint, vi_highlighted_hint)
-                .collect()
+            let mut vec = Vec::with_capacity(terminal.screen_lines() * 2);
+            vec.extend(TextRunIter::from_content(
+                &mut content,
+                highlighted_hint,
+                vi_highlighted_hint,
+            ));
+            vec
         };
         let cursor = content.cursor();
 
--- a/alacritty/src/main.rs
+++ b/alacritty/src/main.rs
@@ -43,6 +43,7 @@ mod message_bar;
 mod panic;
 mod renderer;
 mod scheduler;
+mod text_run;
 mod window_context;
 
 mod gl {
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -954,8 +954,6 @@ impl<'a> RenderApi<'a> {
     }
 
     pub fn render_text_run(&mut self, text_run: TextRun, glyph_cache: &mut GlyphCache) {
-        let TextRunContent { text, zero_widths } = &text_run.content;
-
         // Get font key for cell
         let font_key = Self::determine_font_key(text_run.flags, glyph_cache);
 
@@ -963,12 +961,15 @@ impl<'a> RenderApi<'a> {
             GlyphIter::Hidden
         } else {
             GlyphIter::Shaped(
-                glyph_cache.shape_run(text, font_key, self).expect("read font").into_iter(),
+                glyph_cache
+                    .shape_run(&text_run.content.text, font_key, self)
+                    .expect("read font")
+                    .into_iter(),
             )
         };
 
         for ((mut cell, glyph), zero_width_chars) in
-            text_run.cells().zip(shaped_glyphs).zip(zero_widths.iter())
+            text_run.cells().zip(shaped_glyphs).zip(text_run.content.zero_widths.iter())
         {
             self.add_render_item(&cell, &glyph);
             // Add empty spacer for full width characters
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -1,3 +1,4 @@
+use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::fmt::{self, Display, Formatter};
 use std::hash::BuildHasherDefault;
@@ -6,10 +7,11 @@ use std::{io, ptr};
 
 use bitflags::bitflags;
 use crossfont::{
-    BitmapBuffer, Error as RasterizerError, FontDesc, FontKey, GlyphKey, Rasterize,
+    BitmapBuffer, Error as RasterizerError, FontDesc, FontKey, GlyphId, GlyphKey, Rasterize,
     RasterizedGlyph, Rasterizer, Size, Slant, Style, Weight,
 };
 use fnv::FnvHasher;
+use harfbuzz_rs::{Feature, Owned, UnicodeBuffer};
 use log::{error, info};
 use unicode_width::UnicodeWidthChar;
 
@@ -24,6 +26,7 @@ use crate::display::content::RenderableCell;
 use crate::gl;
 use crate::gl::types::*;
 use crate::renderer::rects::{RectRenderer, RenderRect};
+use crate::text_run::{TextRun, TextRunContent};
 
 pub mod rects;
 
@@ -91,7 +94,7 @@ pub struct TextShaderProgram {
     u_background: GLint,
 }
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, Default)]
 pub struct Glyph {
     tex_id: GLuint,
     multicolor: bool,
@@ -106,9 +109,6 @@ pub struct Glyph {
 }
 
 /// Naïve glyph cache.
-///
-/// Currently only keyed by `char`, and thus not possible to hold different
-/// representations of the same code point.
 pub struct GlyphCache {
     /// Cache of buffered glyphs.
     cache: HashMap<GlyphKey, Glyph, BuildHasherDefault<FnvHasher>>,
@@ -136,6 +136,11 @@ pub struct GlyphCache {
 
     /// Font metrics.
     metrics: crossfont::Metrics,
+
+    /// A map of FontKey to harfbuzz Font.
+    fonts: HashMap<FontKey, Owned<harfbuzz_rs::Font<'static>>, BuildHasherDefault<FnvHasher>>,
+
+    font_features: Box<[Feature]>,
 }
 
 impl GlyphCache {
@@ -143,6 +148,7 @@ impl GlyphCache {
         mut rasterizer: Rasterizer,
         font: &Font,
         loader: &mut L,
+        config: &UiConfig,
     ) -> Result<GlyphCache, crossfont::Error>
     where
         L: LoadGlyph,
@@ -152,9 +158,19 @@ impl GlyphCache {
         // Need to load at least one glyph for the face before calling metrics.
         // The glyph requested here ('m' at the time of writing) has no special
         // meaning.
-        rasterizer.get_glyph(GlyphKey { font_key: regular, character: 'm', size: font.size() })?;
+        rasterizer.get_glyph(GlyphKey {
+            font_key: regular,
+            id: GlyphId::char('m'),
+            size: font.size(),
+        })?;
 
         let metrics = rasterizer.metrics(regular, font.size())?;
+
+        let features: Box<[Feature]> = if config.font.ligatures {
+            Box::new([Feature::new(b"liga", 1, ..), Feature::new(b"calt", 1, ..)])
+        } else {
+            Box::new([])
+        };
 
         let mut cache = Self {
             cache: HashMap::default(),
@@ -166,6 +182,8 @@ impl GlyphCache {
             bold_italic_key: bold_italic,
             glyph_offset: font.glyph_offset,
             metrics,
+            fonts: HashMap::default(),
+            font_features: features,
         };
 
         cache.load_common_glyphs(loader);
@@ -178,7 +196,7 @@ impl GlyphCache {
 
         // Cache all ascii characters.
         for i in 32u8..=126u8 {
-            self.get(GlyphKey { font_key: font, character: i as char, size }, loader, true);
+            self.get(GlyphKey { font_key: font, id: GlyphId::char(i as char), size }, loader, true);
         }
     }
 
@@ -247,6 +265,46 @@ impl GlyphCache {
         FontDesc::new(desc.family.clone(), style)
     }
 
+    pub fn shape_run<L>(
+        &mut self,
+        text_run: &str,
+        font_key: FontKey,
+        loader: &mut L,
+    ) -> Result<Vec<Glyph>, Box<dyn std::error::Error>>
+    where
+        L: LoadGlyph,
+    {
+        if let Entry::Vacant(v) = self.fonts.entry(font_key) {
+            let path = self.rasterizer.font_path(font_key)?;
+            let face = harfbuzz_rs::Face::from_file(path, 0)?;
+            let font = harfbuzz_rs::Font::new(face);
+            v.insert(font);
+        };
+
+        let font = self.fonts.get(&font_key).unwrap();
+        let buffer = UnicodeBuffer::new().add_str(text_run);
+        let result = harfbuzz_rs::shape(font, buffer, &self.font_features);
+        Ok(result
+            .get_glyph_infos()
+            .iter()
+            .map(move |glyph_info| {
+                let codepoint = glyph_info.codepoint;
+
+                // Codepoint of 0 indicates a missing or undefined glyph
+                let id: crossfont::GlyphId = if codepoint == 0 {
+                    GlyphId::char(
+                        text_run.split_at(glyph_info.cluster as usize).1.chars().next().unwrap(),
+                    )
+                } else {
+                    GlyphId::with_glyph_index(codepoint as u32)
+                };
+
+                let glyph_key = GlyphKey { id, font_key, size: self.font_size };
+                self.get(glyph_key, loader, true)
+            })
+            .collect())
+    }
+
     /// Get a glyph from the font.
     ///
     /// If the glyph has never been loaded before, it will be rasterized and inserted into the
@@ -270,8 +328,7 @@ impl GlyphCache {
             Ok(rasterized) => self.load_glyph(loader, rasterized),
             // Load fallback glyph.
             Err(RasterizerError::MissingGlyph(rasterized)) if show_missing => {
-                // Use `\0` as "missing" glyph to cache it only once.
-                let missing_key = GlyphKey { character: '\0', ..glyph_key };
+                let missing_key = GlyphKey { id: GlyphId::placeholder(), ..glyph_key };
                 if let Some(glyph) = self.cache.get(&missing_key) {
                     *glyph
                 } else {
@@ -306,7 +363,7 @@ impl GlyphCache {
         // right side of the preceding character. Since we render the
         // zero-width characters inside the preceding character, the
         // anchor has been moved to the right by one cell.
-        if glyph.character.width() == Some(0) {
+        if glyph.id.as_char().map(|c| c.width()).flatten() == Some(0) {
             glyph.left += self.metrics.average_advance as i32;
         }
 
@@ -337,7 +394,7 @@ impl GlyphCache {
 
         self.rasterizer.get_glyph(GlyphKey {
             font_key: regular,
-            character: 'm',
+            id: GlyphId::char('m'),
             size: font.size(),
         })?;
         let metrics = self.rasterizer.metrics(regular, font.size())?;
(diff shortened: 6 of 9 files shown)


fee1-dead:
@chrisduerr I would appreciate some feedback for the current change. It is likely that not a lot of performance improvements can be made without significantly refactoring the inner `Grid` structure (i.e. moving shaping logic to `alacritty_terminal` and saving results in that structure).

chrisduerr:
> It is likely that not a lot of performance improvements can be made without significantly refactoring the inner Grid structure (i.e. moving shaping logic to alacritty_terminal and saving results in that structure).

If a larger refactoring is necessary to match the current performance, then we cannot move forward without these changes. Ligatures are not worth dropping a single frame for.

fee1-dead:
I think we could go with an implementation similar to kitty's. They continued using cell-based rendering, except they break ligatures into individual cells that have extra data that may point to the special ligature pixmap. Text runs are not broken by the cursor or any style change, so the division is simple and not dynamic.

kitty: (each line is shaped)
 
image

current change: (different colors are shaped individually)

image

chrisduerr:
> Plus when ligatures are disabled the only overhead would be the extra bytes that each cell will take.

Since Alacritty already has dynamic storage in place for cell content that isn't essential, it should be possible to implement this without any extra bits then.

fee1-dead:
I just tried an implementation of this idea and it turned out to have worse performance than the current change, probably because it introduces a lot more complexity. I don't think it is worth it to try improving the performance further, because large refactorings induced headaches for me.

@chrisduerr, to minimize the regression, I'll make it so that harfbuzz is not called when the ligatures option is disabled for folks that would like a performant terminal, which should make the performance impact minimal with the option disabled (I'll test before it gets pushed). If you are still unhappy, then I can disable ligatures by default, does this sound good? This is the only way forward, at least for me, if anyone else wants to take over, I'd be happy to send a diff.

chrisduerr:
> @chrisduerr, to minimize the regression, I'll make it so that harfbuzz is not called when the ligatures option is disabled for folks that would like a performant terminal, which should make the performance impact minimal with the option disabled (I'll test before it gets pushed). If you are still unhappy, then I can disable ligatures by default, does this sound good? This is the only way forward, at least for me, if anyone else wants to take over, I'd be happy to send a diff.

Having options to opt-out of performance is not acceptable. Either Alacritty manages to offer all features at high performance, or it doesn't. Just disabling all the useful features by default as an excuse to pretend like Alacritty has reasonable performance is not productive.

I **strongly** recommend anyone who's not willing to spend some time optimizing the ligature implementation to please stay away from sending PRs related to it. That's really just wasting the time of everyone involved.

xsrvmy:
> Just disabling all the useful features by default as an excuse to pretend like Alacritty has reasonable performance is not productive.

Actually, I would argue that ligatures should be off by default, because some programs might have confusing behavior with ligatures on (eg. using <= as an arrow). It would be best if ligatures can be toggled on and off, or configured to turn on automatically when certain programs are running (eg. less, vim, nvim, emacs, etc.).

chrisduerr:
> Actually, I would argue that ligatures should be off by default, because some programs might have confusing behavior with ligatures on (eg. using <= as an arrow). It would be best if ligatures can be toggled on and off, or configured to turn on automatically when certain programs are running (eg. less, vim, nvim, emacs, etc.).

Oh just to clarify, regardless of what the final ligature implementation might be, it **will** be off by default with a configuration option to toggle it. My comment's goal was solely to point out that having options that significantly impact Alacritty's performance is not an option. Performance is gauged based on the worst-case, not the best-case scenario.

xsrvmy:
@fee1-dead when running `dir` I saw the HT control character rendered at one point, but they fixed themselves somehow.

Logarithmus:
> > It is likely that not a lot of performance improvements can be made without significantly refactoring the inner Grid structure (i.e. moving shaping logic to alacritty_terminal and saving results in that structure).
> 
> If a larger refactoring is necessary to match the current performance, then we cannot move forward without these changes. Ligatures are not worth dropping a single frame for.

Just disable ligatures by default. Those who want to use them and are OK with sacrificing a bit of FPS can enable. It's just that simple. Why are so opposed to merging this feature?

Logarithmus:
> > @chrisduerr, to minimize the regression, I'll make it so that harfbuzz is not called when the ligatures option is disabled for folks that would like a performant terminal, which should make the performance impact minimal with the option disabled (I'll test before it gets pushed). If you are still unhappy, then I can disable ligatures by default, does this sound good? This is the only way forward, at least for me, if anyone else wants to take over, I'd be happy to send a diff.
> 
> Having options to opt-out of performance is not acceptable. Either Alacritty manages to offer all features at high performance, or it doesn't. Just disabling all the useful features by default as an excuse to pretend like Alacritty has reasonable performance is not productive.
> 
> I **strongly** recommend anyone who's not willing to spend some time optimizing the ligature implementation to please stay away from sending PRs related to it. That's really just wasting the time of everyone involved.

Very "nice" & "supportive" attitude. Good luck getting any more pull requests to your repo. Came here after a couple of years of waiting for ligatures support. Will continue happily using `kitty` then.

chrisduerr:
> @chrisduerr Just disable ligatures by default. Those who want to use them and are OK with sacrificing a bit of FPS can enable. It's just that simple. Why are you so opposed to merging this feature?

Merging garbage and slow code is not an option. Either it's implemented properly or not at all. Alacritty will never get any features that opt users out of Alacritty being a reasonably quick terminal emulator, that would completely defeat the purpose of Alacritty.

xsrvmy:
I think I want to look into this a bit. Actually I've tested this with vtebench and didn't observe any significant slowdown. Is there some other test I need to perform?

xsrvmy:
> I strongly recommend anyone who's not willing to spend some time optimizing the ligature implementation to please stay away from sending PRs related to it. That's really just wasting the time of everyone involved.

Isn't the point of a draft PR to allow others to help?

fee1-dead:
In almost all applications or projects, eventually there will be a feature that cannot be added without a regression in performance.

If people really want the feature to be added, I have no idea why it must be "fast" such that it displays content from `/dev/urandom` with a minuscule font size. My proposal was to try and reduce the non-ligatures impact to zero as to give ligatures an experimental status. Clearly @chrisduerr did not like that idea because if all people end up enabling ligatures alacritty will fail to deliver the performance promise.

When developers write something in Rust, their primary focus could be performance, binary size, safety, etc. But I do not think that focus should stay forever. If it is not feature complete, it cannot be "the best terminal emulator". This project started with rather paradoxical views: ([Announcing Alacritty](https://jwilm.io/blog/announcing-alacritty/))

> [Alacritty](https://github.com/jwilm/alacritty) is the result of frustration with existing terminal emulators. Using `vim` inside `tmux` in many terminals was a particularly bad experience. None of them were ever quite fast enough.

Terminal emulators were not always the problem. See [this reddit post](https://www.reddit.com/r/neovim/comments/89roud/why_is_neovim_in_tmux_so_slow/) where some continues to complain about `nvim` being very slow when run inside `tmux`, where the terminal emulator is `alacritty`! If someone tries to use `nvim`'s terminal emulator inside `tmux`, it *is* going to be slow.

> 1. Performance: Alacritty should be the fastest terminal emulator available anywhere.
> 2. Appearance: Alacritty should have beautiful font rendering and look fantastic on all supported platforms.
> 3. Simplicity: Alacritty should be conservative about which features it offers. As we’ve learned from past terminal emulators, it’s far too easy to become bloated. 

Good appearance without the bloat nor bad performance, is, in many ways, impossible to achieve. Alacritty is in a confusing state: do you want good appearance or not? Do you want high performance or not? If you want both, then it will never be the fastest terminal emulator. Being written in Rust does not help. If you just want the high performance, then close #50 and tell everyone that this project is not going to have font ligatures because it causes regressions. Please do not continue to give people hope that it will land if some magic contributor finds a way to implement ligatures without a 2x regression in performance.

chrisduerr:
> Please do not continue to give people hope that it will land if some magic contributor finds a way to implement ligatures without a 2x regression in performance.

Only because you're incapable of implementing it, does not mean it is not possible. It is most certainly possible to implement ligatures without major performance regressions.

fee1-dead:
Sorry, I retract my last sentence on that comment. I did come up with a way without regressions and that is to completely rewrite the grid storage so that it works well with ligatures (each input will only cause reshape of individual lines). It will require the effort of rewriting the entire project and is going to conflict with every other major PR.

For people who want ligatures: consider switching to another project that actually supports ligatures, but if you really want to use alacritty, you could maintain a fork.

gghhkm:
> > Please do not continue to give people hope that it will land if some magic contributor finds a way to implement ligatures without a 2x regression in performance.
> 
> Only because you're incapable of implementing it, does not mean it is not possible. It is most certainly possible to implement ligatures without major performance regressions.

Let's see *you* implement it then, get off your lazy ass and do it if you're so smart otherwise close it and STFU.

Nytelife26:
> Only because you're incapable of implementing it, does not mean it is not possible. It is most certainly possible to implement ligatures without major performance regressions.

Then instead of calling valuable contributors that have sunk time into trying incapable, if you're so sure it can be feasibly done, go do it :)

Nytelife26:
I'm willing to give this a shot, and to try making it performant. No promises, but I'm willing to try. @chrisduerr, how do you suggest benchmarking the main branch and the ligatures PRs to measure progress with minimizing regressions?

fee1-dead:
Nice, @Nytelife26! For me I used [MangoHud](https://github.com/flightlessmango/MangoHud) to monitor the FPS and ran `cat /dev/urandom` with a minuscule font size to compare the performance

Nytelife26:
> I used [MangoHud](https://github.com/flightlessmango/MangoHud) to monitor the FPS and ran `cat /dev/urandom` with a minuscule font size to compare the performance

It appears that the preferred method (according to [a blog post](https://jwilm.io/blog/alacritty-lands-scrollback) is to use vtebench and `time cat <file>`. I'll likely be using that. Anyways, at the moment I'm currently having to rebase the PR to fix merge conflicts.

Is there any chance you could add me to your fork so I can work on the PR, or should I create my own?

chrisduerr:
The recommended way to test PTY performance is by using the ***latest*** version of [`vtebench`](https://github.com/alacritty/vtebench).

However I'd expect anyone trying to seriously verify the performance of any ligature PR to create supplementary benchmarks that stress specific worst-case scenarios. If you're struggling to come up with these, chances are you're not gonna get very far when it comes to providing a performant implementation either.

Nytelife26:
> However I'd expect anyone trying to seriously verify the performance of any ligature PR to create supplementary benchmarks that stress specific worst-case scenarios.

Yeah, I get you. I'll do an additional formal series of tests, likely consisting of the following scenarios.

- Ligatures under the cursor
- Extended series of ligatures
- Printing a reasonably sized source file with and without ligatures
- Ligatures and colour boundaries combined 

If you have any other suggestions, let me know, but those are the ones off of the top of my head. Should cover a lot of fairly common situations. For a more extreme stress test, I'll get it to stream out a random long block of text as fast as possible containing ligatures and then compare it to without.

chrisduerr:
> Edit: I also note that there are no benchmarks in the source code. It would probably be beneficial to add some for more crucial internal functions. That way you can track regressions over time.

We intentionally have no direct Rust benchmarks since they're mostly useless and do not appropriately represent the performance of the actual application.

fee1-dead:
I tested that commit out and compared it with master. vtebench shows similar results, framerates look on par. I'm not going to invest my time into rebasing because I'm not sure if the efforts will be worth it. Thanks @zenixls2 though.

fee1-dead:
To be clear: I was basing my judgement on "Ligatures are not worth dropping a single frame for." I don't know if that is an exaggeration or more serious.

It is unclear how small should performance regressions be, (or whether they should not exist at all), and as to how performance should be measured (by which specific benchmark programs?) for the ligatures feature to be merged.

Until a response for that is made, I will be slightly more convinced that the alacritty terminal does not care about attending to the community's needs, and just want to clench to their own definition of "performance" instead.

RayZ0rr:
I think providing a description such as,  (above the ligature toggling option)  : 

**"Ligature support is currently an experimental feature. Enabling it may lead to performance issues."**

will satisfy both alacrity's reasoning behind feature implementations (which completely doesn't apply to this case because it's experimental) as well as satisfy the community, who have been waiting for ages. We won't have to switch to wezterm or kitty just to get font ligatures which seem rather sad.

kittydoor:
An experimental option in the config, with no promises to maintain except for the intent to move to core functionality if it meets the core requirements with further development, would be the ideal compromise for me as well. Then as the user I can make the choice of what my priorities at that moment are.

chrisduerr:
You cannot add anything to Alacritty without putting a maintenance burden on maintainers. That's just not how it works. And again feature will not be merged until the requirements are met, that's why they're **requirements**.

If you think it is sustainable to maintain a ligature implementation, then feel free to do so in a separate fork (as some users have already done). That way the maintenance is not put on maintainers.

kittydoor:
I'm not pretending it comes for free, and zero extra maintaince burden. I just believe that it might be an acceptable compromise with this and sixel or kitty support, so that people like me who have loved alacritty for so long but their needs have grown over time don't need to abandon ship to get these new needs met.

AlexApps99:
I see a lot of people in this PR and others, posting comments and reactions pressuring the maintainers to merge incomplete features. Here's my 2 cents, addressed broadly to those who want these features merged.

Consider what would happen if an incomplete feature was added to alacritty, and it was never finished by the original creator:

1. The maintainers could finish the feature themselves
    (this would be a massive maintenance burden, completely unfair on the maintainers)
2. The feature could be removed until it was finished by someone else
  (users would kick up a big fuss on the issue tracker, and configs would break)
3. It could be kept experimental indefinitely until finished
  (feature would be buggy and incomplete, code would rot over time, maintainers would have difficulty refactoring and optimizing related code)

There is no way to predict when a feature will be finished, or when a PR author may step away and abandon/postpone work.

As you can see, none of these options are a good choice, which is why I can totally respect the maintainers for not merging these features until they are in a finished state. Installing a fork is much easier than forcing the maintainers to deal with the potential fallout of an incomplete feature.

Please have some empathy for the maintainers, and if you want a feature to be finished, consider finding ways to contribute. Asking for a feature to be finished or merged doesn't move it any further along, it's just a burden for everyone following.

fee1-dead:
I beg to differ. My personal experience with the FOSS model is that incomplete features attract contributors the most, while *completely unimplemented features* are comparably harder. When I started with contributing to Rust (the compiler), I was mainly picking up a feature work that has been incomplete for more than six month. The feature wasn't even an accepted RFC. Am I implementing the feature to introduce more burden on other compiler contributors? Of course not.

> feature would be buggy and incomplete, code would rot over time, maintainers would have difficulty refactoring and optimizing related code

The main concern here is, is Alacritty finished? Is it a complete project already? A useful metric would be how many new features have been added this year. Does alacritty have a roadmap? Are there things worth contributing that can have a realistically significant impact on the end users? I'm not suggesting "just add ligatures already," but rather to articulate some of my observations.

So why is it so hard to maintain ligatures in the code base? The actual code for adding the feature is not too complicated, but the whole project requires refactoring in order to maintain the performance. Is the code going to rot if it eventually get merged? It could, but it wouldn't consist of bugs given how many people have volunteered to get this feature only to find out that it isn't that easy.

kchibisov:
@fee1-dead see https://github.com/alacritty/alacritty/milestone/2

fee1-dead:
@hughesjs: see [this](https://github.com/alacritty/alacritty/pull/5615#issuecomment-997227586). Shaping with harfbuzz is obviously going to reduce the framerate when doing `cat /dev/urandom`. For other benchmarks, I did not find a significant difference between shaping vs. no shaping. There might still be framerate issues when the program outputs a rapid stream of text that are not random, but I still have trouble understanding why the latency in nanoseconds per line is so important.

kchibisov:
>Could we get some sort of meaningful measure of what the actual drop in performance is? I've seen "2x" mentioned above but that's sort of meaningless without knowing what the baseline is. Two times fuck all latency is still fuck all latency.

If you redraw the entire grid the perf on all branches were dropping from 2ms to ~30ms, which is just unusable, given that you clearly start missing frames. The terminal parsing performance thus will increase, because renederer will be very slow.

And that was on a decent hardware, on my phone(PPP) I bet all of that ligatures thing will render frame not in 5ms, but in like 60ms, just because you have poor CPU + GPU.

And be aware, that's a massive hit. I'm not saying that it shouldn't have any hit at all, I'm saying that the hit from ligatures should still keep alacritty renderer at 60fps on low powered hardware, and on ~200 fps on decent one.

foxyseta:
Any updates on this?

luisdavim:
did someone try to rebase this run `vtebench` and see the performance difference?

luisdavim:
Is the impact actually noticeable by a human?
Since this is such a requested feature would it be that bad to have a small performance hit when this is enabled? If the person enabling it would be willing to take it what would be the problem?
And the feature could be iterated on and improved over time.

kchibisov:
>bad to have a small performance hit whe

The impact is not small, the slowdown was around 2-12x on some of my systems compared to current renderer, and on slow hardware we also target the experience is no longer pleasant. There're approaches to do the ligatures faster, they just require effort which one should put into.

artem-nefedov:
I started using Alacritty **because** it puts minimalism and performance above all else, not in spite of it. Otherwise, if it the project went for feature parity instead, it would just become "yet another terminal emulator no.78", and then what is even the point?

Furthermore, there's no such thing as "unnoticeable" performance degradation for something that runs on end user's PC. Everyone's environment is different, and at least someone will definitely notice it.

So, I don't understand why so many people give Chris bad rep for not merging this, despite him just following the core design philosophy of the project by doing so. That being said, I do think contribution guidelines would benefit from making those things more clear.

hughesjs:
Oh here we go with the "speed at the cost of all else" chat again...

hughesjs:
Surely the answer is to just put this behind a config flag? Then users who want it can have it and those who don't don't...

PedroHLC:
Just my two cents:

Alacritty “misses” features I could always compensate for with other tools. Like using tmux instead of having tabs. Or editing the configuration in a file instead of GUI. No toolbars and context menus, but I could always use the keyboard.

I'm totally sure that these “missing” features are what sold Alacritty for me, making it blazingly fast and flexible.

But, I cannot compensate for the missing ligatures in any way.

foxyseta:
> But, I cannot compensate for the missing ligatures in any way.

This. Had to move to WezTerm just because of this.

hughesjs:
> > I started using Alacritty **because** it puts minimalism and performance above all else, not in spite of it. Otherwise, if it the project went for feature parity instead, it would just become "yet another terminal emulator no.78", and then what is even the point?
> > Furthermore, there's no such thing as "unnoticeable" performance degradation for something that runs on end user's PC. Everyone's environment is different, and at least someone will definitely notice it.
> > So, I don't understand why so many people give Chris bad rep for not merging this, despite him just following the core design philosophy of the project by doing so. That being said, I do think contribution guidelines would benefit from making those things more clear.
> 
> +1
> 
> I use alacritty because it is the fastest terminal emulator for me & because it is so simple (configured via toml). There are features I miss (& especially missed when I first started using it) that I really wish could be added. But those features are probably different to a what other people want. If everyone could add all the features they wanted, then it would just be another terminal emulator with some features missing instead of being a terminal emulator whose main feature is performance, which I don't think any other terminal emulator can provide. So, people who love alacritty, love it because of the speed & simplicity, not in spite of it. Regarding putting this behind feature flags, as a user you'll hear no quarrel from me. But as a dev, I imagine it would be nightmare to start adding features behind flags & working to ensure that all possible feature combinations work. You or me could add a few features & call it a day, but the devs have to support it for life.

Not everyone who loves alacritty loves it because of its speed... I love alacritty, despite its use of ligatures, because of its simple config and the fact that it just works really well. I honestly couldn't care less about speed as long as it doesn't become a laggy mess.

utkarshgupta137:
> > But, I cannot compensate for the missing ligatures in any way.

I sometimes miss ligatures too, & I would probably enable them if I couldn't notice a difference in smoothness. I agree it is an irreplaceable feature & table stakes for some users.

> This. Had to move to WezTerm just because of this.

WezTerm is a pretty nice terminal emulator, I used to use it. I'm not sure why you're unsatisfied with WezTerm or rather, why you're still considering Alacritty if WezTerm has all the features you want. My point is that different terms have different tradeoffs & it is probably better to for WezTerm to be a feature-rich term for people who want a lot of different features & Alacritty to be a feature-slim term for people who like the simplicity (& the smoothness that comes with it).

> > I use alacritty because it is the fastest terminal emulator for me & because it is so simple (configured via toml). There are features I miss (& especially missed when I first started using it) that I really wish could be added. But those features are probably different to a what other people want. If everyone could add all the features they wanted, then it would just be another terminal emulator with some features missing instead of being a terminal emulator whose main feature is performance, which I don't think any other terminal emulator can provide. So, people who love alacritty, love it because of the speed & simplicity, not in spite of it. Regarding putting this behind feature flags, as a user you'll hear no quarrel from me. But as a dev, I imagine it would be nightmare to start adding features behind flags & working to ensure that all possible feature combinations work. You or me could add a few features & call it a day, but the devs have to support it for life.
> 
> Not everyone who loves alacritty loves it because of its speed... I love alacritty, despite its use of ligatures, because of its simple config and the fact that it just works really well. I honestly couldn't care less about speed as long as it doesn't become a laggy mess. Now, you're within your rights to disagree, hence why I've suggested a config flag. This isn't just everyone adding all the features they want: this is probably the single most requested feature and one that stops a hell of a lot of people from using alacritty.
> 
> I don't use it anymore if I'm having to do any work in nvim precisely due to the lack of ligatures.

TBC, by speed I meant "smoothness". For me, I had to switch from WezTerm because the cursor was very jerky in neovim (as of 6 months ago at least) (I tend to over use j/k & I've set KeyRepeat on macOS to 15ms).

Again, I've no issue as a user with other people having more choice. I was just saying that I can understand why the devs might not want to support it via feature flags.

hughesjs:
@utkarshgupta137 - I would agree with you if this was a huge amount of effort to implement and maintain, but this PR is a fairly simple couple of hundred lines of code... The maintenance burden wouldn't be large

kchibisov:
I mean, when performance drops to 30fps from ~350fps surely you'll notice difference, which was the case when I tested since my fps dropped like 2-12 times as I said before.

hughesjs:
> I mean, when performance drops to 30fps from ~350fps surely you'll notice difference, which was the case when I tested since my fps dropped like 2-12 times as I said before.

On what hardware? Also, I refer back to my point about config flags... If a user chooses to accept a performance drop, that's fine.

kchibisov:
>On what hardware? Also, I refer back to my point about config flags... If a user chooses to accept a performance drop, that's fine.

I mean, that doesn't matter, such drop in fps is not an option, regardless of option. Like if you can't render at display refresh rate you shouldn't even provide an option, which is the point.

No one forces anyone to use alacritty in the first place if it doesn't meet your desires, you have plenty of terminals around.

hughesjs:
> > On what hardware? Also, I refer back to my point about config flags... If a user chooses to accept a performance drop, that's fine.
> 
> I mean, that doesn't matter, such drop in fps is not an option, regardless of option. Like if you can't render at display refresh rate you shouldn't even provide an option, which is the point.
> 
> No one forces anyone to use alacritty in the first place if it doesn't meet your desires, you have plenty of terminals around.

"If you can't render at display refresh rate" is almost certainly tied to what hardware you're on. I haven't got the time to build this branch and find out, but I would be hugely surprised if my hardware can't render ligatures at well over 100fps.

chrisduerr:
> If a user chooses to accept a performance drop, that's fine.

This is never fine. Performance should never be opt-in or opt-out. Running Alacritty should be fast no matter what you do. A user should not need knowledge of Alacritty's internals to know which options he's allowed to use to avoid performance pitfalls.

hughesjs:
> > If a user chooses to accept a performance drop, that's fine.
> 
> This is never fine. Performance should never be opt-in or opt-out. Running Alacritty should be fast no matter what you do. A user should not need knowledge of Alacritty's internals to know which options he's allowed to use to avoid performance pitfalls.

Well then you can either make it a compile-time config option so advanced users can use it, or just disable it and emit a warning if someone has it enabled and it's dropping frames.

Whilst I don't agree with your mantra of performance above all else, it's your project and you're well within your rights to take that approach. However, I think there are reasonable ways around this and I don't feel as though the project maintainers are actually trying to find a solution that works for both sides of this discussion.

kchibisov:
>Well then you can either make it a compile-time config option so advanced users can use it, or just disable it and emit a warning if someone has it enabled and it's dropping frames.

Compile time option could be achieved by complying the patch you want. Adding disabled by default code upstream is not a great idea.

hughesjs:
> Yeah because detecting which feature of Alacritty might be causing slow downs just so we can support some slow garbage feature nobody needs is a great investment of time.

Dude, it's objectively not a garbage feature when clearly many, many people want it. 

> I don't care about your side of the discussion. If you disagree, you can either implement it yourself or use a different terminal emulator. I have zero interest in compromising if I have nothing to gain but pain.

Like this is fair enough, although we're having this discussion on a PR, so someone clearly has implemented it. So my question then becomes what objective measure needs to be met where you would actually consider merging ligature support? Obviously, I've not read the entire thread on this because it's huge, so sorry if it's already been said, but literally no performance loss is pretty much impossible when adding any feature. So what's the benchmark? It's fast enough that we don't drop any frames on modern hardware? 10 year old hardware?

I'd be willing to have a crack at this when I get the time if I knew what the goalposts were and that if I could get them met then it would be merged.

> Compile time option could be achieved by complying the patch you want. Adding disabled by default code upstream is not a great idea.

I don't know if that's true, many, many common linux tools have code where you selectively compile what features you want. Fair enough if you don't want to go down that route though, I could fairly easily make an AUR entry with this patch (if someone hasn't already).

kchibisov:
>I'd be willing to have a crack at this when I get the time if I knew what the goalposts were and that if I could get them met then it would be merged.

The throughput should stay the same, frame rate could drop only slightly (a *few frames*) when the shaping is actually performed, and the input lag should not change at all.

It also should work just fine on low end CPUs, like arm ones found in pine64 hardware where CPUs are _really slow_.

In general, this issue requires a lot of effort in making it fast.

Be aware that in environments like browsers where massive blocks of text are static it's not really an issue, because this content is shaped once, but in terminal where everything can change at any time, it's more complex.

Like I know how to make it the same perf (in margin of error) when the user just edits text (and it has a lot of exceptions here as well, because text overlapping). But when your whole screen is redrawn it gets _way more_ complicated.

hughesjs:
Would a solution whereby ligatures are only drawn when the screen has been static for a small amount of time be acceptable (on the order of ms)? I don't think anyone would mind too much if ligatures aren't drawn when you've got huge amounts of text scrolling and the screen is being constantly redrawn. That might alleviate some of the performance concerns?

kchibisov:
>Would a solution whereby ligatures are only drawn when the screen has been static for a small amount of time be acceptable (on the order of ms)?

This is just poor UX, at this point just use conceal in vim if you really want them.

hughesjs:
Is it? I would hope you could get it working so that any delay after scrolling was essentially imperceptible. I don't know about you but when I have enough text scrolling through my terminal that performance becomes an issue, I can't really follow what's happening with any level of clarity. Maybe I'm underestimating how much throughput would cause a problem though

fee1-dead:
> some slow garbage feature nobody needs

I think it is important to remind people in this discussion that ligatures isn't just some fancy eye candy for programmers. Many languages **require** ligatures in order to display correctly. See [this](https://en.wikipedia.org/wiki/Complex_text_layout) for an overview of this topic. Many other issues would be resolved with ligatures (#936 #3975 #4025 just to mention a few) but the maintainers simply thinks having full unicode support is "garbage".

Of course, why care about internationalization at all when your only goal is to cling on to the performance? Sorry, you guys can keep on talking about "blazingly fast software" and "fastest terminal emulator" but I'm never going to use it if it has literally inferior font rendering.

kchibisov:
#3975 can't be done without breaking clients, needs negotiation, has nothing to do with ligatures, just shaping.

#936 breaks width as well, can't be done without negotiation with client.

#4025 similar.

All you've linked has nothing to do with ligatures though, ligatures preserve the width, the modifiers don't preserve the width, it's a completely different issue and fixing ligatures won't fix any of that, but ofc you know better.

fee1-dead:
Why do both maintainers like doing things like calling other developers incompetent and saying things like "ofc you know better"? I used the phrase 'ligatures' with an intention for it to mean text shaping, and that is what is done in this PR. If you're going to come up with a way to implement ligatures without getting the benefits of text shaping, you can feel free to go for it. I have not looked at this code for more than a year, and if the width issue exists, it should still be an easy fix.

But really who cares about an insignificant technicality? From my point of view the amount of toxicity displayed is considerably off-putting for any FOSS project, let alone a Rust one.

kchibisov:
>but the maintainers simply thinks having full unicode support is "garbage".

>Why do both maintainers like doing things like calling other developers incompetent and saying things like "ofc you know better"? 

You decided speak for us, you've got reply based on what you wrote. If it offended you, then sorry, but I'm not sure what you've expected from commenting that way. If you disagree you can just _move on_.

>If you're going to come up with a way to implement ligatures without getting the benefits of text shaping, you can feel free to go for it. 

The point is that text shaping similar to what ligatures are doing only works for input that doesn't change width. And I'm not familiar with anything other than ligatures which preserve width. Also, none of the PRs I've tested with ligatures made country flags work, though I don't remember testing it on this PR, since I closed it after massive perf loss.

>and if the width issue exists, it should still be an easy fix.

Width between client and terminal, it's not a rendering sort of thing (unless you generally want to center glyphs within an area they're present). There're some term specific escapes which do negotiation on how both terminal and client calculate them, it's not an easy fix as you say.

>But really who cares about an insignificant technicality?

width is clearly **not** insignificant, it makes the tui app unusable if you do mismatch, because the cursor can't be positioned reliably anymore and you have whole screen of artifacts.

>From my point of view the amount of toxicity displayed is considerably off-putting for any FOSS project, let alone a Rust one.

Yes, not willing to accept every patch in open source is very toxic, I understand, and also getting constant spam on issue telling us how bad we are that we don't accept every patch we don't like.

And when it comes to ligatures PRs it's like the 4-th? iteration of nearly the same code, what would you expect where all solution are basically the same as the first suggested one and which had _proper_ reviews, etc?

kchibisov:
And to be clear, I've rechecked the issue PR and it's 4x slower and none of the linked issues (which are not ligatures) are handled, so flags are broken the same way, etc. Only said eye candy is rendered now.

hughesjs:
Putting the above comments about toxicity aside, as I think everyone involved here could benefit from taking some of the heat out of this discussion.

Are there a script anywhere that runs some objective performance tests that we can use to set a goal? Perhaps, we could define a VM that's suitably specced for a reproducible run?

kchibisov:
>I don't think you got what I meant at all.

I've got what you mean, but behavior on every ligature relating issue is toxic from both sides from the start, so not sure what you really wanted. And for me your behavior was toxic from the start, it's all subjective in the end, especially when you directly starting to call out.

And some necrobumping.

>Sorry, I retract my last sentence on that comment. I did come up with a way without regressions and that is to completely rewrite the grid storage so that it works well with ligatures (each input will only cause reshape of individual lines). It will require the effort of rewriting the entire project and is going to conflict with every other major PR.

I think it was even possible back then when you wrote it, since you have damage information through out the terminal, and you can shape the individual lines, and even do partial redraws (I have PR around to do that, which could resurected). Which is the reason why I said now that I know how to do it for regular common case to be nearly the same as master, but not for full grid change, when e.g. scrolling.

It's not like you wanted to discuss the implementation (though, I wasn't around that time due to _issues_, so maybe @chrisduerr remembers you comming to irc and asking something).


Besides.

It's not like we're not interested at adding shaping at all, it's just requires _work_ and probably communicating with us, and be aware that alacritty is just _one of the projects_ we're doing in our free time, besides full time jobs and other issue. If you have challenges during implementation you can ask on irc (I generally reply to anyone, and you can even make me write patch for you if it's damage/partial rendering related, since I'm interested in that). But just claiming that you know how, but it'll require a rewrite and you won't do that is not productive, for both you and us. Maybe an idea to just make a rewrite of something internally besides the ligatures is not bad in the end, given that some parts of our grid code is just garbage.

And it's the reason why @chrisduerr said that 

>I strongly recommend anyone who's not willing to spend some time optimizing the ligature implementation to please stay away from sending PRs related to it. That's really just wasting the time of everyone involved.

kchibisov:
>Is there a script anywhere that runs some objective performance tests that we can use to set a goal?

If you touch the terminal related code (the part behind the mutex in display/mod.rs) vtebench is enough, however for rendering we don't have great infra other than disabling internal frame throttling and using things like `GALLIUM_HUD=fps` (which may not always be representative), you can also profile rendering time by measuring time you waste to produce a frame (similar to what our render timer is doing (it accounts for GL calls mostly, so could be not that accurate, but it gives some _base_).

fee1-dead:
> I've got what you mean, but behavior on every ligature relating issue is toxic from both sides from the start, so not sure what you really wanted. And for me your behavior was toxic from the start, it's all subjective in the end, especially when you directly starting to call out.

I'm sorry about the confrontational tone in some of my last comments. I will try my best to be more productive here.

I think a lot of the optimizations can be done by reducing the amount of data we feed to harfbuzz. I've read into the documentation again, and the [cluster](https://harfbuzz.github.io/clusters.html#clusters-and-shaping) information (provided to us through [GlyphInfo](https://docs.rs/harfbuzz_rs/2.0.1/harfbuzz_rs/struct.GlyphInfo.html) when shaping through `harfbuzz_rs`) can be very useful. I'm thinking some of the following things:

1. when resizing, we can look at the clusters to determine if any of the ligatures were broken up. If they were (if the line breaks a sequence with the same cluster value), we reshape, if not, we simply reuse the existing information. (might also need to take into account the [`unsafe_to_break`](https://docs.rs/harfbuzz_rs/2.0.1/harfbuzz_rs/struct.GlyphFlags.html#method.unsafe_to_break) flag)
2. when a specific location's text changed, we might simply be able to only use the neighboring clusters for partial reshaping. I'm not _entirely_ sure about whether this is acceptable, because I haven't worked closely with shaping code much, but the last few paragraphs on [this page](https://harfbuzz.github.io/adding-text-to-the-buffer.html) seems to suggest that is is okay, as long as we supply it with context. 
3. Caching a [shaping plan](https://harfbuzz.github.io/shaping-plans-and-caching.html) for the different font faces to avoid the planning cost for each shape run. The safe wrapper doesn't seem to support this, so `harfbuzz_sys` should probably be used.

hughesjs:
> > Is there a script anywhere that runs some objective performance tests that we can use to set a goal?
> 
> If you touch the terminal related code (the part behind the mutex in display/mod.rs) vtebench is enough, however for rendering we don't have great infra other than disabling internal frame throttling and using things like `GALLIUM_HUD=fps` (which may not always be representative), you can also profile rendering time by measuring time you waste to produce a frame (similar to what our render timer is doing (it accounts for GL calls mostly, so could be not that accurate, but it gives some _base_).

Maybe this would be a useful starting point for everyone then, not just this issue. Performance is clearly very important to you, but it's also quite a nebulous thing in a terminal. So, perhaps if we work together to come up with an easily reproduceable and rerunnable testbench, you can set it as an AC on all PRs that everyone can understand.

You up for that? If so, I can create a separate issue for it and take that discussion over there and I *might* even have a chance to look at it over the christmas break.

kchibisov:
>when resizing, we can look at the clusters to determine if any of the ligatures were broken up. If they were (if the line breaks a sequence with the same cluster value), we reshape, if not, we simply reuse the existing information. (might also need to take into account the [unsafe_to_break](https://docs.rs/harfbuzz_rs/2.0.1/harfbuzz_rs/struct.GlyphFlags.html#method.unsafe_to_break) flag)

Resize could actually be fairly complex to optimize (and not sure whether it's worth it)l, given that with reflow involved grid could require attaching rendering metadata to it (but it's separate from the rendering), and in general it could require grid rewrite (it's not reallyl a bad thing, given that the current structure struggles a bit), since
we could benefit from metadata as well to allow users attach content they want to _reflow_.

>when a specific location's text changed, we might simply be able to only use the neighboring clusters for partial reshaping. I'm not entirely sure about whether this is acceptable, because I haven't worked closely with shaping code much, but the last few paragraphs on [this page](https://harfbuzz.github.io/adding-text-to-the-buffer.html) seems to suggest that is is okay, as long as we supply it with context.

This is all tracked through out alacritty/alacritty_terminal now (was before, but now with a bit better abstractions), so that thing sounds possible and what I was thinking of doing. Like we have information on what _roughly_ changed within each line, so we can try to reshape based on that data, which will reduce the amount we're wasting by a _lot_. The challenge is to _store_ clusters somehow, but I think it's a sort of thing you get only when you try to plumb it. 


My main issue is how to deal with full-screen updates and scrolling.

Probably scrolling could be handled with damage information we already have if we try to rotate it inside the viewport, surely we still need to re-damage all the frame, but the information on what was changed inside the lines could be more or less accurate here (less accurate means overdamaging).

As for full-screen updates I don't have an answer, since I never tried shaping and don't know common patterns. Maybe something like `Trie` could help, but I'm not sure that it could actually work here, other than probably clusters have a common start, etc. And I'm also not familiar with harfbuzz much, since ENOTIME to look myself...

kchibisov:
>As for full-screen updates I don't have an answer, since I never tried shaping and don't know common patterns. Maybe something like Trie could help, but I'm not sure that it could actually work here, other than probably clusters have a common start, etc. And I'm also not familiar with harfbuzz much, since ENOTIME to look myself...

Seems like clusters doing something similar if I read the links correctly, at least they can sort of determine the mappings.

kchibisov:
Oh, and besides what was said.

I think the good way to approach that issue would be to just start outside of alacritty and try to shape grid (basically `Vec<Vec<char>>` like we have) of text with various techniques and compare perf when just using freetype vs freetype + harfbuzz shaping. And do that for different patterns of grid updates. The rendering could be omitted all together, since it's not relevant for what we're trying to benchmark, and drawing a little bit larger glyph shouldn't be a problem.

That should be more productive or at least educational, since when doing directly inside alacritty it'll likely just prevent from any progress and one could start fighting the codebase. And as for maintainer it'll really help to see **why** particular approach was used, when you have benchmarks to show it's easier to convince people.

Doing it separately could also benefit others, since they can pick up/play with what was discovered.

hughesjs:
> we already have https://github.com/alacritty/vtebench and https://github.com/alacritty/termbenchbot (and yes, you can use it to trigger perf report on alaritty PRs).

Does it run them on virtual hardware approximating the lowest-power device you'd like to be able to run at full speed? I think that's the key here

kchibisov:
>Does it run them on virtual hardware approximating the lowest-power device you'd like to be able to run at full speed? I think that's the key here

The point is delta, it's all on real hardware, since hardware acceleration, but the system is powered on, run this benchmark, and powers back off. It's sole purpose is to run the benchmarks.

hughesjs:
> > Does it run them on virtual hardware approximating the lowest-power device you'd like to be able to run at full speed? I think that's the key here
> 
> The point is delta, it's all on real hardware, since hardware acceleration, but the system is powered on, run this benchmark, and powers back off. It's sole purpose is to run the benchmarks.

Ahh fair, that'll probably do... I was just thinking less modern hardware might be hit harder than more mordern hardware for certain feature changes, so the delta might not be linear

chrisduerr:
> I was just thinking less modern hardware might be hit harder than more mordern hardware for certain feature changes, so the delta might not be linear

That's why it's running an ancient GPU.

fee1-dead:
> in general it could require grid rewrite

and supporting text shaping where the glyphs' widths change, as you said, would probably require a rewrite as well? But that should probably be after harfbuzz is introduced to the codebase first, right?

> 

> try to shape grid (basically `Vec<Vec<char>>` like we have) of text with various techniques and compare perf when just using freetype vs freetype + harfbuzz shaping. And do that for different patterns of grid updates

I'm not sure if I understand what you meant here. Does `alacritty`/`alacritty_terminal` supply grid updates as some sort of `Event` enum that tells you what needs to be changed?

kchibisov:
>and supporting text shaping where the glyphs' widths change, as you said, would probably require a rewrite as well? But that should probably be after harfbuzz is introduced to the codebase first, right?

They just can't be supported, regardless of grid, because you'll break client width as well, thus the entire screen will be unreadable. There're experimental escapes to negotiate width, but that's about it.

>Does alacritty/alacritty_terminal supply grid updates as some sort of Event enum that tells you what needs to be changed?

It basically gives iterator and we also have a `Vec<LineDamage>` telling what changed in lines, it's at least enough to understand which lines changed.

ChanceNCounter:
Yikes. Maintainers’ uncompromising attitude in the face of overwhelming user demand - yes, the lines count - and the yearlong mistreatment of a would-be contributor, are beyond the pale. I thought I was a user for life, but right after I hit submit on this comment, I’m off to pick another terminal emulator.

Be better.

littleblack111:
any updates?

PR #5699, opened 2021-12-18
Fix CreateNewWindow CLI fallback

### LINKED ISSUE #5659: CreateNewWindow on macos sometimes doesn't start shell correctly
For bug reports, the following information can help speed up the process. Please
describe the bug that you have found and what you would expect to happen
instead.

I have my shell configured:
```
shell:
  program: /bin/zsh
  args:
    - -l
    - -c
    - "tmux attach || tmux"
```

When I start alacritty, then immediately create a new window, the new window is a bare zsh shell, without tmux.

If I then close the second window and create another new one, it correctly runs tmux.

Subsequent new windows seem to randomly either start tmux or not.

### System

OS: macos
Version: alacritty 0.10.0-dev (a1083d18)
Linux/BSD: n/a

### Logs

Crashes: STDERR, STDOUT
```
➤ alacritty -vv
Created log file at "/var/folders/jz/_9dm9vt91dd5jfj40l_3r1dh0000gn/T/Alacritty-86003.log"
[2021-12-04 23:02:43.032512000] [INFO ] [alacritty] Configuration files loaded from:
                                                      "/Users/joe/.config/alacritty/alacritty.yml"
[2021-12-04 23:02:43.033066000] [DEBUG] [alacritty] Using environment locale: en_US.UTF-8
[2021-12-04 23:02:43.068524000] [DEBUG] [alacritty] Estimated DPR: 2
[2021-12-04 23:02:43.068576000] [DEBUG] [alacritty] Estimated window size: Some(PhysicalSize { width: 1320, height: 840 })
[2021-12-04 23:02:43.068590000] [DEBUG] [alacritty] Estimated cell size: 14 x 30
[2021-12-04 23:02:43.111986000] [INFO ] [alacritty] Device pixel ratio: 2
[2021-12-04 23:02:43.113850000] [INFO ] [alacritty] Initializing glyph cache...
[2021-12-04 23:02:43.121498000] [INFO ] [alacritty] ... finished initializing glyph cache in 0.007592041s
[2021-12-04 23:02:43.121524000] [INFO ] [alacritty] Estimated DPR correctly, skipping resize
[2021-12-04 23:02:43.121539000] [INFO ] [alacritty] Cell size: 14 x 30
[2021-12-04 23:02:43.121548000] [INFO ] [alacritty] Padding: 100 x 60
[2021-12-04 23:02:43.121561000] [INFO ] [alacritty] Width: 1320, Height: 840
[2021-12-04 23:02:43.121661000] [INFO ] [alacritty] PTY dimensions: 24 x 80
[2021-12-04 23:02:43.123367000] [INFO ] [alacritty] Initialisation complete
[2021-12-04 23:02:43.165711000] [DEBUG] [alacritty_terminal] [Unhandled CSI] action='q', params=[0], intermediates=[62]
[2021-12-04 23:02:43.175403000] [DEBUG] [alacritty_terminal] [Unhandled CSI] action='h', params=[7727], intermediates=[63]

```
Keyboard and bindings: `alacritty --print-events`

chrisduerr:
The existing behavior for the new CreateNewWindow actions was to always
pass in their own options, which would discard the existing options
configured on the terminal's PTY config.

To fix this the behavior for CreateNewWindow is now the same as for the
initial window creation, the config values are overwritten conditionally
based on their individual presence in the CLI options.

However this also means that all temporary CLI options set on the
"master" Alacritty instance are inherited by all future windows.

Fixes #5659.

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -225,7 +225,7 @@ impl TerminalOptions {
             pty_config.shell = Some(command);
         }
 
-        pty_config.hold = self.hold;
+        pty_config.hold |= self.hold;
     }
 }
 
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -100,8 +100,6 @@ impl Options {
 
     /// Override configuration file with options from the CLI.
     pub fn override_config(&self, config: &mut UiConfig) {
-        self.terminal_options.override_pty_config(&mut config.terminal_config.pty_config);
-
         if let Some(title) = self.title.clone() {
             config.window.title = title
         }
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -100,20 +100,7 @@ impl Options {
 
     /// Override configuration file with options from the CLI.
     pub fn override_config(&self, config: &mut UiConfig) {
-        if let Some(working_directory) = &self.terminal_options.working_directory {
-            if working_directory.is_dir() {
-                config.terminal_config.pty_config.working_directory =
-                    Some(working_directory.to_owned());
-            } else {
-                error!("Invalid working directory: {:?}", working_directory);
-            }
-        }
-
-        if let Some(command) = self.terminal_options.command() {
-            config.terminal_config.pty_config.shell = Some(command);
-        }
-
-        config.terminal_config.pty_config.hold = self.terminal_options.hold;
+        self.terminal_options.override_pty_config(&mut config.terminal_config.pty_config);
 
         if let Some(title) = self.title.clone() {
             config.window.title = title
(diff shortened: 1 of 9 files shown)


kchibisov (changes requested):
I was thinking whether to do so or not when implementing it. Right now I think that we should inherit options.

>However this also means that all temporary CLI options set on the
"master" Alacritty instance are inherited by all future windows.]

Yeah, but this isn't working right now on your PR with `hold` flag. I'm not sure how we should handle them and overwriting options. Initially I've used `Option` to indicate that nothing was provided. One solution would be to wrap everything on `TerminalOptions` in `Option`.

chrisduerr:
> Yeah, but this isn't working right now on your PR with hold flag. I'm not sure how we should handle them and overwriting options. Initially I've used Option to indicate that nothing was provided. One solution would be to wrap everything on TerminalOptions in Option.

That's a good point, but this is trivial to fix. `false` for structopt is just the same as `None`, since it's a flag. So no need to wrap the `hold` in an `Option`, clap also just really doesn't like that. False is the same here.

kchibisov:
I think the problem here with `hold` is a bit more complicated. The thing is that when you pass `hold` on a CLI to the main application there's no way to disable it anymore. This applies to every flag like `hold`. The solution could be to discard all boolean flags back to default values (like it was initially) or add extra flag like `no-hold` (and that would apply to every flag).

kchibisov:
>I'd say there are two reasonable options: Either we keep it the way it currently is, the base CLI flags are basically treated like config settings and inherited everywhere. The other option is to discard all CLI flags from the original client, then we'd probably have to store config/cli separately.

I think discarding for new windows (the way it originally was on this PR) is fine for such flag. Not all, but only boolean flags like hold, since we already have a `hold` flag to pass.

chrisduerr:
> I think discarding for new windows (the way it originally was on this PR) is fine for such flag. Not all, but only boolean flags like hold, since we already have a hold flag to pass.

Discarding only *some* flags seems rather terrible though. At that point there's no consistency for the user?

Also how do you suggest this be implemented, would you discard it completely, ignoring the configuration file option? Or store the CLI/Config stuff separately and only override the former?

kchibisov:
>Also how do you suggest this be implemented, would you discard it completely, ignoring the configuration file option? Or store the CLI/Config stuff separately and only override the former?

Use config option by default, unless CLI option is passed. if `--hold` was passed on CLI apply hold, however if you create a new window, don't apply hold to it, unless it's presented in a config or user explicitly specified `--hold` on CLI for the new window.

chrisduerr:
> Use config option by default, unless CLI option is passed. if --hold was passed on CLI apply hold, however if you create a new window, don't apply hold to it, unless it's presented in a config or user explicitly specified --hold on CLI for the new window.

Yeah, sounds reasonable. Basically treat a new window the same way we'd treat a new instance.

chrisduerr:
@kchibisov If you're around still do you mind testing to see if everything now works as you would expect?

kchibisov (approved):

chrisduerr:
Was kinda scared you just merged the PR with the original commit message, but turns out you're way ahead of me in that regard. Great work.

PR #5712, opened 2021-12-23
Alacritty version 0.10.0-rc1

chrisduerr:


kchibisov:
In general I'm not against starting the RC process, however I'd like to try to add support for passing `--class` and `--title` to `create-window`. If it'll be something simple, we could include in release as well. The thing about `title` and `class` is that you can run only one instance for like everything in your wm.

PR #5719, opened 2021-12-26
Add ability to pass 'title' and 'class' for 'create-window'.

kchibisov:
This adds ability to pass title and class over IPC via create-window
subcommand, so users can run only one instance for windows of different
spurposes in the window managers of their choice.


alacritty/src/window_context.rs
```diff
@@ -50,6 +51,8 @@ pub struct WindowContext {
     font_size: Size,
     mouse: Mouse,
     dirty: bool,
+    /// Initial information to identify the window such as default class and title.
```
chrisduerr, line 54:
```suggestion
```

None of the other fields are documented, so I think it's odd to explicitly document this one. Especially because it's not particularly difficult to understand.

I have to say though `WindowContext` is getting **big**, I guess that's to be expected though since it holds everything important for a window. Maybe breaking it down in the future into separate structs might make sense, but don't think that's what we should do now.

alacritty/src/window_context.rs
```diff
@@ -129,12 +134,14 @@ impl WindowContext {
         if config.terminal_config.cursor.style().blinking {
             event_proxy.send_event(TerminalEvent::CursorBlinkingChange.into());
         }
+        dbg!(identity);
```
chrisduerr, line 137:
```suggestion
```

alacritty/src/window_context.rs
```diff
@@ -184,8 +191,8 @@ impl WindowContext {
         }
 
         // Live title reload.
-        if !config.window.dynamic_title || old_config.window.title != config.window.title {
-            self.display.window.set_title(&config.window.title);
+        if !config.window.dynamic_title || self.identity.title != config.window.identity.title {
+            self.display.window.set_title(&config.window.identity.title);
         }
```
chrisduerr, line 196: Is it me or should this be `config.window.dynamic_title && ...`?

Also since `dynamic_title` is only ever activated when the options does **not** have a title passed, we should probably be able to use the old logic and not store the `identity` on the `WindowContext` unless we remove the override code from `override_config`?

chrisduerr (changes requested):

alacritty/src/cli.rs
```diff
@@ -71,9 +63,9 @@ pub struct Options {
     #[structopt(skip)]
     pub config_options: Value,
 
-    /// Terminal options which could be passed via IPC.
+    /// Options which could be passed via IPC.
```
chrisduerr, line 66:
```suggestion
    /// Options which can be passed via IPC.
```

alacritty/src/event.rs
```diff
@@ -1044,8 +1044,9 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                         }
                     },
                     TerminalEvent::ResetTitle => {
-                        if self.ctx.config.window.dynamic_title {
-                            self.ctx.display.window.set_title(&self.ctx.config.window.title);
+                        let window_config = &self.ctx.config.window;
+                        if window_config.dynamic_title {
```
chrisduerr, line 1048:
```suggestion
                        if window_config.dynamic_title {
                            let window_config = &self.ctx.config.window;
```

alacritty/src/main.rs
```diff
@@ -178,13 +178,13 @@ fn alacritty(options: Options) -> Result<(), String> {
     };
 
     // Event processor.
-    let cli_options = options.terminal_options.clone();
+    let new_window_options = options.window_options.clone();
```
chrisduerr, line 181:
```suggestion
    let window_options = options.window_options.clone();
```

alacritty/src/window_context.rs
```diff
@@ -184,8 +189,8 @@ impl WindowContext {
...
```
chrisduerr, line 194: If the CLI option is specified here and the config is changed, should we really change the title? Shouldn't the CLI parameter have precedence?

In that case we might also get away with using `old_config.window.title` so we don't have to store the field on `WindowContext`. Though we might have to store a flag for it?

chrisduerr (changes requested):

alacritty/src/event.rs
kchibisov: I don't see a reason to do so? I mean it's the same amount of code and you access window_config in `if` anyway, so you can extract it and use in 2 places instead of just one.

alacritty/src/window_context.rs
kchibisov: I think we could make title from command line to have precedence here. However to make this work we should check that title was actually passed on CLI(`!dynamic_title && identity.title != old_config.window.identity.title`). But this won't work when you pass title from config on CLI (which isn't something you'd actually do...).

If you want a flag to indicate whether option was from CLI(besides my idea), you need to pass CLI identity to `WindowContext` and then perform override right in `WindowContext::new` (I really don't want to pass one extra argument).

What solution would you prefer?

chrisduerr: > But this won't work when you pass title from config on CLI

Considering --option the same as changing it in the config is perfectly fine to me. That doesn't need to have precedence.

> I really don't want to pass one extra argument

Same.

alacritty/src/event.rs
chrisduerr: Didn't see that you actually use it in the conditional, thought it was just inside. All good. :+1:

alacritty/src/window_context.rs
```diff
@@ -183,9 +188,12 @@ impl WindowContext {
             self.display.pending_update.dirty = true;
         }
 
-        // Live title reload.
-        if !config.window.dynamic_title || old_config.window.title != config.window.title {
-            self.display.window.set_title(&config.window.title);
+        // Live title reload when title wasn't provided on command line.
+        if !config.window.dynamic_padding
```
chrisduerr, line 192:
```suggestion
        if !config.window.dynamic_title
```

alacritty/src/window_context.rs
```diff
@@ -183,9 +188,12 @@ impl WindowContext {
...
+            && self.identity.title != old_config.window.identity.title
```
chrisduerr, line 193: Not sure exactly what this is supposed to test?

alacritty/src/window_context.rs
```diff
@@ -183,9 +188,12 @@ impl WindowContext {
...
+            && (!config.window.dynamic_title || self.identity.title != config.window.identity.title)
```
chrisduerr, line 194: Not sure exactly what this is supposed to test?

chrisduerr (changes requested):

alacritty/src/window_context.rs
```diff
@@ -183,9 +188,12 @@ impl WindowContext {
...
         }
 
-        // Live title reload.
-        if !config.window.dynamic_title || old_config.window.title != config.window.title {
-            self.display.window.set_title(&config.window.title);
+        // Live title reload only when title wasn't provided on command line.
+        //
+        // We know that the title was provided on the command line when dynamic_title is disaabled
+        // and old_config title isn't equal to the current one.
+        if !config.window.dynamic_title && self.identity.title != old_config.window.identity.title {
```
chrisduerr, line 195: The `||` was in originally to allow changing the title when it was still the same as the original window title. Is this not something we want? I feel like it makes sense but ultimately I don't really care.

alacritty/src/window_context.rs
```diff
@@ -183,9 +188,12 @@ impl WindowContext {
             self.display.pending_update.dirty = true;
         }
 
-        // Live title reload.
-        if !config.window.dynamic_title || old_config.window.title != config.window.title {
-            self.display.window.set_title(&config.window.title);
+        // Live title reload only when title wasn't provided on command line.
```
chrisduerr, line 191:
```suggestion
        // Live title reload only when it wasn't provided on the command line.
```

alacritty/src/window_context.rs
chrisduerr, line 195: So when I have a title specified on the CLI and have the same in the config, it will overwrite it when I change the config?

chrisduerr (changes requested):

alacritty/src/window_context.rs
kchibisov: Yeah, that is the case when it will be overwritten anyway. If we want to handle it as well we should pass `Option<Idenetity>` and overwrite config right in the `WindowContext`.

chrisduerr: Your approach would be to just not bother because not worth? I'd be fine with it tbh.

kchibisov: I mean it should be changed if it was fully provided by the config. Since we compare with old title which should match the old config title in case title change isn't allowed. The case is that it doesn't only work when you pass title on CLI in other cases it'll work.

kchibisov: Yeah, pretty much it. I just override title when it isn't the one you passed on CLI which shouldn't be equal to default title.

alacritty/src/window_context.rs
```diff
@@ -183,9 +194,16 @@ impl WindowContext {
...
+        // Update title on config reload according to the following table.
+        //
+        // │cli │ dynamic_title │ current_title == old_config ││ set_title │
+        // │ Y  │       _       │              _              ││     N     │
+        // │ N  │       Y       │              Y              ││     Y     │
+        // │ N  │       Y       │              N              ││     N     │
+        // │ N  │       N       │              _              ││     Y     │
+
+        if !preserve_title && !config.window.dynamic_title {
+            self.display.window.set_title(&config.window.identity.title);
```
kchibisov, line 206: After looking into this one more time, I think there's no need to reload when `dynamic_title` and current title is the same as configs title. Since we start storing title somewhere(likely on the associated window, and each window will likely has its own title).

kchibisov: The table is left as you provided for now, will tune it later on if you insist that we should reload title in second case.

chrisduerr: > After looking into this one more time, I think there's no need to reload when dynamic_title and current title is the same as configs title.

Why not? Are you saying it's not worth the effort or are you saying that you don't think it would be the ideal behavior?

>  each window will likely has its own title

I'd say this is probably incorrect. There aren't *that* many applications setting custom titles that you're guaranteed to have all your terminals with a dynamic title actually with a dynamic title. Chances are the majority of your one-off terminals are all going to have the same title unless it's been overwritten from the CLI. In that case all of those will remain on the old title while your new one-off terminals will get the new one.

chrisduerr (commented):
See my comment.

alacritty/src/window_context.rs
kchibisov: Should be resolved now.

alacritty/src/window_context.rs
```diff
@@ -183,9 +193,19 @@ impl WindowContext {
...
-        if !config.window.dynamic_title || old_config.window.title != config.window.title {
-            self.display.window.set_title(&config.window.title);
+        // Update title on config reload according to the following table.
+        //
+        // │cli │ dynamic_title │ current_title == old_config ││ set_title │
+        // │ Y  │       _       │              _              ││     N     │
+        // │ N  │       Y       │              Y              ││     Y     │
+        // │ N  │       Y       │              N              ││     N     │
+        // │ N  │       N       │              _              ││     Y     │
+
```
chrisduerr, line 203:
```suggestion
```

chrisduerr (approved):
Should be good after applying the suggestion.

PR #5722, opened 2021-12-27
Fix trailing newline appending to a hints action argument

### LINKED ISSUE #5697: macos click url add "%0A" (newline)
macos click url add "%0A" (newline)

image

image

click this add "%0A"
```
https://open-erp.meituan.com/batchBinding/preBinding?businessId=30&charset=utf-8&timestamp=1639791432156&state=test&developerId=106416&sign=52f350b7fd4a78049d0c0784b4717cfc80a395f7
[nodemon] clean exit - waiting for changes before restart
```

a5ob7r:
This problem happens when a tail character of a hint matched string is on the last column. In the situation, Alacritty pushes a newline character to the string. This may be useful for Copy/Paste but not for command invocations. Command arguments generally don't contain a trailing newline character. So we should remove it for hint action argument.

Fixes #5697.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Strip trailing whitespaces when yanking from a block selection
 - Display area keeps history position when viewport is cleared
 - Commands spawn from the current directory of the foreground shell in Unix-like systems
-- Remove trailing newline from strings taken from hints or some selections
+- Remove trailing newline from strings taken from hints or simple/semantic selections
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Strip trailing whitespaces when yanking from a block selection
 - Display area keeps history position when viewport is cleared
 - Commands spawn from the current directory of the foreground shell in Unix-like systems
+- Remove trailing newline from strings taken from hints or some selections
 
 ### Fixed
 
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -367,10 +367,11 @@ impl<T> Term<T> {
                 res += self.line_to_string(end.line, start.column..end.column, true).trim_end();
             },
             Some(Selection { ty: SelectionType::Lines, .. }) => {
-                res = self.bounds_to_string(start, end);
-                res += "\n";
+                res = self.bounds_to_string(start, end) + "\n";
             },
-            _ => res = self.bounds_to_string(start, end),
+            _ => {
+                res = self.bounds_to_string(start, end);
+            },
         }
 
         Some(res)
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -17,7 +17,7 @@ use crate::config::Config;
 use crate::event::{Event, EventListener};
 use crate::grid::{Dimensions, Grid, GridIterator, Scroll};
 use crate::index::{self, Boundary, Column, Direction, Line, Point, Side};
-use crate::selection::{Selection, SelectionRange};
+use crate::selection::{Selection, SelectionRange, SelectionType};
 use crate::term::cell::{Cell, Flags, LineLength};
 use crate::term::color::{Colors, Rgb};
 use crate::vi_mode::{ViModeCursor, ViMotion};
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Minimum Rust version has been bumped to 1.56.0
 
+### Fixed
+
+- Remove trailing newline from a hint matched text as a hints action argument
```
chrisduerr, line 16: Could you add this to 0.10.0?

alacritty/src/event.rs
```diff
@@ -676,7 +676,12 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             HintAction::Command(command) => {
                 let text = self.terminal.bounds_to_string(*hint.bounds.start(), *hint.bounds.end());
                 let mut args = command.args().to_vec();
-                args.push(text);
+
+                match text.strip_suffix('\n') {
+                    Some(text) => args.push(text.to_string()),
+                    None => args.push(text),
+                }
```
chrisduerr, line 683:
```
let text = text.strip_suffix('\n').map(str::to_owned).unwrap_or(text);
```

There might be an even easier way to do it, but that's the best I could come up with.

chrisduerr (changes requested):

CHANGELOG.md
a5ob7r: Yes.

chrisduerr (approved):

chrisduerr (changes requested):
Actually I'd like to revoke my approval thinking about things again: The behavior should be the same for all actions.

In fact, the `bounds_to_string` function should probably **always** exclude trailing newlines. At least I'd expect selections to behave the same way.

a5ob7r:
> In fact, the bounds_to_string function should probably always exclude trailing newlines. At least I'd expect selections to behave the same way.

Oh, OK. I tried the `Select` hint action for the first time, and I think your request makes sense. By the way, does your `selections` mean just `Select` in hint actions? Do you intend to remove trailing newline character of vi mode selection?

chrisduerr:
> By the way, does your selections mean just Select in hint actions? Do you intend to remove trailing newline character of vi mode selection?

I mean everywhere. We should be consistent.

a5ob7r:
As a Vim user, I want to keep a newline character at the end of the last line only for lines selection if possible.

chrisduerr (changes requested):
Let's not over-complicate things. This works, no need to re-design Alacritty's selection copying when that part of the code clearly worked just fine.

It should still behave consistently for all hint actions though.

alacritty_terminal/src/term/mod.rs
```diff
@@ -348,24 +348,25 @@ impl<T> Term<T> {
     /// Convert the active selection to a String.
     pub fn selection_to_string(&self) -> Option<String> {
```
chrisduerr, line 349: Why did you decide to change the `selection_to_string` behavior now?

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
a5ob7r: The reason is that keep lines selection behavior even if the behaviors of simple/semantics are changed.

> This works, no need to re-design Alacritty's selection copying when that part of the code clearly worked just fine.

I interpreted it as no need to change not related selection behavior to hint action, but It makes sense that `semantics` selection's behavior also is changed because it is very similar to `simple` selection, which is related to `select` hint action.

`block` selection's behavior is also changed in this commit, but it is my mistake. I will revert to previous version.

alacritty_terminal/src/term/mod.rs
```diff
@@ -348,24 +348,29 @@ impl<T> Term<T> {
...
                 }
-            }
-            res += self.line_to_string(end.line, start.column..end.column, true).trim_end();
-        } else {
-            res = self.bounds_to_string(start, end);
+                res += self.line_to_string(end.line, start.column..end.column, true).trim_end();
+            },
+            Some(Selection { ty: SelectionType::Lines, .. }) => {
+                res = self.bounds_to_string(start, end);
+                res += "\n";
```
chrisduerr, line 371:
```suggestion
                res = self.bounds_to_string(start, end) + "\n";
```

Pretty sure this should just work.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Strip trailing whitespaces when yanking from a block selection
 - Display area keeps history position when viewport is cleared
 - Commands spawn from the current directory of the foreground shell in Unix-like systems
+- Remove trailing newline from strings taken from hints or some selections
```
chrisduerr, line 35: "some selections" sounds kinda sketchy. Just say semantic/normal and we are precise without being *much* more verbose.

alacritty_terminal/src/term/mod.rs
```diff
@@ -382,7 +387,7 @@ impl<T> Term<T> {
             res += &self.line_to_string(line, start_col..end_col, line == end.line);
         }
 
-        res
+        res.strip_suffix('\n').map(str::to_owned).unwrap_or(res)
```
chrisduerr, line 390: Why wouldn't we do this in the `line_to_string` function? Currently we add newlines conditionally there just to strip it here. Which seems rather… pointless.

Alacritty will never have `\n` in any cell, the only way there could ever be one is if it's inserted by `line_to_string` because of the last column.

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
chrisduerr: I don't mind the new behavior, just really confused me after my last comment.

alacritty_terminal/src/term/mod.rs
a5ob7r: Um... now Alacritty implicitly uses a trailing newline character of the string returned from `line_to_string` to indicate whether or not the range contains the last column or the end of a line is not wrapped. This is useful to join strings correctly in `bounds_to_string`, especially for wrapline joining. So we would need to treat wrapline by `bounds_to_string` instead of `line_to_string` if we trim a newline character by `line_to_string`. We need to lookup the grid contents to treat wrapline, but I think it is better to do it by `line_to_string` for abstraction.

chrisduerr: Oh yeah wrapline makes sense.

chrisduerr (approved):
LGTM

a5ob7r:
Thank you!

PR #5735, opened 2021-12-30
Fix text joining by block selection contained last column

a5ob7r:
This problem is that copied lines are joined when block selection contains the last column.

Assume a terminal grid like below.

```
aaa
bbb
```

When we select whole lines by block selection and copy it, expected copied text is `aaa\nbbb`, but actual one is `aaabbb`.

This is introduced by 8e584099.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -358,12 +358,9 @@ impl<T> Term<T> {
                     res += self
                         .line_to_string(line, start.column..end.column, start.column.0 != 0)
                         .trim_end();
-
-                    // If the last column is included, newline is appended automatically.
-                    if end.column != self.columns() - 1 {
-                        res += "\n";
-                    }
+                    res += "\n";
                 }
+
                 res += self.line_to_string(end.line, start.column..end.column, true).trim_end();
             },
             Some(Selection { ty: SelectionType::Lines, .. }) => {
@@ -2019,6 +2016,52 @@ mod tests {
     }
 
     #[test]
+    fn simple_selection_works() {
+        let size = SizeInfo::new(5., 5., 1.0, 1.0, 0.0, 0.0, false);
+        let mut term = Term::new(&Config::default(), size, ());
+        let grid = term.grid_mut();
+        for i in 0..4 {
+            if i == 1 {
+                continue;
+            }
+
+            grid[Line(i)][Column(0)].c = '"';
+
+            for j in 1..4 {
+                grid[Line(i)][Column(j)].c = 'a';
+            }
+
+            grid[Line(i)][Column(4)].c = '"';
+        }
+        grid[Line(2)][Column(0)].c = ' ';
+        grid[Line(2)][Column(4)].c = ' ';
+        grid[Line(2)][Column(4)].flags.insert(Flags::WRAPLINE);
+        grid[Line(3)][Column(0)].c = ' ';
+
+        // Multiple lines contain an empty line.
+        term.selection = Some(Selection::new(
+            SelectionType::Simple,
+            Point { line: Line(0), column: Column(0) },
+            Side::Left,
+        ));
+        if let Some(s) = term.selection.as_mut() {
+            s.update(Point { line: Line(2), column: Column(4) }, Side::Right);
+        }
+        assert_eq!(term.selection_to_string(), Some(String::from("\"aaa\"\n\n aaa ")));
+
+        // A wrapline.
+        term.selection = Some(Selection::new(
+            SelectionType::Simple,
+            Point { line: Line(2), column: Column(0) },
+            Side::Left,
+        ));
+        if let Some(s) = term.selection.as_mut() {
+            s.update(Point { line: Line(3), column: Column(4) }, Side::Right);
+        }
+        assert_eq!(term.selection_to_string(), Some(String::from(" aaa  aaa\"")));
+    }
+
+    #[test]
     fn semantic_selection_works() {
         let size = SizeInfo::new(5., 3., 1.0, 1.0, 0.0, 0.0, false);
         let mut term = Term::new(&Config::default(), size, ());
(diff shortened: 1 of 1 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -357,11 +357,7 @@ impl<T> Term<T> {
                 res += self
                     .line_to_string(line, start.column..end.column, start.column.0 != 0)
                     .trim_end();
-
-                // If the last column is included, newline is appended automatically.
-                if end.column != self.columns() - 1 {
-                    res += "\n";
-                }
+                res += "\n";
```
chrisduerr, line 360: We should probably revert to the old code and just trim whitespace?

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
a5ob7r: Oh, do you mean that we should trim whitespaces in `line_to_string` instead of `selection_to_string`? If so, is it for consistence among selections?

chrisduerr: No, I mean that with the change in 8e58409, we over-trimmed. Now we adjust that over-trimming by adding stuff again. That doesn't seem right to me. We should probably just not trim off stuff we don't want gone in the first place.

a5ob7r: Is non `over-trimming` like this?

```rust
// Trim just one newline and any number of other whitespace(except newline) on the end of line.
fn trim_end_weak(s: &str) -> String {
    let mut s = s.to_string();

    if s.ends_with('\n') {
        s.pop();
    }

    s.trim_end_matches(|c: char| c != '\n' && c.is_whitespace()).to_string()
}

#[test]
fn test_trim_end_weak() {
    assert_eq!(trim_end_weak("  hoge"), "  hoge".to_string());
    assert_eq!(trim_end_weak("  hoge\n"), "  hoge".to_string());
    assert_eq!(trim_end_weak("  hoge  "), "  hoge".to_string());
    assert_eq!(trim_end_weak("  hoge  \n"), "  hoge".to_string());
    assert_eq!(trim_end_weak("  hoge  \n  "), "  hoge  \n".to_string());
    assert_eq!(trim_end_weak("  hoge  \n\n"), "  hoge  \n".to_string());
    assert_eq!(trim_end_weak("  hoge  \n  \n"), "  hoge  \n".to_string());
    assert_eq!(trim_end_weak("  hoge  \n  \n  "), "  hoge  \n  \n".to_string());
}
```

chrisduerr: Honestly just `trem_end_matches(' ')` should work. We don't actually care about trimming anything other than empty cell content. If a user outputs text which contains whitespace that isn't actually empty terminal content, we should probably copy it anyway.

a5ob7r: Your comments probably make sense, but it is difficult to do it on the current implementation. Now no way to distinguish whether or not a cell is empty or just whitespace, right? Alacritty treats a space cell as a virtual empty cell on `line_to_string`. I think this makes sense on the current implementation.

BTW, it is awesome if we can distinguish an empty cell and a whitespace cell. It can improve the ideal behavior for the `FirstOccupied` action for vi-mode for me.

chrisduerr: > Now no way to distinguish whether or not a cell is empty or just whitespace, right?

Correct.

> BTW, it is awesome if we can distinguish an empty cell and a whitespace cell. It can improve the ideal behavior for the FirstOccupied action for vi-mode for me.

It isn't, especially because cursor motions allow writing in a different cell without actually ever inserting any whitespace. The only thing terminal emulators can realistically do is store what is an explicit and what is an implicit whitespace, but ultimately it's still all whitespace.

alacritty_terminal/src/term/mod.rs
```diff
@@ -355,16 +355,22 @@ impl<T> Term<T> {
...
 
-                    // If the last column is included, newline is appended automatically.
-                    if end.column != self.columns() - 1 {
+                    res += s;
+
+                    // Every line, except the last line, of block selection, must include a
+                    // trailing newline, but `line_to_string` may return a non-trailing newline
+                    // string when the selection doesn't contain the last column or it is wrapped.
+                    // So we need to append a newline if such a situation happens.
+                    if !s.ends_with('\n') {
```
chrisduerr, line 367: What's the point of the separate variable `s`?

alacritty_terminal/src/term/mod.rs
```diff
@@ -409,7 +415,16 @@ impl<T> Term<T> {
...
-        for column in (cols.start.0..line_length.0).map(Column::from) {
+        // Space trimmed line length.
+        let trimmed_line_length = (cols.start.0..line_length.0)
+            .map(Column::from)
+            .rev()
+            .find(|col| {
+                let cell = &grid_line[*col];
+                cell.c != ' ' || cell.zerowidth().map_or(false, |v| !v.is_empty())
+            })
+            .map_or(cols.start, |col| col + 1);
```
chrisduerr, line 426: Not a fan of having this logic in `line_to_string`.

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
a5ob7r: The reason is for correct string joining even if `s` is an empty string. Empty lines are discarded if not doing so.

a5ob7r:
Sorry for reverting to an old version. Actually, we need to trim whitespaces for block selection in `selection_to_string` instead of `line_to_string` to keep trailing whitespaces of wrapline for simple/lines selection. And I have added a unit test to check it.

chrisduerr (approved):
This patch has been going back-and-forth for far too long already. The current solution certainly isn't ideal or elegant, but it seems to be the simplest fix for 0.10.0 so let's go with this version.

PR #5740, opened 2022-01-02
Use builtin font for box drawing unicode characters

### LINKED ISSUE #5485: [Feature request] Special handling of Block Elements (U+2580 - U+259F)
### Intro

Block element characters are used for terminal fill and shading. 
https://unicode-table.com/en/blocks/block-elements/

The most common characters are:
- `█` (U+2588) Full Block
- `▄` (U+2584) Lower Half Block
- `▀` (U+2580) Upper Half Block
- `▗▘▙▚▛▜▝▞▟` (U+2596 - U+259F) Terminal graphic characters

Currently, they simply use the font's glygh for rendering, which cannot accurately fill a whole cell (or some borders) and instead leave a thin 0.5-1px border.
The following code *should* print a 2x2 full block, but currently it doesn't. There's a thin gap between adjacent cell (both between columns and rows)
`printf '██\n██\n'`

blocks

I tried all installed FreeType fonts including `Noto Sans Mono`, `Free Mono` and `Sarasa Mono` but none of them work. So I think it's not an issue of individual fonts, but some limitations of FreeType itself.

### Proposal 

Specially handle these block-like characters to make them *accurately fully* fill cell (or borders). There are also many terminals did so, including:
- GNOME vte: https://github.com/GNOME/vte/commit/154abade902850afb44115cccf8fcac51fc082f0
- Kitty: https://github.com/kovidgoyal/kitty/commit/e95c789d4bf6e52587b39ff10f419dd4442e92dd
- Konsole: Enabled special treatment by default. Can be disabled by a setting `Use line characters contained in font` in `Appearance` (then it behaves the same as current `alacritty`).

### Affected programs

Programs using these characters to print ASCII art, graphics, or indicators. Some users also use them in shell prompts.

`neofetch`:
nixos

### System

OS: Linux
Version: Current master cf35a2019162c6055e4a4c152976a424ba10869d
Linux: X11, KDE

kchibisov:
This commit adds hand rolled drawing of unicode box drawing[1] and
block elements[2] from ranges U+2500 up to U+259f. While using system
font for such characters will look better most of the time, the
characters tend to overlap or not align, so providing builtin font is
the lesser evil here.

[1] - https://www.unicode.org/charts/PDF/U2500.pdf
[2] - https://www.unicode.org/charts/PDF/U2580.pdf

Fixes #5485.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Display area keeps history position when viewport is cleared
 - Commands spawn from the current directory of the foreground shell in Unix-like systems
 - Remove trailing newline from strings taken from hints or simple/semantic selections
+- Builtin font is now used for box drawing characters from `U+2500` to `U+259f`
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,10 +36,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Strip trailing whitespaces when yanking from a block selection
 - Display area keeps history position when viewport is cleared
 - Commands spawn from the current directory of the foreground shell in Unix-like systems
-<<<<<<< HEAD
 - Remove trailing newline from strings taken from hints or simple/semantic selections
-=======
->>>>>>> 214e1e50 (Resolve style issues)
 - Builtin font is now used for box drawing characters from `U+2500` to `U+259f`
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Strip trailing whitespaces when yanking from a block selection
 - Display area keeps history position when viewport is cleared
 - Commands spawn from the current directory of the foreground shell in Unix-like systems
+<<<<<<< HEAD
 - Remove trailing newline from strings taken from hints or simple/semantic selections
+=======
+>>>>>>> 214e1e50 (Resolve style issues)
 - Builtin font is now used for box drawing characters from `U+2500` to `U+259f`
 
 ### Fixed
--- a/alacritty/src/renderer/box_drawing.rs
+++ b/alacritty/src/renderer/box_drawing.rs
@@ -1,0 +1,583 @@
+//! Hand rolled drawing of unicode [box drawing](http://www.unicode.org/charts/PDF/U2500.pdf)
+//! and [block elements](https://www.unicode.org/charts/PDF/U2580.pdf).
+
+use std::cmp::{max, min};
+
+use crossfont::{BitmapBuffer, Metrics, RasterizedGlyph};
+
+/// Stroke size of a line as a part of the cell width.
+const LINE_STROKE_SIZE: usize = 6;
+
+/// Stroke size of a heavy line as a part of the cell width.
+const HEAVY_LINE_STROKE_SIZE: usize = 3;
+
+// Colors which are used for filling shade variants.
+const COLOR_FILL_ALPHA_STEP_1: RGBPixel = RGBPixel { _r: 50, _g: 50, _b: 50 };
+const COLOR_FILL_ALPHA_STEP_2: RGBPixel = RGBPixel { _r: 125, _g: 125, _b: 125 };
+const COLOR_FILL_ALPHA_STEP_3: RGBPixel = RGBPixel { _r: 200, _g: 200, _b: 200 };
+
+/// Default color used for filling.
+const COLOR_FILL: RGBPixel = RGBPixel { _r: 255, _g: 255, _b: 255 };
+
+/// Returns `Some(RasterizedGlyph)` if character could be drawn with Alacritty's builtin set of
+/// glyphs otherwise `None`.
+pub fn builtin_glyph(character: char, metrics: &Metrics) -> Option<RasterizedGlyph> {
+    match character {
+        // Box drawing characters and block elements.
+        '\u{2500}'..='\u{259f}' => Some(box_drawing(character, metrics)),
+        _ => None,
+    }
+}
+
+fn box_drawing(character: char, metrics: &Metrics) -> RasterizedGlyph {
+    let height = metrics.line_height as usize;
+    let width = metrics.average_advance as usize;
+    let stroke_size = max(width / LINE_STROKE_SIZE, 1);
+    let heavy_stroke_size = max(width / HEAVY_LINE_STROKE_SIZE, 1);
+    let mut canvas = Canvas::new(width, height);
+
+    match character {
+        // Horizonatal dashes: '┄', '┅', '┈', '┉', '╌', '╍'.
+        '\u{2504}' | '\u{2505}' | '\u{2508}' | '\u{2509}' | '\u{254c}' | '\u{254d}' => {
+            let (num_gaps, stroke_size) = match character {
+                '\u{2504}' => (2, stroke_size),
+                '\u{2505}' => (2, heavy_stroke_size),
+                '\u{2508}' => (3, stroke_size),
+                '\u{2509}' => (3, heavy_stroke_size),
+                '\u{254c}' => (1, stroke_size),
+                '\u{254d}' => (1, heavy_stroke_size),
+                _ => unreachable!(),
+            };
+
+            let dash_gap_len = max(width / 8, 1);
+            let dash_len = max(width.saturating_sub(dash_gap_len * num_gaps) / (num_gaps + 1), 1);
+            let y = canvas.center_v();
+            for gap in 0..=num_gaps {
+                let x = min(gap * (dash_len + dash_gap_len), width);
+                canvas.draw_h_line(x as f32, y, dash_len as f32, stroke_size);
+            }
+        },
+        // Vertical dashes: '┆', '┇', '┊', '┋', '╎', '╏'.
+        '\u{2506}' | '\u{2507}' | '\u{250a}' | '\u{250b}' | '\u{254e}' | '\u{254f}' => {
+            let (num_gaps, stroke_size) = match character {
+                '\u{2506}' => (2, stroke_size),
+                '\u{2507}' => (2, heavy_stroke_size),
+                '\u{250a}' => (3, stroke_size),
+                '\u{250b}' => (3, heavy_stroke_size),
+                '\u{254e}' => (1, stroke_size),
+                '\u{254f}' => (1, heavy_stroke_size),
+                _ => unreachable!(),
+            };
+
+            let dash_gap_len = max(height / 8, 1);
+            let dash_len = max(height.saturating_sub(dash_gap_len * num_gaps) / (num_gaps + 1), 1);
+            let x = canvas.center_h();
+            for gap in 0..=num_gaps {
+                let y = min(gap * (dash_len + dash_gap_len), height);
+                canvas.draw_v_line(x, y as f32, dash_len as f32, stroke_size);
+            }
+        },
+        // Horizontal lines: '─', '━', '╴', '╶', '╸', '╺'.
+        // Vertical lines: '│', '┃', '╵', '╷', '╹', '╻'.
+        // Light and heavy line box components:
+        // '┌','┍','┎','┏','┐','┑','┒','┓','└','┕','┖','┗','┘','┙','┚','┛',├','┝','┞','┟','┠','┡',
+        // '┢','┣','┤','┥','┦','┧','┨','┩','┪','┫','┬','┭','┮','┯','┰','┱','┲','┳','┴','┵','┶','┷',
+        // '┸','┹','┺','┻','┼','┽','┾','┿','╀','╁','╂','╃','╄','╅','╆','╇','╈','╉','╊','╋'.
+        // Mixed light and heavy lines: '╼', '╽', '╾', '╿'.
+        '\u{2500}'..='\u{2503}' | '\u{250c}'..='\u{254b}' | '\u{2574}'..='\u{257f}' => {
+            // Left horizontal line.
+            let stroke_size_h1 = match character {
+                '\u{2500}' | '\u{2510}' | '\u{2512}' | '\u{2518}' | '\u{251a}' | '\u{2524}'
+                | '\u{2526}' | '\u{2527}' | '\u{2528}' | '\u{252c}' | '\u{252e}' | '\u{2530}'
+                | '\u{2532}' | '\u{2534}' | '\u{2536}' | '\u{2538}' | '\u{253a}' | '\u{253c}'
+                | '\u{253e}' | '\u{2540}' | '\u{2541}' | '\u{2542}' | '\u{2544}' | '\u{2546}'
+                | '\u{254a}' | '\u{2574}' | '\u{257c}' => stroke_size,
+                '\u{2501}' | '\u{2511}' | '\u{2513}' | '\u{2519}' | '\u{251b}' | '\u{2525}'
+                | '\u{2529}' | '\u{252a}' | '\u{252b}' | '\u{252d}' | '\u{252f}' | '\u{2531}'
+                | '\u{2533}' | '\u{2535}' | '\u{2537}' | '\u{2539}' | '\u{253b}' | '\u{253d}'
+                | '\u{253f}' | '\u{2543}' | '\u{2545}' | '\u{2547}' | '\u{2548}' | '\u{2549}'
+                | '\u{254b}' | '\u{2578}' | '\u{257e}' => heavy_stroke_size,
+                _ => 0,
+            };
+            // Right horizontal line.
+            let stroke_size_h2 = match character {
+                '\u{2500}' | '\u{250c}' | '\u{250e}' | '\u{2514}' | '\u{2516}' | '\u{251c}'
+                | '\u{251e}' | '\u{251f}' | '\u{2520}' | '\u{252c}' | '\u{252d}' | '\u{2530}'
+                | '\u{2531}' | '\u{2534}' | '\u{2535}' | '\u{2538}' | '\u{2539}' | '\u{253c}'
+                | '\u{253d}' | '\u{2540}' | '\u{2541}' | '\u{2542}' | '\u{2543}' | '\u{2545}'
+                | '\u{2549}' | '\u{2576}' | '\u{257e}' => stroke_size,
+                '\u{2501}' | '\u{250d}' | '\u{250f}' | '\u{2515}' | '\u{2517}' | '\u{251d}'
+                | '\u{2521}' | '\u{2522}' | '\u{2523}' | '\u{252e}' | '\u{252f}' | '\u{2532}'
+                | '\u{2533}' | '\u{2536}' | '\u{2537}' | '\u{253a}' | '\u{253b}' | '\u{253e}'
+                | '\u{253f}' | '\u{2544}' | '\u{2546}' | '\u{2547}' | '\u{2548}' | '\u{254a}'
+                | '\u{254b}' | '\u{257a}' | '\u{257c}' => heavy_stroke_size,
+                _ => 0,
+            };
+            // Top vertical line.
+            let stroke_size_v1 = match character {
+                '\u{2502}' | '\u{2514}' | '\u{2515}' | '\u{2518}' | '\u{2519}' | '\u{251c}'
+                | '\u{251d}' | '\u{251f}' | '\u{2522}' | '\u{2524}' | '\u{2525}' | '\u{2527}'
+                | '\u{252a}' | '\u{2534}' | '\u{2535}' | '\u{2536}' | '\u{2537}' | '\u{253c}'
+                | '\u{253d}' | '\u{253e}' | '\u{253f}' | '\u{2541}' | '\u{2545}' | '\u{2546}'
+                | '\u{2548}' | '\u{2575}' | '\u{257d}' => stroke_size,
+                '\u{2503}' | '\u{2516}' | '\u{2517}' | '\u{251a}' | '\u{251b}' | '\u{251e}'
+                | '\u{2520}' | '\u{2521}' | '\u{2523}' | '\u{2526}' | '\u{2528}' | '\u{2529}'
+                | '\u{252b}' | '\u{2538}' | '\u{2539}' | '\u{253a}' | '\u{253b}' | '\u{2540}'
+                | '\u{2542}' | '\u{2543}' | '\u{2544}' | '\u{2547}' | '\u{2549}' | '\u{254a}'
+                | '\u{254b}' | '\u{2579}' | '\u{257f}' => heavy_stroke_size,
+                _ => 0,
+            };
+            // Bottom vertical line.
+            let stroke_size_v2 = match character {
+                '\u{2502}' | '\u{250c}' | '\u{250d}' | '\u{2510}' | '\u{2511}' | '\u{251c}'
+                | '\u{251d}' | '\u{251e}' | '\u{2521}' | '\u{2524}' | '\u{2525}' | '\u{2526}'
+                | '\u{2529}' | '\u{252c}' | '\u{252d}' | '\u{252e}' | '\u{252f}' | '\u{253c}'
+                | '\u{253d}' | '\u{253e}' | '\u{253f}' | '\u{2540}' | '\u{2543}' | '\u{2544}'
+                | '\u{2547}' | '\u{2577}' | '\u{257f}' => stroke_size,
+                '\u{2503}' | '\u{250e}' | '\u{250f}' | '\u{2512}' | '\u{2513}' | '\u{251f}'
+                | '\u{2520}' | '\u{2522}' | '\u{2523}' | '\u{2527}' | '\u{2528}' | '\u{252a}'
+                | '\u{252b}' | '\u{2530}' | '\u{2531}' | '\u{2532}' | '\u{2533}' | '\u{2541}'
+                | '\u{2542}' | '\u{2545}' | '\u{2546}' | '\u{2548}' | '\u{2549}' | '\u{254a}'
+                | '\u{254b}' | '\u{257b}' | '\u{257d}' => heavy_stroke_size,
+                _ => 0,
+            };
+
+            let x_v = canvas.center_h();
+            let y_h = canvas.center_v();
+
+            let v_line_bounds_top = canvas.v_line_bounds(x_v, stroke_size_v1);
+            let v_line_bounds_bot = canvas.v_line_bounds(x_v, stroke_size_v2);
+            let h_line_bounds_left = canvas.h_line_bounds(y_h, stroke_size_h1);
+            let h_line_bounds_right = canvas.h_line_bounds(y_h, stroke_size_h2);
+
+            let size_h1 = max(v_line_bounds_top.1 as i32, v_line_bounds_bot.1 as i32) as f32;
+            let x_h = min(v_line_bounds_top.0 as i32, v_line_bounds_bot.0 as i32) as f32;
+            let size_h2 = width as f32 - x_h;
+
+            let size_v1 = max(h_line_bounds_left.1 as i32, h_line_bounds_right.1 as i32) as f32;
+            let y_v = min(h_line_bounds_left.0 as i32, h_line_bounds_right.0 as i32) as f32;
+            let size_v2 = height as f32 - y_v;
+
+            // Left horizontal line.
+            canvas.draw_h_line(0., y_h, size_h1, stroke_size_h1);
+            // Right horizontal line.
+            canvas.draw_h_line(x_h, y_h, size_h2, stroke_size_h2);
+            // Top vertical line.
+            canvas.draw_v_line(x_v, 0., size_v1, stroke_size_v1);
+            // Bottom vertical line.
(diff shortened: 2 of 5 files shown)


alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+        // Light and heavy line box components:
+        // '┌','┍','┎','┏','┐','┑','┒','┓','└','┕','┖','┗','┘','┙','┚','┛',├','┝','┞','┟','┠','┡',
+        // '┢','┣','┤','┥','┦','┧','┨','┩','┪','┫','┬','┭','┮','┯','┰','┱','┲','┳','┴','┵','┶','┷',
+        // '┸','┹','┺','┻','┼','┽','┾','┿','╀','╁','╂','╃','╄','╅','╆','╇','╈','╉','╊','╋'.
+        // Mixed light and heavy lines: '╼', '╽', '╾', '╿'.
+        '\u{2500}'..='\u{2503}' | '\u{250c}'..='\u{254b}' | '\u{2574}'..='\u{257f}' => {
+            // Left horizontal line.
+            let stroke_size_h1 = match character {
+                '\u{2500}' | '\u{2510}' | '\u{2512}' | '\u{2518}' | '\u{251a}' | '\u{2524}'
+                | '\u{2526}' | '\u{2527}' | '\u{2528}' | '\u{252c}' | '\u{252e}' | '\u{2530}'
```
chrisduerr, line 91: Bunch of glyphs here are more than 2 in sequential order, where `..=` might be beneficial. Did you intentionally write out every single one for consistency to make editing in the future easier without having to break your brain about character ranges?

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+            // Right horizontal line.
+            canvas.draw_h_line(x_h, y_h, size_h2, stroke_size_h2);
+            // Top vertical line.
+            canvas.draw_v_line(x_v, 0., size_v1, stroke_size_v1);
+            // Bottom vertical line.
+            canvas.draw_v_line(x_v, y_v, size_v2, stroke_size_v2);
+        },
+        // Light and double line box components:
+        // '═','║','╒','╓','╔','╕','╖','╗','╘','╙','╚','╛','╜','╝','╞','╟','╠','╡','╢','╣','╤','╥',
+        // '╦','╧','╨','╩','╪','╫','╬'.
```
chrisduerr, line 172: I'd assume the "light" part here was a copy/paste error? These are only double-line box components.

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+
+            // Top vertical line.
+            canvas.draw_v_line(v_lines.0, 0., left_top_size, stroke_size);
+            canvas.draw_v_line(v_lines.1, 0., right_top_size, stroke_size);
+
+            // Bottom vertical line.
+            canvas.draw_v_line(v_lines.0, left_bot_y, bottom_size, stroke_size);
+            canvas.draw_v_line(v_lines.1, right_bot_y, bottom_size, stroke_size);
+        },
+        // Arcs: '╭', '╮', '╯', '╰'.
```
chrisduerr, line 276: I wasn't even sure if I should mention this, but there might be something like a one-off caused by flipping around the arc in edge-cases?

At least when I combine them with the light horizontal dash, the top and bottom look slightly different:
tmp

Notice how at the bottom the thing is completely flat even into the arc, while at the top it seems to step down more aggressively and sooner? I'm pretty sure these two are supposed to be used together (arcs and light horizontal/vertical), so it would probably make sense to ensure they meet up properly.

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+            }
+        },
+        // Parts of full block: '▀', '▁', '▂', '▃', '▄', '▅', '▆', '▇', '▔', '▉', '▊', '▋', '▌',
+        // '▍', '▎', '▏', '▐', '▕'.
+        '\u{2580}'..='\u{2587}' | '\u{2589}'..='\u{2590}' | '\u{2594}' | '\u{2595}' => {
+            let width = width as f32;
+            let height = height as f32;
+            let rect_width = match character {
+                '\u{2589}' => width * 7. / 8.,
+                '\u{258a}' => width * 3. / 4.,
```
chrisduerr, line 325: Does clippy complain about 4./8. here? If not could we please use `/ 8.` for all of them? My head hurts when I have to do math so just counting from 1 to 7 seems much more pleasant to me. :)

Same goes for other places this is done.

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+    RasterizedGlyph { character, top, left: 0, height: height as i32, width: width as i32, buffer }
+}
+
+#[repr(packed)]
+#[derive(Clone, Copy, Debug, Default)]
+struct RGBPixel {
+    _r: u8,
+    _g: u8,
+    _b: u8,
+}
```
chrisduerr, line 426: Why the `repr(packed)` and `_` prefix here? It should also be `RgbPixel` and I think `Pixel` could potentially be omitted completely.

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+        let end_x = min((x + width) as usize, self.width);
+        let start_y = y as usize;
+        let end_y = min((y + height) as usize, self.height);
+        for y in start_y..end_y {
+            let y = y * self.width;
+            self.buffer[start_x + y..end_x + y].fill(color);
+        }
+    }
+
+    /// Naive arbitrary line drawing from (`from_x`, `from_y`) to (`to_x`, `to_y`).
```
chrisduerr, line 502: It's not really arbitrary, is it? Kinda makes it sound like you're going in a zigzag or whatever.

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+        }
+    }
+
+    /// Fills the `Canvas` with the given `Color`.
+    fn fill(&mut self, color: RGBPixel) {
+        self.buffer.fill(color);
+    }
+
+    /// Consumes `Canvas` and returns its underlying storage as raw byte vector.
+    fn into_raw_buffer(self) -> Vec<u8> {
```
chrisduerr, line 548: I think in Rust it's usually just `into_raw`, the buffer can be omitted?

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+        // SAFETY This is safe since we use `repr(packed)` on `Pixel` struct for underlying storage
+        // of the `Canvas` buffer which consists of three u8 values.
```
chrisduerr, line 550: Ah, so this is why you did `repr(packed)` I suppose? But with 3 u8s shouldn't this also work without it? Since they're all single-byte aligned anyway?

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+        unsafe {
+            let capacity = self.buffer.capacity();
+            let len = self.buffer.len() * 3;
+            let buf = self.buffer.as_ptr() as *mut u8;
+            std::mem::forget(self.buffer);
```
chrisduerr, line 555:
```suggestion
            mem::forget(self.buffer);
```

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+        self.buffer.fill(color);
+    }
+
+    /// Consumes `Canvas` and returns its underlying storage as raw byte vector.
+    fn into_raw_buffer(self) -> Vec<u8> {
+        // SAFETY This is safe since we use `repr(packed)` on `Pixel` struct for underlying storage
+        // of the `Canvas` buffer which consists of three u8 values.
+        unsafe {
+            let capacity = self.buffer.capacity();
+            let len = self.buffer.len() * 3;
```
chrisduerr, line 553:
```suggestion
            let len = self.buffer.len() * mem::size_of<RGBPixel>();
```

This should be more resistant to failure I think?

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+            let buf = self.buffer.as_ptr() as *mut u8;
+            std::mem::forget(self.buffer);
+            Vec::from_raw_parts(buf, len, capacity)
```
chrisduerr, line 556: I'm not sure the capacity specified here is correct? Because ultimately the capacity of the old vec is going to be a third of the capacity of the new one? Length **must** be equal or less than capacity and since you're multiplying the length of the original vector by 3, if the capacity is the same as the length this will not hold true.

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+            descent: 4.,
+            underline_position: 2.,
+            underline_thickness: 2.,
+            strikeout_position: 2.,
+            strikeout_thickness: 2.,
+        };
+
+        // Test coverage of box drawing characters.
+        for character in '\u{2500}'..='\u{259f}' {
+            assert!(builtin_glyph(character, &metrics).is_some());
```
chrisduerr, line 581: Might be worthwhile to check the negative case too maybe? Not sure what range you'd pick for that, but maybe something surrounding the box drawing range just to make sure?

chrisduerr (changes requested):
On a quick visual review I'd say the only characters that might look slightly off are the following:

╭	9581	╭	256D	BOX DRAWINGS LIGHT ARC DOWN AND RIGHT
╮	9582	╮	256E	BOX DRAWINGS LIGHT ARC DOWN AND LEFT
╯	9583	╯	256F	BOX DRAWINGS LIGHT ARC UP AND LEFT
╰	9584	╰	2570	BOX DRAWINGS LIGHT ARC UP AND RIGHT
╱	9585	╱	2571	BOX DRAWINGS LIGHT DIAGONAL UPPER RIGHT TO LOWER LEFT
╲	9586	╲	2572	BOX DRAWINGS LIGHT DIAGONAL UPPER LEFT TO LOWER RIGHT
╳	9587	╳	2573	BOX DRAWINGS LIGHT DIAGONAL CROSS

It's not super surprising that curves and diagonals aren't ideal unless you want to get into anti-aliasing, but just thought at the very least I'd point that out. I'm not sure if I'd ever want to see any of these in my terminal with the current font though, or if at that point it wouldn't be better to use my own for these.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Minimum Rust version has been bumped to 1.56.0
 
+### Changed
+
+- Builtin font is now used for box drawing characters from `U+2500` to `U+259f`
```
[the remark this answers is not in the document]
kchibisov: Maybe on a separate PR? I don't really care about powerline glyphs for now and it's also some kind of art. Maybe in a separate PR.

alacritty/src/renderer/mod.rs
```diff
@@ -827,9 +833,9 @@ impl<'a> RenderApi<'a> {
         self.batch.clear();
     }
 
-    /// Render a string in a variable location. Used for printing the render timer, warnings and
+    /// Draw a string in a variable location. Used for printing the render timer, warnings and
     /// errors.
-    pub fn render_string(
+    pub fn draw_string(
```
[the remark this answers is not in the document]
kchibisov: I don't really know, but we have something called `draw`, I think `draw_rect` and I've renamed all of them to make things consistent.

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+#[repr(packed)]
+#[derive(Clone, Copy, Debug, Default)]
+struct RGBPixel {
+    _r: u8,
+    _g: u8,
+    _b: u8,
+}
+
+/// Canvas which is used for simple line drawing operations.
+struct Canvas {
```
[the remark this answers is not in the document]
kchibisov: I mean you can pick arbitrary size, it's just a buffer with some primitives defined to draw on it. I'm not sure how you'd call it otherwise, for me it's natural. And you're not drawing glyphs, you're drawing lines, it just happend that those are glyphs in the end.

alacritty/src/renderer/box_drawing.rs
kchibisov: Yeah, and it was also a sanity check, since you should have the same amount of elements in all of match arms and matches, so you won't miss anything. And if you look on some of them they have like 2 at max contiguous sequential elements, which is pretty much the same amount of chars as putting `|`.

alacritty/src/renderer/box_drawing.rs
kchibisov: Nah, look on the unicode URL, and if you look closely some of them have single light lines in them.

alacritty/src/renderer/box_drawing.rs
kchibisov: packed is just to ensure that everything is packed regardless of what compiler could think here. I don't trust compilers when it comes to layouts. Like it should be byte aligned, but...

alacritty/src/renderer/box_drawing.rs
kchibisov: Oh, I somehow assumed that capacity returns bytes, but it's also in elements....

alacritty/src/renderer/box_drawing.rs
kchibisov: The thing is that if something is missing in the function it'll just straight crash on some `unreachable!()`, so I'm mostly testing those things that nothing is actually missed here. I'm not sure that for negative it makes sense.

alacritty/src/renderer/box_drawing.rs
kchibisov: The glyphs are named the way you should divide them, but ok, can update to / 8.

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
+//! Hand rolled drawing of unicode [box drawing](http://www.unicode.org/charts/PDF/U2500.pdf)
+//! and [block elements](https://www.unicode.org/charts/PDF/U2580.pdf).
+
+use std::cmp::{max, min};
+
+use crossfont::{BitmapBuffer, Metrics, RasterizedGlyph};
+
+/// Stroke size of a line as a part of the cell width.
+const LINE_STROKE_SIZE: usize = 6;
```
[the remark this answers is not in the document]
kchibisov: Sounds reasonable.

alacritty/src/renderer/box_drawing.rs
```diff
@@ -0,0 +1,584 @@
...
+
+/// Default color used for filling.
+const COLOR_FILL: RGBPixel = RGBPixel { _r: 255, _g: 255, _b: 255 };
+
+/// Returns `Some(RasterizedGlyph)` if character could be drawn with Alacritty's builtin set of
+/// glyphs otherwise `None`.
+pub fn builtin_glyph(character: char, metrics: &Metrics) -> Option<RasterizedGlyph> {
+    match character {
+        // Box drawing characters and block elements.
+        '\u{2500}'..='\u{259f}' => Some(box_drawing(character, metrics)),
```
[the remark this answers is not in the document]
kchibisov: Renamed it, I had like font.rs originally, but prefixing with builtin also fine.

kchibisov:
Besides Arc part I've resolved everything else I suppose. I'm not sure what to do with them other than changing the approach of drawing them.

alacritty/src/renderer/mod.rs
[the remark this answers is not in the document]
chrisduerr: Sounds like a good excuse. :)

alacritty/src/renderer/box_drawing.rs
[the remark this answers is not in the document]
chrisduerr: I don't have any better suggestion, so I guess it's fine.

alacritty/src/renderer/box_drawing.rs
chrisduerr: But even the ones with the light lines don't have light line in the name? It's always "single", never "light"?

alacritty/src/renderer/box_drawing.rs
chrisduerr: Ordering might also be an issue, because Rust can freely reorder fileds. So repr is good here.

alacritty/src/renderer/box_drawing.rs
chrisduerr: The `builtin_glyph` function should never panic, right?

chrisduerr (changes requested):
> Besides Arc part I've resolved everything else I suppose. I'm not sure what to do with them other than changing the approach of drawing them.

I mean the mirroring thing could potentially be fixed I'd hope so that at least they look consistent.

As far as them looking bad is concerned it would be possible to just not provide them with our font if we can't draw them better than the font itself. Personally I'd rather have the potential overlapping issue than a glyph that looks completely terrible.

alacritty/src/renderer/box_drawing.rs
kchibisov: box_drawing could panic if you remove some element from pattern matching.

chrisduerr: I mean in that case it would be good to have a test that fails, right?

kchibisov:
I think I've resolved the rest of the issues. Should be fine now.

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,700 @@
...
+            let width = width as f32;
+            let height = height as f32;
+            let rect_width = match character {
+                '\u{2589}' => width * 7. / 8.,
+                '\u{258a}' => width * 6. / 8.,
+                '\u{258b}' => width * 5. / 8.,
+                '\u{258c}' | '\u{2590}' => width * 4. / 8.,
+                '\u{258d}' => width * 3. / 8.,
+                '\u{258e}' => width * 2. / 8.,
+                '\u{258f}' | '\u{2595}' => width / 8.,
```
chrisduerr, line 338:
```suggestion
                '\u{258f}' | '\u{2595}' => width * 1. / 8.,
```

Would maybe even like to see the `|` grouping removed here, just to have **everything** orderly and under each other making it clear what's going on.

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,700 @@
...
+                _ => width,
+            };
+            let (rect_height, y) = match character {
+                '\u{2580}' => (height * 4. / 8., height),
```
chrisduerr, line 342:
```suggestion
                '\u{2580}' => (height * 4. / 8., height * 8. / 1.),
```

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,700 @@
...
+                '\u{2581}' => (height / 8., height / 8.),
```
chrisduerr, line 343:
```suggestion
                '\u{2581}' => (height * 1. / 8., height * 1. / 8.),
```

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,700 @@
...
+                '\u{2582}' => (height * 2. / 8., height * 2. / 8.),
+                '\u{2583}' => (height * 3. / 8., height * 3. / 8.),
+                '\u{2584}' => (height * 4. / 8., height * 4. / 8.),
+                '\u{2585}' => (height * 5. / 8., height * 5. / 8.),
+                '\u{2586}' => (height * 6. / 8., height * 6. / 8.),
+                '\u{2587}' => (height * 7. / 8., height * 7. / 8.),
+                '\u{2594}' => (height / 8., height),
```
chrisduerr, line 350:
```suggestion
                '\u{2594}' => (height * 1. / 8., height * 8. / 8.),
```

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,700 @@
...
+        let delta_x = to_x - from_x;
+        let delta_y = to_y - from_y;
+        let gradient = if delta_x.abs() <= f32::EPSILON { 1. } else { delta_y / delta_x };
+
+        let x_end = f32::round(from_x);
+        let y_end = from_y + gradient * (x_end - from_x);
+        let x_gap = (from_x + 0.5).fract();
+
+        let xpx11 = f32::floor(x_end);
+        let ypx11 = f32::floor(y_end);
```
chrisduerr, line 540: I'd assume these variable names come from the wikipedia entry for the algorithm? In which case it's actually `xpxl1`/`ypxl1`.

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,700 @@
...
+            self.put_pixel(xpx11 + 1., ypx11, color_2);
+        }
+
+        let mut intery = y_end + gradient;
+
+        let x_end = f32::round(to_x);
+        let y_end = to_y + gradient * (x_end - to_x);
+        let x_gap = (to_x + 0.5).fract();
+        let xpx12 = f32::floor(x_end);
+        let ypx12 = f32::floor(y_end);
```
chrisduerr, line 558: I'd assume these variable names come from the wikipedia entry for the algorithm? In which case it's actually `xpxl2`/`ypxl2`.

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,700 @@
...
+                let color_2 = Pixel::gray((intery.fract() * COLOR_FILL._r as f32) as u8);
+                self.put_pixel(x as f32, intery.trunc(), color_1);
+                self.put_pixel(x as f32, intery.trunc() + 1., color_2);
+                intery += gradient;
+            }
+        }
+    }
+
+    /// Draws an arc from `(0, self.center_v())` to `(self.center_h(), 0)` with the specified
+    /// stroke.
```
chrisduerr, line 590: Documenting directionality of the arc might be helpful to understanding the algorithm.

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,700 @@
...
+        let max_transparancy = 0.5;
+
+        for (radious_y, radious_x) in (h_line_bounds.0..h_line_bounds.1)
+            .into_iter()
+            .zip((v_line_bounds.0..v_line_bounds.1).into_iter())
+        {
+            let radious_x = radious_x as f32;
+            let radious_y = radious_y as f32;
+            let radious_x2 = radious_x * radious_x;
+            let radious_y2 = radious_y * radious_y;
```
chrisduerr, line 614:
```suggestion
            let radius_x = radius_x as f32;
            let radius_y = radius_y as f32;
            let radius_x2 = radius_x * radius_x;
            let radius_y2 = radius_y * radius_y;
```

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,700 @@
...
+    fn fill(&mut self, color: Pixel) {
+        self.buffer.fill(color);
+    }
+
+    /// Consumes `Canvas` and returns its underlying storage as raw byte vector.
+    fn into_raw(self) -> Vec<u8> {
+        // SAFETY This is safe since we use `repr(packed)` on `Pixel` struct for underlying storage
+        // of the `Canvas` buffer which consists of three u8 values.
+        unsafe {
+            let capacity = self.buffer.capacity();
```
chrisduerr, line 664:
```suggestion
            let capacity = self.buffer.capacity() * mem::size_of::<Pixel>();
```

chrisduerr (changes requested):

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,700 @@
...
+            canvas.draw_v_line(v_lines.0, 0., left_top_size, stroke_size);
+            canvas.draw_v_line(v_lines.1, 0., right_top_size, stroke_size);
+
+            // Bottom vertical line.
+            canvas.draw_v_line(v_lines.0, left_bot_y, bottom_size, stroke_size);
+            canvas.draw_v_line(v_lines.1, right_bot_y, bottom_size, stroke_size);
+        },
+        // Arcs: '╭', '╮', '╯', '╰'.
+        '\u{256d}' | '\u{256e}' | '\u{256f}' | '\u{2570}' => {
+            canvas.draw_arc(stroke_size);
```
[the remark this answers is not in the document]
kchibisov: You mean in docs or right in function name? Could name like `draw_ellipse_arc_second_quadrant`, will give you precise naming of what it does.

alacritty/src/renderer/builtin_font.rs
kchibisov: nah, it's `* 8. / 8.`

alacritty/src/renderer/builtin_font.rs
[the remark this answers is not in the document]
chrisduerr: I meant right here as a comment. Renaming the function would also be a viable alternative.

alacritty/src/renderer/builtin_font.rs
chrisduerr: Of course it is.

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,751 @@
...
+            canvas.draw_v_line(v_lines.0, 0., left_top_size, stroke_size);
+            canvas.draw_v_line(v_lines.1, 0., right_top_size, stroke_size);
+
+            // Bottom vertical line.
+            canvas.draw_v_line(v_lines.0, left_bot_y, bottom_size, stroke_size);
+            canvas.draw_v_line(v_lines.1, right_bot_y, bottom_size, stroke_size);
+        },
+        // Arcs: '╭', '╮', '╯', '╰'.
+        '\u{256d}' | '\u{256e}' | '\u{256f}' | '\u{2570}' => {
+            canvas.draw_ellipse_arc(stroke_size);
```
chrisduerr, line 325:
```suggestion
            canvas.draw_ellipse_arc(stroke_size);
            
```

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,751 @@
...
+                let color_2 = Pixel::gray((intery.fract() * COLOR_FILL._r as f32) as u8);
+                self.put_pixel(x as f32, intery.trunc(), color_1);
+                self.put_pixel(x as f32, intery.trunc() + 1., color_2);
+                intery += gradient;
+            }
+        }
+    }
+
+    /// Draws a part of ellipse centered in `(0., 0.)` with `self.x_center()` and `self.y_center`
+    /// radius with a given `stroke` in second quadrant of the coordinate system.
```
chrisduerr, line 641:
```suggestion
    /// Draws a part of an ellipse centered in `(0., 0.)` with `self.x_center()` and `self.y_center()`
    /// radius with a given `stroke` in top-right quadrant of the coordinate system.
```

Second is a poor way to describe a quadrant because the numbering is somewhat subjective.

Also naming two values for a radius is kinda confusing and afaik not really the proper terminology for an ellipse. One could use the "proper" terms which are vertex/co-vertex according to Wikipedia but I don't think that's particularly helpful. Just saying width/height might be the easiest to understand?

chrisduerr (changes requested):
Looks like we're almost done, some minor stuff but other than that things are looking good.

alacritty/src/renderer/builtin_font.rs
kchibisov: The thing is that it's not a **top-right** in our case, since our y is towards bottom. Actually quadrants are defined and the first one is where both signs are `+`, and using `vertex` and `co-vertex`, so using proper terminology should be fine.

alacritty/src/renderer/builtin_font.rs
```diff
@@ -0,0 +1,753 @@
...
+                self.put_pixel(x as f32, intery.trunc(), color_1);
+                self.put_pixel(x as f32, intery.trunc() + 1., color_2);
+                intery += gradient;
+            }
+        }
+    }
+
+    /// Draws a part of an ellipse centered in `(0., 0.)` with `self.x_center()` and `self.y_center`
+    /// vertex and co-vertex respectively using a given `stroke` in the first quadrant of the
+    /// `Canvas` coordinate system.
```
chrisduerr, line 643:
```suggestion
    /// Draws a part of an ellipse centered in `(0., 0.)` with `self.x_center()` vertex and `self.y_center` co-vertex
    /// using a given `stroke_size` in the bottom-right quadrant of the  `Canvas` coordinate system.
```

Why do you still refer to it as "first" quadrant instead of using directionality? I'd personally prefer it much more like suggested, assuming it's actually the bottom-right. Otherwise use top-right etc of course.

chrisduerr (changes requested):

alacritty/src/renderer/builtin_font.rs
chrisduerr: > so using proper terminology should be fine.

Strongly disagree. Having some "proper" terminology only makes things easier to look up, it doesn't make them intuitive. Using the direction immediately makes things intuitive, there's no need to look anything up when it's already obvious due to its simple phrasing.

Yes quadrants might be well-defined, but that's already making so many assumptions about the space we're working in (mathematics), the education of the reader (has it been taught to you properly?), the accuracy of the author (did you actually remember it correctly?). There's just so many unnecessary points of conflict.

chrisduerr (approved):

PR #5752, opened 2022-01-07
Alacritty version 0.10.0-rc2

chrisduerr:
This should be the first actually viable RC for Alacritty 0.10.0, I'm pretty sure I've managed to cherry-pick all the remaining bugfixes. Let me know if you think any other commit should be included @kchibisov (or if I picked too much).


kchibisov (changes requested):
That seems fine, however iirc you want to include  #5757 as well?

kchibisov (approved):

PR #5764, opened 2022-01-10
Don't blink when cursor is hidden

kchibisov:
This prevents unnecessary redraws when the terminal is idle, cursor is
hidden and blinking is enabled.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -919,13 +919,16 @@ impl<'a, N: Notify + 'a, T: EventListener> ActionContext<'a, N, T> {
     fn update_cursor_blinking(&mut self) {
         // Get config cursor style.
         let mut cursor_style = self.config.terminal_config.cursor.style;
-        if self.terminal.mode().contains(TermMode::VI) {
+        let vi_mode = self.terminal.mode().contains(TermMode::VI);
+        if vi_mode {
             cursor_style = self.config.terminal_config.cursor.vi_mode_style.unwrap_or(cursor_style);
-        };
+        }
 
         // Check terminal cursor style.
         let terminal_blinking = self.terminal.cursor_style().blinking;
         let blinking = cursor_style.blinking_override().unwrap_or(terminal_blinking);
+        let blinking =
+            blinking && (vi_mode || self.terminal().mode().contains(TermMode::SHOW_CURSOR));
 
         // Update cursor blinking state.
         let timer_id = TimerId::new(Topic::BlinkCursor, self.display.window.id());
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -926,9 +926,8 @@ impl<'a, N: Notify + 'a, T: EventListener> ActionContext<'a, N, T> {
 
         // Check terminal cursor style.
         let terminal_blinking = self.terminal.cursor_style().blinking;
-        let blinking = cursor_style.blinking_override().unwrap_or(terminal_blinking);
-        let blinking =
-            blinking && (vi_mode || self.terminal().mode().contains(TermMode::SHOW_CURSOR));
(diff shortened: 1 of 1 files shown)


alacritty/src/event.rs
```diff
@@ -919,13 +919,16 @@ impl<'a, N: Notify + 'a, T: EventListener> ActionContext<'a, N, T> {
...
+        if vi_mode {
             cursor_style = self.config.terminal_config.cursor.vi_mode_style.unwrap_or(cursor_style);
-        };
+        }
 
         // Check terminal cursor style.
         let terminal_blinking = self.terminal.cursor_style().blinking;
         let blinking = cursor_style.blinking_override().unwrap_or(terminal_blinking);
+        let blinking =
+            blinking && (vi_mode || self.terminal().mode().contains(TermMode::SHOW_CURSOR));
```
chrisduerr, line 931:
```suggestion
        let blinking &= vi_mode || self.terminal().mode().contains(TermMode::SHOW_CURSOR);
```

chrisduerr: Should we really define this as two separate variables, one shadowing the other though? I feel like making the initial variable mutable is far superior. Even better might be just to add the `&& …` at the end of the definition.

alacritty/src/event.rs
```diff
@@ -919,13 +919,16 @@ impl<'a, N: Notify + 'a, T: EventListener> ActionContext<'a, N, T> {
     fn update_cursor_blinking(&mut self) {
         // Get config cursor style.
         let mut cursor_style = self.config.terminal_config.cursor.style;
-        if self.terminal.mode().contains(TermMode::VI) {
+        let vi_mode = self.terminal.mode().contains(TermMode::VI);
+        if vi_mode {
             cursor_style = self.config.terminal_config.cursor.vi_mode_style.unwrap_or(cursor_style);
-        };
+        }
```
chrisduerr, line 925: Do we need to check for `vi_mode || SHOW_CURSOR`, or can we just check for `cursor_style != HIDDEN`?

chrisduerr (changes requested):

alacritty/src/event.rs
kchibisov: This is a config cursor style, it can't be hidden. However the cursor style on a terminal could be hidden. And also `Hidden` on a terminal becomes when creating a `RendererCursor`, which is not available here. So we should check like that.

alacritty/src/event.rs
```diff
@@ -919,13 +919,15 @@ impl<'a, N: Notify + 'a, T: EventListener> ActionContext<'a, N, T> {
...
 
         // Check terminal cursor style.
         let terminal_blinking = self.terminal.cursor_style().blinking;
-        let blinking = cursor_style.blinking_override().unwrap_or(terminal_blinking);
+        let mut blinking = cursor_style.blinking_override().unwrap_or(terminal_blinking);
+        blinking &= vi_mode || self.terminal().mode().contains(TermMode::SHOW_CURSOR);
```
chrisduerr, line 930: You don't like the following?

```suggestion
        let blinking = cursor_style.blinking_override().unwrap_or(terminal_blinking)
            && (vi_mode || self.terminal().mode().contains(TermMode::SHOW_CURSOR));
```

I'm somewhat open to both here, but I'm surprised you didn't go for the solution that is shorter and doesn't require defining the variable as mutable, or modification of it.

Merge whatever you prefer, but I'd favor the suggestion.

chrisduerr (approved):

alacritty/src/event.rs
kchibisov: I had something like that but it looked  way too long for me, and it has more brackets...

chrisduerr: I mean it's objectively shorter and it's not like a single set of brackets is much of a deal.

PR #5765, opened 2022-01-10
Fix block selection expansion with Ctrl + RMB

kchibisov:
When 'ExpandSelection' binding was added only default binding for
RightClick was added, however to expand block selection holding control
when doing a click is required, so this commit adds a binding for
'RMB + Control'.


chrisduerr (approved):
Great catch, thanks.

PR #5766, opened 2022-01-11
Fix use-after-free when formatting Wayland protocol error

kchibisov:


chrisduerr (approved):
I'd assume this is 0.10.0 relevant?

kchibisov:
I guess we can add to v0.10.0, though, it's not like it's easy to trigger this issue and we'll likely be crashing anyway in that case.

chrisduerr:
> I guess we can add to v0.10.0, though, it's not like it's easy to trigger this issue and we'll likely be crashing anyway in that case.

Just to be clear: This issue occurs when winit is in the process of crashing due to a protocol error, thus causing that error to be logged? Debug output isn't logged, so other protocol errors shouldn't run into this, right?

kchibisov:
>Just to be clear: This issue occurs when winit is in the process of crashing due to a protocol error, thus causing that error to be logged? Debug output isn't logged, so other protocol errors shouldn't run into this, right?

Pretty much it.

PR #5768, opened 2022-01-11
Use `with_position` instead of `set_outer_position`

kchibisov:
This uses `with_position` method on a `WindowBuilder` instead of setting
window position on a created window later on.

--
I wasn't able to test this, since it's not something that is allowed on Wayland. So leaving it up to you.

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -9,7 +9,7 @@ use std::sync::atomic::Ordering;
 use std::time::Instant;
 use std::{f64, mem};
 
-use glutin::dpi::{PhysicalPosition, PhysicalSize};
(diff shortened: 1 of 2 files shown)


chrisduerr (approved):
>  I wasn't able to test this, since it's not something that is allowed on Wayland. So leaving it up to you.

You did test that it doesn't explode on Wayland though, right? I'd assume by "not allowed" you just mean it's ignored?

On X11 this seems to work just fine, though the previous one also did for me.

Since this shouldn't introduce any significant changes, it's probably fine to omit the changelog entry?

kchibisov:
>Since this shouldn't introduce any significant changes, it's probably fine to omit the changelog entry?

There's no need, since it's internal change, but for user everything should be the same.

PR #2724, opened 2019-08-16
Track and report damage to compatible compositors

kennylevinsen:
Implement damage tracking and reporting.

This allows compositors to only process damaged (that is, updated) regions of our window buffer, which for larger window sizes (think 4k) should significantly reduce compositing workload under compositors that support/honor it, which is good for performance and battery life.

On Wayland, clients are expected to report always report correct damage, so this makes us a good (or at least, better) citizen there. It can also aid remote desktop (waypipe, RDP, VNC, ...) and other types of screencopy by having damage bubble up correctly.

Damage tracking has been implemented in a somewhat naïve fashion, but I believe it is a fair start.

Note that this currently depends on a fork of glutin where support has been added.

Fixes #3186.

PR #5773, opened 2022-01-12
Add damage tracking and reporting to compatible compositors

### LINKED ISSUE #3186: Properly inform compositors of damaged regions
### System

OS: Linux
Version: alacritty 0.4.0
Linux/BSD: sway

### Logs

N/A

---

alacritty currently appears to be damaging the whole window even when there are little or no changes to the contents. This is important for [energy efficiency](https://mozillagfx.wordpress.com/2019/10/22/dramatically-reduced-power-usage-in-firefox-70-on-macos-with-core-animation/).

See also https://emersion.fr/blog/2019/intro-to-damage-tracking/ for motivation and explanation and https://bugzilla.mozilla.org/show_bug.cgi?id=1484812 for a related firefox issue.

kchibisov:
This allows compositors to only process damaged (that is, updated)
regions of our window buffer, which for larger window sizes (think 4k)
should significantly reduce compositing workload under compositors that
support/honor it, which is good for performance, battery life and lower
latency over remote connections like VNC.

On Wayland, clients are expected to always report correct damage, so
this makes us a good citizen there. It can also aid remote desktop
(waypipe, rdp, vnc, ...) and other types of screencopy by having damage
bubble up correctly.

Fixes #3186.

--
This uses ideas from #2724, since they were performing quite good. For example there's no damage updates in `Term::input` at all, except when wrapping a line. The way damage tracking is implement is that we create an empty information about damage state of each line and store it on a terminal and update damage bounds with left most and right most damaged cells respectively damaging everything between later on.

So this approach allows us to damage only cursor when for example starting writing and ending, and when cursor is unconditionally moved somewhere(jumping with goto, etc), leaving everything that was inserted between bounds with something like `Term::input` automatically damaged.

In cases where damaging becomes complicated we fallback to damaging just entire screen (e.g. when scrolling).

The approach is performing the same as master for me. So this patch is in a good state IMHO, however it requires developing a test suite for it, since not all systems allow to report damage to compositors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
- - Option `font.builtin_box_drawing` to disable the built-in font for drawing box characters
+- Option `font.builtin_box_drawing` to disable the built-in font for drawing box characters
+- Track and report surface damage information to Wayland compositors
 
 ### Changed
 
--- a/alacritty/src/display/damage.rs
+++ b/alacritty/src/display/damage.rs
@@ -1,0 +1,86 @@
+use std::cmp;
+use std::iter::Peekable;
+
+use glutin::Rect;
+
+use alacritty_terminal::term::{LineDamageBounds, SizeInfo, TermDamageIterator};
+
+/// Iterator which converts `alacritty_terminal` damage information into renderer damaged rects.
+pub struct RenderDamageIterator<'a> {
+    damaged_lines: Peekable<TermDamageIterator<'a>>,
+    size_info: SizeInfo<u32>,
+}
+
+impl<'a> RenderDamageIterator<'a> {
+    pub fn new(damaged_lines: TermDamageIterator<'a>, size_info: SizeInfo<u32>) -> Self {
+        Self { damaged_lines: damaged_lines.peekable(), size_info }
+    }
+
+    #[inline]
+    fn rect_for_line(&self, line_damage: LineDamageBounds) -> Rect {
+        let size_info = &self.size_info;
+        let y_top = size_info.height() - size_info.padding_y();
+        let x = size_info.padding_x() + line_damage.left as u32 * size_info.cell_width();
+        let y = y_top - (line_damage.line + 1) as u32 * size_info.cell_height();
+        let width = (line_damage.right - line_damage.left + 1) as u32 * size_info.cell_width();
+        Rect { x, y, height: size_info.cell_height(), width }
+    }
+
+    // Make sure to damage near cells to include wide chars.
+    #[inline]
+    fn overdamage(&self, mut rect: Rect) -> Rect {
+        let size_info = &self.size_info;
+        rect.x = rect.x.saturating_sub(size_info.cell_width());
+        rect.width = cmp::min(size_info.width() - rect.x, rect.width + 2 * size_info.cell_width());
+        rect.y = rect.y.saturating_sub(size_info.cell_height() / 2);
+        rect.height = cmp::min(size_info.height() - rect.y, rect.height + size_info.cell_height());
+
+        rect
+    }
+}
+
+impl<'a> Iterator for RenderDamageIterator<'a> {
+    type Item = Rect;
+
+    fn next(&mut self) -> Option<Rect> {
+        let line = self.damaged_lines.next()?;
+        let mut total_damage_rect = self.overdamage(self.rect_for_line(line));
+
+        // Merge rectangles which overlap with each other.
+        while let Some(line) = self.damaged_lines.peek().copied() {
+            let next_rect = self.overdamage(self.rect_for_line(line));
+            if !rects_overlap(total_damage_rect, next_rect) {
+                break;
+            }
+
+            total_damage_rect = merge_rects(total_damage_rect, next_rect);
+            let _ = self.damaged_lines.next();
+        }
+
+        Some(total_damage_rect)
+    }
+}
+
+/// Check if two given [`glutin::Rect`] overlap.
+fn rects_overlap(lhs: Rect, rhs: Rect) -> bool {
+    !(
+        // `lhs` is left of `rhs`.
+        lhs.x + lhs.width < rhs.x
+        // `lhs` is right of `rhs`.
+        || rhs.x + rhs.width < lhs.x
+        // `lhs` is below `rhs`.
+        || lhs.y + lhs.height < rhs.y
+        // `lhs` is above `rhs`.
+        || rhs.y + rhs.height < lhs.y
+    )
+}
+
+/// Merge two [`glutin::Rect`] by producing the smallest rectangle that contains both.
+#[inline]
+fn merge_rects(lhs: Rect, rhs: Rect) -> Rect {
+    let left_x = cmp::min(lhs.x, rhs.x);
+    let right_x = cmp::max(lhs.x + lhs.width, rhs.x + rhs.width);
+    let y_top = cmp::max(lhs.y + lhs.height, rhs.y + rhs.height);
+    let y_bottom = cmp::min(lhs.y, rhs.y);
+    Rect { x: left_x, y: y_bottom, width: right_x - left_x, height: y_top - y_bottom }
+}
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -1,13 +1,12 @@
 //! The display subsystem including window management, font rasterization, and
 //! GPU drawing.
 
-use std::cmp::min;
 use std::convert::TryFrom;
 use std::fmt::{self, Formatter};
 #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
 use std::sync::atomic::Ordering;
 use std::time::Instant;
-use std::{f64, mem};
+use std::{cmp, mem};
 
 use glutin::dpi::PhysicalSize;
 use glutin::event::ModifiersState;
@@ -15,6 +14,7 @@ use glutin::event_loop::EventLoopWindowTarget;
 #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
 use glutin::platform::unix::EventLoopWindowTargetExtUnix;
 use glutin::window::CursorIcon;
+use glutin::Rect as DamageRect;
 use log::{debug, info};
 use parking_lot::MutexGuard;
 use unicode_width::UnicodeWidthChar;
@@ -24,12 +24,16 @@ use wayland_client::EventQueue;
 use crossfont::{self, Rasterize, Rasterizer};
 
 use alacritty_terminal::ansi::NamedColor;
+use alacritty_terminal::config::MAX_SCROLLBACK_LINES;
 use alacritty_terminal::event::{EventListener, OnResize};
 use alacritty_terminal::grid::Dimensions as _;
 use alacritty_terminal::index::{Column, Direction, Line, Point};
-use alacritty_terminal::selection::Selection;
+use alacritty_terminal::selection::{Selection, SelectionRange};
 use alacritty_terminal::term::cell::Flags;
-use alacritty_terminal::term::{SizeInfo, Term, TermMode, MIN_COLUMNS, MIN_SCREEN_LINES};
+use alacritty_terminal::term::color::Rgb;
+use alacritty_terminal::term::{
+    SizeInfo, Term, TermDamage, TermMode, MIN_COLUMNS, MIN_SCREEN_LINES,
+};
 
 use crate::config::font::Font;
 #[cfg(not(windows))]
@@ -40,6 +44,7 @@ use crate::display::bell::VisualBell;
 use crate::display::color::List;
 use crate::display::content::RenderableContent;
 use crate::display::cursor::IntoRects;
+use crate::display::damage::RenderDamageIterator;
 use crate::display::hint::{HintMatch, HintState};
 use crate::display::meter::Meter;
 use crate::display::window::Window;
@@ -55,6 +60,7 @@ pub mod window;
 
 mod bell;
 mod color;
+mod damage;
 mod meter;
 
 /// Maximum number of linewraps followed outside of the viewport during search highlighting.
@@ -65,6 +71,9 @@ const FORWARD_SEARCH_LABEL: &str = "Search: ";
 
 /// Label for the backward terminal search bar.
 const BACKWARD_SEARCH_LABEL: &str = "Backward Search: ";
+
+/// Color which is used to highlight damaged rects when debugging.
+const DAMAGE_RECT_COLOR: Rgb = Rgb { r: 255, g: 0, b: 255 };
 
 #[derive(Debug)]
 pub enum Error {
@@ -193,6 +202,9 @@ pub struct Display {
     /// Unprocessed display updates.
     pub pending_update: DisplayUpdate,
 
+    is_damage_supported: bool,
+    debug_damage: bool,
+    damage_rects: Vec<DamageRect>,
     renderer: QuadRenderer,
     glyph_cache: GlyphCache,
     meter: Meter,
@@ -319,6 +331,13 @@ impl Display {
         }
 
         let hint_state = HintState::new(config.hints.alphabet());
+        let is_damage_supported = window.swap_buffers_with_damage_supported();
+        let debug_damage = config.debug.highlight_damage;
+        let damage_rects = if is_damage_supported || debug_damage {
+            Vec::with_capacity(size_info.screen_lines())
+        } else {
+            Vec::new()
+        };
 
         Ok(Self {
             window,
@@ -335,6 +354,9 @@ impl Display {
             visual_bell: VisualBell::from(&config.bell),
             colors: List::from(&config.colors),
             pending_update: Default::default(),
+            is_damage_supported,
+            debug_damage,
+            damage_rects,
         })
     }
 
@@ -457,8 +479,56 @@ impl Display {
         self.window.resize(physical);
         self.renderer.resize(&self.size_info);
 
+        if self.collect_damage() {
+            let lines = self.size_info.screen_lines();
+            if lines > self.damage_rects.len() {
+                self.damage_rects.reserve(lines);
+            } else {
+                self.damage_rects.shrink_to(lines);
+            }
+        }
+
         info!("Padding: {} x {}", self.size_info.padding_x(), self.size_info.padding_y());
         info!("Width: {}, Height: {}", self.size_info.width(), self.size_info.height());
+    }
+
+    fn update_damage<T: EventListener>(
+        &mut self,
+        terminal: &mut MutexGuard<'_, Term<T>>,
+        selection_range: Option<SelectionRange>,
+        search_state: &SearchState,
+    ) {
+        let requires_full_damage = self.visual_bell.intensity() != 0.
+            || self.hint_state.active()
+            || search_state.regex().is_some();
+        if requires_full_damage {
+            terminal.mark_fully_damaged();
+        }
+
+        self.damage_highlighted_hints(terminal);
+        let size_info: SizeInfo<u32> = self.size_info.into();
+        match terminal.damage(selection_range) {
+            TermDamage::Full => {
+                let screen_rect =
+                    DamageRect { x: 0, y: 0, width: size_info.width(), height: size_info.height() };
+                self.damage_rects.push(screen_rect);
+            },
+            TermDamage::Partial(damaged_lines) => {
+                let damaged_rects = RenderDamageIterator::new(damaged_lines, size_info);
+                for damaged_rect in damaged_rects {
+                    self.damage_rects.push(damaged_rect);
+                }
+            },
+        }
+        terminal.reset_damage();
+
+        // Ensure that the content requiring full damage is cleaned up again on the next frame.
+        if requires_full_damage {
+            terminal.mark_fully_damaged();
+        }
+
+        // Damage highlighted hints for the next frame as well, so we'll clear them.
+        self.damage_highlighted_hints(terminal);
     }
 
     /// Draw the screen.
@@ -468,7 +538,7 @@ impl Display {
     /// This call may block if vsync is enabled.
     pub fn draw<T: EventListener>(
         &mut self,
-        terminal: MutexGuard<'_, Term<T>>,
+        mut terminal: MutexGuard<'_, Term<T>>,
         message_buffer: &MessageBuffer,
         config: &UiConfig,
         search_state: &SearchState,
@@ -479,6 +549,7 @@ impl Display {
         for cell in &mut content {
             grid_cells.push(cell);
         }
+        let selection_range = content.selection_range();
         let background_color = content.color(NamedColor::Background as usize);
         let display_offset = content.display_offset();
         let cursor = content.cursor();
@@ -490,6 +561,11 @@ impl Display {
 
         let vi_mode = terminal.mode().contains(TermMode::VI);
         let vi_mode_cursor = if vi_mode { Some(terminal.vi_mode_cursor) } else { None };
+
+        if self.collect_damage() {
+            self.damage_rects.clear();
+            self.update_damage(&mut terminal, selection_range, search_state);
+        }
 
         // Drop terminal as early as possible to free lock.
         drop(terminal);
@@ -549,11 +625,9 @@ impl Display {
             self.draw_line_indicator(config, &size_info, total_lines, None, display_offset);
         }
 
-        // Push the cursor rects for rendering.
-        if let Some(cursor) = cursor {
-            for rect in cursor.rects(&size_info, config.terminal_config.cursor.thickness()) {
-                rects.push(rect);
-            }
+        // Draw cursor.
+        for rect in cursor.rects(&size_info, config.terminal_config.cursor.thickness()) {
+            rects.push(rect);
         }
 
         // Push visual bell after url/underline/strikeout rects.
@@ -568,6 +642,10 @@ impl Display {
                 visual_bell_intensity as f32,
             );
             rects.push(visual_bell_rect);
+        }
+
+        if self.debug_damage {
+            self.highlight_damage(&mut rects);
         }
 
         if let Some(message) = message_buffer.message() {
@@ -636,7 +714,12 @@ impl Display {
         #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
         self.request_frame(&self.window);
 
-        self.window.swap_buffers();
+        // Clearing debug highlights from the previous frame requires full redraw.
+        if self.is_damage_supported && !self.debug_damage {
+            self.window.swap_buffers_with_damage(&self.damage_rects);
+        } else {
+            self.window.swap_buffers();
+        }
 
         #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
         if self.is_x11 {
@@ -651,6 +734,7 @@ impl Display {
 
     /// Update to a new configuration.
     pub fn update_config(&mut self, config: &UiConfig) {
+        self.debug_damage = config.debug.highlight_damage;
         self.visual_bell.update_config(&config.bell);
         self.colors = List::from(&config.colors);
     }
@@ -722,7 +806,7 @@ impl Display {
         let num_cols = size_info.columns();
         let label_len = search_label.chars().count();
         let regex_len = formatted_regex.chars().count();
-        let truncate_len = min((regex_len + label_len).saturating_sub(num_cols), regex_len);
+        let truncate_len = cmp::min((regex_len + label_len).saturating_sub(num_cols), regex_len);
         let index = formatted_regex.char_indices().nth(truncate_len).map(|(i, _c)| i).unwrap_or(0);
         let truncated_regex = &formatted_regex[index..];
 
@@ -758,13 +842,15 @@ impl Display {
             return;
         }
 
-        let glyph_cache = &mut self.glyph_cache;
-
         let timing = format!("{:.3} usec", self.meter.average());
         let point = Point::new(size_info.screen_lines().saturating_sub(2), Column(0));
         let fg = config.colors.primary.background;
         let bg = config.colors.normal.red;
 
+        // Damage the entire line.
+        self.damage_from_point(point, self.size_info.columns() as u32);
+
+        let glyph_cache = &mut self.glyph_cache;
         self.renderer.with_api(config, size_info, |mut api| {
             api.draw_string(glyph_cache, point, fg, bg, &timing);
         });
@@ -779,8 +865,26 @@ impl Display {
         obstructed_column: Option<Column>,
         line: usize,
     ) {
+        const fn num_digits(mut number: u32) -> usize {
+            let mut res = 0;
+            loop {
+                number /= 10;
+                res += 1;
+                if number == 0 {
+                    break res;
+                }
+            }
+        }
+
         let text = format!("[{}/{}]", line, total_lines - 1);
         let column = Column(size_info.columns().saturating_sub(text.len()));
+        let point = Point::new(0, column);
+
+        // Damage the maximum possible length of the format text, which could be achieved when
+        // using `MAX_SCROLLBACK_LINES` as current and total lines adding a `3` for formatting.
+        const MAX_LEN: usize = num_digits(MAX_SCROLLBACK_LINES) + 3;
+        self.damage_from_point(Point::new(0, point.column - MAX_LEN), MAX_LEN as u32 * 2);
+
         let colors = &config.colors;
         let fg = colors.line_indicator.foreground.unwrap_or(colors.primary.background);
         let bg = colors.line_indicator.background.unwrap_or(colors.primary.foreground);
@@ -789,8 +893,57 @@ impl Display {
         if obstructed_column.map_or(true, |obstructed_column| obstructed_column < column) {
             let glyph_cache = &mut self.glyph_cache;
             self.renderer.with_api(config, size_info, |mut api| {
-                api.draw_string(glyph_cache, Point::new(0, column), fg, bg, &text);
+                api.draw_string(glyph_cache, point, fg, bg, &text);
             });
+        }
+    }
+
+    /// Damage `len` starting from a `point`.
+    #[inline]
+    fn damage_from_point(&mut self, point: Point<usize>, len: u32) {
+        if !self.collect_damage() {
+            return;
+        }
+
+        let size_info: SizeInfo<u32> = self.size_info.into();
+        let x = size_info.padding_x() + point.column.0 as u32 * size_info.cell_width();
+        let y_top = size_info.height() - size_info.padding_y();
+        let y = y_top - (point.line as u32 + 1) * size_info.cell_height();
+        let width = len as u32 * size_info.cell_width();
+        self.damage_rects.push(DamageRect { x, y, width, height: size_info.cell_height() })
+    }
+
+    /// Damage currently highlighted `Display` hints.
+    #[inline]
+    fn damage_highlighted_hints<T: EventListener>(&self, terminal: &mut Term<T>) {
+        let display_offset = terminal.grid().display_offset();
+        for hint in self.highlighted_hint.iter().chain(&self.vi_highlighted_hint) {
+            for point in (hint.bounds.start().line.0..=hint.bounds.end().line.0).flat_map(|line| {
+                point_to_viewport(display_offset, Point::new(Line(line), Column(0)))
+            }) {
+                terminal.damage_line(point.line, 0, terminal.columns() - 1);
+            }
+        }
+    }
+
+    /// Returns `true` if damage information should be collected, `false` otherwise.
+    #[inline]
+    fn collect_damage(&self) -> bool {
+        self.is_damage_supported || self.debug_damage
+    }
+
+    /// Highlight damaged rects.
+    ///
+    /// This function is for debug purposes only.
+    fn highlight_damage(&self, render_rects: &mut Vec<RenderRect>) {
+        for damage_rect in &self.damage_rects {
+            let x = damage_rect.x as f32;
+            let height = damage_rect.height as f32;
+            let width = damage_rect.width as f32;
+            let y = self.size_info.height() - damage_rect.y as f32 - height;
+            let render_rect = RenderRect::new(x, y, width, height, DAMAGE_RECT_COLOR, 0.5);
+
+            render_rects.push(render_rect);
         }
     }
 
@@ -824,12 +977,14 @@ impl Drop for Display {
 }
 
 /// Convert a terminal point to a viewport relative point.
+#[inline]
 pub fn point_to_viewport(display_offset: usize, point: Point) -> Option<Point<usize>> {
     let viewport_line = point.line.0 + display_offset as i32;
     usize::try_from(viewport_line).ok().map(|line| Point::new(line, point.column))
 }
 
 /// Convert a viewport relative point to a terminal point.
+#[inline]
 pub fn viewport_to_point(display_offset: usize, point: Point<usize>) -> Point {
     let line = Line(point.line as i32) - display_offset;
     Point::new(line, point.column)
--- a/alacritty_terminal/src/selection.rs
+++ b/alacritty_terminal/src/selection.rs
@@ -41,6 +41,7 @@ pub struct SelectionRange {
 
 impl SelectionRange {
     pub fn new(start: Point, end: Point, is_block: bool) -> Self {
+        assert!(start <= end);
         Self { start, end, is_block }
     }
 }
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -1,9 +1,8 @@
 //! Exports the `Term` type which is a high-level API for the Grid.
 
-use std::cmp::{max, min};
 use std::ops::{Index, IndexMut, Range};
 use std::sync::Arc;
-use std::{mem, ptr, str};
+use std::{cmp, mem, ptr, slice, str};
 
 use bitflags::bitflags;
 use log::{debug, trace};
@@ -62,7 +61,7 @@ bitflags! {
         const ALTERNATE_SCROLL    = 0b0000_1000_0000_0000_0000;
         const VI                  = 0b0001_0000_0000_0000_0000;
         const URGENCY_HINTS       = 0b0010_0000_0000_0000_0000;
-        const ANY                 = std::u32::MAX;
+        const ANY                 = u32::MAX;
     }
 }
 
@@ -77,24 +76,24 @@ impl Default for TermMode {
 
 /// Terminal size info.
 #[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq)]
-pub struct SizeInfo {
+pub struct SizeInfo<T = f32> {
     /// Terminal window width.
-    width: f32,
+    width: T,
 
     /// Terminal window height.
-    height: f32,
+    height: T,
 
     /// Width of individual cell.
-    cell_width: f32,
+    cell_width: T,
 
     /// Height of individual cell.
-    cell_height: f32,
+    cell_height: T,
 
     /// Horizontal window padding.
-    padding_x: f32,
+    padding_x: T,
 
     /// Vertical window padding.
-    padding_y: f32,
+    padding_y: T,
 
     /// Number of lines in the viewport.
     screen_lines: usize,
@@ -103,7 +102,54 @@ pub struct SizeInfo {
     columns: usize,
 }
 
-impl SizeInfo {
+impl From<SizeInfo<f32>> for SizeInfo<u32> {
+    fn from(size_info: SizeInfo<f32>) -> Self {
+        Self {
+            width: size_info.width as u32,
+            height: size_info.height as u32,
+            cell_width: size_info.cell_width as u32,
+            cell_height: size_info.cell_height as u32,
+            padding_x: size_info.padding_x as u32,
+            padding_y: size_info.padding_y as u32,
+            screen_lines: size_info.screen_lines,
+            columns: size_info.screen_lines,
+        }
+    }
+}
+
+impl<T: Clone + Copy> SizeInfo<T> {
+    #[inline]
+    pub fn width(&self) -> T {
+        self.width
+    }
+
+    #[inline]
+    pub fn height(&self) -> T {
+        self.height
+    }
+
+    #[inline]
+    pub fn cell_width(&self) -> T {
+        self.cell_width
+    }
+
+    #[inline]
+    pub fn cell_height(&self) -> T {
+        self.cell_height
+    }
+
+    #[inline]
+    pub fn padding_x(&self) -> T {
+        self.padding_x
+    }
+
+    #[inline]
+    pub fn padding_y(&self) -> T {
+        self.padding_y
+    }
+}
+
+impl SizeInfo<f32> {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
         width: f32,
@@ -120,10 +166,10 @@ impl SizeInfo {
         }
 
         let lines = (height - 2. * padding_y) / cell_height;
-        let screen_lines = max(lines as usize, MIN_SCREEN_LINES);
+        let screen_lines = cmp::max(lines as usize, MIN_SCREEN_LINES);
 
         let columns = (width - 2. * padding_x) / cell_width;
-        let columns = max(columns as usize, MIN_COLUMNS);
+        let columns = cmp::max(columns as usize, MIN_COLUMNS);
 
         SizeInfo {
             width,
@@ -139,7 +185,7 @@ impl SizeInfo {
 
     #[inline]
     pub fn reserve_lines(&mut self, count: usize) {
-        self.screen_lines = max(self.screen_lines.saturating_sub(count), MIN_SCREEN_LINES);
+        self.screen_lines = cmp::max(self.screen_lines.saturating_sub(count), MIN_SCREEN_LINES);
     }
 
     /// Check if coordinates are inside the terminal grid.
@@ -151,36 +197,6 @@ impl SizeInfo {
             && x > self.padding_x as usize
             && y <= (self.padding_y + self.screen_lines as f32 * self.cell_height) as usize
             && y > self.padding_y as usize
-    }
-
-    #[inline]
-    pub fn width(&self) -> f32 {
-        self.width
-    }
-
-    #[inline]
-    pub fn height(&self) -> f32 {
-        self.height
-    }
-
-    #[inline]
-    pub fn cell_width(&self) -> f32 {
-        self.cell_width
-    }
-
-    #[inline]
-    pub fn cell_height(&self) -> f32 {
-        self.cell_height
-    }
-
-    #[inline]
-    pub fn padding_x(&self) -> f32 {
-        self.padding_x
-    }
-
-    #[inline]
-    pub fn padding_y(&self) -> f32 {
-        self.padding_y
     }
 
     /// Calculate padding to spread it evenly around the terminal content.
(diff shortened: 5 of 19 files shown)


kchibisov:
@perfbot

kchibisov:
@chrisduerr I've added an option to highlight damage information which were reporting. Should help with testing on systems not supporting damage tracking and also its visualizing.

Could be run via `alacrtitty -o debug.highlight_damage=true`.

Note that in that case it stops reporting damage information to compositor (since you need to cleanup things you've highlighted), but it matches what sway highlights.

chrisduerr:
@kchibisov The damage highlighting seems broken for me, everything is just permanently pink?


https://user-images.githubusercontent.com/8886672/149607705-b97b3e22-3cbd-4d47-9296-b3c4952ad38d.mp4

kchibisov:
@chrisduerr I've rebased and pushed the tests. Should be ready for review.

alacritty_terminal/src/selection.rs
```diff
@@ -41,6 +41,7 @@ pub struct SelectionRange {
 
 impl SelectionRange {
     pub fn new(start: Point, end: Point, is_block: bool) -> Self {
+        assert!(start <= end);
```
chrisduerr, line 44: Did you run into any issues with this, or why add this assert?

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
+use std::cmp;
+
+use glutin::Rect;
+
+use alacritty_terminal::selection::SelectionRange;
+use alacritty_terminal::term::{DamagedLine, RenderableCursor, SizeInfo, TermDamage};
+use alacritty_terminal::vi_mode::ViModeCursor;
+
+/// Minimum part of damaged square when intersecting damaged rectangles.
+pub const MIN_DAMAGED_SQUARE_PART: f32 = 0.4;
```
chrisduerr, line 10:
```suggestion
const MIN_DAMAGED_SQUARE_PART: f32 = 0.4;
```

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+
+/// An error when we for look ahead in iterator, used to try to collect rectangles which pretty
+/// close to each other, but don't fit into `MIN_DAMAGED_SQUARE_PART` just by looking on the first
+/// next damaged rectangle.
+pub const MIN_DAMAGED_SQUARE_PART_ERR: f32 = MIN_DAMAGED_SQUARE_PART / 2.;
```
chrisduerr, line 15:
```suggestion
const MIN_DAMAGED_SQUARE_PART_ERR: f32 = MIN_DAMAGED_SQUARE_PART / 2.;
```

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+use alacritty_terminal::selection::SelectionRange;
+use alacritty_terminal::term::{DamagedLine, RenderableCursor, SizeInfo, TermDamage};
+use alacritty_terminal::vi_mode::ViModeCursor;
+
+/// Minimum part of damaged square when intersecting damaged rectangles.
+pub const MIN_DAMAGED_SQUARE_PART: f32 = 0.4;
+
+/// An error when we for look ahead in iterator, used to try to collect rectangles which pretty
+/// close to each other, but don't fit into `MIN_DAMAGED_SQUARE_PART` just by looking on the first
+/// next damaged rectangle.
```
chrisduerr, line 14: Looking just at this comment I have no clue what you mean, could you maybe try to reword it a bit?

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+pub const MIN_DAMAGED_SQUARE_PART_ERR: f32 = MIN_DAMAGED_SQUARE_PART / 2.;
+
+pub struct RendererDamage<'a> {
```
chrisduerr, line 17: Small comment what this struct is used for might be nice.

This is all damage information required by the renderer for one frame, right?

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+    index: usize,
```
chrisduerr, line 18: I'm not sure it's a good idea to have iterator indices stored on the struct itself? If so it definitely needs to be renamed to make things more clear. But I really don't like having mutable iterator state here.

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+    rect: Option<Rect>,
```
chrisduerr, line 19: What is this exactly, the total damage rectangle, so all other damage combined? Just `rect` doesn't really tell me much.

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+impl<'a> RendererDamage<'a> {
+    pub fn new(
+        terminal_damage: &'a mut TermDamage,
+        size_info: SizeInfo<u32>,
+        cursor: RenderableCursor,
+        vi_cursor: Option<ViModeCursor>,
+        selection: Option<SelectionRange>,
+        display_offset: usize,
+    ) -> Self {
+        let fully_damaged = terminal_damage.take_fully_damaged();
```
chrisduerr, line 33: I really don't like the idea of using a boolean check method to also mutate state. That just doesn't sit right with me.

Should we maybe have a `damage` method on the terminal that returns an enum containing either `Full`, or `Partial(Vec<DamagedLine>)`? Seems like a much more proper solution. Then we can just manually clear damage.

Maybe make that method clear all damage too, then it would at least be a reasonable option. But I'm not sure I would.

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+        let terminal_damage =
+            terminal_damage.damage_mut(cursor, vi_cursor, selection, display_offset);
+
+        let (index, rect) = if fully_damaged {
+            let rect = Rect { x: 0, y: 0, width: size_info.width(), height: size_info.height() };
+            (terminal_damage.len(), rect)
```
chrisduerr, line 39: Collecting all the damage lines into a vector, just to then index that vector to the end to skip iterating it using `len()` seems rather hacky. Really not a fan.

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+        } else {
+            // Terminal cursor is always damaged there's at least one damaged line.
```
chrisduerr, line 41:
```suggestion
            // Terminal cursor is always damaged so there's at least one damaged line.
```

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+            let index = terminal_damage.iter().position(|&line| line.is_damaged()).unwrap();
```
chrisduerr, line 42: Wait, this is an iterator over `Vec<DamagedLine>`, why do we need to check if `line.is_damaged()`. It's a damaged line, so surely they're all damaged?!

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+            // Advance index into the next unprocessed element.
+            (index + 1, rect)
+        };
+
+        Self { index, rect: Some(rect), terminal_damage, size_info }
+    }
+
+    /// Creates a new rectangle in a way that 'lhs' and 'rhs' are inside of it.
+    #[inline]
+    fn intersect(lhs: Rect, rhs: Rect) -> Rect {
```
chrisduerr, line 56: Intersection would be a rectangle containing only area both `lhs` and `rhs` contain. Your comment doesn't describe intersection, it describes a union (since it's rectangles it's even more than a true union).

The code also suggests this isn't an intersection, so please rename the function. I think smithay calls this `merge` for example(https://github.com/Smithay/smithay/blob/master/src/utils/geometry.rs#L955)?

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    #[inline]
+    fn rect_for_line(size_info: &SizeInfo<u32>, line: usize, line_damage: DamagedLine) -> Rect {
+        let x = size_info.padding_x() + line_damage.left as u32 * size_info.cell_width();
+        let y = size_info.height()
+            - size_info.padding_y()
+            - (line + 1) as u32 * size_info.cell_height();
+        let width = (line_damage.right - line_damage.left + 1) as u32 * size_info.cell_width();
+        Rect { x, y, height: size_info.cell_height(), width }
```
chrisduerr, line 71: Same comment as I put in the mod.rs thing: Would be nice if we could improve formatting a bit. I'd suggest extracting cell_width/cell_height maybe.

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+            terminal_damage[index].reset();
+
+            // Advance index into the next unprocessed element.
+            (index + 1, rect)
+        };
+
+        Self { index, rect: Some(rect), terminal_damage, size_info }
+    }
+
+    /// Creates a new rectangle in a way that 'lhs' and 'rhs' are inside of it.
```
chrisduerr, line 54:
```suggestion
    /// Create a new rectangle which contains both 'lhs' and 'rhs'.
```

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+        let x = size_info.padding_x() + line_damage.left as u32 * size_info.cell_width();
+        let y = size_info.height()
+            - size_info.padding_y()
+            - (line + 1) as u32 * size_info.cell_height();
+        let width = (line_damage.right - line_damage.left + 1) as u32 * size_info.cell_width();
+        Rect { x, y, height: size_info.cell_height(), width }
+    }
+
+    /// Adds extra damage area around the `rect` to account for glyphs that are larger than a glyph
+    /// cell or weird wide characters.
```
chrisduerr, line 75: Should use imperative.

```suggestion
    /// Add extra damage area around the rect to account for glyphs that are larger than their cell.
```

I'd assume that by `or weird wide characters` you mean wide characters extending beyond their cell, not that wide characters require over-damage to be rendered correctly? I'd assume the terminal always damages wide characters like two halfwidth.

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+    #[inline]
+    fn overdamage(&self, mut rect: Rect) -> Rect {
+        rect.x = rect.x.saturating_sub(self.size_info.cell_width());
+        rect.width =
+            cmp::min(self.size_info.width() - rect.x, rect.width + 2 * self.size_info.cell_width());
+        rect.y = rect.y.saturating_sub(self.size_info.cell_height() / 2);
+        rect.height =
+            cmp::min(self.size_info.height() - rect.y, rect.height + self.size_info.cell_height());
```
chrisduerr, line 83: Pretty sure we can improve formatting here a good bit just by simply doing `let size_info = &self.size_info` or something like that?

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+
+        rect
+    }
+
+    #[inline]
+    fn rect_square(rect: &Rect) -> u32 {
+        rect.width * rect.height
+    }
```
chrisduerr, line 91:
```suggestion
    #[inline]
    fn rect_area(rect: &Rect) -> u32 {
        rect.width * rect.height
    }
```

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+}
+
+impl<'a> Iterator for RendererDamage<'_> {
```
chrisduerr, line 94: More of a personal note: I haven't reviewed this iterator method yet, would like some feedback first if it should be on `RendererDamage` or not (just to get the review out quicker).

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+    use super::*;
+
+    use crate::config::UiConfig;
+    use alacritty_terminal::grid::Dimensions;
+    use alacritty_terminal::term::{
+        DamagedLine, RenderableCursor as TermRenderableCursor, SizeInfo, Term,
+    };
+
+    #[test]
+    fn damage_fully_damaged() {
```
chrisduerr, line 218: This module is already called `damage` we don't need to prefix every test method with `damage_`?

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    use crate::config::UiConfig;
+    use alacritty_terminal::grid::Dimensions;
+    use alacritty_terminal::term::{
+        DamagedLine, RenderableCursor as TermRenderableCursor, SizeInfo, Term,
+    };
```
chrisduerr, line 215:
```suggestion
    use alacritty_terminal::grid::Dimensions;
    use alacritty_terminal::term::{
        DamagedLine, RenderableCursor as TermRenderableCursor, SizeInfo, Term,
    };
    
    use crate::config::UiConfig;
```

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+        // Update information for the next iteration.
+        self.rect = next_rect;
+        self.index = next_index;
+
+        Some(self.overdamage(rect))
+    }
+}
+
+#[cfg(test)]
+mod test {
```
chrisduerr, line 208: I actually also always get confused by this, but test modules in Alacritty are always `mod tests`, so plural.

Could you maybe also fix this in `builtin_font` while you're at it? I didn't notice that in the review there.

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+        let mut term = Term::new(&UiConfig::default().terminal_config, size, ());
+        let size: SizeInfo<u32> = size.into();
+        let num_cols = term.columns();
+
+        let cursor_point = term.grid().cursor.point;
+        let cursor = TermRenderableCursor { point: cursor_point, shape: Default::default() };
+
+        let term_damage = term.damage();
+        // Ensure it's not damaged.
+        term_damage.take_fully_damaged();
```
chrisduerr, line 273: This confuses me, why do we need to ensure it's not damaged after we *just* started the test? This also applies to your other tests.

Is a newly created terminal fully damaged?

Why is the comment `ensure`, when you're not checking anything? If anything that should say "clear existing damage" or something.

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,345 @@
...
+
+        let mut second_rect = RendererDamage::rect_for_line(
+            &size,
+            second_rect_line,
+            renderer_damage.terminal_damage[second_rect_line],
+        );
+        second_rect.y -= size.cell_height();
+        second_rect.height += size.cell_height();
+        second_rect.width = size.width() - second_rect.x;
+        second_rect = renderer_damage.overdamage(second_rect);
```
chrisduerr, line 307: There's **a lot** of complexity and math in this test. I'm not saying it justifies removal of the test but otoh I'm not able to easily verify this is **correct** either just by looking at it. There could be a billion off-by-ones in there. So if I ever had to make a big change and this test would bother me, I'd probably just toss it.

Just in case you have ideas on how to possibly clean this up, it certainly would be a great candidate.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1,9 +1,8 @@
 //! Exports the `Term` type which is a high-level API for the Grid.
```
chrisduerr, line 1: I'm skipping my review of this file until later. I don't think any of my suggestions should have an effect on this anyway and I don't want to rush it or delay the review until I'm done.

The only thing I can already say for sure is that `DamagedLines` is a terrible name for what it appears to be doing based on what I've read elsewhere.

chrisduerr (changes requested):

alacritty/src/display/mod.rs
```diff
@@ -319,6 +331,13 @@ impl Display {
         }
 
         let hint_state = HintState::new(config.hints.alphabet());
+        let is_damage_supported = window.swap_buffers_with_damage_supported();
+        let debug_damage = config.debug.highlight_damage;
+        let damage_rects = if is_damage_supported || debug_damage {
+            Vec::with_capacity(size_info.screen_lines())
+        } else {
+            Vec::new()
+        };
```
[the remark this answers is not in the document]
kchibisov: Yes, it's purely for upper bound.

alacritty/src/display/mod.rs
```diff
@@ -491,6 +518,46 @@ impl Display {
...
+            // automatically by an iterator.
+            for rect in renderer_damage {
+                self.damage_rects.push(rect)
+            }
+
+            // If we forced the entire screen to be damaged ensure that the next frame will be also
+            // fully damaged just to clean things up.
+            if requires_full_damage {
+                terminal_damage.mark_fully_damaged();
+            }
```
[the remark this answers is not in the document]
kchibisov: Yes, but computing damage is cheap anyways and putting conditionals in functions which update damage will make it slower, since `min, max` are faster than `if false { max }`.

alacritty/src/display/mod.rs
```diff
@@ -491,6 +518,46 @@ impl Display {
...
+                vi_mode_cursor,
+                selection_range,
+                display_offset,
+            );
+
+            // Collect renderer damage information and reset terminal damage. The reset is done
+            // automatically by an iterator.
+            for rect in renderer_damage {
+                self.damage_rects.push(rect)
+            }
```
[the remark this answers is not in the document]
kchibisov: We could reset damage manually and don't do anything in iterator which is doing a reset.

alacritty/src/display/mod.rs
```diff
@@ -636,7 +705,17 @@ impl Display {
         #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
         self.request_frame(&self.window);
 
-        self.window.swap_buffers();
+        if self.is_damage_supported {
+            // We need to redraw entire screen when debugging, so passing empty Vec.
+            let rects: &[DamageRect] = if self.debug_damage { &[] } else { &self.damage_rects };
```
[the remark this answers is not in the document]
kchibisov: They are the same. swap_buffers just passes entire screen damaged under the hood.

alacritty/src/display/mod.rs
```diff
@@ -636,7 +705,17 @@ impl Display {
...
+            self.window.swap_buffers_with_damage(rects);
+        } else {
+            self.window.swap_buffers();
+        }
+
+        if self.collect_damage() {
+            self.damage_rects.clear();
+        }
```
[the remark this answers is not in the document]
kchibisov: I think we can just clear it when starting collecting damaged rects.

alacritty/src/display/mod.rs
```diff
@@ -781,6 +863,11 @@ impl Display {
     ) {
         let text = format!("[{}/{}]", line, total_lines - 1);
         let column = Column(size_info.columns().saturating_sub(text.len()));
+        let point = Point::new(0, column);
+
+        // Damage in a way that it covers most possible length of the text.
+        self.damage_from_point(Point::new(0, point.column - text.len()), text.len() as u32 * 2);
```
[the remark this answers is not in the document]
kchibisov: Because the length is twice the text.len()? I'm overdamaging line indicator in case it grows.

alacritty_terminal/src/selection.rs
kchibisov: Since I'm relying on this behavior when damaging. And from what I can see we also rely that start <= end.

alacritty/src/display/damage.rs
kchibisov: It's all damage information collected from alacritty_terminal.

alacritty/src/display/damage.rs
kchibisov: It's because we're doing look ahead. I can't just do `next()` all the time on terminal_damage, so I need to access it by index, advance ahead and fallback to previous values.

alacritty/src/display/damage.rs
kchibisov: This is the rectangle we'll either yeild `fn next` or its merged result with some other rect. So basically the one we're currently using for merging.

alacritty/src/display/damage.rs
kchibisov: I think I've got an idea how to rewrite it a bit, so it'll make more sense.

alacritty/src/display/damage.rs
kchibisov: Yeah, newly created terminal is always fully damaged.

alacritty_terminal/src/term/mod.rs
kchibisov: Yeah, I have an idea how to do it better. You could just wait and I push updates soon-ish.

alacritty/src/display/damage.rs
chrisduerr: Should update the comment accordingly then.

alacritty/src/display/damage.rs
chrisduerr: There might be a better name then. Will have to look at the iterator method to really understand though probably.

alacritty/src/display/damage.rs
chrisduerr: If it's an iterator, it should be named appropriately. If it's just a data bag for all the damage information, that's different. It shouldn't be both. I feel like it's probably the former though, so a rename might be appropriate?

alacritty_terminal/src/selection.rs
chrisduerr: Hm okay, to what degree do you rely on it? Seems like a `debug_assert!` would be more appropriate? Or does it cause a crash in your damaging logic?

alacritty/src/display/mod.rs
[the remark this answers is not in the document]
chrisduerr: This logic should definitely be documented properly. There's no way anyone would get what exactly the point of this is just by looking at the code.

alacritty/src/display/mod.rs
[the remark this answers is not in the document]
chrisduerr: Can just do `self.damage_supported && !self.debug_damage` then.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Minimum Rust version has been bumped to 1.56.0
 
+### Added
+
+- Track and report damage to compatible compositors (primarily Wayland)
```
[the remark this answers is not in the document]
kchibisov: Well, on X11 with EGL the function `swap_buffers_with_damage` is also present. But let's just state it as Wayland.

alacritty_terminal/src/selection.rs
kchibisov: I'm iterating from start line to end line. I can work around that by manually sorting them, but I'm not sure that I want to do that, since we do order the range that way when creating it.

alacritty/src/display/damage.rs
kchibisov: It actually doesn't damage full width chars properly. It was doing that initially, but then I've removed that part to speed up things. So right now we do that to include wide chars.

kchibisov:
@chrisduerr I've simplified code a bit, so should be better now.

alacritty_terminal/src/term/mod.rs
```diff
@@ -300,6 +474,10 @@ impl<T> Term<T> {
 
         let scroll_region = Line(0)..Line(grid.screen_lines() as i32);
 
+        // Initialize initial damage, so it covers the entire termninal upon launch.
+        let default_cursor_style = config.cursor.style();
+        let damage = TermDamageState::new(num_cols, num_lines, grid.cursor.point);
```
chrisduerr, line 479:
```suggestion
        // Initialize terminal damage, covering the entire terminal upon launch.
        let default_cursor_style = config.cursor.style();
        let damage = TermDamageState::new(num_cols, num_lines, grid.cursor.point);
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -320,9 +498,65 @@ impl<T> Term<T> {
...
+            damage,
         }
     }
 
+    #[must_use]
+    pub fn damage(&mut self, selection: Option<SelectionRange>) -> TermDamage<'_> {
+        // Ensure that we damage entire terminal in insert mode.
+        if self.mode.contains(TermMode::INSERT) {
+            self.is_fully_damaged = true;
+        }
```
chrisduerr, line 511: This doesn't make any sense to me, why would we check `INSERT` when queried for damage?

Applications might enable/disable `INSERT` multiple times per frame, so no matter what the current state of the flag is we can't use it to make any decision on what has happened since the last frame.

Maybe you want to add this code in the mode switch logic?

alacritty_terminal/src/term/mod.rs
```diff
@@ -320,9 +498,65 @@ impl<T> Term<T> {
...
+
+        // Add information about old cursor position and new one if they are not the same, so we
+        // cover everything that was produced by `Term::input`.
+        if self.damage.old_cursor != self.grid.cursor.point {
+            self.damage.damage_point(self.damage.old_cursor);
+        }
```
chrisduerr, line 517: What if the line changed? What if the cursor jumped back and forth since the last frame?

I don't see what benefit this provides beyond `damage_cursor` already done right below.

alacritty_terminal/src/term/mod.rs
```diff
@@ -944,8 +1214,10 @@ impl<T: EventListener> Handler for Term<T> {
             (Line(0), self.bottommost_line())
         };
 
-        self.grid.cursor.point.line = max(min(line + y_offset, max_y), Line(0));
-        self.grid.cursor.point.column = min(col, self.last_column());
+        self.damage_cursor();
+        self.grid.cursor.point.line = cmp::max(cmp::min(line + y_offset, max_y), Line(0));
+        self.grid.cursor.point.column = cmp::min(col, self.last_column());
+        self.damage_cursor();
```
chrisduerr, line 1220: Damaging cursor twice here seems inefficient? Shouldn't the `damage_cursor` function automatically compute a delta to its last position?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1070,6 +1352,7 @@ impl<T: EventListener> Handler for Term<T> {
             self.wrapline();
             return;
         }
+        self.damage_cursor();
```
chrisduerr, line 1355: Again not a fan of the double damage.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1295,26 +1594,18 @@ impl<T: EventListener> Handler for Term<T> {
...
-                    *cell = bg.into();
-                }
-            },
+        let (left, right) = match mode {
+            ansi::LineClearMode::Right => (point.column, Column(self.columns())),
+            ansi::LineClearMode::Left => (Column(0), point.column + 1),
+            ansi::LineClearMode::All => (Column(0), Column(self.columns())),
+        };
+
+        self.damage.damage_line(point.line.0 as usize, left.0, right.0 - 1);
```
chrisduerr, line 1605:
```suggestion
        self.damage.damage_line(point.line.0 as usize, left.0, right.0 - 1);
        
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -1468,6 +1761,7 @@ impl<T: EventListener> Handler for Term<T> {
                 self.tabs.clear_all();
             },
         }
+        self.is_fully_damaged = true;
```
chrisduerr, line 1764: Why does this require a redraw? Shouldn't the content be unaffected by `clear_tabs`?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1632,7 +1929,10 @@ impl<T: EventListener> Handler for Term<T> {
             ansi::Mode::LineFeedNewLine => self.mode.remove(TermMode::LINE_FEED_NEW_LINE),
             ansi::Mode::Origin => self.mode.remove(TermMode::ORIGIN),
             ansi::Mode::ColumnMode => self.deccolm(),
-            ansi::Mode::Insert => self.mode.remove(TermMode::INSERT),
+            ansi::Mode::Insert => {
+                self.mode.remove(TermMode::INSERT);
+                self.is_fully_damaged = true;
+            },
```
chrisduerr, line 1935: Ah, this seems to relate to my comment in `fn damage()`.

I suppose your strategy is to reduce the number of times `fully_damaged` has to be set by only calling it in remove and checking `insert` in `fn damage()` basically? Since one or the other always has to be hit.

chrisduerr (changes requested):

alacritty/src/display/mod.rs
```diff
@@ -457,18 +478,61 @@ impl Display {
         self.window.resize(physical);
         self.renderer.resize(&self.size_info);
 
+        if self.collect_damage() {
+            let fill_rect = DamageRect { x: 0, y: 0, width: 0, height: 0 };
+            self.damage_rects.resize(self.size_info.screen_lines(), fill_rect);
+        }
```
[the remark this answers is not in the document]
kchibisov: I think just reassigning vector is simpler, since resize requires a default arg and `reserve` doesn't shrink...

alacritty/src/display/mod.rs
```diff
@@ -457,18 +478,61 @@ impl Display {
...
+        self.damage_hints(terminal);
+        let size_info: SizeInfo<u32> = self.size_info.into();
+        match terminal.damage(selection_range) {
+            TermDamage::Full => {
+                let screen_rect =
+                    DamageRect { x: 0, y: 0, width: size_info.width(), height: size_info.height() };
+                self.damage_rects.push(screen_rect);
+            },
+            TermDamage::Partical(damaged_lines) => {
+                let damaged_rects = RenderDamageIterator::new(damaged_lines, self.size_info.into());
```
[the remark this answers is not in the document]
kchibisov: it's used in both branches, so I don't see a reason to move it.

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,108 @@
...
+        let mut rect = self.rect_for_line(line);
+        let mut advance = 0;
+
+        let mut ahead_rect = rect;
+        let mut ahead_area = rect_area(&ahead_rect);
+
+        // The amount we advanced by from the last picked rect.
+        let mut advanced_ahead = 0;
+
+        for line in self.damaged_lines.clone() {
```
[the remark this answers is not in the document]
kchibisov: There's no need to and it won't work since we're mutating the iterator and not touching the original until we ensure that we should. Cloning iterator is cheap anyway.

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            ahead_area += next_rect_area;
+            ahead_rect = merged_rect;
+        }
+
+        // Advance iterator in case we've processed more damaged lines when we were looking ahead.
+        for _ in 0..advance {
+            let _ = self.damaged_lines.next();
+        }
```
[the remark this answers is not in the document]
kchibisov: I think we can just straight update iterator.

alacritty_terminal/src/term/mod.rs
```diff
@@ -153,57 +199,173 @@ impl SizeInfo {
...
+    /// Right bound where damage starts.
+    pub right: usize,
+}
+
+impl LineDamageInformation {
     #[inline]
-    pub fn padding_x(&self) -> f32 {
-        self.padding_x
+    pub fn undamaged(line: usize, num_cols: usize) -> Self {
+        Self { line, left: num_cols, right: 0 }
```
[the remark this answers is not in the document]
kchibisov: That's for performance reasons since we don't need to check anything `min` and `max` will work automatically that way.

alacritty_terminal/src/term/mod.rs
```diff
@@ -153,57 +199,173 @@ impl SizeInfo {
...
+}
+
+impl TermDamageState {
+    fn new(num_cols: usize, num_lines: usize, old_cursor: Point) -> Self {
+        let lines = (0..num_lines)
+            .into_iter()
+            .map(|line| LineDamageInformation::undamaged(line, num_cols))
+            .collect();
+
+        Self { num_cols, lines, old_cursor, old_selection: None }
```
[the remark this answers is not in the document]
kchibisov: Could use `Point::default` I guess.

alacritty_terminal/src/term/mod.rs
```diff
@@ -269,6 +431,12 @@ pub struct Term<T> {
     /// Information about cell dimensions.
     cell_width: usize,
     cell_height: usize,
+
+    /// Hint whether terminal should be damaged entirely regardless of the actual damage changes.
+    is_fully_damaged: bool,
+
+    /// Information about damaged cells in terminal.
+    damage: TermDamageState,
```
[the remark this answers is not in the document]
kchibisov: I guess, I'm not sure why I've moved it though.

alacritty_terminal/src/term/mod.rs
kchibisov: There's a code in switch logic already. I just want to always damage if mode `INSERT` is present, mode change only helps with one frame, however if you won't toggle it on the next frame it won't help.

alacritty_terminal/src/term/mod.rs
kchibisov: The point is to cover `Term::input`, if you won't do that you won't damage content written by it properly. As an example look at the test using `input`, that's exactly why damaging old cursor is required, since you expand from it up to new one.

alacritty_terminal/src/term/mod.rs
kchibisov: You change the line here, so you should damage old and new one. We could try moving `damage.last_cursor` around though.

alacritty_terminal/src/term/mod.rs
kchibisov: Yeah, you could look at `INSERT` test, since it's covered in it.

alacritty_terminal/src/term/mod.rs
kchibisov: Moving last_cursor around gets complicated when you start moving it back and forth. I think I'd rather do simpler, but in a way that you can't really fail with collecting damage.

kchibisov:
@chrisduerr I think I've resolved everything, should be fine now.

chrisduerr (changes requested):
This is only a review of the display/mod.rs file, I'll try to get to the other two tomorrow.

alacritty_terminal/src/term/mod.rs
```diff
@@ -300,6 +470,10 @@ impl<T> Term<T> {
 
         let scroll_region = Line(0)..Line(grid.screen_lines() as i32);
 
+        // Initialize terminal damage, covering the entire terminal upon launch.
+        let default_cursor_style = config.cursor.style();
```
chrisduerr, line 474:
```suggestion
        let scroll_region = Line(0)..Line(grid.screen_lines() as i32);
        let default_cursor_style = config.cursor.style();

        // Initialize terminal damage, covering the entire terminal upon launch.
```

Bit confusing to have the comment right above a completely unrelated line.

alacritty_terminal/src/term/mod.rs
```diff
@@ -320,9 +494,64 @@ impl<T> Term<T> {
             selection: None,
             cell_width: size.cell_width as usize,
             cell_height: size.cell_height as usize,
+            damage,
         }
     }
 
+    #[must_use]
+    pub fn damage(&mut self, selection: Option<SelectionRange>) -> TermDamage<'_> {
+        // Ensure that we damage entire terminal in insert mode.
```
chrisduerr, line 503:
```suggestion
        // Ensure the entire terminal is damaged after entering insert mode.
        // Leaving is handled in the ansi mode handler.
```

Documenting everything right here should make it easy for people to understand this logic directly without having to go digging.

alacritty_terminal/src/term/mod.rs
```diff
@@ -320,9 +494,64 @@ impl<T> Term<T> {
...
+        }
+
+        // Always damage current cursor.
+        self.damage_cursor();
+        self.damage.last_cursor = self.grid.cursor.point;
+
+        // Damage Vi cursor if it's present.
+        if self.mode.contains(TermMode::VI) {
+            self.damage_vi_cursor();
+        }
```
chrisduerr, line 521: I forgot, so please remind me: Why does the cursor always have to be damaged even if it did not move? Is this in case only the cursor cell was rewritten?

alacritty_terminal/src/term/mod.rs
```diff
@@ -320,9 +494,64 @@ impl<T> Term<T> {
...
+
+        if self.damage.last_selection != selection {
+            let display_offset = self.grid().display_offset();
+            for selection in self.damage.last_selection.into_iter().chain(selection) {
+                self.damage.damage_selection(selection, display_offset, self.columns());
+            }
+        }
+
+        self.damage.last_selection = selection;
```
chrisduerr, line 530:
```suggestion
        if self.damage.last_selection != selection {
            let display_offset = self.grid().display_offset();
            for selection in self.damage.last_selection.into_iter().chain(selection) {
                self.damage.damage_selection(selection, display_offset, self.columns());
            }
        }
        self.damage.last_selection = selection;
```

Just to ensure a little bit of a stronger bond between these related statements.

alacritty_terminal/src/term/mod.rs
```diff
@@ -320,9 +494,64 @@ impl<T> Term<T> {
...
+
+        if self.damage.is_fully_damaged {
+            TermDamage::Full
+        } else {
+            TermDamage::Partial(TermDamageIterator::new(&self.damage.lines))
+        }
```
chrisduerr, line 536: Do you think it might make sense to check a bit earlier for `is_fully_damaged`? Seems unfortunate to do anything after that's already obvious.

alacritty_terminal/src/term/mod.rs
```diff
@@ -819,8 +1072,17 @@ impl<T> Term<T> {
         cursor_cell.flags = flags;
     }
 
-    pub fn colors(&self) -> &Colors {
-        &self.colors
+    #[inline]
+    fn damage_cursor(&mut self) {
```
chrisduerr, line 1076: > Moving last_cursor around gets complicated when you start moving it back and forth. I think I'd rather do simpler, but in a way that you can't really fail with collecting damage.

Why do we need to move this around? Would it not be possible to reuse the `last_cursor` on `self.damage` as-is?

Would it not be sufficient to always damage the delta to the last cursor position when checking it? Something like this:

```
let last_cursor = mem::replace(&mut self.damage.last_cursor, self.grid.cursor.point);
self.damage.damage_point(last_cursor);
```

Since we always damage the current cursor state at least, this should damage the cursor at every stage of the process, right? Or are there scenarios where the cursor is moved, but damage is skipped? If so that would surely create issues wrt `input`.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2429,6 +2722,288 @@ mod tests {
         assert_eq!(term.grid.cursor.point, Point::new(Line(4), Column(0)));
     }
 
+    #[test]
```
chrisduerr, line 2725: Haven't really reviewed the tests, maybe one comment I could leave about them is that they do look kinda long and complicated. So that's somewhat discouraging. Usually prefer things to be small and easy to understand in case someone needs to change things in the future.

chrisduerr (changes requested):

alacritty/src/display/mod.rs
```diff
@@ -457,18 +479,67 @@ impl Display {
...
+        if requires_full_damage {
+            terminal.mark_fully_damaged();
+        }
+
+        self.damage_highlighted_hints(terminal);
+        let size_info: SizeInfo<u32> = self.size_info.into();
+        match terminal.damage(selection_range) {
+            TermDamage::Full => {
+                let screen_rect =
+                    DamageRect { x: 0, y: 0, width: size_info.width(), height: size_info.height() };
```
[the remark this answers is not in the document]
kchibisov: The problem is that it's glutin type and it doesn't have `new`. I'd used it otherwise...

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,108 @@
+use std::cmp;
+
+use glutin::Rect;
+
+use alacritty_terminal::term::{LineDamageBounds, SizeInfo, TermDamageIterator};
+
+/// Minimum area ratio of efficient damaged rects and resulted merged rect.
+const MIN_DAMAGED_AREA_RATIO: f32 = 0.4;
```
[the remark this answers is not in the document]
kchibisov: I get confused by user suggestion. Assume that you have first rect of area `X` and second of area `Y`. The efficient damage area is area that was actually changed, however we damage more than that. So comparing efficient damage area of both rects to a new one is a ratio of efficient_damage/damage.

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,108 @@
...
+            // achieve the best accuracy, since the best accuracy is achieved if you don't merge
+            // anything. So we try to merge rects if the area ratio is higher than
+            // `NIM_DAMAGED_AREA_RATIO`, if it's lower that that ratio, but higher than
+            // `MIN_DAMAGED_AREA_RATIO_ERR` we try to merge more rectangles until we reach
+            // `MIN_DAMAGED_AREA_RATIO`. If we haven't been able to reach `MIN_DAMAGED_AREA_RATIO`
+            // we abort merging and reporting what've collected so far.
+            if merged_ratio >= MIN_DAMAGED_AREA_RATIO {
+                total_damage = next_merged_rect;
+                // Update iterator position.
+                self.damaged_lines = damaged_lines.clone();
```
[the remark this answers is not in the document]
kchibisov: Do you want to have `advance` approach back? Cloning iterator is pretty cheap, since you just clone offset into underlying collection and pointer to that collection. I thought that it's easier than having more extra variables for no reason.

alacritty/src/display/damage.rs
```diff
@@ -0,0 +1,108 @@
...
+            } else if merged_ratio < MIN_DAMAGED_AREA_RATIO_ERR {
+                break;
```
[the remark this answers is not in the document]
kchibisov: The problem is that without this logic if you have small rectangle right above huge one you issue 2 rectangles to damage, however if you have big rectangle above small one you'll issue just one.

>Have you actually checked how often this makes a significant difference in number of damage rectangles reported, I'd be curious about some application where this significantly changes things.

For me it's vim, since popup window for completions opens under cursor for me, so I damage both of them with the same rect. Without this logic it issues 2 rectangles.

alacritty_terminal/src/term/mod.rs
```diff
@@ -153,57 +199,172 @@ impl SizeInfo {
...
+    fn new(line_damage: &'a [LineDamageBounds]) -> Self {
+        Self { line_damage: line_damage.iter() }
+    }
+}
+
+impl<'a> Iterator for TermDamageIterator<'a> {
+    type Item = LineDamageBounds;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.line_damage.find(|line| line.is_damaged()).copied()
```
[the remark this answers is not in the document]
kchibisov: You can't do that, since find borrows.

alacritty_terminal/src/term/mod.rs
```diff
@@ -153,57 +199,172 @@ impl SizeInfo {
...
+    /// Old selection range.
+    last_selection: Option<SelectionRange>,
+}
+
+impl TermDamageState {
+    fn new(num_cols: usize, num_lines: usize) -> Self {
+        let lines =
+            (0..num_lines).map(|line| LineDamageBounds::undamaged(line, num_cols)).collect();
+
+        Self { is_fully_damaged: true, lines, last_cursor: Point::default(), last_selection: None }
```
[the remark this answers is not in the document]
kchibisov: This won't fit into one line, so I'd rather keep it this way? But if you still prefer `Default::default()` I can change to that.

alacritty_terminal/src/term/mod.rs
```diff
@@ -153,57 +199,172 @@ impl SizeInfo {
...
     }
 
     #[inline]
-    fn screen_lines(&self) -> usize {
-        self.screen_lines
+    fn damage_point(&mut self, point: Point) {
+        let line = point.line.0 as usize;
+        let column = point.column.0 as usize;
+        self.damage_line(line, column, column);
     }
```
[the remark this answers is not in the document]
kchibisov: No, we don't damage offscreen points, however I do ensure that point is in viewport when calling to this function. The terminal point just can't have line and col as negative, since it won't make any sense. However Vi cursor could, and it converts into viewport coordinates before calling it.

Yeah, the comment should help.

alacritty_terminal/src/term/mod.rs
kchibisov: Yes, also it simplifies blinking and shape change.

alacritty_terminal/src/term/mod.rs
kchibisov: They just cover **all** functions which are damaging terminal in one way or another. So those tests are made to hit coverage, and not be short. But they are really simple if you look into them. They basically do `damage -> check result -> reset` and repeat until we cover all functions affected.

alacritty_terminal/src/term/mod.rs
kchibisov: This won't work. Consider the following. You can't use `last_cursor` since it's not accurate, because it doesn't account for changes from `fn input`. Consider the following:


```rust
// You have no damage at all.
term.goto(1,1); // last_cursor is on 0,0, so damaging it and setting to 1,1.
term.input('a');
term.input('b');
term.input('c');
term.goto(2,0); // Damaging 1,1 and setting to 2,0.
// We failed to damaged line properly...
```

And such case will be in other functions where you change the line. So if we want to use last_cursor position we will have to damage it twice here as well, but It just proves that there's no need to carry last_cursor position around, since doing it simpler is better.

In function where cursor can't move to other lines I'm damaging line, not cursor.

alacritty_terminal/src/term/mod.rs
chrisduerr: Yeah I saw that for some of them. Still somewhat annoying to quickly determine if they're actually doing what they're supposed to.

alacritty_terminal/src/term/mod.rs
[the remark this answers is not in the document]
chrisduerr: I do prefer it, even if it doesn't fit.

alacritty/src/display/damage.rs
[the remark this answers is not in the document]
chrisduerr: I don't really like either of these approaches.

alacritty/src/display/damage.rs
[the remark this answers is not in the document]
chrisduerr: My suggestions were based on the incorrect assumption that this represents growth, not effective area size. So some other wording is likely better.

alacritty_terminal/src/term/mod.rs
```diff
@@ -103,7 +102,54 @@ pub struct SizeInfo {
...
+            cell_height: size_info.cell_height as u32,
+            padding_x: size_info.padding_x as u32,
+            padding_y: size_info.padding_y as u32,
+            screen_lines: size_info.screen_lines,
+            columns: size_info.screen_lines,
+        }
+    }
+}
+
+impl<T: Clone + Copy + PartialEq> SizeInfo<T> {
```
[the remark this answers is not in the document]
kchibisov: I just haven't seen the comment.

kchibisov:
I've fixed an issue I've found on irc, should be good now.

kchibisov:
@perfbot

alacritty_terminal/src/term/mod.rs
```diff
@@ -2429,6 +2736,288 @@ mod tests {
...
+        let mut damaged_lines = match term.damage(selection_range) {
+            TermDamage::Full => panic!("Expected partial damage, however got Full"),
+            TermDamage::Partial(damaged_lines) => damaged_lines,
+        };
+        let line = line as usize;
+        // Skip information about cursor on zero line.
+        damaged_lines.next();
+        assert_eq!(damaged_lines.next(), Some(LineDamageBounds { line, left, right }));
+        assert_eq!(damaged_lines.next(), None);
+        term.reset_damage();
```
chrisduerr, line 2792: Still not a fan of these tests. Newlines are seemingly inserted at random, comments are sprinkled throughout being very specific about arbitrary information while not giving any detail on the bigger picture.

The only way I can figure out what you're trying to do is by reading the **entire** testcode and then having to **guess** that maybe that's what you want to do. That's not really maintainable.

Like why do you first test inserting `damage` and then again test `!`? Shouldn't that basically test the same codepaths? I can only assume that you're trying to test the case of a single character separately, but it's basically a 50/50 chance you just duplicated the test randomly. No clue.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2429,6 +2736,288 @@ mod tests {
...
+            TermDamage::Partial(damaged_lines) => damaged_lines,
+        };
+        // Skip information about cursor on zero line.
+        damaged_lines.next();
+        assert_eq!(damaged_lines.next(), Some(LineDamageBounds { line, left, right }));
+        assert_eq!(damaged_lines.next(), None);
+    }
+
+    #[test]
+    fn damage_movements() {
```
chrisduerr, line 2822:
```suggestion
    fn damage_cursor_movements() {
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -2429,6 +2736,288 @@ mod tests {
...
+        let size = SizeInfo::new(10.0, 10.0, 1.0, 1.0, 0.0, 0.0, false);
+        let mut term = Term::new(&Config::default(), size, ());
+        let num_cols = term.columns();
+        // New termianal is always damaged. Reset it since we test when damage is partial.
```
chrisduerr, line 2826: This comment is used in multiple tests but not really phrased particularly well.

I think something like this is what you're trying to say?

```
// Reset terminal for partial damage tests since it's initialized as fully damaged.
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -2429,6 +2736,288 @@ mod tests {
...
+        term.reset_damage();
+
+        term.goto(Line(1), Column(1));
+
+        // NOTE Since [`Term::damage`] always includes cursor we'll be comparing damaged lines
+        // directly, since we're testing that all functions bellow damage cursor in right places.
```
chrisduerr, line 2832: The first part of the sentence I can still follow. After the second "since" I have no idea what you're trying to say.

alacritty_terminal/src/term/mod.rs
chrisduerr, line 2822: Sidenote: This test lends itself much nicer to having a bunch of stuff inside it. It's trivial to see what each small block does.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2429,6 +2736,288 @@ mod tests {
...
+    fn damage_vi_movements() {
+        let size = SizeInfo::new(10.0, 10.0, 1.0, 1.0, 0.0, 0.0, false);
+        let mut term = Term::new(&Config::default(), size, ());
+        let num_cols = term.columns();
+        // New termianal is always damaged. Reset it since we test when damage is partial.
+        term.reset_damage();
+
+        term.toggle_vi_mode();
+        assert!(term.damage.is_fully_damaged);
+        term.reset_damage();
```
chrisduerr, line 2927: Why does vi mode toggle fully damage the terminal?

alacritty_terminal/src/term/mod.rs
```diff
@@ -2429,6 +2736,288 @@ mod tests {
...
+        assert_eq!(term.damage.lines[3], LineDamageBounds { line: 3, left: 5, right: 6 });
+        assert_eq!(term.damage.lines[4], LineDamageBounds { line: 4, left: 5, right: 6 });
+        assert_eq!(term.damage.lines[5], LineDamageBounds { line: 5, left: 5, right: 5 });
+
+        // Ensure that we haven't damaged entire terminal during the test.
+        assert!(!term.damage.is_fully_damaged);
+    }
+
+    #[test]
+    fn should_be_fully_damaged() {
```
chrisduerr, line 2949:
```suggestion
    fn full_damage() {
```

Some tests in here test that things are **not** fully damaged. I think `full_damage` thus fits a little better as a name.

chrisduerr, line 2949: This test is fine too.

chrisduerr (changes requested):
Looks like things are taking shape.

alacritty_terminal/src/term/mod.rs
```diff
@@ -311,7 +489,7 @@ impl<T> Term<T> {
             colors: color::Colors::default(),
             semantic_escape_chars: config.selection.semantic_escape_chars.to_owned(),
             cursor_style: None,
-            default_cursor_style: config.cursor.style(),
+            default_cursor_style,
```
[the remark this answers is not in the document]
kchibisov: Because I've used cursor style in `TermDamageState` before, so I can revert now.

alacritty_terminal/src/term/mod.rs
kchibisov: To cleanup things that you had in Vi mode. Like line indicator, etc.

chrisduerr: Line indicator should just be a single line at worst though, right?

If your argument is that it's not worth it to be more specific that's fine too. But do we **need** to damage the entire screen?

kchibisov: Yeah, I think we can just leave damaging this indicator entirely to alacritty when it leaves Vi mode.

kchibisov: Fixed.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2429,6 +2733,277 @@ mod tests {
...
+        let size = SizeInfo::new(10.0, 10.0, 1.0, 1.0, 0.0, 0.0, false);
+        let mut term = Term::new(&Config::default(), size, ());
+        let num_cols = term.columns();
+        // Reset terminal for partial damage tests since it's initialized as fully damaged.
+        term.reset_damage();
+
+        term.goto(Line(1), Column(1));
+
+        // NOTE We want to test what damage information is being produced by each function
+        // individually and the entire damage that we could get from `[Term::damage]`.
```
chrisduerr, line 2823: What is the purpose of this comment? How does it relate to the code it is annotating?

You're telling me you want to do something, but I see no relation between that and what is commented here? You talk about damage being produced by each function but call no function that produces damage. You talk about damage you get from `Term::damage` but read the damage from `term.damage` directly. It doesn't make sense?

alacritty_terminal/src/term/mod.rs
```diff
@@ -2429,6 +2733,277 @@ mod tests {
...
     }
 
+    #[test]
+    fn damage_public_usage() {
+        let size = SizeInfo::new(10.0, 10.0, 1.0, 1.0, 0.0, 0.0, false);
+        let mut term = Term::new(&Config::default(), size, ());
+        // Reset terminal for partial damage tests since it's initialized as fully damaged.
+        term.reset_damage();
+
+        // Test that we damage input form [`Term::input`].
```
chrisduerr, line 2743:
```suggestion
        // Test that we damage input form [`Term::input`].
        
```

This comment isn't directly related to the code in annotates, but instead applies to everything that follows. In that case I find it helpful to add a newline between the comment and the actual content.

This also applies to the rest of the function and comments in it.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2429,6 +2733,277 @@ mod tests {
...
+            Selection::new(SelectionType::Block, Point::new(Line(line), Column(3)), Side::Left);
+        selection.update(Point::new(Line(line), Column(5)), Side::Left);
+        let selection_range = selection.to_range(&term);
+
+        let mut damaged_lines = match term.damage(selection_range) {
+            TermDamage::Full => panic!("Expected partial damage, however got Full"),
+            TermDamage::Partial(damaged_lines) => damaged_lines,
+        };
+        let line = line as usize;
+        // Skip information about cursor on zero line, since we just check for selection.
```
chrisduerr, line 2775: I'm not sure what you mean by this. Is this suggestion accurate?

```suggestion
        // Skip cursor damage information, since we're just testing selection.
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -2429,6 +2733,277 @@ mod tests {
...
+        damaged_lines.next();
+        assert_eq!(damaged_lines.next(), Some(LineDamageBounds { line, left, right }));
+        assert_eq!(damaged_lines.next(), None);
+        term.reset_damage();
+
+        // Check that existing selection gets damaged when it gets removed.
```
chrisduerr, line 2781:
```suggestion
        // Check that existing selection gets damaged when it is removed.
```

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
```diff
@@ -2819,8 +2823,11 @@ mod tests {
 
         term.goto(Line(1), Column(1));
 
-        // NOTE We want to test what damage information is being produced by each function
-        // individually and the entire damage that we could get from `[Term::damage]`.
+        // NOTE While we can use `[Term::damage]` to access terminal damage information, in the
+        // following tests we will be accessing `term.damage.lines` directly to avoid adding extra
+        // damage information(like cursor and Vi cursor) which we're not testing in the following
+        // cases.
```
chrisduerr, line 2829:
```suggestion
        // NOTE While we can use `[Term::damage]` to access terminal damage information, in the
        // following tests we will be accessing `term.damage.lines` directly to avoid adding extra
        // damage information (like cursor and Vi cursor), which we're not testing.
```

Already has a `in the following tests` in the beginning, best not to repeat ourselves. Also tiny parenthesis mistake.

chrisduerr (changes requested):

chrisduerr (approved):

PR #5790, opened 2022-01-18
Add inline input method support

### LINKED ISSUE #1613: Support inline "input method" input
Which operating system does the issue occur on? Gentoo Linux

On alacritty

2018-10-01_mon_23 53 52

On termite

2018-10-01_mon_23 58 54

Out-of-band input method input is disorienting.

kchibisov:
This commit adds support for inline IME handling. It also makes search
bar draw proper Underline cursor instead of using '_' character.

Fixes #1613.


kchibisov:
@chrisduerr Would like to hear your input on this one, since alacritty part is done. However maybe you suggest something wrt the winit API we're using, so we can tune it.

alacritty/src/display/mod.rs
```diff
@@ -193,11 +194,33 @@ pub struct Display {
...
+#[derive(Debug, Default)]
+pub struct IMEInput {
+    /// The text being preedit.
+    preedit: String,
+
+    /// Byte offset for cursor start in text.
+    cursor_start: Option<usize>,
+
+    /// Byte offset for cursor end in text.
+    _cursor_end: Option<usize>,
```
kchibisov, line 215: This is basically to have a highlight in preedit string while you're at it. We could probably do that by drawing hollow block cursor from cursor_start up to cursor_end, what do you think? For now I don't handle end.

alacritty/src/event.rs
```diff
@@ -1106,6 +1108,26 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                     WindowEvent::KeyboardInput { input, is_synthetic: false, .. } => {
                         self.key_input(input);
                     },
+                    WindowEvent::IME(ime) => match ime {
+                        // TODO decide what to do with it, but in general we should stop update of
+                        // IME input methods.
+                        IME::Enabled => {},
```
kchibisov, line 1114: Will it make sense to check whether ime is present based on this enabled info? We can't check based on preedit string since you don't have it right after inserting text with commit.

crocket:
Does this support fcitx5?

kchibisov:
@crocket it's using text-input-v3, so everything that is using wayland protocol for IME and not dbus should work. So if fcitx is using input-method-v2 it should work. I don't have ability to test fcitx, but it doesn't matter, since they talk to compositor and compositor back to us.

For example the following https://github.com/tadeokondrak/anthywl and https://github.com/emersion/wlhangul works just fine.

rano-oss:
It works great on wayland in my tests, but it fails to start on x11 with this error: 
"error: XMisc("IME Context creation failed")"

pickfire:
I tested it out on x11 with fcitx5, didn't seemed to work, no inline preedit and preedit position is incorrect. Do I have to do something special for the settings? Below I have screenshots with fcitx5 mozc and pinyin.

Screenshot_20220118_180751
Screenshot_20220118_180811

For a comparison, this is how konsole looks like.

Screenshot_20220118_180928

alacritty/src/event.rs
pickfire: What do you mean by ime is present? Like when user toggled ime? IIRC there isn't an event for that until user enter some text.

komi1230:
@rano-oss This issue will be fixed in https://github.com/rust-windowing/winit/pull/2147. Currently, the implementation for X11 is still not done. We are working on that, so wait for a second.

alacritty/src/display/mod.rs
chrisduerr: What is start/end here? In which scenario can the cursor be more than a fullwidth character (I'm not familiar with IME)?

alacritty/src/display/mod.rs
```diff
@@ -570,6 +668,46 @@ impl Display {
             rects.push(visual_bell_rect);
         }
 
+        // Handle search and IME drawing/positioning.
+        match search_state.regex() {
```
chrisduerr, line 672: Is this for IME support in the searchbar?

alacritty/src/event.rs
```diff
@@ -1106,6 +1108,26 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
...
+                        IME::Commit(text) => {
+                            self.ctx.display.ime_input = None;
+                            text.chars().for_each(|ch| self.received_char(ch));
+                        },
+                        // This text is being composed, thus should be shown as inline input to
+                        // the user.
+                        IME::Preedit(text, cursor_start, cursor_end) => {
+                            *self.ctx.dirty = true;
+                            self.ctx.display.ime_input =
+                                Some(IMEInput::new(text, cursor_start, cursor_end));
```
chrisduerr, line 1125: Small comment on the winit API and something I've noticed designing some enum APIs myself: If the parameter here would be a struct instead of a tuple, we could just use the struct 1:1 for our own implementation. But since it's a tuple we need to take that tuple and convert it to our own type.

Generally this makes abstractions a lot more difficult, you can never just take `IME::Preedit(xxx)` and use the entire `xxx` nicely for future processing because then you'll always have to deal with tuple indexing. Using structs instead will automatically allow you to refer to the fields by their name, making the API a lot nicer to use in my experience.

alacritty/src/display/mod.rs
rano-oss: The Cursor covers the whole sentence, often one writes whole sentences:
E.G:
Start: 0, end: 0 = |您好
Start: 0, end: 6* = |您好 (or the whole text is marked or something, I haven't seen it yet)
Start: 6, end: 6* = 您好|

*(these are usually 3 bytes each)

kchibisov: Yeah, it could be treated as selection from start to end in preedit, I also haven't seen something like this, all IMEs I've used just report the same start and end.

alacritty/src/event.rs
kchibisov: On Wayland it comes when you enable input method. So you can clearly skip all input method handling until you've get this. For example you don't have to do anything about IME popup position, which we're currently doing in display, since there's no IME at all.

alacritty/src/display/mod.rs
```diff
@@ -715,8 +830,8 @@ impl Display {
             }
         }
 
-        // Add cursor to show whitespace.
-        formatted_regex.push('_');
+        // Add spacer for cursor to show whitespace.
+        formatted_regex.push(' ');
```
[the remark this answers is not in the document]
kchibisov: Yes, that's why I do it that was and draw cursor above it instead of just drawing it one cell righter.

alacritty/src/event.rs
```diff
@@ -1106,6 +1108,26 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
...
+                        },
+                        IME::Disabled => {
+                            self.ctx.display.ime_input = None;
+                        },
```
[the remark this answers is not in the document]
kchibisov: It's ime enabled/disabled thing. So you could safely stop all ime you have inflight and stop its handling. Will likely require adding a bool on display if we want to make use of it.

alacritty/src/event.rs
kchibisov: Yeah, I also thought that it'd be nice. Will refactor to send struct instead of tuple.

alacritty/src/display/mod.rs
kchibisov: Yes, and we always had it.

alacritty/src/event.rs
kchibisov: >What do you mean by ime is present? Like when user toggled ime? IIRC there isn't an event for that until user enter some text, maybe I am wrong.

No, On Wayland/macOS you known when you're using IME. So if you don't have something like fcitx running you won't ever get Enabled.

rano-oss: Enable comes from the client, and is used to signal that you have entered a text input area. If there is an input method present it will be activated and produce text input, else it will just produce plain old keyboard focus. 
So for @kchibisov's question, if I understood correctly, you can assume that input will come from either a keyboard or from an input method and never both. The compositor should handle which one you receive. You just signal that there is a possibility for text input here

rano-oss:
> @rano-oss This issue will be fixed in [rust-windowing/winit#2147](https://github.com/rust-windowing/winit/pull/2147). Currently, the implementation for X11 is still not done. We are working on that, so wait for a second.

I will, I hope I didn't sound impatient, I was just pointing it out. Maybe this is behind a feature flag and I should have kept my mouth shut, if so I apologize.

alacritty/src/event.rs
rano-oss: @kchibisov I do not see an enable event here? ref: [https://smithay.github.io/wayland-rs/wayland_protocols/unstable/text_input/v3/client/zwp_text_input_v3/enum.Event.html](https://smithay.github.io/wayland-rs/wayland_protocols/unstable/text_input/v3/client/zwp_text_input_v3/enum.Event.html)
Did I miss something?

kchibisov: @rano-oss you won't get Enter until you have IME running, at least that's how it works for me in sway.

rano-oss: Ah that is true, just got confused by the naming. Then I agree with your statement, it makes sense to check that IME is present based on this.
Also thanks, that clarified something for my compositor module.

pickfire: I think this might not apply to X11, so not sure.

kchibisov: Yeah, but those platforms in winit will just send this event on window creation and emit disabled on window death, so it doesn't matter.

chrisduerr: > For example you don't have to do anything about IME popup position, which we're currently doing in display, since there's no IME at all.

Probably easier for us to just update IME position than to check whether we have to or not.

kchibisov: >Probably easier for us to just update IME position than to check whether we have to or not.

I also feel like it, so I don't want to do anything with enabled event.

kchibisov:
For anyone who is using ime on daily basis, do you need a beam(or your config) cursor in the end of preedit of IME input? I think vte/gtk doesn't draw it, but some terminals on macOS do draw it.

rano-oss:
> For anyone who is using ime on daily basis, do you need a beam(or your config) cursor in the end of preedit of IME input? I think vte/gtk doesn't draw it, but some terminals on macOS do draw it.

Yes, the cursor at the end is important to give an indication of where one is when writing. It is typical to move that cursor backwards to E.G change a character somewhere in a sentence before committing the sentence.

kchibisov:
>It is typical to move that cursor backwards to E.G change a character somewhere in a sentence before committing the sentence.

Yeah, I guess I've just haven't seen it in GTK, they only draw cursor when you start moving it backwards, but not when it's in the end. I'm also not quite sure why you'd need a cursor in the end tbqh, but sure, could add it I guess.

rano-oss:
> > It is typical to move that cursor backwards to E.G change a character somewhere in a sentence before committing the sentence.
> 
> Yeah, I guess I've just haven't seen it in GTK, they only draw cursor when you start moving it backwards, but not when it's in the end. I'm also not quite sure why you'd need a cursor in the end tbqh, but sure, could add it I guess.

I think as long as you just accommodate for a cursor at the position of "cursor_end"? And then just let the IME handle cursor placement? And I guess if "cursor_begin != cursor_end" you make the text between marked?

kchibisov:
The X11 IME should work, at least in a way xlib allows it to work. The preedit window at the bottom of the window is xlib feature...

pickfire:
> The X11 IME should work, at least in a way xlib allows it to work. The preedit window at the bottom of the window is xlib feature...

I think that might be the default. But for it to appear under the cursor like other applications we need to set the preedit style to either over-the-spot or on-the-spot (IIRC on-the-spot requires more stuff) and update the preedit position every time the cursor position changed.

kchibisov:
@pickfire no, xlib basically doesn't support it with on-the-spot. You'd need xcb or patch xlib directly, there's no way around it. I'd try sending patch to xlib, since right now when preedit callbacks are used the cursor spot setting requests are discarded by xlib for some 'legacy' reasons.

pickfire:
Ah, xlib vs xcb. Yeah, I think maybe need xcb. Is it a dependency that we shouldn't take in?

kchibisov:
@pickfire yes, because winit is using xlib, so unless we rewrite winit with xcb, we can't use really use it. But we can try patching xlib to allow that, since winit has spot setting in place.

CHANGELOG.md
```diff
@@ -19,11 +19,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support for OpenGL ES 2.0
 - Escape sequence to set underline color (`CSI 58 : 2 : Ps : Ps : Ps m`/`CSI 58 : 5 : Ps m`)
 - Escape sequence to reset underline color (`CSI 59 m`)
+- Support for inline input method
```
pickfire, line 22: Do we want to mention that the caveat is that on x11 only off-the-spot preedit area (no over-the-spot or on-the-spot) is supported due to library limitations (xcb not available)?

kchibisov: I'm not entirely sure. It's kind of common and I've seen other clients with such issue.

@chrisduerr what do you think?

chrisduerr: I have absolutely no clue what these different IME modes are, but chances are if we just say "support XYZ" and Toby's version of XYZ doesn't work he'll come crying on the issue tracker. So we should set the expectations so people are aware of which setups should and shouldn't work now?

kchibisov: Something like `IME is now stuck at the bottom of the window due to libX11 limitations` in CHANGED?

pickfire: `IME is only available at the bottom of the window (off-the-spot preedit area) due to libX11 limitations for x11`

kchibisov: I don't know, I kind of like the one I have in CHANGELOG right now, since it's more about changed.

CHANGELOG.md
```diff
@@ -19,11 +19,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 
 ### Changed
 
 - The `--help` output was reworked with a new colorful syntax
 - OSC 52 is now disabled on unfocused windows
+- Search bar is now respecting cursor thickness
+- On X11 IME popup window is stuck at the bottom of the window due to libX11 limitations
```
chrisduerr, line 29:
```suggestion
- On X11 the IME popup window is stuck at the bottom of the window due to libX11 limitations
```

I think like this is good.

CHANGELOG.md
pickfire: It's also about changed, but at least it needs to mention the unexpected, better than let users dig in to figure out why it didn't work as they expected.

kchibisov:
Windows IME inline bits were also added.

kchibisov:
Closing in favor of #6079.

PR #5795, opened 2022-01-21
Use builtin font to draw powerline symbols

kchibisov:
In addition to box drawing it was decided to also draw powerline
symbols, since those are quite common and rather simple to draw with
present box drawing infra.

### DIFF
--- a/alacritty/src/renderer/text/builtin_font.rs
+++ b/alacritty/src/renderer/text/builtin_font.rs
@@ -1,5 +1,5 @@
 //! Hand-rolled drawing of unicode [box drawing](http://www.unicode.org/charts/PDF/U2500.pdf)
-//! and [block elements](https://www.unicode.org/charts/PDF/U2580.pdf).
+//! and [block elements](https://www.unicode.org/charts/PDF/U2580.pdf), and also powerline symbols.
 
 use std::{cmp, mem, ops};
 
@@ -25,6 +25,8 @@ pub fn builtin_glyph(
     let mut glyph = match character {
         // Box drawing characters and block elements.
         '\u{2500}'..='\u{259f}' => box_drawing(character, metrics, offset),
+        // Powerline symbols: '','','',''
+        '\u{e0b0}'..='\u{e0b3}' => powerline_drawing(character, metrics, offset),
         _ => return None,
     };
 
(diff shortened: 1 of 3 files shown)


kchibisov:
Updated to be against v0.11.0.

chrisduerr (changes requested):
The glyphs don't appear to be perfect:

tmp

Especially with powerline symbols, the addition really only makes sense when things are absolutely perfect. If the slope doesn't start **immediately**, that's likely not something that's gonna be noticed, but if the slope starts too late like this or it's uneven that's gonna be really jarring (I was immediately able to notice this for example).

solonovamax:
> The glyphs don't appear to be perfect:

I'm not sure how you arrived at this.

I've built & tested the branch, and from taking a screenshot and then counting pixels in GIMP, it seems to be accurate to me.

See:
image
image
image
image
image

chrisduerr:
@solonovamax Your images show the same issue I've demonstrated in my screenshot. You just need to take a closer look.

solonovamax:
Could you point out where it is specifically?
I believe that may help with resolving the issue, as if I can't notice it, OP might not be able to notice it. (Unless you've discussed this with them off-platform)

chrisduerr:
> I believe that may help with resolving the issue, as if I can't notice it, OP might not be able to notice it. (Unless you've discussed this with them off-platform)

I believe he's capable of speaking for himself.

solonovamax:
> > I believe that may help with resolving the issue, as if I can't notice it, OP might not be able to notice it. (Unless you've discussed this with them off-platform)
> 
> I believe he's capable of speaking for himself.

I don't want to imply that they can't speak for themselves, I just want to see this merged sooner as this is a feature I would like.

And I apologize if anything I said came off as ride; that was not my intent.

Though, I'll stop commenting unless I have fixes/etc, as this is now becomming off-topic for this thread.

kchibisov:
Does latest commit changes anything @chrisduerr . I really don't see issues with filled rectangles.

picture

chrisduerr:
Since you asked on IRC, this still doesn't look right to me:

tmp

Pretty sure it's because of the anti-aliasing, but there's a really sharp "edge" for the first pixel step while the following pixel steps look far softer, making it look wonky.

Pytness:
Here is what I have came up to.

I use a Canvas twice the size of the glyph, draw the lines fill if necessary, and scale down the canvas to get some sort of antialiasing.
I have tried to line drawing algorithms, Xiaolin's (the one @kchibisov uses)  and Bresenham's.


Examples of other "rendering" methods (made with gimp)
image



Zoomed in:
image

sharkusk:
Would it make sense to support a fallback/override font, where specific code point ranges can be set and used for characters in these ranges?

For example, a user could install a single Nerd Font as the override, then set the specific ranges to support whatever Gyphs they desire (https://github.com/ryanoasis/nerd-fonts/wiki/Glyph-Sets-and-Code-Points).  This would eliminate the need to use patched fonts or handle special symbol rendering in alacritty.

Config would look something like this:

```
font:
  # Special Symbol Override Font
  override:
    # Font family 
    family: Hasklug Nerd Font

   # Nerd Font Code Points
    ranges: 23fb-23fe, 2665, 26a1, 2b58, e000-e00a, e0a0-e0a2, e0a3, e0b0-e0b3, e0b4-e0c8, etc... 
```

chrisduerr:
> Would it make sense to support a fallback/override font, where specific code point ranges can be set and used for characters in these ranges?

No, that would be complete nonsense. This patch is precisely because we **do not** want to use fonts for these glyphs. You can already setup fallback fonts through fontconfig just fine.

alacritty/src/renderer/text/builtin_font.rs
```diff
@@ -495,6 +497,125 @@ fn box_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> Raster
...
+
+    let mut canvas = Canvas::new(width, height);
+
+    let x_end = width as f32 - 1.;
+    let y_end = height as f32 - 1.;
+    let y_center = y_end / 2.;
+
+    // Offset to achieve 90 degree angle in the tip when we can do so just by moving 1px down/up
+    // on the opposite side. When we can't do so, just cut on the tip side.
+    let y_offset = if (height + 1).saturating_sub(2 * width) / 2 == 1 { 1 } else { 0 };
```
chrisduerr, line 514: This is optional, it might be simpler and easier to maintain if we just always do this.

chrisduerr (changes requested):

alacritty/src/renderer/text/builtin_font.rs
```diff
@@ -495,6 +497,125 @@ fn box_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> Raster
...
+    let width = (metrics.average_advance as i32 + offset.x as i32) as usize;
+    // Use one eight of the cell width, since this is used as a step size for block elements.
+    let stroke_size = cmp::max((width as f32 / 8.).round() as usize, 1) as f32;
+
+    let mut canvas = Canvas::new(width, height);
+
+    let y_center = (height - 1) as f32 / 2.;
+    // Start with offset `1` and draw until the intersection of the 2 f(x) = x + 1 and
+    // g(x) = H - x + 1. The intersection happens at 2 * f(x) = H - 2 or f(x) = H/2,
+    // which is `y_center`.
```
chrisduerr, line 511: Nobody will ever understand what this comment means in a week or two lol.

chrisduerr (approved):

alacritty/src/renderer/text/builtin_font.rs
kchibisov: I've tuned it a bit.

Issue #3309, opened 2020-02-07
can't find config file on windows!


catalin560:
`%APPDATA%\alacritty\alacritty.yml` is not a valid path...

kchibisov:
You should put it there yourself. We're not creating a config and other random permanent files for you, it's expected.

catalin560:
where can I find the docs for configuring the terminal?

kchibisov:
everything is mentioned there https://github.com/alacritty/alacritty#configuration

catalin560:
is there a way to integrate it withing vscode and not open a separate instance?

kchibisov:
likely no, vscode uses its own terminal IIRC. So you can't integrate anything into it, I guess. But I'm not an expert here. However if you can open any terminal in vscode or let's say Firefox(random browser), you can likely open alacritty too.

Just to make it clear, we're not cmd.exe,pwsh.exe, etc. We're actually a terminal emulator, not a shell, like cmd.exe and friends.

catalin560:
shame... cmder seems to have a working solution for this https://github.com/cmderdev/cmder/wiki/Seamless-VS-Code-Integration

also I still can't get alacritty to see my config... what is `%APPDATA%` supposed to stand for?
chocolatey installed it in c/tools/alacritty so I assume that the config should be in c/ProgramData/alacritty ??? if so it doesn't work for me

kchibisov:
IIRC you can run alacritty in ConEmu, so in theory if cmder is a conemu fork, you can run alacritty in it, but I'm not an expert here.

The location should be `C:\Users\YOUR_USERNAME\AppData\Roaming\alacritty\alacritty.yml` IIRC.

chrisduerr:
%APPDATA% is how Windows specifies environment variables. So it's not necessarily the location @kchibisov pointed out, though that is most likely. You should check the %APPDATA% environment yourself.

desmap:
> The location should be C:\Users\YOUR_USERNAME\AppData\Roaming\alacritty\alacritty.yml IIRC.

This should be in the README, currently there's a wrong path for Windows user. No deal-breaker but this wastes your users' time.

kchibisov:
>This should be in the README, currently there's a wrong path for Windows user. No deal-breaker but this wastes your users' time.

The path in readme isn't wrong, it's correct, since it's using environment variable, and should auto expand. The path in readme is exactly the same we're using internally, when we're searching for config.

johnnovak:
[...]

PR #5804, opened 2022-01-23
Update YAML config location in README for Windows

dane:
The README states `%APPDATA%\alacritty\alacritty.yml`, but [a comment in issue 3309](https://github.com/alacritty/alacritty/issues/3309#issuecomment-583313866) suggests including the `Roaming` directory in the path. I tried both and `%APPDATA%\Roaming\alacritty\alacritty.yml` worked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ following locations:
 
 On Windows, the config file should be located at:
 
-`%APPDATA%\alacritty\alacritty.yml`
+`%APPDATA%\Roaming\alacritty\alacritty.yml`
 
 ## Contributing
 


chrisduerr:
%APPDATA% is an environment variable, which points to AppData\Roaming. It's correct the way it is.

hardyrafael17:
I wonder how many users have left because of this? which could have a very simple solution.  (Explain it well in the read me). or make the program search for the file in the Home directory, which won't change as a environment variable.

hardyrafael17:
Ohhh Wow, How nice, (I ended up using it because I like the idea and because I like to force myself to solve the problems I face). 

How does something not being intuitive and well explained help this community and this project?  I'm not a Windows user, So it's fine if YOU @chrisduerr think that the most users should know where that variable points to. (it's so counter intuitive, because you search in the internet and in most places I've found it says ~/AppData, but some times it points to /AppData/Roaming. 

(I'm not event trying to understand how they work on windows. (just not interested), I just don't see how your comment helps anyone.

chrisduerr:
I'm not a Windows user either. In fact I don't own any machine with Windows installed on it. There's nothing counter-intuitive about environment variables and you shouldn't have to "search in the internet" for them. If you need someone to hold your hand for something so utterly basic, I'd recommend looking somewhere else.

hardyrafael17:
I'm sorry, you seem to be blocked by some knowledge barrier or something and don't seem to relate to others. I did search the internet because what was intuitive for me wasn't working. Apparently I was not the only one or we would't be having this conversation. (I use linux, Arch Linux by the way), I use neovim as my text editor where I work, I work as as a software developer), I'm not saying I'm smart, I'm just saying I'm trying.

If what you're saying is that people with my level or below my level should look somewhere else, that's fine, I take that. BUT I STILL DON'T SEE HOW THAT HELPS THE COMMUNITY AND THIS PROJECT. 

I see you're a contributor in this project and I thank you for that because I'm liking the application but I still don't see the point in your answer. "Like, come on man! you can't figure it out, try something more basic!" it's a better answer, no answer at all.  Just my opinion.

Not replying anymore, because I also don't see the point in that. It's you way. Millions of people in this world with different views. Probably I'm just plain wrong.

yamanawabi:
@hardyrafael17 Not sure what all the debate is about in here. 

If you open Start->Run and type in %AppData% you will see it will open an Explorer window to C:\Users\youruser\AppData\Roaming. This is a commonly used environment variable.

I don't totally agree with the somewhat apathetic responses from the Alacritty members, but as @chrisduerr mentioned, there isn't any need for a change here.

PR #5811, opened 2022-01-24
Update INSTALL.md

eau-u4f:
OpenBSD now has an official package for alacritty on OpenBSD.


chrisduerr (changes requested):
The INSTALL.md documents building from source. People are expected to know how to install packages through their package manager.

PR #5821, opened 2022-01-26
Fix builtin font lines rendering with font.offset

kchibisov:
This commit takes into account font offset when generating builtin font.

--

I'm not sure about the thickness of those lines and the config option to disable them.  As for a thickness I'd pick something like `width / 4` and as for a config option I'm not sure, but we could add, I guess. What do you think @chrisduerr ?

### DIFF
--- a/alacritty/src/renderer/builtin_font.rs
+++ b/alacritty/src/renderer/builtin_font.rs
@@ -5,6 +5,8 @@ use std::{cmp, mem};
 
 use crossfont::{BitmapBuffer, Metrics, RasterizedGlyph};
 
+use crate::config::ui_config::Delta;
+
 // Colors which are used for filling shade variants.
 const COLOR_FILL_ALPHA_STEP_1: Pixel = Pixel { _r: 192, _g: 192, _b: 192 };
 const COLOR_FILL_ALPHA_STEP_2: Pixel = Pixel { _r: 128, _g: 128, _b: 128 };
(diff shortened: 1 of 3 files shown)


chrisduerr:
> As for a thickness I'd pick something like width / 4

Honestly not sure, for me the thickness looks perfectly fine the way it is. I'd have to actually check a system where that's not the case.

> as for a config option I'm not sure, but we could add, I guess

We certainly could and maybe should. But I'm not sure there's any scenario where our built-in font cannot do something a custom font can? Maybe we should encourage people to improve the built-in font instead.

kchibisov:
>Honestly not sure, for me the thickness looks perfectly fine the way it is. I'd have to actually check a system where that's not the case.

Well, if the font has very thin underlines it's likely won't look good. The thing is that it's based on font metric and not on cell width/height.

With fonts I tend to use the lines look fine though, but I'd assume that some fonts just use thin underlines.

chrisduerr (changes requested):
Currently the built-in font is affected by glyph offsets, but I don't think it should be?
tmp

kchibisov:
I'm not entirely sure. But I guess we can do that for consistency.

alacritty/src/renderer/builtin_font.rs
```diff
@@ -20,12 +20,20 @@ pub fn builtin_glyph(
...
-        _ => None,
-    }
+        '\u{2500}'..='\u{259f}' => box_drawing(character, metrics, offset),
+        _ => return None,
+    };
+
+    // Since we want to ignore `glyph_offset` for builtin font apply it to compensate its addition
+    // when loading glyphs in the renderer.
+    glyph.left -= glyph_offset.x as i32;
+    glyph.top -= glyph_offset.y as i32;
```
chrisduerr, line 34: That's pretty hacky, but I guess it's the easiest way?

alacritty/src/renderer/builtin_font.rs
```diff
@@ -20,12 +20,20 @@ pub fn builtin_glyph(
...
         // Box drawing characters and block elements.
-        '\u{2500}'..='\u{259f}' => Some(box_drawing(character, metrics, offset)),
-        _ => None,
-    }
+        '\u{2500}'..='\u{259f}' => box_drawing(character, metrics, offset),
+        _ => return None,
+    };
+
+    // Since we want to ignore `glyph_offset` for builtin font apply it to compensate its addition
+    // when loading glyphs in the renderer.
```
chrisduerr, line 32:
```suggestion
    // Since we want to ignore `glyph_offset` for the built-in font, subtract it to compensate its addition when loading glyphs in the renderer.
```

chrisduerr (changes requested):

alacritty/src/renderer/builtin_font.rs
kchibisov: Yeah, but that doesn't change glyph loading. Otherwise you have to pass this delta and whether it's builtin or not around, and gets very ugly pretty soon.

chrisduerr (approved):

Issue #5809, opened 2022-01-24
Custom font for box drawing characters


h0adp0re:
Hello, the latest version includes and uses a font for box drawing characters. Is it or will it be possible to use a custom font for those? Or maybe just a boolean switch to keep using whatever it was before, I assume `font.normal`? Please?

The problem I have with the new solution is that the characters are way too thin and look out of place in a terminal environment; vertical lines have gaps in between them; the curves are asymmetrical; the curves don't exactly line up with the vertical lines. None of those issues existed before `0.10.0`.

Seeing as you will also be bundling your own powerline icons in the future, I hope an option will exist to dismiss any built-in fonts, they interfere with what I have already spent time on configuring.

kchibisov:
Are you using `font.offset.y`? I think it'll have gaps only in that case.

h0adp0re:
Yes, I am.

kchibisov:
Yeah, I think buildin font should account for line interval, since normal fonts just draw outside of bounding box here. As for thickness it's the same as underline thickness, so it seems like underlines are also out of place for you.

As for now, not using `font.offset` should make everything contiguous.

h0adp0re:
The font I've selected has all characters under control, and they have a pleasant width.

Underlines are used differently, visually – they're always paired with text. I would then amend my statement: thin strokes look bad to me when in isolation, e.g. as a border of a popup window. It's just too sharp and not easy on the eyes at all.

Can I ask what will the fix be? If you're suggesting removing `font.offset` then why have the option at all when it breaks every box drawing from `0.10.0` onwards? I sincerely hope we can somehow "go back" to the way it was.

deathlyfrantic:
I agree with this issue. For me, the boxes with the custom font are contiguous and look "fine", i.e. no broken transitions, but it's way thinner than the font I use and very obviously out-of-place.

deathlyfrantic:
FWIW in the original issue about this, @chrisduerr pointed out there should be an option to disable this behavior.
https://github.com/alacritty/alacritty/issues/5485#issuecomment-917410712

gkapfham:
Hello! Thanks for all of your hard work on alacritty, I appreciate it and have been using this terminal for years! I've also noticed that for version `alacritty 0.10.0 (8a26dee0)` of alacritty installed on Arch Linux that separating lines (i.e., those in Tmux or in pop-up menus in Neovim) are now noticeably thinner and, in fact, hard for me to see.

[...]

PR #5822, opened 2022-01-26
Add option to control built-in box drawing chars

deathlyfrantic:
Adds a `font.use_built_in_font_for_box_drawing` config option to control whether the built-in font is used for drawing box characters. See #5809 for reference.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -178,6 +178,13 @@
   # it is recommended to set `use_thin_strokes` to `false`.
   #use_thin_strokes: true
 
+  # Use built-in font for box drawing characters.
+  #
+  # If `true`, Alacritty will use a custom built-in font for box drawing
+  # characters (Unicode points 2500 - 259f).
+  #
+  #use_built_in_font_for_box_drawing: false
+
 # If `true`, bold text is drawn using the bright color variants.
 #draw_bold_text_with_bright_colors: false
 
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -351,8 +351,14 @@ impl Display {
             info!("Initializing glyph cache...");
             let init_start = Instant::now();
 
-            let cache =
-                renderer.with_loader(|mut api| GlyphCache::new(rasterizer, &font, &mut api))?;
+            let cache = renderer.with_loader(|mut api| {
+                GlyphCache::new(
+                    rasterizer,
+                    &font,
+                    &mut api,
+                    config.font.use_built_in_font_for_box_drawing,
+                )
+            })?;
 
             let stop = init_start.elapsed();
             let stop_f = stop.as_secs() as f64 + f64::from(stop.subsec_nanos()) / 1_000_000_000f64;
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -137,6 +137,9 @@ pub struct GlyphCache {
 
     /// Font metrics.
     metrics: crossfont::Metrics,
+
+    /// Whether to use the built-in font for box drawing characters.
+    use_built_in_font_for_box_drawing: bool,
 }
 
 impl GlyphCache {
@@ -144,6 +147,7 @@ impl GlyphCache {
         mut rasterizer: Rasterizer,
         font: &Font,
         loader: &mut L,
+        use_built_in_font_for_box_drawing: bool,
     ) -> Result<GlyphCache, crossfont::Error>
     where
         L: LoadGlyph,
@@ -167,6 +171,7 @@ impl GlyphCache {
             bold_italic_key: bold_italic,
             glyph_offset: font.glyph_offset,
             metrics,
+            use_built_in_font_for_box_drawing,
         };
 
         cache.load_common_glyphs(loader);
@@ -268,8 +273,12 @@ impl GlyphCache {
 
         // Rasterize the glyph using the built-in font for special characters or the user's font
         // for everything else.
-        let rasterized = builtin_font::builtin_glyph(glyph_key.character, &self.metrics)
-            .map_or_else(|| self.rasterizer.get_glyph(glyph_key), Ok);
+        let rasterized = if self.use_built_in_font_for_box_drawing {
+            builtin_font::builtin_glyph(glyph_key.character, &self.metrics)
+                .map_or_else(|| self.rasterizer.get_glyph(glyph_key), Ok)
+        } else {
+            self.rasterizer.get_glyph(glyph_key)
+        };
 
         let glyph = match rasterized {
             Ok(rasterized) => self.load_glyph(loader, rasterized),
--- a/alacritty/src/config/font.rs
+++ b/alacritty/src/config/font.rs
@@ -38,6 +38,9 @@ pub struct Font {
 
     /// Font size in points.
     size: Size,
+
+    /// Whether to use the built-in font for box drawing characters.
+    pub use_built_in_font_for_box_drawing: bool,
 }
 
 impl Font {


deathlyfrantic:
I haven't contributed to Alacritty before so I'm sure I've missed things - suggestions welcome. Also please give me a better/shorter name for this option. 😄

kchibisov (changes requested):
I'm not entirely sure how we should approach it yet, since I think the majority of the terminals don't allow you to opt-out from builtin font. While I do think that adding such an option is a good idea. I'd rather try to improve builtin font rendering rather than trying to just feature gate it.

What particular issue are you having with builtin font? So we may work on improving it in that regard.

Though, if option will be added font should be enabled by default, since that how the majority terminal emulators are doing so.

The thing about builtin font thickness is that it's using your font metrics, however it could be changed to use cell_width when doing thickness calculations, something like `width / 4` instead of underline thickness.

deathlyfrantic:
Both wezTerm and iTerm2 allow users to opt out (and both do a much better job of matching font thickness). Terminal.app doesn't have custom box glyphs at all. I don't have any way to check non-macOS terminals. In the screenshots below, "with box glyphs" means the terminal is using its custom drawing code, "without box glyphs" means it is using my specified font's box glyphs:

Screen Shot 2022-01-27 at 8 21 39 AM Screen Shot 2022-01-27 at 8 22 45 AM
Screen Shot 2022-01-27 at 8 20 39 AM Screen Shot 2022-01-27 at 8 20 03 AM
Screen Shot 2022-01-27 at 8 24 43 AM Screen Shot 2022-01-27 at 8 25 20 AM

In all of the above cases I still much prefer the font's box glyphs versus the custom ones, but the difference isn't as extreme in wezTerm and iTerm2.

The Alacritty boxes are so thin that if the contrast between foreground and background is not very high, I can barely see the lines. This is my window separator in Neovim in Alacritty with the box glyphs first, and without (bolder line) second:

Screen Shot 2022-01-27 at 8 35 30 AM Screen Shot 2022-01-27 at 8 35 45 AM

kchibisov:
@deathlyfrantic we're using metrics from your underline's in font, so I'd assume they are also very thin for you? Anyway, Would you mind changing the `stroke_size` in `box_drawing.rs` from `metrics.underline` to `width / 4` and also a multiplier for thick stroke from 3 to 2 and retest? I think it should be better that way.

deathlyfrantic:
With those changes it's much better in my opinion, but it is now thicker than the glyphs of the font.
Screen Shot 2022-01-27 at 12 17 47 PM Screen Shot 2022-01-27 at 12 17 23 PM

kchibisov:
@deathlyfrantic would you mind sharing the patch you've applied to achieve that? It doesn't look like width/4 to me.

kchibisov:
Yeah, that's about right, try `ceil(width / 8)` then.

deathlyfrantic:
```diff
diff --git a/alacritty/src/renderer/builtin_font.rs b/alacritty/src/renderer/builtin_font.rs
index 17e85799..a66e05b5 100644
--- a/alacritty/src/renderer/builtin_font.rs
+++ b/alacritty/src/renderer/builtin_font.rs
@@ -25,8 +25,8 @@ pub fn builtin_glyph(character: char, metrics: &Metrics) -> Option<RasterizedGly
 fn box_drawing(character: char, metrics: &Metrics) -> RasterizedGlyph {
     let height = metrics.line_height as usize;
     let width = metrics.average_advance as usize;
-    let stroke_size = cmp::max(metrics.underline_thickness as usize, 1);
-    let heavy_stroke_size = stroke_size * 3;
+    let stroke_size = cmp::max((width as f32 / 8.0).ceil() as usize, 1);
+    let heavy_stroke_size = stroke_size * 2;
     // Certain symbols require larger canvas than the cell itself, since for proper contiguous
     // lines they require drawing on neighbour cells. So treat them specially early on and handle
     // 'normal' characters later.
```

Screen Shot 2022-01-27 at 12 58 01 PM Screen Shot 2022-01-27 at 12 58 21 PM

kchibisov:
@deathlyfrantic yeah, I think it's looking better. I'd take a look on it myself and send a patch wrt thickness.

chrisduerr:
@deathlyfrantic So using that 1/8th size, would you have any reason to turn off the built-in font?

alacritty.yml
```diff
@@ -178,6 +178,13 @@
   # it is recommended to set `use_thin_strokes` to `false`.
   #use_thin_strokes: true
 
+  # Use built-in font for box drawing characters.
+  #
+  # If `true`, Alacritty will use a custom built-in font for box drawing
+  # characters (Unicode points 2500 - 259f).
+  #
+  #use_built_in_font_for_box_drawing: false
```
kchibisov, line 186: Could we default it to `true`? I'd also name it `builtin_box_drawing`, since we're in `font` section already. The reason behind true, since most terminals I'm aware of has it to true, and it makes sense.

alacritty/src/display/mod.rs
```diff
@@ -351,8 +351,14 @@ impl Display {
...
             let init_start = Instant::now();
 
-            let cache =
-                renderer.with_loader(|mut api| GlyphCache::new(rasterizer, &font, &mut api))?;
+            let cache = renderer.with_loader(|mut api| {
+                GlyphCache::new(
+                    rasterizer,
+                    &font,
+                    &mut api,
+                    config.font.use_built_in_font_for_box_drawing,
```
kchibisov, line 359: You don't need to pass it here, it's already stored on a `font` argument you're passing.

alacritty/src/renderer/mod.rs
```diff
@@ -268,8 +273,12 @@ impl GlyphCache {
 
         // Rasterize the glyph using the built-in font for special characters or the user's font
         // for everything else.
-        let rasterized = builtin_font::builtin_glyph(glyph_key.character, &self.metrics)
-            .map_or_else(|| self.rasterizer.get_glyph(glyph_key), Ok);
+        let rasterized = if self.use_built_in_font_for_box_drawing {
```
kchibisov, line 276: Could you use `then` on bool? It should look like

```rust
self.builtin_box_drawing.then(|| builtin_font::bultin_gyph(glyph_key.character, &self.metrics)).map_or_else(|| self.rasterizer.get_glyph(glyph_key, Ok);
```

kchibisov (changes requested):
I still think that we should likely add this option, since it's something common and provided by others I'm not that against. Though, the default should be `true`.

deathlyfrantic:
> @deathlyfrantic So using that 1/8th size, would you have any reason to turn off the built-in font?

At that size I could live with it. I still find it perplexing that I specify a font and that specification is completely ignored for a range of characters. (I understand other terminals also do this.)

alacritty.yml
deathlyfrantic: That's fine. Since this is already in a release it would be a breaking change.

Other than changing this to `true` here in the default config, is there a way to specify the default in the code?

kchibisov: You need to modify `Default` impl for `Font`. You could try to search for `impl Default` in config code to look at examples.

j4james:
FYI, one reason why users might want to turn this off, is when their font has a very specific style which wouldn't be matched by the built-in rendering. For example, you get some fonts that are designed to produce the scanline effect of old CRT monitors, and that effect would be ruined if the box characters were replaced. ([GlassTTY](http://sensi.org/~svo/glasstty/) is the kind of thing I had in mind, although that particular font doesn't include box characters anyway, so it's not a great example, but you get the idea).

Anyway, I'm not a regular Alacritty user, so I'm not specifically requesting this option - I just wanted to give an example of why users might want to disable this feature.

deathlyfrantic:
I renamed my branch (not sure how I came up with the name `add-built-in-font` 🤔) and didn't realize it would close this PR. I will open a new one with the requested changes.

PR #5823, opened 2022-01-27
Add option to control built-in box drawing chars

deathlyfrantic:
Replacement for #5822. (Sorry!)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Minimum Rust version has been bumped to 1.56.0
 
+### Added
+
+ - Option `font.builtin_box_drawing` to disable the built-in font for drawing box characters
+
 ### Changed
 
 - The `--help` output was reworked with a new colorful syntax
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -178,6 +178,13 @@
   # it is recommended to set `use_thin_strokes` to `false`.
   #use_thin_strokes: true
 
+  # Use built-in font for box drawing characters.
+  #
+  # If `true`, Alacritty will use a custom built-in font for box drawing
+  # characters (Unicode points 2500 - 259f).
+  #
+  #builtin_box_drawing: true
+
 # If `true`, bold text is drawn using the bright color variants.
 #draw_bold_text_with_bright_colors: false
 
--- a/alacritty/src/config/font.rs
+++ b/alacritty/src/config/font.rs
@@ -14,7 +14,7 @@ use crate::config::ui_config::Delta;
 /// field in this struct. It might be nice in the future to have defaults for
 /// each value independently. Alternatively, maybe erroring when the user
 /// doesn't provide complete config is Ok.
-#[derive(ConfigDeserialize, Default, Debug, Clone, PartialEq, Eq)]
+#[derive(ConfigDeserialize, Debug, Clone, PartialEq, Eq)]
 pub struct Font {
     /// Extra spacing per character.
     pub offset: Delta<i8>,
@@ -38,6 +38,9 @@ pub struct Font {
 
     /// Font size in points.
     size: Size,
+
+    /// Whether to use the built-in font for box drawing characters.
+    pub builtin_box_drawing: bool,
 }
 
 impl Font {
@@ -69,6 +72,22 @@ impl Font {
     /// Get bold italic font description.
     pub fn bold_italic(&self) -> FontDescription {
         self.bold_italic.desc(&self.normal)
+    }
+}
+
+impl Default for Font {
+    fn default() -> Font {
+        Self {
+            builtin_box_drawing: true,
+            use_thin_strokes: Default::default(),
+            glyph_offset: Default::default(),
+            bold_italic: Default::default(),
+            italic: Default::default(),
+            offset: Default::default(),
+            normal: Default::default(),
+            bold: Default::default(),
+            size: Default::default(),
+        }
     }
 }
 
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -137,6 +137,9 @@ pub struct GlyphCache {
 
     /// Font metrics.
     metrics: crossfont::Metrics,
+
+    /// Whether to use the built-in font for box drawing characters.
+    builtin_box_drawing: bool,
 }
 
 impl GlyphCache {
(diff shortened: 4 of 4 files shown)


alacritty/src/renderer/mod.rs
```diff
@@ -268,8 +272,12 @@ impl GlyphCache {
...
         // Rasterize the glyph using the built-in font for special characters or the user's font
         // for everything else.
-        let rasterized = builtin_font::builtin_glyph(glyph_key.character, &self.metrics)
-            .map_or_else(|| self.rasterizer.get_glyph(glyph_key), Ok);
+        let rasterized = if self.builtin_box_drawing {
+            builtin_font::builtin_glyph(glyph_key.character, &self.metrics)
+                .map_or_else(|| self.rasterizer.get_glyph(glyph_key), Ok)
+        } else {
+            self.rasterizer.get_glyph(glyph_key)
+        };
```
deathlyfrantic, line 280: I couldn't figure out how to make this work with `.then()`.

kchibisov: Try the following:

```diff
diff --git a/alacritty/src/renderer/mod.rs b/alacritty/src/renderer/mod.rs
index a3184d1c..3d7ecc74 100644
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -272,12 +272,11 @@ impl GlyphCache {

         // Rasterize the glyph using the built-in font for special characters or the user's font
         // for everything else.
-        let rasterized = if self.builtin_box_drawing {
-            builtin_font::builtin_glyph(glyph_key.character, &self.metrics)
-                .map_or_else(|| self.rasterizer.get_glyph(glyph_key), Ok)
-        } else {
-            self.rasterizer.get_glyph(glyph_key)
-        };
+        let rasterized = self
+            .builtin_box_drawing
+            .then(|| builtin_font::builtin_glyph(glyph_key.character, &self.metrics))
+            .flatten()
+            .map_or_else(|| self.rasterizer.get_glyph(glyph_key), Ok);

         let glyph = match rasterized {
             Ok(rasterized) => self.load_glyph(loader, rasterized),
```

kchibisov (changes requested):
You should also add a changelog entry.

h0adp0re:
Thank you for this!

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Minimum Rust version has been bumped to 1.56.0
 
+### Added
+
+- Option `font.builtin_box_drawing` to control whether built-in font is used for drawing box characters
```
deathlyfrantic, line 16: I can't tell if there is a max line length for this file. I can wrap this if it's too long.

alacritty/src/config/font.rs
```diff
@@ -72,6 +75,22 @@ impl Font {
...
+            use_thin_strokes: false,
+            normal: Default::default(),
+            bold: Default::default(),
+            italic: Default::default(),
+            bold_italic: Default::default(),
+            size: Default::default(),
+            builtin_box_drawing: true,
+        }
+    }
+}
```
deathlyfrantic, line 92: As an alternative, the option could be named `disable_builtin_box_drawing`, then the default of `false` wouldn't require this new impl.

kchibisov (changes requested):
One more thing I've noticed, is that you should update `builtin_box_drawing` variable in `update_font_size`, so live config reload will work. After that it should be good to go.

kchibisov (approved):
LGTM

CHANGELOG.md
chrisduerr: It is, 100 columns. So this is a bit too wide.

```
 - Option `font.builtin_box_drawing` to disable the built-in font for drawing box characters
```

This is < 100.

alacritty.yml
```diff
@@ -178,6 +178,13 @@
   # it is recommended to set `use_thin_strokes` to `false`.
   #use_thin_strokes: true
 
+  # Use built-in font for box drawing characters.
+  #
+  # If `true`, Alacritty will use a custom built-in font for box drawing
+  # characters (Unicode points 2500 - 259f).
```
chrisduerr, line 184: @kchibisov Gotta remember to update this for your powerline patch.

alacritty/src/config/font.rs
chrisduerr: I'm generally a in favor of naming config options so the default is `false` rather than `true`, but I don't think this is something even remotely consistently enforced so there's no point being stingy about it when this way around is shorter usually.

**However**, either way the structure of this default impl should be changed. The relavant bit (`builtin_box_drawing`), the only reason why this impl exists at all, is at the far bottom. Making it the last thing you see. Other than that there's `use_thin_strokes: false` which makes it look like someone explicitly deviated from `Default::default()` just to specify this but that's clearly not the case.

The impl should look something like this:

```rust
impl Default for Font {
    fn default() -> Font {
        Self {
            builtin_box_drawing: true,
            use_thin_strokes: Default::default(),
            bold_italic: Default::default(),
            glyph_offset: Default::default(),
            italic: Default::default(),
            offset: Default::default(),
            normal: Default::default(),
            bold: Default::default(),
            size: Default::default(),
        }
    }
}
```

This way it is **obvious** what reason we had to custom-impl `Default`, having all the defaults ordered by length makes it super easy to see at a glance that there's no field accidentally slipping in-between that doesn't use the default value.

chrisduerr (changes requested):

alacritty/src/config/font.rs
deathlyfrantic: I changed it to match this exactly but realized the `glyph_offset` line is longer than the `bold_italic` line. Should I flip them?

alacritty/src/config/font.rs
```diff
@@ -72,6 +75,22 @@ impl Font {
     }
 }
 
+impl Default for Font {
+    fn default() -> Font {
+        Self {
+            builtin_box_drawing: true,
+            use_thin_strokes: Default::default(),
+            bold_italic: Default::default(),
+            glyph_offset: Default::default(),
```
chrisduerr, line 84:
```suggestion
            glyph_offset: Default::default(),
            bold_italic: Default::default(),
```

chrisduerr (approved):

chrisduerr:
@kchibisov Do you think it makes sense to just go for the minimal with on/off toggle, or should we immediately go to a thickness option that also functions as an on/off?

kchibisov:
I'm not sure that you'd need thickness tbqh, I'm not even sure that anyone allows that and it'll likely messup box double lines. The problem with thickness is due to fonts using very thin underilnes or that underlines doesn't scale with font size. So I'd probably go for fixed thickness but based on cell width instead of underlie metrics.

chrisduerr:
> I'm not even sure that anyone allows that

Kitty does, that's what sparked my question.

PR #5832, opened 2022-01-29
Use cell width to compute builtin font thickness

### LINKED ISSUE #5826: Built-in box drawing font curved line glyphs
Hello, since the built-in font has been in use for me for a few days now, one thing about the curved line glyphs has started to bother me more and more – they are drawing a rather strange, oblong, asymmetric curve. Is this intentional or will this be patched in future releases?

asymmetric-curves

kchibisov:
While using underline thickness could sound logical to draw other
lines most fonts don't make underlines thick compared to cell bounding
box if you increase font size. So instead we're using cell width to
scale builtin font nicely.

This commit also adjusts arcs drawing and alignment.

Fixes #5826.
Fixes #5821.

--
I think it should be the last one in series of fix builtin font.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - OSC 104 not clearing colors when second parameter is empty
 - Builtin font lines not contiguous when `font.offset` is used
 - `font.glyph_offset` is no longer applied on builtin font
-- Builiin font arcs alignment
+- Buili-in font arcs alignment
 
 ## 0.10.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - The `--help` output was reworked with a new colorful syntax
+- Builtin font thickness is now based on cell width instead of underline thickness
 
 ### Fixed
 
@@ -26,6 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - OSC 104 not clearing colors when second parameter is empty
 - Builtin font lines not contiguous when `font.offset` is used
 - `font.glyph_offset` is no longer applied on builtin font
+- Builiin font arcs alignment
 
 ## 0.10.0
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -26,6 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - OSC 104 not clearing colors when second parameter is empty
 - Builtin font lines not contiguous when `font.offset` is used
 - `font.glyph_offset` is no longer applied on builtin font
+- Builiin font arcs alignment
```
chrisduerr, line 30:
```suggestion
- Built-in font arcs alignment
```

alacritty/src/renderer/builtin_font.rs
```diff
@@ -39,8 +39,11 @@ pub fn builtin_glyph(
 fn box_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> RasterizedGlyph {
     let height = (metrics.line_height as i32 + offset.y as i32) as usize;
     let width = (metrics.average_advance as i32 + offset.x as i32) as usize;
-    let stroke_size = cmp::max(metrics.underline_thickness as usize, 1);
-    let heavy_stroke_size = stroke_size * 3;
+    // We use one eight of the cell width, since this is used as a step of increasing cell
+    // fulfillness in certain glyphs.
```
chrisduerr, line 43:
```suggestion
    // Use one eighth of the cell width, since this is used as a step size for block elements.
```

alacritty/src/renderer/builtin_font.rs
```diff
@@ -716,6 +745,24 @@ impl Canvas {
...
+            let right = match self.buffer[row..row + self.width].iter().rposition(|p| p._r != 0) {
+                Some(right) => row + right,
+                _ => continue,
+            };
+
+            for index in left + 1..right {
+                self.buffer[index] =
+                    self.buffer[index] + self.buffer[index - 1] / 2 + self.buffer[index + 1] / 2;
+            }
+        }
```
chrisduerr, line 765: What exactly is this needed for? If the arc is multiple pixels "thick", is this code used to fill in everything between the left and right boundary?

chrisduerr (changes requested):

alacritty/src/renderer/builtin_font.rs
kchibisov: Yes, for very large args there's a change that you'd have dots in them, since you don't have enough precision for them.

chrisduerr (approved):
This patch doesn't seem to make any difference with my standard font settings, but things are a little bigger when increasing the font size.

Arc drawing also doesn't seem different to me, but nothing is worse.

PR #5833, opened 2022-01-29
Log on which unix platform event loop is run

kchibisov:
This also fixes a bug where the welcome log message wasn't logged.

--
I just feel like it could save time with XWayland and such.


chrisduerr (approved):
Don't see any way this could hurt, always good to have extra information available.

PR #5841, opened 2022-01-31
Apply `window.opacity` on search bar

kchibisov:
--
I'm not entirely sure that it should be done, but it doesn't look bad.


chrisduerr (commented):
I was just gonna say "why not", but I think the only reasons I can come up with are in favor of not applying transparency to it.

If you think about it, transparency is usually just used for people to be able to see through their terminal and look at the stuff behind it, it's mostly more of a background-color opacity than opacity of all elements within the terminal.

This is also why the option to make all colors transparent should be used with caution. And the search bar is more similar to a fixed background color than it is to the default background color of the window. It's a temporary element where the color is significant and serves a purpose.

So I think it would be best to keep the search bar opaque?

kchibisov:
I guess we should keep it opaque.

PR #5842, opened 2022-02-01
Colored underline

kchibisov:
--
Just testing perf of this, since I think it could be done way better than original approach.


kchibisov:
@perfbot

kchibisov:
@perfbot

kchibisov:
I've tried various approaches and it seems like it's always slower. The closest thing is when no dynamic memory is in use. But that it'll increase runtime memory... Though if we could shorten our pointer it'll help.

picture

chrisduerr:
@perfbot (Just testing something)

chrisduerr:
@perfbot (Just testing something)

PR #5847, opened 2022-02-02
Fix builtin font block elements going invisible

kchibisov:


chrisduerr (approved):
I'd assume this fix is to resolve an issue where small fonts would round down to 0 width?

kchibisov:
I've changed rounding here in general.

chrisduerr (approved):

PR #5860, opened 2022-02-06
Add support for setting prompt marks via OSC 133

### LINKED ISSUE #5850: Consider adding OSC 133
This allows some basic shell integration and the protocol seems like supported by some major modern terminals (kitty/iterm2).

Adding it should be relatively easy and shouldn't impact performance at all.

I'm not sure that adding all of them makes sense for us, but ability to just mark prompt should be fine.

The relevant spec could be found https://iterm2.com/documentation-escape-codes.html.

kchibisov:
Adds support for setting prompt marks with OSC 133 ; A ST. While
there're other markers for prompt `FTCS_PROMPT` (`A` parameter) is
likely the most widely used and useful one.

Fixes #5850.


kchibisov:
@perfbot

chrisduerr (commented):
You mentioned on IRC that a `Vec` might be the best option. Have you tried alternative implementations or is this the first attempt?

kchibisov:
>You mentioned on IRC that a Vec might be the best option. Have you tried alternative implementations or is this the first attempt?

I've tried and I don't like it, since it's more code and you have to put it on a grid into scrolling part.

chrisduerr:
> I've tried and I don't like it, since it's more code and you have to put it on a grid into scrolling part.

Have you considered how we would implement this feature if we wanted to add all of the possible variations instead of just `FTCS_PROMPT`? Would you do it the same way and always store it in the first column?

kchibisov:
>Have you considered how we would implement this feature if we wanted to add all of the possible variations instead of just FTCS_PROMPT? Would you do it the same way and always store it in the first column?

I think yes, and those will be offsets from the fist column to other prompt marks. It'll just move to extra storage.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1046,10 +1046,13 @@ impl<T> Term<T> {
         let c = self.grid.cursor.charsets[self.active_charset].map(c);
         let fg = self.grid.cursor.template.fg;
         let bg = self.grid.cursor.template.bg;
-        let flags = self.grid.cursor.template.flags;
+        let mut flags = self.grid.cursor.template.flags;
 
         let mut cursor_cell = self.grid.cursor_cell();
 
+        // Preserve PROMPT_MARK if we had it.
```
chrisduerr, line 1053:
```suggestion
        // Preserve PROMPT_MARK.
```

Also is this necessary? The prompt shouldn't usually be overwritten?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1600,6 +1603,9 @@ impl<T: EventListener> Handler for Term<T> {
         let bg = cursor.template.bg;
         let point = cursor.point;
 
+        // We don't want to clear prompt markers.
+        let prompt_marked = *self.grid[point.line][Column(0)].flags() & Flags::PROMPT_MARK;
+
```
chrisduerr, line 1608: Is this necessary? Who rewrites a prompt without writing the prompt there?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1694,6 +1703,7 @@ impl<T: EventListener> Handler for Term<T> {
         match mode {
             ansi::ClearMode::Above => {
                 let cursor = self.grid.cursor.point;
+                let prompt_mark = self.grid[cursor.line][Column(0)].flags & Flags::PROMPT_MARK;
```
chrisduerr, line 1706: This seems really strange, there's gonna be nothing there anymore but the prompt line will still be set?

alacritty_terminal/src/term/mod.rs
```diff
@@ -1707,18 +1717,21 @@ impl<T: EventListener> Handler for Term<T> {
                     *cell = bg.into();
                 }
 
+                self.grid[cursor.line][Column(0)].flags.insert(prompt_mark);
                 let range = Line(0)..=cursor.line;
                 self.selection = self.selection.take().filter(|s| !s.intersects_range(range));
             },
             ansi::ClearMode::Below => {
                 let cursor = self.grid.cursor.point;
+                let prompt_mark = self.grid[cursor.line][Column(0)].flags & Flags::PROMPT_MARK;
```
chrisduerr, line 1726: Same here, question if we should keep the prompt mark.

alacritty_terminal/src/term/mod.rs
```diff
@@ -2050,6 +2063,21 @@ impl<T: EventListener> Handler for Term<T> {
         }
     }
 
+    #[inline]
+    fn set_prompt_mark(&mut self) {
+        // Don't mark prompts in alt-screen, since we likely not the one responsible for prompt
+        // marks handling anyway.
+        if self.mode.contains(TermMode::ALT_SCREEN) {
+            return;
+        }
```
chrisduerr, line 2072: Why bother? We could just let them set it? It's not gonna have any negative effect, is it?

alacritty_terminal/src/term/mod.rs
```diff
@@ -2050,6 +2063,21 @@ impl<T: EventListener> Handler for Term<T> {
...
+
+        let mark_point = self.line_search_left(self.grid.cursor.point);
+        trace!("Setting shell prompt mark at: line={}, col={}", mark_point.line, mark_point.column);
```
chrisduerr, line 2075: It's always gonna be column 0, right? So no need to log it?

alacritty_terminal/src/term/search.rs
```diff
@@ -434,6 +434,48 @@ impl<T> Term<T> {
...
     }
+
+    /// Find the next shell prompt.
+    pub fn next_prompt(&self, mut point: Point) -> Option<Point> {
+        // If we're at the start of the prompt go to the next.
+        if point.column == Column(0) {
+            point.line += 1
+        } else {
+            point.column = Column(0);
+        }
```
chrisduerr, line 445: Wouldn't it be easier to just take a line as argument? It's not like the vi cursor can be in front of the first column, so you'll always start in the line after anyway?

alacritty_terminal/src/term/search.rs
```diff
@@ -434,6 +434,48 @@ impl<T> Term<T> {
...
+
+        let end = self.screen_lines() as i32;
+        while point.line < end {
```
chrisduerr, line 448: Should use <= bottommost_line. Makes things a little more intuitive.

alacritty/src/input.rs
```diff
@@ -246,6 +246,55 @@ impl<T: EventListener> Execute<T> for Action {
...
+                };
+
+                if terminal.mode().contains(TermMode::VI) {
+                    terminal.vi_goto_point(prompt_point);
+                } else {
+                    let scroll = Scroll::Delta(point.line.0 - prompt_point.line.0);
+                    terminal.scroll_display(scroll);
+                }
+
+                ctx.mark_dirty();
```
chrisduerr, line 272: Might want to make this a function on `Term`? I think that would be more consistent with the rest of our code.

alacritty/src/input.rs
```diff
@@ -246,6 +246,55 @@ impl<T: EventListener> Execute<T> for Action {
...
             Action::SearchBackward => ctx.start_search(Direction::Left),
+            Action::NextPrompt => {
+                let terminal = ctx.terminal_mut();
+
+                let point = if terminal.mode().contains(TermMode::VI) {
+                    terminal.vi_mode_cursor.point
+                } else {
+                    let display_offset = terminal.grid().display_offset() as i32;
+                    // If we're outside of Vi mode use top most visible line.
+                    Point::new(Line(terminal.screen_lines() as i32 - 1 - display_offset), Column(0))
```
chrisduerr, line 257: This is not the topmost line though, is it? With a display offset of 0, the topmost line would be line `0`. With your logic and a viewport with 30 lines it would be `30 - 1 - 0 = 29`. Which is the bottommost line.

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
kchibisov: It does get overwritten due to the way everyone is sending prompt marks. They send prompt mark and then start writing the prompt, leading to overwriting the cell. The documentation and recommendations state that you should send it before writing the prompt.

alacritty_terminal/src/term/mod.rs
kchibisov: That's what zsh doing right after `precmd`. It calls `ClearBelow`. The `bash` and some other shells call `ClearLine::Right`. Though, we can remove the `above` one.

alacritty/src/input.rs
kchibisov: Yeah, the comment is wrong. I need bottom most line.

alacritty_terminal/src/term/search.rs
kchibisov: The point is when Vi cursor isn't on the first column to jump to first column. And only jump to next/prev prompt from first column. I just think it works more intuitive than jumping straight to some other prompt.

alacritty_terminal/src/term/cell.rs
```diff
@@ -30,6 +30,7 @@ bitflags! {
         const ALL_UNDERLINES            = Self::UNDERLINE.bits | Self::DOUBLE_UNDERLINE.bits
                                         | Self::UNDERCURL.bits | Self::DOTTED_UNDERLINE.bits
                                         | Self::DASHED_UNDERLINE.bits;
+        const PROMPT_MARK               = 0b1000_0000_0000_0000;
```
chrisduerr, line 33: Should be above `ALL_UNDERLINES`. The "special" constants are usually best kept at the bottom because that way one can easily see at a glance that all bits are correctly set for the other constants.

alacritty_terminal/src/term/mod.rs
```diff
@@ -501,6 +501,57 @@ impl<T> Term<T> {
...
+    /// Go to next prompt farthest down in history.
+    pub fn goto_next_prompt(&mut self)
+    where
+        T: EventListener,
+    {
+        let point = if self.mode.contains(TermMode::VI) {
+            self.vi_mode_cursor.point
+        } else {
+            let display_offset = self.grid.display_offset() as i32;
+            // If we're outside of Vi mode use bottom most visible line.
```
chrisduerr, line 513:
```suggestion
            // If we're outside of Vi mode, use bottommost visible line.
```

Topmost/bottommost are one word, as weird as it looks.

alacritty_terminal/src/term/mod.rs
```diff
@@ -501,6 +501,57 @@ impl<T> Term<T> {
...
+            Point::new(Line(self.screen_lines() as i32 - 1 - display_offset), Column(0))
```
chrisduerr, line 514: Why are you using the bottommost line? If I have a prompt visible at the top and another prompt in the middle of my screen, I want to jump to the prompt in the middle of my screen. I don't want to jump to the next offscreen prompt.

This is the same as search, the top of the screen is always to be considered the origin of the search.

This also shouldn't be inverted for `goto_previous`. Even there we should still make it based on the topmost line and not switch to the bottommost. We always want our search match aligned at the top of the viewport.

alacritty_terminal/src/term/mod.rs
```diff
@@ -501,6 +501,57 @@ impl<T> Term<T> {
...
+
+        if self.mode().contains(TermMode::VI) {
+            self.vi_goto_point(prompt_point);
+        } else {
+            let scroll = Scroll::Delta(point.line.0 - prompt_point.line.0);
+            self.scroll_display(scroll);
+        }
+    }
+
+    /// Go to previous prompt farthest up in history.
```
chrisduerr, line 530: We don't go to the previous prompt farthest up in history, that would mean we skip over as many prompts as possible. We go to the previous prompt in history, so the opposite of farthest. Could say closest/nearest up in history but I feel like that's unnecessary.

alacritty_terminal/src/term/mod.rs
```diff
@@ -501,6 +501,57 @@ impl<T> Term<T> {
...
+    pub fn goto_previous_prompt(&mut self)
+    where
+        T: EventListener,
+    {
+        let point = if self.mode.contains(TermMode::VI) {
+            self.vi_mode_cursor.point
+        } else {
+            // If we're outside of Vi mode use topmost visible line.
```
chrisduerr, line 538:
```suggestion
            // If we're outside of Vi mode, use topmost visible line.
```

alacritty_terminal/src/term/mod.rs
chrisduerr: Really not a big fan of this aspect of the escape. Either way though this comment should be fixed.

alacritty_terminal/src/term/search.rs
```diff
@@ -434,6 +434,46 @@ impl<T> Term<T> {
...
     }
+
+    /// Find the next shell prompt.
+    pub fn next_prompt(&self, mut point: Point) -> Option<Point> {
+        // If we're at the start of the prompt go to the next.
+        if point.column == Column(0) {
+            point.line += 1
+        } else {
+            point.column = Column(0);
+        }
```
chrisduerr, line 445:
```suggestion
        point.column = Column(0);
```

A bit shorter like this and probably compiles down to the same thing anyway?

chrisduerr (changes requested):
Really my only major gripe left with this is the whole reset logic and I'm not sure there's a great solution for this. It's somewhat of an issue with the escape sequence itself and I'm not confident that just deviating from what everyone else does is a great idea.

kchibisov:
Given that 586f982 (Allow mode-exclusive bindings in any mode) got merged, I'm no longer interested in this escape sequence. Using uniq symbol in a prompt is a good enough solution for me.

PR #5862, opened 2022-02-07
Fix alacritty_terminal not emitting damage on color change

kchibisov:


alacritty_terminal/src/term/color.rs
```diff
@@ -12,6 +12,9 @@ use crate::ansi::NamedColor;
 /// Number of terminal colors.
 pub const COUNT: usize = 269;
 
+/// Index of the cursor color.
+pub const CURSOR_COLOR_INDEX: usize = 258;
```
chrisduerr, line 16: You shouldn't use a constant. You should use `NamedColor::Cursor as usize`. Otherwise this will explode if we change the order of the colors.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1645,6 +1651,12 @@ impl<T: EventListener> Handler for Term<T> {
     #[inline]
     fn reset_color(&mut self, index: usize) {
         trace!("Resetting color[{}]", index);
+
+        // Damage terminal if we change color and it's not a cursor
```
chrisduerr, line 1655:
```suggestion
        // Damage terminal if the color changed and it's not the cursor.
```

chrisduerr (changes requested):

alacritty_terminal/src/term/color.rs
kchibisov: Oh, I was trying to look for it and haven't found.

chrisduerr (approved):

PR #5863, opened 2022-02-07
Use buffer age to perform partial rendering

### LINKED ISSUE #5843: Perform partial rendering when possible
Since we landed damage tracking we could try to go even further and make use of `EGL_EXT_buffer_age` and `GLX_EXT_buffer_age`. So platforms other than Wayland would get benefits from recent damage tracking addition.

I'm not sure whether macOS/Windows provide something similar, but at least it'll work on X11/Wayland. If someone is aware which extensions to use on macOS/Windows for that let me known.

Also, there's some other API (`EGL_KHR_partial_update`), but I'm not sure that it's widely adopted and I only saw it with lima driver.

kchibisov:
This commit makes uses of buffer age extension and recent
alacritty_terminal damage tracking to perform partial redraws on
platforms supporting `GLX_EXT_buffer_age` and `EGL_EXT_buffer_age`.

The alacritty_terminal damage tracking was enhanced in a way that it
reports damage only up to `occ`, so partial redraws during scrolling
are possible.

Fixes #5843.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -651,8 +651,7 @@ dependencies = [
 [[package]]
 name = "glutin"
 version = "0.28.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "00ea9dbe544bc8a657c4c4a798c2d16cd01b549820e47657297549d28371f6d2"
+source = "git+https://github.com/kchibisov/glutin?branch=buffer-age#fc89d3166bd83c003dc4c123bc622546f117c5c9"
 dependencies = [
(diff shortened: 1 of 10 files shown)


kchibisov:
If folks on Wayland/X11 could test this it'll help a lot.

When testing look for a glitches and if you found them describe what triggers and now they look.

For now I'd say that there're some related to clearing in edge cases, since it's not yet optimized. But most things should work and the latency is decreased dramatically.

kchibisov:
@perfbot

PR #5866, opened 2022-02-08
Fix alpha when using `colors.transparent_background_colors`

kchibisov:
The alpha is expected to be premultiplied from the text shader, so
we should apply it to the background color.


chrisduerr (approved):
Yeah that actually looks a lot more correct.

kchibisov:
Yeah, I've noticed it while doing buffer age, since gl clear wasn't matching with background from shader.

PR #5867, opened 2022-02-09
Add support for drawing dashed and dotted underlines

kchibisov:
This finishes implementation of underline styles provided by
`CSI 4 : [1-5] m` escape sequence.

--

I should still update ref test (it doesn't include dotted and dashed lines). Also I'll experiment how we could approach the structure in `alacritty`. For now just look at them visually and what you'd prefer @chrisduerr. I kind of like dashed, but not sure what to do about dotted.

picture

### DIFF
--- a/alacritty/res/rect.f.glsl
+++ b/alacritty/res/rect.f.glsl
@@ -9,46 +9,121 @@ flat in vec4 color;
 
 out vec4 FragColor;
 
-uniform int isUndercurl;
+uniform int rectKind;
 
 uniform float cellWidth;
 uniform float cellHeight;
 uniform float paddingY;
 uniform float paddingX;
 
-uniform float undercurlThickness;
+uniform float underlinePosition;
+uniform float underlineThickness;
+
 uniform float undercurlPosition;
+
+#define UNDERCURL 1
+#define DOTTED 2
+#define DASHED 3
 
 #define PI 3.1415926538
 
-void main()
-{
-    if (isUndercurl == 0) {
-        FragColor = color;
-        return;
+vec4 draw_undercurl(int x, int y) {
+  // We use `undercurlPosition` as an amplitude, since it's half of the descent
+  // value.
+  float undercurl =
+      -1. * undercurlPosition / 2. * cos(float(x) * 2 * PI / cellWidth) +
+      cellHeight - undercurlPosition;
+
+  float undercurlTop = undercurl + max((underlineThickness - 1), 0);
+  float undercurlBottom = undercurl - max((underlineThickness - 1), 0);
+
+  // Compute resulted alpha based on distance from `gl_FragCoord.y` to the
+  // cosine curve.
+  float alpha = 1.;
+  if (y > undercurlTop || y < undercurlBottom) {
+    alpha = 1. - min(abs(undercurlTop - y), abs(undercurlBottom - y));
+  }
+
+  // The result is an alpha mask on a rect, which leaves only curve opaque.
+  return vec4(color.rgb, alpha);
+}
+
+// When the dot size increases we can use AA to make spacing look even and the
+// dots rounded.
+vec4 draw_dotted_aliased(float x, float y) {
+  int dotNumber = int(x / underlineThickness);
+
+  float radius = underlineThickness / 2.;
+  float centerY = cellHeight - underlinePosition;
+
+  float leftCenter = (dotNumber - dotNumber % 2) * underlineThickness + radius;
+  float rightCenter = leftCenter + 2 * underlineThickness;
+
+  float distanceLeft = sqrt(pow(x - leftCenter, 2) + pow(y - centerY, 2));
+  float distanceRight = sqrt(pow(x - rightCenter, 2) + pow(y - centerY, 2));
+
+  float alpha = max(1 - (min(distanceLeft, distanceRight) - radius), 0);
+  return vec4(color.rgb, alpha);
+}
+
+/// Draw dotted line when dot is just a single pixel.
+vec4 draw_dotted(int x, int y) {
+  int cellEven = 0;
+
+  // Since the size of the dot and its gap combined is 2px we should ensure that
+  // spacing will be even. If the cellWidth is even it'll work since we start
+  // with dot and end with gap. However if cellWidth is odd, the cell will start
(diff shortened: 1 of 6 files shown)


kchibisov:
Should be good now @chrisduerr

chrisduerr:
(just testing something) @perfbot

alacritty/src/renderer/rects.rs
```diff
@@ -274,7 +297,8 @@ impl RectRenderer {
             gl::BindBuffer(gl::ARRAY_BUFFER, 0);
         }
 
-        Ok(Self { vao, vbo, program, rect_vertices: Vec::new(), curl_vertices: Vec::new() })
+        let vertices = [Vec::new(), Vec::new(), Vec::new(), Vec::new()];
```
chrisduerr, line 300:
```suggestion
        let vertices = Default::default();
```

This should work, right? Just a lot shorter. Could maybe even be inlined again.

alacritty/src/renderer/rects.rs
```diff
@@ -293,43 +317,31 @@ impl RectRenderer {
...
-            if !self.curl_vertices.is_empty() {
-                self.program.set_undercurl(true);
-                // Upload accumulated undercurl vertices.
-                gl::BufferData(
-                    gl::ARRAY_BUFFER,
-                    (self.curl_vertices.len() * mem::size_of::<Vertex>()) as isize,
-                    self.curl_vertices.as_ptr() as *const _,
-                    gl::STREAM_DRAW,
-                );
+            // We iterate in reverse order to draw plain rects at the end.
```
chrisduerr: Probably a good idea to mention here *why*.

alacritty/res/rect.f.glsl
```diff
@@ -9,46 +9,120 @@ flat in vec4 color;
...
+  // cellWidth, however since the dot + gap is 2 for 1 px dot we should use 2 *
+  // cellWidth when cellWidth is odd.
+  if (int(cellWidth) % 2 != 0) {
+    cellEven = int((gl_FragCoord.x - paddingX) / cellWidth) % 2;
+  }
 
-    // The result is an alpha mask on a rect, which leaves only curve opaque.
-    FragColor = vec4(color.xyz, alpha);
+  // Since we're using the entire descent area for dots as well we should
+  // compute `y` distance to them, so won't look like bars.
```
chrisduerr, line 81:
```suggestion
  // Since we use the entire descent area for dotted underlines, we limit its
  // height to a single pixel so we don't draw bars instead of dots.
```

I think this might be a little easier to understand.

alacritty/res/rect.f.glsl
```diff
@@ -9,46 +9,120 @@ flat in vec4 color;
...
+    alpha = 0;
+  }
+
+  return vec4(color.rgb, alpha);
+}
+
+vec4 draw_dashed(int x) {
+  // Compute the half of the gap, since on the dash is in the middle of the
+  // cell and half_gap from edges, so when we'll have adjacent dashed cell
+  // gaps will have the same size as dash.
```
chrisduerr, line 93: This comment seem somewhat convoluted I'm not quite sure what you're trying to say but it might be beneficial to keep it a bit shorter and more precise?

alacritty/res/rect.f.glsl
```diff
@@ -9,46 +9,120 @@ flat in vec4 color;
...
+  return vec4(color.rgb, alpha);
+}
+
+/// Draw dotted line when dot is just a single pixel.
+vec4 draw_dotted(int x, int y) {
+  int cellEven = 0;
+
+  // To make spacing evenly spread our dot and its gap should divide
+  // cellWidth, however since the dot + gap is 2 for 1 px dot we should use 2 *
+  // cellWidth when cellWidth is odd.
```
chrisduerr, line 75: I'm not entirely sure what you're trying to say with this tbh.

alacritty/res/rect.f.glsl
```diff
@@ -9,46 +9,120 @@ flat in vec4 color;
...
 
-    // We subtract one, since curve is already 1px thick.
-    float undercurl_top = undercurl + max((undercurlThickness - 1), 0);
-    float undercurl_bottom = undercurl - max((undercurlThickness - 1), 0);
+  // The result is an alpha mask on a rect, which leaves only curve opaque.
+  return vec4(color.rgb, alpha);
+}
 
+// When the dot size increases we can use AA to make spacing look even and the
+// dots rounded when the size allows us.
```
chrisduerr, line 52:
```suggestion
// When the dot size increases we can use AA to make spacing look even and the
// dots rounded.
```

Already mentioned in the beginning that we're talking about increased dot size, no need to say it twice.

chrisduerr (changes requested):
I like this **a lot** more than the undercurl PR. :)

alacritty/res/rect.f.glsl
```diff
@@ -9,46 +9,121 @@ flat in vec4 color;
...
+  if (x % 2 != cellEven) {
+    alpha = 0;
+  }
+
+  return vec4(color.rgb, alpha);
+}
+
+vec4 draw_dashed(int x) {
+  // Since dashes of nearest cells connect with each other our dash len is
+  // basically the half of it.
```
chrisduerr, line 94:
```suggestion
  // Since dashes of adjacent cells connect with each other our dash length is half of the desired total length.
```

alacritty/res/rect.f.glsl
```diff
@@ -9,46 +9,121 @@ flat in vec4 color;
...
+
+/// Draw dotted line when dot is just a single pixel.
+vec4 draw_dotted(int x, int y) {
+  int cellEven = 0;
+
+  // Since the size of the dot and its gap combined is 2px we should ensure that
+  // spacing will be even. If the cellWidth is even it'll work since we start
+  // with dot and end with gap, however if cellWidth is odd we'll even with what
+  // we started with. So in that case we should alternate dot and gap in the
+  // first cell, which could be naturally done by using cell number parity.
```
chrisduerr, line 77:
```suggestion
  // with dot and end with gap. However if cellWidth is odd, the cell will start and end with a dot, creating a dash.
  // To resolve this issue, we invert the pattern every two cells.
```

chrisduerr (changes requested):

PR #5869, opened 2022-02-09
Fix warnings when building only with Wayland feature

kchibisov:
This commit also makes our CI fail hard when warning encountered when
building only for either Wayland or X11.


chrisduerr (changes requested):
I don't like switching to `cargo rustc` why would we do this for these features only?

chrisduerr (approved):

PR #5877, opened 2022-02-10
Clear damage rects right after drawing

kchibisov:
Since we could queue damage before we get into the actual rendering
we should clear it after drawing not before.


chrisduerr (commented):
I'd assume this is required for the font damage to work properly?

kchibisov:
>I'd assume this is required for the font damage to work properly?

yeah...

chrisduerr (approved):

PR #5880, opened 2022-02-11
Config warning: background_opacity is deprecated; use window.opacity instead?

achyuthsai87:
[WARN] See log at /tmp/Alacritty-13795.log ($ALACRITTY_LOG):
Config warning: background_opacity is deprecated; use window.opacity instead

There is no file related to Alacritty and even if we have to create it ourselves it what should I have to type in there.

What is the config warning that is above?

Thanks


chrisduerr (changes requested):
Opening Pull Requests as some form of pseudo-issue is a great way to get yourself banned.

PR #5890, opened 2022-02-15
Account for absolute WAYLAND_DISPLAY paths

mahkoh:
If WAYLAND_DISPLAY contains a '/', we have to replace with with another
character before using WAYLAND_DISPLAY as a path component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Buili-in font arcs alignment
 - Repeated permission prompts on M1 macs
 - Colors being slightly off when using `colors.transparent_background_colors`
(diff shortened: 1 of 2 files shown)


chrisduerr (changes requested):
Seems like a sensible workaround. Out of curiosity, did you actually run into this in the wild or did you have to manually create this scenario?

I'd be curious which compositor uses slashes in their `WAYLAND_DISPLAY`.

Also this should be documented in the changelog, because I'd assume some users might actually run into this if there's compositors out there using slashes in `DISPLAY`.

mahkoh:
> Seems like a sensible workaround. Out of curiosity, did you actually run into this in the wild or did you have to manually create this scenario?

I'm currently writing a new wayland that currently uses an absolute path it WAYLAND_DISPLAY.

kchibisov (approved):

chrisduerr (approved):

PR #5893, opened 2022-02-16
Reuse Rasterizer in Display::new

greg904:
Instead of creating a `Rasterizer` to guess the window dimensions,
dropping it and then creating a new one for the glyph cache, reuse the
same `Rasterizer`.

This prevents the font from being loaded twice during startup.


kchibisov (approved):

kchibisov:
@greg904 It turns out that this wasn't enough to not load font twice, however #5894 should address it.

PR #5894, opened 2022-02-16
Don't load font twice during display creation

kchibisov:
This commit finishes the effort from a64553b to avoid reloading font
twice during startup, since the original issue is with getting font
metrics without building the glyph cache.

--
I've also renamed dpr to scale factor, since I don't really like dpr name in general for UI rendering. Since it's a DPR in its nature only on X11, but on other platforms it's a UI scale factor.

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -231,6 +231,7 @@ impl Display {
         };
 
         // Guess the target window dimensions.
+        debug!("Loading \"{}\" font", &config.font.normal().family);
         let font = &config.font;
         let rasterizer = Rasterizer::new(estimated_scale_factor as f32, font.use_thin_strokes)?;
         let mut glyph_cache = GlyphCache::new(rasterizer, font)?;
@@ -272,6 +273,7 @@ impl Display {
         };
 
         // Load font common glyphs to accelerate rendering.
+        debug!("Filling glyph cache with common glyphs");
         renderer.with_loader(|mut api| {
             glyph_cache.load_common_glyphs(&mut api);
         });
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -5,7 +5,6 @@ use std::convert::TryFrom;
 use std::fmt::{self, Formatter};
 #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
 use std::sync::atomic::Ordering;
-use std::time::Instant;
 use std::{cmp, mem};
 
 use glutin::dpi::PhysicalSize;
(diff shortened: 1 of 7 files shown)


alacritty/src/display/mod.rs
```diff
@@ -255,27 +257,49 @@ impl Display {
...
+            let init_start = Instant::now();
+
+            renderer.with_loader(|mut api| {
+                glyph_cache.load_common_glyphs(&mut api);
+            });
+
+            let stop = init_start.elapsed();
+            let stop_f = stop.as_secs() as f64 + f64::from(stop.subsec_nanos()) / 1_000_000_000f64;
+            info!("... finished initializing glyph cache in {}s", stop_f);
+        }
```
kchibisov, line 289: Do we even need this benchmarking?

alacritty/src/display/mod.rs
```diff
@@ -256,26 +258,42 @@ impl Display {
...
 
-        if let Some(dimensions) = dimensions {
-            if (estimated_dpr - window.dpr).abs() < f64::EPSILON {
-                info!("Estimated DPR correctly, skipping resize");
-            } else {
-                // Resize the window again if the DPR was not estimated correctly.
-                let size = window_size(config, dimensions, cell_width, cell_height, window.dpr);
-                window.set_inner_size(size);
-            }
+        // If the scaling factor changed update the glyph cache and resize latter on.
```
chrisduerr, line 267:
```suggestion
        // If the scaling factor changed update the glyph cache and mark for resize.
```

alacritty/src/display/mod.rs
```diff
@@ -256,26 +258,42 @@ impl Display {
...
+        let should_resize = (estimated_scale_factor - window.scale_factor).abs() > f64::EPSILON;
+        let (cell_width, cell_height) = if should_resize {
+            Self::update_glyph_cache(&mut renderer, &mut glyph_cache, scale_factor, config, font)
+        } else {
```
chrisduerr, line 271: No more estimated correctly log?

alacritty/src/display/mod.rs
```diff
@@ -256,26 +258,42 @@ impl Display {
...
+            let init_start = Instant::now();
+
+            renderer.with_loader(|mut api| {
+                glyph_cache.load_common_glyphs(&mut api);
+            });
+
+            let stop = init_start.elapsed();
+            let stop_f = stop.as_secs() as f64 + f64::from(stop.subsec_nanos()) / 1_000_000_000f64;
+            info!("... finished initializing glyph cache in {}s", stop_f);
         }
```
chrisduerr, line 287: I'm not sure it's appropriate to merge this as-is? A fine-grained benchmark like this doesn't provide much value and our log messages already print timestamps anyway.

alacritty/src/display/mod.rs
```diff
@@ -256,26 +258,42 @@ impl Display {
...
 
-        let padding = config.window.padding(window.dpr);
+        if let Some(dimensions) = should_resize.then(|| dimensions).flatten() {
```
chrisduerr, line 289: dimensions.filter(|_| should_resize) shouldn't require .flatten, so I think it's a little neater maybe?

chrisduerr (changes requested):
There's still a mention of DPR in `alacritty/src/renderer/mod.rs:354`.

alacritty/src/display/mod.rs
kchibisov: Yes, I don't see why we'd need it.

kchibisov:
>There's still a mention of DPR in alacritty/src/renderer/mod.rs:354.

That's crossfont API.

chrisduerr (approved):

PR #5900, opened 2022-02-20
Alacritty version 0.10.1

kchibisov:


chrisduerr (changes requested):
Seems to be missing some files, first thing I can think of would be the alacritty_terminal bump.

chrisduerr (approved):

PR #5902, opened 2022-02-21
Add fallback GLES2 renderer

### LINKED ISSUE #128: Support GPUs without OpenGL 3.3+ drivers
I've been looking for an extremely fast terminal emulator for many years. `xterm` has unfixed rendering bugs and is slower than molasses. I don't like `urxvt`'s lack of memory management. `st` uses more memory than `xterm` (yep!).

Just saw Alacritty on HN and was determined to try it out.

What are your plans for older hardware? I'm currently using a couple of laptops from 2007-2008: a ThinkPad T43 and a ThinkPad T60. Unfortunately the Rust nightly you're currently using is broken on i686 (#126) so I can't test Alacritty on the T43, but it reported the error in the title on my T60 (which has a slightly better GPU, FWIW).

T43 specs:
```
OpenGL vendor string: X.Org R300 Project
OpenGL renderer string: Gallium 0.4 on ATI RV370
OpenGL version string: 2.1 Mesa 11.0.2
OpenGL shading language version string: 1.20

01:00.0 VGA compatible controller: AMD/ATI [Advanced Micro Devices, Inc.] RV370/M22 [Mobility Radeon X300]
```

T60 specs:
```
Extended renderer info (GLX_MESA_query_renderer):
    Vendor: X.Org R300 Project (0x1002)
    Device: ATI RV515 (0x7145)
    Version: 11.2.2
    Accelerated: yes
    Video memory: 128MB

01:00.0 VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] RV515/M54 [Mobility Radeon X1400]
```

Yep, I realize these are kind of the "oh wow" of GPU land :) but I'm genuinely wondering what the best approach would be to support GPUs like these - I've been pondering making my own terminal emulator for a little while (as well as a 2D engine too, actually), but have absolutely no knowledge of OpenGL, Vulkan, etc.

"It would be impossible" is completely understandable; "it would require a nontrivial architecture redesign" also wouldn't be too surprising.

kchibisov:
This commit adds a fallback GLES2 renderer for platforms not supporting
OpenGL 3.3 providing ability to run alacritty on those. While its
performance is slightly slower than glsl3 version it's way faster than
software rendering with sofware emulated GL (e.g. llvmpipe), since
that's what alacritty used on platforms not supporting required OpenGL
version.

Fixes #128.

--
I've managed to run it on pi without any issues.


kchibisov:
Blocked on https://github.com/alacritty/alacritty/pull/5904

README.md
```diff
@@ -47,7 +47,7 @@ For everyone else, the detailed instructions to install Alacritty can be found
 
 ### Requirements
 
-- OpenGL 3.3 or higher
+- OpenGL ES 2.0 or higher
```
kchibisov, line 50: I'm not sure how to adjust it, since I have a feeling that OpenGL 2.1 device without gles2 won't work, since shaders are gles2. Though it's a matter of adjusting syntax in shaders.

chrisduerr: Should be almost irrelevant at that point. But `GLES2 support` might suffice?

kchibisov:
This is ready for review. The feature parity with glsl3 render was reached, fallback stratagy and shaders reuse were added.

chrisduerr:
@perfbot

chrisduerr:
@perfbot

alacritty/src/renderer/shader.rs
```diff
@@ -60,20 +80,31 @@ impl Drop for ShaderProgram {
...
+                source.as_ptr() as *const _,
+                len.as_ptr(),
+            );
             gl::CompileShader(shader.id());
             gl::GetShaderiv(shader.id(), gl::COMPILE_STATUS, &mut success);
         }
 
         if success != GLint::from(gl::TRUE) {
-            return Err(ShaderError::Compile(get_shader_info_log(shader.id())));
+            panic!("{}", ShaderError::Compile(get_shader_info_log(shader.id())));
```
chrisduerr, line 107: Did you intentionally change to panic here or was that just for testing?

chrisduerr (changes requested):
I'll take a look at gles2.rs/gles3.rs and the gles2 shaders later, but the rest of the code looks pretty good already.

chrisduerr:
@perfbot

alacritty/src/renderer/mod.rs
```diff
@@ -82,588 +60,98 @@ impl From<ShaderError> for Error {
...
 
-    #[inline]
-    pub fn size(&self) -> usize {
-        self.len() * size_of::<InstanceData>()
-    }
-
-    pub fn clear(&mut self) {
-        self.tex = 0;
-        self.instances.clear();
+        match &mut self.text_renderer {
```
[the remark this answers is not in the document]
kchibisov: I'm not sure I understand what do you want here. I need to match enum anyway and call the proper implementation. I can't reuse with_api, since Api are different and you can't make closures generic over argument, only functions. If I implement `draw_cells` on enum it'll still require me to do `Self::Glsl3(renderer) => renderer.draw_cells()`, no?

chrisduerr:
@perfbot

chrisduerr:
@perfbot

alacritty/src/renderer/mod.rs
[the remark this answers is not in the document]
chrisduerr: Correct, but you'd only have to do it once. You're currently duplicating the logic because you're calling `draw_calls` in two different locations.

alacritty/src/renderer/shader.rs
```diff
@@ -60,23 +80,34 @@ impl Drop for ShaderProgram {
...
-    fn new(kind: GLenum, source: &'static str) -> Result<Self, ShaderError> {
-        let len: [GLint; 1] = [source.len() as GLint];
+    fn new(
+        shader_version: ShaderVersion,
+        kind: GLenum,
+        source: &'static str,
+    ) -> Result<Self, ShaderError> {
+        let header = shader_version.shader_header();
+        let len: [GLint; 2] = [header.len() as GLint, source.len() as GLint];
+        let source = [header.as_ptr() as *const _, source.as_ptr() as *const _];
```
chrisduerr, line 90: You're assembling the shader here in OpenGL, any specific reason for that? I feel like the more intuitive approach would be to just append the two strings to each other?

alacritty/src/renderer/text/gles2.rs
```diff
@@ -0,0 +1,494 @@
...
+impl Gles2Renderer {
+    pub fn new() -> Result<Self, Error> {
+        info!("Using OpenGL ES 2.0 renderer");
+
+        let program = TextShaderProgram::new(ShaderVersion::Gles2)?;
+        let mut vao: GLuint = 0;
+        let mut vbo: GLuint = 0;
+        let mut ebo: GLuint = 0;
+
+        let mut indices = Vec::with_capacity(BATCH_MAX / 4 * 6);
```
chrisduerr, line 46: Looks like these are vertex indices, right? If we have the space I'd probably go ahead and just name it `vertex_indices` just to make that obvious. Just `indices` isn't super clear on what their purpose is.

alacritty/src/renderer/text/gles2.rs
```diff
@@ -0,0 +1,494 @@
...
+                    );
+                    gl::EnableVertexAttribArray(index);
+
+                    #[allow(unused_assignments)]
+                    {
+                        size += $count * size_of::<$type>();
+                        index += 1;
+                    }
+                };
+            }
```
chrisduerr, line 107: Was considering to suggest abstracting this macro but it's so short that I don't think it's worth it considering one has the divisor while the other doesn't.

alacritty/src/renderer/text/gles2.rs
```diff
@@ -0,0 +1,494 @@
...
+    }
+
+    fn loader_api(&mut self) -> LoaderApi<'_> {
+        LoaderApi {
+            active_tex: &mut self.active_tex,
+            atlas: &mut self.atlas,
+            current_atlas: &mut self.current_atlas,
+        }
+    }
+}
```
chrisduerr, line 221: This trait implementation is basically exactly the same for both renderers, which is usually a sign some other abstraction might be beneficial?

Instead of the `resize` method, you could use a `program` method for example which would allow you to just have one line duplicated which would be all safe code too.

The `draw_cells` could maybe be abstracted to `with_api`? The function exists already anyway and at that point we'd just define `draw_cells` once somewhere instead of doing it twice.

alacritty/src/renderer/text/gles2.rs
```diff
@@ -0,0 +1,494 @@
...
+
+/// Maximum items to be drawn in a batch.
+///
+/// We use the closest dividable by 4(amount of vertices we push for a glyph) number to `u16::MAX`,
+/// since it's maximum possible index in `glDrawElements` in gles2.
```
chrisduerr, line 226:
```suggestion
/// We use the closest number to `u16::MAX` dividable by 4 (amount of vertices we push for a glyph),
/// since it's the maximum possible index in `glDrawElements` in gles2.
```

chrisduerr (changes requested):

alacritty/src/renderer/shader.rs
kchibisov: I mean there's a native way to append shaders, so I don't see why not? Instead of randomly appending rust strings.

chrisduerr: I'm just curious, I'm surprised it works like this at all? And surprises are usually a bad thing.

alacritty/src/renderer/mod.rs
```diff
@@ -1,33 +1,26 @@
...
+use crate::renderer::shader::ShaderError;
 
-pub mod builtin_font;
 pub mod rects;
 mod shader;
+mod text;
+
+use shader::ShaderVersion;
+use text::Glsl3Renderer;
+pub use text::{Gles2Renderer, GlyphCache, LoaderApi, TextRenderer};
```
chrisduerr, line 23:
```suggestion
pub use text::{Gles2Renderer, GlyphCache, LoaderApi, TextRenderer};
use shader::ShaderVersion;
use text::Glsl3Renderer;
```

Not sure if rustfmt did this, but `pub` should usually come first.

alacritty/src/renderer/text/gles2.rs
```diff
@@ -0,0 +1,478 @@
...
+impl<'a> TextRenderer<'a> for Gles2Renderer {
+    type RenderApi = RenderApi<'a>;
+    type RenderBatch = Batch;
+    type Shader = TextShaderProgram;
+
+    fn program(&self) -> &Self::Shader {
+        &self.program
+    }
+
+    fn with_api<'b: 'a, F, T>(&'b mut self, _: &'b SizeInfo, func: F) -> T
```
chrisduerr, line 165: Bit unfortunate this method still seems to basically be completely duplicated? Though I don't think there's any great way to solve this.

kchibisov: Yeah, I don't see how it could be solved, given the types are just different.

alacritty/src/renderer/shader.rs
kchibisov: The documentation clearly states that it concatenates all the sources, since api accepts array of sources. I'm not sure you should be surprised by documented feature.

chrisduerr (approved):

PR #5905, opened 2022-02-22
Omit git version hash when git command fails

nibon7:


chrisduerr (changes requested):
What's the motivation behind this?

nibon7:
While building with release tarball,  `commit_hash`  returns `Some("")`,   and the `VERSION` environment  contains an empty git hash, like `0.11.0-dev ()`.

chrisduerr (approved):

PR #5912, opened 2022-02-25
Add colored underline support

### LINKED ISSUE #4142: Support for setting individual color for underline (<ESC>[58...m)

kchibisov:
This commit adds support for colored underline by packing all the cell
colors and using the mask to access them, so it keeps the cell in the
same size of 24 bytes, but have colored underlines being statically
allocated on the cell.

Fixes #4142.

--

That's what I managed to get on my machine.

colored_underline


kchibisov:
@perfbot

kchibisov:
@perfbot

kchibisov:
@chrisduerr the perf results look good to me and I'm not sure how to improve it further... But packing colors wasn't that bad idea.

chrisduerr:
> But packing colors wasn't that bad idea.

I'm not sure, the code seems rather complicated? For a feature that nobody cares about that seems rather unfortunate?

kchibisov:
@chrisduerr I guess, but it's not exposed in public API and not even visible to anything other than cell.rs internals. And most of them just checking mask and returning color. I'm kind of interested in bringing this feature, but I'm not sure I could find anything better (The other approach is to shorten pointer for Box, but I think it's more complicated).

chrisduerr:
>I guess, but it's not exposed in public API and not even visible to anything other than cell.rs internals.

Public API is kinda irrelevant since we're the only ones using our code anyway? So the way things work internally are just as important as the way things are exposed externally.

> I'm kind of interested in bringing this feature, but I'm not sure I could find anything better

Unfortunately I don't have a clear answer to this either, I'd have to look at it myself and that would likely also only leave me in the same position you're in. But looking at the code the way it is right now, I can't really say that it's elegant or pretty. It's just a bunch of hacky bits cobbled together to try and make things fit.

bluz71:
> I'm not sure, the code seems rather complicated? **For a feature that nobody cares about** that seems rather unfortunate?

@chrisduerr,

It is incorrect to say that nobody cares, I care very much. It's my number one most wanted feature for Alacritty, and has been for years.

Look at PR #4660, there were 41 hearts attached to the first post. That's at least 41 users who care deeply. Same goes for issue #4142, ten thumbs up to the first.

Is it the most important feature in the world, clearly no. But, some folks care. It is not correct to say nobody cares.

For some of us it is extremely important with regards to nice-looking error/lint alerts in editors such as Neovim (for example LSP diagnostics).

chrisduerr:
> It is incorrect to say that nobody cares, I care very much. It's my number one most wanted feature for Alacritty, and has been for years.

Fair enough, though that doesn't improve the code quality. And with the number of terminals that just don't support it at all - and people very willing to still use them - it at least seems to suggest that talking on a bigger scale it just seems to be "nice to have" for most people rather than essential.

Alacritty of course is aimed at the people that use terminals a lot, not just the general Linux populace, so I can see how some of those might enjoy this. But I'd rather try to do it right than going for this mess if there is any option to do so.

bluz71:
The following modern terminals do support coloured underline: kitty, GNOME Terminal, any terminal that uses libvte such as LilyTerm, iTerm2, wezterm and even the suckless st terminal.

Support for the feature is fairly wide nowadays (ignoring old school terminals such as XTerm and Apple Terminal).

Though I do understand your desire for clean code, but it does seem, from the outside that performance parity results in ugly code whilst cleaner code results in a small amount of performance or greater memory use.

Personally I favour an option (maybe defaulting to off) that enables coloured underline with a small loss of performance or increased memory use.

Like native box drawing characters there is a desire for this feature by some in the Alacritty community.

Hopefully the Alacritty team can eventually come to an implementation of this feature that has the least compromises in terms of code cleanliness and performance.

Cheers.

chrisduerr:
> Personally I favour an option (maybe defaulting to off) that enables coloured underline with a small loss of performance or increased memory use.

Alacritty doesn't provide options to opt out of good performance. Never have and never will (as long as I'm around at least). Either we make everything fast, or it won't be merged. @kchibisov already managed to make it fast, now we just need to figure out how to clean it up.

bluz71:
Fair enough, hopefully code cleanliness is possible. I have my fingers crossed.

I should go back to lurking in the shadows; less taking up of Alacritty team time.

Cheers.

kchibisov:
@perfbot

kchibisov:
@perfbot

kchibisov:
@perfbot

alacritty_terminal/src/ansi.rs
```diff
@@ -1392,6 +1394,17 @@ fn attrs_from_sgr_parameters(params: &mut ParamsIter<'_>) -> Vec<Option<Attr>> {
...
             },
             [49] => Some(Attr::Background(Color::Named(NamedColor::Background))),
+            [58] => {
+                let mut iter = params.map(|param| param[0]);
+                parse_sgr_color(&mut iter).map(|color| Attr::UnderlineColor(Some(color)))
+            },
+            [58, params @ ..] => {
+                let rgb_start = if params.len() > 4 { 2 } else { 1 };
+                let rgb_iter = params[rgb_start..].iter().copied();
+                let mut iter = iter::once(params[0]).chain(rgb_iter);
```
chrisduerr, line 1404: This is copy/pasted from the 38/48 escape I'd assume. At this point it might be worth it to extract it to a function? Especially because it's not 100% obvious what this does and a comment about the format might actually help readers.

alacritty_terminal/src/ansi.rs
```diff
@@ -1392,6 +1394,17 @@ fn attrs_from_sgr_parameters(params: &mut ParamsIter<'_>) -> Vec<Option<Attr>> {
...
+
+                parse_sgr_color(&mut iter).map(|color| Attr::UnderlineColor(Some(color)))
+            },
             [90] => Some(Attr::Foreground(Color::Named(NamedColor::BrightBlack))),
```
chrisduerr, line 1408: Missing [59] for reset underline color?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Escape sequence for undercurl, dotted and dashed underlines (`CSI 4 : [3-5] m`)
 - `ToggleMaximized` key binding action to (un-)maximize the active window, not bound by default
 - Support for OpenGL ES 2.0
+- Underline color escape sequence (`CSI 58 : 2 : Ps : Ps : Ps m`/`CSI 58 : 5 : Ps m`)
```
chrisduerr, line 20: CSI 59 should also be documented.

alacritty/src/renderer/rects.rs
```diff
@@ -193,6 +193,9 @@ impl RenderLines {
             return;
         }
 
+        // Strikethrough isn't using the same color underline is using.
+        let color = if flag.contains(Flags::STRIKEOUT) { cell.fg } else { cell.underline };
```
chrisduerr, line 197: Rather than stating it isn't, I'd also state why it's not:

```suggestion
        // The underline color escape does not apply to strikeout.
        let color = if flag.contains(Flags::STRIKEOUT) { cell.fg } else { cell.underline };
```

alacritty_terminal/src/term/cell.rs
```diff
@@ -144,14 +377,16 @@ impl GridCell for Cell {
...
     }
 }
 
 impl From<Color> for Cell {
     #[inline]
     fn from(color: Color) -> Self {
-        Self { bg: color, ..Cell::default() }
+        let mut cell = Cell::default();
+        cell.colors.set_bg(color);
+        cell
```
chrisduerr, line 389: I'd be somewhat worried of the optimization of this, it's probably a bit hot iirc. So changing this to something that is significantly less efficient will probably affect performance at least a little bit?

alacritty_terminal/src/term/cell.rs
```diff
@@ -47,7 +50,244 @@ impl<T: Copy> ResetDiscriminant<T> for T {
...
+
+    #[inline]
+    fn has_underline(&self) -> bool {
+        self.mask & Self::HAS_UNDERLINE != 0
+    }
+}
+
+#[repr(C, packed)]
+#[derive(Copy, Clone)]
+pub union PackedColor {
```
chrisduerr, line 132: You mentioned on IRC you had an idea on how to clean this up, but I don't see any significant difference? This 250 lines of garbage is still something I can't really warm up with.

alacritty_terminal/src/term/cell.rs
```diff
@@ -47,7 +50,244 @@ impl<T: Copy> ResetDiscriminant<T> for T {
...
+        self.bg.into_color(self.mask.bg_ty())
+    }
+
+    #[inline]
+    pub fn underline(&self) -> Option<Color> {
+        if self.mask.has_underline() {
+            Some(self.underline.into_color(self.mask.underline_ty()))
+        } else {
+            None
+        }
```
chrisduerr, line 290: Why bother returning an Option when the underline will always be the fg otherwise anyway? No reason to do this at the callee site, right?

chrisduerr (changes requested):

kchibisov:
Closing in favor of #5912.

jez:
👋 for anyone following this PR it looks like maybe #5935 is the intended PR (#5912 is this PR 😅)

kchibisov:
@jez, yeah, I have too many approaches to the same feature already.

bluz71:
> @jez, yeah, I have too many approaches to the same feature already.

Hopefully you are narrowing in on the best approach. I have my fingers and toes crossed :)

#5935 looks interesting, if I am reading the @perfbot graph correctly (in that PR), your newest colored underline approach has improved performance?

PR #5915, opened 2022-02-26
Fix window layout on macOS when create window

Jamyw7g:
On macOS, when current window is Fullscreen, new window should be Windowed. Otherwise it will create two fullscreen window and a tab bar on the top each window, and the window cannot be closed correctly.
If current window is fullscreen, and new window is windowed, they will tile on the same window as tab. I think this is the best way to arrange windows.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -37,6 +37,7 @@ use alacritty_terminal::term::{ClipboardType, SizeInfo, Term, TermMode};
 use crate::cli::{Options as CliOptions, WindowOptions};
 use crate::clipboard::Clipboard;
 use crate::config::ui_config::{HintAction, HintInternalAction};
+use crate::config::window::StartupMode;
 use crate::config::{self, UiConfig};
 #[cfg(not(windows))]
 use crate::daemon::foreground_process_path;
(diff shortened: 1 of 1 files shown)


chrisduerr (commented):
I don't have access to a macOS machine to test this, but if this breaks the macOS WM shouldn't it be automatically handled by macOS?

I'd assume that if it is legal to open two fullscreen windows, there should be a way to make it functional too?

Jamyw7g:
# Let me show you by videos
 
## Before

https://user-images.githubusercontent.com/30200782/155869119-2055c826-db8e-4dc7-8390-df8558ebef59.mp4


## After
https://user-images.githubusercontent.com/30200782/155868757-d1b478cf-83bf-4030-9715-75d72260b4d7.mp4

chrisduerr (commented):
Seems to me like this is just a curiosity of macOS' tabbing? I don't see how this would be a big problem and I can definitely see how removing this option would hinder some people.

Jamyw7g:
This may be related to the window management mechanism of macOS. When the application is full screen, it will create a new desktop, but the `alacrity` CreateNewWindow is created in the form of tab, which will lead to confusion of the system.
So I change that when the current window is Fullscreen, it will be created in the form of Windowed, so there will be no chaotic window management of appeal.

chrisduerr:
> but the alacritty CreateNewWindow is created in the form of tab, which will lead to confusion of the system.

This is an option *you* opted into though. Others might not have it behave that way. So why should we break *their* workflow just because it mildly inconveniences you?

Jamyw7g:
> This is an option _you_ opted into though.

This is the default behavior. I think if you can, you can test it on macOS. The configuration is as follows,
```
window:
  startup_mode: Fullscreen
```
When the program launched, you can test the operation of creating window, closing window and so on.

chrisduerr:
> This is the default behavior. I think if you can, you can test it on macOS.

It is not, I have tested it.

There might be some option for enhancement here, but this is not it.

PR #5918, opened 2022-02-26
Add a ToggleMaximized key binding action

chrisnc:
enable the set_maximized Window method on all platforms

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Track and report surface damage information to Wayland compositors
 - Escape sequence for undercurl, dotted and dashed underlines (`CSI 4 : [3-5] m`)
+- `ToggleMaximized` key binding action to (un-)maximize the active window, not bound by default
 
 ### Changed
 
@@ -28,7 +29,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
- - Option `font.builtin_box_drawing` to disable the built-in font for drawing box characters
+- Option `font.builtin_box_drawing` to disable the built-in font for drawing box characters
 
 ### Changed
 
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -186,6 +186,9 @@ pub enum Action {
     /// Toggle fullscreen.
     ToggleFullscreen,
 
+    /// Toggle maximized.
+    ToggleMaximized,
+
     /// Toggle simple fullscreen on macOS.
     #[cfg(target_os = "macos")]
     ToggleSimpleFullscreen,
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -249,7 +249,7 @@ impl Display {
         debug!("Estimated cell size: {} x {}", cell_width, cell_height);
 
         // Spawn the Alacritty window.
-        let mut window = Window::new(
+        let window = Window::new(
             event_loop,
             config,
             identity,
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -235,7 +235,7 @@ impl Window {
     }
 
     #[inline]
-    pub fn set_inner_size(&mut self, size: PhysicalSize<u32>) {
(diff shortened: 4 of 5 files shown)


chrisnc:
This change works as expected on macOS and allows me to use:
```
key_bindings:
  - { key: M, mods: Command|Shift, action: Maximize }
```

chrisnc:
It's not obvious to me why the `set_maximized` wrapper was disabled on macOS and Windows before, but interested to know if enabling it is a problem.

chrisduerr:
> It's not obvious to me why the set_maximized wrapper was disabled on macOS and Windows before, but interested to know if enabling it is a problem.

Probably because it wasn't implemented yet.

chrisnc:
As I'm using this, I think I want to make one minor enhancement so it behaves like other things on macOS: pressing the shortcut again should un-maximize it. This seems possible with `winit`, but the `is_maximized` method is not supported on Wayland. I'll do some more experimenting...

chrisduerr (changes requested):
See comment.

chrisnc:
Sounds good. I'll do that. Is it okay if I constrain that action to non-Wayland?

chrisnc:
One other thing: I noticed that some of the methods on Window take `&mut self` but only need `&self`. Do you mind if I change those?

chrisduerr:
>  Edit: the comment on the top-level is_maximized in winit says it's not implemented on Wayland, but the platform_impl for Wayland does have an implementation for it, so maybe that's not necessary and either it already works, or it will start working upon an update.

Restricting it to a specific platform is only necessary when that platform does not offer the API at all and thus would cause a compile/runtime error. If it's just a no-op, we can just call it anyway since it won't have any impact.

In this case however I think you're correct and Wayland does implement `set_maximized`.

> One other thing: I noticed that some of the methods on alacritty's Window take &mut self but only need &self. Do you mind if I change those?

That is fine, yes, as long as there are no `RefCell`s or similar being used in the background which require the mutability hint without the compiler forcing it on us.

chrisduerr:
> It does look like both are implemented and the comment is out of date. I don't have a Wayland system to test this on, though.

Looks to me like the variable is loaded, but never stored. So I'd be surprised if `is_maximized` works on Wayland. You'd probably have to track this internally seeding the initial state using `is_maximized` at startup, or alternatively send a patch upstream.

@kchibisov Do you know why this isn't implemented?

chrisnc:
Ah, okay. I misread the source then. As-is I think this would maximize and then not support un-maximizing, until such time as winit implements `is_maximized` on Wayland. Do you think that would be acceptable? Tracking it within Alacritty seems like overkill.

chrisnc:
Hm, looking more closely, it seems that there's an `Arc` around the maximize, which is cloned and then modified in the event loop. That `maximzied_clone.store(is_maximized, Ordering::Relaxed);` line should be updating the field, no?

chrisduerr:
> Hm, looking more closely, it seems that there's an Arc around the maximized field, which is cloned and then modified in the event loop. That maximzied_clone.store(is_maximized, Ordering::Relaxed); line should be updating the field, no?

Oh, that's a good point I didn't actually spot that because of the typo. So it should work then. Probably best to just spin up sway and try though.

chrisnc:
I'll make a VM and give it a try.

kchibisov:
>@kchibisov Do you know why this isn't implemented?

I don't know, I'll send a patch. Probably I haven't noticed it.

kchibisov:
it's implemented on Wayland, why do you think it's not implemented there?

kchibisov:
I'll test Wayland bits, don't worry. I have everything setup anyway.

chrisnc:
Looks like it was implemented for both in this commit, but the doc comment was not updated:
https://github.com/rust-windowing/winit/commit/c916eb613734b389d7add7efec7f698ca739bae2

chrisduerr (approved):
LGTM

chrisduerr:
Just for the record, since the fullscreen action is called `ToggleFullscreen` and not `ToggleFullscreened` I considered going for `ToggleMaximize` maybe, but that just sounds stupid. Fullscreen is somewhat special in that it is both the verb and the noun so I think `Maximized` is more appropriate here.

chrisnc:
Yes, I was thinking about that, but I think the participle is more consistent so they are both adjectives, and also it matches the winit method.

PR #5922, opened 2022-02-28
Extract `SizeInfo` from alacritty_terminal

kchibisov:
The `SizeInfo` is a SizeInfo used for rendering, which contains
information about padding, and such, however all the terminal need is
number of visible lines and columns.

### DIFF
--- a/alacritty/src/window_context.rs
+++ b/alacritty/src/window_context.rs
@@ -24,6 +24,7 @@ use alacritty_terminal::event_loop::{EventLoop as PtyEventLoop, Msg, Notifier};
 use alacritty_terminal::grid::{Dimensions, Scroll};
 use alacritty_terminal::index::Direction;
 use alacritty_terminal::sync::FairMutex;
+use alacritty_terminal::term::test::TermSize;
 use alacritty_terminal::term::{Term, TermMode};
 use alacritty_terminal::tty;
 
@@ -98,7 +99,7 @@ impl WindowContext {
         // This object contains all of the state about what's being displayed. It's
         // wrapped in a clonable mutex since both the I/O loop and display need to
         // access it.
-        let terminal = Term::new(&config.terminal_config, display.size_info, event_proxy.clone());
+        let terminal = Term::new(&config.terminal_config, &display.size_info, event_proxy.clone());
         let terminal = Arc::new(FairMutex::new(terminal));
 
         // Create the PTY.
@@ -106,7 +107,7 @@ impl WindowContext {
         // The PTY forks a process to run the shell on the slave side of the
         // pseudoterminal. A file descriptor for the master side is retained for
         // reading/writing to the shell.
-        let pty = tty::new(&pty_config, &display.size_info, display.window.x11_window_id())?;
+        let pty = tty::new(&pty_config, display.size_info.into(), display.window.x11_window_id())?;
 
         #[cfg(not(windows))]
         let master_fd = pty.file().as_raw_fd();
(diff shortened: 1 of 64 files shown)


kchibisov:
I'm not sure that it's actually something that will make sense, since I'm not sure how something like sixel deals with terminal padding. But I'm not sure that it really matters, since all that matters for such applications are the size of the window they can actually draw into(the one without padding).

chrisduerr:
> since I'm not sure how something like sixel deals with terminal padding

I mean speaking logically, even graphics rendering should be based on a "virtual" origin that sits at (0, 0) of the terminal content, ignoring potential for padding. Taking padding into account should then be the job of the renderer (`alacritty`).

alacritty_terminal/src/event.rs
```diff
@@ -38,6 +38,9 @@ pub enum Event {
     /// Write some text to the PTY.
     PtyWrite(String),
 
+    /// Request to write the text area size.
+    TextAreaSize(Arc<dyn Fn(WinSize) -> String + Sync + Send + 'static>),
```
chrisduerr, line 42: Our color variant is called `ColorRequest`, should we maybe rename this to `TextAreaSizeRequest` for consistency?

alacritty/src/window_context.rs
```diff
@@ -98,15 +99,15 @@ impl WindowContext {
...
+        let terminal = Term::new(&config.terminal_config, &display.size_info, event_proxy.clone());
         let terminal = Arc::new(FairMutex::new(terminal));
 
         // Create the PTY.
         //
         // The PTY forks a process to run the shell on the slave side of the
         // pseudoterminal. A file descriptor for the master side is retained for
         // reading/writing to the shell.
-        let pty = tty::new(&pty_config, &display.size_info, display.window.x11_window_id())?;
+        let pty = tty::new(&pty_config, display.size_info.into(), display.window.x11_window_id())?;
```
chrisduerr, line 110: What's the point of having `WinSize` when all we do with it is use it to convert `SizeInfo` into it, which then is just used to create `libc`'s `winsize`? Might as well just take `winsize` or `impl ToWinsize` as parameter and save yourself the extra useless type.

chrisduerr (changes requested):

alacritty/src/window_context.rs
kchibisov: It's due to Windows, since it doesn't have libc::winsize.

chrisduerr: The trait could still be used, right?

kchibisov: The thing is that this type should be somewhere, since windows should have some inner type for that. The trait is used under the hood for unix anyway.

How we should proceed with it?

chrisduerr: Can you not take a trait impl instead of a custom type plus a trait impl?

alacritty_terminal/src/term/mod.rs
```diff
@@ -2077,10 +1918,11 @@ impl<T: EventListener> Handler for Term<T> {
 
     #[inline]
     fn text_area_size_pixels(&mut self) {
-        let width = self.cell_width * self.columns();
-        let height = self.cell_height * self.screen_lines();
-        let text = format!("\x1b[4;{};{}t", height, width);
-        self.event_proxy.send_event(Event::PtyWrite(text));
+        self.event_proxy.send_event(Event::TextAreaSizeRequest(Arc::new(move |win_size| {
```
chrisduerr, line 1921:
```suggestion
        self.event_proxy.send_event(Event::TextAreaSizeRequest(Arc::new(move |window_size| {
```

Just 3 characters more and still under the 100 character limit. So probably not worth to cut things here.

chrisduerr (changes requested):
Really don't like naming our struct `WinSize` and `window_size` everywhere. If we need a new struct anyway for cross-platform compatibility, we shouldn't be forced to name it the same way it is on Unix. So `WindowSize` seems superior.

chrisduerr (approved):

PR #5925, opened 2022-03-02
Add option to specify the flashing area of the visual bell

z33ky:
I mostly use fullscreen terminals and having the whole screen flash looks kinda annoying.
I've been using this for a while and thought maybe someone else would like this option. If you don't like it though (as I remember you're not a fan of very minute config options) I'm fine patching the code locally instead of having the feature integrated in upstream.


alacritty/src/config/bell.rs
```diff
@@ -1,11 +1,96 @@
...
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+struct BellRadii {
+    /// Horizontal radius in units of cells.
+    pub horizontal: f32,
+    /// Vertical radius in units of cells.
+    pub vertical: f32,
+}
+
+pub struct Rect {
```
z33ky, line 20: This type feels a bit out-of-place here. It could be moved somewhere else and perhaps also be used inside `renderer::rects::RenderRect`?

kchibisov:
Given that visual bell is the less used option. And I'm not sure why you'd want to bell just some area I don't think it should be upstreamed.

Thanks for the patch though.

Issue #4939, opened 2021-04-04
Colors are rendered incorrectly


simplelife2010:
### System

OS: macOS
Version: alacritty 0.7.2 (5ac8060)

### Steps to reproduce

- Under System Preferences, choose or check your Color Profile (sRGB)
- Disable True Tone under your Mac's Display options
- Start Alacritty and 1 or 2 other Terminal emulators (Terminal.app, iTerm2)
- Make sure the windows are not overlapped by shadows from other windows
- Use the same color scheme (Solarized Light) in all Terminal Emulators
- If using Terminal.app, make sure the same color profile is selected as for your Display when choosing RGB colors
- Open Color Picker (launch-colorpicker)
- Set the color picker to the same color profile as your display (sRGB)
- Set the color picker to display RGB values
- Choose a color from the color scheme that you want to check (Background color: #fdf6e3)
- Check Alacritty and the other Terminal emulators with the color picker: Is the color rendered correctly?

### Expected result

- Color picker shows same RBG value (#fdf6e3)

### Acutal result

- Alacritty: FEF6E1
- iTerm2: FDF6E3
- Terminal.app: FDF6E3
- MacVim: FDF6E3

Comment: Same problem when checking if foreground colors are correct. Red for example is rendered correctly to DC322F in iTerm2, but to EF0C20 in Alacritty, where the difference is clearly visible. Same problem when using other Color Profiles.

simplelife2010:
Correction: When using a P3 Profile, the color values are almost matching in Alacritty. I get the impression that Alacritty color settings is somehow bound to the P3 Profile. Can you confirm that?

In iTerm and Terminal.app the default seems to be sRGB, but can be changed to a different profile if needed.

nixpulvis:
Looks like this was requested before, but closed by the author in #1599. There also seems to have been a bit of work around sRGB formats before to appease some systems in 6debc4f33.

> I get the impression that Alacritty color settings is somehow bound to the P3 Profile. Can you confirm that?

I wouldn't say we're mapped to it so much as we just don't do any conversions (assuming `with_srgb(false)` worked) with OpenGL. Since I doubt we would want alacritty having to deal with color spaces itself, I actually wonder if this is the right default. It seems like it could be the simplest default to get _some_ color to be rendered, but it seems to me that it could easily render the wrong colors, as others have mentioned.

[...]

Issue #3756, opened 2020-05-21
Subpixel fonts aren't rendered correctly


20k:
### System

OS: Windows
Version: alacritty 0.4.2 (f68de37)
Windows: ConPTY

### Logs

Font/Terminal size: 9 with bitstream vera sans mono

Hello! Alacritty doesn't handle rendering subpixel fonts correctly. I thought I'd document the issue in depth:

This is what an example from alacritty looks like:

![8293632 539999992_alacritty](https://user-images.githubusercontent.com/755197/82596219-db069080-9b9e-11ea-93bb-7d023e95ce65.PNG)

This is sample text rendered in a different application that is rendered incorrectly, similar settings:

![LEGACY2](https://user-images.githubusercontent.com/755197/82596260-ebb70680-9b9e-11ea-910c-1557ad2739af.PNG)

This is sample text correctly rendered with sRGB correct blending:

![good1](https://user-images.githubusercontent.com/755197/82596295-f8d3f580-9b9e-11ea-8c9f-03a02d6709a5.PNG)

I had a brief look through the rendering pipeline, as far as I can tell the issue is here:

```// Regular text glyphs.
            vec3 textColor = texture(mask, TexCoords).rgb;
            alphaMask = vec4(textColor, textColor.r);
            color = vec4(fg, 1.0);```

Combined with this piece of code:

```        let windowed_context =
            create_gl_window(window_builder.clone(), &event_loop, false, vsync, size)
                .or_else(|_| create_gl_window(window_builder, &event_loop, true, vsync, size))?;```

This looks like it tries to create a window with sRGB disabled, followed by making one with sRGB enabled if that fails. Currently alacritty only renders text correctly with sRGB enabled, but there's no reason why both can't work

textColor here looks like alpha for subpixel fonts, which is good, although strictly speaking it is alpha, not colour. The blend equation is

```gl::BlendFunc(gl::SRC1_COLOR, gl::ONE_MINUS_SRC1_COLOR);```

Using dual source blending, where src1_color = alphaMask, here the linear subpixels

[...]

PR #5930, opened 2022-03-04
Create srgb gl context by default

kchibisov:
For reference https://github.com/alacritty/alacritty/issues/4939 and
https://github.com/alacritty/alacritty/issues/3756.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Creating the IPC socket failing if WAYLAND_DISPLAY contains an absolute path
 - Crash when resetting the terminal while in vi mode
 - `font.glyph_offset` not live reloading
+- Failure when running on 10-bit color system
+- The colors being slightly different when using srgb displays on macOS
 
 ## 0.10.1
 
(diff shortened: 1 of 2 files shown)


kchibisov:
I think the major problem is that we can't check for available pixel formats before hand.

chrisduerr:
I don't remember, but didn't we do it this way around because one was supposed to be superior while the other would always work? If we put the one that always works first, the other one won't ever be triggered, right?

kchibisov:
It does get triggered on intel or amd on Windows, since it doesn't have non-srgb or so. I haven't found a reference that it was other way around. It was always like that.

chrisduerr (commented):
Does this completely fix the mentioned issues? If so we should add a changelog entry.

kchibisov:
>Does this completely fix the mentioned issues? If so we should add a changelog entry.

I think it would make sense if fix it completely and not pick better default? Though, we can add a note that we prefer srgb surfaces now.

chrisduerr:
> Though, we can add a note that we prefer srgb surfaces now.

I'm not sure people will know what that means, especially without pointing out the alternative? Should probably put it in a way the user can understand it.

kchibisov:
I've verified that it does work on macOS. The fallback is a bit nicer than the chain we had before, and should also accelerate window creation in multiwindow context, given that probing done only ones.

chrisduerr (approved):

PR #5931, opened 2022-03-04
Use round instead of ceil for underline position

kchibisov:
Ceiling underline position results in strikeout text being lower than
it should.

--

I've rechecked the original issue with the font provided there and it doesn't cause any issues.

### DIFF
--- a/alacritty/res/rect.f.glsl
+++ b/alacritty/res/rect.f.glsl
@@ -87,7 +87,7 @@ color_t draw_dotted(float_t x, float_t y) {
 
   // Since we use the entire descent area for dotted underlines, we limit its
   // height to a single pixel so we don't draw bars instead of dots.
-  float_t alpha = 1. - abs(floor(underlinePosition - 0.5) - y);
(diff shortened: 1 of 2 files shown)


chrisduerr (commented):
Do we need to ensure this doesn't cause overflow with the next cell or 0-sized underlines or stuff like that? Usually I'd expect a min/max there somewhere.

kchibisov:
It's about position, not about size, you can't make zero size underline out of it.

chrisduerr:
> It's about position, not about size, you can't make zero size underline out of it. (It was floor originally, if it matters).

Yeah but floor would ensure that it never enters another cell, right? While rounding might cause it to overlap with another cell by one pixel?

kchibisov:
>Yeah but floor would ensure that it never enters another cell, right? While rounding might cause it to overlap with another cell by one pixel?

No, since we've used `ceil` and `floor`. I don't see how it'll go farther, since it wasn't doing so with `ceil`?

chrisduerr (approved):

Issue #2539, opened 2019-06-14
Dragged filename with spaces lacks quotes


cpterry:
# System
|                  |                               |
|------------------|-------------------------------|
| Operating System | Linux     |
| Rust Version     | stable   |
| Display Server   | X11 |

# Summary
Dragging a file with spaces in the filename into Alacritty produces the filename missing quotation marks.

# Behavior
/mnt/storage/test file with spaces

# Expectation
'/mnt/storage/test file with spaces'

chrisduerr:
Duplicate of https://github.com/jwilm/alacritty/issues/1335.

Issue #1335, opened 2018-05-29
dragging\ file\ into\ terminal\ escape\ path


ghost:
https://github.com/jwilm/alacritty/pull/1304#issuecomment-392601639

spaces in the path are not escaped automatically which is a big pain, kitty does not escape them either but Mac terminal does and it makes it much easier/ faster to use

apexo:
This only makes sense when a shell is running. When a text editor (e.g. vim) is running, the current behaviour seems appropriate. However it seems kind of hard to distinguish the two cases (esp. when using tmux), so it's probably more of a question of which use case is more common … I guess the shell …

monouser7dig:
> When a text editor (e.g. vim) is running, the current behaviour seems appropriate.

nixpulvis:
I might be in favor of two different action (with two keybindings) for normal and escaped paste.

jedahan:
is there some way to trigger bracketed paste mode in tmux for this?

chrisduerr:
@zsmestad I appreciate the thought that went into this, however I don't think most of your arguments are super relevant for alacritty.

We've already built in pastejacking prevention, so there's no need for a warning dialog, the code will not automatically be executed (try it yourself with the website example).

Both escaping different patterns of text and an advanced paste dialog seems out of scope, I don't think this is of use for enough users to really provide a great value. Imo pasting files was worth implementing because it's just so incredibly simple with winit, if it would require a lot of work, it wouldn't be worth it.

But those are just my thoughts on the matter.

NickeZ:
@chrisduerr On my machine (ubuntu 18.04) it prints "evil", i.e. the attack works...

chrisduerr:
Oh yeah, apparently `bash` isn't capable of handling this. It works just fine in `zsh`. The only thing alacritty escapes is actual escape sequences right now, I'm sure this could be extended, but that would probably harm zsh users.

monouser7dig:
so after I added this to my config (Mac) and removed the duplicates at the top (maybe Alacritty could warn about duplicates some day)
```
  - { key: Left,     mods: Alt,     chars: "\x1bb"                       }
  - { key: Right,    mods: Alt,     chars: "\x1bf"                       }
```
I can jump by word left and right, if I could now drag a file that would be perfect 😆 just a setting to always escape them would go all the way for me personally and I'd still just say for most others as well...

chrisduerr:
[...]

PR #5937, opened 2022-03-07
Drag'n'drop'ed files now quoted

brainrom:
It's useful when dragging multiple files to use with the console tools.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1145,9 +1145,9 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                         self.on_focus_change(is_focused);
                     },
                     WindowEvent::DroppedFile(path) => {
-						let path: String = path.to_string_lossy().into();
-						let mut pathstr: String = "'".to_owned();
-						pathstr.push_str(&path);
+                        let path: String = path.to_string_lossy().into();
+                        let mut pathstr: String = "'".to_owned();
+                        pathstr.push_str(&path);
                         self.ctx.write_to_pty((pathstr + "' ").into_bytes());
                     },
                     WindowEvent::CursorLeft { .. } => {
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1145,8 +1145,10 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                         self.on_focus_change(is_focused);
                     },
                     WindowEvent::DroppedFile(path) => {
-                        let path: String = path.to_string_lossy().into();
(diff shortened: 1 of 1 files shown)


alacritty/src/event.rs
```diff
@@ -1146,7 +1146,9 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                     },
                     WindowEvent::DroppedFile(path) => {
                         let path: String = path.to_string_lossy().into();
-                        self.ctx.write_to_pty((path + " ").into_bytes());
+                        let mut pathstr: String = "'".to_owned();
```
kchibisov, line 1149: I'm not sure that we should quote them. But why we're not using brackated paste for dnd @chrisduerr ?

brainrom:
Quotes are need to dnd files with spaces and other special symbols in it's names. If I dnd file with spaces in name, I'll just get a two/more different command arguments. Konsole and Gnome-terminal have a such feature.

kchibisov:
I mean that with bracketed paate shell knows that you are pasting and can quote/escape everything itself. I think zsh does something like that by default.

brainrom:
I tried to enable bracketed paste and no luck. Paths still dropped without any escape/quoting. Also I realized that only bracketing isn't enough in case when filename contains ' symbol.

brainrom:
I tested with the regular paste (with Ctrl+Shift+V), it doesn't even try to escape symbols.

brainrom:
Tried the same with the gnome-terminal (as a reference for this feature), same result. So IMHO the best way is to escape and quote filenames on the terminal emulator's side, it will work for all cases.

brainrom:
I looked up gnome-terminal's source code, is uses `g_shell_quote()` to escape file's paths. It does the same, as my patch, but it escapes ' symbol too. So I'll add a ' escaping.

brainrom:
Done. Now it works perfect. Code may looks creepy, because I'm not a Rust programmer, but it works as expected.

chrisduerr (changes requested):
Why did you open a PR to quote dropped files when we **explicitly** denied this feature in the past?

alacritty/src/event.rs
chrisduerr: Just simplicity I suppose. But dropping files with escape sequences in them into the terminal would certainly be an interesting attack vector.

brainrom:
> denied this feature in the past

Where?

chrisduerr:
https://github.com/alacritty/alacritty/issues/1335 / https://github.com/alacritty/alacritty/issues/2539.

chrisduerr:
Closing since there's been no reply to my previous question in almost half a year.

PR #5941, opened 2022-03-09
Fix clippy errors with rust 1.59.0

kchibisov:
Rust 1.59.0 forces to use `unwrap_or` for closure, however the previous
version suggests `unwrap_or_else`, given that access to `len` is contant
`unwrap_or` makes sense, however fixes this clippy error results in
clippy failures with MSRV


chrisduerr (changes requested):
I don't understand, why do we care about clippy errors with Rust 1.59.0?

PR #5947, opened 2022-03-10
Use center of the pixel when aligning undercurl

kchibisov:
Since `x` position in rect shader represents left side of the pixel we
should use the center of it when dealing with contiguous functions.

### DIFF
--- a/alacritty/res/rect.f.glsl
+++ b/alacritty/res/rect.f.glsl
@@ -37,11 +37,14 @@ uniform float_t undercurlPosition;
 color_t draw_undercurl(float_t x, float_t y) {
   // We use `undercurlPosition` as an amplitude, since it's half of the descent
   // value.
-  float_t undercurl = undercurlPosition / 2. * cos(x * 2. * PI / cellWidth)
-                    + undercurlPosition - 1.;
+  //
+  // The `x` represents the left bound of pixel we should add `1/2` to it, so
+  // we compute the undercurl position for the center of the pixel.
+  float_t undercurl = undercurlPosition / 2. * cos((x + 0.5) * 2.
+                    * PI / cellWidth) + undercurlPosition - 1.;
 
-  float_t undercurlTop = undercurl + max((underlineThickness - 1.), 0.);
-  float_t undercurlBottom = undercurl - max((underlineThickness - 1.), 0.);
+  float_t undercurlTop = undercurl + max((underlineThickness - 1.), 0.) / 2.;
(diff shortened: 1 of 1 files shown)


alacritty/res/rect.f.glsl
```diff
@@ -37,8 +37,11 @@ uniform float_t undercurlPosition;
...
-  float_t undercurl = undercurlPosition / 2. * cos(x * 2. * PI / cellWidth)
-                    + undercurlPosition - 1.;
+  //
+  // The `x` represents the left bound of pixel we should add `1/2` to it, so
+  // we compute the undercurl position for the center of the pixel.
+  float_t undercurl = undercurlPosition / 2. * cos((x + 0.5) * 2.
+                    * PI / cellWidth) + undercurlPosition - 1.;
 
   float_t undercurlTop = undercurl + max((underlineThickness - 1.), 0.);
   float_t undercurlBottom = undercurl - max((underlineThickness - 1.), 0.);
```
chrisduerr, line 47: Looking at this code I'm a bit confused. I'd assume that the purpose of the `undercurl` computation is to find the center point of the undercurl?

But why is the top/bottom offset by thickness instead of thickness / 2?

kchibisov: @chrisduerr to make it a bit boolder. We've discussed it before I guess.

kchibisov: The thing is that it was originally `(undercurlThickness - 1.) / 2.` which is correct equation, however I went for even thicker version initially.

https://github.com/alacritty/alacritty/pull/5874/files

chrisduerr: Making it bolder there just doesn't seem to make a lot of sense? Should probably be in a different place?

kchibisov: I mean, where you'd put it else otherwise?

chrisduerr: I mean generally I think the shader should just render how it is told. If I specify a thickness and it renders at another that's kinda confusing. So the best solution would be inside actual Alacritty code and not shader logic imo.

The second best place would be at the beginning of the shader somewhere, where a new `underlineThickness` is calculated, something akin to `thickness *= 2` with a comment on it. That way at least you know it's not a bug but intended.

The way it is right now, it looks like it's a bug. Clearly you're not rendering at underlineThickness, so it couldn't possible be correct. It is, but someone looking at the code doesn't know.

chrisduerr (approved):

PR #5982, opened 2022-03-31
Fix OSC 4 color response format

### LINKED ISSUE #5981: palette query responses are malformed
### System

OS: Linux
Version: alacritty 0.10.0-dev (ebf41d5f)
Linux/BSD: X11/Compiz+XFCE

It appears that Alacritty has recently added support for palette interrogation. That's great! Unfortunately, the responses are malformed. They ought be of the form

`]4;N;rgb:XXXX/XXXX/XXXXST` or `]4;N;rgb:XX/XX/XXST`

instead, Alacritty is replying with

`]N;rgb:XXXX/XXXX/XXXXST`

(note the missing `4;` at the front)

not only does this diverge from other implementations, but it introduces an ambiguity: it is now impossible to distinguish between a reply for the default fg / default bg queries and palette indexes 10 and 11.

this is causing problems for Notcurses with newer Alacritty, manifesting as a bunch of garbage on stdin: https://github.com/dankamongmen/notcurses/issues/2643

chrisduerr:
The commit 60ef17e introduced support for the color query response
escape for OSC 4, however it did omit the `4;` prefix and started the
OSC with just the color index.

This patch fixes this bug and correctly responds to queries with full
OSC 4 format, including prefix plus color index.

Fixes #5981.

### DIFF
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -430,8 +430,8 @@ pub trait Handler {
     /// Set an indexed color value.
     fn set_color(&mut self, _: usize, _: Rgb) {}
 
-    /// Write a foreground/background color escape sequence with the current color.
-    fn dynamic_color_sequence(&mut self, _: u8, _: usize, _: &str) {}
+    /// Respond to a color query escape sequence.
+    fn dynamic_color_sequence(&mut self, _: String, _: usize, _: &str) {}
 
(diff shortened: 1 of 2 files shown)


dankamongmen:
i can confirm that this resolves the problem seen with Notcurses, and that i now successfully retrieve the palette from alacritty.

kchibisov (approved):

Issue #775, opened 2017-09-07
RFC: Optional Scrollbar


chetgurevitch:
@joehillen @tbodt @mattmahn

Seems there's some interest in a scrollbar so I figured I'd open an issue to keep discussion seperate from #657. I don't have a whole lot of knowledge in this area but I'm fairly certain a native implementation isn't possible, or at least that it would require a ton of changes to winit, possibly glutin and/or a new crate. Also [native window decorations are broken in wayland](https://github.com/Smithay/wayland-window/issues/19) and won't necessarily be getting fixed, so even if we did manage a native scrollbar somehow I doubt it'd work there.

Given the above I think we should work on the assumption that a scrollbar would have to be platform agnostic and implemented within alacritty.

rivertam:
Now that scrollback has been implemented, I would like to see this RFC given some more attention. Any indicator that we're currently not at the bottom of the window or that there is scrollback history one could scroll through would be really nice, I think.

I think a very early version of this feature could just be changing the window title to include little details like [Scroll: 88%] or something. That said, that could already exist and I wouldn't know it because I hide the window title with i3.

I wouldn't want to click it or anything. Just an indicator would be nice.

hook-s3c:
maybe also a thumbnail map down the side like some IDEs have

b-fuze:
This would be really nice. At the moment scrolling is taking too long and I'd like if I can just quickly scroll/skim through with a scrollbar.

nixpulvis:
I wonder sometimes how much a `ScrollToLastCommand` action would help with a lot of my long scrolling TBH.

chrisduerr:
Fyi I've looked into that and it's not possible to do that without modifying the shell.

nixpulvis:
@chrisduerr yea, I guess so... Then I guess that's a shell feature I'd want.

chrisduerr:
@nixpulvis iTerm does this and it requires both support from the shell (I believe they use some custom prompt command or something like that) and support from the shell.

So it's in no way a good solution from what I've seen. The only reason it works is because they have full control over the initial state the OS is shipped in.

nirfse:
@chrisduerr Thank you for looking into this. While this is completely out of scope of current issue, maybe it'd be a good idea for consideration to experiment on at least some custom shell, e.g. rust-based Ion Shell (https://github.com/redox-os/ion).

flying-sheep:
[...]

PR #6001, opened 2022-04-15
Show the history line indicator also when scrolled

jturcotte:
There is currently no scrollbar, but also not any kind of visual
indication when the user scrolled the buffer into the history.

There is however the numerical line indicator that is drawn when in
search mode or in vi cursor mode.
This changes also draws the indicator any time where the current
display offset is non-zero.

Related to #775


chrisduerr:
There is no desire to always have the line indicator visible when scrolled into history.

jturcotte:
Alright, thanks for the quick feedback.

PR #6002, opened 2022-04-15
Be less aggressive when copying selection

mahkoh:
1. Don't copy the selection when scrolling: The action of scrolling
   itself should not modify the primary selection. If the left/right
   mouse button is pressed while scrolling, the overall selected text
   will be copied when the button is released.

   Previously this caused the following unexpected behavior: When trying
   to paste text into alacritty, scrolling before pasting, e.g. because
   the user had scrolled up, would cause the to-be-pasted text to be
   replaced by whatever was currently selected in alacritty.

   Significantly, since the middle mouse button is usually a scroll
   wheel, it would not be unexpected for the user to accidentally scroll
   while trying to press the middle mouse button to paste text into
   alacritty.

2. Don't copy the selection upon button release unless the button is the
   left or right button.

   Previously this caused the following unexpected behavier: Pasting
   text into an alacritty window by pressing the middle mouse button
   would replace the primary selection with whatever was currently
   selected in alacritty.

### DIFF
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -614,8 +614,10 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
         let timer_id = TimerId::new(Topic::SelectionScrolling, self.ctx.window().id());
         self.ctx.scheduler_mut().unschedule(timer_id);
 
-        // Copy selection on release, to prevent flooding the display server.
-        self.ctx.copy_selection(ClipboardType::Selection);
+        if let MouseButton::Left | MouseButton::Right = button {
+            // Copy selection on release, to prevent flooding the display server.
+            self.ctx.copy_selection(ClipboardType::Selection);
+        }
     }
 
     pub fn mouse_wheel_input(&mut self, delta: MouseScrollDelta, phase: TouchPhase) {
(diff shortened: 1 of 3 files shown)


alacritty/src/input.rs
```diff
@@ -614,8 +614,10 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
         let timer_id = TimerId::new(Topic::SelectionScrolling, self.ctx.window().id());
         self.ctx.scheduler_mut().unschedule(timer_id);
 
-        // Copy selection on release, to prevent flooding the display server.
-        self.ctx.copy_selection(ClipboardType::Selection);
+        if matches!(button, MouseButton::Left | MouseButton::Right) {
```
chrisduerr, line 617: No reason to use `matches!` here, this should be an `if let` statement instead.

chrisduerr (changes requested):
Your arguments make a lot of sense, I don't see any specific reason why we'd be required to copy on scroll or on release of buttons other than left/right. So this should work just fine.

However since this is likely something people have run into, I think a changelog "fix" entry is appropriate.

chrisduerr (approved):

PR #6016, opened 2022-04-22
Update OpenBSD installation instructions

afh:
Update installation instructions on OpenBSD as [OpenBSD 7.1](http://openbsd.org/71.html), released on April 21nd 2022, provides a prebuilt Alacritty package (`curl -sL http://ftp.openbsd.org/pub/OpenBSD/7.1/packages/amd64/index.txt | grep alacritty`).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -153,7 +153,13 @@ pkg install cmake freetype2 fontconfig pkgconf python3
 
 #### OpenBSD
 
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
The `INSTALL.md` is solely for documenting how to build from source. People should be capable of finding pre-built packages themselves.

PR #5945, opened 2022-03-10
Fix crash when searching during reset

chrisduerr:
This fixes a crash which would occur when the terminal was reset while
searching, causing the vi mode cursor to be outside of the visible area.

This also fixes an issue where the search state reset would incorrectly
clamp the vi mode cursor to the grid, rather than the absolute viewport
position.

While this fix does resolve all crashes when searching while running
`cat /dev/urandom`, it does raise the question if manually clamping the
vi mode cursor in every location where it is modified is the right
choice.

I personally would recommend not applying this patch but instead
clamping the vi mode cursor when it is acquired for rendering, then
removing all the logic that tries to clamp it in other locations.

Fixes #5942.

PR #6041, opened 2022-05-04
Fix a selection range move after leaves vi-mode

a5ob7r:
This patch fixes that the right point of the selection range moves to another point when leaves vi-mode with a selection by ToggleViMode. The cause is that always moves a vi-mode cursor to a search origin whether or not the current search is active.

This problem is a regression which is introduced by #5945.

https://user-images.githubusercontent.com/12132068/166694319-e0a1c16c-ebb8-420d-b7c0-cf34b381f1b2.mp4

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -771,6 +771,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             // If we had search running when leaving Vi mode we should mark terminal fully damaged
             // to cleanup highlighted results.
             if self.search_state.dfas().is_some() {
+                self.search_state.dfas = None;
                 self.terminal.mark_fully_damaged();
             } else {
                 // Damage line indicator and Vi cursor.
@@ -781,7 +782,10 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             self.clear_selection();
         }
 
-        self.cancel_search();
+        if self.search_active() {
+            self.cancel_search();
+        }
+
         self.terminal.toggle_vi_mode();
 
         *self.dirty = true;
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -770,8 +770,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         if self.terminal.mode().contains(TermMode::VI) {
             // If we had search running when leaving Vi mode we should mark terminal fully damaged
             // to cleanup highlighted results.
-            if self.search_state.dfas().is_some() {
-                self.search_state.dfas = None;
+            if self.search_state.dfas.take().is_some() {
                 self.terminal.mark_fully_damaged();
             } else {
                 // Damage line indicator and Vi cursor.


alacritty/src/event.rs
```diff
@@ -844,10 +844,10 @@ impl<'a, N: Notify + 'a, T: EventListener> ActionContext<'a, N, T> {
...
         self.search_state.focused_match = None;
 
-        // The viewport reset logic is only needed for vi mode, since without it our origin is
-        // always at the current display offset instead of at the vi cursor position which we need
-        // to recover to.
-        if !self.terminal.mode().contains(TermMode::VI) {
+        // The viewport reset logic is only needed for vi mode with an active search, since without
+        // it the terminal has no dirty search contents and our origin is always at the current
+        // display offset instead of at the vi cursor position which we need to recover to.
+        if !(self.terminal.mode().contains(TermMode::VI) && self.search_active()) {
```
kchibisov, line 850: Could you simplify this logical expression? I think clippy should suggest using `or` instead of `and` here?

alacritty/src/event.rs
```diff
@@ -844,10 +844,10 @@ impl<'a, N: Notify + 'a, T: EventListener> ActionContext<'a, N, T> {
         // Clear focused match.
         self.search_state.focused_match = None;
 
-        // The viewport reset logic is only needed for vi mode, since without it our origin is
-        // always at the current display offset instead of at the vi cursor position which we need
-        // to recover to.
-        if !self.terminal.mode().contains(TermMode::VI) {
+        // We don't need to reset a vi-mode cursor when at least one of vi-mode and search-mode
+        // isn't active because a vi-mode cursor is invisible, or even if not so, searching can't
+        // change both the cursor position and display offset.
```
chrisduerr, line 849: Why wouldn't searching be able to change both the cursor position and display offset? That's pretty much exactly what search does when in vi mode.

chrisduerr (changes requested):

alacritty/src/event.rs
a5ob7r: It means incremental searching of search-mode doesn't happen, because search-mode isn't active.

What do you think if replaces the last part of above comment with below?

> Any vi-mode cursor position changing by incremental searching of search-mode, which we need to reset, doesn't happen.

chrisduerr: I think that this comment is basically impossible to understand.

chrisduerr: I can do it myself later.

a5ob7r: > I think that this comment is basically impossible to understand.

Really? Oh, that's sad.

> I can do it myself later.

OK, please do.

chrisduerr: Looking into this, I don't think this fix correctly encompasses the scope of this issue. Why the hell are we calling `cancel_search` when there isn't even a search active? We should never even get close to this conditional when search isn't even active.

It's probably best to either early return out of `cancel_search` or put it behind a conditional in `toggle_vi_mode`. While I haven't tested the full scope of these changes, I'd recommend you give that a try. It should probably work just as well?

chrisduerr (changes requested):

a5ob7r:
@chrisduerr Thanks for your requests. I tried both ideas, but unfortunately, both of them aren't appropriate to resolve this issue. Both of them would skip not only `search_reset_state()` but also subsequent cancel logics which clear highlights of searching when leaves vi-mode.

Instead what do you think we call `search_reset_state()` in `cancel_search()` only if vi-mode and search-mode are active? It probably makes sense that no calls a reset logic if search-mode isn't active. In the current implementation, other `cancel_search()` callers are in search-mode, so this change has no effect except vi-mode toggle.

chrisduerr:
> Instead what do you think we call search_reset_state() in cancel_search() only if vi-mode and search-mode are active? It probably makes sense that no calls a reset logic if search-mode isn't active. In the current implementation, other cancel_search() callers are in search-mode, so this change has no effect except vi-mode toggle.

I don't think we should rely on any kind of `cancel_search` behavior when search isn't active. The function either shouldn't be called, or it should be a no-op. Everything else is nonsensical. If we need the function to do something else, it should be renamed appropriate to what it does.

alacritty/src/event.rs
```diff
@@ -771,6 +771,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             // If we had search running when leaving Vi mode we should mark terminal fully damaged
             // to cleanup highlighted results.
             if self.search_state.dfas().is_some() {
+                self.search_state.dfas = None;
```
chrisduerr, line 774: This looks like you want to use `dfas().take().is_some()`?

chrisduerr (changes requested):

alacritty/src/event.rs
a5ob7r: Wow, you're right. I haven't known it. Thank you for pointing it out.

chrisduerr:
I'll test this today or tomorrow, but just to remind myself: Code is all good now. So just need to ensure it actually does what it says it does.

chrisduerr:
I'm still on this, I just got sick. That'll teach me about giving ETAs. :)

chrisduerr (approved):

PR #6062, opened 2022-05-14
Add user contributed script opacity-change.sh

RichardBronosky:
Proposal for inclusion of user contributed scripts. Even if you don't time to verify an abundance of user contributions, I think it's a good idea to give the community some direction.

You can delegate contribution review to the community. Or just let community members have a standard search to run on github (that being `alacritty/scripts/contrib`)

### DIFF
--- a/scripts/contrib/opacity-change.sh
+++ b/scripts/contrib/opacity-change.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+set -eu
+[[ -n ${DEBUG:-} ]] && set -x
+
+#### Example alacritty.yml usage
+#key_bindings:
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
> You can delegate contribution review to the community.

This idea is absolutely ludicrous. Anything in upstream Alacritty is trusted to be verified by Alacritty's maintainers. Adding random scripts that are verified by effectively nobody is not acceptable. Alacritty's upstream repo is also not a dumpster for your alacritty-related dotfiles and scripts.

If you have interest in maintaining a repository with scripts and configs related to Alacritty, please do so separately from upstream. That way users are properly informed about the choices they make when executing that software.

PR #6070, opened 2022-05-20
Disable damage tracking on X11

### LINKED ISSUE #6051: Terminal not fully redrawing under certain conditions
## System

OS: Arch Linux
Version: alacritty 0.11.0-dev (5c51d528)
Linux/BSD: xorg 30bpp (nvidia driver) and Awesome WM

## Problem

When minimizing and then unminimizing an alacritty terminal, or when switching to another window group and switching back to the original window group containing the alacritty terminal, the whole terminal is not redrawn, only the section next to the terminal cursor position is.

## Screenshots

***Correct terminal:***
correct-terminal

***Not fully redrawn terminal:***
not-fully-redrawn-terminal

## Workaround

Causing an event that triggers a full terminal redraw, such as maximizing the window, changing the font size (Ctrl+'+' or Ctrl+'-'), etc., is needed to restore full terminal content.

kchibisov:
Fixes #6051.

### DIFF
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -471,7 +471,23 @@ impl Window {
         self.windowed_context.swap_buffers_with_damage(damage).expect("swap buffes with damage");
     }
 
+    #[cfg(any(target_os = "macos", windows))]
     pub fn swap_buffers_with_damage_supported(&self) -> bool {
+        // Disable damage tracking on macOS/Windows since there's no observation of it working.
+        false
+    }
+
+    #[cfg(not(any(target_os = "macos", windows)))]
+    pub fn swap_buffers_with_damage_supported(&self) -> bool {
+        // On X11 damage tracking is behaving in unexpected ways on some NVIDIA systems. Since
+        // there's no compositor supporting it, damage tracking is disabled on X11.
+        //
+        // For more see https://github.com/alacritty/alacritty/issues/6051.
+        #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
+        if self.window().xlib_window().is_some() {
+            return false;
+        }
+
         self.windowed_context.swap_buffers_with_damage_supported()
     }
 
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -1,6 +1,7 @@
 //! ANSI Terminal Stream Parsing.
 
 use std::convert::TryFrom;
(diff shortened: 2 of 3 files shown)


chrisduerr (changes requested):
Do we really want to disable it just for X11, or is it maybe smart to also disable it for macOS/Windows right away? Are you aware of those systems supporting damage tracking?

kchibisov:
I'm not, at least not via egl.

chrisduerr:
> I'm not, at least not via egl.

So maybe just enable it for linux/bsd then?

kchibisov:
Done. I've also included some minor clippy improvements.

alacritty/src/display/window.rs
```diff
@@ -471,7 +471,24 @@ impl Window {
         self.windowed_context.swap_buffers_with_damage(damage).expect("swap buffes with damage");
     }
 
+    #[cfg(any(target_os = "macos", windows))]
     pub fn swap_buffers_with_damage_supported(&self) -> bool {
+        // Disable damage tracking on macOS/Windows since there's no observation of it working
```
chrisduerr, line 476:
```suggestion
        // Disable damage tracking on macOS/Windows since there's no observation of it working.
```

alacritty/src/display/window.rs
```diff
@@ -471,7 +471,24 @@ impl Window {
...
+        false
+    }
+
+    #[cfg(not(any(target_os = "macos", windows)))]
+    pub fn swap_buffers_with_damage_supported(&self) -> bool {
+        // On X11 damage tracking is behaving in unexpected way on some NVIDIA systems. Given that
+        // there're no compositors supporting damage tracking and there's no such thing on GLX,
+        // which is used by default, disabling it.
+        //
+        // For more see https://github.com/alacritty/alacritty/issues/6051.
```
chrisduerr, line 486:
```suggestion
        // On X11 damage tracking is behaving in unexpected ways on some NVIDIA systems. Since there's no compositor supporting it, damage tracking is disabled on X11.
        //
        // For more see https://github.com/alacritty/alacritty/issues/6051.
```

I think we can omit the GLX part, since the relevant portion would still be the compositor support. That way it reads a little nicer.

chrisduerr (changes requested):

chrisduerr (changes requested):
Should add a changelog entry to mention we fixed something.

chrisduerr (approved):

PR #6071, opened 2022-05-20
Fix builtin box-drawing glyphs for U+2567 and U+2568

algon-320:
I noticed following two characters are rendered incorrectly.

- U+2567 (BOX DRAWINGS UP SINGLE AND HORIZONTAL DOUBLE)
- U+2568 (BOX DRAWINGS UP DOUBLE AND HORIZONTAL SINGLE)

Screenshot taken in the master branch: 
master

This PR makes these characters be rendered correctly.
Screenshot taken after applying this patch: 
fix-builtin-font

`test.sh` I used in the above screenshots:
```
#!/bin/bash
echo -e 'U+2558: \u2558'
echo -e 'U+2559: \u2559'
echo -e 'U+2567: \u2567'
echo -e 'U+2568: \u2568'
```

Reference: http://www.unicode.org/charts/PDF/U2500.pdf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Vi cursor blinking not reset when navigating in search
 - Scrolling and middle-clicking modifying the primary selection
 - Bottom gap for certain builtin box drawing characters
+- Incorrect buili-in glyphs for `U+2567` and `U+2568`
 
 ## 0.10.1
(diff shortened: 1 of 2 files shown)


kchibisov (changes requested):
Would need a changelog entry, otherwise it looks good.

algon-320:
OK, I added a line to `CHANGELOG.md`.

kchibisov (approved):

kchibisov:
Thanks.

PR #6074, opened 2022-05-20
Fix Vi cursor not being damaged on scroll

kchibisov:
When scrolling doesn't actually results in scrolling in Vi mode, but
rather in changing Vi cursor position, scrolling functions were changing
the cursor position. This commit fixes it by always damaging old Vi
cursor if it was present.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -73,6 +73,20 @@ impl Default for TermMode {
     }
 }
 
+/// Convert a terminal point to a viewport relative point.
+#[inline]
+pub fn point_to_viewport(display_offset: usize, point: Point) -> Option<Point<usize>> {
+    let viewport_line = point.line.0 + display_offset as i32;
+    usize::try_from(viewport_line).ok().map(|line| Point::new(line, point.column))
+}
+
+/// Convert a viewport relative point to a terminal point.
+#[inline]
+pub fn viewport_to_point(display_offset: usize, point: Point<usize>) -> Point {
+    let line = Line(point.line as i32) - display_offset;
+    Point::new(line, point.column)
+}
+
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct LineDamageBounds {
     /// Damaged line number.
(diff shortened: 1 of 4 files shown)


alacritty_terminal/src/term/mod.rs
```diff
@@ -927,13 +936,19 @@ impl<T> Term<T> {
...
+    }
+
     /// Damage `Vi` mode cursor.
     #[inline]
     pub fn damage_vi_cursor(&mut self) {
-        let line = (self.grid.display_offset() as i32 + self.vi_mode_cursor.point.line.0)
-            .clamp(0, self.screen_lines() as i32 - 1) as usize;
-        let vi_point = Point::new(line, self.vi_mode_cursor.point.column);
-        self.damage.damage_point(vi_point);
+        self.damage_grid_point(self.vi_mode_cursor.point)
```
chrisduerr, line 951: Might not be worth it to have a method to do so at this point. Seems like that would be pretty concise just inlined.

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
```diff
@@ -373,9 +378,17 @@ impl<T> Term<T> {
         self.damage_cursor();
         self.damage.last_cursor = self.grid.cursor.point;
 
+        // Vi mode doesn't update the terminal content, thus only last vi cursor position and a
+        // new one should be damaged.
+        if let Some(last_vi_cursor_point) = self.damage.last_vi_cursor_point.take() {
+            self.damage.damage_point(last_vi_cursor_point)
+        }
```
chrisduerr, line 385:
```suggestion
        // Vi mode doesn't update the terminal content, thus only last vi cursor position and the
        // new one should be damaged.
        if let Some(last_vi_cursor_point) = self.damage.last_vi_cursor_point.take() {
            self.damage.damage_point(last_vi_cursor_point)
        }
```

alacritty_terminal/src/term/mod.rs
```diff
@@ -927,13 +936,13 @@ impl<T> Term<T> {
...
-            .clamp(0, self.screen_lines() as i32 - 1) as usize;
-        let vi_point = Point::new(line, self.vi_mode_cursor.point.column);
-        self.damage.damage_point(vi_point);
+    fn grid_point_to_viewport(&self, point: Point) -> Point<usize> {
+        let display_offset = self.grid.display_offset() as i32;
+        let lines = self.screen_lines() as i32 - 1;
+        let line = (display_offset + point.line.0).clamp(0, lines) as usize;
+
+        Point::new(line, point.column)
     }
```
chrisduerr, line 946: If this method was made public, we could probably use it in the Alacritty crate, right? I think there are a couple instances where we use something very similar. Though those return an `Option` I think, which is something that should be considered here too since technically there's nothing in this method saying the point has to be within the viewport.

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
kchibisov: I think we had such methods in the past and they always got removed.

Anyway, I've moved those methods from alacritty into alacritty_terminal. Not sure that we should add clamping to bounds...

chrisduerr (approved):

PR #6075, opened 2022-05-21
Add support for GL_EXT_blend_func_extended in GLES2 renderer

anarsoul:
GLES2 has GL_EXT_blend_func_extended extension that enables
dual-source blending, so essentially we can use fragment shader
that is very similar to GLSL3 renderer and do 1 rendering pass
instead of 3 for the text.

Tested on Pinebook with lima driver (ARM Mali 400 GPU)

### DIFF
--- a/alacritty/res/glsl3/text.v.glsl
+++ b/alacritty/res/glsl3/text.v.glsl
@@ -25,7 +25,7 @@ uniform vec4 projection;
 
 uniform int renderingPass;
 
-#define HAS_WIDE_CHAR 2
+#define WIDE_CHAR 2
 
 void main() {
     vec2 projectionOffset = projection.xy;
@@ -43,13 +43,13 @@ void main() {
     bg = backgroundColor / 255.0;
 
     float occupiedCells = 1;
-    if ((int(fg.a) >= HAS_WIDE_CHAR)) {
+    if ((int(fg.a) >= WIDE_CHAR)) {
         // Update wide char x dimension so it'll cover the following spacer.
         occupiedCells = 2;
 
         // Since we don't perform bitwise operations due to limitations of
-        // other renderers we subtract wide char bits keeping only colored.
-        fg.a = round(fg.a - HAS_WIDE_CHAR);
+        // the GLES2 renderer,we subtract wide char bits keeping only colored.
+        fg.a = round(fg.a - WIDE_CHAR);
     }
 
     if (renderingPass == 0) {
--- a/alacritty/res/glsl3/text.v.glsl
+++ b/alacritty/res/glsl3/text.v.glsl
@@ -40,11 +40,12 @@ void main() {
     vec2 cellPosition = cellDim * gridCoords;
 
     fg = vec4(textColor.rgb / 255.0, textColor.a);
+    bg = backgroundColor / 255.0;
 
-    float occupiesCells = 1;
+    float occupiedCells = 1;
     if ((int(fg.a) >= HAS_WIDE_CHAR)) {
         // Update wide char x dimension so it'll cover the following spacer.
-        occupiesCells *= 2;
+        occupiedCells = 2;
 
         // Since we don't perform bitwise operations due to limitations of
         // other renderers we subtract wide char bits keeping only colored.
@@ -53,7 +54,7 @@ void main() {
 
     if (renderingPass == 0) {
         vec2 backgroundDim = cellDim;
-        backgroundDim.x *= occupiesCells;
+        backgroundDim.x *= occupiedCells;
 
         vec2 finalPosition = cellPosition + backgroundDim * position;
         gl_Position =
@@ -73,6 +74,4 @@ void main() {
         vec2 uvSize = uv.zw;
         TexCoords = uvOffset + position * uvSize;
     }
-
-    bg = backgroundColor / 255.0;
 }
--- a/alacritty/res/glsl3/text.v.glsl
+++ b/alacritty/res/glsl3/text.v.glsl
@@ -23,9 +23,9 @@ flat out vec4 bg;
 uniform vec2 cellDim;
 uniform vec4 projection;
 
-uniform int backgroundPass;
+uniform int renderingPass;
 
-#define WIDE_CHAR 1
+#define HAS_WIDE_CHAR 2
 
 void main() {
     vec2 projectionOffset = projection.xy;
@@ -39,12 +39,22 @@ void main() {
     // Position of cell from top-left
     vec2 cellPosition = cellDim * gridCoords;
 
-    if (backgroundPass != 0) {
+    fg = vec4(textColor.rgb / 255.0, textColor.a);
+
+    float occupiesCells = 1;
+    if ((int(fg.a) >= HAS_WIDE_CHAR)) {
+        // Update wide char x dimension so it'll cover the following spacer.
+        occupiesCells *= 2;
+
+        // Since we don't perform bitwise operations due to limitations of
+        // other renderers we subtract wide char bits keeping only colored.
+        fg.a = round(fg.a - HAS_WIDE_CHAR);
+    }
+
+    if (renderingPass == 0) {
         vec2 backgroundDim = cellDim;
-        if ((int(textColor.a) & WIDE_CHAR) != 0) {
-            // Update wide char x dimension so it'll cover the following spacer.
-            backgroundDim.x *= 2;
-        }
+        backgroundDim.x *= occupiesCells;
+
         vec2 finalPosition = cellPosition + backgroundDim * position;
         gl_Position =
             vec4(projectionOffset + projectionScale * finalPosition, 0.0, 1.0);
@@ -65,5 +75,4 @@ void main() {
     }
 
     bg = backgroundColor / 255.0;
-    fg = vec4(textColor.rgb / 255.0, textColor.a);
 }
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -1,10 +1,10 @@
 use std::collections::HashSet;
 use std::ffi::CStr;
 use std::fmt;
-use std::sync::{Mutex, Once};
 
 use crossfont::Metrics;
 use log::info;
+use once_cell::sync::OnceCell;
 
 use alacritty_terminal::index::Point;
 use alacritty_terminal::term::cell::Flags;
@@ -228,34 +228,30 @@ impl GlExtensions {
     ///
     /// This function will lazyly load OpenGL extensions.
     fn contains(extension: &str) -> bool {
-        static OPENGL_EXTENSIONS: Mutex<Option<HashSet<&'static str>>> = Mutex::new(None);
-        static EXTENSIONS_LOADED: Once = Once::new();
+        static OPENGL_EXTENSIONS: OnceCell<HashSet<&'static str>> = OnceCell::new();
 
-        EXTENSIONS_LOADED.call_once(|| unsafe {
-            let extensions = gl::GetString(gl::EXTENSIONS);
-            let mut opengl_extensions = OPENGL_EXTENSIONS.lock().unwrap();
+        OPENGL_EXTENSIONS
+            .get_or_init(|| unsafe {
+                let extensions = gl::GetString(gl::EXTENSIONS);
 
-            if extensions.is_null() {
-                let mut extensions_number = 0;
-                gl::GetIntegerv(gl::NUM_EXTENSIONS, &mut extensions_number);
+                if extensions.is_null() {
+                    let mut extensions_number = 0;
+                    gl::GetIntegerv(gl::NUM_EXTENSIONS, &mut extensions_number);
 
-                *opengl_extensions = Some(
                     (0..extensions_number as gl::types::GLuint)
                         .flat_map(|i| {
                             let extension =
                                 CStr::from_ptr(gl::GetStringi(gl::EXTENSIONS, i) as *mut _);
                             extension.to_str()
                         })
-                        .collect(),
-                );
-            } else {
-                *opengl_extensions = Some(match CStr::from_ptr(extensions as *mut _).to_str() {
-                    Ok(ext) => ext.split_whitespace().collect(),
-                    Err(_) => HashSet::new(),
-                });
-            }
-        });
-
-        OPENGL_EXTENSIONS.lock().unwrap().as_ref().unwrap().contains(extension)
+                        .collect()
+                } else {
+                    match CStr::from_ptr(extensions as *mut _).to_str() {
+                        Ok(ext) => ext.split_whitespace().collect(),
+                        Err(_) => HashSet::new(),
+                    }
+                }
+            })
+            .contains(extension)
     }
 }
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -230,28 +230,30 @@ impl GlExtensions {
     fn contains(extension: &str) -> bool {
         static OPENGL_EXTENSIONS: OnceCell<HashSet<&'static str>> = OnceCell::new();
 
-        OPENGL_EXTENSIONS
-            .get_or_init(|| unsafe {
-                let extensions = gl::GetString(gl::EXTENSIONS);
+        OPENGL_EXTENSIONS.get_or_init(Self::load_extensions).contains(extension)
+    }
 
-                if extensions.is_null() {
-                    let mut extensions_number = 0;
-                    gl::GetIntegerv(gl::NUM_EXTENSIONS, &mut extensions_number);
+    /// Load available OpenGL extensions.
+    fn load_extensions() -> HashSet<&'static str> {
+        unsafe {
+            let extensions = gl::GetString(gl::EXTENSIONS);
 
-                    (0..extensions_number as gl::types::GLuint)
-                        .flat_map(|i| {
-                            let extension =
-                                CStr::from_ptr(gl::GetStringi(gl::EXTENSIONS, i) as *mut _);
-                            extension.to_str()
-                        })
-                        .collect()
-                } else {
-                    match CStr::from_ptr(extensions as *mut _).to_str() {
-                        Ok(ext) => ext.split_whitespace().collect(),
-                        Err(_) => HashSet::new(),
-                    }
+            if extensions.is_null() {
+                let mut extensions_number = 0;
+                gl::GetIntegerv(gl::NUM_EXTENSIONS, &mut extensions_number);
+
+                (0..extensions_number as gl::types::GLuint)
+                    .flat_map(|i| {
+                        let extension = CStr::from_ptr(gl::GetStringi(gl::EXTENSIONS, i) as *mut _);
+                        extension.to_str()
+                    })
+                    .collect()
+            } else {
+                match CStr::from_ptr(extensions as *mut _).to_str() {
+                    Ok(ext) => ext.split_whitespace().collect(),
+                    Err(_) => HashSet::new(),
                 }
-            })
-            .contains(extension)
+            }
+        }
     }
 }
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -230,14 +230,14 @@ impl GlExtensions {
             let extensions = gl::GetString(gl::EXTENSIONS);
             if extensions.is_null() {
                 // We're on the core profile, query extension one by one.
-                Self::from_core_profile()
+                Self::from_core_profile_unchecked()
             } else {
-                Self::from_compatible_profile()
+                Self::from_compatible_profile_unchecked()
             }
         }
     }
 
-    unsafe fn from_core_profile() -> Self {
+    unsafe fn from_core_profile_unchecked() -> Self {
         let mut extensions_number = 0;
         gl::GetIntegerv(gl::NUM_EXTENSIONS, &mut extensions_number);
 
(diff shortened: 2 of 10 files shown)


anarsoul:
I'm a pretty fresh rustacean, so please don't beat me too hard :) Any review is welcome and I'll be more than happy to address the comments.

kchibisov:
>GLES2 has GL_EXT_blend_func_extended extension that enables
dual-source blending, so essentially we can use fragment shader
that is very similar to GLSL3 renderer and do 1 rendering pass
instead of 3 for the text.

gles2 doesn't have this extension, this is gles3 thing. Also the amount of sources should be the same as in glsl3 shader.

kchibisov:
Also, if you want to make use of it you should change glsl3 renderer and its shader, no need to add new shader which is the same as glsl3 shader.

anarsoul:
> gles2 doesn't have this extension, this is gles3 thing.

See https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_blend_func_extended.txt

```
<...>
This extension is written against the OpenGL ES 3.1 (June 4, 2014)
    specification, but can apply to earlier versions back to ES 2.0.
<...>
 GLSL version 100 language is written against The OpenGL ES Shading
    Language (May 12, 2009).
<...>
```

Moreover, I'm the guy who implemented dual-source blending in lima driver, so I can tell you form my personal experience that this extension can be used with GLES2. Mali4x0 is GLES2 class GPU and it supports this extension, however I must admit that lima is the only driver for GLES2-class GPUs I know that exposes this extension.

>  Also the amount of sources should be the same as in glsl3 shader.

Fragment shader is very similar (and actually it's rewritten GLSL3 shader), but ESSL 1.0 and GLSL 3.3 are different enough, so I put it in a separate file. Note that ESSL 1.0 doesn't allow user-defined shader outputs, so EXT_blend_func_extended adds `gl_SecondaryFragColor_EXT`

anarsoul:
> Also, if you want to make use of it you should change glsl3 renderer and its shader, no need to add new shader which is the same as glsl3 shader.

Mali4x0 isn't capable of GL 3.3, but it exposes EXT_blend_func_extended with lima driver. Basically adding support for this extension to glsl3 defeats the purpose for me.

kchibisov:
>Fragment shader is very similar (and actually it's rewritten GLSL3 shader), but ESSL 1.0 and GLSL 3.3 are different enough, so I put it in a separate file. Note that ESSL 1.0 doesn't allow user-defined shader outputs, so EXT_blend_func_extended adds gl_SecondaryFragColor_EXT

You can take a look at rect shader to get inspiration how to tweak glsl3 fragment shader.

Also, do you have blend_func_extended, but not instance rendering, so you can't use glsl3 renderer?

>Mali4x0 isn't capable of GL 3.3, but it exposes EXT_blend_func_extended with lima driver. Basically adding support for this extension to glsl3 defeats the purpose for me.

What extensions lima driver has?

anarsoul:
> Also, do you have blend_func_extended, but not instance rendering, so you can't use glsl3 renderer?

That's correct

>  What extensions lima driver has?

```
GL_VERSION: OpenGL ES 2.0 Mesa 22.2.0-devel (git-53fe6f1084)
GL_RENDERER: Mali400
GL_EXTENSIONS:
    GL_EXT_blend_minmax, GL_EXT_multi_draw_arrays,
    GL_EXT_texture_format_BGRA8888, GL_OES_compressed_ETC1_RGB8_texture,
    GL_OES_depth24, GL_OES_element_index_uint, GL_OES_fbo_render_mipmap,
    GL_OES_mapbuffer, GL_OES_rgb8_rgba8, GL_OES_standard_derivatives,
    GL_OES_stencil8, GL_OES_texture_3D, GL_OES_texture_half_float,
    GL_OES_texture_half_float_linear, GL_OES_texture_npot,
    GL_OES_vertex_half_float, GL_OES_EGL_image, GL_OES_depth_texture,
    GL_OES_packed_depth_stencil, GL_OES_get_program_binary,
    GL_APPLE_texture_max_level, GL_EXT_discard_framebuffer,
    GL_EXT_read_format_bgra, GL_NV_pack_subimage, GL_EXT_frag_depth,
    GL_NV_fbo_color_attachments, GL_OES_EGL_image_external, GL_OES_EGL_sync,
    GL_OES_vertex_array_object, GL_ANGLE_pack_reverse_row_order,
    GL_EXT_texture_rg, GL_EXT_unpack_subimage, GL_NV_draw_buffers,
    GL_NV_read_buffer, GL_NV_read_depth, GL_NV_read_depth_stencil,
    GL_NV_read_stencil, GL_EXT_draw_buffers, GL_EXT_map_buffer_range,
    GL_KHR_debug, GL_KHR_texture_compression_astc_ldr,
    GL_NV_pixel_buffer_object, GL_OES_required_internalformat,
    GL_OES_surfaceless_context, GL_EXT_separate_shader_objects,
    GL_EXT_compressed_ETC1_RGB8_sub_texture, GL_EXT_draw_elements_base_vertex,
    GL_EXT_texture_border_clamp, GL_KHR_context_flush_control,
    GL_OES_draw_elements_base_vertex, GL_OES_texture_border_clamp,
    GL_EXT_blend_func_extended, GL_KHR_no_error,
    GL_KHR_texture_compression_astc_sliced_3d, GL_EXT_clip_control,
    GL_KHR_parallel_shader_compile, GL_MESA_bgra
```

anarsoul:
Here's complete glxinfo output:

[glxinfo.txt](https://github.com/alacritty/alacritty/files/8747887/glxinfo.txt)

kchibisov:
@anarsoul that makes sense, but I'd suggest to refactor glsl3 shader to be gles2 compatible the way it's done in rect renderer and its shader.

If you have troubles doing so, let me know I can try when I have time.

anarsoul:
> I'd suggest to refactor glsl3 shader to be gles2 compatible the way it's done in rect renderer and its shader.

Sure, I can do that.

anarsoul:
CI fails due to clippy error in alacritty/src/scheduler.rs which this PR doesn't touch. Should I fix it in this PR or I can just ignore it since apparently it's broken the same way in master?

kchibisov:
@anarsoul would you mind if I clean up this patch? So it'll be a bit faster, given that changes would be stylistic. And the current code looks like C.

Also, when the patch was added to lima to support blend_func_extended? When I wrote this patch there wasn't support for that on pinephone, and I think it's using the same stuff as pinebook.

anarsoul:
> would you mind if I clean up this patch? So it'll be a bit faster, given that changes would be stylistic.

Sure, go for it

> And the current code looks like C.

Yeah, it's not hard to guess why :)

> Also, when the patch was added to lima to support blend_func_extended? When I wrote this patch there wasn't support for that on pinephone, and I think it's using the same stuff as pinebook.

It's been quite a while, see https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/13873

The change should be present in mesa-22.0

kchibisov:
@anarsoul I've refactored code a bit, would you mind verifying that it still works for you? It does work on my particular machine when forcing gles2 renderer, but testing on a pure gles2 system would be nice.

anarsoul:
It works fine on my laptop with Intel graphics if I force GLES2 renderer, same on Pinebook with lima driver.

anarsoul:
@kchibisov is there anything else blocking this PR from merging?

kchibisov:
>@kchibisov is there anything else blocking this PR from merging?

It's waiting on review from @chrisduerr as you can see.

alacritty/res/glsl3/text.v.glsl
```diff
@@ -39,12 +39,22 @@ void main() {
...
+    fg = vec4(textColor.rgb / 255.0, textColor.a);
+
+    float occupiesCells = 1;
+    if ((int(fg.a) >= HAS_WIDE_CHAR)) {
+        // Update wide char x dimension so it'll cover the following spacer.
+        occupiesCells *= 2;
+
+        // Since we don't perform bitwise operations due to limitations of
+        // other renderers we subtract wide char bits keeping only colored.
+        fg.a = round(fg.a - HAS_WIDE_CHAR);
```
chrisduerr, line 51: Not sure what you mean by this, are bitwise operations not supported on gles2 with extension?

alacritty/src/renderer/mod.rs
```diff
@@ -218,3 +219,55 @@ impl Renderer {
...
+        });
+
+        GlExtensions { extensions: Box::new(extensions) }
+    }
+
+    unsafe fn from_compatible_profile() -> Self {
+        // SAFETY: OpenGL returns a pointer to a `static` string, so using static lifetime.
+        let extensions = gl::GetString(gl::EXTENSIONS);
+        let extensions: Box<dyn Iterator<Item = Result<&'static str, Utf8Error>>> =
+            match std::mem::transmute::<Result<&'_ str, Utf8Error>, Result<&'static str, Utf8Error>>(
```
chrisduerr, line 256: This looks **very** unsafe. Are you just using `transmute` to extend the lifetime here? Because you should **definitely** not do that.

alacritty/src/renderer/mod.rs
```diff
@@ -218,3 +219,55 @@ impl Renderer {
         }
     }
 }
+
+struct GlExtensions {
+    extensions: Box<dyn Iterator<Item = Result<&'static str, Utf8Error>>>,
```
chrisduerr, line 224: A `Box<Iterator>` as storage on a struct seems somewhat unconventional. At that point why not just use a `Vec`?

alacritty/src/renderer/text/gles2.rs
```diff
@@ -444,17 +460,24 @@ pub struct TextShaderProgram {
 
     /// Rendering pass.
     ///
-    /// The rendering is split into 4 passes. One is used for the background and the rest to
-    /// perform subpixel text rendering according to
+    /// For dual source blending, there're 2 passes, one for background, another for text,
+    /// similar to GLES3 renderer.
```
chrisduerr, line 464:
```suggestion
    /// similar to GLES3 the renderer.
```

alacritty/src/renderer/text/gles2.rs
```diff
@@ -444,17 +460,24 @@ pub struct TextShaderProgram {
...
     /// Rendering is split into three passes.
     u_rendering_pass: GLint,
 }
 
 impl TextShaderProgram {
-    pub fn new(shader_version: ShaderVersion) -> Result<Self, Error> {
-        let program = ShaderProgram::new(shader_version, TEXT_SHADER_V, TEXT_SHADER_F)?;
+    pub fn new(shader_version: ShaderVersion, dual_source_blending: bool) -> Result<Self, Error> {
+        let fragment_shader =
+            if dual_source_blending { &super::glsl3::TEXT_SHADER_F } else { &TEXT_SHADER_F };
```
chrisduerr, line 477: Should use imports and fully qualified paths. `super::` just gets confusing.

chrisduerr (changes requested):
Tested on my PinePhone and it seems to work without any issues.

alacritty/src/renderer/text/mod.rs
```diff
@@ -24,11 +24,32 @@ use glyph_cache::{Glyph, LoadGlyph};
 bitflags! {
     #[repr(C)]
     struct RenderingGlyphFlags: u8 {
-        const WIDE_CHAR = 0b0000_0001;
-        const COLORED   = 0b0000_0010;
+        const COLORED   = 0b0000_0001;
+        const WIDE_CHAR = 0b0000_0010;
```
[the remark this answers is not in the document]
kchibisov: To simplify shaders a bit, since we subtract due to lack of bitwise operations on gles2.

alacritty/res/glsl3/text.v.glsl
```diff
@@ -23,9 +23,9 @@ flat out vec4 bg;
 uniform vec2 cellDim;
 uniform vec4 projection;
 
-uniform int backgroundPass;
+uniform int renderingPass;
 
-#define WIDE_CHAR 1
+#define HAS_WIDE_CHAR 2
```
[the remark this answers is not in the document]
kchibisov: I thought that it's a bit clearer, since it's not a bit flag, but rather a packed sum. So 3 and 2 will mean that they have wide char.

alacritty/res/glsl3/text.v.glsl
```diff
@@ -65,5 +75,4 @@ void main() {
     }
 
     bg = backgroundColor / 255.0;
```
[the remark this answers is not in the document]
kchibisov: I guess.

alacritty/res/glsl3/text.v.glsl
```diff
@@ -39,12 +39,22 @@ void main() {
     // Position of cell from top-left
     vec2 cellPosition = cellDim * gridCoords;
 
-    if (backgroundPass != 0) {
+    fg = vec4(textColor.rgb / 255.0, textColor.a);
+
+    float occupiesCells = 1;
+    if ((int(fg.a) >= HAS_WIDE_CHAR)) {
+        // Update wide char x dimension so it'll cover the following spacer.
+        occupiesCells *= 2;
```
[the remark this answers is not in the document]
kchibisov: Ah, yeah. I refactored it a bit weirdly, since it wasn't a multiplier before.

alacritty/res/glsl3/text.v.glsl
kchibisov: >bitwise operations not supported on gles2 with extension

gles2 shaders don't have bitwise operations at all. It doesn't matter if you have extension or not.

alacritty/src/renderer/mod.rs
kchibisov: This is safe, since the pointer from C library is pointing to static string by the opengl specification. So the reference will leave to the end of the program.

I'm just using what specification is tell me, if it wasn't safe I wouldn't extended it in the first place?

What would you suggest? To copy everything?

alacritty/src/renderer/mod.rs
kchibisov: Because we don't collect? It's another iterator that handled 2 sources of extensions. it's purely for efficiency reasons, since we have 2 iterators and they both read static strings from some static storage, so having iterator like this won't allocate anything for elements used in the iterator.

By using `Vec` we'll have to collect and store everything, and there won't be a use for iterator at all. Though, we can likely refactor to some lazy initialized `HashSet<CStr>`.

alacritty/src/renderer/mod.rs
kchibisov: Though, I think I don't really need transmute. I'll try without it, but the thing is that all the references are static here anyway.

alacritty/res/glsl3/text.v.glsl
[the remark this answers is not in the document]
chrisduerr: I'm mostly confused by the `HAS_` prefix tbh. I'd expect boolean prefix to either be `IS_` or nothing at all.

alacritty/src/renderer/mod.rs
chrisduerr: > So the reference will leave to the end of the program.

Is that true for the Rust reference too though? Generally speaking, it seems like a **really** bad idea to extend an existing lifetime with `transmute`. Because even if the compiler doesn't use lifetime information for optimizations, at the very least it would indicate that the original lifetime is incorrect.

kchibisov: Well there's no transmute anymore as you can see. The code is marked as `safe`...

alacritty/res/glsl3/text.v.glsl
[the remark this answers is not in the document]
kchibisov: That's intentional since it's not about identity, but about having, since it's a flag and not a property. So it could have `WIDE_CHAR` and `COLOR`, and it is wide char when the flag is higher than 2. For color it's colored when it's `1` or `3`, so you can't really check for `IS`, it's more about `has`.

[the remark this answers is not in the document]
chrisduerr: `has_widechar` would be more of a function name in that scenario though. For bitflags you'd name the flag itself `WIDE_CHAR` and then check if `flags.contains(WIDE_CHAR)`.

alacritty/res/glsl3/text.v.glsl
chrisduerr: Could you state this explicitly rather than just mentioning "other renderers"? It's not exactly clear what that means here.

alacritty/src/renderer/text/mod.rs
```diff
@@ -24,11 +24,27 @@ use glyph_cache::{Glyph, LoadGlyph};
...
     }
 }
 
+// NOTE: the passes are shared between GLES2 and GLSL3 renderers.
```
chrisduerr, line 32:
```suggestion
/// Rendering passes, for both GLES2 and GLSL3 renderer.
```

alacritty/src/renderer/text/glsl3.rs
```diff
@@ -15,12 +15,12 @@ use crate::renderer::{cstr, Error};
...
 use super::{
-    Glyph, LoadGlyph, LoaderApi, RenderingGlyphFlags, TextRenderApi, TextRenderBatch, TextRenderer,
-    TextShader,
+    Glyph, LoadGlyph, LoaderApi, RenderingGlyphFlags, RenderingPass, TextRenderApi,
+    TextRenderBatch, TextRenderer, TextShader,
 };
 
 // Shader source.
-static TEXT_SHADER_F: &str = include_str!("../../../res/glsl3/text.f.glsl");
+pub(super) static TEXT_SHADER_F: &str = include_str!("../../../res/glsl3/text.f.glsl");
```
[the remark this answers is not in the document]
chrisduerr: This has not been resolved.

alacritty/src/renderer/mod.rs
```diff
@@ -218,3 +220,48 @@ impl Renderer {
...
+                        .collect(),
+                );
+
+                return;
+            }
+
+            OPENGL_EXTENSIONS = match CStr::from_ptr(extensions as *mut _).to_str() {
+                Ok(ext) => Some(ext.split_whitespace().collect()),
+                Err(_) => Default::default(),
+            };
```
chrisduerr, line 257: I think I'd prefer an `else` over an early return here.

alacritty/src/renderer/mod.rs
```diff
@@ -218,3 +220,48 @@ impl Renderer {
...
+
+struct GlExtensions;
+
+impl GlExtensions {
+    /// Check if the given `extension` is supported.
+    ///
+    /// This function will lazyly load OpenGL extensions.
+    fn contains(extension: &str) -> bool {
+        static mut OPENGL_EXTENSIONS: Option<HashSet<&'static str>> = None;
+        static EXTENSIONS_LOADED: Once = Once::new();
```
chrisduerr, line 232: Global static state isn't super common in Rust code usually, but honestly for this kinda stuff I think it makes things pretty convenient. Especially because the underlying OpenGL structures are all static anyway.

alacritty/src/renderer/mod.rs
```diff
@@ -218,3 +220,48 @@ impl Renderer {
...
+                return;
+            }
+
+            OPENGL_EXTENSIONS = match CStr::from_ptr(extensions as *mut _).to_str() {
+                Ok(ext) => Some(ext.split_whitespace().collect()),
+                Err(_) => Default::default(),
+            };
+        });
+
+        unsafe {
```
chrisduerr, line 260: Unfortunate we have to use unsafe here, but I don't think there's any way around this? There might be some crates that manage to hide that unsafe away, but with STD I don't recall a way. Did you check?

alacritty/src/renderer/text/gles2.rs
```diff
@@ -444,17 +458,24 @@ pub struct TextShaderProgram {
 
     /// Rendering pass.
     ///
-    /// The rendering is split into 4 passes. One is used for the background and the rest to
-    /// perform subpixel text rendering according to
+    /// For dual source blending, there're 2 passes, one for background, another for text,
```
chrisduerr, line 461: I don't think it's gramatically incorrect, but "there're" is a bit awkward. So I'd suggest changing it:

```suggestion
    /// For dual source blending, there are 2 passes; one for background, another for text,
```

alacritty/src/renderer/text/gles2.rs
```diff
@@ -375,19 +392,25 @@ impl<'a> TextRenderApi<Batch> for RenderApi<'a> {
...
+                // First text rendering pass.
+                gl::BlendFuncSeparate(gl::ZERO, gl::ONE_MINUS_SRC_COLOR, gl::ZERO, gl::ONE);
+                gl::DrawElements(gl::TRIANGLES, num_indices, gl::UNSIGNED_SHORT, ptr::null());
+
+                // Second text rendering pass.
+                self.program.set_rendering_pass(RenderingPass::SubpixelPass2);
+                gl::BlendFuncSeparate(gl::ONE_MINUS_DST_ALPHA, gl::ONE, gl::ZERO, gl::ONE);
+                gl::DrawElements(gl::TRIANGLES, num_indices, gl::UNSIGNED_SHORT, ptr::null());
+
+                // Third pass.
```
chrisduerr, line 409: Wasn't changed in this PR, but it's kinda strange that we include "text rendering" twice then leave it out for the third. Might as well make it consistent so people don't think the third pass is for something else.

chrisduerr (changes requested):

alacritty/src/renderer/mod.rs
kchibisov: Yeah, the safe way is to put into Mutex.

kchibisov: I'd go with `OnceCell`, since it's a part of our Cargo.lock anyway and will go into rust stdlib eventually.

alacritty/src/renderer/mod.rs
kchibisov: Yeah, but we want a singleton here, since we expose global static immutable object.

alacritty/src/renderer/mod.rs
chrisduerr: That's what I was thinking of when talking about crates, thanks!

alacritty/src/renderer/mod.rs
```diff
@@ -218,3 +220,38 @@ impl Renderer {
...
+impl GlExtensions {
+    /// Check if the given `extension` is supported.
+    ///
+    /// This function will lazyly load OpenGL extensions.
+    fn contains(extension: &str) -> bool {
+        static OPENGL_EXTENSIONS: OnceCell<HashSet<&'static str>> = OnceCell::new();
+
+        OPENGL_EXTENSIONS
+            .get_or_init(|| unsafe {
+                let extensions = gl::GetString(gl::EXTENSIONS);
```
chrisduerr, line 235: Moving this entire closure into a separate function would probably clean things up a lot visually.

chrisduerr (approved):
LGTM

alacritty/src/renderer/mod.rs
chrisduerr: Decided to just do this myself real quick.

kchibisov (approved):
Looks good. I've also tested on macOS to ensure we're not breaking it.

PR #6079, opened 2022-05-23
Bump glutin to 0.29.0

### LINKED ISSUE #5767: Powering off screns results in inconsistent DPI/font-size
Powering off a screen results in inconsistent font/DPI size for all alacritty instances on that screen until the screen resolution is manually changed, including newly spawned alacritty instances.

It appears it gets confused about the `Device pixel ratio:`

Before power off, and expected result after a power on/off cycle:
Note how both alacritty and urxvt have the same, correct, font size.
Screenshot_2022-01-11_07-15-51

After a power cycle:
Notice how only urxvt has the correct font size.
Screenshot_2022-01-11_07-15-26

(the rendered image size only differs in the first screenshot due to neofetch drawing them as a percentage of terminal height, the second one naturally being smaller due to being spawned after and being less tall due to it)

### System

OS: Linux
Version: alacritty 0.9.0 (fed349aa)
Linux/BSD: X11, bspwm, nvidia

### Logs

before and during power off: [alacritty.log.1.txt](https://github.com/alacritty/alacritty/files/7844191/alacritty.log.1.txt)
after power off to demonstrate that the pixel ratio is different: [alacritty.log.2.txt](https://github.com/alacritty/alacritty/files/7844192/alacritty.log.2.txt)

Keyboard and bindings: `alacritty --print-events`
[alacritty_events.log.txt](https://github.com/alacritty/alacritty/files/7844245/alacritty_events.log.txt)

### LINKED ISSUE #5975: Reopening Terminal
When you exit the terminal in after running commands a dot it still under it. If you want to reopen the terminal you first have to quit the program, then click again to open up a terminal.

### System

OS: macOS
Version: alacritty 0.10.1 (2844606)

### Logs

Crashes: STDERR, STDOUT
Keyboard and bindings: `alacritty --print-events`

kchibisov:
Fixes #5975.
Fixes #5767.


kchibisov:
@sshock could you verify that tab issue you've mentioned in winit https://github.com/rust-windowing/winit/issues/2238 gets resolved with this patch? I'm not sure what to look for myself on macOS for that...

CHANGELOG.md
```diff
@@ -19,13 +19,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- `WINIT_WAYLAND_CSD_THEME` env variable to control decorations "dark"/"light" variant
+- Support for inline input method
 
 ### Changed
 
 - The `--help` output was reworked with a new colorful syntax
 - OSC 52 is now disabled on unfocused windows
 - `SpawnNewInstance` no longer inherits initial `--command`
+- Client side decorations should have proper text rendering now on Wayland
+- Search bar is now respecting cursor thickness
```
chrisduerr, line 32: Not sure this should be fixed in this PR?

CHANGELOG.md
```diff
@@ -37,9 +43,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Vi cursor blinking not reset when navigating in search
 - Scrolling and middle-clicking modifying the primary selection
 - Bottom gap for certain builtin box drawing characters
-- Incorrect built-in glyphs for `U+2567` and `U+2568`
 - Character mappings in the DEC special graphics character set (line drawing)
 - Window flickering on resize on Wayland
+- Incorrect built-in glyphs for `U+2567` and `U+2568`
+- Cursor not hiding on GNOME Wayland
+- Font having different scale factor after monitor powering off/on on X11
+- Wrong mouse input when opening a new tabbed window on macOS
```
chrisduerr, line 51: Rather than describing the mouse part here, I'd instead mention that the entire window content was offset. It affected rendering too, even with no mouse usage.

alacritty.yml
```diff
@@ -22,6 +22,20 @@
...
   # available, otherwise `xterm-256color` is used.
   #TERM: alacritty
+  #
+  # WINIT_X11_SCALE_FACTOR
+  #
+  # This value is used to control scale factor used by terminal on X11, by
+  # default it automatically detects scale factor of each monitor and apply
+  # it dynamically by using XRandr. Supported values `randr` or floating point
+  # values, like `"1.0"`.
+  #WINIT_X11_SCALE_FACTOR: randr
```
chrisduerr, line 32: Why did you add this? I really don't want people looking at the config file and thinking it would be a good idea to change this to tweak their scale factor.

alacritty.yml
```diff
@@ -22,6 +22,20 @@
...
+  #
+  # WINIT_WAYLAND_CSD_THEME
+  #
+  # This value is used to pick between Wayland client side decorations variant.
+  # For now only `dark` and `light` are supported.
+  #WINIT_WAYLAND_CSD_THEME: light
```
chrisduerr, line 38: Should we maybe translate this to an Alacritty option if it's relevant enough to show up in our configuration file?

alacritty/build.rs
```diff
@@ -20,7 +20,7 @@ fn main() {
         .unwrap();
 
     #[cfg(windows)]
-    embed_resource::compile("./windows/windows.rc");
+    embed_resource::compile("./windows/alacritty-icon.rc");
```
chrisduerr, line 23: Why did you change that here?

chrisduerr (changes requested):

CHANGELOG.md
kchibisov: It's a part of IME commit.

alacritty.yml
kchibisov: I wasn't sure how to put it and how it should be discovered or not. Since we have certain values and I wasn't sure how to translate them.

alacritty.yml
kchibisov: As an option for config reload? I think we can do that, just not sure that you'd want to.

alacritty/build.rs
kchibisov: The build fails since window-sys creates some lib with `windows` in it's name. So we should have a different name here.

sshock:
> @sshock could you verify that tab issue you've mentioned in winit [rust-windowing/winit#2238](https://github.com/rust-windowing/winit/issues/2238) gets resolved with this patch? I'm not sure what to look for myself on macOS for that...

Yep, it's working!

alacritty/build.rs
chrisduerr: Then I'd suggest `alacritty`?

kchibisov: I think it'll conflict with the lib from the binary, but we can try...

kchibisov: Seems like it's working.

alacritty.yml
kchibisov: I've decided to unify it a bit.

alacritty.yml
kchibisov: I think I'll remove env section bits.

alacritty/src/event.rs
```diff
@@ -1127,6 +1142,44 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
...
+                        // This text is being composed, thus should be shown as inline input to
+                        // the user.
+                        Ime::Preedit(text, cursor_range) => {
+                            // If've got empty preedit we should start processing ordinary keyboard
+                            // input.
+                            if text.is_empty() && cursor_range.is_none() {
+                                self.ctx.display.ime_input.clear_preedit()
+                            } else {
+                                self.ctx.display.ime_input.set_preedit(text, cursor_range);
+                            }
```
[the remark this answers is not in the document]
kchibisov: What platform are you on?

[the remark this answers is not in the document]
jayong93: Oh, I forgot that. I'm on macOS Monterey with M1 Mac.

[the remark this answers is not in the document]
jayong93: I've found one more bug. After I insert one or more korean character, ascii characters (like '0' or '+', etc.) are inserted two per key press.

I guess it is because when `ime_input` is enabled, ascii characters are processed both in `WindowEvent::Ime` and `WindowEvent::KeyboardInput`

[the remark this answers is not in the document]
kchibisov: I've sent a fix upstream, I'll let you know once it merged and get resolved here.

kchibisov:
@chrisduerr I've removed IME commit for now and implemented it like we had before. Will rewrite it and send separately. 

Note, that I have bumped msrv to 1.61 for now. Probably will sort that stuff out tomorrow, since it'll require a winit bump to downgrade it back to 1.57.

alacritty/Cargo.toml
```diff
@@ -25,11 +25,11 @@ fnv = "1"
 serde = { version = "1", features = ["derive"] }
 serde_yaml = "0.8"
 serde_json = "1"
-glutin = { version = "0.28.0", default-features = false, features = ["serde"] }
+glutin = { git = "https://github.com/kchibisov/glutin", branch = "release-029", default-features = false, features = ["serde"] }
```
kchibisov, line 28: I'll release it on demand.

kchibisov: glutin was released.

alacritty/src/event.rs
```diff
@@ -1310,6 +1333,28 @@ impl Processor {
             }
 
             match event {
+                // The event loop just got initialized. Create a window.
+                GlutinEvent::NewEvents(StartCause::Init) => {
+                    // Creating window inside event loop is required for platforms like macOS to
+                    // properly initialize state, like tab management infra.
+                    //
+                    // Othwerwise the first window won't handle tabs.
```
chrisduerr, line 1341:
```suggestion
                    // Creating window inside event loop is required for platforms like macOS to
                    // properly initialize state, like tab management. Othwerwise the first window won't handle tabs.
```

alacritty/src/main.rs
```diff
@@ -62,6 +61,10 @@ use crate::event::{Event, Processor};
 #[cfg(target_os = "macos")]
 use crate::macos::locale;
 
+// TODO refactor to use `ExitCode` when bumping minimum supported rust version to 1.61.
+pub const EXIT_FAILURE: i32 = 1;
+pub const EXIT_SUCCESS: i32 = 0;
```
chrisduerr, line 66: I don't think these constants provide any benefit at all in the current code.

alacritty/src/main.rs
```diff
@@ -79,18 +82,23 @@ fn main() {
...
-        eprintln!("Error: {}", err);
-        process::exit(1);
-    }
+    let exit_code = match result {
+        Ok(code) => code,
+        Err(err) => {
+            eprintln!("Error: {}", err);
+            EXIT_FAILURE
+        },
+    };
```
chrisduerr, line 99: Why bother with exit codes and all that stuff when `Err(err)` is mapped to an exit code anyway? Just so we don't print the error?

chrisduerr (changes requested):

alacritty/src/main.rs
kchibisov: The idea was to propagate the error code from the event loop somehow, since now `run_return` will return non zero code in case of failures.

I basically wanted to indicate an error with non zero exit status nothing more. If you have an idea on how to handle that more reliably let me know.

alacritty/src/main.rs
kchibisov: See the comment wrt `run_return`.

alacritty/src/main.rs
chrisduerr: Why not return the actual error? That seems to be the sensible choice to me.

chrisduerr (changes requested):
See my comment about returning a Result.

alacritty/src/main.rs
```diff
@@ -62,7 +62,7 @@ use crate::event::{Event, Processor};
 #[cfg(target_os = "macos")]
 use crate::macos::locale;
 
-fn main() {
+fn main() -> Result<(), Box<dyn Error>> {
```
chrisduerr, line 65: I've only ever gotten extremely poor results from doing this. I don't think it's a great idea?

alacritty/src/event.rs
```diff
@@ -1310,6 +1337,27 @@ impl Processor {
...
+                        None => return,
+                    };
+
+                    if let Err(result) =
+                        self.create_window(event_loop, proxy.clone(), initial_window_options)
+                    {
+                        // Stash the error and exit with non-zero code.
+                        *loop_error.lock().unwrap() = result.into();
+                        *control_flow = ControlFlow::ExitWithCode(1);
+                        return;
```
chrisduerr, line 1356: Wait can we not just exit with any generic result, which is then returned from the closure?! That's kinda garbage.

Let's maybe just print the error directly inside the `run_return`, then exit upwards with just any random error text?

I **definitely** don't want to throw mutexes at a silly one-off error that can only happen right before the event loop crashes anyway.

chrisduerr (changes requested):

alacritty/src/event.rs
kchibisov: We can sure.

alacritty/src/main.rs
kchibisov: I think that's fine, like it'll print error? At least that's how I've parsed your lets return `Result`, but now I see that you wanted to return it from `run_return`, but you can't do that.

I'm not sure that this is any different than then `println` and then `exit` that what it should do anyway, no?

chrisduerr: Hm, I think my main issue with last time I tried it is that you don't get any backtrace by doing this. Which should be fine in our case since we were just using `println` anyway.

chrisduerr (approved):
I think it would be nice to have the ability to return from winit's `run_return` with any type parameter, but this should be fine for now.

PR #6085, opened 2022-05-25
fix indent yaml

lufeee:
Live config reload and Shell config need to indent.
My environment need to indent once.


kchibisov:
This is a valid yaml with valid indention in a way that removing hash would result in valid config option. `live_config_reload` should have **0** indention. And shell being a top level entry have zero indention as well.

So something is wrong with either your environment or the way you uncomment things.

Issue #6095, opened 2022-05-31
DEC Special Character Set (Line drawing)


kugland:
On commit 1dacc9918 the current table of the DEC Special Character Set has been introduced.

The current values deviate from the ones in XTerm. Gnome-terminal and derivatives, and also KDE’s Konsole follow XTerm in this.

| Character | Alactritty | XTerm |
|-------|-------|-------|
| `b`     | `'\t'` (Tab character) | `U+2409` (Symbol for horizontal tabulation) |
| `c`     | `'\u{000c}'` (Form feed) | `U+240C` (Symbol for form feed) |
| `d`    | `'\r'` (Carriage return) | `U+240D` (Symbol for carriage return) |
| `e`    | `'\n'` (Line feed) | `U+240A` (Symbol for line feed) |
| `i`     | `'\u{000b}'` (Vertical tabulation) | `U+240B` (Symbol for vertical tabulation) |

On Alacritty:
![alacritty](https://user-images.githubusercontent.com/1173932/171280810-1c2f3a25-e1b6-4cd5-900d-1477d8b6b298.png)

On XTerm:
![xterm](https://user-images.githubusercontent.com/1173932/171280660-3ec23574-a6df-414c-a65e-e390b4c6c6fc.png)

chrisduerr:
Good catch, thanks. This bug is actually far older than you suggested: 62294771905917087c484bfc57e4f6ca547d89ba

I suppose that goes to show how rarely these parts of the character set are used.

PR #6096, opened 2022-05-31
Fix for DEC Special Character Set (Line drawing)

kugland:
Fix for #6095.


chrisduerr (changes requested):
Change looks good, could you fix the tests too? I think the easiest solution is probably to just edit the test files manually.

Also for reference: https://vt100.net/docs/vt220-rm/chapter2.html#T2-4

kugland:
> Change looks good, could you fix the tests too? I think the easiest solution is probably to just edit the test files manually.
> 
> Also for reference: https://vt100.net/docs/vt220-rm/chapter2.html#T2-4

I just fixed the tests.

j4james:
While you're in this area, note that this character set is also missing the mapping of `_` to blank. You can see the correct behavior in the XTerm screenshot posted in the original issue (#6095). It's also documented in the VT220 manual linked above.

kugland:
> While you're in this area, note that this character set is also missing the mapping of `_` to blank. You can see the correct behavior in the XTerm screenshot posted in the original issue (#6095). It's also documented in the VT220 manual linked above.

Oh, I didn’t catch this! Now it’s fixed.

kugland:
> While you're in this area, note that this character set is also missing the mapping of `_` to blank. You can see the correct behavior in the XTerm screenshot posted in the original issue (#6095). It's also documented in the VT220 manual linked above.

Same problem as before: this change made the tests fail. I fixed the tests now.

kugland:
With this last commit, the result in Alacritty is the same as in XTerm:

image

chrisduerr (approved):
Thanks for the PR.

I've taken the liberty to add a changelog entry myself and everything should be good to go now.

kugland:
> I've taken the liberty to add a changelog entry myself and everything should be good to go now.

@chrisduerr , in the line above the one you added, ‘built-in’ is written ‘buili-in.’

PR #6105, opened 2022-06-02
Derive Eq for the PartialEq-deriving types

nyurik:
Making Clippy happy by following its [suggestion](https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq) to add Eq trait to some structs and enums.


chrisduerr (changes requested):
Please don't unnecessarily split things into multiple PRs.

nyurik:
@chrisduerr I did this on purpose here because `Eq` could be considered a controversial and unnecessary change, so most projects actually prefer this type of split. I'll be happy to merge it into the #6104

PR #6106, opened 2022-06-02
[noop] Fmt: order trait impl members same as in trait

nyurik:
This is purely a formatting change:  for consistency, order all trait implementations to be in the same order as declared in the trait itself.

Note that trait type definitions are still orderered according to `fmt` rules.


chrisduerr (changes requested):
Pointless change that would just cause completely unnecessary review work.

PR #6109, opened 2022-06-04
Fix flickering on resize on Wayland

kchibisov:


alacritty/src/display/mod.rs
```diff
@@ -622,6 +619,10 @@ impl Display {
         // Resize renderer.
         let physical = PhysicalSize::new(self.size_info.width() as _, self.size_info.height() as _);
         self.window.resize(physical);
+
+        // Ensure we're modifying the correct OpenGL context.
+        self.window.make_current();
```
mahkoh, line 624: Have you checked that there are no opengl calls above this line? When I experimented with this originally, I also moved the code related to the glyph cache below this line.

kchibisov: Hm, yeah, you're right. there's a case when loading fonts, so we should move it past it as well.

kchibisov: I've moved it above other calls, so should be fine.

kchibisov:
@mahkoh would you mind trying it one more time, I think it should work now. At least I don't see any flickering on resize at all anymore.

kchibisov:
Note, this requires testing on other platfroms like macOS before merging, since I'm not sure we can resize before making current on those. As well we should test on X11.

mahkoh:
This fixes the issue for me.

chrisduerr:
> Note, this requires testing on other platfroms like macOS before merging, since I'm not sure we can resize before making current on those. As well we should test on X11.

Can you test on macOS @kchibisov? I have no problem with testing on X11 obviously but I can't test macOS.

kchibisov:
>Can you test on macOS @kchibisov? I have no problem with testing on X11 obviously but I can't test macOS.

Sure, I'll test when I'll get back to my mac. Will acknowledge you afterwards.

Though, the changes are required for Wayland, so they'll stay like that, I could just move a callback for making context current just for macOS, etc.

alacritty/src/renderer/text/glyph_cache.rs
```diff
@@ -267,16 +267,18 @@ impl GlyphCache {
...
-        self.cache = HashMap::default();
+        self.cache = Default::default();
 
         self.load_common_glyphs(loader);
     }
 
-    pub fn update_font_size<L: LoadGlyph>(
+    /// Update the inner font size.
+    ///
+    /// NOTE: to reload renderers font [[`Self::clear_glyph_cache`]] should be called afterwards.
```
chrisduerr, line 277:
```suggestion
    /// NOTE: To reload the renderer's fonts [[`Self::clear_glyph_cache`]] should be called afterwards.
```

alacritty/src/window_context.rs
```diff
@@ -324,6 +324,9 @@ impl WindowContext {
             return;
         }
 
+        // Force the display to process any pending display update.
+        self.display.process_renderer_update();
```
chrisduerr, line 328: This should always be done even if `!self.dirty`? Wouldn't that cause the back buffer to be locked?

alacritty/src/display/mod.rs
```diff
@@ -367,6 +370,15 @@ pub struct Display {
     meter: Meter,
 }
 
+#[derive(Debug, Default, Copy, Clone)]
```
chrisduerr, line 373:
```suggestion
/// Pending renderer updates.
///
/// All renderer updates are cached to be applied just before rendering, to avoid platform-specific rendering issues.
#[derive(Debug, Default, Copy, Clone)]
```

alacritty/src/display/mod.rs
```diff
@@ -425,15 +437,15 @@ impl Display {
...
+            Self::update_font_size(&mut glyph_cache, scale_factor, config, font)
         } else {
             (cell_width, cell_height)
         };
 
         // Load font common glyphs to accelerate rendering.
         debug!("Filling glyph cache with common glyphs");
         renderer.with_loader(|mut api| {
-            glyph_cache.load_common_glyphs(&mut api);
+            glyph_cache.clear_glyph_cache(&mut api);
```
chrisduerr, line 448: I feel like the `clear_glyph_cache` function has a pretty bad name. I actually had to look it up just to make sure it actually loads the common glyphs.

Maybe something like `reset_glpyh_cache` would be more appropriate? Just to indicate it is reset to a specific state, not emptied?

alacritty/src/display/mod.rs
```diff
@@ -568,31 +582,31 @@ impl Display {
         let (mut cell_width, mut cell_height) =
             (self.size_info.cell_width(), self.size_info.cell_height());
 
-        // Ensure we're modifying the correct OpenGL context.
-        self.window.make_current();
+        if pending_update.font().is_some() || pending_update.cursor_dirty() {
+            let pending_renderer_update =
+                self.pending_renderer_update.get_or_insert(Default::default());
+            pending_renderer_update.clear_font_cache = true
```
chrisduerr, line 588:
```suggestion
            let renderer_update = self.pending_renderer_update.get_or_insert(Default::default());
            renderer_update.clear_font_cache = true
```

Just a super short temporary variable, should be fine to use a little bit of a shorthand.

alacritty/src/display/mod.rs
```diff
@@ -568,31 +582,31 @@ impl Display {
...
         }
 
         let (mut width, mut height) = (self.size_info.width(), self.size_info.height());
         if let Some(dimensions) = pending_update.dimensions() {
             width = dimensions.width as f32;
             height = dimensions.height as f32;
+
+            let pending_renderer_update =
+                self.pending_renderer_update.get_or_insert(Default::default());
+            pending_renderer_update.resize = true
```
chrisduerr, line 609:
```suggestion
            let renderer_update = self.pending_renderer_update.get_or_insert(Default::default());
            renderer_update.resize = true
```

alacritty/src/display/mod.rs
```diff
@@ -618,10 +632,36 @@ impl Display {
 
         // Resize terminal.
         terminal.resize(self.size_info);
+    }
+
+    /// Update the state of the actual renderer.
```
chrisduerr, line 637:
```suggestion
    /// Update the state of the renderer.
```

alacritty/src/display/mod.rs
```diff
@@ -618,10 +632,36 @@ impl Display {
...
+    ///
+    /// NOTE: the update to the renderer is split from the display update on purpose, since
```
chrisduerr, line 639:
```suggestion
    /// NOTE: The update to the renderer is split from the display update on purpose, since
```

alacritty/src/display/mod.rs
```diff
@@ -618,10 +632,36 @@ impl Display {
...
+    /// on some platforms, like Wayland, resize and other OpenGL operations must be performed
+    /// right before rendering, otherwise they could lock the back buffer resulting in
+    /// rendering with the buffer of old size.
```
chrisduerr, line 642: Love this comment btw, I wasn't completely following the original issue and this makes it very clear.

alacritty/src/display/mod.rs
```diff
@@ -618,10 +632,36 @@ impl Display {
...
+    ///
+    /// Also, this resolves any flickering, since the resize is now synced with frame callbacks.
```
chrisduerr, line 644:
```suggestion
    /// This also resolves any flickering, since the resize is now synced with frame callbacks.
```

alacritty/src/display/mod.rs
```diff
@@ -618,10 +632,36 @@ impl Display {
...
+    pub fn process_renderer_update(&mut self) {
+        let renderer_update = match self.pending_renderer_update.take() {
+            Some(renderer_update) => renderer_update,
+            _ => return,
+        };
+
+        if renderer_update.resize {
+            // Resize renderer.
```
chrisduerr, line 652:
```suggestion
        // Resize renderer.
        if renderer_update.resize {
```

chrisduerr (changes requested):
I haven't noticed any differences on X11 when resizing.

kchibisov:
Don't see any difference on macOS.

Windows's resize impl on window is a no-op so doesn't need testing. Should be good from this point.

alacritty/src/display/mod.rs
```diff
@@ -548,11 +562,15 @@ impl Display {
     fn clear_glyph_cache(&mut self) {
```
chrisduerr, line 562:
```suggestion
    fn reset_glyph_cache(&mut self) {
```

Just for consistency, the docs probably should also be updated.

alacritty/src/renderer/text/glyph_cache.rs
```diff
@@ -264,19 +264,22 @@ impl GlyphCache {
...
-        self.cache = HashMap::default();
+        self.cache = Default::default();
 
         self.load_common_glyphs(loader);
     }
 
-    pub fn update_font_size<L: LoadGlyph>(
+    /// Update the inner font size.
+    ///
+    /// NOTE: To reload the renderers's fonts [[`Self::reset_glyph_cache`]] should be called
```
chrisduerr, line 277:
```suggestion
    /// NOTE: To reload the renderers's fonts [`Self::reset_glyph_cache`] should be called
```

I'd assume the double brackets were just a typo? Or is there actually any formatting difference in the output with double brackets?

chrisduerr (changes requested):

chrisduerr (approved):

PR #6110, opened 2022-06-04
Fix Open Alacritty here in root of drive on windows

IrishBruse:
Before when opening in root of drive
image

The above Log File
```
[0.003076400s] [ERROR] [alacritty] Invalid working directory: "Z:\""
```

After removing `"` error is gone
image

### DIFF
--- a/alacritty/windows/wix/alacritty.wxs
+++ b/alacritty/windows/wix/alacritty.wxs
@@ -53,7 +53,7 @@
             <!-- Add context menu -->
             <Component Id="ContextMenu" Guid="449f9121-f7b9-41fe-82da-52349ea8ff91">
                 <RegistryKey Root="HKCU" Key="Software\Classes\Directory\Background\shell\Open Alacritty here\command">
-                    <RegistryValue Type="string" Value="[AlacrittyProgramFiles]alacritty.exe --working-directory &quot;%V&quot;" KeyPath="yes"/>
+                    <RegistryValue Type="string" Value="[AlacrittyProgramFiles]alacritty.exe --working-directory &quot;%V" KeyPath="yes"/>
                 </RegistryKey>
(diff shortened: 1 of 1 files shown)


alacritty/windows/wix/alacritty.wxs
```diff
@@ -53,7 +53,7 @@
             <!-- Add context menu -->
             <Component Id="ContextMenu" Guid="449f9121-f7b9-41fe-82da-52349ea8ff91">
                 <RegistryKey Root="HKCU" Key="Software\Classes\Directory\Background\shell\Open Alacritty here\command">
-                    <RegistryValue Type="string" Value="[AlacrittyProgramFiles]alacritty.exe --working-directory &quot;%V&quot;" KeyPath="yes"/>
+                    <RegistryValue Type="string" Value="[AlacrittyProgramFiles]alacritty.exe --working-directory &quot;%V" KeyPath="yes"/>
```
chrisduerr, line 56: This doesn't look correct at all. What makes you believe quoting the start without quoting the end is the correct solution? It seems like you just added a workaround for your specific issue?

chrisduerr (changes requested):

alacritty/windows/wix/alacritty.wxs
IrishBruse: It works dont know why but it does if I edit the registry on my machine to do the same thing it fixes the bug without breaking paths with spaces
`"Z:\""`
^ thats what the error spits out if the ending quote is left in there is an extra " for some reason. Looks to me it add an extra " if the path is a root directory.

chrisduerr: Unfortunately `dont know why` isn't exactly giving me enough confidence to merge this change. Especially because as you've said this only seems to affect root directories?

I'd assume you could check what other applications are doing, but unless you have some Microsoft documentation or some other clear indication this is the correct thing to do, I can't merge this.

PR #6112, opened 2022-06-05
Fix old vi cursor not being properly damaged

kchibisov:


kchibisov:
I actually wonder if we should just move Vi cursor logic before returning full damage to simplify things a bit.

chrisduerr (changes requested):
> I actually wonder if we should just move Vi cursor logic before returning full damage to simplify things a bit, the cost isn't a problem here tbf.

I agree, shouldn't do this twice but just have it deduplicated before the early return.

alacritty_terminal/src/term/mod.rs
```diff
@@ -408,14 +391,31 @@ impl<T> Term<T> {
...
         }
 
+        // Add information about old cursor position and new one if they are not the same, so we
+        // cover everything that was produced by `Term::input`.
+        if self.damage.last_cursor != self.grid.cursor.point {
+            // Cursor cooridanates are always inside viewport even if you have `display_offset`.
+            let point =
+                Point::new(self.damage.last_cursor.line.0 as usize, self.damage.last_cursor.column);
+            self.damage.damage_point(point);
+        }
```
chrisduerr, line 401: I don't see why this would be done when the terminal is fully damaged.

alacritty_terminal/src/term/mod.rs
```diff
@@ -408,14 +391,31 @@ impl<T> Term<T> {
...
+
+        // Always damage current cursor.
+        self.damage_cursor();
+        self.damage.last_cursor = self.grid.cursor.point;
+
         if self.damage.last_selection != selection {
             for selection in self.damage.last_selection.into_iter().chain(selection) {
                 self.damage.damage_selection(selection, display_offset, self.columns());
             }
         }
```
chrisduerr, line 411: I don't see why this would be done when the terminal is fully damaged.

chrisduerr (changes requested):
While I'm fine with pulling shared code to before the early return, I don't see why we would execute a bunch of code we know we'll never need.

alacritty_terminal/src/term/mod.rs
```diff
@@ -372,48 +372,55 @@ impl<T> Term<T> {
...
+        let vi_cursor_point = if self.mode.contains(TermMode::VI) {
+            let vi_cursor_point =
+                point_to_viewport(display_offset, self.vi_mode_cursor.point).unwrap();
+            Some(vi_cursor_point)
+        } else {
+            None
+        };
+
+        let previous_vi_cursor_point =
+            mem::replace(&mut self.damage.last_vi_cursor_point, vi_cursor_point);
```
chrisduerr, line 390:
```suggestion
        let display_offset = self.grid().display_offset();
        let vi_cursor_point = if self.mode.contains(TermMode::VI) {
            let vi_cursor_point =
                point_to_viewport(display_offset, self.vi_mode_cursor.point).unwrap();
            Some(vi_cursor_point)
        } else {
            None
        };
        
        let previous_cursor = mem::replace(&mut self.damage.last_cursor, self.grid.cursor.point);
        let previous_selection = mem::replace(&mut self.damage.last_selection, selection);
        let previous_vi_cursor_point =
            mem::replace(&mut self.damage.last_vi_cursor_point, vi_cursor_point);
```

I think this way around would be pretty cool, this way the mem replace stuff is all together.

chrisduerr (changes requested):
Minor suggestion, I think it should be good after.

chrisduerr (approved):
Made a minor style improvement, should be good now.

PR #6115, opened 2022-06-07
Fix clippy unnecessary closure used to substitute value for `Option::None`

jeremycostanzo:
I was not able to run `cargo clippy` successfully on the repo, perhaps because of rust 1.61?


chrisduerr (changes requested):
Alacritty does not use the latest stable clippy. As such this change is inadmissible since it's incompatible with Alacritty's MSRV.

jeremycostanzo:
I saw in the CI run.
I guess we can easily adapt when bumping the MSRV then :)

Issue #5704, opened 2021-12-19
Vi mode recenter with "zz" keybind


thunderblood101:
For bug reports, the following information can help speed up the process. Please
describe the bug that you have found and what you would expect to happen
instead.

### System

OS: Linux
Version: alacritty 0.9.0 (fed349aa)
Linux/BSD: Sway / Wayland

### Logs (Not really applicable)

Crashes: STDERR, STDOUT
Font/Terminal size: `alacritty -vv`
Keyboard and bindings: `alacritty --print-events`

### Summary
In vi mode (accessed by Ctrl-Shift-Space in the default bindings), it would be nice if the keybind "zz" would recenter the window (by moving the line which the cursor is on to the middle of the screen) as it does in vim (it does **not** have the same behavior in unix vi). 

### To reproduce
1. Enter vi mode by entering Ctrl-Shift-Space (or equivilant for your OS, see the config file for Mac/Windows)
2. Place your cursor in a line not near the center of the screen
3. press "zz"
Expected behavior: Centers the cursor
Current behavior: Nothing

chrisduerr:
Adding an action to center the viewport would certainly be possible, but not on `zz`, I'd assume `z` would serve you just as well though?

SmashMineGame:
This is vim's keybind for it. "z" can be followed by a variety of other keys to get different behaviors. For example "zt" scrolls to put the cursor at the top of the screen.

PR #6116, opened 2022-06-07
Center around vi cursor with z

jeremycostanzo:
closes: #5704

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support for OpenGL ES 2.0
 - Escape sequence to set underline color (`CSI 58 : 2 : Ps : Ps : Ps m`/`CSI 58 : 5 : Ps m`)
 - Escape sequence to reset underline color (`CSI 59 m`)
+- Center view around vi cursor with z
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support for OpenGL ES 2.0
 - Escape sequence to set underline color (`CSI 58 : 2 : Ps : Ps : Ps m`/`CSI 58 : 5 : Ps m`)
 - Escape sequence to reset underline color (`CSI 59 m`)
-- Vi mode keybinding (z) to center view around cursor 
+- Vi mode keybinding (z) to center view around vi mode cursor 
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support for OpenGL ES 2.0
 - Escape sequence to set underline color (`CSI 58 : 2 : Ps : Ps : Ps m`/`CSI 58 : 5 : Ps m`)
 - Escape sequence to reset underline color (`CSI 59 m`)
-- Center view around vi cursor with z
+- Vi mode keybinding (z) to center view around cursor 
 
 ### Changed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -625,6 +625,8 @@
 #   - ToggleBlockSelection
 #   - ToggleSemanticSelection
 #       Toggle semantic selection based on `selection.semantic_escape_chars`.
+#   - CenterAroundCursor
+#       Center view around cursor
 #
 # - Vi mode exclusive cursor motion actions:
 #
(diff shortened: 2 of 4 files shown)


jeremycostanzo:
> 

Thank you very much for your review, I'll have a look tomorrow for the documentation.

I am not too sure on how I could disallow the action to work outside of Vi mode?

Should I make it a ViMotion? It would be a bit different than the others, since this action actually scolls the screen.

Looking at the ViActions, it doesn't seem to be a right fit either since none of them move the cursor or scroll the screen.

kchibisov:
>Should I make it a ViMotion? It would be a bit different than the others, since this action actually scolls the screen.

I think you can check if the user is in ViMode and just no-op? `if term.mode.contains(TermMode::VI)`. You should be able to find similar code around.

I also don't think that moving to actions is something I'd want, though @chrisduerr might have different opinion.

jeremycostanzo:
>  I think you can check if the user is in ViMode and just no-op? if term.mode.contains(TermMode::VI). You should be able to find similar code around.

Isn't that actually the purpose of 
```
+BindingMode::VI, ~BindingMode::SEARCH
```
?

I didn't dive too deep to see how the bitflags are actually used though.

kchibisov:
>Isn't that actually the purpose of

Those are for bindings and for binding mods, however I can define your binding without `Vi` mode. So with your current PR it'll recenter to non-existing vi cursor. What I want is just to be a no-op when the user defines it for some other mode.

The thing you've added in `bindings.rs` is more about config, they don't limit any bindings to any particular mode.

kchibisov:
>Should I make it a ViMotion? It would be a bit different than the others, since this action actually scolls the screen.

Actually, you should move it to `ViAction`, this way you won't need the check iirc. I just forgot that we had something like that for general actions, and not just for motions.

jeremycostanzo:
> > Should I make it a ViMotion? It would be a bit different than the others, since this action actually scolls the screen.
> 
> Actually, you should move it to `ViAction`, this way you won't need the check iirc. I just forgot that we had something like that for general actions, and not just for motions.

Yeah, it probably makes more sense, it would also be more straightforward in the documentation.

alacritty/src/input.rs
```diff
@@ -230,6 +230,16 @@ impl<T: EventListener> Execute<T> for Action {
                     ctx.mark_dirty();
                 }
             },
+            Action::Vi(ViAction::CenterAroundCursor) => {
+                // Center around vi mode cursor.
```
kchibisov, line 234: Not sure that this comment is needed.

alacritty/src/input.rs
```diff
@@ -230,6 +230,16 @@ impl<T: EventListener> Execute<T> for Action {
...
+                let term = ctx.terminal_mut();
```
kchibisov, line 235:
```suggestion
                let term = ctx.terminal();
```

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support for OpenGL ES 2.0
 - Escape sequence to set underline color (`CSI 58 : 2 : Ps : Ps : Ps m`/`CSI 58 : 5 : Ps m`)
 - Escape sequence to reset underline color (`CSI 59 m`)
+- Center view around vi cursor with z
```
kchibisov, line 22:
```suggestion
- Vi mode keybinding (z) to center view around cursor
```

kchibisov (changes requested):

kchibisov (approved):

alacritty.yml
```diff
@@ -625,6 +625,8 @@
 #   - ToggleBlockSelection
 #   - ToggleSemanticSelection
 #       Toggle semantic selection based on `selection.semantic_escape_chars`.
+#   - CenterAroundCursor
+#       Center view around cursor
```
chrisduerr, line 629:
```suggestion
#       Center view around vi mode cursor
```

alacritty/src/config/bindings.rs
```diff
@@ -275,6 +275,8 @@ pub enum ViAction {
     SearchEnd,
     /// Launch the URL below the vi mode cursor.
     Open,
+    /// Centers the screen around the cursor.
```
chrisduerr, line 278:
```suggestion
    /// Centers the screen around the vi mode cursor.
```

chrisduerr (changes requested):
We need to make sure to explicitly point out the vi mode cursor is used, since there is another cursor users might confuse it with.

alacritty.yml
```diff
@@ -625,6 +625,8 @@
 #   - ToggleBlockSelection
 #   - ToggleSemanticSelection
 #       Toggle semantic selection based on `selection.semantic_escape_chars`.
+#   - CenterAroundCursor
```
chrisduerr, line 628:
```suggestion
#   - CenterAroundViCursor
```

Probably also a good idea to explicitly point out that it's the vi cursor in the action name. Just to be safe.

kchibisov: I mean, it's already in the section "Vi mode exclusive actions", so I felt like it's redundant.

chrisduerr: It's not though. Even in vi mode the normal cursor position still exists, so you could center around the write-cursor instead of the vi cursor.

Changing the naming is not about pointing out this is a vi action, but pointing out that you're centering around the vi cursor.

jeremycostanzo:
> We need to make sure to explicitly point out the vi mode cursor is used, since there is another cursor users might confuse it with.

Yes I agree, it doesn't cost anything to add two characters.

It seems that the build failed, is it maybe a runner error?

chrisduerr (approved):
Thanks for the patch.

jeremycostanzo:
> Thanks for the patch.

You're welcome.

Thanks you for the reviews @chrisduerr @kchibisov :)

felipesaev (approved):

Issue #5988, opened 2022-04-05
Alacritty has no support for overline (SGR 53/55)


rbanffy:
In other terminals (VTE, QTermWidget, Microsoft) and screen managers (tmux) there is support for overlined text. This is a good option for status lines on the bottom of the screen or other types of divider where you want to control the line immediately under the divider and don't want to interfere with the bottom line of content that sits above the divider/border.

kchibisov:
We've discussed it with @chrisduerr and I'm not sure why it should be added in the first place, however given that it's widely supported and that it's simple to add maybe we should add it...

What do you think @chrisduerr? It's like strikeout, but higher, clamped to top of the cell iirc...

kchibisov:
It basically works like strikeout and supported by vte/konsole and friends. I personally don't see much benefit in it, since it really looks like shitty underline for the one line above.

rbanffy:
The argument for it is how it's useful for things like status lines (it's not intended to be an underline for the line above).

![Gnome Terminal](https://user-images.githubusercontent.com/184714/161855064-feac0429-affb-403d-bd5a-53410967c6f4.png)

![Konsole](https://user-images.githubusercontent.com/184714/161855509-eda06ccd-3196-4980-aabc-94d0cbbff894.png)

One easy to think use is on the shell prompt, as a separator from previous output. Another one is in terminal splits, where separators can also have titles or other indicators for the panel below.

![IBM 3151](https://user-images.githubusercontent.com/184714/161856429-7536caa1-685e-47b5-b2d4-75faed64d964.png)

chrisduerr:
Seems like just changing the background color would be far superior? The existing implementations you're showcasing look like absolute garbage, so that doesn't really make for a good argument. From those screenshots alone it seems pretty obvious to me that monospace fonts will not properly account for an overline and it will look strange and ugly.

rbanffy:
Personally, I don't like changing the background because I, personally, use color very sparingly. I use that option in my `tmux.conf` (and the default tmux status is that RGB-green line - that won't ever change, as that green line is part of their visual identity and you can see a tmux session from a mile away, making it wildly successful for them) when the target terminal doesn't support, say, overlines. Many physical terminals of ancient times used reverse video (because that's what they had) or underline (and that's probably why Wyse terminals had a line on the top of the screen, a position I find most distracting).

In any case, I can understand not wanting to add a feature just because other competing programs support it.

[...]

PR #4501, opened 2020-11-29
Add OpenConsole support for Windows

fredizzimo:
If the conpty.dll and OpenConsole.exe files from the Windows Terminal project are placed in the same directory as the alacritty executable they will be used instead of implementation provided by Windows.

This way you don't need to wait for new Windows releases to get bug fixes and new features.

Implements https://github.com/alacritty/alacritty/issues/3889, and using a more recent version of OpenConsole, fixes issues like bad mouse support https://github.com/alacritty/alacritty/issues/1663, curror issues, and probably much more.

In the future Alacritty could probably bundle the files with the release to make sure that a compatible version is used, but right now it's up to the user to compile the Windows Terminal project himself and copy the files.

Issue #1647, opened 2018-10-12
Support for SGRs 51, 52, and 53


moonheart08:
51, 52, and 53, referred to as Framed, Encircled, and Overlined, are less commonly used ANSI SGR codes, but implementing them would be a potentially useful thing for allowing users to enhance their experience.

chrisduerr:
This is mentioned here:
http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-048.pdf (page 76 or `- 62 -`)

However, it does not look like this is supported in XTerm. Are you aware of any terminal emulator which already supports this? Or an application which makes use of it? To evaluate the priority of this issue that would probably help put it into perspective a bit.

moonheart08:
Priority is very low. Nothing I know of uses it, however I personally wish to use this feature should I find a terminal emulator that supports it. May try and do this myself, but no promises.

chrisduerr:
While I think that Encircled would be quite difficult to add (there's no functionality to render curves), the PR for proper underline support (#1078) does introduce a function to render horizontal lines.

I'd expect that Overlined would be super simple after that PR is landed and Framed could probably be supported with a bit of additional effort.

It does however seem unlikely that this is going to be added in the near future unless you want to take a look at it yourself. However I'm happy to help you if you want to take a stab at it yourself. As mentioned earlier waiting for #1078 to land would probably make sense though.

chrisduerr:
I don't think adding this provides any value unless it's commonly supported and used. So if people aren't using this anymore, I'd say good riddance.

PR #6122, opened 2022-06-09
Overline (SGR 53/55) support

dropbear32:
The overline attribute (SGR code 53 to enable / 55 to disable) is currently unsupported by Alacritty. This PR adds support for those codes.

---

While overlines are properly drawn by this commit if they are switched to be (e.g.) SGR 1, for some reason if SGR 53 or 55 is passed, they are not present in the parameter list passed to `csi_dispatch`, so this commit is non-functional at present. As I do not know Rust, any help in figuring this bug out would be appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support for OpenGL ES 2.0
 - Escape sequence to set underline color (`CSI 58 : 2 : Ps : Ps : Ps m`/`CSI 58 : 5 : Ps m`)
 - Escape sequence to reset underline color (`CSI 59 m`)
+- Escape sequences for overlines (`CSI 53 m` / `CSI 55 m`)
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


chrisduerr (changes requested):
See https://github.com/alacritty/alacritty/issues/5988 and https://github.com/alacritty/alacritty/issues/1647.

atirit:
From what I understood reading those issues, "low priority" meant that the feature was not high on anyone's list of things to work on, and I will note the following:

> So if people aren't using this anymore, I'd say good riddance.

I use these escape codes in a TUI I'm writing, so they're not entirely unused and they're not low-priority for me (hence why I spent time working on this PR).

Please also see [my comment](https://github.com/alacritty/alacritty/issues/5988#issuecomment-1145525597) in #5988.

If Alacritty explicitly does not and will not support certain escape codes, then a different PR should be merged specifying that in the documentation.

chrisduerr:
> If Alacritty explicitly does not and will not support certain escape codes, then a different PR should be merged specifying that in the documentation.

Feel free, I'd be happy to accept that.

atirit:
> this commit is non-functional at present

*on Windows.

The reason this patch does not work on Windows is because ConPTY seems to remove escape sequences it does not recognize, and the version shipped with (my build of) Windows is older than the version that supports SGR 53/55. This is, as far as I'm aware, not fixable without bundling/linking against a newer version of ConPTY, which has unfortunately been explicitly rejected in #4501.

PR #6123, opened 2022-06-10
Update supported escape code list

atirit:
Overlines (SGR 53/55) are, according to #1647, #5988, and #6122, expressly unsupported by Alacritty (i.e. not only do they not work, but support for them will never be merged). This PR updates the documentation to reflect this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - The `--help` output was reworked with a new colorful syntax
 - OSC 52 is now disabled on unfocused windows
 - `SpawnNewInstance` no longer inherits initial `--command`
+- Updated `doc/escape_support.md` to more accurately reflect supported escape codes
 
 ### Fixed
 
--- a/docs/escape_support.md
+++ b/docs/escape_support.md
@@ -65,7 +65,9 @@ brevity.
 | `CSI l`    | PARTIAL     | See `CSI h` for supported modes                   |
 | `CSI ? l`  | PARTIAL     | See `CSI ? h` for supported modes                 |
 | `CSI M`    | IMPLEMENTED |                                                   |
-| `CSI m`    | IMPLEMENTED |                                                   |
+| `CSI m`    | PARTIAL     | `0..9`, `21..25`, `27..49`, `58`, `59`, `90..97`, |
+|            |             | `100..107`                                        |
+|            | REJECTED    | `53`, `55`                                        |
 | `CSI n`    | IMPLEMENTED |                                                   |
 | `CSI P`    | IMPLEMENTED |                                                   |
 | `CSI SP q` | IMPLEMENTED |                                                   |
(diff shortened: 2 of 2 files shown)


docs/escape_support.md
```diff
@@ -65,7 +65,10 @@ brevity.
 | `CSI l`    | PARTIAL     | See `CSI h` for supported modes                   |
 | `CSI ? l`  | PARTIAL     | See `CSI ? h` for supported modes                 |
 | `CSI M`    | IMPLEMENTED |                                                   |
-| `CSI m`    | IMPLEMENTED |                                                   |
+| `CSI m`    | PARTIAL     | Supported parameters:                             |
+|            |             |   `0`-`9`, `21`-`25`, `27`-`49`, `58`, `59`       |
+|            |             |   `90`-`97`, `100`-`107`                          |
```
chrisduerr, line 70: I'd just keep this as `IMPLEMENTED` with nothing on the right side, since those are pretty much all the SGR escape sequences that I'm aware of.

chrisduerr (changes requested):

docs/escape_support.md
chrisduerr: Those are pretty much all unused and can be considered rejected, yes. Probably not a terrible idea to keep the supported parameters list on the right, but `PARTIAL` I'd still replace with `IMPLEMENTED` since as far as I'm aware we've got pretty much everything in use covered.

atirit: I must again contest the suggestion that SGR 51-55 are unused, but that seems to be a moot point.

chrisduerr: They certainly are and have been explicitly rejected in the past: https://github.com/alacritty/alacritty/issues/1647

docs/escape_support.md
```diff
@@ -65,7 +65,10 @@ brevity.
 | `CSI l`    | PARTIAL     | See `CSI h` for supported modes                   |
 | `CSI ? l`  | PARTIAL     | See `CSI ? h` for supported modes                 |
 | `CSI M`    | IMPLEMENTED |                                                   |
-| `CSI m`    | IMPLEMENTED |                                                   |
+| `CSI m`    | IMPLEMENTED | Supported parameters:                             |
+|            |             |   `0`-`9`, `21`-`25`, `27`-`49`, `58`, `59`       |
+|            |             |   `90`-`97`, `100`-`107`                          |
+|            | REJECTED    | `11`-`19`, `53`, `55`                             |
```
chrisduerr, line 71:
```suggestion
|            | REJECTED    | `11`-`19`, `53-55`                                |
```

chrisduerr (changes requested):

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - The `--help` output was reworked with a new colorful syntax
 - OSC 52 is now disabled on unfocused windows
 - `SpawnNewInstance` no longer inherits initial `--command`
+- Updated `doc/escape_support.md` to more accurately reflect supported escape codes
```
chrisduerr, line 29: No need to document documentation updates.

chrisduerr (approved):
Thanks!

PR #6128, opened 2022-06-14
Use `MaybeUninit<usize>` instead of `usize` in `Storage::swap`

Noratrieb:
`Row` contains pointer bytes, which are not valid for `usize`.
Therefore, `MaybeUninit<usize>` should be used to do the untyped copy.

This should not have any effect on codegen, but I wasn't able to check it manually.

### DIFF
--- a/alacritty_terminal/src/grid/storage.rs
+++ b/alacritty_terminal/src/grid/storage.rs
@@ -1,5 +1,6 @@
 use std::cmp::{max, PartialEq};
 use std::mem;
+use std::mem::MaybeUninit;
 use std::ops::{Index, IndexMut};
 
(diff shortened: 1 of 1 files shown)


chrisduerr:
> This should not have any effect on codegen, but I wasn't able to check it manually.

Why not? The entire point of the code is to generate more optimal assembly, so making sure that this does not affect codegen is kind of a big deal.

Nilstrieb:
I copied some code into Godbolt and it revealed that it's indeed still vectorized. https://godbolt.org/z/n8TvzKP78

chrisduerr (approved):
Thanks!

PR #6135, opened 2022-06-15
Ignore special files for in live config reload

kchibisov:
When using `--config-file /dev/null` with `live_config_reload` each
write to `/dev/null` was forcing alacritty to reload its configuration.

This commit makes alacritty ignore special files from
`live_config_reload`.


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Incorrect built-in glyphs for `U+2567` and `U+2568`
 - Character mappings in the DEC special graphics character set (line drawing)
 - Window flickering on resize on Wayland
+- Config reload when using `/dev/null` as a config file
```
kchibisov, line 43: I'm not even sure what to write here.

chrisduerr: Seems fine to me, though I'm not sure in the `Changed` section this is appropriate?

Maybe in `Fixed` and add an `Unnecessary` to the start?

alacritty/src/config/monitor.rs
```diff
@@ -21,6 +21,11 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
         return;
     }
 
+    // Watch only files.
```
chrisduerr, line 24: I think it would be useful to clarify what we consider "files" here. Explicitly pointing out `/dev/null` would be fine too since that should be the most relevant example.

alacritty/src/config/monitor.rs
```diff
@@ -21,6 +21,11 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
...
+    paths.retain(|path| {
+        path.metadata().map(|metadata| metadata.file_type().is_file()).unwrap_or(false)
```
chrisduerr, line 26:
```suggestion
        path.metadata().map_or(false, |metadata| metadata.file_type().is_file())
```

chrisduerr: There really should be a clippy lint for this.

chrisduerr (changes requested):

alacritty/src/config/mod.rs
```diff
@@ -125,6 +125,7 @@ pub fn load(options: &Options) -> UiConfig {
 
 /// Attempt to reload the configuration file.
 pub fn reload(config_path: &Path, options: &Options) -> Result<UiConfig> {
+    debug!("Reloading configuration file from: {:?}", config_path);
```
chrisduerr, line 128:
```suggestion
    debug!("Reloading configuration file from: {:?}", config_path);
    
```

I'd add a newline, just to create some separation.

alacritty/src/config/monitor.rs
```diff
@@ -21,6 +21,10 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
         return;
     }
 
+    // Exclude char devices like `/dev/null`, sockets, and so on, by checking that file type is a
+    // regular file.
+    paths.retain(|path| path.metadata().map_or(false, |metadata| metadata.file_type().is_file()));
```
chrisduerr, line 26: > Tests whether this file type represents a regular file. The result is mutually exclusive to the results of is_dir and is_symlink; only zero or one of these tests may pass.

This is from the Rust docs on `is_file`. This implies that this code would filter symlinks, which is incorrect. So the code should probably be `is_file() || is_symlink()`?

chrisduerr (changes requested):

alacritty/src/config/monitor.rs
kchibisov: No, the `metadate()` call resolves symlinks.

chrisduerr: Oh word, good catch.

alacritty/src/config/mod.rs
```diff
@@ -125,6 +125,8 @@ pub fn load(options: &Options) -> UiConfig {
...
```
chrisduerr, line 128:
```suggestion
    debug!("Reloading configuration file: {:?}", config_path);
```

Don't really care, but I think this is better with the colon.

chrisduerr (approved):

chrisduerr (approved):

PR #6139, opened 2022-06-16
Add support for hyperlink escape sequence

### LINKED ISSUE #922: Support hyperlinks terminal escapes

kchibisov:
This commit adds support for hyperlink escape sequence
`OSC 8 ; params ; URI ST`.

Fixes #922.


kchibisov:
@perfbot

kchibisov:
@perfbot

kchibisov:
@perfbot

kchibisov:
@perfbot

alacritty/src/display/hint.rs
```diff
@@ -258,30 +381,65 @@ pub fn highlighted_at<T>(
...
+
+    let viewport_start = Line(-(term.grid().display_offset() as i32));
+    let viewport_end = viewport_start + term.bottommost_line();
+
+    let mut match_start = Point::new(point.line, Column(0));
+    let mut match_end = Point::new(point.line, Column(term.columns() - 1));
+    let grid = term.grid();
+
+    // Find start and end line containing the same hyperlink inside viewport that is adjustsent to
+    // the requested point's line.
```
chrisduerr, line 412: Very difficult to understand this comment, should be rephrased.

alacritty/src/display/hint.rs
```diff
@@ -258,30 +381,65 @@ pub fn highlighted_at<T>(
...
+
+    // Find start.
+    loop {
+        let next_line = match_start.line - 1i32;
+        if next_line < viewport_start
+            || !grid[next_line]
+                .into_iter()
+                .any(|cell| cell.hyperlink().map(|h| h == hyperlink).unwrap_or(false))
```
chrisduerr, line 420: It's not immediately clear what this does, should be documented, probably extracted in a separate variable with a nice name.

alacritty/src/display/hint.rs
```diff
@@ -258,30 +381,65 @@ pub fn highlighted_at<T>(
...
+        match_start.line = next_line;
+    }
+
+    // Find end.
+    loop {
+        let next_line = match_end.line + 1i32;
+        if next_line >= viewport_end
+            || !grid[next_line]
+                .into_iter()
+                .any(|cell| cell.hyperlink().map(|h| h == hyperlink).unwrap_or(false))
```
chrisduerr, line 434: It's not immediately clear what this does, should be documented, probably extracted in a separate variable with a nice name.

alacritty_terminal/src/ansi.rs
```diff
@@ -1007,6 +1011,25 @@ where
...
             },
 
+            // Hyperlink.
+            b"8" if params.len() > 2 => {
+                let link_params = params[1];
+                let uri = str::from_utf8(params[2]).unwrap_or_default();
+                if uri.is_empty() {
+                    self.handler.set_hyperlink(None);
+                    return;
+                }
```
chrisduerr, line 1021: This should be made more clear, probably by adding a comment and newline as separation. It's not just some error handling, this is how hyperlinks are stopped.

alacritty_terminal/src/ansi.rs
```diff
@@ -1007,6 +1011,25 @@ where
...
+
+                // Link parameters are in format of `key1=value1:key2=value2`. Currently only key
+                // `id` is defined.
+                let id = link_params
+                    .split(|&b| b == b':')
+                    .find(|kv| kv.starts_with(b"id="))
```
chrisduerr, line 1027: I'd recommend `find_map` and `prefix`?

alacritty_terminal/src/term/cell.rs
```diff
@@ -33,6 +34,53 @@ bitflags! {
     }
 }
 
+/// Counter for not-identified hyperinks.
```
chrisduerr, line 37:
```suggestion
/// Counter for hyperlinks without explicit ID.
```

alacritty_terminal/src/term/cell.rs
```diff
@@ -128,6 +182,28 @@ impl Cell {
     pub fn underline_color(&self) -> Option<Color> {
         self.extra.as_ref()?.underline_color
     }
+
+    /// Set hyperlink.
+    pub fn set_hyperlink(&mut self, hyperlink: Option<Hyperlink>) {
+        let drop_hyperlink = hyperlink.is_none()
+            && self.extra.as_ref().map_or(true, |extra| {
+                !extra.zerowidth.is_empty() || extra.underline_color.is_some()
+            });
```
chrisduerr, line 191:
```suggestion
        let should_drop = hyperlink.is_none()
            && self.extra.as_ref().map_or(true, |extra| {
                !extra.zerowidth.is_empty() || extra.underline_color.is_some()
            });
```

I think that makes things a little more descriptive. Especially because we're dropping extra, not hyperlink.

alacritty_terminal/src/term/cell.rs
```diff
@@ -128,6 +182,28 @@ impl Cell {
...
+
+        if drop_hyperlink {
+            self.extra = None;
+            return;
+        }
+
+        let extra = self.extra.get_or_insert(Default::default());
+        Arc::make_mut(extra).hyperlink = hyperlink;
```
chrisduerr, line 199:
```suggestion
        if drop_hyperlink {
            self.extra = None;
        } else {
            let extra = self.extra.get_or_insert(Default::default());
            Arc::make_mut(extra).hyperlink = hyperlink;
        }
```

Should avoid the extra return, just complicates control flow.

alacritty/src/display/mod.rs
```diff
@@ -60,9 +60,6 @@ mod color;
 mod damage;
```
chrisduerr, line 60: I've skipped this file, but I think this should be enough comments for now.

chrisduerr (changes requested):

alacritty_terminal/tests/ref.rs
```diff
@@ -28,15 +28,36 @@ macro_rules! ref_tests {
 }
 
 ref_tests! {
+    alt_reset
+    clear_underline
```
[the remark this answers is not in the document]
kchibisov: yes, `:sort`

alacritty/src/display/hint.rs
```diff
@@ -165,12 +184,48 @@ impl HintState {
...
+        /// Action for handling the text.
+        action: HintAction,
+
+        /// Terminal range matching the hint.
+        bounds: Match,
+
+        /// Hyperlink.
+        hyperlink: Hyperlink,
+    },
+}
```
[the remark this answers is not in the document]
kchibisov: I also thought about it actually, it'll have the same storage though, just a matter of accessing.

alacritty/src/display/hint.rs
```diff
@@ -238,6 +293,74 @@ impl HintLabels {
...
+        let mut end = start;
+
+        while let Some(next_cell) = display_iter.peek() {
+            if next_cell.hyperlink().as_ref() != Some(&hyperlink) {
+                break;
+            }
+
+            end = next_cell.point;
+            let _ = display_iter.next();
+        }
```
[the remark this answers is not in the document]
kchibisov: The `find_map` finds next unique hyperlink, the `while let` loop finds the end of that particular hyperlink, the point here is to compute meaningful bounds.

However what we can do if we don't care about bounds at all is to always use `find_map` and report ranges for hint as 1 cell. That should probably make some sense, I guess?

alacritty_terminal/src/ansi.rs
```diff
@@ -1007,6 +1011,27 @@ where
                 }
             },
 
+            // Hyperlink.
+            b"8" if params.len() > 2 => {
+                let link_params = params[1];
+                let uri = str::from_utf8(params[2]).unwrap_or_default();
+
+                // The OSC 8 escape sequence must be stopped when getting empty `uri`.
```
chrisduerr, line 1019:
```suggestion
                // The OSC 8 escape sequence must be stopped when getting an empty `uri`.
```

chrisduerr (changes requested):

alacritty/src/display/mod.rs
```diff
@@ -362,9 +359,13 @@ pub struct Display {
     /// The renderer update that takes place only once before the actual rendering.
     pub pending_renderer_update: Option<RendererUpdate>,
 
+    // Mouse point position when highlighting hints.
+    hint_mouse_point: Option<Point>,
+
     is_damage_supported: bool,
     debug_damage: bool,
     damage_rects: Vec<DamageRect>,
+    next_frame_damage_rects: Vec<DamageRect>,
```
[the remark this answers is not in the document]
kchibisov: This is required to clear up URL preview on the next frame in case it'll disappear.

alacritty/src/display/mod.rs
```diff
@@ -964,8 +996,12 @@ impl Display {
 
         // Update cursor shape.
         if highlighted_hint.is_some() {
+            // If mouse changed the line, we'd need to update the preview.
+            dirty = self.hint_mouse_point.map(|p| p.line != point.line).unwrap_or(false);
+            self.hint_mouse_point = Some(point);
             self.window.set_mouse_cursor(CursorIcon::Hand);
```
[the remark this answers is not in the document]
kchibisov: This is accurate, since the only thing we care about is line change, the rest and hint change is being handled earlier in this function.

The particular change was added to catch a situation when the hint stays the same,  but cursor changes the line, so we can change URL preview location if it obscures anything.

alacritty/src/display/mod.rs
```diff
@@ -1010,44 +1046,155 @@ impl Display {
...
+        display_offset: usize,
+    ) {
+        let num_cols = self.size_info.columns();
+        let uris: Vec<Vec<char>> = self
+            .highlighted_hint
+            .iter()
+            .chain(&self.vi_highlighted_hint)
+            .filter_map(|hint| hint.hyperlink().map(|hyperlink| hyperlink.uri()))
+            .map(|uri| Self::format_hyperlink_preview(uri, num_cols))
+            .collect();
```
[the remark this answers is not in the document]
kchibisov: We can, but of form `Option<Vec<char>>; 2`, since we should occupy the entire it? I can adjust that, I think.

alacritty/src/display/mod.rs
```diff
@@ -1010,44 +1046,155 @@ impl Display {
...
+            let point =
+                match term::point_to_viewport(display_offset, Point::new(footer_line, Column(0))) {
+                    Some(point) => point,
+                    None => return,
+                };
+
+            // If the point end up outside viewport don't show preview.
+            if point.line > last_visible_line {
+                break;
+            }
```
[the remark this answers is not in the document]
kchibisov: Yes, if we can't place preview anywhere since all the lines are protected we abort. I'm not sure how it could be done nicely.

alacritty/src/display/mod.rs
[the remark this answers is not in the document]
chrisduerr: You're specifically talking about the mouse though, right? And the obfuscation only checks for obfuscation with the cursors? So a mouse change could never introduce a change unless the "mouse" is the vi mode cursor which clearly isn't a mouse?

alacritty/src/display/mod.rs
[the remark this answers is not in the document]
chrisduerr: The `draw_string` method now takes an iterator, right? So do we need a `Vec` here?

alacritty/src/display/mod.rs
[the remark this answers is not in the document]
chrisduerr: I mean the alternative would also be an option, overlay regardless of overlap. I'm not sure which one is better but I feel like in this scenario overlapping might be?

alacritty/src/display/mod.rs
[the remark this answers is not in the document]
kchibisov: No. the mouse cursor could also be obfuscated, that's the point. If you have terminal of height 2 and have the same hyperlink on both lines starting highlight on first line will result in showing preview in the second line, however if you move to the second line preview will still appear on the second line, since the hint wasn't change, because hint occupies multiple lines and haven't changed, only cursor position was changed in that case.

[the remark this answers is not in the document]
chrisduerr: The mouse cursor is *above* the terminal, it's impossible for hints to render above the cursor. I don't see why the mouse cursor's location would be relevant at all, what you're saying sounds like you're talking about the highlighted URL, not the mouse cursor.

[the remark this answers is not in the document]
kchibisov: I mean that preview will obscure the link, not the cursor as a visual object in particular. Right, I meant to say that whatever we highlight with mouse cursor won't be visible. So we should move preview to show that the link is clickable, since preview isn't meant to be clickable.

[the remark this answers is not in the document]
chrisduerr: That has nothing to do with the mouse then and should be more clearly documented.

alacritty/src/display/mod.rs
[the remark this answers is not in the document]
kchibisov: Well, this case could only be reached when you have 2 lines in total in terminal and you highlight hyperlink hint with both vi cursor and mouse pointer at the same time and those point.

The case is pretty rare and unlikely to ever happen, so I don't think adding extra logic of picking some blacklisted point makes sense? Though, if you insist, what we should pick?

alacritty/src/display/mod.rs
```diff
@@ -1010,44 +1046,155 @@ impl Display {
         bar_text
     }
 
-    /// Draw current search regex.
-    fn draw_search(&mut self, config: &UiConfig, size_info: &SizeInfo, text: &str) {
-        let glyph_cache = &mut self.glyph_cache;
-        let num_cols = size_info.columns();
+    /// Format hyperlink preview to make it fit into the terminal.
+    fn format_hyperlink_preview(text: &str, max_chars: usize) -> Vec<char> {
```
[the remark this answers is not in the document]
kchibisov: I've refactored search and hyperlink to use the same function for both of them. I do like the search code a bit more now since it's much easier to read, compared to random truncation math we had before.

alacritty/src/display/mod.rs
```diff
@@ -964,8 +999,13 @@ impl Display {
 
         // Update cursor shape.
         if highlighted_hint.is_some() {
+            // If mouse changed the line, we should update the hyperlink preview, since the
+            // highlithted hint could be disrupted by the old preview.
```
chrisduerr, line 1003:
```suggestion
            // highlighted hint could be disrupted by the old preview.
```

alacritty/src/display/mod.rs
```diff
@@ -980,74 +1020,203 @@ impl Display {
...
-            formatted_regex.push(c);
-            if c.width() == Some(2) {
-                formatted_regex.push(' ');
+    fn format_search(search_regex: &str, search_label: &str, max_width: usize) -> String {
+        let label_len = search_label.len();
+
+        // If the label doesn't fit into screen return.
+        if label_len > max_width {
+            return search_label[..max_width].to_owned();
+        }
```
chrisduerr, line 1029: This comment makes it sound like we don't render the label at all if it doesn't fit, should probably change it to something like this:

```
// Skip `search_regex` formatting if only label is visible.
```

chrisduerr:
@perfbot

alacritty/src/display/mod.rs
```diff
@@ -980,74 +1020,203 @@ impl Display {
...
+
+        // The search string consists of `search_label` + `search_regex` + `cursor`.
+        let mut bar_text = String::from(search_label);
+        bar_text.extend(
+            Self::format_visible_text(
+                iter::once('_').chain(search_regex.chars().rev()),
```
chrisduerr, line 1035: Wait why do we pass the text in reverse here, without any docs to explain it?

Passing it in reverse just to then iterate over the allocated string and reversing it again seems highly inefficient.

alacritty/src/display/mod.rs
```diff
@@ -980,74 +1020,203 @@ impl Display {
...
             }
+
+            visible_text.push(ch);
+
+            // Add spacers for wide chars.
+            if char_width == 2 {
+                visible_text.push(' ');
+            }
+
+            num_chars += char_width;
```
chrisduerr, line 1092: I recommend putting this in an `else` block rather than `break`ing, it's way too big of a conditional block with the early return to put an early return just to skip 5 lines.

alacritty/src/display/mod.rs
```diff
@@ -980,74 +1020,203 @@ impl Display {
...
-        let mut bar_text = format!("{}{}", search_label, truncated_regex);
+        // The maximum amount of protected lines including the ones we'll show preview on.
+        let max_protected_lines = uris.len() * 2;
+
+        // Lines we shouldn't shouldn't show preview on, because it'll obscure the highlighted
+        // hint.
+        let mut protected_lines = Vec::with_capacity(max_protected_lines);
+        if self.size_info.screen_lines() >= max_protected_lines {
+            // Prefer to show preview even when it'll likely obscure the highlighted hint, when
+            // there's no place left for it.
```
chrisduerr, line 1129: I've tested this and when I have a two-line terminal with search and hover the URL, I don't get any preview.

chrisduerr (changes requested):

alacritty/src/display/mod.rs
kchibisov: We kind of have to, since we should know truncate from the start and we have to know the length. Also we can't return iterator from that method, since we mutate the string, e.g. shorten it Though, I think we somehow look ahead it may work?

I'll try to make it iterator so we can chain `rev`, but I'm not sure how to approach it without, since we shouldn't yeild element we replace later on for example, so we need to know in advance that element is safe to yield.

kchibisov:
I've implemented an iterator for string shortening. Should work now in both directions and place a shoretener for regex strings in search bar.

alacritty/src/display/mod.rs
```diff
@@ -1366,55 +1313,316 @@ fn window_size(
     PhysicalSize::new(width as u32, height as u32)
 }
 
+/// The action performed by [`StrShortener`].
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum TextAction {
+    /// Yeild a spacer.
```
chrisduerr, line 1319:
```suggestion
    /// Yield a spacer.
```

alacritty/src/display/mod.rs
```diff
@@ -1366,55 +1313,316 @@ fn window_size(
...
+    Spacer,
+    /// Terminate state reached.
+    Terminate,
+    /// Yeild a shortener.
```
chrisduerr, line 1323:
```suggestion
    /// Yield a shortener.
```

alacritty/src/display/mod.rs
```diff
@@ -1366,55 +1313,316 @@ fn window_size(
...
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum ShortenDirection {
+    /// Shorten to the start of the string.
+    Left,
+
+    /// Shorten to the end of the string.
+    Right,
+}
+
+/// Iterator that yeild shortened version of the text.
```
chrisduerr, line 1339:
```suggestion
/// Iterator that yields shortened version of the text.
```

alacritty/src/display/mod.rs
```diff
@@ -1366,55 +1313,316 @@ fn window_size(
...
+        mut shortener: Option<char>,
+    ) -> Self {
+        if text.is_empty() {
+            // If we don't have any text don't produce a shortener for it.
+            let _ = shortener.take();
+        }
+
+        let (skip_chars, text_action) = match direction {
+            ShortenDirection::Right => (0, TextAction::Char),
+            ShortenDirection::Left => {
```
chrisduerr, line 1363: Might be easier to early-return for ShortenDirection::Right here? I'm not entirely sure, but this is quite a massive block with two easily avoidable indentations.

chrisduerr (changes requested):
I like the `StrShortener` implementation, in raw LoC count it might be a little more verbose but it's very descriptive and does what it's supposed to do well and clearly.

Could also easily be moved to another module if we want to reduce LoC count in `display/mod.rs`, which probably wouldn't be a terrible idea?

alacritty/src/string.rs
```diff
@@ -0,0 +1,323 @@
...
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum ShortenDirection {
+    /// Shorten to the start of the string.
+    Left,
+
+    /// Shorten to the end of the string.
+    Right,
+}
+
+/// Iterator that yeild shortened version of the text.
```
chrisduerr, line 30:
```suggestion
/// Iterator that yield shortened version of the text.
```

chrisduerr (changes requested):

chrisduerr (approved):

hasufell:
How does this work? Is it released? I tried to build from master and then executed:

```sh
printf '\e]8;;http://example.com\e\\This is a link\e]8;;\e\\\n'
```

It did not produce a clickable link.

kchibisov:
>How does this work? Is it released? I tried to build from master and then executed:

Part of `v0.11.0`. Requires `hyperlinks: true` in the `hints` config (which is default) in case you have custom sections.

You could test with default config `alacritty --config-file /dev/null` Just hovering the mouse should show link preview and highlight the link.

Won't work in tmux unless you build it from master. Likely won't work on Windows, since ConPty.

PR #6148, opened 2022-06-20
Bump dirs to 4.0.0

a5ob7r:
This is to remove dependencies on multiple versions of dirs crate.

There is no breaking change affects alacritty.

https://github.com/dirs-dev/dirs-rs#changelog


a5ob7r:
Thank you.

Issue #6153, opened 2022-06-25
config: `lines` or `columns` not considered if the other one is not set


ssiyad:
### System

OS: Linux
Version: alacritty 0.10.1 ()
Linux/BSD: Wayland, Sway

`window.dimensions.lines` or `window.dimensions.columns` are not considered if the other one is not set

PR #6154, opened 2022-06-25
fix: `lines` or `columns` not conidered if other one is not set

ssiyad:
see #6153


kchibisov (changes requested):
Looks good. Could you add changelog entry for that in `Fixed` section, since it was a bug?

ssiyad:
@kchibisov done

chrisduerr (changes requested):
This is intentional. Otherwise we'd be requesting a column/line count of 0 which doesn't make any sense.

kchibisov:
Hm, I agree with @chrisduerr that sounds a bit strange, but I thought that you just alter certain value, like `lines` and keep columns as default, since you want default width, but just more lines...

chrisduerr:
@kchibisov It's not possible to set just one of these with the operating system APIs available afaik. At least I don't think winit exposes them as such. You always interact with window dimensions through a size, so WxH, never with just width or height.

So as far as I'm concerned, everything works exactly as intended here. @ssiyad where does this feature request come from, what is your usecase? Also are there any other applications that inspired the idea to just set one of the dimensions?

kchibisov:
>You always interact with window dimensions through a size, so WxH, never with just width or height.

It's not actually true? You can set only some specific dimension to certain width and for other you pick some arbitrary value? I understand that you call to system API with both of them. The feature here is that they want to have some dimension being at least e.g. 30 lines.

chrisduerr:
> It's not actually true? You can set only some specific dimension to certain width and for other you pick some arbitrary value?

Yes it is? Winit controls size through `set_inner/outer_size`, you can't set it individually.

> for other you pick some arbitrary value

So why wouldn't the user just add that size to the config explicitly, when we are forced to always send up a size anyway? Provides zero benefit at that point.

ssiyad:
> This is intentional. Otherwise we'd be requesting a column/line count of 0 which doesn't make any sense.

@chrisduerr If this is the case, can we mention that somewhere please? It looks like it is causing some confusion. https://stackoverflow.com/questions/67929417/start-alacritty-with-override-options

chrisduerr:
If you want to change the PR to print a nice warning when only one of them is set, I'm fine with that.

nixpulvis:
Is this something that could be changed in winit? Since I can see some utility in allowing a user to specific a specific width, then having the WM set the height, for example.

kchibisov:
>Is this something that could be changed in winit? Since I can see some utility in allowing a user to specific a specific width, then having the WM set the height, for example.

I don't think that it makes sense, since you need both to draw? And if you want to change just one you can pick the current values and apply them with the mix of old a new.

nixpulvis:
@kchibisov just for my own understanding; how does a window know what size to be initially when the user doesn't mention it specifically?

kchibisov:
>@kchibisov just for my own understanding; how does a window know what size to be initially when the user doesn't mention it specifically?

winit uses some initial size for the window like `800x600`, some systems do tell what the size of the window application should use, when the user hasn't specified the window size themselves. So from what I know there's no clear rule, since even on Wayland it's only a sort of recommendation.

nixpulvis:
I guess all I'm saying is that, if the user only specifies `width = 100 columns`, then it would be nice if the underlying window system would respect that while defaulting to the normal height, for example.

kchibisov:
>I guess all I'm saying is that, if the user only specifies width = 100 columns, then it would be nice if the underlying window system would respect that, while defaulting to the normal height, for example.

That's my expectation as well, though @chrisduerr thinks otherwise afaics.

chrisduerr:
> That's my expectation as well, though @chrisduerr thinks otherwise afaics.

I don't see how you would expect that given that a size is set rather than just one of the dimensions. Unless one of you can point me at an API that would make this possible, which exists on Windows/macOS/X11 AND Wayland, then I don't see how this could be done in winit?

Issue #6145, opened 2022-06-18
Hexadecimal input for --embed


BioBox:
Can you please add suport for hexadecimal xids for the `--embed` option so that I don't have to do these extra steps _every time_ I start my terminal in a tabbed instance?
```fish
set -x xid (tabbed -d)
set plain_id (string sub -s 3 $xid | tr '[a-f]' '[A-F]')
set -x did (echo "obase=10; ibase=16; $plain_id" | bc)
alacritty --embed $did &>/dev/null &
```
As you can see, `tabbed` - as well as every other program I know of that supports xembed, uses hexadecimal. Just a simple check for the `0x` prefix would go a long way.

chrisduerr:
On the one hand, I don't really care what other applications decide to output their IDs with, conversion is trivial and we shouldn't have to change Alacritty any time someone decides to add a new output format.

On the other hand, hexadecimal should be pretty easy to support, so if someone is willing to write a patch, upstreaming shouldn't be a big problem.

PR #6158, opened 2022-06-26
support hexadecimal input for `--embed`

ssiyad:
https://github.com/alacritty/alacritty/issues/6145

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,13 +20,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Escape sequence to set underline color (`CSI 58 : 2 : Ps : Ps : Ps m`/`CSI 58 : 5 : Ps m`)
 - Escape sequence to reset underline color (`CSI 59 m`)
 - Vi mode keybinding (z) to center view around vi mode cursor
+- Accept hexadecimal values starting with `0x` for `--embed`
 
 ### Changed
 
 - The `--help` output was reworked with a new colorful syntax
 - OSC 52 is now disabled on unfocused windows
 - `SpawnNewInstance` no longer inherits initial `--command`
-- `--embed` could accept hexadecimal values starting with `0x`
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - The `--help` output was reworked with a new colorful syntax
 - OSC 52 is now disabled on unfocused windows
 - `SpawnNewInstance` no longer inherits initial `--command`
+- `--embed` could accept hexadecimal values starting with `0x`
 
 ### Fixed
 
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -100,16 +100,7 @@ impl Options {
         }
 
         config.window.dynamic_title &= self.window_options.window_identity.title.is_none();
-        config.window.embed = self.embed.as_ref().and_then(|embed| {
-            // convert hexadecimal to decimal if possible
-            if let (true, Ok(v)) =
-                (embed.starts_with("0x"), u64::from_str_radix(embed.trim_start_matches("0x"), 16))
-            {
-                return Some(v);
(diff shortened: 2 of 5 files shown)


kchibisov (changes requested):
In general the approach looks good.

However I'd probably write something like

```diff
diff --git a/alacritty/src/cli.rs b/alacritty/src/cli.rs
index be377700..babdcafa 100644
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -100,7 +100,13 @@ impl Options {
         }
 
         config.window.dynamic_title &= self.window_options.window_identity.title.is_none();
-        config.window.embed = self.embed.as_ref().and_then(|embed| embed.parse().ok());
+        config.window.embed = self.embed.as_ref().and_then(|embed| {
+            embed
+                .starts_with("0x")
+                .then(|| u64::from_str_radix(&embed[2..], 16).ok())
+                .unwrap_or_else(|| embed.parse().ok())
+        });
+
         config.debug.print_events |= self.print_events;
         config.debug.log_level = max(config.debug.log_level, self.log_level());
         config.debug.ref_test |= self.ref_test;
```

And probably put it into `parse_hex_or_decimanl(&str)` function with a test for it. You can look at `parse_class`, so the end result will look like

```rust
config.window.embed = self.embed.as_ref(),and_then(parse_hex_or_decimal)`
```

kchibisov (changes requested):
Looks good, could you add an entry to `Changed` section in `CHANGELOG.md`, like: `--embed option could accept hexadecimal values by providing `0x`.

Also, you should tune `--help` output, since right now it only states `decimal`, but not hexadecimal starting with `0x`.

alacritty/src/cli.rs
```diff
@@ -173,6 +173,14 @@ fn parse_class(input: &str) -> Result<Class, String> {
     }
 }
 
+/// Convert to hex if possible, else decimal
+fn parse_hex_or_decimal(input: &str) -> Option<u64> {
+    input
+        .starts_with("0x")
+        .then(|| u64::from_str_radix(&input[2..], 16).ok())
+        .unwrap_or_else(|| input.parse().ok())
```
chrisduerr, line 181: I recommend using the `strip_prefix` method here.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - The `--help` output was reworked with a new colorful syntax
 - OSC 52 is now disabled on unfocused windows
 - `SpawnNewInstance` no longer inherits initial `--command`
+- `--embed` could accept hexadecimal values starting with `0x`
```
chrisduerr, line 29: This is a new feature, should be moved to `Added`. "could" is also not appropriate here, I recommend the following:

```
 - Accept hexadecimal values starting with `0x` for `--embed`
```

alacritty/src/cli.rs
```diff
@@ -25,7 +25,7 @@ pub struct Options {
     #[clap(long)]
     pub ref_test: bool,
 
-    /// Defines the X11 window ID (as a decimal integer) to embed Alacritty within.
+    /// Defines the X11 window ID (as a decimal integer or hexadecimal starting with 0x) to embed Alacritty within.
```
chrisduerr, line 28: I think we can do much better than this.

Maybe something like that:

```
/// X11 window ID to embed Alacritty within (decimal or hexadecimal with "0x" prefix).
```

chrisduerr (changes requested):

alacritty/src/cli.rs
```diff
@@ -173,6 +173,14 @@ fn parse_class(input: &str) -> Result<Class, String> {
     }
 }
 
+/// Convert to hex if possible, else decimal
+fn parse_hex_or_decimal(input: &str) -> Option<u64> {
```
kchibisov, line 177: Seems like we should use `c_ulong` instead of `u64` here.

chrisduerr (approved):
LGTM

kchibisov (approved):

ssiyad:
@chrisduerr @kchibisov As a rust newbie, I couldn't have done this without your guidance. Thank you!

PR #6160, opened 2022-06-27
Publish to Windows Package Managar (WinGet)

vedantmgoyal2009:
This action automatically generates manifests for WinGet Community Repository ([microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs)) and submits them.

Before merging this:
1. Please add a GitHub token with `public_repo` scope as a repository secret and rename the secret name in the workflow.
2. Fork [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs) under @alacritty.

### DIFF
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -1,0 +1,13 @@
+name: Publish to WinGet
+on:
+  release:
+    types: [released]
+jobs:
+  publish:
+    runs-on: windows-latest # action can only be run on windows
+    steps:
+      - uses: vedantmgoyal2009/winget-releaser@latest
+        with:
+          identifier: Alacritty.Alacritty
+          installers-regex: '\.msi$' # only .exe files
+          token: ${{ secrets.WINGET_TOKEN }}


kchibisov:
I don't understand why do we need this in the first place. The ones who should be responsible for package manifest for winget should be the ones who is responsible for winget, not us.

From what I can see alacritty is present on winget https://github.com/microsoft/winget-pkgs/blob/6ace2b585a12fddfcefedd74f8a0f226ec1cfe56/manifests/a/Alacritty/Alacritty/0.10.1/Alacritty.Alacritty.locale.en-US.yaml, so I don't think that we need any action here?

chrisduerr (changes requested):
> It is present, but the manifest has to be updated with every new release of Alacritty.

Which is the job of the package maintainer, who shouldn't also be the developer.

vedantmgoyal2009:
Well, here, the package maintainer is github actions :)

kchibisov:
@vedantmgoyal2009 personally I dislike the idea of github workflow, since I think it's not audited? If folks do care about release to winget, they could send patches there, that would at least indicate that they audited the update and that it works...

That's how it works with all linux distros, homebrew, etc. Not sure how those are different from winget.

PR #6165, opened 2022-06-28
Fix resize on Wayland when creating new window

kchibisov:
This fixes the issue when the window that got `CreateWindow` event
gets resized after creating that new window.

Follow-up to 90552e3.

### DIFF
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -495,6 +495,16 @@ impl Window {
         self.windowed_context.resize(size);
     }
 
+    pub fn make_not_current(&mut self) {
+        if self.windowed_context.is_current() {
+            self.windowed_context.replace_with(|context| unsafe {
+                // We do ensure that context is current before any rendering operation due to multi
+                // window support, so we don't need extra "type aid" from glutin here.
+                context.make_not_current().expect("context swap").treat_as_current()
+            });
+        }
+    }
+
     pub fn make_current(&mut self) {
         if !self.windowed_context.is_current() {
             self.windowed_context
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1344,6 +1344,13 @@ impl Processor {
                 GlutinEvent::UserEvent(Event {
                     payload: EventType::CreateWindow(options), ..
                 }) => {
+                    // XXX Ensure that no context is current when creating a new window, otherwise
+                    // it may lock the backing buffer of the surface of current context when asking
+                    // e.g. EGL on Wayland to create a new context.
+                    for window_context in self.windows.values_mut() {
+                        window_context.display.window.make_not_current();
+                    }
+
                     if let Err(err) = self.create_window(event_loop, proxy.clone(), options) {
                         error!("Could not open window: {:?}", err);
                     }


alacritty/src/window_context.rs
```diff
@@ -328,6 +332,14 @@ impl WindowContext {
             // Force the display to process any pending display update.
             self.display.process_renderer_update();
 
+            // XXX it seems like on Wayland when creating a new window the buffer of the old window
+            // may lock leading to resize not being honored, so ask for one more frame to draw in
+            // that particular case.
```
chrisduerr, line 337:
```suggestion
            // XXX: On Wayland when creating a new window the buffer of the old window
            // may lock leading to resize not being honored, so ask for one more frame to draw in
            // that particular case.
```

We should be certain, not guessing.

alacritty/src/event.rs
```diff
@@ -1344,6 +1344,16 @@ impl Processor {
                 GlutinEvent::UserEvent(Event {
                     payload: EventType::CreateWindow(options), ..
                 }) => {
+                    #[cfg(not(any(target_os = "macos", windows)))]
+                    for window_context in self.windows.values_mut() {
+                        // The window is currently current, so redraw it on the next frame as well
+                        // on Wayland, otherwise it could have the buffer of the old size, if
+                        // resize happened.
```
chrisduerr, line 1351:
```suggestion
                        // The window's context is current, so redraw it on the next frame as well
                        // on Wayland, otherwise if a resize happened it could still have the buffer of the old size.
```

chrisduerr (changes requested):
I **really** don't like this patch. I don't doubt it works but just redrawing twice seems like a major hack rather than a proper solution.

alacritty/src/display/window.rs
```diff
@@ -495,6 +495,17 @@ impl Window {
         self.windowed_context.resize(size);
     }
 
+    pub fn make_not_current(&mut self) {
+        if self.windowed_context.is_current() {
+            // SAFETY due to use of multiple windows we always ensure that the context is current
+            // before rendering, so treating it as current here to avoid extra complexity from
+            // addign extra enum.
```
chrisduerr, line 502:
```suggestion
            // adding extra enum.
```

chrisduerr, line 502: This doesn't prove safety at all? If this function is called outside of rendering, the window will not be current.

alacritty/src/event.rs
```diff
@@ -1344,6 +1344,13 @@ impl Processor {
                 GlutinEvent::UserEvent(Event {
                     payload: EventType::CreateWindow(options), ..
                 }) => {
+                    // XXX Ensure that no context is current when creating a new window, otherwise
+                    // it may lock the current context when asking e.g. EGL on Wayland to create
+                    // new context.
```
chrisduerr, line 1349:
```suggestion
                    // XXX Ensure that no context is current when creating a new window, otherwise
                    // it may lock the current context when asking e.g. EGL on Wayland to create a
                    // new context.
```

chrisduerr (changes requested):

alacritty/src/display/window.rs
chrisduerr: > I think the comment is a bit misleading.

That's exactly my problem, the comment is misleading.

chrisduerr (approved):

PR #6172, opened 2022-06-30
Add `cursor.blink_timeout` config option

### LINKED ISSUE #5992: Consider pausing blinking after certain time of inactivity
I think it's quite common among all editors/toolkits I've seen. Could make it non-configurable and be like 30 seconds or so.

kchibisov:
This option should prevent extensive power usage due to cursor blinking
when there's no user activity being performed.

Fixes #5992.

--

I'm not quite sure when we should restart blinking and how we should handle defaults that doesn't make sense, e.g. when timeout is lower than blinking interval. Also, should we update cursor blinking when mouse moves, etc?

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -90,6 +90,7 @@ pub enum EventType {
     Scroll(Scroll),
     CreateWindow(WindowOptions),
     BlinkCursor,
+    BlinkCursorTimeout,
     SearchNext,
 }
 
@@ -180,6 +181,7 @@ pub struct ActionContext<'a, N, T> {
     pub display: &'a mut Display,
     pub message_buffer: &'a mut MessageBuffer,
     pub config: &'a mut UiConfig,
+    pub cursor_blink_timed_out: &'a mut bool,
     pub event_loop: &'a EventLoopWindowTarget<Event>,
     pub event_proxy: &'a EventLoopProxy<Event>,
     pub scheduler: &'a mut Scheduler,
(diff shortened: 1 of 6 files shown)


alacritty/src/event.rs
```diff
@@ -1048,6 +1077,14 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                     self.ctx.display.cursor_hidden ^= true;
                     *self.ctx.dirty = true;
                 },
+                EventType::BlinkCursorTimeout => {
+                    // Disable blinking after timeout reached.
+                    let timer_id = TimerId::new(Topic::BlinkCursor, self.ctx.display.window.id());
+                    self.ctx.scheduler.unschedule(timer_id);
+                    self.ctx.display.cursor_hidden = false;
+                    *self.ctx.blink_cursor_timeout = true;
```
chrisduerr, line 1085: I think `cursor_blink_timed_out` might be a better variable name? Or something like that at least. `blink_cursor_timeout` as a flag name doesn't sit quite right with me.

alacritty_terminal/src/config/mod.rs
```diff
@@ -13,8 +13,11 @@ use crate::ansi::{CursorShape, CursorStyle};
 pub use crate::config::scrolling::{Scrolling, MAX_SCROLLBACK_LINES};
 
 pub const LOG_TARGET_CONFIG: &str = "alacritty_config_derive";
+
 const MIN_BLINK_INTERVAL: u64 = 10;
 
+const MAX_BLINK_TIMEOUT: u64 = 60;
```
chrisduerr, line 19: Do we need a max here? Shouldn't really matter, if the user wants something ridiculous that's up to them.

chrisduerr (changes requested):
> I'm not quite sure when we should restart blinking and how we should handle defaults that doesn't make sense, e.g. when timeout is lower than blinking interval.

I don't think it matters that much how we behave in scenarios that don't make much sense anyway. As long as we're not pathological it should be fine.

> Also, should we update cursor blinking when mouse moves, etc?

I'm not quite sure what the motivation of the original issue is, since I don't really like blinking at all. But I feel like motion alone shouldn't be sufficient, but clicking/dragging should?

alacritty_terminal/src/config/mod.rs
kchibisov: I just don't want to have timer that runs like for a year and then stops your cursor. So capping in a minute made sense.

chrisduerr: Why not?

kchibisov: ¯\_(ツ)_/¯

kchibisov:
>But I feel like motion alone shouldn't be sufficient, but clicking/dragging should?

I've checked other vte, and they don't start blinking when interacting with a mouse. Only when I start typing again. Which I do like, since in general I don't want to start blinking if I've selected something.

But it's not like it's hard to alter it...

chrisduerr:
> I've checked other vte, and they don't start blinking when interacting with a mouse. Only when I start typing again. Which I do like, since in general I don't want to start blinking if I've selected something.

Fine by me tbh.

chrisduerr (approved):

PR #6174, opened 2022-06-30
Add IPC config subcommand

### LINKED ISSUE #472: Runtime configuration via IPC
It would be great it it was possible to make changes to Alacritty's configuration at runtime via some type of IPC such as a Unix domain socket, DBus endpoint, named pipe, web service, or similar. This would not only allow users to make changes to an of Alacritty's configuration parameters from scripts and other utilities but it would help keep the UI from being weighed down with endless config options.

chrisduerr:
This patch adds a new mechanism for changing configuration options
without editing the configuration file, by sending options to running
instances through `alacritty msg`.

Each window will load Alacritty's configuration file by default and then
accept IPC messages for config updates using the `alacritty msg config`
subcommand. By default all windows will be updated, individual windows
can be addressed using `alacritty msg config --window-id
"$ALACRITTY_WINDOW_ID"`.

Each option will replace the config's current value and cannot be reset
until Alacritty is restarted or the option is overwritten with a new
value.

Configuration options are passed in the format `field.subfield=value`,
where `value` is interpreted as yaml.

Closes #472.

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -125,7 +125,7 @@ impl Options {
     }
 }
 
-/// Combine multiple options inta a [`serde_yaml::Value`].
+/// Combine multiple options into a [`serde_yaml::Value`].
 pub fn options_as_value(options: &[String]) -> Value {
     options.iter().fold(Value::default(), |value, option| match option_as_value(option) {
         Ok(new_value) => serde_utils::merge(value, new_value),
@@ -270,7 +270,7 @@ pub enum Subcommands {
 #[derive(Args, Debug)]
 pub struct MessageOptions {
     /// IPC socket connection path override.
-    #[clap(long, short, value_hint = ValueHint::FilePath)]
+    #[clap(short, long, value_hint = ValueHint::FilePath)]
     pub socket: Option<PathBuf>,
 
     /// Message which should be sent.
@@ -309,9 +309,15 @@ pub struct IpcConfig {
     #[clap(required = true, value_name = "CONFIG_OPTIONS")]
     pub options: Vec<String>,
 
-    /// Window ID for the new config
-    #[clap(long, short)]
-    pub window_id: Option<u64>,
+    /// Window ID for the new config.
+    ///
+    /// Use `-1` to apply this change to all windows.
+    #[clap(short, long, allow_hyphen_values = true, env = "ALACRITTY_WINDOW_ID")]
+    pub window_id: Option<i64>,
+
+    /// Clear all runtime configuration changes.
+    #[clap(short, long, conflicts_with = "options")]
+    pub reset: bool,
 }
 
 #[cfg(test)]
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -81,14 +81,7 @@ impl Options {
         let mut options = Self::parse();
 
         // Convert `--option` flags into serde `Value`.
-        for option in &options.option {
-            match option_as_value(option) {
-                Ok(value) => {
-                    options.config_options = serde_utils::merge(options.config_options, value);
-                },
-                Err(_) => eprintln!("Invalid CLI config option: {:?}", option),
-            }
-        }
+        options.config_options = options_as_value(&options.option);
 
         options
     }
@@ -132,7 +125,18 @@ impl Options {
     }
 }
 
-/// Format an option in the format of `parent.field=value` to a serde Value.
+/// Combine multiple options inta a [`serde_yaml::Value`].
+pub fn options_as_value(options: &[String]) -> Value {
+    options.iter().fold(Value::default(), |value, option| match option_as_value(option) {
+        Ok(new_value) => serde_utils::merge(value, new_value),
+        Err(_) => {
+            eprintln!("Ignoring invalid option: {:?}", option);
+            value
+        },
+    })
+}
+
+/// Parse an option in the format of `parent.field=value` as a serde Value.
 fn option_as_value(option: &str) -> Result<Value, serde_yaml::Error> {
     let mut yaml_text = String::with_capacity(option.len());
     let mut closing_brackets = String::new();
@@ -280,9 +284,12 @@ pub struct MessageOptions {
 pub enum SocketMessage {
     /// Create a new window in the same Alacritty process.
     CreateWindow(WindowOptions),
+
+    /// Update the Alacritty configuration.
+    Config(IpcConfig),
 }
 
-/// Subset of options that we pass to a 'create-window' subcommand.
+/// Subset of options that we pass to 'create-window' IPC subcommand.
 #[derive(Serialize, Deserialize, Args, Default, Clone, Debug, PartialEq, Eq)]
 pub struct WindowOptions {
     /// Terminal options which can be passed via IPC.
@@ -292,6 +299,19 @@ pub struct WindowOptions {
     #[clap(flatten)]
     /// Window options which could be passed via IPC.
     pub window_identity: WindowIdentity,
+}
+
+/// Parameters to the `config` IPC subcommand.
+#[cfg(unix)]
+#[derive(Args, Serialize, Deserialize, Default, Debug, Clone, PartialEq, Eq)]
+pub struct IpcConfig {
+    /// Configuration file options [example: cursor.style=Beam].
+    #[clap(required = true, value_name = "CONFIG_OPTIONS")]
+    pub options: Vec<String>,
+
+    /// Window ID for the new config
+    #[clap(long, short)]
+    pub window_id: Option<u8>,
 }
 
 #[cfg(test)]
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -311,7 +311,7 @@ pub struct IpcConfig {
 
     /// Window ID for the new config
     #[clap(long, short)]
-    pub window_id: Option<u8>,
+    pub window_id: Option<u64>,
 }
 
 #[cfg(test)]
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -313,7 +313,7 @@ pub struct IpcConfig {
     ///
     /// Use `-1` to apply this change to all windows.
     #[clap(short, long, allow_hyphen_values = true, env = "ALACRITTY_WINDOW_ID")]
-    pub window_id: Option<i64>,
+    pub window_id: Option<i128>,
 
     /// Clear all runtime configuration changes.
     #[clap(short, long, conflicts_with = "options")]
--- a/alacritty/src/ipc.rs
+++ b/alacritty/src/ipc.rs
@@ -1,11 +1,11 @@
 //! Alacritty socket IPC.
 
+use glutin::window::WindowId;
 use std::ffi::OsStr;
 use std::io::{BufRead, BufReader, Error as IoError, ErrorKind, Result as IoResult, Write};
 use std::os::unix::net::{UnixListener, UnixStream};
 use std::path::PathBuf;
 use std::{env, fs, process};
-use glutin::window::WindowId;
 
 use glutin::event_loop::EventLoopProxy;
 use log::warn;
--- a/alacritty/src/ipc.rs
+++ b/alacritty/src/ipc.rs
@@ -62,6 +62,13 @@ pub fn spawn_ipc_socket(options: &Options, event_proxy: EventLoopProxy<Event>) -
                     let event = Event::new(EventType::CreateWindow(options), None);
                     let _ = event_proxy.send_event(event);
                 },
+                SocketMessage::Config(ipc_config) => {
+                    // TODO: Waiting on window ID patch.
+                    // let window_id = ipc_config.window_id;
+                    let window_id = None;
+                    let event = Event::new(EventType::IpcConfig(ipc_config), window_id);
+                    let _ = event_proxy.send_event(event);
+                },
             }
         }
     });
--- a/alacritty/src/ipc.rs
+++ b/alacritty/src/ipc.rs
@@ -5,6 +5,7 @@ use std::io::{BufRead, BufReader, Error as IoError, ErrorKind, Result as IoResul
 use std::os::unix::net::{UnixListener, UnixStream};
 use std::path::PathBuf;
 use std::{env, fs, process};
+use glutin::window::WindowId;
 
 use glutin::event_loop::EventLoopProxy;
 use log::warn;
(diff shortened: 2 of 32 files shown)


paum3:
This is a great idea! Looking forward to it. thanx

kchibisov:
@chrisduerr could you rebase it? I think last time it you were waiting on winit update, since you needed a way to reference windows. That's why I've avoided reviewing it before.

chrisduerr:
> @chrisduerr could you rebase it? I think last time it you were waiting on winit update, since you needed a way to reference windows. That's why I've avoided reviewing it before.

On that right now actually. Will try to finish it up today.

alacritty/src/cli.rs
```diff
@@ -132,7 +125,18 @@ impl Options {
     }
 }
 
-/// Format an option in the format of `parent.field=value` to a serde Value.
+/// Combine multiple options inta a [`serde_yaml::Value`].
```
kchibisov, line 128:
```suggestion
/// Combine multiple options into a [`serde_yaml::Value`].
```

alacritty/src/window_context.rs
```diff
@@ -81,13 +87,16 @@ impl WindowContext {
...
-            config,
+            &config,
             window_event_loop,
             &identity,
             #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
             wayland_event_queue,
         )?;
 
+        // Export window ID environment variable.
+        env::set_var("ALACRITTY_WINDOW_ID", format!("{:?}", u64::from(display.window.id())));
```
kchibisov, line 98: In general, we should unset it right after spawning child process, otherwise it'll be inherited by the shell scripts user is executing, and given that it contains the window id of the last window, which may be closed, it should be unset right away. So the window id reference will be present only in child.

alacritty/src/window_context.rs
```diff
@@ -246,11 +299,13 @@ impl WindowContext {
...
-                window_id,
+                ..
             } => {
                 let size = (new_inner_size.width, new_inner_size.height);
-                let event =
-                    Event::new(EventType::ScaleFactorChanged(scale_factor, size), window_id);
+                let event = Event::new(
+                    EventType::ScaleFactorChanged(scale_factor, size),
+                    self.display.window.id(),
+                );
```
kchibisov, line 308: Not sure why changing this? We should use the value that came with event.

alacritty/src/ipc.rs
```diff
@@ -1,5 +1,6 @@
 //! Alacritty socket IPC.
 
+use glutin::window::WindowId;
```
kchibisov, line 3: Should move it below to the other glutin import.

extra/alacritty-msg.man
```diff
@@ -10,20 +10,39 @@ making it possible to control Alacritty without directly accessing it.
...
+.TP
+\fB<CONFIG_OPTIONS>...\fR
+Configuration file options [example: cursor.style=Beam]
+.RE
+.TP
+.SH "\tOPTIONS"
+.RS 12
+.TP
+\fB\-w\fR, \fB\-\-window\-id\fR <WINDOW_ID>
+Window ID for the new config
```
kchibisov, line 44: Should reference the env variable you've added.

alacritty/src/cli.rs
```diff
@@ -297,6 +304,19 @@ pub struct WindowOptions {
...
+#[cfg(unix)]
+#[derive(Args, Serialize, Deserialize, Default, Debug, Clone, PartialEq, Eq)]
+pub struct IpcConfig {
+    /// Configuration file options [example: cursor.style=Beam].
+    #[clap(required = true, value_name = "CONFIG_OPTIONS")]
+    pub options: Vec<String>,
+
+    /// Window ID for the new config
+    #[clap(long, short)]
+    pub window_id: Option<u64>,
```
kchibisov, line 317: Could you get the default out of the `env` variable, so you don't have to specify it when sending message? I think that's how most applications do work that way.

Would probably need to have an option to explicitly target all, like `"*"`.

kchibisov (changes requested):

kchibisov:
Also, I think we need some way to reset to the default config somehow, likely via binding or some other command, so it'll use the values from the file.

alacritty/src/cli.rs
chrisduerr: I'm not sure if the most likely usecase is to change all terminals, or just one? Based on that I would choose the default. I've chosen changing all here, but I'm not sure.

kchibisov: I think just one? If you want all you change the config most of the time? We could have an alias to target all though?

chrisduerr: Yeah the main problem is around providing a sensible and nicely documented way to do it for all windows. I'm not really a fan of `"*"` or a dedicated flag, but I don't have any better ideas either.

chrisduerr:
@kchibisov I've decided to go with `-w -1` to apply to all windows, I've chosen this since `-w *` would create annoying issues due to shell expansion and it's much easier to parse/handle than `-w all` where we'd have to write a custom validator to ensure our errors are descriptive enough. Let me know if you think we should do something better.

alacritty/src/ipc.rs
```diff
@@ -62,6 +63,14 @@ pub fn spawn_ipc_socket(options: &Options, event_proxy: EventLoopProxy<Event>) -
                     let event = Event::new(EventType::CreateWindow(options), None);
                     let _ = event_proxy.send_event(event);
                 },
+                SocketMessage::Config(ipc_config) => {
+                    let window_id = ipc_config
+                        .window_id
+                        .filter(|id| id >= &0)
+                        .map(|id| WindowId::from(id as u64));
```
kchibisov, line 70: That looks like `try_into` and `flat_map`.

alacritty/src/window_context.rs
```diff
@@ -109,7 +115,12 @@ impl WindowContext {
         // The PTY forks a process to run the shell on the slave side of the
         // pseudoterminal. A file descriptor for the master side is retained for
         // reading/writing to the shell.
-        let pty = tty::new(&pty_config, display.size_info.into(), display.window.x11_window_id())?;
+        let pty = tty::new(
+            &pty_config,
+            display.size_info.into(),
+            u64::from(display.window.id()),
```
kchibisov, line 121: Do we need from here and not `into()`? I think the most common is `into`? Besides that we have 2 arguments doing the same thing...

alacritty/src/window_context.rs
```diff
@@ -167,33 +182,49 @@ impl WindowContext {
     }
 
     /// Update the terminal window to the latest config.
-    pub fn update_config(&mut self, old_config: &UiConfig, config: &UiConfig) {
-        self.display.update_config(config);
-        self.terminal.lock().update_config(&config.terminal_config);
+    pub fn update_config(&mut self, new_config: Rc<UiConfig>) {
+        let old_config = mem::replace(&mut self.config, new_config);
+
+        // Apply ipc config if there are any overrides.
```
kchibisov, line 188:
```suggestion
        // Apply ipc config if there are overrides.
```

It kind of reads redundant, but I don't have a strong opinion on this one.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -148,7 +148,12 @@ fn default_shell_command(pw: &Passwd<'_>) -> Command {
 }
 
 /// Create a new TTY and return a handle to interact with it.
-pub fn new(config: &PtyConfig, window_size: WindowSize, window_id: Option<usize>) -> Result<Pty> {
+pub fn new(
+    config: &PtyConfig,
+    window_size: WindowSize,
+    alacritty_window_id: u64,
+    x11_window_id: Option<usize>,
```
kchibisov, line 155: Aren't those the same? The window `id` x11 is using is basically winit window id?

extra/alacritty-msg.man
```diff
@@ -10,20 +10,43 @@ making it possible to control Alacritty without directly accessing it.
...
+Configuration file options [example: cursor.style=Beam]
+.RE
+.TP
+.SH "\tOPTIONS"
+.RS 12
+.TP
+\fB\-w\fR, \fB\-\-window\-id\fR <WINDOW_ID>
+Window ID for the new config.
+
+Use `-1` to apply this change to all windows.
```
kchibisov, line 46: What's the idea behind `-1` and not `-a`, the `-a` could stand for `all`, but one is just one? Kind of strange choice.

Ah, so it seems like you're using it as a negative value, but that would mean that it won't work given that your window id could be `u64` that can't feat into `i64`.

An an option if you really want to keep `-1` you should likely use `i128` for that, so it can hold u64.

alacritty/src/cli.rs
```diff
@@ -294,6 +301,25 @@ pub struct WindowOptions {
...
+
+    /// Window ID for the new config.
+    ///
+    /// Use `-1` to apply this change to all windows.
+    #[clap(short, long, allow_hyphen_values = true, env = "ALACRITTY_WINDOW_ID")]
+    pub window_id: Option<i64>,
+
+    /// Clear all runtime configuration changes.
+    #[clap(short, long, conflicts_with = "options")]
+    pub reset: bool,
```
kchibisov, line 320: I was thinking more of having a dedicated binding rather than IPC, but I don't really mind either way.

kchibisov (changes requested):

extra/alacritty-msg.man
kchibisov: It's just happened that you have a u32 under the hood on X11, because XID is u32. However on Wayland we just take a pointer, and pointers have pretty high values.

alacritty/src/ipc.rs
chrisduerr: `flat_map` is not a method on `Option`.

alacritty/src/window_context.rs
chrisduerr: I guess that's a good point, I forgot for a second that we're using Winit's window ID now which is the same as the X11 window id.

alacritty_terminal/src/tty/unix.rs
chrisduerr: They are!

extra/alacritty-msg.man
chrisduerr: > What's the idea behind -1 and not -a, the -a could stand for all, but one is just one? Kind of strange choice.

Much easier to parse and handle. I feel like for the user ultimately it doesn't matter, but if we accept both strings and numbers we have to handle both clap and serde serializers/deserializers for a type that *may* be both. Dealing with potential number parsing crap all over.

> Ah, so it seems like you're using it as a negative value, but that would mean that it won't work given that your window id could be u64 that can't feat into i64.

I'm not sure window ID on unix will ever actually be over i64? That would be quite a massive number already. Going for `i128` just seemed excessive?

> It's just happened that you have a u32 under the hood on X11, because XID is u32. However on Wayland we just take a pointer here, and pointers have pretty high values.

Pretty high values, sure. But > i64 high values? I guess if it's really actual pointers it might be up to u64?

alacritty/src/cli.rs
chrisduerr: Resetting through IPC seems to make sense when it is set through IPC. This way people can easily have both.

alacritty/src/ipc.rs
kchibisov: Oh, that's sad. Since you have `flatten` and `map` I thought that it had `flat_map`?

You can still do something with `try_into` and `ok`, but the way it's fine then.

chrisduerr: Switched to `and_then`.

chrisduerr:
What's your problem with that clap error @kchibisov? It seems sensible to me?

kchibisov (approved):

paum3:
> $ALACRITTY_WINDOW_ID

Hi, how I get the $ALACRITTY_WINDOW_ID value? Thank you!

kchibisov:
It's in `env`.

paum3:
How can I change some parameter eg. font.size from one window to another ? Just setting ALACRITTY_WINDOW_ID value from the other window, does not work.

kchibisov:
They should be opened via `CreateNewWindow` command, not via spawning yet another instance, so they share the same socket.

chrisduerr:
Even when you spawn another instance you can change the config using the `--socket` and `--window-id` flags. You just need to read those environment variables for the target window.

paum3:
Thank you guys,
for some reason I am able to change font.size. When I've tried with other
options eg. colors.primary.background='#0000fe' I got an error: Invalid IPC
config value.
What am I doing wrong?

thanx!

chrisduerr:
You need to quote twice, to indicate it's a string. Like `'"#0000fe"'`.

aswadsun:
I want to run script to create new window and then set some option for that new window inside the same script, for example:
```
alacritty msg create-window
alacritty msg config "font.size=12" -w $NEW_WINDOW_ID
```

After creating new window, how do I get the newly creating alacritty window Id without looking in the window $ALACCRITY_WINDOW_ID manually?

Thank you.
Regards,
Aswad

kchibisov:
Not sure it's possible unless you guess somehow the pid of the window. The alternative is to not use window but instance and pass all you want with `-o`. What we want is to have `-o` for a `create-window` as well.

paum3:
I was measuring what is faster today,  changing  font sizes:
1.  via msg config,
2.  via updating conifig file (by simple program written in c) 

And the conclusion is,:
1. both programs need approx the same time for that, 
but 

there is a big latency (few secs) until the new font size rendered itself , for example from ```normal.font.size=12 ``` to ```normal.font.size=1200``` takes too much time.

rieje:
`$ALACRITTY_SOCKET` is not intended to be a environment variable nor can it be used by multiple instances of Alacritty, right?

Trying to set settings for all windows including new ones--the following is close except it doesn't apply to new windows (as expected since it loads the default Alacritty config by default and accepts IPC message):

    # Wrapper for launching Alacritty
    ! /usr/bin/alacritty msg create-window "$@" && /usr/bin/alacritty "$@"

Use cases are toggling terminal transparency, font sizing, etc.

chrisduerr:
> $ALACRITTY_SOCKET is not intended to be a environment variable nor can it be used by multiple instances of Alacritty, right?

Why wouldn't it be? If we didn't intend to set an environment variable, we wouldn't. One socket corresponds to one instance, but sockets can be used by any instance.

rieje:
> > $ALACRITTY_SOCKET is not intended to be a environment variable nor can it be used by multiple instances of Alacritty, right?
> 
> Why wouldn't it be? If we didn't intend to set an environment variable, we wouldn't. One socket corresponds to one instance, but sockets can be used by any instance.

Whoops I set that to a path in shell config and assumed Alacritty would use that instead.

Anyway, I just copy the config to tmpfs which Alacritty will use and use a script to modify the appropriate settings, taking advantage of the live reload as a solution. I assume there aren't good reasons to have new windows inherit existing in-memory settings as an alternative to the config.

PR #6175, opened 2022-06-30
Use timer based rendering instead of vsync

kchibisov:
Given that alacritty handles all the windows on the main thread using
vsync will result in sequential locks on swap buffers for each window
it'll try to draw at the given time, reducing the frame rate by the
amount of windows being drawn.

Timer based solution solves this since there's no more blocking
involved.

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -2,8 +2,8 @@
 //! GPU drawing.
 
 use std::fmt::{self, Formatter};
-#[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
 use std::sync::atomic::Ordering;
+use std::time::Duration;
 use std::{cmp, mem};
 
 use glutin::dpi::PhysicalSize;
@@ -17,7 +17,7 @@ use log::{debug, info};
 use parking_lot::MutexGuard;
 use serde::{Deserialize, Serialize};
 #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
-use wayland_client::EventQueue;
+use wayland_client::{protocol::wl_surface::WlSurface, Attached, EventQueue};
 
 use crossfont::{self, Rasterize, Rasterizer};
 
@@ -42,23 +42,24 @@ use crate::display::content::RenderableContent;
 use crate::display::cursor::IntoRects;
 use crate::display::damage::RenderDamageIterator;
 use crate::display::hint::{HintMatch, HintState};
-use crate::display::meter::Meter;
+use crate::display::meter::{Meter, Sample};
 use crate::display::window::Window;
-use crate::event::{Mouse, SearchState};
+use crate::event::{Event, EventType, Mouse, SearchState};
 use crate::message_bar::{MessageBuffer, MessageType};
 use crate::renderer::rects::{RenderLines, RenderRect};
 use crate::renderer::{self, GlyphCache, Renderer};
+use crate::scheduler::{Scheduler, TimerId, Topic};
 use crate::string::{ShortenDirection, StrShortener};
 
 pub mod content;
 pub mod cursor;
 pub mod hint;
+pub mod meter;
 pub mod window;
 
 mod bell;
 mod color;
 mod damage;
-mod meter;
 
 /// Label for the forward terminal search bar.
 const FORWARD_SEARCH_LABEL: &str = "Search: ";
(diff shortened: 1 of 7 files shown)


kchibisov:
It was working quite good in Xwayland, though, not sure how it goes in a multi monitor environment.

chrisduerr:
While this does improve the rendering rate with 4 windows from ~35FPS to 40-60 FPS, it also decreases the rendering rate with a single window from a steady 144FPS to 60-100FPS. So I think something's not working quite right.

chrisduerr (changes requested):
See my comment, I don't think this is working properly.

kchibisov:
Could you retest?

kchibisov:
Ah, shouldn't work either. The problem is that with high refresh rate you have less time in general and this time you're spending on processing the event queue as well. So to solve this we should account for the event processing as well, not just for rendering.

In general, it's kind of impressing that event processing takes similar amount of time compared to rendering, especially given that we don't have any interactions. Maybe we're flooding a lot or winit on x11 is slow in general?

Besides that, the real solution is vsync source in winit, however we'd need some sort of fallback. In the mean time I'll try to account for the event processing as well and see how it goes.

chrisduerr:
> In general, it's kind of impressing that event processing takes similar amount of time compared to rendering, especially given that we don't have any interactions. Maybe we're flooding a lot or winit on x11 is slow in general?

I mean we are sending about a billion `Wakeup`s through the event loop when running `yes`, but that isn't limited to X11.

kchibisov:
Could you retest in a mixed refresh rate environment? I think it should work now, at least it can maintain the frame rate.

chrisduerr:
Without mixed refresh rate, I've noticed that this patch takes significantly longer to reach the desired framerate than the old version. Which might be problematic since having to wait *seconds* before getting the full refresh rate might not give you your full performance with short-lived processes.

tmp

tmp

Running with three windows, one on a 60Hz shows some significant instabilities on the 144Hz monitors:

tmp

Though it is steady on the 60Hz one:

tmp

Running three 144Hz windows shows the same results, indicating that it cannot handle three windows at the same time running at 144Hz, while mixing 144/60Hz shouldn't be an issue:

tmp

Sustaining 144 FPS with three separate instances on three 144Hz monitors is not an issue:

tmp

So the massive fluctuations and significant drop in performance is certainly not due to hardware limitations.

alacritty/src/display/mod.rs
```diff
@@ -928,10 +933,15 @@ impl Display {
...
-        // `commit`, which is done by swap buffers under the hood.
-        #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
-        self.request_frame(&self.window);
+        self.meter.add_sample(sample);
+
+        self.draw_render_timer(config);
+
+        // Request the next frame for the given window.
+        //
+        // On Waylad frame event should be requested before swaping buffers, since it requires
```
trimental, line 942: Typo of wayland

chrisduerr: One hell of a lad.

trimental: Practically a full grown man

trimental:
Sorry for chiming in with an idea on something I know little about; but could you do timer based rendering by having a vector of each unique monitor refresh rate and references to the `has_frame` atomic bool of each window using that rate. So `(16ms, [has_frame#1, has_frame#3, has_frame#5])` and `(8ms, [has_frame#2, has_frame#4])`, and then wake up on events for an interval of 16/3 ms and 8/2 ms and iterate through the atomic bools setting them true. 

At least that way each redraw is spaced as far as possible from the last. Won't the windows naturally just desynchronise into randomly spaced redraws relative to one another no matter what the offset is if you set the interval per window?

kchibisov:
The timer is per window with the current monitor rate. So other windows shouldn't matter.

trimental:
Perhaps the reason why it takes time to reach the desired frame rate is because it works better when the windows intervals become desynchronised?

trimental:
How do you know that each window is drawing at a time when other windows aren't though. I thought that's what causes the drop in fps.

trimental:
Sorry I just realised that it's the swap_buffers function that blocks for vsync. I thought they were locked in a shared state between windows. 🤦

kchibisov:
Closing in favor of #6585.

PR #6177, opened 2022-07-01
On macOS and X11 do not render when the window is occluded for performance

trimental:
Rendering when the window is not visible due to it being closed, minimised, occluded by other windows, etc; is extremely cpu/gpu intensive with no benefit to the user.

By implementing occlusion detection, my simple bash script that prints numbers every 0.1 seconds went from ~10% cpu usage to about ~0.4% cpu usage when in the background. I've submitted a winit pull request but I thought I might as well get in early to show off the performance/efficiency benefits.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1204,10 +1204,6 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                     },
                     WindowEvent::Occluded(occluded) => {
                         *self.ctx.occluded = occluded;
-                        if !occluded {
-                            // Render as soon as the window is visible once again.
-                            *self.ctx.dirty = true;
-                        }
                     },
                     WindowEvent::DroppedFile(path) => {
                         let path: String = path.to_string_lossy().into();
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -192,6 +192,7 @@ pub struct ActionContext<'a, N, T> {
     pub search_state: &'a mut SearchState,
     pub font_size: &'a mut Size,
     pub dirty: &'a mut bool,
+    pub occluded: &'a mut bool,
     pub preserve_title: bool,
     #[cfg(not(windows))]
     pub master_fd: RawFd,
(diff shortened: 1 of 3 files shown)


trimental:
Added X11 support

kchibisov:
@trimental Thanks for the PR, but I'd suggest to hold until #6079 since it contains all the changes required for upstream winit. I can apply your patch later, since we'll have rendering related patches as well and they affect the `should_draw` logic.

christianparpart:
FYI: Looks like `Occluded` also means when another window is above your window. I was checking the docs, but I may be wrong, so please correct me then.

Now, from first sight, it might not make sense to render. But what iff the window that's occluding you is actually semi-transparent. Then you'd show stale content. Having semi-transparent windows in 2022 isn't actually a too rare condition, because users seem to like that and configure their desktop as such (e.g. for file managers, or actually terminals, with and without blur effect).

If the window is actually hidden/symbolized, such an optimization makes 100% sense, but the occlusion due to other windows laying on top of you, I fear will generate confusion.

trimental:
@christianparpart Yes you're completely right in all of those assumptions, ie we still need to render when the compositor is presenting previews or behind semi transparent windows.

On macOS both of these are 100% taken into account, so the window is only marked as occluded if it completely cannot be seen in any way. On x11 its a bit harder to know whether all compositors will remember to take care of this, but considering the x11 docs.

```
When the window changes state from viewable and completely unobscured, viewable and partially 
obscured, or not viewable to viewable and fully obscured, the X server generates the event with the 
state member of the XVisibilityEvent structure set to VisibilityFullyObscured.
```

I would definitely interpret that to mean all x11 compositors should account for transparency and previewing, although I haven't fully tested different X11 compositors

trimental:
@kchibisov done :)

chrisduerr (approved):

alacritty/src/event.rs
```diff
@@ -1201,6 +1202,13 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                         self.ctx.update_cursor_blinking();
                         self.on_focus_change(is_focused);
                     },
+                    WindowEvent::Occluded(occluded) => {
+                        *self.ctx.occluded = occluded;
+                        if !occluded {
+                            // Render as soon as the window is visible once again.
+                            *self.ctx.dirty = true;
+                        }
```
kchibisov, line 1210: I'm pretty sure this not required. Since to reset `dirty` you must be not `occluded`. Is it causing issues when the window wasn't changed while being occluded and then present to the display?

kchibisov (changes requested):

alacritty/src/event.rs
trimental: Yes sorry I realised it probably redundant last night and was going to remove it. For some reason my brain thought if it can now be seen then it needs to redraw but that obviously isn't the case. I'll remove it.

kchibisov (approved):

PR #6186, opened 2022-07-06
Fix thin strokes on macOS

### LINKED ISSUE #4616: [Mac Big Sur] 'use_thin_strokes: true' appears to no longer work
Hello Alacritty team,

I was a very happy Alacritty on Catalina user, especially the `use_thin_strokes` feature.

I upgraded to Big Sur and Alacritty's `use_thin_strokes` feature no longer has an effect, setting it to true or false results in no difference (at least with my setup). Note, I did restart Alacritty when toggling that feature.

I have been using Alacritty 0.6.0.

Has Apple changed APIs or is there something else I need to do at my end?

I really like `use_thin_strokes`, so I am sad that it appears to be gone on Big Sur.

Best regards.

### LINKED ISSUE #6108: Crash on startup (macOS): 'A font must have a non-null family name.'
Alacritty crashes on macOS on startup. No custom `alacritty.yml` is defined.

### System

OS: macOS BigSur 11.6.6 (20G624)
Version: alacritty 0.10.1 (2844606)

### Logs

Crashes:
```
$ alacritty 
thread 'main' panicked at 'A font must have a non-null family name.', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/core-text-19.2.0/src/font_descriptor.rs:223:19
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

$ RUST_BACKTRACE=1 alacritty 
thread 'main' panicked at 'A font must have a non-null family name.', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/core-text-19.2.0/src/font_descriptor.rs:223:19
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic_display
   3: core::option::expect_failed
   4: crossfont::darwin::Descriptor::new
   5: crossfont::darwin::Descriptor::to_font
   6: <crossfont::darwin::Rasterizer as crossfont::Rasterize>::load_font
   7: alacritty::renderer::GlyphCache::load_regular_font
   8: alacritty::display::Display::new
   9: alacritty::event::Processor::create_window
  10: alacritty::main
```

chrisnc:
Remove the `font.use_thin_strokes` config, which only did anything on
macOS and only prior to Big Sur. Instead, we will enable or disable
"font smoothing" on macOS based on the `AppleFontSmoothing` user
default.

These changes let users get the "thin strokes" behavior by setting
`AppleFontSmoothing` to 0 with:

```sh
$ defaults write -g AppleFontSmoothing -int 0
```

(Or replace `-g` with `org.alacritty` to apply this setting only to
Alacritty.app, rather than the whole system.)

Add a `removed` config attribute to show helpful warnings to users
who are using config options that don't do anything anymore, and apply
this attribute to `font.use_thin_strokes`.

Bump `crossfont` to 0.5.0 to pick up the new font smoothing behavior.
This release also includes proportional font support, which just
requires an additional `advance` field in `RasterizedGlyph`, as well as
a fix for a crash when trying to load a disabled font.

Fixes #4616.
Fixes #6108.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,10 +34,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `SpawnNewInstance` no longer inherits initial `--command`
 - Blinking cursor will timeout after `5` seconds by default
 - Deprecated `colors.search.bar`, use `colors.footer_bar` instead
+- On macOS, Alacritty now reads `AppleFontSmoothing` from user defaults to control font smoothing
 
 ### Fixed
 
-- Creating the IPC socket failing if WAYLAND_DISPLAY contains an absolute path
+- Creating the IPC socket failing if `WAYLAND_DISPLAY` contains an absolute path
 - Crash when resetting the terminal while in vi mode
 - `font.glyph_offset` not live reloading
 - Failure when running on 10-bit color system
@@ -50,6 +51,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Window flickering on resize on Wayland
 - Unnecessary config reload when using `/dev/null` as a config file
 - Windows `Open Alacritty Here` on root of drive displaying error
+- On macOS, `font.use_thin_strokes` did not work since Big Sur
+- On macOS, trying to load a disabled font would crash
+
+### Removed
+
+- `font.use_thin_strokes` config field; to use thin strokes on macOS, set
+    `AppleFontSmoothing` to 0 with `$ defaults write -g AppleFontSmoothing -int 0`
 
 ## 0.10.1
 
--- a/alacritty/src/config/font.rs
+++ b/alacritty/src/config/font.rs
@@ -22,6 +22,7 @@ pub struct Font {
     /// Glyph offset within character cell.
     pub glyph_offset: Delta<i8>,
 
+    #[config(removed = "set the AppleFontSmoothing user default instead")]
     pub use_thin_strokes: bool,
 
     /// Normal font face.
@@ -79,8 +80,8 @@ impl Default for Font {
     fn default() -> Font {
         Self {
             builtin_box_drawing: true,
-            use_thin_strokes: Default::default(),
             glyph_offset: Default::default(),
+            use_thin_strokes: Default::default(),
             bold_italic: Default::default(),
             italic: Default::default(),
             offset: Default::default(),
--- a/alacritty_config_derive/src/de_struct.rs
+++ b/alacritty_config_derive/src/de_struct.rs
@@ -135,14 +135,14 @@ fn field_deserializer(field_streams: &mut FieldStreams, field: &Field) -> Result
                     config.#ident = serde::Deserialize::deserialize(unused).unwrap_or_default();
                 });
             },
-            "deprecated" => {
-                // Construct deprecation message and append optional attribute override.
-                let mut message = format!("Config warning: {} is deprecated", literal);
+            "deprecated" | "removed" => {
+                // Construct deprecation/removal message with optional attribute override.
+                let mut message = format!("Config warning: {} has been {}", literal, parsed.ident);
                 if let Some(warning) = parsed.param {
                     message = format!("{}; {}", message, warning.value());
                 }
 
-                // Append stream to log deprecation warning.
+                // Append stream to log deprecation/removal warning.
                 match_assignment_stream.extend(quote! {
                     log::warn!(target: #LOG_TARGET, #message);
                 });
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -383,9 +383,9 @@ dependencies = [
 
 [[package]]
 name = "crossfont"
-version = "0.3.2"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1299695a4c6417b7e4a6957bd963478406e148b7b0351e2f2ce31296b0518251"
+checksum = "f66b1c1979c4362323f03ab6bf7fb522902bfc418e0c37319ab347f9561d980f"
 dependencies = [
  "cocoa",
  "core-foundation 0.9.3",
@@ -397,6 +397,8 @@ dependencies = [
  "freetype-rs",
  "libc",
  "log",
+ "objc",
+ "once_cell",
  "pkg-config",
  "servo-fontconfig",
  "winapi 0.3.9",
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -172,12 +172,6 @@
   #  x: 0
   #  y: 0
 
-  # Thin stroke font rendering (macOS only)
-  #
-  # Thin strokes are suitable for retina displays, but for non-retina screens
-  # it is recommended to set `use_thin_strokes` to `false`.
-  #use_thin_strokes: true
-
   # Use built-in font for box drawing characters.
   #
   # If `true`, Alacritty will use a custom built-in font for box drawing
--- a/alacritty/Cargo.toml
+++ b/alacritty/Cargo.toml
@@ -28,7 +28,7 @@ serde_json = "1"
 glutin = { version = "0.28.0", default-features = false, features = ["serde"] }
 notify = "4"
 parking_lot = "0.11.0"
-crossfont = { version = "0.3.1", features = ["force_system_fontconfig"] }
+crossfont = { version = "0.5.0", features = ["force_system_fontconfig"] }
 copypasta = { version = "0.8.0", default-features = false }
 libc = "0.2"
 unicode-width = "0.1"
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -411,7 +411,7 @@ impl Display {
         // Guess the target window dimensions.
         debug!("Loading \"{}\" font", &config.font.normal().family);
         let font = &config.font;
-        let rasterizer = Rasterizer::new(estimated_scale_factor as f32, font.use_thin_strokes)?;
+        let rasterizer = Rasterizer::new(estimated_scale_factor as f32)?;
         let mut glyph_cache = GlyphCache::new(rasterizer, font)?;
         let metrics = glyph_cache.font_metrics();
         let (cell_width, cell_height) = compute_cell_size(config, &metrics);
@@ -487,10 +487,6 @@ impl Display {
         // Clear screen.
         let background_color = config.colors.primary.background;
         renderer.clear(background_color, config.window_opacity());
-
-        // Set subpixel anti-aliasing.
-        #[cfg(target_os = "macos")]
-        crossfont::set_font_smoothing(config.font.use_thin_strokes);
 
         // Disable shadows for transparent windows on macOS.
         #[cfg(target_os = "macos")]
--- a/alacritty/src/renderer/text/builtin_font.rs
+++ b/alacritty/src/renderer/text/builtin_font.rs
@@ -90,6 +90,7 @@ fn box_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> Raster
                 left: 0,
                 height: height as i32,
                 width: width as i32,
+                advance: (0, 0),
                 buffer,
             };
         },
@@ -479,7 +480,15 @@ fn box_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> Raster
 
     let top = height as i32 + metrics.descent as i32;
     let buffer = BitmapBuffer::Rgb(canvas.into_raw());
-    RasterizedGlyph { character, top, left: 0, height: height as i32, width: width as i32, buffer }
+    RasterizedGlyph {
+        character,
+        top,
+        left: 0,
+        height: height as i32,
+        width: width as i32,
+        advance: (0, 0),
+        buffer,
+    }
 }
 
 #[repr(packed)]
--- a/alacritty/src/window_context.rs
+++ b/alacritty/src/window_context.rs
@@ -210,10 +210,6 @@ impl WindowContext {
             self.display.window.set_title(config.window.identity.title.clone());
         }
 
-        // Set subpixel anti-aliasing.
-        #[cfg(target_os = "macos")]
-        crossfont::set_font_smoothing(config.font.use_thin_strokes);
-
         // Disable shadows for transparent windows on macOS.
         #[cfg(target_os = "macos")]
         self.display.window.set_has_shadow(config.window_opacity() >= 1.0);
--- a/alacritty_config_derive/tests/config.rs
+++ b/alacritty_config_derive/tests/config.rs
@@ -35,6 +35,8 @@ struct Test {
     enom_big: TestEnum,
     #[config(deprecated)]
     enom_error: TestEnum,
+    #[config(removed = "it's gone")]
+    gone: bool,
 }
 
 impl Default for Test {
@@ -48,6 +50,7 @@ impl Default for Test {
             enom_small: TestEnum::default(),
             enom_big: TestEnum::default(),
             enom_error: TestEnum::default(),
+            gone: false,
         }
     }
 }
@@ -90,6 +93,7 @@ fn config_deserialize() {
         enom_small: "one"
         enom_big: "THREE"
         enom_error: "HugaBuga"
+        gone: false
     "#,
     )
     .unwrap();
@@ -101,6 +105,7 @@ fn config_deserialize() {
     assert_eq!(test.enom_small, TestEnum::One);
     assert_eq!(test.enom_big, TestEnum::Three);
     assert_eq!(test.enom_error, Test::default().enom_error);
+    assert_eq!(test.gone, false);
     assert_eq!(test.nesting.field1, Test::default().nesting.field1);
     assert_eq!(test.nesting.field2, None);
     assert_eq!(test.nesting.field3, Test::default().nesting.field3);
@@ -116,8 +121,9 @@ fn config_deserialize() {
     ]);
     let warn_logs = logger.warn_logs.lock().unwrap();
     assert_eq!(warn_logs.as_slice(), [
-        "Config warning: field1 is deprecated; use field2 instead",
-        "Config warning: enom_error is deprecated",
+        "Config warning: field1 has been deprecated; use field2 instead",
+        "Config warning: enom_error has been deprecated",
+        "Config warning: gone has been removed; it's gone",
     ]);
 }
 


chrisnc:
See here for what this looks like: https://github.com/alacritty/crossfont/pull/47#issuecomment-1175947714

chrisnc:
I confirmed that doing `defaults write io.alacritty AppleFontSmoothing -int 0` will control this behavior for people who want thin strokes in Alacritty only, but the default rendering (slightly bolder) in the rest of the system.

chrisnc:
I think this is ready for review now. To actually work it will require releasing `crossfont` v0.5.0 and `cargo update -p crossfont` to pick up its new dependencies (`objc` and `once_cell`) in `Cargo.lock`.

I have been running alacritty with these changes for a little while now, and I just updated with the latest `master` of `crossfont`. Everything looks good, and "smoothing" toggles on and off when I expect it to.

chrisduerr:
@chrisnc New crossfont version should be out now.

chrisnc:
Updated with released `crossfont` v0.5.0. Ready for reviews.

CHANGELOG.md
```diff
@@ -44,6 +45,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Character mappings in the DEC special graphics character set (line drawing)
 - Window flickering on resize on Wayland
 - Unnecessary config reload when using `/dev/null` as a config file
+- On macOS, `use_thin_strokes` did not work since Big Sur
+
+### Removed
+
+- Config field `use_thin_strokes`, use `defaults write -g AppleFontSmoothing -int 0` on macOS instead
```
kchibisov, line 52:
```suggestion
- Config field `use_thin_strokes`, use `defaults write -g AppleFontSmoothing -int 0` on macOS instead
```
```suggestion
- Config field `font.use_thin_strokes`, use `defaults write -g AppleFontSmoothing -int 0` on macOS instead
```

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Vi mode keybinding (z) to center view around vi mode cursor
 - Accept hexadecimal values starting with `0x` for `--embed`
 - Config option `cursor.blink_timeout` to timeout cursor blinking after inactivity
+- On macOS, use the `AppleFontSmoothing` user default to decide whether fonts should be "smoothed"
```
kchibisov, line 25: I'd probably put it into changed in the following wording:

"On macOS `AppleFontSmoothing` user default is now being used to control font smoothing"

kchibisov (changes requested):
I'm not sure whether we should actually warn on the old config option.

In general we could throw a warning when the old config option is used and advice to use `AppleFontSmoothing` for that user option. 

What do you think @chrisduerr ?

CHANGELOG.md
chrisnc: Reworded.

CHANGELOG.md
chrisnc: Fixed.

chrisnc:
I've reworded the changelog entry and commit message. I am definitely in favor of emitting a helpful warning or deprecation message for people who have `font.use_thin_strokes` set in their `alacritty.yaml`, but I'm not sure the best way to do that.

chrisnc:
Pushed a separate commit with the deprecation warning. This is how it looks for me:
deprecation warning for font.use_thin_strokes
Let me know what you think.

chrisnc:
Squashed and updated commit message and changelog to reflect deprecation of `font.use_thin_strokes`.

chrisnc:
Fixed some wording and alignment in the changelog, and the ordering of some initializers... I think this is almost to go, pending others' thoughts on the deprecation warning.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Vi mode keybinding (z) to center view around vi mode cursor
 - Accept hexadecimal values starting with `0x` for `--embed`
 - Config option `cursor.blink_timeout` to timeout cursor blinking after inactivity
+- On macOS, Alacritty now reads `AppleFontSmoothing` from user defaults to control font smoothing
```
kchibisov, line 25: Could you move it to `CHANGED`?

CHANGELOG.md
```diff
@@ -44,6 +45,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Character mappings in the DEC special graphics character set (line drawing)
 - Window flickering on resize on Wayland
 - Unnecessary config reload when using `/dev/null` as a config file
+- On macOS, `font.use_thin_strokes` did not work since Big Sur
+
+### Removed
+
+- Deprecated `font.use_thin_strokes` config field; to use thin strokes on macOS, set
```
kchibisov, line 52:
```suggestion
- `font.use_thin_strokes` config field; to use thin strokes on macOS, set
```

alacritty/src/config/font.rs
```diff
@@ -22,6 +22,11 @@ pub struct Font {
     /// Glyph offset within character cell.
     pub glyph_offset: Delta<i8>,
 
+    #[config(deprecated = "To enable thin strokes on macOS, run:\n$ defaults write <domain> \
+                           AppleFontSmoothing -int 0\nwhere <domain> is -g (for system-wide) or \
+                           io.alacritty (for just Alacritty).\nThen reboot your system or \
+                           relaunch Alacritty for the changes to take effect.\nTo disable thin \
+                           strokes, run:\n$ defaults delete <domain> AppleFontSmoothing")]
```
kchibisov, line 29: This is overly verbose and I'm pretty sure we should shorten it. It's also not a deprecated setting, but entirely removed one.

I'd suggest to alter serialization and have a `removed` option (look for "deprecated" in `alacritty_config_derive`.)

The resulted message should be more like
```rust
#config(removed = "The behavior of font smoothing is now controlled with AppleFontSmoothing user default.")]
```

kchibisov (changes requested):

alacritty/src/config/font.rs
chrisnc: Sure; I will trim it a bit and add a `removed` attribute.

CHANGELOG.md
chrisnc: I could, but I put it in `Added` in `crossfont`. I think it makes more sense as an `Added` and a `Removed`. The only `Changed` in `crossfont` was a rename.

chrisnc:
I noticed that previous config deprecations were put into the "Removed" section of the changelog, but the one that just merged is in "Changed". Should it move?

kchibisov:
@chrisnc it could be not entirely clear, but removed means that deprecated option was entirely removed. So it all makes sense.

In that particular case you're just removing without deprecation, so it should be straight up remove without saying that it's deprecated, since it was never deprecated before, and won't be since it'll basically stop working.

That's why I'm saying that it should be `removed`, not `deprecated` as you've originally started here.

CHANGELOG.md
kchibisov: For alacritty it's more of a `CHANGED`, since it was using thin strokes before, but via config option. Now you'll have to use apple user default.

Though. I'm fine with keeping it in added.

chrisnc:
Oh! I was reading the "Deprecated" in those older changelog entries as a verb (meaning it was _just_ deprecated), rather than a past participle, meaning it was previously deprecated and is now removed... It all makes sense now. Could I suggest using "Deprecate" (present tense) for the most recent one in the "Changed" section? This removes the ambiguity.

chrisnc:
> @chrisnc it could be not entirely clear, but removed means that deprecated option was entirely removed. So it all makes sense.

Oh! I was reading the "Deprecated" in those older changelog entries as a verb (meaning it was _just_ deprecated), rather than a past participle, meaning it was previously deprecated and is now removed... It all makes sense now. Could I suggest using "Deprecate" (present tense) for the most recent one in the "Changed" section? This removes the ambiguity.

> In that particular case you're just removing without deprecation, so it should be straight up remove without saying that it's deprecated, since it was never deprecated before, and won't be since it'll basically stop working.
> 
> That's why I'm saying that it should be `removed`, not `deprecated` as you've originally started here.

Yep understood. Changes to this effect are inbound.

alacritty/src/config/font.rs
chrisnc: I'd still like to put the command users need to run in the warning. Not everyone knows what a "user default" is.

CHANGELOG.md
chrisnc: I'll leave it as-is. The default for alacritty was `false`, contra the comment in the `alacritty.yaml`, and it's been not working for two years anyway. Might as well be "added" for most users. ;)

CHANGELOG.md
chrisnc: Fixed.

alacritty/src/config/font.rs
chrisnc: Here is take 2.
use_thin_strokes removal warning

chrisnc:
Ready for review. Latest changes:
 - do a "removal" instead of a "deprecation"
 - add a `removed` attribute to config deserialization and add to the corresponding test
 - shorten the warning message

CHANGELOG.md
```diff
@@ -24,18 +24,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Config option `cursor.blink_timeout` to timeout cursor blinking after inactivity
 - Escape sequence to set hyperlinks (`OSC 8 ; params ; URI ST`)
 - Config `hints.enabled.hyperlinks` for hyperlink escape sequence hint highlight
+- On macOS, Alacritty now reads `AppleFontSmoothing` from user defaults to control font smoothing
+- A `removed` config attribute, to show warnings for config fields that have been removed
```
kchibisov, line 28: This doesn't make any sense for the end users, so could remove it.

kchibisov (changes requested):

CHANGELOG.md
chrisnc: Sure.

alacritty/src/config/font.rs
```diff
@@ -22,6 +22,9 @@ pub struct Font {
     /// Glyph offset within character cell.
     pub glyph_offset: Delta<i8>,
 
+    #[config(removed = "To enable thin strokes on macOS, run:\n$ defaults write -g \
+                        AppleFontSmoothing -int 0\nand relaunch Alacritty. Use io.alacritty \
+                        instead of -g to enable thin strokes in Alacritty only.")]
```
kchibisov, line 27: I don't think you'd need new lines for those actually. I'm pretty sure that our message bar will format everything properly in that case. Could you try it and see how it looks? Also, I don't think we should document how to use it right away. I'd suggest to mention the option that should be used, not how it should be used.

chrisnc: I tried doing it a few different ways, but I also have to get it to pass `rustfmt`, so I need to use the `\` line continuations in order to align the start of each line of the string literal, which forces the use of `\n` for actual newlines in the message.

chrisnc: My concern is that people will not know where to look, and at least some of them may continue to file duplicate issues or pile on to existing ones if they (reasonably) try to set it to `false`/`NO`, or can't figure out how the `defaults` command works.

chrisnc: If I remove the bit about `io.alacritty` vs. `-g` do you think it would be sufficiently short?

kchibisov: if you run `cargo +nightly fmt` won't it format it?

kchibisov: >I tried doing it a few different ways, but I also have to get it to pass rustfmt, so I need to use the \ line continuations in order to align the start of each line of the string literal, which forces the use of \n for actual newlines in the message.

I thought that our message bar will add those automatically or expand across the strings? I'm not sure you'd need an explicit `\n` for those.

chrisnc: Yes, my changes now are `fmt`'ed. My previous attempts failed the build because they did not pass `rustfmt --check`. I'm not quite sure what you'd like me to test regarding newlines. The message now renders as I expect it to, but if I remove the newlines, it'll all be on one line and will wrap differently depending on the window width.

chrisnc: How about this?
```rs
    #[config(removed = "To enable thin strokes on macOS, run:\n\
                        $ defaults write -g AppleFontSmoothing -int 0\n\
                        and relaunch Alacritty.")]
```

kchibisov: >The message now renders as I expect it to, but if I remove the newlines, it'll all be on one line and will wrap differently depending on the window width, and likely splits the command across lines.

Exactly, your case could generate arbitrary new lines, I'd suggest to do that without `\n` and suggest to use `org.alacritty` for settings instead of making it global. After that it should be good to go.

chrisnc: > I thought that our message bar will add those automatically or expand across the strings? I'm not sure you'd need an explicit \n for those.

How do I tell it where the line breaks should go? The warning log is just one string, and it will be one line if there are no newlines. Here is what that looks like for me, with the following config attribute:
```rs
    #[config(removed = "To enable thin strokes on macOS, run: `$ defaults write -g \
                        AppleFontSmoothing -int 0` and relaunch Alacritty.")]
```
removal_warning_oneline

chrisnc: > Exactly, your case could generate arbitrary new lines, I'd suggest to do that without \n and suggest to use org.alacritty for settings instead of making it global. After that it should be good to go.

Oh, I thought you were saying there was some special thing the message bar would do here with actual newlines vs. newline escape sequences. If you just don't want a multi-line log message, that's fine with me, but I thought it would be more readable and easier to select if the command is guaranteed to be on its own line, which is achieved by doing explicit newlines.

chrisnc: Take 3. What do you think? This is with 100 columns.
removal_warning_take_3

chrisnc: And with the newline (and using `org.alacritty`, per your recent update).
removal_warning_take_4

chrisnc: Rebased and pushed the latter just now. If you really don't like the explicit newline I can remove it... This whole thing will be removed eventually anyway, though.

kchibisov: >to select if the command is guaranteed to be on its own line, which is achieved by doing explicit newlines.

You can't select on message bar, so it doesn't matter.

kchibisov (approved):

alacritty/src/config/font.rs
chrisnc: > You can't select on message bar, so it doesn't matter.

ooof... whoops. Let me rethink this then. Sorry for the churn here.

chrisnc:
Really appreciate all the feedback! I don't have any other changes I want to make, so up to you.

kchibisov (changes requested):
I completely forgot about it. Could you add a CHANGELOG entry for https://github.com/alacritty/alacritty/issues/6108 given that you bump crossfont that fixes it?

chrisnc:
> I completely forgot about it. Could you add a CHANGELOG entry for #6108 given that you bump crossfont that fixes it?

Done.

CHANGELOG.md
chrisduerr: This should be moved to `CHANGED`.

CHANGELOG.md
```diff
@@ -26,18 +26,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- On macOS, Alacritty now reads `AppleFontSmoothing` from user defaults to control font smoothing
 
 ### Changed
 
 - The `--help` output was reworked with a new colorful syntax
 - OSC 52 is now disabled on unfocused windows
 - `SpawnNewInstance` no longer inherits initial `--command`
 - Blinking cursor will timeout after `5` seconds by default
-- Deprecated `colors.search.bar`, use `colors.footer_bar` instead
+- Deprecate `colors.search.bar`; use `colors.footer_bar` instead
```
chrisduerr, line 37: This change should be reverted, the existing entry is perfectly fine and imo superior.

alacritty_config_derive/src/de_struct.rs
```diff
@@ -135,14 +135,14 @@ fn field_deserializer(field_streams: &mut FieldStreams, field: &Field) -> Result
                     config.#ident = serde::Deserialize::deserialize(unused).unwrap_or_default();
                 });
             },
-            "deprecated" => {
-                // Construct deprecation message and append optional attribute override.
-                let mut message = format!("Config warning: {} is deprecated", literal);
+            "deprecated" | "removed" => {
+                // Construct deprecation/removal message and append optional attribute override.
+                let mut message = format!("Config warning: {} is {}", literal, parsed.ident);
```
chrisduerr, line 140: I don't think "XYZ is removed" is correct. While "is" might be accurate for deprecation, the past tense would need to be used for removal.

alacritty/src/config/font.rs
```diff
@@ -22,6 +22,8 @@ pub struct Font {
     /// Glyph offset within character cell.
     pub glyph_offset: Delta<i8>,
 
+    #[config(removed = "to enable thin strokes, run:\n$ defaults write org.alacritty \
+                        AppleFontSmoothing -int 0")]
```
chrisduerr, line 26: This seems a bit verbose and I feel like it probably doesn't look great in the message bar. Should probably shorten it and remove the suggested command?

chrisduerr (changes requested):

alacritty/src/config/font.rs
kchibisov: It looks like that https://github.com/alacritty/alacritty/pull/6186#discussion_r917448769 I do think that it's long, but I don't care that much.

chrisduerr: I mean you can't copy from the message bar anyway, so what's the point? Just point users to configure it in macOS or omit it entirely.

kchibisov: Yeah, I was suggesting something like `to control font thin strokes use AppleFontSmoothing user default`.

CHANGELOG.md
chrisnc: I'll change it back, but for context, this wording confused me because messages like "Deprecated <foo>[, use <bar> instead]" appear elsewhere in the changelog in "Removed" sections, except those entries refer to the removal of the deprecated option, rather than the deprecation itself. The use of "deprecated" rather than "deprecate" makes it ambiguous because it's both a verb and an adjective, and you need the context of which section it's in to disambiguate. https://github.com/alacritty/alacritty/pull/6186#issuecomment-1179783098

alacritty_config_derive/src/de_struct.rs
chrisnc: It's valid English; "removed" can be a past participle just like "deprecated", but I agree it's awkward and not common usage... How about "has been removed" (present perfect tense)?

alacritty/src/config/font.rs
chrisnc: We had a lot of back and forth on this one. I think it looks decent (screencap in comment), but I will change it to gesture vaguely at `AppleFontSmoothing` instead if that's preferred. https://github.com/alacritty/alacritty/pull/6186#discussion_r917448769

My main motivation in showing a command is that just mentioning the name of the key and the phrase "user default" is probably not sufficient information for many people learn how to enable this, and you might continue to get issue spam about it. macOS does not expose this setting in its UI anymore; you basically have to google for it. On top of that you need to know that it needs to be set specifically to 0 as an integer for it to work.

chrisnc: Here is a shorter one (pushed just now).
shorter_removed_message

alacritty_config_derive/src/de_struct.rs
kchibisov: No, the point was that the format should have `was`, but for deprecation it should have `is`.

So the main issue is that `format!("Config warning: {} was {}..."`, but for deprecated we need `format!` with `is`.

alacritty/src/config/font.rs
chrisnc: Just link back to this thread when someone files an issue complaining that `AppleFontSmoothing` isn't a supported key in `alacritty.yml` after reading this. 😄

alacritty_config_derive/src/de_struct.rs
chrisnc: It both was removed and still is removed. ;) I can change to "was".

CHANGELOG.md
chrisnc: Whoops, didn't see this one since the conversation was still resolved. Moved.

chrisnc:
I believe all pending feedback has been addressed, but I'll leave the conversations open just in case.

alacritty/src/config/font.rs
chrisnc: Current state as of 18020cd:
was_removed

alacritty_config_derive/src/de_struct.rs
chrisnc: Now using "was" here.

alacritty_config_derive/src/de_struct.rs
```diff
@@ -147,6 +147,18 @@ fn field_deserializer(field_streams: &mut FieldStreams, field: &Field) -> Result
...
+                let mut message = format!("Config warning: {} was removed", literal);
+                if let Some(warning) = parsed.param {
+                    message = format!("{}; {}", message, warning.value());
+                }
+
+                // Append stream to log removal warning.
+                match_assignment_stream.extend(quote! {
+                    log::warn!(target: #LOG_TARGET, #message);
+                });
+            },
```
chrisduerr, line 161: Duplicating this entire block just to separate between is/was seems pointless, why not just branch again?

chrisduerr (changes requested):

alacritty_config_derive/src/de_struct.rs
chrisnc: Done.

chrisnc:
Any other feedback on this?

alacritty_config_derive/src/de_struct.rs
```diff
@@ -135,14 +135,18 @@ fn field_deserializer(field_streams: &mut FieldStreams, field: &Field) -> Result
...
             },
-            "deprecated" => {
-                // Construct deprecation message and append optional attribute override.
-                let mut message = format!("Config warning: {} is deprecated", literal);
+            "deprecated" | "removed" => {
+                // Construct deprecation/removal message and append optional attribute override.
+                let mut verb = "is";
+                if parsed.ident == "removed" {
+                    verb = "was";
+                }
```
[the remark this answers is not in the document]
chrisnc: Pushed an alternative that also avoids the `mut` and does not duplicate the rest of the `format!` call.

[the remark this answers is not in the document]
chrisnc: `message` still needs to be `mut` because it is re-bound right below in the conditional that adds the warning.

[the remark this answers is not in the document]
chrisnc: Working on removing that `mut` as well...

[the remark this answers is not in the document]
chrisnc: Okay, now setting `message` without any `mut`.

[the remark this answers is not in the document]
chrisduerr: You just made it worse…

The point is to get clean, readable code which doesn't unnecessarily assign variables just to immediately throw away their values. Your new code is completely unreadable.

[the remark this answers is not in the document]
chrisnc: I'm a little confused. Where am I assigning a variable and throwing it away? The existing code in master did that in order to append the optional attribute text, and I removed it. Your suggestion did not compile because it removed the `mut` on `message` without removing this assignment, so I assumed you would prefer an implementation that did not require any re-assignments of bindings, and that is what the new code does. What am I missing here?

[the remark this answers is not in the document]
kchibisov: Apply this.

```diff
diff --git a/alacritty_config_derive/src/de_struct.rs b/alacritty_config_derive/src/de_struct.rs
index 42787365..6b1e29c0 100644
--- a/alacritty_config_derive/src/de_struct.rs
+++ b/alacritty_config_derive/src/de_struct.rs
@@ -137,16 +137,11 @@ fn field_deserializer(field_streams: &mut FieldStreams, field: &Field) -> Result
             },
             "deprecated" | "removed" => {
                 // Construct deprecation/removal message with optional attribute override.
-                let message = format!(
-                    "Config warning: {} {} {}{}",
-                    literal,
-                    if parsed.ident == "removed" { "was" } else { "is" },
-                    parsed.ident,
-                    parsed
-                        .param
-                        .map(|warning| format!("; {}", warning.value()))
-                        .unwrap_or_default()
-                );
+                let message = match parsed.ident.as_str() {
+                    "removed" => format!("Config warning: {} was removed", literal),
+                    "deprecated" => format!("Config warning: {} is deprecated", literal),
+                    _ => unreachable!(),
+                };
 
                 // Append stream to log deprecation/removal warning.
                 match_assignment_stream.extend(quote! {
```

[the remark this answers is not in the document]
chrisnc: If there are further changes you want here I'm happy to make them of course, but I don't have much information to go on. My best guess is that you would prefer separate `let` bindings for some or all of the arguments to this `format!` call. My approach before was to use idioms that were already present in the code, which included mutable bindings and discarding values. If there is some other source of style I should be following, let me know.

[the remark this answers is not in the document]
chrisnc: @kchibisov this patch doesn't include the extra warning message from the attribute. I also was attempting to avoid unnecessary duplication, since the first attempt which left "removed" and "deprecated" in separate top-level arms of the match, was shot down for being duplicative.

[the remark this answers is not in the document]
kchibisov: Could rename `let message` to `message_prefix` and combine message later. Like `let message = format!("{} {}", message_prefix, the_reset_you_want)`.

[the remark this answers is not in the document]
chrisnc: The problem is then you need to set `message` to `message_prefix` if there is no attribute parameter, and the code is more verbose than before. I honestly think the code as-is (with a `mut` and a conditional assignment) for this is cleaner, but I interpreted from previous comments that this `mut` should be deleted as well. I will make another attempt that is more readable.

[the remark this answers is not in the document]
chrisnc: Please see the latest. It's one additional line relative to `master`, to choose "was" for "removed", and puts the names in the format string so there isn't just a "{} {} {}". I'll leave the existing `mut` and suffix append alone. I think it's out of scope of this change and there isn't an obviously better way to do it, IMO. If this still is distasteful for some reason, I'd suggest to reconsider just using "is removed". It's both grammatically and semantically correct for what the warning needs to convey, and avoids unnecessary string splicing.

chrisduerr (approved):

chrisduerr (approved):

bluz71:
Nice work to all involved, this will genuinely be useful to a number of Mac users.

Any plans to start a new stable release cycle? The last stable release 0.10.1 was issued on Feb 21. Since then undercurls, coloured underline/undercurl and now thin strokes have come in, all of which really enhance the user experience.

Cheers.

kchibisov:
We're waiting on winit to release. The alacritty release will be shorty after.

bluz71:
Excellent. Looking forward to that.

ciscohack:
how to remove this config and revert it to default value "defaults write io.alacritty AppleFontSmoothing -int 0"

kchibisov:
try org.alacritty.

ciscohack:
> try org.alacritty.

thanks for the update. so now "defaults write io.alacritty AppleFontSmoothing -int 0" this command will not work on macos Sonoma right? and how we revert back the config to default setting

kchibisov:
I mean, just pass `-int 1`. There's also option to reset, the `defaults` probably has docs.

chrisnc:
@ciscohack to remove the setting entirely and go back to the system's default:

```bash
defaults delete -g AppleFontSmoothing
```

or if you only have it set on `org.alacritty`:

```bash
defaults delete org.alacritty AppleFontSmoothing
```

The value is an integer that can be from 0 to 3, but Alacritty just cares about 0 or nonzero.
For this to take effect you will need to restart Alacritty, or reboot, if you are changing the global setting.

The `AppleFontSmoothing` setting does work on Sonoma. The bundle identifier that Alacritty uses on macOS changed from `io.alacritty` to `org.alacritty` some time ago, so that's where it looks for this setting (it will also see it in the global domain, if there's nothing set in `org.alacritty`).

ciscohack:
@chrisnc thanks a lot for the help and command

PR #6190, opened 2022-07-08
Use DeviceId/WindowId::dummy in tests

5225225:
The use of transmute_copy here is UB because a `0` defaults to `i32`, which has size 4 bytes, and (on my platform), a `glutin::event::DeviceId` is 8 bytes, as it is an enum with 2 variants with the `X11` holding a `c_int`, which is another `i32`. Therefore, the `transmute_copy` reads out of bounds of the input pointer.

I just changed it to use the dummy function. Which is still unsafe, but it has the same effect as the transmute_copy does, making a DeviceId with value 0.

WindowId is similar, but holds a u64 or a usize depending on X11/Wayland. Same fix, made it use `dummy`.

In https://github.com/rust-lang/rust/pull/98839 we're making transmute_copy panic at runtime if `U` is larger than `T`.

### DIFF
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -978,7 +978,8 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
 mod tests {
     use super::*;
 
-    use glutin::event::{Event as GlutinEvent, VirtualKeyCode, WindowEvent};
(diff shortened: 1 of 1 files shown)


kchibisov (approved):

kchibisov:
Should run `cargo +nightly fmt` and it'll be good to go.

PR #6192, opened 2022-07-09
Fix error when opening in root of drive

### LINKED ISSUE #4882: Add "Open Alacritty here" in WIN10 context menu throws error when in root directory with a wired solution.
system: windows10 (without WSL)

I solved the problem with a wired solution but I don't know how I did it.

I want to add "Open Alacritty here" in the context menu (right-click). So I search online and find this:
https://gist.github.com/emlautarom1/5ad137f7e3508efd85ecf611d4dd4f51

```
[HKEY_CLASSES_ROOT\Directory\Background\shell\alacritty]
@="Open Alacritty here"
"Icon"="C:\\programdata\\chocolatey\\bin\\alacritty.exe"

[HKEY_CLASSES_ROOT\Directory\Background\shell\alacritty\command]
@="\"C:\\programdata\\chocolatey\\bin\\alacritty.exe\" \"--working-directory\" \"%V%\""
```

When this added to the registry, the last variable of the key-value is `"%V%"`. I don't understand why there's an extra `%` after `%V`. They seem to perform the same.

However, I face two directory situations now that conflicts with each other. D:\ and D:\Program Files

If I use `"%V"`, right-click open Alacritty here at D:\Program Files and D:\nospace will have no problem. However D:\ directory will cause Alacritty to throw an error of `Invalid working directory: "D:\""`. I guess it is because the backslash`\` make `"` an escape character.

Then I use `%V`, D:\  will have no problem. However, D:\Program Files is a problem and Alacritty will not open here. I think it is because you need "" around D:\Program Files to compromise the space between Program Files.

**Then I try `"%V` and it magically solve the problem. I don't know why.**

WezTerm terminal emulator has the same problem. It uses `"%V\\"` but also can't open at D:\ . However, use `"%V` can also solve the problem.

cmd and Powershell don't have this problem, they are fine with `cmd.exe /s /k pushd "%V"` and `powershell.exe -noexit -command Set-Location -literalPath '%V'` registry key value. They can open in all two directories. I guess it might because of some of those variables that pass to them.

IrishBruse:
Fixes #4882 more elegently than in my previous #6110

When opening root drives eg A:\ C:\ Z:\ etc
an error will appear
image

The above Log File
```
[0.003076400s] [ERROR] [alacritty] Invalid working directory: "Z:\""
```

When running the registry command it appears to escape the " duplicating it

I made a simple test app that just prints the command line args to confirm this and this is the output of the command on the root of the A:\
`--working-directory "A:\""`
An extra " is created 
This appears to be an issue with cmd only which is what the registry seems to run as running this command in cmd does the same thing
image

powershell works fine
image

After trying all that I just removed the commandline arg and it worked fine
as it appears windows sets the working directory to the correct place anyways meaning the arg was redundant and just causing parsing/escape issues i also confirmed this with a test application and it sets the working dir with out having %V

I tested this by just opening registry editor and manually changing it in my currently installed alacritty
image

both the root dir and folders with spaces still work
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Character mappings in the DEC special graphics character set (line drawing)
 - Window flickering on resize on Wayland
 - Unnecessary config reload when using `/dev/null` as a config file
+- Windows `Open Alacritty Here` on root of drive displaying error
 
 ## 0.10.1
 
--- a/alacritty/windows/wix/alacritty.wxs
+++ b/alacritty/windows/wix/alacritty.wxs
@@ -53,7 +53,7 @@
             <!-- Add context menu -->
             <Component Id="ContextMenu" Guid="449f9121-f7b9-41fe-82da-52349ea8ff91">
                 <RegistryKey Root="HKCU" Key="Software\Classes\Directory\Background\shell\Open Alacritty here\command">
-                    <RegistryValue Type="string" Value="[AlacrittyProgramFiles]alacritty.exe --working-directory &quot;%V&quot;" KeyPath="yes"/>
+                    <RegistryValue Type="string" Value="[AlacrittyProgramFiles]alacritty.exe" KeyPath="yes"/>
                 </RegistryKey>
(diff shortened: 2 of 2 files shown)


kchibisov (approved):
I wonder if we need a changelog entry for that?

IrishBruse:
Cant hurt just appended that to my commit there with a force push

kchibisov (approved):

PR #6193, opened 2022-07-10
add "sudo" to Debian/Ubuntu command

AliKZ07:
hello thats need sudo access


kchibisov:
In general users are aware that privileged access is required. `sudo` is just one of the common ways to get it.

PR #6198, opened 2022-07-11
Fix hyperlinks not being keyboard actionable

kchibisov:
This fixes a typo in 694a52b which was filtering all hyperlinks, except
only duplicated ones when opening them with keyboard.

### DIFF
--- a/alacritty_terminal/src/event.rs
+++ b/alacritty_terminal/src/event.rs
@@ -99,7 +99,7 @@ pub trait EventListener {
     fn send_event(&self, _event: Event) {}
 }
 
-/// It listens to the void and tests.
+/// Null sink for events.
 pub struct VoidListener;
 
 impl EventListener for VoidListener {}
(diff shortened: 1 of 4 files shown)


alacritty_terminal/src/event.rs
```diff
@@ -99,6 +99,7 @@ pub trait EventListener {
     fn send_event(&self, _event: Event) {}
 }
 
-/// Placeholder implementation for tests.
-#[cfg(test)]
```
kchibisov, line 103: Having it as `cfg(test)` prevents `alacritty` crate for using it for tests. So I've added a dummy type pure for tests, so we can use `mock_term`. Though, I can move it to `alacritty_terminal::term::tests` so it'll be more clear what it used for.

chrisduerr (changes requested):

alacritty_terminal/src/event.rs
```diff
@@ -99,7 +99,7 @@ pub trait EventListener {
     fn send_event(&self, _event: Event) {}
 }
 
-/// It listens to the void and tests.
+/// Null sink for events.
```
chrisduerr, line 102: Sometimes life just isn't fair. :smile:

chrisduerr (approved):

Issue #4897, opened 2021-03-15
Window shadows is missing in MacOS version


lucastrvsn:
### System

OS: macOS Big Sur 11.2.3 Apple Sillicon
Version: `alacritty 0.8.0-dev (3159b60)` using arm_ci branch

For comparison, this is `Terminal.app` and `alacritty`:

![Captura de Tela 2021-03-15 às 10 20 49](https://user-images.githubusercontent.com/8826449/111159855-4a0f0480-8578-11eb-8e04-ae27d0c139ba.png)

![Captura de Tela 2021-03-15 às 10 20 27](https://user-images.githubusercontent.com/8826449/111159838-45e2e700-8578-11eb-9e7c-0293512d51c6.png)

chrisduerr:
Do you use transparency?

lucastrvsn:
@chrisduerr Yes. Removing the `background_opacity` option from my `alacritty.yml` the shadows appears. Maybe we need some warning on the `alacritty.yml` template config file to it?

chrisduerr:
I personally don't really think a shadow is significant enough to justify that, but if you want to send a PR yourself feel free.

PR #6200, opened 2022-07-11
issue #4897; force shadow even if opacity is assigned.

sarashino:
#4897


chrisduerr (changes requested):
IIRC the reason why shadow is disabled with transparent windows is because it's buggy in macOS. So unfortunately Alacritty doesn't have much of a choice here.

And I definitely don't have any interest adding a useless config option for people to opt-into broken behavior. If that's desired, people can just compile things from source.

@kchibisov Do you know if the transparency bug is still present with the latest macOS version?

kchibisov:
>@kchibisov Do you know if the transparency bug is still present with the latest macOS version?

I'd try to take a look, but I guess it's still there, since this bugs chrome as well iirc.

kchibisov:
I've tested on macOS and shadow is still broken. The main issue is that it shows the back buffer with it from time to time, and some artifacts. So I don't think there's a way to have it for transparent windows unless it'll work without bugging.

Issue #6178, opened 2022-07-04
Having specific strings colored by default


Ok-pass:
Would it be possible to implement a color setting in hints.regex or elsewhere, that would allow regex strings to always appear highlighted? 
For example something like
```
hints:
  enabled:
  - regex: "(https:|http:)\
            [^\u0000-\u001F\u007F-\u009F<>\"\\s{-}\\^⟨⟩`]+"
    color:
      background: '#3388ff'
      foreground: '#000000'
```
to have URLs colored. 
I'm looking for a way to quickly spot full URL paths in large amounts of text.

An alternative is to have the strings matched by a regex highlighted when in Hints mode, similar to how it is in Search mode.

System
OS: Linux
Version: alacritty 0.10.1

chrisduerr:
I don't think it's a good idea to always highlight regex matches for performance reasons.

> An alternative is to have the strings matched by a regex highlighted when in Hints mode, similar to how it is in Search mode.

Isn't this already the case? When you start hint mode, all matches should be highlighted.

Ok-pass:
> When you start hint mode, all matches should be highlighted.

As long as I'm not missing something: only the hints labels are highlighted, not the strings themselves:
![3](https://user-images.githubusercontent.com/60102027/177080762-f3e2f0f7-1bde-4a19-892a-7a7c4491ffd4.png)

Unlike in Search mode where the full string is colored:
![2](https://user-images.githubusercontent.com/60102027/177081647-ec8ffe5f-24fa-4fa4-8830-4a2bf31bcc76.png)

colors.hints only has settings for:

- start # First character in the hint label
- end # All characters after the first one in the hint label

My full colors section, if it would help.
```
colors:
  primary:
    background: '#050505'
    foreground: '#eeeeee'
  cursor:
    text: '#080808'
    cursor: '#33ee33'
  vi_mode_cursor:
    text: '#080808'
    cursor: '#33ee33'
  search:
    matches:
      foreground: '#080808'
      background: '#3388ff'
    focused_match:
      foreground: '#080808'
      background: '#33ff88'
    bar:
      background: '#3388ff'
      foreground: '#080808'
  hints:
    start:
      foreground: '#080808'
      background: '#33ff88'
    end:
      foreground: '#0000ff'
      background: '#3388ff'
  line_indicator:
    foreground: '#080808'
    background: '#00ff00'
  selection:
    text: CellBackground
    background: CellForeground

  normal:
    black:   '#000000'
[...]

PR #6207, opened 2022-07-14
issue #6178 - fixed hint highlighting

aaron-loa:
#6178 
I found the reason why only the first character of the hint match got highlighted and tried to fix it.

line 448
` self.labels[self.matches.index].get(label_position).copied().map(|c| (c, is_first)`

If label_position was not 0 then None got returned, thats what caused the issue. I changed it to 0, everything seems to be working now, nothing should be broken.
Also added underlining for the rest of the hint.
alacritty_screenshot


aaron-loa:
Hmm, by multiple character labels, do you mean when there are too many visible matches on the same screen, so you need more than one character to select them?

aaron-loa:
Oh ok, with this change it works like this

First it only shows 1 label:
alacritty_1

Then when you press that button, then another label shows up in the place of the previous one:
alacritty_2

I thought this was intended, but just checked with the release version, I see how it should work now.

aaron-loa:
Now it is supposed to work as intended.
alacritty_3

Instead of iterating through only the labels of the match, now it iterates through the whole regex match, underlining after the labels

alacritty/src/display/content.rs
```diff
@@ -222,18 +222,21 @@ impl RenderableCell {
...
         let colors = &content.config.colors;
         let mut character = cell.c;
-
-        if let Some((c, is_first)) =
-            content.hint.as_mut().and_then(|hint| hint.advance(viewport_start, cell.point))
-        {
-            let (config_fg, config_bg) = if is_first {
-                (colors.hints.start.foreground, colors.hints.start.background)
+        let mut flags = cell.flags;
+        if content.hint.as_mut().map(|hint| hint.matches.advance(cell.point)).unwrap_or(false) {
```
chrisduerr, line 226:
```suggestion
        if content.hint.as_mut().map_or(false, |hint| hint.matches.advance(cell.point)) {
```

alacritty/src/display/content.rs
```diff
@@ -222,18 +222,21 @@ impl RenderableCell {
...
+            if let Some((c, is_first)) =
+                content.hint.as_mut().and_then(|hint| hint.advance(viewport_start, cell.point))
+            {
+                let (config_fg, config_bg) = if is_first {
+                    (colors.hints.start.foreground, colors.hints.start.background)
+                } else {
+                    (colors.hints.end.foreground, colors.hints.end.background)
+                };
+                character = c;
+                Self::compute_cell_rgb(&mut fg, &mut bg, &mut bg_alpha, config_fg, config_bg);
```
chrisduerr, line 236:
```suggestion
                Self::compute_cell_rgb(&mut fg, &mut bg, &mut bg_alpha, config_fg, config_bg);
                
                character = c;
```

Doesn't make sense to interrupt the color determination/computation with the character assignment when the character doesn't depend on the color determination.

alacritty/src/display/content.rs
```diff
@@ -428,11 +429,6 @@ impl<'a> Hint<'a> {
     ///
     /// The tuple's [`bool`] will be `true` when the character is the first for this hint.
     fn advance(&mut self, viewport_start: Point, point: Point) -> Option<(char, bool)> {
-        // Check if we're within a match at all.
-        if !self.matches.advance(point) {
-            return None;
-        }
```
chrisduerr, line 434: Removing this seems very wrong.

chrisduerr (changes requested):

alacritty/src/display/content.rs
aaron-loa: I removed it because if that function is called then it's already evaluated that we are already inside a match, first line here:

`if content.hint.as_mut().map_or(false, |hint| hint.matches.advance(cell.point)) {`
`if let Some((c, is_first)) = content.hint.as_mut().and_then(|hint| hint.advance(viewport_start, cell.point)){`

I can put it back, just to be extra safe, or if something else calls that function.

chrisduerr:
Sorry for the delay, but I'm not super pleased with how messy the current solution is with how simple the patch really should be, so I'd like to look at it myself over the weekend.

Just to let you know why I haven't put down another review.

alacritty/src/display/content.rs
```diff
@@ -420,14 +422,19 @@ struct Hint<'a> {
     labels: &'a Vec<Vec<char>>,
 }
 
+enum HintType {
+    Label(char, bool),
+    AfterLabel,
+}
```
chrisduerr, line 428: First of all I think it would probably make sense to break this out into three different variants, that way we don't have to match on the boolean after matching on the enum.

I also think that `HintPosition` might be a better name.

alacritty/src/display/content.rs
```diff
@@ -222,18 +222,22 @@ impl RenderableCell {
...
+                    let (config_fg, config_bg) = if is_first {
+                        (colors.hints.start.foreground, colors.hints.start.background)
+                    } else {
+                        (colors.hints.end.foreground, colors.hints.end.background)
+                    };
+                    Self::compute_cell_rgb(&mut fg, &mut bg, &mut bg_alpha, config_fg, config_bg);
+                    character = c;
+                },
+                HintType::AfterLabel => flags |= Flags::UNDERLINE,
+            }
```
chrisduerr, line 240: I think on this own the approach taken here isn't bad, the problem really is just that this method is an incredibly ugly behemoth that is slowly becoming unmaintainable.

I don't think there's anything simple you can do about this though, so I'd probably just accept the additional complexity for this PR.

chrisduerr (changes requested):
Since this is a user-facing change, we need a changelog entry.

alacritty/src/display/content.rs
aaron-loa: Sorry of the late reply, I agree, probably can't be solved as easily as I thought.
I don't really like this solution as well, but can't come up with a cleaner way, so I think I will just close the pr.

PR #6208, opened 2022-07-15
Audit and simplifying of dirty fields

greg904:
This PR simplifies the code related to knowing when the terminal should be redrawn:
* It tries to separate the `dirty` (in `WindowContext`) and the `hint_highlighted_dirty` mechanisms by being more explicit about which is needed.
* It stops setting the `dirty` field when it's not needed, which removes extra redraws that would take place when scrolling, typing and could impact latency (see previous PR #6206) without introducing too much complexity (only simple conditions).
* It tries to remove the usage of the `dirty` field, so that the terminal can act as the source of truth and the code can be simplified. Note that this is not done everywhere. For example, the search functionality is still relying on the `dirty` field.

It should probably be read commit by commit. The commits are very small and I tried to explain what I did in the messages.

### DIFF
--- a/alacritty/src/window_context.rs
+++ b/alacritty/src/window_context.rs
@@ -311,7 +311,7 @@ impl WindowContext {
             );
         }
 
-        if self.dirty || self.mouse.hint_highlight_dirty {
+        if self.mouse.hint_highlight_dirty {
             self.dirty |= self.display.update_highlighted_hints(
                 &terminal,
                 config,
--- a/alacritty/src/window_context.rs
+++ b/alacritty/src/window_context.rs
@@ -327,7 +327,7 @@ impl WindowContext {
             return;
         }
 
-        if self.dirty {
+        if self.dirty || terminal.has_damage() {
             // Force the display to process any pending display update.
             self.display.process_renderer_update();
 
--- a/alacritty/src/window_context.rs
+++ b/alacritty/src/window_context.rs
@@ -312,8 +312,8 @@ impl WindowContext {
         }
 
         if self.mouse.hint_highlight_dirty {
-            self.dirty |= self.display.update_highlighted_hints(
-                &terminal,
+            self.display.update_highlighted_hints(
+                &mut terminal,
                 config,
                 &self.mouse,
                 self.modifiers,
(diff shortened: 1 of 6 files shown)


alacritty/src/window_context.rs
```diff
@@ -327,7 +327,7 @@ impl WindowContext {
             return;
         }
 
-        if self.dirty {
+        if self.dirty || terminal.has_damage() {
```
kchibisov, line 330: The right condition would never be true, when the left is false.

In general all of your damage related changes must be discarded, since they are slowing down perf when actually rendering a frame. I'd suggest to not touch damage the way it's right now, hence it's optimal in terms of perf the way it's right now.


In general I'd suggest to focus on dirty bit.

kchibisov (changes requested):

greg904:
> The right condition would never be true, when the left is false.

In 5add7c03bd3e285425236f47d39efd879809acaa, we stop setting the field to true and rely on the terminal to check if the selection changed instead so that we can't actually miss one.

> In general all of your damage related changes must be discarded, since they are slowing down perf when actually rendering a frame.

Is there a way to test this? With vtebench at least, I don't see slowing down:
output

kchibisov:
I mean that Pty will send a `Wakeup`, which will set `dirty` bit. Those events are triggered if and only there's a terminal update, which means that it's dirty, so the dirty check on the terminal doesn't make any sense in a context of deciding to render or not, since Pty tells us that anyway.

The damaging is more rendering, etc related. In general you're enlarging the time we're under the lock if we check for damage the way you do, since later we collect damage again, leading to iterating damage rects more than needed.

greg904:
> I mean that Pty will send a Wakeup, which will set dirty bit. Those events are triggered if and only there's a terminal update, which means that it's dirty, so the dirty check on the terminal doesn't make any sense in a context of deciding to render or not, since Pty tells us that anyway.

Wakeup is sent when the pty reads something new, but not when ActionContext changes the selection or the cursor position because of a user event. The idea was that because the terminal already tracks the cursor position and selection changes, it might as well help us here, instead of having the additional `dirty` that needs to be kept in sync.

> The damaging is more rendering, etc related. In general you're enlarging the time we're under the terminal lock(which we must minimize) if we check for damage the way you do, since later we collect damage again, leading to iterating damage rects more than needed.
> 
> Note, you're not seeing any performance difference because your right condition is never executed, like at all on vte bench, since we're always dirty. I'm just saying that all damage changes are redundant and should be left as is.
> 
> The point of damaging wrt selection is to not convert it multiple times to a range, since it's not free iirc.

What about modifying Term to add a setter for selection, which would set the damage, and using that setter from ActionContext? It would fix this and it would even make the render path not call to_range at all if the selection hasn't changed unlike in master.

kchibisov:
Closing in favor of #6219

PR #6211, opened 2022-07-17
Replace `.map().unwrap_or()` with `.map_or()`.

chrisnc:
Use a `match` with a guard instead of a `.map().filter().unwrap_or()`
chain on an `Option`.

### DIFF
--- a/alacritty_terminal/src/ansi.rs
+++ b/alacritty_terminal/src/ansi.rs
@@ -1155,8 +1155,9 @@ where
         let mut params_iter = params.iter();
         let handler = &mut self.handler;
 
-        let mut next_param_or = |default: u16| {
-            params_iter.next().map(|param| param[0]).filter(|&param| param != 0).unwrap_or(default)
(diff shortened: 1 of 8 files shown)


alacritty_terminal/src/ansi.rs
```diff
@@ -1155,8 +1155,9 @@ where
         let mut params_iter = params.iter();
         let handler = &mut self.handler;
 
-        let mut next_param_or = |default: u16| {
-            params_iter.next().map(|param| param[0]).filter(|&param| param != 0).unwrap_or(default)
+        let mut next_param_or = |default: u16| match params_iter.next() {
+            Some(&[param, ..]) if param != 0 => param,
```
chrisnc, line 1159: This will change the behavior when the `param` slice has length `0`. Before this would `panic`, now it will return `default`. From my reading of the code in `vte` code this should never happen, but can leave this change out if it's not preferred.

kchibisov (approved):

PR #6216, opened 2022-07-19
Fix inverted condition for cell's extra clearing

### LINKED ISSUE #6215: hyperlink text is cut short by ECMA-48 SGR reset sequence
### To reproduce

```shell
▸ printf '\e]8;;https://github.com/alacritty\e\\\e[36mAla\e[2mcri\e[0mtty\e]8;;\e\\\n'
Alacritty
└────┴──── the resulting link spans this part only
```

The same test work as expected in kitty and gnome-terminal.

### System

OS: Linux
Version: alacritty 0.11.0-dev (2a676dfa)
Linux/BSD: X11

kchibisov:
Fixes #6215.


alacritty_terminal/src/term/cell.rs
```diff
@@ -186,7 +186,7 @@ impl Cell {
     pub fn set_hyperlink(&mut self, hyperlink: Option<Hyperlink>) {
         let should_drop = hyperlink.is_none()
             && self.extra.as_ref().map_or(true, |extra| {
-                !extra.zerowidth.is_empty() || extra.underline_color.is_some()
+                !extra.zerowidth.is_empty() || extra.underline_color.is_none()
             });
```
chrisduerr, line 190: Is the zerowidth logic correct here? Shouldn't it be `zerowidth.is_empty() && underline_color.is_none()`?

We only want to drop if zerowidth and color are both not set, right?

chrisduerr (changes requested):

chrisduerr (approved):

PR #6217, opened 2022-07-19
Avoid panic / exit when there are IO errors on tty initialization

mikayla-maki:
We want to use alacritty_terminal as a library in our app. We had an issue with the entire application crashing on terminal launch when EMFILE errors caused the TTY initialization to fail. This patch updates the `unix.rs` file to pass results up if there are errors opening the pty handles or reading the pw entry.

### DIFF
--- a/alacritty_terminal/src/tty/unix.rs
+++ b/alacritty_terminal/src/tty/unix.rs
@@ -33,14 +33,14 @@ macro_rules! die {
 }
 
 /// Get raw fds for master/slave ends of a new PTY.
-fn make_pty(size: winsize) -> (RawFd, RawFd) {
+fn make_pty(size: winsize) -> Result<(RawFd, RawFd)> {
     let mut window_size = size;
     window_size.ws_xpixel = 0;
     window_size.ws_ypixel = 0;
 
-    let ends = openpty(Some(&window_size), None).expect("openpty failed");
+    let ends = openpty(Some(&window_size), None)?;
 
-    (ends.master, ends.slave)
+    Ok((ends.master, ends.slave))
 }
 
 /// Really only needed on BSD, but should be fine elsewhere.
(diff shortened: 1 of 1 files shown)


chrisduerr (approved):
Seems very reasonable to me. Are the other two instances where we call `die!` here fine for your usecase?

Generally I'd say just exploding is usually a bad thing to do for a library, so we might want to consider changing these too?

Also I'd be curious what you use `alacritty_terminal` for if you're willing to share.

mikayla-maki:
> Are the other two instances where we call die! here fine for your usecase?

Honestly, I'm unsure. Those other two instances of `die!()` are for `ioctl(2)`. Checking the man page (https://man7.org/linux/man-pages/man2/ioctl.2.html), the only kinds of errors it produces seem to be programmer errors like 'fd is not a file descriptor' or 'bad argument format'. The only way the user's enviroment could produce these errors is if something freaky happened like the pty file descriptors being closed externally. Further, reporting these errors back up to the application seems difficult. One of them (the one in `impl OnResize for Pty > on_resize()`) is behind a channel send with no return arguments. Given the difficulty of handling the errors and my guess that `ioctl` failures are unlikely, it seems economical to leave them in and wait for a crash report to provide further context.

> Also I'd be curious what you use alacritty_terminal for if you're willing to share.

I'm using it as the back end for the https://zed.dev/ terminal component! One of the founders just gave your github account access to the alpha, feel free to try it out if you're on a mac :D

chrisduerr:
> Given the difficulty of handling the errors and my guess that ioctl failures are unlikely, it seems economical to leave them in and wait for a crash report to provide further context.

Yeah that checks out, can always improve on things later.

I believe right now there's not a lot of people actively using `alacritty_terminal` as a library, since it was originally just part of alacritty until it was extracted. So if you have any feedback or questions I'd be very interested in further building out this usecase.

mikayla-maki:
I'll let you know as I run into them! It's been surprisingly smooth sailing so far. The `alacritty_terminal` crate does almost everything. The only part I've had to really pay attention to is translating inputs (mouse, keys, etc.) to the correct escape sequences.

PR #6219, opened 2022-07-20
Remove redundant dirty updates

kchibisov:
In some cases dirty was set without any ui update leading
to extra redraws, this commit resolves this.

Co-authored-by: Greg Depoire--Ferrer <greg@gregdf.com>

cc @greg904

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -217,16 +217,17 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
 
         self.terminal.scroll_display(scroll);
 
+        let lines_changed = old_offset - self.terminal.grid().display_offset() as i32;
+
         // Keep track of manual display offset changes during search.
         if self.search_active() {
-            let display_offset = self.terminal.grid().display_offset();
-            self.search_state.display_offset_delta += old_offset - display_offset as i32;
+            self.search_state.display_offset_delta += lines_changed;
         }
 
+        let vi_mode = self.terminal.mode().contains(TermMode::VI);
+
         // Update selection.
-        if self.terminal.mode().contains(TermMode::VI)
-            && self.terminal.selection.as_ref().map_or(true, |s| !s.is_empty())
-        {
+        if vi_mode && self.terminal.selection.as_ref().map_or(false, |s| !s.is_empty()) {
             self.update_selection(self.terminal.vi_mode_cursor.point, Side::Right);
         } else if self.mouse.left_button_state == ElementState::Pressed
             || self.mouse.right_button_state == ElementState::Pressed
@@ -236,7 +237,8 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             self.update_selection(point, self.mouse.cell_side);
         }
 
-        *self.dirty = true;
+        // Update dirty if actually scrolled or we're in the Vi mode.
+        *self.dirty |= lines_changed != 0 || vi_mode;
     }
 
     // Copy text selection.
@@ -258,8 +260,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     }
 
     fn clear_selection(&mut self) {
-        self.terminal.selection = None;
-        *self.dirty = true;
+        *self.dirty |= self.terminal.selection.take().map_or(false, |s| !s.is_empty());
     }
 
     fn update_selection(&mut self, mut point: Point, side: Side) {
@@ -286,9 +287,8 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
 
     fn start_selection(&mut self, ty: SelectionType, point: Point, side: Side) {
         self.terminal.selection = Some(Selection::new(ty, point, side));
-        *self.dirty = true;
-
         self.copy_selection(ClipboardType::Selection);
+        *self.dirty = true;
     }
 
     fn toggle_selection(&mut self, ty: SelectionType, point: Point, side: Side) {
@@ -421,13 +421,11 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         *self.font_size = max(*self.font_size + delta, Size::new(FONT_SIZE_STEP));
         let font = self.config.font.clone().with_size(*self.font_size);
         self.display.pending_update.set_font(font);
-        *self.dirty = true;
     }
 
     fn reset_font_size(&mut self) {
         *self.font_size = self.config.font.size();
         self.display.pending_update.set_font(self.config.font.clone());
-        *self.dirty = true;
     }
 
     #[inline]
@@ -435,7 +433,6 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         if !self.message_buffer.is_empty() {
             self.display.pending_update.dirty = true;
             self.message_buffer.pop();
-            *self.dirty = true;
         }
     }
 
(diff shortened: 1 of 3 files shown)


alacritty/src/event.rs
```diff
@@ -286,9 +287,8 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
 
     fn start_selection(&mut self, ty: SelectionType, point: Point, side: Side) {
         self.terminal.selection = Some(Selection::new(ty, point, side));
-        *self.dirty = true;
-
         self.copy_selection(ClipboardType::Selection);
+        *self.dirty = true;
```
chrisduerr, line 291: Why was this moved? I actually kinda like putting it under the statement that actually requires the dirty flag.

The `copy_selection` doesn't care about dirty or not, right? So why not have the statement closer to what actually caused the dirty change?

alacritty/src/event.rs
```diff
@@ -651,13 +647,13 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         let timer_id = TimerId::new(Topic::BlinkCursor, self.display.window.id());
         if self.scheduler.unschedule(timer_id).is_some() {
             self.schedule_blinking();
-            self.display.cursor_hidden = false;
+            if mem::replace(&mut self.display.cursor_hidden, false) {
+                *self.dirty = true;
+            }
```
chrisduerr, line 652: I'd add an empty line above and put a quick comment that we only need to redraw if the cursor is currently hidden.

Also since this is a `bool` you can use `mem::take` instead.

alacritty/src/event.rs
```diff
@@ -716,6 +712,8 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
                 self.terminal.vi_goto_point(*hint_bounds.start());
             },
         }
+
+        self.mark_dirty();
```
chrisduerr, line 716: Which action here requires being marked dirty? We shouldn't just randomly sprinkle this around because it seems convenient. Most of the actions taken here don't seem to require any damage?

The only things I could imagine would be `MoveViModeCursor` and `Select`?

alacritty/src/event.rs
```diff
@@ -930,7 +928,6 @@ impl<'a, N: Notify + 'a, T: EventListener> ActionContext<'a, N, T> {
     fn exit_search(&mut self) {
         self.display.pending_update.dirty = true;
         self.search_state.history_index = None;
-        *self.dirty = true;
```
chrisduerr, line 933: This seems **really** wrong. There definitely should be some dirtiness going on when exiting the search. At the very least when `focused_match != None`.

alacritty/src/event.rs
```diff
@@ -1084,14 +1080,13 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                     // Disable blinking after timeout reached.
                     let timer_id = TimerId::new(Topic::BlinkCursor, self.ctx.display.window.id());
                     self.ctx.scheduler.unschedule(timer_id);
-                    self.ctx.display.cursor_hidden = false;
                     *self.ctx.cursor_blink_timed_out = true;
+                    self.ctx.display.cursor_hidden = false;
```
chrisduerr, line 1084: Nice, looks much more pretty this way.

alacritty/src/event.rs
```diff
@@ -1184,7 +1177,6 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                     },
                     WindowEvent::Focused(is_focused) => {
                         self.ctx.terminal.is_focused = is_focused;
-                        *self.ctx.dirty = true;
```
chrisduerr, line 1187: Removing this without cursor blinking might cause issues? We **must** have a dirty flag set whenever focus changed because of the box cursor change.

chrisduerr (changes requested):

alacritty/src/event.rs
```diff
@@ -236,7 +237,8 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             self.update_selection(point, self.mouse.cell_side);
         }
 
-        *self.dirty = true;
+        // Update dirty if actually scrolled or we're in the Vi mode.
+        *self.dirty |= lines_changed != 0 || vi_mode;
```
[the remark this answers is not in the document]
kchibisov: Since scroll can move Vi cursor.

alacritty/src/event.rs
kchibisov: I've sorted them by length.

alacritty/src/event.rs
kchibisov: Yes, due to Vi movements it's here.

alacritty/src/event.rs
kchibisov: There's dirty already `self.display.pending_update.dirty = true;`. No need to set it twice.

alacritty/src/event.rs
kchibisov: Oh, forgot about it, will update.

alacritty/src/event.rs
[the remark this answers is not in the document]
chrisduerr: Even if the display doesn't move at all?

alacritty/src/event.rs
chrisduerr: Yeah I'm not sure that's the best choice here, especially since there's a newline in between?

alacritty/src/event.rs
chrisduerr: Should only be in the vi branches then probably?

alacritty/src/event.rs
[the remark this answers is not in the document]
kchibisov: Yes, we clamp it to viewport boundaries from what I can see.

```rust
// Clamp vi mode cursor to the viewport.
let viewport_start = -(self.grid.display_offset() as i32);
let viewport_end = viewport_start + self.bottommost_line().0;
let vi_cursor_line = &mut self.vi_mode_cursor.point.line.0;
*vi_cursor_line = cmp::min(viewport_end, cmp::max(viewport_start, *vi_cursor_line));
```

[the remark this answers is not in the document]
chrisduerr: Yeah but that shouldn't do anything, right? Like the vi cursor should always be clamped, scrolling isn't meant just for clamping?

I thought the clamping here was only necessary if we actually do scroll?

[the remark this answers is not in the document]
kchibisov: Hm, seems like you are right here. We don't need an extra check for vi mod scroll.

alacritty/src/event.rs
```diff
@@ -714,6 +716,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
                 }
 
                 self.terminal.vi_goto_point(*hint_bounds.start());
+                self.mark_dirty();
```
chrisduerr, line 719: Do we also need to `mark_dirty` in the `HintInternalAction::Select`, or is that handled by `update_selection`? I'd assume the latter?

alacritty/src/event.rs
```diff
@@ -1184,7 +1182,11 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                     },
                     WindowEvent::Focused(is_focused) => {
                         self.ctx.terminal.is_focused = is_focused;
-                        *self.ctx.dirty = true;
+
+                        // When the unfocused hollow is used we must redraw on focus change.
+                        if self.ctx.config.terminal_config.cursor.unfocused_hollow {
+                            *self.ctx.dirty = false;
```
chrisduerr, line 1188:
```suggestion
                            *self.ctx.dirty = true;
```

chrisduerr (changes requested):

alacritty/src/event.rs
kchibisov: Yes, update selection does that.

chrisduerr (approved):

PR #6221, opened 2022-07-21
Register shells launched by Alacritty as tty sessions on macOS

### LINKED ISSUE #3420: Alacritty sessions aren't registered as TTYS on OSX Catalina
When executing the `who` command I normally see the console session (osx debug) and all iterm/terminal sessions on my Macbook. This is however not the case with Alacritty, none of my alacritty sessions (nor tmux sessions that I create within an Alacritty session) are shown in `who`. Is there a reason for alacritty somehow flying under the system's radar?

Who command executed in iterm 
Screenshot 2020-03-04 at 10 14 43

Who command in alacritty with Iterm no longer running:
Screenshot 2020-03-04 at 10 16 32

### System

OS: macOS
Version: alacritty 0.4.1 (4d9253b)

### Logs

Font/Terminal size:

```
[2020-03-04 10:22] [INFO] Welcome to Alacritty
[2020-03-04 10:22] [INFO] Configuration loaded from "/Users/peter/.config/alacritty/alacritty.yml"
[2020-03-04 10:22] [DEBUG] Estimated DPR: 1
[2020-03-04 10:22] [DEBUG] Estimated Cell Size: 6 x 15
[2020-03-04 10:22] [DEBUG] Estimated Dimensions: Some((1804.0, 1204.0))
[2020-03-04 10:22] [INFO] Device pixel ratio: 1
[2020-03-04 10:22] [INFO] Initializing glyph cache...
[2020-03-04 10:22] [INFO] ... finished initializing glyph cache in 0.030991822s
[2020-03-04 10:22] [INFO] Estimated DPR correctly, skipping resize
[2020-03-04 10:22] [INFO] Cell Size: 6 x 15
[2020-03-04 10:22] [INFO] Padding: 2 x 2
[2020-03-04 10:22] [INFO] Width: 1804, Height: 1204
[2020-03-04 10:22] [INFO] PTY Dimensions: Line(80) x Column(300)

```
Keyboard and bindings: unrelated

chrisnc:
Unless the `shell` config is specified, launch the user's shell with:

```sh
login -flp $USER /bin/sh -c "exec -a -shell /path/to/shell"
```

On macOS, just running a shell prefixed by `-` is not sufficient to be
registered as a login session for things like `w` and `logname`.
However, using the `login` command changes the directory to `$HOME`
before running the program by default, which is not desired. The `-l`
flag disables this behavior, but also skips prepending `-` to the
executed program, so shells will not run as login shells. Instead we
just do this part ourselves with `exec -a`. The result is login shells
that run in the intended directory and are registered as tty sessions.

Update `alacritty.yml` to reflect that Alacritty will use `$SHELL` or
the user's login shell on Linux, BSD, and macOS.

Fixes #3420

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -434,8 +434,7 @@
 # shell.
 #
 # Default:
-#   - (macOS) /bin/bash --login
-#   - (Linux/BSD) user login shell
+#   - (Linux/BSD/macOS) `$SHELL` or the user's login shell, if `$SHELL` is unset
 #   - (Windows) powershell
 #shell:
 #  program: /bin/bash
--- a/alacritty_terminal/src/tty/unix.rs
+++ b/alacritty_terminal/src/tty/unix.rs
@@ -1,8 +1,5 @@
 //! TTY related functionality.
 
-use std::borrow::Cow;
-#[cfg(not(target_os = "macos"))]
-use std::env;
 use std::ffi::CStr;
 use std::fs::File;
 use std::io::{Error, ErrorKind, Result};
@@ -10,7 +7,7 @@ use std::mem::MaybeUninit;
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 use std::os::unix::process::CommandExt;
 use std::process::{Child, Command, Stdio};
-use std::ptr;
+use std::{env, ptr};
 
 use libc::{self, c_int, winsize, TIOCSCTTY};
 use log::error;
@@ -21,7 +18,7 @@ use nix::sys::termios::{self, InputFlags, SetArg};
 use signal_hook::consts as sigconsts;
 use signal_hook_mio::v0_6::Signals;
 
-use crate::config::{Program, PtyConfig};
+use crate::config::PtyConfig;
 use crate::event::{OnResize, WindowSize};
 use crate::tty::{ChildEvent, EventedPty, EventedReadWrite};
 
@@ -121,17 +118,33 @@ impl Pty {
     }
 }
 
-#[cfg(target_os = "macos")]
-fn default_shell(pw: &Passwd<'_>) -> Program {
-    let shell_name = pw.shell.rsplit('/').next().unwrap();
-    let argv = vec![String::from("-c"), format!("exec -a -{} {}", shell_name, pw.shell)];
-
-    Program::WithArgs { program: "/bin/bash".to_owned(), args: argv }
+/// Look for a shell in the `$SHELL` environment variable, then in `passwd`.
+fn default_shell(pw: &Passwd<'_>) -> String {
+    env::var("SHELL").unwrap_or_else(|_| pw.shell.to_owned())
 }
 
 #[cfg(not(target_os = "macos"))]
-fn default_shell(pw: &Passwd<'_>) -> Program {
-    Program::Just(env::var("SHELL").unwrap_or_else(|_| pw.shell.to_owned()))
+fn default_shell_command(pw: &Passwd<'_>) -> Command {
+    Command::new(default_shell(pw))
+}
+
+#[cfg(target_os = "macos")]
+fn default_shell_command(pw: &Passwd<'_>) -> Command {
+    let shell = default_shell(pw);
+    let shell_name = shell.rsplit('/').next().unwrap();
+
+    // On macOS, use the `login` command so the shell will appear as a tty session.
+    let mut login_command = Command::new("/usr/bin/login");
+
+    // Exec the shell with argv[0] prepended by '-' so it becomes a login shell.
+    // `login` normally does this itself, but `-l` disables this.
+    let exec = format!("exec -a -{} {}", shell_name, shell);
+
+    // -f: Bypasses authentication for the already-logged-in user.
+    // -l: Skips changing directory to $HOME and prepending '-' to argv[0].
+    // -p: Preserves the environment.
+    login_command.args(["-flp", pw.name, "/bin/sh", "-c", &exec]);
+    login_command
 }
 
 /// Create a new TTY and return a handle to interact with it.
@@ -148,15 +161,13 @@ pub fn new(config: &PtyConfig, window_size: WindowSize, window_id: Option<usize>
     let mut buf = [0; 1024];
     let pw = get_pw_entry(&mut buf)?;
 
-    let shell = match config.shell.as_ref() {
-        Some(shell) => Cow::Borrowed(shell),
-        None => Cow::Owned(default_shell(&pw)),
+    let mut builder = if let Some(shell) = config.shell.as_ref() {
+        let mut cmd = Command::new(shell.program());
+        cmd.args(shell.args());
+        cmd
+    } else {
+        default_shell_command(&pw)
     };
-
-    let mut builder = Command::new(shell.program());
-    for arg in shell.args() {
-        builder.arg(arg);
-    }
 
     // Setup child stdin/stdout/stderr as slave fd of PTY.
     // Ownership of fd is transferred to the Stdio structs and will be closed by them at the end of
@@ -170,10 +181,6 @@ pub fn new(config: &PtyConfig, window_size: WindowSize, window_id: Option<usize>
     builder.env("LOGNAME", pw.name);
     builder.env("USER", pw.name);
     builder.env("HOME", pw.dir);
-
-    // Set $SHELL environment variable on macOS, since login does not do it for us.
-    #[cfg(target_os = "macos")]
-    builder.env("SHELL", config.shell.as_ref().map_or(pw.shell, Program::program));
 
     if let Some(window_id) = window_id {
         builder.env("WINDOWID", format!("{}", window_id));
@@ -210,7 +217,7 @@ pub fn new(config: &PtyConfig, window_size: WindowSize, window_id: Option<usize>
     }
 
     // Prepare signal handling before spawning child.
-    let signals = Signals::new(&[sigconsts::SIGCHLD]).expect("error preparing signal handling");
+    let signals = Signals::new([sigconsts::SIGCHLD]).expect("error preparing signal handling");
 
     match builder.spawn() {
         Ok(child) => {
@@ -231,8 +238,12 @@ pub fn new(config: &PtyConfig, window_size: WindowSize, window_id: Option<usize>
             Ok(pty)
         },
         Err(err) => Err(Error::new(
-            ErrorKind::NotFound,
-            format!("Failed to spawn command '{}': {}", shell.program(), err),
+            err.kind(),
+            format!(
+                "Failed to spawn command '{}': {}",
+                builder.get_program().to_string_lossy(),
+                err
+            ),
         )),
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Windows `Open Alacritty Here` on root of drive displaying error
 - On macOS, `font.use_thin_strokes` did not work since Big Sur
 - On macOS, trying to load a disabled font would crash
+- On macOS, Alacritty sessions did not appear in the list of tty sessions for `w` and `who`
 
 ### Removed
 


alacritty_terminal/src/tty/unix.rs
```diff
@@ -232,7 +242,11 @@ pub fn new(config: &PtyConfig, window_size: WindowSize, window_id: Option<usize>
         },
         Err(err) => Err(Error::new(
             ErrorKind::NotFound,
-            format!("Failed to spawn command '{}': {}", shell.program(), err),
+            format!(
+                "Failed to spawn command '{}': {}",
```
chrisnc, line 246: Wondering if we should just log the full command here.

chrisnc:
TODO: update the comments in `alacritty.yml` to reflect the default shell behaviors.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -121,17 +118,38 @@ impl Pty {
...
+    // launch the shell so programs like `w`, `who`, and `logname` work.
+    #[cfg(target_os = "macos")]
+    let mut command = if shell_config.is_some() {
+        Command::new(shell)
+    } else {
+        let mut login_command = Command::new("/usr/bin/login");
+        // -f bypasses authentication for the already-logged-in user
+        // -l makes it not a "login" shell and prevents login from changing directories to $HOME
+        // exec the shell with argv[0] prepended by - so the shell run as a login shell anyway
+        login_command.args(["-fl", pw.name, "/bin/zsh", "-c"]);
```
chrisduerr, line 141: I'd say especially for older machines `zsh` is far more likely to not exist than `bash`. Sure bash might be deprecated as a login shell, but I can't imagine they'd get rid of it, breaking any script ever.

I'm not sure if `sh` is just some symlink to one of those, but it certainly would be another option.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -121,17 +118,38 @@ impl Pty {
...
+        login_command.arg(format!("exec -a -{} {}", shell.rsplit('/').next().unwrap(), shell));
```
chrisduerr, line 142: Why is this split into a separate line, that just seems incredibly error-prone.

Also I'd like to keep the shell_name variable extracted, because that way it's going to be much more obvious what the whole split next unwrap crap is actually doing.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -121,17 +118,38 @@ impl Pty {
...
+        .unwrap_or_else(|| pw.shell.to_owned());
 
-    Program::WithArgs { program: "/bin/bash".to_owned(), args: argv }
-}
+    #[cfg(not(target_os = "macos"))]
+    let mut command = Command::new(shell);
+
+    // On macOS, unless the user set the `shell` config in alacritty.yml, use the `login` command to
+    // launch the shell so programs like `w`, `who`, and `logname` work.
+    #[cfg(target_os = "macos")]
```
chrisduerr, line 133: This whole thing should not clutter up the codebase for people just trying to figure out how this works on sensible linux/bsd systems. There's a good reason why the macos stuff was extracted into a separate function.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -121,17 +118,38 @@ impl Pty {
...
+    };
 
-#[cfg(not(target_os = "macos"))]
-fn default_shell(pw: &Passwd<'_>) -> Program {
-    Program::Just(env::var("SHELL").unwrap_or_else(|_| pw.shell.to_owned()))
+    // If the user specified a shell in alacritty.yml, the arguments are added here.
+    // Note that if we wanted to implicitly invoke these via login on macOS also, the arguments
+    // would need to be added to the `format!("exec ...")` above, but we only take that path if
+    // there is no shell config, and hence no arguments to add.
+    shell_config.map(|prog| command.args(prog.args()));
```
chrisduerr, line 150: What's the point of doing this when we know there's no shell config for one of the macos branches anyway?

Also `-e` can override the shell too, not just the config file. It also seems completely irrelevant what we *could* do here, if anything we should document reasons why we don't.

alacritty_terminal/src/tty/unix.rs
chrisduerr: We shouldn't.

chrisduerr (changes requested):

alacritty_terminal/src/tty/unix.rs
chrisnc: `/bin/sh` appears to be a minimal GNU `bash`, but not a symlink to `/bin/bash`. I will switch to that rather than pick `zsh`.

I was curious about this though, and it looks like `zsh` has existed in macOS since 10.0:
https://opensource.apple.com/releases/
https://github.com/apple-oss-distributions/distribution-macOS/tree/mac-os-x-100

alacritty_terminal/src/tty/unix.rs
chrisnc: It's split because `format` returns an owned string, while all the other arguments are borrowed strings, so I found it clearer and more concise to put them first rather than in one big array with `.to_owned` everywhere. I will refactor this a bit and keep the `shell_name` extracted.

alacritty_terminal/src/tty/unix.rs
chrisnc: Sure, I'll make separate versions of the function.

alacritty_terminal/src/tty/unix.rs
chrisnc: This statement covers adding shell arguments for all platforms, and the comment is just explaining why the arguments can be added directly to the command. This will be obviated by separate functions.

alacritty_terminal/src/tty/unix.rs
chrisduerr: Extract the format, put it into the array as slice?

chrisnc: Sure.

chrisnc:
Updated to address feedback.

One thing I'm a little curious about is this line, which is still there:

```
    // Set $SHELL environment variable on macOS, since login does not do it for us.
    #[cfg(target_os = "macos")]
    builder.env("SHELL", config.shell.as_ref().map_or(pw.shell, Program::program));
```

When I run `login -f $USER /usr/bin/env` I see `SHELL` set to my login shell. I will go back and look at the discussion about this and see what I'm missing and whether this may not be needed anymore.

chrisnc:
It looks like `iTerm2` does not set `SHELL` itself, it only overwrites it if it's already in the environment and iTerm's own settings specify a different shell: https://github.com/gnachman/iTerm2/blob/master/sources/shell_launcher.c#L29-L41

The older iTerm shell launcher code referenced before also errors out if `SHELL` is unset:
https://github.com/gnachman/iTerm2/blob/f6b54d66b5e371b9e69a0881e9608e13e710fda1/sources/shell_launcher.c#L33-L36, so I think it's safe to assume it will be set and remove the special behavior here. Now I'm wondering if `SHELL` should be overridden on all platforms _if_ Alacritty has a different shell configured...

chrisnc:
Updated to also use `login -p` like `iTerm` does, which allows a few macOS and Alacritty-related environment variables to propagate to the shell as they did before...

alacritty.yml
```diff
@@ -429,13 +429,13 @@
...
-# You can set `shell.program` to the path of your favorite shell, e.g.
+# You can set `shell.program` to the path of your favorite shell, e.g.,
 # `/bin/fish`. Entries in `shell.args` are passed unmodified as arguments to the
 # shell.
 #
 # Default:
-#   - (macOS) /bin/bash --login
-#   - (Linux/BSD) user login shell
+#   - (Linux/BSD/macOS) use the `$SHELL` environment variable, or the user's
+#     login shell if `$SHELL` is unset
```
chrisduerr, line 438: Shouldn't be necessary to document `$SHELL` here. The previous Linux/BSD stuff was just fine tbh.

alacritty.yml
```diff
@@ -429,13 +429,13 @@
 
 # Shell
 #
-# You can set `shell.program` to the path of your favorite shell, e.g.
+# You can set `shell.program` to the path of your favorite shell, e.g.,
```
chrisduerr, line 432: Why did you put commas after all the `e.g.`s in this file? That's not correct…

alacritty_terminal/src/tty/unix.rs
```diff
@@ -121,17 +118,30 @@ impl Pty {
...
 }
 
-#[cfg(target_os = "macos")]
-fn default_shell(pw: &Passwd<'_>) -> Program {
-    let shell_name = pw.shell.rsplit('/').next().unwrap();
-    let argv = vec![String::from("-c"), format!("exec -a -{} {}", shell_name, pw.shell)];
-
-    Program::WithArgs { program: "/bin/bash".to_owned(), args: argv }
+fn default_shell(pw: &Passwd<'_>) -> String {
+    // Look for a shell in SHELL, then in passwd.
```
chrisduerr, line 122: Might as well extract this as a method doc.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -121,17 +118,30 @@ impl Pty {
...
-    Program::Just(env::var("SHELL").unwrap_or_else(|_| pw.shell.to_owned()))
+fn default_shell_command(pw: &Passwd<'_>) -> Command {
+    Command::new(default_shell(pw))
+}
+
+#[cfg(target_os = "macos")]
+fn default_shell_command(pw: &Passwd<'_>) -> Command {
+    // On macOS, use the `login` command to launch the shell so programs like `w`, `who`, and
+    // `logname` work.
+    let shell = default_shell(pw);
```
chrisduerr, line 135: This comment above the `default_shell` is somewhat confusing.

I'd recommend putting a newline between shell_name and login_command, then putting the comment above the actual `login` command spawn.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -121,17 +118,30 @@ impl Pty {
...
+    let shell_name = shell.rsplit('/').next().unwrap();
+    let mut login_command = Command::new("/usr/bin/login");
+    // -f bypasses authentication for the already-logged-in user
+    // -l tells login that this is not a login session and not to change directories to $HOME
+    // -p preserves the environment
+    // exec the shell with argv[0] prepended by `-` so it becomes a login shell anyway
+    let exec = format!("exec -a -{} {}", shell_name, shell);
+    login_command.args(["-flp", pw.name, "/bin/sh", "-c", &exec]);
+    login_command
```
chrisduerr, line 144: This is somewhat of a mess, the comment just squished between two lines without any whitespace above it, documenting the two following lines in opposite order.

The exec comment should be above the exec line, the flag comments should be above the args line. If this code is so intricate that it requires documentation above every single line, then that's perfectly fine.

Also what does `-l` actually do? Can you maybe post the `login --help` output? Having `-l` be `no-login` seems odd.

And also please format the exec comment to be capitalized with a full stop at the end and maybe change the args comment to use `-f: Bypasses authentication for the logged-in user.` as a comment format.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -148,16 +158,14 @@ pub fn new(config: &PtyConfig, window_size: WindowSize, window_id: Option<usize>
     let mut buf = [0; 1024];
     let pw = get_pw_entry(&mut buf)?;
 
-    let shell = match config.shell.as_ref() {
-        Some(shell) => Cow::Borrowed(shell),
-        None => Cow::Owned(default_shell(&pw)),
+    let mut builder = if let Some(prog) = config.shell.as_ref() {
+        let mut cmd = Command::new(prog.program());
+        cmd.args(prog.args());
+        cmd
```
chrisduerr, line 164:
```suggestion
    let mut builder = if let Some(shell) = config.shell.as_ref() {
        let mut cmd = Command::new(shell.program());
        cmd.args(shell.args());
        cmd
```

alacritty_terminal/src/tty/unix.rs
```diff
@@ -167,13 +175,7 @@ pub fn new(config: &PtyConfig, window_size: WindowSize, window_id: Option<usize>
...
 
     // Setup shell environment.
-    builder.env("LOGNAME", pw.name);
-    builder.env("USER", pw.name);
-    builder.env("HOME", pw.dir);
-
-    // Set $SHELL environment variable on macOS, since login does not do it for us.
-    #[cfg(target_os = "macos")]
-    builder.env("SHELL", config.shell.as_ref().map_or(pw.shell, Program::program));
+    builder.envs([("LOGNAME", pw.name), ("USER", pw.name), ("HOME", pw.dir)]);
```
chrisduerr, line 178: Please revert this change.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -232,7 +234,11 @@ pub fn new(config: &PtyConfig, window_size: WindowSize, window_id: Option<usize>
         },
         Err(err) => Err(Error::new(
             ErrorKind::NotFound,
-            format!("Failed to spawn command '{}': {}", shell.program(), err),
+            format!(
+                "Failed to spawn command '{}': {}",
+                builder.get_program().to_string_lossy(),
```
chrisduerr, line 239: This will just print `login` on macOS, right?

chrisduerr (changes requested):

alacritty.yml
chrisnc: I looked into it more and I think it's an American English vs. British English thing... I'll revert if you want to follow British English style guides, but there aren't British spellings everywhere (like "favourite" on this line), so I assumed American English would be preferred. https://jakubmarian.com/comma-after-i-e-and-e-g/

alacritty_terminal/src/tty/unix.rs
chrisnc: Right, that's why I asked earlier if we should print the whole command, but you replied that we should not.

alacritty_terminal/src/tty/unix.rs
chrisnc: Yes, I agree it's a little odd. Here is `man login` for the flags I'm using:

```
    -f      When a user name is specified, this option indicates that proper authentication has already been done and that no password need be requested.  This option
             may only be used by the super-user or when an already logged in user is logging in as themselves.

             With the -f option, an alternate program (and any arguments) may be run instead of the user's default shell.  The program and arguments follows the user
             name.

     -l      Tells the program executed by login that this is not a login session (by convention, a login session is signalled to the program with a hyphen as the first
             character of argv[0]; this option disables that), and prevents it from chdir(2)ing to the user's home directory.  The default is to add the hyphen (this is
             a login session).

     -p      By default, login discards any previous environment.  The -p option disables this behavior.
 ```

Comment (chrisnc) [reply alacritty_terminal/src/tty/unix.rs:144]: And sure, I can re-order the comments. The order it is now was due to adding the `&exec` to the same `args` call rather than after it in its own `arg` call, per the other feedback.

File: alacritty_terminal/src/tty/unix.rs (hunk 12 above)
Comment (chrisnc) [reply alacritty_terminal/src/tty/unix.rs:178]: Just `env` * 3 vs. `envs` part? Will do.

What are your thoughts on setting `SHELL` on macOS? My thinking right now is we shouldn't do it anywhere. It's kind of fraught, and the behavior right now results in people who set `/usr/bin/login /bin/bash` or something as their shell via `alacritty.yml` have `$SHELL` set to `/usr/bin/login`, which isn't desirable.

File: alacritty.yml (hunk 6 above)
Comment (chrisnc) [reply alacritty.yml:438]: Hm... in the code `SHELL` takes precedence over the user's login shell though. Do you want to change that behavior?

File: alacritty_terminal/src/tty/unix.rs (hunk 13 above)
Comment (chrisnc) [reply alacritty_terminal/src/tty/unix.rs:244]: Pushed a version that just prints the whole command, and re-uses the triggering error's `kind`, rather than assuming `NotFound`.

File: alacritty_terminal/src/tty/unix.rs (hunk 12 above)
Comment (chrisnc) [reply alacritty_terminal/src/tty/unix.rs]: Reverted `envs`. Will leave this thread open to talk about setting `SHELL`.

File: alacritty_terminal/src/tty/unix.rs (hunk 10 above)
Comment (chrisnc) [reply alacritty_terminal/src/tty/unix.rs]: Updated.

File: alacritty.yml (hunk 6 above)
Comment (chrisduerr) [reply alacritty.yml:438]: No the current behavior is correct. I guess we could keep this but I'd try to at the very least keep it in a single line.

File: alacritty.yml (hunk 7 above)
Comment (chrisduerr) [reply alacritty.yml:432]: Doesn't matter if it's british or american, no reason to change something that is correct. You're just introducing unnecessary changes.

File: alacritty_terminal/src/tty/unix.rs (hunk 12 above)
Comment (chrisduerr) [reply alacritty_terminal/src/tty/unix.rs]: Honestly don't know about `$SHELL`, clearly at some point someone somewhere thought it was necessary, but I don't really care enough about macOS to hunt down exactly the reason why.

If it works without it on the latest fresh macOS version, I'm fine with removing it. If people complain we can always add it again.

File: alacritty_terminal/src/tty/unix.rs (hunk 13 above)
Comment (chrisduerr) [reply alacritty_terminal/src/tty/unix.rs:243]: No printing just `login` is fine, I was just trying to confirm its behavior.

Pushing the whole command will just be confusing to users, considering there's a bunch of exec and argument concatenation going on, it might even make some macOS users think there's something wrong with their Alacritty.

File: alacritty_terminal/src/tty/unix.rs (hunk 12 above)
Comment (chrisnc) [reply alacritty_terminal/src/tty/unix.rs]: Sounds good. Will leave it out.

File: alacritty.yml (hunk 6 above)
Comment (chrisnc) [reply alacritty.yml:438]: Sure, I'll shorten it a bit.

File: alacritty.yml (hunk 7 above)
Comment (chrisnc) [reply alacritty.yml:432]: Ok. To American readers it is wrong to omit the comma, which is why I changed it initially, but I'll take them out.

File: alacritty_terminal/src/tty/unix.rs (hunk 13 above)
Comment (chrisnc) [reply alacritty_terminal/src/tty/unix.rs:243]: Sure, will leave it.

**Comment by chrisnc** [comment]
Updated to address latest feedback.

File: alacritty_terminal/src/tty/unix.rs [hunk 14]
```diff
@@ -121,17 +118,33 @@ impl Pty {
...
+
+    // Exec the shell with argv[0] prepended by '-' so it becomes a login shell.
+    // `login` normally does this itself, but `-l` disables this.
+    let exec = format!("exec -a -{} {}", shell_name, shell);
+
+    // -f: Bypasses authentication for the already-logged-in user.
+    // -l: Skips changing directory to $HOME and prepending '-' to argv[0].
+    // -p: Preserves the environment.
+    login_command.args(["-flp", pw.name, "/bin/sh", "-c", &exec]);
+    login_command
```
chrisduerr, line 147: Reads like a novel by Charles Dickens, much nicer now. :+1:

chrisduerr (approved):
Thanks for your work, macOS contributions are always very welcome.

alacritty_terminal/src/tty/unix.rs
chrisnc: More like a greek tragedy in 3 acts...

PR #6223, opened 2022-07-24
Fix clippy lints

chrisnc:
Fix clippy lints

`first()` instead of `.get(0)`

No `let _: ()`

`derive Eq` where possible

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -134,7 +134,7 @@ impl From<glutin::ContextError> for Error {
 }
 
 /// Terminal size info.
-#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq)]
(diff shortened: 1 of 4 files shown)


chrisnc:
Ha... the clippy in CI prefers `unwrap_or_else` just because it's a function call. `cargo +nightly clippy` complains in the opposite direction. I tried it on compiler explorer and the generated code is the same.

chrisduerr:
> Ha... the clippy in CI prefers unwrap_or_else just because it's a function call. cargo +nightly clippy complains in the opposite direction. I tried it on compiler explorer and the generated code is the same.

Yeah, just run clippy on oldstable. This is intentional, it just takes some time for things to propagate.

kchibisov (approved):

chrisduerr (approved):

PR #6226, opened 2022-07-26
Fix visible regex match on tall viewports

pdelong:
The end of the search window is currently calculated using the viewport start instead of the end.  The observed behavior is that all hinting stops suddenly after line 101.  This was introduced in #6139 when the code was refactored into this file from `display/content.rs`.


kchibisov (approved):

kchibisov:
Thanks.

PR #6228, opened 2022-07-27
Allow external applications to manage the grid buffer

mikayla-maki:
I am building a renderer for the `alacritty_terminal` crate and I want to estimate the throughput of data in the terminal in order to scale my renderer's frame rate between 10fps and 120fps. To do so, I'd like to be able to periodically clear portions of the grid with `grid.update_history()`. However, this is method takes a mutable reference to the grid and there is no way to get the current grid mutably. This PR adds such a method to the Term API.


kchibisov:
Unfortunately I don't quite get the reason behind mutating it to measure throughput of data. How do you want to scale those?

mikayla-maki:
@kchibisov My renderer is embedded in a code editing application and both the application and the terminal event loop are running on a single process. When the terminal is running a script that hammers stdout (e.g. `yes`, or cat-ing a million line file), the entire application freezes up due to congestion around the terminal mutex. In order to reduce congestion, I want to throttle the rate at which I grab the terminal mutex based on the throughput of data through the terminal. To estimate throughput, I've decided to try allocating a grid buffer that is twice as large as I want to show the user. With this extra large buffer I can measure the difference between the 'real' and 'visible' sizes on every frame and use `grid.update_history(visible_size); grid.update_history(real_size);` to clear out the difference for measurement on the next frame. If y'all have a suggestion for a better way to measure throughput, I'd love to implement it.

kchibisov:
I think I understand the issue, and the real solution for you to schedule frames in your editor, so it won`t lock mutex for a while. In general terminal is running on a different thread (see `PTY reader`) so the only issue here is that you lock it longer than you should.

Anyway, your approach doesn't make any sense to me. The only reliable way to measure throughput is to broadcast the amount of data we read from the pty. See `event_loop.rs`. Will probably require adding a getter and resetter for it on terminal.

chrisduerr:
I generally don't have a problem with making this method public, it's certainly useful from an API perspective. But I can guarantee your approach won't solve your problem.

Just think about running `yes` in the alternate screen buffer. There's no history there and the update rate is still the exact same.

Generally high update frequency is not necessarily connected to **new** lines being output. Lots of terminal applications update frequently but do so in-place. Mutex congestion can be an issue with that, but generally it should only be problematic by limiting throughput when rendering too much, since the PTY is in a different thread. As @kchibisov suggested scheduling frames independently from output rate is probably a better strategy.

chrisduerr (changes requested):
Seems like you've blocked pushes to your PR branch, so I'm just going to suggest the following right here:

```diff
diff --git a/alacritty_terminal/src/term/mod.rs b/alacritty_terminal/src/term/mod.rs
index 5bd7f8c6..8317e701 100644
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -580,15 +580,11 @@ impl<T> Term<T> {
     }

     /// Access to the raw grid data structure.
-    ///
-    /// This is a bit of a hack; when the window is closed, the event processor
-    /// serializes the grid state to a file.
     pub fn grid(&self) -> &Grid<Cell> {
         &self.grid
     }

-    /// Mutable access for swapping out the grid during tests or for other applications
-    ///which want to help manage the back buffer.
+    /// Mutable access to the raw grid data structure.
     pub fn grid_mut(&mut self) -> &mut Grid<Cell> {
         &mut self.grid
     }
```

That should keep things simple and make the grid easy to play around with for library consumers. Doesn't matter if you still need it or someone else does in the future.

mikayla-maki:
Thank you both for the feedback, and for pre-catching some bugs for me :D I’ll make those changes and rework my approach. I really appreciate the broader perspective!

mikayla-maki:
@chrisduerr I've updated those doc comments, unsure why the build is failing thought. I'll try to see about enabling pushes on branches in the future!

chrisduerr:
> I've updated those doc comments, unsure why the build is failing thought. I'll try to see about enabling pushes on branches in the future!

There's something wrong with SourceHut's Arch CI. Unrelated to this PR.

chrisduerr (approved):

PR #6233, opened 2022-07-31
Implement WheelUp and WheelDown actions (simulate wheel rolls)

c4eater:
Alacritty is my main terminal emulator but I've always been finding its Scroll actions overly speedy - even the ScrollHalfPageUp / ScrollHalfPageDown ones. The mouse wheel shows just the right scrolling speed but there is no equivalent keyboard action. So this commit brings in WheelUp / WheelDown actions which are simulating the mouse wheel rolls. The scrolling speed for those can be controlled via the "multiplier" setting so "multiplier" now affects both the wheel and WheelUp / WheelDown. Action names are subject to discussion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Config option `cursor.blink_timeout` to timeout cursor blinking after inactivity
 - Escape sequence to set hyperlinks (`OSC 8 ; params ; URI ST`)
 - Config `hints.enabled.hyperlinks` for hyperlink escape sequence hint highlight
+- Added WheelUp and WheelDown actions to scroll by multiple lines as if the mouse wheel was rolled
 
 ### Changed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -104,7 +104,8 @@
   # Specifying '0' will disable scrolling.
   #history: 10000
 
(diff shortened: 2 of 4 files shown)


chrisduerr:
There's already a `ScrollLineUp` and `ScrollLineDown` action. Yes that is not affected by the mouse multiplier, which is good because that would make zero sense.

c4eater:
> Yes that is not affected by the mouse multiplier

And that is the problem. Scrolling with ScrollLineUp / ScrollLineDown is slow and is not suitable for viewing a large file. WheelUp / WheelDown has the right speed, plus, unlike ScrollLineUp / ScrollLineDown, it is tunable.

Please reconsider reopening the PR.

chrisduerr:
Then bind it multiple times. Mousewheel scroll having the "right" speed is pure coincidence and it's absolute nonsense to add a special keyboard binding to it.

c4eater:
> Then bind it multiple times. 

I don't think this is a robust solution.

> and it's absolute nonsense to add a special keyboard binding to it.

So that's the naming problem / reference to mouse wheel that is the problem for you? OK, suggest I renamed the new actions to MultiLineUp / MultiLineDown. Does that address the issue you have with the semantics?

chrisduerr:
> I don't think this is a robust solution.

I don't really care what you think about it. This will never be merged.

c4eater:
Wow that's a really quick jump from a technical discussion to "get lost".
OK, point taken.

PR #6254, opened 2022-08-11
Include extra directory in crate

### LINKED ISSUE #6242: Publish `extra` dir on crates io as well
Debian requires all the assets bundled inside the crate from crates.io.

chrisduerr:
This moves all the files from the extra directory into the alacritty
directory, to automatically bundle it when building the package for
crates.io.

Closes #6242.


chrisduerr:
This seems like the easiest solution to the problem, but I'm open to other suggestions if you have some @kchibisov.

kchibisov:
I think specifying paths on what to bundle in Cargo.toml is better? Since we should also include `alacritty.yml`. If we just move it will create certain work for packages, since you'd need to adjust paths you take stuff to install.

In general I'd keep layout as is and maybe symlink stuff we need inside alacritty crate?

chrisduerr:
> In general I'd keep layout as is and maybe symlink stuff we need inside alacritty crate?

I've decided to not do this because we already have a symlink from `./extra/windows` to `./alacritty/windows`. I feel like it might get somewhat convoluted when we have `./alacritty/extra/windows -> ./extra/windows -> ./alacritty/windows`? Like this is getting pretty messy?

kchibisov:
I think what we can do is to move `alacritty/windows` into `extra` and then just symlink the entire extra? I don't see how it could be any messy. We'd also be able to remove `alacritty.png` symlink, since we'll have single extra symlink.

chrisduerr:
If we move `alacritty/windows` into `extra`, you won't be able to build on Windows anymore.

kchibisov:
Hm, maybe we can move the entire extra, add windows folder to it and put symlink from old location?

chrisduerr:
> Hm, maybe we can move the entire extra, add windows folder to it and put symlink from old location?

Isn't that exactly what I did in this PR?

kchibisov (changes requested):
This looks good. Could you add symlink to alacritty.yml into `alacritty` dir, like we have for `LICENSE` and such. Should be good to go afterwards.

kchibisov:
Same review. Should add a config symlink and it's good to go.

kchibisov (approved):

PR #6259, opened 2022-08-12
Mention the `--config-file` flag in README.md

mol-lux:
I was trying to find a documented way of specifying a custom config file. However, I was not able to find anything in the documentation.

Now, of course, when you run `alacritty --help`, the option is right there, but I only found it after some digging. However, I wish this had been mentioned in a more obvious location.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,8 @@ following locations:
 3. `$HOME/.config/alacritty/alacritty.yml`
 4. `$HOME/.alacritty.yml`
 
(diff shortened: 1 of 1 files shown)


chrisduerr:
There's a manpage plus a very detailed `--help` output. Additionally pretty much every terminal ever has a way to override the configuration file.

This is just redundant information bloating up our documentation. There's no reason to specify all our CLI flags in the readme.

PR #6264, opened 2022-08-14
[docs] use curl instead of local files if possible

nyurik:
Some commands could be done using the in-memory temporary files instead of requiring users to clone the repo.  This way alacritty can be installed with `cargo install`, and the terminfo can be updated without any `git clone.  Although it might be worth while to set up a separate installation section for these tricks?

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -292,10 +292,12 @@ infocmp alacritty
 ```
 
 If it is not present already, you can install it globally with the following
-command:
+command. You can also use a local `extra/alacritty.info` instead of the `<(curl ...)`:
 
 ```
-sudo tic -xe alacritty,alacritty-direct extra/alacritty.info
+sudo -i
+tic -xe alacritty,alacritty-direct <(curl https://raw.githubusercontent.com/alacritty/alacritty/master/extra/alacritty.info)
+exit
 ```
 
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
No, users should be cloning the repository.

nyurik:
@chrisduerr but you already offer a non-cloning variant with `cargo install alacritty` -- why not help those users and document how to set up other things?

chrisduerr:
I'd rather remove `cargo install` as an installation method. Which quite honestly nobody should be using anyway.

nyurik:
Why? This is a fairly standard way for many rust utilities like lsd, and very convenient.  Cloning on the other hand involves extra steps that later have to be cleaned up. In sure there are many ppl who began using various programs just because they were so easy to install. Why purposefully make it more difficult to gain new users??

kchibisov:
Alacritty is available in lots of distributions and using your distro package manager is always a preferred way. With cargo install you don't get extra files, terminfo(it's in ncurses now, but not on macOs for example), man pages, icons, desktop entries, and other documentation, like configuration file. So cloning is still required for all of that.

Using `cargo install` isn't that much useful here, unfortunately. Folks who want to install from crates io will always find it.

PR #6265, opened 2022-08-15
Add inline input method support

### LINKED ISSUE #1613: Support inline "input method" input

kchibisov:
This commit adds support for inline IME handling. It also makes the
search bar use underline cursor instead of using '_' character.

Fixes #1613.

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -20,8 +20,9 @@ use serde::{Deserialize, Serialize};
 use wayland_client::EventQueue;
 
 use crossfont::{self, Rasterize, Rasterizer};
+use unicode_width::UnicodeWidthChar;
 
-use alacritty_terminal::ansi::NamedColor;
+use alacritty_terminal::ansi::{CursorShape, NamedColor};
 use alacritty_terminal::config::MAX_SCROLLBACK_LINES;
 use alacritty_terminal::event::{EventListener, OnResize, WindowSize};
 use alacritty_terminal::grid::Dimensions as TermDimensions;
@@ -38,7 +39,7 @@ use crate::config::window::{Dimensions, Identity};
 use crate::config::UiConfig;
 use crate::display::bell::VisualBell;
 use crate::display::color::List;
-use crate::display::content::RenderableContent;
+use crate::display::content::{RenderableContent, RenderableCursor};
 use crate::display::cursor::IntoRects;
 use crate::display::damage::RenderDamageIterator;
 use crate::display::hint::{HintMatch, HintState};
@@ -46,7 +47,7 @@ use crate::display::meter::Meter;
 use crate::display::window::Window;
 use crate::event::{Mouse, SearchState};
 use crate::message_bar::{MessageBuffer, MessageType};
-use crate::renderer::rects::{RenderLines, RenderRect};
+use crate::renderer::rects::{RenderLine, RenderLines, RenderRect};
 use crate::renderer::{self, GlyphCache, Renderer};
 use crate::string::{ShortenDirection, StrShortener};
 
@@ -362,6 +363,9 @@ pub struct Display {
     /// The renderer update that takes place only once before the actual rendering.
     pub pending_renderer_update: Option<RendererUpdate>,
 
+    /// The ime on the given display.
+    pub ime: Ime,
+
     // Mouse point position when highlighting hints.
     hint_mouse_point: Option<Point>,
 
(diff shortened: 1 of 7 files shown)


alacritty/src/event.rs
```diff
@@ -803,6 +807,9 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             self.cancel_search();
         }
 
+        // We don't need IME in Vi mode.
+        self.window().set_ime_allowed(was_in_vi_mode);
```
chrisduerr, line 811: Is it correct to enable IME whenever leaving vi mode? Or should we just disable it if we're entering and it is enabled?

alacritty/src/event.rs
```diff
@@ -1216,12 +1227,39 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
...
                         }
                     },
-                    WindowEvent::Ime(ime) => {
-                        if let Ime::Commit(text) = ime {
+                    WindowEvent::Ime(ime) => match ime {
+                        Ime::Commit(text) => {
+                            // Clear preedit.
+                            self.ctx.display.ime.set_preedit(None);
+
+                            // Don't suppress the chars comming from the ime commit.
```
chrisduerr, line 1235:
```suggestion
                            // Don't suppress the chars coming from the IME commit.
```

chrisduerr (changes requested):
Preliminary review with a couple things I noticed.

alacritty/src/display/mod.rs
```diff
@@ -374,6 +378,77 @@ pub struct Display {
...
+        Default::default()
+    }
+
+    #[inline]
+    pub fn set_enabled(&mut self, is_enabled: bool) {
+        if is_enabled {
+            self.enabled = is_enabled
+        } else {
+            // Clear state when disabling IME.
+            *self = Default::default();
```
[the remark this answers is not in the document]
kchibisov: Well, we ensure that we don't have preedit. That's the point. We use preedit to determine whether we should show cursor or not. In general it can be like that, if the server sends you disable right away or you change the focus of the window.

alacritty/src/display/mod.rs
```diff
@@ -374,6 +378,77 @@ pub struct Display {
...
+    /// The preedit text.
+    text: String,
+
+    /// Byte offset for cursor start into the preedit text.
+    ///
+    /// `None` means that the cursor is invisible.
+    cursor_byte_offset: Option<usize>,
+
+    /// The cursor offset from the end of the preedit in char width.
+    cursor_end_offset: Option<usize>,
```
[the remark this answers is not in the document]
kchibisov: Right, I use end offset to determine how to shorten the text. If the cursor is more than `num_cols` from the end, I start shortening from the cursor position.

alacritty/src/event.rs
kchibisov: It's otherwise you won't have IME when leaving it?

alacritty/src/display/mod.rs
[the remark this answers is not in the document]
chrisduerr: Hm maybe a little easy to get confused here just by the variable names. But I don't have any concrete suggestions for improvement.

alacritty/src/event.rs
chrisduerr: Yeah, but are there scenarios where we leave vi mode but still don't want to allow IME? Because otherwise you're forcefully reenabling it here, right? I can't think of anything right now though, so I'd assume we **always** want IME enabled while VI mode isn't enabled?

kchibisov: There's no. We use IME by default when in out side of Vi mode.

If you look closely, we start with `IME` enabled, then if we enter `Vi` we disable it. In search mode we just enable it and disable when going back. So it's like consistent. Whenever we can type we enable IME, otherwise(Vi) disable.

alacritty/src/event.rs
```diff
@@ -1216,12 +1227,39 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
...
+                            self.ctx.display.ime.set_preedit(None);
+
+                            // Don't suppress the chars coming from the IME commit.
+                            let suppress = mem::take(self.ctx.suppress_chars);
                             for ch in text.chars() {
-                                self.received_char(ch)
+                                self.received_char(ch);
                             }
-                        }
+                            *self.ctx.suppress_chars = suppress;
```
chrisduerr, line 1240: Suppress chars would only ever be triggered by key bindings that were entered while the IME is active. The fact that you're looking out for this here means there's a chance keybindings might be read while IME is performed? Do we want this or do we want to disable bindings during IME?

It seems strange to temporarily disable suppression here.

chrisduerr (changes requested):

alacritty/src/display/mod.rs
```diff
@@ -853,6 +928,55 @@ impl Display {
...
+                let search_text = Self::format_search(regex, search_label, size_info.columns());
+
+                // Render the search bar.
+                self.draw_search(config, &search_text);
+
+                // Draw search bar cursor.
+                let line = size_info.screen_lines();
+                let column = Column(search_text.chars().count() - 1);
+
+                if self.ime.preedit().is_none() {
```
chrisduerr, line 948:
```suggestion
                // Add cursor to search bar if IME is not active.
                if self.ime.preedit().is_none() {
```

alacritty/src/display/mod.rs
```diff
@@ -853,6 +928,55 @@ impl Display {
...
+                    let fg = config.colors.footer_bar_foreground();
+                    let shape = CursorShape::Underline;
+                    let cursor = RenderableCursor::new(Point::new(line, column), shape, fg, false);
+                    rects.extend(
+                        cursor.rects(&size_info, config.terminal_config.cursor.thickness()),
+                    );
+                }
+
+                // Compute IME position.
```
chrisduerr, line 957:
```suggestion
```

Nothing being computed here.

alacritty/src/display/mod.rs
```diff
@@ -853,6 +928,55 @@ impl Display {
             rects.push(visual_bell_rect);
         }
 
+        // Handle search and IME positioning.
```
chrisduerr, line 931:
```suggestion
        // Handle IME positioning and search.
```

Would just like to make it clear we're not handling search positioning here. Could maybe also change it from just `search` to `search rendering` or `search drawing`.

alacritty/src/event.rs
```diff
@@ -476,6 +476,9 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             };
         }
 
+        // Enable IME so we can input into the search bar with it when we were in Vi mode.
```
chrisduerr, line 479:
```suggestion
        // Enable IME so we can input into the search bar with it if we were in Vi mode.
```

alacritty/src/event.rs
```diff
@@ -803,6 +807,9 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             self.cancel_search();
         }
 
+        // We don't need IME in Vi mode.
```
chrisduerr, line 810:
```suggestion
        // We don't want IME in Vi mode.
```

Not just that it's unnecessary, but that it would actively be harmful, right?

chrisduerr (changes requested):

chrisduerr (approved):

PR #6274, opened 2022-08-17
Added action to either copy selection to clipboard or paste from clipb.

simmihugs:
Added an action similar to the right click behavior in Windows terminal/cmd

With configuration setting

```yml
 mouse_bindings:
  - { mouse: Right,                 action: CopySelectionOrPasteClipboard }
```

Right-click on a selection, copies selection to clipboard (does not paste) and deselects
Right-click without selection, pastes from clipboard

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -125,6 +125,9 @@ pub enum Action {
     /// Paste contents of selection buffer.
     PasteSelection,
 
+    /// Paste clipboard if nothing is selected, or copy selection to clipboard and clear selection.
+    CopySelectionOrPasteClipboard,
+
     /// Increase font size.
     IncreaseFontSize,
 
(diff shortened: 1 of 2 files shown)


chrisduerr:
What good does this option do? You say you added it because it is a feature in Windows terminal/cmd, but it just seems like a terrible feature to me.

simmihugs:
Allows for 2mouse button workflow

Personally my mouse has a very shaky mouse wheel (when clicked, scrolling is fine) - so having an action like this, it allows to use 2 mouse buttons 
left=selection
right=paste or copy

rather then 3 mouse buttons
left=selection
right=copy
scroll wheel=paste

I feel this behavior is intuitive, if I have something selected, it gets copied, if I have not selected something, what would a copying do good? so paste from clipboard. Also, middle click remains a option if that's more your thing.

I did not add it because it is a feature in windows, I mentioned this, because I encountered this behavior when using windows. I realized it fits my workflow and it does usually not exist in linux terminals. So if someone is looking also for this behavior might find it easier. I encountered this question, how do I get this for multiple terminals, st, termite, gnome terminal - usually the response is, to remap `PasteSelection` to right click, but this is a different behavior.

chrisduerr:
Not a fan at all of conditional stuff based on the clipboard, your workflow just sounds like you're trying to work around a hardware bug which is not appropriate for upstream Alacritty.

PR #6276, opened 2022-08-20
Fix general and instance class set on window

ulrikdem:
This fixes a regression introduced in 7d708d5, which caused the general
and instance class to be swapped.


chrisduerr (approved):
Oh good catch, looks like previously `with_class` accepted it in the opposite order to what `with_name` does. So this should be the correct order now.

PR #6282, opened 2022-08-23
Rework `--class` CLI option

### LINKED ISSUE #6279: CLI argument `--class` not setting `app_id` in Wayland
### The Problem
Running `alacritty --class test` on Wayland results in an alacritty window with `app_id` set to `Alacritty` (the default value).
**Expected:** An aclacritty window with `app_id` set to `test`.

I tried setting `class:general:test` in the configuration file and the `app_id` of new alacritty windows are indeed set to `test`. So I assume that somehow the CLI value is ignored (probably because it sets the `instance` name and not `general`, see below).

### Context
This is probably related to https://github.com/alacritty/alacritty/commit/23915cfc5343859329731c50114f701997c29b1f, and in turn https://github.com/alacritty/alacritty/commit/7d708d53f7ecb4c64db10ae44e7f4fca47c29a85 that caused the initial regression.

I'm not very familiar with Rust or the Alacritty codebase. However I see that `winit` changed the API to use `with_name()` instead of `with_app_id()/with_class()` (https://github.com/rust-windowing/winit/pull/2250/commits/a78e801f13231f773e5d19818097f76f53a8b6d0) and this triggered the above changes. There I see that on Wayland the class `general` name is used as `app_id` and `instance` name is ignored.

The problem seems to be that `--class` sets the instance name instead (the previous invocation of `with_app_id()` used the `instance` name too) but the new `with_name()` ignores the instance name on Wayland.

### System

OS: Linux
Version: alacritty 0.11.0-dev (23915cfc)
Linux: Wayland, sway

kchibisov:
This commit swaps the order of `general` and `instance` arguments
and also sets `instance` to `general` when only one argument was
provided. This should make this option behave like in other terminals
on X11, since they set either both or general by default, but
not instance like alacritty.

Fixes #6279.

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -11,7 +11,7 @@ use serde_yaml::Value;
 
 use alacritty_terminal::config::{Program, PtyConfig};
 
-use crate::config::window::{Class, Identity, DEFAULT_NAME};
+use crate::config::window::{Class, Identity};
 use crate::config::{serde_utils, UiConfig};
 
 /// CLI options for the main Alacritty executable.
@@ -159,19 +159,16 @@ fn option_as_value(option: &str) -> Result<Value, serde_yaml::Error> {
 
 /// Parse the class CLI parameter.
 fn parse_class(input: &str) -> Result<Class, String> {
-    match input.find(',') {
-        Some(position) => {
-            let general = input[position + 1..].to_owned();
-
-            // Warn the user if they've passed too many values.
-            if general.contains(',') {
-                return Err(String::from("Too many parameters"));
-            }
-
-            Ok(Class { instance: input[..position].into(), general })
+    let (general, instance) = match input.split_once(',') {
+        // Warn the user if they've passed too many values.
+        Some((_, instance)) if instance.contains(',') => {
+            return Err(String::from("Too many parameters"))
         },
-        None => Ok(Class { instance: input.into(), general: DEFAULT_NAME.into() }),
-    }
+        Some((general, instance)) => (general, instance),
+        None => (input, input),
+    };
+
+    Ok(Class::new(general, instance))
 }
 
 /// Convert to hex if possible, else decimal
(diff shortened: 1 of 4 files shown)


alacritty/src/cli.rs
```diff
@@ -161,16 +161,16 @@ fn option_as_value(option: &str) -> Result<Value, serde_yaml::Error> {
 fn parse_class(input: &str) -> Result<Class, String> {
     match input.find(',') {
         Some(position) => {
-            let general = input[position + 1..].to_owned();
+            let instance = input[position + 1..].to_owned();
```
chrisduerr, line 164: Not sure why we haven't done this, but seems like this would be much cleaner with `split_once`, right?

extra/alacritty.man
```diff
@@ -32,10 +32,12 @@ Increases the level of verbosity (the max level is \fB\-vvv\fR)
 Prints version information
 .SH "OPTIONS"
 .TP
-\fB\-\-class\fR <instance> | <instance>,<general>
+\fB\-\-class\fR <general> | <general>,<instance>
 Defines the window class hint on Linux [default: Alacritty,Alacritty]
 
-On Wayland the instance class sets the `app_id`, while the general class is ignored.
+When only general class was passed it'll set instance to the same value.
```
chrisduerr, line 38:
```suggestion
When only the general class is passed, instance will be set to the same value.
```

extra/alacritty.man
```diff
@@ -32,10 +32,12 @@ Increases the level of verbosity (the max level is \fB\-vvv\fR)
...
+
+On Wayland the instance class sets the `app_id`, while the instance class is ignored.
```
chrisduerr, line 40: On Wayland the general class sets the `app_id`, right? That's why `--class whatever` wasn't working?

chrisduerr (changes requested):

GreyXor:
lgtm

alacritty/src/cli.rs
kchibisov: It was too new when this was written.

extra/alacritty.man
kchibisov: Right.

chrisduerr: Seems like you didn't fix this?

chrisduerr (changes requested):

extra/alacritty.man
kchibisov: I did, at least I've applied you suggestion. Maybe github bugged for you?

chrisduerr: > On Wayland the instance class sets the `app_id`

It's the general class, not the instance class.

Also I didn't make a suggestion?

chrisduerr (approved):

jonaz:
@kchibisov you forgot to update the --help info? It now differs from man pages? 
```
OPTIONS:
        --class <instance> | <instance>,<general>
            Defines window class/app_id on X11/Wayland [default: Alacritty]
```
vs man
```
--class <general> | <general>,<instance>
```

kchibisov:
Seems like it.

PR #6310, opened 2022-09-07
Update alacritty-term+scanlines to align with macOS design language

samholmes:
This PR is to get the bounding size of the app icon to more closely follow the [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/foundations/app-icons/) for macOS icons

alacritty-term+scanlines
Screen Shot 2022-09-06 at 21 29 40


samholmes:
Explanation?

kchibisov:
Same as it was before for this kind of PR given that it's not a first time someone from macOS sents an icon just for macOS.

https://github.com/alacritty/alacritty/pull/4726

samholmes:
@kchibisov There appears to be no explanation there either.

Do we observe demand for a platform specific icon (macOS)? Yes. Why do we not meet this demand especially if the effort is willingly absorbed by community contributors? What's the reasoning?

kchibisov:
I thought that PR linked to that, but that's the issue.

https://github.com/alacritty/alacritty/issues/3926

The problem is that we don't want to maintain custom themes for all OSes. You're free to patch homebrew installation or your own installation to use whatever icon you feel like.

starkovv:
For those who want to change the default Alacritty app icon to sexier one that @samholmes suggested:

1. Download the new PNG from the original post.
2. Open that PNG in Preview app.
3. Click on the menu `File`, then hold `Option` key, then select `Save As`.
4. In opened window again hold `Option` key and click on file extensions drop-down list and select ICNS. Click `Save`.
5. In Finder, go to `/Applications/Alacritty.app/Contents/Resources`.
6. Replace existing `alacritty.icns` icon with the brand new one you just saved.
7. Restart your Alacritty app.

twhiting:
Just chiming in here. This seems like a small amount of work to improve the macos experience for alot of people.

It's a real pain to dig for instructions every time I refresh my macos image and need to fix the icon on macos. It has already wasted a good 30m - 1h of my life as i've done this at least 4 times by now.

Multiply that by how many users on MacOs and it seems like a small hill to die on for an improved experience. Small amount of work for a good image for the application :)

kchibisov:
If only macOS has a package manager that can provide a custom icon for you. A lot of linux distros do ship custom alacritty icon since they use KDE by default and KDE can't handle our SVG...

It just you're trying to fix the wrong problem here. The problem is that you want to ship custom icon via brew, but brew can't do anything, since it's not really a functional package manager... Like I don't see a reason to not patch source or even rebuild packages with their
authorized signatures, so users don't have issues with signing (like most linux package managers do).

PR #6318, opened 2022-09-09
Don't enable multisampling

kchibisov:
It never worked in the first place, because we were never requesting it in glutin, and it provides no value given that textures are already antialiased.


chrisduerr (approved):
Makes a lotta sense. Was wondering why we had that anyway.

PR #6333, opened 2022-09-14
Update unicode-width to Unicode 15

chrisduerr:
This applies all not-breaking dependency changes, including an update to unicode-width. This brings support for Unicode 15 with its new widths table so new glyphs and Emojis are shown correctly.


Cargo.lock
```diff
@@ -2039,9 +2036,9 @@ checksum = "c811ca4a8c853ef420abd8592ba53ddbbac90410fab6903b3e79972a631f7680"
 
 [[package]]
 name = "winit"
-version = "0.27.2"
```
kchibisov, line 2042: Winit requires manual changes in a separate branch.

chrisduerr: Yeah we can just get the other PR merged first.

Issue #6336, opened 2022-09-14
No application icon on Openbox


kitmonisit:
Related 
#1591: Missing Icon in KDE (breaks Freedesktop specification)
#2166: Icon not showing properly on Openbox

### System

OS: Linux
Version: alacritty 0.12.0-dev (bf3e8822)
Linux/BSD: X11 Openbox

### Logs

Crashes: None
Font/Terminal size: N/A
Keyboard and bindings: N/A

kitmonisit:
Font size in the tint2 taskbar is set to 64 x 64.

Wrong icon shows up in the tint2 panel when using the 64 x 64 [alacritty-term.png](https://github.com/kitmonisit/alacritty/blob/47b250be8f54c7b4d0589873b8953f4c347389cd/extra/logo/compat/alacritty-term.png)
![image](https://user-images.githubusercontent.com/1175380/190329642-882a1a87-f5e4-4bbe-a5d6-5c45ba4b907a.png)

Alt-tab display also shows the wrong icon
![image](https://user-images.githubusercontent.com/1175380/190330047-71776121-a676-46ea-b7a2-b178f167aea8.png)

chrisduerr:
Most likely an issue with either tint2 or you failed to install Alacritty properly. The icon works just fine.

kitmonisit:
I followed the instructions in `INSTALL.md`, but it still does not show up. Probably an issue with tint2.

If I change it to the 512 x 512 [alacritty-term+scanlines.png](https://github.com/kitmonisit/alacritty/blob/47b250be8f54c7b4d0589873b8953f4c347389cd/extra/logo/compat/alacritty-term+scanlines.png) on [alacritty/src/display/window.rs](https://github.com/kitmonisit/alacritty/blob/47b250be8f54c7b4d0589873b8953f4c347389cd/alacritty/src/display/window.rs), I get the icon, even if I do not do the extra steps in `INSTALL.md`

![image](https://user-images.githubusercontent.com/1175380/190331533-9190b55f-d730-40d7-8d64-4ba9e768a043.png)

chrisduerr:
I mean tint2 should probably follow the xdg icon spec for looking up icons to begin with. So the fact that changing `window.rs` does anything at all already kinda indicates that it's doing some weird shit.

kitmonisit:
Yes, tint2 is an application panel. I'm not sure which one does the Alt-Tab display, openbox or tint2.

kitmonisit:
[...]

PR #6337, opened 2022-09-14
Fixes #2166 and #6336: application icon with correct relative path

### LINKED ISSUE #2166: Icon not showing properly on Openbox
Which operating system does the issue occur on? Arch

I noticed that no matter what i do, the alacritty doesn't show with the "utilities-terminal" which is written on the .desktop file config, neither any other icon, even the default icon openbox uses (which is on /usr/share/pixmaps and also shows on some electrons applications i have, like SimpleNote, is not used for Alacritty), it only shows a default icon everywhere, like my taskbar, my app switcher and so on. Is there anything i can do, is this only happening here? Pic above of the blank icon on tint2 with alacritty open on taskbar. (the middle one, ofc)

2019-03-09-034253_126x50_scrot

EDIT: added the app switcher, so i guess the problem is not on tint, just something with alacritty and gtk icon not working around properly, i think

2019-03-09-034555_1920x1080_scrot

kitmonisit:
The original alacritty-term.png does not show up.  Thus changed it to alacritty-term+scanlines.png.

### DIFF
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -59,7 +59,7 @@ use crate::gl;
 
 /// Window icon for `_NET_WM_ICON` property.
 #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
-static WINDOW_ICON: &[u8] = include_bytes!("../../extra/logo/compat/alacritty-term.png");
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
This makes no sense.

PR #6355, opened 2022-09-23
Rebranding script for users who don't know how to write own engine

buahaha:
There is a script that will update files with a "brand" like Nutek, or MyApp, switch the icons, to not break the license and it's in python3.


chrisduerr (changes requested):
We will not add a script for rebranding Alacritty. This is a necessity to comply with licensing if you want to make your own thing out of Alacritty, but that doesn't mean we actively encourage and support doing so.

PR #6369, opened 2022-09-30
Fix icon decoding on X11

Hugal31:
glutin is waiting for a RGBA with 8-bit depth, but our icon is 16-bit depth. Normalize the color data when decoding the icon.

Without this, I had a `winit::icon::BadIcon::DimensionsVsPixelCount` when loading the icon on Linux/X11.

### DIFF
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -324,7 +324,8 @@ impl Window {
     pub fn get_platform_window(identity: &Identity, window_config: &WindowConfig) -> WindowBuilder {
         #[cfg(feature = "x11")]
         let icon = {
-            let decoder = Decoder::new(Cursor::new(WINDOW_ICON));
+            let mut decoder = Decoder::new(Cursor::new(WINDOW_ICON));
+            decoder.set_transformations(png::Transformations::normalize_to_color8());
             let mut reader = decoder.read_info().expect("invalid embedded icon");
             let mut buf = vec![0; reader.output_buffer_size()];
             let _ = reader.next_frame(&mut buf);
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -330,6 +330,7 @@ impl Window {
             let mut buf = vec![0; reader.output_buffer_size()];
             let _ = reader.next_frame(&mut buf);
             Icon::from_rgba(buf, reader.info().width, reader.info().height)
+                .expect("invalid embedded icon format")
         };
 
         let builder = WindowBuilder::new()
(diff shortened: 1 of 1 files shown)


chrisduerr:
Which Alacritty version did you build? I'm not running into any errors on Linux.

Hugal31:
I'm trying on the last upstream commit (e9ee8dc). A bisect shows the bug appears since 7d708d5.
You can run with this patch to make the error more visible:
```patch
diff --git a/alacritty/src/display/window.rs b/alacritty/src/display/window.rs
index 6c88f56f..582856e9 100644
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -331,6 +331,11 @@ impl Window {
             Icon::from_rgba(buf, reader.info().width, reader.info().height)
         };

+        if let std::result::Result::<&Icon, &glutin::window::BadIcon>::Err(err) = icon.as_ref() {
+            log::error!("could not load icon: {:?}", err);
+        }
+
         let builder = WindowBuilder::new()
             .with_title(&identity.title)
             .with_name(&identity.class.general, &identity.class.instance)
```

I'm using Ubuntu 20.04 with XFCE 4.14 if that helps.

chrisduerr:
> You can run with this patch to make the error more visible:

Ah okay, so the issue is just that the icon doesn't show up, it doesn't crash or anything?

chrisduerr (changes requested):
A big reason why this error went undetected is that we just pass `icon.ok()` to `with_window_icon` and thus any error is silently ignored and we just pretend that no icon was specified.

However, we **always** load and decode an icon on X11 and we're certain that this should work without any errors. So to avoid having a silent failure like this in the future, I'd actually change the code to be more verbose here.

We could return a result since we already `expect` the `read_info`, or alternatively we could also `expect` icon decode errors. Generally I think I'd expect returning a Result but I'm not sure if that can be easily handled by the caller.

Hugal31:
Yes, this cause no crash, it only silently fails to add the icon. In fact, on my setup, it shows the icon on the window decorator but not on the XFCE desktop miniatures somehow, but I don't know if it is specific to my setup.

I don't think we should return a result. Either we consider this is a "build" (not a runtime) error and panic (by `expect`ing the result), or we consider this can happen log a warning because a missing icon is not a big deal. But if we return a result, we will have to handle it and because this is the function supposed to create the WindowBuilder, we will still be left in a useless state.

Personally i'll go with a warning.

chrisduerr:
We do **not** want a warning here, since this should **always** succeed. The image we pass in is compiled into our application so it will either always fail or never. If it always fail we want it to be as noisy as possible so we can fix it asap.

chrisduerr (approved):

PR #6375, opened 2022-10-03
Fixed typo in comments

CosmosAtlas:


chrisduerr:
Fixing a singular typo isn't really a beneficial contribution to the project. That's just unnecessary noise.

CosmosAtlas:
I agree, not worth a separate commit. : )

Maybe integrate it with another commit that touches the same file.

Issue #6373, opened 2022-10-02
Make Control+Insert a default keybinding for Copy


heisencoder:
By default on Linux/Windows, alacritty does not bind the Control+Insert keyboard shortcut to Copy (or CopySelection), but does bind Shift+Insert to Paste.  This lacks symmetry and breaks the workflow of users who use the standard Control/Shift+Insert for copy/paste.  These shortcuts are particularly useful in the context of terminals because they are the same across most other Windows/Linux apps and provide a way to avoid the disaster of accidentally pressing Control+C in the terminal via muscle memory.

The request here is simple: Add Control+Insert with a default binding to the Copy action.

Note that I tried binding this action to CopySelection, but was unable to get it working to my satisfaction on Linux, so am suggesting just binding this to Copy.

### Specific suggested changes:

1. In `fn common_keybindings()` in bindings.rs, add this line:

```rust
    bindings!(
    ...
        Insert,   ModifiersState::CTRL, ~BindingMode::VI; Action::Copy;
    ...)
```

2. Add this default to the `key_bindings` section of `alacritty.yml`:

```yaml
key_bindings:
  ...
   # (Windows, Linux, and BSD only)
   # - { key: Insert,         mods: Control,                         action: Copy             }
```

Thanks for consideration of this feature request!  

### System

* OS: Linux/BSD/Windows (Not MacOS)
* Version: alacritty 0.10.1 ()
* Linux/BSD: X11, i3wm with compton

chrisduerr:
Control + Insert to *copy* makes exactly zero sense to me. Which other applications on Linux have this binding enabled by default?

heisencoder:
Control+Insert has a long history between both Linux and Windows, and it also has support among many Linux applications.

See this excellent article describing the history: https://www.howtogeek.com/804030/the-origins-of-ctrlc-ctrlv-ctrlx-and-ctrlz-explained/

When Windows 1.0 was first introduced in 1985, it used Control+Insert for Copy, and it wasn't until Windows 3.1 (in 1992) when it also supported Ctrl+C for Copy (based on the Apple Lisa and Later Macintosh line).  Linux came out around that same time and most of the Linux-based UIs copied many aspects from Windows, including the shortcuts for cut/copy/paste and undo.

[...]

PR #6380, opened 2022-10-07
Bind Ctrl+Ins to Copy on Windows/Linux

heisencoder:
See #6373


chrisduerr:
I'd just like to point out that I generally don't review draft PRs. So if you're hoping for feedback on this PR, please request a review or remove the draft status.

heisencoder:
> 

My apologies -- this is still intentionally in draft because I haven't been able to validate yet that this change works. It appears that the new change has no effect when I rebuild and run with no config.  I don't understand how Shift+Insert reliably Pastes, but the new Control+Insert default does not Copy.  I can, however, add a new line to the config to enable binding Control+Insert to Copy, so I'm currently very confused, but haven't had time to debug further.

chrisduerr:
Closing because dead draft.

PR #6390, opened 2022-10-12
Update glutin to 0.30.0

### LINKED ISSUE #1268: Couldn't find any pixel format that matches the criterias
Using Arch Linux with X11 which is fully up too date, an error occurs when trying to open alacritty.

```
Alacritty encountered an unrecoverable error:

Error creating GL context; Couldn't find any pixel format that matches the criterias.
```

I'm using a NVIDIA Geforce 760 GTX video card with the latest propietary video drivers (396.18).

kchibisov:
The glutin 0.30.0 update decouples glutin from winit which
provides us with basis for a multithreaded renderer. This
also improves robustness of our configuration picking,
context creation, and surface handling.

As an example we're now able to start on systems without a vsync,
we don't try to build lots of contexts to check if some config works,
and so on.

That also brings us possibility to handle context losses, but that's
a future work.

Fixes #1268.

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -335,7 +335,6 @@ impl DisplayUpdate {
 /// The display wraps a window, font rasterizer, and GPU renderer.
 pub struct Display {
     pub size_info: SizeInfo,
-    pub window: Window,
 
     /// Hint highlighted by the mouse.
     pub highlighted_hint: Option<HintMatch>,
@@ -365,23 +364,26 @@ pub struct Display {
     /// The ime on the given display.
     pub ime: Ime,
 
-    // OpenGL context.
-    context: Replaceable<PossiblyCurrentContext>,
-
-    // OpenGL surface.
-    surface: Surface<WindowSurface>,
-
     // Renderer.
     renderer: Renderer,
 
     // Mouse point position when highlighting hints.
     hint_mouse_point: Option<Point>,
 
+    surface: Surface<WindowSurface>,
+
+    // XXX OpenGL context must be dropped after the renderer.
+    context: Replaceable<PossiblyCurrentContext>,
+
     debug_damage: bool,
     damage_rects: Vec<DamageRect>,
     next_frame_damage_rects: Vec<DamageRect>,
     glyph_cache: GlyphCache,
     meter: Meter,
+
+    // XXX window should be dropped in the end, otherwise `surface` and other GL operations could
+    // have issues during freeing resources.
+    pub window: Window,
 }
 impl Display {
     pub fn new(
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -364,16 +364,8 @@ pub struct Display {
     /// The ime on the given display.
     pub ime: Ime,
 
-    // Renderer.
-    renderer: Renderer,
-
     // Mouse point position when highlighting hints.
     hint_mouse_point: Option<Point>,
-
-    surface: Surface<WindowSurface>,
-
-    // XXX OpenGL context must be dropped after the renderer.
-    context: Replaceable<PossiblyCurrentContext>,
 
     debug_damage: bool,
     damage_rects: Vec<DamageRect>,
@@ -381,8 +373,13 @@ pub struct Display {
     glyph_cache: GlyphCache,
     meter: Meter,
 
-    // XXX window should be dropped in the end, otherwise `surface` and other GL operations could
-    // have issues during freeing resources.
+    // XXX All fields below are ordered in the order they should be dropped.
+    renderer: Renderer,
+
+    surface: Surface<WindowSurface>,
+
+    context: Replaceable<PossiblyCurrentContext>,
+
     pub window: Window,
 }
 impl Display {
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -2,22 +2,22 @@
 //! GPU drawing.
 
 use std::fmt::{self, Formatter};
+use std::num::NonZeroU32;
+use std::ops::{Deref, DerefMut};
 #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
 use std::sync::atomic::Ordering;
 use std::{cmp, mem};
 
-use glutin::dpi::PhysicalSize;
-use glutin::event::ModifiersState;
-use glutin::event_loop::EventLoopWindowTarget;
-#[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
-use glutin::platform::unix::EventLoopWindowTargetExtUnix;
-use glutin::window::CursorIcon;
-use glutin::Rect as DamageRect;
-use log::{debug, info};
+use glutin::context::{NotCurrentContext, PossiblyCurrentContext};
+use glutin::prelude::*;
+use glutin::surface::{Rect as DamageRect, Surface, SwapInterval, WindowSurface};
+
+use log::{debug, info, warn};
 use parking_lot::MutexGuard;
 use serde::{Deserialize, Serialize};
-#[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
-use wayland_client::EventQueue;
+use winit::dpi::PhysicalSize;
+use winit::event::ModifiersState;
+use winit::window::CursorIcon;
 
 use crossfont::{self, Rasterize, Rasterizer};
 use unicode_width::UnicodeWidthChar;
@@ -33,9 +33,9 @@ use alacritty_terminal::term::color::Rgb;
 use alacritty_terminal::term::{self, Term, TermDamage, TermMode, MIN_COLUMNS, MIN_SCREEN_LINES};
 
 use crate::config::font::Font;
+use crate::config::window::Dimensions;
 #[cfg(not(windows))]
 use crate::config::window::StartupMode;
-use crate::config::window::{Dimensions, Identity};
 use crate::config::UiConfig;
 use crate::display::bell::VisualBell;
 use crate::display::color::List;
@@ -84,8 +84,8 @@ pub enum Error {
     /// Error in renderer.
     Render(renderer::Error),
 
-    /// Error during buffer swap.
-    Context(glutin::ContextError),
+    /// Error during context operations.
+    Context(glutin::error::Error),
 }
 
 impl std::error::Error for Error {
@@ -128,8 +128,8 @@ impl From<renderer::Error> for Error {
     }
 }
 
-impl From<glutin::ContextError> for Error {
-    fn from(val: glutin::ContextError) -> Self {
+impl From<glutin::error::Error> for Error {
+    fn from(val: glutin::error::Error) -> Self {
         Error::Context(val)
     }
 }
@@ -343,8 +343,7 @@ pub struct Display {
     /// Hint highlighted by the vi mode cursor.
     pub vi_highlighted_hint: Option<HintMatch>,
 
-    #[cfg(not(any(target_os = "macos", windows)))]
-    pub is_x11: bool,
+    pub is_wayland: bool,
 
     /// UI cursor visibility for blinking.
     pub cursor_hidden: bool,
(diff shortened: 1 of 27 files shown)


kchibisov:
The actual glutin release will be used after review, so I can breaking change glutin if needed for now.

alacritty/src/display/mod.rs
```diff
@@ -366,179 +374,89 @@ pub struct Display {
...
+
         let metrics = glyph_cache.font_metrics();
         let (cell_width, cell_height) = compute_cell_size(config, &metrics);
 
-        // Guess the target window size if the user has specified the number of lines/columns.
-        let dimensions = config.window.dimensions();
-        let estimated_size = dimensions.map(|dimensions| {
-            window_size(config, dimensions, cell_width, cell_height, estimated_scale_factor)
-        });
+        // Resize the window to account for user size.
```
chrisduerr, line 416: This comment got mangled. It shouldn't matter how big our users are. :smile:

alacritty/src/display/mod.rs
```diff
@@ -366,179 +374,89 @@ pub struct Display {
...
-        let (cell_width, cell_height) = if should_resize {
-            Self::update_font_size(&mut glyph_cache, scale_factor, config, font)
-        } else {
-            (cell_width, cell_height)
-        };
+        // Make the context current.
+        let context = context.make_current(&surface)?;
+
+        // We need to load them only once per instance, but only after we make our context current
+        // due to WGL limitations.
```
chrisduerr, line 442:
```suggestion
        // We need to load OpenGL functions once per instance, but only after we make our context current
        // due to WGL limitations.
```

"them" is kinda confusing if you read the comment before looking at the code.

alacritty/src/display/mod.rs
```diff
@@ -582,13 +500,14 @@ impl Display {
...
-            StartupMode::Maximized if is_x11 => window.set_maximized(true),
+            #[cfg(not(any(target_os = "macos", windows)))]
+            StartupMode::Maximized if !is_wayland => window.set_maximized(true),
             _ => (),
         }
 
         let hint_state = HintState::new(config.hints.alphabet());
-        let is_damage_supported = window.swap_buffers_with_damage_supported();
+
+        let is_damage_supported = is_wayland;
```
chrisduerr, line 510: `is_wayland` is shorter than `is_damage_supported`. So might as well check for `is_wayland` directly?

If another platform adds support in the future, we can always change it again.

alacritty/src/display/mod.rs
```diff
@@ -597,8 +516,20 @@ impl Display {
             (Vec::new(), Vec::new())
         };
 
+        // We use vsync everywhere except wayland.
+        if !is_wayland {
+            if let Err(err) =
+                surface.set_swap_interval(&context, SwapInterval::Wait(NonZeroU32::new(1).unwrap()))
+            {
+                // This is not really an error we should broadcast to the user.
```
chrisduerr, line 524: This comment is a bit strange, stating we shouldn't do something without any explanation.

And I'm also questioning whether the statement is even true. If vsync could not be set, it seems like something very relevant to broadcast? I'd assume this could happen when vsync was disabled in the driver? That's definitely something the user might care about, since we don't operate correctly without vsync.

alacritty/src/display/mod.rs
```diff
@@ -622,6 +553,37 @@ impl Display {
         })
     }
 
+    #[inline]
+    pub fn gl_context(&self) -> &PossiblyCurrentContext {
+        self.context.get()
+    }
+
+    pub fn make_not_current(&mut self) {
+        if self.context.get().is_current() {
```
chrisduerr, line 562: What's the benefit of checking if the context is current here? Do things explode when trying to make a not-current context not-current, or is this just an efficiency thing?

alacritty/src/display/mod.rs
```diff
@@ -622,6 +553,37 @@ impl Display {
...
+            self.context.replace_with(|context| {
+                context.make_not_current().expect("failed to disable context").treat_as_current()
```
chrisduerr, line 564: "make_not_current" -> "treat_as_current" sounds **very** wrong.

alacritty/src/event.rs
```diff
@@ -1327,21 +1339,167 @@ impl Processor {
         }
     }
 
+    /// Create the GL display.
+    fn create_gl_display(
```
chrisduerr, line 1343: Questionable these OpenGL functions should be in `alacritty/src/event.rs`? Maybe move to display?

alacritty/src/event.rs
```diff
@@ -1327,21 +1339,167 @@ impl Processor {
...
+        let preference = DisplayApiPreference::Cgl;
+
+        #[cfg(windows)]
+        let preference = DisplayApiPreference::Wgl(Some(_raw_window_handle.unwrap()));
+
+        #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
+        let preference = DisplayApiPreference::GlxThenEgl(Box::new(unix::register_xlib_error_hook));
+
+        #[cfg(all(not(feature = "x11"), not(any(target_os = "macos", windows))))]
+        let preference = DisplayApiPreference::Egl;
```
chrisduerr, line 1357: Why do you have to do all this manually? The goal for `glutin` is to provide an abstraction, right? Couldn't it provide a `PlatformApiPreference` that automatically works?

alacritty/src/event.rs
```diff
@@ -1327,21 +1339,167 @@ impl Processor {
...
+        proxy: EventLoopProxy<Event>,
+        options: WindowOptions,
+    ) -> Result<(), Box<dyn Error>> {
+        let raw_display_handle = event_loop.raw_display_handle();
+
+        let mut identity = self.config.window.identity.clone();
+        options.window_identity.override_identity_config(&mut identity);
+
+        // Windows has different order of GL platform initialization compared to any other platform,
+        // it requires window first.
```
chrisduerr, line 1435:
```suggestion
        // it requires the window first.
```

alacritty/src/event.rs
```diff
@@ -1327,21 +1339,167 @@ impl Processor {
...
+        #[cfg(windows)]
+        let window = Some(Window::new(event_loop, &self.config, &identity)?);
+        #[cfg(not(windows))]
+        let window: Option<Window> = None;
+
+        let raw_window_handle = window.as_ref().map(|w| w.raw_window_handle());
```
chrisduerr, line 1441: What's the point of the as_ref().map when it's always going to be None on macos/linux and some on windows anyway? Don't need to create a temporary `window` variable at all.

alacritty/src/event.rs
```diff
@@ -1327,21 +1339,167 @@ impl Processor {
...
+            Window::new(
+                event_loop,
+                &self.config,
+                &identity,
+                #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
+                self.wayland_event_queue.as_ref(),
+                #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
+                gl_config.x11_visual(),
+            )?
+        };
```
chrisduerr, line 1457: Just `cfg` this? The branch here is just confusing.

alacritty/src/event.rs
```diff
@@ -1327,21 +1339,167 @@ impl Processor {
...
         event_loop: &EventLoopWindowTarget<Event>,
         proxy: EventLoopProxy<Event>,
         options: WindowOptions,
     ) -> Result<(), Box<dyn Error>> {
-        let window_context = WindowContext::new(
-            self.config.clone(),
-            &options,
+        // Get any window and take its GL config and display to build a new context.
+        let (gl_display, gl_config) = {
+            let gl_context = self.windows.iter().next().unwrap().1.display.gl_context();
```
chrisduerr, line 1479:
```suggestion
            let gl_context = self.windows.first().unwrap().1.display.gl_context();
```

chrisduerr (changes requested):
Overall doesn't look as bad as I expected tbh. Just a lot of Glutin stuff bleeding into Alacritty now in `event.rs`.

alacritty/src/display/mod.rs
kchibisov: We just did that in the past, I'd assume that it could be faster?

alacritty/src/display/mod.rs
kchibisov: Maybe the name is wrong? We're doing that already, nothing new here. Given that your context is possibly current and you basically tell it to stay like that and not go into `NotCurrent`.

alacritty/src/event.rs
kchibisov: Because some requires things like `unix::register_xlib_error_hook` which may not be present at all, or `RawWindowHandle` without a reason. I thought how to abstract it in a way that you can still use glutin for example without WGL and use that interface, but I have no real ideas...

If you have any let me know.

If we could have preferences without extra data it could provide something like that.

alacritty/src/event.rs
kchibisov: What do you mean? I need to initialize window on windows before doing anything else and keep it around? You mean that I don't need that variable on non windows? Then yes, but it's less code to write that way and compiler will optimize out that `None` parts.

alacritty/src/event.rs
kchibisov: It's a `HashMap` there's no `first`.

alacritty/src/event.rs
kchibisov: I've added just more cfgs here, we'd still need `RWHD` though, so I just assign it to `None`.

alacritty/src/display/mod.rs
kchibisov: Seems fine.

alacritty/src/display/mod.rs
kchibisov: It was like that before IIRC, it tries to say user configured size.

alacritty/src/display/mod.rs
kchibisov: I'll keep warn for now, since the errors about us crashing due to missing vsync is close to `None`.

kchibisov:
Just as a note, that would need testing on macOS, windows, in vm, and on things like pinephone to check for update parity.

kchibisov:
Windows/macOS tested and seems to work fine with multi windows.

alacritty/src/display/mod.rs
```diff
@@ -582,23 +472,35 @@ impl Display {
...
         } else {
             (Vec::new(), Vec::new())
         };
 
+        // We use vsync everywhere except wayland.
+        if !is_wayland {
+            if let Err(err) =
+                surface.set_swap_interval(&context, SwapInterval::Wait(NonZeroU32::new(1).unwrap()))
+            {
+                // In general we expect vsync to be present for optimal working.
```
chrisduerr, line 495:
```suggestion
```

The warning is already there, no need for the comment.

alacritty/src/display/mod.rs
chrisduerr: Maybe the function should be renamed to `treat_as_possibly_current`? That way it's clear that we don't assume it's current but we just treat it as a context that might be?

alacritty/src/display/mod.rs
```diff
@@ -607,21 +509,50 @@ impl Display {
...
+    pub fn make_not_current(&mut self) {
+        if self.context.get().is_current() {
+            self.context.replace_with(|context| {
+                context.make_not_current().expect("failed to disable context").treat_as_current()
+            });
+        }
+    }
+
+    pub fn make_current(&self) {
+        self.context.make_current(&self.surface).expect("failed to make context current.")
```
chrisduerr, line 539: We have a check in `make_not_current` to ensure it is current, but we have no check in here to ensure it is not current. Seems a bit strange I'd say unless we have a good reason for this. Otherwise I'd prefer consistency.

alacritty/src/renderer/platform.rs
```diff
@@ -0,0 +1,111 @@
+//! The graphics platform that is used by the renderer.
```
chrisduerr, line 1:
```suggestion
//! The graphics platform that is used by the renderer.

```

alacritty/src/renderer/platform.rs
```diff
@@ -0,0 +1,111 @@
...
+) -> GlutinResult<NotCurrentContext> {
+    let context_attributes = ContextAttributesBuilder::new()
+        .with_context_api(ContextApi::OpenGl(Some(Version::new(3, 3))))
+        .build(raw_window_handle);
+
+    unsafe {
+        if let Ok(gl_context) = gl_display.create_context(gl_config, &context_attributes) {
+            Ok(gl_context)
+        } else {
+            // Fallback.
```
chrisduerr, line 84: This comment provides no value. We know it's a fallback since it's in the `else` branch of an error handler. If you want to keep this comment, you should provide more context.

alacritty/src/window_context.rs
```diff
@@ -68,33 +75,111 @@ pub struct WindowContext {
...
+        let mut identity = config.window.identity.clone();
+        options.window_identity.override_identity_config(&mut identity);
+
+        // Windows has different order of GL platform initialization compared to any other platform,
+        // it requires the window first.
+        #[cfg(windows)]
+        let window = Window::new(event_loop, &config, &identity)?;
+
+        #[cfg(windows)]
+        let raw_window_handle = Some(window.raw_window_handle());
```
chrisduerr, line 98:
```suggestion
        // Windows has different order of GL platform initialization compared to any other platform,
        // it requires the window first.
        #[cfg(windows)]
        let window = Window::new(event_loop, &config, &identity)?;
        #[cfg(windows)]
        let raw_window_handle = Some(window.raw_window_handle());
```

Probably better to keep the Windows stuff bundled together. Avoids accidentally having something written in between.

alacritty/src/window_context.rs
```diff
@@ -68,33 +75,111 @@ pub struct WindowContext {
...
+        options: WindowOptions,
+        #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
+        wayland_event_queue: Option<&EventQueue>,
+    ) -> Result<Self, Box<dyn Error>> {
+        let raw_display_handle = event_loop.raw_display_handle();
+
+        let mut identity = config.window.identity.clone();
+        options.window_identity.override_identity_config(&mut identity);
+
+        // Windows has different order of GL platform initialization compared to any other platform,
```
chrisduerr, line 92:
```suggestion
        // Windows has different order of GL platform initialization compared to any other platform;
```

alacritty/src/event.rs
chrisduerr: Yeah I like it better this way.

chrisduerr (changes requested):

kchibisov:
This is now uses the latest beta of glutin instead of git commit, which should be the latest one before the release.

alacritty/src/display/mod.rs
```diff
@@ -607,21 +508,55 @@ impl Display {
...
+                    .make_not_current()
+                    .expect("failed to disable context")
+                    .treat_as_possibly_current()
+            });
+        }
+    }
+
+    pub fn make_current(&self) {
+        if !self.context.get().is_current() {
+            self.context.make_current(&self.surface).expect("failed to make context current.")
```
chrisduerr, line 542:
```suggestion
            self.context.make_current(&self.surface).expect("failed to make context current")
```

Just tiny consistency thing.

chrisduerr (changes requested):
Super tiny thing, but other than that we should be all good.

chrisduerr (approved):

kchibisov:
Will merge after glutin release.

alacritty/src/display/mod.rs
```diff
@@ -365,23 +364,26 @@ pub struct Display {
...
+
     debug_damage: bool,
     damage_rects: Vec<DamageRect>,
     next_frame_damage_rects: Vec<DamageRect>,
     glyph_cache: GlyphCache,
     meter: Meter,
+
+    // XXX window should be dropped in the end, otherwise `surface` and other GL operations could
+    // have issues during freeing resources.
+    pub window: Window,
```
chrisduerr, line 386: I'd definitely place all the things that require manual dropping at the end together. Splitting it up like this seems like a recipe for disaster.

This seems fairly complicated so I'd be fine with a manual `Drop` impl too.

chrisduerr (changes requested):

kchibisov:
@chrisduerr I've updated with `ManuallyDrop`.

alacritty/src/display/mod.rs
```diff
@@ -369,176 +370,64 @@ pub struct Display {
...
+
+    context: ManuallyDrop<Replaceable<PossiblyCurrentContext>>,
+
     debug_damage: bool,
     damage_rects: Vec<DamageRect>,
     next_frame_damage_rects: Vec<DamageRect>,
-    renderer: Renderer,
     glyph_cache: GlyphCache,
     meter: Meter,
 }
```
chrisduerr, line 384:
```suggestion
}

```

chrisduerr (changes requested):

chrisduerr (approved):

Issue #6213, opened 2022-07-18
Problem with the default color scheme


sshock:
As someone who has used Alacritty for a while now, and been involved in [various discussions](https://github.com/alacritty/alacritty/issues?q=commenter%3Asshock), and even made an important [contribution](https://github.com/rust-windowing/winit/pull/2239) to fix a [bug](https://github.com/alacritty/alacritty/issues/5876), please hear me out.

Let me start by saying that I recognize there is no "right" or "wrong" when it comes to a default color scheme.  Indeed there are many [different defaults](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors) that have been used over time.  So of course Alacritty is free to choose whatever scheme and that doesn't make it "wrong".

But it also doesn't mean that the choice that has been made is the best choice, or even a good choice.

I would like to give a concrete reason for why the "Tomorrow Night" scheme is not a good choice for a default.  And it's not because the [green is yellowish](https://github.com/alacritty/alacritty/issues/1561) or because red and bright red seem backwards to me.

It's because the colors are too close and that breaks usability.

The problem is several colors are not distinct enough to make them distinguishable, especially the bright vs non-bright of a color.  They are too similar, and that makes some apps unusable due to the text being unreadable.

Here are Alacritty's default colors:

<img width="335" alt="alacritty 256" src="https://user-images.githubusercontent.com/1693761/179561271-22d95e6d-81a1-4ce4-988e-b22351670d1e.png">

In contrast, here is my current color scheme, which I consider to be more sane:

<img width="356" alt="good 256" src="https://user-images.githubusercontent.com/1693761/179561327-a0ff784d-e1b7-432f-abd9-2cd15ce6e0c3.png">

It may not be readily apparent what is the big problem.  After all, Alacritty's colors do look pretty smooth and nice.  And the color scheme I am using is positively garish.

Hopefully the next picture with every color on top of every color will help illustrate the problem:

<img width="735" alt="alacritty 16" src="https://user-images.githubusercontent.com/1693761/179563316-cf6a0b5f-7bf2-419f-b937-6c35a191daf8.png">

And here is the color scheme I am using for comparison:

<img width="738" alt="good 16" src="https://user-images.githubusercontent.com/1693761/179563358-5cf513a8-4d08-4f88-b7da-2d5d0c196dfd.png">

Maybe the problem still isn't apparent, but let's take a look at some specific spots, starting with green on green:

[...]

PR #6438, opened 2022-10-22
Change default bright colors to bright versions of the normal colors

sshock:
Set the bright colors to the normal colors ("Tomorrow Night") multiplied by a brightness factor of 1.34 (capped at 100%).

This follows the proposed approach given in the [original issue](https://github.com/alacritty/alacritty/issues/3404) from Mar 2020 to rework the default bright colors.

More precisely, take each normal color, convert RGB to HSV, then multiply the V by 1.34, and finally convert back to RGB.

There are just two minor deviations:
1. The normal yellow already has a brightness of 94.1%, so changing the brightness to 100% doesn't result in a big enough difference; so for bright yellow the saturation was also changed, from 51.7 to 40, which gives it a noticeable difference.
2. For bright black the brightness was just changed from 12.9% to 40%; 40% is exactly the same brightness level we are currently using; the formula just doesn't work to go from black to bright black because multiplying black brightness by 1.34 doesn't really change it much.

The new color scheme has all of the following advantages:
* Bright colors are now easily distinguishable from their normal counterparts.
* Bright red is now brighter than normal red instead of darker.
* Bright cyan is now brighter than normal cyan instead of being almost identical and actually slightly darker.
* Overall contrast is greatly improved across the board (see bar chart below).

Luminance calculations (using formula `0.2126*R + 0.7152*G + 0.0722*B`) for Red and Cyan:
* **Red BEFORE:**
  * Normal = 0.23299420863212106
  * Bright = 0.20219445780928133
* **Red AFTER:**
  * Normal = 0.23299420863212106
  * Bright = 0.3825685577896843
* **Cyan BEFORE:**
  * Normal = 0.4564908235539082
  * Bright = 0.4431834311169868
* **Cyan AFTER:**
  * Normal = 0.4564908235539082
  * Bright = 0.8842685999154053

Here is a visual comparison of the current (old) colorscheme versus the new one with this PR.  The screenshots show all color combinations:

**OLD**

old

**NEW**

new

Finally, here is a bar chart showing the contrast delta (X axis has a point for each color combination; Y axis shows the new contrast minus the old contrast):

contrast_delta

See https://github.com/alacritty/alacritty/issues/6213 for more details.

Some history:

The original issue was "solved" by incorporating the 8 colors from "Tomorrow Night Bright" as the bright colors, but there was no logical reason for doing that as that color scheme is just another nice 8-color scheme and has no correlation to "Tomorrow Night" other than having a similar name.  Any other random color scheme could have been chose for the bright colors and would have resulted in the same problem.

As a result, the original issue was only partially solved as many bright colors remained hard to distinguish from their normal counterpart, since both the hue and brightness remained very close (and yes it makes sense for normal and bright to have the same hue, but not the same brightness).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.12.0-dev
 
+### Changed
+
+- Default bright colors are now derived from `Tomorrow Night` rather than using `Tomorrow Night Bright`
+
 ### Fixed
 
 - `--help` output for `--class` does not match man pages
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -293,14 +293,14 @@
 
   # Bright colors
   #bright:
-  #  black:   '#666666'
-  #  red:     '#d54e53'
-  #  green:   '#b9ca4a'
-  #  yellow:  '#e7c547'
-  #  blue:    '#7aa6da'
-  #  magenta: '#c397d8'
-  #  cyan:    '#70c0b1'
-  #  white:   '#eaeaea'
+  #  black:   '#5a6066'
+  #  red:     '#ff8080'
+  #  green:   '#f2fd8b'
+  #  yellow:  '#ffdd99'
+  #  blue:    '#add8ff'
+  #  magenta: '#eec6fa'
+  #  cyan:    '#b9fff5'
+  #  white:   '#fbfffc'
 
   # Dim colors
   #
--- a/alacritty/src/config/color.rs
+++ b/alacritty/src/config/color.rs
@@ -227,14 +227,14 @@ pub struct BrightColors {
 impl Default for BrightColors {
     fn default() -> Self {
         BrightColors {
-            black: Rgb { r: 0x66, g: 0x66, b: 0x66 },
-            red: Rgb { r: 0xd5, g: 0x4e, b: 0x53 },
-            green: Rgb { r: 0xb9, g: 0xca, b: 0x4a },
-            yellow: Rgb { r: 0xe7, g: 0xc5, b: 0x47 },
-            blue: Rgb { r: 0x7a, g: 0xa6, b: 0xda },
-            magenta: Rgb { r: 0xc3, g: 0x97, b: 0xd8 },
-            cyan: Rgb { r: 0x70, g: 0xc0, b: 0xb1 },
-            white: Rgb { r: 0xea, g: 0xea, b: 0xea },
+            black: Rgb { r: 0x5a, g: 0x60, b: 0x66 },
+            red: Rgb { r: 0xff, g: 0x80, b: 0x80 },
+            green: Rgb { r: 0xf2, g: 0xfd, b: 0x8b },
+            yellow: Rgb { r: 0xff, g: 0xdd, b: 0x99 },
+            blue: Rgb { r: 0xad, g: 0xd8, b: 0xff },
+            magenta: Rgb { r: 0xee, g: 0xc6, b: 0xfa },
+            cyan: Rgb { r: 0xb9, g: 0xff, b: 0xf5 },
+            white: Rgb { r: 0xfb, g: 0xff, b: 0xfc },
         }
     }
 }


chrisduerr:
Honestly the improvements don't really seem significant enough to justify making any changes and the process used to generate these changes is flawed anyway.

I'm curious what @kchibisov thinks about them, if he feels like this would make for a significant improvement. But I don't see this solving anything realistically.

kchibisov:
>I'm curious what @kchibisov thinks about them, if he feels like this would make for a significant improvement. But I don't see this solving anything realistically.

I have issues distinguishing colors with the new proposed theme, especially bright green and bright yellow are basically the same on small characters. I don't have issues like that with our default theme.

I also can't really repro the theme contrast at all, since I use smaller font I can't see anything when using white as background, since everything is invisible for me, but it shouldn't matter. The contrast with blue backgrounds is also strange.

From my experience in terminal you mostly all the time deal with:

1. Things on blue background (a lot ncurses apps look like that).
2. Red, green, and yellow(and their bright variants) mixed together. So they should clearly be distinguishable from each other.
3. Other colors are usually used against black background, rarely against blue.
4. The only colored backgrounds I'm aware of are `blue`, `white`, and `default bg`. Other colors are not used for that. At least I've never seen them being used.

While this theme tries to be better on average It's not better on the most frequently used color combinations, which is worse in my opinion. So I'm not sure we are getting with it anywhere.

The theme that has best contrast that I know of is dark gruvbox(don't use the provided ones, you need to alter white on it) with `high-contrast` bg. It was designed around 16 colors and on nearly every combination is clearly visible and distinguishable. But I'm not saying that it should be changed to use it, since it's arbitrary.

sshock:
> Honestly the improvements don't really seem significant enough to justify making any changes

The improvements are very significant.  Several of the bright colors are way too close to the normal colors right now, making them indistinguishable.  It is important for the brights to be distinguishable from their normal counterparts (even when they're not on top of each other), otherwise you wouldn't have created issue https://github.com/alacritty/alacritty/issues/3404 in the first place.

Also, I have the numbers that clearly show an enormous improvement in contrast across the board.

So we have:
* Matching user expectations of having bright colors actually be brighter than the normal colors.
* A very significant improvement in contrast for usability.
* Normal colors remain unchanged; only brights are affected.

> and the process used to generate these changes is flawed anyway.

* What's flawed about it?
* It's **your own strategy!** that **you recommended** in issue https://github.com/alacritty/alacritty/issues/3404
* It's the same strategy as is being used for the dim colors (in the opposite direction).

What's flawed is the idea that you can take two unrelated 8-color color schemes and use one of them for the 8 normals and the other one for 8 brights.  That's a flawed strategy that results in bright colors not being brighter than the normals.  TN and TNBright, despite having similar names, are just two different 8-color color schemes; the colors in TNBright are not bright versions of the colors in TN, and they are not all brighter.  Many of them are very close in brightness and some are even darker.  This is easily explained as a flawed strategy.

> I'm curious what @kchibisov thinks about them, if he feels like this would make for a significant improvement. But I don't see this solving anything realistically.

I look forward to hearing from @kchibisov as well.

sshock:
> I have issues distinguishing colors with the new proposed theme, especially bright green and bright yellow are basically the same on small characters. I don't have issues like that with our default theme.

Bright green and bright yellow indeed don't have very good contrast, though based on the contrast calculation, their contrast is actually worse in the current color scheme with a contrast of 1.0745, versus the new contrast of 1.1958.

But despite the calculation, I'll give this one to you because even for myself, I find this color combination easier to read with the old (current) scheme.

> I also can't really repro the theme contrast at all, since I use smaller font I can't see anything when using white as background, since everything is invisible for me, but it shouldn't matter.

Are you saying the colors are harder to read on a white background?  I get that.  With that one I am seeing a mixed bag with some being way easier to read in the new scheme (like +GRN/WHT) while others got worse (like +BLU/WHT).

FWIW, Here's what the contrast calculations show for colors with a white background:

on_white

> The contrast with blue backgrounds is also strange.

Can you be more specific than "strange"?

> From my experience in terminal you mostly all the time deal with:
> 
>     1. Things on blue background (a lot ncurses apps look like that).

The readability of colors on blue background is WAY better in the new color scheme.  This is a HUGE win:

blues

>     2. Red, green, and yellow(and their bright variants) mixed together. So they should clearly be distinguishable from each other.

Some of these have improved and some got worse.  Here is the full spreadsheet with contrast calculations:

[contrast_comparison.ods](https://github.com/alacritty/alacritty/files/9893633/contrast_comparison.ods)

>     3. Other colors are usually used against black background, rarely against blue.

All of the colors continue to be very readable on top of black.

blacks

>     4. The only colored backgrounds I'm aware of are `blue`, `white`, and `default bg`. Other colors are not used for that. At least I've never seen them being used.
> 
> While this theme tries to be better on average It's not better on the most frequently used color combinations, which is worse in my opinion. So I'm not sure we are getting with it anywhere.

* It's better on black (the most common bg).
* It's better on blue (next most common bg).
* It's half better and half worse on white.

So overall, BETTER!!

> The theme that has best contrast that I know of is dark gruvbox(don't use the provided ones, you need to alter white on it) with `high-contrast` bg. It was designed around 16 colors and on nearly every combination is clearly visible and distinguishable. But I'm not saying that it should be changed to use it, since it's arbitrary.

Yeah, personally I use a theme with a very excellent contrast, but it is also very plain and ugly, so I purposely decided to keep with the Tomorrow Night theme and only fix the bright colors to keep the change minimal.

kchibisov:
The approach is flowed in a first place since we compare the colors in a perfect world, however the situation is that on real monitors they look all differently. I do prefer old theme since it's something that was designed and not generated, though the bright colors are arbitrary, but at least they are clear. With the new theme it doesn't matter how good it is on paper the end result when you look at it is worse and that's all that matters. Even you concluded that it's hard to see the green/yellow combinations.

sshock:
> The approach is flowed in a first place since we compare the colors in a perfect world, however the situation is that on real monitors they look all differently. I do prefer old theme since it's something that was designed and not generated, though the bright colors are arbitrary, but at least they are clear. With the new theme it doesn't matter how good it is on paper the end result when you look at it is worse and that's all that matters. Even you concluded that it's hard to see the green/yellow combinations.

If generating the bright colors is flawed, then why was it not flawed to generate the dim colors?

The bright green / bright yellow is hard to see, but that is just one color combination; most color combinations improved and are more readable, including most colors on the most common backgrounds of blue and black that you specifically called out.

Take a look at a side by side comparison (current on left, new on right) of the colors on the most common background, black.

blacks

All of these colors on black are very nice and still very readable!

Now take a look at a side by side comparison of the colors on the very common background, blue:

blues

The new colors are WAY better and more readable!!

Finally, for completeness here are the colors on white background, where I admit some improved and some got worse:

whites

But overall with the white background, we really only end up with one color that is really bad and hard to read (+BLU).  But the old color scheme also had one color that was really bad and hard to read on white (+GRN).

So I argue that it is incorrect to say "the end result when you look at it is worse", at least if you're talking about contrast and readability, which is what matters most for a default color scheme.  The colors on white background are a wash with some better and some worse, and only one bad before and after, and the colors on black have better contrast, but I'll ignore this since they already had good contrast, so that leaves blue, but with blue the new color scheme clearly wins with much better readability!

Why aren't these two things important to you for new users?
* The expectation that bright colors actually be brighter than the normal colors.
* The readability (and therefore usability) of colors on a very common blue background.

If generating the brights is so horribly flawed, then you should work on picking a new designed 16-color color scheme.  But in the meantime, I think it would be wise to merge this PR because it really does improve readability and user expectations.

kchibisov:
>If generating the bright colors is flawed, then why was it not flawed to generate the dim colors?

dim colors are used so rarely that no-one cares about them.

Besides test for me the Red, Green, and Yellow are better and much easier to read and when they are adjacent to each other in
text. The bright green and yellow for me a just tinted white. With worse LCDs you'd have both green and yellow just the same.

> the end result when you look at it is worse

By that I mean when you look at them on average LCD panels, for me on 2 of my monitors with sRGB profiles bright yellow and green are basically the same color. And I have pretty common panels. The problem is not contrast against background and not with large font sizes like you use, but with small font sizes and 1 DPR monitors.

sshock:
> By that I mean when you look at them on average LCD panels, for me on 2 of my monitors with sRGB profiles bright yellow and green are basically the same color. And I have pretty common panels. The problem is not contrast against background and not with large font sizes like you use, but with small font sizes and 1 DPR monitors.

Alright, so I went ahead and tested 4 LCD monitors just now, a lenovo laptop, a macbook pro, and two acer monitors for my desktop, one of which I went in and switched the color profile to sRGB.

On all four I made the font small and tested bright green on bright yellow and vice-versa.

In all four cases, with the new color scheme I was still able to read the text, whereas with the old (current) color scheme it was washed out and I could not read it!

I don't know what's going on with all your monitors (and I've never even heard of a DPR monitor) is, but do you think it would be fair if we asked a few other people to test this scenario?

kchibisov:
DPR meas device pixel ratio. I was talking about lower device pixel ratios. Yeah, if the majority who is using the default theme would like your changes I'm not that against, but maybe changed the yellow/green colors a bit.

I'm just used to more warmer themes and looking at our default cold theme and more colder updated one is just hard for me.

sshock:
> DPR meas device pixel ratio. I was talking about lower device pixel ratios. Yeah, if the majority who is using the default theme would like your changes I'm not that against, but maybe changed the yellow/green colors a bit.

Ok, cool.  Here is a test I created with small text of bright green on bright yellow and vice-versa.

Would you mind tagging a few people and asking them whether "Before" or "After" is easier to read?

test2

I'm not sure what I was thinking when earlier I said "despite the calculation, I'll give this one to you because even for myself, I find this color combination easier to read with the old (current) scheme."

Because now as I am testing, and also the smaller I make the text, the more pronounced it becomes for me that this new color scheme actually greatly improves the readability of these two colors against each other.

> I'm just used to more warmer themes and looking at our default cold theme and more colder updated one is just hard for me.

Is it because the brights are too bright and makes it feel harsh?  I wonder if I were to redo it but use a smaller brightness factor like 1.25 instead of 1.34 if that would be more pleasing?  I could try that if you would like.

kchibisov:
The point is to not compare against, but near each other. The point is that you can't really say whether it's green or not with bright variants. If you have color issues it'll be the same color for you.

kchibisov:
But I mean that's likely true with the current colors as well.

sshock:
> The point is to not compare against, but near each other. The point is that you can't really say whether it's green or not with bright variants. If you have color issues it'll be the same color for you.

Oh, I see.  I'm sorry I didn't understand what you were actually talking about until now.

I'm trying it now, and yeah, they do seem pretty close:

small_bright_green_vs_bright_yellow

> But I mean that's likely true with the current colors as well.

Yeah, I never brought this up before because it didn't bother me too much, but one of the things I've always noticed about Tomorrow Night is that the green is actually quite yellow-ish, like a puke green that is halfway between green and yellow, which makes them less distinguishable.

And I can see how green and yellow could look even more similar in the bright version.

So I think it's kinda an existing problem too, but maybe it's not as bad when the greens and yellows don't actually get brighter.  Anyway, perhaps just a small tweak here would help.

chrisduerr:
> It's your own strategy! that you recommended

Only because it was my suggestion as most likely to be fruitful, doesn't mean it's flawless. I think I've been pretty clear from the beginning that continuously changing the colorscheme is unlikely to be a beneficial effort.

I'm also not claiming that this PR is not an improvement at all. But how much more bikeshedding about the colorscheme will there be? This discussion alone already makes it pretty clear that the improvements to the colorscheme are moderate at best and more changes might be necessary to improve things even further. So clearly that would just drive further and further bikeshedding.

I personally have zero interest in changing the colorscheme in Alacritty every other year. So I don't think it's a good idea to merge anything unless most people can agree that the colorscheme definitely doesn't need any more tweaking after that.

sshock:
> I'm also not claiming that this PR is not an improvement at all. But how much more bikeshedding about the colorscheme will there be? This discussion alone already makes it pretty clear that the improvements to the colorscheme are moderate at best and more changes might be necessary to improve things even further. So clearly that would just drive further and further bikeshedding.

Yeah, although to me the improvements here are more than moderate, and I thought I had found a great solution we could all get behind, and I do think we reached a pretty good place where both you and @kchibisov do not disagree with the specific improvements I have called out, at the end of the day for you the improvements here are not significant enough and for @kchibisov there are a couple specific issues with the new scheme, so it's just not going to work because like you I don't think I really want to spend the time going back and forth with little tweaks.

> I personally have zero interest in changing the colorscheme in Alacritty every other year. So I don't think it's a good idea to merge anything unless most people can agree that the colorscheme definitely doesn't need any more tweaking after that.

Yeah, I totally get it.  I'll close this out.

chrisduerr:
Btw, I still want to thank you for looking into this @sshock. You certainly did your due diligence exploring the field with available opportunities.

Right now I don't think the situation is dire enough to redesign our colors, considering I rarely ever get any negative feedback about them. But chances are if there's ever a motion away from our current theme, it would likely be to the theme of one of the maintainers since that way you'll probably have one person convinced already at least (though I know at least my theme isn't fit for widestream adoption), given that that theme is somewhat friendly for newcomers and provides better contrasts.

Issue #6440, opened 2022-10-23
Top padding, or content vertical align


qRoC:
Hello, I use Alacritty on MacOS with settings:

```yaml
window:
  startup_mode: SimpleFullscreen
font:
  size: 14
  offset:
    y: 1
```

I have free space at the bottom of the window:

<img width="42" alt="Screenshot 2022-10-23 at 20 51 33" src="https://user-images.githubusercontent.com/1789565/197407706-fcf63b74-49b4-4a4a-b3e8-037694145d07.png">

Is it possible to vertical align the content to the bottom or specify the padding only for the top side?

kchibisov:
Hm, not. The only thing you can use is `dynamic_padding` to make it evenly spread.

Issue #4297, opened 2020-10-08
Feature: More granular padding configurability


Stakker:
If you have Alacritty permanently living in one static location on your desktop without the title bar, then it would also be useful for us fussy desktop tweakers to be able to change the padding separately for each edge to get your layout just right.

I know it's not a super common use case, but I'd imagine it's probably also a fairly easy feature to implement :-)

chrisduerr:
This is a possible solution/workaround for https://github.com/alacritty/alacritty/issues/1671.

ike:
Would this be essentially solved by @sodiumjoe's solution in #1290?

chrisduerr:
Yes, though the comments in #1290 still apply. The benefit of this is questionable and a usecase has not sufficiently been elaborated.

Doing it because it's a "fairly easy feature to implement", is not a good reason to add new features. Features have maintenance cost and should always come with a significant improvement to Alacritty's users. Especially with a configuration file change all changes have to immediately fight the inconvenience of the grown config size as a hurdle.

ike:
Where do you land on the macos-specific-fix <-> padding-configurable spectrum?

chrisduerr:
Making padding configurable by itself is not a reason. And this is not a fix for the macos issue at all.

ike:
Gotcha -- has the macos transparent padding issue been addressed in a different way? I have read #1671 but am curious if there are other workarounds or preferred methods.

chrisduerr:
It has not.

Stakker:
Yeah fighting bloat & complexity is a good goal (and it aligns with Alacritty's "vision"), so I'll survive without the extra conf :-)

ike:
So perhaps this is a `wont-implement` feature?

jvantuyl:
One other comment on this one.  With Apple adding the "notch" into the top of MacBook screens, there's yet-another-reason-to-add-configurable padding.

While I'm all for solving concrete problems, we can't be psychic and we can't be prescient.  There are a lot of different reasons that we know and that we don't know that configurable padding would help.  Granular config of padding solves them all... categorically... forever.

chrisduerr:
This is not a fix for titlebar height or any other nonsense Apple might come up with in the future. It's at best a hacky workaround. As such the issues should be properly solved and I will happily accept PRs doing so.

> While I'm all for solving concrete problems, we can't be psychic and we can't be prescient.

[...]

PR #6442, opened 2022-10-24
Terminal content align

qRoC:
Now is possible to align content based on padding. This is a very important functionality for displays with rounded corners.

For example, new MacBooks have rounded top corners, which makes it impractical to use `window.startup_mode: SimpleFullscreen`. The current solution was to specify `padding.y` losing space at the bottom, now you can do this:

```yaml
window:
  valign: bottom
  padding:
    y: 10
```

Another problem is the inability to specify where to leave space for indentation requirements. ( #6440 ).

This PR adds 2 new properties:
 - `window.valign`: `top`/`middle`/`bottom`, where `middle` same as `dynamic_padding` for vertical align.
 - `window.halign`: `left`/`center`/`right`, where `center ` same as `dynamic_padding` for horizontal align.

`dynamic_padding` now is deprecated.

Examples:

```yaml
  valign: bottom
  halign: center
  padding:
    y: 10
    x: 20
```

Screenshot 2022-10-24 at 13 29 58

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Option for padding align (`window.valign`, `window.halign`)
+- Options for padding align (`window.valign`, `window.halign`)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.12.0-dev
 
+### Added
+
+- Option for padding align (`window.valign`, `window.halign`)
+
 ### Fixed
 
 - `--help` output for `--class` does not match man pages
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -50,8 +50,10 @@
   #  x: 0
   #  y: 0
 
-  # Spread additional padding evenly around the terminal content.
-  #dynamic_padding: false
+  # Vertical align for terminal content.
+  #valign: middle
+  # Horizontal align for terminal content.
+  #halign: center
 
   # Window decorations
   #
--- a/alacritty/src/config/window.rs
+++ b/alacritty/src/config/window.rs
@@ -138,10 +138,10 @@ impl WindowConfig {
     #[inline]
     pub fn align(&self) -> (VerticalAlign, HorizontalAlign) {
         if self.valign == Default::default() && self.halign == Default::default() && self.dynamic_padding {
-            return (VerticalAlign::Middle, HorizontalAlign::Center)
+            return (VerticalAlign::Middle, HorizontalAlign::Center);
         }
 
-        return (self.valign, self.halign)
+        (self.valign, self.halign)
     }
 
     #[inline]
--- a/alacritty/src/config/window.rs
+++ b/alacritty/src/config/window.rs
@@ -40,7 +40,13 @@ pub struct WindowConfig {
     decorations_theme_variant: Option<String>,
 
     /// Spread out additional padding evenly.
+    #[config(deprecated = "use window.valign and window.halign instead")]
     pub dynamic_padding: bool,
+
+    /// Vertical align for terminal content.
+    pub valign: VerticalAlign,
+    /// Horizontal align for terminal content.
+    pub halign: HorizontalAlign,
 
     /// Use dynamic title.
     pub dynamic_title: bool,
@@ -70,6 +76,8 @@ impl Default for WindowConfig {
             decorations_theme_variant: Default::default(),
             gtk_theme_variant: Default::default(),
             dynamic_padding: Default::default(),
+            valign: Default::default(),
+            halign: Default::default(),
             identity: Identity::default(),
             opacity: Default::default(),
             padding: Default::default(),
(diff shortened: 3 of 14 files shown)


kchibisov:
I guess what we want is https://github.com/alacritty/alacritty/issues/4297 with basically offsets from each side, so we have 4 options.

The current way looks a bit weird, especially given you have `padding` from all directions in code, but have weird exposure for that for the user?

But it could be indeed a bit weird how to handle dynamic padding if you can specify it from all the edges and where to add extra padding.

>This is a very important functionality for displays with rounded corners.

No, it's not, for rounded corners you have system to tell you the safe zones of the display, so something like `winit` could tell us about them and we won't ever draw there. It has nothing to do with the padding in-particular.

qRoC:
Padding for each side is more obvious setting, but complicated because we have a "remainder" where we can't put content. This will force the user to manually indent all the time. This option is really more difficult to understand, but is not inferior in flexibility and solves this problem.

I need to pin content to the bottom:

```yaml
window:
  valign: bottom
```

I need left padding:

```yaml
window:
  halign: right
  padding:
    x: 10
```

I need left and right padding:

```yaml
window:
  halign: center
  padding:
    x: 10
```

> No, it's not, for rounded corners you have system to tell you the safe zones of the display, so something like winit could tell us about them and we won't ever draw there. It has nothing to do with the padding in-particular.

This is just an example, I was trying to say that in some modes (`SimpleFullscreen`) this solves this problem.

kchibisov:
Yeah, I'm not sure about the remainder either, would suggest to wait for @chrisduerr here.

I do understand the need to add remainder to top and not to bottom. The way it's added to bottom and right is because it's simpler.

chrisduerr:
I do not understand why this would be necessary at all. If hardware is the reason you suggest this, it seems absolutely ridiculous to me.

Adding the extra space to the top/left makes very little sense to me.

kchibisov:
@chrisduerr we had similar issue wrt titlebar and such, so that's why the original issue exist. There could be a reason for ascetics, since some apps have status bars at the bottom and one may want to perfectly align it.

I think though that it's overly complicated and having regular padding to offset each side sounds better. Maybe just offset top even?

The hardware limitations is not an issue to solve here, since as it was stated the right solution is inside the winit for it. However the titlebar thing is likely to be solved with padding.

chrisduerr:
> @chrisduerr we had similar issue wrt titlebar and such, so that's why the original issue exist.

Aligning to bottom is **not** a fix to the titlebar issue. So this patch doesn't really qualify for that. Yes the titlebar issue exists, but it should be fixed properly not "accidentally" by adding a completely unrelated feature.

> some apps have status bars at the bottom and one may want to perfectly align it.

Some apps also have status bars at the top and bottom. The best way to align the top and bottom of the terminal for general usage is to use dynamic padding, that's why I added and use this option. Adding an option to align to the bottom doesn't solve anything, it just moves the issues from the bottom to the top.

> I think though that it's overly complicated and having regular padding to offset each side sounds better. Maybe just offset top even?

I'm okay-ish with setting individual padding for each side. But I've yet to find an actual usecase for that which isn't just a hacky workaround. If the goal is just to fix the top padding thing, then we should add the **exact** height of the titlebar to the top padding automatically, not have the user figure it out on their own.

qRoC:
And what will the individual padding give? Constantly changing the configuration on any changes? 

> The best way to align the top and bottom of the terminal for general usage is to use dynamic padding

Maybe for you, but most users only use the bottom status bar (tmux, vim, etc).

This PR makes it possible to bind both to the bottom and in the middle(dynamic padding) for you.

kchibisov:
Individual padding will give you a constant padding of the size of the titlebar from the the top? With your approach you'd have to adjust as well? Since you need a padding of `notch_size` from the top already it's not like you can purely rely on the top padding.

But as it was mentioned you're trying to solve problem with a hacks, since what you want is to draw content only into safe area? Which is completely different thing, since you need to obtain safe area from the system.

>And what will the individual padding give? Constantly changing the configuration on any changes?

It'll give you a constant offset from the top, but it's really not a thing you want and trying to solve the notch issue that way is just a not go.

To be fair, the titlebar issue could be solved as well with the help of winit, buy providing a safe area for that as well.

@chrisduerr abstracting away from the notch issue, since it's irrelevant here, what they want is to add extra padding not to the bottom, but to the top. Dynamic padding would make it even, but they in particular don't want even, but want their stuff be 0px from the bottom, which I sort of think that is reasonable, but the option is niche (Not sure I've seen it in any terminal).

kchibisov:
The issue about safe area is here https://github.com/alacritty/alacritty/issues/6444

chrisduerr:
> Dynamic padding would make it even, but they in particular don't want even, but want their stuff be 0px from the bottom

I don't see why you'd want "0px from the bottom" though for any system with reasonable hardware. The only cause for this 0 distance from bottom request is broken hardware.

Considering that text is usually top-left to bottom-right, having all the padding in the top is just going to make things even more annoying on most normal systems.

qRoC:
@kchibisov You can't set padding. It's not obvious - you must calculate for concrete display and settings (display height, font size, font offset)
For example your display height is 1000 and row height is 26.

case: `padding_top: 11`
result: `padding_top: 11, padding_bottom: 1`

case: `padding_top: 12`
result: `padding_top: 12`

case: `padding_top: 13`
result: `padding_top: 13, padding_bottom: 25`

case: `padding_top: 38`
result: `padding_top: 38`

So the concept of padding doesn't work. For this reason I added aligns, no more.

bric3:
On macOs there is `transparent` decorations, however this makes the text appear under the buttons. In order to have the transparent decorations but keep the buttons, one can use the padding, but then the vertical padding is also applied to the bottom (which is logical).

I believe the `valign` would have been solution for this esthetic issue. Or maybe there's something else to do or consider?

```
window:
    padding:
        x: 5
        y: 30

    dynamic_padding: true

    # Window decorations
    #
    # Values for `decorations`:
    #
    # Values for `decorations` (macOS only):
    #     - transparent: Title bar, transparent background and title bar buttons
    #     - buttonless: Title bar, transparent background and no title bar buttons
    decorations: transparent
```

qRoC:
> I believe the valign would have been solution for this esthetic issue

@bric3 Hello, yes. For example, my config(macOS):

```yaml
window:
  startup_mode: SimpleFullscreen
  decorations: none
  valign: bottom

font: # Hack Nerd Font Mono
  size: 14
  offset:
    y: 3
```

I don't have unused space at the bottom, which is aesthetically terrible.

PR #6453, opened 2022-10-27
Adding yaml schema (WORK IN PROGRESS)

imochoa:
**WORK IN PROGRESS**

Making it easier to modify the configuration YAML file using the schema!
(Supported by most YAML editors, I use Neovim & VSCode and they should work nicely)

# VSCode example
If you use VSCode, you can set the schema used by the [YAML language server](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) by creating a file `.vscode/settings.json` that tells the editor what schema to use:
```json
{
    "yaml.schemas": {
        "Path/URL to schema": "globPattern of files to apply it to",
        "https://raw.githubusercontent.com/imochoa/alacritty/adding-yaml-schema/extra/alacritty-config-schema.json": "/*"
    }
}
```
> Note that `/*` will match all files. You should set the path so that it only affects your alacritty config file

This is what it looks like in VSCode:
image

Here is an example of it complaining about the data type:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.12.0-dev
 
+### Fixed
+
+- `--help` output for `--class` does not match man pages
+- Cursor and underlines always being black on very old hardware
+- Crash when using very low negative `font.offset`
+- Startup failure on macOS with default config when system `/bin/sh` is `dash`
+- Artifacts in corners for maximized window with CSD on Wayland
+- Dotted underline not shown on macOS
+- Underline on macOS always being at the bottom of the cell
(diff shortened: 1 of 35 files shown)


chrisduerr (changes requested):
This is just completely unnecessary maintenance burden.

imochoa:
@chrisduerr Are you sure about that? 
To each their own I guess, but having a schema for the data structures you expect other people to use is just good practice, not a "maintenance burden" (especially in this case where these files are modified by hand)

The user experience is infinitely better than having to wonder why your 900 line YAML file is not working as you expect.
Or having to read the source code...

chrisduerr:
Yeah imagine having to figure out all these types yourself. Wish we'd provide a file documenting all the viable parameters for these options. That would be really nice.

imochoa:
Imagine being able to distinguish between documentation and schemas. That would be really nice

PR #6457, opened 2022-10-30
Update to notify v5 via notify_debouncer_mini

jamessan:
v5 notify moved the debounced API into its own crate,
notify_debouncer_mini.  The debounced API doesn't provide details on the
type of event that happened, just that a list of events or errors
happened.  Therefore, trigger reload on any event for a matching path.

Disable default features for notify_debouncer_mini since std::sync::mpsc
is being used, so there's no need to pull in crossbeam-channel.

### DIFF
--- a/alacritty/src/config/monitor.rs
+++ b/alacritty/src/config/monitor.rs
@@ -3,7 +3,8 @@ use std::sync::mpsc;
 use std::time::Duration;
 
 use log::{debug, error};
-use notify::{watcher, DebouncedEvent, RecursiveMode, Watcher};
+use notify_debouncer_mini::new_debouncer;
+use notify_debouncer_mini::notify::RecursiveMode;
 use winit::event_loop::EventLoopProxy;
 
 use alacritty_terminal::thread;
@@ -40,8 +41,8 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
 
     // The Duration argument is a debouncing period.
     let (tx, rx) = mpsc::channel();
-    let mut watcher = match watcher(tx, DEBOUNCE_DELAY) {
-        Ok(watcher) => watcher,
+    let mut debouncer = match new_debouncer(DEBOUNCE_DELAY, None, tx) {
+        Ok(debouncer) => debouncer,
         Err(err) => {
             error!("Unable to watch config file: {}", err);
             return;
@@ -61,6 +62,7 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
         parents.sort_unstable();
         parents.dedup();
 
+        let watcher = debouncer.watcher();
         // Watch all configuration file directories.
         for parent in &parents {
             if let Err(err) = watcher.watch(parent, RecursiveMode::NonRecursive) {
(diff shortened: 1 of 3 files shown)


alacritty/src/config/monitor.rs
```diff
@@ -69,26 +70,20 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
...
             };
 
-            match event {
-                DebouncedEvent::Rename(_, path)
-                | DebouncedEvent::Write(path)
-                | DebouncedEvent::Create(path)
-                | DebouncedEvent::Chmod(path)
-                    if paths.contains(&path) =>
-                {
+            if let Ok(events) = res {
```
chrisduerr, line 81: Why are you ignoring this error completely?

jamessan: Because that's what the code was doing before, too.

chrisduerr: The code before didn't have a result here.

jamessan: The error was encoded in the [DebouncedEvent](https://docs.rs/notify/4.0.17/notify/enum.DebouncedEvent.html#variant.Error).  With the new API, they made it [an actual Result](https://docs.rs/notify-debouncer-mini/0.2.1/notify_debouncer_mini/type.DebounceEventResult.html).

chrisduerr:
I feel like this patch is just a downgrade in functionality, so I'm not quite sure if it should be merged.

Do you know why the individual events aren't broadcasted anymore?

kchibisov:
@chrisduerr one thing to consider is that BSD will have normal wait instead of active polling each 50ms.

I'm not sure if we miss that much though?

chrisduerr:
Yeah that's a good point, I saw the new kqueue dependency. So I guess this closes that one open bug we had about it.

I'm not really opposed to this. I just find it weird they've changed the API this way and I'd like to understand why?

jamessan:
https://github.com/notify-rs/notify/blob/main/UPGRADING_V4_TO_V5.md#precise-events--debouncing gives their reasoning.

alacritty/src/config/monitor.rs
```diff
@@ -69,26 +71,20 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
         }
 
         loop {
-            let event = match rx.recv() {
-                Ok(event) => event,
+            let res = match rx.recv() {
+                Ok(res) => res,
```
chrisduerr, line 75: Let's just also log the error for `res` here and be done with it. Wasted too much time already on this dependency bump.

So both `Err(err)` and `Err(Err(err))` should print debug logs, just to make sure we're not ignoring serious issues.

Probably easiest to use one error message for both, no need to be that specific in a debug message unseen by most anyway.

chrisduerr (changes requested):

chrisduerr:
This will now print a warning when the configuration file is deleted while there are still Alacritty instances open which used that configuration file. But I think that's a sensible thing to warn about anyway, so should be fine.

jamessan:
Thanks for the review and merge. I think this was the last blocker for getting the Debian package uploaded.

chrisduerr:
Oh you should have mentioned that from the beginning. That's certainly always a good argument to bump dependencies.

PR #6481, opened 2022-11-10
Don't change cursor visibility when the window loses focus.

### LINKED ISSUE #6452: Mouse doesn't show up with mouse hide_on_typing set to true
### System

OS: macOS 13.0 (22A380)
Version: lacritty 0.11.0 (8dbaa0b)

### Logs

First enable the `hide_on_typing` in the config.
```
mouse:
  hide_when_typing: true
```

Steps to reproduce:

1. Launch Alacritty.
2. Type something in the terminal and don't move your mouse, the cursor is hidden.
3. Switch to a different app using `command` + `tab`.
4. Switch back to Alacritty using `command` + `tab` and don't type anything yet.
5. Swipe fingers on the touchpad or move the mouse, the cursor doesn't show up.

Expected behavior:
- In step 5, the cursor should show up.

chrisnc:
Fixes #6452.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Underline on macOS always being at the bottom of the cell
 - Crash with `OT-SVG` fonts on Linux/BSD
 - Crash during text compose on old GNOME under Wayland
+- Mouse cursor staying hidden after window regains focus on macOS Ventura
 
 ## 0.11.0
 
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1213,8 +1213,6 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
 
                         if is_focused {
                             self.ctx.window().set_urgent(false);
-                        } else {
-                            self.ctx.window().set_mouse_visible(true);
                         }
 
                         self.ctx.update_cursor_blinking();


chrisnc:
In my testing, the cursor still became visible after switching away and back to the original window without any mouse input on both Windows and X11, so I believe the call to make the cursor visible on losing focus was never necessary on those platforms. On macOS, the cursor will now hide again when switching back to a window in which it was originally hidden due to typing, but now mouse movement will reveal it again, which isn't happening on Ventura currently.

kchibisov:
I know that it could cause issues for some Wayland users, but they have a miss configured setup anyway...

chrisnc:
> I know that it could cause issues for some Wayland users, but they have a miss configured setup anyway...

Interesting. I'll have to test this one next... I also still need to test on Monterey.

kchibisov:
I mean on Wayland it'll break only of the application doesn't draw the cursor at all, so it'll have a cursor of the previous application which is hidden. It just on Wayland the cursor is drawn by the client and not the system compositor, so that's why it's different. I don't care about such things at all, since clearly the applications must draw cursor if they care about it. Though, I'd suggest test on GNOME since it could not draw cursor when you move cursor to background or something like that.

chrisnc:
Thanks for the context. I will give that a try and leave a comment with the results.

chrisduerr (changes requested):
Could you add a changelog entry, considering this affects 0.11.0?

SheldonZhong:
It seems winit doesn't track the visibility state on Windows and Wayland. Would there potential performance / power efficiency issue? Without the visibility state, every mouse movement triggers a call to winit's `set_cursor_visible` on those platforms.

chrisnc:
> It seems winit doesn't track the visibility state on Windows and Wayland. Would there potential performance / power efficiency issue? Without the visibility state, every mouse movement triggers a call to winit `set_cursor_visible`'s underlying API on those platforms.

Oof, you're right... I will revert those changes.

chrisnc:
Added changelog, removed all other changes besides dropping the call on loss of focus. I confirmed this behaves the same was on Monterey as on Ventura. (Each window remembers whether the cursor was hidden, and the cursor hides and unhides when it should.) I attempted to test on Wayland with a fresh install of Debian configured to use GNOME, but for some reason the session is still x11; possibly there's something about my system that Wayland doesn't support. Any suggestions appreciated, or if someone else could volunteer to test this on Wayland, that would be great.

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Underline on macOS always being at the bottom of the cell
 - Crash with `OT-SVG` fonts on Linux/BSD
 - Crash during text compose on old GNOME under Wayland
+- On macOS Ventura, mouse movement would not unhide the cursor if the window
+    lost and then regained focus while the cursor was hidden
```
chrisduerr, line 22:
```suggestion
- Mouse cursor staying hidden after focus change on macOS Ventura
```

Let's keep these changes more concise.

chrisduerr (changes requested):

chrisduerr (approved):

PR #6530, opened 2022-11-26
Handle GPU resets

kchibisov:
This should keep alacritty working after the GPU reset getting detected.

### DIFF
--- a/alacritty/build.rs
+++ b/alacritty/build.rs
@@ -17,6 +17,7 @@ fn main() {
 
     Registry::new(Api::Gl, (3, 3), Profile::Core, Fallbacks::All, [
         "GL_ARB_blend_func_extended",
+        "GL_KHR_robustness",
         "GL_KHR_debug",
     ])
     .write_bindings(GlobalGenerator, &mut file)
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -562,8 +562,8 @@ impl Display {
         let is_current = self.context.get().is_current();
 
         // Attempt to make the context current if it's not.
-        let mut was_context_reset = if is_current {
-            false
+        let context_loss = if is_current {
+            self.renderer.was_context_reset()
         } else {
             match self.context.make_current(&self.surface) {
                 Err(err) if err.error_kind() == ErrorKind::ContextLost => {
@@ -574,9 +574,7 @@ impl Display {
             }
         };
 
-        was_context_reset |= self.renderer.was_context_reset();
-
-        if !was_context_reset {
+        if !context_loss {
             return;
         }
 
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -8,7 +8,10 @@ use std::num::NonZeroU32;
 use std::ops::{Deref, DerefMut};
 use std::time::{Duration, Instant};
 
+use glutin::config::GetGlConfig;
 use glutin::context::{NotCurrentContext, PossiblyCurrentContext};
+use glutin::display::GetGlDisplay;
+use glutin::error::ErrorKind;
 use glutin::prelude::*;
 use glutin::surface::{Surface, SwapInterval, WindowSurface};
 
@@ -33,6 +36,7 @@ use alacritty_terminal::term::{
 };
 use alacritty_terminal::vte::ansi::{CursorShape, NamedColor};
 
+use crate::config::debug::RendererPreference;
 use crate::config::font::Font;
 use crate::config::window::Dimensions;
 #[cfg(not(windows))]
@@ -385,6 +389,7 @@ pub struct Display {
     hint_mouse_point: Option<Point>,
 
     renderer: ManuallyDrop<Renderer>,
+    renderer_preference: Option<RendererPreference>,
 
     surface: ManuallyDrop<Surface<WindowSurface>>,
 
@@ -513,6 +518,7 @@ impl Display {
             context: ManuallyDrop::new(Replaceable::new(context)),
             visual_bell: VisualBell::from(&config.bell),
             renderer: ManuallyDrop::new(renderer),
+            renderer_preference: config.debug.renderer,
             surface: ManuallyDrop::new(surface),
             colors: List::from(&config.colors),
             frame_timer: FrameTimer::new(),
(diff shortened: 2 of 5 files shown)


kchibisov:
This branch is blocked, unless I can find a system to reliably test it. Eventually sway will get support for reset recovering, but for now when GPU reset happens system compositor is the first thing that dies so the fact that alacritty will die doesn't matter.

There also will be a driver protocol to handle such a thing.

matu3ba:
> This branch is blocked, unless I can find a system to reliably test it. Eventually sway will get support for reset recovering, but for now when GPU reset happens system compositor is the first thing that dies so the fact that alacritty will die doesn't matter.

Arcan claims to handle GPU resets + more gracefully. [technical details](https://arcan-fe.com/2017/12/24/crash-resilient-wayland-compositing/). [repo](https://github.com/letoram/arcan)

kchibisov:
I mean sway will have support for stuff like that eventually, it'll also require a special protocol though.

YaLTeR:
Got it to recover successfully on an AMD dGPU in a laptop, sway 1.10, kernel 6.11.11.
```
[75.892549293s] [INFO ] [alacritty] GPU reset (innocent)
[75.895038005s] [ERROR] [alacritty] [gl_render] GL_CONTEXT_LOST in context lost
[75.895105231s] [ERROR] [alacritty] [gl_render] GL_CONTEXT_LOST in context lost
[75.895132933s] [ERROR] [alacritty] [gl_render] GL_CONTEXT_LOST in context lost
[75.895157971s] [ERROR] [alacritty] [gl_render] GL_CONTEXT_LOST in context lost
[75.895182176s] [ERROR] [alacritty] [gl_render] GL_CONTEXT_LOST in context lost
[75.895253811s] [ERROR] [alacritty] [gl_render] GL_CONTEXT_LOST in context lost
[75.895281102s] [ERROR] [alacritty] [gl_render] GL_CONTEXT_LOST in context lost
[75.895304166s] [ERROR] [alacritty] [gl_render] GL_CONTEXT_LOST in context lost
[75.895326999s] [ERROR] [alacritty] [gl_render] GL_CONTEXT_LOST in context lost
[75.895357155s] [ERROR] [alacritty] [gl_render] GL_CONTEXT_LOST in context lost
[75.895379908s] [ERROR] [alacritty] [gl_render] GL_CONTEXT_LOST in context lost
[75.903346426s] [INFO ] [alacritty] Running on AMD Radeon RX 6700M (radeonsi, navi22, LLVM 19.1.5, DRM 3.59, 6.11.11-300.fc41.x86_64)
[75.903459529s] [INFO ] [alacritty] OpenGL version 4.6 (Core Profile) Mesa 24.3.2, shader_version 4.60
[75.903480799s] [INFO ] [alacritty] GPU reset notifications are enabled
[75.903500155s] [INFO ] [alacritty] Using OpenGL 3.3 renderer
[75.908700058s] [DEBUG] [alacritty] Enabled debug logging for OpenGL
[75.919608000s] [DEBUG] [alacritty] Recovered window WindowId(94729299219344) from gpu reset
[75.935177160s] [DEBUG] [alacritty_terminal] New num_cols is 159 and num_lines is 83
[75.935811363s] [INFO ] [alacritty] Padding: 2 x 2
[75.935873921s] [INFO ] [alacritty] Width: 1276, Height: 1548
```

kchibisov:
>[75.895357155s] [ERROR] [alacritty] [gl_render] GL_CONTEXT_LOST in context lost

@chrisduerr btw, should we log those as error if we can recover? it does let the user know that gpu crashed, but it just opens message bar...

chrisduerr:
> should we log those as error if we can recover? it does let the user know that gpu crashed, but it just opens message bar...

What does the user gain by knowing? It's not really Alacritty's responsibility to inform the user about this.

alacritty/build.rs
```diff
@@ -17,6 +17,7 @@ fn main() {
 
     Registry::new(Api::Gl, (3, 3), Profile::Core, Fallbacks::All, [
         "GL_ARB_blend_func_extended",
+        "GL_KHR_robustness",
```
chrisduerr, line 20: How widely is this supported?

alacritty/src/display/mod.rs
```diff
@@ -1413,7 +1469,7 @@ impl Drop for Display {
     fn drop(&mut self) {
         // Switch OpenGL context before dropping, otherwise objects (like programs) from other
         // contexts might be deleted when dropping renderer.
-        self.make_current();
+        self.make_current(false);
```
chrisduerr, line 1472: Are we just not recovering here because it's not necessary? Seems kinda odd to have a parameter solely for the drop impl.

alacritty/src/renderer/mod.rs
```diff
@@ -281,6 +284,45 @@ impl Renderer {
...
+            };
+
+            info!("GPU reset ({})", reason);
+
+            true
+        }
+    }
+
+    /// Log reset notifications.
+    fn log_reset_notification_status() -> bool {
```
chrisduerr, line 307: Why does this return a boolean?

kchibisov:
>What does the user gain by knowing? It's not really Alacritty's responsibility to inform the user about this.

we just has general handler that does error logging on opengl errors, though, I guess for context loss we can suppress it.

alacritty/src/display/mod.rs
```diff
@@ -552,9 +558,59 @@ impl Display {
...
+            match self.context.make_current(&self.surface) {
+                Err(err) if err.error_kind() == ErrorKind::ContextLost => {
+                    info!("Context lost for window {:?}", self.window.id());
+                    true
+                },
+                _ => false,
+            }
+        };
+
+        was_context_reset |= self.renderer.was_context_reset();
```
chrisduerr, line 579: Is there a realistic benefit to this not just being inside the `is_current` branch? Sure it might help if the context resets right after the `make_current` call, but that could also happen right after this call, right? So in the end this should only really be necessary if `is_current` is true?

alacritty/src/display/mod.rs
```diff
@@ -552,9 +558,59 @@ impl Display {
...
+            false
+        } else {
+            match self.context.make_current(&self.surface) {
+                Err(err) if err.error_kind() == ErrorKind::ContextLost => {
+                    info!("Context lost for window {:?}", self.window.id());
+                    true
+                },
+                _ => false,
+            }
+        };
```
chrisduerr, line 577:
```suggestion
        // Attempt to make the context current, if it is not.
        let mut was_context_reset = if is_current {
            false
        } else {
            match self.context.make_current(&self.surface) {
                Err(err) if err.error_kind() == ErrorKind::ContextLost => {
                    info!("Context lost for window {:?}", self.window.id());
                    true
                },
                _ => false,
            }
        };
```

I think a comment here is nice to explain that our goal here is not just checking if the context was reset, but actually performing the make_current call.

alacritty/src/display/mod.rs
```diff
@@ -552,9 +558,59 @@ impl Display {
...
+
+        was_context_reset |= self.renderer.was_context_reset();
+
+        if was_context_reset && recover {
+            let gl_display = self.context.display();
+            let gl_config = self.context.config();
+            let raw_window_handle = Some(self.window.raw_window_handle());
+            let context =
+                renderer::platform::create_gl_context(&gl_display, &gl_config, raw_window_handle)
+                    .expect("failed to recreate context.");
```
chrisduerr, line 587:
```suggestion
                platform::create_gl_context(&gl_display, &gl_config, raw_window_handle)
                    .expect("failed to recreate context.");
```

alacritty/src/display/mod.rs
```diff
@@ -552,9 +558,59 @@ impl Display {
...
+                    true
+                },
+                _ => false,
+            }
+        };
+
+        was_context_reset |= self.renderer.was_context_reset();
+
+        if was_context_reset && recover {
+            let gl_display = self.context.display();
```
chrisduerr, line 582: This is a pretty big chunk of code inside this conditional. Could probably move it to a function or early-return.

chrisduerr (changes requested):

alacritty/src/display/mod.rs
kchibisov: yeah, we don't recover since it doesn't matter, otherwise we create render and drop it.

chrisduerr: Would *probably* be fine tbh. It's **extremely** unlikely to ever happen anyway.

Like who manages to shut down Alacritty right between rendering at a billion frames a second and a GPU reset? Even if we crash at that point most people would never notice.

alacritty/src/display/mod.rs
kchibisov: I feel like it could lie/desync/not propagate, it's opengl at the end of the day.

chrisduerr: But isn't the same true for this function too? How likely are the two to be different?

alacritty/src/renderer/mod.rs
```diff
@@ -281,6 +285,49 @@ impl Renderer {
         }
     }
 
+    /// Get the context reset status.
+    pub fn was_context_reset(&self) -> bool {
+        // If robustness is not supproted, don't use its functions.
```
chrisduerr, line 290:
```suggestion
        // If robustness is not supported, don't use its functions.
```

chrisduerr (changes requested):

alacritty/src/display/mod.rs
kchibisov: The context is done by WSI, but the OpenGL is by driver itself, though, eventually the `is_current` path will be triggered.

chrisduerr (approved):

PR #6535, opened 2022-11-27
Add ALACRITTY_RENDERER environment variable

kchibisov:
This should help trouble shooting the renderer being created and different renderer options to determine when something like dual-source rendering isn't working.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -888,6 +888,12 @@
   #   - Trace
   #log_level: Warn
 
+  # Renderer to force
+  #   - glsl3
+  #   - gles2
+  #   - gles2_pure
+  #renderer: None
+
   # Print all received window events.
   #print_events: false
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -888,7 +888,7 @@
   #   - Trace
   #log_level: Warn
 
-  # Renderer to force
+  # Renderer override.
   #   - glsl3
   #   - gles2
   #   - gles2_pure
--- a/alacritty/src/config/debug.rs
+++ b/alacritty/src/config/debug.rs
@@ -49,7 +49,7 @@ pub enum RendererPreference {
     #[serde(rename = "glsl3")]
     Glsl3,
 
-    /// GLES 2 renderer, but can use extensions like dual source blending.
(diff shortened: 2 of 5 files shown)


chrisduerr (commented):
Why did you decide to go with an environment variable rather than a configuration option?

As far as I remember, this would be the first thing we expose as an environment variable in Alacritty directly.

kchibisov:
Hm, I'm just used that most libs control the renderer to use that way. Could add a debug option for that. It's not like we directly expose it. It'll be used in a context with 'could you test renderer XXX'.

Do you want a `debug` option for that?

kchibisov:
I've updated to use the config option.

alacritty.yml
```diff
@@ -888,6 +888,12 @@
   #   - Trace
   #log_level: Warn
 
+  # Renderer to force
```
chrisduerr, line 891:
```suggestion
  # Renderer override.
```

alacritty/src/config/debug.rs
```diff
@@ -18,6 +20,9 @@ pub struct Debug {
     /// Highlight damage information produced by alacritty.
     pub highlight_damage: bool,
 
+    /// The renderer alacritty should be using
```
chrisduerr, line 23:
```suggestion
    /// The renderer alacritty should be using.
```

alacritty/src/config/debug.rs
```diff
@@ -32,6 +37,23 @@ impl Default for Debug {
...
+/// The renderer configuration options.
+#[derive(Deserialize, Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum RendererPreference {
+    /// OpenGL 3.3 renderer.
+    #[serde(rename = "glsl3")]
+    Glsl3,
+
+    /// GLES 2 renderer, but can use extensions like dual source blending.
+    #[serde(rename = "gles2")]
+    Gles2,
```
chrisduerr, line 54:
```suggestion
    /// GLES 2 renderer, with optional extensions like dual source blending.
    #[serde(rename = "gles2")]
    Gles2,
```

chrisduerr (approved):

chrisduerr (approved):

PR #6555, opened 2022-12-07
Fix damage computation on boundaries

kchibisov:
Given that the Rect started to use signed integers saturating_sub became irrelevant and no clamp to zero were performed. This commit uses max instead to fix it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -676,9 +676,9 @@ dependencies = [
 
 [[package]]
 name = "glutin"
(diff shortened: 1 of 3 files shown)


chrisduerr (changes requested):
This does affect master, right? So we probably need a changelog entry.

kchibisov:
No it doesn't, it's due to glutin update.

chrisduerr (approved):

PR #6559, opened 2022-12-10
Use -T as the short form for --title

jamessan:
Debian-based distributions provide a standard interface to launch a terminal via the x-terminal-emulator name.  In order for a terminal emualtor to satisfy that interface, it must

* Be VT100 compatiable
* Support the "-e <command> <args>" CLI option
* Support the "-T <title>" CLI option

Adjust the short form of --title accordingly, providing -t as an alias to avoid breaking any existing usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.12.0-dev
 
+### Added
+
+- Uppercase `-T` short form for `--title`
+
 ### Fixed
 
 - `--help` output for `--class` does not match man pages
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.12.0-dev
 
+### Changed
+
+- `-t` to `-T` for short form of `--title`
```
chrisduerr, line 12: Let's add this to the `Added` section and then just make it " - Uppercase `-T` short form for `--title`". Not sure if we'll ever really deprecate or remove `-t` now that we've had it for so long. So probably best to just accept both going forward.

chrisduerr (changes requested):
I've confirmed that this is the shortcut for at least XTerm, URxvt and Kitty. So makes some sense for consistency.

chrisduerr (approved):

PR #6577, opened 2022-12-20
Hack in tpp-bypass compatibility

DelusionalLogic:
The purpose of this PR is to explore the possibility of circumventing ConPTY for wsl on windows. This is my first real rust code, and it's been years since I touched the windows api. It's a best effort and any comments ar appreciated.

With that out of the way. I've been having some issues with the vim scrolling in wsl and being unable to use color 0 (black). It turns out that both issues are caused by ConPTY parsing the VTE codes from linux, breaking them in the process. It turns out that alacritty isn't the first project to have that issue. Mintty for example solved that in their [wsltty](https://github.com/mintty/wsltty) project by using a project called [wslbridge2](https://github.com/Biswa96/wslbridge2) that, using some undocumented COM objects and a process on either side of the wsl vm, redirects the terminal through some AF_VSOCK sockets, circumventing ConPTY. It works a treat, but it seems a little complicated and brittle to me. Unfortunately wslbridge is tied to cygwin, so it's not easily usable from rust and therefore not directly applicable to alacritty.

Another interesting project that has a solution to the ConPTY problem is [terminalpp](https://github.com/terminalpp/terminalpp). I've never heard of the terminal, but they have a subproject called [tpp-bypass](https://github.com/terminalpp/terminalpp/tree/master/tpp-bypass) which smuggles the raw terminal connection inside the stdin and stdout stream of wsl, no ConPTY involved at all. This project also avoids any external channel for resizes by encoding that control information inside the input stream of the linux pty. This design means it only needs a single executable on the linux side to establish and forward the pty.

I've tried to implement a backend for alacritty that can talk to tpp-bypass. That's the code you see in this PR. The pty output stream is passed directly to alacritty whereas the input stream is redirected through a thread that escapes the characters that need escaping. There may be some race condition between that thread and resize, but that could be solved with a mutex.

To test this out you'll need to compile tpp-bypass for your wsl distro. Be aware that the directions in the tpp-bypass repo do not work. You have to run cmake in the root directory, and then copy the stamp.h file into the tpp-bypass subdirectory.

With tpp-bypass installed into your `/home/<username>/.local/bin` folder you can run this backend with the commandline `alacritty.exe -e "wsl -e /home/<username>/.local/bin/tpp-bypass TERM=alacritty"` to connect directly to your wsl terminal without ConPTY.

To reiterate. This isn't ready for merging, but I'm looking for comments on style and direction. I think this could resolve a lot of VTE related issues on windows, but I will admit it's a bit of a hack.

### DIFF
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -1,18 +1,17 @@
-use std::io::Error;
-use std::os::windows::io::IntoRawHandle;
+use std::io::{Error, Read, Write};
+use std::os::windows::io::{AsRawHandle, FromRawHandle, IntoRawHandle};
 use std::{mem, ptr};
 
 use mio_anonymous_pipes::{EventedAnonRead, EventedAnonWrite};
 
 use windows_sys::core::PWSTR;
-use windows_sys::Win32::Foundation::{HANDLE, S_OK};
-use windows_sys::Win32::System::Console::{
-    ClosePseudoConsole, CreatePseudoConsole, ResizePseudoConsole, COORD, HPCON,
+use windows_sys::Win32::Foundation::{
+    DuplicateHandle, SetHandleInformation, DUPLICATE_SAME_ACCESS, HANDLE, HANDLE_FLAG_INHERIT,
 };
 use windows_sys::Win32::System::Threading::{
-    CreateProcessW, InitializeProcThreadAttributeList, UpdateProcThreadAttribute,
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
I have zero interest in merging some strange hacky API bypass for Window's ConPTY.

The ConPTY is Windows' official API targeted at exactly what Alacritty is using it for. If Microsoft can't figure that out, it should be fixed upstream in Windows not with some shitty hacks in Alacritty.

PR #6585, opened 2022-12-24
User timer based rendering instead of vsync

### LINKED ISSUE #824: Redraw randomly hangs until a keypress
Using Alacritty (commit 6916537858d12e02478d618339d1b9f0e73a677a) on Ubuntu 16.04 with X11.

It is better described with the following screencast. Here I'm repeatedly opening and closing `tig`, so the output should be the same in each iteration. The screen often stays in a partially redrawed state (marked by mouse cursor movement). Then it finishes redraw correctly when I'm pressing a key.

screencast

kchibisov:
Fixes #824.

--
For me it was keeping +- target frame rate with multiple instances. Would suggest to test with multiple monitors **at different refresh rate**, since I can't test it myself...

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -1562,58 +1562,47 @@ impl<T> DerefMut for Replaceable<T> {
 
 /// The frame timer state.
 pub struct FrameTimer {
-    /// The optimal delay for the renderer.
-    swap_timeout: Duration,
+    /// Base timestamp used to compute sync points.
+    base: Instant,
 
-    /// The previous error that we've used to offset the `swap_delay`.
-    previous_error: i64,
+    /// The last timestamp we synced to.
+    last_synced_timestamp: Instant,
 
-    /// The timestamp of the previous frame.
-    frame_timestamp: Instant,
-
-    /// The refresh rate we've used to compute the `swap_delay`.
+    /// The refresh rate we've used to compute sync timestamps.
     refresh_interval: Duration,
 }
 
 impl FrameTimer {
     pub fn new() -> Self {
-        Self {
-            swap_timeout: Duration::ZERO,
-            previous_error: 0,
-            frame_timestamp: Instant::now(),
-            refresh_interval: Duration::ZERO,
-        }
+        let now = Instant::now();
+        Self { base: now, last_synced_timestamp: now, refresh_interval: Duration::ZERO }
     }
 
     /// Compute the delay that we should use to achieve the target frame
     /// rate.
     pub fn compute_timeout(&mut self, refresh_interval: Duration) -> Duration {
+        let now = Instant::now();
+
         // Handle refresh rate change.
         if self.refresh_interval != refresh_interval {
+            self.base = now;
+            self.last_synced_timestamp = now;
             self.refresh_interval = refresh_interval;
-            self.frame_timestamp = Instant::now();
-            self.swap_timeout = refresh_interval;
             return refresh_interval;
         }
 
-        // Compute the error.
-        let error = self.refresh_interval.as_micros() as i64
-            - self.frame_timestamp.elapsed().as_micros() as i64;
+        let next_frame = self.last_synced_timestamp + self.refresh_interval;
 
-        // Don't update error and swap timeout if we've waited for too long.
-        if error < self.refresh_interval.as_micros() as i64 {
-            // Average previously used error with the current one to reduce momentum from the
-            // error sign changes.
-            self.previous_error = (self.previous_error + error) / 2;
-
-            let new_swap_timeout = self.swap_timeout.as_micros() as i64 + self.previous_error;
-            let new_swap_timeout =
-                Duration::from_micros(u64::try_from(new_swap_timeout).unwrap_or_default());
-            self.swap_timeout = new_swap_timeout;
+        if next_frame < now {
+            let elapsed_micros = (now - self.base).as_micros() as u64;
+            let refresh_micros = self.refresh_interval.as_micros() as u64;
+            self.last_synced_timestamp =
+                now - Duration::from_micros(elapsed_micros % refresh_micros);
+            Duration::ZERO
+        } else {
+            self.last_synced_timestamp = next_frame;
+            next_frame - now
         }
-
-        self.frame_timestamp = Instant::now();
-        self.swap_timeout
     }
 }
 
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -369,7 +369,7 @@ pub struct Display {
     pub ime: Ime,
 
     /// The state of the timer for frame scheduling.
-    pub frame_timer_info: FrameTimerInfo,
+    pub frame_timer: FrameTimer,
 
     // Mouse point position when highlighting hints.
     hint_mouse_point: Option<Point>,
@@ -510,7 +510,7 @@ impl Display {
             vi_highlighted_hint: None,
             is_wayland,
             cursor_hidden: false,
-            frame_timer_info: FrameTimerInfo::new(),
+            frame_timer: FrameTimer::new(),
             visual_bell: VisualBell::from(&config.bell),
             colors: List::from(&config.colors),
             pending_update: Default::default(),
@@ -1412,7 +1412,7 @@ impl Display {
         let monitor_vblank_interval =
             Duration::from_micros((1000. * monitor_vblank_interval) as u64);
 
-        let swap_timeout = self.frame_timer_info.compute_timeout(monitor_vblank_interval);
+        let swap_timeout = self.frame_timer.compute_timeout(monitor_vblank_interval);
 
         let window_id = self.window.id();
         let timer_id = TimerId::new(Topic::Frame, window_id);
@@ -1561,7 +1561,7 @@ impl<T> DerefMut for Replaceable<T> {
 }
 
 /// The frame timer state.
-pub struct FrameTimerInfo {
+pub struct FrameTimer {
     /// The optimal delay for the renderer.
     swap_timeout: Duration,
 
@@ -1575,7 +1575,7 @@ pub struct FrameTimerInfo {
     refresh_interval: Duration,
 }
 
-impl FrameTimerInfo {
+impl FrameTimer {
     pub fn new() -> Self {
         Self {
             swap_timeout: Duration::ZERO,
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -510,7 +510,7 @@ impl Display {
             vi_highlighted_hint: None,
             is_wayland,
             cursor_hidden: false,
-            frame_timer_info: Default::default(),
+            frame_timer_info: FrameTimerInfo::new(),
             visual_bell: VisualBell::from(&config.bell),
             colors: List::from(&config.colors),
             pending_update: Default::default(),
@@ -1561,51 +1561,59 @@ impl<T> DerefMut for Replaceable<T> {
 }
 
 /// The frame timer state.
-#[derive(Default)]
 pub struct FrameTimerInfo {
     /// The optimal delay for the renderer.
-    swap_delay: Option<Duration>,
+    swap_timeout: Duration,
+
+    /// The previous error that we've used to offset the `swap_delay`.
+    previous_error: i64,
 
     /// The timestamp of the previous frame.
-    frame_timestamp: Option<Instant>,
+    frame_timestamp: Instant,
 
     /// The refresh rate we've used to compute the `swap_delay`.
     refresh_interval: Duration,
 }
 
 impl FrameTimerInfo {
+    pub fn new() -> Self {
+        Self {
+            swap_timeout: Duration::ZERO,
+            previous_error: 0,
+            frame_timestamp: Instant::now(),
+            refresh_interval: Duration::ZERO,
+        }
+    }
+
     /// Compute the delay that we should use to achieve the target frame
     /// rate.
     pub fn compute_timeout(&mut self, refresh_interval: Duration) -> Duration {
-        let swap_timeout =
-            match self.swap_delay.filter(|_| self.refresh_interval == refresh_interval) {
-                Some(swap_delay) => swap_delay,
-                None => {
-                    // If the refresh rate has changed, reset the timings.
-                    self.refresh_interval = refresh_interval;
-                    self.frame_timestamp = Some(Instant::now());
-                    self.swap_delay = Some(refresh_interval);
-                    return refresh_interval;
-                },
-            };
+        // Handle refresh rate change.
+        if self.refresh_interval != refresh_interval {
+            self.refresh_interval = refresh_interval;
+            self.frame_timestamp = Instant::now();
+            self.swap_timeout = refresh_interval;
+            return refresh_interval;
+        }
 
-        let new_swap_timeout = match self.frame_timestamp {
-            Some(last_frame_timestamp) => {
-                // Compute the error.
-                let error = self.refresh_interval.as_micros() as i64
-                    - last_frame_timestamp.elapsed().as_micros() as i64;
-                let new_swap_delay = swap_timeout.as_micros() as i64 + error;
-                Duration::from_micros(u64::try_from(new_swap_delay).unwrap_or_default())
-            },
-            // If we don't have any information about the last frame, just return
-            // previous target.
-            None => swap_timeout,
-        };
+        // Compute the error.
+        let error = self.refresh_interval.as_micros() as i64
+            - self.frame_timestamp.elapsed().as_micros() as i64;
 
-        self.swap_delay = Some(new_swap_timeout);
-        self.frame_timestamp = Some(Instant::now());
+        // Don't update error and swap timeout if we've waited for too long.
+        if error < self.refresh_interval.as_micros() as i64 {
+            // Average previously used error with the current one to reduce momentum from the
+            // error sign changes.
+            self.previous_error = (self.previous_error + error) / 2;
 
-        new_swap_timeout
+            let new_swap_timeout = self.swap_timeout.as_micros() as i64 + self.previous_error;
+            let new_swap_timeout =
+                Duration::from_micros(u64::try_from(new_swap_timeout).unwrap_or_default());
+            self.swap_timeout = new_swap_timeout;
+        }
+
+        self.frame_timestamp = Instant::now();
+        self.swap_timeout
     }
 }
 
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -368,7 +368,7 @@ pub struct Display {
     /// The ime on the given display.
     pub ime: Ime,
 
-    /// The state of the timer based for frame scheduling.
+    /// The state of the timer for frame scheduling.
     pub frame_timer_info: FrameTimerInfo,
 
     // Mouse point position when highlighting hints.
@@ -492,7 +492,9 @@ impl Display {
         };
 
         // Disable vsync.
-        let _ = surface.set_swap_interval(&context, SwapInterval::DontWait);
+        if let Err(err) = surface.set_swap_interval(&context, SwapInterval::DontWait) {
+            info!("Failed to disable vsync: {}", err);
+        }
 
         Ok(Self {
             window,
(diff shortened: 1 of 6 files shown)


alacritty/src/display/mod.rs
```diff
@@ -487,14 +492,8 @@ impl Display {
...
-        // We use vsync everywhere except wayland.
-        if !is_wayland {
-            if let Err(err) =
-                surface.set_swap_interval(&context, SwapInterval::Wait(NonZeroU32::new(1).unwrap()))
-            {
-                warn!("Error setting vsync: {:?}", err);
-            }
-        }
+        // Disable vsync.
+        let _ = surface.set_swap_interval(&context, SwapInterval::DontWait);
```
kchibisov, line 496: I'm not sure how to handle failure of this. It could fail just because there's no extension. Maybe we should distinguish the failure from no extension and from the actual changing?

alacritty/src/display/mod.rs
```diff
@@ -367,6 +368,9 @@ pub struct Display {
     /// The ime on the given display.
     pub ime: Ime,
 
+    /// The state of the timer based for frame scheduling.
+    pub frame_timer_info: FrameTimerInfo,
```
chrisduerr, line 372:
```suggestion
    /// The state of the timer for frame scheduling.
    pub frame_timer_info: FrameTimerInfo,
```

alacritty/src/display/mod.rs
chrisduerr: I'm fine with ignoring it. We could log it maybe?

alacritty/src/display/mod.rs
```diff
@@ -969,7 +969,9 @@ impl Display {
         // Frame event should be requested before swaping buffers, since it requires surface
         // `commit`, which is done by swap buffers under the hood.
         #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
```
chrisduerr, line 971: I think this `cfg` can be removed? Sure it's a runtime check more if Wayland is disabled but it should basically be free and the function exists everywhere now.

Otherwise I'd suggest adding cfgs to all the `!self.is_wayland` places too, we should just be consistent.

alacritty/src/display/mod.rs
```diff
@@ -969,7 +969,9 @@ impl Display {
         // Frame event should be requested before swaping buffers, since it requires surface
         // `commit`, which is done by swap buffers under the hood.
```
chrisduerr, line 970:
```suggestion
        // Frame event should be requested before swapping buffers on Wayland, since it requires surface
        // `commit`, which is done by swap buffers under the hood.
```

alacritty/src/display/mod.rs
```diff
@@ -1371,23 +1377,38 @@ impl Display {
...
-        should_draw.store(false, Ordering::Relaxed);
+            return;
+        }
 
-        // Request a new frame.
-        surface.frame().quick_assign(move |_, _, _| {
-            should_draw.store(true, Ordering::Relaxed);
-        });
+        // Get the display vblank interval in micro seconds.
+        let monitor_vblank_interval = Duration::from_micros(
```
chrisduerr, line 1396: Why are you using microseconds here? I find that extremely strange considering basically everyone uses either ms or nanos usually.

Just for the sub-millis accuracy without having to use nanos which would be overly precise?

alacritty/src/display/mod.rs
```diff
@@ -1371,23 +1377,38 @@ impl Display {
...
+            (1e6 / (self
```
chrisduerr, line 1397: I find the `1e6` notation difficult to grasp at a glance and I rarely see that in source code likely for that reason. Sure it's shorter, but is it more readable than `1_000_000`?

If you go into the billions and higher it might be more helpful, but I think people will recognize much faster that `1_000_000` is a million rather than `1e6`.

This also applies to the `1e3` used later which is even worse since it's just 1000.

alacritty/src/display/mod.rs
```diff
@@ -1371,23 +1377,38 @@ impl Display {
...
+                .window
+                .current_monitor()
+                .and_then(|monitor| monitor.refresh_rate_millihertz())
+                .unwrap_or(60_000) as f64
+                / 1e3)) as u64,
```
chrisduerr, line 1402: What's the point of this division? Wouldn't it be easier to just do `1_000_000_000 / 60_000` rather than `1_000_000 / (60_000 / 1000)`?

alacritty/src/display/mod.rs
```diff
@@ -1371,23 +1377,38 @@ impl Display {
...
+        );
+
+        let swap_delay = self.frame_timer_info.compute_delay(monitor_vblank_interval);
```
chrisduerr, line 1405: Not sure `delay` is the right word here? Maybe `timeout` is more appropriate? Though I don't love it either.

alacritty/src/display/mod.rs
```diff
@@ -1529,6 +1550,54 @@ impl<T> DerefMut for Replaceable<T> {
     }
 }
 
+/// The frame timer state.
+#[derive(Default)]
+pub struct FrameTimerInfo {
+    /// The optimal delay for the renderer.
+    swap_delay: Option<Duration>,
+
+    /// The time stamp of the previous contiguos frame.
```
chrisduerr, line 1559:
```suggestion
    /// The timestamp of the previous contiguous frame.
```

Besides the typo, what is "contiguous frame" supposed to mean here? Isn't it already implied by "previous" that we're talking about the frame right before this?

I think just this should be better:

```suggestion
    /// The timestamp of the previous frame.
```

chrisduerr, line 1559:
```suggestion
    /// The timestamp of the previous contiguous frame.
```

Besides the typo, what is "contiguous frame" supposed to mean here? Isn't it already implied by "previous" that we're talking about the frame right before this?

I think just this should be better:

```suggestion
    /// The timestamp of the previous frame.
```

alacritty/src/display/mod.rs
```diff
@@ -1529,6 +1550,54 @@ impl<T> DerefMut for Replaceable<T> {
...
+        // If the refresh rate has changed, reset the timings.
+        if self.refresh_interval != refresh_interval || self.swap_delay.is_none() {
+            self.refresh_interval = refresh_interval;
+            self.frame_timestamp = Some(Instant::now());
+            self.swap_delay = Some(refresh_interval);
+            return refresh_interval;
+        }
+
+        // At this point we know that `swap_delay` has some value.
+        let swap_delay = self.swap_delay.unwrap();
```
chrisduerr, line 1579: Why not use this:

```rust
let swap_delay = match self.sway_delay.filter(|_| self.refresh_interval == refresh_interval) {
    Some(swap_delay) => swap_delay,
    None => {
        // Reset stuff
    },
};
```

Maybe a little worse for performance, but should basically be irrelevant.

alacritty/src/display/mod.rs
```diff
@@ -1529,6 +1550,54 @@ impl<T> DerefMut for Replaceable<T> {
...
+
+        let new_swap_delay = match self.frame_timestamp {
+            Some(last_frame_timestamp) => {
+                // Compute the error.
+                let error = self.refresh_interval.as_micros() as i64
+                    - last_frame_timestamp.elapsed().as_micros() as i64;
+                let new_swap_delay = swap_delay.as_micros() as i64 + error;
```
chrisduerr, line 1586: Assuming that a frame took a particularly long time, that will mean that immediately the next frame will get hurt a lot too. But I don't think there's any immediate connection between the last frame and the next frame, right?

So shouldn't we try to "stabilize" towards some rolling average rather than flip-flopping between a massive and tiny delay?

What you seem to be doing is basically just to try and **start** rendering every *refresh_rate* ms, while realistically what I think we'd want is to try and **finish** rendering every *refresh_rate* ms.

So I feel like we should add a rolling average to this and subtract it as a "buffer" for how long we'd expect rendering to take. Then keep your logic to adjust for when that buffer was insufficient or to much to ensure we're staying on track. But that way we should be able to aim to finish in time rather than starting in time.

alacritty/src/event.rs
```diff
@@ -1474,6 +1474,13 @@ impl Processor {
                             WinitEvent::RedrawEventsCleared,
                         );
                     }
+
+                    // Update the scheduler after we've done everything, so
+                    // the deadline will be more correct.
+                    *control_flow = match scheduler.update() {
+                        Some(instant) => ControlFlow::WaitUntil(instant),
+                        None => ControlFlow::Wait,
+                    };
```
chrisduerr, line 1483: This comment is useful for my review, but "more correct" as a relative term will be completely useless when you don't have the comparison to the old version anymore.

I'd suggest rewording to something like this:

```suggestion
                    // Update the scheduler after event processing, to ensure the event loop deadline is as accurate as possible.
                    *control_flow = match scheduler.update() {
                        Some(instant) => ControlFlow::WaitUntil(instant),
                        None => ControlFlow::Wait,
                    };
```

chrisduerr, line 1483: This comment is useful for my review, but "more correct" as a relative term will be completely useless when you don't have the comparison to the old version anymore.

I'd suggest rewording to something like this:

```suggestion
                    // Update the scheduler after event processing, to ensure the event loop deadline is as accurate as possible.
                    *control_flow = match scheduler.update() {
                        Some(instant) => ControlFlow::WaitUntil(instant),
                        None => ControlFlow::Wait,
                    };
```

chrisduerr, line 1483: This comment is useful for my review, but "more correct" as a relative term will be completely useless when you don't have the comparison to the old version anymore.

I'd suggest rewording to something like this:

```suggestion
                    // Update the scheduler after event processing, to ensure the event loop deadline is as accurate as possible.
                    *control_flow = match scheduler.update() {
                        Some(instant) => ControlFlow::WaitUntil(instant),
                        None => ControlFlow::Wait,
                    };
```

chrisduerr (changes requested):

alacritty/src/display/mod.rs
kchibisov: That's a good question. Simply ns is not enough, so I've picked next... Could use nanos if you feel like it.

chrisduerr: Micros should be fine I guess. The explanation makes enough sense where people shouldn't be completely confused reading this.

alacritty/src/display/mod.rs
kchibisov: aye, it's now a previous frame, it was a bit different before.

alacritty/src/display/mod.rs
kchibisov: >Assuming that a frame took a particularly long time, that will mean that immediately the next frame will get hurt a lot too.

The frame will fire immediately in this case. Not sure how correct this will be though.

>But I don't think there's any immediate connection between the last frame and the next frame, right?

The connection was initially made for contiguous frames, but it was showing worse perf then what it's right now. The initial idea was to compute the optimal `timeout` for contiguous frames and use it as a `default` when we don't know sufficient information about 'how much time we should timeout to render at the monitor refresh rate'.

>So shouldn't we try to "stabilize" towards some rolling average rather than flip-flopping between a massive and tiny delay?

It was slower, but maybe I was doing it wrong. Basically what I was trying to do is when we hit `draw` and we skip it due to frame not being dirty I was resetting the last timestamp. But I think what we could try to do is when the error gets too big as in it zeros the delay we might want to ignore adjusting the delay in this case. The large error means that we likely wasn't ready to draw.

>What you seem to be doing is basically just to try and start rendering every refresh_rate ms, while realistically what I think we'd want is to try and finish rendering every refresh_rate ms.

I think the latter is what I'm trying to do since I schedule the frame after all the rendering, so I'm accounting not for the start but for the end. The place is after all the rendering operations and I'm scheduling for that. If I were doing the same thing before the `swap_buffers` then yes, it'll be the former.

>So I feel like we should add a rolling average to this and subtract it as a "buffer" for how long we'd expect rendering to take. 

Rendering is already accounted since we do that after `gl::Finish` and `swap_buffers`, it could be wise to call the `gl::Finish` on other platforms though, however it should be the case on other platforms already with the swap buffers.


>Then keep your logic to adjust for when that buffer was insufficient or to much to ensure we're staying on track. But that way we should be able to aim to finish in time rather than starting in time.

We aiming on finish time already

So to sum it up, what could be adjusted is behavior when swap delay get zeroed, since it indeed breaks some scheduling, however when I was trying to do so with the method I've described above I got lower fps then with the current approach, which makes sense, since I was rendering with 0 delay the next frame as well or something like that.

I also used various techniques, like using `Meter` for averaging the error, but not sure whether it got me anywhere, I think I should retry at least, since last time I did that with wrong scheduler logic and haven't bothered to retest once I changed the scheduler timing re-computation.

alacritty/src/display/mod.rs
```diff
@@ -985,6 +985,8 @@ impl Display {
             self.renderer.finish();
         }
 
+        // XXX request for the rendering frame after swap buffers, so the time
+        // for finishing gl operations will be accounted in to the timeout.
```
chrisduerr, line 989:
```suggestion
        // XXX: Request the new frame after swapping buffers, so the time to finish OpenGL operations is accounted for in the timeout.
```

kchibisov:
@chrisduerr I've updated the logic of rescheduling. It's a bit more stable when running `yes` and also accounts for long delays...

alacritty/src/display/mod.rs
```diff
@@ -367,6 +368,9 @@ pub struct Display {
     /// The ime on the given display.
     pub ime: Ime,
 
+    /// The state of the timer for frame scheduling.
+    pub frame_timer_info: FrameTimerInfo,
```
chrisduerr, line 372:
```suggestion
    pub frame_timer: FrameTimer,
```

"Info" is a pretty useless suffix usually, I think we can just remove it.

alacritty/src/display/mod.rs
```diff
@@ -1376,23 +1385,40 @@ impl Display {
...
+                .window
+                .current_monitor()
+                .and_then(|monitor| monitor.refresh_rate_millihertz())
+                .unwrap_or(60_000) as f64;
 
-        // Mark that window was drawn.
-        should_draw.store(false, Ordering::Relaxed);
+        // Now convert it to micro seconds.
+        let monitor_vblank_interval =
+            Duration::from_micros((1000. * monitor_vblank_interval) as u64);
```
chrisduerr, line 1413:
```suggestion
            Duration::from_micros(1000 * monitor_vblank_interval as u64);
```

alacritty/src/display/mod.rs
```diff
@@ -1534,6 +1560,63 @@ impl<T> DerefMut for Replaceable<T> {
...
+            self.swap_timeout = refresh_interval;
+            return refresh_interval;
+        }
+
+        // Compute the error.
+        let error = self.refresh_interval.as_micros() as i64
+            - self.frame_timestamp.elapsed().as_micros() as i64;
+
+        // Don't update error and swap timeout if we've waited for too long.
+        if error < self.refresh_interval.as_micros() as i64 {
```
chrisduerr, line 1604: I'm somewhat concerned about this logic here. It will cause an immediate scheduling which I guess is fine but that will mean our "clock" gets reset. Thus the new frame is not in line with our previous frames.

Shouldn't we aim to render at the next "modulo frame interval" time, so that our rendering stays on track and keeps its consistent clock?

Otherwise I'm afraid that if we regularly skip drawing (like every other frame this happens), we'd get very inconsistent latencies (not technically worse, just inconsistent).

alacritty/src/display/mod.rs
```diff
@@ -1534,6 +1560,63 @@ impl<T> DerefMut for Replaceable<T> {
...
+            // Average previously used error with the current one to reduce momentum from the
+            // error sign changes.
+            self.previous_error = (self.previous_error + error) / 2;
+
+            let new_swap_timeout = self.swap_timeout.as_micros() as i64 + self.previous_error;
```
chrisduerr, line 1609: Hm I'm a little concerned with this. Previously since we adjusted for the error **exactly** we'd obviously always draw in `swap_timeout` intervals. But now that we adjust for this error, we have no reference point for our "default" clock cycle anymore. So we'd drift off course since the error correction with the previous frame is not precise.

Instead of taking the `swap_timeout` I feel like we should be calculating the time until the next clock cycle (based on start time) and use that instead?

chrisduerr (changes requested):

alacritty/src/display/mod.rs
kchibisov: No, you'll drop the fractional part?

alacritty/src/display/mod.rs
kchibisov: But I'm not using `zero`? I'm just using whatever we previously accumulated, just don't recompute and update error?

alacritty/src/display/mod.rs
kchibisov: >point for our "default" clock cycle anymore. So we'd drift off course since the error correction with the previous frame is not precise.

We use `refresh_interval` for that? So we have a reference, we just try to converge to it slower.

>Instead of taking the swap_timeout I feel like we should be calculating the time until the next clock cycle (based on start time) and use that instead?

Could you elaborate on that? You mean that? For now we use the refresh interval to cap us at refresh rate, do you want some arbitrary logic like get `Instant::now()` at startup, and then draw only on  `start + X * modulo`? I'm not sure it'll be any better to be honest...

alacritty/src/display/mod.rs
chrisduerr: Ah yeah that's fair, I wasn't considering that we want to keep the fractional precision of the variable.

alacritty/src/display/mod.rs
chrisduerr: Yeah it's fair that you're not using zero, but you're also not trying to adjust for the clock cycle at all. It doesn't really matter if you're using zero, last `swap_interval` or `swap_timeout` here, none of these will reliably match up with your previous clock cycles.

alacritty/src/display/mod.rs
chrisduerr: > We use refresh_interval for that? So we have a reference, we just try to converge to it slower.

You have an interval that converges using two historical datapoints. But let's assume this averaging has introduced a 1 microsecond drift in one direction due to rounding. You're never accounting for that drift in the future since all you're doing is just taking the last two errors and averaging them. As a result a drift of 1 microsecond is introduced, shifting your clock permanently.

> Could you elaborate on that? You mean that? For now we use the refresh interval to cap us at refresh rate, do you want some arbitrary logic like get Instant::now() at startup, and then draw only on start + X * modulo? I'm not sure it'll be any better to be honest...

I'm not sure about the exact implementation or your suggestion, but basically yeah. We're already using `elapsed()` to calculate the error, what I'd suggest basically is just doing the same but keeping our clock consistent with the start of the clock rather than consistent with the last frame. That should improve frame consistency.

alacritty/src/display/mod.rs
```diff
@@ -1534,6 +1560,52 @@ impl<T> DerefMut for Replaceable<T> {
...
+        if self.refresh_interval != refresh_interval {
+            self.base = now;
+            self.last_synced_timestamp = now;
+            self.refresh_interval = refresh_interval;
+            return refresh_interval;
+        }
+
+        let next_frame = self.last_synced_timestamp + self.refresh_interval;
+
+        if next_frame < now {
```
chrisduerr, line 1596:
```suggestion
        if next_frame < now {
        // Redraw immediately if we haven't drawn in over `refresh_interval` microseconds.
```

I think adding some comments here would be nice.

alacritty/src/display/mod.rs
```diff
@@ -1534,6 +1560,52 @@ impl<T> DerefMut for Replaceable<T> {
...
+            let elapsed_micros = (now - self.base).as_micros() as u64;
+            let refresh_micros = self.refresh_interval.as_micros() as u64;
+            self.last_synced_timestamp =
+                now - Duration::from_micros(elapsed_micros % refresh_micros);
+            Duration::ZERO
+        } else {
+            self.last_synced_timestamp = next_frame;
```
chrisduerr, line 1603:
```suggestion
            // Redraw on the next `refresh_interval` clock tick.
            self.last_synced_timestamp = next_frame;
```

I think adding some comments here would be nice.

chrisduerr (approved):

PR #6590, opened 2022-12-27
Disable depth and stencil buffers

### LINKED ISSUE #2881: Disable stencil/depth buffers
Currently Alacritty initializes both the depth and stencil buffer, which leads to additional video memory usage.

This has been disabled in 3475e449870b382cda4ea6d48f980577cd8c929e, however some X11 setups had issues with this and it lead to Windows getting initialized with garbage on newer amdgpu Systems.

Because of this the commit has been reverted in https://github.com/jwilm/alacritty/pull/2882.

Some additional info can be found in https://github.com/jwilm/alacritty/pull/2875.

chrisduerr:
This patch reduces the active GPU memory consumption by disabling the depth and stencil buffers. During original testing it reduced GPU memory usage on Linux by almost a third.

This is a reintroduction of previously reverted patch 3475e44.

Closes #2881.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Blurry fonts when changing padding size at runtime
 - Crash while typing on Wayland
 - Multi-line semantic bracket selection
+- Reduced GPU memory usage
 
 ## 0.11.0
 
--- a/alacritty/src/renderer/platform.rs
+++ b/alacritty/src/renderer/platform.rs
@@ -42,7 +42,10 @@ pub fn pick_gl_config(
     gl_display: &Display,
     raw_window_handle: Option<RawWindowHandle>,
 ) -> Result<Config, String> {
-    let mut default_config = ConfigTemplateBuilder::new().with_transparency(true);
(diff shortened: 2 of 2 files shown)


chrisduerr:
I've tested this on my 6900 XT, which should be using a modern amdgpu driver and I wasn't able to reproduce the original issue. However I've never reproduced the original issue either, so I can't confirm for sure that this doesn't just reintroduce it.

kchibisov:
This would need broader testing, since setting both of them to 0 could change the driver picking on Windows from what I've understood. It's correct that we don't need any of them, however the hardware might not have the accelerated formats without stencil/depth, since you usually need them and our case is not that common.

chrisduerr:
> This would need broader testing, since setting both of them to 0 could change the driver picking on Windows from what I've understood.

If someone on Windows wants to test it, it would be much appreciated. But otherwise I'd also be fine with just giving it a shot to see if someone on Windows complaints. If it doesn't work on Windows the best solution would probably be to just `cfg` it anyway? So we wouldn't completely remove it.

kchibisov:
Does it decrease the video memory usage for you btw?

chrisduerr:
31720 KiB -> 19432 Kib, so my reduction is above a third even (~39% improvement).

kchibisov (approved):
Tested the rest of the platforms.

PR #6594, opened 2022-12-29
Fix crash when one dimension of the window is zero

kchibisov:
This fixes a crash on Windows when the user resizes the window to the point that it has the height of zero. The crash was introduced by the glutin update, since it requires non-zero sizes for the resize.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1176,13 +1176,13 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                 match event {
                     WindowEvent::CloseRequested => self.ctx.terminal.exit(),
                     WindowEvent::Resized(size) => {
-                        // Minimizing the window sends a Resize event with zero width and
-                        // height. But there's no need to ever actually resize to this.
-                        // ConPTY has issues when resizing down to zero size and back.
-                        //
                         // The user could also resize the window to have zero height
-                        // due to the titlebar.
-                        #[cfg(windows)]
+                        // due to the window decorations and such, so ignore zero sizes.
+                        //
+                        // Also on Windows, minimizing the window sends a Resize event with zero
+                        // width and height. But there's no need to ever
+                        // actually resize to this. ConPTY has issues when
+                        // resizing down to zero size and back.
                         if size.width == 0 || size.height == 0 {
                             return;
                         }
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1179,8 +1179,11 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                         // Minimizing the window sends a Resize event with zero width and
                         // height. But there's no need to ever actually resize to this.
                         // ConPTY has issues when resizing down to zero size and back.
+                        //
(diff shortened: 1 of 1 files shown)


alacritty/src/event.rs
```diff
@@ -1179,8 +1179,11 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                         // Minimizing the window sends a Resize event with zero width and
                         // height. But there's no need to ever actually resize to this.
                         // ConPTY has issues when resizing down to zero size and back.
+                        //
+                        // The user could also resize the window to have zero height
+                        // due to the titlebar.
                         #[cfg(windows)]
-                        if size.width == 0 && size.height == 0 {
+                        if size.width == 0 || size.height == 0 {
```
chrisduerr, line 1186: Is this actually creating ConPTY issues, or is our code struggling with the zero size? You mentioned winit, so if that's a change introduced to winit we might have to add this check to all platforms?

chrisduerr (changes requested):

alacritty/src/event.rs
kchibisov: I mean you can't resize the window to the size of zero, you can't have a buffer that way? Our code will have issues with zero sizes, because you can't make a zero buffer, but that's nothing we can do other than ignoring...

kchibisov: Note, that it's probably wise to ignore zero sizes on all platforms, just in case, but yet I haven't seen them coming anywhere other than windows...

chrisduerr: > I mean you can't resize the window to the size of zero, you can't have a buffer that way? Our code will have issues with zero sizes, because you can't make a zero buffer, but that's nothing we can do other than ignoring...

Can you not do that on other platforms though? I'm pretty sure I added an i3 option to limit my window size because I can resize it to 0x0.

chrisduerr: Ah it seems like X11 backend might be clamped to 1x1 internally. I can make the window completely disappear but it still reports a resize event of 1x1.

alacritty/src/event.rs
```diff
@@ -1176,13 +1176,13 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                 match event {
                     WindowEvent::CloseRequested => self.ctx.terminal.exit(),
                     WindowEvent::Resized(size) => {
-                        // Minimizing the window sends a Resize event with zero width and
-                        // height. But there's no need to ever actually resize to this.
-                        // ConPTY has issues when resizing down to zero size and back.
-                        //
                         // The user could also resize the window to have zero height
```
chrisduerr, line 1179: Having `also` here doesn't make any sense anymore since it's the first sentence now.

Suggesting this:

```
// Ignore resize events to zero in any dimension, to avoid issues with Winit and the ConPTY.
//A 0x0 resize will also occur when the window is minimized on Windows.
```

I think this should cover all the information we currently have but provide it in a little more structured manner.

chrisduerr (changes requested):

chrisduerr (approved):
Assuming this isn't present in the latest stable version so we don't need a changelog entry.

Issue #6060, opened 2022-05-13
SpawnNewInstance should not pass --command (-e) argument


WebOfNakedFancies:
When alacritty is run with --command (-e) option, SpawnNewInstance is also passed this argument.

My specific use case was running "alacritty -e ranger", and I expected SpawnNewInstance to give me a terminal for the directory I am currently parsing. Instead, it launched a new alacritty instance with ranger running.
When I use ranger without invoking alacritty's --command option, it works as expected.

I don't know if there's some other use case benefiting from the current behavior.

I propose that it should be ignored like the --working-directory argument already is, as in the below code 

https://github.com/alacritty/alacritty/blob/9bbb296d1dc4f72e845695f22e903bcc1d66d5a3/alacritty/src/event.rs#L369-L372

chrisduerr:
You're absolutely correct. In fact if you do `alacritty msg create-window`, it should already behave as you would expect. Apparently it's just the `SpawnNewInstance` action that doesn't behave as intended. So you can use the command as a work around until this is fixed.

Alkeryn:
I just looked it up whilst switching from termite, nice to see someone opened the issue just 2 days ago !

conraythefirst:
Is there any alternative way to spawn a new instance through a key binding that actually inherits the --command param?
I understand why it's unexpected behaviour for SpawnNewInstance but I grew to like it and made heavy use of it.

PR #6614, opened 2023-01-11
add DuplicateInstance Action

conraythefirst:
Added Action 'DuplicateInstance' wich spawns a new instance but inherits all args.
Useful when working with --command arguments.
Reaction to #6060

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -180,6 +180,9 @@ pub enum Action {
     /// Spawn a new instance of Alacritty.
     SpawnNewInstance,
 
+    /// Duplicate instance of Alacritty.
+    DuplicateInstance,
+
     /// Create a new Alacritty window.
     CreateNewWindow,
 
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -409,13 +409,6 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         // Reuse the arguments passed to Alacritty for the new instance.
         #[allow(clippy::while_let_on_iterator)]
         while let Some(arg) = env_args.next() {
-            // On unix, the working directory of the foreground shell is used by `start_daemon`.
-            #[cfg(not(windows))]
-            if arg == "--working-directory" {
-                let _ = env_args.next();
-                continue;
-            }
-
             args.push(arg);
         }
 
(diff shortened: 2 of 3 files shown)


chrisduerr:
Why are you using `--command` arguments that you want propagated, instead of using Alacritty's config for it? I'm not sure this patch makes sense to me.

conraythefirst:
I use it in combination with suckless tabbed and also for ssh or tmux sessions

I run alacritty with something like
`alacritty --config-file ~/.config/alacritty/alacritty-tabbed.yml --embed $tabbed_xid --class Shell -t $host -e bash -c "ssh $host"`

It's convenient to be able to just "clone" your session for debugging stuff etc

chrisduerr:
So couldn't you just put the `bash -c "ssh $host"` as the shell in the `alacritty-tabbed.yml`?

conraythefirst:
That config is not only for ssh sessions, but some general config to make it look good with my monitor setup and tabbed.

Since $host can be one of many hosts it would involve constantly rewriting that file. With multiple sessions on multiple hosts that might get tricky.
Before #6060 SpawnNewInstance would actually behave that way and I liked it and have gotten used to it, so I'm trying to reintroduce that functionality.

kchibisov:
That could be useful if you work with an editor opened from the desktop entry, so you don't have to go through-out the desktop entry stuff again in your launcher. I'm not sure how common is that.

But we have a situation where we also have windows and you might do the same for them? Since I always have exactly one instance and just open new windows. So such change would be useless to me, given that even apps from launcher are opened via IPC.

Maybe I'd be fine with adding `DuplicateInstance` and `DuplicateWindow` given that if you work on a remote host you might spawn window via IPC with `-e` to **ssh** into the host, so you can open new windows from that one automatically connecting to the one you want.

As alternative as it was stated you should create temporary config files, which is basically your file + derivation in `programs.shell`, which is way less ergonomic.

conraythefirst:
Is there a way to open a new window within the same instance that inherits --command? 
That's the goal anyway, "cloning" the current window. It doesn't need to be a new instance necessarily.

chrisduerr:
Config options are inherited, CLI options are not. That's the way it works for all options.

chrisduerr:
Going to close this PR, since even if we add a feature like this it won't look like this PR.

nixpulvis:
As a watcher from a distance, I'm wondering if it might get to be time to allow actions to have options themselves so-as to avoid having a million similar action names in the enumeration?

chrisduerr:
@nixpulvis This has been considered, but we could never come up with a design that would be remotely clean enough to make it sensible.

That said if the goal is just to create another instance with flags, you can just use `command` for that.

PR #6625, opened 2023-01-15
Add support for horizontal scrolling

### LINKED ISSUE #2185: Horizontal mouse scroll wheel
Those of us on macOS have horizontal scroll wheels.  When "all motion mouse tracking" is enabled Alacritty sends mouse wheel scroll events as CSI M 64 and CSI M 65.

Can we have an option to enable generation of horizontal scroll wheel events, as buttons 8 and 9, which I believe would generate CSI M 128 and CSI M 129 respectively.

https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Other-buttons

kchibisov:
This adds support for horizontal mouse scrolling in mouse mode and alternative scrolling modes.

Fixes #2185.

### DIFF
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -631,18 +631,19 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
 
     pub fn mouse_wheel_input(&mut self, delta: MouseScrollDelta, phase: TouchPhase) {
         match delta {
-            MouseScrollDelta::LineDelta(_columns, lines) => {
-                let new_scroll_px = lines * self.ctx.size_info().cell_height();
-                self.scroll_terminal(f64::from(new_scroll_px));
+            MouseScrollDelta::LineDelta(columns, lines) => {
+                let new_scroll_px_x = columns * self.ctx.size_info().cell_width();
+                let new_scroll_px_y = lines * self.ctx.size_info().cell_height();
+                self.scroll_terminal(new_scroll_px_x as f64, new_scroll_px_y as f64);
             },
             MouseScrollDelta::PixelDelta(lpos) => {
                 match phase {
                     TouchPhase::Started => {
                         // Reset offset to zero.
-                        self.ctx.mouse_mut().scroll_px = 0.;
+                        self.ctx.mouse_mut().accumulated_scroll = Default::default();
                     },
                     TouchPhase::Moved => {
-                        self.scroll_terminal(lpos.y);
+                        self.scroll_terminal(lpos.x, lpos.y);
                     },
                     _ => (),
                 }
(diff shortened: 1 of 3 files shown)


alacritty/src/input.rs
```diff
@@ -631,37 +631,52 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
 
         if self.ctx.mouse_mode() {
-            self.ctx.mouse_mut().scroll_px += new_scroll_px;
+            self.ctx.mouse_mut().accumulated_scroll.x += new_scroll_x_px;
+            self.ctx.mouse_mut().accumulated_scroll.y += new_scroll_y_px;
 
-            let code = if new_scroll_px > 0. { 64 } else { 65 };
-            let lines = (self.ctx.mouse().scroll_px / height).abs() as i32;
+            let code = if new_scroll_y_px > 0. { MOUSE_WHEEL_UP } else { MOUSE_WHEEL_DOWN };
+            let lines = (self.ctx.mouse().accumulated_scroll.y / height).abs() as i32;
```
chrisduerr, line 668: This `lines` coupled with the iterator should take care of the case where no line is actually scrolled vertically, right?

chrisduerr (approved):
I don't think I have any good way to test this, but the code looks trivial and correct. So should be all good.

alacritty/src/input.rs
kchibisov: Yes, it was working for a long time already :p

kchibisov:
>I don't think I have any good way to test this, but the code looks trivial and correct. So should be all good.

I don't think that I've found a client actually using this. But I've at least compared with xterm + mouse mode to get it the same codes on my mouse with horizontal wheel.

chrisduerr:
> I don't think that I've found a client actually using this. But I've at least compared with xterm + mouse mode to get it the same codes on my mouse with horizontal wheel.

`less` can do horizontal scrolling, right? At least iirc it has an option to show `>` at the end of the line and allow you to scroll horizontally.

kchibisov:
>less can do horizontal scrolling, right? At least iirc it has an option to show > at the end of the line and allow you to scroll horizontally.

Yeah, it scrolls. But it's alternate scroll. I was more interested in mouse mode.

chrisduerr:
Try htop, it seems to support horizontal scrolling with arrow keys at least.

kchibisov:
htop is using mouse mode, so it's irrelevant.

chrisduerr:
> Yeah, it scrolls. But it's alternate scroll. I was more interested in mouse mode.

I thought you were looking for an application using mouse mode?

kchibisov:
You mean to test? Nah, I know that we produce the right escapes compared to xterm.

nixpulvis:
I'm testing this with `less` on a macOS laptop with a touchpad. My first thought is that it might be worth taking a note from other applications which only allow continuous vertical or horizontal scrolls but not both at the same time. I'm finding myself having to scroll diagonally to say on an edge of the view and completely unable to scroll reliably when not on an edge.

Perhaps something like #6712 could help. But I haven't gotten to testing that yet, since I need to setup my tablet again.

nixpulvis:
Reading #6747, I've now tested on master, sorry for not doing this first. For some reason I thought that change had already been rolled in and my lazy ass forgot to do due diligence.

Seems like the diagonal cutoff basically implements what I was expecting with the U example above.

Thanks :)

PR #6626, opened 2023-01-15
Alacritty version 0.11.1-rc1

chrisduerr:
---

This is a somewhat big patch release, which is why I've actually started out with the latest master and just dropped patches one might consider "too breaking".

It might be worth considering just making this into a 0.12.0 release?


roland-rollo:
Personally I think even without few things from master, this is too big for 11.1 version. Looking at changelog in master, releasing a 12.0 version looks more logical.

kchibisov:
Yeah, we decided like that on irc. Just haven't changed anything here.

chrisduerr:
I'll close this PR for now to not cause any confusion. It's not like this is going to be useful for a 0.12.0 release anyway.

Issue #6202, opened 2022-07-11
Include alacritty.yaml in Mac Alacritty.app


rektide:
Hi. In #6199 I asked for an easy way to get the default `alacritty.yaml`. @chrisduerr countered distributions ought include this.

> If your repository does have a package, I'd suggest recommending to them that they include the repo's configuration file.

I'm requesting the MacOS distribution on the release page include an `alacritty.yaml`.

chrisduerr:
What exactly would you do with the `alacritty.yml`, were it to be included in the `dmg`? I doubt there would be a lot of people moving it out of there and I'd actually discourage that usually.

kchibisov:
@chrisduerr we do have like everything(completions, man page) in Alacritty.app, the problem is that we don't have a config there :/ 

Seems like it should be added, given that we copy a bunch of stuff there as well.

mhmdanas:
> What exactly would you do with the alacritty.yml, were it to be included in the dmg? I doubt there would be a lot of people moving it out of there and I'd actually discourage that usually.

Homebrew does put reference configs in `/usr/local/share/examples` (for instance, it puts one for Mirage in `/usr/local/share/examples/mirage/settings.py`), and I for one would certainly appreciate having a reference config on my system instead of having to open it on Alacritty's repo on GitHub.

kchibisov:
Was done in #6639.

PR #6639, opened 2023-01-22
Include alacritty.yml in dmg

h-leth:
Ref  #6202

Updated Makefile to include alacritty.yml in dmg

Let me know if anything is wrong/need to be changed


chrisduerr:
I still don't see a reasonable response to https://github.com/alacritty/alacritty/issues/6202#issuecomment-1181529918.

What is your goal with the config in the dmg? We do not want users to copy it in their Alacritty config directory.

kchibisov:
The main point is that it's for the version you shipped and available offline? It's not like you copy it, but rather look into it as a reference.

kchibisov:
It's something common when packaging stuff on linux, since you also package the config into the docs of the binary. On most distros over config is shipped already and put into the `/usr/share/docs/alacritty/examples`. The `dmg` is basically the same for macOS.

h-leth:
It's up to you guys to decide what to do. I just tried to close a open issue.

chrisduerr:
> The main point is that it's for the version you shipped and available offline? It's not like you copy it, but rather look into it as a reference.

The thing is literally nobody will do this though. Everyone will just copy it to the config directory and complain about issues with uncommenting shit.

No macOS user is aware of some global directory for documentation on configuration files. And the dmg will be removed after installation.

Makefile
```diff
@@ -47,6 +48,7 @@ $(APP_NAME)-%: $(TARGET)-%
 	@mkdir -p $(APP_COMPLETIONS_DIR)
 	@gzip -c $(MANPAGE) > $(APP_EXTRAS_DIR)/alacritty.1.gz
 	@gzip -c $(MANPAGE-MSG) > $(APP_EXTRAS_DIR)/alacritty-msg.1.gz
+	@cp -fp $(CONFIGFILE) $(APP_EXTRAS_DIR)/
```
chrisduerr, line 51: Can you move this down to the other lines that are just `cp -fp`, just for consistency?

chrisduerr (changes requested):
I guess we can just add it and tell users they're dumb when they copy it over and shit breaks.

h-leth:
Can rename it on copy to eg. `alacritty_example.yml`. But I'll think literally no one is going to look for it inside the .app file.

chrisduerr (approved):

chrisduerr:
`alacritty.yml` should be fine, not interested in people opening bug reports because they copied it over and changing things didn't work.

PR #6640, opened 2023-01-23
Add .d convention for a configuration directory in bash completions

CristopherA96:
Hi,

## Intro
I was installing alacritty in Ubuntu 20.04 LTS and when I add the bash completions, every time that I opened the terminal, the next message was shown.

 **bash: .: /home/<my_user>/.bash_completion: is a directory**
 
## Solution
So, the way that I solved this, was by adding the suffix `.d` to the configuration directory `.bash_completion`. Like this `.bash_completion.d` 

```sh
mkdir -p ~/.bash_completion.d
cp extra/completions/alacritty.bash ~/.bash_completion.d/alacritty
echo "source ~/.bash_completion.d/alacritty" >> ~/.bashrc
```

The reason to do that was because most of Linux distributions consider `.bash_completion` as a configuration file not a configuration directory.

## Conclusion
I think this may be helpful for people when installing alacritty and the bash completions.

Thanks for your time

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -376,6 +376,14 @@ cp extra/completions/alacritty.bash ~/.bash_completion/alacritty
 echo "source ~/.bash_completion/alacritty" >> ~/.bashrc
 ```
 
+Most Linux distributions consider that `~/.bash_completion` is a configuration file, not a configuration directory. To avoid name clashing, there is a convention for creating a configuration directory, and it is to add the suffix `.d` at the end of the directory `~/.bash_completion.d`. If you have any issue, you can run:
+
+```sh
+mkdir -p ~/.bash_completion.d
+cp extra/completions/alacritty.bash ~/.bash_completion.d/alacritty
+echo "source ~/.bash_completion.d/alacritty" >> ~/.bashrc
+```
+
 #### Fish
 
 To install the completions for fish, run


chrisduerr:
> The reason to do that was because most of Linux distributions consider .bash_completion as a configuration file not a configuration directory.

Source? This just sounds like you messed up your shell configuration. Why would a distro's configuration load your user's `~/.bash_completion` as a file by default?

CristopherA96:
It is a name clashing if I do not do that.

But it is okay, if it is not necessary.

CristopherA96:
It was only the way to solve the message every time that I opened
alacritty. Because the problem is with the bash_completion in the default
config bashrc. So there is a name clashing, and only doing the instructions
that you have in the INSTALL.md file, it showed me that message. I try in
another Debian based distro and it happened the same.

chrisduerr:
> in the default config bashrc

Which default bashrc are you talking about? `/etc/bash.bashrc` or `~/.bashrc?
Because `/etc/bash.bashrc` is the only default and as far as I know it should only load `/usr/share/bash-completion/bash_completion`.

CristopherA96:
In the ~/.bashrc I have the bash_completion instructions which load
/usr/share/bash-completion/bash_completion, and in the /etc/bash.bashrc I
have the same instructions but commented. I mean with "default" because it
is the configuration that comes in a new distro installation, so I did not
change anything in the ~/.bashrc because it is a new installation.

chrisduerr:
Everything in `~/.bashrc` is done by you. If `/etc/bash.bashrc` doesn't load the file in home, then you screwed up your system.

PR #6645, opened 2023-01-25
Make gles2 renderer actually gles2

kchibisov:


kchibisov:
This is draft only because I haven't made glutin release with functionality I use.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Uppercase `-T` short form for `--title`
 - Support for horizontal scrolling in mouse mode and alternative scrolling modes
 - Support for fractional scaling on Wayland with wp-fractional-scale protocol
+- Support for running on GLES context.
```
chrisduerr, line 15:
```suggestion
- Support for running on GLES context
```

alacritty/src/renderer/platform.rs
```diff
@@ -81,21 +81,31 @@ pub fn create_gl_context(
...
-                .with_profile(GlProfile::Compatibility)
-                .with_context_api(ContextApi::OpenGl(Some(Version::new(2, 1))))
-                .build(raw_window_handle);
-            gl_display.create_context(gl_config, &context_attributes)
-        }
+    let mut profiles = [
+        ContextAttributesBuilder::new()
+            .with_context_api(ContextApi::OpenGl(Some(Version::new(3, 3))))
+            .build(raw_window_handle),
+        // Try gles before OpenGL 2.1 as it tends to be more stable.
```
chrisduerr, line 88: Curious why you say this? Is this also the case on macOS for example?

alacritty/src/renderer/text/atlas.rs
```diff
@@ -144,11 +158,24 @@ impl Atlas {
...
+                    // explicit copy.
+                    if self.is_gles_context {
+                        let mut new_buffer = Vec::with_capacity(buffer.len() / 3 * 4);
+                        for rgb in buffer.chunks_exact(3) {
+                            new_buffer.push(rgb[0]);
+                            new_buffer.push(rgb[1]);
+                            new_buffer.push(rgb[2]);
+                            new_buffer.push(u8::MAX);
+                        }
+                        (gl::RGBA, Cow::Owned(new_buffer))
```
chrisduerr, line 171: Would it make sense to do this at an earlier point, on the `RasterizedGlyph` directly? Just so we can avoid pointlessly moving from one vec to a freshly allocated second one.

chrisduerr (changes requested):

alacritty/src/renderer/platform.rs
kchibisov: macOS is always a OpenGL 3.3 renderer.

alacritty/src/renderer/text/atlas.rs
kchibisov: Do you mean inside the crossfont?

alacritty/src/renderer/platform.rs
kchibisov: We might swap them, it just far less easier to achieve CTS conformance on gles then on OpenGL 2.1, and that's what mostly tested.

alacritty/src/renderer/text/atlas.rs
chrisduerr: Is it a `Vec` or a `&Vec`/`&[]`? I forgot.

kchibisov: It's an owning `Vec` that is passed to you from crossfont.

chrisduerr: Then it should be possible for us to mutate it?

kchibisov: Yeah, you need to pass rasterized glyph as mut and such... Be aware that OpenGL 3.3 will also copy under the hood from what I understand.

chrisduerr (approved):

PR #6648, opened 2023-01-26
create new windows with Cmd+N by default

chrisduerr:
This changes the default Cmd+N binding on macOS to create a new window rather than spawning a new instance.

Initially this change was held back for further testing of the multi-window feature. At this point all significant issues found with it have been fixed so it should be ready for prime-time now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `font.glyph_offset.y` is now applied to underline/strikeout
 - Always use sRGB color space on macOS
 - Erase in line after the last column will no longer clear the last column
+- Open new windows by default with macOS `Cmd`+`N` binding
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


kchibisov (changes requested):
Given that you change default behavior you should update the changelog entry.

kchibisov (approved):

utkarshgupta137:
On macOS, most apps use `Cmd + T` to open a new tab in the same window (CreateNewWindow) & `Cmd + N` to open a new separate window (SpawnNewInstance). Any reason why `Cmd + N` was changed instead of using `Cmd + T` like most other mac apps?

utkarshgupta137:
> On macOS, most apps use `Cmd + T` to open a new tab in the same window (CreateNewWindow) & `Cmd + N` to open a new separate window (SpawnNewInstance). Any reason why `Cmd + N` was changed instead of using `Cmd + T` like most other mac apps?

Please ignore my comment, it looks like `CreateNewWindow` is creating a new tab due to this setting in mac:

image

So I guess it makes sense to keep `Cmd + N` for people who've not changed that setting.

axieax:
Is there a particular reason why the default isn't `Cmd + T` for creating a new tab (`CreateNewWindow`) and `Cmd + N` for creating a new window (`SpawnNewInstance`), as would be expected for any MacOS shortcut? Happy to make a PR to address this change.

chrisduerr:
There's no difference between creating a new tab or window in Alacritty. It just depends on your system configuration.

PR #6649, opened 2023-01-26
Prefer environment variables over passwd

TitanNano:
If the `passwd` lookup for any reason fails, the entire TTY creation fails as well. It is actually more desirable to check the matching environment variables for `USER`, `LONGNAME` and `HOME` first before trying to access the `passwd`. This is actually already being done for `SHELL`.

### DIFF
--- a/alacritty_terminal/src/tty/unix.rs
+++ b/alacritty_terminal/src/tty/unix.rs
@@ -118,19 +118,55 @@ impl Pty {
     }
 }
 
-/// Look for a shell in the `$SHELL` environment variable, then in `passwd`.
-fn default_shell(pw: &Passwd<'_>) -> String {
-    env::var("SHELL").unwrap_or_else(|_| pw.shell.to_owned())
+/// User information that is required for a new shell session.
+struct ShellUser {
+    user: String,
+    home: String,
+    shell: String,
+}
+
+impl ShellUser {
+    /// look for shell, username, longname, and home dir in the respective environment variables
+    /// before falling back on looking in to `passwd`.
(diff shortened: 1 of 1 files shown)


chrisduerr:
> If the passwd lookup for any reason fails, the entire TTY creation fails as well.

"For any reason"? What reason would that be?

> It is actually more desirable to check the matching environment variables for USER, LONGNAME and HOME first before trying to access the passwd.

Source?

chrisduerr (changes requested):
Generally I'd say the environment variables have a much higher chance of failing than the `passwd` lookup. Not being able to read `passwd` will break **numerous** applications that try to do the same thing.

I'd be fine with looking up the environment variables first and falling back to passwd if those are unset. But environment variables alone are insufficient.

TitanNano:
> Generally I'd say the environment variables have a much higher chance of failing than the `passwd` lookup. Not being able to read `passwd` will break **numerous** applications that try to do the same thing.
> 
> I'd be fine with looking up the environment variables first and falling back to passwd if those are unset. But environment variables alone are insufficient.

The patch currently tries to fetch each of the environment variables and if a variable fails it falls back to the matching value in the Passwd struct.

chrisduerr:
> The patch currently tries to fetch each of the environment variables and if a variable fails it falls back to the matching value in the Passwd struct.

Oh sorry, I didn't look at the patch close enough.

chrisduerr:
Will try to get a review done today or tomorrow, just fyi.

chrisduerr (changes requested):

chrisduerr (approved):

PR #6653, opened 2023-01-28
Fix notify doing active polling

### LINKED ISSUE #6652: Debouncer is constantly polling the events
With the notify v4 we were doing passive waiting, however update to notify v5 changed that and now we have active polling that wakes up every 10ms draining the battery.

kchibisov:
The `notify-debouncer-mini` spawn a thread which checks the events every timeout, which is not desired since we want to avoid active polling.

This commit re-implements debouncer based on the `RecommendedWatcher` without adding an extra thread on top and not doing any busy-waiting.

Fixes #6652.

### DIFF
--- a/alacritty/src/config/monitor.rs
+++ b/alacritty/src/config/monitor.rs
@@ -1,20 +1,19 @@
 use std::path::PathBuf;
-use std::sync::mpsc;
-use std::time::Duration;
+use std::sync::mpsc::{self, RecvTimeoutError};
+use std::time::{Duration, Instant};
 
 use log::{debug, error};
-use notify_debouncer_mini::new_debouncer;
-use notify_debouncer_mini::notify::RecursiveMode;
+use notify::{Config, EventKind, RecommendedWatcher, RecursiveMode, Watcher};
 use winit::event_loop::EventLoopProxy;
 
 use alacritty_terminal::thread;
 
 use crate::event::{Event, EventType};
 
-#[cfg(any(target_os = "linux", target_os = "macos", target_os = "windows"))]
 const DEBOUNCE_DELAY: Duration = Duration::from_millis(10);
-#[cfg(not(any(target_os = "linux", target_os = "macos", target_os = "windows")))]
-const DEBOUNCE_DELAY: Duration = Duration::from_millis(1000);
+
+/// The fallback for `RecommendedWatcher` polling.
+const FALLBACK_POLLING_TIMEOUT: Duration = Duration::from_secs(1);
 
 pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
     // Don't monitor config if there is no path to watch.
@@ -41,8 +40,11 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
 
     // The Duration argument is a debouncing period.
     let (tx, rx) = mpsc::channel();
-    let mut debouncer = match new_debouncer(DEBOUNCE_DELAY, None, tx) {
-        Ok(debouncer) => debouncer,
+    let mut watcher = match RecommendedWatcher::new(
+        tx,
+        Config::default().with_poll_interval(FALLBACK_POLLING_TIMEOUT),
+    ) {
+        Ok(watcher) => watcher,
         Err(err) => {
             error!("Unable to watch config file: {}", err);
             return;
@@ -62,7 +64,6 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
         parents.sort_unstable();
         parents.dedup();
 
-        let watcher = debouncer.watcher();
         // Watch all configuration file directories.
         for parent in &parents {
             if let Err(err) = watcher.watch(parent, RecursiveMode::NonRecursive) {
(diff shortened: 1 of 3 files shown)


alacritty/src/config/monitor.rs
```diff
@@ -62,32 +64,68 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
...
             if let Err(err) = watcher.watch(parent, RecursiveMode::NonRecursive) {
                 debug!("Unable to watch config directory {:?}: {}", parent, err);
             }
         }
 
+        // The current debouncing time.
+        let mut debouncing_time = None;
+
+        // The events accumulated during the debounce period.
+        let mut received_events = Vec::with_capacity(8);
```
chrisduerr, line 78: Is there any point in initializing this to a size of 8? Seems somewhat arbitrary and likely won't make any difference in practice?

alacritty/src/config/monitor.rs
```diff
@@ -62,32 +64,68 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
...
+            // We use `recv_timeout` to debounce the events comming from the watcher and reduce
+            // the amount of config reloads.
+            let event = if let Some(debouncing_time) = debouncing_time.as_mut() {
+                let time_waited = Instant::now();
+                let event = rx.recv_timeout(*debouncing_time);
+                // Adjust the debouncing time, so we debounce at least each `DEBOUNCE_DELAY`.
+                *debouncing_time = debouncing_time.saturating_sub(time_waited.elapsed());
+                event
+            } else {
+                debouncing_time = Some(DEBOUNCE_DELAY);
```
chrisduerr, line 90: Wouldn't it be more accurate to just store `Instant::now() + DEBOUNCE_DELAY` here and then pass that to `recv_timeout` rather than subtracting from the debounce time? The latter will definitely not be accurate, but I feel like the former is also going to be simpler code?

alacritty/src/config/monitor.rs
```diff
@@ -62,32 +64,68 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
...
+                rx.recv().map_err(Into::into)
+            };
+
+            match event {
+                Ok(Ok(event)) => {
+                    // Receiving any event should reset debouncing.
```
chrisduerr, line 96: It shouldn't and you're not actually resetting debouncing here? So the comment seems wrong?

chrisduerr (changes requested):

alacritty/src/config/monitor.rs
kchibisov: I just picked a value that will avoid any allocation.

alacritty/src/config/monitor.rs
kchibisov: The `recv_timeout` only accept `Duration` unfortunately, there's a nightly API that accepts `Instant` that will simplify the code, but we can't use it. We subtract to block at max of `DEBOUNCE_DELAY`.

alacritty/src/config/monitor.rs
kchibisov: Yes, you're correct. I was resetting in the first impl, but decided that I should change that to not do so, to prevent infinite wait if you update config each `9ms` for example.

alacritty/src/config/monitor.rs
chrisduerr: If anything it would avoid re-allocation, right? And the value is arbitrary, isn't it? Like it could easily be bigger than 8.

alacritty/src/config/monitor.rs
chrisduerr: Wouldn't it still be easier to just sub now from the deadline rather than recalculating the elapsed time every time? That just seems less efficient.

alacritty/src/config/monitor.rs
kchibisov: The amount of events is `4` at max from what I've tested. Since you could have write/rename things. You can print the amount of events we're getting on writing to the file or replacing the file.

alacritty/src/config/monitor.rs
kchibisov: But how do you know what to sub? You need to sub the amount you've blocked, and the amount you've blocked is the elapsed time.

alacritty/src/config/monitor.rs
chrisduerr: That number seems arbitrary though? It will be completely different on other platforms.

alacritty/src/config/monitor.rs
chrisduerr: All you need to do is `deadline - now` to get the duration until an instant.

kchibisov: But the `now` could be in the future and thus you panic?

alacritty/src/config/monitor.rs
```diff
@@ -62,32 +64,63 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
...
+        let mut debouncing_deadline: Option<Instant> = None;
+
+        // The events accumulated during the debounce period.
+        let mut received_events = Vec::new();
+
         loop {
-            let events = match rx.recv() {
-                Ok(Ok(events)) => events,
+            // We use `recv_timeout` to debounce the events comming from the watcher and reduce
+            // the amount of config reloads.
```
chrisduerr, line 82:
```suggestion
            // We use `recv_timeout` to debounce the events coming from the watcher and reduce
            // the amount of config reloads.
```

alacritty/src/config/monitor.rs
```diff
@@ -62,32 +64,63 @@ pub fn watch(mut paths: Vec<PathBuf>, event_proxy: EventLoopProxy<Event>) {
...
+            let event = if let Some(debouncing_deadline) = debouncing_deadline {
```
chrisduerr, line 83: Should use match here.

chrisduerr (changes requested):

chrisduerr (approved):

PR #6660, opened 2023-01-30
Update winit to 0.28

### LINKED ISSUE #3187: Properly inform compositors of opaque regions
### System

OS: Linux
Version: alacritty 0.4.0
Linux/BSD: wayland

### Logs

N/A

---

For efficiency reasons alacritty should inform compositors about opaque portions of the window when alacritty is not set to use partial transparency. This allows compositors to avoid composition for the areas of the screen that are opaque.

On wayland this can be achieved with `wl_surface::set_opaque_region` protocol call.

### LINKED ISSUE #6515: Window clipping when maximized and `decorations` set to `none` on Windows
image

### LINKED ISSUE #6558: Support 'decorations_theme_variant' option on macOS?
Hello Alacritty team,

I prefer my desktop environment to use a light desktop theme, but  I much prefer a dark terminal.

On Linux the `window: decorations_theme_variant` allows one to override the desktop light/dark setting, I set it to dark which looks excellent since the title bar will now be dark blending perfectly with the dark terminal itself.

On macOS Alacritty uses the overall desktop setting (light or dark) to render the window title bar.
Is is possible to override that with a `window: decorations_theme_variant` approach?
Or does Alacritty take no part in rendering the title bar?
Basically is there a technical reason why `window: decorations_theme_variant` only applies to Linux GTK/Wayland and not in addition to macOS and Windows?

iTerm2 does provide a GUI option that mimics the Alacritty `window: decorations_theme_variant` approach FYIW.

Not the biggest or most important issue I know.
If possible to do, this would be excellent.
If impossible to do, then close this issue.

Best regards.

### LINKED ISSUE #6644: IME purpose should be set to terminal
The on-screen keyboard GNOME opens when using Alacritty lacks modifier keys. That's because Alacritty does not specify the input is for a terminal. In wayland, this can be done by [set_content_type](https://docs.rs/wayland-protocols/latest/wayland_protocols/wp/text_input/zv3/client/zwp_text_input_v3/struct.ZwpTextInputV3.html#method.set_content_type) to [ContentPurpose::Terminal](https://docs.rs/wayland-protocols/latest/wayland_protocols/wp/text_input/zv3/client/zwp_text_input_v3/enum.ContentPurpose.html#variant.Terminal).

The difficulty here is that ZwpTextInputV3 is handled by winit (in https://github.com/rust-windowing/winit/blob/master/src/platform_impl/linux/wayland/seat/text_input/mod.rs), so winit needs to add an IME purpose too (not specific to Wayland). I think this would be reasonable (although support in the other platform impls would be limited, macos at least [supports some](https://developer.apple.com/documentation/appkit/nstextcontenttype)). Alternatively, Alacritty would have to do IME handling manually on Wayland.

kchibisov:
Fixes #6644.
Fixes #6558.
Fixes #6515.
Fixes #3187.

### DIFF
--- a/alacritty/src/config/window.rs
+++ b/alacritty/src/config/window.rs
@@ -49,7 +49,7 @@ pub struct WindowConfig {
     /// Background opacity from 0.0 to 1.0.
     pub opacity: Percentage,
 
-    // XXX This wants SerdeReplace...
+    /// Controls which `Option` key should be treated as `Alt`.
     #[cfg(target_os = "macos")]
     pub option_as_alt: OptionAsAlt,
 
--- a/alacritty/src/config/window.rs
+++ b/alacritty/src/config/window.rs
@@ -6,6 +6,9 @@ use serde::de::{self, MapAccess, Visitor};
 use serde::{Deserialize, Deserializer, Serialize};
 use winit::window::{Fullscreen, Theme};
 
+#[cfg(target_os = "macos")]
+use winit::platform::macos::OptionAsAlt;
+
 use alacritty_config_derive::{ConfigDeserialize, SerdeReplace};
 use alacritty_terminal::config::{Percentage, LOG_TARGET_CONFIG};
 use alacritty_terminal::index::Column;
@@ -46,6 +49,10 @@ pub struct WindowConfig {
     /// Background opacity from 0.0 to 1.0.
     pub opacity: Percentage,
 
+    // XXX This wants SerdeReplace...
+    #[cfg(target_os = "macos")]
+    pub option_as_alt: OptionAsAlt,
+
     /// Pixel padding.
     padding: Delta<u8>,
 
@@ -67,6 +74,8 @@ impl Default for WindowConfig {
             opacity: Default::default(),
             padding: Default::default(),
             dimensions: Default::default(),
+            #[cfg(target_os = "macos")]
+            option_as_alt: Default::default(),
         }
     }
 }
--- a/alacritty/src/config/window.rs
+++ b/alacritty/src/config/window.rs
@@ -4,7 +4,7 @@ use std::os::raw::c_ulong;
 use log::{error, warn};
 use serde::de::{self, MapAccess, Visitor};
 use serde::{Deserialize, Deserializer, Serialize};
-use winit::window::Fullscreen;
+use winit::window::{Fullscreen, Theme};
 
 use alacritty_config_derive::{ConfigDeserialize, SerdeReplace};
 use alacritty_terminal::config::{Percentage, LOG_TARGET_CONFIG};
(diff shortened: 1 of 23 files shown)


alacritty/src/config/window.rs
```diff
@@ -111,13 +111,15 @@ impl WindowConfig {
...
-    pub fn decorations_theme_variant(&self) -> Option<&str> {
-        self.gtk_theme_variant
-            .as_ref()
-            .or(self.decorations_theme_variant.as_ref())
-            .map(|theme| theme.as_str())
+    pub fn decorations_theme_variant(&self) -> Option<Theme> {
+        self.gtk_theme_variant.as_ref().or(self.decorations_theme_variant.as_ref()).and_then(
+            |theme| match theme.to_lowercase().as_str() {
+                "dark" => Some(Theme::Dark),
+                "light" => Some(Theme::Light),
```
chrisduerr, line 119: I'd assume `Theme` does not implement serialize/deserialize already? Would be easier to just use that in our config.

alacritty/src/display/window.rs
```diff
@@ -199,7 +197,15 @@ impl Window {
             #[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
             wayland_surface,
             scale_factor,
-        })
+        };
+
+        // Enable IME.
+        window.set_ime_allowed(true);
```
chrisduerr, line 203: Why was this moved?

alacritty/src/display/window.rs
```diff
@@ -404,6 +404,9 @@ impl Window {
 
     pub fn set_ime_allowed(&self, allowed: bool) {
         self.window.set_ime_allowed(allowed);
+        if allowed {
+            self.window.set_ime_purpose(ImePurpose::Terminal);
```
chrisduerr, line 408: Does this have to be reset every time? Or can we just set it once at startup?

chrisduerr (approved):

alacritty/src/config/window.rs
kchibisov: We might want to change it from string to enum?

alacritty/src/display/window.rs
kchibisov: to set both enable and ime purpose

alacritty/src/config/window.rs
chrisduerr: Not sure what you mean.

alacritty/src/display/window.rs
kchibisov: I think it persists internally in winit.

chrisduerr: Would be nice to just set it at startup once.

alacritty/src/config/window.rs
kchibisov: I've updated to use the `enum` instead.

alacritty/src/config/window.rs
```diff
@@ -52,6 +49,10 @@ pub struct WindowConfig {
     /// Background opacity from 0.0 to 1.0.
     pub opacity: Percentage,
 
+    // XXX This wants SerdeReplace...
+    #[cfg(target_os = "macos")]
+    pub option_as_alt: OptionAsAlt,
+
```
kchibisov, line 55: @chrisduerr, Could you help to resolve this one? The struct wants `SerdeReplace`. The type has all sorts of traits implemented already including serde's serialize/deserialize... Right now what I see is to wrap, but I don't want to wrap....

chrisduerr: The thing is that `SerdeReplace` needs to be implemented for this enum to work, so we get the `replace` function.

Since we do not own it, we obviously can't use `ConfigDeserialize`, so what you should do is add it to `impl_replace` in `alacritty_config/src/lib.rs`. Should be pretty simple, but probably easier for you with access to a mac.

kchibisov: I'm just not happy that other crates started depending on winit.

chrisduerr: You mean `alacritty_config`? It also depends on log that's just kinda the nature of the crate.

chrisduerr (approved):

Issue #2572, opened 2019-06-21
Ability to choose fall-back font


rupor-github:
Windows 10

Due to switch to DirectWrite information about fallback font (font linking, windows registry) is ignored. Quoting Microsoft:

"DirectWrite has fallback data that is not read from the registry or in any way configurable. In Windows 8.1, though, APIs have been introduced that allow an app to specify its own fallback. (It's similar to the WPF APIs for creating a composite font definition.)"

Is there a way to specify fallback font explicitly? If not - it will be very helpful to have, it is not always possible to use nerd font patch...

chrisduerr:
This is largely addressed by #957 and mainly just waiting for a reasonable implementation. However other issues have had priority for now, so I can't make any guarantees about when an implementation is going to land.

PR #6663, opened 2023-02-01
Optional fallback font support natively in Alacritty (ie. support NerdFont symbols with any font)

sharkusk:
Patching fonts with extra symbols can be a pain (unless a pre-patched NerdFont is already available).  Even then, the patched font is not exactly the same as the original.  OS level fallback font configuration may be possible with Linux, but is not viable with MacOS and difficult to figure out with Windows.  Adding support for an optional fallback font directly in Alacritty makes life easier for those of us who use shiny n/vim/tmux configurations.  All that is needed is a single installed NerdFont specified as the fallback, and all of the NF symbols become available across all system fonts.

A couple notes on these changes:
1. No effort has been made to match the top/bottom offsets of the symbols used for power line symbols.  This is not needed since native support for those will be included in Alacritty going forward.
2. I tried to keep the changes minimal and straightforward.  This is my first time with Rust, so I'm guessing there may be a more elegant way to handle the main logic.  Feedback is very much appreciated.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -155,7 +155,14 @@
     #family: monospace
 
     # The `style` can be specified to pick a specific face.
-    #style: Bold Italic
+# 
+  # Fallback font used for glphs that cannot be found in normal font
+  #fallback:
+    # Font family
+    #
+    # If the fallback family is not specified, it will fall back to the
+    # value specified for the normal font.
+    #family: Hasklug Nerd Font
 
   # Point size
   #size: 11.0
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -155,7 +155,8 @@
     #family: monospace
 
     # The `style` can be specified to pick a specific face.
-# 
+    #style: Bold Italic
+ 
   # Fallback font used for glphs that cannot be found in normal font
   #fallback:
     # Font family
--- a/alacritty/src/config/font.rs
+++ b/alacritty/src/config/font.rs
@@ -37,6 +37,9 @@ pub struct Font {
     /// Bold italic font face.
     bold_italic: SecondaryFontDescription,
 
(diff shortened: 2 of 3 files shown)


chrisduerr (changes requested):
Yeah this implementation is completely useless. Having font fallback **after** all the other fonts will not provide a useful result. If you wanted to actually implement this feature, you'd have to fallback **before** the system's font fallback.

sharkusk:
Hmm. Did you actually try it before calling it useless?

Based on your encouraging feedback I'm guessing you want a pre-font that is checked before the main font (along the lines of the line draw feature.) IMHO, this seems like a bad approach for a fallback. The fallback font should be used as the last resort, when all other possibilities have been checked. (Thus reason it's called "fallback".)

For a pre-font approach you can use a bunch of range checking in the config (which I considered), but that seems to be overkill for what most users need, a bit slow since nerd fonts have a lot of small ranges, and error prone when new ranges and symbols are added over time.

chrisduerr:
This approach will destroy itself on the average system with 6000 fonts installed. The issue is never not having any fallback, because we have font fallback built in and if there was actually a font with that glyph we would just pick it. The issue is with font fallback not falling back according to the user's preference.

sharkusk:
Thanks for getting back to me.  I agree this feature does not allow a user to pick and chose a glyph override from each of their 6,000 installed fonts (aka "symbol_map" in kitty).

As mentioned in my description, both MacOS and Windows do not have a good mechanism to allow users to specify fallback fonts.  This feature gives those users an easy way to add a fallback font when using Alacritty.

There may be users other than myself (#2572 for example) that would find this more limited approach useful while waiting for #957.

PR #6664, opened 2023-02-02
Fix: Set IME purpose before enable

lukaslihotzki:
@kchibisov, thanks for releasing winit and updating alacritty.

As I understood, setting the IME purpose after enabling may have no effect, so simply do it before enabling.

### DIFF
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -172,8 +172,8 @@ impl Window {
         window.set_cursor_icon(current_mouse_cursor);
 
         // Enable IME.
-        window.set_ime_allowed(true);
         window.set_ime_purpose(ImePurpose::Terminal);
+        window.set_ime_allowed(true);
 
         // Set initial transparency hint.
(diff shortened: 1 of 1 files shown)


kchibisov:
No it doesn't matter when you enable it. It's an atomic change to the current state. It only works with the enable commit or after. if it doesn't it's a bug of your IME.

kchibisov:
Or do you mean that some IMEs 

>Subsequent attempts to update them may have no effect. The values remain valid until the next committed enable or disable request.

Treat implicit setting of this as content purpose change, so they don't bother changing?

lukaslihotzki:
Yes, I did this PR because of the sentence you quoted, so the enable request is after the set_content_type.

kchibisov:
And you know that your IME breaks because of this?

kchibisov:
I'd keep it that way, since it's cleaner. I might change the entire API in winit wrt IME anyway, to account for atomic updates, but unlikely soon.

PR #6665, opened 2023-02-02
Update ndk-sys to workaround cargo checksum issue

oxalica:
This causes sporadic errors when fetching dependencies, especially in clean sandboxed build like Nix.
See https://github.com/rust-lang/cargo/issues/11412

> The package [ndk-sys](https://crates.io/crates/ndk-sys) has a latest version 0.4.0 and a more recent yanked version 0.4.0+25.0.8775105. Cargo is downloading the 0.4.0 crate, but comparing the hash with the checksum of the yanked version, and therefore failing with a "failed to verify checksum of ndk-sys v0.4.0" error.


kchibisov (approved):
I think I've seen this one on nix as well.

PR #6673, opened 2023-02-04
Switch to TOML configuration format

### LINKED ISSUE #6592: Switch to TOML for configuration file
I'm quite honestly tired of looking at YAML files with broken indentation. While YAML is an awful format even for experienced users, clearly it's completely impossible for inexperienced ones to grasp its concept. And the documentation of our configuration file with everything commented out by default probably doesn't help.

Since as far as I know all previous issues with TOML have been fixed and escape sequences can be easily specified in TOML strings now, I think it would make sense to switch to TOML as our default format.

In a transitory period we could probably get away with just loading existing YAML files and writing them out as TOML, to allow for backward compatibility. This will probably produce somewhat messy TOML files, but I can't imagine an auto-generated TOML file would be much worse than a YAML file.

The only major drawbacks I see is that this is a rather big change that will no doubt annoy a bunch of people and some more advanced features of YAML (like anchors) are not supported in TOML. But nobody *really* needs anchors and I kinda like annoying people.

@kchibisov You proposed this a while ago on IRC, do you think this is a bad idea?

See https://github.com/alacritty/alacritty/issues/1217 for prior discussions.

chrisduerr:
This switches Alacritty's default configuration format from yaml to toml. While yaml is still supported, it is done by converting it to toml and should be removed entirely in the future.

All existing features were persisted based on my testing. Behavior should not change much, though `--option` might have slightly different behavior since the entire line is not interpreted as one line of toml.

TODO:
 - [x] Migrate alacritty.yml => Might want to reconsider defaults being commented out?
 - [X] Write/Test an `alacritty migrate` tool which converts the configuration file for the user
 - [x] Warn on unused keys
 - [x] Find solution to `key_bindings` at the top requirement

Closes #6592.


chrisduerr:
While I do intend to finish this up with the remaining todos in the near future, I'm curious about what you think about this approach @kchibisov. For someone who doesn't want an automatic migration, all the code should be done already and behavior doesn't need any changes afaik.

alacritty/src/config/mod.rs
```diff
@@ -186,18 +214,17 @@ fn parse_config(
         contents = contents.split_off(3);
     }
 
+    // Convert YAML to TOML as a transitionary fallback mechanism.
+    let extension = path.extension().unwrap_or_default();
+    if (extension == "yaml" || extension == "yml") && !contents.trim().is_empty() {
```
kchibisov, line 219: From what I remember we don't allow `yaml` extension, just `yml`.

```suggestion
    if (extension == "yml") && !contents.trim().is_empty() {
```

alacritty/src/config/bindings.rs
```diff
@@ -1010,19 +1010,20 @@ impl<'a> Deserialize<'a> for RawBinding {
...
+                            match value.as_integer() {
+                                Some(scancode) => {
+                                    if scancode > i64::from(u32::MAX) {
+                                        return Err(<V::Error as Error>::custom(format!(
+                                            "Invalid key binding, scancode too big: {}",
+                                            scancode
+                                        )));
+                                    }
+                                    key = Some(Key::Scancode(scancode as u32));
+                                },
```
kchibisov, line 1023: Maybe do explicit check for negative numbers?

alacritty.toml
```diff
@@ -0,0 +1,905 @@
+# Configuration for Alacritty, the GPU enhanced terminal emulator.
```
kchibisov, line 1: That's not a `toml`.

kchibisov (commented):
I can't say much, given that the code does very little useful. Most of it is simply conversion from `yaml` to `toml`, which is simply replace `serde_yaml` with `toml`.

From what I can see the `migrate` thing isn't written yet, and the entire config file is still `yaml`. If you plan on doing `migrate` I'd suggest to convert it file by file separately without `import` chaining. I'd also suggest to output to user provided path, instead of overwriting files. Maybe add a suggestion on how to migrate it.

chrisduerr:
>  If you plan on doing migrate I'd suggest to convert it file by file separately without import chaining. I'd also suggest to output to user provided path, instead of overwriting files. Maybe add a suggestion on how to migrate it.

Since `.toml` and `.yml` aren't in conflict, my plan was just to go through every single configuration file, parse it with yaml, then write the toml to `alacritty.toml` in the same location the `.yml` was in. I think that should work pretty well and should be completely non-destructive. Then we can just ask people to delete their yaml files.

This also gives me an idea though: I kinda gotten used to loading the old stuff first, but I don't think we should be doing that. Right now yaml is loaded first and toml tried second. We should swap that around. For config values it makes sense to check the old one first because of the way our config struct is built, but for configuration files it should be trivial to just always pick the best available format.

I'm not entirely sure if I want to rewrite import paths to switch them from `yml` to `toml` since that requires changes in two places, but I think there shouldn't be much risk for failure.

alacritty/src/config/mod.rs
chrisduerr: Correct, but I intentionally accept both here since `--config-file` is a thing which some people might use in their weird setups. It costs nothing and should be a little safer.

Of course people could also do `--config-file /tmp/whatever`, but if you do that without a file suffix then fuck you, you deserve the error.

alacritty.toml
chrisduerr: Shhh, don't tell anyone.

kchibisov:
>Since .toml and .yml aren't in conflict, my plan was just to go through every single configuration file, parse it with yaml, then write the toml to alacritty.toml in the same location the .yml was in. I think that should work pretty well and should be completely non-destructive. Then we can just ask people to delete their yaml files.

The location might be RO.

>This also gives me an idea though: I kinda gotten used to loading the old stuff first, but I don't think we should be doing that. Right now yaml is loaded first and toml tried second. We should swap that around. For config values it makes sense to check the old one first because of the way our config struct is built, but for configuration files it should be trivial to just always pick the best available format.

Yeah, that's expected to try new first and then fallback to the old one.

>I'm not entirely sure if I want to rewrite import paths to switch them from yml to toml since that requires changes in two places, but I think there shouldn't be much risk for failure.

Nah, I think we should provide a manual way to convert all of them, writing might not work in the first place. Though you could provide option with output and option with automatic writing.

alacritty/src/config/bindings.rs
chrisduerr: We need to check for both negative numbers **and** positive numbers that are too big though. But there's no need why we would convert twice, so we can just use `u32::try_from` instead.

chrisduerr:
> The location might be RO.

In which case we just print an error and exit the `migrate` tool. I'd assume this might happen with some shared config setups in like `/etc` or whatever, but it shouldn't be common enough to where we somehow need to explicitly support it.

If you have a system admin config and you run it without privileges we just fail and you can rerun it with the privileged user.

> Nah, I think we should provide a manual way to convert all of them, writing might not work in the first place. Though you could provide option with output and option with automatic writing.

How likely is it that writing won't work though? I'd assume you're thinking about your weird ass nixOS system?

But having an output-only option seems sensible.

kchibisov:
>How likely is it that writing won't work though? I'd assume you're thinking about your weird ass nixOS system?

Yeah, your config is a symlink to RO filesystem, so you can't even write to it with 'root'.

chrisduerr:
Testing that @perfbot is back online.

fee1-dead:
> This switches Alacritty's default configuration format from toml to yaml

from toml to yaml? this should preferably be changed

chrisduerr:
@perfbot test

WuerfelDev:
Regarding configuration files, I'd love to have some fallback keys (maybe with `keyname_fallback = "value"` since toml doesn't allow duplicate keys). Maybe that's something we could integrate?
As example: currently I want to use `window: {decorations: buttonless}` on mac but when I use my config on linux I get an error. If there was a fallback I wouldn't have to maintain two versions of the same config file

amtoine:
i like this idea :yum: 

what is the status of the PR? :open_mouth:

chrisduerr:
> what is the status of the PR?

I don't think there's any rush to migrate. Currently the only thing somewhat uncertain is documentation I suppose.

amtoine:
> > what is the status of the PR?
> 
> I don't think there's any rush to migrate. Currently the only thing somewhat uncertain is documentation I suppose.
>
oh yeah of course, just curious to see a new format :yum:

docs/features.md
```diff
@@ -77,7 +77,7 @@ file. If an application captures your mouse clicks, which is indicated by a
 change in mouse cursor shape, you're required to hold <kbd>Shift</kbd> to bypass
 that.
 
-[configuration file]: ../alacritty.yml
+[configuration file]: ../alacritty.toml
```
kchibisov, line 80: I don't think we have it.

extra/man/alacritty-bindings.5.scd
```diff
@@ -0,0 +1,478 @@
...
+:[
+:  _"Copy"_
+|  _"L"_
+:  _"Control"_
+:[
+:  _"ClearLogNotice"_
+|  _"L"_
+:  _"Control"_
+:  _"~Vi|~Search"_
+:  *chars*: _"\x0c"_
```
kchibisov, line 48: I'd assume you've ensured that `\x` bindings do work? I remember they were not a part of initial spec.

extra/man/alacritty.5.scd
```diff
@@ -0,0 +1,800 @@
...
+
+	Allowed values are hexadecimal colors like _#ff00ff_, or
+	_CellForeground_/_CellBackground_, which references the affected cell.
+
+	Default: _{ text = "CellBackground", background = "CellForeground" }_
+
+*normal*
+
+	*black* <string>
+		Default: _"#1d1f21"_
```
kchibisov, line 336: I see that you want me to port this as well...

alacritty/src/migrate.rs
```diff
@@ -0,0 +1,130 @@
...
+use std::fs;
+use std::path::Path;
+
+use toml::Value;
+
+use crate::cli::MigrateOptions;
+use crate::config;
+
+/// Handle migration.
+pub fn migrate(options: MigrateOptions) {
```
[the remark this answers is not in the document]
chrisduerr: Seems like the easiest solution to me?

docs/features.md
chrisduerr: It wasn't actually a link anyway.

alacritty/src/migrate.rs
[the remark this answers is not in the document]
kchibisov: I mean, it's a bit weird that we have normal error handling in other sub command and have lazy `eprintln` + `exit` here.

extra/man/alacritty.5.scd
chrisduerr: :)

extra/man/alacritty-bindings.5.scd
chrisduerr: They do not work… yet.

alacritty/src/config/mod.rs
```diff
@@ -179,58 +208,101 @@ fn parse_config(
...
+                error!(target: LOG_TARGET_CONFIG, "Unable to import config {:?}: {}", path, err)
+            },
+        }
+    }
+
+    merged
+}
+
+// TODO: Merge back with `load_imports` once `alacritty migrate` is dropped.
+//
```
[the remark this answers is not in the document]
kchibisov: That `//` comments flow into `///` comments.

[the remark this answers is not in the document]
chrisduerr: I see no issue with it tbh.

ngortheone:
I like this idea. 
What are the technical reasons for the switch?
Is yaml parsing too slow? Yaml dependencies too bloated? ...

Or simply out of pure hatred of yaml?
https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell

kchibisov:
folks can't write their configs with yaml because of indention.

kchibisov:
also, we could warn on unused fields now.

chrisduerr:
> Or simply out of pure hatred of yaml?

Yes.

chrisduerr:
> also, we could warn on unused fields now.

We not only could, we do. It is implemented already in this PR.

kchibisov (changes requested):
I think now I went through all the thing.

extra/man/alacritty.5.scd
```diff
@@ -0,0 +1,800 @@
...
+
+		If the bright foreground color is not set, it is automatically
+		calculated based on the foreground color.
+
+		Default: _"#828482"_
+
+	*bright_foreground* <string>
+
+		If the bright foreground color is not set, or
+		_draw\_bold\_text\_with\_bright\_colors_ is _false_, the normal
```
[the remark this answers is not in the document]
chrisduerr: I'm not sure what you mean by this.

extra/man/alacritty.5.scd
```diff
@@ -0,0 +1,800 @@
...
+
+		Default: _"#1d1f21"_
+
+	*background* <string>
+
+		Default: _"#c5c8c6"_
+
+	*dim_foreground* <string>
+
+		If the bright foreground color is not set, it is automatically
```
kchibisov, line 242:
```suggestion
		If the dim foreground color is not set, it is automatically
```

I think it's `dim`?

extra/man/alacritty.5.scd
```diff
@@ -0,0 +1,800 @@
...
+
+		Default: _"#828482"_
+
+	*bright_foreground* <string>
+
+		If the bright foreground color is not set, or
+		_draw\_bold\_text\_with\_bright\_colors_ is _false_, the normal
+		foreground color will be used.
+
+		Default: _"#eaeaea"_
```
kchibisov, line 253:
```suggestion
		Default: _None_
```

We don't set it, meaning that we don't use it, unlike dim which is used, but auto calculated.

extra/man/alacritty.5.scd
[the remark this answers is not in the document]
kchibisov: I wanted something like:

```
This will be used only when it's set and _draw_bold_text_with_bright_colors is true.
```

Or something similar.

INSTALL.md
```diff
@@ -333,6 +333,8 @@ Installing the manual page requires the additional dependencies `gzip` and `scdo
 sudo mkdir -p /usr/local/share/man/man1
 scdoc < extra/man/alacritty.1.scd | gzip -c | sudo tee /usr/local/share/man/man1/alacritty.1.gz > /dev/null
 scdoc < extra/man/alacritty-msg.1.scd | gzip -c | sudo tee /usr/local/share/man/man1/alacritty-msg.1.gz > /dev/null
+scdoc < extra/man/alacritty.5.scd | gzip -c | sudo tee /usr/local/share/man/man5/alacritty.5.gz > /dev/null
+scdoc < extra/man/alacritty-bindings.5.scd | gzip -c | sudo tee /usr/local/share/man/man5/alacritty-bindings.5.gz > /dev/null
```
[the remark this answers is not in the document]
chrisduerr: I don't think it's relevant tbh.

extra/man/alacritty.5.scd
kchibisov: Also, the default in the config is `None`, when you look into color.rs, so it could be changed either here or in the configuration to specify the value explicitly. It's also a bit weird that we have `Dim` colors defined in the config, but not use `None`.

it's also weird that we have `Dim` colors defined in the `color.rs`, but they actually never used, but still being computed on demand.

alacritty/src/cli.rs
```diff
@@ -30,17 +28,18 @@ pub struct Options {
...
     pub config_file: Option<PathBuf>,
 
-    /// Specify alternative configuration file [default: %APPDATA%\alacritty\alacritty.yml].
+    /// Specify alternative configuration file [default: %APPDATA%\alacritty\alacritty.toml].
     #[cfg(windows)]
     #[clap(long, value_hint = ValueHint::FilePath)]
     pub config_file: Option<PathBuf>,
 
-    /// Specify alternative configuration file [default: $HOME/.config/alacritty/alacritty.yml].
+    /// Specify alternative configuration file [default: $HOME/.config/alacritty/alacritty.toml].
```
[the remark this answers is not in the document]
chrisduerr: I don't think it's relevant for macOS users though? They basically never have it set.

docs/features.md
```diff
@@ -77,8 +77,6 @@ file. If an application captures your mouse clicks, which is indicated by a
 change in mouse cursor shape, you're required to hold <kbd>Shift</kbd> to bypass
 that.
 
-[configuration file]: ../alacritty.yml
```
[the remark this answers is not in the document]
chrisduerr: Oh I thought this was never used because I couldn't find it. I didn't think it would be broken across lines.

alacritty_config_derive/src/config_deserialize/de_struct.rs
```diff
@@ -42,20 +42,25 @@ pub fn derive_deserialize<T>(
...
+                            unused.insert(key, value);
                         },
                     }
                 }
 
                 #flatten
 
+                // Warn about unused keys.
+                for key in unused.keys() {
+                    log::warn!(target: #LOG_TARGET, "Unused config key: {}", key);
```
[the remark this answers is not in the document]
chrisduerr: Add it where?

alacritty/src/migrate.rs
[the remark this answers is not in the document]
chrisduerr: I want to keep it this way because it gives us more control over the error messages.

We're not just trying to print errors and bail here, but instead we're trying to have nice human-readable CLI output.

That's why I did it this way, letting Rust's `?` handle error printing is usually worse.

alacritty/src/config/ui_config.rs
```diff
@@ -90,29 +97,36 @@ impl Default for UiConfig {
...
 }
 
 impl UiConfig {
     /// Generate key bindings for all keyboard hints.
     pub fn generate_hint_bindings(&mut self) {
+        let key_bindings = if self.keyboard.bindings.0.len() >= self.key_bindings.0.len() {
+            &mut self.keyboard.bindings.0
+        } else {
+            &mut self.key_bindings.0
+        };
```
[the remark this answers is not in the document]
chrisduerr: It's because of our default bindings.

alacritty/src/migrate.rs
```diff
@@ -0,0 +1,130 @@
...
+    // Migrate config imports.
+    if !options.skip_imports {
+        migrate_imports(options, &mut config, recursion_limit)?;
+    }
+
+    // Convert to TOML format.
+    let toml = toml::to_string(&config).map_err(|err| format!("conversion error: {err}"))?;
+    let new_path = format!("{prefix}.toml");
+
+    if options.dry_run && !options.silent {
```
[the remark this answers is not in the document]
chrisduerr: > Maybe we should move deprecated fields into the right sections automatically?

I'm looking into this.

> Also, each keybinding resulted in separate [[key_bindings]] after conversion, though, not sure what we should do about that.

I don't think there's any way we'll be able to format the keybindings nicely.

extra/man/alacritty.5.scd
chrisduerr: I believe we've intentionally specified colors here so people have an easier time copying Alacritty's colorscheme.

Your suggestion is technically correct, but I think in this very specific scenario it's better to be a little bit wrong.

alacritty/src/cli.rs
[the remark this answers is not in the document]
kchibisov: That's true, but it doesn't deny a fact that it's used as first.

extra/man/alacritty.5.scd
kchibisov: The thing is that it's not computed and such color is not used, so there's no such color and the value written here is arbitrary.

For dim colors it's correct, but not here.

alacritty_config_derive/src/config_deserialize/de_struct.rs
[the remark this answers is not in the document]
kchibisov: In the `CHANGELOG`.

kchibisov (changes requested):

alacritty/src/migrate.rs
```diff
@@ -0,0 +1,215 @@
...
+        Ok(true)
+    } else {
+        let next_table_value = table.entry(path[0]).or_insert_with(|| Value::Table(Table::new()));
+
+        let next_table = match next_table_value.as_table_mut() {
+            Some(next_table) => next_table,
+            None => return Err(format!("invalid `{}` table", path[0])),
+        };
+
+        insert_node_if_empty(next_table, &path[1..], node)
```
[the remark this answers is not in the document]
chrisduerr: Why? Doing it recursively is significantly cleaner imo and there's no chance this will bust the stack.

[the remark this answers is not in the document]
kchibisov: it's just you have tail recursion, so it's already sort of iteration, but it's up to you. If it was the library code I'd force iteration as it's efficient, but here it's probably fine.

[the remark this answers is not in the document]
chrisduerr: Yeah the **sole** reason for this is to make the code cleaner, which I think in this scenario should have priority.

A lot of times recursion can be confusing, but this is one of those things that just maps naturally to it imo.

alacritty/src/migrate.rs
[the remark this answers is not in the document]
chrisduerr: I don't think there's a way to format them nicely automatically.

extra/man/alacritty.5.scd
```diff
@@ -0,0 +1,823 @@
...
+
+	Allow terminal applications to change Alacritty's window title.
+
+	Default: _true_
+
+*class* { instance = <string>, general = <string> } _(Linux/BSD only)_
+
+	Window class
+
+	On Wayland, _general_ is used as app_id and _instance_ is ignored.
```
kchibisov, line 125:
```suggestion
	On Wayland, _general_ is used as _app\_id_ and _instance_ is ignored.
```

kchibisov (approved):

amtoine:
i've been subscribed to this for a while and i'm super happy to see it land :star_struck:

.github/workflows/release.yml
```diff
@@ -87,4 +89,3 @@ jobs:
           ./.github/workflows/upload_asset.sh ./extra/completions/_alacritty $GITHUB_TOKEN
           ./.github/workflows/upload_asset.sh ./extra/linux/Alacritty.desktop $GITHUB_TOKEN
           ./.github/workflows/upload_asset.sh ./extra/alacritty.info $GITHUB_TOKEN
-          ./.github/workflows/upload_asset.sh ./alacritty.yml $GITHUB_TOKEN
```
aboqasem, line 90: Hello, my I know if there will be an alternative way to get the default TOML config with comments?

aboqasem: Ok it's already been asked: https://github.com/alacritty/alacritty/issues/6592#issuecomment-1603702638

PR #6675, opened 2023-02-04
Resize the window by cell dimensions

### LINKED ISSUE #388: When resizing, use row/column precision, not pixel precision.

kchibisov:
This should resize window by cell dimensions granularity instead of using pixels.

Fixes #388.

--

Probably will need a config option for that, idk, given that it might not be desirable behavior for everyone.

It works fine on both macOS/GNOME.


chrisduerr:
> Probably will need a config option for that, idk, given that it might not be desirable behavior for everyone.

Let's maybe go without at first and then see if people complain? I'd imagine for floating people they might not even realize anyway and for tiling it doesn't matter.

chrisduerr (approved):

PR #6695, opened 2023-02-12
Add touch input support

### LINKED ISSUE #3671: Add touchscreen support

chrisduerr:
This patch builds upon the prior work by @4z3 and @bytbox to add touchscreen support to Alacritty. While some inspiration was taken from @4z3's patch, it was rewritten from scratch.

This patch supports 4 basic touch interactions:
 - Tap
 - Scroll
 - Select
 - Zoom

Tap allows emulating the mouse to enter a single LMB click. While it would be possible to add more complicated mouse emulation including support for RMB and others, it's likely more confusing than anything else and could conflict with other more useful touch actions.

Scroll and Select are started by horizontal or vertical dragging. While selection isn't particularly accurate with a fat finger, it works reasonably well and the separation from selection through horizontal and vertical start feels pretty natural.

Since horizontal drag is reserved for selection we do not support horizontal scrolling inside the terminal. While it would be possible to somewhat support it by starting a selection with vertical movement and then scrolling horizontally afterwards, it would likely just confuse people so it was left out.

Zoom is pretty simple in just changing the font size when a two-finger pinch gesture is used. Performance of this is pretty terrible especially on low-end hardware since this obviously isn't a cheap operation, but it seems like a worthwhile addition since small touchscreen devices are most likely to need frequent font size adjustment to make output readable.

Closes #3671.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -2,7 +2,7 @@
 
 use std::borrow::Cow;
 use std::cmp::{max, min};
-use std::collections::{HashMap, VecDeque};
+use std::collections::{HashMap, HashSet, VecDeque};
 use std::error::Error;
 use std::ffi::OsStr;
 use std::fmt::Debug;
@@ -19,7 +19,8 @@ use log::{debug, error, info, warn};
 use wayland_client::{Display as WaylandDisplay, EventQueue};
 use winit::dpi::PhysicalSize;
 use winit::event::{
-    ElementState, Event as WinitEvent, Ime, ModifiersState, MouseButton, StartCause, WindowEvent,
+    ElementState, Event as WinitEvent, Ime, ModifiersState, MouseButton, StartCause,
+    Touch as TouchEvent, WindowEvent,
 };
 use winit::event_loop::{
     ControlFlow, DeviceEventFilter, EventLoop, EventLoopProxy, EventLoopWindowTarget,
@@ -65,6 +66,9 @@ const MAX_SEARCH_WHILE_TYPING: Option<usize> = Some(1000);
 
 /// Maximum number of search terms stored in the history.
 const MAX_SEARCH_HISTORY_SIZE: usize = 255;
+
+/// Touch zoom speed.
+const TOUCH_ZOOM_FACTOR: f32 = 0.01;
 
 /// Alacritty events.
 #[derive(Debug, Clone)]
@@ -186,6 +190,7 @@ pub struct ActionContext<'a, N, T> {
     pub terminal: &'a mut Term<T>,
     pub clipboard: &'a mut Clipboard,
     pub mouse: &'a mut Mouse,
+    pub touch: &'a mut Touch,
     pub received_count: &'a mut usize,
     pub suppress_chars: &'a mut bool,
     pub modifiers: &'a mut ModifiersState,
@@ -336,6 +341,11 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     #[inline]
     fn mouse(&self) -> &Mouse {
         self.mouse
+    }
+
+    #[inline]
+    fn touch(&mut self) -> &mut Touch {
+        self.touch
     }
 
     #[inline]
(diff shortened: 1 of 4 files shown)


4z3:
Thanks for pushing this forward! I've tested this patch and it works for me as expected.

4z3:
There is one minor difference from the way this branch works vs. my branch (which is a rebased version of https://github.com/alacritty/alacritty/pull/4346):

When zooming, then it's not possible to keep one finger on the touchscreen while swiping with the other to zoom out or in.  Instead, the zooming gesture has to be done without interruption.

For me that's fine, but the previous behavior is more inline with other touch-enabled stuff.

chrisduerr:
> When zooming, then it's not possible to keep one finger on the touchscreen while sweeping with the other to zoom out or in. Instead, the zooming gesture has to be done without interruption.

Yeah I've intentionally made this change to keep things simpler and at least personally I've never used pinch gestures any other way. I might just be speaking from my personal habit, but either way it should be fairly simple to extend the existing functionality to allow for one finger to be released if people do desire it.

alacritty/src/input.rs
```diff
@@ -735,6 +745,118 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+
+                    // Apply motion since touch start.
+                    self.on_touch_motion(touch);
+                } else if delta_y.abs() > MAX_TAP_DISTANCE {
+                    // Update gesture state.
+                    *touch_purpose = Touch::Scroll(*start);
+
+                    // Apply motion since touch start.
+                    self.on_touch_motion(touch);
+                }
```
kchibisov, line 802: That looks kind of wrong, why you don't compute the actual distance between the points and apply some logic based on that, what's the purpose of separate `x` first and then maybe `y` second handling? Simply to have `Scroll` separate?

Maybe we might use `2` fingers for scrolling, or other terminals scroll with just one finger?

alacritty/src/input.rs
```diff
@@ -735,6 +745,118 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+        let touch_purpose = self.ctx.touch();
+        match touch_purpose {
+            // Simulate LMB clicks.
+            Touch::Tap(start) => {
+                let start_location = start.location;
+                *touch_purpose = Default::default();
+
+                self.mouse_moved(start_location);
+                self.mouse_input(ElementState::Pressed, MouseButton::Left);
+                self.mouse_input(ElementState::Released, MouseButton::Left);
```
kchibisov, line 834: That should consider slots, otherwise you can 3-ple click with 3 fingers? So the tap should be counted from first/last insert or maybe when there's only a single finger at a time.

alacritty/src/event.rs
```diff
@@ -1016,12 +1026,63 @@ impl<'a, N: Notify + 'a, T: EventListener> ActionContext<'a, N, T> {
...
+    Scroll(TouchEvent),
+    Zoom(TouchZoom),
+    Tap(TouchEvent),
+    Invalid(HashSet<u64>),
+}
+
+/// Touch zooming state.
+#[derive(Debug)]
+pub struct TouchZoom {
+    slots: (TouchEvent, TouchEvent),
```
kchibisov, line 1044: Maybe a bit more descriptive than tuple? not sure how to guess what it supposed to do. Are you trying to keep to slots here?

alacritty/src/input.rs
```diff
@@ -735,6 +745,118 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+        }
+    }
+
+    /// Handle end of touch input.
+    pub fn on_touch_end(&mut self, touch: TouchEvent) {
+        // Finalize the touch motion up to the release point.
+        self.on_touch_motion(touch);
+
+        let touch_purpose = self.ctx.touch();
+        match touch_purpose {
```
kchibisov, line 826:
```suggestion
        let touch_purpose = self.ctx.touch();
        *touch_purpose = match touch_purpose {
```

And then simply return from each branch. Should do the same for `on_touch_motion`, since you're already doing so in `on_touch_start`.

kchibisov (changes requested):
Also need a CHANGELOG entry.

alacritty/src/input.rs
4z3: Twofingerscrolling on a touchscreen seems pretty unergonomic. I quickly checked some terminals, most don't respond to touch at all, but at least konsole also scrolls with one finger. And of course browsers scroll with one finger.

alacritty/src/input.rs
chrisduerr: As soon as a third finger is put on the touchscreen the invalid state is entered and you need to release all touch points to start any further touch input.

alacritty/src/input.rs
chrisduerr: Won't work due to borrow-checker issues. You can't hold the touch purpose across `self.mouse_input` since it's a mutable borrow.

kchibisov: But you've already done it before with `std::mem::take`?

chrisduerr: No, mem::take does not circumvent the borrowchecker issues, it just makes it so you don't have to clone anything. You're still holding `touch_purpose` until reassignment mutably.

kchibisov: Then you should change the other instance where you use `mem::take` for that?

kchibisov (approved):
From my testing this works good, except the Zoom handler. Though, I guess the issue is that we're actually doing IO and most time is spent re-matching the font...

thorstenweber83:
Love it!
Just as i moved to a x1 yoga and thought i had to live with kitty's weird selection behavior, alacritty comes along with touch support :partying_face:

PR #6699, opened 2023-02-14
Update INSTALL.md

nixpulvis:
Well, I'm testing out... wait which OS am I on right now...

Anyway, this seems more canonical now, no?

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -84,7 +84,7 @@ to build Alacritty. Here's an apt command that should install all of them. If
 something is still found to be missing, please open an issue.
 
 ```sh
-apt-get install cmake pkg-config libfreetype6-dev libfontconfig1-dev libxcb-xfixes0-dev libxkbcommon-dev python3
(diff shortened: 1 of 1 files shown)


chrisduerr:
Is there *any* reason to actually change this? Doesn't `apt` internally still call `apt-get` anyway?

If this works just fine and there's no sign that it will ever break, I don't see why would change it?

nixpulvis:
I got an error, apparently the system I was on didn't have `apt-get` but did have `apt`. Hard to say which is more popular, but I do think `apt` is pretty ubiquitous.

Let me double check the state of Ubuntu and Debian LTS stuff and report back.

kchibisov (approved):
Legacy systems have just `apt-get`(very old ones). More modern had both, and I think the transition is towards  `apt` anyway. At least on `Debian 11` I have `apt`.

chrisduerr:
> I got an error, apparently the system I was on didn't have apt-get but did have apt

Which system is this?

nixpulvis:
Pretty sure it was just the latest Ubuntu release. Haven't verified that.

PR #6705, opened 2023-02-15
Implement basic smoothing of the cursor movement

GregTheMadMonk:
This is made by simply caching the previous cursor position and interpolating between it and the current position from frame to frame. This means rendering all of the frames, and an increased GPU load (which is comparable to watching a YouTube video), but that's just an optional feature that's off by default. Maybe the possible performance hit should be specified in the config template comments.
Also, turning the feature on switches block cursor to underline, because block cursor is rendered via replacing cell's fore-/background colors and you couldn't really interpolate there, unless there's a bigger change to the code.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -425,6 +425,12 @@
   # Specifying '0' will disable timeout for blinking.
   #blink_timeout: 5
 
+  # Cursor movement smoothing factor
+  # Cursor position each frame is determined via `old_pos + (new_pos - old_pos) / factor`
+  # The more the factor - the slower the cursor movement. 0.0 turns smoothing off
+  # 'Block' cursor shape gets replaced with 'Undeline' if this is on
+  #smooth_factor: 0.0
+
   # If this is `true`, the cursor will be rendered as a hollow box when the
   # window is not focused.
   #unfocused_hollow: true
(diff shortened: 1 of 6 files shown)


chrisduerr (changes requested):
Not even remotely interested in this feature.

GregTheMadMonk:
Rude 😐

chrisduerr:
If you didn't see that coming, that's 100% on you.

It should be obvious to anyone that Alacritty isn't the right project for adding pointless UI options that reduces performance. Doing so after issues specifically requesting this feature have been shot down is quite frankly just a rude waste of my time.

GregTheMadMonk:
You wasted more time insisting on being an asshole after closing the issue than it would take any person to read this entire pull request.

> Then do it yourself lol.

I did. What's your problem now?

merowingen:
This is a great feature that was very well received when it was released in kitty. I know a lot of people who switched to kitty only because of it. But kitty is *nix-only so I've wonder if there is any chances that alacritty one day might have something similar.

TheAjaykrishnanR:
> This is a great feature that was very well received when it was released in kitty. I know a lot of people who switched to kitty only because of it. But kitty is *nix-only so I've wonder if there is any chances that alacritty one day might have something similar.

alacritty wont but [wezterm](https://github.com/wezterm/wezterm/discussions/1136) has a discussion on this and would most likely have an implementation in the future. Also I never understood people already on linux switching to kitty from alacritty, precisely as to why they chose alacritty to begin with.

PR #6710, opened 2023-02-18
Add `window.resize_increments` config option

kchibisov:
Given how bugged the resize increments are on X11, it's better to disable it by default.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -101,6 +101,11 @@
   # auto pick-up. Set this to `None` to use the default theme variant.
   #decorations_theme_variant: None
 
+  # Resize increments
+  #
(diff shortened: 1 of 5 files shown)


alacritty.yml
```diff
@@ -101,6 +101,11 @@
   # auto pick-up. Set this to `None` to use the default theme variant.
   #decorations_theme_variant: None
 
+  # Resize increments
+  #
+  # Prefer rezining window by descrite steps equal to cell dimensions.
```
chrisduerr, line 106:
```suggestion
  # Prefer resizing window by discrete steps equal to cell dimensions.
```

chrisduerr (changes requested):

chrisduerr (approved):

PR #6712, opened 2023-02-18
Relax horizontal scrolling

kchibisov:
This should trigger horizontal scrolling with touchpads only when the amount of pixel change towards x is highter than towards y.


chrisduerr (approved):

kchibisov:
I've changed the logic to be based on angles, and not just higher.

chrisduerr (approved):

nixpulvis:
I was expecting to see a 45 degree line, or am I confused?

kchibisov:
>I was expecting to see a 45 degree line, or am I confused?

You don't really want to see horizontal scrolling in terminal in any form unless you actually try to do that.

nixpulvis:
Interesting. I'll have to play with it in a bit. This is exciting to see though!

PR #6732, opened 2023-02-24
Fix busy rendering when the same warning persists

kchibisov:
When the same warning is thrown on the each rendering iteration, it'll force alacritty to always render. This commit changes the message queue to add logs only when they are not 
yet queued, preventing 'sticky' log notices from taking over the message bar buffer.

--

Just a side note, on how this was noticed. When calling a function in winit which is not implemented, winit will throw a warning saying that it's not implemented. When running alacritty with -vvv options that was introducing busy rendering loop and in-ability to look
into the trace log of alacritty, since winit was taking over the message bar queue. Now I just have a constant warning from the message bar due to winit, but I can at least see the
trace log...

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1207,7 +1207,7 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                     self.ctx.display.cursor_hidden = false;
                     *self.ctx.dirty = true;
                 },
-                // Add message only in case it's not already queued.
+                // Add message only if it's not already queued.
                 EventType::Message(message) if !self.ctx.message_buffer.is_queued(&message) => {
                     self.ctx.message_buffer.push(message);
                     self.ctx.display.pending_update.dirty = true;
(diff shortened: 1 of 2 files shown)


alacritty/src/event.rs
```diff
@@ -1266,7 +1267,8 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                 },
                 #[cfg(unix)]
                 EventType::IpcConfig(_) => (),
-                EventType::ConfigReload(_) | EventType::CreateWindow(_) => (),
+                EventType::ConfigReload(_) | EventType::CreateWindow(_) | EventType::Message(_) => {
+                },
```
chrisduerr, line 1271: Did `rustfmt` add these `{}` instead of `()`?

alacritty/src/event.rs
```diff
@@ -1207,7 +1207,8 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                     self.ctx.display.cursor_hidden = false;
                     *self.ctx.dirty = true;
                 },
-                EventType::Message(message) => {
+                // Add message only in case it's not already queued.
```
chrisduerr, line 1210:
```suggestion
                // Add message only if it's not already queued.
```

Doesn't really matter, but this is a little shorter and reads nicer imo.

chrisduerr (changes requested):

alacritty/src/event.rs
kchibisov: Yes, because it doesn't fit into the same line anymore.

chrisduerr (approved):

PR #6738, opened 2023-02-26
Add globbing support to config import directive

codepeon:
config.d directories are useful, lets have that.


chrisduerr (changes requested):
This issue has insufficient explanation for why this is something that should be added to Alacritty.

It also adds a new import for the previously unused `glob` crate. While this is an official rust-lang crate, I'm still skeptical about adding more dependencies for this.

codepeon:
I see, and i agree with you regarding adding glob.
There are no real issue I'm tying to solve, this just makes life easier with a split config.
I'm fine with you closing this PR.
If not closed tomorrow evening I'll update the commit message with a better explanation.

chrisduerr:
>  There are no real issue I'm tying to solve, this just makes life easier with a split config.

Can you elaborate on how your configuration file is split?

codepeon:
My config structure looks like this, where these are git tracked.

```
alacritty
├── alacritty.yml
└── config.d
    ├── colors.yml
    ├── fonts.yml
    └── hints.yml
```

Right now i have the following import statement in alacritty.yml

```
import:
 - ~/.config/alacritty/config.d/colors.yml
 - ~/.config/alacritty/config.d/fonts.yml
 - ~/.config/alacritty/config.d/hints.yml
```

With this this PR that would be:

```
import
 - ~/.config/alacritty/config.d/*.yml
```

This allows for untracked files for machine specific config fragments which
does not show as changes to the tracked standard config.

```
alacritty
├── alacritty.yml
└── config.d
    ├── colors.yml
    ├── fonts.yml
    ├── local_untracked_options_for_macbook.yml
    └── hints.yml
```

chrisduerr:
Yeah I think your configuration files are just unnecessarily fragmented. A big issue with globbing is import order. Since configuration files can overwrite each other, the order matters. You wouldn't want to leave that up to randomness or alphabetics.

So I think forcing people to specify every file is actually better, since that way the users are aware of the import order and less likely to report issues about things not working properly.

PR #6739, opened 2023-02-26
Send D&D input through paste action

chrisduerr:
Treating D&D like paste allows using D&D to input text into areas other than the PTY, like the search bar.


kchibisov:
Should add a changelog entry for that.

kchibisov (approved):

Issue #6747, opened 2023-03-01
Unable to disable horizontal scrolling via the config


TheDen:
I sometimes use the trackpad to scroll up/down in vim or selection via a vert block. However, now with horizontal scrolling I can't use the trackpad only scroll vertically, since slight x-axis movements break this vertical-only flow.

### System

OS: macOS
Version: alacritty 0.12.0-rc1 (9de64a5)

chrisduerr:
Can you try the latest master? The X-Axis sensitivity has been reduced significantly.

TheDen:
Tested with build off `master` and it's perfect 👍

PR #6751, opened 2023-03-02
add unmount target + use printf over echo + update PHONY list

TheDen:
PR to update the `Makefile`, changes include

* Adds `unmount` target which ejects the Alacritty volume via `hdiutil`
* Replaces instances of `echo` with `printf`, so now it's consistent 
* Adds missing `all`, `test`, and now `unmount` to `.PHONY`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ $(APP_NAME)-%: $(TARGET)-%
 	@touch -r "$(APP_BINARY)" "$(APP_DIR)/$(APP_NAME)"
 	@codesign --remove-signature "$(APP_DIR)/$(APP_NAME)"
 	@codesign --force --deep --sign - "$(APP_DIR)/$(APP_NAME)"
-	@echo "Created '$(APP_NAME)' in '$(APP_DIR)'"
+	@printf "Created '$(APP_NAME)' in '$(APP_DIR)'\n"
 
 dmg: $(DMG_NAME)-native ## Create an Alacritty.dmg
 dmg-universal: $(DMG_NAME)-universal ## Create a universal Alacritty.dmg
@@ -68,14 +68,17 @@ $(DMG_NAME)-%: $(APP_NAME)-%
 		-fs HFS+ \
 		-srcfolder $(APP_DIR) \
 		-ov -format UDZO
-	@echo "Packed '$(APP_NAME)' in '$(APP_DIR)'"
+	@printf "Packed '$(APP_NAME)' in '$(APP_DIR)'\n"
+
+unmount:
+	@hdiutil detach $$(diskutil info Alacritty | awk -F ':' '/Mount Point/ {print $$2}' | awk '{print $$(NF)}')
 
 install: $(INSTALL)-native ## Mount disk image
 install-universal: $(INSTALL)-native ## Mount universal disk image
 $(INSTALL)-%: $(DMG_NAME)-%
 	@open $(DMG_DIR)/$(DMG_NAME)
 
-.PHONY: app binary clean dmg install $(TARGET) $(TARGET)-universal
+.PHONY: all app binary clean dmg install test unmount $(TARGET) $(TARGET)-universal
 
 clean: ## Remove all build artifacts
 	@cargo clean
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ $(DMG_NAME)-%: $(APP_NAME)-%
 		-ov -format UDZO
 	@printf "Packed '$(APP_NAME)' in '$(APP_DIR)'\n"
 
-unmount:
+unmount: ## Detach universal disk image
 	@hdiutil detach $$(diskutil info Alacritty | awk -F ':' '/Mount Point/ {print $$2}' | awk '{print $$(NF)}')
 
 install: $(INSTALL)-native ## Mount disk image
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
Why?

TheDen:
> Why?

Which part are you asking about?

chrisduerr:
Why are you sending this PR? What is the motivation behind it? There has been zero demand for it to my knowledge.

TheDen:
> Why are you sending this PR? What is the motivation behind it? here has been zero demand for it to my knowledge.

I want to, was building and installing a lot, and idc

TheDen:
Closing this. A reviewer asking snarky one word replies about motives and demands of a PR that simply does minor updates to a `Makefile`...yeah @chrisduerr totally not toxic behaviour...

nixpulvis:
Asking for the motivation for a change is completely reasonable and he even elaborated when asked. As it isn't clearly solving any open issues or known sticking points for people, it can be said to be a pet idea you had. No shame in that, but also no expectation of being merged.

Let's avoid name calling shall we?

TheDen:
> Asking for the motivation for a change is completely reasonable and he even elaborated when asked.

Agree, but for me it was the tone. idk maybe I expected better comms from this project so it was jarring

> Let's avoid name calling shall we?

Agreed, but calling out someone's behaviour isn't the same thing as name calling

Look I'm sure chrisduerr isn't a jerk, I get busy and tired too and slip. It might be good to add a [PR template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) with these common questions, there's an issues template in this repo, and I found it useful.

nixpulvis:
OK, so my curiosity is piqued.

@TheDen:

- Why are you adding a `test` `.PHONY` target when there is no `test`?
- Why are you changing `@echo` to `@printf` when it requires more code?
- Why are you adding a way to `detach` when there is no corresponding `attach`?

As for your last comment. I'd just say that each project and community has their own norms and expectations. I would read through other closed pull requests and issues to get a sense of things personally. It shouldn't really need to be stated that motivating a change is a good idea, especially when the change is somewhat unjustified, as I've pointed out above.

This isn't about being a jerk or not. This is about making good thoughtful changes to the source.

TheDen:
Okay, yeah maybe let's start over

>Why are you adding a test .PHONY target when there is no test?

I use [checkmake](https://github.com/mrtazz/checkmake) and `test` is a required phony target for the `minphony` rule. It passes everything else. I can remove it of course.

>Why are you changing @echo to @printf when it requires more code?

`printf` is also used in the `help` target, so IMO we ought to just use one rather than both `echo` and `printf`. Since `printf` offers better portability, I think it's the right choice.

>Why are you adding a way to detach when there is no corresponding attach?

The `hdiutil create` command under the `dmg` targets creates and mounts the volume. So after running `make install`, it's useful to unmount it via the cli, since `Finder.app` can be buggy when unmounting (at least for me). Maybe the naming of the target and whatnot could be better?

Motivation-wise I was building for [this](https://github.com/alacritty/alacritty/issues/6747) playing around with different [values](https://github.com/alacritty/alacritty/pull/6712/files) and changing my mind after installing it, it got annoying using `Finder.app`,and I saw improvement whilst editing the `Makefile` and thought I'd share them. I mean these are small details, and I usually focus on details, so  I didn't really think much of it and thought the changes were obvious which was perhaps an err on my part.

nixpulvis:
Strange, I even tested with `make dmg`:

```
Created 'Alacritty.app' in 'target/release/osx'
Packing disk image...
..........................
created: /Users/nixpulvis/Developer/alacritty/target/release/osx/Alacritty.dmg
Packed 'Alacritty.app' in 'target/release/osx'
```

Maybe I don't know where it's mounted, but I don't see it. Sorry, I'm not as familiar with macOS as I once was.

nixpulvis:
P.S. funny, I just commented on the horizontal scrolling myself here: https://github.com/alacritty/alacritty/pull/6625#issuecomment-1452990057 😉

nixpulvis:
Oh, you're doing a `make install`, well yea... all that does is open the DMG. I feel like this change is over the top macOS developer handholding though. Frankly, I'd rather see the `install` target removed before this PR merged, since I prefer less code to more code, especially in the build systems.

TheDen:
Yeah I think I shouldn't have been using `make install`, also subsequent multiple `make installs` creates multiple mounts
image

```bash
$ df /Volumes/Alacritty*
Filesystem                      512-blocks      Used  Available Capacity iused      ifree %iused  Mounted on
/dev/disk4s1                         16800     11040       5760    66%      22 4294967257    0%   /Volumes/Alacritty
/dev/disk5s1                         16800     11040       5760    66%      22 4294967257    0%   /Volumes/Alacritty 1
/dev/disk6s1                         16800     11040       5760    66%      22 4294967257    0%   /Volumes/Alacritty 2
```

I think they're distinct mounts, and generally an unergonomic flow MacOS users have gotten used to with mounts. Agree on dropping the `install` target

TheDen:
Removed the `install` targets, and `unmount` target. Also dropped `test` from `.PHONY` since the project doesn't have `makecheck` as a pass req

Running `git grep 'make install'` returns nothing, so AFAIK it's not really used, or at least referenced outside the `Makefile`

chrisduerr (changes requested):
So now this PR basically just removes features that were previously supported? There's really no benefits being provided here.

PR #6762, opened 2023-03-08
Use `thin` instead of `fat` LTO

kchibisov:
The performance between `thin` and `fat` is in the margin of error, however `thin` LTO is way faster to build.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,6 @@ members = [
 ]
 
 [profile.release]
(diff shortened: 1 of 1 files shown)


chrisduerr:
Will be a couple more days probably before perfbot is back, if it doesn't find any delta we can switch.

chrisduerr (approved):
Perfbot seems happy, so I am too.

PR #6784, opened 2023-03-15
Update INSTALL.md

cratolabs:
Since mostly bash is the default shell in the terminal, so if someone using multiple shells, then first go to fish shell to make sure $fish_complete_path[1] is recognized by shell and then run commands to install the completions for fish.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -378,7 +378,13 @@ echo "source ~/.bash_completion/alacritty" >> ~/.bashrc
 
 #### Fish
 
-To install the completions for fish, run
+Make sure you are in fish shell or run
+
+```
+fish
+```
+
+Once you are in fish shell, to install the completions for fish, run
 
 ```
 mkdir -p $fish_complete_path[1]


INSTALL.md
```diff
@@ -378,7 +378,13 @@ echo "source ~/.bash_completion/alacritty" >> ~/.bashrc
...
 #### Fish
 
-To install the completions for fish, run
+Make sure you are in fish shell or run
+
+```
+fish
+```
+
+Once you are in fish shell, to install the completions for fish, run
```
kchibisov, line 387: Maybe something like that? At least it should give you an idea how to adjust it, but the suggestion should be oneline imho.

```suggestion
To install the completions for fish, from inside the fish shell, run
```

chrisduerr (changes requested):
What motivated you to make this change? I've never heard anyone running into this issue. To me this seems kinda obvious?

cratolabs:
> 

I run the command and it created [1]  folder in same directory and not in the fish completion directory.
Then i get to know i am on different shell and the variable is only recognised in fish shell.

chrisduerr (approved):

chrisduerr:
New revision seems concise enough.

PR #6797, opened 2023-03-18
Fix `;` character in URI OSC 8 payload

kchibisov:
The special character `;` can be not URL-encoded, thus it'll add extra parameter in the payload. Handle it joining extra params with the `;` as a separator.


alacritty_terminal/src/ansi.rs
```diff
@@ -1014,7 +1014,14 @@ where
...
                 let link_params = params[1];
-                let uri = str::from_utf8(params[2]).unwrap_or_default();
+                let mut uri = str::from_utf8(params[2]).unwrap_or_default().to_string();
+                // XXX The escape sequence is of form 'OSC 8 ; params ; URI ST, where
+                // URI is URL-encoded, however `;` is a special character and might be
+                // passed as is, thus we should rebuild the URI.
+                for param in params[3..].iter() {
+                    uri.push(';');
+                    uri.push_str(str::from_utf8(param).unwrap_or_default());
+                }
```
chrisduerr, line 1024:
```suggestion

                // NOTE: The escape sequence is of form 'OSC 8 ; params ; URI ST', where
                // URI is URL-encoded. However `;` is a special character and might be
                // passed as is, thus we need to rebuild the URI.
                let mut uri = str::from_utf8(params[2]).unwrap_or_default().to_string();
                for param in params[3..].iter() {
                    uri.push(';');
                    uri.push_str(str::from_utf8(param).unwrap_or_default());
                }
```

chrisduerr (changes requested):

chrisduerr (approved):

Issue #1628, opened 2018-10-05
Feature request: undercurl support


petobens:
As per this [comment](https://github.com/jwilm/alacritty/pull/1078#issuecomment-426444304).

nixpulvis:
I believe this is the relevant spec: https://sw.kovidgoyal.net/kitty/protocol-extensions.html#colored-and-styled-underlines

bluz71:
Hello Alacritty team,

I thought it might be handy to expand this item with some pertinent details.

Proper *undercurl* support actually involves two feature enhancements; those being: rendering actual undercurls, and secondly colouring underlines & undercurls separately from the character they sit beneath.

kitty I believe was the first terminal to get the ball rolling, with the relevant spec listed by @nixpulvis above.

- GNOME Terminal, and other [VTE](https://gitlab.gnome.org/GNOME/vte)-based terminals, added support for coloured undercurls a couple years back via [bug 721761](https://bugzilla.gnome.org/show_bug.cgi?id=721761).

- Vim added support for undercurls [here](https://github.com/vim/vim/issues/1306).

- Neovim added support for undercurls [here](https://github.com/neovim/neovim/pull/9052).

- [tmux](https://github.com/tmux/tmux) staged the introduction of undercurls, first via just undercurls in version 3.0 and then coloured undercurls & underlines in 3.1. tmux [#1492](https://github.com/tmux/tmux/issues/1492) is the relevant issue. Note, plain undercurls require `set -as terminal-overrides ',*:Smulx=\E[4::%p1%dm'` in `tmux.conf` and coloured underlines & undercurls also require `set -as terminal-overrides ',*:Setulc=\E[58::2::%p1%{65536}%/%d::%p1%{256}%/%{255}%&%d::%p1%{255}%&%d%;m'`.

- [mintty](https://github.com/mintty/mintty) it appears also supports coloured undercurls which they note in their [text attributes and rendering](https://github.com/mintty/mintty/wiki/Tips#text-attributes-and-rendering) documentation; [this discussion](https://github.com/kovidgoyal/kitty/issues/226#issuecomment-381665023) may also be of interest.

[...]

PR #6803, opened 2023-03-20
add Setulc to terminfo

ArtAndreev:
Add `Setulc` to terminfo, so colored underlines will be supported by default (no need to use `terminal-overrides` in tmux, for example). Works for `alacritty-direct` TERM.

### DIFF
--- a/extra/alacritty.info
+++ b/extra/alacritty.info
@@ -22,6 +22,7 @@ alacritty-direct|alacritty with direct color indexing,
 	setaf=\E[%?%p1%{8}%<%t3%p1%d%e38\:2\:\:%p1%{65536}%/%d\:%p1%{256}
           %/%{255}%&%d\:%p1%{255}%&%d%;m,
     setb@, setf@,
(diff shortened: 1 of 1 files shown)


ArtAndreev:
I noticed, that underlines are not colored in my neovim inside tmux, but are colored outside (alacritty host). Found how to enable colored underlines there: https://github.com/alacritty/alacritty/issues/1628#issuecomment-663931243

Tmux supports it: https://linuxcommand.org/lc3_man_pages/tmux1.html (section `TERMINFO EXTENSIONS`)

And alacritty, of course.

kchibisov:
The kitty link doesn't suggest the colors though? It's simply `Su`. 

I know that tmux has

```
set-option -ga terminal-features ",alacritty:usstyle"
```

Is this capability is a part of `ncurses`, if so then we may add it.

ArtAndreev:
I haven't found info about `Setulc` in ncurses. Looks like, only vim/neovim and tmux support this. Alacritty also shows colored underlines correctly. So okay, will close then.

kchibisov:
nvim uses style underline escape sequence as heuristic to enable colors.

PR #6804, opened 2023-03-20
alacritty: prefer alacritty-direct over alacritty terminfo

ArtAndreev:
By default, alacritty sets `TERM` to `alacritty` even if `alacritty-direct` terminfo exists (typically installed by `tic` from `alacritty.info`). So the default behavior forces use of not truecolor term. I suggest to check for `alacritty-direct` terminfo first, then for `alacritty` and fallback to `xterm-256color` as before.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -19,9 +19,9 @@
   #
   # This value is used to set the `$TERM` environment variable for
   # each instance of Alacritty. If it is not present, alacritty will
-  # check the local terminfo database and use `alacritty` if it is
-  # available, otherwise `xterm-256color` is used.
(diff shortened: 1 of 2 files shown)


kchibisov:
No, that will break lots of programs (I don't even use alacritty-direct, so it'll force me to change config, as well as many others).

Direct terminfo shouldn't be used as default, since truecolors are not desired most of the time. Same applies for xterm as well, they use xterm-256color, even though, there's a direct version. If your software requires direct(I think only emacs or plain vim), use it, but most don't.

ArtAndreev:
OK, thanks for clarification, sorry for noise

PR #6807, opened 2023-03-22
Fix selection rotation on the last line

### LINKED ISSUE #6698: Selection wrong when search end on last line
When the entire terminal is filled and a non-vi search is started for something in the last line, then the selection will shift up a line once the selection is confirmed.

chrisduerr:
This fixes an issue with terminal resizes when the selection is on the
last line. Alacritty would fail to rotate lines and keep the selection
in the same line index whenever the terminal line count was grown or
shrunk.

This issue occurred due to the range passed to the selection's rotate
function still being based on the old terminal size, which caused the
initial or target state of the rotation to be outside of the terminal
bounds.

Closes #6698.

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -588,11 +588,10 @@ impl Display {
         });
     }
 
+    // XXX: this function must not call to any `OpenGL` related tasks. Renderer updates are
+    // performed in [`Self::process_renderer_update`] right befor drawing.
+    //
     /// Process update events.
-    ///
-    /// XXX: this function must not call to any `OpenGL` related tasks. Only logical update
-    /// of the state is being performed here. Rendering update takes part right before the
-    /// actual rendering.
     pub fn handle_update<T>(
         &mut self,
         terminal: &mut Term<T>,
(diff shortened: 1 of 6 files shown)


alacritty/src/display/mod.rs
```diff
@@ -666,14 +665,11 @@ impl Display {
         self.size_info = new_size;
     }
 
+    // NOTE: Renderer updates are split off, since platforms like Wayland require resize and other
+    // OpenGL operations to be performed right before rendering. Otherwise they could lock the
+    // back buffer and render with the previous state. This also solves flickering during resizes.
+    //
     /// Update the state of the renderer.
```
kchibisov, line 672: Split this into clippy fixes.

alacritty/src/display/mod.rs
```diff
@@ -588,11 +588,10 @@ impl Display {
         });
     }
 
+    // XXX: this function must not call to any `OpenGL` related tasks. Renderer updates are
+    // performed in [`Self::process_renderer_update`] right befor drawing.
+    //
     /// Process update events.
```
kchibisov, line 594: Split this into clippy fixes.

kchibisov (changes requested):
Also requires a CHANGELOG entry.

alacritty/src/display/mod.rs
chrisduerr: I'll squash the clippy fixes with the rest, I don't think it should be relevant.

I just made it separate commits to make the review a little easier.

kchibisov: Ok.

chrisduerr:
Should merge https://github.com/alacritty/alacritty/pull/6808 before this.

kchibisov:
Just so you know, CHANGELOG entry is missing here, you likely forgot to push it.

chrisduerr:
> Just so you know, CHANGELOG entry is missing here, you likely forgot to push it.

I don't know what you're talking about. It was clearly there already. :sweat_smile:

kchibisov (approved):

PR #6808, opened 2023-03-22
Bump CHANGELOG development version to 0.13.0-dev

chrisduerr:
This is only an update to the development version and does not represent a stable release.

The original bump to 0.13.0-dev was made in 9cf2c89, but did not make any changes to the changelog. This adds a 0.13.0-dev changelog entry to master so new changes are properly added to the correct place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,13 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 0.12.0-dev
+## 0.13.0-dev
+
+### Fixed
+
+- Character `;` inside the `URI` in `OSC 8` sequence breaking the URI
+
+## 0.12.0
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,13 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 0.12.0-dev
+## 0.13.0-dev
+
+### Fixed
+
+- Character `;` inside the `URI` in `OSC 8` sequence breaking the URI
```
kchibisov, line 12: Don't you want to add it into v0.12? I've intentionally scheduled it there given how trivial it is.

chrisduerr: I don't think it's actually a significant change and I want to make the 0.12.0 release this weekend. So I'd rather not make another RC and pull in even more stuff.

I feel like 0.12.0 is good where it is and I'd only add new stuff when there is serious breakages that were introduced in this specific version.

kchibisov (approved):

PR #6816, opened 2023-03-25
Add `window.use_srgb` config option for macOS

### LINKED ISSUE #6815: [0.12.0][macOS] Colors washed out after updating
After upgrading to `0.12.0`, the colors in neovim appear all washed out. It is probably due to #6602. I understand the author's reasons for wanting sRGB for color work, but I prefer the previous versions' "vivid" look. Please introduce a config option for the same.

### System

OS: macOS
Version: alacritty 0.12.0 (5a72819)

### Screenshot
I don't know if you can tell the difference, but the colors in the new version look really washed out to me - akin to watching HDR content on a non-HDR display.

image

utkarshgupta137:
Adds a config option to use the default colorspace on macOS.

It may make sense for many users to use the sRGB color space, but some, like myself, prefer the vivid look of the default color space. I don't do any color work & I don't care if the colors are "accurate" are not. Setting this option to false restores colors prior to 0.12.0.

Closes #6815

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Character `;` inside the `URI` in `OSC 8` sequence breaking the URI
 - Selection on last line not updating correctly on resize
+- Added `use_srgb` option for macOS (default true)
 
 ## 0.12.0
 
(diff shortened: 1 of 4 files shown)


xiaogaozi:
The old behavior (do not use sRGB color space) is much better for me, sadly I can only roll back to the old behavior by forking the project and apply this PR at the moment.

PR #6772, opened 2023-03-11
Port from mio to polling

notgull:
As `mio` v0.6 is currently a "ticking timebomb", it's been requested to port from `mio` to `polling`. This PR ports the `event_loop` system from `mio` to the `polling` crate, which serves a similar purpose.

Draft, since I haven't done Windows support yet. This is also a breaking change for `alacritty_terminal`.

PR #6846, opened 2023-04-08
Port from mio to polling, take 2

### LINKED ISSUE #6486: Update Mio
It seems alacritty_terminal is still using Mio v0.6, which isn't supported any more. Any thing I can do to help move it to Mio v0.8?

### LINKED ISSUE #7104: alacritty fails to build in Windows on ARM
this error appears when try to build for `aarch64-pc-windows-gnullvm` target
```console
error[E0412]: cannot find type `ULONG_PTR` in this scope
    --> C:\Users\Administrator\.cargo\registry\src\index.crates.io-6f17d22bba15001f\winapi-0.2.8\src\basetsd.rs:86:19
     |
  86 | pub type SIZE_T = ULONG_PTR;
     |                   ^^^^^^^^^ not found in this scope
```
that's because there is winapi 0.2, which doesn't support arm64 target. is there any possibility to update winapi package?

notgull:
Supersedes #6772 by replacing the `mio` crate with the `polling` crate. Now that smol-rs/polling#96 has been merged, we should be at full feature parity with `mio` v0.6 now.

Draft because I haven't ported the Windows code over to `polling` yet, but from what I understand that is low-priority.

Fixes #7104.
Fixes #6486.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -713,12 +713,9 @@ dependencies = [
 
 [[package]]
 name = "fastrand"
-version = "1.9.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e51093e27b0797c359783294ca4f0a911c270184cb10f85783b118614a1501be"
-dependencies = [
- "instant",
-]
+checksum = "25cbce373ec4653f1a01a31e8a5e5ec0c622dc27ff9c4e6606eefef5cbbed4a5"
 
 [[package]]
 name = "fdeflate"
@@ -1012,15 +1009,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e05c02b5e89bff3b946cedeca278abc628fe811e604f027c45a8aa3cf793d0eb"
 dependencies = [
  "libc",
-]
-
-[[package]]
-name = "instant"
-version = "0.1.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7a5bbe824c507c5da5956355e86a746d82e0e1464f65d862cc5e71da70e94b2c"
-dependencies = [
- "cfg-if",
 ]
 
 [[package]]
@@ -1505,9 +1493,9 @@ checksum = "8afb450f006bf6385ca15ef45d71d2288452bc3683ce2e2cacc0d18e4be60b58"
 
 [[package]]
 name = "piper"
-version = "0.2.0"
+version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "563ed39531c3ede9d6d76c99cb0f200c46a572ec1a10bbd62d7e0fcf5b1a1d39"
+checksum = "668d31b1c4eba19242f2088b2bf3316b82ca31082a8335764db4e083db7485d4"
 dependencies = [
  "atomic-waker",
  "fastrand",
@@ -2065,7 +2053,7 @@ version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "401dc1020e10f74d38616c1f1ab92ccd85dc902705a29d0730e0fbea8534f91a"
 dependencies = [
- "bitflags 2.3.3",
+ "bitflags 2.4.0",
  "cursor-icon",
  "log",
  "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,6 +103,7 @@ dependencies = [
  "miow",
  "nix 0.26.4",
  "parking_lot",
+ "piper",
  "polling",
  "regex-automata",
  "serde",
@@ -711,6 +712,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "fastrand"
+version = "1.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e51093e27b0797c359783294ca4f0a911c270184cb10f85783b118614a1501be"
+dependencies = [
+ "instant",
+]
+
+[[package]]
 name = "fdeflate"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -819,6 +829,12 @@ checksum = "76ee7a02da4d231650c7cea31349b889be2f45ddb3ef3032d2ec8185f6313fd2"
 dependencies = [
  "libc",
 ]
+
+[[package]]
+name = "futures-io"
+version = "0.3.28"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4fff74096e71ed47f8e023204cfd0aa1289cd54ae5430a9523be060cdb849964"
 
 [[package]]
 name = "gethostname"
@@ -996,6 +1012,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e05c02b5e89bff3b946cedeca278abc628fe811e604f027c45a8aa3cf793d0eb"
 dependencies = [
  "libc",
+]
+
+[[package]]
+name = "instant"
+version = "0.1.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7a5bbe824c507c5da5956355e86a746d82e0e1464f65d862cc5e71da70e94b2c"
+dependencies = [
+ "cfg-if",
 ]
 
 [[package]]
@@ -1477,6 +1502,17 @@ name = "pin-project-lite"
 version = "0.2.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8afb450f006bf6385ca15ef45d71d2288452bc3683ce2e2cacc0d18e4be60b58"
+
+[[package]]
+name = "piper"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "563ed39531c3ede9d6d76c99cb0f200c46a572ec1a10bbd62d7e0fcf5b1a1d39"
+dependencies = [
+ "atomic-waker",
+ "fastrand",
+ "futures-io",
+]
 
 [[package]]
 name = "pkg-config"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,9 +22,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.5"
+version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c378d78423fdad8089616f827526ee33c19f2fddbd5de1629152c9593ba4783"
+checksum = "ea5d730647d4fadd988536d06fecce94b7b4f2a7efdae548f1cf4b63205518ab"
 dependencies = [
  "memchr",
 ]
@@ -85,7 +85,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "serde",
- "syn",
+ "syn 2.0.37",
  "toml 0.7.8",
 ]
 
@@ -145,9 +145,9 @@ checksum = "fc7eb209b1518d6bb87b283c20095f5228ecda460da70b44f0802523dea6da04"
 
 [[package]]
 name = "anstream"
-version = "0.5.0"
+version = "0.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b1f58811cfac344940f1a400b6e6231ce35171f614f26439e80f8c1465c5cc0c"
+checksum = "2ab91ebe16eb252986481c5b62f6098f3b698a45e34b5b98200cf20dd2484a44"
 dependencies = [
  "anstyle",
  "anstyle-parse",
@@ -159,15 +159,15 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.2"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "15c4c2c83f81532e5845a733998b6971faca23490340a418e9b72a3ec9de12ea"
+checksum = "7079075b41f533b8c61d2a4d073c4676e1f8b249ff94a393b0595db304e0dd87"
 
 [[package]]
 name = "anstyle-parse"
-version = "0.2.1"
+version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "938874ff5980b03a87c5524b3ae5b59cf99b1d6bc836848df7bc5ada9643c333"
+checksum = "317b9a89c1868f5ea6ff1d9539a69f45dffc21ce321ac1fd1160dfa48c8e2140"
 dependencies = [
  "utf8parse",
 ]
@@ -183,9 +183,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle-wincon"
-version = "2.1.0"
+version = "3.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "58f54d10c6dfa51283a066ceab3ec1ab78d13fae00aa49243a45e4571fb79dfd"
+checksum = "f0699d10d2f4d628a98ee7b57b289abbc98ff3bad977cb3152709d4bf2330628"
 dependencies = [
  "anstyle",
  "windows-sys 0.48.0",
@@ -193,15 +193,15 @@ dependencies = [
 
 [[package]]
 name = "arrayref"
-version = "0.3.7"
+version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6b4930d2cb77ce62f89ee5d5289b4ac049559b1c45539271f5ed4fdc7db34545"
+checksum = "a4c527152e37cf757a3f78aae5a06fbeefdb07ccc535c980a3208ee3060dd544"
 
 [[package]]
 name = "arrayvec"
-version = "0.7.4"
+version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
+checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "as-raw-xcb-connection"
@@ -211,9 +211,9 @@ checksum = "2d5f312b0a56c5cdf967c0aeb67f6289603354951683bc97ddc595ab974ba9aa"
 
 [[package]]
 name = "atomic-waker"
-version = "1.1.1"
+version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1181e1e0d1fce796a03db1ae795d67167da795f9cf4a39c37589e85ef57f26d3"
+checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
 name = "autocfg"
@@ -223,9 +223,9 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "base64"
-version = "0.13.1"
+version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
+checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
 name = "bitflags"
@@ -288,9 +288,9 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.13.0"
+version = "3.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a3e2c3daef883ecc1b5d58c15adae93470a91d425f3532ba1695849656af3fc1"
+checksum = "0d261e256854913907f67ed06efbc3338dfe6179796deefc1ff763fc1aee5535"
 
 [[package]]
 name = "bytemuck"
@@ -320,12 +320,11 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.83"
+version = "1.0.73"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f1174fb0b6ec23863f8b971027804a42614e347eafb0a95bf0b12cdae21fc4d0"
+checksum = "2fff2a6927b3bb87f9595d67196a70493f627687a71d87a0d692242c33f58c11"
 dependencies = [
  "jobserver",
- "libc",
 ]
 
 [[package]]
@@ -357,9 +356,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.4.2"
+version = "4.4.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a13b88d2c62ff462f88e4a121f17a82c1af05693a2f192b5c38d14de73c19f6"
+checksum = "d04704f56c2cde07f43e8e2c154b43f216dc5c92fc98ada720177362f953b956"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -367,9 +366,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.4.2"
+version = "4.4.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2bb9faaa7c2ef94b2743a21f5a29e6f0010dff4caa69ac8e9d6cf8b6fa74da08"
+checksum = "0e231faeaca65ebd1ea3c737966bf858971cd38c3849107aa3ea7de90a804e45"
 dependencies = [
  "anstream",
  "anstyle",
@@ -379,9 +378,9 @@ dependencies = [
 
 [[package]]
 name = "clap_complete"
-version = "4.4.1"
+version = "4.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4110a1e6af615a9e6d0a36f805d5c99099f8bab9b8042f5bc1fa220a4a89e36f"
+checksum = "e3ae8ba90b9d8b007efe66e55e48fb936272f5ca00349b5b0e89877520d35ea7"
 dependencies = [
  "clap",
 ]
@@ -395,7 +394,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.37",
 ]
 
 [[package]]
@@ -416,18 +415,18 @@ dependencies = [
 
 [[package]]
 name = "cmake"
-version = "0.1.50"
+version = "0.1.48"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a31c789563b815f77f4250caee12365734369f942439b7defd71e18a48197130"
+checksum = "e8ad8cef104ac57b68b89df3208164d228503abbdce70f6880ffa3d970e7443a"
 dependencies = [
  "cc",
 ]
 
 [[package]]
 name = "cocoa"
-version = "0.24.1"
+version = "0.24.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f425db7937052c684daec3bd6375c8abe2d146dca4b8b143d6db777c39138f3a"
+checksum = "6f63902e9223530efb4e26ccd0cf55ec30d592d3b42e21a28defc42a9586e832"
 dependencies = [
  "bitflags 1.3.2",
  "block",
@@ -441,9 +440,9 @@ dependencies = [
 
 [[package]]
 name = "cocoa-foundation"
-version = "0.1.1"
+version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "931d3837c286f56e3c58423ce4eba12d08db2374461a785c86f672b08b5650d6"
+checksum = "7ade49b65d560ca58c403a479bb396592b155c0185eada742ee323d1d68d6318"
 dependencies = [
  "bitflags 1.3.2",
  "block",
@@ -472,9 +471,9 @@ dependencies = [
 
 [[package]]
 name = "concurrent-queue"
-version = "2.2.0"
+version = "2.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "62ec6771ecfa0762d24683ee5a32ad78487a3d3afdc0fb8cae19d2c5deb50b7c"
+checksum = "f057a694a54f12365049b0958a1685bb52d567f5593b355fbf685838e873d400"
 dependencies = [
  "crossbeam-utils",
 ]
@@ -505,9 +504,9 @@ dependencies = [
 
 [[package]]
 name = "core-foundation-sys"
-version = "0.8.4"
+version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e496a50fda8aacccc86d7529e2c1e0892dbd0f898a6b5645b5561b89c3210efa"
+checksum = "5827cebf4670468b8772dd191856768aedcb1b0278a04f989f7766351917b9dc"
 
 [[package]]
 name = "core-graphics"
@@ -524,12 +523,13 @@ dependencies = [
 
 [[package]]
 name = "core-graphics-types"
-version = "0.1.2"
+version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2bb142d41022986c1d8ff29103a1411c8a3dfad3552f87a4f8dc50d61d4f4e33"
+checksum = "3a68b68b3446082644c91ac778bf50cd4104bfb002b5a6a7c44cca5a2c70788b"
 dependencies = [
  "bitflags 1.3.2",
  "core-foundation",
+ "foreign-types 0.3.2",
  "libc",
 ]
 
@@ -556,9 +556,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-channel"
-version = "0.5.8"
+version = "0.5.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a33c2bf77f2df06183c3aa30d1e96c0695a313d4f9c453cc3762a6db39f99200"
+checksum = "c2dd04ddaf88237dc3b8d8f9a3c1004b506b54b3313403944054d23c0870c521"
 dependencies = [
  "cfg-if",
  "crossbeam-utils",
@@ -566,9 +566,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-utils"
-version = "0.8.16"
+version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5a22b2d63d4d1dc0b7f1b6b2747dd0088008a9be28b6ddf0b1e7d335e3037294"
+checksum = "edbafec5fa1f196ca66527c1b12c2ec4745ca14b50f1ad8f9f6f720b55d11fac"
 dependencies = [
  "cfg-if",
 ]
@@ -638,7 +638,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "330c60081dcc4c72131f8eb70510f1ac07223e5d4163db481a04a0befcffa412"
 dependencies = [
- "libloading 0.8.0",
+ "libloading",
 ]
 
 [[package]]
@@ -663,13 +663,13 @@ dependencies = [
 
 [[package]]
 name = "embed-resource"
-version = "1.8.0"
+version = "1.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e62abb876c07e4754fae5c14cafa77937841f01740637e17d78dc04352f32a5e"
+checksum = "936c1354206a875581696369aef920e12396e93bbd251c43a7a3f3fa85023a7d"
 dependencies = [
  "cc",
  "rustc_version",
- "toml 0.5.11",
+ "toml 0.5.8",
  "vswhom",
  "winreg",
 ]
@@ -682,9 +682,9 @@ checksum = "5443807d6dff69373d433ab9ef5378ad8df50ca6298caf15de6e52e24aaf54d5"
 
 [[package]]
 name = "errno"
-version = "0.3.3"
+version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "136526188508e25c6fef639d7927dfb3e0e3084488bf202267829cf7fc23dbdd"
+checksum = "add4f07d43996f76ef320709726a556a9d4f965d9410d8d0271132d2f8293480"
 dependencies = [
  "errno-dragonfly",
  "libc",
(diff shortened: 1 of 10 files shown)


kchibisov:
>Draft because I haven't ported the Windows code over to polling yet, but from what I understand that is low-priority.

Yeah, the windows code could be as dirty as it can, because it's a "best effort" platform, though, we must have an implementation for it in case we want to merge it.

Even doing the thing with `thread` as we did before is fine for windows.

Also, don't forget to include the `Cargo.lock` update, since right now it's missing.

kchibisov:
@davidhewitt sorry for pinging you, but given that you're the maintainer of mio-anonymous-pipe and the author of alacritty's ConPty integration bits, do you know whether your crate could be easily ported to `polling`, given that `mio v0.6.x` is not maintained anymore?

We'd like to remove `mio v0.6.x` from our dependency graph before the next alacritty release to remove future incompatible code, thx.

notgull:
It should be easy enough once smol-rs/polling#104 lands, which would make it possible to push arbitrary events to the `Poller`, allowing much more flexibility in how external sources are handled.

davidhewitt:
I wouldn't publish a new version of `mio-anonymous-pipe` with polling; the library is very small and it's named specifically for `mio`. 

The whole idea of that library is just to create a thread which reads from the pipe and pushes it into a buffer which implements the `mio` events. It's quite small (less than 500 lines) so should be straightforward to create a fork of that code which uses `polling`. Could maybe even inline directly into Alacritty depending on how much smaller the `polling` implementation would be.

I don't have much time at the moment, so I wouldn't be able to do this for you. Potentially some time can free up a few months down the line.

kchibisov:
>I wouldn't publish a new version of mio-anonymous-pipe with polling; the library is very small and it's named specifically for mio.

I've asked given that mio version you were using is not maintained anymore, so your crate is sort of timed bombed due to future incompat code in mio `0.6`

kchibisov:
>Right, I don't think it would be hard to publish a new version on a new mio, but I don't think that's what you need here? 

Right, but it's more like, we don't have plans on keep using `mio`, because it's `Edge` only and we won't use mio in any upstream crate anyway for that reason. It's more of a notice at this point that your crate is sort of dead with the `mio` `0.6`, but I guess no-one needs to
overlapped IO on windows besides alacritty anyway.

I think @notgull is more or less fine. It's just a matter of a code reuse, since I have a feeling that the core parts of it should stay the same, so the code could be taken from your crate. 

Though, I think given that the crate has `mio-anonymous-pipe`, it's better to stick with `mio`, so the only update to it could be to `0.8.0` mio, but not sure that it'll be used.

notgull:
Ported the Windows code over to use `polling` now. I just need to test it on my Windows box and we should be good to go.

notgull:
Seems to work on my Windows box, everything appears to be clean. I think this is ready to be reviewed/merged now.

alacritty_terminal/src/event_loop.rs
```diff
@@ -313,17 +335,12 @@ where
             let mut state = State::default();
             let mut buf = [0u8; READ_BUFFER_SIZE];
 
-            let mut tokens = (0..).map(Into::into);
-
-            let poll_opts = PollOpt::edge() | PollOpt::oneshot();
-
-            let channel_token = tokens.next().unwrap();
-            self.poll.register(&self.rx, channel_token, Ready::readable(), poll_opts).unwrap();
+            let poll_opts = polling::PollMode::EdgeOneshot;
```
kchibisov, line 338: Thinking about it, do we really need `OneShot`? Shouldn't we be fine with just `Edge`?

alacritty_terminal/src/tty/windows/blocking.rs
```diff
@@ -0,0 +1,285 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//! Code for running a reader/writer on another thread while driving it through `polling`.
+
+use piper::{pipe, Reader, Writer};
+use polling::os::iocp::{CompletionPacket, PollerIocpExt};
+use polling::{Event, PollMode, Poller};
+
```
kchibisov, line 7: Do we really need this code here? I'd assume some other clients could be interested in it? We also don't have SPDX prefixes in the project in case they match our licenses.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -17,9 +15,13 @@ use windows_sys::Win32::System::Threading::{
 
 use crate::config::PtyConfig;
 use crate::event::{OnResize, WindowSize};
+use crate::tty::windows::blocking::{UnblockedReader, UnblockedWriter};
 use crate::tty::windows::child::ChildExitWatcher;
 use crate::tty::windows::{cmdline, win32_string, Pty};
 
+// Adjust to taste. Too high means higher memory consumption, too low means over-synchronization.
+const PIPE_CAPACITY: usize = 4 * 1024 * 1024;
```
kchibisov, line 23: Why that particular value was picked? We have a `READ_BUFFER_SIZE` inside the event_loop, so maybe it should be tuned based on it?

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -12,27 +14,43 @@ use windows_sys::Win32::System::WindowsProgramming::INFINITE;
...
+    if let Some(interest) = interest.as_ref() {
+        interest.poller.post(CompletionPacket::new(interest.event)).ok();
+    }
 }
 
 pub struct ChildExitWatcher {
     wait_handle: AtomicPtr<c_void>,
-    event_rx: Receiver<ChildEvent>,
+    event_rx: mpsc::Receiver<ChildEvent>,
+    interest: Arc<Mutex<Option<Interest>>>,
```
kchibisov, line 44: Should it really be behind the `Mutex`? Don't we have a way to have `&mut` in the chain?

kchibisov (changes requested):
I haven't got to actually testing it.

alacritty_terminal/src/tty/windows/blocking.rs
notgull: I would definitely be willing to extract this into a `polling-utils`-esque crate for other users.

alacritty_terminal/src/event_loop.rs
```diff
@@ -73,13 +97,25 @@ impl event::Notify for Notifier {
...
 impl event::OnResize for Notifier {
     fn on_resize(&mut self, window_size: WindowSize) {
-        let _ = self.0.send(Msg::Resize(window_size));
+        self.0.send(Msg::Resize(window_size));
+    }
+}
+
+pub struct EventLoopSender {
+    sender: Sender<Msg>,
+    poller: Arc<polling::Poller>,
```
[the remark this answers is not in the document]
notgull: Passing around the `Poller` is necessary on Windows at least for passing packets to the IOCP.

alacritty_terminal/src/tty/windows/child.rs
notgull: It is also used by the child event handler, so this is needed to share it between the two threads.

alacritty_terminal/src/event_loop.rs
kchibisov: Though, from what I see now the main reason for `OneShot` is because we change between read/write modes.

notgull: Looks like everything still works on Linux under `Oneshot`.

alacritty_terminal/src/tty/windows/conpty.rs
notgull: It's the default pipe capacity used in the `blocking` crate. I've fixed it to just equal `READ_BUFFER_SIZE`.

alacritty_terminal/src/event_loop.rs
kchibisov: I was curious to have just `Edge` and remove `OneShot`.

alacritty_terminal/src/tty/windows/conpty.rs
kchibisov: Maybe it should be benchmarked, with e.g. vtebench, but from what I remember conpty is so slow that it doesn't matter what we pick.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -20,7 +20,7 @@ use crate::tty::windows::child::ChildExitWatcher;
 use crate::tty::windows::{cmdline, win32_string, Pty};
 
 // Adjust to taste. Too high means higher memory consumption, too low means over-synchronization.
```
chrisduerr, line 22: This comment needs updating.

kchibisov:
I've tested on windows, and it was sort of working.

alacritty_terminal/src/tty/windows/conpty.rs
kchibisov: I also meant that we can use `8 * BUFFER` or something based on that if it's "better".

kchibisov:
Also "tb" macOS.

alacritty_terminal/src/tty/windows/conpty.rs
kchibisov: We could check with `vtebench` whether changing anything results in change in performance.

chrisduerr:
@perfbot

alacritty_terminal/src/event_loop.rs
```diff
@@ -39,14 +36,41 @@ pub enum Msg {
     Resize(WindowSize),
 }
 
-/// The main event!.. loop.
+struct PeekableReceiver<T> {
```
chrisduerr, line 39: This structure seems a bit strange and inefficient to me. Did you add it solely for synchronized updates? Why was it necessary if it worked without this before?

Also it seems like you're silently ignoring channel failure. How does this impact the behavior of the channel's death compared to master?

alacritty_terminal/src/event_loop.rs
```diff
@@ -336,29 +353,29 @@ where
...
-                for event in events.iter() {
-                    match event.token() {
-                        token if token == channel_token => {
-                            if !self.channel_event(channel_token, &mut state) {
-                                break 'event_loop;
-                            }
-                        },
+                // Handle channel events, if there are any.
+                if !self.drain_recv_channel(&mut state) {
+                    break 'event_loop;
```
chrisduerr, line 373: Seems like you can just use `break` here?

alacritty_terminal/src/tty/mod.rs
```diff
@@ -24,18 +25,20 @@ pub trait EventedReadWrite {
...
     ) -> io::Result<()>;
-    fn reregister(&mut self, _: &mio::Poll, _: mio::Ready, _: mio::PollOpt) -> io::Result<()>;
-    fn deregister(&mut self, _: &mio::Poll) -> io::Result<()>;
+    fn reregister(
+        &mut self,
+        _: &Arc<polling::Poller>,
+        _: polling::Event,
+        _: polling::PollMode,
+    ) -> io::Result<()>;
+    fn deregister(&mut self, _: &Arc<polling::Poller>) -> io::Result<()>;
```
chrisduerr, line 38: Should probably import this stuff at the top of the file.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -5,23 +5,30 @@ use std::fs::File;
...
 use signal_hook::consts as sigconsts;
-use signal_hook_mio::v0_6::Signals;
 
 use crate::config::PtyConfig;
 use crate::event::{OnResize, WindowSize};
 use crate::tty::{ChildEvent, EventedPty, EventedReadWrite};
 
+// We use the same interest for reading and writing.
+pub(crate) const PTY_READ_TOKEN: usize = 0;
+pub(crate) const PTY_WRITE_TOKEN: usize = 0;
```
chrisduerr, line 27: Why are they separate constants then? Why not make it a `PTY_READ_WRITE_TOKEN`?

Also couldn't this just be an enum to avoid accidental conflicts?

alacritty_terminal/src/tty/unix.rs
```diff
@@ -254,7 +261,14 @@ pub fn new(config: &PtyConfig, window_size: WindowSize, window_id: u64) -> Resul
     }
 
     // Prepare signal handling before spawning child.
-    let signals = Signals::new([sigconsts::SIGCHLD]).expect("error preparing signal handling");
+    let signals = {
+        let (sender, recv) = UnixStream::pair()?;
+
+        // Register the recv end of the pipe for SIGCHLD.
+        signal_hook::low_level::pipe::register(sigconsts::SIGCHLD, sender)?;
```
chrisduerr, line 268: Definitely shouldn't specify the fully qualified path here.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -302,89 +316,87 @@ impl EventedReadWrite for Pty {
...
 }
 
 impl EventedPty for Pty {
     #[inline]
     fn next_child_event(&mut self) -> Option<ChildEvent> {
-        self.signals.pending().next().and_then(|signal| {
-            if signal != sigconsts::SIGCHLD {
-                return None;
-            }
+        use std::io::Read;
```
chrisduerr, line 372: Please don't ever add imports inside a function like this.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -302,89 +316,87 @@ impl EventedReadWrite for Pty {
...
+                // We received the signal.
+            },
 
-    #[inline]
-    fn child_event_token(&self) -> mio::Token {
-        self.signals_token
+            Err(e) => {
+                // We didn't receive the signal.
+                // Either an error occurred or we got EAGAIN. We don't really care which one.
+                if e.kind() != std::io::ErrorKind::WouldBlock {
```
chrisduerr, line 384: No fully qualified path here please.

alacritty_terminal/src/tty/windows/blocking.rs
chrisduerr: Why is a different license necessary here to begin with?

alacritty_terminal/src/tty/windows/blocking.rs
```diff
@@ -0,0 +1,285 @@
...
+use std::task::{Context, Poll, Wake, Waker};
+use std::{io, thread};
+
+struct Registration {
+    interest: Mutex<Option<Interest>>,
+    end: End,
+}
+
+#[derive(Copy, Clone)]
+enum End {
```
chrisduerr, line 20: The enum name here isn't very descriptive.

alacritty_terminal/src/tty/windows/blocking.rs
```diff
@@ -0,0 +1,285 @@
...
+        // Create a new pipe.
+        let (reader, mut writer) = pipe(pipe_capacity);
+        let interest = Arc::new(Registration {
+            interest: Mutex::<Option<Interest>>::new(None),
+            end: End::Reader,
+        });
+
+        // Spawn the reader thread.
+        thread::Builder::new()
+            .name("alacritty-tty-reader-thread".into())
```
chrisduerr, line 63: Pretty sure we have a utility to spawn a named thread that should make this a little less verbose.

chrisduerr (changes requested):

alacritty_terminal/src/event_loop.rs
notgull: The channel was previously registered into the event loop, while this version isn't. I added a panic on channel death.

alacritty_terminal/src/tty/unix.rs
notgull: They have to be separate in Windows but not Unix. They need to be `usize`, since that's the token type `polling` takes.

alacritty_terminal/src/tty/windows/blocking.rs
notgull: Removed the notice. I thought there was an Apache 2.0 license, my bad

alacritty_terminal/src/event_loop.rs
notgull: That seems to work properly as well.

alacritty_terminal/src/tty/windows/blocking.rs
kchibisov: The crate is apache only, so you should keep the notice if you want. Though, you can just move the part of you want to your utils.

alacritty_terminal/src/event_loop.rs
kchibisov: I mean, it's level, so it should work...

alacritty_terminal/src/tty/windows/blocking.rs
chrisduerr: As the original code author @notgull can do whatever he wants with the code he wrote himself. This includes inclusion in his own projects under whatever license he wants.

The only difference this would make is that this would give a different license option to people using this code (like ourselves). But since Apache and MIT are compatible licenses this isn't really a major factor.

kchibisov:
I think something like that should work.

```diff
diff --git a/alacritty_terminal/src/event_loop.rs b/alacritty_terminal/src/event_loop.rs
index ac2b7123..4eb9f0da 100644
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -199,7 +199,7 @@ where
     ) -> EventLoop<T, U> {
         let (tx, rx) = mpsc::channel();
         EventLoop {
-            poll: polling::Poller::new().expect("create mio Poll").into(),
+            poll: polling::Poller::new().expect("create Poll").into(),
             pty,
             tx,
             rx: PeekableReceiver::new(rx),
@@ -338,10 +338,12 @@ where
             let mut state = State::default();
             let mut buf = [0u8; READ_BUFFER_SIZE];
 
-            let poll_opts = polling::PollMode::Edge;
+            let poll_opts = polling::PollMode::Level;
+
+            let mut interest = polling::Event::readable(0);
 
             // Register TTY through EventedRW interface.
-            self.pty.register(&self.poll, polling::Event::readable(0), poll_opts).unwrap();
+            self.pty.register(&self.poll, interest, poll_opts).unwrap();
 
             let mut events = Vec::with_capacity(1024);
 
@@ -394,7 +396,7 @@ where
                         },
 
                         token if token == tty::PTY_READ_TOKEN || token == tty::PTY_WRITE_TOKEN => {
-                            /*
+                            /* FIXME: once polling supports extra reported modes bring it back.
                             #[cfg(unix)]
                             if UnixReady::from(event.readiness()).is_hup() {
                                 // Don't try to do I/O on a dead PTY.
@@ -431,13 +433,16 @@ where
                     }
                 }
 
-                // Register write interest if necessary.
-                let mut interest = polling::Event::readable(0);
-                if state.needs_write() {
-                    interest.writable = true;
+                if state.needs_write() != interest.writable {
+                    if state.needs_write() {
+                        interest.writable = true;
+                    } else {
+                        interest.writable = false;
+                    }
+
+                    // Reregister with new interest.
+                    self.pty.reregister(&self.poll, interest, poll_opts).unwrap();
                 }
-                // Reregister with new interest.
-                self.pty.reregister(&self.poll, interest, poll_opts).unwrap();
             }
 
             // The evented instances are not dropped here so deregister them explicitly.

```

alacritty_terminal/src/event_loop.rs
kchibisov: Could you move it and related impl the bottom of the file.

alacritty_terminal/src/event_loop.rs
```diff
@@ -158,28 +197,28 @@ where
         hold: bool,
         ref_test: bool,
     ) -> EventLoop<T, U> {
-        let (tx, rx) = channel::channel();
+        let (tx, rx) = mpsc::channel();
```
kchibisov, line 200: Could you move `impl` below `EventLoop` struct definition while at it?

kchibisov (changes requested):
It would be nice if you could provide the https://github.com/alacritty/vtebench/ results from windows platform wrt changing that or anonymous pipe size(it's zero now, which is meant default).

If running `vtebench` results in no change it's likely that windows is just bottlenecks on contpy. It was like that in the past for example and benchmark which runs like 10ms on the same machine was running 1 second(or even more) with conpty.

alacritty_terminal/src/event_loop.rs
kchibisov: it's still not there, it should be at line 59(60).

alacritty_terminal/src/event_loop.rs
```diff
@@ -336,29 +327,29 @@ where
...
-                                break 'event_loop;
-                            }
-                        },
+                // Handle channel events, if there are any.
+                if !self.drain_recv_channel(&mut state) {
+                    break;
+                }
 
-                        token if token == self.pty.child_event_token() => {
+                for event in events.iter() {
```
kchibisov, line 350: Can't we `drain` here instead of `clear` above?

notgull:
Sorry that I haven't updated in a while! It's a combination of starting a new job, and the fact that if I look at the `polling-utils` code again I think I will die of a heart attack. I'll finish this at some point in the near future.

chrisduerr:
No worries, take all the time you need.

kchibisov:
We may keep the windows part in tree for a while, it's not like I really mind it.

notgull:
> We may keep the windows part in tree for a while, it's not like I really mind it.

Ok, sounds good. Writing the Windows part of the code in a generic way is an effort that I don't want to undertake for the time being.

I've rebased on the latest `master`

alacritty_terminal/src/tty/windows/conpty.rs
kchibisov: Could you drop this comment?

kchibisov (changes requested):
Just one minor comment from me.

I'll also need to retest on Windows again just in case.

kchibisov:
@perfbot

kchibisov:
Windows ack.

kchibisov (approved):

alacritty_terminal/src/tty/unix.rs
```diff
@@ -2,26 +2,34 @@
...
 use crate::config::PtyConfig;
 use crate::event::{OnResize, WindowSize};
 use crate::tty::{ChildEvent, EventedPty, EventedReadWrite};
 
+// We use the same interest for reading and writing.
+pub(crate) const PTY_READ_TOKEN: usize = 0;
+pub(crate) const PTY_WRITE_TOKEN: usize = 0;
+
+// Interest in new child events.
+pub(crate) const PTY_CHILD_EVENT_TOKEN: usize = 1;
```
chrisduerr, line 31: Is there a way to get these tokens in a way that is actually maintainable? Figuring out which token ID is free when it's just constants all over the place is basically impossible.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -267,9 +282,9 @@ pub fn new(config: &PtyConfig, window_size: WindowSize, window_id: u64) -> Resul
             let mut pty = Pty {
                 child,
                 file: unsafe { File::from_raw_fd(master) },
-                token: mio::Token::from(0),
+                token: 0,
                 signals,
-                signals_token: mio::Token::from(0),
+                signals_token: 0,
```
chrisduerr, line 287: These are just arbitrary token values unrelated to our read/write tokens I'd assume. Not great they're also zero tokens.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -302,89 +317,85 @@ impl EventedReadWrite for Pty {
     #[inline]
     fn register(
         &mut self,
-        poll: &mio::Poll,
-        token: &mut dyn Iterator<Item = mio::Token>,
-        interest: mio::Ready,
-        poll_opts: mio::PollOpt,
+        poll: &Arc<polling::Poller>,
+        mut interest: polling::Event,
+        poll_opts: polling::PollMode,
```
chrisduerr, line 322: Not a fan of full imports again, this applies all over the place.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -302,89 +317,85 @@ impl EventedReadWrite for Pty {
...
     ) -> Result<()> {
-        self.token = token.next().unwrap();
-        poll.register(&EventedFd(&self.file.as_raw_fd()), self.token, interest, poll_opts)?;
+        self.token = PTY_READ_TOKEN;
```
chrisduerr, line 324: Is there any point in this? It'll always be 0 anyway.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -302,89 +317,85 @@ impl EventedReadWrite for Pty {
...
-        self.signals.pending().next().and_then(|signal| {
-            if signal != sigconsts::SIGCHLD {
-                return None;
-            }
+        // See if there has been a SIGCHLD.
+        let mut buf = [0u8; 1];
+        match self.signals.read(&mut buf) {
+            Ok(_) => {
+                // We received the signal.
+            },
```
chrisduerr, line 378:
```suggestion
            // We received the signal.
            Ok(_) => (),
```

alacritty_terminal/src/tty/unix.rs
```diff
@@ -302,89 +317,85 @@ impl EventedReadWrite for Pty {
...
-                    None
-                },
-                Ok(None) => None,
-                Ok(_) => Some(ChildEvent::Exited),
-            }
-        })
-    }
+            Err(e) => {
+                // We didn't receive the signal.
+                // Either an error occurred or we got EAGAIN. We don't really care which one.
```
chrisduerr, line 382:
```suggestion
            Err(e) => {
```

alacritty_terminal/src/tty/windows/blocking.rs
```diff
@@ -0,0 +1,276 @@
...
+
+use piper::{pipe, Reader, Writer};
+use polling::os::iocp::{CompletionPacket, PollerIocpExt};
+use polling::{Event, PollMode, Poller};
+
+use std::io::prelude::*;
+use std::marker::PhantomData;
+use std::sync::{Arc, Mutex};
+use std::task::{Context, Poll, Wake, Waker};
+use std::{io, thread};
```
chrisduerr, line 11: STD imports go at the top.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -220,8 +221,8 @@ pub fn new(config: &PtyConfig, window_size: WindowSize) -> Option<Pty> {
         }
     }
 
-    let conin = EventedAnonWrite::new(conin);
-    let conout = EventedAnonRead::new(conout);
+    let conin = UnblockedWriter::new(conin, PIPE_CAPACITY);
+    let conout = UnblockedReader::new(conout, PIPE_CAPACITY);
```
chrisduerr, line 225: Are we double-buffering the PTY on Windows? That seems pretty inefficient.

Have you done any performance testing on Windows?

chrisduerr (changes requested):

alacritty_terminal/src/tty/windows/conpty.rs
kchibisov: That's sort of true, but it was the case before, nothing changed here.

alacritty_terminal/src/event_loop.rs
```diff
@@ -414,19 +299,152 @@ where
...
-            let _ = self.poll.deregister(&self.rx);
             let _ = self.pty.deregister(&self.poll);
 
             (self, state)
         })
     }
 }
+
+/// Helper type which tracks how much of a buffer has been written.
+struct Writing {
```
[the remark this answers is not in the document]
kchibisov: This was the case before, so if you don't like the name you can rename it on master or suggest a name change.

alacritty_terminal/src/event_loop.rs
```diff
@@ -374,17 +261,15 @@ where
                             }
                         },
 
-                        token
-                            if token == self.pty.read_token()
-                                || token == self.pty.write_token() =>
-                        {
-                            #[cfg(unix)]
+                        token if token == tty::PTY_READ_TOKEN || token == tty::PTY_WRITE_TOKEN => {
+                            /* FIXME: once polling supports extra reported modes bring it back.
```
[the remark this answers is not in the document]
kchibisov: In general, I don't think it provides that much value given that it worked only on Linux before.

But see https://github.com/smol-rs/polling/issues/112 , I remember that it'll require a breaking bump of polling.

alacritty_terminal/Cargo.toml
```diff
@@ -36,10 +35,9 @@ vte = { version = "0.11.1", default-features = false, features = ["ansi", "serde
 [target.'cfg(unix)'.dependencies]
 nix = { version = "0.26.2", default-features = false, features = ["term"] }
 signal-hook = "0.3.10"
-signal-hook-mio = { version = "0.2.1", features = ["support-v0_6"] }
 
 [target.'cfg(windows)'.dependencies]
-mio-anonymous-pipes = "0.2"
+piper = "0.2"
```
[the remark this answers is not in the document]
kchibisov: It's this https://github.com/smol-rs/piper/ . It's maintained.

alacritty_terminal/src/event_loop.rs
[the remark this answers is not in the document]
chrisduerr: I'm pretty sure without this it might cause occasional errors on exit?

alacritty_terminal/Cargo.toml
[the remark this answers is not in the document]
chrisduerr: I see a PR from two months ago with a trivial change that has seen zero action. So I have my doubts.

alacritty_terminal/src/event_loop.rs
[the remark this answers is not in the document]
kchibisov: It could cause an error when you try to read again, but we have those errors on anything other than linux already (if we even have them). You may even notice that this change was added as enhancement in the past.

alacritty_terminal/Cargo.toml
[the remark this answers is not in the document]
kchibisov: @chrisduerr from dependabot....

alacritty_terminal/src/event_loop.rs
```diff
@@ -313,17 +205,13 @@ where
...
             let mut buf = [0u8; READ_BUFFER_SIZE];
 
-            let mut tokens = (0..).map(Into::into);
-
-            let poll_opts = PollOpt::edge() | PollOpt::oneshot();
-
-            let channel_token = tokens.next().unwrap();
-            self.poll.register(&self.rx, channel_token, Ready::readable(), poll_opts).unwrap();
+            let poll_opts = polling::PollMode::Level;
+            let mut interest = polling::Event::readable(0);
```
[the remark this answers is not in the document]
notgull: The zero is just a placeholder for the token to be interested in.

alacritty_terminal/Cargo.toml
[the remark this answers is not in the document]
notgull: `piper` was extracted somewhat recently from `blocking`, which is already used in production in various places. The only reason no changes have been made is because `piper` has already been scrutinized very closely through its prior use in `blocking`. Not to mention it's not used directly in many places at the moment, so there's no issues to be filed (and therefore no changes to be made). For more context see [this](https://github.com/smol-rs/async-channel/issues/1) issue.

alacritty_terminal/src/event_loop.rs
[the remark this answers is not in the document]
notgull: As @kchibisov commented, this is just an optimization for now. It can be added back once we add greater event granularity to `polling`.

alacritty_terminal/src/tty/unix.rs
notgull: It clarifies that the token is representative of the read token.

alacritty_terminal/Cargo.toml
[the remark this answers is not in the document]
chrisduerr: > @chrisduerr from dependabot....

Dependabot doesn't make random changes unless requested. If requested and not acted upon, there's clearly *some* maintenance issues.

alacritty_terminal/src/event_loop.rs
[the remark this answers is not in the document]
chrisduerr: Either way I don't like the documentation in the comments. If this is something we want to track, it should be removed from the code and added to the issue tracker.

alacritty_terminal/src/tty/unix.rs
chrisduerr: How is that any different from just initializing it to the read token from the start?

alacritty_terminal/src/tty/unix.rs
notgull: We could just have an `Iterator` like we did in the old code. This was how it was done originally but it was requested that we move to this constant based approach instead.

alacritty_terminal/src/tty/unix.rs
notgull: Yes, these are empty tokens.

alacritty_terminal/src/tty/unix.rs
notgull: This mirrors how it was done in the `mio` code. If you want I can change it to be initialized in this way from the start.

ognevnydemon:
is there any progress? just for interest

notgull:
I've updated to the new breaking release of `polling`, version 3.0. This adds some new (checked) unsafe code to this crate.

alacritty_terminal/src/event_loop.rs
```diff
@@ -158,28 +66,28 @@ where
...
             terminal,
             event_proxy,
             hold,
             ref_test,
         }
     }
 
-    pub fn channel(&self) -> Sender<Msg> {
-        self.tx.clone()
+    pub fn channel(&self) -> EventLoopSender {
```
nixpulvis, line 82: Might not be a bad idea to rename this function to be a bit more clear.

Maybe `send_channel`?

alacritty_terminal/src/event_loop.rs
```diff
@@ -313,17 +207,15 @@ where
...
+            let mut interest = PollingEvent::readable(0);
 
             // Register TTY through EventedRW interface.
-            self.pty.register(&self.poll, &mut tokens, Ready::readable(), poll_opts).unwrap();
+            unsafe {
+                self.pty.register(&self.poll, interest, poll_opts).unwrap();
+            }
 
-            let mut events = Events::with_capacity(1024);
+            let mut events = Events::with_capacity(NonZeroUsize::new(128).unwrap());
```
nixpulvis, line 218: Why the large decrease in size?

alacritty_terminal/src/event_loop.rs
```diff
@@ -337,29 +229,29 @@ where
...
+                if let Err(err) = self.poll.wait(&mut events, timeout) {
                     match err.kind() {
                         ErrorKind::Interrupted => continue,
                         _ => panic!("EventLoop polling error: {err:?}"),
                     }
                 }
 
                 // Handle synchronized update timeout.
-                if events.is_empty() {
+                if events.is_empty() && self.rx.peek().is_none() {
```
nixpulvis, line 241: Is there no way to setup the abstractions so that the calls to `poll` and the event buffer are always in sync, thus avoiding this extra peak condition?

alacritty_terminal/Cargo.toml
```diff
@@ -36,10 +35,9 @@ vte = { version = "0.12.0", default-features = false, features = ["ansi", "serde
...
```
chrisduerr, line 40:
```suggestion
piper = "0.2.1"
```

alacritty_terminal/src/event_loop.rs
chrisduerr: Not a bad suggestion, but won't block on this since it's unrelated to this PR.

alacritty_terminal/src/event_loop.rs
chrisduerr: +1 to this question. I don't really mind, but I'd like a reason stated.

alacritty_terminal/src/event_loop.rs
```diff
@@ -374,17 +266,10 @@ where
...
-                            if token == self.pty.read_token()
-                                || token == self.pty.write_token() =>
-                        {
-                            #[cfg(unix)]
-                            if UnixReady::from(event.readiness()).is_hup() {
-                                // Don't try to do I/O on a dead PTY.
-                                continue;
-                            }
+                        token if token == tty::PTY_READ_TOKEN || token == tty::PTY_WRITE_TOKEN => {
+                            // FIXME: once polling supports extra reported modes bring it back.
```
chrisduerr, line 270: This should be an issue. Comments are useless and just get lost.

Also missing what "it" is here. Bring what back?

alacritty_terminal/src/tty/mod.rs
```diff
@@ -22,20 +25,15 @@ pub trait EventedReadWrite {
...
-        _: mio::Ready,
-        _: mio::PollOpt,
-    ) -> io::Result<()>;
-    fn reregister(&mut self, _: &mio::Poll, _: mio::Ready, _: mio::PollOpt) -> io::Result<()>;
-    fn deregister(&mut self, _: &mio::Poll) -> io::Result<()>;
+    /// # Safety
+    ///
+    /// The underlying sources must outlive their registration in the `Poller`.
+    unsafe fn register(&mut self, _: &Arc<Poller>, _: Event, _: PollMode) -> io::Result<()>;
+    fn reregister(&mut self, _: &Arc<Poller>, _: Event, _: PollMode) -> io::Result<()>;
```
chrisduerr, line 32: I'd assume that reregister has mostly the same properties as register. Is this safety note not also true for reregister?

alacritty_terminal/src/tty/unix.rs
chrisduerr: +1

alacritty_terminal/src/tty/unix.rs
chrisduerr: Would probably be cleaner this way. The mio code wasn't perfect, there's no need to keep things consistent for the sake of being similar to the mio code when there's a better solution available.

chrisduerr (changes requested):

alacritty_terminal/src/tty/unix.rs
```diff
@@ -300,91 +316,80 @@ impl EventedReadWrite for Pty {
...
-                    error!("Error checking child process termination: {}", e);
-                    None
-                },
-                Ok(None) => None,
-                Ok(_) => Some(ChildEvent::Exited),
-            }
-        })
-    }
+            },
+        }
```
kchibisov, line 382: Looks like `if let Err(err)`

alacritty_terminal/src/tty/unix.rs
```diff
@@ -300,91 +316,80 @@ impl EventedReadWrite for Pty {
...
 
-    #[inline]
-    fn child_event_token(&self) -> mio::Token {
-        self.signals_token
+        // Match on the child process.
+        match self.child.try_wait() {
+            Err(e) => {
+                error!("Error checking child process termination: {}", e);
```
kchibisov, line 387:
```suggestion
            Err(err) => {
                error!("Error checking child process termination: {err}");
```

kchibisov (changes requested):

alacritty_terminal/src/event_loop.rs
```diff
@@ -337,29 +229,29 @@ where
                 let timeout =
                     handler.sync_timeout().map(|st| st.saturating_duration_since(Instant::now()));
 
-                if let Err(err) = self.poll.poll(&mut events, timeout) {
+                events.clear();
```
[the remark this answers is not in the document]
notgull: `polling` v3.0.0 has a separate `Events` structure, with only `iter()` and `clear()`.

alacritty_terminal/src/tty/mod.rs
notgull: "register" clones the raw file descriptor into the `Poller`, which is an unsafe operation. "reregister" does not do this.

Cargo.lock
```diff
@@ -365,42 +357,40 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.3.3"
+version = "4.4.2"
```
kchibisov, line 360: Could you please, update the lock file only by the means of added dependencies? So you actually won't have broken tests due to clap change? It'll also be easier for us to test things and see what was added.

You could though, remove dups if you see a way to do so.

kchibisov (approved):
Tested macOS/Windows

chrisduerr (approved):

chrisduerr:
Thanks for working on this.

PR #6847, opened 2023-04-08
Add missing terminfo capabilities

chrisduerr:
This adds capabilities for focus in/out (XF, kxIN, kxOUT) and bracketed paste (BD, BE, PE, PS).


extra/alacritty.info
```diff
@@ -108,3 +108,5 @@ alacritty+common|base fragment for alacritty,
     hs, dsl=\E]2;\007, fsl=^G, tsl=\E]2;,
     Smulx=\E[4\:%p1%dm,
     Sync=\EP=%p1%ds\E\\,
+    XF, kxIN=\e[I, kxOUT=\e[0,
```
kchibisov, line 111:
```suggestion
    XF, kxIN=\E[I, kxOUT=\E[0,
```

Maybe put `XF` with other caps above, but I don't mind it being here?

kchibisov (changes requested):
Do we need any `CHANGELOG` entry for that?

extra/alacritty.info
chrisduerr: I like having all the things that go together on the same line. Makes it nice and easy to blame too.

kchibisov (approved):

PR #6849, opened 2023-04-09
Improve keybinding flexibility

### LINKED ISSUE #4073: Keybinding Specific Search

chrisduerr:
Previously keybindings for the same key would be evaluated in order using the mode set while evaluating the first binding. This makes it impossible to combine keybindings across mode changes when the action is restricted to a mode entered through a previous binding.

To make it easier to combine keybindings, this patch reevaluates the mode for every keybinding.

To further improve the ability of users to write macros like these, the `chars` action has also been rewired to go through `reiceived_char`, allowing it to automatically work for Alacritty's UI (i.e. search), rather than just writing to the PTY.

Closes #4073.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -218,56 +218,6 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         self.notifier.notify(val);
     }
 
-    fn received_char(&mut self, c: char) {
-        // Don't insert chars when we have IME running.
-        if self.display().ime.preedit().is_some() {
-            return;
-        }
-
-        // Handle hint selection over anything else.
-        if self.display().hint_state.active() && !*self.suppress_chars {
-            self.hint_input(c);
-            return;
-        }
-
-        // Pass keys to search and ignore them during `suppress_chars`.
-        let search_active = self.search_active();
-        if *self.suppress_chars || search_active || self.terminal().mode().contains(TermMode::VI) {
-            if search_active && !*self.suppress_chars {
-                self.search_input(c);
-            }
-
-            return;
-        }
-
-        self.on_typing_start();
-
-        if self.terminal().grid().display_offset() != 0 {
-            self.scroll(Scroll::Bottom);
-        }
-        self.clear_selection();
-
-        let utf8_len = c.len_utf8();
-        let mut bytes = vec![0; utf8_len];
-        c.encode_utf8(&mut bytes[..]);
-
-        #[cfg(not(target_os = "macos"))]
-        let alt_send_esc = true;
-
-        // Don't send ESC when `OptionAsAlt` is used. This doesn't handle
-        // `Only{Left,Right}` variants due to inability to distinguish them.
-        #[cfg(target_os = "macos")]
-        let alt_send_esc = self.config().window.option_as_alt != OptionAsAlt::None;
-
-        if alt_send_esc && *self.received_count() == 0 && self.modifiers().alt() && utf8_len == 1 {
-            bytes.insert(0, b'\x1b');
-        }
-
-        self.write_to_pty(bytes);
-
-        *self.received_count() += 1;
-    }
-
     /// Request a redraw.
     #[inline]
     fn mark_dirty(&mut self) {
@@ -774,9 +724,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
                 self.clipboard.store(ClipboardType::Clipboard, text);
             },
             // Write the text to the PTY/search.
-            HintAction::Action(HintInternalAction::Paste) => {
-                self.paste(&text);
-            },
+            HintAction::Action(HintInternalAction::Paste) => self.paste(&text, true),
             // Select the text.
             HintAction::Action(HintInternalAction::Select) => {
                 self.start_selection(SelectionType::Simple, *hint_bounds.start(), Side::Left);
@@ -832,13 +780,25 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         }
     }
 
+    /// Handle beginning of terminal text input.
+    fn on_terminal_input_start(&mut self) {
+        self.on_typing_start();
+        self.clear_selection();
+
+        if self.terminal().grid().display_offset() != 0 {
+            self.scroll(Scroll::Bottom);
+        }
+    }
+
     /// Paste a text into the terminal.
-    fn paste(&mut self, text: &str) {
+    fn paste(&mut self, text: &str, bracketed: bool) {
         if self.search_active() {
             for c in text.chars() {
                 self.search_input(c);
             }
-        } else if self.terminal().mode().contains(TermMode::BRACKETED_PASTE) {
+        } else if bracketed && self.terminal().mode().contains(TermMode::BRACKETED_PASTE) {
+            self.on_terminal_input_start();
+
             self.write_to_pty(&b"\x1b[200~"[..]);
 
             // Write filtered escape sequences.
@@ -851,6 +811,8 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
 
             self.write_to_pty(&b"\x1b[201~"[..]);
         } else {
+            self.on_terminal_input_start();
+
             // In non-bracketed (ie: normal) mode, terminal applications cannot distinguish
             // pasted data from keystrokes.
             // In theory, we should construct the keystrokes needed to produce the data we are
@@ -1338,7 +1300,7 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                         self.key_input(input);
                     },
                     WindowEvent::ModifiersChanged(modifiers) => self.modifiers_input(modifiers),
-                    WindowEvent::ReceivedCharacter(c) => self.ctx.received_char(c),
+                    WindowEvent::ReceivedCharacter(c) => self.received_char(c),
                     WindowEvent::MouseInput { state, button, .. } => {
                         self.ctx.window().set_mouse_visible(true);
                         self.mouse_input(state, button);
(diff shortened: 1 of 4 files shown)


kchibisov (changes requested):
You're missing changelog entry for that.

kchibisov:
Seems like the repeat of normal bindings broke, like `Backspace`, `F2`, and similar. They are executing only once now.

alacritty/src/event.rs
```diff
@@ -834,12 +782,15 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
     }
 
     /// Paste a text into the terminal.
-    fn paste(&mut self, text: &str) {
+    fn paste(&mut self, text: &str, bracketed: bool) {
+        self.on_typing_start();
+        self.clear_selection();
```
kchibisov, line 787: From what I can see, we don't call them when we have input.

I think something like that should reflect the logic we have inside the `received_char`.

```rust
    /// Paste a text into the terminal.
    fn paste(&mut self, text: &str, bracketed: bool) {
        if self.search_active() {
            for c in text.chars() {
                self.search_input(c);
            }
            return;
        }

        // NOTE: probably refactor to have the same logic with `received_char`.
        self.on_typing_start();

        if self.terminal().grid().display_offset() != 0 {
            self.scroll(Scroll::Bottom);
        }
        self.clear_selection();
        
        if bracketed && self.terminal().mode().contains(TermMode::BRACKETED_PASTE) {
            self.write_to_pty(&b"\x1b[200~"[..]);

            // Write filtered escape sequences.
            //
            // We remove `\x1b` to ensure it's impossible for the pasted text to write the bracketed
            // paste end escape `\x1b[201~` and `\x03` since some shells incorrectly terminate
            // bracketed paste on its receival.
            let filtered = text.replace(['\x1b', '\x03'], "");
            self.write_to_pty(filtered.into_bytes());

            self.write_to_pty(&b"\x1b[201~"[..]);
        } else {
            // In non-bracketed (ie: normal) mode, terminal applications cannot distinguish
            // pasted data from keystrokes.
            // In theory, we should construct the keystrokes needed to produce the data we are
            // pasting... since that's neither practical nor sensible (and probably an impossible
            // task to solve in a general way), we'll just replace line breaks (windows and unix
            // style) with a single carriage return (\r, which is what the Enter key produces).
            self.write_to_pty(text.replace("\r\n", "\r").replace('\n', "\r").into_bytes());
        }
    }

```

chrisduerr: > From what I can see, we don't call them when we have input.

What do you mean by this?

kchibisov: I think I've meant:

From what I can see, we don't **always** call them when we have input.

The behavior I've suggested should reflect the `received_char` function.

alacritty/src/event.rs
```diff
@@ -783,12 +782,15 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
-    fn paste(&mut self, text: &str) {
+    fn paste(&mut self, text: &str, bracketed: bool) {
         if self.search_active() {
             for c in text.chars() {
                 self.search_input(c);
             }
-        } else if self.terminal().mode().contains(TermMode::BRACKETED_PASTE) {
+        } else if bracketed && self.terminal().mode().contains(TermMode::BRACKETED_PASTE) {
+            self.on_typing_start();
+            self.clear_selection();
```
kchibisov, line 792: What's the purpose to ignore the early return to have less duplicate code? The check for non-zero before scrolling was pulled to not damage entire screen, but I there could be a logic inside the scrolling itself to not damage when nothing changed, I don't remember.

chrisduerr: There is no early return here. What are you talking about?

kchibisov: I'm talking about, why you don't use early return to remove like 4 lines of duplicates.

chrisduerr: Early return where exactly?

kchibisov: I've posted the original diff with the early return though, and all the code needed.

https://github.com/alacritty/alacritty/pull/6849#discussion_r1161350483

The only issue is that `NOTE` could be extracted, so it's the same in `received_char` and `paste`.

chrisduerr: That original code is much worse than what I've implemented here though.

kchibisov: I just don't like that we have two else branches with near the same code right now, which you could call right before taking both of the branches, this is just error prone when you update only one and forget about the other. I'd maybe even refactor this function logic out and use the same in `received_char` as well, to not break it.

chrisduerr: Your suggestion just makes things more verbose and more fragile by increasing the control-flow complexity though.

chrisduerr: I just had it before the branching before, which behaved the same way as far as I could tell. But you wanted to change that so this is imo the best alternative.

kchibisov: Maybe something like this then:

```diff
diff --git a/alacritty/src/event.rs b/alacritty/src/event.rs
index 70c88a52..6c34ea61 100644
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -692,6 +692,17 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         }
     }
 
+    /// Handle keyboard typing into the terminal.
+    ///
+    /// This function must be called when the user typing results in writing to the
+    /// pty.
+    #[inline]
+    fn on_terminal_typing_start(&mut self) {
+        self.on_typing_start();
+        self.scroll(Scroll::Bottom);
+        self.clear_selection();
+    }
+
     /// Process a new character for keyboard hints.
     fn hint_input(&mut self, c: char) {
         if let Some(hint) = self.display.hint_state.keyboard_input(self.terminal, c) {
@@ -788,8 +799,7 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
                 self.search_input(c);
             }
         } else if bracketed && self.terminal().mode().contains(TermMode::BRACKETED_PASTE) {
-            self.on_typing_start();
-            self.clear_selection();
+            self.on_terminal_typing_start();
 
             self.write_to_pty(&b"\x1b[200~"[..]);
 
@@ -802,11 +812,8 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             self.write_to_pty(filtered.into_bytes());
 
             self.write_to_pty(&b"\x1b[201~"[..]);
-
-            self.scroll(Scroll::Bottom);
         } else {
-            self.on_typing_start();
-            self.clear_selection();
+            self.on_terminal_typing_start();
 
             // In non-bracketed (ie: normal) mode, terminal applications cannot distinguish
             // pasted data from keystrokes.
@@ -815,8 +822,6 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             // task to solve in a general way), we'll just replace line breaks (windows and unix
             // style) with a single carriage return (\r, which is what the Enter key produces).
             self.write_to_pty(text.replace("\r\n", "\r").replace('\n', "\r").into_bytes());
-
-            self.scroll(Scroll::Bottom);
         }
     }
 
diff --git a/alacritty/src/input.rs b/alacritty/src/input.rs
index 4e256672..95a9e207 100644
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -116,6 +116,7 @@ pub trait ActionContext<T: EventListener> {
     fn search_direction(&self) -> Direction;
     fn search_active(&self) -> bool;
     fn on_typing_start(&mut self) {}
+    fn on_terminal_typing_start(&mut self) {}
     fn toggle_vi_mode(&mut self) {}
     fn hint_input(&mut self, _character: char) {}
     fn trigger_hint(&mut self, _hint: &HintMatch) {}
@@ -993,12 +994,7 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
             return;
         }
 
-        self.ctx.on_typing_start();
-
-        if self.ctx.terminal().grid().display_offset() != 0 {
-            self.ctx.scroll(Scroll::Bottom);
-        }
-        self.ctx.clear_selection();
+        self.ctx.on_terminal_typing_start();
 
         let utf8_len = c.len_utf8();
         let mut bytes = vec![0; utf8_len];

```

I'm just worried, because we already had in the past breakages in terminal input due to inconsistent handling across the code base, such as, where latency regressed by 1 frame due to dirty bit altering early on or not doing various checks.

alacritty/src/event.rs
```diff
@@ -800,14 +802,21 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
...
+
             // In non-bracketed (ie: normal) mode, terminal applications cannot distinguish
             // pasted data from keystrokes.
             // In theory, we should construct the keystrokes needed to produce the data we are
             // pasting... since that's neither practical nor sensible (and probably an impossible
             // task to solve in a general way), we'll just replace line breaks (windows and unix
             // style) with a single carriage return (\r, which is what the Enter key produces).
             self.write_to_pty(text.replace("\r\n", "\r").replace('\n', "\r").into_bytes());
+
+            self.scroll(Scroll::Bottom);
```
kchibisov, line 819:
```suggestion
```

alacritty/src/event.rs
```diff
@@ -800,14 +802,21 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             self.write_to_pty(filtered.into_bytes());
 
             self.write_to_pty(&b"\x1b[201~"[..]);
+
+            self.scroll(Scroll::Bottom);
```
kchibisov, line 806:
```suggestion
```

kchibisov (changes requested):

kchibisov (approved):

PR #6876, opened 2023-04-20
feat: update default macOS new window bindings

axieax:
This PR addresses the unconventional bindings introduced by #6648, adding the following macOS bindings:

- `Cmd + T` for `CreateNewWindow` (mimics opening a new tab if system-enabled)
- `Cmd + N` for `SpawnNewInstance` (opens a separate Alacritty window)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - Mode-specific bindings can now be bound in any mode for easier macros
+- Update default macOS bindings for `Cmd`+`N` and `Cmd`+`T` to open windows as expected
 
 ### Fixed
 
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -875,7 +875,8 @@
   #- { key: M,              mods: Command,                    action: Minimize              }
   #- { key: Q,              mods: Command,                    action: Quit                  }
   #- { key: W,              mods: Command,                    action: Quit                  }
-  #- { key: N,              mods: Command,                    action: CreateNewWindow       }
+  #- { key: T,              mods: Command,                    action: CreateNewWindow       }
+  #- { key: N,              mods: Command,                    action: SpawnNewInstance       }
   #- { key: F,              mods: Command|Control,            action: ToggleFullscreen      }
   #- { key: F,              mods: Command, mode: ~Search,     action: SearchForward         }
   #- { key: B,              mods: Command, mode: ~Search,     action: SearchBackward        }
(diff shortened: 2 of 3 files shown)


utkarshgupta137:
This is the behavior I expect since I've changed the following setting:
image

With these keybindings, Cmd+T opens a new tab & Cmd+N opens a new window like every other macOS app.
I don't think changing these bindings will affect people who've not changed the above setting.

chrisduerr (changes requested):
`CreateNewWindow` can open a new window just like `SpawnNewInstance` does.

utkarshgupta137:
> `CreateNewWindow` can open a new window just like `SpawnNewInstance` does.

If we change the above-mentioned setting in macOS, `CreateNewWindow` opens a new OS tab whereas `SpawnNewInstance` opens a new OS window. So there is a difference.

kchibisov:
Be aware that `SpawnNewInstance` is not something macOS applications do at all.

chrisduerr:
> If we change the above-mentioned setting in macOS, CreateNewWindow opens a new OS tab whereas SpawnNewInstance opens a new OS window. So there is a difference.

Pretending that's the difference between the two actions is just a gross misrepresentation of what they do.

axieax:
Is there a way to specifically open a new MacOS tab for new windows (perhaps as a window option)?

chrisduerr:
> Is there a way to specifically open a new MacOS tab for new windows (perhaps as a window option)?

There is not, which is why there is no such binding in Alacritty by default.

utkarshgupta137:
@axieax Are you talking about something like this:
image

If so, change the following setting in system preferences:
image

Then bind `Cmd + T` to `CreateNewWindow` (& maybe `Cmd + N` to `SpawnNewInstance` to open a separate macOS window). See alacritty.yml in this PR. 

Bonus tip: You can use `Cmd + Shift + [ / ]` to go to the left or right tab. (This shortcut works for most macOS apps)

axieax:
@utkarshgupta137 yep that's the tab view I'm referring to, and it would be nice for there to be a way to specifically target whether to open a "tab" or a "window" without having to change the global system settings just for this one application.

utkarshgupta137:
> @utkarshgupta137 yep that's the tab view I'm referring to, and it would be nice for there to be a way to specifically target whether to open a "tab" or a "window" without having to change the global system settings just for this one application.

The "tabs" you're seeing in alacritty is a feature of the macOS's window manager. Alacritty itself doesn't support tabs. I'm not sure if macOS supports changing this on a per-app basis. I think this should work: https://apple.stackexchange.com/a/458847/387376

axieax:
Got this to work now, thanks for all the help @utkarshgupta137!! 🎉🎉

PR #6898, opened 2023-05-02
Add `ALACRITTY_EXTRA_LOG_TARGETS` env variable

kchibisov:
This should help with troubleshooting the dependency crates like winit.

### DIFF
--- a/alacritty/src/logging.rs
+++ b/alacritty/src/logging.rs
@@ -13,6 +13,7 @@ use std::time::Instant;
 use std::{env, process};
 
 use log::{self, Level, LevelFilter};
+use once_cell::sync::Lazy;
 use winit::event_loop::EventLoopProxy;
 
 use alacritty_terminal::config::LOG_TARGET_CONFIG;
@@ -26,6 +27,17 @@ pub const LOG_TARGET_IPC_CONFIG: &str = "alacritty_log_ipc_config";
 
 /// Name for the environment variable containing the log file's path.
 const ALACRITTY_LOG_ENV: &str = "ALACRITTY_LOG";
+
+/// Name for the environment varibale containing extra logging targets.
+///
+/// The targets are semicolon separated.
+const ALACRITTY_EXTRA_LOG_TARGETS_ENV: &str = "ALACRITTY_EXTRA_LOG_TARGETS";
+
+/// User configurable extra log targets to include.
+static EXTRA_LOG_TARGETS: Lazy<Vec<String>> = Lazy::new(|| {
+    env::var(ALACRITTY_EXTRA_LOG_TARGETS_ENV)
+        .map_or(Vec::new(), |targets| targets.split(';').map(ToString::to_string).collect())
+});
 
 /// List of targets which will be logged by Alacritty.
 const ALLOWED_TARGETS: &[&str] = &[
@@ -167,7 +179,7 @@ fn create_log_message(record: &log::Record<'_>, target: &str, start: Instant) ->
 fn is_allowed_target(level: Level, target: &str) -> bool {
     match (level, log::max_level()) {
         (Level::Error, LevelFilter::Trace) | (Level::Warn, LevelFilter::Trace) => true,
-        _ => ALLOWED_TARGETS.contains(&target),
+        _ => ALLOWED_TARGETS.contains(&target) || EXTRA_LOG_TARGETS.iter().any(|t| t == target),
     }
 }
 


alacritty/src/logging.rs
```diff
@@ -167,7 +180,10 @@ fn create_log_message(record: &log::Record<'_>, target: &str, start: Instant) ->
 fn is_allowed_target(level: Level, target: &str) -> bool {
     match (level, log::max_level()) {
         (Level::Error, LevelFilter::Trace) | (Level::Warn, LevelFilter::Trace) => true,
-        _ => ALLOWED_TARGETS.contains(&target),
+        _ => {
+            ALLOWED_TARGETS.contains(&target)
+                || EXTRA_LOG_TARGETS.as_ref().map_or(false, |elt| elt.contains(&target))
```
chrisduerr, line 185:
```suggestion
                || EXTRA_LOG_TARGETS.as_ref().map_or(false, |targets| targets.contains(&target))
```

Took me too long to figure out what the hell an "elt" is supposed to be lol.

alacritty/src/logging.rs
```diff
@@ -27,6 +28,18 @@ pub const LOG_TARGET_IPC_CONFIG: &str = "alacritty_log_ipc_config";
...
+///
+/// The targets are semicolon separated.
+const ALACRITTY_EXTRA_LOG_TARGETS_ENV: &str = "ALACRITTY_EXTRA_LOG_TARGETS";
+
+/// User configurable extra log targets to include.
+static EXTRA_LOG_TARGETS: Lazy<Option<Vec<&str>>> = Lazy::new(|| {
+    // The logging should be valid for the static lifetime.
+    let targets = Box::leak(env::var(ALACRITTY_EXTRA_LOG_TARGETS_ENV).ok()?.into_boxed_str());
+    Some(targets.split(';').collect())
+});
```
chrisduerr, line 41: Can we not use a String here?

chrisduerr (changes requested):

alacritty/src/logging.rs
kchibisov: Do you mean using `String` as a whole and search in split string? I don't understand what do you mean exactly.

alacritty/src/logging.rs
kchibisov: But they are on the same line...

alacritty/src/logging.rs
chrisduerr: I'm just a bit confused about the necessity of the boxed string leakage.

kchibisov: You can't leak the string in rust directly yet, so you have to do something like that.

chrisduerr: You can use `Lazy<Option<Vec<String>>>` though, right?

kchibisov: I can, but it'll mean to re-allocate everything without changing the semantics. It's just less efficient for the same thing.

alacritty/src/logging.rs
```diff
@@ -167,7 +179,12 @@ fn create_log_message(record: &log::Record<'_>, target: &str, start: Instant) ->
 fn is_allowed_target(level: Level, target: &str) -> bool {
     match (level, log::max_level()) {
         (Level::Error, LevelFilter::Trace) | (Level::Warn, LevelFilter::Trace) => true,
-        _ => ALLOWED_TARGETS.contains(&target),
+        _ => {
+            ALLOWED_TARGETS.contains(&target)
+                || EXTRA_LOG_TARGETS
+                    .as_ref()
+                    .map_or(false, |targets| targets.iter().any(|t| t == target))
```
kchibisov, line 186: `contains` expects `&String` and not `&str`, so we have this mess here.

chrisduerr: Should just change it to `Vec<String>` instead of `Option<Vec<String>>` by using an empty vec by default? That should simplify things.

alacritty/src/logging.rs
```diff
@@ -27,6 +28,20 @@ pub const LOG_TARGET_IPC_CONFIG: &str = "alacritty_log_ipc_config";
...
+/// The targets are semicolon separated.
+const ALACRITTY_EXTRA_LOG_TARGETS_ENV: &str = "ALACRITTY_EXTRA_LOG_TARGETS";
+
+/// User configurable extra log targets to include.
+static EXTRA_LOG_TARGETS: Lazy<Vec<String>> = Lazy::new(|| {
+    if let Ok(targets) = env::var(ALACRITTY_EXTRA_LOG_TARGETS_ENV) {
+        targets.split(';').map(ToString::to_string).collect()
+    } else {
+        Vec::new()
+    }
```
chrisduerr, line 42: map_or?

chrisduerr (approved):

PR #6919, opened 2023-05-11
Add EmptySelection key binding mode

EBADBEEF:
With this, a bind can be configured to trigger based on the presence or absence of a selection.

For example, this config emulates windows terminal style copy paste:
- { mouse: Right,  mode: ~EmptySelection, mods: Shift, action: Copy }
- { mouse: Right,  mode: ~EmptySelection, mods: Shift, action: ClearSelection }
- { mouse: Right,  mode: EmptySelection,  mods: Shift, action: PasteSelection }

In the above example, when no text is selected shift-rightclick will paste. If text is selected, then shift-rightclick will copy and clear the selection.

### DIFF
--- a/alacritty.yml
+++ b/alacritty.yml
@@ -757,6 +757,7 @@
 #    - Search
 #    - Alt
 #    - Vi
+#    - EmptySelection
 #
 #    A `~` operator can be used before a mode to apply the binding whenever
 #    the mode is *not* active, e.g. `~Alt`.
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -770,17 +770,19 @@ bitflags! {
         const ALT_SCREEN          = 0b0000_0100;
         const VI                  = 0b0000_1000;
         const SEARCH              = 0b0001_0000;
+        const EMPTY_SELECTION     = 0b0010_0000;
     }
 }
 
 impl BindingMode {
-    pub fn new(mode: &TermMode, search: bool) -> BindingMode {
+    pub fn new(mode: &TermMode, search: bool, empty_selection: bool) -> BindingMode {
         let mut binding_mode = BindingMode::empty();
         binding_mode.set(BindingMode::APP_CURSOR, mode.contains(TermMode::APP_CURSOR));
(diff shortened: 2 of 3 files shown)


kchibisov:
Given how small the patch is it's probably fine as is (especially given that we have search_active), but I know that @chrisduerr has "reasons" for such thing.

chrisduerr:
I see no good reason why this should be added. Toggling behavior based on a selection existing somewhere is asinine and I have no interest in restricting future development to ensure this behavior keeps working.

EBADBEEF:
@chrisduerr It's not asinine to me, and I bet there are at least a few others like me. I urge you to reconsider including this feature.

Maybe you are repulsed by this suggestion because of stuffing this into "mode" where you never imagined it. Or maybe like a gamer using inverted mouse ;-) It's not perfect, but I argue it still fits.

I think the maintenance burden is minimal because you already have actions dealing with selection. Are you really going to keep clear_selection() but not selection_is_empty()? It seems trivial to keep it along with the current feature set.

Now... how can I justify this feature... it just feels good!

I like clearing the selection on copy because it gives useful feedback. Then what is "copy" supposed to do with no selection? It's a waste! It is intuitive to have it paste. I used to like the X11 primary clipboard (save_to_clipboard option), but these days I prefer not to use it.

I have not used the Vi or Search modes because I'm usually inside of tmux or screen (often remote), and they are not integrated with my system clipboard. Therefore, I reach for the mouse to share the clipboard with outside programs. I could assign right click to copy and middle click to paste, but it can be hard to middle click on a laptop. I will often use the keyboard to paste (with shift-insert) but other times going back and forth copying and pasting it is useful to not have to use the keyboard to paste.

I like this UX from windows terminal, though I dislike it in in other ways like holding shift while copying strips out newlines (how am I supposed to copy from mouse-enabled terminal program without shift?!) and it is really slow and takes >16GB of ram just to build. I saw it was really simple to implement in alacritty so I went for it. I really appreciate how all the necessary parts were there. This was the first rust program I have ever looked at.

Regardless of your decision, thanks for your work. I guess a "no" is better than languishing in purgatory forever, so thanks for being active, too.

PR #6927, opened 2023-05-16
Improve renderer debuggability

kchibisov:
Make the renderer more debuggable by adding extra logging and using `GL_KHR_debug` when running with the debug log level.

### DIFF
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -1,12 +1,12 @@
 use std::collections::HashSet;
 use std::ffi::{CStr, CString};
-use std::fmt;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::{fmt, ptr};
 
 use crossfont::Metrics;
 use glutin::context::{ContextApi, GlContext, PossiblyCurrentContext};
 use glutin::display::{GetGlDisplay, GlDisplay};
-use log::info;
+use log::{debug, error, info, warn, LevelFilter};
 use once_cell::sync::OnceCell;
 
 use alacritty_terminal::index::Point;
@@ -102,13 +102,17 @@ impl Renderer {
             });
         }
 
-        let (version, renderer) = unsafe {
-            let renderer = CStr::from_ptr(gl::GetString(gl::RENDERER) as *mut _);
-            let version = CStr::from_ptr(gl::GetString(gl::SHADING_LANGUAGE_VERSION) as *mut _);
-            (version.to_string_lossy(), renderer.to_string_lossy())
+        let (shader_version, gl_version, renderer) = unsafe {
+            let renderer = CStr::from_ptr(gl::GetString(gl::RENDERER) as *mut _).to_string_lossy();
+            let shader_version =
+                CStr::from_ptr(gl::GetString(gl::SHADING_LANGUAGE_VERSION) as *mut _)
+                    .to_string_lossy();
+            let gl_version = CStr::from_ptr(gl::GetString(gl::VERSION) as *mut _).to_string_lossy();
+            (shader_version, gl_version, renderer)
         };
 
-        info!("Running on {}", renderer);
+        info!("Running on {renderer}");
+        info!("OpenGL version {gl_version}, shader_version {shader_version}");
 
         let is_gles_context = matches!(context.context_api(), ContextApi::Gles(_));
 
@@ -117,7 +121,7 @@ impl Renderer {
             Some(RendererPreference::Glsl3) => (true, true),
             Some(RendererPreference::Gles2) => (false, true),
             Some(RendererPreference::Gles2Pure) => (false, false),
-            None => (version.as_ref() >= "3.3" && !is_gles_context, true),
+            None => (shader_version.as_ref() >= "3.3" && !is_gles_context, true),
         };
 
         let (text_renderer, rect_renderer) = if use_glsl3 {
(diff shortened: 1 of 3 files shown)


kchibisov:
@chrisduerr You can test with `cargo run -- -o debug.renderer=gles2_pure -vv` and applying:

```patch
diff --git a/alacritty/src/renderer/platform.rs b/alacritty/src/renderer/platform.rs
index cb123793..05ba35c8 100644
--- a/alacritty/src/renderer/platform.rs
+++ b/alacritty/src/renderer/platform.rs
@@ -102,10 +102,10 @@ pub fn create_gl_context(
 ) -> GlutinResult<NotCurrentContext> {
     let debug = log::max_level() >= LevelFilter::Debug;
     let mut profiles = [
-        ContextAttributesBuilder::new()
-            .with_debug(debug)
-            .with_context_api(ContextApi::OpenGl(Some(Version::new(3, 3))))
-            .build(raw_window_handle),
+        // ContextAttributesBuilder::new()
+        //     .with_debug(debug)
+        //     .with_context_api(ContextApi::OpenGl(Some(Version::new(3, 3))))
+        //     .build(raw_window_handle),
         // Try gles before OpenGL 2.1 as it tends to be more stable.
         ContextAttributesBuilder::new()
             .with_debug(debug)
diff --git a/alacritty/src/renderer/text/atlas.rs b/alacritty/src/renderer/text/atlas.rs
index 9d4cb401..94d713a2 100644
--- a/alacritty/src/renderer/text/atlas.rs
+++ b/alacritty/src/renderer/text/atlas.rs
@@ -160,18 +160,18 @@ impl Atlas {
                     multicolor = false;
                     // Gles context doesn't allow uploading RGB data into RGBA texture, so need
                     // explicit copy.
-                    if self.is_gles_context {
-                        let mut new_buffer = Vec::with_capacity(buffer.len() / 3 * 4);
-                        for rgb in buffer.chunks_exact(3) {
-                            new_buffer.push(rgb[0]);
-                            new_buffer.push(rgb[1]);
-                            new_buffer.push(rgb[2]);
-                            new_buffer.push(u8::MAX);
-                        }
-                        (gl::RGBA, Cow::Owned(new_buffer))
-                    } else {
+                    // if !self.is_gles_context {
+                    //     let mut new_buffer = Vec::with_capacity(buffer.len() / 3 * 4);
+                    //     for rgb in buffer.chunks_exact(3) {
+                    //         new_buffer.push(rgb[0]);
+                    //         new_buffer.push(rgb[1]);
+                    //         new_buffer.push(rgb[2]);
+                    //         new_buffer.push(u8::MAX);
+                    //     }
+                    //     (gl::RGBA, Cow::Owned(new_buffer))
+                    // } else {
                         (gl::RGB, Cow::Borrowed(buffer))
-                    }
+                    // }
                 },
                 BitmapBuffer::Rgba(buffer) => {
                     multicolor = true;

```

alacritty/src/renderer/mod.rs
```diff
@@ -131,6 +135,16 @@ impl Renderer {
             (text_renderer, rect_renderer)
         };
 
+        // Enable debug logging for OpenGL as well.
+        if log::max_level() >= log::LevelFilter::Debug && GlExtensions::contains("GL_KHR_debug") {
```
chrisduerr, line 139:
```suggestion
        if log::max_level() >= LevelFilter::Debug && GlExtensions::contains("GL_KHR_debug") {
```

alacritty/src/renderer/mod.rs
```diff
@@ -131,6 +135,16 @@ impl Renderer {
...
+            debug!("Enabled debug logging for renderer");
```
chrisduerr, line 140: Why is this necessary? To see if `GL_KHR_debug` is supported?

Also this should be "debug logging for OpenGL", not renderer.

alacritty/src/renderer/mod.rs
```diff
@@ -131,6 +135,16 @@ impl Renderer {
...
+            unsafe {
+                gl::Enable(gl::DEBUG_OUTPUT);
+                gl::Enable(gl::DEBUG_OUTPUT_SYNCHRONOUS);
+                gl::DebugMessageCallback(Some(gl_debug_log), std::ptr::null_mut());
```
chrisduerr, line 144:
```suggestion
                gl::DebugMessageCallback(Some(gl_debug_log), ptr::null_mut());
```

alacritty/src/renderer/platform.rs
```diff
@@ -69,6 +71,23 @@ pub fn pick_gl_config(
...
+                    write!(msg, "\n  hardware_accelerated: {:?}", gl_config.hardware_accelerated());
+                let _ = write!(msg, "\n  config_api: {:?}", gl_config.api());
+                let _ = write!(
+                    msg,
+                    "\n  supports_transparency: {:?}",
+                    gl_config.supports_transparency()
+                );
+                let _ = write!(msg, "\n  srgb_capable: {:?}", gl_config.srgb_capable());
+                debug!("{}", msg);
+            }
```
chrisduerr, line 89: This seems like the worst possible way to write this. Wouldn't it be easier to just use `format!`?

chrisduerr (changes requested):

alacritty/src/renderer/mod.rs
kchibisov: Yeah, I want to check if extension is here.

alacritty/src/renderer/platform.rs
kchibisov: I think I've found a worse alternative.

alacritty/src/renderer/platform.rs
```diff
@@ -69,6 +70,24 @@ pub fn pick_gl_config(
...
         if let Some(gl_config) = gl_config {
+            debug!(
+                r#"Picked GL Config:
+  buffer_type: {:?}
+  alpha_size: {}
+  num_samples: {}
+  hardware_accelerated: {:?}
+  supports_transparency: {:?}
+  config_api: {:?}
+  srgb_capable: {}"#,
```
chrisduerr, line 81: I feel like this would be much better like this:

```suggestion
                "Picked GL Config:\n\
                  buffer_type: {:?}\n\
...
```

chrisduerr (changes requested):

chrisduerr (approved):

PR #6929, opened 2023-05-16
Update dependencies

kchibisov:
This is a series updating dependencies we're using to the latest version whenever possible. Each meaningful update is represented as a separate commit.


alacritty_config_derive/src/config_deserialize/de_enum.rs
```diff
@@ -14,7 +14,16 @@ pub fn derive_deserialize(ident: Ident, generics: Generics, data_enum: DataEnum)
     for variant in data_enum.variants.iter().filter(|variant| {
         // Skip deserialization for `#[config(skip)]` fields.
         variant.attrs.iter().all(|attr| {
-            !crate::path_ends_with(&attr.path, "config") || attr.tokens.to_string() != "(skip)"
```
chrisduerr, line 17: Can this function be removed?

chrisduerr (changes requested):

alacritty_config_derive/src/config_deserialize/de_enum.rs
kchibisov: I think I've tried, but it's actually required in other place. I could probably inline it there, but it looks sort of ugly.

chrisduerr (approved):

PR #6933, opened 2023-05-18
Change the default colorscheme

kchibisov:
The current colorscheme has poor contrast when it comes to yellow and green colors. It also has issues with the common white on blue and white on green. Morever the default background color is blue-ish.

The proposed colorscheme is inspired by base16-dark and addresses the issues of the current theme. While it may look a bit boring it uses the colors most dark themes tend to pick, which makes it look more natural in generic setups. It's also worth saying that it's variant is used by alacritty.org website.

--

The dim colors are not here at the moment, since I'm not sure how they should be configured(I think we're using the generated ones in the current theme as well).

Certain colors could be tweaked slightly. The white color was picked to have reasonable contrast with `blue` and `green` but to not look as bright as the original base16 themes have, they use `#d8d8d8`, though we can use it as well.

picture

picture


chrisduerr (changes requested):
I see zero reason why I would want to merge this. It's just the same exact issue with all other colorscheme changes all over again. It doesn't help at all.

kchibisov:
I'd probably just keep it around. It's just I've seen complains about yellow/green of the default theme again. The issue is that changing the yellow/green is not really working due to the blue background, normal green looks weird on it.

We could use other theme, but I just don't like that the default one has poor contrast.

Also, if look at the terminal themes at the https://github.com/chriskempson/tomorrow-theme they all differ to what we have. They have slightly different settings for normal colors and use the tomorrow night colors as the bright version. Surely some colors do match, but our bright colors are just glue from the high contrast theme and the regular, and the high contrast is what have issues.

Alternative solution would probably use the orange from tomorrow night instead of one of the yellows, but it could too radical.

chrisduerr:
Mixing and matching colors to create some kind of hodgepodge colorscheme will just make things more complicated without helping anyone.

kchibisov:
Just for the context. The theme is 

https://github.com/chriskempson/base16-vim/blob/master/colors/base16-classic-dark.vim as normal

and https://github.com/chriskempson/base16-vim/blob/master/colors/base16-default-dark.vim as bright.

Except the changes:

background is `#1c1c1c` instead of `#151515`(could be `#181818`, that's what we use on alacritty.org), but it could be changed.
foreground is `#c8c8c8` instead of `#d0d0d0`(could be `#d8d8d8`, we use `#aaaaaa`, but that's arbitrary).

red is not the one used by the classic theme, but could be reverted back to what it should be.

kchibisov:
As per discussion on IRC some agreement was made to use base16 classic dark, which is what was proposed but differed in the `red` color.

The only difference is `bright-black`(uses `#6b6b6b` to make it more readable with the colors such theme is using for `cursorline`(it's `#585858` on `#282828` by default, which is not that great) in various editors.

alacritty/src/config/color.rs
```diff
@@ -140,7 +140,7 @@ impl Default for FocusedMatchColors {
     fn default() -> Self {
         Self {
             background: CellRgb::Rgb(Rgb { r: 0x00, g: 0x00, b: 0x00 }),
-            foreground: CellRgb::Rgb(Rgb { r: 0xff, g: 0xff, b: 0xff }),
+            foreground: CellRgb::Rgb(Rgb { r: 0xf4, g: 0xbf, b: 0x75 }),
         }
     }
```
chrisduerr, line 145: This is not what you have in the configuration file and I'd assume the config values would be much more distinguishable.

alacritty.yml
```diff
@@ -278,9 +279,10 @@
   # Color used for the footer bar on the bottom, used by search regex input,
   # hyperlink URI preview, etc.
   #
+  # By default, these will use the opposing primary color.
```
chrisduerr, line 282:
```suggestion
  # By default, these will use the inverse primary color.
```

Since "inverse text" is the usual way terminal emulators refer to swapping foreground/background, I think it's good to use the well known language here.

alacritty.yml
```diff
@@ -294,39 +296,39 @@
...
-  #  cyan:    '#5b7d78'
-  #  white:   '#828482'
+  #  black:   '#0f0f0f'
+  #  red:     '#712b2b'
+  #  green:   '#5f6f3a'
+  #  yellow:  '#a17e4d'
+  #  blue:    '#456877'
+  #  magenta: '#704d68'
+  #  cyan:    '#4d7770'
+  #  white:   '#8e8e8e'
```
chrisduerr, line 331: These are calculated using Alacritty's "basic" luminance function?

alacritty.yml
```diff
@@ -196,12 +196,13 @@
...
-# Colors (Tomorrow Night)
+# Colors (base16 classic dark is used for normal colors and base16 default dark
+# for the bright ones)
 #colors:
   # Default colors
   #primary:
-  #  background: '#1d1f21'
-  #  foreground: '#c5c8c6'
+  #  background: '#181818'
+  #  foreground: '#d8d8d8'
```
chrisduerr, line 205: This is the original base16 classic white, right?

I'm not sure if I'm just used to my grey but this is **very** bright on top of `#181818`.

chrisduerr (changes requested):
This is the full color contrast chart:
tmp

For me personally normal and bright colors are basically indistinguishable. I'm not sure if these aren't too close.

One interesting test might be putting this on a stock KDE light DE. It's basically the brightest thing out there so it should be a good test for how this performs outside of the dark environment I'm personally used to (I'm happy to test this myself before merge).

alacritty.yml
kchibisov: I just copied the exact same line from the different option, so I'd change both of them.

alacritty.yml
kchibisov: yes.

kchibisov:
>For me personally normal and bright colors are basically indistinguishable. I'm not sure if these aren't too close.

I'd say that it's true on your picture but when you look at the text with them it's easier to tell, like I can distinguish them quite easily. The only issue is `red` and that's why I had it tweaked in my settings initially.

alacritty.yml
kchibisov: The original has `#d0d0d0` against the `#151515`, and the default dark has `#d8d8d8` against the `#181818`, so if we use `#181818` I've picked the `#d8d8d8`. It's indeed bright, but it provides better contrast with white on blue and white on green, which are widely used.

chrisduerr:
Honestly my main concern with this is the contrast between normal and bright white. I cannot tell the difference personally:

tmp

alacritty.yml
```diff
@@ -196,12 +196,13 @@
 # If `true`, bold text is drawn using the bright color variants.
 #draw_bold_text_with_bright_colors: false
 
-# Colors (Tomorrow Night)
+# Colors (base16 classic dark is used for normal colors and base16 default dark
```
chrisduerr, line 199: Probably easier to move this above the normal and bright sections?

chrisduerr (changes requested):

chrisduerr (changes requested):
I cannot in good faith approve this when realistically only the normal colors are properly useful.

We've had a messy combination of two themes as our last theme and that was one of its big criticisms, so I cannot in good faith move forward doing the same thing again.

The bright theme should either be a different theme that is so distinctly different, that nobody is going to complain, or it should be automatically generated like our dim colors (which I see as likely the best option).

kchibisov:
The thing is with this combination is that the one is lighter variant of another, however it's not bright enough to say so most of the time, however if you increase the components brightness you'll get base16 default dark.

Are these colors any good:

```yml
  bright:
    black:   '#6b6b6b'
    red:     '#C17171'
    green:   '#ACBE83'
    yellow:  '#F7CF97'
    blue:    '#8FB7C7'
    magenta: '#BF97B7'
    cyan:    '#97C7BF'
    white:   '#f8f8f8'

```

I got red-cyan by increasing brightness by 25%. You could play with generating bright colors yourself with https://colorkit.co/palette/ac4242-90a959-f4bf75-6A9FB5-aa759f-75B5AA/ (default normal colors).

chrisduerr (approved):

PR #6934, opened 2023-05-18
Add ability to zoom to alternate font size directly

reavessm:
I share my screen during video calls a fair bit during work.  On these calls, it's common to need to increase the font size so the other person can read my terminal.  I can't be bothered to type Ctrl+Plus mutliple times, so it'd be nice to bind a key to put my terminal into "Big Font Mode", then I can ResetFontSize when I'm done.  I chose Key9 for this default binding since the ResetFontSize binding defaults to Key0.

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -134,6 +134,9 @@ pub enum Action {
     /// Reset font size to the config value.
     ResetFontSize,
 
+    /// Set font size to specific value
+    ZoomFont,
+
     /// Scroll exactly one page up.
     ScrollPageUp,
 
@@ -675,6 +678,7 @@ fn common_keybindings() -> Vec<KeyBinding> {
             +BindingMode::VI, ~BindingMode::SEARCH; Action::ClearSelection;
         Insert,   ModifiersState::SHIFT, ~BindingMode::VI; Action::PasteSelection;
         Key0,     ModifiersState::CTRL;  Action::ResetFontSize;
+        Key9,     ModifiersState::CTRL;  Action::ZoomFont;
         Equals,   ModifiersState::CTRL;  Action::IncreaseFontSize;
         Plus,     ModifiersState::CTRL;  Action::IncreaseFontSize;
         NumpadAdd,      ModifiersState::CTRL;  Action::IncreaseFontSize;
--- a/alacritty/src/config/font.rs
+++ b/alacritty/src/config/font.rs
@@ -40,6 +40,9 @@ pub struct Font {
     /// Font size in points.
     size: Size,
 
+    /// Font size in points, when zoomed int.
+    zoom_size: Size,
+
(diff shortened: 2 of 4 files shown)


reavessm:
While this is currently functional, I'm leaving this in draft status until two things are completed (at least):

- [ ] I'm not really sure how to parse this from the config file, but I also haven't really looked into that yet
- [ ] When I `cargo run` this, I get an error saying, `[0.023804604s] [ERROR] [alacritty_config_derive] Config error: key_bindings: invalid type: unit value, expected a sequence`.  I believe this is due to how I'm setting the default value of 25. in the default keybinding, but I'm not really sure.

I plan to do some  more debugging with this later today, but I wanted to go ahead and get conversation going on this.

kchibisov:
We have font size increment set internally to 0.5, maybe we could add a configurable value for it instead, so you could define some arbitrary value?

Alternatively, you can set font size for the current running window with `alacritty msg config font.size=20` and then `alacritty msg config -r` to reset it back.

chrisduerr (changes requested):
Bloating up the config with some random preferences is definitely not the right way about expanding Alacritty's functionality.

As @kchibisov has said this is already pretty trivial with IPC.

reavessm:
That `alacritty msg config` is a pretty neat trick.  Apparently Fedora 37's repos are a bit old for alacritty because I don't have that option.

@kchibisov is there a way to run that from an alacritty keybinding?  Because that's the problem  I was trying to solve.  This PR is a one-click solution.

@chrisduerr I'm not really sure how one line is bloating the config, especially when it's optional.  And I wouldn't exactly say a new feature is a random preference.  If you guys don't want the feature, that's fine, but let's call it what it is.   I'm not really sure what the correct procedure is for requesting features but I figured writing the code would be the most maintainable.

reavessm:
Also, if you're going to write a script somewhere to run `alacritty msg config font.size=20`, it makes the most sense to go ahead and put that in the config next to the regular font size so you can see what it's set to.  When I'm operating my computer for my day-to-day work, I don't remember what my font size is.  It my be 10, it may be 15.  So manually setting the font size in an ad-hoc manner doesn't make sense either.  If I set font.size and font.zoom_size next to each other, I can make sure the ratio is exactly how I want it.

If I can't have a simple keybinding in alacritty and have to remember what my font size currently is, I might as well just spam Ctrl+Plus a bunch of time and hope for the best.

kchibisov:
I mean you can `set` and `reset` to default, assuming that you have 2 modes basically.

kchibisov:
You can execute any program from the alacritty key bindings, the alacritty 0.12.0 is shipped on fedora 38 for example.

chrisduerr:
> I'm not really sure how one line is bloating the config

Right. One feature for one user. Extrapolate to thousands of people and you got a long config file.

PR #6952, opened 2023-05-24
Fix 216-color cube generation

chrisduerr:
This fixes a regression introduced in cb7ad5b which swapped the green and blue values when constructing the 216-color RGB cube.


Andy-Python-Programmer (approved):
Next time I will be more careful with mass regex search and replaces :^)

kchibisov (approved):

PR #6954, opened 2023-05-25
Don't copy Glyph on GlyphCache hit

kchibisov:
The Glyph is being passed around by reference, so doing the copy is not desired.

--

It might look ugly a bit given that missing glyph handling must be moved, but it's actually
don't try to copy lots of glyph data around on each renderer call.


chrisduerr (changes requested):
What inspired this change? Saying this is a bit ugly compared to the current code is an understatement, so there should be some good justification.

kchibisov:
The glyph is ~32 bytes and we copy it for each cell, so if we assume 77 * 180 cells we copy 400 KB of data for no reason for each frame, given that we only use a reference, and not the copy.


And with my code we copy 4x less, so it's simply faster?

chrisduerr:
> And with my code we copy 4x less, so it's simply faster?

Says who? Unless you've benchmarked it, it's not faster. Making the code more complicated if there's no real world benefit to it isn't useful.

Issue #4754, opened 2021-02-01
Report other mouse buttons


remme123:
I've tried map mouse button X1 & X2 in vim, but nothing happened when I clicked. And double click also heppened nothing.

remme123:
vim map command: map <X1Mouse> :echo '1'<cr>

chrisduerr:
What are mouse buttons X1 and X2?

remme123:
As described in the help of vim:
> The X1 and X2 buttons refer to the extra buttons found on some mice.  The
'Microsoft Explorer' mouse has these buttons available to the right thumb.
Currently X1 and X2 only work on Win32 and X11 environments.

chrisduerr:
Are you aware of any other terminal emulator supporting this, or does this just work in gvim?

In theory supporting this in Alacritty would be possible and it currently does not correctly send these mouse buttons, but the way this is phrased with Win32 and X11 makes me question if it is supported by vim inside the terminal.

remme123:
Yes, I've only seen it in gvim. It will be great if alacritty supports this feature.

nixpulvis:
@remme123 I'm assuming you didn't mean to close this issue, reopening.

chrisduerr:
It's possible that this cannot be supported in terminal vim, simply because vim doesn't support it. But that doesn't mean that Alacritty can't provide support for it.

sophrosyne97:
@chrisduerr I am interested in closing this issue, I just have a couple of questions.

What exactly is meant by "Report other mouse buttons"? Does this just mean that we should allow a user to map to whatever extra mouse buttons they have, or does report mean something more like actually showing them that these buttons are available somehow? In on_mouse_press() match statement in input.rs I see a comment saying "Can't properly report more than three buttons..", would properly handling most extra mouse buttons by returning a value for each one here be a start towards this?

In looking into mapping extra mouse buttons I've briefly looked through the code and saw that you use a MouseButton enum with one of the variants being a tuple struct Other(u16). I am assuming this u16 is for referring to the particular button value? For example, using xev I know that the extra buttons on my mouse refer to button8 and button9; continuing off what I asked in the previous paragraph about input.rs, would MouseButton::Other(8) properly match to button8 somehow? I am not really sure how to properly read the correct extra mouse button and then handle it with the MouseButton enum.

[...]

PR #6957, opened 2023-05-26
Fix Report other mouse buttons #4754

marianomarciello:
This enables mapping mouse buttons in mouse_bindings and also enables mouse buttons signal in vim/nvim. So it should be possibile to map buttons X1 and X2 as stated in the vim documentation or at least to have the same behavior of other terminals.


alacritty/src/input.rs
```diff
@@ -623,8 +622,7 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
                 MouseButton::Left => 0,
                 MouseButton::Middle => 1,
                 MouseButton::Right => 2,
-                // Can't properly report more than three buttons.
-                MouseButton::Other(_) => return,
+                MouseButton::Other(x) => x.try_into().unwrap(),
```
chrisduerr, line 625: I really doubt this reports mouse buttons correctly. Did you just guess at what value should be reported?

chrisduerr (changes requested):

alex-courtis:
Tested with the following:

```yaml
mouse_bindings:
  - { mouse: Left, chars: 'L' }
  - { mouse: Middle, chars: 'M' }
  - { mouse: Right, chars: 'R' }
  - { mouse: 4, chars: '4' }
  - { mouse: 5, chars: '5' }
  - { mouse: 6, chars: '6' }
  - { mouse: 7, chars: '7' }
  - { mouse: 8, chars: '8' }
  - { mouse: 9, chars: '9' }
```

LMR were sent however numbered were not.

chrisduerr:
https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Other-buttons

alex-courtis:
> https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Other-buttons
> 
> For terminal emulators it is important we're reporting the **exact same** things as every other terminal emulator. I recommend testing XTerm for comparison.

FYI I used st to debug the standard xterm control sequences: https://github.com/alacritty/alacritty/issues/4754#issuecomment-939402976

alacritty/src/input.rs
marianomarciello: Hi @chrisduerr  thanks for your time. I have tested this with my mouse that has one additional button (button 8) and the number reported is 8. Then I used xbindkeys to bind button 9, 10 and 11 with my left shift and the value reported was 9, 10 and 11 respectively.  Should I follow the convention linked by you ? (https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Other-buttons)

chrisduerr: Yes you should.

alex-courtis:
> Can you provide me the mapping of your buttons ? (You can use xev for this).

Wayland and X11 have different behaviour regarding button numbers.

X11:
- xev reports button press 1 to 9 as expected
  - 1 : left
  - 2 : middle
  - 3 : right
  - 4 : wheel up
  - 5 : wheel down
  - 6 : wheel left
  - 7 : wheel right
  - 8 : side rear
  - 9 : side front
- Buttons 8 to 9 are seen by alacritty and send 8, 9
- Buttons 4 to 7 are NOT seen by alacritty

Wayland:
- wev reports differently to xev
  - 272(left) : left
  - 274(middle) : middle
  - 273(right) : right
  - axis events : wheel up/down/left/right
  - 275(side) : side rear
  - 276(extra) : side front
- xev reports same as under X11
- Buttons 275 and 276 are seen by alacritty and send S, E
- Buttons 4 to 9 are NOT seen by alacritty

This behaviour does not appear to be any different to master.

```yaml
mouse_bindings:
  - { mouse: Left, chars: 'L' }
  - { mouse: Middle, chars: 'M' }
  - { mouse: Right, chars: 'R' }
  - { mouse: 4, chars: '4' }
  - { mouse: 5, chars: '5' }
  - { mouse: 6, chars: '6' }
  - { mouse: 7, chars: '7' }
  - { mouse: 8, chars: '8' }
  - { mouse: 9, chars: '9' }
  - { mouse: 272, chars: 'l' }
  - { mouse: 273, chars: 'r' }
  - { mouse: 274, chars: 'm' }
  - { mouse: 275, chars: 'S' }
  - { mouse: 276, chars: 'E' }
```

Mouses tested: logitech g502, steelseries rival 3. Model does not appear to be relevant.

kchibisov:
Axis events are not send via buttons in the first place by our upstream library, and won't be sent, because it's not a button.

The thing is that every button should be normalized to a value it should, which is not the case and that's why it's questionable. I'm pretty sure on macOS/Windows the situation will be different and that event on X11/Wayland it's behaving not the way it should.

I don't think application generally support mouse buttons other than `back` and `forward` (which is 4 and 5 or 6/7, idk). Thus I'd suggest to translate only them and there's a patch upstream to namely expose such buttons in winit https://github.com/rust-windowing/winit/pull/2770

marianomarciello:
> 

Hi @kchibisov  thanks for your time.
I think `back` and `forward` buttons are not that useful in a terminal. I would prefer to report those buttons in a way that we can then bind them to a specific action (using mouse_bindings). `back` and `forward` can then be enabled using mouse_bindings.

marianomarciello:
> 

Hi @alex-courtis. I think the Mousewheel is handled with the `MouseWheel` field in the `WindowEvent` enum (from the winit crates). While MouseButtons (like button 8..9) are handled using the enum `MouseButton` (from the winint creates). So long story short I think that the Mousewheel events are handled before the MouseButton code, so they can not reach the code that I am changing. I think this can be part of another PR. I will try to fix the value reported (using the xterm convention) and then switch to the MouseWheel issue.

alex-courtis:
> I think this can be part of another PR. I will try to fix the value reported (using the xterm convention) and then switch to the MouseWheel issue.

That's fantastic. Please let me know when you need more testing.

kchibisov:
I'm still not sure how it should all work and translate, surely it might work on X11, but outside of it, I highly doubt it.

However the new winit API defined `Back` and `Forward` on all the platforms and those are `8` and `9` in your PR (the same will map to Wayland).

marianomarciello:
Hi @kchibisov thanks again for your time.
I only tested it in X11, @alex-courtis tested it in wayland too. The `8` that I was talking about is because my hardware has this mapping. So my `Back` button is button 8. I think this is not so rare.  So there is no guarantee that some buttons correspond to some functionality. Maybe it is better to let the user decide the mapping in the configuration file ?

kchibisov:
Well, the thing is that reporting for such buttons will be broken most of the time, because Xterm references X11, where it'll work, because Xorg remaps the buttons, but not on Wayland, macOS, Windows, where there's no such thing.

What I've suggested is to map `Back` and `Forward` to `8` and `9` and report it the way it should according to Xterm, as for the rest, I'm not sure we should report them, because we can't translate them to the way xterm spec expected them to be. (On Wayland you have 0x110+, because those are coming from libinput iirc).

alex-courtis:
> What I've suggested is to map `Back` and `Forward` to `8` and `9`

That appears to be a reliable standard; seen across all mouse hardware I've owned.

marianomarciello:
So we need to avoid reporting 8 and 9 ? Maybe I need to take a look into the button mapping after this PR [https://github.com/rust-windowing/winit/pull/2770](https://github.com/rust-windowing/winit/pull/2770), maybe everything works out of the box after this rework :)

kchibisov:
@marianomarciello the said branch merged, we now get button events out of winit which we should be able to map to `8` and `9` for the mouse mode from what I can say.

If the intention is to use them in the config, you can bind `Back` and `Forward` buttons.

chrisduerr:
Closing the PR since it's dead.

VishnuSanal:
> What I've suggested is to map `Back` and `Forward` to `8` and `9`

@chrisduerr this was the only thing that was missing in this PR, right?

chrisduerr:
https://github.com/alacritty/alacritty/pull/6957#issuecomment-1595028204

PR #6958, opened 2023-05-28
Update to the new winit keyboard API

kchibisov:


kchibisov:
That's a reference what we get when pressing `ctrl + shift + c`.

```
[1.288686403s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(94882226519600)), event: KeyboardInput { device_id: DeviceId(Wayland(DeviceId)), event: KeyEvent { physical_key: KeyI, logical_key: Character("C"), text: Some("C"), location: Standard, state: Pressed, repeat: false, platform_specific: KeyEventExtra { key_without_modifiers: Character("c"), text_with_all_modifiers: Some("\u{3}") } }, is_synthetic: false } }
[1.349699787s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(94882226519600)), event: KeyboardInput { device_id: DeviceId(Wayland(DeviceId)), event: KeyEvent { physical_key: KeyI, logical_key: Character("C"), text: Some("C"), location: Standard, state: Released, repeat: false, platform_specific: KeyEventExtra { key_without_modifiers: Character("c"), text_with_all_modifiers: Some("\u{3}") } }, is_synthetic: false } }

```

kchibisov:
This branch should sort of work now, except for macOS users given that `OptionAsAlt` is not ported yet.

kchibisov:
As it was discussed on IRC, I've added transition from old winit keys into their modern counterparts.

kchibisov:
I left all complexity wrt `char: ` bindings and adding support for direct key code references out of this PR for now, since it's not yet clear whether we really want them.

kchibisov:
The one more question is what should we do with the `Numpad` bindings, we had them before, but with the current winit, they are not encoded into the virtual key code directly, but rather reside in the `position` option.

We still have information whether a binding was from a numpad or not, but it could require config changes, such as adding an implicit field default to `None` or `Standard`.

We could also drop support for old numpad keys in the config and solely rely on our new keys, that would require users to delete a few lines, but I don't think that's an issue. I'd prefer this variant btw.

afh:
For what it's worth alacritty has seen config changes in the past that required users to make modifications to their config.

@kchibisov could you give an example of a change that a user would need to make to replace the old numpad keys with the new keys in their config?

kchibisov:
@afh there's no difference between them and normal keys now. Unless you manually define special set of bindings just for numpad, which differ from regular keys there shouldn't be an issue.

So all `NumpadSubtract/Add` are now just `-`,`+`, etc.

chrisduerr:
@perfbot (just testing stuff)

afh:
Thanks for clarifying, @kchibisov. Personally I do not use NumPad keys specifically, yet I could see someone wanting to do that. Would it be possible and useful to introduce a `Numpad` modifier to `Control`, `Alt`, or `Shift` when specifying a key binding?

For example:
```yaml
key_bindings:
   { key: Minus, mods: Numpad, action: DecreaseFontSize }
```

kchibisov:
It's just not clear how default bindings should work then, should we separately define numpad bindings or not, because it affects all the keys. What we get now from the winit is

```
Key { code: Character("-"), position: Numpad }
Key { code: Character("-"), position: Standard }
```

We could add implicit `Standard` field to all the bindings, and translate `Numpad` to `Numpad` bindings, and still have the bindings separate... That could work, but not sure how convenient it could be.

kchibisov:
The numpad thing in the config should be resolved now.

alacritty/src/input.rs
```diff
@@ -917,16 +941,43 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
         }
 
+        let text = key.text_with_all_modifiers().unwrap_or_default();
+
         // All key bindings are disabled while a hint is being selected.
         if self.ctx.display().hint_state.active() {
-            *self.ctx.suppress_chars() = false;
+            for ch in text.chars() {
+                self.ctx.hint_input(ch);
+            }
```
chrisduerr, line 980:
```suggestion
            for c in text.chars() {
                self.ctx.hint_input(c);
            }
```

Far more common abbreviation I think.

alacritty/src/input.rs
```diff
@@ -939,102 +990,67 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
-        if self.ctx.display().ime.preedit().is_some() {
+        // Key bindings suppress the character input.
+        if self.process_key_bindings(&key) {
             return;
         }
 
-        // Handle hint selection over anything else.
-        if self.ctx.display().hint_state.active() && !suppress_chars {
-            self.ctx.hint_input(c);
+        // TODO maybe we should just accept text?
```
chrisduerr, line 998: TODO must be resolved, do whatever you think works best I don't really mind.

alacritty/src/input.rs
```diff
@@ -939,102 +990,67 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
         #[cfg(not(target_os = "macos"))]
         let alt_send_esc = true;
 
-        // Don't send ESC when `OptionAsAlt` is used. This doesn't handle
-        // `Only{Left,Right}` variants due to inability to distinguish them.
         #[cfg(target_os = "macos")]
-        let alt_send_esc = self.ctx.config().window.option_as_alt != OptionAsAlt::None;
+        let alt_send_esc = {
+            let option_as_alt = self.ctx.config().window.option_as_alt;
+            // Only send ESC for an `Alt` key.
```
chrisduerr, line 1020: What does this comment mean?

alacritty/src/input.rs
```diff
@@ -939,102 +990,67 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+            option_as_alt == OptionAsAlt::Both
+                || (option_as_alt == OptionAsAlt::OnlyLeft
+                    && self.ctx.modifiers().lalt_state() == ModifiersKeyState::Pressed)
+                || (option_as_alt == OptionAsAlt::OnlyRight
+                    && self.ctx.modifiers().ralt_state() == ModifiersKeyState::Pressed)
```
chrisduerr, line 1025: This could probably be neatly implemented on `OptionAsAlt` and clean up the code here significantly.

alacritty/src/input.rs
```diff
@@ -939,102 +990,67 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+        };
 
-        if alt_send_esc
-            && *self.ctx.received_count() == 0
-            && self.ctx.modifiers().alt()
-            && utf8_len == 1
-        {
-            bytes.insert(0, b'\x1b');
+        let mut bytes = Vec::with_capacity(text.as_bytes().len() + 1);
```
chrisduerr, line 1028:
```suggestion
        let mut bytes = Vec::with_capacity(text.len() + 1);
```

This is identical, is it not?

alacritty/src/input.rs
```diff
@@ -939,102 +990,67 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+        if alt_send_esc && self.ctx.modifiers().state().alt_key() && text.len() == 1 {
```
chrisduerr, line 1029: Checking for the alt key again here when it's basically done already on macOS seems unfortunate.

alacritty/src/input.rs
```diff
@@ -939,102 +990,67 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+            bytes.push(b'\x1b');
         }
+        bytes.extend_from_slice(text.as_bytes());
```
chrisduerr, line 1032: This always reallocates the text, but wouldn't it be more efficient to `insert` like before which would only reallocate when `alt` is held?

alacritty/src/config/bindings.rs
```diff
@@ -365,11 +346,26 @@ macro_rules! bindings {
...
+    (KeyBinding, $key:literal, $location:expr) => {
+        BindingKey::Keycode { key: Character($key.into()), location: $location }
+    };
+    (KeyBinding, $key:expr,) => {
+        BindingKey::Keycode { key: $key, location: KeyLocation::Standard }
+    };
+    ($ty:ident, $key:expr,) => {
+        $key
+    };
+}
```
chrisduerr, line 362: Why is this a separate macro?

alacritty/src/config/bindings.rs
```diff
@@ -1023,7 +981,9 @@ impl<'a> Deserialize<'a> for RawBinding {
                                     },
                                 },
                                 None => {
-                                    key = Some(Key::deserialize(value).map_err(V::Error::custom)?);
+                                    let k =
+                                        BindingKey::deserialize(value).map_err(V::Error::custom)?;
+                                    key = Some(k);
```
chrisduerr, line 986: I'd assume this was done in an attempt to improve formatting? Seems like it didn't work out, might as well inline it.

alacritty/src/config/bindings.rs
```diff
@@ -734,23 +628,83 @@ pub fn platform_key_bindings() -> Vec<KeyBinding> {
...
         match u32::deserialize(value.clone()) {
-            Ok(scancode) => Ok(Key::Scancode(scancode)),
+            Ok(scancode) => Ok(BindingKey::Scancode(KeyCode::from_scancode(scancode))),
             Err(_) => {
-                let keycode = VirtualKeyCode::deserialize(value).map_err(D::Error::custom)?;
-                Ok(Key::Keycode(keycode))
+                let keycode = String::deserialize(value.clone()).map_err(D::Error::custom)?;
+                let (key, location) = if keycode.len() == 1 {
+                    (Key::Character(keycode.to_lowercase().into()), KeyLocation::Standard)
+                } else {
```
chrisduerr, line 649: Why are you checking for `keycode.len() == 1` here? Shouldn't it be based on the number of characters not bytes?

chrisduerr (changes requested):

extra/man/alacritty-bindings.5.scd
```diff
@@ -38,11 +38,11 @@ configuration. See *alacritty*(5) for full configuration format documentation.
 :[
 :[
 :  _"Copy"_
-|  _"L"_
```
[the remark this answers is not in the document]
kchibisov: Do you mean why I've changed the case to lower one? I think I've replied to you on IRC wrt that, and it was simply to clarify that you can write lowercase letters even for `Shift` bindings and that you should look on the lowercase latter.

If you like uppercase more I can change back...

alacritty/src/input.rs
kchibisov: But we do `with_capacity` so it doesn't reallocate anything?

alacritty/src/config/bindings.rs
kchibisov: You can try to make it not a separate macro, but for me it's way more readable than what was before.

I've tried to encode all that special case handling into our main macro, but it was not always working because it was matching `Character("l")` in other branch after the first invocation and so on.

alacritty/src/input.rs
kchibisov: I don't think adding a special trait just for macOS is worth it, however I could encapsulate all of that into a separate function on input.

extra/man/alacritty-bindings.5.scd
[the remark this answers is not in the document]
chrisduerr: I think uppercase looks cleaner and is easier to identify.

alacritty/src/input.rs
```diff
@@ -472,14 +477,14 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+        let modifiers = self.ctx.modifiers().state();
+        if modifiers.shift_key() {
             mods += 4;
         }
-        if modifiers.alt() {
+        if modifiers.alt_key() {
             mods += 8;
         }
-        if modifiers.ctrl() {
+        if modifiers.control_key() {
```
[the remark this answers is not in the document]
chrisduerr: Yeah just seems like something that is unnecessary breakage when this could work without a change.

alacritty/src/input.rs
chrisduerr: I guess.

alacritty/src/input.rs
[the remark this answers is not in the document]
kchibisov: There were other breaking changes though to this state, like some keys got renamed to `super` and `control` from the `logo` and `ctrl` respectively.

extra/man/alacritty-bindings.5.scd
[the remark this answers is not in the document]
kchibisov: Reverted.

alacritty/src/config/bindings.rs
chrisduerr: I made some minor adjustments (mainly wrapping everything in double `{{ }}` but the alternatives I could come up with were less readable than this solution.

The best way to simplify these macros would likely be to switch away from `bindings!` and go for `binding!`, but I'm not interested in doing that right now and I'm sure neither are you.

extra/man/alacritty.5.scd
```diff
@@ -656,11 +656,14 @@ This section documents the *[keyboard]* table of the configuration file.
 
 	*key* <string>
 
-		Identifier of the binding's key, for example: _"A"_, _"F1"_, or
-		_"Key0"_.
-
-		A full list with available key codes can be found here:++
-https://docs.rs/winit/\*/winit/event/enum.VirtualKeyCode.html#variants
+		The regular keys like _"A"_, _"0"_, and _"Я"_ could be mapped directly
```
chrisduerr, line 659: Much prefer `Я` here, it's way more clear.

chrisduerr, line 659:
```suggestion
		The regular keys like _"A"_, _"0"_, and _"Я"_ can be mapped directly
```

extra/man/alacritty.5.scd
```diff
@@ -656,11 +656,14 @@ This section documents the *[keyboard]* table of the configuration file.
...
+		without any special syntax. Full list of named keys like _"F1"_ or a
+		syntax for dead keys could be found here:++
```
chrisduerr, line 661:
```suggestion
		without any special syntax. Full list of named keys like _"F1"_ and the
		syntax for dead keys can be found here:++
```

chrisduerr (changes requested):

chrisduerr (approved):

afh:
Thank you so much for working on this, @kchibisov, and keep up the great work!! 🎉 I'm happy to see this merged and look forward to the next alacritty release (candidate) as it should resolve a long standing issue I had with alacritty on OpenBSD (see rust-windowing/winit#2032).

PR #6961, opened 2023-05-29
Update path of changed doc files in Makefile

maierd:
It looks like the paths to the recently changed doc files (https://github.com/alacritty/alacritty/commit/e3746e49a117ebf8d1f43715554c5a1cf8d4a116) were not updated in the Makefile, with this change they now point to the files again.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,6 +14,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+      - name: Install dependencies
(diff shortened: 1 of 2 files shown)


kchibisov (changes requested):
The patch should use `scdoc` in the first place and not simply compress `scd` files.

@chrisduerr but I thought that we don't build man pages on macOS?

chrisduerr (approved):
Oopsie.

kchibisov (approved):

PR #6963, opened 2023-05-29
Improve Alacritty's short description

zer0-x:
The current short description is:
```
A fast, cross-platform, OpenGL terminal emulator
```

1. I think that we should not say `OpenGL`, but what `OpenGL` achieve, which is [hardware-accelerated](https://en.wikipedia.org/wiki/Hardware_acceleration) [rendering](https://en.wikipedia.org/wiki/Rendering_(computer_graphics)). However, we can mention `OpenGL` in a longer description where technical stuff should be mentioned.

2. I don't think that we should mention `cross-platform` in files like [`extra/linux/Alacritty.desktop`](https://github.com/alacritty/alacritty/blob/d94cb6be99b098b0084d0a54dc4e655cd9c0e47c/extra/linux/Alacritty.desktop) and [`extra/linux/org.alacritty.Alacritty.appdata.xml`](https://github.com/alacritty/alacritty/blob/d94cb6be99b098b0084d0a54dc4e655cd9c0e47c/extra/linux/org.alacritty.Alacritty.appdata.xml), since they are targeting single platform (Linux).

---

I would like to hear other ideas and suggestions, before it being merged *( or not, I don't know 🤣 )*.


chrisduerr (changes requested):
No.

zer0-x:
Sorry for wasting you time, but is there any explanation?

I thought it would make the description simpler and more clear to understand.

chrisduerr:
It's neither simpler, nor clear to understand. It's less precise for no reason at all and unnecessarily inconsistent when the goal of the description is completely unrelated to the platform Alacritty is shipped on.

PR #6977, opened 2023-06-04
refactor: Replace once_cell to std lib

howjmay:


kchibisov:
Thanks for the patch, but we can't bump our MSRV.

PR #6987, opened 2023-06-11
Update windows-sys to 0.48.0

proski:


chrisduerr:
What's the motivation behind this change?

proski:
I was debugging an unrelated issue with vim for Windows (it shows `m` on startup) and noticed that the versions of windows-sys are different across the project. So I tried updating them and noticed that `INFINITE` was moved to `windows_sys::Win32::System::Threading`. Even though updating windows-sys made no difference to that issue, I didn't want to throw away that change, as somebody would have to make it one day.

chrisduerr (approved):
Sounds good. I've taken the liberty to also apply all automatic updates using `cargo update`.

Issue #6976, opened 2023-06-03
Allow hints to persist after selection


djl:
foot has a neat feature as part of it's URL hints mode where hint characters are persisted after a hint has been selected. This makes it easy to quickly select multiple hints without having to hit the same key binding repeatedly after each one.

Is this something Alacritty could add?

chrisduerr:
Adding a boolean flag to stay in hint mode after selection seems like a reasonable addition.

Should be a great first issue too if someone is interested.

SonuBardai:
Hey! I've set up a local env and have played around with Alacritty for a bit. Just wanted to get some more clarity on the hints feature. So, if I hit `Ctrl+Shift+U` I see these vimium style keybindings that highlight hyperlinks, and we want them to persist even after the user triggers them? Should this be an opt-in from the yml config file? 

<img width="699" alt="Screenshot 2023-06-11 at 15 21 24" src="https://github.com/alacritty/alacritty/assets/67749330/8c1bf52e-0680-4230-bb17-54f94c5431d1">

djl:
How it works in foot is that the hints are immediately re-enabled after a hint has been selected. Effectively the same as hitting the shortcut again after the selection.

For the config, I imagine this would be on a per-enabled-hint basis. e.g.

```
hints:
  alphabet: "jfkdls;ahgurieowpq"
  enabled:
  - regex: "(ipfs:|ipns:|magnet:|mailto:|gemini:|gopher:|https:|http:|news:|file:|git:|ssh:|ftp:)[^\u0000-\u001F\u007F-\u009F<>\"\\s{-}\\^\u27e8\u27e9`]+"
    hyperlinks: true
    command: xdg-open
    persist: true
    post_processing: true
    mouse:
      enabled: true
      mods: None
    binding:
      key: O
      mods: Control|Shift
  - regex: "(ipfs:|ipns:|magnet:|mailto:|gemini:|gopher:|https:|http:|news:|file:|git:|ssh:|ftp:)[^\u0000-\u001F\u007F-\u009F<>\"\\s{-}\\^\u27e8\u27e9`]+"
    hyperlinks: true
    command: xdg-open
    post_processing: true
    mouse:
      enabled: true
      mods: None
    binding:
      key: U
      mods: Control|Shift
```

The first block has `persist: true` while the second block just uses the default of `persist: false`.

kchibisov:
>Should this be an opt-in from the yml config file?

yes, with the option named `persist` which was named above.

[...]

PR #6992, opened 2023-06-11
Issue #6976: Hints persist after selection

SonuBardai:
Closes https://github.com/alacritty/alacritty/issues/6976

https://github.com/alacritty/alacritty/assets/67749330/0490fa7c-d905-448f-97b3-71da8d41417c

### Hints with multiple characters

https://github.com/alacritty/alacritty/assets/67749330/b85eef4e-fa8e-4aa3-a4cf-e3042d32aa60


alacritty.yml
```diff
@@ -518,6 +518,7 @@
   #           [^\u0000-\u001F\u007F-\u009F<>\"\\s{-}\\^⟨⟩`]+"
   #   hyperlinks: true
   #   command: xdg-open
+  #   persist: true
```
chrisduerr, line 521: It needs to be documented in the section above what this option actually does.

alacritty/src/config/ui_config.rs
```diff
@@ -255,6 +255,7 @@ impl Default for Hints {
             enabled: vec![Hint {
                 content,
                 action,
+                persist: true,
```
chrisduerr, line 258: The default should be false. Most people don't want to open multiple URIs and this causes additional friction when having to dismiss the hints.

alacritty/src/display/hint.rs
```diff
@@ -150,7 +150,9 @@ impl HintState {
             let bounds = self.matches[index].clone();
             let action = hint.action.clone();
 
-            self.stop();
+            if !hint.persist {
+                self.stop();
+            }
 
             // Hyperlinks take precedence over regex matches.
             let hyperlink = term.grid()[*bounds.start()].hyperlink();
```
chrisduerr, line 158:
```suggestion
            // Exit hint mode unless it requires explicit dismissal.
            if !hint.persist {
                self.stop();
            }
            
            let bounds = self.matches[index].clone();
            let action = hint.action.clone();

            // Hyperlinks take precedence over regex matches.
            let hyperlink = term.grid()[*bounds.start()].hyperlink();
            
```

The order of operations here is a bit confusing, not really related to your PR but I'd fix it while we're at it. This way the data aggregation for the return value is separate from the modification to the state itself.

The little comment on the conditional will also hopefully make things more clear.

chrisduerr (changes requested):
This needs a changelog entry.

kchibisov (dismissed):
This doesn't work correctly when the hint contains multiple characters.

What it should do is to reset the hint state as in no char was input after the user typed the last char with `persist` for the hint. So you simply could start again. Right now it's getting stuck on the last char left from original input.

extra/man/alacritty.5.scd
```diff
@@ -616,6 +616,11 @@ _action_ or a _command_.
 
 		The _enabled_ field controls if the hint should be underlined when
 		hovering over the hint text with all _mods_ pressed.
+	
+	*persist* <boolean>
+
+		Set to _false_ by default. When this is _true_, hints remain persistent 
+		after selection.
```
chrisduerr, line 623: Can we move this up to the other boolean options? I think that structures a little nicer.

Same goes for the order in the type definition of `enabled`.

alacritty/src/display/hint.rs
SonuBardai: I'm trynna see how I can fix this after reorganizing the operations. Still a little rusty at Rust, sry! :P

```rs
error[E0502]: cannot borrow `*self` as mutable because it is also borrowed as immutable
   --> alacritty/src/display/hint.rs:154:17
    |
142 |         let hint = self.hint.as_ref()?;
    |                    ------------------ immutable borrow occurs here
...
154 |                 self.stop();
    |                 ^^^^^^^^^^^ mutable borrow occurs here
...
158 |             let action = hint.action.clone();
    |                          ------------------- immutable borrow later used here

For more information about this error, try `rustc --explain E0502`.
```

`stop` needs a mutable reference to `self`, but we're creating an immutable reference above.

SonuBardai: Ah, I see what the issue is. Calling `self.stop()` clears `self.hint` and `self.matches` ([ref](https://github.com/alacritty/alacritty/blob/7257c49313984f88c2f6b550a81ce838c75deda2/alacritty/src/display/hint.rs#L63-L68)). So, trying to access `self.matches[index].clone()` would cause a panic because of an index out of bounds. I think the order of the operations must be kept as is.. unless there's a different way to do this that I'm missing?

chrisduerr (changes requested):
Looks good other than this last formatting thing.

extra/man/alacritty.5.scd
```diff
@@ -585,6 +585,11 @@ _action_ or a _command_.
 		there are characters likely not to be part of the hint (e.g. a trailing
 		_._). This is most useful for URIs and applies only to _regex_ matches.
 
+	*persist* <boolean>
+
+		Set to _false_ by default. When this is _true_, hints remain persistent 
+		after selection.
```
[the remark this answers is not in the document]
SonuBardai: I've added it to the default section underneath the options. Lmk if we also want to add it this way ^

chrisduerr (approved):
Good catch on the other missing defaults, thanks.

PR #6993, opened 2023-06-11
Change the way we refer to IRC chat

kchibisov:
--

I'm not quite sure that I like what I wrote, but whatever... It's not like most know what IRC is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,10 @@ a daily driver.
 
 Precompiled binaries are available from the [GitHub releases page](https://github.com/alacritty/alacritty/releases).
 
+Join [`#alacritty`] on libera.chat if you have questions or looking for a quick help.
+
+[`#alacritty`]: https://web.libera.chat/gamja/?channels=#alacritty
+
 ## Features
 
 You can find an overview over the features available in Alacritty [here](./docs/features.md).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -100,13 +102,10 @@ good fit for Alacritty. This means you won't find things like tabs or splits
...
-
-Alacritty discussions can be found in `#alacritty` on Libera.Chat.
-
 ## License
 
 Alacritty is released under the [Apache License, Version 2.0].
 
 [Apache License, Version 2.0]: https://github.com/alacritty/alacritty/blob/master/LICENSE-APACHE
 [tmux]: https://github.com/tmux/tmux
+[`#alacritty`]: https://web.libera.chat/gamja/?channels=#alacritty
```
chrisduerr, line 111: Can we move this under the paragraph itself? I don't like moving the definitions to the end of the file it makes it rather difficult to find where things are actually used.

chrisduerr (changes requested):

README.md
kchibisov: Should we do the same wrt tmux?

chrisduerr: Probably. I'm not blocking the PR on it but it would make sense to have it consistent.

chrisduerr (approved):

Issue #4794, opened 2021-02-15
Windows: Use ConPty from OpenConsole instead of ConHost


levirs565:
I have some issue with Alacritty. Then, I try reproduce the issue with WezTerm and Windows Terminal. However, I cannot reproduce the issue in Wezterm and Windows Terminal. Even, I make WezTerm config equally with Alacritty. However, I am still cannot repoduce the issue. Then, It turns out that WezTerm and Windows Terminal use ConPty from OpenConsole whereas Alacritty use ConPty from ConHost.

OpenConsole is ConHost clone. Its maintained by Microsoft. OpenConsole source code is come from ConHost source code. OpenConsole have more feature than ConHost. You can see difference in https://github.com/microsoft/terminal/issues/8101. 

You can see how WezTerm use OpenConsole in this file https://github.com/wez/wezterm/blob/a034730373c341aa078765c081fcb32168b1974f/pty/src/win/psuedocon.rs .

My issue is gitui not rendered correctly **only when `$env:TERM`  is not blank**.  

![Annotation 2021-02-15 153022](https://user-images.githubusercontent.com/42236775/107922551-bfe87600-6fa2-11eb-8ad1-38d130e62a48.png)

When I set `$env:TERM` blank. It works.

![Annotation 2021-02-15 153022](https://user-images.githubusercontent.com/42236775/107923144-a693f980-6fa3-11eb-93a4-6919c06890c1.png)

This issue can be reproduced using PowerShell Core directly (Directly ConHost). 

![Annotation 2021-02-15 153556](https://user-images.githubusercontent.com/42236775/107923070-8d8b4880-6fa3-11eb-9a04-98f930124cb8.png)

When `$env:TERM` in PS Core is blank. Gitui work. This issue cannot reproduced in Windows Terminal and WezTerm even when `$env:TERM` is `xterm-256color`.

![Annotation 2021-02-15 154945](https://user-images.githubusercontent.com/42236775/107924394-72213d00-6fa5-11eb-8c31-4e8c439c2c38.png)

Then, I create experiment. I rename `conpty.dll` to `conpty_12.dll` in WezTerm installation. This will make WezTerm use ConHost. Yeah, WezTerm not render Gitui correclty when `$env:TERM` is `xterm-256color`.

![Annotation 2021-02-15 155220](https://user-images.githubusercontent.com/42236775/107924623-dc39e200-6fa5-11eb-86cf-91c56321c0e6.png)

[...]

PR #6994, opened 2023-06-12
Load and use conpty.dll if available

proski:
I was asked by @kchibisov to put this change in review. I realize that similar PRs were rejected in the past for different reasons. However, the sad fact is that Microsoft has not backported their conpty changes to Windows 10 even though they are many years old. And the same Microsoft is benefiting from the bundled `conpty.dll` in their Windows Terminal. They have no urgency to backport their fixes, but they are leaving Alacritty in the cold if it sticks to the stable API.

Windows 10 is still widespread in the corporate world, it's not my OS of choice by any means, but I'm trying to make my development environment comfortable.

The original change was made by @grueslayer but I touched pretty much every line in that patch. I did my best to get this patch in a good shape for it to be accepted. The missing parts are the documentation and the configuration. I'd like to get some feedback first.

If there are any concerns about the use of unstable API, let's consider how to mitigate them. We can have a configuration option with detailed comments about the implications. The option can be disabled by default. What would be a good name for the option? `use_experimental_conpty`? Should it be boolean or have full path to `conpty.dll`?

If there is an option and it's enabled, it would be good to warn the user if the `conpty.dll` loading fails and fallback is used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Warnings for unused configuration file options
 - Config option `persist` in `hints` config section
+- Support for dynamically loading conpty.dll on Windows
 
 ### Changed
 
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -1,14 +1,18 @@
+use log::info;
 use std::io::Error;
 use std::os::windows::io::IntoRawHandle;
 use std::{mem, ptr};
 
 use mio_anonymous_pipes::{EventedAnonRead, EventedAnonWrite};
 
-use windows_sys::core::PWSTR;
+use windows_sys::core::{HRESULT, PWSTR};
 use windows_sys::Win32::Foundation::{HANDLE, S_OK};
 use windows_sys::Win32::System::Console::{
     ClosePseudoConsole, CreatePseudoConsole, ResizePseudoConsole, COORD, HPCON,
 };
+use windows_sys::Win32::System::LibraryLoader::{GetProcAddress, LoadLibraryW};
+use windows_sys::{s, w};
+
 use windows_sys::Win32::System::Threading::{
     CreateProcessW, InitializeProcThreadAttributeList, UpdateProcThreadAttribute,
     EXTENDED_STARTUPINFO_PRESENT, PROCESS_INFORMATION, PROC_THREAD_ATTRIBUTE_PSEUDOCONSOLE,
@@ -20,9 +24,70 @@ use crate::event::{OnResize, WindowSize};
 use crate::tty::windows::child::ChildExitWatcher;
 use crate::tty::windows::{cmdline, win32_string, Pty};
 
+/// Load the pseudoconsole API from conpty.dll if possible, otherwise use the
+/// standard Windows API.
+///
+/// The conpty.dll from the Windows Terminal project
+/// supports loading OpenConsole.exe, which offers many improvements and
+/// bugfixes compared to the standard conpty that ships with Windows.
+///
+/// The conpty.dll and OpenConsole.exe files will be searched in PATH and in
+/// the directory where Alacritty's executable is located.
+type CreatePseudoConsoleFn =
+    unsafe extern "system" fn(COORD, HANDLE, HANDLE, u32, *mut HPCON) -> HRESULT;
+type ResizePseudoConsoleFn = unsafe extern "system" fn(HPCON, COORD) -> HRESULT;
+type ClosePseudoConsoleFn = unsafe extern "system" fn(HPCON);
+
+struct ConptyApi {
+    create: CreatePseudoConsoleFn,
+    resize: ResizePseudoConsoleFn,
+    close: ClosePseudoConsoleFn,
+}
+
+impl ConptyApi {
+    fn new() -> Self {
+        match Self::load_conpty() {
+            Some(conpty) => {
+                info!("Using conpty.dll for pseudoconsole");
+                conpty
+            },
+            None => {
+                // Cannot load conpty.dll - use the standard Windows API.
+                info!("Using Windows API for pseudoconsole");
+                Self {
+                    create: CreatePseudoConsole,
+                    resize: ResizePseudoConsole,
+                    close: ClosePseudoConsole,
+                }
+            },
+        }
+    }
+
+    /// Try loading ConptyApi from conpty.dll library.
+    fn load_conpty() -> Option<Self> {
+        type LoadedFn = unsafe extern "system" fn() -> isize;
+        unsafe {
+            let hmodule = LoadLibraryW(w!("conpty.dll"));
+            if hmodule == 0 {
+                return None;
+            }
+            let create_fn = GetProcAddress(hmodule, s!("CreatePseudoConsole"))?;
+            let resize_fn = GetProcAddress(hmodule, s!("ResizePseudoConsole"))?;
+            let close_fn = GetProcAddress(hmodule, s!("ClosePseudoConsole"))?;
+
+            Some(Self {
+                create: mem::transmute::<LoadedFn, CreatePseudoConsoleFn>(create_fn),
+                resize: mem::transmute::<LoadedFn, ResizePseudoConsoleFn>(resize_fn),
+                close: mem::transmute::<LoadedFn, ClosePseudoConsoleFn>(close_fn),
+            })
+        }
+    }
+}
+
 /// RAII Pseudoconsole.
 pub struct Conpty {
     pub handle: HPCON,
+    api: ConptyApi,
 }
 
 impl Drop for Conpty {
@@ -31,7 +96,7 @@ impl Drop for Conpty {
         // conout pipe has already been dropped by this point.
         //
         // See PR #3084 and https://docs.microsoft.com/en-us/windows/console/closepseudoconsole.
-        unsafe { ClosePseudoConsole(self.handle) }
+        unsafe { (self.api.close)(self.handle) }
     }
 }
 
@@ -39,6 +104,7 @@ impl Drop for Conpty {
 unsafe impl Send for Conpty {}
 
 pub fn new(config: &PtyConfig, window_size: WindowSize) -> Option<Pty> {
+    let api = ConptyApi::new();
     let mut pty_handle: HPCON = 0;
 
     // Passing 0 as the size parameter allows the "system default" buffer
@@ -50,7 +116,7 @@ pub fn new(config: &PtyConfig, window_size: WindowSize) -> Option<Pty> {
 
     // Create the Pseudo Console, using the pipes.
     let result = unsafe {
-        CreatePseudoConsole(
+        (api.create)(
             window_size.into(),
             conin_pty_handle.into_raw_handle() as HANDLE,
             conout_pty_handle.into_raw_handle() as HANDLE,
@@ -158,7 +224,7 @@ pub fn new(config: &PtyConfig, window_size: WindowSize) -> Option<Pty> {
     let conout = EventedAnonRead::new(conout);
 
     let child_watcher = ChildExitWatcher::new(proc_info.hProcess).unwrap();
-    let conpty = Conpty { handle: pty_handle as HPCON };
+    let conpty = Conpty { handle: pty_handle as HPCON, api };
 
     Some(Pty::new(conpty, conout, conin, child_watcher))
 }
@@ -170,7 +236,7 @@ fn panic_shell_spawn() {
 
 impl OnResize for Conpty {
     fn on_resize(&mut self, window_size: WindowSize) {
-        let result = unsafe { ResizePseudoConsole(self.handle, window_size.into()) };
+        let result = unsafe { (self.api.resize)(self.handle, window_size.into()) };
         assert_eq!(result, S_OK);
     }
 }
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -45,6 +45,7 @@ windows-sys = { version = "0.48", features = [
     "Win32_System_Console",
     "Win32_Foundation",
     "Win32_Security",
+    "Win32_System_LibraryLoader",
     "Win32_System_Threading",
     "Win32_System_WindowsProgramming",
 ]}


alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -20,9 +23,63 @@ use crate::event::{OnResize, WindowSize};
...
+/// standard Windows API. The conpty.dll from the Windows Terminal project
+/// supports loading OpenConsole.exe, which offers many improvements and
+/// bugfixes compared to the standard conpty that ships with Windows.
+///
+/// In particular, the new conpty.dll and OpenConsole.exe enable mouse
+/// reporting, which not yet supported by the Windows API on Windows 10 as of
+/// June 2023.
+///
+/// The conpty.dll and OpenConsole.exe files will be searched in PATH and in
+/// the directory where Alacritty executable is located.
```
chrisduerr, line 36:
```suggestion
/// Load the pseudoconsole API from conpty.dll if possible, otherwise use the
/// standard Windows API.
///
/// The conpty.dll from the Windows Terminal project
/// supports loading OpenConsole.exe, which offers many improvements and
/// bugfixes compared to the standard conpty that ships with Windows.
///
/// The conpty.dll and OpenConsole.exe files will be searched in PATH and in
/// the directory where Alacritty's executable is located.
```

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -20,9 +23,63 @@ use crate::event::{OnResize, WindowSize};
...
+struct ConptyApi {
+    create: unsafe extern "system" fn(COORD, HANDLE, HANDLE, u32, *mut HPCON) -> HRESULT,
+    resize: unsafe extern "system" fn(HPCON, COORD) -> HRESULT,
+    close: unsafe extern "system" fn(HPCON),
+}
+
+impl ConptyApi {
+    /// Try loading ConptyApi from conpty.dll library.
+    fn load_conpty() -> Option<Self> {
+        // Unsafe because of Windows API calls.
```
chrisduerr, line 46: This comment provides no value. If you want to make a comment on an `unsafe` block you should comment why it's **safe**. I don't need a comment to tell me an unsafe block contains unsafe code.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -20,9 +23,63 @@ use crate::event::{OnResize, WindowSize};
...
+                return None;
+            }
+            let create_fn = GetProcAddress(hmodule, s!("CreatePseudoConsole"))?;
+            let resize_fn = GetProcAddress(hmodule, s!("ResizePseudoConsole"))?;
+            let close_fn = GetProcAddress(hmodule, s!("ClosePseudoConsole"))?;
+
+            Some(Self {
+                create: mem::transmute(create_fn),
+                resize: mem::transmute(resize_fn),
+                close: mem::transmute(close_fn),
```
chrisduerr, line 59: What are the actual types here? You probably shouldn't be using transmute.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -20,9 +23,63 @@ use crate::event::{OnResize, WindowSize};
...
+            })
+        }
+    }
+
+    /// Return the conpty API.
+    fn new() -> Self {
+        if let Some(conpty) = Self::load_conpty() {
```
chrisduerr, line 66: This is `unwrap_or`.

chrisduerr (changes requested):

kchibisov:
We should also add an entry to the `FaQ` section in README wrt windows bugs and why you might want to try conpty.dll from windows terminal.

As well as a changelog entry.

alacritty_terminal/src/tty/windows/conpty.rs
proski: OK, removing.

alacritty_terminal/src/tty/windows/conpty.rs
proski: I would be happy to use something less intrusive, but I don't know if there is a way. Quoting the error message without the `transmute`:

```
error[E0308]: mismatched types
  --> alacritty_terminal\src\tty\windows\conpty.rs:54:25
   |
54 |                 create: create_fn,
   |                         ^^^^^^^^^ incorrect number of function parameters
   |
   = note: expected fn pointer `unsafe extern "system" fn(COORD, isize, isize, u32, *mut isize) -> i32`
              found fn pointer `unsafe extern "system" fn() -> isize`
```

The other three functions are also reported to have signature `unsafe extern "system" fn() -> isize`

kchibisov: Given that those are pointers, try `as _`. `std::mem::transmute` shouldn't be used on pointers generally, because you have alternative like I've mentioned, and it's error prone if you change pointer to something else.

proski: It doesn't work. I defined the needed types and made a function `build()` to allow the `as _` syntax, but the cast is still not allowed:
```
error[E0605]: non-primitive cast: `unsafe extern "system" fn() -> isize` as `unsafe extern "system" fn(COORD, isize, isize, u32, *mut isize) -> i32`
  --> alacritty_terminal\src\tty\windows\conpty.rs:61:30
   |
61 |             Some(Self::build(create_fn as _, resize_fn as _, close_fn as _))
   |                              ^^^^^^^^^^^^^^ invalid cast
```
Please see https://stackoverflow.com/questions/46134477/how-can-i-call-a-raw-address-from-rust

proski:
I've addressed the code changes. I'll submit the documentation part shortly.

alacritty_terminal/src/tty/windows/conpty.rs
kchibisov: Hm, yes.

Could we do something like that then:

```diff
diff --git a/alacritty_terminal/src/tty/windows/conpty.rs b/alacritty_terminal/src/tty/windows/conpty.rs
index 8770aa6b..e780a46c 100644
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -23,6 +23,31 @@ use crate::event::{OnResize, WindowSize};
 use crate::tty::windows::child::ChildExitWatcher;
 use crate::tty::windows::{cmdline, win32_string, Pty};
 
+type LoadedFn = unsafe extern "system" fn() -> isize;
+type CreatePseudoConsoleFn =
+    unsafe extern "system" fn(COORD, HANDLE, HANDLE, u32, *mut HPCON) -> HRESULT;
+type ResizePseudoConsoleFn = unsafe extern "system" fn(HPCON, COORD) -> HRESULT;
+type ClosePseudoConsoleFn = unsafe extern "system" fn(HPCON);
+
+/// RAII Pseudoconsole.
+pub struct Conpty {
+    pub handle: HPCON,
+    api: ConptyApi,
+}
+
+impl Drop for Conpty {
+    fn drop(&mut self) {
+        // XXX: This will block until the conout pipe is drained. Will cause a deadlock if the
+        // conout pipe has already been dropped by this point.
+        //
+        // See PR #3084 and https://docs.microsoft.com/en-us/windows/console/closepseudoconsole.
+        unsafe { (self.api.close)(self.handle) }
+    }
+}
+
+// The ConPTY handle can be sent between threads.
+unsafe impl Send for Conpty {}
+
 /// Load the pseudoconsole API from conpty.dll if possible, otherwise use the
 /// standard Windows API.
 ///
@@ -33,12 +58,24 @@ use crate::tty::windows::{cmdline, win32_string, Pty};
 /// The conpty.dll and OpenConsole.exe files will be searched in PATH and in
 /// the directory where Alacritty's executable is located.
 struct ConptyApi {
-    create: unsafe extern "system" fn(COORD, HANDLE, HANDLE, u32, *mut HPCON) -> HRESULT,
-    resize: unsafe extern "system" fn(HPCON, COORD) -> HRESULT,
-    close: unsafe extern "system" fn(HPCON),
+    create: CreatePseudoConsoleFn,
+    resize: ResizePseudoConsoleFn,
+    close: ClosePseudoConsoleFn,
 }
 
 impl ConptyApi {
+    /// Return the conpty API.
+    fn new() -> Self {
+        Self::load_conpty().unwrap_or({
+            // Cannot load conpty.dll - use the standard Windows API.
+            Self {
+                create: CreatePseudoConsole,
+                resize: ResizePseudoConsole,
+                close: ClosePseudoConsole,
+            }
+        })
+    }
+
     /// Try loading ConptyApi from conpty.dll library.
     fn load_conpty() -> Option<Self> {
         unsafe {
@@ -51,45 +88,14 @@ impl ConptyApi {
             let close_fn = GetProcAddress(hmodule, s!("ClosePseudoConsole"))?;
 
             Some(Self {
-                create: mem::transmute(create_fn),
-                resize: mem::transmute(resize_fn),
-                close: mem::transmute(close_fn),
+                create: mem::transmute::<LoadedFn, CreatePseudoConsoleFn>(create_fn),
+                resize: mem::transmute::<LoadedFn, ResizePseudoConsoleFn>(resize_fn),
+                close: mem::transmute::<LoadedFn, ClosePseudoConsoleFn>(close_fn),
             })
         }
     }
-
-    /// Return the conpty API.
-    fn new() -> Self {
-        Self::load_conpty().unwrap_or({
-            // Cannot load conpty.dll - use the standard Windows API.
-            Self {
-                create: CreatePseudoConsole,
-                resize: ResizePseudoConsole,
-                close: ClosePseudoConsole,
-            }
-        })
-    }
 }
 
-/// RAII Pseudoconsole.
-pub struct Conpty {
-    pub handle: HPCON,
-    api: ConptyApi,
-}
-
-impl Drop for Conpty {
-    fn drop(&mut self) {
-        // XXX: This will block until the conout pipe is drained. Will cause a deadlock if the
-        // conout pipe has already been dropped by this point.
-        //
-        // See PR #3084 and https://docs.microsoft.com/en-us/windows/console/closepseudoconsole.
-        unsafe { (self.api.close)(self.handle) }
-    }
-}
-
-// The ConPTY handle can be sent between threads.
-unsafe impl Send for Conpty {}
-
 pub fn new(config: &PtyConfig, window_size: WindowSize) -> Option<Pty> {
     let api = ConptyApi::new();
     let mut pty_handle: HPCON = 0;

```

chrisduerr (approved):
Should be good once the doc changes are in.

proski:
All comments addressed, including the documentation.

kchibisov (approved):
While the changes are LGTM. I'm not sure that the accent should be on mouse, while it's just a general issue with Windows and ConPty. It's also not clear where the users should get the conpty.dll. I'd assume they can't just use windows terminal for that?

proski:
Those are good points. I'm not familiar with other issues (apart from an `m` character appearing in Mingw64 vim on startup). The mouse support is a showstopper for me.

I was feeling uneasy about referring to Wezterm specifically, but I can do it. Windows Terminal ships `OpenConsole.exe` but not `conpty.dll`. I don't think it would be reasonable to ask users to build `conpty.dll` from the Windows Terminal sources.

Maybe Alacritty should just copy those files from WezTerm and ship them? It should be legal, they are under MIT license: https://github.com/wez/wezterm/tree/main/assets/windows/conhost

Perhaps it would be reasonable to have a configuration option in that case so that users can use the system conpty if they want (especially on Windows 11).

kchibisov:
I would suggest to maybe point to these files, but I'm not sure we should ship them. The windows users who care could bring them.

We might just add to a paragraph in FaQ saying from where you can obtain them, the one which is official repo and a 3rd party builds.

j4james:
FYI, there are plans to make conpty available as a nuget package at some point (at least that's my understanding). The issue tracking that is here: https://github.com/microsoft/terminal/issues/15065. Not that that helps you now, but just so you know this process should get easier in the future.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Warnings for unused configuration file options
 - Config option `persist` in `hints` config section
+- Support for loading conpty.dll on Windows to enable mouse reporting
```
chrisduerr, line 19:
```suggestion
- Support for dynamically loading conpty.dll on Windows
```

I agree that mentioning specific features this enables is not appropriate.

README.md
```diff
@@ -109,6 +109,17 @@ niceties like a GUI config editor.
...
+**_Why is mouse reporting not working on Windows?_**
+
+Alacritty for Windows is using programming interface known as conpty to
+communicate with the programs running in the terminal. That interface supports
+mouse events on Windows 11 but not yet on Windows 10.
+
+It is possible to make Alacritty to use the updated conpty code by copying
+files named conpty.dll and OpenConsole.exe to any directory in PATH or to the
+directory where the Alacritty executable is located. Those files can be found
+in other terminal emulators for Windows.
```
chrisduerr, line 121: There's numerous issues with the default ConPTY. This section should be way more generic.

It's also pointlessly detailed, nobody cares what it's called or how it works, they just want a fix.

chrisduerr (changes requested):

Grueslayer:
@proski The original author is @fredizzimo I only altered code from using winapi to windows _sys. 

„It's also pointlessly detailed, nobody cares what it's called or how it works, they just want a fix.“

1.5 years later after saying 

„I'm not going to lose any sleep over issues that Microsoft can easily solve themselves whenever they feel like it.“

does not match my personal opinion of good community, so please keep me out.

fredizzimo:
I still have my version which worked when I updated it 2 years ago https://github.com/fredizzimo/alacritty/commit/5121f4aaa09a921c236701c202a0f4c9a3276978 and the original version was made 2 and a half year ago here https://github.com/alacritty/alacritty/pull/4501

I don't need any credit for it, it's just a simple change, especially the first version. But I agree with @Grueslayer, if you are going to credit someone it should be the original author. But I don't share his opinions about the community. I have had several great interactions with @kchibisov, and especially today.

kchibisov:
I mean generally speaking, the conpty issue is affecting microsoft as well, because they still have cmd.exe and such, which I doubt link to conpty.dll or ship it, so they have broken application on their own.

As a popular project it could be an option to stand again a behavior which is obviously wrong, but in years it wasn't changed, given that it's not an issue for the microsoft. At least they sort of wanted to provide a package on nuget (not sure if it'll help Windows 10 users).

If the point is to have a trailer on the commit (`Suggested-by` then it's fine).

Anyway, that's rather offtopic given that the opinion changed and the patch is now accepted.

README.md
kchibisov: Maybe something like that

```markdown
**_I'm on Windows and applications don't work like in other terminals?_**

Alacritty on Windows is relying on the system libraries by default, which are
quite out of date most of the time, for example your mouse might not work in
vim on Windows 10. Other applications, like Windows Terminal, ship updated
versions of the required libraries which are known to work better, however
they are not officially provided by the microsoft [yet].

Alacritty doesn't ship those libraries, but it could still load them if they are
present in your `PATH`. Some other terminal emulators for windows do build their
own copies of such libraries, so you could take those libs from them.


[yet]: https://github.com/microsoft/terminal/issues/15065

```

fredizzimo:
No, we are referring to the message in the description that says:
> The original change was made by @Grueslayer

I believe both of us are requesting that to either be removed or giving credit to the right person. Could you do that @proski?

As for adding `Suggested-by` me, that's still wrong, since it was suggested by @levirs565 here https://github.com/alacritty/alacritty/issues/4794, I just implemented the suggestion. 

`Co-Authored-By:` would be more appropriate, if any credit are given since the code in this PR resembles https://github.com/fredizzimo/alacritty/commit/5121f4aaa09a921c236701c202a0f4c9a3276978 pretty strongly, it's just split in two functions and cleaned up slightly. I was very new to Rust when I wrote it and also did not care about the quality since the PR was already closed, which can be seen especially with the way I did the null termination.

Anyway, now hard feelings here, I just want the credits to be either correct or not given at all.

PS. I don't think the Windows console was ever broken (any more than it always was), it just did not use any features that needed the fixes from Conpty

proski:
Addressed all comments. Rebased on top of the master branch. No code changes.

kchibisov:
>PS. I don't think the Windows console was ever broken (any more than it always was), it just did not use any features that needed the fixes from Conpty

There was an edge case with CJK chars where writing them broke resizing or such. It's not about just mouse mode, conpty maintains the grid of the terminal with its own reflow and resize logic, so it's really complicated, it's also limiting performance.

README.md
proski: I don't think we should drop the file names, `conpty.dll`  and `OpenConsole.exe`. That would make the documentation less helpful. Also, I'm not sure `OpenConsole.exe` can be called a library. Please feel free to fix my text in `README.md`.

proski:
Rebased.
Updated credits.
Improved FAQ entry in `README.md`.
No code changes.

README.md
proski: I've applied some of the suggestions.

kchibisov (approved):

README.md
chrisduerr: I think this is still significantly too verbose. It should be easy to explain this in a single paragraph. It's not necessary to mention potential issues at all, if people are running into problems we just need to point them at things they can try. We also don't need to document in Alacritty how other terminal emulators behave.

When it comes to how to get this to work, we don't need to really mention anything in the FAQ at all. I'd just add a section to the install.md.

chrisduerr (changes requested):
See my comment on the docs.

proski:
Removed all changes to `README.md`. If any changes are needed, please feel free to make them without my participation.

chrisduerr (approved):

iamFIREcracker:
Hey, thanks for this: I am really forward to being able to use my mouse again to select a Vim windows/tmux pane.

Quick question: I am not familiar with the project release cycle, but when can we expect these changes to be released?  I have seen couple of tags, lately, but none of them seemeed to include this change (b9c886872d1202fc9302f68a0bedbb17daa35335).

Thanks again!

jimporter:
> You can always build yourself or trigger msi build in your repo (the way we do releases).

Done [here](https://github.com/jimporter/alacritty/actions/runs/5450531021) in a personal fork for anyone else who's impatient.

PR #7008, opened 2023-06-17
Suppress clippy warning due to known false positive

proski:
Reference: https://github.com/rust-lang/rust-clippy/issues/10893

Apparently, the clippy analysis started ignoring direct stores to the cloned object with the recent versions of Rust.


kchibisov:
We run clippy only on MSRV, so this bug is not affecting us(you can see CI passes). It's true that stable clippy linting is broken right now, but it was like that even before when lints were not compatible with the msrv.

PR #7010, opened 2023-06-17
Fix parsing of integer font size

proski:
Config file conversion broke parsing of the font size value if it's written as an integer. TOML integers are signed.


kchibisov (approved):

chrisduerr (changes requested):
This just allows negative values, which we should not support.

I'd have to look into the options myself, but the simplest solution would probably be to just add a custom error when the value is negative.

kchibisov:
@chrisduerr we already allow negative values due to accepting floating point values. We'd have to rewrite the logic for the floats as well, so given the fix preserved old bug with negative values, I've approved it as is.

chrisduerr (approved):
Yeah that's probably fair.

PR #7014, opened 2023-06-18
Update README.md

borngraced:
`man alacritty` specified config file with `.yml` ext


kchibisov:
Your man pages are for the version you have in release, not the current master. Master branch is using toml, thus you see `toml` in the readme.

tip: look at `git blame` if you're not sure why things the way they are.

Issue #2492, opened 2019-05-30
Alacritty Icon Jumping on Mac Dock


ngklingler:
Which operating system does the issue occur on? macOS

If on linux, are you using X11 or Wayland? Not relevant

Version: 0.3.2

On macOS the alacritty icon in the dock will repeatedly jump when alacritty is not focused. It seems most likely to do this if an SSH connection was broken, but even after I change focus to alacritty and run other commands successfully it continues to jump after I change focus. It seems the only way to resolve it is to close and reopen alacritty, which is a bit annoying.

If anyone knows of a solution or knows what in the code is causing this please let me know. Thanks in advance for any and all help!

ngklingler:
To reproduce run alacritty on macOS, SSH into a server, then disconnect your wifi or something to break the connection. When you switch focus to another application the icon should start jumping

chrisduerr:
That almost sounds like an urgent flag is set on the window, without it getting cleared properly.

Can you reproduce this by running `sleep 3 && echo -e "\a"`? That should ring the bell and be a little easier to reproduce than having to use ssh.

Is it correct that Alacritty will just continue to request focus non-stop, or does it stop for some period and then just start again later?

ngklingler:
So when I run what you just mentioned it will request focus, I switch focus to alacritty and that seems to clear it since when I switch focus away it stops jumping. But with SSH when I switch focus to alacritty it stops jumping until I switch away again, and it seems nonstop until I close and reopen alacritty

ngklingler:
FWIW instead of closing and reopening alacritty I can just open a new tmux window and close the previous one, or start a new tmux session. Possible tmux is repeatedly sending the terminal bell when SSH fails?

chrisduerr:
That definitely sounds a bit odd, if it was an Alacritty bug I would have expected the `\a` bell character to cause the same issues.

I'd assume you can't reproduce this with another terminal emulator though?

ngklingler:
Got it somewhat reproduced with another emulator. Terminal.app will jump once when I leave focus, but doesn't repeatedly jump like alacritty does. Regardless it seems like a tmux/ssh issue versus an alacritty issue. In the meantime do you happen to know of a way to disable this behavior? I have set the duration of the visual bell to 0 but it still occurs. If you don't know no worries, you've been super helpful in troubleshooting this issue and I appreciate that.

chrisduerr:
> In the meantime do you happen to know of a way to disable this behavior?

[...]

PR #7019, opened 2023-06-20
🦥 Disable icon jumping when a bell is received

boxmein:
This PR is essentially a manual fix for #2492 and #2950. It changes the terminal bell event handler so that it no longer sets the `"urgent"` flag on the window. macOS uses the urgent flag to detect if the dock icon should be jumping.

As discussed in #2950, alacritty will expect the inner process to set a terminal mode flag instead. However, in my experience, it is almost impossible to find what is sending the bell. For me, the optimal fix was just reacting to bells normally *without* causing the dock icon to jump.

I assume the PR isn't really a candidate for release - but it may help someone get rid of an annoying icon jump issue.

I track alacritty in my forked repository and add this commit on top, so that there's a version of alacritty without the dock icon jumping functionality.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1246,12 +1246,6 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                     },
                     TerminalEvent::Wakeup => *self.ctx.dirty = true,
                     TerminalEvent::Bell => {
-                        // Set window urgency hint when window is not focused.
-                        let focused = self.ctx.terminal.is_focused;
-                        if !focused && self.ctx.terminal.mode().contains(TermMode::URGENCY_HINTS) {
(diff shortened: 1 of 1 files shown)


kchibisov:
But you could set the mode to disable jumping from the `shell` launch/propmt and you'll never see it jumping regardless of what you run?

kchibisov:
Yes, the mentioned escape sequence.

kchibisov:
Your terminal here is tmux, you could disable bell in it.

boxmein:
It already is disabled :) my config file includes these lines:

```
set-option -g bell-action none
set-option -g visual-bell off
```

mrded:
TL;DR: https://github.com/alacritty/alacritty/pull/4304
```
$ printf "\e[?1042l"
```

PR #7034, opened 2023-06-26
Ignore scrolling multiplier when using touchscreen

4z3:
This causes the scrolling multiplier to be ignored when scrolling using the touchscreen. Without this change, scrolling is out of sync with the position of the finger on the touchscreen.

_(Before this patch, with the default scrolling multiplier of 3, scrolling appears to be in sync when scrolling in the primary buffer, but when scrolling in the alternate screen buffer, e.g. when using weechat or vim, then it will be too slow.)_

### DIFF
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -63,9 +63,6 @@ const MIN_SELECTION_SCROLLING_HEIGHT: f64 = 5.;
 /// Number of pixels for increasing the selection scrolling speed factor by one.
 const SELECTION_SCROLLING_STEP: f64 = 20.;
 
-/// Touch scroll speed.
-const TOUCH_SCROLL_FACTOR: f64 = 0.35;
-
 /// Distance before a touch input is considered a drag.
 const MAX_TAP_DISTANCE: f64 = 20.;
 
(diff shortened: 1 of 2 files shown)


chrisduerr (changes requested):
Why not just set the scroll multiplier to 1 in your config?

chrisduerr:
I didn't get around to testing it this weekend, so it'll have to wait a couple more days. Just wanted to let you know that it might take till the end of the week until I get around to it.

chrisduerr:
It looks like I won't be able to get around to this today either, so I'm not giving any further ETAs. But since this does affect me on my phone, I will definitely keep track of it and look into it as soon as my compositor is working properly again.

chrisduerr (approved):
Sorry for the massive delay, I finally got around to testing this on my phone and it is an absolutely massive improvement.

Thanks for the patch.

alacritty/src/input.rs
```diff
@@ -828,7 +827,7 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
                 let delta_y = touch.location.y - last_touch.location.y;
                 *touch_purpose = TouchPurpose::Scroll(touch);
 
-                self.scroll_terminal(0., delta_y * TOUCH_SCROLL_FACTOR);
+                self.scroll_terminal(0., delta_y, 1.0);
```
kchibisov, line 830: Could we have extra comment why it's done that way, so it won't get removed by accident?

PR #7048, opened 2023-07-01
features.md: Fix configuration link typo

NReilingh:
Saw a square bracket standing in for the letter 'c'; attempted to fix with a markdown relative link instead of just removing it.


chrisduerr (changes requested):
My goal was to remove the link so people wouldn't be driven to the poorly rendered scd and instead relied on the manpage.

So I think just removing the link is best.

chrisduerr (approved):

PR #7051, opened 2023-07-02
Fix legacy bindings taking precedence over new ones

### LINKED ISSUE #7050: Keybindings don't work on Linux after migration to TOML
I recently built version 13-dev and as suggested at the first start I did the migration of my config from YML to TOML.
I don't see regressions on Mac so far, but on Linux all of my keybindings stopped working.
To reproduce, I simplified my bindings to this:
```
[[keyboard.bindings]]
chars = "\u001B"
key = "Left"
mods = "None"
```

But, it doesn't work.

### System

OS: Linux
Version: alacritty 0.13.0-dev (4ad1aa32)
Linux/BSD: Wayland, Gnome, Fedora

### Logs

```
[2.008352824s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(94497932021376)), event: KeyboardInput { device_id: DeviceId(Wayland(DeviceId)), input: KeyboardInput { scancode: 105, state: Pressed, virtual_keycode: Some(Left), modifiers: (empty) }, is_synthetic: false } }
[2.083860136s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(94497932021376)), event: KeyboardInput { device_id: DeviceId(Wayland(DeviceId)), input: KeyboardInput { scancode: 105, state: Released, virtual_keycode: Some(Left), modifiers: (empty) }, is_synthetic: false } }

# cat
^[[D
```

kchibisov:
They were compared by len, but it's was wrong from the start, since a user provided binding could remove more than one builtin binding, so it was impossible for users to use their own bindings.

The most reliable way to do so is to use `Option`, given that we fill default during deserialization.

Fixes #7050.

### DIFF
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -80,11 +80,11 @@ pub struct UiConfig {
 
     /// Keybindings.
     #[config(deprecated = "use keyboard.bindings instead")]
-    key_bindings: KeyBindings,
+    key_bindings: Option<KeyBindings>,
 
     /// Bindings for the mouse.
     #[config(deprecated = "use mouse.bindings instead")]
-    mouse_bindings: MouseBindings,
+    mouse_bindings: Option<MouseBindings>,
 
     /// Configuration file imports.
     ///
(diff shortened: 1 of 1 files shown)


chrisduerr (approved):
I'm a bit confused why I didn't do this for the start, but if there was a reason, I forgot it. So LGTM.

PR #7055, opened 2023-07-05
Update mentions of key_bindings and mouse_bindings in the manual

proski:


extra/man/alacritty.5.scd
```diff
@@ -642,12 +642,12 @@ of the configuration file.
 
 		Mouse button which needs to be pressed to trigger this binding.
 
-	*action* <key_bindings.action> | "ExpandSelection"
+	*action* <keyboard_bindings_action> | "ExpandSelection"
```
kchibisov, line 645:
```suggestion
	*action* <keyboard.bindings_action> | "ExpandSelection"
```

kchibisov (changes requested):

kchibisov (approved):

chrisduerr (changes requested):
I think we should document keyboard/mouse as separate sections for the sake of consistency.

proski:
@chrisduerr Please explain. The documentation is already using separate sections. The only overlap is in the actions. I can move actions to a separate section.

chrisduerr:
`keyboard` and `mouse` are TOML tables and should be in dedicated sections. No other section is collapsed the way you did it.

kchibisov:
@proski if it's not clear what is wanted here, the mouse and bindings sections are both in the `# Bindings`, however what was suggested instead is that `mouse_bindings` will be moved to `# Mouse` section and will be called not `mouse.bindings`, but just `bindings`.

The same for keyboard, however the `# Bindings` should be renamed to `# Keyboard` at this point and `key_bindings` will become just `bindings`.

That should be just moving sections around.

The `# Keyboard` section should have entry docs like `# Mouse` one.

proski:
The manual sections reflect the top-level sections of the TOML file now.

kchibisov (approved):

chrisduerr (approved):

PR #7059, opened 2023-07-07
Escape file paths using single quotes when dragging and dropping file.

EsfoNL:


kchibisov:
Thanks for the patch, but this is not desired. Shell can clearly escape on its own, because text is written with the escape sequence. For example zsh can do auto escaping when pasting itself.

PR #7080, opened 2023-07-13
Mention all the actions in the man page

kchibisov:
A lot of actions were left without a notice, so the only way to figure out some of them was to browse source code. An example of such actions were `Maximize` and `Minimize`, since we don't have a binding for them, so they were left even in alacritty-bindings(5).

Explicitly list all the bindings we have. The search bindings were also not accuratelly restricted, since we allow them to be executed inside regardless of mode.

### DIFF
--- a/extra/man/alacritty.5.scd
+++ b/extra/man/alacritty.5.scd
@@ -689,14 +689,70 @@ https://docs.rs/winit/\*/winit/keyboard/enum.Key.html
 
 	*action*
 
+		*ReceiveChar*
+			Allow receiving char input.
+		*None*
+			No action.
+		*Paste*
+			Paste contents of system clipboard.
+		*Copy*
+			Store current selection into clipboard.
+		*IncreaseFontSize*
+			Increase font size.
+		*DecreaseFontSize*
+			Decrease font size.
+		*ResetFontSize*
+			Reset font size to the config value.
+		*ScrollPageUp*
+			Scroll exactly one page up.
+		*ScrollPageDown*
+			Scroll exactly one page down.
+		*ScrollHalfPageUp*
+			Scroll half a page up.
+		*ScrollHalfPageDown*
+			Scroll half a page down.
+		*ScrollLineUp*
+			Scroll one line up.
+		*ScrollLineDown*
+			Scroll one line down.
+		*ScrollToTop*
+			Scroll all the way to the top.
+		*ScrollToBottom*
+			Scroll all the way to the bottom.
+		*ClearHistory*
+			Clear the display buffer(s) to remove history.
+		*Hide*
+			Hide the Alacritty window.
+		*Minimize*
+			Minimize the Alacritty window.
+		*Quit*
+			Quit Alacritty.
+		*ClearLogNotice*
+			Clear warning and error notices.
+		*SpawnNewInstance*
+			Spawn a new instance of Alacritty.
+		*CreateNewWindow*
+			Create a new Alacritty window.
+		*ToggleFullscreen*
+			Toggle fullscreen.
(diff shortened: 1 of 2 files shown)


extra/man/alacritty.5.scd
```diff
@@ -689,75 +689,123 @@ https://docs.rs/winit/\*/winit/keyboard/enum.Key.html
...
+		*ToggleViMode*
+			Toggle vi mode.
+		*ReceiveChar*
+			Allow receiving char input.
+		*SearchForward*
+			Start a forward buffer search.
+		*SearchBackward*
+			Start a backward buffer search.
+		*None*
+			No action.
```
kchibisov, line 747: Not sure whether we should mention this one here as well.

chrisduerr (changes requested):

extra/man/alacritty.5.scd
```diff
@@ -689,75 +689,123 @@ https://docs.rs/winit/\*/winit/keyboard/enum.Key.html
...
+			Move to end of previous semantically separated word.
+		WordLeft*
+			Move to end of semantically separated word.
+		WordRight*
+			Move to start of whitespace separated word.
+		WordLeftEnd*
+			Move to start of next whitespace separated word.
+		WordRightEnd*
+			Move to end of previous whitespace separated word.
+		Bracket*
```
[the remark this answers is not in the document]
kchibisov: Also, should we somehow emphasize that it's all about Vi cursors?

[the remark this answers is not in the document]
kchibisov: I got it wrong because we had different spacing in enums... Now it should be fine.

extra/man/alacritty.5.scd
chrisduerr: For sure, I'd actually move `ReceiveChar` and `None` to the top probably?

extra/man/alacritty.5.scd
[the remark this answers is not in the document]
chrisduerr: We could add a subsection with a little "vi mode bindings" header.

chrisduerr (changes requested):

extra/man/alacritty.5.scd
```diff
@@ -689,14 +689,73 @@ https://docs.rs/winit/\*/winit/keyboard/enum.Key.html
...
+		*ClearSelection*
+			Clear active selection.
+		*ToggleViMode*
+			Toggle vi mode.
+		*SearchForward*
+			Start a forward buffer search.
+		*SearchBackward*
+			Start a backward buffer search.
+
+		_Extra vi mode actions:_
```
kchibisov, line 749: I wanted to emphasize that it's extra actions and not only actions. So you could still do other actions in it.

chrisduerr: Maybe rather than "extra vi mode actions" we should name it "Vi mode exclusive actions"?

That's how we've talked about this in the past.

chrisduerr (changes requested):

extra/man/alacritty.5.scd
kchibisov: The thing is that you can bind them outside Vi mode due to pipelining we've added.

chrisduerr: But they're exclusive to the vi mode. They only work in vi mode.

kchibisov: Maybe just `Vi mode actions` then?

chrisduerr: Sounds good to me.

chrisduerr (approved):

Issue #6859, opened 2023-04-14
`colors.transparent_background_colors` does not apply to selections


imsuck:
### System
OS: Linux
Version: alacritty 0.12.0 (5a728195)
Linux/BSD: X11, picom and i3

The configuration option `colors.transparent_background_colors` does not not apply to selections. It doesn't matter whether you set `colors.selection.background` or not.

No selections
![image](https://user-images.githubusercontent.com/49095435/231982808-10474589-e5d7-4d40-97da-a2e034052b22.png)
Selection without setting `colors.selection.background`
![image](https://user-images.githubusercontent.com/49095435/231982893-f3ad3036-1a08-4c3c-9fd3-fe16d2959b4b.png)
Selection with `colors.selection.background`
![image](https://user-images.githubusercontent.com/49095435/231982993-15881cfb-a22a-40a0-87f1-446e67ca1876.png)

SonuBardai:
question: Is this the issue? We expect the selected color to have the same alpha value as the bg? 

https://github.com/alacritty/alacritty/assets/67749330/67724d23-48d1-46df-955a-8e7f88c0efd8

kchibisov:
@SonuBardai yes, the `window.opacity` value should be used for selection as well.

kchibisov:
Well, you could open a PR and we could discuss it there.

PR #7085, opened 2023-07-15
Fix: apply transparent-background-colors for selections

SonuBardai:
Closes https://github.com/alacritty/alacritty/issues/6859

https://github.com/alacritty/alacritty/assets/67749330/c57bf30d-a182-4033-a568-ac7ebfa89a29

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - IME popup now tries to not obscure the current cursor line
 - The double click threshold was raised to `400ms`
 - OSC 52 paste ability is now **disabled by default**; use `terminal.osc52` to adjust it
+- Apply `colors.transparent_background_colors` for selections, hints, and search matches
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Default Vi key bindings for `Last`/`First` actions not working on X11/Wayland
 - Cut off wide characters in preedit string
 - Scrolling on touchscreens
-- Applying `colors.transparent_background_colors` for selections, hints, and search matches
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Default Vi key bindings for `Last`/`First` actions not working on X11/Wayland
 - Cut off wide characters in preedit string
 - Scrolling on touchscreens
+- Fixed applying `colors.transparent_background_colors` for selections
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Default Vi key bindings for `Last`/`First` actions not working on X11/Wayland
 - Cut off wide characters in preedit string
 - Scrolling on touchscreens
-- Fixed applying `colors.transparent_background_colors` for selections, hints, and search matches
+- Applying `colors.transparent_background_colors` for selections, hints, and search matches
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Default Vi key bindings for `Last`/`First` actions not working on X11/Wayland
 - Cut off wide characters in preedit string
 - Scrolling on touchscreens
-- Fixed applying `colors.transparent_background_colors` for selections
+- Fixed applying `colors.transparent_background_colors` for selections, hints, and search matches
 
 ### Removed
 
--- a/alacritty/src/display/content.rs
+++ b/alacritty/src/display/content.rs
@@ -232,27 +232,13 @@ impl RenderableCell {
             } else {
                 (colors.hints.end.foreground, colors.hints.end.background)
             };
-            Self::compute_cell_rgb(
-                content.config,
-                &mut fg,
-                &mut bg,
-                &mut bg_alpha,
-                config_fg,
-                config_bg,
-            );
+            Self::compute_cell_rgb(&mut fg, &mut bg, &mut bg_alpha, config_fg, config_bg);
 
             character = c;
         } else if is_selected {
             let config_fg = colors.selection.foreground;
             let config_bg = colors.selection.background;
-            Self::compute_cell_rgb(
-                content.config,
-                &mut fg,
-                &mut bg,
-                &mut bg_alpha,
-                config_fg,
-                config_bg,
-            );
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `Maximized` startup mode not filling the screen properly on GNOME Wayland
 - `OptionAsAlt` with `OnlyLeft`/`OnlyRight` settings not working properly on macOS
 - Default Vi key bindings for `Last`/`First` actions not working on X11/Wayland
+- Fixed applying `colors.transparent_background_colors` for selections
```
SonuBardai, line 43: I wasn't sure if all fixes go into the changelog. Lmk if the fix isn't big enough to have here and I can remove it.

alacritty/src/display/content.rs
```diff
@@ -239,6 +239,9 @@ impl RenderableCell {
             let config_fg = colors.selection.foreground;
             let config_bg = colors.selection.background;
             Self::compute_cell_rgb(&mut fg, &mut bg, &mut bg_alpha, config_fg, config_bg);
+            if content.config.colors.transparent_background_colors {
```
SonuBardai, line 242: I went with the simplest solution. The alternative I thought of was to add `transparent_background_colors` as an optional argument to the `compute_cell_rgb`, and we change the `compute_cell_rgb` to something like this:

```rust
    fn compute_cell_rgb(
        cell_fg: &mut Rgb,
        cell_bg: &mut Rgb,
        bg_alpha: &mut f32,
        fg: CellRgb,
        bg: CellRgb,
        transparent_background_colors: bool
    ) {
        let old_fg = mem::replace(cell_fg, fg.color(*cell_fg, *cell_bg));
        *cell_bg = bg.color(old_fg, *cell_bg);

        if !transparent_background_colors && bg != CellRgb::CellBackground {
            *bg_alpha = 1.0;
        }
    }
```

And while calling this on Line 241 here we do:
```rust
Self::compute_cell_rgb(&mut fg, &mut bg, &mut bg_alpha, config_fg, config_bg, true);
```

chrisduerr (changes requested):
I think this fails to sufficiently address the issue. There's still many other branches in this function where transparent background isn't handled properly.

SonuBardai:
> I think this fails to sufficiently address the issue. There's still many other branches in this function where transparent background isn't handled properly.

Cool! Thanks for that headsup @chrisduerr. I think this was the branch you're referring to? ie. if the fg value is set the same as the bg value. This is after the change (and with fg=bg set):

Screenshot 2023-07-18 at 10 22 25

chrisduerr:
> Cool! Thanks for that headsup @chrisduerr. I think this was the branch you're referring to? ie. if the fg value is set the same as the bg value. This is after the change (and with fg=bg set):

No, I'm talking about multiple scenarios. Search for example also doesn't check for this.

SonuBardai:
Do we also want the alpha value to apply to Hints? There might be a much easier solution if we want transparency to be applied to all branches.
```rust
if let Some((c, is_first)) = content.hint.as_mut().and_then(|hint| hint.advance(viewport_start, cell.point)) {
    // Hints
} else if is_selected {
    // Selection
} else if content.search.as_mut().map_or(false, |search| search.advance(cell.point)) {
    // Search results
    let (config_fg, config_bg) = if focused {
        // Focused search results
    } else {
        // Other search results
    };
}
// All other renderable cells
```

kchibisov:
I think we want to apply alpha to every background, as the option states. So hints, search, etc.

alacritty/src/display/content.rs
```diff
@@ -287,19 +286,9 @@ impl RenderableCell {
...
-        fg: CellRgb,
-        bg: CellRgb,
-    ) {
+    fn compute_cell_rgb(cell_fg: &mut Rgb, cell_bg: &mut Rgb, fg: CellRgb, bg: CellRgb) {
         let old_fg = mem::replace(cell_fg, fg.color(*cell_fg, *cell_bg));
         *cell_bg = bg.color(old_fg, *cell_bg);
-
-        if bg != CellRgb::CellBackground {
-            *bg_alpha = 1.0;
-        }
```
SonuBardai, line 302: Is there a reason we do this? I understand that if the bg is different from the actual background then we're setting the alpha to 1.0, but this was the main reason why hints, selections, and search matches were opaque

alacritty/src/display/content.rs
```diff
@@ -366,10 +355,10 @@ impl RenderableCell {
     /// is computed.
     #[inline]
     fn compute_bg_alpha(config: &UiConfig, bg: Color) -> f32 {
-        if bg == Color::Named(NamedColor::Background) {
-            0.
-        } else if config.colors.transparent_background_colors {
+        if config.colors.transparent_background_colors {
             config.window_opacity()
```
SonuBardai, line 359: Is there a need to having this order for the checks? I wasn't able to understand why we do it this way

chrisduerr: If it's the same as the background color, we're not drawing anything anyway. So it doesn't matter what you set your `window_opacity` to. Making it *less* transparent than 0 will just cause it to be rendered, which is a waste of time since you can't see it anyway.

alacritty/src/display/content.rs
SonuBardai: Thankyou for the explanation!

SonuBardai:
Updates: We're passing the UI config to the `compute_cell_rgb` function. I reverted the changes from the previous solution and now we're keeping the `bg_alpha` at 0 if the bg is same as the cell bg. If it's different then we check if the config has `transparent_background_colors` turned on and assign the alpha value based on it. 


https://github.com/alacritty/alacritty/assets/67749330/623c8443-eb11-4864-9e78-f67ccdd6ad63

chrisduerr (changes requested):
I've been thinking, is there actually any scenario where `config.colors.transparent_background_color` is set and we do **not** want a background to be transparent? I can't really think of one.

It might be easiest to just check if the config is set at the end and if not, we override the alpha. Though we should make sure that the alpha isn't already smaller (realistically the only smaller value possible will be zero, so checking for that should be fine too).

SonuBardai:
https://github.com/alacritty/alacritty/assets/67749330/6896d574-ab93-49a5-a9fe-d68907c65895

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `OptionAsAlt` with `OnlyLeft`/`OnlyRight` settings not working properly on macOS
 - Default Vi key bindings for `Last`/`First` actions not working on X11/Wayland
 - Cut off wide characters in preedit string
+- Applying `colors.transparent_background_colors` for selections, hints, and search matches
```
chrisduerr, line 45: I think this fits better into `Changed`.

alacritty/src/display/content.rs
```diff
@@ -256,6 +256,10 @@ impl RenderableCell {
             Self::compute_cell_rgb(&mut fg, &mut bg, &mut bg_alpha, config_fg, config_bg);
         }
 
+        if bg_alpha > 0. && content.config.colors.transparent_background_colors {
+            bg_alpha = content.config.window_opacity();
+        }
```
chrisduerr, line 261: A comment would be helpful summarizing what this block does.

chrisduerr (changes requested):

chrisduerr (approved):

chrisduerr (approved):

PR #7092, opened 2023-07-18
Use bracketed paste only for multi-char IME input

kchibisov:
Some IME setups do only `commit` single char input, like fcitx5 when doing ru input.


chrisduerr (changes requested):
Can you remind me what the motivation behind this was?

kchibisov:
@chrisduerr I think commit message has an example? I mainly want to have single char input go without bracketed paste, so it's less payload on editor for no reason.

However for large preedit lines, we want bracketed paste, so the highlighting logic and such won't run on a per char basis.

chrisduerr (approved):

PR #7093, opened 2023-07-18
Skip whitespaces for wide chars in preedit

kchibisov:
While we skip the spacers for the wide characters in the grid due to them having a proper flags, the draw_string method was generating the cells with incorrect flags leading to wide chars being cut off.

--

I'm not sure I like this fix.

### DIFF
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -8,6 +8,7 @@ use glutin::context::{ContextApi, GlContext, PossiblyCurrentContext};
 use glutin::display::{GetGlDisplay, GlDisplay};
 use log::{debug, error, info, warn, LevelFilter};
 use once_cell::sync::OnceCell;
+use unicode_width::UnicodeWidthChar;
 
 use alacritty_terminal::index::Point;
 use alacritty_terminal::term::cell::Flags;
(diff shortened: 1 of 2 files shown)


kchibisov:
CI failures seem unrelated.

alacritty/src/renderer/mod.rs
```diff
@@ -175,15 +176,30 @@ impl Renderer {
...
-            bg_alpha: 1.0,
-            fg,
-            bg,
-            underline: fg,
+        let mut skip_next = false;
+        let cells = string_chars.enumerate().filter_map(|(i, character)| {
+            if skip_next {
+                skip_next = false;
+                return None;
+            }
```
chrisduerr, line 184: Isn't it easier to just increment i twice for wide chars?

chrisduerr (changes requested):

alacritty/src/renderer/mod.rs
kchibisov: But it's a part of enumerate and we need to skip a space? Like the issue is that we have a space after a wide char here.

I decided to avoid rewriting all of the handling to do `(char, width)` like iterators because it was getting really messy.

chrisduerr: Nobody forces you to use enumerate? Seems simler to roll your own loop.

kchibisov: yeah, but you need to skip `space` regardless? Like the point is to skip next character and not advance the `i` with `+2`.

chrisduerr (approved):

PR #7094, opened 2023-07-18
Don't use mutable references unnecessarily

proski:
This fixes all issues found by the nightly clippy.


chrisduerr (approved):
Usually I'd recommend against making nightly clippy changes since they can break oldstable, but these seem rather uneventful and sensible so I suppose it should be fine.

PR #7101, opened 2023-07-21
Make URL_REGEX more strict

SomeoneToIgnore:
Current URL_REGEX matches things like `http::trace::on_request::log_parameters` which are not valid URLs, the patch fixes this, by requiring `//` after all protocols that should have it.

### DIFF
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -25,7 +25,7 @@ use crate::config::window::WindowConfig;
 
 /// Regex used for the default URL hint.
 #[rustfmt::skip]
-const URL_REGEX: &str = "(ipfs:|ipns:|magnet:|mailto:|gemini://|gopher://|https://|http://|news:|file://|git://|ssh:|ftp://)\
+const URL_REGEX: &str = "(ipfs:|ipns:|magnet:|mailto:|gemini://|gopher://|https://|http://|news:|file:|git://|ssh:|ftp://)\
                          [^\u{0000}-\u{001F}\u{007F}-\u{009F}<>\"\\s{-}\\^⟨⟩`]+";
 
 #[derive(ConfigDeserialize, Clone, Debug, PartialEq)]
@@ -550,11 +550,11 @@ impl Eq for LazyRegexVariant {}
 
 #[cfg(test)]
 mod tests {
+    use super::*;
+
     use alacritty_terminal::term::test::mock_term;
 
     use crate::display::hint::visible_regex_match_iter;
-
-    use super::*;
 
     #[test]
     fn positive_url_parsing_regex_test() {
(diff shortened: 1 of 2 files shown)


SomeoneToIgnore:
Thank you for the man pages pointer, I've also updated the dev guide to mention this bit.

I don't have strong intentions to merge this, feel free to close this if you want to.
Originally, I copied this regex from this repo since I wanted to match regex in other terminal and thought that this regex will get me covered when matching selections on hover — and when found out that there's only a small bit missing, decided to upstream this.

CONTRIBUTING.md
```diff
@@ -86,7 +86,7 @@ Latency is another important factor for Alacritty. On X11, Windows, and macOS th
 Code should be documented where appropriate. The existing code can be used as a guidance here and
 the general `rustfmt` rules can be followed for formatting.
 
-If any change has been made to the `config.rs` file, it should also be documented in the man pages.
+If any change has been made to the `config.rs`, `ui_config.rs` files, it should also be documented in the man pages.
```
kchibisov, line 89: I'm not sure that this should be changed that way.  And I don't think it's a part of this patch in the first place.

SomeoneToIgnore: Reverted.

kchibisov (approved):

kchibisov:
I'd still wait for @chrisduerr to ack this one though.

alacritty/src/config/ui_config.rs
```diff
@@ -25,7 +25,7 @@ use crate::config::window::WindowConfig;
 
 /// Regex used for the default URL hint.
 #[rustfmt::skip]
-const URL_REGEX: &str = "(ipfs:|ipns:|magnet:|mailto:|gemini:|gopher:|https:|http:|news:|file:|git:|ssh:|ftp:)\
+const URL_REGEX: &str = "(ipfs:|ipns:|magnet:|mailto:|gemini://|gopher://|https://|http://|news:|file://|git://|ssh:|ftp://)\
```
chrisduerr, line 28: I'm not sure `file://` should be enforced. I wouldn't be surprised to see `file:/home/user/whatever` in the wild and it generally works just fine.

alacritty/src/config/ui_config.rs
```diff
@@ -547,3 +547,60 @@ impl PartialEq for LazyRegexVariant {
...
 }
 impl Eq for LazyRegexVariant {}
+
+#[cfg(test)]
+mod tests {
+    use alacritty_terminal::term::test::mock_term;
+
+    use crate::display::hint::visible_regex_match_iter;
+
+    use super::*;
```
chrisduerr, line 557: Super should be the first import.

chrisduerr (changes requested):

alacritty/src/config/ui_config.rs
SomeoneToIgnore: Sure, done.

alacritty/src/config/ui_config.rs
SomeoneToIgnore: Fixed.

chrisduerr (approved):

PR #7106, opened 2023-07-22
Use ahash instead of fnv and regular hash function

kchibisov:
After evaluation of the ahash with the data alacritty uses it was discovered that it's 1.5-2x times faster when getting the already hashed values, which is the primary cases for alacritty's renderer.

Given that ahash is generally faster, all the HashSet and HashMap's inside the alacritty were changed to use it as a hasher function.

--

You can play with the benchmark here https://git.sr.ht/~kchibisov/hash-bench it uses `cargo bench`.

My results were

```
running 6 tests
test benches::ahash_get    ... bench:      38,873 ns/iter (+/- 1,023)
test benches::ahash_insert ... bench:      84,969 ns/iter (+/- 2,716)
test benches::fnv_get      ... bench:      57,062 ns/iter (+/- 982)
test benches::fnv_insert   ... bench:      90,400 ns/iter (+/- 1,016)
test benches::std_get      ... bench:     164,550 ns/iter (+/- 3,032)
test benches::std_insert   ... bench:     238,054 ns/iter (+/- 7,363)
```

As you can see, the `get` perf of `ahash` is way better than other options.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,9 +9,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "ahash"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
+dependencies = [
+ "cfg-if 1.0.0",
+ "getrandom",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
 name = "alacritty"
 version = "0.13.0-dev"
 dependencies = [
+ "ahash",
  "alacritty_config",
  "alacritty_config_derive",
  "alacritty_terminal",
@@ -23,7 +36,6 @@ dependencies = [
  "crossfont",
  "dirs",
  "embed-resource",
- "fnv",
  "gl_generator",
  "glutin",
  "home",
--- a/alacritty/Cargo.toml
+++ b/alacritty/Cargo.toml
@@ -23,11 +23,11 @@ path = "../alacritty_config"
 version = "0.1.2-dev"
 
 [dependencies]
+ahash = { version = "0.8.3", features = ["no-rng"] }
 bitflags = "2.2.1"
 clap = { version = "4.2.7", features = ["derive", "env"] }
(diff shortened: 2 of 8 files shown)


notgull:
`rustc-hash` and `nohash` are also worth consideration in my opinion. They're generally faster than `ahash` but have quadratic worst case times in certain conditions.

kchibisov:
Hm, `nohash` is not an option, since we do hash strings in some places, but with `rustc_hash` the perf is:

```
running 8 tests
test benches::ahash_get         ... bench:      38,139 ns/iter (+/- 1,172)
test benches::ahash_insert      ... bench:      84,573 ns/iter (+/- 1,167)
test benches::fnv_get           ... bench:      56,900 ns/iter (+/- 1,658)
test benches::fnv_insert        ... bench:      90,352 ns/iter (+/- 2,182)
test benches::rustc_hash_get    ... bench:      29,420 ns/iter (+/- 266)
test benches::rustc_hash_insert ... bench:      76,016 ns/iter (+/- 696)
test benches::std_get           ... bench:     161,716 ns/iter (+/- 1,001)
test benches::std_insert        ... bench:     236,779 ns/iter (+/- 1,968)

```

The benchmark only considers the `renderer` structure, but given that most of the data is like we have in the renderer, it seems like `rustc_hash` could be better for us.

kchibisov:
Though, maybe the benchmark I've used is not that representative, but `ahash` also compares all of them https://github.com/tkaitchuck/aHash/blob/master/compare/readme.md#Speed

The `FxHash` is what the `rustc_hash` is, and it's as they say good on primitives of certain types, but not in general. In alarcitty case, we never hash a structure with
nulls, like you can't really have them in our types, so that's why the `rustc_hash` worked for us. But I'd argue that the difference is not that big compared to `fnv` hash in absolute values.

chrisduerr (changes requested):
Why did you even test this? I'd imagine the impact on Alacritty's performance for this change is negligible.

kchibisov:
I was mostly curious, but why use slower code in the first place?

chrisduerr (changes requested):
I'd much prefer using STD's `HashMap` with just setting the random state/hasher.

kchibisov:
@chrisduerr it's a `std` hashmap though, it's just a type alias.

chrisduerr:
> @chrisduerr it's a std hashmap though, it's just a type alias.

Which is why I'd prefer using STD's hashmap, instead of hiding this behind a type alias.

chrisduerr (approved):

Issue #6999, opened 2023-06-13
Example toml config


kenricqq:
With the migration from yml over to toml, settings can look quite different. Something like alacritty.toml as an example would be very helpful.

chrisduerr:
What exactly would you need that example for? If your goal is to migrate your own configuration file, I would recommend using `alacritty migrate`.

kchibisov:
Also, if there're man pages, which are not that bad rendered even on github. So the need for a config is questionable, given that it's extra maintenance.

kenricqq:
Thanks for the tip, and def recommend you guys put `alacritty migrate` in readme.md.

As for example page, yes, it might mean extra maintenance, but it's going to attract more new users in the long run due to the increased ease of use of having an example. Anyway, just a suggestion for the taking.

kchibisov:
it's in the changelog, but it's probably better to put it in the `warning`.

chrisduerr:
Not interested in adding an example configuration file at this point.

cideM:
Since this is the issue people will stumble on when they can no longer find the convenient example configuration file in the root repository, I'll just add this here.

At the time of writing this comment you are unlikely to have an Alacritty version installed that already ships the new man pages. So just go [here](https://github.com/alacritty/alacritty/blob/master/extra/man/alacritty.5.scd) to read the man page on the configuration on GitHub.

Eventually you'll be able to get there with `man 5 alacritty`.

kchibisov:
You're using alacritty version not supporting toml yet, you need to build alacritty yourself if you want a toml config.

0323pin:
If people really need an example, feel free to [take what you want](https://codeberg.org/pin/netbsd-marswm/raw/branch/main/.config/alacritty/alacritty.toml) 

I'm always running my builds from git-HEAD and had no issues migrating to `toml`.
Regards

antonwnk:
Ugh I see, bad oversight on my side. Assumed toml was the recommended config already in 12.1
Sorry and thanks for the answers!

tonymcneil:
[...]

PR #7107, opened 2023-07-22
Use inline toml for defaults in alacritty(5)

kchibisov:
Alacritty doesn't provide TOML config file by default and documents everything inside the man page. The choice was made to improve the delivery of the config documentation updates as well as defaults update to the end users, given they don't copy the config anymore.

However for some users man pages were harder to work with, given they couldn't simply copy paste anything into their config files anymore and have to construct everything based on the docs, which takes a lot of time. This commit aims to solve exactly that issue by inlining TOML tables and also providing examples for sections which have complicated defaults, so the users could simply copy paste them into their configs, because they are valid TOML.

Links: https://github.com/alacritty/alacritty/issues/6999

--

I'm not sure how to deliver cross platform defaults inside the inline TOML. Maybe provide a comment like `# command = "xdg_open"`? Because for example hints section default won't work unless you put the `command` into it.

### DIFF
--- a/extra/man/alacritty.5.scd
+++ b/extra/man/alacritty.5.scd
@@ -13,7 +13,7 @@ can be found at https://toml.io/en/v1.0.0.
 
 This section documents the root level of the configuration file.
 
-*import* [<string>,]
+*import* = [_"<string>"_,]
 
 	Import additional configuration files
 
@@ -24,7 +24,13 @@ This section documents the root level of the configuration file.
 	All imports must either be absolute paths starting with _/_, or paths
 	relative to the user's home directory starting with _~/_.
 
-*shell* <string> | { program = <string>, args = [<string>,] }
+	Example:
+		import = [++
+  _"~/.config/alacritty/base16-dark.toml"_,++
+  _"~/.config/alacritty/keybindings.toml"_,++
+]
+
+*shell* = _"<string>"_ | { program = _"<string>"_, args = [_"<string>"_,] }
 
 	You can set _shell.program_ to the path of your favorite shell, e.g.
 	_/bin/zsh_. Entries in _shell.args_ are passed as arguments to the shell.
@@ -33,20 +39,25 @@ This section documents the root level of the configuration file.
 		Linux/BSD/macOS: _$SHELL_ or the user's login shell, if _$SHELL_ is unset++
 Windows: _"powershell"_
 
-*working_directory* <string> | "None"
+	Example:
+		*[shell]*++
+program = _"/bin/zsh"_++
+args = [_"-l"_]
+
+*working_directory* = _"<string>"_|_"None"_
 
 	Directory the shell is started in. When this is unset, or _None_, the
 	working directory of the parent process will be used.
 
-	Default: _None_
+	Default: _"None"_
 
-*live_config_reload* <boolean>
+*live_config_reload* = _true_|_false_
 
 	Live config reload (changes require restart)
 
 	Default: _true_
 
-*ipc_socket* <boolean> _(unix only)_
+*ipc_socket* = _true_|_false_ # _(unix only)_
 
 	Offer IPC using _alacritty msg_
 
@@ -58,16 +69,15 @@ All key-value pairs in the *env* section will be added as environment variables
 for any process spawned by Alacritty, including its shell. Some entries may
 override variables set by alacritty itself.
 
-```
-[env]
-TERM = "alacritty"
-```
+Example:
+	*[env]*++
+WINIT_X11_SCALE_FACTOR = _"1.0"_
 
 # WINDOW
 
 This section documents the *[window]* table of the configuration file.
 
-*dimensions* { columns = <integer>, lines = <integer> }
+*dimensions* = { columns = _<integer>_, lines = _<integer>_ }
 
 	Window dimensions (changes require restart)
 
@@ -76,22 +86,22 @@ This section documents the *[window]* table of the configuration file.
 	least _2_, while using a value of _0_ for columns and lines will fall back
 	to the window manager's recommended size
 
-	Default: _{ column = 0, lines = 0 }_
+	Default: { column = _0_, lines = _0_ }
 
-*padding* { x = <integer>, y = <integer> }
+*padding* = { x = _<integer>_, y = _<integer>_ }
 
 	Blank space added around the window in pixels. This padding is scaled
 	by DPI and the specified value is always added at both opposing sides.
 
-	Default: _{ x = 0, y = 0 }_
+	Default: { x = _0_, y = _0_ }
 
-*dynamic_padding* <boolean>
+*dynamic_padding* = _true_|_false_
 
 	Spread additional padding evenly around the terminal content.
 
 	Default: _false_
 
-*decorations* "Full" | "None" | "Transparent" | "Buttonless"
+*decorations* = _"Full"_|_"None"_|_"Transparent"_|_"Buttonless"_
 
 	Window decorations
 
@@ -106,32 +116,32 @@ This section documents the *[window]* table of the configuration file.
 
 	Default: _"Full"_
 
-*opacity* <float>
+*opacity* = _<float>_
 
 	Background opacity as a floating point number from _0.0_ to _1.0_. The value
 	\_0.0_ is completely transparent and _1.0_ is opaque.
 
 	Default: _1.0_
 
-*startup_mode* "Windowed" | "Maximized" | "Fullscreen" | "SimpleFullscreen"
+*startup_mode* = _"Windowed"_|_"Maximized"_|_"Fullscreen"_|_"SimpleFullscreen"_
 
 	Startup mode (changes require restart)
 
 	Default: _"Windowed"_
 
-*title* <string>
+*title* = _"<string>"_
 
 	Window title
 
 	Default: _"Alacritty"_
 
-*dynamic_title* <boolean>
+*dynamic_title* = _true_|_false_
 
 	Allow terminal applications to change Alacritty's window title.
 
 	Default: _true_
 
-*class* { instance = <string>, general = <string> } _(Linux/BSD only)_
+*class* = { instance = _"<string>"_, general = _"<string>"_ } # _(Linux/BSD only)_
 
 	Window class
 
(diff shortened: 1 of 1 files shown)


extra/man/alacritty.5.scd
```diff
@@ -7,7 +7,9 @@ Alacritty - TOML configuration file format
 # SYNTAX
 
 Alacritty's configuration file uses the TOML format. The format's specification
-can be found at https://toml.io/en/v1.0.0.
+can be found at _https://toml.io/en/v1.0.0_. Each _Default:_ and _Example:_
+entry contains valid TOML config which could be copied into the configuration
+files.
```
chrisduerr, line 12:
```suggestion
can be found at _https://toml.io/en/v1.0.0_. Every _Default:_ and _Example:_
entry is valid TOML and can be copied directly into the configuration file.
```

extra/man/alacritty.5.scd
```diff
@@ -33,6 +42,12 @@ This section documents the root level of the configuration file.
 		Linux/BSD/macOS: _$SHELL_ or the user's login shell, if _$SHELL_ is unset++
 Windows: _"powershell"_
 
+	*Example:*
+
+		*[shell]*++
+program = _"/bin/bash"_++
+args = _["-l"]_
+
 *working_directory* <string> | "None"
```
chrisduerr, line 51: Is there a reason why we don't document examples for simple K/V pairs? Wouldn't it be better to have an example than the default, so you can copy-paste it?

extra/man/alacritty.5.scd
```diff
@@ -24,6 +26,13 @@ This section documents the root level of the configuration file.
 	All imports must either be absolute paths starting with _/_, or paths
 	relative to the user's home directory starting with _~/_.
 
+	Example:
+
+		import = [++
+  _"~/.my_fancy_colorscheme.toml"_,++
```
chrisduerr, line 32: This tilde will be transformed into one that cannot be used in the shell.

extra/man/alacritty.5.scd
```diff
@@ -308,122 +302,145 @@ Colors are specified using their hexadecimal values with a _#_ prefix: _#RRGGBB_
...
+
+	*[colors.dim]*++
+black   = _"#0f0f0f"_++
+red     = _"#712b2b"_++
+green   = _"#5f6f3a"_++
+yellow  = _"#a17e4d"_++
+blue    = _"#456877"_++
+magenta = _"#704d68"_++
+cyan    = _"#4d7770"_++
+white   = _"#8e8e8e"_
```
chrisduerr, line 443: This seems excessive. Wouldn't it be better if each block was below its actual table?

extra/man/alacritty.5.scd
```diff
@@ -495,44 +512,45 @@ This section documents the *[cursor]* table of the configuration file.
...
+	*[cursor]*++
+blink_timeout     = _5_++
+blinking_interval = _750_++
+thickness         = _0.15_++
+unfocused_hollow  = _true_++
+vi_mode_style     = _"None"_
+
+	*[cursor.style]*++
+shape = _"Block"_++
+blinking = _"Off"_
```
chrisduerr, line 553: Again these are two tables in one block. I think separation would be better.

extra/man/alacritty.5.scd
```diff
@@ -633,14 +656,19 @@ _action_ or a _command_.
...
+alphabet = _"jfkdls;ahgurieowpq"_
+
+	*[[hints.enabled]]*++
+command         = _"xdg-open"_++
+regex           = _"(mailto:|gemini:|gopher:|https:|http:|news:|file:|git:|ssh:|ftp:)[^\\u0000-\\u001F\\u007F-\\u009F<>\\"\\\\s{-}\\\\^⟨⟩`]+"_++
+hyperlinks      = _true_++
+post_processing = _true_++
+persist         = _false_++
+mouse.enabled   = _true_++
+binding         = { key = _"U"_, mods = _"Control | Shift"_ }
```
chrisduerr, line 671: Also think splitting would be better here.

chrisduerr (changes requested):

extra/man/alacritty.5.scd
kchibisov: I wasn't sure how to document here because it's a global section.

extra/man/alacritty.5.scd
kchibisov: What do you mean, we support `~` and I've tested this import explicitly.

extra/man/alacritty.5.scd
kchibisov: I just wanted to provide a full color scheme here.

extra/man/alacritty.5.scd
chrisduerr: `working_directory = "whatever"`

extra/man/alacritty.5.scd
chrisduerr: When generating this with scdoc, I get the following:

```
"˜/.my_fancy_colorscheme.toml"
```

This does **not** work.

extra/man/alacritty.5.scd
chrisduerr: That's fair but it seems excessive.

extra/man/alacritty.5.scd
kchibisov: As discussed, it's an issue with `groff`.

kchibisov:
I would suggest to reread the entire thing just in case.

extra/man/alacritty.5.scd
```diff
@@ -24,6 +26,13 @@ This section documents the root level of the configuration file.
 	All imports must either be absolute paths starting with _/_, or paths
 	relative to the user's home directory starting with _~/_.
 
+	Example:
+
+		import = [++
+  _"~/.my_fancy_colorscheme.toml"_,++
+  _"/home/user/.linux_extra.toml"_,++
```
chrisduerr, line 33: I think it would be best to encourage proper config file organization. So we should maybe use example directories that start with ~/.config/alacritty/` at least.

extra/man/alacritty.5.scd
```diff
@@ -52,16 +67,21 @@ Windows: _"powershell"_
 
 	Default: _true_
 
+Default:
+
+	live_config_reload = _true_++
+working_directory = _"None"_
```
chrisduerr, line 73: This is just duplicated? We shouldn't have things documented twice.

You can just document the defaults as toml.

extra/man/alacritty.5.scd
```diff
@@ -52,16 +67,21 @@ Windows: _"powershell"_
...
 override variables set by alacritty itself.
 
-```
-[env]
-TERM = "alacritty"
-```
+Example:
+
+	*[env]*++
+TERM = _"alacritty"_
```
chrisduerr, line 84: Overriding `TERM` is usually a bad idea. I wonder if we could use a better example variable here? Maybe something from winit?

extra/man/alacritty.5.scd
```diff
@@ -144,19 +148,26 @@ This section documents the *[window]* table of the configuration file.
...
+	*[window]*++
+dimensions      = { columns = _0_, lines = _0_ }++
+padding         = { x = _0_, y = _0_ }++
+dynamic_padding = _false_++
+decorations     = _"Full"_++
+opacity         = _1.0_++
+startup_mode    = _"Windowed"_++
+title           = _"Alacritty"_++
+dynamic_title   = _true_++
+decorations_theme_variant = _"None"_
```
chrisduerr, line 170: Not a fan of having all the defaults documented in a massive blog so far from its definition.

I don't want to scroll through half a manpage just to get the documentation for that one field I'm looking for. Especially because I don't even know if that documentation is still going to come. If I'm just looking at that one field, how am I supposed to know there's more?

This obviously applies to all the toml blocks.

extra/man/alacritty.5.scd
```diff
@@ -168,59 +179,52 @@ This section documents the *[scrolling]* table of the configuration file.
...
 
 *bold* { family = <string>, style = <string> }
 
-	If the family is not specified, it will fall back to the value specified for
-	the normal font.
-
-	Default: _{ style = "Bold" }_
+	If the _family_ is not specified, it will fall back to the value specified
+	for the _normal font_. If the _style_ is not specified, it will fallback
+	to the _default weight and slant_ for this style.
```
chrisduerr, line 205: Might be nice to add a `Defaut:` "heading" here so people know that this is only relevant if they're pondering defaults?

chrisduerr (changes requested):

extra/man/alacritty.5.scd
kchibisov: But the whole point was to make it like that so you can copy sections into the config and probably assemble the config out of man page pretty reliably?  I don't see any issue with sections like that.

extra/man/alacritty.5.scd
kchibisov: Hm, I think I just forgot to remove `default`

extra/man/alacritty.5.scd
kchibisov: That's a good idea.

extra/man/alacritty.5.scd
kchibisov: I don't quite understand, they could clearly rely on this behavior when omitting a specific field.

extra/man/alacritty.5.scd
chrisduerr: I **really** don't like that which is why I didn't write the manpage like this to begin with.

I think it's a bloody mess and makes the entire manpage unreadable.

extra/man/alacritty.5.scd
chrisduerr: It's a default though, so if you change it to *anything* the entire paragraph is useless.

But this is not explicitly pointed out ahead of time, basically the entire paragraph needs to be read to understand that.

kchibisov:
I've completely changed the way it was written/handled.

chrisduerr (changes requested):
Maybe it will be easier to split this into multiple PRs? Leaving out the more controversial stuff could at least get the fixes merged quicker.

kchibisov:
The non-controversial parts of the PR were merged separately.

It was also discussed on IRC that doing color codding and more explicit enums were fine, so it may be submitted separately later.

PR #7109, opened 2023-07-22
Add `terminal` config section to control OSCs

### LINKED ISSUE #3386: Add config option to restrict use of certain OSCs
Some `OSC`s could be not desirable for some users. As an example `OSC 7` or `OSC 50` are a bit strange, but one could make use of them. Also some folks could decide to disable `OSC 52`. The motivation for banning some of them is security (`OSC 52`) or annoyance (`OSC 7`).

kchibisov:
Some environments demand certain OSC sequences to be disabled or some escape sequence could require handling which is out of scope of alacritty, but could be done by external script (OSC 777).

Added section for now just handles the `OSC 52` sequence and changes its default to be `OnlyCopy`, which is handy for remote copy, but `Paste` is redundant because normal `Paste` hotkey could be used as well.

Fixes #3386.

### DIFF
--- a/alacritty_terminal/src/config/mod.rs
+++ b/alacritty_terminal/src/config/mod.rs
@@ -31,8 +31,33 @@ pub struct Config {
     /// Cursor configuration.
     pub cursor: Cursor,
 
+    /// Terminal specific settings.
+    pub terminal: Terminal,
+
     #[config(flatten)]
     pub pty_config: PtyConfig,
+}
+
+#[derive(ConfigDeserialize, Clone, Debug, PartialEq, Eq, Default)]
+pub struct Terminal {
+    // OSC 52 handling (clipboard handling).
+    pub osc52: Osc52,
+}
+
+#[derive(ConfigDeserialize, Clone, Debug, PartialEq, Eq, Default)]
+pub enum Osc52 {
+    /// The handling of the escape sequence is disabled.
+    Disabled,
+    /// Only copy sequence is accepted.
+    ///
+    /// This option is the default as a compromiss between entirely
+    /// disabling it (the most secure) and allowing `paste` (the less secure).
+    #[default]
+    OnlyCopy,
+    /// Only paste sequence is accepted.
+    OnlyPaste,
+    /// Both are accepted.
+    CopyPaste,
 }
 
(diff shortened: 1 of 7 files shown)


alacritty_terminal/src/config/mod.rs
```diff
@@ -31,10 +31,35 @@ pub struct Config {
...
+    ///
+    /// This option is the default as a compromiss between entirely
+    /// disabling it (the most secure) and allowing `paste` (the less secure).
+    #[default]
+    OnlyCopy,
+    /// Only paste sequence is accepted.
+    OnlyPaste,
+    /// Both are accepted.
+    CopyPaste,
+}
```
chrisduerr, line 61: Arguably bitflags would be easier here inside of Alacritty, but I suppose the main reason for an enum was to allow easy (de)serialization?

extra/man/alacritty.5.scd
```diff
@@ -534,6 +534,20 @@ This section documents the *[cursor]* table of the configuration file.
...
+
+This section documents the *[terminal]* table of the configuration file.
+
+*osc52* "Disabled" | "OnlyCopy" | "OnlyPaste" | "CopyPaste"
+
+	Controls the ability to write to the system clipboard with the _OSC 52_
+	escape sequence. While this escape sequence is useful to copy contents
+	from the remote server, the ability to paste is not that necessary
+	given that normal `Paste` action could be used instead and such
+	sequence could be abused.
```
chrisduerr, line 547:
```suggestion
	Controls the ability to write to the system clipboard with the _OSC 52_
	escape sequence. While this escape sequence is useful to copy contents
	from the remote server, allowing any application to read from the clipboard can be easily abused while not providing significant benefits over explicitly pasting text.
```

My main gripe with this comment was the "not that necessary" part sounding kinda imprecise. I think the new suggestion should be a little more direct.

chrisduerr (changes requested):

alacritty_terminal/src/config/mod.rs
kchibisov: Yeah, and we don't have that many variants.

chrisduerr (approved):

PR #7125, opened 2023-08-02
Implement kitty's keyboard protocol

### LINKED ISSUE #6378: Add support for the kitty keyboard protocol
I'm developing a terminal application using [crossterm](https://github.com/crossterm-rs/crossterm), and I would like to use the number keys as modifiers instead of ctrl and alt. That can be done through the use of [`PushKeyboardEnhancementFlags`](https://docs.rs/crossterm/latest/crossterm/event/struct.PushKeyboardEnhancementFlags.html), whereupon my program will receive [`Release`](https://docs.rs/crossterm/latest/crossterm/event/enum.KeyEventKind.html#variant.Release) and [`Repeat`](https://docs.rs/crossterm/latest/crossterm/event/enum.KeyEventKind.html#variant.Repeat) events in addition to just [`Press`](https://docs.rs/crossterm/latest/crossterm/event/enum.KeyEventKind.html#variant.Press). However, it seems like that functionality is dependent on the terminal implementing the [kitty keyboard protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/), which alacritty doesn't.

Are there any plans for implementing such a feature, or is there a better way to achieve the same goal (arbitrary keys as modifiers) in alacritty?

kchibisov:
The protocol enables robust key reporting for the applications, so they could bind more keys and the user won't have collisions with the legacy control keys.

Fixes #6378.
Links: https://sw.kovidgoyal.net/kitty/keyboard-protocol

--

Only draft because of vte. I've also hasn't updated man pages, because the modes could be renamed. Besides everything should work...

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -7,7 +7,9 @@ use serde::de::{self, Error as SerdeError, MapAccess, Unexpected, Visitor};
 use serde::{Deserialize, Deserializer};
 use toml::Value as SerdeValue;
 use winit::event::MouseButton;
-use winit::keyboard::{Key, KeyCode, KeyLocation, ModifiersState, NamedKey, PhysicalKey};
+use winit::keyboard::{
+    Key, KeyCode, KeyLocation as WinitKeyLocation, ModifiersState, NamedKey, PhysicalKey,
+};
 use winit::platform::scancode::PhysicalKeyExtScancode;
 
 use alacritty_config_derive::{ConfigDeserialize, SerdeReplace};
@@ -413,10 +415,13 @@ macro_rules! trigger {
         BindingKey::Keycode { key: Key::Character($key.into()), location: $location }
     }};
     (KeyBinding, $key:literal,) => {{
-        BindingKey::Keycode { key: Key::Character($key.into()), location: KeyLocation::Standard }
+        BindingKey::Keycode { key: Key::Character($key.into()), location: KeyLocation::Any }
+    }};
+    (KeyBinding, $key:ident, $location:expr) => {{
+        BindingKey::Keycode { key: Key::Named(NamedKey::$key), location: $location }
     }};
     (KeyBinding, $key:ident,) => {{
-        BindingKey::Keycode { key: Key::Named(NamedKey::$key), location: KeyLocation::Standard }
+        BindingKey::Keycode { key: Key::Named(NamedKey::$key), location: KeyLocation::Any }
     }};
     (MouseBinding, $base:ident::$button:ident,) => {{
         $base::$button
@@ -432,6 +437,8 @@ pub fn default_mouse_bindings() -> Vec<MouseBinding> {
     )
 }
 
+// NOTE: key sequences which are not present here, like F5-F20, PageUp/PageDown codes are
+// built on the fly in input/keyboard.rs.
 pub fn default_key_bindings() -> Vec<KeyBinding> {
     let mut bindings = bindings!(
         KeyBinding;
@@ -439,56 +446,28 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
         Copy,  +BindingMode::VI; Action::ClearSelection;
         Paste, ~BindingMode::VI; Action::Paste;
         "l",       ModifiersState::CONTROL; Action::ClearLogNotice;
-        "l",       ModifiersState::CONTROL, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x0c".into());
-        Tab,       ModifiersState::SHIFT,   ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[Z".into());
-        Backspace, ModifiersState::ALT,     ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b\x7f".into());
-        Backspace, ModifiersState::SHIFT,   ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x7f".into());
+        "l",       ModifiersState::CONTROL; Action::ReceiveChar;
         Home,      ModifiersState::SHIFT, ~BindingMode::ALT_SCREEN; Action::ScrollToTop;
         End,       ModifiersState::SHIFT, ~BindingMode::ALT_SCREEN; Action::ScrollToBottom;
         PageUp,    ModifiersState::SHIFT, ~BindingMode::ALT_SCREEN; Action::ScrollPageUp;
         PageDown,  ModifiersState::SHIFT, ~BindingMode::ALT_SCREEN; Action::ScrollPageDown;
-        Home,      ModifiersState::SHIFT, +BindingMode::ALT_SCREEN, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[1;2H".into());
-        End,       ModifiersState::SHIFT, +BindingMode::ALT_SCREEN, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[1;2F".into());
-        PageUp,    ModifiersState::SHIFT, +BindingMode::ALT_SCREEN, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[5;2~".into());
-        PageDown,  ModifiersState::SHIFT, +BindingMode::ALT_SCREEN, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[6;2~".into());
+        // App cursor mode.
         Home,       +BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1bOH".into());
-        Home,       ~BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[H".into());
         End,        +BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1bOF".into());
-        End,        ~BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[F".into());
         ArrowUp,    +BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1bOA".into());
-        ArrowUp,    ~BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[A".into());
         ArrowDown,  +BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1bOB".into());
-        ArrowDown,  ~BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[B".into());
         ArrowRight, +BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1bOC".into());
-        ArrowRight, ~BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[C".into());
         ArrowLeft,  +BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1bOD".into());
-        ArrowLeft,  ~BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[D".into());
-        Backspace,  ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x7f".into());
-        Insert,     ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[2~".into());
-        Delete,     ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[3~".into());
-        PageUp,     ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[5~".into());
-        PageDown,   ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[6~".into());
-        F1,         ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1bOP".into());
-        F2,         ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1bOQ".into());
-        F3,         ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1bOR".into());
-        F4,         ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1bOS".into());
-        F5,         ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[15~".into());
-        F6,         ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[17~".into());
-        F7,         ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[18~".into());
-        F8,         ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[19~".into());
-        F9,         ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[20~".into());
-        F10,        ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[21~".into());
-        F11,        ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[23~".into());
-        F12,        ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[24~".into());
-        F13,        ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[25~".into());
-        F14,        ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[26~".into());
-        F15,        ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[28~".into());
-        F16,        ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[29~".into());
-        F17,        ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[31~".into());
-        F18,        ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[32~".into());
-        F19,        ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[33~".into());
-        F20,        ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1b[34~".into());
-
+        // Legacy keys handling which can't be automatically encoded.
+        F1,         ~BindingMode::VI, ~BindingMode::SEARCH, ~BindingMode::REPORT_ALL_KEYS_AS_ESC, ~BindingMode::DISAMBIGUATE_ESC_CODES; Action::Esc("\x1bOP".into());
+        F2,         ~BindingMode::VI, ~BindingMode::SEARCH, ~BindingMode::REPORT_ALL_KEYS_AS_ESC, ~BindingMode::DISAMBIGUATE_ESC_CODES; Action::Esc("\x1bOQ".into());
+        F3,         ~BindingMode::VI, ~BindingMode::SEARCH, ~BindingMode::REPORT_ALL_KEYS_AS_ESC, ~BindingMode::DISAMBIGUATE_ESC_CODES; Action::Esc("\x1bOR".into());
+        F4,         ~BindingMode::VI, ~BindingMode::SEARCH, ~BindingMode::REPORT_ALL_KEYS_AS_ESC, ~BindingMode::DISAMBIGUATE_ESC_CODES; Action::Esc("\x1bOS".into());
+        Tab,       ModifiersState::SHIFT,   ~BindingMode::VI,   ~BindingMode::SEARCH, ~BindingMode::REPORT_ALL_KEYS_AS_ESC; Action::Esc("\x1b[Z".into());
+        Tab,       ModifiersState::SHIFT | ModifiersState::ALT, ~BindingMode::VI, ~BindingMode::SEARCH, ~BindingMode::REPORT_ALL_KEYS_AS_ESC; Action::Esc("\x1b\x1b[Z".into());
+        Backspace, ~BindingMode::VI, ~BindingMode::SEARCH, ~BindingMode::REPORT_ALL_KEYS_AS_ESC; Action::Esc("\x7f".into());
+        Backspace, ModifiersState::ALT,     ~BindingMode::VI, ~BindingMode::SEARCH, ~BindingMode::REPORT_ALL_KEYS_AS_ESC; Action::Esc("\x1b\x7f".into());
+        Backspace, ModifiersState::SHIFT,   ~BindingMode::VI, ~BindingMode::SEARCH, ~BindingMode::REPORT_ALL_KEYS_AS_ESC; Action::Esc("\x7f".into());
         // Vi mode.
         Space, ModifiersState::SHIFT | ModifiersState::CONTROL, ~BindingMode::SEARCH; Action::ToggleViMode;
         Space, ModifiersState::SHIFT | ModifiersState::CONTROL, +BindingMode::VI, ~BindingMode::SEARCH; Action::ScrollToBottom;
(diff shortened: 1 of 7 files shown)


Nyabinary:
Any status on this?

kchibisov:
it needs a release of the vte and some minor changes.

Nyabinary:
Any ETA on when the new vte release will come?

kchibisov:
Please, don't spam the PR. Everything will happen eventually.

alacritty/src/input.rs
```diff
@@ -1192,6 +1243,251 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+        Key::Character("8") if csi_u_numpad => ("57407".into(), 'u'),
+        Key::Character("9") if csi_u_numpad => ("57408".into(), 'u'),
+        Key::Character(".") if csi_u_numpad => ("57409".into(), 'u'),
+        Key::Character("/") if csi_u_numpad => ("57410".into(), 'u'),
+        Key::Character("*") if csi_u_numpad => ("57411".into(), 'u'),
+        Key::Character("-") if csi_u_numpad => ("57412".into(), 'u'),
+        Key::Character("+") if csi_u_numpad => ("57413".into(), 'u'),
+        Key::Enter if csi_u_numpad => ("57414".into(), 'u'),
+        Key::Character("=") if csi_u_numpad => ("57415".into(), 'u'),
+        // KP_SEPARATOR if csi_u_numpad => ("57415".into(), 'u'),
```
chrisduerr, line 1310: Why are there randomly commented out values?!

alacritty/src/input.rs
```diff
@@ -1192,6 +1243,251 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+        Key::ArrowDown if csi_u_numpad => ("57420".into(), 'u'),
+        Key::PageUp if csi_u_numpad => ("57421".into(), 'u'),
+        Key::PageDown if csi_u_numpad => ("57422".into(), 'u'),
+        Key::Home if csi_u_numpad => ("57423".into(), 'u'),
+        Key::End if csi_u_numpad => ("57424".into(), 'u'),
+        Key::Insert if csi_u_numpad => ("57425".into(), 'u'),
+        Key::Delete if csi_u_numpad => ("57426".into(), 'u'),
+        // KP_BEGIN if csi_u_numpad => ("57427".into(), 'u'),
+        // Handle common keys.
+        Key::ArrowLeft if mods.is_empty() && !send_event_type => ("".into(), 'D'),
```
chrisduerr, line 1323: This is a giant mess and I cannot imagine Rust manages to optimize all these conditionals well, especially when function calls are involved. Have you checked what kind of code this generates?

alacritty/src/input.rs
```diff
@@ -1192,6 +1243,251 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+            let payload = if mode.contains(TermMode::KEYBOARD_REPORT_ALTERNATE_KEYS)
+                && codepoint != base_codepoint
+            {
+                format!("{codepoint}:{base_codepoint}")
+            } else {
+                codepoint.to_string()
+            };
+
+            (payload.into(), 'u')
+        },
```
chrisduerr, line 1438: Having such a big code block in such an already complex match statement is unacceptable.

alacritty_terminal/src/config/mod.rs
```diff
@@ -40,8 +40,16 @@ pub struct Config {
...
+    /// OSC 52 handling (clipboard handling).
     pub osc52: Osc52,
+
+    /// Handle kitty keyboard protocol.
+    ///
+    /// Set to `true` only when all of the kitty keyboard protocol modes are supported.
+    ///
+    /// The default is `false`.
+    #[config(skip)]
+    pub kitty_keyboard_protocol: bool,
```
chrisduerr, line 52: This is not the place to store random terminal options.

chrisduerr (changes requested):

alacritty/src/config/bindings.rs
```diff
@@ -793,6 +700,8 @@ bitflags! {
         const ALT_SCREEN          = 0b0000_0100;
         const VI                  = 0b0000_1000;
         const SEARCH              = 0b0001_0000;
+        const DISAMBIGUATE_KEYS   = 0b0010_0000;
```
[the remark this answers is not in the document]
kchibisov: from the spec. it's also not user exposed.

alacritty/src/input.rs
```diff
@@ -1002,7 +1002,33 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+
+        if key.state == ElementState::Released {
+            if mode.contains(TermMode::KEYBOARD_REPORT_EVENT_TYPES)
+                && !mode.contains(TermMode::VI)
+                && !self.ctx.search_active()
+                && !self.ctx.display().hint_state.active()
+            {
+                // NOTE: echoing the key back on release is how it's done in kitty/foot and
+                // it's how it should be done according to the kitty author
+                // https://github.com/kovidgoyal/kitty/issues/6516#issuecomment-1659454350
```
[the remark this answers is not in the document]
kchibisov: I don't remember the question, but I reworded the comment.

alacritty/src/input.rs
```diff
@@ -1002,7 +1002,33 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+                let bytes: Cow<'static, [u8]> = match key.logical_key.as_ref() {
+                    Key::Tab => [b'\t'].as_slice().into(),
+                    Key::Enter => [b'\r'].as_slice().into(),
+                    Key::Delete => [b'\x7f'].as_slice().into(),
+                    Key::Escape => [b'\x1b'].as_slice().into(),
```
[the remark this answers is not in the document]
kchibisov: I need to do it manually for `Backspace`(because winit has different code) and I don't want to deal with `Option` from invoking `to_text`, even though I can unwrap it.

alacritty/src/config/bindings.rs
```diff
@@ -714,12 +626,48 @@ pub fn platform_key_bindings() -> Vec<KeyBinding> {
...
+            (self, other),
+            (_, KeyLocation::Any)
+                | (KeyLocation::Any, _)
+                | (KeyLocation::Standard, KeyLocation::Standard)
+                | (KeyLocation::Numpad, KeyLocation::Numpad)
+        )
+    }
+}
+
+impl Eq for KeyLocation {}
```
chrisduerr, line 669: You can just add that to the derive, no?

alacritty/src/config/bindings.rs
```diff
@@ -731,23 +679,26 @@ impl<'a> Deserialize<'a> for BindingKey {
...
+                        "Right" => (Key::Named(ArrowRight), KeyLocation::Any),
+                        "At" => (Key::Character("@".into()), KeyLocation::Any),
+                        "Colon" => (Key::Character(":".into()), KeyLocation::Any),
+                        "Period" => (Key::Character(".".into()), KeyLocation::Any),
+                        "LBracket" => (Key::Character("[".into()), KeyLocation::Any),
+                        "RBracket" => (Key::Character("]".into()), KeyLocation::Any),
+                        "Semicolon" => (Key::Character(";".into()), KeyLocation::Any),
+                        "Backslash" => (Key::Character("\\".into()), KeyLocation::Any),
+
+                        // The keys which has alternative on numeric pad.
```
chrisduerr, line 699:
```suggestion
                        // The keys which have alternatives on the numeric pad.
```

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,592 @@
...
+    ) -> bool {
+        if mode.contains(TermMode::REPORT_ALL_KEYS_AS_ESC) {
+            true
+        } else if mode.contains(TermMode::DISAMBIGUATE_ESC_CODES) {
+            let on_numpad = key.location == KeyLocation::Numpad;
+            let is_escape = key.logical_key == Key::Named(NamedKey::Escape);
+            is_escape || (!mods.is_empty() && mods != ModifiersState::SHIFT) || on_numpad
+        } else {
+            // `Delete` key has always text attached to it, but it's a named key, thus needs
+            // to be excluded here as well.
```
chrisduerr, line 117:
```suggestion
            // `Delete` key always has text attached to it, but it's a named key, thus needs
            // to be excluded here as well.
```

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,592 @@
...
+            }
+        }
+    }
+}
+
+/// Build the keyboard escape sequence based on the given `key`, `mods`, and `mode`.
+///
+/// The key sequences for `APP_KEYPAD` and alike are handled inside the bindings.
+#[inline(never)]
+fn build_keyboard_esc_sequence(key: KeyEvent, mods: ModifiersState, mode: TermMode) -> Vec<u8> {
```
chrisduerr, line 222: `should_build_key_sequence` and `build_keyboard_esc_sequence` are inconsistent.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,592 @@
...
+            let timer_id = TimerId::new(Topic::DelayedSearch, self.ctx.window().id());
+            let scheduler = self.ctx.scheduler_mut();
+            if let Some(timer) = scheduler.unschedule(timer_id) {
+                scheduler.schedule(timer.event, TYPING_SEARCH_DELAY, false, timer.id);
+            }
+        }
+    }
+}
+
+/// Build the keyboard escape sequence based on the given `key`, `mods`, and `mode`.
```
chrisduerr, line 218:
```suggestion
/// Build a key's escape sequence based on the given `key`, `mods`, and `mode`.
```

It's not an escape sequence for the keyboard, but for a key.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,592 @@
...
+
+impl KeyEscSequenceBuildingContext {
+    /// Prepare key with the text attached to it.
+    fn prepare_text_base(&self, key: &KeyEvent) -> Option<KeyEscSequenceBase> {
+        match key.logical_key.as_ref() {
+            Key::Character(character) => {
+                // We could have compose sequience, which has more than one character. Thus
+                // ignore it when we have to encode all.
+                if self.kitty_encode_all && character.chars().count() > 1 {
+                    Some(KeyEscSequenceBase::new("0".into(), 'u'))
```
chrisduerr, line 316: Why are we passing 0 as payload here? Why is the character just 'u'?

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,592 @@
...
+                } else {
+                    let character = character.chars().next().unwrap();
+                    let base_character = character.to_lowercase().next().unwrap();
+
+                    let codepoint = u32::from(character);
+                    let base_codepoint = u32::from(base_character);
+
+                    // NOTE Base layouts are ignored, since winit doesn't expose this information
```
chrisduerr, line 324:
```suggestion
                    // NOTE: Base layouts are ignored, since winit doesn't expose this information
```

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,592 @@
...
+            NamedKey::ArrowUp => (if omit_base { "" } else { "1" }, 'A'),
+            NamedKey::ArrowDown => (if omit_base { "" } else { "1" }, 'B'),
+            NamedKey::Home => (if omit_base { "" } else { "1" }, 'H'),
+            NamedKey::End => (if omit_base { "" } else { "1" }, 'F'),
+            NamedKey::PageUp => ("5", '~'),
+            NamedKey::PageDown => ("6", '~'),
+            NamedKey::Insert => ("2", '~'),
+            NamedKey::Delete => ("3", '~'),
+            NamedKey::F1 => (if self.kitty_seq && omit_base { "" } else { "1" }, 'P'),
+            NamedKey::F2 => (if self.kitty_seq && omit_base { "" } else { "1" }, 'Q'),
```
chrisduerr, line 416: This is an incomprehensible mess.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,592 @@
...
+                if self.kitty_encode_all {
+                    let left = key.location == KeyLocation::Left;
+                    base = match named {
+                        NamedKey::CapsLock => "57358",
+                        NamedKey::NumLock => "57360",
+                        NamedKey::Shift => {
+                            if left {
+                                "57441"
+                            } else {
+                                "57447"
```
chrisduerr, line 496: Way too much nesting here.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,592 @@
...
+bitflags::bitflags! {
+    /// The modifiers encoding for escape sequnce
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    struct EscSequenceModifiers : u8 {
+        const SHIFT   = 0b0000_0001;
+        const ALT     = 0b0000_0010;
+        const CONTROL = 0b0000_0100;
+        const SUPER   = 0b0000_1000;
+        // NOTE kitty protocol defines additional modiffiers to what is present here, like
+        // Capslock, but it's not a modifier as per winit.
```
chrisduerr, line 572:
```suggestion
        // NOTE: Kitty protocol defines additional modifiers to what is present here, like
        // Capslock, but it's not a modifier as per winit.
```

chrisduerr (changes requested):

alacritty/src/input/keyboard.rs
kchibisov: Hm, I read it wrongly initially. Basically we should send `0` only when assoc text is requested and when report_all is also set.

See https://sw.kovidgoyal.net/kitty/keyboard-protocol/#text-as-code-points .

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,600 @@
...
+                (base, 'Q')
+            },
+            NamedKey::F3 => {
+                // F3 diverges from alacritty's terminfo for kitty protocol.
+                if self.kitty_seq {
+                    ("13", '~')
+                } else {
+                    ("1", 'R')
+                }
+            },
```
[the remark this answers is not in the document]
kchibisov: There's nothing to `omit`, and yes it's different because we have special key defined in terminfo.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,600 @@
...
+            (NamedKey::Alt, KeyLocation::Left) => "57443",
+            (NamedKey::Super, KeyLocation::Left) => "57444",
+            (NamedKey::Hyper, KeyLocation::Left) => "57445",
+            (NamedKey::Meta, KeyLocation::Left) => "57446",
+            (NamedKey::Shift, _) => "57447",
+            (NamedKey::Control, _) => "57448",
+            (NamedKey::Alt, _) => "57449",
+            (NamedKey::Super, _) => "57450",
+            (NamedKey::Hyper, _) => "57451",
+            (NamedKey::Meta, _) => "57452",
```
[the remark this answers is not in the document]
kchibisov: I can, but then it won't be `&'static str`

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,600 @@
...
+impl KeyEscSequenceBuildingContext {
+    /// Prepare key with the text attached to it.
+    fn prepare_text_base(&self, key: &KeyEvent) -> Option<KeyEscSequenceBase> {
+        let character = match key.logical_key.as_ref() {
+            Key::Character(character) => character,
+            _ => return None,
+        };
+
+        // Character from winit is non-empty.
+        if character.chars().count() == 1 {
```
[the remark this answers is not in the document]
kchibisov: It was more of a reminder that `count == 0` impossible to happen.

alacritty_terminal/src/term/mod.rs
```diff
@@ -1193,6 +1268,63 @@ impl<T: EventListener> Handler for Term<T> {
...
+        self.set_keyboard_mode(mode.into(), KeyboardModesApplyBehavior::Replace);
+    }
+
+    #[inline]
+    fn pop_keyboard_modes(&mut self, to_pop: u16) {
+        if !self.config.kitty_keyboard {
+            return;
+        }
+
+        trace!("Attemting to pop {to_pop} keyboard mode from the stack");
```
chrisduerr, line 1310:
```suggestion
        trace!("Attemting to pop {to_pop} keyboard modes from the stack");
```

Should be either plural or (s), otherwise someone's going to think it's a specific mode.

alacritty_terminal/src/term/mod.rs
```diff
@@ -668,6 +724,12 @@ impl<T> Term<T> {
             self.inactive_grid.reset_region(..);
         }
 
+        mem::swap(&mut self.keyboard_mode_stack, &mut self.inactive_keyboard_mode_stack);
+        self.set_keyboard_mode(
+            self.keyboard_mode_stack.last().copied().unwrap_or(KeyboardModes::NO_MODE).into(),
```
chrisduerr, line 729: If you move this to a different line it might format nicer.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,589 @@
...
+impl KeyEscSequenceBase {
+    fn new(payload: Cow<'static, str>, terminator: KeyEscSequenceTerminator) -> Self {
+        Self { payload, terminator }
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum KeyEscSequenceTerminator {
+    /// The terminator is based on terminfo.
+    Tinfo(char),
```
chrisduerr, line 546: Bad name, you're saving like 3 characters.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,589 @@
...
+use alacritty_terminal::event::EventListener;
+use alacritty_terminal::term::TermMode;
+use winit::platform::modifier_supplement::KeyEventExtModifierSupplement;
+
+use crate::config::{Action, BindingKey, BindingMode};
+use crate::event::TYPING_SEARCH_DELAY;
+use crate::input::{ActionContext, Execute, Processor};
+use crate::scheduler::{TimerId, Topic};
+
+use KeyEscSequenceTerminator::{Kitty, Tinfo};
```
chrisduerr, line 20: Importing enum variants should **never** be a thing.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,589 @@
...
+        suppress_chars.unwrap_or(false)
+    }
+
+    /// Handle key release.
+    fn key_release(&mut self, key: KeyEvent, mode: TermMode, mods: ModifiersState) {
+        if mode.contains(TermMode::REPORT_EVENT_TYPES)
+            && !mode.contains(TermMode::VI)
+            && !self.ctx.search_active()
+            && !self.ctx.display().hint_state.active()
+        {
```
chrisduerr, line 187: Early return.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,589 @@
...
+            let bytes: Cow<'static, [u8]> = match key.logical_key.as_ref() {
+                // NOTE echo the key back on release to follow kitty/foot behavior. When
```
chrisduerr, line 189:
```suggestion
                // NOTE: Echo the key back on release to follow kitty/foot behavior. When
```

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,589 @@
...
+    let modifiers = mods.into();
+
+    let kitty_seq = mode.intersects(
+        TermMode::REPORT_ALL_KEYS_AS_ESC
+            | TermMode::DISAMBIGUATE_ESC_CODES
+            | TermMode::REPORT_EVENT_TYPES,
+    );
+
+    let kitty_encode_all = mode.contains(TermMode::REPORT_ALL_KEYS_AS_ESC);
+    // The `Press` is ignored here, since it's a default and no need to report it.
```
chrisduerr, line 234: What is this comment supposed to mean?

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,589 @@
...
+            NamedKey::AudioVolumeUp => ("57439", Kitty),
+            NamedKey::AudioVolumeMute => ("57440", Kitty),
+            _ => return None,
+        };
+
+        Some(KeyEscSequenceBase::new(base.into(), terminator))
+    }
+
+    /// Prepare control characters (e.g. Enter) and modifiers.
+    fn prepare_control_and_mods(&self, key: &KeyEvent) -> Option<KeyEscSequenceBase> {
```
chrisduerr, line 481: What does `prepare` mean in this context?

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
```diff
@@ -959,6 +1020,19 @@ impl<T> Term<T> {
...
     }
+
+    #[inline]
+    fn set_keyboard_mode(&mut self, mode: TermMode, apply: KeyboardModesApplyBehavior) {
+        let active_mode = self.mode & TermMode::KITTY_KEYBOARD_PROTOCOL;
+        self.mode &= !TermMode::KITTY_KEYBOARD_PROTOCOL;
+        let new_mode = match apply {
+            KeyboardModesApplyBehavior::Replace => mode,
+            KeyboardModesApplyBehavior::Union => active_mode.union(mode),
+            KeyboardModesApplyBehavior::Difference => active_mode.difference(mode),
```
chrisduerr, line 1031: Naming here is bad for both enum and variants.

alacritty_terminal/src/term/mod.rs
```diff
@@ -43,33 +44,70 @@ const TITLE_STACK_MAX_DEPTH: usize = 4096;
...
+        );
+        mode.set(
+            TermMode::REPORT_ALL_KEYS_AS_ESC,
+            value.contains(KeyboardModes::REPORT_ALL_KEYS_AS_ESC),
+        );
+        mode.set(
+            TermMode::REPORT_ASSOCIATED_TEXT,
+            value.contains(KeyboardModes::REPORT_ASSOCIATED_TEXT),
+        );
+        mode
```
chrisduerr, line 110: Formatting needs fixing. This can easily be done in two lines for each set call.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,584 @@
...
+        // NOTE: Kitty protocol defines additional modifiers to what is present here, like
+        // Capslock, but it's not a modifier as per winit.
+    }
+}
+
+impl SequenceModifiers {
+    /// Get the value which should be passed to escape sequence.
+    pub fn esc_sequence_encoded(self) -> u8 {
+        self.bits() + 1
+    }
```
chrisduerr, line 572: Should be imperative same for SequenceTerminator.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,584 @@
...
+            NamedKey::Enter => "13",
+            NamedKey::Escape => "27",
+            NamedKey::Space => "32",
+            NamedKey::Backspace => "127",
+            _ => "",
+        };
+
+        if !self.kitty_encode_all && base.is_empty() {
+            return None;
+        }
```
chrisduerr, line 497: Needs a comment for what this branch is for.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,584 @@
...
+            NamedKey::Home => (one_based, SequenceTerminator::Normal('H')),
+            NamedKey::End => (one_based, SequenceTerminator::Normal('F')),
+            NamedKey::ArrowLeft => (one_based, SequenceTerminator::Normal('D')),
+            NamedKey::ArrowRight => (one_based, SequenceTerminator::Normal('C')),
+            NamedKey::ArrowUp => (one_based, SequenceTerminator::Normal('A')),
+            NamedKey::ArrowDown => (one_based, SequenceTerminator::Normal('B')),
+            NamedKey::F1 => (one_based, SequenceTerminator::Normal('P')),
+            NamedKey::F2 => (one_based, SequenceTerminator::Normal('Q')),
+            NamedKey::F3 => {
+                // F3 in kitty protocol diverges from alacritty's terminfo.
```
chrisduerr, line 414: Why is this? Is Kitty's default reply different without the protocol enabled?

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,584 @@
...
+
+    /// Try building from [`NamedKey`].
+    fn try_from_named(&self, key: &KeyEvent) -> Option<SequenceBase> {
+        let named = match key.logical_key {
+            Key::Named(named) => named,
+            _ => return None,
+        };
+
+        // Omit payload when the base is 1, since it's a common thing to do and some clients break
+        // if don't do so.
```
chrisduerr, line 398: This is not a proper sentence.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,584 @@
...
+            // text.
+            Some(SequenceBase::new("0".into(), SequenceTerminator::Kitty))
+        } else {
+            None
+        }
+    }
+
+    /// Try building from numpad key.
+    ///
+    /// `None` is returned when the key neither known or numpad.
```
chrisduerr, line 347:
```suggestion
    /// `None` is returned when the key is neither known nor numpad.
```

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,584 @@
...
+    kitty_seq: bool,
+    /// Encode all the keys according to the protocol.
+    kitty_encode_all: bool,
+    /// Report event types.
+    kitty_event_type: bool,
+    modifiers: SequenceModifiers,
+}
+
+impl SequenceBuildingContext {
+    /// Try building sequence from the event emitting text.
```
chrisduerr, line 308:
```suggestion
    /// Try building sequence from the event's emitting text.
```

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,584 @@
...
+    if mode.contains(TermMode::REPORT_ASSOCIATED_TEXT)
+        && key.state != ElementState::Released
+        && (modifiers.is_empty() || modifiers == SequenceModifiers::SHIFT)
+    {
+        if let Some(text) = key.text {
+            let mut codepoints = text.chars().map(u32::from);
+            if let Some(codepoint) = codepoints.next() {
+                payload.push_str(&format!(";{codepoint}"));
+            }
+            // Push the rest of the chars.
```
chrisduerr, line 283: The rest of what to where? This comment doesn't provide any useful information.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,584 @@
...
+        payload.push(':');
+        let event_type = match key.state {
+            _ if key.repeat => '2',
+            ElementState::Pressed => '1',
+            ElementState::Released => '3',
+        };
+        payload.push(event_type);
+    }
+
+    // Associated text is not reported for the control/alt/logo key presses.
```
chrisduerr, line 273:
```suggestion
    // Associated text is not reported when the control/alt/logo key is pressed.
```

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,584 @@
...
+    let modifiers = mods.into();
+
+    let kitty_seq = mode.intersects(
+        TermMode::REPORT_ALL_KEYS_AS_ESC
+            | TermMode::DISAMBIGUATE_ESC_CODES
+            | TermMode::REPORT_EVENT_TYPES,
+    );
+
+    let kitty_encode_all = mode.contains(TermMode::REPORT_ALL_KEYS_AS_ESC);
+    // When the key event represents a `Press` it'll be encoded as `1`, so we can omit it.
```
chrisduerr, line 234: Default is usually zero though? Why can we omit it **because** it's a 1?

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
kchibisov: it's in vte.

alacritty/src/input/keyboard.rs
kchibisov: Kitty uses different values to us by default.

alacritty/src/input/keyboard.rs
kchibisov: Isn't it 1 actually? You can call a lot of escapes without argument (like moving) and it'll do 1. The same with these sequences, and impls in kitty,foot, etc. 1 is always omitted when it has nothing attached to it, like `1:3`.

j4james: Numeric parameters (as in `CUP`, `ICH`, `DL`) will often default to 1, but selective parameters (as in `SGR`, `ED`, and `TBC`) would typically default to 0. So if this were a standard sequence, you'd expect the key event options to be 0, 1, and 2, with a default of 0. But Kitty defined them as 1, 2, and 3, so the default really had to be 1. It's a bit unconventional, but that's just the way it is.

chrisduerr: Should rephrase this comment to explicitly state that the default is 1 and that's why we can omit it.

alacritty/src/input/keyboard.rs
```diff
@@ -0,0 +1,584 @@
...
+    /// Encode all the keys according to the protocol.
+    kitty_encode_all: bool,
+    /// Report event types.
+    kitty_event_type: bool,
+    modifiers: SequenceModifiers,
+}
+
+impl SequenceBuildingContext {
+    /// Try building sequence from the event's emitting text.
+    fn try_from_textual(&self, key: &KeyEvent) -> Option<SequenceBase> {
```
chrisduerr, line 308: I've just noticed this, but why do all these `try_from` impls not return `Self`?

chrisduerr (changes requested):

alacritty/src/input/keyboard.rs
kchibisov: That's because we use `Context` to build common concepts, it's not really a builder. Alternatively we'd need to carry the value inside this builder, which is not good.

The design is simply to have common args passed across the functions. I had a version with `Self` but I wasn't sure about it, since it was less clear.

chrisduerr: It's just the name that doesn't make sense here, I'm not opposed to the mechanism.

A `try_from` method that takes `self` and doesn't return `Self` is just nonsense. Maybe just rename the struct to `SequenceBuilder` and make it `try_build_textual` etc?

kchibisov: >A try_from method that takes self and doesn't return Self is just nonsense. Maybe just rename the struct to SequenceBuilder and make it try_build_textual etc?

I guess, I applied the suggestion.

chrisduerr (approved):

kamlendras:
Good job!

jfly:
Thank you @kchibisov, I'm very excited to try this out!

PR #7133, opened 2023-08-06
Copy global IPC options for new windows

### LINKED ISSUE #7128: option to persist config runtime config on create-window
Hiya~

I use `alacritty msg config` to switch some stuff around at runtime dynamically, and (unless i'm missing something), `alacritty msg create-window` doesn't seem to persist those runtime config changes to newly spawned windows.

Is this intentional behavior, and if not, would you consider having an option that copied runtime config to a newly created window when spawned via create-window?

chrisduerr:
This patch stores all options set for the Window ID `-1` and automatically applies them to new windows after their creation.

This in theory makes it possible to have a fully dynamic "default config" without having to reapply it for every new window.

Closes #7128.


DirectXMan12:
this does what i wanted for my current usecase (i.e. lgtm :+1: from that regard)

kchibisov (approved):

PR #7143, opened 2023-08-13
Don't report alternate scroll while search/vi active

kchibisov:
Reporting them is a bit confusing, since it could result in unpredictable input, when the user will try to scroll with the mouse wheel.

The mouse mode reporting was left untouched, since it was intended to be like that, but it could be alterated later, if it'll be undesirable either

### DIFF
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -713,6 +713,8 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
             .mode()
             .contains(TermMode::ALT_SCREEN | TermMode::ALTERNATE_SCROLL)
             && !self.ctx.modifiers().state().shift_key()
+            && !self.ctx.terminal().mode().contains(TermMode::VI)
+            && !self.ctx.search_active()
         {
             self.ctx.mouse_mut().accumulated_scroll.x += new_scroll_x_px * multiplier;
             self.ctx.mouse_mut().accumulated_scroll.y += new_scroll_y_px * multiplier;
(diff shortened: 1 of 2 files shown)


alacritty/src/input.rs
```diff
@@ -713,6 +713,8 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
             .mode()
             .contains(TermMode::ALT_SCREEN | TermMode::ALTERNATE_SCROLL)
             && !self.ctx.modifiers().state().shift_key()
+            && !self.ctx.terminal().mode().contains(TermMode::VI)
```
chrisduerr, line 716: I understand search, but why not during vi mode?

chrisduerr (changes requested):

alacritty/src/input.rs
chrisduerr: Actually even during search I'm not sure it's necessary to disable this? Did you run into issues?

kchibisov: Yeah, I tried to scroll in my irc client during VI mode, and nearly pasted some arbitrary string, because it was sending arrow key.

chrisduerr: That's more about mouse mode than anything else though, right?

kchibisov: It was alternate scroll though, I'm still not sure why we keep it in Vi mode.

chrisduerr: Why wouldn't you? Only because vi mode is active, doesn't mean you don't want to scroll?

kchibisov: Well, yeah, but Vi mode itself won't scroll and the scrolling is not native, it's just we try to send `Arrow` keys. At this point could just sent arrow keys directly.

chrisduerr: I don't see why we'd change anything here. Why wouldn't we send arrow keys, that's what we send without vi mode too.

chrisduerr: You can scroll the history while vi mode is active, so I don't see why alternate scroll wouldn't work.

Either both should work or neither.

PR #7145, opened 2023-08-15
Fix scaling for offset and glyph offset

simonfall:
# Issue

Currently, offsets and glyph offsets are not scaled with the display scale factor and applied as-is. When using a single display this is not an issue because the offset values can be manually set to achieve the desired results. When using both low-DPI and high-DPI displays however, the missing scaling leads to inconsistent cell size proportions (see attached screenshot).

Screenshot showing cell size inconsistency

# Changes

This PR adds scaling for both offsets (applied in cell size calculation) and glyph offsets (applied in glyph cache). The changes require the glyph cache to be aware of the scale factor.

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -400,7 +400,7 @@ impl Display {
         let rasterizer = Rasterizer::new(scale_factor)?;
 
         debug!("Loading \"{}\" font", &config.font.normal().family);
-        let mut glyph_cache = GlyphCache::new(rasterizer, &config.font)?;
+        let mut glyph_cache = GlyphCache::new(rasterizer, &config.font, window.scale_factor)?;
 
         let metrics = glyph_cache.font_metrics();
         let (cell_width, cell_height) = compute_cell_size(config, &metrics, window.scale_factor);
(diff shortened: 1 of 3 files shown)


kchibisov:
The change is a bit questionable, because some may desire to adjust based on physical pixels and not be constrained by logical pixel grid. It could also weirdly work with non integer scale factors.

simonfall:
I agree that the simplicity of adjusting based on physical pixels is great and should be kept. I still think that there is some value in having an option which allows to use offsets and move a window between displays without changing the window's content (I've attached a screenshot with the same window moved between two displays with different scale factors). Would it be OK to add an option for logical pixels and keep the default based on physical pixels?

high_dpi_vs_low_dpi

Regarding non-integer scale factors: The current fix truncates the calculated physical offsets towards zero but this could also be changed to rounding instead.

kchibisov:
>I agree that the simplicity of adjusting based on physical pixels is great and should be kept. I still think that there is some value in having an option which allows to use offsets and move a window between displays without changing the window's content (I've attached a screenshot with the same window moved between two displays with different scale factors). Would it be OK to add an option for logical pixels and keep the default based on physical pixels?

Maybe due to us scaling the font sizes, actually scaling the offsets is more desirable, but it could lead to situation where you can't offset the way you want. I guess we also have something similar when it comes to padding, and maybe some other metrics, since those are generally using the physical pixels.

simonfall:
Yes, but if that's important there could be a config option to offset based on logical vs on physical pixels. What's your suggested way of proceeding with this, @kchibisov?

PR #7149, opened 2023-08-16
Remove winit dependency from alacritty_config crate

maxbrunsfeld:
The [Zed](https://zed.dev) editor uses the `alacritty_terminal` crate as a library. Currently, that requires that we compile `winit` and its dependencies as part of our build, because the `alacritty_config` depends on `winit` for the `OptionAsAlt` type.

This PR removes the dependency on `winit` from the `alacritty_config` crate. I did this by creating a verbatim copy of the `OptionAsAlt` enum in the `alacritty::config::window` module, and wrote a function that converts this enum into its `winit` equivalent.

It adds a little bit of boilerplate to that module, for converting between the two types, but is pretty simple, and makes the `alacritty_terminal` crate more convenient to use as a library on macOS.

### DIFF
--- a/alacritty/src/config/window.rs
+++ b/alacritty/src/config/window.rs
@@ -7,7 +7,7 @@ use serde::{Deserialize, Deserializer, Serialize};
 use winit::window::{Fullscreen, Theme};
 
 #[cfg(target_os = "macos")]
-use winit::platform::macos::OptionAsAlt;
+use alacritty_terminal::config::OptionAsAlt;
 
 use alacritty_config_derive::{ConfigDeserialize, SerdeReplace};
 use alacritty_terminal::config::{Percentage, LOG_TARGET_CONFIG};
@@ -51,7 +51,7 @@ pub struct WindowConfig {
 
     /// Controls which `Option` key should be treated as `Alt`.
     #[cfg(target_os = "macos")]
-    pub option_as_alt: OptionAsAlt,
+    option_as_alt: OptionAsAlt,
 
     /// Resize increments.
     pub resize_increments: bool,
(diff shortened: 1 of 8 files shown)


alacritty/src/config/window.rs
```diff
@@ -137,6 +137,16 @@ impl WindowConfig {
...
         self.startup_mode == StartupMode::Maximized
     }
+
+    #[cfg(target_os = "macos")]
+    pub fn option_as_alt(&self) -> winit::platform::macos::OptionAsAlt {
+        match self.option_as_alt {
+            OptionAsAlt::OnlyLeft => winit::platform::macos::OptionAsAlt::OnlyLeft,
+            OptionAsAlt::OnlyRight => winit::platform::macos::OptionAsAlt::OnlyRight,
+            OptionAsAlt::Both => winit::platform::macos::OptionAsAlt::Both,
+            OptionAsAlt::None => winit::platform::macos::OptionAsAlt::None,
```
chrisduerr, line 147: Shouldn't use the full path here. Just import it as `WinitOptionsAsAlt` or something.

chrisduerr (changes requested):
Seems easy enough.

alacritty_terminal/src/config/mod.rs
```diff
@@ -215,6 +215,22 @@ impl From<CursorBlinking> for bool {
...
+    OnlyRight,
+
+    /// Both `Option` keys are treated as `Alt`.
+    Both,
+
+    /// No special handling is applied for `Option` key.
+    #[default]
+    None,
+}
+
```
kchibisov, line 233: This is only used in alacritty, the type has no use in `alacritty_terminal`, because it doesn't do anything related to keyboard input, just simply move it into `alacritty`.

kchibisov (changes requested):

alacritty_terminal/src/config/mod.rs
maxbrunsfeld: Ha, what a great point. Thanks.

chrisduerr (approved):

kchibisov (approved):

chrisduerr:
Do you need this released and if so is there any urgency?

chrisduerr:
I was hoping 0.13 wouldn't be **too** far off, but I'll keep it in mind if it's not coming along as quickly as I'd like.

PR #7173, opened 2023-08-26
Update regex-automata to v0.3.6

### LINKED ISSUE #7097: Hint regex with large chars amount. Hanging and continues memory consumption
### Problem

With the `Qm[0-9A-Za-z]{44}` as a hint regex pattern, Alacritty hangs upon activating Hints and memory consumption starts to growth continuously.

```yml
# ... rest of the config
hints:
  enabled:
    - regex: 'Qm[0-9A-Za-z]{44}'
      action: Copy
      binding:
        key: U
        mods: Control|Shift
```

But no issues with `sha256:([0-9a-f]{64})` pattern.

```yml
# ... rest of the config
hints:
  enabled:
    - regex: 'sha256:([0-9a-f]{64})'
      action: Copy
      binding:
        key: U
        mods: Control|Shift
```
No any hangs, hints highlighted almost immediately

### System

Version: 0.12.2
OS: Ubuntu 20.04
Compositor: compton
WM: xmonad

### Logs

Crashes: NONE

Font/Terminal size:
[screenshot](https://github.com/alacritty/alacritty/assets/29537473/e70f66a4-9ef0-4e26-b012-0c8c985db0b0)

Keyboard and bindings:
After activating Hints with the mentioned pattern Alacritty hangs and stops reporting any events with `--report-events` option

chrisduerr:
This seems like a sensible first step before looking into #7097.

---

While this in theory fixes #7097 by erroring-out, I'd like to look into sparse regexes with a follow-up PR that hopefully solves the issue completely and doesn't error out.

### DIFF
--- a/alacritty_terminal/src/term/search.rs
+++ b/alacritty_terminal/src/term/search.rs
@@ -2,7 +2,12 @@ use std::cmp::max;
 use std::mem;
 use std::ops::RangeInclusive;
 
-use regex_automata::{dense, DenseDFA, Error as RegexError, DFA};
+pub use regex_automata::dfa::dense::BuildError;
+use regex_automata::dfa::dense::{Builder, Config, DFA};
+use regex_automata::dfa::Automaton;
+use regex_automata::nfa::thompson::Config as ThompsonConfig;
+use regex_automata::util::syntax::Config as SyntaxConfig;
+use regex_automata::Anchored;
 
 use crate::grid::{BidirectionalIterator, Dimensions, GridIterator, Indexed};
 use crate::index::{Boundary, Column, Direction, Point, Side};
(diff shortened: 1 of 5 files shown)


alacritty_terminal/src/term/search.rs
```diff
@@ -12,47 +17,46 @@ use crate::term::Term;
 /// Used to match equal brackets, when performing a bracket-pair selection.
 const BRACKET_PAIRS: [(char, char); 4] = [('(', ')'), ('[', ']'), ('{', '}'), ('<', '>')];
 
+/// Maximum DFA size to prevent pathological regexes taking down the entire system.
+const MAX_DFA_SIZE: usize = 100_000_000;
```
kchibisov, line 21: How was it computed?

alacritty_terminal/src/term/search.rs
```diff
@@ -629,80 +665,91 @@ mod tests {
...
+        let regex = RegexSearch::new("te.*ing").unwrap();
         let start = Point::new(Line(0), Column(11));
         let end = Point::new(Line(0), Column(0));
-        assert_eq!(term.regex_search_left(&dfas, start, end), Some(end..=start));
+        assert_eq!(term.regex_search_left(&regex, start, end), Some(end..=start));
+    }
+
+    #[test]
+    fn end_on_multibyte_unicode() {
+        let term = mock_term("testвосиб");
```
kchibisov, line 716: That string looks suspicious, but whatever.

alacritty_terminal/src/term/search.rs
```diff
@@ -212,6 +228,7 @@ impl<T> Term<T> {
         let mut c = cell.c;
 
         let mut point = iter.point();
+        let mut last_point = point; // TODO: Hacky af
```
kchibisov, line 231: This TODO should be removed.

alacritty_terminal/src/term/search.rs
```diff
@@ -227,20 +244,30 @@ impl<T> Term<T> {
...
-            if dfa.is_match_or_dead_state(state) {
-                if dfa.is_dead_state(state) {
-                    break;
-                } else {
-                    regex_match = Some(point);
+                // Matches require one additional BYTE of lookahead, so we check the match state for
+                // the first byte of every new character to determine if the last character was a
+                // match.
+                if i == 0 && regex.is_match_state(state) {
+                    regex_match = Some(last_point);
```
kchibisov, line 253: it's not really obvious how it works, I guess the `last_point` is the nearest on to this cell and match is justified by the start of the next cell or when we reach the end?

kchibisov (changes requested):

alacritty_terminal/src/term/search.rs
chrisduerr: Suspicious how? I just copied it from a different test.

alacritty_terminal/src/term/search.rs
chrisduerr: It wasn't computed. I just picked a random number, checked how long it would hang Alacritty up for, then decided that was acceptable.

alacritty_terminal/src/term/search.rs
chrisduerr: New regex-automata does lookaround and requires ingesting one additional byte before deciding if something is a match or not.

So we need to process one more byte, then if it is a match it was the last point that was an actual match.

alacritty_terminal/src/term/search.rs
kchibisov: Huh, ok then, it's just a suffix of my surname.

alacritty_terminal/src/term/search.rs
kchibisov: Yeah, I just wanted to ensure how it works.

kchibisov (approved):

PR #7176, opened 2023-08-26
Add EGL_OVER_GLX environment variable

### LINKED ISSUE #7056: Window is getting stuck until a key press with NVIDIA drivers
Running on an up-to-date Arch install with gnome and Nvidia using default gnome settings. About two months ago, everything worked well, but since then there is serious input lag during use. I've been waiting for a while and using kitty in the meantime hoping an update to nvidia drivers or gnome could solve the problem, but I'm still experiencing this issue. I'm wondering now if this is something that would need to be adjusted on Alacritty's end.

RolfSievert:
After v0.11.0 rendering sometimes paused indefinitely until a keystroke was registered, but not necessarily rendering the change caused by the new keystroke. This issue seems to be related to GLX in conjunction with Nvidia drivers on linux systems, and using EGL instead fixes the issue.

Closes #7056

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Copy global IPC options (`-w -1`) for new windows
 - Bindings to create and navigate tabs on macOS
 - Support startup notify protocol to raise initial window on Wayland/X11
+- Debug option `prefer_egl` to prioritize EGL over other display APIs
 
 ### Changed
 
--- a/alacritty/src/renderer/platform.rs
+++ b/alacritty/src/renderer/platform.rs
@@ -21,15 +21,24 @@ use winit::window::raw_window_handle::{RawDisplayHandle, RawWindowHandle};
 pub fn create_gl_display(
     raw_display_handle: RawDisplayHandle,
     _raw_window_handle: Option<RawWindowHandle>,
+    _prefer_egl: bool,
 ) -> GlutinResult<Display> {
     #[cfg(target_os = "macos")]
     let preference = DisplayApiPreference::Cgl;
 
     #[cfg(windows)]
-    let preference = DisplayApiPreference::Wgl(Some(_raw_window_handle.unwrap()));
+    let preference = if _prefer_egl {
+        DisplayApiPreference::EglThenWgl(Some(_raw_window_handle.unwrap()))
+    } else {
+        DisplayApiPreference::WglThenEgl(Some(_raw_window_handle.unwrap()))
+    };
 
     #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
-    let preference = DisplayApiPreference::GlxThenEgl(Box::new(x11::register_xlib_error_hook));
+    let preference = if _prefer_egl {
+        DisplayApiPreference::EglThenGlx(Box::new(x11::register_xlib_error_hook))
+    } else {
+        DisplayApiPreference::GlxThenEgl(Box::new(x11::register_xlib_error_hook))
+    };
 
     #[cfg(all(not(feature = "x11"), not(any(target_os = "macos", windows))))]
     let preference = DisplayApiPreference::Egl;
--- a/extra/man/alacritty.5.scd
+++ b/extra/man/alacritty.5.scd
@@ -935,6 +935,13 @@ relied upon.
 
 	Default: _false_
 
+*prefer_egl* <boolean>
+
+    Use EGL as display API if the current platform allows it. Note that
+    transparency may not work with EGL on Linux/BSD.
+
+    Default: _false_
+
 # SEE ALSO
 
 *alacritty*(1), *alacritty-msg*(1), *alacritty-bindings*(5)
--- a/alacritty/src/config/debug.rs
+++ b/alacritty/src/config/debug.rs
@@ -23,6 +23,9 @@ pub struct Debug {
     /// The renderer alacritty should be using.
     pub renderer: Option<RendererPreference>,
 
+    /// Use EGL as display API if the current platform allows it.
+    pub prefer_egl: bool,
+
     /// Record ref test.
     #[config(skip)]
     pub ref_test: bool,
@@ -38,6 +41,7 @@ impl Default for Debug {
             highlight_damage: Default::default(),
             ref_test: Default::default(),
             renderer: Default::default(),
+            prefer_egl: Default::default(),
         }
     }
 }
--- a/alacritty/src/window_context.rs
+++ b/alacritty/src/window_context.rs
@@ -92,8 +92,11 @@ impl WindowContext {
         #[cfg(not(windows))]
         let raw_window_handle = None;
 
-        let gl_display =
-            renderer::platform::create_gl_display(raw_display_handle, raw_window_handle)?;
+        let gl_display = renderer::platform::create_gl_display(
+            raw_display_handle,
+            raw_window_handle,
+            config.debug.prefer_egl,
+        )?;
         let gl_config = renderer::platform::pick_gl_config(&gl_display, raw_window_handle)?;
 
         #[cfg(not(windows))]


RolfSievert:
Some thoughts/questions:
- The environment variable should perhaps be passed as an `Option` to `preferred_display_api` so that `platform.rs` not dependent on the environment. I'm new to rust so please advice if you see any design issues.
- Is it a better option to skip the new environment variable entirely and just use EGL over GLX?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support for keybindings with dead keys
 - `Back`/`Forward` mouse buttons support in bindings
 - Copy global IPC options (`-w -1`) for new windows
+- Environment variable `EGL_OVER_GLX` for setting display api priority
```
kchibisov, line 23: We should call this variable like `PREFER_EGL`, and make it works reliably across the platforms.

CHANGELOG.md
```diff
@@ -36,6 +37,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - The double click threshold was raised to `400ms`
 - OSC 52 paste ability is now **disabled by default**; use `terminal.osc52` to adjust it
 - Apply `colors.transparent_background_colors` for selections, hints, and search matches
+- EGL is preferred over GLX when not on Windows or macOs systems
```
kchibisov, line 40: We can't do that, it'll break the existing systems because transparency doesn't work with EGL and mesa. And probably doesn't work with nvidia either.

alacritty/src/renderer/platform.rs
```diff
@@ -20,23 +22,36 @@ use winit::platform::x11;
...
+            DisplayApiPreference::GlxThenEgl(Box::new(x11::register_xlib_error_hook))
+        }
+    }
 
-    let display = unsafe { Display::new(raw_display_handle, preference)? };
-    log::info!("Using {}", { display.version_string() });
-    Ok(display)
+    #[cfg(all(not(feature = "x11"), not(any(target_os = "macos", windows))))]
+    return DisplayApiPreference::Egl;
 }
```
kchibisov, line 55: Could you just merge both functions and not pass `DisplayApi` to the function? there's no need to that, you can pick the display based on the env variable just fine.

alacritty_terminal/src/tty/mod.rs
```diff
@@ -70,6 +70,10 @@ pub fn setup_env(config: &Config) {
     // Advertise 24-bit color support.
     env::set_var("COLORTERM", "truecolor");
 
+    // When not on a Windows or macOs system, EGL is preferred over GLX by default. May be
+    // overridden by user's config below.
+    env::set_var("EGL_OVER_GLX", "true");
```
kchibisov, line 75: Don't set a variable like that, it's user controllable. So completely remove those lines.

extra/man/alacritty.5.scd
```diff
@@ -61,6 +61,7 @@ override variables set by alacritty itself.
 ```
 [env]
 TERM = "alacritty"
+EGL_OVER_GLX = "true"
```
kchibisov, line 64: Just use `PREFER_EGL=1` it's more common for `env` to use numbers.

alacritty/src/renderer/platform.rs
```diff
@@ -20,23 +22,36 @@ use winit::platform::x11;
...
-    let preference = DisplayApiPreference::Wgl(Some(_raw_window_handle.unwrap()));
+    return DisplayApiPreference::Wgl(Some(_raw_window_handle.unwrap()));
 
     #[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
-    let preference = DisplayApiPreference::GlxThenEgl(Box::new(x11::register_xlib_error_hook));
+    {
+        let egl_over_glx = match env::var("EGL_OVER_GLX") {
+            Ok(egl) => egl,
+            Err(_) => panic!("Could not find environment variable 'EGL_OVER_GLX'"),
+        };
```
kchibisov, line 44: No need to do any sort of panics like that it's just error prone, simply handle good value and use prefer otherwise use standard behavior.

So it'll be 
```rust
Ok("1") => DisplayApiPreference::EglThenGlx(Box::new(x11::register_xlib_error_hook)), // Or wgl
_ => // default
```

kchibisov (changes requested):
Thanks.

I wanted to add something like this a while ago as well, just to simply testing different platforms.

extra/man/alacritty.5.scd
chrisduerr: Should probably just add it to our debug section?

kchibisov: Yeah, that as well.

alacritty_terminal/src/tty/mod.rs
RolfSievert: I noticed that the variable stays set even when removed from the config (and restarting), that is why I added this line. Perhaps it requires a restart of the terminal spawning alacritty to reset properly? I'll remove this anyway if you don't see an issue.

CHANGELOG.md
RolfSievert: Do you mean that the flag should affect macOs and Windows systems as well?

kchibisov: Yes, it should work on all platforms.

alacritty_terminal/src/tty/mod.rs
chrisduerr: Shouldn't be a thing, but switching to a config option easily solves this too.

CHANGELOG.md
RolfSievert: Fixed :)

extra/man/alacritty.5.scd
RolfSievert: Numbers aren't currently supported it looks like. Is the string "1" ok?

RolfSievert: Do you mean that it should be added as a Debug option or just added to the section?

alacritty/src/renderer/platform.rs
RolfSievert: Thanks for the example, see the fix :)

alacritty/src/renderer/platform.rs
```diff
@@ -1,5 +1,7 @@
 //! The graphics platform that is used by the renderer.
 
+#[cfg(any(windows, all(feature = "x11", not(any(target_os = "macos", windows)))))]
+use std::env;
```
kchibisov, line 4: Just inline `std` in such case.

alacritty/src/renderer/platform.rs
```diff
@@ -26,10 +28,20 @@ pub fn create_gl_display(
     let preference = DisplayApiPreference::Cgl;
 
     #[cfg(windows)]
-    let preference = DisplayApiPreference::Wgl(Some(_raw_window_handle.unwrap()));
+    let preference = match env::var("PREFER_EGL") {
+        Ok(egl) if egl == "1" => {
+            DisplayApiPreference::EglThenWgl(Some(_raw_window_handle.unwrap()))
+        },
+        _ => DisplayApiPreference::Wgl(Some(_raw_window_handle.unwrap())),
+    };
```
kchibisov, line 36: You can do something like that to match directly without `if`.

```rust
    #[cfg(windows)]
    let preference = match std::env::var("PREFER_EGL").as_ref().map(|s| s.as_str()) {
        Ok("1") => DisplayApiPreference::EglThenWgl(Some(_raw_window_handle.unwrap())),
        _ => DisplayApiPreference::Wgl(Some(_raw_window_handle.unwrap())),
    };

```

extra/man/alacritty.5.scd
```diff
@@ -61,6 +61,7 @@ override variables set by alacritty itself.
 ```
 [env]
 TERM = "alacritty"
+PREFER_EGL = "1"
```
kchibisov, line 64: Maybe we should document this in the `debug` section as well... Or somehow en `env` section, what do you think @chrisduerr ? Maybe we should document other alacritty specific variables?

kchibisov (changes requested):

extra/man/alacritty.5.scd
chrisduerr: There should not be an environment variable for it. Just a debug flag, that's it.

Environment variables are just hacky.

kchibisov: Yeah, I just thought about env variable, since initially I wanted it inside the `glutin`, but in alacritty it's indeed doesn't make that much sense.

kchibisov:
Could you change impl to `bool` debug flag one? It'll be basically the same as it's now, just no `env` stuff.

RolfSievert:
> Could you change impl to `bool` debug flag one? It'll be basically the same as it's now, just no `env` stuff.

Yeah sure! Perhaps this PR should't close the issue itself but act as a temporary solution to those with the problem, what do you think?

kchibisov:
Yeah, it shouldn't fix the issue itself.

alacritty/src/renderer/platform.rs
```diff
@@ -21,15 +21,25 @@ use winit::platform::x11;
 pub fn create_gl_display(
     raw_display_handle: RawDisplayHandle,
     _raw_window_handle: Option<RawWindowHandle>,
+    #[cfg(any(windows, all(feature = "x11", not(any(target_os = "macos", windows)))))]
+    prefer_egl: bool,
```
chrisduerr, line 25: Would be easier to just add this on all platforms and avoid the cfg on the caller.

extra/man/alacritty.5.scd
```diff
@@ -880,6 +880,12 @@ relied upon.
 
 	Default: _false_
 
+*prefer_egl* <boolean>
+
+    Use EGL as display API if the current platform allows it.
+
+    Default: _false_
```
chrisduerr, line 887: This should definitely mention the drawbacks.

chrisduerr (changes requested):

alacritty/src/renderer/platform.rs
RolfSievert: So we're okay with passing unused parameters? I was a bit hesitant about this since it's only a bool, and I can agree with you in this case.

kchibisov: yeah, for debug config options it's fine to do so.

alacritty/src/cli.rs
```diff
@@ -72,6 +72,10 @@ pub struct Options {
     /// Subcommand passed to the CLI.
     #[clap(subcommand)]
     pub subcommands: Option<Subcommands>,
+
+    /// Use EGL as display API if the current platform allows it.
+    #[clap(long)]
+    pub prefer_egl: bool,
```
kchibisov, line 78: We don't special options like that for debug settings, they could be set via the `-o`.

kchibisov (changes requested):
Remove the cli option, because it's already exposed via cli, so extra setting is redundant.

extra/man/alacritty.5.scd
```diff
@@ -880,6 +880,13 @@ relied upon.
 
 	Default: _false_
 
+*prefer_egl* <boolean>
+
+    Use EGL as display API if the current platform allows it. Note that
+    transparency does not work with EGL.
```
kchibisov, line 886: Maybe something like that?

```suggestion
    Use EGL as display API if the current platform allows it. Note that
    transparency may not work with EGL on Linux/BSD.
```

alacritty/src/renderer/platform.rs
```diff
@@ -21,15 +21,24 @@ use winit::platform::x11;
...
 ) -> GlutinResult<Display> {
     #[cfg(target_os = "macos")]
     let preference = DisplayApiPreference::Cgl;
 
     #[cfg(windows)]
-    let preference = DisplayApiPreference::Wgl(Some(_raw_window_handle.unwrap()));
+    let preference = if _prefer_egl {
+        DisplayApiPreference::EglThenWgl(Some(_raw_window_handle.unwrap()))
+    } else {
+        DisplayApiPreference::Wgl(Some(_raw_window_handle.unwrap()))
```
kchibisov, line 33: Why it's not `WglThenEgl`?

RolfSievert: Good point!

kchibisov (approved):

chrisduerr (approved):

CHANGELOG.md
guidocella: Mesa 24.1 finally fixed background transparency on EGL on X11. It always worked on Nvidia.

PR #7177, opened 2023-08-27
Update dependencies

chrisduerr:
This patch applies all breaking and non-breaking dependencies, to ensure everything is using the latest available package versions.

The only things which weren't updated were mio and wayland-client.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -60,7 +60,6 @@ dependencies = [
  "serde_yaml",
  "toml 0.8.2",
  "unicode-width",
- "wayland-client 0.31.1",
  "windows-sys 0.48.0",
  "winit",
  "x11-dl",
@@ -1582,15 +1581,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "quick-xml"
-version = "0.30.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eff6510e86862b57b210fd8cbe8ed3f0d7d600b9c2863cd4549a2e033c66e956"
-dependencies = [
- "memchr",
-]
-
-[[package]]
 name = "quote"
 version = "1.0.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1864,7 +1854,7 @@ dependencies = [
  "memmap2",
  "nix 0.26.4",
  "thiserror",
- "wayland-backend 0.1.2",
+ "wayland-backend",
  "wayland-client 0.30.2",
  "wayland-cursor 0.30.0",
  "wayland-protocols 0.30.1",
@@ -2219,20 +2209,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "wayland-backend"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "19152ddd73f45f024ed4534d9ca2594e0ef252c1847695255dae47f34df9fbe4"
-dependencies = [
- "cc",
- "downcast-rs",
- "nix 0.26.4",
- "scoped-tls",
- "smallvec",
- "wayland-sys 0.31.1",
-]
-
-[[package]]
 name = "wayland-client"
 version = "0.29.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2257,20 +2233,8 @@ dependencies = [
  "bitflags 1.3.2",
  "calloop",
  "nix 0.26.4",
- "wayland-backend 0.1.2",
+ "wayland-backend",
  "wayland-scanner 0.30.1",
-]
-
-[[package]]
-name = "wayland-client"
-version = "0.31.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1ca7d52347346f5473bf2f56705f360e8440873052e575e55890c4fa57843ed3"
-dependencies = [
- "bitflags 2.4.0",
- "nix 0.26.4",
- "wayland-backend 0.3.2",
- "wayland-scanner 0.31.0",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


Cargo.lock
```diff
@@ -337,43 +337,42 @@ dependencies = [
...
-checksum = "7f6b5c519bab3ea61843a7923d074b04245624bb84a64a8c150f5deb014e388b"
+checksum = "586a385f7ef2f8b4d86bddaa0c094794e7ccbfe5ffef1f434fe928143fc783a5"
 dependencies = [
  "clap",
 ]
 
 [[package]]
 name = "clap_derive"
-version = "4.3.2"
+version = "4.4.0"
```
kchibisov, line 361: This is `1.70` msrv btw.

Cargo.lock
```diff
@@ -818,9 +830,9 @@ dependencies = [
 
 [[package]]
 name = "glutin"
-version = "0.30.9"
+version = "0.30.10"
```
kchibisov, line 833: I've updated it here https://github.com/alacritty/alacritty/pull/7079/files , but it's till in review...

Cargo.lock
```diff
@@ -1664,15 +1669,15 @@ checksum = "e1cf6437eb19a8f4a6cc0f7dca544973b0b78843adbfeb3683d1a94a0024a294"
...
-version = "1.1.0"
+version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d29ab0c6d3fc0ee92fe66e2d99f700eab17a8d57d1c1d3b748380fb20baa78cd"
+checksum = "94143f37725109f92c262ed2cf5e59bce7498c01bcc1502d7b9afe439a4e9f49"
 
 [[package]]
 name = "sctk-adwaita"
-version = "0.6.0"
+version = "0.6.1"
```
kchibisov, line 1678: This is also updated on the winit branch iirc.

alacritty/Cargo.toml
```diff
@@ -32,13 +32,13 @@ glutin = { version = "0.30.4", default-features = false, features = ["egl", "wgl
 home = "0.5.5"
 libc = "0.2"
 log = { version = "0.4", features = ["std", "serde"] }
-notify = "5.1.0"
+notify = "6.1.1"
```
kchibisov, line 35: It hasn't fixed anything? Like partial writes, etc?

kchibisov (changes requested):

kchibisov:
I'd also suggest to handle https://github.com/alacritty/alacritty/pull/6846 as well as https://github.com/alacritty/alacritty/pull/7079 before this, because they sort of shuffle deps and you generate more work for them as well.

Cargo.lock
chrisduerr: I'm happy to wait for the winit branch.

alacritty/Cargo.toml
chrisduerr: Nothing particularly relevant as far as I can see:
https://github.com/notify-rs/notify/blob/HEAD/CHANGELOG.md#notify-600-2023-05-17

alacritty/Cargo.toml
```diff
@@ -55,12 +55,11 @@ xdg = "2.5.0"
 png = { version = "0.17.5", default-features = false, optional = true }
 
 [target.'cfg(target_os = "macos")'.dependencies]
-cocoa = "0.24.0"
+cocoa = "0.25.0"
 objc = "0.2.2"
 
 [target.'cfg(not(any(target_os="windows", target_os="macos")))'.dependencies]
 x11-dl = { version = "2", optional = true }
-wayland-client = { version = "0.29.0", features = ["dlopen"], optional = true }
```
chrisduerr, line 63: @kchibisov was this added to force the `dlopen` feature, or did we actually use it in the past? It seems like that got removed from the latest wayland-client versions.

If the feature name was just changed and we want this back, let me know.

Cargo.lock
chrisduerr: sid has 1.70.0, so this should be fine.

alacritty/Cargo.toml
kchibisov: you can remove it, it's fine.

kchibisov (approved):
Should be fine after CI fix.

Cargo.lock
```diff
@@ -2015,9 +2039,15 @@ checksum = "1dd624098567895118886609431a7c3b8f516e41d30e0643f03d94592a147e36"
...
+version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c0edd1e5b14653f783770bce4a4dabb4a5108a5370a5f5d8cfe8710c361f6c8b"
+checksum = "e51733f11c9c4f72aa0c160008246859e340b00807569a0da0e7a1079b27ba85"
+
+[[package]]
+name = "unsafe-libyaml"
+version = "0.2.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f28467d3e1d3c6586d8f25fa243f544f5800fec42d97032474e17222c2b75cfa"
```
kchibisov, line 2050: Actually, could you roll back this part of the update? We'll drop yml anyway, and I don't want to deal with this mess. It's a libyaml transpiled into rust, so basically a 'nice way to add a security nightmare' sort of things.

kchibisov (changes requested):
drop yml update.

Cargo.lock
chrisduerr: I mean yaml is just a security nightmare anyway. I don't see how keeping an unmaintained version would be more secure.

And yaml probably isn't going away soon.

kchibisov: I mean, they are both not maintained. It's just one is a massive transpilation of a C library, and the rust crate is a small safe rust parser. They're not maintained from the same year.

kchibisov (approved):
I still don't like the yml change, but given that fixes for the buffer/stack overflows in libyaml are getting backported, it's probably not that bad, given that rust yaml lib used before was based on the `libyaml` and had the same CVEs.

PR #7179, opened 2023-08-28
Start terminal even if `working_directory` is invalid, show a warning

proski:
Failing to start due to invalid working directory makes it necessary to use some other tool to fix the configuration.

`std::process::Command` doesn't distinguish errors due to invalid shell and invalid start directory. Instead, change directory for the main process and warn if it fails.

### DIFF
--- a/alacritty_terminal/src/tty/unix.rs
+++ b/alacritty_terminal/src/tty/unix.rs
@@ -2,15 +2,16 @@
 
 use std::ffi::CStr;
 use std::fs::File;
-use std::io::{Error, ErrorKind, Result};
+use std::io::{Error, ErrorKind, Read, Result, Write};
 use std::mem::MaybeUninit;
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 use std::os::unix::process::CommandExt;
 use std::process::{Child, Command, Stdio};
 use std::{env, ptr};
 
+use bitflags::bitflags;
 use libc::{self, c_int, winsize, TIOCSCTTY};
-use log::error;
+use log::{error, warn};
 use mio::unix::EventedFd;
 use nix::pty::openpty;
 #[cfg(any(target_os = "linux", target_os = "macos"))]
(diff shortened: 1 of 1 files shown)


proski:
`std::process::Command` returns exactly the same error (No such file or directory) whether the working directory or the shell is missing. There is no way to figure out what's work other than checking the existence of the shell and the working directory, which is racy and duplicates code.

This PR makes the main process take care of the current directory. It might seem heavy-handed to change directory of the main process, but it's aligned with user expectations, and it makes logging easy.

proski:
alacritty-screenshot

proski:
I'm currently trying another approach that uses sending status of `pre_exec` over the pty file handles from the child to the parent process. It might be more extensible and and won't require directory change in the parent process. Marking this PR as draft for now.

proski:
I've remade the PR to change directory in `pre_exec`. The parent process doesn't change the directory, only the child process does. I added a mechanism for passing warnings to the parent process. That mechanism is used for two other warnings that were not handled well (setting session ID and controlling terminal).

Maybe it would be possible to use terminal events (like those used to set the title) to deliver warnings, I'll have a look (no promises). In any case, I would appreciate comments.

proski:
Changed to use `std::io::{Read,Write}` instead of libc functions. Removed `die!`, all errors are handled in a better way.

kchibisov (commented):
In general the idea to handle error here and just start alacritty sounds good to me.

I guess the reason you've picked approach like that is because it's less racy than checking the directly before hand and doing a fallback? Also, how the windows parts should look like, since we have similar issues with it, I believe, but if you're not interested in it, then I understand.

chrisduerr (changes requested):
We're intentionally handling this only in `--working-directory` but not in the config because it doesn't make sense to start Alacritty with a broken working directory.

Why would we want this?

Also if we do want this, this isn't the way to go about it. It's way too complicated. Just use whatever `--working-directory` does already.

proski:
Sorry for not responding earlier. I'll redo the whole PR if I have a chance. I'm going to rewrite the code using a dedicated socket pair and free-form log messages. The pseudoterminal socket is not reliable due to control character conversion. I do believe that "it's better to ask for forgiveness than for permission" in software, as it eliminates duplication of sanity checks. Closing this PR for now.

PR #7181, opened 2023-08-29
Fix crash with anchored searches

chrisduerr:
While this does **not** enable the use of anchors (`^`) in user regexes, it does prevent Alacritty from crashing when attempting to do so.


alacritty_terminal/src/term/search.rs
```diff
@@ -8,6 +8,7 @@ use regex_automata::dfa::Automaton;
 use regex_automata::nfa::thompson::Config as ThompsonConfig;
 use regex_automata::util::syntax::Config as SyntaxConfig;
 use regex_automata::Anchored;
+use regex_automata::Input;
```
kchibisov, line 11: These can clearly be merged.

kchibisov (approved):

PR #7196, opened 2023-09-05
Remove obsolete ansicode.txt file

BuriedInTheGround:


kchibisov (approved):
I still sort of think that this file shouldn't exist, because the docs are present at https://invisible-island.net/xterm/ctlseqs/ctlseqs.html.

BuriedInTheGround:
I have to say that I don't disagree. The file is historically present in the repo, but documentation on ANSI sequences is easily found online. Personally, it makes no difference to me whether the file stays or not, I leave that decision to the maintainers of alacritty.

kchibisov:
The thing is that this file is basically an original pdf beeing rewritten, we do provide links to the docs that should be used on the wiki. The links on the wiki page provide the extensive reference https://github.com/alacritty/alacritty/wiki/ANSI-References .

It just I know that none of the maintainers use this file. But I'd rather handle the deletion separately with @chrisduerr .

BuriedInTheGround:
@kchibisov if you agree, I'll delete the file and update the PR.

kchibisov:
aye, you can delete the file.

chrisduerr (approved):

PR #7199, opened 2023-09-07
Test man-pages compilation on the CI

kchibisov:


.builds/linux.yml
```diff
@@ -18,13 +19,23 @@ environment:
...
-  - test: |
-      cd alacritty
-      cargo test
   - rustfmt: |
       cd alacritty
       rustup toolchain install nightly -c rustfmt
       cargo +nightly fmt -- --check
+  - man-pages: |
+      cd alacritty
+      printf "Compiling alacritty.1.scd" > /dev/null
```
chrisduerr, line 28: Shouldn't this be written to stdout, instead of to /dev/null? I'd assume it's because `scdoc` doesn't reference the input file and thus errors cannot clearly identify a file.

.builds/linux.yml
```diff
@@ -18,13 +19,23 @@ environment:
...
+      scdoc < extra/man/alacritty.1.scd > /dev/null
+      printf "Compiling alacritty-msg.1.scd" > /dev/null
+      scdoc < extra/man/alacritty-msg.1.scd > /dev/null
+      printf "Compiling alacritty.5.scd" > /dev/null
+      scdoc < extra/man/alacritty.5.scd > /dev/null
+      printf "Compiling alacritty-bindings.5.scd" > /dev/null
+      scdoc < extra/man/alacritty-bindings.5.scd > /dev/null
+  - test: |
+      cd alacritty
+      cargo test
```
chrisduerr, line 38: Is there a reason why test was moved after rustfmt?

chrisduerr (changes requested):

.builds/linux.yml
kchibisov: to make faster jobs before, so they fail sooner.

.builds/linux.yml
kchibisov: I'd suggest to look into the CI output, it's better to understand like that. Basically all redirection and the way output appear is not consistent, so it's just to help identify the command being run.

kchibisov: that's how it looks without redirection https://builds.sr.ht/~kchibisov/job/1054018#task-man-pages

chrisduerr: So what's the point of having it in the first place?

kchibisov: yeah, so I can at least know which command failed.

chrisduerr (approved):

PR #7204, opened 2023-09-08
Fix regex memory usage

### LINKED ISSUE #7097: Hint regex with large chars amount. Hanging and continues memory consumption

chrisduerr:
This fixes an issue where regexes with a large number of possible states
would consume excessive memory, since the entire DFA was compiled ahead
of time.

To solve this, the DFA is now built at runtime using `regex-automata`'s
hybrid DFA.

There are however still some checks performed ahead of time, causing
errors with obscenely large regexes (`[0-9A-Za-z]{999999999}`), which
shouldn't cause any issues.

A regex which is large, but not large enough to fail the NFA
construction (like `[0-9A-Za-z]{999999}`) will cause a long search of
the entire grid, but will complete and show the match.

Closes #7097.

### DIFF
--- a/alacritty_terminal/src/term/search.rs
+++ b/alacritty_terminal/src/term/search.rs
@@ -1,4 +1,5 @@
 use std::cmp::max;
+use std::error::Error;
 use std::mem;
 use std::ops::RangeInclusive;
 
@@ -16,9 +17,6 @@ use crate::term::Term;
 /// Used to match equal brackets, when performing a bracket-pair selection.
 const BRACKET_PAIRS: [(char, char); 4] = [('(', ')'), ('[', ']'), ('{', '}'), ('<', '>')];
 
-/// Maximum DFA size to prevent pathological regexes taking down the entire system.
-const MAX_DFA_SIZE: usize = 100_000_000;
-
 pub type Match = RangeInclusive<Point>;
 
 /// Terminal regex search state.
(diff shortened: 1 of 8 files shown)


alacritty_terminal/src/term/search.rs
```diff
@@ -201,8 +222,29 @@ impl<T> Term<T> {
         end: Point,
         direction: Direction,
         anchored: bool,
-        regex: &impl Automaton,
+        regex: &mut LazyDfa,
     ) -> Option<Point> {
+        match self.regex_search_internal(start, end, direction, anchored, regex) {
+            Ok(regex_match) => regex_match,
+            Err(err) => {
+                log::warn!("Regex exceeded complexity limit");
```
kchibisov, line 230: Maybe we should say which regex somehow? Also, import `warn!` and `debug!`.

kchibisov (changes requested):

alacritty_terminal/src/term/search.rs
chrisduerr: I considered that but it's only really viable if we store the pattern in the dfa and I didn't want to do that just for log messages. It should be pretty clear anyway since it's triggered by user actions usually (and should be very rare).

alacritty_terminal/src/term/search.rs
```diff
@@ -245,27 +287,27 @@ impl<T> Term<T> {
                 };
 
                 // Since we get the state from the DFA, it doesn't need to be checked.
```
nixpulvis, line 289: Perhaps update this comment since it doesn't need to mention being checked anymore?

alacritty_terminal/src/term/search.rs
```diff
@@ -881,19 +923,34 @@ mod tests {
...
+    }
+
+    #[test]
+    fn runtime_cache_error() {
+        let term = mock_term(&str::repeat("i", 9999));
+
+        let mut regex = RegexSearch::new("[0-9A-Za-z]{9999}").unwrap();
+        let start = Point::new(Line(0), Column(0));
+        let end = Point::new(Line(0), Column(9999));
+        assert_eq!(term.regex_search_right(&mut regex, start, end), None);
```
nixpulvis, line 954: Would it be worth making these errors a little more descriptive, since they could in theory be hit by a crazy user configuration. Always nice to help tame insanity imo.

kchibisov (approved):

PR #7208, opened 2023-09-11
Added mouse "leader_mod" option

JezerM:
Previously, to use a mouse binding (or for executing a hint action) the used needed to press the "Shift" modifier. This new option allows to change this prefix/leader modifier.

As a Mac user, I prefer to use `Command+Click` to open a URL instead of `Shift+Click`, so this change could fix this preference issue.


kchibisov:
It's a bit more complicated, since what you're talking about is entirely about a mouse mode. Since normal alacritty config opens links without an extra modifiers. I'm not sure that we should configure it though, given that `Shift` is very common for that and it's non configurable most of the time.

JezerM:
Yeah, I'm talking about the situation when a program is capturing the mouse (as documented), in which case you must use `Shift` modifier to open the link. I think that the option to change it to another Modifier should be there.

PR #7211, opened 2023-09-11
Make clippy ignore `skip(0)`, it's intentional

proski:


proski:
Only nightly is affected, so perhaps it's not very urgent.

An alternative would be to refactor the code to avoid using the `Skip` type as the holder of the text being shortened. But I cannot think of any simple change at the moment.

proski:
Need to find a way to keep older clippy happy.

chrisduerr (changes requested):
I'm aware of this issue, but no clippy changes will be merged unless at least stable is affected. Everything else would just cause unnecessary code churn.

PR #7212, opened 2023-09-12
Add missing position config docs

chrisduerr:


extra/man/alacritty.5.scd
```diff
@@ -88,6 +88,14 @@ This section documents the *[window]* table of the configuration file.
 
 	Default: _{ column = 0, lines = 0 }_
 
+*position* { x = <integer>, y = <integer> }
+
+	Window startup position
+
+	Specified in number of pixels.
+
+	If the position is not set, the window manager will handle placement.
```
kchibisov, line 97:
```suggestion
*position* "None" | { x = <integer>, y = <integer> }

	Window startup position

	Specified in number of pixels.

	If the position is _"None"_, the window manager will handle placement.
	
	Default: _"None"_
```

kchibisov (changes requested):
It's not quite accurate, that should fix it.

kchibisov (approved):

PR #7214, opened 2023-09-13
Add max_intensity config option for visual bell

ARitz-Cracker:
I like the idea of a visual bell, but I also think that having the flash colour be 100% opaque is a bit too visually jarring. This PR adds a `bell.max_intensity` config option which allows users to set an initial opacity for the flash.

### DIFF
--- a/alacritty/src/config/bell.rs
+++ b/alacritty/src/config/bell.rs
@@ -2,10 +2,10 @@ use std::time::Duration;
 
 use alacritty_config_derive::ConfigDeserialize;
 
-use alacritty_terminal::config::Program;
+use alacritty_terminal::config::{Percentage, Program};
 use alacritty_terminal::term::color::Rgb;
 
-#[derive(ConfigDeserialize, Clone, Debug, PartialEq, Eq)]
+#[derive(ConfigDeserialize, Clone, Debug, PartialEq)]
 pub struct BellConfig {
     /// Visual bell animation function.
     pub animation: BellAnimation,
@@ -15,6 +15,9 @@ pub struct BellConfig {
 
     /// Visual bell flash color.
     pub color: Rgb,
+
+    /// Maximum opacity of the flash color, from 0.0 to 1.0.
+    pub max_intensity: Percentage,
 
     /// Visual bell duration in milliseconds.
     duration: u16,
@@ -27,6 +30,7 @@ impl Default for BellConfig {
             animation: Default::default(),
             command: Default::default(),
             duration: Default::default(),
+            max_intensity: Default::default(),
         }
     }
 }
--- a/alacritty/src/display/bell.rs
+++ b/alacritty/src/display/bell.rs
@@ -11,6 +11,9 @@ pub struct VisualBell {
 
     /// The last time the visual bell rang, if at all.
     start_time: Option<Instant>,
+
+    /// Visual bell intensity multiplier
+    intensity_multiplier: f32,
 }
 
 impl VisualBell {
@@ -23,6 +26,7 @@ impl VisualBell {
 
     /// Get the currently intensity of the visual bell. The bell's intensity
     /// ramps down from 1.0 to 0.0 at a rate determined by the bell's duration.
+    /// It is then multiplied with the intensity_multiplier.
     pub fn intensity(&self) -> f64 {
         self.intensity_at_instant(Instant::now())
     }
@@ -42,7 +46,7 @@ impl VisualBell {
 
     /// Get the intensity of the visual bell at a particular instant. The bell's
     /// intensity ramps down from 1.0 to 0.0 at a rate determined by the bell's
-    /// duration.
+    /// duration. It is then multiplied with the intensity_multiplier.
     pub fn intensity_at_instant(&self, instant: Instant) -> f64 {
         // If `duration` is zero, then the VisualBell is disabled; therefore,
         // its `intensity` is zero.
@@ -93,7 +97,7 @@ impl VisualBell {
 
                 // Since we want the `intensity` of the VisualBell to decay over
                 // `time`, we subtract the `inverse_intensity` from 1.0.
-                1.0 - inverse_intensity
+                (1.0 - inverse_intensity) * self.intensity_multiplier as f64
             },
         }
     }
@@ -110,6 +114,7 @@ impl From<&BellConfig> for VisualBell {
             animation: bell_config.animation,
             duration: bell_config.duration(),
             start_time: None,
+            intensity_multiplier: bell_config.max_intensity.as_f32(),
         }
     }
 }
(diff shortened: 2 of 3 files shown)


chrisduerr (changes requested):
Just change the color.

ARitz-Cracker:
> Just change the color.

That doesn't create the effect that I want, whatever the colour is, the text being completely obscured is what's too visually jarring for me.

kchibisov:
Maybe we should just start from the transparent color instead of fully opaque one? I also don't think we need an option for that, but picking starting alpha around 85 should be good?

ARitz-Cracker:
> Maybe we should just start from the transparent color instead of fully opaque one? I also don't think we need an option for that, but picking starting alpha around 85 should be good?

But that's the thing, you think `0.85` is good, and it might be for your setup, but I have a near-pitch-black background colour, and with my patch in place, I'm using a value of `0.15`, because that's a value that noticeably flashes the background while leaving the text mostly unaffected.

Either way, I don't understand the resistance to this addition, users can already completely customize their colour pallet which makes any one-size-fits-all solution non-ideal, the next version of alacritty that's going to be released is going to be a "breaking" version change, and I've already completed all the (fairly trivial) work needed to make this happen.

kchibisov:
The whole issue is adding lots of options for the features, which are not really used. It's not that hard to apply a patch to software... I'm still not sure why starting from some opacity level, maybe 50 shouldn't work with the short animation times.

ARitz-Cracker:
> The whole issue is adding lots of options for the features, which are not really used. It's not that hard to apply a patch to software... I'm still not sure why starting from some opacity level, maybe 50 shouldn't work with the short animation times.

I'm sorry, but I still don't understand your position, the `README.md` of this repo advertises that alacritty "allows for extensive configuration" if that is still true, then I thought that this would be a welcome change, especially since the changes to the codebase to allow for this customization was literally less than 10 lines of actual code. Since the moment you allow users to customize all their colours, any assumption that you may have as to how every colour setting interacts with every other colour setting is invalid. Again, you just suggested a different value, which again, may work for some setups, but not for others.

With the colour pallet that I use, a `0.15` opacity is how I can flash the background while leaving the text (with its potential varying colours) mostly unaffected. But I am also not going to assume that everyone else in the entire world would want the same effect, nor would I assume any arbitrary initial opacity value would work with all colour pallets. Which is why my PR is "allow for config of the initial opacity" and not "reduce the initial opacity to 0.15 because that's what I like"

T-O-R-U-S:
> Just change the color.

Honestly I feel like this take is just really shallow, and maybe a little condescending? Considering you can already pick the colour of the flash, I don't think being able to pick the alpha of that colour as well is too far of a stretch. I mean, you guys are already going so far as to change the opacity to an arbitrary value, but for some reason are opposed to this value being customizable?

I think this change is quite justified. And should be renamed to `bell_opacity` instead of `max_intensity`, more self-explanatory that way.

nixpulvis:
I'll just jump in here to say, color / contrast arguments have historically been somewhat difficult. That said, given that there isn't currently a way to configure this, it may be more acceptable. I don't personally use the visual bell, so my opinion might be less meaningful, but I can see where @ARitz-Cracker is coming from with the desire to use transparency. I think I'd have to test it for myself on many different color schemes to see how much configuration is needed, but I without doing that, I'll just say that we currently allow changing the `animation` and `duration`, so color doesn't seem like a far reach to ask.

I think that this PR is the wrong approach however. I would rather see something that allows setting the color with an alpha channel. Then testing the string for the alpha channel and using that to decide on the transparency of the flash fadeout. Or changes to the `animation` configuration. Though I do think that option is a bit bloated already.

kchibisov:
The thing is that the option is not really used at all, and adding more config options to a rarely used option adds an extra maintenance. I've suggested to use alpha other than 1, because it won't change anything.

>I'd have to test it for myself on many different color schemes to see how much configuration is needed, but I without doing that, I'll just say that we currently allow changing the animation and duration, so color doesn't seem like a far reach to ask.

@nixpulvis we allow, duration, animation, and color. They want a starting alpha. Which is why I'm suggesting to use a default alpha value around 0.5, because I still have no clue which terminal even allows you to set **4** options for a visual bell, which majority users simply disable because it's a destructing thing. It's even disabled by default, because constant flashing during output could cause issue, and for some group of ppl they could be non pleasant.

>I think that this PR is the wrong approach however. I would rather see something that allows setting the color with an alpha channel. Then testing the string for the alpha channel and using that to decide on the transparency of the flash fadeout. Or changes to the animation configuration. Though I do think that option is a bit bloated already

The issue is that animations rely on changing alpha, and the function is on the segment from [0;1] resulting in values in the same segment. The start is from one and it *fades*.

nixpulvis:
> The issue is that animations rely on changing alpha

@kchibisov I guess this is why I was thinking it _may_ be worth looking at a refactor of the whole `animation` configuration. But again, I don't use this personally and I know larger refactors like this need to be strongly motivated to make up for the breakage for current users.

I also would like to **reduce** configuration complexity, not increase it here.

ARitz-Cracker:
@nixpulvis the datatypes used for colour configs are all RGB, not RGBA, so that would require creating a new data type just to throw it away immediately afterwards, which adds complexity. Multiplying the output of the animation function @kchibisov mentioned with a float from 0 to 1 was the least complex/destructive/increase-in-maintenece way I could think of to add the ability to have opacity customization. It's literally just multiplying a number with another number, I acknowledge that OSS projects are resistant to changes that require an increase in maintenance, something which I considered when coming up with this implementation.

My entire goal is to have the visual bell be less "non-pleasent", and provide a way for people to have that experience with differing colour pallets.

Sidenote, I have created a new commit which renames my proposed new config option from `max_intensity` to `max_opacity` but it seems like this PR didn't update, so if you guys are willing to have this addition, I can open a new PR with the updated name.

nixpulvis:
I don't see how an RGBA struct is an increase in maintenance, seeing as it may prove useful in other places. Whereas this is a one off configuration option, which can only be described as an increase in complexity, albeit a small one.

That said, I'm more interested in keeping the `bell` config section short, so that may be my bias.

ARitz-Cracker:
> I don't see how an RGBA struct is an increase in maintenance, seeing as it may prove useful in other places. Whereas this is a one off configuration option, which can only be described as an increase in complexity, albeit a small one.
> 
> That said, I'm more interested in keeping the `bell` config section short, so that may be my bias.

The thing is, the bell colour config is directly used to draw the rectangle over the entire window [as seen here](https://github.com/alacritty/alacritty/blob/34b5dbacd28cd1abaedf1d81cc0ebe57aa44a086/alacritty/src/display/mod.rs#L867C14-L867C14) so if we were going to make an RGBA config struct (which I am willing to do) I'd have to `impl From<Rgba> for Rgb` that would strip off the alpha value. This is something I am willing to do in a new PR tomorrow. The alpha value in the config will be used to derive the `intensity_multiplier` I have already.

If this is the preferred approach to getting customizable opacity for the bell flash, then I'll definitely be willing to submit a new PR with those additions.

PR #7225, opened 2023-09-19
Add inline vi mode search

### LINKED ISSUE #7203: vi mode find with `f` and `t`
I'm not sure if this is a bug report or feature request, but since the [docs](https://github.com/alacritty/alacritty/blob/master/extra/man/alacritty.5.scd#keyboard) don't mention it, it's probably a feature request.

### System

OS: NixOS
Version: 0.12.0
Linux/BSD: sway/wayland

### Logs

I thought it might be related to https://github.com/alacritty/alacritty/issues/4111 but it seems to be picking up the keys.
```
$ alacritty --print-events
...
[33.647854165s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(94491515078192)), event: KeyboardInput { device_id: DeviceId(Wayland(DeviceId)), input: KeyboardInput { scancode: 33, state: Pressed, virtual_keycode: Some(F), modifiers: (empty) }, is_synthetic: false } }
[33.648697908s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(94491515078192)), event: KeyboardInput { device_id: DeviceId(Wayland(DeviceId)), input: KeyboardInput { scancode: 33, state: Released, virtual_keycode: Some(F), modifiers: (empty) }, is_synthetic: false } }
[38.334091594s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(94491515078192)), event: KeyboardInput { device_id: DeviceId(Wayland(DeviceId)), input: KeyboardInput { scancode: 20, state: Pressed, virtual_keycode: Some(T), modifiers: (empty) }, is_synthetic: false } }
[38.405158533s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(94491515078192)), event: KeyboardInput { device_id: DeviceId(Wayland(DeviceId)), input: KeyboardInput { scancode: 20, state: Released, virtual_keycode: Some(T), modifiers: (empty) }, is_synthetic: false } }
```

chrisduerr:
This patch adds inline search to vi mode using `f`/`F` and `t`/`T` as default bindings. The behavior matches that of vim.

Closes #7203.

---

There's a very good chance I missed an edge case or 16, so this probably needs a bit of manual testing.

### DIFF
--- a/alacritty_terminal/src/term/search.rs
+++ b/alacritty_terminal/src/term/search.rs
@@ -513,7 +513,25 @@ impl<T> Term<T> {
     }
 
     /// Find left end of semantic block.
-    pub fn semantic_search_left(&self, mut point: Point) -> Point {
+    #[must_use]
+    pub fn semantic_search_left(&self, point: Point) -> Point {
+        match self.inline_search_left(point, &self.semantic_escape_chars) {
+            Ok(point) => self.grid.iter_from(point).next().map_or(point, |cell| cell.point),
+            Err(point) => point,
+        }
+    }
+
+    /// Find right end of semantic block.
+    #[must_use]
+    pub fn semantic_search_right(&self, point: Point) -> Point {
+        match self.inline_search_right(point, &self.semantic_escape_chars) {
+            Ok(point) => self.grid.iter_from(point).prev().map_or(point, |cell| cell.point),
+            Err(point) => point,
+        }
+    }
+
+    /// Searching to the left, find the next character contained in `needles`.
+    pub fn inline_search_left(&self, mut point: Point, needles: &str) -> Result<Point, Point> {
         // Limit the starting point to the last line in the history
         point.line = max(point.line, self.topmost_line());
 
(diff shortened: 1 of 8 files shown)


chrisduerr:
@nixpulvis I didn't get to do much testing today, but I think your specific issue should be resolved. At least I couldn't immediately repro it anymore. Let me know if you find anything else.

nixpulvis:
Let me try and do another lookover tomorrow, back on my ol' trusty Linux machine. This Mac is unable to run the branch until the PTY winit thing is resolved I guess.

alacritty_terminal/src/term/search.rs
```diff
@@ -450,41 +468,41 @@ impl<T> Term<T> {
...
-            if cell.point.column == last_column && !cell.flags.contains(Flags::WRAPLINE) {
-                break; // cut off if on new line or hit escape char
+            if !cell.flags.intersects(wide) && needles.contains(cell.c) {
+                return Ok(point);
             }
 
-            point = cell.point;
+            if point.column == last_column && !cell.flags.contains(Flags::WRAPLINE) {
+                break;
+            }
```
kchibisov, line 479: This code doesn't match all semantics for the `semantic_search` since it seems like we need to compensate inside the `break` as well, because the point will be different.

Though, the simple fix would be to move the `point = cell.point` like it was before, and do a `return Ok(cell.point)`.

This will be exactly like before when we account for compensation for `Ok` you have.

alacritty_terminal/src/term/search.rs
```diff
@@ -450,41 +468,41 @@ impl<T> Term<T> {
...
             point = cell.point;
 
+            if !cell.flags.intersects(wide) && needles.contains(cell.c) {
+                return Ok(point);
+            }
+
             if point.column == last_column && !cell.flags.contains(Flags::WRAPLINE) {
-                break; // cut off if on new line or hit escape char
+                break;
             }
```
kchibisov, line 502: The same as above, you can make like it was before and do `return Ok(cell.point)` and then doing an advance. Though, in this case it doesn't really matter since we do it before `break`....

kchibisov (changes requested):
The man pages are feeling lonely.

nixpulvis:
Still getting the dreaded "thread 'PTY reader' panicked", so I'll have to merge the updates in winit before I can get around to more on this.

If v0.13 lands before this, then ignore this comment.

alacritty_terminal/src/term/search.rs
chrisduerr: I've intentionally changed it though, otherwise I would have kept it the same? Can you actually provide an example where this introduces a regression? Because my testing didn't show any changes.

kchibisov: I haven't found it myself, because only one part about last column is different, but last column means that we reached end anyway. The other code suggests that as well though.

PR #7234, opened 2023-09-22
Fix regex not matching the first column

kchibisov:
Fixes 73276b62 (Update regex-automata to v0.3.6)


chrisduerr:
https://github.com/alacritty/alacritty/pull/7264

PR #7244, opened 2023-09-28
Add support for OSC 22 escape sequence

### LINKED ISSUE #7119: Add support for OSC 22
This allows setting a cursor shape, the names are X11 names which should be just
passed as is to winit. We may want to have an option to disable it.

The use case is to have a better mouse mode support in some applications.

The escape is supported by: foot, xterm. The docs are on https://invisible-island.net/xterm/ctlseqs/ctlseqs.html . If the shape doesn't exist we can ignore the event.

kchibisov:
Add support for OSC 22 escape sequence to allow changing mouse cursor shapes. While the escape sequence could help some users with the mouse UX in TUI applications using mouse mode, it could obscure common mouse mode indication which is the `Text` cursor. Thus OSC 22 is disabled by default and could be enabled with `terminal.osc22` configuration option.

Fixes #7119.

--

I'm not sure from where to `re-export` the `CursorIcon` inside the `alacritty_terminal`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Bindings to create and navigate tabs on macOS
 - Support startup notify protocol to raise initial window on Wayland/X11
 - Debug option `prefer_egl` to prioritize EGL over other display APIs
+- Escape sequence to change mouse cursor shape (OSC 22);
+    use `terminal.osc22` to enable it
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


rockorager:
I created an issue on kitty for support and Kovid ended up writing a pretty nice spec for this. It expands on the xterm implementation by adding a query for supported names, as well as a stack based approach if desired. Also fleshes out how it should work with other terminal features

https://github.com/kovidgoyal/kitty/blob/master/docs/pointer-shapes.rst

chrisduerr:
@perfbot (just testing things)

PR #7264, opened 2023-10-03
Fix regex matches ending on multiline

chrisduerr:
This fixes an issue where the reverse search for the regex start would
truncate a character when ending on a newline, since it was omitting the
EOI check in that case.

This also fixes a separate issue which caused regexes which capture
empty strings (e.g.: `.*`) to always report a match.

This is a regression introduced in 73276b6.

### DIFF
--- a/alacritty_terminal/src/term/search.rs
+++ b/alacritty_terminal/src/term/search.rs
@@ -47,7 +47,7 @@ impl RegexSearch {
         let right_fdfa = LazyDfa::new(
             search,
             config.clone(),
-            syntax_config.clone(),
+            syntax_config,
             thompson_config.clone(),
             Direction::Right,
             false,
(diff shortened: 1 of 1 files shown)


kchibisov:
This looks much better, however I still noticed a few things different or not ideal:

- `[a-z]*` or any sequence with `*` works as the one with `+`. I find it a bit confusing, at least;
- the labels for hints are still very off with some regexes containing the `|`.

The following regex stopped working, while it worked before:

`regex = "(([\\-\\._a-z0-9]*/[^/ ]*)+/?)|((ipfs:|ipns:|magnet:|mailto:|gemini:|gopher:|https:|http:|news:|file:|git:|ssh:|ftp:)[^\u0000-\u001F\u007F-<>\"\\s{-}\\^⟨⟩`]+)"`

The example is `https://github.com, it'll highlight the `/github`, but before it was highlighting the longest match, so `https://github.com`. It's getting fixed if you swap the order, is it could be not a big deal in the end. Though, still weird.

kchibisov (changes requested):
See above.

kchibisov (approved):

PR #7277, opened 2023-10-08
Use rustix-openpty instead of nix

kchibisov:
This should remove our dependency on nix which tends to do breaking changes pretty often resulting in lots of their deps being in the tree.

The upstream libraries we use also move to rustix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -100,11 +100,11 @@ dependencies = [
  "libc",
  "log",
  "miow",
- "nix 0.27.1",
  "parking_lot",
  "piper",
  "polling",
  "regex-automata",
+ "rustix-openpty",
  "serde",
  "serde_json",
  "serde_yaml",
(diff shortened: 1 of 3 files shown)


alacritty/src/window_context.rs
```diff
@@ -191,7 +191,7 @@ impl WindowContext {
 
         // Create the PTY.
         //
-        // The PTY forks a process to run the shell on the slave side of the
+        // The PTY forks a process to run the shell on the user side of the
```
chrisduerr, line 194: ?

alacritty_terminal/src/event_loop.rs
```diff
@@ -275,11 +275,10 @@ where
...
                                 {
-                                    // On Linux, a `read` on the master side of a PTY can fail
-                                    // with `EIO` if the client side hangs up.  In that case,
-                                    // just loop back round for the inevitable `Exited` event.
-                                    // This sucks, but checking the process is either racy or
-                                    // blocking.
+                                    // On Linux, a `read` on the controller side of a PTY can fail
+                                    // with `EIO` if the client side hangs up.  In that case, just
+                                    // loop back round for the inevitable `Exited` event. This
+                                    // sucks, but checking the process is either racy or blocking.
```
chrisduerr, line 281: Please change this shit back.

chrisduerr (changes requested):

chrisduerr (approved):

PR #7296, opened 2023-10-15
Fix `window.decorations_theme_variant` reload

### LINKED ISSUE #7295: Reload window decorations when config changes
When `decorations_theme_variant` is set to `None` Alacritty auto picks the system theme variant. But if the user changes the system theme from dark to light or vice versa and changes Alacritty's color scheme to match it (i.e. via a shell script), the new config will be reloaded correctly, but the window decorations will remain unchanged and the only alternative is to close and reopen Alacritty.

[alacritty_decoration.webm](https://github.com/alacritty/alacritty/assets/39863381/9293193f-9be9-47d9-a8d3-3727c31f9247)

### Steps

- Make sure `decorations_theme_variant` is set to `None`
- Use Wayland client side decorations
- Switch between light and dark themes (you can use `gsettings set org.gnome.desktop.interface color-scheme "default"` and `gsettings set org.gnome.desktop.interface color-scheme "prefer-dark"`) and change anything in Alacritty config just to force reload

### Expected behavior

Alacritty checks if the system theme variant has changed and reloads the window decorations when reloading the config; or just reload the window decorations along with the rest of the configuration

### System

OS: Linux
Version: alacritty 0.12.3 (5efb0690)
Linux/BSD: GNOME 45 Wayland

kchibisov:
The live reload handling wasn't introduced when the option got added.

Fixes #7295.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Scrolling on touchscreens
 - Double clicking on CSD titlebar not always maximizing a window on Wayland
 - Excessive memory usage when using regexes with a large number of possible states
+- `window.decorations_theme_variant` not live reloading
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


chrisduerr (approved):
Ah, I thought #7295 was about the gsettings update not being updated automatically. LGTM

brunolpsousa:
Wow, that was quick 😁

I can confirm that this has fixed #7295 for me.

Thank you very much!

PR #7220, opened 2023-09-16
Update winit to 0.29.2

kchibisov:
Fixes #7236.
Fixes #7201.
Fixes #6848.
Fixes #3601.
Fixes #3108.
Fixes #2453.

PR #7297, opened 2023-10-15
Fix display freeze on Windows during bell animation

### LINKED ISSUE #7236: Bell causes window rendering updates to stop on Windows
When I do something in WSL that causes a terminal bell, the Alacritty window ceases to update until I move the window to another monitor, resize it, etc.

I bisected the issue to this commit: https://github.com/alacritty/alacritty/commit/a189861880aafcce760e3a142bc2d391f2801f64

### System

OS: Windows
Version: master

kchibisov:
This also fixes an issue of `RedrawRequested` not being throttled for regular events when timer was in use.

Measuring time between `RedrawRequsted` on Windows also showed that events are not throttled, thus we still need our timer.

Fixes #7236.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1481,6 +1481,9 @@ impl Processor {
                 return;
             }
 
+            // NOTE when moving event to be handled without batching, the `dirty` must be changed
+            // to `display.window.request_redraw`, otherwise we could have a situation with an
+            // empty queue, but dirty bit being set.
             match event {
                 // The event loop just got initialized. Create a window.
                 WinitEvent::Resumed => {
@@ -1512,9 +1515,7 @@ impl Processor {
                 }) => {
                     if let Some(window_context) = self.windows.get_mut(&window_id) {
                         window_context.dirty = true;
-                        if window_context.display.window.has_frame {
-                            window_context.display.window.request_redraw();
-                        }
+                        window_context.display.window.request_redraw();
                     }
                 },
                 // NOTE: This event bypasses batching to minimize input latency.
@@ -1567,7 +1568,7 @@ impl Processor {
                         WinitEvent::RedrawRequested(window_id),
                     );
 
-                    window_context.draw(&mut scheduler);
+                    window_context.draw(&proxy, &mut scheduler);
                 },
                 // Process all pending events.
                 WinitEvent::AboutToWait => {
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -99,6 +99,7 @@ pub enum EventType {
     BlinkCursor,
     BlinkCursorTimeout,
     SearchNext,
+    Redraw,
     Frame,
 }
 
@@ -1254,6 +1255,7 @@ impl input::Processor<EventProxy, ActionContext<'_, Notifier, EventProxy>> {
                 EventType::Message(_)
                 | EventType::ConfigReload(_)
                 | EventType::CreateWindow(_)
+                | EventType::Redraw
                 | EventType::Frame => (),
             },
             WinitEvent::WindowEvent { event, .. } => {
@@ -1505,8 +1507,12 @@ impl Processor {
 
                     info!("Initialisation complete");
                 },
-                // NOTE: This event bypasses batching to minimize input latency.
+                // NOTE: These events bypasses batching to minimize input latency.
                 WinitEvent::UserEvent(Event {
+                    window_id: Some(window_id),
+                    payload: EventType::Redraw,
+                })
+                | WinitEvent::UserEvent(Event {
                     window_id: Some(window_id),
                     payload: EventType::Terminal(TerminalEvent::Wakeup),
                 }) => {
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1481,9 +1481,6 @@ impl Processor {
                 return;
             }
 
-            // NOTE when moving event to be handled without batching, the `dirty` must be changed
-            // to `display.window.request_redraw`, otherwise we could have a situation with an
-            // empty queue, but dirty bit being set.
             match event {
                 // The event loop just got initialized. Create a window.
                 WinitEvent::Resumed => {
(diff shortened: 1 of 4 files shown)


alacritty/src/event.rs
```diff
@@ -1481,6 +1481,9 @@ impl Processor {
                 return;
             }
 
+            // NOTE when moving event to be handled without batching, the `dirty` must be changed
+            // to `display.window.request_redraw`, otherwise we could have a situation with an
+            // empty queue, but dirty bit being set.
```
kchibisov, line 1486: I'm not sure whether I should comment like that, or add a special `catch-all` in `AboutToWait` when queue is empty, but we have a dirty bit set. Probably better to
have a catch-all in `AboutToWait`?

alacritty/src/window_context.rs
```diff
@@ -487,10 +491,7 @@ impl WindowContext {
...
         }
 
-        // Request a redraw.
-        //
-        // Even though redraw requests are squashed in winit, we try not to
-        // request more if we haven't received a new frame request yet.
+        // We draw from `RedrawRequested`, so no need to request it, since we clear dirty bit.
         if self.dirty && !self.occluded && !matches!(event, WinitEvent::RedrawRequested(_)) {
             self.display.window.request_redraw();
         }
```
chrisduerr, line 497: The comment says we don't need to request it, but the code requests it. There's also no dirty bit clearing going on here, instead the dirty bit is being checked.

To me the two seem to have nothing to do with each other.

chrisduerr (changes requested):

alacritty/src/window_context.rs
```diff
@@ -391,7 +391,11 @@ impl WindowContext {
 
         // Request immediate re-draw if visual bell animation is not finished yet.
         if !self.display.visual_bell.completed() {
-            self.display.window.request_redraw();
+            let window_id = Some(self.display.window.id());
+            // Winit is not reliable on at least Windows when asking for `Redraw` during the
+            // `RedrawRequsted`.
```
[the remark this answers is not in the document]
kchibisov: @chrisduerr it's a windows itself, as I linked. Winit will need to change how it works entirely and not rely on `WM_PAINT` for drawing.

>Also we shouldn't just be abusing other events just because they might work. If we need our own event, we should define it properly.

They not **might**, they work. Because they are user events. Windows itself preventing winit from requesting a redraw without any signs.

alacritty/src/window_context.rs
kchibisov: The comment try to explain why we have `!matches`.

alacritty/src/window_context.rs
```diff
@@ -415,7 +420,21 @@ impl WindowContext {
         event: WinitEvent<Event>,
     ) {
         match event {
-            WinitEvent::AboutToWait | WinitEvent::RedrawRequested(_) => {
+            WinitEvent::AboutToWait => {
+                // Skip further event handling with no staged updates.
+                if self.event_queue.is_empty() {
+                    // Request a redraw in-case we have dirty flag set without events in the queue.
+                    // This could happen when event excluded from batching set `dirty` flag, but
+                    // forgot to call `redraw_requested`.
```
kchibisov, line 428: That's the handling for the missed `redraw_requested` I had a `NOTE` before. I don't really like it, but it protects against not asking for redraw at all.

The issue is because when we handle events without batching, we'll have an `empty` queue here, which means that we won't ask for the redraw event at all.

alacritty/src/window_context.rs
```diff
@@ -391,7 +392,11 @@ impl WindowContext {
 
         // Request immediate re-draw if visual bell animation is not finished yet.
         if !self.display.visual_bell.completed() {
-            self.display.window.request_redraw();
+            let window_id = Some(self.display.window.id());
+            // Winit on Windows discards `redraw_requested` calls from the `RedrawRequested` event,
+            // thus use a `Wakeup` which is delivered reliably and also sets the `dirty` flag.
```
kchibisov, line 397: I hope this is a bit more clear. I've updated winit docs to say that it doesn't work though. And yes, it's a Windows behavior, which will take time to fix in winit, since it should use something else to schedule redraws.

The bug is not new, it's just new for us, since we now draw on `RedrawRequsted`.

alacritty/src/window_context.rs
```diff
@@ -391,7 +392,11 @@ impl WindowContext {
...
+            let _ = event_proxy
+                .send_event(Event::new(EventType::Terminal(TerminalEvent::Wakeup), window_id));
```
chrisduerr, line 399: Using terminal events for windowing isn't an option imo.

alacritty/src/window_context.rs
chrisduerr, line 428: This just sounds like a workaround. Either the event which "forgot" to call `redraw_requested` should be fixed or the comment needs rewording. Right now it just sounds redundant unless other code is buggy.

alacritty/src/window_context.rs
```diff
@@ -487,10 +506,8 @@ impl WindowContext {
             self.mouse.hint_highlight_dirty = false;
         }
 
-        // Request a redraw.
-        //
-        // Even though redraw requests are squashed in winit, we try not to
-        // request more if we haven't received a new frame request yet.
+        // Don't call `request_redraw` when event is `RedrawRequested` since the `dirty` flag
+        // represents the current frame, but redraw is for the `next` frame.
```
chrisduerr, line 510:
```suggestion
        // represents the current frame, but redraw is for the next frame.
```

chrisduerr (changes requested):

alacritty/src/window_context.rs
kchibisov: What's your suggestion then? To add one more event telling to refresh the display?

alacritty/src/window_context.rs
kchibisov: I mean, I could just remove this.

kchibisov:
Handled in https://github.com/alacritty/alacritty/pull/7220 due to upstream resolution of the issue.

PR #7300, opened 2023-10-17
Prefer exact matches for bindings in mouse mode

### LINKED ISSUE #7292: Text pasted twice on Shift-MiddleButton in mouse tracking apps (vim, tmux)
### System

OS: MacOS 14.0
Version: current `master` branch, current `update-winit-final` branch

### Configuration

```
[mouse.bindings]]
mouse = "Middle"
mods = "Shift"
action = "Paste"
```

### Logs

Events logged with `--print-events` and the current `update-winit-final` branch (dfbb01d6b17c112be34ed7ebebf0227647919d80).

```
*** Shift is already held, mouse wheel is pressed
[17.065938291s] [INFO ] [alacritty] winit event: NewEvents(WaitCancelled { start: Instant { tv_sec: 18849, tv_nsec: 810943958 }, requested_resume: None })
[17.067660666s] [INFO ] [alacritty] winit event: DeviceEvent { device_id: DeviceId(DeviceId), event: Button { button: 2, state: Pressed } }
[17.067787416s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(5191806272)), event: CursorMoved { device_id: DeviceId(DeviceId), position: PhysicalPosition { x: 434.0, y: 22.0 } } }
[17.067893041s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(5191806272)), event: MouseInput { device_id: DeviceId(DeviceId), state: Pressed, button: Middle } }
[17.067969541s] [INFO ] [alacritty] winit event: AboutToWait
*** Output from an added debug statement in the beginning of execute()
[alacritty/src/input.rs:162] &self = Paste
[alacritty/src/input.rs:162] &self = PasteSelection
[17.074582958s] [INFO ] [alacritty] winit event: NewEvents(WaitCancelled { start: Instant { tv_sec: 18854, tv_nsec: 874886916 }, requested_resume: None })
[17.074667666s] [INFO ] [alacritty] winit event: UserEvent(Event { window_id: Some(WindowId(WindowId(5191806272))), payload: Terminal(Wakeup) })
[17.074735083s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(5191806272)), event: RedrawRequested }
[17.082676833s] [INFO ] [alacritty] winit event: AboutToWait
[17.082799583s] [INFO ] [alacritty] winit event: NewEvents(WaitCancelled { start: Instant { tv_sec: 18854, tv_nsec: 884327708 }, requested_resume: None })
[17.082853083s] [INFO ] [alacritty] winit event: UserEvent(Event { window_id: Some(WindowId(WindowId(5191806272))), payload: Terminal(Wakeup) })
[17.082888125s] [INFO ] [alacritty] winit event: UserEvent(Event { window_id: Some(WindowId(WindowId(5191806272))), payload: Frame })
[17.082927416s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(5191806272)), event: RedrawRequested }
[17.088560208s] [INFO ] [alacritty] winit event: AboutToWait
[17.088627791s] [INFO ] [alacritty] winit event: NewEvents(WaitCancelled { start: Instant { tv_sec: 18854, tv_nsec: 890160041 }, requested_resume: None })
[17.088665208s] [INFO ] [alacritty] winit event: UserEvent(Event { window_id: Some(WindowId(WindowId(5191806272))), payload: Frame })
[17.088693750s] [INFO ] [alacritty] winit event: AboutToWait
*** Releasing mouse wheel, still holding Shift
[19.807858583s] [INFO ] [alacritty] winit event: NewEvents(WaitCancelled { start: Instant { tv_sec: 18854, tv_nsec: 890260333 }, requested_resume: None })
[19.809546500s] [INFO ] [alacritty] winit event: DeviceEvent { device_id: DeviceId(DeviceId), event: Button { button: 2, state: Released } }
[19.809694333s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(5191806272)), event: CursorMoved { device_id: DeviceId(DeviceId), position: PhysicalPosition { x: 434.0, y: 22.0 } } }
[19.809806791s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(5191806272)), event: MouseInput { device_id: DeviceId(DeviceId), state: Released, button: Middle } }
[19.809885750s] [INFO ] [alacritty] winit event: AboutToWait
```

### Context

I'm used to holding Shift when cutting and pasting, as I spend a lot of time in applications that track mouse events. However, Shift-MiddleButton doesn't past when the mouse is not being tracked (e.g. shell without tmux). So I tried adding a binding for Shift-MiddleButton (see config above). Now the bare shell is working as expected, but text is pasted twice in tmux and neovim.

As you can see, two actions happen when the mouse wheel is clicked, `Paste` and `PasteSelection`. That appears to be issue to me. An event should not trigger two actions.

With an empty config file, only one action is printed by `dbg!(&self);` in the beginning of `execute()` in `alacritty/src/input.rs`, that's `PasteSelection`.

proski:
Only consider bindings without Shift if there are no actions defined for the actual mouse event.

Fixes #7292

### DIFF
--- a/alacritty/src/input.rs
+++ b/alacritty/src/input.rs
@@ -1003,17 +1003,24 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
         let mode = BindingMode::new(self.ctx.terminal().mode(), self.ctx.search_active());
         let mouse_mode = self.ctx.mouse_mode();
         let mods = self.ctx.modifiers().state();
+        let mouse_bindings = self.ctx.config().mouse_bindings().to_owned();
 
-        for i in 0..self.ctx.config().mouse_bindings().len() {
-            let mut binding = self.ctx.config().mouse_bindings()[i].clone();
+        // If mouse mode is active, also look for bindings without shift.
+        let mut check_fallback = mouse_mode && mods.contains(ModifiersState::SHIFT);
 
-            // Require shift for all modifiers when mouse mode is active.
-            if mouse_mode {
-                binding.mods |= ModifiersState::SHIFT;
-            }
-
+        for binding in &mouse_bindings {
             if binding.is_triggered_by(mode, mods, &button) {
                 binding.action.execute(&mut self.ctx);
+                check_fallback = false;
+            }
+        }
+
+        if check_fallback {
+            let fallback_mods = mods & !ModifiersState::SHIFT;
+            for binding in &mouse_bindings {
+                if binding.is_triggered_by(mode, fallback_mods, &button) {
+                    binding.action.execute(&mut self.ctx);
+                }
             }
         }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Apply `colors.transparent_background_colors` for selections, hints, and search matches
 - Underline full hint during keyboard selection
 - Synchronized updates now use `CSI 2026` instead of legacy `DCS` variant
+- In mouse mode with `Shift` pressed, mouse bindings without `Shift` are only triggered
+    if no exact binding (i.e. one with `Shift`) is found.
 
 ### Fixed
 
--- a/extra/man/alacritty.5.scd
+++ b/extra/man/alacritty.5.scd
@@ -583,9 +583,10 @@ This section documents the *[mouse]* table of the configuration file.
 	See _keyboard.bindings_ for full documentation on _mods_, _mode_, _action_,
 	and _chars_.
 
-	To trigger mouse bindings when an application running within Alacritty
-	captures the mouse, the `Shift` modifier is automatically added as a
-	requirement.
+	When an application running within Alacritty captures the mouse, the `Shift`
+	modifier can be used to suppress mouse reporting. If no action is found for
+	the event, actions for the event without the `Shift` modifier are triggered
+	instead.
 
 	*mouse* "Middle" | "Left" | "Right" | "Back" | "Forward" | <number>
 


proski:
The implementation follows the suggestion I've got in #7292. The actions defined for the actual mouse event (with Shift modifier if Shift is actually pressed) take priority. The actions for the bindings with Shift removed are only considered is there are no actions for the actual event.

I tried to make implementation as simple as the original. `clone()` is only called on the actions that match the event, not on every binding. The `binding` variable is immutable now. It iterates over the bindings naturally, without access by index. The modification is made to the actual event modifiers, not to the bindings, which makes the logic more transparent (Shift is removed from the event, not added to the binding).

Once there is any exact match, fallback matches are not populated anymore.

proski:
The code is working well for me, I'm using it as my normal terminal. Both middle-click and Shift-middle-click are pasting exactly once everywhere. I had to disable paste on middle-click in vim to prevent double paste without Shift. All other mouse operations are working as expected.

An alternative approach is to put actions to a set, which would remove duplicate actions. That would require hash support on actions and some other types. I don't think anyone would want one action to run multiple times for an event. However, `Paste` and `PasteSelection` are different events, so some users can end up with both being emitted for one event.

alacritty/src/input.rs
```diff
@@ -984,18 +984,28 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+                }
+                actions.push(binding.action.clone());
             }
+            if use_fallback && binding.is_triggered_by(mode, fallback_mods, &button) {
+                actions.push(binding.action.clone());
+            }
+        }
+
+        for action in actions {
+            action.execute(&mut self.ctx);
```
kchibisov, line 1008: Could you simply take the code which was there before and do 2 loops? The amount of bindings is very low so it's better than `vec` approach.

Just do normal loop without `mouse_mode` part and set some flag to `true` when you actually triggered some binding.

And then, `if mouse_mode && !sent_binding` do the ho the same loop, but with `SHIFT` applied? First loop won't need to `clone` even, I think.

proski:
I've implemented 2 loops as you requested. Unfortunately, I have to use the same iteration by index and cloning in both loops, and I cannot find a way to avoid it without executing actions outside the loops, which would need a vector or some other data structure.

proski:
I was able to get rid of indices and `clone()` by using `to_owned()` for `mouse_bindings`. That just one memory allocation per function call, as opposed to allocations for every binding on `clone()`.

alacritty/src/input.rs
```diff
@@ -984,19 +984,26 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
 
-        for i in 0..self.ctx.config().mouse_bindings().len() {
-            let mut binding = self.ctx.config().mouse_bindings()[i].clone();
-
-            // Require shift for all modifiers when mouse mode is active.
-            if mouse_mode {
-                binding.mods |= ModifiersState::SHIFT;
-            }
+        // If mouse mode is active, also look for bindings without Shift
+        let mut fallback_needed = mouse_mode && mods.contains(ModifiersState::SHIFT);
```
chrisduerr, line 990:
```suggestion
        let mut check_fallback = mouse_mode && mods.contains(ModifiersState::SHIFT);
```

alacritty/src/input.rs
```diff
@@ -984,19 +984,26 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
 
+        for binding in &mouse_bindings {
             if binding.is_triggered_by(mode, mods, &button) {
+                fallback_needed = false;
                 binding.action.execute(&mut self.ctx);
```
chrisduerr, line 995:
```suggestion
                binding.action.execute(&mut self.ctx);
                fallback_needed = false;
```

chrisduerr (changes requested):
Missing a changelog entry.

proski:
Addressed the comments.

chrisduerr:
If you want another review, please request it.

kchibisov (approved):

alacritty/src/input.rs
```diff
@@ -1003,17 +1003,24 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+        let mouse_bindings = self.ctx.config().mouse_bindings().to_owned();
 
-        for i in 0..self.ctx.config().mouse_bindings().len() {
-            let mut binding = self.ctx.config().mouse_bindings()[i].clone();
-
-            // Require shift for all modifiers when mouse mode is active.
-            if mouse_mode {
-                binding.mods |= ModifiersState::SHIFT;
-            }
+        // If mouse mode is active, also look for bindings without Shift
```
chrisduerr, line 1008:
```suggestion
        // If mouse mode is active, also look for bindings without shift.
```

chrisduerr (changes requested):

alacritty/src/input.rs
proski: Done

chrisduerr (approved):
Thanks for the PR.

PR #7321, opened 2023-10-25
Fix warnings about unused imports

proski:
`pub use crate::config::mouse::Mouse` is not needed even for tests.
`pub use crate::config::bindings::Binding` is only needed for tests.

Tested in 8 combinations:
```
cargo {+stable,+nightly} clippy {,--all-targets} {,--release}
```

### DIFF
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -25,11 +25,11 @@ mod bindings;
 mod mouse;
 
 use crate::cli::Options;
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
There's no reason to open multiple different PRs for all these tiny style changes. It just creates unnecessary noise.

proski:
OK, I'll make a single PR.

PR #7322, opened 2023-10-26
Fix clippy warnings

proski:
Tested in 8 combinations:
```
cargo {+stable,+nightly} clippy {,--all-targets} {,--release}
```

* Fixed warnings about unused imports. One import is only used for tests, another import is not needed at all.
* Replaced `get(0)` with `first()` in tests as suggested by nightly clippy.

### DIFF
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -25,11 +25,11 @@ mod bindings;
 mod mouse;
 
 use crate::cli::Options;
-pub use crate::config::bindings::{
-    Action, Binding, BindingKey, BindingMode, MouseAction, SearchAction, ViAction,
-};
 #[cfg(test)]
-pub use crate::config::mouse::Mouse;
+pub use crate::config::bindings::Binding;
+pub use crate::config::bindings::{
+    Action, BindingKey, BindingMode, MouseAction, SearchAction, ViAction,
+};
 pub use crate::config::ui_config::UiConfig;
 
 /// Maximum number of depth for the configuration file imports.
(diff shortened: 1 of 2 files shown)


alacritty/src/string.rs
```diff
@@ -50,8 +50,9 @@ impl<'a> StrShortener<'a> {
         }
 
         if direction == ShortenDirection::Right {
+            let skip_chars = 0; // suppress clippy warning about `skip(0)` in Rust 1.73+
             return Self {
-                chars: text.chars().skip(0),
+                chars: text.chars().skip(skip_chars),
```
chrisduerr, line 55: This is pretty garbage, why not just add a clippy exception?

chrisduerr (changes requested):

alacritty/src/string.rs
proski: That warning is showing with clippy from Rust 1.73.0 (the current stable). We are supporting Rust 1.70.0 and newer. Clippy prior to 1.73.0 doesn't understand `#[allow(clippy::iter_skip_zero)]`. So I'm also allowing `unknown_lint` and adding a comment why it's needed.

alacritty/src/string.rs
```diff
@@ -50,6 +50,8 @@ impl<'a> StrShortener<'a> {
         }
 
         if direction == ShortenDirection::Right {
+            // clippy::iter_skip_zero appears in Rust 1.73, ignore it for earlier versions.
+            #[allow(unknown_lints, clippy::iter_skip_zero)]
```
chrisduerr, line 54: Oldstable clippy is the only thing that matters. If there's an issue fixing before oldstable starts complaining that's fine, but exceptions shouldn't be added before oldstable requires it.

chrisduerr (changes requested):

alacritty/src/string.rs
proski: OK, I'm removing that change. I'll try to find an acceptable solution, but I don't see any good approach at the moment.

chrisduerr (approved):

PR #7332, opened 2023-10-29
Bump glutin to 0.31.1

kchibisov:
This fixes a crash on startup with macOS Sonoma.

--

Crash is not present with glutin 0.30.x series.


chrisduerr (approved):
I'd assume changelog entry is irrelevant because the bug doesn't exist on stable.

PR #7337, opened 2023-11-02
Report errors during OpenGL initialization

proski:
Alacritty crashes on Windows in a VM (eg. in virtsh on Linux or in UTM on MacOS) due to shader version reported as NULL. Alacritty shows absolutely no information about the crash, even in the verbose mode. This PR makes two improvements:

* Instead of a crash, an error is returned to the caller.
* The Windows build of Alacritty shows the error in a message box and on the the parent console.

### DIFF
--- a/alacritty/src/renderer/mod.rs
+++ b/alacritty/src/renderer/mod.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::HashSet;
 use std::ffi::{CStr, CString};
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -47,12 +48,16 @@ pub static GL_FUNS_LOADED: AtomicBool = AtomicBool::new(false);
 pub enum Error {
     /// Shader error.
     Shader(ShaderError),
+
+    /// Other error.
+    Other(String),
 }
 
 impl std::error::Error for Error {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         match self {
             Error::Shader(err) => err.source(),
+            Error::Other(_) => None,
         }
     }
 }
@@ -63,6 +68,9 @@ impl fmt::Display for Error {
             Error::Shader(err) => {
                 write!(f, "There was an error initializing the shaders: {}", err)
             },
+            Error::Other(err) => {
+                write!(f, "{}", err)
+            },
         }
     }
 }
@@ -70,6 +78,12 @@ impl fmt::Display for Error {
 impl From<ShaderError> for Error {
     fn from(val: ShaderError) -> Self {
         Error::Shader(val)
+    }
+}
+
+impl From<String> for Error {
+    fn from(val: String) -> Self {
+        Error::Other(val)
     }
 }
 
@@ -83,6 +97,25 @@ enum TextRendererProvider {
 pub struct Renderer {
     text_renderer: TextRendererProvider,
     rect_renderer: RectRenderer,
+}
+
+/// Wrapper around gl::GetString with error checking and reporting.
+fn gl_get_string(
+    string_id: gl::types::GLenum,
+    description: &str,
+) -> Result<Cow<'static, str>, Error> {
+    unsafe {
+        let string_ptr = gl::GetString(string_id);
+        match gl::GetError() {
+            gl::NO_ERROR if !string_ptr.is_null() => {
+                Ok(CStr::from_ptr(string_ptr as *const _).to_string_lossy())
+            },
+            gl::INVALID_ENUM => {
+                Err(format!("OpenGL error requesting {}: invalid enum", description).into())
+            },
+            error_id => Err(format!("OpenGL error {} requesting {}", error_id, description).into()),
+        }
+    }
 }
 
 impl Renderer {
@@ -104,14 +137,9 @@ impl Renderer {
             });
         }
 
-        let (shader_version, gl_version, renderer) = unsafe {
-            let renderer = CStr::from_ptr(gl::GetString(gl::RENDERER) as *mut _).to_string_lossy();
-            let shader_version =
-                CStr::from_ptr(gl::GetString(gl::SHADING_LANGUAGE_VERSION) as *mut _)
-                    .to_string_lossy();
-            let gl_version = CStr::from_ptr(gl::GetString(gl::VERSION) as *mut _).to_string_lossy();
-            (shader_version, gl_version, renderer)
-        };
+        let shader_version = gl_get_string(gl::SHADING_LANGUAGE_VERSION, "shader version")?;
+        let gl_version = gl_get_string(gl::VERSION, "OpenGL version")?;
+        let renderer = gl_get_string(gl::RENDERER, "renderer version")?;
 
         info!("Running on {renderer}");
         info!("OpenGL version {gl_version}, shader_version {shader_version}");


alacritty/src/renderer/mod.rs
```diff
@@ -105,12 +118,17 @@ impl Renderer {
...
+            let shader_version = gl::GetString(gl::SHADING_LANGUAGE_VERSION);
+            let gl_version = gl::GetString(gl::VERSION);
+            let renderer = gl::GetString(gl::RENDERER);
+            if shader_version.is_null() || gl_version.is_null() || renderer.is_null() {
+                return Err("OpenGL support is missing, cannot continue".into());
+            }
+            (
+                CStr::from_ptr(shader_version as *mut _).to_string_lossy(),
+                CStr::from_ptr(gl_version as *mut _).to_string_lossy(),
+                CStr::from_ptr(renderer as *mut _).to_string_lossy(),
```
kchibisov, line 130: Can't we use `glGetError`, I think it should work?

kchibisov (changes requested):

alacritty/src/renderer/mod.rs
proski: `gl::GetError()` returns 0 at all times. `shader_version` is NULL, so it crashes in `to_string_lossy()`. `gl_version` is `1.1.0`, `renderer` is `GDI Generic`.

kchibisov: I guess the driver is just straight broken, since `glGetString` returns `zero` when the `enum` is not valid, but enum variant is sort of valid, though, I guess not for the 1.1 spec, which we don't support. But still `glGetError` probably should return. though...

kchibisov: Also, at which point you've called `glGetError` right after the calling `shader_version`, since otherwise it won't work, I think.

kchibisov (approved):

kchibisov:
it's not like any of that could happen with any sane driver, but only with something that is below 2.0, which is basically windows in a vm.

kchibisov:
Alacritty can run in a vm on windows with ANGLE btw, you just need it and force EGL with debug option, probably will run on some software DX renderer.

alacritty/src/main.rs
```diff
@@ -205,7 +205,14 @@ fn alacritty(options: Options) -> Result<(), Box<dyn Error>> {
     // FIXME patch notify library to have a shutdown method.
     // config_reloader.join().ok();
 
-    // Without explicitly detaching the console cmd won't redraw it's prompt.
+    // Panic on error on Windows, that way the error message is shown in a message box.
+    // It is done before FreeConsole() so the message appears on the console as well.
+    #[cfg(windows)]
+    if let Err(err) = result {
+        panic!("{}", err);
+    }
```
chrisduerr, line 213: I don't like the idea of sidestepping error handling to cause an explicit panic, doesn't really make sense.

Especially not in here, considering there's other places this function could error out from. If anything, our error handling should be consistent?

chrisduerr (changes requested):

alacritty/src/main.rs
kchibisov: there's only one place and it's an event loop builder, but I think message box won't show if you don't build event loop on windows. Not sure about that though.

It's also tricky that you need to call it before `FreeConsole`, otherwise you won't log into pty, so I don't see other place you'd through it on windows.

The good thing about side stepping panic like that, that it's visible for the users so they can report something at least.

chrisduerr: That just sounds like the `FreeConsole` code should be reworked.

alacritty/src/renderer/mod.rs
proski: My mistake, `gl::GetError` returns 1280 (`GL_INVALID_ENUM`) immediately after requesting `gl::SHADING_LANGUAGE_VERSION`.

alacritty/src/main.rs
proski: I'm open to suggestions. Ideally, all fatal errors from Alacritty should be reported through GUI on all platforms, as we cannot expect users to run Alacritty from another terminal.

alacritty/src/renderer/mod.rs
kchibisov: yeah, so we should propagate it.

proski:
The new revision is using `gl::getError`.

alacritty/src/main.rs
kchibisov: The fatal error could be due to crash of the event loop, so it's effectively not possible to do so.

Only windows always has windows around, but not X11/Wayland, you need to somehow draw that sort of thing. In general, you should send information to crash report service, which is system log, basically, so you'll see crashes inside the `journalctl`, etc, since rust's panic doesn't do that sort of thing.

kchibisov: You should run `FreeConsole` from inside the `fn main`, from the `match` there you should get a `let result` and then do the same panic logic there.

kchibisov: Also, do we really need a `panic`? Can't we call some `Win32` API function with panic message and do a normal return?

proski: I've extracted the message box from the panic handler. The message box is called directly without panic.

I'm not sure about moving `FreeConsole` to `main`, but I'll give it a try later.

alacritty/src/renderer/mod.rs
```diff
@@ -105,12 +118,17 @@ impl Renderer {
...
-            let renderer = CStr::from_ptr(gl::GetString(gl::RENDERER) as *mut _).to_string_lossy();
-            let shader_version =
-                CStr::from_ptr(gl::GetString(gl::SHADING_LANGUAGE_VERSION) as *mut _)
-                    .to_string_lossy();
-            let gl_version = CStr::from_ptr(gl::GetString(gl::VERSION) as *mut _).to_string_lossy();
-            (shader_version, gl_version, renderer)
+            let shader_version = gl::GetString(gl::SHADING_LANGUAGE_VERSION);
+            if gl::GetError() != gl::NO_ERROR {
+                return Err("OpenGL shader support is missing, cannot continue".into());
+            }
```
kchibisov, line 124: My point was to sanity check all calls to `gl::GetString` and get an actual error message, because I think gl allows you to format error, like we do in other cases, based on enum variant.

alacritty/src/main.rs
```diff
@@ -205,7 +205,19 @@ fn alacritty(options: Options) -> Result<(), Box<dyn Error>> {
...
+    // If an error has occurred, show it in on the console and in a message box.
+    // It is done before FreeConsole() so the error appears on the console.
+    #[cfg(windows)]
+    if let Err(ref err) = result {
+        eprintln!("Error: {}", err);
+        if let Some(source) = err.source() {
+            eprintln!("Source: {}", source);
+        }
+        panic::error_message_box(&err.to_string(), "An error has occurred");
+    }
```
kchibisov, line 218: The point was to consistently show this message box for all errors encountered, so it should be not here, but in `fn main` where the all errors are being returned, though, I'd assume only for `alacritty` itself we need a message box.

kchibisov (dismissed):

nixpulvis:
Just curious, would it help other platforms to do a similar thing? Should this be done in a more cross-platform manner?

kchibisov:
>Just curious, would it help other platforms to do a similar thing? Should this be done in a more cross-platform manner?

no general API for stuff like that cross platform. The thing is that logs are sort of written into system logger on macOS (Console thing) and on linux we can write into log.

alacritty/src/renderer/mod.rs
proski: I could not find a way to format the error using gl, but I was able to wrap all calls to check for errors and report then with a caller-provided description (which is user-friendlier than a macro would be).

alacritty/src/main.rs
proski: I've removed the GUI and the top-level code from the PR.

alacritty/src/main.rs
proski: I've removed that code.

proski:
I've removed the "upper level" part, as it's inevitably platform specific and I don't have time to get it to a satisfactory condition. The only remaining commit deals with detecting and reporting errors during OpenGL initialization. It's still a significant improvement, as it would prevent crash on systems with deficient OpenGL support.

chrisduerr (approved):

PR #7347, opened 2023-11-07
[docs] add cargo install command in INSTALL.md

melMass:
Small one, but there is no real mention of "installing" per se from source in the install doc.
As the "usual" `cargo install --path .` did not work I think it makes sense to specify there.

Feel free to discard or edit it

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -250,7 +250,9 @@ filling in this section of the README.
 ### Linux / Windows / BSD
 
 ```sh
-cargo build --release
+cargo build --release // will put the bin in target/release/
+// or
+cargo install --path alacritty // will install alacritty globally (in ~/.cargo/bin)
 ```
 
 On Linux/BSD, if it is desired to build Alacritty without support for either the
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -249,13 +249,17 @@ filling in this section of the README.
 
 ### Linux / Windows / BSD
 
+Running `build` will output the binary in `target/release/`
 ```sh
-cargo build --release // will put the bin in target/release/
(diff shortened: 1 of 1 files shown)


melMass:
You might want to limit the CI to not run for md edits 😬  (`paths-ignore: ***/*.md`)

INSTALL.md
```diff
@@ -249,11 +249,17 @@ filling in this section of the README.
...
 cargo build --release
 ```
 
-On Linux/BSD, if it is desired to build Alacritty without support for either the
+Alternatively you can use `install` that will install it globally (in `~/.cargo/bin` by default)
+```
+cargo install --path alacritty 
+```
+
+On Linux/BSD, if it is desired to build (or install) Alacritty without support for either the
```
kchibisov, line 262: I'm not sure about this one though. Since we tell where the binary is, and users can copy it if they want.

Telling to do an `install` like that is a bit wrong, simply because we have `features` and missing them in `install` will restart the build without those features. This command with `install` is also not really correct, because you need to pass `--locked` or `--frozen` for installs in general.

INSTALL.md
```diff
@@ -249,11 +249,17 @@ filling in this section of the README.
 
 ### Linux / Windows / BSD
 
+Running `build` will output the binary in `target/release/`
```
kchibisov, line 252: This is mentioned in the end of the section.

kchibisov:
Thanks. But I don't think that we should add more information than already is to how to use `cargo`, just `cargo build` like we have is more than enough already, and saying where the binary is, is more than enough in my opinion for such way of installing.

PR #7349, opened 2023-11-08
Add `--option` argument to `create-window`

### LINKED ISSUE #6238: `alacritty msg create-window` should support configuration
I've been using `alacritty msg create-window` to manage multiple terminals at once, which works great.
However, it does not support configuration flags, such as `-o=...` or `--config-file ...`.

I would imagine this could be due to technical reasons, as settings like `ipc_socket` or `debug.*` only make sense from an instance, rather than an individual window.

I request that window-specific configuration is implemented in a way that overrides the instance configuration, using the same flags as currently exposed on the main executable.

If an instance-specific configuration flag such as `ipc_socket` or `debug.*` is set to a mismatching value, either on the command line or in a file, when it doesn't make sense to override, alacritty could throw an error.

I'm guessing that configuration would need to be refactored for this change to be implemented, because it's probably not structured per-window at this time, so I don't expect it to be easy to implement.

### System

OS: Arch Linux
Version: `alacritty 0.10.1`
Linux: Wayland, Sway

### Logs

```
$ alacritty msg create-window -o="colors.primary.background='#FFFFFF'"
error: Found argument '-o' which wasn't expected, or isn't valid in this context

USAGE:
    alacritty msg create-window [FLAGS] [OPTIONS]

For more information try --help
```

chrisduerr:
This patch adds a new CLI parameter to the `create-window` subcommand,
matching the existing `--option` parameter when creating a new Alacritty
instance.

This parameter allows setting up the initial window configuration from
the CLI without having to call `alacritty msg config`, making sure that
all options are set appropriately right from the start.

Closes #6238.

---

This is split into two patches, the first one being technically optional. However it hopefully makes understanding these changes a little easier.

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -1,17 +1,20 @@
 use std::cmp::max;
+use std::ops::{Deref, DerefMut};
 use std::path::PathBuf;
+use std::rc::Rc;
 
+use alacritty_config::SerdeReplace;
 use clap::{ArgAction, Args, Parser, Subcommand, ValueHint};
 use log::{self, error, LevelFilter};
 use serde::{Deserialize, Serialize};
 use toml::Value;
 
-use crate::logging::LOG_TARGET_CONFIG;
 use alacritty_terminal::tty::Options as PtyOptions;
 
 use crate::config::ui_config::Program;
 use crate::config::window::{Class, Identity};
 use crate::config::UiConfig;
+use crate::logging::LOG_TARGET_IPC_CONFIG;
 
 /// CLI options for the main Alacritty executable.
 #[derive(Parser, Default, Debug)]
@@ -60,7 +63,7 @@ pub struct Options {
 
     /// CLI options for config overrides.
     #[clap(skip)]
-    pub config_options: Vec<(String, Value)>,
+    pub config_options: ParsedOptions,
 
     /// Options which can be passed via IPC.
     #[clap(flatten)]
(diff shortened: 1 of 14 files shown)


alacritty/src/cli.rs
```diff
@@ -322,13 +325,82 @@ pub struct IpcConfig {
...
+        let mut config_options = Vec::new();
+
+        let logger_started = logging::is_initialized();
+
+        for option in options {
+            let parsed = match toml::from_str(option) {
+                Ok(parsed) => parsed,
+                Err(err) => {
+                    if logger_started {
+                        error!("Ignoring invalid CLI option '{option}': {err}");
```
kchibisov, line 346: I'm not really a fun of logging like that, because it'll be visible for all windows
and there's no clear way to solve this issue. And ti's not like we're preventing the window from being spawned in the first place like that.

Thus, I think it'll be nice to have this warning only for a newly created window.

kchibisov: This is also missing a `target` so there's no way to clear messages by invoking `alacritty msg config` again, so the only way to manually `Ctrl + L`.

alacritty/src/cli.rs
```diff
@@ -301,6 +293,17 @@ pub struct WindowOptions {
     #[cfg(target_os = "macos")]
     /// The window tabbing identifier to use when building a window.
     pub window_tabbing_id: Option<String>,
+
+    /// Override configuration file options [example: cursor.style=Beam].
```
kchibisov, line 297: Something like that? 

```suggestion
    /// Override configuration file options [example: cursor.style="Beam"].
```

kchibisov (changes requested):

kchibisov:
Also, you should update `alacritty-msg(1)`, probably `alacritty(1)` while at it ("Beam") and also say how the override works, and add a `CHANGELOG` entry.

alacritty/src/cli.rs
chrisduerr: > I'm not really a fun of logging like that, because it'll be visible for all windows

But it affects all windows?

> Thus, I think it'll be nice to have this warning only for a newly created window.

So you want it to **only** write to CLI? I'd be okay with that tbh.

chrisduerr:
> Also, you should update alacritty-msg(1), probably alacritty(1) while at it ("Beam") and also say how the override works, and add a CHANGELOG entry.

There were a lot of things wrong with the manpage, so I've tried to fix it all in this patch. Kinda meh but probably worth it.

alacritty/src/cli.rs
kchibisov: >But it affects all windows?

It depends on what happens, since you get this log from `alacritty msg create-window -o font.size=12e`, so it affects only the newly created window, thus only this new window should have this log message.

>So you want it to only write to CLI? I'd be okay with that tbh.

At least in the current state of things.

kchibisov (changes requested):
Looks fine except of CI and that `error!` log without a target.

kchibisov (approved):

PR #7355, opened 2023-11-11
Fix visual bell getting stuck on macOS

### LINKED ISSUE #7325: Visual bell animation on Mac is really long
On macOS, the visual bell animation lasts for around 1 second regardless of what non-zero duration is in the config file. This seems to be the case since commit 80d4daccc307622f8e604af4f36ac89ecccf5db7.

Before that commit, visual bell animation worked fine.

kchibisov:
Fixes #7325.

### DIFF
--- a/alacritty/src/window_context.rs
+++ b/alacritty/src/window_context.rs
@@ -381,7 +381,7 @@ impl WindowContext {
         // Request immediate re-draw if visual bell animation is not finished yet.
         if !self.display.visual_bell.completed() {
             // We can get an OS redraw which bypasses alacritty's frame throttling, thus
-            // mark the window as dirty when we don't have frame yet.
+            // marking the window as dirty when we don't have frame yet.
             if self.display.window.has_frame {
                 self.display.window.request_redraw();
             } else {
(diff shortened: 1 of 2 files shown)


alacritty/src/window_context.rs
```diff
@@ -380,7 +380,13 @@ impl WindowContext {
 
         // Request immediate re-draw if visual bell animation is not finished yet.
         if !self.display.visual_bell.completed() {
-            self.display.window.request_redraw();
+            // We can get an OS redraw which by-passes alacritty's _frame_ throttling, thus
```
chrisduerr, line 383:
```suggestion
            // We can get an OS redraw which bypasses alacritty's frame throttling, thus
```

alacritty/src/window_context.rs
```diff
@@ -380,7 +380,13 @@ impl WindowContext {
...
+            // mark the window as dirty when we don't have frame yet.
```
chrisduerr, line 384:
```suggestion
            // marking the window as dirty when we don't have frame yet.
```

chrisduerr (changes requested):

chrisduerr (approved):

PR #7358, opened 2023-11-12
Fix Vi cursor not being dirty when scrolling

kchibisov:


alacritty/src/event.rs
```diff
@@ -274,8 +274,12 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             self.update_selection(point, self.mouse.cell_side);
         }
 
+        // Scrolling inside Vi mode moves the cursor, so start typing.
+        if vi_mode {
+            self.on_typing_start();
+        }
         // Update dirty if actually scrolled or we're in the Vi mode.
-        *self.dirty |= lines_changed != 0;
+        *self.dirty |= lines_changed != 0 || vi_mode;
```
kchibisov, line 282: The comment above had actually a reason. and hes the bug and the comment were from the same commit.

alacritty/src/event.rs
```diff
@@ -274,8 +274,12 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
             self.update_selection(point, self.mouse.cell_side);
         }
 
+        // Scrolling inside Vi mode moves the cursor, so start typing.
+        if vi_mode {
+            self.on_typing_start();
+        }
```
chrisduerr, line 280:
```suggestion
        }
        
```

alacritty/src/event.rs
chrisduerr: This will **always** mark the terminal as dirty when scrolling in vi mode. Can't we avoid this when the vi cursor hasn't moved?

chrisduerr (changes requested):

alacritty/src/event.rs
kchibisov: We either need to pass `vi` position` or rewrite scrolling we're doing. The issue is that we rotate Vi cursor inside the `execute` separatelly to this scroll function.

I also don't think that adding more granular handling here worth the complexity, since if you scroll but we're cursor doesn't move it means that you do it in place...

chrisduerr: What do you mean by "execute separately to this scroll function"? Isn't this done by `terminal.scroll_display`? Couldn't we just check before and after?

chrisduerr (changes requested):

chrisduerr (approved):

PR #7362, opened 2023-11-12
Simplify powerline drawing algorithm

kchibisov:
Iterate over points in line instead of drawing it right away
and then finding it in the buffer.

Fixes: 4a26667060 (Use builtin font to draw powerline symbols)

### DIFF
--- a/alacritty/src/renderer/text/builtin_font.rs
+++ b/alacritty/src/renderer/text/builtin_font.rs
@@ -511,22 +511,22 @@ fn powerline_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> 
     let mut canvas = Canvas::new(width, height);
 
     let slope = 1;
-    let f_y0 = 1;
-    let g_y0 = height as i32 - f_y0 - 1;
+    let top_y = 1;
+    let bottom_y = height as i32 - top_y - 1;
 
     // Start with offset `1` and draw until the intersection of the f(x) = slope * x + 1 and
     // g(x) = H - slope * x - 1 lines. The intersection happens when f(x) = g(x), which is at
-    // x = H/2 - 1.
+    // x = (H - 2) / (2 * slope).
     let x_intersection = (height as i32 + 1) / 2 - 1;
 
-    let top_line = LineIterator::new(slope, f_y0, 0, x_intersection);
-    let bottom_line = LineIterator::new(-slope, g_y0, 0, x_intersection);
+    let top_line = (0..x_intersection).map(|x| line_equation(slope, x, top_y));
+    let bottom_line = (0..x_intersection).map(|x| line_equation(-slope, x, bottom_y));
 
     // Inner lines to make arrows thicker.
-    let mut top_inner_line =
-        LineIterator::new(slope, f_y0 + extra_thickness, 0, x_intersection - extra_thickness);
-    let mut bottom_inner_line =
-        LineIterator::new(-slope, g_y0 - extra_thickness, 0, x_intersection - extra_thickness);
+    let mut top_inner_line = (0..x_intersection - extra_thickness)
+        .map(|x| line_equation(slope, x, top_y + extra_thickness));
+    let mut bottom_inner_line = (0..x_intersection - extra_thickness)
+        .map(|x| line_equation(-slope, x, bottom_y - extra_thickness));
 
     // NOTE: top_line and bottom_line have the same amount of iterations.
     for (p1, p2) in top_line.zip(bottom_line) {
@@ -889,32 +889,9 @@ fn calculate_stroke_size(cell_width: usize) -> usize {
     cmp::max((cell_width as f32 / 8.).round() as usize, 1)
 }
 
-/// Iterator over the `f(x) = slope * x + offset` yielding (x, f(x))
-/// on each iteration.
-struct LineIterator {
-    slope: i32,
-    offset: i32,
-    x: i32,
-    width: i32,
-}
-
-impl LineIterator {
-    fn new(slope: i32, offset: i32, from_x: i32, width: i32) -> Self {
-        Self { x: from_x - 1, width, slope, offset }
-    }
-}
-
-impl Iterator for LineIterator {
-    type Item = (f32, f32);
-
-    fn next(&mut self) -> Option<Self::Item> {
-        self.x += 1;
-        if self.x >= self.width {
-            None
-        } else {
-            Some((self.x as f32, (self.slope * self.x + self.offset) as f32))
-        }
-    }
+/// `f(x) = slope * x + offset` equation.
+fn line_equation(slope: i32, x: i32, offset: i32) -> (f32, f32) {
+    (x as f32, (slope * x + offset) as f32)
 }
 
 #[cfg(test)]
--- a/alacritty/src/renderer/text/builtin_font.rs
+++ b/alacritty/src/renderer/text/builtin_font.rs
@@ -1,7 +1,7 @@
 //! Hand-rolled drawing of unicode [box drawing](http://www.unicode.org/charts/PDF/U2500.pdf)
 //! and [block elements](https://www.unicode.org/charts/PDF/U2580.pdf), and also powerline symbols.
 
-use std::{cmp, mem, ops};
+use std::{cmp, iter, mem, ops};
 
 use crossfont::{BitmapBuffer, Metrics, RasterizedGlyph};
 
@@ -519,8 +519,8 @@ fn powerline_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> 
     // x = H/2 - 1.
     let intersection = (height as i32 + 1) / 2 - 1;
 
-    let mut f_x = LineEquation::new(slope, f_y0, 0, intersection);
-    let mut g_x = LineEquation::new(-slope, g_y0, 0, intersection);
+    let f_x = LineEquation::new(slope, f_y0, 0, intersection);
+    let g_x = LineEquation::new(-slope, g_y0, 0, intersection);
 
     // Inner functions to make arrows thicker.
     let mut f_x_inner =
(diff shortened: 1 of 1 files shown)


alacritty/src/renderer/text/builtin_font.rs
```diff
@@ -15,6 +15,11 @@ const COLOR_FILL_ALPHA_STEP_3: Pixel = Pixel { _r: 64, _g: 64, _b: 64 };
 /// Default color used for filling.
 const COLOR_FILL: Pixel = Pixel { _r: 255, _g: 255, _b: 255 };
 
+const POWERLINE_TRIANGLE: char = '\u{e0b0}';
+const POWERLINE_ARROW: char = '\u{e0b1}';
+const POWERLINE_TRIANGLE_FLIPPED: char = '\u{e0b2}';
+const POWERLINE_ARROW_FLIPPED: char = '\u{e0b3}';
```
chrisduerr, line 21: Can we change this from flipped to RTL? RTL should be a pretty common abbreviation for text/unicode stuff and might be a little more descriptive than "flipped".

alacritty/src/renderer/text/builtin_font.rs
```diff
@@ -500,106 +506,50 @@ fn box_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> Raster
...
-                break;
-            }
-        }
-        let mut y2 = height / 2;
-        for row in height / 2..height {
-            if buffer[row * width]._r != 0 {
-                y2 = row;
+    // Start with offset `1` and draw until the intersection of the f(x) = x + 1 and
+    // g(x) = H - x - 1 lines. The intersection happens when f(x) = g(x), which is at
+    // x = H/2 - 1.
```
chrisduerr, line 519: I mean the intersection is always at height / 2, isn't it? This comment seems pointlessly complex for that.

Our height offset of 1 at the top and bottom is irrelevant since it applies to both the top and the bottom, thus not changing the center point.

Unless you're talking about the X position of the intersection, which I'd assume you're not here since the value ignores `slope`.

alacritty/src/renderer/text/builtin_font.rs
```diff
@@ -500,106 +506,50 @@ fn box_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> Raster
...
+    let intersection = (height as i32 + 1) / 2 - 1;
+
+    let f_x = LineEquation::new(slope, f_y0, 0, intersection);
+    let g_x = LineEquation::new(-slope, g_y0, 0, intersection);
+
+    // Inner functions to make arrows thicker.
+    let mut f_x_inner =
+        LineEquation::new(slope, f_y0 + extra_thickness, 0, intersection - extra_thickness);
+    let mut g_x_inner =
+        LineEquation::new(-slope, g_y0 - extra_thickness, 0, intersection - extra_thickness);
```
chrisduerr, line 529: These aren't functions, are they?

alacritty/src/renderer/text/builtin_font.rs
```diff
@@ -500,106 +506,50 @@ fn box_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> Raster
...
+
+    // NOTE f_x and g_x have the same amount of iterations.
```
chrisduerr, line 531: Not sure this is relevant, at the very least I'd want a `:` after `NOTE`.

alacritty/src/renderer/text/builtin_font.rs
```diff
@@ -500,106 +506,50 @@ fn box_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> Raster
...
+    for (p1, p2) in iter::zip(f_x, g_x) {
```
chrisduerr, line 532:
```suggestion
    for (p1, p2) in f_x.zip(g_x) {
```

I think this should just work since `LineEquation` implements iterator and it's shorter?

alacritty/src/renderer/text/builtin_font.rs
```diff
@@ -500,106 +506,50 @@ fn box_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> Raster
...
+        if character == POWERLINE_TRIANGLE || character == POWERLINE_TRIANGLE_FLIPPED {
+            canvas.draw_rect(0., p1.1, p1.0 + 1., 1., COLOR_FILL);
+            canvas.draw_rect(0., p2.1, p2.0 + 1., 1., COLOR_FILL);
+        } else if character == POWERLINE_ARROW || character == POWERLINE_ARROW_FLIPPED {
+            let p3 = f_x_inner.next().unwrap_or(p2);
+            let p4 = g_x_inner.next().unwrap_or(p1);
+
+            // Once we reach the canvas end fill the area between f(x) and g(x) to make a proper
+            // cut.
```
chrisduerr, line 541: This took me a second to figure out, I think it might be nice to be a little more verbose:

```
// If we can't fit the entire arrow in the cell, we cut off the tip of the arrow by drawing a rectangle between the two lines.
```

alacritty/src/renderer/text/builtin_font.rs
```diff
@@ -500,106 +506,50 @@ fn box_drawing(character: char, metrics: &Metrics, offset: &Delta<i8>) -> Raster
...
-        for row in height / 2..height {
-            if buffer[row * width]._r != 0 {
-                y2 = row;
+    // Start with offset `1` and draw until the intersection of the f(x) = x + 1 and
+    // g(x) = H - x - 1 lines. The intersection happens when f(x) = g(x), which is at
+    // x = H/2 - 1.
+    let intersection = (height as i32 + 1) / 2 - 1;
+
+    let f_x = LineEquation::new(slope, f_y0, 0, intersection);
+    let g_x = LineEquation::new(-slope, g_y0, 0, intersection);
```
chrisduerr, line 523: Honestly thinking about it `f_x` and `g_x` are pretty terrible names. Would be much easier to understand all of this by just naming them after what they actually are. Something like `top_line` and `bottom_line` maybe.

The whole `_x` thing is just confusing because it's basically referring to a variable that isn't actually passed to the line function (we just call `.next()`). It makes sense mathematically but the concept just doesn't translate.

alacritty/src/renderer/text/builtin_font.rs
```diff
@@ -960,6 +883,39 @@ impl Canvas {
     }
 }
 
+/// Size of the stroke to use.
+fn calculate_stroke_size(width: usize) -> usize {
```
chrisduerr, line 887:
```suggestion
/// Compute line width.
fn calculate_stroke_size(cell_width: usize) -> usize {
```

My main problem here is the "to use" suffix in your docs which is confusing (to use for what?). Decided to reword it completely to provide additional value over the function name itself (synonyms are more useful to figure out purpose than using the same words again).

alacritty/src/renderer/text/builtin_font.rs
```diff
@@ -960,6 +883,39 @@ impl Canvas {
...
+    // Use one eight of the cell width, since this is used as a step size for block elements.
+    cmp::max((width as f32 / 8.).round() as usize, 1)
+}
+
+/// Equation for `f(x) = slope * x + offset.
+struct LineEquation {
```
chrisduerr, line 893: This is a bad name. It's not an equation it's an iterator. An equation would map to a function in programming so it would just be `fn f(x: i32) -> i32` or whatever.

alacritty/src/renderer/text/builtin_font.rs
```diff
@@ -960,6 +883,39 @@ impl Canvas {
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.x += 1;
+        if self.x >= self.width {
+            None
+        } else {
+            Some((self.x as f32, (self.slope * self.x + self.offset) as f32))
+        }
+    }
+}
```
chrisduerr, line 917: Does having this as an iterator actually provide any value over defining it as a function and just iterating over a range? X is just incrementing by `1` anyway, so this is basically the same as `0..width.map(fun)`, isn't it?

chrisduerr (changes requested):

alacritty/src/renderer/text/builtin_font.rs
kchibisov: Yeah, because at some point I want inner functions to return `None` and `unwrap_or` on them? It's just _way_ simpler that way otherwise you need to build up more complex logic.

alacritty/src/renderer/text/builtin_font.rs
kchibisov: Oh, I'm talking about _x_, it's just slope is already included due to it being `+1` and `-1`. I've added `slope` explicitly. What I find is intersetion of the lines on `x` not `y`.

alacritty/src/renderer/text/builtin_font.rs
kchibisov: e.g.

```rust
            let p3 = top_inner_line.next().unwrap_or(p2);
            let p4 = bottom_inner_line.next().unwrap_or(p1);
```

chrisduerr: But a mapped range iterator would produce the exact same result?

alacritty/src/renderer/text/builtin_font.rs
chrisduerr: If the slope is changed, this will be wrong though.

alacritty/src/renderer/text/builtin_font.rs
kchibisov: Hm, let me try to see how it looks.

alacritty/src/renderer/text/builtin_font.rs
kchibisov: I've updated to be more general and not slope specific.

chrisduerr (approved):

PR #7365, opened 2023-11-13
Use DEC mode 2026 for Sync capability

gpanders:
Alacritty has supported mode 2026 for synchronized updates for a few
months [1], but the terminfo entry still uses the old DCS sequence
originally supported by iTerm2. Since many other terminal emulators and
applications seem to be [standardizing around 2026](https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036), change the terminfo
entry to use SM instead of DCS.

Note that even iTerm2's wiki now recommends using `DECSET`/`DECRST` instead of DCS: https://gitlab.com/gnachman/iterm2/-/wikis/synchronized-updates-spec#decsetdecreset

[1]: https://github.com/alacritty/alacritty/commit/47d500770a7fe306d710ca2cd9a0758c27a639a2


extra/alacritty.info
```diff
@@ -107,6 +107,6 @@ alacritty+common|base fragment for alacritty,
     Ms=\E]52;%p1%s;%p2%s\007, Se=\E[0 q, Ss=\E[%p1%d q,
     hs, dsl=\E]2;\007, fsl=^G, tsl=\E]2;,
     Smulx=\E[4\:%p1%dm,
-    Sync=\EP=%p1%ds\E\\,
+    Sync=\E[?2026%?%p1%{1}%-%tl%eh%;,
```
chrisduerr, line 110: Did you copy this from a different terminfo or write yourself?

gpanders: The entry is copied from [another terminal emulator](https://mitchellh.com/ghostty) which I’m helping to beta test. I didn’t write the original entry, but have since updated it.

kchibisov: @chrisduerr foot's entry is the same here if it matters.

chrisduerr: I'm mostly just curious because terminfo syntax is a massive pita. But it looks about right just the trailing semicolon confused me.

> foot's entry is the same here if it matters.

I just checked and it's not. It's missing the trailing semicolon.

chrisduerr: Can you just remove the trailing `%;`? It's not that this is wrong a semicolon here should work just fine, but I don't think there's any benefit to having it so might as well reduce the statement's complexity.

chrisduerr (changes requested):

extra/alacritty.info
gpanders: Wezterm has the same entry as foot, without the trailing semicolon. According to the [terminfo man page](https://man7.org/linux/man-pages/man5/terminfo.5.html ) the trailing `%;` is required for a conditional expression. Whether or not it actually matters in practice, though, I’m less sure.

chrisduerr (approved):

PR #7367, opened 2023-11-14
Damage only terminal inside `alacritty_terminal`

### LINKED ISSUE #7111: Cursor is not damaged when blinking and scrolled up

kchibisov:
The damage tracking was including selection and vi_cursor which were rendering viewport related, however all the damage tracking inside the `alacritty_terminal` was _terminal viewport_ related, meaning that it should be affected by `display_offset`.

Refactor the damage tracking so `alacritty_terminal` is only tracking actual terminal updates and properly applying display offset to them, while `alacritty` pulls this damage into its own UI damage state.

Fixes #7111.


alacritty/src/display/mod.rs
```diff
@@ -652,11 +648,19 @@ impl Display {
...
-        terminal.resize(new_size);
+            // Resize terminal.
+            terminal.resize(new_size);
+
+            // Resize damage tracking.
+            self.damage_tracker.resize(new_size.screen_lines(), new_size.columns());
+        }
 
         // Check if dimensions have changed.
         if new_size != self.size_info {
```
chrisduerr, line 666: We already check if the dimensions have changed further down here, is there a reason to check twice?

alacritty/src/display/mod.rs
```diff
@@ -502,27 +499,25 @@ impl Display {
...
-            meter: Meter::new(),
             size_info,
-            ime: Ime::new(),
-            highlighted_hint: None,
-            vi_highlighted_hint: None,
-            cursor_hidden: false,
+            hint_state,
+            glyph_cache,
+            damage_tracker,
+            raw_window_handle,
```
chrisduerr, line 506: Descending order would be much better here. :(

alacritty/src/display/damage.rs
```diff
@@ -1,20 +1,197 @@
...
+                }
+            }
+        }
+    }
+}
+
+/// Damage state for the rendering frame.
+#[derive(Debug, Default)]
+pub struct FrameDamage {
+    /// The entire frame got redrawn.
```
chrisduerr, line 144:
```suggestion
    /// The entire frame needs to be redrawn.
```

alacritty/src/display/damage.rs
```diff
@@ -1,20 +1,197 @@
...
+    damage_all: bool,
```
chrisduerr, line 145:
```suggestion
    full: bool,
```

`damage_all` sounds more like a verb than an adjective. Should probably be `all`, though `full` sounds better in this context imo. `full_damage` would be another option if you want to be more verbose, but we're talking about `FrameDamage` already anyway so it shouldn't be necessary.

alacritty/src/display/damage.rs
```diff
@@ -1,20 +1,197 @@
...
+    /// Terminal lines damaged in the given frame.
+    lines: Vec<LineDamageBounds>,
+    /// Rectangular regions damage in the given frame.
+    rects: Vec<Rect>,
+}
+
+impl FrameDamage {
+    /// Damage line for the given frame.
```
chrisduerr, line 153:
```suggestion
    /// Damage line for this frame.
```

There isn't really a frame given here, since that would imply parameters. You could argue that `&self` is a parameter but it doesn't really make sense in the context of a method doc on a struct.

alacritty/src/display/damage.rs
```diff
@@ -1,20 +1,197 @@
...
+    /// Add a damage rectangle.
+    #[inline]
+    pub fn add_rect(&mut self, x: i32, y: i32, width: i32, height: i32) {
+        self.rects.push(Rect { x, y, width, height });
+    }
+
+    fn reset(&mut self, num_lines: usize, num_cols: usize) {
+        self.damage_all = false;
+        self.rects.clear();
+        self.lines.clear();
```
chrisduerr, line 179: I don't think this is necessary, but just to make sure: Do we have to deal with runaway memory here?

The line damage should always be limited to the viewport size which will be irrelevant for memory. But rects could potentially be a lot I'd imagine with a ton of underlines. I'd imagine the maximum number of rects you'd get in two frames wouldn't be enough to justify shrinking the vec on reset, but something to consider maybe.

alacritty/src/display/damage.rs
```diff
@@ -1,20 +1,197 @@
...
+        } else {
+            let lines_damage = RenderDamageIterator::new(
+                TermDamageIterator::new(&self.frames[0].lines, 0),
+                &size_info,
+            );
+            lines_damage.chain(self.frames[0].rects.iter().copied()).collect()
+        }
+    }
+
+    /// Set the selection adding the damage for the current frame.
```
chrisduerr, line 105:
```suggestion
    /// Add the current frame's selection damage.
```

"Set the selection" seems misleading.

alacritty/src/display/damage.rs
```diff
@@ -1,20 +1,197 @@
...
+        mut selection: Option<SelectionRange>,
+        display_offset: usize,
+    ) {
+        mem::swap(&mut self.old_selection, &mut selection);
+
+        if self.frame().damage_all {
+            return;
+        }
+
+        if selection != self.old_selection {
```
chrisduerr, line 117:
```suggestion
        if self.frame().damage_all || selection == self.old_selection {
            return;
        }

```

alacritty/src/display/damage.rs
```diff
@@ -1,20 +1,197 @@
...
+            for selection in self.old_selection.into_iter().chain(selection) {
+                // self.damage_selection_impl(selection, display_offset);
```
chrisduerr, line 119: ?

alacritty/src/display/damage.rs
```diff
@@ -1,20 +1,197 @@
...
+                let display_offset = display_offset as i32;
+                let last_visible_line = self.screen_lines as i32 - 1;
+                let columns = self.columns;
+
+                // Don't damage invisible selection.
+                if selection.end.line.0 + display_offset < 0
+                    || selection.start.line.0.abs() < display_offset - last_visible_line
+                {
+                    continue;
+                };
```
chrisduerr, line 129:
```suggestion
                // Ignore invisible selection.
                if selection.end.line.0 + display_offset < 0
                    || selection.start.line.0.abs() < display_offset - last_visible_line
                {
                    continue;
                };
```

Ignore seems a bit better as a verb here because ultimately we can't actually damage it even if we wanted to, since damage is only tracked within the viewport.

alacritty/src/display/damage.rs
```diff
@@ -1,20 +1,197 @@
...
+
+                let start = cmp::max(selection.start.line.0 + display_offset, 0);
+                let end = (selection.end.line.0 + display_offset).clamp(0, last_visible_line);
+                for line in start as usize..=end as usize {
```
chrisduerr, line 133: Might be easier to put the usize at the ends of start/end here. There's parenthesis around everything anyway and that way the range becomes much easier to read.

alacritty/src/display/damage.rs
```diff
@@ -1,20 +1,197 @@
...
+            return;
+        }
+
+        if let Some(vi_cursor) = self.old_vi_cursor {
+            self.frame().damage_point(vi_cursor);
+        }
+
+        if let Some(vi_cursor) = vi_cursor {
+            self.frame().damage_point(vi_cursor);
+        }
```
chrisduerr, line 89: It's probably unnecessary, but you could potentially add the `damage_all` check to `damage_point` and all the other frame methods.

That way it would be guaranteed that none of that will ever cause actual damage generation since it's unnecessary.

That said, these operations should be so cheap I'm not sure the early return is even worth it here.

alacritty_terminal/src/term/mod.rs
```diff
@@ -417,27 +395,25 @@ impl<T> Term<T> {
...
     }
 
+    /// Collect the information about the changes in the lines, which
+    /// could be used to minimize the amount of drawing operations.
+    ///
+    /// The user controlled elements, like `Vi` mode cursor and `Selection` **are not** part of the
+    /// collected damage state. Those could easily be tracked by comparing their old and new
+    /// value between adjacent frames.
+    ///
+    /// After reading damage [`reset_damage`] should be called.
```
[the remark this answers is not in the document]
kchibisov: Just to allow to call this function more than once, like maybe you want to check that some part is damaged and run custom logic. In our use case it's always _take_ though.

alacritty/src/display/mod.rs
kchibisov: They are not equivalent, because we could change width without changing dimensions, like you the change resulted only in the padding change.

alacritty/src/display/mod.rs
kchibisov: It won't be a pyramid though, because we have descending for defaults, thus it means we need ascending for regular fields so they meet at some point?

alacritty/src/display/damage.rs
kchibisov: You're confusing which rects we're talking about. The damage rects here has nothing to do with underlines. It's for the UI elements outside of the viewport, so the amount is limited to 1-2 rectangles.

For now it's used in a single place with the message bar as you can see.

alacritty/src/display/damage.rs
kchibisov: I decided to not add because that it's cheap, since it's zero conditions. The same reason we don't have a check in alacritty_terminal since it's more expensive to check than do an actual write.

alacritty_terminal/src/term/mod.rs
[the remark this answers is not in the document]
chrisduerr: This function isn't particularly cheap though, so wouldn't you instead take the damage and then operate on that, instead of calling this function multiple times?

It also mutates the damage based on the pending state, so it's not idempotent.

alacritty/src/display/mod.rs
chrisduerr: Fair enough though that seems extremely unlikely to happen in practice.

alacritty/src/display/mod.rs
chrisduerr: Realistically it should always be descending. Otherwise the order has to be changed completely and that's not possible because then default would be at the top.

alacritty/src/display/damage.rs
```diff
@@ -1,20 +1,196 @@
...
+            debug: false,
+            old_vi_cursor: None,
+            old_selection: None,
+            frames: Default::default(),
+        };
+        tracker.resize(screen_lines, columns);
+        tracker
+    }
+
+    #[inline]
```
chrisduerr, line 44: Just in general I don't think inline annotations in alacritty will make a big difference. Generally that should be inlined anyway and if not there might be a good reason for it (though realistically I don't see why this wouldn't be).

alacritty_terminal/src/term/mod.rs
```diff
@@ -417,30 +395,30 @@ impl<T> Term<T> {
...
         let previous_cursor = mem::replace(&mut self.damage.last_cursor, self.grid.cursor.point);
-        let previous_selection = mem::replace(&mut self.damage.last_selection, selection);
-        let previous_vi_cursor_point =
-            mem::replace(&mut self.damage.last_vi_cursor_point, vi_cursor_point);
 
-        // Early return if the entire terminal is damaged.
-        if self.damage.is_fully_damaged {
+        // NOTE: the terminal damage is tracked from the bottom, however scrolling could damage
+        // the viewport due to lines going off the scrollback, hence we don't check for display
+        // offset when we have full damage.
```
chrisduerr, line 420: I'm not sure this is the right place for this comment. If there's full damage I don't think anyone would assume you'd need to check anything? So the detail provided here might be correct, but it's useless.

Also the "from the bottom" here is confusing. I'd assume what you mean is that terminal damage is tracked only for the viewport. If that's from the bottom or top of the viewport is irrelevant, the important bit is that it's the viewport we care about.

chrisduerr (changes requested):

alacritty_terminal/src/term/mod.rs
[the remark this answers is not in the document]
kchibisov: I'm not sure I like the extra complexity with doing the `take` approach, simply because you need to attach the `&mut` to the iterator itself and then reset damage when you drop it, it's just messy.

The way `&mut` is done in this function is because you need current cursor to _finalize_ the damage tracking state, otherwise it'll be wrong. The computation is not that complex though, since it basically does some assignments to make the damage valid.

alacritty_terminal/src/term/mod.rs
kchibisov: I have no idea where to put this comment as well, because it all sounds like I'm explaining how damage works when we have all the parts in place and test covered. This comment is neither `Why` nor `What`, it's just states the fact on how things happened to work.

>I'd assume what you mean is that terminal damage is tracked only for the viewport.

Viewport is what you _view_, which is not the case, because once you've scrolled your viewport is moved to that offset. I mean the area input can write to, but it's not a viewport.

alacritty/src/display/damage.rs
kchibisov: Yeah, but I want to give a hint that it should be inlined. We already add annotations like that to other small functions through out the alacritty code base.

chrisduerr (approved):

PR #7381, opened 2023-11-20
Add inode/directory compatible desktop entry

simvux:
Some terminals support being set as the default file manager so that UI elements such as `Browse Files` can open the terminal but `cd`'d into that directory. 

Kitty accomplishes this by having a separate desktop entry that supports being a file manager and a ssh client. 
Alacritty's CLI doesn't really have a flag to allow it to be both a SSH client and a file manager under the same flag, however the desktop entry file provided here supports at least the file manager part. 
```sh
$ xdg-mime query default inode/directory
Alacritty-open.desktop
```

Long term, maybe alacritty should introduce a `--uri` flag to it's CLI which parses the URI to be either a path, or a compatible protocol such as `ssh://` to then act accordingly? That way we could have one generalized desktop entry that supports numerous mimetypes.

### DIFF
--- a/extra/linux/Alacritty-open.desktop
+++ b/extra/linux/Alacritty-open.desktop
@@ -1,0 +1,13 @@
+[Desktop Entry]
+Type=Application
+Exec=alacritty --working-directory %U
+TryExec=alacritty
+Icon=Alacritty
+Terminal=false
+Categories=System;TerminalEmulator;
+Name=Alacritty
+GenericName=Terminal
+Comment=Open URLs with Alacritty
+StartupWMClass=Alacritty
+Actions=New;
+MimeType=inode/directory;text/*;x-scheme-handler/alacritty;x-scheme-handler/ssh;
--- a/extra/linux/Alacritty-open.desktop
+++ b/extra/linux/Alacritty-open.desktop
@@ -10,4 +10,4 @@ GenericName=Terminal
 Comment=Open URLs with Alacritty
 StartupWMClass=Alacritty
 Actions=New;
-MimeType=inode/directory;text/*;x-scheme-handler/alacritty;x-scheme-handler/ssh;
+MimeType=inode/directory;text/*;x-scheme-handler/alacritty;


kchibisov:
Can't we have other action in the same desktop entry? I think you can have multiple for multiple things.

simvux:
It *might* be possible by adding numerous Actions in the desktop entry however; from what I can tell there's no way to assign a specific mimetype to an action. So I think the only way to implement it is to add URI parsing code in Alacritty so that multiple mimetypes can be put under the same action. Similarly to `kitty +open <URI>`. 

Desktop entry actions are more to add extra initialization options in context menus

simvux:
If we are to add parsing for this in Alacritty then all of these mimetypes could be put under the original `Alacritty.desktop` if we edge-case the argument being empty or missing (in case of the desktop entry being ran normally). Although whether that's less or more hacky is up for interpretation. 

I'm not all that familiar with the xdg file entries so it's also entirely possible that there's a much better way of doing this. I was mainly just looking at how Kitty was able to open path links, and worked from there.

simvux:
> I mean, usually mime type opening in terminals is done by having `-e` and just passing command. Like the way you can open tui `nvim` or htop right in alacritty. The same way you can open ssh inside alacritty.

Ye but that's not actually registered in xdg. For example; if I press a ssh link then that can't automatically launch in an Alacritty terminal. Because Alacritty is not registered to handle `x-scheme-handler/ssh`.

kchibisov:
Then you need a desktop entry for `ssh`? alacritty is not an ssh client by any means, we can just run thing.

simvux:
afaik that's not really how these specifications are meant to work. It's up to the applications themselves to register which formats they can handle (which they do in their respective desktop entries). There's no `ssh` desktop entry which uses some standardized `-e` flag onto the `TerminalEmulator` xdg category. 

Again; the problem we're trying to solve here is that pressing links or UI elements that could spawn alacritty if the user so wishes currently can't. For example; if a user that uses alacritty also just uses their standard terminal for ssh (which would be most Linux users), then pressing a `ssh://192.168.0.33/` link somewhere or opening that URI in the web browser would currently present the user with a set of `ssh://` compatible programs to choose from (or pick the default). However; Alacritty is not registered as one of them. Even though most alacritty users are used to using alacritty for ssh. 

Since I also navigate files in the terminal, the same thing applies to just general paths. 

We could add a `Alacritty-ssh.desktop` file which sets the `Exec` file to something like `alacritty -e "ssh %U"` but that seems rather hacky in comparison. 

If you think that terminals like kitty are scope-creeping what they should include in their packaging by registering to these mimetypes then that's pretty reasonable. Users who would like their terminal to be the default for these sort of things are also the kind of users who wouldn't mind creating their own desktop entries. But it's a feature worth considering.

kchibisov:
alacritty has no idea what to run to handle `ssh://` links, what if my client is not named `ssh` in the first place, but `mosh` or something else? This should be really done the way `nvim`, `htop`, or other programs do that. So it'll work with _any_ terminal out there.

simvux:
Hm, ye I suppose that could make more sense for `ssh` specifically. But what about file browsing? 

In that case there's no application we're trying to spawn inside a terminal. The objective is to spawn the terminal itself (just in the right directory).

kchibisov:
Usually file browsing is left to GUI application, which has option to `open new terminal here`. The way `here` could be done is by fork + cwd + execve, so it doesn't even need to know about what type of terminal it is or its flags. It should only know that it's a `Terminal`, which users could set with `Terminal`. The terminal will properly inherit cwd and open in the directory GUI manager wanted it.

simvux:
I wouldn't want to go through an extra file manager just to have it open a terminal. 

To give you an example. In Steam I might want to `Right Click -> Manage -> Browse local files` on a game and have that open Alacritty into that directory. 

Or in some graphical application I might want to click `Open logs folder` and have that open Alacritty into that directory. 

I don't use graphical file managers, I navigate and manage my filesystem through my terminal.

simvux:
Yes. 

TUI file managers probably just have a desktop entry Exec'ing themselves with `Terminal=true` so it runs on the default terminal. 

Regardless, up to you whether you find that the use-case is too niche for this to be merged. At least the PR now exists so that people who are asking for the same thing can just copy it.

chrisduerr:
Users are always welcome to make special desktop entries themselves for things like `alacritty -e ssh` or whatever. But that makes assumptions about other applications and thus should be left up to the user.

PR #7384, opened 2023-11-23
Create only one branch per major release

kchibisov:
Having a separate branch for each release makes it harder to maintain without an actual benefit, since every release from the major version is linear, so creating branches doesn't make any sense.

They also collapse with the tag names leading to ambiguous refs.

--

We've discussed it at some point, and I remember you've agreed. The goal is to make it simpler.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,14 +113,14 @@ documentation comments.
 Alacritty's release process aims to provide stable and well tested releases without having to hold
 back new features during the testing period.
 
-To achieve these goals, a new branch is created for every new release. Both the release candidates
+To achieve these goals, a new branch is created for every major release. Both the release candidates
 and the final version are only committed and tagged in this branch. The master branch only tracks
 development versions, allowing us to keep the branches completely separate without merging releases
 back into master.
 
 The exact steps for an exemplary `0.2.0` release might look like this:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -113,14 +113,14 @@ documentation comments.
...
-To achieve these goals, a new branch is created for every new release. Both the release candidates
+To achieve these goals, a new branch is created for every major release. Both the release candidates
 and the final version are only committed and tagged in this branch. The master branch only tracks
 development versions, allowing us to keep the branches completely separate without merging releases
 back into master.
 
 The exact steps for an exemplary `0.2.0` release might look like this:
   1. Initially, the version on the latest master is `0.2.0-dev`
-  2. A new `v0.2.0` branch is created for the release
+  2. A new `v0.2.x` branch is created for the release
```
chrisduerr, line 123:
```suggestion
  2. A new `v0.2` branch is created for the release
```

Probably better to just omit it? That way nobody has to get confused by which letter to use, especially once it becomes `v1` eventually.

chrisduerr (changes requested):

CONTRIBUTING.md
kchibisov: Hm, I guess it's fine, given that versions have 3 digits.

chrisduerr (approved):

PR #7386, opened 2023-11-23
update windows-sys to 0.52.0

deid84:
rebuilt and retested without warnings or errors, dependency update seems to work fine


kchibisov:
Every dependency is using 0.48 version, so there's no reason to update to 0.51, unless the rest is updated, and that won't happen until the next major release of the majority of the deps.

PR #7387, opened 2023-11-23
Improve man pages docs

kchibisov:
--

Just view each commit individually with its message.


kchibisov:
The CI fails due to me changing help for some CLI stuff, will regenerate after ack on specific commit.

chrisduerr (approved):

PR #7388, opened 2023-11-24
Fix ime position

lilydjwg:
In the past, only the bottom left corner point was passed to ime, so the position was `point.line + 1` to account for current line. However, a rectangle is now passed, and the coordinate passed should be the top left corner of the rectangle. The current line height should not be added in this case.

### DIFF
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -413,7 +413,7 @@ impl Window {
     /// Adjust the IME editor position according to the new location of the cursor.
     pub fn update_ime_position(&self, point: Point<usize>, size: &SizeInfo) {
         let nspot_x = f64::from(size.padding_x() + point.column.0 as f32 * size.cell_width());
-        let nspot_y = f64::from(size.padding_y() + (point.line + 1) as f32 * size.cell_height());
+        let nspot_y = f64::from(size.padding_y() + point.line as f32 * size.cell_height());
 
         // Exclude the rest of the line since we edit from left to right.
         let width = size.width as f64 - nspot_x;


kchibisov (changes requested):
Which platform are you on? Like this surely breaks IME for me, so it overlaps the content, but I'd also need to test this on GNOME.

kchibisov:
I think something like that should do the job.

```diff
diff --git a/alacritty/src/display/window.rs b/alacritty/src/display/window.rs
index 7dc6375e..14df091d 100644
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -112,6 +112,7 @@ pub struct Window {
     /// Current window title.
     title: String,
 
+    is_x11: bool,
     current_mouse_cursor: CursorIcon,
     mouse_visible: bool,
 }
@@ -188,15 +189,17 @@ impl Window {
 
         let scale_factor = window.scale_factor();
         log::info!("Window scale factor: {}", scale_factor);
+        let is_x11 = matches!(window.raw_window_handle(), RawWindowHandle::Xlib(_));
 
         Ok(Self {
-            current_mouse_cursor,
-            mouse_visible: true,
             requested_redraw: false,
-            window,
             title: identity.title,
+            current_mouse_cursor,
+            mouse_visible: true,
             has_frame: true,
             scale_factor,
+            window,
+            is_x11,
         })
     }
 
@@ -412,11 +415,17 @@ impl Window {
 
     /// Adjust the IME editor position according to the new location of the cursor.
     pub fn update_ime_position(&self, point: Point<usize>, size: &SizeInfo) {
+        // NOTE: X11 doesn't support cursor area, so we need to offset manually to not obscure
+        // the text.
+        let offset = if self.is_x11 { 1 } else { 0 };
         let nspot_x = f64::from(size.padding_x() + point.column.0 as f32 * size.cell_width());
-        let nspot_y = f64::from(size.padding_y() + (point.line + 1) as f32 * size.cell_height());
+        let nspot_y =
+            f64::from(size.padding_y() + (point.line + offset) as f32 * size.cell_height());
 
-        // Exclude the rest of the line since we edit from left to right.
-        let width = size.width as f64 - nspot_x;
+        // NOTE: some compositors don't like excluding too much and try to render popup at the
+        // bottom right corner of the provided area, so exclude just the full-width char to not
+        // obscure the cursor and not render popup at the end of the window.
+        let width = size.cell_width() as f64 * 2.;
         let height = size.cell_height as f64;
 
         self.window.set_ime_cursor_area(
```

kchibisov:
>In the past, only the bottom left corner point was passed to ime

This is also not correct, the Top-right corner is what is always passed, the issue you see is likely because the logic to not-obscure the cursor moves it even further that way.

lilydjwg:
Oh I didn't think about x11. I'm on wayland (KDE / wlroots). Your patch sounds reasonable.

>  In the past, only the bottom left corner point was passed to ime

I meant the cursor rect. It used to pass a width=height=0 rect to wayland.

kchibisov:
>I meant the cursor rect. It used to pass a width=height=0 rect to wayland.

Yeah, but we were offsetting manually for that.

kchibisov:
See https://github.com/alacritty/alacritty/pull/7389

kchibisov:
Also, if you want me to remove your from commit trailer for whatever reason, let me know.

PR #7390, opened 2023-11-24
Damage entire window on font size change

kchibisov:
Font size could change without changing the cell dimensions, like becoming slightly higher/wider.

Fixes: 40160c5d (Damage only terminal inside `alacritty_terminal`)

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -618,7 +618,8 @@ impl Display {
 
             info!("Cell size: {} x {}", cell_width, cell_height);
 
-            // NOTE: fonts can change glyph sizes keeping the same cell size.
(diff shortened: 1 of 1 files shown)


alacritty/src/display/mod.rs
```diff
@@ -617,6 +617,9 @@ impl Display {
             cell_height = cell_dimensions.1;
 
             info!("Cell size: {} x {}", cell_width, cell_height);
+
+            // NOTE: fonts can change glyph sizes keeping the same cell size.
```
chrisduerr, line 621:
```suggestion
            // Mark entire terminal as damaged since glyph size could change without cell size changes.
```

chrisduerr (changes requested):

chrisduerr (approved):

PR #7391, opened 2023-11-24
Bump dependencies

kchibisov:
Update dependencies with `cargo update` as well as do explicit update of winit, crossfont, and regex-automata.

### DIFF
--- a/alacritty_terminal/Cargo.toml
+++ b/alacritty_terminal/Cargo.toml
@@ -21,7 +21,7 @@ libc = "0.2"
 log = "0.4"
 parking_lot = "0.12.0"
 polling = "3.0.0"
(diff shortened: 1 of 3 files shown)


alacritty_terminal/Cargo.toml
```diff
@@ -21,7 +21,7 @@ libc = "0.2"
 log = "0.4"
 parking_lot = "0.12.0"
 polling = "3.0.0"
-regex-automata = "0.3.6"
+regex-automata = "0.4.3"
```
kchibisov, line 24: I've read the history and haven't found an obvious reason, but I'd suggest you to double check on that matter.

chrisduerr (approved):

PR #7405, opened 2023-12-01
Don't use `::*` for enum variants in bindings

kchibisov:


alacritty/src/config/bindings.rs
```diff
@@ -735,15 +733,15 @@ impl<'a> Deserialize<'a> for BindingKey {
                 } else {
                     // Translate legacy winit codes into their modern counterparts.
                     match keycode.as_str() {
-                        "Up" => (Key::Named(ArrowUp), KeyLocation::Standard),
```
chrisduerr, line 738: Couldn't this just be `NamedKey::ArrowUp`? I'd much prefer that since it's much easier to actually understand for someone writing new bindings.

alacritty/src/config/bindings.rs
```diff
@@ -412,25 +410,25 @@ macro_rules! bindings {
...
-        $key
+    (MouseBinding, $button:ident,) => {{
+        MouseButton::$button
     }};
 }
 
 pub fn default_mouse_bindings() -> Vec<MouseBinding> {
     bindings!(
         MouseBinding;
-        MouseButton::Right;                            MouseAction::ExpandSelection;
```
chrisduerr, line 431: Honestly preferred the old one, how is anyone supposed to figure this out with just `Right`?

chrisduerr (changes requested):

alacritty/src/config/bindings.rs
kchibisov: it's conversion code and not bindings. Not sure what to do here?

alacritty/src/config/bindings.rs
chrisduerr: It worked just fine before, didn't it?

kchibisov: Because of `expr`, but `Right` is not expr, since it complains that the type is not here, so you need either `ident` or `tt`, but you can't use `ident`, because you have literals, so the only option is `tt`, but `::` is token, so you get the issue...

kchibisov: The macro is common with keyboard which has literals.

chrisduerr (changes requested):

chrisduerr (approved):

PR #7407, opened 2023-12-01
Optimize undercurl shader

kchibisov:
This removes the if and lowers amount of operations.


kchibisov:
arch broke again, sigh...

PR #7421, opened 2023-12-06
Don't set `LC_ALL` on macOS

### LINKED ISSUE #7360: Don't set `LC_ALL` on macOS
It seems like this is not really needed anymore and breaks on Sonoma. I'm not
even sure why we tried to do it that way...

zefr0x:
Closes #7360

I wasn't able to test it in `x86_64-apple-darwin` or `aarch64-apple-darwin`, but it works in `x86_64-unknown-linux-gnu`.
Edited: Great, there is a workflow for macos.


kchibisov:
You just broke macOS entirely. We need to set locale similarly to what we have, but tune to what iterm, etc do. Without it there won't be any locale set at all for macOS.

PR #7424, opened 2023-12-07
Don't emit text for NamedKey without text repr

### LINKED ISSUE #7423: Arrow keys print  (up, down, left, right) instead of moving the cursor.
Since cb03806e2ab85674c45e87e1bb24dfe2fd1a918c pressing the arrow keys prints one of these characters:
- UP:  
- Down: 
- Left: 
- Right: 

I obviously expected that they'd make the cursor go one character in the respective direction instead.

### System

OS: macOS
Version: alacritty 0.13.0-dev (cb03806e)

### Logs

`alacritty --print-events` log of a window where I pressed "Arrow Up", "Arrow Down", "Arrow Left", "Arrow Right" and then closed the window with "cmd+w": [print-events.log](https://github.com/alacritty/alacritty/files/13595220/print-events.log)

kchibisov:
When the key doesn't have textual representation we shouldn't emit the text for them, since they are processed via bindings.

Fixes #7423.

### DIFF
--- a/alacritty/src/input/keyboard.rs
+++ b/alacritty/src/input/keyboard.rs
@@ -110,13 +110,13 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
             return true;
         }
 
-        let disambiguate = if mode.contains(TermMode::DISAMBIGUATE_ESC_CODES) {
+        let mut disambiguate = false;
+        if mode.contains(TermMode::DISAMBIGUATE_ESC_CODES) {
             let on_numpad = key.location == KeyLocation::Numpad;
             let is_escape = key.logical_key == Key::Named(NamedKey::Escape);
-            is_escape || (!mods.is_empty() && mods != ModifiersState::SHIFT) || on_numpad
-        } else {
-            false
-        };
+            disambiguate =
+                is_escape || (!mods.is_empty() && mods != ModifiersState::SHIFT) || on_numpad;
+        }
 
         match key.logical_key {
             _ if disambiguate => true,
--- a/alacritty/src/input/keyboard.rs
+++ b/alacritty/src/input/keyboard.rs
@@ -107,15 +107,22 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
         mods: ModifiersState,
     ) -> bool {
         if mode.contains(TermMode::REPORT_ALL_KEYS_AS_ESC) {
-            true
-        } else if mode.contains(TermMode::DISAMBIGUATE_ESC_CODES) {
+            return true;
+        }
+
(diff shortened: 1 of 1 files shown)


alacritty/src/input/keyboard.rs
```diff
@@ -107,15 +107,22 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
-            text.is_empty() || key.logical_key == Key::Named(NamedKey::Delete)
+            false
+        };
+
+        match key.logical_key {
+            _ if disambiguate => true,
+            // Exclude all the named keys unless they have textual representation.
+            Key::Named(named) => named.to_text().is_none(),
+            _ => text.is_empty(),
         }
```
kchibisov, line 126: This should be a _common fallback_ and not `else` clause, otherwise `Arrow` will be omitted (note that you can't face it in vim, because cursor keys, so it's handled before everything else).

RC-14:
Can confirm that this fixes #7423.

alacritty/src/input/keyboard.rs
```diff
@@ -107,15 +107,22 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
-            true
-        } else if mode.contains(TermMode::DISAMBIGUATE_ESC_CODES) {
+            return true;
+        }
+
+        let disambiguate = if mode.contains(TermMode::DISAMBIGUATE_ESC_CODES) {
             let on_numpad = key.location == KeyLocation::Numpad;
             let is_escape = key.logical_key == Key::Named(NamedKey::Escape);
             is_escape || (!mods.is_empty() && mods != ModifiersState::SHIFT) || on_numpad
         } else {
```
chrisduerr, line 117: The else branch is just `false`, so this can probably be written as a single conditional?

You could still keep the variables extracted for documentation purposes, but they're also pretty self-documenting.

alacritty/src/input/keyboard.rs
```diff
@@ -107,15 +107,22 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
-            // `Delete` key always has text attached to it, but it's a named key, thus needs to be
-            // excluded here as well.
-            text.is_empty() || key.logical_key == Key::Named(NamedKey::Delete)
+            false
+        };
+
+        match key.logical_key {
+            _ if disambiguate => true,
```
chrisduerr, line 122: Not sure this has any business being inside this match statement. Though I guess it's more concise than the 3-4 lines for a separate conditional.

chrisduerr (changes requested):

alacritty/src/input/keyboard.rs
kchibisov: yeah, I just don't like `if true { true }` branch.

chrisduerr (approved):

PR #7433, opened 2023-12-14
Update alacritty versions

chrisduerr:
This is only an update to the development versions and does not represent a stable release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,9 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 0.13.0-dev
+## 0.14.0-dev
+
+## 0.13.0
 
 ### Packaging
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 0.13.0-dev
+## 0.14.0-dev
+
+## 0.13.0
```
kchibisov, line 10: I think we always used `rc` here? Don't remember.

kchibisov (changes requested):
All the dev bumps except alacritty itself should be non-breaking.

CHANGELOG.md
chrisduerr: > On master, the version is bumped to `0.3.0-dev` and the `-dev` suffix is removed from the `0.2.0-dev` changelog

Should be fine according to our guideline.

chrisduerr:
> All the dev bumps except alacritty itself should be non-breaking.

Yeah I wasn't actually sure about that. But I guess it's consistent with what we did in the past.

kchibisov (approved):

PR #7439, opened 2023-12-16
Fix the order of `ClearSelection` in Vi mode

### LINKED ISSUE #7438: Can not copy when `Ctrl+Shift+C` is pressed in Vi mode
### System

OS: Windows
Version: alacritty 0.13.0-rc1 (ada35ae)

### Issue

When I switch to Vi mode in Alacritty 0.13.0-rc1, press `v` to select a portion of text, and then press `Ctrl+Shift+C`, the selected text **cannot** be copied to the clipboard.

And if I set key bindings in `alacritty.toml` as follows, I can copy the selected text successfully in Vi mode:

```
[keyboard]
bindings = [
    { key = "C", mods = "Control|Shift", action = "Copy" },
    { key = "C", mods = "Control|Shift", mode = "Vi|~Search", action = "ClearSelection" },
]
```

This problem did not occur in version 0.12.3, but it first appeared in version 0.13.0-rc1

kchibisov:
Fixes #7438.

--

Maybe we should do that without a binding instead, e.g. when copying from inside the Vi mode, so it's not config order dependent and consistent with user added bindings.

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -545,12 +545,12 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
 fn common_keybindings() -> Vec<KeyBinding> {
     bindings!(
         KeyBinding;
-        "c",    ModifiersState::CONTROL | ModifiersState::SHIFT, +BindingMode::VI, ~BindingMode::SEARCH; Action::ClearSelection;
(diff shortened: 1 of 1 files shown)


chrisduerr:
> Maybe we should do that without a binding instead, e.g. when copying from inside the Vi mode, so it's not config order dependent and consistent with user added bindings.

No.

chrisduerr (approved):

PR #7446, opened 2023-12-19
Account for option_as_alt when doing kitty protocol

### LINKED ISSUE #7443: Alt + 4 wrongly sends escape sequence inside Alacritty+Neovim (Swedish/Nordic kb layout!)
OS: macOS Ventura 13.6.1
Version: Current commit: https://github.com/alacritty/alacritty/tree/a10fb8adc0857633be428d673ab8d36d2a90193a (but same issues since kitty key protocol)
Keyboard layout: Swedish/Nordic

option_as_alt = "None"

Key events:

Same output for command line inside Alacritty, Inside Neovim, Inside Neovim inside Tmux:
```
[122.479315289s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(140634396691760)), event: KeyboardInput { device_id: DeviceId(DeviceId), event: KeyEvent { physical_key: Code(AltLeft), logical_key: Named(Alt), text: None, location: Left, state: Pressed, repeat: false, platform_specific: KeyEventExtra { text_with_all_modifiers: None, key_without_modifiers: Named(Alt) } }, is_synthetic: false } }
[122.483869688s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(140634396691760)), event: KeyboardInput { device_id: DeviceId(DeviceId), event: KeyEvent { physical_key: Code(Digit4), logical_key: Character("$"), text: Some("$"), location: Standard, state: Pressed, repeat: false, platform_specific: KeyEventExtra { text_with_all_modifiers: Some("$"), key_without_modifiers: Character("4") } }, is_synthetic: false } }
[122.550189254s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(140634396691760)), event: KeyboardInput { device_id: DeviceId(DeviceId), event: KeyEvent { physical_key: Code(Digit4), logical_key: Character("$"), text: None, location: Standard, state: Released, repeat: false, platform_specific: KeyEventExtra { text_with_all_modifiers: Some("$"), key_without_modifiers: Character("4") } }, is_synthetic: false } }
[122.554484575s] [INFO ] [alacritty] winit event: WindowEvent { window_id: WindowId(WindowId(140634396691760)), event: KeyboardInput { device_id: DeviceId(DeviceId), event: KeyEvent { physical_key: Code(AltLeft), logical_key: Named(Alt), text: None, location: Left, state: Released, repeat: false, platform_specific: KeyEventExtra { text_with_all_modifiers: None, key_without_modifiers: Named(Alt) } }, is_synthetic: false } }
```

_Alt + 4_ inside Alacritty command line outputs **$**

_Alt + 4_ inside Alacritty+Neovim from Insert mode it switches to Normal mode and then sends the keycode which result in my caret ending up at the end of the line instead of inserting a **$**

_Alt + 4_ inside Alacritty+Tmux+Neovim correctly inserts **$** inside Neovim

I had a similar issue when Alacritty switched from "alt_send_esc" to "option_as_alt", "option_as_alt" = "None" has been working correctly ever since the following issue was filed:
https://github.com/alacritty/alacritty/issues/6637#issuecomment-1420837553

Back then i had to comment out a section that appended an escape sequence/bytes to make it work, after that issue a fix was implemented and it has worked natively ever since then.

But now something has changed (i think it changed after switching to the kitty keyboard protocol?), it is as if the option "None" doesn't work anymore, it seems to have no effect, but i have not tested it out properly, instead; this time i simply commented out code again to get it to work as i expect it to:

https://github.com/alacritty/alacritty/blob/master/alacritty_terminal/src/term/mod.rs#L94-L95

After commenting out/removing:

```
let disambiguate_esc_codes = value.contains(KeyboardModes::DISAMBIGUATE_ESC_CODES);
mode.set(TermMode::DISAMBIGUATE_ESC_CODES, disambiguate_esc_codes);
```

Once again Alacritty works as expected, always inserting a **$** no matter if i am just using Alacritty cmdline, Alacritty + Neovim, Alacritty + Tmux + Neovim

I have only used it for 1 hour so far but so far there has been no negative issues for me without those lines, i am sure there is some unforseen issues i will run into in the future but for now, Alacritty works as expected

So, in latest commit, it seems like **option_as_alt = "None"** isn't respected, or something else is going on?

And it is weird that it works in Alacritty+Tmux+Neovim but not Alacritty+Neovim, i don't know how Tmux is handling events, is it stripping out the escape sequence?

Anyway, with my "fix" i can use Alacritty properly again but it not a nice solution

kchibisov:
By default `Alt` is not a real `Alt` on macOS, so we shouldn't treat it as a modifier.

Fixes #7443.

### DIFF
--- a/alacritty/src/input/keyboard.rs
+++ b/alacritty/src/input/keyboard.rs
@@ -122,6 +122,7 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
                 if named.to_text().is_some() {
                     alt_send_esc
                 } else {
+                    // Treat `Alt` as modifier for named keys without text, like ArrowUp.
                     self.ctx.modifiers().state().alt_key()
                 }
             },
--- a/alacritty/src/input/keyboard.rs
+++ b/alacritty/src/input/keyboard.rs
@@ -78,13 +78,16 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
             return;
         }
 
+        // Mask `Alt` modifier from input when we won't send esc.
+        let mods = if self.alt_send_esc(&key, text) { mods } else { mods & !ModifiersState::ALT };
+
         let build_key_sequence = Self::should_build_sequence(&key, text, mode, mods);
 
         let bytes = if build_key_sequence {
             build_sequence(key, mods, mode)
         } else {
             let mut bytes = Vec::with_capacity(text.len() + 1);
-            if self.alt_send_esc() && text.len() == 1 {
+            if mods.alt_key() {
                 bytes.push(b'\x1b');
             }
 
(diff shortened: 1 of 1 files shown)


alacritty/src/input/keyboard.rs
```diff
@@ -99,6 +102,33 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
...
+                        && self.ctx.modifiers().ralt_state() == ModifiersKeyState::Pressed))
+        };
+
+        match key.logical_key {
+            Key::Named(named) => {
+                if named.to_text().is_some() {
+                    alt_send_esc
+                } else {
+                    self.ctx.modifiers().state().alt_key()
+                }
```
chrisduerr, line 126: What is this branch for? I'd assume it's for keys that have no characters attached, so is it specifically to handle modifier presses themselves? Alt still shouldn't send esc if that's the case on macOS, right?

chrisduerr (changes requested):

alacritty/src/input/keyboard.rs
kchibisov: That's for `Arrow`/`PageUp` keys where `Alt` should actually be `Alt` and not `Option` because they have alternate character we're trying to avoid with all of that mess.

chrisduerr (approved):

chrisduerr (approved):

PR #7450, opened 2023-12-21
Keep IME always enabled on X11

### LINKED ISSUE #7195: Vi Mode Freezes
Vi Mode freezes. I have to close the terminal instance and open a new one.

Two things worth noting in my current setup:
- The mentioned issue occurs intermittently
- The mentioned issue occurs only when alacrity is launched using the shortcut, never when launched from the command line
- I use https://github.com/jeffreytse/zsh-vi-mode, perhaps this is causing contention?

### System

OS: Linux
Version: alacritty 0.13.0-dev (33306142)
X11, ubuntu:GNOME, GNOME Shell

kchibisov:
Fixes #7195.

### DIFF
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -410,7 +410,10 @@ impl Window {
     }
 
     pub fn set_ime_allowed(&self, allowed: bool) {
-        self.window.set_ime_allowed(allowed);
+        // NOTE: don't manipulate X11 IME at runtime since it tends to break with some IMEs.
+        if !self.is_x11 {
+            self.window.set_ime_allowed(allowed);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


alacritty/src/display/window.rs
```diff
@@ -410,7 +410,10 @@ impl Window {
     }
 
     pub fn set_ime_allowed(&self, allowed: bool) {
-        self.window.set_ime_allowed(allowed);
+        // NOTE: don't manipulate X11 IME at runtime since it tends to break with some IMEs.
```
chrisduerr, line 413:
```suggestion
        // Skip runtime IME manipulation on X11 since it breaks some IMEs.
```

chrisduerr (changes requested):

chrisduerr (approved):

PR #7458, opened 2023-12-24
Apply modifiers before presses in kitty protocol

kchibisov:
While this doesn't handle releases with multiple identical modifiers pressed, the release can't work reliable anyway, since one modifier could be pressed before focusing the window, thus tracking modifiers based on the keysym values won't work as it was suggested by kitty author.

Links: https://github.com/kovidgoyal/kitty/issues/6913


chrisduerr (approved):
I hate it.

kchibisov:
>I hate it.

same.

PR #7468, opened 2023-12-28
Derive `Clone` for `EventLoopSender`

HyperHamster:
`.spawn()` consumes `EventLoop` making it impossible to acquire more senders by calling `.channel()` after the fact. It is possible to produce all the senders you would need in advance by calling `.channel()` multiple times before calling `.spawn()` but this is very unidiomatic and would both pollute the namespace and hinder readability. I am proposing this change under the assumption that it is generally always expected that Sender (new)types implement `Clone`.

### DIFF
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -349,6 +349,7 @@ impl event::OnResize for Notifier {
     }
 }
 
+#[derive(Clone)]
 pub struct EventLoopSender {
     sender: Sender<Msg>,
     poller: Arc<polling::Poller>,


chrisduerr:
Curious about your goal with this? Are you trying to write to the PTY from multiple different threads? How are you synchronizing this?

HyperHamster:
This pull request doesn't affect synchronization since you're using `std::sync::mpsc::Sender` and `polling::Poller` which both already correctly handle synchronization across threads internally; Artificially limiting the conditions under which `EventLoopSender` may be cloned does nothing to protect users from synchronization issues.

But to sate your curiosity I have implemented an Alacritty front-end in Discord via a [Serenity](https://github.com/serenity-rs/serenity) bot. This bot handles input from multiple sources and potentially users at once including both raw bytestring input and pre-programmed byte sequences bound to [clickable buttons](https://discord.com/developers/docs/interactions/message-components#buttons) for keys like `Esc` and `Backspace` that cannot be easily typed into a text field by users. The Serenity standard framework is heavily `async` so I rely on being able to clone the `EventLoopSender` (previously `mio_extras::channel::Sender` in `0.19`) associated with a given terminal session as necessary when I need to send input from the `tokio` runtime to the separate thread(s) running the Alacritty event loop providing the bot with terminal emulation.

The code for this is currently private as it in no way secure and only meant to played with in a private server with friends; That and I have no interest in hosting/maintaining a public bot.

kchibisov:
I wonder whether we can plug our event loop into async loop in such a case btw?

I'm fine with `Clone`, but maybe we can do it a bit better? Like inserting `poller` into async runtime should be fine, probably?

kchibisov (approved):

chrisduerr (approved):
Just needs rebase.

PR #7480, opened 2023-12-28
Fix `alacritty migrate` with nonexistent imports

### LINKED ISSUE #7473: alacritty migrate: Configuration file migration failed: parsing error: Unable to locate config file
Fresh after upgrading to alacritty 0.13.0, I saw in the release notes that there's been a migration from yaml to toml. I ran `alacritty migrate` as instructed, but it failed with:
```
Configuration file migration failed:
    "/home/user/.config/alacritty/alacritty.yml": parsing error: Unable to locate config file
```
Obviously, the file is there. There is a parsing error, but the exact cause seems to be misreported.

After playing with it a bit, I found that running `alacritty migrate -i` (skipping imports) succeeds. The only import in my yaml was `colorscheme.yml`. To be honest, I don't remember what exactly that is, but I've decided to create this issue, because `Unable to locate config file` is misleading and maybe this will help other people.

It's unfortunate that `migrate` fails with a vague error on a [former valid config](https://github.com/Randoragon/dotfiles/blob/b45d7534595d6e3cc6550d616294f8e5bb4c49a4/alacritty/.config/alacritty/alacritty.yml). I don't mean this to come off as picky though, thanks for taking the effort of making this migration smooth!

### System

OS: Arch Linux, X11, bspwm, picom

chrisduerr:
Closes #7473.

### DIFF
--- a/alacritty/src/migrate.rs
+++ b/alacritty/src/migrate.rs
@@ -96,7 +96,7 @@ fn migrate_config(
     if options.dry_run && !options.silent {
         // Output new content to STDOUT.
         println!(
-            "\nv-----Start TOML for {path:?}-----v\n\n{toml}\n^-----End TOML for {path:?}-----^"
+            "\nv-----Start TOML for {path:?}-----v\n\n{toml}\n^-----End TOML for {path:?}-----^\n"
         );
     } else if !options.dry_run {
         // Write the new toml configuration.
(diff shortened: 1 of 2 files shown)


alacritty/src/migrate.rs
```diff
@@ -124,7 +124,23 @@ fn migrate_imports(
...
+            new_imports.push(Value::String(import.to_string_lossy().into()));
+            continue;
+        }
+
         let new_path = migrate_config(options, &import, recursion_limit - 1)?;
+
+        // Print new import path.
+        if options.dry_run {
+            println!("Successfully migrated import {import:?} to {new_path:?}");
+        }
```
chrisduerr, line 142: Just added this because it seems like a good idea to have.

kchibisov: Do we need it only for `dry-run` though?

chrisduerr: Otherwise it's printed twice. Since every wet run includes a dry run, we only log stuff during the dry run.

kchibisov (commented):
probably a changelog entry needed?

kchibisov (approved):

kchibisov (approved):

PR #7489, opened 2023-12-29
Passthrough potential errors for `EventLoopSender`

HyperHamster:
For this change I'm not sure if the error enum should be made `#[non_exhaustive]` or not, I'm simply not familiar enough with the codebase or your plans for the future to make that determination myself. I think it's fine that you chose to ignore these errors internally but it felt like bad practice to not pass through potential errors on a `pub` method.

### DIFF
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -2,7 +2,7 @@
 
 use std::borrow::Cow;
 use std::collections::VecDeque;
-use std::fmt::{self, Formatter, Display};
+use std::fmt::{self, Display, Formatter};
 use std::fs::File;
 use std::io::{self, ErrorKind, Read, Write};
 use std::marker::Send;
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -2,6 +2,7 @@
 
 use std::borrow::Cow;
 use std::collections::VecDeque;
+use std::fmt::{self, Formatter};
 use std::fs::File;
 use std::io::{self, ErrorKind, Read, Write};
 use std::marker::Send;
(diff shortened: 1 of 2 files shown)


alacritty_terminal/src/event_loop.rs
```diff
@@ -339,13 +340,37 @@ impl event::Notify for Notifier {
...
+            EventLoopSendError::Send(x) => x.fmt(f),
+        }
+    }
+}
+
+impl std::error::Error for EventLoopSendError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            EventLoopSendError::Io(x) => Some(x),
+            EventLoopSendError::Send(x) => Some(x),
```
chrisduerr, line 372: `x` is not a good variable name.

alacritty_terminal/src/event_loop.rs
```diff
@@ -339,13 +340,37 @@ impl event::Notify for Notifier {
...
+pub enum EventLoopSendError {
+    Io(io::Error),
+    Send(mpsc::SendError<Msg>),
+}
+
+impl Display for EventLoopSendError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            EventLoopSendError::Io(x) => x.fmt(f),
+            EventLoopSendError::Send(x) => x.fmt(f),
```
chrisduerr, line 363: `x` is not a good variable name.

chrisduerr (changes requested):
I don't really mind this, but handling these errors is rarely ever useful.

alacritty_terminal/src/event_loop.rs
HyperHamster: Renamed to `err` to conform with other `std::error::Error` `impl`s in the project.

alacritty_terminal/src/event_loop.rs
HyperHamster: Renamed to `err` to conform with other `std::fmt::Display` `impl`s in the project.

kchibisov (approved):

chrisduerr (approved):

PR #7493, opened 2023-12-29
Send associated text for shifted numbers with kitty

### LINKED ISSUE #7492: Certain modifiers fail in `kakoune`
I found at least two type of modifiers that stopped working on `0.13` that worked before on `0.12`.

-  `ctrl + [` does not trigger `ESC`-equivalent behaviour in `kakoune`.
    - Interestingly enough the same action continues to work in `vim`?
- `alt + shift` combinations fail to register, so I checked eg. `alt + shift + !` (append command output), `alt + K` (keep not matching), `alt + N` (extend reverse search).

I assume these issues have a wider impact than just `kakoune` but I found it the easiest to reproduce it there. It possibly might be a duplicate considering a wave of keyboard input related issues lately but none so far seems to match exactly that. Sorry in advance.

### System

OS: Arch Linux
Linux/BSD: Wayland, sway

### Logs

This should be quite trivial to reproduce, can provide logs if necessary though.

Thanks a lot for looking into it. Fantastic project and amazing work, lots of love <3

kchibisov:
Fixes #7492.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `Alt` bindings requiring composed key rather than pre-composed one on macOS
 - `Alt + Control` bindings not working on Windows
 - `chars = "\u000A"` action in bindings inserting `\n`
+- Alternate keys not sent for `Shift + <number>` when using kitty protocol
+- Alternative keys being swapped in kitty protocol implementation
 
 ## 0.13.0
 
--- a/alacritty/src/input/keyboard.rs
+++ b/alacritty/src/input/keyboard.rs
@@ -350,17 +350,25 @@ impl SequenceBuilder {
             let character = character.chars().next().unwrap();
             let base_character = character.to_lowercase().next().unwrap();
 
-            let codepoint = u32::from(character);
-            let base_codepoint = u32::from(base_character);
+            let alternate_key_code = u32::from(character);
+            let mut unicode_key_code = u32::from(base_character);
+
+            // Try to get the base for keys which change based on modifier, like `1` for `!`.
+            match key.key_without_modifiers().as_ref() {
+                Key::Character(unmodded) if alternate_key_code == unicode_key_code => {
+                    unicode_key_code = u32::from(unmodded.chars().next().unwrap_or(base_character));
+                },
+                _ => (),
+            }
 
             // NOTE: Base layouts are ignored, since winit doesn't expose this information
             // yet.
             let payload = if self.mode.contains(TermMode::REPORT_ALTERNATE_KEYS)
-                && codepoint != base_codepoint
+                && alternate_key_code != unicode_key_code
             {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - `alacritty migrate` failing with nonexistent imports
+- Associated text not send for `Shift + <number>` when using kitty protocol
```
chrisduerr, line 13:
```suggestion
- Associated text not sent for `Shift + <number>` when using kitty protocol
```

alacritty/src/input/keyboard.rs
```diff
@@ -338,7 +338,15 @@ impl SequenceBuilder {
             let base_character = character.to_lowercase().next().unwrap();
 
             let codepoint = u32::from(character);
-            let base_codepoint = u32::from(base_character);
+            let mut base_codepoint = u32::from(base_character);
+
+            // Try to get the base for keys which change based on modifier, like `1` for `!`.
+            match key.key_without_modifiers().as_ref() {
+                Key::Character(unmodded) if codepoint == base_codepoint => {
```
chrisduerr, line 345: This might be cleaner with `matches!`?

chrisduerr (changes requested):

alacritty/src/input/keyboard.rs
kchibisov: you can't use matches here from what I can see, since you need to `bind`.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - `alacritty migrate` failing with nonexistent imports
+- Associated text not sent for `Shift + <number>` when using kitty protocol
```
glaeqen, line 13: Is it an accurate changelog entry? It seems to fix any `shift + alt + <char>` key combination? Just making sure that the changelog entry is as precise as possible 👍🏻

kchibisov: Yeah, I'd need one more entry.

alacritty/src/input/keyboard.rs
```diff
@@ -337,17 +337,25 @@ impl SequenceBuilder {
...
             // yet.
             let payload = if self.mode.contains(TermMode::REPORT_ALTERNATE_KEYS)
-                && codepoint != base_codepoint
+                && alternate_key_code != unicode_key_code
             {
-                format!("{codepoint}:{base_codepoint}")
+                format!("{unicode_key_code}:{alternate_key_code}")
             } else {
-                codepoint.to_string()
+                alternate_key_code.to_string()
```
kchibisov, line 358: Naming is not what this protocol is good at with this thing... I used the words they use so it's easier to at least match with the docs.

chrisduerr (approved):

PR #7515, opened 2023-12-31
Make pre-0.13.0 configuration a note in README

kchibisov:
Rendered

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,10 +55,6 @@ For everyone else, the detailed instructions to install Alacritty can be found
 
 ## Configuration
 
-**Alacritty versions before 0.13.0 are using YAML configuration files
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
What does this contribute? I don't see how this makes anything better and I'm not sure what this proprietary markdown looks like on crates.io for example.

kchibisov:
Hm, that's fair, it probably look broken on crates.io.

Maybe we should just move it below, since 0.13.0 is out?

chrisduerr:
> Maybe we should just move it below, since 0.13.0 is out?

Sure, I'm also fine with just removing it.

kchibisov:
I've just removed it.

chrisduerr (approved):

PR #7529, opened 2024-01-02
Mouse bindings with numbers don't work in 0.13.0

### LINKED ISSUE #7527: Mouse bindings with numbers don't work as of 0.13
I upgraded to 0.13, ran `migrate` and am greeted with an error message:

```
[521.148322971s] [ERROR] [alacritty_config_derive] Config error: invalid type: integer `9`, expected Left, Right, Middle, Back, Forward, or a number from 0 to 65536
                                                   in `mouse`
                                                   ; ignoring binding
```

This is my binding:
```
[[mouse.bindings]]
action = "DecreaseFontSize"
mods = "Control"
mouse = 8
```

I narrowed the problem down to the deserializer which "only" contains `visit_str` and `visit_u64`

https://github.com/alacritty/alacritty/blob/659550ee34fe1b45ea176b88e60901d277d210a3/alacritty/src/config/bindings.rs#L849-L890

I'm just going to assume that this worked with `serde_json` (which I assume was used before) but for toml the numbers seem to be interpreted as signed by default.

Implementing a `visit_i64` method makes this work:

```rust
            fn visit_i64<E>(self, value: i64) -> Result<MouseButtonWrapper, E>
            where
                E: de::Error,
            {
                match value {
                    0..=65536 => Ok(MouseButtonWrapper(MouseButton::Other(value as u16))),
                    _ => Err(E::invalid_value(Unexpected::Signed(value), &self)),
                }
            }
```

I can submit a PR if this is an acceptable solution.

### System

OS: Linux (Arch)
Linux/BSD: X11/bspwm

lfrancke:
not having a visitor for this would mean numbers for mouse buttons wouldn't work

/fixes #7527

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Blurry window when using `window.dimensions` on some Wayland compositors
 - IME input lagging behind on X11
 - xdotool modifiers input not working correctly on X11
+- Parsing numbers fails for mouse bindings
 
 ## 0.13.0
 
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -860,6 +860,16 @@ impl<'a> Deserialize<'a> for MouseButtonWrapper {
                 f.write_str("Left, Right, Middle, Back, Forward, or a number from 0 to 65536")
             }
 
+            fn visit_i64<E>(self, value: i64) -> Result<MouseButtonWrapper, E>
+            where
+                E: de::Error,
+            {
+                match value {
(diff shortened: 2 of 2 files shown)


chrisduerr:
Interesting, it seems like this might be due to the TOML change. Is the `visit_u64` method still necessary?

lfrancke:
It isn't being called anymore, no. But it doesn't hurt either should the behavior change upstream or should you ever switch to a different toml library with slightly different behavior.

I could add a comment why both are implemented or remove one.
I'd leave it in but it's up to you.

lfrancke:
The toml spec says

> Arbitrary 64-bit signed integers (from −2^63 to 2^63−1) should be accepted and handled losslessly. If an integer cannot be represented losslessly, an error must be thrown.

https://toml.io/en/v1.0.0#integer

chrisduerr (approved):
Should be fine, I was just curious.

PR #7536, opened 2024-01-02
Ignore null values in `alacritty migrate`

kchibisov:
This should help with broken YAML configurations by throwing nulls away, which are not representable in toml.


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.14.0-dev
 
+### Changed
+
+- `alacritty migrate` will ignore null mappings in yaml instead of erroring out.
```
chrisduerr, line 12:
```suggestion
- `alacritty migrate` will ignore null values in yaml instead of erroring out
```

alacritty/src/config/mod.rs
```diff
@@ -230,7 +230,8 @@ pub fn deserialize_config(path: &Path) -> Result<Value> {
             "YAML config {path:?} is deprecated, please migrate to TOML using `alacritty migrate`"
         );
 
-        let value: serde_yaml::Value = serde_yaml::from_str(&contents)?;
+        let mut value: serde_yaml::Value = serde_yaml::from_str(&contents)?;
+        prune_yaml_nulls(&mut value);
```
chrisduerr, line 234: Won't this also remove nulls which are recognized as error already?

alacritty/src/config/mod.rs
```diff
@@ -318,6 +319,26 @@ pub fn imports(
...
+            sequence.is_empty()
+        } else {
+            value.is_null()
+        }
+    }
+
+    if walk(value) {
+        // When the value itself is null return the mapping.
+        *value = serde_yaml::Value::Mapping(Default::default());
+    }
```
chrisduerr, line 339: This seems extremely hacky for effectively not much benefit.

chrisduerr (changes requested):

alacritty/src/config/mod.rs
kchibisov: This will only remove unit types, I'm not sure you can craft `nul` as error.

alacritty/src/config/mod.rs
kchibisov: That's the only way to translate the empty config though.

alacritty/src/config/mod.rs
chrisduerr: If you do `window:` it will have a unit type and be an error.

alacritty/src/config/mod.rs
chrisduerr: We could always just not and force users to actually fix their yaml first.

kchibisov: Yeah, but empty config is a valid toml config as well, so it seems like our bug.

alacritty/src/config/mod.rs
kchibisov: But do we care? Like removing it during translation is better than forcing users to figure out what to actually fix, because there's no clear indication on what is being broken, and they could have an error before for that, no?

Like I don't really want to write code based on broken config that throws an error before, I write it based on the configs that worked before but not anymore.

alacritty/src/config/mod.rs
chrisduerr: Yeah that's fair, I'm just not sure about this solution in particular. And removing stuff without even printing a log warning definitely shouldn't be happening I think.

kchibisov: I can log what I remove though, but null itself is empty, so there's nothing to print, since nothing to remove. it's just a conversion is weird in that case.

chrisduerr: Yeah with empty config you wouldn't want to log obviously.

alacritty/src/config/mod.rs
```diff
@@ -318,6 +319,32 @@ pub fn imports(
     Ok(import_paths)
 }
 
+/// Prune the nulls from the YAML.
```
chrisduerr, line 322: Maybe we should add a comment *why* one would want to do this?

So something like "to ensure toml compatibility" or something.

alacritty/src/config/mod.rs
```diff
@@ -318,6 +319,32 @@ pub fn imports(
...
+fn prune_yaml_nulls(value: &mut serde_yaml::Value, warn_pruned: bool) {
+    fn walk(value: &mut serde_yaml::Value, warn_pruned: bool) -> bool {
+        if let Some(mapping) = value.as_mapping_mut() {
+            mapping.retain(|key, value| {
+                let retain = !walk(value, warn_pruned);
```
chrisduerr, line 327: This is recursive, right? There doesn't seem a limit to this, so I could probably craft a config that would run into a stack overflow here?

alacritty/src/migrate.rs
```diff
@@ -60,6 +62,7 @@ fn migrate_config(
     options: &MigrateOptions,
     path: &Path,
     recursion_limit: usize,
+    warn_pruned: bool,
```
chrisduerr, line 65: Rather than adding another option here, we should probably just use the `options.dry_run` option to ensure consistency with other code. That way we also make sure to **always** warn on the dry run, which ensures that even on failure everything is logged.

chrisduerr (changes requested):

alacritty/src/config/mod.rs
kchibisov: Unless the config itself is recursive you should not run into limit? I can limit it just in case if you feel afraid of.

alacritty/src/config/mod.rs
```diff
@@ -318,6 +322,36 @@ pub fn imports(
...
+                }
+                retain
+            });
+            mapping.is_empty()
+        } else if let Some(sequence) = value.as_sequence_mut() {
+            sequence.retain_mut(|value| !walk(value, depth + 1, warn_pruned));
+            sequence.is_empty()
+        } else {
+            value.is_null()
+        }
```
chrisduerr, line 346: Should probably use a `match` statement here.

chrisduerr (changes requested):

chrisduerr (approved):

PR #7537, opened 2024-01-02
Tell what to do with unused keys

kchibisov:
--

¯\_(ツ)_/¯

### DIFF
--- a/alacritty_config_derive/src/config_deserialize/de_struct.rs
+++ b/alacritty_config_derive/src/config_deserialize/de_struct.rs
@@ -58,7 +58,8 @@ pub fn derive_deserialize<T>(
 
                 // Warn about unused keys.
                 for key in unused.keys() {
-                    log::warn!(target: #LOG_TARGET, "Unused config key: {}", key);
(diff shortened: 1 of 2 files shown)


chrisduerr (changes requested):
I don't think anyone is having struggles understanding that part? This just bloats up the message bar.

kchibisov:
>I don't think anyone is having struggles understanding that part? This just bloats up the message bar.

There were a few where people were not sure what to do. Some thought that they haven't added something.

chrisduerr:
> There were a few where people were not sure what to do. Some thought that they haven't added something.

Yeah but I think that's mostly confusion about getting an error when they didn't before, not that they're confused about what to do.

chrisduerr:
> it's a message bar you see pretty rare

Especially right now people are **constantly** running into this. That's the only reason why you changed it in the first place. So our goal should be to actually be constructive.

kchibisov:
yeah, and there's clear a "don't know what to do" situation, because they never get a warning for unused key before.

PR #7538, opened 2024-01-02
Fix replacing optional fields

### LINKED ISSUE #7518: Setting window position from CLI works only for one axis
Upgrading from alacritty 0.12 I found myself unable to set *both* `x` and `y` positions from CLI, but I can set either *one*. Examples:

```
# This command applies only `x=444` option and ignores `y=333` option:
alacritty -o window.position.y=333 window.position.x=444

# This command applies only `y=333` option and ignores `x=444` option:
alacritty -o window.position.x=444 window.position.y=333
```
It feels like a latter position drops the former one.

Interestingly, if I set both `x` and `y` in the alacritty.toml and run alacritty with neither `x` nor `y` specified from the CLI, then both positions get applied correctly.

https://github.com/alacritty/alacritty/issues/7267 is possibly related.

### System

OS: Debian
Version: alacritty 0.13.0
Linux: X11, BSPWM

chrisduerr:
This fixes an issue with the default `SerdeReplace` implementation where it would never recurse through options but always replace the entire option with the new value.

Closes #7518.

### DIFF
--- a/alacritty_config/src/lib.rs
+++ b/alacritty_config/src/lib.rs
@@ -3,7 +3,6 @@ use std::error::Error;
 use std::path::PathBuf;
 
 use log::LevelFilter;
-use winit::platform::wayland::Theme;
 use serde::Deserialize;
 use toml::Value;
 
@@ -34,7 +33,6 @@ impl_replace!(
     String,
     PathBuf,
     LevelFilter,
-    Theme,
 );
 
 fn replace_simple<'de, D>(data: &mut D, value: Value) -> Result<(), Box<dyn Error>>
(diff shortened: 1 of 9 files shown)


alacritty_config/src/lib.rs
```diff
@@ -1,14 +1,17 @@
 use std::collections::HashMap;
 use std::error::Error;
+use std::path::PathBuf;
 
 use log::LevelFilter;
+use winit::platform::wayland::Theme;
```
kchibisov, line 6:
```suggestion
use winit::window::Theme;
```

kchibisov (approved):

PR #7595, opened 2024-01-09
Fix `debug.renderer="gles2pure"` documentation

chrisduerr:
Patch 5685ce8bf changed the allowed values of the `debug.renderer` enum, prohibiting the usage of `_` in the `Gles2Pure` variant. This patch updates the documentation to correct for that.

### DIFF
--- a/extra/man/alacritty.5.scd
+++ b/extra/man/alacritty.5.scd
@@ -996,7 +996,7 @@ relied upon.
 	Example:
 		_ALACRITTY_EXTRA_LOG_TARGETS="winit;vte" alacritty -vvv_
 
-*renderer* = _"glsl3"_ | _"gles2"_ | _"gles2_pure"_ | _"None"_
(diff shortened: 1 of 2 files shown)


extra/man/alacritty.5.scd
```diff
@@ -996,7 +996,7 @@ relied upon.
 	Example:
 		_ALACRITTY_EXTRA_LOG_TARGETS="winit;vte" alacritty -vvv_
 
-*renderer* = _"glsl3"_ | _"gles2"_ | _"gles2_pure"_ | _"None"_
+*renderer* = _"glsl3"_ | _"gles2"_ | _"gles2pure"_ | _"None"_
```
kchibisov, line 999: can't we bring back how it was?

chrisduerr: We could, but that would require proc macro changes. Not sure that's worth it for a debug option?

kchibisov (approved):

PR #7601, opened 2024-01-10
Allow specifying all config keys on all platforms

### LINKED ISSUE #7592: Warnings when using OS-specific keys on alternative OSes
### System

OS: Linux/macOS
Version: alacritty 0.14.0-dev
Linux/BSD: X11/i3WM

I share my Alacritty config between my work Mac and my personal laptop. I've recently migrated my config to TOML and I now see a warning about unused keys for `option_as_alt`. This is a valid key on Mac OS, but not on Linux, so I'm receiving the warning only on Linux.

Is there a way to suppress this warning?

chrisduerr:
Closes #7592.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - CLI env variables clearing configuration file variables
 - Vi inline search/semantic selection expanding across newlines
 
+### Changed
+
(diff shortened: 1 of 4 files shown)


kchibisov:
@chrisduerr won't we warn on bindings as well, like we have more specifics there?

chrisduerr:
> won't we warn on bindings as well, like we have more specifics there?

Oh you mean platform-specific binding actions? I forgot about those.

kchibisov (approved):
Will need a rebase.

PR #7610, opened 2024-01-11
ALT + arrows default key binding

smasyutin:
ALT/Option + arrows is the default behavior of MacOS Terminal, kitty, and VSCode terminal: jump by word-by-word instead of moving char-by-char, which is extremely handy when editing a (long) command.

With YAML config it was possible as [described here](https://stackoverflow.com/questions/58516583/move-cursor-word-by-word-in-alacritty) :

```
key_bindings:
  ...
  - { key: Right, mods: Alt, chars: "\x1BF" }
  - { key: Left,  mods: Alt, chars: "\x1BB" }
```

However, with the TOML configuration of Alactritty it is not possible to get this behavior, thus I propose to add it as default key bindings.

PS: this is my first PR to alacritty and the first in Rust, so I'm open to any suggestions to fix/update/make it better.

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -452,6 +452,8 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
         End,       ModifiersState::SHIFT, ~BindingMode::ALT_SCREEN; Action::ScrollToBottom;
         PageUp,    ModifiersState::SHIFT, ~BindingMode::ALT_SCREEN; Action::ScrollPageUp;
         PageDown,  ModifiersState::SHIFT, ~BindingMode::ALT_SCREEN; Action::ScrollPageDown;
+        ArrowRight, ModifiersState::ALT; Action::Esc("\x1bF".into());
+        ArrowLeft,  ModifiersState::ALT; Action::Esc("\x1bB".into());
         // App cursor mode.
         Home,       +BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1bOH".into());
         End,        +BindingMode::APP_CURSOR, ~BindingMode::VI, ~BindingMode::SEARCH; Action::Esc("\x1bOF".into());


kchibisov:
>However, with the TOML configuration of Alactritty it is not possible to get this behavior, thus I propose to add it as default key bindings.

Everything is possible, use `\U001BF` and `\U` notation.

The `Alt + Left` and `Alt + Right` should use `^[[1;3D^[[1;3C` respectively, which they use.

smasyutin:
@kchibisov thanks for the tip. It works. However, in macOS Terminal, kitty, VSCode it works out-of-box (surprisingly, it is doesn't in iTerm2)

Why couldn't we have it at alacritty _by default_?

kchibisov:
The default bindings are the ones I said. If macOS developed their own in-house things it's up to macOS, I'm not going to adopt macOS specific bindings, the standard encoding is what I said.

smasyutin:
@kchibisov apologies for not being entirely clear. The config changes you suggest do work as expected on macOS. Thanks again.

```
{ key = "ArrowRight", mods = "Alt", chars = "\u001BF" },
{ key = "ArrowLeft",  mods ="Alt", chars ="\u001BB" }
```

It is not about macOS, it is about Alacritty macOS **users** who choose between different options on the market and try to find whatever matches their needs the best. Right now it is **additional** configuration that users need to apply, which is **on** _by default_ for majority of competitors. Thus my suggestion is still to accept my PR (or any variation of it) and make it _by default_ with no additional user configuration, so that Alacritty gets feature parity with the competitors.

PS: sorry for the long discussion

kchibisov:
>majority of competitors

You have wrong mindset, we're not competing with anyone. It's a free software.

> Thus my suggestion is still to accept my PR (or any variation of it) and make it by default with no additional user configuration, so that Alacritty gets feature parity with the competitors.

Your PR lacks _technical_ motivation on why it's needed, these keys are not standard. On what is more or less standard see https://invisible-island.net/xterm/ctlseqs/ctlseqs.html what you suggest is not. You don't even need alacritty bindings for what you want, just bind standard keys in the `shell` you're using.

Like 
```zsh
bindkey '^[[1;3C' forward-word
bindkey '^[[1;3D' backward-word
```

in zsh, will work in every terminal, and cross platform, because the sequences are also cross platform.

smasyutin:
> > majority of competitors
> 
> You have wrong mindset, we're not competing with anyone. It's a free software.

You barely know me, the best practice is to keep your hypothesis about my personality outside of the discussion.

If something is free, it doesn't mean it has no competitors. If you do not care about users, why open source? If you do care, why wouldn't you listen to them?

> 
> > Thus my suggestion is still to accept my PR (or any variation of it) and make it by default with no additional user configuration, so that Alacritty gets feature parity with the competitors.
> 
> Your PR lacks _technical_ motivation on why it's needed, these keys are not standard. On what is more or less standard see https://invisible-island.net/xterm/ctlseqs/ctlseqs.html what you suggest is not. You don't even need alacritty bindings for what you want, just bind standard keys in the `shell` you're using.
> 

These keys are standard for Terminal.app, kitty and terminal within VSCode.

> Like
> 
> ```shell
> bindkey '^[[1;3C' forward-word
> bindkey '^[[1;3D' backward-word
> ```
> 
> in zsh, will work in every terminal, and cross platform, because the sequences are also cross platform.

Thanks for another tip how to make it work without changing Alacritty. I noted and applied it. It works. Thanks.

Still, it is **very** different from user experience with Terminal.app, kitty and terminal within VSCode.

kchibisov:
>If something is free, it doesn't mean it has no competitors. If you do not care about users, why open source? If you do care, why wouldn't you listen to them?

You should say _alternative_ not competitors we don't compete with anyone, that's the key thing here. We accept patches, but there's a line, like you should provide a justification for what you send, and it's like that in general with any software. Our goal is not to please every user by adding every possible feature.

>These keys are standard for Terminal.app, kitty and terminal within VSCode.

Being standard because `Terminal.app` created them? We have a clear encoding table how to encode arrow keys, we encode them as they should be encoded, you suggest encoding that is developed by `Terminal.app` devs, why should we care about it in the first place where we have standard? Maybe bug apple to use the standard encoding schema? By they surely care since last time they bothered to update ncurses was > 10 years ago for example.

>Still, it is very different from user experience with Terminal.app, kitty and terminal within VSCode.

You're free to use alternatives if you don't like particular software.

smasyutin:
> > If something is free, it doesn't mean it has no competitors. If you do not care about users, why open source? If you do care, why wouldn't you listen to them?
> 
> You should say _alternative_ not competitors we don't compete with anyone, that's the key thing here. We accept patches, but there's a line, like you should provide a justification for what you send, and it's like that in general with any software. Our goal is not to please every user by adding every possible feature.

OK, I got it. So my justification that "other software is easier to use and switch to" is not credible enough, right? Btw, if you could state what the goal is, it would help.

> 
> > These keys are standard for Terminal.app, kitty and terminal within VSCode.
> 
> Being standard because `Terminal.app` created them? We have a clear encoding table how to encode arrow keys, we encode them as they should be encoded, you suggest encoding that is developed by `Terminal.app` devs, why should we care about it in the first place where we have standard? Maybe bug apple to use the standard encoding schema? By they surely care since last time they bothered to update ncurses was > 10 years ago for example.

I will repeat once again that **3 (three) other alternatives do that**. Not all of them are macOS specific. So please stop your bias towards Apple. It's more than that. Beyond that, if I want to switch from another alternative -- it is harder than expected, thus more chances I will stay where I am. So less chances Alacritty will get one more user. So back to "why open source?" question. Btw, this is my second attempt to switch, still I'm hesitant.
 
> > Still, it is very different from user experience with Terminal.app, kitty and terminal within VSCode.
> 
> You're free to use alternatives if you don't like particular software.

Based on this PR & discussion 

I do like 
- Alactritty software, when it works
- technical information about ways to make things work w/o changing Alactritty

I do not like
- judgmental way maintainers communicate
- unclarity about the philosophy of adding new features

I can use alternatives, I can build on my own fork. Still, the reason I created this PR and continue this discussion is that I'd like to make Alacritty better and more accessible software.

kchibisov:
>I will repeat once again that 3 (three) other alternatives do that. Not all of them are macOS specific. So please stop your bias towards Apple. It's more than that. Beyond that, if I want to switch from another alternative -- it is harder than expected, thus more chances I will stay where I am. So less chances Alacritty will get one more user. So back to "why open source?" question. Btw, this is my second attempt to switch, still I'm hesitant.

Please, show me what these escape mean and how the said encoding is done. You're aware that they only work because apple decided to add bindings like that in the system zsh configuration similarly the way I told you to bind keys?

>Not all of them are macOS specific

They could add specific stuff for macOS. Again, by adding these bindings you're breaking existing clients which **rely** on standard keys, which we **send**.

>unclarity about the philosophy of adding new features

We add standard documented escape sequences, we don't add something in-house developed. If you can show why these escapes escape in the way you propose to add them for said `Alt` keys, please, let me know. I've linked you an xterm table because it lists all common encodings, there's https://sw.kovidgoyal.net/kitty/keyboard-protocol which we support and this is also not part of it, there's even http://www.leonerd.org.uk/hacks/fixterms/ which is also says nothing about what you add, so from my point of view it's just something apple developed and everyone adopted because users complained that keys don't work.

There're similar escape sequences specific to `Terminal.app` (so called proprietary sequences) which work because apple is the one configuring the terminal globally (This is behavior is not apple specific, it's also done by e.g. fedora which sets vte specific escapes by checking `VTE_VERSION`).

smasyutin:
> > I will repeat once again that 3 (three) other alternatives do that. Not all of them are macOS specific. So please stop your bias towards Apple. It's more than that. Beyond that, if I want to switch from another alternative -- it is harder than expected, thus more chances I will stay where I am. So less chances Alacritty will get one more user. So back to "why open source?" question. Btw, this is my second attempt to switch, still I'm hesitant.
> 
> Please, show me what these escape mean and how the said encoding is done. 

I'm not sure I get the question.

> You're aware that they only work because apple decided to add bindings like that in the system zsh configuration similarly the way I told you to bind keys?

Hm, I was not aware about it. However, on macOS Sonoma that I use `/etc/zshrc` has some bindings (see below), but none for `forward-word` & `backward-word`. I was not able to find that `forward/backward-word` are defined anywhere. Not in a `/etc/zshrc_Apple_Terminal` either. So, I'd still assume that those are built-in into terminal apps.

```

# Default key bindings
[[ -n ${key[Delete]} ]] && bindkey "${key[Delete]}" delete-char
[[ -n ${key[Home]} ]] && bindkey "${key[Home]}" beginning-of-line
[[ -n ${key[End]} ]] && bindkey "${key[End]}" end-of-line
[[ -n ${key[Up]} ]] && bindkey "${key[Up]}" up-line-or-search
[[ -n ${key[Down]} ]] && bindkey "${key[Down]}" down-line-or-search

# Default prompt
```

The opinion that terminal app could behave _the same for every shell and machine_ is very relevant for IT & DevOps folks. In case bindings are a part of the terminal app allows to have same key bindings regardless of the machine I connect to and default shell set up. Alacritty approach could be different. 

> 
> > Not all of them are macOS specific
> 
> They could add specific stuff for macOS. Again, by adding these bindings you're breaking existing clients which **rely** on standard keys, which we **send**.
> 
> > unclarity about the philosophy of adding new features
> 
> We add standard documented escape sequences, we don't add something in-house developed. If you can show why these escapes escape in the way you propose to add them for said `Alt` keys, please, let me know. I've linked you an xterm table because it lists all common encodings, there's https://sw.kovidgoyal.net/kitty/keyboard-protocol which we support and this is also not part of it, there's even http://www.leonerd.org.uk/hacks/fixterms/ which is also says nothing about what you add, so from my point of view it's just something apple developed and everyone adopted because users complained that keys don't work.
> 
> There're similar escape sequences specific to `Terminal.app` (so called proprietary sequences) which work because apple is the one configuring the terminal globally (This is behavior is not apple specific, it's also done by e.g. fedora which sets vte specific escapes by checking `VTE_VERSION`).

I was not supposed to make any changes to keyboard protocol. The intent was to get key bindings that are convenient and widely adopted (I will not call them standard any more). However, I see that I put them in line with all the bindings that you put for specific Alacritty modes. So I should agree, my suggestions are off here.

To close the discussion: 
-- I got that providing any key bindings that compliment / override user shell is not in Alacritty approach.
-- I learned a lot about terminal apps and shell from this discussion. Thanks for your patience and sharing all the details.

kchibisov:
>I'm not sure I get the question.

The current arrow keys you're seeing are defined by the xterm which has them from the original VT terms. If go on the link I've posted multiple times https://invisible-island.net/xterm/ctlseqs/ctlseqs.html You can spot the following table

image

In PC keys section, it explicitly states how all the keys must be encoded. A lot of applications, like tmux, vim, nvim, use this encoding and not the one developed by apple for this specific case.

I've did some research on what the meaning is and it's just default `zsh`/`bash` bindings. What is equivalent in is `Alt + b`/`Alt + f` which you may use if you have `option_as_alt` defined.

I've also checked iterm2 and it behaves the same as alacritty (though, you said that, I just had it around). So a cross platform way to do things would be press `Alt + f` and `Alt + b`, sending them for arrows doesn't make any sense.

>I'd still assume that those are built-in into terminal apps.

Terminal just sends characters, shell (e.g. zsh) interprets them, some terminals make your shell preload arbitrary config files (so called shell integration feature).

>The opinion that terminal app could behave the same for every shell and machine is very relevant for IT & DevOps folks. In case bindings are a part of the terminal app allows to have same key bindings regardless of the machine I connect to and default shell set up. Alacritty approach could be different.

Bindings to navigate a prompt are part of the shell, not terminal. Terminals must just send the correct encoding, which we do, so adding incorrect encoding you want doesn't make any sense.

PR #7616, opened 2024-01-13
Only keep "compat" SVG icons

proski:
The old SVG icons show a white triangle in KDE where it's not supposed to be. The difference in the look of the icons is negligible. Move `compat` icons over the default icons.

--

Observed on Fedora 39 Asahi on aarch64, all packages are up to date. In fact, even the GIMP open dialog shows that white triangle in preview.

I believe the fix belongs to Alacritty. I don't feel it would be reasonable to ask KDE developers to use more advanced SVG libraries for icons. Asking a distro to use the compat icons would only fix the issue for that particular distro.

I haven't touched the PNG images, they remain in `compat`. They could probably be removed. Programs to generate PNG from SVG are ubiquitous.


kchibisov:
@proski in general if the distro uses `KDE` as a main option I'd suggest them to use `compat` icons. Like I use compat icon on gentoo because there're a lot of people using KDE.

The white triangle is QT bug and should be fixed.

kchibisov:
>I haven't touched the PNG images, they remain in compat. They could probably be removed. Programs to generate PNG from SVG are ubiquitous

png is required as embeded icon iirc.

kchibisov:
distributions could ship compat icons, but generally this Qt bug is open for a long time. Like the bug is in Qt itself, it's not like they use simple library. Keep in mind that even picosvg can handle it.

PR #7642, opened 2024-01-19
Update README.md

eddiehazel:
tabs are supported

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ usecases.
 
 Alacritty has many great features, but not every feature from every other
 terminal. This could be for a number of reasons, but sometimes it's just not a
-good fit for Alacritty. This means you won't find things like tabs or splits
+good fit for Alacritty. This means you won't find things like splits
 (which are best left to a window manager or [terminal multiplexer][tmux]) nor
 niceties like a GUI config editor.
 


kchibisov:
tabs are supported only on platforms where window manager provides them. We don't implement them ourselves.

eddiehazel:
is there any way to get splits in the windows on the the macos?

i love it because it is fast and rust but i can't use it because i can't cope with using tmux the whole while

sometimes i love to party like it's 1989 but i can't work like that 😂

kchibisov:
There're tiling window managers for macOS, so you have splits everywhere.

eddiehazel:
thank you for the advice! congratulations on the invention of there're - from one wordsmith to another

i am guessing you are not the macos user so i will mention albeit unsolicited that the said tiling manager are not the best and the fact that macos in in 'flux' between the fullscreen version and the prexisting version that amethyst leverages mean at present Allacritty is not viable on macos at least for me for this reason

more info... https://chrishannah.me/using-a-tiling-window-manager-on-macos

thanks 😌

kchibisov:
I used https://github.com/koekeishiya/yabai iirc last time I had to use macOS. But generally a lot of editors do tiling as well (e.g. vim).

eddiehazel:
"had to" 😂 

yeah i know the answer ultimately is "use a real OS" 😁 

i am definitely the atypical case and i understand completely the need to be stringent to discourage the scope creep

having said that i am "stuck" in MacOS as i am sure many others are and this breaks both the usage paradigms so it makes this software unusable which is sad. the yabai thing is a turn off because it is a hack and difficult to implement and vim or similar suffers from second class keyboard shortcuts. we are supposed to be avoiding spending an entire week configuring the perfect OS over here : )

so certainly not trying to convince you but i am also quite interested in  how those splits would be added

i am in the middle of learning rust (WHICH I LOVE DEARLY! 🦀 😍)  perhaps i could try for myself!

btw lmk if there is a better forum for this conversation than a closed PR 😂

eddiehazel:
also if there is any mac specific tasks that need doing perhaps i can help, i am not exactly time rich but i am keen to learn the rust from the masters 🙇

PR #7644, opened 2024-01-20
Fix text accessory view overlaps with text when changing IME on macOS

uiryuu:
When changing IME on macOS (or typing in a language that uses a word selection window), the text accessory view will overlap with the text. Add an offset of `cell_height` to prevent this from happening. Found the code to deal with this problem for X11, so I just added another check to offset for macOS.

Before:
image

After:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Text accessory overlaps with text when changing input methods on macOS
 - `alacritty migrate` failing with nonexistent imports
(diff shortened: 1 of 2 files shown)


kchibisov:
if you change the area actually does it change anything? The way it's excluded is that upstream doesn't have support for that on X11, but macOS has it implemented so it could be a bug upstream instead.

In general, I'd direct to fix it there instead of trying to workaround here (note that X11 just can't do that iirc with xim, so that's why it's excluded).

uiryuu:
Thanks for your prompt reply. I'll try to fix this in upstream.

kchibisov:
yeah, at least try opening issue there, so it won't be forgotten.

Issue #6749, opened 2023-03-01
option_as_alt config error in rc 0.12.0-rc1


pitoniak32:
## Issue Description:
I updated from the latest release `0.11.0` to `0.12.0-rc1` because I was looking to get access to the `option_as_alt: Both` config option on my mac for using it as a Meta key with `zellij`.

### Current Behavior:
Config errors, and not able to use option key as alt.

### Expected Behavior:
Ability to use new option without config errors.

### System

OS: macOS Monterey 12.6.1 (Apple M1 Chip)
Version: Version 0.12.0-rc1 (1)

### Logs
<img width="519" alt="Screen Shot 2023-03-01 at 11 42 21 AM" src="https://user-images.githubusercontent.com/84917393/222205643-3d4c97ac-e430-4fbb-a92a-3bd259e7fd40.png">

kchibisov:
I doubt your editor knows about which options alacritty supports.

Not sure how you test the behavior as well.

mkatychev:
@kchibisov @pitoniak32   this is most likely a json schema configuration that's being run through a YAML lsp:
[json-schema spec](https://json-schema.org/)
[json-schema tore](https://www.schemastore.org/json/)
And the schema in question (found through the schema store above):
https://github.com/distinction-dev/alacritty-schema/

Freed-Wu:
Comes from <https://github.com/distinction-dev/alacritty-schema/issues/3>.

How about move [alacritty config json schema](https://github.com/distinction-dev/alacritty-schema) to alacritty repo in order to make this json schema keep latest with every publishing of new version with alacritty?

PR #7646, opened 2024-01-20
Add json schema (#6749)

Freed-Wu:
Screenshot from 2024-01-20 23-24-42

### DIFF
--- a/docs/alacritty-schema.json
+++ b/docs/alacritty-schema.json
@@ -1,0 +1,1061 @@
+{
+  "$id": "alacritty-schema.json",
+  "$schema": "https://json-schema.org/draft-07/schema#",
+  "$comment": "https://alacritty.org/config-alacritty.html",
+  "title": "Alacritty Configuration",
+  "description": "Configuration schema for [Alacritty](https://github.com/alacritty/alacritty), the GPU enhanced terminal emulator",
+  "definitions": {
+    "Alacritty.Cursor.Style": {
+      "type": "object",
+      "description": "Cursor Style",
+      "properties": {
+        "shape": {
+          "type": "string",
+          "default": "Block",
+          "description": "Cursor shape\n\n Values for `shape`:\n   - ▇ Block\n   - _ Underline\n   - | Beam",
+          "enum": ["Block", "Beam", "Underline"]
+        },
+        "blinking": {
+          "description": "# Cursor blinking state\n\n Values for `blinking`:\n   - Never: Prevent the cursor from ever blinking\n   - Off: Disable blinking by default\n   - On: Enable blinking by default\n   - Always: Force the cursor to always blink",
+          "type": "string",
+          "default": "Off",
+          "enum": ["Never", "Off", "On", "Always"]
+        }
+      },
+      "additionalProperties": false
+    },
+    "Alacritty.Command": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "program": {
+          "type": "string"
+        },
+        "args": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        }
+      }
+    },
+    "Alacritty.MouseBinding": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "mouse": {
+          "type": "string",
+          "enum": ["Left", "Middle", "Right"]
+        },
+        "action": {
+          "type": "string"
+        }
+      },
+      "required": ["mouse", "action"]
+    },
+    "Alacritty.KeyBindings.Key": {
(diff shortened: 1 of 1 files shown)


kchibisov:
Thanks, but we're not interested in maintaining it.

it's also not clear what benefit it provides, since it can't be destributed and picked up by each program automatically, unlike the manual page with has a standard procedure of discoverability (`apropos alacritty`) and reading `man 5 alacritty`.

Freed-Wu:
> it's also not clear what benefit it provides

Sorry for my forgetfulness to explain.

Refer <https://github.com/cmhughes/latexindent.pl/issues/206>, [json schema](https://json-schema.org/) allow any editors supporting LSP to complete the configurations. In this screenshot, we complete `window.opacity` of `alacritty.toml`.

Before this PR, there is another [repo](https://github.com/distinction-dev/alacritty-schema) to maintain it. Due to <https://github.com/distinction-dev/alacritty-schema/issues/3>, I think move it to upstream will be better to get updating.

If you still reject it, please reply me. I'll push it to [schemastore](https://www.schemastore.org/). However, I think the problem like #6749 will occur again when it cannot keep latest with official alacritty repo.

Freed-Wu:
> it can't be destributed and picked up by each program automatically

It shouldn't be distributed. Refer <https://github.com/cmhughes/latexindent.pl/issues/206>.

kchibisov:
As I've said, we're not interested in that. You already have documentation distributed and as I said this file can't be used by us, since it requires actual `.toml` config, but we don't have it and don't want to have it, and it can't be provided for each distro by the package, you point to some distribution of the schemes, but I'd rather not provide any of that at all, since such scheme can represent only the latest version, but not the version you're using, which is also solved by `man` since it's distributed along side your alacritty installation if you're using package manager (brew, any linux, doesn't matter).

Freed-Wu:
> You already have documentation distributed

Json schema is not documentation. :cry: It just be used for completion.

@KillDozerX2, @tplk maintain a json schema for old version alacritty. I'll take a look about if they can accept it.

kchibisov:
but it doesn't provide any value and can't guarantee the correct completions that's the thing. The point in the linked issue is the docs mismatch between the version user had and the one in schema.

And the popup completion you have confuses you, since it tells you things for the version schema was generated for, not the actual version you have, so yes, it's also documentation. Anything that tells you the options you have and what they mean also a form of documentation, otherwise why you have text embeded in the schema telling what the options are?

So to not confuse users, like it was on the issue, I'd rather not provide any schema at all, so they can use the standard flow to discover what options they have and what they mean for their version.

Freed-Wu:
> since it tells you things for the version schema was generated for, not the actual version you have

The json schema of this repo should server latest version's config. IMO, if this program keep compatibility, the new version's json schema should validate old version's config successfully. Even if you add some breaking changes, schemastore allow you use two schemas like:

```json
    {
      "name": "alacritty",
      "description": "alacritty config",
      "fileMatch": [
        "alacritty.toml"
      ],
      "url": "/the/url/of/new/version's/config",
      "versions": {
        "0.1.0": "/the/url/of/an/old/version's/config"
        // ...
      }
    },
```

But maintain many different version's json schemas are too troublesome. :cry:

Freed-Wu:
In this problem, I guess the maintainer of schemastore know more? @hyperupcall @madskristensen

hyperupcall:
From my experience, most people use JSON Schema for completion and the convenience of having the documentation right within their editor. Some people use it for validation, but that is less common. Similar to projects such as [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) nothing is guaranteed in terms of absolute correctness, but I think it's still incredibly useful despite that.

It is difficult to maintain schemas for multiple versions if there is no way to autogenerate them. But, having schemas for multiple versions is usually not necessary if the shape of the config file is roughly the same, and keys are mostly added after ever release). In that case, the schema may "incorrectly" include config keys that have been removed in later versions (and visa versa) but the tradeoff is that the schema still provides hints and or does not error during validation for the people who have not yet updated. For example, that is the case with [clangd](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/clangd.json). There are plenty of tricks you can do to fix this from a different angle, especially with more recent versions of JSON Schema, but I haven't seen them much - probably because the time invested isn't worth the 100% correctness.

If the schema cannot be added upstream, then we can always accept them at SchemaStore. :)
About errors like in #6749, you can get around them by using `additionalProperties: true`. The tradeoff with that, is that validation will no longer error when it sees an unrecognized key, of course. But, you will still get documentation hints. Usually we discourage that, but if you really think it is important, then that is an option. Conversely, doing so will mean config keys added in newer versions of Alacritty are more likely not to be in the schema - no errors will be generated for people to notice (and then fix the schema).

PR #7651, opened 2024-01-21
Update IME position more frequently

uiryuu:
Currently the IME position is only updated when the IME is enabled. So the initial position of the IME will be wrong if the IME is not enabled. Changed to update position even if the IME is off.

Steps to reproduce the bug:
- Change typing language to English (turn off IME)
- Open Alacritty
- Start typing using IME, the initial position will be wrong

Or:
- Open Alacritty
- Type in English for a few characters
- Switch to another language which uses IME, the position will be wrong

The bug might only reproducible for some languages' IME depending on whether the IME immediately pup up the candidate window or not.

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -900,8 +900,8 @@ impl Display {
         };
 
         // Handle IME.
-        if self.ime.is_enabled() {
-            if let Some(point) = ime_position {
+        if let Some(point) = ime_position {
+            if self.ime.is_enabled() {
                 let (fg, bg) = if search_state.regex().is_some() {
                     (config.colors.footer_bar_foreground(), config.colors.footer_bar_background())
                 } else {
@@ -909,6 +909,8 @@ impl Display {
                 };
 
                 self.draw_ime_preview(point, fg, bg, &mut rects, config);
+            } else {
+                self.window.update_ime_position(point, &self.size_info);
             }
         }
 


alacritty/src/display/mod.rs
```diff
@@ -900,15 +900,17 @@ impl Display {
         };
 
         // Handle IME.
-        if self.ime.is_enabled() {
-            if let Some(point) = ime_position {
+        if let Some(point) = ime_position {
```
kchibisov, line 903: Winit should indicate that the IME got enabled and you should schedule an update from it, always updating doesn't make any sense and the input into terminal should correctly reposition, since you'll likely have preedit going.

uiryuu: I tried a little bit, the `Ime::Enabled` is only generated when the user starts typing, so we have no idea when the user switched the IME. On macOS, when the user is changing the IME (or just toggle caps lock, see the picture below), the text accessory view will appear on the last position you set.

image
Screenshot 2024-01-22 at 17 26 26

So with the current implementation, the user will never get a correct text accessory view position when they are enabling caps lock on macOS.

kchibisov: What happens when you press `CapsLock`, do we have any keyboard events in general for it?

nixpulvis:
I assume this was closed because the author wasn't interested in working on it anymore, but I can confirm that it does help with things on macOS at least. I don't have anything that uses IME on my linux setup yet so I can't really comment there.

I'll just add a small nitpick that the IME position should be offset down by about half the line height so it doesn't overlap with the cursor itself (see image):

Screenshot 2024-05-29 at 11 29 45 AM

Having the IME positioned correctly really helps me when I accidentally enable caps lock or switch input languages and don't notice because my eyes aren't focused on the top left corner of my terminal.

alacritty/src/display/mod.rs
kchibisov: I mean, it's maybe not delivered from the winit itself. Generally I'm not that much against updating the IME more frequently just on macOS, but I'd like to know a better entry point for such stuff in general.

PR #7658, opened 2024-01-24
Don't report associated text only for C0/C1

### LINKED ISSUE #7657: kitty keyboard encoding reports associated text with some control characters
When using kitty keyboard encoding, keys which map to C0 control codes emit associated text:

Affected keys: tab, enter, backspace, escape

No other terminal implementing kitty keyboard encoding follows this behavior (kitty, foot, wezterm, ghostty).

### System

OS: Linux
Version: alacritty 0.13.1 (fe2a3c56)
Linux/BSD: wayland / sway

### Logs

Keyboard and bindings: [Alacritty-459396.log](https://github.com/alacritty/alacritty/files/14038242/Alacritty-459396.log)

kchibisov:
This has a side effect that we'll have text reported for Alt+Shift+T and similar, but only C0/C1 should be excluded and Alt+Shift+T is emitting neither, thus regular `T` will be reported.

Fixes #7657.

--

@rockorager

### DIFF
--- a/alacritty/src/input/keyboard.rs
+++ b/alacritty/src/input/keyboard.rs
@@ -275,11 +275,18 @@ fn build_sequence(key: KeyEvent, mods: ModifiersState, mode: TermMode) -> Vec<u8
     let context =
         SequenceBuilder { mode, modifiers, kitty_seq, kitty_encode_all, kitty_event_type };
 
+    let associated_text = key.text_with_all_modifiers().filter(|text| {
+        mode.contains(TermMode::REPORT_ASSOCIATED_TEXT)
+            && key.state != ElementState::Released
+            && !text.is_empty()
+            && !is_control_character(text)
+    });
+
     let sequence_base = context
         .try_build_numpad(&key)
         .or_else(|| context.try_build_named(&key))
         .or_else(|| context.try_build_control_char_or_mod(&key, &mut modifiers))
-        .or_else(|| context.try_build_textual(&key));
+        .or_else(|| context.try_build_textual(&key, associated_text));
 
     let (payload, terminator) = match sequence_base {
         Some(SequenceBase { payload, terminator }) => (payload, terminator),
@@ -289,10 +296,7 @@ fn build_sequence(key: KeyEvent, mods: ModifiersState, mode: TermMode) -> Vec<u8
     let mut payload = format!("\x1b[{}", payload);
 
     // Add modifiers information.
-    if kitty_event_type
-        || !modifiers.is_empty()
-        || (mode.contains(TermMode::REPORT_ASSOCIATED_TEXT) && key.text.is_some())
-    {
+    if kitty_event_type || !modifiers.is_empty() || associated_text.is_some() {
         payload.push_str(&format!(";{}", modifiers.encode_esc_sequence()));
     }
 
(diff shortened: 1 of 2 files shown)


alacritty/src/input/keyboard.rs
```diff
@@ -644,3 +642,10 @@ impl From<ModifiersState> for SequenceModifiers {
         modifiers
     }
 }
+
+/// Check that the text represents either C0 or C1 codepoint.
```
chrisduerr, line 646:
```suggestion
/// Check whether the text represents a C0 or C1 codepoint.
```

alacritty/src/input/keyboard.rs
```diff
@@ -644,3 +642,10 @@ impl From<ModifiersState> for SequenceModifiers {
...
+fn is_c0c1_codepoint(text: &str) -> bool {
+    let codepoint = text.chars().next().map(u32::from).unwrap_or(0);
```
chrisduerr, line 648: This would mean that an empty string is a c0 or c1 codepoint. Which I'd say is inaccurate.

I also think that looking at chars/u32s is unnecessary here. Just getting `bytes()` should be fine since no valid unicode character (and `&str` is always valid unicode) can start with a C0/C1 character?

alacritty/src/input/keyboard.rs
```diff
@@ -644,3 +642,10 @@ impl From<ModifiersState> for SequenceModifiers {
...
+    // Include 0x7f as control character.
```
chrisduerr, line 649: This comment strikes me as odd because it doesn't document what the code does nor does it explain why `0x7f` is special and included.

Suggestion:

```
// 0x7f (DEL) is included here since it has a dedicated control code (`^?`) which generally does not match the reported text (`^H`), despite not technically being part of C0 or C1.
```

Not sure if this explanation matches your reasoning, but I'd assume it's something like that.

chrisduerr (changes requested):
What motivated you to change this considering the following statement:

> There's no wording saying that I can't do so and I don't see why I shouldn't. libxkbcommon explicitly tells me that I have text, thus, I send text.

alacritty/src/input/keyboard.rs
kchibisov: Yeah, maybe checking for empty should be better on callee side.

kchibisov:
>What motivated you to change this considering the following statement:

https://github.com/kovidgoyal/kitty/commit/90e1ba77812dde95920cd595e56c51f420b0f13e

alacritty/src/input/keyboard.rs
kchibisov: I still can't remove the len check due to compose input, since it could be an actual string in the end of the day.

alacritty/src/input/keyboard.rs
```diff
@@ -644,3 +643,11 @@ impl From<ModifiersState> for SequenceModifiers {
         modifiers
     }
 }
+
+/// Check whether the text represents a C0 or C1 codepoint.
+fn is_c0c1_codepoint(text: &str) -> bool {
+    let codepoint = text.bytes().next().unwrap();
+    // 0x7f (DEL) is included here since it has a dedicated control code (`^?`) which generally
+    // does not match the reported text (`^H`), despite not technically being part of C0 or C1.
+    text.bytes().len() == 1 && (codepoint < 0x20 || (0x7f..=0x9f).contains(&codepoint))
```
chrisduerr, line 652: Is there any difference between `text.bytes().len()` and `text.len()`? Isn't `text.len()` the byte length anyway?

And even if it was the character length instead, it would still work since the only characters that start with the control codes are the C0 and C1s.

alacritty/src/input/keyboard.rs
```diff
@@ -644,3 +643,11 @@ impl From<ModifiersState> for SequenceModifiers {
         modifiers
     }
 }
+
+/// Check whether the text represents a C0 or C1 codepoint.
+fn is_c0c1_codepoint(text: &str) -> bool {
```
chrisduerr, line 648: Might want to change the name because of `0x7f`? Maybe just `is_control_character` with a comment which specific control characters this function detects?

chrisduerr (changes requested):

alacritty/src/input/keyboard.rs
kchibisov: No, there's no.

The issue is that C0/C1 could be a part of the large compose insert, no? But I guess we should throw it away anyway.

chrisduerr: I have no problem with checking for length here. I think it makes sense and I don't think we should report this as a control character when the length is bigger than 1.

All I was saying is that you can change `text.bytes().len()` to `text.len()`.

chrisduerr (approved):

PR #7661, opened 2024-01-25
Clamp end point before getting selection range

MoSal:
An out of range end column, for example, can cause out of range access later.

### DIFF
--- a/alacritty_terminal/src/selection.rs
+++ b/alacritty_terminal/src/selection.rs
@@ -285,6 +285,7 @@ impl Selection {
             return None;
         }
         start.point = start.point.grid_clamp(term, Boundary::Grid);
+        end.point = end.point.grid_clamp(term, Boundary::Grid);
 
         match self.ty {
             SelectionType::Simple => self.range_simple(start, end, columns),


alacritty_terminal/src/selection.rs
```diff
@@ -285,6 +285,7 @@ impl Selection {
             return None;
         }
         start.point = start.point.grid_clamp(term, Boundary::Grid);
+        end.point = end.point.grid_clamp(term, Boundary::Grid);
```
kchibisov, line 288: Is the column out of range? Because we already check the line on the line above?

MoSal: > Is the column out of range?

Yes. And the selection was semantic. The code run ended with an out-of-range index here:

https://github.com/alacritty/alacritty/blob/79b686df419d90e5557d1b7000f51f012986c141/alacritty_terminal/src/grid/mod.rs#L601

This is with a terminal that depends on `alacritty_terminal`. I don't know if this can be triggered with `alacritty` proper or not.

Also, the check above obviously doesn't check `> bottommost_line`.

kchibisov: >This is with a terminal that depends on alacritty_terminal. I don't know if this can be triggered with alacritty proper or not.

I got that, since it can't be triggered with alacritty itself, since we probably clamp on the callee side.

In general, I'm fine with such clamping, since it's consistent.

chrisduerr (approved):
Tested it and couldn't find any regressions, so I don't see a reason not to merge this.

PR #7681, opened 2024-02-01
Avoid panic due to unwrap

### LINKED ISSUE #7680: EventLoop::new() can panic with "Bad File Descriptor"
We are using Alacritty at Zed to provide the embedded terminal, and across the whole userbase see this panic a few times a day:

```
thread 'main' panicked at 'create mio Poll: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }'
/Users/administrator/.cargo/git/checkouts/alacritty-afea874b09a502a5/3330614/alacritty_terminal/src/event_loop.rs:163
<terminal::TerminalBuilder>::new
<project::Project>::create_terminal
// more zed stuff....
```

Although we have struggled to reproduce this problem in situ, we suspect this happens while the filesystem is quite heavily loaded (we're restoring a workspace for example https://github.com/zed-industries/zed/issues/6835). That said, this is not reliably reproducible for those users.

I think the best next step is relatively straightforward, which is to pass the result out of `EventLoop::new()` instead of unwrapping, so that the caller can handle it.

### System

OS: macOS
Version: [Zed](https://zed.dev) using alacritty_terminal v0.21

### Logs

```
thread 'main' panicked at 'create mio Poll: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }'
/Users/administrator/.cargo/git/checkouts/alacritty-afea874b09a502a5/3330614/alacritty_terminal/src/event_loop.rs:163
<terminal::TerminalBuilder>::new
<project::Project>::create_terminal
// more zed stuff....
```

ConradIrwin:
Fixes #7680

Although this doesn't address the root cause, it will allow us to stop crashing the entire process when it does.

### DIFF
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -214,8 +214,9 @@ where
             let mut interest = PollingEvent::readable(0);
 
             // Register TTY through EventedRW interface.
-            unsafe {
-                self.pty.register(&self.poll, interest, poll_opts).unwrap();
+            if let Err(err) = unsafe { self.pty.register(&self.poll, interest, poll_opts) } {
+                error!("EventLoop register error: {err:?}");
+                return (self, state);
             }
 
             let mut events = Events::with_capacity(NonZeroUsize::new(1024).unwrap());
@@ -236,7 +237,10 @@ where
                 if let Err(err) = self.poll.wait(&mut events, timeout) {
                     match err.kind() {
                         ErrorKind::Interrupted => continue,
-                        _ => panic!("EventLoop polling error: {err:?}"),
+                        _ => {
+                            error!("EventLoop polling error: {err:?}");
+                            break 'event_loop;
+                        },
                     }
                 }
 
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -215,7 +215,7 @@ where
 
             // Register TTY through EventedRW interface.
             if let Err(err) = unsafe { self.pty.register(&self.poll, interest, poll_opts) } {
-                error!("EventLoop register error: {err:?}");
+                error!("Event loop registration error: {}", err);
                 return (self, state);
             }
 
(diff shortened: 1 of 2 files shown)


alacritty_terminal/src/event_loop.rs
```diff
@@ -213,8 +214,9 @@ where
             let mut interest = PollingEvent::readable(0);
 
             // Register TTY through EventedRW interface.
-            unsafe {
-                self.pty.register(&self.poll, interest, poll_opts).unwrap();
+            if let Err(err) = unsafe { self.pty.register(&self.poll, interest, poll_opts) } {
+                error!("EventLoop register error: {err:?}");
```
chrisduerr, line 218: Errors are generally better logged with their `Display` impl. Also think this reads a little nicer as "registration error" rather than "register error", since it's not an error with a register (like a CPU register).

Should also be `Event loop`, rather than `EventLoop`.

alacritty_terminal/src/event_loop.rs
```diff
@@ -235,7 +237,10 @@ where
                 if let Err(err) = self.poll.wait(&mut events, timeout) {
                     match err.kind() {
                         ErrorKind::Interrupted => continue,
-                        _ => panic!("EventLoop polling error: {err:?}"),
+                        _ => {
+                            error!("EventLoop polling error: {err:?}");
```
chrisduerr, line 241: Same comment about event loop/display.

chrisduerr (changes requested):
Seems sensible enough. Just minor style changes.

ConradIrwin:
Thanks for the fast review, updated!

chrisduerr (approved):
Do you need this released?

ConradIrwin:
Yes please! (Though if not we can depend on the GitHub version for a bit)

kchibisov:
I'd just schedule it to be part of alacritty 0.13.2 which may happen during the next couple of weeks.

ConradIrwin:
Thanks!

PR #7685, opened 2024-02-02
Alacritty Terminal version 0.21.0

chrisduerr:


kchibisov:
Do we want to remove the `-dev` stuff from the `alacritty_terminal`?

chrisduerr:
> Do we want to remove the -dev stuff from the alacritty_terminal?

Yes, since this is not a development version? It's a release.

kchibisov:
@chrisduerr but we usually do that on separate branches and if you want to do something like that on master, you should probably update the `CONTRIBUTING.md` wrt how `alacritty_terminal` is released.

chrisduerr:
> @chrisduerr but we usually do that on separate branches and if you want to do something like that on master, you should probably update the CONTRIBUTING.md wrt how alacritty_terminal is released.

Oh yeah good point, I had no intention of merging this into master. I wanted this to be on the same branch I have on my fork.

kchibisov (approved):

Issue #5789, opened 2022-01-16
I want to set the path relative to the config file in import.


mondy:
When the following configuration files are in the same directory, I want to load `default.yml` in `alacritty.yml`.
- alacritty.yml
- default.yml

alacritty.yml:

```
import:
  - default.yml
```

default.yml:

```
font:
  normal:
    family: fooooo
```

Originally, the relative path of import should be from the configuration file. Currently, it seems to be looking at the path relative to the current directory when the program is executed.
Can we at least introduce a prefix (like a ~/) to point to the directory where the configuration file is located?

### System

Version: alacritty 0.9.0 (fed349a)

mondy:
#4157 found.
It turns out that this is a duplicate issue.
I hope this issue will be fixed if possible.

Issue #4164, opened 2020-08-24
Import config files from home dir or relative


gempir:
I love the addition of importing config files in 3c3e6870dedad56b270f5b65ea57d5a6e46b1de6
Now i can finally split up OS specific configs easily. 

The issue still is I can't find a way to import from the home directory or relative.
I tried using 

- $HOME
- ~/

Is there something I'm missing or is that feature just not supported yet, since this did whole feature just came to master.

### System

OS: Ubuntu 20.04 / macOS 10.15.6
Version: alacritty 0.6.0-dev (6cfcd7c)
Linux/BSD: X11 + i3 on Ubuntu

Edit:

I found out `.config/alacritty/myfile.yml` is working. So without any prefix it always starts in the home directory. Is that always going to stay like that or should it be possible to make it absolute?

enizor:
See #4157. For short: the implementation does not support expansion yet, and it may not be desirable.

chrisduerr:
> I found out .config/alacritty/myfile.yml is working. So without any prefix it always starts in the home directory. Is that always going to stay like that or should it be possible to make it absolute?

This is not true, the path is simply relative to where you started Alacritty. That's not going to change, but it is not reliable either.

PR #7690, opened 2024-02-05
Support relative path imports from config file

caojoshua:
Currently any relative path import will import relative from the alacritty executable. I believe its not useful. This PR makes the imports relative from the config file.

Nested relative imports work. For example, `A` can import `B` relative from `A`'s path, and then `B` can import `C` relative from `C`'s path.

One caveat is that `alacritty migrate` will expand relative paths to absolute paths. I noticed that this is the same behavior in that it expands `~/` (which was added in https://github.com/alacritty/alacritty/pull/4449).


chrisduerr (changes requested):
Using relative imports at all seems like a poor choice. Why is this something we want to support?

kchibisov:
yeah, relative path are a bit confusing, especially when it comes to symbolic links, etc. paths relative to the home dir(which we support) are way more clear imho.

caojoshua:
> Using relative imports at all seems like a poor choice. Why is this something we want to support?

Why do relative imports seem like a poor choice?

My use case is that in my configs, I have a git submodule for [alacritty-theme](https://github.com/alacritty/alacritty-theme). The configs are setup such that the relative paths are always consistent between `alacritty.toml` and the theme I want to import. Absolute paths are not guaranteed to be consistent across systems.

There are at least a handful of users who would like this support https://github.com/alacritty/alacritty/issues/4164 https://github.com/alacritty/alacritty/issues/5789

kchibisov:
it's not predictable with symlinks though, since you have 2 choices basically. And relative from home only breaks when you use windows and unix systems at the same time, I think, but the rest is always the `~/.config/alacritty/` or `~/`. Though you could put your imports into the same place relative to `HOME`, which we already support.

caojoshua:
> it's not predictable with symlinks though, since you have 2 choices basically.

I'm not understanding why this is an issue. The path will be relative from where Alacritty reads the `alacritty.toml`. I don't think there will ever be 2 choices. In general, programming and scripting languages allow for relative paths and I don't think they create extra confusion. 

> And relative from home only breaks when you use windows and unix systems at the same time

This does not apply to me, but this would help people under this use case

> Though you could put your imports into the same place relative to HOME

Yes this is true. There are solutions without relative paths. I am suggesting supporting relative path imports as a quality of life feature.

mxaddict:
I think relative paths are a good addition (Assuming that it's relative to the main config file)

Would allow for configs like so:
```toml
import = [
    "./tokyo-night.toml",
    "./common.toml",
]
```

Compared to what I currently have which is:
```toml
import = [
    "~/.config/alacritty/tokyo-night.toml",
    "~/.config/alacritty/common.toml",
]
```

melMass:
> And relative from home only breaks when you use windows and unix systems at the same time, I think, but the rest is always the `~/.config/alacritty/` 

That's exactly why I need relative paths personally, I share common dotfiles for Windows native, wsl and macos.

My solution is indeed symlinking on all os to `~/alacrity-themes` while storing this folder next to my config

alacritty/src/config/mod.rs
```diff
@@ -313,6 +319,12 @@ pub fn imports(
             path = home_dir.join(stripped);
         }
 
+        if path.is_relative() {
+            if let Some(config_path) = base_path.parent() {
+                path = config_path.join(path)
```
chrisduerr, line 324:
```suggestion
            if let Some(base_path) = base_path.parent() {
                path = base_path.join(path)
```

Better to keep consistent naming to avoid confusion.

chrisduerr (changes requested):
I've contemplated this for a bit and I do think it makes sense to add, though there is some risk for breakage.

This PR uses the unresolved path if the configuration file is a symlink, which I think is the only thing that really makes sense. If people have a symlinked config, they should also symlink the imports.

I've considered if an alternative implementation would be changing Alacritty's working directory, but that would just break things and conflict with the working directory config option. We could set it just for subprocesses spawned by bindings, but I don't think there's any actual reason for why we'd bother.

chrisduerr (approved):

kchibisov (changes requested):
This should be documented in the man pages.

caojoshua:
Added docs. This is the plain text version I can see locally.

```
           All imports must either be absolute paths starting with /, paths relative to the user's home directory starting with ~/, or paths relative from the current config file.

           Example:
               import = [
                    "~/.config/alacritty/base16-dark.toml",
                    "~/.config/alacritty/keybindings.toml",
                    "alacritty-theme/themes/gruvbox_dark.toml",
               ]
```

I also moved this change to `0.14.0-dev` in the changelog

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Notable changes to the `alacritty_terminal` crate are documented in its
 - Pressing `Alt` with unicode input will now add `ESC` like for ASCII input
 - Decorations use opaque style and system window background on macOS
 - No longer source `~/.zshenv` on macOS
+- Support relative path imports from config files
```
kchibisov, line 18: I'm pretty sure it's `Added`, but changed is also fine.

kchibisov (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ Notable changes to the `alacritty_terminal` crate are documented in its
 
 ## 0.14.0-dev
 
+### Added
+- Support relative path imports from config files
```
chrisduerr, line 14:
```suggestion
### Added

- Support relative path imports from config files
```

chrisduerr (changes requested):

chrisduerr (approved):

PR #7703, opened 2024-02-08
Remove extra TIOCSWINSZ ioctl on startup

kchibisov:
The openpty call already performs it, thus no need to call it one more with the exact same size since it confuses some applications.


chrisduerr (approved):
It really shouldn't cause any issues though…

kchibisov:
@chrisduerr yeah...

PR #7709, opened 2024-02-10
Update dependencies

chrisduerr:
At the current point in time Alacritty's master does not build due to one of its dependencies. This seems like a good time to apply non-breaking dependency updates, which also fixes this issue in the process.

### DIFF
--- a/alacritty/src/input/mod.rs
+++ b/alacritty/src/input/mod.rs
@@ -19,6 +19,8 @@ use winit::dpi::PhysicalPosition;
 use winit::event::{
     ElementState, Modifiers, MouseButton, MouseScrollDelta, Touch as TouchEvent, TouchPhase,
 };
+#[cfg(target_os = "macos")]
+use winit::event_loop::EventLoopWindowTarget;
 use winit::keyboard::ModifiersState;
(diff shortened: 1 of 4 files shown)


alacritty/src/input/mod.rs
```diff
@@ -103,7 +102,6 @@ pub trait ActionContext<T: EventListener> {
     fn pop_message(&mut self) {}
     fn message(&self) -> Option<&Message>;
     fn config(&self) -> &UiConfig;
-    fn event_loop(&self) -> &EventLoopWindowTarget<Event>;
```
chrisduerr, line 106: Turns out this was never used anywhere.

kchibisov:
I would clarify that doesn't build with _nightly_.

chrisduerr:
Closing since I don't want to update MSRV to 1.74.0 at this point, but it seems required by clap.

kchibisov:
you can bump just `ahash` if you want the thing to get build on hightly again.

chrisduerr:
> you can bump just ahash if you want the thing to get build on hightly again.

I don't really care that much since it'll probably fix itself anyway sooner or later.

Issue #7712, opened 2024-02-10
[feature request] Would be nice to have an `ALACRITTY_OS` env added


mxaddict:
Would be nice to have an env value set by allacritty that tells the current os.

Something like `ALACRITTY_OS` which would have the current OS.

So when running on macOS `ALACRITTY_OS` would be `darwin`

On linux `ALACRITTY_OS` would be `linux`

On windows `ALACRITTY_OS` would be `winblowz`

Or something like that.

Then possibly allow substitution of `env` values during `import` in configs like:

```toml
import = [
"~/.config/alacritty/${ALACRITTY_OS}.toml",
"~/.config/alacritty/${SOME_OTHER_ENV}.toml",
]
```

mxaddict:
Let me know if a feature like this would be welcome, I can create a PR for this.

chrisduerr:
OS-specific imports are not desired. Imports are already optional, if you don't want something to be loaded just don't put it there.

mxaddict:
This is not a OS specific import feature really, it's more like adding a `OS` to the env values.

Which could be used for other reason (I have some fish scripts that use `KITTY_OS` for example

Also, import with env parsing can be for setting themes via a simple env value.

These are just 2 use cases for this, but I'm sure there can be more.

mxaddict:
As for the OS env value, can be as simple as :

```diff
diff --git a/alacritty_terminal/src/tty/mod.rs b/alacritty_terminal/src/tty/mod.rs
index d1bb023c..e54a1eca 100644
--- a/alacritty_terminal/src/tty/mod.rs
+++ b/alacritty_terminal/src/tty/mod.rs
@@ -91,6 +91,8 @@ pub fn setup_env() {
     let terminfo = if terminfo_exists("alacritty") { "alacritty" } else { "xterm-256color" };
     env::set_var("TERM", terminfo);

+    env::set_var("OS", std::env::consts::OS);
+
     // Advertise 24-bit color support.
     env::set_var("COLORTERM", "truecolor");
```

mxaddict:
Tested on a local build of alacritty:
<img width="436" alt="Screenshot 2024-02-10 at 10 31 28 PM" src="https://github.com/alacritty/alacritty/assets/1060905/47654326-49fa-4e91-b3ff-014e57b94818">

kchibisov:
You can detect OS yourself in shell scripts. e.g. with `uname` (if it doesn't exist, it's likely windows).

[...]

PR #7713, opened 2024-02-10
Added new env value for `OS`

mxaddict:
Screenshot 2024-02-10 at 10 31 28 PM


mxaddict:
This is in relation to #7712

mxaddict:
Is it really that bad to have an env with the current OS value?

Issue #6872, opened 2023-04-19
In system monitoring tools like bottom,gotop the graph part is not displayed properly


yogeshdcool:
### System

OS: Linux
Version: alacritty 0.12.0 (5a728195)
Linux/BSD: X11,dwm,picom

### Problem
__When using system monitoring (bottom), the graph part is not displayed properly  
I tried changing the font but nothing changed__

### This is how it looks in alacritty
![alacritty](https://user-images.githubusercontent.com/84368844/232988884-f013e47e-77c9-4bb1-bb48-c185d3a28ab5.png)

### This is how it looks in kitty
![kitty](https://user-images.githubusercontent.com/84368844/232988965-dee02054-584e-4ddd-a486-4ffd3d0fc81f.png)

### Alacritty and Kitty
![diff](https://user-images.githubusercontent.com/84368844/232989449-43ef48f9-5fd2-4769-866c-0e2ece49af67.png)

chrisduerr:
Working as intended.

yogeshdcool:
> Working as intended.

Can you please say what do you mean?

lbfalvy:
@kchibisov Would a PR for this likely be accepted?

kchibisov:
Well you can try and see how it looks.

chrisduerr:
For most of our internal glyphs the motivation behind drawing them internally is that they're purposefully included into Unicode to enable text-based UIs, maybe with the small exception of powerline which still doesn't really serve any other purpose.

I find it hard to justify custom-drawing braille symbols, considering they have a very specific purpose which is **not** for rendering graphs in the terminal (which is what they're commonly abused for).

kchibisov:
tbf, braille has a purpose in real world, but on the screen where you can't fill them with your fingers they are quite useless.

j4james:
Braille on a computer screen can still be used by sighted or visually paired people to learn the system, or even just to get a general idea of what it looks like when reading an article on the subject. The whole point of the font is to simulate the look of real-world Braille.

You may still decide that pseudo graphs are of more value to your users than legitimate Braille usage, but I think it's unfair to suggest that Braille on a computer screen is useless.

lbfalvy:
It's not that braille on a computer screen is useless, but specifically in a terminal emulator I think it's more likely to appear either as graphics or an unused side effect of a workflow that involves both tts and a braille display, than as learning material.

[...]

Issue #5547, opened 2021-10-16
Help: Blocky Braille Characters


SystematicError:
Hey! I recently saw [this](https://st.suckless.org/patches/boxdraw/) called boxdraw, which makes the terminal font more pleasing for TUI's. I want something like the pixelated braille characters shown in the patch, however I'm not sure how to do this in Alacritty.

chrisduerr:
You write your own patch.

RainbowShatter:
is there a tutorial for that

chrisduerr:
If you look at the closed PR associated with this issue, you should be able to figure out how to get it working pretty easily. Though my recommendation would just be installing a font that does what you want.

RainbowShatter:
> If you look at the closed PR associated with this issue, you should be able to figure out how to get it working pretty easily. Though my recommendation would just be installing a font that does what you want.

it's because I want to know if I can combine ubraille font and meslo nerd fonts together

RainbowShatter:
> If you look at the closed PR associated with this issue, you should be able to figure out how to get it working pretty easily. Though my recommendation would just be installing a font that does what you want.

what closed PR? i dont know what that is

Issue #4192, opened 2020-08-31
braille not rendering properly 


Titaniumtown:
## System
OS: Linux
Version: alacritty 0.5.0
Linux/BSD: X11, Gnome

Braille doesn't display properly in tools like ytop and bottom. 
![Screenshot from 2020-08-31 14-38-58](https://user-images.githubusercontent.com/11786225/91754646-c8392a00-eb97-11ea-9e11-be467dbb0418.png)

kchibisov:
Looks like a braille from your font, I don't see anything wrong with it?

Titaniumtown:
It should be a singular dot. On gnome-terminal the issue isn't there. 
![image](https://user-images.githubusercontent.com/11786225/91757206-f28ce680-eb9b-11ea-9239-4be588afa166.png)

Titaniumtown:
Ok, so I followed https://www.reddit.com/r/archlinux/comments/gf2vgb/the_braille_fonts_dont_show_properly_anywhere/ 

> Install a font like ttf-symbola and if you have gnu-free-fonts installed, remove it

and it works. Strange the issue is only on alacritty.

kchibisov:
Alacritty is using **your** font, gnome-terminal as well as some other terminals are **not** doing so, instead they do render their builtin artificial font when facing those glyphs, line drawing glyphs, etc, meaning that there's no way you can ever see those glyphs from any font you're using.

PR #7714, opened 2024-02-10
Custom drawing logic for Braille symbols

lbfalvy:
Support for TUI programs that use braille characters to draw graphs, as requested in #6872 , in #5547 and in #4192 .

Notes:
- In 6872 I suggested that this could be separately configured in case users want to read actual Braille text in the terminal, I haven't implemented that yet, not sure if there's real demand and whether we're saving on config options.
- Antialiasing isn't done with an actual circle-square overlap algo, it's good enough by my standards, maybe the esteemed maintainers disagree.
- The size of the dots is totally arbitrary.

Example on my fairly low resolution display:
Screenshot_20240210_140951


chrisduerr (changes requested):
Not interested in this feature, especially because it can be **easily** provided by fonts themselves.

The leading motivator behind Alacritty's custom font is that they cannot reliably be implemented in the fonts themselves. This is not true for braille.

PR #7721, opened 2024-02-11
Occupied entries are correctly updated when using IndexMut<RangeToInc…

placintaalexandru:
[Here](https://github.com/alacritty/alacritty/blob/master/alacritty_terminal/src/grid/row.rs#L209), the number of occupied columns is updated correctly, but, on `IndexMut<RangeToInclusive<Column>>` trait impl, the occupied uses `*index.end` instead of `*index.end + 1`

### DIFF
--- a/alacritty_terminal/src/grid/row.rs
+++ b/alacritty_terminal/src/grid/row.rs
@@ -291,7 +291,7 @@ impl<T> Index<RangeToInclusive<Column>> for Row<T> {
 impl<T> IndexMut<RangeToInclusive<Column>> for Row<T> {
     #[inline]
     fn index_mut(&mut self, index: RangeToInclusive<Column>) -> &mut [T] {
-        self.occ = max(self.occ, *index.end);
+        self.occ = max(self.occ, *index.end + 1);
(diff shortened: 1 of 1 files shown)


chrisduerr (approved):
Was this the source of a bug or how did you spot this issue?

It definitely looks correct to me, though the code in question is pretty fragile so I'm surprised it was never encountered.

placintaalexandru:
I want to start contributing to alacritty so I started looking through the code.

chrisduerr:
Turns out nobody found this bug because this code isn't used. At least not by Alacritty, it still makes sense to patch this for other `alacritty_terminal` consumers.

placintaalexandru:
@chrisduerr, where should one start to look to become a contributor?

chrisduerr:
> where should one start to look to become a contributor

You've just contributed to Alacritty, making you a contributor. That's all it takes.

If you're looking for issues to work on, the milestones are usually what is most pressing. If those are too complicated, the backlog has tags for easy issues. If you have any questions feel free to reach out on IRC.

PR #7725, opened 2024-02-12
Fix regional scrolling leaking into history

chrisduerr:
This fixes an issue where a scrolling region that does not start at the top of the screen would still rotate lines into history when scrolling the content "upwards".


kchibisov:
@perfbot

kchibisov (changes requested):
Could you send your macOS changes in msrv PR?

kchibisov (approved):

PR #7726, opened 2024-02-12
Use dynamic MSRV for oldstable CI

chrisduerr:
Instead of manually specifying the oldstable version in all our CI
scripts, it is now pulled from the `Cargo.toml` which simplifies the
update process.

The contributing guide has also been updated to not include the explicit
version and its wording has been loosened a bit to correctly represent
current maintenance practices.

### DIFF
--- a/alacritty/Cargo.toml
+++ b/alacritty/Cargo.toml
@@ -7,7 +7,7 @@ description = "A fast, cross-platform, OpenGL terminal emulator"
 readme = "README.md"
 homepage = "https://github.com/alacritty/alacritty"
 edition = "2021"
-rust-version = "1.70.0"
+rust-version = "1.72.0"
 
 [dependencies.alacritty_terminal]
 path = "../alacritty_terminal"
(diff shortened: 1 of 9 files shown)


alacritty/Cargo.toml
```diff
@@ -7,7 +7,7 @@ description = "A fast, cross-platform, OpenGL terminal emulator"
 readme = "README.md"
 homepage = "https://github.com/alacritty/alacritty"
 edition = "2021"
-rust-version = "1.70.0"
+rust-version = "1.72.0"
```
kchibisov, line 10: A bit ahead of debian, but we likely don't care? 0.8.7 of ahash will work just fine.

chrisduerr: We also have some time until making another release. I was hoping that until then 1.72.0 is supported.

kchibisov: I guess, just add your macOS changes you did in other PR and it should be good.

kchibisov (approved):

PR #7729, opened 2024-02-13
Replace takes into account alias attribute

### LINKED ISSUE #7720: CLI -o config overrides should accept aliases as documented
### Expected behavior
As indicated by `man alacritty`, the `-o` option can be used to override configuration from the `alacritty.toml` config file with the properties specified by `man 5 alacritty`. One would therefore expect that
```alacritty -o 'colors.cursor.cursor="CellBackground"'```
leads to an invisible cursor.

### Actual behavior
Instead, it gives the error
```Unable to override option 'colors.cursor.cursor="CellBackground"': Field 'cursor' does not exist```
and no change to the background color is effected.

### Workaround
Inspecting the source code, this appears to stem from the fact that the `colors.cursor` configuration option is an `InvertedCellColors` struct, for which `cursor` is an alias for a property named `background`. Indeed,
```alacritty -o 'colors.cursor.background="CellBackground"'```
yields the desired result. The `-o` overrides should resolve those aliases since only the aliases are used in the documentation.

### Possible cause
Digging a bit deeper, I think the relevant part of the code is the function `derive_deserialize` in `alacritty/alacritty_config_derive/src/config_deserialize/de_struct.rs` which implements `SerdeReplace` without using the result of `fields_deserializer` which includes the substitution of aliases.

### System

OS: Linux 6.7.4-arch1-1
Version: alacritty 0.13.1 (fe2a3c56)
Compositor: sway

### PR DESCRIPTION
**Pull request by placintaalexandru**
Fixes #7720

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Autokey no longer working with alacritty on X11
 - Freeze when moving window between monitors on Xfwm
 - Mouse cursor not changing on Wayland when cursor theme uses legacy cursor icon names
+- Config keys are available under proper names
 
 ### Changed
 
--- a/alacritty_config_derive/src/serde_replace.rs
+++ b/alacritty_config_derive/src/serde_replace.rs
@@ -44,7 +44,10 @@ pub fn derive_recursive<T>(
 ) -> TokenStream2 {
     let GenericsStreams { unconstrained, constrained, .. } =
         crate::generics_streams(&generics.params);
-    let replace_arms = match_arms(&fields);
+    let replace_arms = match match_arms(&fields) {
+        Err(e) => return e.to_compile_error(),
+        Ok(replace_arms) => replace_arms,
+    };
 
     quote! {
         #[allow(clippy::extra_unused_lifetimes)]
@@ -75,7 +78,7 @@ pub fn derive_recursive<T>(
 }
 
 /// Create SerdeReplace recursive match arms.
-fn match_arms<T>(fields: &Punctuated<Field, T>) -> TokenStream2 {
+fn match_arms<T>(fields: &Punctuated<Field, T>) -> Result<TokenStream2, syn::Error> {
     let mut stream = TokenStream2::default();
     let mut flattened_arm = None;
 
@@ -88,19 +91,42 @@ fn match_arms<T>(fields: &Punctuated<Field, T>) -> TokenStream2 {
         let flatten = field
             .attrs
             .iter()
+            .filter(|attr| (*attr).path().is_ident("config"))
             .filter_map(|attr| attr.parse_args::<Attr>().ok())
             .any(|parsed| parsed.ident.as_str() == "flatten");
 
         if flatten && flattened_arm.is_some() {
-            return Error::new(ident.span(), MULTIPLE_FLATTEN_ERROR).to_compile_error();
+            return Err(Error::new(ident.span(), MULTIPLE_FLATTEN_ERROR));
         } else if flatten {
             flattened_arm = Some(quote! {
                 _ => alacritty_config::SerdeReplace::replace(&mut self.#ident, value)?,
             });
         } else {
+            // Extract all `#[config(alias = "...")]` attribute values.
+            let aliases = field
+                .attrs
+                .iter()
+                .filter(|attr| (*attr).path().is_ident("config"))
+                .filter_map(|attr| attr.parse_args::<Attr>().ok())
+                .filter(|parsed| parsed.ident.as_str() == "alias")
+                .map(|parsed| {
+                    let value = parsed
+                        .param
+                        .ok_or_else(|| format!("Field \"{}\" has no alias value", ident))?
+                        .value();
+
+                    if value.trim().is_empty() {
+                        return Err(format!("Field \"{}\" has an empty alias value", ident));
+                    }
+
+                    Ok(value)
+                })
+                .collect::<Result<Vec<String>, String>>()
+                .map_err(|msg| Error::new(ident.span(), msg))?;
+
             stream.extend(quote! {
-                #literal => alacritty_config::SerdeReplace::replace(&mut self.#ident, next_value)?,
-            });
+                    #(#aliases)|* | #literal => alacritty_config::SerdeReplace::replace(&mut
+            self.#ident, next_value)?,         });
         }
     }
 
@@ -109,5 +135,5 @@ fn match_arms<T>(fields: &Punctuated<Field, T>) -> TokenStream2 {
         stream.extend(flattened_arm);
     }
 
-    stream
+    Ok(stream)
 }
--- a/alacritty_config_derive/tests/config.rs
+++ b/alacritty_config_derive/tests/config.rs
@@ -23,7 +23,7 @@ impl Default for TestEnum {
 
 #[derive(ConfigDeserialize)]
 struct Test {
-    #[config(alias = "noalias")]
+    #[config(alias = "field1_alias")]
     #[config(deprecated = "use field2 instead")]
     field1: usize,
     #[config(deprecated = "shouldn't be hit")]
@@ -39,6 +39,9 @@ struct Test {
     enom_error: TestEnum,
     #[config(removed = "it's gone")]
     gone: bool,
+    #[config(alias = "multiple_alias1")]
+    #[config(alias = "multiple_alias2")]
+    multiple_alias_field: usize,
 }
 
 impl Default for Test {
@@ -53,6 +56,7 @@ impl Default for Test {
             enom_big: TestEnum::default(),
             enom_error: TestEnum::default(),
             gone: false,
+            multiple_alias_field: 0,
         }
     }
 }
@@ -70,6 +74,7 @@ struct Test2<T: Default> {
 
 #[derive(ConfigDeserialize, Default)]
 struct Test3 {
+    #[config(alias = "flatty_alias")]
     flatty: usize,
 }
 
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by placintaalexandru** [comment]
Running `cargo run -- -o 'colors.cursor.cursor="#ff0000"'` works and makes the cursor red:
image

File: alacritty_config_derive/src/serde_replace.rs [hunk 1]
```diff
@@ -98,9 +98,26 @@ fn match_arms<T>(fields: &Punctuated<Field, T>) -> TokenStream2 {
                 _ => alacritty_config::SerdeReplace::replace(&mut self.#ident, value)?,
             });
         } else {
-            stream.extend(quote! {
-                #literal => alacritty_config::SerdeReplace::replace(&mut self.#ident, next_value)?,
-            });
+            // Check if #[config(alias = "some alias")], and, if present, extract its value
```
chrisduerr, line 101:
```suggestion
            // Extract `#[config(alias = "...")]` attribute values.
```

alacritty_config_derive/src/serde_replace.rs
```diff
@@ -98,9 +98,26 @@ fn match_arms<T>(fields: &Punctuated<Field, T>) -> TokenStream2 {
...
+            let alias_value = field
+                .attrs
+                .iter()
+                .filter_map(|attr| attr.parse_args::<Attr>().ok())
+                .find(|parsed| parsed.ident.as_str() == "alias")
```
chrisduerr, line 106: Find here only allows for a single alias. I think having multiple `#[config(alias = "xxx")]` annotations is technically supported even if it might not be used right now. This should also be taken into account here.

chrisduerr (changes requested):

alacritty_config_derive/src/serde_replace.rs
placintaalexandru: Oh, sorry. I was not aware of this. I'll update

placintaalexandru: hello @chrisduerr,
I've just updated according to your request + test to check the multiple `#[config(alias = "some_alias")]` part

alacritty_config_derive/src/serde_replace.rs
```diff
@@ -98,9 +99,28 @@ fn match_arms<T>(fields: &Punctuated<Field, T>) -> TokenStream2 {
...
-            });
+            // Extract all `#[config(alias = "...")]` attribute values.
+            let aliases = field
+                .attrs
+                .iter()
+                .filter(|attr| (*attr).path().is_ident("config"))
+                .filter_map(|attr| attr.parse_args::<Attr>().ok())
+                .filter(|parsed| parsed.ident.as_str() == "alias")
+                .map(|parsed| parsed.param.map(|v| v.value()))
+                .collect::<Option<Vec<_>>>();
```
chrisduerr, line 110: This will be `None` if *any* of the attributes has no value. That seems incorrect to me.

It should just be collected into a vec and empty elements should be skipped. Or even better a compile error is produced if there is no value.

alacritty_config_derive/src/serde_replace.rs
```diff
@@ -98,9 +99,28 @@ fn match_arms<T>(fields: &Punctuated<Field, T>) -> TokenStream2 {
...
+                Some(aliases) => {
+                    stream.extend(quote! {
+                            #(#aliases)|* | #literal => alacritty_config::SerdeReplace::replace(&mut
+                    self.#ident, next_value)?,         });
+                },
+                None => {
+                    stream.extend(quote! {
+                        #literal => alacritty_config::SerdeReplace::replace(&mut self.#ident,
+            next_value)?,         });
+                },
```
chrisduerr, line 122: This branching is technically unnecessary.

alacritty_config_derive/tests/config.rs
```diff
@@ -189,6 +194,30 @@ fn replace_derive() {
...
+#[test]
+fn replace_derive_using_multiple_aliases() {
+    let mut test = Test::default();
+
+    for (alias, value) in [("multiple_alias1", 6), ("multiple_alias2", 7)] {
+        let toml_value = toml::from_str(&format!("{}={}", alias, value)).unwrap();
+        test.replace(toml_value).unwrap();
+
+        assert_eq!(test.multiple_alias_field, value);
+    }
```
chrisduerr, line 218: I feel like this would actually be simpler unrolled.

alacritty_config_derive/tests/config.rs
```diff
@@ -189,6 +194,30 @@ fn replace_derive() {
     assert_eq!(test.nesting.newtype, NewType(9));
 }
 
+#[test]
+fn replace_derive_using_alias() {
+    let mut test = Test::default();
+
+    assert_ne!(test.field1, 9);
+
+    let value = toml::from_str("noalias=9").unwrap();
```
chrisduerr, line 203: I'm not sure why this field has the alias "noalias", but would probably best to change that. It makes no sense to me.

chrisduerr (changes requested):

alacritty_config_derive/src/serde_replace.rs
placintaalexandru: @chrisduerr, so when encounter `#[config(alias = "")]` we should panic?

alacritty_config_derive/src/serde_replace.rs
```diff
@@ -98,9 +99,30 @@ fn match_arms<T>(fields: &Punctuated<Field, T>) -> TokenStream2 {
...
+                    }
+
+                    value
+                })
+                .collect::<Vec<_>>();
+
             stream.extend(quote! {
-                #literal => alacritty_config::SerdeReplace::replace(&mut self.#ident, next_value)?,
-            });
+                    #(#aliases)|* | #literal => alacritty_config::SerdeReplace::replace(&mut
```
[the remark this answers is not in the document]
placintaalexandru: sorry, i did not understand this one

alacritty_config_derive/src/serde_replace.rs
```diff
@@ -98,9 +99,30 @@ fn match_arms<T>(fields: &Punctuated<Field, T>) -> TokenStream2 {
...
+                .filter(|parsed| parsed.ident.as_str() == "alias")
+                .map(|parsed| {
+                    let value = parsed
+                        .param
+                        .unwrap_or_else(|| panic!("Field \"{}\" has no alias value", ident))
+                        .value();
+
+                    if value.is_empty() {
+                        panic!("Field \"{}\" has an empty alias value", ident);
+                    }
```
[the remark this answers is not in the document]
placintaalexandru: thanks for the comment. I did not know that. In addition I noticed that also the compilation message regarding the usage of multiple `#[config(flatten)]` attributes was not correctly reported: I could've used as many `#[config(flatten)]` inside a struct, but the compilation message was something weird because of the way how `replace_arms` was used directly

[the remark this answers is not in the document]
placintaalexandru: Now, the correct compilation messages are reported

alacritty_config_derive/src/serde_replace.rs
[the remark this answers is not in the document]
chrisduerr: Doesn't really matter, your solution works. I just posted this for future reference.

alacritty_config_derive/src/serde_replace.rs
```diff
@@ -46,36 +46,41 @@ pub fn derive_recursive<T>(
...
-                            let value = value.clone();
-
-                            match field.as_str() {
-                                #replace_arms
-                                _ => {
-                                    let error = format!("Field \"{}\" does not exist", field);
-                                    return Err(error.into());
-                                },
-                            }
+    match replace_arms {
```
chrisduerr, line 49: This is going to be **far** cleaner if you early return on error. Just to deal with the already ridiculous amount of indentation.

alacritty_config_derive/src/serde_replace.rs
```diff
@@ -88,19 +93,42 @@ fn match_arms<T>(fields: &Punctuated<Field, T>) -> TokenStream2 {
         let flatten = field
             .attrs
             .iter()
+            .filter(|attr| (*attr).path().is_ident("config"))
             .filter_map(|attr| attr.parse_args::<Attr>().ok())
             .any(|parsed| parsed.ident.as_str() == "flatten");
 
         if flatten && flattened_arm.is_some() {
-            return Error::new(ident.span(), MULTIPLE_FLATTEN_ERROR).to_compile_error();
```
chrisduerr, line 95: I'd assume this is what caused the errors previously, since the compile errors were converted too early?

alacritty_config_derive/src/serde_replace.rs
```diff
@@ -88,19 +93,42 @@ fn match_arms<T>(fields: &Punctuated<Field, T>) -> TokenStream2 {
...
+                .filter(|parsed| parsed.ident.as_str() == "alias")
+                .map(|parsed| {
+                    let value = parsed
+                        .param
+                        .ok_or_else(|| format!("Field \"{}\" has no alias value", ident))?
+                        .value();
+
+                    if value.is_empty() {
+                        return Err(format!("Field \"{}\" has an empty alias value", ident));
+                    }
```
chrisduerr, line 122: Bit of an odd question, but have you tested stuff like `" "`? So empty strings.

When converted to tokens the spaces should be ignored, so I'd assume this would bypass this error handling but still fail to compile. I'd imagine you'd have to `value.trim().is_empty()` to handle this properly?

chrisduerr (changes requested):

alacritty_config_derive/src/serde_replace.rs
placintaalexandru: Yes, inside the match statement, where `replace_arms` where used, the expanded was something like this:
```rust
match field {
     "cursor" | "backround" => {},
      <WEIRD_FLATTEN_ERROR>
}
```

so the compiler was complaining that the match was incorrect

alacritty_config_derive/src/serde_replace.rs
```diff
@@ -46,6 +46,12 @@ pub fn derive_recursive<T>(
         crate::generics_streams(&generics.params);
     let replace_arms = match_arms(&fields);
 
+    if let Err(e) = replace_arms {
+        return e.to_compile_error();
+    }
+
+    let replace_arms = replace_arms.unwrap();
```
chrisduerr, line 53:
```
let replace_arms = match whatever {
    Ok(replace_arms) => replace_arms,
    Err(...) => ...,
}
```

chrisduerr (changes requested):

chrisduerr (changes requested):
Just add a changelog entry and it should be good to go.

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Autokey no longer working with alacritty on X11
 - Freeze when moving window between monitors on Xfwm
 - Mouse cursor not changing on Wayland when cursor theme uses legacy cursor icon names
+- Deriving `ConfigDeserialize` takes the `#[config(alias = "...")]` attribute(s) into account
+- Compilation errors are correctly reported when deriving `ConfigDeserialize` macro
```
chrisduerr, line 32: We don't document things for developers, we document them for users.

So the message you should be adding is that config keys are now all available under their proper name.

chrisduerr (changes requested):

CHANGELOG.md
placintaalexandru: thank you. i did not know

chrisduerr (approved):

PR #7732, opened 2024-02-15
[RFC] Experimental retry-loop to ameliorate clipboard error on Windows

cwedgwood:
Sometimes setting the clipboard on Windows fails, this is an experiment to address this.

### DIFF
--- a/alacritty/src/clipboard.rs
+++ b/alacritty/src/clipboard.rs
@@ -63,9 +63,28 @@ impl Clipboard {
             _ => &mut self.clipboard,
         };
 
-        clipboard.set_contents(text.into()).unwrap_or_else(|err| {
-            warn!("Unable to store text in clipboard: {}", err);
-        });
+        // Sometimes updates to the Windows clipboard fails, IDKW so
+        // for now put in a retry-loop to mitigate this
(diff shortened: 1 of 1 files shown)


kchibisov:
Have you checked whether this is present with 5.0 version of the windows clipboard? In general, I think consumer of the API should do thing like that and what we should do is it indicate that the clipboard is busy?

chrisduerr (changes requested):
Definitely not something appropriate for including into Alacritty. This is just a dirty hack and there's no technical necessity for this as far as I'm aware.

cwedgwood:
> Have you checked whether this is present with 5.0 version of the windows clipboard? In general, I think consumer of the API should do thing like that and what we should do is it indicate that the clipboard is busy?

I have not, I couldn't get copypasta to accommodate this.

cwedgwood:
@kchibisov @chrisduerr 

As a point of reference after 4 weeks of using this change I hit a single incident where it failed (after 3 attempts), so implementation details aside, whilst the concept helps it is insufficient.

Searching for similar errors I see other reports of transient clipboard access problems.

For now, I am going to rebuild using the main branch in the hope that some of the dependencies there have inadvertently addressed this.



Thanks!

PR #7777, opened 2024-02-26
Add motion repeat for vi-mode

Monochrome-Sauce:
This pull request allows us to type `100j` to move 100 rows down, or `42l` to move 42 columns to the right. The `count` is applied for every `ViMotion` that changes the cursor's position.

### DIFF
--- a/alacritty/src/config/bindings.rs
+++ b/alacritty/src/config/bindings.rs
@@ -495,7 +495,6 @@ pub fn default_key_bindings() -> Vec<KeyBinding> {
         ArrowDown,                          +BindingMode::VI, ~BindingMode::SEARCH; ViMotion::Down;
         ArrowLeft,                          +BindingMode::VI, ~BindingMode::SEARCH; ViMotion::Left;
         ArrowRight,                         +BindingMode::VI, ~BindingMode::SEARCH; ViMotion::Right;
-        "0",                                +BindingMode::VI, ~BindingMode::SEARCH; ViMotion::First;
         "$",      ModifiersState::SHIFT,    +BindingMode::VI, ~BindingMode::SEARCH; ViMotion::Last;
         Home,                               +BindingMode::VI, ~BindingMode::SEARCH; ViMotion::First;
         End,                                +BindingMode::VI, ~BindingMode::SEARCH; ViMotion::Last;
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -206,13 +206,19 @@ impl Default for InlineSearchState {
     }
 }
 
+#[derive(Default)]
 pub struct MotionState {
+    /// Digit history.
+    ///
+    /// When a `ViMotion` is processed or a non-digit character is added,
+    /// `MotionState` discards its history of entered digits.
     history: String,
 }
 
+/// Vi motion state for repeating actions.
 impl MotionState {
     pub fn add_character(&mut self, ch: char) {
-        if ch.is_digit(10) {
+        if ch.is_ascii_digit() {
             self.history.push(ch);
         }
         else {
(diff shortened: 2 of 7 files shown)


Monochrome-Sauce:
Note: I removed `"0", +BindingMode::VI, ~BindingMode::SEARCH; ViMotion::First;` in `default_key_bindings()` because of the conflict with repeating a motion that includes a zero.

chrisduerr (changes requested):
What if I want to add a vi mode binding for a number?

The reason why this wasn't added is that it's fragile and extremely vi-specific, while our vi mode in general can use any combination of keyboard bindings.

Monochrome-Sauce:
Then, what about I make it trigger in a way similar to search-mode?
For example type `.` to take digits and then the first non-digit (or `Enter` followed by anything) will trigger the motion repeatedly.

Another option would be a setting in `alacritty.toml` to toggle between digits-as-binding (default) and digits-as-count.

chrisduerr:
> Then, what about I make it trigger in a way similar to search-mode?
For example type . to take digits and then the first non-digit (or Enter followed by anything) will trigger the motion repeatedly.

Now you're just making up things that nobody else does in this fashion. It's essentially completely unintuitive at which point you might as well just put it in a fork just for yourself.

> Another option would be a setting in alacritty.toml to toggle between digits-as-binding (default) and digits-as-count.

This will never happen.

kchibisov:
I think the way vim works is that it removes the option to enter that specific number once you add number as a binding. In a way that if you bind `4` and you try to start with `4` it'll dispatch, but if you start with other number and type `4` it'll work fine.

I don't really have opinion whether this should be or not included though.

chrisduerr:
> I think the way vim works is that it removes the option to enter that specific number once you add number as a binding. In a way that if you bind 4 and you try to start with 4 it'll dispatch, but if you start with other number and type 4 it'll work fine.

Vim bindings are very complicated, including case-by-case logic that adds timeouts for keybindings when conflicting bindings exist. This is does **NOT** work well in vim and there's zero reason to introduce that atrocity to Alacritty.

Monochrome-Sauce:
> > Then, what about I make it trigger in a way similar to search-mode?
> > For example type . to take digits and then the first non-digit (or Enter followed by anything) will trigger the motion repeatedly.
> 
> Now you're just making up things that nobody else does in this fashion. It's essentially completely unintuitive at which point you might as well just put it in a fork just for yourself.

Having the option for repeat is better than not.


> > Another option would be a setting in alacritty.toml to toggle between digits-as-binding (default) and digits-as-count.
> 
> This will never happen.

Why not?

chrisduerr:
> Having the option for repeat is better than not.

No it isn't. A poorly implemented feature is much worse than nothing.

Monochrome-Sauce:
Ok, so we disagree on one front.

Now, mind elaborating on why a toggle in `alacritty.toml` will never happen?

Monochrome-Sauce:
@chrisduerr 
At the very least tell me, what is your idea of a good implementation and how would it be different from the naive implementation? I would love to work on it.

chrisduerr:
>  At the very least tell me, what is your idea of a good implementation and how would it be different from the naive implementation? I would love to work on it.

I didn't add this to Alacritty because I don't think there is any implementation of this feature that is **good**.

Monochrome-Sauce:
What are your requirements for an implementation to be "good"?

chrisduerr:
> What are your requirements for an implementation to be "good"?

If it's confusing to users, it's bad. If it relies on timeouts, it's bad. If there's a chance that it breaks existing bindings, or conflicts with them, it's bad.

A good implementation shouldn't have any obvious drawbacks and be acceptable to push to **all** users without massive backlash.

Monochrome-Sauce:
Thanks for replying with honesty throughout this conversation.

Anyway, those are fair requirements. I'll try to think of a solution (as impossible as it may sound) in my spare time.

chrisduerr:
> So how does alacritty handle the 3rd and 4th requirements for new features?

I mean generally speaking if you're asking how we don't break *user's* bindings, then the answer is that default bindings are always overwritten by user bindings. If you're asking how we managed not to break support for old user bindings by adding new user binding features, then the answer is just that there never is any need for it.

PR #7787, opened 2024-02-29
cargo update smithay-client-toolkit

Friz64:
alacritty will now benefit from https://github.com/Smithay/client-toolkit/pull/427.

Should we run a general `cargo update` instead?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Modifiers occasionally getting desynced on X11
 - Autokey no longer working with alacritty on X11
 - Freeze when moving window between monitors on Xfwm
+- Mouse cursor not changing on Wayland when cursor theme uses legacy cursor icon names
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


kchibisov:
Thanks, I thought that it was long merged, but it just everyone forgot about the release.

>Should we run a general cargo update instead?

No, it'll make backport a bit harder.

kchibisov (changes requested):
Could you add a changelog entry for that?

Friz64:
like this?

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - No unused-key warnings will be emitted for OS-specific config keys
 - Use built-in font for sextant symbols from `U+1FB00` to `U+1FB3B`
 - Kitty encoding is not used anymore for uncommon keys unless the protocol enabled
+- Bumped smithay-client-toolkit to 0.18.1
```
kchibisov, line 36: It should be in a `FIXED` section and say something like

```suggestion
- Mouse cursor not changing with cursor themes using legacy cursor names on Wayland
```

kchibisov (approved):

Issue #7778, opened 2024-02-27
Add a way to set env variables for alacritty's child process only


SomeoneToIgnore:
Somewhat related to https://github.com/alacritty/alacritty/issues/7753 since caused by the same use case: Zed editor uses alacritty to spawn things and recently it gained the ability to spawn arbitrary `bash -c foobar.sh` or `python3 something.py` commands using Alacritty's `alacritty_terminal::tty::Shell::new`

Certain tasks would need more env variables, e.g. various python scripts may require different venv set up taken from the project.
For that, certain env variables have to be set in order to run the task properly.
If Zed uses `std::env::` API, it will set those for the entire Zed process which may cause bad side effects. 

How appropriate would it be to extend the `alacritty_terminal::tty::Shell` struct with `env: HashMap<String, String>` config and set it in https://github.com/alacritty/alacritty/blob/37349c4566a1431b5becbb649dec536ecbe18e71/alacritty_terminal/src/tty/unix.rs#L228-L230
?

kchibisov:
Don't see why not, given that it's racy to do manually and rely on inheritance. Probably we shouldn't set `ALACRITTY_WINDOW_ID` and require `window_id` in the first place, but do it on the caller side.

There's `options` where you can add `env`.

PR #7788, opened 2024-02-29
Allow setting terminal env vars via pty options

SomeoneToIgnore:
Closes https://github.com/alacritty/alacritty/issues/7778 by adding another field to `alacritty_terminal::tty::Options` and using those in `unix.rs` during the shell command building.

Some questions came up along the way:

* Are `tty::setup_env` vars supposed to be process-specific too in the unix case?
https://github.com/alacritty/alacritty/blob/8218c203032e6f42a20f31112c5e72abf6c3af83/alacritty_terminal/src/tty/mod.rs#L92 

Seems like not, but wanted to double check.

* There are no explicit *.md doc files related to alacritty_terminal, but should I mention this breaking change into root CHANGELOG.md? Something else?

* re: https://github.com/alacritty/alacritty/issues/7778#issuecomment-1966045017 change
Was the idea to check that new `env` for the id? I'm glad to omit that bit out of the PR unless it's not something simple to do along the way.

### DIFF
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -167,7 +167,13 @@ impl UiConfig {
     /// Derive [`PtyOptions`] from the config.
     pub fn pty_config(&self) -> PtyOptions {
         let shell = self.shell.clone().map(Into::into);
-        PtyOptions { shell, working_directory: self.working_directory.clone(), hold: false }
+        PtyOptions {
+            shell,
+            working_directory: self.working_directory.clone(),
+            hold: false,
+            #[cfg(not(windows))]
+            env: self.env.clone(),
+        }
     }
 
     /// Generate key bindings for all keyboard hints.
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -172,7 +172,7 @@ impl UiConfig {
             working_directory: self.working_directory.clone(),
             hold: false,
             #[cfg(not(windows))]
-            env: self.env.clone(),
+            env: HashMap::new(),
         }
     }
 
--- a/alacritty/src/config/ui_config.rs
+++ b/alacritty/src/config/ui_config.rs
@@ -171,7 +171,6 @@ impl UiConfig {
             shell,
             working_directory: self.working_directory.clone(),
             hold: false,
-            #[cfg(not(windows))]
             env: HashMap::new(),
         }
     }
--- a/alacritty_terminal/src/tty/mod.rs
+++ b/alacritty_terminal/src/tty/mod.rs
@@ -1,5 +1,6 @@
 //! TTY related functionality.
 
+use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::{env, io};
@@ -29,6 +30,10 @@ pub struct Options {
 
     /// Remain open after child process exits.
     pub hold: bool,
+
+    /// Extra environment variables.
+    #[cfg(not(windows))]
+    pub env: HashMap<String, String>,
 }
 
 /// Shell options.
--- a/alacritty_terminal/src/tty/mod.rs
+++ b/alacritty_terminal/src/tty/mod.rs
@@ -32,7 +32,6 @@ pub struct Options {
     pub hold: bool,
 
     /// Extra environment variables.
-    #[cfg(not(windows))]
     pub env: HashMap<String, String>,
 }
 
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -1,4 +1,4 @@
-use log::info;
+use log::{info, warn};
 use std::collections::{HashMap, HashSet};
 use std::ffi::OsStr;
 use std::io::Error;
@@ -256,12 +256,17 @@ fn convert_custom_env(custom_env: &HashMap<String, String>) -> Option<Vec<u16>> 
     let mut converted_block = Vec::new();
     let mut all_env_keys = HashSet::new();
     for (custom_key, custom_value) in custom_env {
-        let custom_key = OsStr::new(custom_key);
-        if all_env_keys.insert(custom_key.to_ascii_uppercase()) {
+        let custom_key_os = OsStr::new(custom_key);
+        if all_env_keys.insert(custom_key_os.to_ascii_uppercase()) {
             add_windows_env_key_value_to_block(
                 &mut converted_block,
-                custom_key,
+                custom_key_os,
                 OsStr::new(&custom_value),
+            );
+        } else {
+            warn!(
+                "Omitting environment variable pair with duplicate key: \
+                 '{custom_key}={custom_value}'"
             );
         }
     }
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -241,8 +241,9 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
     Some(Pty::new(conpty, conout, conin, child_watcher))
 }
 
-// https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa#parameters
-// https://learn.microsoft.com/sk-SK/previous-versions/troubleshoot/windows/win32/createprocess-cannot-eliminate-duplicate-variables#environment-variables
+// Windows environment variables are case-insensitive, and
+// https://learn.microsoft.com/en-us/previous-versions/troubleshoot/windows/win32/createprocess-cannot-eliminate-duplicate-variables#environment-variables
+// mentions that the caller is responsible for deduplicating environment variables, so do that here while converting.
 fn convert_custom_env(custom_env: &HashMap<String, String>) -> Option<Vec<u16>> {
     // Windows inherits parent's env when no `lpEnvironment` parameter is specified.
     if custom_env.is_empty() {
@@ -277,7 +278,9 @@ fn convert_custom_env(custom_env: &HashMap<String, String>) -> Option<Vec<u16>> 
     Some(converted_block)
 }
 
-// According to the `lpEnvironment` parameter description in the link above:
+// According to the `lpEnvironment` parameter description:
+// https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa#parameters
+//
 // > An environment block consists of a null-terminated block of null-terminated strings. Each
 // string is in the following form:
 // >
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -1,9 +1,12 @@
 use log::info;
+use std::collections::{HashMap, HashSet};
+use std::ffi::OsStr;
 use std::io::Error;
+use std::os::windows::ffi::OsStrExt;
 use std::os::windows::io::IntoRawHandle;
 use std::{mem, ptr};
 
-use windows_sys::core::{HRESULT, PWSTR};
+use windows_sys::core::{IInspectable, HRESULT, PWSTR};
 use windows_sys::Win32::Foundation::{HANDLE, S_OK};
 use windows_sys::Win32::System::Console::{
     ClosePseudoConsole, CreatePseudoConsole, ResizePseudoConsole, COORD, HPCON,
@@ -13,8 +16,8 @@ use windows_sys::{s, w};
 
 use windows_sys::Win32::System::Threading::{
     CreateProcessW, InitializeProcThreadAttributeList, UpdateProcThreadAttribute,
-    EXTENDED_STARTUPINFO_PRESENT, PROCESS_INFORMATION, PROC_THREAD_ATTRIBUTE_PSEUDOCONSOLE,
-    STARTF_USESTDHANDLES, STARTUPINFOEXW, STARTUPINFOW,
+    CREATE_UNICODE_ENVIRONMENT, EXTENDED_STARTUPINFO_PRESENT, PROCESS_INFORMATION,
+    PROC_THREAD_ATTRIBUTE_PSEUDOCONSOLE, STARTF_USESTDHANDLES, STARTUPINFOEXW, STARTUPINFOW,
 };
 
 use crate::event::{OnResize, WindowSize};
@@ -200,7 +203,15 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
 
     let cmdline = win32_string(&cmdline(config));
     let cwd = config.working_directory.as_ref().map(win32_string);
-
+    let mut creation_flags = EXTENDED_STARTUPINFO_PRESENT;
+    let custom_env_block = convert_custom_env(&config.env);
+    let custom_env_block_pointer = match &custom_env_block {
+        Some(custom_env_block) => {
+            creation_flags |= CREATE_UNICODE_ENVIRONMENT;
+            custom_env_block.as_ptr() as IInspectable
+        },
+        None => ptr::null_mut(),
+    };
     let mut proc_info: PROCESS_INFORMATION = unsafe { mem::zeroed() };
     unsafe {
         success = CreateProcessW(
@@ -209,8 +220,8 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
             ptr::null_mut(),
             ptr::null_mut(),
             false as i32,
-            EXTENDED_STARTUPINFO_PRESENT,
-            ptr::null_mut(),
+            creation_flags,
+            custom_env_block_pointer,
             cwd.as_ref().map_or_else(ptr::null, |s| s.as_ptr()),
             &mut startup_info_ex.StartupInfo as *mut STARTUPINFOW,
             &mut proc_info as *mut PROCESS_INFORMATION,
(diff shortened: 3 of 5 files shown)


alacritty_terminal/src/tty/mod.rs
```diff
@@ -29,6 +30,10 @@ pub struct Options {
 
     /// Remain open after child process exits.
     pub hold: bool,
+
+    /// Extra environment variables.
+    #[cfg(not(windows))]
```
kchibisov, line 35: Why we don't do so on windows?

alacritty/src/config/ui_config.rs
```diff
@@ -167,7 +167,13 @@ impl UiConfig {
     /// Derive [`PtyOptions`] from the config.
     pub fn pty_config(&self) -> PtyOptions {
         let shell = self.shell.clone().map(Into::into);
-        PtyOptions { shell, working_directory: self.working_directory.clone(), hold: false }
+        PtyOptions {
+            shell,
+            working_directory: self.working_directory.clone(),
+            hold: false,
+            #[cfg(not(windows))]
+            env: self.env.clone(),
```
kchibisov, line 175: This should be _empty_ since we set env for alacritty iself so it's inherited.

The rest could be a bit confusing with the way things work currently.

kchibisov (changes requested):

alacritty_terminal/src/tty/mod.rs
SomeoneToIgnore: Mainly due to my 0 knowledge of how things work for Windows.

As far as I see, there are two options to spawn a terminal in Windows:
* poweshell commandline 
https://github.com/alacritty/alacritty/blob/8218c203032e6f42a20f31112c5e72abf6c3af83/alacritty_terminal/src/tty/windows/mod.rs#L125

would it be ok to use `$env1:value1;$env2:value2;..everything else` or what the approach would be

* `ConptyApi`
https://github.com/alacritty/alacritty/blob/8218c203032e6f42a20f31112c5e72abf6c3af83/alacritty_terminal/src/tty/windows/conpty.rs#L107

Presumably, `CreateProcessW` has to get more parameters?

kchibisov: There's a command called `CreateProcessW` which we use, it has option to pass environment. Keep in mind that it sounds like `lpEnvironment` discards all the parent process envs, which we likely shouldn't be doing and rather merge/overwrite, but it needs testing.

It's the **only** way, the other one is the default thing which we shouldn't speculate with.

SomeoneToIgnore: I'm somewhat stuck with forming the right parameter form for the env variables.
Here's a simple example of an attempt to add some ad-hoc env variable as an argument

```diff
diff --git a/alacritty_terminal/src/tty/windows/conpty.rs b/alacritty_terminal/src/tty/windows/conpty.rs
index 9731b4f0..c3605316 100644
--- a/alacritty_terminal/src/tty/windows/conpty.rs
+++ b/alacritty_terminal/src/tty/windows/conpty.rs
@@ -1,9 +1,11 @@
 use log::info;
+use std::ffi::{OsStr, OsString};
 use std::io::Error;
+use std::os::windows::ffi::OsStrExt;
 use std::os::windows::io::IntoRawHandle;
 use std::{mem, ptr};
 
-use windows_sys::core::{HRESULT, PWSTR};
+use windows_sys::core::{IInspectable, HRESULT, PWSTR};
 use windows_sys::Win32::Foundation::{HANDLE, S_OK};
 use windows_sys::Win32::System::Console::{
     ClosePseudoConsole, CreatePseudoConsole, ResizePseudoConsole, COORD, HPCON,
@@ -201,6 +203,24 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
     let cmdline = win32_string(&cmdline(config));
     let cwd = config.working_directory.as_ref().map(win32_string);
 
+    let env_block_wide: Vec<u16> = //config.env
+        Some(("FOO", "FOO_VALUE"))
+        .iter()
+        .flat_map(|(key, value)| {
+            let wide_string: Vec<u16> = OsStr::new(key)
+                .encode_wide()
+                .chain(std::iter::once(b'=' as u16))
+                .chain(OsStr::new(value).encode_wide())
+                .chain(std::iter::once(0)) // Null terminator for the individual string
+                .collect();
+            wide_string
+        })
+        .chain(std::iter::once(0)) // Add an extra null terminator to the end
+        .collect();
+
+    let another_env_block_wide =
+        another::format_env_block(&[(OsString::from("FOO"), OsString::from("FOO_VALUE"))]);
+
     let mut proc_info: PROCESS_INFORMATION = unsafe { mem::zeroed() };
     unsafe {
         success = CreateProcessW(
@@ -210,7 +230,9 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
             ptr::null_mut(),
             false as i32,
             EXTENDED_STARTUPINFO_PRESENT,
-            ptr::null_mut(),
+            Some(another_env_block_wide).as_ref().map(|a| a.as_ptr()).unwrap_or(ptr::null())
+                as IInspectable,
+            // ptr::null_mut(),
             cwd.as_ref().map_or_else(ptr::null, |s| s.as_ptr()),
             &mut startup_info_ex.StartupInfo as *mut STARTUPINFOW,
             &mut proc_info as *mut PROCESS_INFORMATION,
@@ -249,3 +271,43 @@ impl From<WindowSize> for COORD {
         COORD { X: columns as i16, Y: lines as i16 }
     }
 }
+
+mod another {
+    use std::{
+        collections::HashSet,
+        ffi::{OsStr, OsString},
+        os::windows::ffi::{OsStrExt, OsStringExt},
+    };
+
+    pub fn format_env_block(env: &[(OsString, OsString)]) -> Vec<u16> {
+        fn to_uppercase(s: &OsStr) -> OsString {
+            OsString::from_wide(
+                &s.encode_wide()
+                    .map(
+                        |c| {
+                            if c < 128 {
+                                (c as u8 as char).to_ascii_uppercase() as u16
+                            } else {
+                                c
+                            }
+                        },
+                    )
+                    .collect::<Vec<_>>(),
+            )
+        }
+        let mut pruned: Vec<_> = {
+            let mut seen = HashSet::<OsString>::new();
+            env.iter().rev().filter(|&(k, _)| seen.insert(to_uppercase(k))).collect()
+        };
+        pruned.reverse();
+        let mut block = vec![];
+        for (k, v) in pruned {
+            block.extend(k.encode_wide());
+            block.push('=' as u16);
+            block.extend(v.encode_wide());
+            block.push(0);
+        }
+        block.push(0);
+        block
+    }
+}
```


with the `mod another` using https://github.com/hniksic/rust-subprocess/blob/5e89ac093f378bcfc03c69bdb1b4bcacf4313ce4/src/popen.rs#L1080

`Some(another_env_block_wide).as_ref().map(|a| a.as_ptr()).unwrap_or(ptr::null())` contraption also comes from that code, but I did try a regular version too with the same result:

```
panicked at alacritty_terminal\src\tty\windows\conpty.rs:257:5:
Unable to spawn shell: The parameter is incorrect. (os error 87)
error: process didn't exit successfully: target\debug\alacritty.exe (exit code: 0xc000013a, STATUS_CONTROL_C_EXIT)
```

So looks like I convert the parameter incorrectly, but I do not understand why does not it work for `rust-subprocess` crate code too.
Will poke around more, but really appreciate any pointers.

SomeoneToIgnore: And if I add the `|= CREATE_UNICODE_ENVIRONMENT` to the `dwcreationflags`, it does not output the error, but very briefly opens and closes the process, which also seems wrong..

SomeoneToIgnore: Added, thank you for the pointers.

https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa#parameters

* we do indeed inherit the env on default
> If this parameter is NULL, the new process uses the environment of the calling process.

* for non-defaults, have to deduplicate the variables, case-insensitive
https://learn.microsoft.com/sk-SK/previous-versions/troubleshoot/windows/win32/createprocess-cannot-eliminate-duplicate-variables#environment-variables

* I've cast the pointer to `IInspectable` logical type, but not sure if it's applicable, but other `*mut c_void` types there neither look applicable.

Seems to work:
telegram-cloud-photo-size-2-5407061907099212075-y

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -230,6 +241,35 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
...
+fn convert_custom_env(custom_env: &HashMap<String, String>) -> Option<Vec<u16>> {
+    let mut result = Vec::new();
+    if custom_env.is_empty() {
+        return None;
+    }
+
+    let mut all_env_keys = HashSet::new();
+    for (custom_key, custom_value) in custom_env {
+        let custom_key = OsStr::new(custom_key);
+        if all_env_keys.insert(custom_key.to_ascii_uppercase()) {
```
kchibisov, line 253: Is it required to `uppercase` it? Since on e.g. linux you can mix case.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -230,6 +241,35 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
...
+            result.extend(custom_key.encode_wide());
+            result.push('=' as u16);
+            result.extend(OsStr::new(custom_value).encode_wide());
+            result.push(0);
+        }
+    }
+    for (inherited_key, inherited_value) in std::env::vars_os() {
```
kchibisov, line 260: I'm also pretty sure that you can merge both iterators, just need a `map` in one and then `chain` the other.

```suggestion
    // Pull the current a process environment after to not overwrite user provided one.
    for (inherited_key, inherited_value) in std::env::vars_os() {
```

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -230,6 +241,35 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
...
+        if all_env_keys.insert(inherited_key.to_ascii_uppercase()) {
+            result.extend(inherited_key.encode_wide());
+            result.push('=' as u16);
+            result.extend(inherited_value.encode_wide());
+            result.push(0);
+        }
+    }
+    result.push(0);
```
kchibisov, line 268: Document why it's done like that. Like it's not really clear why you need `0`, `0` at the end.

kchibisov (changes requested):

alacritty_terminal/src/tty/windows/conpty.rs
SomeoneToIgnore: Windows env names are case-insensitive and we're required to deduplicate those before sending to the process, so some, upper or lower, common case is needed for that?
Windows, at least before, used to prefer upper case, e.g. PATH, USERNAME, TEMP.

On Linux, env names are case-sensitive and there's no need to do anything like that, as HashMap<String, String> deduplicates things automatically.

alacritty_terminal/src/tty/windows/conpty.rs
SomeoneToIgnore: The problem with that is that `custom_ref` is a reference type, but `std::env::var_os()` is an iterator of owned strings.
So it's not that easy to combine those, but I've extracted the string creation method and documented it, as proposed.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -230,6 +241,54 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
...
+                &inherited_value,
+            );
+        }
+    }
+
+    converted_block.push(0);
+    Some(converted_block)
+}
+
+// According to the `lpEnvironment` parameter description in the link above:
```
kchibisov, line 280: Could you just move the link here instead?

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -230,6 +241,54 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
     Some(Pty::new(conpty, conout, conin, child_watcher))
 }
 
+// https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa#parameters
+// https://learn.microsoft.com/sk-SK/previous-versions/troubleshoot/windows/win32/createprocess-cannot-eliminate-duplicate-variables#environment-variables
```
kchibisov, line 245: Use `en-us` in links.

kchibisov (changes requested):
Just some minor nits.

alacritty_terminal/src/tty/windows/conpty.rs
SomeoneToIgnore: Sure, also added a few words about case-insensitivity for the remaining link.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -241,8 +241,9 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
     Some(Pty::new(conpty, conout, conin, child_watcher))
 }
 
-// https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa#parameters
-// https://learn.microsoft.com/sk-SK/previous-versions/troubleshoot/windows/win32/createprocess-cannot-eliminate-duplicate-variables#environment-variables
+// Windows environment variables are case-insensitive, and
+// https://learn.microsoft.com/en-us/previous-versions/troubleshoot/windows/win32/createprocess-cannot-eliminate-duplicate-variables#environment-variables
+// mentions that the caller is responsible for deduplicating environment variables, so do that here while converting.
```
chrisduerr, line 246: Docs should be written either using markdown syntax with putting long links beyond the URL, or since these comments are never rendered in markdown they should be written in plain text while adding references below, the way you did it with your other comment.

In its current state this is unreadable.

chrisduerr (changes requested):

alacritty_terminal/src/tty/windows/conpty.rs
SomeoneToIgnore: Ok, moved the URL in the very bottom of the comment.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -200,7 +203,15 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
...
-
+    let mut creation_flags = EXTENDED_STARTUPINFO_PRESENT;
+    let custom_env_block = convert_custom_env(&config.env);
+    let custom_env_block_pointer = match &custom_env_block {
+        Some(custom_env_block) => {
+            creation_flags |= CREATE_UNICODE_ENVIRONMENT;
+            custom_env_block.as_ptr() as *mut std::ffi::c_void
+        },
+        None => ptr::null_mut(),
+    };
```
chrisduerr, line 214: Please don't just put everything in massive blocks with no documentation again. This could is much more readable with sensible spacing and a comment or two. Otherwise nobody will know what is going on at a glance.

alacritty_terminal/src/tty/windows/conpty.rs
```diff
@@ -230,6 +241,58 @@ pub fn new(config: &Options, window_size: WindowSize) -> Option<Pty> {
...
+    // Windows inherits parent's env when no `lpEnvironment` parameter is specified.
+    if custom_env.is_empty() {
+        return None;
+    }
+
+    let mut converted_block = Vec::new();
+    let mut all_env_keys = HashSet::new();
+    for (custom_key, custom_value) in custom_env {
+        let custom_key = OsStr::new(custom_key);
+        if all_env_keys.insert(custom_key.to_ascii_uppercase()) {
```
chrisduerr, line 258: `custom_env` is a `HashMap`, so `custom_key` is known to not exist in `all_env_keys`. What is the point of checking if it was already inserted? And even if it was, wouldn't we want to overwrite it rather than use the *first* value specified?

chrisduerr (changes requested):

alacritty_terminal/src/tty/windows/conpty.rs
SomeoneToIgnore: > What is the point of checking if it was already inserted

To deduplicate it case-insensitively: `HashMap::from([("one_key", "one_value"), ("ONE_KEY", "another value")])` will contain two keys but from Windows' standpoint, both keys are the same.

> wouldn't we want to overwrite it rather than use the first value specified

Since it's a `HashMap`, is there such thing as the first value, really? 

I have thought that it's dangerous to rely on hash-based collections' order of iteration, so we cannot safely state that the user env input deserialized into a `HashMap` will have the same order of entries.

What we do want, is to accept user only env vars in case of key matching case-insensitively, so I have used the same code for both to make it look less bloated.

------------------

I'm somewhat disenchanted now: there's a comment above, saying 
```
// Windows environment variables are case-insensitive, and the caller is responsible for
// deduplicating environment variables, so do that here while converting.
```
and here's the `to_ascii_uppercase` call — should I remove the entire doc then, if it's useless? Or adjust somehow?

alacritty_terminal/src/tty/windows/conpty.rs
SomeoneToIgnore: Sorry, quite hard to come from another world with another style requirements.

I've added a small comment (which is also dangerous in this repo, given how many review notes were for those, but let's see) and a new linebreak to separate the whole arguments' preparation block out.

alacritty_terminal/src/tty/windows/conpty.rs
chrisduerr: > and here's the to_ascii_uppercase call — should I remove the entire doc then, if it's useless? Or adjust somehow?

Nah it's fine, I just didn't consider the implications of this for the initial hashmap.

It might make sense to warn in the else case though, which also would make it more clear to anyone reading the code? People should be made aware that environment variables they're defining have no effect, though maybe that should just be a debug statement for configs shared with linux?

SomeoneToIgnore: Yeah, somewhat tricky to log it for all cases, I pushed a change to WARN log only windows part only for duplicates in the user config.

* Unix part with case sensitivity makes `HashMap<String, String>` interface "intuitive" and all deduplications already happen in the user code
* Both Windows and Unix versions are expected to deduplicate OS env vars, so no logging is needed there
* The "unintuitive" case-insensitive Windows part got logs for the rest

SomeoneToIgnore:
I think I've addressed all the feedback (thank you for that again), but let me know if something is hanging on my side still.

kchibisov:
@SomeoneToIgnore you should re-request review with github UI when you think so.

chrisduerr (approved):

chrisduerr:
@kchibisov Feel free to merge if you have nothing to add on.

kchibisov (approved):

Issue #7753, opened 2024-02-19
A way to know exit code after `alacritty_terminal::tty::Shell::new` finishes


SomeoneToIgnore:
I use `alacritty_terminal::tty::Shell::new` options when creating `tty::new` for some bash-like command.
It works well, shows command's output in the terminal and sends `Event::Exit` after the command finishes execution, but there seems to be no way to know what is the `ExitStatus` of that command?

There's a way to get a reference to a `Child` via `Pty::child` (https://docs.rs/alacritty_terminal/latest/alacritty_terminal/tty/struct.Pty.html#method.child)
but the pty itself has to be moved into the `EventLoop` in order to init the terminal:
https://docs.rs/alacritty_terminal/latest/alacritty_terminal/event_loop/struct.EventLoop.html#method.new and the `Child` is not `Clone`.

I could get child's pid, but using `libc` on that pid as
```rust
fn get_exit_code(pid: u32) -> Option<i32> {
    let mut status = 0;
    let waitpid_result = unsafe { libc::waitpid(pid as libc::pid_t, &mut status, libc::WNOHANG) };
    if waitpid_result == -1 {
        None
    } else {
        if libc::WIFEXITED(status) {
            Some(libc::WEXITSTATUS(status))
        } else if libc::WIFSIGNALED(status) {
            None
        } else {
            None
        }
    }
}
```
does not seem to yield any result, I get `None`.

Is there a way to know the `ExitStatus` of that bash-like command? 
If not, would it be appropriate to add a way to do so? Should it be a field in the `Event::Exit` or is there a better way to design this?

kchibisov:
There's no way to do so, and adding exit code event sounds like not a bad idea.

PR #7794, opened 2024-03-01
Send exit code events on child process exit

SomeoneToIgnore:
Closes https://github.com/alacritty/alacritty/issues/7753

### DIFF
--- a/alacritty_terminal/src/event.rs
+++ b/alacritty_terminal/src/event.rs
@@ -52,6 +52,9 @@ pub enum Event {
 
     /// Shutdown request.
     Exit,
+
+    /// Child process exited with an error code.
+    ExitCode(i32),
 }
 
 impl Debug for Event {
@@ -69,6 +72,7 @@ impl Debug for Event {
             Event::Wakeup => write!(f, "Wakeup"),
             Event::Bell => write!(f, "Bell"),
             Event::Exit => write!(f, "Exit"),
+            Event::ExitCode(code) => write!(f, "ExitCode({code})"),
         }
     }
 }
--- a/alacritty_terminal/src/event.rs
+++ b/alacritty_terminal/src/event.rs
@@ -54,7 +54,7 @@ pub enum Event {
     Exit,
 
     /// Child process exited with an error code.
-    ExitCode(i32),
+    ChildExit(i32),
 }
 
 impl Debug for Event {
@@ -72,7 +72,7 @@ impl Debug for Event {
             Event::Wakeup => write!(f, "Wakeup"),
             Event::Bell => write!(f, "Bell"),
             Event::Exit => write!(f, "Exit"),
-            Event::ExitCode(code) => write!(f, "ExitCode({code})"),
+            Event::ChildExit(code) => write!(f, "ChildExit({code})"),
         }
     }
 }
--- a/alacritty_terminal/src/event.rs
+++ b/alacritty_terminal/src/event.rs
@@ -53,8 +53,8 @@ pub enum Event {
     /// Shutdown request.
     Exit,
 
-    /// Child process exited with an error code.
-    ExitCode(i32),
+    /// Child process exited, successfully or not.
+    ChildExit(bool),
 }
 
 impl Debug for Event {
@@ -72,7 +72,7 @@ impl Debug for Event {
             Event::Wakeup => write!(f, "Wakeup"),
             Event::Bell => write!(f, "Bell"),
             Event::Exit => write!(f, "Exit"),
-            Event::ExitCode(code) => write!(f, "ExitCode({code})"),
+            Event::ChildExit(sucess) => write!(f, "ChildExit({sucess})"),
         }
     }
 }
--- a/alacritty_terminal/src/event.rs
+++ b/alacritty_terminal/src/event.rs
@@ -53,8 +53,8 @@ pub enum Event {
     /// Shutdown request.
     Exit,
 
-    /// Child process exited, successfully or not.
-    ChildExit(bool),
+    /// Child process exited with an error code.
+    ExitCode(i32),
 }
 
 impl Debug for Event {
@@ -72,7 +72,7 @@ impl Debug for Event {
             Event::Wakeup => write!(f, "Wakeup"),
             Event::Bell => write!(f, "Bell"),
             Event::Exit => write!(f, "Exit"),
-            Event::ChildExit(sucess) => write!(f, "ChildExit({sucess})"),
+            Event::ExitCode(code) => write!(f, "ExitCode({code})"),
         }
     }
 }
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -259,7 +259,8 @@ where
                 for event in events.iter() {
                     match event.key {
                         tty::PTY_CHILD_EVENT_TOKEN => {
-                            if let Some(tty::ChildEvent::Exited(code)) = self.pty.next_child_event()
+                            if let Some(tty::ChildEvent::Exited(success)) =
+                                self.pty.next_child_event()
                             {
                                 if self.hold {
                                     // With hold enabled, make sure the PTY is drained.
@@ -268,9 +269,7 @@ where
                                     // Without hold, shutdown the terminal.
                                     self.terminal.lock().exit();
                                 }
-                                if let Some(code) = code {
-                                    self.event_proxy.send_event(Event::ExitCode(code));
-                                }
+                                self.event_proxy.send_event(Event::ChildExit(success));
                                 self.event_proxy.send_event(Event::Wakeup);
                                 break 'event_loop;
                             }
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -261,15 +261,15 @@ where
                         tty::PTY_CHILD_EVENT_TOKEN => {
                             if let Some(tty::ChildEvent::Exited(code)) = self.pty.next_child_event()
                             {
+                                if let Some(code) = code {
+                                    self.event_proxy.send_event(Event::ChildExit(code));
+                                }
                                 if self.hold {
                                     // With hold enabled, make sure the PTY is drained.
                                     let _ = self.pty_read(&mut state, &mut buf, pipe.as_mut());
                                 } else {
                                     // Without hold, shutdown the terminal.
                                     self.terminal.lock().exit();
-                                }
-                                if let Some(code) = code {
-                                    self.event_proxy.send_event(Event::ChildExit(code));
                                 }
                                 self.event_proxy.send_event(Event::Wakeup);
                                 break 'event_loop;
--- a/alacritty_terminal/src/event_loop.rs
+++ b/alacritty_terminal/src/event_loop.rs
@@ -259,8 +259,7 @@ where
                 for event in events.iter() {
                     match event.key {
                         tty::PTY_CHILD_EVENT_TOKEN => {
-                            if let Some(tty::ChildEvent::Exited(success)) =
-                                self.pty.next_child_event()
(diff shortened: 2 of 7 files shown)


chrisduerr:
> requires having a process handle, which is absent in child_exit_callback

The context passed into the callback is up to us. So you can maybe just pass in the process handle yourself?

SomeoneToIgnore:
Thank you, that seems to work and now I have the codes in both platforms.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -31,7 +32,11 @@ extern "system" fn child_exit_callback(ctx: *mut c_void, timed_out: BOOLEAN) {
     }
 
     let event_tx: Box<_> = unsafe { Box::from_raw(ctx as *mut ChildExitSender) };
-    let _ = event_tx.sender.send(ChildEvent::Exited);
+    let child_handle = event_tx.child_handle.load(Ordering::Relaxed) as HANDLE;
+    let mut exit_code = 0_u32;
```
kchibisov, line 36: Could you use the actual type of this value? `LPDWORD` or however windows-sys calls it.

alacritty_terminal/src/tty/mod.rs
```diff
@@ -68,7 +68,7 @@ pub trait EventedReadWrite {
 #[derive(Debug, PartialEq, Eq)]
 pub enum ChildEvent {
     /// Indicates the child has exited.
-    Exited,
+    Exited(Option<i32>),
```
kchibisov, line 71: Do we really need option instead of just saying that `0` is fine? Or maybe we should properly propagate an error from these calls?

alacritty_terminal/src/event_loop.rs
```diff
@@ -259,15 +259,18 @@ where
...
                                     // With hold enabled, make sure the PTY is drained.
                                     let _ = self.pty_read(&mut state, &mut buf, pipe.as_mut());
                                 } else {
                                     // Without hold, shutdown the terminal.
                                     self.terminal.lock().exit();
                                 }
-
+                                if let Some(code) = code {
+                                    self.event_proxy.send_event(Event::ExitCode(code));
+                                }
```
kchibisov, line 273: I think this should be not `ExitCode`, but `ChildExit` or something like that and then we follow it with e.g `exit`, so your sending should be above, since right now you send

```
Exit,
ExitCode
```

Which is a bit _weird_, also you may want to know the exit code for `--hold`, but then `ExitCode` is weird, because it could mean that terminal exited with that code.

kchibisov (changes requested):

alacritty_terminal/src/tty/windows/child.rs
SomeoneToIgnore: There seems to be no `LPDWORD` or even `DWORD` type in the `windows_sys` crate, and looks like tjeu use `*mut u32` directly:

* https://docs.rs/windows-sys/latest/windows_sys/Win32/System/Threading/fn.GetExitCodeProcess.html

* https://github.com/microsoft/windows-rs/blob/1f40da8ffc44adba3f9a866287108e80223d0a81/crates/libs/sys/src/Windows/Win32/System/Threading/mod.rs#L140

* https://github.com/microsoft/windows-rs/blob/1f40da8ffc44adba3f9a866287108e80223d0a81/crates/libs/windows/src/Windows/Win32/System/Threading/mod.rs#L840

alacritty_terminal/src/tty/mod.rs
SomeoneToIgnore: Yeah, this way is even better, I've switched over to `sucess: bool` model.

alacritty_terminal/src/event_loop.rs
SomeoneToIgnore: I've used `ChildExit`.

alacritty_terminal/src/tty/windows/child.rs
kchibisov: then it's fine.

alacritty_terminal/src/event.rs
```diff
@@ -69,6 +72,7 @@ impl Debug for Event {
             Event::Wakeup => write!(f, "Wakeup"),
             Event::Bell => write!(f, "Bell"),
             Event::Exit => write!(f, "Exit"),
+            Event::ChildExit(sucess) => write!(f, "ChildExit({sucess})"),
```
kchibisov, line 75:
```suggestion
            Event::ChildExit(success) => write!(f, "ChildExit({success})"),
```

kchibisov (changes requested):

alacritty_terminal/src/event.rs
```diff
@@ -52,6 +52,9 @@ pub enum Event {
 
     /// Shutdown request.
     Exit,
+
+    /// Child process exited, successfully or not.
+    ChildExit(bool),
```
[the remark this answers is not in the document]
SomeoneToIgnore: I do not understand what it should be then: 
* https://doc.rust-lang.org/std/process/struct.Child.html#method.try_wait returns `Result<Option<ExitStatus>>`
* https://github.com/alacritty/alacritty/blob/992011a4cd9a35f197acc0a0bd430d89a0d01013/alacritty_terminal/src/tty/windows/mod.rs#L117 has similar ideas: we might disconnect from waiting before registering the fact that the process was stopped, ergo we cannot really go and call `GetExitCodeProcess` for this case?

I also do not want to use https://doc.rust-lang.org/std/io/type.Result.html since I have to potentially map every 421-like error code to some `Error` variant or go with https://doc.rust-lang.org/std/io/struct.Error.html#method.other which is also not good — technically, child process cancellations are not _errors_, just a different exit code returned.


So, I will get back to the `Option<i32>` model on this case and will be glad to see your alternative proposals.

alacritty_terminal/src/event_loop.rs
```diff
@@ -259,15 +259,17 @@ where
...
+                            {
                                 if self.hold {
                                     // With hold enabled, make sure the PTY is drained.
                                     let _ = self.pty_read(&mut state, &mut buf, pipe.as_mut());
                                 } else {
                                     // Without hold, shutdown the terminal.
                                     self.terminal.lock().exit();
                                 }
-
+                                self.event_proxy.send_event(Event::ChildExit(success));
```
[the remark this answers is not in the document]
SomeoneToIgnore: Thank you, moved upwards.

alacritty_terminal/src/event.rs
```diff
@@ -52,6 +52,9 @@ pub enum Event {
 
     /// Shutdown request.
     Exit,
+
+    /// Child process exited with an error code.
+    ChildExit(i32),
```
chrisduerr, line 57: Just to make sure: You don't care about children that get killed? Since those obviously won't have an exit code.

alacritty_terminal/src/tty/mod.rs
chrisduerr: Just to be clear here: `Option` does **not** come from `try_wait`, in that scenario the process hasn't quit. It is an `Option` because every process on Unix can either exit, or be killed by a signal. Processes killed by a signal have no exit status.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -32,7 +33,11 @@ extern "system" fn child_exit_callback(ctx: *mut c_void, timed_out: BOOLEAN) {
...
+    let status = unsafe { GetExitCodeProcess(child_handle, &mut exit_code) };
+    let exit_code = if status > 0 { Some(exit_code as i32) } else { None };
+    let _ = event_tx.sender.send(ChildEvent::Exited(exit_code));
```
chrisduerr, line 40: Some logical separation of the separate steps with blank lines would be nice here. Otherwise that's just a very dense block that does "everything".

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -145,6 +154,10 @@ mod tests {
         poller.wait(&mut events, Some(WAIT_TIMEOUT)).unwrap();
         assert_eq!(events.iter().next().unwrap().key, PTY_CHILD_EVENT_TOKEN);
         // Verify that at least one `ChildEvent::Exited` was received.
-        assert_eq!(child_exit_watcher.event_rx().try_recv(), Ok(ChildEvent::Exited));
+        assert_eq!(
+            child_exit_watcher.event_rx().try_recv(),
+            Ok(ChildEvent::Exited(Some(1))),
+            "Killed process should report non-zero exit code"
```
chrisduerr, line 160: The message and assertion do not match. Should say we expect an exit code of `1` if that's what we're asserting.

I'd go so far and say this message provides no additional benefit and if we wanted to add a message the most useful thing to put there would be the exit status we actually got.

chrisduerr (changes requested):

alacritty_terminal/src/event.rs
SomeoneToIgnore: I do not think so, but my use cases might be not very mainstream (judging by the fact that I have created 3 feature requests related to them), so ready to adjust the logic the way it seems more appropriate.

All my 3 cases are related to Zed and this part of code: https://github.com/zed-industries/zed/blob/1081ba7a62ed67ba7dae0e2304842e82a589f25e/crates/terminal/src/terminal.rs#L322-L338

* interactive shell session

Does not really matter how and when children exit: `Event::Exit` is used for hiding corresponding UI elements, and it will be sent anyway, no dependence on the child being killed or not.

* (future) collab terminal with tty bytes sent over the server into another instance

Either no children will be used at all, or, if we'll come up with a magic wrapper to make output easier, its child exit should be ignored as not something exposed to the user => no need to know if it's killed too

* running bash-like commands that user defined in Zed elsewhere

Here is what it's all about: those commands spawn a dedicated, task tabs that will only show that command's output — I would want to e.g. change the terminal tab icon based on various child exit statuses, but right now all Alacritty has for shutdown reporting is `Event::Exit`.

If we can rely on Alacritty always sending `Event::ChildExit` on such command finish (which seems to work, based on my tests), killed and all other non-natural abrupts can be derived from the fact that we've already received `Event::Exit`but no `Event::ChildExit` was sent before.

Works for me and does not seem like we need to know about the killing here too.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -32,7 +33,13 @@ extern "system" fn child_exit_callback(ctx: *mut c_void, timed_out: BOOLEAN) {
     }
 
     let event_tx: Box<_> = unsafe { Box::from_raw(ctx as *mut ChildExitSender) };
-    let _ = event_tx.sender.send(ChildEvent::Exited);
+
+    let mut exit_code = 0_u32;
+    let child_handle = event_tx.child_handle.load(Ordering::Relaxed) as HANDLE;
+    let status = unsafe { GetExitCodeProcess(child_handle, &mut exit_code) };
+    let exit_code = if status > 0 { Some(exit_code as i32) } else { None };
```
chrisduerr, line 40: > If the function succeeds, the return value is nonzero.

The return type of `GetExitCodeProcess` is an integer and the documentation states that anything nonzero indicates success. As such it would be possible to get a return value of `-1`, which is smaller than 0 but still a valid success return value.

Generally you want to handle stuff like this using `status != FALSE`, where `FALSE` is the constant imported from windows_sys.

chrisduerr (changes requested):

alacritty_terminal/src/tty/windows/child.rs
SomeoneToIgnore: Indeed, thank you for catching that, fixed.

chrisduerr (approved):

chrisduerr:
@kchibisov Feel free to just merge if you don't have any major issues left with this.

kchibisov (approved):

PR #7796, opened 2024-03-01
Expose more process info

JunkuiZhang:
This PR introduces two functions that are equivalent to their Unix counterparts:
```rust
// unix
let pty = alacritty_terminal::tty::new();
let fd = pty.file().as_raw_fd();
let pid = pty.child().id();
// Windows
let fd = pty.fd();
let pid = pty.pid()
```

### DIFF
--- a/alacritty_terminal/src/tty/windows/child.rs
+++ b/alacritty_terminal/src/tty/windows/child.rs
@@ -1,5 +1,6 @@
 use std::ffi::c_void;
 use std::io::Error;
+use std::num::NonZeroU32;
 use std::sync::atomic::{AtomicPtr, Ordering};
 use std::sync::{mpsc, Arc, Mutex};
 
@@ -43,7 +44,7 @@ pub struct ChildExitWatcher {
     event_rx: mpsc::Receiver<ChildEvent>,
     interest: Arc<Mutex<Option<Interest>>>,
     child_handle: HANDLE,
-    pid: u32,
+    pid: Option<NonZeroU32>,
 }
 
 impl ChildExitWatcher {
@@ -68,7 +69,7 @@ impl ChildExitWatcher {
         if success == 0 {
             Err(Error::last_os_error())
         } else {
-            let pid = unsafe { GetProcessId(child_handle) };
+            let pid = unsafe { NonZeroU32::new(GetProcessId(child_handle)) };
             Ok(ChildExitWatcher {
                 wait_handle: AtomicPtr::from(wait_handle as *mut c_void),
                 event_rx,
@@ -102,7 +103,7 @@ impl ChildExitWatcher {
     }
 
     /// Retrieve the Process ID associated to the underlying child process.
-    pub fn pid(&self) -> u32 {
+    pub fn pid(&self) -> Option<NonZeroU32> {
         self.pid
     }
 }
--- a/alacritty_terminal/src/tty/windows/child.rs
+++ b/alacritty_terminal/src/tty/windows/child.rs
@@ -8,7 +8,7 @@ use polling::{Event, Poller};
 
 use windows_sys::Win32::Foundation::{BOOLEAN, HANDLE};
 use windows_sys::Win32::System::Threading::{
-    RegisterWaitForSingleObject, UnregisterWait, INFINITE, WT_EXECUTEINWAITTHREAD,
+    GetProcessId, RegisterWaitForSingleObject, UnregisterWait, INFINITE, WT_EXECUTEINWAITTHREAD,
     WT_EXECUTEONLYONCE,
 };
 
@@ -42,6 +42,8 @@ pub struct ChildExitWatcher {
     wait_handle: AtomicPtr<c_void>,
     event_rx: mpsc::Receiver<ChildEvent>,
     interest: Arc<Mutex<Option<Interest>>>,
+    child_handle: HANDLE,
+    pid: u32,
 }
 
 impl ChildExitWatcher {
@@ -66,10 +68,13 @@ impl ChildExitWatcher {
         if success == 0 {
             Err(Error::last_os_error())
         } else {
+            let pid = unsafe { GetProcessId(child_handle) };
             Ok(ChildExitWatcher {
                 wait_handle: AtomicPtr::from(wait_handle as *mut c_void),
                 event_rx,
                 interest,
+                child_handle,
+                pid,
             })
         }
     }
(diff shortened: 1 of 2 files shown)


chrisduerr (commented):
Could you go into why you need this?

JunkuiZhang:
> Could you go into why you need this?

If this PR is accepted, it will make it easier to implement related features in[ Zed, a cross-platform editor](https://github.com/zed-industries/zed), as the terminal functionality within the editor is currently not fully developed on Windows. Thank you very much!

### related codes
https://github.com/zed-industries/zed/blob/91d1146d976bcf257e289d9523995b4c28d07c0e/crates/terminal/src/terminal.rs#L384C1-L404C39
```rust
let pty = match tty::new(
    &pty_options,
    TerminalSize::default().into(),
    window.window_id().as_u64(),
) {
    Ok(pty) => pty,
    Err(error) => {
        bail!(TerminalError {
            directory: working_directory,
                shell,
                source: error,
         });
    }
};

#[cfg(unix)]
let (fd, shell_pid) = (pty.file().as_raw_fd(), pty.child().id());

// todo("windows")
#[cfg(windows)]
let (fd, shell_pid) = (-1, 0);
```

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -85,6 +90,14 @@ impl ChildExitWatcher {
     pub fn deregister(&self) {
         *self.interest.lock().unwrap() = None;
     }
+
+    pub fn fd(&self) -> isize {
+        self.child_handle
+    }
```
kchibisov, line 96: This is generally not a `FD`? You should also document what exactly is getting returned here.

alacritty_terminal/src/tty/windows/mod.rs
```diff
@@ -47,6 +47,14 @@ impl Pty {
     ) -> Self {
         Self { backend: backend.into(), conout: conout.into(), conin: conin.into(), child_watcher }
     }
+
+    pub fn fd(&self) -> isize {
+        self.child_watcher.fd()
+    }
```
kchibisov, line 53: You should also document what exactly is getting returned here, because `isize` is clearly confusing. Maybe add a new type for that.

Also, can you generally close the alacritty terminal that way? Though, I think if you close it in between it won't create issues other than telling that pty got closed?

kchibisov (changes requested):
Would suggest in general always include the motivation behind the changes in commit message, so it'll be a bit clear next time.

alacritty_terminal/src/tty/windows/child.rs
JunkuiZhang: Sorry for the inconvenience, I'm still learning, not only coding but also how to collaborate. English is not my native language, so sometimes my ideas may not be expressed very accurately, please understand. Here, the return is the handle of the process. In the Windows environment, a handle is represented as an `isize`. Returning isize really confusing here, I will write as the `HANDLE` type. This handle can mainly be used to obtain the corresponding process ID (`pid`) through the Windows API `GetProcessID`.
I will convert this pr to draft.

alacritty_terminal/src/tty/windows/mod.rs
JunkuiZhang: Since its the process handle, people could do anything to it. Maybe make this function `unsafe`?
As this is just my initial idea, I wanted to get this information directly, so I simply wrote codes to get them. Given that this PR is still very immature, I have marked it as a draft. Any suggestions are welcome

alacritty_terminal/src/tty/windows/child.rs
kchibisov: No, I mean, if you name it as `fn handle` and not `fn fd` it'll be better, and also if you add a `///` explaining you're actually returning, for example with a link to microsoft docs for that.

Also, it should be for the type visible for users, so they can get a grasp of what it is on docsrs.

alacritty_terminal/src/tty/windows/mod.rs
kchibisov: I think keeping it safe is fine given that functions to manipulate with that `HANDLE` will be unsafe in rust, so it's kind of the same as with pointer, because getting it is safe, but touching is unsafe.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -85,6 +90,21 @@ impl ChildExitWatcher {
...
+    /// and the handle is only guaranteed to be valid while the hosted application
+    /// has not yet been destroyed.
+    /// If you terminate the process using this handle, the terminal will get a
+    /// timeout error, and the child watcher will emit an `Exited` event.
+    pub fn raw_handle(&self) -> HANDLE {
+        self.child_handle
+    }
+
+    /// Retrieve the Process ID associated to the underlying child process.
+    pub fn pid(&self) -> u32 {
```
kchibisov, line 105: Maybe we should use `Option<NonZeroU32>` for that? Since the value is 0 when it errors and it generally can't be _zero_. Though, I'm not strong on that.

alacritty_terminal/src/tty/windows/child.rs
```diff
@@ -85,6 +90,21 @@ impl ChildExitWatcher {
     pub fn deregister(&self) {
         *self.interest.lock().unwrap() = None;
     }
+
+    /// Retrieve the process handle of the underlying child process.
+    /// This function does **not** pass ownership of the raw handle to you,
+    /// and the handle is only guaranteed to be valid while the hosted application
+    /// has not yet been destroyed.
+    /// If you terminate the process using this handle, the terminal will get a
+    /// timeout error, and the child watcher will emit an `Exited` event.
```
kchibisov, line 99: Just a bit less dense.

```suggestion
    /// Retrieve the process handle of the underlying child process.
    ///
    /// This function does **not** pass ownership of the raw handle to you, and the
    /// handle is only guaranteed to be valid while the hosted application has not
    /// yet been destroyed.
    ///
    /// If you terminate the process using this handle, the terminal will get
    /// timeout error, and the child watcher will emit an `Exited` event.
```

kchibisov (changes requested):
Just minor nits.

alacritty_terminal/src/tty/windows/child.rs
JunkuiZhang: > Maybe we should use `Option<NonZeroU32>` for that? Since the value is 0 when it errors and it generally can't be _zero_. Though, I'm not strong on that.

Oh, I forgot the error handling here, sorry.

alacritty_terminal/src/tty/windows/child.rs
JunkuiZhang: Done.

kchibisov (approved):

chrisduerr (approved):

kchibisov:
Thanks.

PR #7813, opened 2024-03-06
fix some typos

thirdkeyword:


chrisduerr (approved):
Generally fixing two insignificant typos like this doesn't provide value and only causes maintenance overhead.

But no reason not to merge it I suppose, just don't open PRs like this for repos unless you're actually contributing positively in some fashion.

PR #7816, opened 2024-03-07
Install script for Debian, and installation instructions 

LinuxUser255:
Hello, 

I created an install script for Debian, based on the instructions found on your INSTALL.md:
https://github.com/LinuxUser255/alacritty/blob/master/INSTALL.md

The goal was to make it faster and easier to install Alacritty on Debian, as it is not in their apt package manager.
I also added instructions on running the script and setting up the config file, alacritty.toml on the README.md.  

The install_alacritty.sh is also documented. 

Any questions, etc, let me know.

Thanks,
- Chris

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,10 +40,54 @@ You can find an overview over the features available in Alacritty [here](./docs/
 ## Installation
 
 Alacritty can be installed by using various package managers on Linux, BSD,
-macOS and Windows.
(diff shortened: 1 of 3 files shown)


kchibisov:
Thanks, but not interested. 

>as it is not in their apt package manager.

it's packages by both `Debian` and `Ubuntu`.

PR #4763, opened 2021-02-05
Support for graphics in the terminal

ayosec:
This patch adds support for graphics in the terminal. Graphics can be sent as Sixel images.

## New features

* Ability to render graphics in the grid.
* Support for Sixel images.

## Sixel parser

The Sixel parser is based on the [SIXEL GRAPHICS EXTENSION](https://archive.org/details/bitsavers_decstandar0VideoSystemsReferenceManualDec91_74264381/page/n907/mode/2up) chapter in a DEC manual.

The support is complete except for the pixel aspect ratio parameters. According to the manual, a Sixel image can specify that it expects a specific shape for the pixels in the device, but in none of terminals that I checked these parameters had any effect: they always assume a 1:1 ratio. Also, I didn't find any Sixel generator that emits a different ratio. To avoid extra complexity in the parser, it always assume 1:1 when the image is built.

There are two new terminal modes:

* `SixelScrolling` (`80`)

    If enabled, new graphics are inserted at the cursor position, and they can scroll the grid.

    If disabled, new graphics are inserted at top-left, and they are limited by the height of the window.

* `SixelPrivateColorRegisters` (`1070`)

    If enabled, every Sixel parser has its own color palette.

    If disabled, Sixel images can share a color palette.

    Initially I didn't plan to support this mode, since it seems to be specific to xterm, but when I was testing applications using Sixel I found that mpv uses it to reuse a palette between video frames. Since mpv is based on libsixel, I guess that this feature could be used by more applications.

Both modes are enabled by default.

The function to convert HLS colors to RGB is a direct port of the implementation of the same [function in xterm](https://github.com/ThomasDickey/xterm-snapshots/blob/xterm-364/graphics.c#L1076-L1158). I verified that the function emits the same results in all combinations of values 0, 30, 60, 90, and 100 in every color component. Only a few of these combinations were left in the tests to reduce the noise in the code.

To test the parser there are Sixel images generated with 3 different applications. For each one, there is a `.rgba` file in the same directory with the expected RGBA values. The commands to produce these files are in [`alacritty_terminal/tests/sixel/README.md`](https://github.com/ayosec/alacritty/blob/graphics/alacritty_terminal/tests/sixel/).

###  Byte `90` as DCS

Sixel images using byte `90` as DCS are not supported. DCS can be either `ESC P` or `90`, but the `vte` crate only recognizes `ESC P`. I guess that this is because `90` can be a continuation byte in a UTF-8 sequence (two most significant bits are `10`), so it can be a valid input from users.

[...]

PR #7822, opened 2024-03-09
Fix pty's pixelsize being zero until resize

kchibisov:
117719b3 removed the extra call for TIOCSWINSZ, however the first call was not setting the pixelsize for _unknown_ reason, which regressed some clients.

Fixes: 117719b3 (Remove extra TIOCSWINSZ)

--

It was like that since the start of alacritty, and then it was just carried. The second call to resize the PTY just masked the issue.

### DIFF
--- a/alacritty_terminal/src/tty/unix.rs
+++ b/alacritty_terminal/src/tty/unix.rs
@@ -4,7 +4,6 @@ use std::ffi::CStr;
 use std::fs::File;
 use std::io::{Error, ErrorKind, Read, Result};
 use std::mem::MaybeUninit;
-use std::os::fd::OwnedFd;
 use std::os::unix::io::{AsRawFd, FromRawFd};
 use std::os::unix::net::UnixStream;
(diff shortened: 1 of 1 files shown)


chrisduerr (commented):
> unknown reason

It's not an unknown reason. `ws_xpixel` and `ws_ypixel` are unused according to kernel documentation.

Which application is relying on these?

kchibisov:
>It's not an unknown reason. ws_xpixel and ws_ypixel are unused according to kernel documentation.

https://github.com/alacritty/alacritty/pull/4763#issuecomment-1986964873

We also set it via `on_resize` so we should either never set it or we should always set it. 

Right now we don't set it for create, but we set it later, and it was also always set before, so don't see why we shouldn't stick to behavior we had.

chrisduerr (approved):
I suppose there's no reason not to set it if we have the information, even if it shouldn't be used. It's also just simpler, which is nice.

PR #7824, opened 2024-03-09
Fix hint `Select` action for hyperlink escape

### LINKED ISSUE #7766: hints.enabled action="Select" mouse.enabled=true results in selection of the whole line (instead of just the hint)
Using the following configuration snippet:
```
#Select hint text
[[hints.enabled]]
action          = "Select"
hyperlinks      = true
mouse.enabled   = true
#mouse.mods      = "Alt"
```
The Select action select the whole line containing the hint, not just the hint text. 
For example:
If I do something like `ls -l --hyperlink=auto` only the files get underlined as I hover over them, but when clicked the whole line is selected.

This doesn't change even if I uncoment the `mouse.mods` line.

Is this the intended behavior?

The man 5 alacritty states:
```
Select
                   Select the hint's text.
```
So I would expect only the hint's text to be selected.

If it's not a bug / the current behavior is already established, would you consider something like "SelectHintOnly" action?

### System

OS: Linux (Fedora 38)
Version: alacritty 0.13.1
Linux/BSD: X11, KDE5 / KWin

chrisduerr:
This fixes an issue where the `Select` action for hyperlink escape text
would select the entire line, instead of selecting only the hyperlink
itself.

It also changes the way hyperlinks with the same ID are highlighted,
removing the restriction of being on consecutive lines and instead
highlighting all visible cells that correspond to the matching
hyperlink.

Closes #7766.

---

Had this patch around for a while but never sent it. It definitely changes the behavior a bit but imo the code is both simpler and works better.

### DIFF
--- a/alacritty/src/display/hint.rs
+++ b/alacritty/src/display/hint.rs
@@ -204,7 +204,8 @@ pub struct HintMatch {
 impl HintMatch {
     #[inline]
     pub fn should_highlight(&self, point: Point, pointed_hyperlink: Option<&Hyperlink>) -> bool {
-        self.bounds.contains(&point) && self.hyperlink.as_ref() == pointed_hyperlink
+        self.hyperlink.as_ref() == pointed_hyperlink
+            && (self.hyperlink.is_some() || self.bounds.contains(&point))
     }
 
     #[inline]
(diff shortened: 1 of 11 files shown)


alacritty/src/input/mod.rs
```diff
@@ -1050,12 +1050,9 @@ impl<T: EventListener, A: ActionContext<T>> Processor<T, A> {
         let point = self.ctx.mouse().point(&self.ctx.size_info(), display_offset);
         let hyperlink = self.ctx.terminal().grid()[point].hyperlink();
 
-        // Function to check if mouse is on top of a hint.
-        let hint_highlighted = |hint: &HintMatch| hint.should_highlight(point, hyperlink.as_ref());
-
         if let Some(mouse_state) = self.message_bar_cursor_state() {
             mouse_state
-        } else if self.ctx.display().highlighted_hint.as_ref().map_or(false, hint_highlighted) {
+        } else if hyperlink.is_some() {
```
chrisduerr, line 1055: This change isn't strictly necessary, but it seems to just work so why not?

kchibisov: Hm, the reason this works is because the logic is bit inconsistent. We basically set the cursor from the renderer and from this function, so that's why the code here works even though it doesn't check the regular region anyway.

In the current state of things you can remove all the logic around `hyperlink` and `::Pointer` and it'll still work. 

I don't really like that it's like that though since it's a bit error prone?

alacritty/src/display/hint.rs
```diff
@@ -204,7 +204,8 @@ pub struct HintMatch {
 impl HintMatch {
     #[inline]
     pub fn should_highlight(&self, point: Point, pointed_hyperlink: Option<&Hyperlink>) -> bool {
-        self.bounds.contains(&point) && self.hyperlink.as_ref() == pointed_hyperlink
+        self.hyperlink.as_ref() == pointed_hyperlink
+            && (self.hyperlink.is_some() || self.bounds.contains(&point))
```
kchibisov, line 208: What the reason to check `is_some`, I think it's already covered?

alacritty/src/display/hint.rs
```diff
@@ -400,49 +401,30 @@ pub fn highlighted_at<T>(
...
-
-        let line_contains_hyperlink = grid[next_line]
-            .into_iter()
-            .any(|cell| cell.hyperlink().map_or(false, |h| h == hyperlink));
-
-        if !line_contains_hyperlink {
+    let mut match_start = point;
+    let mut iter = grid.iter_from(point);
+    while let Some(cell) = iter.prev() {
+        if cell.hyperlink().map_or(false, |link| link == hyperlink) {
```
kchibisov, line 423: you can still use `for` here.

kchibisov (changes requested):
CHANGELOG entry.

alacritty/src/display/hint.rs
chrisduerr: If it's a hyperlink escape, we ALWAYS highlight regardless of the bounds. If we don't check for `is_some`, we'd only highlight with in bounds.

alacritty/src/display/hint.rs
chrisduerr: How?

alacritty/src/input/mod.rs
kchibisov: Pretty much yes. I'm not sure what to do about this function though, because clearly the check could be removed, though, maybe there's an edge case...

alacritty/src/display/hint.rs
kchibisov: Hm, you can not without changing the type in alacritty_terminal.

alacritty/src/input/mod.rs
chrisduerr: Yeah honestly I was kinda worried that this patch would only change the mouse cursor when on a hyperlink escape, but that doesn't seem to be the case.

chrisduerr: Actually I found why this is necessary and my changes broke things:

If you're in mouse mode and hover over a URL, this will change mouse shape without shift.

alacritty/src/display/hint.rs
chrisduerr: Yeah it's not a double-ended iterator unfortunately.

kchibisov (approved):

PR #7836, opened 2024-03-15
Bump dependencies

chrisduerr:
This bumps all dependencies that can be updated without introducing a build failure.

---

Might make sense to wait for sctk-adwaita to update crossfont to 0.8.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.14.0-dev
 
-### Packaging
-
-- Minimum Rust version has been bumped to 1.72.0
-
 ### Added
 
 - Default `Home`/`End` bindings in Vi mode mapped to `First`/`Last` respectively
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.8.8"
+version = "0.8.11"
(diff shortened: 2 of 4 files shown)


chrisduerr:
Debian only has 1.70 in SID, while we need at least 1.72. This patch would require 1.74. So it might be best to hold back the clippy update until the next release.

kchibisov:
I don't really want to bump MSRV for patch version, because patch will not be pulled by e.g. debian. The deps we have right now are _fine_.

chrisduerr:
> The deps we have right now are fine.

I think it makes sense to bump deps for non-breaking stuff? Especially for macOS/Windows users.

kchibisov:
The `windows-sys` doesn't really matter, since winit uses the `0.48`, as well as `glutin` so majority of windows code will use `0.48` and our small portion using `0.52` will just increase binary for no reason.

As for bugfixes it's fine, but `ahash` is the one bumping `MSRV` even further I think, so we should just bump everything except it if you want to cherry-pick.

You can split bumps into multiple commits if you want so, but windows stuff I'll bump along with winit 0.30/glutin 0.32 bump.

chrisduerr:
> The windows-sys doesn't really matter, since winit uses the 0.48, as well as glutin so majority of windows code will use 0.48 and our small portion using 0.52 will just increase binary for no reason.

We already are pulling 0.52 anyway, so no reason not to.

> As for bugfixes it's fine, but ahash is the one bumping MSRV even further I think, so we should just bump everything except it if you want to cherry-pick.

ahash seems to build fine with 1.72.0.

kchibisov:
>ahash seems to build fine with 1.72.0.

That's correct, but we'll need 1.70 for cherry pick, and ahash wants 1.72, thus we'd need to downgrade it when picking or make it a separate bump or something.

kchibisov (approved):

PR #7837, opened 2024-03-15
Alacritty version 0.13.2-rc1

chrisduerr:
---

I've cherry-picked everything since 0.13.1 basically, I don't think there were any noteworthy things that should be held back.
Only patches I didn't cherry pick were updates to versions, since that was easier to do in the final RC1 commit.

I'd like to wait for https://github.com/alacritty/alacritty/pull/7836 before making a release.


kchibisov (changes requested):
This commit seems to have message from other commit d5ada921443eeaea24df7694f75a8bfb1e5d978b

MSRV bump should be dropped.

If you're going to include more `alacritty_terminal` stuff, maybe we should look into #7788 as well.

chrisduerr:
> This commit seems to have message from other commit d5ada92

Which one is 'this commit'?

kchibisov:
>Which one is 'this commit'?

The one I've linked, which is `this`. Other commit is next to it in the log.

kchibisov:
>The 1.72.0 one? Why? It's necessary for other patches.

It's not though, it's due to bump of `ahash`, you can safely drop it. I can do that myself though.

kchibisov:
>The one I've linked, which is this. Other commit is next to it in the log.

Hm, it's just the commit has the wrong title originally as well, maybe messed up when merging them, since they are adjacent.

kchibisov (changes requested):
`alacritty_config` is missing a version bump.

chrisduerr:
> alacritty_config is missing a version bump.

Oh I didn't even realize I changed something there. Good catch.

chrisduerr:
@kchibisov force-pushed the update into the RC commit.

kchibisov (approved):

Issue #7774, opened 2024-02-26
window.position.y config doesn't work as expected


dnjmn:
The `window.position.y` doesn't seem to work no matter what i set it to.
No matter what the value of `y` is, the window position remains as shown:

![Screenshot 2024-02-26 at 1 26 49 PM](https://github.com/alacritty/alacritty/assets/49618740/36576e99-9bc9-4742-9aaa-686e724de0b1)

### System

OS:macOS
Version: `alacritty 0.13.1 (fe2a3c5)`

### Logs

Crashes: N/A
Font/Terminal size:
```
~ ❯ alacritty -vv                                                                                                             ✘ INT 26s 13:28:43
Created log file at "/var/folders/gh/lq4p_dbs7231yv74z999rpx80000gn/T/Alacritty-11294.log"
[0.000005333s] [INFO ] [alacritty] Welcome to Alacritty
[0.000222750s] [INFO ] [alacritty] Version 0.13.1 (fe2a3c5)
[0.000719000s] [INFO ] [alacritty] Configuration files loaded from:
                                     "/Users/dhananjaymeena/.config/alacritty/alacritty.toml"
                                     "/Users/dhananjaymeena/.config/alacritty/colorScheme/nightfly.toml"
[0.000970041s] [DEBUG] [alacritty] Using environment locale: C/UTF-8/C/C/C/C
[0.019687416s] [INFO ] [alacritty] Using Apple CGL
[0.029296333s] [DEBUG] [alacritty] Picked GL Config:
                                     buffer_type: Some(Rgb { r_size: 8, g_size: 8, b_size: 8 })
                                     alpha_size: 8
                                     num_samples: 0
                                     hardware_accelerated: true
                                     supports_transparency: Some(true)
                                     config_api: Api(OPENGL)
                                     srgb_capable: true
[0.040800208s] [INFO ] [alacritty] Window scale factor: 1
[0.041644666s] [DEBUG] [alacritty] Loading "Hack Nerd Font Mono" font
[0.067410625s] [INFO ] [alacritty] Running on Apple M1
[0.067447333s] [INFO ] [alacritty] OpenGL version 4.1 Metal - 88, shader_version 4.10
[0.067463041s] [INFO ] [alacritty] Using OpenGL 3.3 renderer
[0.070950833s] [DEBUG] [alacritty] Filling glyph cache with common glyphs
[0.079256333s] [INFO ] [alacritty] Cell size: 10 x 20
[...]

PR #7839, opened 2024-03-17
Fix window positioning issue on macOS

mteit:
PR addresses an issue related to window positioning on macOS. The problem arises when setting the window to visible after resizing, which leads to inconsistent window positioning behavior after resizing. Specifically, the configuration window's position becomes dependent on the window size, causing it to appear at the bottom left corner of the window instead of the expected top left corner.

To fix this issue, I adjusted the order of operations to set the window visible before resizing.

This PR is linked to the corresponding GitHub issue: [Link to GitHub issue](https://github.com/alacritty/alacritty/issues/7774)

### DIFF
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -411,10 +411,12 @@ impl Display {
         let metrics = glyph_cache.font_metrics();
         let (cell_width, cell_height) = compute_cell_size(config, &metrics);
 
-        // Resize the window to account for the user configured size.
         // In macOS, for keeping the window position after resizing
         // at the top left window corner, it must be visible.
+        #[cfg(target_os = "macos")]
         window.set_visible(true);
+
+        // Resize the window to account for the user configured size.
         if let Some(dimensions) = config.window.dimensions() {
             let size = window_size(config, dimensions, cell_width, cell_height, scale_factor);
             window.request_inner_size(size);
@@ -481,6 +483,9 @@ impl Display {
         if config.window.resize_increments {
             window.set_resize_increments(PhysicalSize::new(cell_width, cell_height));
         }
+
+        #[cfg(not(target_os = "macos"))]
+        window.set_visible(true);
 
         #[allow(clippy::single_match)]
         #[cfg(not(windows))]
--- a/alacritty/src/display/mod.rs
+++ b/alacritty/src/display/mod.rs
@@ -16,6 +16,8 @@ use log::{debug, info};
 use parking_lot::MutexGuard;
 use raw_window_handle::RawWindowHandle;
 use serde::{Deserialize, Serialize};
+#[cfg(target_os = "macos")]
+use winit::dpi::PhysicalPosition;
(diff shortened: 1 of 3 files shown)


alacritty/src/display/mod.rs
```diff
@@ -412,6 +412,9 @@ impl Display {
         let (cell_width, cell_height) = compute_cell_size(config, &metrics);
 
         // Resize the window to account for the user configured size.
+        // In macOS, for keeping the window position after resizing
+        // at the top left window corner, it must be visible.
+        window.set_visible(true);
```
chrisduerr, line 417: Did you test this on other platforms? The current location of this call is where it is for a reason iirc.

chrisduerr (changes requested):

alacritty/src/display/mod.rs
mteit: Unfortunately, I wasn't able to test the changes on other platforms. However, I've made adjustments to ensure platform-specific behavior. The call to set the window visible is now conditional based on the target platform, so it should maintain the expected behavior on macOS while not affecting other platforms. Is this approach acceptable to you?

chrisduerr:
>  marked this pull request as draft

Why did you switch the PR to draft status? Did you encounter any issues?

mteit:
> Why did you switch the PR to draft status? Did you encounter any issues?
@chrisduerr I encountered several unexpected behaviors while testing the PR.
Initially, the fix addressed the issue with window position along the `y` axis when `window.decorations = "None"`, ensuring that the position adhered to the configuration in other cases.
However, the commit unexpectedly resolved an issue where `window.dimensions` was restricted in width when `window.decorations = "Full"|"Transparent"|"Buttonless"`.
Despite this fix, I discovered a new issue: when `window.decorations = "Full"|"Transparent"|"Buttonless"`, the window position along the `x` axis starts at 346 pixels (on my display), deviating from the expected behavior. Importantly, this issue persists both on my branch and on the master branch.

So, I'm going to fix the `x` axis issue, and then I will switch the PR status.

alacritty/src/display/mod.rs
```diff
@@ -411,12 +413,23 @@ impl Display {
...
         if let Some(dimensions) = config.window.dimensions() {
             let size = window_size(config, dimensions, cell_width, cell_height, scale_factor);
             window.request_inner_size(size);
         }
 
+        // Set the window position to account for the user configured position.
+        #[cfg(target_os = "macos")]
+        if let Some(position) = config.window.position {
+            window.set_outer_position(PhysicalPosition::new(position.x as u32, position.y as u32));
+        }
```
chrisduerr, line 431: Did you just add this using trial and error? We're already setting the position on startup, why are you setting it again?

This does not seem like the proper solution, it just looks like we're moving it a bunch hoping that one of the operations will stick.

chrisduerr (changes requested):

chrisduerr:
PR is clearly dead. If someone wants to pick it up, please open a new PR.

Issue #7833, opened 2024-03-13
How drawing of cursor 'Block' works ?


iksi4prs:
I want to make some changes in how a cursor is drawn.
It works fine when the shape is not 'Block'.
When it is block, I always see some another "paint" which I can't find where it comes from.
In the code, I see following:
There is comment for  `// Draw cursor.` in [display/mod.rs](https://github.com/alacritty/alacritty/blob/master/alacritty/src/display/mod.rs)
Then in `fn rects`, in [display/cursor.rs](https://github.com/alacritty/alacritty/blob/master/alacritty/src/display/cursor.rs)
the following code;
```
 match self.shape() {
            CursorShape::Beam => beam(x, y, height, thickness, self.color()),
            CursorShape::Underline => underline(x, y, width, height, thickness, self.color()),
            CursorShape::HollowBlock => hollow(x, y, width, height, thickness, self.color()),
            _ => CursorRects::default(),
        }
```
Even if I change the value of the default (`_ =>`) to return something else than default, eg to return `hollow`, but with some distinct colors, I see that the result is not only the rects of `hollow`, but instead it is the result of hollow drawn over some other block rectangle.
Anyone here that can help/explain ?
10x.

kchibisov:
Block cursor is just a change in the cell background, so it's effectively the same as you hide the cursor and change the color for the cell yourself.

kchibisov:
https://github.com/alacritty/alacritty/blob/master/alacritty/src/display/content.rs#L163

kchibisov:
In general if you have dev related questions try ask us on IRC first, though we may be away.

PR #7840, opened 2024-03-17
Added function and explanation-comments, for how drawing of 'block' shape cursor is done

iksi4prs:
The drawing of 'block' shape cursor is different from other cursors,
and not clear at first, because implemented in different way, and in other file.
I've add a function that wraps original code, with explanation.
All logic is kept not changed, also kept default handler in `match self.shape()`, to avoid any changes due to un-mentioned values of `CursorShape`.

See also
[https://github.com/alacritty/alacritty/issues/7833](https://github.com/alacritty/alacritty/issues/7833)

### DIFF
--- a/alacritty/src/display/cursor.rs
+++ b/alacritty/src/display/cursor.rs
@@ -32,6 +32,7 @@ impl IntoRects for RenderableCursor {
             CursorShape::Beam => beam(x, y, height, thickness, self.color()),
             CursorShape::Underline => underline(x, y, width, height, thickness, self.color()),
             CursorShape::HollowBlock => hollow(x, y, width, height, thickness, self.color()),
+            CursorShape::Block => block(),
             _ => CursorRects::default(),
         }
     }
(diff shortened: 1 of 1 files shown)


alacritty/src/display/cursor.rs
```diff
@@ -90,3 +91,9 @@ fn hollow(x: f32, y: f32, width: f32, height: f32, thickness: f32, color: Rgb) -
         index: 0,
     }
 }
+
+/// For CursorShape::Block, drawing is done by changing 'bg' of cell,
+/// not drawing rect (see content.rs), so return default/empty value.
+fn block() -> CursorRects {
+    CursorRects::default()
+}
```
chrisduerr, line 99: How does this help?

chrisduerr (changes requested):

alacritty/src/display/cursor.rs
iksi4prs: @chrisduerr , please check #7833 again, Kirill Chibisov explained it there.

PR #7872, opened 2024-03-27
Support terminal-wg's BiDi draft proposal in alacritty_terminal

MoSal:
https://terminal-wg.pages.freedesktop.org/bidi

Implementation is hidden behind a `bidi_draft` crate feature, and depends on alacritty/vte#112.

Public API is limited to:
 * `Cell::bidi_mode()` method which returns a `BidiMode` enum value.
 * `Cell::bidi_box_mirroring()` method which returns a boolean.
 * `Term` boolean field `bidi_disable_arrow_key_swapping`.

Cell `BidiFlags` is only used internally since relations/interactions between individual flags are not straight forward, and could lead to erroneous behavior if it's all left for API consumers to figure out.

`BidiDir` named fields exist in all `BidiMode` variants. This is done deliberately (instead of e.g. a `BidiMode` struct with a `BidiDir` field) to signify the different purpose `BidiDir` serves in each mode.

### DIFF
--- a/alacritty_terminal/src/term/mod.rs
+++ b/alacritty_terminal/src/term/mod.rs
@@ -25,6 +25,8 @@ use crate::vte::ansi::{
     KeyboardModesApplyBehavior, NamedColor, NamedMode, NamedPrivateMode, PrivateMode, Rgb,
     StandardCharset,
 };
+#[cfg(feature = "bidi_draft")]
+use crate::vte::ansi::{ScpCharPath, ScpUpdateMode};
 
 pub mod cell;
 pub mod color;
@@ -274,6 +276,16 @@ pub struct Term<T> {
 
     pub selection: Option<Selection>,
 
+    #[cfg(feature = "bidi_draft")]
+    /// BiDi-supporting terminal emulators should swap the right and left arrow keys by
+    /// default when in RTL paragraph context. This is done to actually match the physical
+    /// movement of the cursor in RTL text in such contexts.
+    ///
+    /// This is a global (private) mode of the terminal. The escape sequence `\e[?1243l`
+    /// disables swapping and sets this field to `true`. The escape sequence `\e[?1243h`
+    /// restores the default, setting this field to `false` again.
+    pub bidi_disable_arrow_key_swapping: bool,
+
     /// Currently active grid.
     ///
     /// Tracks the screen buffer currently in use. While the alternate screen buffer is active,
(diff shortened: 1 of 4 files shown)


chrisduerr:
Seems like there's no performance impact, it's possible that the u8 fits inside the existing cell's padding though I'm too lazy to do the math. That said I still think it would be better stored in the extra data if this ever goes anywhere.

MoSal:
> Seems like there's no performance impact, it's possible that the u8 fits inside the existing cell's padding though I'm too lazy to do the math. That said I still think it would be better stored in the extra data if this ever goes anywhere.

Yeah, I benched `no_bidi`, `bidi_cell`, and `bidi_extra`, and the difference was in the noise.

Does perfbot bench with `--all-features` though? If not, then results above benched the SCP change alone!

----

I will add some tests next. Is anything else required of me?

chrisduerr:
> Does perfbot bench with --all-features though? If not, then results above only benched the SCP change!

It does not, good point. I've temporarily enabled it as a default feature for testing.

chrisduerr:
@perfbot

alacritty_terminal/src/term/mod.rs
```diff
@@ -2082,6 +2193,15 @@ impl<T: EventListener> Handler for Term<T> {
         let mode = match mode {
             ansi::Mode::Named(mode) => mode,
             ansi::Mode::Unknown(mode) => {
+                #[cfg(feature = "bidi_draft")]
+                // BDSM
```
chrisduerr, line 2197: :eyes:

MoSal: BI-DIRECTIONAL SUPPORT MODE

chrisduerr:
@MoSal Are you still interested in this? I'm definitely happy to collaborate if that avoids fragmentation between `alacritty_terminal` consumers.

vaknin:
Hey @chrisduerr,

I want to reopen this BiDi work. Quick question about the feature flag before I submit a new PR.
Since bidi_draft is a compile-time feature, users installing from package managers won't have bidi-text support unless packagers enable it.

Should it:
1. Stay compile-only
2. be enabled by default
3. be a runtime config option, e.g. via alacritty.toml

chrisduerr:
> Since bidi_draft is a compile-time feature, users installing from package managers won't have bidi-text support unless packagers enable it.

The way this was designed, bidi is only added to alacritty_terminal, not to Alacritty. So that's why it's a feature, that way Alacritty doesn't need to compile code it doesn't use.

Issue #7868, opened 2024-03-27
On Windows "Open Alacritty Here" not work well


ZhiZe-ZG:
The “Open Alacritty Here” option in the context menu will open Alacritty, but the shell path will still be at home instead of the current path.

The value of `HKEY_CLASSES_ROOT\Directory\Background\shell\Open Alacritty here\command` is just `C:\Program Files\Alacritty\alacritty.exe`.

I installed Alacritty by winget.

### System

OS: Windows 11 Pro (23H2)
Version: alacritty 0.13.2 (bb8ea18)

### Logs

Crashes: not crashed
Font/Terminal size: 
```
[0.101764800s] [INFO ] [alacritty] Using WGL
[0.107716100s] [DEBUG] [alacritty] Picked GL Config:
                                     buffer_type: Some(Rgb { r_size: 8, g_size: 8, b_size: 8 })
                                     alpha_size: 8
                                     num_samples: 0
                                     hardware_accelerated: true
                                     supports_transparency: Some(false)
                                     config_api: Api(OPENGL | GLES1 | GLES2)
                                     srgb_capable: true
[0.109633600s] [DEBUG] [alacritty] Loading "IosevkaZhiZeX00 Nerd Font" font
[0.112669600s] [INFO ] [alacritty] Running on NVIDIA GeForce GTX 1060 6GB/PCIe/SSE2
[0.112760400s] [INFO ] [alacritty] OpenGL version 3.3.0 NVIDIA 536.23, shader_version 3.30 NVIDIA via Cg compiler
[0.112907800s] [INFO ] [alacritty] Using OpenGL 3.3 renderer
[0.116413500s] [DEBUG] [alacritty] Enabled debug logging for OpenGL
[0.116505800s] [DEBUG] [alacritty] Filling glyph cache with common glyphs
[0.139609100s] [INFO ] [alacritty] Cell size: 10 x 26
[0.139723600s] [INFO ] [alacritty] Padding: 0 x 0
[0.139811800s] [INFO ] [alacritty] Width: 800, Height: 624
[0.158401800s] [INFO ] [alacritty] PTY dimensions: 24 x 80
[0.159064900s] [INFO ] [alacritty_terminal] Using Windows API for pseudoconsole
[0.162429500s] [INFO ] [alacritty] Initialisation complete
[...]

PR #7875, opened 2024-03-28
Update wxs to fix Open Alacritty Here in Windows

ZhiZe-ZG:
The previous msi installer generated incorrect registry values. This causes the working path of alacritty started by "Open Alacritty Here" in the right-click menu is not in the current path.

This patch corrects the registry generated values.

The above modifications have been tested on Windows 11 Pro (23H2).

### DIFF
--- a/alacritty/windows/wix/alacritty.wxs
+++ b/alacritty/windows/wix/alacritty.wxs
@@ -41,7 +41,7 @@
       <!-- Add context menu -->
       <Component Id="ContextMenu" Guid="449f9121-f7b9-41fe-82da-52349ea8ff91" Directory="TARGETDIR">
          <RegistryKey Root="HKCU" Key="Software\Classes\Directory\Background\shell\Open Alacritty here\command">
-            <RegistryValue Type="string" Value="[AlacrittyProgramFiles]alacritty.exe &quot;--working-directory&quot; &quot;%v.&quot;" KeyPath="yes" />
+            <RegistryValue Type="string" Value='[AlacrittyProgramFiles]alacritty.exe --working-directory "%v"' KeyPath="yes" />
          </RegistryKey>
          <RegistryKey Root="HKCU" Key="Software\Classes\Directory\Background\shell\Open Alacritty here">
             <RegistryValue Type="string" Name="Icon" Value="[AlacrittyProgramFiles]alacritty.exe" />
--- a/alacritty/windows/wix/alacritty.wxs
+++ b/alacritty/windows/wix/alacritty.wxs
@@ -41,7 +41,7 @@
       <!-- Add context menu -->
       <Component Id="ContextMenu" Guid="449f9121-f7b9-41fe-82da-52349ea8ff91" Directory="TARGETDIR">
          <RegistryKey Root="HKCU" Key="Software\Classes\Directory\Background\shell\Open Alacritty here\command">
-            <RegistryValue Type="string" Value="[AlacrittyProgramFiles]alacritty.exe" KeyPath="yes" />
+            <RegistryValue Type="string" Value="[AlacrittyProgramFiles]alacritty.exe &quot;--working-directory&quot; &quot;%v.&quot;" KeyPath="yes" />
          </RegistryKey>
          <RegistryKey Root="HKCU" Key="Software\Classes\Directory\Background\shell\Open Alacritty here">
             <RegistryValue Type="string" Name="Icon" Value="[AlacrittyProgramFiles]alacritty.exe" />


ZhiZe-ZG:
Related issue: #7868

ZhiZe-ZG:
@chrisduerr

alacritty/windows/wix/alacritty.wxs
```diff
@@ -41,7 +41,7 @@
       <!-- Add context menu -->
       <Component Id="ContextMenu" Guid="449f9121-f7b9-41fe-82da-52349ea8ff91" Directory="TARGETDIR">
          <RegistryKey Root="HKCU" Key="Software\Classes\Directory\Background\shell\Open Alacritty here\command">
-            <RegistryValue Type="string" Value="[AlacrittyProgramFiles]alacritty.exe" KeyPath="yes" />
+            <RegistryValue Type="string" Value="[AlacrittyProgramFiles]alacritty.exe &quot;--working-directory&quot; &quot;%v.&quot;" KeyPath="yes" />
```
chrisduerr, line 44: Is it necessary to quote thhe `--working-directory` flag here? And does `\"` work or does it need to be `&quot;`?

chrisduerr (changes requested):

alacritty/windows/wix/alacritty.wxs
ZhiZe-ZG: If omit `--working-directory`, for example:

```shell
alacritty C:/User/
```

will lead:

```
error: unrecognized subcommand 'C:/User/'

Usage: alacritty.exe [OPTIONS] [COMMAND]
```

And `\"` isn't standard escape character in XML, use `\"` will terminating string after `\`. That caused a error when wix processing in my test.

ZhiZe-ZG: Quoting  `--working-directory` and `%v`, follow the writing method of git shell and cmd. Without quotation, perhaps also work. But add quotes to prevent some unexpected problems

ZhiZe-ZG: As we all know, Windows handles command line parameters differently from Unix. So put it in a string to prevent parsing problems. VS Code even quotes the executable path itself.

ZhiZe-ZG: So now there are so many styles of writing:
1.
```
"[AlacrittyProgramFiles]alacritty.exe --working-directory %v"
```
2.
```
"[AlacrittyProgramFiles]alacritty.exe --working-directory &quot;%v&quot;"
```
3.
```
"[AlacrittyProgramFiles]alacritty.exe &quot;--working-directory&quot; &quot;%v&quot;"
```
4.
```
&quot;"[AlacrittyProgramFiles]alacritty.exe&quot; &quot;--working-directory&quot; &quot;%v&quot;"
```

which one do you choose?

I still recommend using quotation marks, especially 4.

This can avoid many problems caused by encoding or path spaces.

chrisduerr: Quoting the path is fine, as long as it is required and it doesn't work with a space in the path otherwise (that should be simple to test).

So if quotes around the path is required, then 2 would be the best option. There's no reason to quote something if it works without it. That just makes things unreadable.

ZhiZe-ZG: In my test, using single quotes not work:
1.
```
<RegistryValue Type="string" Value="'[AlacrittyProgramFiles]alacritty.exe' '--working-directory' '%v'" KeyPath="yes" />
```
But this work:
2.
```
<RegistryValue Type="string" Value='"[AlacrittyProgramFiles]alacritty.exe" "--working-directory" "%v"' KeyPath="yes" />
```
Even if the installation path set by msi has spaces (and working directory also has spaces), the following writing works:
3.
```
<RegistryValue Type="string" Value='[AlacrittyProgramFiles]alacritty.exe --working-directory %v' KeyPath="yes" />
```
This should be due to some patches in new versions of Windows. This is not guaranteed to work on older versions of Windows. 

Therefore, software such as VS Code will use quotation marks. This can be seen as a defensive programming technique.

In order to be consistent with the overall style of the project, I would like to ask you to choose a version between 2 and 3.

ZhiZe-ZG:
@chrisduerr 

I conducted a new test and found that when there are Han characters and spaces in the path, the quotation marks before and after `%v` cannot be omitted.

I haven't found that omitting other quotes causes problems at the moment.

chrisduerr (approved):
LGTM, thanks for testing this.

Issue #7626, opened 2024-01-15
macOS: Creating a new tab when terminal is in SimpleFullScreen mode opens tab in a new window


sl4m:
I haven't seen anyone else report this. When alacritty is not in SimpleFullScreen mode, new tab creation works as expected. Is this a known issue?

### System

OS: macOS Sonoma latest
Version: alacritty 0.13.1 (fe2a3c5)

### Logs

Crashes: N/A
Font/Terminal size:

```
[0.000003500s] [INFO ] [alacritty] Welcome to Alacritty
[0.000789041s] [INFO ] [alacritty] Version 0.13.1 (fe2a3c5)
[0.001582041s] [INFO ] [alacritty] Configuration files loaded from:
                                     "/Users/skim/.alacritty.toml"
                                     "/Users/skim/.config/colours/tomorrow-night-256.toml"
[0.002712458s] [DEBUG] [alacritty] Using environment locale: en_US.UTF-8
[0.028418541s] [INFO ] [alacritty] Using Apple CGL
[0.043234666s] [DEBUG] [alacritty] Picked GL Config:
                                     buffer_type: Some(Rgb { r_size: 8, g_size: 8, b_size: 8 })
                                     alpha_size: 8
                                     num_samples: 0
                                     hardware_accelerated: true
                                     supports_transparency: Some(true)
                                     config_api: Api(OPENGL)
                                     srgb_capable: true
[0.063517291s] [INFO ] [alacritty] Window scale factor: 2
[0.064318666s] [DEBUG] [alacritty] Loading "Mensch" font
[0.083453875s] [INFO ] [alacritty] Running on Apple M2 Ultra
[0.083484625s] [INFO ] [alacritty] OpenGL version 4.1 Metal - 88, shader_version 4.10
[0.083495333s] [INFO ] [alacritty] Using OpenGL 3.3 renderer
[0.087088625s] [DEBUG] [alacritty] Filling glyph cache with common glyphs
[0.088921458s] [INFO ] [alacritty] Cell size: 18 x 36
[0.088941375s] [INFO ] [alacritty] Padding: 2 x 2
[0.088950250s] [INFO ] [alacritty] Width: 1600, Height: 1200
[0.175626250s] [INFO ] [alacritty] PTY dimensions: 33 x 88
[0.178481875s] [INFO ] [alacritty] Initialisation complete
[0.209736083s] [DEBUG] [alacritty_terminal] New num_cols is 284 and num_lines is 79
[...]

PR #7885, opened 2024-03-30
Fix macOS creating tab in startup_mode = Fullscreen

Uziniii:
When startup_mode was set to Fullscreen, creating a new tab would cause the tab to be set as Fullscreen too
Tab being Fullscreen cause a new window to be created instead of doing a tab


chrisduerr (changes requested):
What is the motivation behind this PR? This sounds extremely opinionated and I do not see what startup mode has to do with anything here, considering fullscreen can be entered at any point during runtime.

Uziniii:
That linked to this issue #7626 
sorry i forgot to link it

chrisduerr:
You're not fixing the issue though.

Issue #7906, opened 2024-04-15
Move `GENERAL` options into `[general]` section


kchibisov:
The fact that this section should be at the top is confusing for some users, thus make a dedicated section for it and write an automatic migration rules to move things into it.

Certain fields, like `shell` could be moved into other sections, like `[terminal]`, however things like `live_config_reload` can not, and are enabled for the whole instance.

Maybe `[instance]` section could be used instead, but it feels not general enough.

PR #7908, opened 2024-04-16
Add: move GENERAL config to [general], live_config_reload to [instanc…

Uziniii:
…e], shell to [terminal]

#7906


nixpulvis:
What other than the update to the man pages and CHANGELOG is needed here?

chrisduerr:
> What other than the update to the man pages and CHANGELOG is needed here?

The code is clearly in a draft state and I have no interest in taking a serious look without that changing. At that point I might as well just toss it and do it myself.

chrisduerr:
Closing since at this point this patch is more likely to stall this feature than actually help ship it.

PR #7917, opened 2024-04-20
Fix config TOCTOU

mattfbacon:
There is a TOCTOU because we explicitly check if the file exists before reading it. We should just check for a "not found" error from the read, instead.

### DIFF
--- a/alacritty/src/config/mod.rs
+++ b/alacritty/src/config/mod.rs
@@ -263,13 +263,12 @@ fn load_imports(config: &Value, config_paths: &mut Vec<PathBuf>, recursion_limit
             },
         };
 
-        if !path.exists() {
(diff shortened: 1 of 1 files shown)


chrisduerr:
Have you actually been able to reproduce this in practice?

mattfbacon:
WDYM, is the problem not self-apparent?

chrisduerr:
> WDYM, is the problem not self-apparent?

I'm curious about the motivation behind this pull request. If it's just a theoretical issue that is fine, however I'd like to be aware if this actually caused real-world issues.

mattfbacon:
It is primarily a theoretical issue, but it could have a real effect, because presumably the "not found" case was special-cased for a reason, to print a better error message, and that code path could be missed if the TOCTOU mismatch occurs.

chrisduerr (approved):

PR #7925, opened 2024-04-21
Fix #7924

### LINKED ISSUE #7924: When `startup_mode = "Fullscreen"` both Cmd+T and Cmd+N start opening tabs in a new window
For me it is `startup_mode = "Fullscreen"` that causes this bug. Both Cmd+T and Cmd+N start opening tabs in a new window with this line in config.

Sonoma 14.4, Alacritty 0.13.1

https://github.com/alacritty/alacritty/assets/87249792/0aa38ebe-486e-4dc7-8ccf-38e28d396307

_Originally posted by @Konnko in https://github.com/alacritty/alacritty/issues/7626#issuecomment-1987005907_

Uziniii:
#7924

### DIFF
--- a/alacritty/src/window_context.rs
+++ b/alacritty/src/window_context.rs
@@ -31,6 +31,7 @@ use alacritty_terminal::tty;
 
 use crate::cli::{ParsedOptions, WindowOptions};
 use crate::clipboard::Clipboard;
+use crate::config::window::StartupMode;
 use crate::config::UiConfig;
 use crate::display::window::Window;
 use crate::display::Display;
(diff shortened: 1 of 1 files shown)


chrisduerr (changes requested):
This is not an acceptable solution. In fact it's not a solution at all.

alacritty/src/display/window.rs
```diff
@@ -163,14 +163,19 @@ impl Window {
             window_builder = window_builder.with_embed_parent_window(parent_window_id);
         }
 
+        #[cfg(target_os = "macos")]
+        let fullscreen = if tabbing_id.is_some() { None } else { config.window.fullscreen() };
```
chrisduerr, line 167: You're just disabling fullscreen when tabbing is enabled, but that can be easily done by the user themselves if that's actually what they want.

I don't know how macOS works here exactly, but I would want the fullscreen window to get the tabs. I'm not actually sure what happens when a window with multiple tabs is made fullscreen, but I'd assume it doesn't just get split into multiple windows?

Also if we wanted to prohibit having these two things set together, it should be a configuration warning/error at least. We shouldn't just make these changes and not report back to the user that their selected configuration is invalid.

chrisduerr (changes requested):

alacritty/src/display/window.rs
Uziniii: If you look at the video in the issue you will see that when he create a tab that make another window into fullscreen mode so we should disable tab when start mode fullscreen is set or make it a config

Also when making alacritty fullscreen without startup mode tabs work perfectly fine

And when startup mode is set exiting alacritty leave buggy window in fullscreen while alacritty is exited

PR #7927, opened 2024-04-22
Fix dynamic title override for multiple windows

### LINKED ISSUE #6836: Setting title doesn't work in terminals launched via 'msg create-window' when instance was created with `--title`
Hello,

When I launch a terminal with "alacritty" that terminal's title is updated with the running command / pwd as I do stuff in it. When I launch` a terminal with "alacritty msg create-window" the title isn't being updated, it just stays as "Alacritty"

the zsh snippet responsible for updating the terminals title is:
```
case $TERM in
    xterm*|alacritty)
        preexec () {
	    print -Pn "\e]0;$1\a"
	}
	precmd () {
	    print -Pn "\e]0;$PWD\a"
	}
        ;;
esac
```

### System

OS: Fedora Linux 39 (Rawhide Prerelease)
Version: alacritty 0.10.1
Linux/BSD: X11 i3wm

chrisduerr:
This fixes an issue where Windows spawned after the initial one through
IPC or bindings would not update their title due to the initial window
having its title set through the CLI.

Title changes are still inhibited for additional windows when they are
spawned through `alacritty msg create-window` with the `--title` CLI
option added.

Closes #6836.

### DIFF
--- a/alacritty/src/cli.rs
+++ b/alacritty/src/cli.rs
@@ -92,7 +92,6 @@ impl Options {
             config.ipc_socket |= self.socket.is_some();
         }
 
-        config.window.dynamic_title &= self.window_options.window_identity.title.is_none();
         config.window.embed = self.embed.as_ref().and_then(|embed| parse_hex_or_decimal(embed));
         config.debug.print_events |= self.print_events;
         config.debug.log_level = max(config.debug.log_level, self.log_level());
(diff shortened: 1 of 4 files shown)


alacritty/src/cli.rs
```diff
@@ -92,7 +92,6 @@ impl Options {
             config.ipc_socket |= self.socket.is_some();
         }
 
-        config.window.dynamic_title &= self.window_options.window_identity.title.is_none();
```
chrisduerr, line 95: I'm not 1000% sure removing this is safe, but it seems like the `preserve_title` option on `WindowContext` should do the same thing this operation did in the past. So I believe this should be redundant.

The advantage of not doing it here is that if we do it on the config directly, we have no way to distinguish between having it set in the config and having it set on the initial CLI. While storing the two separately allows us to make that distinction.

kchibisov (approved):

Issue #4474, opened 2020-11-24
(macOS) Mission Control's animations are slow when switching from Alacritty.


ftpd:
### System

OS: macOS Big Sur 11.0.1
Version: alacritty 0.5.0 (a6681e3)

### The bug:
There is this functionality of macOS called 'Mission Control', which user can invoke (via shortcut, key on Apple keyboard or trackpad gesture) to show all windows from current workspace to quickly switch between them.
When I'm using it to switch to anything from Alacritty (like: Alacritty is the focused app/window while invoking Mission Control), everything is laggy and slow. When I'm invoking it 'from' any different app, it's smooth and fast.

Here is the screencast I recorded to show the issue: https://insomniac.pl/f/alacritty-slow-mission-control.mov (it's Apple QuickTime, I can convert to other format if you need it).

chrisduerr:
This is a duplicate of https://github.com/alacritty/alacritty/issues/4410.

As stated there, I don't see how Alacritty could do anything about this? Mission control is completely outside the control of Alacritty and I'm not aware of any "laggy mission control" setting.

How many other OpenGL applications have you tested? What makes you think that this is a bug **inside of Alacritty**, other than the issue showing up with Alacritty?

jonleopard:
I don't get the same behavior on my machine. Whenever I switch context with Mission Control all behaves smoothly _most_ of the time. Occasionally, I have noticed some UI lag across multiple applications (specifically when dragging windows from one screen to the other) with Big Sur. I also have two monitors plugged into my machine which causes more rendering work for my already lackluster GPU (I'm using a macbook pro). 

If any lagging is occurring it's likely and issue with macOS Big Sur or your hardware.

ftpd:
@chrisduerr thanks for posting out this second issue. I searched before creating this one, but I had to miss it somehow. I don't know any other software with OpenGL (or rather: I'm not aware which apps are using OpenGL and which are not - as an end-user and a person not so much interested what graphical engine (?) is used by the app, I think I can not to care about such details). Sorry for assuming it's Alacritty's fault - the problem occurred after starting using it, so my natural reaction was blaming the software itself. Can you list like two of other OpenGL popular (so: trusted) and free app, so I can install and test?

@jonleopard which version of macOS are you using? I won't blame the hardware, as it always worked smoothly for everything. Just yesterday it stopped, right after I started using Alacritty, but like @chrisduerr said, it might be global OpenGL problem.

chrisduerr:
[...]

PR #7928, opened 2024-04-23
MacOS handles window opacity better

sfsam:
On MacOS, when the window is configured to be opaque, draw an opaque title bar and proper window border.
This is achieved by giving the window the standard background color when it is opaque.

The image below shows the current translucent drawing for an opaque window versus this pull request's opaque drawing.
alacritty

### DIFF
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -23,6 +23,7 @@ use {
     cocoa::appkit::NSColorSpace,
     cocoa::base::{id, nil, NO, YES},
     objc::{msg_send, sel, sel_impl},
+    objc::runtime::{Class, Object},
     winit::platform::macos::{OptionAsAlt, WindowBuilderExtMacOS, WindowExtMacOS},
 };
 
@@ -167,7 +168,7 @@ impl Window {
             .with_title(&identity.title)
             .with_theme(config.window.theme())
             .with_visible(false)
-            .with_transparent(true)
+            .with_transparent(config.window_opacity() < 1.)
             .with_blur(config.window.blur)
             .with_maximized(config.window.maximized())
             .with_fullscreen(config.window.fullscreen())
@@ -343,6 +344,9 @@ impl Window {
 
     pub fn set_transparent(&self, transparent: bool) {
         self.window.set_transparent(transparent);
+
+        #[cfg(target_os = "macos")]
+        set_transparent_macos(&self.window, transparent);
     }
 
     pub fn set_blur(&self, blur: bool) {
@@ -494,3 +498,25 @@ fn use_srgb_color_space(window: &WinitWindow) {
         let _: () = msg_send![raw_window, setColorSpace: NSColorSpace::sRGBColorSpace(nil)];
     }
 }
+
+#[cfg(target_os = "macos")]
+fn set_transparent_macos(window: &WinitWindow, transparent: bool) {
+    let raw_window = match window.raw_window_handle() {
+        RawWindowHandle::AppKit(handle) => handle.ns_window as id,
+        _ => return,
+    };
+
+    // If transparent, fill background with NSColor.clearColor,
+    // otherwise fill with NSColor.windowBackgroundColor so that MacOS
+    // properly draws the window's border and opaque title bar.
+    unsafe {
+        let color_class = Class::get("NSColor").unwrap();
+        let bg_color: *const Object = if transparent {
+            msg_send![color_class, clearColor]
+        } else {
+            msg_send![color_class, windowBackgroundColor]
+        };
+        let _: () = msg_send![raw_window, setBackgroundColor: bg_color];
+    }
+}
+


chrisduerr (changes requested):
If you're looking for this functionality, it should be added to Winit.

kchibisov:
IIRC we rejected such thing upstream because it was too specific and you can do it very easy downstream. it's also a bit opinionated since you can change the color pretty much arbitrary, so you need to define coloring APIs, etc.

Maybe in the future it'll be a bit easier to manipulate things like that because the macOS stuff will be exposed directly, but for now I don't mind such change. Though, I'm not sure whether it's how the rest of the apps behave by default, since macOS tends to blend background color into decorations and what not.

sfsam:
FWIW, I didn't intend for this PR to be opinionated. It is applying MacOS's `windowBackgroundColor` only if the window is opaque. This makes MacOS draw the window with proper borders and opaque title bar. You can still set Alacritty's `opacity` config option to get transparency. In fact, if you set `opacity` to `0.9999`, this PR will draw much like current Alacritty. That said, maybe this functionality belongs in winit.

MacOS `windowBackgroundColor` isn't a specific color. It is computed by the OS depending on a number of factors (dark mode, accessibility, etc).
https://developer.apple.com/documentation/appkit/nscolor/1528630-windowbackgroundcolor?language=objc

Here are some apps on a black background with Dark Mode on to illustrate how the windows are drawn:

pr

kchibisov:
>Is it not a standard thing to have transparent for transparent windows? I don't know how it actually works. If there's no such standard then I'd prefer just always keeping Alacritty's decorations opaque.

You can set arbitrary color for decorations. If winit will do so it could conflict with users doing so as well, since you can set a color with alpha iirc. Maybe winit should do that by default, but it's hard to understand what the default is...

We could add it here and then try to upstream in a sane way, but I know that e.g. neovide uses this in a different manner with a custom color, etc...

kchibisov:
>Seems like a great addition to winit without all that much macOS bullshit. Why would other applications like Alacritty not want this is the way I’d think about it.

The point is that other applications sometimes change this color on their own, so winit can't do that implicitly by default, it must be explicit, and making it explicit is a bit cumbersome because defining colors, etc sounds a bit out of place for winit.

We could re-iterate it upstream, ofc, but I'm not sure in the current state things will be added upstream.

madsmtm:
I think the missing border is due to a call to `set_has_shadow`, which is done by Alacritty when the window's opacity is turned down.

Regarding transparency, maybe the problem is rather that Winit sets the background color to `NSColor::clearColor()` when initializing with transparency? If we didn't do that, then the window title is opaque (like on most other applications, and which generally seems to be the default) regardless of whether the window _contents_ is opaque.

And then users can opt-in to a transparent window title bar using some other method, either the manual approach here, or using some API we provide in Winit?

kchibisov:
we call `set_has_shadow(true)` once you change the opacity to `1.0`, so it should be basically as default? Or does the order matter here.

sfsam:
If you create a new Mac app in Xcode, it will automatically create an `NSWindow` for you with the `backgroundColor` property set to `NSColor.windowBackgroundColor`. This is the default behavior. This window draws the way this PR does. If you later explicitly change the window's `backgroundColor` to `NSColor.clearColor`, the window draws like current Alacritty.

madsmtm:
I've opened https://github.com/rust-windowing/winit/pull/3657 to correct the behaviour in Winit - after that, the default background color is always `+[NSColor windowBackgroundColor]`.

The rejected PR for setting the background color is https://github.com/rust-windowing/winit/pull/2314, btw.

kchibisov:
Is this patch still needed after this https://github.com/alacritty/alacritty/pull/7935 ?

sfsam:
After https://github.com/alacritty/alacritty/pull/7935, Alacritty still doesn't draw window borders. What's more, it also doesn't respect the config option`[window] decorations_theme_variant = "Dark"`. Instead, it draws a white title bar.

I tried applying this PR to https://github.com/alacritty/alacritty/pull/7935, but it has a conflict: `WindowBuilderExtMacOS` was changed to `WindowAttributesExtMacOS`. If I make that change and apply the rest of this PR, it now has no effect. The title bar is still white and there are no borders.

nixpulvis:
IDK, it seems to still be drawing a dark title bar, but no border as desired. I didn't follow the PRs in winit, so I'm not sure what I expect at the moment, but I think agree with the intentions of this PR at least.

nixpulvis:
@kchibisov seems like this PR is still functional. In fact it seems to improve the performance of the app switcher (Expose) since the window is now longer partially transparent. This was an outstanding issue for a while and may actually be a partial fix for #4474.

nixpulvis:
Here's a video showing the performance difference of Mission Control.

`master`: 


`mac_transparency` (this branch):

nixpulvis:
Also, here's the old/new behavior when dragging Alacritty windows between spaces.

`master`:  

`mac_transparency` (this branch):

nixpulvis:
@kchibisov leave me a day to recollect myself, I may have tested this wrong. Now when I compile `master` it seems the title bar is not transparent as intended. It's entirely possible I did something wrong this morning while I was testing a few different branches.

kchibisov:
In general, I'll re-look into this myself. Though, I'm limited to macOS 12.6 iirc.

LeileiChui:
> @kchibisov leave me a day to recollect myself, I may have tested this wrong. Now when I compile `master` it seems the title bar is not transparent as intended. It's entirely possible I did something wrong this morning while I was testing a few different branches.给我一天时间回忆一下，我可能测试错了。现在，当我编译 `master` 时，标题栏似乎不按预期透明。今天早上我在测试几个不同的分支时完全有可能做错了什么。

After
```
commit 48c088a50c977dfaf2a5a7052a5ddd39071e6063
Author: Kirill Chibisov <contact@kchibisov.com>
Date:   2024-05-04 20:51:56

    Bump winit to 0.30.0
```
alacritty's transparency seems to be broken.

gersonsosa:
> alacritty's transparency seems to be broken.

I can confirm this, I noticed a few days ago, I don't see any issue in winit, not sure how to know whether this is an alacritty or winit issue.

madsmtm:
I've posted https://github.com/alacritty/alacritty/pull/7965 which extends upon this one with a few further fixes.

chrisduerr:
Closing in favor of #7965.

PR #7935, opened 2024-04-27
Bump winit to 0.30.0

kchibisov:
--

The first commit does breaking changes, the other moves from `run` to `run_app` API.

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -18,6 +18,7 @@ use crossfont::Size as FontSize;
 use glutin::display::{Display as GlutinDisplay, GetGlDisplay};
 use log::{debug, error, info, warn};
 use raw_window_handle::HasRawDisplayHandle;
+use winit::application::ApplicationHandler;
 use winit::event::{
     ElementState, Event as WinitEvent, Ime, Modifiers, MouseButton, StartCause,
     Touch as TouchEvent, WindowEvent,
@@ -47,7 +48,7 @@ use crate::display::hint::HintMatch;
 use crate::display::window::Window;
 use crate::display::{Display, Preedit, SizeInfo};
 use crate::input::{self, ActionContext as _, FONT_SIZE_STEP};
-use crate::logging::LOG_TARGET_CONFIG;
+use crate::logging::{LOG_TARGET_CONFIG, LOG_TARGET_WINIT};
 use crate::message_bar::{Message, MessageBuffer};
 use crate::scheduler::{Scheduler, TimerId, Topic};
 use crate::window_context::WindowContext;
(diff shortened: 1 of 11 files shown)


alacritty/src/event.rs
```diff
@@ -64,6 +65,386 @@ const MAX_SEARCH_HISTORY_SIZE: usize = 255;
...
 
+/// The event processor.
+///
+/// Stores some state from received events and dispatches actions when they are
+/// triggered.
+pub struct Processor {
+    clipboard: Clipboard,
+    scheduler: Scheduler,
+    initial_window_options: Option<WindowOptions>,
+    pub initial_window_error: Option<Box<dyn Error>>,
```
chrisduerr, line 76: I'd put `pub` stuff at the top for cleaner separation?

alacritty/src/event.rs
```diff
@@ -64,6 +65,386 @@ const MAX_SEARCH_HISTORY_SIZE: usize = 255;
...
+                | WindowEvent::Destroyed
+                | WindowEvent::ThemeChanged(_)
+                | WindowEvent::HoveredFile(_)
+                | WindowEvent::Moved(_)
+        )
+    }
+}
+
+impl ApplicationHandler<Event> for Processor {
+    // The event loop just got initialized. Create a window.
```
chrisduerr, line 197: Should use triple-slash docs on functions so they're available as function documentation.

I also don't think this comment is accurate? It's `resume`, not `created` right? So this might be called multiple times in theory? At least that's what the name suggests. The two sentences can also be combined into one probably.

alacritty/src/event.rs
```diff
@@ -64,6 +65,386 @@ const MAX_SEARCH_HISTORY_SIZE: usize = 255;
...
+    }
+
+    fn window_event(
+        &mut self,
+        _event_loop: &ActiveEventLoop,
+        window_id: WindowId,
+        event: WindowEvent,
+    ) {
+        if self.config.debug.print_events {
+            info!("winit event: {:?}", event);
```
chrisduerr, line 223: You didn't change this, but honestly it might make sense to set the log target here rather than adding `winit event: ` to every event. Having a `print_events` artificial target would probably shorten the logs a bit. No point in prefixing it with the `alacritty` target when the events are fundamentally not from alacritty.

alacritty/src/event.rs
```diff
@@ -64,6 +65,386 @@ const MAX_SEARCH_HISTORY_SIZE: usize = 255;
...
+        match event {
+            WindowEvent::RedrawRequested => {
+                window_context.handle_event(
+                    #[cfg(target_os = "macos")]
+                    _event_loop,
+                    &self.proxy,
+                    &mut self.clipboard,
+                    &mut self.scheduler,
+                    WinitEvent::WindowEvent { window_id, event },
+                );
```
chrisduerr, line 245: Maybe I'm blind, but isn't this the same in both branches? If so, why not just move it before the match statement?

alacritty/src/event.rs
```diff
@@ -64,6 +65,386 @@ const MAX_SEARCH_HISTORY_SIZE: usize = 255;
...
+            },
+        }
+    }
+
+    fn new_events(&mut self, event_loop: &ActiveEventLoop, cause: StartCause) {
+        let _ = (event_loop, cause);
```
chrisduerr, line 263: What exactly is this for? It does nothing from what I can tell.

alacritty/src/event.rs
```diff
@@ -64,6 +65,386 @@ const MAX_SEARCH_HISTORY_SIZE: usize = 255;
...
+        // Update the scheduler after event processing to ensure
+        // the event loop deadline is as accurate as possible.
+        let control_flow = match self.scheduler.update() {
+            Some(instant) => ControlFlow::WaitUntil(instant),
+            None => ControlFlow::Wait,
+        };
+        event_loop.set_control_flow(control_flow);
+    }
+
+    fn exiting(&mut self, _event_loop: &ActiveEventLoop) {
```
chrisduerr, line 421: Couldn't the body of this function just be a drop impl on `self`? I don't see which parts of this require being called manually before exiting the event loop.

Shouldn't reinvent destructors.

alacritty/src/main.rs
```diff
@@ -210,17 +211,15 @@ fn alacritty(mut options: Options) -> Result<(), Box<dyn Error>> {
...
-    // config_reloader.join().ok();
-
     // Without explicitly detaching the console cmd won't redraw it's prompt.
     #[cfg(windows)]
     unsafe {
         FreeConsole();
     }
 
+    // FIXME patch notify library to have a shutdown method.
+    // config_reloader.join().ok();
```
chrisduerr, line 221: I can't believe this comment is still around. :smile: 

Config reloader isn't even a valid identifier for any variables in this scope.

It is definitely possible to shut down the watcher cleanly at this point and I might throw up a patch for that myself in a second. For this PR I recommend just removing this entirely.

chrisduerr (changes requested):

alacritty/src/event.rs
kchibisov: It's true that it could be called multiple times, but only on Android.

alacritty/src/event.rs
kchibisov: Yeah, I forgot to remove.

alacritty/src/event.rs
kchibisov: The `self` is `&mut` borrowed and not moved into the loop, so you can not `move` self, you can probably move `event_loop` here in though, but that's about it?

alacritty/src/event.rs
chrisduerr: I'm a little bit concerned about this potentially changing in the future. I guess it's fine for now, but at the very least that needs to be documented.

alacritty/src/event.rs
chrisduerr: Things will get dropped after this anyway though I'd assume? What's the necessity of doing it in `exiting` rather than later down the line?

kchibisov: I mean, maybe you have something around in your state you don't want to drop? Passing the loop in also doesn't work with APIs which don't destroy the loop after exit.

chrisduerr: Just to make sure I understand this correctly: Both the EGL display AND the clipboard **must** be dropped before the event loop, because otherwise things explode?

And event loop shutdown automatically drops the event loop, so you can't take a return value from `event_loop.run` or whatever and manually drop that later?

kchibisov: The must be dropped because they _borrow_ the resource, the display pointer, which is getting destroyed with the loop. There's no way to prevent this sort of things in general, because no way to indicate that to event loop.

Returning the event loop back out of run and making it invalid for further reruns is probably good, but it'll still require the client to keep things _sound_.

chrisduerr: Couldn't the event loop just not destroy the display pointer?

kchibisov: And leak the resource? It could, I guess, but there're plans to just ref-count it.

alacritty/src/event.rs
kchibisov: I think we should just use `new_events`.

alacritty/src/logging.rs
```diff
@@ -28,6 +28,9 @@ const ALACRITTY_LOG_ENV: &str = "ALACRITTY_LOG";
 /// Logging target for config error messages.
 pub const LOG_TARGET_CONFIG: &str = "alacritty_config_derive";
 
+/// Logging target for winit events.
+pub const LOG_TARGET_WINIT: &str = "winit";
```
chrisduerr, line 32: How frequently does winit log? I could see this easily going out of control.

alacritty/src/event.rs
```diff
@@ -66,6 +65,374 @@ const MAX_SEARCH_HISTORY_SIZE: usize = 255;
...
+    #[cfg(unix)]
+    global_ipc_options: ParsedOptions,
+    cli_options: CliOptions,
+    config: Rc<UiConfig>,
+}
+
+impl Processor {
+    /// Create a new event processor.
+    ///
+    /// Takes a writer which is expected to be hooked up to the write end of a PTY.
```
chrisduerr, line 89: You just copied this, but it's wrong. Might as well remove it.

alacritty/src/event.rs
```diff
@@ -66,6 +65,374 @@ const MAX_SEARCH_HISTORY_SIZE: usize = 255;
...
+    }
+}
+
+impl ApplicationHandler<Event> for Processor {
+    fn resumed(&mut self, _event_loop: &ActiveEventLoop) {
+        // We use `new_events` because we target only desktop.
+    }
+
+    fn new_events(&mut self, event_loop: &ActiveEventLoop, cause: StartCause) {
+        if cause == StartCause::Init {
```
chrisduerr, line 202: Early return for a lil less indentation?

alacritty/src/event.rs
```diff
@@ -66,6 +65,374 @@ const MAX_SEARCH_HISTORY_SIZE: usize = 255;
...
+                | WindowEvent::HoveredFile(_)
+                | WindowEvent::Moved(_)
+        )
+    }
+}
+
+impl ApplicationHandler<Event> for Processor {
+    fn resumed(&mut self, _event_loop: &ActiveEventLoop) {
+        // We use `new_events` because we target only desktop.
+    }
```
chrisduerr, line 199: I don't think this comment is necessary. Not using a handler is perfectly fine and the two aren't really the same.

chrisduerr (changes requested):

alacritty/src/event.rs
kchibisov: Last time the style for thing like that was the one I picked.

I can do early return, but I don't think in this particular case it's really needed, because indention is low.

alacritty/src/logging.rs
chrisduerr: Could go for `alacritty_winit_event` just to be precise that this isn't actually logged from winit?

There's zero reason to be brief here, since it's always used through the constant anyway.

alacritty/src/event.rs
chrisduerr: The thing is when you say "instead" it assumes that there's something that *should* be handled here. However there's no **necessity** to do any handling here.

So if there's nothing you need to handle on resume, why claim to do it in `new_events` instead? We don't need to explicitly handle resume, we just use `new_events` to handle startup.

alacritty/src/event.rs
chrisduerr: I mean it's 4 levels of indentation already. With kernel indentation rules that would be 32 columns of indentation alone for the deepest level. I wouldn't call that "low".

If there were plans to handle different causes in the future, it might be different. But I don't see any immediate reason why we **ever** would handle different causes, and if we did, we'd probably switch to a match anyway. So I fail to see how an early return isn't the best option here.

chrisduerr (approved):

PR #7936, opened 2024-04-28
Fix shutdown of config monitor

chrisduerr:
This implements a coordinated shutdown of the config monitor by sending an event to its thread and waiting for the thread to terminate.

### DIFF
--- a/alacritty/src/config/monitor.rs
+++ b/alacritty/src/config/monitor.rs
@@ -1,9 +1,13 @@
 use std::path::PathBuf;
-use std::sync::mpsc::{self, RecvTimeoutError};
+use std::sync::mpsc::{self, RecvTimeoutError, Sender};
+use std::thread::JoinHandle;
 use std::time::{Duration, Instant};
 
-use log::{debug, error};
-use notify::{Config, EventKind, RecommendedWatcher, RecursiveMode, Watcher};
+use log::{debug, error, warn};
+use notify::{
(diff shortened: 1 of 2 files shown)


alacritty/src/config/monitor.rs
```diff
@@ -15,115 +19,139 @@ const DEBOUNCE_DELAY: Duration = Duration::from_millis(10);
...
+    /// Synchronously shut down the monitor.
+    pub fn shutdown(self) {
+        // Request shutdown.
+        let mut event = NotifyEvent::new(EventKind::Other);
+        event = event.set_info("shutdown");
+        let _ = self.shutdown_tx.send(Ok(event));
+
+        // Wait for thread to terminate.
+        if let Err(err) = self.thread.join() {
+            warn!("config monitor shutdown failed: {err:?}");
```
chrisduerr, line 154: Using debug here since display isn't implemented on this error for some reason. But I don't think it would ever fail anyway.

kchibisov (approved):
I think the original issue was part of the `notify` crate spawning their own thread and there was no way to `join` it.

If you look into watchers, they do send `Shutdown` for example, but none of them actually wait/join anything, so it still kind of doesn't wait properly, I think.

I'm not sure that it matters much for this patch, since we generally want to _join_, it's just, not sure that the original `FIXME` is actually fixed.

PR #7938, opened 2024-04-29
Add config file locations into config manpage

akx:
I was trying to look for this information in https://alacritty.org/config-alacritty.html, but it wasn't there! Had to spelunk around a bit to find it in the readme instead...

### DIFF
--- a/extra/man/alacritty.5.scd
+++ b/extra/man/alacritty.5.scd
@@ -19,7 +19,7 @@ following locations on UNIX systems:
 . `$HOME/.config/alacritty/alacritty.toml`
 . `$HOME/.alacritty.toml`
 
-On Windows, the config file should be located at:
+On Windows, the config file will be looked for in:
 
 . `%APPDATA%\alacritty\alacritty.toml`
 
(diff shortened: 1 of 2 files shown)


nixpulvis:
Might also be a good idea to mention #7690 if/when that lands. Not that that should block this PR, just something that came to my mind.

extra/man/alacritty.5.scd
```diff
@@ -9,6 +9,20 @@ Alacritty - TOML configuration file format.
...
+
+Alacritty doesn't create the config file for you, but it looks for one in the
+following locations on UNIX systems:
+
+. `$XDG_CONFIG_HOME/alacritty/alacritty.toml`
+. `$XDG_CONFIG_HOME/alacritty.toml`
+. `$HOME/.config/alacritty/alacritty.toml`
+. `$HOME/.alacritty.toml`
+
+On Windows, the config file should be located at:
```
chrisduerr, line 22: We should reuse the "looks for"/search verbs here. There shouldn't be a config file there, it's all optional.

chrisduerr (changes requested):

chrisduerr (approved):

PR #7940, opened 2024-04-30
Add `from_file_descriptors()` to `tty::unix`

someone13574:
Adds a function to create a `Pty` handle from file descriptors opened outside of Alacritty. This does not break backward compatibility.

### DIFF
--- a/alacritty_terminal/CHANGELOG.md
+++ b/alacritty_terminal/CHANGELOG.md
@@ -12,7 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Added `tty::unix::from_fd()` to create a TTY from a pre-opened PTY's file-descriptors
+- `tty::unix::from_fd()` to create a TTY from a pre-opened PTY's file-descriptors
 
 ### Changed
 
--- a/alacritty_terminal/CHANGELOG.md
+++ b/alacritty_terminal/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.23.0-dev
 
+### Added
+
+- Added `tty::unix::from_fd()` to create a TTY from a pre-opened PTY's file-descriptors
+
 ### Changed
 
 - **`Term` is not focused by default anymore**
--- a/alacritty_terminal/src/tty/unix.rs
+++ b/alacritty_terminal/src/tty/unix.rs
@@ -4,6 +4,7 @@ use std::ffi::CStr;
 use std::fs::File;
 use std::io::{Error, ErrorKind, Read, Result};
 use std::mem::MaybeUninit;
+use std::os::fd::OwnedFd;
 use std::os::unix::io::{AsRawFd, FromRawFd};
 use std::os::unix::net::UnixStream;
 use std::os::unix::process::CommandExt;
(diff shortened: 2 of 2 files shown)


alacritty/Cargo.toml
```diff
@@ -12,7 +12,7 @@ rust-version = "1.70.0"
 
 [dependencies.alacritty_terminal]
 path = "../alacritty_terminal"
-version = "0.23.0-dev"
```
nixpulvis, line 15: Probably shouldn't be touched here.

nixpulvis (commented):
Is there a reason for this change? A use case for a consumer application?

someone13574:
> Is there a reason for this change? A use case for a consumer application?

I'm working on flatpak support for [Zed](https://github.com/zed-industries/zed), which uses `alacritty_terminal` for it's builtin terminal, but in order to properly spawn the shell process on the host system one needs to pass the file descriptors of the command been spawned to the `--forward-fd` argument [flatpak-spawn](https://man7.org/linux/man-pages/man1/flatpak-spawn.1.html). This doesn't work however as alacritty doesn't expose the file descriptors until we have already passed the shell command to use, thus the need open the PTY ourselves so the fd's are available.

alacritty_terminal/src/tty/unix.rs
```diff
@@ -184,6 +185,16 @@ fn default_shell_command(shell: &str, user: &str) -> Command {
 pub fn new(config: &Options, window_size: WindowSize, window_id: u64) -> Result<Pty> {
     let pty = openpty(None, Some(&window_size.to_winsize()))?;
     let (master, slave) = (pty.controller, pty.user);
+    from_file_descriptors(config, window_id, master, slave)
+}
+
+/// Create a new TTY from a Pty's file descriptors.
+pub fn from_file_descriptors(
```
chrisduerr, line 192: I think `fd` is actually more universally used than `file_descriptor`.

```suggestion
pub fn from_fd(
```

alacritty_terminal/src/tty/unix.rs
```diff
@@ -184,6 +185,16 @@ fn default_shell_command(shell: &str, user: &str) -> Command {
 pub fn new(config: &Options, window_size: WindowSize, window_id: u64) -> Result<Pty> {
     let pty = openpty(None, Some(&window_size.to_winsize()))?;
     let (master, slave) = (pty.controller, pty.user);
+    from_file_descriptors(config, window_id, master, slave)
+}
+
+/// Create a new TTY from a Pty's file descriptors.
```
chrisduerr, line 191: I'm a bit bothered by TTY being all capitals, but Pty being camel case. I'd just upper them all.

chrisduerr (changes requested):
Seems simple enough and the stuff you have to do outside is also trivial (`openpty`).

So this solution makes more sense than basically exposing a function that is the same as `openpty` just with some `spawn` call attached.

nixpulvis (approved):
😄

Just letting @chrisduerr make the final review.

alacritty_terminal/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## 0.23.0-dev
 
+### Added
+
+- Added `tty::unix::from_fd()` to create a TTY from a pre-opened PTY's file-descriptors
```
chrisduerr, line 15:
```suggestion
- `tty::unix::from_fd()` to create a TTY from a pre-opened PTY's file-descriptors
```

If we add `Added` to every entry in the Added section, we're not really gaining anything.

chrisduerr (changes requested):
Another tiny nitpick. Doubt I can find anything else.

chrisduerr (approved):

someone13574:
> @someone13574 Do you need this released?

Yes, that would be awesome.

chrisduerr:
No promises on the ETA, but I'll try and get a release out this week.

PR #7950, opened 2024-05-04
Don't execute `zsh` startup files as part of our default macOS wrapper

### LINKED ISSUE #7886: ZSH pollutes environment on macOS even with `$SHELL` set to `bash`
With `$SHELL` set to `/bin/bash` `~/.zshenv` still seems to get run, which is not really great. This is a result of the choice to execute the shell in a [very roundabout way](https://github.com/alacritty/alacritty/blob/d4f2f8577f763df059653dfab733dfe6ddc06913/alacritty_terminal/src/tty/unix.rs#L179) through `zsh -c` by default. Is this really necessary? Wouldn't `/usr/bin/login -fp $user $SHELL` be sufficient as we seem to counteract the `-l` immediately using an `exec -a` anyway.

If that doesn't work for some reason I can't foresee, perhaps zsh should be executed with `-f`/`--no-rcs`

### System

OS: macOS
Version: `alacritty 0.13.2 (bb8ea18)`

nixpulvis:
Given a `~/.zshenv` with `echo hit` inside, I've tested with both `chsh -s /bin/sh` which does not print `hit` anymore, while `chsh -s /bin/zsh` continues to print `hit` as it does on master.

fixes #7886

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Notable changes to the `alacritty_terminal` crate are documented in its
 
 - Pressing `Alt` with unicode input will now add `ESC` like for ASCII input
 - Decorations use opaque style and system window background on macOS
+- No longer source `~/.zshenv` on macOS
 
 ### Fixed
 
--- a/alacritty_terminal/src/tty/unix.rs
+++ b/alacritty_terminal/src/tty/unix.rs
@@ -177,7 +177,7 @@ fn default_shell_command(shell: &str, user: &str) -> Command {
     // -p: Preserves the environment.
     //
     // XXX: we use zsh here over sh due to `exec -a`.
-    login_command.args(["-flp", user, "/bin/zsh", "-c", &exec]);
+    login_command.args(["-flp", user, "/bin/zsh", "-fc", &exec]);
     login_command
 }
 


chrisduerr (changes requested):
Have you considered any other zsh options? I'm curious if this includes `--no-globalrcs`.

nixpulvis:
> Have you considered any other zsh options? I'm curious if this includes --no-globalrcs.

I've considered it, but after reading [the man page](https://man.archlinux.org/man/zshoptions.1.en#GLOBAL_RCS) for the options I don't think it will have any effect at the moment for an average macOS user. The only file in `/etc/` on my macOS system is `/etc/zshrc` (which was there by default, but doesn't seem to be used).

It's worth mentioning that `~/.zshrc` is also not run on current master. So this patch really is just effecting the run of `~/.zshenv` as #7886 requests.

@Learath2 I'm curious if you have any input, or could verify this patch.

Learath2:
This patch is exactly what I'm running locally. I haven't hit any issues with it after running it for a month. As for adding `-d`/`--no-global-rcs`, I think it might be a good idea just to futureproof, but I also don't see it currently changing anything.

nixpulvis:
I was wondering this myself, but after a brief discussion in #alacritty, it seems like the stance is that users who run their own `shell.program` should do the wrapping themselves if they care about it.

I'm personally still not convinced, since I hardly understand the need for the wrapping, I assume most users are also unaware of the need and will therefor do something like `shell.program = "/bin/bash"`. In fact this is evident if you scan through the comments of many users in our issue tracker and elsewhere online.

But that could be considered a separate issue to this PR either way.

kchibisov:
>The only reason I didn't contribute this patch myself was that I was wondering whether we should be wrapping the users shell setting from the config the same way we wrap the one set by chsh. Then I got distracted by some other work.

The point was to run login shell with `-` as prefix. iterm2 for example has a separate binary for that iirc, but it effectively does the same thing as `zsh` wrapping we're doing.

And I don't think it wraps user specified programs, since users can properly start their shells.

chrisduerr:
Most importantly `login` will automatically change to the home directory. So if you change your shell and your working directory config option doesn't work anymore, you can fix that yourself. If you don't care about that, then you don't have to care about it, it's just that Alacritty's default shell **has** to care about it.

chrisduerr (changes requested):
Seems like the only thing that gets sourced now is `/etc/zsh/zshenv` and I didn't see an easy solution to fix that. Should be fine as-is, probably isn't going to make things worse at least.

Could you add a changelog entry documenting these changes?

chrisduerr (approved):

PR #7952, opened 2024-05-06
Implement dragging and resizing of borderless window

hborchardt:
Hi, I like using alacritty without decorations (on windows, that's the only alternative to full title bar), but I found it difficult to move and resize the window.

I saw that winit provides the `drag_window` and `drag_resize_window` functions for the use case of defining custom click target areas to move and resize a window.

So I implemented moving the window when dragging the top padding, and resizing it when dragging the bottom-right corner padding, as well as corresponding `CursorIcon` changes.

I hope you like it, I am happy about any criticism.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Notable changes to the `alacritty_terminal` crate are documented in its
 
 ## 0.14.0-dev
 
+### Added
+
+- Moving and resizing of windows without decorations by dragging the padding at the top or bottom-right
+
 ### Changed
 
 - Pressing `Alt` with unicode input will now add `ESC` like for ASCII input
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -34,8 +34,8 @@ use winit::monitor::MonitorHandle;
 #[cfg(windows)]
 use winit::platform::windows::IconExtWindows;
 use winit::window::{
-    CursorIcon, Fullscreen, ImePurpose, Theme, UserAttentionType, Window as WinitWindow,
-    WindowAttributes, WindowId,
+    CursorIcon, Fullscreen, ImePurpose, ResizeDirection, Theme, UserAttentionType,
+    Window as WinitWindow, WindowAttributes, WindowId,
 };
 
 use alacritty_terminal::index::Point;
@@ -370,9 +370,8 @@ impl Window {
         }
     }
 
-    pub fn drag_resize_window(&self) {
-        if let Err(err) = self.window.drag_resize_window(winit::window::ResizeDirection::SouthEast)
-        {
+    pub fn drag_resize_window(&self, direction: ResizeDirection) {
+        if let Err(err) = self.window.drag_resize_window(direction) {
             debug!("Unable to initiate resizing the window: {}", err);
         }
     }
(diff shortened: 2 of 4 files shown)


nixpulvis:
Upon first look at this PR a lot of things come to mind:

1. The current state of window options on macOS are somewhat overly complicated, i.e. does anyone actually use "Transparent"?
2. This begs for the ability to change `window.padding.top` without changing `window.padding.bottom`, which I know we've had requests for before, but have been rejected for some reason
3. It's unclear to me if dragging in the left/right/bottom padding area should also allow moving the window, since we're going outside the standard UX a bit anyway. I might be inclined to say yes.
4. It remains impossible to select text in the top row of a window with `decoration = "Buttonless"` unless the padding is large enough. It would be nice to change the cursor in these cases as well.

All that said, I don't expect all these points to necessarily be addressed in a single PR, I just view them as related.

As for this PR, I did notice some bugs with the cursor setting. The easiest to replicate is if you start a selection and drag into the top padding area. The cursor will flicker between a move cursor and the default. I also don't know why the SE corner is special, I would expect to be able to resize from any corner, no?

chrisduerr:
> does anyone actually use "Transparent"?

Yes

> but have been rejected for some reason

No valid usecases.

> It remains impossible to select text in the top row of a window with decoration = "Buttonless"

This is a separate issue.

nixpulvis:
>> does anyone actually use "Transparent"?

> Yes

Really? Because without `window.padding.top` as an option, this is basically guaranteed to draw the traffic lights on top of text. Who would want that?

hborchardt:
> 1. window options on macOS are somewhat overly complicated

I can't comment on macOS as I don't use it. I just try to make the decoration-less mode more usable.

> 2. change `window.padding.top`

Personally I have no issue with the small padding I have to hit to move the window, and I like that it is slim and symmetric and configurable as it is.

> 3. dragging in the left/right/bottom padding area should also allow moving the window

Moving the window using the other paddings is not quite intuitive to me, and is not necessary. I imagine it as a collapsed title bar, not the border gaining dragging powers.

> 4. `decoration = "Buttonless"`

Can't comment on macOS.

> ... start a selection and drag into the top padding area. The cursor will flicker.

Will try to fix that, thanks.

> I also don't know why the SE corner is special

I went for the minimalistic approach here (or call me lazy 😄), with my implementation you can already position and resize the window arbitrarily. There's also a precedent of things that are only resizable using the bottom-right corner, like this textarea field in the browser for example. But if people would require to resize from all corners, that can surely be done.

nixpulvis:
I would strongly push for resizing from both bottom corners at least.

hborchardt:
I addressed the flicker issue, enabled resizing on the south-east corner as well, and in general refactored my changes to put things in better places and make it easier to read and extend, I hope. Feel free to have a second try.

nixpulvis:
Sadly my Windows machine is currently broken and testing this on macOS seems to yield different results (all corners and edges allow resizing, as expected). I'll leave it to someone else to confirm this PR unless I get the time to put windows on an old laptop I've got lying around again.

nixpulvis:
@chrisduerr any thoughts on this. I don’t want to completely forget about it. I’ll have some more time again soon.

chrisduerr:
> any thoughts on this?

I have no interest in this feature at all, because it is completely unnecessary on any sane WM.

hborchardt:
I respect that, I just hope I will find a sane WM for Windows then

PR #7959, opened 2024-05-07
Fix mouse mode bindings with multiple actions

EBADBEEF:
The following config was broken:
```
  [mouse]
  bindings = [
  { mouse = "Right", mods = "Shift", action = "Copy"            },
  { mouse = "Right", mods = "Shift", action = "ClearSelection"  },
  ]
```

Only the first action was applied. Change to allow more than one exact match in mouse mode with shift held, but keep the logic to not allow fallback search if any exact match was found.

Fixes: 1a143d11 (Fix trigger of normal bindings in mouse mode)


chrisduerr (changes requested):
Could you add a changlog entry?

The code looks good to me.

chrisduerr (approved):

PR #7960, opened 2024-05-08
escape the '&' char in windows links

zangold:
The & character is special in windows, so escape it when passing strings that contain '&' as arguments to cmd.exe.

This fixes a bug where ctrl-clicking on a link with multiple parameters would only open the substring of the link up to (and not including) the first ampersand, e.g.,

http://www.foo.bar/index.html?foo=bar&baz=wat

would direct me to:

http://www.foo.bar/index.html?foo=bar

### DIFF
--- a/alacritty/src/event.rs
+++ b/alacritty/src/event.rs
@@ -1111,6 +1111,10 @@ impl<'a, N: Notify + 'a, T: EventListener> input::ActionContext<T> for ActionCon
         match &hint.action() {
             // Launch an external program.
             HintAction::Command(command) => {
+                // Windows requires an escape character (^) before occurrences of ampersand (&)
+                #[cfg(windows)]
+                let text = text.replace('&', "^&");
+
                 let mut args = command.args().to_vec();
                 args.push(text);
                 self.spawn_daemon(command.program(), &args);


chrisduerr:
Are you suggesting this is something inherent to spawning *any* command on Windows? Because I don't see how that would be true.

zangold:
I'm having difficulty finding any official documentation from microsoft on this, but some stackoverflow answers seem to suggest that cmd.exe uses '&' as a way to specify multiple commands on the same line, equivalent to putting a semicolon in a line in bash.

On further review, it seems as though delimiting the URL with double quotes also fixes this issue:
- `start "" http://www.foo.bar/index.html?foo=bar&baz=wat` opens the shortened link (and prints a message about `baz` not being a valid command)
- `start "" "http://www.foo.bar/index.html?foo=bar&baz=wat"` opens the full link, without the above error message about `baz`

I can update the PR to quote-delimit the URL instead of escaping the & character, if you prefer.

chrisduerr:
Commands launched by hints don't have to execute `cmd.exe`, it can be executed by any command. Your PR doesn't make sense to me.

zangold:
I think you might be right that there is a better approach to this -- per the Rust docs on std::process, it sounds like passing args into CreateProcessW is janky, at best: https://doc.rust-lang.org/std/process/index.html#windows-argument-splitting

It's worth noting that, because ampersands aren't correctly escaped in the daemon processes that alacritty launches, this is actually a security concern -- I think one would be able to construct some text with hyperlink terminal directives that can execute an arbitrary command on the user's computer if they control-click on it.

I'll see if I can come up with a solution that fixes the control-click behaviour and tries to sanitize the args passed into CreateProcessW.

chrisduerr:
> I'll see if I can come up with a solution that fixes the control-click behaviour and tries to sanitize the args passed into CreateProcessW.

Seems like you couldn't. If you or anyone else is still interested in this, please open an issue.

PR #7963, opened 2024-05-10
Add `strength` factor and setting for visual bell

fishBone000:
The visual bell has a color setting, however it's more like a "solid" color.  
For example, user might want to set the color to a less intense one (#808080 rather than #FFFFFF) to not flash his/her eyes blind, but setting it too low will actually make the terminal flashed in "black", instead of flash in dim white. Setting it to a moderate value can be a good solution, but for bright areas on the terminal, it's still flashing "black".  

(I know this is actually not a big issue, but I have been itchy to make it better)

This `strength` factor solves this problem. This factor is applied to visual bell intensity. Users now can have softer (or stronger, if they wish) bell flashes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Notable changes to the `alacritty_terminal` crate are documented in its
 
 ## 0.14.0-dev
 
+### Added
+
+- New `strength` factor and setting for the visual bell effect
+
 ### Changed
 
 - Pressing `Alt` with unicode input will now add `ESC` like for ASCII input
--- a/alacritty/src/config/bell.rs
+++ b/alacritty/src/config/bell.rs
@@ -18,6 +18,9 @@ pub struct BellConfig {
 
     /// Visual bell duration in milliseconds.
     duration: u16,
+
+    /// Stength of visual effect, in percent.
+    strength: u16,
 }
 
 impl Default for BellConfig {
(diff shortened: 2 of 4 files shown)


chrisduerr (changes requested):
I don't see how this is useful compared to just precomputing the alpha against your background color?

fishBone000:
:( True, but at least you don't need to precompute don't you think?  
Plus the background color can be temporarily changed by TUI programs, like nvim with themes installed, and old-fashion-looked nmtui.

chrisduerr:
@kchibisov Any feelings about this? Realistically a properly configured terminal doesn't have a different background color inside and outside of vim. So the only real benefit I see for this is with transparent setups, but I don't particularly care about those anyway.

kchibisov:
I also don't really see a value in this for the following reasons:

1. You can precompute yourself.
2. If you change colors between vim/not-vim you need a different value of the strength, because I don't see how it actually solves anything given that it's just changes the alpha of the color.

Like I could only suggest (as I did with some other issue) to add special cased animation that just not that intense, etc.

You can also change the bell color in the current window dynamically with `IPC` from vim hooking them to enter/close, so I don't think it's a _big_ issue here.

fishBone000:
Alright...Thought this PR might be appealing, I will keep this PR as a patch for myself then.

> Like I could only suggest (as I did with some other issue) to add special cased animation that just not that intense, etc.

Hmm sounds good, I also want more variant animations, but I don't know Rust and I don't want this feature that much. 

Anyways, closing this PR.

PR #7965, opened 2024-05-10
Improve transparency on macOS

madsmtm:
Extension upon https://github.com/alacritty/alacritty/pull/7928, see that for discussion. I've improved a bit on the window background, by setting the alpha channel to follow the window opacity.

When `window.opacity < 1`, we do two things:
- Set the title bar as transparent (new).
- Disable window border shadows (existing behaviour).

### DIFF
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -22,6 +22,7 @@ use std::fmt::{self, Display, Formatter};
 use {
     cocoa::appkit::NSColorSpace,
     cocoa::base::{id, nil, NO, YES},
+    objc::runtime::{Class, Object},
     objc::{msg_send, sel, sel_impl},
     winit::platform::macos::{OptionAsAlt, WindowAttributesExtMacOS, WindowExtMacOS},
 };
@@ -345,6 +346,8 @@ impl Window {
 
     pub fn set_transparent(&self, transparent: bool) {
         self.window.set_transparent(transparent);
+        #[cfg(target_os = "macos")]
+        set_background_transparent_macos(&self.window, transparent)
     }
 
     pub fn set_blur(&self, blur: bool) {
@@ -494,5 +497,26 @@ fn use_srgb_color_space(window: &WinitWindow) {
 
     unsafe {
         let _: () = msg_send![raw_window, setColorSpace: NSColorSpace::sRGBColorSpace(nil)];
+    }
+}
+
+/// If transparent, set background color to NSColor.clearColor, otherwise set
+/// it to NSColor.windowBackgroundColor so that MacOS properly draws the
+/// window's border and opaque title bar.
+#[cfg(target_os = "macos")]
+fn set_background_transparent_macos(window: &WinitWindow, transparent: bool) {
+    let raw_window = match window.raw_window_handle() {
+        RawWindowHandle::AppKit(handle) => handle.ns_window as id,
+        _ => return,
+    };
+
+    unsafe {
+        let color_class = Class::get("NSColor").unwrap();
+        let bg_color: *const Object = if transparent {
+            msg_send![color_class, clearColor]
+        } else {
+            msg_send![color_class, windowBackgroundColor]
+        };
+        let _: () = msg_send![raw_window, setBackgroundColor: bg_color];
     }
 }
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -349,7 +349,9 @@ impl Window {
     pub fn set_transparent(&self, transparent: bool) {
         self.window.set_transparent(transparent);
         #[cfg(target_os = "macos")]
-        set_background_transparent_macos(&self.window, transparent)
+        set_background_transparent_macos(&self.window, transparent);
+        #[cfg(target_os = "macos")]
+        set_titlebar_transparent_macos(&self.window, transparent);
     }
 
     pub fn set_blur(&self, blur: bool) {
@@ -521,4 +523,17 @@ fn set_background_transparent_macos(window: &WinitWindow, transparent: bool) {
         };
         let _: () = msg_send![raw_window, setBackgroundColor: bg_color];
     }
+}
+
+/// Set the transparency of the window titlebar on macOS.
+#[cfg(target_os = "macos")]
+fn set_titlebar_transparent_macos(window: &WinitWindow, transparent: bool) {
+    let raw_window = match window.raw_window_handle() {
+        RawWindowHandle::AppKit(handle) => handle.ns_window as id,
+        _ => return,
+    };
+
+    let transparent = if transparent { YES } else { NO };
+
+    let _: () = unsafe { msg_send![raw_window, setTitlebarAppearsTransparent: transparent] };
 }
--- a/alacritty/src/display/window.rs
+++ b/alacritty/src/display/window.rs
@@ -168,6 +168,8 @@ impl Window {
             .with_title(&identity.title)
             .with_theme(config.window.theme())
             .with_visible(false)
+            // Needed on X11 and perhaps on Windows to enable configuring
+            // transparency later on.
             .with_transparent(true)
             .with_blur(config.window.blur)
             .with_maximized(config.window.maximized())


alacritty/src/display/window.rs
```diff
@@ -182,8 +183,11 @@ impl Window {
         window.set_ime_allowed(true);
         window.set_ime_purpose(ImePurpose::Terminal);
 
-        // Set initial transparency hint.
-        window.set_transparent(config.window_opacity() < 1.);
+        // Set initial background opacity and transparency
+        #[cfg(target_os = "macos")]
+        set_titlebar_transparent(&window, config.window_opacity() < 1.);
+        #[cfg(target_os = "macos")]
+        set_background_opacity(&window, config.window_opacity() as _);
```
chrisduerr, line 190: You're not calling `window.set_transparent` here anymore, why is that?

alacritty/src/display/window.rs
```diff
@@ -345,6 +349,13 @@ impl Window {
...
     pub fn set_transparent(&self, transparent: bool) {
         self.window.set_transparent(transparent);
+        #[cfg(target_os = "macos")]
+        set_titlebar_transparent(&self.window, transparent);
+    }
+
+    pub fn set_background_opacity(&self, _opacity: f32) {
+        #[cfg(target_os = "macos")]
+        set_background_opacity(&self.window, _opacity as _)
     }
```
chrisduerr, line 359: Is there ever a scenario where you'd want to call `set_background_opacity` without setting transparency? Seems like the two functions can just be combined.

alacritty/src/display/window.rs
```diff
@@ -496,3 +507,33 @@ fn use_srgb_color_space(window: &WinitWindow) {
...
+#[cfg(target_os = "macos")]
+fn set_background_opacity(window: &WinitWindow, opacity: CGFloat) {
+    let raw_window = match window.raw_window_handle() {
+        RawWindowHandle::AppKit(handle) => handle.ns_window as id,
+        _ => return,
+    };
+
+    // If transparent, fill background with NSColor.clearColor,
+    // otherwise fill with NSColor.windowBackgroundColor so that MacOS
+    // properly draws the window's border and opaque title bar.
```
chrisduerr, line 520: This sounds like the window's background is filled, but surely that's not what is happening, right? Just the titlebar background?

alacritty/src/display/window.rs
```diff
@@ -496,3 +507,33 @@ fn use_srgb_color_space(window: &WinitWindow) {
         let _: () = msg_send![raw_window, setColorSpace: NSColorSpace::sRGBColorSpace(nil)];
     }
 }
+
+#[cfg(target_os = "macos")]
+fn set_background_opacity(window: &WinitWindow, opacity: CGFloat) {
```
chrisduerr, line 512: These macOS-specific functions should indicate their macOS-specificity. They're also lacking function documentation.

alacritty/src/display/window.rs
```diff
@@ -167,7 +168,7 @@ impl Window {
             .with_title(&identity.title)
             .with_theme(config.window.theme())
             .with_visible(false)
-            .with_transparent(true)
+            .with_transparent(config.window_opacity() < 1.)
```
chrisduerr, line 171: I believe we've had it this way in the past and that changed. Are you sure this is safe on all platforms? Because I'm not sure that this doesn't change Alacritty's behavior.

alacritty/src/display/window.rs
```diff
@@ -182,8 +183,11 @@ impl Window {
         window.set_ime_allowed(true);
         window.set_ime_purpose(ImePurpose::Terminal);
 
-        // Set initial transparency hint.
-        window.set_transparent(config.window_opacity() < 1.);
+        // Set initial background opacity and transparency
```
chrisduerr, line 186:
```suggestion
        // Set initial background opacity and transparency.
```

chrisduerr (changes requested):
Missing changelog entry.

alacritty/src/display/window.rs
madsmtm: Because it's called in `WindowBuilder::with_transparent` instead.

chrisduerr: But only on macOS. Contrary to Apple's belief, there are other operating systems. :)

chrisduerr: That was also called before though? Are you suggesting a winit update has removed the necessity for this on all platforms?

sfsam:
I'm seeing the same thing I observed in the discussion of #7928: No borders and a white title bar even though `decorations_theme_variant = "Dark"` is set. I think this is due to something that was changed in the move to winit 0.3.0. I'm on MacOS Sonoma 14.4.1. FWIW, my window config:
```
[window]
decorations_theme_variant = "Dark"
decorations = "full"
option_as_alt = "OnlyLeft"
resize_increments = true
dimensions = { columns = 100, lines = 24 }
```

nixpulvis:
> Set the title bar as transparent (new)

Seems to me this may be undesired. I hate to think we need yet another macOS specific configuration for it though.

alacritty/src/display/window.rs
madsmtm: Yes, `WindowBuilder::with_transparent` nowadays does the exact same thing as `window.set_transparent` (call `setOpaque:`). In the past, there was confusion about this, because `WindowBuilder::with_transparent` set the window background as well (but it no longer does since https://github.com/rust-windowing/winit/pull/3657).

kchibisov:
The end window opacity is not as it should be, it's just generally more _opaque_ compared to what it was before or glitches.

alacritty/src/display/window.rs
madsmtm: They probably could be merged, but I think it makes sense to keep them as separate function to self-document specifically what it is that they do.

alacritty/src/display/window.rs
madsmtm: You're right that the behaviour here is different on X11 and Windows, didn't consider other platforms when making this. Have reverted the change.

madsmtm:
I've updated the PR and reverted some of the changes I probably mistakenly thought were for the better, though I'm too unsure of what the desired behaviour is (I don't like transparent windows myself, and don't use Alacritty).

The first two commits should bring Alacritty back in line with what it was pre winit 0.30.

Would be happy if someone else would take over for the rest.

nixpulvis:
I should be able to finish testing and implementing this PR, though my macOS API knowledge is limited and it may take me a little while to get around to everything. Right now it seems the main things that need changing are with respect to the titlebar and fixing whatever is causing #7969. At first I didn't notice that issue myself, but now transparency is broken for me as well, without any changes to the code or configuration, so it seems some macOS intermittent issue. Fun.

I assume the best way to do this would be to make a new PR with the current state of this one and push to that myself as I make progress?

chrisduerr:
> I assume the best way to do this would be to make a new PR with the current state of this one and push to that myself as I make progress?

Probably easiest this way, yes.

nixpulvis:
So after catching up on this issue, it seems like https://github.com/rust-windowing/winit/pull/3722 fixes the main issue (#7969), so what's left is to figure out what we actually want, if anything.

I'm unsure that having a transparent titlebar is the right default because it can be achieved with `window.decorations = "Transparent"` (however, I still think that could use with some improvement so the content area doesn't overlap with the titlebar buttons, but that's an existing concern).

I also don't know if it's better to have drop shadows or not with a transparent background. My instinct would be to have them, but given that we don't have them already, I'm fine leaving it as is.

Looking back at #7928, it seems like everything there has been addressed.

Here's a screenshot of an alacritty window (`opacity = 0.75`) from master with winit pointing to https://github.com/rust-windowing/winit/pull/3722:

Screenshot 2024-06-10 at 10 35 02 AM

and here's one without `opacity` set:

Screenshot 2024-06-10 at 10 40 00 AM

Yes the transparent window's titlebar is a less dark color, but I don't see the red bleeding into it like it appears to be in the original PR and I believe this behavior looks reasonable.

madsmtm:
I'll note another option: Use `-[NSWindow setAlphaValue:]`, this also sets the opacity of all the decorations (the border and the title bar) (in this mode, the application should internally render as-if it was opaque)

nixpulvis:
@madsmtm I was looking at that the other day. Might be a better solution in that it might look better, but I'm not sure it's worth platform specific code if `winit` still implements transparency the other way. We don't want to change how things work on Linux/Windows.

kchibisov:
@sfsam I'd suggest to bring it upstream if after the https://github.com/alacritty/alacritty/pull/8043 the issue will be still the same, because I can not repro it myself and everything works as expected(or before).

It's true that internally winit touched a lot of code on macOS, so there's a chance that with 14+ something is not working, given that I can only test version 12.

kchibisov:
Closing in favor of https://github.com/alacritty/alacritty/pull/8043

PR #7971, opened 2024-05-12
improved default color scheme

singh-jagjot:
The default color scheme of alacritty has 'color.normal.black' as "#181818" which is the same as 'color.primary.background' due to which text with 'normal.black' color is impossible to view. This pull request fixes that issue by making the value of 'normal.color' to '#333333' (just a lighter shade). This will improve the experience of default color scheme users.

Before:
Screenshot 2024-05-12 at 1 19 42 PM

After:
Screenshot 2024-05-12 at 1 20 05 PM

### DIFF
--- a/alacritty/src/config/color.rs
+++ b/alacritty/src/config/color.rs
@@ -201,7 +201,7 @@ pub struct NormalColors {
 impl Default for NormalColors {
     fn default() -> Self {
         NormalColors {
-            black: Rgb::new(0x18, 0x18, 0x18),
+            black: Rgb::new(0x33, 0x33, 0x33),
             red: Rgb::new(0xac, 0x42, 0x42),
             green: Rgb::new(0x90, 0xa9, 0x59),
             yellow: Rgb::new(0xf4, 0xbf, 0x75),


singh-jagjot:
@chrisduerr Is there any specific reason for closing this? Just curious. I believe this is an improvement, text in the terminal is visible to a certain extent after this fix. I know this is just a minor change, but one can use the default color scheme without any hiccups. 🙂

chrisduerr:
I'm not accepting any colorscheme changes since that topic is far too subjective.

singh-jagjot:
@chrisduerr Thanks for the reply. I do agree with you and I'm not trying to put my colorscheme above others. It's just one color. This issue kept me searching for fixes. Without this fix, the only possible solution to this will be to use another theme or edit alacritty.toml manually. Because the background and normal.black color is same in default color scheme, you cannot read the terminal in case text is of normal.black color. If you think this is still unnecessary, that's fine by me. 🙂

kchibisov:
>Because the background and normal.black color is same in default color scheme, you cannot read the terminal in case text is of normal.black color. If you think this is still unnecessary, that's fine by me. 🙂

it's like that in 99% of the themes. Just use bright black, a lot of software which uses regular one does it in bold to request bright, for whatever reason. Better to just patch such software to use bright black as they want to.

PR #7996, opened 2024-05-23
fix IO Safety violation from dropping `OwnedFd` multiple times

jakobhellermann:
fixes https://github.com/zed-industries/zed/issues/12114

> When running zed (cargo zed) on linux, the process crashes trying to open the alacritty terminal with
>`fatal runtime error: IO Safety violation: owned file descriptor already closed`

The error only happens on nightly because they recently enabled late debug assertions in std which work on the compiled binary: https://github.com/rust-lang/rust/commit/1ba00d9cb2fcfef464b6a188fa3a7543c66eecaa

I don't know much about linux fd handling so I don't know if `dup`ing here is the correct solution, I got this suggestion from chatgpt:
> In Rust, when you want to attach the same file descriptor (fd) to multiple streams (like stdin and stdout) for a Command, you need to duplicate the file descriptor to avoid having it dropped twice. You can use the nix crate to achieve this safely. The dup system call duplicates the file descriptor, giving you a new one that you can safely pass to another Stdio instance.

### DIFF
--- a/alacritty_terminal/src/tty/unix.rs
+++ b/alacritty_terminal/src/tty/unix.rs
@@ -212,8 +212,8 @@ pub fn from_fd(config: &Options, window_id: u64, master: OwnedFd, slave: OwnedFd
     };
 
     // Setup child stdin/stdout/stderr as slave fd of PTY.
-    builder.stdin(rustix_openpty::rustix::io::dup(&slave)?);
-    builder.stderr(rustix_openpty::rustix::io::dup(&slave)?);
+    builder.stdin(slave.try_clone()?);
+    builder.stderr(slave.try_clone()?);
     builder.stdout(slave);
 
     // Setup shell environment.
--- a/alacritty_terminal/src/tty/unix.rs
+++ b/alacritty_terminal/src/tty/unix.rs
@@ -212,9 +212,6 @@ pub fn from_fd(config: &Options, window_id: u64, master: OwnedFd, slave: OwnedFd
     };
 
     // Setup child stdin/stdout/stderr as slave fd of PTY.
-    // Ownership of fd is transferred to the Stdio structs and will be closed by them at the end of
-    // this scope. (It is not an issue that the fd is closed three times since File::drop ignores
-    // error on libc::close.).
     builder.stdin(rustix_openpty::rustix::io::dup(&slave)?);
     builder.stderr(rustix_openpty::rustix::io::dup(&slave)?);
     builder.stdout(slave);
--- a/alacritty_terminal/src/tty/unix.rs
+++ b/alacritty_terminal/src/tty/unix.rs
@@ -5,10 +5,10 @@ use std::fs::File;
 use std::io::{Error, ErrorKind, Read, Result};
 use std::mem::MaybeUninit;
 use std::os::fd::OwnedFd;
-use std::os::unix::io::{AsRawFd, FromRawFd};
+use std::os::unix::io::AsRawFd;
 use std::os::unix::net::UnixStream;
 use std::os::unix::process::CommandExt;
-use std::process::{Child, Command, Stdio};
+use std::process::{Child, Command};
 use std::sync::Arc;
 use std::{env, ptr};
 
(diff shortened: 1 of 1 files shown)


alacritty_terminal/src/tty/unix.rs
```diff
@@ -212,12 +212,9 @@ pub fn from_fd(config: &Options, window_id: u64, master: OwnedFd, slave: OwnedFd
...
     // Setup child stdin/stdout/stderr as slave fd of PTY.
-    // Ownership of fd is transferred to the Stdio structs and will be closed by them at the end of
-    // this scope. (It is not an issue that the fd is closed three times since File::drop ignores
-    // error on libc::close.).
-    builder.stdin(unsafe { Stdio::from_raw_fd(slave_fd) });
-    builder.stderr(unsafe { Stdio::from_raw_fd(slave_fd) });
-    builder.stdout(unsafe { Stdio::from_raw_fd(slave_fd) });
+    builder.stdin(rustix_openpty::rustix::io::dup(&slave)?);
+    builder.stderr(rustix_openpty::rustix::io::dup(&slave)?);
+    builder.stdout(slave);
```
chrisduerr, line 217: The slave FDs were previously closed in the child, is there a reason why you're not doing this anymore?

alacritty_terminal/src/tty/unix.rs
```diff
@@ -212,12 +212,9 @@ pub fn from_fd(config: &Options, window_id: u64, master: OwnedFd, slave: OwnedFd
...
 
     // Setup child stdin/stdout/stderr as slave fd of PTY.
-    // Ownership of fd is transferred to the Stdio structs and will be closed by them at the end of
-    // this scope. (It is not an issue that the fd is closed three times since File::drop ignores
-    // error on libc::close.).
-    builder.stdin(unsafe { Stdio::from_raw_fd(slave_fd) });
-    builder.stderr(unsafe { Stdio::from_raw_fd(slave_fd) });
-    builder.stdout(unsafe { Stdio::from_raw_fd(slave_fd) });
+    builder.stdin(rustix_openpty::rustix::io::dup(&slave)?);
+    builder.stderr(rustix_openpty::rustix::io::dup(&slave)?);
```
chrisduerr, line 216: Definitely shouldn't go through `rustix_openpty` here.

chrisduerr (changes requested):

alacritty_terminal/src/tty/unix.rs
kchibisov, line 217: You can just `try_clone` the `fd`, you don't need the `dup` from the `rustix` iirc.

kchibisov (changes requested):
I think the issue is that because the FD is the same everywhere the `drop` will fail in one of the instances, regardless how you close the fd and whether the fd was raw.

In general, this should be _fine_.

alacritty_terminal/src/tty/unix.rs
kchibisov: it's all the same as before, it's just `fd`'s are valid in each call so the `assert` in `drop`  doesn't get triggered instead of closing multiple times.

chrisduerr (approved):

kchibisov (approved):

bjorn3:
This looks like a correct fix. I disagree that the previous behavior was harmless though: If right after the first close another thread opens a new file and it happens to be assigned the same fd as the tty slave, the next close would close the file that the other thread just opened.

kchibisov:
I guess it's safe for `alacritty` itself, but not for other programs during stuff in parallel to tty opening, since alacritty is single threaded when it comes to fd opens, etc.

ConradIrwin:
We can (probably?) revert https://github.com/alacritty/alacritty/pull/7681 now this is merged.

nixpulvis:
Is there really any reason to revert it though? Better to show an error without crashing than to crash, no?

chrisduerr:
> Is there really any reason to revert it though? Better to show an error without crashing than to crash, no?

No reason to revert, no.

Issue #8002, opened 2024-05-25
Replace cstr! macro with C-string literal syntax


dan-arnold:
Currently, in the renderer [there is a macro](https://github.com/alacritty/alacritty/blob/cacdb5bb3b72bad2c729227537979d95af75978f/alacritty/src/renderer/mod.rs#L36) called `cstr!` which is used to get a pointer to a string literal as a CStr.

From <https://doc.rust-lang.org/nightly/edition-guide/rust-2021/c-string-literals.html>:
> Starting with Rust 1.77, C-strings can be written using C-string literal syntax with the c or cr prefix.
> ...
> Now, C-strings can be written directly using literal syntax notation, which will generate a value of type [&core::ffi::CStr](https://doc.rust-lang.org/nightly/core/ffi/struct.CStr.html) which is automatically terminated with a NUL byte.

This is the same behavior as the `cstr!` macro.

I think we should remove the `cstr!` macro and replace calls to it with `c"mystring"` instead. However, this change would require a bump of the rust-version to 1.77.

chrisduerr:
I don't mind this change, but widespread support for 1.77.0 is probably still like half a year out.

PR #8003, opened 2024-05-25
feat: Remove unsafe cstr macro, bump rust-version to 1.77

dan-arnold:
Per #8002, this PR removes the `cstr!` macro and replaces it with C-string literal syntax. Because that syntax wasn't introduced until Rust 1.77, I have bumped the rust-version as well. All tests pass and benchmarked performance seems excellent. It may need to be bumped in other Cargo.toml files as well, but am unsure, please advise. Thanks.


chrisduerr (changes requested):
MSRV is too high.

PR #8016, opened 2024-06-02
Arch Linux - package changed it's name from pkg-config to pkgconf

roland-5:
https://archlinux.org/packages/core/x86_64/pkgconf/


chrisduerr (changes requested):
Using `pkg-config` still works just fine, so I see no reason to change it.

PR #8026, opened 2024-06-08
Bump glutin to 0.32.0

kchibisov:


alacritty/src/display/window.rs
```diff
@@ -444,14 +444,15 @@ impl Window {
     /// This prevents rendering artifacts from showing up when the window is transparent.
     #[cfg(target_os = "macos")]
     pub fn set_has_shadow(&self, has_shadows: bool) {
-        let raw_window = match self.raw_window_handle() {
-            RawWindowHandle::AppKit(handle) => handle.ns_window as id,
+        let ns_view = match self.window.window_handle().unwrap().as_raw() {
```
chrisduerr, line 447:
```suggestion
        let ns_view = match self.raw_window_handle() {
```

This should still work just fine, right? We have access to self here, not just window.

chrisduerr (changes requested):

alacritty/src/display/window.rs
kchibisov: Yeah, it should, I just did it the same way as other function.

chrisduerr (approved):

PR #8063, opened 2024-06-24
star search preliminaries

ramseyharrison:


alacritty/src/config/bindings.rs
```diff
@@ -327,6 +327,8 @@ pub enum ViAction {
     InlineSearchNext,
     /// Jump to the previous inline search match.
     InlineSearchPrevious,
+    ///Star search
+    StarSearch,
```
chrisduerr, line 331: Since you can rebind things it doesn't make sense to name them after their binding. It should be named after what it does.

```suggestion
    /// Start forward search for the word at the current cursor position.
    SearchCurrentWordForward,
```

chrisduerr (changes requested):
I'm not reviewing unformatted code.

chrisduerr (changes requested):
This does not behave the way vim does and we do not need a separate search function for it. Also I'd expect my suggestions to either be applied or to add a comment on why a different approach was made, otherwise we're just wasting our time here.

I would expect `*` to put me in an identical state as manually starting a search for the word under the cursor would.

chrisduerr:
> If I'm not mistaken, in vim, pressing * above something like "-apple" will move the cursor to the next instance of "apple", not "-apple". Same thing for many symbols, but not for example, for '_'. For instance, pressing * above "_apple" will in fact search for the text "_apple".

The separator characters we can just copy from the existing word motion, but what is far more important here is that it not only moves the cursor to the next matching word, it also stores this word as the current search, allowing you to jump between instances with n/N etc.

> My first implementation pays no heed to this however, but if the cursor is midway throughout a word, it does first backtrack to the beginning of a word and then moves forwards to its end. It constructs a regexp from the text between these two points, and then launches a search. Words in this case are delimited by spaces, which is not the way vim does it. That's what this first version was trying to do, but perhaps I'm thinking about it wrong.

There should be no reason for any of your own code after you have found the word in question. We already have code to search for text, it should be reused.

ramseyharrison:
> The separator characters we can just copy from the existing word motion, but what is far more important here is that it not only moves the cursor to the next matching word, it also stores this word as the current search, allowing you to jump between instances with n/N etc.

Thanks, this clarifies what I need to do.

ramseyharrison:
I've gone ahead and rewritten my pull request,  I think this might be more in line with what you wanted.
Thanks!

chrisduerr:
Just to keep you updated: I don't plan to make any more reviews of this PR at this point. If I feel like working on it, I'll just clean it up myself.

PR #8069, opened 2024-06-28
Fix search bug with wrapline on first character

### LINKED ISSUE #8060: Bug: backward search on single line goes in cycles
When using backward search, it sometimes goes in cycles without moving a cursor to some search result. In example video, before resizing, everything works fine, but after resizing, it cycles only on several last elements on the line, instead of cycling through all occurrences

https://github.com/alacritty/alacritty/assets/37012324/34cbb259-9d3d-4de5-83df-f92ca66735ba

### System

OS: Linux, Wayland, Sway
Version: alacritty 0.13.2

### Logs

No logs

chrisduerr:
This fixes an issue where an inline search in the left direction would incorrectly assume that the first cell searched would not contain the `WRAPLINE` flag, causing the second search for the match end to terminate prematurely.

Closes #8060.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Notable changes to the `alacritty_terminal` crate are documented in its
 - Leaking FDs when closing windows on Unix systems
 - Config emitting errors for nonexistent import paths
 - Kitty keyboard protocol reporting shifted key codes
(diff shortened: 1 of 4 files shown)


kchibisov (changes requested):
Needs a CHANGELOG entry.

kchibisov (approved):

PR #8085, opened 2024-07-11
Make font size step configurable

lordi:
For me, the hardcoded delta for font size increase (1px) always felt too small. A quick zoom in, e.g., during a screencast, currently requires too many button presses. For comparison, check a browser increases font size on Ctrl-+. For me, alacritty should behave the same.

This will introduce a new configuration variable for the font size step. The default is still 1px to not break anything. In my config, I would set it to ~4-6px.


chrisduerr:
You can already do what you want without any changes to Alacritty simply by binding the action multiple times.

PR #8093, opened 2024-07-18
Document `window.position` not working on Wayland

kchibisov:
--

I think that won't hurt.


kchibisov:
I've added docs for the other option which is also not working yet on anything other than X11/macOS so it won't confuse users.

chrisduerr (approved):

